PR #2223, opened 2020-06-12
[luci-interpreter]Add L2Normalize Kernel

struss:
This commit add L2Normalize Kernel on luci-interpreter.


compiler/luci-interpreter/src/kernels/L2Normalize.test.cpp
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
```
jinevening, line 3: If we used the same number as TFLite for testing, it would be safe to add TensorFlow copyright.

@lemmaa , Do we have to add the phrase for TensorFlow copyright when using the same test dataset as TFLite?

lemmaa: @jinevening , I'm not sure, but I think it would be nice to specify that. To be precise, the copyright displayed in the corresponding source code should be copied as it is. Would you point-out the original source code for reference?

jinevening: > Would you point-out the original source code for reference?

The test data is the same with [this test](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/l2norm_test.cc#L71-L78) in TFLite.

@struss , Can you add the below phrase at line 3?
`Copyright 2017 The TensorFlow Authors. All Rights Reserved.`

struss: @jinevening Okay, I will add that on all test cases that referring from tensorflow.

compiler/luci-interpreter/src/KernelBuilder.cpp
```diff
@@ -180,6 +181,16 @@ std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleFullyConnected *n
...
 }
 
+std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleL2Normalize *node)
+{
+  assert(node->arity() == 1);
+
+  const Tensor *input = getInputTensor(node->x());
+  Tensor *output = getOutputTensor(node);
+
+  return std::make_unique<kernels::L2Normalize>(input, output);
```
binarman, line 191: `CircleL2Normalize` is inherited from `LuciNodeMixin<LuciNodeTrait::FusedActFunc>`, so we have `fusedActivationFunction` attribute for this operator.
Need to support it in kernels, or add assert/throw, so we will not miss.

s-barannikov: Note that TFLite does not support activations other than kNone:
https://github.com/tensorflow/tensorflow/blob/e5bf8de410005de06a7ff5393fafdf832ef1d4ad/tensorflow/lite/kernels/l2norm.cc#L64-L66

struss: Sorry for missing that. I know only kNone is supported, but forget to write param about this. I will make update commit for this. sorry for make confused.

compiler/luci-interpreter/src/kernels/L2Normalize.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+void L2Normalize::evalQuantized() const
+{
+  tflite::L2NormalizationParams op_params{};
+  op_params.input_zero_point = _input->zero_point();
+  tflite::optimized_ops::L2Normalization(op_params, getTensorShape(_input),
+                                         getTensorData<uint8_t>(_input), getTensorShape(_output),
+                                         getTensorData<uint8_t>(_output));
+}
```
s-barannikov, line 80: (optional) These functions are very simle, they could just be inlined.

compiler/luci-interpreter/src/kernels/L2Normalize.h
```diff
@@ -0,0 +1,48 @@
...
+namespace luci_interpreter
+{
+namespace kernels
+{
+
+class L2Normalize : public KernelWithParams<L2NormParams>
+{
+public:
+  L2Normalize(const Tensor *input, Tensor *output, const L2NormParams &params);
+
```
binarman, line 32: @struss 
Sorry, I merged changes that caused conflicts here =_=
See https://github.com/Samsung/ONE/pull/2335/files#diff-fb2aa6e7e47ff3dba307e302d6f5a265R33 for details.

You need to rebase on to master and apply this change:
```suggestion

  std::vector<const Tensor *> getInputTensors() const override { return {_input}; }
  std::vector<Tensor *> getOutputTensors() const override { return {_output}; }
```

struss: Okay! I will add this and update a commit. never mind!

struss:
Oh! something goes wrong! I will make new PR for it. Sorry!

struss:
Sorry for making confused. while rebasing some error occur and I make a New branch from current master and make a new PR #2394.

PR #2225, opened 2020-06-12
[nnfwapi] Use `nnfw_set_tensorinfo`

wateret:
Use `nnfw_set_tensorinfo` rather than `nnfw_applytensorinfo`.
`nnfw_applytensorinfo` will be deprecated in 1.7.0.

They both do the same thing but `nnfw_applytensorinfo`'s name is
misleading and there was a little fix in its signature.

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -382,7 +382,8 @@ NNFW_STATUS nnfw_session::apply_tensorinfo(uint32_t index, nnfw_tensorinfo ti)
   {
     if (isStateInitialized())
     {
-      std::cerr << "Error during apply_tensorinfo : should be run after load_model" << std::endl;
+      std::cerr << "Error during set_input_tensorinfo : should be run after load_model"
+                << std::endl;
       return NNFW_STATUS_ERROR;
     }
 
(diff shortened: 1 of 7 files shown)


runtime/onert/api/src/nnfw_api.cc
```diff
@@ -253,11 +253,11 @@ NNFW_STATUS nnfw_register_custom_op_info(nnfw_session *session, const char *id,
   return session->register_custom_operation(id, info->eval_function);
 }
 
-NNFW_STATUS nnfw_apply_tensorinfo(nnfw_session *session, uint32_t index,
-                                  nnfw_tensorinfo tensor_info)
+NNFW_STATUS nnfw_set_input_tensorinfo(nnfw_session *session, uint32_t index,
+                                      nnfw_tensorinfo tensor_info)
```
hseok-oh, line 257: @wateret Could you explain why this naming change need? This method is not defined and below already have `nnfw_set_input_tensorinfo(nnfw_session, uint32_t, const nnfw_tensorinfo*)`

hseok-oh:
@wateret Please resolve conflict.

runtime/onert/api/src/nnfw_api.cc
wateret: Oops. It shouldn't have been changed. Thanks!

hseok-oh:
@wateret Please resolve conflict

Issue #2222, opened 2020-06-12
[luci] exports wrong output


mhs4670go:
I tried to optimize in-house model. But, there is an error that all the custom ops in the graph are replaced with the same custom ops.

### Before
![image](https://user-images.githubusercontent.com/24587354/84486331-3debd300-acd8-11ea-9596-655f3e01647d.png)

### After
![image](https://user-images.githubusercontent.com/24587354/84486217-0aa94400-acd8-11ea-9bd2-26acb41f4b58.png)

seanshpark:
I don't understand what the problem is. Please explain this to me on-line or off-line.

PR #2226, opened 2020-06-12
[luci] fix custom op export bug

mhs4670go:
This draft fixes bug when luci exports custom op

Related : #2222

### DIFF
--- a/compiler/luci/export/src/CircleExporterUtils.cpp
+++ b/compiler/luci/export/src/CircleExporterUtils.cpp
@@ -90,13 +90,13 @@ namespace luci
 
 uint32_t SerializedModelData::registerBuiltinOpcode(circle::BuiltinOperator builtin_code)
 {
-  auto it = _operator_codes.find(OpCode{builtin_code, ""});
+  auto it = _operator_codes.find(OpCode{builtin_code});
   if (it != _operator_codes.end())
   {
     return it->second;
   }
   auto idx = static_cast<uint32_t>(_operator_codes.size());
-  _operator_codes.emplace(OpCode{builtin_code, ""}, idx);
+  _operator_codes.emplace(OpCode{builtin_code}, idx);
   return idx;
 }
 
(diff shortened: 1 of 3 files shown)


mhs4670go:
This bug occurred because `std::unordered_map<OpCode, std::string> _custom_operator_codes;` variable. When we run `registerCustomOpcode`, its custom code(above `std::string`) was overwritten as it has the same `OpCode`(key duplicated).

mhs4670go:
@seanshpark PTAL:)

compiler/luci/export/src/CircleExporterUtils.cpp
```diff
@@ -90,21 +90,26 @@ namespace luci
 
 uint32_t SerializedModelData::registerBuiltinOpcode(circle::BuiltinOperator builtin_code)
 {
-  auto it = _operator_codes.find(OpCode{builtin_code});
+  auto it = _operator_codes.find(OpCode{builtin_code, ""});
```
seanshpark, line 93: I think we don't need to add `, ""` for `custom_code`

compiler/luci/export/src/CircleExporterUtils.cpp
```diff
@@ -90,21 +90,26 @@ namespace luci
...
   if (it != _operator_codes.end())
   {
     return it->second;
   }
   auto idx = static_cast<uint32_t>(_operator_codes.size());
-  _operator_codes.emplace(OpCode{builtin_code}, idx);
+  _operator_codes.emplace(OpCode{builtin_code, ""}, idx);
```
seanshpark, line 99: same here

compiler/luci/export/src/SerializedData.h
```diff
@@ -29,8 +29,19 @@ namespace luci
 struct OpCode
 {
   circle::BuiltinOperator opcode;
-
-  bool operator==(const OpCode &rhs) const { return opcode == rhs.opcode; }
+  std::string custom_code;
```
seanshpark, line 32: (optional)
```cxx
std::string custom_code{""};
```

PR #2228, opened 2020-06-12
Enable OSX build for onert

chunseoklee:
tested on :
- OSX version : osx mojave 10.14.6
- g++ version :
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/usr/include/c++/4.2.1
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

- libaec 611a94be2488ed549 : manually build and install to /usr/local/
- boost 1.67.0 : manually build(./bootstrap.sh --with-libraries="program_options") and install to /usr/local


chunseoklee:
On my 2017 macbook(m3 processor) does not finish build without PARALLEL_BUILD=0.

glistening:
@chunseoklee I tried to `brew install boost` instead of building by myself. However, it saids `program_options` is missing. Now, I know why you've built `boost` by yourself.

tests/tools/nnpackage_run/src/nnpackage_run.cc
```diff
@@ -34,6 +34,10 @@
 #include <unordered_map>
 #include <vector>
 
+#ifdef __clang__
+#include <libgen.h>
```
glistening, line 38: @chunseoklee I've found that there are two version of `basename`. One gets `const char*`, another gets `char*`. Mac OS only supports latter, which requires to include `<libgen.h>`. I've thought using a same version of `basename`, however, it requires to make a copy of string. So. your approach is reasonable.

tests/tools/tflite_run/src/tflite_run.cc
```diff
@@ -356,8 +359,8 @@ int main(const int argc, char **argv)
     std::string model_basename;
     std::string backend_name = default_backend_cand;
     {
-      model_basename = basename(args.getTFLiteFilename().c_str());
-      // should remove extension(such as .tflite)
+      std::vector<char> vpath(args.getTFLiteFilename().begin(), args.getTFLiteFilename().end() + 1);
```
glistening, line 362: @chunseoklee It could be invalid in c++03, which does not guarantee the buffer for string is continuous and terminated with null like c++11 or later. However, it is 2020, I don't think we need to support c++03. Or we may explicitly push back `\0` after copying `begin()` to `end()`.

tests/tools/tflite_run/src/tflite_run.cc
```diff
@@ -37,6 +37,9 @@
 #include <algorithm>
 #include <vector>
 
+#ifdef __clang__
+#include <libgen.h>
+#endif
```
wateret, line 42: As we copy and pass `char *` below instead of `const char *`,

```suggestion
#include <libgen.h>
```

wateret:
> @chunseoklee I tried to `brew install boost` instead of building by myself. However, it said `program_options` is missing. Now, I know why you've built `boost` by yourself.

With `brew install boost` here's what I've tried on nnpackage_run.

```cmake
nnfw_find_package(Boost REQUIRED)
```

I found this does not work. Didn't figure out why. When I replace this line with

```cmake
find_package(Boost 1.58.0 QUIET COMPONENTS log program_options filesystem system)

# And also
target_link_libraries(nnpackage_run Boost::program_options) # ${BOOST_PROGRAM_OPTIONS_LIBRARY}=Boost::program_options
```

Then it works.

So there is two things I would like to ask:

1. Why `nnfw_find_package` does not work?
2. Why do we have this if branch? (For aarch64? but which OS?)
```cmake
if(BUILD_BOOST)
  # We have to use Boost::program_options, instead of boost_program_optinos
  # Boost::program_options provides the full path for our own built boost.
  target_link_libraries(nnpackage_run Boost::program_options)
else()
  # We cannot use `Boost::program_options` on aarch64
  # because it uses boost 1.65.1, which requires cmake >= 3.9.3
  # to identify Boost::program_options imported target correctly,
  # However, it uses cmake 3.5. (old version).
  target_link_libraries(nnpackage_run boost_program_options)
endif()
```

glistening:
@wateret 

> 1. Why nnfw_find_package does not work?
> 2. Why do we have this if branch? (For aarch64? but which OS?)

It depends on both cmake version and boost version whether `Boost::program_options` could be used or not. It is not dependent to architecture. It would be nice if we can write the condition that would work for various condition.

wateret:
> `find_package` (that `nnfw_find_package` calls) succeed. But some system which uses old cmake version, `Boost::program_options` has nothing.

I checked CMake 3.19(or 3.18) on My Mac. And `infra/nnfw/cmake/BoostConfig.cmake` is not called and I did not figure out why. But directly calling `find_package` succeeds.

> It depends on both cmake version and boost version whether `Boost::program_options` could be used or not. It is not dependent to architecture. It would be nice if we can write the condition that would work for various condition.

`${Boost_PROGRAM_OPTIONS_LIBRARY}` might work for both cases? (Just my rough guess)

glistening:
@wateret 

> `${Boost_PROGRAM_OPTIONS_LIBRARY}` might work for both cases? (Just my rough guess)

I am not sure before testing. It is the reason I've added `if (BUILD_BOOST)` condition while keeping original statement in `else`, which has been working for our supported platform and CI.

You may test including OS X. I don't have mac in work, I can test OS X only in my home. :)

wateret:
I checked `${Boost_PROGRAM_OPTIONS_LIBRARY}` works on my Mac(Of course only if `nnfw_find_package` is succeeded) but did not try for Android.

wateret:
I tried building on my macbook pro at office (Catalina), I still see this:

```
ld: library not found for -lboost_program_options
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [tests/tools/tflite_run/tflite_run] Error 1
make[1]: *** [tests/tools/tflite_run/CMakeFiles/tflite_run.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

I have boost lib installed via brew.

wateret:
I checked that @chunseoklee built Boost himself.

wateret:
Why don't we merge this first? I would like to continue on this work.

wateret:
> find_package (that nnfw_find_package calls) succeed. But some system which uses old cmake version, Boost::program_options has nothing.

From my environment `nnfw_find_package` didn't work and replacing it with `find_package(boost 1.58.0 ...)`, it was successful

glistening:
> From my environment nnfw_find_package didn't work and replacing it with find_package(boost 1.58.0 ...), it was successful

I don't know why. On my mac envionment, `nnfw_find_package` + using `Boost::program_options` work. There are so many possible combination. If your solution works for all officially supported platforms (`x86_64`, `arm ubuntu`, `aarch64 ubuntu`, `arm tizen`, `android` + coming another one), I am okay. :)

PR #2212, opened 2020-06-12
DRAFT fix for real models with options

seanshpark:
on-going draft to support real models with user options

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #2194, opened 2020-06-11
Compile FE: problems with in-house models


seanshpark:
Let's gather problems that came from our in-house models

seanshpark:
Problems
- [ ] GraphBuilder validation Errors with UNPACK : input shape is unknown
- [ ] GraphBuilder validation Errors with GREATER output type is not BOOL

seanshpark:
TODO
- [ ] Introduce an option for the user to turn off validations.
   - Validations are useful and I don't to remove the codes.

seanshpark:
done for now

PR #2229, opened 2020-06-14
[luci] Introduce env

seanshpark:
This will introduce luci env module that provides compile execution optional settings

### DIFF
--- a/compiler/luci/env/src/UserSettings.cpp
+++ b/compiler/luci/env/src/UserSettings.cpp
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


seanshpark:
For #2194, from draft #2212

compiler/luci/env/src/UserSettings.cpp
```diff
@@ -0,0 +1,73 @@
...
+  switch (key)
+  {
+    case Key::MuteWarnings:
+      _MuteWarnings = value;
+      break;
+    case Key::DisableValidation:
+      _DisableValidation = value;
+      break;
+    default:
+      break;
```
llFreetimell, line 44: (Optional) How about throwing `runtime_error` if invalid key is given?

compiler/luci/env/src/UserSettings.cpp
```diff
@@ -0,0 +1,73 @@
...
+bool UserSettingsImpl::get(const Key &key) const
+{
+  switch (key)
+  {
+    case Key::MuteWarnings:
+      return _MuteWarnings;
+    case Key::DisableValidation:
+      return _DisableValidation;
+    default:
+      break;
```
llFreetimell, line 57: (optional) ditto

compiler/luci/env/src/UserSettings.cpp
seanshpark: seems reasonable.

PR #2231, opened 2020-06-14
[circle2circle] Revise help message

seanshpark:
This will revise help message to show category of optimization


seanshpark:
For #2194, from draft #2212

PR #2232, opened 2020-06-15
[luci] Revise log to show warnings with settings

seanshpark:
This will revise log to show warnings
- WARN will be always shown but can be muted with settings


seanshpark:
For #2194, from draft #2212

PR #2233, opened 2020-06-15
[circle2circle] Support two settings

seanshpark:
This will revise circle2circle to support two settings mute warnings and disable validation


seanshpark:
For #2194, from draft #2212

seanshpark:
Needs rebase to use `luci/env`

PR #2235, opened 2020-06-15
[luci] Fix comment of Log with WARN

seanshpark:
This will fix comment of Log with WARN
- WARN is now available


seanshpark:
For #2194, from draft #2212

PR #2237, opened 2020-06-15
[luci] Revise import/export Unpack

seanshpark:
This will revise import and export of Unpack with disable validation and add warning messages


seanshpark:
For #2194, from draft #2212

compiler/luci/import/src/Nodes/CircleUnpack.cpp
```diff
@@ -100,7 +124,6 @@ void CircleUnpackGraphBuilder::build(const circle::OperatorT &op,
   node->axis(options->axis);
 
   assert(outputs.size() > 0);
-  assert(int32_t(outputs.size()) == options->num);
```
seanshpark, line 103: this is moved to `validate`

compiler/luci/import/src/Nodes/CircleUnpack.cpp
```diff
@@ -27,19 +30,40 @@ namespace luci
...
     return false;
 
-  if (static_cast<int32_t>(outputs.size()) != options->num)
-    return false;
+  // NOTE real models may have mismatch
+  if (!settings->get(luci::UserSettings::Key::DisableValidation))
+  {
+    if (static_cast<int32_t>(outputs.size()) != options->num)
+      return false;
+  }
```
mhs4670go, line 49: Maybe this line is unnecessary because its condition is same with 63 line.

compiler/luci/import/src/Nodes/CircleUnpack.cpp
```diff
@@ -27,19 +30,40 @@ namespace luci
...
 
   if (options->num < 0)
     return false;
 
+  if (int32_t(outputs.size()) != options->num)
```
mhs4670go, line 54: (optional) maybe `static_cast` like above?

compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -1629,9 +1643,15 @@ void OperationExporter::visit(luci::CircleUnpack *node)
         break;
       }
     }
+    // NOTE real models may not use all of the outputs
     if (!found)
     {
-      INTERNAL_EXN("Invalid Unpack output");
```
seanshpark, line 1634: as a graph, not using all of the outputs is not an error.
using assert is OK but throw is not appropriate.

PR #2239, opened 2020-06-15
[luci] Revise import Greater Op

seanshpark:
This will revise import of Greater Op to option to disable dtype validation and show warnings for this

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleGreater.cpp
+++ b/compiler/luci/import/src/Nodes/CircleGreater.cpp
@@ -18,6 +18,9 @@
 
 #include <luci/IR/Nodes/CircleGreater.h>
 
+#include <luci/UserSettings.h>
+#include <luci/Log.h>
+
 #include <loco.h>
 
 namespace luci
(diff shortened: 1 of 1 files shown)


seanshpark:

compiler/luci/import/src/Nodes/CircleGreater.cpp
```diff
@@ -39,8 +46,18 @@ bool CircleGreaterGraphBuilder::validate(const ValidateArgs &args) const
   if (tensors[inputs[0]]->type != tensors[inputs[1]]->type)
     return false;
 
+  // NOTE: real models do have output dtype NOT BOOL
```
llFreetimell, line 49: Sentence is little bit confusing... @.@ Which one of them?

- Real models often have output dtype, which is NOT BOOL.
- Real models actually do not have output dtype, which is NOT BOOL

seanshpark: > do have

I thought output dtype should be boolean, but it is possible that models can have not boolean. So. `do have` is emphasis  on this.

PR #2204, opened 2020-06-12
Revise libbenchmark for nnpackage_run and tflite_run

YongseopKim:
- Expose the only Phases and Result classes
- Phases: a role for acting benchmark such as latency and memory
- Result: the result of benchmark

Signed-off-by: Yongseop Kim <yons.kim@samsung.com>

PR #2243, opened 2020-06-15
fix bug on runner

YongseopKim:
- #2204 cause this bug
  - without warmup option, it doesn't measure memory
- change the default warmup value
  - originally it was 0 but, on the past, it actually does warmup once for measuring memory. therefore it has nothing to be changed

Signed-off-by: Yongseop Kim <yons.kim@samsung.com>


YongseopKim:
PTAL @chunseoklee and @periannath

YongseopKim:
ping @Samsung/nnfw_committers

Issue #2166, opened 2020-06-11
Compiler FE: change linux only header


mhs4670go:
For #2074, we need to build `tflchef` `circlechef` `circle-verify` etc. But, they include some header file that exists only in linux(e.g. `sys/mman.h`)

Let's change this header to something compatible with Windows.

mhs4670go:
@seanshpark Above modules includes `sys/*` headers for loading `.tflite` or `.circle`. 

How about changing the way it loads to `ONE/compiler/tflite2circle/src/TFLModel.cpp` sytle?

It comes from [here](https://google.github.io/flatbuffers/flatbuffers_guide_use_cpp.html).

seanshpark:
OK, let's try not to use `sys/mman.h` file

PR #2247, opened 2020-06-15
[circlechef] Revise the way it loads file

mhs4670go:
This commit revises the way it loads file in circlechef

Related : #2166


compiler/circlechef/tools/reverse/Driver.cpp
```diff
@@ -31,19 +33,9 @@ int entry(int argc, char **argv)
...
   // Load TF lite model from a circle file
-  std::unique_ptr<circlechef::RawModel> rawmodel = circlechef::load_circle(argv[1]);
-  if (rawmodel == nullptr)
-  {
-    std::cerr << "ERROR: Failed to load circle '" << argv[1] << "'" << std::endl;
-    return 255;
-  }
-
-  const circle::Model *tflmodel = rawmodel->model();
-  if (tflmodel == nullptr)
```
seanshpark, line 42: can you add this check?

PR #2248, opened 2020-06-15
[onert] Fix assert checking error by optional input

hseok-oh:
Fix invalid assert checking by optional input (fully-connected bias)

### DIFF
--- a/runtime/onert/backend/cpu/ops/OperationUtils.cc
+++ b/runtime/onert/backend/cpu/ops/OperationUtils.cc
@@ -82,7 +82,7 @@ void GetQuantizedConvolutionMultiplier(const Tensor *input, const Tensor *filter
                                        const Tensor *bias, const Tensor *output, double *multiplier)
 {
   const double input_product_scale = input->data_scale() * filter->data_scale();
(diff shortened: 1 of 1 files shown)


runtime/onert/backend/cpu/ops/OperationUtils.cc
```diff
@@ -82,7 +82,7 @@ void GetQuantizedConvolutionMultiplier(const Tensor *input, const Tensor *filter
                                        const Tensor *bias, const Tensor *output, double *multiplier)
 {
   const double input_product_scale = input->data_scale() * filter->data_scale();
-  const double bias_scale = bias->data_scale();
+  const double bias_scale = (bias != nullptr) ? bias->data_scale() : input_product_scale;
```
glistening, line 85: @hseok-oh What version of tensorflow are you using? I am looking into `v2.2`, assertion check is not checked if `bias` is `null`. Anyway, this code looks correct because it will pass assert when `bias` is `null`.

hseok-oh: This PR is to fix svace warning.

PR #2003, opened 2020-06-08
[DRAFT][onert] Adjust calling order of `StaticInferer` in multiple subgraphs

ragmani:
For issue #1999

This draft adjusts calling order of `StaticInferer` in multiple subgraphs

Signed-off-by: ragmani <ragmani0216@gmail.com>

Issue #1999, opened 2020-06-08
[onert] Adjust calling order of `StaticInferer` in multiple subgraphs


ragmani:
Now `StaticInferer` calls inferring shapes for each independent `Graph`.
However, subgraph's inputs can be inferred by controlflow operation of parent subgraph. To infer actual shape of inputs of subgraphs, we make `Staticinferer` calls inferring shapes in the same order as actual execution.

ragmani:
Done.

PR #2249, opened 2020-06-15
[onert] Support multiple subgraph with While op in StaticInferer.

ragmani:
For issue : #1999
Draft PR : #2003

This commit makes StaticInferer supports multiple subgraph with While op.

### DIFF
--- a/runtime/onert/core/src/util/shapeinf/While.cc
+++ b/runtime/onert/core/src/util/shapeinf/While.cc
@@ -23,18 +23,128 @@ namespace shape_inference
 
 void StaticInferer::visit(const ir::operation::While &op)
 {
-  for (const auto input_idx : op.getInputs())
+  auto &cond_graph = _lowered_subgs.at(op.param().cond_subg_index)->graph();
+  auto &body_graph = _lowered_subgs.at(op.param().body_subg_index)->graph();
+  const auto inputs = op.getInputs();
+  const auto &outputs = op.getOutputs();
+
+  // re-sizing input shapes of then subgraph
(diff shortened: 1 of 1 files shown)


runtime/onert/core/src/util/shapeinf/While.cc
```diff
@@ -23,18 +23,120 @@ namespace shape_inference
...
+        (cond_input.shape() != body_output.shape()))
+    {
+      check_unpredictable_dynamic = true;
+      break;
+    }
+  }
+
+  if (check_unpredictable_dynamic)
+  {
+    // Set inputs of body subgraph
```
hyunsik-yoon, line 94: when only `i`th input of cond and while was _unpredictable_,
the code below set _all_ inputs dynamic.
Could you explain the reason?

ragmani: While operation makes body subgraph looping. We cannot predict how many the subgraph loops and which outputs change to dynamic by the loop. So I added the code to set all inputs to dynamic if one of outputs change to dynamic.

runtime/onert/core/src/util/shapeinf/While.cc
```diff
@@ -23,18 +23,128 @@ namespace shape_inference
...
+    for (const auto &input_index : body_inputs)
+    {
+      auto &input = body_graph.operands().at(input_index);
+      if (!input.isConstant())
```
ragmani, line 98: This was added by force pushing.

hseok-oh:
@hyunsik-yoon PTAL

PR #2251, opened 2020-06-15
[libbenchmark] fix bug on peakMemory

YongseopKim:
- caused by #2204
- handle WARMUP as EXECUTE
  - it's because measuring memory on EXECUTE takes overhead a quite


YongseopKim:
PTAL @chunseoklee and @periannath

PR #2256, opened 2020-06-15
[luci] Introduce ResolveCustomOpAddPass

mhs4670go:
This commit introduces ResolveCustomOpAddPass to luci

Draft: #2227

### DIFF
--- a/compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
@@ -28,22 +28,16 @@ namespace
 // NOTE This function assumes there is only one BroadcastTo node among its inputs.
 int32_t get_broadcastTo_index_among_inputs_of(luci::CircleCustom *cop)
 {
-  auto input0 = dynamic_cast<const luci::CircleCustomOut *>(cop->inputs(0));
-  if (input0)
+  for (uint32_t idx = 0; cop->numInputs(); idx++)
   {
-    auto broadcastTo = dynamic_cast<luci::CircleCustom *>(input0->input());
-    assert(broadcastTo);
-    if (broadcastTo->custom_code() == "BroadcastTo")
-      return 0;
-  }
-
-  auto input1 = dynamic_cast<const luci::CircleCustomOut *>(cop->inputs(1));
-  if (input1)
-  {
-    auto broadcastTo = dynamic_cast<luci::CircleCustom *>(input1->input());
-    assert(broadcastTo);
-    if (broadcastTo->custom_code() == "BroadcastTo")
-      return 1;
+    auto input = dynamic_cast<const luci::CircleCustomOut *>(cop->inputs(idx));
+    if (input)
+    {
+      auto broadcastTo = dynamic_cast<luci::CircleCustom *>(input->input());
+      assert(broadcastTo);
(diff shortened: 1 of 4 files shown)


compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -0,0 +1,109 @@
...
+ *                   \           /
+ *                    \         /
+ *                    [CircleAdd]
+ */
+bool resolve_with_BroadcastTo(luci::CircleCustom *addv2)
+{
+  const luci::CircleCustomOut *input1 =
+      dynamic_cast<const luci::CircleCustomOut *>(addv2->inputs(1));
+
+  if (input1 != nullptr)
```
seanshpark, line 47: what if `addv2->inputs(0)` is `BroadcastTo` ?

mhs4670go: @seanshpark 
Well, I followed our in-house model because already this pass is highly dependent on our model. If some model has `AddV2` which has `BroadcastTo` as first argument, I'll implement it then.

seanshpark: How about introduce a function that returns 0 or 1 or -1 as idx
- 0 : `BroadcastTo` is input(0)
- 1: `BroadcastTo` is input(1)
- -1: there is no `BroadcastTo` input
and then we can use this value for `BroadcastTo` and `1 - idx` as opposite.

mhs4670go:
@seanshpark 
PTAL:)

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -0,0 +1,131 @@
...
+  auto input0 = dynamic_cast<const luci::CircleCustomOut *>(cop->inputs(0));
+  if (input0)
+  {
+    auto broadcastTo = dynamic_cast<luci::CircleCustom *>(input0->input());
+    assert(broadcastTo);
+    if (broadcastTo->custom_code() == "BroadcastTo")
+      return 0;
+  }
+
+  auto input1 = dynamic_cast<const luci::CircleCustomOut *>(cop->inputs(1));
```
seanshpark, line 40: (optional) we can loop idx = 0; idx < 2; ++idx

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -0,0 +1,131 @@
...
+  if (broadcastTo_idx == -1)
+    return false;
+
+  auto input = dynamic_cast<const luci::CircleCustomOut *>(addv2->inputs(broadcastTo_idx));
+  auto broadcastTo = dynamic_cast<luci::CircleCustom *>(input->input());
+
+  auto add = addv2->graph()->nodes()->create<luci::CircleAdd>();
+  add->fusedActivationFunction(luci::FusedActFunc::NONE);
+  add->x(addv2->inputs(1 - broadcastTo_idx));
+  add->y(broadcastTo->inputs(0));
```
seanshpark, line 80: `0` -> `broadcastTo_idx`

mhs4670go:
@seanshpark 
PTAL:)

PR #2257, opened 2020-06-15
[nnfwapi] Move HowToUseNNFWAPI.md

wateret:
Move HowToUseNNFWAPI.md to the new dir from the legacy dir.


docs/howto/how-to-use-nnfw-api.md
```diff
@@ -1,14 +1,16 @@
-# Prepare nnpackage
+# How to use NNFW API
```
lemmaa, line 1:
```suggestion
# How to Use NNFW API
```
Trivial, but to make it the same as the other cases.

PR #2258, opened 2020-06-15
[onert] Fix Select V2 operation broadcasting

hseok-oh:
Fix to brodcasting between input-true and input-false


KimDongEon:
LGTM

PR #2267, opened 2020-06-15
[circle2circle] add ResolveCustomOpAddPass option

mhs4670go:
This commit adds ResolveCustomOpAddPass option to circle2circle


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -89,6 +89,10 @@ int entry(int argc, char **argv)
     options->enable(Algorithms::FuseInstanceNorm);
     return 0;
   };
+  argparse["--resolve_customop_add"] = [&options](const char **) {
```
seanshpark, line 92: Can you please add also to the showing help message?

mhs4670go:
@seanshpark PTAL:)

PR #2268, opened 2020-06-15
[luci] fix for loop condition in ResolveCustomOpAddPass

mhs4670go:
This commit fixes for-loop condition in ResolveCustomOpAddPass

### DIFF
--- a/compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
@@ -20,7 +20,7 @@
 
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/AttrFusedActFunc.h>
-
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -20,15 +20,15 @@
 
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/AttrFusedActFunc.h>
-
+#include <iostream>
```
seanshpark, line 23: Please add an empty space after this

seanshpark: Is this needed?

mhs4670go:
@seanshpark PTAL:)

PR #2276, opened 2020-06-16
[luci-interpreter] add missing header

mhs4670go:
This commit adds missing header to luci-interpreter


jinevening:
Why did you add the header file? Is there any benefit? (Just for curiosity. I'm not a cpp expert)

mhs4670go:
@jinevening This missing file causes an error during Windows build:)

binarman:
@jinevening 
`std::string` is used on line 92/93, so `string` header is needed.

Probably it didn't fail under linux because `string` header is included in some other std lib file.

PR #2279, opened 2020-06-16
[nnpkg_run] support building nnpkg_run without hdf5

glistening:
`nnpackage_run` could be built without hdf5, of course,
it will not be capable of loading and dumping.

If `hdf5` is not found, it will silently build `nnpackage_run`
without loading/dumping functionality.

### DIFF
--- a/tests/tools/nnpackage_run/src/args.cc
+++ b/tests/tools/nnpackage_run/src/args.cc
@@ -37,8 +37,10 @@ void Args::Initialize(void)
     ("help,h", "Print available options")
     ("version", "Print version and exit immediately")
     ("nnpackage", po::value<std::string>()->required())
+#if defined(ONERT_HAVE_HDF5) && ONERT_HAVE_HDF5 == 1
     ("dump,d", po::value<std::string>()->default_value(""), "Output filename")
     ("load,l", po::value<std::string>()->default_value(""), "Input filename")
+#endif
     ("num_runs,r", po::value<int>()->default_value(1), "The number of runs")
     ("warmup_runs,w", po::value<int>()->default_value(1), "The number of warmup runs")
     ("gpumem_poll,g", po::value<bool>()->default_value(false), "Check gpu memory polling separately")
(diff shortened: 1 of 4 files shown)


glistening:
@chunseoklee As you asked last Thur or Fri, `nnpackage_run` could be built without hdf5. Of course, it cannot load or dump if `hdf5` does not exist. 

<sub><b>no load and no dump </b></sub>
```
$ Product/x86_64-linux.debug/out/bin/nnpackage_run --help
nnpackage_run

Usage: Product/x86_64-linux.debug/out/bin/nnpackage_run path to nnpackage root directory [<options>]


General options:
  -h [ --help ]                   Print available options
  --version                       Print version and exit immediately
  --nnpackage arg
  -r [ --num_runs ] arg (=1)      The number of runs
  -w [ --warmup_runs ] arg (=1)   The number of warmup runs
  -g [ --gpumem_poll ] arg (=0)   Check gpu memory polling separately
  -m [ --mem_poll ] arg (=0)      Check memory polling
  -p [ --write_report ] arg (=0)  Write report
                                  {exec}-{nnpkg}-{backend}.csv will be generated.
                                  e.g. nnpackage_run-UNIT_Add_000-acl_cl.csv.
                                  {nnpkg} name may be changed to realpath if you use symbolic-link.
```

If you don't have `hdf5` (especially on android), you don't need to prepare `hdf5`.

tests/tools/nnpackage_run/src/args.cc
```diff
@@ -37,8 +37,10 @@ void Args::Initialize(void)
     ("help,h", "Print available options")
     ("version", "Print version and exit immediately")
     ("nnpackage", po::value<std::string>()->required())
+#if defined(ONERT_HAVE_HDF5) && ONERT_HAVE_HDF5 == 1
```
periannath, line 40: (Optional) Should this macro have value 1? It seems defining macro w/o value would be enough.

glistening: @periannath I like this style. Because I and someone like me expected `USE_NNAPI=0` would not use nnapi. However, it checked only whether `USE_NNAPI` is defined or not. Thus, it worked same to `USE_NNAPI=1`.

Issue #2260, opened 2020-06-15
Compiler FE: While as network output


seanshpark:
When While Op is used as the network output but not all outputs are used current Compiler fontend doesn't work as expected.

From `While_001` recipe
```
operation {
  type: "While"
  input: "ifm1"
  input: "ifm4"
  output: "ofm1"
  output: "ofm2"
  while_options {
    body_subgraph_index: 2
    cond_subgraph_index: 1
  }
}
input: "ifm1"
input: "ifm4"
output: "ofm1"
#output: "ofm2" <--- ofm2 is not used as an output
name: "Main"
```
this will not work.

seanshpark:
To do...
- [ ] Siblings of multiple outputs need to be taken care
- [ ] ...

seanshpark:
Shape of this While output is set as int32[1,0]
- from the condition of shapes, this should be same as WHILE_COND network input
- which happens to be int32[], and this is not same as int32[1,0]

What should we do with this?

seanshpark:
what to do with `Siblings of multiple outputs need to be taken care` ?
- nodes are removed from RemoveDeadNode pass
   - prevent this
- ...

seanshpark:
done

PR #2284, opened 2020-06-16
[res] Unpack and While for testing

seanshpark:
This will introduce test materials that have valid unknown shape and unused multiple outputs


seanshpark:
For #2260 from draft #2261

mhs4670go:
@seanshpark maybe needs to fix sign-off :)

seanshpark:
I think I need to manually fix sign-off inside commit message when merge time.

PR #2287, opened 2020-06-16
[luci] Support VERBOSE in logging

seanshpark:
This will enable support of VERBOSE in logging

### DIFF
--- a/compiler/luci/log/src/Log.cpp
+++ b/compiler/luci/log/src/Log.cpp
@@ -38,6 +38,11 @@ template <> bool safecast<bool>(const char *s, const bool &value)
   return (s == nullptr) ? value : (std::stoi(s) != 0);
 }
 
+template <> int safecast<int>(const char *s, const int &value)
+{
+  return (s == nullptr) ? value : std::stoi(s);
+}
+
 } // namespace
 
 //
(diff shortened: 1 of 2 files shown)


seanshpark:
For #2260 from draft #2261

compiler/luci/log/src/Log.cpp
```diff
@@ -65,6 +70,10 @@ LoggerConfig::LoggerConfig()
 
   // Turn on info logging if LUCI_LOG is set as non-zero value
   _show_info = safecast<bool>(std::getenv("LUCI_LOG"), false);
+
+  // Turn on verbose logging if LUCI_LOG is set to some level
+  // VERBOSE(l, 1) will be visible with LUCI_LOG=2 and VERBOSE(l, 2) with LUCI_LOG=3 and so on
```
mhs4670go, line 75: I'm just confused with the number. Isn't it `VERBOSE(l, 1) will be visible with LUCI_LOG=1`?

seanshpark: `VERBOSE(l, level)` will be visible for higher than the `level`, that I've discovered while running with this. that is `VERBOSE(l, 0)` is same(?) as `INFO(l)`

PR #2288, opened 2020-06-16
[luci] Introduce IndexTensorOutputs

seanshpark:
This will introduce IndexTensorOutputs in GraphBuilderContext to cache Tensor index that are outputs of operators


seanshpark:
For #2260 from draft #2261

PR #2289, opened 2020-06-16
[luci] Revise CircleCustom import

seanshpark:
This will revise CircleCustom import to remove unnecessary enroll of node with index


seanshpark:
For #2260 from draft #2261

Issue #2112, opened 2020-06-10
nnpackage_run should support model with dynamic shape


chunseoklee:
nnpackage_run uses model's input shapes during inference. 
To test inference model with dynamic shape, nnpackag_run should provide a way to set various input shapes before inference. This can be done by using `nnfw_apply_tensorinfo` in nnpackage_run.

chunseoklee:
cc @hyunsik-yoon

hyunsik-yoon:
This is done with #2290.

PR #2290, opened 2020-06-16
[onert] make nnpackage_run receive tensor shape

hyunsik-yoon:
This makes `nnpakcage_run` receive new tensor shapes to call `nnfw_set_input_tensorinfo()`.

To run some model, calling `nnfw_set_input_tensorinfo()` is required and with this commit, `nnpackage_run()` receives new tensor shape from command.

parent issue: #2112

Note: `--shape_compile` works but `--shape_exec` does not work because `nnfw_input_tensorinfo()` is not yet fully implemented.

### DIFF
--- a/tests/tools/nnpackage_run/src/args.cc
+++ b/tests/tools/nnpackage_run/src/args.cc
@@ -16,10 +16,45 @@
 
 #include "args.h"
 
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/foreach.hpp>
 #include <iostream>
 
 namespace nnpkg_run
 {
+
+// param shape_str is a form of, e.g., "[1, [2, 3], 3, []"
+void setShape(TensorShapeMap &shape_map, const std::string &shape_str)
+{
+  using boost::property_tree::ptree;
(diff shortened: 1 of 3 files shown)


tests/tools/nnpackage_run/src/args.cc
```diff
@@ -16,11 +16,46 @@
 
 #include "args.h"
 
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/foreach.hpp>
```
glistening, line 21: @hyunsik-yoon I and other members (e.g. @wateret) would like to make `nnpackage_run` boost-free. Thus, I removed `boost::filesystem` dependency. `program_options` are the only dependency. I would like to implement without further `boost`.

wateret: What about using jsoncpp? we already have it.

hyunsik-yoon: OK. I will fix it using jsoncpp.

wateret:
@hyunsik-yoon PTAL at #2325 before switching to jsoncpp.

hyunsik-yoon:
replaced by #2404

PR #2170, opened 2020-06-11
[onert] Add a condition and converting function in shape inference of reshape

ragmani:
For issue #2075

This commit adds a condition and converting function in shape inference of reshape
  - Add a function converting shape buffer to ir::Shape
  - Add a the condition whether 2nd param and output shape are the same

Signed-off-by: ragmani <ragmani0216@gmail.com>

PR #2292, opened 2020-06-16
[onert] Treat shape tensor as flat tensor in shape inference of reshape

periannath:
- This commit treats shape tensor as flat tensor in shape inference of reshape
  - Shape inference for reshape can't accept shape tensor with rank 2 or higher
  - Treat shape tensor as flat tensor to accept rank 2 or higher shape tensor

---

Related PR : #2170

### DIFF
--- a/runtime/onert/core/src/util/shapeinf/Reshape.cc
+++ b/runtime/onert/core/src/util/shapeinf/Reshape.cc
@@ -91,8 +91,7 @@ void StaticInferer::visit(const ir::operation::Reshape &op)
     const auto *shape_buf = reinterpret_cast<const int32_t *>(shape.data()->base());
     assert(shape_buf);
 
-    assert(shape.shape().rank() == 1);
-    new_shape = convertShape(shape_buf, shape.shape().dim(0), input.shape().num_elements());
+    new_shape = convertShape(shape_buf, shape.shape().num_elements(), input.shape().num_elements());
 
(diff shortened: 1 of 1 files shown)


periannath:
After offline disscussion with @ragmani, I'll update this PR to support 2D or higher tensors as reshape input. (Ignoring dimension with size 1)

periannath:
@ragmani @hyunsik-yoon I've updated this PR. PTAL

runtime/onert/core/src/util/shapeinf/Reshape.cc
```diff
@@ -91,8 +91,7 @@ void StaticInferer::visit(const ir::operation::Reshape &op)
     const auto *shape_buf = reinterpret_cast<const int32_t *>(shape.data()->base());
     assert(shape_buf);
 
-    assert(shape.shape().rank() == 1);
-    new_shape = convertShape(shape_buf, shape.shape().dim(0), input.shape().num_elements());
+    new_shape = convertShape(shape_buf, shape.shape().num_elements(), input.shape().num_elements());
```
hyunsik-yoon, line 94: Could you also change the name of second param (`const int32_t rank`) of `convertShape()` to align with the real meaning?

periannath: @hyunsik-yoon Thanks. I've updated.

PR #2298, opened 2020-06-16
[luci] MultiOutputDetector and allocateCircleTensorInfo

seanshpark:
This will introduce MultiOutputDetector class to revise storing of multiple output tensors
also allocateCircleTensorInfo method that actually store CircleTensorInfo


seanshpark:
For #2260 from draft #2261

seanshpark:

PR #2303, opened 2020-06-16
[luci] Revise copy_shape_dtype

seanshpark:
This will revise copy_shape_dtype method to handle nodes that did not run shape/dtype inference
- some multiple output nodes may not be connected to the output
- for this case shape/dtype inference does not run
- shape/dtype of that node will be used from the import code or as unknown


seanshpark:
For #2260 from draft #2261

seanshpark:

PR #2305, opened 2020-06-16
[luci] Revise DeadNodeQueryServiceImpl

seanshpark:
This will revise DeadNodeQueryServiceImpl to skip virtual nodes for multiple outputs when the real node is active


seanshpark:
For #2260 from draft #2261

PR #2072, opened 2020-06-10
[luci] Enable test of logex codes

seanshpark:
This will enable test of logex codes to get higher test coverage

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #2309, opened 2020-06-16
[luci] Revise readverify with higher LUCI_LOG

seanshpark:
This will revise readverify script with higher LUCI_LOG to include VERBOSE codes
- 100 is chosen to some huge value that value 100 itself is not important
- this is needed as the purpose of widen testing


seanshpark:
Related #2072 and #2287

seanshpark:

PR #2310, opened 2020-06-16
[luci] Revise import of While

seanshpark:
This will revise import of While to set also shape and dtype from the file


seanshpark:
For #2260 from draft #2261

seanshpark:

PR #2313, opened 2020-06-16
[luci] Revise shape infer log to verbose

seanshpark:
This will revise shape infer method to use verbose for logging


seanshpark:
For #2260 from draft #2261

PR #2314, opened 2020-06-16
[luci] Revise graph validate_shape_dtype log format

seanshpark:
This will revise graph validate_shape_dtype method log format to include "[luci]" prefix


seanshpark:
For #2260 from draft #2261

PR #2317, opened 2020-06-16
[onert] Support BCQ's new feature

dayo09:
- Add cluster as an input

### DIFF
--- a/runtime/onert/core/include/ir/operation/BCQFullyConnected.h
+++ b/runtime/onert/core/include/ir/operation/BCQFullyConnected.h
@@ -37,7 +37,8 @@ public:
     INPUT = 0,
     WEIGHTS_SCALES,
     WEIGHTS_BINARY,
-    BIAS
+    BIAS,
+    WEIGHTS_CLUSTERS,
   };
 
   struct Param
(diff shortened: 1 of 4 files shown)


runtime/onert/core/include/ir/operation/BCQFullyConnected.h
```diff
@@ -37,7 +37,8 @@ class BCQFullyConnected : public Operation
     INPUT = 0,
     WEIGHTS_SCALES,
     WEIGHTS_BINARY,
-    BIAS
+    BIAS,
+    WEIGHTS_CLUSTER,
```
llFreetimell, line 41: Actually, I am receiving as **CLUSTERS**, not `CLUSTER` :(

runtime/onert/core/src/ir/operation/BCQFullyConnected.cc
```diff
@@ -31,7 +31,7 @@ void BCQFullyConnected::accept(OperationVisitor &v) const { v.visit(*this); }
 
 BCQFullyConnected::BCQFullyConnected(const OperandIndexSequence &inputs,
                                      const OperandIndexSequence &outputs, const Param &param)
-    : Operation{OperandConstraint::createExact(4u), inputs, outputs}, _param{param}
+    : Operation{OperandConstraint::createExact(5u), inputs, outputs}, _param{param}
```
llFreetimell, line 34: I don't know well about runtime so I have a question..!
Does `createExact(5u)` allow to get only four inputs?

runtime/onert/core/include/ir/operation/BCQFullyConnected.h
dayo09: I agree. It's fixed.

runtime/onert/core/src/ir/operation/BCQFullyConnected.cc
dayo09: @llFreetimell Yes, I will checkout

PR #2319, opened 2020-06-16
[luci] Revise shape inf of While

seanshpark:
This will revise shape inference of While op where current shape of matching Cond graph is not same as shape of node from mode file

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.cpp
@@ -2198,7 +2198,17 @@ public:
     auto cond_graph_inputs = cond_graph->inputs();
     auto cond_graph_input = cond_graph_inputs->at(cond_in->index());
 
-    return loco::NodeShape{*cond_graph_input->shape()};
+    auto cond_graph_input_shape = *cond_graph_input->shape();
+    auto this_shape = own_shape(node);
+
+    if (!(this_shape == cond_graph_input_shape))
+    {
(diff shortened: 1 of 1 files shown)


seanshpark:
For #2260 from draft #2261

seanshpark:

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -2198,7 +2198,17 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
     auto cond_graph_inputs = cond_graph->inputs();
     auto cond_graph_input = cond_graph_inputs->at(cond_in->index());
 
-    return loco::NodeShape{*cond_graph_input->shape()};
+    auto cond_graph_input_shape = *cond_graph_input->shape();
+    auto this_shape = own_shape(node);
+
+    if (!(this_shape == cond_graph_input_shape))
```
jinevening, line 2204: Is there a case when this happens?

I thought this will not happen.

seanshpark: Yes, out in-house model has different shape. where shape for Cond has `[]` but main `WHILE` has `[1,0]`

seanshpark: Look https://github.com/Samsung/ONE/issues/2260#issuecomment-644474454

PR #2320, opened 2020-06-16
[luci] Revise create_circleconst

seanshpark:
This will revise import create_circleconst() method to skip creation of CircleConst if the tensor is output of some other operator


seanshpark:
For #2260 from draft #2261

seanshpark:

PR #2322, opened 2020-06-16
[luci] Revise export allocateCircleTensor

seanshpark:
This will revise allocateCircleTensor to separate NoOp and MultipleOutputOps and use allocateCircleTensorInfo as common method


seanshpark:
For #2260 from draft #2261

seanshpark:

PR #2323, opened 2020-06-16
[tflchef] Support SegmentSum operator

d-krylov:
Support SegmentSum operator in tflchef


compiler/tflchef/core/src/OpChef.def
```diff
@@ -71,6 +71,8 @@ OP_CHEF(ReverseSequence, ReverseSequenceChefFactory)
 OP_CHEF(Round, RoundChefFactory)
 OP_CHEF(Rsqrt, RsqrtChefFactory)
 OP_CHEF(ScatterNd, ScatterNdChefFactory)
+OP_CHEF(SegmentSum,SegmentSumChefFactory)
+OP_CHEF(Shape, ShapeChefFactory)
```
mhs4670go, line 75: This might be added mistakenly.

compiler/tflchef/proto/tflchef.proto
```diff
@@ -492,7 +496,7 @@ message Operation {
   optional Pool2DOptions l2pool2d_options = 203;
   optional ReducerOptions all_options = 204;
   optional ReducerOptions reduce_min_options = 205;
-
+  optional SegmentSumOptions segment_sum_options = 206;
```
binarman, line 499: removed empty line

compiler/tflchef/core/src/OpChef.def
binarman: Indeed!
```suggestion
```

compiler/tflchef/proto/tflchef.proto
binarman:
```suggestion
  optional SegmentSumOptions segment_sum_options = 206;

```

PR #2325, opened 2020-06-16
[nnpkgrun] Introduce `--output_sizes`

wateret:
As NNFW API requires user to prepare enough buffer sizes for outputs if
a model is dynamic, this commit introduces an command line option for
`nnpackage_run`.

Usage:

```
./nnpackage_run package_path --output_sizes="[0, 40, 2, 80]"
```

Will allocate 40 bytes for output 0 and 80 bytes for output 1.

### DIFF
--- a/tests/tools/nnpackage_run/CMakeLists.txt
+++ b/tests/tools/nnpackage_run/CMakeLists.txt
@@ -32,7 +32,7 @@ target_include_directories(nnpackage_run PRIVATE src)
 target_include_directories(nnpackage_run PRIVATE ${Boost_INCLUDE_DIRS})
 
 target_link_libraries(nnpackage_run onert_core onert tflite_loader)
-target_link_libraries(nnpackage_run tensorflow-lite ${LIB_PTHREAD} dl nnfw_lib_tflite)
+target_link_libraries(nnpackage_run tensorflow-lite ${LIB_PTHREAD} dl nnfw_lib_tflite jsoncpp)
 target_link_libraries(nnpackage_run nnfw-dev)
 if(BUILD_BOOST)
   # We have to use Boost::program_options, instead of boost_program_optinos
--- a/tests/tools/nnpackage_run/src/args.cc
+++ b/tests/tools/nnpackage_run/src/args.cc
@@ -17,6 +17,44 @@
 #include "args.h"
 
 #include <iostream>
+#include <json/json.h>
+
+namespace
+{
+
+// This function parses a json object and returns as a vector of integers
+// For example,
+// [0, [1, 2, 3, 4], 3, 40, 4, []] in JSON
+// is converted to:
+// {
+//  0 -> [1, 2, 3, 4]
+//  3 -> 40
+//  4 -> []
+// } in std::unordered_map. Note that the value type is still Json::Value.
+std::unordered_map<uint32_t, Json::Value> argArrayToMap(const Json::Value &jsonval)
+{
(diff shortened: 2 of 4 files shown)


hyunsik-yoon:
Can I ask a question? If we have n outputs and only 1 among k th output is dynamic, we have to specify _all_ sizes from 0th to kth outputs (where k < n) ?
(https://github.com/Samsung/ONE/pull/2325/files#diff-1acc035a02fb45aa47c0f553a16a3d9cR211)
I guess this is difficult to use. Maybe we need to add such thing in, e.g., readme, or arg description.

I think it would be better to align the format of this commit and format of input shapes param(which I wrote yesterday, https://github.com/Samsung/ONE/pull/2290) too. There, input shapes can be given, like, `[0, [1, 2], 2, []]` where 0th input will be `[1, 2]` and 2nd input will be `[ ]` (1th input can be omitted). I thought that skipping some input whose shape is obvious would be better for user of nnpackage_run. I learned this from last year when _nncc compiler_ has `*.info` file where all shapes of inputs should be provided... _nnfw_ people had hard time to write `*.info`.

If you have some reason to design the format of outputs sizse, please go ahead.
In case of input shape, I think that I'd better stick to the format shown at https://github.com/Samsung/ONE/pull/2325/files#diff-1acc035a02fb45aa47c0f553a16a3d9cR211

wateret:
@hyunsik-yoon I know this commit is not perfect and here's my thoughts on that.

- Just for clarification unlike input shapes output buffers do not need shape info, it just needs the buffer's size
- For non-dyanmic inputs please refer [here](https://github.com/Samsung/ONE/pull/2325/files#diff-49d11279d07f1b30af524275983fe6daR46), or we could use `false` or `null` as well
- If the array size is greater than number of inputs, the trailing values are ignored. If number of inputs are greater than the array size, the unspecified ones are considered as static inputs. - I did this since it is easy to implement and it makes sense a bit. We can change it strictly match those two
- I will add an instruction once we've decided the format

And I agree with we should have the same style. Here's my thoughts on my style(array of shapes/bufsize and index is the key) and your style(explicit key and value).

- As inputs and outputs are arrays without holes so having it as an array also makes sense
- Your style is more useful when there is a lot of inputs/outputs and want to specify just few of them and my style is more useful when there are only a couple of inputs/outputs in the model. And I did this as I think most of the tests have few inputs and outputs
- `[0, [1, 2], 2, []]` does not look so readable to me. I was thinking that `{"0": [1, 2], "2": []}` would make more sense
    - NOTE: strictly speaking, JSON spec says a key must be a string

With these reasons I would suggest **accepting both the dictionary style(`{}`) and my array style(`[]`)**. What about it?

wateret:
After the discussion with @hyunsik-yoon , I am going to follow @hyunsik-yoon 's way!

hyunsik-yoon:
After the discussion with @wateret, we are going to follow his suggestion to use dictionary format instead of list format!
BTW this feature is getting serious.
One of the test I ran:
```
odroid@odroid:~/work/1/ONE$ BACKENDS=cpu  ${ONE_PATH}/Product/out/bin/nnpackage_run  ${MODEL_PATH}/subg_0_0_37 --shape_compile '[0, [1, 10]]' --load ${MODEL_PATH}/large_test_input_0.h5 
Error during nnfw_session::run : ExecutorBase: output buffer size is less than output tensor size
```

wateret:
@hyunsik-yoon I found that non-string keys cannot be parsed with JSONCPP. Let's get back to your solution.

PR #2327, opened 2020-06-16
[res] Add SegmentSum operator recipe

d-krylov:
Add SegmentSum operator recipe

### DIFF
--- a/res/TensorFlowLiteRecipes/SegmentSum_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/SegmentSum_000/test.recipe
@@ -1,3 +1,6 @@
+# TODO We need to check if toco generates models like described in this recipe.
+# Toco from TF 2.2.0 was not able to convert SegmentSum operator.
+
 operand {
   name: "ifm"
   type: FLOAT32
--- a/res/TensorFlowLiteRecipes/SegmentSum_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/SegmentSum_000/test.recipe
@@ -1,0 +1,26 @@
+operand {
+  name: "ifm"
+  type: FLOAT32
+  shape { dim: 1 dim: 4 dim: 4 dim: 3 }
+}
+operand {
+name: "segment_ids"
+  type: INT32
+  shape { dim: 4 }
+  filler {
+    tag: "explicit"
+    arg: "1" arg: "4" arg: "4" arg: "3" 
+  }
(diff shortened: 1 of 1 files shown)


res/TensorFlowLiteRecipes/SegmentSum_000/test.recipe
```diff
@@ -0,0 +1,27 @@
...
+  type: FLOAT32
+  shape { dim: 1 dim: 4 dim: 4 dim: 3 }
+}
+operand {
+name: "segment_ids"
+  type: INT32
+  shape { dim: 4 }
+  filler {
+    tag: "explicit"
+    arg: "1" arg: "4" arg: "4" arg: "3"
```
binarman, line 12: This operator need to be investigated.
I tried to create such operator in TF and it failed. On the other hand TFLite 2.2 interpreter did interpret model created by this recipe (though I am not sure it did something sane).

I have two consernt about this code:
1) number of elements in segment_ids should be equal to first dimension of input
2) According TF documentation "The standard segment_* functions assert that the segment indices are sorted. "

binarman: I tried to fill input tensor with ones, and got garbage in the output, I think tflite interpret worked just because it assumes that all checks are done in TF.

binarman:
I pushed changes that work in tensorflow, but I was not able to get expected results in TFLite for some reason.

Maybe this operator is too new and actually not supported in tflite (letest toco can not convert model with this operator).

I propose to proceed as-is for now and fix it when it is needed

seanshpark:
> I propose to proceed as-is for now and fix it when it is needed

Can you add a note like this ?
```
# NOTE we need more investigation when running this model
```

binarman:
@seanshpark 
I checked again, I made a mistake in "interpreter" script =_=
Finally I was able to get expected output, so I consider this "likely" correct =)

P.S. Still was not able to convert TF model in toco.

res/TensorFlowLiteRecipes/SegmentSum_000/test.recipe
```diff
@@ -0,0 +1,27 @@
+operand {
```
binarman, line 1:
```suggestion
# TODO We need to check if toco generates models like described in this recipe.
# Toco from TF 2.2.0 was not able to convert SegmentSum operator.

operand {
```

PR #2328, opened 2020-06-16
[docs] Remove "How to Use API"

wateret:
It looks like "How to Use API" means NNFW API which has its document
already.

### DIFF
--- a/docs/howto/how-to-use-api.md
+++ b/docs/howto/how-to-use-api.md
@@ -1,1 +1,0 @@
-# How to Use API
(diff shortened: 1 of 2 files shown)


lemmaa:
@wateret , I was hoping that this document would give an overall description of the API provided by the ONE runtime. About both NNFW API and Android NNAPI binding. In addition, I thought it would be good to introduce the history of the change in API support of the ONE runtime. What are your thoughts?

wateret:
@lemmaa That would be nice. I'm not sure about the history but I agree on introducing both NN API and NNFW API.

PR #2331, opened 2020-06-16
[tflite2circle] Support Where operator

d-krylov:
Support Where operator in tflite2circle


compiler/tflite2circle/src/BuildBuiltinOptions/WhereOptions.cpp
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+#include "WhereOptions.h"
+
+namespace tflite2circle
+{
+
+flatbuffers::Offset<circle::WhereOptions>
+build_circle_WhereOptions(flatbuffers::FlatBufferBuilder &fb, const tflite::Operator *)
```
lucenticus, line 23: It seems that you forgot `op`
```suggestion
build_circle_WhereOptions(flatbuffers::FlatBufferBuilder &fb, const tflite::Operator *op)
```

mhs4670go: @lucenticus `op` is not used in this function. This prevents unused variable warning.

lucenticus: @mhs4670go  Ok, thanks for clarifying!

PR #2334, opened 2020-06-16
[luci] Fix for unused var release build error

seanshpark:
This will fix unused variable error in release build in export

### DIFF
--- a/compiler/luci/export/src/CircleTensorExporter.cpp
+++ b/compiler/luci/export/src/CircleTensorExporter.cpp
@@ -132,6 +132,7 @@ private:
   {
     auto outs = loco::succs(node);
     assert(outs.size() == count);
+    (void)count; // for unused variable error in release build
     for (auto out : outs)
     {
       auto circle_out = loco::must_cast<luci::CircleNode *>(out);


seanshpark:
For #2260 from draft #2261

seanshpark:
@nnfw-bot test nncc-debug

seanshpark:

compiler/luci/export/src/CircleTensorExporter.cpp
```diff
@@ -132,6 +132,7 @@ class MultiOutputDetector final : public luci::CircleNodeMutableVisitor<bool>
   {
     auto outs = loco::succs(node);
     assert(outs.size() == count);
+    (void)count; // for unused variable error in release build
```
llFreetimell, line 135: I saw the full draft codes, but I couldn't find why `count` is needed :(
Will this variable be used for later?
If not, how about following code?
```cpp
void store_outputs(luci::CircleNode *node, uint32_t) // Remove count
```

llFreetimell, line 135: (Optional) How about change above assert as runtime_error?

seanshpark: Behind this is that import will check the condition and throw if not. This assert is just to be sure the condition is OK.

PR #1980, opened 2020-06-07
[WIP][luci-interpreter] Multiple Graph support

s-barannikov:
ONE-DCO-1.0-Signed-off-by: Sergei Barannikov <s.barannikov@samsung.com>

PR #2335, opened 2020-06-16
[luci-interpreter] Add tensor accessors to kernels

s-barannikov:
Draft PR #1980.

Add `getInputTensors` and `getOutputTensors` methods to kernels.

### DIFF
--- a/compiler/luci-interpreter/src/core/Kernel.h
+++ b/compiler/luci-interpreter/src/core/Kernel.h
@@ -19,6 +19,8 @@
 
 #include "luci_interpreter/core/Tensor.h"
 
+#include <vector>
+
 namespace luci_interpreter
 {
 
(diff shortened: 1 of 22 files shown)


jinevening:
Why does a kernel need input/output tensor accessor ? I've seen the draft PR, but couldn't find a place where a kernel calls the accessor.

s-barannikov:
@jinevening 
`getOutputTensors` is used [here](https://github.com/Samsung/ONE/blob/435326e659d6b87b688429a3b6dc294b1a5c5306/compiler/luci-interpreter/src/core/RuntimeGraph.cpp#L73).
I added `getInputTensors` for symmetry, there are no uses of it yet.

s-barannikov:
@binarman 
There is no luci graph when `getOutputTensors` is called. I also think these two methods are natural for kernels to have.

s-barannikov:
@jinevening As soon the need arises :)

PR #2336, opened 2020-06-16
[luci] Remove duplicates in MultiOutputDetector

seanshpark:
This will remove duplicate codes in export MultiOutputDetector class


seanshpark:
For #2260 from draft #2261

seanshpark:

Issue #2275, opened 2020-06-16
[onert] int64 support needed for some op


hyunsik-yoon:
We need int64 support for the following to run out test model, large_test.tflite:
([large_test.tflite info](https://github.sec.samsung.net/STAR/nnfw/issues/11334#issuecomment-719756))

Can someone check and help?
- [x] matrixBandPart
- [x] ArgMax
- [x] Cast

```
Operator 20: CUSTOM(MatrixBandPart)
        Fused Activation: NONE
        Input Tensors[136, 3, 4]
                Tensor  136 : buffer  137 |  Empty | FLOAT32 | Memory 4.0B   | Shape Scalar (b'transformer_c3po/attention_bias_lower_triangle/attention_bias_local/ones')
                Tensor    3 : buffer    4 | Filled | INT64   | Memory 8.0B   | Shape Scalar (b'transformer_c3po/attention_bias_lower_triangle/attention_bias_local/Cast')
                Tensor    4 : buffer    5 | Filled | INT64   | Memory 8.0B   | Shape Scalar (b'transformer_c3po/attention_bias_lower_triangle/attention_bias_local/Cast_1')
        Output Tensors[137]
                Tensor  137 : buffer  138 |  Empty | FLOAT32 | Memory 4.0B   | Shape Scalar (b'transformer_c3po/attention_bias_lower_triangle/attention_bias_local/MatrixBandPart')

Operator 907: ARG_MAX
        Fused Activation: NONE
        Input Tensors[1063, 48]
                Tensor 1063 : buffer 1957 |  Empty | FLOAT32 | Memory 4.0B   | Shape Scalar (b'transformer_c3po/while/Squeeze')
                Tensor   48 : buffer  942 | Filled | INT32   | Memory 4.0B   | Shape Scalar (b'transformer_c3po/get_timing_signal_1d/range/delta1')
        Output Tensors[1064]
                Tensor 1064 : buffer 1958 |  Empty | INT64   | Memory 8.0B   | Shape Scalar (b'transformer_c3po/while/ArgMax')

Operator 908: CAST
        Fused Activation: NONE
        Input Tensors[1064]
                Tensor 1064 : buffer 1958 |  Empty | INT64   | Memory 8.0B   | Shape Scalar (b'transformer_c3po/while/ArgMax')
        Output Tensors[1065]
[...]

PR #2339, opened 2020-06-17
[onert] support int64 for Cast operation

intom:
parent issue : [#2275](https://github.com/Samsung/ONE/issues/2275)

Add type to support int64 cast operation

ONE-DCO-1.0-Signed-off-by: sungho22.lee <sungho22.lee@samsung.com>


hyunsik-yoon:
Can we have some test with pretty big number > max(int32) just in case? :-)

ragmani:
> Can we have some test with pretty big number > max(int32) just in case? :-)

We cannot add it as generated unittest because nnapi does not support the type. If we want to have some test with the type, we should create tflite files for the test.

hyunsik-yoon:
@intom , please go see https://github.com/Samsung/ONE/issues/2344

hseok-oh:
@intom Please remove NO_MERGE label if it is ready to review and merge.

PR #2342, opened 2020-06-17
[luci] More tests for Unpack and While

seanshpark:
This will enable more tests for Unpack and While
- has unused multiple outputs
- has unknown shape tensor that is not constant


seanshpark:
For #2260 from draft #2261

seanshpark:

Issue #696, opened 2020-05-12
[Compiler] Post-training Quantization Support


jinevening:
**Update: 2020/06/05**. layer-wise quantization -> channel-wise quantization.
List of channel-wise quantized Ops (conv, depthwise_conv, fc, transpose_conv)

**Update: 2020/05/22**. Step 0 was added to consider quantization loss when recording min/max.

We plan to support post-training quantization. Quantization algorithm is channel-wise affine quantization using representative dataset (similar to: TensorFlow [calibrator](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/python/optimize/calibrator.py#L57)).

### Quantization using representative dataset 

**Brief overview**: Record min-max values of tensors while performing inference on the given representative dataset. Next, calculate zero point and scale using the recorded min-max values. Finally, quantize the model using the derived zero point and scale.
(Quantization formula: real value = scale * (quantized value - zero point))

<center><img src="https://user-images.githubusercontent.com/5449554/82645539-faa8c200-9c4d-11ea-94b9-50f49531966f.png" width="600"></center>

**Step 0. Quantize and dequantize the weights of the model (by luci)**
**Input**: circle model (fp32)
**Output**: circle model (fp32) whose weights are quantized and dequantized.
**Details**: Users run ```circle2circle``` with the option ```--quantize_dequantize_weights```. ```circle2circle``` imports the fp32 circle model, runs an optimization pass named ```QuantizeDequantizeWeightsPass```, and exports the model as the fp32 circle model. ```QuantizeDequantizeWeightsPass``` quantizes weights and then dequantizes the weights back. This is to further consider quantization loss during profiling.

**Step 1. Record min-max value of each tensor while running the representative dataset (by luci-interpreter)**
**Input**: circle model (fp32), representative dataset (~~format is TBD. hdf5?~~ format is hdf5)
**Output**: circle model (fp32) where min-max values of tensors are saved in [QuantizationParameters](https://github.com/Samsung/ONE/blob/master/nnpackage/schema/circle_schema.fbs#L67).
**Details**: Users run executable named ```record-minmax```, which invokes luci-interpreter to perform inference on the given dataset. Whenever performing inference on each data, ```record-minmax``` records the moving average of min and the moving average of max for each tensor. After the whole dataset is fed to the interpreter, the recorded average min-max values are saved in the ```QuantizationParameters``` of each tensor in the circle model. 

**Step 2. Quantize the model using the embedded information (by luci)**
[...]

PR #2346, opened 2020-06-17
[luci] quantize convolution's weight to layer-wise uint8

meejeong:
- apply to quantize convolution's only weights and bias to layer-wise 
- datatype : uint8
- asymmetric quantization

Related to: #696, https://github.sec.samsung.net/STAR/nnapps/issues/379

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -111,8 +111,8 @@ std::pair<loco::Node *, loco::Node *> get_input_weight_of_bias(CircleNode *node)
   return std::make_pair(nullptr, nullptr);
 }
 
-void quant_bias(CircleConst *node, float input_scale, float weight_scale, float *scaling_factor,
-                int64_t *zp)
+void asym_quant_bias(CircleConst *node, float input_scale, float weight_scale,
+                     float *scaling_factor, int64_t *zp)
 {
   float scale = input_scale * weight_scale;
   const float scaling_factor_inv = (scale == 0) ? 0 : 1.0 / scale;
@@ -127,10 +127,12 @@ void quant_bias(CircleConst *node, float input_scale, float weight_scale, float 
 
   node->dtype(loco::DataType::S32);      // change the type of tensor
   node->size<loco::DataType::S32>(size); // resize tensor
-  const int32_t kScale = std::numeric_limits<int32_t>::max();
+  const int32_t kMinScale = std::numeric_limits<int32_t>::lowest();
+  const int32_t kMaxScale = std::numeric_limits<int32_t>::max();
   for (int i = 0; i < static_cast<int32_t>(size); ++i)
   {
-    node->at<loco::DataType::S32>(i) = std::min(kScale, std::max(-kScale, quantized_values[i]));
+    node->at<loco::DataType::S32>(i) =
+        std::min(kMaxScale, std::max(kMinScale, quantized_values[i]));
   }
   *scaling_factor = scale;
   *zp = 0;
@@ -147,32 +149,19 @@ bool is_quantized(const CircleNode *node)
          node->dtype() == loco::DataType::S32;  // bias
 }
 
-void asym_wquant_with_minmax(CircleConst *node, float min, float max, float *scaling_factor,
-                             int64_t *zp)
+void asym_wquant(CircleConst *node, float min, float scaling_factor)
 {
   const int32_t kMinScale = 0;
   const int32_t kMaxScale = 255;
 
   uint32_t size = node->size<loco::DataType::FLOAT32>();
-  if (min == max)
-  {
-    node->dtype(loco::DataType::U8);      // change the type of tensor
-    node->size<loco::DataType::U8>(size); // resize tensor
-    for (int i = 0; i < static_cast<int32_t>(size); ++i)
-      node->at<loco::DataType::U8>(i) = 0;
 
-    *scaling_factor = 1;
-    *zp = 0;
-    return;
-  }
-
-  compute_asym_scale_zp(min, max, scaling_factor, zp);
-  const float scaling_factor_inv = 1.0 / *scaling_factor;
+  const float scaling_factor_inv = 1.0 / scaling_factor;
   std::vector<int32_t> quantized_values(size);
   for (int i = 0; i < static_cast<int32_t>(size); ++i)
   {
-    quantized_values[i] = static_cast<int32_t>(
-        std::round(*zp + node->at<loco::DataType::FLOAT32>(i) * scaling_factor_inv));
+    auto data = node->at<loco::DataType::FLOAT32>(i);
+    quantized_values[i] = static_cast<int32_t>(std::round((data - min) * scaling_factor_inv));
   }
 
   node->dtype(loco::DataType::U8);      // change the type of tensor
@@ -270,12 +259,14 @@ struct QuantizeActivation final : public luci::CircleNodeMutableVisitor<bool>
 
 struct QuantizeBias final : public luci::CircleNodeMutableVisitor<bool>
 {
-  QuantizeBias(loco::DataType input, loco::DataType output) : input_type(input), output_type(output)
+  QuantizeBias(loco::DataType input, loco::DataType output, QuantizationGranularity gr)
+      : input_type(input), output_type(output), granularity(gr)
   {
   }
 
   loco::DataType input_type;
   loco::DataType output_type;
+  QuantizationGranularity granularity;
 
   // Quantize bias node
   bool visit(luci::CircleNode *node)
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -111,29 +111,31 @@ std::pair<loco::Node *, loco::Node *> get_input_weight_of_bias(CircleNode *node)
...
+void asym_quant_bias(CircleConst *node, float input_scale, float weight_scale,
+                     float &scaling_factor, int64_t &zp)
 {
   float scale = input_scale * weight_scale;
   const float scaling_factor_inv = (scale == 0) ? 0 : 1.0 / scale;
 
   uint32_t size = node->size<loco::DataType::FLOAT32>();
   std::vector<int32_t> quantized_values(size);
-  for (int i = 0; i < static_cast<int32_t>(size); ++i)
+  for (uint32_t i = 0; i < size; ++i)
```
seanshpark, line 122: This change doesn't seem to be related with this PR

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -111,29 +111,31 @@ std::pair<loco::Node *, loco::Node *> get_input_weight_of_bias(CircleNode *node)
   return std::make_pair(nullptr, nullptr);
 }
 
-void quant_bias(CircleConst *node, float input_scale, float weight_scale, float *scaling_factor,
-                int64_t *zp)
+void asym_quant_bias(CircleConst *node, float input_scale, float weight_scale,
+                     float &scaling_factor, int64_t &zp)
```
seanshpark, line 115: changing ptr to ref is also doesn't seem to be related with this PR

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -111,29 +111,31 @@ std::pair<loco::Node *, loco::Node *> get_input_weight_of_bias(CircleNode *node)
...
         static_cast<int32_t>(std::round(node->at<loco::DataType::FLOAT32>(i) * scaling_factor_inv));
   }
 
   node->dtype(loco::DataType::S32);      // change the type of tensor
   node->size<loco::DataType::S32>(size); // resize tensor
-  const int32_t kScale = std::numeric_limits<int32_t>::max();
-  for (int i = 0; i < static_cast<int32_t>(size); ++i)
+  const int32_t kMinScale = std::numeric_limits<int32_t>::lowest();
+  const int32_t kMaxScale = std::numeric_limits<int32_t>::max();
+  for (uint32_t i = 0; i < size; ++i)
```
seanshpark, line 132: ditto

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -147,37 +149,24 @@ bool is_quantized(const CircleNode *node)
...
-    *zp = 0;
-    return;
-  }
 
-  compute_asym_scale_zp(min, max, scaling_factor, zp);
-  const float scaling_factor_inv = 1.0 / *scaling_factor;
+  const float scaling_factor_inv = 1.0 / scaling_factor;
   std::vector<int32_t> quantized_values(size);
-  for (int i = 0; i < static_cast<int32_t>(size); ++i)
+  for (uint32_t i = 0; i < size; ++i)
```
seanshpark, line 161: ditto

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -147,37 +149,24 @@ bool is_quantized(const CircleNode *node)
...
-    quantized_values[i] = static_cast<int32_t>(
-        std::round(*zp + node->at<loco::DataType::FLOAT32>(i) * scaling_factor_inv));
+    auto data = node->at<loco::DataType::FLOAT32>(i);
+    quantized_values[i] = static_cast<int32_t>(std::round((data - min) * scaling_factor_inv));
   }
 
   node->dtype(loco::DataType::U8);      // change the type of tensor
   node->size<loco::DataType::U8>(size); // resize tensor
-  for (int i = 0; i < static_cast<int32_t>(size); ++i)
+  for (uint32_t i = 0; i < size; ++i)
```
seanshpark, line 169: ditto

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -147,37 +149,24 @@ bool is_quantized(const CircleNode *node)
          node->dtype() == loco::DataType::S32;  // bias
 }
 
-void asym_wquant_with_minmax(CircleConst *node, float min, float max, float *scaling_factor,
-                             int64_t *zp)
+void asym_wquant_with_minmax(CircleConst *node, float min, float scaling_factor)
```
jinevening, line 152: It would be better to change the function name.

Maybe something like `asym_wquant_with_min_scale` or simply `asym_wquant`.

I'd preferred the second one.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -345,24 +343,22 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
         {
-          auto data = circle_const->at<loco::DataType::FLOAT32>(i);
-          min = data < min ? data : min;
-          max = data > max ? data : max;
+          // will be implemented.
+        }
+        // Find min/max per layer-wise
+        else
+        {
+          // Quantize using recorded min/max
```
jinevening, line 354: recorded min/max -> recorded quantparam

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
meejeong: I also agree the second one :-) I will apply it.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -292,22 +283,28 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<bool>
...
+    {
+      assert(input->quantparam()->scale.size() == 1); // Only support per-layer quant
+      auto input_scale = input->quantparam()->scale[0];
+
+      assert(weight->quantparam()->scale.size() == 1); // Only support per-layer quant
+      auto weight_scale = weight->quantparam()->scale[0];
+
+      auto circle_const = loco::must_cast<luci::CircleConst *>(node);
+      float scaling_factor;
+      int64_t zp;
```
mhs4670go, line 300: Please initialize it to prevent warning in static analysis

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -345,24 +343,22 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
-          auto data = circle_const->at<loco::DataType::FLOAT32>(i);
-          min = data < min ? data : min;
-          max = data > max ? data : max;
+          // will be implemented.
+        }
+        // Find min/max per layer-wise
+        else
+        {
+          // Quantize using recorded quantparam
+          auto quantparam = circle_node->quantparam();
```
mhs4670go, line 355: maybe it would be better to assert `quantparam != nullptr`.

meejeong: I applied it :-)

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -270,12 +259,14 @@ struct QuantizeActivation final : public luci::CircleNodeMutableVisitor<bool>
 
 struct QuantizeBias final : public luci::CircleNodeMutableVisitor<bool>
 {
-  QuantizeBias(loco::DataType input, loco::DataType output) : input_type(input), output_type(output)
+  QuantizeBias(loco::DataType input, loco::DataType output, QuantizationGranularity granularity)
+      : input_type(input), output_type(output), granularity(granularity)
```
seanshpark, line 263: `granularity(granularity)` is same variable. does this work correctly?

meejeong: when I execute this code, it works properly. However, I can change it if you think it is confused.

seanshpark: Ah.. thanks, let's change it like `granularity(gr)`  :)

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -318,13 +315,14 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<bool>
  */
 struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
 {
-  QuantizeWeights(loco::DataType input, loco::DataType output)
-      : input_type(input), output_type(output)
+  QuantizeWeights(loco::DataType input, loco::DataType output, QuantizationGranularity granularity)
+      : input_type(input), output_type(output), granularity(granularity)
```
seanshpark, line 319: same here :)

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
mhs4670go: AFAIK, It works well

mhs4670go: AFAIK, it works well:)

s-barannikov: This indeed works well, but we usually use underscore prefix for private variables, e.g. "_granularity".
`QuantizeBias` should be a class rather than a struct, and the variables should be private.
It's just a note, please proceed without my approval if others find changes OK.

PR #2352, opened 2020-06-17
[luci] Add `--no-cache-dir` option of pip in tests using pip

ragmani:
Fix https://github.com/Samsung/ONE/issues/457#issuecomment-645172551

This commit adds `--no-cache-dir` option of pip in tests using pip.

### DIFF
--- a/compiler/luci-value-test/CMakeLists.txt
+++ b/compiler/luci-value-test/CMakeLists.txt
@@ -29,8 +29,8 @@ set(REQUIREMENTS_BIN_PATH "${CMAKE_CURRENT_BINARY_DIR}/${REQUIREMENTS_FILE}")
 add_custom_command(
   OUTPUT ${REQUIREMENTS_BIN_PATH}
   COMMAND ${CMAKE_COMMAND} -E copy ${REQUIREMENTS_SRC_PATH} ${REQUIREMENTS_BIN_PATH}
-  COMMAND ${VIRTUALENV}/bin/python -m pip install --upgrade pip setuptools --timeout 100
-  COMMAND ${VIRTUALENV}/bin/python -m pip install -r requirements.txt --upgrade --timeout 100
+  COMMAND ${VIRTUALENV}/bin/python -m pip install --upgrade pip setuptools --no-cache-dir --timeout 100
+  COMMAND ${VIRTUALENV}/bin/python -m pip install -r requirements.txt --no-cache-dir --upgrade --timeout 100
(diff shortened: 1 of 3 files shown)


mhs4670go:
IMHO, I guess this is not a cache problem. Because TensorFlow is installed in a new virtual environment every build. It might be network problem, which makes tensorflow file invalid. just my guess though:)

ragmani:
@jinevening 
I will update it in this PR.

ragmani:
> IMHO, I guess this is not a cache problem. Because TensorFlow is installed in a new virtual environment every build. It might be network problem, which makes tensorflow file invalid. Just my guess though:)

@mhs4670go 
Can I get the information which network problem makes tensorflow file invalid?

seanshpark:
>  TensorFlow is installed in a new virtual environment every build

Can we use overlay (persistant storage to docker) ?

ragmani:
> Can we use overlay (persistant storage to docker) ?

It is possible with mounting the volume without clean build. But @mhs4670go would mean `virtualenv`.

mhs4670go:
@ragmani Well, I don't know how network between package server works at all. I just found out two solutions about this error in Google.
1. add `--no-cache-dir`
2. rerun command
If this test always failed, it would be cache problem. But, this error occurs intermittently. That is why I thought it was. 

@seanshpark It might be possible. In overlay folder `-DNNCC_OVERLAY_DIR=/overlay`, we might be able to create virtual environment first.

seanshpark:
> It is possible with mounting the volume without clean build. But @mhs4670go would mean virtualenv.

Yes, if we can put virtual env folder as an overlay, we don't have to build (download) it every time.
But we may need to set the overlay for every build machine...

seanshpark:
I think this change is OK as @mhs4670go confirms :)

ragmani:
> If this test always failed, it would be cache problem. But, this error occurs intermittently. That is why I thought it was.

@mhs4670go 
Well, I think it does not guarantee it would be not cache problem even if the error occurs intermittently. If `pip` in `virtualenv` uses the default cache, the error occurs at times when installed packages's hash code is updated.

ragmani:
> Yes, if we can put virtual env folder as an overlay, we don't have to build (download) it every time.
But we may need to set the overlay for every build machine...

@seanshpark 
I see what you mean now. But I have no idea where is installed packages of pip and how to set the overaly to pip in a virtualenv.

seanshpark:
#2359 is created for using overlay :)

Issue #1907, opened 2020-06-04
[Compiler FE] Supporting new BCQ Format


llFreetimell:
Draft PR : #2023

Format of BCQ operations is upgraded as follow.
- More information is needed so the number of input will be increased
- Format and interpretation of quantized weight/input is changed
- Inserting `Transpose` or changing `axis` of `Gather` is needed often

~~Original codes for handling original BCQ format will be kept for a while.
However, those codes will be deprecated eventually in the near future.
In conclusion, `FuseNewBCQPass` will be introduced and it will substitute old `FuseBCQPass` in the future.~~
_(20200617)UPDATE : Old BCQ will be substituted by new BCQ **immediately**_

### Todo
- [x] ~~Update schema of circle~~ (Not needed)
- [x] Update `CircleBCQFullyConnected` #2340
- [x] Update `CircleBCQGather` #2340
- [x] ~~Update `circledump`~~ (Not needed)
- [x] Update tests in `luci/lang` #2446
- [x] Update `logex` in `luci` #2349
- [x] Update `import` in `luci` #2353
- [x] Update `export` in `luci` #2447
- [x] Update `ShapeInference` in `luci/Service` #2452
- [x] Update `FuseBCQPass` in `luci` #2487 #2646
- [x] Add testcase (#1517)

llFreetimell:
AFAIK, We don't need to keep old BCQ version, so I will apply this change soon.

llFreetimell:
Huge problem occurred.... if values of constant nodes are same, they are removed.........

llFreetimell:
### 20200619 Update

#### Situation
When `const1`, `const2`, `const3` have all same values, they are removed to one constant node in TFv2 converter as follow.
```
[before converting in TFv2]
const1 --- output1
const2 --- output2
const3 --- output3

[after converting in TFv2]
const1 --- output1
const1 --- output2 (const2 is removed)
const1 --- output3 (const3 is removed)
```

#### Workaround
When we use TFv1 converter, above constant nodes are not removed as follow.
```
[before converting in TFv1]
const1 --- output1
const2 --- output2
const3 --- output3

[after converting in TFv1]
const1 --- Reshape --- output1
const2 --- Reshape --- output2
const3 --- Reshape --- output3
```

#### Solution
I will find another solution to handle it because there will be some cases that we should use TFv2 converter.

llFreetimell:
### 20200623 Update

[...]

PR #2353, opened 2020-06-17
[luci] Update import of BCQ operations

llFreetimell:
Parent Issue : #1907

New BCQ operations have one more inputs.
This commit will update import of BCQ operations to handle additional input.

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleBCQFullyConnected.cpp
+++ b/compiler/luci/import/src/Nodes/CircleBCQFullyConnected.cpp
@@ -25,7 +25,7 @@ namespace luci
 
 bool CircleBCQFullyConnectedGraphBuilder::validate(const ValidateArgs &args) const
 {
-  if (args.op.inputs.size() != 4)
+  if (args.op.inputs.size() != 5)
     return false;
 
   return true;
@@ -41,6 +41,16 @@ CircleNode *CircleBCQFullyConnectedGraphBuilder::build_node(const circle::Operat
   node->weights_scales(inputs[1]);
   node->weights_binary(inputs[2]);
   node->bias(inputs[3]);
+  node->weights_clusters(inputs[4]);
+
+  // TODO Find and move to appropriate place for setting optional input
+  if (auto bias = dynamic_cast<luci::CircleOutputExclude *>(node->bias()))
+  {
+    // bias is not used for type inference, but node itself should have a type
+    bias->dtype(loco::DataType::FLOAT32);
+
+    // bias is not used for shape inference
+  }
 
   const auto *options = op.builtin_options.AsBCQFullyConnectedOptions();
   node->weights_hidden_size(options->weights_hidden_size);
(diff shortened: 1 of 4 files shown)


compiler/luci/import/src/Nodes/CircleBCQFullyConnected.cpp
```diff
@@ -41,6 +43,7 @@ CircleNode *CircleBCQFullyConnectedGraphBuilder::build_node(const circle::Operat
   node->weights_scales(inputs[1]);
   node->weights_binary(inputs[2]);
   node->bias(inputs[3]);
+  node->weights_clusters(inputs[4]);
```
seanshpark, line 46: maybe add `if (args.op.inputs.size() >= 5)` or something?

compiler/luci/import/src/Nodes/CircleBCQGather.cpp
```diff
@@ -40,6 +42,7 @@ CircleNode *CircleBCQGatherGraphBuilder::build_node(const circle::OperatorT &op,
   node->input_scales(inputs[0]);
   node->input_binary(inputs[1]);
   node->indices(inputs[2]);
+  node->input_clusters(inputs[3]);
```
seanshpark, line 45: ditto

compiler/luci/import/src/Nodes/CircleBCQFullyConnected.cpp
llFreetimell: Hmm... if I understand correctly, that checking is already done at line 30 :)
```cpp
  if (args.op.inputs.size() != 4 && args.op.inputs.size() != 5)
    return false;
```

seanshpark: if `args.op.inputs.size() == 4 ` then `inputs[4]` will make out of bounds error

llFreetimell: @seanshpark Thanks for the sharp review :) I will fix related issues first and then ping again..!

compiler/luci/import/src/Nodes/CircleBCQFullyConnected.cpp
```diff
@@ -41,6 +41,16 @@ CircleNode *CircleBCQFullyConnectedGraphBuilder::build_node(const circle::Operat
...
+
+  // TODO Find and move to appropriate place for setting optional input
+  if (auto bias = dynamic_cast<luci::CircleOutputExclude *>(node->bias()))
+  {
+    // bias is not used for type inference, but node itself should have a type
+    bias->dtype(loco::DataType::FLOAT32);
+
+    // bias is not used for shape inference
```
seanshpark, line 52: comment seems duplicate

llFreetimell: It is different with above comment, `type inference` and `shape inference` :)

PR #2356, opened 2020-06-17
[luci] Separate DeadNodeQueryService

seanshpark:
This will separate DeadNodeQueryServiceImpl class to own file

### DIFF
--- a/compiler/luci/lang/src/DeadNodeQueryService.cpp
+++ b/compiler/luci/lang/src/DeadNodeQueryService.cpp
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


compiler/luci/lang/src/DeadNodeQueryService.cpp
```diff
@@ -0,0 +1,78 @@
...
+  auto output_nodes = std::set<loco::Node *>(output_nodes_vec.begin(), output_nodes_vec.end());
+  auto active_nodes = loco::active_nodes(output_nodes_vec);
+
+  if (active_nodes.find(node) != active_nodes.end())
+    return false;
+  // input and output nodes are not dead node even if it is not active.
+  if (input_nodes.find(node) != input_nodes.end())
+    return false;
+  if (output_nodes.find(node) != output_nodes.end())
+    return false;
```
jinevening, line 56: It seems that `active_nodes` always contains `output_nodes`. If so, this conditional statement would be redundant.

compiler/luci/lang/src/DeadNodeQueryService.cpp
```diff
@@ -0,0 +1,78 @@
...
+    if (circle_node->accept(&d))
+    {
+      assert(node->arity() == 1);
+      loco::Node *real_node = node->arg(0);
+      if (active_nodes.find(real_node) != active_nodes.end())
+        return false;
+      if (input_nodes.find(real_node) != input_nodes.end())
+        return false;
+      if (output_nodes.find(real_node) != output_nodes.end())
+        return false;
```
jinevening, line 71: ditto

compiler/luci/lang/src/DeadNodeQueryService.cpp
seanshpark: This change is to separate to a file and would be better to change in another PR.

jinevening: Ok. No problem for this PR.

Issue #2338, opened 2020-06-17
[onert] segmentation fault during nnfw_close_session()


glistening:
For model below,

![image](https://user-images.githubusercontent.com/12553213/84840426-7265e300-b07a-11ea-8899-a8b4ecb6881f.png)

onert got segmentation fault during `nnfw_close_session()`.

I've run on `x86_64` with c1aaf15.

```
$ Product/x86_64-linux.debug/out/bin/nnpackage_run --nnpackage d2-5/
... warmup 1 takes 0.308 ms
... run 1 takes 0.069 ms
Segmentation fault (core dumped)
```

This model has no problem with `tflite_run`.

nnpackage with golden data: [d2-5.tar.gz](https://github.com/Samsung/ONE/files/4789672/d2-5.tar.gz)

#### call stack

```
#0  __GI___libc_free (mem=0x40c63c9bc01cca5b) at malloc.c:2951
...
#5  0x00007ffff7b7befb in std::vector<int, std::allocator<int> >::~vector (this=0x6bac58, __in_chrg=<optimized out>) at /usr/include/c++/9/bits/stl_vector.h:680
#6  0x00007ffff3b522c4 in nnfw::cker::FCTempArena::~FCTempArena (this=0x6bac20, __in_chrg=<optimized out>) at /home/brian/z/ONE/compute/cker/include/cker/operation/FullyConnected.h:31
...
#52 0x00007ffff7b755c7 in nnfw_close_session (session=0x6b24e0) at /home/brian/z/ONE/runtime/onert/api/src/nnfw_api.cc:68
```

<details> <summary> full call stack </summary>

```
#0  __GI___libc_free (mem=0x40c63c9bc01cca5b) at malloc.c:2951
#1  0x00007ffff7b80364 in __gnu_cxx::new_allocator<int>::deallocate (this=0x6bac58, __p=0x40c63c9bc01cca5b) at /usr/include/c++/9/ext/new_allocator.h:128
#2  0x00007ffff7b7ee3c in std::allocator_traits<std::allocator<int> >::deallocate (__a=..., __p=0x40c63c9bc01cca5b, __n=2676019642441686) at /usr/include/c++/9/bits/alloc_traits.h:470
#3  0x00007ffff7b7d908 in std::_Vector_base<int, std::allocator<int> >::_M_deallocate (this=0x6bac58, __p=0x40c63c9bc01cca5b, __n=2676019642441686) at /usr/include/c++/9/bits/stl_vector.h:351
[...]

PR #2372, opened 2020-06-17
[onert] fix StridedSlice to use inferred shape

glistening:
It fixes StridedSlice to use inferred shape.

### DIFF
--- a/runtime/onert/backend/cpu/KernelGenerator.cc
+++ b/runtime/onert/backend/cpu/KernelGenerator.cc
@@ -933,12 +933,11 @@ void KernelGenerator::visit(const ir::operation::StridedSlice &node)
   auto begin_mask = node.param().begin_mask;
   auto end_mask = node.param().end_mask;
   auto shrink_axis_mask = node.param().shrink_axis_mask;
-  auto rank = node.param().rank;
 
   auto fn = std::make_unique<ops::StridedSliceLayer>();
 
   fn->configure(input_alloc, starts_alloc, ends_alloc, strides_alloc, output_alloc, begin_mask,
-                end_mask, shrink_axis_mask, rank);
+                end_mask, shrink_axis_mask);
 
   _return_fn = std::move(fn);
 }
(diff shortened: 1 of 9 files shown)


glistening:
It fixes https://github.com/Samsung/ONE/issues/2338#issuecomment-645084710.

```
$ Product/x86_64-linux.debug/out/bin/nnpackage_run --nnpackage d2-5
... warmup 1 takes 0.958 ms
... run 1 takes 0.169 ms
MODEL_LOAD   takes 4.863 ms
PREPARE      takes 15.028 ms
EXECUTE      takes 0.169 ms
- MEAN     :  0.169 ms
- MAX      :  0.169 ms
- MIN      :  0.169 ms
- GEOMEAN  :  0.169 ms
```

runtime/onert/backend/cpu/KernelGenerator.cc
```diff
@@ -933,12 +933,11 @@ void KernelGenerator::visit(const ir::operation::StridedSlice &node)
   auto begin_mask = node.param().begin_mask;
   auto end_mask = node.param().end_mask;
   auto shrink_axis_mask = node.param().shrink_axis_mask;
-  auto rank = node.param().rank;
 
   auto fn = std::make_unique<ops::StridedSliceLayer>();
 
   fn->configure(input_alloc, starts_alloc, ends_alloc, strides_alloc, output_alloc, begin_mask,
-                end_mask, shrink_axis_mask, rank);
+                end_mask, shrink_axis_mask);
```
ragmani, line 940: Maybe, you should change getting static rank to get dynamic rank

glistening: @ragmani https://github.com/Samsung/ONE/issues/2338#issuecomment-645084710 happens for static inference. I think dynamic case is also same - they need to use inferred shape, not the first given shape. Could you please explain more why two cases should be changed differently?

ragmani: You are right. I haven't seen it properly.

Issue #2293, opened 2020-06-16
[onert] Make nnfw_input_tensorinfo() work after calling nnfw_set_input_tensorinfo()


hyunsik-yoon:
while working for [large_test.tflite](https://github.sec.samsung.net/STAR/nnfw/issues/11334#issuecomment-719756), I found the this issue.

The following sequence should work:

assuming that input tensor 0's shape is [1, 2] in tflite file, currently line 4 has error.

- `nnfw_load_model_from_file()`
- `nnfw_prepare()`
- `nnfw_set_input_tensorinfo(0, [2, 2])`
- `nnfw_input_tensorinfo(0)`       --> this still returns [1, 2]
- `nnfw_run()`

Anyone who's interested in this work, please assign yourself and start working!
Thanks!

wateret:
As I worked on `nnfw_output_tensorinfo`, I will do this.

PR #2373, opened 2020-06-17
[nnfwapi] Update behavior of nnfw_input_tensorinfo

wateret:
As `nnfw_input_tensorinfo` works differently for each session state,
this commit makes it to work properly for PREPARED and FINISHED state.

For #2293


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -417,7 +417,7 @@ NNFW_STATUS nnfw_session::apply_tensorinfo(uint32_t index, nnfw_tensorinfo ti)
     // overwrite input shape with the shape from ti
     input.info().shape(new_shape);
   }
-  else // when called after nnfw_session::prepare()
+  else // when called before nnfw_session::prepare()
```
hyunsik-yoon, line 420: shouldn't this be `after`?

wateret:
@hyunsik-yoon Fixed, PTAL :)

PR #2097, opened 2020-06-10
[onert] Using mmap and munmap during model load

periannath:
This commit uses mmap and munmap to reduce peak memory usage
  - Loads model file using mmap at base_loader
  - Unmap mapped memory after weight and bias data are copied into CachedData

---

For issue: #2077

Issue #2077, opened 2020-06-10
[onert] Reducing peak memory usage during model loading


periannath:
onert consumes almost twice as much memory as the model size during model loading.

### Model loading process
1. Read a whole model file into memory (`_buffer` variable)
1. Load operators from `_buffer` and create Graph
1. Load weight and bias data and copy them into `CachedData` :point_left: **Peak memory**
1. Deallocate `_buffer` from memory

At step 3, weight and bias data are duplicated at `_buffer` and `CachedData`.

### How to reduce peak memory usage? (list of options)

1. Deallocate memory area for weight and bias data when it is copied into `CachedData`
   - Later phases are not affected
   - How to deallocate memory area for weight and bias data?
1. Using `mmap` to read model file and `CachedData` has `mmap` address
   - Deallocation using `munmap` is needed at dtor of `CachedData`
   - Any problem when nnapi is used?

periannath:
### Profiling memory usage for option 1

- Commit : #2097 
- Model : inception_v3_2018_04_27
- command
	```bash
	$ BACKENDS=cpu valgrind --pages-as-heap=yes --tool=massif --detailed-freq=1 ./Product/armv7l-linux.debug/out/bin/nnpackage_run --nnpackage ../nnpkg_tst/inception_v3_2018_04_27/
	```

Using valgrind to profile memory usage for `nnpackage_run`
Valgrind will terminate after model prepare because some neon instruction is not supported at valgrind

**onert master**

![mass_no_mmap_point](https://user-images.githubusercontent.com/10445076/84240913-ab153200-ab39-11ea-9896-968cdd8d61db.png)

**onert draft**

![massif_mmap_point](https://user-images.githubusercontent.com/10445076/84240895-a6e91480-ab39-11ea-913d-ce926f574096.png)

- Memory usage during model loading is almost same as model size

periannath:
### Measure model load time for option 1

**onert master**
```bash
$ BACKENDS=cpu ./Product/armv7l-linux.release/out/bin/nnpackage_run --nnpackage ../nnpkg_tst/inception_v3_2018_04_27
... run 0 takes 1263.64 ms
===================================
MODEL_LOAD takes 411.48 ms
PREPARE    takes 281.398 ms
EXECUTE    takes 
- Min:  1263.64 ms
- Max:  1263.64 ms
- Mean: 1263.64 ms
- GeoMean: 1263.65 ms
===================================
```


**onert draft**
```bash
[...]

PR #2385, opened 2020-06-18
[onert] Using munmap to release copied memory during model load

periannath:
- This commit uses munmap to eagerly unloads mmaped memory
  - Peak memory usage during model load will be reduced
  - Change type of `_base` from `char *` to `uint8_t *`
    - Flatbuffer returns `const char *` for buffer
    - Both pointers should have same type for pointer subtraction
    - Change type of `_base` to match pointer type

---

For issue : #2077
From draft #2097

### DIFF
--- a/runtime/onert/frontend/base_loader/include/base_loader.h
+++ b/runtime/onert/frontend/base_loader/include/base_loader.h
@@ -62,7 +62,7 @@ public:
    * @param graph reference on subgraphs
    */
   explicit BaseLoader(std::unique_ptr<ir::Subgraphs> &subgs)
-      : _base{nullptr}, _subgraphs(subgs), _model{nullptr}
+      : _base{nullptr}, _pagesize(getpagesize()), _subgraphs(subgs), _model{nullptr}
   {
   }
 
@@ -81,6 +81,7 @@ protected:
   ir::Activation convertActivation(ActivationFunctionType type);
   ir::DataType tensorTypeToDataType(TensorType type);
   ir::OperandIndex tensorIdxToOperandIdx(int32_t tensorIdx);
+  void deallocateMmappedArea(uint8_t *ptr, size_t size);
 
   // Create operands form tflite::Tensor
   ir::OperandIndex loadOperand(const Tensor *tensor, ir::Graph &subg);
@@ -170,7 +171,9 @@ protected:
 
 protected:
   // Base address for mapped region for loading (if needed)
-  char *_base;
+  uint8_t *_base;
+  // Memory page size
+  int32_t _pagesize;
   // Reference on loadable subgraphs
   std::unique_ptr<ir::Subgraphs> &_subgraphs;
   const Model *_model;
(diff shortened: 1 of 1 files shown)


runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -331,9 +352,12 @@ ir::OperandIndex BaseLoader<LoaderDomain, SpecificLoader>::loadOperand(const Ten
 
   // Constant tensors are indicated by non-empty data.
   const auto *data = _model->buffers()->Get(tensor->buffer())->data();
+
+  int32_t pagesize = getpagesize();
```
YongseopKim, line 356: I expect the pagesize would not be changed while running, I think. If so(and I'm not wrong), the pagesize could be a static member of this class?

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -270,6 +266,31 @@ BaseLoader<LoaderDomain, SpecificLoader>::BaseLoader::tensorIdxToOperandIdx(int3
   return isOptionalInputTensor(tensorIdx) ? ir::OperandIndex() : _tensor_to_operand[tensorIdx];
 }
 
+template <typename LoaderDomain, typename SpecificLoader>
+void BaseLoader<LoaderDomain, SpecificLoader>::BaseLoader::deallocateMmappedArea(uint8_t *ptr,
```
YongseopKim, line 270: I'd like that `UmmappedArea` seems better. It's just opinion.

hseok-oh:
@periannath Please check sign-off

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -270,6 +266,31 @@ BaseLoader<LoaderDomain, SpecificLoader>::BaseLoader::tensorIdxToOperandIdx(int3
...
+                                                                                 size_t size,
+                                                                                 int32_t pagesize)
+{
+  // Calculate offset from base address of mapped region
+  auto unaligned_offset_start = ptr - _base;
```
YongseopKim, line 275: Personally, `auto` keyword has been used too much, I think. For me, `unaligned_offset_start` could be `uint8_t*` or `uint32_t`. Below `aligned_offset_start` is calculated by division and multiplication. If so that could be `double`. (If that is your intention, ignore my comment.)

In short, how about using the exact type instead of `auto`(type deduction)?

periannath:
@hseok-oh Thanks. I'll add it soon.

runtime/onert/frontend/base_loader/include/base_loader.h
periannath: @YongseopKim I fount that pointer subtraction should use `ptrdiff_t` type. https://en.cppreference.com/w/cpp/types/ptrdiff_t.
I used auto keyword here because I don't know correct type for `ptrdiff_t` + `size_t`.
I'll find out correct type and use exact type here.

runtime/onert/frontend/base_loader/include/base_loader.h
periannath: @YongseopKim You're right. It will not change while running. I'll define `pagesize` as private member of BaseLoader. (I think it don't have to be static)

periannath:
@hseok-oh @YongseopKim I've updated this PR. PTAL.

Issue #1176, opened 2020-05-20
Compiler FE: Line coverage 


seanshpark:
Line coverage report until enabled in CI

Previous: #117

Related: `Compiler Frontend test count` #103

Add : Report of [daily-compiler-coverage](http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/)

---

### How-to run coverage test

1) configure and build
```
infra/scripts/configure_compiler_coverage.sh
CC=$(CC) CXX=$(CXX) ./nncc build -j6
```
2) run test
```
./nncc test
```
3) run coverage report
```
./nnas gen-coverage-report compiler
```
   - You'll see something like this when finished
   ```
   Writing directory view page.
   Overall coverage rate:
     lines......: 87.6% (29248 of 33404 lines)
     functions..: 85.5% (12496 of 14615 functions)
   ```
   - You may have to delete previous `coverage` report folder
4) open `coverage/html/index.html` file

hseok-oh:
Internal CI show daily line coverage report

Overview
http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/

Daily report
http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/1/cobertura/

seanshpark:
When adding new components to compiler we need to do updates to these files also
- `infra/scripts/configure_compiler_coverage.sh`
- `infra/nncc/command/utcount`

seanshpark:
Coverage report as of now
- http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/13/cobertura/

seanshpark:
Line coverage @ CI server http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/

seanshpark:
http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage

Closing...

PR #2386, opened 2020-06-18
[infra] revise configure_compiler_coverage

seanshpark:
This will revise configure_compiler_coverage script to include recent foder module


seanshpark:
Related #1176

Issue #103, opened 2020-04-20
Compiler Frontend test count


seanshpark:
Records of positive/negative unit test count of '20 projects

Related: `Compiler FE: Line coverage` #1176

seanshpark:
With #2066, `No tests were found!!!` won't appear.

seanshpark:
As of 2020-06-20

  | 2020-04-20 |   | 2020-06-19 |  
-- | -- | -- | -- | --
Module | Total | NEG | Total | NEG
angkor | 88 | 0 | 88 | 0
cwrap | 4 | 0 | 4 | 0
pepper-str | 4 | 0 | 4 | 0
pepper-strcast | 1 | 0 | 1 | 0
pp | 16 | 0 | 16 | 0
stdex | 4 | 0 | 4 | 0
oops | 3 | 0 | 3 | 0
pepper-assert | 0 | 0 | 0 | 0
hermes | 12 | 0 | 12 | 0
hermes-std | 2 | 0 | 2 | 0
loco | 148 | 0 | 154 | 3
locop | 5 | 0 | 5 | 0
locomotiv | 58 | 0 | 58 | 0
logo-core | 2 | 0 | 2 | 0
logo | 4 | 0 | 4 | 0
foder | - | - | 0 | 0
safemain | 0 | 0 | 0 | 0
mio-circle | 0 | 0 | 0 | 0
mio-tflite | 0 | 0 | 0 | 0
tflite2circle | 0 | 0 | 0 | 0
luci | 30 | 0 | 506 | 377
luci-interpreter | - | - | 55 | 0
luci-value-test | - | - | 0 | 0
record-minmax | - | - | 0 | 0
circle2circle | - | - | 1 | 1
  |   |   |   |  
TOTAL | 381 | 0 | 919 |  
새로 추가 | 0 |   | 538 |  
추가 POS : NEG |   |   | 157 | 381
POS - NEG |   |   |   | 224

seanshpark:
CC @lemmaa 

Test count started at April 20, which we can use ab2a31ca6f34d3b5fa6d5f4f4268a919b806177c.
This is not the exact commit but I think it won't give us big difference.

seanshpark:
Count of April 20 with TCM: Total 391, NEG 4

seanshpark:
Count of 3a0309f1f268cdcbb912027130316f498b25c4ce
- Total TCs: 979
- Positive TCs: 571 = +180
- Negative TCs: 408 = +404
[...]

PR #2387, opened 2020-06-18
[infra] revise utcount

seanshpark:
This will revise utcount to include recent added foder module


seanshpark:
Related #103

Issue #2088, opened 2020-06-10
Test for post-training quantization


jinevening:
**Updated 2020/06/24**: The test framework now tests the end-to-end quantization steps (fake quantization, record minmax, quantization).

We are implementing post-training quantization (#696). This issue is about `quantization-value-test`, which tests the quantized values of a circle model.

![image](https://user-images.githubusercontent.com/5449554/85508474-19282180-b62f-11ea-84b4-b2f6b899f00d.png)

Above figure shows the overview of `quantization-value-test`. More details about the test process is described below.

**Step 1. Fake quantization**
Run `circle2circle` with `--quantize_dequantize_weights` option.
Dump the fake-quantized model with `circle-tensordump`.
Compare the dumped model with the expected output in "expected_outputs/<model_name>/fake_quantization/<tensor_name>"
The expected output should include
 (1) values of weights (for conv, transposed_conv, depthwise_conv, fc layers)

**Step 2. Record moving avg of min and moving avg of max for activations**
Run `record-minmax` with the fake-quantized model (input data is saved in "test_inputs/<model_name>/<record_number>")
Dump the minmax-recorded model with `circle-tensordump`.
Compare the dumped model with the expected output in "expected_outputs/<model_name>/record_minmax/<tensor_name>"
The expected output should include
 (1) min/max of activations

**Step 3. Quantization**
Run `circle2circle` with `--quantize_with_minmax` option.
Dump the quantized model with `circle-tensordump`.
Compare the dumped model with the expected output in "expected_outputs/<model_name>/quantization/<tensor_name>"
The expected output should include
 (1) scale, zero point of activations
 (2) scale, zero point, values of weights
 (3) scale, values (weights) of bias

jinevening:
**Format of the tensor file (.h5)**

We use hdf5 file (.h5) to save tensors. Hierarchy of the file is as below.

```
Group "/"
  ㄴGroup <tensor_name>
      ㄴDataset "weights"
      ㄴDataset "zero_point" Shape (1, n)
      ㄴDataset "scale" Shape (1, n)
```
`tensor_name` : name of tensor (`/` is replaced with `_`)
`weights` : weights of the tensor
`zero_point` : zero points (shape: n = 1 for layer-wise quant, n = dimension of channel for channel-wise quant)
`scale` : scale (shape: same with zero_point)

Example: tensors related to a conv layer, channel-wise quant for filters and bias

```
Group "/"
  ㄴGroup "conv_filters"
[...]

PR #2390, opened 2020-06-18
[record-minmax-test] Revise input gen script

jinevening:
This revises the input generation script

---
Related to #2088


compiler/record-minmax-conversion-test/gen_h5_random_inputs.py
```diff
@@ -22,7 +22,6 @@
 
 # Build TFLite interpreter. (to get the information of model input)
 interpreter = tf.lite.Interpreter(model)
-interpreter.allocate_tensors()
```
mhs4670go, line 25: This line is removed because it it for interpreting, isn't it?

jinevening: The interpreter is used only for getting the input information, so tensors need not be allocated.

PR #2391, opened 2020-06-18
[onert] don't throw divide by zero error in float div

glistening:
divide by zero is well defined in IEEE specificaiton.
It should set the output as +Inf or -Inf, instead of throwing error.

### DIFF
--- a/compute/cker/include/cker/operation/BinaryArithmeticOps.h
+++ b/compute/cker/include/cker/operation/BinaryArithmeticOps.h
@@ -32,6 +32,16 @@ namespace cker
 
 namespace
 {
+template <typename T> T divide(const T &a, const T &b)
+{
+  if (b == 0)
+    throw std::runtime_error("Divide by zero");
+
+  return a / b;
+}
+
+template <> float divide<float>(const float &a, const float &b) { return a / b; }
+
 template <typename T>
 const std::function<T(const T &, const T &)> GetBinaryArtithmeticFn(BinaryArithmeticOpType type)
 {
@@ -51,13 +61,7 @@ const std::function<T(const T &, const T &)> GetBinaryArtithmeticFn(BinaryArithm
     }
     case BinaryArithmeticOpType::DIV:
     {
-      return [](const T &a, const T &b) -> T {
-        if (b == 0)
-        {
-          throw std::runtime_error("Divide by zero");
-        }
-        return a / b;
-      };
+      return divide<T>;
     }
     case BinaryArithmeticOpType::POW:
     {


glistening:
For example below, tflite does not check and throw exception even it encounters `0`.

image

compute/cker/include/cker/operation/BinaryArithmeticOps.h
```diff
@@ -51,13 +51,7 @@ const std::function<T(const T &, const T &)> GetBinaryArtithmeticFn(BinaryArithm
...
     case BinaryArithmeticOpType::DIV:
     {
-      return [](const T &a, const T &b) -> T {
-        if (b == 0)
-        {
-          throw std::runtime_error("Divide by zero");
-        }
-        return a / b;
-      };
+      return [](const T &a, const T &b) -> T { return a / b; };
```
chunseoklee, line 54: What If T is INT ?

wateret, line 54: What about

```suggestion
      return divide;
```

And have divide function as a static or something.

```cpp
template<T>
T divide(const T& a, const T &b)
{
        if (b == 0)	
        {	
          throw std::runtime_error("Divide by zero");	
        }	
        return a / b;	
      };
}

template <>
float divide(const float &a, const float &b)
{
  return a/b;
}
```

glistening:
@wateret, @chunseoklee 
I've introduced specialized `GetBinaryArtithmeticFn` for `float` to apply this change for only `float`.

compute/cker/include/cker/operation/BinaryArithmeticOps.h
glistening: @wateret I've tried, but I got 
```
In file included from /home/brian/z/ONE/runtime/onert/core/src/interp/operations/BinaryArithmeticOps.cc:17:
/home/brian/z/ONE/compute/cker/include/cker/operation/BinaryArithmeticOps.h:44:1: error: type qualifiers ignored on function return type [-Werror=ignored-qualifiers]
   44 | const float divide(const float &a, const float &b)
      | ^~~~~
/home/brian/z/ONE/compute/cker/include/cker/operation/BinaryArithmeticOps.h:44:13: error: template-id ‘divide<>’ for ‘const float nnfw::cker::divide(const float&, const float&)’ does not match any template declaration
   44 | const float divide(const float &a, const float &b)
      |             ^~~~~~
/home/brian/z/ONE/compute/cker/include/cker/operation/BinaryArithmeticOps.h:34:3: note: candidate is: ‘template<class T> T nnfw::cker::divide(const T&, const T&)’
   34 | T divide(const T& a, const T &b)`
```

glistening:
@wateret, @chunseoklee, @periannath I've updated again, PTAL.

compute/cker/include/cker/operation/BinaryArithmeticOps.h
wateret: Fixed the comment!

PR #2394, opened 2020-06-18
[luci-interpreter]Add L2Normalize Kernel

struss:
This commit add L2Normalize Kernel on luci-interpreter.


compiler/luci-interpreter/src/kernels/L2Normalize.h
```diff
@@ -0,0 +1,51 @@
...
+  L2Normalize(const Tensor *input, Tensor *output, const L2NormParams &params);
+
+  std::vector<const Tensor *> getInputTensors() const override { return {_input}; }
+  std::vector<Tensor *> getOutputTensors() const override { return {_output}; }
+
+  void configure() override;
+  void execute() const override;
+
+private:
+  inline void evalFloat() const;
```
s-barannikov, line 40: `inline` keyword is useless here.

compiler/luci-interpreter/src/kernels/L2Normalize.cpp
```diff
@@ -0,0 +1,83 @@
...
+      break;
+    case DataType::U8:
+      evalQuantized();
+      break;
+    default:
+      throw std::runtime_error("Unsupported type.");
+  }
+}
+
+inline void L2Normalize::evalFloat() const
```
s-barannikov, line 64: `inline` keyword is useless here.
What I meant by "inlining" in my previous comment is incorporating the function body into its caller (optional).
Please remove the keyword or inline the functions.

struss: Oh, Sorry. I will make a Optional change. Thanks for telling me that problem.

compiler/luci-interpreter/src/kernels/L2Normalize.h
struss: Thanks, I will remove this.

PR #2397, opened 2020-06-18
[luci-interpreter] Rework observing functionality support

s-barannikov:
Draft PR #1980.

Add a layer of abstraction to avoid dependency of runtime entities on IR entities.

### DIFF
--- a/compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
@@ -59,8 +59,14 @@ private:
   void createKernels(const loco::Graph *graph);
 
   const loco::Graph *_main_graph = nullptr;
-  std::unique_ptr<class TensorMap> _tensor_map;
-  std::unique_ptr<class KernelMap> _kernel_map;
+  std::unique_ptr<class TensorMap> _node_to_tensor;
+
+  // Kernels, in execution order.
+  std::vector<std::unique_ptr<class Kernel>> _kernels;
+
+  // Observer functionality support.
+  std::unique_ptr<struct RuntimeToIR> _runtime_to_ir;
+  std::unique_ptr<class Hook> _hook;
   std::vector<ExecutionObserver *> _observers;
 };
 
--- a/compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
@@ -66,7 +66,7 @@ private:
 
   // Observer functionality support.
   std::unique_ptr<struct RuntimeToIR> _runtime_to_ir;
-  std::unique_ptr<class Hook> _hook;
+  std::unique_ptr<class EventNotifier> _event_notifier;
   std::vector<ExecutionObserver *> _observers;
 };
 
--- a/compiler/luci-interpreter/src/core/Hook.h
+++ b/compiler/luci-interpreter/src/core/Hook.h
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef LUCI_INTERPRETER_CORE_HOOK_H
+#define LUCI_INTERPRETER_CORE_HOOK_H
+
+namespace luci_interpreter
+{
+
+// Used at execution stage to tell the interpreter that the runtime state has changed in some way.
+class Hook
+{
+public:
+  virtual ~Hook() = default;
+
+  virtual void postTensorWrite(Tensor *tensor) const = 0;
+};
+
+} // namespace luci_interpreter
+
+#endif // LUCI_INTERPRETER_CORE_HOOK_H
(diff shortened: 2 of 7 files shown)


compiler/luci-interpreter/src/core/Hook.h
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+#ifndef LUCI_INTERPRETER_CORE_HOOK_H
+#define LUCI_INTERPRETER_CORE_HOOK_H
+
+namespace luci_interpreter
+{
+
+// Used at execution stage to tell the interpreter that the runtime state has changed in some way.
+class Hook
```
s-barannikov, line 24: Please suggest a better name.

s-barannikov:
@jinevening 
Could you check that quantizator is still working? I can't run the tests because I have newer version of Ubuntu, which lacks outdated TensorFlow python packages.

jinevening:
> Could you check that quantizator is still working? I can't run the tests because I have newer version of Ubuntu, which lacks outdated TensorFlow python packages.

`record-minmax-conversion-test` runs without error.

BTW, the test runs python on `venv`, so it should be able to run on higher versions of Ubuntu (does it depend on Ubuntu version?).

s-barannikov:
@jinevening 
Strictly speaking, it depends on Python version.
Ubuntu 20.04 comes with Python 3.8.2 onboard, which has  2.2.0rc1, 2.2.0rc2, 2.2.0rc3, 2.2.0rc4, 2.2.0 TensorFlow package versions.
Downgrading python is not a solution for me.

compiler/luci-interpreter/src/core/Hook.h
jinevening: How about Notifier?

This class notifies an event (change of runtime state according to your comment) to the interpreter.

s-barannikov: > How about Notifier?

I also considered this option. `Notifier` sounds too ambigous. I can't say that `Hook` is more specific though :)

jinevening:
I'm curious if we can expose `_node_to_tensor` (using a getter), so that an observer can use it.

E.g., If we allow an observer to access `_node_to_tensor`, the observer will be able to do more things (e.g., getting tensor data of the node's inputs).

compiler/luci-interpreter/src/core/Hook.h
jinevening: `EventNotifier` sounds good to me :)

seanshpark:
> Ubuntu 20.04 comes with Python 3.8.2 onboard, which has 2.2.0rc1, 2.2.0rc2, 2.2.0rc3, 2.2.0rc4, 2.2.0 TensorFlow package versions.

Out default TF is 1.13.1

s-barannikov:
@seanshpark AFAIK TensorFlow versions are backward compatible. Our are there any issues?

seanshpark:
>TensorFlow versions are backward compatible. Our are there any issues?

I don't know.  You wrote

> I can't run the tests because I have newer version of Ubuntu, which lacks outdated TensorFlow python packages.

What is your problem?

s-barannikov:
@seanshpark I can only install 2.2.0 version and its release candidates. 1.13 is unavailable in the repo.
I suppose 2.2.0 would also work well for the purpose of testing, because versions shold be backward compatible, but I did not try.
We could at least make the version configurable via some cmake option. Right now it is hardcoded in requirements.txt.

compiler/luci-value-test/requirements.txt
```diff
@@ -1,2 +1,2 @@
-tensorflow==1.13.1
+tensorflow==2.2.0
```
jinevening, line 1: This change seems not relevant to this PR.

s-barannikov: Oh, sorry, I committed this accidently. I will revert it

seanshpark:
> I can only install 2.2.0 version and its release candidates. 

So can you change your platform that has 1.13.1 ?

compiler/luci-value-test/requirements.txt
jinevening: Now I saw the above comment. If tf version is backward compatible, it's ok to change the version.

s-barannikov:
@seanshpark Yes, I can. But we can't force external contributors to do the same.

seanshpark:
> Yes, I can. But we can't force external contributors to do the same.

My point is our target TF is 1.13.1 and we should work on a platform that supports this product.
Do you mean that we should support TF 2.2 for Ubuntu 20.04 users?

compiler/luci-value-test/requirements.txt
seanshpark: I don't think this is about backward compatibility. Please remove this change.
And this file is different from other files.

s-barannikov:
@seanshpark I don't know if it is 100% working solution. It shold be, but who knows.
It seems that it would be better to install the required version (1.13), but I don't know how to do that without downgrading OS or Python.

seanshpark:
>  downgrading OS or Python.

Have you tried docker with Ubuntu 18.04 ?

s-barannikov:
> > downgrading OS or Python.
> 
> Have you tried docker with Ubuntu 18.04 ?

No. I've never tried building the project in docker. I will try, thank you for suggestion.

compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
```diff
@@ -59,8 +59,14 @@ class Interpreter
...
-  std::unique_ptr<class TensorMap> _tensor_map;
-  std::unique_ptr<class KernelMap> _kernel_map;
+  std::unique_ptr<class TensorMap> _node_to_tensor;
+
+  // Kernels, in execution order.
+  std::vector<std::unique_ptr<class Kernel>> _kernels;
+
+  // Observer functionality support.
+  std::unique_ptr<struct RuntimeToIR> _runtime_to_ir;
+  std::unique_ptr<class EventNotifier> _event_notifier;
```
binarman, line 69: I didn't get why you need `RuntimeToIR` structure and `EventNotifier`, could you clarify?

s-barannikov: It is in the description. To make runtime entities independent of IR.
Please see the draft for details.

binarman: My question was not about purpose of these members, but about purpose of separate structures.
From my point EventNotifier can be inlined, This applies to RuntimeToIR too.

Well, Let's return to this discussion, when everything is working =)

PR #2398, opened 2020-06-18
[onert] Allow more than 4-D

ragmani:
This commit allows more than 4-D.
  - Between the same layouts : up to 6-D
  - Betwwen the different layouts : up to 4-D

### DIFF
--- a/runtime/onert/core/src/util/shapeinf/Permute.cc
+++ b/runtime/onert/core/src/util/shapeinf/Permute.cc
@@ -56,16 +56,21 @@ void DynamicInferer::visit(const ir::operation::Permute &op)
   // getting output shapes
   auto new_shape = input_shape;
   // Permute is a special operation that layouts of input/output may be different
+  assert(new_shape.rank() <= ir::Shape::MAX_RANK);
   if (new_shape.rank() >= 4)
   {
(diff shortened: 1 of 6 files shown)


ragmani:
For fix [the issue](https://github.sec.samsung.net/STAR/nnfw/issues/11487)

runtime/onert/core/src/util/shapeinf/Permute.cc
```diff
@@ -56,16 +56,21 @@ void DynamicInferer::visit(const ir::operation::Permute &op)
   // getting output shapes
   auto new_shape = input_shape;
   // Permute is a special operation that layouts of input/output may be different
+  assert(new_shape.rank() <= ir::Shape::MAX_RANK);
   if (new_shape.rank() >= 4)
```
hyunsik-yoon, line 60: should this be also changed to  the following? In case of rank 5, I don't think it is meaningful to convert NHWC NCHW conversion.
```suggestion
  if (new_shape.rank() == 4)
```

hyunsik-yoon: I guess this can be committed in next PR.

ragmani: It is meaningful for the assertions to check `new_shape.rank() == 4`

PR #2404, opened 2020-06-18
[onert] make nnpackage_run receive tensor shape

hyunsik-yoon:
This makes nnpakcage_run receive new tensor shapes to call nnfw_set_input_tensorinfo().

To run some model, calling nnfw_set_input_tensorinfo() is required.

### DIFF
--- a/tests/tools/nnpackage_run/src/args.cc
+++ b/tests/tools/nnpackage_run/src/args.cc
@@ -130,11 +130,11 @@ void Args::Initialize(void)
          "{exec}-{nnpkg}-{backend}.csv will be generated.\n"
          "e.g. nnpackage_run-UNIT_Add_000-acl_cl.csv.\n"
          "{nnpkg} name may be changed to realpath if you use symbolic-link.")
-    ("shape_compile", po::value<std::string>()->default_value("[]"),
+    ("shape_prepare", po::value<std::string>()->default_value("[]"),
          "set shape of specified tensor before compilation\n"
          "e.g. '[0, [1, 2], 2, []]' to set 0th tensor to [1, 2] and 2nd tensor to [].\n")
-    ("shape_exec", po::value<std::string>()->default_value("[]"),
-         "set shape of specified tensor right before execution\n"
+    ("shape_run", po::value<std::string>()->default_value("[]"),
+         "set shape of specified tensor right before running\n"
          "e.g. '[1, [1, 2]]` to set 1st tensor to [1, 2].\n")
 
     ;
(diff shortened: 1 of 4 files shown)


tests/tools/nnpackage_run/src/args.cc
```diff
@@ -92,6 +130,13 @@ void Args::Initialize(void)
          "{exec}-{nnpkg}-{backend}.csv will be generated.\n"
          "e.g. nnpackage_run-UNIT_Add_000-acl_cl.csv.\n"
          "{nnpkg} name may be changed to realpath if you use symbolic-link.")
+    ("shape_compile", po::value<std::string>()->default_value("[]"),
+         "set shape of specified tensor before compilation\n"
+         "e.g. '[0, [1, 2], 2, []]' to set 0th tensor to [1, 2] and 2nd tensor to [].\n")
+    ("shape_exec", po::value<std::string>()->default_value("[]"),
```
glistening, line 136: @hyunsik-yoon What about `shape_prepare` and `shape_run`? Our API provides `nnfw_prepare()` and `nnfw_run()` and these values are applied before corresponding API call.

tests/tools/nnpackage_run/src/args.cc
```diff
@@ -54,6 +54,44 @@ std::unordered_map<uint32_t, Json::Value> argArrayToMap(const Json::Value &jsonv
...
+{
+  Json::Value root;
+  Json::Reader reader;
+  if (!reader.parse(shape_str, root, false))
+  {
+    std::cerr << "Invalid JSON format for output_sizes \"" << shape_str << "\"\n";
+    exit(1);
+  }
+
+  auto arg_map = argArrayToMap(root);
```
glistening, line 68: @hyunsik-yoon `argArrayToMap` returns `std::unordered_map<uint32_t, Json::Value>`, and `TensorShapeMap = std::unordered_map<int, std::vector<int>>`. What happens if `key` is bigger than int32 max? Perhaps we may change the return type of argArrayToMap to <int32_t, ...> since it is only used for setting `output_size`.

tests/tools/nnpackage_run/src/args.cc
hyunsik-yoon: Thanks. Fixed.

tests/tools/nnpackage_run/src/args.cc
hyunsik-yoon: @glistening, Fixed `TensorShapeMap` to use key as `uint32_t`. :-)

Issue #2393, opened 2020-06-18
[onert] Support new_shape option for reshape operation


periannath:
`Reshape` operation has `new_shape` option.

**Reshape**
```bash
Inputs {
  0: a tensor
  1: ignored
}
Outputs {
  0: a tensor with the same elements as the input but with the new shape
}
Options {
  new_shape
}
```

From : https://www.tensorflow.org/lite/guide/ops_compatibility

This option is ignored in onert.

Let's support `new_shape` option for `reshape` operation.

- [x] Add `new_shape` param in `reshape` IR #2405 
- [x] Reads `new_shape` option in base_loader #2405
- [x] Update shape inference for reshape #2467

hyunsik-yoon:
AFAIK, 
if the second param is const, this option and the shape of output tensor was same.
if the second param is _not_ const, i am not sure what the option value will be.

seanshpark:
Please refer res/TensorFlowLiteRecipes/Reshape_000 and other Reshape that compile have met with this.

periannath:
@seanshpark Thanks! I'll check them.

periannath:
Done.

PR #2405, opened 2020-06-18
[onert] Support new_shape option for reshape opration

periannath:
This commit supports new_shape option for reshape opration
- Add `new_shape` param in reshape IR
- Read `new_shape` option in base_loader

---

For issue : #2393

### DIFF
--- a/runtime/onert/core/include/ir/operation/Reshape.h
+++ b/runtime/onert/core/include/ir/operation/Reshape.h
@@ -37,12 +37,22 @@ public:
     SHAPE = 1
   };
 
+  struct Param
+  {
+    std::vector<int32_t> new_shape;
+  };
+
 public:
-  Reshape(const OperandIndexSequence &inputs, const OperandIndexSequence &outputs);
+  Reshape(const OperandIndexSequence &inputs, const OperandIndexSequence &outputs,
+          const Param &param);
 
 public:
   void accept(OperationVisitor &v) const override;
   OpCode opcode() const final { return OpCode::Reshape; }
+  const Param &param() const { return _param; }
+
+private:
+  Param _param;
 };
 
 } // namespace operation
--- a/runtime/onert/frontend/nnapi/wrapper/OperationFactory.cc
+++ b/runtime/onert/frontend/nnapi/wrapper/OperationFactory.cc
@@ -365,7 +365,9 @@ OperationFactory::OperationFactory()
     OperandIndexSequence inputs{init_param.inputs[0], init_param.inputs[1]};
     OperandIndexSequence outputs{init_param.outputs[0]};
 
-    return new operation::Reshape{inputs, outputs};
+    operation::Reshape::Param param{};
+
+    return new operation::Reshape{inputs, outputs, param};
   };
 
   _map[ANEURALNETWORKS_FULLY_CONNECTED] = [](const OperationFactory::Param &init_param,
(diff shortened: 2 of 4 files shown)


glistening:
@periannath It introduces `new_shape` param in IR. It changes both frontend. What about changing title of PR and commit?

runtime/onert/core/include/ir/operation/Reshape.h
```diff
@@ -37,12 +37,25 @@ class Reshape : public Operation
     SHAPE = 1
   };
 
+  struct Param
+  {
+    // Please see tensorflow/lite/c/builtin_op_data.h and reshape.cc.
+    // tensorflow lite supports only for ndim <= 8.
+    int shape[8];
+    int num_dimensions;
```
ragmani, line 45: How about change these param to a `std::vector`?

periannath: @ragmani I just copied that code from TFLite, and they use above structure because they can't use dynamic data in Params. I agree with your opinion that it is good to use `std::vector`. I'll update it soon.

periannath:
@glistening @ragmani I've updated. PTAL

runtime/onert/frontend/nnapi/wrapper/OperationFactory.cc
```diff
@@ -365,7 +365,9 @@ OperationFactory::OperationFactory()
     OperandIndexSequence inputs{init_param.inputs[0], init_param.inputs[1]};
     OperandIndexSequence outputs{init_param.outputs[0]};
 
-    return new operation::Reshape{inputs, outputs};
+    operation::Reshape::Param param{};
+
+    return new operation::Reshape{inputs, outputs, param};
```
glistening, line 370: @periannath You don't initialize `param.new_shape`. To complete `nnapi` frontend, you have to modify `nnapi_delegate.cpp` also.

periannath: @glistening I've checked NNAPI spec and `new_shape` option is not specified in NNAPI. I just added params at there since constructor of `Reshape` needs argument param. If `new_shape` option is not specified, reshape will use second input as shape of output tensor.

runtime/onert/core/include/ir/operation/Reshape.h
```diff
@@ -37,12 +37,22 @@ class Reshape : public Operation
     SHAPE = 1
   };
 
+  struct Param
```
hseok-oh, line 40: IMO, it is strange because `new_shape` is duplicated as input and parameter.

hseok-oh: @periannath Could you explain me reason of duplication `new_shape`?

periannath: @hseok-oh It is specified in TFLite specification as I described in #2393. I and @ragmani think `new_shape` param is defined for constant shape and `shape` tensor is defined for dynamic shape. However, it seems `shape` tensor is used for both cases. I've uploaded this PR to support `reshape` operator without `shape` tensor.

hseok-oh: I've checked tensorflow lite kernel. Tensorflow lite check number of inputs and use parameter if node has only one input, otherwise use second input. IMO we should use similar implementation because we have IR definition based on tensorflow lite schema.

PR #2415, opened 2020-06-18
[onert] Fix implicit instantiation of undefined template in OpCode

ragmani:
Fix https://github.com/Samsung/ONE/issues/566#issuecomment-645876524

This commit Fix implicit instantiation of undefined template in OpCode.


wateret:
Merging this before tests as it is urgent.

PR #2419, opened 2020-06-18
[onert] Remove rank from params of ir for operations

ragmani:
This commit removes rank from params of ir for operations.
  - Remove rank param from ir
  - Replace using rank with getting shape in backends
  - Adjust shape inference of Pack and Unpack


wateret:
I'm really glad that we can remove `rank` params, but AFAIR param `rank` was introduced as a workaround for ACL. So I wonder if using backend shape is good for ACL now.

ragmani:
@wateret 
Yes, the param rank was introduced as a workaround because of the shape changes and permutations of only-4d. This param must be removed to support dynamic tensor. And I think it would be okay to remove this param if we can do not the shape changes. So I'm going to remove `ShapeFixer` from `onert`

PR #2424, opened 2020-06-18
[onert] fix Range operation in cpu backend

dayo09:
- instead of using pointer as T, use value as T

### DIFF
--- a/compute/cker/include/cker/operation/Range.h
+++ b/compute/cker/include/cker/operation/Range.h
@@ -28,23 +28,36 @@ namespace cker
 {
 template <typename T> inline int GetSize(T start, T limit, T delta)
 {
+  if (!((start > limit && delta < 0) || (start < limit && delta > 0)))
+  {
+    throw std::runtime_error("Range: invalid input values");
+  }
+
   int size = (std::is_integral<T>::value
-                  ? ((std::abs(*limit - *start) + std::abs(*delta) - 1) / std::abs(*delta))
-                  : std::ceil(std::abs((*limit - *start) / *delta)));
+                  ? ((std::abs(limit - start) + std::abs(delta) - 1) / std::abs(delta))
+                  : std::ceil(std::abs((limit - start) / delta)));
   return size;
 }
+
 template <typename T>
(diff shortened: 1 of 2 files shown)


dayo09:
This operator doesn't have any tests. It isn't within nnapi, so I think even oneapi test should be updated.

hyunsik-yoon:
1. can you also mention the `start` value change into the body of PR message? I was confused with this PR since some lines are not relate to template type change.

2. can you also add parent issue?

Thanks!

ragmani:
@hyunsik-yoon 
> 1. can you also mention the `start` value change into the body of PR message? I was confused with this PR since some lines (to fix `start`) are not relate to template type change.

- Before
`T start = start_data;` <---- T is `float *` or `int32_t *`
- After
`T value = start_value;` <---- T is `float` or `int32_t`

ragmani:
> This operator doesn't have any tests. It isn't within nnapi, so I think even oneapi test should be updated.

@dayo09 
This operator was introduced into `NeuralNetworksEx.h` and have nnapi tests in `tests/nnapi/specs/Ex/`

hyunsik-yoon:
@dayo09 
I guess this was written for https://github.sec.samsung.net/STAR/nnfw/issues/11494
and could you mention this fixes #11494?

hyunsik-yoon:
FYI, the test still fails
```
odroid@odroid:~/work/1/ONE$ BACKENDS=cpu ${ONE_PATH}/Product/out/bin/nnpackage_run ${MODEL_ZOO_ENMT_PATH}/subg_0_0_191 --shape_run '[0, [1, 10]]' --load ${MODEL_ZOO_ENMT_PATH}/large_test_input_0.h5
nnpackage_run: /home/eric/work/1/ONE/compute/cker/include/cker/operation/Range.h:57: void nnfw::cker::Range(const T*, const T*, const T*, T*) [with T = int]: Assertion `value < limit_value' failed.
Aborted
```

both `value` and `limit_value` are 10

dayo09:
@hyunsik-yoon I think that is wrong assert, you can just delete it.

ragmani:
I think `num_elements` value is different from actual the num of elements.

PR #2428, opened 2020-06-18
[luci] quantize convolution's weight to channel-wise uint8

meejeong:
- apply to quantize convolution's weights and bias to channel-wise
- datatype : uint8
- asymmetric quantization

Related to: #696, https://github.sec.samsung.net/STAR/nnapps/issues/379

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -110,8 +110,8 @@ std::pair<loco::Node *, loco::Node *> get_input_weight_of_bias(CircleNode *node)
   return std::make_pair(nullptr, nullptr);
 }
 
-void asym_quant_bias(CircleConst *node, float input_scale, float weight_scale,
-                     float *scaling_factor, int64_t *zp)
+void asym_quant_bias_per_layer(CircleConst *node, float input_scale, float weight_scale,
+                               float *scaling_factor, int64_t *zp)
 {
   float scale = input_scale * weight_scale;
   const float scaling_factor_inv = (scale == 0) ? 0 : 1.0 / scale;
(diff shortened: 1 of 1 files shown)


seanshpark:
@meejeong , there seems to be multiple changes here
- new implementations
- change method names
- and more ? :)

it would be better to split these but you can go as-is.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -137,6 +137,97 @@ void asym_quant_bias(CircleConst *node, float input_scale, float weight_scale,
...
+{
+  float scale{0};
+
+  uint32_t size = node->size<loco::DataType::FLOAT32>();
+  std::vector<int32_t> quantized_values(size);
+
+  for (uint32_t i = 0; i < size; ++i)
+  {
+    scale = input_scale * weight_scale[i];
+    scaling_factor[i] = (scale == 0) ? 0 : 1.0 / scale;
```
jinevening, line 214: This looks like the inverse of scaling factor.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -290,7 +431,32 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<bool>
...
-      // will be implemented.
+      assert(input->quantparam()->scale.size() == 1); // input scale's layer-wise
+      auto input_scale = input->quantparam()->scale[0];
+
+      assert(weight->quantparam() != nullptr); // weight scale's channel-wise
+      auto weight_scale = weight->quantparam()->scale;
+
+      auto circle_const = loco::must_cast<luci::CircleConst *>(node);
+
+      uint32_t size = circle_const->size<loco::DataType::FLOAT32>();
```
jinevening, line 442: The number of channels is the same with the size of bias (am I right?). If so, can you add an assert here?
```
assert(size == weight_scale.size());
```

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: `asym_quant_bias_per_layer` saves non-inversed scale as is.

meejeong:
@seanshpark , I uploaded ```new implementations``` and ```change method names``` together in this PR :-(. Next time, I will take care about it. Always thank you for your comment

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
meejeong: In case of ```asym_quant_bias_per_layer```, inversed scale is also used as [here](https://github.com/Samsung/ONE/blob/c6d68fb491d21bdc669024036f8455e378265e16/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp#L116-L125)

meejeong: > This looks like the inverse of scaling factor. Is it intended?

For 8bit quantization, we used asymmetric quantization method.
So, 
we use equation as below
```
Out = Sin*Sw/Sout ((Input - Zin) * (Weight - Zw) + Bias/Sw*Sin)
```

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
meejeong: I will apply it

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: > In case of asym_quant_bias_per_layer, inversed scale is also used as here

It is used, but the saved scale is the non-inversed one.
https://github.com/Samsung/ONE/blob/c6d68fb491d21bdc669024036f8455e378265e16/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp#L136

meejeong: oh, I see. I didn't catch up.

mhs4670go:
(Optional) When you have to change the codes, you may just add a new commit instead of force-push if its change is not trivial. Because reviewers might have trouble recognizing what part has been changed.

PR #2432, opened 2020-06-19
[onert] Fix Range op's assert() and adds tests for `Range` op

hyunsik-yoon:
This fixes range's assert and adds tests.

for https://github.com/Samsung/ONE/pull/2424#issuecomment-646397867

### DIFF
--- a/compute/cker/include/cker/operation/Range.h
+++ b/compute/cker/include/cker/operation/Range.h
@@ -21,6 +21,7 @@
 #include "cker/Shape.h"
 
 #include <cmath>
+#include <stdexcept>
 
 namespace nnfw
 {
@@ -53,8 +54,6 @@ inline void Range(const T *start_data, const T *limit_data, const T *delta_data,
   {
     output_data[i] = value;
     value += delta_value;
-
-    assert(value < limit_value);
   }
 }
 
--- a/compute/test/CMakeLists.txt
+++ b/compute/test/CMakeLists.txt
@@ -1,0 +1,13 @@
+set(TEST_COMPUTE test_compute)
+
+file(GLOB_RECURSE TESTS "*.cc")
+
+add_executable(${TEST_COMPUTE} ${TESTS})
+
+target_link_libraries(${TEST_COMPUTE} nnfw_lib_cker)
+target_link_libraries(${TEST_COMPUTE} gtest)
+target_link_libraries(${TEST_COMPUTE} gtest_main)
+target_link_libraries(${TEST_COMPUTE} ${LIB_PTHREAD} dl)
+add_test(${TEST_COMPUTE} ${TEST_COMPUTE})
+
+install(TARGETS ${TEST_COMPUTE} DESTINATION unittest)
--- a/compute/test/cker/Range.cc
+++ b/compute/test/cker/Range.cc
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2019 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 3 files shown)


hyunsik-yoon:
I have one curiosity while writing tests.

Could  the type of  start, limit, delta be different like the following?
```
   const int start = 0;
   const int limit = 10;
   const float delta = 1.5;
```
Similar example is in https://www.tensorflow.org/api_docs/python/tf/range (the second one) although it does not mention the type explicitly.

hyunsik-yoon:
@hseok-oh I added a test for cker. Could you check if the dir is OK? Thanks.

hyunsik-yoon:
Ah. @hseok-oh is day off today. @wateret, @chunseoklee could you check?

compute/test/cker/Range.cc
```diff
@@ -0,0 +1,72 @@
...
+    const float limit = 1;
+    const float delta = -0.5;
+    std::vector<float> expected = {
+        3, 2.5, 2, 1.5,
+    };
+    std::vector<float> actual(expected.size());
+    nnfw::cker::Range<float>(&start, &limit, &delta, actual.data());
+
+    for (int i = 0; i < actual.size(); i++)
+      ASSERT_EQ(actual[i], expected[i]);
```
wateret, line 58: What about using `ASSERT_FLOAT_EQ` for floating point value comparison? It could make an error when delta value is different.

wateret, line 58: What about using `ASSERT_FLOAT_EQ` for floating point value comparison? It could make an error when delta value is different.

compute/test/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+
+add_executable(${TEST_COMPUTE} ${TESTS})
+
+target_link_libraries(${TEST_COMPUTE} nnfw_lib_cker)
+target_link_libraries(${TEST_COMPUTE} gtest)
+target_link_libraries(${TEST_COMPUTE} gtest_main)
+target_link_libraries(${TEST_COMPUTE} ${LIB_PTHREAD} dl)
+add_test(${TEST_COMPUTE} ${TEST_COMPUTE})
+
+install(TARGETS ${TEST_COMPUTE} DESTINATION unittest)
```
wateret, line 13: Yes, all the binaries installed in `unittest` will run on CI.

FYI it will run multiple times with different executors (this is out of scope, so never mind 🥴)

compute/test/cker/Range.cc
hyunsik-yoon: fixed. thanks.

compute/cker/include/cker/operation/Range.h
```diff
@@ -53,8 +54,6 @@ inline void Range(const T *start_data, const T *limit_data, const T *delta_data,
   {
     output_data[i] = value;
     value += delta_value;
-
-    assert(value < limit_value);
```
wateret, line 57: Can I ask why you removed this? Wouldn't this be double-check `GetSize` result?

ragmani: If `delta_value` is a negative value, this assertion makes an error.

hyunsik-yoon: also 
considering [start = 1, delta 1, limit = 2]
after 1st iteration output[0] = 1, value == 2, so assert fails.

wateret: Oh then this assertion should have been at the top of the loop. And should check if `delta_value` is negative or positive.

dayo09: @wateret 
There is this line.
```c++
//compute/cker/include/cker/operation/Range.h
//line 36-39
if (!((start > limit && delta < 0) || (start < limit && delta > 0)))
  {	  {
    throw std::runtime_error("Range: invalid input values");
  }	  }
```

ragmani: The assertion is optional and a defensive code. If there is no change, it will be filtered by the code you said. So you don't have to add it if you think it is duplicated.

PR #2271, opened 2020-06-15
[onert] Add custom(broadcast_to) operation.

Hyunjun85:
1. Add Custom op.
2. Add its static tensor TC.
3. Add shape inference and test case.

Issue: https://github.com/Samsung/ONE/issues/2265

ONE-DCO-1.0-Signed-off-by: H Kim <hyunjun2.kim@samsung.com>

PR #2433, opened 2020-06-19
[onert] Add custom(Addv2) operation

KimDongEon:
addv2 call add operation.

Signed-off-by: Dongeon kim good.kim@samsung.com

### DIFF
--- a/runtime/nnapi-header/include/NeuralNetworksEx.h
+++ b/runtime/nnapi-header/include/NeuralNetworksEx.h
@@ -518,7 +518,42 @@ typedef enum {
    * Outputs:
    * * 0: An output tensor.
    */
-  ANEURALNETWORKS_EINSUM_EX = 50037
+  ANEURALNETWORKS_EINSUM_EX = 50037,
+
+  /** Adds two tensors, element-wise.
+   *
+   * Takes two input tensors of identical type and compatible dimensions. The output
+   * is the sum of both input tensors, optionally modified by an activation function.
+   *
+   * Two dimensions are compatible when:
+   *     1. they are equal, or
+   *     2. one of them is 1
+   *
+   * The size of the output is the maximum size along each dimension of the input operands.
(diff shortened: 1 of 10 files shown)


runtime/nnapi-header/include/NeuralNetworksEx.h
```diff
@@ -518,7 +518,42 @@ typedef enum {
...
+   *
+   * Inputs:
+   * * 0: A tensor.
+   * * 1: A tensor of the same type, and compatible dimensions as input0.
+   * * 2: An INT32 value, and has to be one of the {@link FuseCode} values.
+   *      Specifies the activation to invoke on the result of each addition.
+   *
+   * Outputs:
+   * * 0: The sum, a tensor of the same type as input0.
+   */
```
Hyunjun85, line 555: It's good to explain why you pass 50038.

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -95,6 +95,7 @@ template <typename LoaderDomain, typename SpecificLoader> class BaseLoader
   template <typename Param> void loadPool2D(Param &param, const Pool2DOptions *options);
 
   // Operations
+  void LoadAddV2(const Operator *op, ir::Graph &subg);
```
Hyunjun85, line 98: Where is its implementation?
it shows only forward declare.

KimDongEon: In the loadCustom function, AddV2 calls loadAdd(). Therefore, we will remove LoadAddV2.

runtime/nnapi-header/include/NeuralNetworksEx.h
KimDongEon: The enum value 50038 is used in PR #2271. Therefore, 50039 is used in this PR.

PR #2434, opened 2020-06-19
[luci-interpreter] Add RuntimeModule and RuntimeGraph classes

s-barannikov:
Draft PR #1980.

Add `RuntimeModule` and `RuntimeGraph` classes.


compiler/luci-interpreter/src/Interpreter.cpp
```diff
@@ -250,7 +271,7 @@ void Interpreter::interpret()
     kernel->execute();
```
binarman, line 271: So, new classes are not used in this PR?

compiler/luci-interpreter/src/Interpreter.cpp
```diff
@@ -171,31 +192,30 @@ class HookImpl final : public EventNotifier
...
-  RuntimeToIR &_loader_map;
+  RuntimeToIR &_runtime_to_ir;
   const std::vector<ExecutionObserver *> &_observers;
 };
 
 Interpreter::Interpreter(const luci::Module *module)
 {
   _runtime_to_ir = std::make_unique<RuntimeToIR>();
-  _event_notifier = std::make_unique<HookImpl>(*_runtime_to_ir, _observers);
+  _event_notifier = std::make_unique<EventNotifierImpl>(*_runtime_to_ir, _observers);
```
jinevening, line 207: If runtime module is used in the `interpret` method, `Interpreter` may not need `_event_notifier`.

compiler/luci-interpreter/src/core/RuntimeGraph.h
```diff
@@ -0,0 +1,62 @@
...
+  const std::vector<Tensor *> &getOutputTensors() const { return _output_tensors; }
+
+  void addKernel(std::unique_ptr<Kernel> &&kernel);
+
+  void configure();
+
+  void execute() const;
+
+private:
+  RuntimeModule *_owning_module;
```
jinevening, line 51: (optional) In my opinion, just `_module` is enough. But it is ok to keep the current name.

compiler/luci-interpreter/src/core/RuntimeModule.h
```diff
@@ -0,0 +1,61 @@
...
+
+  void configure() { getMainGraph()->configure(); }
+
+  void execute() const { getMainGraph()->execute(); }
+
+private:
+  RuntimeGraph *getMainGraph() const { return _graphs[0].get(); }
+
+  EventNotifier *const _event_notifier;
+  std::vector<std::unique_ptr<RuntimeGraph>> _graphs;
```
jinevening, line 56: Do we need to save the pointers?

compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
```diff
@@ -58,8 +59,9 @@ class Interpreter
   void createTensors(const loco::Graph *graph);
   void createKernels(const loco::Graph *graph);
 
-  const loco::Graph *_main_graph = nullptr;
-  std::unique_ptr<class TensorMap> _node_to_tensor;
+  std::unique_ptr<class RuntimeModule> _runtime_module;
+  class RuntimeGraph *_main_runtime_graph;
```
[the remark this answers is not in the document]
s-barannikov: Yes, this is a temporary solution. It will be removed when ModuleLoader is introduced.

compiler/luci-interpreter/src/Interpreter.cpp
```diff
@@ -137,14 +137,35 @@ void Interpreter::createTensors(const loco::Graph *graph)
...
     }
 
+    _node_to_tensor.emplace(node, tensor.get());
     _runtime_to_ir->tensor_to_node.emplace(tensor.get(), node);
-    _node_to_tensor->setTensor(node, std::move(tensor));
+
+    _main_runtime_graph->addTensor(std::move(tensor));
   }
+
+  // Set input tensors.
```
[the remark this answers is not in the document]
s-barannikov: It is only 60 lines :)
I will extract part of the code to make it even smaller.

compiler/luci-interpreter/src/Interpreter.cpp
```diff
@@ -137,14 +137,35 @@ void Interpreter::createTensors(const loco::Graph *graph)
...
+    input_tensors[i] = _node_to_tensor.at(input_nodes[i]);
+  }
+  _main_runtime_graph->setInputTensors(input_tensors);
+
+  // Set output tensors.
+  auto output_nodes = loco::output_nodes(const_cast<loco::Graph *>(graph));
+  std::vector<Tensor *> output_tensors(output_nodes.size());
+  for (size_t i = 0; i < output_nodes.size(); ++i)
+  {
+    const auto *node = loco::must_cast<const luci::CircleOutput *>(output_nodes[i]);
```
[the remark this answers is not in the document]
s-barannikov: It is not a check. I have to cast it to `CircleOutput` to be able to call `node->from()` method on the next line.

compiler/luci-interpreter/src/Interpreter.cpp
s-barannikov: It is a horrible mistake, thank you.

compiler/luci-interpreter/src/Interpreter.cpp
s-barannikov: `Interpreter` is the owner of `EventNotifierImpl` instance, and `RuntimeModule` accesses it via `EventNotifier` interface.
`EventNotifierImpl` cannot be moved into `RuntimeModule`, because `EventNotifierImpl` depends on IR, while `core` component must not.
(If I understood you correctly.)

compiler/luci-interpreter/src/core/RuntimeModule.h
s-barannikov: Yes. The source `Module` and `Graph`s are loaded into the "runtime" and then discarded. This is done in `Interpreter`'s constructor.
`Interpreter::execute` works with "loaded" graphs (`RuntimeModule` / `RuntimeGraph` classes), so I have to store them in order to use them.

compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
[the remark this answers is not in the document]
s-barannikov: I added a TODO comment.

compiler/luci-interpreter/src/Interpreter.cpp
s-barannikov: Fixed

compiler/luci-interpreter/src/Interpreter.cpp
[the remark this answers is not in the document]
s-barannikov: Fixed

s-barannikov:
@jinevening @binarman PTAL

compiler/luci-interpreter/src/Interpreter.cpp
jinevening: > EventNotifierImpl cannot be moved into RuntimeModule, because EventNotifierImpl depends on IR, while core component must not.

I'm not sure I understood the answer. CMakeLists.txt in the `src/core` directory links luci_lang (this is IR, isn't it?) to luci_interpreter_core (it is used in KernelParams.h).

You may have some design principles for luci-interpreter, but it is difficult for me to understand them without explanation (I'm not very familiar with cpp software designs). Can you explain a little bit more about the restrictions or principles of luci-interpreter? Even a brief explanation would be greatly helpful for me to review future PRs.

compiler/luci-interpreter/src/core/RuntimeModule.h
jinevening: Sorry, my point was that how about just using `std::vector<RuntimeGraph>`.

compiler/luci-interpreter/src/Interpreter.cpp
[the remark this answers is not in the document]
binarman: Ok, I just was confused because loco Node has method `arg`, so you can write simple `output_nodes[i]->arg(0)`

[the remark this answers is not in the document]
s-barannikov: I did not know this when I initially wrote this code, @jinevening pointed this out later.
But I think some additional checks will not do any harm.

s-barannikov:
@jinevening 
> CMakeLists.txt in the `src/core` directory links luci_lang (this is IR, isn't it?) to luci_interpreter_core (it is used in KernelParams.h).

Unfortunately, yes. `Interpreter` reuses `DataType` class and  `Padding`, and `Activation` enums from the IR. Ideally, they should be replicated in the `Interpreter`, but they are currently not.

> Can you explain a little bit more about the restrictions or principles of luci-interpreter?

I am planning to prepare design diagram this week (probably will be ready on next monday). It should lit some light on the design decisions.
I should have done this some time ago. Sorry about this.

> Even a brief explanation would be greatly helpful for me to review future PRs.

Currently, there are two main components, `Interpreter` and `core`. `Loader` component is going to be introduced.

* `Interpreter` is the external interface for loading and executing a model. It is high-level class. It depends on IR and all other components.
* `core` component contains classes which are the "runtime" counterpart of the IR model. Consider it as a program loaded into memory. It is lowest level class, it should not have external dependencies.
* `Loader` component is used to translate the IR model into the runtime model, as its name suggest. It does therefore depend on both IR and `core` (but not on `Interpreter`). It is middle-level class.

---

About `EventNotifier` class.
We need to notify registered observers after executing a kernel. `Kernel` is part of the core component (which should not have external dependencies), but `ExecutionObserver::postTensorWrite` depends on IR (it has CircleNode as a parameter). This means that we cannot call it directly. So, how to do this?
This is done in two steps.
1. A kernel uses `EventNotifierImpl::postTensorWrite` to tell the `Interpreter` that the tensor has changed. Note that `EventNotifierImpl` is part of the `core` component (the same component `Kernel` belongs to).
2. `Interpreter` recieves notification via `EventNotifierImpl` class, and looks up `RuntimeToIR` mapping to get the corresponding `CircleNode`. This `CircleNode` is then used to notify external observers (passed into `ExecutionObserver::postTensorWrite`).

Introducing `EventNotifierImpl` class in order to solve the dependency problem is called dependency inversion principle, one of the ground principles of software design. The leter "D" in [SOLID](https://en.wikipedia.org/wiki/SOLID) stands for [DIP -- dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle).

jinevening:
Thank you for the detailed explanation. So, the point is that you want to make the `core` (composed of lowest level classes) separate from the IR. That was the reason for making RuntimeModule, RuntimeGraph, and EventNotifier, etc.

> I am planning to prepare design diagram this week (probably will be ready on next monday). It should lit some light on the design decisions.

I'm looking forward to seeing the design document :)

compiler/luci-interpreter/src/core/RuntimeModule.h
s-barannikov: It is a little easier with pointers. In case of `std::vector<RuntimeGraph>`, when a new element is inserted, references to old elements may be invalidated (if insertion caused resizing of the vector).

PR #2435, opened 2020-06-19
[onert] fix Range operation's invalid assert

dayo09:
- In case of delta is minus, assert fails while the operation is valid

### DIFF
--- a/compute/cker/include/cker/operation/Range.h
+++ b/compute/cker/include/cker/operation/Range.h
@@ -53,8 +53,6 @@ inline void Range(const T *start_data, const T *limit_data, const T *delta_data,
   {
     output_data[i] = value;
     value += delta_value;
-
-    assert(value < limit_value);
   }
 }
 


dayo09:
https://github.sec.samsung.net/STAR/nnfw/issues/11494

hyunsik-yoon:
related to https://github.com/Samsung/ONE/pull/2432

ragmani:
Could you add checking that delta is not equal to 0?

ragmani:
How about changing this PR like https://github.com/Samsung/ONE/pull/2432#discussion_r442659878?

dayo09:
@ragmani 
```c++
if (!((start > limit && delta < 0) || (start < limit && delta > 0)))
  {	  {
    throw std::runtime_error("Range: invalid input values");	    throw std::runtime_error("Range: invalid input values");
  }	  }
```
Above the code, there's this line I added at the previous PR.
if delta == 0, the condition is !((false)||(false)) So it throws exception.

ragmani:
What if `start == limit` and `delta == 0`?

ragmani:
@dayo09 

Oh I see. You are right.
Anyway, `assert(value < limit_value);` has removed by #2432
Could you close this PR if you won't change this PR like https://github.com/Samsung/ONE/pull/2432#discussion_r442659878

dayo09:
@ragmani  sure.

PR #2438, opened 2020-06-19
[onert] use std::is_floating_point to skipp divide by zero check.

glistening:
It uses `std::is_floating_point` instead of introducing `divide`.

### DIFF
--- a/compute/cker/include/cker/operation/BinaryArithmeticOps.h
+++ b/compute/cker/include/cker/operation/BinaryArithmeticOps.h
@@ -32,16 +32,6 @@ namespace cker
 
 namespace
 {
-template <typename T> T divide(const T &a, const T &b)
-{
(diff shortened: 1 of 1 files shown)


compute/cker/include/cker/operation/BinaryArithmeticOps.h
```diff
@@ -61,7 +51,14 @@ const std::function<T(const T &, const T &)> GetBinaryArtithmeticFn(BinaryArithm
     }
     case BinaryArithmeticOpType::DIV:
     {
-      return divide<T>;
+      if (std::is_floating_point<T>::value)
+        return [](const T &a, const T &b) -> T { return a - b; };
```
wateret, line 55:
```suggestion
        return [](const T &a, const T &b) -> T { return a / b; };
```

PR #2439, opened 2020-06-19
[luci] Set quantization type of outputs

jinevening:
This sets quantization type of outputs

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -581,6 +581,18 @@ bool QuantizeWithMinMaxPass::run(loco::Graph *g)
     circle_node->accept(&qb);
   }
 
+  for (auto node : loco::output_nodes(g))
+  {
+    auto circle_node = loco::must_cast<luci::CircleOutput *>(node);
+    if (static_cast<luci::CircleNode *>(circle_node->from())->dtype() == _output_dtype)
+    {
+      circle_node->dtype(_output_dtype);
+      auto graph_outputs = g->outputs();
+      auto graph_output = graph_outputs->at(circle_node->index());
+      graph_output->dtype(_output_dtype);
+    }
+  }
+
   INFO(l) << "QuantizeWithMinMaxPass End" << std::endl;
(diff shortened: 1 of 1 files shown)


seanshpark:
@jinevening , can you please explain purpose of this change? I can guess but I'm not sure.

jinevening:
@seanshpark Currently, the quantization pass only quantizes values in circle nodes. This raises an assert (the output type of loco graph should be the same with the type of CircleOutput) when circle2circle calls `validate`.

This PR sets the quantization type to the output of loco graph (if `from` of CircleOutput is quantized) to avoid the assertion.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -581,6 +581,18 @@ bool QuantizeWithMinMaxPass::run(loco::Graph *g)
     circle_node->accept(&qb);
   }
 
+  for (auto node : loco::output_nodes(g))
```
seanshpark, line 584: Can you please add a commnet what this `for` loop does as you explained in the comment?

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -581,6 +581,18 @@ bool QuantizeWithMinMaxPass::run(loco::Graph *g)
...
+  {
+    auto circle_node = loco::must_cast<luci::CircleOutput *>(node);
+    if (static_cast<luci::CircleNode *>(circle_node->from())->dtype() == _output_dtype)
+    {
+      circle_node->dtype(_output_dtype);
+      auto graph_outputs = g->outputs();
```
seanshpark, line 590: `graph_outputs ` seems to be loop independent

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
seanshpark: Something like this maybe...
1) update output nodes dtype
2) update output dtype

jinevening: I've added a comment.

PR #2444, opened 2020-06-19
[nnfwapi] Remove ConfigSource

wateret:
nnfwapi does not access to ConfigSource as it is not session based,
and we have an API for setting configuration for each session.
So this commit decouples `IConfigSource` from nnfwapi.


wateret:
I'm thinking removing `nnfw_debug_session` after this PR.

PR #2449, opened 2020-06-19
DRAFT Compiler FE: Instance norm pass 

d-krylov:
Draft for Instance norm pass. 


d-krylov:
Very crude draft. More and more review, please!

compiler/luci/pass/src/FuseInstanceNormPassV2.cpp
```diff
@@ -0,0 +1,314 @@
...
+  // TODO What if input is NCHW format in Circle?
+  auto red_indices = dynamic_cast<luci::CircleConst *>(mean->reduction_indices());
+  if (not red_indices)
+    return false;
+  if (red_indices->rank() != 1)
+    return false;
+  std::set<int32_t> red_indices_set;
+  {
+    // TODO Currently only support S32, support other types
+    assert(red_indices->dtype() == loco::DataType::S32);
```
kvochko, line 53: I think it would be better to throw an exception here, assertions are usually disabled in release builds

compiler/luci/pass/include/luci/Pass/FuseInstanceNormCommon.h
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_FUSE_INSTANCE_NORM_COMMON_H__
+#define __LUCI_FUSE_INSTANCE_NORM_COMMON_H__
```
binarman, line 18: Could you clarify, why this is "common"?
For now it is included only once

compiler/luci/pass/src/FuseInstanceNormPassV2.cpp
seanshpark: I think if we do not meet required state of the node for the Pass, we can just return `false` not to run any transformation.

compiler/circle2circle-dredd-recipe-test/test.lst
```diff
@@ -10,7 +10,7 @@
 ## TFLITE RECIPE
 
 Add(Net_InstanceNorm_001 PASS fuse_instnorm)
-# Add(Net_InstanceNorm_002 PASS fuse_instnorm)
+Add(Net_InstanceNorm_003 PASS fuse_instnorm_v2)
```
seanshpark, line 13: Can you describe why you added new network example `Net_InstanceNorm_003 `?

res/TensorFlowLiteRecipes/Net_InstanceNorm_003/test.rule
```diff
@@ -0,0 +1,8 @@
+# To check if custom op InstanceNorm is converted to circle builtin op
+
+RULE    "VERIFY_FILE_FORMAT"      $(verify_file_format) '=' 1
+
+RULE    "INSTANCE_NORM_EXIST"     $(op_count INSTANCE_NORM) '=' 1
+RULE    "NO_RESHAPE"              $(op_count RESHAPE) '=' 3
```
seanshpark, line 6: maybe this is `RESHAPE_EXIST`

compiler/luci/pass/src/FuseInstanceNormPassV2.cpp
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/**
+ * Instance norm pattern would be fused like following diagram:
+ *
+ *    [In] --------------------------- CircleInstanceNorm --- [Out]
+ *                                     / /
+ *    const_as_gamma --- TFLReshape --- /
+ *                                     /
+ *    const_as_beta ---- TFLReshape ---
```
seanshpark, line 242: Maybe use `CircleReshape` instead of `TFLReshape `?

res/TensorFlowLiteRecipes/Net_InstanceNorm_003/test.rule
seanshpark: there is 2 Reshape in the diagram in comment for `fuse_instance_norm` . which is correct?

compiler/luci/pass/include/luci/Pass/FuseInstanceNormCommon.h
```diff
@@ -0,0 +1,126 @@
...
+
+template <class ARG_TYPE_1, class ARG_TYPE_2>
+inline NodeFiller<ARG_TYPE_1, ARG_TYPE_2> fill(ARG_TYPE_1 **arg_1, ARG_TYPE_2 **arg_2)
+{
+  return NodeFiller<ARG_TYPE_1, ARG_TYPE_2>{arg_1, arg_2};
+}
+
+template <class ARG_TYPE_1, class ARG_TYPE_2>
+template <class COMM_NODE>
+bool NodeFiller<ARG_TYPE_1, ARG_TYPE_2>::with_commutative_args_of(const COMM_NODE *node)
```
seanshpark, line 80: I don't see `NodeFiller` used in `FuseInstanceNormPassV2 `, maybe still work in progress?

compiler/luci/pass/src/FuseInstanceNormPassV2.cpp
```diff
@@ -0,0 +1,314 @@
...
+ *
+ * Note
+ *  - 'const_as_gamma' and 'const_as_beta' are from original graph
+ *  - Value of 'const_as_epsilon' would be copied to CircleInstanceNorm's attribute
+ *  - TFLReshape is added as CircleInstanceNorm only accept 1D tensor
+ *  - 'CircleConst --- TFLReshape' is expected to be fused in constant folding for Reshape
+ */
+void fuse_instance_norm(const InstanceNormPattern &p)
```
seanshpark, line 250: Can you provide `fuse_instance_norm` in existing `FuseInstanceNormPass` ? 
I think it would be better not to provide new `FuseInstanceNormPassV2`. This introduces more branchs and codes.

seanshpark:
Thank you! Code looks good!
Just one more thing; you've modified the recipe file, I think it's OK. Can you add a comment on top of the recipe file with the version number of TensorFlow you used to generate the source model(TF .pb, .tflite) ?
```
#
# This was generated by tflchef-reverse from TF version xxxx model
# visit https://github.com/Samsung/ONE/issues/1741 for more information
#
```

binarman:
Everything from this PR is merged, closing

PR #2454, opened 2020-06-22
[circlechef] Fix static analysis warning

llFreetimell:
This commit will fix warning when `flat_array` is `nullptr`.


seanshpark:
How did you encounter `flat_array` to be nullptr? Depends on the caller we can also do like
- we can use `assert` for this
- we may have to return empty vertor

llFreetimell:
@seanshpark 
Actually, I never encountered such cases but analyzer told me like that :(
Then may I change to return empty vector when `flat_array == nullptr` ?

seanshpark:
> Then may I change to return empty vector when flat_array == nullptr ?

I'm not sure because we didn't actually got the problem from real cases.
I think we can go as this change and fix later when something actually happens.

Issue #2455, opened 2020-06-22
Compiler FE: Support AddN Op


seanshpark:
- [x]  res/TensorFlowPythonExamples
- [x]  res/TensorFlowLiteRecipes
- [x]  tflchef
- [x]  tflite2circle
- [x]  tfldump, circledump
- [x]  luci.IR
- [x]  luci.Import
- [x]  luci.Export
- [x]  luci.logex
- [x]  luci.Service
- [x]  luci.tests

For #1606

seanshpark:
```
Operator Codes: [order] OpCodeName (OpCode Enum)
[0] ADD_N (code: 106, version: 1)

Buffers: B(index) (length) values, if any
B(0) (0)
B(1) (0)
B(2) (0)
B(3) (0)
B(4) (0)
B(5) (16) 0x31 0x2e 0x31 0x34 0x2e 0x30 00 00 00 00 00 00 00 00 00 00 ...

-------------------------------------------------------------------
Sub-Graph: #0 main

Operands: T(subgraph index : tensor index) TYPE (shape) (shape_signature) B(buffer index) OperandName
T(0:0) FLOAT32 (1, 3, 3, 5) B(1) input_01
T(0:1) FLOAT32 (1, 3, 3, 5) B(2) input_02
T(0:2) FLOAT32 (1, 3, 3, 5) B(3) input_03
T(0:3) FLOAT32 (1, 3, 3, 5) B(4) add

Operators: O(subgraph index : operator index) OpCodeName
    Option(values) ... <-- depending on OpCode
    I T(tensor index) OperandName <-- as input
    O T(tensor index) OperandName <-- as output
O(0:0) ADD_N
    I T(0:0) input_01
    I T(0:1) input_02
    I T(0:2) input_03
    O T(0:3) add

Inputs/Outputs: I(input)/O(output) T(tensor index) OperandName
I T(0:0) input_01
I T(0:1) input_02
I T(0:2) input_03
O T(0:3) add

```

seanshpark:
done

PR #2458, opened 2020-06-22
DRAFT: Compiler FE support AddN Op

seanshpark:
on-going draft code to support AddN Op


seanshpark:
For #2455

seanshpark:
all landed

PR #2462, opened 2020-06-22
[luci] Fix static analysis warning

llFreetimell:
This commit will fix warning when `node` is `nullptr`.


compiler/luci/export/src/CircleTensorExporter.cpp
```diff
@@ -187,6 +187,9 @@ class MultiOutputDetector final : public luci::CircleNodeMutableVisitor<bool>
 
 void allocateCircleTensor(CircleNode *node, CircleTensorContext &ctx)
 {
+  if (node == nullptr)
+    throw std::runtime_error("Allocation is impossible because node is nullptr");
```
seanshpark, line 191: (optional) `allocateCircleTensor failed: node is nullptr`

PR #2467, opened 2020-06-22
[onert] Add shape inference for reshape with new_shape option

periannath:
- Shape inference of reshape uses new_shape option
  - If new shape is given by second input tensor, use that tensor as new shape
  - If above condition is not met, use new_shape option as new shape

---

For issue : #2393

### DIFF
--- a/runtime/onert/core/src/util/shapeinf/Reshape.cc
+++ b/runtime/onert/core/src/util/shapeinf/Reshape.cc
@@ -75,25 +75,41 @@ void StaticInferer::visit(const ir::operation::Reshape &op)
     return;
   }
 
-  if (op.getInputs().size() == 1)
+  // New shape is given by second input tensor
+  if (op.getInputs().size() == 2)
   {
-    // no change on output shape
-    return;
+    // Let's check the second input
+    const auto shape_idx{op.getInputs().at(ir::operation::Reshape::Input::SHAPE)};
+    const auto &shape = _operands.at(shape_idx);
+
+    if (shape.isConstant())
+    {
+      const auto *shape_buf = reinterpret_cast<const int32_t *>(shape.data()->base());
+      assert(shape_buf);
+
+      ir::Shape new_shape =
+          convertShape(shape_buf, shape.shape().num_elements(), input.shape().num_elements());
+
+      // if shape is from Const, TFLC put the shape of output into tensor
+      if (new_shape != output.shape())
+      {
+        // change on output shape
+        output.info().shape(new_shape);
+      }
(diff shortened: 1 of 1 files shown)


glistening:
@periannath Do you have test that you'd like to solve from #2393 and this PR?

periannath:
@glistening I've created TFLite file from `res/TensorFlowLiteRecipes/Reshape_000/test.recipe`. It has one input tensor and get `new_shape` as option.

runtime/onert/core/src/util/shapeinf/Reshape.cc
```diff
@@ -75,25 +75,41 @@ void StaticInferer::visit(const ir::operation::Reshape &op)
...
+    const auto shape_idx{op.getInputs().at(ir::operation::Reshape::Input::SHAPE)};
+    const auto &shape = _operands.at(shape_idx);
 
-  // Let's check the second input
-  const auto shape_idx{op.getInputs().at(ir::operation::Reshape::Input::SHAPE)};
-  const auto &shape = _operands.at(shape_idx);
-
-  ir::Shape new_shape;
-  if (shape.isConstant())
+    if (shape.isConstant())
```
ragmani, line 85: Does `shape` operand have a higher priority than `op.param().new_shape`?

runtime/onert/core/src/util/shapeinf/Reshape.cc
```diff
@@ -75,25 +75,41 @@ void StaticInferer::visit(const ir::operation::Reshape &op)
     return;
   }
 
-  if (op.getInputs().size() == 1)
+  // New shape is given by second input tensor
+  if (op.getInputs().size() == 2)
```
ragmani, line 79: Does shape operand have a higher priority than op.param().new_shape?

glistening: @ragmani Yes. According to tflite's `reshape` implementation, `input` has higher priority than `param` though it is different to https://www.tensorflow.org/lite/guide/ops_compatibility. As I understand, `tflite` seems to use `param` when the shape is determined before runtime, it uses `input` for dynamic shape. If `reshape` has `input`, `input` should be handled with high priority.

glistening:
@periannath Do you have the test in your pc? If you're not going to run in CI for some reason, could you please attach the test in comment?

periannath:
@glistening Here is the generated tflite file. [reshape_000.zip](https://github.com/Samsung/ONE/files/4816197/reshape_000.zip)

Issue #2460, opened 2020-06-22
[onert] Mixed backend test fail


hseok-oh:
Mixed backend test (debug build) fail after #2398

```
$ ./infra/scripts/test_ubuntu_runtime_mixed.sh

============================================
Verification with nnapi_test ...

1..36
not ok 1 - MODELS/inception_module
not ok 2 - MODELS/mobilenet
ok 3 - add/1D
ok 4 - add/4D
ok 5 - average_pool_2d/aligned
ok 6 - average_pool_2d/avgpool1
ok 7 - average_pool_2d/avgpool2
ok 8 - cast
ok 9 - concat/2D
ok 10 - concat/concat1
ok 11 - concat/concat2
ok 12 - conv_2d/convolution1
ok 13 - conv_2d/convolution2
ok 14 - depthwise_conv_2d/depthconv1
ok 15 - depthwise_conv_2d/depthconv2
ok 16 - div/broadcast
not ok 17 - fullyconnected/fc1
ok 18 - fullyconnected/hybrid
ok 19 - fullyconnected/matmul2x2
ok 20 - fullyconnected/weights_as_input
ok 21 - max
ok 22 - max_pool_2d/maxpool1
ok 23 - max_pool_2d/maxpool2
ok 24 - mean
ok 25 - min
ok 26 - mul/broadcast
ok 27 - pack
ok 28 - reduce_max
ok 29 - rsqrt
ok 30 - slice
ok 31 - softmax
ok 32 - squeeze
ok 33 - strided_slice
ok 34 - sub/broadcast
ok 35 - tanh
ok 36 - transpose

Verification failed... exit code: 1
============================================
```

hseok-oh:
cc/ @ragmani 

```
$ ACL_LAYOUT="NCHW" OP_BACKEND_Conv2D="cpu" \
OP_BACKEND_MaxPool2D="acl_cl" OP_BACKEND_AvgPool2D="acl_neon" \
BACKENDS="acl_cl;acl_neon;cpu" \
./Product/out/bin/nnapi_test \
  tests/scripts/framework/cache/fullyconnected/fc1/fullyconnected_test.tflite 
[NNAPI TEST] Run T/F Lite Interpreter without NNAPI
[NNAPI TEST] Run T/F Lite Interpreter with NNAPI

[...]

PR #2483, opened 2020-06-22
[onert] Allow permutation for rank < 4

hseok-oh:
Change assert condition on layout check on permutation to rank <= 4


hseok-oh:
For #2460

PR #2485, opened 2020-06-22
[circle-tensordump] Fix bug on dumping uint8

jinevening:
This fixes a bug on dumping uint8 values

### DIFF
--- a/compiler/circle-tensordump/src/Dump.cpp
+++ b/compiler/circle-tensordump/src/Dump.cpp
@@ -95,7 +95,7 @@ void print_buffer(std::ostream &os, uint32_t buff_idx, const flatbuffers::Vector
   os << std::endl;
 }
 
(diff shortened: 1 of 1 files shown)


compiler/circle-tensordump/src/Dump.cpp
```diff
@@ -95,7 +95,7 @@ void print_buffer(std::ostream &os, uint32_t buff_idx, const flatbuffers::Vector
   os << std::endl;
 }
 
-} // namepsace
+} // namespace
```
seanshpark, line 98: This fix is not a bug-fix but a typo-fix but let's go as-is

jinevening: Thank you :) I did not want to include this, but this was automatically included by the format checker.

PR #1964, opened 2020-06-05
[onert] Change cpu backend impl to use ITensor

wateret:
In order to reuse tensor from another backend, `***Layer`s
must accept rather cpu's specific tensor(`cpu::Tensor`).
This commit makes cpu backend use `ITensor` in Layer implementations(To
accept tensors from other backends).

Part of #1325

ONE-DCO-1.0-Signed-off-by: Hanjoung Lee <hanjoung.lee@samsung.com>

PR #2492, opened 2020-06-22
[onert] Introduce IPortableTensor

wateret:
This commit introduces `IPortableTensor` which can be used for
cpu-based no-padding tensors.

Backends that use `IPortableTensor` derivatives can reuse each others'
tensors without copying. This can be done by not adding Permute op in
`PermutationInsertionPass` (Will be implemented).

Part of #1325 (an alternative of #1964)

### DIFF
--- a/runtime/onert/backend/cpu/TensorBuilder.cc
+++ b/runtime/onert/backend/cpu/TensorBuilder.cc
@@ -94,20 +94,19 @@ void TensorBuilder::allocate()
 
 std::shared_ptr<ITensor> TensorBuilder::tensorAt(const ir::OperandIndex &ind)
 {
-  auto found = _tensor_reg->find(ind);
-  if (found == _tensor_reg->end())
-    return nullptr;
+  return _tensor_reg->getITensor(ind);
+}
 
-  return found->second;
+std::shared_ptr<IPortableTensor> TensorBuilder::portableAt(const ir::OperandIndex &ind)
+{
+  return _tensor_reg->getPortableTensor(ind);
 }
 
 void TensorBuilder::iterate(const IterateFunction &fn) { _static_tensor_mgr->iterate(fn); }
 
 std::shared_ptr<cpu_common::Tensor> TensorBuilder::at(const ir::OperandIndex &ind)
 {
-  auto found = _tensor_reg->find(ind);
-  assert(found != _tensor_reg->end());
-  return found->second;
+  return _tensor_reg->getManagedTensor(ind);
 }
 
 std::unique_ptr<ITensorManager> TensorBuilder::releaseStaticTensorManager(void)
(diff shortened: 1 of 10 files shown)


wateret:
Class Inheritance diagram:

image

`IPortableTensor` derivatives can be reused(registered to `ITensorRegistry` as external tensor) in other backends that also uses `IPortableTensor`.

runtime/onert/backend/cpu/TensorBuilder.cc
```diff
@@ -94,20 +94,19 @@ void TensorBuilder::allocate()
...
 std::shared_ptr<ITensor> TensorBuilder::tensorAt(const ir::OperandIndex &ind)
 {
-  auto found = _tensor_reg->find(ind);
-  if (found == _tensor_reg->end())
-    return nullptr;
+  return _tensor_reg->getITensor(ind);
+}
 
-  return found->second;
+std::shared_ptr<IPortableTensor> TensorBuilder::portableAt(const ir::OperandIndex &ind)
```
ragmani, line 100: When is this method used?

wateret: Oops, I missed explanation. This will be the next PR - https://github.com/wateret/ONE/commit/a1dbab5bec7e035b7a556cc40dbf3cc24c16ccf8 .

All calls `_tensor_builder->at()` will be replaced to `_tensor_builder->portableAt()` in CPU kernel generator and all `ops` classes in cpu backend is going to have `IPortableTensor` instead of `cpu_common::Tensor`.

ragmani: Could you add an assertion that checks portable tensor's layout in this method like `assert(_tensor_reg->getPortableTensor(ind)->layout() == ir::Layout::NHWC);`?

wateret: Good point. Well, I have been thinking we can handle Layouts since just being portable tensors does not mean that all those tensors are reusable - we also have layout info in `PermuteFactor`. So it will skip inserting Permute operation only if layouts match.

---

For more details,

I'm going to remove Source/Sink when starting execution. Model Inputs are going to be treated as tensors in `controlflow`. That means Permute op will do the job of Source/Sink. This makes it `PermutationInsertionPass` to do everything so we can optimize it further.

For now `PermutationInsertionPass` always makes Permute operation since it cannot reuse tensors from other backends. But with this work, it can reuse `IPortableTensor` tensors so can skip inserting `Permute`s, **only if these conditions are met**.

- The tensor that we are processing is `IPortableTensor` the backend that uses that tensor can use portable tensors.
- Layouts must match (We have this in `PermuteFactor`)

These conditions will be implemented in `PermutationInsertionPass`.

Plus, this will also make it possible that, for example, if we have another CPU backend(let's say its id `cpu2` with different kernels), and some operations use cpu2 backend, then it will not insert any `Pemute` operations.

wateret: With a discussion with @ragmani , I will add an assertion to CPU backend!

PR #2497, opened 2020-06-22
[res] Add example for ReverseSequence

d-krylov:
Add example for ReverseSequence


res/TensorFlowPythonExamples/examples/reverse_sequence/__init__.py
```diff
@@ -0,0 +1,4 @@
+import tensorflow as tf
+
+in_ = tf.compat.v1.placeholder(dtype=tf.float32, shape=(4, 8), name="Hole")
+op_ = tf.compat.v1.reverse_sequence(in_, [7, 2, 3, 5])
```
binarman, line 4: Could you check this example, I think `seq_dim` or `seq_axis` arguments are missing.
maybe also `batch_dim`/`batch_axis`, need check this too

PR #2500, opened 2020-06-23
[onert] use template parameter for BinaryArithmetic op types

glistening:
Previously, it uses runtime parameter for op type (ADD, SUB, ...).
However, it is determined before runtime. It replaces it with
template parameter. It gives about 1~2% improvement for some model.

### DIFF
--- a/runtime/onert/backend/cpu/ops/AddLayer.cc
+++ b/runtime/onert/backend/cpu/ops/AddLayer.cc
@@ -32,7 +32,6 @@ void AddLayer::addFloat32()
   float output_activation_min = 0, output_activation_max = 0;
   CalculateActivationRange(_activation, &output_activation_min, &output_activation_max);
   nnfw::cker::BinaryArithmeticOpParam op_params;
-  op_params.type = nnfw::cker::BinaryArithmeticOpType::ADD;
   op_params.float_activation_max = output_activation_max;
   op_params.float_activation_min = output_activation_min;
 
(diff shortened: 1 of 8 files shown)


runtime/onert/backend/cpu/ops/AddLayer.cc
```diff
@@ -32,22 +32,21 @@ void AddLayer::addFloat32()
...
-  op_params.type = nnfw::cker::BinaryArithmeticOpType::ADD;
   op_params.float_activation_max = output_activation_max;
   op_params.float_activation_min = output_activation_min;
 
   const bool need_broadcast =
       nnfw::cker::ProcessBroadcastShapes(getTensorShape(_lhs), getTensorShape(_rhs), &op_params);
   if (need_broadcast)
   {
-    nnfw::cker::BroadcastBinaryArithmeticOp<float>(
+    nnfw::cker::BroadcastBinaryArithmeticOp<nnfw::cker::BinaryArithmeticOpType::ADD>(
```
periannath, line 42: I think it is better to specifiy variable type in here. 
```suggestion
    nnfw::cker::BroadcastBinaryArithmeticOp<nnfw::cker::BinaryArithmeticOpType::ADD, float>(
```

glistening:
@periannath Specifying `uint8_t` gives error:

```
/home/brian/z/ONE/compute/cker/include/cker/operation/reference/BinaryArithmeticOps.h: In instantiation of ‘void nnfw::cker::reference::BinaryArithmeticOp(const nnfw::cker::BinaryArithmeticOpParam&, const nnfw::cker::Shape&, const T*, const nnfw::cker::Shape&, const T*, const nnfw::cker::Shape&, T*, const std::function<T(const T&, const T&)>&) [with T = unsigned char]’:
/home/brian/z/ONE/compute/cker/include/cker/operation/BinaryArithmeticOps.h:197:32:   required from ‘void nnfw::cker::BinaryArithmeticOp(const nnfw::cker::BinaryArithmeticOpParam&, const nnfw::cker::Shape&, const T*, const nnfw::cker::Shape&, const T*, const nnfw::cker::Shape&, T*) [with nnfw::cker::BinaryArithmeticOpType op_type = nnfw::cker::BinaryArithmeticOpType::ADD; T = unsigned char]’
/home/brian/z/ONE/runtime/onert/backend/cpu/ops/AddLayer.cc:127:78:   required from here
/home/brian/z/ONE/compute/cker/include/cker/operation/reference/BinaryArithmeticOps.h:43:50: error: no matching function for call to ‘ActivationFunctionWithMinMax(unsigned char, const int32_t&, const int32_t&)’
   43 |     output_data[i] = ActivationFunctionWithMinMax(fn(input1_data[i], input2_data[i]),
      |                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   44 |                                                   params.quantized_activation_min,
      |                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   45 |                                                   params.quantized_activation_max);
      |                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/brian/z/ONE/compute/cker/include/cker/operation/reference/BinaryArithmeticOps.h:23,
```

periannath:
@glistening I think both are fine. Do as you prefer :)

glistening:
@periannath I've tried to fix build error the above after specifying `uint8_t`. However, specifying both `op_type` and `uint8_t` leads to using two template parameter version function. It gives wrong answer since it does not handle `uint8_t` case. To use overloaded template function (same name but has one template parameter `op_type`), we need to pass 1 template parameter only like this PR.

PR #2503, opened 2020-06-23
MinLayer supports INT32 type on cpu backend

chunseoklee:
- MinLayer supports INT32 type on cpu backend


runtime/onert/backend/cpu/ops/MinLayer.cc
```diff
@@ -36,6 +36,13 @@ void MinLayer::minFloat32()
                          getTensorShape(_output), reinterpret_cast<float *>(_output->buffer()));
 }
 
+void MinLayer::minInt32()
```
hseok-oh, line 39: We can use template to reduce code duplication.

PR #2504, opened 2020-06-23
[tools] Fix arguments of numpy function

llFreetimell:
`Numpy` function `random_sample` has only one argument.
This commit will apply this fix.


llFreetimell:
This PR is wrong, I will create another PR.

Issue #2505, opened 2020-06-23
Compiler FE: Support SparseToDense Op


struss:
Parent Issue : #1606 
Draft PR : #2507 

- [x] res/TensorFlowPythonExamples #2521
- [x] tflchef #2524
- [x] tflite2circle #2529
- [x] tfldump #2526
- [x] circledump #2527
- [x] res/TensorFlowLiteRecipes #2600 
- [x] luci.IR #2605 
- [x] luci.Import #2641 
- [x] luci.Export #2640 
- [x] luci.logex #2642 
- [x] luci.Service #2643 
- [x] luci.tests #2686

struss:
Done! Thanks for all!

PR #2507, opened 2020-06-23
[Draft]Enable SPARSE_TO_DENSE Draft

struss:
This commit is draft for `SPARSE_TO_DENSE` Operation.

Parent Issue : #2505


compiler/luci/lang/src/Nodes/CircleSparseToDense.test.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+#include "luci/IR/Nodes/CircleSparseToDense.h"
+
+#include "luci/IR/CircleDialect.h"
+
+#include <gtest/gtest.h>
+
+TEST(CircleSparseToDenseTest, constructor_P)
```
seanshpark, line 23: If you look in to other tests, there are also `---_NEG` tests. Can you please add them also?
And we don't have to put `_P` suffix for normal test cases :)

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -1676,6 +1676,39 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
...
+        LUCI_ASSERT(output_shape_node->dtype() == loco::DataType::S32,
+                    "Only support int32 CircleConst");
+
+        if (output_shape_node->rank() != 1)
+          INTERNAL_EXN_V("Only support rank 1 CircleConst",
+                         oops::to_uint32(output_shape_node->rank()));
+
+        shape.rank(output_shape_node->dim(0).value());
+
+        for (uint32_t axis = 0; axis < shape.rank(); ++axis)
```
seanshpark, line 1698: Is `output_shape_node->rank()` always 1 or current implementation only supports for 1 and later we may support for rank() > 1 ?
If this is for former case, we won't need to run a for loop. If second case, may be add `TODO support rank() > 1` so that loop will be used in the future...

compiler/luci/lang/src/Nodes/CircleSparseToDense.test.cpp
struss: Okay I will add it.

seanshpark:
others looks good !

compiler/luci/service/src/CircleShapeInferenceRule.cpp
struss: In implementation of TF V1.14, output_shape_node always rank 1.

seanshpark: Ah... I didn't read the code correctly: `shape.rank(output_shape_node->dim(0).value());`. My mistake.
Code looks OK :)

struss:
Thanks! I will make sub PR for this Operation.

struss:
Done! Thanks for all that help me!

Issue #2509, opened 2020-06-23
Compiler FE: Support LogSoftmax Op


seanshpark:
- [x]  tflchef
- [x]  tflite2circle
- [x]  tfldump, circledump
- [x]  res/TensorFlowPythonExamples
- [x]  res/TensorFlowLiteRecipes
- [x]  luci.IR
- [x]  luci.Import
- [x]  luci.Export
- [x]  luci.logex
- [x]  luci.Service
- [x]  luci.tests

For #1606

seanshpark:
https://github.com/tensorflow/docs/blob/r1.13/site/en/api_docs/python/tf/nn/log_softmax.md

seanshpark:
```
input = tf.compat.v1.placeholder(tf.float32, shape=[2,3,4,5])
output = tf.compat.v1.nn.log_softmax(input)
```
-->
```
Operator Codes: [order] OpCodeName (OpCode Enum)
[0] LOG_SOFTMAX (code: 50, version: 1)

Buffers: B(index) (length) values, if any
B(0) (0)
B(1) (0)
B(2) (0)

-------------------------------------------------------------------
Sub-Graph: #0 (noname)

Operands: T(subgraph index : tensor index) TYPE (shape) (shape_signature) B(buffer index) OperandName
T(0:0) FLOAT32 (2, 3, 4, 5) B(1) LogSoftmax
T(0:1) FLOAT32 (2, 3, 4, 5) B(2) Placeholder
    Quantization: min(0) max(255)

Operators: O(subgraph index : operator index) OpCodeName
    Option(values) ... <-- depending on OpCode
    I T(tensor index) OperandName <-- as input
    O T(tensor index) OperandName <-- as output
O(0:0) LOG_SOFTMAX
    I T(0:1) Placeholder
    O T(0:0) LogSoftmax

Inputs/Outputs: I(input)/O(output) T(tensor index) OperandName
I T(0:1) Placeholder
O T(0:0) LogSoftmax
```

seanshpark:
```
input = tf.compat.v1.placeholder(tf.float32, shape=[2,3,4,5])
output = tf.compat.v1.nn.log_softmax(input, axis=1)
```
--> TF
![image](https://user-images.githubusercontent.com/4616940/85358424-ae9fb480-b54e-11ea-9ded-eb64b6f52590.png)

--> TF lite
[...]

PR #2510, opened 2020-06-23
DRAFT Compiler FE support log_softmax

seanshpark:
on-going draft to support log_softmax operator


seanshpark:
For #2509

seanshpark:
all landed

PR #2515, opened 2020-06-23
[circledump] Comment for LogSoftmax

seanshpark:
This will add comment for LogSoftmax as there is no options to print


seanshpark:
For #2509, from draft #2510

PR #2518, opened 2020-06-23
[tflite2circle] Enable LogSoftmax Op

seanshpark:
This will enable tflite to circle conversion of LogSoftmax Op


seanshpark:
For #2509, from draft #2510

seanshpark:

PR #2530, opened 2020-06-23
[DRAFT][onert] Remove unnecessary copies during loop of while

ragmani:
This draft removes unnecessary copies during loop of while.


ragmani:
Done

PR #2535, opened 2020-06-23
[luci] Shape,dtype inf for LogSoftmax IR

seanshpark:
This will enable shape and dtype inference for LogSoftmax IR


seanshpark:
For #2509, from draft #2510

PR #2536, opened 2020-06-23
[luci] Export LogSoftmax IR

seanshpark:
This will enable export of LogSoftmax IR


seanshpark:
For #2509, from draft #2510

PR #2540, opened 2020-06-23
[onert] Remove configure decl of If/WhileLayer

wateret:
Remove `configure` method declaration of If/WhileLayer of controlflow
backend which does not have implementation.


ragmani:

PR #2542, opened 2020-06-23
[luci-interpreter] Add If kernel

s-barannikov:
Add implementation of `If` kernel.


compiler/luci-interpreter/src/core/RuntimeGraph.h
```diff
@@ -33,7 +33,7 @@ class RuntimeGraph
 public:
   explicit RuntimeGraph(RuntimeModule *owning_module) : _owning_module(owning_module) {}
 
-  void addTensor(std::unique_ptr<Tensor> &&tensor);
+  Tensor *addTensor(std::unique_ptr<Tensor> &&tensor);
```
binarman, line 36: Maybe add new method
`Tensor *addTensor(loco::DataType dtype, Shape shape, AffineQuantization quant, std::string)` ?

compiler/luci-interpreter/src/kernels/If.cpp
```diff
@@ -0,0 +1,104 @@
...
+    const auto &input_tensors = graph->getInputTensors();
+    for (size_t i = 0; i < _inputs.size(); ++i)
+    {
+      assert(input_tensors[i]->element_type() == _inputs[i]->element_type());
+      input_tensors[i]->resize(_inputs[i]->shape());
+    }
+    graph->configure();
+  }
+
+  // For now, output tensors of both graphs should have the same shape.
```
jinevening, line 54: Maybe they should have the same type as well?

compiler/luci-interpreter/src/core/RuntimeGraph.cpp
```diff
@@ -68,17 +69,23 @@ void RuntimeGraph::execute() const
   EventNotifier *event_notifier = _owning_module->getEventNotifier();
 
   // Notify the observers that the input tensors have changed.
-  for (Tensor *input_tensor : getInputTensors())
+  if (event_notifier != nullptr)
```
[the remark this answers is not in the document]
s-barannikov: I did not get it. What do you mean?

compiler/luci-interpreter/src/kernels/If.cpp
s-barannikov: I added an assertion.

compiler/luci-interpreter/src/core/RuntimeGraph.cpp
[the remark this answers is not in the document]
binarman: I mean, this pointer must be `!= nullptr` to consider RuntimeGraph constructed correctly.
Nevermind, I am just a little surprised.

PR #2545, opened 2020-06-23
[luci] Support Where operator

d-krylov:
Support Where operator in luci
    

compiler/luci/import/src/GraphBuilderRegistry.cpp
```diff
@@ -153,7 +154,9 @@ GraphBuilderRegistry::GraphBuilderRegistry()
   // BuiltinOperator_SQUARE = 92,
   // BuiltinOperator_UNIQUE = 103,
   // BuiltinOperator_REVERSE_V2 = 105,
-  // BuiltinOperator_WHERE = 109,
+  // BuiltinOperator_ADD_N = 106,
+  // BuiltinOperator_RANK = 110,
+  // BuiltinOperator_REVERSE_SEQUENCE = 112,
```
seanshpark, line 159: Please check this again. These doesn't seem to be here.

seanshpark:
Please add for `logex`

compiler/luci/lang/include/luci/IR/Nodes/CircleWhere.h
```diff
@@ -0,0 +1,59 @@
...
+class CircleWhere final : public VariadicArityNode<CircleNodeImpl<CircleOpcode::WHERE>>
+{
+public:
+  CircleWhere(uint32_t arity) : VariadicArityNode<CircleNodeImpl<CircleOpcode::WHERE>>(arity)
+  {
+    assert(arity >= 1);
+  }
+
+public:
+  uint32_t numValues(void) const { return arity(); }
```
binarman, line 42: I think we can add named Accessors here.
At least one - `cond`

compiler/luci/lang/include/luci/IR/Nodes/CircleWhere.h
```diff
@@ -0,0 +1,65 @@
...
+public:
+  loco::Node *cond() const { return at(0)->node(); }
+  void cond(loco::Node *node) { at(0)->node(node); }
+
+  loco::Node *x() const
+  {
+    if (arity() == 3)
+      return at(1)->node();
+    return nullptr;
+  }
```
binarman, line 51: @seanshpark 
What do you think about this design?
According TF documentation where operator can be executed in two modes: with 1 input (condition) and 3 inputs (condition, x, y). I tried to express it explicitly.
Old design was generic: any number of inputs can be set in where operator.

If you think old design is better, we can revert changes.

seanshpark: I think it's OK, I'd like to ask @parjong also :)

compiler/luci/lang/include/luci/IR/Nodes/CircleWhere.h
```diff
@@ -0,0 +1,65 @@
...
+  void x(Node *node) { at(1)->node(node); }
```
seanshpark, line 52: for i input this will fail. we need to check :)

compiler/luci/lang/include/luci/IR/Nodes/CircleWhere.h
seanshpark: if this returns `nullptr` caller needs to check if it's null or not. this `if` conditional branch can happen with `arity() == ` condition. so condition check has not much difference.
so, this may be about how methods are user friendly. current is +1 :)

parjong: > According TF documentation where operator can be executed in two modes

@binarman Could you please check TFLITE documentation or implementation? TF and TFLITE are similar in most cases, but we need to check this again.

compiler/luci/lang/include/luci/IR/Nodes/CircleWhere.h
binarman: Do you mean we should do nothing in this case?

For now it will throws an exception, I added tests (`constructor_without_x_NEG` and `constructor_without_y_NEG`) for this.

seanshpark: Ok. That sounds nice.

compiler/luci/lang/include/luci/IR/Nodes/CircleWhere.h
binarman: I checked what is generated by converter for TF Where and found that it generates TFLite Where for 1 input mode and TFLite Select for 3 input mode.
I also checked TFLite implementation and found that it is not able to run 3-input Where operator.

Sooo, I think I need to remove 3 input version of Where operator from this PR =)

compiler/luci/lang/include/luci/IR/Nodes/CircleWhere.h
binarman: see https://github.com/Samsung/ONE/pull/2545#discussion_r447363169

compiler/luci/lang/include/luci/IR/Nodes/CircleWhere.h
seanshpark: @binarman , FYI, @parjong is on vacation today...

Issue #2550, opened 2020-06-24
Compiler FE: install check


seanshpark:
Let's check installation

command
```
BUILD_TYPE=Release ./nncc build install
```

Whitelist to build
- https://github.com/Samsung/ONE/blob/master/infra/nncc/command/utcount#L12

Todo
- find missing components
- revise cmake for missing components

seanshpark:
Makefile for build and install (WIP)
```
BUILD_TYPE?=Release
TYPE_FOLDER=$(shell echo $(BUILD_TYPE) | tr A-Z a-z)

CURRENT_DIR=$(shell pwd)
BUILDFOLDER=build

BUILD_TYPE_FOLDER=$(BUILDFOLDER)/$(TYPE_FOLDER)
LOCAL_INSTALL_FOLDER=$(CURRENT_DIR)/$(BUILDFOLDER)/$(TYPE_FOLDER).install

NNCC_CFG_OPTION+= -DCMAKE_BUILD_TYPE=$(BUILD_TYPE)
NNCC_CFG_OPTION+= -DCMAKE_INSTALL_PREFIX="$(LOCAL_INSTALL_FOLDER)"

REL_BUILD_ITEMS:=angkor;cwrap;pepper-str;pepper-strcast;pp;stdex
REL_BUILD_ITEMS+=;oops;pepper-assert
REL_BUILD_ITEMS+=;hermes;hermes-std
REL_BUILD_ITEMS+=;loco;locop;locomotiv;logo-core;logo
REL_BUILD_ITEMS+=;safemain;mio-circle;mio-tflite
REL_BUILD_ITEMS+=;foder;circle-verify
REL_BUILD_ITEMS+=;tflite2circle
REL_BUILD_ITEMS+=;luci
REL_BUILD_ITEMS+=;luci-interpreter
REL_BUILD_ITEMS+=;luci-value-test
REL_BUILD_ITEMS+=;circle2circle;record-minmax
REL_BUILD_ITEMS+=;tflchef;circlechef
REL_BUILD_ITEMS+=;tf2tfliteV2
REL_BUILD_ITEMS+=;record-minmax-conversion-test
REL_BUILD_ITEMS+=;tf2tfliteV2-conversion-test
REL_BUILD_ITEMS+=;tflite2circle-conversion-test

_SPACE_:=
_SPACE_+=
REL_BUILD_WHITELIST=$(subst $(_SPACE_),,$(REL_BUILD_ITEMS))

[...]

PR #2554, opened 2020-06-24
[record-minmax] install target

seanshpark:
This will revise cmake to install target to bin folder


seanshpark:
For #2550

PR #2558, opened 2020-06-24
[nnpkgrun] coverity: Fix except thrown in noexcept

wateret:
Fix exceptions thrown in noexcept function. This commit makes it not
`noexcept` and put in a try block so the exceptions are handled.


wateret:
Dup #2562

PR #2561, opened 2020-06-24
[cker] Remove deadcode

wateret:
Remove deadcode as `label` never can be negative.


wateret:
If this code is imported, then I'll close this.

PR #2563, opened 2020-06-24
[onert] catch uncaught exception in event observers

glistening:
Svace complains the destructor could throw exception.
This patch will catch the possible exceptions.

### DIFF
--- a/runtime/onert/core/src/exec/ExecutionObservers.cc
+++ b/runtime/onert/core/src/exec/ExecutionObservers.cc
@@ -74,7 +74,17 @@ ChromeTracingObserver::ChromeTracingObserver(const std::string &filepath, const 
 {
 }
 
-ChromeTracingObserver::~ChromeTracingObserver() { _recorder.writeToFile(_ofs); }
+ChromeTracingObserver::~ChromeTracingObserver()
+{
+  try
+  {
+    _recorder.writeToFile(_ofs);
+  }
+  catch (const std::exception &e)
+  {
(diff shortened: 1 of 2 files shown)


ragmani:
Could you inform what kind of exception it is?

glistening:
@ragmani Coverity says `Json::LogicError` could be thrown in `writeSNPEBenchmark(os)`. If ID `1140307` is resolved, it could be disappear. We may wait for `1140307`.

runtime/onert/core/src/exec/ExecutionObservers.cc
```diff
@@ -74,7 +74,17 @@ ChromeTracingObserver::ChromeTracingObserver(const std::string &filepath, const
...
 }
 
-ChromeTracingObserver::~ChromeTracingObserver() { _recorder.writeToFile(_ofs); }
+ChromeTracingObserver::~ChromeTracingObserver()
+{
+  try
+  {
+    _recorder.writeToFile(_ofs);
+  }
+  catch (const std::runtime_error &e)
```
ragmani, line 83: `Json::LogicError` is inherited from `std::exception`, not `std::runtime_error`. Please change it to `std::exception`
Refer to declaration as below
`JSONCPP_NORETURN void throwLogicError(JSONCPP_STRING const &msg) { throw LogicError(msg); }`
`class JSON_API LogicError : public Exception`
`class JSON_API Exception : public std::exception`

Issue #2522, opened 2020-06-23
Resolve coverity defect on master


chunseoklee:
~~link : https://code.sec.samsung.net/ahub/service/analyses/128858~~
internal coverity analysis report : https://code.sec.samsung.net/ahub/service/analyses/130150

## Summary
update : 2020/06/24 12:31
21 defects 
| WID | resolved? | assignee | internal defect link |
| ----|    -----| ----| -- |
|    1140718     |      #2557             | @wateret | |
|    1140582     |         #2558          | @wateret | |
| 1140580 | Done - Intentional | @wateret | |
| 1140579 |#2559 | @wateret | |
| 1140578 | #2558 |@wateret | |
| 1140462 | #2561 |@wateret | |
| 1140458 | external (from TF) | @glistening | |
| 1140457 | Done | @glistening | |
| 1140307 | Done - False positive |@wateret | |
| 1140306 | #2563 | @glistening | |
| 1140305 | #2563 | @glistening | |
| 1139635 | Done | @glistening | |
| 1139634 | Done | @glistening | |
| 1134443 | Done | @glistening | |
| 1091162 | #2562 | @ragmani  | |
| 1091099 | #2562 | @ragmani  | |
| 1085213 | Done - False positive | @ragmani  | |
| 1045362 | #2560 | @ragmani  |  |
| 1045361 | #2560 | @ragmani |   |
| 839456 | #2556 | @ragmani  | |
|111659 | external | | |

hseok-oh:
Many defects are related with test binaries

chunseoklee:
cc @glistening @wateret @ragmani

glistening:
@wateret, @ragmani I will start from 1/3 line (1140458) to bottom.

wateret:
@periannath http://10.113.139.115:8081/reports.htm#v13368/p10494/fileInstanceId=259850887&defectInstanceId=92817200&mergedDefectId=1140580 Could you please take a look? This can be marked as false positive?

wateret:
For 1085213 and 1140307

It had been marked as False Positive.

http://10.113.139.115:8081/reports.htm#v13368/p10180/fileInstanceId=209067200&defectInstanceId=75293460&mergedDefectId=1085213

wateret:
Marked 1085213 and 1140307 as "False Positive".

Marked 1140580 as "Intentional".

chunseoklee:
[...]

PR #2564, opened 2020-06-24
[onert] Fix bad free with using Json::Value in nnfw_api_internal

ragmani:
For issue #2522

This commit fixes bad free with using Json::Value in nnfw_api_internal.
  - Change some Json::Value variables to rvalue references


wateret:
> since the message operator [] frees array "models"

I'm not sure this will fix that error.

PR #2565, opened 2020-06-24
[res] TFLite recipes for Conv2D, DepsthwiseConv2D

jinevening:
This adds TFLite recipes used to test quantized values

---
Related to #2088


jinevening:
These recipes will be used to test quantized models.
(We need models whose weights (and bias) filled with explicit values)

PR #2569, opened 2020-06-24
[onert] converity: Fix EventRecorder

wateret:
`Json::Value::operator[]` could throw execption if `Value` type is
non-null and non-object. So I hope this make coverity ensure that
`root` is an object.

An alternative for #2563


wateret:
Hmm.. #2563 looks better.

PR #2571, opened 2020-06-24
[onert] cpu: Add an assertion in KernelGenerator

wateret:
Add an assertion to check if tensors' layouts are NHWC in `backend::cpu::KernelGenerator`.

This is useful since we are going to reuse tensors from other backends.
So there may be some chance that we have non-NHWC tensors accidentally.

It has been discussed in #2492.

### DIFF
--- a/runtime/onert/backend/cpu/KernelGenerator.cc
+++ b/runtime/onert/backend/cpu/KernelGenerator.cc
@@ -122,6 +122,12 @@ void KernelGenerator::visit(const ir::OpSequence &op_seq)
 
     for (const auto &ind : (node.getInputs() | ir::Remove::UNDEFINED) + node.getOutputs())
     {
+      auto portable_tensor = _tensor_builder->portableAt(ind);
+      if (portable_tensor)
+      {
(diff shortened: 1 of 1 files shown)


runtime/onert/backend/cpu/KernelGenerator.cc
```diff
@@ -122,6 +122,7 @@ void KernelGenerator::visit(const ir::OpSequence &op_seq)
     for (const auto &ind : (node.getInputs() | ir::Remove::UNDEFINED) + node.getOutputs())
     {
       auto tensor = _tensor_builder->at(ind);
+      assert(tensor->layout() == ir::Layout::NHWC); // CPU backend only support NHWC
```
wateret, line 125: I think this should be in below if statement. I will fix.

wateret:
@ragmani I would like make this affected for external tensors(all IPortableTensors) but `IPortableTensor`s are not ready. It needs#2573 and some works for `TensorRegistry`. I think this PR should wait for it until then 😃

wateret:
@ragmani It is ready. PTAL :)

PR #2573, opened 2020-06-24
[onert] Make cpu backend ops use IPortableTensor

wateret:
Make cpu backend ops use IPortableTensor so it can accept tensors from
other backends.

Draft: #1325


hseok-oh:
@wateret Please resolve conflict

wateret:
@ragmani Rebased, PTAL :)

PR #2585, opened 2020-06-24
[tools] CMake: Unify linking boost

wateret:
Use variable `${Boost_PROGRAM_OPTIONS_LIBRARY}` rather than plain
strings. This may resolve both cases(if-else branch).


wateret:
It seems like for Android, most tools are disabled so I was not able to try on Android.

wateret:
`infra/nnfw/cmake/packages/BoostConfig.cmake` eventually calls(no matter it is built or external) system's `find_package` so `Boost_<C>_LIBRARY` will be always set if found. And it is available in any 3.x versions.

Issue #2484, opened 2020-06-22
Compiler FE: introduce tf2nnpackage-value-remote-test


mhs4670go:
There is a test that tests the flow of end-to-end from compiler to runtime(nnpackage test). But this test uses deprecated module as of now, which is why I will introduce a new test.

For test, compiler should pass test resources;nnpackage(circle,info,input,golden). The tests I will introduce from now have two purposes.
- end-to-end test
- To extract test resources more easily.
  - To do this, all resources for testing will be created during the cmake process. This is to conveniently use resources for testing even before actual ctest, or to stash it more easily in CI.

The overall flow is shown in the figure below.

![image](https://user-images.githubusercontent.com/24587354/85271552-e870ac80-b4b5-11ea-9261-d5e1254af199.png)

Related: #2113

PR #2586, opened 2020-06-24
[DRAFT] Introduce tf2nnpackage-value-remote-test

mhs4670go:
This draft introduces tf2nnpackage-value-remote-test

WIP..

Related : #2484 


compiler/tf2nnpackage-value-remote-test/CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
+# TODO add pbtxt
+
+find_package(HDF5 COMPONENTS CXX QUIET)
+
+if(NOT HDF5_FOUND)
+  return()
```
seanshpark, line 6: how about adding a message before return so that we can identify just returned

compiler/tf2nnpackage-value-remote-test/src/TestDataGenerator.cpp
```diff
@@ -0,0 +1,187 @@
...
+      // create value
+      std::vector<hsize_t> dims(input_node->rank());
+      for (uint32_t d = 0; d < input_node->rank(); d++)
+      {
+        dims.at(d) = input_node->dim(d).value();
+        assert(dims.at(d) >= 0);
+      }
+      auto dataspace = std::make_unique<H5::DataSpace>(dims.size(), dims.data());
+      auto dataset = std::make_unique<H5::DataSet>(input_file.createDataSet(
+          "value/" + std::to_string(input_index), H5::PredType::NATIVE_FLOAT, *dataspace));
```
jinevening, line 123: There can be non-float inputs. It would be better to set the datatype based on the type of `input_node`.

mhs4670go: @jinevening I'll add TODO.

PR #2589, opened 2020-06-24
[onert] Introducing ConstMemoryManager

hyunsik-yoon:
`DynamicMemoryManager` was originally used for managing _const_ tensors and later, used for _dynamic_ tensors.

This PR creates `ConstMemoryManager`, which is a copy of `DynamicMemoryManager`. However, `deallocate(ind)` method, which was written only for _dynamic_ tensors, was removed from `ConstMemoryManager`.

### DIFF
--- a/runtime/onert/core/include/backend/cpu_common/MemoryManager.h
+++ b/runtime/onert/core/include/backend/cpu_common/MemoryManager.h
@@ -29,6 +29,9 @@ namespace backend
 namespace cpu_common
 {
 
+/**
+ * @brief MemoryManager for static tensors whose memory is allocated at compilation phase
+ */
 class MemoryManager : public backend::IMemoryManager
 {
 public:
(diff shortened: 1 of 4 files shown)


ragmani:
Is there another reason why you introduce `ConstMemeryManager` except `deallocate(ind)`? If not, I would like to keep the memory manager of constants as `DynamicMemoryManager`.

hyunsik-yoon:
hmm.. I am planning to modifying `DynamicMemoryManager`. Well, I will also check if modified `DynamicMemoryManager` also work with const. If there's some problem, I will create `ConstMemoryManager`.

ragmani:
@hyunsik-yoon 
Some kernels can free memory allocated by `DynamicMemoryManager` if constant tensor of those kernels doesn't the allocated memory in any places. But it is not used `deallocate(ind)`, it just reset `Allocator`'s base to nullptr.
Anyway, I'm wondering why introducing `ConstMemoryManager` instead of `DynamicMemoryManager` for constant tensors. Could you tell me about your plan modifying `DynamicMemoryManager` in more detail?

hyunsik-yoon:
Please refer to #2612.
I was afraid that managing memory for const and dynamic would be different but it seemed that #2612 work both for const and dynamic tensors.

hyunsik-yoon:
https://github.com/Samsung/ONE/pull/2612 was merged.

PR #2597, opened 2020-06-25
[mio-circle] Revise cmake depends

seanshpark:
This will revise cmake command depends with existing SCHEMA_FILE


seanshpark:

PR #2601, opened 2020-06-25
[onert/1.4] Use flatbuffers dev package on gbs build

hseok-oh:
GBS build uses flatbuffers devel package and link libflatbuffers.so to use tensorflow lite
Change test build default: not build


hseok-oh:
For #2592
Cherry-pick #2601 and change `test_build` default value.

PR #2611, opened 2020-06-25
[onert] Support Split int32 on cpu

ragmani:
This commit makes Split supports int32 on cpu backend.


runtime/onert/backend/cpu/ops/OperationUtils.h
```diff
@@ -194,7 +194,39 @@ bool HaveSameShapes(const Tensor *input1, const Tensor *input2);
 
 int32_t CalculateInputRadius(int input_integer_bits, int input_left_shift);
 
-uint32_t sizeOfData(OperandType type, const std::vector<uint32_t> &dimensions);
```
hseok-oh, line 197: @ragmani Can't we fix this vector type to `vector<int32_t>` not using template?

ragmani: Yes, we can fix it to `vector<int32_t>` because this function is not being used now. Do you want it?

hseok-oh: Yes. I prefer to not using template.

PR #2612, opened 2020-06-25
[onert] Enhancing memory management for dynamic tensor

hyunsik-yoon:
This fixes some (possible) memory issue of dynamic tensor like the following:

0. `DynamicMemoryManager::_mem_alloc_map` now stores tensors only when they are allocated.
1. `allocate()` checks if the tensor is already allocated, preventing duplicate applcations
2. When calling `deallocate(tensor_index)`, its memory is explicitly released
   - Previously, its ref count was decreased so the memory might not be released if other still hold the shared_ref.
3. After `deallocInput()` and `deallocSubghraphOutput()`, input/output tensors are explicitly marked as "the tensor does not have memory allocated". 
   - Previously, the tensor might have `_allocator` even after `dealloc()`, and `Tensor::allocator != nullptr`, which is not our expectation.

### DIFF
--- a/runtime/onert/core/src/backend/cpu_common/DynamicTensorManager.cc
+++ b/runtime/onert/core/src/backend/cpu_common/DynamicTensorManager.cc
@@ -115,10 +115,13 @@ void DynamicTensorManager::deallocInput(ir::OperationIndex op_ind)
   auto &input_set = find->second;
   for (auto input_ind : input_set)
   {
-    if (!_tensors->getManagedTensor(input_ind)->is_dynamic())
+    auto *tensor = _tensors->getManagedTensor(input_ind).get();
+    if (!tensor->is_dynamic())
       continue;
 
     _dynamic_mem_mgr->deallocate(input_ind);
+    tensor->resetBuffer();
+
     VERBOSE(DynamicTensorManager) << "Deallocating #" << input_ind.value()
                                   << " (input of op_ind: " << op_ind.value() << ")" << std::endl;
   }
@@ -126,10 +129,13 @@ void DynamicTensorManager::deallocInput(ir::OperationIndex op_ind)
 
 void DynamicTensorManager::deallocSubgraphOutput(ir::OperandIndex output_ind)
 {
-  if (!_tensors->getManagedTensor(output_ind)->is_dynamic())
+  auto *tensor = _tensors->getManagedTensor(output_ind).get();
+  if (!tensor->is_dynamic())
     return;
 
   _dynamic_mem_mgr->deallocate(output_ind);
+  tensor->resetBuffer();
+
   VERBOSE(DynamicTensorManager) << "Deallocating #" << output_ind.value()
                                 << " (output of a subgraph)" << std::endl;
 }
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/backend/cpu_common/DynamicTensorManager.cc
```diff
@@ -115,21 +115,27 @@ void DynamicTensorManager::deallocInput(ir::OperationIndex op_ind)
   auto &input_set = find->second;
   for (auto input_ind : input_set)
   {
-    if (!_tensors->getManagedTensor(input_ind)->is_dynamic())
+    auto *tensor = _tensors->getManagedTensor(input_ind).get();
+    if (!tensor->is_dynamic())
       continue;
 
     _dynamic_mem_mgr->deallocate(input_ind);
+    tensor->resetBuffer();
```
ragmani, line 123: Is this necessary here?

wateret: > 3. After `deallocInput()` and `deallocSubghraphOutput()`, input/output tensors are explicitly marked as "the tensor does not have memory allocated". 
>    - Previously, the tensor might have `_allocator` even after `dealloc()`, and `Tensor::allocator != nullptr`, which is not our expectation.

This may be the reason? Not sure though😅

runtime/onert/core/src/backend/cpu_common/DynamicTensorManager.cc
```diff
@@ -115,21 +115,27 @@ void DynamicTensorManager::deallocInput(ir::OperationIndex op_ind)
...
 
 void DynamicTensorManager::deallocSubgraphOutput(ir::OperandIndex output_ind)
 {
-  if (!_tensors->getManagedTensor(output_ind)->is_dynamic())
+  auto *tensor = _tensors->getManagedTensor(output_ind).get();
+  if (!tensor->is_dynamic())
     return;
 
   _dynamic_mem_mgr->deallocate(output_ind);
+  tensor->resetBuffer();
```
wateret, line 137: Just to check if I understood correctly,

These two calls must always be called together?

runtime/onert/core/src/backend/cpu_common/DynamicTensorManager.cc
hyunsik-yoon: `deallocate(..)` eventually releases calls `allocator.release()` which resets `_base` of `allocator`.
but this does not mean that `tensor->allocacator != nullptr`.
This could make some possible error later, e.g., in `Tensor.cc`, we check if tensor has memory or not by calling, e.g., `assert(tensor->allocator == nullptr)` but such call is currently not right after calling `deallocate(..)`

runtime/onert/core/src/backend/cpu_common/DynamicTensorManager.cc
hyunsik-yoon: I thought that calling these two calls together is safer.

Issue #2609, opened 2020-06-25
Compile FE: problems with in-house models (2)


seanshpark:
Let's gather problems that came from our in-house models

previous alike: #2194

seanshpark:
Problems
- [ ] shape inference for CircleRange: inputs may be other than CircleConst
- [ ] shape inference for CircleExpandDims : need to fix to follow range of  [ -(rank+1), rank ]

seanshpark:
done

PR #2618, opened 2020-06-25
DRAFT Compiler FE fix for in-house models

seanshpark:
Draft code of fixes to support in-house models


seanshpark:
For #2609

seanshpark:
all landed

PR #2620, opened 2020-06-25
[luci] Fix shape inference

seanshpark:
This will fix shape inference for CircleRange, CircleStridedSlice that may have non CircleConst inputs
and CircleExpandDims axis range


seanshpark:

PR #2625, opened 2020-06-25
[res] TFlite schema from hash for latest version

seanshpark:
This will introduce tflite schema from specific hash to utilize BatchMatMul Op


seanshpark:
For #2609 , from draft #2618

llFreetimell:
@seanshpark AFAIK, this is for supporting `tf-nightly-cpu==2.3.0.dev20200623`. If it is right, using `e5023a1738cce7efcdf9d87863b85c80ab2f8c9e` seems better because `tf-nightly-cpu==2.3.0.dev20200623` is released by  `e5023a1738cce7efcdf9d87863b85c80ab2f8c9e`.

Issue #2617, opened 2020-06-25
[circle-quantizer] Separate quantization functionalities from circle2circle


jinevening:
Currently, quantization is supported by `circle2circle`, which not only performs quantization but also various optimizations. To keep the tool smaller and simpler, it would be better to make a separate tool for quantization.

For this, following things must be done.

- [x] 1. Remove the codes related to quantization from circle2circle #2621

- [x] 2. Update `quantize` function in luci (`quantize` should be revised for being called alone) #2629

- [x] 3. Introduce a new executable circle-quantizer #2645

- [x] 4. Update pota-quantization-value-test #2668

jinevening:
@seanshpark I think 1-2-3 is the least number of steps. 3 must be done after 2 (otherwise, the circle-quantizer will not work properly). And it is better to do 2 after 1 (2 may affect the current circle2circle).

seanshpark:
Have to thought of the name for this? `circlequantizer` ?

jinevening:
@seanshpark I got it. For now, I will use `circle-quantizer` for the consistency with circle-inspect, circle-verify, etc. I will avoid using `_` as much as possible :)

jinevening:
I close this issue because all tasks were done.

PR #2629, opened 2020-06-25
[luci] Shape/type inference in quantize

jinevening:
This does shape/type inference in quantize

---
Related to #2617

### DIFF
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -161,7 +161,7 @@ void CircleOptimizer::quantize(loco::Graph *g) const
 
   logo::Phase phase;
 
-  // Do Shape/Type inference (this must be done before luci::validate is called)
+  // Do Shape/Type inference
   phase.emplace_back(std::make_unique<luci::ShapeInferencePass>());
   phase.emplace_back(std::make_unique<luci::TypeInferencePass>());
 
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -158,6 +158,17 @@ void CircleOptimizer::quantize(loco::Graph *g) const
                                            str_to_granularity(granularity));
     quantizer.run(g);
   }
+
+  logo::Phase phase;
+
+  // Do Shape/Type inference (for validation)
```
seanshpark, line 164: Don't exactly undersand why this is `for validation`.  May be add some more comments as notes?

jinevening: This is for passing `luci::validate`.

`circle-quantizer` will call `luci::validate` for validation after quantization (like `circle2circle` calls `luci::validate` after optimization). To pass the `luci::validate`, type and shape must be inferred in advance.

seanshpark: Purpose of the shape inference and dtype inference is to set dtype and shape information to nodes that may not have it for new added by some transformations. 
Purpose of `validate` is to do minimal validation of the graph, including shape and dtype existance.
So doing shape and dtype inference actually is not for validation but for the model itself.
Here, dtype may be updated so may have to run dtype inference to update the information.
For shape, I'm not sure we need to run.

As for conclusion of the question, I think we can remove th comment `(for validation)`

seanshpark: I think shape inference should be executed. If only quantize is called, export may require both inference...

PR #2632, opened 2020-06-25
[onert] Move IPermuteFunction.h

hseok-oh:
Move IPermuteFunction.h into src/


hseok-oh:
I've updated to fix build error. PTAL

PR #2644, opened 2020-06-25
[onert] Change param axes of Mean operation to input

ragmani:
This commit changes param axes of Mean operation to input.
  - Add helper functions converting axes from buffer for each backends
  - Modify ir::operation::Mean, base_loader, OperationFactory, KernelGenerators

### DIFF
--- a/runtime/onert/backend/cpu/KernelGenerator.cc
+++ b/runtime/onert/backend/cpu/KernelGenerator.cc
@@ -82,6 +82,38 @@
 
 #include <stdexcept>
 
+namespace
+{
+
+using namespace onert;
+
+std::vector<int32_t> getReducerAxes(const ir::Operand &axes)
+{
+  std::vector<int32_t> ret;
+
+  assert(axes.isConstant());
+  switch (axes.typeInfo().type())
+  {
+    case ir::DataType::INT32:
+    {
+      const auto tmp = axes.asVector<int32_t>();
+      ret.insert(ret.begin(), tmp.begin(), tmp.end());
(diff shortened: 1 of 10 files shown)


wateret:
Just one question. Does it work for dynamic AXES with acl_cl/neon?

ragmani:
> Just one question. Does it work for dynamic AXES with acl_cl/neon?

No. There are some issues. The issue directly related to AXES is that most kernels using AXES don't have AXES as tensor. But our runtime does not support dynamic without tensor.

runtime/onert/backend/cpu/KernelGenerator.cc
```diff
@@ -1188,15 +1219,17 @@ void KernelGenerator::visit(const ir::operation::Mean &node)
 {
   const auto output_index{node.getOutputs().at(0)};
   const auto input_index{node.getInputs().at(ir::operation::Mean::INPUT)};
+  const auto axes_index{node.getInputs().at(ir::operation::Mean::AXES)};
 
-  const auto axes = node.param().axes;
+  const auto &axes = _ctx.at(axes_index);
+  const auto axes_vec = getReducerAxes(axes);
```
hyunsik-yoon, line 1225: I wonder what happen if axes is output of other operation.
Shouldn't we check `if (axes.isConstant())` to call this function?

ragmani: There is no handling dynamic axes now. I will add an assertion instead.

PR #2645, opened 2020-06-25
Introduce circle-quantizer for pota quantization

jinevening:
This introduces circle-quantizer for post-training affine quantization

---
Related to #2617


jinevening:
**For reviewers**

Most of the code was copied from circle2circle. The only meaningful difference is that circle-quantizer calls `quantize` instead of `optimize`.

PR #2650, opened 2020-06-25
[tflite2circle] Support BatchMatMul

seanshpark:
This will enable conversion from tflite to circle of BatchMatMul Op


seanshpark:
For #2609 , from draft #2618

seanshpark:

Issue #2623, opened 2020-06-25
[onert] nnpkg_test.sh passes when Not comparable (different dimension tensors)


dayo09:
- I think this can make confusion while model debugging so I just made this issue, please note.

### Passed
```bash
BACKENDS="cpu" tools/nnpackage_tool/nnpkg_test/nnpkg_test.sh -i ../model_bcq bcq_tran
[  Run  ] ../model_bcq/bcq_tran 	Pass
[Compare] ../model_bcq/bcq_tran 	Pass
```

### Opening log file
```bash
group1   group2
---------------------------------------
    x      x                   
    x      x    /0             
    x      x    /1             

group  : </value> and </value>
0 differences found
dataset: </value/0> and </value/0>
Not comparable: </value/0> has rank 2, dimensions [8x1536], max dimensions [8x1536]
and </value/0> has rank 4, dimensions [1x6x2x1536], max dimensions [1x6x2x1536]
0 differences found
dataset: </value/1> and </value/1>
Not comparable: </value/1> has rank 4, dimensions [1x6x2x1536], max dimensions [1x6x2x1536]
and </value/1> has rank 2, dimensions [8x1536], max dimensions [8x1536]
0 differences found
```
- __0 differences found__ : it was not comparable so there was 0 differences. I think this is the reason why it 'passed' even it's not passed

dayo09:
### Unexpected behavior of h5diff
- h5diff exits with 0 when "Some objects are not comparable" : http://hdf-forum.184993.n3.nabble.com/h5diff-exits-with-0-when-quot-Some-objects-are-not-comparable-quot-td4025252.html

dayo09:
### Non-comparable of h5diff
- https://support.hdfgroup.org/pubs/rfcs/RFC_h5diff_NonComparable.pdf
- According to the report above, there are five non-comparable case
  - Empty datasets 
  - Different datatype classes or H5T_TIME class or H5T_COMPOUND class 
  - Different dataspace ranks 
  - Different dataspace dimensions
  - Different sign properties
  - Invalid numeric operation in relative error calculation
- It's printing below with -v(verbose) set, exit 0.
```bash
--------------------------------
Some objects are not comparable
--------------------------------
Use -c for a list of non-comparable objects. 
```

PR #2658, opened 2020-06-26
[onert] Fix nnpkg_test.sh to fail on 'not comparable'

dayo09:
- Issue
   - Test passes when it's 'not comparable'(dimension unmatch, rank unmatch, zero tensor..)
- Changes
   - Fail on 'not comparable'
   - Print `dumped.log` even if it passes, to check
- related issue : #2623

### DIFF
--- a/tests/scripts/nnpkg_test.sh
+++ b/tests/scripts/nnpkg_test.sh
@@ -111,13 +111,26 @@ expected="$nnpkg/metadata/tc/expected.h5"
 
 echo -n "[Compare] $nnpkg "
 
-if $difftool -d 0.001 -v "$dumped" "$expected" /value >& "$dumped.log"; then
-  echo -e "\tPass"
-  rm "$dumped" "$dumped.log"
-else
+test_fail()
+{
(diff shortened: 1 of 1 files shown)


dayo09:
https://github.com/Samsung/ONE/issues/2623#issuecomment-649906051
> ### Non-comparable of h5diff
> * https://support.hdfgroup.org/pubs/rfcs/RFC_h5diff_NonComparable.pdf
> * According to the report above, there are five non-comparable case
>   
>   * Empty datasets
>   * Different datatype classes or H5T_TIME class or H5T_COMPOUND class
>   * Different dataspace ranks
>   * Different dataspace dimensions
>   * Different sign properties
>   * Invalid numeric operation in relative error calculation
> * It's printing below with -v(verbose) set, exit 0.
> 
> ```shell
> --------------------------------
> Some objects are not comparable
> --------------------------------
> Use -c for a list of non-comparable objects. 
> ```

hseok-oh:
@glistening PTAL

Issue #42, opened 2020-04-17
Compiler FE: Update build module list for release


seanshpark:
Lets update build module list for release of '20
- This list will be used for TestCase, LineCoverage and SAM

seanshpark:
Let's update for new `circle-quantizer` module

seanshpark:
Update with TF 2.3.0-rc0 official release from previous nightly version.

seanshpark:
```
tf2nnpkg.20200630
```
is missing

seanshpark:
I think we can now close this.

PR #2666, opened 2020-06-26
[infra] Add circle-quantizer to compiler coverage

seanshpark:
This will add recently introduced circle-quantizer to compiler coverage


seanshpark:
Related #42

PR #2669, opened 2020-06-26
Fix wrong parameter on random_sample

chunseoklee:
- Remove type parameter from call


llFreetimell:
Can you fix Line 141 either? :)

glistening:
@chunseoklee You've removed the parameter for `np.array`. `random_sample` takes 1 parameter, which is correct. Also, removed version raises error for my given input. I would like to roll back after checking what was your error.

PR #2670, opened 2020-06-26
[Draft] Check whether 'NPROCS' work

cgbahk:
Let's see whether `NPROCS` setting works

Related: #2673


cgbahk:
Checked it's not working. See #2673

PR #2672, opened 2020-06-26
[onert] Add some dimension checks for MatrixBandPart

dayo09:
- Add some dimension checks throwing runtime error


dayo09:
- https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/matrix-band-part
```
input: Rank k tensor.
num_lower: 0-D tensor. Number of subdiagonals to keep. If negative, keep entire lower triangle.
num_upper: 0-D tensor. Number of superdiagonals to keep. If negative, keep entire upper triangle.
```

PR #2676, opened 2020-06-26
[pota-value-test] Conv2D layer-wise uint8 test

jinevening:
This adds Conv2D_004 quantization value test

---
Related to #2088


seanshpark:
Q) how did you get the numbers? can you write a how-to document so that others or future you can reproduce?

jinevening:
I generated those values based on the Python scripts written by the app part. I will see if I can make the how-to document, but I cannot assure to make it for general models, because the script was very hard-coded.

seanshpark:
> but I cannot assure to make it for general models, because the script was very hard-coded.

I think specific for this material, NOT GENERAL, is OK for now. This is to document the steps as no one will make another one when another kind or value of material is needed.

Issue #2655, opened 2020-06-26
[onert] Improvement performance of OpSequences::findOperation


periannath:
This issue comes from https://github.com/Samsung/ONE/issues/2610#issuecomment-649280855

`OpSequences::findOperation` traverses whole operations to find target operation.

It seems that function can be optimized in many ways.

1. Early stop when operation is found
1. Create <OperationIndex, OpSequenceIndex> table.
1. ...

I'll implement options and compare the performance.

- [x] Implement early stop when operation is found
- [x] Implement creation of <OperationIndex, OpSequenceIndex> table
- [x] Compare performance
- [x] Upload PRs #2678 #2816

periannath:
### Current master

```bash
$ BACKENDS=cpu ./Product/armv7l-linux.release/out/bin/nnpackage_run --nnpackage ..
... run 1 takes 12.232 ms
===================================
MODEL_LOAD   takes 34.830 ms
PREPARE      takes 1533.274 ms
EXECUTE      takes 12.232 ms
- MEAN     :  12.232 ms
- MAX      :  12.232 ms
- MIN      :  12.232 ms
- GEOMEAN  :  12.232 ms
===================================

```

periannath:
### Measurement of current master

```bash
$ BACKENDS=cpu ./Product/armv7l-linux.release/out/bin/nnpackage_run --nnpackage ..
... run 1 takes 12.232 ms
===================================
MODEL_LOAD   takes 34.830 ms
PREPARE      takes 1533.274 ms
EXECUTE      takes 12.232 ms
- MEAN     :  12.232 ms
- MAX      :  12.232 ms
- MIN      :  12.232 ms
- GEOMEAN  :  12.232 ms
===================================
```

- Prepare : **1533 ms**

periannath:
### Testing early stop

#### Master
https://github.com/Samsung/ONE/blob/52bc738295f267dd75c265e9cee3e5ff1e855a06/runtime/onert/core/src/ir/OpSequences.cc#L73-L84
- Prepare time : **1533 ms**

#### Trial 1 : Add return function

```patch
     for (const auto &op_idx : object.operations())
     {
       if (op_idx == operation_index)
+      {
         ret = index;
+        return;
+      }
     }
   });
   return ret;
```
- Prepare time : **1060 ms**

periannath:
### Testing early stop (cont'd)

#### Trial 2 : Access operations directly instead of using `iterate` function

[...]

PR #2678, opened 2020-06-26
[onert] Skip unnecessary traverse in OpSequences::findOperation

periannath:
- This commit skips unnecessary traverse in OpSequences::findOperation
  - Return immediately after the return value is found

---

Related issue : #2655


periannath:
ping @Samsung/nnfw_committers

PR #2691, opened 2020-06-29
DRAFT Support TFlite BatchMatMul

seanshpark:
On going draft to support TFlite BatchMatMul


seanshpark:
all landed

Issue #2652, opened 2020-06-26
Compiler FE: custom to built-in chef migration


seanshpark:
To support our in-house model, we found that BatchMatMul Op is needed as built-in operator.
TensorFlow official 2.2 converts this to Custom Op so current recipe is to generate Custom Op.

From the task #2609, we will support as built-in Op.
This issue is about custom operator circle chef to built-in of tflite chef migration.

seanshpark:
Naming issue: add `Custom` prefix?

seanshpark:
BatchMatMul test material is created with 
- CircleRecipes/BatchMatMul_000/test.recipe
- compiler/circlechef/core/src/Op/BatchMatMul.cpp

To-be
- TensorFlowLiteRecipes/BatchMatMul_000/test.recipe
- compiler/tflchef/core/src/Op/BatchMatMul.cpp

seanshpark:
done

PR #2693, opened 2020-06-29
[res] CircleBatchMatMul_000

seanshpark:
This will introduce CircleBatchMatMul_000, copy of BatchMatMul_000 to avoid name conflict with tflite BatchMatMul_000


seanshpark:
For #2652, from draft #2691

mhs4670go:
We already have `BatchMatMul_000` in `CircleRecipes`. Can't we just change its name to `CircleBatchMatMul_000`?

seanshpark:
>  Can't we just change its name to CircleBatchMatMul_000?

luci test.lst file uses BatchMatMul_000 and drops `Missing recipe of 'BatchMatMul_000' test`

PR #2700, opened 2020-06-29
[circle2circle dredd] Use CircleBatchMatMul_000

seanshpark:
This will revise to use CircleBatchMatMul_000 for Circle BatchMatMul Op


seanshpark:
For #2652, from draft #2691

PR #2701, opened 2020-06-29
[res] temporary comment for BatchMatMulV2_001

seanshpark:
This will temporary comment batch_matmul_options before revise to adj_x


seanshpark:
For #2652, from draft #2691

PR #2704, opened 2020-06-29
[luci-interpreter] set detphwiseConv2d dilation

mhs4670go:
This commit enables depthwiseConv2d to set dilation from graph


mhs4670go:
@seanshpark merge plz:)

PR #2706, opened 2020-06-29
[tflchef] Support BatchMatMul

seanshpark:
This will enable support of BatchMatMul
- fix option to use adj_x and adj_y


seanshpark:
For #2652, from draft #2691

seanshpark:

PR #2713, opened 2020-06-29
[res] Remove old circle BatchMatMul

seanshpark:
This will remove BatchMatMul circle recipe that is not used anymore


seanshpark:
For #2652, from draft #2691

seanshpark:
Will be rebase after #2714 lands

Issue #2654, opened 2020-06-26
Docs: supported-operations.md


seanshpark:
Let's work on this document

seanshpark:
@hseok-oh , please ping me when runtime op list in mosaic is updated :)

hseok-oh:
@seanshpark mosaic is updated

seanshpark:
done

PR #2715, opened 2020-06-29
[docs] Revise overview supported-operations.md

seanshpark:
This will update overview supported-operations.md file with supported operators as of '20-06-26


seanshpark:
For #2654

PR #2716, opened 2020-06-30
[res] introduce tflite recipe BatchMatMul_000

seanshpark:
This will introduce tflite recipe BatchMatMul_000


seanshpark:
For #2652, from draft #2691

PR #2717, opened 2020-06-30
[luci] Test BatchMatMul

seanshpark:
This will enable test of BatchMatMul Op


seanshpark:
For #2652, from draft #2691

PR #2718, opened 2020-06-30
[cker] Add scalar case to the function to check matching of shapes

ragmani:
This commit adds scalar case to the function to check matching of shapes.


hseok-oh:
@ragmani It makes `shape: [1]` and `check_shape: [1, 1, 1]` return true. Is it OK?

ragmani:
@hseok-oh I'm not sure it is OK. I will add checking rank.

hseok-oh:
@llFreetimell PTAL

PR #2723, opened 2020-06-30
[infra] Update abseil-cpp version

hseok-oh:
- Update abseil-cpp latest release version
- Runtime (tensorflow-lite lib) uses updated abseil-cpp


infra/cmake/packages/AbseilSourceConfig.cmake
```diff
@@ -12,10 +12,14 @@ function(_AbseilSource_import)
   #
   # The last change of "48cd2c3f351" was commited on 2018.09.27
```
parjong, line 13: Why don't you remove this line?

PR #2725, opened 2020-06-30
Rename PARALLEL_BUILD to ALL_PROC_BUILD

cgbahk:
This commit renames top level option `PARALLEL_BUILD` to `ALL_PROC_BUILD`
which describes its role better.

Signed-off-by: Cheongyo Bahk <ch.bahk@samsung.com>

---

IMHO, option name `PARALLEL_BUILD` is ambiguous as it does not describe its role correctly. Parallel build possible with setting `NPROCS`.

What it actually does is to set `NPROCS` as a number of cpu processor in host system.

Please propose better name if you have :)

### DIFF
--- a/Makefile.template
+++ b/Makefile.template
@@ -4,7 +4,7 @@ BUILD_TYPE?=Debug
 CROSS_BUILD?=0
 HOST_OS?=linux
 TARGET_OS?=linux
-PARALLEL_BUILD?=1
+ALL_PROC_BUILD?=1
 COVERAGE_BUILD?=0
 BENCHMARK_ACL_BUILD?=0
 OPTIONS?=
@@ -78,7 +78,7 @@ ifneq ($(ANDROID_BOOST_ROOT),)
   OPTIONS+= -DANDROID_BOOST_ROOT=$(ANDROID_BOOST_ROOT)
 endif
 
-ifeq ($(PARALLEL_BUILD),1)
+ifeq ($(ALL_PROC_BUILD),1)
 	# Get number of processors (linux only for now)
 	ifeq ($(HOST_OS),linux)
 		NPROCS?=$(shell grep -c ^processor /proc/cpuinfo)
@@ -140,7 +140,7 @@ endif
 	touch $(TIMESTAMP_CONFIGURE)
 
 build_internal: $(BUILD_FOLDER)
-	NNFW_WORKSPACE="$(WORKSPACE)" NPROCS="$(NPROCS)" PARALLEL_BUILD="$(PARALLEL_BUILD)" ./nnfw build
+	NNFW_WORKSPACE="$(WORKSPACE)" NPROCS="$(NPROCS)" ALL_PROC_BUILD="$(ALL_PROC_BUILD)" ./nnfw build
 	rm -rf $(BUILD_ALIAS)
 	ln -s $(BUILD_FOLDER) $(BUILD_ALIAS)
 	touch $(TIMESTAMP_BUILD)
--- a/infra/nnfw/command/build
+++ b/infra/nnfw/command/build
@@ -11,9 +11,9 @@ fi
 # TODO Use argument instead of environment variable
 HOST_OS=${HOST_OS:-linux}
 NPROCS=${NPROCS:-1}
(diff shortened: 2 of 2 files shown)


seanshpark:
I've know this as parallel build... https://www.gnu.org/software/make/manual/html_node/Parallel.html
`All proc build` is new to me... :)

cgbahk:
> I've know this as parallel build...

Right. It's still parallel build.

This PR is after all `type/discussion`, so let me explain my intention with example :)

||It reads like...|but actually...|
|--|--|--
|`NPROCS=4` `PARALLEL_BUILD=1`|I want to do parallel build with `-j4!`|Build with `-j$(nproc)`, in case of CircleCI `-j88`
|`NPROCS=4` `PARALLEL_BUILD=0`|Give `NPROCS`, but no parallel build? I'm :confused:|Build with `-j4`

My intension in this PR is to fill the gap between `It reads like...` and `but actually...` section.
Other solution might be just removing `PARALLEL_BUILD` option. IMO `NPROCS` is enough and explicit. It also not used in CI currently.

> `All proc build` is new to me... :)

I am not 100% confident with this name too :sob:

seanshpark:
Maybe we should do parallem build with `-j$(PARALLEL_BUILD)` if `PARALLEL_BUILD >= 1`
and no `-j` option when `PARALLEL_BUILD` is 0 or not defined...

cgbahk:
> Maybe we should do parallel build with ...

That's exactly what `NPROCS` options doing, as I understand :) That is why I said 



> Other solution might be just removing `PARALLEL_BUILD` option. IMO `NPROCS` is enough and explicit. It also not used in CI currently.

seanshpark:
>  IMO NPROCS is enough and explicit. It also not used in CI currently.

`NPROCS` is strange to me as an option or switch for parallel build...

cgbahk:
Then `DEGREE_OF_PARLLELISM`? :smile: It has usecase https://docs.oracle.com/database/121/SUTIL/GUID-B3EAC57A-3AEB-42F9-9438-4CB077BF7383.htm#SUTIL3978

hseok-oh:
>Build with -j$(nproc), in case of CircleCI -j88
>Build with -j4

This is a bug. We can fix by below patch.

```diff
--- a/infra/nnfw/command/build
+++ b/infra/nnfw/command/build
@@ -10,15 +10,15 @@ fi
 # Set parallel build
 # TODO Use argument instead of environment variable
 HOST_OS=${HOST_OS:-linux}
-NPROCS=${NPROCS:-1}
 PARALLEL_BUILD=${PARALLEL_BUILD:-1}
 
-if [ "${PARALLEL_BUILD}" == "1" ]; then
+if [ "${PARALLEL_BUILD}" = "1" ]; then
   # Get number of processors (linux only for now)
-  if [ "${HOST_OS}" == "linux" ]; then
-    NPROCS="$(grep -c ^processor /proc/cpuinfo)"
+  if [ "${HOST_OS}" = "linux" ]; then
+    NPROCS=${NPROCS:-(grep -c ^processor /proc/cpuinfo)}
   fi
 fi
+NPROCS=${NPROCS:-1}
 
 cd ${BUILD_PATH}
 make -j ${NPROCS} "$@"
```

We can remove `PARALLEL_BUILD` because we can disable parallel build by `NPROCS=1`.
`PARALLEL_BUILD` comes from https://github.sec.samsung.net/star/nnfw/pull/1328

cgbahk:
> We can remove `PARALLEL_BUILD` because we can disable parallel build by `NPROCS=1`.

I vote for this option :)

seanshpark:
> https://github.sec.samsung.net/star/nnfw/pull/1328

Let's invite the author :) @lemmaa , PTAL

lemmaa:
> > We can remove PARALLEL_BUILD because we can disable parallel build by NPROCS=1.
> > PARALLEL_BUILD comes from https://github.sec.samsung.net/star/nnfw/pull/1328
> 
> Let's invite the author :) @lemmaa , PTAL

@hseok-oh , @seanshpark , sounds good! I agree. :)

cgbahk:
Let's remove `PARALLEL_BUILD` option. I would make new PR

PR #2720, opened 2020-06-30
[Draft/luci-interpreter]Update *.test.cpp in luci-interpreter

struss:
This commit Update *.test.cpp to test uint8 case on operations used in Resnet.

ONE-DCO-1.0-Signed-off-by: KiDeuk Bang <rrstrous@nate.com>

PR #2728, opened 2020-06-30
[luci-interpreter] Test quantized Add kernels for ResNet-v1

struss:
Parent Issue: #2730
Draft PR: #2720

This commit update `Add` operator test case.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Add.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Add.test.cpp
@@ -34,128 +34,138 @@ float GetTolerance(float min, float max)
   return kQuantizedStep;
 }
 
-float getMinValue(std::initializer_list<float> data)
+TEST(AddTest, Uint8)
 {
-  return 0 > std::min<float>(data) ? std::min<float>(data) : 0;
-}
-
-float getMaxValue(std::initializer_list<float> data)
-{
-  return 0 < std::max<float>(data) ? std::max<float>(data) : 0;
-}
-
-template <typename T>
-void Check(std::initializer_list<int32_t> input1_shape, std::initializer_list<int32_t> input2_shape,
-           std::initializer_list<int32_t> output_shape, std::initializer_list<float> input1_data,
-           std::initializer_list<float> input2_data, std::initializer_list<float> output_data)
-{
-  float kQuantizedTolerance = GetTolerance(getMinValue(input1_data) + getMinValue(input2_data),
-                                           getMaxValue(input1_data) + getMaxValue(input2_data));
-  std::pair<float, int32_t> input1_quant_param =
-      quantizationParams<T>(getMinValue(input1_data), getMaxValue(input1_data));
-  std::pair<float, int32_t> input2_quant_param =
-      quantizationParams<T>(getMinValue(input2_data), getMaxValue(input2_data));
-  Tensor input1_tensor{getElementType<T>(),
-                       input1_shape,
-                       {{input1_quant_param.first}, {input1_quant_param.second}},
-                       ""};
-  Tensor input2_tensor{getElementType<T>(),
-                       input2_shape,
-                       {{input2_quant_param.first}, {input2_quant_param.second}},
-                       ""};
-  if (std::is_floating_point<T>::value)
-  {
-    input1_tensor.writeData(input1_data.begin(), input1_data.size() * sizeof(T));
-    input2_tensor.writeData(input2_data.begin(), input2_data.size() * sizeof(T));
-  }
-  else
-  {
-    std::vector<T> quantized_input1_value =
-        quantize<T>(input1_data, input1_quant_param.first, input1_quant_param.second);
-    std::vector<T> quantized_input2_value =
-        quantize<T>(input2_data, input2_quant_param.first, input2_quant_param.second);
-    input1_tensor.writeData(quantized_input1_value.data(),
-                            quantized_input1_value.size() * sizeof(T));
-    input2_tensor.writeData(quantized_input2_value.data(),
-                            quantized_input2_value.size() * sizeof(T));
-  }
-  std::pair<float, int32_t> output_quant_param =
-      quantizationParams<T>(getMinValue(input1_data) + getMinValue(input2_data),
-                            getMaxValue(input1_data) + getMaxValue(input2_data));
-  Tensor output_tensor =
-      makeOutputTensor(getElementType<T>(), output_quant_param.first, output_quant_param.second);
-
-  AddParams params{};
-  params.activation = Activation::RELU;
-
-  Add kernel(&input1_tensor, &input2_tensor, &output_tensor, params);
-  kernel.configure();
-  kernel.execute();
-
-  if (std::is_floating_point<T>::value)
-  {
-    EXPECT_THAT(extractTensorData<T>(output_tensor), ElementsAreArray(ArrayFloatNear(output_data)));
-  }
-  else
-  {
-    EXPECT_THAT(dequantize<T>(extractTensorData<T>(output_tensor), output_tensor.scale(),
-                              output_tensor.zero_point()),
-                ElementsAreArray(ArrayFloatNear(output_data, kQuantizedTolerance)));
-  }
-  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray(output_shape));
-}
-
-template <typename T> class AddTest : public ::testing::Test
-{
-};
-
-using DataTypes = ::testing::Types<float, uint8_t>;
-TYPED_TEST_CASE(AddTest, DataTypes);
-
-TYPED_TEST(AddTest, TotalTest)
-{
-  std::initializer_list<int32_t> input1_shape{2,3,1,2};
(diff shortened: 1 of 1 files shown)


compiler/luci-interpreter/src/kernels/Add.test.cpp
```diff
@@ -27,59 +27,118 @@ namespace
...
 
-    EXPECT_THAT(extractTensorData<float>(output_tensor),
-                ::testing::ElementsAreArray(ArrayFloatNear(test_outputs[i], 0.0001f)))
-        << "With shape number " << i;
+template <typename T>
+void Check(std::initializer_list<int32_t> input1_shape, std::initializer_list<int32_t> input2_shape,
+           std::initializer_list<int32_t> output_shape, std::initializer_list<float> input1_data,
+           std::initializer_list<float> input2_data, std::initializer_list<float> output_data)
+{
+  float kQuantizedTolerance = GetTolerance(getMinValue(input1_data) + getMinValue(input2_data),
```
jinevening, line 52: I don't understand the meaning of this code. Isn't this element-wise add?
input 1 = [1, 2, 3]
input 2 = [3, 2, 1]
input 1 + input 2 = [4, 4, 4] (min = 4, max = 4)
According to your code, min = 2, max = 6.

compiler/luci-interpreter/src/kernels/Add.test.cpp
```diff
@@ -27,59 +27,118 @@ namespace
...
+        quantize<T>(input1_data, input1_quant_param.first, input1_quant_param.second);
+    std::vector<T> quantized_input2_value =
+        quantize<T>(input2_data, input2_quant_param.first, input2_quant_param.second);
+    input1_tensor.writeData(quantized_input1_value.data(),
+                            quantized_input1_value.size() * sizeof(T));
+    input2_tensor.writeData(quantized_input2_value.data(),
+                            quantized_input2_value.size() * sizeof(T));
+  }
+  std::pair<float, int32_t> output_quant_param =
+      quantizationParams<T>(getMinValue(input1_data) + getMinValue(input2_data),
```
jinevening, line 83: same as above

compiler/luci-interpreter/src/kernels/Add.test.cpp
```diff
@@ -27,59 +27,118 @@ namespace
...
-    Tensor output_tensor = makeOutputTensor(DataType::FLOAT32);
+  float kQuantizedStep = (max - min) / 255.0;
+  return kQuantizedStep;
+}
 
-    AddParams params{};
-    params.activation = Activation::RELU;
+float getMinValue(std::initializer_list<float> data)
+{
+  return 0 > std::min<float>(data) ? std::min<float>(data) : 0;
```
jinevening, line 39: Why do you make the minimum value as 0? The data can have negative values as well.

compiler/luci-interpreter/src/kernels/Add.test.cpp
```diff
@@ -27,59 +27,118 @@ namespace
...
+  if (std::is_floating_point<T>::value)
+  {
+    EXPECT_THAT(extractTensorData<T>(output_tensor), ElementsAreArray(ArrayFloatNear(output_data)));
+  }
+  else
+  {
+    EXPECT_THAT(dequantize<T>(extractTensorData<T>(output_tensor), output_tensor.scale(),
+                              output_tensor.zero_point()),
+                ElementsAreArray(ArrayFloatNear(output_data, kQuantizedTolerance)));
   }
```
jinevening, line 104: Can you add tests with exchanged inputs as the previous code did?

compiler/luci-interpreter/src/kernels/Add.test.cpp
jinevening: @seanshpark They are values, not shapes :)  [1, 2, 3] is a vector of length 3, whose values are 1, 2, and 3.

seanshpark:
please merge without my approval or I'll merge when all other reviewers approved

compiler/luci-interpreter/src/kernels/Add.test.cpp
struss: Okay, I will add it.

compiler/luci-interpreter/src/kernels/Add.test.cpp
struss: on Quantization sequence, it must contains value 0 between min and max(on TFlite). so, if min value in data should be less than 0. also, max value in data should be also more than 0.
https://github.com/tensorflow/tensorflow/blob/00fad90125b18b80fe054de1055770cfb8fe4ba3/tensorflow/lite/kernels/test_util.h#L357

compiler/luci-interpreter/src/kernels/Add.test.cpp
struss: in case, if input1 and input2 have a same shape, it will be [4,4,4]. but in case if input1 shape [1,3] and input2 shape [3,1], then output would be [[2,3,4],[3,4,5],[4,5,6]]. I just thought min value is a possible minimal value that inferred from input values. so, it is sum of min value in each inputs. and max value is also the same.

compiler/luci-interpreter/src/kernels/Add.test.cpp
```diff
@@ -27,59 +27,118 @@ namespace
...
+void Check(std::initializer_list<int32_t> input1_shape, std::initializer_list<int32_t> input2_shape,
+           std::initializer_list<int32_t> output_shape, std::initializer_list<float> input1_data,
+           std::initializer_list<float> input2_data, std::initializer_list<float> output_data)
+{
+  float kQuantizedTolerance = GetTolerance(getMinValue(input1_data) + getMinValue(input2_data),
+                                           getMaxValue(input1_data) + getMaxValue(input2_data));
+  std::pair<float, int32_t> input1_quant_param =
+      quantizationParams<T>(getMinValue(input1_data), getMaxValue(input1_data));
+  std::pair<float, int32_t> input2_quant_param =
+      quantizationParams<T>(getMinValue(input2_data), getMaxValue(input2_data));
```
jinevening, line 57: Please move the quantization-related codes to the proper place. E.g., these lines do not have to be called if T is float.

compiler/luci-interpreter/src/kernels/Add.test.cpp
struss: Sorry for make confused. I will use the value in TFlite example value.

compiler/luci-interpreter/src/kernels/Add.test.cpp
struss: Oh, I found that In TFlite test case, it split uint8(or int8) case and float case for some reason. so, I just add a new test case for uint8 case.

compiler/luci-interpreter/src/kernels/Add.test.cpp
```diff
@@ -27,6 +27,94 @@ namespace
...
+{
+  std::initializer_list<int32_t> base_shape = {2, 3, 1, 2};
+  std::initializer_list<float> base_data = {-0.3f, 2.3f, 0.9f,  0.5f, 0.8f, -1.1f,
+                                            1.2f,  2.8f, -1.6f, 0.0f, 0.7f, -2.2f};
+  std::initializer_list<int32_t> test_shapes[] = {
+      {1, 1, 3, 2}, {1, 3, 1, 2}, {2, 1, 3, 1}, {2, 3, 1, 1}};
+  std::initializer_list<float> test_data = {0.2f, 0.3f, -0.4f, 0.5f, 1.0f, 0.9f};
+  std::initializer_list<int32_t> output_shapes[] = {
+      {2, 3, 3, 2}, {2, 3, 1, 2}, {2, 3, 3, 2}, {2, 3, 1, 2}};
+  std::vector<std::vector<float>> output_datas = {
```
jinevening, line 47: how about just `output_data` ? data is plural.

compiler/luci-interpreter/src/kernels/Add.test.cpp
```diff
@@ -27,6 +27,94 @@ namespace
...
+    Add kernel(&input1_tensor, &input2_tensor, &output_tensor, params);
+    kernel.configure();
+    kernel.execute();
+
+    EXPECT_THAT(dequantize<uint8_t>(extractTensorData<uint8_t>(output_tensor),
+                                    output_tensor.scale(), output_tensor.zero_point()),
+                ElementsAreArray(ArrayFloatNear(output_datas[i], kQuantizedTolerance)));
+    EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray(output_shapes[i]));
+  }
+  for (int i = 0; i < 4; i++)
```
jinevening, line 87: How about i < output_data.size() ?

jinevening: Adding a comment would help to read the code. For example, `// Re-run with exchanged inputs.`

struss: Oh I update it, but not applied. sorry......

compiler/luci-interpreter/src/kernels/Add.test.cpp
struss: oh! sorry, I will fix it!

PR #2724, opened 2020-06-30
DRAFT Compiler FE use TensorFlowSource-2.3.0-rc0

seanshpark:
On-going draft to use TensorFlowSource-2.3.0-rc0 release

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #2722, opened 2020-06-30
Compiler FE: Use TensorFlow 2.3.0-rc0


seanshpark:
As TensorFlow has official release canidate for 2.3.0, let's use this version in Compiler frontend
- and remove previous codes using commit hash

seanshpark:
We can't use `2.3.0-rc0` as version number in CMake.
Document says the format is like `major[.minor[.patch[.tweak]]]`
- https://cmake.org/cmake/help/v3.0/command/find_package.html

seanshpark:
done

PR #2729, opened 2020-06-30
[res] TensorFlowLiteSchema from 2.3.0-rc0

seanshpark:
This will introduce TensorFlowLiteSchema from 2.3.0-rc0 version


seanshpark:
For #2722, from draft #2724

PR #2738, opened 2020-06-30
DRAFT Compiler FE: Enable MatrixSetDiag operator

d-krylov:
Enable MatrixSetDiag operator


compiler/luci/lang/src/Nodes/CircleMatrixSetDiag.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+  luci::CircleMatrixSetDiag matrix_set_diag_node;
+
+  ASSERT_EQ(luci::CircleDialect::get(), matrix_set_diag_node.dialect());
+  ASSERT_EQ(luci::CircleOpcode::MATRIX_SET_DIAG, matrix_set_diag_node.opcode());
+
+  ASSERT_EQ(nullptr, matrix_set_diag_node.input());
+  ASSERT_EQ(nullptr, matrix_set_diag_node.diagonal());
+}
+
+TEST(CircleMatrixSetDiagTest, input_NEG)
```
binarman, line 35: I am not sure this is negative test.
According wikipedia: "Negative testing is a method of testing an application or system that ensures that the plot of the application is according to the requirements and can handle the unwanted input and user behavior"

compiler/luci/lang/src/Nodes/CircleMatrixSetDiag.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+  ASSERT_EQ(nullptr, matrix_set_diag_node.diagonal());
+}
+
+TEST(CircleMatrixSetDiagTest, arity_NEG)
+{
+  luci::CircleMatrixSetDiag matrix_set_diag_node;
+
+  ASSERT_NO_THROW(matrix_set_diag_node.arg(0));
+  ASSERT_NO_THROW(matrix_set_diag_node.arg(1));
+  ASSERT_THROW(matrix_set_diag_node.arg(2), std::out_of_range);
```
binarman, line 59: Indeed, this is a "negative" test.
`matrix_set_diag_node.arg(2)` is invalid usage of interface and it should throw.

compiler/luci/lang/src/Nodes/CircleMatrixSetDiag.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+TEST(CircleMatrixSetDiagTest, visit_mutable_NEG)
+{
+  struct TestVisitor final : public luci::CircleNodeMutableVisitor<void>
+  {
+  };
+
+  luci::CircleMatrixSetDiag matrix_set_diag_node;
+
+  TestVisitor tv;
+  ASSERT_THROW(matrix_set_diag_node.accept(&tv), std::exception);
```
binarman, line 71: Not sure this test checks operator functionality.
Looks like test for Visitor =)

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -1023,6 +1023,27 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
...
+    LUCI_ASSERT(rank == input_shape.rank() - 1, "diagonal rank = input rank - 1");
+
+    for (uint32_t i = 0; i < rank - 1; i++)
+    {
+      LUCI_ASSERT(diagonal_shape.dim(i) == input_shape.dim(i), "diagonal dims = input dims");
+    }
+
+    auto dim = std::min(input_shape.dim(rank - 1).value(), input_shape.dim(rank).value());
+
+    LUCI_ASSERT(dim == diagonal_shape.dim(rank - 1), "Max diag len error");
```
binarman, line 1042: I am not sure about our conventions, maybe @seanshpark can tell more, but for me this looks like an overkill =)

I propose to move checks in separate function that returns `true` if everything is ok.
Leave one line, like `LUCI_ASSERT(check_function(input_shape, diagonal_shape), "Wrong diagonal shape")` in this function.

If we do so, this function will have only 3 lines, which is better for readability. (and could help avoid "Unused variable" compilation errors in case we redefine `LUCI_ASSERT`)

compiler/luci/lang/src/Nodes/CircleMatrixSetDiag.test.cpp
seanshpark: Yes, just to be sure visitor is working for this node :)

compiler/luci/service/src/CircleShapeInferenceRule.cpp
seanshpark: I think it's both is OK, add assert for validations and remove assert for readability.
Just for information, some strange real models do have some violations and these asserts help to find them :)

seanshpark:
Looks good to me :)

binarman:
Everything is merged, thank you!

Issue #2021, opened 2020-06-08
 Compiler FE: Enable Where Operator  


d-krylov:
- [x] tflchef #2330 
- [x] luci.IR #2545 
- [x] luci.Import  #2545 
- [x] luci.Export   #2545 
- [x] luci.logex  #2545 
- [x] luci.Service   #2545 
- [x] res/TensorFlowPythonExamples 
- [x] res/TensorFlowLiteRecipes  #2332 #2741 
- [x] tflite2circle  #2331 
- [x] luci.tests  #2545 
- [x] tfldump  nothing to do
- [x] circledump  nothing to do

for #1606

draft 1 #2060 
draft 2 #2741

binarman:
Everything is done, closing

PR #2740, opened 2020-06-30
[DRAFT] Compiler FE: Enable Where operator version 2

binarman:
Support in luci + recipes


binarman:
second attempt for #2021

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -2011,6 +2011,13 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
     return loco::NodeShape{output_shape};
   }
 
+  loco::NodeShape visit(const luci::CircleWhere *node) final
+  {
+    // return empty shape, because it is known only on runtime
```
seanshpark, line 2016: For this case, please use
```
loco::TensorShape shape = own_shape(node);
return loco::NodeShape{shape};
```

res/TensorFlowLiteRecipes/Where_000/test.recipe
```diff
@@ -1,33 +1,19 @@
...
-  type: FLOAT32
-  shape { dim: 4 dim: 4 dim: 4 dim: 3 }
+  shape { dim: 4 dim: 2 dim: 3}
 }
 operand {
   name: "ofm"
-  type: FLOAT32
-  shape { dim: 4 dim: 4 dim: 4 dim: 3 }
+  type: INT64
+  shape { } # output shape is 2d, but partially unknown, because first dimension depends on number of "true" values in input so in our case it is (?, 3)
```
seanshpark, line 9: Can you please split this comment so that length would be in 80 or 100 columns ? Just for better readability :)

seanshpark:
Looks good! (with one minor comment)

res/TensorFlowLiteRecipes/Where_000/test.recipe
binarman: Fixed, thank you.

P.S. I also changed comment in second recipe to "shape is intentionally omitted", because I found that is the comment used in other recipes =)

seanshpark:
👍

binarman:
Everything is merged.
Thank you!

PR #2742, opened 2020-07-01
[cmake] Introduce TensorFlowSource-2.3.0-rc0Config

seanshpark:
This will introduce TensorFlowSource-2.3.0-rc0Config.cmake file to aquire 2.3.0-rc0 with `nnas_find_package()` method


seanshpark:
For #2722, from draft #2724

PR #2737, opened 2020-06-30
[Draft] Support explicit expression of optional input tensor

llFreetimell:
Parent Issue : #2734 

Until now, there is no way to express optional input tensor explicitly.
This draft is for supporting explicit expression of optional input tensor.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

Issue #2734, opened 2020-06-30
[Compiler] Support explicit expression of optional input tensor in chef


llFreetimell:
### Background
We sometimes want to test when optional input tensor is not exist.
For example, `FullyConnected` in TFLite can have no `bias` and it can be **implicitly** expressed as follow.
```
operand {
  name: "in"
  type: FLOAT32
  shape { dim: 1 dim: 1 dim: 1 dim: 4 }
}
operand {
  name: "weight"
  type: FLOAT32
  shape { dim: 2 dim: 4 }
}
operand {
  name: "out"
  type: FLOAT32
  shape { dim: 1 dim: 2 }
}
operation {
  type: "FullyConnected"
  fullyconnected_options {
    activation: NONE
  }
  input: "in"
  input: "weight"
  output: "out"
}
input: "in"
input: "weight"
output: "out"
```

### Problem

However, how should we handle for the case when the optional input tensor is not in the last dimension?
Because `tflchef` and `circlechef` have no **explicit** expression method for optional input tensor, there is no way for it.
This problem is occurred when I was handling #1517.

### Solution

When we give blank("") as input, we can interpret this as empty optional input tensor.
```
operand {
  name: "in"
  type: FLOAT32
  shape { dim: 1 dim: 1 dim: 1 dim: 4 }
}
operand {
  name: "weight"
  type: FLOAT32
  shape { dim: 2 dim: 4 }
}
operand {
  name: "out"
  type: FLOAT32
  shape { dim: 1 dim: 2 }
}
operation {
  type: "FullyConnected"
  fullyconnected_options {
    activation: NONE
  }
  input: "in"
  input: "weight"
  input: ""  <---------------- Here is suggestion
  output: "out"
}
input: "in"
input: "weight"
output: "out"
```

### Draft

- #2737

llFreetimell:
All related PRs are merged, done!

PR #2744, opened 2020-07-01
[tflchef] Support optional input tensor expression

llFreetimell:
Parent Issue : #2734
Full Draft : #2737

Until now, there is no way to express optional input tensor explicitly in `tflchef`.
This commit will make it possible.

### DIFF
--- a/compiler/tflchef/core/src/ModelChef.cpp
+++ b/compiler/tflchef/core/src/ModelChef.cpp
@@ -347,7 +347,14 @@ GeneratedModel cook(const ::tflchef::ModelRecipe &model_recipe)
     // Tensor Name -> Tensor ID mapping (per Graph)
     std::map<std::string, int32_t> symbol_table;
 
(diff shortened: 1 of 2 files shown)


compiler/tflchef/core/src/ModelChef.cpp
```diff
@@ -593,7 +600,14 @@ GeneratedModel cook(const ::tflchef::ModelRecipe &model_recipe)
     // Tensor Name -> Tensor ID mapping (per Graph)
     std::map<std::string, int32_t> symbol_table;
 
-    auto lookup = [&symbol_table](const std::string &name) { return symbol_table.at(name); };
+    auto lookup = [&symbol_table](const std::string &name) {
+      if (symbol_table.find(name) != symbol_table.end())
+        return symbol_table.at(name);
+      else if (name == "")
+        return -1;
```
seanshpark, line 607: (optional) instead of using `-1`, maybe use `INPUT_EMPTY` or something?

PR #2748, opened 2020-07-01
[onert] Release constant tensor when ruy kernel runs on cpu

YongseopKim:
Release constant tensor(weight) after ruy kernel runs on cpu if possible
- if batch size <= 4, ruy kernel caches the weight so that we can
release the weight.

### DIFF
--- a/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
+++ b/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
@@ -111,7 +111,7 @@ void FullyConnectedLayer::fullyConnectedHybrid()
       getTensorShape(_bias), reinterpret_cast<const float *>(_bias ? _bias->buffer() : nullptr),
       getTensorShape(_output), reinterpret_cast<float *>(_output->buffer()), temp_arena);
 
-  if (_cached_weights == nullptr || _freed_weights)
+  if (_cached_weights == nullptr || _is_weights_freed)
     return;
 
   auto weight_tensor = dynamic_cast<const Tensor *>(_weights);
@@ -121,7 +121,7 @@ void FullyConnectedLayer::fullyConnectedHybrid()
     tensor->decrease_ref();
     if (tensor->buffer() == nullptr) // ref == 0?
     {
-      _freed_weights = true;
+      _is_weights_freed = true;
     }
   }
 #endif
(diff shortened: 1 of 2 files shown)


runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
```diff
@@ -100,6 +101,44 @@ void FullyConnectedLayer::fullyConnectedHybrid()
...
+
+  // TODO Remove const_cast
+  auto weight_tensor = dynamic_cast<const Tensor *>(_weights);
+  if (weight_tensor)
+  {
+    // TODO Remove this workaround
+    // buffer will have deallocated memory address because ruy kernel uses
+    // it as cache key
+    auto tensor = const_cast<Tensor *>(weight_tensor);
+    auto buffer = tensor->buffer();
```
wateret, line 133: To remove the workaround, can we make `buffer` to be a member of `FullyConnectedLayer`?

YongseopKim: Oh, that's a good idea. I'll try. Thanks.

YongseopKim:
PTAL @wateret and @periannath

runtime/onert/backend/cpu/ops/FullyConnectedLayer.h
```diff
@@ -65,6 +67,11 @@ class FullyConnectedLayer : public ::onert::exec::IFunction
 
   ir::Activation _activation;
   std::unique_ptr<nnfw::cker::FCTempArena> _temp_arena;
+
+#ifdef USE_RUY_GEMV
+  bool _freed_weights = false;        // weights is already freed?
```
periannath, line 72: (Optional) I think `_weights_freed` or `_is_weights_freed` would be better.

YongseopKim: `_is_weights_freed` looks better, thanks.

runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
```diff
@@ -95,11 +96,35 @@ void FullyConnectedLayer::fullyConnectedHybrid()
...
+      getTensorShape(_output), reinterpret_cast<float *>(_output->buffer()), temp_arena);
+
+  if (_cached_weights == nullptr || _is_weights_freed)
+    return;
+
+  auto weight_tensor = dynamic_cast<const Tensor *>(_weights);
+  if (weight_tensor)
+  {
+    auto tensor = const_cast<Tensor *>(weight_tensor);
+    tensor->decrease_ref();
```
wateret, line 121: Ah, still need it as non-const due to `decrease_ref`! 😢

Issue #2671, opened 2020-06-26
[onert] Scalar/Vector tensor and OperationValidator check


dayo09:
I have been confused about how to handle scalar and vector input tensor.
This issue is to make some points, please tell me if something's wrong.

### Scalar(0-dim) Vector(1-dim) input tensors
- some operations have input tensors of 0-dim or 1-dim.
- 0-dim tensor : loaded as a param, usually int or float. Or loaded as a Tensor
- 1-dim tensor :  loaded as a vector, usually vector<int>. Or loaded as a Tensor 

### Pros and Cons of loading them as param / as tensor
(please tell me if I know wrongly)
#### as tensor
- different data type can be supported
#### as param 
- runtime's tensor allocation gets less load
- param values can be used while OperationValidator check (otherwise ConstnatnInit runs after OperationValidator so it cannot be checked)

dayo09:
## Operator Validation during runtime
- Compiler::compile() runs OperationValidator and then CreateExecutor- which contains ConstInit or so.
- So actually some of operation-validating is done during runtime
- (suggestion) Why don't we have OP_REQUIRES-like macro all over the cpu kernels? Will be useful.

hyunsik-yoon:
_param_ is decided at compilation time, meaning param value is in a model file. (e.g., const..)
_tensor_ can be read at compilation (when it is a const) time but also also be read at runtime (output of op).
In other words, in pbtxt file, there are some attributes, e.g., `strides` is always provided by model file and it is also encoded into tflite file, so that is treated as `param`. However `ker` is _input_, which could be a scalar, vector, or 2d tensor. Regardless of its rank, _input_ needs to be treated as _tensor_ rather than _param_.

```
node {
  name: "ofm"
  op: "Conv2D"
  input: "ifm"
  input: "ker"
  attr {
    key: "T"
    value { type: DT_FLOAT }
  }
  attr {
    key: "data_format"
    value { s: "NHWC" }
  }
  attr {
    key: "dilations"
    value {
      list { i: 1 i: 1 i: 1 i: 1 }
    }
  }
  attr {
    key: "padding"
    value { s: "VALID" }
  }
  attr {
    key: "strides"
    value {
      list { i: 1 i: 1 i: 1 i: 1 }
    }
  }
}

```

dayo09:
@hyunsik-yoon Thank you for the explanation between Tensor and Param. However I meant somewhat differnet. I think my explanation was vague. 
[...]

Issue #2431, opened 2020-06-19
[onert] Determining on/off values for one-hot quant8


dr-venkman:
The data types of `on_value` and `off_value` within `one-hot` are defined as `float` within onert. While this works for float32, how should it be handled for the case of quant8 asymm? There are two related questions I would like to know about, which are as follows:

1. Tfllite models the above values as tensors, and so can interpret them according to the type of computation (`float` or `uint8_t`). As onert defines them to be strictly `float`, should we typecast and clamp the values within (0, 255) range? 

For example, if we consider `float on_value = 300; float off_value = 0;` will they be converted to `uint8_t q_on_value = 255; uint8_t q_off_value = 0;`?

2. Will there be scale, offset parameters to consider for `on_value` and `off_value`? If yes, then, we will need additional parameters to be added under `runtime/onert/core/include/ir/operation/OneHot.h`.

Please let me know what approach works best.

dayo09:
```c++
// ir/operation/onehot.h
class OneHot : public Operation
{
public:
  enum Input
  {
    INDICES = 0,
    DEPTH = 1,
    ON_VALUE = 2,
    OFF_VALUE = 3,
  };

  struct Param
  {
    int depth;       // comes from input tensor, not from OneHotOptions
    float on_value;  // comes from input tensor, not from OneHotOptions
    float off_value; // comes from input tensor, not from OneHotOptions
    int axis;
  };
```

- It seems that IR already has ON_VALUE and OFF_VALUE as Input Tensor.
- At base_loader.h, it converts ON_VALUE and OFF_VALUE from Input into Param.
- Therefore, you don't need to change IR, so there is no following concerns about other backends. Just changing base_loader will do. If it passes as Input Tensor instead of converting into param, I think it will work.

dr-venkman:
Thanks for the explanation, I will get started with the implementation following this. Regarding the second question (about scale and offsets), I see that tflite does not convert between on/off tensor scales and output scales, so I can safely ignore that concern.

glistening:
@dr-venkman I've just read this issue. I started to update `OneHot` yesterday. I will push it today or tomorrow.

dr-venkman:
@glistening Thanks, I have just about completed quant8 support for comparison operators from last week. Please go ahead with one-hot, I will find other operations to work on shortly.

hseok-oh:
#2759 is merged

PR #2759, opened 2020-07-01
[onert] OneHot will store depth and on/off value in input tensor

glistening:
OneHot had stored depth and on/off value in param.
It works for constant float input.
However, it need to be updated to cover those cases:

- other data types like int32, quantized 8bit
- depth and on/off values is determined at runtime

Thus, it will save depth and on/off value as input tensor like tflite does.

It includes the change in NeuralNetworksEx.h.

### DIFF
--- a/tests/nnapi/specs/Ex/one_hot_ex_dynamic_nnfw.mod.py
+++ b/tests/nnapi/specs/Ex/one_hot_ex_dynamic_nnfw.mod.py
@@ -27,9 +27,10 @@ model = Model()
 indice_shape = [2, 2]
 indice_input = [1, 2, 0, 2]
 
-depth = Int32Scalar("depth", 3)
-onvalue = Float32Scalar("on", 1.) # default value is 1.
-offvalue = Float32Scalar("off", 0.) # default value is 0.
+depth = Input("depth", "TENSOR_INT32", "{1}")
+onvalue = Input("onvalue", "TENSOR_FLOAT32", "{1}")
+offvalue = Input("offvalue", "TENSOR_FLOAT32", "{1}")
+
 axis0 = Int32Scalar("axis", -1) # default value is -1.
 model_output0 = Output("output", "TENSOR_FLOAT32", "{2, 2, 3}")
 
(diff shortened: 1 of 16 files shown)


glistening:
Related: #2431, #2671 

For a model below, which can know `off_value` at runtime,

image

#### Before
```
$ BACKENDS=cpu Product/armv7l-linux.release/out/bin/nnpackage_run \
--nnpackage onehot_offvalue_from_input
Segmentation fault
```

#### After
```
$ BACKENDS=cpu tests/scripts/nnpkg_test.sh onehot_offvalue_from_input
[  Run  ] ./onehot_offvalue_from_input 	Pass
[Compare] ./onehot_offvalue_from_input 	Pass
```

[onehot_offvalue_from_input.zip](https://github.com/Samsung/ONE/files/4855987/onehot_offvalue_from_input.zip)

I've just found `acl_neon` value test fails. I will fix it.

glistening:
@dayo09, @hseok-oh  I've confirmed `acl_neon` is fixed. PTAL.

tests/nnapi/specs/Ex/one_hot_ex_dynamic_nnfw.mod.py
```diff
@@ -27,9 +27,10 @@
 indice_shape = [2, 2]
 indice_input = [1, 2, 0, 2]
 
-depth = Int32Scalar("depth", 3)
-onvalue = Float32Scalar("on", 1.) # default value is 1.
-offvalue = Float32Scalar("off", 0.) # default value is 0.
+depth = Input("depth", "TENSOR_INT32", "{1}")
+onvalue = Input("onvalue", "TENSOR_FLOAT32", "{1}")
+offvalue = Input("offvalue", "TENSOR_FLOAT32", "{1}")
```
periannath, line 32: @glistening It seems the off value is wrong. I think it should be 0.

glistening: @periannath `{1}` is not the value of `offvalue`. It is the shape of `offvalue` tensor. You can find the value of `offvalue` in `offvalue_data`.

periannath: @glistening Thanks, I didn't know that.

PR #2762, opened 2020-07-01
[res] Remove temporary TF lite schema

seanshpark:
This will remove temporary TensorFlow lite schema before 2.3.0-rc0 release


seanshpark:
For #2722, from draft #2724

PR #2767, opened 2020-07-01
[record-minmax-conversion-test] Add FullyConnected_002 in test list

llFreetimell:
Parent Issue : #2734
Full Draft : #2737

This commit will add `FullyConnected_002` in the test list,
because thie test could not be enable for now.


jinevening:
Why do you add the commented test? Are you going to enable it later?

llFreetimell:
@jinevening Is this list only contains passed tests? I added `FullyConnected_002` only because `test.lst` contains `FullyConnected` operation. If I was wrong, I will close this PR :)

jinevening:
> Is this list only contains passed tests?

Yes, `test.lst` only has the whitelist. If #2734 is not related to `record-minmax`, it would be better to close this PR :) Thanks.

llFreetimell:
As the test list only includes passed tests, I will close this PR!

PR #2763, opened 2020-07-01
DRAFT Compiler FE introduce souschef

seanshpark:
on-going draft of souschef; common codes for circlechef and tflchef

ONE-DCO-1.0-ONE-DCO-1.0-ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #2753, opened 2020-07-01
Compiler FE: reduce duplicate codes


seanshpark:
There are lots of duplicate codes in our modules for '20 compiler frontend.
Lets' reduce them.

- [x] extract common codes in _circlechef_ and _tflchef_ to _souschef_ (new module)
- [x] similar patterns in _logo_ ReorderDecodePass.cpp
- [x] use foder model loaders (Model.cpp)
- [x] quantization in pass

seanshpark:
close for now

PR #2768, opened 2020-07-01
[souschef] Introduce souschef

seanshpark:
This will introduce souschef module from common codes of tflchef and circlechef
- use namespace souschef
- code itself is not changed from tflchef/circlechef
- adjusted header header file guards


seanshpark:
For #2753, from draft #2763

compiler/souschef/include/souschef/Data/Constant.h
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __SOURCHEF_DATA_CONSTANT_H__
+#define __SOURCHEF_DATA_CONSTANT_H__
```
mhs4670go, line 18: maybe souschef?

compiler/souschef/include/souschef/DataChef.def
```diff
@@ -0,0 +1,19 @@
...
+
+// DATA_CHEF(TYPE, NAME, FACTORY_CLASS)
+//  "TYPE" SHOULD BE an enum tag of tflchef::TensorType
+DATA_CHEF(FLOAT32, constant, ConstantDataChefFactory<float>)
+DATA_CHEF(BOOL, constant, ConstantDataChefFactory<bool>)
+DATA_CHEF(UINT8, constant, ConstantDataChefFactory<uint8_t>)
+DATA_CHEF(INT32, constant, ConstantDataChefFactory<int32_t>)
+DATA_CHEF(INT64, constant, ConstantDataChefFactory<int64_t>)
+DATA_CHEF(INT64, explicit, ExplicitDataChefFactory<int64_t>)
+DATA_CHEF(INT32, explicit, ExplicitDataChefFactory<int>)
```
llFreetimell, line 13: Just curious, compared with `DATA_CHEF(INT32, constant, ConstantDataChefFactory<int32_t>)`, is there any reason for using `int` instead of `int32_t`?

seanshpark: Hum... not sure, it was copied from existing codes. I thin we should use int32_t for this

PR #2773, opened 2020-07-01
[infra] Add souschef to coverage

seanshpark:
This will add souschef module to code coverage report script


seanshpark:
For #2753, from draft #2763

PR #2775, opened 2020-07-01
[infra] Add souschef to preset 20200630

seanshpark:
This will add souschef module to preset 20200630
- adding new preset requires chain updates into CI; update this file for a while


seanshpark:
For #2753, from draft #2763

PR #2783, opened 2020-07-01
[tflchef] Use souschef to remove common codes

seanshpark:
This will revise tflchef to use souschef to remove common codes with circlechef


seanshpark:
For #2753, from draft #2763

seanshpark:

Issue #2761, opened 2020-07-01
[gen_golden] error on float input


glistening:
`gen_golden` fails for tflite in [onehot_offvalue_from_input.zip](https://github.com/Samsung/ONE/files/4856084/onehot_offvalue_from_input.zip)

```
root@white:/opt/z/ONE# tools/nnpackage_tool/gen_golden/gen_golden.py onehot_offvalue_from_input.tflite
Traceback (most recent call last):
  File "tools/nnpackage_tool/gen_golden/gen_golden.py", line 150, in <module>
    interpreter.set_tensor(input_details[idx]['index'], input_values[idx])
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/lite/python/interpreter.py", line 404, in set_tensor
    self._interpreter.SetTensor(tensor_index, value)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/lite/python/interpreter_wrapper/tensorflow_wrap_interpreter_wrapper.py", line 149, in SetTensor
    return _tensorflow_wrap_interpreter_wrapper.InterpreterWrapper_SetTensor(self, i, value)
ValueError: Cannot set tensor: Got value of type NOTYPE but expected type FLOAT32 for input 3, name: decoder/one_hot/off_value 
```

If I roll back #2669, or explicitly pass `np.float32` to `np.array`,

```diff
$ git diff
diff --git a/tools/nnpackage_tool/gen_golden/gen_golden.py b/tools/nnpackage_tool/gen_golden/gen_golden.py
index d78e0a8..4373d8f 100755
--- a/tools/nnpackage_tool/gen_golden/gen_golden.py
+++ b/tools/nnpackage_tool/gen_golden/gen_golden.py
@@ -136,7 +136,7 @@ if __name__ == '__main__':
                 input_values.append(
                     np.array(np.random.randint(0, 255, this_shape, this_dtype)))
             elif this_dtype == np.float32:
-                input_values.append(np.array(np.random.random_sample(this_shape)))
+                input_values.append(np.array(np.random.random_sample(this_shape), np.float32))
             elif this_dtype == np.bool_:
                 # generate random integer from [0, 2)
                 input_values.append(
```

it is solved.

I am using `numpy 1.18.4` and `tf 2.2.0`.

cc @chunseoklee, @llFreetimell

[...]

PR #1330, opened 2020-05-22
[nnpackage_tool] Adding supported type for pb

llFreetimell:
_Continued from #1317_

Until now, only float random input values are generated for `.pb`.
This commit will add supported type for `.pb`

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

PR #2785, opened 2020-07-02
[gen_golden] cast to float32 for float32 input

glistening:
As np.random.random_sample() generates float64 type array,
it it necessary to cast to `float32`.

### DIFF
--- a/tools/nnpackage_tool/gen_golden/gen_golden.py
+++ b/tools/nnpackage_tool/gen_golden/gen_golden.py
@@ -92,7 +92,8 @@ if __name__ == '__main__':
                 input_values.append(
                     np.array(np.random.randint(0, 255, this_shape, this_dtype)))
             elif this_dtype == np.float32:
-                input_values.append(np.array(np.random.random_sample(this_shape)))
+                input_values.append(
+                    np.random.random_sample(this_shape).astype(np.float32))
             elif this_dtype == np.bool_:
                 # generate random integer from [0, 2)
                 input_values.append(
@@ -136,7 +137,8 @@ if __name__ == '__main__':
                 input_values.append(
                     np.array(np.random.randint(0, 255, this_shape, this_dtype)))
             elif this_dtype == np.float32:
-                input_values.append(np.array(np.random.random_sample(this_shape)))
+                input_values.append(
+                    np.random.random_sample(this_shape).astype(np.float32))
             elif this_dtype == np.bool_:
                 # generate random integer from [0, 2)
                 input_values.append(


glistening:
Related #2761 

@chunseoklee or @llFreetimell Could you please check it works on your environment with test case which made you push #2669.

Currently #2669 shows error on `float` input at least for me and other member.

chunseoklee:
@glistening failed with tf2.3-rc0 and numpy 1.18.5
``` sh
(tf-2.3-rc0) ~/d/O/t/n/gen_golden> python ./gen_golden.py -o out_dir ~/dev/models/Bixby/NLU/eNLU_0619_contextual_bert_matmul/eNLU_0619_contextual_bert_matmul_frozen.pb
2020-07-02 18:56:48.751870: W tensorflow/stream_executor/platform/default/dso_loader.cc:59] Could not load dynamic library 'libcudart.so.10.1'; dlerror: libcudart.so.10.1: cannot open shared object file: No such file or directory
2020-07-02 18:56:48.751896: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
  File "./gen_golden.py", line 96, in <module>
    np.random.random_sample(this_shape).astype(this_dtype))
TypeError: data type not understood
(tf-2.3-rc0) ~/d/O/t/n/gen_golden> pip list | grep numpy
numpy                  1.18.5
```

glistening:
@chunseoklee I've tested with your environment using docker. It worked for me because the difference is that you're testing with `.pb`, and I am using `.tflite`. For `.tflite`, `this_dytpe` is numpy type. Thus, it is okay. For `.pb`, `this_dtype` is non-numpy type. It is returned type by tensorflow API though it is comparable with `numpy.dtype`. I fixed by using explicit numpy dtype (`np.float32`) for both case.

I am not sure that `gen_golden.py` would work for `.pb` for other data type that was introduced by #1330.
Because it seems to assume `this_dtype` is `numpy type` like `.tflite`, and `float` doesn't work.

@llFreetimell Did you tested for various inputs of `pb`? If not, we need to check and confirm the change of #1330.

llFreetimell:
This PR works for me and my `numpyt` version is 1.18.1.

> @llFreetimell Did you tested for various inputs of pb? If not, we need to check and confirm the change of #1330.

Before #1330 is merged, only `float32` type was supported by `gen_golden.py` but I needed to generate for `int32`.
So I just copied the code from the case when `elif extension == ".tflite":` in `gen_golden.py`.
And models what I have seen usually have `float32` and `int32` inputs and I was able to check only for those cases.

glistening:
@llFreetimell Okay. `float` and `int32` are checked. For other types, it need to be checked. I am curious how it would work for `bool` and `uint8`. I will test them.

glistening:
@chunseoklee Could you please check this PR works? After this is merged, I will push another PR after checking other types.

glistening:
@llFreetimell PTAL

PR #2789, opened 2020-07-02
[tflchef] Revise cook to remove dup codes

seanshpark:
This will revise cook method to remove duplicate codes to produce main and subgraphs


seanshpark:
For #2753, from draft #2784

seanshpark:

PR #2791, opened 2020-07-02
[onert] Add override keyword for FullyConnectedLayer method

hseok-oh:
Override IFunction's run() method


hseok-oh:
It will fix android build fail.

PR #2793, opened 2020-07-02
[logo] Common code for ReorderDecodePass

seanshpark:
This will add common code in ReorderDecodePass to reduce duplicate codes


seanshpark:
For #2753, from draft #2784

seanshpark:

Issue #2769, opened 2020-07-01
[onert] Running nnpackage_run with verbose log produces segfault on android


periannath:
I recently tried to run `nnpackage_run` on android with verbose log for debugging.
Running `nnpackage_run` with `ONERT_LOG_ENABLE=1` produces segmentation fault for every model.

#### Execution result
```bash
$ ONERT_LOG_ENABLE=1 BACKENDS=cpu ./nnfw_alpha/nnpackage_run --nnpackage mobilenet_v2_1.0_224/                                                                                            
[Compiler] [Compiler] ==== Compiler Options ====
[Compiler] backend_list             : cpu
[Compiler] trace_filepath           : 
[Compiler] graph_dump_level         : 0
[Compiler] op_seq_max_node          : 0
[Compiler] executor                 : Linear
[Compiler] manual_scheduler_options : (Too many things to print)
[Compiler] he_scheduler             : false
[Compiler] he_profiling_mode        : false
[Compiler] disable_compile          : false
[Compiler] fp16_enable              : false
[Compiler] [loadBackend] Successfully loaded 'cpu' - libbackend_cpu.so
Segmentation fault (core dumped) 
```
- Segmentation fault is occured

#### Generated crash dump file
- Crash dump file is created at `/data/tombstones/`
- The ndk-stack tool allows you to symbolize stack traces from adb logcat or a tombstone
  - https://developer.android.com/ndk/guides/ndk-stack

```bash
$ adb shell cat /data/tombstones/tombstone_13 | ./ndk-stack -sym ~/ONE/Product/out/lib
********** Crash dump: **********
Build fingerprint: 'samsung/y2qsqw/y2q:10/QP1A.190711.020/G986USQE1ATDA:eng/test-keys'
#00 0x0000000000056d54 /data/local/tmp/nnfw_alpha/nnpackage_run (std::__ndk1::basic_ostream<char, std::__ndk1::char_traits<char>>::sentry::sentry(std::__ndk1::basic_ostream<char, std::__ndk1::char_traits<char>>&)+60) (BuildId: 53cd438c0bccf6ad550e4c14532de31724e1d0c9)
                                                                 std::__ndk1::basic_ostream<char, std::__ndk1::char_traits<char> >::sentry::sentry(std::__ndk1::basic_ostream<char, std::__ndk1::char_traits<char> >&)
[...]

PR #2794, opened 2020-07-02
Use shared library variant of libc++ for android build

periannath:
- This commit uses shared library variant of libc++ for android build
  - nnpackage_run with verbose log produces segfault on android
  - It seems linking static c++ runtime for multiple libraries produces this error
  - Using shared c++ runtime fixes this issue
- `libc++_shared.so` file in ndk is installed in lib folder for android build

---

Realted issue : #2769

### DIFF
--- a/infra/nnfw/cmake/buildtool/cross/toolchain_aarch64-android.cmake
+++ b/infra/nnfw/cmake/buildtool/cross/toolchain_aarch64-android.cmake
@@ -13,6 +13,8 @@ endif(NOT DEFINED NDK_DIR)
 set(ANDROID_ABI arm64-v8a)
 set(ANDROID_API_LEVEL 29)
 set(ANDROID_PLATFORM android-${ANDROID_API_LEVEL})
+set(ANDROID_STL c++_shared)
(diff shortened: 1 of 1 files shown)


periannath:
This PR comes from https://github.com/Samsung/ONE/issues/2769#issuecomment-652731540. Thanks to @YongseopKim

YongseopKim:
@periannath How about adding below?
```
set(ANDROID_STL_LIB ${NDK_DIR}/sources/cxx-stl/llvm-libc++/libs/arm64-v8a/libc++_shared.so)
install(FILES ${ANDROID_STL_LIB} DESTINATION lib)
```

Then,
```
dragon@loki:~/Works/github/ONE$ ls ./Product/out/lib/libc++_shared.so 
./Product/out/lib/libc++_shared.so
```

periannath:
@YongseopKim Thanks! It would be better. I've updated.

periannath:
@dhdh-oh I've updated. PTAL

hseok-oh:
I'll merge without @dhdh-oh 's approve because it delayed 4 days.

PR #2798, opened 2020-07-02
[onert] Fix getReducerAxes bug

hseok-oh:
getReducerAxes return invalid vector because of pointer operation bug


hseok-oh:
Bug in #2747
It will fix test fail on debug build.

Issue #2730, opened 2020-06-30
[luci-interpreter] Test quantized kernels for ResNet-v1


struss:
Parent issue: #1157 , #1612 
Draft : #2720 

Update Testcase for Operator
- [x] Add #2728 
- [x] ArgMax #2731 
- [x] Conv2D #2868 
- [x] MaxPool2D #2801
- [x] Mean #2800 
- [x] Pad #2795

struss:
All Uint8 Test case and shape inference merged! Done! Thanks to all that help this!

PR #2800, opened 2020-07-02
[luci-interpreter] Test quantized Mean kernels for ResNet-v1

struss:
Parent Issue: #2730 
Draft PR: #2720

This commit update `Mean` operator test case.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Mean.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Mean.test.cpp
@@ -27,6 +27,9 @@ namespace
 
 using namespace testing;
 
+// for quantized Add, the error shouldn't exceed step
+float GetTolerance(int min, int max) { return (max - min) / 255.0; }
+
 TEST(MeanTest, FloatKeepDims)
 {
   std::vector<float> input_data = {1.0,  2.0,  3.0,  4.0,  5.0,  6.0,  7.0,  8.0,
@@ -46,8 +49,10 @@ TEST(MeanTest, FloatKeepDims)
   kernel.execute();
 
   std::vector<float> ref_output_data{10.5, 12.5, 14.5};
+  std::initializer_list<int32_t> ref_output_shape{1, 3, 1};
   EXPECT_THAT(extractTensorData<float>(output_tensor),
               ElementsAreArray(ArrayFloatNear(ref_output_data)));
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray(ref_output_shape));
 }
 
 TEST(MeanTest, FloatKeepDims4DMean)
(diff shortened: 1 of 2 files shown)


compiler/luci-interpreter/src/kernels/Mean.test.cpp
```diff
@@ -92,8 +99,70 @@ TEST(MeanTest, FloatNotKeepDims)
...
+}
+
+TEST(MeanTest, Uint8NotKeepDims)
+{
+  float kQuantizedTolerance = GetTolerance(-1.0, 1.0);
+  std::vector<float> input_data = {0.4, 0.2, 0.3, 0.4, 0.5, 0.6};
+  std::pair<float, int32_t> quant_param = quantizationParams<uint8_t>(-1.0f, 1.0f);
+
+  std::vector<int32_t> axis_data{1};
+  // Tensor input_tensor = makeInputTensor<DataType::FLOAT32>({3, 2}, input_data);
```
jinevening, line 145: Can you remove this code?

compiler/luci-interpreter/src/kernels/Mean.test.cpp
```diff
@@ -92,8 +99,70 @@ TEST(MeanTest, FloatNotKeepDims)
...
+}
+
+TEST(MeanTest, Uint8KeepDims)
+{
+  float kQuantizedTolerance = GetTolerance(-1.0, 1.0);
+  std::vector<float> input_data = {0.4, 0.2, 0.3, 0.4, 0.5, 0.6};
+  std::pair<float, int32_t> quant_param = quantizationParams<uint8_t>(-1.0f, 1.0f);
+
+  std::vector<int32_t> axis_data{1};
+  // Tensor input_tensor = makeInputTensor<DataType::FLOAT32>({3, 2}, input_data);
```
jinevening, line 115: ditto

compiler/luci-interpreter/src/kernels/Mean.test.cpp
struss: Okay. I will remove it!

compiler/luci-interpreter/src/kernels/Mean.test.cpp
```diff
@@ -27,6 +27,9 @@ namespace
 
 using namespace testing;
 
+// for quantized Add, the error shouldn't exceed step
+float GetTolerance(int min, int max) { return (max - min) / 255.0; }
```
jinevening, line 31: This is not just for quantized Add (see the comment).

Plus, min and max seem to be float, not int.

If GetTolerance is used in all quantized tests, it had better be placed in "kernels/TestUtils.h". Then, the function can have a more general form, e.g., `float getTolerance(float min, float max, int quantize_steps)`. `quantize_steps` will be 255 for uint8 quantization.

compiler/luci-interpreter/src/kernels/Mean.test.cpp
struss: Okay. I will remove it!

compiler/luci-interpreter/src/kernels/Mean.test.cpp
struss: Oh I will fix it!

PR #2801, opened 2020-07-02
[luci-interpreter] Test quantized MaxPool2D kernels for ResNet-v1

struss:
Parent Issue: #2730 
Draft PR: #2720

This commit update `MaxPool2D` operator test case.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/MaxPool2D.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/MaxPool2D.test.cpp
@@ -56,10 +56,10 @@ TEST(MaxPool2DTest, Float)
       1, 2, //
       5, 6, //
   };
-  std::initializer_list<int32_t> ref_output_shape{1,2,2,1};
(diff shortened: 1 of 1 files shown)


compiler/luci-interpreter/src/kernels/MaxPool2D.test.cpp
```diff
@@ -53,8 +53,43 @@ TEST(MaxPool2DTest, Float)
...
+  std::vector<uint8_t> quantize_input =
+      quantize<uint8_t>(input_data, quant_param.first, quant_param.second);
+  input_tensor.writeData(quantize_input.data(), quantize_input.size() * sizeof(uint8_t));
+
+  Pool2DParams params{};
+  params.padding = Padding::VALID;
+  params.filter_height = 2;
+  params.filter_width = 2;
+  params.stride_height = 2;
+  params.stride_width = 2;
```
binarman, line 80: optional: better to use different values for strides and filters, so we can catch HW mixing errors

PR #2802, opened 2020-07-02
[circlechef] Use class from souschef

seanshpark:
This will revise circlechef to use Registry and RangedArguments classes from souschef


seanshpark:
For #2753, from draft #2784

PR #2804, opened 2020-07-02
[logo] Revise ReorderDecodePass to reduce dup codes

seanshpark:
This will revise ReorderDecodePass to reduce duplicate codes


seanshpark:
For #2753, from draft #2784

PR #2813, opened 2020-07-02
[Draft][Onert] Add util for onert exception

KimDongEon:
Add a util to classify exception handling.

existing
throw std::runtime_error("")
You can classify and use exception handling that was used only in one way.

ex)
throw onert::util::exception::NotSupportedTypeException("");

Signed-off-by: Dongeon kim <good.kim@samsung.com>

### DIFF
--- a/runtime/onert/core/include/util/OnertException.h
+++ b/runtime/onert/core/include/util/OnertException.h
@@ -23,8 +23,6 @@ namespace onert
 {
 namespace util
 {
-namespace exception
-{
 
 class OnertException : public std::exception
 {
@@ -66,7 +64,6 @@ public:
   NotImplementedException(std::string _m) : OnertException("NotImplementedException : " + _m) {}
 };
 
-} // namespace exception
 } // namespace util
 } // namespace onert
 
--- a/runtime/onert/core/include/util/OnertException.h
+++ b/runtime/onert/core/include/util/OnertException.h
@@ -34,10 +34,10 @@ public:
   virtual const char *what() const throw() { return message.c_str(); }
 };
 
-class ExecutionFailedException : public OnertException
+class OutOfRangeException : public OnertException
 {
 public:
-  ExecutionFailedException(std::string _m) : OnertException("ExecutionFailedException : " + _m) {}
+  OutOfRangeException(std::string _m) : OnertException("OutOfRangeException : " + _m) {}
 };
 
 class FileNotFoundException : public OnertException
@@ -46,22 +46,57 @@ public:
   FileNotFoundException(std::string _m) : OnertException("FileNotFoundException : " + _m) {}
 };
 
-class NotSupportedTypeException : public OnertException
-{
-public:
-  NotSupportedTypeException(std::string _m) : OnertException("NotSupportedTypeException : " + _m) {}
-};
-
 class InvalidValueException : public OnertException
 {
 public:
   InvalidValueException(std::string _m) : OnertException("InvalidValueException : " + _m) {}
 };
 
+class NotSupportedTypeException : public OnertException
+{
+public:
+  NotSupportedTypeException(std::string _m) : OnertException("NotSupportedTypeException : " + _m) {}
+};
+
+class NotYetSupportedTypeException : public OnertException
+{
+public:
+  NotYetSupportedTypeException(std::string _m)
+      : OnertException("NotYetSupportedTypeException : " + _m)
+  {
+  }
+};
+
+class NotSupportedOperationException : public OnertException
+{
+public:
+  NotSupportedOperationException(std::string _m)
+      : OnertException("NotSupportedOperationException : " + _m)
+  {
+  }
+};
+
+class NotYetSupportedOperationException : public OnertException
+{
+public:
+  NotYetSupportedOperationException(std::string _m)
+      : OnertException("NotYetSupportedOperationException : " + _m)
+  {
+  }
+};
+
 class NotImplementedException : public OnertException
 {
 public:
   NotImplementedException(std::string _m) : OnertException("NotImplementedException : " + _m) {}
+};
+
+class NotYetImplementedException : public OnertException
+{
+public:
+  NotYetImplementedException(std::string _m) : OnertException("NotYetImplementedException : " + _m)
+  {
+  }
 };
 
 } // namespace util
--- a/runtime/onert/core/include/util/OnertException.h
+++ b/runtime/onert/core/include/util/OnertException.h
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


runtime/onert/core/include/util/OnertException.h
```diff
@@ -0,0 +1,73 @@
...
+#ifndef __ONERT_UTIL_ONERTEXCEPTION_H__
+#define __ONERT_UTIL_ONERTEXCEPTION_H__
+
+#include <iostream>
+
+namespace onert
+{
+namespace util
+{
+namespace exception
```
wateret, line 26: These days I think we are trying to reduce namespace nesting. What about just `onert::util`? 

~~Even `onert` looks good to me, as they are exception classes.~~ Hmm.. but it breaks consistency of the namespace rule.

wateret: LGTM other than it.

runtime/onert/core/include/util/OnertException.h
```diff
@@ -0,0 +1,73 @@
...
+class InvalidValueException : public OnertException
+{
+public:
+  InvalidValueException(std::string _m) : OnertException("InvalidValueException : " + _m) {}
+};
+
+class NotImplementedException : public OnertException
+{
+public:
+  NotImplementedException(std::string _m) : OnertException("NotImplementedException : " + _m) {}
```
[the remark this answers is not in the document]
KimDongEon: I think... It is good to simplify the classification of exceptions to the minimum.

For this reason, for example,
Whether or not there is a support plan, it is the same that has not been implemented.
Additional classify there can be confusing for user trying to debug.

I think you can use '//' comments if needed.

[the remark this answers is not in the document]
KimDongEon: In addition, I think that even in the case of NotSupportedTypeException, it can be used equally whether the datatype is not supported or the shape is not supported.

runtime/onert/core/include/util/OnertException.h
KimDongEon: Remove namespace exception.

runtime/onert/core/include/util/OnertException.h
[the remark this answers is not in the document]
KimDongEon: I classify NotImplementedException -> NotYetImplementedException and NotYetSupportedException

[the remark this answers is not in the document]
hyunsik-yoon: I am not sure the difference between `NotYetImplementedException` and `NotYetSupportedException`.

runtime/onert/core/include/util/OnertException.h
```diff
@@ -0,0 +1,76 @@
...
+{
+  std::string message;
+
+public:
+  OnertException(std::string _m) : message(_m) {}
+
+  virtual const char *what() const throw() { return message.c_str(); }
+};
+
+class ExecutionFailedException : public OnertException
```
hyunsik-yoon, line 37: It seems like this is a super exception for all, e.g., when an op is not yet implemented, 
`ExecutionFailedException` can be thrown or also `NotYetImplementedException` can be thrown.
So we may need some guideline when to use `ExecutionFailedException` rather than `NotYetImplementedException`.

runtime/onert/core/include/util/OnertException.h
```diff
@@ -0,0 +1,76 @@
...
+  ExecutionFailedException(std::string _m) : OnertException("ExecutionFailedException : " + _m) {}
+};
+
+class FileNotFoundException : public OnertException
+{
+public:
+  FileNotFoundException(std::string _m) : OnertException("FileNotFoundException : " + _m) {}
+};
+
+class NotSupportedTypeException : public OnertException
```
hyunsik-yoon, line 49: case 1) Some datatype is not supported (e.g., int128 :-P , meaning out of spec)
case 2) but some is not yet supported (e.g., int32 support for some ops).
I wonder if this is this for case 1).
Should we throw `NotYetImplementedException` in case of case 2)?
It's somewhat confusing which one to throw...

runtime/onert/core/include/util/OnertException.h
KimDongEon: Rethinking.. ExecutionFailed seems to contain all other exceptions.

So I think you can delete ExecutionFailed.

hyunsik-yoon: Some questions I have are...

I wonder what exception i have to use while writing code:
1) when system fails to alloc memory
2) whey system fails to find a tensor in map
...

I thought that `ExecutionFailedException` fits for such cases. 

However, when an input value or shape is wrong, exception will fail. Does that mean `InvalidValueException` should inherit `ExecutionFailedException`?

FYI, due to exception design needs consideration of complex cases, _one_ compiler team decided to use only two types of exceptions. :-P UserError (named `OopsException` that needs to be reported to users) and InternalError (internal logic error). :-)

wateret: I think we need different kind of exceptions. As runtime is getting complicated it is not trivial to know which exception is thrown at which point. And we are going to handle the errors differently.

For example, 

Sometimes we need to report the error to the user.

https://github.com/Samsung/ONE/blob/6735e767658b604e91876160641537dc84a87575/runtime/onert/core/src/exec/ExecutorBase.cc#L263-L264

Then we could throw `InsufficientOutputSizeError` rather than `runtime_error` and it may be caught in `catch (InsufficientOutputSizeError e) { return NNFW_INSUFFICIENT_OUTPUT_SIZE; }` of `nnfw_run`. And in another part of execution it could throw something else.

% When I left a comment "ExecutionFailedException", I did not had many thoughts and it looks too ambiguous 😓

wateret:
As @hyunsik-yoon mentioned, it would be nicer to write doxygen comments along with the code. If it is too hard to design at once, I think it is also good to introduce one exception at once and evolve the whole exception system gradually.

KimDongEon:
@hyunsik-yoon @wateret 
Exceptions were further classified
It was classified into 9 types.

1. OutOfRangeException
2. FileNotFoundException
3. InvalidValueException
4. NotSupportedTypeException
5. NotYetSupportedTypeException
6. NotSupportedOperationException
7. NotYetSupportedOperationException
8. NotImplementedException
9. NotYetImplementedException

From 4 to 9, it was classified as Not and NotYet.
NotSupportedOperationException and NotImplementedException can be confused.
To add a description, use NotImplementedException when there is a part that is not implemented within the supported operation.

runtime/onert/core/include/util/OnertException.h
KimDongEon: I agree with the opinion that ExecutionFailedException is too ambiguous.
I think it is important to identify the detailed cause when executing ExecutionFail.

Therefore, ExecutionFailedException was removed and more detailed Exceptions were added.

runtime/onert/core/include/util/OnertException.h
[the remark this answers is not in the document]
KimDongEon: Add NotSupportedOperationException and NotYetSupportedOperationException 

NotImplementedException and NotYetImplementedException can be used when there is a part that is supported but not yet implemented. tell me your opinions.

runtime/onert/core/include/util/OnertException.h
KimDongEon: add NotSupportedTypeException and NotYetSupportedTypeException

wateret:
Thank you for you work. I would like ask some questions.

- When `OutOfRangeException` will be thrown? Is this something to do with `std::out_of_range` thrown by `std::map::at`?
- AFAIR onert does not do file I/O(it is all done from frontend) so `FileNotFoundException` is not needed
- What is `InvalidValueException`?
- Which exception will be thrown in the example from https://github.com/Samsung/ONE/pull/2813#discussion_r450190011 ?
- What are the differences between Not and NotYet?

PR #2816, opened 2020-07-02
[onert] Introduce sequence_index cache to optimize findOperation

periannath:
- This commit introduces sequence_index cache (`_seq_indexes`) to optimize OpSequences::findOperation
  - `_seq_indexes` saves `OperationIndex` to `OpSequenceIndex` mapping
  - `_seq_indexes` is maintained during insertion, search, and removal of operations
  - `findOperation` returns OpSequenceIndex in constant time if entry is exist in `_seq_indexes`

---

Related issue : #2655

### DIFF
--- a/runtime/onert/core/include/ir/OpSequence.h
+++ b/runtime/onert/core/include/ir/OpSequence.h
@@ -66,15 +66,14 @@ public:
 public:
   void remove(const OperationIndex &index);
 
+  bool exist(const OperationIndex &index) const;
+
 public:
   Layout getLayout() const { return _layout; }
 
 public:
   std::vector<OperationIndex>::const_iterator begin() const { return _operations.begin(); }
   std::vector<OperationIndex>::const_iterator end() const { return _operations.end(); }
-
-private:
-  bool exist(const OperationIndex &index) const;
 
 private:
   OperandIndexSequence _inputs;
--- a/runtime/onert/core/src/ir/OpSequences.cc
+++ b/runtime/onert/core/src/ir/OpSequences.cc
@@ -30,12 +30,45 @@ OpSequenceIndex OpSequences::emplace(const OperationIndex &index, Layout layout)
 {
   std::unique_ptr<OpSequence> op_seq = std::make_unique<OpSequence>(layout);
   op_seq->appendOperation(index);
-  return push(std::move(op_seq));
+  const OpSequenceIndex &seq_index = push(std::move(op_seq));
+  cacheSequenceIndex(seq_index, index);
+  return seq_index;
 }
 
 OpSequenceIndex OpSequences::emplace(std::unique_ptr<OpSequence> &&op_seq)
 {
-  return push(std::move(op_seq));
+  auto &operations = op_seq->operations();
+  const OpSequenceIndex &seq_index = push(std::move(op_seq));
+  for (const auto &op_idx : operations)
+  {
+    cacheSequenceIndex(seq_index, op_idx);
+  }
+  return seq_index;
+}
+
+void OpSequences::cacheSequenceIndex(const OpSequenceIndex &seq_index,
+                                     const OperationIndex &op_index) const
+{
+  _seq_indexes.emplace(op_index, seq_index);
+}
+
+OpSequenceIndex *OpSequences::findSequenceIndex(const OperationIndex &operation_index) const
+{
+  // If opration_index is cached, return sequence_index from cache
+  if (_seq_indexes.count(operation_index))
+  {
+    auto &op_seq_index = _seq_indexes.at(operation_index);
+    if (_objects.count(op_seq_index) && _objects.at(op_seq_index)->exist(operation_index))
+    {
+      return &op_seq_index;
+    }
+    else
+    {
+      _seq_indexes.erase(operation_index);
+      return nullptr;
+    }
+  }
+  return nullptr;
 }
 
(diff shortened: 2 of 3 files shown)


periannath:
This PR produces value mismatch for some tests.

```bash
17:37:06  not ok 24 - MODELS/inception_module
17:37:06  not ok 25 - MODELS/mobilenet
```

I'll find out and fix the problem.

periannath:
I found that check code for OpSequence removal was missing. I've updated to fix this problem.

periannath:
@glistening I've updated. PTAL

runtime/onert/core/src/ir/OpSequences.cc
```diff
@@ -30,12 +30,46 @@ OpSequenceIndex OpSequences::emplace(const OperationIndex &index, Layout layout)
...
+  if (_seq_indexes.count(operation_index))
+  {
+    auto &op_seq_index = _seq_indexes.at(operation_index);
+    if (_objects.count(op_seq_index) && _objects.at(op_seq_index)->exist(operation_index))
+    {
+      return &op_seq_index;
+    }
+    else
+    {
+      _seq_indexes.erase(operation_index);
```
glistening, line 68: @periannath What about making a function for erasing the cache slot like `cacheSequenceIndex`?

runtime/onert/core/include/ir/OpSequence.h
```diff
@@ -66,16 +66,15 @@ class OpSequence
 public:
   void remove(const OperationIndex &index);
 
+  bool exist(const OperationIndex &index) const;
+
```
glistening, line 70: @periannath It is not necessary to move `exist` to `public`.

runtime/onert/core/src/ir/OpSequences.cc
```diff
@@ -30,12 +30,46 @@ OpSequenceIndex OpSequences::emplace(const OperationIndex &index, Layout layout)
...
+    cacheSequenceIndex(seq_index, op_idx);
+  }
+  return seq_index;
+}
+
+void OpSequences::cacheSequenceIndex(const OpSequenceIndex &seq_index,
+                                     const OperationIndex &op_index) const
+{
+  _seq_indexes.insert(
+      std::pair<const OperationIndex &, const OpSequenceIndex &>(op_index, seq_index));
```
glistening, line 53: @periannath You may use `_seq_indexes.emplace(op_index, seq_index)`.

runtime/onert/core/src/ir/OpSequences.cc
glistening: @periannath As we talked offline, I would like to introduce nested class `Cache` with `add`, `remove`, `find`. Instead of `cacheSequenceIndex`, `_seq.indexes.erase(...)` and `findSequenceIndex`.

runtime/onert/core/include/ir/OpSequence.h
periannath: @glistening Thanks! I'll update it.

periannath:
@glistening I've updated following your 2nd and 3rd comments. PTAL

runtime/onert/core/src/ir/OpSequences.cc
periannath: @glistening I have thought about it, but I can't find a proper name for it.

runtime/onert/core/src/ir/OpSequences.cc
periannath: @glistening Thanks! I'll update it.

runtime/onert/core/include/ir/OpSequence.h
glistening: @periannath It's my mistake. It is `OpSequence`, not `OpSequences`.

wateret:
Thank you for the enhancement but I have a little concern. This will add a restriction to `OpSequence` that it cannot remove an operation(it has `remove` method) once it is added to `OpSequences`. For now it is OK but this may be affected when we do optimizations like operation fusing. Actually we had an optimization pass(`PermutationEliminationPass`) but it is disabled.

By the way is this much better than having a map in both `OpSequence` and `OpSequences`, then it can find two constant time searches.

periannath:
@wateret This PR handles removal of operations. `findSequenceIndex` always checks whether cached operation exist on `OpSequence` or not.
If that operation is removed from `OpSequence`, `findOperation` will fallback to search whole `OpSequences`.

wateret:
@periannath Actually I was talking about users calling `OpSequence::remove` directly not through `OpSequences`. But it seems like there is no use like that. And I just wanted to avoid that we should keep it in mind.

So I asked this.

> By the way is this much better than having a map in both `OpSequence` and `OpSequences`? which can find two constant time searches.

periannath:
@wateret `findSequenceIndex` always checks whether the cached index is valid or not. If an operation is removed from OpSequence by `OpSequence::remove`, that operation will be invalidated since it does not exist in cached OpSequence.

When I checked some models, most OpSequence has 1 or 2 operations. I think traversing `OperationIndex` vector would be enough for those OpSequence with few operations. Maintaining another map for `OpSequence` would incur additional memory overhead for a small benefit.

wateret:
@periannath Sorry about that, I misread the code 😅

wateret:
Ready to get merged?

PR #2818, opened 2020-07-02
Append delay/sleep to runner between runs

YongseopKim:
Append delay/sleep to nnpackage_run, tflite_run and tflite_run_2_2_0
between runs as ms units


YongseopKim:
PTAL @chunseoklee

PR #2823, opened 2020-07-02
[circle2circle] Change the way models are loaded

mhs4670go:
This commit changes the way models are loaded

Related : #2753 


mhs4670go:
@seanshpark This pr is okay:)

PR #2837, opened 2020-07-02
[onert] Nit: Fix bug in IPermuteFunction

wateret:
Size of the data type must be multiplied for copy size.


ragmani:
:eyes:

PR #2835, opened 2020-07-02
[DRAFT] Remove Source/Sink copies at execution begin/end

wateret:
Remove Source/Sink copies at execution begin/end, and make user's buffers as a `UserTensor` of `controlflow` backend and insert Permute to the graph instead of Source/Sink.

This is pre-work before Input/Output tensor copy elimination(but this work is like 90% and actual copy elimination is just 10%).

This also includes many other bug fixes.

Rebased version of #1325

PR #2838, opened 2020-07-02
[onert] Allow empty weight buffer in cpu FC impl

wateret:
Soon, weight's buffer could be `null` when it is a model input.

Part of #2835

### DIFF
--- a/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
+++ b/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
@@ -175,9 +175,14 @@ void FullyConnectedLayer::prepare()
     {
       // TODO If it's possible to extract precaching from ruy kernel,
       // place this instead of below code
-      assert(_weights->buffer() != nullptr);
-      // buffer will be used by ruy kernel as a cache key
-      _cached_weights = _weights->buffer();
+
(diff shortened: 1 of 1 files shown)


wateret:
@YongseopKim Besides this, does it consider weights as input case?

YongseopKim:
@wateret I'm not sure but it could be enough possible if fc-hybrid & weights as input. (Of course, I've not seen any case until now.) This is good for preparing for handling exceptions in advice.

wateret:
@YongseopKim I think `fullyconnected/weights_as_input` in NNAPI test tests it. But it is OK since it is run only once.

YongseopKim:
@wateret Normal fc should not enter the code. I'll fix the only `fc_hybrid` to enter the code.

YongseopKim:
@wateret In short, the code to save cache key(weight buffer address) should be entered by the only fc-hybrid to use ruy kernel.
 If the fc-hybrid has weights as input, the code would be entered. But the `fullyconnected/weights_as_input` 's weight type is not uint8 so that the fc should not enter the code. That's what I mean. Could you take a look at https://github.com/Samsung/ONE/pull/2857 ?

wateret:
@YongseopKim Oh I may have mixed up with `null` buffer and hybrid FC. I think you're right. Thank you for checking. Anyway what I wanted to say is the backend should consider if it is const or not. 😄

PR #2739, opened 2020-06-30
[Draft] Add BCQ recipes

llFreetimell:
Parent Issue : #1517 

This draft is for adding BCQ recipes.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

Issue #1517, opened 2020-05-27
[compiler] Add BCQ testcase


llFreetimell:
Currently, BCQ testcases for compiler are absent.
Let's add testcases by using `circlechef`.

### Todo (Draft)
- [x] circlechef/circle #2859
- [x] circlechef/core #2858
- [x] circlechef/proto #2846
- [x] res/CircleRecipes/BCQFullyConnected #2788
- [x] res/CircleRecipes/BCQGather #2923
- [x] Enable luci/tests #2924

### Draft

- #2739

llFreetimell:
I think that it would be better to create BCQ testcases at internal repository.

seanshpark:
> I think that it would be better to create BCQ testcases at internal repository.

I was thinking of creating simple tflite recipe file for the test material.
Can we create this recipe in this repo?

llFreetimell:
I agree for creating simple recipe file for test.

However, as BCQ format is upgraded, interpretation of BCQ operation inputs becomes harder than before.
As a result, creating a recipe file may consume much time I think and we have a high priority to run some models.

Therefore, I want to proceed with real test material which is generated by real model at first, and then make a simple recipe file :)

By the way, there is a possibility to use `circlechef`, not `tflchef` for the test material :)

llFreetimell:
### 20200630 Update

Main comment is edited

llFreetimell:
Related PRs are all merged, close!

PR #2846, opened 2020-07-03
[circlechef] Update proto for BCQ operations

llFreetimell:
Parent Issue : #1517
Draft : #2739

This commit will update `circlechef.proto` to include BCQ operations.

### DIFF
--- a/compiler/circlechef/proto/circlechef.proto
+++ b/compiler/circlechef/proto/circlechef.proto
@@ -68,6 +68,16 @@ message InstanceNormOptions {
   optional Activation activation = 2 [default = NONE];
 }
 
+message BCQFullyConnectedOptions {
+  optional int32 weights_hidden_size = 1 [default = 0];
+  optional Activation activation = 2 [default = NONE];
+}
+
+message BCQGatherOptions {
+  optional int32 input_hidden_size = 1 [default = 0];
+  optional int32 axis = 2 [default = 0];
+}
+
 message Operation {
   optional string type = 1;
   repeated string input = 2;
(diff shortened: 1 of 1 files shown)


compiler/circlechef/proto/circlechef.proto
```diff
@@ -68,13 +68,25 @@ message InstanceNormOptions {
   optional Activation activation = 2 [default = NONE];
 }
 
+message BCQFullyConnectedOptions {
+  optional int32 weights_hidden_size = 1 [default = 0];
+  optional Activation activation = 2 [default = NONE];
```
seanshpark, line 73: usually I've met activation in the first. is it OK to swap these two parameters?

compiler/circlechef/proto/circlechef.proto
```diff
@@ -68,13 +68,25 @@ message InstanceNormOptions {
...
+}
+
+message BCQGatherOptions {
+  optional int32 input_hidden_size = 1 [default = 0];
+  optional int32 axis = 2 [default = 0];
```
seanshpark, line 78: ditto, is axis first then input_hidden_size OK?

compiler/circlechef/proto/circlechef.proto
llFreetimell: I followed below order in circle schema not to be confused.

https://github.com/Samsung/ONE/blob/bd60a8ed0f25786d91e1949c810e624887a49bb7/nnpackage/schema/circle_schema.fbs#L991-L999

And `InstanceNormOptions` in this file also have activation as second.
Nevertheless if you don't like this, I will change it :)

seanshpark: Ah.. let's follow source circle schema :)

PR #2844, opened 2020-07-02
DRAFT Compiler FE reduce LOC-Class

seanshpark:
on-going draft to reduce LOC-class from our software analysis tool report

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #2843, opened 2020-07-02
Compiler FE reduce lines of code in class


seanshpark:
One of our internal software analysis tool reports lines of codes are high.
- compiler/luci/export/src/CircleOperationExporter.cpp
- compiler/luci/service/src/CircleShapeInferenceRule.cpp
- compiler/luci/logex/src/FormattedGraph.cpp
- compiler/locomotiv/src/NodeExecution.h

Let's try to reduce these

lemmaa:
FYI https://github.sec.samsung.net/RS7-Architectural-Refactoring/SAM-RefactoringGuide

seanshpark:
I think we can remove asserts in each visit methods.

seanshpark:
done this for now

PR #2851, opened 2020-07-03
[luci] introduce export_simple

seanshpark:
This will introduce common export_simple method to export a typical operator
- CircleCeil and CircleCos are revised to call export_simple to prevent unused compilation error


seanshpark:
For #2843, from draft #2844

seanshpark:

PR #2853, opened 2020-07-03
[onert] Revise parallel build setting

hseok-oh:
- Remove parallel setting in build command
- Set default parallel setting in Makefile.template


hseok-oh:
I've reverted default parallel setting on linux because many `Makefile.template` users build without `NPROCS`.

PR #2857, opened 2020-07-03
[onert] Avoid preparing cache for ruy kernel on fc layer, cpu

YongseopKim:
Avoid preparing cache for ruy kernel of fc layer on cpu.
The only fc hybrid will use ruy kernel to cache.

### DIFF
--- a/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
+++ b/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
@@ -164,6 +164,13 @@ void FullyConnectedLayer::run()
 void FullyConnectedLayer::prepare()
 {
 #ifdef USE_RUY_GEMV
+  // The only fc hybrid will use ruy kernel
+  if (_input->data_type() != OperandType::FLOAT32 ||
+      _weights->data_type() != OperandType::QUANT_INT8_SYMM)
+  {
+    return;
+  }
+
   // NOTE. The condition to enable caching on ruy kernel can be changed according to ruy's version
   const int rows = getTensorShape(_weights).Dims(0);
   if (rows % 4 == 0)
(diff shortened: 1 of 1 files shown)


runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
```diff
@@ -164,6 +164,13 @@ void FullyConnectedLayer::run()
 void FullyConnectedLayer::prepare()
 {
 #ifdef USE_RUY_GEMV
+  // The only fc hybrid will use ruy kernel
```
wateret, line 167: What about leaving a TODO comment that this is a workaround and it should check if the weight is const.

ragmani: It should also check if the input is dynamic.

YongseopKim:
Sorry for late. Could you take a look please? @wateret and @ragmani

runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
```diff
@@ -164,13 +164,17 @@ void FullyConnectedLayer::run()
...
+  // TODO This is workaround
   // The only fc hybrid will use ruy kernel
   if (_input->data_type() != OperandType::FLOAT32 ||
       _weights->data_type() != OperandType::QUANT_INT8_SYMM)
   {
     return;
   }
 
+  // For now, dynamic input cannot support
+  assert(_input->is_dynamic() == false);
```
ragmani, line 176: Why did you add this as an assertion?
If `_input` is dynamic, this method is not called?

YongseopKim: Is this your comment's intention? I understood that this fc doesn't support dynamic tensor as input. Exactly what does it 
mean?

> It should also check if the input is dynamic.

YongseopKim: FYI I'm not good at dynamic tensor code.

ragmani: If so, could you add a TODO comment instead of the assertion.

YongseopKim:
PTAL

PR #2859, opened 2020-07-03
[circlechef] Introduce BCQ operations circle directory

llFreetimell:
Parent Issue : #1517
Draft : #2739

This commit will add BCQ operations in `circlechef/circle` directory.

### DIFF
--- a/compiler/circlechef/circle/src/Op/BCQFullyConnected.cpp
+++ b/compiler/circlechef/circle/src/Op/BCQFullyConnected.cpp
@@ -24,7 +24,22 @@ namespace circlechef
 void CircleOpBCQFullyConnected::filler(const circle::Operator *op, CircleImport *import,
                                        circlechef::ModelRecipe *model_recipe) const
 {
-  // Nothing to do with filler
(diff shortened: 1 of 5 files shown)


compiler/circlechef/circle/src/Op/BCQFullyConnected.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "Convert.h"
+
+namespace circlechef
+{
+
+void CircleOpBCQFullyConnected::filler(const circle::Operator *op, CircleImport *import,
+                                       circlechef::ModelRecipe *model_recipe) const
+{
+  // Nothing to do with filler
```
seanshpark, line 27: .recipe file in the draft has
```
operand {
  name: "w_a"
  type: FLOAT32
  shape { dim: 32 }
  filler {
    tag: "gaussian"
    arg: "0.0"
    arg: "1.0"
  }
}
```
and seems we need to do something in this method.

llFreetimell:
@seanshpark Now recipes are all created in the draft PR and added missing filler codes here, PTAL :)

Issue #2743, opened 2020-07-01
Compiler FE: Common Intermediate Artifacts


mhs4670go:
In the middle of building our compiler module, we make several intermediate artifacts. But there are many overlapping processes because multiple modules make the same output(e.g. `.tflite`, `.circle`).

How about making an intermediate output beforehand and using it?

In this module(?), these will be created.
- recipe
- tensorflow lite
- circle
- nnpackage
- nnpackage test data(input.h5, expected.h5)

[This draft](https://github.com/Samsung/ONE/pull/2586) can be changed to this module I think.

seanshpark:
@mhs4670go , can you describe in form of

AS-IS
- each source resides in `res/TensorFlowLiteRecipes` folder
- each modules in `compiler/` will make a copy of this recipe to it's build output folder
- each module will do it's conversion or call other module(s) to convert to intermediate file and do it's own conversion

TO-BE
- ...

mhs4670go:
AS-IS

- each source resides in res/TensorFlowLiteRecipes folder
- each modules in compiler/ will make a copy of this recipe to it's build output folder
- each module will do it's conversion or call other module(s) to convert to intermediate file and do it's own conversion

TO-BE
- `common-artifacts` module will do conversion of all intermediate file to prevent duplication of copies
- each module will be able to get required resources from `common-artifacts` module.

I think copy of recipe is redundant as well but it seems to be okay with the concept of putting resources together.

PR #2860, opened 2020-07-03
[common-artifacts] Introduce common artifacts

mhs4670go:
This commit introduces common-artifacts to our compiler

Related : #2743 
Draft : #2586 


compiler/common-artifacts/CMakeLists.txt
```diff
@@ -0,0 +1,162 @@
...
+  set(INPUT_BIN_PATH "${CMAKE_CURRENT_BINARY_DIR}/${INPUT_HDF5_FILE}")
+  
+  set(EXPECTED_HDF5_FILE "${RECIPE}${OPT_FORMAT}.expected.h5")
+  set(EXPECTED_BIN_PATH "${CMAKE_CURRENT_BINARY_DIR}/${EXPECTED_HDF5_FILE}")
+  
+  if(NOT DEFINED NO_TCGEN_${RECIPE})
+    # Generate input.h5, expected.h5
+    add_custom_command(OUTPUT ${INPUT_BIN_PATH} ${EXPECTED_BIN_PATH}
+      COMMAND $<TARGET_FILE:testDataGenerator> ${MODEL_FILE}
+      DEPENDS $<TARGET_FILE:testDataGenerator> ${MODEL_FILE}
```
jinevening, line 130: testDataGenerator should accept a circle file (am I right?). 

If a user does not generate a circle file using circlize(~), ${MODEL_FILE} will be a tflite model. (MODEL_FORMAT is tflite if NO_OPTIMIZE_${RECIPE} is defined)

Have you tested this case?

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -0,0 +1,234 @@
...
+  uint32_t input_index = 0;
+  for (uint32_t g = 0; g < circle_model->subgraphs()->size(); g++)
+  {
+    const auto input_nodes = loco::input_nodes(module->graph(g));
+    for (const auto &node : input_nodes)
+    {
+      const auto *input_node = dynamic_cast<const luci::CircleInput *>(node);
+      std::string name = input_node->name();
+      if (name.find(":0") == std::string::npos)
+        name += ":0";
```
jinevening, line 148: Is this a convention of OneRT ? This will add ":0" even if the name has ":1" or ":2".

compiler/common-artifacts/exclude.lst
```diff
@@ -0,0 +1,160 @@
+# circlize : Exclude from tflite-to-circle conversion(tflite2circle)
```
jinevening, line 1: Just for curiosity. Is there a case that uses `circlize` ? AFAIK, circle schema is a superset of tflite schema, which means that any tflite model can be converted to circle model.

compiler/common-artifacts/README.md
```diff
@@ -0,0 +1,3 @@
+# common-artifacts
+
+`common-artifacts` is a module that produces intermediate artifacts that are commonly generated for compiler testing.
```
jinevening, line 3: Can you add some more information about the artifacts? E.g,
```
List of intermediate artifacts
recipe
tflite
circle
circle (applied all optimizations in circle2circle)
input data for nnpackage (.h5)
expected output data for nnpackage (.h5)
```

compiler/common-artifacts/CMakeLists.txt
mhs4670go: @jinevening That's a good point! I didn't think of this.

compiler/common-artifacts/exclude.lst
mhs4670go: @jinevening Actually I thought like what you said at the first place. But as @seanshpark said, it is for developer's convenience. When we implement a new operator, dividing the step will help in gradual development. If there's no exclusion of "circlize", we will get an error because of NYI in `tflite2circle`.

compiler/common-artifacts/src/TestDataGenerator.cpp
mhs4670go: Good point! it's my mistake.

mhs4670go:
@jinevening PTAL:)

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -144,8 +144,8 @@ int entry(int argc, char **argv)
     {
       const auto *input_node = dynamic_cast<const luci::CircleInput *>(node);
       std::string name = input_node->name();
-      if (name.find(":0") == std::string::npos)
-        name += ":0";
+      if (name.find(":") == std::string::npos)
+        name += ":";
```
jinevening, line 148: This will make input name "~~:".

Maybe `name += ":" + std::to_string(<index_of_input_node>)` ?

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -195,8 +195,8 @@ int entry(int argc, char **argv)
     {
       const auto *output_node = dynamic_cast<const luci::CircleOutput *>(node);
       std::string name = output_node->name();
-      if (name.find(":0") == std::string::npos)
-        name += ":0";
+      if (name.find(":") == std::string::npos)
+        name += ":";
```
jinevening, line 199: ditto

mhs4670go:
@jinevening PTAL:)

PR #2868, opened 2020-07-03
[luci-interpreter] Test quantized Conv2D kernels for ResNet-v1

struss:
Parent Issue: #2730 
Draft PR: #2720

This commit update `Conv2D` operator test case.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Conv2D.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Conv2D.test.cpp
@@ -65,8 +65,119 @@ TEST(Conv2DTest, Float)
       11, 16, 7, 20, // row = 0
       0,  40, 0, 44, // row = 1
   };
+  std::vector<int32_t> ref_output_shape{1, 2, 2, 2};
   EXPECT_THAT(extractTensorData<float>(output_tensor),
               ElementsAreArray(ArrayFloatNear(ref_output_data)));
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray(ref_output_shape));
+}
+
+TEST(Conv2DTest, FloatCheck)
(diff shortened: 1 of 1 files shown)


jinevening:
@binarman @s-barannikov PTAL

compiler/luci-interpreter/src/kernels/Conv2D.test.cpp
```diff
@@ -65,8 +65,119 @@ TEST(Conv2DTest, Float)
       11, 16, 7, 20, // row = 0
       0,  40, 0, 44, // row = 1
   };
+  std::vector<int32_t> ref_output_shape{1, 2, 2, 2};
   EXPECT_THAT(extractTensorData<float>(output_tensor),
               ElementsAreArray(ArrayFloatNear(ref_output_data)));
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray(ref_output_shape));
+}
+
+TEST(Conv2DTest, FloatCheck)
```
binarman, line 74: If I got it right this is float version of actual quantized test.
I found it difficult to link these functions if you didn't see this PR.

Optional:
Move input/reference data to some shared place (for example set of functions that fills tensors), so these tests will be related.

struss: Oh, Sorry. I forget to remove that. while Adding Uint8 test case, some values from output tensor is different from what I expected in Uint8 case. so, I check the same value with float operation, and finally I found that I didn't make bias value quantized. sorry for confused!
and I will move input/reference data to some shared place on next PR with some common function move to testutils! Thanks!(maybe it will be a refactoring of test...)

seanshpark:
@s-barannikov PTAL

PR #2871, opened 2020-07-03
[luci] Support pass for new InstanceNorm graph

d-krylov:
Support pass for new InstanceNorm graph


compiler/luci/pass/src/FuseInstanceNormPass.cpp
```diff
@@ -381,13 +550,28 @@ namespace luci
...
+
   for (auto node : loco::active_nodes(loco::output_nodes(g)))
   {
-    auto add = dynamic_cast<luci::CircleAdd *>(node);
-    if (not add)
-      continue;
+    auto reshape = dynamic_cast<luci::CircleReshape *>(node);
+    if (not reshape)
+    {
+      add = dynamic_cast<luci::CircleAdd *>(node);
```
binarman, line 561: I also think this logic is part of matcher, so we can move it inside.

compiler/luci/pass/src/FuseInstanceNormPass.test.cpp
```diff
@@ -0,0 +1,64 @@
...
+  setShape(const_node, {8});
+  EXPECT_FALSE(is_quasi_1D_with_dummy_dim(&const_node, 8));
+
+  setShape(const_node, {1, 2, 1, 8, 1});
+  EXPECT_FALSE(is_quasi_1D_with_dummy_dim(&const_node, 8));
+
+  setShape(const_node, {8, 3});
+  EXPECT_FALSE(is_quasi_1D_with_dummy_dim(&const_node, 8));
+
+  setShape(const_node, {8, 1});
```
binarman, line 55: @d-krylov 
I am not sure about this case, could you comment?

compiler/luci/pass/src/FuseInstanceNormPass.cpp
binarman: Upd. I looked again, need to think how to refactor this beter. Fix it later.

compiler/luci/pass/src/FuseInstanceNormPass.test.cpp
d-krylov: I ihink, EXPECT_FALSE. Because, if input shape = {8} -> all shapes in model = {1,8,1}

compiler/luci/pass/src/FuseInstanceNormPass.test.cpp
```diff
@@ -0,0 +1,64 @@
+/*
```
binarman, line 1: @seanshpark 
I've added this file for internal functions testing, is it ok?

I am also thinking about negative tests for this optimization (creating whole graph in luci IR).
Is it ok to create graph in this file, and not by recipe?

seanshpark: > I've added this file for internal functions testing, is it ok?

Seems OK. I'm not sure what concern you have for adding tests :)

> Is it ok to create graph in this file, and not by recipe?

Yes, graph from codes (adding nodes with C++ codes) is also OK.

compiler/luci/pass/src/FuseInstanceNormPass.cpp
```diff
@@ -175,7 +190,53 @@ bool is_instance_mean(luci::CircleMean *mean)
...
+  //
+  auto input = mean->input();
+  if (not loco::shape_known(input))
+    return false;
+  auto input_shape = loco::shape_get(input).as<loco::TensorShape>();
+  if (input_shape.rank() != 5)
+    return false;
+
+  //
+  // CHECK 2) 'reduction indices' is CircleConst of value [1,2,4], that is HW of NHWC
```
s-barannikov, line 206: I did not get it. The input is 5-D and there should be three reduction incides -- 1, 2 and 4. How does it relate to3 HW and NHWC?

compiler/luci/pass/src/FuseInstanceNormPass.cpp
```diff
@@ -381,13 +551,28 @@ namespace luci
...
   bool changed = false;
+  luci::CircleAdd *add;
+  InstanceNormPattern::PatternVersion pv;
+
   for (auto node : loco::active_nodes(loco::output_nodes(g)))
   {
-    auto add = dynamic_cast<luci::CircleAdd *>(node);
-    if (not add)
-      continue;
+    auto reshape = dynamic_cast<luci::CircleReshape *>(node);
```
s-barannikov, line 559: (Optional) You could just check the opcode.

compiler/luci/pass/src/FuseInstanceNormPass.cpp
```diff
@@ -114,8 +115,6 @@ bool NodeFiller<ARG_TYPE_1, ARG_TYPE_2>::with_commutative_args_of(const COMM_NOD
 } // namespace
 
 // Helper to check detail
-namespace
```
seanshpark, line 117: Q) is there any particular reason to remove this?

binarman: Yes, I removed this so I can write unit tests for these functions in *.test.cpp file

compiler/luci/pass/src/FuseInstanceNormPass.cpp
```diff
@@ -175,7 +190,53 @@ bool is_instance_mean(luci::CircleMean *mean)
...
 }
 
-} // namespace
+bool is_instance_mean_v1(luci::CircleMean *mean)
+{
+  //
+  // CHECK 1) input is rank 5
+  //
+  auto input = mean->input();
+  if (not loco::shape_known(input))
```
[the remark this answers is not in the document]
binarman: I don't want to change this (at least in this PR), because old code in this file is written in this manner.

compiler/luci/pass/src/FuseInstanceNormPass.cpp
binarman: `reshape` is used on the line 569, so we need this variable anyway. I think we can leave it as-is for now.

compiler/luci/pass/src/FuseInstanceNormPass.cpp
binarman: If I got it right, 5-d should be always `N x H x W x C x 1`, so 4-th index just removes redundant dimension

binarman:
@seanshpark 
oops, sorry, I pushed changes in comments, could you approve again?

PR #2873, opened 2020-07-03
[onert] Fix getting OpSequence inputs

wateret:
As not only the first operation can have inputs, we should count all
the operations other which are not an output of the current
`OpSequence`.


wateret:
Before/After

Screen Shot 2020-07-03 at 7 38 23 PM → Screen Shot 2020-07-03 at 7 23 54 PM

YongseopKim:
ping @Samsung/nnfw_committers

PR #2874, opened 2020-07-03
[onert] Do nothing on Permute dynamic inference

wateret:
Do nothing on Permute dynamic shape inference. It will be done in
the kernel.

Part of #2835


ragmani:
I think it is okay to remove the entire function.

hyunsik-yoon:
maybe leaving a comment (like the following?) could help future readers. 

> `shape inference and output allocation will be done in the kernel.`

wateret:
@ragmani @hyunsik-yoon Thank you guys! I updated the commit accordingly.

PR #2885, opened 2020-07-05
[luci] Revise export to reduce class LOC T to Z

seanshpark:
This will revise export to reduce class LOC from CircleTanh to last


seanshpark:
@sk-namkoong @sw4670-chae @hj1-jeong

PR #2892, opened 2020-07-06
[onert] dont' allocate memory for small size in reduce operation

glistening:
axis and temporary are usually smaller than 5.
It will use pre allocated array for those case instead of allocation.

### DIFF
--- a/compute/cker/include/cker/operation/Reduce.h
+++ b/compute/cker/include/cker/operation/Reduce.h
@@ -128,8 +128,10 @@ public:
       return;
 
     // prepare space for temp_index and resolved_axis
-    _temp_index.resize(temp_index_size);
-    _resolved_axis.resize(resolved_axis_size);
+    if (temp_index_size > kMaxSmallSize)
+      _temp_index.resize(temp_index_size);
+    if (resolved_axis_size > kMaxSmallSize)
+      _resolved_axis.resize(resolved_axis_size);
     _prepared = true;
   }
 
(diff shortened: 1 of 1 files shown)


compute/cker/include/cker/operation/Reduce.h
```diff
@@ -148,23 +150,31 @@ class Reduce
...
+  inline int32_t *temp_index_data(void)
+  {
+    return _temp_index.size() ? _temp_index.data() : _temp_index_small;
+  }
 
 private:
   std::vector<int> _temp_index;
   std::vector<int> _resolved_axis;
   bool _prepared;
+  static const int kMaxSmallSize = 4;
```
periannath, line 175: (Optional) How about define this variable as `constexpr`?
```suggestion
  static constexpr int kMaxSmallSize = 4;
```

glistening: @periannath What is different between `const` and `constexp` in this case? It is clearly `const`. Does it generate better code?

periannath: @glistening 

> The primary difference between const and constexpr variables is that the initialization of a const variable can be deferred until run time. A constexpr variable must be initialized at compile time. All constexpr variables are const.

From https://docs.microsoft.com/en-us/cpp/cpp/constexpr-cpp?view=vs-2019

`constexpr` might be faster than `const` since its value is evaluated at compile tile.

glistening: @periannath As I said, it is `int`. Compiler knows it is `const` at compilation time. Many answers from SO also says there is no difference for those cases. What you're quote is general story. `constexpr` is not faster for `const int` case.

glistening: @periannath  I think the statement below does not come from your reference. It is your opinion.

> `constexpr` might be faster than const since its value is evaluated at compile tile.

https://docs.microsoft.com/en-us/cpp/cpp/constexpr-cpp?view=vs-2019 says

>  And when a value is computed at compile time instead of run time, it helps your program run faster and use less memory.

For `const int` in this case, it is clearly computed at compile time, it is the reason many (including me) says there is no difference for this case.

periannath: @glistening 
> For const int in this case, it is clearly computed at compile time, it is the reason many (including me) says there is no difference for this case.

You're right. I just found that there's no difference for this compile-time constant expression. I think `constexpr` is more clear than `const`, but it seems like a matter of taste.

PR #2893, opened 2020-07-06
[luci] Revise shape inf to reduce class LOC

seanshpark:
This will introduce 4 common methods to reduce class LOC in CircleShapeInferenceRule.cpp


seanshpark:
For #2843, from draft #2844

seanshpark:

PR #2894, opened 2020-07-06
[onert][cker] Introduce checking constant into OperandInfo

ragmani:
This commit introduces checking constant into OperandInfo.
  - Introduce isConstant() into OperandInfo
  - Add is_constant() into ITensor
  - Make cpu::ConvolutionLayer uses is_constant() of Tensor.
  - Apply constant and dynamic for ConvolutionLayer::prepare()


wateret:
@YongseopKim FYI this will introduce `ITensor::is_constant`, which can be used for Layers check if the tensors(weights) are constants.

PR #2903, opened 2020-07-06
[luci] Introduce common methods for NodeSummaryBuilder

seanshpark:
This will introduce several common methods for CircleNodeSummaryBuilder
- call these methods once not to make unused code compilation error


seanshpark:
For #2843 from draft #2844

PR #2904, opened 2020-07-06
[locomotiv] Revise AvgPool2D to reduce class LOC

seanshpark:
This will revise AvgPool2D to reduce class LOC
- this change is an experiment and if the score doesn't change, I'll roll back


seanshpark:
For #2843 from draft #2844

seanshpark:
Value decreased from 975 to 952

PR #2907, opened 2020-07-06
[luci] Use common for CircleNodeSummaryBuilder

seanshpark:
This will revise CircleNodeSummaryBuilder to use common methods to reduce class LOC


seanshpark:
For #2843 from draft #2844

seanshpark:

PR #2909, opened 2020-07-06
[tf2nnpackage-value-remote-test] Introduce e2e test

mhs4670go:
This commit introduces tf2nnpackage-value-remote-test to compiler

Related: #2484 
Draft: #2586 

### DIFF
--- a/compiler/tf2nnpackage-value-remote-test/CMakeLists.txt
+++ b/compiler/tf2nnpackage-value-remote-test/CMakeLists.txt
@@ -1,14 +1,13 @@
-# Read "test.lst" if exists
-include("test.lst" OPTIONAL)
+include("test.lst")
 
 # Do not make test if there are no remote machine information
 if(NOT REMOTE_IP)
-  message("tf2nnpackage-value-remote-test: need to set REMOTE IP")
+  message(STATUS "tf2nnpackage-value-remote-test: need to set REMOTE IP")
   return()
 endif(NOT REMOTE_IP)
 
 if(NOT REMOTE_USER)
-  message("tf2nnpackage-value-remote-test: need to set REMOTE_USER")
+  message(STATUS "tf2nnpackage-value-remote-test: need to set REMOTE_USER")
   return()
 endif(NOT REMOTE_USER)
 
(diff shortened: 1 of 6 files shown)


compiler/tf2nnpackage-value-remote-test/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+    list(APPEND TEST_NAMES ${DIR})
+  endif()
+endforeach()
+
+get_target_property(ARTIFACTS_SRC_PATH testDataGenerator SOURCE_DIR)
+
+macro(circlize)
+endmacro()
+macro(optimize)
+endmacro()
```
seanshpark, line 40: above two macro is empty. are these intended?

compiler/tf2nnpackage-value-remote-test/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
+# Read "test.lst" if exists
+include("test.lst" OPTIONAL)
```
seanshpark, line 2: 1) other tests require `test.lst` and optionally read `test.local.lst`. how about using this name?
2) I don't see mandatory test list file... it's a bit strange. maybe add a NOTE about your intention.

mhs4670go: First, `test.lst` is mandatory because this test need `REMOTE_IP` and `REMOTE_USER`. But if I remove `OPTIONAL` keyword, configure step will stop. So, I set it to optional and return when there is no `REMOTE_IP` and `REMOTE_USER` below. 

Second, right there is no mandatory test list file this is why below some macro are empty. Actually, in `common-artifacts`, all of the modules `tf2nnpackage-value-remote-test` need to test - tflchef, tflite2circle, circle2circle, luci-interpreter - are tested. That is, `common-artifacts` already includes this modules except runtime test.

So, there is no need `test.lst`. In this module, only the runtime test is performed, and for this runtime test, generation of test data is required. And, `tcgenerate` in `${ARTIFACTS_SRC_PATH}/exclude.lst` means it won't generate test data, which is why below macro excludes specific opearators from runtime test.

seanshpark: As talked off-line, we can add an empty `test.lst`, maybe with a comment at the first line.

compiler/tf2nnpackage-value-remote-test/CMakeLists.txt
seanshpark: maybe adding a comment for empty macro would be preferable

mhs4670go:
@seanshpark PTAL:)

Issue #2910, opened 2020-07-06
[onert] Debug build runtime test fail


hseok-oh:
```
19:35:53  [ RUN      ] GeneratedTests.abs_dynamic_nnfw
19:35:53  nnapi_gtest: /home/jenkins_slave/workspace/nnfw/master/push-cross-aarch64-runtime-test-debug/runtime/onert/core/include/ir/OperandInfo.h:121: bool onert::ir::OperandInfo::isConstant() const: Assertion `!isDynamic()' failed.
19:35:53  /jenkins/workspace/nnfw/master/push-cross-aarch64-runtime-test-debug/tests/scripts/unittest.sh: line 71: 32181 Aborted                 (core dumped) $TEST_BIN $(get_gtest_option)
19:35:53  /jenkins/workspace/nnfw/master/push-cross-aarch64-runtime-test-debug/Product/out/unittest/nnapi_gtest failed... return code: 134
19:35:53  ============================================
19:35:53  Finishing set 4: /jenkins/workspace/nnfw/master/push-cross-aarch64-runtime-test-debug/Product/out/unittest/nnapi_gtest...
19:35:53  ============================================
```

Fail after #2894

hseok-oh:
Resolved

PR #2912, opened 2020-07-06
[onert] Fix assert condition in isConstant()

hseok-oh:
Assert for impossible case: constant and dynamic operand


hseok-oh:
For #2910

PR #2917, opened 2020-07-06
[circle2circle-dredd-recipe-test] Enable resolve_customop_matmul pass

kvochko:
This enables resolve_customop_matmul pass for MatMul_000 recipe in
circle2circle-dredd-recipe-test

For #1854 
Draft #2121 

ONE-DCO-1.0-Signed-off-by: Andrey Kvochko <a.kvochko@samsung.com>

PR #2915, opened 2020-07-06
[circle2circle] Support custom matmul operator

kvochko:
This adds resolve_customop_matmul option to circle2circle

For #1854 
Draft #2121 

ONE-DCO-1.0-Signed-off-by: Andrey Kvochko <a.kvochko@samsung.com>

PR #2121, opened 2020-06-10
[DRAFT] Implement CUSTOM(MatMul) operation

kvochko:
This enables import of a custom matmul operation and rewrites it as
Transpose+FullyConnected when possible.

For #1854

Issue #1854, opened 2020-06-03
Compile FE: CUSTOM(MatMul) to FC+TRANSPOSE


seanshpark:
Latest (TF 2.2) version TF lite converter converts `MatMul` Op to `FULLY_CONNECTED` + `TRANSPOSE`.
But, old version drops it as `CUSTOM(MatMul)`.

We may need to add a Pass to do this if our import reads this Op.

seanshpark:
From #1675

binarman:
@seanshpark 
> We may need to add a Pass to do this if our import reads this Op.

Do you mean `CUSTOM(MatMul)` to `FULLY_CONNECTED` conversion?
I also want to point, that latest TF is not able to convert all variants of `MatMul` operator.
For example, we still can get custom op for `tf.linalg.matmul(x, y, transpose_a=True, transpose_b=False)`

seanshpark:
> Do you mean CUSTOM(MatMul) to FULLY_CONNECTED conversion?

Yes.. 

> I also want to point, that latest TF is not able to convert all variants of MatMul operator.

Yes, not all but it is good if we can convert any possible cases.

kvochko:
@seanshpark is this task still available? If you don't mind, I'd like to take a crack at it

seanshpark:
>  is this task still available? If you don't mind, I'd like to take a crack at it

Yes! it is available :) Please go ahead!

seanshpark:
@kvochko, finished?

kvochko:
Yes, it's all done!

seanshpark:
Thank you!!!

PR #2918, opened 2020-07-06
[luci] Enable custom matmul op in luci tests

kvochko:
This enables tests for custom MatMul operator

For #1854 
Draft #2121 

Needs to be rebased on #2915 and #2917 

ONE-DCO-1.0-Signed-off-by: Andrey Kvochko <a.kvochko@samsung.com>


seanshpark:
Although #2915 has not landed but the test passed. I think test material is somehow not dependent on #2915.

PR #2920, opened 2020-07-06
[infra] Turn on optimization in tf2nnpkg.20200630

seanshpark:
This will enable all optimization options in tf2nnpkg.20200630 when calling circle2circle


seanshpark:
Related #42

PR #2921, opened 2020-07-06
[tf2circleV2] Introduce tf2circleV2

seanshpark:
This will introduce tf2circleV2 program that converts TF model to circle model


seanshpark:
closing as one-* commands replaces this

PR #2928, opened 2020-07-07
[tests/tools] Make nnapi_test runs nnapi_delegate twice

ragmani:
For issue #2929

This commit makes nnapi_test runs nnapi_delegate twice.

### DIFF
--- a/tests/tools/nnapi_test/src/nnapi_test.cc
+++ b/tests/tools/nnapi_test/src/nnapi_test.cc
@@ -57,7 +57,9 @@ int main(const int argc, char **argv)
     {
       seed = std::atoi(argv[2]);
     }
-    return nnfw::tflite::RandomTestRunner::make(seed).run(builder);
+    auto runner = nnfw::tflite::RandomTestRunner::make(seed);
+    runner.compile(builder);
+    return runner.run(2);
   }
   catch (const std::exception &e)
   {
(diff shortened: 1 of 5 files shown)


YongseopKim:
@ragmani I've heard that nnapi_test failes twice because of dealloc const tensors which are already cached by ruy kernel. I think you should append disable code on prepare() of FullyConnectedLayer on cpu. I'll investigate why this happes.

ragmani:
@YongseopKim 

I got it. If so, could you assign #2929 to yourself?
And please review again this patch. This enables to verify running `nnapi_test` twice.

tests/tools/nnapi_test/src/nnapi_test.cc
```diff
@@ -57,7 +57,9 @@ int main(const int argc, char **argv)
     {
       seed = std::atoi(argv[2]);
     }
-    return nnfw::tflite::RandomTestRunner::make(seed).run(builder);
+    auto runner = nnfw::tflite::RandomTestRunner::make(seed);
+    runner.compile(builder);
+    return runner.run(2);
```
hseok-oh, line 62: I have two concern

- It seems strange `nnapi_test` itself always run twice
- It increases CI test time

ragmani: > It seems strange `nnapi_test` itself always run twice

Right, I will change running count to parameter of `nnapi_test`

> It increases CI test time

I think it is not big because it consumes only execution time once more.

PR #2950, opened 2020-07-07
[luci] Support MatrixSetDiag operator

d-krylov:
Support MatrixSetDiag operator in luci
    

compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -630,6 +631,20 @@ void OperationExporter::visit(luci::CircleLogSoftmax *node)
...
+  uint32_t op_idx = md.registerBuiltinOpcode(circle::BuiltinOperator_MATRIX_SET_DIAG);
+  std::vector<int32_t> inputs_vec{get_tensor_index(node->input()),
+                                  get_tensor_index(node->diagonal())};
+  std::vector<int32_t> outputs_vec{get_tensor_index(static_cast<loco::Node *>(node))};
+  auto inputs = builder.CreateVector(inputs_vec);
+  auto outputs = builder.CreateVector(outputs_vec);
+  auto options = CreateMatrixSetDiagOptions(builder);
+  auto op_offset = CreateOperator(builder, op_idx, inputs, outputs,
+                                  circle::BuiltinOptions_MatrixSetDiagOptions, options.Union());
+  gd._operators.push_back(op_offset);
```
seanshpark, line 645: Can you please use `export_simple()` ? Purpose is to decrease total LOC of this class for our SE guide; so that this class doesn't get huge.

seanshpark:
Was there some problem with testing? You've removed items in the test list file.

seanshpark:
MatrixDiag landed and there are conflicts. Please resolve conflicts.

binarman:
@seanshpark 
> Was there some problem with testing? You've removed items in the test list file.

Yes, this is due to recent introduction of `common-artifacts` component that runs tests for all recipes except listed in `exclude.lst` file.

Consider these facts:
1. We can not merge luci tests before recipe (luci tests will fail)
2. We can not merge recipe before luci (common-artifact will fail on circle2circle testsuite)

So now we decided to merge components in following order:
1. luci (without tests)
1. recipe
1. luci tests

seanshpark:
> So now we decided to merge components in following order:

Thanks for the explanation!

`common-artifacts` gave some side effects... I'll post an issue to solve or make it easy to merge codes...

PR #2953, opened 2020-07-08
[onert] Use optimized logistic cpu kernel for aarch64 only

periannath:
- This commit uses optimized logistic cpu kernel for aarch64 only
  - Optimized logistic kernel produces numerical error except for aarch64
  - Use reference kernel in platforms other than aarch64 to deal with value mismatch


hseok-oh:
LGTM. 

We may need to update `eigen` version using `cker`.

Issue #2943, opened 2020-07-07
[onert] Do not use FunctionSequenceForDynamicTensor when all tensors are static


hyunsik-yoon:
When all tensors are static, using `FunctionSequenceForDynamicTensor` will hurt the performance.
In such case we can run `FunctionSequence`.

Let's make code not to use `FunctionSequenceForDynamicTensor` when function sequence does not need to handle dynamic tensors.

previously, 

```
# model does not have dynamic tensor
...
nnfw_prepare() --> at this stage we create `FunctionSequenceForDynamicTensor` when backend is CPU
nnfw_run() --> dynamic inferer is called, creating a bit of overhead
```

We should consider the following case:
```
# model does not have dynamic tensor
...
nnfw_prepare() --> at this stage we create `FunctionSequence` since model does not have dynamic tensor!
nnfw_set_input_tensorinfo() ---> what? creating a dynamic tensor?
nnfw_run()
```

So, we should decide to call dynamic shape inferer or not **at execution time** (at the beginning of `nnfw_run()`).

hyunsik-yoon:
with pr #2957, performance enhances by about 3.2%.
(cpu governer: performance, release build, odroid xu4, ubuntu 18.04)

```
# previously

$ git log
commit 918142191db4b97b2166c30209d2924ea86892c0 (upstream/master, master)
Author: Hyeongseok Oh <hseok82.oh@samsung.com>
Date:   Fri Jul 10 13:23:48 2020 +0900
...

$ ./Product/armv7l-linux.release/out/bin/nnpackage_run --nnpackage /home/data/model_zoo_private/E2EASR/ASR_MoChA_divide_enc/tflite_8bit/enc2 -w 10 -r 100
===================================
MODEL_LOAD   takes 5.027 ms
PREPARE      takes 166.066 ms
EXECUTE      takes 12.821 ms
- MEAN     :  12.821 ms
- MAX      :  13.188 ms
- MIN      :  12.474 ms
- GEOMEAN  :  12.820 ms
===================================

# after cherry-picking this PR 

$ ./Product/armv7l-linux.release.pr.2957/out/bin/nnpackage_run --nnpackage /home/data/model_zoo_private/E2EASR/ASR_MoChA_divide_enc/tflite_8bit/enc2 -w 10 -r 100
===================================
MODEL_LOAD   takes 5.172 ms
PREPARE      takes 169.360 ms
EXECUTE      takes 12.414 ms
- MEAN     :  12.414 ms
[...]

PR #2957, opened 2020-07-08
DRAFT/WIP : Use FuncionSequence when a model has static tensors only

hyunsik-yoon:
**DRAFT/WIP**

For #2943

This draft make `FunctionSequence` to be used when a model has static tensors only.

Previously, with CPU backend, FunctionSequenceForDynamicTensor was always used and this make some performance overhead with, e.g., `deallocInput()` (/cc @glistening)

### DIFF
--- a/runtime/onert/core/include/exec/FunctionSequence.h
+++ b/runtime/onert/core/include/exec/FunctionSequence.h
@@ -18,6 +18,7 @@
 #define __ONERT_EXEC_FUNCTION_SEQUENCE_H__
 
 #include <memory>
+#include <cassert>
 #include <vector>
 #include <functional>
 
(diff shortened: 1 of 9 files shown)


runtime/onert/core/include/compiler/StaticShapeInference.h
```diff
@@ -55,16 +55,30 @@ class StaticInferer : public ir::OperationVisitor
    *        when running kernel.
    * @param op_seq sequence of operations
    */
-  void infer(const ir::OpSequence &op_seq)
+  bool infer(const ir::OpSequence &op_seq)
   {
+    bool ret_has_dynamic_tensor = false;
+
+    // TODO Eliminate _has_dynamic_tensor by, e.g., introducing bool accept(..)
```
wateret, line 62: I personally prefer to not introducing `bool accept()`. We need other return types in other part of code. But it is not worth introducing all return types we need. So I ended up with having only one `accept` function.

runtime/onert/core/include/exec/FunctionSequence.h
```diff
@@ -72,33 +74,76 @@ class FunctionSequence : public IFunction
...
-                                    backend::IDynamicTensorManager *dyn_tensor_manager)
-      : _op_seq(op_seq), _operations_ctx(operations),
-        _dyn_shape_inferer(std::move(dyn_shape_inferer)), _dyn_tensor_manager(dyn_tensor_manager)
-  { /* empty */
+  /**
+   * @brief Prepare to run FunctionSequence which "might" handle dynamic tensor
+   * @note  Calling this does not mean that run() will handle dynamic tensor.
+   *        hasDynamicTensor(true) will make run() will handle dynamic tensor.
+   */
+  void dynamicTensorContext(const ir::OpSequence *op_seq, const ir::Operations *operations,
```
wateret, line 89: - It should be named `dynamic_tensor_context` or `setDynamicTensorContext` according to our coding convention
- What about packing 5 argument as a struct?
- What about moving this to the constructor? Calling this function seems like a must

hseok-oh:
Outdated, and some parts are already merged.

Issue #2809, opened 2020-07-02
How can I release with external libraries?


jinevening:
I'm writing this issue to ask helps for "release". I'm new to release, so any suggestion can be very helpful.
~~(@seanshpark Can you tell me the difference between "release" and https://github.com/Samsung/ONE/releases ?)~~

## Problem

I made a module (`record-minmax`) using external libraries (hdf5 and boost). How can I release the project with external libraries?

### boost

I found the cmake config file for boost in `infra/cmake/packages/BoostConfig.cmake`. The file downloads the boost library from the external download server written in `BoostSourceConfig.cmake` and builds the library.

I used the boost in `record-minmax` like this. `nnas_find_package(Boost QUITE COMPONENTS program_options)`. Should I do anything more for the release with boost?

### hdf5

I found the cmake config file for hdf5 in `infra/nnfw/cmake/packages/HDF5Config.cmake`. The file finds hdf5 in two ways.
1. Specifying the location of hdf5 library (H5Cpp.h, libhdf5.a, libhdf5_cpp.a) 
2. Finding hdf5 pre-installed in the system (find_package(HDF5 COMPONENTS CXX QUIET))

My program (`record-minmax`) is placed under the compiler directory, so I cannot use `nnfw_find_package`. Instead, I just used `find_package`, which is the same as the second way of the `HDF5Config.cmake`.

So, our customers should install hdf5 in advance. If the customers use Ubuntu, this may be done easily (sudo apt-get install libhdf5-dev), but it would be not simple in other platforms.

It seems that hdf5 can be downloaded and built with cmake (like [this](https://portal.hdfgroup.org/display/support/Building+HDF5+with+CMake)). But for this, we need CMake minimum version 3.12 (3.15 is recommended) [link](https://www.hdfgroup.org/2020/03/release-of-hdf5-1-12-0-newsletter-172/). I have no idea about the cmake version of our release (Can anyone tell me?).

cc @mhs4670go @cgbahk

seanshpark:
> Can you tell me the difference between "release" and https://github.com/Samsung/ONE/releases 

https://github.com/Samsung/ONE/releases  is the URL where we get the "Releases" files or artifacts

seanshpark:
What I think is
- run `record-minmax` without any problem is the purpose
- we may not have to build boost or hdf5 from source

To experience the problem of the end-user
- prepare a clean linux machine; virtual machine can be a solution
[...]

PR #2965, opened 2020-07-08
[DRAFT] workaround of hdf5 static build

mhs4670go:
This draft is a workaround of hdf5 static build

Related : #2809 


mhs4670go:
By setting `HDF5_USE_STATIC_LIBRARIES` to ON, `libhdf5` is linked statically. But, external libraries that is needed for hdf5(libz, libsz) are still needed dynamically. So in this draft I replace shared library to static library which exists together.

compiler/circle-tensordump/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 if(NOT TARGET mio_circle)
   return()
 endif(NOT TARGET mio_circle)
-
+set(HDF5_USE_STATIC_LIBRARIES OFF)
```
seanshpark, line 4: is it possible to introduce a macro in compiler cmake and call that?

mhs4670go: It's possible but I think this setting is really hdf5-specific. Maybe I'd better write hdf5 build script..

jinevening:
👍  Can you share the package file created in this way?

glistening:
[HDF5 license](https://support.hdfgroup.org/ftp/HDF5/releases/COPYING) says:

```
Copyright Notice and License Terms for 
HDF5 (Hierarchical Data Format 5) Software Library and Utilities
-----------------------------------------------------------------------------
...

2. Redistributions in binary form must reproduce the above copyright notice, 
   this list of conditions, and the following disclaimer in the documentation 
   and/or materials provided with the distribution.
```

Maybe we need to write hdf5 copyright somewhere if `hdf5` is included in our release package.

seanshpark:
https://support.hdfgroup.org/ftp/HDF5/releases/COPYING says in the botton,

> HDF5 is available with the SZIP compression library but SZIP is not part 
of HDF5 and has separate copyright and license terms. See SZIP Compression 
in HDF Products (www.hdfgroup.org/doc_resource/SZIP/) for further details.

compiler/record-minmax/CMakeLists.txt
```diff
@@ -19,8 +22,16 @@ add_executable(record-minmax ${DRIVER} ${SOURCES})
 target_include_directories(record-minmax PRIVATE include)
 target_include_directories(record-minmax PRIVATE ${Boost_INCLUDE_DIRS})
 target_include_directories(record-minmax PRIVATE ${HDF5_INCLUDE_DIRS})
-
-target_link_libraries(record-minmax ${HDF5_CXX_LIBRARIES})
+unset(HDF5_STATIC_LIBRARIES)
+foreach(LIB ${HDF5_CXX_LIBRARIES})
+  if(LIB MATCHES libsz.so OR LIB MATCHES libz.so)
+    string(REGEX REPLACE "[.]so$" ".a" LIB ${LIB})
```
seanshpark, line 28: As this does not apply to Windows platform, how about extracting this block of new codes to a function or a macro and call it. inside the macro or function it can branch depending on the platform
(I'm not sure branching by the platform is good or bad)

glistening: @mhs4670go, @seanshpark 

1) I guessed it would work on windows because we're using mingw or something for windows build, not native windows at least last Dec. Maybe I miss something. Could you let me know why this does nt apply to window platform?

@mhs4670go

2) As I know, `find_package` should have searched `*.a` if `HDF5_USE_STATIC_LIBRARIES` is set before `find_package`.

compiler/record-minmax/CMakeLists.txt
```diff
@@ -19,8 +22,16 @@ add_executable(record-minmax ${DRIVER} ${SOURCES})
 target_include_directories(record-minmax PRIVATE include)
 target_include_directories(record-minmax PRIVATE ${Boost_INCLUDE_DIRS})
 target_include_directories(record-minmax PRIVATE ${HDF5_INCLUDE_DIRS})
-
-target_link_libraries(record-minmax ${HDF5_CXX_LIBRARIES})
+unset(HDF5_STATIC_LIBRARIES)
```
glistening, line 25: @mhs4670go Why this line is necessary?

compiler/record-minmax/CMakeLists.txt
```diff
@@ -19,8 +22,16 @@ add_executable(record-minmax ${DRIVER} ${SOURCES})
...
+foreach(LIB ${HDF5_CXX_LIBRARIES})
+  if(LIB MATCHES libsz.so OR LIB MATCHES libz.so)
+    string(REGEX REPLACE "[.]so$" ".a" LIB ${LIB})
+    endif()
+    list(APPEND HDF5_STATIC_LIBRARIES ${LIB})
+endforeach()
+find_library(LIB_AEC NAMES "libaec.a")
```
glistening, line 32: @mhs4670go `find_library` will set `LIB_AEC` cache variable. Maybe you need to unset `LIB_AEC` before `find_library`.

compiler/record-minmax/CMakeLists.txt
mhs4670go: This is cache problem. Even if I set `HDF5_USE_STATIC_LIBRARIES`, cmake doesn't set it `*.a` because it's already `*.so`.

glistening:
@mhs4670go 

We may opt out `libsz` by building `hdf5` by ourselves without `sz` if its license could be problematic.

```sh
$ mkdir build_x86_64 && cd build_x86_64

# you don't need --without-szlib because it is off by default
$ ../configure --enable-cxx

Linking Options:
----------------
...
                Extra libraries: -lz -ldl -lm 
```

Note there is no `-lsz`.

```
$ make all install
$ tree hdf5/lib
├── libhdf5.a
├── libhdf5.la
├── libhdf5.settings
├── libhdf5.so -> libhdf5.so.100.0.0
├── libhdf5.so.100 -> libhdf5.so.100.0.0
├── libhdf5.so.100.0.0
├── libhdf5_cpp.a
├── libhdf5_cpp.la
├── libhdf5_cpp.so -> libhdf5_cpp.so.100.0.0
├── libhdf5_cpp.so.100 -> libhdf5_cpp.so.100.0.0
├── libhdf5_cpp.so.100.0.0
├── libhdf5_hl.a
├── libhdf5_hl.la
├── libhdf5_hl.so -> libhdf5_hl.so.100.0.0
├── libhdf5_hl.so.100 -> libhdf5_hl.so.100.0.0
├── libhdf5_hl.so.100.0.0
├── libhdf5_hl_cpp.a
├── libhdf5_hl_cpp.la
├── libhdf5_hl_cpp.so -> libhdf5_hl_cpp.so.100.0.0
├── libhdf5_hl_cpp.so.100 -> libhdf5_hl_cpp.so.100.0.0
└── libhdf5_hl_cpp.so.100.0.0
```

Also, `libaec` could be removed also as it is needed by `libsz`.

glistening:
@mhs4670go 

We may opt out `libsz` by building `hdf5` by ourselves without `sz` if its license could be problematic.

```sh
$ mkdir build_x86_64 && cd build_x86_64

# you don't need --without-szlib because it is off by default
$ ../configure --enable-cxx

Linking Options:
----------------
...
                Extra libraries: -lz -ldl -lm 
```

Note there is no `-lsz`.

```
$ make all install
$ tree hdf5/lib
├── libhdf5.a
├── libhdf5.la
├── libhdf5.settings
├── libhdf5.so -> libhdf5.so.100.0.0
├── libhdf5.so.100 -> libhdf5.so.100.0.0
├── libhdf5.so.100.0.0
├── libhdf5_cpp.a
├── libhdf5_cpp.la
├── libhdf5_cpp.so -> libhdf5_cpp.so.100.0.0
├── libhdf5_cpp.so.100 -> libhdf5_cpp.so.100.0.0
├── libhdf5_cpp.so.100.0.0
├── libhdf5_hl.a
├── libhdf5_hl.la
├── libhdf5_hl.so -> libhdf5_hl.so.100.0.0
├── libhdf5_hl.so.100 -> libhdf5_hl.so.100.0.0
├── libhdf5_hl.so.100.0.0
├── libhdf5_hl_cpp.a
├── libhdf5_hl_cpp.la
├── libhdf5_hl_cpp.so -> libhdf5_hl_cpp.so.100.0.0
├── libhdf5_hl_cpp.so.100 -> libhdf5_hl_cpp.so.100.0.0
└── libhdf5_hl_cpp.so.100.0.0
```

Also, `libaec` could be removed also as it is needed by `libsz`.

mhs4670go:
@glistening You're right. Actually, these drat is very workaround. What you said applies in https://github.com/Samsung/ONE/pull/2989 :)

Issue #2968, opened 2020-07-08
luci-interpter with empty CircleConst


seanshpark:
Current `Net_ZeroDim_001` is excluded in `common-artifacts` module with a problem in _luci_

problem:
> compiler/luci/lang/src/Nodes/CircleConst.cpp:42: const typename loco::DataTypeImpl<DT>::Type& luci::CircleConst::at(uint32_t) const [with loco::DataType DT = (loco::DataType)10; typename loco::DataTypeImpl<DT>::Type = float; uint32_t = unsigned int]: Assertion `n < size<DT>()' failed.

Let's fix this problem

seanshpark:
this recipe has
```
operand {
  name: "perm"
  type: INT32
  shape {
    dim: 4
  }
  filler {
    tag: "explicit"
    arg: "0"
    arg: "2"
    arg: "1"
    arg: "3"
  }
}
```

seanshpark:
To reproduce
```
cd build/debug/compiler/common-artifacts
./testDataGenerator Net_ZeroDim_001.circle
```

seanshpark:
As `Net_ZeroDim_001` has operators that current _luci-interpreter_ does not support, simple Add would be better

This as `Add_001/test.recipe`
```
operand {
  name: "ifm1"
  type: FLOAT32
  shape { dim: 1 dim: 0 dim: 1 }
}
operand {
  name: "ifm2"
  type: FLOAT32
  shape { dim: 1 dim: 0 dim: 1}
  filler {
    tag: "explicit"
  }
}
operand {
  name: "ofm"
  type: FLOAT32
  shape { dim: 1 dim: 0 dim: 1 }
}
operation {
  type: "Add"
  input: "ifm1"
  input: "ifm2"
  output: "ofm"
  add_options {
    activation: NONE
  }
}
input: "ifm1"
output: "ofm"
```

seanshpark:
@s-barannikov , can you please check with the draft #2971 ?
And possible, can you provide any guides to fix this? or it is OK if you fix this issue or point out it's another modules problem.

get draft and apply this change
```diff

index 4e7fab2..198f254 100644
--- a/compiler/common-artifacts/exclude.lst
+++ b/compiler/common-artifacts/exclude.lst
@@ -7,7 +7,7 @@ optimize(Where_001)

 # tcgenerate : Exclude from test data generation(TestDataGenerator)
 tcgenerate(Abs_000)
[...]

PR #2971, opened 2020-07-08
DRAFT empty constant issue with luci-interpreter

seanshpark:
on-going draft to fix empty constant issue with luci-interpreter


seanshpark:
For #2968

compiler/luci-interpreter/src/core/Tensor.cpp
```diff
@@ -36,6 +36,8 @@ void Tensor::readData(void *data_ptr, size_t data_size) const
 {
   const size_t element_size = getDataTypeSize(element_type());
   const int32_t num_elements = shape().num_elements();
+  if (num_elements <= 0)
```
s-barannikov, line 39: I would instead assert that `data_ptr != nullptr && data_size != 0` after checking `data_size != num_elements * element_size` below. IMHO it is incorrect to call this function for empty tensor.

compiler/luci-interpreter/src/core/Tensor.cpp
```diff
@@ -47,6 +49,8 @@ void Tensor::writeData(const void *data_ptr, size_t data_size)
...
```
s-barannikov, line 52: Same

compiler/luci-interpreter/src/core/Tensor.cpp
```diff
@@ -59,6 +63,8 @@ void Tensor::resize(const Shape &new_shape)
   _shape = new_shape;
   const size_t element_size = getDataTypeSize(_element_type);
   const int32_t num_elements = _shape.num_elements();
+  if (num_elements <= 0)
+    return;
```
s-barannikov, line 67: Could you please add a comment near the `_data` field that it will be nullptr for empty tensors?

seanshpark:
@s-barannikov , thanks for the review and I've update the change. please take a look again :)

seanshpark:
all landed

PR #2974, opened 2020-07-08
[luci] Update FuseBCQPass

llFreetimell:
This commit will update following things
- Add a new rule for finding prefix of node name
- Handle when `do_w_x` is `int32` type

### DIFF
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -125,7 +125,7 @@ public:
 
     if (_do_w_x[prefix]->dtype() == loco::DataType::S32)
       return _do_w_x[prefix]->at<loco::DataType::S32>(0) == 1;
-    else if(_do_w_x[prefix]->dtype() == loco::DataType::BOOL)
+    else if (_do_w_x[prefix]->dtype() == loco::DataType::BOOL)
       return _do_w_x[prefix]->at<loco::DataType::BOOL>(0);
     else
       throw std::runtime_error("do_w_x should be int or bool");
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -38,6 +38,16 @@ const std::string node_name_prefix(luci::NodeName node_name)
 {
   std::string prefix = node_name;
 
+  if (prefix.find("ReadVariableOp/resource/") != std::string::npos)
+  {
+    const auto start_index = prefix.find("ReadVariableOp/resource/");
+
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -38,6 +38,16 @@ const std::string node_name_prefix(luci::NodeName node_name)
 {
   std::string prefix = node_name;
 
+  if (prefix.find("ReadVariableOp/resource/") != std::string::npos)
+  {
+    const auto start_index = prefix.find("ReadVariableOp/resource/");
```
seanshpark, line 43: As `"ReadVariableOp/resource/"` appears twice, how about adding a variable const std::string or const char * for this?

compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -48,6 +58,11 @@ const std::string node_name_prefix(luci::NodeName node_name)
     const auto index = prefix.find("kernel/");
     prefix = prefix.substr(0, index - 1);
   }
+  else if (prefix.find("/bcqinfo_") != std::string::npos)
+  {
+    const auto index = prefix.find("/bcqinfo_");
```
seanshpark, line 63: same here

seanshpark: for `"Tensordot/"` we can change it in another PR if you accept

compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -107,7 +122,11 @@ class BCQConverter final
   bool do_w_x(luci::CircleConst *node)
   {
     const auto prefix = node_name_prefix(node->name());
-    return _do_w_x[prefix]->at<loco::DataType::BOOL>(0);
+
+    if (_do_w_x[prefix]->dtype() == loco::DataType::S32)
+      return _do_w_x[prefix]->at<loco::DataType::S32>(0) == 1;
+    else
+      return _do_w_x[prefix]->at<loco::DataType::BOOL>(0);
```
seanshpark, line 129: maybe adding an assert for `_do_w_x[prefix]->dtype()` be `loco::DataType::S32` or `loco::DataType::BOOL` ?

compiler/luci/pass/src/FuseBCQPass.cpp
llFreetimell: Then, I will change all the things in another PR for consistency :)

llFreetimell:
@seanshpark Because of format-checker fail, please PTAL again.. sorry :(

Issue #2951, opened 2020-07-07
Compiler FE: Enable SelectV2 Op


ai-moiseev:
- [x] res/TensorFlowPythonExamples #2976 
- [x] tflchef #3149 
- [x] tflite2circle #3148 
- [x] tfldump #3150 
- [x] circledump #3151 
- [x] res/TensorFlowLiteRecipes #3190 
- [x] luci.IR #3153 
- [x] luci.Import #3184
- [x] luci.Export #3185 
- [x] luci.logex  #3181 
- [x] luci.Service #3183 
- [x] luci.tests #3187 
- [x] common artifacts #3152 

For #1606
Draft #3091

lemmaa:
@ai-moiseev , Alexander Moiseev, welcome to join the project. :)

binarman:
@ai-moiseev
don't forget to disable interpreter tests in common-artifacts,
without it you are not able to commit recipes.

binarman:
Everything is done, closing

PR #2976, opened 2020-07-08
[res] Add example for WhereV2 operator

ai-moiseev:
For #2951 


binarman:
I think it will be better to name PR `[res] Add example for WhereV2 operator` because this PR about TensorFlow python API. This API provides WhereV2 operator, not Select

PR #2982, opened 2020-07-08
[luci-interpreter] Fix for empty tensors

seanshpark:
This will fix for empty tensors; where one of dimension in tensor is zero


seanshpark:
For #2968, from draft #2971

PR #2985, opened 2020-07-09
[res] Add recipe with zero size tensor

seanshpark:
This will introduce Add_001 recipe with zero size tensor


seanshpark:
For #2968, from draft #2971

seanshpark:

PR #2986, opened 2020-07-09
[luci] enable test of Add_001

seanshpark:
This will enable test of Add_001 having zero size tensor


seanshpark:
For #2968, from draft #2971

Issue #3001, opened 2020-07-09
Compiler FE: Introduce argument parser


mhs4670go:
We are currently implementing a small argument parser in the driver of each module.

However, the implementation of this is different module by module, and it is difficult to maintain.

Recently, a driver using the boost library came in, which complicated the release process.

So, I introduce an argument parser that is very lightweight and easy to use:)

mhs4670go:
This is an example of arser.
```cxx
Arser arser;

arser.add_argument("--input_path")
      .nargs(1)
      .type(arser::DataType::STR)
      .help("input path of this program.")
      .required();
arser.add_argument("--output_path")
      .nargs(1)
      .type(arser::DataType::STR)
      .help("output path of this program.")
      .required(true);
arser.add_argument("--training_data")
      .nargs(5)
      .type(arser::DataType::INT32_VEC)
      .help("give traning data to this program.")
      .required();

Prompt prompt("./ml --input_path /I/am/in.put --output_path I/am/out.put "
                "--training_data 2 43 234 3 334");
/* act */
arser.parse(prompt.argc(), prompt.argv());
/* assert */
EXPECT_TRUE(arser["--input_path"]);
EXPECT_EQ("/I/am/in.put", arser.get<std::string>("--input_path"));
EXPECT_TRUE(arser["--output_path"]);
EXPECT_EQ("I/am/out.put", arser.get<std::string>("--output_path"));
EXPECT_TRUE(arser["--training_data"]);
std::vector<int32_t> data = arser.get<std::vector<int32_t>>("--training_data");
EXPECT_EQ(2, data.at(0));
EXPECT_EQ(43, data.at(1));
EXPECT_EQ(234, data.at(2));
EXPECT_EQ(3, data.at(3));
EXPECT_EQ(334, data.at(4));
```

PR #3002, opened 2020-07-09
[arser] Introduce arser

mhs4670go:
This commit introduces arser to our compiler

Related : #3001 

### DIFF
--- a/compiler/arser/tests/arser.test.cpp
+++ b/compiler/arser/tests/arser.test.cpp
@@ -1,0 +1,179 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


jinevening:
Overall, it looks good. I think `boost::program_option` can be replaced with `arser`.

(optional) Supporting a default value would make arser even better.

compiler/arser/tests/arser.test.cpp
```diff
@@ -0,0 +1,179 @@
...
+  arser.add_argument("--output_path")
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .help("output path of this program.")
+      .required(true);
+  arser.add_argument("--training_data")
+      .nargs(5)
+      .type(arser::DataType::INT32_VEC)
+      .help("give traning data to this program.")
+      .required();
```
parjong, line 161: Looks like a C++ version of Python argparse. Am I right?

mhs4670go: Rigth:) That looks simple.

PR #3012, opened 2020-07-09
DRAFT: Make record-minmx use arser instead of boost

jinevening:
On going draft to make record-minmax use arser instead of boost::program_option

---
Related to #2809


jinevening:
To merge this draft without violating our rule (no simultaneous changes to multiple modules in one PR), five PRs are needed.

1. disable record-minmax-conversion-test

2. disable pota-quantization-value-test

3. change record-minmax to use arser

4. enable record-minmax-conversion-test

5. enable pota-quantization-value-test

seanshpark:
If `record-minmax` argument is going to be changed, you need to change related drivers using this.
- tf2nnpkg and others

PR #3026, opened 2020-07-10
[onert] Remove the member `_constants` from TensorBuilders

### LINKED ISSUE #2996: [onert] performance regression after #2894
I've found that my target model takes 2.5x time (0.55 ms →1.4 ms).

After several measuring with recent changes, I've found it is caused by #2894.

Before #2894, 0.52 ms
After #2894, 1.36 ms

@ragmani Could you please take a look?

cc @wateret, @hseok-oh

ragmani:
Fix #2996

This commit removes the member `_constants` from TensorBuilders.
  - Make TensorBuilders uses `isConstant` of OperandInfo instead of `_constants`
  - Fix wrong checking constant of cpu tensor


glistening:
@ragmani I took a look for the PR content. LGTM for content also.

PR #3036, opened 2020-07-10
DRAFT one-cmds

seanshpark:
on-going draft to provide one- CLI commands


seanshpark:
done for now

PR #3040, opened 2020-07-10
[onert] renaming StaticInferer and DynamicInferer

hyunsik-yoon:
`StaticInfer` is renamed to `StaticShapeInfer` and `DynamicInferer` to `DynamicShapeInferer`.

Previously, their full name (with namespace) was `util::shape_inference::StaticInferer` (namespace has `Shape`)
so their name does not need to have `"Shape"`. 
Currently, their full name is `compiler::SaticInferer` so name without `"Shape"` seems not very specific.


ragmani:

wateret:
Oops I made a conflict. Please rebase 😅

PR #3041, opened 2020-07-10
[onert] Move random input generation into randomgen

glistening:
It moves random input generation code into randomgen.*.
Also, it cleans up trivials - put header in order, removing blank lines.

### DIFF
--- a/tests/tools/nnpackage_run/src/h5formatter.cc
+++ b/tests/tools/nnpackage_run/src/h5formatter.cc
@@ -18,10 +18,8 @@
 #include "nnfw.h"
 #include "nnfw_util.h"
 
-#include <cassert>
 #include <iostream>
 #include <stdexcept>
-#include <cstdlib>
 #include <H5Cpp.h>
 
 namespace nnpkg_run
@@ -31,7 +29,7 @@ static const char *h5_value_grpname = "value";
 void H5Formatter::loadInputs(const std::string &filename, std::vector<Allocation> &inputs)
 {
   uint32_t num_inputs;
-  NNPR_ENSURE_STATUS(nnfw_input_size(session, &num_inputs));
+  NNPR_ENSURE_STATUS(nnfw_input_size(session_, &num_inputs));
   try
   {
     // Turn off the automatic error printing.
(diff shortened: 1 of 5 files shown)


tests/tools/nnpackage_run/src/h5formatter.h
```diff
@@ -29,12 +29,12 @@ namespace nnpkg_run
...
 {
 public:
-  H5Formatter(nnfw_session *sess) : session(sess) {}
+  H5Formatter(nnfw_session *sess) : session_(sess) {}
   void loadInputs(const std::string &filename, std::vector<Allocation> &inputs);
   void dumpOutputs(const std::string &filename, std::vector<Allocation> &outputs);
 
 private:
-  nnfw_session *session;
+  nnfw_session *session_;
```
glistening, line 37: I will create other PR for renaming `session` to `session_` if reviewer prefers.

tests/tools/nnpackage_run/src/h5formatter.cc
```diff
@@ -18,10 +18,8 @@
 #include "nnfw.h"
 #include "nnfw_util.h"
 
-#include <cassert>
 #include <iostream>
 #include <stdexcept>
-#include <cstdlib>
```
glistening, line 24: Removing unnecessary headers. It could be in other PR if reviewer prefers

tests/tools/nnpackage_run/src/nnpackage_run.cc
```diff
@@ -32,22 +32,11 @@
 #include <chrono>
 #include <cstdlib>
 #include <iostream>
+#include <libgen.h>
```
glistening, line 35: Putting headers in order could be another PR if reviewer prefers

YongseopKim:
The title would be better if [onert] -> [nnpkg_run].

tests/tools/nnpackage_run/src/randomgen.cc
```diff
@@ -0,0 +1,74 @@
...
+template <class T> void randomData(nnfw::misc::RandomGenerator &randgen, void *data, uint64_t size)
+{
+  for (uint64_t i = 0; i < size; i++)
+    reinterpret_cast<T *>(data)[i] = randgen.generate<T>();
+}
+
+void RandomGenerator::generate(std::vector<Allocation> &inputs)
+{
+  // generate random data
+  const int seed = 1;
```
YongseopKim, line 36: How about changing `seed` not static but dynamic? I mean such as time()...

glistening: @YongseopKim Thank you for comment.

First, it is beyond this PR.

Second, there has been discussion about constant seed vs. changing seed. I am for const since it gives deterministic result. If we use different seed for each CI run, it is hard to know the regression comes from different seed or PR itself.

If someone or we would like to strengthen the test by testing more input values (it is good), we may add additional argument like #2622, and run several times with different seed daily between midnight to dawn when CI is likely to be idle.

PR #3042, opened 2020-07-10
[onert] Make FunctionSequence::run() handle dynamic tensor

hyunsik-yoon:
This makes `FunctionSequence::run()` handle dynamic tensor.

Please refer to #2943 for the background. 
In summary,
- we should decide if we'll call dynamic shape inferer or not **at execution time**.
- previous way (deciding `FunctionSequence` or `FunctionSequenceForDynamicTensor` at compilation time) does not work.

So, I'd like to use `FunctionSequence` only, and remove `FunctionSequenceForDynamicTensor`.

draft: #2957

### DIFF
--- a/runtime/onert/core/include/exec/FunctionSequence.h
+++ b/runtime/onert/core/include/exec/FunctionSequence.h
@@ -72,10 +72,70 @@ public:
     }
   }
 
+public: // context for dynamic tensor
+  /**
+   * @brief Prepare to run FunctionSequence which "might" handle dynamic tensor
+   * @note  Calling this does not mean that run() will handle dynamic tensor.
+   *        enableDynamicShapeInferer(true) will make run() will handle dynamic tensor.
+   */
+  void dynamicTensorContext(const ir::OpSequence *op_seq, const ir::Operations *operations,
+                            std::unique_ptr<exec::DynamicInferer> dynamic_shape_inferer,
+                            std::shared_ptr<backend::ITensorRegistry> tensor_registry,
+                            backend::IDynamicTensorManager *dynamic_tensor_manager)
+  {
+    _op_seq = op_seq;
+    _operations = operations;
+    _dynamic_shape_inferer = std::move(dynamic_shape_inferer);
+    _tensor_registry = tensor_registry;
+    _dynamic_tensor_manager = dynamic_tensor_manager;
+  }
+
+  /**
+   * @brief Call this function if this FunctionSequence handles dynamic tensors.
+   * @note When dynamic tensors are handled, enableDynamicShapeInferer(true) must be called before
+   *       run(). If not called, run() assumes that all tensors are static.
+   */
+  void enableDynamicShapeInferer(bool enable) { _enable_dynamic_shape_inferer = enable; }
+
+  const ir::OpSequence *opSeq()
+  {
+    assert(_op_seq);
+    return _op_seq;
+  }
+
+  const ir::Operations *operations()
+  {
+    assert(_operations);
+    return _operations;
+  }
+
+  std::shared_ptr<backend::ITensorRegistry> &tensorRegistry()
+  {
+    assert(_tensor_registry.get());
+    return _tensor_registry;
+  }
+
+  backend::IDynamicTensorManager *dynamicTensorManager()
+  {
+    assert(_dynamic_tensor_manager);
+    return _dynamic_tensor_manager;
+  }
+
 protected:
   std::vector<std::unique_ptr<IFunction>> _functions;
+
+protected: // context to run this sequence when this sequence may handle dynamic tensors
+  bool _enable_dynamic_shape_inferer = false;
+  const ir::OpSequence *_op_seq = nullptr;
+  const ir::Operations *_operations = nullptr;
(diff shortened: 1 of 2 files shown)


wateret:
I'm taking a look at the draft first.

runtime/onert/core/include/exec/FunctionSequence.h
```diff
@@ -72,10 +72,70 @@ class FunctionSequence : public IFunction
...
+public: // context for dynamic tensor
+  /**
+   * @brief Prepare to run FunctionSequence which "might" handle dynamic tensor
+   * @note  Calling this does not mean that run() will handle dynamic tensor.
+   *        enableDynamicShapeInferer(true) will make run() will handle dynamic tensor.
+   */
+  void dynamicTensorContext(const ir::OpSequence *op_seq, const ir::Operations *operations,
+                            std::unique_ptr<exec::DynamicInferer> dynamic_shape_inferer,
+                            std::shared_ptr<backend::ITensorRegistry> tensor_registry,
+                            backend::IDynamicTensorManager *dynamic_tensor_manager)
```
wateret, line 84: - It should be named `dynamic_tensor_context` or `setDynamicTensorContext` according to our coding convention
- What about packing 5 argument as a struct?
- What about moving this to the constructor? Calling this function seems like a must

hyunsik-yoon: @wateret, thanks for the comment.
1. Do you mean the name of method? I spent time to search our naming convention but it was hard to find. To person like me who recently joined in onert, the coding convention info is hard to find. :crying_cat_face: 
   - Considering your comment, if i want to create a method to _get a big cat_, I wonder you meant that the following follows the right convention?
      - `Cat big_cat(void) { .. }`
      - `Cat getBigCat(void) { .. }`
2. Considering your comment, how about storing the last three params (`dynamic_shape_inferer`, `tensor_registry`, `dynamic_tensor_manager`) into `BackendContext`? by passing `BackendContext` into the ctor of `KernelConstructor`, we don't need to pass these three here. 
3. Regarding ctor of `FunctionSequence`... In acl, `KernelGenerator` use the following ctor:
   - https://github.com/Samsung/ONE/blob/master/runtime/onert/backend/acl_cl/KernelGenerator.cc#L229-L230
      - too many ctors make usage of `FunctionSequence` too complex.
   - I propose the following change into `FunctionSequence`:
      - ctor `FunctionSequence()`  // for backend supporting static tensors only
      - ctor `FunctionSequence(struct DynamicTensorContext)` // for backend supporting dynamic tensors
      - method `append(arg1, arg2, arg3, ...)` which will replace previous ctor `template <typename... Args> FunctionSequence(Args &&... args) { initialize(std::move(args)...); }`

wateret: > 1. Do you mean the name of method? I spent time to search our naming convention but it was hard to find. To person like me who recently joined in onert, the coding convention info is hard to find. 😿
>    
>    * Considering your comment, if i want to create a method to _get a big cat_, I wonder you meant that the following follows the right convention?
>      
>      * `Cat big_cat(void) { .. }`
>      * `Cat getBigCat(void) { .. }`

Both are correct. I think the whole team(not only runtime) talked about this long ago, .

- All function and method names must be camelCase
- (Exception) For getters and setters, they also can be the same with the field name without prefix `_`
    - In a class, the naming rule must be consistent(Do not mix up those two ways in a class)

> 2. Considering your comment, how about storing the last three params (`dynamic_shape_inferer`, `tensor_registry`, `dynamic_tensor_manager`) into `BackendContext`? by passing `BackendContext` into the ctor of `KernelConstructor`, we don't need to pass these three here.
> 3. Regarding ctor of `FunctionSequence`... In acl, `KernelGenerator` use the following ctor:
>    
>    * https://github.com/Samsung/ONE/blob/master/runtime/onert/backend/acl_cl/KernelGenerator.cc#L229-L230
>      
>      * too many ctors make usage of `FunctionSequence` too complex.
>    * I propose the following change into `FunctionSequence`:
>      
>      * ctor `FunctionSequence()`  // for backend supporting static tensors only
>      * ctor `FunctionSequence(struct DynamicTensorContext)` // for backend supporting dynamic tensors
>      * method `append(arg1, arg2, arg3, ...)` which will replace previous ctor `template <typename... Args> FunctionSequence(Args &&... args) { initialize(std::move(args)...); }`

Ok then I'm good with the as-is way 😄  I'm just worried about have too many states in a class.

hyunsik-yoon: @wateret, thanks for the info about convention. I am now crystal-clear about the convention. I will fix.

>  I'm just worried about have too many states in a class.
I don't like many states either. So I tried to pull `run()` out of `FunctionSequence`... however, changing acl code (`KernelGenerator` builds sort of depth-1 tree of `FunctionSequence`) was difficult.

>> I propose the following change into FunctionSequence...
Let me try to fix this change in next PRs.

hyunsik-yoon: Name was fixed. PTAL.

PR #3044, opened 2020-07-10
[onert] Find a model has dynamic tensor at compilation time

hyunsik-yoon:
For #2943 
Related PR: #3042
draft: #2957

This commit make `StaticShapeInferer` (at compilation time) find if a model has dynamic tensor or not and keep the boolean flag into `OpSequence`.

If a model has dynamic tensor (e.g., `reshape` op's output tensor if the 2nd input is not const),
executor will use this flag (in next PR) to run `DynamicShapeInferer` at execution time.

Since `FunctionSequence` (generated from `OpSequence`) is the unit of `run()`,
I put `OpSequence` to keep this flag.

### DIFF
--- a/runtime/onert/core/include/ir/OpSequence.h
+++ b/runtime/onert/core/include/ir/OpSequence.h
@@ -81,8 +81,8 @@ public:
    *        or dynamic output;
    *        @c false if all operations' inputs and outputs are static tensors
    */
-  void hasDynamicTensor(bool hasDynamicTensor) { _has_dynamic_tensor = hasDynamicTensor; }
-  bool hasDynamicTensor() const { return _has_dynamic_tensor; }
+  void has_dynamic_tensor(bool has_dynamic_tensor) { _has_dynamic_tensor = has_dynamic_tensor; }
+  bool has_dynamic_tensor() const { return _has_dynamic_tensor; }
 
 private:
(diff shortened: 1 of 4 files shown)


runtime/onert/core/include/ir/OpSequence.h
```diff
@@ -75,13 +75,25 @@ class OpSequence
...
   std::vector<OperationIndex>::const_iterator end() const { return _operations.end(); }
 
+public:
+  /**
+   * @brief Set @c true if any operation in this opSequence has dynamic input
+   *        or dynamic output;
+   *        @c false if all operations' inputs and outputs are static tensors
+   */
+  void hasDynamicTensor(bool hasDynamicTensor) { _has_dynamic_tensor = hasDynamicTensor; }
+  bool hasDynamicTensor() const { return _has_dynamic_tensor; }
```
wateret, line 85: The name of Getter and Setter should be `has_dynamic_tensor` according to our coding convention.

hyunsik-yoon: This will be fixed after getting answer for https://github.com/Samsung/ONE/pull/3042#discussion_r454075584.

hyunsik-yoon: Fixed. Thanks.

PR #3047, opened 2020-07-10
[record-minmax] Use percentile instead of moving avg

jinevening:
This uses percentile for recording instead of moving avg

### DIFF
--- a/compiler/record-minmax/include/RecordFunction.h
+++ b/compiler/record-minmax/include/RecordFunction.h
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <vector>
+#include <cassert>
+#include <algorithm>
+#include <cmath>
+#include <numeric>
+#include <stdexcept>
+
+namespace record_minmax
+{
+
+/**
+ * @brief  getNthPercentile calculates the n-th percentile of input vector (0.0 <= n <= 100.0)
+ *         linear interpolation is used when the desired percentile lies between two data points
+ */
+float getNthPercentile(std::vector<float> &vector, float percentile)
+{
+  assert(!vector.empty());
+
+  if (percentile < 0 || percentile > 100)
+    throw std::runtime_error("Percentile must be ranged from 0 to 100");
+
+  if (percentile == 0.0)
+    return vector.front();
+
+  if (percentile == 100.0)
+    return vector.back();
+
+  std::vector<float> copy;
+  copy.assign(vector.begin(), vector.end());
+  std::sort(copy.begin(), copy.end());
+
+  int i = static_cast<int>(std::floor((copy.size() - 1) * percentile / 100.0));
+
+  float percent_i = static_cast<float>(i) / static_cast<float>(copy.size() - 1);
+  float fraction = (percentile / 100.0 - percent_i) / ((i + 1.0) / (copy.size() - 1.0) - percent_i);
+  float res = copy[i] + fraction * (copy[i + 1] - copy[i]);
+  return res;
+}
+
+/**
+ * @brief  getMovingAverage calculates the weighted moving average of input vector
+ *         The initial value is the first element of the vector
+ */
+float getMovingAverage(std::vector<float> &vector, float alpha)
+{
+  assert(!vector.empty());
+  assert(alpha >= 0.0 && alpha <= 1.0);
+
+  float curr_avg = vector[0];
+  for (size_t i = 1; i < vector.size(); i++)
+  {
+    curr_avg = curr_avg * alpha + vector[i] * (1.0 - alpha);
+  }
+  return curr_avg;
+}
+
+} // namespace record_minmax
--- a/compiler/record-minmax/include/RecordFunction.h
+++ b/compiler/record-minmax/include/RecordFunction.h
@@ -30,8 +30,6 @@ namespace record_minmax
  */
 float getNthPercentile(std::vector<float> &vector, float percentile)
 {
-  assert(!vector.empty());
-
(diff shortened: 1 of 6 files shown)


jinevening:
The test data were generated using `numpy.percentile`.

compiler/record-minmax/include/RecordFunction.h
```diff
@@ -0,0 +1,75 @@
...
+  if (percentile == 100.0)
+    return vector.back();
+
+  std::vector<float> copy;
+  copy.assign(vector.begin(), vector.end());
+  std::sort(copy.begin(), copy.end());
+
+  int i = static_cast<int>(std::floor((copy.size() - 1) * percentile / 100.0));
+  int clamped_i = std::max(0, std::min(static_cast<int>(copy.size() - 2), i));
+
```
meejeong, line 50: @jinevening , I can't understand L48-L49 meaning

jinevening: Sorry, it was unnecessary. I've updated the code.

compiler/record-minmax/include/RecordFunction.h
```diff
@@ -0,0 +1,73 @@
...
+    return vector.front();
+
+  if (percentile == 100.0)
+    return vector.back();
+
+  std::vector<float> copy;
+  copy.assign(vector.begin(), vector.end());
+  std::sort(copy.begin(), copy.end());
+
+  int i = static_cast<int>(std::floor((copy.size() - 1) * percentile / 100.0));
```
seanshpark, line 48: (optional) it would be better to use variable with a meaning instead of `i`

compiler/record-minmax/include/RecordFunction.h
```diff
@@ -0,0 +1,73 @@
...
+
+  float percent_i = static_cast<float>(i) / static_cast<float>(copy.size() - 1);
```
seanshpark, line 50: if `copy.size() == 1` this will fail with some system error. but it would be better to show an error message with help message what to fix. from `copy` I don't know what this so I cann't suggest a meaningful message.

compiler/record-minmax/tests/RecordFunction.test.cpp
```diff
@@ -0,0 +1,88 @@
...
+                           -3.23002354,  28.37688474,  -39.56923256, -19.60519504, -35.69124391,
+                           36.72073486,  -32.02635031, -35.33275436, -49.52181541, -32.22861975,
+                           35.68981239,  -27.81677304, 28.81541331,  43.04085581,  -9.7852105};
+
+  EXPECT_FLOAT_NEAR(-47.20005765319782, getNthPercentile(input, 1));
+  EXPECT_FLOAT_NEAR(-42.23149604452366, getNthPercentile(input, 3.14));
+  EXPECT_FLOAT_NEAR(49.40456084968194, getNthPercentile(input, 99));
+}
+
+TEST(GetNthPercentileTest, OutOfBoundary_NEG)
```
seanshpark, line 80: There are several postive test. Can you add some more negative tests?

compiler/record-minmax/include/RecordFunction.h
jinevening: I've changed the name to `index` (I could not think of a better name..).

compiler/record-minmax/include/RecordFunction.h
jinevening: I added a path to handle the single element case.

compiler/record-minmax/tests/RecordFunction.test.cpp
jinevening: I've added one more negative test for handling an empty vector.

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -160,9 +145,17 @@ void RecordMinMax::profileData(const std::string &input_data_path)
...
+    if (mode == "percentile")
+    {
+      min = getNthPercentile(minmax.min_vector, min_percentile);
+      max = getNthPercentile(minmax.max_vector, max_percentile);
+    }
+    else if (mode == "moving_average")
+    {
+      min = getMovingAverage(minmax.min_vector, 0.9);
+      max = getMovingAverage(minmax.max_vector, 0.9);
+    }
```
seanshpark, line 158: if the `mode` is not of the two, `min`/`max` will be undefined. Is the a check code anywhere for `mode` ?

jinevening: I've added a check code. Thanks.

compiler/record-minmax/include/RecordFunction.h
```diff
@@ -0,0 +1,78 @@
...
+ * @brief  getMovingAverage calculates the weighted moving average of input vector
+ *         The initial value is the first element of the vector
+ */
+float getMovingAverage(std::vector<float> &vector, float alpha)
+{
+  assert(!vector.empty());
+  assert(alpha >= 0.0 && alpha <= 1.0);
+
+  float curr_avg = vector[0];
+  for (size_t i = 1; i < vector.size(); i++)
```
oj9040, line 71: If using MovingAverage, you need to compute in a unit of mini-batch. That is because it is highly possible to include the long-tailed range if the batch size =1. 
Here, the size of minibatch is recommended to set 16 based on our experiments.

That is,
min <- alpha*min + (1-alpha)*min(min_batch)
max <- alpha*max + (1-alpha)*max(max_batch)

To implement this, you need double for-loops. In the in-loop you compute min/max of mini-batch and In the out-loop, you update moving average, and

jinevening: @oj9040 Thanks for the comment. 

I agree to use a batch, because it will reduce the sensitivity of min/max values to the order of input dataset. I have several questions about the comments

> it is highly possible to clip too much range if the batch size =1. On the other hand, it is possible to include long-tailed range if the batch size is too large.

AFAIK, moving average is computed using the whole data, not the clipped data.

> To implement this, you need double for-loops. In the in-loop you compute min/max of mini-batch and in the out-loop, you update moving average.

In the in-loop you compute ~~min/max~~ (maybe average?) of mini-batch

+) This PR is for introducing `percentile`, so I'll make another PR to update `getMovingAverage` to use batch :)

oj9040: Moving average is computed using the given whole data.
In in-loop min or max is computed, not average.

jinevening: > In in-loop min or max is computed, not average because moving average of average smooths out too much.

Ok, so just min/max of the minibatch is used and the rest of the values are thrown away. I'll send you a review request when I make the code.

jinevening:
@seanshpark Can you merge this PR? I'll make a new PR for handling @oj9040 's comments.

jinevening:
@seanshpark Can you merge this PR? I'll make another PR for handling @oj9040 's comments.

PR #3048, opened 2020-07-10
[onert] Add quant8 support for maximum cker

hasw7569:
Add quant8 support for maximum cker

ONE-DCO-1.0-Signed-off-by: Leehansol <hansolu.lee@samsung.com>

### DIFF
--- a/runtime/onert/backend/cpu/ops/MaxLayer.cc
+++ b/runtime/onert/backend/cpu/ops/MaxLayer.cc
@@ -29,21 +29,25 @@ namespace cpu
 namespace ops
 {
 
-void MaxLayer::maxFloat32()
+template <typename T> void MaxLayer::maximum()
 {
-  nnfw::cker::Max<float>(getTensorShape(_lhs), reinterpret_cast<const float *>(_lhs->buffer()),
-                         getTensorShape(_rhs), reinterpret_cast<const float *>(_rhs->buffer()),
-                         getTensorShape(_output), reinterpret_cast<float *>(_output->buffer()));
+  nnfw::cker::Max<T>(getTensorShape(_lhs), reinterpret_cast<const T *>(_lhs->buffer()),
+                     getTensorShape(_rhs), reinterpret_cast<const T *>(_rhs->buffer()),
+                     getTensorShape(_output), reinterpret_cast<T *>(_output->buffer()));
 }
 
 void MaxLayer::maxQuant8()
 {
-  // TODO Check whether cker for quant8 max produces correct results
-  // nnfw::cker::Max<uint8_t>(
-  //     getTensorShape(_lhs), reinterpret_cast<const uint8_t*>(_lhs->buffer()),
-  //     getTensorShape(_rhs), reinterpret_cast<const uint8_t*>(_rhs->buffer()),
-  //     getTensorShape(_output), reinterpret_cast<uint8_t*>(_output->buffer()));
-
+  if (_lhs->data_scale() == _rhs->data_scale() && _lhs->data_scale() == _output->data_scale())
+  {
+    if (_lhs->data_offset() == _rhs->data_offset() && _lhs->data_offset() == _output->data_offset())
+    {
+      return nnfw::cker::Max<uint8_t>(
+          getTensorShape(_lhs), reinterpret_cast<const uint8_t *>(_lhs->buffer()),
+          getTensorShape(_rhs), reinterpret_cast<const uint8_t *>(_rhs->buffer()),
+          getTensorShape(_output), reinterpret_cast<uint8_t *>(_output->buffer()));
+    }
+  }
   throw std::runtime_error("Max NYI for quantized");
 }
 
@@ -63,7 +67,7 @@ void MaxLayer::run()
 {
   if (_lhs->data_type() == OperandType::FLOAT32)
   {
-    maxFloat32();
+    maximum<float>();
   }
   else if (_lhs->data_type() == OperandType::QUANT_UINT8_ASYMM)
   {
--- a/tests/nnapi/specs/V1_2/maximum_quant8_nnfw.mod.py
+++ b/tests/nnapi/specs/V1_2/maximum_quant8_nnfw.mod.py
@@ -1,0 +1,34 @@
+#
+# Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+# Copyright (C) 2020 The Android Open Source Project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+i1 = Input("input0", "TENSOR_QUANT8_ASYMM", "{3, 1, 2}, 1.0f, 100")
+i2 = Input("input1", "TENSOR_QUANT8_ASYMM", "{3, 1, 2}, 1.0f, 100")
+i3 = Output("output0", "TENSOR_QUANT8_ASYMM", "{3, 1, 2}, 1.0f, 100")
+
+model = Model().Operation("MAXIMUM", i1, i2).To(i3)
+
+# Example 1. Input in operand 0,
+input0 = {i1: # input 0
+          [129, 127, 125, 149, 123, 124],
+          i2: # input 1
+          [125, 127, 129, 151, 121, 124]}
+
+output0 = {i3: # output 0
+           [129, 127, 129, 151, 123, 124]}
+
+# Instantiate an example
+Example((input0, output0))
--- a/runtime/onert/backend/cpu/ops/MaxLayer.h
+++ b/runtime/onert/backend/cpu/ops/MaxLayer.h
@@ -39,7 +39,7 @@ public:
   }
 
 public:
-  void maxFloat32();
+  template <typename T> void maximum();
 
   void maxQuant8();
 
--- a/tests/nnapi/nnapi_gtest.skip.noarch.interp
+++ b/tests/nnapi/nnapi_gtest.skip.noarch.interp
@@ -259,6 +259,7 @@ GeneratedTests.maximum_broadcast
 GeneratedTests.maximum_broadcast_quant8
 GeneratedTests.maximum_dynamic_nnfw
 GeneratedTests.maximum_overflow
+GeneratedTests.maximum_quant8_nnfw
 GeneratedTests.maximum_simple
 GeneratedTests.maximum_simple_quant8
 GeneratedTests.mean


ragmani:

runtime/onert/backend/cpu/ops/MaxLayer.cc
```diff
@@ -38,12 +38,16 @@ void MaxLayer::maxFloat32()
...
+  {
+    if (_lhs->data_offset() == _rhs->data_offset() && _lhs->data_offset() == _output->data_offset())
+    {
+      return nnfw::cker::Max<uint8_t>(
+          getTensorShape(_lhs), reinterpret_cast<const uint8_t *>(_lhs->buffer()),
+          getTensorShape(_rhs), reinterpret_cast<const uint8_t *>(_rhs->buffer()),
+          getTensorShape(_output), reinterpret_cast<uint8_t *>(_output->buffer()));
+    }
+  }
   throw std::runtime_error("Max NYI for quantized");
```
YiHyunJin, line 51: Please delete it.

runtime/onert/backend/cpu/ops/MaxLayer.cc
```diff
@@ -38,12 +38,16 @@ void MaxLayer::maxFloat32()
 
 void MaxLayer::maxQuant8()
 {
-  // TODO Check whether cker for quant8 max produces correct results
-  // nnfw::cker::Max<uint8_t>(
-  //     getTensorShape(_lhs), reinterpret_cast<const uint8_t*>(_lhs->buffer()),
-  //     getTensorShape(_rhs), reinterpret_cast<const uint8_t*>(_rhs->buffer()),
-  //     getTensorShape(_output), reinterpret_cast<uint8_t*>(_output->buffer()));
-
+  if (_lhs->data_scale() == _rhs->data_scale() && _lhs->data_scale() == _output->data_scale())
```
ragmani, line 41: I think we should make `MaxLayer` supports qasymm8 if `scale` and `offset` are different. Are you going to do it?

runtime/onert/backend/cpu/ops/MaxLayer.cc
ragmani: Probably, it is necessary because above `if` block doesn't support qaunt8 fully.

runtime/onert/backend/cpu/ops/MaxLayer.cc
hasw7569: I agree.
actually we already have testcase with diffrent scale and offset. (**maximum.mod.py** source file)

but according to following [document](https://www.tensorflow.org/lite/performance/quantization_spec).
tensorflow seems to support some cases with the same scale and offset.

But this document looks old.
If I'm wrong, please let me know to fix this PR.

runtime/onert/backend/cpu/ops/MaxLayer.cc
hasw7569: I left this line for cases where the scale and offset are different.
I am not sure if I need to support some cases where the scale and offset are different.
Solving this will remove this line : )

runtime/onert/backend/cpu/ops/MaxLayer.cc
hseok-oh: Tensorflow lite kernel for `Maximum` operation don't support qasymm8 for different `scale` and `offset`. So we may not need to support different `scale` and `offset` for tflite (and circle) model inference now.

If we need to support different `scale` and `offset` to support NNAPI specification and pass nnapi generation test, we should port android nn runtime's kernel, not tensorflow lite kernel. But it is not our requirement.

ragmani: @hasw7569 

> but according to following [document](https://www.tensorflow.org/lite/performance/quantization_spec).
tensorflow seems to support some cases with the same scale and offset.
`
MAXIMUM
  Input 0:
    data_type  : int8
    range      : [-128, 127]
    granularity: per-tensor
  Output 0:
    data_type  : int8
    range      : [-128, 127]
    granularity: per-tensor
  restriction: Input and outputs must all have same scale/zero_point
`
Tensorflow lite kernel for Maximum operation (2.3-rc0 version) supports only 
It seems to describes about int8 quantization . But you implemented to support `QUANT_UINT8_ASYMM`. Both are different quantization.

Until NNAPI V1.2(Android API level 29), Maximum operation supports `TENSOR_QUANT8_ASYMM`, but not `TENSOR_QUANT8_ASYMM_SIGNED`. Our runtime supports until nnapi v1.2 now.
Tensorflow lite kernel for Maximum operation (2.3-rc0 version) supports signed quant8(`TENSOR_QUANT8_ASYMM_SIGNED`).
So, I think we implement these two quantizations separately.
1. Supporting `TENSOR_QUANT8_ASYMM` for the nnapi unitest to work
2. Supporting `TENSOR_QUANT8_ASYMM_SIGNED` for a test tflite model to works

If your goal was to supporting a tflite model, you should implements it as second way.

hseok-oh: @ragmini @hasw7569 

As @ragmini commented, document https://www.tensorflow.org/lite/performance/quantization_spec describes in8 quantization. So to check tensorflow lite's quantization spec, we should check kernel implementation.

But as I commented above, tensorflow lite (2.3-rc0)'s `uint8 qasymm8` kernel also support `maximum` kernel for same `scale` and `offset`. And other operations `uint8 qasymm8` kernel also same spec with `int8`. 

After we defined `nnfw api`, `NNAPI` is not official API for our runtime, so `NNAPI` frontend is optional frontend to use tensorflow lite and test code (nnapi generation tests).

ragmani: > But as I commented above, tensorflow lite (2.3-rc0)'s `uint8 qasymm8` kernel also support `maximum` kernel for same scale and offset.

It is strange. Why does the kernel support only the type for same `scale` and `offset`. Tensorflow lite (2.3-rc)'s nnapi delegate converts `TfLiteQuantizationParams` to `ANEURALNETWORKS_TENSOR_QUANT8_ASYMM` as it is? How  does tensorflow lite makes it has inputs and output of the operation as the same `scale` and `offset`? Does `TOCO` or TFLC make a model with `maximum` operation like that? I don't know it.
Anyway, I think we cannot sure `uint qasymm8` type of `maximum` operation until we see a model that includes the operation in the type. 
@hasw7569 Have you seen any model with `maximum` operation in `uint qasymm8`?


>  And other operations `uint8 qasymm8` kernel also same spec with `int8`.

@hseok-oh 
I disagree with that other operations `uint8 qasymm8` kernel also same spec with `int8`. `uint qasymm8` and `int8` are different spec. All `int8` can use the kernels for `uint qasymm8`, but `uint qasymm8` cannot use the kernels for `int8`. Only when `zero_point` is 128, it is possible.


> After we defined nnfw api, NNAPI is not official API for our runtime, so NNAPI frontend is optional frontend to use tensorflow lite and test code (nnapi generation tests).

I agree with you.

hseok-oh: >How does tensorflow lite makes it has inputs and output of the operation as the same scale and offset? Does TOCO or TFLC make a model with maximum operation like that? I don't know it.

I don't know. But tensorflow lite kernel assume that input & output have same `scale` and `offset`, so maybe `TFLC` make them same. 

https://github.com/tensorflow/tensorflow/blob/99fea8da0d98fb271b60b58cfa5755f2bd430079/tensorflow/lite/kernels/maximum_minimum.cc#L79-L84

```cpp
struct MaximumOp {
  template <typename data_type>
  static data_type op(data_type el1, data_type el2) {
    return el1 > el2 ? el1 : el2;
  }
};
```

https://github.com/tensorflow/tensorflow/blob/99fea8da0d98fb271b60b58cfa5755f2bd430079/tensorflow/lite/kernels/maximum_minimum.cc#L93-L104
```cpp
template <KernelType kernel_type, typename data_type, typename op_type>
void TFLiteOperation(TfLiteContext* context, TfLiteNode* node,
                     const OpContext& op_context) {
  reference_ops::MaximumMinimumBroadcastSlow(
      GetTensorShape(op_context.input1),
      GetTensorData<data_type>(op_context.input1),
      GetTensorShape(op_context.input2),
      GetTensorData<data_type>(op_context.input2),
      GetTensorShape(op_context.output),
      GetTensorData<data_type>(op_context.output),
      op_type::template op<data_type>);
}
```

https://github.com/tensorflow/tensorflow/blob/99fea8da0d98fb271b60b58cfa5755f2bd430079/tensorflow/lite/kernels/internal/reference/maximum_minimum.h#L52-L57
```cpp
    auto maxmin_func = [&](int indexes[N]) {
      output_data[SubscriptToIndex(output_desc, indexes)] =
          op(input1_data[SubscriptToIndex(desc1, indexes)],
             input2_data[SubscriptToIndex(desc2, indexes)]);
    };
    NDOpsHelper<N>(output_desc, maxmin_func);
```

hseok-oh: >I disagree with that other operations uint8 qasymm8 kernel also same spec with int8

`uint8 qasymm8` kernel and `int8` kernel don't share code. `uint8 qasymm8` kernel is just legacy code and `int8` kernel is different code. But they have same spec in document for `int8` when I checked each operations. If `int8` allow different `scale` and `offset`, `uint8 qasymm8` also allow. If `int8` support only same `scale` and `offset`, `uint8 qasymm8` also support same `scale` and `offset.

ragmani: @hseok-oh 

> I don't know. But tensorflow lite kernel assume that input & output have same scale and offset, so maybe `TFLC` make them same.

I hope so.

tests/nnapi/specs/V1_2/maximum_quant8.mod.py
```diff
@@ -0,0 +1,33 @@
+#
```
ragmani, line 1: Could you change this file name to `maximum_quant8_nnfw.mod.py`?
Our rule to create a new nnapi test file is to append `_nnfw`

tests/nnapi/specs/V1_2/maximum_quant8_nnfw.mod.py
```diff
@@ -0,0 +1,33 @@
+#
```
ragmani, line 1: Please add our copyright text here.

ragmani: Please leave android copyright text too.

runtime/onert/backend/cpu/ops/MaxLayer.cc
```diff
@@ -29,21 +29,25 @@ namespace cpu
...
+  {
+    if (_lhs->data_offset() == _rhs->data_offset() && _lhs->data_offset() == _output->data_offset())
+    {
+      return nnfw::cker::Max<uint8_t>(
```
ragmani, line 45: Please remove `return`

hasw7569: If you remove **return**, line 51 is always called.

So I think **return** is necessary

ragmani: Ah I see. I forgot it.

hasw7569:
@dr-venkman @intom @YiHyunJin PTAL

Issue #2939, opened 2020-07-07
[luci-interpreter] Dynamic tensor support


s-barannikov:
Shapes of some tensors can only be inferred if operator's input _data_ is known. For example, the shape of the output tensor of `Transpose` operator depends on the data of the `perm` input tensor.

Currently, the interpreter does not support cases when shapes of output tensors cannot be inferred at `configure` stage. With the example above that means that the interpreter only supports `Transpose` operators which have constant `perm` input tensor.

I counted 36 operators with such "dynamic" tensors in upstream TensorFlow Lite. Below are the most interesting of them.

* `NonMaxSuppression`
  Depending on the number of inputs the operator has two or three outputs. The first output tensor is static, the shapes of the other output tensors depend on the third input tensor (`max_output_size`).
* `ReduceAny`, `ReduceMax`, `ReduceMin`, `ReduceProd`, `Sum`
  The `axis` input tensor affects the shape of the output tensor and also the shapes of one or two temporary tensors.
* `TransposeConv`
  - `weights` input tensor affects the shape of the `transposed_weights` temporary tensor and does not affect the shape of the output tensor. There seems to be a mistake, because the _value_ of `weights` should not affect the shape of `transposed_weights`, the _shape_ should.
* `If`
  - If any of subgraphs has dynamic output tensors, all output tensors of the operator are dynamic.
  - If the shape of an output tensor of the `then` subgraph differs from the shape of the corresponding output tensor of the `else` subgraph, then the corresponding tensor of the operator is dynamic.
* `While`
  Similar to `If` plus the `cond` subgraph can also contain dynamic tensors. Nevertheless, the output tensor of the `cond` subgraph is always a scalar.
* `Unique`
  The output tensor is always dynamic. The operator has to be actually executed in order to calculate the output shape.
* `ReadVariable`
  The output tensor is always dynamic. The data for the tensor is read from "subgraph resources" which we do not currently support.

There are a couple more of interesting operations, with static output shape:

* `Rank`, `Shape`
  The _shape_ of the input tensor affects the _shape_ and the _content_ of the output tensor. That makes the outputs of these operators "constant" in some sense.

s-barannikov:
@seanshpark @jinevening @binarman FYI

binarman:
Is it done?
Or you want to do something more for this task?

s-barannikov:
There are possible performance improvements I would like to implement. I'll close it for now, and reopen or link if needed.

PR #3051, opened 2020-07-11
[luci-interpreter] Dynamic shapes support

s-barannikov:
For issue #2939.

Call `configure` and `execute` methods in interleaved manner.


s-barannikov:
This is the simplest solution. This will slow down repeated execution of the model a bit, because `configure` will be called each time the model is executed.

Issue #3018, opened 2020-07-10
Compile frontend drivers


seanshpark:
WIP

Current frontend comamnds that are generated
- tf2tfliteV2.py
- tflite2circle
- circle2circle
- record-minmax
- circle-quantizer
- model2nnpkg.sh
- tf2nnpkg

User exposed commands to introduce
- one-import ( * -> circle)
   - one-import-tf
   - one-import-tflite
   - one-import-onnx (later)
- one-optimize (circle -> circle)
- one-quantize (circle -> circle)
- one-pack (circle -> nnpkg)
- one-codegen (circle -> binary)

User exposed commands to introduce for several steps at once
- one-compile : import - optimize - quantize
- one-build : import - optimzie - quantize - codegen
- one-build-pack : import - optimzie - quantize - pack
- one [command] 

one
- command driver to invoke commands
- command can be import, optimize, quantize, compile and build

one-import [model type] ...
- first argument
   - tf for TensorFlow model
   - tflite for TensorFlow lite model
- this will invoke tf2tflite.py, tflite2circle

one-optimize
- this will invoke circle2circle

one-quantize
- this will invoke record-minmax, circle-quantizer

one-compile
- this will invoke compiler backend

one-build
- this will call all above

seanshpark:
CC @lemmaa , @parjong , @chunseoklee

seanshpark:
In this month, July, as we have to freeze code in 20th, preparing three one-import, one-optimize and one-quantize seems possible.

lemmaa:
@seanshpark , as we discussed off-line, I agree much with this. :)

seanshpark:
@lemmaa , what about current `tf2nnpkg` ?

seanshpark:
tf2tfliteV2.py needs to run in our TF 2.3.0rc0 (as of now) and to match this, we provide python virtual environment
- restricting user to run our tools in specific virtual environment for precondition doesn't seem to be good

it would be better to provide these commands tools with bash script

lemmaa:
> @lemmaa , what about current `tf2nnpkg` ?

@seanshpark , how about `one build-pack` ? This is a complete replacement for `tf2nnpkg` which takes `tf` as input and creates `nnpackage` as output. Apart from this, we can have a `one pack` or `one pkg` that takes a `circle` as input and makes `nnpackage`.


```
$ dotnet --help
.NET Core SDK (3.1.301)
Usage: dotnet [runtime-options] [path-to-application] [arguments]

[...]

PR #3054, opened 2020-07-13
[one-cmds] Introduce one-cmds for end users

seanshpark:
This will introduce one-cmds module that provides end users comamnd one-import and one-import-tf


seanshpark:
For #3018, from draft #3036

seanshpark:

compiler/one-cmds/one-import-tf
```diff
@@ -0,0 +1,114 @@
...
+rm -rf "${OUTPUT_PATH}.log"
+
+# generate temporary tflite file
+echo "python" "${DRIVER_PATH}/tf2tfliteV2.py" --v2 --input_path ${INPUT_PATH} \
+--input_arrays ${INPUT_ARRAYS} --input_shapes ${INPUT_SHAPES} \
+--output_path "${TMPDIR}/${MODEL_NAME}.tflite" \
+--output_arrays ${OUTPUT_ARRAYS} > "${OUTPUT_PATH}.log"
+echo " " >> "${OUTPUT_PATH}.log"
+
+python "${DRIVER_PATH}/tf2tfliteV2.py" --v2 --input_path ${INPUT_PATH} \
```
llFreetimell, line 102: Simple question..! Why same commands executed twice?
In my sight, printing log is the only difference @.@

seanshpark: which is excuted twice ?

llFreetimell: Ah, my mistake, sorry :(

Issue #2960, opened 2020-07-08
[luci-interpreter] Interface to access tensors in preOperatorExecute and postOperatorExecute


jinevening:
luci-interpreter supports hooks called before/after an operator is executed (`preOperatorExecute` and `postOpeartorExecute`). Currently, both hooks have only one parameter `const luci::CircleNode *`, as follows.

``` cpp
virtual void preOperatorExecute(const luci::CircleNode *node);
virtual void postOperatorExecute(const luci::CircleNode *node);
```

This interface allows very limited use cases, because we cannot access the contents of tensors by using CircleNode.

To allow the access of tensor values inside `preOperatorExecute`/`postOperatorExecute`,  I'd like to add one more parameter (`node_to_tensor`) as follows.

``` cpp
virtual void preOperatorExecute(const luci::CircleNode *node, std::unordered_map<const loco::Node *, Tensor *> &node_to_tensor);
virtual void postOperatorExecute(const luci::CircleNode *node, std::unordered_map<const loco::Node *, Tensor *> &node_to_tensor);
```

jinevening:
@s-barannikov PTAL. I think this may change the interface of many classes in luci-interpreter (especially ~~Loader).

seanshpark:
If `node_to_tensor` doesn't depend on `node` then how about adding something like `preNetworkExecute` to pass `node_to_tensor` once before all the operators?

jinevening:
> how about adding something like preNetworkExecute to pass node_to_tensor once before all the operators?

This is a good idea. AFAIK, `node_to_tensor` is not updated after it is created. My draft (#2961) is adding `node_to_tensor` to pre/postOperatorExecute, but it would be better to make `preNetworkExecute` and pass `node_to_tensor` to it.

jinevening:
I think introducing `preNetworkExecute` and `postNetworkExecute` is a good idea. Currently users cannot access the input/output of the model (pre/postOperatorExecute is not called for CircleInput and CircleOutput).

The parameters of pre/postNetworkExecute would be the inputs of a model (a vector of CircleInput) and the outputs of a model (a vector of CircleOutput), respectively.

s-barannikov:
@jinevening 
[...]

PR #3055, opened 2020-07-13
[luci-interpreter] Add getTensorData to interpreter

jinevening:
This adds getTensorData to interpreter

---
Related to #2960

### DIFF
--- a/compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
@@ -26,6 +26,7 @@
 
 #include <memory>
 #include <vector>
+#include <unordered_map>
 
 namespace luci_interpreter
 {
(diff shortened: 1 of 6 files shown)


jinevening:
Interpreter::getTensorData returns `Tensor` for now, but this will change when Interpreter is decoupled from Tensor.

compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
```diff
@@ -60,11 +61,14 @@ class Interpreter
 
   void attachObserver(ExecutionObserver *observer);
 
+  inline Tensor *getTensorData(const loco::Node *node) { return _node_to_tensor[node]; }
```
s-barannikov, line 64: - `inline` is redundant here. All methods defined within class body are implicitly `inline`.
- Is it necessary to return non-constant `Tensor`?
- Since it now returns `Tensor`, I think the name of the method should be changed to reflect this.

jinevening: > inline is redundant here. All methods defined within class body are implicitly inline.

Thanks.

> Is it necessary to return non-constant Tensor?

No, it is not necessary yet. If we need such an interface, it will be possible to make a new function e.g, `getMutableTensor`.

> Since it now returns Tensor, I think the name of the method should be changed to reflect this.

I've changed the name to `getTensor`. Thanks.

compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
```diff
@@ -60,11 +61,14 @@ class Interpreter
 
   void attachObserver(ExecutionObserver *observer);
 
+  const Tensor *getTensor(const loco::Node *node) { return _node_to_tensor[node]; }
```
s-barannikov, line 64: The method could be `const` but that does not change much.

jinevening:
@nnfw-bot test pr-nncc-debug

jinevening:
@nnfw-bot test all

jinevening:
@seanshpark Can you merge this?

PR #3056, opened 2020-07-13
[tflite2circle] Revise to use arser

mhs4670go:
This commit revises driver to use arser

Related : #3001 


seanshpark:
This change without any options will drop
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Invalid argument. You must have missed some argument.
Aborted (core dumped)
```
and `core dumped` doesn't look good.

seanshpark:
@mhs4670go , please show usage message if any arguments are missing or invalid

mhs4670go:
@seanshpark Okay. I'll update `arser` for error message in another PR.

seanshpark:
I need to check with this PR. I'll wait for new _arser_ and rebase to this PR.

mhs4670go:
@seanshpark PTAL:)

compiler/tflite2circle/driver/Driver.cpp
```diff
@@ -15,26 +15,45 @@
...
 
 int entry(int argc, char **argv)
 {
-  if (argc != 3)
+  arser::Arser arser;
+
+  arser.add_argument("tflite")
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .help("Give the path of tflite file you want to convert");
```
seanshpark, line 33: Can you revise this message to "Source tflite file path to convert"

compiler/tflite2circle/driver/Driver.cpp
```diff
@@ -15,26 +15,45 @@
...
+  arser.add_argument("circle")
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .help("Give the path of circle file after conversion");
```
seanshpark, line 37: Can you revise this message to "Target circle file path"

mhs4670go:
@seanshpark I added simple description. PTAL:)

compiler/tflite2circle/driver/Driver.cpp
```diff
@@ -25,7 +25,7 @@
 
 int entry(int argc, char **argv)
 {
-  arser::Arser arser;
+  arser::Arser arser{"tflite2circle is a Tensorflow Lite to Circle model converter"};
```
seanshpark, line 28: Let's use case like this: `Tensorflow lite to circle mode`

PR #3058, opened 2020-07-13
[infra] Update preset 20200630

seanshpark:
This will update preset 20200630
- add one-cmds for end user commands
- use tensorflow-cpu==2.3.0rc0: it's a bit smaller and our 2.x is 2.3.0rc0


seanshpark:
For #3018, from draft #3036

PR #3061, opened 2020-07-13
[circle-tensordump] Revise to use arser

mhs4670go:
This commit revises to use arser

Related : #3001


compiler/circle-tensordump/driver/Driver.cpp
```diff
@@ -25,66 +26,41 @@
...
-    std::cerr << std::endl;
-    print_help(argv[0]);
-    return EXIT_FAILURE;
-  }
+  arser::Arser arser{
+      "circle-tensordump allows users to retrieve tensor information from a Circle model file"};
 
-  // Simple argument parser (based on map)
-  std::map<std::string, OptionHook> argparse;
+  arser.add_argument("circle").nargs(1).type(arser::DataType::STR).help("Circle file path to dump");
```
llFreetimell, line 34: Just curious, not using `--circle` is intended?

mhs4670go: @llFreetimell Yes. Double dash is for optional argument. And without it is for positional argument.

PR #3065, opened 2020-07-13
[gen-core] temporary cmake change

mhs4670go:
This commit changes cmake find_package mode temporarily


compiler/gen-core/CMakeLists.txt
```diff
@@ -1,4 +1,9 @@
-find_package(HDF5 COMPONENTS CXX QUIET)
+# TODO This will be nnas_find_package
```
seanshpark, line 1: (optional) `TODO This will be nnas_find_package` seems strange expression

What I would use: `TODO Revise this to use nnas_find_package`

PR #2989, opened 2020-07-09
[DRAFT] hdf5 build in configuration

mhs4670go:
This draft builds hdf5 in cmake configuration.

Related : #2809 
ONE-DCO-1.0-Signed-off-by: seongwoo <mhs4670go@naver.com>

PR #3067, opened 2020-07-13
[nnc] temporary cmake change

mhs4670go:
This commit changes cmake find_package mode temporarily

### DIFF
--- a/compiler/nnc/cmake/config.cmake
+++ b/compiler/nnc/cmake/config.cmake
@@ -21,7 +21,13 @@ set(NNC_INSTALL_LIB_PATH ${NNC_INSTALL_PATH}/lib) # directory that contains othe
 #
 # find necessary packages
 #
-find_package(HDF5 COMPONENTS CXX QUIET)
+# TODO This will be nnas_find_package
+find_package(HDF5 COMPONENTS CXX QUIET CONFIG)
+
+if(NOT HDF5_FOUND)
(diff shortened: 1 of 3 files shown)


s-barannikov:
It looks strange. What is it for?

mhs4670go:
@s-barannikov As soon as `HDF5Config.cmake` is landed, it will change to just `nnas_find_package`.

For release, we have to static-link `hdf5` library to `record-minmax`. But, it is difficult to achieve this with `find_package` supported by cmake because dynamic and static link are mixed. So, I'll achieve this by building `hdf5` in cmake configure-step and then link it.

[This](#2989) is a full draft. But for merging this draft without violating our rule(one commit, one module), I have to use a little trick.

Here is the reason why I have to commit like this.

If I merge above draft in this order,
1. `HDF5Config.cmake` merge
2. change `find_package` to `nnas_find_package` module by module -> error

It would occur an error. Because `find_package` seek to `FindHDF5.cmake` and `nnas_find_package` seek to `HDF5Config.cmake`.

If I merge in this order,
1. change `find_package` to `nnas_find_package` module by module -> error
2. `HDF5Config.cmake` merge

It would occur an error because there is no HDF5-config.cmake

So, if `CONFIG` file was not found, I had to add logic to find `MODULE` file.
1. add above logic
2. `HDF5Config.cmake` merge
3. remove above logic and just add `nnas_find_package`

Issue #3053, opened 2020-07-13
[onert] Performance regression on quantized mobilnet inference


hseok-oh:
Example model: `mobilenet_v1_1.0_224_quant.tflite`

### release/1.6.0, cpu backend

```
$ BACKENDS=cpu USE_NNAPI=1 ./Product/out/bin/tflite_run -w 1 -r 5 mobilenet_v1_1.0_224_quant.tflite

===================================
MODEL_LOAD   takes 7.318 ms
PREPARE      takes 0.173 ms
EXECUTE      takes 57.994 ms
- MEAN     :  57.994 ms
- MAX      :  58.268 ms
- MIN      :  57.762 ms
- GEOMEAN  :  57.993 ms
===================================

```

### master (99b5d3055102d3aaacfeb2e3b014c2897a022507), cpu backend

```
$ BACKENDS=cpu USE_NNAPI=1 ./Product/out/bin/tflite_run -w 1 -r 5 mobilenet_v1_1.0_224_quant.tflite

===================================
MODEL_LOAD   takes 8.258 ms
PREPARE      takes 0.110 ms
EXECUTE      takes 662.754 ms
- MEAN     :  662.754 ms
- MAX      :  663.705 ms
- MIN      :  661.943 ms
- GEOMEAN  :  662.753 ms
===================================

```

hseok-oh:
Performance down after #2894

commit 510ff81d71abb49c856e397b90ad3ccfbc23107e:

```
===================================
MODEL_LOAD   takes 9.616 ms
PREPARE      takes 0.131 ms
EXECUTE      takes 661.396 ms
- MEAN     :  661.396 ms
- MAX      :  662.595 ms
- MIN      :  660.704 ms
- GEOMEAN  :  661.396 ms
===================================
```

commit 0f161d7929516e2bff839b2d6709db160ea0ff2b (510ff81d71abb49c856e397b90ad3ccfbc23107e's parent commit)

```
===================================
MODEL_LOAD   takes 9.138 ms
PREPARE      takes 0.189 ms
EXECUTE      takes 53.630 ms
- MEAN     :  53.630 ms
- MAX      :  53.864 ms
- MIN      :  53.221 ms
- GEOMEAN  :  53.629 ms
===================================
```

hseok-oh:
After #3072 

```
===================================
MODEL_LOAD   takes 7.928 ms
PREPARE      takes 0.110 ms
EXECUTE      takes 44.991 ms
[...]

PR #3072, opened 2020-07-13
[onert] Fix performance regression by Convolution operation

hseok-oh:
- Call prepare() on cpu backend run
- Always use optimized kernel for qasymm8

### DIFF
--- a/compute/cker/include/cker/operation/Conv.h
+++ b/compute/cker/include/cker/operation/Conv.h
@@ -119,19 +119,9 @@ public:
                        params.stride_height);
     }
 
-    if (_need_im2col)
-    {
-      uint8_t *im2col_raw_data = _im2col_data.data();
-      optimized::Conv(params, input_shape, input_data, filter_shape, filter_data, bias_shape,
-                      bias_data, output_shape, output_data, _im2col_shape, im2col_raw_data);
-    }
-    // TODO Apply optimized kernel if the optimized kernel is faster without im2col than the
-    // reference kernel
-    else
-    {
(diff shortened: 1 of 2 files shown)


hseok-oh:
Fix #3053

compute/cker/include/cker/operation/Conv.h
```diff
@@ -119,19 +119,9 @@ class Conv
...
-    // TODO Apply optimized kernel if the optimized kernel is faster without im2col than the
-    // reference kernel
-    else
-    {
-      reference::Conv(params, input_shape, input_data, filter_shape, filter_data, bias_shape,
-                      bias_data, output_shape, output_data);
-    }
+    uint8_t *im2col_raw_data = _im2col_data.data();
+    optimized::Conv(params, input_shape, input_data, filter_shape, filter_data, bias_shape,
+                    bias_data, output_shape, output_data, _im2col_shape, im2col_raw_data);
```
ragmani, line 124: `im2col_raw_data` can be `nullptr`, but `optimized::Conv()` does not supports it as `nullptr`.

hseok-oh: If `_need_im2col` is false, `im2col_raw_data` is not used in `optimized::Conv()`. Tensorflow lite also uses this feature for optimization.

https://github.com/Samsung/ONE/blob/42a0337242798aaf8c5d8f2644f3ce58b848d19f/compute/cker/include/cker/operation/optimized/Conv.h#L107-L136

ragmani: I see, thanks for your explanation.

PR #3080, opened 2020-07-13
[one-cmds] Fix typo

seanshpark:
This will fix typo of one-import-tf/tflite


seanshpark:
For #3018, from draft #3036

PR #3086, opened 2020-07-13
[onert] Update version eigen

hseok-oh:
- Update version eigen used for cker computation library (eigen for 2.3.0-rc0)
- TFLite 1.13.1 library uses eigen for 1.13.1
- Update copyright eigen_sptial_convolutions-inl.h


hseok-oh:
@chunseoklee @glistening Could you check performance regression on internal models?

chunseoklee:
> Could you check performance regression on internal models by this PR?

Not tested for all internal model, but IMHO no regression shown for several models.

glistening:
For one model, it slows about 1%. For another (smaller), it gets 0.55 ms → 0.41 ms. I measured both on xu4.

compute/cker/include/cker/eigen/eigen_spatial_convolutions-inl.h
```diff
@@ -1245,9 +1274,6 @@ struct gemm_pack_rhs<
     eigen_assert(stride == 0);
     eigen_assert(offset == 0);
 
-    (void)stride;
-    (void)offset;
-
```
glistening, line 1250: @hseok-oh Without these lines, android release build got errors due to unused variables.

hseok-oh:
I've updated. PTAL
- Resolve conflict
- Fix android build error
- Update copyright

glistening:
I tested on android with updated PR, which can build for android.

On s20, I have similar result on xu4.
big model gets 1 % slower, small model gets faster significantly.

PR #3090, opened 2020-07-13
[infra/cmake] Introduce TensorFlowProtoText 1.13.1

hseok-oh:
Introduce ProtoText for TensorFlow 1.13.1


parjong:
@seanshpark @hseok-oh You may remove [tfgraph-xform](https://github.com/Samsung/ONE/tree/master/compiler/tfgraph-xform) if this makes it difficult to maintain. As I remember, ``tfgraph-xform`` is the only module that needs this package.

Issue #677, opened 2020-05-11
[onert] Write a draft for removing input/output tensor copy


wateret:
Now each backend has its own memory manager and always need to copy for model inputs/outputs.

Let's support external tensor and avoid copying if possible.

- Let the backend have external tensor
- Sourcing/Sinking at the beginning of execution are removed
- Add Permute operation if necessary (between input tensor from user and runtime internal)

As-is:

![image](https://user-images.githubusercontent.com/10582184/81561279-ece17a00-93cd-11ea-96dc-11cb697e71cc.png)

To-be:

Two cases, permute needed, uneeded.

As Sourcing/Sinking in the execution is removed, everything is expressed explicitly in Graph IR.

![image](https://user-images.githubusercontent.com/10582184/81636787-ad5b7200-944f-11ea-87e7-efa376d72ab5.png)

wateret:
Done in #1325. Let's get them merged.

PR #3102, opened 2020-07-14
[onert] Manage UserTensor in controlflow backend

wateret:
Not only `cpu_common::UserTensor`, `UserTensor` is supported in
controlflow backend. As a workaround, two tensor registries are exist
for both types of tensors.

- Introduce `ITensorRegistry::getManagedITensor`
- Introduce `controlflow::DynamicTensorManager`
    - Bascially a copy of `cpu_common::DynamicTensorManager`
    - `UserTensor` handling is added

Draft PR : #2835

### DIFF
--- a/runtime/onert/core/include/backend/ITensorRegistry.h
+++ b/runtime/onert/core/include/backend/ITensorRegistry.h
@@ -17,6 +17,8 @@
 #ifndef __ONERT_BACKEND_ITENSOR_REGISTRY__
 #define __ONERT_BACKEND_ITENSOR_REGISTRY__
 
+#include <memory>
+
 #include "ir/Index.h"
 #include "backend/ITensor.h"
 
@@ -33,10 +35,17 @@ struct ITensorRegistry
   virtual ~ITensorRegistry() = default;
 
   /**
-   * @brief Returns pointer of ITensor
+   * @brief Returns pointer of ITensor among managed and external tensors
    * @note  Return tensor cannot be used longer than dynamic tensor manager
    */
   virtual std::shared_ptr<ITensor> getITensor(const ir::OperandIndex &) = 0;
+  /**
+   * @brief Returns pointer of ITensor among managed tensors
+   *
+   * Unlike @c getITensor , this function only searches from managed tensors
+   * @note  Return tensor cannot be used longer than dynamic tensor manager
+   */
+  virtual std::shared_ptr<ITensor> getManagedITensor(const ir::OperandIndex &) = 0;
 };
 
 } // namespace backend
@@ -67,6 +76,11 @@ public:
     auto external_tensor = _external.find(ind);
     if (external_tensor != _external.end())
       return external_tensor->second;
+    return getManagedTensor(ind);
+  }
+
+  std::shared_ptr<ITensor> getManagedITensor(const ir::OperandIndex &ind) override
+  {
     return getManagedTensor(ind);
   }
 
--- a/runtime/onert/core/src/backend/controlflow/Backend.h
+++ b/runtime/onert/core/src/backend/controlflow/Backend.h
@@ -66,7 +66,7 @@ public:
     auto tb = std::make_shared<TensorBuilder>();
     context->tensor_builder = tb;
     context->constant_initializer = std::make_shared<ConstantInitializer>(operands, tb);
-    context->kernel_gen = std::make_shared<KernelGenerator>(graph);
+    context->kernel_gen = std::make_shared<KernelGenerator>(graph, tb);
     context->tensor_register = nullptr;
     context->optimizer = nullptr;
     return context;
--- a/runtime/onert/core/src/backend/controlflow/DynamicTensorManager.cc
+++ b/runtime/onert/core/src/backend/controlflow/DynamicTensorManager.cc
@@ -1,0 +1,154 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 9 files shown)


wateret:
@ragmani @hyunsik-yoon PTAL :)

runtime/onert/core/include/backend/ITensorRegistry.h
```diff
@@ -33,10 +35,17 @@ struct ITensorRegistry
   virtual ~ITensorRegistry() = default;
 
   /**
-   * @brief Returns pointer of ITensor
+   * @brief Returns pointer of ITensor among managed and external tensors
```
ragmani, line 38: does this `external tensors` mean `External Tensor` in #3119?

wateret: No. More precisely, it is "portable tensors that are from other backends".

hyunsik-yoon:
1. I am not sure of background of `UserTensor`. `UserTensor` seems to be input tensors and output tensors of a graph. Could you explain the relation of input&output tensors of a graph and `UserTensor` of control flow backend?

2. Also I am not sure about the intention behind introducing dynamic tensor manger of control flow. Could you explain?

wateret:
> 1. I am not sure of background of `UserTensor`. `UserTensor` seems to be input tensors and output tensors of a graph. Could you explain the relation of input&output tensors of a graph and `UserTensor` of control flow backend?

1. What I'm doing is: Remove Source/Sink(copying from/to user buffers to/from backend's tensors when starting/ending execution) and make it present in the graph(a permute operation)
2. In the second figure from https://github.com/Samsung/ONE/issues/677#issue-615846767, USER OPERAND 0 and 1 is going to be part of the graph IR
3. So, in order to express users' buffer as a `ITensor`, `UserTensor` is introduced

> 2. Also I am not sure about the intention behind introducing dynamic tensor manger of control flow. Could you explain?

This is not necessarily relevant to my commit as it has been used before this commit(it uses `cpu_common::DynamicTensorManager` before this commit). Anyway I'm going to explain:

- Control flow backend also creates `cpu_common::Tensor`s to handle IF and WHILE ops. And I/O tensors for those operations could be dynamic
- `UserTensor`'s shape and buffer could change by `nnfw_set_input_tensorinfo()`

wateret:
Please understand this has some workarounds. Otherwise this work is hard to proceed. I rebased 2-3 times for this work already, so now I would like to keep some workarounds and merge this quickly then fix. 😃

hyunsik-yoon:
Thanks for the explanation.

In case of the following graph (https://github.com/Samsung/ONE/issues/677#issue-615846767) where there is no control flow ops,
```
#0 = placeholder
#1 = relu(#0)
```
is control flow backend involved in removing copy?

Well, basically my curiosity was... 'do we need control flow backend even though there is no control flow op?`. :-)

After writing the above question... I kinda remember some offline discussion where control flow backend can do more than processing "control flow op". Is this PR the case?

wateret:
> After writing the above question... I kinda remember some offline discussion where control flow backend can do more than processing "control flow op". Is this PR the case?

You're right. I needed a backend to express model in/outputs. Rather than introducing another one, I decided to use "controlflow" backend. If the name looks misleading, we may change the name later. (I'm thinking "builtin" backend)

wateret:
Please note that #2835 does not remove copies, it just changes the way of handling model's IO tensors to be present in Graph IR, just same as other tensors.

i.e. "Treat inputs/outputs specially, and copy inputs/outputs on start/end of the execution" → "Treat inputs/outputs as a tensor then insert Permute operation if needed, just like other tensors"

After doing this, model IO tensors are just same as other tensors, then we can have possibilities to optimize for some cases(do not insert Permute between backends that use portable tensors). And this applies to the same with non-IO tensors which means that there will be no Permute between those backends.

### Before #2835

There is no user-given buffers. copying inputs/outputs is done by `Executor::execute()`, before/after run.

Screen Shot 2020-07-15 at 9 18 44 PM

### Apply #2835

Those 3 controlflow tensors are the tensors with user-given buffers. Now we can see user's buffer. And now that they are in Graph, we can further optimize!

Screen Shot 2020-07-15 at 9 20 40 PM

hyunsik-yoon:
Thanks for the deep explanation! :+1: :heart: 

We should rename "control flow backend" someday... 
The name is really misleading readers. :-) 
(e.g., at first glance I thought, "is UserTensor for the input and output of while/if op?" :-P )

runtime/onert/core/include/backend/ITensorRegistry.h
hyunsik-yoon: When I first read this comment, I also thought that `ExternalTensor` in #3119. :-P 
Maybe we need different term since #3119 obtained copyright(?) of external tensor. :-)

ragmani: For different term, I have another question. What does `managed` mean?

wateret:
@hyunsik-yoon @ragmani PTAL :)

hyunsik-yoon:
There are some threads under https://github.com/Samsung/ONE/pull/3102#pullrequestreview-448749329 :-)

runtime/onert/core/include/backend/ITensorRegistry.h
wateret: - Managed : All tensors that the backend directly created
- External : Tensors(portable) that are created by other backends
- Managed ∪ External : All tensors that are accessible from the backend

Sometimes we need to check if a tensor is managed or external.

ragmani: I have no idea. "native" and "immigrant"?

PR #3103, opened 2020-07-14
[one-cmds] introduce one-prepare-venv

seanshpark:
This will introduce one-prepare-venv and its document to prepare python virtual environment with TensorFlow


seanshpark:
For #3018, from draft #3036

Issue #3100, opened 2020-07-14
Compiler FE: common venv


mhs4670go:
In our module test, we configure python virtual environment. But since each module makes own virtual env, many duplicate occur, which makes build and test spend more time than it should be.

To make this better, I'll configure it in `common-artifacts` once for each version.

Then `common-artifacts` layout will be like this.
```
common-artifacts
├── venv_1_13_1
├── venv_2_3_0_rc0
...
```

PR #3105, opened 2020-07-14
[common-artifacts] common venv

mhs4670go:
This commit creates common venv to common artifacts

Related : #3100 

### DIFF
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -1,5 +1,45 @@
-# TODO add pbtxt
+#[[ Generate common python virtual enviornment ]]
+find_package(PythonInterp 3 QUIET)
+find_package(PythonLibs 3 QUIET)
 
+if(NOT ${PYTHONINTERP_FOUND})
+  message(STATUS "Build common-artifacts: FALSE (Python3 is missing)")
+  return()
+endif()
+
+if(${PYTHON_VERSION_MINOR} LESS 3)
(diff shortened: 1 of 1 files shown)


compiler/common-artifacts/CMakeLists.txt
```diff
@@ -1,5 +1,45 @@
...
+  return()
+endif()
+
+if(${PYTHON_VERSION_MINOR} LESS 3)
+  message(STATUS "Build common-artifacts: FALSE (You need to install Python version higher than 3.3)")
+  return()
+endif()
+
+# Create python virtual environment with tensorflow 1.13.1
+set(VIRTUALENV "${CMAKE_CURRENT_BINARY_DIR}/venv_1_13_1")
```
seanshpark, line 16: Latest 1.13.x is 1.13.2. Can you change to use 1.13.2 ?

compiler/common-artifacts/CMakeLists.txt
```diff
@@ -1,5 +1,45 @@
...
 
+add_custom_command(
+  OUTPUT ${VIRTUALENV}
+  COMMAND ${PYTHON_EXECUTABLE} -m venv ${VIRTUALENV}
+)
+
+# Create requirements.txt and install required pip packages
+set(REQUIREMENTS_FILE "requirements.txt")
```
seanshpark, line 24: Is it ok with this PR without `requirements.txt` file?

mhs4670go: It's okay because I just create that file below by echo

Issue #3123, opened 2020-07-14
Compiler FE: daily compiler coverage failed


seanshpark:
http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/ fails

```
02:27:14  -- Configure TFLITE2CIRCLE-CONVERSION-TEST
02:27:14  CMake Error at /home/jenkins_slave_coverage/workspace/nnfw/master/daily-compiler-coverage/compiler/tflite2circle-conversion-test/CMakeLists.txt:41 (get_target_property):
02:27:14    get_target_property() called with non-existent target "testDataGenerator".
02:27:14  
02:27:14  
02:27:14  -- Configure TFLITE2CIRCLE-CONVERSION-TEST - Done
02:27:14  -- Configuring incomplete, errors occurred!
```

from recent no 31 : http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/31/console

seanshpark:
@mhs4670go , do you know what may be the problem?

seanshpark:
Seems `infra/scripts/configure_compiler_coverage.sh` needs update

mhs4670go:
`tflite2circle-conversion-test` should be built after `common-artfacts` is built. So I think we should add `common-artifacts` to script.

seanshpark:
need to add `arser` too

seanshpark:
done

PR #3125, opened 2020-07-14
[infra] Fix configure_compiler_coverage

seanshpark:
This will add recent common-artifacts to configure_compiler_coverage list


seanshpark:
For #3123

PR #3141, opened 2020-07-14
[onert] Fix DotDumper to show multiple edges

wateret:
- Show multiple edges if an operand is used multiple times
  by iterating over operations rather than operands
  (operand does not have order and duplications)
- Change creating a new `shared_ptr` object to a raw pointer of existing
  object
- Variable/method name changes : edge -> out_edge


wateret:
Screen Shot 2020-07-14 at 7 12 21 PM → Screen Shot 2020-07-14 at 7 12 05 PM

PR #3142, opened 2020-07-14
[arser] Support multiple string argument

mhs4670go:
This commit enables arser support string vector type

### DIFF
--- a/compiler/arser/tests/arser.test.cpp
+++ b/compiler/arser/tests/arser.test.cpp
@@ -206,3 +206,26 @@ TEST(BasicTest, MultipleFloatValue)
 
   EXPECT_THROW(arser.get<std::vector<int>>("--add_float"), std::runtime_error);
 }
+
+TEST(BasicTest, MultipleStringValue)
+{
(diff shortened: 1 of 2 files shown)


compiler/arser/tests/arser.test.cpp
```diff
@@ -206,3 +206,26 @@ TEST(BasicTest, MultipleFloatValue)
...
 }
+
+TEST(BasicTest, MultipleStringValue)
+{
+  /* arrange */
+  Arser arser;
+
+  arser.add_argument("--three_color")
+      .nargs(3)
+      .type(arser::DataType::STR_VEC)
```
wateret, line 217: (Just curiosity) What if I do something like this?

```cpp
      .nargs(3)
      .type(arser::DataType::STR)
```

mhs4670go: @wateret Then, it will take three inputs but user won't be able to "get" those values. `arser.get<std::string>` gives an error with type mismatch

Issue #3132, opened 2020-07-14
Compiler FE: Adding versions to commands


seanshpark:
Current command tools doesn't have version display options.
Let's provide `--version, -v` option for all commands.

To minimize changes when upgrade, let's make a single library that provides latest version number, string.

seanshpark:
done

PR #3161, opened 2020-07-15
DRAFT Compiler version library and command

seanshpark:
on-going draft to provide version string/number to compiler frontend commands


seanshpark:
For #3132

seanshpark:
all landed

PR #3162, opened 2020-07-15
[luci/interpreter]Add assert on maxPooling2D condition check.

struss:
This commit add input tensor condition check on luci/interpreter kernel.


struss:
Some operation with different quantization param on input tensor and output tensor make wrong result. so, should checking is needed if it is quantized operation.
Just referenced from V2.1 TFlite source.
https://github.com/tensorflow/tensorflow/blob/3ffdb91f122f556a74a6e1efd2469bfe1063cb5c/tensorflow/lite/kernels/pooling.cc#L92

PR #3165, opened 2020-07-15
[onert] Implement float/qasymm8 SpaceToBatchNDLayer in cpu backend

dayo09:
- float : padding with 0
- qasymm8 : padding with output's data offset


runtime/onert/core/src/compiler/OperationValidator.cc
```diff
@@ -1333,5 +1333,6 @@ void OperationValidator::visit(const ir::operation::MatrixBandPart &node)
   OP_REQUIRES(_ctx.at(num_upper_index).shape().rank() == 0); // num_lower must be scalar
   OP_REQUIRES(_ctx.at(num_lower_index).shape().rank() == 0); // num_upper must be scalar
 }
+
```
dr-venkman, line 1336: Perhaps revert this if not required?

dayo09: @dr-venkman I removed it, thanks.

runtime/onert/backend/cpu/ops/SpaceToBatchNDLayer.cc
```diff
@@ -0,0 +1,112 @@
...
+namespace ops
+{
+SpaceToBatchNDLayer::SpaceToBatchNDLayer()
+    : _input(nullptr), _block_shape(nullptr), _padding(nullptr), _output(nullptr)
+{
+  // DO NOTHING
+}
+
+template <typename T> void SpaceToBatchNDLayer::SpaceToBatchND(int32_t pad_value)
+{
```
dr-venkman, line 38: I understand the code below follows from Resizing the output tensor from TFLite. Except that in this case, it appears to be a validity check. Would it be possible to resize the output tensor here if the dimensions do not match?

dayo09: @dr-venkman No, resizing output tensor is not included here. This only checks dimension and throws error. 
We have dynamic shape inferer for dynamic tensor resizing. I am now working on implementation of shape inferer for this operator. After that, I will check if this dimension checking can be removed or can be changed into assertion.

runtime/onert/backend/cpu/ops/SpaceToBatchNDLayer.cc
```diff
@@ -0,0 +1,112 @@
...
+
+  nnfw::cker::SpaceToBatchND(
+      params, getTensorShape(_input), reinterpret_cast<const T *>(_input->buffer()),
+      getTensorShape(_block_shape), reinterpret_cast<const int32_t *>(_block_shape->buffer()),
+      getTensorShape(_padding), reinterpret_cast<const int32_t *>(_padding->buffer()),
+      getTensorShape(_output), reinterpret_cast<T *>(_output->buffer()));
+}
+
+void SpaceToBatchNDLayer::spaceToBatchNDFloat32() { SpaceToBatchND<float>(0); }
+
```
hasw7569, line 77: As I checked, `_output->data_offset()` is always 0 in this case.
so I think you can use data_offset() also here.

runtime/onert/backend/cpu/ops/SpaceToBatchNDLayer.cc
```diff
@@ -0,0 +1,112 @@
...
+  _padding = padding;
+  _output = output;
+}
+
+void SpaceToBatchNDLayer::run()
+{
+  if (_input->data_type() == OperandType::FLOAT32)
+  {
+    spaceToBatchNDFloat32();
+  }
```
hasw7569, line 98: It seems like unnecessary function call.
is there any reason to use this function ?

runtime/onert/backend/cpu/ops/SpaceToBatchNDLayer.cc
dayo09: @hasw7569 The value might be zero but I think it will be safer not to use data_offset for float values as it is not defined. The operator spec includes float's data offset as zero so I made the flow to fork with each designated value.

runtime/onert/backend/cpu/ops/SpaceToBatchNDLayer.cc
dayo09: The value might be zero but I think it will be safer not to use data_offset for float values as it is not defined. The operator spec includes float's data offset as zero so I made the flow to fork with each designated value.

runtime/onert/backend/cpu/ops/SpaceToBatchNDLayer.cc
hasw7569: Oh, the tflite kernel is also using `0`.

runtime/onert/backend/cpu/ops/SpaceToBatchNDLayer.cc
dayo09: I will explicit the meaning of pad_value in this case...

hasw7569: `And those other layers defines internal functions for each datatypes`
you are right, some layers defines internal function for each datatypes, but some layers are not.

I think the optimization has not been done because we have ported the kernel one by one for each data type.
So, if we don't have a layer policy, I think we should discuss it.

dayo09: @hasw7569 I'd be happy to participate in discussion. I changed the code for now to look more intuitive,  redundant bridge functions. Please check.

dayo09: (You may already have this in mind, but FYI, we creates `issue` for this kind of discussion and usually many people take parts in.)

hasw7569: ok : )
I will create issue about it.

dayo09:
@dr-venkman PTAL

runtime/onert/backend/cpu/ops/SpaceToBatchNDLayer.cc
```diff
@@ -0,0 +1,114 @@
...
+SpaceToBatchNDLayer::SpaceToBatchNDLayer()
+    : _input(nullptr), _block_shape(nullptr), _padding(nullptr), _output(nullptr)
+{
+  // DO NOTHING
+}
+
+// TO DO : move into shape inferer
+void SpaceToBatchNDLayer::checkDimension()
+{
+  const int kSpatialDimensionNum = 2;
```
dr-venkman, line 40: The value for `kSpatialDimensionNum` is set to 2, which seems to deviate from TFlite implementation where the corresponding value is `input_size->size -2`. Or is it that  both expressions are similar? Please ignore comment if trivial.

dayo09: This is a limitation on tflite kernel. You can find in tensorflow operation spec.  It only support like this :  
- input's rank must be 4 (ex, [1, 4, 4, 8])
- spatial dimension must be size of 2 (ex, [2,2])

dayo09:
@hasw7569 PTAL

PR #3167, opened 2020-07-15
[onert] Add a function to check if any graph input is dynamic

hyunsik-yoon:
For #2943
Related PR: #3044
draft: #2957

This adds a helper function to check if any graph input is dynamic.

FYI, To decide to run `DynamicShapeInferer`, the following is considered:

1. Are all tensors are static at compilation time?
2. Are all input tensors are static?

If both are true, `DynamicShapeInferer` will _not_ be run.

This function is to check the second case.

### DIFF
--- a/runtime/onert/core/src/exec/ExecutorBase.h
+++ b/runtime/onert/core/src/exec/ExecutorBase.h
@@ -143,6 +143,12 @@ private:
   }
 
 protected:
+  /**
+   * @brief Returns @c true if any input tensor is dynamic; @c false if all are static tensors
+   */
+  bool hasDynamicInput();
+
+protected:
   ExecutionObservee _subject;
   std::shared_ptr<ir::OperationIndexMap<int64_t>> _indexed_ranks;
   std::unique_ptr<ir::LoweredGraph> _lowered_graph;
@@ -157,11 +163,6 @@ protected:
 private:
   void handleDynamicInputTensor(ir::IOIndex input_index, const IODescription &desc);
 };
-
-/**
- * @brief Returns @c true if any input tensor is dynamic; @c false if all are static tensors
- */
-bool hasDynamicTensor(const std::vector<std::shared_ptr<onert::backend::ITensor>> &input_tensors);
 
 } // namespace exec
 } // namespace onert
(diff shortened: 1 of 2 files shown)


runtime/onert/core/src/exec/ExecutorBase.h
```diff
@@ -158,6 +158,11 @@ class ExecutorBase : public IExecutor
   void handleDynamicInputTensor(ir::IOIndex input_index, const IODescription &desc);
 };
 
+/**
+ * @brief Returns @c true if any input tensor is dynamic; @c false if all are static tensors
+ */
+bool has_dynamic_input(const std::vector<std::shared_ptr<onert::backend::ITensor>> &input_tensors);
```
wateret, line 164: All C++ function names must be in `camelCase` 😄

hyunsik-yoon: Ah. still confused with coding convention in ONE compiler source code. :-) will be fixed.

runtime/onert/core/src/exec/ExecutorBase.h
```diff
@@ -158,6 +158,11 @@ class ExecutorBase : public IExecutor
   void handleDynamicInputTensor(ir::IOIndex input_index, const IODescription &desc);
 };
 
+/**
+ * @brief Returns @c true if any input tensor is dynamic; @c false if all are static tensors
+ */
+bool hasDynamicTensor(const std::vector<std::shared_ptr<onert::backend::ITensor>> &input_tensors);
```
ragmani, line 164: Where will you use this function?

hyunsik-yoon: Could you find the code in https://github.com/Samsung/ONE/pull/2957/files#diff-d508434b4530bfdc88a7965944a65966R55 ?

ragmani: If so, how about moving this function in `ExecutorBase` as protected method?

hyunsik-yoon: That would be better. Thanks. Fixed and committed.

Issue #3144, opened 2020-07-14
Write "How to Use NNAPI Binding"


wateret:
Parent : #2472

https://github.com/Samsung/ONE/blob/master/docs/howto/how-to-use-nnapi-binding.md

PR #3177, opened 2020-07-15
[docs] Write "How to Use NNAPI Binding"

wateret:
Write "How to Use NNAPI Binding" briefly as NN API has its own
documents.


wateret:
For #3144

PR #3188, opened 2020-07-15
[onert] Deallocate memory used by CustomKernel

hyunsik-yoon:
This deallocates memory allocated for `_param.inputs` and `_param.outputs`.

### DIFF
--- a/runtime/onert/api/src/CustomKernel.cc
+++ b/runtime/onert/api/src/CustomKernel.cc
@@ -67,6 +67,15 @@ public:
 Kernel::Kernel(const nnfw_custom_eval evalFunction)
     : _params(), _userdata(nullptr), _userdata_size(0), _evalFunction(evalFunction)
 {
+  _params.inputs = _params.outputs = nullptr;
+}
+
+Kernel::~Kernel()
+{
+  if (_params.inputs)
+    delete[] _params.inputs;
+  if (_params.outputs)
+    delete[] _params.outputs;
 }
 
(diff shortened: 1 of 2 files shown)


runtime/onert/api/src/CustomKernel.cc
```diff
@@ -67,6 +67,15 @@ class APIConverter
...
 {
+  _params.inputs = _params.outputs = nullptr;
+}
+
+Kernel::~Kernel()
+{
+  if (_params.inputs)
+    delete[] _params.inputs;
+  if (_params.outputs)
+    delete[] _params.outputs;
```
glistening, line 78: @hyunsik-yoon I am not sure it is good to assume `_params.inputs` and `_params.outputs` are allocated with `new []`.

i) We need to specify `_params.inputs` and `_params.outputs` are allocated with `new [ ]`, not with `new`.
ii) In addition, `nnfw_custom_kernel_params` is part of NNAPI. It should be able to support `C`.

There are some custom kernels (mostly internally builtin custom kernels). After investigating how they allocate them, we need to modify them also.

glistening: @hyunsik-yoon Thank you for informing me that  `_params.*puts` are allocated in:
```c++
void Kernel::configure(CustomKernelConfigParams &&inParams)
{
 ...
  _params.inputs = new nnfw_operand[_params.ninputs];
}
```

hyunsik-yoon: :-)

PR #3189, opened 2020-07-15
[onert] Write "Heterogeneous Execution"

wateret:
Write "Heterogeneous Execution" which is a big feature of ONE Runtime.


docs/runtime/heterogeneous-execution.md
```diff
@@ -0,0 +1,35 @@
...
+
+This means that choosing a backend for an operation should be careful. We must take Permute(copy) overhead into account. That cost could dominate actual computation's so choosing the fastest one is not always the best choice.
+
+Sometime later, we are going to introduce a concept of compatible backends and they can share their tensors freely. So for those cases, Permute operations are not inserted.
+
+## Automatic Scheduling
+
+Now we know that it could be useful to have a lot of backends, but one might wonder how we can schedule backends effectively. When a model is large, it is not trivial to do it manually by a user.
+
+There is a profiling-based automatic scheduler(experimental). It first runs every operation in sequence with every backend it has. And it records the duration of the execution for each. Then that data can be used to schedule. The scheduler assigns backends according to a certain algorithm with the given data. And the user can execute it with Parallel Scheduler.
```
hyunsik-yoon, line 33: Can we also have link to `profiling-based automatic scheduler(experimental)`? Some file link would be great.

wateret:
```suggestion
There is a [profiling-based automatic scheduler](/runtime/onert/core/src/compiler/HEScheduler.h)(experimental). It first runs every operation in sequence with every backend it has. And it records the duration of the execution for each. Then that data can be used to schedule. The scheduler assigns backends according to a certain algorithm with the given data. And the user can execute it with Parallel Scheduler.
```

wateret:
One more review from anyone please?

PR #3191, opened 2020-07-15
[foder] revise error message

seanshpark:
This will revise error message to make consistant with other drivers


seanshpark:
From draft #3161

PR #3194, opened 2020-07-16
[arser] fix type mismatch

seanshpark:
This will fix type mismatch with argc


seanshpark:
CC @mhs4670go

PR #3204, opened 2020-07-16
[tfldump] Use standard library

mhs4670go:
This commit gets tfldump to use standard library


mhs4670go:
@nnfw-bot test nncc-debug

PR #3212, opened 2020-07-16
[luci-interpreter]Add Squeeze kernels on luci-interpreter

struss:
This commit add `Squeeze` operation kernels on luci-interpreter.

### DIFF
--- a/compiler/luci-interpreter/src/core/KernelParams.h
+++ b/compiler/luci-interpreter/src/core/KernelParams.h
@@ -123,6 +123,12 @@ struct StridedSliceParams
   int32_t shrink_axis_mask;
 };
 
+struct SqueezeParams
+{
+  int8_t squeeze_dims_count;
+  int32_t squeeze_dims[4];
+};
+
 struct TransposeConvParams
 {
   Padding padding;
--- a/compiler/luci-interpreter/src/core/KernelParams.h
+++ b/compiler/luci-interpreter/src/core/KernelParams.h
@@ -22,6 +22,7 @@
 #include <luci_interpreter/core/DataType.h>
 
 #include <cstdint>
+#include <vector>
 
 namespace luci_interpreter
 {
@@ -125,8 +126,7 @@ struct StridedSliceParams
 
 struct SqueezeParams
 {
-  int8_t squeeze_dims_count;
-  int32_t squeeze_dims[4];
+  std::vector<int32_t> squeeze_dims;
 };
 
 struct TransposeConvParams
--- a/compiler/luci-interpreter/src/kernels/Squeeze.cpp
+++ b/compiler/luci-interpreter/src/kernels/Squeeze.cpp
@@ -75,23 +75,12 @@ void Squeeze::configure()
 
 void Squeeze::execute() const
 {
-  size_t num_input = 1;
-  size_t num_output = 1;
-  for (int i = 0; i < _input->shape().num_dims(); i++)
-  {
-    num_input = num_input * _input->shape().dim(i);
-  }
-  for (int i = 0; i < _output->shape().num_dims(); i++)
-  {
-    num_output = num_output * _output->shape().dim(i);
-  }
-
-  assert(num_input * getDataTypeSize(_input->element_type()) ==
-         num_output * getDataTypeSize(_output->element_type()));
+  assert(_input->shape().num_elements() == _output->shape().num_elements());
 
   const auto *input_data = _input->data<void>();
   auto *output_data = _output->data<void>();
-  std::memcpy(output_data, input_data, getDataTypeSize(_input->element_type()) * num_input);
+  std::memcpy(output_data, input_data,
+              getDataTypeSize(_input->element_type()) * _input->shape().num_elements());
 }
 
 } // namespace kernels
--- a/compiler/luci-interpreter/src/kernels/Squeeze.cpp
+++ b/compiler/luci-interpreter/src/kernels/Squeeze.cpp
@@ -86,20 +86,12 @@ void Squeeze::execute() const
     num_output = num_output * _output->shape().dim(i);
   }
 
-  assert(num_input * sizeof(getDataTypeSize(_input->element_type())) ==
-         num_output * sizeof(getDataTypeSize(_output->element_type())));
+  assert(num_input * getDataTypeSize(_input->element_type()) ==
+         num_output * getDataTypeSize(_output->element_type()));
 
-  switch (_input->element_type())
-  {
-    case DataType::FLOAT32:
-      memcpy(_output->data<float>(), _input->data<float>(), sizeof(float) * num_input);
-      break;
-    case DataType::U8:
-      memcpy(_output->data<uint8_t>(), _input->data<uint8_t>(), sizeof(uint8_t) * num_input);
-      break;
-    default:
-      throw std::runtime_error("Unsupported type.");
-  }
+  const auto *input_data = _input->data<void>();
+  auto *output_data = _output->data<void>();
+  std::memcpy(output_data, input_data, getDataTypeSize(_input->element_type()) * num_input);
 }
 
 } // namespace kernels
--- a/compiler/luci-interpreter/src/kernels/Squeeze.cpp
+++ b/compiler/luci-interpreter/src/kernels/Squeeze.cpp
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "kernels/Squeeze.h"
+
+#include "kernels/Utils.h"
+
+#include <stdexcept>
+
+namespace luci_interpreter
+{
+
+namespace kernels
+{
+
+Squeeze::Squeeze(const Tensor *input, Tensor *output, const SqueezeParams &params)
+    : KernelWithParams<SqueezeParams>(params), _input(input), _output(output)
+{
+}
+
+void Squeeze::configure()
+{
+  int input_num_dims = _input->shape().num_dims();
+  int num_squeeze_dims = params().squeeze_dims_count;
+  const int *squeeze_dims = params().squeeze_dims;
+  assert(input_num_dims <= 8);
+  bool should_squeeze[8] = {false};
+  int num_squeezed_dims = 0;
+  if (num_squeeze_dims == 0)
+  {
+    for (int idx = 0; idx < input_num_dims; ++idx)
+    {
+      if (_input->shape().dim(idx) == 1)
+      {
+        should_squeeze[idx] = true;
+        ++num_squeezed_dims;
+      }
+    }
+  }
+  else
+  {
+    for (int idx = 0; idx < num_squeeze_dims; ++idx)
+    {
+      int current = squeeze_dims[idx] < 0 ? squeeze_dims[idx] + input_num_dims : squeeze_dims[idx];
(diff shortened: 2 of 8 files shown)


compiler/luci-interpreter/src/kernels/Squeeze.cpp
```diff
@@ -0,0 +1,106 @@
...
+  {
+    case DataType::FLOAT32:
+      memcpy(_output->data<float>(), _input->data<float>(), sizeof(float) * num_input);
+      break;
+    case DataType::U8:
+      memcpy(_output->data<uint8_t>(), _input->data<uint8_t>(), sizeof(uint8_t) * num_input);
+      break;
+    default:
+      throw std::runtime_error("Unsupported type.");
+  }
```
s-barannikov, line 102: It can be easily done without switch and support all other types, see the implementation of `Reshape` op.
Also, prefer `std::memcpy` over `memcpy`.

compiler/luci-interpreter/src/loader/KernelBuilder.cpp
```diff
@@ -455,6 +456,22 @@ std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleStridedSlice *nod
...
 
+std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleSqueeze *node)
+{
+  assert(node->arity() == 1);
+
+  const Tensor *input = getInputTensor(node->input());
+  Tensor *output = getOutputTensor(node);
+
+  SqueezeParams params{};
+  params.squeeze_dims_count = node->squeeze_dims().size();
```
s-barannikov, line 467: It should be checked that the number of dimensions does not exceed the size of `SqueezeParams::squeeze_dims`.

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
```diff
@@ -0,0 +1,85 @@
...
+  EXPECT_THAT(extractTensorData<float>(output_tensor),
+              ElementsAreArray(ArrayFloatNear({1.0,  2.0,  3.0,  4.0,  5.0,  6.0,  7.0,  8.0,
+                                               9.0,  10.0, 11.0, 12.0, 13.0, 14.0, 15.0, 16.0,
+                                               17.0, 18.0, 19.0, 20.0, 21.0, 22.0, 23.0, 24.0})));
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray({24}));
+}
+
+TEST(SqueezeTest, Uint8)
+{
+  // There is No Uint8 case on TFlite Test, So I just add Quantization Parameters.
```
s-barannikov, line 56: Why don't just do a TYPED_TEST (see `Transpose`, `Unpack`, `Split` for examples)?
Although we use TFLite sources, we do not have to copy them.

compiler/luci-interpreter/src/kernels/Squeeze.cpp
struss: Okay. I will check `Reshape` kernel and update this kernel.

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
struss: Okay, I will fix it!

compiler/luci-interpreter/src/loader/KernelBuilder.cpp
struss: okay, I will add a assert before it that checks the size of node->squeeze_dims() exceed the size of `SqueezeParam::squeeze_dims`(4)

compiler/luci-interpreter/src/kernels/TestUtils.h
```diff
@@ -71,7 +71,6 @@ std::vector<::testing::Matcher<float>> ArrayFloatNear(const std::vector<float> &
 template <typename T>
 inline std::vector<T> quantize(const std::vector<float> &data, float scale, int32_t zero_point)
 {
-  assert(!std::is_floating_point<T>::value);
```
jinevening, line 74: I think we had better keep this assert. 

If you really want to remove this, you should change the below code `numeric_limits<T>::min ()` to `numeric_limits<T>::lowest()`.

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
```diff
@@ -26,58 +26,51 @@ namespace
...
+           std::initializer_list<float> input_data, std::initializer_list<float> output_data,
+           DataType element_type, int8_t squeeze_dims_count, std::vector<int32_t> squeeze_dims)
 {
-  std::vector<float> input_data{1.0,  2.0,  3.0,  4.0,  5.0,  6.0,  7.0,  8.0,
-                                9.0,  10.0, 11.0, 12.0, 13.0, 14.0, 15.0, 16.0,
-                                17.0, 18.0, 19.0, 20.0, 21.0, 22.0, 23.0, 24.0};
-  Shape input_shape = {1, 24, 1};
-  Tensor input_tensor = makeInputTensor<DataType::FLOAT32>(input_shape, input_data);
-  Tensor output_tensor = makeOutputTensor(DataType::FLOAT32);
+  std::pair<float, int32_t> quant_param = quantizationParams<T>(-127.f, 128.f);
```
jinevening, line 34: Do we have to test with quantizationParams? Squeeze does not change quantizationParams.

A similar case (`Transpose.test.cpp`), which does not change data, does not use quantizationParams for testing uint8 case.

compiler/luci-interpreter/src/kernels/TestUtils.h
struss: Okay, I got it! i will change `numeric_limits<T>::min()` to `numeric_limits<T>::lowest()`.

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
struss: okay. I will remove this and update the test cases.

compiler/luci-interpreter/src/kernels/TestUtils.h
jinevening: If you don't use `quantizationParams` for testing with float values, this assert does not have to be removed.

struss: Okay, I will update it!

compiler/luci-interpreter/src/kernels/Squeeze.cpp
```diff
@@ -0,0 +1,98 @@
...
+    {
+      output_shape.dim(out_idx++) = _input->shape().dim(in_idx);
+    }
+  }
+  _output->resize(output_shape);
+}
+
+void Squeeze::execute() const
+{
+  size_t num_input = 1;
```
s-barannikov, line 78: Why don't just use `shape()->num_elements()`?

compiler/luci-interpreter/src/kernels/Squeeze.cpp
```diff
@@ -0,0 +1,98 @@
...
+  for (int i = 0; i < _input->shape().num_dims(); i++)
+  {
+    num_input = num_input * _input->shape().dim(i);
+  }
+  for (int i = 0; i < _output->shape().num_dims(); i++)
+  {
+    num_output = num_output * _output->shape().dim(i);
+  }
+
+  assert(num_input * getDataTypeSize(_input->element_type()) ==
```
s-barannikov, line 89: Can the input and output data types differ?

compiler/luci-interpreter/src/kernels/Squeeze.cpp
struss: Oh, Thanks. I will use that function!

compiler/luci-interpreter/src/kernels/Squeeze.cpp
struss: on configure function, it will be checked. so it won't be differ. so I will just check the element number.

compiler/luci-interpreter/src/core/KernelParams.h
```diff
@@ -123,6 +123,12 @@ struct StridedSliceParams
   int32_t shrink_axis_mask;
 };
 
+struct SqueezeParams
+{
+  int8_t squeeze_dims_count;
+  int32_t squeeze_dims[4];
```
binarman, line 129: Why not use vector?

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
```diff
@@ -0,0 +1,78 @@
...
+namespace kernels
+{
+namespace
+{
+
+using namespace testing;
+
+template <typename T>
+void Check(std::initializer_list<int32_t> input_shape, std::initializer_list<int32_t> output_shape,
+           std::initializer_list<float> input_data, std::initializer_list<float> output_data,
```
s-barannikov, line 31:
```suggestion
           std::initializer_list<T> input_data, std::initializer_list<T> output_data,
```

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
```diff
@@ -0,0 +1,78 @@
...
+           DataType element_type, int8_t squeeze_dims_count, std::vector<int32_t> squeeze_dims)
+{
+  Tensor input_tensor{element_type, input_shape, {}, ""};
+  std::vector<T> update_input_data;
```
s-barannikov, line 35: With the suggestion above, this is redundant.

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
```diff
@@ -0,0 +1,78 @@
...
+};
+
+using DataTypes = ::testing::Types<float, uint8_t>;
+TYPED_TEST_CASE(SqueezeTest, DataTypes);
+
+TYPED_TEST(SqueezeTest, TotalTest)
+{
+  Check<TypeParam>(
+      /*input_shape=*/{1, 24, 1}, /*output_shape=*/{24},
+      /*input_data=*/{1.0,  2.0,  3.0,  4.0,  5.0,  6.0,  7.0,  8.0,  9.0,  10.0, 11.0, 12.0,
```
s-barannikov, line 69: The numbers should be integers. Integers will be implicitly converted to float or int8_t as needed.

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
s-barannikov: See `Transpose` op test for an example.

compiler/luci-interpreter/src/core/KernelParams.h
struss: Oh, sorry, I refer some wrong referred source. I will change into a int vector.

s-barannikov: It is indeed a simple array in TFLite, but we don't need to copy them.

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
struss: Okay I will fix it.

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
struss: Okay I will fix it.

compiler/luci-interpreter/src/kernels/Squeeze.test.cpp
struss: Okay I will fix it.

compiler/luci-interpreter/src/core/KernelParams.h
struss: yes. I saw wrong reference. sorry for confused.

compiler/luci-interpreter/src/loader/KernelBuilder.cpp
```diff
@@ -455,6 +456,23 @@ std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleStridedSlice *nod
...
 
+std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleSqueeze *node)
+{
+  assert(node->arity() == 1);
+
+  const Tensor *input = getInputTensor(node->input());
+  Tensor *output = getOutputTensor(node);
+
+  SqueezeParams params{};
+  assert(node->squeeze_dims().size() <= 4);
```
binarman, line 467: optional:
I think this assert is not necessary anymore, since `SqueezeParams::squeeze_dims` accept any number of dimension now

s-barannikov: AFAIK TFLite implementation only supports up to 4 dimensions.

PR #3215, opened 2020-07-16
[circledump] Use arser

mhs4670go:
This commit makes circledump use arser

Related : #3001 


mhs4670go:
@seanshpark @llFreetimell PTAL:)

Issue #3174, opened 2020-07-15
[Compiler] Handling circle operation version in luci


llFreetimell:
### Background

For now, `op_version` in `circle` was only used in `circledump`.
However, `op_version` can be used when circle operation is imported or exported.
Let's apply circle operation version to `luci` !

### Todo

- [x] Introduce `op_version` in `luci/lang` #3264
- [x] Apply `op_version` to `luci`
    - [x] `luci/export` #3301
    - [x] `luci/import` #3280
- [x] Update `tflchef`
    - [x] `proto` #3266
    - [x] chef #3302
- [x] Update `circlechef`
    - [x] `proto` #3265
    - [x] chef #3303
- [x] Introduce new rule for checking op version in `tfl-inspect` #3281 #3321
- [x] Introduce new rule for checking op version in `circle-inspect` #3288 #3323
- [x] Introduce new rule for checking op version in `dredd` #3317
- [x] Add new testcase to exclude list in `common-artifact` #3334
- [x] Add testcases with rule file in `res` #3332
- [x] Enable new testcase in `dredd` test #3345

### Draft
 
- #3216

### Reference
- Version control code in TFLite
    - https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/toco/tflite/op_version.cc
    - https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/register.cc
- Version administration
    - https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/g3doc/guide/ops_version.md

llFreetimell:
### Tensorflow Lite operator versioning

```python
import tensorflow as tf # Tested on TF 2.3rc0

if __name__ == "__main__":
  
  sess = tf.compat.v1.Session()
  tf.compat.v1.disable_eager_execution()

  c0 = tf.compat.v1.constant(1,dtype=tf.float32,shape=[1,2,5,5],name="c0")
  c1 = tf.compat.v1.constant(1,dtype=tf.float32,shape=[1,2,5,5],name="c1")
  r1 = tf.compat.v1.nn.depthwise_conv2d_native(c0,c1,[1,2,2,1],'VALID',dilations=[1,1,1,1]) # Version 1
  bias1 = tf.compat.v1.constant(2.2,dtype=tf.float32,shape=[25])
  ba1 = tf.compat.v1.nn.bias_add(r1,bias1)
  p1 = tf.reshape(ba1,[-1])

[...]

PR #3216, opened 2020-07-16
[Draft] [luci] Applying circle operation version

llFreetimell:
Parent issue : #3174

This draft is for applying circle operation version in `luci`.

### DIFF
--- a/compiler/circle-inspect/src/Dump.cpp
+++ b/compiler/circle-inspect/src/Dump.cpp
@@ -133,3 +133,38 @@ void DumpConv2DWeight::run(std::ostream &os, const circle::Model *model)
 }
 
 } // namespace circleinspect
+
+namespace circleinspect
+{
+
+void DumpOperatorVersion::run(std::ostream &os, const circle::Model *model)
+{
+  std::map<std::string, int32_t> op_version_map;
+
+  circleinspect::Reader reader(model);
+
+  assert(reader.num_subgraph() == 1);
+  reader.select_subgraph(0);
+
+  auto ops = reader.operators();
+
+  // Dump operators' version
+  for (uint32_t i = 0; i < ops->Length(); ++i)
+  {
+    const auto op = ops->Get(i);
+
+    auto op_name = reader.opcode_name(op);
+    auto op_version = reader.opcodes().at(op->opcode_index())->version();
+
+    if (op_version_map.find(op_name) == op_version_map.end() ||
+        op_version_map[op_name] < op_version)
+      op_version_map[op_name] = op_version;
+  }
+
+  for (auto op : op_version_map)
+  {
+    os << op.first << "," << op.second << std::endl;
+  }
+}
+
+} // namespace circleinspect
--- a/compiler/circlechef/core/src/ModelChef.cpp
+++ b/compiler/circlechef/core/src/ModelChef.cpp
@@ -164,16 +164,19 @@ OpChefRegistry &op_chef_registry(void)
 }
 
 /// @brief This will prepare a set of unique builtin codes in the mode recipe
-std::set<circle::BuiltinOperator>
-gather_builtincode_set(const ::circlechef::ModelRecipe &model_recipe)
+std::map<circle::BuiltinOperator, int32_t>
+gather_builtincode_map(const ::circlechef::ModelRecipe &model_recipe)
 {
-  std::set<circle::BuiltinOperator> builtin_set;
+  std::map<circle::BuiltinOperator, int32_t> builtin_map;
   for (const auto &operation : model_recipe.operation())
   {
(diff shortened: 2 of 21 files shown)


seanshpark:
have you checked the version numbers in all operators in the tflite/circle schema? and how the version is managed in generated code?

compiler/circle-inspect/src/Dump.cpp
```diff
@@ -133,3 +133,38 @@ void DumpConv2DWeight::run(std::ostream &os, const circle::Model *model)
...
+{
+
+void DumpOperatorVersion::run(std::ostream &os, const circle::Model *model)
+{
+  std::map<std::string, int32_t> op_version_map;
+
+  circleinspect::Reader reader(model);
+
+  assert(reader.num_subgraph() == 1);
+  reader.select_subgraph(0);
```
seanshpark, line 147: What about multiple graph models?

compiler/circle-inspect/src/Dump.cpp
```diff
@@ -133,3 +133,38 @@ void DumpConv2DWeight::run(std::ostream &os, const circle::Model *model)
...
+  for (uint32_t i = 0; i < ops->Length(); ++i)
+  {
+    const auto op = ops->Get(i);
+
+    auto op_name = reader.opcode_name(op);
+    auto op_version = reader.opcodes().at(op->opcode_index())->version();
+
+    if (op_version_map.find(op_name) == op_version_map.end() ||
+        op_version_map[op_name] < op_version)
+      op_version_map[op_name] = op_version;
```
seanshpark, line 161: This will permit only max version per OpCode as I see. What is the use case for this in inspection?

compiler/circlechef/core/src/ModelChef.cpp
```diff
@@ -185,11 +188,14 @@ gather_builtincode_set(const ::circlechef::ModelRecipe &model_recipe)
       auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
       if (op_chef->code() == circle::BuiltinOperator_CUSTOM)
         continue;
-      builtin_set.insert(op_chef->code());
+
+      if (builtin_map.find(op_chef->code()) == builtin_map.end() ||
+          builtin_map[op_chef->code()] < operation.version())
+        builtin_map[op_chef->code()] = operation.version();
```
seanshpark, line 194: This will permit only maximum number of version per OpCode if  I read your code correctly.
Is there no model that has same OpCode with different version?

compiler/circlechef/core/src/ModelChef.cpp
```diff
@@ -164,16 +164,19 @@ OpChefRegistry &op_chef_registry(void)
 }
 
 /// @brief This will prepare a set of unique builtin codes in the mode recipe
-std::set<circle::BuiltinOperator>
-gather_builtincode_set(const ::circlechef::ModelRecipe &model_recipe)
+std::map<circle::BuiltinOperator, int32_t>
```
seanshpark, line 167: I'm not sure about sorting order of `set` vs `map`. Is this really OK with the file format?

seanshpark: CC @parjong

compiler/luci/export/src/CircleExporterImpl.cpp
```diff
@@ -109,7 +109,7 @@ encodeOperatorCodes(FlatBufferBuilder &builder, std::unordered_map<luci::OpCode,
     uint32_t idx = it.second;
     if (it.first.opcode != BuiltinOperator_CUSTOM)
     {
-      operator_codes_vec[idx] = CreateOperatorCode(builder, it.first.opcode);
+      operator_codes_vec[idx] = CreateOperatorCode(builder, it.first.opcode, 0, it.first.version);
```
seanshpark, line 112: I can't find modified `CreateOperatorCode` of yours int this PR.

compiler/circlechef/core/src/ModelChef.cpp
llFreetimell: As I said at offline, it seems to be.
I wrote some memo related with it in https://github.com/Samsung/ONE/issues/3174#issuecomment-660802695 :)

compiler/circle-inspect/src/Dump.cpp
llFreetimell: I will handle it later because `DumpConv2DWeight` like functions also have same asserts :)

seanshpark: OK, I think circle-inspect and tfl-inspect needs revise :)

seanshpark: I've posted #3278

llFreetimell:

Issue #2359, opened 2020-06-17
Compiler FE: use overlay to reduce pip error


mhs4670go:
There are some tests having dependency to python virtual environment, which is kind of a burden of our CI. That is why pip network error occurs very often recently.

To make our CI happy, we can create the virtual environment to overlay.

Let's do this:)

seanshpark:
```diff
@@ -38,8 +38,8 @@ bool CircleCeilGraphBuilder::validate(const ValidateArgs &args) const
 }
 
 CircleNode *CircleCeilGraphBuilder::build_node(const circle::OperatorT &,
-                                                const std::vector<CircleNode *> &inputs,
-                                                loco::Graph *graph) const
+                                               const std::vector<CircleNode *> &inputs,
+                                               loco::Graph *graph) const
 {
   auto *node = graph->nodes()->create<CircleCeil>();
   node->x(inputs[0]);
```

PR #3223, opened 2020-07-16
[common-artifacts] Create python venv in overlay

mhs4670go:
This commit creates python venv in overlay

Related: #2359 


seanshpark:
@nnfw-bot test nncc-release

PR #3237, opened 2020-07-16
[record-minmax] Use batch to compute moving avg

jinevening:
This uses batch to compute moving avg

### DIFF
--- a/compiler/record-minmax/src/RecordMinMax.cpp
+++ b/compiler/record-minmax/src/RecordMinMax.cpp
@@ -153,8 +153,8 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
     }
     else if (mode == "moving_average")
     {
-      min = getMovingAverage(minmax.min_vector, 0.9);
-      max = getMovingAverage(minmax.max_vector, 0.9);
+      min = getMovingAverage(minmax.min_vector, 0.9, 16, true);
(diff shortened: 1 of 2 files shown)


compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -153,8 +153,8 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
     }
     else if (mode == "moving_average")
     {
-      min = getMovingAverage(minmax.min_vector, 0.9);
-      max = getMovingAverage(minmax.max_vector, 0.9);
+      min = getMovingAverage(minmax.min_vector, 0.9, 16, true);
```
seanshpark, line 156: value here is 0.9 for `alpha`.Is it like 90% of min_vector?

jinevening: > Is it like 90% of min_vector?

100% of data are used to compute the moving average.

`alpha` is the parameter of the equation of exponentially weighted moving avg ([EWMA](https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average)). EWMA is updated as below.

`EWMA = alpha * EWMA + (1 - alpha) * new_data`

So, `alpha` is the weight where the previous data contributes to the next computed value.

seanshpark: Thanks for the explanation

PR #3242, opened 2020-07-16
[circle-quantizer] Use arser

seanshpark:
This will revise to use arser for arguments parsing


compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -35,104 +36,81 @@ using OptionHook = std::function<int(const char **)>;
...
+            "output_dtype(uint8) granularity(layer)");
 
-  // TODO use better parsing library (ex: boost.program_options)
-  argparse["--quantize_with_minmax"] = [&options](const char **argv) {
-    options->enable(Algorithms::QuantizeWithMinMax);
-
-    if (argv[0] == nullptr || argv[1] == nullptr || argv[2] == nullptr)
-      throw std::runtime_error("--quantize_with_minmax must have three following parameters.");
+  arser.add_argument("input").nargs(1).type(arser::DataType::STR).help("Input circle model");
+  arser.add_argument("output").nargs(1).type(arser::DataType::STR).help("Output circle model");
```
jinevening, line 69: Maybe `--input` and `--output`?

seanshpark: it's positional so add without `--`

PR #3232, opened 2020-07-16
DRAFT: imgdata2hdf5, tool to convert raw img data to hdf5

jinevening:
On going draft to land imgdata2hdf5

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to #3213

Issue #3213, opened 2020-07-16
[imgdata2hdf5] Python script to convert raw image data to hdf5


jinevening:
The input dataset for the argument of `record-minmax` has the form of hdf5. Users may have difficulty in making hdf5 file, so we will provide `imgdata2hdf5.py`, a Python script that converts raw data to hdf5 (only for image data for now)

![image](https://user-images.githubusercontent.com/5449554/87745755-b985f600-c829-11ea-82c0-a16e9fb133d8.png)

Input of `imgdata2hdf5.py` : data list (text file)
Output of `imgdata2hdf5.py` : hdf5 file which can be used by `record-minmax`

jinevening:
## Example
```
python imgdata2hdf5.py \
> --input_path=tmp/imgdata/data
> --data_list=tmp/imgdata/datalist.txt
> --output_path=tmp/imgdata/imgdata.hdf5
```

## Arguments
```
  -h, --help            Show this help message and exit
  -i INPUT_PATH, --input_path INPUT_PATH
                        Path to the directory where the raw image data files are saved.
  -l DATA_LIST, --data_list DATA_LIST
                        Path to the text file which lists the raw image data files to be converted.
  -o OUTPUT_PATH, --output_path OUTPUT_PATH
                        Path to the output hdf5 file.
```

seanshpark:
Just curious. if we gather all the image files in the `input_path`, then `data_list` is needless.
if some files are tobe exluced then not all files are used so list is needed. is this the reason of `data_list` file existance?
or if `data_list` all items has absolute path then `input_path` maybe needless.

jinevening:
> if some files are tobe exluced then not all files are used so list is needed. is this the reason of data_list file existance?

Yes. data_list is for allowing users to exclude un-wanted files.

> just another thought, if data_list all items has absolute path then input_path maybe needless.

I agree with you.

> if there are hundreds or thousands images, preparing this data_list is a task.
I think not doing this task would help users...
or is there some simple way for this?

I agree that it can be a burden for some users. Using data_list is an idea from @lemma (AFAIK, SNPE quantization tool supports such an interface). @lemmaa what do you think of this issue?

lemmaa:
@jinevening , @seanshpark , shouldn't we support both `--input_path` and `--data_list`? 
Both have their own advantages. `--input_path` eliminates the inconvenience of list creation, and `--data_list` supports selective test set configuration without moving the target file. (You can mix and use data from multiple directories) 

[...]

PR #3250, opened 2020-07-17
[imgdata2hdf5] Tool to convert raw imgdata to hdf5

jinevening:
This adds imdata2hdf5, a tool to convert raw image data to hdf5

---
Related to #3213 Draft PR #3232

### DIFF
--- a/compiler/imgdata2hdf5/imgdata2hdf5.py
+++ b/compiler/imgdata2hdf5/imgdata2hdf5.py
@@ -45,7 +45,7 @@ with open(data_list, 'r') as f:
 # Input files
 num_converted = 0
 for imgdata in datalist:
-    with open(imgdata, 'r') as f:
+    with open(imgdata, 'rb') as f:
         sample = group.create_group(str(num_converted))
         num_converted += 1
         filename = os.path.basename(imgdata)
--- a/compiler/imgdata2hdf5/imgdata2hdf5.py
+++ b/compiler/imgdata2hdf5/imgdata2hdf5.py
@@ -27,12 +27,19 @@ group = h5_file.create_group("value")
 # If this flag is set in the hdf5 file, record-minmax will skip type/shape check
 group.attrs['rawData'] = '1'
 
+if os.path.isfile(data_list) == False:
+    raise SystemExit("No such file. " + data_list)
+
 # Data list
 datalist = []
 with open(data_list, 'r') as f:
     lines = f.readlines()
     for line in lines:
-        datalist.append(line.rstrip())
+        filename = line.rstrip()
+        if os.path.isfile(filename):
+            datalist.append(filename)
+        else:
+            raise SystemExit("No such file. " + filename)
 
 # Input files
 num_converted = 0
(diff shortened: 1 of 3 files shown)


seanshpark:
as a random testing, I gave one text file as an item to list and got
```
  File "imgdata2hdf5.py", line 45, in <module>
    raw_data = bytearray(f.read())
TypeError: string argument without an encoding
```
and then one binary file as an item and got
```
  File "./imgdata2hdf5.py", line 45, in <module>
    raw_data = bytearray(f.read())
  File "/.../lib/python3.5/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
```

compiler/imgdata2hdf5/imgdata2hdf5.py
```diff
@@ -0,0 +1,59 @@
...
+# Data list
+datalist = []
+with open(data_list, 'r') as f:
+    lines = f.readlines()
+    for line in lines:
+        filename = line.rstrip()
+        if os.path.isfile(filename):
+            datalist.append(filename)
+        else:
+            raise SystemExit("No such file. " + filename)
```
seanshpark, line 42: this will stop when there is an empty line or only with spaces that users can use as some delimiter

seanshpark:
Still i get
```
  File "./imgdata2hdf5.py", line 52, in <module>
    raw_data = bytearray(f.read())
TypeError: string argument without an encoding
``` 
with some dummy text file as an input file

seanshpark:
I used random text file as image file and a text list file with something like
```
/home/userid/some_random_textfile.txt

```
and an empty space at the bottom

jinevening:
@seanshpark Can you upload your input text file? So I can solve your problem

seanshpark:
> Can you upload your input text file? So I can solve your problem

Text file is random. you can use any file. You can event copy this text and use it as the input file.

compiler/imgdata2hdf5/imgdata2hdf5.py
jinevening: I changed the code to support skipping of empty lines

jinevening:
@seanshpark Sorry, I could not reproduce the below error

```
  File "./imgdata2hdf5.py", line 52, in <module>
    raw_data = bytearray(f.read())
TypeError: string argument without an encoding
```

I used random.txt, whose contents is

```
Text file is random. you can use any file. You can even copy this text and use it as the input file.
```

The output was 

```
No such file. Text file is random. you can use any file. You can even copy this text and use it as the input file
```

I also tried with data list file with 'random.txt', but it just ran well (and I think this is what we wanted).

Can you please upload your file? It can be easily attached in the comment.

seanshpark:
I used your `Text file is random. you can use any file. You can even copy this text and use it as the input file.` file and got same error.

I'm not good at python but `TypeError: string argument without an encoding` can be a hint.

seanshpark:
```
$ python --version
Python 3.5.2
```
and I'm running in TF2.3.0rc0 virtual environment

jinevening:
> $ python --version
Python 3.5.2

This was the reason. I used Python 2.7, which does not generate the error. I've fixed the problem, so now the error has gone in both Python 3.X and Python 2.7.

PR #3251, opened 2020-07-17
[record-minmax] Allow to process hdf5 of raw data

jinevening:
This allows to process hdf5 filled with raw data

---
Related to #3213 Draft PR #3232

### DIFF
--- a/compiler/record-minmax/src/HDF5Importer.cpp
+++ b/compiler/record-minmax/src/HDF5Importer.cpp
@@ -62,6 +62,10 @@ DataType toInternalDtype(const H5::DataType &h5_type)
   return DataType::Unknown;
 }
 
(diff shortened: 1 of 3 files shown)


compiler/record-minmax/src/HDF5Importer.cpp
```diff
@@ -62,6 +62,10 @@ DataType toInternalDtype(const H5::DataType &h5_type)
   return DataType::Unknown;
 }
 
+void readTensorData(H5::DataSet &tensor, uint8_t *buffer)
+{
+  tensor.read(buffer, H5::PredType::NATIVE_UINT8);
+}
```
seanshpark, line 68: please add empty line after this

PR #3268, opened 2020-07-17
[infra] Add pybind11 library to infra

jinevening:
This adds pybind11 library for dalgona project

---
Related to https://github.sec.samsung.net/STAR/nnfw/issues/11537

### DIFF
--- a/infra/cmake/packages/Pybind11Config.cmake
+++ b/infra/cmake/packages/Pybind11Config.cmake
@@ -1,0 +1,21 @@
+function(_Pybind11_import)
+  nnas_find_package(Pybind11Source QUIET)
+
+  if(NOT Pybind11Source_FOUND)
+    set(Pybind11_FOUND FALSE PARENT_SCOPE)
+    return()
+  endif(NOT Pybind11Source_FOUND)
+
+  nnas_include(ExternalBuildTools)
+  ExternalBuild_CMake(CMAKE_DIR   ${Pybind11Source_DIR}
+                      BUILD_DIR   ${CMAKE_BINARY_DIR}/externals/PYBIND11/build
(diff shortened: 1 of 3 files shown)


seanshpark:
If not urgent, I'd like this to land this after release branch takes off.

jinevening:
@seanshpark This is not urgent :) let's land this after release.

infra/cmake/packages/Pybind11Config.cmake
```diff
@@ -0,0 +1,14 @@
+function(_Pybind11_import)
+  nnas_find_package(Pybind11Source QUIET)
+
+  if(NOT Pybind11Source_FOUND)
+    set(Pybind11_FOUND FALSE PARENT_SCOPE)
+    return()
+  endif(NOT Pybind11Source_FOUND)
+
+  add_subdirectory(${Pybind11Source_DIR} ${CMAKE_CURRENT_BINARY_DIR}/pybind11)
```
parjong, line 9: I'm not sure about the convention over this ``), but ``CMAKE_CURRENT_BINARY_DIR`` may be dangerous ;-)

jinevening: I'm not very familiar with cmake. Can you explain a little bit more about the dangerous cases? It will be helpful for me to avoid such cases.

parjong: This will create a pybind11 directory for for the first ``find_pacakge`` call is made. There is no guarantee that there is no duplication ;-D

hseok-oh: @jinevening You can refer `infra/cmake/packages/GTestConfig.cmake` or `infra/cmake/packages/FlatBuffersConfig.cmake` for external library build.

PR #3276, opened 2020-07-20
[arser] Implement lexical_cast for boolean type

mhs4670go:
This commit implements lexical_cast for boolean type

### DIFF
--- a/compiler/arser/include/arser/arser.h
+++ b/compiler/arser/include/arser/arser.h
@@ -40,6 +40,13 @@ template <typename T> T lexical_cast(const std::string &str)
   ss >> data;
   return data;
 }
+template <> bool lexical_cast(const std::string &str)
+{
+  bool data = true;
+  if (str == "false" || str == "False" || str == "FALSE" || str == "0")
+    data = false;
+  return data;
+}
 
 } // namespace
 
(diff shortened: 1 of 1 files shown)


seanshpark:
Where is this used?

mhs4670go:
@seanshpark Actually, there is an error when boolean type is passed from user as of now. For bug fix.

ragmani:
@nnfw-bot test nncc-release

seanshpark:
> For bug fix.

Then it would be nice if you could add a link or add some explanation about the error.

compiler/arser/include/arser/arser.h
```diff
@@ -40,6 +40,13 @@ template <typename T> T lexical_cast(const std::string &str)
   ss >> data;
   return data;
 }
+template <> bool lexical_cast(const std::string &str)
```
seanshpark, line 43: please add an empty space in between

mhs4670go:
@seanshpark You're right.
FYI, when user "get"s the value that he registered, "get" method calls `lexical_cast`.
```cxx
template <typename T> T lexical_cast(const std::string &str)
{
  std::istringstream ss;
  ss.str(str);
  T data;
  ss >> data;
  return data;
}
```
But this implementation can't cover boolean type. This is why I've posted this PR:)

PR #3285, opened 2020-07-20
[arser] Fix bug

mhs4670go:
This commit fixes bug to cover when nargs is 0


seanshpark:
When is nargs == 0 ?

mhs4670go:
@seanshpark 
```cxx
arser.add_argument("--fuse_bcq")
      .nargs(0)
      .required(false)
      .help("This will fuse operators and apply Binary Coded Quantization");
```

This way:)

PR #3290, opened 2020-07-20
[circle-inspect] Use standard library

mhs4670go:
This commit makes circle-inspect use standard library


compiler/circle-inspect/driver/Driver.cpp
```diff
@@ -57,9 +57,9 @@ int entry(int argc, char **argv)
   std::vector<std::unique_ptr<circleinspect::DumpInterface>> dumps;
 
   if (arser["--operators"])
-    dumps.push_back(std::move(stdex::make_unique<circleinspect::DumpOperators>()));
+    dumps.push_back(std::move(std::make_unique<circleinspect::DumpOperators>()));
```
s-barannikov, line 60: `std::move` is redundant here :)

mhs4670go: Oh, This is because `make_unique` results in prvalue and `push_back` will call the `unique_ptr` move constructor when constructing the element to be appended to the vector, doesn't it?

Thanks for tip:)

I'll remove above in another PR.

s-barannikov: > Oh, This is because `make_unique` results in prvalue and `push_back` will call the `unique_ptr` move constructor when constructing the element to be appended to the vector, doesn't it?

Yes, this is exactly what happens

PR #3288, opened 2020-07-20
[circle-inspect] Introduce DumpOperatorVersion

llFreetimell:
Parent Issue : #3174
Draft : #3216

This commit will introduce DumpOperatorVersion, which dumps
the version of circle operator.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

PR #3299, opened 2020-07-20
[circle-inspect] Remove redundant std::move

mhs4670go:
This commit remoes redundant std::move


mhs4670go:
@llFreetimell Maybe you might apply this change in #3288

PR #3301, opened 2020-07-20
[luci] Export operator version

llFreetimell:
Parent Issue : #3174
Draft : #3216

This commit will enable operator version exporting

### DIFF
--- a/compiler/luci/export/src/CircleExporterUtils.cpp
+++ b/compiler/luci/export/src/CircleExporterUtils.cpp
@@ -88,15 +88,18 @@ circle::MirrorPadMode to_circle_mirrorpadmode(luci::MirrorPadMode mode)
 namespace luci
 {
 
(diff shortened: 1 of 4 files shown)


compiler/luci/export/src/CircleExporterUtils.cpp
```diff
@@ -88,15 +88,18 @@ circle::MirrorPadMode to_circle_mirrorpadmode(luci::MirrorPadMode mode)
 namespace luci
 {
 
-uint32_t SerializedModelData::registerBuiltinOpcode(circle::BuiltinOperator builtin_code)
+uint32_t SerializedModelData::registerBuiltinOpcode(circle::BuiltinOperator builtin_code,
+                                                    const int32_t op_version = 0)
```
seanshpark, line 92: please do not use default value `= 0`

llFreetimell: I will remove default value to prevent the case that someone forgot to put the value :)

PR #3302, opened 2020-07-20
[tflchef] Revise set of BuiltinCode

llFreetimell:
Parent Issue : #3174
Draft : #3216

Until now, `builtin_code_set` did not include operator version.
This commit will revise it to include the version information by
changing it from `std::set` to `std::map`.


mhs4670go:
I'm not sure but I think there are no multiple version in single model. Does it need to store its version?

Issue #3272, opened 2020-07-20
[tf2tfliteV2] Mobilenet V2 conversion fail on daily package test


hseok-oh:
Daily package generation & test job fail: `mobilenet_v2_1.0_224` nnpackage generation test by using `tools/nnpackage_tool/nncc-tc-to-nnpkg-tc/nncc-tc-to-nnpkg-tc.sh`

Input/output info file:
```
$ cat mobilenet_v2_1.0_224.info 
input,  input:0,  TF_FLOAT,   [1,224,224,3]
output, MobilenetV2/Predictions/Reshape_1:0,  TF_FLOAT,   [1,1001]
```

command: `tools/nnpackage_tool/nncc-tc-to-nnpkg-tc/nncc-tc-to-nnpkg-tc.sh -o nnpkg-tcs -i models mobilenet_v2_1.0_224`
```
03:05:02  2020-07-18 03:04:43.566398: I tensorflow/core/platform/cpu_feature_guard.cc:142] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN)to use the following CPU instructions in performance-critical operations:  AVX2 FMA
03:05:02  To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
03:05:02  2020-07-18 03:04:43.608157: I tensorflow/core/platform/profile_utils/cpu_utils.cc:104] CPU Frequency: 3096300000 Hz
03:05:02  2020-07-18 03:04:43.648506: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x5a03930 initialized for platform Host (this does not guarantee that XLA will be used). Devices:
03:05:02  2020-07-18 03:04:43.648563: I tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): Host, Default Version
03:05:03  2020-07-18 03:04:44.027042: I tensorflow/core/grappler/devices.cc:78] Number of eligible GPUs (core count >= 8, compute capability >= 0.0): 0 (Note: TensorFlow was not compiled with CUDA or ROCm support)
03:05:03  2020-07-18 03:04:44.027224: I tensorflow/core/grappler/clusters/single_machine.cc:356] Starting new session
03:05:03  2020-07-18 03:04:44.110052: I tensorflow/core/grappler/optimizers/meta_optimizer.cc:816] Optimization results for grappler item: graph_to_optimize
[...]

PR #3307, opened 2020-07-20
[one-cmds] Revise one-import-tf with both v1, v2

seanshpark:
This will revise one-import-tf to try with both v1 and then v2 options


seanshpark:
Related #3272

seanshpark:
CC @hseok-oh

seanshpark:
I think we may have to drop this but provide an option for v1 or v2

PR #3308, opened 2020-07-21
[one] Rename "..._alloc" to ".._tensor" in KernelGenerator

hyunsik-yoon:
This renames `..._alloc` to `..._tensor` in `KernelGenerator`.

Reason)
- AFAIK, the name `..._alloc` was originated from PACL. I current n onert, `..._alloc` actually means `..._tensor` so `..._tensor` is better name. `..._alloc` makes code reading difficult since reader assume that it is some kind of allocated memory.
- `KernelGenerator` runs at compile time and it is **not** always true that all tensor buffer was be allocated when the tensor is a dynamic tensor. Therefore, the name `alloc` is not always correct.

### DIFF
--- a/runtime/onert/backend/acl_cl/KernelGenerator.cc
+++ b/runtime/onert/backend/acl_cl/KernelGenerator.cc
@@ -76,15 +76,15 @@ void KernelGenerator::visit(const ir::operation::BatchToSpaceND &node)
   const auto block_size_index{
       node.getInputs().at(ir::operation::BatchToSpaceND::Input::BLOCK_SIZE)};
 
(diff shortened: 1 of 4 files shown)


wateret:
LGTM, but what about merging this after create a branch for 1.7.0? Otherwise it is hard for our internal repo to sync up with master as it has too many changes to cpu backend.

hyunsik-yoon:
@wateret no problem. please merge when 1.7.0 work is done. :-)

wateret:
I will approve once the branch is created. I guess it is going to be done within an hour.

PR #3326, opened 2020-07-21
[circle2circle] Remove redundant code

mhs4670go:
This commit removes redundant codes in circle2circle


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -33,15 +33,12 @@
 #include <map>
```
s-barannikov, line 33: No longer used

PR #3332, opened 2020-07-21
[res] Add DepthwiseConv2D testcase which has version 2

llFreetimell:
Parent Issue : #3174
Draft : #3216

This commit will add `DepthwiseConv2D` test case which version 2 is included.

### DIFF
--- a/res/TensorFlowLiteRecipes/DepthwiseConv2D_003/test.recipe
+++ b/res/TensorFlowLiteRecipes/DepthwiseConv2D_003/test.recipe
@@ -1,0 +1,44 @@
+operand {
+  name: "ifm"
+  type: FLOAT32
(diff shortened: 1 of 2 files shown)


seanshpark:
Have you checked with `compiler/common-artifacts/exclude.lst` file?

llFreetimell:
@seanshpark Yes, I planned to add below codes.
https://github.com/Samsung/ONE/blob/2f81517a767a8ec22b4912ffaaccaeb997e4f766/compiler/common-artifacts/exclude.lst#L47

And because they are different module, I thought that I should add recipe first and then update list.
However, @mhs4670go said to me the order should be reversed.
I will create that PR first and then rebase this after landing :)

PR #3341, opened 2020-07-21
[tfl-verify] Use arser

mhs4670go:
This commit makes tfl-verify use arser

Related : #3001 

### DIFF
--- a/compiler/tfl-verify/src/Driver.cpp
+++ b/compiler/tfl-verify/src/Driver.cpp
@@ -35,7 +35,7 @@ int entry(int argc, char **argv)
   {
     std::cout << err.what() << std::endl;
     std::cout << arser;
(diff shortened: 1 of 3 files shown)


compiler/tfl-verify/src/Driver.cpp
```diff
@@ -16,22 +16,31 @@
...
-    std::cerr << std::endl;
-    std::cerr << "USAGE: " << argv[0] << " [tflite]" << std::endl;
-    return 255;
+    arser.parse(argc, argv);
   }
+  catch (const std::runtime_error &err)
+  {
+    std::cout << err.what() << std::endl;
+    std::cout << arser;
+    return 0;
```
jinevening, line 38: Maybe return 255?

PR #3349, opened 2020-07-21
[tflchef] change return value

mhs4670go:
This commit changes return value when an error is occurred


mhs4670go:
@nnfw-bot test nncc-debug

PR #3354, opened 2020-07-21
[onert] UserTensor::is_constant returns false

wateret:
This commit makes `UserTensor::is_constant` return false.


wateret:
+ UserTensor is a tensor from users which are model inputs and outputs. So it is always non-const.

PR #3363, opened 2020-07-21
[vconone] Introduce version provider

seanshpark:
This will introduce version number provide vconone library for one-* commands and command line tools


parjong:
(Optional) ``vcon`` looks better for me than ``vconone`` as I already told you offline ;-)

seanshpark:
For #3132 , from draft #3161

PR #3365, opened 2020-07-22
[one-cmds] revise exit code

seanshpark:
This will revise exit code of one-* commands to exit 255 for usage


seanshpark:
For #3132 , from draft #3161

PR #3369, opened 2020-07-22
[one-cmds] use command name

seanshpark:
This will revise one-codegen and one-import to show it's explicit name
- before will show full path which is unnecessary to show the path


seanshpark:
For #3132 , from draft #3161

PR #3372, opened 2020-07-22
[onert] Optimization for model input copies

wateret:
Do not insert Permute operation when tensors are copied from controlflow
to cpu(model input cases).

### DIFF
--- a/runtime/onert/core/src/compiler/ExecutorFactory.cc
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.cc
@@ -201,16 +201,33 @@ ExecutorFactory::initializeModelIOTensors(ir::LoweredGraph &lowered_graph,
     // Add tensor to controlflow TensorRegistry.
     cf_tensor_builder->setUserTensor(ind, tensor);
     ret.push_back(tensor);
-
-    // Set other tensors as external tensors
-    for (auto &tensor_builder : tensor_builders)
-    {
-      // FIXME This is a workaround registering all user tensors to all backends
-      // FIXME Handle when it is failed
-      tensor_builder->setExternalTensor(ind, tensor);
-    }
   }
   return ret;
+}
+
+void ExecutorFactory::prepareExternalTensors(ir::LoweredGraph &lowered_graph,
+                                             TensorBuilders &tensor_builders)
+{
+  lowered_graph.op_seqs().iterate(
+      [&](const ir::OpSequenceIndex &op_seq_index, const ir::OpSequence &op_seq) {
+        auto lower_info = lowered_graph.getLowerInfo(op_seq_index);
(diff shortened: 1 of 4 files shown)


runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -201,18 +201,35 @@ ExecutorFactory::initializeModelIOTensors(ir::LoweredGraph &lowered_graph,
...
-      tensor_builder->setExternalTensor(ind, tensor);
-    }
   }
   return ret;
 }
 
+void ExecutorFactory::prepareExternalTensors(ir::LoweredGraph &lowered_graph,
+                                             TensorBuilders &tensor_builders)
+{
+  lowered_graph.iterateTopolOpSeqs(
```
ragmani, line 211: Does this method require topological sort of `OpSequences`?

wateret: I think not.

wateret:
Before/After example

Screen Shot 2020-07-15 at 9 20 40 PM → Screen Shot 2020-07-22 at 2 00 57 PM

runtime/onert/core/src/compiler/ExecutorFactory.cc
ragmani: If so, you can use `op_seqs()` of `LoweredGraph`  instead of `iterateTopolOpSeqs()`

runtime/onert/core/src/ir/pass/PermutationInsertionPass.cc
```diff
@@ -62,27 +62,22 @@ void PermutationInsertionPass::callback(const OperandIndex &index, Operand &obje
     auto insert_set = operand_li->use_factors() - operand_li->def_factors();
     auto def_factor = operand_li->def_factors().getOnlyElement();
 
-    auto compatible_backends = [](auto /* backend1 */, auto /* backend2 */) {
-      // TODO If other issues for Permute elimination are resolved, enable this
-      return false;
-      /*
+    auto compatible_backends = [](auto backend1, auto backend2) {
```
hyunsik-yoon, line 65: IMHO this function is confusing because cpu backend and cf backend are compatible but the reverse is not compatible.

wateret: This is temporary and it is going to be compatible in both ways. I'll add more description

runtime/onert/core/src/compiler/ExecutorFactory.cc
wateret: Ok thanks!

wateret:
@hyunsik-yoon @ragmani PTAL. Thank you for the review!

PR #3377, opened 2020-07-22
[infra] Add vconone to compiler_modules

seanshpark:
This will add vconone module to compiler_modules list and nnpkg_resources list


seanshpark:
For #3132 , from draft #3161

PR #3381, opened 2020-07-22
[record-minmax] apply ENABLE_TEST switch

seanshpark:
This will enable test when ENABLE_TEST is set


seanshpark:
For #3132 , from draft #3161

PR #3382, opened 2020-07-22
[tflite2circle] show version

seanshpark:
This will provide --version argument to show version information


seanshpark:
For #3132 , from draft #3161

PR #3386, opened 2020-07-22
[onert] cpu: Refine Compare kernel dispatching

wateret:
Refine cpu backend's CompareLayer kernel dispatching to remove code
duplications. Changing switch-case to array-lookup.

This is for better score on CodeDuplication of SAM.

### DIFF
--- a/runtime/onert/backend/cpu/ops/CompareLayer.cc
+++ b/runtime/onert/backend/cpu/ops/CompareLayer.cc
@@ -17,6 +17,7 @@
 
 #include "OperationUtils.h"
 
+#include <assert.h>
 #include <cker/operation/Comparison.h>
 using namespace nnfw::cker;
 namespace onert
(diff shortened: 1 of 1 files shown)


wateret:
@dr-venkman PTAL

runtime/onert/backend/cpu/ops/CompareLayer.cc
```diff
@@ -52,95 +53,40 @@ void compareQuant8(const IPortableTensor *lhs, const IPortableTensor *rhs, IPort
...
+      Broadcast4DSlowEqualWithScaling,   Broadcast4DSlowNotEqualWithScaling,
+      Broadcast4DSlowGreaterWithScaling, Broadcast4DSlowGreaterEqualWithScaling,
+      Broadcast4DSlowLessWithScaling,    Broadcast4DSlowLessEqualWithScaling,
+  };
+  static const CompareFunction non_broadcast_fns[] = {
+      EqualWithScaling,        NotEqualWithScaling, GreaterWithScaling,
+      GreaterEqualWithScaling, LessWithScaling,     LessEqualWithScaling,
+  };
+
+  // Assumes these enum values to be in the order like this
```
dr-venkman, line 71: I agree with the overall approach, but having to check the conditions below everytime on execution is something to optimize. Will shifting lines 72 -- 78 to `configure` function, change the overhead? Similar argument for `broadcast_fns` and `non_broadcast_fns` could be considered.

wateret: `static_assert` is a compile time check, so it does not run at runtime.

wateret:
> `compareScalar` could be refined as well, by analogy.

You're right. Will work on that too when this gets landed.

runtime/onert/backend/cpu/ops/CompareLayer.cc
dr-venkman: Looks good then :).

PR #3392, opened 2020-07-22
[tflite_run_2_2_0] Fix android build error for tflite_run_2_2_0

periannath:
- This commit fixes android build error for tflite_run_2_2_0
  - `cannot find -lrt` error occurs for android build
  - `-lrt` link option should be removed since librt is included in bionic lib
  - Add `-lrt` option when target is not android


periannath:
ping @Samsung/nnfw_committers

PR #3396, opened 2020-07-22
DRAFT revise nncc-tc-to-nnpkg-tc

seanshpark:
This will add --v2 option to use TF 2.x interface in tf2nnpkg and revise nncc-tc-to-nnpkg-tc
- current tc models will correctly convert with TF 1.x interface
- there is no .v2 file so it will be TF 1.x

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@samsung.com>

PR #3398, opened 2020-07-23
[infra] revise tf2nnpkg.20200630

seanshpark:
This will revise tf2nnpkg.20200630 usage to split in multiple lines
- return 255 as showing usage does not do what it should do
- remove [OPTION] as other options are not handled and/or -o is not optional


seanshpark:
From #3396

PR #3400, opened 2020-07-23
[infra] introduce --v2 option to tf2nnpkg.20200630

seanshpark:
This will introduce --v2 option to tf2nnpkg.20200630 and use --v1 option as default


seanshpark:
For #3272, from draft #3396

PR #3405, opened 2020-07-23
[tools] revise nncc-tc-to-nnpkg-tc

seanshpark:
This will revise nncc-tc-to-nnpkg-tc to call tf2nnpkg with --v2 option if modelfilename.v2 file exist


seanshpark:
For #3272, from draft #3396

ragmani:

ragmani:

PR #3411, opened 2020-07-23
[doc] guide to use ln insteand of cp for Makefile.template

glistening:
Currently build guide says use `cp -n ` for Makefile.
It will guide to use `ln -s` so that they don't need to
update Makefile continually.

### DIFF
--- a/docs/howto/how-to-build-runtime.md
+++ b/docs/howto/how-to-build-runtime.md
@@ -13,7 +13,7 @@ In the Ubuntu, you can easily install it with the following command.
 
 ```
 $ sudo apt-get install cmake libboost-all-dev
-``` 
+```
 
 If your linux system does not have the basic development configuration, you will need to install more packages. A list of all packages needed to configure the development environment can be found in the https://github.com/Samsung/ONE/blob/master/infra/docker/Dockerfile.1804 file.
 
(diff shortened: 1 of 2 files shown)


lemmaa:
@glistening , the purpose seems to be different. AFAIK, the reason for using `cp -n` is to keep the local version regardless of the update in git repo. :)

glistening:
@lemmaa I've made this PR because I heard some member has trouble by not-updating `Makefile` after copying `Makefile.template` and I've never updated `Makefile.template`. I am not sure what is majority.

hseok-oh:
@lemmaa @glistening Then we can guide to use `make -f Makefile.template`

lemmaa:
This is part of how to use make tool, and I don't know if we should go into that detail. :( 
If there have such a happening and need to resolve it, but I don't think so, I prefer @hseok-oh 's way.

wateret:
I'm OK with either way. But I also think, if the official guide suggests linking, then what about just renaming `Makefile.template` to `Makefile`?

wateret:
AFAIR `Makefile.template` is introduced for CI. If it is correct, it would be much better to remove uses of Makefile and replace with `./nnfw`.

glistening:
I like @hseok-oh's suggestion. For @wateret's suggestion, we may create another issue because it is related but beyond this PR's topic and it may take a long to get conclusion.

Issue #3374, opened 2020-07-22
Compiler FE: check exclude list of common-artifacts


mhs4670go:
As a result of making `common-artifacts`, I got to know there are a few bugs or NYI stuffs.

Let's update one more time to find out more detailed reason.

PR #3412, opened 2020-07-23
[res] Fix TransposeConv_000 bug

mhs4670go:
This commit fixes TransposeConv_000 bug

Related : #3374 


seanshpark:
Can you provide more about the bug? And it would be nice to provide kernel shape information like `IOHW` for TransposedConv.

mhs4670go:
FYI, in tensorflow, filter is a 4-D Tensor with the same type as input and shape [height,width, output_channels, in_channels].
But, in tensorflow lite, it is changed to [output_channels, height, width, in_channels].

PR #3413, opened 2020-07-23
Use RUY_THREADS=1

chunseoklee:
Many applications shows better performance with RUY_THREADS=1


periannath:
@chunseoklee How about changing `kDefaultNumThreadpoolThreads` variable in `RuySupport.h`? That variable is used when number of ruy thread is not given by environment variable.

chunseoklee:
Updated as commented

PR #3415, opened 2020-07-23
[luci] Add validation to CircleTransposeConv

mhs4670go:
This commit adds validation to CircleTransposeConv

Related : #3374 

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleTransposeConv.cpp
+++ b/compiler/luci/import/src/Nodes/CircleTransposeConv.cpp
@@ -30,6 +30,13 @@ bool CircleTransposeConvGraphBuilder::validate(const ValidateArgs &args) const
   if (args.op.inputs.size() != 3)
     return false;
 
+  const auto &inputs = args.op.inputs;
+  const auto &tensors = args.reader.tensors();
+  const auto &filter_tensor = tensors.at(inputs[1]);
+  const auto &ifm_tensor = tensors.at(inputs[2]);
+  if (ifm_tensor.get()->shape.at(3) != filter_tensor.get()->shape.at(3))
+    return false;
+
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


compiler/luci/import/src/Nodes/CircleTransposeConv.cpp
```diff
@@ -30,6 +30,13 @@ bool CircleTransposeConvGraphBuilder::validate(const ValidateArgs &args) const
   if (args.op.inputs.size() != 3)
     return false;
 
+  const auto &inputs = args.op.inputs;
+  const auto &tensors = args.reader.tensors();
+  const auto &filter_tensor = tensors.at(inputs[1]);
+  const auto &ifm_tensor = tensors.at(inputs[2]);
+  if (ifm_tensor.get()->shape.at(3) != filter_tensor.get()->shape.at(3))
```
seanshpark, line 37: please use rank value not `3`

mhs4670go: @seanshpark 
Is it what I think it means?
```suggestion
  auto rank = ifm_tensor.get()->shape.size();
  if (ifm_tensor.get()->shape.at(rank - 1) != filter_tensor.get()->shape.at(rank - 1))
```

seanshpark: > Is it what I think it means?

I think so. If this is the condition for TransposeConv. If we check with `3` then this code will break if rank is 2 or 3

seanshpark: I don't know the detail but is `ifm_tensor.get()->shape.at(rank - 1) != filter_tensor.get()->shape.at(rank - 1)` must hold?

mhs4670go: Well, as [document](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d_transpose#args) said, ifm and filter must be 4-D tensor. So, I think rank is always 4.

seanshpark: OK then let's add rank check too if it's 4 for ifm and filter.

PR #3421, opened 2020-07-23
[onert] Ignore bias tensor with all zero values in fully connected layer

periannath:
- This commit ignores bias tensor with all zero values in fully connected layer
  - For bias tensor with values, fully connected copies its value into output tensor
  - If bias tensor has all zero values, we can ignore that bias tensor to remove memcpy
  - Finds bias tensor with all zero values in prepare phase and mark it as optional tensor

### DIFF
--- a/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
+++ b/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
@@ -18,6 +18,7 @@
 
 #include "../Tensor.h"
 #include <cker/operation/FullyConnected.h>
+#include <cker/TensorUtils.h>
 
 namespace onert
 {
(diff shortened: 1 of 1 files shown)


runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
```diff
@@ -167,6 +168,12 @@ void FullyConnectedLayer::run()
 
 void FullyConnectedLayer::prepare()
 {
+  const int bias_size = getTensorShape(_bias).FlatSize();
```
YongseopKim, line 171: Hmm.. if _bias is nulltpr, it could cause segment fault.

periannath: @YongseopKim Thanks! I've fixed this issue.

periannath:
It seems the bias value is not used in nnapi. I'll check and fix it.

periannath:
@YongseopKim I've fixed CI error. PTAL

PR #3427, opened 2020-07-23
[tools] make update-version work on both linux and macOS

glistening:
- BSD sed, which comes from macOS, does not allow "-i" without
passing extension. Thus, sed command is replaced with perl.
- bash read doesn't work without quoting string.

### DIFF
--- a/tools/update_version/update-version
+++ b/tools/update_version/update-version
@@ -40,11 +40,12 @@ fi
 
 version=$1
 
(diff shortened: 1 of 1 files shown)


glistening:
@chunseok I've found there is no portable solution using `sed` that works on both `ubuntu` and `macOS`. Thus, `update-version` will use `perl`, which works on both `ubuntu` and `macOS` and comes without additional installation.

Issue #1088, opened 2020-05-19
[luci-interpreter] Unit testing TODOs


s-barannikov:
* Add tests for KernelBuilder.
* Check output shapes in kernel tests.
* Improve checks for consistency of types and shapes in `configure` and `execute`.
* Add negative kernel tests.

struss:
@s-barannikov Is it okay for you if I create functions in testutil for output tensor shape check?

s-barannikov:
@struss 
Shure. I just have one wish: it would be nice to see what exactly differs in shapes in case a test fails.

struss:
@s-barannikov okay. I will make a draft PR for this. thanks a lot.

seanshpark:
Left items: `Add negative kernel tests.`

PR #3433, opened 2020-07-24
[luci-interpreter] Add more `KernelBuilder` tests

s-barannikov:
Issue #1088.

Add more tests of `KernelBuilder`.

### DIFF
--- a/compiler/luci-interpreter/src/loader/KernelBuilder.cpp
+++ b/compiler/luci-interpreter/src/loader/KernelBuilder.cpp
@@ -492,6 +492,17 @@ std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleSqueeze *node)
   return std::make_unique<kernels::Squeeze>(input, output, params);
 }
 
+std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleTranspose *node)
+{
+  assert(node->arity() == 2);
+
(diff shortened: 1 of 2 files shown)


seanshpark:
Please add negative tests too with `_NEG` suffix in the name.

s-barannikov:
@seanshpark 
Provided the correct circle model, the functions being tested will never fail. I can't see what negative tests should look like. Can you suggest something?

seanshpark:
> Can you suggest something?

I haven't reviewed the codes. I also can't suggest. It's about our SE requirements to provide negative testing.
If we keep adding only postive tests, our project may be rejected.

seanshpark:
You can add negative tests in another PR but please add within our project timeline.

s-barannikov:
> It's about our SE requirements to provide negative testing.
If we keep adding only postive tests, our project may be rejected.

OK, I will bear this in mind.

jinevening:
IMHO, one way to make a negative test is to make a test expected to throw an exception (caused by wrong behaviors).

Building a kernel for `CircleInput`, `CircleOutput`, or `CircleConst` will throw an exception. It would be possible to make negative tests for those Ops using EXPECT_THROW.

s-barannikov:
@jinevening Thank you for the idea. I will add them in other PR.

PR #3440, opened 2020-07-24
[luci] Add CONSTANT OpCode

mhs4670go:
This commit adds CONSTANT OPCode. It will be replaced soon


seanshpark:
This originated from Windows build. Windows has error where `CONST` is compiled as `const` keyword.

Issue #3218, opened 2020-07-16
[Compiler/luci] 'quantized_dimension' info not passed


cgbahk:
For CWQ*, `quantized_dimension` in tflite/circle schema is essential. 

(*Channle-wise quantization, or per-axis quantization. See [tensorflow/quantization_spec.md](https://github.com/tensorflow/tensorflow/blob/9c20fbff6c251b05c15f2b4f840722c78bf38233/tensorflow/lite/g3doc/performance/quantization_spec.md#per-axis-vs-per-tensor))

https://github.com/Samsung/ONE/blob/74682f8bd5f74dddf46254ec81f7762e276e4d0f/nnpackage/schema/circle_schema.fbs#L83-L91

## Issue
But as I understand, `quantized_dimension` is not dealt with in luci or any compiler projects yet.
```
$ git grep quantized_dimension -- compiler

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h:// TODO Add 'quantized_dimension' field for per-channel case when IR provides it.
compiler/mir/src/mir_tflite_importer/schema/schema.fbs:  quantized_dimension:int;
```
https://github.com/Samsung/ONE/blob/74682f8bd5f74dddf46254ec81f7762e276e4d0f/compiler/luci/import/src/CircleReader.cpp#L152-L173
https://github.com/Samsung/ONE/blob/74682f8bd5f74dddf46254ec81f7762e276e4d0f/compiler/luci/export/src/CircleTensorExporter.cpp#L305

As I understand, we can imply this value from operator type as a work-around(e.g. 3 for DEPTHWISE_CONV_2D, 0 for CONV_2D), but this approach might be dangerous.

jinevening:
> As I understand, we can imply this value from operator type as a work-around(e.g. 3 for DEPTHWISE_CONV_2D, 0 for CONV_2D), but this approach might be dangerous.

FYI, `circle-quantizer` is taking the above approach. https://github.com/Samsung/ONE/blob/master/compiler/luci/pass/src/QuantizationUtils.cpp#L108

jinevening:
To support channel-wise quantization (including quantized_dimension) in luci-interpreter, we need to change `class Tensor`. 

Currently, `Tensor` has following functions.

```cpp
  float scale() const
  {
    assert(_quantization.scale.size() == 1);
    return _quantization.scale[0];
  }

  float zero_point() const
  {
    assert(_quantization.zero_point.size() == 1);
[...]

PR #3442, opened 2020-07-24
[luci-interpreter] Add quantized_dimension to luci-interpreter

jinevening:
This adds quantized_dimension to luci-interpreter

---
Related to #3218

### DIFF
--- a/compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
@@ -79,12 +79,11 @@ private:
 //
 // Note that due to historical and performance reasons, per-tensor quantization uses unsigned
 // integer types, while per-channel uses signed types assuming 'zero_point' == 0.
-//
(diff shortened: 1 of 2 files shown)


compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
```diff
@@ -96,17 +95,15 @@ class Tensor
 
   const Shape &shape() const { return _shape; }
 
-  float scale() const
-  {
-    assert(_quantization.scale.size() == 1);
-    return _quantization.scale[0];
-  }
+  float scale() const { return _quantization.scale[0]; }
```
seanshpark, line 98: You've removed `assert(_quantization.scale.size() == 1);`. Is this safe?

jinevening: It would be safer to leave the assert. I've updated the code. Thanks.

jinevening:
@s-barannikov PTAL

PR #3443, opened 2020-07-24
[infra/dockerfile] Update Dockerfile

hseok-oh:
- Remove ENV for proxy: proxy is pre-defined argument
- Upgrade pip after install: remove fail when install numpy


hseok-oh:

PR #3449, opened 2020-07-24
[Compiler] Compiler bugfix in 1.6.1

llFreetimell:
This commit will fix following bugs
- `circle2circle` never ended because of `ResolveCustomOpBatchMatMulPass`
    - When custom op is not `BatchMatMul`, `changed` is always true.
- BCQ is applied incorrectly

### DIFF
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -38,6 +38,16 @@ const std::string node_name_prefix(luci::NodeName node_name)
 {
   std::string prefix = node_name;
 
+  if (prefix.find("ReadVariableOp/resource/") != std::string::npos)
+  {
+    const auto start_index = prefix.find("ReadVariableOp/resource/");
+
+    const auto left_prefix = prefix.substr(0, start_index);
+    const auto right_prefix = prefix.substr(start_index + 24);
+
+    prefix = left_prefix + right_prefix;
+  }
+
   if (prefix.find("Tensordot/") != std::string::npos)
   {
     const auto index = prefix.find("Tensordot/");
@@ -47,6 +57,11 @@ const std::string node_name_prefix(luci::NodeName node_name)
   {
     const auto index = prefix.find("kernel/");
     prefix = prefix.substr(0, index - 1);
+  }
+  else if (prefix.find("/bcqinfo_") != std::string::npos)
+  {
+    const auto index = prefix.find("/bcqinfo_");
+    prefix = prefix.substr(0, index);
   }
 
   return prefix;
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -107,7 +122,13 @@ class BCQConverter final
...
   {
     const auto prefix = node_name_prefix(node->name());
-    return _do_w_x[prefix]->at<loco::DataType::BOOL>(0);
+
+    if (_do_w_x[prefix]->dtype() == loco::DataType::S32)
+      return _do_w_x[prefix]->at<loco::DataType::S32>(0) == 1;
+    else if (_do_w_x[prefix]->dtype() == loco::DataType::BOOL)
+      return _do_w_x[prefix]->at<loco::DataType::BOOL>(0);
+    else
+      throw std::runtime_error("do_w_x should be int or bool");
```
seanshpark, line 131: stop the program is not good. can you just return false?

compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -268,7 +289,7 @@ bool FuseBCQPass::run(loco::Graph *g)
         bcq_gather->input_clusters(converter.packed_clusters(params));
 
         const auto binary_hidden_size =
-            loco::must_cast<luci::CircleConst *>(bcq_gather->input_binary())->dim(0).value() * 32;
+            loco::must_cast<luci::CircleConst *>(bcq_gather->input_binary())->dim(1).value() * 32;
```
seanshpark, line 292: what is 0 and 1? please add a comment what this index is.

compiler/luci/pass/src/FuseBCQPass.cpp
llFreetimell: I agree that stopping the program is not good :)

However, if `false` is returned, it can cause unintended behavior because we don't know what was original intention.
For example, if `dtype` is `FLOAT32` and the value was 0.5, it is not `true` nor `false`. Thus I thought that user should be notified BCQ cannot be applied.

I can change the codes to return `false` at `run()` function when the information is invalid, but if then, should I apply the diff to 1.7.0 either? Because this code is included at 1.7.0 too :(

compiler/luci/pass/src/FuseBCQPass.cpp
llFreetimell: I will add :)
But should I apply this to 1.7.0 too?

seanshpark: > But should I apply this to 1.7.0 too?

Yes please

compiler/luci/pass/src/FuseBCQPass.cpp
seanshpark: For our optimization pass, return false is do nothing. I wish this will apply here also.

> should I apply the diff to 1.7.0 either?

yes please.

llFreetimell: > For our optimization pass, return false is do nothing

I said about `bool do_w_x(luci::CircleConst *node)`, not optimization pass :)
In short, if `do_w_x` returns `false`, it means that we would ignore unexpected behavior.

Anyway, I will change related codes.

seanshpark:
@chunseoklee  , please merge :)

PR #3455, opened 2020-07-24
[luci-interpreter] Remove useless DataType setting in tests

s-barannikov:
The data type does not matter for the purpose of the tests.

### DIFF
--- a/compiler/luci-interpreter/src/loader/KernelBuilder.test.cpp
+++ b/compiler/luci-interpreter/src/loader/KernelBuilder.test.cpp
@@ -43,28 +43,24 @@ using namespace testing;
 class KernelBuilderTest : public Test
 {
 protected:
-  luci::CircleInput *createInputNode(loco::DataType dtype)
-  {
-    return createNode<luci::CircleInput>(dtype);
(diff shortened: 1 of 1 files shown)


compiler/luci-interpreter/src/loader/KernelBuilder.test.cpp
```diff
@@ -43,28 +43,24 @@ using namespace testing;
...
-  template <typename NodeT> NodeT *createNode(loco::DataType output_type)
+  template <typename NodeT, typename... Args> NodeT *createNode(Args &&... args)
   {
-    auto *node = _graph.nodes()->create<NodeT>();
-    node->dtype(output_type);
+    auto *node = _graph.nodes()->create<NodeT>(std::forward<Args>(args)...);
+    // The actual type does not matter for the purpose of the tests.
+    // NOTE The type is meaningless for nodes with multiple outputs (corresponding *Out nodes carry
+    //  actual output types).
+    node->dtype(loco::DataType::FLOAT32);
```
binarman, line 54: If this type is meaningless, why set it?

PR #3457, opened 2020-07-26
[luci-value-tests] enable tests

seanshpark:
This will enable test in luci-value-tests that is now working


seanshpark:
Related #3438

Issue #3444, opened 2020-07-24
[res] Add U8 recipe: L2_NORMALIZATION


underflow101:
Parent Issue: #1880 

Let's enable U8 recipe for L2_Normalization Op.

Issue #1880, opened 2020-06-04
Compiler FE: Adding U8 recipies


seanshpark:
As you can see in https://github.com/Samsung/ONE/tree/master/res/TensorFlowLiteRecipes folder, there are sub folders of recipe that we support(some are work in progress).
- recipe is a protocol buffer file, a soure file to generate TF lite file
- tflchef is a tool that actually generates the file
- we use human readable text source so that editing is easy only with a text editor

There are two data types that we support in major: Float32 and Uint8.
All the Ops have Float32. But some have Uint8 recipe.

This task is to create Uint8 recipes that are missing
- https://www.tensorflow.org/lite/performance/quantization_spec shows list of Ops that TensorFlow defined.
- This document provides specifications for Int8 but we use **Uint8**.
- Related paper with **Uint8**: https://arxiv.org/abs/1712.05877
  - _Quantization and Training of Neural Networks for Efficient Integer-Arithmetic-Only Inference_

Things to do
- Write `test.recipe` file. You can copy-paste from existing ones
- Add item to https://github.com/Samsung/ONE/blob/master/compiler/luci/tests/test.lst file
   - configure - build - test should pass
- Fix problems if anything goes wrong
- One recipe at a time, that is one recipe per PR

---

You can also fire an issue (optional), `Add` for example
```
[res] Add U8 recipe: ADD

Let's enable U8 recipe for Add Op
```
This will let others, including me, know that one is already working on it.

Steps (not mandatory but usually this may happen) 
- Create an issue with title like `[res] Add U8 recipe: ADD` for `ADD` Op, and some descriptions
- It would be nice if you work on one Op at a time to concentrate on the operator
- I'll update the table of this issue that others know you are working on like #1606 (I'll be notified by github from the new issue)
- create a PR and work on this topic; update PR with Work In Progress codes and this may help getting faster feedbacks
- get a review from guys working on this topic, like @jinevening
- apply feedback changes if any
- land the PR and done !

How to test
```
./nncc configure
./nncc build
./nncc test
```

seanshpark:
Operators in https://www.tensorflow.org/lite/performance/quantization_spec 
Operator | status
-- | --
ADD |
AVERAGE_POOL_2D |
CONCATENATION |
CONV_2D |
DEPTHWISE_CONV_2D |
FULLY_CONNECTED |
L2_NORMALIZATION |
LOGISTIC |
MAX_POOL_2D |
MUL |
RESHAPE |
[...]

PR #3459, opened 2020-07-27
[res/TensorFlowLiteRecipes] Add L2Normalize_U8_000

underflow101:
Parent Issue: #1880
Fired Issue: #3444 

This commit adds recipes for L2Normalize_U8_000. (Derived from SOS Mine Project)

Please review this Work in Progress codes @seanshpark , @jinevening
I'll be happy to get feedback and change to make improvement.

Thank you.

Signed-off-by: underflow101 <ikarus125@gmail.com>

### DIFF
--- a/res/TensorFlowLiteRecipes/L2Normalize_U8_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/L2Normalize_U8_000/test.recipe
@@ -1,0 +1,22 @@
+operand {
+  name: "ifm1"
+  type: UINT8
+  shape { dim: 1 dim: 4 dim: 4 dim: 3 }
+  quant { min: 0 max: 255 scale: 0.0078125 zero_point: 128}
+}
+operand {
+  name: "ofm"
+  type: UINT8
+  shape { dim: 1 dim: 4 dim: 4 dim: 3 }
+  quant { min: 0 max: 255 scale: 0.0078125 zero_point: 128}
+}
+operation {
+  type: "L2Normalize"
+  l2norm_options {
+    activation: NONE
+  }
+  input: "ifm1"
+  output: "ofm"
+}
+input: "ifm1"
+output: "ofm"
--- a/res/TensorFlowLiteRecipes/L2Normalize_U8_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/L2Normalize_U8_000/test.recipe
@@ -2,13 +2,13 @@ operand {
   name: "ifm1"
   type: UINT8
   shape { dim: 1 dim: 4 dim: 4 dim: 3 }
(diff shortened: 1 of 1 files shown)


res/TensorFlowLiteRecipes/L2Normalize_U8_000/test.recipe
```diff
@@ -0,0 +1,22 @@
+operand {
+  name: "ifm1"
+  type: UINT8
+  shape { dim: 1 dim: 4 dim: 4 dim: 3 }
+  quant { min: 0 max: 255 scale: 0.0078125 zero_point: 128}
```
jinevening, line 5: It would be better to make `scale` 1.

In `circle-quantizer` (and others doing affine quantization), `scale` is calculated as below.

scale = (max - min) / (qmax - qmin)

Since this recipe is for u8 quantization, qmax = 255 and qmin = 0. So, scale would be `(max - min) / 255`.

underflow101: I'm sorry, I made a mistake. I should fix min & max
Scale should be fixed to 0.0078125 for 2 reasons:
    1. According to guide([Tensorflow Quantization](https://www.tensorflow.org/lite/performance/quantization_spec)),
        L2Normalization has scale of (1.0 / 128.0)
    2. If I set scale to 1.0, ./nncc build fails and outputs error message: `testDataGenerator: /home/bearpaek/data/git/opensource/ONE/compiler/luci-interpreter/src/kernels/L2Normalize.cpp:42: virtual void luci_interpreter::kernels::L2Normalize::configure(): Assertion `output()->scale() == (1. / 128.)' failed.`

So I think it is reasonable to keep scale as 0.0078125, and change max to 2.

I will commit again after changing max.

Thank you for your kind guidance, @jinevening .

jinevening: Nice finding. We should use the scale 0.0078125, as you mentioned. The exact value of max is 255 * 0.0078125 = 1.9921875, but it would be ok to just use 2 (min/max are typically not important).

TensorFlow restricts the scale and zero point of several Ops.

log_softmax: {16.0/256.0, 127}  ({scale, zp})
logistic: {1 / 256.0, -128}
l2_normalization: {1 / 128.0, 0}
softmax: {1 / 256.0, -128}
tanh: {1 / 128.0, 0}

If you are going to make a recipe for one of those operators, please use the specified values as you did in this PR. Thanks.

underflow101: Thank you for comment @jinevening ,
I will do the same for the rest of the recipes! :)

PR #3229, opened 2020-07-16
[onert] Introduce StaticTensorManager for cpu backend

YongseopKim:
Introduce StaticTensorManager for cpu backend. Compared to
cpu_common::StaticTensorManager, cpu::StaticTensorManager doesn't use
const tensor manager because it handles const tensors as ExternalTensor.
**For now, it is not used but will be soon.**

ONE-DCO-1.0-Signed-off-by: Yongseop Kim <yons.kim@samsung.com>

Issue #3039, opened 2020-07-10
[onert] Enable const backend tensor to use Data from external on the only cpu backend


YongseopKim:
Parent issue: https://github.com/Samsung/ONE/issues/2580
draft: https://github.com/Samsung/ONE/pull/3037

Enable const backend tensor on cpu backend to use Data from external while not allocating any memory for the tensor.

Current `cpu_common` is used by cpu, controlflow and an internal backend together. If `cpu_common` is fixed for using data from external, I'm not sure the controlflow and the internal backend would be fine. Therefore, I'll make it on the only cpu backend.

YongseopKim:
Current `cpu_common` is used by cpu, controlflow and an internal backend together. If `cpu_common` is fixed for using data from external, I'm not sure the controlflow and the internal backend would be fine. Therefore, I'll make it on the only cpu backend.

To do it, the following will be changed

class level
- as-is
  - cpu_common::TensorRegistry (will not be changed)
    - = PortableTensorRegistryTemplate<cpu_common::Tensor>
    - is used by cpu_common::StaticTensorManager and cpu_common::DynamicTensorManager
    - has cpu_common::Tensor as managed and IPortableTensor as external
  - cpu_common::StaticTensorManager
    - build const/non_const tensors as cpu_common::Tensor
    - allocate const/non_const tensors
  - cpu_common::DynamicTensorManager (will not be changed)
    - allocate dynamic tensors
  - cpu_common::Tensor
    - io tensor or const tensor
    - dynamic or not dynamic
  - IConstantInitializer
  - Data
to-be
- cpu::StaticTensorManager <- cpu_common::StaticTensorManager
  - build non_const tensors as cpu_common::Tensor
  - build const tensors as cpu::ExternalTensor
  - allocate the only non_const tensors
- cpu_common::Tensor
  - the only io tensor
  - dynamic or not dynamic
- cpu::ExternalTensor
  - -> cpu_common::Tensor
  - the only const tensor
  - not dynamic tensor
  - the `external` means external of backend so that I make it to take Data not the only {External|MMaped}Data
- IConstantInitializer
  - append `registerExternalInitializer` to enable ExternalTensor to get data from external
- Data
  - append `shareData` for reference count

YongseopKim:
cc/ @chunseoklee @periannath @dhdh-oh and @Samsung/nnfw 

If interested in, anyone can review https://github.com/Samsung/ONE/pull/3037 with the above comment for the explanation.

[...]

PR #3037, opened 2020-07-10
[onert] Fix const tensors to use ExternalData on cpu

YongseopKim:
- ExternalTensor: get Data from external backend
- IConstantInitializer::registerExternalInitializer: set tensor to use a
data from external
- StaticTensorManager's const mem mgr does nothing

ONE-DCO-1.0-Signed-off-by: Yongseop Kim <yons.kim@samsung.com>

PR #3460, opened 2020-07-27
[onert] Fix StaticTensorManager to use the only NativeTensors on cpu

YongseopKim:
Fix cpu::StaticTensorManager to use the only NativeTensors

### DIFF
--- a/runtime/onert/backend/cpu/StaticTensorManager.cc
+++ b/runtime/onert/backend/cpu/StaticTensorManager.cc
@@ -61,12 +61,12 @@ void StaticTensorManager::buildTensor(const ir::OperandIndex &ind,
   if (as_const)
   {
     auto tensor = std::make_shared<ExternalTensor>(tensor_info, backend_layout);
-    _tensors->setMigrantTensor(ind, tensor);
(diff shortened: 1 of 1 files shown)


YongseopKim:
Fix https://github.com/Samsung/ONE/pull/3229#issuecomment-664098155

YongseopKim:
for #3039
draft #3037

YongseopKim:
PTAL @wateret @chunseoklee and @periannath

runtime/onert/backend/cpu/StaticTensorManager.cc
```diff
@@ -97,8 +97,6 @@ void StaticTensorManager::iterate(const std::function<void(const ir::OperandInde
 {
   for (const auto &it : _tensors->managed_tensors())
```
periannath, line 98: This line is not part of this PR, but it seems that this line should be `_tensors->native_tensors()` since `MigrantTensor` becomes `NativeTensor`.

wateret: I'll do that, thanks.

PR #3462, opened 2020-07-27
[onert] Unify LSTM kernelgen of ACL backends

wateret:
Unify LSTM kernel generation function to avoid code duplication.


wateret:
This is for getting better score on SAM.

Related : #3454

PR #3466, opened 2020-07-27
[luci-interpreter]Add DepthToSpace kernel on luci-interpreter

struss:
This commit add `DepthToSpace` kernel on luci-interpreter.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/DepthToSpace.cpp
+++ b/compiler/luci-interpreter/src/kernels/DepthToSpace.cpp
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "DepthToSpace.h"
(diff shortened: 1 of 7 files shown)


compiler/luci-interpreter/src/kernels/DepthToSpace.cpp
```diff
@@ -0,0 +1,90 @@
...
+  }
+  const int block_size = params().block_size;
+  const int32_t input_height = input()->shape().dim(1);
+  const int32_t input_width = input()->shape().dim(2);
+  const int32_t input_channels = input()->shape().dim(3);
+  int32_t output_height = input_height * block_size;
+  int32_t output_width = input_width * block_size;
+  int32_t output_channels = input_channels / block_size / block_size;
+
+  assert(input_height == output_height / block_size);
```
binarman, line 55: nit pick: when this could happen?

struss:
> I am confused between title and committed file.
> SpaceToDepth? DepthToSpace?

Oh, Sorry. That's my mistake. I will fix it. I Add kernel which name is `DepthToSpace`

compiler/luci-interpreter/src/kernels/DepthToSpace.cpp
struss: I thought those cases will not happen. but, in TFlite case, it also check so I added.

seanshpark:
@intom , PTAL

compiler/luci-interpreter/src/kernels/DepthToSpace.test.cpp
```diff
@@ -0,0 +1,60 @@
...
+  params.block_size = 2;
+
+  DepthToSpace kernel = DepthToSpace(&input_tensor, &output_tensor, params);
+  kernel.configure();
+  kernel.execute();
+
+  EXPECT_THAT(extractTensorData<TypeParam>(output_tensor),
+              ::testing::ElementsAreArray(output_data));
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray(output_shape));
+}
```
seanshpark, line 56: Please add negative test cases

PR #3467, opened 2020-07-27
[onert] Add custom(stateless_random_uniform) operation.

YiHyunJin:
This pr is draft for stateless_random_uniform.


runtime/onert/core/include/ir/operation/StatelessRandomUniform.h
```diff
@@ -0,0 +1,52 @@
...
+namespace operation
+{
+
+class StatelessRandomUniform : public Operation
+{
+public:
+  enum Input
+  {
+    INPUT = 0,
+    SEED = 1
```
glistening, line 37: @YiHyunJin Please check what would be tflite's input and property of `stateless_uniform` after reading [tf.random.stateless_uniform](https://www.tensorflow.org/api_docs/python/tf/random/stateless_uniform). You seem to read the input from `tf`.

- Our input could be `seed` and `min`, `max`:
- `seed` could be as input tensor as you wrote.
- I cannot find `min` and `max`, which are passed by tf attributes.
  - Note the data type of `min` and `max` could be `float32` or other types.
- Current `INPUT` is the shape of `input` in `tf` world. We may use this information to determine output tensor's shape.

YiHyunJin: okay i will check.

glistening: @YiHyunJin As you informed me this morning, the default value for `min` and `max` is `0` and `1`. We may provide the simplest implementation without `min` and `max` since our current model requires (min, max) = (very small positive number, 1).

glistening:
@hseok-oh Could you please take a look ? It is a custom operator. @YiHyunJin has confirmed the target model works with this PR.

PR #3472, opened 2020-07-27
[nnpkgrun] Refine Args impl

wateret:
Use `notifier` of `boost::program_options` to reduce branches in `Parse`
method rather than `if` statement for each argument.

This is for better score on CyclomaticComplexity.


wateret:
Just waiting for the SAM results before landing.

PR #3480, opened 2020-07-28
[luci] Add to protect overflow about asymmetric quantization

meejeong:
### [luci] Add to protect overflow about asymmetric quantization
- It can be occured to overflow when you computed fixed bias value.
```
int32 fixed_bias = float_bias / (input scale * weight scale(too small));
```

### DIFF
--- a/compiler/luci/pass/src/QuantizationUtils.cpp
+++ b/compiler/luci/pass/src/QuantizationUtils.cpp
@@ -62,8 +62,8 @@ void compute_asym_scale_zp(float min, float max, float &scaling_factor, int64_t 
   const int32_t kMaxScale = 255;
   const double qmin_double = kMinScale;
   const double qmax_double = kMaxScale;
-  const double rmin = std::fmin(0, min);
-  const double rmax = std::fmax(0, max);
+  double rmin = std::fmin(0, min);
+  double rmax = std::fmax(0, max);
 
   double scale = (rmax - rmin) / (qmax_double - qmin_double);
   double zero_point_double = 0;
@@ -83,6 +83,7 @@ void compute_asym_scale_zp(float min, float max, float &scaling_factor, int64_t 
     assert(min >= 0 && max >= 0);
     nudged_zero_point = kMinScale;
     scale = max / (qmax_double - qmin_double);
+    rmin = 0;
     if (min > 0 && max > 0)
       WARN(l) << "The minimum and maximum values are all positive." << std::endl;
   }
@@ -97,6 +98,13 @@ void compute_asym_scale_zp(float min, float max, float &scaling_factor, int64_t 
   {
     assert(min < 0 && max >= 0);
     nudged_zero_point = static_cast<uint8_t>(std::round(zero_point_double));
+  }
+
+  // protect scale from being very low due to overflow
+  if (scale < 1e-5)
+  {
+    scale = 1e-5;
+    nudged_zero_point = static_cast<uint8_t>(std::round(qmin_double - rmin / scale));
   }
 
   nudged_min = static_cast<float>((qmin_double - nudged_zero_point) * scale);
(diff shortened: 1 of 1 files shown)


seanshpark:
@cgbahk , can you please take a look of this change? I think your expertise is needed :)

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -83,6 +83,7 @@ void compute_asym_scale_zp(float min, float max, float &scaling_factor, int64_t
     assert(min >= 0 && max >= 0);
     nudged_zero_point = kMinScale;
     scale = max / (qmax_double - qmin_double);
+    rmin = 0;
```
jinevening, line 86: If min >= 0 (according to the assert in line 83), `rmin` would be already 0 due to line 65. Do we have to set rmin again?

If rmin is not changed, we can keep `const` in line 65~66.

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -99,6 +100,13 @@ void compute_asym_scale_zp(float min, float max, float &scaling_factor, int64_t
     nudged_zero_point = static_cast<uint8_t>(std::round(zero_point_double));
   }
 
+  // protect scale from being very low due to overflow
+  if (scale < 1e-5)
```
jinevening, line 104: Just curiosity. How did you get the number (1e-5) ? Is it selected empirically?

meejeong: yes. When we tested by calculating the scale factor range that does not overflow, it was empirically **1e-5**.

compiler/luci/pass/src/QuantizationUtils.cpp
meejeong: oh, I didn't catch it. Thank you.

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -99,6 +99,13 @@ void compute_asym_scale_zp(float min, float max, float &scaling_factor, int64_t
...
+  {
+    scale = 1e-5;
+    nudged_zero_point = static_cast<uint8_t>(std::round(qmin_double - rmin / scale));
```
cgbahk, line 106: (optional)

I had nasty bug before due to undefined behavior of `static_cast<uint*_t>(float_value)` when `float_value` is out of scope of destination type.

c++'s implicit conversion rule for float->integral:
> If the value cannot fit into the destination type, the behavior is undefined (even when the destination type is unsigned, modulo arithmetic does not apply).

_(from [cppreference.com](https://en.cppreference.com/w/cpp/language/implicit_conversion#Floating.E2.80.93integral_conversions))_

I strongly recommend to make a safer wrapper like:
```c++
uint8_t my_cast(float f)
{
  assert(std::numeric_limits<uint8_t>::min() <= f);
  assert(f <= std::numeric_limits<uint8_t>::max());
  return static_cast<uint8_t>(f);
}
```

seanshpark: I think @cgbahk 's comment is worth applying.

seanshpark:
@meejeong , are you going to update this PR?

meejeong:
> @meejeong , are you going to update this PR?

@seanshpark , I think that @cgbahk 's comment is needed. So, I want to apply it. Which one is better to amend it in this PR, or upload another PR for applying @cgbahk ' comment?

seanshpark:
> Which one is better to amend it in this PR, or upload another PR

Actually it was optional and reviewers approved this PR, you can send another PR if you like :)

meejeong:
> > Which one is better to amend it in this PR, or upload another PR
> 
> Actually it was optional and reviewers approved this PR, you can send another PR if you like :)

I will upload another PR for applying @cgbahk 's comment

PR #3483, opened 2020-07-28
[onert] Fix tensor registration

wateret:
Revise `runTensorRegistration` to register only operands that the
backend defines.

### DIFF
--- a/runtime/onert/core/src/compiler/ExecutorFactory.cc
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.cc
@@ -133,6 +133,7 @@ void ExecutorFactory::initializeBackendContext(ir::LoweredGraph *lowered_graph)
 }
 
 void ExecutorFactory::runTensorRegistration(ir::LoweredGraph *lowered_graph,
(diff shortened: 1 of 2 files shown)


runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -149,32 +150,17 @@ void ExecutorFactory::runTensorRegistration(ir::LoweredGraph *lowered_graph,
...
-          {
-            const auto &operand_lower_info =
-                lowered_graph->getLowerInfo(index)->def_factors().getOnlyElement();
-
-            // E.g., permute (CPU) -> tensor A -> MaxPool2D(acl_cl)
-            // op.getOutputs() of permute (CPU) returns tensor A
-            // but tensor A belongs to the backend of acl_cl.
-            // So, we have to make this tensor NOT registered for CPU.
-            if (operand_lower_info.backend() != backend)
-              continue;
```
wateret, line 167: Now we are now based on `BackendContext::operand_list`, this case does not occur.

wateret:
Turns out not necessary

Issue #3355, opened 2020-07-21
Compiler FE: Separate the virtual environment creation part from create-package


mhs4670go:
There are codes that creates virtual environment in preset which invoked by `create-package` command.

But, sometimes we don't need a virtual environment. 

So it would be better to separate these part:)

mhs4670go:
Here is the part related virtual environment
```shell
function preset_install()
{
 ...
  # Create python virtual enviornment 
  python3 -m venv "${NNAS_INSTALL_PREFIX}/bin/venv"

  # Install tensorflow
  source "${NNAS_INSTALL_PREFIX}/bin/venv/bin/activate"
  python -m pip --default-timeout=1000 --trusted-host pypi.org --trusted-host files.pythonhost.org \
    install -U pip setuptools
  python -m pip --default-timeout=1000 --trusted-host pypi.org --trusted-host files.pythonhost.org \
    install tensorflow-cpu==2.3.0rc0
}
```

Actually, I think we can **remove** above part because we've introduced `one-cmds` which includes `one-prepare-venv`.

AFAIK, `daily-publish-package` is the only CI test that uses `create-package` more precisely, `infra/scripts/docker_build_nncc.sh`. So, there would be added just `one-prepare-venv`.

PR #3486, opened 2020-07-28
[scripts] create python virtual env

mhs4670go:
This commit adds python virtual env creation step to docker_build_nncc.

We originally install Python with `./nnas create-package` command, but I split the process in two for convenience. There is no need to change CI script.

Related : #3355 


hseok-oh:
@nnfw-bot test nncc-debug
@nnfw-bot test nncc-release

PR #3492, opened 2020-07-28
[luci-interpreter] Add CIRCLECONST CircleOpCode case

mhs4670go:
This commit adds CONSTANT CircleOpCode case.

`CONST` will be deprecated soon.
Related : #3390 

### DIFF
--- a/compiler/luci-interpreter/src/loader/GraphLoader.cpp
+++ b/compiler/luci-interpreter/src/loader/GraphLoader.cpp
@@ -70,6 +70,7 @@ bool isExecutableNode(const luci::CircleNode *node)
   {
     // These nodes denote inputs / outputs of a graph.
     case luci::CircleOpcode::CONST:
+    case luci::CircleOpcode::CIRCLECONST:
     case luci::CircleOpcode::CIRCLEINPUT:
     case luci::CircleOpcode::CIRCLEOUTPUT:
     // The following nodes denote outputs of multiple-output nodes.


s-barannikov:
@mhs4670go What is the difference from CONST? Is it deprecated?

mhs4670go:
@s-barannikov Please refer to [this comment](https://github.com/Samsung/ONE/issues/3390#issuecomment-662316157). In windows build, compiler sees CONST as `const`. I don't know why this happens..

s-barannikov:
Windows headers define `CONST` as `const`:
https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types

mhs4670go:
@s-barannikov Thank you for your information! 

One weird thing is that unit test of this don't give any compile errors. `WinDef.h` header might be included in nncc build somehow.

seanshpark:
I'll wait for a test something like this
```
#ifdef CONST
#undef CONST
#endif // CONST
```

I'm not sure if this is also needed...
```
#ifdef _WIN32
...
#endif // _WIN32
```

s-barannikov:
BTW There are some inconsistency for now:
- `CIRCLEINPUT` and `CircleInput` node
- `CIRCLECONST` and `CircleConst` node
- `CONST` and `CircleConst` node

May be name it as `CIRCLECONST`?

seanshpark:
> May be name it as CIRCLECONST?

`CIRCLE~~~` nodes  are Circle virtual nodes but `CONST` is real (?) node.

s-barannikov:
@seanshpark Not sure what are the virtual nodes. Is it the nodes which do some computation? If so, there is no much difference to me between input and const -- one has dynamic data (provided by the user) and the other has static data (stored in the model). Or is it something other?

seanshpark:
@s-barannikov , `virtual nodes` word I used are nodes that are not in the circle operators but are defined for convenience in _luci_.
These are listed in end of `compiler/luci/lang/include/luci/IR/CircleNodes.lst` file.

seanshpark:
@mhs4670go , although Windows build may work with `#undef CONST`, I think we may need to rename `CONST` as `CIRCLECONST`. If you wish to do the changes, please go on.

mhs4670go:
@s-barannikov @jinevening PTAL:)

PR #3499, opened 2020-07-28
[circle-tensordump] Support dumping of quantized_dimension

jinevening:
This supports dumping of quantized_dimension


mhs4670go:
It would be better to support scalar type in `write_data_to_hdf5`. I'll revise this later.

Issue #3458, opened 2020-07-26
[luci-value-tests] failed items for existing kernels


seanshpark:
These items failed where I could see kernel files in `compiler/luci-interpreter/src/kernels` folder

- ArgMax_000
- ArgMax_001
- ArgMax_002
- ArgMax_003
- ELU_000
- FullyConnected_002 : no bias recipe
- If_000
- If_001
- LeakyRelu_000
- LocalResponseNormalization_000
- SpaceToDepth_000
- Split_000
- Unpack_000
- Unpack_001
- Unpack_002

For quantized
- ArgMax_U8_000
- ArgMax_U8_001
- ArgMax_U8_002
- ArgMax_U8_003
- DepthwiseConv2D_U8_001

Kernel not supported yet
- FullyConnected_U8_000
- Softmax_U8_000

seanshpark:
Related #3457

struss:
Currently, FullyConnected_U8 and Softmax_U8 is not supported yet.

jinevening:
> LeakyRelu_000 :
output value not match. TFlite mplementation change in 1.15

> ELU, IF :
Tensorflow v1.13 version not supported. ELU added in 1.14, If add in 1.15

#3561 may help to resolve the above issues.


> DepthwiseConv2D_U8_001 :
Channel wise quantize example. current TFlite Error ValueError: QuantizationParam has 4 scale values (only 1 is supported).

FYI, upgrading TF to 2.3.0-rc0 can resolve the above issue, but the test will still fail because luci-interpreter currently does not support per-channel quantized DepthwiseConv2D kernel.

struss:
> > LeakyRelu_000 :
> > output value not match. TFlite mplementation change in 1.15
> 
> > ELU, IF :
> > Tensorflow v1.13 version not supported. ELU added in 1.14, If add in 1.15
> 
> #3561 may help to resolve the above issues.
> 

Everything works okay but, Mean kernel have an error around 1.
This comes from `TFlite::round`, `TFlite::round`, `TFlite::round` and `std::round`,`std::min`,`std::max`.(only difference in code)

struss:
> 
> > DepthwiseConv2D_U8_001 :
> > Channel wise quantize example. current TFlite Error ValueError: QuantizationParam has 4 scale values (only 1 is supported).
> 
[...]

PR #3513, opened 2020-07-29
[luci-value-test]Update luci-value-test list on SpaceToDepth.

struss:
For : #3458 

This commit enable `SpaceToDepth` on luci-value-test.


compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -890,7 +890,7 @@ void OperationExporter::visit(luci::CircleSpaceToDepth *node)
 {
   export_simple(node, circle::BuiltinOperator_SPACE_TO_DEPTH,
                 circle::BuiltinOptions_SpaceToDepthOptions,
-                CreateSpaceToDepthOptions(builder).Union());
+                CreateSpaceToDepthOptions(builder, node->block_size()).Union());
```
seanshpark, line 893: maybe unintended change?

struss: Sorry for make changing in two module in on PR!

I will make a PR for Export first and change luci value test later.

struss:
I will make a PR for Export first and change luci value test later.

PR #3507, opened 2020-07-29
DRAFT Compiler FE Fix TCM issues

seanshpark:
on-going draft to fix TCM issues

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #3506, opened 2020-07-28
Compiler FE: Fix TCM no assertion issues


seanshpark:
Refer comment https://github.com/Samsung/ONE/issues/2756#issuecomment-665014662

seanshpark:
- [ ] compiler/locop/src/FormattedTensorShape.test.cpp | 25
- [ ] compiler/luci/lang/src/Nodes/CircleIf.test.cpp | 46
- [ ] compiler/luci/lang/src/Nodes/CircleIf.test.cpp | 41
- [ ] compiler/luci-interpreter/src/kernels/L2Normalize.test.cpp | 48
- [ ] compiler/record-minmax/tests/RecordFunction.test.cpp | 66
- [ ] compiler/record-minmax/tests/RecordFunction.test.cpp | 80
- [ ] compiler/record-minmax/tests/RecordFunction.test.cpp | 37
- [ ] compiler/record-minmax/tests/RecordFunction.test.cpp | 52
- [ ] compiler/record-minmax/tests/RecordFunction.test.cpp | 29
- [ ] compiler/luci-interpreter/src/kernels/Logistic.test.cpp | 52
- [ ] compiler/luci/lang/src/Module.test.cpp | 21
- [ ] compiler/luci-interpreter/src/kernels/LeakyRelu.test.cpp | 66
- [ ] compiler/luci-interpreter/src/kernels/LeakyRelu.test.cpp | 51
- [ ] compiler/locomotiv/src/Node/BiasEncode.test.cpp | 95
- [ ] ompiler/locomotiv/src/Node/BiasEncode.test.cpp | 93
- [ ] compiler/luci/lang/src/Nodes/CircleWhile.test.cpp | 41
- [ ] compiler/luci/lang/src/Nodes/CircleWhile.test.cpp | 46
- [ ] compiler/locomotiv/src/Node/MatMul.test.cpp | 159
- [ ] compiler/locomotiv/src/Node/MatMul.test.cpp | 121
- [ ] compiler/luci/import/src/Importer.test.cpp | 23
- [ ] compiler/luci-interpreter/src/kernels/TransposeConv.test.cpp | 93
- [ ] compiler/luci-interpreter/src/kernels/TransposeConv.test.cpp | 87
- [ ] compiler/luci-interpreter/src/kernels/TransposeConv.test.cpp | 73
- [ ] compiler/luci-interpreter/src/kernels/TransposeConv.test.cpp | 61
[...]

PR #3521, opened 2020-07-29
[hermes] fix test for TCM

seanshpark:
This will revise unit test to do testing

### DIFF
--- a/compiler/hermes/src/hermes.test.cpp
+++ b/compiler/hermes/src/hermes.test.cpp
@@ -18,6 +18,9 @@
 
 #include <gtest/gtest.h>
 
+namespace
+{
+
 class Logger final : public hermes::Source
 {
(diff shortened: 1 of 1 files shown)


seanshpark:
For #3506, from draft #3507

compiler/hermes/src/hermes.test.cpp
```diff
@@ -18,7 +18,23 @@
...
 #include <gtest/gtest.h>
 
-TEST(HermesTest, simple_usecase)
+class Logger final : public hermes::Source
 {
-  // TO BE FILLED
+public:
+  Logger(hermes::Context *ctx);
+  ~Logger();
+};
```
parjong, line 26: (Optional) It would be better to warp this class with anonymous namespace

PR #3525, opened 2020-07-29
[locop] Fix for TCM

seanshpark:
This will revise unit tests for TCM


seanshpark:
For #3506, from draft #3507

PR #3539, opened 2020-07-30
[record-minmax] Remove CONST CircleOpcode

mhs4670go:
This commit removes CONST CircleOpcode.

CONST CircleOpcode has been changed to CIRCLECONST.

Related : #3390 


jinevening:
One question.

What happens if a model has a constant tensor? Did you change luci/import or luci/export to handle that? 

If import/export were not updated yet, this change will disable `record-minmax` to handle contant tensors.

mhs4670go:
@jinevening Please refer to `compiler/luci/lang/include/luci/IR/CircleNodes.lst`.
```
CIRCLE_NODE(CONST, void)
...
CIRCLE_NODE(CIRCLECONST, luci::CircleConst)
```
As you can see, `CONST` already means dummy.

PR #3542, opened 2020-07-30
[onert] Make EdgeConsistencyChecker verbose

wateret:
Make EdgeConsistencyChecker print error message.

### DIFF
--- a/runtime/onert/core/src/ir/verifier/Verifier.cc
+++ b/runtime/onert/core/src/ir/verifier/Verifier.cc
@@ -79,13 +79,47 @@ bool EdgeConsistencyChecker::verify(const Graph &graph) const
   operations.iterate([&](const OperationIndex &index, const Operation &node) {
     for (auto operand_index : node.getInputs() | ir::Remove::UNDEFINED)
     {
-      auto &operand = graph.operands().at(operand_index);
-      mismatches += (operand.getUses().contains(index) ? 0 : 1);
+      try
+      {
+        auto &operand = graph.operands().at(operand_index);
+        bool operand_has_use = operand.getUses().contains(index);
+        if (!operand_has_use)
+        {
+          VERBOSE(EdgeConsistencyChecker) << "[ERROR] EDGE MISMATCH : Missing USE edge - Operand "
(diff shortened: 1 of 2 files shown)


runtime/onert/core/src/ir/verifier/Verifier.cc
```diff
@@ -79,13 +79,47 @@ bool EdgeConsistencyChecker::verify(const Graph &graph) const
...
+                                          << std::endl;
+          mismatches += 1;
+        }
+      }
+      catch (const std::out_of_range &e)
+      {
+        VERBOSE(EdgeConsistencyChecker)
+            << "[ERROR] OPEARAND NOT FOUND : Operation " << index << " has Operand "
+            << operand_index << ", but the operand object is not present in the graph" << std::endl;
+        mismatches += 1;
```
ragmani, line 99: Why don't you add throwing an exception?
I think this looks like an exception, but not a mismatch.

wateret: Verifiers are just designed to be debug mode. And I would like verifiers to never throw. Rather than that, it can just return `false` and the handling is the caller's responsibility.

If an operation has an edge to an operand but there is no such operand, that means the graph is invalid. And you can think of `mismatch` is just a number of graph inconsistency check count.

ragmani: > Verifiers are just designed to be debug mode. And I would like verifiers to never throw. Rather than that, it can just return false and the handling is the caller's responsibility.

If so, how about adding `noexcept` specifier to this function?

> If an operation has an edge to an operand but there is no such operand, that means the graph is invalid. And you can think of mismatch is just a number of graph inconsistency check count

I see. I don't think that graph is invalid is one of the mismatches, but I agree it is one of the graph inconsistency.

wateret: OK then, I will

- Add `noexcept` specifier
- Rename `mismatch` variable to `errors`
- Print total errors (My new idea)

Thank you very much!

PR #3547, opened 2020-07-30
[onert] Unify FC kernelgen of ACL backends

wateret:
Unify Pool2D kernel generation function to avoid code
duplication.

The unified function is used by 3 pooling op.

- Min
- Max
- L2

### DIFF
--- a/runtime/onert/backend/acl_common/AclKernelGen.h
+++ b/runtime/onert/backend/acl_common/AclKernelGen.h
@@ -21,6 +21,7 @@
 #include <ir/Operands.h>
 
 #include <ir/operation/LSTM.h>
+#include <arm_compute/runtime/CL/CLFunctions.h>
 
 namespace onert
 {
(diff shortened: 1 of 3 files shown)


wateret:
This is for SAM score improvement.

runtime/onert/backend/acl_common/AclKernelGen.h
```diff
@@ -262,6 +263,53 @@ kernelGenFullyConnected(const ir::operation::FullyConnected &node, const ir::Ope
...
+  VERBOSE(Pool2DParam) << "STRIDE_W: " << stride.horizontal << std::endl;
+  VERBOSE(Pool2DParam) << "PAD(T): " << padding.top << std::endl;
+  VERBOSE(Pool2DParam) << "PAD(B): " << padding.bottom << std::endl;
+  VERBOSE(Pool2DParam) << "PAD(L): " << padding.left << std::endl;
+  VERBOSE(Pool2DParam) << "PAD(R): " << padding.right << std::endl;
+
+  auto ofm_tensor = tensor_builder->at(ofm_index).get();
+  auto ifm_tensor = tensor_builder->at(ifm_index).get();
+
+  bool exclude_padding = (pooling_type == ::arm_compute::PoolingType::AVG);
```
wateret, line 300: I don't know why, but just to keep the same semantics with the original code. (exclude padding is true only for AVG)

ragmani: `PoolingLayerInfo`'s `exclude_padding` means output tensor's shape does not have padding. It only have value as its dimension value. 
Our spec of pool operation, however, output tensor's shape have padding. Please always set it as `true`.

wateret: Got it, thanks!

PR #3548, opened 2020-07-30
[onert/test] Set config and cache directory for run_test.sh

hseok-oh:
Set config and cache directory to support different config and cache directory for some models
It will help to merge with nnpackage gtest model install

---

part of #3514


tests/scripts/framework/run_test.sh
```diff
@@ -98,6 +100,12 @@ do
         --run=*)
             RUN_TEST=${i#*=}
             ;;
+        --configdir=*)
+            TEST_ROOT_PATH=${i#*=}
```
ragmani, line 104: Is the config directory the same as `TEST_ROOT_PATH`?

hseok-oh: >Is the config directory the same as `TEST_ROOT_PATH`?

Yes.

PR #3550, opened 2020-07-30
[onert] Implement cpu backend OP(Relu6)

YiHyunJin:
Add Relu6 operation


ragmani:
TFLite provides optimized kernel for supporting Relu. How about introducing Relu kernel from TFLite's optimized kernel into cker?

YiHyunJin:
@ragmani 
I have checked.
However, there is no optimization code for Relu6.

ragmani:
@YiHyunJin 
Okay. I just check it. It is strange :confused:

ragmani:
Could you update `tests/nnapi/nnapi_gtest.skip.x86_64-linux.cpu` too?

Issue #1495, opened 2020-05-26
[oneapi] Introduce more `NNFW_STATUS` values (error codes)


wateret:
Currently only two error codes are available.

```c
typedef enum {
  /** Successful */
  NNFW_STATUS_NO_ERROR = 0,
  /** Failed */
  NNFW_STATUS_ERROR = 1,
} NNFW_STATUS;
```

Let's create some more error codes that are more helpful to users.

wateret:
Will work on in in July.

wateret:
We may refer to `ResultCode` in https://android.googlesource.com/platform/frameworks/ml/+/refs/heads/master/nn/runtime/include/NeuralNetworks.h

wateret:
If one introduces an error code, all these must be done.
Let's say we are defining `NNFW_OUT_OF_MEMORY`.

- Introduce a new enum value `NNFW_OUT_OF_MEMORY` in `NNFW_STATUS`
- Update doxygen comment in `nnfw.h` for all API functions that can return `NNFW_OUT_OF_MEMORY`
- Update nnfwapi_gtest corresponding test cases `NNFW_STATUS_ERROR` → `NNFW_OUT_OF_MEMORY`

cc: @KimDongEon

PR #3554, opened 2020-07-30
[nnfwapi] Introduce NNFW_STATUS_OUT_OF_MEMORY

wateret:
Introduce `NNFW_STATUS_OUT_OF_MEMORY` to `NNFW_STATUS`.

Other situations that should return `NNFW_STATUS_OUT_OF_MEMORY` will be
handled later.


wateret:
Part of #1495

runtime/onert/api/src/nnfw_api.cc
```diff
@@ -58,8 +59,9 @@ NNFW_STATUS nnfw_create_session(nnfw_session **session)
 {
   NNFW_RETURN_ERROR_IF_NULL(session);
 
-  *session = new nnfw_session();
-
+  *session = new (std::nothrow) nnfw_session();
+  if (*session == nullptr)
```
ragmani, line 63: It would be better to add check nullptr of `session` too.

PR #3436, opened 2020-07-24
[Draft] Manage ruy_context on runtime

periannath:
- Create ExternalContext class which has ruy context
  - CPU backend has ExternalContext
  - KernelGenerator and FullyConnectedLayer has ExternalContext
  - Pass ruy_context to cker::FullyConnectedHybrid
- Link ruy library to onert

ONE-DCO-1.0-Signed-off-by: JiHwan Yeo <jihwan.yeo@samsung.com>

---

For issue : #3318

Issue #3318, opened 2020-07-21
[onert] Move ruy context from cker to runtime


periannath:
Related : #2929 #3225

Ruy context has a collection of utilities and it is needed to run kernels in ruy library.
Currently, cker has ruy context and it is defined as `thread_local`. 

### Issue
This implementation runs well on most cases but produces error when one runtime thread is re-created after destruction.

At the first run, the weight value for hybrid fully connected is cached by ruy library and weight value is freed.
In thread deconstruction, cached weight value in ruy context is freed.
In the second run on a re-created thread, the runtime doesn't have weight value and ruy context also doesn't have cached weight value.
ONE Runtime will produce segfault for this case.

### Todo
To solve this problem, I'll move ruy context from cker to runtime.
Each session will have its ruy context and the above problem will be solved.

/cc @chunseoklee @YongseopKim @dhdh-oh

periannath:
### How to implement?
- Introduce `ExternalContext`
- Backend has `ExternalContext`
#### Option 1
- Classes from `Backend` to `cker::FullyConnectedHybrid` has `ExternalContext`
  - `Backend`, `KernelGenerator`, `FullyConnectedLayer` gets `ExternalContext` during its initialization
  - Pass `ExternalContext` at kernel run
- Pros
  - Easy to implement
- Cons
  - Some classes has unnecessary `ExternalContext` as its class member
#### Option 2
- Pass `ExternalContext` as argument
  - `LinearExecutor` -> `FunctionSequence` -> `FullyConnectedLayer`
- Pros
  - `ExternalContext` is maintained only at `Backend`
- Cons
  - Many layers are affected
    - `XXLayer::run()` -> `XXLayer::run(ExternalContext *)`

YongseopKim:
This will touch runtime's execution. cc/ @Samsung/nnfw

periannath:
I realized several things via offline discussing with @ragmani.

Option 2 makes onert_core depends on ruy library since onert_core has `LinearExecutor`.
Each subgraph has `Backend` and `BackendContext` and models with control flow can create multiple subgraphs. In this case, multiple ruy context will be created.

Managing `ExternalContext` inside `BackendContext` would be an option.

ragmani:
And @periannath talked about multiple ruy contexts. A flaw of multiple ruy contexts is that they create multiple thread pools unnecessary. Multiple thread pools creates many alive threads at the same time.

periannath:
[...]

PR #3559, opened 2020-07-30
[onert] Manage ruy_context on runtime

periannath:
- Create `ExternalContext` class which has ruy context
  - CPU `BackendConext` class has `ExternalContext`
  - KernelGenerator and FullyConnectedLayer has shared_ptr of `ExternalContext`
- Link ruy library to libbackend_cpu.so

---

For issue : #3318 
From draft #3436

### DIFF
--- a/runtime/onert/backend/cpu/BackendContext.h
+++ b/runtime/onert/backend/cpu/BackendContext.h
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __ONERT_BACKEND_CPU_BACKEND_CONTEXT_H__
+#define __ONERT_BACKEND_CPU_BACKEND_CONTEXT_H__
+
+#include <backend/BackendContext.h>
+#include "ExternalContext.h"
+
+namespace onert
+{
+namespace backend
+{
+namespace cpu
+{
+
+class BackendContext : public onert::backend::BackendContext
+{
+public:
+  BackendContext(const Backend *backend, const ir::Graph *graph,
+                 std::shared_ptr<ITensorBuilder> tensor_builder = nullptr,
+                 std::shared_ptr<IConstantInitializer> constant_initializer = nullptr,
+                 std::shared_ptr<IKernelGenerator> kernel_gen = nullptr,
+                 std::shared_ptr<ITensorRegister> tensor_register = nullptr,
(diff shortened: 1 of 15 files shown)


runtime/onert/backend/cpu/BackendContext.h
```diff
@@ -0,0 +1,55 @@
...
+                 std::shared_ptr<IKernelGenerator> kernel_gen = nullptr,
+                 std::shared_ptr<ITensorRegister> tensor_register = nullptr,
+                 std::shared_ptr<IOptimizer> optimizer = nullptr)
+      : onert::backend::BackendContext(backend, graph, tensor_builder, constant_initializer,
+                                       kernel_gen, tensor_register, optimizer),
+        _external_context(new ExternalContext)
+  {
+  }
+
+  std::shared_ptr<ExternalContext> external_context() { return _external_context; }
```
ragmani, line 45: Could you add a comment about https://github.com/Samsung/ONE/issues/3318#issuecomment-664100840?

periannath: @ragmani Thanks. I've added NOTE and TODO comments.

runtime/onert/backend/cpu/ExternalContext.h
```diff
@@ -0,0 +1,64 @@
...
+public:
+  ExternalContext() : _ruy_context(new ruy::Context)
+  {
+    SetMaxNumThreads(onert::util::getConfigInt(onert::util::config::RUY_THREADS));
+#ifdef USE_RUY_GEMV
+    _ruy_context->cache_policy = ruy::kCacheLHSOnNarrowMul;
+#endif
+  }
+
+  void SetMaxNumThreads(int max_num_threads)
```
wateret, line 47: Method names must start with small "s"

periannath: @wateret Thanks. I've fixed it.

wateret:
I may not fully understand the whole code, but I was not able to see the reason why `IExternalContext` should be part of `BackendContext` yet. I also looked at #3436 . Could you please explain why?

periannath:
@wateret I just intended to manage `IExternalContext` somewhere inside ONE runtime. At first, I defined it as part of cpu `Backend` and it is moved onto cpu `BackendContext` after offline discussion with @ragmani https://github.com/Samsung/ONE/issues/3318#issuecomment-664081927. 

If you are talking about why runtime have to manage ruy context, defining ruy context as thread_local in cker creates many issues for multi-thread environment. One ruy context managed by one session will solve multi-thread issues.

wateret:
@periannath Thank you for explanation, just one more thing to ask. Where does `BackendContext::external_context()` get called except backend context generation? And where does `IExternalContext::setMaxNumThreads` get called except `ExeternalContext` constructor?

periannath:
@wateret `BackendContext::external_context()` is only called at backend context generation, and `IExternalContext::setMaxNumThreads` is only called at `ExternalContext` constructor.
`IExternalContext::setMaxNumThreads` can be used if user want to change number of threads used by ruy library between inference, but currently we don't have API to support that feature.

wateret:
Would that mean if there is no need to call `setMaxNumThreads` from runtime core, `ExternalContext` may not necessarily be part of `BackendContext`?

periannath:
@wateret As far as one runtime session has one `ExternalContext`, `ExternalContext` can be moved into another class. It is not related to `setMaxNumThreads`.
I used `BackendContext` because it seems proper class for now. Having `ExternalContext` inside runtime core would make `libonert_core.so` depends on ruy library. In current implementation, `libbackend_cpu.so` depends on ruy library.

PR #3561, opened 2020-07-30
[common-artifacts] Add venv for tf 2.3.0-rc0

jinevening:
This adds venv for tf 2.3.0-rc0

### DIFF
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -15,14 +15,23 @@ endif()
 # Create python virtual environment with tensorflow 1.13.2
 set(VIRTUALENV_OVERLAY "${NNCC_OVERLAY_DIR}/venv_1_13_2")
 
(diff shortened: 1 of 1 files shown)


compiler/common-artifacts/CMakeLists.txt
```diff
@@ -15,14 +15,23 @@ endif()
 # Create python virtual environment with tensorflow 1.13.2
 set(VIRTUALENV_OVERLAY "${NNCC_OVERLAY_DIR}/venv_1_13_2")
```
mhs4670go, line 16: How about changing `VIRTUALENV_OVERLAY` to `VIRTUALENV_OVERLAY_TF_1_13_2` as well?

jinevening: I've updated the variable name.

underflow101:
Thanks for adding TF 2.3.0!! :)

PR #3564, opened 2020-07-30
[luci-value-test]Update luci-value-test for more test cases.

struss:
For #3458 

This commit update luci_eval_verifier.py and EvalTester.cpp for accepting more test cases.

1. Add more input DataType case for `If` kernel.
2. Add more output DataType case for `ArgMax` kernel.
3. Update for supporting multiple output.

### DIFF
--- a/compiler/luci-value-test/luci_eval_verifier.py
+++ b/compiler/luci-value-test/luci_eval_verifier.py
@@ -59,7 +59,8 @@ subprocess.run(
 # Compare the results.
 for idx in range(len(interpreter.get_output_details())):
     output_details = interpreter.get_output_details()[idx]
-    output_data = np.fromfile(circle_model + ".output" + str(idx), output_details["dtype"])
+    output_data = np.fromfile(circle_model + ".output" + str(idx),
+                              output_details["dtype"])
     shape_file = open(circle_model + ".output" + str(idx) + ".shape", 'r')
     output_shape = [int(i) for i in shape_file.read().split(',')]
     luci_output_data = np.reshape(output_data, output_shape)
--- a/compiler/luci-value-test/luci_eval_verifier.py
+++ b/compiler/luci-value-test/luci_eval_verifier.py
@@ -35,6 +35,10 @@ for i in range(num_inputs):
         input_data = np.array(
             np.random.randint(0, 256, size=input_details["shape"]),
             input_details["dtype"])
+    elif input_details["dtype"] == np.bool_:
+        input_data = np.array(
+            np.random.choice(a=[True, False], size=input_details["shape"]),
+            input_details["dtype"])
     else:
         raise SystemExit("Unsupported input dtype")
 
@@ -45,9 +49,7 @@ for i in range(num_inputs):
 interpreter.invoke()
 
 # Get reference output data.
-assert len(interpreter.get_output_details()) == 1  # TODO: Support multiple outputs
 output_details = interpreter.get_output_details()[0]
-ref_output_data = interpreter.get_tensor(output_details["index"])
 
 # Execute luci interpreter.
 subprocess.run(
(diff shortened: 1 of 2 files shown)


compiler/luci-value-test/tester/src/EvalTester.cpp
```diff
@@ -141,24 +149,30 @@ int entry(int argc, char **argv)
...
+    auto shape_str = std::to_string(output_node->dim(0).value());
+    if (i != 0)
+    {
+      shape_str = "\n" + shape_str;
+    }
+    for (int i = 1; i < output_node->rank(); i++)
+    {
+      shape_str += ",";
+      shape_str += std::to_string(output_node->dim(i).value());
+    }
```
jinevening, line 173: 1. Why do you use a different format from multiple inputs?

2. This code is a little bit confusing. Please use a different variable name (`i`).

3. It would be better to make a string in the loop and write the string to the file outside the loop. (to reduce # of file IO)

compiler/luci-value-test/tester/src/EvalTester.cpp
```diff
@@ -39,9 +39,17 @@ void readDataFromFile(const std::string &filename, char *data, size_t data_size)
...
-  std::ofstream fs(filename, std::ofstream::binary);
+  std::ofstream fs;
+  if (append)
+  {
+    fs = std::ofstream(filename, std::ofstream::app);
+  }
+  else
+  {
+    fs = std::ofstream(filename, std::ofstream::binary);
+  }
```
jinevening, line 52: (optional) It would be possible to reduce the code as below

```
void writeDataToFile(const std::string &filename, const char *data, size_t data_size, ios_base::openmode mode)
{
  std::ofstream fs = std::ofstream(filename, mode);
```

struss: Okay, I will fix it.

compiler/luci-value-test/tester/src/EvalTester.cpp
struss: 1.
2. that's my mistake. I will fix it.
3. Okay, I will make a Change.

struss: 1. I will change multiple output file format. 
2. that's my mistake. I will fix it.
3. Okay, I will make a Change.

compiler/luci-value-test/tester/src/EvalTester.cpp
```diff
@@ -141,24 +141,25 @@ int entry(int argc, char **argv)
...
-  // (ex: Add.circle.output.shape)
-  // TODO: Use HDF5 file format
-  writeDataToFile(output_file, output_data.data(), output_data.size());
-  auto shape_str = std::to_string(output_node->dim(0).value());
-  for (int i = 1; i < output_node->rank(); i++)
+  for (int i = 0; i < module->graph()->outputs()->size(); i++)
   {
-    shape_str += ",";
-    shape_str += std::to_string(output_node->dim(i).value());
+    const auto *output_node = dynamic_cast<const luci::CircleOutput *>(output_nodes[i]);
```
jinevening, line 146: It would be better to use `loco::must_cast` instead of dynamic_cast.

compiler/luci-value-test/tester/src/EvalTester.cpp
```diff
@@ -141,24 +141,25 @@ int entry(int argc, char **argv)
...
+    std::vector<char> output_data(getTensorSize(output_node));
+    interpreter.readOutputTensor(output_node, output_data.data(), output_data.size());
+
+    // Output data is written in ${output_file}
+    // (ex: Add.circle.output)
+    // Output shape is written in ${output_file}.shape
+    // (ex: Add.circle.output.shape)
+    // TODO: Use HDF5 file format
```
jinevening, line 154: Can you rewrite the comment?

struss: Okay, I will fix it!

compiler/luci-value-test/tester/src/EvalTester.cpp
struss: Okay! and also I fix input node casting part as a `loco::must_cast`!

compiler/luci-value-test/luci_eval_verifier.py
```diff
@@ -44,39 +48,63 @@
...
+                                 " does not match with " + circle_model)
+        elif output_details["dtype"] == np.int32:
+            if np.allclose(
+                    luci_output_data,
+                    interpreter.get_tensor(
+                        interpreter.get_output_details()[idx]["index"]),
+                    rtol=0,
+                    atol=0) == False:
+                raise SystemExit("Execution result of " + tflite_model +
+                                 " does not match with " + circle_model)
```
mhs4670go, line 103: The codes look similar except `float32` case. There are plans to add some algorithm for each data?

struss: I think may be not.

Issue #3362, opened 2020-07-21
Compiler FE: Enable PadV2 Op


a-zh:
- [x] res/TensorFlowPythonExamples #3685 
- [x] tflchef #3684 
- [x] tflite2circle #3683
- [x] tfldump #3753
- [x] circledump #3681
- [x] res/TensorFlowLiteRecipes #3676 
- [x] luci.IR #3682 
- [x] luci.Import #3680 
- [x] luci.Export #3679 
- [x] luci.logex #3678 
- [x] luci.Service #3677 
- [x] luci.tests  #3675 
- [x] common artifacts #3752

Parent Issue : #1606, For #277
Draft: https://github.com/Samsung/ONE/pull/3566

a-zh:
`PadV2` example python code:
```
#!/usr/bin/python3

import tensorflow as tf
import numpy as np
# Construct a basic model.
root = tf.train.Checkpoint()

root.f = tf.function(lambda x, y: tf.raw_ops.PadV2(input=x, paddings=y, constant_values=1))

# Create the concrete function.
x = np.ones(shape=[1, 1, 1, 1], dtype=np.float32)
paddings = np.array([[1,1], [2,2], [3,3], [4,4]], dtype=np.int32)
concrete_func = root.f.get_concrete_function(x, paddings)
converter = tf.lite.TFLiteConverter.from_concrete_functions([concrete_func])
tflite_model = converter.convert()
with open("model.tflite", "wb") as f:
    f.write(tflite_model)
 ```

`TensorFlow 2.0.0-rc1` binary tf graph:  [pad_v2.zip](https://github.com/Samsung/ONE/files/4959174/pad_v2.zip)

`Netron` graph:
![pad_v2](https://user-images.githubusercontent.com/39411904/88160049-d13cf000-cc16-11ea-98a0-76b402fe59a7.png)

**Comment by a-zh** [comment]
PAD_V2 circledump output:

```
Dump: pad_v2.circle

===================================================================
Model version: 0
 # sub graphs: 1

Operator Codes: [order] OpCodeName (OpCode Enum)
[0] PADV2 (code: 60, version: 1)

Buffers: B(index) (length) values, if any
B(0) (0)
B(1) (4) 00 00 0x80 0x3f
B(2) (0)
B(3) (0)
B(4) (0)
B(5) (5) 0x31 0x2e 0x39 0x2e 0x30 ...

-------------------------------------------------------------------
Sub-Graph: #0 (noname)

Data Format:
[...]

PR #3566, opened 2020-07-30
[Draft] Compiler FE: Enable PadV2 Op

a-zh:
Issue: https://github.com/Samsung/ONE/issues/3362


compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -717,6 +718,12 @@ void OperationExporter::visit(luci::CirclePad *node)
                 CreatePadOptions(builder).Union());
 }
 
+void OperationExporter::visit(luci::CirclePadV2 *node)
+{
+  export_simple(node, circle::BuiltinOperator_PADV2, circle::BuiltinOptions_PadV2Options,
+                CreatePadOptions(builder).Union());
```
binarman, line 724:
```suggestion
                CreatePadV2Options(builder).Union());
```

compiler/luci/import/src/Nodes/CirclePadV2.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace luci
+{
+
+bool CirclePadV2GraphBuilder::validate(const ValidateArgs &args) const
+{
+  if (args.op.inputs.size() != 3)
+    return false;
+
+  // TODO do attribute checks
```
binarman, line 31: No attributes to check, this todo is redundant
Instead you can check number of outputs

compiler/tflchef/tflite/src/Op/PadV2.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "Convert.h"
+
+namespace tflchef
+{
+
+void TFliteOpPadV2::filler(const tflite::Operator *op, TFliteImport *import,
+                           tflchef::ModelRecipe *model_recipe) const
+{
+  // filler for second input
+  const auto &inputs = *op->inputs();
```
binarman, line 28: Use `fill_tensor_to_import(1, import);` defined in `FillerHelper.h`
for example see `SegmentSum.cpp`

This implementations has some flaws: for example if input is not constant this code could fail with segfault.

a-zh:
@nnfw-bot test format-checker

compiler/luci/export/src/CircleOperationExporter.cpp
a-zh: Done

compiler/luci/import/src/Nodes/CirclePadV2.cpp
a-zh: Done

compiler/tflchef/tflite/src/Op/PadV2.cpp
a-zh: Done

binarman:
Everything is merged, closing

PR #3501, opened 2020-07-28
Draft: [dalgona] Python interface for dynamic analysis of DNN

jinevening:
On going draft to introduce dalgona, a dynamic analysis tool for DNN

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to https://github.sec.samsung.net/STAR/nnfw/issues/11537

PR #3570, opened 2020-07-31
[dalgona] Introduce python interface to analyze DNN

jinevening:
This introduces dalgona, python interface to analyze DNN during inference

---
Draft PR: #3501


seanshpark:
This PR has huge code and this makes hard to review.
It would be nice if you could split the codes so that reviewers can focus and understand better with the codes.

PR #3576, opened 2020-07-31
[luci] Change conditions to calculate scale for all positive & negative valeus

meejeong:
- I used ```zero_point_double``` value to distinguish to calculate scale factor algorithm. However, it had some problems when min/max values is too close to zero values. So, I changed conditions. (It is my big mistake T.T)

- I apply to @cgbahk 's comment from here.

ONE-DCO-1.0-Signed-off-by: mia park mia.park@samsung.com

### DIFF
--- a/compiler/luci/pass/src/QuantizationUtils.cpp
+++ b/compiler/luci/pass/src/QuantizationUtils.cpp
@@ -24,6 +24,13 @@
 namespace luci
 {
 
+uint8_t fp32_to_uint8_cast(float f)
+{
+  assert(std::numeric_limits<uint8_t>::min() <= f);
+  assert(f <= std::numeric_limits<uint8_t>::max());
+  return static_cast<uint8_t>(f);
+}
+
 void compute_sym_scale_zp(float min, float max, float &scaling_factor, int64_t &zp,
                           float &nudged_min, float &nudged_max)
 {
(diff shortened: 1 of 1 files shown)


jinevening:
Can you change the commit message?

For example,
Title: `[luci] Change conditions to calculate scale`
Contents : `This changes conditions to calculate scale`

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -78,15 +85,15 @@ void compute_asym_scale_zp(float min, float max, float &scaling_factor, int64_t
   }
   else
     zero_point_double = qmin_double - rmin / scale;
-  if (zero_point_double <= qmin_double)
+  if (min >= 0)
   {
     assert(min >= 0 && max >= 0);
     nudged_zero_point = kMinScale;
     scale = max / (qmax_double - qmin_double);
```
jinevening, line 92: Is this line necessary?

if min >= 0, it seems that the scale is the same as the value set in line 75.

`double scale = (rmax - rmin) / (qmax_double - qmin_double);`

jinevening: I think line 100 can be removed as well.

PR #3579, opened 2020-07-31
emit_rusage only on debug mode

chunseoklee:
`emit_rusage` causes severe overhead on measurement. Let's enable this on debug build

### DIFF
--- a/runtime/onert/core/src/util/EventCollector.cc
+++ b/runtime/onert/core/src/util/EventCollector.cc
@@ -54,6 +54,7 @@ private:
   std::string _ts;
 };
 
+#ifdef DEBUG
 inline void emit_rusage(EventRecorder *rec, const std::string &ts)
 {
   struct rusage ru;
@@ -81,6 +82,7 @@ inline void emit_rusage(EventRecorder *rec, const std::string &ts)
     rec->emit(evt);
   }
 }
+#endif
 
 } // namespace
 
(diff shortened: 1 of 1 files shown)


runtime/onert/core/src/util/EventCollector.cc
```diff
@@ -99,6 +99,9 @@ void EventCollector::onEvent(const Event &event)
       break;
   }
 
-  // Trace resource usage per each event notification
+// TODO: Add resurece measurement(e.g. RSS)
+// when ready with low overhead in release build
+#ifdef DEBUG
```
hseok-oh, line 104: It will make unused function warning on clang release build.

chunseoklee: inline does not make it work on release mode ? I will test on android build

hseok-oh: @chunseoklee I've checked and got build error

```
/home/nfs/git/ONE/runtime/onert/core/src/util/EventCollector.cc:57:13: error: unused function 'emit_rusage'
      [-Werror,-Wunused-function]
inline void emit_rusage(EventRecorder *rec, const std::string &ts)
            ^
1 error generated.
runtime/onert/core/CMakeFiles/onert_core.dir/build.make:2519: recipe for target 'runtime/onert/core/CMakeFiles/onert_core.dir/src/util/EventCollector.cc.o' failed
make[3]: *** [runtime/onert/core/CMakeFiles/onert_core.dir/src/util/EventCollector.cc.o] Error 1
```

chunseoklee: @hseok-oh I have added commit which declare emit_rusage on debug build.

PR #3596, opened 2020-08-02
[tf2tfliteV2] Import Keras

seanshpark:
This will enable tf2tfliteV2 to import Keras files

### DIFF
--- a/compiler/tf2tfliteV2/tf2tfliteV2.py
+++ b/compiler/tf2tfliteV2/tf2tfliteV2.py
@@ -130,9 +130,17 @@ def _v1_convert(flags):
         ]
         input_shapes = dict(list(zip(input_arrays, input_shapes_list)))
 
-    converter = tf.compat.v1.lite.TFLiteConverter.from_frozen_graph(
-        flags.input_path, _parse_array(flags.input_arrays),
-        _parse_array(flags.output_arrays), input_shapes)
+    retry_with_keras = False
+    try:
+        converter = tf.compat.v1.lite.TFLiteConverter.from_frozen_graph(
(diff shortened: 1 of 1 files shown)


seanshpark:
@mhs4670go , PTAL and any comments are welcome :)

mhs4670go:
@seanshpark I think we should split the argument options according to the input model format. Because in above change, I think it causes keras model converting api error even if it is any graph def related error, which makes it hard to get original exception. And splitting the arg looks more convenient to the user.

How about this change?
```python
# Input model format
    model_format_arg = parser.add_mutually_exclusive_group(required=True)
    model_format_arg.add_argument("--graph_def", action="store_true", help="Use graph def file")
    model_format_arg.add_argument("--saved_model", action="store_true", help="Use saved model")
    model_format_arg.add_argument("--keras_model", action="store_true", help="Use keras model")
```

seanshpark:
Adding additional arguments for model type is OK but may break existing codes that uses this.
I want to understand your proposal: what are the releations with existing `--v1` and `--v2` ?

seanshpark:
@mhs4670go , If you can provide a change that doesn't break existing command lines and provide a way to import tf-keras models, it would be great :) I will close this PR for that.

mhs4670go:
@seanshpark 
> I want to understand your proposal: what are the releations with existing --v1 and --v2 ?


Since both `--v1` and `--v2` options will support the above model type, model type and version options will be independent.

> If you can provide a change that doesn't break existing command lines and provide a way to import tf-keras models, it would be great :) I will close this PR for that.

Okay:) I'll post the PRs.

seanshpark:
Close as @mhs4670go has landed better codes

Issue #3497, opened 2020-07-28
[res] Add U8 recipe: RESIZE_BILINEAR


underflow101:
Parent Issue: #1880

Let's enable U8 recipe for ResizeBilinear_000 Op.

underflow101:
I'm keep getting error below:

`circle2circle: /home/bear/data/opensource/ONE/compiler/luci/service/src/CircleTypeInferenceRule.cpp:521: virtual loco::DataType {anonymous}::TypeInferenceAlgorithm::visit(const luci::CircleOutput*): Assertion `output_dtype == loco::dtype_get(node->from())' failed.`

This error occurs even if I omit `min` & `max` for `quant`.

Meanwhile, I'll try another recipes. (Seems like other things work well.)

seanshpark:
Please copy-paste recipe here or post a PR for this recipe so that I could also investigate the problem you have, if you like.

underflow101:
@seanshpark , my basic recipe is like this:

`operand {
  name: "ifm1"
  type: UINT8
  shape { dim: 1 dim: 4 dim: 4 dim: 3 }
  quant { min: 0 max: 255 scale: 1.0 zero_point: 0 }
}
operand {
  name: "size"
  type: INT32
  shape { dim: 2 }
  filler {
    tag: "constant" arg: "16" arg: "16"
  }
  quant { min: 0 max: 255 scale: 1.0 zero_point: 0}
}
operand {
  name: "ofm"
  type: UINT8
  shape { dim: 1 dim: 16 dim: 16 dim: 3 }
  quant { min: 0 max: 255 scale: 1.0 zero_point: 0 }
}
operation {
  type: "ResizeBilinear"
  input: "ifm1"
  input: "size"
  output: "ofm"
  resize_bilinear_options {
    align_corners: false
    half_pixel_centers: false
  }
}
input: "ifm1"
output: "ofm"`

Resize_Bilinear basically has restriction like this:
> restriction: Input and outputs must all have same scale/zero_point

What I don't get to understand is the error message I get.
(`Assertion output_dtype == loco::dtype_get(node->from())' failed.`)

ifm1 & ofm data type is still the same UINT8, but it seems like I'm having an issue with that.

seanshpark:
There are currently two issues; (1) need to fix luci type inference (2) interpreter needs implementation
For (1), I'll post a fix, (2) need to add to `common-artifacts` exclude list; I'll post a change for this too.

seanshpark:
[...]

PR #3599, opened 2020-08-03
[luci] Fix type inference for ResizeBilinear

seanshpark:
This will fix type infererence for ResizeBilinear Op


seanshpark:
For https://github.com/Samsung/ONE/issues/3497#issuecomment-667747984

PR #3600, opened 2020-08-03
[common-artifacts] Exclude ResizeBilinear_U8

seanshpark:
This will exclude ResizeBilinear_U8_000 recipe as luci-interpreter needs implementation


seanshpark:
https://github.com/Samsung/ONE/issues/3497#issuecomment-667747984

Issue #3603, opened 2020-08-03
Compiler FE: use TensorFlow 2.3.0 official


seanshpark:
As TensorFlow 2.3.0 official version is released, let's use this instead of 2.3.0rc0.
- https://github.com/tensorflow/tensorflow/releases/tag/v2.3.0

seanshpark:
done

PR #3606, opened 2020-08-03
DRAFT Compiler FE Use TF2.3.0

seanshpark:
Use TensorFlow 2.3.0 official version for Compiler frontend


seanshpark:
For #3603

seanshpark:
all landed

PR #3607, opened 2020-08-03
[res] TFlite schema 2.3.0

seanshpark:
This will introduce TFlite schema file version 2.3.0


seanshpark:
For #3603, from draft #3606

Issue #3597, opened 2020-08-03
[onert] introduce sparsity (loader, fullyconnected)


glistening:
I will start to use `sparsity` field of `tflite`.

TensorFlow Lite schema supports optimized operator for spare tensor using `sparsity` field of `tensor`, which is already our part of both `tflite` and `circle`.

```js
table Tensor {
 
  // Parameters to encode a sparse tensor. See the example in
  // tensorflow/lite/testdata/sparse_tensor.json.
  sparsity:SparsityParameters;  // Optional.
 
}

// Parameters to encode a sparse TfLite tensor.
table SparsityParameters {
  // The traversal order of the dimensions defined in the `shape` field of the
  // conceptual dense tensor. For a n-dimensional tensors with dims (d0, d1,
  // ..., dn-1),
  //   - if not block sparse, the traversal_order is just a permutation of (d0,
  //     ..., dn-1). For example, a 2-D matrix stored in row-major order would
  //     have traversal_order = (d0, d1).
  //   - if block sparse with a k-dimensional block (0 <= k <= n), the
  //     traversal_order has n + k elements. The first n elements are still a
  //     permutation of (d0, ..., dn-1). The lask k elements are a permutation
  //     of (dn, ..., dn+k-1), defining how to traverse a block internally. For
  //     example, a 2-D matrix with 2-D blocks, both stored in row-major order
  //     would have traversal_order = (d0, d1, d2, d3).
  traversal_order:[int];
  // For an n-dimensional tensor with a k-dimensional block (0 <= k <= n),
  // stores how a block dimension in (dn, ..., dn+k-1) maps to the original
  // tensor dimension in (d0, ..., dn).
  // It's stored in the order of (dn, ..., dn+k-1).
  // If not block-sparse, this field is NULL.
  block_map:[int];
  // In the traversal order defined above, the metadata needed for
  // each dimension to locate the non-zero values in the original dense tensor.
  // The size of the dim_metadata array = the size of the traversal_order array
  // = n + k.
  dim_metadata:[DimensionMetadata];
}

// Variable-typed buffer to store the index metadata for a sparse dimension.
// The widest type is Int32 instead of UInt32 because tensor's shape is a int32
// vector. We don't want the per-dimensional index to overflow that range.
union SparseIndexVector {
  Int32Vector,
  Uint16Vector,
  Uint8Vector
}

table DimensionMetadata {
  // Whether a dimension is dense or sparse.
  format:DimensionType;
[...]

PR #3609, opened 2020-08-03
[draft] [onert] introduce sparsity for sparse weight fully connected

glistening:
Related: #3597 

It bring the code for sparse weight 2d fully connected.

- This PR is based on tensorflow lite's sparsed_op implementation.
- It uses `sparsity` field of `tflite`.
- It is composed of loader, ir, kernel.
  - Please focus on loader and ir part since I will replace kernel implementation to other which shows better performance.

#### To Do
- [x] Remove hard coded ConstantInitializer~ ←sparse weight has smaller size than normal weight tensor. It is okay.
- [x] Remove `w0_size`, which may be removed by using output's dimension.
- [x] Support `bias` (which is **necessary** for my target model).

### DIFF
--- a/runtime/onert/core/include/ir/TypeInfo.h
+++ b/runtime/onert/core/include/ir/TypeInfo.h
@@ -42,6 +42,8 @@ public:
   float scale() const { return _scale; }
   int32_t offset() const { return _offset; }
   bool sparse() const { return _sparse; }
+  const uint16_t *w1_segments() const { return _w1_segments.data(); }
+  const uint16_t *w1_indices() const { return _w1_indices.data(); }
 
 public:
   void type(const DataType type) { _type = type; }
--- a/compute/cker/include/cker/operation/FullyConnected.h
+++ b/compute/cker/include/cker/operation/FullyConnected.h
@@ -208,6 +208,66 @@ inline void FullyConnectedHybrid(const FullyConnectedParams &params, const Shape
   return;
 }
 
+inline void FullyConnectedSparseWeight(const FullyConnectedParams &params, const Shape &input_shape,
+                                       const float *input_data, const Shape &weights_shape,
+                                       const float *weights_data, const Shape &bias_shape,
+                                       const float *bias_data, const Shape &output_shape,
+                                       float *output_data, int w0_size, const uint16_t *w1_segments,
+                                       const uint16_t *w1_indices)
+{
+  UNUSED_RELEASE(params);
+  UNUSED_RELEASE(input_shape);
+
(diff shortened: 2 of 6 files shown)


runtime/onert/core/include/ir/TypeInfo.h
```diff
@@ -32,22 +33,40 @@ class TypeInfo
...
   {
   }
 
 public:
   DataType type() const { return _type; }
   float scale() const { return _scale; }
   int32_t offset() const { return _offset; }
+  bool sparse() const { return _sparse; }
+  int w0_size() const { return _w0_size; }
+  const uint16_t *w1_segments() const { return _w1_segments.data(); }
```
periannath, line 46: In tflite spec, this type can be INT32, UINT16, and UINT8. In my opinion, TypeInfo should support all three sparse types.

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -355,8 +374,67 @@ ir::OperandIndex BaseLoader<LoaderDomain, SpecificLoader>::loadOperand(const Ten
     if (details != nullptr)
       throw std::runtime_error("Custom Quantization is not supported");
   }
+  // Sparsity
+  auto src_sparsity = tensor->sparsity();
+  int w0_dense_size = 0;
+  std::vector<uint16_t> w1_segments;
```
periannath, line 380: This vector is created every time even if operand is not sparse. It might produce some overhead.

wateret:
LGTM. But I just can't imagine how it would look like if a sparse tensor was not const.

runtime/onert/core/include/ir/TypeInfo.h
glistening: @periannath Yes, we can do later if necessary. Note that we don't support all operator and input data type in tflite specification.

runtime/onert/frontend/base_loader/include/base_loader.h
glistening: @periannath I guess it would be almost zero overhead for creating empty vector. Do you have any suggestion?

periannath: @glistening empty `std::vector` takes 24 bytes on x64 linux.

```bash
std::cout << sizeof(std::vector<int>) << std::endl;
```

I agree with you. This overhead seems negligible.

>  Do you have any suggestion?

Maybe read sparsity after creating `type_info`?

glistening: @periannath I've updated. Looks much better. Thanks.

PR #3610, opened 2020-08-03
[infra] Revise comment for farmhash

seanshpark:
Revise farmhash note for Tensorflow version to 2.3.0


seanshpark:
For #3603, from draft #3606

Issue #3601, opened 2020-08-03
[tf2tfliteV2] Support more model format


mhs4670go:
Currently, `tf2tfliteV2` supports only `graph_def` format.

It's time to extend the supported model format(e.g. keras, saved model).

After this, user will need to specify the format he or she wants to convert.

PR #3613, opened 2020-08-03
[tf2tfliteV2] Support more model format

mhs4670go:
This commit enables tf2tfliteV2 to support more model format.

Related : #3601 

### DIFF
--- a/compiler/tf2tfliteV2/tf2tfliteV2.py
+++ b/compiler/tf2tfliteV2/tf2tfliteV2.py
@@ -48,6 +48,27 @@ def _get_parser():
     converter_version.add_argument(
         "--v2", action="store_true", help="Use TensorFlow Lite Converter 2.x")
 
+    # Input model format
+    model_format_arg = parser.add_mutually_exclusive_group()
+    model_format_arg.add_argument(
+        "--graph_def",
+        action="store_const",
+        dest="model_format",
+        const="graph_def",
+        help="Use graph def file")
+    model_format_arg.add_argument(
+        "--saved_model",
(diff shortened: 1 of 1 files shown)


compiler/tf2tfliteV2/tf2tfliteV2.py
```diff
@@ -48,6 +48,15 @@ def _get_parser():
...
         "--v2", action="store_true", help="Use TensorFlow Lite Converter 2.x")
 
+    # Input model format
+    model_format_arg = parser.add_mutually_exclusive_group(required=True)
+    model_format_arg.add_argument(
+        "--graph_def", action="store_true", help="Use graph def file")
+    model_format_arg.add_argument(
+        "--saved_model", action="store_true", help="Use saved model")
+    model_format_arg.add_argument(
+        "--keras_model", action="store_true", help="Use keras model")
```
seanshpark, line 58: all three can be given and that is somewhat strange

compiler/tf2tfliteV2/tf2tfliteV2.py
```diff
@@ -48,6 +48,15 @@ def _get_parser():
     converter_version.add_argument(
         "--v2", action="store_true", help="Use TensorFlow Lite Converter 2.x")
 
+    # Input model format
+    model_format_arg = parser.add_mutually_exclusive_group(required=True)
+    model_format_arg.add_argument(
+        "--graph_def", action="store_true", help="Use graph def file")
```
seanshpark, line 54: can you set this as default so it will work same as given or not given?

PR #3616, opened 2020-08-03
[one-cmds] Use TF2.3.0

seanshpark:
This will revise to TensorFlow 2.3.0 version


seanshpark:
For #3603, from draft #3606

PR #3617, opened 2020-08-03
[TensorFlowPhtyonExample] `WHILE` example throwing incompatible shape

hyunsik-yoon:
This adds `WHILE` example that throws incompatible shape with `tf2tfliteV2`

```
$ python ../../compiler/tf2tfliteV2/tf2tfliteV2.py --v2 -i while_2.pbtxt -o while_2.pbtxt -I Hole:0 -O Output:0

 Exception: venv/tf-2.3/lib/python3.6/site-packages/tensorflow/python/eager/lift_to_graph.py:339:0:
 error: body function result type tensor<1x1xi32> is incompatible with result type tensor<1x0xi32>
 at index 0
 ...
 note: see current operation: %1:2 = "tf.While"(%0, %arg0)
 {body = @_functionalize_body_00, cond = @_functionalize_cond_00, device = "", is_stateless = false, output_shapes = [], parallel_iterations = 10 : i64}
 : (tensor<1x0xi32>, tensor<1x1xi32>) -> (tensor<1x0xi32>, tensor<1x1xi32>)
```


seanshpark:
I think it is OK to leave existing `while` as is.

hyunsik-yoon:
Fixed. Thanks.

PR #3620, opened 2020-08-03
[common-artifacts] Use TensorFlow 2.3.0

seanshpark:
This will revise to use TensorFlow version 2.3.0


seanshpark:
For #3603, from draft #3606

PR #3634, opened 2020-08-04
[luci-value-test]Scalar output write update.

struss:
For #3458
This commit enable Scalar output write on shape file.
Scalar shape output write as a 1.

### DIFF
--- a/compiler/luci-value-test/tester/src/EvalTester.cpp
+++ b/compiler/luci-value-test/tester/src/EvalTester.cpp
@@ -154,7 +154,7 @@ int entry(int argc, char **argv)
     writeDataToFile(std::string(output_file) + std::to_string(i), output_data.data(),
                     output_data.size());
     // In case of Tensor output is Scalar value.
-    // undefined output will not appear in output of interpreter.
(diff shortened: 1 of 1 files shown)


compiler/luci-value-test/tester/src/EvalTester.cpp
```diff
@@ -153,14 +153,23 @@ int entry(int argc, char **argv)
     // (ex: Add.circle.output0.shape)
     writeDataToFile(std::string(output_file) + std::to_string(i), output_data.data(),
                     output_data.size());
-    auto shape_str = std::to_string(output_node->dim(0).value());
-    for (int j = 1; j < output_node->rank(); j++)
+    // In case of Tensor output is Scalar value.
+    // undefined output will not appear in output of interpreter.
```
jinevening, line 157: (optional) Can you change the comment? 

"undefined output will not appear in output of interpreter." -> "The output tensor with rank 0 is treated as a scalar with shape (1)"

struss: Okay! I think those comment is more clear!

PR #3635, opened 2020-08-04
[onert] Introduce new PermutationEliminationPass

wateret:
- Introduce new PermutationEliminationPass
- Change approach for Permute insertion
    - Insert all Permute operations needed
    - Then remove unnecessary Permute (Check if the tensor is portable)

Part of #8533

### DIFF
--- a/runtime/onert/core/src/ir/LoweredGraph.cc
+++ b/runtime/onert/core/src/ir/LoweredGraph.cc
@@ -23,6 +23,7 @@
 #include "pass/ConstantLoweringPass.h"
 #include "pass/PermutationOperationPass.h"
 #include "pass/PermutationInsertionPass.h"
+#include "pass/PermutationEliminationPass.h"
 #include "ir/GraphIterator.h"
 #include "verifier/Verifier.h"
 #include "backend/Backend.h"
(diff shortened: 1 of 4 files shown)


runtime/onert/core/src/ir/pass/PermutationInsertionPass.cc
```diff
@@ -60,35 +60,8 @@ void PermutationInsertionPass::callback(const OperandIndex &index, Operand &obje
...
-      // Check exceptional cases that Permute ops are not inserted
-      if (factor.layout() == def_factor.layout() &&
-          compatible_backends(factor.backend(), def_factor.backend()))
-      {
-        VERBOSE(PermutationInsertionPass) << "Permutation Insertion is skipped for operand "
-                                          << index << " / as the tensor is compatible with backend "
-                                          << factor.backend()->config()->id() << std::endl;
-        factor_to_index.emplace(factor, index);
-        continue;
-      }
```
wateret, line 90: Let all Permute operations inserted, for the cases of different PermuteFactors between def and use.

runtime/onert/core/src/ir/LoweredGraph.cc
```diff
@@ -123,6 +124,9 @@ LoweredGraph::LoweredGraph(const Graph &graph, const compiler::CompilerOptions &
     pass::PermutationInsertionPass pi_pass(*this);
     pi_pass.run();
 
+    pass::PermutationEliminationPass pe_pass(*this);
+    pe_pass.run();
```
wateret, line 128: All permute operations are inserted from `PermutationInsertionPass`, now it is time to remove some.

PR #3637, opened 2020-08-04
[Draft/luci-interpreter]Add Optional input tensor case handle.

struss:
Parent Issue #3458 on fully connected 002 test case.

This commit is Draft for optional Tensor Input handling case update.

### DIFF
--- a/compiler/luci-interpreter/src/loader/KernelBuilder.cpp
+++ b/compiler/luci-interpreter/src/loader/KernelBuilder.cpp
@@ -74,9 +74,12 @@ const Tensor *KernelBuilder::getInputTensor(const loco::Node *node) const
   return tensor;
 }
 
-const Tensor *KernelBuilder::getOptionalInputTensor(const loco::Node *node) const
+const Tensor *KernelBuilder::getOptionalInputTensor(loco::Node *node) const
 {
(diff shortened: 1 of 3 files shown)


compiler/luci-interpreter/src/loader/KernelBuilder.cpp
```diff
@@ -74,9 +74,12 @@ const Tensor *KernelBuilder::getInputTensor(const loco::Node *node) const
   return tensor;
 }
 
-const Tensor *KernelBuilder::getOptionalInputTensor(const loco::Node *node) const
+const Tensor *KernelBuilder::getOptionalInputTensor(loco::Node *node) const
 {
-  // TODO Revise this when optional inputs are implemented in the IR.
+  if (dynamic_cast<luci::CircleOutputExclude *>(node))
```
s-barannikov, line 79: It is not necessary to remove `const` on the parameter if apply the suggested change:
```suggestion
  if (dynamic_cast<const luci::CircleOutputExclude *>(node))
```

struss: oh! Thanks I will make a change and make a official PR for it.

struss:
If everything is fine, then I will make a PR for this!

PR #3645, opened 2020-08-04
[onert] Separate roles of _buffer and _allocator in Tensor.h

hyunsik-yoon:
This separates the roles of `_buffer` and `_allocator` in _Tensor.h_.

Previously,
- memory pointer can be accessed sometimes by `_buffer` and sometimes by `_allocator.base()`, which was hard to understand.

But now,
- memory pointer can be accessed _only_ by `_buffer`
- `_allocator` exists only to have shared _ownership_ of memory area

### DIFF
--- a/runtime/onert/core/include/backend/cpu_common/Tensor.h
+++ b/runtime/onert/core/include/backend/cpu_common/Tensor.h
@@ -45,19 +45,32 @@ public:
 
 public:
   // Only one of two method 'setBuffer' must be called once
+
+  /**
+   * @brief Set the Buffer object. This method is called for static and non-const tensor
+   */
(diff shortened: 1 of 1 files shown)


wateret:
Just to double-check, this means that the allocator never re-allocate a buffer, doesn't it?

hyunsik-yoon:
It can be re-allocated by calling `overwriteBuffer(..)` (caller needs to dealloc before re-alloc by calling `overwriteBuffer`)

wateret:
Got it, thanks. IMHO this commit does not make the code look easier nor harder.

(Beyond the commit) "Allocator" should have been named "Allocation", I think.

PR #3533, opened 2020-07-29
[onert] Introduce new PermutationEliminationPass

wateret:
Introduce new PermutationEliminationPass
Introduce interface `ITensor::dynamic_tensor_manager()`

ONE-DCO-1.0-Signed-off-by: Hanjoung Lee <hanjoung.lee@samsung.com>

PR #3647, opened 2020-08-04
[onert] Eliminate copy for model outputs

wateret:
Eliminate copies for model outputs(only for cpu backend).

### DIFF
--- a/runtime/onert/core/src/ir/pass/PermutationEliminationPass.cc
+++ b/runtime/onert/core/src/ir/pass/PermutationEliminationPass.cc
@@ -46,54 +46,107 @@ void PermutationEliminationPass::visit(const operation::Permute &node)
     auto in_backend_id = in_def_factor.backend()->config()->id();
     auto out_backend_id = out_def_factor.backend()->config()->id();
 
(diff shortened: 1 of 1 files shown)


wateret:
Draft : #3533 (Second commit)

hyunsik-yoon:
It took me long time to figure out why we need the line 91 to line 105. T.T

IMHO, if we provide more comment especially for _pass_, it would be helpful for readers and future developers.
From the experience of compiler, comments such as the following helped a lot:
Some examples are..
- https://github.com/Samsung/ONE/blob/master/compiler/moco-tf/src/Canonicalization/AvgPoolCanonicalizer.cpp#L36-L51
- https://github.com/Samsung/ONE/blob/master/compiler/exo/src/Pass/MergeConcatNodesPass.cpp#L50-L83
- https://github.com/Samsung/ONE/blob/master/compiler/logo/src/Passes/ReorderDecodePass.cpp#L101-L161

wateret:
@hyunsik-yoon Sure, I will leave some comments with ascii figures, soon enough.

Plus, we need some refactoring to reduce efforts for keeping integrity for OpSequence, Operation and Operand.

Issue #3649, opened 2020-08-04
[nnfwapi] nnfw header file management & policy


wateret:
There is a [the discussion in our internal repo](https://github.sec.samsung.net/STAR/nnfw/issues/10796), about nnfw headers.

- `nnfw.h` : Official API (public)
- ~~`nnfw_dev.h`~~ → `nnfw_experimental.h` : Experimental API (public)
- ~~`nnfw_debug.h`~~ → `nnfw_internal.h` : API only for internal developers (private)

And there was a conclusion, but I did not do the action 😓 . So I'm going to do it now.

- Rename `nnfw_dev.h` to `nnfw_experimental.h`
- Rename `nnfw_debug.h` to `nnfw_internal.h`

wateret:
Done!

PR #3650, opened 2020-08-04
[nnfwapi] Rename nnfw headers

wateret:
- Rename `nnfw_dev.h` to `nnfw_experimental.h`
- Rename `nnfw_debug.h` to `nnfw_internal.h`

For #3649

### DIFF
--- a/docs/howto/how-to-build-runtime.md
+++ b/docs/howto/how-to-build-runtime.md
@@ -107,7 +107,7 @@ $ tree -L 3 ./Product/out
 │   │   ├── NeuralNetworksEx.h
 │   │   ├── NeuralNetworksExtensions.h
 │   │   ├── NeuralNetworks.h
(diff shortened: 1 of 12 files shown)


chunseoklee:
> I am still not sure it is right that nnfw_set_config and nnfw_get_config are in nnfw_internal.h. How app developer can set config without exposing this APIs?

We need to discussion and provide this kind of feature one by one.

wateret:
> I am still not sure it is right that nnfw_set_config and nnfw_get_config are in nnfw_internal.h. How app developer can set config without exposing this APIs?

They just can't. Probably if something is needed, I will just add a new specific API rather than general ones like those.

glistening:
@wateret Thank you for answer. I agree that let's add new API when we are requested.

wateret:
@glistening Right, that's the current strategy 😄

Issue #3361, opened 2020-07-21
Compiler FE: Enable NonMaxSuppressionV4 Op


ai-moiseev:
- [x] res/TensorFlowPythonExamples #3692 
- [x] tflchef #3694 
- [x] tflite2circle #3695 
- [x] tfldump #3696 
- [x] circledump #3697 
- [x] res/TensorFlowLiteRecipes #3704 
- [x] luci.IR #3698 
- [x] luci.Import #3702 
- [x] luci.Export #3703 
- [x] luci.logex #3700 
- [x] luci.Service #3701 
- [x] luci.tests #3705 
- [x] common artifacts #3699 

For #1606
Draft #3652

binarman:
Everything is done, closing

PR #3652, opened 2020-08-04
DRAFT Compiler FE: Enable NonMaxSuppressionV4 Op

ai-moiseev:
Enable NonMaxSuppressionV4 operator.

For #3361


compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -2019,6 +2025,30 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
...
+      INTERNAL_EXN("CircleNonMaxSuppressionV4 IR is not configured correctly");
+
+    auto index = node->index();
+    if (index == 1)
+      return loco::TensorShape({0});
+
+    assert(index == 0);
+
+    auto unknown = loco::TensorShape{loco::Dimension()};
+    auto max_output_size = dynamic_cast<const luci::CircleConst *>(nmsv4->max_output_size());
```
ai-moiseev, line 2043: I'm not sure if this is a right way to get constant value to determine output shape, so I'm planning to update this code section.

compiler/luci/lang/include/luci/IR/Nodes/CircleNonMaxSuppressionV4.h
```diff
@@ -0,0 +1,63 @@
...
+
+public:
+  bool pad_to_max_output_size(void) const { return _pad_to_max_output_size; }
+  void pad_to_max_output_size(bool pad_to_max_output_size)
+  {
+    _pad_to_max_output_size = pad_to_max_output_size;
+  }
+
+private:
+  bool _pad_to_max_output_size{0};
```
seanshpark, line 58: Please use `{false}` for `bool`

compiler/luci/service/src/CircleShapeInferenceRule.cpp
seanshpark: Yes, this is OK :)

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -2019,6 +2025,30 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
...
+    if (max_output_size == nullptr)
+      return unknown; // we need CircleConst for max output size
+    LUCI_ASSERT(max_output_size->dtype() == S32, "Only support int32 for max_output_size");
+
+    auto max_output_size_value = uint32_t(max_output_size->at<S32>(0));
```
seanshpark, line 2048: To be sure, we need to check element count > 0 or == 1 to access its content.

compiler/luci/service/src/CircleTypeInferenceRule.cpp
```diff
@@ -572,6 +577,13 @@ struct TypeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::DataT
     return then_graph_output->dtype();
   }
 
+  loco::DataType visit(const luci::CircleNonMaxSuppressionV4Out *node) final
+  {
+    (void)node;
+    assert(node->index() == 0 || node->index() == 1);
+    return loco::DataType::S32;
```
seanshpark, line 584: @jinevening , is this OK for quantization?

seanshpark:
@ai-moiseev, Thank you! Code looks good with some minor comments. Please go on with PRs :)

compiler/luci/service/src/CircleTypeInferenceRule.cpp
jinevening: I think it is ok for quantization, because TFLite does not have a quantized kernel for this Op.

I have another question to @ai-moiseev .

TFLite kernel for NonMaxSuppression has three outputs: selected_indices, selected_scores, num_selected_indices. But this draft assumes it has only two outputs. Why did you assume there are only two outputs?

ai-moiseev: @jinevening TFLite kernel has additional output ``selected_scores`` in case when ``is_soft_nms`` is true and the node gets marked as NonMaxSuppressionV5 in tflite model. Otherwise it's saved as NonMaxSuppressionV4 with ``selected_indices`` and ``num_selected_indices`` outputs.

compiler/luci/lang/include/luci/IR/Nodes/CircleNonMaxSuppressionV4.h
ai-moiseev: Done, fixed it :)

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -2019,6 +2025,30 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
...
+    if (index == 1)
+      return loco::TensorShape({0});
+
+    assert(index == 0);
+
+    auto unknown = loco::TensorShape{loco::Dimension()};
+    auto max_output_size = dynamic_cast<const luci::CircleConst *>(nmsv4->max_output_size());
+    if (max_output_size == nullptr || max_output_size->size<S32>() < 1)
+      return unknown; // we need CircleConst for max output size
+    LUCI_ASSERT(max_output_size->dtype() == S32, "Only support int32 for max_output_size");
```
binarman, line 2046: This assert is duplicated in line 2044 (https://github.com/Samsung/ONE/pull/3652/files#diff-aada5b5648e693d6ff6ca4b253f1b255R2044) in `size<S32>` call.
If you want to check type, please move this assert before the condition.

compiler/luci/lang/include/luci/IR/Nodes/CircleNonMaxSuppressionV4.h
```diff
@@ -0,0 +1,63 @@
...
+
+public:
+  bool pad_to_max_output_size(void) const { return _pad_to_max_output_size; }
+  void pad_to_max_output_size(bool pad_to_max_output_size)
+  {
+    _pad_to_max_output_size = pad_to_max_output_size;
+  }
+
+private:
+  bool _pad_to_max_output_size{false};
```
binarman, line 58: Do you use this attribute in some shape inference or something?

compiler/tflchef/tflite/src/Op/NonMaxSuppressionV4.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace tflchef
+{
+
+void TFliteOpNonMaxSuppressionV4::filler(const tflite::Operator *op, TFliteImport *import,
+                                         tflchef::ModelRecipe *model_recipe) const
+{
+  const auto &inputs = *op->inputs();
+
+  // for input max_output_size
+  const tflite::Tensor *tensor = import->tensors()->Get(inputs[2]);
```
binarman, line 30: You can use `fill_tensor_to_import` function from `FillerHelper.h`
I think only difference is it can not check types, like you did with assert.

compiler/luci/lang/include/luci/IR/Nodes/CircleNonMaxSuppressionV4.h
ai-moiseev: Nope, since there's no such option in tflite representation, I guess it's not needed. I've removed it in commit below.

compiler/tflchef/tflite/src/Op/NonMaxSuppressionV4.cpp
ai-moiseev: Good to know about this helper. I've preserved type checks and used this method to simplify filler processing code.

ai-moiseev:
Thank everyone for comments :) I've addressed them in additional draft patches.

compiler/luci/service/src/CircleTypeInferenceRule.cpp
jinevening: @ai-moiseev , I didn't know that. Thanks.

binarman:
Everything is merged, thank you!

Issue #3632, opened 2020-08-04
[cker/ruy] EXPERIMENTAL_RUY_FEATURE flag not working for android


periannath:
`EXPERIMENTAL_RUY_FEATURE` flag is introduced in #3446. 

This flag works well on arm32 linux, but it is not working on arm64 android.

I'll investigate the reason of this error. 

/cc @YongseopKim

periannath:
`dynamic_cast` in `FullyConnectedLayer.cc` produces nullptr for correct casting
https://github.com/Samsung/ONE/blob/60dd70837f2331c8f69555ed7b8dd7e03becbce4/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc#L135
- This `dynamic_cast` works correctly on arm32 linux
- Produces nullptr only at arm64 android

I found some pages explain about `dynamic_cast` in Android
- [NDK document for common problems](https://android.googlesource.com/platform/ndk/+/master/docs/user/common_problems.md#rtti_exceptions-not-working-across-library-boundaries)

  > Problem: Exceptions are not being caught when thrown across shared library boundaries, or dynamic_cast is failing.
  Solution: Add a key function to your types. A key function is the first non-pure, out-of-line virtual function for a type. For an example, see the discussion on Issue 533.

- [How to add a key function](https://github.com/android/ndk/issues/533#issuecomment-335977747)

### In summary
- `dynamic_cast` is not working across shared library boundaries (NDK only)
  - Declared type of `_weight` is `IPortableTensor`  defined at `libonert_core.so`
  - Actual type of `_weight` is `cpu_common::Tensor` defined at `libbackend_cpu.so`
- Solution
  - Method 1 : Add a key function to `IPortableTensor` and `cpu_common::Tensor`
    - Problem : Thers's no cpp file for `IPortableTensor`
  - Method 2 : Using `static_cast` instead of `dynamic_cast`
    - Problem : Type check for polymorphism is missing
    - Maybe use `nnfw::misc::polymorphic_downcast` instead of `static_cast`?

periannath:
Method 2 runs well on android. I've uploaded PR #3654.

periannath:
Fixed by #3654.

PR #3654, opened 2020-08-05
[cker/ruy] Use polymorphic_downcast to fix dynamic_cast error on android

periannath:
- `dynamic_cast` is not working across shared library boundaries (Android NDK limitation)
- `EXPERIMENTAL_RUY_FEATURE` flag uses `dynamic_cast` and it is not working on android
- Use polymorphic_downcast instead of dynamic_cast to fix this issue

---

For issue : #3632

### DIFF
--- a/runtime/libs/misc/include/misc/polymorphic_downcast.h
+++ b/runtime/libs/misc/include/misc/polymorphic_downcast.h
@@ -27,7 +27,9 @@ namespace misc
 
 template <typename DstType, typename SrcType> inline DstType polymorphic_downcast(SrcType *x)
 {
+#ifndef __ANDROID__
   assert(dynamic_cast<DstType>(x) == x);
+#endif
   return static_cast<DstType>(x);
 }
 
--- a/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
+++ b/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
@@ -19,6 +19,7 @@
 #include "../Tensor.h"
 #include <cker/operation/FullyConnected.h>
 #include <cker/TensorUtils.h>
+#include <misc/polymorphic_downcast.h>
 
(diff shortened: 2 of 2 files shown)


hseok-oh:
@periannath 

1. `dynamic_cast` is using many place. Should we fix them all?
2. If `dynamic_cast` is not working on android build, we should fix `nnfw::misc::polymorphic_downcast` because it will not work on debug build. `dynamic_cast` is used in `nnfw::misc::polymorphic_downcast` on assert.

periannath:
@hseok-oh Not all `dynamic_cast` have to be replaced. 

Two conditions should be met to cause this issue.
1. Classes defined across shared library boundaries 
1. Classes don't have a key function 

We can fix `nnfw::misc::polymorphic_downcast` or use `static_cast` instead of `nnfw::misc::polymorphic_downcast`.

hseok-oh:

periannath:
@YongseopKim and I will find general solution to fix `nnfw::misc::polymorphic_downcast`.

hseok-oh:
Then we need build flag for `libs/misc` to ignore `assert` check on android build...

https://github.com/Samsung/ONE/blob/c17a51bb620df2c040757891ea188b08ccc3cdf5/runtime/libs/misc/include/misc/polymorphic_downcast.h#L28-L38

glistening:
For early history about `polymorphic_downcast` against `dynamic_cast`, see [1 year ago comments](https://github.sec.samsung.net/STAR/nnfw/pull/5383#discussion_r200073).

periannath:
@hseok-oh @YongseopKim I've add `__ANDROID__` macro check in `polymorphic_downcast.h` to avoid assert in android. Also, removing redundant check code in `FullyConnectedLayer.cc` since static_cast does not return nullptr. PTAL.

PR #3661, opened 2020-08-05
[common-artifacts] Add fixed_seed option to TestDataGenerator

mhs4670go:
This commit adds fixed_seed option to TestDataGenerator.

Sometimes we want it to generate same inputs to compare two network with
same inputs for testing optimizing pass or sth.


compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -62,10 +63,12 @@ template <typename T> void geneate_random_data(std::mt19937 &gen, void *data, ui
   }
 }
 
-void fill_random_data(void *data, uint32_t size, loco::DataType dtype)
+void fill_random_data(void *data, uint32_t size, loco::DataType dtype, uint32_t seed = 0)
```
seanshpark, line 66: we usually didn't use default values for methods. can you introduce new `fill_random_data` with 4 parameters and then let `fill_random_data` with 3 call with 4 arguments for intermediate state and then gradually remove with 3 ?

mhs4670go:
@seanshpark merge plz:)

PR #3666, opened 2020-08-05
[luci] Use common artifacts

mhs4670go:
This commit makes luci-tests use common artifacts.

### DIFF
--- a/compiler/luci/requires.cmake
+++ b/compiler/luci/requires.cmake
@@ -1,3 +1,4 @@
+require("common-artifacts")
 require("foder")
 require("loco")
 require("locop")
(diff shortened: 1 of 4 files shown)


compiler/luci/requires.cmake
```diff
@@ -1,3 +1,4 @@
+require("common-artifacts")
```
seanshpark, line 1: common-artifacts uses luci. we cannot apply this change.

mhs4670go: @seanshpark Oh right. I first thought that since `common-artifacts` uses `luci` only when optimizing(`circle2circle`), this luci-test also uses what `common-artifacts` uses like recipe, `tfl-chef`(recipe -> tflite), `tflite2circle`(tflite -> circle).

But, this PR makes us have to add `common-artifacts` to preset, which looks bad:)

Issue #2929, opened 2020-07-07
[onert] Running tflite_run several times produces segfault


ragmani:
I recently tried to run tflite_run. But it produced a segmentation fault.
Running the model included fullyconnected hybrid by using tflite_run with --num_runs produces a segmentation fault.

- Once
```bash
$ USE_NNAPI=1 BACKENDS="cpu" ./Product/armv7l-linux.release/out/bin/tflite_run --num_runs 1 model_with_fc_hybrid.tflite 
nnapi function 'ANeuralNetworksModel_create' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
nnapi function 'ANeuralNetworksModel_addOperand' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
nnapi function 'ANeuralNetworksModel_setOperandValue' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
nnapi function 'ANeuralNetworksModel_addOperationEx' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
nnapi function 'ANeuralNetworksModel_addOperation' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
nnapi function 'ANeuralNetworksModel_identifyInputsAndOutputs' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
nnapi function 'ANeuralNetworksModel_finish' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
nnapi function 'ANeuralNetworksCompilation_create' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
nnapi function 'ANeuralNetworksCompilation_finish' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
input tensor indices = [5,4,2,3,]
nnapi function 'ANeuralNetworksExecution_create' is loaded from '/home/odroid/nfs/Product/armv7l-linux.release/out/bin/../lib/libneuralnetworks.so' 
[...]

PR #3668, opened 2020-08-05
[onert] Enable constant tensor deallocation if tensor is cached

periannath:
- Constant tensor deallocation is introduced at #2748 but it is disabled since it produces error with multi-thread execution #2929
- Multi-thread issue is now fixed by #3559
- Remove temporal macro `EXPERIMENTAL_RUY_FEATURE` to enable deallocation feature
  - Deallocation is only enabled on ARM architecture
  - Ruy library is not used on x86

### DIFF
--- a/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
+++ b/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
@@ -115,8 +115,6 @@ void FullyConnectedLayer::fullyConnectedHybrid()
       getTensorShape(_output), reinterpret_cast<float *>(_output->buffer()), temp_arena,
       _external_context->ruy_context());
 
-// TODO Remove this ifdef
-#ifdef EXPERIMENTAL_RUY_FEATURE
   if (_cached_weights == nullptr || _is_weights_freed)
     return;
 
@@ -148,7 +146,6 @@ void FullyConnectedLayer::fullyConnectedHybrid()
       _is_weights_freed = true;
     }
   }
-#endif
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
```diff
@@ -198,8 +195,7 @@ void FullyConnectedLayer::prepare()
     }
   }
 
-#ifdef USE_RUY_GEMV
-#ifdef EXPERIMENTAL_RUY_FEATURE
+#if defined(__ARM_NEON__) && defined(USE_RUY_GEMV)
```
YongseopKim, line 198: if Ruy library is not used on x86, is `#if defined(__ARM_NEON__)` unnecessary?

periannath: @YongseopKim Without `#if defined(__ARM_NEON__)`, constant tensor is deallocated even if it is not cached by ruy library on x86.

YongseopKim: Ah, It was meaning that - if`Ruy library is not used on x86`, the `USE_RUY_GEMV` would be not defined so that this code would not built. I guess your meanging is the ruy library is built and used on x86?

YongseopKim: FYI, the reason why I thought so is the word `USE_RUY_GEMV` expresses the availiability of ruy.

periannath: @YongseopKim Ah, I understand your point. On x86 architecture, ruy library is built, and `USE_RUY_GEMV` is defined. Although ruy library is built, it is not used by fullyconneted layer on x86.

YongseopKim: Okay. If so, I think it's necessary to make ruy library not built on x86 regardless of this PR. (If the change would land, of course, this defined macro should be fixed.)

PR #3677, opened 2020-08-05
[luci] Enable PadV2 operator in services

a-zh:
This enables PadV2 support in services

Draft: #3566
Issue: #3362

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.cpp
@@ -1126,9 +1126,14 @@ public:
     const loco::DataType S32 = loco::DataType::S32;
 
     auto input_shape = loco::shape_get(node->input()).as<loco::TensorShape>();
-    auto paddings = loco::must_cast<luci::CircleConst *>(node->paddings());
+    auto paddings = dynamic_cast<luci::CircleConst *>(node->paddings());
 
-    // TODO support non-const case
+    if (!paddings)
+    {
+      auto node_shape = own_shape(node);
+      return loco::NodeShape{node_shape};
+    }
+
     // TODO support other data type
     LUCI_ASSERT(paddings->dtype() == S32, "Only support int 32 for now");
     LUCI_ASSERT(paddings->rank() == 2, "paddings should be rank 2")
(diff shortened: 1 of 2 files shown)


compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -1121,6 +1121,40 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
     return loco::NodeShape{output_shape};
   }
 
+  loco::NodeShape visit(const luci::CirclePadV2 *node) final
+  {
+    const loco::DataType S32 = loco::DataType::S32;
```
binarman, line 1126: nitpick
In this situation it is better to use `constexpr`, because compiler has more guarantees for optimizations.

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -1121,6 +1121,40 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
...
+
+    auto input_shape = loco::shape_get(node->input()).as<loco::TensorShape>();
+    auto paddings = loco::must_cast<luci::CircleConst *>(node->paddings());
+
+    // TODO support non-const case
```
binarman, line 1131: My I think we can just return without computing anything in this case.
We can not do anything anyway.

seanshpark: > without computing anything in this case.

I think this would do
```cxx
loco::NodeShape unknown;
return unknown;
```

seanshpark: Or we could just return shape of the node from file
```
auto this_shape = own_shape(node);
return loco::NodeShape{this_shape};
```

compiler/luci/service/src/CircleShapeInferenceRule.cpp
binarman: I looked at code around, there are no `constexpr`s, so I'll leave this piece as-is.

PR #3683, opened 2020-08-05
[tflite2circle] Enable PadV2 operator

a-zh:
This enables PadV2 operator in tflite2circle

Draft: #3566
Issue: #3362


compiler/tflite2circle/src/BuildBuiltinOptions/PadV2Options.cpp
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "PadV2Options.h"
+
+#include <cassert>
```
ai-moiseev, line 19: Is cassert needed here?
I think there's several files which contain `#include <cassert>`, but never use it, and this line gets copied over :)

a-zh: Done

PR #3684, opened 2020-08-05
[tflchef] Enable PadV2 operator

a-zh:
This enables PadV2 operator in tflchef

Draft: #3566
Issue: #3362


compiler/tflchef/tflite/src/Op/PadV2.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "FillerHelper.h"
+
+namespace tflchef
+{
+
+void TFliteOpPadV2::filler(const tflite::Operator *op, TFliteImport *import,
+                           tflchef::ModelRecipe *model_recipe) const
+{
+  // Filler for indices and shape
+  fill_tensor_to_import(1, import);
```
kvochko, line 28: I think input 2 (`constant_values`) needs a filler too. Could you also fix the comment above this line?

a-zh: Done

PR #3685, opened 2020-08-05
[res] TensorFlowPythonExamples for PadV2

a-zh:
This will introduce TensorFlowPythonExamples for PadV2 op

Draft: #3566
Issue: #3362

### DIFF
--- a/res/TensorFlowPythonExamples/examples/PadV2/__init__.py
+++ b/res/TensorFlowPythonExamples/examples/PadV2/__init__.py
@@ -1,0 +1,7 @@
+import tensorflow as tf
+import numpy as np
+
+input_ = np.ones(shape=[1, 1, 1, 1], dtype=np.float32)
+paddings_ = np.array([[1, 1], [2, 2], [3, 3], [4, 4]], dtype=np.int32)
+constant_values_ = 1
+op_ = tf.raw_ops.PadV2(input=input_, paddings=paddings_, constant_values=constant_values_)
--- a/res/TensorFlowPythonExamples/examples/PadV2/__init__.py
+++ b/res/TensorFlowPythonExamples/examples/PadV2/__init__.py
@@ -2,6 +2,7 @@ import tensorflow as tf
 import numpy as np
 
 input_ = tf.compat.v1.placeholder(shape=[1, 1, 1, 1], dtype=tf.float32)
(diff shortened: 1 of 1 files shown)


res/TensorFlowPythonExamples/examples/PadV2/__init__.py
```diff
@@ -0,0 +1,7 @@
+import tensorflow as tf
+import numpy as np
+
+input_ = np.ones(shape=[1, 1, 1, 1], dtype=np.float32)
+paddings_ = np.array([[1, 1], [2, 2], [3, 3], [4, 4]], dtype=np.int32)
+constant_values_ = 1
+op_ = tf.raw_ops.PadV2(input=input_, paddings=paddings_, constant_values=constant_values_)
```
seanshpark, line 7: With TF 1.13.1, I got this error
```
AttributeError: module 'tensorflow' has no attribute 'raw_ops'
```

a-zh: This is expected behavior for TF less 2.0. Do you want more readable message for test?

seanshpark: Our default TF is 1.13.1. If fails in 1.13.1 we cannot use this example.

a-zh: I'm sorry. I cannot find `PadV2` for `TF 1.13.2`. `PadV2` - a new operator which not available on `TF 1.13.2`. I don't have any ideas for fix it.

seanshpark: Please add a comment at the top like
```
# NOTICE Run this script in TensorFlow 2.x as PadV2 is not available in 1.13.2
```

binarman: @seanshpark 
I checked this operator in tf 1.13 again and found solution, now it works.

seanshpark:
@binarman , thanks!

binarman:
FYI @seanshpark 

@kvochko @AShedko and @a-zh are on vacation for two weeks (will be back 24th August)
We can not get approves from them so I think we should remove/ignore their status in existing PRs

PR #3702, opened 2020-08-06
[luci/import] Add NonMaxSuppressionV4 op

ai-moiseev:
Support NonMaxSuppressionV4 in luci importer.

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
+++ b/compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Import/Nodes/CircleNonMaxSuppressionV4.h"
+
+#include <luci/IR/Nodes/CircleNonMaxSuppressionV4.h>
+#include <luci/IR/Nodes/CircleNonMaxSuppressionV4Out.h>
+
+#include <loco.h>
+#include <oops/UserExn.h>
+
+namespace luci
+{
+
+bool CircleNonMaxSuppressionV4GraphBuilder::validate(const ValidateArgs &args) const
+{
+  const auto &inputs = args.op.inputs;
+  const auto &outputs = args.op.outputs;
+
+  if (inputs.size() != 5)
+    return false;
+  if (outputs.size() != 2)
+    return false;
+
+  const auto &tensors = args.reader.tensors();
+  const auto &boxes_tensor = tensors.at(inputs[0]);
+  if (boxes_tensor->shape.size() != 2)
+    return false;
+  if (boxes_tensor->shape.at(1) != 4)
+    return false;
+  if (boxes_tensor->shape.at(0) != tensors.at(inputs[1])->shape.at(0))
+    return false;
+
+  if (tensors.at(inputs[2])->type != circle::TensorType_INT32)
+    return false;
+  if (tensors.at(inputs[3])->type != circle::TensorType_FLOAT32)
+    return false;
+  if (tensors.at(inputs[4])->type != circle::TensorType_FLOAT32)
+    return false;
+
+  return true;
+}
+
+/**
+ * @brief  NonMaxSuppressionV4 Node builder
+ *
+ * @note   Current loco does not provide multiple outputs
+ *         We will create multiple NonMasSuppressionV4Oout nodes to emulate this
+ */
+
+void CircleNonMaxSuppressionV4GraphBuilder::build(const circle::OperatorT &op,
+                                                  GraphBuilderContext *context) const
+{
+  assert(context != nullptr);
+
+  auto graph = context->graph();
+
+  const std::vector<int32_t> &inputs = op.inputs;
+  const std::vector<int32_t> &outputs = op.outputs;
+  const auto &tensors = context->reader()->tensors();
+  const auto &opcodes = context->reader()->opcodes();
+  auto tensors_ptr = context->reader()->tensors_ptr();
+  assert(tensors_ptr != nullptr);
+
+  std::vector<CircleNode *> input_nodes;
+  for (const int32_t input_tensor_index : inputs)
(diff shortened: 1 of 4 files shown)


compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  // Create CircleNonMaxSuppressionV4
+  auto node = graph->nodes()->create<CircleNonMaxSuppressionV4>();
+  node->boxes(input_nodes[0]);
+  node->scores(input_nodes[1]);
+  node->max_output_size(input_nodes[2]);
+  node->iou_threshold(input_nodes[3]);
+  node->score_threshold(input_nodes[4]);
+
+  assert(outputs.size() == 2);
```
a-zh, line 92: Do you really agree with excluding check of size for release build?

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
```diff
@@ -0,0 +1,123 @@
...
+  assert(context != nullptr);
+
+  auto graph = context->graph();
+
+  const std::vector<int32_t> &inputs = op.inputs;
+  const std::vector<int32_t> &outputs = op.outputs;
+  const auto &tensors = context->reader()->tensors();
+  const auto &opcodes = context->reader()->opcodes();
+  auto tensors_ptr = context->reader()->tensors_ptr();
+  assert(tensors_ptr != nullptr);
```
a-zh, line 76: Do you really agree with excluding check on nullptr for release build?

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
```diff
@@ -0,0 +1,123 @@
...
+ * @brief  NonMaxSuppressionV4 Node builder
+ *
+ * @note   Current loco does not provide multiple outputs
+ *         We will create multiple NonMasSuppressionV4Oout nodes to emulate this
+ */
+
+void CircleNonMaxSuppressionV4GraphBuilder::build(const circle::OperatorT &op,
+                                                  GraphBuilderContext *context) const
+{
+  assert(context != nullptr);
```
a-zh, line 67: Do you really agree with excluding check on nullptr for release build?

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
```diff
@@ -0,0 +1,123 @@
...
+    // Let's use name of output 0 as NonMaxSuppressionV4 name
+    const circle::TensorT &output_tensor = *tensors[outputs[0]];
+    node->name(tensor_name(output_tensor));
+    node->op_version(opcodes[op.opcode_index].get()->version);
+
+    // NOTE We don't set quantization for NonMaxSuppressionV4 itself but to virtual outputs
+  }
+
+  // Create virtual outputs of NonMaxSuppressionV4
+  for (int32_t n = 0; n < int32_t(outputs.size()); ++n)
```
a-zh, line 103: Do you really need signed `int32_t` for `n`? And you really sure with C-casting result?

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+bool CircleNonMaxSuppressionV4GraphBuilder::validate(const ValidateArgs &args) const
+{
+  const auto &inputs = args.op.inputs;
+  const auto &outputs = args.op.outputs;
+
+  if (inputs.size() != 5)
+    return false;
+  if (outputs.size() != 2)
+    return false;
```
a-zh, line 36: For me, integers looks better in named constant form: ONE, TWO...

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  const auto &tensors = args.reader.tensors();
+  const auto &boxes_tensor = tensors.at(inputs[0]);
```
a-zh, line 39: Maybe we can get more semantic info for `input[]` elements. Something like this `auto perfect_name=input[ZERO];`.

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  std::vector<CircleNode *> input_nodes;
+  for (const int32_t input_tensor_index : inputs)
+  {
+    input_nodes.push_back(context->nodefinder()->node(input_tensor_index));
+  }
+
+  // Create CircleNonMaxSuppressionV4
+  auto node = graph->nodes()->create<CircleNonMaxSuppressionV4>();
+  node->boxes(input_nodes[0]);
```
a-zh, line 86: Maybe we can get more semantic info for `input_nodes[]` elements. Something like this `auto input_nodex=input[ZERO];`.

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
ai-moiseev: Every other node import validation implementation uses regular integer literals, so I'm not sure if this is needed here.

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
ai-moiseev: Done, replaced it with size_t.

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
ai-moiseev: I've referenced other implementations of multi-output graph builders (CircleSplit, CircleIf), all of them use assertion checking. I don't have complete understanding of architecture, but I suppose that context should be valid in end-user scenario.

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
ai-moiseev: As in previous case, I haven't found no place in project which uses this kind of notation, so I think it's not needed here.

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
binarman: I agree with Alexander, I think it is better to leave numbers, because this is how rest of importer looks like.

compiler/luci/import/src/Nodes/CircleNonMaxSuppressionV4.cpp
binarman: Context _must_ be passed to this function, otherwise it is considered as invalid usage and a logic error in program. We can not recover after it properly in runtime.
Assert here is fine.

binarman:
Merging without @a-zh approval, because he is on vacation and will be available only 24-th of August

PR #3708, opened 2020-08-07
[circle-tensordump] Check whether option is provided or not

mhs4670go:
This commit add condition that checks whether one option is provided or
not.

circle-tensordump needs one option for how to print but arser doesn't
support required exclusive option yet.

### DIFF
--- a/compiler/circle-tensordump/driver/Driver.cpp
+++ b/compiler/circle-tensordump/driver/Driver.cpp
@@ -49,6 +49,13 @@ int entry(int argc, char **argv)
     return 255;
   }
 
+  if (arser["--tensors_to_hdf5"] == arser["--tensors"])
+  {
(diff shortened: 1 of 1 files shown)


compiler/circle-tensordump/driver/Driver.cpp
```diff
@@ -49,6 +49,13 @@ int entry(int argc, char **argv)
     return 255;
   }
 
+  if (arser["--tensors_to_hdf5"] == arser["--tensors"])
```
seanshpark, line 52: I wonder what `arser["--tensors_to_hdf5"]` and `arser["--tensors"]` return... :)

mhs4670go: @seanshpark That is a boolean value for whether the option has been passed or not.

PR #3712, opened 2020-08-07
[luci] Change the way accessing vectors

mhs4670go:
This commit changes the way accessing vectors.

[] operators allows access beyond its actual size.


mhs4670go:
@seanshpark merge plz:)

Issue #3667, opened 2020-08-05
Compiler FE: Introducing CircleFileExporter to reduce duplicate codes


jinevening:
When saving a luci module to a circle file, we use `CircleExporter`, which requires `CircleExporter::Contract` for its argument. For example,

```
  luci::CircleExporter exporter;
  CircleExpContract contract(module.get(), output_path);     // Implementation of CircleExporter::Contract
  if (!exporter.invoke(&contract))
  {
    std::cerr << "ERROR: Failed to export '" << output_path << "'" << std::endl;
    return 255;
  }
```

The problem is that `CircleExporter::Contract` needs to be implemented in each project, increasing duplicate codes. I found four modules have `CircleExpContract.h` and `CircleExpContract.cpp` which have the same implementation of `CircleExporter::Contract`.

```
hjj@hjj-ubuntu:~/ONE/compiler$ find . -name CircleExpContract*
./luci-value-test/tester/src/CircleExpContract.h
./luci-value-test/tester/src/CircleExpContract.cpp
./record-minmax/src/CircleExpContract.h
./record-minmax/src/CircleExpContract.cpp
./circle-quantizer/include/CircleExpContract.h
./circle-quantizer/src/CircleExpContract.cpp
./circle2circle/include/CircleExpContract.h
./circle2circle/src/CircleExpContract.cpp
```

I'd like to make `CircleFileExporter`, a class that exports a luci module to a circle file without need for implementing `CircleExporter::Contract`, to remove the duplicate codes.

One issue is to determine where to implement `CircleFileExporter` (making a separate module ? or under the `luci` directory?).

seanshpark:
I think method to store to a file with filename parameter and others can be provided in `CircleExporter.h` file as inline function.

jinevening:
> I think method to store to a file with filename parameter and others can be provided in CircleExporter.h file as inline function.

Making a method rather than a separate class would be simpler. I'll make `exportToFile` in `CircleExporter` class.

jinevening:
I close this issue because all related PRs were merged.

PR #3715, opened 2020-08-07
DRAFT: Introducing exportToFile to reduce duplicate codes

jinevening:
This introduces exportToFile to reduce duplicate codes

---
Related to #3667


jinevening:
**For reviewers**:  Please review `CircleExporter.h` and `CircleExporter.cpp`, which have the implementation of `exportToFile`.

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -185,9 +184,7 @@ void RecordMinMax::saveModel(const std::string &output_model_path)
 {
   // Export to output Circle file
   luci::CircleExporter exporter;
-  CircleExpContract contract(_module.get(), output_model_path);
-
-  if (!exporter.invoke(&contract))
+  if (!exporter.exportToFile(_module.get(), output_model_path))
```
seanshpark, line 187: `CircleExpContract ` is introduced to separate and now you binded again.
This is not I hoped.

jinevening: Maybe I misundertood your comment in https://github.com/Samsung/ONE/issues/3667#issuecomment-670206599.

Can you explain a little bit more?

seanshpark: I think the comment is somwhat misleading (I forgot somepart of the design)
`CircleExporter` which generates export data and storing the data was decoupled by `CircleExpContract`.
This was by design.
Misleading part is that comment says storing the data to file can be included in `CircleExporter` module which I didn't think is good thing todo.

What I think is exporting to file by some methould should be independent of `CircleExporter`.
`CircleExpContract` just calls that method.

jinevening: @seanshpark  How about making `CircleFileExpContract.h` under the luci? Then, the final code would look like below.

```
#include <luci/CircleExporter.h>
#include <luci/CircleFileExpContract.h>
...

luci::CircleExporter exporter; 
CircleFileExpContract contract(_module.get(), output_model_path);

  if (!exporter.invoke(&contract))
...
```

seanshpark: > How about making CircleFileExpContract.h under the luci? 

there are several modules in luci. which one or a new one?

jinevening: > there are several modules in luci. which one or a new one?


Maybe under the `luci/export/include/luci`? (where `CircleExporter.h` resides)

seanshpark: OK :) let's try

jinevening:
@seanshpark I've changed the implementation to decouple exporting and saving. PTAL

seanshpark:
Looks good :)

PR #3717, opened 2020-08-07
[oenrt] Fix batch_to_space op

YiHyunJin:
reinterpret_cast<const int32_t *>(_crops); -> reinterpret_cast<const int32_t *>(_crops->buffer());
_crops->buffer() should put in the crops buffer.
Not a pointer to IPortableTensor.

Change data type. int -> int32_t

### DIFF
--- a/compute/cker/include/cker/operation/BatchToSpaceND.h
+++ b/compute/cker/include/cker/operation/BatchToSpaceND.h
@@ -87,8 +87,8 @@ inline void BatchToSpaceND(const Shape &unextended_input1_shape, const T *input1
   const int32_t block_shape_height = block_shape_data[0];
   const int32_t block_shape_width = block_shape_data[1];
 
-  const int crops_top = crops_data[0];
-  const int crops_left = crops_data[2];
+  const int32_t crops_top = crops_data[0];
+  const int32_t crops_left = crops_data[2];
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/ir/operation/BatchToSpaceND.cc
```diff
@@ -31,8 +31,8 @@ void BatchToSpaceND::accept(OperationVisitor &v) const { v.visit(*this); }
 
 BatchToSpaceND::BatchToSpaceND(const OperandIndexSequence &inputs,
                                const OperandIndexSequence &outputs)
-    : Operation{OperandConstraint::createExact(3u), inputs, outputs}
 {
+    : Operation{OperandConstraint::createExact(3u), inputs, outputs}
```
intom, line 35: "3u" mean input size?

YiHyunJin: When doing setInputs, it checks if the number of inputs is correct.

```
void Operation::setInputs(const OperandIndexSequence &indexes)
{
  assert(_input_constr.check(indexes.size()));
  _inputs = indexes;
}
```

hseok-oh:
@intom PTAL

PR #3706, opened 2020-08-06
[nnfwapi] Introduce model-building API

wateret:
- Introduce `nnfw_load_circle_from_buffer` to `nnfw_internal.h`
- Introduce `CircleGen` to nnfw_api test, which is the in-memory circle builder
- Sample tests with `CircleGen`

https://github.com/Samsung/ONE/issues/3503#issuecomment-669049510

PR #3718, opened 2020-08-07
[runtime] Make circle_schema as a separate proj

wateret:
Make circle_schema as a separate CMake target to be reusable from other
modules.


wateret:
Use : #3706

runtime/onert/frontend/circle_schema/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_library(circle_schema INTERFACE)
+
+target_link_libraries(circle_schema INTERFACE flatbuffers::flatbuffers)
+
+target_include_directories(circle_schema INTERFACE include)
```
wateret, line 5: Please also check the directory path.

PR #3721, opened 2020-08-07
[onert/test] Use installed list and model config

hseok-oh:
- Change list and config install path: tests -> test
- Use installed list file: remove source dependency


hseok-oh:
PTAL

PR #3723, opened 2020-08-07
[pota-value-test] Enable quantization value test

jinevening:
This enables quantization value test


jinevening:
I disabled the tests for Conv2D_004 and DepthwiseConv2D_002, because the previous golden data is not valid as the quantization algorithm was changed.

I did not include the new golden data in this PR for easy review. The new golden data will be uploaded in the following PRs.

PR #3728, opened 2020-08-10
[infra/script] Fix bug in docker_build_nncc.sh

hseok-oh:
- Fix tar exclude file option: --except -> --exclude
- Fix option order


seanshpark:
files are mixed...

hseok-oh:
@seanshpark Thanks. I've updated. PTAL

Issue #3727, opened 2020-08-10
About cleaning up `rwx` properties of ONE compiler installed files.


lemmaa:
After installing the ONE compiler, the `rwx` properties of the executable file are varied. Unless there is a specific reason, is it not good to unify in a single form? 755 or 775 would be common.

```
total 4880
drwxrwxr-x 2 sjlee sjlee    4096  8월 10 10:54 .
drwxrwxr-x 6 sjlee sjlee    4096  8월 10 10:54 ..
-rwxr-xr-x 1 sjlee sjlee   89624  8월 10 10:54 circle2circle
-rwxr-xr-x 1 sjlee sjlee   84024  8월 10 10:54 circle-quantizer
-rwxrwxr-x 1 sjlee sjlee    2142  8월 10 10:50 generate_bcq_output_arrays
-rwxr-xr-x 1 sjlee sjlee    1519  8월 10 10:54 model2nnpkg.sh
-rwxrwxr-x 1 sjlee sjlee    1670  8월  4 19:26 one-codegen
-rwxrwxr-x 1 sjlee sjlee    1598  8월  4 19:26 one-import
-rwxrwxr-x 1 sjlee sjlee    4295  8월  4 19:27 one-import-bcq
-rwxrwxr-x 1 sjlee sjlee    3556  8월  4 19:26 one-import-tf
-rwxrwxr-x 1 sjlee sjlee    1871  8월  4 19:26 one-import-tflite
-rwxrwxr-x 1 sjlee sjlee    3652  8월  4 19:26 one-optimize
-rwxrwxr-x 1 sjlee sjlee    1901  8월  4 19:27 one-pack
-rwxrwxr-x 1 sjlee sjlee    1692  8월  4 19:27 one-prepare-venv
-rwxrwxr-x 1 sjlee sjlee    4567  8월  4 19:26 one-quantize
-rwxr-xr-x 1 sjlee sjlee   19720  8월 10 10:51 one-version
-rwxrwxr-x 1 sjlee sjlee    3561  8월 10 10:50 preserve_bcq_info
-rwxr-xr-x 1 sjlee sjlee 3722744  8월 10 10:54 record-minmax
-rwxr-xr-x 1 sjlee sjlee  313880  6월 17 14:07 tf2circle
-rwxr-xr-x 1 sjlee sjlee    2590  8월 10 10:54 tf2nnpkg
-rwxr-xr-x 1 sjlee sjlee   76992 12월  9  2019 tf2tflite
[...]

PR #3731, opened 2020-08-10
[bcq-tools] Install with matching access

seanshpark:
This will revise to install with group read-excute to match others


seanshpark:
For #3727

PR #3735, opened 2020-08-10
[infra] Use user defined value by default

mhs4670go:
This commit enables create-package to get NPROC var from user as well as
from system.

### DIFF
--- a/infra/packaging/build
+++ b/infra/packaging/build
@@ -85,7 +85,7 @@ function join_by
 # Invoke "preset_configure" function that the preset provides
 preset_configure
 
-NPROC=$(cat /proc/cpuinfo | grep -c processor)
+NPROC=${NPROC:-$(cat /proc/cpuinfo | grep -c processor)}
 cmake --build . -- -j$((NPROC/2)) all
 cmake --build . -- install
 # Install NN Package tools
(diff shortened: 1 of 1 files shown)


infra/packaging/build
```diff
@@ -85,7 +85,7 @@ function join_by
 # Invoke "preset_configure" function that the preset provides
 preset_configure
 
-NPROC=$(cat /proc/cpuinfo | grep -c processor)
+NPROC=${NPROC:-$(cat /proc/cpuinfo | grep -c processor)}
```
parjong, line 88: Oh... is this working ;-)? It would be better to add "echo" below to check whether this works as expected.

mhs4670go: [Reference](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion)
```
${parameter:-word}

If parameter is unset or null, the expansion of word is substituted. Otherwise, the value of parameter is substituted.
```
@parjong You mean adding "echo" to real PR or just for test?

parjong: It would be better to keep that change in the actual PR as we did for preset :)

parjong: As a user, I would like to see whether my configuration is applied as expected

PR #3738, opened 2020-08-10
[cker] Fix wrong calculation code in MatrixBandPart

llFreetimell:
When `input_shape` is 2-rank shape, `batch_num` is calculated as 0.
This commit will fix it.

### DIFF
--- a/compute/cker/include/cker/operation/MatrixBandPart.h
+++ b/compute/cker/include/cker/operation/MatrixBandPart.h
@@ -33,7 +33,7 @@ void MatrixBandPart(const T num_lower_diags, const T num_upper_diags, const Shap
   auto last_dim = input_shape.DimensionsCount() - 1;
 
   T batch_num = 1;
-  for (int dim = 0; dim < last_dim - 2; dim++)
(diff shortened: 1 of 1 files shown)


compute/cker/include/cker/operation/MatrixBandPart.h
```diff
@@ -32,10 +32,10 @@ void MatrixBandPart(const T num_lower_diags, const T num_upper_diags, const Shap
 {
   auto last_dim = input_shape.DimensionsCount() - 1;
 
-  T batch_num = 0;
+  T batch_num = 1;
   for (int dim = 0; dim < last_dim - 2; dim++)
```
ragmani, line 36: `last_dim - 2` should also be changed to `input_shape.DimensionsCount() - 2`. If I'm right, could you change it too?

llFreetimell, line 36:
```suggestion
  for (int dim = 0; dim < input_shape.DimensionsCount() - 2; dim++)
```

PR #3739, opened 2020-08-10
[onert/test] Install tflite download model config

hseok-oh:
Install tflite download model config for test


hseok-oh:
CI model test is not working by this bug

PR #3742, opened 2020-08-10
[luci] Fix quantization for transpose conv

jinevening:
This fixes quantization for transpose conv


jinevening:
This is a bug fix. TRANSPOSE_CONV is not correctly quantized without this code.

PR #3752, opened 2020-08-10
[common-artifacts] Disable interpreter tests for PadV2

binarman:
Add PadV2 operator in exclude list, because interlreter does not work with this operator

ONE-DCO-1.0-Signed-off-by: Anton Zhukov <a.zhukov@samsung.com>


binarman:
Issue #3362

PR #3753, opened 2020-08-10
[tfldump] Fix comment for PadV2 op

binarman:
Add comment in tflite dumper for PadV2 operator


binarman:
issue #3362

PR #3754, opened 2020-08-10
[res] TF example while_3

seanshpark:
This will introduce another while example slightly modified from while_2
with alternative implementation not using Size operator and
using Placeholder of i instead of constant number so that
constant folding for the operators won't happen

### DIFF
--- a/res/TensorFlowPythonExamples/examples/while_3/__init__.py
+++ b/res/TensorFlowPythonExamples/examples/while_3/__init__.py
@@ -1,0 +1,33 @@
+import tensorflow as tf
+
+x = tf.compat.v1.placeholder(shape=[1, None], dtype=tf.int32, name='Hole')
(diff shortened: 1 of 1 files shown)


res/TensorFlowPythonExamples/examples/while_3/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+    rs = tf.compat.v1.shape(ii)
+    r1 = rs[1]
+    return tf.compat.v1.less(r1, 10)
+
+
+def b(ii):
+    return tf.concat([ii, x], axis=1)
+
+
+# this loop changs i's shape from [1, 0] -> [1, 1] -> [1, 2] -> ... -> [1, 10]
```
hyunsik-yoon, line 17:
```suggestion
# this loop changes i's shape from [1, 0] -> [1, 1] -> [1, 2] -> ... -> [1, 10]
```

Issue #1623, opened 2020-05-28
[Compiler] Create more negative test cases


llFreetimell:
Related issue : #103

### Background

Negative test case is for checking runtime error, segmentation fault, or any unintended behavior.

As we can see in #103, the number of negative test cases is much lower than the number of positive test cases. It is recommended that the number of negative ones should be equal or larger than the number of positive ones.

So let's create more negative test cases!

#### Example code

https://github.com/Samsung/ONE/blob/a200ae65e40278f21a0d12696a611eb241cdb01e/compiler/luci/lang/src/Nodes/CircleIf.test.cpp#L50-L55

#### Convention

Test name should be end with `_NEG`

#### Where to add

Any module which includes some test cases. `luci`, `luci-interpreter`, etc.

seanshpark:
closing old issue

PR #3756, opened 2020-08-11
[luci-interpreter]Negative Tests for Add kernel

struss:
For #1623 

This commit add negative tests on `Add` kernel on luci-interpreter.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Add.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Add.test.cpp
@@ -169,6 +169,33 @@ TEST(AddTest, Float)
   }
 }
 
+TEST(AddTest, Input_Output_Type_NEG)
+{
+  Tensor input1_tensor = makeInputTensor<DataType::FLOAT32>({1}, {1.f});
+  Tensor input2_tensor = makeInputTensor<DataType::S32>({1}, {2});
+  Tensor output_tensor = makeOutputTensor(DataType::FLOAT32);
+
+  AddParams params{};
+  params.activation = Activation::RELU;
+
+  Add kernel(&input1_tensor, &input2_tensor, &output_tensor, params);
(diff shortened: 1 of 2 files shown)


compiler/luci-interpreter/src/kernels/Add.test.cpp
```diff
@@ -169,6 +169,33 @@ TEST(AddTest, Float)
...
+{
+  Tensor input1_tensor = makeInputTensor<DataType::FLOAT32>({1}, {1.f});
+  Tensor input2_tensor = makeInputTensor<DataType::S32>({1}, {2});
+  Tensor output_tensor = makeOutputTensor(DataType::FLOAT32);
+
+  AddParams params{};
+  params.activation = Activation::RELU;
+
+  Add kernel(&input1_tensor, &input2_tensor, &output_tensor, params);
+  EXPECT_ANY_THROW(kernel.configure());
```
struss, line 182: @llFreetimell On assert condition is not caught using `EXPECT_ANY_THROW`. Could I replace `EXPECT_ANY_THROW` into `EXPECT_DEATH` with assert?

seanshpark: CI test passed.  Can you explain more about your concern? 

> assert condition

you've changed from `assert` to `throw`. what do you mean by `assert condition` ?

struss: Oh, currently All Input validation case and some invalid value catch is made up with `assert`. so, I have 2 option, change into throw or keep that and use `EXPECT_DEATH` as a checker.

seanshpark: `assert` will work differently with release build. let's use `throw` with testing.

PR #3766, opened 2020-08-11
[luci-interpreter]Negative Tests for ArgMax kernel

struss:
For #1623

This commit add negative tests on `ArgMax` kernel on luci-interpreter.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/ArgMax.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/ArgMax.test.cpp
@@ -93,6 +93,21 @@ TYPED_TEST(ArgMaxTest, MultiDimensions)
                             /*dimension_data=*/{3}, /*output_data=*/{3, 1});
 }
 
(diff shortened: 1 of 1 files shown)


struss:
@seanshpark @jinevening Change the context.
only add NEG case for Unsupported types.

compiler/luci-interpreter/src/kernels/ArgMax.test.cpp
```diff
@@ -93,6 +93,21 @@ TYPED_TEST(ArgMaxTest, MultiDimensions)
                             /*dimension_data=*/{3}, /*output_data=*/{3, 1});
 }
 
+TEST(ArgMaxTest, UnsupportedType)
```
seanshpark, line 96: `_NEG` suffix if negative test?

struss: Oh I forget! thanks!

PR #3767, opened 2020-08-11
[onert] Enable assertion in PortableTensorRegistry

wateret:
Now that the workaround is gone, so we can enable the assertion in
`PortableTensorRegistryTemplate`.

Plus, null pointer in TensorRegistry is never allowed.

### DIFF
--- a/runtime/onert/core/include/backend/ITensorRegistry.h
+++ b/runtime/onert/core/include/backend/ITensorRegistry.h
@@ -110,22 +110,20 @@ public:
 
   bool setMigrantTensor(const ir::OperandIndex &ind, const std::shared_ptr<IPortableTensor> &tensor)
   {
-    // TODO Uncomment this as two tensors for an index is not allowed.
-    //      But now it is temporarily allowed as a workaround. External one hides Managed one.
-    // auto itr = _native.find(ind);
-    // if (itr != _native.end() && itr->second != nullptr && tensor != nullptr)
-    //  throw std::runtime_error{
-    //      "Tried to set an migrant tensor but an native tensor already exists."};
+    assert(tensor != nullptr);
+    auto itr = _native.find(ind);
+    if (itr != _native.end())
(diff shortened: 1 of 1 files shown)


runtime/onert/core/include/backend/ITensorRegistry.h
```diff
@@ -110,12 +110,9 @@ template <typename T_Tensor> class PortableTensorRegistryTemplate : public ITens
...
   bool setMigrantTensor(const ir::OperandIndex &ind, const std::shared_ptr<IPortableTensor> &tensor)
   {
-    // TODO Uncomment this as two tensors for an index is not allowed.
-    //      But now it is temporarily allowed as a workaround. External one hides Managed one.
-    // auto itr = _native.find(ind);
-    // if (itr != _native.end() && itr->second != nullptr && tensor != nullptr)
-    //  throw std::runtime_error{
-    //      "Tried to set an migrant tensor but an native tensor already exists."};
+    auto itr = _native.find(ind);
+    if (itr != _native.end() && itr->second != nullptr && tensor != nullptr)
```
hyunsik-yoon, line 114: `tensor != nullptr` may need another error message? 
(or maybe simply `assert(tensor)`)

wateret:

runtime/onert/core/include/backend/ITensorRegistry.h
wateret: I just remembered what that means. If tensor is null, I thought it means unsetting the tensor. I'm not sure that's necessary, though.

ragmani: If so, this condition should be changed to like below
```suggestion
    if (itr != _native.end() || tensor == nullptr)
```

wateret: After offline discussion with @ragmani , we concluded that null entry in TensorRegistry is not allowed. Then the condition will be way simpler. Thank you for pointing that out!

PR #3770, opened 2020-08-11
[luci-interpreter]Negative Tests for AveragePool2D kernel

struss:
For #1623

This commit add negative tests on `AveragePool2D` kernel on luci-interpreter.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/AveragePool2D.cpp
+++ b/compiler/luci-interpreter/src/kernels/AveragePool2D.cpp
@@ -35,6 +35,14 @@ AveragePool2D::AveragePool2D(const Tensor *input, Tensor *output, const Pool2DPa
 
 void AveragePool2D::configure()
 {
+  if (input()->element_type() != output()->element_type())
+  {
+    throw std::runtime_error("Input Tensor and Output Tensor Type must be same");
+  }
+  if (input()->shape().num_dims() != 4)
+  {
+    throw std::runtime_error("Input Tensor Shape must be 4-D");
+  }
   const Shape &input_shape = input()->shape();
 
   const int32_t batches = input_shape.dim(0);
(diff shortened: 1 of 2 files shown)


jinevening:
I'm not sure it is ok to check the things (type/shape/quantparam) inside the kernels, because this will make interpreter slow.

IMHO, those checks should be performed in the `import` stage. 

@s-barannikov What do you think of this?

seanshpark:
> IMHO, those checks should be performed in the import stage.

@jinevening , @s-barannikov , there is already merged #3756 . Please check this too.

s-barannikov:
@jinevening 
I don't have strong opinion on that.

On one hand, things like data type should be correct at the time the interpreter is called. `assert` should be enough for such checks. On the other, shapes of outputs are evaluated as operators are executed, and cannot always be statically known to be guaranteed to be correct.

I don't think that adding any checks of the form `if (!condition) throw exception` would slow down the interpreter -- such checks only add are a couple of processor instructions to the hot path, until the `condition` involves some significant amount of computations (comparable to the complexity of the operation itself). But at the same time, adding too many checks would make the code less understandable, we certainly should not check _everything_: TFLite has its own code well tested, and we should not test what is already tested.

In short, I don't mind adding such checks. But it may be worth to introduce some macro analogous to TFLIte's `TFLITE*CHECK*` to reduce the number of `if` statements to make the code more readable.

s-barannikov:
@seanshpark Is the requirement for the negative tests only a [necessary] formality?

seanshpark:
>  Is the requirement for the negative tests only a [necessary] formality?

Um... not sure about `[necessary] formality`  :)
What I know is that our project process requires the number of (new) negative test cases should be higher than (new) positive test cases.

jinevening:
I'm ok with this change as @s-barannikov approved. Please go ahead.

s-barannikov:
I mean, if it is a formality, we could add some dummy tests. For example, try to execute kernels with unsupported data type, such as FLOAT64.

seanshpark:
> if it is a formality, we could add some dummy tests. For example, try to execute kernels with unsupported data type, such as FLOAT64.

Yes, obvious tests like your example is also OK I guess :)
Personally I think they may alarm un-intended changes in the future.

PR #3773, opened 2020-08-11
[onert] Merge operation input dynamic shape checking

hseok-oh:
In static shape inference, use common dynamic input shape checking for all operations except controlflow operation

### DIFF
--- a/runtime/onert/core/src/compiler/StaticShapeInference.cc
+++ b/runtime/onert/core/src/compiler/StaticShapeInference.cc
@@ -25,6 +25,64 @@ namespace onert
 namespace compiler
 {
 
+bool StaticShapeInferer::infer(const ir::OpSequence &op_seq)
+{
+  bool has_dynamic_tensor = false;
+
+  for (const auto &operation_idx : op_seq.operations())
+  {
+    auto &op = _operations.at(operation_idx);
+    auto opcode = op.opcode();
+
+    _return_has_dynamic_tensor = false; // this is used as a return value inside operation's visit()
+
+    // IF: need shape inference for then, else
+    // While: need shape inference for condition, body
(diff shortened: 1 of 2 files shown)


ragmani:
I would be like not to merge things checking dynamic shape. The conditions for making output tensors dynamic can be different for each operation. For example, some operations make output tensor dynamic even if all their input tensors are not dynamic.

hseok-oh:
>some operations make output tensor dynamic even if all their input tensors are not dynamic.

This PR is just merging input dynamic tensor. For example, constant checking for some operation (ex. `ExpandDims`) is remain in each visitor.

ragmani:
@hseok-oh 
I see. I misunderstood it as merging everybody. Thanks for explanation.

runtime/onert/core/src/compiler/StaticShapeInference.cc
```diff
@@ -25,6 +25,64 @@ namespace onert
...
+bool StaticShapeInferer::infer(const ir::OpSequence &op_seq)
+{
+  bool has_dynamic_tensor = false;
+
+  for (const auto &operation_idx : op_seq.operations())
+  {
+    auto &op = _operations.at(operation_idx);
+    auto opcode = op.opcode();
+
+    _return_has_dynamic_tensor = false; // this is used as a return value inside operation's visit()
```
ragmani, line 37: (optional) It would be nice to change this member variable to a local variable.

hseok-oh: OK. I'll fix it.

hseok-oh: It cannot be changed to local variable because this variable is used in visitors.

ragmani: Okay. I did miss out it.

Issue #3776, opened 2020-08-11
Compiler FE: Enable NonMaxSuppressionV5 Op


ai-moiseev:
- [ ] res/TensorFlowPythonExamples
- [ ] tflchef
- [ ] tflite2circle
- [ ] tfldump
- [ ] circledump
- [ ] res/TensorFlowLiteRecipes
- [ ] luci.IR
- [ ] luci.Import
- [ ] luci.Export
- [ ] luci.logex
- [ ] luci.Service
- [ ] luci.tests
- [ ] common artifacts

For #1606

PR #3777, opened 2020-08-11
DRAFT Compiler FE: Enable NonMaxSuppressionV5 Op

ai-moiseev:
Enable NonMaxSuppressionV5 operator.

For #3776


compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -2065,6 +2071,34 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
...
+  {
+    const loco::DataType S32 = loco::DataType::S32;
+
+    auto nmsv5 = dynamic_cast<const luci::CircleNonMaxSuppressionV5 *>(node->input());
+    if (nmsv5 == nullptr)
+      INTERNAL_EXN("CircleNonMaxSuppressionV5 IR is not configured correctly");
+
+    auto index = node->index();
+    if (index == 2)
+      return loco::TensorShape({0});
```
seanshpark, line 2084: Could you provide a reference for this?

compiler/luci/service/src/CircleTypeInferenceRule.cpp
```diff
@@ -589,6 +594,13 @@ struct TypeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::DataT
     return loco::DataType::S32;
   }
 
+  loco::DataType visit(const luci::CircleNonMaxSuppressionV5Out *node) final
+  {
+    (void)node;
+    assert(node->index() >= 0 || node->index() <= 2);
+    return loco::DataType::S32;
```
seanshpark, line 601: TensorFlow op
```
REGISTER_OP("NonMaxSuppressionV5")
    .Input("boxes: T")
    .Input("scores: T")
    .Input("max_output_size: int32")
    .Input("iou_threshold: T")
    .Input("score_threshold: T")
    .Input("soft_nms_sigma: T")
    .Output("selected_indices: int32")
    .Output("selected_scores: T")
    .Output("valid_outputs: int32")
    .Attr("T: {half, float} = DT_FLOAT")
    .Attr("pad_to_max_output_size: bool = false")
```
Second op ` .Output("selected_scores: T")` is not `int32`  . Can you provide where I could reference that all output types are `S32`?

ai-moiseev: Actually, it's float32 in tflite kernel, I've used this file for reference: https://github.com/tensorflow/tensorflow/blob/2a96849f478627ad03b8f20f8c71e85d6d7480e3/tensorflow/lite/kernels/non_max_suppression.cc#L58

I've updated this section to use float32, thank you for this note :)

compiler/luci/service/src/CircleShapeInferenceRule.cpp
ai-moiseev: Ouput `num_selected_indices` is a scalar, maybe I've used wrong 0D tensor shape representation here?
Here's how it's shape is defined in tflite kernel:
https://github.com/tensorflow/tensorflow/blob/2a96849f478627ad03b8f20f8c71e85d6d7480e3/tensorflow/lite/kernels/non_max_suppression.cc#L58

seanshpark: Thanks for the reference. I think current code is OK.

seanshpark:
Code seems OK :) Please proceed with posting PRs.

PR #3808, opened 2020-08-13
[onert] filter optional tensor during checkDynamicInput

glistening:
Operator may have index -1 for input tensor index (optional tensor).
It filters out -1 index to avoid `_Map_base::at`.

### DIFF
--- a/runtime/onert/core/src/compiler/StaticShapeInference.cc
+++ b/runtime/onert/core/src/compiler/StaticShapeInference.cc
@@ -64,7 +64,7 @@ bool StaticShapeInferer::infer(const ir::OpSequence &op_seq)
 
 bool StaticShapeInferer::checkDynamicInput(const ir::Operation &op)
 {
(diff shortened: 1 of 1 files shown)


runtime/onert/core/src/compiler/StaticShapeInference.cc
```diff
@@ -64,7 +64,7 @@ bool StaticShapeInferer::infer(const ir::OpSequence &op_seq)
 
 bool StaticShapeInferer::checkDynamicInput(const ir::Operation &op)
 {
-  for (auto input_idx : op.getInputs())
+  for (auto input_idx : op.getInputs() | ir::Remove::UNDEFINED)
```
ragmani, line 67: (optional) It would be nice to add `ir::Remove::DUPLICATED` too.

glistening:
@hseok-oh, @ragmani I've updated. PTAL

PR #3810, opened 2020-08-13
[onert] Fix weight deallocation to work on android (1.8.0 release)

periannath:
- Weight deallocation is enabled when `__ARM_NEON__` flag is exist
  - This flag is legacy and not supported on android
  - It is replaced by `__ARM_NEON`
- Use both `__ARM_NEON` and `__ARM_NEON__` flags to enable weight deallocation on android

### DIFF
--- a/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
+++ b/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
@@ -220,7 +220,7 @@ void FullyConnectedLayer::prepare()
     }
   }
 
-#if defined(__ARM_NEON__) && defined(USE_RUY_GEMV)
+#if (defined(__ARM_NEON__) || defined(__ARM_NEON)) && defined(USE_RUY_GEMV)
   // TODO This is workaround
   // The only fc hybrid will use ruy kernel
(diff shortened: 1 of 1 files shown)


runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc
```diff
@@ -220,7 +220,7 @@ void FullyConnectedLayer::prepare()
     }
   }
 
-#if defined(__ARM_NEON__) && defined(USE_RUY_GEMV)
+#if defined(__ARM_NEON) && defined(USE_RUY_GEMV)
```
ragmani, line 223: Could you change this like below?
```suggestion
#if (defined(__ARM_NEON__) || defined(__ARM_NEON)) && defined(USE_RUY_GEMV)
```

periannath: @ragmani Thanks! I've applied your suggestion.

periannath:
@chunseoklee I've updated to use both flags following https://github.com/Samsung/ONE/pull/3810#discussion_r469685913. PTAL

PR #3812, opened 2020-08-13
[onert] Change param ReduceType to template in base_loader

ragmani:
This commit changes param ReduceType to template in base_loader.

### DIFF
--- a/runtime/onert/frontend/base_loader/include/base_loader.h
+++ b/runtime/onert/frontend/base_loader/include/base_loader.h
@@ -130,8 +130,8 @@ protected:
   void loadSquaredDifference(const Operator *op, ir::Graph &subg);
   void loadTanh(const Operator *op, ir::Graph &subg);
   void loadTranspose(const Operator *op, ir::Graph &subg);
-  void loadReduce(const Operator *op, ir::Graph &subg,
-                  ir::operation::Reduce::ReduceType reduce_type);
+  template <ir::operation::Reduce::ReduceType reduce_type>
+  void loadReduce(const Operator *op, ir::Graph &subg);
   void loadReduceAll(const Operator *op, ir::Graph &subg);
   void loadReverseV2(const Operator *op, ir::Graph &subg);
   void loadPad(const Operator *op, ir::Graph &subg);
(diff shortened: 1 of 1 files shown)


runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -933,8 +933,8 @@ void BaseLoader<LoaderDomain, SpecificLoader>::loadTranspose(const Operator *op,
 }
 
 template <typename LoaderDomain, typename SpecificLoader>
-void BaseLoader<LoaderDomain, SpecificLoader>::loadReduce(
-    const Operator *op, ir::Graph &subg, ir::operation::Reduce::ReduceType reduce_type)
+template <ir::operation::Reduce::ReduceType reduce_type>
+void BaseLoader<LoaderDomain, SpecificLoader>::loadReduce(const Operator *op, ir::Graph &subg)
```
hseok-oh, line 937: @ragmani Could you explain me what is benefit by this change? IMHO, it just makes a little bit large code size without benefit.

ragmani: Right. This changes makes a little bit larger code size. But this changes make compiler knows value of `reduce_type` at compile time. It can allows compiler to inline the call.
And the gcc compiler handles whether to allow implicitly inline templates or not with a flag `no-implicit-inline-templates`

wateret: Actually I don't see its impact in terms of inlining. The as-is code can be also inlined and then it doesn't have any advantage.

Even though it does, this part does not look like performance-critical part so having less code size looks better here.

ragmani: I have confirmed about it. as-is code was not been inline. 
- template
https://godbolt.org/z/nfYj4K
- arguments
https://godbolt.org/z/Kezffh

Anyway I am going to recover template codes in `base_loader` to use it as a parameter if you are thinking this part does not look like performance-critical part.

wateret: Templates are inlined because of its nature, and IMHO gcc did not inline `print` as the size of  is too large to inline. (I believe that there is a threshold for inlining a function)

wateret: https://godbolt.org/z/95Ynv9 I added gcc attribute `__attribute__((always_inline))` to force inlining, and it looks like it gives even further optimized code.

Issue #3809, opened 2020-08-13
[onert] Refactoring : Distinguish roles of TensorBuilder and TensorRegistry


wateret:
I'd like to distinguish roles of `TensorBuilder` and `TensorRegistry`, I will try to make a draft for it.

And I'm thinking fixed roles for these like

- TensorBuilder : A builder object that creates tensors. Collects tensor info and creates a TensorRegistry as a result.
- TensorRegistry : A registry of tensor objects for both static and dynamic ones

Things to consider

- [x] `TensorRegistry` for acl backends
- [x] Remove `TensorBuilder::tensorAt`(since it is a duplication with `TensorRegistry`'s)
- [x] Remove `TensorBuilder::iterate`
- [x] ~~Make static tensors also registered in `TensorRegistry`~~
- [x] `TensorRegistry` as part of `BackendContext`
- [x] `TensorBuilders` → `TensorRegstries`?

PR #3813, opened 2020-08-13
[onert] Refactoring: Controlflow DynamicTensorMan

wateret:
- Unify two tensor registries in `controlflow::DynamicTensorManager`
- Do not register model input/outputs during compilation which makes
  unused (cpu) tensors

### DIFF
--- a/runtime/onert/core/src/backend/controlflow/DynamicTensorManager.cc
+++ b/runtime/onert/core/src/backend/controlflow/DynamicTensorManager.cc
@@ -17,6 +17,7 @@
 #include "DynamicTensorManager.h"
 
 #include "util/logging.h"
+#include "ir/DataType.h"
 
 namespace onert
 {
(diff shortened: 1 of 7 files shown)


wateret:
This is necessary for proceeding #3809

hyunsik-yoon:
Not a scope of this PR but I have a question.
While reading `setMigrantTensor`, if there are another backend B with 100 native tensors, I wonder which tensors (native to backend B) among the 100 tensors should be set with `setMigrantTensor`.
It seemed that `setMigrantTensor` is called only for IO tensors. So, how about renaming it to, e.g., `setMigrantUserTensor`?

wateret:
> It seemed that setMigrantTensor is called only for IO tensors. So, how about renaming it to, e.g., setMigrantUserTensor?

@hyunsik-yoon Good point. For now it is true. But the logic is registering migrant tensors for the tensor indices that is used but not registered as native tensors.

When the permutation elimination pass' removal condition becomes general(not just between CF and CPU backend), other tensors as well as I/O tensors are also added as migrant tensors too.

PR #3710, opened 2020-08-07
[onert] Remove unused function for munmap

periannath:
- This commit removes `deallocateMmappedArea` function
  - `deallocateMmappedArea` is now integrated into `MMapedData` by #2764
  - This function is no longer used

ONE-DCO-1.0-Signed-off-by: JiHwan Yeo <jihwan.yeo@samsung.com>

PR #2764, opened 2020-07-01
[onert] Use MMapedData instead of CachedData

YongseopKim:
Use MMapedData derived from ExternalData  instead of CachedData

ONE-DCO-1.0-Signed-off-by: Yongseop Kim <yons.kim@samsung.com>

PR #3817, opened 2020-08-13
[onert] Use CachedData instead of MMapedData (1.8.0)

periannath:
- MMapedData reduces load/prepare time, but increases inference time
- Use CachedData for weight tensor for inference performance
  - Includes Revert "[onert] Remove unused function for munmap (#3710)"

---

- This PR partially reverts https://github.com/Samsung/ONE/pull/2764


chunseoklee:
PTAL @Samsung/nnfw

Issue #3811, opened 2020-08-13
[onert] assertion failure during DIV with different shape


glistening:
After #3782 is merged, my target model got assertion failure:

```
$ Product/x86_64-linux.debug/out/bin/nnpackage_run --nnpackage dec
Package Filename dec
nnpackage_run: /home/brian/z/ONE/compute/cker/include/cker/Shape.h:290: int nnfw::cker::MatchingFlatSize(const nnfw::cker::Shape&, Ts ...) [with Ts = {nnfw::cker::Shape, nnfw::cker::Shape}]: Assertion `checkMatching(shape, std::forward<Ts>(check_shapes)...)' failed.
Aborted (core dumped)
```

It is caused #3782 roll backed the change of #2153.

We may provide specialize version of `eval` for `DIV`.

```
template <nnfw::cker::BinaryArithmeticOpType arithmetic_type, typename T>
void eval(const IPortableTensor *lhs, const IPortableTensor *rhs, IPortableTensor *output,
          nnfw::cker::BinaryArithmeticOpParam op_params)
```

For release build, my model works.

glistening:
I've made #3819. I've confirmed tensorflow lite `v2.3.0` uses `!HaveSameShape` for checking broadcast for `div`, `pow`, `mul`, `add`, `sub`, ... However, I just roll back `div` only in #3819 because changing condition may affect several model's performance or correctness in our source.

PR #3782, opened 2020-08-12
[onert] Unify binary arithmetic operations to BinaryArithmetic

ragmani:
For issue #3625

This commit unifies binary arithmetic operations to BinaryArithmetic for reducing code duplication.
  - BinaryArithmetic ops: Add, Sub, Mul, Div
  - Unify binary arithmetic operations
  - Remove checking workaround in HESchedule

Signed-off-by: ragmani <ragmani0216@gmail.com>

PR #2153, opened 2020-06-11
[onert] div will check requires_broadcast with haveSameShape

glistening:
Problem:
When div has a tensor with 1 element size and a scalar as inputs,
ProcessBroadcastShapes says broadcast is not required.
However, the path for non-broadcast requires the inputs has same shape.

Solve:
It will check requires_broadcast with haveSameShape as tensorflow lite does.

ONE-DCO-1.0-Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>

PR #3819, opened 2020-08-13
[onert] div will use MatchElementsSize instead of MatchFlatSize

glistening:
MatchFlatSize checks whether tww inputs are same dimension.
It will cause assertion failure. Thus it will use MatchElementsSize.

### DIFF
--- a/compute/cker/include/cker/operation/reference/BinaryArithmeticOps.h
+++ b/compute/cker/include/cker/operation/reference/BinaryArithmeticOps.h
@@ -37,7 +37,7 @@ inline void BinaryArithmeticOp(const BinaryArithmeticOpParam &params, const Shap
                                const T *input2_data, const Shape &output_shape, T *output_data,
                                const std::function<T(const T &, const T &)> &fn)
 {
-  const int32_t flat_size = MatchingFlatSize(input1_shape, input2_shape, output_shape);
+  const int32_t flat_size = MatchingElementsSize(input1_shape, input2_shape, output_shape);
   for (int i = 0; i < flat_size; ++i)
   {
     output_data[i] = ActivationFunctionWithMinMax(fn(input1_data[i], input2_data[i]),
@@ -53,7 +53,7 @@ inline void BinaryArithmeticOp(const BinaryArithmeticOpParam &params, const Shap
                                float *output_data,
                                const std::function<float(const float &, const float &)> &fn)
 {
-  const int size = MatchingFlatSize(input1_shape, input2_shape, output_shape);
+  const int size = MatchingElementsSize(input1_shape, input2_shape, output_shape);
   for (int i = 0; i < size; i++)
   {
     output_data[i] =


runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
```diff
@@ -34,8 +34,12 @@ template <nnfw::cker::BinaryArithmeticOpType arithmetic_type, typename T>
 void eval(const IPortableTensor *lhs, const IPortableTensor *rhs, IPortableTensor *output,
           nnfw::cker::BinaryArithmeticOpParam op_params)
 {
-  const bool need_broadcast =
-      nnfw::cker::ProcessBroadcastShapes(getTensorShape(lhs), getTensorShape(rhs), &op_params);
+  bool need_broadcast = false;
+  if (arithmetic_type == nnfw::cker::BinaryArithmeticOpType::DIV)
+    need_broadcast = !HaveSameShapes(lhs, rhs);
```
ragmani, line 39: It would be better to fix `ProcessBroadcastShapes()` than change this to call `HaveSameShapes`. `need_broadcast` is used for calling functions of `cker`.

hseok-oh: It may not share `ProcessBroadcastingShapes()` implement because `DIV` operation comes from different tflite version. If we fix `ProcessBroadcastingShapes()`, `ProcessBroadcastingShapes()` may have operation checking and use different flow.

ragmani: I see, I just checked TFLite interpreter code. `ProcessBroadcastingShapes()` is older version. I s old version faster than new version of `v2.3.0`?

glistening: @ragmani As I wrote in https://github.com/Samsung/ONE/issues/3811#issuecomment-673289378 yesterday, recent versions (`v2.3.0` and perhaps including `v2.2.0` and so on) use `!HavsSameShape`. I did not update the condition since I did not measure all models and am not sure for performance. It may be done by another issue. @hseok-oh, @periannath and I already discussed and choose this solution in #2153. This commit roll back to previous one before #3782 for `Div`.

glistening:
@ragmani, @hseok-oh 
@periannath told me that tensorflow lite uses both of `HavaSameShape()` (in prepare...) and `ProcessBroadcastingShapes()` (in eval...). I've updated to use `MatchElementsSize` the same way in recent tensorflow lite. It does not check dimension count, thus assertion failure not occurs.

runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
ragmani: I got it.

Issue #3820, opened 2020-08-13
[onert] acl_cl backend fails to initialize


dr-venkman:
I am currently running a tflite model on a 64-bit target platform running Tizen. The build was accordingly set to `aarch64`, which worked ok. However, when I force `nnpackage_run` to use `acl_cl` backend, it gives off a segmentation fault soon after it fails to initialize.

Please find the details below:

```
$. ONERT_LOG_ENABLE=1 BACKENDS="acl_cl" $PROJECT_ROOT/Product/out/bin/nnpackage_run -w 100 -r 100 ./my_nnpackage/

Package Filename ./my_nnpackage/
[Compiler] [Compiler] ==== Compiler Options ====
[Compiler] backend_list             : acl_cl
[Compiler] trace_filepath           : 
[Compiler] graph_dump_level         : 0
[Compiler] op_seq_max_node          : 0
[Compiler] executor                 : Linear
[Compiler] manual_scheduler_options : (Too many things to print)
[Compiler] he_scheduler             : false
[Compiler] he_profiling_mode        : false
[Compiler] disable_compile          : false
[Compiler] fp16_enable              : false
[Compiler] [loadBackend] Successfully loaded 'acl_cl' - libbackend_acl_cl.so
[onert_backend_create] 'acl_cl' loaded
[loadBackend] acl_cl backend initialization failed. Don't use this backend
[onert_backend_create] 'acl_cl' unloaded
Segmentation fault (core dumped)
```

On further inspection, it turns out that the check for OpenCL fails at the following line (at location `runtime/onert/backend/acl_cl/Config.cc: bool Config::initialize()`):
```
if (!arm_compute::opencl_is_available())
  {
    return false;
  }
```
I assume that the implementation details for `opencl_is_available()` are in `lib_armcompute.so`. Could anyone please suggest how to troubleshoot the issue further? I checked for `lib_armcompute.so`, and `libOpenCL.so` at the target, and they exist as symlinks shown below:

```
sh-3.2# ls -l /usr/lib/libarm_compute.so                                                                                                                                                                  
-rwxr-xrwx 1 root root 67322276 Jul  8  2020 /usr/lib/libarm_compute.so

sh-3.2# ls /usr/lib/libOpenCL.so -l                                                                                                                                                                        
lrwxrwxrwx 1 root root 30 Feb 19 03:46 /usr/lib/libOpenCL.so -> /usr/lib64/driver/libOpenCL.so
[...]

PR #3829, opened 2020-08-14
[onert] Throw when no backends loaded

### LINKED ISSUE #3826: [onert] Handle when no backends were loaded
https://github.com/Samsung/ONE/issues/3820#issuecomment-673859405

wateret:
Throw when no backends were loaded rather than assertion.

Add a regression test.

Resolve #3826


wateret:
@dr-venkman It would be nice if you can check #3820 does not crash(Segfault) but just fails execution.

wateret:
Merging as this commit is useful regardless of #3820 .

PR #3823, opened 2020-08-14
DRAFT: Support requantization from int8 to uint8

jinevening:
On going draft to support requantization from int8 to uint8

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to #3779

Issue #3779, opened 2020-08-12
[circle-quantizer] Support requantization from int8 tflite to uint8 circle


jinevening:
We'll support requantization between models quantized with different schemes. This will enable
1) to run models quantized by other frameworks on uint8-based hardware.
2) to use test data generated by other frameworks in ONE.

In this issue, we support requantization from int8 tflite (quantized by TF 2.x) to uint8 circle.

### Workflow

int8 tflite -> int8 circle (by tflite2circle) -> uint8 circle (by circle-quantizer)

### Requantization (int8 -> uint8)

TF 2.x int8 quantization 
activation: range -128\~127, type int8
weights: range -127~127, type int8
bias: type int32

ONE uint8 quantization
activation: range 0\~255, type uint8
weights: range 0~255, type uint8
bias: type int32

For requantization, we shift the zero points and the weights of the source model (int8 tflite) by 128 while keeping the values of scales.

mhs4670go:
@jinevening Just curiosity, what is the purpose of quantization with TF 2.x? I thought `circle-quantizer` would do that.

jinevening:
> Just curiosity, what is the purpose of quantization with TF 2.x?

The major reason is to generate the test data (inference result) for CWQ uint8 quantized model. Since there is no proved kernel for CWQ uint8 quantized model, we're going to make the test data by running CWQ int8 quantized model on TFLite interpreter and converting the result to uint8.

Another reason is to run models quantized by TF 2.x on the hardware based on uint8 quantization.

s-barannikov:
@jinevening 
> For requantization, we shift the zero points and the weights of the source model (int8 tflite) by 128 while keeping the values of scales.

Is this transformation correct? Should not the zero_point == 0 in the int8 case?

jinevening:
> Is this transformation correct? Should not the zero_point == 0 in the int8 case (isn't it symmetric)?

I think this is correct (please tell me if it is wrong).

In tflite int8 model, zero point of weights is always zero as you mentioned.

For requantization (int8 -> uint8), we add 128 to the weights values to shift the int8 weights (-127\~127) to uint8 range (1\~255). (0 is not used as symmetric int8 uses only 254 numbers). zero point should be also shifted as the range is changed.

This can be shown with simple math.

real value = scale * (Q_i8 - zp_i8) = scale * (Q_u8 - zp_u8)

where,
scale: int8 scale (same with uint8 scale)
Q_i8 : int8 quantized weights
[...]

PR #3830, opened 2020-08-14
[luci] Support for requantization from int8 to uint8

jinevening:
This supports requantization from int8 to uint8

---
Related to #3779 Draft PR #3823

### DIFF
--- a/compiler/luci/pass/src/RequantizePass.cpp
+++ b/compiler/luci/pass/src/RequantizePass.cpp
@@ -1,0 +1,241 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


compiler/luci/pass/src/RequantizePass.cpp
```diff
@@ -0,0 +1,276 @@
...
+        continue;
+
+      // Check if this was not quantized (only quantized tensors are requantized)
+      if (circle_node->quantparam() == nullptr)
+        continue;
+
+      if (_input_type == loco::DataType::S8 && _output_type == loco::DataType::U8)
+      {
+        // Check if this is not bias (bias is not requantized when int8 -> uint8)
+        auto iw = get_input_weight_of_bias(circle_node);
```
mhs4670go, line 167: Could you explain about `get_input_weight_of_bias` function more detail? It seems there is no other place that uses this function. But then, return value doesn't seem necessary. I mean, just true, false values are enough.

jinevening: You're right. The return values are not used in this pass. They are used to calculate the scales of bias in QuantizedWithMinMaxPass. I just copied the function from the source code. I've updated the code to remove the redundant codes.

jinevening: Also, I've updated the code to correctly handle const tensors (not weights nor bias). PTAL.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

parjong (approved):

PR #3852, opened 2020-08-18
[nnpackage_tool] update tflite2circle

glistening:
- Update tensorflow schema version v1.12.0 -> v1.13.1
- Remove unnecessary default value writings in json


tools/nnpackage_tool/tflite2circle/tflite2circle.sh
```diff
@@ -73,7 +73,7 @@ name=${tflite_base%.*}
 # convert
 
 mkdir -p "${outdir}"
-${flatc} -o ${outdir} --defaults-json --strict-json -t ${tflite_schema} -- $1
+${flatc} -o ${outdir} --strict-json -t ${tflite_schema} -- $1
```
mhs4670go, line 76: Is there any reason to remove this option? All default values are unnecessary?

glistening: @mhs4670go Yes, default value in json is for human. flatbuffer does not save value for default values.

mhs4670go (approved):
LGTM

PR #3857, opened 2020-08-18
[luci-interpreter]Update to use Tensorflow 2.3.0 on luci-interpreter.

struss:
This commit update to use Tensorflow 2.3.0 on luci-interpreter.

  1. Update cmake file in luci-interpreter kernel to include some file in Tensorflow(Mean kernel use this.)
  2. Update kernels for changed spec from 2.1.0 to 2.3.0.(`LeakyRelu`, `Mul`)
	`LeakyRelu` : Arguments are changed.
	`Mul` : kernel changed to reference to optimized.
  3. Update Kernel that accept more tensor.(handle 3~4 Tensor both).`TransposeConv`
	currently, accept nullptr on 4th Tensor. it will update later.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/CMakeLists.txt
+++ b/compiler/luci-interpreter/src/kernels/CMakeLists.txt
@@ -63,7 +63,8 @@ set(SOURCES
     TransposeConv.h
     TransposeConv.cpp
     Unpack.h
-    Unpack.cpp)
+    Unpack.cpp
+    ${TensorFlowSource_DIR}/tensorflow/lite/kernels/internal/quantization_util.cc)
 
 list(APPEND SOURCES Utils.h Utils.cpp)
 
@@ -71,6 +72,7 @@ add_library(luci_interpreter_kernels STATIC ${SOURCES})
 set_target_properties(luci_interpreter_kernels PROPERTIES POSITION_INDEPENDENT_CODE ON)
 target_include_directories(luci_interpreter_kernels PUBLIC ${LUCI_INTERPRETER_SOURCE_DIR})
 target_include_directories(luci_interpreter_kernels SYSTEM PRIVATE
+    "${TensorFlowRuySource_DIR}"
     "${TensorFlowGEMMLowpSource_DIR}"
     "${TensorFlowEigenSource_DIR}"
     "${TensorFlowSource_DIR}")
--- a/compiler/luci-interpreter/src/kernels/CMakeLists.txt
+++ b/compiler/luci-interpreter/src/kernels/CMakeLists.txt
@@ -63,10 +63,9 @@ set(SOURCES
     TransposeConv.h
     TransposeConv.cpp
     Unpack.h
-    Unpack.cpp
-    ${TensorFlowSource_DIR}/tensorflow/lite/kernels/internal/quantization_util.cc)
+    Unpack.cpp)
 
-list(APPEND SOURCES Utils.h Utils.cpp)
+list(APPEND SOURCES Utils.h Utils.cpp ${TensorFlowSource_DIR}/tensorflow/lite/kernels/internal/quantization_util.cc)
 
 add_library(luci_interpreter_kernels STATIC ${SOURCES})
 set_target_properties(luci_interpreter_kernels PROPERTIES POSITION_INDEPENDENT_CODE ON)
--- a/compiler/luci-interpreter/src/kernels/TransposeConv.cpp
+++ b/compiler/luci-interpreter/src/kernels/TransposeConv.cpp
@@ -30,8 +30,8 @@ namespace kernels
 {
 
 TransposeConv::TransposeConv(const Tensor *output_shape, const Tensor *filter, const Tensor *input,
-                             Tensor *output, const TransposeConvParams &params)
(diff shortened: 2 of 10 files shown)


compiler/luci-interpreter/src/kernels/TransposeConv.test.cpp
```diff
@@ -50,8 +50,8 @@ void Check(std::initializer_list<int32_t> output_shape_shape,
   params.stride_height = stride_height;
   params.stride_width = stride_width;
 
-  TransposeConv kernel(&output_shape_tensor, &weight_tensor, &input_data_tensor, &output_tensor,
-                       params);
+  TransposeConv kernel(&output_shape_tensor, &weight_tensor, &input_data_tensor, nullptr,
```
s-barannikov, line 53: It would be nice to test non-zero bias.

compiler/luci-interpreter/src/loader/KernelBuilder.test.cpp
```diff
@@ -693,6 +693,7 @@ TEST_F(KernelBuilderTest, TransposeConv)
   auto *output_shape = createInputNode();
   auto *filter = createInputNode();
   auto *input = createInputNode();
+  auto *bias = createInputNode();
```
s-barannikov, line 696: There should be corresponding `checkTensor` below.

compiler/luci-interpreter/src/kernels/CMakeLists.txt
```diff
@@ -63,14 +63,16 @@ set(SOURCES
     TransposeConv.h
     TransposeConv.cpp
     Unpack.h
-    Unpack.cpp)
+    Unpack.cpp
+    ${TensorFlowSource_DIR}/tensorflow/lite/kernels/internal/quantization_util.cc)
```
s-barannikov, line 67: Please move this file to the list below (line 69).

struss: Okay, I will fix it.

compiler/luci-interpreter/src/kernels/TransposeConv.test.cpp
struss: okay, I will add a new test case for non-zero bias test case. after that, I will make a notification.

compiler/luci-interpreter/src/loader/KernelBuilder.test.cpp
struss: Currently, `CircleTransposeConv` doesn't have a bias term. so after update luci commented on 3, it will be updated more.

s-barannikov: I see. In this case this line is unnecessary.
Please add a TODO comment to not forget to update the test when luci IR has support for the bias term.

struss: Okay I will add it.

struss: Add TODO and some comments that will be comment out after luci updated.

compiler/luci-interpreter/src/kernels/TransposeConv.cpp
```diff
@@ -106,8 +106,9 @@ void TransposeConv::evalFloat() const
   op_params.output_multiplier = _output_multiplier;
   tflite::reference_ops::TransposeConv(
       op_params, getTensorShape(input()), getTensorData<float>(input()), getTensorShape(filter()),
-      getTensorData<float>(filter()), getTensorShape(output()), getTensorData<float>(output()),
-      tflite::RuntimeShape(), (float *)nullptr);
+      getTensorData<float>(filter()), getTensorShape(bias()), getTensorData<float>(bias()),
+      getTensorShape(output()), getTensorData<float>(output()), tflite::RuntimeShape(),
+      (float *)nullptr);
```
s-barannikov, line 111: (just for information) You don't have to cast nullptr to the targe type, it is implicitly cast.

compiler/luci-interpreter/src/kernels/TransposeConv.test.cpp
```diff
@@ -26,15 +26,15 @@ namespace
 
 using namespace testing;
 
-template <typename T>
+template <typename T, typename B>
```
s-barannikov, line 29: It seems that B is always `float`. Is in necessary to add this template parameter?

s-barannikov (approved):
LGTM

compiler/luci-interpreter/src/kernels/TransposeConv.test.cpp
struss: in Quantized case, it will the bias params are `int32`. so, for those cases, I split into T and B

s-barannikov: I see, thanks.

compiler/luci-interpreter/src/kernels/TransposeConv.cpp
struss: Oh thanks! I will keep in mind!

llFreetimell (approved):
LGTM

PR #3863, opened 2020-08-19
[onert] Throw an exception when tensor_registry is nullptr

hyunsik-yoon:
This changes `assert` to `throw` when `tensor_register == nullptr`.

When we run some model with several subgraphs with cpu backend, 
if `BACKEND=cpu` _is not provided by mistake_, assert checks nullptr. However, with _release_ build, onert keeps running and dies with segmentation fault. 

So changing `assert` to `throw` seems better to prevent this mistake.


wateret:
Actually I'm working on every backend must have a TensorRegistry. It is almost done. And then I think this should be `assert`. Could you wait for that a bit?

ragmani:
@wateret  Is it done?

wateret:
@hyunsik-yoon Can this be closed?

hyunsik-yoon:
yeap. thanks.

PR #3866, opened 2020-08-19
[onert] Upgrade cmake version for Android docker file

hyunsik-yoon:
This upgrades cmake version from 3.10.2 to 3.16.2 for Android docker file.

Reason:
  - with 3.10.2, errors like "cannot find cmake target flatbuffer::flatbuffer" is thrown several times during building onert.

### DIFF
--- a/infra/docker/Dockerfile.1804
+++ b/infra/docker/Dockerfile.1804
@@ -6,7 +6,17 @@ ARG UBUNTU_MIRROR
 RUN apt-get update && apt-get -qqy install software-properties-common
 
 # Build tool
-RUN apt-get update && apt-get -qqy install build-essential cmake scons git lcov g++-arm-linux-gnueabihf g++-aarch64-linux-gnu
+RUN apt-get update && apt-get -qqy install build-essential scons git lcov g++-arm-linux-gnueabihf g++-aarch64-linux-gnu
+
+# Download cmake (`apt-get install cmake` installs version 3.10.2 but we need more recent version.
+# With 3.10.2, we will get some error, e.g., "Cannot find a target flatbuffer::flatbuffer"
+RUN apt-get -qqy install libssl-dev wget
+WORKDIR /tmp
+RUN wget -q https://cmake.org/files/v3.16/cmake-3.16.2.tar.gz
+RUN tar -xvzf cmake-3.16.2.tar.gz
+RUN cd cmake-3.16.2 && ./bootstrap --prefix=/usr/local && make && make install
+RUN ln -s /usr/local/bin/cmake /usr/bin/cmake
+RUN rm -rf cmake-3.16.2*
 
 # Install extra dependencies (Caffe, nnkit)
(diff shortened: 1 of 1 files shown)


glistening:
@hyunsik-yoon Does the error occurs only for android?
@chunseoklee You also had same problem. Did it also happen for android build?

wateret:
What about just updating it to 16.04? When I first introduced `Dockerfile.1804`, it was just to use CMake 3.10, nothing more. So if we change the default version of CMake, I think we can just do it to 16.04 then we can remove 18.04.

hseok-oh:
If you want to use latest cmake versoin, I recommand to use official repo: `apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'`

hyunsik-yoon:
> @hyunsik-yoon Does the error occurs only for android?

As for me, I build onert for XU4 with `./infra/scripts/docker_build_cross_arm_runtime.sh` with Bionic rootfs and have not experience this error before.

I just hope that other people should not experience same error and spend their time to solve it. :-)

glistening:
> What about just updating it to 16.04? When I first introduced `Dockerfile.1804`, it was just to use CMake 3.10, nothing more. So if we change the default version of CMake, I think we can just do it to 16.04 then we can remove 18.04.

You had tried to [update CMake default version on Sep 26, 2019](https://github.sec.samsung.net/STAR/nnfw/pull/7763#issuecomment-388665). But we couldn't because of tizen cmake version requirement.

glistening:
@hseok-oh 

> If you want to use latest cmake versoin, I recommand to use official repo: `apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'`

On Ubuntu 18.04, adding this repo doesn't work since it installed some version older than `v3.16.2`, which shows same error.

glistening:
I think two solution has its own drawbacks.

1. Update the cmake version required version for `onert`
- We need to check tizen's cmake version requirement.
- All users has to update their cmake version. Adding repository doesn't work. He/She need to download and build and install.

2. Update 18.04 docker image like this PR
- Someone who don't use our docker still see the same error.

Thus, it would be better to find solution that works with cmake 3.10.x or the default version after adding cmake repo, if possible.

hyunsik-yoon:
Maybe I was in hurry to decide this. :-)
Like @glistening mentioned, this was caused not by CMake(?). I guess this needs some more test. 
Let me change this PR into **draft** and close some time later to leave some record about problem on Android build.

wateret:
> You had tried to [update CMake default version on Sep 26, 2019](https://github.sec.samsung.net/STAR/nnfw/pull/7763#issuecomment-388665). But we couldn't because of tizen cmake version requirement.

I see, thanks. Anyways the docker image is not necessarily 18.04. We may just change it to 20.04 rather than building CMake in 18.04. (If the cmake version really matters 😄 )

PR #3880, opened 2020-08-19
[onert] Remove `ITensorBuilder::tensorAt`

wateret:
Remove `ITensorBuilder::tensorAt` to distinguish the roles of
`ITensorBuilder` and `ITensorRegistry`. A builder is a builder and a
registry contains built tensors.

Note that KernelGenerator and ConstantInitializer now use TensorRegistry
to find tensor objects which takes most part of the changes.

### DIFF
--- a/runtime/onert/backend/acl_common/AclTensorRegistry.h
+++ b/runtime/onert/backend/acl_common/AclTensorRegistry.h
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+#ifndef __ONERT_BACKEND_ACL_COMMON_ACL_TENSOR_REGISTRY_H__
+#define __ONERT_BACKEND_ACL_COMMON_ACL_TENSOR_REGISTRY_H__
+
 #include "backend/ITensorRegistry.h"
 
 namespace onert
@@ -43,6 +46,8 @@ public:
     return getITensor(ind);
   }
 
+  auto getAclTensor(const ir::OperandIndex &ind) { return _tensor_mgr->at(ind); }
+
 private:
   T_AclTensorManager *_tensor_mgr;
 };
@@ -50,3 +55,5 @@ private:
 } // namespace acl_common
 } // namespace backend
 } // namespace onert
+
+#endif // __ONERT_BACKEND_ACL_COMMON_ACL_TENSOR_REGISTRY_H__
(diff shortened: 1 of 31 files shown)


wateret:
Part of #3809

runtime/onert/backend/acl_common/AclTensorRegistry.h
```diff
@@ -43,10 +46,14 @@ template <typename T_AclTensorManager> class AclTensorRegistry : public ITensorR
     return getITensor(ind);
   }
 
+  auto getAclTensor(const ir::OperandIndex &ind) { return _tensor_mgr->at(ind); }
```
hyunsik-yoon, line 49: I wonder the reason of introducing this method instead of using `getITensor()`. Could you explain a bit? Thanks.

runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -221,7 +221,7 @@ void ExecutorFactory::prepareExternalTensors(ir::LoweredGraph &lowered_graph,
           // If an OpSequence input/output tensor does not have a own tensor object,
           // it must be using external tensors, so find the tensor from other tensor builders and
           // set the tensor to this tensor builder if portable
-          if (!backend_ctx->tensor_builder->tensorAt(ind))
+          if (!backend_ctx->tensor_builder->tensorRegistry()->getITensor(ind))
```
hyunsik-yoon, line 224: Not a scope of this PR but
what if we store `tensorRegistry` into `backend_ctx` instead of getting it via `tensor_builder`, e.g., 

```
          if (!backend_ctx->tensorRegistry()->getITensor(ind))
```

Since `tensorRegistry` is a kinda global (?) registry whose lifetime spans from compilation to the end of execution. Therefore, I guess that it would be easier if we access `tensorRegistry` through one object (`backend_ctx`).

wateret: Good point. It is going to be done, planned in #3809 already.

runtime/onert/backend/acl_common/AclTensorRegistry.h
wateret: It returns the actual implementation of `ITensor`(`CLTensor` or `NETensor`). If it was a raw pointer we can just use [Covariant Return Type](https://www.lwithers.me.uk/articles/covariant.html), but it is not possible with `shared_ptr`.

Plus, I made the return type `auto` otherwise the template code will get more complicated.

hyunsik-yoon (approved):
LGTM

Issue #3859, opened 2020-08-19
Compile FE: Merge same constants


seanshpark:
If there are two or more Constant nodes that are identical in content, we can think of removing duplicates
For example , for this network
```
[ifm] ----- [Add] ----- [Add] --- [ofm]
 [c1] --/    [c2] --/
```
we could change like this if `c1` and `c2` are identical
```
[ifm] ----- [Add] ----- [Add] --- [ofm]
 [c1] --/-----------/
```

Let's find out about this.

seanshpark:
Things to do
- [ ] prepare a small network with identical constants and some differenent values
- [ ] a pass that updates the network
- [ ] model difference?
- [ ] run in OneRT

mhs4670go:
I think this pass will have to be executed after another fusion-like pass has been executed. Maybe other pass could be related as well.

seanshpark:
Prepared draft code checks inside a graph, that is sub-graph in circle file.
Our target inhouse model is like indentical constant across different sub-graphs.
Current luci cannot support this.
Need to do some design...

seanshpark:
How to, in view of TF lite file,
- For Buffers that are identical,
- Update the Tensor that points to the second and above Buffers to the first Buffer
- Remove duplicate (second and above) Buffers

Overal how to, in _luci_
- We need a Class to store Buffer: say CircleFileBuffer
- Update CircleConst to point to the CircleFileBuffer object instead of holding the content itself

How to in Import
- ...

seanshpark:
@llFreetimell  finished this :)

llFreetimell:
For records, #4754 resolved it :)

PR #3897, opened 2020-08-20
DRAFT: Compiler FE: Remove identical constants

seanshpark:
On-going draft to reduce storage by removing identical constants


seanshpark:
For #3859

seanshpark:
close as not used

PR #3896, opened 2020-08-20
[draft][onert/test] Operation backend test by using nnfw API

hseok-oh:
- Move GenModelTest to header file
- Introduce backend setting method in GenModelTest
- Move test code for operation backend test

Signed-off-by: Hyeongseok Oh <hseok82.oh@samsung.com>

PR #3911, opened 2020-08-21
[onert/test] Split GenModelTests.cc

hseok-oh:
- Header file for GenModelTest class
- Divide test for each operation

---

part of #3896


wateret:
What about changing the folder name`ops` to something like `one_op_tests`? `ops` could be confusing as other types of tests also exist.

tests/nnfw_api/src/GenModelTests.h
```diff
@@ -0,0 +1,108 @@
+/*
```
wateret, line 1: As the class name is `GenModelTest` so what about renaming this file to `GenModelTest.h`?

hseok-oh:

hseok-oh:

Issue #3881, opened 2020-08-19
Compiler FE: Analysis score is dropping


seanshpark:
One of our internal software analysis tool reports the score is getting low: 4.07 as of today.

Let's get this higer.

seanshpark:
Previous related issue: #2843

seanshpark:
done for now

PR #3914, opened 2020-08-21
[luci] Reduce duplicate codes in Conv2D shape-inf

seanshpark:
This will reduce duplicate codes in Conv2D type shape inference


seanshpark:
For #3881

PR #3905, opened 2020-08-21
DRAFT Compiler FE care for analysis score

seanshpark:
on-going draft to care our static analysis score

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #2919, opened 2020-07-06
Compiler FE Refactoring locomotiv NodeExecution class


seanshpark:
We can reduce LOC of class `NodeExecution` in _locomotiv_ like #2904 change.

From #2843

Purpose of this refactoring is to reduce total sum of `Lines Of Code` in the class.

seanshpark:
done

PR #3917, opened 2020-08-21
[locomotiv] Reduce class codes D

seanshpark:
This will rename methods to reduce class lines of codes of NodeExecution


seanshpark:
For #2919 from draft #3905

PR #3920, opened 2020-08-21
[luci] Import validate helpers to reduce dup codes

seanshpark:
This will introduce ValidateHelpers to reduce duplicate codes


seanshpark:
For #3881 from draft #3905

llFreetimell (approved):
LGTM

Issue #3764, opened 2020-08-11
[base_loader] Add GenModel test cases to improve Code Coverage


wateret:
Here's the list of functions in `base_loader.h::loadOperation` that never be tested.

- [ ] loadSelect(op, subg);
- [ ] loadSquaredDifference(op, subg);
- [ ] loadReduce(op, subg, ir::operation::Reduce::ReduceType::ANY);
- [ ] loadReverseV2(op, subg);
- [ ] loadPrelu(op, subg);
- [ ] loadSplit(op, subg);
- [ ] loadSplitV(op, subg);
- [ ] loadUnpack(op, subg);
- [ ] loadCast(op, subg);
- [ ] loadOneHot(op, subg);
- [ ] loadAbs(op, subg);
- [ ] loadCos(op, subg);
- [ ] loadSin(op, subg);
- [ ] loadReduce(op, subg, ir::operation::Reduce::ReduceType::PROD);
- [ ] loadArgMax(op, subg);
- [ ] loadLog(op, subg);
- [ ] loadRound(op, subg);
- [ ] loadPow(op, subg);
- [ ] loadLogicalNot(op, subg);
- [ ] loadLogicalOr(op, subg);
- [ ] loadFill(op, subg);
- [ ] loadZerosLike(op, subg);
- [ ] loadTile(op, subg);
- [ ] loadRange(op, subg);
- [ ] loadBatchMatMul(op, subg);
- [ ] loadLogSoftmax(op, subg);
- [ ] loadQuantize(op, subg);
- [ ] loadSpaceToDepth(op, subg);

## Howto?

TBD

glistening:
@wateret Alternatively, we may add nnpkgtc by creating models for each uncovered operators. It has several advantages than GenModel (at least for me):

i) models could be used to measure memory and performance by it is nnpackage.
ii) input and expected value are easy to create. Just run `tflite2nnpkgtc.sh`.

I usually create `tflite` by a) using `flatc` with hand-created `json` or b) splitting an op from existing model.

s-barannikov:
> It has several advantages than GenModel

There is one disadvantage:
- if a test fails, it does not necessarily mean that the problem is in the loader, it can be anywhere in the runtime pipeline.

If the intention is to just improve coverage, then it's OK I guess :)

hyunsik-yoon:
Interesting this is
I think we could write `CircleGen::save(filename)` method, which can write a circle file. 
[...]

PR #3932, opened 2020-08-21
[onert] Fix conv2d op for dilation

YiHyunJin:
When using MultiThread, both dilation data must be 1.
When obtaining padding data, dilation data is used.

### DIFF
--- a/compute/cker/include/cker/operation/Conv.h
+++ b/compute/cker/include/cker/operation/Conv.h
@@ -61,11 +61,13 @@ public:
   }
 
   void prepare(const Shape &filter_shape, const float *filter_data, PaddingType padding_type,
-               bool &is_replaced_weights)
+               bool &is_replaced_weights, uint32_t dilationWidthFactor,
+               uint32_t dilationHeightFactor)
   {
     if (!_prepared)
     {
-      if (usableMultiThreaded(padding_type))
+      if (usableMultiThreaded(padding_type) && dilationWidthFactor == 1 &&
+          dilationHeightFactor == 1)
       {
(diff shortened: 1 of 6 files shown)


compute/cker/include/cker/operation/Conv.h
```diff
@@ -61,11 +61,13 @@ class Conv
...
   void prepare(const Shape &filter_shape, const float *filter_data, PaddingType padding_type,
-               bool &is_replaced_weights)
+               bool &is_replaced_weights, uint32_t dilationWidthFactor,
+               uint32_t dilationHeightFactor)
   {
     if (!_prepared)
     {
-      if (usableMultiThreaded(padding_type))
+      if (usableMultiThreaded(padding_type) && dilationWidthFactor == 1 &&
+          dilationHeightFactor == 1)
```
glistening, line 70: @YiHyunJin Could you please move additional conditions of dilation into `usableMultiThreaded`?

glistening (approved):
LGTM. (optional) It would be good if we can provide test case for this case (= conv2d with `same` padding and non-one dilation).

hseok-oh:
@YiHyunJin @glistening You can add test in `tests/nnfw_api/src/one_op_tests` later.

glistening:
@hseok-oh As I wrote in https://github.com/Samsung/ONE/issues/3764#issuecomment-672364693, alternatively we may use nnpkgtc [conv2d_dilation.zip](https://github.com/Samsung/ONE/files/5116245/conv2d_dilation.zip), which @YiHyunJin and me are used to test this op. nnpkgtc was created using `select_operator.py`.

PR #3935, opened 2020-08-23
[luci] Shape inf reduce LOC A-B

seanshpark:
This will relocate implementation of shape inference to reduce class LOC


seanshpark:
For #3881 from draft #3905

mhs4670go (approved):
LGTM

Issue #3565, opened 2020-07-30
Compiler FE: Batch Norm folding in Transposed Conv2D


mhs4670go:
BN(Batch Normalization) is a popular method used in neural network as it often reduces training time and improves generalization to some extent.

During inference, BN uses approximated per-channel mean and variance, which makes compilers have it merged with the preceding convolution layer. It saves computational resources and simplifies the network. 

Actually, fusion of Convolution and BN is already done by the compiler. Going further, I want to **see if it is possible to fuse with TCONV**(Transposed convolution) which looks similar to convolution.

### TODO
- [x] Mathematically check if it is possible
- [x] Check at the code level
- [x] Write a small network including TCONV + BN(mul+add)
- [x] Write a optimization pass : (TCONV+mul fusion) + add
- [x] Value test with ~~runtime(NYI)~~ luci-interpreter
- [x] Create TCONV IR having bias attribute
- [x] Update a optimization pass : TCONV+mul+add fusion

mhs4670go:
## Check at the code level
```python
import tensorflow as tf
import numpy as np

tf.compat.v1.disable_eager_execution()

def batch_norm(X, scale, offset):
    mean = np.mean(X)
    std = np.std(X)
    return (X - mean) / std * scale + offset

X = np.arange(1,5).reshape((1,2,2,1))
W = np.ones(9).reshape((3,3,1,1))
input_ = tf.compat.v1.constant(X, dtype=tf.float32)
filter_ = tf.compat.v1.constant(W, dtype=tf.float32)
tconv_ = tf.compat.v1.nn.conv2d_transpose(input_, filter_, output_shape=(1,4,4,1), strides=[1,1,1,1], padding='VALID')

with tf.compat.v1.Session() as sess:
    tconv_out = sess.run(tconv_)

scale = 1.05
offset = 0.015
# BN with numpy
print('-' * 10, 'numpy batch_norm', '-' * 10)
print(batch_norm(tconv_out, scale, offset))

# BN with tensorflow
print('-' * 10, 'tensorflow batch_norm', '-' * 10)
scale_ = tf.compat.v1.constant([1.05], dtype=tf.float32)
offset_ = tf.compat.v1.constant([0.015], dtype=tf.float32)
mean_ = tf.compat.v1.constant([np.mean(tconv_out)], dtype=tf.float32)
variance_ = tf.compat.v1.constant([np.var(tconv_out)], dtype=tf.float32)
[...]

PR #3940, opened 2020-08-24
[luci-interpreter] Add assert condition of CircleTransposeConv

mhs4670go:
This commit adds assert condition of CircleTransposeConv.

Bias attribute will be introduced.

Related : #3565 


s-barannikov (approved):
LGTM
I suppose this is a temporary change?

mhs4670go:
@s-barannikov Right. `node->arity() == 3` condition will be removed after bias attribute comes in Transposed Conv2D IR.

PR #3951, opened 2020-08-24
[onert] Set dyn_tensor_manager for static tensors

wateret:
Set `dynamic_tensor_manager` for static tensors in controlflow backend.
This is needed as any static tensors could turn into dynamic.

% Same approach has been applied for cpu backend, but not for controflow
yet.


wateret:
@hyunsik-yoon This will fix the problem you mentioned.

hyunsik-yoon (approved):
LGTM Thanks!

PR #3953, opened 2020-08-24
[nnfwapi] Simple WHILE test

wateret:
Add simple WHILE tests. As WHILE alone cannot make a meaningful test,
Add and Less operator are also used.

Introduce multi-subgraph support for `CircleGen`.


hseok-oh (approved):
LGTM

PR #3955, opened 2020-08-24
[luci] Shape inf common for Pad

seanshpark:
This will provide common methods for MirrorPad, Pad and PadV2 nodes to reduce duplicate codes


seanshpark:
For #3881 from draft #3905

mhs4670go (approved):
LGTM

PR #3967, opened 2020-08-25
[tests] exclude TransposeConv test from nnpkg test of acl_cl, acl_neon

mhs4670go:
This commit excludes TransposeConv test from nnpkg test of ac_cl,
acl_neon.

Since runtime doesn't support optional input, it will be excluded until
runtime support it.

Related : #3870 


hseok-oh (approved):
LGTM

PR #3971, opened 2020-08-25
[luci] Introduce FuseBatchNormWithTConv pass

mhs4670go:
This commit introduces FuseBatchNormWithTConv pass to luci.

Draft : #3660
Related : #3565

### DIFF
--- a/compiler/luci/pass/src/FuseBatchNormWithTConv.cpp
+++ b/compiler/luci/pass/src/FuseBatchNormWithTConv.cpp
@@ -1,0 +1,159 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


compiler/luci/pass/src/FuseBatchNormWithTConv.cpp
```diff
@@ -0,0 +1,159 @@
...
+bool FuseBatchNormWithTConvPass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto tconv = dynamic_cast<luci::CircleTransposeConv *>(node);
+    if (not tconv)
+      continue;
+
+    changed |= fused_batch_norm_with_tconv(tconv);
```
seanshpark, line 153: if once `changed` is true, I'm not sure `fused_batch_norm_with_tconv(tconv);` will be called when optimized.
for safety, 
```cxx
if (fused_batch_norm_with_tconv(tconv))
  changed = true;
```

mhs4670go: @seanshpark Maybe you're talking about short-circuit. Since `|=` is bit operation rather than logical operation, it won't short-circuit.

seanshpark: Have to tested?

mhs4670go: Maybe I think you mean "Shouldn't it be tested?". In the next PR, circle dredd test will be posted. But for testing this, I'll post a PR with another dredd test that tests the network that has more than two TCONV.

BTW, even if it did short-circuit, it would run well because `PhaseRunner` runs it until there's no pass that make a change.

seanshpark (approved):
LGTM

compiler/luci/pass/src/FuseBatchNormWithTConv.cpp
seanshpark: I ment test for `|=` :) 
As our mighty `PhaseRunner` will run until there is no more fusion, I think it's OK.

Issue #3831, opened 2020-08-14
Compiler FE: Support optional and mutually exclusive inputs in one-cmds


mhs4670go:
As of now, one-cmds support simple argument parsing. But it need to be upgraded.

For example, current `one-import-tf` has below interface.
```shell
Usage: one-import-tf
    --version Show version information and exit
    --input_path <path/to/tfmodel>
    --output_path <path/to/circle>
    --input_arrays <names of the input arrays, comma-separated>
    --input_shapes <input shapes, colon-separated>
    --output_arrays <names of the output arrays, comma-separated>
    --v2 Use TensorFlow 2.x interface (default is 1.x interface)
```
Although `input_shapes` works only without `--v2`, it is still not properly separated.

So, it would be better if we can support optional input or mutually exclusive group.

PR #3978, opened 2020-08-25
[one-cmds] Revise one-import-tf

mhs4670go:
This commit revises one-import-tf.

- `--input_shapes` and `--v2` becomes exclusive
- `--input_shapes` becomes optional

Related : #3831

### DIFF
--- a/compiler/one-cmds/one-import-tf
+++ b/compiler/one-cmds/one-import-tf
@@ -84,10 +84,7 @@ while [ "$#" -ne 0 ]; do
 done
 
 if [ -n ${INPUT_SHAPES} ] && [ ${TF_INTERFACE} = "--v2" ]; then
-  echo "Error: shape cannot be assigned to dynamic shape in v2 version."
-  echo ""
-  usage
-  exit 2
+  echo "Warning: if --v2 option is used, shape will be ignored"
 fi
 
 if [ -z ${INPUT_PATH} ] || [ ! -e ${INPUT_PATH} ]; then
--- a/compiler/one-cmds/one-import-tf
+++ b/compiler/one-cmds/one-import-tf
@@ -83,6 +83,13 @@ while [ "$#" -ne 0 ]; do
   esac
 done
 
+if [ -n ${INPUT_SHAPES} ] && [ ${TF_INTERFACE} = "--v2" ]; then
+  echo "Error: shape cannot be assigned to dynamic shape in v2 version."
+  echo ""
+  usage
+  exit 2
+fi
+
 if [ -z ${INPUT_PATH} ] || [ ! -e ${INPUT_PATH} ]; then
   echo "Error: input model not found"
   echo ""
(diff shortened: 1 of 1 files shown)


compiler/one-cmds/one-import-tf
```diff
@@ -83,6 +83,13 @@ while [ "$#" -ne 0 ]; do
   esac
 done
 
+if [ -n ${INPUT_SHAPES} ] && [ ${TF_INTERFACE} = "--v2" ]; then
+  echo "Error: shape cannot be assigned to dynamic shape in v2 version."
+  echo ""
+  usage
+  exit 2
```
seanshpark, line 90: I think it would be better to show a warning message and not end the program for backward compatibility.
- message something like, `If --v2 option is used, shape will be ignored`

compiler/one-cmds/one-import-tf
```diff
@@ -117,16 +124,18 @@ show_err_onexit()
...
+CONVERT_SCRIPT+="--output_arrays ${OUTPUT_ARRAYS} "
+if [ ! -z ${INPUT_SHAPES} ]; then
+  CONVERT_SCRIPT+="--input_shapes ${INPUT_SHAPES} "
+fi
 
-python "${DRIVER_PATH}/tf2tfliteV2.py" ${TF_INTERFACE} --input_path ${INPUT_PATH} \
---input_arrays ${INPUT_ARRAYS} --input_shapes ${INPUT_SHAPES} \
---output_path "${TMPDIR}/${MODEL_NAME}.tflite" \
---output_arrays ${OUTPUT_ARRAYS} >> "${OUTPUT_PATH}.log" 2>&1
+echo ${CONVERT_SCRIPT} > "${OUTPUT_PATH}.log"
```
seanshpark, line 136: this is nice :)

compiler/one-cmds/one-import-tf
```diff
@@ -117,16 +121,18 @@ show_err_onexit()
...
---input_arrays ${INPUT_ARRAYS} --input_shapes ${INPUT_SHAPES} \
---output_path "${TMPDIR}/${MODEL_NAME}.tflite" \
---output_arrays ${OUTPUT_ARRAYS} > "${OUTPUT_PATH}.log"
-echo " " >> "${OUTPUT_PATH}.log"
+CONVERT_SCRIPT="python ${DRIVER_PATH}/tf2tfliteV2.py ${TF_INTERFACE} "
+CONVERT_SCRIPT+="--input_path ${INPUT_PATH} "
+CONVERT_SCRIPT+="--input_arrays ${INPUT_ARRAYS} "
+CONVERT_SCRIPT+="--output_path ${TMPDIR}/${MODEL_NAME}.tflite "
+CONVERT_SCRIPT+="--output_arrays ${OUTPUT_ARRAYS} "
+if [ ! -z ${INPUT_SHAPES} ]; then
```
seanshpark, line 129: maybe check `TF_INTERFACE` is not `--v2` ?

seanshpark (approved):
LGTM

compiler/one-cmds/one-import-tf
seanshpark: We can go as-is as `tf2tfliteV2.py` will ignore :)

mhs4670go:
merge plz:)

PR #3982, opened 2020-08-25
[onert/test] Random uint8 input value generation for tflite_run

hseok-oh:
tflite_run set random value for uint8 input tensor

### DIFF
--- a/tests/tools/tflite_run/src/tflite_run.cc
+++ b/tests/tools/tflite_run/src/tflite_run.cc
@@ -220,13 +220,16 @@ int main(const int argc, char **argv)
         // Generate unsigned 8-bit integer input
         auto tensor_view = nnfw::tflite::TensorView<uint8_t>::make(*interpreter, o);
 
-        uint8_t value = 0;
+        auto fp = static_cast<uint8_t (nnfw::misc::RandomGenerator::*)(
+            const ::nnfw::misc::tensor::Shape &, const ::nnfw::misc::tensor::Index &)>(
+            &nnfw::misc::RandomGenerator::generate<uint8_t>);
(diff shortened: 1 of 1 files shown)


tests/tools/tflite_run/src/tflite_run.cc
```diff
@@ -220,13 +220,16 @@ int main(const int argc, char **argv)
...
+            const ::nnfw::misc::tensor::Shape &, const ::nnfw::misc::tensor::Index &)>(
+            &nnfw::misc::RandomGenerator::generate<uint8_t>);
+        const nnfw::misc::tensor::Object<uint8_t> data(tensor_view.shape(),
+                                                       std::bind(fp, randgen, _1, _2));
 
         nnfw::misc::tensor::iterate(tensor_view.shape())
             << [&](const nnfw::misc::tensor::Index &ind) {
-                 // TODO Generate random values
+                 const auto value = data.at(ind);
                  tensor_view.at(ind) = value;
```
glistening, line 232: @hseok-oh I spent about 1 hour to understand the code of this PR.

After studying, I've found that

- static_cast is here to disambiguate `generate()` and `generate(shape, index)` member function pointer
- use `std::bind()` to generate generator for `nnfw::misc::tensor::Object`.

However, we don't use shape and index parameters to generate random value.

I would write

```c++
tensor_view.at(ind) = value;
```

to

```c++
tensor_view.at(ind) = randgen.generate<uint8_t>();
```

glistening (approved):
LGTM

tests/tools/tflite_run/src/tflite_run.cc
hseok-oh: I copied this code from `runtime/libs/tflite/src/RandomTestRunner.cpp` :)

PR #3991, opened 2020-08-25
[onert] Change Permutation Elimination condition

wateret:
Change the condition of Permutation Elimination to be more general.

### DIFF
--- a/runtime/onert/core/src/compiler/pass/PermutationEliminationPass.cc
+++ b/runtime/onert/core/src/compiler/pass/PermutationEliminationPass.cc
@@ -37,19 +37,17 @@ void PermutationEliminationPass::visit(const ir::operation::Permute &node)
   auto in_operand = node.getInputs().at(0);
   auto out_operand = node.getOutputs().at(0);
 
-  // Check if two tensors are both portable
-  // TODO Make this general, this is just a workaround to check two tensors are portable
+  // Check if two tensors are both portable if not, we can't eliminate the node
   {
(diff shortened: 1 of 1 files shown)


wateret:
I manually tested that this worked fine with another cpu-style backend.

runtime/onert/core/src/compiler/pass/PermutationEliminationPass.cc
```diff
@@ -37,19 +37,17 @@ void PermutationEliminationPass::visit(const ir::operation::Permute &node)
...
-    auto out_backend_id = out_def_factor.backend()->config()->id();
+    auto in_config = in_def_factor.backend()->config();
+    auto out_config = out_def_factor.backend()->config();
 
-    // TODO Fix this workaround that removes only Permute between cpu and controlflow backend.
-    //      This should be general.
-    if (!((in_backend_id == backend::controlflow::Config::ID && out_backend_id == "cpu") ||
-          (in_backend_id == "cpu" && out_backend_id == backend::controlflow::Config::ID)))
+    // FIXME Supporting dynamic tensor does not exactly mean those are portable.
+    //       It may need to have another config option for checking if each uses `IPortableTensor`.
```
wateret, line 49: @ragmani Earlier I told you that a config value for this would not be necessary. However it turns out that it is not that easy in the current structure. 😢

ragmani: It doesn't matter much in my opinion. 😃

ragmani (approved):
LGTM

PR #3994, opened 2020-08-25
[souschef] common as_vector as_dataset

seanshpark:
This will introduce common as_vector as_dataset methods from circlechef and tflchef


seanshpark:
For #3881 from draft #3905

mhs4670go (approved):
LGTM

PR #3995, opened 2020-08-25
[souschef] Introduce Dims

seanshpark:
This will introduce Dims related helpers, common of circlechef and tflchef

### DIFF
--- a/compiler/souschef/src/Dims.cpp
+++ b/compiler/souschef/src/Dims.cpp
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


seanshpark:
For #3881 from draft #3905

compiler/souschef/src/Dims.cpp
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "souschef/Dims.h"
+
+// NOTE Do NOT delete this file; this file checks the completeness of 'Dims.h'
```
mhs4670go, line 19: Could you explain about this more detail? I mean what the completeness you said is.

seanshpark: Like other header only files, (you can search `compiler` folder with this string), this cpp will include the header `Dims.h` and compile.
If not, `Dims.h` won't be compiled and CI won't tell if there is compilation error.

mhs4670go (approved):
LGTM

PR #3996, opened 2020-08-26
[luci] Clear quantparam before fake quantization

jinevening:
This clears quantparam before doing fake quantization


jinevening:
This code clears all quantparams of a circle model at the first stage of quantization (--quantize_dequantize_weights).

jinevening:
@seanshpark I've updated the code to explicitly use `nullptr` (and it works well). Thanks!

seanshpark (approved):
LGTM

Issue #1635, opened 2020-05-29
[runtime] Tracking static analysis warning


hseok-oh:
This issue is to track internal static analysis warning on runtime build

hseok-oh:
Analysis result at 6c69521 

- `tests/nnfw_api/src/RegressionTests.cc`: resource leak

hseok-oh:
Analysis result at 1bf9f07

- Exception without catch
  - `tlite_loader.cc:218` `ANeuralNetworksExecution.cc:101`
- Resource leak
  - `RegressionTEsts.cc:26` `RegressionTEsts.cc:33`

hseok-oh:
Analysis result at 9f71611646200181ad9f2ec672f4b34088412e5c
- svace: 1 critical, 1 major
  - `FullyConnectedLayer.cc`
  - `ShapeInference.cc`
- svace security: no warning
- coverity: 2 critical
  - `Data.h`
  - `FullyConnectedLayer.cc`

hseok-oh:
Analysis result at d96ab686207a5a76fd72ae62c531d8c50b4d84a5: 0 criticals on svace, svace secutiry, coverity

hseok-oh:
Analysis result at fe5ab2be0a10d5334f51bf7551eca9f43eb09c63 
- 1 major warning on svace
  - `compute/cker/include/cker/operation/Helper/PhiloxRandom.h`:264  INTEGER_OVERFLOW
   - Mark as `intentional`
- 1 major warning on coverity
  - `compute/cker/include/cker/operation/SplitV.h`:46 Useless call
- No warning on svace security

hseok-oh:
Analysis result at 198e4f622d13703428bb00c82862edd8bd64192b

- 7 critical warnings and 1 major warning  on coverity
  - `jsoncpp.cpp` (4 criticals): ignore 
  - `OperationFactory.cc` (2 criticals): param initialize
  - acl-ex `CLReduceOperation.cpp` (2 criticals): 
  - `cpu backend `OpeationUtils.h` (1 major):
- No warning on svace and svace security

chunseoklee:
For 1.9.0 release, we should resolve 12 major defect on latest master : 

https://code.sec.samsung.net/ahub/service/analyses/175944

glistening:
Some defects in `View.h` caused by #3987.

chunseoklee:
@ragmani PTAL https://code.sec.samsung.net/ahub/service/analyses/175944

[...]

PR #3997, opened 2020-08-26
[onert/nnapi] Handle param initialize warning

hseok-oh:
- Throw exception when CONV2D input count is unsupported
- Initialize ARGMAX output type param: always INT32 type by NNAPI specification

---

Related issue: https://github.com/Samsung/ONE/issues/1635#issuecomment-680395192

### DIFF
--- a/runtime/onert/frontend/nnapi/wrapper/OperationFactory.cc
+++ b/runtime/onert/frontend/nnapi/wrapper/OperationFactory.cc
@@ -630,6 +630,10 @@ OperationFactory::OperationFactory()
       param.activation =
           NNAPIConvert::getFusedActivation(operands.at(activation_index).asScalar<FuseCode>());
     }
+    else
+    {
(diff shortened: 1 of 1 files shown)


lemmaa (approved):
LGTM.

runtime/onert/frontend/nnapi/wrapper/OperationFactory.cc
```diff
@@ -1415,6 +1419,8 @@ OperationFactory::OperationFactory()
 
     operation::ArgMax::Param param;
     param.axis = operands.at(OperandIndex{init_param.inputs[1]}).asScalar<std::int32_t>();
+    // NNAPI ARGMAX output type is always int32
+    param.output_type = DataType::INT32;
```
lemmaa, line 1423: It seems we should check this kind of specification handling for all other NNAPI operators as well. :)

PR #3965, opened 2020-08-25
[onert] Make nnpackage_run read shape info from input h5 file

hyunsik-yoon:
Parent issue: #3840

This adds `h5` into `--shape_prepare` and `--shape_run` param, which reads shape from h5 file.

For example,
```
$ ./Product/out/bin/nnpackage_run --nnpackage ${NNPACKAGE} --load ${H5_INPUT_FILE} \
  --shape_run h5 ...
```

Of course, explicit shape also work:

```
$ ./Product/out/bin/nnpackage_run --nnpackage ${NNPACKAGE} --load ${H5_INPUT_FILE} \
  --shape_run "[0,[10,1]]" ...
```

Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>

PR #3998, opened 2020-08-26
[onert] Do nothing when 'passed shape of nnfw_set_tensorinfo() == model shape' 

hyunsik-yoon:
This makes `nnfw_set_input_tensorinfo()` do nothing when 'passed shape == model shape'.

related: https://github.com/Samsung/ONE/pull/3965#issuecomment-679777002

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -481,26 +481,27 @@ NNFW_STATUS nnfw_session::apply_tensorinfo(uint32_t index, nnfw_tensorinfo ti)
     }
   }
 
+  auto ind = primary_subgraph()->getInputs().at(index);
+  auto &input = primary_subgraph()->operands().at(ind);
+
+  onert::ir::Shape new_shape(ti.rank);
+  for (int32_t i = 0; i < ti.rank; i++)
+    new_shape.dim(i) = ti.dims[i];
+
+  // if passed shape is same with the shape of model, do nothing
+  if (input.info().shape() == new_shape)
+    return NNFW_STATUS_NO_ERROR;
+
   if (!isStatePreparedOrFinishedRun())
   {
     // In this case, if we apply input shape in primary_subgraph, it will propagate after
     // compilation and excution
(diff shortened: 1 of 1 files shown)


glistening:
@hyunsik-yoon 

> but in this case setting shape before run makes runtime to infer shape during running. If static shape of input is same shape set with set_input(), it is better for runtime not to work in dynamic way. I will work on this.

Could you please explain how this PR get runtime not to work in dynamic way when new shape == model shape?

wateret (approved):

hyunsik-yoon:
@glistening Previously, the following line always called, which make model input tensor `dynamic`. So the op using model input tensor sets their output tensor dynamic, and so on.
```
    _execution->changeInputShape(onert::ir::IOIndex(index), new_shape);
```
With this PR, the above line is not called when newly set shape is same with model input. Therefore, model input tensor stays `static`.

ragmani (approved):

glistening:
@hyunsik-yoon Yesterday, I've also guessed `changeInputShape` would make the input tensor `dynamic`. However, I thought `changeInputShape` just set `input_shape_signature`.

```c++
void Execution::changeInputShape(const ir::IOIndex &index, const ir::Shape &new_shape)
{
  // This should be called BEFORE setInput.
  if (_io_desc.inputs.at(index.value()) != 0)
    throw std::runtime_error("Error in calling order");

  _io_desc.input_shape_signature[index] = new_shape;
}
```

Today, I investigated more, then I've found how it is related to `dynamic`.

```c++
auto input_shape = desc.input_shape_signature.find(ir::IOIndex{i});
if (input_shape != desc.input_shape_signature.end())
{
  tensor->set_dynamic();
  tensor->setShape(input_shape->second);
}
```
```c++
struct IODescription
{
  std::vector<std::unique_ptr<InputDesc>> inputs;
  std::vector<std::unique_ptr<OutputDesc>> outputs;
  // Contains shape of input set by set_input_tensorinfo
  std::unordered_map<ir::IOIndex, ir::Shape> input_shape_signature;
};
```

`input_shape_signature` is used to contain the input shape from `set_input_tensorinfo`.

PR #4003, opened 2020-08-27
[acl-ex] Check reduce axis set size before use

hseok-oh:
Check reduced axis set has at least one element

---

Related issue: https://github.com/Samsung/ONE/issues/1635#issuecomment-680395192

### DIFF
--- a/compute/ARMComputeEx/src/runtime/CL/functions/CLReduceOperation.cpp
+++ b/compute/ARMComputeEx/src/runtime/CL/functions/CLReduceOperation.cpp
@@ -59,6 +59,8 @@ Status CLReduceOperation::validate(const ITensorInfo *input, const ITensorInfo *
   const size_t num_of_kernels = axis.size();
   const size_t num_of_interm_tensors = num_of_kernels - (keep_dims ? 1 : 0);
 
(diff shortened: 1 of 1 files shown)


compute/ARMComputeEx/src/runtime/CL/functions/CLReduceOperation.cpp
```diff
@@ -59,6 +59,8 @@ Status CLReduceOperation::validate(const ITensorInfo *input, const ITensorInfo *
   const size_t num_of_kernels = axis.size();
   const size_t num_of_interm_tensors = num_of_kernels - (keep_dims ? 1 : 0);
 
+  ARM_COMPUTE_RETURN_ERROR_ON(num_of_kernels < 1);
```
glistening, line 62: @hseok-oh is `axis.size() == 0` not allowed?

hseok-oh: `axis.size() == 0` means there is no axis to reduce. And it makes invalid result to calculate `num_of_interm_tensors` and invalid access to `interm_tensors` array.

glistening (approved):

ragmani (approved):
LGTM

Issue #3625, opened 2020-08-03
[onert] Unify similar operations into one


wateret:
Just like `Reduce` operation, what about unifying operations that basically do the same thing?
e.g.) `Add` and `Sub` - have same params and same Input/Output semantics

For example, we already have done that for `Reduce` ops.

```cpp
class Reduce : public Operation
{
public:
  // ...
  enum class ReduceType
  {
    ALL,
    ANY,
    MAX,
    MEAN,
    MIN,
    PROD,
    SUM
  };

  struct Param
  {
    ReduceType reduce_type;
    bool keep_dims;
  };

```

### WORK LIST

- [x] Binary arithmetic (Has activation)
    - Add
    - Sub
    - Mul
    - Div
- [x] Binary elementwise
    - LogicalOr
    - LogicalAnd
    - Max
    - Min
- [x] Pool2D
    - MaxPool2D
    - AvgPool2D
    - L2Pool2D

- [x] Activation elementwise
    - Elu
    - Logistic
    - Relu
    - Tanh

- [x] Unary elementwise
    - Abs
    - Cast
    - Cos
    - Dequantize
    - Exp
    - Floor
    - Log
    - LogicalNot
    - Neg
    - Quantize
    - Round
    - RSqrt
    - Sin
    - Sqrt
    - Square
    - ZerosLike

### Candidates (not strictly analyzed, just roughly grouped)

....

wateret:
@ragmani 👍 for volunteering!

wateret:
I'm expecting effects of this work

- Removing duplicate code makes code more readable
- Better SAM score : DC(Duplicated Code), CBO and LOC at the same time

PR #4016, opened 2020-08-27
[onert] Remove duplicated codes in base_loader

ragmani:
For issue #3625

This commit removes duplicated codes in base_loader.
  - Remove a template parameter from class BaseLoader
  - Unify loading operation to using loadOperationTo()

### DIFF
--- a/runtime/onert/frontend/base_loader/include/base_loader.h
+++ b/runtime/onert/frontend/base_loader/include/base_loader.h
@@ -110,6 +110,8 @@ protected:
 private:
   virtual std::unique_ptr<ir::Graph> loadSubgraph(const SubGraph *subg) = 0;
   // Operations
+  template <typename OpIR, typename... Args>
+  const OpIR *loadOperationTo(const Operator *op, ir::Graph &subg, Args &&... args);
   void loadConv2D(const Operator *op, ir::Graph &subg);
   void loadDepthwiseConv2D(const Operator *op, ir::Graph &subg);
   void loadTransposeConv(const Operator *op, ir::Graph &subg);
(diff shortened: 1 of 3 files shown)


wateret (dismissed):
LGTM

wateret:
It may be nice to request review to the original author of base_loader(I don't know who 😅 ) too.

ragmani:
> It may be nice to request review to the original author of base_loader(I don't know who sweat_smile ) too.

@s-barannikov Could you review this PR?

wateret (approved):
LGTM

s-barannikov (commented):
It seems that the PR mixes refactoring (e.g. inlining load* methods and introducing `loadOprationTo`) and changing logic (e.g. removing template parameter and introducing virtual method).
It would be much easier to review if the refactoring part is merged separately, or at least in a separate commit(s). Is it possible to split the changes? It would be really helpful.

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -71,6 +71,8 @@ template <typename LoaderDomain, typename SpecificLoader> class BaseLoader
   {
   }
 
+  virtual std::unique_ptr<ir::Graph> loadSubgraph(const SubGraph *subg) = 0;
```
s-barannikov, line 74: (historical note)
It was intended to use static polymorphism (via SpecificLoader) instead of the dynamic one (via virtual methods).
The workflow suggested was as follows:
- Initially, all the code is contained in BaseLoader.
- When Circle implementation of a specific concept begins to differ from the TFLite counterpart (as happened with SubGraphs and some Circle-specific operators), the related code is completely removed from the base loader and implemented in both derived loaders.
- When there become too many differences between the two loaders, the base loader can be removed, leaving two completely independent loaders (Circle and TFLite).
---
I must admit that the intention is not clear from the code, so the implementation of the idea might not be the best.

Virtual methods can also help to achieve the same goal. Applying virtual methods for such purposes is known as "template method pattern" (although it has nothing to do with C++ templates, it's just a name clash). It is well known pattern, so the code using it should be more clear to developers than the former approach using static polymorphism.

Please note that the virtual function should be `private` in this case (in both base and derived classes). More on this [here](http://www.gotw.ca/publications/mill18.htm).

PS
It is worth noting that each virtual method added to BaseLoader is actually two methods (one for each loader). This is because TFLiteLoader and CircleLoader derive from different base classes, thanks to template parameter of BaseLoader.

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -106,67 +108,31 @@ template <typename LoaderDomain, typename SpecificLoader> class BaseLoader
...
-  void loadStridedSlice(const Operator *op, ir::Graph &subg);
-  void loadUnpack(const Operator *op, ir::Graph &subg);
   void loadComparison(const Operator *op, ir::Graph &subg);
-  void loadEinsum(const Operator *op, ir::Graph &subg);
-  void loadOneHot(const Operator *op, ir::Graph &subg);
-  void loadShape(const Operator *op, ir::Graph &subg);
-  void loadIf(const Operator *op, ir::Graph &subg);
-  void loadWhile(const Operator *op, ir::Graph &subg);
+
+  // FIXME Remove these methods
```
s-barannikov, line 131: I think it is ok to have some methods to keep the `switch` smaller.

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -1804,89 +1250,113 @@ void BaseLoader<LoaderDomain, SpecificLoader>::loadOperation(const Operator *op,
...
+
+      loadOperationTo<ir::operation::OneHot>(op, subg, param);
       return;
+    }
     case BuiltinOperator::BuiltinOperator_SHAPE:
-      loadShape(op, subg);
-      return;
-    case BuiltinOperator::BuiltinOperator_REDUCE_PROD:
-      loadReduce<ir::operation::Reduce::ReduceType::PROD>(op, subg);
+      // ir::operation::Shape::Param param;
```
s-barannikov, line 1264: Is this commented out intentionally?

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -1558,135 +902,151 @@ void BaseLoader<LoaderDomain, SpecificLoader>::loadArgMax(const Operator *op, ir
...
+      param.beta = options->beta();
+      loadOperationTo<ir::operation::Softmax>(op, subg, param);
       return;
+    }
     case BuiltinOperator::BuiltinOperator_CONCATENATION:
-      loadConcatenation(op, subg);
+    {
+      ir::operation::Concat::Param param;
+      const auto *options = op->builtin_options_as_ConcatenationOptions();
+      // Axis
```
s-barannikov, line 1001: I understand that this is cut & paste, but as it is a refactoring, I would suggest to remove such obvious comments.

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -1558,135 +902,151 @@ void BaseLoader<LoaderDomain, SpecificLoader>::loadArgMax(const Operator *op, ir
   subg.addOperation(std::move(new_op));
 }
 
-template <typename LoaderDomain, typename SpecificLoader>
-void BaseLoader<LoaderDomain, SpecificLoader>::loadPow(const Operator *op, ir::Graph &subg)
+template <typename LoaderDomain>
+template <typename OpIR, typename... Args>
+const OpIR *BaseLoader<LoaderDomain>::loadOperationTo(const Operator *op, ir::Graph &subg,
```
s-barannikov, line 907: This is a great helper that reduces amount of boilerplate significantly. Why don't submit it (and its uses) first? This would make the reviewing much easier...

s-barannikov:
Looks good (aside from minor comments), but please try to split this into smaller chunks.

glistening:
+1 for https://github.com/Samsung/ONE/pull/4016#pullrequestreview-476977366. I've experienced a regression from a big change like this PR, which prevent careful review.

runtime/onert/frontend/base_loader/include/base_loader.h
hseok-oh: I agree with @s-barannikov comment. I think `switch` must smaller and simple pattern as possible.

hseok-oh:
Some part agree, and some part disagree.

Agree
- Introduce `loadOperationTo()`
- Introduce virtual method and remove typename of template class `BaseLoader`

Disagree
- Inlining parameter parsing in switch-case. I prefer to use simple switch-case just calling one method, then call common method in operator specific method.

runtime/onert/frontend/base_loader/include/base_loader.h
ragmani: Okay I will remove it.

runtime/onert/frontend/base_loader/include/base_loader.h
ragmani: This comment already existed in the original code. I don't know why this didn't commented out in detail. But It seems unnecessary. I will remove it.

runtime/onert/frontend/base_loader/include/base_loader.h
ragmani: I think these operation's IR should be changed some params to input that can be tensor of backend. After it, the duplicated codes can be removed in this method. I will change this commend from removing these methods to remove duplicated codes in these methods.

ragmani:
> Inlining parameter parsing in switch-case. I prefer to use simple switch-case just calling one method, then call common method in operator specific method.

I see. I will keep some method for simple switch-case.

runtime/onert/frontend/base_loader/include/base_loader.h
ragmani: Thanks for your kindly explanation. I will move the virtual function to be `private`.

ragmani:
> It seems that the PR mixes refactoring (e.g. inlining load* methods and introducing loadOprationTo) and changing logic (e.g. removing template parameter and introducing virtual method).

Right. I will change this PR to the two separated commit within this PR.

runtime/onert/frontend/base_loader/include/base_loader.h
ragmani: I divided this PR to two separated commits.

runtime/onert/frontend/circle/src/circle_loader.cc
```diff
@@ -91,7 +91,7 @@ class CircleLoader final : public base_loader::BaseLoader<LoaderDomain, CircleLo
     }
   }
 
-  std::unique_ptr<ir::Graph> loadSubgraph(const circle::SubGraph *circle_subg)
+  std::unique_ptr<ir::Graph> loadSubgraph(const circle::SubGraph *circle_subg) override
```
s-barannikov, line 94: Should be private

runtime/onert/frontend/tflite/src/tflite_loader.cc
```diff
@@ -78,7 +78,7 @@ class TFLiteLoader final : public base_loader::BaseLoader<LoaderDomain, TFLiteLo
     }
   }
 
-  std::unique_ptr<ir::Graph> loadSubgraph(const onert_tflite::SubGraph *tflite_subg)
+  std::unique_ptr<ir::Graph> loadSubgraph(const onert_tflite::SubGraph *tflite_subg) override
```
s-barannikov, line 81: Same

s-barannikov (approved):
LGTM

runtime/onert/frontend/circle/src/circle_loader.cc
ragmani: Thanks for your pointing out. It's my mistake.

runtime/onert/frontend/tflite/src/tflite_loader.cc
ragmani: Thanks.

s-barannikov (approved):

hseok-oh (approved):

glistening (approved):
LGTM

ragmani:
PTAL @wateret

hseok-oh:
It's better to merge after branch out release 1.9.0, IMO.

ragmani:
> It's better to merge after branch out release 1.9.0, IMO.

@hseok-oh Okay

wateret:
Will merge once branch 1.9.0 is created.

hseok-oh:
Release 1.9.0 branch is created.
@wateret PTAL

PR #4017, opened 2020-08-27
[onert] CF KernelGen use IDynamicTensorManager

wateret:
Make controlfow KernelGen use `IDynamicTensorManager` rather than
`TensorBuilder`, as Dynamic Tensor Manager is what it really needs.


hyunsik-yoon (approved):
LGTM

Issue #3757, opened 2020-08-11
[nncc] Assign input's Shape in Circle<op>Out node instance


dr-venkman:
ShapeInference for `unique`'s output index needs to follow from it's input tensor, as per definition. From the draft pull request in #2829, this is expected to happen at the function `loco::NodeShape visit(const luci::CircleUniqueOut *node)  {.. } `.

However, I am not able to access the input shape as seamlessly as I was able to do so with `loco::NodeShape visit(const luci::CircleUnique *node)`, where `auto input_shape = loco::shape_get(node->input()).as<loco::TensorShape>();` fetched me the input shape without issues.

In `loco::NodeShape visit(const luci::CircleUniqueOut *node)  {.. } `, please refer to the following lines as reference:

```
    auto unique = dynamic_cast<const luci::CircleUnique *>(node->input());
    if (unique == nullptr)
    {
      INTERNAL_EXN("CircleUnique IR is not configured correctly");
    }

    auto unique_shape = loco::shape_get(unique).as<loco::TensorShape>();
```

Here, `unique_shape` is not what I was expecting, i.e. the input operand shape. My question in general is whether it is indeed possible to access the input shape from within the above function. Kindly let me know. Thank you in advance.

seanshpark:
All other nodes input shapre reference uses codes like
```
auto unique_shape = loco::shape_get(unique).as<loco::TensorShape>();
```

> not what I was expecting,

what do you exect and what is the value you get? how did you get that value?

dr-venkman:
> All other nodes input shapre reference uses codes like
> 
> ```
> auto unique_shape = loco::shape_get(unique).as<loco::TensorShape>();
> ```
> 
> > not what I was expecting,
> 
> what do you exect and what is the value you get? how did you get that value?

I was expecting to retreive the input shape and assign it's rank and dim[0] to an output shape, just as I had done previously inside `loco::NodeShape visit(const luci::CircleUnique *node)`. 

In this case, however, `unique_shape` does not have a well-defined `_dims` vector that i can access and assign values from. In the former case, `input_shape` had `rank() == 1`, and had the dimension values set within `_dims`.

seanshpark:
>  i can access and assign values from

where is `from` ?

seanshpark:
Can you be more specific to real numbers? I can't catch what exactly the problem is.

dr-venkman:
@seanshpark, I am currently testing with the following recipe:

```
operand {
  name: "ifm"
[...]

PR #4019, opened 2020-08-28
[TensorFlowLiteRecipes] fix unique op recipe

mhs4670go:
This commit fixes unique op recipe.

When I compare it with tfilte file generated from `res/TensorFlowPythonExamples`, there's no shape rather than zero dim.

### DIFF
--- a/res/TensorFlowLiteRecipes/Unique_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/Unique_000/test.recipe
@@ -6,7 +6,7 @@ operand {
 operand {
   name: "ofm"
   type: FLOAT32
-  shape { dim: 0 }
+  shape { }
 }
 operand {
   name: "ofm_idx"
--- a/res/TensorFlowLiteRecipes/Unique_001/test.recipe
+++ b/res/TensorFlowLiteRecipes/Unique_001/test.recipe
@@ -6,7 +6,7 @@ operand {
 operand {
   name: "ofm"
   type: FLOAT32
-  shape { dim: 0 }
+  shape { }
 }
 operand {
   name: "ofm_idx"
--- a/res/TensorFlowLiteRecipes/Unique_002/test.recipe
+++ b/res/TensorFlowLiteRecipes/Unique_002/test.recipe
@@ -6,7 +6,7 @@ operand {
 operand {
   name: "ofm"
   type: INT32
-  shape { dim: 0 }
+  shape { }
 }
 operand {
   name: "ofm_idx"
--- a/res/TensorFlowLiteRecipes/Unique_003/test.recipe
+++ b/res/TensorFlowLiteRecipes/Unique_003/test.recipe
@@ -6,7 +6,7 @@ operand {
 operand {
   name: "ofm"
   type: INT32
-  shape { dim: 0 }
+  shape { }
 }
 operand {
   name: "ofm_idx"
--- a/res/TensorFlowLiteRecipes/Unique_U8_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/Unique_U8_000/test.recipe
@@ -7,7 +7,7 @@ operand {
 operand {
   name: "ofm"
   type: UINT8
-  shape { dim: 0 }
+  shape { }
 }
 operand {
   name: "ofm_idx"
--- a/res/TensorFlowLiteRecipes/Unique_U8_001/test.recipe
+++ b/res/TensorFlowLiteRecipes/Unique_U8_001/test.recipe
@@ -7,7 +7,7 @@ operand {
 operand {
   name: "ofm"
   type: UINT8
-  shape { dim: 0 }
+  shape { }
 }
 operand {
   name: "ofm_idx"


seanshpark:
@mhs4670go , please refer #3757
With `shape { }` , test is OK and this confuses me compared with @dr-venkman 's https://github.com/Samsung/ONE/issues/3757#issuecomment-675377402 comment

mhs4670go:
@seanshpark Well, you think this pr has an error?
AFAIK, if there is no shape in flatbuffers file, it's gonna be imported as UNKNOWN shape, which will be exported as `own_shape`, which is unknown. I mean, I don't think there is an error.

seanshpark:
>  I mean, I don't think there is an error.

Not an error, if you look at #3757 , @dr-venkman  wrote `shape { }` has an error so we concluded to use ` shape { dim: 0 }`
But now,  `shape { }` is working OK. I also think `shape { }` should work but @dr-venkman wrote it didn't.
@dr-venkman, can you please describe more about what was the problem?

dr-venkman:
> > I mean, I don't think there is an error.
> 
> Not an error, if you look at #3757 , @dr-venkman wrote `shape { }` has an error so we concluded to use ` shape { dim: 0 }`
> But now, `shape { }` is working OK. I also think `shape { }` should work but @dr-venkman wrote it didn't.
> @dr-venkman, can you please describe more about what was the problem?

@seanshpark , @mhs4670go , I remember from the issue #3757 that unless `shape` was assigned a dimension (`dim: 0` in this case) `ofm` was assigned a rank 0, and was therefore treated as a scalar. As a result, the `UniqueOut` node for `ofm` was never visited, and therefore it's shape inferencing was not handled inside the `visit(const luci::CircleUniqueOut *node) method`. Assigning `dim: 0` forces a rank 1 assignment to `ofm` and makes it possible to visit the `UniqueOut` node.

I do not know what changed between then and now, but if the default rank is now a value other than 0, then there would be no need to assign `dim : 0`.

dr-venkman (commented):
if the default rank assignment for `ofm` is a value other then 0, then this change should be fine.

seanshpark:
Current shape inference for CircleUniqueOut codes branch by value of index: return shape is different.
As this change passes, there maybe some holes in our testing.
@mhs4670go, if it's OK for you, I'd like to do some tracing.

mhs4670go:
I think the result of this PR - pass the test - seems right.
First, if there is no shape in recipe, it would import as NOSHAPE.
```cxx
if (tensors_ptr->Get(outputs[n])->shape() == nullptr)
      nodeout->shape_status(ShapeStatus::NOSHAPE);
```
Then, when the shape inference runs, its shape is assigned as its own shape.
```cxx
bool is_shape_none = (circle_node->shape_status() == ShapeStatus::NOSHAPE);
  bool is_scalar = (circle_node->rank() == 0);

  if (is_shape_undefined)
    shape = circle_node->accept(&alg);
  else
  {
    if (is_shape_none || is_scalar)
      shape = own_shape(circle_node); // here
    else
      shape = circle_node->accept(&alg);
  }
```

So, there is no place that causes error.

seanshpark:
> ofm was never visited, and therefore it's shape inferencing was not handled inside the visit(const luci::CircleUniqueOut *node) method

For index 0, `ofm`, result shape was set by the line @mhs4670go pointed 
```cxx
      shape = own_shape(circle_node); // here
``` 
and this is OK. 
for index 1, `ofm_idx`, it is handled inside `loco::NodeShape infer_unique_out(const luci::CircleUniqueOut *node)` and
also OK.  So as @mhs4670go pointed, current code is working OK.

> I do not know what changed between then and now

Actully there is no changes besides refactoring done by me.
I wonder what was the problem @dr-venkman was struggling.

seanshpark (approved):
LGTM

dr-venkman:
> > ofm was never visited, and therefore it's shape inferencing was not handled inside the visit(const luci::CircleUniqueOut *node) method
> 
> For index 0, `ofm`, result shape was set by the line @mhs4670go pointed
> 
> ```c++
>       shape = own_shape(circle_node); // here
> ```

Does it then mean that `ofm`'s ShapeInferencing is not handled inside `loco::NodeShape infer_unique_out(const luci::CircleUniqueOut *node)`? If so, then this is what I was looking to avoid, based on previous comments. The reasoning was that, similar to TypeInferencing, Shapeinferencing for UniqueOut should be handled inside `loco::NodeShape infer_unique_out(const luci::CircleUniqueOut *node)`, with the `node->index()` checks handling `ofm` and `ifm_idx` separately.
I did manage to pass the testcases earlier without having to handle `ofm` explicitly inside `loco::NodeShape infer_unique_out(const luci::CircleUniqueOut *node)`. 

> 
> and this is OK.
> for index 1, `ofm_idx`, it is handled inside `loco::NodeShape infer_unique_out(const luci::CircleUniqueOut *node)` and
> also OK. So as @mhs4670go pointed, current code is working OK.
> 
> > I do not know what changed between then and now
> 
> Actully there is no changes besides refactoring done by me.
> I wonder what was the problem @dr-venkman was struggling.

seanshpark:
> The reasoning was that, similar to TypeInferencing, Shapeinferencing for UniqueOut should be handled inside loco::NodeShape infer_unique_out(const luci::CircleUniqueOut *node)

Thanks for explaining what have happened. 
I think there was some misunderstanding. There is no restriction that all shape inference MUST be inside `infer_unique_out` method. For common cases with unknown shape (or scalar), to avoid all the shape inference code modification, `shape = own_shape(circle_node); // here` code was introduced.

seanshpark:
@dr-venkman , @mhs4670go requested your review.

dr-venkman (approved):
LGTM

dr-venkman:
> > The reasoning was that, similar to TypeInferencing, Shapeinferencing for UniqueOut should be handled inside loco::NodeShape infer_unique_out(const luci::CircleUniqueOut *node)
> 
> Thanks for explaining what have happened.
> I think there was some misunderstanding. There is no restriction that all shape inference MUST be inside `infer_unique_out` method. For common cases with unknown shape (or scalar), to avoid all the shape inference code modification, `shape = own_shape(circle_node); // here` code was introduced.

Thanks for clarifying, I will keep in mind when implementing new operators in future.

PR #4020, opened 2020-08-28
[onert] Fix log generation fail for LeakyRelu

hseok-oh:
Fix ElementwiseActivation node name print for LeakyRelu


ragmani (approved):
LGTM

PR #4026, opened 2020-08-28
[onert/test] Build nnpackage_run_random for memory check

hseok-oh:
- Build nnpackage_run_random: build without hdf5 load/dump
- For comparison memory usage with tflite_run: hdf5 library spent about 3MB

### DIFF
--- a/tests/tools/nnpackage_run/CMakeLists.txt
+++ b/tests/tools/nnpackage_run/CMakeLists.txt
@@ -15,31 +15,44 @@ nnfw_find_package(Boost REQUIRED program_options)
 nnfw_find_package(Ruy QUIET)
 nnfw_find_package(HDF5 QUIET)
 
-if (HDF5_FOUND)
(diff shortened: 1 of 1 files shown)


glistening:
@hseok-oh I don't want to diverse `nnpackage_run` just for difference of using random or hdf5. We don't create different name of `nnpackage_run` for each configuration (e.g. `nnpackage_run_benchmark`, `nnpackage_run_memory_profiling`, `nnpackage_run_random_profiling' ,...

glistening:
My suggestion is:

- If you don't use `hdf5` feature from `nnpackage_run`, build without `hdf5` features. You don't need to create a new name of `nnpackage_run`.
- If you don't use `hdf5` and you're sure how much memory `hdf5` library uses (it seems about 3MB), just subtract 3MB from measured memory.

hseok-oh:
OK. I'll close this PR.

Issue #4033, opened 2020-08-31
Support gelu operator


seanshpark:
Let's check new `gelu` operator in TF nightly (2.4.0-dev) version

seanshpark:
tflite file from using `tf2tfliteV2.py` script: [gelu.tflite.zip](https://github.com/Samsung/ONE/files/5149413/gelu.tflite.zip)

seanshpark:
CC @chunseoklee

seanshpark:
Run in TF2.3.4 virtual env
```
    op_ = tf.nn.gelu(in_, approximate=True, name="Output")
AttributeError: module 'tensorflow._api.v2.nn' has no attribute 'gelu'
```

chunseoklee:
> Run in TF2.3.0 virtual env. shows an error
> 
> ```
>     op_ = tf.nn.gelu(in_, approximate=True, name="Output")
> AttributeError: module 'tensorflow._api.v2.nn' has no attribute 'gelu'
> ```

This error is resolved by using TF-nightyly ?

seanshpark:
> This error is resolved by using TF-nightyly ?

Yup :)

seanshpark:
Added gelu.pbtxt file at https://github.com/Samsung/ONE/issues/4033#issuecomment-683660496

seanshpark:
I think this can be closed

PR #4034, opened 2020-08-31
[res] gelu TF examples

seanshpark:
This will add TF examples for gelu operator

### DIFF
--- a/res/TensorFlowPythonExamples/examples/gelu/__init__.py
+++ b/res/TensorFlowPythonExamples/examples/gelu/__init__.py
@@ -1,0 +1,7 @@
+# NOTE please use TF2.4.0-dev or above to use gelu op
+import tensorflow as tf
+
+tf.compat.v1.disable_eager_execution()
+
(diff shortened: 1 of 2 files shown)


seanshpark:
For  #4033

mhs4670go (approved):
LGTM

mhs4670go:
@nnfw-bot test nncc-debug

mhs4670go:
@nnfw-bot test nncc-release

mhs4670go:
@seanshpark Just in my opinion, even though of course we officially supports tensorflow 1.X, sometimes we run this python script with tensorflow 2.X, which throws an error because tensorflow 2.X needs to be eager execution disabled like this PR. How about adding the command that disables eager execution to python script by default?

seanshpark:
> How about adding the command that disables eager execution to python script by default?

Need to get opinion from @parjong  :)

For me, I'm note sure about this at the moment...

seanshpark:
Not sure why debug/release build fails... :(

mhs4670go:
@seanshpark Hmmm.. neither no I..

seanshpark:
close as another #4046 passed

PR #4050, opened 2020-09-01
[onert/cmake] Introduce option build without hdf5

hseok-oh:
Introduce cmake option variable to support build without hdf5

---

Related issue: https://github.com/Samsung/ONE/pull/4026#issuecomment-682377045


seanshpark (approved):
LGTM

PR #4056, opened 2020-09-01
[luci] add condition that builds CircleSparseToDense node

mhs4670go:
This commit adds condition that builds CircleSparseToDense node.

Somethimes there is no builtin option.

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleSparseToDense.cpp
+++ b/compiler/luci/import/src/Nodes/CircleSparseToDense.cpp
@@ -42,7 +42,8 @@ CircleNode *CircleSparseToDenseGraphBuilder::build_node(const circle::OperatorT 
   node->default_value(inputs.at(3));
 
   const auto *options = op.builtin_options.AsSparseToDenseOptions();
-  node->validate_indices(options->validate_indices);
+  if (options)
+    node->validate_indices(options->validate_indices);
 
   return node;
 }


compiler/luci/import/src/Nodes/CircleSparseToDense.cpp
```diff
@@ -42,7 +42,8 @@ CircleNode *CircleSparseToDenseGraphBuilder::build_node(const circle::OperatorT
   node->default_value(inputs.at(3));
 
   const auto *options = op.builtin_options.AsSparseToDenseOptions();
-  node->validate_indices(options->validate_indices);
+  if (options)
```
llFreetimell, line 45: When `options` is `nullptr`, then what is default value of that option?

mhs4670go: Hmm.. `nullptr` is the default value.

llFreetimell: Ah I mean... when options is nullptr, then `validate_indices` is `true` or `false`?
Because when some option value is optional, **Optional** keyword is included in fbs file.
For example, `Shape` operation has following options,
```
table ShapeOptions {
  // Optional output type of the operation (int32 or int64). Defaults to int32.
  out_type : TensorType;
}
```
and `LSTM` operation has following options.
```
  cell_clip: float; // Optional, 0.0 means no clipping
  proj_clip: float; // Optional, 0.0 means no clipping
```

mhs4670go: @llFreetimell That is false.
```cxx
struct SparseToDenseOptionsT : public flatbuffers::NativeTable {
  typedef SparseToDenseOptions TableType;
  bool validate_indices;
  SparseToDenseOptionsT()
      : validate_indices(false) {
  }
};
```
FYI, all unspecified fields will default to 0 or NULL.

llFreetimell: Hmm... if then, should we modify as following or go as it is?
```cpp
  if (options)
    node->validate_indices(options->validate_indices);
  else
    node->validate_indices(false);
```

seanshpark: current source has
```cxx
class CircleSparseToDense final
  ...
  bool _validate_indices{true};
```
so, if options == nullptr, what should we set to `node->validate_indices()` ?

seanshpark:
@nnfw-bot test nncc-debug

compiler/luci/import/src/Nodes/CircleSparseToDense.cpp
mhs4670go: I think we should change the default value of `_validate_indices` to false.

mhs4670go: So, this PR can be merged as it is and I'll post another PR that has contents I said above.

seanshpark: OK

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

PR #4059, opened 2020-09-01
[onert] Reduce memory usage in Conv qasymm uint8 operation

hseok-oh:
Move allocate for im2col space from configure to run  because im2col data cannot reuse

### DIFF
--- a/compute/cker/include/cker/operation/Conv.h
+++ b/compute/cker/include/cker/operation/Conv.h
@@ -54,11 +54,7 @@ inline void TransposeFloatTensor(const float *input_data, const nnfw::cker::Shap
 class Conv
 {
 public:
-  Conv()
-      : _modified_filter_data(), _im2col_data(), _im2col_shape(4), _need_im2col(false),
-        _prepared(false)
-  {
-  }
+  Conv() : _modified_filter_data(), _im2col_shape(4), _need_im2col(false), _prepared(false) {}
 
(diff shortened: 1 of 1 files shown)


hseok-oh:
quantized inception v3: 73556 KB -> 42448 KB

hseok-oh:
PTAL

compute/cker/include/cker/operation/Conv.h
```diff
@@ -121,9 +117,10 @@ class Conv
                        params.stride_height);
     }
 
-    uint8_t *im2col_raw_data = _im2col_data.data();
+    int im2col_size = _need_im2col ? _im2col_shape.FlatSize() : 1;
+    uint8_t im2col_data[im2col_size];
```
glistening, line 121: @hseok-oh `im2col_data` will be allocated from stack at runtime. Allocation may fail if `im2col_size` is big.

glistening (commented):
@hseok-oh Do you have running time comparison?

compute/cker/include/cker/operation/Conv.h
hseok-oh: OK. I'll use heap for this.

hseok-oh:
>Do you have running time comparison?

Before fix: 340 ms
Fix using stack only: 340 ms
Fix using heap only: 400 ms

glistening (approved):
LGTM

PR #4063, opened 2020-09-02
[luci] Change default value of CircleSparseToDense's attribute

mhs4670go:
This commit changes default value of CircleSparseToDense's attribute.

All unspecified value in schema will be 0 or null or sth by default.

### DIFF
--- a/compiler/luci/lang/include/luci/IR/Nodes/CircleSparseToDense.h
+++ b/compiler/luci/lang/include/luci/IR/Nodes/CircleSparseToDense.h
@@ -49,7 +49,7 @@ public:
   void validate_indices(bool validate_indices) { _validate_indices = validate_indices; }
 
 private:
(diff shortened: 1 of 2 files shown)


seanshpark (approved):
LGTM

seanshpark:
> 11:03:11 [ RUN      ] CircleSparseToDenseTest.constructor
11:03:11 /opt/jenkins_slave/workspace/nnfw/master/pr-nncc-debug/compiler/luci/lang/src/Nodes/CircleSparseToDense.test.cpp:36: Failure
11:03:11       Expected: true
11:03:11 To be equal to: stb_node.validate_indices()
11:03:11       Which is: false
11:03:11 [  FAILED  ] CircleSparseToDenseTest.constructor (0 ms)

seanshpark (approved):
LGTM

PR #4029, opened 2020-08-31
DRAFT: On going draft for concat qparam propagation

jinevening:
On going draft to propagate qparam of concat Op to its input tensors

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to #4004

Issue #4004, opened 2020-08-27
Propagating quantization parameters of Concatenation Op


jinevening:
## In brief
We can remove rescaling in Concat Op by propagating the quantization parameters of the output of Concat Op to its inputs.

## Motivation

Let's assume inputs of Concat Op are quantized with different quantization parameters (scale, zp).

Concat Op not only just concatenates the input tensors but also rescales the values of the input tensors to make them have the same quantization parameters with the output tensor. Rescaling requires floating point division, multiplication and rouding ([link](https://github.com/Samsung/ONE/blob/master/compute/cker/include/cker/operation/Concatenation.h#L140-L146)), introducing computational overheads. Moreover, for the integer-based hardware such as npu, rescaling is approximated using shift and multiplication, which inevitably causes loss of precision.

![image](https://user-images.githubusercontent.com/5449554/91246426-bff18100-e78a-11ea-88c0-a9c17e139937.png)

For example, in the above graph (captured in Inception V3), Concat Op rescales the values of the input tensors (1)~(4) to have the same quantization parameters with the output tensor (5).

## Optimization chance

If we know the quantization parameters of the output tensor of the Concat Op before running the model (i.e., activations were statically quantized), we can avoid rescaling in Concat Op by overwriting the output tensor's quantization parameters to the input tensors.

## Issues

1. Subsequent concats.
  \- If an output of concat Op is given as an input of another concat Op, the quantization parameters of the later concat can be propagated to the upper concat. For example, in the below figure, quantparams of (5) can be propagated to (1), (2), (3), (4).

![image](https://user-images.githubusercontent.com/5449554/91371902-3ce52f00-e84d-11ea-910c-2acb923561ac.png)

 \- It seems that TFLite converter converts subsequent concats into one concat. (Subsequent concats in the pb file of InceptionV3 are replaced with one concat in tflite). Since ONE is using TFLite converter, **we will not consider subsequent concats for now**.

2. We should not propagate quantparams to an input tensor produced to other Ops.
 \- Overwriting quantparams to such an input tensor would affect the result of other Ops (e.g., concat in AI sound model).
 \- For example, in the below figure, if quantparams of (3) is propagated to (1) and (2), the result of `conv2` will be distorted as the quantparams of (1), i.e., the input of `conv2`, changes. Therefore we only overwrite (3) to (2) and leave (1) as-is.

[...]

PR #4064, opened 2020-09-02
[luci] Propagate concat qparam to its inputs

jinevening:
This propagates concat qparam to its inputs

---
Related to #4004 Draft PR #4029

### DIFF
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -19,6 +19,7 @@
 #include "luci/Pass/FuseBatchNormWithTConv.h"
 #include "luci/Pass/FuseBCQPass.h"
 #include "luci/Pass/FuseInstanceNormPass.h"
+#include "luci/Pass/PropagateConcatenationQparamPass.h"
 #include "luci/Pass/ResolveCustomOpAddPass.h"
 #include "luci/Pass/ResolveCustomOpBatchMatMulPass.h"
 #include "luci/Pass/ResolveCustomOpMatMulPass.h"
@@ -209,6 +210,9 @@ void CircleOptimizer::quantize(loco::Graph *g) const
     luci::QuantizeWithMinMaxPass quantizer(str_to_dtype(input_dtype), str_to_dtype(output_dtype),
                                            str_to_granularity(granularity));
     quantizer.run(g);
+
+    luci::PropagateConcatenationQparamPass propagator;
+    propagator.run(g);
   }
 
   // Requantize
--- a/compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
+++ b/compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/PropagateConcatenationQparamPass.h"
+
+#include "PropagateConcatenationQparamPassInternal.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/AttrFusedActFunc.h>
+#include <luci/Log.h>
+
+namespace
+{
+
+bool is_uint8_quantized(luci::CircleNode *node)
+{
+  if (node->dtype() == loco::DataType::U8 && node->quantparam() != nullptr &&
+      node->quantparam()->scale.size() > 0 && node->quantparam()->zerop.size() > 0)
+    return true;
+
+  return false;
+}
+
+void overwrite_quantparam(luci::CircleConcatenation *concat, luci::CircleNode *target)
+{
+  auto concat_qparam = concat->quantparam();
+  assert(concat_qparam != nullptr);
+
+  auto target_qparam = target->quantparam();
+  assert(target_qparam != nullptr);
+  target_qparam->min = concat_qparam->min;
+  target_qparam->max = concat_qparam->max;
+  target_qparam->scale = concat_qparam->scale;
+  target_qparam->zerop = concat_qparam->zerop;
+  target_qparam->quantized_dimension = concat_qparam->quantized_dimension;
+}
+
+} // namespace
+
+namespace luci
+{
+
+void propagate_concat_quantparam(luci::CircleConcatenation *concat)
+{
+  // Check if concat is uint8-quantized
+  if (!is_uint8_quantized(concat))
+    return;
+
+  // Check if concat has no fused activation function
+  if (concat->fusedActivationFunction() != luci::FusedActFunc::NONE)
+    return;
+
+  const auto num_inputs = concat->numValues();
+  for (uint32_t i = 0; i < num_inputs; i++)
+  {
+    auto node = static_cast<luci::CircleNode *>(concat->arg(i));
+
+    // Skip if this input is not uint8-quantized
+    if (!is_uint8_quantized(node))
+      continue;
+
+    // Skip if concat and node have different quantized_dimension
+    if (node->quantparam()->quantized_dimension != concat->quantparam()->quantized_dimension)
+      continue;
+
+    // Skip if this input is CONCAT Op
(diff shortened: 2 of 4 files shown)


seanshpark (approved):
LGTM

compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
```diff
@@ -0,0 +1,131 @@
...
+    if (not concat)
+      continue;
+
+    // Propagate qparam of concat to its inputs if
+    // (1) concat is uint8-quantized
+    // (2) concat has no fused activation function
+    // (3) the input is uint8-quantized
+    // (4) the input is not concatenation Op
+    // (5) the input is not produced to Ops other than concat
+    propagate_concat_quantparam(concat);
```
s-barannikov, line 125: It should also take into accout `CircleInput` and `CircleConst`.

compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -209,6 +210,9 @@ void CircleOptimizer::quantize(loco::Graph *g) const
     luci::QuantizeWithMinMaxPass quantizer(str_to_dtype(input_dtype), str_to_dtype(output_dtype),
                                            str_to_granularity(granularity));
     quantizer.run(g);
+
+    luci::PropagateConcatenationQparamPass propagator;
+    propagator.run(g);
```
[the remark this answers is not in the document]
jinevening: There is no solid reason for that.

I made it as a Pass simply for the consistency with other optimizations. It would be ok to change it to a function if there is a clear benefit.

compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
```diff
@@ -0,0 +1,131 @@
...
+  if (concat->fusedActivationFunction() != luci::FusedActFunc::NONE)
+    return;
+
+  const auto num_inputs = concat->numValues();
+  for (uint32_t i = 0; i < num_inputs; i++)
+  {
+    auto node = static_cast<luci::CircleNode *>(concat->arg(i));
+
+    // Skip if this input is not uint8-quantized
+    if (!is_uint8_quantized(node))
```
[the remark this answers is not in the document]
jinevening: I'll add an assert for safety.

compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
jinevening: For `CircleInput`, I think it is ok to just overwrite qparam to the input node. The updated qparam is saved in the model, so users can give an input quantized with a proper qparam to the model.

For `CircleConst`,we have to quantize constant values according to propagated qparam. I'll add code for that. Thanks!

s-barannikov: > The updated qparam is saved in the model, so users can give an input quantized with a proper qparam to the model.

So, the user will have to requantize the input data because of the optimization? Well, I would not be happy ;D
Or I'm missing something?

jinevening: > So, the user will have to requantize the input data because of the optimization?

I think users do not have to 're'quantize the input data. Concat qparam is propagated when fp32 model is quantized to u8 model. Therefore users will only see the overwritten quantization parameters (users will not see the original qparam ever), which means input quantization is performed once.

s-barannikov: Oh, I see, thanks.

compiler/luci/pass/src/CircleOptimizer.cpp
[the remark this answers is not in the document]
s-barannikov: The benefit is - less dependencies and less code to support. Never mind, I was just curious.

seanshpark (approved):
LGTM

jinevening:
@s-barannikov  I've added the code to quantize const tensors (a test for quantized values will be included in the following PR).  PTAL.

compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
```diff
@@ -0,0 +1,161 @@
...
+        auto data = const_node->at<loco::DataType::FLOAT32>(i);
+        quantized_values[i] = static_cast<int32_t>(std::round(data * scaling_factor_inv) + zerop);
+      }
+
+      const_node->dtype(loco::DataType::U8);      // change the type of tensor
+      const_node->size<loco::DataType::U8>(size); // resize tensor
+      for (uint32_t i = 0; i < size; ++i)
+      {
+        const_node->at<loco::DataType::U8>(i) = std::min(255, std::max(0, quantized_values[i]));
+      }
```
s-barannikov, line 113: IMHO It's worth extracting into a function.

compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
```diff
@@ -0,0 +1,161 @@
...
+    if (succs.size() != 1)
+      continue;
+
+    assert(succs.find(concat) != succs.end());
+
+    // Quantize constant values
+    if (node->opcode() == luci::CircleOpcode::CIRCLECONST)
+    {
+      luci::CircleConst *const_node = loco::must_cast<luci::CircleConst *>(node);
+      assert(const_node->dtype() == loco::DataType::FLOAT32);
```
s-barannikov, line 88: Not sure I understand this.
If the concat has quantization parameters and U8 type, how can its input be float?

jinevening: QuantizeWithMinMaxPass quantizes output tensor of concat, while it does not quantize constant inputs of concat.
(Currently the only quantized constant tensors are weights and bias of conv, dw_conv, tconv, fc layers)

So when PropagateConcatenationQparamPass is called, the output tensor of concat is already quantized but const inputs of concat are not yet quantized.

compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
jinevening: It would enhance readability. I've updated the code.

compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
s-barannikov: Do you mean that without this optimization the graph would be invalid?
I think that any transformation should leave the graph in valid state. This allows to make them independent on each other, and requires less "internal knowledge" of each transformation.

jinevening: Our quantizer does not have a general function to handle constant inputs of an Op (as mentioned above, we only handle constant inputs for conv, dw_conv, tconv, fc layer). This is because we need to quantize constant inputs in an operator-specific way. For example, ReduceMean can receive a constant input as `axis`, which should not be quantized. So I wanted to implement quantization logics for each operator as the need arises (ex: a new pattern occurs in target models).

> Do you mean that without this optimization the graph would be invalid?
> I think that any transformation should leave the graph in valid state. This allows to make them independent on each other, and requires less "internal knowledge" of each transformation.

Agreed. I think we have to throw an exception for the unsupported cases (const inputs of unsupported Ops) to prevent the graph from being invalid. I'll make another PR for that.

s-barannikov: > I'll make another PR for that.

👍 Do you want to make it first, or would you like to merge this one now?

> This is because we need to quantize constant inputs in an operator-specific way. For example, ReduceMean can receive a constant input as `axis`, which should not be quantized. 

I see. As a suggestion, I saw some kind of black- or white-listing function for that in TFLite, and it works for them well. I could try to find the exact place if you want.

seanshpark (approved):
LGTM

compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
jinevening: > Do you want to make it first, or would you like to merge this one now?

I'd like to merge this now because it seems to be a separate issue.

> I see. As a suggestion, I saw some kind of black- or white-listing function for that in TFLite, and it works for them well. I could try to find the exact place if you want.

It would be very helpful to specify the exact place for the code. Thanks :)

s-barannikov (approved):

compiler/luci/pass/src/PropagateConcatenationQparamPass.cpp
s-barannikov: @jinevening Here is the relevant [code](https://github.com/tensorflow/tensorflow/blob/312c11c4ecdc2c466895fc3be0a16d60c692422e/tensorflow/lite/tools/optimize/operator_property.cc#L67-L994) and [data structures](https://github.com/tensorflow/tensorflow/blob/312c11c4ecdc2c466895fc3be0a16d60c692422e/tensorflow/lite/tools/optimize/operator_property.h#L57-L126).
The code may look scary, but half of the function is just LSTM case.

mhs4670go (approved):
LGTM

PR #4070, opened 2020-09-02
[luci] Make luci::Importer accept files

s-barannikov:
This reduces amount of boilerplate code needed to load a model from file before passing to luci::Importer.


s-barannikov:
@seanshpark What do you think of it?

seanshpark:
Code is reduced: this is good :)
Actually importer was implemented as-is for not having file related codes inside.
I wanted to separate import itself and where to import from: just interface to access graphs(?)
So I'm not confident with the changes -_-;

seanshpark:
> What if I add the new file-based interface while keeping existing one ....

Is there any needs in SRR or by you for the changes?

PR #4085, opened 2020-09-04
[luci] Test for concat qparam propagation

jinevening:
This is a test for concat qparam propagation

---
Related to #4004 Draft PR #4029

### DIFF
--- a/compiler/luci/pass/src/PropagateConcatenationQparamPass.test.cpp
+++ b/compiler/luci/pass/src/PropagateConcatenationQparamPass.test.cpp
@@ -1,0 +1,224 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

compiler/luci/pass/src/PropagateConcatenationQparamPass.test.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    concat_node.values(0, &input_1);
+    concat_node.values(1, &input_2);
+
+    addQuantParam(concat_node, {3.14}, {77});
+    addQuantParam(input_1, {1.0}, {1});
+    addQuantParam(input_2, {2.0}, {2});
+  }
+
+  ~SimpleConcatGraph()
```
s-barannikov, line 57: Is it necessary to define the destructor and explicitly reset the pointers?

s-barannikov (approved):

compiler/luci/pass/src/PropagateConcatenationQparamPass.test.cpp
jinevening: Yes. If `input_1` or `input_2` is used by `concat_node` when the destructor is called, the below assertion occurs in `loco/src/IR/Node.cpp`.

```cpp
Node::~Node()
{
  // To detect dangling references
  assert(_uses.size() == 0);
}
```

s-barannikov: I see. This is probably because the nodes are not linked to Graph inputs / outputs.

PR #4086, opened 2020-09-04
DRAFT compiler documents

seanshpark:
on-going draft to add compiler related documents

### DIFF
--- a/docs/howto/how-to-build-compiler.md
+++ b/docs/howto/how-to-build-compiler.md
@@ -1,7 +1,114 @@
 # How to Build Compiler
 
+This document is based on the system where Ubuntu Desktop Linux 18.04 LTS is installed with default
+settings, and can be applied in other environments without much difference. For reference, the
+development of our project started in the Ubuntu Desktop Linux 16.04 LTS environment.
+As of now, to build in 16.04, please use gcc 7.x or above.
+
 ## Build Requires
+
+If you are going to build this project, the following modules must be installed on your system:
+
+- CMake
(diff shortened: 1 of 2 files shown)


docs/howto/how-to-build-compiler.md
```diff
@@ -1,7 +1,106 @@
...
+To run unit tests:
+```
+$ ./nncc test
+```
+
+Above steps will build all the modules in the compiler folder. There are modules that are currently not active.
+To buld only as of now active modules of the compiler, we provide a preset of modules. Latest is `20200630`.
+
+```
+$ ./nnas create-package --preset 20200630 --prefix build/install
```
mhs4670go, line 78: Not much related this PR, but it would be good to make users be able to build the latest preset without any preset.

docs/howto/how-to-build-compiler.md
```diff
@@ -1,7 +1,106 @@
...
 
+If you are going to build this project, the following modules must be installed on your system:
+
+- CMake
+- Boost C++ libraries
+
+In the Ubuntu, you can easily install it with the following command.
+
+```
+$ sudo apt-get install cmake libboost-all-dev
```
jinevening, line 15: `apt-get` -> `apt` or vice versa for consistency with the command in line 23?

docs/howto/how-to-build-compiler.md
seanshpark: I think we can omit `--preset 20200630` as there is default, currently `20200630`

PR #4088, opened 2020-09-04
[onert] Log subgraph calls during execution

wateret:
Log subgraph calls during execution so we can trace the execution stack.


ragmani (approved):
LGTM

PR #4089, opened 2020-09-04
[compiler] bump version to 1.10.0

seanshpark:
This will revise vconone version to 1.10.0


mhs4670go (approved):
LGTM

PR #4091, opened 2020-09-04
[nnfwapi] Log cond subgraph call of While op

wateret:
cond subgraph call logging was missing.


chunseoklee (approved):

ragmani (approved):
LGTM

Issue #4053, opened 2020-09-01
[onert] Mismatch between model schema and IR in some operations


ragmani:
Onert IR of some operations are incorrectly defined compared to model schema. In model schema, these operations have an input as input that can be non-constant. But they have it as a param in IR. It must be fixed to input that can be non-constant.

#### The operations
- ArgMax : axis
- ResizeNearestNeighbor : size
- ResizeBilinear : size
- Split : axis
- Tile : multiplies
- Transpose : permute factor

ragmani:
Done.

PR #4092, opened 2020-09-04
[onert] Add an optional input to ir::ResizeBilinear

ragmani:
For issue #4053

This commit adds an optional input to ir::ResizeBilinear for supporting non-constant.

### DIFF
--- a/runtime/onert/core/include/ir/operation/ResizeBilinear.h
+++ b/runtime/onert/core/include/ir/operation/ResizeBilinear.h
@@ -34,10 +34,12 @@ public:
   enum Input
   {
     INPUT = 0,
+    SIZE = 1,
   };
 
   struct Param
   {
+    // If the input SIZE exists in inputs, height_out and width_out are not set. Ignore these params
     int32_t height_out;
     int32_t width_out;
     bool align_corners;
--- a/runtime/onert/backend/acl_cl/KernelGenerator.cc
+++ b/runtime/onert/backend/acl_cl/KernelGenerator.cc
@@ -879,7 +879,6 @@ void KernelGenerator::visit(const ir::operation::Permute &node)
 void KernelGenerator::visit(const ir::operation::ResizeBilinear &node)
 {
   const auto ofm_index{node.getOutputs().at(0)};
-
   const auto ifm_index{node.getInputs().at(ir::operation::ResizeBilinear::Input::INPUT)};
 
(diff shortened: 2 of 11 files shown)


runtime/onert/core/include/ir/operation/ResizeBilinear.h
```diff
@@ -34,10 +34,12 @@ class ResizeBilinear : public Operation
   enum Input
   {
     INPUT = 0,
+    SIZE = 1,
   };
 
   struct Param
   {
+    // If the input SIZE exists in inputs, height_out and width_out are not set. Ignore these params
```
s-barannikov, line 42: `height_out` and `width_out` are used in nnapi frontend only, am I right?
Is it possible to create a new tensor (size) from these two variables in the nnapi frontend and then remove them?

ragmani: It is not simple. According to nnapi spec, `height_out` and `width_out` each have their own index. This makes it is not simple to create a new tensor from these two operands.
First, if `height_out` and `width_out` are constant, their values will be set by the api `ANeuralNetworksModel_setOperandValue` respectively. But there is no operation information in the api. This means that runtime should find and match them to a new operand (size). If IR of this operation doesn't have `height_out` and `width_out`, where should runtime works it within frontend.
Second, if `height_out` and `width_out` are non-constant, it's really complicated.Their values will be set by the api `ANeuralNetworksExecution_setInput` or outputs of other operations respectively. `ANeuralNetworksExecution_setInput` don't have operation information too. This would cause a similar problem to the first. Furthermore, the works to find and match them make the performance degradation in execution time. The critical problem is the case when these are when they are output of the other operations. I can't even think of them as the output of other operations because this operation has them as a input(size) in model of tflite, But nnapi treats it as two inputs. I think this means that the two operands never be non-constant or that nnapi has a spec error.

wateret:
Please add some test cases :)

ragmani:
> Please add some test cases :)

@wateret
Unfortunately, this changes don't support the optional(size) input with nnapi. For details, refer to https://github.com/Samsung/ONE/pull/4092#discussion_r484176366.
The optional input is int32 type, But I don't know how to test int32 input without using nnapi.

s-barannikov (approved):
LGTM +0.5

wateret:
@ragmani I got that. Once #4114 , you probably are able to create one :) I'll merge this and leave an issue for a test case.

wateret (approved):
LGTM

PR #4096, opened 2020-09-04
[onert] Move wrong param axis of Split to inputs

ragmani:
For issue #4053

This commit moves wrong param axis of Split to inputs.


hyunsik-yoon (dismissed):
It would be better if we have test cases for these ops (e.g., split, argmax, etc) when axis is not a const.

ragmani:
> It would be better if we have test cases for these ops (e.g., split, argmax, etc) when axis is not a const.

Okay, I will add test cases.

hyunsik-yoon (approved):
LGTM :-)

wateret (approved):
LGTM

PR #4101, opened 2020-09-07
[luci] Propagate concat qparam during quantization

jinevening:
This changes PropagateConcatenationQparamPass to a function inside QuantizeWithMinMaxPass

  - To guarantee the validity of graph before/after applying Pass

---
Related to #4004


jinevening:
This is refactoring of #4064 and #4085.

[This comment](https://github.com/Samsung/ONE/pull/4064#discussion_r483361713) is the reason for the refactoring.

seanshpark (dismissed):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -516,6 +559,74 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
+    if (succs.size() != 1)
+      continue;
+
+    assert(succs.find(concat) != succs.end());
+
+    // Quantize constant values
+    if (node->opcode() == luci::CircleOpcode::CIRCLECONST)
+    {
+      luci::CircleConst *const_node = loco::must_cast<luci::CircleConst *>(node);
+      assert(const_node->dtype() == loco::DataType::FLOAT32);
```
s-barannikov, line 609: Looks redundant in the view of the check on the next line. It could also break a negative test in debug mode (if it relies on throwing an exception).

s-barannikov (approved):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
seanshpark: I'll wait for this before merge...

jinevening: > It could also break a negative test in debug mode

@s-barannikov  Agreed. I've removed the check. PTAL

seanshpark (approved):
LGTM

s-barannikov (approved):

PR #4118, opened 2020-09-07
[res/TensorFlowLiteRecipes] Add LogSoftmax_U8_000

underflow101:
Parent Issue: #1880
Fired Issue: #4119 

This commit enable U8 recipe for LogSoftmax_000 Op.

Please review this PR, @seanshpark .
I'll be happy to get feedback and change to make improvement.

Thank you.

(Derived from SOS Mine Project)

Signed-off-by: underflow101 <ikarus125@gmail.com>

### DIFF
--- a/res/TensorFlowLiteRecipes/LogSoftmax_U8_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/LogSoftmax_U8_000/test.recipe
@@ -1,0 +1,21 @@
+operand {
+  name: "ifm"
+  type: UINT8
+  shape { dim: 1 dim: 3 dim: 3 dim: 2 }
+  quant { min: -10 max: 10 scale: 0.0625 zero_point: 255 }
+}
+operand {
+  name: "ofm"
+  type: UINT8
+  shape { dim: 1 dim: 3 dim: 3 dim: 2 }
+  quant { min: -10 max: 10 scale: 0.0625 zero_point: 255 }
+}
+operation {
+  type: "LogSoftmax"
+  log_softmax_options {
+  }
+  input: "ifm"
+  output: "ofm"
+}
+input: "ifm"
+output: "ofm"
--- a/res/TensorFlowLiteRecipes/LogSoftmax_U8_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/LogSoftmax_U8_000/test.recipe
@@ -2,13 +2,13 @@ operand {
   name: "ifm"
   type: UINT8
   shape { dim: 1 dim: 3 dim: 3 dim: 2 }
-  quant { min: -10 max: 10 scale: 0.0625 zero_point: 255 }
+  quant { min: -4.952 max: 4.939 scale: 0.0388 zero_point: 128 }
 }
 operand {
   name: "ofm"
   type: UINT8
   shape { dim: 1 dim: 3 dim: 3 dim: 2 }
-  quant { min: -10 max: 10 scale: 0.0625 zero_point: 255 }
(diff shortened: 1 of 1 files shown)


res/TensorFlowLiteRecipes/LogSoftmax_U8_000/test.recipe
```diff
@@ -0,0 +1,21 @@
+operand {
+  name: "ifm"
+  type: UINT8
+  shape { dim: 1 dim: 3 dim: 3 dim: 2 }
+  quant { min: -10 max: 10 scale: 0.0625 zero_point: 255 }
```
seanshpark, line 5: Is there a reason for using 255 as ZP?
I'm not good at quantization... , if ZP is 255, how is near max value, say 9.9 represented?

jinevening: > if ZP is 255, how is near max value, say 9.9 represented?

If zp is 255, only negative or zero values can be expressed.

@underflow101 , I think quantparam should be changed to something like below.

Input qparam
min: -4.952, max: 4.939, scale: 0.0388, zero_point: 128

Output qparam
min: -15.9375, max: 0, scale: 0.0625, zero_point: 255

underflow101: Hmm... This must be from my lack of understanding. I apologize for this. I will re-commit after editing.

I re-calculated input's quantizationparams according to `TestUtils.h` under `luci-interpreter/src/kernel` dir, and understood why zero_point of ifm has to be 128.

Thank you so much for your advice @seanshpark @jinevening.

But why `min: -4.952` & `max: 4.939`?
[reference link](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/activations_test.cc#L1987-L2008)
> `      /*input=*/{TensorType_UINT8, {2, 4}, -10, 10},`

The reference shows input of `kmin = -10`, `kmax = 10`.

jinevening: > But why min: -4.952 & max: 4.939?

I just picked the numbers from another example. 

> The reference shows input of kmin = -10, kmax = 10.

You don't have to use the min/max values in the test code for the recipe. You can use any min/max for ifm. The important thing is that scale and zero point should be calculated based on the corresponding min/max values.

underflow101: >You don't have to use the min/max values in the test code for the recipe. You can use any min/max for ifm. The important thing is that scale and zero point should be calculated based on the corresponding min/max values.

Oh, now I fully understood!! :D

Thank you so much for your explanation!

jinevening: You can set min/max to -10/+10, same as the values in the test code. If so, please calculate scale and zero point using `quantizationParams` in `TestUtils.h`.

jinevening (approved):
LGTM

res/TensorFlowLiteRecipes/LogSoftmax_U8_000/test.recipe
underflow101: > You can set min/max to -10/+10, same as the values in the test code. If so, please calculate scale and zero point using quantizationParams in TestUtils.h.

If I set min/max to -10/+10, scale would be 0.078431 and zero_point would be 128...!
I'll do better from now on :)
Thank you so much, @jinevening :D

seanshpark (approved):
LGTM thank you!

PR #4121, opened 2020-09-08
[onert/tool] Rootfs generator for ubuntu 20.04

hseok-oh:
install_rootfs.sh supports ubuntu 20.04 host/target (ex. odroid-xu4)


lemmaa (approved):
LGTM. Thanks!

seanshpark (approved):
LGTM

hyunsik-yoon:
Does this mean that ubuntu 20.04 is also one of the targets we officially support?

hseok-oh:
>Does this mean that ubuntu 20.04 is also one of the targets we officially support?

If you focus on **officially**, maybe not yet. Currently we don't support 14.04, aarch64 16.04, armv7l 18.04 officially.

PR #4126, opened 2020-09-08
[circledump] Revise dump function that dumps flatbuffers vector

mhs4670go:
This commit revises dump function that dumps flatbuffers vector.

This revision is for dumping uint8_t dtype. Without this change, ostream
would print uint8_t as character.


seanshpark (approved):
LGTM

PR #4114, opened 2020-09-07
[nnfwapi] Test : support other tensor types

wateret:
Support more tensor types in GenModelTest.

- Introduce helper `uniformTCD` for generating test cases with all same
  data types of model inputs/outputs
- Now `TestCaseData` has buffers with `uint8_t` which indicates that
  those are byte buffers(it does not mean uint8)
- Fix the workarounds in Rank tests

ONE-DCO-1.0-Signed-off-by: Hanjoung Lee <hanjoung.lee@samsung.com>

PR #4127, opened 2020-09-08
[onert] Fix Split operation input index

hseok-oh:
Split operation's 1st operand is axis, and 2nd operand is input tensor

---

Related issue: #4116


hseok-oh:
We can add positive test for this after #4114 . And I found that it's not easy to add negative test because `OperationValidator` is called after static shape inference.

ragmani (approved):
LGTM

hseok-oh:

wateret (approved):
LGTM

PR #4128, opened 2020-09-08
[DRAFT][onert] introduce 16x1 block sparsity

glistening:
It will introduce block sparsity with 16x1 block size.
It includes the change of loader, ir, layer, cpu backend kernel.


glistening:
I've succeed to run my model with draft PR (= 16x1 block sparsity kernel).
I will push smaller PRs from this draft.

PR #4130, opened 2020-09-08
[luci-interpreter]Negative Tests for DepthwiseConv2D kernel

struss:
For #1623 

This commit add negative tests on `DepthwiseConv2D` kernel on luci-interpreter.


jinevening (approved):
LGTM

(optional) There exist a lot of duplicate codes. It would be better to make a class that creates a DepthwiseConv2D operator with different settings to reduce code size.

struss:
Oh I got it. I will update test codes with single function that check configure phase with different data. Thanks!

s-barannikov (approved):
LGTM

PR #4136, opened 2020-09-08
[onert] Disable dim_correction of input of acl StridedSlice kernels

ragmani:
Acl's StridedSlice kernels checks if input's rank and the number of starts/ends are equal. To prevent error by the checks, this commit disables dim_correction of input of acl StridedSlice kernels.


hseok-oh (approved):
LGTM

wateret (approved):
LGTM

PR #4141, opened 2020-09-08
[luci] Support quantization of const inputs

jinevening:
This supports quantization of constant inputs

### DIFF
--- a/compiler/luci/pass/src/QuantizationUtils.cpp
+++ b/compiler/luci/pass/src/QuantizationUtils.cpp
@@ -36,7 +36,6 @@ void asymmetric_wquant_with_minmax_per_layer(CircleConst *node, float min, float
                                              float &scaling_factor, int64_t &zp, float &nudged_min,
                                              float &nudged_max)
 {
-
   const int32_t kMinScale = 0;
   const int32_t kMaxScale = 255;
 
--- a/compiler/luci/pass/src/QuantizationUtils.cpp
+++ b/compiler/luci/pass/src/QuantizationUtils.cpp
@@ -31,6 +31,7 @@ uint8_t fp32_to_uint8_cast(float f)
   return static_cast<uint8_t>(f);
 }
 
+// Per-layer quantization of weights (const tensor) using given min/max values
 void asymmetric_wquant_with_minmax_per_layer(CircleConst *node, float min, float max,
                                              float &scaling_factor, int64_t &zp, float &nudged_min,
                                              float &nudged_max)
--- a/compiler/luci/pass/src/QuantizationUtils.cpp
+++ b/compiler/luci/pass/src/QuantizationUtils.cpp
@@ -31,6 +31,36 @@ uint8_t fp32_to_uint8_cast(float f)
   return static_cast<uint8_t>(f);
 }
 
+void asymmetric_wquant_with_minmax_per_layer(CircleConst *node, float min, float max,
+                                             float &scaling_factor, int64_t &zp, float &nudged_min,
+                                             float &nudged_max)
+{
+
+  const int32_t kMinScale = 0;
+  const int32_t kMaxScale = 255;
+
+  uint32_t size = node->size<loco::DataType::FLOAT32>();
+  compute_asym_scale_zp(min, max, scaling_factor, zp, nudged_min, nudged_max);
+  const float scaling_factor_inv = 1.0 / scaling_factor;
+  std::vector<int32_t> quantized_values(size);
+  for (uint32_t i = 0; i < size; ++i)
+  {
+    // clipping
+    auto data = node->at<loco::DataType::FLOAT32>(i);
+    data = data < nudged_min ? nudged_min : data;
+    data = data > nudged_max ? nudged_max : data;
+    quantized_values[i] =
+        static_cast<int32_t>(std::round((data - nudged_min) * scaling_factor_inv));
+  }
+
+  node->dtype(loco::DataType::U8);      // change the type of tensor
+  node->size<loco::DataType::U8>(size); // resize tensor
+  for (uint32_t i = 0; i < size; ++i)
+  {
+    node->at<loco::DataType::U8>(i) = std::min(kMaxScale, std::max(kMinScale, quantized_values[i]));
+  }
+}
+
 void compute_sym_scale_zp(float min, float max, float &scaling_factor, int64_t &zp,
                           float &nudged_min, float &nudged_max)
 {
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -585,54 +585,73 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
 };
 
 /**
- * @brief QuantizeConstInputs quantizes const input tensors
- * @details Quantize values using min/max of const values
+ * @brief Quantize const input tensors using min/max of const values
  */
-struct QuantizeConstInputs final : public luci::CircleNodeMutableVisitor<bool>
+void quantize_const_inputs(luci::CircleNode *node, loco::DataType output_type)
 {
-  QuantizeConstInputs(loco::DataType input, loco::DataType output)
-      : input_type(input), output_type(output)
+  auto opcode = node->opcode();
+  auto arity = node->arity();
+
+  loco::Node *input_node{nullptr};
+  luci::CircleConst *const_node{nullptr};
+
+  switch (opcode)
   {
-  }
+    case luci::CircleOpcode::CONV_2D:
+    case luci::CircleOpcode::DEPTHWISE_CONV_2D:
+    case luci::CircleOpcode::FULLY_CONNECTED:
+    case luci::CircleOpcode::TRANSPOSE_CONV:
+      // Handled in QuantizeWeights and QuantizeBias
+      break;
 
-  loco::DataType input_type;
-  loco::DataType output_type;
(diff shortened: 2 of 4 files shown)


jinevening:
This PR can solve [this issue](https://github.com/Samsung/ONE/pull/4064#discussion_r483368658) (throwing an exception if const tensors are not correctly quantized)

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -557,6 +584,117 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
+ */
+struct QuantizeConstInputs final : public luci::CircleNodeMutableVisitor<bool>
+{
+  QuantizeConstInputs(loco::DataType input, loco::DataType output)
+      : input_type(input), output_type(output)
+  {
+  }
+
+  loco::DataType input_type;
+  loco::DataType output_type;
```
s-barannikov, line 599: Do they have to be `public`?

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -557,6 +584,117 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
+  {
+    LOGGER(l);
+    INFO(l) << "QuantizeConstInputs visit node: " << node->name() << std::endl;
+    auto opcode = node->opcode();
+    auto arity = node->arity();
+
+    loco::Node *input_node{nullptr};
+    luci::CircleConst *const_node{nullptr};
+
+    switch (opcode)
```
s-barannikov, line 611: It's quite strange to have a `switch` inside a visitor 😄  Either of them is superflous.

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -31,6 +31,36 @@ uint8_t fp32_to_uint8_cast(float f)
   return static_cast<uint8_t>(f);
 }
 
+void asymmetric_wquant_with_minmax_per_layer(CircleConst *node, float min, float max,
```
s-barannikov, line 34: Could you please write a brief comment what this function is supposed to do? I coulnd't grasp it from the name.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: I've changed the code not to use visitor pattern.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: I've changed the struct to the function (not to use visitor pattern).

compiler/luci/pass/src/QuantizationUtils.cpp
jinevening: This function was used to quantize weights in `QuantizeDequantizeWeightsPass.cpp`. It can be used to quantize a constant tensor, so I moved the function to `QuantizationUtils.h` and `QuantizationUtils.cpp` to use it here. 

I've added a brief comment on the function.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -557,6 +584,101 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
+  loco::Node *input_node{nullptr};
+  luci::CircleConst *const_node{nullptr};
+
+  switch (opcode)
+  {
+    case luci::CircleOpcode::CONV_2D:
+    case luci::CircleOpcode::DEPTHWISE_CONV_2D:
+    case luci::CircleOpcode::FULLY_CONNECTED:
+    case luci::CircleOpcode::TRANSPOSE_CONV:
+      // Handled in QuantizeWeights and QuantizeBias
```
s-barannikov, line 604: Do the weights and the bias differ significantly that they are not handled here?

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -557,6 +584,101 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
+      break;
+
+    case luci::CircleOpcode::CONCATENATION:
+      // Handled in propagate_concat_quantparam
```
s-barannikov, line 608: Same question for the concatenation inputs.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -557,6 +584,101 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
+          else
+          {
+            throw std::runtime_error("NYI data type");
+          }
+        }
+      }
+      break;
+
+    default:
+      for (uint32_t i = 0; i < arity; i++)
```
s-barannikov, line 671: May be it is worth to list operators with "default" behavior explicitly and throw an exception in the "default:" case.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
s-barannikov: Thanks. Personally I'd go with visitor (I don't like big functions and especially long switchs).

jinevening: I'd like to use a function here because implementing a visitor for each operator would make lots of duplicate codes.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: Well, I think so. We have to consider something more to quantize them (e.g., granularity (lwq, cwq), weight clipping (maybe implemented in the future), different data type and quantization timing for bias (int32)).

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: Since we need to check several conditions to propagate concat qparam, I think it is better to be handled separately.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: The 'default' behavior to quantize a const tensor would be quantizing with the const values (implemented in `quant_const` function). But it should be applied according to opcode.

I'd like to make this function work with the allow list because I don't fully understand the behavior of all operators. Currently, the listed operators are only the ones I'm familiar with. It will be possible to add operators in this switch statement as needed.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: Current implementation of `propagate_concat_quantparam` does not quantize const inputs in some cases, e.g., if concat has a fused activation function, const inputs are not quantized now. To fix this, I'll make another PR after this PR is landed.

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -31,6 +31,37 @@ uint8_t fp32_to_uint8_cast(float f)
   return static_cast<uint8_t>(f);
 }
 
+// Per-layer quantization of weights (const tensor) using given min/max values
+void asymmetric_wquant_with_minmax_per_layer(CircleConst *node, float min, float max,
+                                             float &scaling_factor, int64_t &zp, float &nudged_min,
+                                             float &nudged_max)
+{
+
```
seanshpark, line 39: maybe addition line?

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
s-barannikov: What I'm suggesting is to throw exception "by default", rather that quantize all operator inputs. It would allow to catch operators that are not currently white-listed.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
s-barannikov: Just thought that everyting in one place would be better :)

s-barannikov (dismissed):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: > What I'm suggesting is to throw exception "by default", rather that quantize all operator inputs. It would allow to catch operators that are not currently white-listed.

I think the current code is doing the same thing. Exception is being thrown in the `default:` case when unsupported Op (Ops not in the whitelist) has a const input. In the default case, inputs are not quantized.

compiler/luci/pass/src/QuantizationUtils.cpp
jinevening: I removed the line. Thanks.

jinevening:
@s-barannikov @seanshpark  PTAL

seanshpark (approved):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
s-barannikov: I see now, thank you.

s-barannikov (approved):

PR #4142, opened 2020-09-08
[luci] Prevent quantizing bias of TCONV Op

jinevening:
This prevents quantizing bias of TCONV Op


seanshpark (approved):
LGTM

PR #4146, opened 2020-09-08
[onert] Remove release Tensors

wateret:
Remove tensor manager moves. We don't have to move tensor managers from
TensorBuilder to Executor since TensorBuilder is also alive until the
session destroys.


hyunsik-yoon (approved):
LGTM

ragmani (approved):
LGTM

PR #4147, opened 2020-09-08
[luci-interpreter]Negative Tests for Elu kernel

struss:
For #1623 

This commit add negative tests on `Elu` kernel on luci-interpreter.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Elu.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Elu.test.cpp
@@ -73,21 +73,6 @@ TEST(EluTest, InOutTypeMismatch_NEG)
   EXPECT_ANY_THROW(kernel.configure());
 }
 
-TEST(EluTest, InvalidType_NEG)
-{
-  Shape input_shape{1, 2, 4, 1};
-  std::vector<uint8_t> input_data{
-      0, 6, 2,  4, //
-      3, 2, 10, 1, //
-  };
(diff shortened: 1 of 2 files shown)


compiler/luci-interpreter/src/kernels/Elu.test.cpp
```diff
@@ -59,6 +59,35 @@ TEST(EluTest, SimpleElu)
...
+      3, -2, 10, -0.1, //
+  };
+  Tensor input_tensor = makeInputTensor<DataType::FLOAT32>(input_shape, input_data);
+  Tensor output_tensor = makeOutputTensor(DataType::U8);
+
+  Elu kernel(&input_tensor, &output_tensor);
+  EXPECT_ANY_THROW(kernel.configure());
+}
+
+TEST(EluTest, InvalidType_NEG)
```
struss, line 76: @s-barannikov 
a little bit question about this. Currently not support for this type, but later, this would be enabled for other types.
Should I leave this or remove this?

s-barannikov: I'd remove it. IMHO negative tests should cover _invalid_ cases, not _unsupported_.
The difference between _invalid_ and _unsupported_ can be vague, but this one is not the case :)

struss: Okay I will remove this! Thanks for comment!

jinevening (approved):
LGTM

s-barannikov (approved):
LGTM

PR #4150, opened 2020-09-08
[luci-interpreter] Add support for relu activation

ai-moiseev:
Add support for relu activation op in luci-interpreter.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Relu.cpp
+++ b/compiler/luci-interpreter/src/kernels/Relu.cpp
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "kernels/Relu.h"
+#include "kernels/Utils.h"
+
+#include <tensorflow/lite/kernels/internal/optimized/optimized_ops.h>
+
+#include <stdexcept>
+
+namespace luci_interpreter
+{
+
+namespace kernels
+{
+
+Relu::Relu(const Tensor *input, Tensor *output) : Kernel({input}, {output}) {}
+
+void Relu::configure()
+{
+  LUCI_INTERPRETER_CHECK(input()->element_type() == output()->element_type());
+
+  if (input()->element_type() == DataType::U8)
+  {
+    double multiplier = input()->scale() / output()->scale();
+    quantizeMultiplier(multiplier, &_output_multiplier, &_output_shift);
+  }
+  output()->resize(input()->shape());
+}
+
+void Relu::execute() const
+{
+  switch (input()->element_type())
+  {
+    case DataType::FLOAT32:
+      evalFloat();
+      break;
+    case DataType::U8:
+      evalQuantized();
+      break;
+    default:
+      throw std::runtime_error("Unsupported type.");
+  }
+}
+
+void Relu::evalFloat() const
+{
+  const auto input_data = getTensorData<float>(input());
+  const auto input_shape = getTensorShape(input());
+  auto output_data = getTensorData<float>(output());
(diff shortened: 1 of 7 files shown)


compiler/luci-interpreter/src/kernels/Relu.cpp
```diff
@@ -0,0 +1,77 @@
...
+void Relu::evalQuantized() const
+{
+  const auto input_data = getTensorData<uint8_t>(input());
+  const auto input_zero_point = input()->zero_point();
+  const auto size = getTensorShape(input()).FlatSize();
+  auto output_data = getTensorData<uint8_t>(output());
+
+  for (auto i = decltype(size){0}; i < size; ++i)
+  {
+    output_data[i] = input_data[i] >= input_zero_point ? input_data[i] : input_zero_point;
```
s-barannikov, line 72: This does not seem what TFLite does in U8 case.
IIUC, input and output quantization parameters do not have to match. If they are different, the value should be "requantized".
Please check out `ReluPrepare` and `ReluEval` in `lite/kernels/activations.cc`.

compiler/luci-interpreter/src/kernels/Relu.test.cpp
```diff
@@ -0,0 +1,117 @@
...
+{
+namespace kernels
+{
+namespace
+{
+
+using namespace testing;
+
+template <typename T>
+void Check(std::initializer_list<int32_t> input_shape, std::initializer_list<int32_t> output_shape,
```
s-barannikov, line 31: In other tests, this function is used to check several data types (typename T). You use it only for `float`, so why don't just inline it?)

compiler/luci-interpreter/src/kernels/Relu.test.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  Relu kernel(&input_tensor, &output_tensor);
+  kernel.configure();
+  kernel.execute();
+
+  EXPECT_THAT(dequantize(extractTensorData<uint8_t>(output_tensor), output_tensor.scale(),
+                         output_tensor.zero_point()),
+              ElementsAreArray(ArrayFloatNear(ref_output_data, kQuantizedTolerance)));
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray({1, 2, 3, 1}));
+
```
jinevening, line 92: Can you add the test for the quantized value for uint8 kernels (e.g., like [logsoftmax](https://github.com/Samsung/ONE/blob/master/compiler/luci-interpreter/src/kernels/LogSoftmax.test.cpp#L81-L82))?

If it is difficult to make reference data, using the test data in [TFLite](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/activations_test.cc#L661-L682) will be ok.

compiler/luci-interpreter/src/kernels/Relu.test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  SUCCEED();
```
jinevening, line 93: Seems unnecessary.

compiler/luci-interpreter/src/kernels/Relu.cpp
ai-moiseev: Done, replaced with call to quantized ReluX method. Also added test-case for re-quantization.

compiler/luci-interpreter/src/kernels/Relu.test.cpp
ai-moiseev: Done, inlined to test case body.

compiler/luci-interpreter/src/kernels/Relu.test.cpp
ai-moiseev: Done, I've added two `uint8_t` value checking test cases.

compiler/luci-interpreter/src/kernels/Relu.test.cpp
ai-moiseev: Done, removed it. Was taken from other test cases, thought it was some kind of convention :)

s-barannikov: It is used for static analysis somehow. If the test body does not contain any of "EXPECT_*", "ASSERT_*", then the test is considered invalid, despite of the fact that they could be called function called from the test.

compiler/luci-interpreter/src/kernels/Relu.test.cpp
s-barannikov: It seems unused now -_-

jinevening (dismissed):
LGTM (`Check` function in the test code needs to be removed)

ai-moiseev:
> LGTM (`Check` function in the test code needs to be removed)

@jinevening Done, removed `Check` from `*.test.cpp`

jinevening (approved):

s-barannikov (approved):

Issue #4157, opened 2020-09-09
[onert/Android] NDK's C++ dynamic_cast support


wateret:
To recap,

- On Android, `dynamic_cast` always have been working
- However it has an issue when **Across Library Boundaries** (looks like it is an issue of Android NDK/Loader, not 100% sure)
- To workaround it, we must ensure a key function exists for the classes

---

We had some problems(a couple of times) `dynamic_cast` not working properly on Android.

And we assumed that `dynamic_cast` is not supported on Android. However [the official document](https://developer.android.com/ndk/guides/cpp-support#rt) says

> As with exceptions, RTTI is supported by libc++, but is disabled by default in ndk-build. CMake and standalone toolchains have RTTI enabled by default.

With CMake, RTTI is enabled. And I double-checked typeinfo is generated by:

```
$ nm -C libonert_core.so | ack 'typeinfo'
00000000035c190 V typeinfo for long (long)
0000000000357378 V typeinfo for onert::exec::IExecutor* (std::__ndk1::unique_ptr<onert::compiler::LoweredGraph, std::__ndk1::default_delete<onert::compiler::LoweredGraph> >, onert::compiler::CompilerOptions const&, std::__ndk1::shared_ptr<std::__ndk1::unordered_map<onert::util::Index<unsigned int, onert::ir::SubgraphIndexTag>, std::__ndk1::unique_ptr<onert::exec::IExecutor, std::__ndk1::default_delete<onert::exec::IExecutor> >, std::__ndk1::hash<onert::util::Index<unsigned int, onert::ir::SubgraphIndexTag> >, std::__ndk1::equal_to<onert::util::Index<unsigned int, onert::ir::SubgraphIndexTag> >, std::__ndk1::allocator<std::__ndk1::pair<onert::util::Index<unsigned int, onert::ir::SubgraphIndexTag> const, std::__ndk1::unique_ptr<onert::exec::IExecutor, std::__ndk1::default_delete<onert::exec::IExecutor> > > > > > const&)
000000000035b180 V typeinfo for void (onert::interp::ExecEnv const*, onert::ir::Operation const&)
000000000035b0d0 V typeinfo for void (onert::interp::ExecEnv*, onert::ir::Operation const&)
0000000000354140 V typeinfo for void (onert::ir::Operand const&, onert::backend::ITensor&)
...
```

And then I found this.

[...]

PR #4158, opened 2020-09-09
[onert] Add destructor to cpu_common::Tensor

wateret:
As it is used as a base class, having a virtual destructor makes it
safer.

It also resolves the issue on Android, as it creates a key function by
defining it as non-inline.

Related : #4157

### DIFF
--- a/runtime/onert/core/include/backend/cpu_common/Tensor.h
+++ b/runtime/onert/core/include/backend/cpu_common/Tensor.h
@@ -33,6 +33,7 @@ class Tensor : public IPortableTensor
 {
 public:
   Tensor() = delete;
+  virtual ~Tensor();
 
 public:
   Tensor(const ir::OperandInfo &info, const ir::Layout layout,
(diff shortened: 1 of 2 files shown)


chunseoklee (approved):

hseok-oh (approved):

ragmani (approved):
LGTM

s-barannikov:
Should not the destructor be in IPortableTensor instead?

wateret:
@s-barannikov Are you talking about "key function" from the issue #4157?

s-barannikov:
@wateret As far as I know it is sufficient to add a virtual destructor to the very base class, and all the inherited classes will have virtual destructors automatically generated. `IPortableTensor` already has virtual destructor, so from this point of view this change is redundant.
If the change helps to resolve #4157, that's nice.

wateret:
@s-barannikov 

> As far as I know it is sufficient to add a virtual destructor to the very base class, and all the inherited classes will have virtual destructors automatically generated.

Oh I didn't know about this 😃

wateret:
So the virtual destructor in `IPortableTensor` would be redundant.

s-barannikov:
> So the virtual destructor in `IPortableTensor` would be redundant.

Yes, and in all other classes directly or indirectly derived from ITensor. But I don't know if this is true from the point of view of the issue with across-library calls you mentioned.

PR #4162, opened 2020-09-09
[bcq-tools] Modify unique value generator

llFreetimell:
If `new_shape` attribute of `Reshape` operation is `[1, -1]`, preserved `do_w_x` can be duplicated with it and it can cause
unintended conversion.
This commit will fix it by modifying the starting value to -1500000001, which is prime number.

### DIFF
--- a/compiler/bcq-tools/preserve_bcq_info
+++ b/compiler/bcq-tools/preserve_bcq_info
@@ -50,7 +50,7 @@ def preserve_bcq_info(flags):
 
     We use negative values to preserve BCQ information because
     positive values may cause some confusion with real BCQ information values.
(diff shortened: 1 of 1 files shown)


seanshpark:
> which is prime number.

Just curiois... `-1500000001` happend to be a prime number?

llFreetimell:
@seanshpark 
Just to decrease probability of value duplication, I selected `-1500000001` among big prime numbers... @.@
I thought that "-1500000001 will not be duplicated because it cannot be obtained by multiplication and it is too big..!"
As this value is not mandatory you can use another value! :smile:

seanshpark (approved):
LGTM

PR #4173, opened 2020-09-10
[luci-interpreter]Add negative cases for If kernel

struss:
For #1623 

This commit add negative cases for  `If` kernel on luci-interpreter.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/If.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/If.test.cpp
@@ -106,6 +106,36 @@ TEST(IfTest, CondFalse)
   EXPECT_THAT(extractTensorData<float>(output), ElementsAreArray(ArrayFloatNear({5, 14})));
 }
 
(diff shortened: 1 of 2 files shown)


compiler/luci-interpreter/src/kernels/If.test.cpp
```diff
@@ -88,9 +88,9 @@ TEST(IfTest, CondTrue)
   EXPECT_THAT(extractTensorData<float>(output), ElementsAreArray(ArrayFloatNear({6, 9})));
 }
 
-TEST(IfTest, CondFalse)
```
jinevening, line 91: It seems that this test (`CondFalse`) is removed. Is it intended?

struss: Oh, That's my mistake. I will fix it.

jinevening (approved):
LGTM

s-barannikov (approved):
LGTM

PR #4178, opened 2020-09-10
[onert] Revise DynamicMemoryManager

wateret:
Change key type to pointer as we are changing it to work without index.

Draft : #4175


hyunsik-yoon (approved):
LGTM

Issue #3889, opened 2020-08-20
Compiler FE: Support SparsityParameters


mhs4670go:
Let's support `SparsityParameters` in `Tensors`.

seanshpark:
done ?

mhs4670go:
@seanshpark I'll closed after implementing sparsifier.

PR #4190, opened 2020-09-10
[DRAFT] Support Sparse tensor

mhs4670go:
This draft is to support Sparse tensor.

Related : #3889


compiler/luci/import/src/CircleReader.cpp
```diff
@@ -193,6 +246,14 @@ void copy_tensor_attributes(const circle::TensorT &tensor, CircleNode *node)
     if (quantparam)
       node->quantparam(std::move(quantparam));
   }
+
+  const auto *sparsity = tensor.sparsity.get();
+  if (sparsity != nullptr)
+  {
+    auto sparsityparam = luci_sparsityparam(sparsity);
+    if (sparsity)
```
seanshpark, line 254: `sparsity` is not null here :)

PR #4194, opened 2020-09-10
[onert] Move `applyShape` to `ITensor`

wateret:
Move `applyShape` to `ITensor`.

- Replace `DynamicTensorManager::applyShape` with `ITensor::applyShape`
- `cpu_common::Tensor` now has `DynamicMemoryManager` rather than
  `DynamicTensorManager`
- Remove unused DynamicTensorManager in `controlflow::UserTensor`

After this, we can finally remove `DynAllocInfo` stuff.

Draft : #4175

### DIFF
--- a/runtime/onert/core/src/backend/controlflow/UserTensor.cc
+++ b/runtime/onert/core/src/backend/controlflow/UserTensor.cc
@@ -16,6 +16,9 @@
 
 #include "UserTensor.h"
 
+#include "util/Exceptions.h"
+#include "ir/DataType.h"
+
 namespace onert
 {
 namespace backend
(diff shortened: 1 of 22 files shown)


runtime/onert/core/src/backend/controlflow/UserTensor.cc
```diff
@@ -35,6 +38,16 @@ size_t UserTensor::calcOffset(const ir::Coordinates &coords) const
   return offset;
 }
 
+bool UserTensor::applyShape(const ir::Shape &new_shape)
+{
+  // User tensors cannot be reallocated.
```
hyunsik-yoon, line 43: I wonder whether this is true.
Is there any case where input size increases while calling `nnfw_run()` multiple times?

wateret: This is the description for `UserTensor`.

```cpp
/**
 * @brief Tensor object that is for Input and Output tensors from the user.
 *
 * This class is a wrapped buffer that is allocated by the user. So it does not have resposibility
 * on allocation nor deallocation. All the model input/output tensors are wrapped with this class
 * for execution.
 *
 */
```

This tensor is just a wrapper of a buffer given by `nnfw_set_input/output`, so the buffer can never be changed during execution.

wateret: Are you talking about `nnfw_run` could make some changes the input tensors? We don't have such models yet and not sure those models are useful. Even if those cases exist, my answer would still be the same.

hyunsik-yoon: I see. Thanks.

hyunsik-yoon (approved):
LGTM

PR #4219, opened 2020-09-11
[luci] Quantize const input of concat

jinevening:
This quantizes const input of concat

### DIFF
--- a/compiler/luci/pass/src/PropagateConcatenationQparam.test.cpp
+++ b/compiler/luci/pass/src/PropagateConcatenationQparam.test.cpp
@@ -37,7 +37,7 @@ void addQuantParam(luci::CircleNode &node, const std::vector<float> &scale,
   node.quantparam(std::move(quantparam));
 }
 
-int quantize(float f, luci::CircleQuantParam *qparam)
+int32_t quantize(float f, luci::CircleQuantParam *qparam)
 {
   float scale = qparam->scale[0];
   int64_t zp = qparam->zerop[0];
--- a/compiler/luci/pass/src/PropagateConcatenationQparam.test.cpp
+++ b/compiler/luci/pass/src/PropagateConcatenationQparam.test.cpp
@@ -18,6 +18,7 @@
 
 #include <luci/IR/CircleQuantParam.h>
 
+#include <math.h>
 #include <vector>
 
 #include <gtest/gtest.h>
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/PropagateConcatenationQparam.test.cpp
```diff
@@ -36,6 +37,14 @@ void addQuantParam(luci::CircleNode &node, const std::vector<float> &scale,
   node.quantparam(std::move(quantparam));
 }
 
+int quantize(float f, luci::CircleQuantParam *qparam)
```
seanshpark, line 40: return type is `int`. in body `zp` is `int64_t`. Maybe use same type `int64_t`(or `int32_t`) for consistancy?

compiler/luci/pass/src/PropagateConcatenationQparam.test.cpp
```diff
@@ -195,22 +203,11 @@ TEST(PropagateConcatenationQparam, propagate_concat_quantparam)
...
-  g.concat_node.dtype(loco::DataType::S8);
-  luci::propagate_concat_quantparam(&g.concat_node);
-  EXPECT_FLOAT_EQ(g.concat_node.quantparam()->scale[0], 3.14);
-  EXPECT_EQ(g.concat_node.quantparam()->zerop[0], 77);
-  EXPECT_FLOAT_EQ(g.input_1.quantparam()->scale[0], 1.0);
-  EXPECT_EQ(g.input_1.quantparam()->zerop[0], 1);
-  EXPECT_FLOAT_EQ(g.input_2.quantparam()->scale[0], 2.0);
-  EXPECT_EQ(g.input_2.quantparam()->zerop[0], 2);
-  g.concat_node.dtype(loco::DataType::U8);
-
```
jinevening, line 213: I removed this test case because `propagate_concat_quantparam` is now part of `QunatizeWithMinMaxPass`, so dtype of concat is always uint8 (I added `assert` for that condition in [here ](https://github.com/Samsung/ONE/pull/4219/files#diff-40cdd42b1975a4d2a2cc451a01dbebe9R712)).

compiler/luci/pass/src/PropagateConcatenationQparam.test.cpp
jinevening: I've changed it to `int32_t` for consistency. Thanks.

seanshpark (approved):
LGTM

s-barannikov (approved):
LGTM

mhs4670go (approved):
LGTM

PR #4220, opened 2020-09-11
[release_tool] fix minor bug

mhs4670go:
This commit fixes minor bug of git_release script.

When we upload an asset, the name of the uploaded asset should be a
basename.


hseok-oh (approved):
LGTM

PR #4221, opened 2020-09-11
[res] TFLiteRecipes for Ops with const input

jinevening:
This adds TFLiteRecipes for Ops with const input


jinevening:
These recipes are Ops where one of its input is a constant tensor. This will be used for testing quantization of constant inputs.

seanshpark (approved):
LGTM

PR #4225, opened 2020-09-11
[record-minmax] Generate random inputs including neg numbers

jinevening:
This generates random inputs including negative numbers

### DIFF
--- a/compiler/record-minmax-conversion-test/gen_h5_random_inputs.py
+++ b/compiler/record-minmax-conversion-test/gen_h5_random_inputs.py
@@ -39,8 +39,8 @@ for i in range(num_data):
 
     for j in range(len(input_details)):
         input_detail = input_details[j]
-        input_data = np.array(
-            np.random.random_sample(input_detail["shape"]), input_detail["dtype"])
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

compiler/record-minmax-conversion-test/gen_h5_random_inputs.py
```diff
@@ -39,8 +39,8 @@
 
     for j in range(len(input_details)):
         input_detail = input_details[j]
-        input_data = np.array(
-            np.random.random_sample(input_detail["shape"]), input_detail["dtype"])
+        input_data = np.array(10 * np.random.random_sample(input_detail["shape"]) - 5,
```
mhs4670go, line 42: It would be better to add comment that shows the range of its random data. Maybe [-5, 5]?

jinevening: Added comment. Thanks.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #4228, opened 2020-09-11
[onert] Argmax acl cl int64

dr-venkman:
Support int64 output type for Argmax on acl/cl backend.

ONE-DCO-1.0-Signed-off-by: venkat.iyer <venkat.iyer@samsung.com>


hseok-oh:
Please add unittest.

dr-venkman:
I tried extending the testcase to handle a new `TENSOR_INT64` type. However, it fails to generate a suitable `argmax_float_x_nnfw.example.cpp` upon compilation. Adding int64 testcases seems a non-trivial task. Could you please let me know if there are testcases existing already that deal with int64? Alternatively, would it help to open an issue and reference it here?

hseok-oh:
@dr-venkman Use `one_op_tests` (`tests/nnfw_api/src/one_op_tests`), not nnapi test. NNAPI don't have INT64 type. There is no INT64 type example, but you can extend type easily than NNAPI.

dr-venkman:
> @dr-venkman Use `one_op_tests` (`tests/nnfw_api/src/one_op_tests`), not nnapi test. NNAPI don't have INT64 type. There is no INT64 type example yet, but you can extend type easily than NNAPI.

@hseok-oh , Thanks, PTAL.

tests/nnfw_api/src/one_op_tests/ArgMax.cc
```diff
@@ -40,6 +40,28 @@ TEST_F(GenModelTest, OneOp_ArgMax_AxisToConst)
   SUCCEED();
 }
 
+TEST_F(GenModelTest, OneOp_ArgMax64_AxisToConst)
```
hseok-oh, line 43: Please change name `OneOp_ArgMax64_AxisToConst` to `OneOp_ArgMax_Int64_AxisToConst`

hseok-oh:
@ragmani Could you check this PR?

compute/ARMComputeEx/src/runtime/CL/functions/CLArgMinMaxLayerEx.cpp
```diff
@@ -0,0 +1,205 @@
+/*
```
intom, line 1: please add copyright

tests/nnfw_api/src/one_op_tests/ArgMax.cc
dr-venkman: Thanks for the comment, I have modified accordingly.

compute/ARMComputeEx/src/runtime/CL/functions/CLArgMinMaxLayerEx.cpp
dr-venkman: Added to all modified files under armcomputeex, kindly review. Thanks.

hseok-oh (approved):

intom (approved):
LGTM

PR #4229, opened 2020-09-11
[onert] add sparse fullyconnected 16x1 kernel

glistening:
It adds fullyconnected 16x1 bcsr kernel based on lpcnet code.

### DIFF
--- a/compute/cker/include/cker/operation/FullyConnectedSparse16x1.h
+++ b/compute/cker/include/cker/operation/FullyConnectedSparse16x1.h
@@ -1,0 +1,125 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


compute/cker/include/cker/operation/FullyConnectedSparse16x1.h
```diff
@@ -0,0 +1,125 @@
...
+  {
+    VectorBatchVectorAssign(bias_data, output_depth, batches, output_data);
+  }
+  else
+  {
+    ZeroVector(output_data, batches * output_depth);
+  }
+  for (int b = 0; b < batches; ++b)
+  {
+    for (int idx_0 = 0; idx_0 < (output_depth / 16); ++idx_0)
```
YiHyunJin, line 87: how about this?

```
int depth_size = output_depth / 16;

for (int b = 0; b < batches; ++b)
  {
    for (int idx_0 = 0; idx_0 < depth_size ; ++idx_0)
..
..
```

glistening: @YiHyunJin I believe it makes no difference since the compiler will optimize it for us. However, I will change.

wateret:
@glistening Could you add a test for this? (now or later)

glistening:
@wateret Yes, I will. I've checked the correctness by running the target model. And I can create nnpkgtc right now. But for `CircleGen`, I requires more changes and work for supporting `sparsity` in circlegen and writing testcase in c++. It can be done in another PRs.

wateret (approved):
LGTM

glistening:
@YiHyunJin PTAL

YiHyunJin (approved):
LGTM

PR #4242, opened 2020-09-14
[onert] add sparse fullyconnected 16x1 kernel

glistening:
It adds fullyconnected 16x1 bcsr kernel based on lpcnet code.


glistening:
It is a part of #4128.

wateret (approved):
LGTM

glistening:
I've pushed this PR since it is not in master. I've found that I've already pushed the same PR, it was pending.

PR #4256, opened 2020-09-14
[onert/docs] Write backend-api.md

wateret:
This commit fills out document `backend-api.md`.

### DIFF
--- a/docs/runtime/api.md
+++ b/docs/runtime/api.md
@@ -28,6 +28,6 @@ For usage, refer to [Howto : NNFW API](../howto/how-to-use-nnfw-api.md).
 
 ## Backend API
 
-We provide Backend API to allow anyone to implement so one can extend the runtime in terms of operation computation.
+Backend API enables anyone to extend the runtime in terms of operation computation and memory management.
 
 For detailed descriptions, refer to [Backend API](../runtime/backend-api.md).
--- a/docs/runtime/api.md
+++ b/docs/runtime/api.md
@@ -28,8 +28,6 @@ For usage, refer to [Howto : NNFW API](../howto/how-to-use-nnfw-api.md).
 
 ## Backend API
 
-Backend API is defined by One Runtime.
-
-Backend API is about actual computation of operations and memory management for operands. In order to allow different kinds of computation units or computation libraries, One Runtime defines Backend API to support user defined operation kernels and memory manager. It contains a lot of C++ headers which are subject to change.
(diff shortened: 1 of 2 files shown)


hseok-oh (dismissed):
LGTM

docs/runtime/api.md
```diff
@@ -28,8 +28,6 @@ For usage, refer to [Howto : NNFW API](../howto/how-to-use-nnfw-api.md).
 
 ## Backend API
 
-Backend API is defined by One Runtime.
-
-Backend API is about actual computation of operations and memory management for operands. In order to allow different kinds of computation units or computation libraries, One Runtime defines Backend API to support user defined operation kernels and memory manager. It contains a lot of C++ headers which are subject to change.
+We provide Backend API to allow anyone to implement so one can extend the runtime in terms of operation computation.
```
glistening, line 31: @wateret 

- What do you mean by `so` here? Shared library?
- Also, the explanation here is different from `backend-api.md`. It does not say about memory management. Which one is correct? Here or `backend-api.md`?

wateret: - "so" here is an adverb, please let me know it is grammatically wrong or have a better sentence
- Both are correct. If I should summarize "Backend" as a one word, it would be "operation computation". Just wanted to keep this sentence short

I changed this part to avoid duplication, I could just remove the whole sentence 😄  or please suggest a sentence.

wateret:
@glistening PTAL :)

docs/runtime/api.md
glistening: >     * "so" here is an adverb, 

Thank you for answer. 

> please let me know it is grammatically wrong or have a better sentence

I don't see grammatical error. I was just confused whether `so` is for `implement` or for following sentence.

> have a better sentence

I think you mean:

Backend API enables anyone to extend the runtime in terms of operation computation.

( I am not sure it is better sentence. I just removed the words that give less information at least to me. )

> Both are correct. If I should summarize "Backend" as a one word, it would be "operation computation".

I thought `computation` and `memory allocation` are different concept like `ComputeLibrary`'s `Core` and `Runtime`. `operation computation` is okay since details will be explained in `backend-api.md`.

wateret:
```suggestion
Backend API enables anyone to extend the runtime in terms of operation computation and memory management.
```

wateret: Updated the sentence, thanks!

hseok-oh (approved):

glistening (approved):
LGTM

PR #4258, opened 2020-09-14
[res] Add FullyConnected network that contains sparse tensor

mhs4670go:
This commit adds FullyConnected network that contains sparse tensor.

Related : #3889
Draft : #4190


seanshpark (approved):
LGTM

Issue #4151, opened 2020-09-09
Compiler FE: Enable UniDirectionalSequenceLSTM Op


struss:
Parent Issue : #1606
Draft PR : #4260 

- [x] res/TensorFlowPythonExamples #4283
- [x] tflchef #4292 #4377
- [x] tflite2circle #4305
- [x] tfldump #4342
- [x] circledump #4343 
- [x] common-artifacts ~#4367~ #4494 
- [x] res/TensorFlowLiteRecipes #4530
- [x] luci.IR #4380
- [x] luci.Import #4502 
- [x] luci.Export #4503 
- [x] luci.logex #4501 
- [x] luci.Service #4504 
- [x] luci.tests #4546

struss:
Reference page
https://www.tensorflow.org/lite/convert/rnn#example

struss:
I think variable tensor and this operator kernel is different work. so, I will just add this kernel with except recipe that contains variable tensor(on common-artifacts level). after add this kernel I think we need to talk about variable tensor.

struss:
Done.... but Currently recipe_001 is not pass a test for circle2circle. that recipe need to add concept of `is_variable` on luci.

PR #4260, opened 2020-09-14
[DRAFT]Enable UnidirectionalSequenceLSTM Draft

struss:
For #4151 

This commit is draft for `UnidirectionalSequenceLSTM` Operation.


compiler/common-artifacts/exclude.lst
```diff
@@ -11,6 +11,7 @@ optimize(Unique_002)
 optimize(Unique_003)
 optimize(Unique_U8_000)
 optimize(Unique_U8_001)
+optimize(UnidirectionalSequenceLSTM_000)
```
seanshpark, line 14: plz fix sort

struss: Okay I will fix it! Thanks.

compiler/luci/import/src/Nodes/CircleUnidirectionalSequenceLSTM.cpp
```diff
@@ -0,0 +1,70 @@
...
+  node->output_gate_bias(inputs.at(15));
+  node->projection_weights(inputs.at(16));
+  node->projection_bias(inputs.at(17));
+  node->activation_state(inputs.at(18));
+  node->cell_state(inputs.at(19));
+  node->input_layer_norm_coefficients(inputs.at(20));
+  node->forget_layer_norm_coefficients(inputs.at(21));
+  node->cell_layer_norm_coefficients(inputs.at(22));
+  node->output_layer_norm_coefficients(inputs.at(23));
+  const auto *options = op.builtin_options.AsUnidirectionalSequenceLSTMOptions();
```
seanshpark, line 62: maybe add an empty line above for readability

compiler/luci/lang/include/luci/IR/Nodes/CircleUnidirectionalSequenceLSTM.h
```diff
@@ -0,0 +1,109 @@
...
+  void cell_clip(float cell_clip) { _cell_clip = cell_clip; }
+  float proj_clip(void) const { return _proj_clip; }
+  void proj_clip(float proj_clip) { _proj_clip = proj_clip; }
+  bool time_major(void) const { return _time_major; }
+  void time_major(bool time_major) { _time_major = time_major; }
+
+private:
+  float _cell_clip = 0.0;
+  float _proj_clip = 0.0;
+  bool _time_major{false};
```
seanshpark, line 104: plz use one style `=` or `{}`. and float is `0.0f`

compiler/tflchef/tflite/src/Op/UnidirectionalSequenceLSTM.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace tflchef
+{
+
+void TFliteOpUnidirectionalSequenceLSTM::filler(const tflite::Operator *op, TFliteImport *import,
+                                                tflchef::ModelRecipe *model_recipe) const
+{
+  // Nothing to do with filler
+}
+tflchef::Operation *
```
seanshpark, line 29: plz add empty line above

compiler/luci/import/src/Nodes/CircleUnidirectionalSequenceLSTM.cpp
struss: Thanks! I will fix it!

compiler/luci/lang/include/luci/IR/Nodes/CircleUnidirectionalSequenceLSTM.h
struss: Thanks! I will fix it! I will use style`=`. sorry for confuse.

compiler/tflchef/tflite/src/Op/UnidirectionalSequenceLSTM.cpp
struss: Thanks! I will fix it!

struss:
currently, working on `res/TensorFlowPythonExamples`

mhs4670go:
@struss I think the recipe that will be merged is same with the example of [here](https://www.tensorflow.org/lite/convert/rnn#from_saved_model). Because that must be validate. Or you can add another one as `*_001.recipe`.

[unidirectional_sequence_lstm.tar.gz](https://github.com/Samsung/ONE/files/5229733/unidirectional_sequence_lstm.tar.gz)
This file is the one that can be made from keras model of above example.

First, below graph is made from the recipe you post here.
image

I have some questions and suggestion about this.

1. Where did you get the attribute list - its name, the number of attributes, etc.?

Here is the code where `CircleUnidirectionalSequenceLSTM` is implemented.
```cxx
loco::Node *input(void) const { return at(0)->node(); }
void input(loco::Node *node) { at(0)->node(node); }

loco::Node *input_to_input_weights(void) const { return at(1)->node(); }
void input_to_input_weights(loco::Node *node) { at(1)->node(node); }
...
loco::Node *cell_layer_norm_coefficients(void) const { return at(22)->node(); }
void cell_layer_norm_coefficients(loco::Node *node) { at(22)->node(node); }
loco::Node *output_layer_norm_coefficients(void) const { return at(23)->node(); }
void output_layer_norm_coefficients(loco::Node *node) { at(23)->node(node); }
```

And, the graph I uploaded is like below.
image

The op in the graph has got 20 inputs, which isn't same with the number of what `CircleUnidirectionalSequenceLSTM` do. There might be optional inputs or sth.

2. I'm not sure but `UnidirectionalSequenceLSTM` often has `ActivationFunctionType_TANH`. Below is the code where TOCO converts it to its tflite(`tensorflow/lite/toco/tflite/operator.cc`).
```cxx
class UnidirectionalSequenceLstm
    : public BuiltinOperator<
          UnidirectionalSequenceLstmOperator,
          ::tflite::UnidirectionalSequenceLSTMOptions,
          ::tflite::BuiltinOptions_UnidirectionalSequenceLSTMOptions> {
 public:
  using BuiltinOperator::BuiltinOperator;
  flatbuffers::Offset<TfLiteOptions> WriteOptions(
      const TocoOperator& op,
      flatbuffers::FlatBufferBuilder* builder) const override {
    // Current toco converter only supports tanh, no clip.
    return ::tflite::CreateUnidirectionalSequenceLSTMOptions(
        *builder, /*fused_activation_function=*/
        ::tflite::ActivationFunctionType_TANH, // !!!!!NOTE HERE
        /*cell_clip=*/0.0,
        /*proj_clip=*/0.0,
        /*time_major=*/true);
  }
```

So, you should add `TANH` value to `Activation` enum.

mhs4670go:
@struss 
3. There are some inputs that can be constants as graph represented. It means there need some implementation of `filler` of this OP.

struss:
> 1. Where did you get the attribute list - its name, the number of attributes, etc.?

I found all reference from `tensorflow/lite/kernels/unidirectional_sequence_lstm.cc` and `lstm_shared.h` on tensorflow version 2.3.
on `lstm_shared.h` comment like [below](https://github.com/tensorflow/tensorflow/blob/ee598066c4cb31ec5ed3106e61ba99ef004a4bae/tensorflow/lite/kernels/lstm_shared.h#L22-L24)
```
// For full inputs kernel (24-inputs).
// Please note the 20-input full kernel is deprecated and only kept
// here for backward compatibility.
```
but, I also agree that recipe for above is needed, so I will add another one as `*_001.recipe`.
> 2. I'm not sure but UnidirectionalSequenceLSTM often has ActivationFunctionType_TANH. Below is the code where TOCO converts it to its tflite(tensorflow/lite/toco/tflite/operator.cc).

Thanks for pointing out. I will fix this.

> 3. There are some inputs that can be constants as graph represented. It means there need some implementation of `filler` of this OP.

Currently, I am checking kernel file in `tensorflow/lite/kernel/unidirectional_sequence_lstm.cc`. If some constants are represented, I will fix all of it. Thanks for pointing out!

struss:
and also add a `TANH` and `SIGN_BIT` value to `FusedActFunc` in luci/AttrFusedActFunc.h

struss:
Activation Are added on PR #4325 and #4327

struss:
> @struss
> 3. There are some inputs that can be constants as graph represented. It means there need some implementation of `filler` of this OP.

I think Tensor inputs except input[0, 17, 18] could be constant because all Tensor except those named as `*_weights` or `*_bias` or `*_Coefficients`(a diagonal matrix).

struss:
while working about this, state for this(input[19] and input[20]) is consist of `variable tensor`.

struss:
Done but need to consider #4396

PR #4261, opened 2020-09-14
[onert] DotDumper : Always show In/Out tensors

### LINKED ISSUE #4094: [onert] DotDumper : Print operands that are not used by any operations
Even with level 2, operands that are not used by any operations are not shown.

For example, in `nnfw_api_gtest`, `GenModelTest.OneOp_If` subgraph 1 and 2 shows nothing. But they both actually have a constant output operand.

wateret:
Always show In/Out tensors, especially when a output tensor is const.

Fix #4094


hseok-oh (approved):
LGTM

PR #4267, opened 2020-09-14
[onert] Throw when output buffer is unset

wateret:
Throw when output buffer is unset and this leads to returning error
when trying to `nnfw_run` but `nnfw_set_output` is not called for an
output.


hyunsik-yoon (approved):
LGTM

hseok-oh (approved):
LGTM

PR #4271, opened 2020-09-15
[onert] Rename "External" to "Migrant"

hyunsik-yoon:
This renames a term `external` to `migrant`.

(We renamed `external tensor` to `migrant tensor` but some had not been renamed.)


wateret (approved):
LGTM

PR #4074, opened 2020-09-03
[Draft] [cker/ruy] Update RUY library package version

periannath:
Update RUY library version used by cker: same with tensorflow 2.3

ONE-DCO-1.0-Signed-off-by: JiHwan Yeo <jihwan.yeo@samsung.com>

---

For issue : #3231

Issue #4047, opened 2020-09-01
[cker/ruy] Always use PrepackedCache for fast inference time


periannath:
Ruy library transforms weight data into packed format since its kernel only accepts packed format.

When some conditions are met, weight data is prepacked at frist run and packed data is used for following runs. If not, weight data is transformed into packed format in every inference.

Prepack can reduce inference time by removing packing phase but it consumes more memory for prepacked weight. Ruy library has cache policies to balance inference time and memory usage.

Now ONE can deallocate original weight after it is prepacekd #2749. Additional memory usage by prepacking is removed in ONE. 

I think we can reduce inference time by packing all the weight used by ruy library. Cache policy to cache every weight is introduced in the latest ruy library, so we need to upgrade ruy library first. 

- [x] Upgrade ruy library #3231 
- [x] Always prepack weight in ruy library #4273

/cc @chunseoklee @YongseopKim @dhdh-oh

periannath:
Done.

Issue #3231, opened 2020-07-16
[cker/ruy] Upgrade ruy library


periannath:
I think it is better to upgrade ruy library to latest version.

#### Pros
- Performance improvement?
- Changes in `ruy::PrepacekdCached` class
  - ~~It seems we can check whether weight is cached or not~~
  - ~~Candidate solution for #3225~~

Source code link and cmake file should be updated to use latest ruy library.

~~I'll upgrade ruy library and checks its performance.~~

Reopens this issue since cache policy in latest ruy library is needed.

- [x] Add cpuinfo library to infra #4248
- [x] Add cpuinfo tar ball to packaging #4250
- [x] Upgrade cmake file for ruy library #4273
- [x] Upgrade ruy tar ball #4274

/cc @YongseopKim

periannath:
I'll use https://github.com/google/ruy/commit/34ea9f4993955fa1ff4eb58e504421806b7f2e8f for ruy library since it is used in tensorflow v2.3.0-rc1.

periannath:
### Changes in `PrepackedCache`

#### Cache key in previous ruy library
```cc
     auto cache_key = std::make_pair(reinterpret_cast<void*>(params->run_kernel),                                                                                                                               
                                     params->src[Side::kLhs].data);  
```
  - Function initialized in ruy library and weight data are used as key

#### Cache key in updated ruy library
```cc
  struct Key {
    // The source matrix's data pointer.
    const void* src_data;
    // The packed matrix's layout, see PEMat::layout.
    PMatLayout packed_layout;
    // The packed matrix's zero point (for integer-quantized matrices only).
    std::int32_t zero_point;
  };
```
  - Weight data, packed matrix layout, and zero point are used as key
  - PackedMatrix created inside ruy library is needed to search cache
  - Still, it is hard to search `PrepackedCache` outside of ruy library...

periannath:
### Changes in cache policy
- Cache policy is changed in recent ruy library

#### Cache policy in previous ruy library
- kNoCache : No cache
- kCacheLHSOnNarrowMul (**default**) : Cache when `batch_size <= 4`

#### Cache policy in updated ruy library
- kNeverCache : No cache
- kCacheIfLargeSpeedup (**default**) : Cache when `batch size <= cols of packed matrix`
  - `cols of packed matrix` : Can be 2, 4, or 8. Depends on platform and kernel type
[...]

PR #4273, opened 2020-09-15
[cker/ruy] Update ruy library version

periannath:
- This commit updates ruy library version
  - The same as ruy library used by tensorflow v2.3.0
  - New dependency is added (cpuinfo)
- Change cker to use updated ruy library functions

---

For issue: #3231 #4047 
From draft: #4074


compute/cker/include/cker/NeonTensorUtils.h
```diff
@@ -74,10 +76,33 @@ inline int32_t AccumulateNeonLane(const int32x4_t lane)
 } // namespace
 
 #ifdef __aarch64__
+// The implementation of dotprod detection is copied from ruy's internal
+// function DetectDotprod().
+// At the moment it's only implemented on Linux ARM64. Consider syncing again
+// with ruy in the future to share improvements.
+#if defined __linux__ && defined __aarch64__
```
YongseopKim, line 83: The `defined __arch64__` is duplicated.

runtime/onert/core/CMakeLists.txt
```diff
@@ -13,6 +13,9 @@ target_link_libraries(onert_core PRIVATE nnfw_coverage)
 target_link_libraries(onert_core PRIVATE dl ${LIB_PTHREAD})
 target_link_libraries(onert_core PRIVATE jsoncpp)
 target_link_libraries(onert_core INTERFACE ruy_instrumentation)
+if (ANDROID)
+  target_link_libraries(onert_core INTERFACE log)
```
YongseopKim, line 17: Is it necessary?

compute/cker/include/cker/NeonTensorUtils.h
periannath: @YongseopKim Thanks! I'll fix it.

runtime/onert/core/CMakeLists.txt
periannath: @YongseopKim Android build fails without this line. I'm not sure this is the correct way to fix linking error, but it works. 

**Android Linking Dependency** 

onert_core -> cker -> ruy -> cpuinfo -> clog -> log

onert_core needs to link `liblog.so`, but it fails without above line.

periannath:
#4274 is merged into this PR.

runtime/onert/core/CMakeLists.txt
YongseopKim: I got it, thank you for your answer.

hseok-oh (dismissed):
LGTM

runtime/onert/core/CMakeLists.txt
hseok-oh: Then can't we link `log` to ruy`?

periannath: @hseok-oh Same error is generated when I linked `log` to `ruy`. 

```bash
/ndk/toolchains/llvm/prebuilt/linux-x86_64/lib/gcc/aarch64-linux-android/4.9.x/../../../../aarch64-linux-android/bin/ld: warning: liblog.so, needed by libonert_core.so, not found (try using -rpath or -rpath-link)
```

wateret: That's odd, though...

hseok-oh: Then maybe cpuinfo's header file requires `log`, and onert_core uses cpuinfo's header. Did you link `log` to `ruy` with `PRIVATE` keyword?

periannath: @hseok-oh I've tried both `INTERFACE` and `PUBLIC` keywords but the error remains.

periannath: @hseok-oh @wateret Adding `-fuse-ld=gold` to linker flags solves this issue. I found this solution from https://github.com/android/ndk/issues/1088. I'll update this PR soon.

wateret: @periannath If that has too much work and it is sure that is a bug, we may leave some comments with the current changes.

periannath: @wateret It seems that this is caused by bug in BFD linker. To solve this problem, we have to use gold linker instead of BFD linker in android build. The code change is small but I'm not sure about its side-effect. 
If it is okay, I want to review this PR as is and discuss about android linker in another issue or PR.

periannath:
@YongseopKim PTAL

runtime/onert/core/CMakeLists.txt
wateret: @periannath Sure, but could you leave some comments here?

periannath: @wateret I've added comments as you suggested.

wateret (approved):
LGTM

hseok-oh (approved):

YongseopKim (approved):
LGTM

PR #4275, opened 2020-09-15
[preset] Remove deprecated command

mhs4670go:
This commit removes deprecated command TF2NNPKG_FOR_WINDOWS.

### DIFF
--- a/infra/packaging/preset/20200731_windows
+++ b/infra/packaging/preset/20200731_windows
@@ -29,7 +29,6 @@ function preset_configure()
   # TODO Use "nncc configure" and "nncc build"
   cmake \
     -G "MSYS Makefiles" \
-    -DTF2NNPKG_FOR_WINDOWS=ON \
     -DUSE_PROTOBUF_LEGACY_IMPORT=ON \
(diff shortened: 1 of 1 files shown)


seanshpark:
Search shows `TF2NNPKG_FOR_WINDOWS` is used to use `filesystem_windows.cpp`  or `filesystem_linux.cpp`(default) in old `tf2nnpkg`. It would be better to remove using code first if this is really not used.

mhs4670go:
@seanshpark You mean old `tf2nnpkg` should be removed first? As you said, `TF2NNPKG_FOR_WINDOWS` is used in old `tf2nnpkg` but `20200731_windows` preset doesn't include old `tf2nnpkg` as a `REQUIRED_UNITS` which is why I post this PR.

seanshpark:
> You mean old tf2nnpkg should be removed first?

Nope. 

> 20200731_windows preset doesn't include old tf2nnpkg as a REQUIRED_UNITS which is why I post this PR.

OK got it :)

seanshpark (approved):
LGTM

PR #4277, opened 2020-09-15
[onert] Change shared_ptr for tensors to unique_ptr

hyunsik-yoon:
This commit enables the following:
- replace things like `shared_ptr<ITensor>` to `unique_ptr`, making `TensorRegistry` have ownership
- methods like `getITensor()` do not return `shared_ptr` but return _raw pointer_ to increase performance.

Please refer to [parent discussion](https://github.sec.samsung.net/STAR/nnfw/issues/11743) and [performance note showing that now `getITensor() is x2 faster](https://github.sec.samsung.net/STAR/nnfw/issues/11743#issuecomment-831227).

I check this commit with valgrind and there's no leak or invalid access due to this commit.

Note:
- I changed `shared_ptr` to `unique_ptr` in CPU code, However, for acl source code, I left `shared_ptr` in the internals acl code.

### DIFF
--- a/runtime/onert/core/src/exec/IPermuteFunction.h
+++ b/runtime/onert/core/src/exec/IPermuteFunction.h
@@ -56,7 +56,7 @@ public:
     auto dst_it = _dst_tensors.begin();
     while (src_it != _src_tensors.end())
     {
-      const auto src_tensor = *src_it;
+      auto src_tensor = *src_it;
       auto dst_tensor = *dst_it;
       if (src_tensor != dst_tensor)
       {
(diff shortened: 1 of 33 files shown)


runtime/onert/core/src/exec/IPermuteFunction.h
```diff
@@ -56,7 +56,7 @@ class IPermuteFunction : public IFunction
     auto dst_it = _dst_tensors.begin();
     while (src_it != _src_tensors.end())
     {
-      const auto src_tensor = *src_it;
+      auto src_tensor = *src_it;
```
hyunsik-yoon, line 59: with `const`, compiling `access(fn)` fails.

seanshpark:
Title typo `unque_ptr `

ragmani (approved):
LGTM

wateret (approved):
LGTM

PR #4279, opened 2020-09-15
[nnfwapi] Test : Remove a GenModelTestContext ctor

wateret:
Remove the `GenModelTestContext` ctor that accepts `CircleBuffer` as
lvalue reference. This is to prevent when a CircleBuffer is moved to
somewhere but the caller does not know about it.


hyunsik-yoon (approved):
LGTM

PR #4283, opened 2020-09-15
[Res/TFPythonExamples]Add UnidirectionalLSTM operation case

struss:
Parent Issue : #4151 
Draft PR : #4260 

This commit add `UnidirectionalLSTM` operation related case.

according to [this](https://colab.research.google.com/github/tensorflow/tensorflow/blob/master/tensorflow/lite/examples/experimental_new_converter/Keras_LSTM_fusion_Codelab.ipynb), `UnidirectionalSequenceLSTM` is created from `tf.keras.layers.LSTM`.

### DIFF
--- a/res/TensorFlowPythonExamples/examples/unidirectional_sequence_LSTM/__init__.py
+++ b/res/TensorFlowPythonExamples/examples/unidirectional_sequence_LSTM/__init__.py
@@ -1,0 +1,4 @@
+import tensorflow as tf
+
+in_ = tf.compat.v1.placeholder(dtype=tf.float32, shape=[28, 28, 3], name="Hole")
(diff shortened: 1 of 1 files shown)


mhs4670go (approved):
LGTM

seanshpark:
When I run this in TF1.13.2, output is a bit large, complicated model, pbtxt size 173,878B (for my case).
Is this correct?

struss:
> When I run this in TF1.13.2, output is a bit large, complicated model, pbtxt size 173,878B (for my case).
> Is this correct?

I'm not sure but following reference above, the created model is 854KB but converted TFlite model is just 17KB.

seanshpark (approved):
LGTM

PR #4287, opened 2020-09-15
[onert] Throw when odd OperationConstraint

wateret:
Throw when the number of inputs is odd for an operation.

- Introduce expected model-loading failure to GenModelTest
- Fix some operations' constraint that were wrong


hseok-oh (approved):
LGTM

PR #4292, opened 2020-09-15
[tflchef]Add UnidirectionalSequenceLSTM Operation

struss:
Parent Issue : #4151
Draft PR : #4260

This commit add `UnidirectionalSequenceLSTM` operation on TFlchef.


seanshpark (dismissed):
LGTM

seanshpark (dismissed):

mhs4670go (dismissed):
LGTM

struss:
I need to add filler for input constant. after update, I will notify to all reviewers. Sorry for bothering.

mhs4670go (dismissed):
LGTM

struss:
@mhs4670go sorry for update. inputs[0](input), inputs[17](outputstate), input[18](cellstate) are exclude from filler because this Tensor could not be constants.

mhs4670go:
@struss Actually, it would be okay without that condition as `fill_tensor_to_import` checks if it has a buffer. But better to be safe than error:) Please add a comment as you said about it because it would be hard to know what index mean and why they are excluded.

struss:
> @struss Actually, it would be okay without that condition as `fill_tensor_to_import` checks if it has a buffer. But better to be safe than error:) Please add a comment as you said about it because it would be hard to know what index mean and why they are excluded.

Okay! I will add a comment on it.

seanshpark (approved):
LGTM

compiler/tflchef/tflite/src/Op/UnidirectionalSequenceLSTM.cpp
```diff
@@ -0,0 +1,65 @@
...
+void TFliteOpUnidirectionalSequenceLSTM::filler(const tflite::Operator *op, TFliteImport *import,
+                                                tflchef::ModelRecipe *model_recipe) const
+{
+  const std::vector<int32_t> &inputs = as_index_vector(op->inputs());
+  assert(inputs.size() == 24);
+
+  for (int32_t i = 0; i < inputs.size(); i++)
+  {
+    // Except for Input 0, 17 and 18.
+    // Each Input mean  Input Tensor, OutputState Tensor and CellState Tensor.
```
seanshpark, line 34: `mean  Input` : seems something is missing...

struss: Oh I will fix it.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #4293, opened 2020-09-15
[onert] Introduce PassRunner

wateret:
Introduce a utility class PassRunner.

- `PassRunner` is a composite class of `Pass`es
- This class is for grouping some passes and dumping the graph after
  each pass


hyunsik-yoon (approved):
LGTM

PR #4301, opened 2020-09-16
[ndarray] Do not link coverage lib

wateret:
Do not link coverage lib since it is not used by onert.


seanshpark (approved):
LGTM

hseok-oh:
Can't we remove this lib? We can port TFLite's kernel if we need to support `NON_MAX_SUPPRESSION_V4` or `NON_MAX_SUPPRESSION_V5`.

wateret:
@hseok-oh It is never used, so I would like to remove it if possible.

wateret:
Will close if #4304 lands

PR #4304, opened 2020-09-16
[ndarray] Get rid of ndarray

wateret:
Get rid of ndarray which is never used.


hseok-oh (approved):
LGTM

Issue #41, opened 2020-04-17
[oneapi] Make enough test cases


wateret:
Let's make some test cases for *oneapi*.

We may refer to Android NN's test cases.

# NOTE

- The number of **negative test cases** must be equal or greater than the **positive ones**.
    - **I won't accept the commits that only contain positive tests**

# End of May Goals

- Validation Tests (Total 75+) : Focus on validation with simple models
    - 30+ positive TCs
    - 45+ negative TCs
- Model Tests (Total 8+) : Focus on different kind of models with specific scenarios
    - 3+ positive TCs
    - 5+ negative TCs

wateret:
Done!

PR #4307, opened 2020-09-16
[onert] Fix incorrect result of Cast op with bool type input in acl_cl backend

ragmani:
For issue #4309

This commit fixes results mismatch errors when using cast ops with bool type input.
  - Add a acl_cl kernel for cast operation with bool type input
  - Applied the kernel
  - Add some nnfw_api testcases

### DIFF
--- a/compute/ARMComputeEx/arm_compute/core/CL/kernels/CLCastBoolKernel.h
+++ b/compute/ARMComputeEx/arm_compute/core/CL/kernels/CLCastBoolKernel.h
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


wateret (changes requested):
LGTM, but please add negative test cases(2 or more). It is due to the release criteria. 😞

ragmani:
> LGTM, but please add negative test cases(2 or more). It is due to the release criteria. disappointed

@wateret 
How to create negative test cases unexpected result?

ragmani:

wateret:
Negative tests are not necessarily relevant to this PR. This is just for release criteria, to continuously keep the ratio of positive and negative TCs. You can also add negative tests with a separate PR.

It is announced in #41 .

> - The number of **negative test cases** must be equal or greater than the **positive ones**.
>     - **I won't accept the commits that only contain positive tests**
>

wateret:
For example, these days I started making some invalid models - https://github.com/Samsung/ONE/blob/master/tests/nnfw_api/src/one_op_tests/LogSoftmax.cc

ragmani:
> Negative tests are not necessarily relevant to this PR. This is just for release criteria, to continuously keep the ratio of positive and negative TCs. You can add negative tests with a separate PR if irrelevant, but please make them at the same time.

@wateret 
To add some irrelevant negative tests to this PR such as invalid model tests, I have to add some commits that fix error handling for the invalid model. It would make this PR big and multirole for fixing two or more error. Is it right?

wateret (approved):
LGTM, agreed on adding negative tests after landing.

PR #4312, opened 2020-09-16
[onert] Optimize DynamicTensorManager::deallocInput()

hyunsik-yoon:
This optimizes `DynamicTensorManager::deallocInput()`.
(Also refer to [here](https://github.sec.samsung.net/STAR/nnfw/issues/11758))

**Background**

When handling a model with dynamic tensors, `deallocInput()` is always called after running a kernel of op.
If a kernel is small, overhead of [ `deallocInput()` * number of ops ] could be relatively big.
For this reason, optimization of `deallocInput()` is necessary.

**How to**

1. Eliminate usage of tensor index
2. This also removes the calling number of `getITensor()`, resulting in more optimization.

**Result**

(assuming b05eeec7 result is 100, tested on Odroid XU4)

   case         |  b05eeec7  | with #4277 | more with this commit
----------|----------|----------|---------
total `getITensors()`s   |     100    |     45.4   |   32.3
total `deallocInputs()`s |     100    |     68.9   |   58.8

### DIFF
--- a/runtime/onert/core/include/backend/IDynamicTensorManager.h
+++ b/runtime/onert/core/include/backend/IDynamicTensorManager.h
@@ -41,10 +41,10 @@ public:
   /**
    * @brief Plan when to delete a tensor. Note this planning is done at compilation time.
    * @param op_ind        operation index
-   * @param operand_ind   operand index of input operand of first param op. Operand can be static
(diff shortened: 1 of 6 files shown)


ragmani (approved):
LGTM

wateret (approved):
LGTM

PR #4315, opened 2020-09-16
[tests] Scheduler profile for Android

seanshpark:
This will introduce test_scheduler_with_profiling and related scripts for Android


wateret (approved):
LGTM

PR #4319, opened 2020-09-17
[onert] Add Tensor's reference count interfaces

dayo09:
- Add functions needed to unused-tensor's early release nnfw#11749

### DIFF
--- a/runtime/onert/backend/cpu/Tensor.h
+++ b/runtime/onert/backend/cpu/Tensor.h
@@ -95,6 +95,21 @@ public:
     }
   }
 
+  /**
+   * @brief Reset reference count to zero and release data
+   */
+  void reset_ref() override
+  {
+    assert(_data != nullptr);
+    assert(_num_references > 0);
+    _num_references = 0;
+
+    _data.reset();
+    _buffer = nullptr;
+  }
+
+  int32_t num_references() override { return _num_references; }
+
 private:
   std::shared_ptr<const ir::Data> _data;
 };
--- a/runtime/onert/core/include/backend/cpu_common/Tensor.h
+++ b/runtime/onert/core/include/backend/cpu_common/Tensor.h
@@ -119,6 +119,7 @@ public:
 
     ++_num_references;
   }
+
   virtual void decrease_ref()
   {
     assert(_buffer != nullptr || _allocator != nullptr);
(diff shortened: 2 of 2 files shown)


periannath (commented):
How about adding `assert(_num_references > 0);` at `buffer()` function? I think it is better to prevent access of released buffer.

periannath (approved):
LGTM

dayo09:
### Issue : using buffer() when _num_reference == 0

- ConvolutionLayer.cc
- At first prepare(), is_transposed == 0, so transform _kernel(original weight) into _conv_kernel->_modified_filter_data and decrease reference
- At run(), convFloat32(), calls kernel with _kernel->buffer(). In computation, the value is not used and _modified_filter_data is used.

So there is no bug. So the assertion should be removed?

```c++
  void operator()(const ConvParams &params, const Shape &input_shape, const float *input_data,
                  const Shape &filter_shape, const float *filter_data, const Shape &bias_shape,
                  const float *bias_data, const Shape &output_shape, float *output_data)
  {
    if (usableMultiThreaded(params.padding_type, params.dilation_width_factor,
                            params.dilation_height_factor))
    {
      bool transposed_in_execution = false;
      if (!_prepared)
      {
        // This means that filter is not constant
        // TODO Apply optimized kernel if multithreaded kernel is slower than optimized kernel by
        // transposing filter data
        transposeFilter(filter_shape, filter_data, transposed_in_execution);
      }
      multithreaded::Conv(params, input_shape, input_data, filter_shape, &_modified_filter_data[0],
                          bias_shape, bias_data, output_shape, output_data);
    }
    else
    {
      // TODO Support optimized kernel
      reference::Conv(params, input_shape, input_data, filter_shape, filter_data, bias_shape,
                      bias_data, output_shape, output_data);
    }
  }
```

periannath:
> So the assertion should be removed?

I think assertion should be removed to bypass convolution case. Maybe we can add assertion to `buffer()` function later considering convolution case in another PR.

dayo09:
@periannath Yep I removed the assert for now. I will be able to add the assert if the #4324 PR is accepted and merged.
@wateret PTAL

periannath (dismissed):
LGTM

hseok-oh:
Please fix commit message. `nnfw#11749` is invalid link.

wateret (dismissed):
LGTM

wateret:
@YongseopKim PTAL

dayo09:
@hseok-oh I updated the commit msg.
@periannath @YongseopKim @wateret PTAL

YongseopKim (approved):
LGTM.

Can I adress my idea? (Optional. Not necessary to follow my idea.)

decrease_ref() and reset_ref() have a duplicated part. It could be extracted to like `reset()`. And the `reset()` could be override.

Therefore,

Tensor has decrease_ref() and reset_ref() as non virtual but does reset() as virtual. And ExternalTensor overrides the only reset().

wateret (approved):
LGTM

periannath (approved):
LGTM

PR #4328, opened 2020-09-17
[onert] Fix wrong input size constraint of Pack op

hyunsik-yoon:
This fixes wrong input size constraint of Pack.

Situation:

- In a model I am working on, there is a `Pack` op like the following and it has only one input. :open_mouth: 

```
PACK
    ValuesCount(1) Axis(0)
    I  strided_slice_7
    O  Reshape_2/shape
```

- I checked with Tensorflow doc, and there is no constraints on the size of inputs.
- I checked with code of TFLite kernel, and only constraint I've seen was
  number of input tensors >= 1.


wateret (approved):
LGTM

PR #4333, opened 2020-09-17
[nnfwapi] Test : Add "Add" operation tests

wateret:
Add "Add" operation tests - 1 positive / 3 negative.

ONE-DCO-1.0-Signed-off-by: Hanjoung Lee <hanjoung.lee@samsung.com>

PR #4334, opened 2020-09-17
[onert] Fix when a model has const tensor output

### LINKED ISSUE #4231: [onert] Crash when a model output tensor is const
Minimal-reproducible example(to be added as a regression test):

```cpp
TEST_F(GenModelTest, DISABLED_UnusedConstOutputOnly)
{
  // A single tensor which is constant
  CircleGen cgen;
  uint32_t const_buf = cgen.addBuffer(std::vector<float>{9, 8, 7, 6});
  int out_const = cgen.addTensor({{1, 2, 2, 1}, circle::TensorType::TensorType_FLOAT32, const_buf});
  cgen.setInputsAndOutputs({}, {out_const});

  _context = std::make_unique<GenModelTestContext>(cgen.finish());
  _context->addTestCase(uniformTCD<float>({}, {{9, 8, 7, 6}}));
  _context->addTestCase(uniformTCD<float>({}, {{9, 8, 7, 6}}));
  _context->setBackends({"acl_cl", "acl_neon", "cpu"});

  SUCCEED();
}

TEST_F(GenModelTest, DISABLED_UnusedConstOutputAndAdd)
{
  // A single tensor which is constant + an Add op
  CircleGen cgen;
  uint32_t rhs_buf = cgen.addBuffer(std::vector<float>{5, 4, 7, 4});
  uint32_t const_buf = cgen.addBuffer(std::vector<float>{9, 8, 7, 6});
  int lhs = cgen.addTensor({{1, 2, 2, 1}, circle::TensorType::TensorType_FLOAT32});
  int rhs = cgen.addTensor({{1, 2, 2, 1}, circle::TensorType::TensorType_FLOAT32, rhs_buf});
  int out = cgen.addTensor({{1, 2, 2, 1}, circle::TensorType::TensorType_FLOAT32});
  int out_const = cgen.addTensor({{1, 2, 2, 1}, circle::TensorType::TensorType_FLOAT32, const_buf});
  cgen.addOperatorAdd({{lhs, rhs}, {out}}, circle::ActivationFunctionType_NONE);
  cgen.setInputsAndOutputs({lhs}, {out, out_const});

  _context = std::make_unique<GenModelTestContext>(cgen.finish());
  _context->addTestCase(uniformTCD<float>({{1, 3, 2, 4}}, {{6, 7, 9, 8}, {9, 8, 7, 6}}));
  _context->addTestCase(uniformTCD<float>({{0, 1, 2, 3}}, {{5, 5, 9, 7}, {9, 8, 7, 6}}));
  _context->setBackends({"acl_cl", "acl_neon", "cpu"});

  SUCCEED();
}
```

wateret:
Fix when a model has const tensor output

- Introduce ConstantOutputPass to resolve the problem when there is some
  output tensors that are also constants.

Resolve #4231

### DIFF
--- a/runtime/onert/core/include/ir/OperandInfo.h
+++ b/runtime/onert/core/include/ir/OperandInfo.h
@@ -117,6 +117,7 @@ public:
 
   MemAllocType memAllocType() const { return _alloc_type; }
   void setAsConstant() { _const = true; }
+  void setAsNonConst() { _const = false; }
   bool isConstant() const
   {
     // Impossible case: constant and dynamic operand
--- a/runtime/onert/core/src/compiler/pass/ConstantOutputPass.cc
+++ b/runtime/onert/core/src/compiler/pass/ConstantOutputPass.cc
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ConstantOutputPass.h"
+
+#include "ir/Graph.h"
+#include "ir/operation/Permute.h"
+#include "util/logging.h"
+
+namespace onert
+{
+namespace compiler
+{
+namespace pass
+{
+
+void ConstantOutputPass::callback(const ir::OperandIndex &ind, ir::Operand &obj)
+{
+  if (!_graph.getOutputs().contains(ind) || !obj.isConstant())
+    return;
+
+  auto permute_input_ind = _graph.addOperand(obj.shape(), obj.typeInfo());
+  auto &permute_input_obj = _graph.operands().at(permute_input_ind);
+
+  // Move the const data
+  permute_input_obj.data(obj.shareData());
+  obj.releaseData();
+  obj.info().setAsNonConst();
+
+  using ir::operation::Permute;
+  auto permute_obj = std::make_unique<Permute>(permute_input_ind, ind, Permute::Type::COPY);
+  auto permute_ind = _graph.operations().push(std::move(permute_obj));
+
+  permute_input_obj.insertUse(permute_ind);
+  obj.setDef(permute_ind);
+
+  VERBOSE(ConstantOutputPass) << "Permute Op inserted for a constant ouput, node index : "
+                              << permute_ind << std::endl;
+  VERBOSE(ConstantOutputPass) << "  - Input (inserted) Operand : " << permute_input_ind
+                              << std::endl;
+  VERBOSE(ConstantOutputPass) << "  - Output(original) Operand : " << ind << std::endl;
+}
+
+} // namespace pass
+} // namespace compiler
+} // namespace onert
--- a/tests/nnfw_api/src/GenModelTests.cc
+++ b/tests/nnfw_api/src/GenModelTests.cc
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @file This file contains miscellaneous GenModelTest test cases.
+ *
+ */
+
+#include "GenModelTest.h"
+
+#include <memory>
+
+TEST_F(GenModelTest, UnusedConstOutputOnly)
+{
+  // A single tensor which is constant
+  CircleGen cgen;
+  uint32_t const_buf = cgen.addBuffer(std::vector<float>{9, 8, 7, 6});
+  int out_const = cgen.addTensor({{1, 2, 2, 1}, circle::TensorType::TensorType_FLOAT32, const_buf});
+  cgen.setInputsAndOutputs({}, {out_const});
+
+  _context = std::make_unique<GenModelTestContext>(cgen.finish());
+  _context->addTestCase(uniformTCD<float>({}, {{9, 8, 7, 6}}));
+  _context->setBackends({"acl_cl", "acl_neon", "cpu"});
+
+  SUCCEED();
+}
+
+TEST_F(GenModelTest, UnusedConstOutputAndAdd)
+{
+  // A single tensor which is constant + an Add op
+  CircleGen cgen;
+  uint32_t rhs_buf = cgen.addBuffer(std::vector<float>{5, 4, 7, 4});
+  uint32_t const_buf = cgen.addBuffer(std::vector<float>{9, 8, 7, 6});
+  int lhs = cgen.addTensor({{1, 2, 2, 1}, circle::TensorType::TensorType_FLOAT32});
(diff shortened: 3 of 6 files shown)


wateret:
This commit has 2 positive tests and #4333 has 2 more negative tests.

tests/nnfw_api/src/GenModelTests.cc
```diff
@@ -0,0 +1,60 @@
...
+{
+  // A single tensor which is constant
+  CircleGen cgen;
+  uint32_t const_buf = cgen.addBuffer(std::vector<float>{9, 8, 7, 6});
+  int out_const = cgen.addTensor({{1, 2, 2, 1}, circle::TensorType::TensorType_FLOAT32, const_buf});
+  cgen.setInputsAndOutputs({}, {out_const});
+
+  _context = std::make_unique<GenModelTestContext>(cgen.finish());
+  _context->addTestCase(uniformTCD<float>({}, {{9, 8, 7, 6}}));
+  _context->setBackends({"cpu"});
```
hyunsik-yoon, line 36: I wonder if this does not work with acl or neon.
I am not sure but if it is because there is no op that can be run with acl backend,
what if we inserts some op like `concat(input)` instead of `permute`?

runtime/onert/core/include/ir/OperandInfo.h
```diff
@@ -117,6 +117,7 @@ class OperandInfo
 
   MemAllocType memAllocType() const { return _alloc_type; }
   void setAsConstant() { _const = true; }
+  void setAsNonConst() { _const = false; }
```
ragmani, line 120: If a constant operand is used as both output of graph and input of a node, the operand is not constant?

tests/nnfw_api/src/GenModelTests.cc
wateret: Oops, I removed those while working on it. I will add acl_cl and neon backends. As Permute is run by `controlflow` so it does not have an issue.

runtime/onert/core/include/ir/OperandInfo.h
wateret: Good point, I think I can work on it after getting this merged.

I'm thinking changing the node's input to the newly created constant tensor in `ConstantOutputPass`.

```
( )   : non-input/output tensor
(( )) : output tensor
[ ]   : operation

Original Model
(( Const Output )) -> [ Some Operation ] -> ( Some Tensor ) -> ...

After performing ConstantOutputPass (AS-IS)
( Const ) -> (( Output )) -> [ Some Operation ] -> ( Some Tensor ) -> ...

After performing ConstantOutputPass (TO-BE)
( Const ) -> (( Output )) 
         \-> [ Some Operation ] -> ( Some Tensor ) -> ...
```

hyunsik-yoon (approved):
LGTM

runtime/onert/core/include/ir/OperandInfo.h
ragmani: > Good point, I think I can work on it after getting this merged.

What if the constant operand is used in multiple backends? If so, I think `setAsNonConst()` must be called after being copied to each backend by calling `ConstantInsertionPass`. But it seems to be called before `ConstantInsertionPass` in this PR.

wateret: Wouldn't my solution resolve that case?

```
After performing ConstantOutputPass (TO-BE)
( Const ) -> (( Output )) 
         \-> [ Some Operation ] -> ( Some Tensor ) -> ...
```

Then other operations use a constant.

ragmani: I'm not sure if this pr wouldn't cover that case. It wouldn't probably cover that case.

Can your solution cover this case?
```
( Const ) -> [ Permute ] -> (( Output )) 
         \-> [ Some Operation on cpu ] -> ( Some Tensor ) -> ...
         \-> [ Some Operation on acl_cl ] -> ( Some Tensor ) -> ...
```
I think your solution should cover this case too.

wateret: Yes, as this pass is run before lowering, so those cases are handled in the later pass `PermutationInsertionPass`.

ragmani:
```
== After PermutationInsertionPass ==
( Const ) -> [ Permute ] -> (( Output ))
         \-> [ Some Operation on cpu ] -> ( Some Tensor ) -> ...
         \-> [ Permute ] -> ( Permuted Const ) -> [ Some Operation on acl_cl ] -> ( Some Tensor ) -> ...
```

Does `\-> [ Permute ] -> ( Permuted Const ) -> [ Some Operation on acl_cl ]` mean that a permute node is inserted for constant operand? Is it going to run every executions? Why don't you resolve it by `ConstantInitializer`?

ragmani:
```
After performing ConstantOutputPass (AS-IS)
( Const ) -> (( Output )) -> [ Some Operation ] -> ( Some Tensor ) -> ...
```

"AS-IS" means this PR? I don't find the test to verify it in this PR. So I'm not sure if this PR resolves it. Could you add a test for this case?

wateret: > Why don't you resolve it by ConstantInitializer?

Oh I probably knew something wrong. 😢  Anyway the thing is, as far as I know, our runtime can handle this graph. What `ConstantOutputPass` does is:

1. Create a permute and const input (This PR, as-is)
2. Change operation inputs to the newly-created-const (to-be-done)

```
( Const ) -> [ Permute ] -> (( Output ))
         \-> [ Some Operation on cpu ] -> ( Some Tensor ) -> ...
         \-> [ Some Operation on acl_cl ] -> ( Some Tensor ) -> ...
```

And after that, what I described about the runtime was probably wrong. It seems like `ConstantInsertionPass` generates other constants(rather than creating Permute) and then `ConstantIntializer` fills all of them, and I guess it is what you meant. It it may look like this:

```
== After PermutationInsertionPass ==
( Const ) -> [ Permute ] -> (( Output ))
         \-> [ Some Operation on cpu ] -> ( Some Tensor ) -> ...
( Const ) -> [ Some Operation on acl_cl ] -> ( Some Tensor ) -> ...
```

> "AS-IS" means this PR? 

Yes.

>  I don't find the test to verify it in this PR. So I'm not sure if this PR resolves it. Could you add a test for this case?

Sure, but can I do that later? I also need to prepare some negative tests. And this is originally for #4231 .

runtime/onert/core/src/compiler/pass/ConstantOutputPass.cc
```diff
@@ -0,0 +1,59 @@
...
+  permute_input_obj.data(obj.shareData());
+  obj.releaseData();
+  obj.info().setAsNonConst();
+
+  using ir::operation::Permute;
+  auto permute_obj = std::make_unique<Permute>(permute_input_ind, ind, Permute::Type::COPY);
+  auto permute_ind = _graph.operations().push(std::move(permute_obj));
+
+  permute_input_obj.insertUse(permute_ind);
+  obj.setDef(permute_ind);
```
ragmani, line 48: From https://github.com/Samsung/ONE/pull/4334/files#r492524848

It seems to make the graph like below
```
== Original model ==
(( Const Output )) 
         \-> [ Some Operation on cpu ] -> ( Some Tensor ) -> ...
         \-> [ Some Operation on acl_cl ] -> ( Some Tensor ) -> ...

== After ConstantOutputPass ==
( Const ) -> [ Permute ] -> (( Output (Non-const) ))  -> [ Some Operation on cpu ] -> ( Some Tensor ) -> ...
                                                      \-> [ Some Operation on acl_cl ] -> ( Some Tensor ) -> ...

== After ConstantInsertionPass ==
Nothing. "Output" is not const.

== After PermutationInsertionPass ==
( Const ) -> [ Permute ] -> (( Output (Non-const) ))  -> [ Some Operation on cpu ] -> ( Some Tensor ) -> ...
                                                      \->  [ Permute ] -> ( Permuted Tensor )  -> [ Some Operation on acl_cl ] -> ( Some Tensor ) -> ...
```

Is it right? If so, please do not change operation's const operand to non-const. There are some problems as follows:
1. Those operations using "Output" cannot get constant data from the "Output" operand in KernelGenerator. In model, the operand is constant and has data, but in graph after running passes, the operand doen't have data.
2. Some kernels may cache constant data with transposing it for performance. But if they cannot cache non-constant data.
3. Performance will be decrease by that increased "Permute" node .

ragmani: This problems would be fixed by @wateret in other PR.

ragmani (approved):
LGTM

runtime/onert/core/src/compiler/pass/ConstantOutputPass.cc
wateret: @ragmani Thank you for detailed review 😄

Issue #4244, opened 2020-09-14
[onert] An error by using SubTensor as input of Pack operation on acl backend


ragmani:
![Subtensor error](https://user-images.githubusercontent.com/18050611/93043957-54664980-f68e-11ea-9921-cc066d796020.PNG)

```bash
$ USE_NNAPI=1 BACKENDS="acl_cl" ./Product/armv7l-linux.debug/out/bin/nnapi_test temp.tflite 
[NNAPI TEST] Run T/F Lite Interpreter without NNAPI
[NNAPI TEST #1] Run T/F Lite Interpreter with NNAPI
...
[ALLOC] base pointer: 0x30b870
[acl_cl_KernelGenerator_Concat] Concat eliminated
[EXCEPTION] in extend_parent_shape src/core/SubTensorInfo.cpp:45: (coords.x() != 0) || (coords.y() != 0)
[NNAPI::Compilation] finish: Fail to compile
Aborting since NNAPI returned failure nnapi_delegate.cc:1098
[onert_backend_create] 'acl_cl' unloaded
```

ragmani:
Done

ragmani:
#4266 will be fix the error caused by #4266

PR #4335, opened 2020-09-17
[onert] Fixes dim_correction incorrectly applied to Unpack

ragmani:
For issue https://github.com/Samsung/ONE/issues/4244

This commit fixes dim_correction incorrectly applied to Unpack of acl backends

### DIFF
--- a/runtime/onert/backend/acl_cl/KernelGenerator.cc
+++ b/runtime/onert/backend/acl_cl/KernelGenerator.cc
@@ -1470,7 +1470,7 @@ void KernelGenerator::visit(const ir::operation::Unpack &node)
   for (const auto &output_index : node.getOutputs())
     output_indexes.emplace_back(output_index);
 
-  auto input = _tensor_reg->getAclTensor(input_index)->handle();
+  auto input_tensor = _tensor_reg->getAclTensor(input_index);
   std::vector<arm_compute::ICLTensor *> outputs;
   for (const auto &output_index : output_indexes)
     outputs.emplace_back(_tensor_reg->getAclTensor(output_index)->handle());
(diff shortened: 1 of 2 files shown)


runtime/onert/backend/acl_cl/KernelGenerator.cc
```diff
@@ -1482,26 +1482,19 @@ void KernelGenerator::visit(const ir::operation::Unpack &node)
   axis = acl_common::ToARMComputeAxis(input_rank, axis, frontend_layout, backend_layout).value();
 
   // Disable applied dim_correction
-  for (const auto &output_index : output_indexes)
+  if (input_tensor->num_dimensions() != input_tensor->info()->num_dimensions())
```
hyunsik-yoon, line 1485: I have a question. I just assume that one is number of dim from one IR and another is number of dim from acl. Is my assumption correct?

It is hard to understand that both returns number of dimensions but their values are different... (Not the scope of this PR but we may need some enhancement later, I guess...)

ragmani: input_tensor->num_dimensions() -> Wrapper tensor `backend::ITensor`
input_tensor->info()->num_dimensions() -> Tensor of backend `arm_compute::ITensor`

> I have a question. I just assume that one is number of dim from one IR and another is number of dim from acl. Is my assumption correct?

Yes, It is correct.

> It is hard to understand that both returns number of dimensions but their values are different... (Not the scope of this PR but we may need some enhancement later, I guess...)

Okay, I will change it to getting rank from ir.

hyunsik-yoon: Thanks!

hyunsik-yoon (approved):
LGTM

wateret (approved):
LGTM

Issue #1612, opened 2020-05-28
[luci-interpreter] Operators to enable (2)


jinevening:
Next step of #1157.

You can learn how to add operators to luci-interpreter from [previous PRs](https://github.com/Samsung/ONE/pulls?q=luci+interpreter+kernel+in%3Atitle+is%3Apr+is%3Aclosed).
(ex: Pad #1509, AveragePool2D #1218, Reshape #1099, FullyConnected #1087)

If you want to participate in operator implementation, pick an operator, put your name into the PR column and replace it with the link to the PR once you are done.

**Top priority**
Operators | Status (float) | Status (uint8) | PR
-- | :--: | :--: | --
logistic | done | done | #1810 
transpose | done | done | #1858
transpose_conv | done | done | #2033

**Long-term goal** (in order from easy to difficult)

Activations | Status (float) | Status (uint8) | PR
-- | :--: | :--: | --
elu | done | &mdash; |#2147
leaky_relu | done | done |#2160
prelu | done | done | #4039
relu | done | done |  #4150 
relu6 | done | done | #4192
tanh | done | done | #3622

Element-wise | Status (float) | Status (uint8) | PR
-- | :--: | :--: | --
logical_or | only bool | only bool | #4716 
div | done | done | #4425 
equal | done | done | #4336
floor | done | &mdash; | #4402
floor_div(floordiv) | done | &mdash; | #4421
greater | done | done | #4336
greater_equal | done | done | #4336
less | done | done | #4336
less_equal | done | done | #4336
not_equal | done | done | #4336
pow | done | &mdash; | #4361
rsqrt | done | &mdash; | #3992
sqrt | done | &mdash; | #3904 
sub(subtract) | done | done | #4464

Space transformation | Status (float) | Status (uint8) | PR
-- | :--: | :--: | --
depth_to_space | done | done | #3466 
resize_bilinear | done | done | #4163 
resize_nearestneighbor | done | done | #4106 
reverse | done | done | #3435 
slice | done | done | #3385 
space_to_depth | done | done | #3230 
split | done | done | #1742
squeeze | done | done | #3212 
strided_slice | done  | done  | #3049 

Other | Status (float) | Status (uint8) | PR
-- | :--: | :--: | --
l2_normalize | done | done |#2394
l2_pool | done | &mdash; |#2321
local_response_normalize | done | &mdash; |#2350

s-barannikov:
@jinevening Why exactly these operators? Are they from some target models?
[...]

PR #4336, opened 2020-09-17
[luci-interpreter] Support for comparison kernels

ai-moiseev:
Add support for comparison ops in luci-interpreter:
* Greater
* GreaterEqual
* Less
* LessEqual
* Equal
* NotEqual

For #1612 

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Equal.cpp
+++ b/compiler/luci-interpreter/src/kernels/Equal.cpp
@@ -1,0 +1,113 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "kernels/Equal.h"
+#include "kernels/Utils.h"
+
+#include <tensorflow/lite/kernels/internal/reference/comparisons.h>
+
+#include <stdexcept>
+
+namespace luci_interpreter
+{
+
+namespace kernels
+{
+
+Equal::Equal(const Tensor *x, const Tensor *y, Tensor *output) : Kernel({x, y}, {output}) {}
+
+void Equal::configure()
+{
+  LUCI_INTERPRETER_CHECK(x()->element_type() == y()->element_type());
+  LUCI_INTERPRETER_CHECK(output()->element_type() == DataType::BOOL);
+
+  if (x()->element_type() == DataType::U8)
+  {
+    quantizeMultiplier(output()->scale() / x()->scale(), &_x_multiplier, &_x_shift);
+    quantizeMultiplier(output()->scale() / y()->scale(), &_y_multiplier, &_y_shift);
+  }
+  output()->resize(calculateShapeForBroadcast(x()->shape(), y()->shape()));
+}
+
+void Equal::execute() const
+{
+  switch (x()->element_type())
+  {
+    case DataType::FLOAT32:
+      evalFloat();
+      break;
+    case DataType::U8:
+      evalQuantized();
+      break;
+    default:
+      throw std::runtime_error("Unsupported type.");
+  }
+}
+
+void Equal::evalFloat() const
+{
+  const auto x_data = getTensorData<float>(x());
+  const auto y_data = getTensorData<float>(y());
+  auto output_data = getTensorData<bool>(output());
+
+  tflite::ComparisonParams op_params;
+  op_params.is_broadcast = x()->shape() != y()->shape();
+
+  if (op_params.is_broadcast)
+  {
+    tflite::reference_ops::Broadcast4DSlowEqual(op_params, getTensorShape(x()), x_data,
+                                                getTensorShape(y()), y_data,
+                                                getTensorShape(output()), output_data);
+  }
+  else
+  {
+    tflite::reference_ops::Equal(op_params, getTensorShape(x()), x_data, getTensorShape(y()),
+                                 y_data, getTensorShape(output()), output_data);
+  }
+}
+
+void Equal::evalQuantized() const
+{
+  const auto x_data = getTensorData<uint8_t>(x());
+  const auto y_data = getTensorData<uint8_t>(y());
+  auto output_data = getTensorData<bool>(output());
+
+  tflite::ComparisonParams op_params;
+  op_params.left_shift = 8;
+  op_params.input1_offset = -x()->zero_point(); // Note the '-'
+  op_params.input1_shift = -_x_shift;           // Note the '-'
+  op_params.input1_multiplier = _x_multiplier;
+  op_params.input2_offset = -y()->zero_point(); // Note the '-'
+  op_params.input2_shift = -_y_shift;           // Note the '-'
+  op_params.input2_multiplier = _y_multiplier;
+  op_params.is_broadcast = x()->shape() != y()->shape();
+
+  if (op_params.is_broadcast)
+  {
+    tflite::reference_ops::Broadcast4DSlowEqualWithScaling(op_params, getTensorShape(x()), x_data,
+                                                           getTensorShape(y()), y_data,
+                                                           getTensorShape(output()), output_data);
+  }
+  else
+  {
+    tflite::reference_ops::EqualWithScaling(op_params, getTensorShape(x()), x_data,
+                                            getTensorShape(y()), y_data, getTensorShape(output()),
+                                            output_data);
+  }
+}
+
+} // namespace kernels
+} // namespace luci_interpreter
--- a/compiler/luci-interpreter/src/kernels/Equal.cpp
+++ b/compiler/luci-interpreter/src/kernels/Equal.cpp
@@ -36,8 +36,8 @@ void Equal::configure()
 
   if (x()->element_type() == DataType::U8)
   {
-    quantizeMultiplier(output()->scale() / x()->scale(), &_x_multiplier, &_x_shift);
-    quantizeMultiplier(output()->scale() / y()->scale(), &_y_multiplier, &_y_shift);
+    quantizeMultiplierSmallerThanOneExp(x()->scale(), &_x_multiplier, &_x_shift);
+    quantizeMultiplierSmallerThanOneExp(y()->scale(), &_y_multiplier, &_y_shift);
   }
   output()->resize(calculateShapeForBroadcast(x()->shape(), y()->shape()));
 }
@@ -88,10 +88,10 @@ void Equal::evalQuantized() const
   tflite::ComparisonParams op_params;
   op_params.left_shift = 8;
   op_params.input1_offset = -x()->zero_point(); // Note the '-'
-  op_params.input1_shift = -_x_shift;           // Note the '-'
+  op_params.input1_shift = _x_shift;
   op_params.input1_multiplier = _x_multiplier;
   op_params.input2_offset = -y()->zero_point(); // Note the '-'
-  op_params.input2_shift = -_y_shift;           // Note the '-'
+  op_params.input2_shift = _y_shift;
   op_params.input2_multiplier = _y_multiplier;
   op_params.is_broadcast = x()->shape() != y()->shape();
 
--- a/compiler/luci-interpreter/src/kernels/Equal.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Equal.test.cpp
@@ -1,0 +1,184 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "kernels/Equal.h"
+#include "kernels/TestUtils.h"
+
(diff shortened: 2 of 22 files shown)


compiler/luci-interpreter/src/kernels/Greater.cpp
```diff
@@ -0,0 +1,114 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "kernels/Greater.h"
+#include "kernels/Utils.h"
+
+#include <tensorflow/lite/kernels/internal/types.h>
```
s-barannikov, line 20: Is it used?

compiler/luci-interpreter/src/kernels/Greater.test.cpp
```diff
@@ -0,0 +1,226 @@
...
+{
+  Tensor x_tensor = makeInputTensor<DataType::FLOAT32>({1}, {1.f});
+  Tensor y_tensor = makeInputTensor<DataType::U8>({1}, {1});
+  Tensor output_tensor = makeOutputTensor(DataType::BOOL);
+
+  Greater kernel(&x_tensor, &y_tensor, &output_tensor);
+  EXPECT_ANY_THROW(kernel.configure());
+}
+
+TEST(Relu6Test, Input_Output_Type_NEG)
```
s-barannikov, line 203:
```suggestion
TEST(GreaterTest, InputOutputType_NEG)
```

compiler/luci-interpreter/src/kernels/Greater.test.cpp
```diff
@@ -0,0 +1,226 @@
...
+  Tensor output_tensor = makeOutputTensor(DataType::FLOAT32);
+
+  Greater kernel(&x_tensor, &y_tensor, &output_tensor);
+  EXPECT_ANY_THROW(kernel.configure());
+}
+
+TEST(Relu6Test, Invalid_Input_Type_NEG)
+{
+  Tensor x_tensor = makeInputTensor<DataType::S64>({1}, {1});
+  Tensor y_tensor = makeInputTensor<DataType::S64>({1}, {1});
```
s-barannikov, line 216: It is not invalid, it is unsupported (yet). May be you meant unmatching input types?

compiler/luci-interpreter/src/kernels/Greater.cpp
ai-moiseev: Nope, it gets included with `reference/comparisons.h`, forgot to remove it during prototyping.

compiler/luci-interpreter/src/kernels/Greater.test.cpp
```diff
@@ -0,0 +1,226 @@
...
+  Greater kernel(&x_tensor, &y_tensor, &output_tensor);
+  kernel.configure();
+  kernel.execute();
+
+  EXPECT_THAT(extractTensorData<bool>(output_tensor), ::testing::ElementsAreArray(ref_output_data));
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray({3, 3}));
+}
+
+// Choose min / max in such a way that there are exactly 256 units to avoid rounding errors.
+const float F_MIN = -128.0 / 128.0;
```
a-zh, line 90: -1.0?

compiler/luci-interpreter/src/kernels/Greater.test.cpp
ai-moiseev: Type mismatch gets tested in testcase above. Since S64 is on roadmap, I'll remove this TC.

compiler/luci-interpreter/src/kernels/Greater.test.cpp
ai-moiseev: Yes, I've left this to clarify quantized value ranges, like [here](https://github.com/Samsung/ONE/blob/6c669d42e47b34e5376a629c6dd5ece00a1b4565/compiler/luci-interpreter/src/kernels/Relu.test.cpp#L87).

compiler/luci-interpreter/src/kernels/Greater.test.cpp
ai-moiseev: Done, fixed it.

compiler/luci-interpreter/src/kernels/Greater.h
```diff
@@ -0,0 +1,53 @@
...
+
+  void configure() override;
+  void execute() const override;
+
+private:
+  void evalFloat() const;
+  void evalQuantized() const;
+
+private:
+  int32_t _x_multiplier = 0;
```
a-zh, line 44: The first underscore symbol can be reserved by system libraries for macros, function, global variable names.

compiler/luci-interpreter/src/kernels/Greater.test.cpp
```diff
@@ -0,0 +1,215 @@
...
+namespace kernels
+{
+namespace
+{
+
+using namespace testing;
+
+TEST(GreaterTest, FloatSimple)
+{
+  std::vector<float> x_data{
```
a-zh, line 32: Maybe add the `using std::vector`? It's ok for .cpp I mean.
And add explicit `#include <vector>`?

compiler/luci-interpreter/src/kernels/Greater.test.cpp
```diff
@@ -0,0 +1,215 @@
...
+      0.5, 0.7, 0.9, // Row 1
+      1,   0,   -1,  // Row 2
+      -1,  0,   1,   // Row 3
+  };
+
+  std::vector<float> y_data{
+      0.9, 0.7, 0.5, // Row 1
+  };
+
+  std::vector<bool> ref_output_data{
```
a-zh, line 71: As I remember `vector<bool>` has a some fundamental issues. Please, look at this point https://alenacpp.blogspot.com/2005/06/vector.html

compiler/luci-interpreter/src/kernels/Greater.test.cpp
```diff
@@ -0,0 +1,215 @@
...
+  kernel.configure();
+  kernel.execute();
+
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray({1, 3, 4, 1}));
+  EXPECT_THAT(extractTensorData<bool>(output_tensor), ::testing::ElementsAreArray(ref_output_data));
+}
+
+TEST(GreaterTest, Input_Type_Mismatch_NEG)
+{
+  Tensor x_tensor = makeInputTensor<DataType::FLOAT32>({1}, {1.f});
```
a-zh, line 195: Why we use mixing code with native `float` and `DataType::FLOAT32`? Its OK? Should we check the platform architecture and support cross compiler?

compiler/luci-interpreter/src/kernels/Greater.h
ai-moiseev: It seems to be a project convention to use underscore for private fields, as other kernel implementations follow this notation.

s-barannikov: It is reserved when it is followed by a capital case letter or by another underscore.

compiler/luci-interpreter/src/kernels/Greater.test.cpp
s-barannikov: It has one issue - you can't get a reference to an element of the vector. But we don't need to.

compiler/luci-interpreter/src/kernels/Greater.test.cpp
s-barannikov: Ideally, we should check it, but if we need to support an architecture which does not have IEEE-754 floats we have much more problems than just that.

compiler/luci-interpreter/src/kernels/Greater.test.cpp
s-barannikov: AFAIK `ElementsAreArray` also works with other containers (like std::array), `std::initializer_list` and even with C arrays.
But if we want to switch to it, we should do it in all tests.

compiler/luci-interpreter/src/kernels/Greater.test.cpp
ai-moiseev: `using namespace` is totally fine in cpps, but I've decided to stick with existing test cases namespace verbose formatting style. Maybe more refactoring for all testcases should be done, but it's not strictly necessary IMO.

compiler/luci-interpreter/src/kernels/Equal.cpp
```diff
@@ -0,0 +1,113 @@
...
+Equal::Equal(const Tensor *x, const Tensor *y, Tensor *output) : Kernel({x, y}, {output}) {}
+
+void Equal::configure()
+{
+  LUCI_INTERPRETER_CHECK(x()->element_type() == y()->element_type());
+  LUCI_INTERPRETER_CHECK(output()->element_type() == DataType::BOOL);
+
+  if (x()->element_type() == DataType::U8)
+  {
+    quantizeMultiplier(output()->scale() / x()->scale(), &_x_multiplier, &_x_shift);
```
s-barannikov, line 39: This looks wrong. Output has type `bool`, how can it have `scale`?

compiler/luci-interpreter/src/kernels/Equal.test.cpp
```diff
@@ -0,0 +1,184 @@
...
+      0.9, 0.6, 0.55, 0.5, // Row 1
+      -1,  0,   0.05, 1,   // Row 2
+  };
+
+  std::vector<bool> ref_output_data{
+      false, true, false, false, // Row 1
+      false, true, true,  false, // Row 2
+  };
+
+  std::pair<float, int32_t> quant_param = quantizationParams<uint8_t>(F_MIN, F_MAX);
```
s-barannikov, line 112: `x` and `y` should have different quantization parameters (to make the test stronger and catch mistakes like above).

compiler/luci-interpreter/src/kernels/Equal.cpp
```diff
@@ -0,0 +1,113 @@
...
+void Equal::evalQuantized() const
+{
+  const auto x_data = getTensorData<uint8_t>(x());
+  const auto y_data = getTensorData<uint8_t>(y());
+  auto output_data = getTensorData<bool>(output());
+
+  tflite::ComparisonParams op_params;
+  op_params.left_shift = 8;
+  op_params.input1_offset = -x()->zero_point(); // Note the '-'
+  op_params.input1_shift = -_x_shift;           // Note the '-'
```
jinevening, line 91: Is it necessary to negate shift?

compiler/luci-interpreter/src/kernels/Equal.cpp
jinevening, line 39: There are some checks about the scale of x and y and the shift value in TFLite code (See `QuantizeMultiplierSmallerThanOneExp`). Is it safe to skip the checks?

ai-moiseev: Fixed it, thanks. Got mixed up input and output scaling parameters, output has no longer effect on input scaling.

compiler/luci-interpreter/src/kernels/Equal.test.cpp
ai-moiseev: Added different quantization parameters in Equal/NotEqual tests. Less/Greater/etc. already had it in `UInt8QuantizedRescale` tests.

compiler/luci-interpreter/src/kernels/Equal.cpp
ai-moiseev: Not anymore, wrong input scale calculation was used, fixed it.

compiler/luci-interpreter/src/kernels/Equal.cpp
ai-moiseev: Done, I've used `quantizeMultiplierSmallerThanOneExp` here, it should provide required checks.

compiler/luci-interpreter/src/kernels/GreaterEqual.test.cpp
```diff
@@ -0,0 +1,215 @@
...
+      false, true,  true, true,  // Row 1
+      true,  false, true, false, // Row 2
+  };
+
+  std::pair<float, int32_t> quant_param = quantizationParams<uint8_t>(F_MIN, F_MAX);
+  Tensor x_tensor =
+      makeInputTensor<DataType::U8>({1, 2, 4, 1}, quant_param.first, quant_param.second, x_data);
+  Tensor y_tensor =
+      makeInputTensor<DataType::U8>({1, 2, 4, 1}, quant_param.first, quant_param.second, y_data);
+  Tensor output_tensor = makeOutputTensor(DataType::BOOL, quant_param.first, quant_param.second);
```
s-barannikov, line 115: bool can't be quantized.
There are several identical places in the PR.

compiler/luci-interpreter/src/kernels/Greater.test.cpp
```diff
@@ -0,0 +1,215 @@
...
+{
+  std::vector<float> x_data{
+      0.5, 0.6, 0.7,  0.9, // Row 1
+      1,   0,   0.05, -1,  // Row 2
+  };
+
+  std::vector<float> y_data{
+      0.9, 0.7,  0.6, 0.5, // Row 1
+      -1,  0.05, 0,   1,   // Row 2
+  };
```
s-barannikov, line 135: There should be a case with x == y and result == **false**.

compiler/luci-interpreter/src/kernels/GreaterEqual.test.cpp
```diff
@@ -0,0 +1,215 @@
...
+{
+  std::vector<float> x_data{
+      0.5, 0.6, 0.7,  0.9, // Row 1
+      1,   0,   0.05, -1,  // Row 2
+  };
+
+  std::vector<float> y_data{
+      0.9, 0.7,  0.6, 0.5, // Row 1
+      -1,  0.05, 0,   1,   // Row 2
+  };
```
s-barannikov, line 135: There should be a case with x == y and result == **true**.
Please check other tests too.

compiler/luci-interpreter/src/kernels/GreaterEqual.test.cpp
ai-moiseev: Done, removed quantization parameters from `makeOutputTensor` calls.

compiler/luci-interpreter/src/kernels/Greater.test.cpp
ai-moiseev: Done

compiler/luci-interpreter/src/kernels/GreaterEqual.test.cpp
ai-moiseev: Done, updated comparison cases to add cases with comparison of equal values

s-barannikov (approved):
LGTM, thank you

jinevening (approved):
LGTM

a-zh (approved):
LGTM

PR #4341, opened 2020-09-18
[onert] Add Keyfunction to IPortableTensor to resolve NDK's dynamic_cast issue

dayo09:
- Related Issue
  - https://github.com/Samsung/ONE/issues/4157
  - https://github.sec.samsung.net/STAR/nnfw/issues/11762

### DIFF
--- a/runtime/onert/core/src/backend/cpu_common/Tensor.cc
+++ b/runtime/onert/core/src/backend/cpu_common/Tensor.cc
@@ -26,7 +26,7 @@ namespace backend
 namespace cpu_common
 {
 
-Tensor::~Tensor() = default;
+Tensor::~Tensor() {}
 
 size_t Tensor::calcOffset(const ir::Coordinates &coords) const
 {
--- a/runtime/onert/backend/cpu/Tensor.cc
+++ b/runtime/onert/backend/cpu/Tensor.cc
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


runtime/onert/core/src/backend/cpu_common/Tensor.cc
```diff
@@ -23,10 +23,13 @@ namespace onert
 {
 namespace backend
 {
+// `dynamic_cast` not working across library boundaries on NDK
+// With this as a key function, `dynamic_cast` works across dl
+IPortableTensor::~IPortableTensor() {}
```
dayo09, line 28: I am sure that this works as a keyfunction successfully but I am worried if defining super class's function in child class like this.. @wateret

hyunsik-yoon:
I wonder if we can have something like `~Tensor() = default;` in `.h` file instead of introducing an empty body in a separate `.cc` file. Could you explain the reason a bit? Thanks.

runtime/onert/core/src/backend/cpu_common/Tensor.cc
```diff
@@ -26,7 +26,7 @@ namespace backend
 namespace cpu_common
 {
 
-Tensor::~Tensor() = default;
+Tensor::~Tensor() {}
```
wateret, line 29: I thought those are same.

wateret (approved):
LGTM

wateret:
> I wonder if we can have something like `~Tensor() = default;` in `.h` file instead of introducing an empty body in a separate `.cc` file. Could you explain the reason a bit? Thanks.

@hyunsik-yoon This is intended, just to create a key function. Please refer to #4157 .

hyunsik-yoon:
@wateret Ah. `non-inlined`. Thanks. :-)

periannath (approved):
LGTM

PR #4347, opened 2020-09-21
[pota-value-test] Add more single Op tests

jinevening:
This adds more single Op tests


jinevening:
This adds five tests for quantizing a single Op model. See `test.lst` only.

seanshpark:
There are lots of .txt files with all numbers. How did you made these? 
I'd like to know if there can be any way to add a program to generate these instead of adding these files.

seanshpark (approved):
LGTM

jinevening:
> There are lots of .txt files with all numbers. How did you made these?
I'd like to know if there can be any way to add a program to generate these instead of adding these files.

@seanshpark I generated them using some Python scripts and dalgona. I'll add explanation on how to generate test data in README.

PR #4349, opened 2020-09-21
[onert] Enable dynamic_cast assert for android in polymorphic_downcast

periannath:
- This commit enables dynamic_cast assert for android in polymorphic_downcast
  - Check for `__ANDROID__` macro is added to avoid assertion #3654
  - dynamic_cast now works well on android #4341

### DIFF
--- a/runtime/libs/misc/include/misc/polymorphic_downcast.h
+++ b/runtime/libs/misc/include/misc/polymorphic_downcast.h
@@ -27,9 +27,7 @@ namespace misc
 
 template <typename DstType, typename SrcType> inline DstType polymorphic_downcast(SrcType *x)
 {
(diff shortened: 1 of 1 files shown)


YongseopKim:
This should be tested on the internal app by measuring memory.

periannath:
@YongseopKim This commit just removes macro that disables assert check on android. It doesn't affect any program logic at all. `FullyConnectedLayer` still uses `polymorphic_downcast` and it uses `static_cast`.

YongseopKim:
@periannath Thank you for your confirmation. Because of [this](https://github.sec.samsung.net/STAR/nnfw/pull/11718), I asked.

hseok-oh (approved):

PR #4356, opened 2020-09-21
[bcq-tools] Add missing license in bcq-tools

llFreetimell:
This commit will add missing license in `bcq-tools`

### DIFF
--- a/compiler/bcq-tools/generate_bcq_output_arrays
+++ b/compiler/bcq-tools/generate_bcq_output_arrays
@@ -1,5 +1,20 @@
 #!/usr/bin/env python3
 
+# Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+# Copyright (C) 2018 The TensorFlow Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


compiler/bcq-tools/generate_bcq_output_arrays
```diff
@@ -1,5 +1,20 @@
 #!/usr/bin/env python3
 
+# Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+# Copyright (C) 2018 The TensorFlow Authors
```
seanshpark, line 4: does this code has direct copy from tensorflow?

llFreetimell: There is no direct code from Tensorflow but the codes import TensorFlow. So I thought that we should write the license author like `tf2tfliteV2.py`

seanshpark: tf2tfliteV2.py has copied code from TensorFlow :)
If there is no direct copy code then we can skip this if you agree.

llFreetimell: Oh, I didn't know that @.@
Frankly speaking, I totally agree with you but I want to be checked that we are really correct.
So after it is checked, I will remove it :)

llFreetimell:
This will be handled in another PR. Close this!

PR #4357, opened 2020-09-21
[luci] Ignore shape and dtype check for CircleOutputExclude

llFreetimell:
`CircleOutputExclude` is for not exporting graph output.
Therefore shape and dtype check for `CircleOutputExclude` is not needed.


seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #4358, opened 2020-09-21
[luci-interpreter]Update test case for Logistic

struss:
This commit update test case for `Logistic` on luci-interpreter

1. Add Uint8 Test case.
2. Add Negative Test case.


s-barannikov (approved):
LGTM

PR #4363, opened 2020-09-21
[luci] Fix core algorithm in FuseBCQPass

llFreetimell:
This commit will revise `FuseBCQPass` algorithm to handle following things.
- In case of input shape has unknown dimension, `input_hidden_size` and `weights_hidden_size` is set by constant tensor, not input.
- `Transpose` operation should be inserted only after `BCQGather` when `do_w_x` is false.


seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #4367, opened 2020-09-22
[common-artifacts]Exclude from example creation.

struss:
Parent Issue : #4151
Draft PR : #4260

This commit add `UnidirectionalSequenceLSTM` to exclusion list using TC generation.

### DIFF
--- a/compiler/common-artifacts/exclude.lst
+++ b/compiler/common-artifacts/exclude.lst
@@ -5,6 +5,8 @@
 
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
+optimize(UnidirectionalSequenceLSTM_000)
+optimize(UnidirectionalSequenceLSTM_001)
 optimize(Unique_000)
 optimize(Unique_001)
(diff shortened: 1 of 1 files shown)


mhs4670go (approved):
LGTM

compiler/common-artifacts/exclude.lst
```diff
@@ -5,6 +5,8 @@
 
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
+optimize(UnidirectionalSequenceLSTM_000)
+optimize(UnidirectionalSequenceLSTM_001)
```
mhs4670go, line 9: Isn't it possible to optimize?

struss: I will check this. Thanks for pointing out.

struss: I found something to fix on `luci/import` and `tflchef` for running through circle2circle. after all fix it, I will update this file!

struss:
I will close this and reopen after update `tflchef` and `luci`.

Issue #4352, opened 2020-09-21
Calculated when axis is -1 in reduce operation.


YiHyunJin:
```
inline bool ReduceImpl(const In *input_data, const Shape &input_shape, const Shape &,
                       const int *axis, const int num_axis, int *input_iter,
                       Out reducer(const Out current, const In in), Out *output_data)
{
  const auto input_dims = input_shape.DimsData();
  const auto input_num_dims = input_shape.DimensionsCount();

  if (num_axis == 1 && (axis[0] == -1 || axis[0] == input_num_dims - 1)) {
    int input_size = 1;
    int reduce_size = 0;
    for (int idx = 0; idx < input_num_dims - 1; idx++) {
      input_size *= input_dims[idx];
    }
    reduce_size = input_dims[input_num_dims - 1];
    for (int idx = 0; idx < input_size; idx++) {
      for (int r_idx = 0; r_idx < reduce_size; r_idx++) {
        if (r_idx == 0) {
          output_data[idx] = input_data[idx * reduce_size];
        }
        else {
          output_data[idx] = reducer(output_data[idx], input_data[idx * reduce_size + r_idx]);
        }
      }
    }
    return true;
  }
....
....
....
```

How about this?

If the axis is -1, we can access the index simply as above.

So I added this, And tested any models. The result is an 8% increase in speed.
(1.446 -> 1.340)

glistening:
> How about this?
> 
> If the axis is -1, we can access the index simply as above.
> 
> So I added this, And tested any models. The result is an 8% increase in speed.
> (1.446 -> 1.340)

Great! What about create a PR and get review? Could you please remove me from assignee? Perhaps you just want to notify this issue.

PR #4368, opened 2020-09-22
[onert] Calculated when axis is -1 in reduce operation

YiHyunJin:
If the axis is -1, we can access the index simply.
So I added this, And tested any models. The result is an 8% increase in speed.
(1.446 -> 1.340)

### DIFF
--- a/compute/cker/include/cker/operation/Reduce.h
+++ b/compute/cker/include/cker/operation/Reduce.h
@@ -39,6 +39,32 @@ inline bool ReduceImpl(const In *input_data, const Shape &input_shape, const Sha
   const auto input_num_dims = input_shape.DimensionsCount();
 
   // Reset input iterator.
+  if (num_axis == 1 && (axis[0] == -1 || axis[0] == input_num_dims - 1))
+  {
+    int input_size = 1;
+    int reduce_size = 0;
+    for (int idx = 0; idx < input_num_dims - 1; idx++)
+    {
+      input_size *= input_dims[idx];
+    }
+    reduce_size = input_dims[input_num_dims - 1];
+    for (int idx = 0; idx < input_size; idx++)
+    {
+      for (int r_idx = 0; r_idx < reduce_size; r_idx++)
+      {
+        if (r_idx == 0)
+        {
+          output_data[idx] = input_data[idx * reduce_size];
+        }
+        else
+        {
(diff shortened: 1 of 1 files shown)


YiHyunJin:
https://github.com/Samsung/ONE/issues/4352

compute/cker/include/cker/operation/Reduce.h
```diff
@@ -39,19 +39,47 @@ inline bool ReduceImpl(const In *input_data, const Shape &input_shape, const Sha
...
+        {
+          output_data[idx] = reducer(output_data[idx], input_data[idx * reduce_size + r_idx]);
+        }
+      }
+    }
+    return true;
   }
-  // Iterate through input_data.
-  do
+  else
```
glistening, line 67: @YiHyunJin I think we don't need `else` because the special case of `axis = -1` is already handled and early exited.

glistening:
@ragmani or/and @hseok-oh Could you please take a look? As I remember, you worked for reduce operation.

compute/cker/include/cker/operation/Reduce.h
```diff
@@ -39,19 +39,47 @@ inline bool ReduceImpl(const In *input_data, const Shape &input_shape, const Sha
   const auto input_num_dims = input_shape.DimensionsCount();
 
   // Reset input iterator.
-  for (int idx = 0; idx < input_num_dims; ++idx)
+  if (num_axis == 1 && (axis[0] == -1 || axis[0] == input_num_dims - 1))
```
hseok-oh, line 42: `axis[0] == -1` is needless because `ResolveAxis()` makes `-1` to `input_num_dims - 1`.

wateret:
> So I added this, And tested any models. The result is an 8% increase in speed.
> (1.446 -> 1.340)

@YiHyunJin 👍 Could you let us know which model it was(if you can tell)? And is that 8% for the whole model execution? not only Reduce operations

YiHyunJin:
> @YiHyunJin 👍 Could you let us know which model it was(if you can tell)? And is that 8% for the whole model execution? not only Reduce operations
I can't tell you which model it is.. :(
And 8% is whole model execution.

glistening (approved):
LGTM

hseok-oh (approved):

PR #4369, opened 2020-09-22
[cmn-art] enable Unique optimize

seanshpark:
This will remove Unique op from exclude list as it works


mhs4670go (approved):
LGTM

Issue #4373, opened 2020-09-22
[Compiler][luci] Introduce new optimization pass, FoldDequantize


llFreetimell:
Parent Issue : #4353

---

### What to do

Remove `Dequantize` operation when following pattern is appeared.

#### Pattern 1
- `Gather` is input of `Dequantize`
- Quantized constant value is input of `Gather`
![Dequant](https://user-images.githubusercontent.com/24720192/93846447-016c4200-fcdf-11ea-9ed2-6276d6122611.png)

#### Pattern 2
- Input of `Dequqntize` is constant value

### How it works

`Gather` operation do not change the value itself. So we can apply `Dequantize` operation in advance and thus we can substitute input of `Gather` to dequantized constant value.

### Todo
- [x] Support `Dequantize` operation in Compiler FE (Only some of them are mandatory) #4382
- [x] Introduce `FoldDequantize` pass in `luci/pass` #4404
- [x] ~~Reorder optimizations in circle optimizer #4423~~ Not needed
- [x] Support `S8` in `CircleTensorExporter` #4453
- [x] Introduce `FoldDequantize` pass in `circle2circle` #4448

### Draft

#4374

llFreetimell:
Done!

PR #4374, opened 2020-09-22
[Draft][Compiler] Introduce Dequantize op and new optimization

llFreetimell:
Parent Issue : #4373

This draft is for introducing `Dequantize` opeartion in `luci`
and related new optimization pass, `FoldDequantize`.


llFreetimell:
/cc @YongseopKim 

```sh
ONE$ nncc configure && nncc build
ONE$ build/compiler/circle2circle/circle2circle --all /path/to/model /path/to/output/model
```

compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -96,6 +96,12 @@ int entry(int argc, char **argv)
       .default_value(false)
       .help("This will convert Custom(Matmul) to Matmul operator");
 
+  arser.add_argument("--fold_dequantize")
```
seanshpark, line 99: Q) if this his folding, there would be a result of folding. what would the result be?
or is this like remove `dequantize` ?

compiler/luci/pass/src/FoldDequantize.cpp
```diff
@@ -0,0 +1,122 @@
...
+                    ((double)const_node->at<loco::DataType::S8>(i) -
+                     const_node->quantparam()->zerop.at(qd)) *
+                    const_node->quantparam()->scale.at(qd);
+              else if (const_node->dtype() == loco::DataType::U8)
+                circle_output->at<loco::DataType::FLOAT32>(i) =
+                    ((double)const_node->at<loco::DataType::U8>(i) -
+                     const_node->quantparam()->zerop.at(qd)) *
+                    const_node->quantparam()->scale.at(qd);
+              else
+                assert(false && "unsupported fold dequant type");
```
seanshpark, line 108: it would be better not to assert, just use LOG and return false (or do nothing)

compiler/circle2circle/src/Circle2Circle.cpp
llFreetimell: As mentioned in https://github.com/Samsung/ONE/issues/4373#issue-706091190, it is to remove `Dequantize`.

compiler/luci/pass/src/FoldDequantize.cpp
llFreetimell: Yeap, for now I was just to share the working codes so most of those things will be modified better than now :)

compiler/circle2circle/src/Circle2Circle.cpp
seanshpark: If this is remove, could you please use `remove` , `eliminate` or any other words to reflect it's removed?

llFreetimell: This pass is removing `Dequantize` but main characteristic is folding constant nodes which is used by `Dequantize` by pre-calculating.
If we name it like `RemoveDequantize` or `EliminateDequantize`, it seems that this pass is just removing `Dequantize` without any other actions.
Also, I think that **fold** is reflecting the elimination characteristic too,  just like Constant**Fold** :)

seanshpark: OK, thanks for the explanation.

llFreetimell:

PR #4377, opened 2020-09-22
[tflchef]Update UnidirectionalSequenceLSTM

struss:
Parent Issue : #4151
Draft PR : #4260

This commit update filler of `UnidirectionalSequenceLSTM` to fill all tensors if exciplit value provided.

### DIFF
--- a/compiler/tflchef/tflite/src/Op/UnidirectionalSequenceLSTM.cpp
+++ b/compiler/tflchef/tflite/src/Op/UnidirectionalSequenceLSTM.cpp
@@ -30,12 +30,6 @@ void TFliteOpUnidirectionalSequenceLSTM::filler(const tflite::Operator *op, TFli
 
   for (int32_t i = 0; i < inputs.size(); i++)
   {
-    // Except for Input 0, 17 and 18.
-    // Each Input mean Input[0](=Input Tensor), Input[17](=OutputState Tensor) and
-    // Input[18](=CellState Tensor).
-    // This could be updated from previous input or User Given data, so This could not be Const
-    if (i == 0 || i == 17 || i == 18)
-      continue;
     if (inputs[i] != -1)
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

compiler/tflchef/tflite/src/Op/UnidirectionalSequenceLSTM.cpp
```diff
@@ -30,12 +30,6 @@ void TFliteOpUnidirectionalSequenceLSTM::filler(const tflite::Operator *op, TFli
 
   for (int32_t i = 0; i < inputs.size(); i++)
   {
-    // Except for Input 0, 17 and 18.
-    // Each Input mean Input[0](=Input Tensor), Input[17](=OutputState Tensor) and
-    // Input[18](=CellState Tensor).
-    // This could be updated from previous input or User Given data, so This could not be Const
-    if (i == 0 || i == 17 || i == 18)
-      continue;
```
mhs4670go, line 38: As comment said, this could not be CONST. Have it changed?

mhs4670go: This might as well be merged as we're not sure about this. It could be changed after testing with real model.

mhs4670go (approved):
LGTM

compiler/tflchef/tflite/src/Op/UnidirectionalSequenceLSTM.cpp
struss: @mhs4670go I have a little question. is there a way to create variable tensor using tflchef?

struss:
On precise check, state is composed of variable tensor. so, implementation above is right(I thought) so I close this as temporary.

PR #4375, opened 2020-09-22
DRAFT: Fuse Add Op with TConv Op

jinevening:
On going draft to fuse Add Op with TConv Op

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

PR #4378, opened 2020-09-22
[luci] Fuse Add Op with TConv Op

jinevening:
This fuses Add Op with TConv Op

---
Draft PR: #4375

### DIFF
--- a/compiler/luci/pass/src/FuseAddWithTConvPass.cpp
+++ b/compiler/luci/pass/src/FuseAddWithTConvPass.cpp
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FuseAddWithTConvPass.h"
+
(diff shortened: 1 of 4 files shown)


compiler/luci/pass/src/FuseAddWithTConvPass.cpp
```diff
@@ -0,0 +1,114 @@
...
+  if (not add)
+    return false;
+  if (add->dtype() != loco::DataType::FLOAT32)
+    return false;
+  if (add->fusedActivationFunction() != luci::FusedActFunc::NONE &&
+      add->fusedActivationFunction() != luci::FusedActFunc::RELU6)
+    return false;
+
+  // get addition
+  auto addition = dynamic_cast<luci::CircleConst *>(add->y());
```
seanshpark, line 62: can this be `add->x()` too?

compiler/luci/pass/src/FuseAddWithTConvPass.cpp
```diff
@@ -0,0 +1,114 @@
...
+bool FuseAddWithTConvPass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto tconv = dynamic_cast<luci::CircleTransposeConv *>(node);
+    if (not tconv)
+      continue;
+
+    changed |= fuse_add_with_tconv(tconv);
```
seanshpark, line 108: (optional) I think this was copied from other passes, but as `changed` is bool and this `|=` is bitwise operation, it's a bit strange... can you revise this as for boolean value?

compiler/luci/pass/src/FuseAddWithTConvPass.cpp
jinevening: I think so, because add is commutative. I've updated the code. Thanks!

compiler/luci/pass/src/FuseAddWithTConvPass.cpp
jinevening: I've updated the code. PTAL.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM thank you!

PR #4379, opened 2020-09-22
[res] Add recipe and test rule for FuseAddWithTConvPass

jinevening:
This adds recipes and test rules for FuseAddWithTConvPass

---
Draft PR: #4375

### DIFF
--- a/res/TensorFlowLiteRecipes/Net_TConv_Add_001/test.recipe
+++ b/res/TensorFlowLiteRecipes/Net_TConv_Add_001/test.recipe
@@ -1,0 +1,119 @@
+operand {
+  name: "filter"
+  type: FLOAT32
+  shape {
+    dim: 1
+    dim: 3
+    dim: 3
+    dim: 2
+  }
+  filler {
+    tag: "gaussian"
+    arg: "0.0"
+    arg: "0.1"
+  }
+  quant {
+    quantized_dimension: 0
+  }
+}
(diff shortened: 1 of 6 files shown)


jinevening:
There are two recipes. One for testing a positive case, where add is fused with tconv. Another one is for testing a negative case, where add is not fused because the dimension of add operand does not match with the first dimension of tconv filter.

res/TensorFlowLiteRecipes/Net_TConv_Add_001/test.recipe
```diff
@@ -0,0 +1,119 @@
...
+  transpose_conv_options {
+    padding: VALID
+    stride_w: 1
+    stride_h: 1
+  }
+}
+operation {
+  type: "Add"
+  input: "conv2d_transpose"
+  input: "Addition_add_param"
```
seanshpark, line 112: (optional) maybe we could add another recipe with two inputs swapped
```
operation {
  type: "Add"
  input: "Addition_add_param"
  input: "conv2d_transpose"
```

jinevening: I've added a new recipe for checking commutativity. Thanks.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #4389, opened 2020-09-23
[circle2circle] Fuse add with tconv

jinevening:
This adds --fuse_add_with_tconv option

---
Draft PR: #4375

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -60,6 +60,12 @@ int entry(int argc, char **argv)
   arser.add_argument("--all").nargs(0).required(false).default_value(false).help(
       "Enable all optimize options");
 
+  arser.add_argument("--fuse_add_with_tconv")
+      .nargs(0)
+      .required(false)
+      .default_value(false)
+      .help("This will fuse Add operator to Transposed Convolution operator");
+
   arser.add_argument("--fuse_batchnorm_with_tconv")
       .nargs(0)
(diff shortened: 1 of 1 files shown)


mhs4670go (approved):
LGTM

compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -130,6 +136,8 @@ int entry(int argc, char **argv)
     options->enable(Algorithms::ResolveCustomOpBatchMatMul);
     options->enable(Algorithms::ResolveCustomOpMatMul);
   }
+  if (arser.get<bool>("--fuse_add_with_tconv"))
+    options->enable(Algorithms::FuseAddWithTConv);
```
seanshpark, line 140: Is it possible to add also to `--all` ?

seanshpark: @mhs4670go , was there any particular reason for `fuse_batchnorm_with_tconv` not added to `all` ?

mhs4670go: @seanshpark When we were discussing the pass, we decided to add it to `--all` section as the path became more mature. There are still few places to use this path, and I'm gonna put it in after some more tests are added, which by the way, is the reason why I don't close related issue yet.

seanshpark: OK, thanks!

seanshpark (approved):
LGTM

PR #4392, opened 2020-09-23
[tf2tfliteV2] Revise option for I/O nodes

seanshpark:
This will revise tf2tfliteV2.py input_arrays and output_arrays be optional as saved_model and keras don't use them but are required only for graphdef


mhs4670go (approved):
LGTM

PR #4394, opened 2020-09-23
[DRAFT] Introduce Sparsifier

mhs4670go:
This draft is to implement Sparsifier.

Related : #3889

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -123,6 +123,41 @@ int entry(int argc, char **argv)
   arser.add_argument("input").nargs(1).type(arser::DataType::STR).help("Input circle model");
   arser.add_argument("output").nargs(1).type(arser::DataType::STR).help("Output circle model");
 
+  // sparsification argument
+  arser.add_argument("--sparsify_tensor")
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .required(false)
+      .help("Tensor name that you want to sparsify");
+
+  arser.add_argument("--sparsify_traversal_order")
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .required(false)
+      .default_value("0,1,2,3")
+      .help("Traversal order of dimensions. Default value: 0,1,2,3");
+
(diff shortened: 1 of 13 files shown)


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -117,6 +117,41 @@ int entry(int argc, char **argv)
   arser.add_argument("input").nargs(1).type(arser::DataType::STR).help("Input circle model");
   arser.add_argument("output").nargs(1).type(arser::DataType::STR).help("Output circle model");
 
+  // sparsification argument
+  arser.add_argument("--sparsify_tensor")
```
seanshpark, line 121: It would be better to make multiple options to one option.

seanshpark: Q) is sparsify be possible for multiple tensors? then do we have to run multiple times of circle2circle for this?

seanshpark: I'm not sure but thought of something like this;
```
--sparsify="TheTensor;NameForSparsify,0,d,10,0/Another;Tensor;Name,0,d,20,1"
```

mhs4670go: @seanshpark 
> Q) is sparsify be possible for multiple tensors? then do we have to run multiple times of circle2circle for this?

Yes. User passes specific tensor name that he/she wants to sparsify.

> It would be better to make multiple options to one option.

Actually, it is a little bit hard to make them one because the length of its array can be different for each tensor. To make it less verbose, I set a default value that is commonly used. Moreover, I will introduce a short option to `arser`.

seanshpark:
From #4486, I could imagine some negative test cases;
- with absent options
- with wrong options, like `Sparsify_format=,,,,`
- none existing tensor name for `Sparsify_tensor_name`
- `Sparsify_tensor_name` tensor with INT8 tensor

If possible, could add some tests (unit test is nice, or one-cmd test is also nice) later ?

mhs4670go:
@seanshpark Okay. I'll test what you listed:)

PR #4404, opened 2020-09-24
[luci] Introduce FoldDequantize pass in luci

llFreetimell:
Parent Issue : #4373
Draft : #4374

This commit will introduce `FoldDequantize` pass in `luci`.

### DIFF
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -16,6 +16,7 @@
 
 #include "luci/CircleOptimizer.h"
 
+#include "luci/Pass/FoldDequantizePass.h"
 #include "luci/Pass/FuseAddWithTConvPass.h"
 #include "luci/Pass/FuseBatchNormWithTConv.h"
 #include "luci/Pass/FuseBCQPass.h"
@@ -136,6 +137,10 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   if (_options->query(Options::Algorithm::FuseAddWithTConv))
   {
     phase.emplace_back(std::make_unique<FuseAddWithTConvPass>());
+  }
+  if (_options->query(Options::Algorithm::FoldDequantize))
+  {
+    phase.emplace_back(std::make_unique<luci::FoldDequantizePass>());
   }
 
   // Shape inference is needed for added nodes doing above transformations
--- a/compiler/luci/pass/src/FoldDequantizePass.cpp
+++ b/compiler/luci/pass/src/FoldDequantizePass.cpp
@@ -1,0 +1,200 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FoldDequantizePass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+#include <loco/Service/TypeInference.h>
+
+namespace
+{
+
+bool is_hybrid_kernel_supported(loco::Node *node)
+{
+  if (dynamic_cast<luci::CircleFullyConnected *>(node) != nullptr)
+    return true;
+
+  return false;
+}
+
+bool is_foldable_const(luci::CircleConst *node)
+{
+  if (node->dtype() == loco::DataType::S8)
+    return true;
+  if (node->dtype() == loco::DataType::U8)
+    return true;
+
+  return false;
+}
+
+luci::CircleConst *dequantized_const_node(luci::CircleConst *const_node)
+{
+  assert(const_node->quantparam() != nullptr);
+
+  auto g = const_node->graph();
+  auto new_const_node = g->nodes()->create<luci::CircleConst>();
+
+  new_const_node->dtype(loco::DataType::FLOAT32);
+  new_const_node->rank(const_node->rank());
+  uint32_t dim_size = 1;
+  for (uint32_t i = 0; i < new_const_node->rank(); ++i)
+  {
+    new_const_node->dim(i) = const_node->dim(i);
+    dim_size *= const_node->dim(i).value();
+  }
+  new_const_node->size<loco::DataType::FLOAT32>(dim_size);
+  new_const_node->shape_status(luci::ShapeStatus::VALID);
+
+  const int32_t q_dim = const_node->quantparam()->quantized_dimension;
+  const int32_t q_dim_value = const_node->dim(q_dim).value();
+
+  int32_t right_count = q_dim_value;
+  for (uint32_t i = q_dim + 1; i < const_node->rank(); ++i)
+    right_count *= const_node->dim(i).value();
+
+  if (const_node->dtype() == loco::DataType::S8)
+  {
+    for (uint32_t i = 0; i < const_node->size<loco::DataType::S8>(); ++i)
+    {
+      uint32_t qd = (i % right_count) / (right_count / q_dim_value);
+      if (qd >= const_node->quantparam()->zerop.size())
+        qd = 0;
+
+      new_const_node->at<loco::DataType::FLOAT32>(i) =
+          (float)(const_node->at<loco::DataType::S8>(i) - const_node->quantparam()->zerop.at(qd)) *
+          const_node->quantparam()->scale.at(qd);
+    }
+  }
+  else
+  {
+    for (uint32_t i = 0; i < const_node->size<loco::DataType::U8>(); ++i)
+    {
+      uint32_t qd = (i % right_count) / (right_count / q_dim_value);
+      if (qd >= const_node->quantparam()->zerop.size())
+        qd = 0;
+
+      new_const_node->at<loco::DataType::FLOAT32>(i) =
+          (float)((int)const_node->at<loco::DataType::U8>(i) -
+                  const_node->quantparam()->zerop.at(qd)) *
+          const_node->quantparam()->scale.at(qd);
+    }
+  }
+
+  return new_const_node;
+}
+
+bool replace_const_node(loco::Node *node, luci::CircleConst *const_node)
+{
+  if (auto gather = dynamic_cast<luci::CircleGather *>(node))
+  {
+    gather->params(dequantized_const_node(const_node));
+    gather->dtype(loco::DataType::FLOAT32);
+    return true;
+  }
+  else
+  {
+    // TODO Support more ops
+    return false;
+  }
+}
+
+} // namespace
+
+namespace luci
+{
+
+/**
+ *
+ * Folding pattern 1 - When input of Dequantize is foldable constant
+ *
+ * [Before]
+ *     quantized_const_input ---------- Dequantize ---------- Op ---
+ *                             +-- Op1_with_quant_input ---
+ *                             +-- Op2_with_quant_input ---
+ *
+ * [After]
+ *   dequantized_const_input -------------------------------- Op ---
+ *
+ *     quantized_const_input ----- Op1_with_quant_input ---
+ *                             +-- Op2_with_quant_input ---
+ *
+ *
+ * Folding pattern 2 - When input of Dequantize uses quantized output value
+ *
+ * [Before]
+ *     quantized_const_input ----- Gather ----- Dequantize --- Op ---
+ *                             +-- Op1_with_quant_input ---
+ *                             +-- Op2_with_quant_input ---
+ *
+ * [After]
+ *   dequantized_const_input ------Gather -------------------- Op ---
+ *
+ *     quantized_const_input ----- Op1_with_quant_input ---
+ *                             +-- Op2_with_quant_input ---
+ *
+ *
+ */
+bool FoldDequantizePass::run(loco::Graph *g)
+{
+  bool changed = false;
+
+  for (auto node : loco::all_nodes(g))
+  {
+    if (auto circle_dequant = dynamic_cast<luci::CircleDequantize *>(node))
+    {
+      if (auto const_input = dynamic_cast<luci::CircleConst *>(circle_dequant->input()))
+      {
+        // Pattern 1 - When input of Dequantize is foldable constant
+        if (is_foldable_const(const_input))
+        {
+          loco::replace(circle_dequant).with(dequantized_const_node(const_input));
+          changed = true;
+        }
+      }
+    }
+    else if (auto const_node = dynamic_cast<luci::CircleConst *>(node))
+    {
+      if (const_node->quantparam() != nullptr && is_foldable_const(const_node))
+      {
+        for (auto const_node_user : loco::succs(const_node))
+        {
+          // If user is hybrid kernel supported operation, do not dequantize
+          if (is_hybrid_kernel_supported(const_node_user))
+            continue;
+
+          auto users = loco::succs(const_node_user);
+          if (users.size() > 1)
+            continue;
+
+          // Pattern 2 - When input of Dequantize uses quantized output value
+          if (auto dequant = dynamic_cast<luci::CircleDequantize *>(*users.begin()))
+          {
+            if (replace_const_node(const_node_user, const_node))
+            {
+              loco::replace(dequant).with(const_node_user);
+              changed = true;
+            }
+          }
+        }
+      }
+    }
+  }
+
+  return changed;
+}
+
+} // namespace luci
--- a/compiler/luci/pass/include/luci/CircleOptimizer.h
+++ b/compiler/luci/pass/include/luci/CircleOptimizer.h
@@ -42,6 +42,7 @@ public:
       QuantizeDequantizeWeights,
       QuantizeWithMinMax,
       Requantize,
+      FoldDequantize,
     };
 
     enum AlgorithmParameters
--- a/compiler/luci/pass/include/luci/Pass/FoldDequantizePass.h
+++ b/compiler/luci/pass/include/luci/Pass/FoldDequantizePass.h
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_FOLD_DEQUANTIZE_PASS_H__
+#define __LUCI_FOLD_DEQUANTIZE_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to fold Dequantize, which can be folded by constant inputs
+ *
+ */
+struct FoldDequantizePass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::FOLD_DEQUANTIZE"; }
+
+  bool run(loco::Graph *g) final;
+};
+
+} // namespace luci
+
+#endif // __LUCI_FOLD_DEQUANTIZE_PASS_H__


compiler/luci/pass/src/FoldDequantize.cpp
```diff
@@ -0,0 +1,133 @@
...
+        changed = true;
+      }
+    }
+    else if (auto const_node = dynamic_cast<luci::CircleConst *>(node))
+    {
+      if (const_node->quantparam() != nullptr && is_foldable_const(const_node))
+      {
+        for (auto s : loco::succs(const_node))
+        {
+          if (!is_quant_supported_op(s))
```
jinevening, line 71: AFAIK, successor of const node should be `dequantize` to apply this pass (Am I right?).

Plus, if the const_node is used by anther Op, const values should not be dequantized.

jinevening: AFAIK, this pass should check the below pattern.

```
const - gather (or other Ops that do not change the values ) - dequantize
```

This code does not seem to check the pattern (Am I wrong?).

llFreetimell: You are right. For now, this pass only checks whether the operation among `Dequantize` and `Const` is `FullyConnected` or not. The list of ops which quantized value can be used is written in `is_quant_supported_op`.

If the operation is not `FullyConnected`, this pass will fold the constant inputs.

llFreetimell: I changed the code as you said :)

llFreetimell:
@jinevening Sorry for late, I applied your comment :) PTAL!

compiler/luci/pass/src/FoldDequantize.cpp
```diff
@@ -0,0 +1,158 @@
...
+bool FoldDequantizePass::run(loco::Graph *g)
+{
+  bool changed = false;
+
+  for (auto node : loco::all_nodes(g))
+  {
+    if (auto circle_dequant = dynamic_cast<luci::CircleDequantize *>(node))
+    {
+      auto input_node = circle_dequant->input();
+      if (auto const_input = dynamic_cast<luci::CircleConst *>(input_node))
```
jinevening, line 126: `const_input` must not be used by another Op, which means `loco::succs(const_input)` should return only one element `circle_dequant`. Otherwise, dequantize should not be folded.

compiler/luci/pass/src/FoldDequantize.cpp
```diff
@@ -0,0 +1,158 @@
...
+      {
+        for (auto s : loco::succs(const_node))
+        {
+          if (!is_hybrid_supported_op(s))
+          {
+            // Input can be pre-dequantized
+            changed |= replace_const_node(s, dequantized_const_node(const_node));
+          }
+        }
+      }
```
jinevening, line 151: Similar comment here. Dequantize must be folded only when `const_node` is used by one Op (`s`), and `s` is used by dequantize Op.

compiler/luci/pass/src/FoldDequantize.cpp
llFreetimell: As you said, this code do not remove the original `const_input` :)
This case is like following.
```
[Before]
const_input ---------------------- Dequantize ------- Op1 --- ...
              +-- Op2(not Dequantize) --- ...
              +-- Op3(not Dequantize) --- ...

[After]
dequantized_const_input ---------- Dequantize ------- Op1 --- ...

const_input ----- Op2(not Dequantize) --- ...
              +-- Op3(not Dequantize) --- ...
```

compiler/luci/pass/src/FoldDequantize.cpp
llFreetimell: This case is like following diagram.
```
[Before]
const_input ---------------------- Gather ------- Dequantize --- Op1 ...
              +-- Op2(not Dequantize) --- ...
              +-- Op3(not Dequantize) --- ...

[After]
dequantized_const_input ---------- Gather ------- Op1 --- ...

const_input ----- Op2(not Dequantize) --- ...
              +-- Op3(not Dequantize) --- ...
```

For now, only `Gather` operation can be done like above because of [this function](https://github.com/Samsung/ONE/pull/4404/files#diff-8931584a400d08cf537c82b45ee311adR46)

compiler/luci/pass/src/FoldDequantize.cpp
```diff
@@ -0,0 +1,158 @@
...
+  uint32_t dim_size = 1;
+  for (uint32_t i = 0; i < new_const_node->rank(); ++i)
+  {
+    new_const_node->dim(i) = const_node->dim(i);
+    dim_size *= const_node->dim(i).value();
+  }
+  new_const_node->size<loco::DataType::FLOAT32>(dim_size);
+  new_const_node->shape_status(luci::ShapeStatus::VALID);
+
+  const int32_t q_dim = const_node->quantparam()->quantized_dimension;
```
jinevening, line 74: quantized_dimension is only used for channel-wise quantization. Do you assume `const_node` is quantized in channel-wise?

compiler/luci/pass/src/FoldDequantize.cpp
```diff
@@ -0,0 +1,158 @@
...
+  {
+    // TODO Support more ops
+    return false;
+  }
+}
+
+luci::CircleConst *dequantized_const_node(luci::CircleConst *const_node)
+{
+  auto g = const_node->graph();
+  auto new_const_node = g->nodes()->create<luci::CircleConst>();
```
jinevening, line 61: I think it would be simpler if you just overwrite values in the existing node instead of creating a new node.

compiler/luci/pass/src/FoldDequantize.cpp
jinevening: Oops, I misunderstood the code. Sorry for confusion.

compiler/luci/pass/src/FoldDequantize.cpp
jinevening: Thanks for the explanation. I misunderstood the code :)

compiler/luci/pass/src/FoldDequantize.cpp
llFreetimell: I didn't know that `quantized_dimension` is only used for channel-wise quantization but the code operates whatever the kind of quantization is:-)
When quantization is not channel-wise, quantized_dimension would have only one scale and one zero point.
In that case, because of Line 86 or 99, all the element becomes to use the only one scale and zero point.

compiler/luci/pass/src/FoldDequantize.cpp
llFreetimell: @jinevening I forgot to insert comment about the optimization :) I will add this, thanks :)

compiler/luci/pass/src/FoldDequantize.cpp
jinevening: > When quantization is not channel-wise, quantized_dimension would have only one scale and one zero point.
> If this is incorrect, I should fix the code :(

That's correct.

compiler/luci/pass/src/FoldDequantize.cpp
```diff
@@ -0,0 +1,188 @@
...
+    else if (auto const_node = dynamic_cast<luci::CircleConst *>(node))
+    {
+      if (const_node->quantparam() != nullptr && is_foldable_const(const_node))
+      {
+        for (auto s : loco::succs(const_node))
+        {
+          if (!is_hybrid_supported_op(s))
+          {
+            // Input can be pre-dequantized
+            changed |= replace_const_node(s, dequantized_const_node(const_node));
```
jinevening, line 178: This code may create a new const node even though it is not used (for all cases where `s` is not gather).

Would it be ok?
(I'm not very familiar with how loco handles nodes. It seems that a node is created and registered to `ObjectPool`. The node may not be automatically released although it is not used)

llFreetimell: If newly created const nodes are not used, it means that they are dead nodes.
Thus following optimization will remove all of them :)

https://github.com/Samsung/ONE/blob/4b1c4721248887d68a18f97d6903e459675163fc/compiler/luci/pass/src/CircleOptimizer.cpp#L144

compiler/luci/pass/src/FoldDequantize.cpp
```diff
@@ -0,0 +1,188 @@
...
+        changed = true;
+      }
+    }
+    else if (auto const_node = dynamic_cast<luci::CircleConst *>(node))
+    {
+      if (const_node->quantparam() != nullptr && is_foldable_const(const_node))
+      {
+        for (auto s : loco::succs(const_node))
+        {
+          if (!is_hybrid_supported_op(s))
```
jinevening, line 175: It seems that this code dequantizes const_node even though there is no dequantize Op.

I think successor of `s` must be dequantize Op.

compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include "luci/CircleOptimizer.h"
 
+#include "luci/Pass/FoldDequantize.h"
```
jinevening, line 19: (optional) It would be better to follow the naming convention. :) `FoldDequantizePass.h`

llFreetimell: Oops It's totally my mistake, thanks!

compiler/luci/pass/src/FoldDequantize.cpp
llFreetimell: If const_node is quantized but there is no `Dequantize` operation, I thought that it is just like `FullyConnected` which input can be quantized tensor but output is still FLOAT32.

But in theory, unintended behavior can be occurred... I will fix it!

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include "luci/Pass/FoldDequantizePass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+#include <loco/Service/TypeInference.h>
+
+namespace
+{
+
+bool is_hybrid_supported_op(loco::Node *node)
```
mhs4670go, line 26: Actually, this name makes it hard to know about what this function does. Rename `hybrid_supported`.. to `hybrid_sth_supported` would be help:)

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -0,0 +1,190 @@
...
+      if (const_node->quantparam() != nullptr && is_foldable_const(const_node))
+      {
+        for (auto s : loco::succs(const_node))
+        {
+          auto s_user = loco::succs(s);
+          if (!is_hybrid_kernel_supported(s) && s_user.size() == 1 &&
+              dynamic_cast<luci::CircleDequantize *>(*s_user.begin()) != nullptr)
+          {
+            // Input can be pre-dequantized
+            changed |= replace_const_node(s, dequantized_const_node(const_node));
```
mhs4670go, line 180: It doesn't need to change the dtype of `s`? I mean change from `U8` or `S8` to `F32`.

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -0,0 +1,190 @@
...
+    return true;
+
+  return false;
+}
+
+bool replace_const_node(loco::Node *node, luci::CircleConst *const_node)
+{
+  if (auto gather = dynamic_cast<luci::CircleGather *>(node))
+  {
+    gather->params(const_node);
```
jinevening, line 48: gather's dtype should be changed to the type of `const_node`.

(optional) It would be good to add an assertion to check if dtype of const_node is float32.

compiler/luci/pass/src/FoldDequantizePass.cpp
jinevening, line 180: This leaves dequantize Op after replacing the const_node. It would be better to call `loco::replace(s_user).with(s)`.

llFreetimell: Dequantize is changing U8/S8 to F32... what do you mean @.@?

llFreetimell: I was confused with other pattern, thanks :)

compiler/luci/pass/src/FoldDequantizePass.cpp
jinevening: This is the same comment with @mhs4670go 's comment (https://github.com/Samsung/ONE/pull/4404#discussion_r494710604).

compiler/luci/pass/src/FoldDequantizePass.cpp
mhs4670go: `s` has got U8/S8 dtype first. After `replace_const_node` function is called, `s` has F32 params. It is internally applied somewhere?

llFreetimell: Ah, I understood it now. If `s` is replaced, it means that `FoldDequantize` cause some changes.
Then `TypeInfrencePass` will be done at `CircleOptimizer`.
In that phase, the inferred type will be changed :)

jinevening: > In that phase, the inferred type will be changed :)

Why not change the dtype of CircleNode as well? Is there a reason to keep the previous dtype of `s`?

Mismatch between the inferred type and the dtype of CircleNode seems risky. For example, let's assume another pass needs to use the dtype of `s`. Which type the pass should use? inferred type (fp32) or dtype saved in CircleNode (u8 or s8) ?

llFreetimell: > Mismatch between the inferred type and the dtype of CircleNode seems risky

Yes, it is very risky. It should be changed in view of long term.

> Which type the pass should use? inferred type (fp32) or dtype saved in CircleNode (u8 or s8) ?

**Inferred type**. During optimizations or transformations, new nodes can be generated. But if we decide to use saved dtype in CircleNode, we should always inference the shape and dtype for them. Furthermore, `ShapeInferencePass` and `TypeInferencePass` are implemented to do exactly that thing.

jinevening: > It will be done at TypeInferencePass.

I can't find the code that changes the dtype of CircleNode in TypeInferencePass. Can you tell me the location of the code?

Even though TypeInferencePass changes the dtype, it is safe to change the dtype here, because other passes can access the dtype of `s` before TypeInferencePass is called. This may cause a bug very difficult to find.

> Inferred type. During optimizations or transformations, new nodes can be generated. But if we decide to use saved dtype in CircleNode, we should always inference the shape and dtype for them. Furthermore, ShapeInferencePass and TypeInferencePass are implemented to do exactly that thing.

I didn't know that. There is only one pass that uses `dtype_get` (ResolveCustomOpMatMulPass). Most passes use dtype() in CircleNode. As I said above, it would be safe to set the correct dtype inside an optimization pass to avoid the type-mismatch bugs.

llFreetimell: > I can't find the code that changes the dtype of CircleNode in TypeInferencePass. Can you tell me the location of the code?

My words causes misunderstanding, sorry. `TypeInferencePass` do not change the dtype of CircleNode but change loco dtype.

> Even though TypeInferencePass changes the dtype, it is safe to change the dtype here, because other passes can access the dtype of s before TypeInferencePass is called. This may cause a bug very difficult to find.

Hmm... I agree with your concern but do not agree with the solution.
Until now, fortunately, those errors are not found but it can be occurred for later. So we should resolve this issue.
However, updating shape and dtype in every pass seems not good because

- Inferencing shape and dtype in each pass requires implementation of shape and dtype inference code. It can be crashed with the logic in `CircleTypeInferenceRule.cpp` and `CircleShapeInferenceRule.cpp`. In short, if some of shape is wrong, we don't know where is wrong and if some rule of shape/dtype inference is changed, we should change every pass related with it.
- If some pass generate many nodes, for example 10 nodes, implementation for all of them is really tough.

So I suggest to run `ShapeInferencePass` and `TypeInferencePass` after each pass ended at `CircleOptimizer`.
If then, when each passes ended, all of nodes will have updated shape and dtype.

> Most passes use dtype() in CircleNode. 

Yes, `FuseBCQPass` which I wrote do not use `loco::shape_get` and `loco::dtype_get` neither :(
I think that we should fix those things in every pass.

jinevening: > However, updating shape and dtype in every pass seems not good because

It seems that there was a misunderstanding. What I suggested was just adding one line of code in this pass.

`s->dtype(const_node->dtype());`

Adding one more line for assertion would be better.

`assert(const_node->dtype() == loco::DataType::Float32);`

> So I suggest to run ShapeInferencePass and TypeInferencePass after each pass ended at CircleOptimizer.
If then, when each passes ended, all of nodes will have updated shape and dtype.

This is not a complete solution, because type/shape are not updated until each pass ends. For example, let's assume a pass changes a type of a node at the beginning. The pass can access the type of the node again (before exiting). In this case, TypeInferencePass is not called yet (because the pass is not finished), so the type is not updated.

I think we had better use dtype instead of dtype_get inside a pass. And, dtype of CircleNode should be correctly updated whenever changes occur.

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -0,0 +1,190 @@
...
+      {
+        // Input of Dequantize is constant
+        loco::replace(circle_dequant).with(dequantized_const_node(const_input));
+        changed = true;
+      }
+      else if (loco::dtype_get(input_node) == loco::DataType::FLOAT32)
+      {
+        // Input is already dequantized
+        loco::replace(circle_dequant).with(input_node);
+        changed = true;
```
mhs4670go, line 166: When does this happen? I mean, the node that has F32 dtype just before Dequantize node.

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -0,0 +1,190 @@
...
+  bool changed = false;
+
+  for (auto node : loco::all_nodes(g))
+  {
+    if (auto circle_dequant = dynamic_cast<luci::CircleDequantize *>(node))
+    {
+      auto input_node = circle_dequant->input();
+      if (auto const_input = dynamic_cast<luci::CircleConst *>(input_node))
+      {
+        // Input of Dequantize is constant
```
mhs4670go, line 158: You should check whether it is U8/S8 or not.

compiler/luci/pass/src/FoldDequantizePass.cpp
llFreetimell: Result of offline discussion is as following
- For now, insert dtype setting code
- However, the fundamental issues should be handled in the near future.
    - Risk of redundant shape/dtype information among CircleNode and locoNode
    - shape/dtype validity before and after each pass
    - Recommendation method for getting shape/dtype in each pass

compiler/luci/pass/src/FoldDequantizePass.cpp
llFreetimell: Hmm... in theory, this case can occur but it seems that we don't need to handle this for now.
I will remove this case.

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -0,0 +1,198 @@
...
+  uint32_t dim_size = 1;
+  for (uint32_t i = 0; i < new_const_node->rank(); ++i)
+  {
+    new_const_node->dim(i) = const_node->dim(i);
+    dim_size *= const_node->dim(i).value();
+  }
+  new_const_node->size<loco::DataType::FLOAT32>(dim_size);
+  new_const_node->shape_status(luci::ShapeStatus::VALID);
+
+  const int32_t q_dim = const_node->quantparam()->quantized_dimension;
```
seanshpark, line 60: Please add `assert(const_node->quantparam() != nullptr)` at top of this method

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -0,0 +1,198 @@
...
+  auto g = const_node->graph();
+  auto new_const_node = g->nodes()->create<luci::CircleConst>();
+
+  new_const_node->dtype(loco::DataType::FLOAT32);
+  new_const_node->rank(const_node->rank());
+  uint32_t dim_size = 1;
+  for (uint32_t i = 0; i < new_const_node->rank(); ++i)
+  {
+    new_const_node->dim(i) = const_node->dim(i);
+    dim_size *= const_node->dim(i).value();
```
seanshpark, line 55: Q) for rare unknown shaped constants, one of the dim value can be 0. is there no problem for this case?

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -0,0 +1,198 @@
...
+{
+
+/**
+ *
+ * Folding pattern 1 - When input of Dequantize is foldable constant
+ *
+ * [Before]
+ *     quantized_const_input ----- Dequantize ----- Op1 ---
+ *                             +-- Op2 ---
+ *                             +-- Op3 ---
```
seanshpark, line 125: `Op2` and `Op3` seems not aligned

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -0,0 +1,198 @@
...
+ *     quantized_const_input ----- Op2 ---
+ *                             +-- Op3 ---
+ *
+ *
+ * Folding pattern 2 - When input of Dequantize uses quantized output value
+ *
+ * [Before]
+ *     quantized_const_input ----- Gather ------ Dequantize --- Op1 ---
+ *                             +-- Op2 ---
+ *                             +-- Op3 ---
```
seanshpark, line 139: ditto

compiler/luci/pass/src/FoldDequantizePass.cpp
llFreetimell: Hmm... I am okay to add `assert` but can you explain the difference with https://github.com/Samsung/ONE/pull/4434#discussion_r495626783?

FYI, `const_node->quantparam()` is being checked at https://github.com/Samsung/ONE/pull/4404/files#diff-378da9ae08b1d64c34d114cfebd0da6cR169 :)

compiler/luci/pass/src/FoldDequantizePass.cpp
llFreetimell: Yes, if original tensor has no shape and thus dim value is 0, `new_const_node` would have same zero dimension.
As a result, any calculation code for dequantizing will not be executed because `const_node` size is 0 :)
The only difference with original const node is it's dtype, which changes from `U8/S8` to `F32`.

compiler/luci/pass/src/FoldDequantizePass.cpp
llFreetimell: I couldn't catch the point of comment :(
The name of `Op2` and `Op3` is weird or position is strange? Or any other thing?

compiler/luci/pass/src/FoldDequantizePass.cpp
seanshpark: > I am okay to add assert but can you explain the difference with #4434 (comment)?

For this method only, there should be some condition for `const_node->quantparam()` is not null.
You can also return if this condition is not met.

For `difference with #4434 (comment)?`: 
1) if the assert in #4434 can happen in any model then Pass may assert in debug that may break the execution; it would be better to just skip for that node. For release version, the result could be unpredeictable.
2) if this doesn't happen in real models but you want to be sure for debugging; it is like debug condition, then it is OK. But the code is like `if {condition) ... else assert()` which doesn't seem to be for this case. For the logic, I think it would be better with something like `if (condition) ... else { LOG(...); continue }` or `assert(condition);` with some comment `This conditon must always be true`

compiler/luci/pass/src/FoldDequantizePass.cpp
seanshpark:
```
 *     quantized_const_input ----- Dequantize ----- Op1 ---
 *                                              +-- Op2 ---
 *                                              +-- Op3 ---
```

compiler/luci/pass/src/FoldDequantizePass.cpp
llFreetimell: Thanks for explanation, I will add `assert` here :)

compiler/luci/pass/src/FoldDequantizePass.cpp
llFreetimell: Ah, original one is intended. If `quantized_const_input` is used for some operations, only input of `Dequantize` will be dequantized and the other operations will not be dequantized.
But the names of example may cause confusion :( I will fix it.

seanshpark: OK, the meaning is like nodes that share `quantized_const_input ` as input :)

jinevening (dismissed):
LGTM

seanshpark (approved):
LGTM

llFreetimell:
@jinevening @seanshpark 
I am sorry, updated commit was not uploaded :( PTAL...

And can we merge without @mhs4670go's approval?
He said that we can merge without his approval before he went vacation :)

llFreetimell:
@nnfw-bot test copyright-checker

seanshpark (approved):
LGTM

jinevening (approved):

PR #4348, opened 2020-09-21
[draft][onert] Divide OperationValidator

hseok-oh:
Divide OperationValidator -> OperationValidator + ShapeValidator 
- ShapeValidator: don't move
- OperationValidator: first step for compile

It is draft (test for some operation only)

Step 1: Rename OperationValidator -> ShapeValidator
Step 2: Introduce OperationValidator (check type / check param independent with shape inference)
(Step 3: Merge ShapeValidator into ShapeInference)

Signed-off-by: Hyeongseok Oh <hseok82.oh@samsung.com>

---

Related issue: #4171

PR #4406, opened 2020-09-24
[onert] Introduce OperationValidator

hseok-oh:
Introduce OperationValidator to check shape independent feature
Move shape independent check for ElementwiseActivation
Add negative test for LeakyRelu

---

Part of #4348

### DIFF
--- a/runtime/onert/core/src/compiler/Compiler.cc
+++ b/runtime/onert/core/src/compiler/Compiler.cc
@@ -18,6 +18,7 @@
 
 #include "ParamChecker.h"
 #include "ExecutorFactory.h"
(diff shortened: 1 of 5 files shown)


hyunsik-yoon (approved):
LGTM

wateret (approved):
LGTM, but I still see some code that should be moved to `OperationValidator` from `ShapeValidator`.

hseok-oh:
>I still see some code that should be moved to OperationValidator from ShapeValidator. For example, WHILE node just checks the number of input/outputs, not their shapes.

Right. So I remained TODO comment.

PR #4411, opened 2020-09-24
[nnfw_api] Introduce FullyConnected16x1 sparsity testcase

glistening:
It introduces SparsityParams and its child structures in CircleGen also.

### DIFF
--- a/tests/nnfw_api/src/CircleGen.cc
+++ b/tests/nnfw_api/src/CircleGen.cc
@@ -38,8 +38,15 @@ uint32_t CircleGen::addBuffer(const uint8_t *buf, size_t size)
 
 uint32_t CircleGen::addTensor(const TensorParams &params)
 {
-  int ind = curSubgCtx().tensors.size();
+  uint32_t ind = curSubgCtx().tensors.size();
   curSubgCtx().tensors.emplace_back(buildTensor(params));
+  return ind;
+}
+
+uint32_t CircleGen::addTensor(const TensorParams &params, const SparsityParams &sp)
+{
+  uint32_t ind = curSubgCtx().tensors.size();
+  curSubgCtx().tensors.emplace_back(buildTensor(params, sp));
   return ind;
 }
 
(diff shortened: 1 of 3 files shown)


tests/nnfw_api/src/CircleGen.cc
```diff
@@ -43,6 +43,13 @@ uint32_t CircleGen::addTensor(const TensorParams &params)
   return ind;
 }
 
+uint32_t CircleGen::addTensor(const TensorParams &params, const SparsityParams &sp)
+{
+  int ind = curSubgCtx().tensors.size();
```
mhs4670go, line 48: (optional) 
```suggestion
  uint32_t ind = curSubgCtx().tensors.size();
```

tests/nnfw_api/src/CircleGen.cc
```diff
@@ -298,6 +305,44 @@ flatbuffers::Offset<circle::Tensor> CircleGen::buildTensor(const TensorParams &p
...
+  flatbuffers::Offset<flatbuffers::Vector<flatbuffers::Offset<circle::DimensionMetadata>>>
+      dim_metadata;
+
+  traversal_order = _fbb.CreateVector(sp.traversal_order);
+  block_map = _fbb.CreateVector(sp.block_map);
+
+  std::vector<flatbuffers::Offset<circle::DimensionMetadata>> dim_metadata_vec;
+  for (auto &it : sp.dim_metadata)
+  {
+
```
mhs4670go, line 322: unintended space?

tests/nnfw_api/src/CircleGen.h
```diff
@@ -54,6 +54,47 @@ class CircleGen
...
 
+  using SparseIndexVectorType = circle::SparseIndexVector;
+  using SparseDimensionType = circle::DimensionType;
+
+  struct SparseIndexVector
+  {
+    SparseIndexVector(){};
+    ~SparseIndexVector(){};
+    std::vector<uint16_t> u16;
+  };
```
mhs4670go, line 65:
```suggestion
  struct SparseIndexVector
  {
    SparseIndexVector(){}
    ~SparseIndexVector(){}
    std::vector<uint16_t> u16;
  };
```
And maybe ctor and ctor could be removed as well. Is there any reason of this?

tests/nnfw_api/src/CircleGen.h
```diff
@@ -54,6 +54,47 @@ class CircleGen
...
+  {
+    DimMetaData() = delete;
+    DimMetaData(SparseDimensionType format, std::vector<uint16_t> array_segments,
+                std::vector<uint16_t> array_indices)
+        : _format{format},
+          _array_segments_type(SparseIndexVectorType::SparseIndexVector_Uint16Vector),
+          _array_indices_type(SparseIndexVectorType::SparseIndexVector_Uint16Vector)
+    {
+      _array_segments.u16 = array_segments;
+      _array_indices.u16 = array_indices;
```
mhs4670go, line 77:
```suggestion
        : _format{format},
          _array_segments_type(SparseIndexVectorType::SparseIndexVector_Uint16Vector),
          _array_indices_type(SparseIndexVectorType::SparseIndexVector_Uint16Vector),
          __array_segments{array_segments}, _array_indices{array_indices}
    {
```

mhs4670go: And, there are three kind of types in `SparseIndexVector` - INT32, U16, U8. Maybe you will just support only U16 type for convenience, won't you?

wateret (dismissed):
LGTM

tests/nnfw_api/src/CircleGen.h
glistening: @mhs4670go Yes, I will update if it become necessary. Just adding a field and function will work. However, it is not necessary and I don't want to spend my time. I guess it is unlikely to have a model that requires `Int32` index or segments, which means its sparse dimension has > 65535 elements. We may have `uint8`, when sparse dimension max index <= 256 or num_nonzero <= 256.

tests/nnfw_api/src/CircleGen.cc
glistening: @mhs4670go Thank you. I've updated. (including `addTensor` with 1 parameter version ­­— which the code is originated.)

mhs4670go (approved):
LGTM

glistening:
@wateret @mhs4670go's comments are applied. PTAL.

wateret (approved):
LGTM

PR #4419, opened 2020-09-24
[onert] (trivial) Use 'buf' postfix on inferShape functions' parameters using buffer

dayo09:
- put _buf into parameter's name to clarify that it's buffer from tensor
- change _data to _buf (for consistency)

### DIFF
--- a/runtime/onert/core/include/util/ShapeInference.h
+++ b/runtime/onert/core/include/util/ShapeInference.h
@@ -50,7 +50,7 @@ ir::Shape inferBatchMatMulShape(const ir::Shape &lhs_shape, const ir::Shape &rhs
 ir::Shape inferBCQFullyConnectedShape(const ir::Shape &in_shape, const ir::Shape &cluster_shape,
                                       const int32_t *cluster_buf);
 
-ir::Shape inferBroadcastToShape(const ir::Shape wshape, const int32_t *shape_buffer);
+ir::Shape inferBroadcastToShape(const ir::Shape shp_shape, const int32_t *shp_buf);
 
 ir::Shape inferConcatShape(const Shapes &in_shapes, const ir::operation::Concat::Param &param);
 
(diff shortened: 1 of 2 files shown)


runtime/onert/core/include/util/ShapeInference.h
```diff
@@ -47,7 +47,7 @@ ir::Shape inferArgMaxShape(const ir::Shape &input_shape, int axis, int rank);
 ir::Shape inferBatchMatMulShape(const ir::Shape &lhs_shape, const ir::Shape &rhs_shape,
                                 const ir::operation::BatchMatMul::Param &param);
 
-ir::Shape inferBroadcastToShape(const ir::Shape wshape, const int32_t *shape_buffer);
+ir::Shape inferBroadcastToShape(const ir::Shape shp_shape, const int32_t *shape_buf);
```
hyunsik-yoon, line 50: I am not sure of the meaning of `shp_shape`.
how about something like `src_shape` and `dest_shape_buf`?

dayo09: @hyunsik-yoon 
That's the confusing part.

shp_shape means that  `SHAPE` tensor's shape.
The input tensor's name is 'SHAPE', so shp_shape means 'shape of the `SHAPE` tensor' and shape_buf means 'buffer of the `SHAPE` tensor'. 

I followed the other function's cases.
I think src_shape and dest_shape_buf doesn't fit the underlying meaning too. it's no source and dest, it's actually shape and buffer of the shape tensor..
May I add some explanation then?

hyunsik-yoon: I see. Let's just leave as you named. :-)

hyunsik-yoon:
please resolve conflict. :-)

hyunsik-yoon (approved):
LGTM

PR #4338, opened 2020-09-17
[Draft] Revise BCQ fusing algorithm

llFreetimell:
Parent Issue : #4337 

This draft is for revising `FuseBCQPass` algorithm.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

Issue #4337, opened 2020-09-17
[Compiler] Refactoring BCQ for removing hard coding


llFreetimell:
### Background

Until now, `FuseBCQPass` has some hard coding algorithm.
Specifically, BCQ information node and BCQ applicable operation is connected with their names.
The names become different during conversion(pb->tflite) and thus all the naming rules must be recognized in advance to apply BCQ. This was hard coding algorithm but works well for now.
However, from tf-nightly 2.4, names in subgraphs become totally different and as a result this hard coding algorithm will not work for sure.
Therefore we should find another solution to connect BCQ information node and BCQ applicable operation.

One more thing, some optimizations during conversion sometimes make some loss of BCQ information.
This was handled by additional python script but we cannot sure another optimization do not disturb applying BCQ.
So this should be handled together.

### What To Do

One of immutable characteristic between TensorFlow and TFLite is the sequence of output.
Even TFLite converter transform and optimize a model, value and sequence of output must be same.
So one of solution originated from this idea.
Briefly, we will insert metadata of BCQ as first output and then package BCQ information nodes and BCQ applicable operation as continuous sequence next to original outputs like following example.
```
[Original output]
out1, out2

[BCQ information and metadata inserted]
METADATA, out1, out2, node1/bcqinfoA, node1/bcqinfoB, node1, ..., nodeN/bcqinfoA, nodeN/bcqinfoB, nodeN
```
And after BCQ is fused, model will be reverted to have only original outputs.
For more details, please refer to comment of each code.

### Todo List
- [x] Introduce new bcq-tool, `generate_bcq_metadata` #4420
- [x] Revise `generate_bcq_output_arrays` #4422 #4452 #4461 #4493 #4506 #4510
- [x] Update validation of circle graph not to check shape of `CircleOutputExclude` #4357
- [x] Refactoring `FuseBCQPass` #4434 #4457 #4462 #4511 #4523
- [x] Update `tf2nnpkg` #4496
- [x] Update `one-import-bcq` #4495
- [x] Remove `preserve_bcq_info` #4520 #4524
- [x] Final Check
- (Good to have) New optimization passes in `luci` to handle not optimized operations (#4353)

### Draft

Draft code : #4338

- Most of comments in code are not written yet. Those will be written soon.
- If there is some difference between above description and draft codes, the description is correct. Differences will be removed by additional implementations.

llFreetimell:
Done!

PR #4420, opened 2020-09-24
[bcq-tools] Introduce metadata generator

llFreetimell:
Parent Issue : #4337
Full Draft : #4338

This commit will introduce new metadata generator, `generate_bcq_metadata`

### DIFF
--- a/compiler/bcq-tools/README.md
+++ b/compiler/bcq-tools/README.md
@@ -76,3 +76,38 @@ const(value=[1, 2, 3, -3], name='const3')
 ### Caution
 
 - Generated output_arrays will be start with comma.
+
+## generate_bcq_metadata
+
+### Purpose
+
+To apply BCQ, additional information which is not included in original model is needed.
+For example, without metadata, we cannot know from where to where BCQ related nodes are included. Metadata for BCQ is needed due to those reason and `generate_bcq_metadata` will generate metadata in the model according to each BCQ version.
+
+### How to use
+
+```bash
+generate_bcq_metadata \
+--input_path /path/to/original_model.pb \
+--output_path /path/to/metadata_inserted_model.pb
+--output_arrays output1,output2,...,outputN
+```
+
+### How it works
+
+First of all, metadata will be generated as following description.
+```
+< Generated Metadata in BCQ version 1 >
+[0] Starting magic number                = {-2e9 + 27}
+[1] Version of BCQ                       = {1}
+[2] The number of original model outputs = {N | N > 0}
+[3] Bundle size                          = {7, 8}
+[4] Ending magic number                  = {2e9 - 27}
+```
+- BCQ version 1
+    - Two magic numbers, starting and ending magic number, are used for indicating that the node includes BCQ metadata. To decrease value duplication probability, prime number is used and the value is inserted not only at the beginning but also at the end.
+    - The word **bundle** means that a set of BCQ information and BCQ applicable operation. If six BCQ information nodes are used for one operation, the six information nodes and the other one operation are packaged as **bundle**. Then, in this case, the bundle size will be 6 + 1 = 7.
+
+### Caution
+
+- If there is no BCQ information in original model, any changes will be applied.
--- a/compiler/bcq-tools/README.md
+++ b/compiler/bcq-tools/README.md
@@ -95,7 +95,7 @@ generate_bcq_metadata \
 
 ### How it works
 
-First of all, metadata will be generated as following description.
+Metadata will be generated as following description.
 ```
 < Generated Metadata in BCQ version 1 >
 [0] Starting magic number                = {-2e9 + 27}
@@ -105,7 +105,7 @@ First of all, metadata will be generated as following description.
 [4] Ending magic number                  = {2e9 - 27}
 ```
 - BCQ version 1
-    - Two magic numbers, starting and ending magic number, are used for indicating that the node includes BCQ metadata. To decrease value duplication probability, prime number is used and the value is inserted not only at the beginning but also at the end.
+    - Two magic numbers, starting and ending magic number, are used for indicating that the model includes BCQ metadata. To decrease value duplication probability, prime number is used and the value is inserted not only at the beginning but also at the end.
     - The word **bundle** means that a set of BCQ information and BCQ applicable operation. If six BCQ information nodes are used for one operation, the six information nodes and the other one operation are packaged as **bundle**. Then, in this case, the bundle size will be 6 + 1 = 7.
 
 ### Caution
--- a/compiler/bcq-tools/README.md
+++ b/compiler/bcq-tools/README.md
@@ -81,8 +81,8 @@ const(value=[1, 2, 3, -3], name='const3')
 
 ### Purpose
 
-To apply BCQ, additional information which is not included in original model is needed.
-For example, without metadata, we cannot know from where to where BCQ related nodes are included. Metadata for BCQ is needed due to those reason and `generate_bcq_metadata` will generate metadata in the model according to each BCQ version.
+`generate_bcq_metadata` is for appending metadata as output of a model which includes BCQ information.
+The appended metadata is used for connecting BCQ related operations and constant nodes.
 
 ### How to use
 
--- a/compiler/bcq-tools/generate_bcq_metadata
+++ b/compiler/bcq-tools/generate_bcq_metadata
@@ -167,7 +167,6 @@ def determine_bcq_version(flags):
 
     When BCQ information does not exist, -1 is returned.
     """
-
     bcq_version = -1
 
     original_graph = load_graph(flags.input_path)
@@ -202,7 +201,7 @@ def generate_bcq_metadata(flags):
         err_msg = "BCQ version of the model(v{}) ".format(model_version)
         err_msg += "is higher than "
         err_msg += "the version supported by this program(v{})".format(program_version)
-        raise Exception(err_msg)
+        raise SystemExit(err_msg)
 
     tf.io.write_graph(result_graph_def, '.', flags.output_path, False)
 
--- a/compiler/bcq-tools/generate_bcq_metadata
+++ b/compiler/bcq-tools/generate_bcq_metadata
@@ -192,11 +192,13 @@ def generate_bcq_metadata(flags):
     """
     version = determine_bcq_version(flags)
 
-    # When there is not BCQ information, do nothing
     if version == 1:
         result_graph_def = generate_bcq_metadata_v1(flags)
-    else:
+    elif version == -1:
+        # When there is no BCQ information, do nothing
         result_graph_def = load_graph(flags.input_path)
+    else:
+        raise Exception("BCQ version is not supported yet")
 
     tf.io.write_graph(result_graph_def, '.', flags.output_path, False)
 
--- a/compiler/bcq-tools/generate_bcq_metadata
+++ b/compiler/bcq-tools/generate_bcq_metadata
@@ -190,15 +190,19 @@ def generate_bcq_metadata(flags):
         - Metadata based on each BCQ format
         - Magic number indicating end
     """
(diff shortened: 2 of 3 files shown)


compiler/bcq-tools/generate_bcq_metadata
```diff
@@ -0,0 +1,214 @@
...
+    new_node.name = "one_compiler/bcqinfo_one_metadata"
+    new_node.attr["dtype"].CopyFrom(
+        tf.core.framework.attr_value_pb2.AttrValue(type=tf.int32.as_datatype_enum))
+    new_node.attr["value"].tensor.CopyFrom(metadata_tensor)
+    return new_graph_def
+
+
+def determine_bcq_version(flags):
+    """
+    CAUTION : For now, BCQ has only version and thus always returns 1 when BCQ
```
seanshpark, line 164: `BCQ has only version` -> `BCQ has only one version` ?

compiler/bcq-tools/README.md
```diff
@@ -76,3 +76,38 @@ const(value=[1, 2, 3, -3], name='const3')
...
+```bash
+generate_bcq_metadata \
+--input_path /path/to/original_model.pb \
+--output_path /path/to/metadata_inserted_model.pb
+--output_arrays output1,output2,...,outputN
+```
+
+### How it works
+
+First of all, metadata will be generated as following description.
```
seanshpark, line 98: `First of all` can be omitted.

compiler/bcq-tools/README.md
```diff
@@ -76,3 +76,38 @@ const(value=[1, 2, 3, -3], name='const3')
...
+```
+< Generated Metadata in BCQ version 1 >
+[0] Starting magic number                = {-2e9 + 27}
+[1] Version of BCQ                       = {1}
+[2] The number of original model outputs = {N | N > 0}
+[3] Bundle size                          = {7, 8}
+[4] Ending magic number                  = {2e9 - 27}
+```
+- BCQ version 1
+    - Two magic numbers, starting and ending magic number, are used for indicating that the node includes BCQ metadata. To decrease value duplication probability, prime number is used and the value is inserted not only at the beginning but also at the end.
```
seanshpark, line 108: `the node includes` : I'm don't quite understand what the `node` is.

compiler/bcq-tools/README.md
```diff
@@ -76,3 +76,38 @@ const(value=[1, 2, 3, -3], name='const3')
 ### Caution
 
 - Generated output_arrays will be start with comma.
+
+## generate_bcq_metadata
+
+### Purpose
+
+To apply BCQ, additional information which is not included in original model is needed.
```
seanshpark, line 84: Somewhat vague sentence.

compiler/bcq-tools/generate_bcq_metadata
```diff
@@ -0,0 +1,214 @@
...
+        help="Full filepath of the input file.",
+        required=True)
+    parser.add_argument(
+        "-o",
+        "--output_path",
+        type=str,
+        help="Full filepath of the output file.",
+        required=True)
+    parser.add_argument(
+        "-a",
```
seanshpark, line 49: `tf2tfliteV2` uses `-O` for this. Can you use `-O` for consistancy?

compiler/bcq-tools/README.md
llFreetimell: I will change to `the model includes`

compiler/bcq-tools/README.md
llFreetimell: Which point is vague?
If `additional information` is vague, next sentence is saying with an example. However if other thing is vague, could you let me know? :)

seanshpark: Yes, `additional information` is somewhat vague.

compiler/bcq-tools/README.md
```diff
@@ -76,3 +76,38 @@ const(value=[1, 2, 3, -3], name='const3')
...
+For example, without metadata, we cannot know from where to where BCQ related nodes are included. Metadata for BCQ is needed due to those reason and `generate_bcq_metadata` will generate metadata in the model according to each BCQ version.
```
seanshpark, line 85: if `metadata` is addition information, it would be better to add this explanation.

seanshpark: `from where to where`; is this like word to word translation from Korean?
I'm not sure the meaning is unstandable

seanshpark: `For example` : if this is one of several examples, using `for example` seems suitable. If not, I think `For example` can be omitted.

seanshpark: Can this be what your intension is?
> `generate_bcq_metadata` is to append metadata in BCQ model that is used to connect BCQ related operators and constant nodes.

llFreetimell: > if metadata is addition information, it would be better to add this explanation.

OK, I will add it.

> from where to where; is this like word to word translation from Korean? I'm not sure the meaning is unstandable

I searched about this expression and it seems be used in English but... I am not sure :(

> For example : if this is one of several examples, using for example seems suitable. If not, I think For example can be omitted.

Metadata has more information than the example :)

> Can this be what your intension is?
> generate_bcq_metadata is to append metadata in BCQ model that is used to connect BCQ related operators and constant nodes.

Quite simillar :)


I will fix the sentence with applying your opinion!

seanshpark (dismissed):
LGTM

seanshpark:
@jinevening , please review as much as you understand :)

compiler/bcq-tools/generate_bcq_metadata
```diff
@@ -0,0 +1,214 @@
...
+        - Metadata based on each BCQ format
+        - Magic number indicating end
+    """
+    version = determine_bcq_version(flags)
+
+    # When there is not BCQ information, do nothing
+    if version == 1:
+        result_graph_def = generate_bcq_metadata_v1(flags)
+    else:
+        result_graph_def = load_graph(flags.input_path)
```
jinevening, line 199: I think it had better exit the script if `version != 1` rather than make the same model.

compiler/bcq-tools/generate_bcq_metadata
```diff
@@ -0,0 +1,214 @@
...
+        elif num_of_bcqinfo != len(infonames):
+            is_valid = False
+
+    # The number of BCQv1 information should be 6 or 7
+    if num_of_bcqinfo != 6 and num_of_bcqinfo != 7:
+        is_valid = False
+
+    # If BCQ information is invalid, return original model
+    if is_valid == False:
+        return original_graph_def
```
jinevening, line 134: Similar comment. How about printing error message (e.g., "invalid BCQ information") and exiting the script?

compiler/bcq-tools/generate_bcq_metadata
llFreetimell: Because of users of `tf2nnpkg`, we should make same model :(
More specifically, users both who want to apply BCQ and who do not know BCQ, they use same command as following.
```sh
tf2nnpkg --graphdef model.pb --info model.info -o .
```
Because of this situation, when a model has no BCQ information, we should copy original model.

seanshpark: I think this user scenatio needs validation for `version != 1`

- In the future: when there is a BCQ file with version 2 or above and run with this version of script: what should we do?

llFreetimell: > I think this user scenatio needs validation for version != 1

If there is something I misunderstand, please let me know :)

At first, I wrote a code to throw an error when version or something related with BCQ is invalid.
However, I modified all the codes not to do anything after I received a review, "Just do nothing when BCQ information is invalid seems better than throwing an error. When model without BCQ is invalid, it should be died. However, when a model with BCQ is wrong, we can ignore only BCQ and generate a model without BCQ".
Following code is also written according to the policy.
https://github.com/Samsung/ONE/blob/c63a98f465d0c4c9854cbb37ade23bf22eddd2e3/compiler/luci/pass/src/FuseBCQPass.cpp#L508

So, should we change the policy?

---

> In the future: when there is a BCQ file with version 2 or above and run with this version of script: what should we do?

When new BCQ version is introduced, related implementations in `luci`, `bcq-tools` and so on should be done in advance of course.

jinevening: > Because of this situation, when a model has no BCQ information, we should copy original model.

I understand we should copy original model when a model has no BCQ information. But that does not mean that the tool must copy original model when the model has wrong version.

IMHO, a program should exit with error messages when its input is **invalid**. To me, it seems that a model without BCQ information is **valid**, while a model with wrong version is **invalid**.

It's my personal opinion (I'm not familiar with BCQ), so you can ignore it. :)

seanshpark: > When new BCQ version is introduced, related implementations in luci, bcq-tools and so on should be done in advance of course.

Not that case. Please read again.

seanshpark: Case 1: model version == script version: OK
Case 2: model version < script version: will be handled in the future version of script
Case 3: model version > script version: show message("cannot handle higher version") and stop maybe?

llFreetimell: @seanshpark 

> Case 2: model version < script version: will be handled in the future version of script

This is OK either. When script version is 3, it means that the script can handle BCQ version 1,2,3.

> Case 3: model version > script version: show message("cannot handle higher version") and stop maybe?

I have no objection for showing some messages and stopping :)
However as I wrote in https://github.com/Samsung/ONE/pull/4420#discussion_r495758729, I received an opinion, **Do not stop and just ignore BCQ**, and have written related codes according to the policy.
What I am curious is, **should I write a code to stop for such cases from now on** :-) ?

seanshpark: > I received an opinion, Do not stop and just ignore BCQ

Maybe we can ask the person that gave you this opinion :)

llFreetimell: Hmm..... frankly speaking.... it was you O.O;;
Then I will follow to showing message and stop...!

compiler/bcq-tools/generate_bcq_metadata
```diff
@@ -0,0 +1,216 @@
...
+    """
+    version = determine_bcq_version(flags)
+
+    if version == 1:
+        result_graph_def = generate_bcq_metadata_v1(flags)
+    elif version == -1:
+        # When there is no BCQ information, do nothing
+        result_graph_def = load_graph(flags.input_path)
+    else:
+        raise Exception("BCQ version is not supported yet")
```
seanshpark, line 201: One more thing :)
Can you show file version and this version number? Maybe like;
```
BCQ version of the file(number) is higher than the program(number). Please upgrade this program.
```

seanshpark (dismissed):
LGTM thank you!

compiler/bcq-tools/generate_bcq_metadata
```diff
@@ -190,13 +190,19 @@ def generate_bcq_metadata(flags):
...
         result_graph_def = generate_bcq_metadata_v1(flags)
-    else:
+    elif model_version == -1:
+        # When there is no BCQ information, do nothing
         result_graph_def = load_graph(flags.input_path)
+    else:
+        err_msg = "BCQ version of the model(v{}) ".format(model_version)
+        err_msg += "is higher than "
+        err_msg += "the version supported by this program(v{})".format(program_version)
+        raise Exception(err_msg)
```
jinevening, line 205: This will raise unhandled Exception (which is not accepted by our release testing team).

Please change the code as below.

```suggestion
        raise SystemExit(err_msg)
```

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

PR #4421, opened 2020-09-24
[luci-interpreter] Support for FloorDiv op

ai-moiseev:
Add support for FloorDiv op in luci-interpreter.

For #1612

### DIFF
--- a/compiler/luci-interpreter/src/kernels/FloorDiv.cpp
+++ b/compiler/luci-interpreter/src/kernels/FloorDiv.cpp
@@ -19,7 +19,7 @@
 #include "kernels/Utils.h"
 
 #include <tensorflow/lite/kernels/internal/reference/binary_function.h>
-#include <math.h>
+#include <cmath>
 
 namespace luci_interpreter
 {
@@ -55,7 +55,7 @@ void FloorDiv::execute() const
 void FloorDiv::evalFloat() const
 {
   auto FloorDivFunc = [](float x, float y) -> float {
-    return std::floor(std::divides<double>()(static_cast<double>(x), static_cast<double>(y)));
+    return std::floor(static_cast<double>(x) / static_cast<double>(y));
   };
 
   if (x()->shape() != y()->shape())
--- a/compiler/luci-interpreter/src/kernels/FloorDiv.cpp
+++ b/compiler/luci-interpreter/src/kernels/FloorDiv.cpp
@@ -58,19 +58,26 @@ void FloorDiv::evalFloat() const
     return std::floor(static_cast<double>(x) / static_cast<double>(y));
   };
 
+  const auto x_data = getTensorData<float>(x());
+  const auto y_data = getTensorData<float>(y());
+
+  // Check the denominator
+  for (int i = 0; i < getTensorShape(y()).FlatSize(); ++i)
+  {
+    LUCI_INTERPRETER_CHECK(y_data[i] != 0);
+  }
+
   if (x()->shape() != y()->shape())
   {
     tflite::reference_ops::BroadcastBinaryFunction4DSlow<float, float, float>(
-        getTensorShape(x()), getTensorData<float>(x()), getTensorShape(y()),
-        getTensorData<float>(y()), getTensorShape(output()), getTensorData<float>(output()),
-        FloorDivFunc);
+        getTensorShape(x()), x_data, getTensorShape(y()), y_data, getTensorShape(output()),
+        getTensorData<float>(output()), FloorDivFunc);
   }
   else
   {
     tflite::reference_ops::BinaryFunction<float, float, float>(
-        getTensorShape(x()), getTensorData<float>(x()), getTensorShape(y()),
-        getTensorData<float>(y()), getTensorShape(output()), getTensorData<float>(output()),
-        FloorDivFunc);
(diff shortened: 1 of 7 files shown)


compiler/luci-interpreter/src/kernels/FloorDiv.cpp
```diff
@@ -0,0 +1,78 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "kernels/FloorDiv.h"
+#include "kernels/Utils.h"
+
+#include <tensorflow/lite/kernels/internal/reference/binary_function.h>
+#include <math.h>
```
s-barannikov, line 22:
```suggestion
#include <cmath>
```

compiler/luci-interpreter/src/kernels/FloorDiv.cpp
```diff
@@ -0,0 +1,78 @@
...
+      break;
+    default:
+      throw std::runtime_error("Unsupported type.");
+  }
+}
+
+void FloorDiv::evalFloat() const
+{
+  auto FloorDivFunc = [](float x, float y) -> float {
+    return std::floor(std::divides<double>()(static_cast<double>(x), static_cast<double>(y)));
```
s-barannikov, line 58: Why don't just
```suggestion
    return std::floor(static_cast<double>(x) / static_cast<double>(y));
```

compiler/luci-interpreter/src/kernels/FloorDiv.test.cpp
```diff
@@ -0,0 +1,119 @@
...
+namespace kernels
+{
+namespace
+{
+
+using namespace testing;
+
+TEST(FloorDivTest, FloatSimple)
+{
+  std::initializer_list<int32_t> x_shape{2, 3};
```
s-barannikov, line 32: Is there a specific reason not to use `std::vector` for shapes?

compiler/luci-interpreter/src/kernels/FloorDiv.cpp
```diff
@@ -0,0 +1,78 @@
...
+  };
+
```
jinevening, line 60: Please check denominator (`y`) is non-zero before calling TFLite kernels.

It would be even better to add a test for a divide-by-zero case.

compiler/luci-interpreter/src/kernels/FloorDiv.cpp
ai-moiseev: Fixed

compiler/luci-interpreter/src/kernels/FloorDiv.test.cpp
ai-moiseev: It's used here to feed both `makeInputTensor` and `ElementsAreArray` down below without need for conversion. `makeInputTensor` uses to implicitly create `Shape`, but `Shape` doesn't have `begin()`/`end()` to use it in default gmock matcher.

compiler/luci-interpreter/src/kernels/FloorDiv.cpp
ai-moiseev: Isn't division-by-zero a defined behavior for IEEE 754 floats? I've tested it with constant tensors and it gives valid output with `inf`/`-inf` values (or `nan` in case of `0/0`).

I've updated test cases to cover this situations and specific case for NaN. Please correct me if I understood it wrong.

compiler/luci-interpreter/src/kernels/FloorDiv.cpp
ai-moiseev: No reason, probably, tflite implementation was taken as [reference](https://github.com/tensorflow/tensorflow/blob/b36436b087bd8e8701ef51718179037cccdfc26e/tensorflow/lite/kernels/floor_div.cc#L46). Replaced with simple division.

compiler/luci-interpreter/src/kernels/FloorDiv.cpp
jinevening: TF tensors can have inf/-inf/nan (https://www.tensorflow.org/api_docs/python/tf/math/is_inf), so it makes sense to support div-by-zero.

But, I saw the check code from the TFLite kernel. If we allow div-by-zero, luci-interpreter may work differently from TFLite interpreter. I'm not sure this is desirable. @s-barannikov what do you think of this?

ai-moiseev: Haven't noticed this check in tflite, thanks! It seems that tflite doesn't support zero-denominators then, I've added check and `_NEG` testcase for it.

compiler/luci-interpreter/src/kernels/FloorDiv.test.cpp
s-barannikov: Please use `vector`. It is both consistent and avoids confusion -- `initializer_list`s were designed for other purpose. It is quite unusual to see them as local variables.

ai-moiseev: Done. Replaced with `vector` for reference shapes and `Shape`s.

s-barannikov (approved):
LGTM

jinevening (approved):
LGTM

PR #4423, opened 2020-09-24
[luci] Reorder the sequence of optimizations

llFreetimell:
Parent Issue : #4373
Draft : #4374

Sometimes, some optimization passes need inferenced shape or dtype.
However for now, shape inference and type inference are done at last.
This commit will reorder the sequence of optimizations
and thus another optimization passes can use inferenced shape or dtype.

### DIFF
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -109,7 +109,7 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   logo::Phase phase;
 
   /* TRANSFORM DECLARATION BEGIN */
-  
+
   // Shape inference is needed for added nodes doing below transformations
   phase.emplace_back(std::make_unique<luci::ShapeInferencePass>());
   phase.emplace_back(std::make_unique<luci::TypeInferencePass>());
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -109,6 +109,12 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   logo::Phase phase;
 
   /* TRANSFORM DECLARATION BEGIN */
+  
+  // Shape inference is needed for added nodes doing below transformations
+  phase.emplace_back(std::make_unique<luci::ShapeInferencePass>());
+  phase.emplace_back(std::make_unique<luci::TypeInferencePass>());
+  phase.emplace_back(std::make_unique<logo::RemoveDeadNodeWithQueryPass>());
+
   if (_options->query(Options::Algorithm::ResolveCustomOpAdd))
   {
     phase.emplace_back(std::make_unique<luci::ResolveCustomOpAddPass>());
@@ -138,10 +144,6 @@ void CircleOptimizer::optimize(loco::Graph *g) const
     phase.emplace_back(std::make_unique<FuseAddWithTConvPass>());
   }
 
-  // Shape inference is needed for added nodes doing above transformations
-  phase.emplace_back(std::make_unique<luci::ShapeInferencePass>());
-  phase.emplace_back(std::make_unique<luci::TypeInferencePass>());
-  phase.emplace_back(std::make_unique<logo::RemoveDeadNodeWithQueryPass>());
   /* TRANSFORM DECLARATION END */
 
   ProgressReporter prog(g, logo::PhaseStrategy::Saturate);
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -110,10 +110,9 @@ void CircleOptimizer::optimize(loco::Graph *g) const
 
   /* TRANSFORM DECLARATION BEGIN */
 
-  // Shape inference is needed for added nodes doing below transformations
+  // Shape and type inference is needed for doing below transformations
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -109,6 +109,12 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   logo::Phase phase;
 
   /* TRANSFORM DECLARATION BEGIN */
+
+  // Shape inference is needed for added nodes doing below transformations
+  phase.emplace_back(std::make_unique<luci::ShapeInferencePass>());
+  phase.emplace_back(std::make_unique<luci::TypeInferencePass>());
+  phase.emplace_back(std::make_unique<logo::RemoveDeadNodeWithQueryPass>());
```
jinevening, line 116: Some passes create dead nodes (ex: FoldDequantizePass), so this pass should be called after such a pass is run.

llFreetimell: If  other passes create dead nodes, those nodes will be removed after the iteration.
However, considering the efficiency, I will apply your opinion :)

jinevening: > those nodes will be removed after the iteration.

AFAIK, a phase runner repeats to run passes only when a pass changes a graph (returns changed=true). FoldDequantizePass can return (changed = false) even though it creates a dead node.

https://github.com/Samsung/ONE/pull/4404#discussion_r494253979

compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -109,6 +109,11 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   logo::Phase phase;
 
   /* TRANSFORM DECLARATION BEGIN */
+
+  // Shape and type inference is needed for doing below transformations
+  phase.emplace_back(std::make_unique<luci::ShapeInferencePass>());
+  phase.emplace_back(std::make_unique<luci::TypeInferencePass>());
```
jinevening, line 115: Some passes can change the type of nodes. Is it ok not to update the types after those passes run?

jinevening: > Some passes can change the type of nodes.

If so, the phase runner will run the whole passes again, so it will be ok. Never mind this comment :)

compiler/luci/pass/src/CircleOptimizer.cpp
llFreetimell: Oh..... I did another mistake on `FoldDequantizePass`... if any changes occurred in some pass, it should return true but I didn't..... really thanks for that.

jinevening:
>  some optimization passes need inferenced shape or dtype.

Can you explain it a little bit more? Is there a case where shape and dtype cannot be inferred inside an optimization pass ?

llFreetimell:
@jinevening 

In `FoldDequantizePass`, following code includes `loco::dtype_get`
```cpp
    if (auto circle_dequant = dynamic_cast<luci::CircleDequantize *>(node))
    {
      auto input_node = circle_dequant->input();
      if (auto const_input = dynamic_cast<luci::CircleConst *>(input_node))
      {
        // Input of Dequantize is constant
        loco::replace(circle_dequant).with(dequantized_const_node(const_input));
        changed = true;
      }
      else if (loco::dtype_get(input_node) == loco::DataType::FLOAT32)
      {
        // Input is already dequantized
        loco::replace(circle_dequant).with(input_node);
        changed = true;
      }
    }
``` 
The reason why `dtype_get` is used instead of `input_node->dtype()` is that result of `ShapeInference` and `TypeInference` is annotated as `loco::AnnotatedItem`, not by circle attributes.
In short, if there are some changes for shape or dtype during some `luci::Pass`, shape and dtype of circle node is not changed but loco node's annoted item is changed.
So we should use `loco::dtype_get` or `loco::shape_get` to get the updated shape or dtype.
However, if we do not infer shape and dtype at first like now, `loco::dtype_get` and `loco::shape_get` will return an error because they have no annoted item

jinevening:
> The reason why dtype_get is used instead of input_node->dtype() is that result of ShapeInference and TypeInference is annotated as loco::AnnotatedItem, not by circle attributes.

Why do you use the result of TypeInference (instead of input_node->dtype())? Is there a case where the type should be inferred (such as dynamic type?)?

llFreetimell:
> Why do you use the result of TypeInference (instead of input_node->dtype())? Is there a case where the type should be inferred (such as dynamic type?)?

In case of the result of another passes. Following example shows what I want to say.
```
[Before]
Op1(output type : S8) ----- Dequantize --- Op2 ---

[After phase 1]
Op1(output type : FLOAT32) ----- Dequantize --- Op2 ---

[After phase 2]
Op1(output type : FLOAT32) ----- Op2 ---
```

I didn't see such cases but normally, using `loco::dtype_get` is recommended instead of dtype in CircleNode.

seanshpark:
Optimization algorithms run several times until there is nothing changed.
If the first run can't run cause there is no shape or no type, then there will be at the second run.
If your code is not working with this steps, then please explain what the problem is.

llFreetimell:
> Optimization algorithms run several times until there is nothing changed.
If the first run can't run cause there is no shape or no type, then there will be at the second run.
If your code is not working with this steps, then please explain what the problem is.

Then we may use `loco::shape_known` for checking those cases, I understood :)
Close this PR!

PR #4430, opened 2020-09-25
[onert] Support neon optimization for softmax

YiHyunJin:
In softmax, it use neon optimizaion for exp operation.
So I added this, And tested any models. The result is an 1.5% increase in speed.
(1.346 -> 1.327)


glistening:
@YiHyunJin As I guessed, this neon kernel (used in our model app) gives a slightly different value and lead errors like:

> pr-cross-runtime-test-pipeline-release/tests/nnapi/include/TestHarness.h:303: Failure 
> The difference between expected and actual is 1.0877847671508789e-05, which exceeds fpRange, 
> where expected evaluates to 0.086128540337085724, 
> actual evaluates to 0.086139418184757233, 

This kernel should be enclosed with build flag, that is not eabled by default, which is enabled only in the case that this difference is acceptable like our model.

compute/cker/include/cker/operation/Exp.h
```diff
@@ -27,6 +30,46 @@ namespace nnfw
 namespace cker
 {
 
+#ifdef USE_NEON
```
glistening, line 33:
```suggestion
#if defined(USE_FAST_SOFTMAX) && defined(USE_NEON)
```

compute/cker/include/cker/operation/Exp.h
```diff
@@ -19,6 +21,7 @@
 #define __NNFW_CKER_EXP_H__
 
 #include "cker/Shape.h"
+#include "cker/neon/neon_check.h"
```
glistening, line 24:
```suggestion
#if defined(USE_FAST_SOFTMAX)
#include "cker/neon/neon_check.h"
#endif
```

compute/cker/include/cker/operation/SoftMax.h
```diff
@@ -22,6 +22,8 @@
 #include "cker/Utils.h"
 #include "cker/Types.h"
 #include "cker/eigen/Utils.h"
+#include "cker/neon/neon_check.h"
+#include "cker/operation/Exp.h"
```
glistening, line 26:
```suggestion
#if defined(USE_FAST_SOFTMAX)
#include "cker/neon/neon_check.h"
#include "cker/operation/Exp.h"
#endif
```

compute/cker/include/cker/operation/SoftMax.h
```diff
@@ -90,11 +92,35 @@ inline void Softmax(const float *in, const int input_size, const int batch_size,
 
     // Compute the normalized sum of exps.
     float exp_sum = 0.0;
+#ifdef USE_NEON
```
glistening, line 95:
```suggestion
#if defined(USE_FAST_SOFTMAX) && defined(USE_NEON)
```

glistening:
@YiHyunJin Could you finish this PR?

1. Enclose this PR's change with `USE_FAST_SOFTMAX` or something else you'd like as I suggested
2. Introduce config variable `USE_FAST_SOFTMAX` in `CfgOptionFlags.cmake` with default `OFF`. 
3. Define `USE_FAST_SOFTMAX` only when `USE_FAST_SOFTMAX` is `ON`.

infra/nnfw/cmake/CfgOptionFlags.cmake
```diff
@@ -48,6 +48,11 @@ option(BUILD_TFLITE_TEST "Build tensorflow lite test" OFF)
 option(BUILD_TFLITE_CLASSIFY_APP "Build tflite_classify app" OFF)
 option(BUILD_UBEN "Build micro-benchmark (uben) suite" OFF)
 option(BUILD_MLAPSE "Build mlapse benchmark toolkit" OFF)
+option(BUILD_FAST_SOFTMAX "Build mlapse benchmark toolkit" OFF)
```
glistening, line 51:
```suggestion
option(USE_FAST_SOFTMAX "Use fast but approximated softmax" OFF)
```

infra/nnfw/cmake/CfgOptionFlags.cmake
```diff
@@ -48,6 +48,11 @@ option(BUILD_TFLITE_TEST "Build tensorflow lite test" OFF)
...
+if(BUILD_FAST_SOFTMAX)
+    ADD_DEFINITIONS("-DUSE_FAST_SOFTMAX")
+endif(BUILD_FAST_SOFTMAX)
+
```
glistening, line 55: @YiHyunJin Please move those lines out of `CfgOptionFlags.cmake`. There are no logic in this file. `compute/cker/CMakeLists.txt` looks good.

glistening (changes requested):

glistening:
@YiHyunJin If we provides approximate but fast softmax kernel as build option defaulted off, it cannot be used unless building again. It may be better to provide runtime option instead of build option so that it can be selected by other runtime configuration. However, as I know, we need to provide a way for user to change runtime configuration.

cc @wateret

YiHyunJin:
The effect was insignificant, so it was closed.

PR #4432, opened 2020-09-25
[onert] Move comparison type validation check

hseok-oh:
Move comparision operation data type validation check to OperationValidator
Introduce positive and negative test for Equal
Introduce test result comparison method for boolean

### DIFF
--- a/runtime/onert/core/src/compiler/ShapeValidator.cc
+++ b/runtime/onert/core/src/compiler/ShapeValidator.cc
@@ -109,16 +109,9 @@ void ShapeValidator::visit(const ir::operation::BatchToSpaceND &node)
   OP_REQUIRES(input_shape.C == output_shape.C);
 }
 
-void ShapeValidator::visit(const ir::operation::Comparison &node)
+void ShapeValidator::visit(const ir::operation::Comparison &)
 {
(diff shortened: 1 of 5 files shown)


runtime/onert/core/src/compiler/ShapeValidator.cc
```diff
@@ -109,18 +109,6 @@ void ShapeValidator::visit(const ir::operation::BatchToSpaceND &node)
   OP_REQUIRES(input_shape.C == output_shape.C);
 }
 
-void ShapeValidator::visit(const ir::operation::Comparison &node)
-{
-  const auto output_index{node.getOutputs().at(0)};
```
hyunsik-yoon, line 114: Since comparison may needs some shape validation (e.g., lhs.shape == rhsl.shape)
How about adding `// TODO write shape validation of comparison`?

hseok-oh: OK. I'll add comment

hyunsik-yoon (approved):
LGTM. Oh. unused param error.

hyunsik-yoon (approved):
LGTM

wateret (approved):
LGTM

PR #4434, opened 2020-09-25
[luci] Rewrite FuseBCQPass running code to handle metadata

llFreetimell:
Parent Issue : #4337
Full Draft : #4338

As applying method is changed to use metadata, this commit will rewrite
`FuseBCQPass` ruinning code to handle it.

### DIFF
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -499,7 +499,8 @@ bool FuseBCQPass::run(loco::Graph *g)
       const auto bundle_cnt = metadata_node->at<loco::DataType::S32>(3);
 
       BCQFuser<1> fuser{original_output_cnt, bundle_cnt};
-      changed |= fuser.fuseBCQ(g);
+      if (fuser.fuseBCQ(g))
+        changed = true;
     }
     else
     {
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -17,6 +17,7 @@
 #include "luci/Pass/FuseBCQPass.h"
 
 #include <luci/IR/CircleNodes.h>
+#include <luci/Log.h>
 
 #include <cassert>
 #include <string>
@@ -502,7 +503,8 @@ bool FuseBCQPass::run(loco::Graph *g)
     }
     else
     {
-      assert(false && "Not supported BCQ version");
+      LOGGER(l);
+      WARN(l) << "Not supported BCQ version is found." << std::endl;
     }
 
     // Remove all of BCQ information nodes iff there is no change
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -445,38 +456,70 @@ bool FuseBCQPass::run(loco::Graph *g)
...
-  if (changed && version_node != nullptr)
-  {
-    // If BCQ is applied and version node was found, remove the node.
-    loco::replace(version_node).with(createNoOp(version_node));
+      BCQFuser<1> fuser{original_output_cnt, bundle_cnt};
+      changed |= fuser.fuseBCQ(g);
+    }
+    else
+    {
+      assert(false && "Not supported BCQ version");
```
seanshpark, line 505: Please do not assert for Pass.

compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -445,38 +456,70 @@ bool FuseBCQPass::run(loco::Graph *g)
...
+    if (bcq_version == 1)
+    {
+      const auto bundle_cnt = metadata_node->at<loco::DataType::S32>(3);
 
-  if (changed && version_node != nullptr)
-  {
-    // If BCQ is applied and version node was found, remove the node.
-    loco::replace(version_node).with(createNoOp(version_node));
+      BCQFuser<1> fuser{original_output_cnt, bundle_cnt};
+      changed |= fuser.fuseBCQ(g);
```
seanshpark, line 501: Please do not use bitwise for bool.

llFreetimell: If the result of `fuseBCQ(g)` is always true, we can write as `changed = fuser.fuseBCQ(g)` because `changed` will be always `true`. However, when `fuseBCQ(g)` returns `false`, `changed` should not be changed.
In addition, after #2694 issue was found, some of other passes in `luci` are implemented like above.
https://github.com/Samsung/ONE/blob/39a5a952463aeb40fae3390b239983326100b9a4/compiler/luci/pass/src/FuseBatchNormWithTConv.cpp#L153

Do you have any reason not to use bitwise operator?

seanshpark: > In addition, after #2694 issue was found,

I'd like to change all the codes that uses `|=`

> Do you have any reason not to use bitwise operator?

I don't think boolean is not bit value. it's just boolean, true of false;

Please use like
```
if (fuser.fuseBCQ(g))
   changed = true;
```

seanshpark (approved):
LGTM thank you!

jinevening (approved):
@seanshpark  

I skimmed through the code, but my review for this PR is not very reliable, because I don't have any background knowledge about BCQ codes.

seanshpark:
>  I don't have any background knowledge about BCQ codes.

Me too :)

PR #4437, opened 2020-09-28
[loco] Revise Nodes.test

seanshpark:
This will revise Nodes.test to use ASSERT_TRUE, ASSERT_FALSE for boolean type


jinevening (approved):
LGTM

PR #4441, opened 2020-09-28
[onert] Handle errors when Pool2D invalid params

wateret:
Handle errors when Pool2D operations have invalid params.
Add 5 nagative test cases.

NOTE Negative stride values are not allowed for Pool2D but it is
allowed for some other operations.


wateret:
The valid value ranges are referred from https://github.com/tensorflow/tensorflow/blob/c603937ed5cb3cc753bf988656937d7233ecd76f/tensorflow/python/kernel_tests/pool_test.py#L115-L119

glistening (approved):
LGTM

PR #4443, opened 2020-09-28
[onert] Remove deallocSubgraphOutput method

wateret:
Remove deallocSubgraphOutput in TensorManager class. Now all subgraph
outputs are all user-given buffers, we do not need this method. (already
never used.)


hyunsik-yoon (approved):
LGTM

PR #4456, opened 2020-09-29
[nnfwapi] Add a While test

wateret:
Add a While operation test with two inputs.


ragmani (approved):
LGTM

PR #4458, opened 2020-09-29
[onert] Refine onert/graph test cases

wateret:
This commit refines onert/graph test cases.

- Rename some tests (to be negative tests)
    - According to the rule I made - "If a test contains both positive
      and negative tests, it is a negative test."
    - Some tests must be renamed to be negative tests
- Add new negative tests


hseok-oh (approved):
LGTM

PR #4462, opened 2020-09-29
[luci] Support new metadata for BCQ in FuseBCQPass

llFreetimell:
Parent Issue : #4337
Full Draft : #4338

This commit will enable supporting new metadata for BCQ in `FuseBCQPass` as followings
- Pattern for figuring out `Einsum` is added.
- `int32_t` is used for prefix instead of `string`
- Redundant getters are removed

### DIFF
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -45,11 +45,18 @@ public:
     for (auto node : output_nodes)
     {
       auto output_node = loco::must_cast<luci::CircleOutput *>(node);
+
+      /**
+       * First output of model is metadata for BCQ. Please refer to following example.
+       *
+       * When original_output_cnt is 2,
+       * BCQ_METADATA, original_output_1, original_output_2, BCQ_INFO_1, ...
+       */
       if ((int)output_node->index() > _original_output_cnt)
       {
         const auto prefix = (output_node->index() - (_original_output_cnt + 1)) / (_bundle_cnt);
-        const MetadataType metadata_type =
-            (MetadataType)((output_node->index() - (_original_output_cnt + 1)) % (_bundle_cnt));
+        const MetadataType metadata_type = static_cast<MetadataType>(
+            (output_node->index() - (_original_output_cnt + 1)) % (_bundle_cnt));
         const auto circle_node = loco::must_cast<luci::CircleNode *>(output_node->from());
         add_BCQ_info_node(prefix, metadata_type, circle_node);
       }
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -20,60 +20,7 @@
 #include <luci/Log.h>
 
 #include <cassert>
-#include <string>
 #include <set>
-
-namespace
-{
-
-/**
- * @brief Circle nodes including BCQ information and a circle node to which BCQ will be applied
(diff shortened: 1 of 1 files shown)


seanshpark:
Code looks OK but I don't understand the logic.

compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -93,32 +40,43 @@ template <> class BCQFuser<1>
...
   bool fuseBCQ(loco::Graph *g)
   {
-    for (auto node : loco::all_nodes(g))
+
+    const auto output_nodes = loco::output_nodes(g);
+    for (auto node : output_nodes)
     {
-      if (auto circle_const = dynamic_cast<luci::CircleConst *>(node))
+      auto output_node = loco::must_cast<luci::CircleOutput *>(node);
+      if ((int)output_node->index() > _original_output_cnt)
```
mhs4670go, line 48: I'm not sure but `_orginal_output_cnt` might be variable that has the number of outputs without bcq related outputs. Then, isn't it `output_node->index() >= _original_output_cnt` that satisfy the condition as you intended? Because if the number of original output is 3, bcq node index that starts from 0 would starts from 3.

llFreetimell: You are understanding about the code very well but just one more thing, first output of BCQ model is metadata :)
Following example may help you
```
original_output_cnt = 2

BCQ_METADATA, original_output_1, original_output_2, BCQ_info1, BCQ_info2, ...
```

mhs4670go: @llFreetimell How about adding the comment as what you said?

compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -93,32 +40,43 @@ template <> class BCQFuser<1>
...
       {
-        add_BCQ_info_node(circle_const);
+        const auto prefix = (output_node->index() - (_original_output_cnt + 1)) / (_bundle_cnt);
+        const MetadataType metadata_type =
+            (MetadataType)((output_node->index() - (_original_output_cnt + 1)) % (_bundle_cnt));
```
mhs4670go, line 52: (optional) It would be better to use `static_cast` in c++. It does cast in compile time and make it more readable.

compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -93,32 +40,43 @@ template <> class BCQFuser<1>
...
-          bcq_gather->input_scales(get_alpha(params));
-          bcq_gather->input_binary(get_packed_binary_code(params));
+          bcq_gather->input_scales(_alpha[prefix]);
+          bcq_gather->input_binary(_packed_binary_code[prefix]);
           bcq_gather->indices(gather->indices());
-          bcq_gather->input_clusters(packed_clusters(params));
+          bcq_gather->input_clusters(packed_clusters(g, prefix));
 
-          if (do_w_x(params))
+          if (_do_w_x[prefix]->at<loco::DataType::BOOL>(0))
```
mhs4670go, line 79: Is it safe to access map without index checking?

llFreetimell: I will add that validation to `is_bcqinfo_valid` in another PR :)

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #4463, opened 2020-09-30
[onert] Arithmetic kernels without fused activation

binarman:
Introduced kernels for non-fused arithmetic operations: Add, Mul, Sub

### DIFF
--- a/compute/cker/include/cker/operation/BinaryArithmeticOps.h
+++ b/compute/cker/include/cker/operation/BinaryArithmeticOps.h
@@ -19,6 +19,7 @@
 #define __NNFW_CKER_BINARY_ARITHMETIC_OPS_H__
 
 #include <functional>
+#include <limits>
 #include "cker/operation/optimized/BinaryArithmeticOps.h"
 #include "cker/operation/reference/BinaryArithmeticOps.h"
 #include "cker/Shape.h"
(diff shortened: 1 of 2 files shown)


krayzemli:
This PR most likely will conflict with newer PR #4465. Please, check PR #4465 first.

glistening:
@binarman I think you have the experimental number for our target model. Could you please write down how much improvement it bring? Please write `before` and `after`'s number of `nnpackage_run --nnpackage our_model -w 300 -r 1200` on your android aarch64 device.

binarman:
@glistening 
I've tested changes again like this (`--num_runs=10000 --warmup_runs=10`):
```
for i in $(seq 20); do taskset -a 128 ./nnpackage_run --num_runs=10000 --warmup_runs=10 --nnpackage package 2>&1 | grep " MEAN"; sleep 3; done
```

Using large number of runs I can get more or less stable results.
In short I got 1.2096 ms -> 1.2011 ms (0.65% improvement :man_shrugging:)

Full times output are:
With patch: [1.205, 1.199, 1.197, 1.206, 1.203, 1.236, 1.206, 1.186, 1.197, 1.234, 1.169, 1.202, 1.207, 1.193, 1.190, 1.197, 1.198, 1.192, 1.199, 1.206]
Without patch: [1.221, 1.194, 1.192, 1.225, 1.230, 1.207, 1.201, 1.223, 1.203, 1.209, 1.191, 1.188, 1.223, 1.204, 1.209, 1.212, 1.195, 1.218, 1.236, 1.211]

P.s. I also found that some BIG cores are faster then others on my phone, I didn't get why this is happening yet, but at least they are stable. So do not be scared by ~1.2 ms vs ~0.4 ms on your device.

binarman:
As @krayzemli wrote he included these changes in his PR, so we can close this one.

PR #4440, opened 2020-09-28
DRAFT: test one-cmds

seanshpark:
On-going draft to test installed one-cmds

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@samsung.com>

Issue #4326, opened 2020-09-17
Compiler FE: Testing one-cmds


seanshpark:
Let's prepare stricpt to test one-cmds command line tools to automate #4155 testing.
Later we can add this scipt in CI.

seanshpark:
Test objects
- one-import
- one-import-tf
- one-import-tflite
- one-import-bcq
- one-optimize
- one-quantize
- one-pack
- one-codegen

seanshpark:
Test should be done with installed components:
- a test script for each test case
- a driver script to run individual test scripts
- test material files
- python virtual env folder

seanshpark:
Done

Issue #4155, opened 2020-09-09
Compiler FE: testing with models (1.9.0)


seanshpark:
This issue is to describe compiler frontend typical positve usage and negative usage.
- positive usage: typical usage of our compiler frontend with supported models.
- negative usage: compiler frontend results of unsupported models or invalid files

This issue assumes
- description is based on recent as-of-today 1.8.0 release
- run in Ubuntu 16.04 or 18.04 running on x86-64

Similar issue: #1179
Revision of #3864

seanshpark:
## Download and install 
- download from https://github.com/Samsung/ONE/releases/tag/1.9.0
   - https://github.com/Samsung/ONE/releases/download/1.9.0/nncc-1.9.0.tar.gz
- extract files. this will look like this
   - <details>

      ```
      $ tree
      .
      ├── bin
      │   ├── circle2circle
      │   ├── circle-quantizer
      │   ├── generate_bcq_output_arrays
      │   ├── model2nnpkg.sh
      │   ├── one-codegen
      │   ├── one-import
      │   ├── one-import-bcq
      │   ├── one-import-tf
      │   ├── one-import-tflite
      │   ├── one-optimize
      │   ├── one-pack
      │   ├── one-prepare-venv
      │   ├── one-quantize
      │   ├── one-version
      │   ├── preserve_bcq_info
      │   ├── record-minmax
      │   ├── tf2nnpkg
      │   ├── tf2tfliteV2.py
      │   └── tflite2circle
      ├── doc
      │   ├── how-to-prepare-virtualenv.txt
      │   └── how-to-use-one-commands.txt
      ├── lib
      │   ├── libloco.so
      │   ├── libluci_env.so
      │   ├── libluci_export.so
      │   ├── libluci_import.so
      │   ├── libluci_interpreter.so
      │   ├── libluci_lang.so
      │   ├── libluci_logex.so
      │   ├── libluci_log.so
      │   ├── libluci_pass.so
      │   └── libluci_service.so
      └── nncc-1.9.0.tar.gz
      ```

      </details>
- add above `bin` folder to your `PATH` envrionment
- prepare python virtual environment. for more information, read `how-to-prepare-virtualenv.txt` file in `doc` folder.
   ```
   $ one-prepare-venv
   ```

seanshpark:
## Test case 1: Typical positive usage

Convert to `.circle` file
```
one-import tf \
--input_path ./inception_v3.pb \
--output_path ./inception_v3.circle \
[...]

PR #4467, opened 2020-10-05
[one-cmds] Tests for one-import-bcq

seanshpark:
This will introduce test scripts for one-import-bcq


seanshpark:
For #4326, from draft #4440, test case from #4155 (comment)

mhs4670go (approved):
LGTM

PR #4469, opened 2020-10-05
[one-cmds] Tests for one-pack

seanshpark:
This will introduce test scripts for one-pack


seanshpark:
For #4326, from draft #4440, test case from https://github.com/Samsung/ONE/issues/4155#issuecomment-690903095

mhs4670go (approved):
LGTM

PR #4470, opened 2020-10-05
[one-cmds] Tests for one-quantize

seanshpark:
This will introduce test scripts for one-quantize

### DIFF
--- a/compiler/one-cmds/tests/one-quantize_001.test
+++ b/compiler/one-cmds/tests/one-quantize_001.test
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+# Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


seanshpark:
For #4326, from draft #4440, test case from https://github.com/Samsung/ONE/issues/4155#issuecomment-690903095

compiler/one-cmds/tests/one-quantize_001.test
```diff
@@ -0,0 +1,54 @@
...
+trap trap_err_onexit ERR
+
+inputfile="./inception_v3.circle"
+outputfile="./inception_v3.quantized.circle"
+
+rm -rf ${outputfile}
+
+# to create inception_v3.circle
+if [[ ! -s ${inputfile} ]]; then
+  /bin/bash one-import_001.test >> /dev/null
```
mhs4670go, line 35: Maybe  `one-import_001.test` will be merged later?

seanshpark: `one-import_001.test` is in another PR. Test is not enabled yet so it is OK.

mhs4670go:
Except import test, they tests the process that follow import. So, `one-import_001` can be run in the prepare section first, which reduces test time I think.

seanshpark:
> So, one-import_001 can be run in the prepare section first

I'll post another PR for `prepare_test_materials.sh` file patch

> which reduces test time I think.

I've updated with `one-import_001.test` PR so that `circle` file is not deleted.
Current code to call to `one-import_001.test` would be good to stay as-is for safety and will be called when file doesn't exist.

mhs4670go (approved):
LGTM

PR #4471, opened 2020-10-05
[onert] optimize getTensorShape(IPortableTensor*)

glistening:
getTensorShape make a temporary copy for shape.
Then, it is copied again in nnfw::shape::cker::Shape constructor.
It will remove unnecessary copy and allocation.

In addition, it adds final to cpu backend's dimension().
With final keyword, it gives better performance.


hseok-oh (approved):

periannath (approved):
LGTM

PR #4476, opened 2020-10-05
[luci] Revise SparsityParam

mhs4670go:
This commit revises sparsity parameter.

Related : #3889
Draft: #4394

### DIFF
--- a/compiler/luci/lang/include/luci/IR/SparsityParam.h
+++ b/compiler/luci/lang/include/luci/IR/SparsityParam.h
@@ -18,6 +18,8 @@
 #define __LUCI_IR_SPARSITYPARAM_H__
 
 #include <cstdint>
+#include <stdexcept>
+#include <utility>
 #include <vector>
 
 namespace luci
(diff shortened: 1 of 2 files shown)


compiler/luci/lang/include/luci/IR/SparsityParam.h
```diff
@@ -37,23 +39,186 @@ enum SparseIndexVectorType
   U8,
 };
 
-struct DimMetaData
+class SparseIndexVector
 {
+public:
+  SparseIndexVector() = default;
```
seanshpark, line 45: I think default CTOR should be removed
```cxx
SparseIndexVector() = delete;
```
Current DTOR uses `_type` to deletes what was created

seanshpark: Maybe remove copy CTOR too?

mhs4670go:
```cxx
DimMetaData(DimensionType format, int32_t dense_size) : _format{format}, _dense_size{dense_size}
  {
    // DO NOTHING
  }
```
`DimMetaData` can be two types, e.g., dense, sparse. when it is sparse, `DimMetaData` calls above CTOR that calls default `SparseIndexVector` CTOR.

seanshpark: Ah... didn't read the code carefully :)

seanshpark (approved):
LGTM

PR #4479, opened 2020-10-05
[one-cmds] Install tests scripts

seanshpark:
This will install tests scripts for one-cmd when TEST is enabled


mhs4670go (approved):
LGTM

PR #4486, opened 2020-10-05
[luci] Introduce SparsifyTensorPass

mhs4670go:
This commit introduces SparsifyTensorPass to luci.

Related : #3889
Draft: #4394

### DIFF
--- a/compiler/luci/pass/src/SparsifyTensorPass.cpp
+++ b/compiler/luci/pass/src/SparsifyTensorPass.cpp
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


compiler/luci/pass/src/SparsifyTensorPass.cpp
```diff
@@ -0,0 +1,117 @@
...
+        assert(_block_size.at(i) == dim_metadata.at((_format.size() + i) * 2).at(0));
+        sparsityparam->dim_metadata.emplace_back(DimensionType::DENSE, _block_size.at(i));
+      }
+      cop->sparsityparam(std::move(sparsityparam));
+      break;
+    }
+    default:
+      break;
+  }
+  return true;
```
seanshpark, line 95: This will return true event type is not FLOAT32... is this OK?

mhs4670go:
@seanshpark PTAL:)

compiler/luci/pass/src/SparsifyTensorPass.cpp
```diff
@@ -0,0 +1,123 @@
...
+        sparsify_tensor<loco::DataType::S32>(cop);
+        break;
+      case loco::DataType::S8:
+        sparsify_tensor<loco::DataType::S8>(cop);
+        break;
+      case loco::DataType::FLOAT32:
+        sparsify_tensor<loco::DataType::FLOAT32>(cop);
+        break;
+      default:
+        std::runtime_error("SparsifyTensorPass: Unsupported dtype.");
```
seanshpark, line 111: I'm not sure it would be better to just `continue` or stop the program with `throw`

@llFreetimell , what do you think?

llFreetimell: I think that throwing is better.
For `FuseBCQPass`, it was included in `--all` option at `circle2circle` thus `continue` was better than `throw`.
However, `SparsifyTensorPass` is not included in `--all` option and users of this Pass should give valid `_tensor_name`.

Therefore, for me, `throw` looks better :)

llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

PR #4487, opened 2020-10-05
[one-cmds] Allow to call test script from outside test directory

hseok-oh:
Allow to call test script without changing current directory


seanshpark (approved):
LGTM thank you!

PR #4488, opened 2020-10-05
[one-cmds] Create python symbolic link

hseok-oh:
one-prepare-venv create python symbolic link
Can use virtual environment python by setting PATH [install_dir]/bin only


seanshpark (approved):
LGTM

seanshpark:
@hseok-oh , is it safe to add a symbolic link to python executable?

I'm not good at python virtual env...
`venv/bin/python3` itself is symbolic link to system `python`.
if we run this symbolic link of `python` in `venv` folder,
can it load wrong libraries from the system not from the `venv/lib` ?

hseok-oh:
@seanshpark My local environment don't have tensorflow-cpu package. But virtual environment by using this PR and test without setting PATH to installed venv passed. It seems work well.

seanshpark:
@hseok-oh , OK thanks! Let's go and fix if anything goes wrong :)

Issue #4490, opened 2020-10-06
[one-cmds] User notification if venv is not prepared


seanshpark:
With installed commands,
- User may not read documents
- User may not prepare virtual environment

As of result, `one-import-tf` may cause this error

> ImportError: No module named tensorflow

@parjong proposed

> It may be better to show some warning (e.g. Run one-prepare-venv) if venv does not exist.

seanshpark:
Condition: if `venv` folder is not found
Action: show message `Virtual environment not found. Please run 'one-prepare-venv' command.`
For: `one-import-tf`, `one-import-bcq`

seanshpark:
Done

PR #4491, opened 2020-10-06
[one-cmds] Show error if venv not prepared

seanshpark:
This will revise commands that require venv to show error message and a guide if not prepared


mhs4670go (approved):
LGTM

seanshpark:
For #4490

parjong (approved):
👍

PR #4494, opened 2020-10-06
[common-artifacts]Exclude from example creation.

struss:
Parent Issue : #4151
Draft PR : #4260

This commit add `UnidirectionalSequenceLSTM` to exclusion list using TC generation.

### DIFF
--- a/compiler/common-artifacts/exclude.lst
+++ b/compiler/common-artifacts/exclude.lst
@@ -5,7 +5,6 @@
 
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
-optimize(UnidirectionalSequenceLSTM_000) # This will be removed after all luci components done.
 optimize(UnidirectionalSequenceLSTM_001) # This recipe contains is_variable Tensor
 
 ## CircleRecipes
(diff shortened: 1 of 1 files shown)


compiler/common-artifacts/exclude.lst
```diff
@@ -5,6 +5,8 @@
 
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
+optimize(UnidirectionalSequenceLSTM_000) # This will be removed after all luci components done.
```
mhs4670go, line 8: Does `UnidirectionalSequenceLSTM_000` contain things `circle2circle` cannot run like `is_variable`? AFAIK, it doesn't need to add it to `optimize` exclude list if related PRs are merged proper order.

struss: No, it is model that composed of all weight, bias and etc as input. If I add recipe at the last part, those line is not needed. I will remove that and add recipe on last part(after luci component add.)

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #4497, opened 2020-10-06
[release_tool] fix minor bug

mhs4670go:
This commit makes `git_release.sh` occur an error when its system
doesn't have enhanced getopt instead of echo.


llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

PR #4498, opened 2020-10-06
Uninitialized 'quantized_dimension'

cgbahk:
Let's see if it works.


cgbahk:
At least it seems not failed test

PR #4499, opened 2020-10-06
[circle2circle] Add sparsify-related option

mhs4670go:
This commit adds sparsify-related option to circle2circle.

Related : #3889
Draft: #4394

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -214,6 +214,11 @@ int entry(int argc, char **argv)
     if (arser["--sparsify_block_size"])
       options->param(AlgorithmParameters::Sparsify_block_size,
                      arser.get<std::string>("--sparsify_block_size"));
+    else
+    {
+      std::cerr << "ERROR: Block size not provided" << std::endl;
+      return 255;
+    }
     options->param(AlgorithmParameters::Sparsify_block_map,
                    arser.get<std::string>("--sparsify_block_map"));
   }
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -123,6 +123,41 @@ int entry(int argc, char **argv)
   arser.add_argument("input").nargs(1).type(arser::DataType::STR).help("Input circle model");
   arser.add_argument("output").nargs(1).type(arser::DataType::STR).help("Output circle model");
 
+  // sparsification argument
+  arser.add_argument("--sparsify_tensor")
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .required(false)
+      .help("Tensor name that you want to sparsify");
+
+  arser.add_argument("--sparsify_traversal_order")
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .required(false)
+      .default_value("0,1,2,3")
+      .help("Traversal order of dimensions. Default value: 0,1,2,3");
+
+  arser.add_argument("--sparsify_format")
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .required(false)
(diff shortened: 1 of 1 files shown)


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -167,6 +202,22 @@ int entry(int argc, char **argv)
...
+    options->param(AlgorithmParameters::Sparsify_traversal_order,
+                   arser.get<std::string>("--sparsify_traversal_order"));
+    options->param(AlgorithmParameters::Sparsify_format,
+                   arser.get<std::string>("--sparsify_format"));
+    if (arser["--sparsify_block_size"])
+      options->param(AlgorithmParameters::Sparsify_block_size,
+                     arser.get<std::string>("--sparsify_block_size"));
+    options->param(AlgorithmParameters::Sparsify_block_map,
+                   arser.get<std::string>("--sparsify_block_map"));
+  }
```
seanshpark, line 219: Can this block work when only `sparsify_tensor` option is provided?

compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -123,6 +123,41 @@ int entry(int argc, char **argv)
...
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .required(false)
+      .help("Tensor name that you want to sparsify");
+
+  arser.add_argument("--sparsify_traversal_order")
+      .nargs(1)
+      .type(arser::DataType::STR)
+      .required(false)
+      .default_value("0,1,2,3")
```
seanshpark, line 137: Just curiois about how to test this, what are the limits or conditions of the numbers?

compiler/circle2circle/src/Circle2Circle.cpp
mhs4670go: @seanshpark Yes. But it would be the tensor far from what user think it would be. Actually, it only supports block-sparsification. It should be expanded to general sparsification later.

compiler/circle2circle/src/Circle2Circle.cpp
mhs4670go: It is a little bit complicated to make the explanation short. It comes from circle schema as below.
```flatbuffers
// Parameters to encode a sparse TfLite tensor.
table SparsityParameters {
  // The traversal order of the dimensions defined in the `shape` field of the
  // conceptual dense tensor. For a n-dimensional tensors with dims (d0, d1,
  // ..., dn-1),
  //   - if not block sparse, the traversal_order is just a permutation of (d0,
  //     ..., dn-1). For example, a 2-D matrix stored in row-major order would
  //     have traversal_order = (d0, d1).
  //   - if block sparse with a k-dimensional block (0 <= k <= n), the
  //     traversal_order has n + k elements. The first n elements are still a
  //     permutation of (d0, ..., dn-1). The lask k elements are a permutation
  //     of (dn, ..., dn+k-1), defining how to traverse a block internally. For
  //     example, a 2-D matrix with 2-D blocks, both stored in row-major order
  //     would have traversal_order = (d0, d1, d2, d3).
  traversal_order:[int];
  // For an n-dimensional tensor with a k-dimensional block (0 <= k <= n),
  // stores how a block dimension in (dn, ..., dn+k-1) maps to the original
  // tensor dimension in (d0, ..., dn).
  // It's stored in the order of (dn, ..., dn+k-1).
  // If not block-sparse, this field is NULL.
  block_map:[int];
  // In the traversal order defined above, the metadata needed for
  // each dimension to locate the non-zero values in the original dense tensor.
  // The size of the dim_metadata array = the size of the traversal_order array
  // = n + k.
  dim_metadata:[DimensionMetadata];
}
```

seanshpark (approved):
LGTM

PR #4502, opened 2020-10-06
[luci/import]Add UnidirectionalSequenceLSTM on luci/import

struss:
Parent Issue : #4151
Draft PR : #4260

This commit add `UnidirectionalSequenceLSTM` on luci/import

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleUnidirectionalSequenceLSTM.cpp
+++ b/compiler/luci/import/src/Nodes/CircleUnidirectionalSequenceLSTM.cpp
@@ -36,93 +36,38 @@ CircleNode *CircleUnidirectionalSequenceLSTMGraphBuilder::build_node(
 {
   auto *node = graph->nodes()->create<CircleUnidirectionalSequenceLSTM>();
   node->input(inputs.at(0));
-
   node->input_to_input_weights(inputs.at(1)); // Optional
-  if (auto input_to_input_weights =
-          dynamic_cast<luci::CircleOutputExclude *>(node->input_to_input_weights()))
-  {
-    input_to_input_weights->dtype(loco::DataType::FLOAT32);
-  }
   node->input_to_cell_weights(inputs.at(2));
   node->input_to_forget_weights(inputs.at(3));
   node->input_to_output_weights(inputs.at(4));
-
   node->recurrent_to_input_weights(inputs.at(5)); // Optional
-  if (auto recurrent_to_input_weights =
-          dynamic_cast<luci::CircleOutputExclude *>(node->recurrent_to_input_weights()))
-  {
-    recurrent_to_input_weights->dtype(loco::DataType::FLOAT32);
-  }
   node->recurrent_to_cell_weights(inputs.at(6));
   node->recurrent_to_forget_weights(inputs.at(7));
   node->recurrent_to_output_weights(inputs.at(8));
-
-  node->cell_to_input_weights(inputs.at(9)); // Optional
-  if (auto cell_to_input_weights =
-          dynamic_cast<luci::CircleOutputExclude *>(node->cell_to_input_weights()))
-  {
-    cell_to_input_weights->dtype(loco::DataType::FLOAT32);
-  }
+  node->cell_to_input_weights(inputs.at(9));   // Optional
   node->cell_to_forget_weights(inputs.at(10)); // Optional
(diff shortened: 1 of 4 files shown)


compiler/luci/import/src/Nodes/CircleUnidirectionalSequenceLSTM.cpp
```diff
@@ -0,0 +1,137 @@
...
+  {
+    input_to_input_weights->dtype(loco::DataType::FLOAT32);
+  }
+  node->input_to_cell_weights(inputs.at(2));
+  node->input_to_forget_weights(inputs.at(3));
+  node->input_to_output_weights(inputs.at(4));
+
+  node->recurrent_to_input_weights(inputs.at(5)); // Optional
+  if (auto recurrent_to_input_weights =
+          dynamic_cast<luci::CircleOutputExclude *>(node->recurrent_to_input_weights()))
```
seanshpark, line 52: as `recurrent_to_input_weights()` is `at(5)->node();`  and we have in `loco`, `loco::Node *arg(uint32_t n)` we can revise this as
```cxx
auto recurrent_to_input_weights = dynamic_cast<luci::CircleOutputExclude *>(node->arg(5));
```
then we can add a loop and change all `if (...) { ... }` to one line with an array having all optional index `[5, 9, 10, ...]`...

something like
```cxx
std::vector<int32_t> optionals = { 5, 9, 10, ... }
for (auto optional : optionals)
{
  if (auto inp = dynamic_cast<luci::CircleOutputExclude *>(node->arg(optional)))
  {
    inp ->dtype(loco::DataType::FLOAT32);  
  }
}
```

seanshpark: it would be nice if you could add a comment about the reason why we call `dtype(loco::DataType::FLOAT32);`...

struss: Oh, Okay I will fix it!

struss: > it would be nice if you could add a comment about the reason why we call `dtype(loco::DataType::FLOAT32);`...

Oh, I forget. This Data type depends on `input_to_output_weights` type. I will fix it.

compiler/luci/import/src/Nodes/CircleUnidirectionalSequenceLSTM.cpp
```diff
@@ -0,0 +1,83 @@
...
+  return true;
+}
+
+CircleNode *CircleUnidirectionalSequenceLSTMGraphBuilder::build_node(
+    const circle::OperatorT &op, const std::vector<CircleNode *> &inputs, loco::Graph *graph) const
+{
+  auto *node = graph->nodes()->create<CircleUnidirectionalSequenceLSTM>();
+  node->input(inputs.at(0));
+  node->input_to_input_weights(inputs.at(1)); // Optional
+  node->input_to_cell_weights(inputs.at(2));
```
seanshpark, line 40: Q) if `input_to_input_weights` didn't exist as it's optional then is this `input_to_cell_weights` be `inputs.at(2)` or `inputs.at(1)` ?
Usually we had optional at the end so I'm not sure existing current codes will work for this PR.
Can you make a recipe for a test that some of the optional nodes in the middle is absent and this works?

struss: Okay I will test it on my draft!

struss: >Q) if `input_to_input_weights` didn't exist as it's optional then is this `input_to_cell_weights` be `inputs.at(2)` or `inputs.at(1)` ?
@seanshpark as I tested on my draft, I found that inputs have `luci::CircleOutputExclude` on empty input(if options are not provided case). So, inputs.at(2) will be in `input_to_cell_weights`.

seanshpark: OK thanks!
Please add your test material (recipe files I guiess) too :)

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #4504, opened 2020-10-06
[luci/service]Add UnidirectionalSequenceLSTM on luci/service

struss:
Parent Issue : #4151
Draft PR : #4260

This commit add Type Inference/Shape Inference on luci.

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.cpp
@@ -2393,14 +2393,14 @@ public:
     return infer_transpose_conv(node);
   }
 
+  loco::NodeShape visit(const luci::CircleUnpack *node) final { return infer_unpack(node); }
+
   loco::NodeShape visit(const luci::CircleUnidirectionalSequenceLSTM *node) final
   {
     return infer_unidirectionalsequencelstm(node);
   }
 
   loco::NodeShape visit(const luci::CircleUnique *node) final { return infer_unique(node); }
-
-  loco::NodeShape visit(const luci::CircleUnpack *node) final { return infer_unpack(node); }
 
   loco::NodeShape visit(const luci::CircleWhere *node) final { return use_own(node); }
 
--- a/compiler/luci/service/src/CircleShapeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.cpp
@@ -1613,7 +1613,7 @@ loco::NodeShape infer_unidirectionalsequencelstm(const luci::CircleUnidirectiona
   auto input_shape = loco::shape_get(node->input()).as<loco::TensorShape>();
   auto recurrent_to_output_weights =
       loco::shape_get(node->recurrent_to_output_weights()).as<loco::TensorShape>();
-  auto rank = static_cast<int32_t>(input_shape.rank());
+  auto rank = input_shape.rank();
   loco::TensorShape output_shape;
   output_shape.rank(rank);
   for (int32_t i = 0; i < rank - 1; i++)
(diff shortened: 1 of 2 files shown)


compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -2377,10 +2393,15 @@ class ShapeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::NodeS
...
 
-  loco::NodeShape visit(const luci::CircleUnpack *node) final { return infer_unpack(node); }
+  loco::NodeShape visit(const luci::CircleUnidirectionalSequenceLSTM *node) final
+  {
+    return infer_unidirectionalsequencelstm(node);
+  }
 
   loco::NodeShape visit(const luci::CircleUnique *node) final { return infer_unique(node); }
 
+  loco::NodeShape visit(const luci::CircleUnpack *node) final { return infer_unpack(node); }
```
seanshpark, line 2403: I think this change be in separate PR :)

struss: Okay. When I add a this, I found this is not alphabetic order. so I just thought revise this, but this is not a part of adding new OP. so, I will rollback it.

seanshpark (approved):
LGTM thank you!

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -1608,6 +1608,22 @@ loco::NodeShape infer_unpack(const luci::CircleUnpack *node)
   return loco::NodeShape{output_shape};
 }
 
+loco::NodeShape infer_unidirectionalsequencelstm(const luci::CircleUnidirectionalSequenceLSTM *node)
+{
+  auto input_shape = loco::shape_get(node->input()).as<loco::TensorShape>();
+  auto recurrent_to_output_weights =
+      loco::shape_get(node->recurrent_to_output_weights()).as<loco::TensorShape>();
+  auto rank = static_cast<int32_t>(input_shape.rank());
```
mhs4670go, line 1616: Is there any reason of this `static_cast`?

struss: `input_shape.rank()` is a type of `uint32_t`. so, I just use static_cast.

mhs4670go: @struss got it. But it might as well goes on as it is without casting because rank shouldn't be minus value, which is why most of `rank` methods returns `uint32_t` I think.

struss: Oh, I will update without casting. thanks for pointing out!

mhs4670go (approved):
LGTM

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -1608,6 +1608,22 @@ loco::NodeShape infer_unpack(const luci::CircleUnpack *node)
...
 
+loco::NodeShape infer_unidirectionalsequencelstm(const luci::CircleUnidirectionalSequenceLSTM *node)
+{
+  auto input_shape = loco::shape_get(node->input()).as<loco::TensorShape>();
+  auto recurrent_to_output_weights =
+      loco::shape_get(node->recurrent_to_output_weights()).as<loco::TensorShape>();
+  auto rank = input_shape.rank();
+  loco::TensorShape output_shape;
+  output_shape.rank(rank);
+  for (int32_t i = 0; i < rank - 1; i++)
```
seanshpark, line 1619: `rank` is `uint32_t` then this may drop compilation error of type mismatch.

struss: Oh I will fix `i` as `uint32_t`.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #4508, opened 2020-10-06
[onert] Optimize StridedSlice

YiHyunJin:
when only last dimension used, we have easy accesse to index.
So I optimized it for that situation.
And tested any models. The result is an 3.5% increase in speed.
(1.233 -> 1.191)

### DIFF
--- a/compute/cker/include/cker/operation/StridedSlice.h
+++ b/compute/cker/include/cker/operation/StridedSlice.h
@@ -260,11 +260,39 @@ template <typename T>
 inline void StridedSlice(const StridedSliceParams &op_params, const Shape &unextended_input_shape,
                          const T *input_data, const Shape &unextended_output_shape, T *output_data)
 {
-  // Note that the output_shape is not used herein.
-  StridedSliceParams params_copy = op_params;
-
   assert(unextended_input_shape.DimensionsCount() <= 4);
   assert(unextended_output_shape.DimensionsCount() <= 4);
+
+  bool optimize = true;
+  int st_count = op_params.strides_count;
+  for (int idx = 0; idx < st_count - 1; idx++)
+  {
+    const int start = StartForAxis(op_params, unextended_input_shape, idx);
+    const int stop = StopForAxis(op_params, unextended_input_shape, 0, start);
+    if (start != 0 || stop != 0)
+    {
+      optimize = false;
+      break;
+    }
+  }
+
+  if (optimize)
+  {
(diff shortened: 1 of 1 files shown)


compute/cker/include/cker/operation/StridedSlice.h
```diff
@@ -260,12 +260,38 @@ template <typename T>
...
+  bool optimize = true;
+  int st_count = op_params.strides_count;
+  for (int idx = 0; idx < st_count - 1; idx++)
+  {
+    if (op_params.start_indices[idx] != 0 || op_params.stop_indices[idx] != 0)
+    {
+      optimize = false;
+      break;
+    }
+  }
```
glistening, line 275: @YiHyunJin Why don't you check in KernelGenerator?

wateret:
> @YiHyunJin Why don't you check in KernelGenerator?

How was this resolved? It looks like the param check is done in the kernel. Just adding some more description, for example,

- Split the optimized kernel and original kernel into two separate functions
- As we know the params at prepare(compile) time, move the param check code in KernelGenerator
- Then we are able to bind the most proper kernel for that single operation node, and no need to check it at execution time

BTW it looks good itself and this work can be done after merging this PR.

wateret (changes requested):
Please add a test case (for code coverage and verification)

wateret:

glistening:
`android-runtime-builder` fails because downloading 'http://sourceforge.net/projects/boost/files/boost/1.58.0/boost_1_58_0.tar.gz' failed.

glistening:
I've write testcase for this PR using CircleGen. Please see #4525.

wateret:
@YiHyunJin Could you please respond https://github.com/Samsung/ONE/pull/4508#issuecomment-704742171 ?

YiHyunJin:
@wateret It looks good to do after `buildStridedSliceParams`.
So I checked in `StridedSlice`.

> BTW it looks good itself and this work can be done after merging this PR.
Yes~

glistening (approved):
LGTM

wateret (approved):
LGTM

glistening:
@YiHyunJin This morning, I've tested with merged PR. In my environment (a bit different model), this PR does not make differences for our target model.

1.010 → 1.005 (about 0.49% improvement)

YiHyunJin:
@glistening  I checked this problem and i fixed.
https://github.com/Samsung/ONE/pull/4543

Issue #4278, opened 2020-09-15
Compiler FE: enhance one-cmds arguments interface


seanshpark:
Problem:
- Users or upper level tools(like SDK) will call our one-cmds command line tools with lots of arguments
- One time calling is OK but periodic calling with long arguments may be inconvenient (we may copy-paste or use bash history)
- With one or two argument changes to test artifacts

One solution that came up:
- To provide a easy-to-edit simple text configuration file(like old ini or cfg file) that can cover all the arguments
- We can then use existing arguments to override the configuration file items

Let's talk about this...
- Objective of this issue is to define the solution/specification to solve this problem
- And to provide this in Oct or Nov release

seanshpark:
An example
```
one-import tf -c detectimage.cfg
```
where there is `detectimage.pb` file as an input model

seanshpark:
## specification

To be added

mhs4670go:
## suggestion
### `one-compile.config`
```bash
# This configuration file is used to configure the parameters and initial settings for ONE command driver.

# input_path                    Path to input model
input_path=""
# output_path                   Path to output model
output_path=""

### one-import-tf ### one-import-bcq ###
# input_arrays                  Names of the input arrays, comma-separated
input_arrays=""
# input_shapes                  Input shapes, colon-separated
input_shapes=""
# output_arrays                 Names of the output arrays, comma-separated
output_arrays=""
# v2                            Use TensorFlow 2.x interface (default is 1.x interface)
v2="false"

### one-import-tflite ###

### one-optimize ###
# all                           Enable all optimization algorithms
all="true"
# fuse_bcq                      Enable FuseBCQ Pass
fuse_bcq="false"
# fuse_instnorm                 Enable FuseInstanceNormalization Pass
fuse_instnorm="false"
# resolve_customop_add          Enable ResolveCustomOpAddPass Pass
resolve_customop_add="false"
# resolve_customop_batchmatmul  Enable ResolveCustomOpBatchMatMulPass Pass
resolve_customop_batchmatmul="false"
# resolve_customop_matmul       Enable ResolveCustomOpMatMulPass Pass
resolve_customop_matmul="false"

### one-quantize ###
# input_dtype                   Input data type (supported: float32, default=float32)
input_dtype=""
# quantized_dtype               Output quantized data type (supported: uint8, default=uint8)
quantized_dtype=""
# granularity                   Quantize granularity (supported: layer, channel, default=layer)
granularity=""
[...]

PR #4509, opened 2020-10-06
[DRAFT] Introduce one-build

mhs4670go:
This draft is for one-build.

You can run `one-import-tf` with this draft as of now. Please review the configuration file.
```bash
$  ./one-build -c one-build.cfg
```
Once the configuration specification is approved, I'll write the rest of `one-build`.
Related : #4278 

### DIFF
--- a/compiler/one-cmds/one-build
+++ b/compiler/one-cmds/one-build
@@ -61,7 +61,7 @@ LONG_OPTS="\
 help,\
 version,\
 config:\
-`### configuration option ###`\
+`### configuration options ###`\
 input_path:,\
 output_path:,\
 `# one-import-tf, one-import-bcq`\
@@ -72,8 +72,28 @@ import_input_shapes:,\
 import_output_arrays:,\
 import_v2:,\
 import_tf_output_path:,\
-import_bcq_output_path:"
-# TO BE FILLED
+import_bcq_output_path:,\
+`# one-import-tflite`\
+import_tflite_output_path:,\
+`# one-optimize`\
+optimize_all:,\
+optimize_fuse_bcq:,\
+optimize_fuse_instnorm:,\
+optimize_resolve_customop_add:,\
+optimize_resolve_customop_batchmatmul:,\
+optimize_resolve_customop_matmul:,\
+optimize_output_path:,\
+`# one-quantize`\
+quantize_input_dtype:,\
+quantize_quantized_dtype:,\
+quantize_granularity:,\
+quantize_min_percentile:,\
+quantize_max_percentile:,\
+quantize_mode:,\
+quantize_input_data:,\
+quantize_output_path:,\
+`# one-pack`\
+pack_output_path:"
 
 OPTS=$(getopt --options "$SHORT_OPTS" --longoptions "$LONG_OPTS" --name "$0" -- "$@")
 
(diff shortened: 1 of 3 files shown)


seanshpark:
A. I think we need to consider for upcoming tasks:
- import TF saved model, keras model: no need of input/output arrays and shapes
- import ONNX: also no need of input/output arrays and shapes

but, I think there's no critical issues for these
please simulate about what changes are needed to the codes

B. Using flat `.cfg`
- not organized/structured so it's a bit hard to distinguish arguments that are depend on
   - we may use comments to separate visually
   - not sure this will be good for users and problem solving in the future.
- as @lemmaa suggested using `INI` can solve this issue; or using json format.

compiler/one-cmds/one-build
```diff
@@ -0,0 +1,165 @@
...
+
+# test if getopt is enhanced version
+getopt --test > /dev/null
+if [ $? -ne 4 ]; then
+  echo "[ERROR] Your system doesn't have enhanced getopt"
+  exit 2
+fi
+
+# usage
+function Usage()
```
glistening, line 27: @mhs4670go (optional) without `function`, it become more short and portable.

hyunsik-yoon:
It would be much better if `-h` option shows _what this script is for_ (purpose of this script). :-)
(At a glance of this PR title, I thought that it is some kinda of build script for ONE (compiler + runtime) for CI and if `-h` provides such info, it will be better. :-)

mhs4670go:
@hyunsik-yoon Well, I think so. We can rename it to `one-compile`.

seanshpark:
Leaving an idea from short off-line discussion

*How about providing several templates like `one-build.cfg` for each format?*
- There are items that must be filled and others left alone (don't care)
   - users may not know which item to fill
- Instead, we provide template of cfg for each formats(TF, BCQ, TFLITE and later TF/SavedModel, TF/Keras, ONNX)
   - users should fill for each items
   - but, still there are lots of common (optimize, quantize... ) optionals

seanshpark:
If may confuse by the this PR title but this tool (one-build) will be installed as a command line.

seanshpark:
> I thought that it is some kinda of build script for ONE (compiler + runtime) for CI 

@hyunsik-yoon , the file is `compiler/one-cmds/one-build` not in `tools/*` or such.

compiler/one-cmds/one-build.cfg
```diff
@@ -0,0 +1,79 @@
...
+output_path="inception_v3.circle"
+
+# driver to use
+use_one_import_tf="true"
+use_one_import_bcq="false"
+use_one_import_tflite="false"
+use_one_optimize="true"
+use_one_quantize="false"
+use_one_pack="true"
+use_one_codegen="false"
```
mhs4670go, line 21: I reordered `use` options to make it more readable. Users first select what driver to use and then fill the rest of other options. If users miss some options, the script occurs an error.

compiler/one-cmds/one-build.cfg
```diff
@@ -0,0 +1,79 @@
...
+
+### one-import-tf ### one-import-bcq ###
```
seanshpark, line 23: Would be be better if add a separate line like this?
```
# ----------------------------------------------------------------------------
### one-import-tf ### one-import-bcq ###
```

compiler/one-cmds/one-build.cfg
```diff
@@ -0,0 +1,79 @@
...
+optimize_resolve_customop_batchmatmul="false"
+# resolve_customop_matmul       Enable ResolveCustomOpMatMulPass Pass
+optimize_resolve_customop_matmul="false"
+# optimize_output_path          (Optional) Path to intermediate output of one-optimize
+optimize_optimize_output_path=""
+
+### one-quantize ###
+# input_dtype                   Input data type (supported: float32, default=float32)
+quantize_input_dtype=""
+# quantized_dtype               Output quantized data type (supported: uint8, default=uint8)
```
seanshpark, line 60: We now surpport `int16`

compiler/one-cmds/one-build
```diff
@@ -0,0 +1,165 @@
...
+  echo "See \"one-build.cfg\" for more details."
+  echo ""
+  echo "[OPTIONS TO OVERWRITE]"
+  echo "Sometimes you want to change certain options without touching the configuration file."
+  echo "If you pass the option directly to the command line, the option is processed prior to "
+  echo "the configuration file. A list of option names can be found in \"one-build.cfg\"."
+}
+
+# version
+DRIVER_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
seanshpark, line 43: (optional) maybe move `DRIVER_PATH` above `# version` ?

mhs4670go:
@seanshpark You can use `one-build` as of now:) Maybe needs to do more tests.

mhs4670go:
I'll post a new PR because `one-build` design will be changed a little bit.

PR #4511, opened 2020-10-06
[luci] Check more validation for BCQ information

llFreetimell:
Parent Issue : #4337
Full Draft : #4338

This commit will add more validation for BCQ information

### DIFF
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -267,12 +267,123 @@ private:
 
   bool is_bcqinfo_valid()
   {
-    // do_w_x should be int32 or bool type
+    std::set<int32_t> ref_prefix_set;
+
     for (auto n : _do_w_x)
     {
-      if (n.second->dtype() != loco::DataType::BOOL && n.second->dtype() != loco::DataType::S32)
+      // do_w_x should be BOOL type
+      if (n.second->dtype() != loco::DataType::BOOL)
+      {
+        LOGGER(l);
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -267,13 +267,124 @@ template <> class BCQFuser<1>
...
+      {
+        LOGGER(l);
+        WARN(l) << "FuseBCQPass : alpha has wrong type" << std::endl;
+        return false;
+      }
+
+      if (ref_prefix_set.find(n.first) == ref_prefix_set.end())
+      {
+        LOGGER(l);
+        WARN(l) << "FuseBCQPass : Invalid prefix is found" << std::endl;
```
seanshpark, line 299: Q) from the condition, is the meaning like `Required prefix not found. Invalid file` ?

llFreetimell: At first, I wanted to check that `prefix` is valid when we access to `map`.
However, it looks weird to add the assertion in this function.
I will remove this check in this PR and create another PR for it :)

seanshpark (approved):
LGTM thank you

mhs4670go:
@llFreetimell Just curiosity, I think the pass spits out WARNING instead of error because it shouldn't be stopped even if there are some mismatch with what it should be. But if there are some wrong condition, the converted model will be wrong model. Is it worth to generate it instead of throwing an error?

llFreetimell:
@mhs4670go Answered at offline :)

PR #4512, opened 2020-10-07
[onert] Fix a bug with consecutive calls of nnfw_run()

hyunsik-yoon:
This fix a bug with consecutive calls of nnfw_run() when 1st call uses
static input tensor but dynamnic input tensor is used from the second call.

Related issue [here](https://github.sec.samsung.net/STAR/nnfw/issues/11748).

### DIFF
--- a/runtime/onert/core/src/exec/LinearExecutor.cc
+++ b/runtime/onert/core/src/exec/LinearExecutor.cc
@@ -51,8 +51,10 @@ void LinearExecutor::executeImpl()
     _subject.notifyJobBegin(this, op_seq, backend);
 
     auto &fn_seq = code.fn_seq;
-    bool handle_dynamic_tensor = op_seq->has_dynamic_tensor() || hasDynamicInput();
 
+    fn_seq->initRunning();
+
+    bool handle_dynamic_tensor = op_seq->has_dynamic_tensor() || hasDynamicInput();
     fn_seq->enableDynamicShapeInferer(handle_dynamic_tensor);
     fn_seq->run();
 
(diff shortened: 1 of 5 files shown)


wateret (dismissed):
LGTM

runtime/onert/core/src/exec/LinearExecutor.cc
```diff
@@ -51,8 +51,10 @@ void LinearExecutor::executeImpl()
     _subject.notifyJobBegin(this, op_seq, backend);
 
     auto &fn_seq = code.fn_seq;
-    bool handle_dynamic_tensor = op_seq->has_dynamic_tensor() || hasDynamicInput();
 
+    fn_seq->initRunning();
```
glistening, line 55: I am going to remove the overhead related to `dynamic_inferer` in fn_sequence like `dynamic_cast`. It is about 3~4% penalty. I would like to move this out of every func sequence run if possible.

hyunsik-yoon: You're correct. This is a quick fix and should be optimized later.
I will add `todo` into comment mentioning this and keep eye on this.
Thanks.

glistening (approved):
LGTM

wateret (approved):
LGTM

runtime/onert/core/src/exec/LinearExecutor.cc
glistening: > Please note fn_seq->run() is hot spot. 

I am sorry. I was confused.

initRunning() is not called for every function, but for every function sequence.

`FunctionSequence::run()` has a loop for iterating functions.
`dynamic_cast` is in the loop, and `initRunning` is out of this loop.
Thus, it is not big as I first thought.

chunseoklee (approved):

PR #4517, opened 2020-10-07
[luci] Fix static warning in FoldDequantizePass

llFreetimell:
This commit will fix static warning in `FoldDequantizePass`

### DIFF
--- a/compiler/luci/pass/src/FoldDequantizePass.cpp
+++ b/compiler/luci/pass/src/FoldDequantizePass.cpp
@@ -33,6 +33,9 @@ bool is_hybrid_kernel_supported(loco::Node *node)
 
 bool is_foldable_const(luci::CircleConst *node)
 {
+  if (node->quantparam() == nullptr)
+    return false;
+
   if (node->dtype() == loco::DataType::S8)
     return true;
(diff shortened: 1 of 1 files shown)


seanshpark (dismissed):
LGTM

compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -43,7 +46,10 @@ bool is_foldable_const(luci::CircleConst *node)
 
 luci::CircleConst *dequantized_const_node(luci::CircleConst *const_node)
 {
-  assert(const_node->quantparam() != nullptr);
+  if (const_node->quantparam() == nullptr)
+  {
+    throw std::runtime_error("Given constant node cannot be dequantized");
```
mhs4670go, line 51: How about adding some more information about why this node can't be dequantized,, like `the node doesn't have a quant-related parameter~`.

llFreetimell, line 51:
```suggestion
    throw std::runtime_error("Given constant node has no quantization parameter");
```

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #4519, opened 2020-10-07
[Draft] [onert] Support UnidirectionalSequenceLSTM

ragmani:
For issue #3851

Signed-off-by: ragmani <ragmani0216@gmail.com>

Issue #3851, opened 2020-08-18
[onert] Implement UNIDIRECTIONAL_SEQUENCE_LSTM on onert


hseok-oh:
Let's implement UNIDIRECTIONAL_SEQUENCE_LSTM operation on onert

ragmani:
Is it from the backend fee?

hseok-oh:
>Is it from any backend free?

CPU first.

ragmani:
Okay, I will do this.

ragmani:
This operation is the similar `LSTM` operation. `LSTM` performs a single time step, but `UNIDIRECTIONAL_SEQUENCE_LSTM` performs all time step in lstm. So these two operations has difference as follows:
- `LSTM` operation performs a single time step in lstm with an 2-D input. But `UNIDIRECTIONAL_SEQUENCE_LSTM` performs all time steps with dynamic input unrolled the along the time dimension. And the input is either time-major or batch-major.
- `LSTM` operation generates outputs with cell state. But `UNIDIRECTIONAL_SEQUENCE_LSTM` doesn't generate cell state because it is not necessary after performing lstm.

| | LSTM | UNIDIRECTIONAL_SEQUENCE_LSTM |
|--|:--:|:--:|
| Time steps | single step | all steps |
| Inputs | 2-D | 3-D with a dimension for max time steps |
| Outputs | with cell state | without cell state |

ragmani:
Now, onert does not support optional inputs of LSTM. So I will work on that in this task.

ragmani:
A tflite model with `UNIDIRECTIONAL_SEQUENCE_LSTM` for MNIST : [unidirectional_sequence_lstm.zip](https://github.com/Samsung/ONE/files/5347307/unidirectional_sequence_lstm.zip)

ragmani:
`UNIDIRECTIONAL_SEQUENCE_LSTM` have state inputs that can be constant, but haven't data.

ragmani:
Running above tflite model works well with #4519 .
```bash
$ BACKENDS="cpu" ./Product/armv7l-linux.debug/out/bin/nnpackage_run --nnpackage usl/unidirectional_sequence_lstm -w 1 -r 1 -d output.h5 -l usl/unidirectional_sequence_lstm/metadata/tc/input.h5
Package Filename usl/unidirectional_sequence_lstm
===================================
MODEL_LOAD   takes 17.498 ms
PREPARE      takes 86.187 ms
[...]

PR #4533, opened 2020-10-07
[onert] Introduces a function `exist` into TensorRegistry

ragmani:
For issue #3851
Draft PR #4519

This commit introduces a function `exist` that checks an operand exists in TensorRegistry as a tensor.

### DIFF
--- a/runtime/onert/core/include/backend/ITensorRegistry.h
+++ b/runtime/onert/core/include/backend/ITensorRegistry.h
@@ -59,6 +59,13 @@ struct ITensorRegistry
    * @return false if not supported
    */
   virtual bool setMigrantTensor(const ir::OperandIndex &, IPortableTensor *) { return false; }
+  /**
+   * @brief Check if a tensor with the given index exists
+   *
+   * @param[in] index Index of the tensor to be returned
+   * @return true if such entry exists otherwise false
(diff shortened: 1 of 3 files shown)


runtime/onert/core/include/backend/ITensorRegistry.h
```diff
@@ -59,6 +59,13 @@ struct ITensorRegistry
    * @return false if not supported
    */
   virtual bool setMigrantTensor(const ir::OperandIndex &, IPortableTensor *) { return false; }
+  /**
+   * @brief Get the tensor that is associated with the given index
```
hyunsik-yoon, line 63:
```suggestion
   * @brief Check if a tensor with the given index exists
```

wateret:
`getITensor` returns nullptr when it does not exist, would we still need this function? 😃

runtime/onert/core/include/backend/ITensorRegistry.h
ragmani: Thanks for your pointing out. I will fix it.

ragmani:
> `getITensor` returns nullptr when it does not exist, would we still need this function?

It would be replaced it. I close this pr.

PR #4535, opened 2020-10-07
[onert] Fix when an operand that is input & output

wateret:
Insert a Permute operation and an operand to make an input can be copied
to an output. This is necessary since our API lets users give different
buffer pointers even though a tensor is an input and an output at the
same time.

- Introduce OddOutputPass to handle when an output is also an input
- Fix controlflow::DynamicTensorManager bug
- Add restriction to `PermutationInsertionPass` so the inserted Permute
  op can never be gone
- Add two tests

### DIFF
--- a/runtime/onert/core/src/compiler/pass/OddOutputPass.h
+++ b/runtime/onert/core/src/compiler/pass/OddOutputPass.h
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


runtime/onert/core/src/compiler/pass/OddOutputPass.h
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * Create an operand and insert a Permute(copy) op between them. And change the output to be the
+ * newly generated operand.
+ *
+ * e.g.)
+ *
+ * ```
+ * ((#0 Input0 and also Output0))
+ * becomes
+ * ((#0 Input0)) -> [#0 Permute] -> ((#1 Output0))
```
wateret, line 54: Visual(?) example 😄

dayo09 (approved):
LGTM

wateret:
As it is urgent, I'm merging this now.

PR #4541, opened 2020-10-07
[luci-interpreter] Support S16 AveragePool2D and MaxPool2D ops

s-barannikov:
Reuse TFLite's implementation for S16 AveragePool2D and MaxPool2D ops.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/AveragePool2D.cpp
+++ b/compiler/luci-interpreter/src/kernels/AveragePool2D.cpp
@@ -18,6 +18,7 @@
 
 #include "kernels/Utils.h"
 
+#include <tensorflow/lite/kernels/internal/reference/integer_ops/pooling.h>
 #include <tensorflow/lite/kernels/internal/reference/pooling.h>
 
 #include <stdexcept>
@@ -61,11 +62,13 @@ void AveragePool2D::configure()
       computePadding(_params.stride_width, 1, input_width, _params.filter_width, output_width);
   if (input()->element_type() == DataType::U8)
   {
-    if (input()->scale() != output()->scale() || input()->zero_point() != output()->zero_point())
-    {
-      throw std::runtime_error(
-          "Quantization param for Input and output must be same(scale or zero-point)");
-    }
+    LUCI_INTERPRETER_CHECK(std::abs(output()->scale() - input()->scale()) <= 1.0e-6);
+    LUCI_INTERPRETER_CHECK(output()->zero_point() == input()->zero_point());
+  }
+  else if (input()->element_type() == DataType::S16)
+  {
+    LUCI_INTERPRETER_CHECK(std::abs(output()->scale() - input()->scale()) <= 1.0e-6);
+    LUCI_INTERPRETER_CHECK(input()->zero_point() == 0 && output()->zero_point() == 0);
   }
   output()->resize({batches, output_height, output_width, depth});
 }
@@ -79,6 +82,9 @@ void AveragePool2D::execute() const
       break;
     case DataType::U8:
       evalQuantized();
+      break;
+    case DataType::S16:
+      evalSInt16();
       break;
     default:
       throw std::runtime_error("Unsupported type.");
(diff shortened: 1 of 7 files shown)


s-barannikov:
Luckily, TFLite had these implemented already. Unfortunately, there are no optimized versions of the kerels.

compiler/luci-interpreter/src/kernels/AveragePool2D.cpp
```diff
@@ -61,11 +62,13 @@ void AveragePool2D::configure()
       computePadding(_params.stride_width, 1, input_width, _params.filter_width, output_width);
   if (input()->element_type() == DataType::U8)
   {
-    if (input()->scale() != output()->scale() || input()->zero_point() != output()->zero_point())
-    {
-      throw std::runtime_error(
-          "Quantization param for Input and output must be same(scale or zero-point)");
-    }
+    LUCI_INTERPRETER_CHECK(std::abs(output()->scale() - input()->scale()) <= 1.0e-6);
```
binarman, line 65: Where did you get 1.0e-6 delta?
Looks like input and output scales should be totally equal.

s-barannikov: This is what TFLite does.
I'm ok with equality comparison too.
@jinevening what do you think?

compiler/luci-interpreter/src/kernels/MaxPool2D.cpp
```diff
@@ -54,10 +55,15 @@ void MaxPool2D::configure()
       computePadding(_params.stride_width, 1, input_width, _params.filter_width, output_width);
 
   output()->resize({batches, output_height, output_width, depth});
-  if (input()->element_type() == DataType::U8 || input()->element_type() == DataType::S8)
+  if (input()->element_type() == DataType::U8)
   {
-    assert(input()->scale() == output()->scale());
-    assert(input()->zero_point() == output()->zero_point());
+    LUCI_INTERPRETER_CHECK(std::abs(output()->scale() - input()->scale()) <= 1.0e-6);
```
binarman, line 60: Here too

compiler/luci-interpreter/src/kernels/AveragePool2D.cpp
jinevening: > @jinevening what do you think?

I think it would be safe to follow the TFLite code because we are using their kernel. :)

FYI, our quantizer does not have such a restriction (input scale == ouput scale) for average_pool.

s-barannikov: > FYI, our quantizer does not have such a restriction (input scale == ouput scale) for average_pool.

Is it intentional?

s-barannikov: Just for the case, TFLite only checks that the scales are (almost) equal, it does not make use of them otherwise.

compiler/luci-interpreter/src/kernels/AveragePool2D.test.cpp
```diff
@@ -112,6 +112,38 @@ TEST(AveragePool2DTest, Uint8_1)
...
+  std::vector<float> input_data{
+      -4, -3, -2, -1, 0,  //
+      1,  2,  3,  4,  5,  //
+      6,  7,  8,  9,  10, //
+  };
+  std::vector<float> ref_output_data{
+      0, 1.5, //
+      4.5, 6, //
+  };
+  Tensor input_tensor = makeInputTensor<DataType::S16>(input_shape, 0.5, 0, input_data);
```
jinevening, line 128: `makeInputTensor` quantizes a tensor using `quantize` function in `TestUtils.h`. The `quantize` function quantizes a tensor using the full range of a given data type (i.e., from `std::numeric_limits<T>::lowest()` to `std::numeric_limits<T>::max()`), which works well for uint8 quantization.

However, int16 quantization uses a symmetric range (from `std::numeric_limits<T>::lowest() +1` to `std::numeric_limits<T>::max()`). So we should extend `quantize` function in `TestUtils.h` to support symmetric quantization.

s-barannikov: Good catch, thanks!

compiler/luci-interpreter/src/kernels/AveragePool2D.cpp
jinevening: > Is it intentional?

I have no idea. Some restrictions may be added in the future for performance, but the current instruction model does not have such a restriction.

> Just for the case, TFLite only checks that the scales are (almost) equal, it does not make use of them otherwise.

I think that the purpose of checking the equality of scales is to avoid 'rescaling'. It seems that TFLite kernel does not perform rescaling under the assumption that the scales of input and output are the same (or almost equal).

s-barannikov: > I have no idea. Some restrictions may be added in the future for performance, but the current instruction model does not have such a restriction.

Would you want to remove this restriction from interpreter too (the kernel will have to be reimplemented by hand), or is it OK to go as it is for now?

compiler/luci-interpreter/src/kernels/AveragePool2D.test.cpp
s-barannikov: Fixed

compiler/luci-interpreter/src/kernels/AveragePool2D.cpp
jinevening: I think we can go as is for now. :)

jinevening (approved):
LGTM

binarman (approved):
LGTM

compiler/luci-interpreter/src/kernels/AveragePool2D.cpp
s-barannikov: > I think we can go as is for now. :)

OK. I'm just a bit worried about using interpreter with models quantized with our quantizer tool.

ai-moiseev:
LGTM

ai-moiseev (approved):

ai-moiseev (approved):
LGTM

PR #4547, opened 2020-10-08
[onert] Fix when duplicated output tensors used

wateret:
just like the case when "an operand that is input & output", do the
same for an operand which is referred as a model output two or more
times.


wateret:

wateret:

hseok-oh (approved):

dayo09 (approved):
LGTM

PR #4548, opened 2020-10-08
[one-cmds] Introduce FoldDequantize in one-optimize

llFreetimell:
Parent issue : #4373 

This commit will introduce `FoldDequantizePass` in `one-optimize`

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

PR #4550, opened 2020-10-08
[one-cmds] Introduce one-build

mhs4670go:
This commit introduces one-build to one-cmds.

Related: #4278 
Draft: #4509 

### DIFF
--- a/compiler/one-cmds/one-build
+++ b/compiler/one-cmds/one-build
@@ -1,0 +1,536 @@
+#!/bin/bash
+
+# Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


seanshpark (approved):
LGTM

seanshpark:
@llFreetimell , would it required to add https://github.com/Samsung/ONE/pull/4548 options too to `one-build`?

compiler/one-cmds/one-build
```diff
@@ -0,0 +1,536 @@
...
+    *)
+      echo "[ERROR] getopt internal error"
+      exit 2
+      ;;
+  esac
+done
+
+# source configuration file
+if [ -e "$CONFIG_FILE" ]; then
+  source "$CONFIG_FILE"
```
jinevening, line 245: Executing the config file looks quite risky to me. If the config file does not have the form we want, it can do numerous unexpected behaviors. For example, if config file has content `rm -rf .`, one-build will erase all files in the current directory. Or it can download ransomware from web using `wget`.

This may not be acceptable by our release test team. I think it will be better to use a text format (json or ini) for config file and parse the file to get the options.

jinevening: FYI, here is a simple ini parser written in bash script.  https://gist.github.com/splaspood/1473761

llFreetimell:
@seanshpark It seems yes, we may add the `--fold_dequantize` option

mhs4670go:
`one-build` will change its design:)

PR #4551, opened 2020-10-08
[onert] fix typo in ShapeValidator.cc

dayo09:
- fix Reduce operator's log print


hseok-oh:
It seems better to remove log code in `ShapeValidator`. It's legacy code.

dayo09:
@hseok-oh I was wondering about the purpose of this code. In that case, I will make a new PR to remove them.

PR #4562, opened 2020-10-08
[onert] Move param and I/O size check

hseok-oh:
Move param value check and input/output size check from ShapeValidator to OperationValidator


hyunsik-yoon (approved):
LGTM

Issue #4339, opened 2020-09-17
[luci-interpreter] Support 16-bit int quantized operators


binarman:
We need to start support of 16-bit quantization, this issue related to interpreter support

First priority operators are:

Operator | Status
--- | ---
ADD | #4569
AVERAGE_POOL_2D | #4541
CONV_2D | #4590
DEPTHWISE_CONV | #4591
MAX_POOL_2D | #4541
MEAN | #4571
MUL | #4568
PRELU | #4570
RELU | #4572
TRANSPOSE_CONV | #4593

+cc @s-barannikov @seanshpark @jinevening @lucenticus @parjong

jinevening:
How are you going to support these operators? Are you going to manually implement them?

s-barannikov:
Some of them can be implemented using function from TFLite, at least partially. Others will have to be impemented manually.

binarman:
This task is done in the the form we planned it.

I know there were a discussion about channel-wise quantization support with @jinevening 
Do we have other issues we need to take into account?
What is the priority of related tasks (i.e. channel-wise quantization)?

jinevening:
> What is the priority of related tasks (i.e. channel-wise quantization)?

The target model for testing int16 quantization is FSRCNN.

FSRCNN has depthwise_conv, conv, transpose_conv, so supporting CWQ kernels for them would be high priority.

After that, we may test if luci-interperter can run int16-quantized FSRCNN. :)

binarman:
@jinevening I created separate issue for this activity, closing this one

PR #4568, opened 2020-10-08
[luci-interpreter] Support S16 Mul op

s-barannikov:
Issue #4339.

Add support for S16 `Mul` operator.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Mul.cpp
+++ b/compiler/luci-interpreter/src/kernels/Mul.cpp
@@ -17,10 +17,10 @@
 
 #include "kernels/Mul.h"
 
+#include "kernels/BinaryOpCommon.h"
 #include "kernels/Utils.h"
 
 #include <tensorflow/lite/kernels/internal/optimized/optimized_ops.h>
-#include <tensorflow/lite/kernels/internal/reference/process_broadcast_shapes.h>
 
 #include <stdexcept>
 
@@ -36,7 +36,13 @@ Mul::Mul(const Tensor *input1, const Tensor *input2, Tensor *output, const MulPa
 
 void Mul::configure()
 {
-  assert(input1()->element_type() == input2()->element_type());
+  LUCI_INTERPRETER_CHECK(input1()->element_type() == input2()->element_type());
+  if (input1()->element_type() == DataType::S16)
+  {
+    LUCI_INTERPRETER_CHECK(input1()->zero_point() == 0 && input2()->zero_point() == 0 &&
+                           output()->zero_point() == 0);
+  }
+
   output()->resize(calculateShapeForBroadcast(input1()->shape(), input2()->shape()));
 }
 
@@ -46,6 +52,9 @@ void Mul::execute() const
   {
     case DataType::FLOAT32:
       evalFloat();
+      break;
+    case DataType::S16:
+      evalQuantizedS16();
       break;
     default:
       throw std::runtime_error("Unsupported type.");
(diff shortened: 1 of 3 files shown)


Hyunjun85 (approved):

compiler/luci-interpreter/src/kernels/Mul.test.cpp
```diff
@@ -78,6 +78,68 @@ TEST(MulTest, Float)
...
+    params.activation = Activation::RELU;
+
+    Mul kernel(&input1_tensor, &input2_tensor, &output_tensor, params);
+    kernel.configure();
+    kernel.execute();
+
+    EXPECT_THAT(extractTensorShape(output_tensor),
+                ::testing::ElementsAreArray(ref_output_shapes[i]))
+        << "With shape number " << i;
+    EXPECT_THAT(dequantizeTensorData(output_tensor), FloatArrayNear(ref_outputs[i], 0.001))
```
jinevening, line 117: I found the below code in TFLite.

```cpp
// For quantized Mul, the error shouldn't exceed (2*step + step^2).
...
const float kQuantizedStepInt16 = 2.0 / 32767.0;
const float kQuantizedToleranceInt16 =
    2.0 * kQuantizedStepInt16 + kQuantizedStepInt16 * kQuantizedStepInt16
```

How about using `2.0 * kQuantizedStepInt16 + kQuantizedStepInt16 * kQuantizedStepInt16` instead of `0.001` ?

s-barannikov: I just don't understand the formula :) I can also see another one in `TEST(QuantizedMulOpTest, NoActivationInt16Scaled)`, which is:
```C++
  const float kQuantizedToleranceInt16Scaled =
      6.0 * kQuantizedStepInt16 + kQuantizedStepInt16 * kQuantizedStepInt16;
```

Also, TFLite has its own way of quantizing input tensors, so I'm not sure these formulae will fit anyway.

But I think I can adjust scales to improve precision.

s-barannikov: @jinevening PTAL, does it looks better now? If it's OK, I will apply the same concept to `Add` in sibling PR.

jinevening: > Also, TFLite has its own way of quantizing input tensors, so I'm not sure these formulae will fit anyway.

I agree.

One heuristic we've used in uint8 was to use one step (=scale) of quantized values as the tolerance. In this example, the scale of output is 0.01, and tolerance is 1e-5 (in `FloatArrayNear`). So I think it is ok to use the current values.

(optional) Smaller scale would be better for testing precision (previous scales look better to me). How about using the previous scales and setting the tolerance to `n X scale of output tensor` (maybe `n=1` as uint8 quant)? Showing the relationship between scale and tolerance would improve readability.

compiler/luci-interpreter/src/kernels/Mul.cpp
```diff
@@ -36,7 +36,13 @@ Mul::Mul(const Tensor *input1, const Tensor *input2, Tensor *output, const MulPa
 
 void Mul::configure()
 {
-  assert(input1()->element_type() == input2()->element_type());
+  LUCI_INTERPRETER_CHECK(input1()->element_type() == input2()->element_type());
+  if (input1()->element_type() == DataType::S16)
+  {
+    LUCI_INTERPRETER_CHECK(input1()->zero_point() == 0 && input2()->zero_point() == 0 &&
```
binarman, line 42: Should output quantization be same as input? I mean is operators (S16, S16) -> U8 valid?
Or this is checked somewhere else?

binarman (commented):
LGTM

compiler/luci-interpreter/src/kernels/Mul.cpp
s-barannikov: No, I'll add the check.

compiler/luci-interpreter/src/kernels/Mul.test.cpp
s-barannikov: > (optional) Smaller scale would be better for testing precision (previous scales look better to me). How about using the previous scales and setting the tolerance to n X scale of output tensor (maybe n=1 as uint8 quant)? Showing the relationship between scale and tolerance would improve readability.

OK, I'll try.

compiler/luci-interpreter/src/kernels/Mul.cpp
s-barannikov: Fixed.

binarman (approved):
LGTM

compiler/luci-interpreter/src/kernels/Mul.test.cpp
s-barannikov: Done, PTAL

jinevening (approved):
LGTM

s-barannikov:
@Hyunjun85 PTAL

Hyunjun85 (approved):
LGTM

PR #4570, opened 2020-10-08
[luci-interpreter] Support S16 Prelu op

s-barannikov:
Issue #4339.

Add support for S16 `Prelu` operator.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Prelu.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Prelu.test.cpp
@@ -164,6 +164,52 @@ TEST(PreluTest, Uint8Broadcast)
               ::testing::ElementsAreArray(ref_quant_output_data));
 }
 
+TEST(PreluTest, SInt16Simple)
+{
+  std::vector<float> input_data{-0.8f, 0.2f, 0.9f, 0.7f, 0.1f, -0.4f};
+  std::vector<float> alpha_data{0.5f, 0.5f, 0.5f, 0.25f, 1.0f, 0.25f};
+  std::vector<float> ref_output_data{-0.4f, 0.2f, 0.9f, 0.7f, 0.1f, -0.1f};
+
+  Tensor input_tensor = makeInputTensor<DataType::S16>({1, 2, 3, 1}, 0.1, 0, input_data);
+  Tensor alpha_tensor = makeInputTensor<DataType::S16>({1, 2, 3, 1}, 0.1, 0, alpha_data);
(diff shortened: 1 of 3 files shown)


compiler/luci-interpreter/src/kernels/Prelu.test.cpp
```diff
@@ -164,6 +164,52 @@ TEST(PreluTest, Uint8Broadcast)
...
+  Tensor input_tensor = makeInputTensor<DataType::S16>({1, 2, 3, 1}, 0.1, 0, input_data);
+  Tensor alpha_tensor = makeInputTensor<DataType::S16>({1, 2, 3, 1}, 0.1, 0, alpha_data);
+  Tensor output_tensor = makeOutputTensor(DataType::S16, 0.1, 0);
+
+  Prelu kernel(&input_tensor, &alpha_tensor, &output_tensor);
+  kernel.configure();
+  kernel.execute();
+
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray({1, 2, 3, 1}));
+  EXPECT_THAT(dequantizeTensorData(output_tensor), FloatArrayNear(ref_output_data));
```
jinevening, line 182: How about using quantized tolerance as uint8 kernel ?

Code from `activations_test.cc`  in TFLite.
```cpp
const float kQuantizedToleranceInt16 = 2 * (1. / 4096);
```

jinevening: The rest looks good to me.

s-barannikov: > Code from `activations_test.cc`  in TFLite.

Thanks. It seems too rought though and I don't understand where these numbers came from.
I'd rather use output scale as a tolerance if current 1.0e-5 is too narrow.

s-barannikov: > I'd rather use output scale as a tolerance if you think the current 1.0e-5 is too narrow.

@jinevening Would that be OK to you?

jinevening: > @jinevening Would that be OK to you?

Yes, it is ok to me. Actually, the current one is also ok because 1.0e-5 is smaller than output scale (0.1).

jinevening (approved):

ai-moiseev (approved):

binarman (approved):
LGTM

PR #4571, opened 2020-10-08
[luci-interpreter] Support S16 Mean op

s-barannikov:
Issue #4339.

Add support for S16 `Mean` operator (special case of reduction over H & W axes of NHWC tensor).

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Mean.cpp
+++ b/compiler/luci-interpreter/src/kernels/Mean.cpp
@@ -130,8 +130,13 @@ Mean::Mean(const Tensor *input, const Tensor *axes, Tensor *output, const Reduce
 
 void Mean::configure()
 {
-  assert(input()->element_type() == output()->element_type());
-  assert(axes()->element_type() == DataType::S32);
+  LUCI_INTERPRETER_CHECK(input()->element_type() == output()->element_type());
+  LUCI_INTERPRETER_CHECK(axes()->element_type() == DataType::S32);
+  if (input()->element_type() == DataType::S16)
+  {
+    LUCI_INTERPRETER_CHECK(input()->zero_point() == 0 && output()->zero_point() == 0);
+  }
+
   const Shape &input_shape = input()->shape();
   int input_num_dims = input_shape.num_dims();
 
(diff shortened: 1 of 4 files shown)


compiler/luci-interpreter/src/kernels/Mean.cpp
```diff
@@ -169,6 +174,9 @@ void Mean::execute() const
     case DataType::U8:
       evalQuantized();
       break;
+    case DataType::S16:
+      evalSInt16();
```
binarman, line 178: I think we need to stick to one naming convention to avoid ambiguity.
I mean this is quantized computations too, so It would be better to name eval functions like
`evalQuantizedU8` and `evalQuantizedS16`

compiler/luci-interpreter/src/kernels/Mean.cpp
```diff
@@ -245,5 +253,74 @@ void Mean::evalQuantized() const
...
+  const auto *input_data = getTensorData<int16_t>(input());
+  auto *output_data = getTensorData<int16_t>(output());
+
+  const Shape &input_shape = input()->shape();
+  const Shape &output_shape = output()->shape();
+
+  const auto *axes_data = getTensorData<int32_t>(axes());
+  const int num_axes = axes()->shape().num_elements();
+
+  constexpr int32_t output_min = -std::numeric_limits<int16_t>::max();
```
binarman, line 267: nitpick.
If this should be particular value maybe add it explicitly?

s-barannikov: It's easier to make a mistake with explicit numbers.

compiler/luci-interpreter/src/kernels/Mean.cpp
s-barannikov: Fixed.

binarman (dismissed):
LGTM

jinevening (dismissed):
LGTM

s-barannikov:
@binarman @jinevening I've resolved conflicts, please reapprove.

jinevening (approved):

PR #4583, opened 2020-10-12
[onert/loader] Check if subgraph index is valid

wateret:
- For If and While operation, check if subgraph index values are valid
- Add 5 negative tests for each operation

### DIFF
--- a/runtime/onert/frontend/base_loader/include/base_loader.h
+++ b/runtime/onert/frontend/base_loader/include/base_loader.h
@@ -157,6 +157,14 @@ private:
   void loadSpaceToDepth(const Operator *op, ir::Graph &subg);
   void loadLeakyRelu(const Operator *op, ir::Graph &subg);
 
+  void verifySubgraphIndex(int subg_index)
+  {
+    const auto num_subgraphs = _model->subgraphs()->size();
+    if (subg_index < 0 || subg_index >= static_cast<int32_t>(num_subgraphs))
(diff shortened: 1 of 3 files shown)


runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -157,6 +157,14 @@ template <typename LoaderDomain> class BaseLoader
   void loadSpaceToDepth(const Operator *op, ir::Graph &subg);
   void loadLeakyRelu(const Operator *op, ir::Graph &subg);
 
+  void verifySubgraphIndex(int subg_index)
```
hseok-oh, line 160: Looks better to check on `OperationValidator`, IMO.

ragmani: `OperationValidator` cannot access other subgraphs. Anyway I also think that it would be better to verify this in core of `onert`.

wateret: @hseok-oh Maybe we should distinguish which is checked from model-loading or from compilation.

hseok-oh: If we need to check on model-loading, we can move calling `OperationValidator` from `Compiler::compiler()` to `Graph::finishBuilding()`

hseok-oh (approved):

ragmani (approved):

PR #4584, opened 2020-10-12
[onert] Add InputAndOutput GenModelTest

wateret:
Add a TC for when an operand is both Input and Output and is used by an
operation.


wateret:
@llFreetimell and I had a talk about this TC.

llFreetimell (approved):
LGTM :)

PR #4586, opened 2020-10-12
[one-cmds] Add missing declaration

llFreetimell:
Missing declaration causes `unary operator expected` error.
This commit will fix it.


seanshpark (approved):
LGTM

PR #4588, opened 2020-10-12
[mio-tf] Use tensorflow 2.3 source code

hseok-oh:
Use tensorflow 2.3 source code to refer protobuf specification

### DIFF
--- a/compiler/mio-tf/CMakeLists.txt
+++ b/compiler/mio-tf/CMakeLists.txt
@@ -1,6 +1,6 @@
 nnas_find_package(Protobuf QUIET)
 # TensorFlowSource package is used to use ~.proto files
-nnas_find_package(TensorFlowSource EXACT 1.12 QUIET)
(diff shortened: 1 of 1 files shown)


compiler/mio-tf/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 nnas_find_package(Protobuf QUIET)
 # TensorFlowSource package is used to use ~.proto files
-nnas_find_package(TensorFlowSource EXACT 1.12 QUIET)
+nnas_find_package(TensorFlowSource EXACT 2.3 QUIET)
```
seanshpark, line 3: Currently mio-tf is not active project and not sure changing to 2.3 is valid. (I assume this is to remove 1.12 source related things)
I may think future me or others should do the validation :)

seanshpark (approved):
LGTM

PR #4591, opened 2020-10-12
[luci-interpreter] Support S16 DepthwiseConv2D op

s-barannikov:
Issue #4339.

Add support for S16 `DepthwiseConv2D` operator.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
+++ b/compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
@@ -45,7 +45,11 @@ void DepthwiseConv2D::configure()
   // (4) | int8  int8   int32 int8   | quantized per channel
   // (5) | int16 int8   int64 int16  | quantized per channel 16x8
   //
-  // We only support (1) and (3) for now.
+  // We only support (1) and (3) for now, and additionally the following:
+  //     | input filter bias  output |
+  // ----+---------------------------+
+  // (5) | int16 int16  int64 int16  |
+  //
   if (input()->element_type() == DataType::FLOAT32 && filter()->element_type() == DataType::FLOAT32)
   {
     LUCI_INTERPRETER_CHECK(bias() == nullptr || bias()->element_type() == DataType::FLOAT32);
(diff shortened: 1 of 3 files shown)


compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
```diff
@@ -171,5 +182,91 @@ void DepthwiseConv2D::evalQuantized() const
...
+  const int32_t output_width = output_shape.dim(2);
+
+  const int32_t stride_height = _params.stride_height;
+  const int32_t stride_width = _params.stride_width;
+  const int32_t dilation_height_factor = _params.dilation_height_factor;
+  const int32_t dilation_width_factor = _params.dilation_width_factor;
+  const int32_t depth_multiplier = _params.depth_multiplier;
+
+  const double effective_output_scale =
+      getQuantizedConvolutionMultipler(input()->scale(), filter()->scale(), output()->scale());
```
jinevening, line 212: This assumes per-layer quantization. Can you add a kernel for per-channel quantization?

If we have a per-channel kernel, per-layer kernel may be easily replaced with the per-channel kernel (by calling per-channel kernel with the same scales/zero_points for all channels).

jinevening: FYI, we can check whether dw_conv is quantized per-channel or not by checking if the number of scales/zero_points of a filter is greater than 1.

s-barannikov: > This assumes per-layer quantization. Can you add a kernel for per-channel quantization?

There are no kernels with per-channel quantization support yet (even U8).
Per-channel quantization support should be more or less straightforward, but it is still a big work. It also does not seem that urgent, so I'd like to proceed as it is for now, if you don't mind.

jinevening: > It also does not seem that urgent, so I'd like to proceed as it is for now, if you don't mind.

I don't mind :) This is not urgent as you said.

jinevening (approved):
LGTM

compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
```diff
@@ -45,7 +45,11 @@ void DepthwiseConv2D::configure()
   // (4) | int8  int8   int32 int8   | quantized per channel
   // (5) | int16 int8   int64 int16  | quantized per channel 16x8
   //
-  // We only support (1) and (3) for now.
+  // We only support (1) and (3) for now, and additionally the following:
+  //     | input filter bias  output |
+  // ----+---------------------------+
+  // (5) | int16 int16  int64 int16  |
```
binarman, line 51: nitpick: 5-th item already in the list )

binarman (approved):
LGTM

compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
s-barannikov: Just copied from conv2d =\

PR #4592, opened 2020-10-12
[luci-interpreter] Add TransposeConv U8 test

s-barannikov:
* Add a test for U8 quantized `TransposeConv`.
* Fix found errors.
* Move padding evaluation into `configure` method.

Signed-off-by: Sergei Barannikov <s.barannikov@samsung.com>

### DIFF
--- a/compiler/luci-interpreter/src/kernels/TransposeConv.cpp
+++ b/compiler/luci-interpreter/src/kernels/TransposeConv.cpp
@@ -43,18 +43,6 @@ void TransposeConv::configure()
   assert(input()->element_type() == DataType::FLOAT32 || input()->element_type() == DataType::U8);
   assert(input()->element_type() == output()->element_type());
   assert(input()->shape().dim(3) == filter()->shape().dim(3));
-  if (input()->element_type() == DataType::U8)
-  {
-    _scratch_tensor =
-        std::make_unique<Tensor>(DataType::S32, output()->shape(), AffineQuantization{}, "");
-    double real_multiplier = 0.0;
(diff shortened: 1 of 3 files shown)


ai-moiseev (approved):
LGTM

compiler/luci-interpreter/src/kernels/TransposeConv.cpp
```diff
@@ -43,25 +43,38 @@ void TransposeConv::configure()
...
+      computePadding(params().stride_height, 1, output_height, filter_height, unused_output_height);
+  _padding_width =
+      computePadding(params().stride_width, 1, output_width, filter_width, unused_output_width);
+
+  if (input()->element_type() == DataType::U8)
+  {
+    _scratch_tensor =
+        std::make_unique<Tensor>(DataType::S32, output()->shape(), AffineQuantization{}, "");
+    const double input_product_scale = input()->scale() * filter()->scale();
+    assert(input_product_scale >= 0);
```
struss, line 74: What about using `LUCI_INTERPERTER_CHECK` on this?

struss (approved):
LGTM

compiler/luci-interpreter/src/kernels/TransposeConv.cpp
s-barannikov: This is a bit out of scope of this PR. I think it can be done later. We also should fix Conv2D and DepthwiseConv2D the similar way, or extract this into a function.

binarman (approved):
LGTM

PR #4593, opened 2020-10-12
[luci-interpreter] Support S16 TransposeConv op

s-barannikov:
Issue #4339.

Add support for S16 `TransposeConv` operator.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/TransposeConv.cpp
+++ b/compiler/luci-interpreter/src/kernels/TransposeConv.cpp
@@ -40,7 +40,8 @@ void TransposeConv::configure()
   assert(output_shape()->shape().num_dims() == 1);
   assert(input()->shape().num_dims() == 4);
   assert(filter()->shape().num_dims() == 4);
-  assert(input()->element_type() == DataType::FLOAT32 || input()->element_type() == DataType::U8);
+  assert(input()->element_type() == DataType::FLOAT32 || input()->element_type() == DataType::U8 ||
+         input()->element_type() == DataType::S16);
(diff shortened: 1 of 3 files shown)


jinevening (approved):
LGTM

compiler/luci-interpreter/src/kernels/TransposeConv.cpp
```diff
@@ -135,5 +141,93 @@ void TransposeConv::evalQuantized() const
...
+      {
+        for (int32_t out_c = 0; out_c < output_depth; ++out_c)
+        {
+          int64_t acc = scratch_data[calcOffset(output_shape, batch, out_y, out_x, out_c)];
+          if (bias_data)
+          {
+            acc += bias_data[out_c];
+          }
+          int32_t scaled_acc =
+              tflite::MultiplyByQuantizedMultiplier(acc, _output_multiplier, _output_shift);
```
jinevening, line 220: `_output_multiplier` is a scalar value, which means that TCONV is quantized per-tensor. It will be better to change `_output_multiplier` to a vector so that we can support per-channel quantization.

s-barannikov: @jinevening Please see my comment in the other PR https://github.com/Samsung/ONE/pull/4591#discussion_r505125806. Modifying the kernel to support per-channel quantization does not seem difficult, but testing infrastruture is not ready for this yet. There are no U8 kernels with per-channel support, too. IMHO per-channel quantization support should be a separate task.

binarman (approved):
LGTM

PR #4595, opened 2020-10-13
[onnxkit] Use ONNX 1.6.0

seanshpark:
This will revise onnxkit to use ONNX version 1.6.0


parjong (approved):

llFreetimell:
@seanshpark It seems that ONNX v1.7.0 which was released at May is the latest version. Is there any reason for not using the latest version?

seanshpark:
> Is there any reason for not using the latest version?

There are 1.4.1 and 1.6.0 used in our code and this is jus to change to use one version, 1.6.0.
I didn't know about 1.7.0. Maybe later we can upgrade to latest version.

llFreetimell (approved):
LGTM

PR #4597, opened 2020-10-13
[infra/cmd] Introduce user mode nnas docker-run

hseok-oh:
Introduce user mode docker-run to reduce ownership change overhead
User mode run may fail if command require home directory or/and username (ex. gbs build)


seanshpark (approved):
LGTM

PR #4598, opened 2020-10-13
[onert] Replace dyn_cast with polymorphic_downcast

wateret:
Replace dyn_cast with polymorphic_downcast in DynamicTensorManager as
this code always assumes that the downcast is successful. And for some
models this part could be performance-critical as it is run during
inference execution.


hyunsik-yoon (approved):
LGTM

PR #4603, opened 2020-10-13
[onert] Support neon kernel in reduceSum

YiHyunJin:
When aixe is -1, it support neon kernel.
The sum is implemented using 'vaddq_f32'.
And tested any models. The result is an 4.5% increase in speed.
(1.148 -> 1.099)


glistening:
0.889 → 0.867

glistening (approved):
LGTM

Issue #4544, opened 2020-10-08
CPU backend getTensorShape overhead


hseok-oh:
@wateret (beyond this PR)

Please note that `getTensorShape` overhead is significant for models which has a lot of op, no 1~2 dominant kernel. It would be good if we can remove the overhead of `getTensorShape`. I think the complex hierarchy hinders the compiler's optimization. Do you have any idea or suggestion?

_Originally posted by @glistening in https://github.com/Samsung/ONE/pull/4538#issuecomment-705253011_

wateret:
It does not seem like there is a way doing it in `configure`. So I think reducing `getTensorShape` overhead is pretty much all I can come up with. To do so, we may define a raw pointer protocol for shape info.

Let me give a rough example,

Define a raw pointer format. A simple vector, size(rank) comes first. Say we have a pointer `ptr` with size (4 * (RANK+1) bytes).

- Address `ptr+0` : RANK (4 bytes)
- Address `ptr+4` : DIMENSION VALUE ARRAY (4 * RANK bytes)

Then introduce `void* ITensor::getRawShape()`. And make `cker::Shape` just a wrapper of this pointer.

Plus, it might be further optimized if the pointer(`void *`) could be the object itself and use `reinterpret_cast`.

krayzemli:
Since we know that cpu backend always gets backend::cpu_common::Tensor as its input, we could make a static cast for each tensor first and then get a constant reference to its _info field with a special Tensor-specific method, and use it directly. Unificating ir::Shape and cker::Shape the conversion overhead may be reduced even more.

I don't get the reason why our ITensor interface is so abstract, with an individual getter for every thing, when we could just keep some general tensor descriptor structure inside common for all tensor types and return a constant reference to it with a single call, probably even non-virtual. That is an approach used in MNN, for example.

krayzemli:
Besides, getTensorShape is not the only problem which introduces tensor size specific preprocessing overhead. Since dynamic tensors are not common for neural networks, we better make some onTensorInfoChanged method in addition to configure() and make all necessary preprocessing there, including caching necessary tensor shapes. When tensor shape, data type, layout, or something like that changes, some mechanism would trigger onTensorInfoChanged method call of an affected operator before calling run()

wateret:
@krayzemli 

[...]

PR #4611, opened 2020-10-13
[onert] static tensor info caching in BinaryArithmeticLayer

krayzemli:
Keep tensor shapes and derived broadcast parameters in binary arithmetic CPU ops implementation, and reuse them when all tensors turned out to be static tensors.

This change brings 5%+ improvement on you-know-what oversized-graph network.

It relies on shape inference logic always marking output tensor of an op as dynamic once any of its input tensors is dynamic. **Please verify if it is correct before merging.**

### DIFF
--- a/compute/cker/include/cker/Shape.h
+++ b/compute/cker/include/cker/Shape.h
@@ -136,6 +136,21 @@ public:
     std::memcpy(dst_dims, dims_data, dimensions_count * sizeof(int32_t));
   }
 
+  inline void ReplaceWith(const Shape &other)
+  {
+    ReplaceWith(other.DimensionsCount(), other.DimsData());
+  }
+
+  inline void ReplaceWith(Shape &&other)
+  {
+    Resize(0);
+    std::swap(_size, other._size);
+    if (_size <= kMaxSmallSize)
+      std::copy(other._dims, other._dims + kMaxSmallSize, _dims);
+    else
+      _dims_pointer = other._dims_pointer;
+  }
+
   template <typename T> inline void BuildFrom(const T &src_iterable)
   {
     const int dimensions_count = std::distance(src_iterable.begin(), src_iterable.end());
--- a/runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
+++ b/runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
@@ -56,8 +56,13 @@ template <nnfw::cker::BinaryArithmeticOpType arithmetic_type, typename T> struct
 
   void operator()(const IPortableTensor *lhs, const IPortableTensor *rhs, IPortableTensor *output)
   {
+    // Assume dynamic tensors never become static and static ones never change shape since
+    // configure()
     if (output->is_dynamic())
       updateCache(lhs, rhs, output);
+    else
+      assert(_lhs_shape == getTensorShape(lhs) && _rhs_shape == getTensorShape(rhs) &&
+             _output_shape == getTensorShape(output));
     auto lhs_buffer = reinterpret_cast<const T *>(lhs->buffer());
     auto rhs_buffer = reinterpret_cast<const T *>(rhs->buffer());
     auto output_buffer = reinterpret_cast<T *>(output->buffer());
--- a/runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
+++ b/runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
@@ -30,31 +30,54 @@ namespace ops
 namespace
 {
 
-template <nnfw::cker::BinaryArithmeticOpType arithmetic_type, typename T>
-void eval(const IPortableTensor *lhs, const IPortableTensor *rhs, IPortableTensor *output,
-          nnfw::cker::BinaryArithmeticOpParam op_params)
+template <nnfw::cker::BinaryArithmeticOpType arithmetic_type, typename T> struct Eval
 {
-  const auto lhs_shape = getTensorShape(lhs);
-  const auto rhs_shape = getTensorShape(rhs);
-  const bool need_broadcast = nnfw::cker::ProcessBroadcastShapes(lhs_shape, rhs_shape, &op_params);
-  if (need_broadcast)
+  nnfw::cker::Shape _lhs_shape;
+  nnfw::cker::Shape _rhs_shape;
+  nnfw::cker::Shape _output_shape;
+  nnfw::cker::BinaryArithmeticOpParam _op_params;
+  bool _need_broadcast;
+
+  Eval(const IPortableTensor *lhs, const IPortableTensor *rhs, IPortableTensor *output,
+       nnfw::cker::BinaryArithmeticOpParam op_params)
+      : _op_params(std::move(op_params))
   {
-    nnfw::cker::BroadcastBinaryArithmeticOp<arithmetic_type>(
-        op_params, lhs_shape, reinterpret_cast<const T *>(lhs->buffer()), rhs_shape,
-        reinterpret_cast<const T *>(rhs->buffer()), getTensorShape(output),
-        reinterpret_cast<T *>(output->buffer()));
-    return;
+    if (!output->is_dynamic())
+      updateCache(lhs, rhs, output);
   }
 
-  nnfw::cker::BinaryArithmeticOp<arithmetic_type>(
-      op_params, lhs_shape, reinterpret_cast<const T *>(lhs->buffer()), rhs_shape,
-      reinterpret_cast<const T *>(rhs->buffer()), getTensorShape(output),
-      reinterpret_cast<T *>(output->buffer()));
-}
+  void updateCache(const IPortableTensor *lhs, const IPortableTensor *rhs, IPortableTensor *output)
+  {
+    _lhs_shape.ReplaceWith(getTensorShape(lhs));
+    _rhs_shape.ReplaceWith(getTensorShape(rhs));
+    _output_shape.ReplaceWith(getTensorShape(output));
+    _need_broadcast = nnfw::cker::ProcessBroadcastShapes(_lhs_shape, _rhs_shape, &_op_params);
+  }
+
+  void operator()(const IPortableTensor *lhs, const IPortableTensor *rhs, IPortableTensor *output)
+  {
+    if (output->is_dynamic())
+      updateCache(lhs, rhs, output);
+    auto lhs_buffer = reinterpret_cast<const T *>(lhs->buffer());
+    auto rhs_buffer = reinterpret_cast<const T *>(rhs->buffer());
+    auto output_buffer = reinterpret_cast<T *>(output->buffer());
+    if (_need_broadcast)
+    {
+      nnfw::cker::BroadcastBinaryArithmeticOp<arithmetic_type>(
+          _op_params, _lhs_shape, lhs_buffer, _rhs_shape, rhs_buffer, _output_shape, output_buffer);
+    }
+    else
+    {
+      nnfw::cker::BinaryArithmeticOp<arithmetic_type>(
+          _op_params, _lhs_shape, lhs_buffer, _rhs_shape, rhs_buffer, _output_shape, output_buffer);
+    }
+  }
+};
 
 template <nnfw::cker::BinaryArithmeticOpType arithmetic_type>
 std::function<void(const IPortableTensor *, const IPortableTensor *, IPortableTensor *)>
-generateKernelGeneric(const IPortableTensor *lhs, const ir::Activation activation,
+generateKernelGeneric(const IPortableTensor *lhs, const IPortableTensor *rhs,
+                      IPortableTensor *output, const ir::Activation activation,
                       nnfw::cker::BinaryArithmeticOpParam op_params)
 {
   switch (lhs->data_type())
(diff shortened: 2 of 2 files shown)


glistening:
cc @hyunsik-yoon, @wateret

runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
```diff
@@ -30,31 +30,54 @@ namespace ops
...
+  {
+    _lhs_shape.ReplaceWith(getTensorShape(lhs));
+    _rhs_shape.ReplaceWith(getTensorShape(rhs));
+    _output_shape.ReplaceWith(getTensorShape(output));
+    _need_broadcast = nnfw::cker::ProcessBroadcastShapes(_lhs_shape, _rhs_shape, &_op_params);
+  }
+
+  void operator()(const IPortableTensor *lhs, const IPortableTensor *rhs, IPortableTensor *output)
+  {
+    if (output->is_dynamic())
```
periannath, line 59: @hyunsik-yoon Is there any case where tensor is dynamic during `compile` and becomes static during `inference`? If that case occurs, `_lhs_shape` will be used without initialize.

glistening: According to https://github.com/Samsung/ONE/issues/4544#issuecomment-707460292, it can happen.

hyunsik-yoon: Correct. BTW we don't have such test case yet so I am writing the case.

hyunsik-yoon: Designing cache strategy for various cases will take time.

For simple start, how about not using (or updating) the cache if `output->is_dynamic()`?
In other words, when `output->is_dynamic()`, the code works just like the previous way (by calling `getTensorShape`).

I am not sure how many tensors you want to cache. If you eventually consider to expand the scope of cache to all tensors in a model, I tried to calculate an imaginary huge model where there is, e.g., 1000 operations and works with dynamic tensors (meaning hit ratio of cache for static tensor shape is low and simple caching strategy waste memory),
the cache size would be [ 2.5 tensors / op * 1000  op * 28Bytes (size of nnfw::cker::Shape) = 68KB ]

I guess that we can eventually provide complex cache strategy or provide some flag to invalidate caching for the imaginary model I mentioned.

hyunsik-yoon: for https://github.com/Samsung/ONE/pull/4611#discussion_r504403760

> Correct. BTW we don't have such test case yet so I am writing the case.

it will take some time. please refer to https://github.com/Samsung/ONE/issues/4625

periannath: With offline discussion with @hyunsik-yoon, I've noticed that caching tensor shape for dynamic output may slow down inference when model has dynamic tensor. I agree with his suggestion that removes `updateCache` when output is dynamic.

krayzemli: @hyunsik-yoon Can you be more specific with the scenario when dynamic tensor becomes static again? As I see from code, static shapes are inferenced in compile-time only, and the only other way of shape inference is the dynamic one which sets a new shape with applyShape method (at least for binary arithmetic op) which again always sets dynamic flag of an output.

krayzemli: @periannath This is more theoretic, as `updateCache` does things that are done anyway, the only slowdown is the single call to `is_dynamic()` and probably `cker::Shape` copy, but copy can be eliminated by making additional `getTensorShape` signature that takes `Shape` reference and fills existing object instead of creating a new one.

glistening: > @hyunsik-yoon Can you be more specific with the scenario when dynamic tensor becomes static again? As I see from code, static shapes are inferenced in compile-time only, and the only other way of shape inference is the dynamic one which sets a new shape with applyShape method (at least for binary arithmetic op) which again always sets dynamic flag of an output.

@krayzemli In current implementation, you may be right. However, there is no rule and much reason that we forbid dynamic shape to static shape.

compute/cker/include/cker/Shape.h
```diff
@@ -136,6 +136,21 @@ class Shape
...
+  }
+
+  inline void ReplaceWith(Shape &&other)
+  {
+    Resize(0);
+    std::swap(_size, other._size);
+    if (_size <= kMaxSmallSize)
+      std::copy(other._dims, other._dims + kMaxSmallSize, _dims);
+    else
+      _dims_pointer = other._dims_pointer;
```
binarman, line 151: remove else and replace  this line with 
```suggestion
    std::swap(_dims_pointer, other._dims_pointer);
```
current implementation could lead to:
- double free in case `self._size > kMaxSmallSIze` and `other._size > kMaxSmallSize` (sizes at call time) or segfault on accessing this array
- memory leak in case `self._size > kMaxSmallSIze` and `other._size <= kMaxSmallSize` (sizes at call time)

runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
binarman: @hyunsik-yoon
I want to clarify memory requirements:
From my perspective ~100 KB overhead is not a big deal in case of neural networks, I think it is a fare trade-off if it can improve performance. Are you agree or I misunderstand something?

hyunsik-yoon: @krayzemli

> Can you be more specific with the scenario when dynamic tensor becomes static again? 

It depends on the result of https://github.com/Samsung/ONE/issues/4625.

> This is only theoretic, as updateCache does things that are done anyway, the only slowdown is the single call to is_dynamic() and probably Shape::ReplaceWith(Shape&&) rvalue-copy, but this copy can be eliminated by making additional getTensorShape signature that takes Shape& reference and fills existing object instead of creating a new one.

Assuming the following model (similar with the test model I am working on):
(note that `#0` means 0th tensor)

```
#0 = input of shape [n] and n is the number of input from user
#1 = const of value [20]
WHILE #1 > 0
  #1 = SUB (#1 , 1)
  #0 = ADD (#0, #0)
#2 (output) = #0
```

application code is
```
nnfw_prepare();

while (input is `q`)
   n = read from user
   nnfw_set_input_tensor(#0, [n])
   nnfw_run()
```

`n` changes over time. In addition there is a WHILE loop. So caching tensorshape for `ADD` is overhead in this scenario.
`SUB` is also similar. Cache hit ratio is almost 0% in this case.

compute/cker/include/cker/Shape.h
hyunsik-yoon: @binarman for https://github.com/Samsung/ONE/pull/4611#discussion_r505117981
I agree now. :-)
Maybe we will find ways to reduce memory when we started to get strong request to reduce memory later. :-)

runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
hyunsik-yoon: @binarman for https://github.com/Samsung/ONE/pull/4611#discussion_r505117981
I agree now. :-)
Maybe we will find ways to reduce memory when we started to get strong request to reduce memory later. :-)

compute/cker/include/cker/Shape.h
krayzemli: @binarman self._size is always 0 after Resize(0) in the beginning, all memory is freed, so other._size becomes 0 after the size swap.

krayzemli: By the way, I don't know the reason why operator= is disabled for the Shape, so I've made new `ReplaceWith` signatures.

binarman: > this->_size is always 0 after Resize(0)

Then it is ok, thank you.

> By the way, I don't know the reason why operator= is disabled

I think to avoid unintentional memory copies.
Though it was long ago, maybe @hseok-oh can remember the reason.

runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
krayzemli: @hyunsik-yoon 
> So caching tensorshape for `ADD` is overhead in this scenario.
> `SUB` is also similar. Cache hit ratio is almost 0% in this case.

The only overhead here is that cached parameters are stored in Eval object local memory (so they can be reused on the cache hit) rather than on the stack (where they can't be reused). There is some very small overhead of Shape::ReplaceWith, but it can be easily eliminated (see https://github.com/krayzemli/ONE/tree/binShapeCache2 ). Since we don't have real dynamic networks for experimenting with, we can't actually measure performance impact of cache miss.

compute/cker/include/cker/Shape.h
hseok-oh: >By the way, I don't know the reason why operator= is disabled

`Shape` class comes from tensorflow lite's `RuntimeShep` class. IMO, it is disabled for blocking unexpected copy.

runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
hyunsik-yoon: @krayzemli 

My curiosity is not about its overhead but more about whether this line is meaningful or not.

IMHO, this line is will provide improvement when _previous_ `nnfw_run()`'s shape is same with those in the _next_ `nnfw_run()` (case A) 
but this line is not meaningful when those shapes are different most of time (case B). 

Case B will be some language model where input size varies whenever `nnfw_run()` is called like [in the example I mentioned](https://github.com/Samsung/ONE/pull/4611/files#r505173665).

I wonder what would be the real example for case A.
If we don't have a specific example for case A, IMHO there is no point to leave this line (regardless of how much  overhead it adds or not.) considering [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

If we don't know case A at this moment, IMHO how about removing this line and adds some comment like following?
```
// TODO Adds lines like the following when a model using dynamic tensors needs shape caching
// if (output->is_dynamic())
//      updateCache(lhs, rhs, output);
```

krayzemli: This line updates (invalidates) the cache every run (since a tensor became dynamic, it is never static again), without this line dynamic networks become broken. Only static models benefit from the cache reuse,.

hyunsik-yoon: @krayzemli Oh. I see. Then let's leave as is. :-)

glistening: > This line updates (invalidates) the cache every run (since a tensor became dynamic, it is never static again), without this line dynamic networks become broken. Only static models benefit from the cache reuse,.

@krayzemli It seems you still assume once it become dynamic, it cannot be back to static. We may make a rule internally after discussion. However, I've never heard explicitly about this kind of rule. Thus, I don't think it is always true.

krayzemli: @glisting To make tensor static again, static shape inferer must be rerun, and right now, there is no such mechanism in the runtime. Introducing such a mechanism will trigger a lot of redesign anyway. I think I could start such a redesign soon after my probation.

binarman: @krayzemli 
I'll try to reword what @glistening said:
- We do not have agreement about dynamic shapes yet and we don't have any checking mechanism in runtime that will ensure properties of tensors.
- Many people are working/worked/will work on runtime so we can not be sure code will stay as it is today.

So we can not rely on 
> now, there is no such mechanism in the runtime

> Introducing such a mechanism will trigger a lot of redesign anyway

I am not sure this is hard to introduce, there are a lot of logic in data-flow executor, for example. I believe some magic related to shapes could be done there.

Anyway, if you want to redesign something - feel free to start a conversation, so everyone can understand your ideas =)

krayzemli: @binarman Would adding a formal assert solve the problem then?

glistening: @binarman Thank you for rewording to clarifying my statements.

@krayzemli Yes, if you can add `assert` here, it will make `onert` easy to debug in not-allowed but happened case.

glistening (approved):
LGTM. Though I would like to get `assert` or runtime check (https://github.com/Samsung/ONE/pull/4611#discussion_r508111163),  it better be checked in early stage rather than `BinaryArithmetic`.

glistening:
@krayzemli I wonder whether this PR is live, can others resume the review? Or are you preparing something to improve?

krayzemli:
@glistening

I'm not sure how to improve this PR, I'm just waiting for someone to merge it, or be more specific on desired changes.

glistening:
@krayzemli Okay. I am not asking you to improve it. It is the reason I approved.

@periannath, @hyunsik-yoon, @binarman, (@hseok-oh)  PTAL. Please give your opinion with this PR as it is.

periannath (approved):
LGTM. I think `assert` in line 64 covers the concern I mentioned in https://github.com/Samsung/ONE/pull/4611/files#r504390712.

hseok-oh (approved):

hyunsik-yoon (approved):
LGTM

glistening:
@binarman PTAL

binarman (approved):
LGTM
Hope assert will ensure dynamic tensor invariant

glistening:
> LGTM
> Hope assert will ensure dynamic tensor invariant

Yes, we have to add runtime check in earlier phase than `BinaryArithmeticLayer` after we have conclusion in #4625.

PR #4614, opened 2020-10-13
[onert] Devirtualize calls in getTensorShape

binarman:
Added non virtual get_info method to IPortableTensor class that helped to remove redundant virtual method calls.

### DIFF
--- a/runtime/onert/core/include/backend/IPortableTensor.h
+++ b/runtime/onert/core/include/backend/IPortableTensor.h
@@ -18,6 +18,7 @@
 #define __ONERT_BACKEND_I_PORTABLE_TENSOR_H__
 
 #include "backend/ITensor.h"
+#include "ir/OperandInfo.h"
 #include "ir/Sparsity.h"
 
 namespace onert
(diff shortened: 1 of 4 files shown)


binarman:
On my device I've measured improvement 0.444 ms -> 0.424 ms (~4.5%)

I did 5 measurements using `taskset -a 64 ./nnpackage_run --nnpackage package --warmup_runs=100 --num_runs=10000` then averaged them: [0.418, 0.425, 0.429, 0.427, 0.422] vs [0.438, 0.445, 0.444, 0.446, 0.448]

runtime/onert/core/include/backend/IPortableTensor.h
```diff
@@ -37,12 +38,18 @@ namespace backend
 class IPortableTensor : public ITensor
 {
 public:
+  IPortableTensor(const ir::OperandInfo &info) : _info(info) {}
+
   virtual ~IPortableTensor();
   virtual const ir::Sparsity *sparsity() const { return nullptr; }
+  const ir::OperandInfo &get_info() const { return _info; }
```
binarman, line 45: As you can see I moved _info field from derived classes here.
I suppose this is fine, but I need to verify it with code owner, I hope I added right persons into review =)

Whole idea of this changes is: many calls to small virtual methods are expensive, and this PR removes calls to `ITensor::dimension` and `ITensor::num_dimensions` in cpu_backend.

glistening (approved):
LGTM

wateret (approved):
LGTM

wateret:
@binarman What about making it something like https://github.com/Samsung/ONE/issues/4544#issuecomment-705304811 or https://github.com/Samsung/ONE/issues/4544#issuecomment-707725495 eventually? Directly using `ir::Shape` is not an option as kernel modules must not depend on onert core.

binarman:
> What about making it something like #4544 (comment) or #4544 (comment) eventually? (Directly using ir::Shape is not an option as kernel modules must not depend on onert core.)

Thank you for this explanation, I missed this conversation.
If we do not want to `kernel -> core` dependency we definitely need to rework this PR.

PR #4618, opened 2020-10-14
[onert] Check nullptr when dynamic_cast

hseok-oh:
Check nullptr and throw exception if dynamic_cast result is nullptr

---

Related issue:  https://github.com/Samsung/ONE/issues/1635#issuecomment-708121712


wateret:
This code intended, do we have another way to workaround?

hseok-oh:
>This code intended, do we have another way to workaround?

Mark as `intended` on `svace` and `coverity` issue, and ignore.

wateret:
Sounds better, and FYI this code is likely to be refactored by next release.

PR #4630, opened 2020-10-15
[onert/cpu] Initialize ReduceLayer reduce type field

hseok-oh:
Initialize _reduceType field and use all target
New reduce type for initialize and debug: kInvliad
Fix exception log

---

Related issue: https://github.com/Samsung/ONE/issues/1635#issuecomment-708121712


ragmani (approved):
LGTM

PR #4632, opened 2020-10-15
[onert] Install armcompute .so file only

hseok-oh:
When installing arm compute library for test, copy .so file only


ragmani (approved):
LGTM

Issue #4631, opened 2020-10-15
[onert] Coverity warning by munmap in base_loader


hseok-oh:
http://10.113.139.115:8081/reports.htm#v13368/p10494/fileInstanceId=292815710&defectInstanceId=104472563&mergedDefectId=1153376

>A non-heap pointer is placed on the free list, likely causing a crash later.
>
>In onert::​base_loader::​BaseLoader<onert::​tflite_loader::​<unnamed>::​LoaderDomain>::​deallocateMmappedArea(unsigned char *, unsigned int): Free of an address-of expression, which can never be heap allocated (CWE-590)

CWE-590: http://cwe.mitre.org/data/definitions/590.html

https://github.com/Samsung/ONE/blob/e2e63858960770b0a63eb16873a01f959e9c3391/runtime/onert/frontend/base_loader/include/base_loader.h#L293-L301

This implementation is intrudeced by #4007

@periannath Could you resolve this issue before 1.10 release?

periannath:
I'll check and fix it ASAP.

hseok-oh:
Resolved

PR #4636, opened 2020-10-15
[onert] Match mmap-munmap pair in base_loader

periannath:
- Use mmap and munmap to read tensor data for CachedData
- Removes unnecessary unmap helper function

---

For issue #4631


hseok-oh (approved):
LGTM

PR #4648, opened 2020-10-16
[onert/cker] Define temporary variable to avoid warning

hseok-oh:
Fix build fail  on gcc toolchain 9.x by warning: non-const parameter aliases with const parameter

---

Related issue: #4645


glistening (approved):
LGTM. Without this commit, android build fails.

Issue #4615, opened 2020-10-14
[circle-quanitzer] Quantize CPN (Cascaded Pyramid Network)


jinevening:
Let's quantize CPN.

Original paper: https://arxiv.org/abs/1711.07319

jinevening:
ResizeBilinear: Inputs are image and new_size. new_size should not be quantized.

jinevening:
IN CPN, multiple TCONV layers share one tensor as their weights.

Currently, circle-quantizer does not support quantization of weights shared by multiple layers. I'll implement the functionality.

jinevening:
Done

PR #4649, opened 2020-10-16
[luci] Support quantization of shared tensor

jinevening:
This supports quantization of a tensor shared by multiple operators

---
Related to: https://github.com/Samsung/ONE/issues/4615

### DIFF
--- a/compiler/luci/pass/src/QuantizationUtils.cpp
+++ b/compiler/luci/pass/src/QuantizationUtils.cpp
@@ -24,6 +24,52 @@
 namespace luci
 {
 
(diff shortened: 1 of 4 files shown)


jinevening:
Previously, a tensor shared by multiple operators was not quantized. Below code blocked the quantization.

``` cpp
  if (succs.size() != 1) // assume weights is used by only one node
    return false;
```

I've removed the code and enabled quantization of a tensor shared by multiple operators with the same opcode, e.g., weights tensor shared by multiple TCONV operators in CPN.

seanshpark (approved):
LGTM

Issue #4444, opened 2020-09-28
Channel-wise INT16 quantization support


jinevening:
We need to support channel-wise INT16 quantization in one-quantize.

- [x] circle-tensordump
- [x] luci
- [x] pota-quantization-value-test (for 12 operators)
- [x] one-quantize

jinevening:
I reopened this issue to support int64 bias.

Things to do
- [ ] Remove test data for int16 in pota-quantization-value-test
- [ ] Implement int64 bias code (after [this issue](https://github.sec.samsung.net/AIP/Trinity/issues/139) is resolved)
- [ ] Add new test data for int16 in pota-quantization-value-test
- [ ] Give examples for `one-quantize` with int16 option

jinevening:
### Example

`one-quantize` can use int16 quantization by passing `int16` to `--quantized_dtype` and `channel` to `--granularity`.

```
one-quantize \
--input_dtype float32 \
--quantized_dtype int16 \
--granularity channel \
--input_path ./inception_v3.circle \
--input_data ./inception_v3_test_data.h5 \
--output_path ./inception_v3.quantized.circle
```

### Negative example

We only support CWQ int16 quantization. If you try to do int16 quantization with `layer` (default granularity), the below error message will be printed.

```
one-quantize \
--input_dtype float32 \
--quantized_dtype int16 \
--input_path ./inception_v3.circle \
--input_data ./inception_v3_test_data.h5 \
--output_path ./inception_v3.quantized.circle
```

```
ERROR: Layer-wise quantization only supports uint8 dtype.
```

jinevening:
I close this issue as all jobs required for int64 bias are done.

PR #4654, opened 2020-10-16
[luci] Add exception for quantized_dtype of LWQ

jinevening:
This throws exception when quantized_dtype of LWQ is not uint8

---
Related to: https://github.com/Samsung/ONE/issues/4444


jinevening:
INT16 does not support LWQ, so I added exception code for the case.

seanshpark:
(1) if SOME model CAN have this condition, then it would be better to SKIP quantization.
(2) if ALL model MUST have this condition when running this routine,it is NECESSARY to throw.

is this the case of (2) ?

seanshpark:
Seems (1) :)

seanshpark (approved):
LGTM

jinevening:
> Seems (2) :)

Yes, currently quantized_dtype of LWQ should be uint8 for all models. :)

PR #4655, opened 2020-10-16
[onert] Allows variable tensor

ragmani:
For issue #3851

This commit allows variable tensor.
  - Introduces a field checking variable tensor into OperandInfo
  - Apply variable tensor

### DIFF
--- a/runtime/onert/core/include/ir/OperandInfo.h
+++ b/runtime/onert/core/include/ir/OperandInfo.h
@@ -66,8 +66,9 @@ public:
    * @param[in] alloc_type  When the thesor needs memory allocation
    */
   OperandInfo(const Shape &shape, const TypeInfo &typeInfo, MemAllocType alloc_type,
-              bool is_const = false)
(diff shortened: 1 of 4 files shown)


wateret (changes requested):
Could you please explain what is difference between non-const(`IsConstant()` is false) and variable tensors?

ragmani:
> Could you please explain what is difference between non-const(IsConstant() is false) and variable tensors?

@wateret 
Variable tensor means mutable tensor. And TFLite interpreter only supports variable tensor without buffers in the model.

wateret (approved):
LGTM, I heard the explanation offline.

- `is_variable` is an explicit flag which is present in tflite schema
- No operation generates it (like constants)
- It does not have data (unlike constants)
- Only one operation is allowed to use a variable tensor
- Seems like it is used as a temp space to store the current state (not 100% sure)

hseok-oh (approved):

PR #4660, opened 2020-10-16
[onert/acl] Fix AclTensorManager::at

wateret:
Return nullptr if the tensor is not present instead of throw.


ragmani (approved):
LGTM

PR #4665, opened 2020-10-19
[record-minmax] Catch H5Exception to prevent runtime abort

jinevening:
This catches H5Exception and rethrows it as std exception


jinevening:
For reviewers: Note that this PR just wraps the existing code (using HDF5) with try-catch block. This does not change the main program logic.

seanshpark (approved):
LGTM

Issue #4407, opened 2020-09-24
[onert] Concat test fail


hseok-oh:
On my environment (odroid-xu4, ubuntu 20.04)

`GenModelTest` concat test added on PR #4112 fail 
(`GenModelTest.OneOp_Concat_Subtensor_4D`, `GenModelTest.OneOp_Concat_Subtensor_2D`)

Fail on debug/release build

```
$ ./Product/out/unittest_standalone/nnfw_api_gtest --gtest_filter=GenModelTest.OneOp_Concat_*
Note: Google Test filter = GenModelTest.OneOp_Concat_*
[==========] Running 3 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 3 tests from GenModelTest
[ RUN      ] GenModelTest.OneOp_Concat_ShareSubTensor
[       OK ] GenModelTest.OneOp_Concat_ShareSubTensor (406 ms)
[ RUN      ] GenModelTest.OneOp_Concat_Subtensor_4D
/home/nfs/git/ONE/tests/nnfw_api/src/GenModelTest.h:389: Failure
The difference between refval and val is 7, which exceeds 0.001, where
refval evaluates to 1,
val evaluates to 8, and
0.001 evaluates to 0.001.
e == 0
/home/nfs/git/ONE/tests/nnfw_api/src/GenModelTest.h:389: Failure
The difference between refval and val is 2, which exceeds 0.001, where
refval evaluates to 2,
val evaluates to 0, and
0.001 evaluates to 0.001.
e == 1
/home/nfs/git/ONE/tests/nnfw_api/src/GenModelTest.h:389: Failure
The difference between refval and val is 3, which exceeds 0.001, where
refval evaluates to 3,
val evaluates to 0, and
0.001 evaluates to 0.001.
e == 2
/home/nfs/git/ONE/tests/nnfw_api/src/GenModelTest.h:389: Failure
The difference between refval and val is 4, which exceeds 0.001, where
refval evaluates to 4,
val evaluates to 0, and
0.001 evaluates to 0.001.
e == 3
/home/nfs/git/ONE/tests/nnfw_api/src/GenModelTest.h:389: Failure
The difference between refval and val is 5, which exceeds 0.001, where
refval evaluates to 5,
val evaluates to 0, and
0.001 evaluates to 0.001.
e == 4
/home/nfs/git/ONE/tests/nnfw_api/src/GenModelTest.h:389: Failure
The difference between refval and val is 11, which exceeds 0.001, where
[...]

PR #4668, opened 2020-10-19
[onert] Fix concat elimination bug

hseok-oh:
Don't use concat elimination if axis is last dimension and backend tensor uses padding (acl-cl backend)

---

Related issue: #4407


wateret (approved):
LGTM

PR #4670, opened 2020-10-19
[nnfwapi] Add experimental API TCs

wateret:
TCs for `nnfw_input_tensorindex` and `nnfw_output_tensorindex`.

- 2 positive tests
- 10 negative tests

ONE-DCO-1.0-Signed-off-by: Hanjoung Lee <hanjoung.lee@samsung.com>

PR #4671, opened 2020-10-19
[nnfwapi] Add more tests for Cast op

wateret:
- Extract CircleGen generation into a separate function
- Add two more tests


wateret:
Negative TCs are added in #4670 (8 more negative tests)

glistening (approved):
LGTM

PR #4679, opened 2020-10-20
[OneOpTest] Enable concat test on all backends

hseok-oh:
Enable concat test on all backends: acl-cl, acl-neon, cpu


wateret (approved):
LGTM

PR #4684, opened 2020-10-20
[nnfw_api] Introduce NNFW_TYPE_TENSOR_QUANT8_ASYMM_SIGNED type

hseok-oh:
- Introduce NNFW_TYPE_TENSOR_QUANT8_ASYMM_SIGNED type
- TFLite and circle model int8 type is matched with QUANT_INT8_ASYMM
  - QUANT_INT8_SYMM is used for weight tensor only


runtime/onert/core/src/compiler/OperationValidator.cc
```diff
@@ -107,7 +107,7 @@ void OperationValidator::visit(const ir::operation::ElementwiseUnary &node)
   if (node.param().op_type == ir::operation::ElementwiseUnary::Type::DEQUANTIZE)
   {
     OP_REQUIRES((_ctx.at(input_index).typeInfo().type() == ir::DataType::QUANT_UINT8_ASYMM) ||
-                (_ctx.at(input_index).typeInfo().type() == ir::DataType::QUANT_INT8_SYMM));
+                (_ctx.at(input_index).typeInfo().type() == ir::DataType::QUANT_INT8_ASYMM));
```
periannath, line 110: `Dequantize` TCs get input as `TENSOR_QUANT8_SYMM` type. It seems that TC fails at here.

hseok-oh: @periannath I've updated. PTAL

periannath (dismissed):
LGTM

wateret (changes requested):
Please add `static_assert` too 😄 

https://github.com/Samsung/ONE/blob/cf2d40929e1f2cf992ea7ac49de53d787e668d2f/runtime/onert/api/src/nnfw_api.cc#L24-L29

hseok-oh:
@periannath @wateret I've updated. PTAL

periannath (approved):
LGTM

wateret (approved):
LGTM

PR #4686, opened 2020-10-20
[nnfwapi] Add ReduceMax tests

wateret:
Make ShapeInferer to throw rather than `assert`.
Add ReduceMax GenModel tests - 1 positive / 4 negatives.


hyunsik-yoon:
I think the CircleGen test should fail compilation against `OperationValidator`. (rather than `ShapeInference.cc`)
Maybe we can add `reduce` into `OperationValidator` in next PR.

wateret:
@hyunsik-yoon "Axes" is a tensor so I thought it could change every time in an execution. That's why I wanted to throw here. What about it?

hyunsik-yoon (approved):
LGTM

PR #4688, opened 2020-10-20
[luci] int64 bias for INT16 quantization

jinevening:
This supports int64 bias for INT16 quantization

---
Related to: https://github.com/Samsung/ONE/issues/4444

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -248,12 +248,9 @@ void int16_quant_bias_per_tensor(CircleConst *node, float input_scale,
 
   node->dtype(loco::DataType::S64);      // change the type of tensor
   node->size<loco::DataType::S64>(size); // resize tensor
-  const int64_t kMinScale = std::numeric_limits<int64_t>::lowest();
-  const int64_t kMaxScale = std::numeric_limits<int64_t>::max();
   for (uint32_t i = 0; i < size; ++i)
   {
-    node->at<loco::DataType::S64>(i) =
-        std::min(kMaxScale, std::max(kMinScale, quantized_values[i]));
+    node->at<loco::DataType::S64>(i) = quantized_values[i];
   }
 }
 
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -226,6 +226,37 @@ void quant_bias_per_channel(CircleConst *node, float input_scale, std::vector<fl
   }
 }
 
+void int16_quant_bias_per_tensor(CircleConst *node, float input_scale,
+                                 std::vector<float> &weight_scale,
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

meejeong (dismissed):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -226,6 +226,37 @@ void quant_bias_per_channel(CircleConst *node, float input_scale, std::vector<fl
...
+        static_cast<int64_t>(std::round(node->at<loco::DataType::FLOAT32>(i) * scaling_factor_inv));
+  }
+
+  node->dtype(loco::DataType::S64);      // change the type of tensor
+  node->size<loco::DataType::S64>(size); // resize tensor
+  const int64_t kMinScale = std::numeric_limits<int64_t>::lowest();
+  const int64_t kMaxScale = std::numeric_limits<int64_t>::max();
+  for (uint32_t i = 0; i < size; ++i)
+  {
+    node->at<loco::DataType::S64>(i) =
```
s-barannikov, line 255: This clamping seems unnecessary as the `quantized_values` is vector<int64_t>.
Not sure about the lowest bound though, should it be -2^32 or -(2^32-1)?
In any case this loop could be fused into the loop above :)

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -465,7 +496,22 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<bool>
...
       std::vector<int64_t> zp(size);
 
-      quant_bias_per_channel(circle_const, input_scale, weight_scale, scaling_factor, zp);
+      if (output_type == loco::DataType::U8)
+      {
+        quant_bias_per_channel(circle_const, input_scale, weight_scale, scaling_factor, zp);
+      }
+      else if (output_type == loco::DataType::S16)
+      {
+        // Bias is quantized per-tensor in int16 quantization
```
s-barannikov, line 505: Is it OK to do per-tensor quantization in `granularity == ChannelWise` branch?

jinevening: Yes. That's how we decide the scale of bias in CWQ. (we should use one scale for the bias for some reason)

s-barannikov: I see, thanks

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: Clamping seems unncessary. I've updated the code.

It is not simple to fuse the loops because the type/shape of node is changed between the loops.

s-barannikov (approved):

seanshpark (approved):
LGTM

meejeong (approved):
LGTM

jinevening:
@seanshpark Can you merge this PR? :)

PR #4689, opened 2020-10-20
[onert/cpu] Fix Add&Sub assert bug

hseok-oh:
Fix assert bug: cker param lhs and rhs offset is negative


ragmani (approved):
LGTM

Issue #4528, opened 2020-10-07
Compiler FE: RNN and LSTMCell


seanshpark:
Let's enable `RNN` and `LSTMCell` two operators that was forgotten...
- RNN: #1712
- LSTMCell: missed somehow (LSTM at #1477)

For #1928

seanshpark:
Like #4526, made TF example script
```py
import tensorflow as tf

sess = tf.Session()

in_ = tf.compat.v1.placeholder(dtype=tf.float32, shape=(1, 2, 3), name="Hole")
op_ = tf.keras.layers.RNN(tf.keras.layers.LSTMCell(4))
ex_ = op_(in_)
out = tf.identity(ex_, name="Output")

# we need to save checkpoint to freeze dropped model
init = tf.global_variables_initializer()
sess.run(init)

saver = tf.train.Saver()
saver.save(sess, './ckpt/LSTMCell.ckpt')

# use below command to freeze this model after running tfpem.py
'''
freeze_graph --input_graph LSTMCell.pbtxt \
--input_binary=false \
--input_checkpoint=./ckpt/LSTMCell.ckpt \
--output_node_names=Output \
--output_graph LSTMCell_fr.pbtxt
'''
```
But drops `IndexError: list index out of range` error when run `freeze_graph`

Environment: TF 1.13.x

seanshpark:
CC @lucenticus

seanshpark:
Although not freezed model (with variables) this command can convert to circle file with `--v2` option
```
one-import-tf --input_path LSTMCell.pbtxt --output_path LSTMCell.circle --input_arrays Hole --output_arrays Output --v2
```
I think there are no additional Ops to enable: as you can see from the circle dump:
```
Operator Codes: [order] OpCodeName (OpCode Enum)
[0] CUSTOM(VarHandleOp) (code: 32, version: 1)
[1] CUSTOM(TensorArrayV3) (code: 32, version: 1)
[2] TRANSPOSE (code: 39, version: 1)
[3] CUSTOM(TensorArrayScatterV3) (code: 32, version: 1)
[4] WHILE (code: 119, version: 1)
[5] CUSTOM(TensorArraySizeV3) (code: 32, version: 1)
[6] RANGE (code: 96, version: 1)
[7] CUSTOM(TensorArrayGatherV3) (code: 32, version: 1)
[8] STRIDED_SLICE (code: 45, version: 1)
[9] LESS (code: 58, version: 1)
[...]

PR #4698, opened 2020-10-20
[tf2tfliteV2] Support conversion of recurrent Keras models

binarman:
Added support of models with lstm, rnn and gru layers

### DIFF
--- a/compiler/tf2tfliteV2/tf2tfliteV2.py
+++ b/compiler/tf2tfliteV2/tf2tfliteV2.py
@@ -18,6 +18,7 @@
 import tensorflow as tf
 import argparse
 import sys
+import tempfile
 
 from google.protobuf.message import DecodeError
 from google.protobuf import text_format as _text_format
@@ -142,6 +143,22 @@ def _parse_array(arrays, type_fn=str):
     return list(map(type_fn, arrays.split(",")))
 
 
+def _v1_convert_keras(model_path):
+    model = tf.keras.models.load_model(model_path)
+    try_saved_model = False
+    for l in model.layers:
+        if l.name in ["lstm", "simple_rnn", "gru"]:
+            try_saved_model = True
+            break
+    if try_saved_model:
+        tmpd = tempfile.TemporaryDirectory().name
+        model.save(tmpd)
+        converter = tf.compat.v1.lite.TFLiteConverter.from_saved_model(tmpd)
+    else:
+        converter = tf.compat.v1.lite.TFLiteConverter.from_keras_model_file(model_path)
+    return converter
+
+
 def _v1_convert(flags):
     if flags.model_format == "graph_def":
         if not flags.input_arrays:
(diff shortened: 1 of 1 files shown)


binarman:
For now I checked that original keras and generated tflite models behaves equal. (I've run them with several inputs)
Still need to check that circle model is fine too (will do this using luci-interpreter), but this is probably is out of this PR scope.

compiler/tf2tfliteV2/tf2tfliteV2.py
```diff
@@ -142,6 +143,22 @@ def _parse_array(arrays, type_fn=str):
     return list(map(type_fn, arrays.split(",")))
 
 
+def _v1_convert_keras(model_path):
+    model = tf.keras.models.load_model(model_path)
+    try_saved_model = False
+    for l in model.layers:
+        if l.name in ["lstm", "simple_rnn", "gru"]:
```
seanshpark, line 150: Can you please add some comments behind this? 
The codes seems to save `keras mode` to `saved model file` and then get the converter...

binarman: You are right, this is exactly what this code does, I tried several combinations (see list below) and at first look this PR solution looked like the only one working, so I created this PR.

Though, my methodology is a mess, (I did not check all tf versions and ways), so then I carefully analyzed results and maybe this PR is not necessary.
Looks like RNN models are properly converted only since some version of TF (like 2.1.0) by one-step converters.

What I tried:
- tf1.13.1 (checkpoint) -> freeze_graph(1.13.1 and 2.0.0) -> TFLiteConverter.from_frozen_graph (**failed** on graph freeze)
- tf2.3.1 (checkpoint) -> freeze_graph(1.13.1 and 2.0.0) -> TFLiteConverter.from_frozen_graph (**failed** on graph freeze)
- tf1.13.1 (**keras model**) -> keras load + keras save (tf 2.3.1, 2.1.0, 2.2.1) -> TFLiteConverter.from_saved_model (v2) (**SUCCESS**)
- tf1.13.1 (**keras model**) -> tflite_convert (tf 2.0.0)(v1 and v2) (**failed** with ValueError: Cannot find the Placeholder op that is an input to the ReadVariableOp.)
- tf1.13.1 (saved model) -> tflite_convert (1.13.1) (**failed** with segfault)
- tf1.13.1 (saved model) -> tflite_convert (2.0.0) (**failed** with "ValueError: Cannot find the Placeholder op that is an input to the ReadVariableOp.")
- tf1.13.1 (saved model) -> tflite_convert (2.3.1) (**failed**  with exception during exception handling)
- tf1.13.1 (saved model) -> TFLiteConverter.from_saved_model (2.3.1, toco/MLIR versions of converter) (**failed** with exception during exception handling)

binarman:
I consider this PR does not add any new functionality (see https://github.com/Samsung/ONE/issues/4528#issuecomment-716957802), closing it

PR #4700, opened 2020-10-20
[luci-interpreter] Simple tensor memory management

krayzemli:
Reduce tensor data memory usage. Allocate lazily on demand. Deallocate when not needed any more.

Theoretical tensor memory usage on inception V3: 90.9 MB (weights) + 200.6 MB (other) --> 90.9 MB + 8.8 MB

luci-value-tester peak usage: 501 MB --> 386 MB (same as preparing the interpreter and skipping inference)

### DIFF
--- a/compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
@@ -118,13 +118,13 @@ public:
 
   template <typename T> const T *data() const
   {
-    assert(!!_data);
+    assert(_data_allocated);
     return reinterpret_cast<const T *>(_data.get());
   }
 
   template <typename T> T *data()
   {
-    if (!_data)
+    if (!_data_allocated)
       allocate();
     return reinterpret_cast<T *>(_data.get());
   }
@@ -143,6 +143,7 @@ private:
   AffineQuantization _quantization;
   std::unique_ptr<uint8_t[]> _data;
   std::string _name;
+  bool _data_allocated;
 };
 
 } // namespace luci_interpreter
--- a/compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
@@ -107,15 +107,27 @@ public:
     return _quantization.zero_point[0];
   }
 
+  void allocate();
+  void deallocate();
+
   const std::vector<float> &scales() const { return _quantization.scale; }
 
   const std::vector<int32_t> &zero_points() const { return _quantization.zero_point; }
 
   int32_t quantized_dimension() const { return _quantization.quantized_dimension; }
 
-  template <typename T> const T *data() const { return reinterpret_cast<const T *>(_data.get()); }
+  template <typename T> const T *data() const
+  {
+    assert(!!_data);
+    return reinterpret_cast<const T *>(_data.get());
+  }
 
-  template <typename T> T *data() { return reinterpret_cast<T *>(_data.get()); }
+  template <typename T> T *data()
+  {
+    if (!_data)
+      allocate();
+    return reinterpret_cast<T *>(_data.get());
+  }
 
   const std::string &name() const { return _name; }
 
--- a/compiler/luci-interpreter/src/core/RuntimeGraph.cpp
+++ b/compiler/luci-interpreter/src/core/RuntimeGraph.cpp
@@ -19,6 +19,7 @@
 #include "core/RuntimeModule.h"
 
 #include <algorithm>
+#include <unordered_map>
 
 namespace luci_interpreter
 {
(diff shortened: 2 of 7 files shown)


compiler/luci-interpreter/src/core/RuntimeGraph.h
```diff
@@ -46,13 +46,18 @@ class RuntimeGraph
...
+
   RuntimeModule *_owning_module;
   std::vector<std::unique_ptr<Tensor>> _tensors;
   std::vector<Tensor *> _input_tensors;
   std::vector<Tensor *> _output_tensors;
 
   // Kernels in execution order.
   std::vector<std::unique_ptr<Kernel>> _kernels;
+  // Tensors that are not used anymore after given op
+  mutable std::vector<std::vector<Tensor *>> _tensor_dealloc_plan;
```
binarman, line 60: I propose to add some non-constant method, like "prepare" and do allocation/deallocation planning there, so we do not need mutable magic here.

Other way is to add incremental planning in addKernel method, but I think this will complicate code a little.

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
```diff
@@ -107,15 +107,27 @@ class Tensor
...
+  {
+    assert(!!_data);
+    return reinterpret_cast<const T *>(_data.get());
+  }
 
-  template <typename T> T *data() { return reinterpret_cast<T *>(_data.get()); }
+  template <typename T> T *data()
+  {
+    if (!_data)
+      allocate();
```
binarman, line 128: New memory management logic looks a little entangled to me.
You allocate memory at first write(?) access to tensor from kernels (implicitly), deallocates it on resize(implicitly again) and on demand from RuntimeGraph.

Maybe we can move intermediate tensors management on single level of interpreter, in RuntimeGraph?
I mean not only deallocate tensors, but also allocate needed before kernel executes.
This also applies to im2col and other aux tensors: kernel explicitly allocates it and deallocates internal tensors, so there are less hidden behavior.

P.s. Note that `allocate` method is not used outside of `Tensor` class, that looks strange.

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
```diff
@@ -107,15 +107,27 @@ class Tensor
...
   const std::vector<float> &scales() const { return _quantization.scale; }
 
   const std::vector<int32_t> &zero_points() const { return _quantization.zero_point; }
 
   int32_t quantized_dimension() const { return _quantization.quantized_dimension; }
 
-  template <typename T> const T *data() const { return reinterpret_cast<const T *>(_data.get()); }
+  template <typename T> const T *data() const
+  {
+    assert(!!_data);
```
binarman, line 121: Just a nitpick: do you need explicit conversion to bool?

compiler/luci-interpreter/src/core/RuntimeGraph.h
krayzemli: `mutable` usage is typical for storing cached data, like here. This cache does not keep any additional state, it duplicates the information already available in other fields. One way to hide `mutable` is using `unique_ptr` indirection, interpreter uses such an approach many timer (like, say, keeping temporary tensors). Adding `prepare` non-const method would complicate the interface if the class.

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
krayzemli: Allocating memory transparently on first use is a classic approach in memory management, and it solves the half of the problem of automatic memory management. It secures from use-after-free and complies with KISS principle, in contrast to explicit allocation calls. Using manual allocation everywhere would just make interpreter code more complicated without any clear benefit.

Old `resize` unconditionally reallocates (i.e. deallocates and allocates again) the memory, it does not copy data or something, so the new behavior is fully compatible, since allocation will be done on a first write. 

For dynamic shape inferring, auto-allocation and explicit free approach is simple and effective. Introducing static memory management will  make necessary rewriting quite a lot of the interpreter code, since it is not designed for static shape inferring now.

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
krayzemli: Yes, I like this two-char idiom.

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
binarman: > complies with KISS principle

It is not =_=
The thing is this code violates KISS and SRP, because you merge memory management logic into unexpected places (at least from my perspective).

I see you like lazy computations. I agree that they could be extremely useful for applications where some computations are likely not to happen at all or we can benefit from gathering some chunks of computations and then execute them all at once.

Unfortunately I don't see how this code benefits from lazy model (though I believe memory manager can use this technic in some cases), so from my perspective this code is more complex then it should be.

Though I think we can go as-is for now.

compiler/luci-interpreter/src/core/RuntimeGraph.cpp
```diff
@@ -73,6 +155,10 @@ void RuntimeGraph::execute() const
     // TODO The `configure` method should only be called if the outputs of an operator need to be
     //  resized.
     kernel->configure();
+#if 0
```
binarman, line 158: Need to mark it with TODO, so we can find this place later
```suggestion
// TODO decide where to allocate memory, and uncomment/remove this if
#if 0
```

binarman (approved):
LGTM

I think we can proceed as-is

krayzemli:
Less code -- less problems, less code maintenance cost. Explicit allocation logic is not needed right now. Automatic one works well. It might be necessary for complex memory planning algorithms (see runtime memory planners as an example), or when some operator needs to reads an uninitialized tensor memory (for whatever strange reason). I added `#if 0` just to make you happy, there is nothing to do there now, till the time when kernel configure phase becomes separated from kernel execute phase, and more complicated memory planners are introduced. If you need to know the places where the memory may be allocated, you just look for all non-const `Tensor::data` calls.

PR #4715, opened 2020-10-21
[luci] Quantize constants of more Ops

jinevening:
This quantizes constants of six more Ops


seanshpark (approved):
LGTM

Issue #4625, opened 2020-10-14
Make rule for multiple calling of nnfw_set_input_tensorinfo() and nnfw_run()


hyunsik-yoon:
Rule for multiple calling of `nnfw_set_input_tensorinfo()` and `nnfw_run()` is not clear.
How about the following rule?

/cc @Samsung/nnfw 

Rule 1) `nnfw_set_input_tensorinfo()` will **always** set its input tensor _dynamic_ (see case A) and B) in the example below)

Rule 2) Once `nnfw_set_input_tensorinfo(new_shape)` is called, `new_shape` will be used for the successive calls of `nnfw_run()` until another `nnfw_set_input_tensorinfo()` is called. (see case C) and D))

For example: assuming a model has an input of [1, 128]

```
nnfw_prepare();
nnfw_run();  // static input tensor input with [1, 128] 

nnfw_set_input_tensorinfo([1, 128]); // app developer shows and intention to change the shape
nnfw_run();  // *dynamic* input tensor with [1, 128] --> case A)

nnfw_set_input_tensorinfo([5, 128]);
nnfw_run();  // *dynamic* input tensor with [5, 128] --> case B)

// once nnfw_set_input_tensorinfo([5, 128]); is called for the previous nnfw_run()
// let the next nnfw_run() also use [5, 128]
nnfw_run();  // *dynamic* input tensor with [5, 128] --> case C)
nnfw_run();  // same with case C)

nnfw_set_input_tensorinfo([1, 128]);
nnfw_run();  // *dynamic* input tensor with [1, 128] --> case D)

```

FAQ)
- Can we run a model with static input tensor in case of case A) or D)?
  - Consider the following example, (`#0` means `tensor with index 0`)

   ```
   #0 = input 0 of shape [1]
   #1 = input 1 of shape [2]
   #2 = relu(#0)
   #3 = reshape(const of shape [8] , #1)
   #4 = add(#2, #3)
   ```

    - after `nnfw_prepare()`, #3 and #4 become dynamic and others are static --> state A)
    - after calling `nnfw_set_input_tensorinfo(#0, [2,2])` and `nnfw_run()`, all tensors become dynamic
    - then, what if we call `nnfw_set_input_tensorinfo(#0, [1])` and `nnfw_run()` again to run with _static_ input tensor?
      - since all tensors became dynamic, we have to move back to the state A)
      - this means that we have to store state after `nnfw_prepare()`: code will be complex and more memory is required
[...]

PR #4718, opened 2020-10-21
Rewrite set_input_tensorinfo(shape) where shape==model input shape

hyunsik-yoon:
Parent issue: https://github.com/Samsung/ONE/issues/4625

This reverts commit b124ad26ceee32416a735c826c5ea715f8493e40
and adds test cases.

Previously, e.g., if the model input 0's shape is _S0_ and `set_input_tensorinfo(0, S0)` is called,
we don't treat 0th tensor as dynamic tensor.

However, this caused some issue at (1) in the following case. (please refer to #4625) 
So, we decided to treat (1) as well as (0) as dynamic tensor (in the discussion at #4625).
This change can be done by reverting b124ad26ceee32416a735c826c5ea715f8493e40.

```
// assume that input 0's shape in model file is S0
set_input_tensorinfo(0, S0) ---> (0)
nnfw_run()
// set to something different, say, S1
set_input_tensorinfo(0, S1) // now input is dynamic
nnfw_run()
// back to S0?
set_input_tensorinfo(0, S0) ---> (1). This tries to go back to static tensor again, which crashes at this moment.
nnfw_run()
```

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -492,27 +492,26 @@ NNFW_STATUS nnfw_session::apply_tensorinfo(uint32_t index, nnfw_tensorinfo ti)
     }
   }
 
-  auto ind = primary_subgraph()->getInputs().at(index);
-  auto &input = primary_subgraph()->operands().at(ind);
-
-  onert::ir::Shape new_shape(ti.rank);
-  for (int32_t i = 0; i < ti.rank; i++)
-    new_shape.dim(i) = ti.dims[i];
-
-  // if passed shape is same with the shape of model, do nothing
-  if (input.info().shape() == new_shape)
-    return NNFW_STATUS_NO_ERROR;
-
   if (!isStatePreparedOrFinishedRun())
   {
     // In this case, if we apply input shape in primary_subgraph, it will propagate after
     // compilation and excution
+    auto ind = primary_subgraph()->getInputs().at(index);
+    auto &input = primary_subgraph()->operands().at(ind);
+
+    onert::ir::Shape new_shape(ti.rank);
+    for (int32_t i = 0; i < ti.rank; i++)
+      new_shape.dim(i) = ti.dims[i];
 
     // overwrite input shape with the shape from ti
     input.info().shape(new_shape);
   }
   else // when called after nnfw_session::prepare()
   {
+    onert::ir::Shape new_shape(ti.rank);
+    for (int32_t i = 0; i < ti.rank; i++)
+      new_shape.dim(i) = ti.dims[i];
+
     _execution->changeInputShape(onert::ir::IOIndex(index), new_shape);
   }
 
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -492,6 +492,10 @@ NNFW_STATUS nnfw_session::apply_tensorinfo(uint32_t index, nnfw_tensorinfo ti)
     }
   }
 
+  onert::ir::Shape new_shape(ti.rank);
+  for (int32_t i = 0; i < ti.rank; i++)
+    new_shape.dim(i) = ti.dims[i];
+
   if (!isStatePreparedOrFinishedRun())
   {
     // In this case, if we apply input shape in primary_subgraph, it will propagate after
@@ -499,19 +503,11 @@ NNFW_STATUS nnfw_session::apply_tensorinfo(uint32_t index, nnfw_tensorinfo ti)
     auto ind = primary_subgraph()->getInputs().at(index);
     auto &input = primary_subgraph()->operands().at(ind);
 
-    onert::ir::Shape new_shape(ti.rank);
-    for (int32_t i = 0; i < ti.rank; i++)
-      new_shape.dim(i) = ti.dims[i];
-
     // overwrite input shape with the shape from ti
     input.info().shape(new_shape);
   }
   else // when called after nnfw_session::prepare()
   {
-    onert::ir::Shape new_shape(ti.rank);
-    for (int32_t i = 0; i < ti.rank; i++)
-      new_shape.dim(i) = ti.dims[i];
-
     _execution->changeInputShape(onert::ir::IOIndex(index), new_shape);
   }
 
--- a/tests/nnfw_api/src/ModelTestDynamicTensor.cc
+++ b/tests/nnfw_api/src/ModelTestDynamicTensor.cc
@@ -564,3 +564,224 @@ TEST_F(TestIfDynamicModelLoaded, run_verify)
   for (int i = 0; i < actual_output0.size(); ++i)
     ASSERT_NEAR(if_dynamic_output0[i], actual_output0[i], 0.00001);
 }
+
+class CombinationTest1 : public ::testing::Test
+{
+protected:
+  void SetUp() override
+  {
+    CircleGen cgen;
+
+    // Creating a graph which has dynamic tensors after compilation.
+    // #3 and #4 are dynamic. This model was used to check if internal dynamic tensors could
+    // make any side-effect.
+    //
+    // #0 = input 0 of shape [1]
+    // #1 = input 1 of shape [2]
+    // #2 = cast(#0, int to float)
+    // #3 = reshape(const of shape [4] , #1)
+    // #4 = add(#2, #3)
+
+    constexpr circle::TensorType CIRCLE_DTYPE = circle::TensorType::TensorType_FLOAT32;
+
+    int cast_in = cgen.addTensor({{1}, circle::TensorType::TensorType_INT32});
+    int cast_out = cgen.addTensor({{1}, circle::TensorType::TensorType_FLOAT32});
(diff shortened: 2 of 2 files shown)


hyunsik-yoon:
This PR is somewhat complex. I will split this into some.

krayzemli:
Is the problem really related to dynamic shapes? What I see here is wrong comparison not with the actual execution model tensor shape, but with original model graph operand shape passed to the compiler, which is not updated together with dynamic tensors. This comparison makes it impossible to return the shape back to original one.

krayzemli:
The reason is wrong comparison in the logic. Tried to fix the bug in PR #4748, please take a look.

hyunsik-yoon:
b124ad2 was intended to make small optimization but it creates some bugs internally and caused problem.

The difference between this PR (reverting b124ad2) and your PR #4748 is 

- This PR: always change input shape when `nnfw_set_input_tensorinfo()` is called.
- your PR #4748: change input shape when the shape is different from previous input shape

IMHO they both work same except the following case:
- This PR: treats input as dynamic tensor with 2nd and 4th run (please refer to the example below)
- Your PR #4748: treats input as static tensor with 2nd and dynamic tensor with 4th run 

It is a taste of selection... IMHO I prefer this PR since it is _consistent_ policy just like I mentioned in #4625 and I think this is important with API.
`"Invoking set_input_tensorinfo() will make ONE runtime treat input tensor as dynamic tensor."`

```
// 1st run: model file has input shape [1]
nnfw_prepare();
nnfw_run();

// 2nd run: change to [1] (same shape with 1st run)
set_input_tensorinfo(0, [1]);
nnfw_run();

// 3rd run: change to [100]
set_input_tensorinfo(0, [100]);
nnfw_run();

// 4th run: change to [1] again
set_input_tensorinfo(0, [1]);
nnfw_run();
```

hyunsik-yoon:
Rebased to resolve conflict.

krayzemli:
What was the purpose of PR #3998 then? And what is the reason of changing the mind?

If I get everything right, dynamic/static execution mode is an internal to the API and invisible/transparent to the user. There is no API function to know whether the tensor is dynamic or static.

I see the only usecase of forcing the tensor into dynamic mode: when the user wants to reduce the performance overhead of shape reinferring, and an input of interest enters the graph very close to the output, so that only few tensors need to be reinferred in dynamic mode, while static reinferring would process the whole graph again. But this case will be actual only since we implement static reinferring logic after resize, which is not supported now. And even in that case an explicit flag forcing the dynamic mode would be better.

hyunsik-yoon:
> What was the purpose of PR #3998 then? And what is the reason of changing the mind?

At that time it was added for an app that uses model of, e.g., [unknown, 100] and unknown batch is always 1. So, #3998 provide benefit of performance when batch is always 1. At that time, we didn't deeply thought about the case when input can be shuffled (static -> dynamic -> static). However, recently we see such case and such case made one runtime crashed. So we started to discuss #4625.
After #4625, we can guide app developer that if their model always use batch==1, `set_input_tensorinfo()` should not be called for performance.

> If I get everything right, dynamic/static execution mode is an internal to the API and invisible/transparent to the user. There is no API function to know whether the tensor is dynamic or static.

One of big concerns of app developers (and also model developers) is performance. In many companies including Samsung, there are some strict performance criteria with apps that uses AI.
IMHO some explanation about `dynamic and static tensors` should be written to developers since it can make impact on performance. So I think we'd better mention the following somewhere:

```
set_input_tensorinfo() will make input tensors dynamic, which will degrade the performance a little bit.
Therefore, if a model can be run without calling set_input_tensorinfo(), 
it is recommended not to call set_input_tensorinfo().
```

It is worth discussion explicit `resizing` and shape inference on whole graph after resizing. I see the benefit with inference when input shape changes after some multiple calls of nnfw_run().

krayzemli:
@hyunsik-yoon 

Anyway, I don't see the reason in introducing an optimization first (PR #3998) and then removing it back (rather than fixing a bug in it). Now the user will have to do more programming work in his code to handle this special case explicitly (when, say, batch size is defined by some command line parameter), while it is only our implementation-related problem that resizing input turns on less effective dynamic mode, and it may change in the future, and an explicit optimization the user has made in his code will become obsolete. Also I don't see the reason for now of why the user may want to turn on dynamic inferring explicitly. But even if such a reason existed, this switching should be accomplished by an additional flag parameter or a separate API function, rather than by the fact of resizing an input. Teaching a user to rely on such a fact may introduce backward-compatibility problems in the future.

hyunsik-yoon:
please refer to my comment at https://github.com/Samsung/ONE/issues/4625#issuecomment-715267688

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -492,27 +492,26 @@ NNFW_STATUS nnfw_session::apply_tensorinfo(uint32_t index, nnfw_tensorinfo ti)
...
+    onert::ir::Shape new_shape(ti.rank);
+    for (int32_t i = 0; i < ti.rank; i++)
+      new_shape.dim(i) = ti.dims[i];
 
     // overwrite input shape with the shape from ti
     input.info().shape(new_shape);
   }
   else // when called after nnfw_session::prepare()
   {
+    onert::ir::Shape new_shape(ti.rank);
```
krayzemli, line 511: new_shape calculation is duplicate in both if branches, it better be moved outside.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -492,27 +492,26 @@ NNFW_STATUS nnfw_session::apply_tensorinfo(uint32_t index, nnfw_tensorinfo ti)
...
+    for (int32_t i = 0; i < ti.rank; i++)
+      new_shape.dim(i) = ti.dims[i];
+
     _execution->changeInputShape(onert::ir::IOIndex(index), new_shape);
```
krayzemli, line 515: Adding a check `if (_execution->getInputShape(onert::ir::IOIndex{index}) != new_shape)` will be good optimization that prevents activation of dynamic mode when not necessary. Unless it is proved to break something

hyunsik-yoon: Thanks. Fixed.

runtime/onert/api/src/nnfw_api_internal.cc
hyunsik-yoon: Thanks. Fixed.

runtime/onert/api/src/nnfw_api_internal.cc
hyunsik-yoon: I have some concern with this change. Could you please refer to https://github.com/Samsung/ONE/pull/4718#issuecomment-714894837?

One other thing is that what `_execution->changeInputShape(..)`  does is simply entering new shape information into a map inside `IODescription`. IMHO elimination of calling  `_execution->changeInputShape(..)` by adding one more condition (`if (_execution->getInputShape(onert::ir::IOIndex{index}) != new_shape)`) does not provide optimization.

krayzemli: The main purpose of this optimization is to **prevent creating** an item in that map when two conditions are met: a) the tensor is not yet registered in the map of dynamic tensors, b) the new shape is the same as the static shape. Of course, there are many ways to implement that logic, and moving it from nnfw_api to the inside of `changeInputShape` would be even more appropriate. But following the idea of PR #3998, moving the comparison from the wrong place to the right place would fix the bug with minimum code change.

hyunsik-yoon: The purpose of this PR is not to follow the optimization of #3998 but to follow the rule 1) and 2) in https://github.com/Samsung/ONE/issues/4625#issue-721225511.

What you propose is one way to solve current crash but I mentioned in
https://github.com/Samsung/ONE/pull/4718#issuecomment-714894837, it is difficult to communicate. 

I tried to write down how I will communicate if I accept your suggestion and it is complex:
- If input shape is changed, ONERT treats the input tensor as dynamic
- But once the input tensor becomes dynamic, the input will remain dynamic if the shape changes. (see 5th run in https://github.com/Samsung/ONE/pull/4718#issuecomment-714894837)

Also, IMHO another drawback with your suggestion is that... when user creates a model with Tensorflow with shape e.g., `(None, 100)`, converted tflite file will have `(1, 100)` shape. However, users won't know the `None` will be converted to to `1` in tflite file. So, they will have hard time to understand why the shape was not changed with 2nd run and it is static... and why 4, 5th run is dynamic. It is too difficult to understand.

So, IMHO, I think rules in https://github.com/Samsung/ONE/issues/4625#issue-721225511 and the change in this PR is better for API.

hyunsik-yoon:
PTAL

hyunsik-yoon:
@wateret @krayzemli
PTAL

krayzemli:
LGTM, though I'd prefer an additional optimization in `changeInputShape`

hyunsik-yoon:
@krayzemli Thanks for the review. Could you approve? (you didn't approve and added only a comment)

@wateret PTAL

krayzemli (approved):

wateret (approved):
LGTM

PR #4724, opened 2020-10-21
[OneOpTest] Quantization tensor condition

hseok-oh:
- Support INT8 type
- Restrict zero-point and scale


wateret (approved):
LGTM

PR #4627, opened 2020-10-14
[onert] DRAFT: Enable combinations of nnfw_set_input_tensorinfo() and nnfw_run()

hyunsik-yoon:
DRAFT/WIP

This is a draft to enable combinations of nnfw_set_input_tensorinfo() and nnfw_run().

parent issue: #4625

Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>

PR #4738, opened 2020-10-22
[onert] make new_shape param optional in CircleGen::addOperatorReshape

hyunsik-yoon:
This makes `new_shape`, the second of `CircleGen::addOperatorReshape` optional.

When optional, `new_shape` will be `nullptr` just like
`circle::CreateReshapeOptionsDirect(.., const std::vector<int32_t> *new_shape = nullptr)`
method.

Originated from draft #4627

### DIFF
--- a/tests/nnfw_api/src/GenModelTests.cc
+++ b/tests/nnfw_api/src/GenModelTests.cc
@@ -193,3 +193,99 @@ TEST_F(GenModelTest, OneTensor_ConstAndThreeOutputs)
 
   SUCCEED();
 }
+
+TEST_F(GenModelTest, Reshape_with_shape_param_as_const)
+{
+  CircleGen cgen;
+  auto i32 = circle::TensorType::TensorType_INT32;
+
+  int input = cgen.addTensor({{4}, i32});
+
+  std::vector<int32_t> new_shape_data{2, 2}; // const of value [2, 2]
+  uint32_t new_shape_buf = cgen.addBuffer(new_shape_data);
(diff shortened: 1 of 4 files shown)


tests/nnfw_api/src/GenModelTests.cc
```diff
@@ -193,3 +193,99 @@ TEST_F(GenModelTest, OneTensor_ConstAndThreeOutputs)
 
   SUCCEED();
 }
+
+TEST_F(GenModelTest, Reshape_with_shape_param_as_const)
```
wateret, line 197: What about removing `param_` in the test name? "shape param" sounds like it is an operation's parameter which is not. a tensor, to me.

hyunsik-yoon: What I meant was param of Reshape op.

in https://github.com/Samsung/ONE/pull/4738/files#diff-4ee790f1f7d02d361d6728f1b1c16729ff0f9afffbed34de279a2a3a3edb482bR211, 

I meant the following **bold** param.
cgen.addOperatorReshape({{input, new_shape}, {out}}, **&new_shape_data**); 

Since the main change of this PR is about handling param as an option (if omitted, param shape is `nullptr`)
I intended to add two test cases: one with param, another without param.

wateret: Oh thank you for explanation! I forgot Reshape has "new shape" as a param as well as an input.

wateret (approved):
LGTM

PR #4741, opened 2020-10-22
[onert] Refactoring MoelTestDynamicTensor.cc

hyunsik-yoon:
This refactors `MoelTestDynamicTensor.cc`:
1. Rename `set_input_output` to `setInputOutput` (came case)
2. Make `setInputOutput()` accept template inputs
3. Create `verifyOutput()` func

Originated from draft #4627
Excerpted from #4718

### DIFF
--- a/tests/nnfw_api/src/ModelTestDynamicTensor.cc
+++ b/tests/nnfw_api/src/ModelTestDynamicTensor.cc
@@ -33,7 +33,7 @@
   {                                                                    \
     if (result != NNFW_STATUS_NO_ERROR)                                \
       throw std::runtime_error("returning error on calling nnfw api"); \
-  } while (false);
+  } while (false)
 
 template <class CPP_TYPE> struct nnfw_type;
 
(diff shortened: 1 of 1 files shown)


tests/nnfw_api/src/ModelTestDynamicTensor.cc
```diff
@@ -22,31 +22,89 @@
...
+// if there's any error while setting input or output:
+//
+//  EXPECT_ANY_THROW(setInputOutput(...));
+//
+#define THROW_WHEN_NNFW_ERROR(result)                                  \
+  do                                                                   \
+  {                                                                    \
+    if (result != NNFW_STATUS_NO_ERROR)                                \
+      throw std::runtime_error("returning error on calling nnfw api"); \
+  } while (false);
```
wateret, line 36: I would like to suggest 

```suggestion
  } while (false)
```

I think it is more common used without `;`. (to force developers always put `;` when they use this macro)

wateret:
👍 I think some functions can be reused from other tests, later.

tests/nnfw_api/src/ModelTestDynamicTensor.cc
hyunsik-yoon: super!

hyunsik-yoon:
PTAL

wateret (approved):
LGTM

PR #4747, opened 2020-10-22
[onert] Use atomic finished flag in the Execution

krayzemli:
Investigating the runtime code, I noticed that bool finished flag in Execution may be accessed from two unsynchronized threads at the same time. That should be safe in practice in this particular case, but not correct in theory.

I made the flag atomic, added flag reset in startAsync, and also added some async-related runtime checks.

What do you think?

### DIFF
--- a/runtime/onert/core/src/exec/Execution.cc
+++ b/runtime/onert/core/src/exec/Execution.cc
@@ -32,6 +32,15 @@ Execution::Execution(const std::shared_ptr<ExecutorMap> &executors) : _executors
   _io_desc.outputs.resize(primary_subg.getOutputs().size());
 }
 
+void Execution::assertAsyncInactive()
+{
+  assert(!_exec_thread || !_exec_thread->joinable());
+  if (_exec_thread && _exec_thread->joinable())
+    waitFinish();
+}
+
+Execution::~Execution() { assertAsyncInactive(); }
+
 void Execution::changeInputShape(const ir::IOIndex &index, const ir::Shape &new_shape)
 {
   // This will be used later to set input tensor dynamic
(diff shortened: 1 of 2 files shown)


hseok-oh:
Could you tell us scenario when `Execution` instance is accessed by two (or more) threads? `Execution` is designed to be accessed from only one thread for inference. `Executor` is designed to be accessed from many threads.

krayzemli:
The theoretical case is calling `isFinished` or `getOutputShape` between `startExecute` and `waitFinish`. This is not possible when using the API, but only when `Execution` is used directly. Another possible problem fixed with this PR is a crash that may happen if a session is closed without calling `waitFinish` after an asynchronous execution. These are all abnormal cases, but no checks are made to prevent them.

hseok-oh:
@krayzemli 

>Another possible problem fixed with this PR is a crash that may happen if a session is closed without calling waitFinish after an asynchronous execution. These are all abnormal cases, but no checks are made to prevent them.

Handling this case looks reasonable as you commented, IMO.

hseok-oh:
@wateret @krayzemli I found out `nnfw_api` reuses `Execution` instance if `nnfw_run` is called many times. But it isn't intended usage. 

IMHO, to reuse `Execution` instance, `Execution` needs some changes
- Status variable: `prepare` -> `executing` -> `finished` 
  - Atomic variable as @krayzemli 's change in this PR
- Block calling `setInputLayout` and `setOutputLayout`, `setInput`, `setOutput`, `changeInputShape` between `startExecute` and `waitFinish`
- `reset` API to prepare inference again
  - `reset` if `setInputLayout`, `setOutputLayout`, `setInput`, `setOutput`, `changeInputShape`, `execute`, `startExecute` is called

hseok-oh:
@wateret @krayzemli Please forgot above comment. `nnfw_api` already handle internal status.

hseok-oh:
`nnfw_api` reuses `Execution` instance if `nnfw_run` is called many times. But it isn't intended usage (follow `ANeuralnetworksExecution`).

There is two choice

- Restrict `Execution`'s usage 
  - Disallow to call `Execution`'s API except getters when execution finished
  - Need to change `nnfw_api_internal`'s implementation
- Change to reuse `Execution`
  - Handle status in `Execution`
  - Restrict to call `Execution`'s API except getters anyway, but support `reset` API to reuse `Execution`

hseok-oh:
My opinion

- `std::atomic` type `_finished` looks good
- Introducing `assertAsyncInactive` and calling on destructor looks good
- Let's discuss on other place how to handle `Execution`'s status

runtime/onert/core/src/exec/Execution.cc
```diff
@@ -32,6 +32,15 @@ Execution::Execution(const std::shared_ptr<ExecutorMap> &executors) : _executors
   _io_desc.outputs.resize(primary_subg.getOutputs().size());
 }
 
+void Execution::assertAsyncInactive()
+{
+  assert(!_exec_thread || !_exec_thread->joinable());
+  if (_exec_thread && _exec_thread->joinable())
```
wateret, line 38: Here it checks the same condition right after an assertion. we usually don't do something like this, as assertions works only for debug mode and used to catch ONE developer's fault early, not the user's. Can I ask what your intention is here?

runtime/onert/core/src/exec/Execution.cc
```diff
@@ -32,6 +32,15 @@ Execution::Execution(const std::shared_ptr<ExecutorMap> &executors) : _executors
   _io_desc.outputs.resize(primary_subg.getOutputs().size());
 }
 
+void Execution::assertAsyncInactive()
```
wateret, line 35: This name looks misleading. Should be `ensureWaitFinish` or something.

runtime/onert/core/src/exec/Execution.cc
krayzemli: Intention is to trigger assert in debug mode and silently fix the problem in release mode.

hseok-oh:
Outdated

PR #4748, opened 2020-10-22
[onert] Bugfix in nnfw_set_tensorinfo() 

krayzemli:
Compare new shape with the runtime shape rather than with a model default shape n the logic of avoiding dynamic shape infer mode.

This is a fix of the bug introduced in PR #3998.

Related: PR #4718, issue #4625


seanshpark:
> Related: PR #4718, issue #4625

I don't think it is nice to post another persons work as another PR work in progress.
I would be better to comment as `change` so that you can also contribute to that PR.

krayzemli:
Okay, let it be the change.

PR #4484, opened 2020-10-05
[Draft] Use ruy library for FP32 fully connected operation 

periannath:
This commit uses ruy library for FP32 fully connected operation

ONE-DCO-1.0-Signed-off-by: JiHwan Yeo <jihwan.yeo@samsung.com>

---

For issue #4482

PR #4749, opened 2020-10-23
[onert] Remove the calls of CalculateActivationRange in FullyConnected

glistening:
It removes the calls of CalculateActivationRange in FullyConnected.
Because the calculated float_activation_{min,max} are not used any longer.


periannath:
`float_activation_{min,max}` might be used by later kernel implementations. Ruy library uses those values for FullyConnected FP32 computation #4484.

It would be better to comment out `float_activation_{min,max}` inside `Types.h`.

ragmani (approved):
LGTM

glistening:
@periannath I've updated as you commented.

periannath (approved):
LGTM

Issue #4414, opened 2020-09-24
[onert] Optimizes PermuteLayer of controlflow


ragmani:
Now, there are two reasons why `PermuteLayer` has a bit poor performance.

#### Problems
1. `PermuteLayer` does not use multi-threading.
2. For accessing memory of "arm_compute::ICLTensor", runtime must call `arm_compute::ICLTensor::map()`. Our runtime calls the `arm_compute::ICLTensor::map()` with the blocking flag set to always true. 

#### The way to resolve above problems
This task is going to optimize `PermuteLayer` of `controlflow` by resolving above two reasons.

- [x] 1. To make `PermuteLayer` uses multi-threading by using thread pool of `ruy::Context`.
- [x] 2. To make our runtime calls `clEnqueueWriteBuffer()` or `clEnqueueReadBuffer()` instead of `arm_compute::ICLTensor::map()` when output or intput of `PermuteLater` is tensor of `acl_cl` backend.
- [x] 3. To cache offsets of tensors in PermuteLayer

Draft PR : #4395 

/cc @Samsung/nnfw_committers @periannath

hyunsik-yoon:
I have one curiosity. Can we use the benefit of verctorization when we copy memory?

ragmani:
@hyunsik-yoon 
> I have one curiosity. Can we use the benefit of verctorization when we copy memory?

I think it's a good point. But I don't have a plan to use it yet for the following reason.
We can use vectorization (SIMD). But I'm not sure if it will always be benefit of performance. To use vectorization(SIMD), we have to use instructions of load and store. It is depending on how big and fast the caches are in target device. In the other words, using SIMD for copying memory could decrease performance than `memcpy()`, contrary to our expectations.

ragmani:
>  2. To make our runtime calls arm_compute::ICLTensor::map() by distinguishing how to set the blocking flag value.

I tried to change using `map()` to use `clEnqueueMapBuffer()` for only writing CLTensors. But I failed it because there is no way how to get memory buffer without mapping. The function always map data to buffer even though I try all the available flags.
So, I'm going to try to use `clEnqueueWriteBuffer()` instead of `clEnqueueMapBuffer()`.

ragmani:
#### Effects
> 1. To make `PermuteLayer` uses multi-threading by using thread pool of `ruy::Context`.

Depending on some conditions such as device or model, using multi-threading can improve or can worsens performance. So we have to set the number of threads well.

[...]

PR #4395, opened 2020-09-23
[Draft][onert] Optimizes PermuteLayer 

ragmani:

PR #4751, opened 2020-10-23
[onert] Optimizes PermuteLayer of controlflow by ruy

ragmani:
For issue #4414
Draft PR #4395

This commit optimizes PermuteLayer of controlflow by using ruy's thread pool.

### DIFF
--- a/runtime/onert/core/src/backend/controlflow/ExternalContext.h
+++ b/runtime/onert/core/src/backend/controlflow/ExternalContext.h
@@ -19,7 +19,11 @@
 
 #include <backend/IExternalContext.h>
 #include <util/ConfigSource.h>
+
 #include <ruy/context.h>
+#include <ruy/context_get_ctx.h>
+#include <ruy/ctx.h>
+#include <ruy/tune.h>
 
 namespace
 {
@@ -37,9 +41,10 @@ namespace controlflow
 class ExternalContext : public IExternalContext
 {
 public:
-  ExternalContext() : _ruy_context(nullptr)
+  ExternalContext() : _ruy_context(std::make_unique<ruy::Context>())
   {
-    // setMaxNumThreads(onert::util::getConfigInt(onert::util::config::RUY_THREADS));
+    setMaxNumThreads(onert::util::getConfigInt(onert::util::config::RUY_THREADS));
+    initPerThreadState();
   }
 
   void setMaxNumThreads(int max_num_threads)
@@ -50,6 +55,19 @@ public:
   }
 
   ruy::Context *ruy_context() const { return _ruy_context.get(); }
+
+private:
+  void initPerThreadState()
+  {
+    // Initialize per-thread state.
+    const int thread_count = _ruy_context->max_num_threads();
+    auto ctx = ruy::get_ctx(_ruy_context.get());
+    ctx->EnsureThreadSpecificResources(thread_count);
+    for (int i = 0; i < thread_count; i++)
+    {
+      ctx->GetThreadSpecificTuningResolver(i)->SetTuning(ctx->explicit_tuning());
+    }
+  }
 
 private:
   const std::unique_ptr<ruy::Context> _ruy_context;
--- a/runtime/onert/core/CMakeLists.txt
+++ b/runtime/onert/core/CMakeLists.txt
@@ -2,6 +2,8 @@ file(GLOB_RECURSE SOURCES "src/*.cc")
 file(GLOB_RECURSE TESTS "*.test.cc")
 list(REMOVE_ITEM SOURCES ${TESTS})
 
+nnfw_find_package(Ruy REQUIRED)
+
 add_library(onert_core SHARED ${SOURCES})
 set_target_properties(onert_core PROPERTIES POSITION_INDEPENDENT_CODE ON)
 target_include_directories(onert_core PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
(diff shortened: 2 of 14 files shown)


runtime/onert/core/src/backend/controlflow/ExternalContext.h
```diff
@@ -19,7 +19,13 @@
 
 #include <backend/IExternalContext.h>
 #include <util/ConfigSource.h>
+
+#ifdef USE_RUY_GEMV
```
periannath, line 23: `USE_RUY_GEMV` macro is defined to check whether or not to use ruy library for hybrid fully connected layer. This macro might be undefined even if ruy library is supported.

Ruy library is currently built on every targets, I think we can remove this macro check. Or it would be better to define a new macro which becomes true when ruy library is supported and use that macro here.

periannath (commented):
This PR is too large, so I checked codes related to ruy library and external context.

runtime/onert/core/src/backend/controlflow/ExternalContext.h
ragmani: Thanks for your information. How about `USE_RUY` as new macro?

runtime/onert/core/src/backend/controlflow/ExternalContext.h
```diff
@@ -36,10 +42,11 @@ namespace controlflow
 // TODO Unify this with cpu::ExternalContext
 class ExternalContext : public IExternalContext
 {
+#ifdef USE_RUY_GEMV
 public:
-  ExternalContext() : _ruy_context(nullptr)
+  ExternalContext() : _ruy_context(std::make_unique<ruy::Context>())
   {
-    // setMaxNumThreads(onert::util::getConfigInt(onert::util::config::RUY_THREADS));
+    setMaxNumThreads(onert::util::getConfigInt(onert::util::config::RUY_THREADS));
```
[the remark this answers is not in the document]
ragmani: Okay, I will do so.

runtime/onert/core/src/backend/controlflow/ExternalContext.h
periannath: @ragmani `USE_RUY` looks good to me!

ragmani: Okay, I will do so.

runtime/onert/core/src/backend/controlflow/ExternalContext.h
```diff
@@ -36,10 +42,12 @@ namespace controlflow
 // TODO Unify this with cpu::ExternalContext
 class ExternalContext : public IExternalContext
 {
+#ifdef USE_RUY
```
hyunsik-yoon, line 45: before offline talk, I thought that we use RUY only for calculation. In here, we use RUY for thread library, not calculation. If you comment that here, it would be easier to read the code.

runtime/onert/core/src/backend/controlflow/ExternalContext.h
hyunsik-yoon: If we always build RUY when we build ONERT, I wonder if we need to use `#ifdef USE_RUY` . Could this line be removed?

runtime/onert/core/src/backend/controlflow/ExternalContext.h
ragmani: In here, that identifier(USE_RUY) prevents build failure by `ruy::Context` when that identifier was not defined. And the `ruy::Context` must be required for using RUY. I think it seem not to be related to code readability.

runtime/onert/core/src/backend/controlflow/ExternalContext.h
ragmani: I think `onert` have to support `PermuteLayer` without using ruy. I use `#ifdef USE_RUY` to support it.

wateret:
@periannath @ragmani 

Would we need `USE_RUY` flag? I mean, unless there is some environment that ruy is not available, it would be nice to have only one implement with `ruy`.

ragmani:
> Would we need USE_RUY flag? I mean, unless there is some environment that ruy is not available, it would be nice to have only one implement with ruy.

If there is no any environment that ruy is not available, there is no reason why `ExternalContext` wraps `Ruy::Context`. If we are going to keep only the implementation with `ruy`, it would be better to remove `ExternalContext`.

periannath:
@hyunsik-yoon @wateret 
After offline discussion with @ragmani, we agreed with the opinion that we don't need `USE_RUY` flag.

periannath:
(Optional) In this PR, both `cpu` and `controlflow` backend have their own `ruy::Context`. If we set `RUY_THREADS` to 4, onert will create 8 threads (4 for `cpu` and 4 for `controlflow`).
If both backend can share one `ruy::Context`, we can reduce the number of ruy threads.

wateret:
> If there is no any environment that ruy is not available, there is no reason why `ExternalContext` wraps `Ruy::Context`. If we are going to keep only the implementation with `ruy`, it would be better to remove `ExternalContext`.

> After offline discussion with @ragmani, we agreed with the opinion that we don't need `USE_RUY` flag. Since ruy library is supported on all targets, we don't have to consider the environment that doesn't have ruy library.

Both sound great to me 😃

ragmani:
I removed USE_RUY flag.

infra/nnfw/cmake/packages/Ruy/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ list(APPEND RUY_INCLUDES "${RuySource_DIR}")
 add_library(ruy STATIC ${RUY_SRCS})
 target_include_directories(ruy SYSTEM PUBLIC ${RUY_INCLUDES})
 target_compile_options(ruy PRIVATE -O3)
+target_compile_definitions(ruy PUBLIC USE_RUY)
```
periannath, line 31: @ragmani I think this line should be removed.
```suggestion
```

ragmani: Thanks. I will remove it.

periannath (approved):
LGTM

wateret (approved):
LGTM

hseok-oh (approved):

hyunsik-yoon (approved):

PR #4754, opened 2020-10-23
[luci] Model size optimization when exporting model

llFreetimell:
This commit will decrease model size by following optimizations
- Remove redundant empty buffers
- Remove redundant buffers with same values

### DIFF
--- a/compiler/luci/export/src/CircleTensorExporter.cpp
+++ b/compiler/luci/export/src/CircleTensorExporter.cpp
@@ -388,8 +388,7 @@ bool has_same_values(luci::CircleConst *lhs, luci::CircleConst *rhs)
   return true;
 }
 
-uint32_t get_buffer_id(FlatBufferBuilder &builder, SerializedModelData &md,
-                             luci::CircleConst *node)
+uint32_t get_buffer_id(FlatBufferBuilder &builder, SerializedModelData &md, luci::CircleConst *node)
 {
   if (node != nullptr)
   {
--- a/compiler/luci/export/src/CircleTensorExporter.cpp
+++ b/compiler/luci/export/src/CircleTensorExporter.cpp
@@ -343,6 +343,82 @@ encodeSparsityParameters(FlatBufferBuilder &builder, luci::SparsityParam *sparsi
                                                 &sparsityparam->block_map, &dim_metadata_vec);
 }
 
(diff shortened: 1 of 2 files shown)


jinevening (dismissed):
LGTM

One suggestion is to change the function name  `optimized_buffer_id`. IMHO, just `get_buffer_id` or `generate_buffer_id` looks better.

jinevening (approved):
Thanks!

seanshpark:
What model have you tested with this? Have you tested the model with OneRT?

compiler/luci/export/src/CircleTensorExporter.cpp
```diff
@@ -351,16 +426,11 @@ void exportOpDefinedTensor(const CircleTensoInfo &info, FlatBufferBuilder &build
...
-  auto buffer =
-      info.content() == nullptr ? encodeOpBuffer(builder) : encodeOpBuffer(builder, info.content());
-
   auto quantparam = encodeQuantizationParameters(builder, info.quantparam());
 
   auto sparsityparam = encodeSparsityParameters(builder, info.sparsityparam());
 
-  auto buffer_id = static_cast<uint32_t>(md._buffers.size());
-  md._buffers.push_back(buffer);
+  auto buffer_id = get_buffer_id(builder, md, info.content());
```
mhs4670go, line 433: @llFreetimell 
I'm not sure but as original code said, buffer is created even if `info.content()` is `nullptr`. But changed code doesn't. Is this one of what you intended?

llFreetimell: Yes it was intended.
When `info.content()` is `nullptr`, empty buffer is newly generated every time.
However, it can be substituted by buffer id 0, which is representative empty buffer.

llFreetimell:
@seanshpark I tested it with internal models, and ran well in ONE runtime.

mhs4670go (approved):
LGTM

compiler/luci/export/src/CircleTensorExporter.cpp
seanshpark: > However, it can be substituted by buffer id 0, which is representative empty buffer.

I'm not sure this is OK but I'll go with your decision.

seanshpark (approved):
LGTM

PR #4758, opened 2020-10-23
[DRAFT] Inroduce one-cmds python interface

mhs4670go:
This draft is to introduce python interface to one-cmds.

Related test will be added soon.

Related : #4278 


seanshpark:
Although all commands are python scripts, I've removed `.py` and wanted to act as a program without extension.
Is there any particular reason to add `.py` extension?

compiler/one-cmds/one_build.py
```diff
@@ -0,0 +1,118 @@
...
+        action='store_true',
+        help='show program\'s version number and exit')
+
+    # configuration file
+    config_group = parser.add_argument_group('configuration arguments')
+    config_group.add_argument(
+        '-C', '--config', type=str, help='run with configuation file')
+    config_group.add_argument(
+        '-S',
+        '--section',
```
seanshpark, line 43: Do we really need this to get from the user?

compiler/one-cmds/one_build.py
```diff
@@ -0,0 +1,118 @@
...
+    """verify given arguments"""
+    # check if required arguments is given
+    if not _utils._is_valid_attr(args, 'config'):
+        parser.error('-C/--config argument is required')
+
+
+def _get_driver_name(driver_name):
+    return {
+        'one-import-bcq': 'one_import_bcq.py',
+        'one-import-tf': 'one_import_tf.py',
```
seanshpark, line 69: Why do you leave `one-import-tf` if you add `one_import_tf.py` ?

compiler/one-cmds/one_build.py
```diff
@@ -0,0 +1,118 @@
...
+        'one-import.tflite': 'one_import_tflite.py',
```
seanshpark, line 70: where did `one-import.tflite` came from? isn't it `one-import-`tflite ?

compiler/one-cmds/one_build.py
mhs4670go: @seanshpark There are two options for users. If `--section` option is given, just run the section. If option is not given, all section in config file will be run sequentially.

seanshpark: > If option is not given, all section in config file will be run sequentially.

This means, config file is not just provides static information but a control information from `run sequentially.`.
I don't this is a good idea. It would be better not to include this implict behaviour.

seanshpark: In another point, users may want to get controllability of execution sequence somehow . Let's think about this...

mhs4670go:
```bash
$ ./runtestall.sh
one-import-bcq_001.test SUCCESS
one-import-bcq_neg_001.test SUCCESS
one-import-bcq_neg_002.test SUCCESS
one-import-bcq_neg_003.test SUCCESS
one-import-bcq_neg_004.test SUCCESS
one-import-bcq_neg_005.test SUCCESS
one-import-bcq_neg_006.test SUCCESS
one-import-bcq_neg_007.test SUCCESS
one-import-bcq_neg_008.test SUCCESS
one-import-bcq_neg_009.test SUCCESS
one-import_001.test SUCCESS
one-import_neg_001.test SUCCESS
one-import_neg_002.test SUCCESS
one-import_neg_003.test SUCCESS
one-import_neg_004.test SUCCESS
one-import_neg_005.test SUCCESS
one-import_neg_006.test SUCCESS
one-import_neg_007.test SUCCESS
one-import_neg_008.test SUCCESS
one-import_neg_009.test SUCCESS
one-import_neg_010.test SUCCESS
one-optimize_001.test SUCCESS
one-optimize_neg_001.test SUCCESS
one-optimize_neg_002.test SUCCESS
one-optimize_neg_003.test SUCCESS
one-pack_001.test SUCCESS
one-pack_neg_001.test SUCCESS
one-pack_neg_002.test SUCCESS
one-pack_neg_003.test SUCCESS
one-quantize_001.test SUCCESS
one-quantize_neg_001.test SUCCESS
one-quantize_neg_002.test SUCCESS
one-quantize_neg_003.test SUCCESS
one-quantize_neg_004.test SUCCESS
one-quantize_neg_005.test SUCCESS
one-quantize_neg_006.test SUCCESS
one-quantize_neg_007.test SUCCESS
one-quantize_neg_008.test SUCCESS
one-quantize_neg_009.test SUCCESS
one-quantize_neg_010.test SUCCESS
one-quantize_neg_011.test SUCCESS
one-quantize_neg_012.test SUCCESS
one-quantize_neg_013.test SUCCESS
```

PTAL:)

compiler/bcq-tools/generate_bcq_output_arrays.py
```diff
@@ -233,13 +233,13 @@ def print_bcqinfo_output_arrays_v1(flags):
                 f_arrays.write("," + prefix + "/bcqinfo_qbits_of_clusters")
                 f_arrays.write("," + fuseop)
                 if has_dequant_weight:
-                    f_arrays.write("," + prefix + "/bcqinfo_dequant_weight")
+                    f_arrays.write("," + prefix + "/bcqinfo_1dequant_weight")
```
seanshpark, line 236: there's a `1` in `bcqinfo_1dequant_weight`.

compiler/one-cmds/how-to-use-one-commands.txt
```diff
@@ -20,6 +20,16 @@ Compilation flow for NPU
...
 
 
+Prerequisite
+------------
+
+Before you use one commands, please prepare a python environment with TensorFlow installed.
+You can use `one-prepare-venv`(ubuntu) script where it automatically installs python3 virtual
+environment with Tensorflow-cpu.
+
+If you have prepared Python as a virtual environment, don't forget to activate it.
```
seanshpark, line 30: Previouslly we didn't let users to activate but the shell script did it. Now we use python script and this makes activation not possible (or we don't know how-to).
What I could think of right now...
- add a script to activate and deactivate
- let one-import-tf as-is using bash
- ...

compiler/one-cmds/one-build.cfg
```diff
@@ -0,0 +1,20 @@
+[driver-to-run]
+one-import-tf=True
+one-import-tflite=False
+one-import-bcq=False
+one-optimize=True
+one-quantize=False
+one-pack=False
```
seanshpark, line 7: Let's set to True

compiler/one-cmds/one-build.cfg
```diff
@@ -0,0 +1,20 @@
+[driver-to-run]
```
seanshpark, line 1: How about naming this section as `one-build` ?

seanshpark, line 1: How about naming this filas as `one-build.cfg.template` ?

compiler/one-cmds/one-codegen
```diff
@@ -14,53 +14,74 @@
...
+
+import one_utils as _utils
+
+
+def _get_parser():
+    parser = argparse.ArgumentParser(description='command line tool for code generation')
+
+    _utils._add_default_arg(parser)
+
+    parser.add_argument('-b', '--backend', type=str, help='backend name to use')
```
seanshpark, line 31: Please add the version option

compiler/one-cmds/one-build
```diff
@@ -0,0 +1,123 @@
...
+def main():
+    # parse arguments
+    parser = _get_parser()
+    args = _parse_arg(parser)
+
+    # verify arguments
+    _verify_arg(parser, args)
+
+    # parse configuration file
+    config = _parse_cfg(args)
```
seanshpark, line 96: sequence of `_verify_arg` -> `_parse_cfg`: how does this work if both command line arg and cfg file exist?

compiler/one-cmds/one-codegen
```diff
@@ -14,53 +14,74 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-DRIVER_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-
-function Usage()
-{
-  echo "Usage: one-codegen [BACKEND] ..."
-  echo "Available BACKEND drivers:"
```
seanshpark, line 22: please add this part

seanshpark:
Please add tests that uses `.cfg` files

compiler/bcq-tools/generate_bcq_output_arrays.py
```diff
@@ -255,6 +255,196 @@ def print_bcq_output_arrays(flags):
         raise SystemExit(err_msg)
 
 
+def get_bcq_version(input_path):
```
seanshpark, line 258: Why is this added?

compiler/one-cmds/CMakeLists.txt
```diff
@@ -8,6 +10,7 @@ set(ONE_COMMAND_FILES
     one-pack
     one-codegen
     one-prepare-venv
+    one_utils.py
```
seanshpark, line 13: Adding files in `ONE_COMMAND_FILES` will make it executable. Please add another list for `one-build.cfg` and `one_utils.py`.

seanshpark, line 13: following `one*` for `one_utils.py` seems confusing for command name similarity. can we use another name that doesn't start with `one` ?

compiler/one-cmds/tests/one-import-bcq_neg_003.test
```diff
@@ -21,7 +21,7 @@ filename="${filename_ext%.*}"
 
 trap_err_onexit()
 {
-  if grep -q "input model not found" "${filename}.log"; then
```
seanshpark, line 24: This error message followed https://github.com/Samsung/ONE/issues/4155#issuecomment-690903095
@lemmaa , do we need to keep this or can we update this to follow new tools?

compiler/one-cmds/one_utils.py
```diff
@@ -0,0 +1,137 @@
...
+    parser.add_argument(
+        '-v',
+        '--version',
+        action='store_true',
+        help='show program\'s version number and exit')
+
+    # configuration file
+    parser.add_argument('-C', '--config', type=str, help='run with configuation file')
+    # section name that you want to run in configuration file
+    parser.add_argument('-S', '--section', type=str, help=argparse.SUPPRESS)
```
seanshpark, line 35: do we need this? as we discussed off-line, we could remove this for the first version.

compiler/bcq-tools/generate_bcq_output_arrays.py
mhs4670go: @seanshpark print_* method will be deprecated. Because it returns its output as a file, which is redundant if we use python script. So, get_* methods return its output as python things.

compiler/one-cmds/one-build
mhs4670go: @seanshpark As we've discussed earlier, arg overrides cfg.

compiler/one-cmds/one_utils.py
mhs4670go: @seanshpark if `argparse.SUPPRESS` is given to `help` kwarg, user can't see its option. It is internal or private option for implementation convenience and scalability.

compiler/one-cmds/one-build
seanshpark: Why I asked is funcion call sequence is `_verify_arg` then `_parse_cfg`.
One may think (1) set the `args` with command line (2) override `args` with cfg file.
So, can we swap these?

compiler/bcq-tools/generate_bcq_output_arrays.py
```diff
@@ -255,6 +255,196 @@ def print_bcq_output_arrays(flags):
...
+            ret_output_arrays.append(prefix + '/bcqinfo_size_of_clusters')
+            ret_output_arrays.append(prefix + '/bcqinfo_qbits_of_clusters')
+            ret_output_arrays.append(fuseop)
+            if has_dequant_weight:
+                ret_output_arrays.append(prefix + '/bcqinfo_dequant_weight')
+
+    return ret_output_arrays
+
+
+def get_bcq_output_arrays(input_path, output_arrays):
```
seanshpark, line 432: > Why is this added?

Maybe I should have asked here...  this file is in `bcq-tools` not in `one-cmds`

compiler/one-cmds/one_utils.py
seanshpark: Can user add this in argument and override it? Users can still access the source code.

compiler/one-cmds/one-build
mhs4670go: Unfortunately, I can't. Actually, I have to verify `config` - configuration file path - argument  before parsing configuration file. It seems natural...

seanshpark: Sorry, not `_verify_arg`...
it is about `_parse_arg` -> `_parse_cfg` sequence.

mhs4670go: @seanshpark Right. It might seem unnatural as you said. But as with above, conf file path first has to be given before parsing conf file. So I should parse arg first.

compiler/one-cmds/one_utils.py
mhs4670go: Yes. Umm.. I think there is no demerit even if user finds out this option from source code. If you worries about misbehave like randomly user gives `-S`, we can print warning at this time.

compiler/bcq-tools/generate_bcq_output_arrays.py
mhs4670go: Oh, right. I will split this file when I post a PR.

compiler/one-cmds/tests/one-import-bcq_neg_003.test
seanshpark: Had a off-line talk with @lemma and it is OK to change these :)

compiler/one-cmds/one_utils.py
seanshpark: OK, got it. I think no need to add warning message. Using -`S` is possible but it's up to the user to use unofficial arguments and get misbehaviour.

compiler/one-cmds/one-build.cfg
mhs4670go: @seanshpark How about `one-build.template.cfg`? Editor programs like vim or sth recognize its format as a configuration  file from last extension. 
`one-build.cfg.template`
image
`one-build.template.cfg`
image

compiler/one-cmds/how-to-use-one-commands.txt
mhs4670go: @seanshpark I used this magic:)
```bash
#!/usr/bin/env bash
''''export SCRIPT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"                  # '''
''''export PY_PATH=${SCRIPT_PATH}/venv/bin/python                                       # '''
''''test -f ${PY_PATH} && exec ${PY_PATH} "$0" "$@"                                     # '''
''''echo "Error: Virtual environment not found. Please run 'one-prepare-venv' command." # '''
```
First, it will be run as a bash script. And then, check if python path exists. If it does, execute it. Triple quote will be comment then.

compiler/one-cmds/CMakeLists.txt
```diff
@@ -36,6 +34,34 @@ foreach(ONE_COMMAND IN ITEMS ${ONE_COMMAND_FILES})
           
 endforeach(ONE_COMMAND)
 
+set(ONE_UTILITIY_FIELS
```
seanshpark, line 37: typo `ONE_UTILITIY_FILES` ?

mhs4670go: Oh, Hawk's eye!

compiler/one-cmds/CMakeLists.txt
```diff
@@ -34,17 +34,17 @@ foreach(ONE_COMMAND IN ITEMS ${ONE_COMMAND_FILES})
           
 endforeach(ONE_COMMAND)
 
-set(ONE_UTILITIY_FIELS
+set(ONE_UTILITY_FIELS
```
seanshpark, line 37: You've fixed to remove `I` in `UTILITY`- you have good eyes too
What I saw `ONE_UTILITY_FILES` in `FIELS` -> `FILES`

compiler/one-cmds/how-to-use-one-commands.txt
seanshpark: This is very interesting!

mhs4670go: @seanshpark Actually, there is pros and cons.
`one-import-tf.py` + `one-import` = increase the number of files
`one-import-tf`(bash + python as above) = need manual format check and no ide formatting

PR #4744, opened 2020-10-22
[onert] Fix when PermutationEliminationPass off

wateret:
Fix bugs when PermutationEliminationPass off

ONE-DCO-1.0-Signed-off-by: Hanjoung Lee <hanjoung.lee@samsung.com>

PR #4761, opened 2020-10-23
[onert] Fix dynamic shape inference of ArgMax

wateret:
Fix dynamic shape inference of ArgMax. The shape could change depending
on the value of axis. So if output is dynamic we should do shape
inference again.


wateret:
From #4744

hyunsik-yoon (approved):
LGTM

PR #4769, opened 2020-10-26
DRAFT: Fuse Pre-activation batchnorm pass

jinevening:
On going draft to add FusePreActivationBatchNormPass

---
Related to: #4773
See [this comment](https://github.sec.samsung.net/AIP/Trinity/issues/141#issuecomment-862601) for more details.

### DIFF
--- a/compiler/luci/pass/src/FusePreActivationBatchNormPass.test.cpp
+++ b/compiler/luci/pass/src/FusePreActivationBatchNormPass.test.cpp
@@ -1,0 +1,264 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


compiler/luci/pass/src/FusePreActivationBatchNormPass.test.cpp
```diff
@@ -0,0 +1,264 @@
...
+{
+
+/**
+ *  Simple graph for test
+ *
+ *  BEFORE
+ *
+ *          [Conv]  W + bias
+ *            |
+ *          [Add]
```
seanshpark, line 36: If this is `[Add]`, there should be another input(s) but the diagram doesn't provide them...

jinevening: The pass only requires one CONV, so I made a a test with a single input for simplicity. Another node can be added, but it may not affect the test.

seanshpark: Um... does this work also? (sorry, I drawn it in horz direction)
```
[Conv] --- [Add] --- [Mul] --- [Add+Relu] -- [Conv] --- [Add]
[Conv] --/         \----------------------------------/

jinevening: >  I'm curious if the transformations works if there is two Conv and then Add...

Yes, it works. In that case, `beta/gamma` is added to the bias of one Conv. I'll add one more Conv to show the case.

mhs4670go:
How about adding dredd test?

jinevening:
> How about adding dredd test?

I'll add it in the following PR.

PR #4777, opened 2020-10-26
[onert] Reduce namespace in OperationValidator

hseok-oh:
OperationValidator is moved from onert::compiler to onert::ir


hyunsik-yoon (approved):
LGTM

Issue #4773, opened 2020-10-26
Fuse pre-activation batchnorm


jinevening:
### Background

In inference models (tflite or circle), batchnorm is usually fused with the preceding CONV layer for better performance.

ResNet V2 has non-trivial structure called _pre-activation_ to improve model generalization. (<-> activation after CONV, i.e., post-activation). (paper link: https://arxiv.org/pdf/1603.05027v3.pdf)

![image](https://user-images.githubusercontent.com/5449554/97135420-8d1c3700-1793-11eb-9062-4eed6cb624a9.png)

In post-activation models (a), `tflite_convert` automatically fuses batchnorm into the preceding CONV layer (depicted as `weight`).

In pre-activation models (b), however, it is not simple to fuse batchnorm computations into nearby CONV layers.

### Goal

Introduce a new pass named `FusePreActivationBatchNormPass` that fuses batchnorm computations of pre-activations into nearby CONV layers. And finally remove all batchnorm computations in ResNet V2.

### Progress

- [x] Add `FusePreActivationBatchNormPass` in luci
- [x] Add test (luci)
- [x] Add test (dredd)
- [x] Add a pass to fuse remaining activation functions (`FuseActivationFunctionPass`) #4833
- [x] Add a pass to change negative values in gamma to positive (`MakeBatchNormGammaPositivePass`)
- [ ] Update one-build

CC @periannath

periannath:
### Performance measurement

- Model : ResNet v2

#### Device : S20+ exynos

| Backend | Before fusion | After fusion |
|--|--|--|
| cpu | 478.5 | 469.0
| acl_cl | 143.1 | 134.8

#### Device : odroid-xu4

| Backend | Before fusion | After fusion |
|--|--|--|
| cpu | 1254.1 | 1224.4
| acl_cl | 662.4 | 651.4

jinevening:
I close this issue as all tasks are done.

To sum up, we could successfully fuse 17 of 17 pre-activations in ResNet50 V2.

seanshpark:
> I close this issue as all tasks are done.

last one is not checked :)

jinevening:
> last one is not checked :)

I forgot to check the box. Thanks! :)

periannath:
## Performance comparison of quantized model

- Model : ResNet v2
- CPU doesn't support RELU for uint8

#### Device : S20+ exynos

| Backend | Before fusion | After fusion | Speedup |
|--|--|--|--|
[...]

PR #4780, opened 2020-10-26
[luci] Pass to fuse pre-activation batchnrom

jinevening:
This adds a pass to fuse pre-activation batchnorm

---
Related to: #4773 
Draft PR: #4769

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/FusePreActivationBatchNormPass.h
+++ b/compiler/luci/pass/include/luci/Pass/FusePreActivationBatchNormPass.h
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_FUSE_PRE_ACTIVATION_BATCH_NORM_PASS_H__
+#define __LUCI_FUSE_PRE_ACTIVATION_BATCH_NORM_PASS_H__
+
+#include <logo/Pass.h>
+#include <luci/IR/CircleNodes.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to fuse batch normalization of pre-activation
+ */
+struct FusePreActivationBatchNormPass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::FusePreActivationBatchNormPass"; }
+
+  bool run(loco::Graph *g) final;
+
+  std::vector<luci::CircleMul *> _mul_list;
+  std::vector<luci::CircleAdd *> _add_list;
+  std::vector<luci::CircleSub *> _sub_list; // inserted during fusion
+};
+
+} // namespace luci
+
+#endif // __LUCI_FUSE_PRE_ACTIVATION_BATCH_NORM_PASS_H__
--- a/compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
+++ b/compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
@@ -41,23 +41,21 @@ bool is_non_negative(const luci::CircleConst *node)
 bool is_batchnorm_mul(const luci::CircleMul *mul, luci::CircleNode *&pred_node,
                       luci::CircleConst *&gamma)
 {
-  auto x = loco::must_cast<luci::CircleNode *>(mul->x());
-  auto y = loco::must_cast<luci::CircleNode *>(mul->y());
+  auto x = dynamic_cast<luci::CircleConst *>(mul->x());
+  auto y = dynamic_cast<luci::CircleConst *>(mul->y());
 
   luci::CircleNode *pred = nullptr;
   luci::CircleConst *constant = nullptr;
 
-  if (x->opcode() == luci::CircleOpcode::CIRCLECONST &&
-      y->opcode() != luci::CircleOpcode::CIRCLECONST)
+  if (x != nullptr && y == nullptr)
   {
-    pred = y;
-    constant = loco::must_cast<luci::CircleConst *>(x);
+    pred = loco::must_cast<luci::CircleNode *>(mul->y());
+    constant = x;
   }
-  else if (x->opcode() != luci::CircleOpcode::CIRCLECONST &&
-           y->opcode() == luci::CircleOpcode::CIRCLECONST)
+  else if (x == nullptr && y != nullptr)
   {
-    pred = x;
-    constant = loco::must_cast<luci::CircleConst *>(y);
+    pred = loco::must_cast<luci::CircleNode *>(mul->x());
+    constant = y;
   }
   else
   {
@@ -125,12 +123,11 @@ namespace luci
  *  Swap MUL/ADD if they are from batch normalization
  *
  *  BEFORE
- *
  *           [Mul]  gamma
  *             |
  *        [Add + Relu]  beta
- *  AFTER
  *
+ *  AFTER
  *           [Add]  beta/gamma
  *             |
  *           [Mul]  gamma
@@ -155,7 +152,7 @@ bool swap_mul_add(luci::CircleAdd *add, std::vector<luci::CircleMul *> &mul_list
     return false;
 
   if (beta->dtype() != loco::DataType::FLOAT32 || gamma->dtype() != loco::DataType::FLOAT32)
-    throw std::runtime_error("FusePreActivationBatchNormPass only supports fp32 model");
+    throw std::runtime_error("FusePreActivationBatchNormPass only supports Float32 model");
 
   if (!is_non_negative(gamma))
     return false;
--- a/compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
+++ b/compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
@@ -1,0 +1,243 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FusePreActivationBatchNormPass.h"
+#include "FusePreActivationBatchNormPassInternal.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/Log.h>
+
+namespace
+{
+
+// Check if all elements are non-negative
+bool is_non_negative(const luci::CircleConst *node)
+{
+  assert(node->dtype() == loco::DataType::FLOAT32);
+
+  uint32_t size = node->size<loco::DataType::FLOAT32>();
+  for (uint32_t i = 0; i < size; ++i)
+  {
+    if (node->at<loco::DataType::FLOAT32>(i) < 0)
+      return false;
+  }
+  return true;
+}
+
+// Check if mul is batchnorm mul
+bool is_batchnorm_mul(const luci::CircleMul *mul, luci::CircleNode *&pred_node,
+                      luci::CircleConst *&gamma)
+{
+  auto x = loco::must_cast<luci::CircleNode *>(mul->x());
+  auto y = loco::must_cast<luci::CircleNode *>(mul->y());
+
+  luci::CircleNode *pred = nullptr;
+  luci::CircleConst *constant = nullptr;
+
+  if (x->opcode() == luci::CircleOpcode::CIRCLECONST &&
+      y->opcode() != luci::CircleOpcode::CIRCLECONST)
+  {
+    pred = y;
+    constant = loco::must_cast<luci::CircleConst *>(x);
+  }
+  else if (x->opcode() != luci::CircleOpcode::CIRCLECONST &&
+           y->opcode() == luci::CircleOpcode::CIRCLECONST)
+  {
+    pred = x;
+    constant = loco::must_cast<luci::CircleConst *>(y);
+  }
+  else
+  {
+    return false;
+  }
+
+  if (constant->rank() != 1)
+    return false;
+
+  auto channel_dim = constant->dim(0);
+  if (!(channel_dim == mul->dim(mul->rank() - 1)))
+    return false;
+
+  pred_node = pred;
+  gamma = constant;
+  return true;
+}
+
+// Check if add is batchnorm add
+bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::CircleConst *&beta)
+{
+  auto x = loco::must_cast<luci::CircleNode *>(add->x());
+  auto y = loco::must_cast<luci::CircleNode *>(add->y());
+
+  luci::CircleMul *pred = nullptr;
+  luci::CircleConst *constant = nullptr;
+
+  if (add->fusedActivationFunction() != luci::FusedActFunc::RELU)
+    return false;
+
+  if (x->opcode() == luci::CircleOpcode::CIRCLECONST && y->opcode() == luci::CircleOpcode::MUL)
+  {
+    pred = loco::must_cast<luci::CircleMul *>(y);
+    constant = loco::must_cast<luci::CircleConst *>(x);
+  }
+  else if (x->opcode() == luci::CircleOpcode::MUL && y->opcode() == luci::CircleOpcode::CIRCLECONST)
+  {
+    pred = loco::must_cast<luci::CircleMul *>(x);
+    constant = loco::must_cast<luci::CircleConst *>(y);
+  }
+  else
+  {
+    return false;
+  }
+
+  if (constant->rank() != 1)
+    return false;
+
+  auto channel_dim = constant->dim(0);
+  // Assumption: Layout is channel-last
+  if (!(channel_dim == add->dim(add->rank() - 1)))
+    return false;
+
+  mul = pred;
+  beta = constant;
+  return true;
+}
+
+} // namespace
+
+namespace luci
+{
+
+/**
+ *  Swap MUL/ADD if they are from batch normalization
+ *
+ *  BEFORE
+ *
+ *           [Mul]  gamma
+ *             |
+ *        [Add + Relu]  beta
+ *  AFTER
+ *
+ *           [Add]  beta/gamma
+ *             |
+ *           [Mul]  gamma
+ *             |
+ *           [Relu]
+ */
+bool swap_mul_add(luci::CircleAdd *add, std::vector<luci::CircleMul *> &mul_list,
+                  std::vector<luci::CircleAdd *> &add_list)
+{
+  luci::CircleNode *pred_node = nullptr;
+  luci::CircleMul *mul = nullptr;
+  luci::CircleConst *beta = nullptr;
+  luci::CircleConst *gamma = nullptr;
+
+  if (!is_batchnorm_add(add, mul, beta))
+    return false;
+
+  if (loco::succs(mul).size() != 1)
+    return false;
+
+  if (!is_batchnorm_mul(mul, pred_node, gamma))
+    return false;
+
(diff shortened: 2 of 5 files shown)


jinevening:
See https://github.sec.samsung.net/AIP/Trinity/issues/141 for more details.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FusePreActivationBatchNormPass.h"
+#include "FusePreActivationBatchNormPassInternal.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/Log.h>
+
+#include <iostream>
```
mhs4670go, line 23: I'm not sure but maybe it is traces of debugging?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+
+namespace
+{
+
+// Check if all elements are non-negative
+bool is_non_negative(luci::CircleConst *node)
```
seanshpark, line 29: if `node` is input and we don't change the contents, let's use like `const luci::CircleConst *node`

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+  for (uint32_t i = 0; i < size; ++i)
+  {
+    if (node->at<loco::DataType::FLOAT32>(i) < 0)
+      return false;
+  }
+  return true;
+}
+
+// Check if mul is batchnorm mul
+bool is_batchnorm_mul(luci::CircleMul *mul, luci::CircleNode *&pred_node, luci::CircleConst *&gamma)
```
seanshpark, line 43: same reasone, can we use `const luci::CircleMul *mul`

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+{
+  auto x = static_cast<luci::CircleNode *>(mul->x());
+  auto y = static_cast<luci::CircleNode *>(mul->y());
+
+  luci::CircleNode *pred;
+  luci::CircleConst *constant;
```
seanshpark, line 49: @llFreetimell , can these give warning from static code analysis or OK as below codes set them?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+  if (!(channel_dim == mul->dim(mul->rank() - 1)))
+    return false;
+
+  pred_node = pred;
+  gamma = constant;
+  return true;
+}
+
+// Check if add is batchnorm add
+bool is_batchnorm_add(luci::CircleAdd *add, luci::CircleMul *&mul, luci::CircleConst *&beta)
```
seanshpark, line 81: same reason, `const luci::CircleAdd *add`

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+    if (filter->rank() != 4)
+      return nullptr;
+
+    if (filter->dim(3).value() != channel_size)
+      return nullptr;
+
+    if (loco::succs(filter).size() != 1)
+      return nullptr;
+
+    return conv;
```
seanshpark, line 140: (optional) `return dynamic_cast<luci::CircleConv2D *>(node);`

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+  }
+  // MUL can be fused with CONV across MEAN
+  // i.e., MUL-MEAN-CONV -> MEAN-CONV
+  // This is for handling the last part of ResNetV2
+  else if (opcode == luci::CircleOpcode::MEAN)
+  {
+    auto mean = static_cast<luci::CircleMean *>(node);
```
seanshpark, line 147: `static_cast` ?

seanshpark: maybe use `loco::must_cast` ?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+  auto y = static_cast<luci::CircleNode *>(mul->y());
+
+  luci::CircleNode *pred;
+  luci::CircleConst *constant;
+
+  if (x->opcode() == luci::CircleOpcode::CIRCLECONST &&
+      y->opcode() != luci::CircleOpcode::CIRCLECONST)
+  {
+    pred = y;
+    constant = static_cast<luci::CircleConst *>(x);
```
seanshpark, line 55: `static_cast` ? maybe use `loco::must_cast`

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+    if (node->at<loco::DataType::FLOAT32>(i) < 0)
+      return false;
+  }
+  return true;
+}
+
+// Check if mul is batchnorm mul
+bool is_batchnorm_mul(luci::CircleMul *mul, luci::CircleNode *&pred_node, luci::CircleConst *&gamma)
+{
+  auto x = static_cast<luci::CircleNode *>(mul->x());
```
seanshpark, line 45: `static_cast` ? maybe use `loco::must_cast`

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+      y->opcode() != luci::CircleOpcode::CIRCLECONST)
+  {
+    pred = y;
+    constant = static_cast<luci::CircleConst *>(x);
+  }
+  else if (x->opcode() != luci::CircleOpcode::CIRCLECONST &&
+           y->opcode() == luci::CircleOpcode::CIRCLECONST)
+  {
+    pred = x;
+    constant = static_cast<luci::CircleConst *>(y);
```
seanshpark, line 61: `static_cast` ? maybe use `loco::must_cast`

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+    if (filter->rank() != 4)
+      return nullptr;
+
+    if (filter->dim(0).value() != channel_size)
+      return nullptr;
+
+    if (loco::succs(filter).size() != 1)
+      return nullptr;
+
+    return conv;
```
seanshpark, line 202: (optiona) `return dynamic_cast<luci::CircleConv2D *>(node);`

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+      bias->at<loco::DataType::FLOAT32>(i) += beta->at<loco::DataType::FLOAT32>(i);
+    else
+      bias->at<loco::DataType::FLOAT32>(i) -= beta->at<loco::DataType::FLOAT32>(i);
+  }
+  return true;
+}
+
+void update_conv_weights_with_gamma(luci::CircleConv2D *conv, luci::CircleConst *gamma)
+{
+  auto filter = dynamic_cast<luci::CircleConst *>(conv->filter());
```
seanshpark, line 261: please use `loco::must_cast`

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,701 @@
...
+          uint32_t offset = o * filter_height_dim * filter_width_dim * filter_in_dim +
+                            h * filter_width_dim * filter_in_dim + w * filter_in_dim + i;
+          filter->at<loco::DataType::FLOAT32>(offset) *= gamma->at<loco::DataType::FLOAT32>(i);
+        }
+      }
+    }
+  }
+}
+
+luci::CircleSub *insert_sub(luci::CircleNode *pred, luci::CircleConst *beta)
```
seanshpark, line 285: `const luci::CircleNode *pred` and `const luci::CircleConst *beta` ?

seanshpark:
using `static_cast` here seems OK but please use `dynamic_cast` or `loco::must_cast` for consistancy with other codes.

seanshpark:
if the argument is only used as `read` please add `const`

seanshpark:
this PR code is a bit huge. for better  review, I strongly suggest to split codes. It is very hard to consentrate for a long time reading and understanding the codes.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: You're right. I forgot to remove it. Thanks.

jinevening:
> using static_cast here seems OK but please use dynamic_cast or loco::must_cast for consistancy with other codes.

Ok, I'll follow the convention. I've changed static_cast to loco::must_cast when casting from loco node to circle node. 

> if the argument is only used as read please add const

I've changed them for possible cases.

> this PR code is a bit huge. for better review, I strongly suggest to split codes. It is very hard to concentrate for a long time reading and understanding the codes and then thinking about what may go wrong.

I understand it is hard. I've split the previous PR and uploaded only the first stage of the pass. Now FusePreActivationBatchNormPass.cpp is 244 lines. PTAL :)

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Changing them to const generates compilation error (because of no interface implementation).

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,243 @@
...
+                      luci::CircleConst *&gamma)
+{
+  auto x = loco::must_cast<luci::CircleNode *>(mul->x());
+  auto y = loco::must_cast<luci::CircleNode *>(mul->y());
+
+  luci::CircleNode *pred = nullptr;
+  luci::CircleConst *constant = nullptr;
+
+  if (x->opcode() == luci::CircleOpcode::CIRCLECONST &&
+      y->opcode() != luci::CircleOpcode::CIRCLECONST)
```
seanshpark, line 51: (optional) we can also use tlike this
```
auto x = dynamic_cast<luci::CircleConst*>(mul->x());
auto y = dynamic_cast<luci::CircleConst*>(mul->y());

if (x != nullptr && y == nullptr)
```

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,243 @@
...
+ *           [Mul]  gamma
+ *             |
+ *        [Add + Relu]  beta
+ *  AFTER
+ *
+ *           [Add]  beta/gamma
+ *             |
+ *           [Mul]  gamma
+ *             |
+ *           [Relu]
```
seanshpark, line 138:
```suggestion
 *  BEFORE
 *           [Mul]  gamma
 *             |
 *        [Add + Relu]  beta
 *
 *  AFTER
 *           [Add]  beta/gamma
 *             |
 *           [Mul]  gamma
 *             |
 *           [Relu]
```

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,243 @@
...
+    return false;
+
+  if (loco::succs(mul).size() != 1)
+    return false;
+
+  if (!is_batchnorm_mul(mul, pred_node, gamma))
+    return false;
+
+  if (beta->dtype() != loco::DataType::FLOAT32 || gamma->dtype() != loco::DataType::FLOAT32)
+    throw std::runtime_error("FusePreActivationBatchNormPass only supports fp32 model");
```
seanshpark, line 158: maybe `fp32` --> `F32` ?

seanshpark (approved):
LGTM thank you!

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
llFreetimell: Is it correct to die when `gamma->dtype()` is not supported or just returning `false` is correct?

jinevening: Changed to Float32 :)

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Changed. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: I've changed the code.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: I've added initialization code

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,240 @@
...
+  auto x = loco::must_cast<luci::CircleNode *>(add->x());
+  auto y = loco::must_cast<luci::CircleNode *>(add->y());
+
+  luci::CircleMul *pred = nullptr;
+  luci::CircleConst *constant = nullptr;
+
+  if (add->fusedActivationFunction() != luci::FusedActFunc::RELU)
+    return false;
+
+  if (x->opcode() == luci::CircleOpcode::CIRCLECONST && y->opcode() == luci::CircleOpcode::MUL)
```
seanshpark, line 89: (optional) as you changed the pattern to check the node type in `is_batchnorm_mul()`, how about apply here too?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,240 @@
...
+    return false;
+
+  if (loco::succs(mul).size() != 1)
+    return false;
+
+  if (!is_batchnorm_mul(mul, pred_node, gamma))
+    return false;
+
+  if (beta->dtype() != loco::DataType::FLOAT32 || gamma->dtype() != loco::DataType::FLOAT32)
+    throw std::runtime_error("FusePreActivationBatchNormPass only supports Float32 model");
```
llFreetimell, line 155: Is it correct to die when `dtype` are not supported or is it correct to return `false`?

jinevening: > Is it correct to die when dtype are not supported or is it correct to return false?

I think this is a matter of choice, not correctness :)

In my opinion, a non-float model is a completely wrong input to this pass, so I decided to throw an exception.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: This code is more complicated to apply the pattern, because it checks opcode is not only CIRCLECONST but also MUL.

llFreetimell (approved):
LGTM

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
seanshpark: OK :)

seanshpark (approved):
LGTM

jinevening:
@mhs4670go PTAL

compiler/luci/pass/include/luci/Pass/FusePreActivationBatchNormPass.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+struct FusePreActivationBatchNormPass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::FusePreActivationBatchNormPass"; }
+
+  bool run(loco::Graph *g) final;
+
+  std::vector<luci::CircleMul *> _mul_list;
+  std::vector<luci::CircleAdd *> _add_list;
+  std::vector<luci::CircleSub *> _sub_list; // inserted during fusion
```
mhs4670go, line 37: This variable will be used later?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,240 @@
...
+    {
+      beta->at<loco::DataType::FLOAT32>(i) = 1;
+    }
+    else
+    {
+      // If g is 0, we use a small value (empirically determined)
+      if (g == 0)
+        g = 1e-10;
+      beta->at<loco::DataType::FLOAT32>(i) = b / g;
+    }
```
mhs4670go, line 190:
```suggestion
      // If g is 0, we use a small value (empirically determined)
      if (g == 0)
        g = 1e-10;
      beta->at<loco::DataType::FLOAT32>(i) = b / g;
    
```
Isn't it possible?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -0,0 +1,240 @@
...
+  // Valid pattern was not detected. Fast exit.
+  if (!changed)
+    return false;
+
+  // Step 2. Fuse MUL with the next CONV
+
+  // Step 3. Fuse ADD with the preceding CONV and insert SUB
+
+  // Step 4. Fuse SUB to CONV (SUB -> ADD <- CONV pattern)
+
```
mhs4670go, line 236: Just curiosity, I'm not sure but if the fusion proceeds sequentially, what if the condition that checks if it can keep going or not is fail? I mean, if Step3 returns false, what can we restore the graph?

compiler/luci/pass/include/luci/Pass/FusePreActivationBatchNormPass.h
jinevening: Yes, `_sub_list` will be used later.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: If b ==0 and g == 0, beta will be 1 in the previous code and 0 in the suggested code.

Actually, this is a corner case that may not happen in real models. I made it to 1 for simple testing (https://github.com/Samsung/ONE/pull/4769/files#diff-ce6d441c9cf90c4ae46084e6c0b00d6d0031c90b455ca36afa662e7e335156b4R179).

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: > I'm not sure but if the fusion proceeds sequentially, what if the condition that checks if it can keep going or not is fail? I mean, if Step3 returns false, how can we restore the graph?

The graph is sequentially changed as much as possible. It is not restored even if the next step fails. This is possible because each step is safe (= produces the same result after applied).

mhs4670go (approved):
LGTM

PR #4784, opened 2020-10-27
[luci] Fuse mul with conv (for FusePreActivationBatchNormPass)

jinevening:
This adds a logic to fuse mul with conv for FusePreActivationBatchNormPass

---
Related to: #4773
Draft PR: #4769

### DIFF
--- a/compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
+++ b/compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
@@ -114,10 +114,156 @@ bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::C
   return true;
 }
 
+luci::CircleConv2D *get_forward_conv2d(luci::CircleNode *node, uint32_t channel_size)
+{
+  auto opcode = node->opcode();
+  if (opcode == luci::CircleOpcode::CONV_2D)
+  {
+    auto conv = loco::must_cast<luci::CircleConv2D *>(node);
+    auto filter = dynamic_cast<luci::CircleConst *>(conv->filter());
+
+    if (filter == nullptr)
+      return nullptr;
+
+    if (filter->rank() != 4)
+      return nullptr;
+
+    if (filter->dim(3).value() != channel_size)
+      return nullptr;
+
+    if (loco::succs(filter).size() != 1)
+      return nullptr;
+
+    return conv;
+  }
+  // MUL can be fused with CONV across MEAN
+  // i.e., MUL-MEAN-CONV -> MEAN-CONV
+  // This is for handling the last part of ResNetV2
+  else if (opcode == luci::CircleOpcode::MEAN)
+  {
+    auto mean = loco::must_cast<luci::CircleMean *>(node);
+    auto axis = mean->reduction_indices();
+    auto axis_const = dynamic_cast<luci::CircleConst *>(axis);
+    if (not axis_const)
+      return nullptr;
+
+    assert(axis_const->dtype() == loco::DataType::S32);
+    auto axis_size = axis_const->size<loco::DataType::S32>();
+    for (uint32_t i = 0; i < axis_size; ++i)
+    {
+      // Reduction axis must not be the channel index
+      // Assumption: Layout is channel-last
+      if (axis_const->at<loco::DataType::S32>(i) == static_cast<int32_t>(node->rank() - 1))
+        return nullptr;
+    }
+
+    auto succ = loco::succs(node);
+    if (succ.size() != 1)
+      return nullptr;
+
+    auto succ_node = loco::must_cast<luci::CircleNode *>(*succ.begin());
+
+    return get_forward_conv2d(succ_node, channel_size);
+  }
+  else
+  {
+    return nullptr;
+  }
+}
+
+void update_conv_weights_with_gamma(const luci::CircleConv2D *conv, const luci::CircleConst *gamma)
+{
+  auto filter = loco::must_cast<luci::CircleConst *>(conv->filter());
+
+  uint32_t filter_out_dim = filter->dim(0).value();
+  uint32_t filter_height_dim = filter->dim(1).value();
+  uint32_t filter_width_dim = filter->dim(2).value();
+  uint32_t filter_in_dim = filter->dim(3).value();
+  for (uint32_t o = 0; o < filter_out_dim; o++)
+  {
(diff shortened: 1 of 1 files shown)


jinevening:
For reviewers: This is the second stage of fusing pre-activation BN.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -114,11 +114,157 @@ bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::C
...
 }
 
+luci::CircleConv2D *get_forward_conv2d(luci::CircleNode *node, uint32_t channel_size)
+{
+  auto opcode = node->opcode();
+  if (opcode == luci::CircleOpcode::CONV_2D)
+  {
+    auto conv = loco::must_cast<luci::CircleConv2D *>(node);
+    auto filter = dynamic_cast<luci::CircleConst *>(conv->filter());
+
```
seanshpark, line 124: maybe remove this empty line to follow like below `else` block or add space in `else` block

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -114,11 +114,157 @@ bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::C
...
+bool fuse_mul_with_conv(luci::CircleMul *mul)
+{
+  luci::CircleNode *pred_node = nullptr;
+  luci::CircleConst *gamma = nullptr;
+
+  if (!is_batchnorm_mul(mul, pred_node, gamma))
+    return false;
+
+  auto mul_succ = loco::succs(mul);
+  assert(mul_succ.size() == 1);
```
seanshpark, line 230: is this condition checked somewhere else or is there any possibility to meet this in some networks?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -114,11 +114,157 @@ bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::C
...
+    if (conv == nullptr)
+      fusable = false;
+  }
+
+  if (fusable)
+  {
+    for (auto s : relu_succ)
+    {
+      // Find the next CONV
+      auto conv = get_forward_conv2d(loco::must_cast<luci::CircleNode *>(s), channel_size);
```
seanshpark, line 251: `get_forward_conv2d` can return `nullptr`. is this OK?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: The first one. This is checked in swap_mul_add, so it should be true. Do you want to remove this assert?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -114,11 +114,157 @@ bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::C
   return true;
 }
 
+luci::CircleConv2D *get_forward_conv2d(luci::CircleNode *node, uint32_t channel_size)
```
mhs4670go, line 117:
```suggestion
luci::CircleConv2D *get_forward_conv2d(const luci::CircleNode *node, uint32_t channel_size)
```
This can be const as well.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: It is checked in front of this code. If `get_forward_conv2d` returns nullptr, `fusable` is false, so execution does not reach here.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -114,11 +114,157 @@ bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::C
...
+  }
+  else
+  {
+    return nullptr;
+  }
+}
+
+void update_conv_weights_with_gamma(const luci::CircleConv2D *conv, const luci::CircleConst *gamma)
+{
+  auto filter = loco::must_cast<luci::CircleConst *>(conv->filter());
```
mhs4670go, line 176: maybe assert(conv != nullptr) or sth?

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: Updated. Thanks.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
jinevening: `conv` is always not null in this code. `update_conv_weights_with_gamma` is used only in this file (defined in anonymous namespace), so I did not insert assert.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
seanshpark: > The first one.

OK :)

> Do you want to remove this assert?

Nope, `assert` tells that this should be the condition for this function to work.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
seanshpark: ah, this is called twice...

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
seanshpark: maybe adding `assert` may help reading the code as that would be the condition of this function to work...

jinevening: But I'll add assert for safety.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #4783, opened 2020-10-27
[onert] While op test - failed test case


wateret:
When executing this model, it runs successfully, but the output tensor value mismatches.

Mismatches just one element index [0], in output 1.

```
Note: Google Test filter = GenModelTest.OneOp_While_Large:GenModelTest.OneOp_If_Regression
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from GenModelTest
[ RUN      ] GenModelTest.OneOp_While_Large
/home/hanjoung/ws/gh/Samsung/ONE/tests/nnfw_api/src/GenModelTest.h:393: Failure
The difference between refval and val is 9, which exceeds 0.001, where
refval evaluates to 109,
val evaluates to 100, and
0.001 evaluates to 0.001.
Output #1, Element Index : 0
[  FAILED  ] GenModelTest.OneOp_While_Large (33 ms)
[----------] 1 test from GenModelTest (33 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (34 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] GenModelTest.OneOp_While_Large
```

```cpp
TEST_F(GenModelTest, OneOp_While_Regression)
{
  // The model looks just like the below pseudocode
  //
  // function model(x, data)
  // {
  //   // `data` does not do anything but passed to while's cond and body subgraphs
  //   // to measure copy overhead between subgraphs
  //   while (x < 100.0)
  //   {
  //     x = x + 1.0;
  //   }
  //   return (x, data)
  // }

  const int kElems = 4;
  const std::vector<int32_t> shape{kElems};

  CircleGen cgen;
  uint32_t incr_buf = cgen.addBuffer(std::vector<float>{1});
  uint32_t incr_data_buf = cgen.addBuffer(std::vector<float>(kElems, 1));
  uint32_t end_buf = cgen.addBuffer(std::vector<float>{100});

  // primary subgraph
  {
    int x_in = cgen.addTensor({{1}, circle::TensorType_FLOAT32});
    int d_in = cgen.addTensor({shape, circle::TensorType_FLOAT32});
    int x_out = cgen.addTensor({{1}, circle::TensorType_FLOAT32});
    int d_out = cgen.addTensor({shape, circle::TensorType_FLOAT32});
[...]

Issue #4138, opened 2020-09-08
[onert] Copy elimination on subgraph calls 


wateret:
Remove redundant copies between subgraph calls.

I'm thinking a way that making controlflow tensors can indirect another tensor. And when calling/returning just change the pointing tensor

wateret:
This will take longer than I thought. I made a draft that works for some cases(not all, not uploaded yet). And I found out that I need to resolve other issues first.

glistening:
@wateret I am happy to hear "copy elimination` and not not so happy because my model has only one subgraph.

`onert` has lots of redundant copies. Whenever I compare c++ native implementation and `onert` working`, I almost always find `onert` make redundant object (which requires allocations and copies). For example, `Reshape`, `StridedSlice` (when it is used like referencing index of array) ..., do you have plan to optimize those cases also?

wateret:
> For example,Reshape, StridedSlice` (when it is used like referencing index of array) ..., do you have plan to optimize those cases also?

@glistening Sorry I don't get what you mean by that. Are you talking about compile-time copies or runtime copies? If it is the later, could you please elaborate the example? 😄

glistening:
@wateret I meant the latter. For example, see ReshapeLayer.cc in cpu backend.

```c++
void ReshapeLayer::run() { reshapeGeneric(); }

void ReshapeLayer::reshapeGeneric()
{
  size_t count = _input->total_size();
  memcpy(_output->buffer(), _input->buffer(), count);
}
```

wateret:
@glistening Thank you for your explanation. That is a very good point. For the Reshape case it looks like `onert` must support in-place operations, currently input and output must differ. StridedSlice looks trickier, but basically the same.

That one is present in my task list for 2021(in our secret repo),

> - Peephole Optimizations
>     - In-place calculation for supported operations (No need to allocate another tensor for outputs, the caluclation is done in the inputs)

I'm not saying I must be the one who works on it, though.

wateret:
I just had second thoughts that Reshape and StridedSlice cases(and other in-place ops) can be handled from backend side, somewhat similar to Concat Elimination(acl backends).

glistening:
@wateret In my target model, almost all `reshape`'s source is not used any longer. I would like to handle this case first in a few weeks. If you don't have plan, I will try.

s-barannikov:
[...]

PR #4785, opened 2020-10-27
[onert] Fix subgraphs IO tensor lifetime

wateret:
Make inputs alive until the run ends

The as-is implementation could reuse input's memory from outputs, but
that could cause memory misuse for While op which performs copying
inputs and outputs.

### DIFF
--- a/runtime/onert/core/src/compiler/Linear.cc
+++ b/runtime/onert/core/src/compiler/Linear.cc
@@ -119,9 +119,12 @@ void Linear::planTensors(const compiler::LoweredGraph &lowered_graph,
     tensor_builder_map[ind] = tensor_builder;
   });
 
+  const auto io_tensors =
+      (graph.getInputs() + graph.getOutputs()) | ir::Remove::DUPLICATED | ir::Remove::UNDEFINED;
+
   // If a tensor is model output, increase the use of the tensor.
   // This aim is same to above one.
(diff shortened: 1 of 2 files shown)


wateret:
For #4783

periannath:
It looks like this PR will increase memory usage since input tensors are alive during exeuction of `OpSequence`.  How about allocate input tensors only for the subgraph with `While` operation?

wateret:
You're right. I thought about that but it doesn't seem like it is that simple. 😓 

- In theory, there is no restriction that a subgraph is used once for just one If or While op
    - So to handle this, we must scan all If/While ops to make sure a subgraph is used by If or While
- Actually this will have no effect once #4138 lands, as it totally changes how to handle I/O tensors
- I'm also working on the thing that any subgraph can be primary graph on #4138, which may be necessary for NN API 1.3 support

Please also note that this change does not affect primary graphs as they use `UserTensor` which is a user-given tensor and they are handled by `planTensors` AFAIR. This means that the only disadvantage for this is that `If` subgraphs could have higher memory use. But I also think #4138 will resolve that.

periannath:
@wateret Thanks for your detailed explanation. I was concerned about general models without `IF` op, and it looks good since this PR doesn't affect their memory usage.

periannath (approved):
LGTM

PR #4786, opened 2020-10-27
[nnfwapi-internal] Add negative tests for config

wateret:
- Add negative tests for config
- Handle error correctly when session is null


hseok-oh (approved):
LGTM

lemmaa (approved):
LGTM

PR #4788, opened 2020-10-27
DRAFT: Make negative BN gamma to positive

jinevening:
On going draft to make a pass to enforce positive batchnorm gamma

---
Related to: #4773

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -90,6 +90,12 @@ int entry(int argc, char **argv)
       .default_value(false)
       .help("This will fuse operators to InstanceNorm operator");
 
+  arser.add_argument("--make_batchnorm_gamma_positive")
+      .nargs(0)
+      .required(false)
+      .default_value(false)
+      .help("This will make negative gamma of BatchNorm into a small positive value (1e-10)");
+
   arser.add_argument("--resolve_customop_add")
       .nargs(0)
       .required(false)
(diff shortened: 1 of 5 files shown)


jinevening:
### Background

`FusePreActivationBatchNormPass` can be applied only when batchnorm gamma is positive. `MakeBatchNormGammaPositivePass` manually changes negative gamma to positive, allowing users to apply `FusePreActivationBatchNormPass` for those models with negative gammas.

As depicted in the comment, this pass can change the execution result, so users should carefully use this pass.

seanshpark:
Code looks OK

mhs4670go:
@jinevening This pass uses only when user want to run `FusePreActivationBatchNormPass` Pass? If so, how about making these two pass into anther one pass? Not codes level, interface level.

jinevening:
> If so, how about making these two pass into anther one pass?

That was my first plan :) I've changed my mind after conversation with @parjong . He said that we had better make `MakeBatchNormGammaPositivePass` as a separate pass and make a user decide whether to use this option or not.

mhs4670go:
@jinevening Right. I thought if `FusePreActivationBatchNormPass` is the only pass that follows `MakeBatchNormGammaPositivePass`, the decision whether to use `MakeBatchNormGammaPositivePass` or not is same with the one whether to use `ONE_PASS_INCLUDING_TWO_PASS` or not. Maybe it seems not the only one I guess as the conversion you said above.

jinevening:
> Okay. I thought if FusePreActivationBatchNormPass is the only pass that follows MakeBatchNormGammaPositivePass, the decision whether to use MakeBatchNormGammaPositivePass or not is same with the one whether to use ONE_PASS_INCLUDING_TWO_PASS or not. Maybe it seems not the only one I guess as the conversion you said above.

My explanation might have confused you. `FusePreActivationBatchNormPass` is an independent pass. It can be used without `MakeBatchNormGammaPositivePass`. It will remove batchnorm with non-negative gamma and leave batchnorm with negative gamma.

`MakeBatchNormGammaPositivePass` just helps to increase the applicability of `FusePreActivationBatchNormPass` for some special models that the change does not affect the model accuracy (e.g., ResNet V2).

Issue #4735, opened 2020-10-22
Rename developer resources


hseok-oh:
We need to change `nnas` to `one`

### Docker image

`nnfw/nnas:latest` -> `nnfw/one-devtools:latest` (alias tag `nnfw/one-devtools:xenial-x86_64`)
`nnfw/nnas:1804` -> `nnfw/one-devtools:bionic-x86_64`

We need new docker account for buildtools image (and release images) instead of `nnfw`

### Command

`nnas` -> `one-dev`

ex) Format checking command

```
./one-dev format
```

Anything else?

PR #4791, opened 2020-10-27
[infra] Build docker image with ubuntu codename

hseok-oh:
Use ubuntu codename for docker build parameter
build-docker-image "--extension" parameter will be deprecated
Change dockerfile path by using ubuntu codename


infra/command/build-docker-image
```diff
@@ -26,10 +29,20 @@ do
...
+      else
+        echo "Invalid extension"
+        exit 1
+      fi
       shift
       shift
       ;;
+    --codename)
+      UBUNTU_CODENAME=$2
+      DOCKER_TAG=$2
```
lemmaa, line 43: I cannot understand this line. Are the rules for generating image names different when using `--extension` and `--codename`?

hseok-oh: `--extension` didn't set image names. `--codename` will set tag automatically if user don't set name explicitly by using `--tag` parameter or `DOCKER_IMAGE_NAME` environment variable

lemmaa: I mean that the :`tag` part of image name, if I run this script with `--codename bionic`, the DOCKER_TAG will set as `bionic` here, so default image name will be `nnfw/nnas:bionic`, right?

hseok-oh: >so image name will be nnfw/nnas:bionic, right?

Right. 

I suggested to rename docker image (#4735), and it prepares renaming with ubuntu codename. When we support different architecture (ex. aarch64), tag name will include target architecture later.

lemmaa: Aha, I understood. I have missed that issue. 👍

lemmaa (approved):
LGTM!

PR #4797, opened 2020-10-27
[res] Recurrent keras models

binarman:
Added LSTM, RNN and GRU keras models to res/TensorFlowPythonExamples/

for #4528


seanshpark (approved):
LGTM

PR #4685, opened 2020-10-20
[Draft][luci] Support dynamic shape inference in luci

llFreetimell:
Parent Issue : #4372

This draft is for supporting dynamic shape inference in `luci` with `shape_signature`

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

Issue #4372, opened 2020-09-22
[Compiler][luci] Support dynamic shape inference in luci


llFreetimell:
### Background

When shape has unknown dimension, `shape_signature` is inserted in TFLite.
However, after `circle2circle` is executed, all the `shape_signature` are removed because `luci` do not handle it.
It means that Compiler do not support dynamic shape inference for now and as a result dynamic shape information is removed.
To resolve this issue, let's support dynamic shape inference in `luci`.

### Candidate Solutions

- :heavy_check_mark: Introduce new service, `ShapeSignatureInference`
- Add dynamic shape inference code in current implementation
    - Unless `loco` design is changed, we cannot change attribute(except shape) of circle node because of following constraints.
https://github.com/Samsung/ONE/blob/e5b1842133329f2abc95e2fa403cade3ea10169a/compiler/luci/service/src/CircleShapeInferenceRule.cpp#L2486-L2493
    - Introducing `shape_signature` in `loco` seems not good because
        - Modifying `loco` is dangerous because it can cause unexpected problem in other components
        - `shape_siganture` is characteristic of Circle, not common.

### Todo

- [x] Introduce `ShapeSignature` in `luci/lang` #4798 #4849 #4958 #5043
- [x] Support `ShapeSiagnture`  in `luci/export` and `luci/import` #4978 #5032
- [x] Apply `ShapeSignature` in shape validation codes in `luci` #5031
- [x] Introduce `ShapeSignature` in `tflchef` #4959 #4980
- [x] Introduce `ShapeSignature` in `circlechef` #4960 #5015
- [x] Introduce `ShapeSignatureInference` in `luci/Service` #4979 #5052
- [x] Introduce `ShapeSignatureInferencePass` in `luci/Pass` #5035
- Implement `ShapeSignatureInferenceAlgorithm` for each nodes (127 nodes)
    - This is handled in #5037
- [x] Enable `ShapeSignatureInferencePass` in `CircleOptimizer` and `luci/export` #5161

### Draft codes

- #4685 : Overall design is complete

llFreetimell:
Done!

PR #4798, opened 2020-10-27
[luci] Introduce ShapeSignature in luci/IR

llFreetimell:
Parent Issue : #4372
Draft PR : #4685

This commit will introduce `ShapeSiganture` in `luci/IR`

### DIFF
--- a/compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
+++ b/compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
@@ -33,7 +33,8 @@ public:
     _shape_signature = shape_signature;
   }
 
-  const std::vector<int32_t> &as_vector() const { return _shape_signature; }
+public:
+  std::vector<int32_t> as_vector() { return _shape_signature; }
 
   int32_t dim(uint32_t d) const { return _shape_signature.at(d); }
   int32_t &dim(uint32_t d) { return _shape_signature.at(d); }
--- a/compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
+++ b/compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
```diff
@@ -0,0 +1,49 @@
...
+  ShapeSignature() = default;
+
+  ShapeSignature(const std::vector<int32_t> &shape_signature)
+  {
+    _shape_signature = shape_signature;
+  }
+
+  const std::vector<int32_t> as_vector() const { return _shape_signature; }
+
+  const int32_t &dim(uint32_t d) const { return _shape_signature.at(d); }
```
seanshpark, line 38: `const int32_t` vs `int32_t` in return type... do we need to add `const` for this?

compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
```diff
@@ -0,0 +1,49 @@
...
+{
+public:
+  ShapeSignature() = default;
+
+  ShapeSignature(const std::vector<int32_t> &shape_signature)
+  {
+    _shape_signature = shape_signature;
+  }
+
+  const std::vector<int32_t> as_vector() const { return _shape_signature; }
```
seanshpark, line 36: Will this return a copy of `_shape_signature` ? if so, is this your intenstion?

compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
llFreetimell: At first, I was not sure so I referenced following code...
https://github.com/Samsung/ONE/blob/92f5e69b4a8ac45a051e3f62768915931b38376f/compiler/loco/include/loco/IR/NodeMixins.h#L67-L68

I think that we may need to control the modification authority.

compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
llFreetimell: This function is for following (draft)code in #4685 so it was my intension :)

https://github.com/Samsung/ONE/blob/b11505e89ac277f4177d913d9bdd4de24af3a5bf/compiler/luci/export/src/CircleTensorExporter.cpp#L256

seanshpark: As it returns `const` and it's const method, caller cannot modify it. I think it would be better to return reference for performance.

compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
seanshpark: `const int32_t &dim()` as `int32_t` is just a number, not a class or struct, I think it would be same for `const int32_t dim()`
For this I'm not sure `const` is usefull for the caller. just `int32_t dim()` wouldn't be better for usage?

compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
seanshpark: As we talked off-line, I think `std::vector<int32_t> as_vector() { ... }` can hide internal implementation and caller can R/W vector :)

seanshpark (approved):
LGTM thank you!

PR #4816, opened 2020-10-28
[one-cmds] Revise one-codegen as python script

mhs4670go:
This commit revises one-codegen as python script.

Related : #4278
Draft: #4758

### DIFF
--- a/compiler/one-cmds/one-codegen
+++ b/compiler/one-cmds/one-codegen
@@ -1,4 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
+''''export SCRIPT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"                  # '''
+''''export PY_PATH=${SCRIPT_PATH}/venv/bin/python                                       # '''
+''''test -f ${PY_PATH} && exec ${PY_PATH} "$0" "$@"                                     # '''
+''''echo "Error: Virtual environment not found. Please run 'one-prepare-venv' command." # '''
+''''exit 255                                                                            # '''
 
 # Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
 #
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

compiler/one-cmds/one-codegen
```diff
@@ -14,53 +19,93 @@
...
+    backends_list = []
+    for cand in files:
+        if cand.endswith('-compile'):
+            # 8 : length of '-compile'
+            backends_list.append(cand[:-8])
+    return backends_list
+
+
+def _get_parser():
+    parser = argparse.ArgumentParser(description='command line tool for code generation')
```
seanshpark, line 43: `command` --> `Command` if there is no other message in front of this.
Please check with all other PRs.

seanshpark: confirmed in another PR :)

compiler/one-cmds/one-codegen
```diff
@@ -1,4 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
+''''export SCRIPT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"                  # '''
+''''export PY_PATH=${SCRIPT_PATH}/venv/bin/python                                       # '''
+''''test -f ${PY_PATH} && exec ${PY_PATH} "$0" "$@"                                     # '''
+''''echo "Error: Virtual environment not found. Please run 'one-prepare-venv' command." # '''
+''''exit 255                                                                            # '''
```
parjong, line 6: @_@???

compiler/one-cmds/one-codegen
```diff
@@ -1,4 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
+''''export SCRIPT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"                  # '''
+''''export PY_PATH=${SCRIPT_PATH}/venv/bin/python                                       # '''
+''''test -f ${PY_PATH} && exec ${PY_PATH} "$0" "$@"                                     # '''
```
parjong, line 4: Ah... this line is the key..

parjong: What does ``''''`` mean in BASH?

parjong: I guess that it is a string in Python.

mhs4670go: Right. `''''    #'''` becomes string in python. ' ' ' ' means `''` + `''`, two string without contents. Another example is like..
```bash
$ 'l''s'
$ "l""s"
```
and so on.

parjong (approved):
;-)

PR #4824, opened 2020-10-28
[one-cmds] Introduce one-build

mhs4670go:
This commit introduces one-build to one-cmds.

Related : #4278
Draft: #4758


seanshpark:
CC @struss , @llFreetimell , @jinevening

compiler/one-cmds/CMakeLists.txt
```diff
@@ -34,6 +35,7 @@ foreach(ONE_COMMAND IN ITEMS ${ONE_COMMAND_FILES})
 endforeach(ONE_COMMAND)
 
 set(ONE_UTILITY_FILES
+    one-build.template.cfg
```
seanshpark, line 38: I think we need to revise the document about this and also `one-build`

seanshpark (approved):
LGTM

compiler/one-cmds/one-build
```diff
@@ -0,0 +1,134 @@
...
+        'one-pack': 'one-pack',
+        'one-codegen': 'one-codegen'
+    }[driver_name]
+
+
+def _parse_cfg(args):
+    config = configparser.ConfigParser()
+    parsed = config.read(os.path.expanduser(getattr(args, 'config')))
+    if not parsed:
+        raise AssertionError('not found given configuration file')
```
jinevening, line 70: Does the message correctly describe the situation? I mean, let's assume there exists a configuration file, but the file has wrong contents and thus cannot be parsed. Is it checked here? If so, the message should be changed to something like 'Configuration file cannot be parsed.'

(optional) How about capitalizing the first character? 'Not found ~~.'

(optional) How about using `SystemExit` or `RuntimeError` instead of `AssertionError` ? Assertion is usually used for debugging purpose.
(https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement)

mhs4670go: @jinevening `configparser.read` returns the list of configuration file that was parsed successfully. Being `parsed successfully` means here that the file exists. And wrong contents will be caught by `configparser` library. BTW, I'll change the error to `FileNotFoundError`. Thank you for your suggestion.

mhs4670go:
@seanshpark @jinevening @llFreetimell PTAL:)

llFreetimell (dismissed):
LGTM

jinevening (dismissed):
LGTM

compiler/one-cmds/how-to-use-one-commands.txt
```diff
@@ -20,8 +20,73 @@ Compilation flow for NPU
...
+
+You can see a template file for how to write a configuration file in `one-build.template.cfg`.
+
+[options to write]
+
+Sometimes you want to change certain options without touching the configuration file. If you
+pass the option directly to the command line, the option is processed prior to the configura-
+tion file. A list of options can be found in each driver's help message with `-h` option.
+
+e.g. $ ./one-import tf -C my-conf.cfg -i path/to/overwrite.pb
```
seanshpark, line 51: e.g.
```
$ ./one-import tf -C my-conf.cfg -i path/to/overwrite.pb
```
?

compiler/one-cmds/how-to-use-one-commands.txt
```diff
@@ -20,8 +20,73 @@ Compilation flow for NPU
...
+
+
+one-build
+---------
+
+one-build is an integrated driver that can execute one-commands at once. It's nice to run each
+driver individually, but sometimes you'll want to put together the most frequently used comman-
+ds and run them all at once. You can do this with one-build and its configuration file.
```
seanshpark, line 59: We don't have to jalign justify. Please split before `commands`

compiler/one-cmds/how-to-use-one-commands.txt
```diff
@@ -20,8 +20,73 @@ Compilation flow for NPU
...
+
+For one-build, the configuration file needs 'one-build' section that consists of list of dri-
+ver.
```
seanshpark, line 62: same here, `driver.` not `dri-` + `ver.`

compiler/one-cmds/how-to-use-one-commands.txt
mhs4670go: @seanshpark You mean it's gotta need more detail?

compiler/one-cmds/how-to-use-one-commands.txt
mhs4670go: @seanshpark Okay. I just wanted to make it align. Being neat freak:)

seanshpark: Other blocks didn't do it like this.

compiler/one-cmds/how-to-use-one-commands.txt
seanshpark: No, content is same. Just format change with surrounding with "```"

mhs4670go:
@seanshpark @jinevening @llFreetimell PTAL:)

llFreetimell (approved):
LGTM
// This PR can be merged without my approval

Hyunjun85 (approved):

seanshpark (approved):
LGTM thank you!

seanshpark:
as @jinevening is absent today, I'll merge without his approval; he approved before :)

Issue #4693, opened 2020-10-20
[onert] Crash when PermutationEliminationPass(optimization) is off


wateret:
If disabling `PermutationInsertionPass`, it crashes like

```diff
diff --git a/runtime/onert/core/src/compiler/LoweredGraph.cc b/runtime/onert/core/src/compiler/LoweredGraph.cc
index 7df465d7..58356092 100644
--- a/runtime/onert/core/src/compiler/LoweredGraph.cc
+++ b/runtime/onert/core/src/compiler/LoweredGraph.cc
@@ -124,7 +124,7 @@ LoweredGraph::LoweredGraph(const ir::Graph &graph, const CompilerOptions &option
       .run();
 
   // Optimization passes
-  pass::PassRunner{}.append(std::make_unique<pass::PermutationEliminationPass>(*this)).run();
+  //pass::PassRunner{}.append(std::make_unique<pass::PermutationEliminationPass>(*this)).run();
 
   VERBOSE(OpSequences) << "Dump after permutation insertion" << std::endl;
   dumpOpSequences(_op_seqs, _graph.operations());
```

```
$ Product/armv7l-linux.debug/out/unittest_standalone/nnfw_api_gtest
[ RUN      ] GenModelTest.OneOp_Transpose_RegularTranspose
nnfw_api_gtest: /home/hanjoung/ws/gh/Samsung/ONE/runtime/onert/core/src/backend/cpu_common/MemoryPlanner.cc:122: virtual void onert::backend::cpu_common::WICPlanner::claim(const OperandIndex&, size_t): Assertion `size != 0' failed.
[1]    19376 abort      ONERT_LOG_ENABLE= GRAPH_DOT_DUMP=
```

As it is just an optimization pass so it should work without this pass.

wateret:
Looks like it has two problems.

- Assert on zero-sized allocations(static tensor manager)
- ArgMax shape inference has an error?

wateret:
Fixed

PR #4825, opened 2020-10-28
[onert/cpu] Fix LSTMLayer optional inputs

wateret:
When LSTMLayer gets an optional input of size 0, make the corresponding
input to be null, as the kernel implementation assumes it is ungiven
only if it is null.


wateret:
This is found when PermutationEliminationPass is off #4693 .

An alternative fix would be https://github.com/Samsung/ONE/pull/4744/files#diff-1d82a41619491e95fa0905ea29a806c73023f3cd41d4bfca47b9d5235e48fdb5R71 (always return null for 0-sized tensors) but this way looks more general.

ragmani (approved):
LGTM

hseok-oh (approved):

hseok-oh:
@nnfw-bot test x64-builder

Issue #4793, opened 2020-10-27
Clarifying odd(?) shapes


wateret:
While I was working on #4693 , I got some questions about tensor shape. There are some shapes that are not clear to me. I would appreciate if someone can clarify.

I'm going to use `[?, ?, ?]` notation for shapes.

Shape | Description | What I know | Size in onert
-- | -- | -- | --
`[]` | Zero-dimension shape | Scalar type | 1
`[0]` | 1D 0-sized | Literally 0-sized array(no data) | 0
`[1, 0, 3, 4]` | 0 for some particular dims | ? | 0
`[1]` |   | Could be a dummy shape for unspecified tensor? | 1

Questions

- Is it correct `[]` is size of 1, while `[0]` is 0 ?
- Is it there actual difference `[]` between `[1]`?
- `[0]` tensor vs. optional tensor - it looks like `[0]` tensor can be replaced by an optional tensor should we allow both?
- AFAIK unspecified(dynamic) tensor is represented as `[1]`. As long as its shape signature(tflite) is `-1`, does shape value here still matter?
- What does `[1, 0, 3, 4]` mean? is it useful in any way?
- Are these semantics identical for tensorflow/tflite(circle)/NNAPI?

cc: @Samsung/nnfw @samsung/nncc

hseok-oh:
>Are these semantics identical for tensorflow/tflite(circle)/NNAPI?

`NNAPI` define zero dimension when dimension value is unspecified.

```
 * A tensor operand type of specified rank but some number of
 * unspecified dimensions is represented by setting dimensionCount to
 * the rank and each unspecified dimension to 0.
 *
 * Available since API level 27.
 *
 * Starting at API level 29, a tensor operand type of unspecified rank is
 * represented by setting dimensionCount to 0 and dimensions to NULL (just as if
 * it were a scalar operand type).
```

seanshpark:
What I assume and not sure...

> Is it correct [] is size of 1, while [0] is 0 ?

Yes

> Is it there actual difference [] between [1]?

[] is scalar. but [1] is size 1 vector. I think memory structure is the same.

> [0] tensor vs. optional tensor - it looks like [0] tensor can be replaced by an optional tensor should we allow both?

Don't know either...

> AFAIK unspecified(dynamic) tensor is represented as [1].
> Does shape value here still matter? if its shape signature(tflite) is -1

[...]

PR #4826, opened 2020-10-28
[onert/cpu] Allow size 0 allocation

wateret:
As we have shapes like `[0]`, this commit removes assertions when
size 0 is passed to memory planners.

### DIFF
--- a/runtime/onert/core/src/backend/cpu_common/MemoryPlanner.cc
+++ b/runtime/onert/core/src/backend/cpu_common/MemoryPlanner.cc
@@ -27,8 +27,6 @@ namespace cpu_common
 
 void BumpPlanner::claim(const ir::OperandIndex &ind, size_t size)
 {
-  assert(size != 0);
-
   Block blk{_capacity, size};
   _mem_plans[ind] = blk;
   _capacity += size;
@@ -59,8 +57,6 @@ void BumpPlanner::release(const ir::OperandIndex &ind)
 //    the previous claim_base_offset.
 void FirstFitPlanner::claim(const ir::OperandIndex &ind, size_t size)
 {
-  assert(size != 0);
-
   // Find the right position for claiming
   uint32_t next_offset = 0;
   for (auto &mem_claim : _claim_table)
@@ -119,8 +115,6 @@ WICPlanner::WICPlanner()
 
 void WICPlanner::claim(const ir::OperandIndex &ind, size_t size)
 {
-  assert(size != 0);
-
   _operands.emplace(size, ind);
   _interference_graph[ind].insert(_interference_graph[ind].end(), _live_operands.cbegin(),
                                   _live_operands.cend());


wateret:
Draft : #4744 
Related : #4793

periannath (approved):
LGTM

runtime/onert/core/src/backend/cpu_common/MemoryPlanner.cc
```diff
@@ -27,8 +27,6 @@ namespace cpu_common
 
 void BumpPlanner::claim(const ir::OperandIndex &ind, size_t size)
 {
-  assert(size != 0);
-
```
glistening, line 31: @wateret Is it safe to just remove `assert`? What happens when `size=0`? It may cause `malloc(0)`, It would be good if `size=0` case is handled explicitly.

hseok-oh:
@nnfw-bot test x64-builder

runtime/onert/core/src/backend/cpu_common/MemoryPlanner.cc
wateret: The memory planners never return null. They have one big memory chunk and the actual buffer address is calculated `base + offset`.

However I agree that we need some exact policy for 0-sized tensors in general(not only these static planners but also dyanmic allocators). We need to define what to handle 0-sized tensors. #4744 has a different approach from this PR, that always return null for 0-sized tensors.

wateret: However I ended up with allowing 0-sized tensor allocation since I think we may have to distinguish these two:

- optional tensor which were not given  (In python, `None`)
- a tensor with no values (In python, `[]`)

And this case is the second, but with the way from #4744, we won't be able to distinguish those.

glistening: > The memory planners never return null. They have one big memory chunk and the actual buffer address is calculated `base + offset`.
> 

@wateret Can tensor A and B have same address if tensor A is allocated then next tensor B (with size 0) is allocated? If some implementation (like cache) would like to use the address as an unique key, it may break the assumption.

wateret: > Can tensor A and B have same address if tensor A is allocated then next tensor B (with size 0) is allocated?

Yes.

> If some implementation (like cache) would like to use the address as an unique key, it may break the assumption.

Good point. But would that actually break that caching assumption? I took a look at 

https://github.com/Samsung/ONE/blob/083789b16ef07d6e6ab14134db6537077fe7e76f/runtime/onert/backend/cpu/ops/FullyConnectedLayer.cc#L244-L251

but I don't see it is an issue that a tensor buffer is the same address with another tensor buffer. Maybe there is another case that would be an issue? Please correct me if I'm wrong.

glistening: @wateret For `FullyConnectedHybrid()`, I am not sure because I don't know the details how ruy handles the cache keys. Furthermore, I don't know the all cases in `onert`. Thus, I am not 100% sure this change is safe.

Before this change, it was checked with `assert`. It should be very rare case. I would choose to handle this case in early stage if possible. `[0]` is used for `unspecified` dimension, however, I think it is specified during running by static or dynamic shape inferer.

wateret: > `[0]` is used for `unspecified` dimension, however, I think it is specified during running by static or dynamic shape inferer.

You're right. So this should be more about nnapi frontend than allocating 0.

---

Anyways now that I got interested in 0-size tensors so I will keep working on this too. I think we could make some cases when there are 0-size tensors intentionally. I'll dig into it.

> @wateret For `FullyConnectedHybrid()`, I am not sure because I don't know the details how ruy handles the cache keys. Furthermore, I don't know the all cases in `onert`. Thus, I am not 100% sure this change is safe.

About caching, I still think there is no issue, so I am going to keep this way(two tensors could have a same buffer pointer). Once we find that could restrict use of ruy or other kernels then we can restrict it - "the returned tensor pointers must be distinct for the all tensors that interfere their lifetimes"

wateret:
Close this and re-work as commented https://github.com/Samsung/ONE/pull/4826#discussion_r514985026 .

runtime/onert/core/src/backend/cpu_common/MemoryPlanner.cc
wateret: @glistening I had second thoughts that we may still need to allow this. Is there a way to distinguish `[0]` is a dynamic tensor or actually a 0-size tensor in NN API?

- According to https://github.com/Samsung/ONE/issues/4793#issuecomment-717086013 it may mean it is always a dynamic tensor. (Then would that mean NN API can't express 0-size tensor?) Then as we assume the unspecified dimension value to 1 in tflite, we can go this way.
- If both are possible, we should allocate 0-size tensor in the static tensor memory area, and if we don't change anything in our allocators, it may return the same pointer for different tensors, but as I mentioned above, I don't think that is an issue.

PR #4827, opened 2020-10-28
[onert] Optimize PermuteLayer of controlflow by using enqueue functions

ragmani:
For issue #4414
Draft PR #4395

This commit Optimize PermuteLayer of controlflow by using enqueue functions instead of calling map().

### DIFF
--- a/runtime/onert/core/include/backend/ITensor.h
+++ b/runtime/onert/core/include/backend/ITensor.h
@@ -103,6 +103,17 @@ public:
    * @note  Higer dimension will be placed on front.
    */
   virtual ir::Shape getShape() const;
+
+  virtual bool is_subtensor() const { return false; }
+  virtual bool needMemoryMap() const { return false; }
+  virtual void enqueueWriteBuffer(const void *, bool)
+  {
+    throw std::runtime_error("This backend does not support enqueueWriteBuffer");
+  }
+  virtual void enqueueReadBuffer(void *, bool)
+  {
+    throw std::runtime_error("This backend does not support enqueueReadBuffer");
+  }
 };
 
 } // namespace backend
--- a/runtime/onert/core/src/backend/controlflow/kernel/PermuteLayer.cc
+++ b/runtime/onert/core/src/backend/controlflow/kernel/PermuteLayer.cc
@@ -166,6 +166,19 @@ void PermuteLayer::appendPermuteTasks(const ITensor *src_tensor, ITensor *dst_te
   _tasks_map[src_tensor] = std::move(tasks);
 }
 
+void PermuteLayer::runPermuteTasks(backend::ITensor *src, uint8_t *dst_buffer)
+{
+  assert(src->getShape().num_elements() * ir::sizeOfDataType(src->data_type()) <=
+         src->total_size());
+  std::vector<PermuteWorkerTask> &tasks = _tasks_map.at(src);
+  for (size_t i = 0; i < tasks.size(); ++i)
+  {
+    tasks.at(i).setBuffers(src->buffer(), dst_buffer);
+  }
+  assert(tasks.size() >= 1);
+  _external_context->ruy_context()->mutable_thread_pool()->Execute(tasks.size(), tasks.data());
+}
+
 void PermuteLayer::run()
 {
   assert(_src_tensors.size() == _dst_tensors.size());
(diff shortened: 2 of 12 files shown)


ragmani:
@nnfw-bot test all

ragmani:
@nnfw-bot test x64-builder

hseok-oh (dismissed):

runtime/onert/core/include/backend/ITensor.h
```diff
@@ -103,6 +103,17 @@ class ITensor
    * @note  Higer dimension will be placed on front.
    */
   virtual ir::Shape getShape() const;
+
+  virtual bool is_subtensor() const { return false; }
+  virtual bool hasClBuffer() const { return false; }
+  virtual void enqueueWriteBuffer(const void *)
```
hyunsik-yoon, line 109: Shouldn't it better for ITensor to have names neutral to specific backend?
Or how about something like the following?

- hasClBuffer -> needMemoryMap: // some backend needs CPU memory to be mapped to, e.g., GPU memory or NPU memory.
- enqueueWriteBuffer -> mapMemory

ragmani: > - hasClBuffer -> needMemoryMap: // some backend needs CPU memory to be mapped to, e.g., GPU memory or NPU memory.

Okay. I will.

> - enqueueWriteBuffer -> mapMemory

This method is not to map memory. How about changing it to writeBuffer?

hyunsik-yoon: how about writeMappedMemory ?

ragmani: This method is not related to mapping data. It writes a buffer to device memory of the backend tensor from host memory.

hyunsik-yoon: I see. The concern I have is that if the name is _too general_ like `writeBuffer`, backend developer may be curious why CPU tensor cannot support `writeBuffer`. In that sense I like `enqueueWriteBuffer` (your first naming) would be better than `writeBuffer`. Or some long name like `writeBufferToDevice`(?) would also work.

ragmani: I think the original(enqueueWriteBuffer) looks the best among them.

runtime/onert/core/include/backend/ITensor.h
```diff
@@ -103,6 +103,17 @@ class ITensor
...
+  virtual bool is_subtensor() const { return false; }
+  virtual bool needMemoryMap() const { return false; }
+  virtual void enqueueWriteBuffer(const void *, bool)
+  {
+    throw std::runtime_error("This backend does not support enqueueWriteBuffer");
+  }
+  virtual void enqueueReadBuffer(void *, bool)
+  {
+    throw std::runtime_error("This backend does not support enqueueReadBuffer");
+  }
```
wateret, line 116: Frankly I'm not a big fan of these introduced, but I also admit that there is another way that can be done quickly. We may accept this and we'll do some refactoring later.

wateret (dismissed):
LGTM

hseok-oh (dismissed):

runtime/onert/core/src/backend/controlflow/kernel/PermuteLayer.cc
```diff
@@ -217,36 +230,54 @@ void PermuteLayer::run()
...
+            // This is more effective than multi-threading
+            src->access([&](backend::ITensor &) { dst->enqueueWriteBuffer(src->buffer(), false); });
+          }
+          else
+          {
+            _buffers_map[dst].reserve(dst->total_size());
+            auto dst_buffer = _buffers_map[dst].data();
+
+            src->access([&](backend::ITensor &) { runPermuteTasks(src, dst_buffer); });
+            dst->enqueueWriteBuffer(dst_buffer, false);
```
ragmani, line 264: This block can be slower than calling ` runPermuteTasks(src, dst->buffer());` if padding size is big.

hyunsik-yoon (approved):
LGTM

hseok-oh (approved):

ragmani:
@wateret PTAL

wateret (approved):
LGTM

PR #4845, opened 2020-10-30
[onert] Add newline to log

seanshpark:
This will add missing newline to logging in LoweredGraph.cc

### DIFF
--- a/runtime/onert/core/src/compiler/LoweredGraph.cc
+++ b/runtime/onert/core/src/compiler/LoweredGraph.cc
@@ -61,7 +61,7 @@ LoweredGraph::LoweredGraph(const ir::Graph &graph, const CompilerOptions &option
     // we should change it back(throw if backend is not loaded) later.
     if (!backend)
     {
-      VERBOSE(LoweredGraph) << "Cannot load backend - " << backend_str;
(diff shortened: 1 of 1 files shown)


hyunsik-yoon (approved):
LGTM

chunseoklee (approved):

runtime/onert/core/src/compiler/LoweredGraph.cc
```diff
@@ -459,7 +459,7 @@ void LoweredGraph::dumpLowerInfo()
   {
     if (!e.second.empty())
     {
-      VERBOSE(Lower) << e.second;
+      VERBOSE(Lower) << e.second << std::endl;
```
wateret, line 462: Nit: actually the whole if statement here was inserted by accident. My bad 😢

seanshpark: I think it's not your mistake. I need to revert this change. After adding this, log had extra line after `Use Backends` item...
```
    - Def Backends    : { cpu(NHWC) }
    - Use Backends    : { cpu(NHWC) }

[Lower] Operand #7 LowerInfo
```

PR #4849, opened 2020-10-30
[luci] Add const to a function in CircleShapeSignature

llFreetimell:
Parent Issue : #4372 

If `as_vector()` function do not have `const` keyword, build is failed.
This commit will fix it.

Build fail as following
https://github.com/Samsung/ONE/blob/b11505e89ac277f4177d913d9bdd4de24af3a5bf/compiler/luci/export/src/CircleTensorExporter.cpp#L256


seanshpark (approved):
LGTM

Issue #4830, opened 2020-10-29
Compiler FE: conversion of pytorch to TF with Ops


seanshpark:
As next step of #4795, let's check our list of Ops.

seanshpark:
@underflow101 , I've done with my first target of operators in my list.
If you like, you can add other operators that converts to `tflite` with `ptem.py`
And it is OK to add that don't convert to `tflite` with a `Note` like #4898 or  #4900

underflow101:
> @underflow101 , I've done with my first target of operators in my list.
> If you like, you can add other operators that converts to `tflite` with `ptem.py`
> And it is OK to add that don't convert to `tflite` with a `Note` like #4898 or #4900

Thank you, @seanshpark ,
I will add more PyTorch examples. And thank you for your prior examples! :)
I will take a look in #4898 or #4900 which doesn't convert to `tflite`.

seanshpark:
Done :)

PR #4852, opened 2020-11-01
[res] pth examples for div,mul,sub

seanshpark:
This will introduce div, mul and sub examples for PyTorchExamples


seanshpark:
For #4830

mhs4670go (approved):
LGTM

PR #4865, opened 2020-11-02
[res] PyTorchExamples for Activations

seanshpark:
This will introduice PyTorchExamples to ReLU and several activations


seanshpark:
For #4830

mhs4670go (approved):
LGTM

Issue #4828, opened 2020-10-28
[acl_cl] Support dilation on DepthwiseConv2D


chunseoklee:
base_loader does not parse dilation parameter. 
Let's implement it. 
ComputeLibrary is ready to process DepthwiseConv2D with dilation.

hseok-oh:
I'll check this issue

chunseoklee:
Done by #4873 #4840

PR #4872, opened 2020-11-02
[onert] Allow size 0 tensors

wateret:
Allow size 0 tensors in CPU memory planner.

### DIFF
--- a/tests/nnfw_api/src/one_op_tests/Add.cc
+++ b/tests/nnfw_api/src/one_op_tests/Add.cc
@@ -84,6 +84,25 @@ TEST_F(GenModelTest, OneOp_Add_VarToVarSame)
   SUCCEED();
 }
 
+TEST_F(GenModelTest, OneOp_Add_VarToVarSize0)
+{
+  CircleGen cgen;
+  int a = cgen.addTensor({{0}, circle::TensorType::TensorType_FLOAT32});
+  int b = cgen.addTensor({{0}, circle::TensorType::TensorType_FLOAT32});
+  int c = cgen.addTensor({{0}, circle::TensorType::TensorType_FLOAT32});
+  int m = cgen.addTensor({{0}, circle::TensorType::TensorType_FLOAT32});
+  int out = cgen.addTensor({{0}, circle::TensorType::TensorType_FLOAT32});
+  cgen.addOperatorAdd({{a, b}, {m}}, circle::ActivationFunctionType_NONE);
+  cgen.addOperatorAdd({{m, c}, {out}}, circle::ActivationFunctionType_NONE);
+  cgen.setInputsAndOutputs({a, b, c}, {out});
+
+  _context = std::make_unique<GenModelTestContext>(cgen.finish());
+  _context->addTestCase(uniformTCD<float>({{}, {}, {}}, {{}}));
+  _context->setBackends({"cpu"});
+
+  SUCCEED();
+}
+
 TEST_F(GenModelTest, neg_OneOp_Add_InvalidType)
 {
   CircleGen cgen;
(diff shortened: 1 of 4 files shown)


wateret:
This is derived PR from #4826 .

@glistening I made a test case when `[0]` size tensors are needed. PTAL 😃

tests/nnfw_api/src/one_op_tests/Add.cc
```diff
@@ -84,6 +84,25 @@ TEST_F(GenModelTest, OneOp_Add_VarToVarSame)
...
+  int b = cgen.addTensor({{0}, circle::TensorType::TensorType_FLOAT32});
+  int c = cgen.addTensor({{0}, circle::TensorType::TensorType_FLOAT32});
+  int m = cgen.addTensor({{0}, circle::TensorType::TensorType_FLOAT32});
+  int out = cgen.addTensor({{0}, circle::TensorType::TensorType_FLOAT32});
+  cgen.addOperatorAdd({{a, b}, {m}}, circle::ActivationFunctionType_NONE);
+  cgen.addOperatorAdd({{m, c}, {out}}, circle::ActivationFunctionType_NONE);
+  cgen.setInputsAndOutputs({a, b, c}, {out});
+
+  _context = std::make_unique<GenModelTestContext>(cgen.finish());
+  _context->addTestCase(uniformTCD<float>({{}, {}, {}}, {{}}));
```
hyunsik-yoon, line 100: Shouln't this be the following?
```suggestion
  _context->addTestCase(uniformTCD<float>({{0}, {0}, {0}}, {{0}}));
```

wateret: Since the shape is `{0}`, the buffer must be empty. Please note that the former is "shape" and the latter is "data".

hyunsik-yoon (approved):
LGTM

periannath (approved):
LGTM

glistening:
@wateret As I mentioned #4828, I would not allow zero size tensor because it may cause several unnatural situation at least for me.

i) it may cause different tensors have same address of buffer.
ii) it requires several other modules to define how it will handle for zero size tensors. (e.g. memory planner, cache, ...) though you are sure it would be safe. I am not sure because I don't know every cases.

In addition, I am still not sure that memory planner should take account for zero size tensor. Before memory planning, unspecified tensors should be specified by either static shape inference or dynamic shape inference.

If others agree for this change, I would not block ths PR. However please I cannot approve this PR because of the reasons the above.

wateret:
@glistening 

> As I mentioned #4828, I would not allow zero size tensor because it may cause several unnatural situation at least for me.

This "Add" TC could look unnatural, but it is just to havve a simple test case. In real world, there's some cases of shape `[0]` tensors. For ReduceMax op, axis tensor(an input) could be shape of `[0]`, which does not reduce in any axis. This may look unnatural but it could actually happen, unlike the "Add" TC.

> it requires several other modules to define how it will handle for zero size tensors. (e.g. memory planner, cache, ...) though you are sure it would be safe. I am not sure because I don't know every cases.

I'm not sure too, as I don't know every case too. But it looks fine for the cases we've talked about.

> unspecified tensors should be specified by either static shape inference or dynamic shape inference.

I thought so before, but changed my mind due to the reason in https://github.com/Samsung/ONE/pull/4826#discussion_r515016276 .

wateret:
If it turns out that we need to go with @glistening 's opinion later, I think we can fix this with the way from https://github.com/Samsung/ONE/pull/4826#discussion_r515016276 . So I'll merge this first. Let's see then fix if a problem occurs!

periannath:
FYI, TFLite allows zero-sized allocation and it returns nullptr. https://github.com/tensorflow/tensorflow/blob/4fee5e9fa5a6efe1180f9c3e108083ecf9fae93a/tensorflow/lite/simple_memory_arena.h#L59

glistening:
TFLite returns nullptr. It looks better because several different tensors have same non-zero address as its tensor data buffer.

PR #4875, opened 2020-11-02
[bcq-tools] Add removed files

mhs4670go:
This commit adds removed files for backward compatibility.


seanshpark (approved):
LGTM

PR #4878, opened 2020-11-02
[res] PyTorchExamples of Reshape and others

seanshpark:
This will introduce PyTorchExamples of Flatten, cat, permute, reshape,
slice, split and squeeze


seanshpark:
For #4830

mhs4670go (approved):
LGTM

PR #4893, opened 2020-11-03
[res] PyTorchExamples of Softmax

seanshpark:
This will introduce Softmax activation


seanshpark:
For #4830 , that was missed from previous activation Ops

mhs4670go (approved):
LGTM

PR #4898, opened 2020-11-03
[res] PyTorchExamples for logical Ops

seanshpark:
This will introduce PyTorchExamples for logical_and, logical_or and logical_xor
- These have problem exporting to ONNX as of now


seanshpark:
For #4830

mhs4670go (approved):
LGTM

Issue #4901, opened 2020-11-03
Make ChromeTracing handle multiple subgraphs (executors)


hyunsik-yoon:
Currently Chrome tracing does not show a model with multiple subgraphs (executors). 

Let's make ChromeTracing handle multiple subgraphs.

hyunsik-yoon:
Draft: #4903 

Sample screen with the following model:

```
y=neg(0)
while less ( y, 100 )
  y = add (y, 10)
z = cos(y)
```

You can see 3 subgraphs: primary, while_condition, while_body
- purple bars: labeled `subgraph`
- orange op: `neg`
- pink op: `cos`
- green op: `comparison`
- blue op: `add`

![image](https://user-images.githubusercontent.com/63768207/97960783-698d6800-1df5-11eb-97c1-8a633c4f0397.png)

zoom in:
![image](https://user-images.githubusercontent.com/63768207/97961203-3ac3c180-1df6-11eb-8e54-515b22f0c098.png)

wateret:
I hope we have a subgraph index rather than the object's pointer address, later on. I guess there is no way to get the subgraph index info for now.

hyunsik-yoon:
Putting the exact subgraph index seems not easy 
- We don't want to put tflite subgraph index into `Graph` object
- So we need to have some mapping info class of IR->IR' (e.g., tflite subgraph ->graph pointer)

Maybe what we need is to _distinguish_ subgraphs in human readable way?
For now we could use something like `Subgraph A`, `Subgraph B`, etc.

seanshpark:
> Maybe what we need is to distinguish subgraphs in human readable way?

FYI, Subgraphs in tflite and circle has human readable text name.

hyunsik-yoon:
Now shape info was added.

When is this necessary?
- if you want to find an op whose input shape is [1, 2, 3], you can find it by typing "1 2 3" in _search bar_ now.
- Another format "[1, 2, 3]" would be better but Chrome browser's Json parser regards "[1, 2, 3]" as a list and it cannot be text-searchable.

![image](https://user-images.githubusercontent.com/63768207/98082785-a6706200-1ebc-11eb-9341-b3777020fcac.png)

hyunsik-yoon:
All merged.

I'd like to leave how I have tested:

0. How to check the result (assuming that log name is `m.log`)
   - check manually `m.log`, `m.log.table.md` 
   - check manyally `m.log.chrome.json`
[...]

PR #4905, opened 2020-11-03
[onert] Add more ops into OneOp_While test

hyunsik-yoon:
This add `cast` before calling `while` and 
add another `cast` after calling `while`. 

This is to check if input and output of `while` is handled well.

for issue https://github.com/Samsung/ONE/issues/4901 to check if we also see other ops well before and after `while`.

### DIFF
--- a/runtime/onert/core/src/backend/controlflow/kernel/PermuteLayer.cc
+++ b/runtime/onert/core/src/backend/controlflow/kernel/PermuteLayer.cc
@@ -55,7 +55,10 @@ void PermuteLayer::optimize()
     {
       auto src = *src_it;
       auto dst = *dst_it;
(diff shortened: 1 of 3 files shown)


wateret (changes requested):
What about adding another test for this?

hyunsik-yoon:
You want to keep `oneOp` test for only one op? I see. :-)

wateret:
@hyunsik-yoon Breaking "one-op" category is OK. I would allow some tests that have more than one ops are in "one-op" test dir. But I would like to keep the model tests that have just one operation. I find them useful when debugging. 😃

hyunsik-yoon:
I moved the same test to `GenModelTests.cc`.
Since I have to add one more negative test, I added a negative test.
Thanks.

wateret (approved):
LGTM

PR #4917, opened 2020-11-03
[luci-interpreter] Support CWQ convolution op 

binarman:
Support convolution with channel-wise quantized weights and
layer-wise quantized activations (input and output of conv layer)

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Conv2D.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Conv2D.test.cpp
@@ -229,16 +229,16 @@ TEST(Conv2DTest, SInt16_CWQ_weights)
       7, 8, // row = 1, col 1
   };
   std::vector<float> filter_data{
-      4,  -3, // out = 0
-      1, -3,  // out = 1
-      5,  -3, // out = 2
+      4, -3, // out = 0
+      1, -3, // out = 1
+      5, -3, // out = 2
   };
   std::vector<float> bias_data{1, 10, 5};
   std::vector<float> ref_output_data{
-      0, 5, 4,   // row 0, col 0
-      1, 1, 8,   // row 0, col 1
-      3, 0, 12,  // row 1, col 0
-      5, 0, 16,  // row 1, col 1
+      0, 5, 4,  // row 0, col 0
+      1, 1, 8,  // row 0, col 1
+      3, 0, 12, // row 1, col 0
+      5, 0, 16, // row 1, col 1
   };
 
   float input_scale = 0.25f;
--- a/compiler/luci-interpreter/src/kernels/Conv2D.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Conv2D.test.cpp
@@ -215,6 +215,62 @@ TEST(Conv2DTest, SInt16)
   EXPECT_THAT(dequantizeTensorData(output_tensor), FloatArrayNear(ref_output_data));
 }
 
+TEST(Conv2DTest, SInt16_CWQ_weights)
+{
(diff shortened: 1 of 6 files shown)


compiler/luci-interpreter/src/kernels/Conv2D.test.cpp
```diff
@@ -215,6 +215,62 @@ TEST(Conv2DTest, SInt16)
...
+      3, 0, 12, // row 1, col 0
+      5, 0, 16, // row 1, col 1
+  };
+
+  float input_scale = 0.25f;
+  float output_scale = 0.05f;
+  std::vector<float> filter_scales = {0.25f, 0.2f, 0.1f};
+  std::vector<float> bias_scales;
+  for (int i = 0; i < filter_scales.size(); ++i)
+    bias_scales.push_back(filter_scales[i] * input_scale);
```
jinevening, line 249: This may not affect the test result, but I leave this comment FYI.

In int16 quantization, bias has only one scale which is the greatest one among `filter_scales[i] * input_scale`.
(https://github.com/Samsung/ONE/blob/master/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp#L236-L240)

This is because of the limitation of some ISA.

binarman: > I should have shared this with you..

This is not a problem, I should ask it myself earlier =)

I am not sure I got the idea.

for now computation of one output element looks like this:
```
for (output channels)
{
  accumulator = 0;
  for (filter spatial dimensions)
  {
    accumulator += input[spatial][channel]*filter[spatial][channel];
  }
  accumulator += bias[channel];  // this line assume that scales of bias and accumulator are identical, otherwise we can not add values like this
  output_data = rescale(accumulator);
}
```

According your statement we should do something like this:
```
for (output channels)
{
  accumulator = 0;
  for (filter spatial dimensions)
  {
    accumulator += input[spatial][channel]*filter[spatial][channel];
  }
  output_data = quantized_sum(accumulator, bias[channel]);
}
```
* quantized_sum represents sum of two quantized numbers with different scales.

I don't see other algorithms, could you point me what I misunderstand?

binarman: When I am talking about equality of bias and partial sum scales I mean this:
for example, we have only one input value, one kernel value and bias:
- in = in_q * in_scale
- k = k_q * k_scale
- b = b_q * b_scale

full sum: in*k + b = (in_q*k_q) * (in_scale*k_scale) + b_q * b_scale

this equal to (in_q*k_q + b_q) * (in_scale*k_scale) only in case b_scale == in_scale*k_scale

jinevening:
```
accumulator += bias[channel];  // this line assume that scales of bias and accumulator are identical, otherwise we can not add values like this
```

Strictly speaking, the above comment is true. But AFAIK, our algorithm team has decided to just add bias, ignoring the scale difference. This is the [related issue](https://github.sec.samsung.net/AIP/Trinity/issues/139). Sorry for the Korean.

@meejeong , Can you check this issue?

jinevening: @binarman Sorry, I misunderstood the issue.

There is no need to maintain one scale (we need one shift amount value and this is irrelevant to the number of scales). So, we should maintain the scale for each channel, i.e., # of scales == # of channels.

Please keep this implementation. I'll change the quantizer. Thanks for pointing out the issue.

jinevening (approved):
LGTM

PR #4919, opened 2020-11-04
[onert/cpu] Use reference for BinaryArithmeticOpParam

hseok-oh:
generateKernelGeneric uses reference to get BinaryArithmeticOpParam parameter
Set default value for all BinaryArithmeticOpParam field


hseok-oh:
It will resolve coverity warning CID 1193045

wateret (approved):
LGTM

PR #4924, opened 2020-11-04
[nnfwapi] Test: revise the way to add TestCaseData

wateret:
Revise the way to add TestCaseData, to make them one-liners for many
cases. `TestCaseData` methods return `this` for convenience.

### DIFF
--- a/tests/nnfw_api/src/one_op_tests/ResizeBilinear.cc
+++ b/tests/nnfw_api/src/one_op_tests/ResizeBilinear.cc
@@ -47,10 +47,9 @@ TEST_F(GenModelTest, OneOp_ResizeBilinear_SizeToVar)
   cgen.setInputsAndOutputs({in, size}, {out});
 
   _context = std::make_unique<GenModelTestContext>(cgen.finish());
-  TestCaseData tcd;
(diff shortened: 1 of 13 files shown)


tests/nnfw_api/src/one_op_tests/ResizeBilinear.cc
```diff
@@ -47,10 +47,9 @@ TEST_F(GenModelTest, OneOp_ResizeBilinear_SizeToVar)
   cgen.setInputsAndOutputs({in, size}, {out});
 
   _context = std::make_unique<GenModelTestContext>(cgen.finish());
-  TestCaseData tcd;
-  tcd.addInput(std::vector<int32_t>{3, 3});
-  tcd.addInput(std::vector<float>{1, 1, 2, 2});
-  tcd.addOutput(std::vector<float>{1, 1, 1, 1.666666667, 1.666666667, 1.666666667, 2, 2, 2});
```
wateret, line 53: It never was added as a test case 😮 . And I tried to enable it but the results do not match.

hseok-oh (approved):
LGTM

PR #4925, opened 2020-11-04
[luci/logex] Log memory address

cgbahk:
This commit adds memory address info for each node on logging.

ONE-DCO-1.0-Signed-off-by: Cheongyo Bahk <cg.bahk@gmail.com>

---

IMO it is sometimes useful to log memory address of nodes.
Use case: when error occurred on node `0xBAADF00D` -> We can directly search in log what and where the node is.

It will looks
```
; Mem = 0x559a39c0bfa0
%0 = circle.CIRCLEINPUT()
; Mem = 0x559a39dbc030
%1 = circle.CIRCLECONST(...)
```


seanshpark (approved):
LGTM

PR #4928, opened 2020-11-04
[luci] Fix bias quantization of int16 CWQ

jinevening:
Per-channel quantization of bias for int16 CWQ


jinevening:
To reviewers: This changes per-tensor bias quantization to per-channel in CWQ.

seanshpark (approved):
LGTM

PR #4931, opened 2020-11-05
[luci] Change traversal order in FusePreActBNPass

jinevening:
This changes traversal order of Add op from (x->y) to (y->x)


jinevening:
To reviewers: This is for testing with internal model (golden data has the switched form).

compiler/luci/pass/src/FusePreActivationBatchNormPass.test.cpp
```diff
@@ -146,6 +156,8 @@ class SimpleGraph
   luci::CircleConst *pred_conv_filter;
   luci::CircleConst *pred_conv_bias;
   luci::CircleConv2D *pred_conv2;
+  luci::CircleConst *pred_conv2_filter;
```
seanshpark, line 159: (maybe in another PR) for static code analysis tool warnings, would be better to initialize with nullptr
```
luci::CircleConst *pred_conv2_filter = nullptr;
```
for this and other members...

seanshpark (approved):
LGTM

Issue #4874, opened 2020-11-02
[circle-quantizer] Quantize FastStyleTransfer model


jinevening:
Let's quantize FastStyleTransfer (FST) model.
(Model repo: https://github.com/lengstrom/fast-style-transfer)

A major issue on quantizing FST is instance normalization. In FST, instance normalization is implemented as below.

![instnorm_pattern](https://user-images.githubusercontent.com/5449554/97844445-61b6c080-1d2e-11eb-8a5f-facd74307ce6.png)

Current `FuseInstanceNormPass` does not capture this pattern, so we need to update the pass to support this new pattern.

**ToDo**
- [x] Update `FuseInstanceNormPass` to support the new pattern
- [x] Support InstanceNorm Op in luci-interpreter #4902
- [x] Support InstanceNorm quantization in circle-quantizer

jinevening:
Quantization failed as luci-interpreter fails to run Conv2D.

This is because a buffer for im2col is too large (about 6.7 GB), thus interpreter fails to allocate the buffer (std::bad_alloc is thrown)).

I've made Conv2D fall back to the reference kernel that does not use im2col (#4933) if memory allocation fails.

jinevening:
Now FST can be quantized with fused IN. I close this PR.

PR #4933, opened 2020-11-05
[luci-interpreter] Fallback logic for Conv2D if bad_alloc

jinevening:
Fall back to reference kernel if memory allocation fails

---
Related to: https://github.com/Samsung/ONE/issues/4874

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Conv2D.cpp
+++ b/compiler/luci-interpreter/src/kernels/Conv2D.cpp
@@ -110,8 +110,15 @@ void Conv2D::configure()
     const int input_depth = input_shape.dim(3);
     Shape im2col_shape{batches, output_height, output_width,
                        input_depth * filter_height * filter_width};
-    _im2col =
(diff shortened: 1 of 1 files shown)


compiler/luci-interpreter/src/kernels/Conv2D.cpp
```diff
@@ -110,8 +110,15 @@ void Conv2D::configure()
...
-    _im2col =
-        std::make_unique<Tensor>(input()->element_type(), im2col_shape, AffineQuantization{}, "");
+    try
+    {
+      _im2col =
+          std::make_unique<Tensor>(input()->element_type(), im2col_shape, AffineQuantization{}, "");
+    }
+    catch (std::bad_alloc &ba)
+    {
+      // Failed memory allocation
```
seanshpark, line 120: maybe add `_im2col = nullptr` ?

jinevening: I've added the assignment statement. Thanks.

seanshpark (approved):
LGTM

PR #4940, opened 2020-11-05
[luci-interpreter] Support DepthwiseConv2D channel-wise quantization

binarman:
Supports channel-wise quantization of weights in DepthwiseConv2D sint16 kernel

ONE-DCO-1.0-Signed-off-by: Alexander Efimov <a.efimov@samsung.com>

PR #4941, opened 2020-11-05
[luci-interpreter] Support Transpose conv with channel-wise quantization 

binarman:
Support channel-wise quantization of weights in Depthwise conv 2d kernel

need merge #4940 first

### DIFF
--- a/compiler/luci-interpreter/src/kernels/TransposeConv.h
+++ b/compiler/luci-interpreter/src/kernels/TransposeConv.h
@@ -20,16 +20,22 @@
 #include "core/Kernel.h"
 #include "core/KernelParams.h"
 
+#include <memory>
+
 namespace luci_interpreter
 {
 namespace kernels
 {
+
+class ChannelQuantMultipliers;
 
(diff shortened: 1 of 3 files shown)


compiler/luci-interpreter/src/kernels/TransposeConv.h
```diff
@@ -52,8 +58,7 @@ class TransposeConv : public KernelWithParams<TransposeConvParams>
   int32_t _padding_width{};
   // The scaling factor from input to output (aka the 'real multiplier') can
   // be represented as a fixed point multiplier plus a left shift.
-  int32_t _output_multiplier = 0;
-  int _output_shift = 0;
+  std::unique_ptr<std::vector<ChannelQuantMultipliers>> _quant_multipliers;
```
jinevening, line 61: Is it necessary to use unique_ptr ? Why not just use a vector?

binarman: You are right, vector is enough. I was careless and left this artifact of some experiments.

jinevening (approved):
LGTM

PR #4943, opened 2020-11-06
[draft] Remove memcpy in reshape

YiHyunJin:
When reshape operating, we don't use memcpy.
Because, input data and output data are same.
So only move memony point.

TODO -
The input buffer should deallocated after use output buffer.
Modified to be usable in neon and acl.

### DIFF
--- a/runtime/onert/backend/cpu/StaticTensorManager.cc
+++ b/runtime/onert/backend/cpu/StaticTensorManager.cc
@@ -42,10 +42,25 @@ void StaticTensorManager::allocateNonconsts(void)
   {
     const auto &ind = pair.first;
     auto tensor = pair.second.get();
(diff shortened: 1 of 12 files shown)


wateret:
Many thanks for your effort! Just in case if you want to land this(just drafting is OK), this commit wouldn't likely get accepted since:

- **The changes in onert core(`ir` and `compiler`) are not necessary.** We can find which operand is an input of Reshape in the backend side. It might have been introduced for convenience, but it gives more responsibilities to the existing classes
- It would be nice to achieve it without `_as_reshape` as we would have something similar with `_as_shape` for many ops(e.g. Relu can be done in-place)
    - Or at least we can unify the way for all in-place operations

% Please also add me to the reviewers, once you make PRs for merge related to this work.

PR #4948, opened 2020-11-07
[luci-interpreter] Support PReLU with channel-wise quantization

binarman:
Support PReLU with CWQ alpha argument and layer-wise quantized input/output

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Prelu.cpp
+++ b/compiler/luci-interpreter/src/kernels/Prelu.cpp
@@ -34,20 +34,41 @@ Prelu::Prelu(const Tensor *input, const Tensor *alpha, Tensor *output)
 {
 }
 
+Prelu::~Prelu()
+{
+  // Destructor declared to delete vector of alpha quantized data properly
+}
+
 void Prelu::configure()
 {
   LUCI_INTERPRETER_CHECK(input()->element_type() == output()->element_type());
   LUCI_INTERPRETER_CHECK(alpha()->element_type() == output()->element_type());
+  LUCI_INTERPRETER_CHECK(input()->scales().size() <= 1);
+  LUCI_INTERPRETER_CHECK(output()->scales().size() <= 1);
 
-  if (input()->element_type() == DataType::U8 || input()->element_type() == DataType::S16)
+  if (input()->element_type() == DataType::U8)
   {
-    if (input()->element_type() == DataType::S16)
-    {
-      LUCI_INTERPRETER_CHECK(input()->zero_point() == 0 && alpha()->zero_point() == 0 &&
-                             output()->zero_point() == 0);
-    }
+    LUCI_INTERPRETER_CHECK(alpha()->scales().size() <= 1);
+    _alpha_multipliers.resize(1);
     double alpha_multiplier = input()->scale() * alpha()->scale() / output()->scale();
-    quantizeMultiplier(alpha_multiplier, &_output_multiplier_alpha, &_output_shift_alpha);
+    quantizeMultiplier(alpha_multiplier, &_alpha_multipliers[0].multiplier,
+                       &_alpha_multipliers[0].shift);
+    double identity_multiplier = input()->scale() / output()->scale();
+    quantizeMultiplier(identity_multiplier, &_output_multiplier_identity, &_output_shift_identity);
+  }
+  else if (input()->element_type() == DataType::S16)
+  {
+    LUCI_INTERPRETER_CHECK(input()->zero_point() == 0 && output()->zero_point() == 0);
+    for (size_t channel = 0; channel < alpha()->zero_points().size(); ++channel)
+    {
+      LUCI_INTERPRETER_CHECK(alpha()->zero_points()[channel] == 0);
+    }
+
+    std::vector<double> real_multipliers =
+        getQuantizedConvolutionMultiplers(input()->scale(), alpha()->scales(), output()->scale());
+
+    _alpha_multipliers = quantizeMultipliers(real_multipliers);
+
     double identity_multiplier = input()->scale() / output()->scale();
     quantizeMultiplier(identity_multiplier, &_output_multiplier_identity, &_output_shift_identity);
   }
@@ -109,8 +130,8 @@ void Prelu::evalQuantized() const
   op_params.output_offset = output()->zero_point();
   op_params.output_shift_1 = _output_shift_identity;
   op_params.output_multiplier_1 = _output_multiplier_identity;
-  op_params.output_shift_2 = _output_shift_alpha;
-  op_params.output_multiplier_2 = _output_multiplier_alpha;
+  op_params.output_shift_2 = _alpha_multipliers[0].shift;
+  op_params.output_multiplier_2 = _alpha_multipliers[0].multiplier;
 
   if (input()->shape() != alpha()->shape())
   {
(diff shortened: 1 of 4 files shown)


jinevening:
It seems that this code assumes alpha of PRelu is quantized per-channel (alpha has multiple scales and zero points). But alpha has only one scale and one zero point (both in cwq and lwq), i.e., it is quantized per-tensor. I think the previous implementation is enough to run int16 quantized PRelu.

AFAIK, there is no plan for per-channel quantization of alpha of PRelu for now, but it may be supported in the future. This PR will be useful at that time.

binarman:
> It seems that this code assumes alpha of PRelu is quantized per-channe

Yes, this implementation assumes that input and output are LWQ and alpha is CWQ.
I created this PR because it is possible to effectively implement such behavior in HW.
Also note that CWQ for input/output is not supported because it seems there are no way to get channel-wise quantized feature maps by other operators anyway.

> I AFAIK, there is no plan for per-channel quantization of alpha of PRelu for now, but it may be supported in the future. This PR will be useful at that time.

Ok! Is it better to close PR or leave it open for now?

jinevening:
> Ok! Is it better to close PR or leave it open for now?

I think we don't have to close this. I think quantization spec may be updated soon. Let's leave it open.

compiler/luci-interpreter/src/kernels/Prelu.test.cpp
```diff
@@ -241,6 +324,56 @@ TEST(PreluTest, Invalid_Input_Type_NEG)
...
+  std::vector<float> scales{1.f, 1.f};
+  std::vector<int32_t> zerop{0, 0};
+  std::vector<float> dummy_data(4, 0.0f);
+  Tensor input_tensor = makeInputTensor<DataType::U8>({2, 2}, 1.f, 0, dummy_data);
+  Tensor alpha_tensor = makeInputTensor<DataType::U8>({2, 2}, scales, zerop, 0, dummy_data);
+  Tensor output_tensor = makeInputTensor<DataType::U8>({2, 2}, 1.f, 0, dummy_data);
+
+  Prelu kernel(&input_tensor, &alpha_tensor, &output_tensor);
+  EXPECT_ANY_THROW(kernel.configure());
+}
```
jinevening, line 364: Can you remove this test? circle-quantizer now support U8 CWQ PRelu (kernel is not yet implemented).

compiler/luci-interpreter/src/kernels/Prelu.cpp
```diff
@@ -128,25 +149,97 @@ void Prelu::evalQuantized() const
...
+  else
+  {
+    tflite::RuntimeShape input_shape = getTensorShape(input());
+    tflite::RuntimeShape alpha_shape = getTensorShape(alpha());
+    tflite::RuntimeShape output_shape = getTensorShape(output());
+    const int16_t *input_data = input()->data<int16_t>();
+    const int16_t *alpha_data = alpha()->data<int16_t>();
+    int16_t *output_data = output()->data<int16_t>();
+
+    constexpr int N = 5;
```
jinevening, line 192: Maybe N = 4?

compiler/luci-interpreter/src/kernels/Prelu.cpp
```diff
@@ -128,25 +149,97 @@ void Prelu::evalQuantized() const
...
+    if (input_shape == alpha_shape)
+    {
+      const int flat_size = tflite::MatchingElementsSize(input_shape, alpha_shape, output_shape);
+      int quantized_dim = alpha()->quantized_dimension();
+
+      int32_t outer_dims_size = 1;
+      for (int i = 0; i < quantized_dim; ++i)
+        outer_dims_size *= input_shape.Dims(i);
+      int32_t quant_dim_size = alpha_shape.Dims(alpha()->quantized_dimension());
+      int32_t inner_dims_size = flat_size / outer_dims_size / quant_dim_size;
```
jinevening, line 205: It seems that this code assumes `quantized_dimension` can be any integer in [0, rank()-1], but there is a restriction in our quantization spec.

In CWQ (**channel-wise** quantization), `quantized_dimension` of alpha must be the channel dimension, which is the last dimension of the tensor (= rank - 1). And the length of `quantized_dimension` must be the same with that of the last dimension of input (=channel size). 

For example, let's assume input shape is [1,4,4,64]. Here, alpha shape is something like [1,1, 64], and the `quantized_dimension` is 2 (rank - 1). If alpha shape is [64], then the quantized_dimension will be 0 (rank - 1).

So, `inner_dims_size` in this code is always 1.

compiler/luci-interpreter/src/kernels/Prelu.cpp
```diff
@@ -128,25 +149,97 @@ void Prelu::evalQuantized() const
...
+    auto fn = [this](int16_t input_val, int16_t alpha_val) {
+      const ChannelQuantMultipliers pos_mult{_output_shift_identity, _output_multiplier_identity};
+      const ChannelQuantMultipliers &neg_mult = _alpha_multipliers[0];
+      return evalElemS16Prelu(input_val, alpha_val, pos_mult, neg_mult);
+    };
+
+    BinaryOpBroadcastSlow(getTensorShape(input()), getTensorData<int16_t>(input()),
+                          getTensorShape(alpha()), getTensorData<int16_t>(alpha()),
+                          getTensorShape(output()), getTensorData<int16_t>(output()), fn);
+  }
```
jinevening, line 182: We only support CWQ for s16, so I think this code can be removed.

compiler/luci-interpreter/src/kernels/Prelu.cpp
```diff
@@ -128,25 +149,97 @@ void Prelu::evalQuantized() const
...
+                                                             identity_mult.shift)
+                     : tflite::MultiplyByQuantizedMultiplier(
+                           input_val * alpha_val, alpha_mult.multiplier, alpha_mult.shift);
+  const int32_t clamped_output = std::min(quantized_max, std::max(quantized_min, output_val));
+  return clamped_output;
+}
+
+void Prelu::evalQuantizedS16() const
+{
+  bool channel_wise = alpha()->scales().size() != 1;
```
jinevening, line 170: This code is misleading when the channel size is one. It would be better to check whether alpha is quantized per-channel as below.

```cpp
LUCI_INTERPRETER_CHECK(alpha()->scales().size() == alpha()->shape().dim(alpha()->quantized_dimension()));
```

compiler/luci-interpreter/src/kernels/Prelu.cpp
```diff
@@ -128,25 +149,97 @@ void Prelu::evalQuantized() const
...
+      assert(alpha_shape.DimensionsCount() <= N);
+      assert(output_shape.DimensionsCount() <= N);
+
+      tflite::NdArrayDesc<N> desc1{};
+      tflite::NdArrayDesc<N> desc2{};
+      tflite::NdArrayDesc<N> output_desc{};
+      tflite::NdArrayDescsForElementwiseBroadcast(input_shape, alpha_shape, &desc1, &desc2);
+      tflite::CopyDimsToDesc(tflite::RuntimeShape::ExtendedShape(N, output_shape), &output_desc);
+
+      const int quantized_dim = alpha()->quantized_dimension() + N - alpha()->shape().num_dims();
```
jinevening, line 232: If the model was quantized correctly, `alpha()->quantized_dimension() + N - alpha()->shape().num_dims()` would be always `N - 1`.

compiler/luci-interpreter/src/kernels/Prelu.cpp
```diff
@@ -128,25 +149,97 @@ void Prelu::evalQuantized() const
...
+            output_data[offset] =
+                evalElemS16Prelu(input_data[offset], alpha_data[offset], pos_mult, neg_mult);
+          }
+        }
+    }
+    else
+    {
+      assert(input_shape.DimensionsCount() <= N);
+      assert(alpha_shape.DimensionsCount() <= N);
+      assert(output_shape.DimensionsCount() <= N);
```
jinevening, line 224: Maybe N = 4 ? (currently N=5)

binarman: I took this limitation somewhere from tflite (but not from prelu, probably it was some binary operation, like sub).
Current implementation in tflite 2.3 sets N to 4, so you are right.

binarman:
@jinevening 
Sorry for long delay.

I checked specification and simplified code as much as I could.
Assumptions i used:
- all dimensions of alpha shape should be equal to 1, except the last one
- the last one alpha dimension is quantized
- number of elements in alpha is equal last dimension of input and output tensors (it is not checker anywhere for now)

Could you verify that this matches your vision?

compiler/luci-interpreter/src/kernels/Prelu.cpp
```diff
@@ -34,20 +34,52 @@ Prelu::Prelu(const Tensor *input, const Tensor *alpha, Tensor *output)
...
-      LUCI_INTERPRETER_CHECK(input()->zero_point() == 0 && alpha()->zero_point() == 0 &&
-                             output()->zero_point() == 0);
+      LUCI_INTERPRETER_CHECK(alpha()->zero_points()[channel] == 0);
     }
-    double alpha_multiplier = input()->scale() * alpha()->scale() / output()->scale();
-    quantizeMultiplier(alpha_multiplier, &_output_multiplier_alpha, &_output_shift_alpha);
+    // Prelu specific checks for CWQ
+    LUCI_INTERPRETER_CHECK(alpha()->quantized_dimension() == alpha()->shape().num_dims() - 1);
+    LUCI_INTERPRETER_CHECK(alpha()->scales().size() ==
+                           alpha()->shape().dim(alpha()->quantized_dimension()));
```
jinevening, line 70: Compilation failed in this line. Can you fix the issue?

jinevening:
> Sorry for long delay.

No problem. It was my fault that did not clarify the quantization spec. Sorry for that.

> Could you verify that this matches your vision?

Yes, they match with our quantization spec :) It would be even better to add a check for the last assumption somewhere (maybe in `configure`). Thanks.

jinevening (approved):
LGTM Thank you!

Issue #4944, opened 2020-11-06
[rawdata2hdf5] Prepare rawdata2hdf5 to be included in 1.11.0


jinevening:
Let's prepare `rawdata2hdf5` to be included in 1.11.0.

- [x] More error handling
- [x] Tests in one-cmds
- [x] Document (how-to-create-hdf5-dataset.txt)

jinevening:
I close this issue as all tasks are done.

PR #4953, opened 2020-11-09
[rawdata2hdf5] Preparation to be included in release

jinevening:
This includes changes for release
1. Generate rawdata2hdf5 (dropped .py)
2. Improve exception handling

---
Related to: https://github.com/Samsung/ONE/issues/4944

### DIFF
--- a/compiler/rawdata2hdf5/rawdata2hdf5
+++ b/compiler/rawdata2hdf5/rawdata2hdf5
@@ -83,7 +83,7 @@ def create_hdf5(data_list, output_path):
                 if os.path.isfile(filename):
                     datalist.append(filename)
                 else:
-                    raise Exception("No such file. " + filename)
+                    raise FileNotFoundError("No such file. " + filename)
 
     # Input files
     num_converted = 0
--- a/compiler/rawdata2hdf5/rawdata2hdf5
+++ b/compiler/rawdata2hdf5/rawdata2hdf5
@@ -24,7 +24,6 @@ import numpy as np
 import argparse
 import glob
 import os
-import sys
 
 
 def get_parser():
@@ -105,21 +104,13 @@ def create_hdf5(data_list, output_path):
 
 
 def main():
-    try:
-        parser = get_parser()
-        args = parser.parse_args()
-        verify_args(parser, args)
-        create_hdf5(args.data_list, args.output_path)
-        return 0
+    parser = get_parser()
 
-    except UnicodeDecodeError:
-        print("UnicodeDecode Error. data_list may not be the text file.")
-        return 1
+    args = parser.parse_args()
 
-    except Exception as e:
-        print(e)
-        return 1
+    verify_args(parser, args)
 
+    create_hdf5(args.data_list, args.output_path)
 
 if __name__ == '__main__':
-    sys.exit(main())
+    main()
--- a/compiler/rawdata2hdf5/rawdata2hdf5
+++ b/compiler/rawdata2hdf5/rawdata2hdf5
@@ -71,7 +71,7 @@ def create_hdf5(data_list, output_path):
     group.attrs['rawData'] = '1'
 
     if os.path.isfile(data_list) == False:
-        raise Exception("No such file. " + data_list)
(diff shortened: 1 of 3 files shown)


seanshpark (approved):
LGTM

compiler/rawdata2hdf5/rawdata2hdf5
```diff
@@ -99,17 +105,21 @@ def create_hdf5(data_list, output_path):
...
+        print(e)
+        return 1
 
 
 if __name__ == '__main__':
-    try:
-        main()
-    except UnicodeDecodeError:
-        raise SystemExit("UnicodeDecode Error. data_list may not be the text file.")
+    sys.exit(main())
```
mhs4670go, line 125: I'm not sure but this try-catch statement is necessary? If you raise the error in the function, it will be exit with desired return value.

jinevening: > If you raise the error in the function, it will be exit with desired return value.

At first I wrote the code as you said, but the code did not work correctly during testing.

The test codes save logs as follows.

```
rawdata2hdf5 \
--data_list ${inputfile} \
--output_path ${outputfile} > ${filename}.log
```

When I raised an error in the function, the log was not written to `${filename}.log` (don't know why..). The logs were correctly written to the file when I use the function `print`.

mhs4670go: @jinevening Please add `2>&1` at the end of the command.
```bash
$ rawdata2hdf5 \
--data_list ${inputfile} \
--output_path ${outputfile} > ${filename}.log 2>&1
```

jinevening: > @jinevening Please add 2>&1 at the end of the command.

@mhs4670go  Thanks for the suggestion. But is it ok to use unhandled exception?

In cpp program (circle-quantizer), the release test team did not accept unhandled exceptions.

If unhandled exception is ok in Python program, I'll remove try-catch statement.

The log looks as below.

```
hjj@hjj-ubuntu:~/ONE.2/install/test$ ../bin/rawdata2hdf5 -l wronglist.txt -o out.h5
  File "../bin/rawdata2hdf5", line 118, in <module>
    main()
  File "../bin/rawdata2hdf5", line 115, in main
    create_hdf5(args.data_list, args.output_path)
  File "../bin/rawdata2hdf5", line 88, in create_hdf5
    raise Exception("No such file. " + filename)
Exception: No such file. **non-existing-file.data**
```

CC  @seanshpark

jinevening:
After conversation with @seanshpark , we decided to remove try-catch. I've updated the code PTAL @mhs4670go

seanshpark (approved):
LGTM

compiler/rawdata2hdf5/rawdata2hdf5
```diff
@@ -66,7 +71,7 @@ def create_hdf5(data_list, output_path):
     group.attrs['rawData'] = '1'
 
     if os.path.isfile(data_list) == False:
-        raise SystemExit("No such file. " + data_list)
+        raise Exception("No such file. " + data_list)
```
mhs4670go, line 74: Maybe this error would be more specific.
```suggestion
        raise FileNotFoundError("No such file. " + data_list)
```

compiler/rawdata2hdf5/rawdata2hdf5
```diff
@@ -78,7 +83,7 @@ def create_hdf5(data_list, output_path):
                 if os.path.isfile(filename):
                     datalist.append(filename)
                 else:
-                    raise SystemExit("No such file. " + filename)
+                    raise Exception("No such file. " + filename)
```
mhs4670go, line 86: ditto

jinevening, line 86:
```suggestion
                    raise FileNotFoundError("No such file. " + filename)
```

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #4955, opened 2020-11-09
[one-cmds] Add tests for rawdata2hdf5

jinevening:
This adds tests for rawdata2hdf5

---
Related to: #4944


compiler/one-cmds/tests/prepare_test_materials.sh
```diff
@@ -49,6 +43,51 @@ if [[ ! -s "bcq.pb" ]]; then
...
+if [[ ! -s "img_files" ]]; then
+    rm -rf img_files
+    wget https://github.com/Samsung/ONE/files/5499172/img_files.zip
+    unzip img_files.zip
+fi
+
+# prepare preprocessing script and raw data files for quantization test
+if [[ ! -s "preprocess.py" ]]; then
+    rm -f preprocess.py
+    cat > preprocess.py << EOF
```
mhs4670go, line 55: Just curiosity, is there any reason why you generates `preprocess.py` in script? It could be added as a python file.

jinevening: No special reason. I put the code in this script simply because the name of the script is `prepare_test_materials.sh`.

If a separate file looks better, I'll move this code to the new file.

seanshpark: If possible, can you embed the function of preprocess.py inside prepare_test_materials.sh ?
you can add the codes as a function if you like.

seanshpark: maybe just provide `prepare_hdf5files.py` ?

mhs4670go: @jinevening Well, it depends on your preference.
FYI, in my opinion, `cat` command with redirect uses when logging the outputs that is derived from this commands in our repo like this.
```bash
cat > "${PREFIX}.log" <(
      exec 2>&1

      rm -f "${PASSED_TAG}"

      set -ex
      # Create a pb model
      "${TFKIT_PATH}" encode "${TESTCASE_DIR}/test.pbtxt" "${PREFIX}.pb"

      # Do inference
      "${NNKIT_RUN_PATH}" \
        --backend "${NNKIT_TF_BACKEND_PATH}" \
        --backend-arg "${PREFIX}.pb" \
        --backend-arg "${TESTCASE_DIR}/test.info"
      set +ex

      touch "${PASSED_TAG}"
    )   
```
second, `.py` will be more readable in ide and make it easy to maintain and format check later.

jinevening: > second, .py will be more readable in ide and make it easy to maintain and format check later.

It would be better to make a separate file. I'll move the code to a new file named `preprocess_images.py`.

CC @seanshpark

jinevening: I've updated the code PTAL :)

seanshpark:
Please do not change file attributes: `compiler/one-cmds/tests/prepare_test_materials.sh  100644 → 100755`
And please check for new files too.

jinevening:
> Please do not change file attributes: compiler/one-cmds/tests/prepare_test_materials.sh 100644 → 100755

My mistake.. I've recovered the attribute. Thanks.

compiler/one-cmds/tests/preprocess_images.py
```diff
@@ -0,0 +1,34 @@
...
+list_file = 'datalist.txt'
+
+if os.path.exists(output_dir):
+    shutil.rmtree(output_dir, ignore_errors=True)
+os.makedirs(output_dir)
+
+for (root, _, files) in os.walk(input_dir):
+    datalist = open(list_file, 'w')
+    for f in files:
+        image = PIL.Image.open(root + '/' + f)
```
mhs4670go, line 28: Is it okay not to close after opening? With "with" block might helpful.

jinevening: I've updated to use 'with'. Thanks.

mhs4670go (approved):
LGTM

seanshpark:
I'll merge this after #4953 lands

seanshpark (approved):
LGTM

PR #4870, opened 2020-11-02
[Draft] Use ruy library on Conv2D FP32

periannath:
- This commit uses ruy library on Conv2D FP32

ONE-DCO-1.0-Signed-off-by: JiHwan Yeo <jihwan.yeo@samsung.com>

---

Related issue : #4863

Issue #4863, opened 2020-11-02
[onert] Optimize Conv2D in cpu backend


periannath:
onert uses `Eigen` library to compute Conv2D layer but it is slower than `ruy` library.

Let's optimize Conv2D kernel by using `ruy` library.

---

Draft is done. Let's upload PRs.

- [x] Update im2col condition #4956
- [x] Add Conv2D GenModelTest #5019
- [x] Introduce Conv2D ruy kernel in ~cker~ `compute/ruy` #5023
- [x] Introduce RUY backend #5067

periannath:
As discussed offline, I'll introduce Conv2D ruy kernel as new backend.

periannath:
### How to use ruy for Conv2D
- `OP_BACKEND_Conv2D=ruy BACKENDS="cpu;ruy"` is required to use ruy backend on Conv2D operation
  - `OP_BACKEND_Conv2D=ruy` : Use ruy backend on all Conv2D operations
  - `BACKENDS="cpu;ruy"` : Loads cpu and ruy backend and use cpu backend as default backend

seanshpark:
@periannath , if possible can you add some documents about using `ruy` backend?
- maybe at https://github.com/Samsung/ONE/tree/master/docs/howto

periannath:
@seanshpark I'm not sure what kind of documentation you want to be added. `How to use ruy backend`? or `Implementation details about ruy backend`? 

We don't have any document about backend right now. Should `ruy` backend be special case for documentation? I think general documentation about backend should be added first.

seanshpark:
> I'm not sure what kind of documentation you want to be added.

Just leaving https://github.com/Samsung/ONE/issues/4863#issuecomment-730088944 in this issue would be hard for others to know about this. As of start, just with this text content is ok, for me :) Later your or me or others can add more things in the document.

periannath:
@seanshpark Okay. Then, I'll add document about backend configuration. (Including `OP_BACKEND_[name]` and `OP_BACKEND_MAP`)

PR #4956, opened 2020-11-09
[onert] Update condition to use im2col in Convolution layer

periannath:
- This PR updates condition to use im2col in Convolution layer
  - Im2col is needed in two cases, dilated_im2col and non_dilated_im2col
  - Runtime allocates im2col buffer only for non_dilated_im2col
- Adds allocation of im2col buffer for dilated_im2col

---

For issue #4863 
From draft #4870


hseok-oh (approved):
LGTM

PR #4958, opened 2020-11-09
[luci] Add ShapeSignature attribute in CircleNode

llFreetimell:
Parent Issue : #4372

This commit will add `ShapeSignature` attribute in `CircleNode`

### DIFF
--- a/compiler/luci/lang/include/luci/IR/CircleNodeDecl.h
+++ b/compiler/luci/lang/include/luci/IR/CircleNodeDecl.h
@@ -20,6 +20,7 @@
 #include <loco/IR/Dialect.h>
 #include <loco/IR/Node.h>
 #include <loco/IR/NodeMixins.h>
+#include <luci/IR/CircleShapeSignature.h>
 #include <luci/IR/PropertyShapeStatus.h>
 
 #include "CircleOpcode.h"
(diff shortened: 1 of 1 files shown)


compiler/luci/lang/include/luci/IR/CircleNodeDecl.h
```diff
@@ -61,6 +62,9 @@ struct CircleNode : public loco::Node,
     _sparsityparam = std::move(sparsityparam);
   }
 
+  ShapeSignature shape_signature(void) const { return _shape_signature; }
+  void shape_signature(ShapeSignature ss) { _shape_signature = ss; }
```
seanshpark, line 66: compared to `ShapeStatus`, `ShapeSignature` is a class with `std::vector` member.
These two will always make a one more more copies of this member when using these.
Is this your intention?

llFreetimell: I didn't thought about this point, I will fix it. Thanks!

seanshpark: `void shape_signature(ShapeSignature ss) { _shape_signature = ss; }` vs
`void shape_signature(ShapeSignature &ss) { _shape_signature = ss; }` : can you describe the difference?

llFreetimell: It was good chance to study about it :)

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

Issue #4966, opened 2020-11-10
[onert] Test fail by #4718


hseok-oh:
Test fail by #4718

Last 50 lines of output:
```[...truncated 10783 lines...]
[2020-11-10T03:35:41.166Z] -- Installing: /home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/armv7l-linux.release/out/unittest_standalone/tflite_test
[2020-11-10T03:35:41.166Z] -- Set runtime path of "/home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/armv7l-linux.release/out/unittest_standalone/tflite_test" to "$ORIGIN/../lib:$ORIGIN/"
[2020-11-10T03:35:41.167Z] -- Installing: /home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/armv7l-linux.release/out/bin/opencl_info
[2020-11-10T03:35:41.167Z] -- Set runtime path of "/home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/armv7l-linux.release/out/bin/opencl_info" to "$ORIGIN/../lib:$ORIGIN/:/home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/armv7l-linux.release/overlay/lib"
[2020-11-10T03:35:41.167Z] make[1]: Leaving directory '/home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/armv7l-linux.release/obj'
[2020-11-10T03:35:41.167Z] rm -rf /home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/out
[2020-11-10T03:35:41.167Z] ln -s /home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/armv7l-linux.release/out /home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/out
[2020-11-10T03:35:41.167Z] touch /home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/armv7l-linux.release/INSTALL
[2020-11-10T03:35:41.167Z] cp /home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/armv7l-linux.release/overlay/lib/libarm_compute*.so /home/jenkins_slave/workspace/nnfw/master/push-nnpackage-test/Product/out/lib || true
[2020-11-10T03:35:41.167Z] packaging test suite
[...]

PR #4970, opened 2020-11-10
[onert] Fix nnpackage_run with load data fail

hseok-oh:
Don't resize input when load input data shape is same with model

---

For #4966


seanshpark (approved):
LGTM

hyunsik-yoon (approved):
LGTM

PR #4971, opened 2020-11-10
[onert] Test for set_input_tensorinfo() with a model having two inputs

hyunsik-yoon:
Add tests for `set_input_tensorinfo()` with a model having two inputs 
and make sure some combination of calling `set_input_tensorinfo()` to each inputs work OK.

Parent issue: #4625
After #4718


wateret (approved):
LGTM

PR #4850, opened 2020-10-30
[Draft][luci] Refactoring luci/Service

llFreetimell:
Parent Issue : #4796
Related Issue : #4372
**NOTE : First commit is based on #4685**

---

This draft is for refctoring `luci/Service` for removing dependency with `loco`

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

PR #4979, opened 2020-11-10
[luci] Introduce inference rule of shape signature

llFreetimell:
Parent Issue : #4372 

This commit will introduce `CircleShapeSignatureInferenceRule`.

### DIFF
--- a/compiler/luci/service/src/CircleShapeSignatureInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeSignatureInferenceRule.cpp
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Service/CircleShapeSignatureInferenceRule.h"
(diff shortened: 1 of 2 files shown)


compiler/luci/service/src/CircleShapeSignatureInferenceRule.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+  ShapeSignatureInferenceAlgorithm alg;
+
+  shape_signature = circle_node->accept(&alg);
+
+  VERBOSE(l, 1) << "[luci] Shape Signature( " << circle_node->name() << " )";
+  VERBOSE(l, 1) << "    before: " << circle_node->shape_signature();
+  VERBOSE(l, 1) << "     after: " << shape_signature;
+
+  return true;
```
seanshpark, line 57: can we change to void return type? don't know usage of always return true here.

llFreetimell: The reason why I set return type to `bool` was that we should return `false` when inference of shape_signature was failed because of some reason.
For now, as I couldn't find such cases, it returns always `true` for now.
However, I am not sure that there are no such failure cases in the future.

FYI, current shape/dtype inference always return `true`.

https://github.com/Samsung/ONE/blob/872e4e85afd06096be1d383ba769597823f4f46e/compiler/luci/service/src/CircleShapeInferenceRule.cpp#L2486-L2514

This will be changed like following codes in #4850

```cpp
bool New_CircleShapeInferenceRule::infer(const luci::CircleNode *circle_node,
                                         loco::TensorShape &shape) const
{
  LOGGER(l);
  VERBOSE(l, 1) << "[luci] shape: " << circle_node->name();
  VERBOSE(l, 1) << "              before: " << ShapeInferenceHelper::circle_shape(circle_node);

  if (!inputs_shape_ready(circle_node))
  {
    VERBOSE(l, 1) << " -> after: shape of some inputs are not ready to be inferenced";
    return false;
  }

  New_ShapeInferenceAlgorithm alg;
  shape = circle_node->accept(&alg);
  VERBOSE(l, 1) << " -> after: " << shape;

  return true;
}
```

seanshpark (approved):
LGTM

PR #4980, opened 2020-11-10
[tflchef] Add testcase for shape signature

llFreetimell:
Parent Issue : #4372

This commit will add a testcase for shape signature in `tflchef`


seanshpark (approved):
LGTM

Issue #4268, opened 2020-09-14
Compiler FE: Support more TF 2.x formats


seanshpark:
Time to prepare support of saved model and keras model formats.
- [x] tf2tfliteV2.py 
- [x] test materials
- [x] one-import-tf
- [x] ~revise documents~
- [x] add tests

seanshpark:
SavedModel: https://www.tensorflow.org/guide/saved_model

seanshpark:
SavedModel example
- https://www.tensorflow.org/api_docs/python/tf/saved_model/Asset#example

execution result
```
INFO:tensorflow:Assets written to: /tmp/saved_model/assets
tf.Tensor(b'Hello World!\n', shape=(), dtype=string)
```
```
$ cd /tmp/new_location
$ tree
.
├── assets
│   └── file.txt
├── saved_model.pb
└── variables
    ├── variables.data-00000-of-00001
    └── variables.index

2 directories, 4 files
```
`saved_model.pb` read from netron:
![image](https://user-images.githubusercontent.com/4616940/93149237-c8592e00-f730-11ea-87ee-26fc10cffc95.png)

seanshpark:
Saved model example
- https://www.tensorflow.org/api_docs/python/tf/saved_model/save#example_usage

a little modification: add name to input, set input shape to (3, 3)
```py
import tensorflow as tf

class Adder(tf.Module):

  @tf.function(input_signature=[tf.TensorSpec(shape=(3,3), dtype=tf.float32, name="input")])
  def add(self, x):
    return x + x + 1.

to_export = Adder()

tf.saved_model.save(to_export, '/tmp/adder')
```

saved model pb
- ![image](https://user-images.githubusercontent.com/4616940/93153340-0576ed80-f73c-11ea-8fa3-fdfcbd025c4a.png)
- this is main graph
- there are several sub graphs
- `__inference_add_134` sub graph looks like this;
- ![image](https://user-images.githubusercontent.com/4616940/93153465-47079880-f73c-11ea-9071-5e559bfa3172.png)

seanshpark:
Actually we don't need to provide input, output node names for `saved_model`

seanshpark:
Will continue after #4278 is done

seanshpark:
Time to resume :)

seanshpark:
python code to create test saved_model;

```py
import tensorflow as tf

class Adder(tf.Module):
[...]

PR #4981, opened 2020-11-10
DRAFT: one-import-tf support saved_model and keras

seanshpark:
on-going draft to support saved_model and keras model in one-import-tf


seanshpark:
For #4268

seanshpark:
all landed

PR #4983, opened 2020-11-10
[one-cmds] Add URL info for test materials

seanshpark:
This will add github issue URL info for test materials where the
download source exists


jinevening (approved):
LGTM

PR #4984, opened 2020-11-10
[one-cmds] Fix actions for image preprocess

seanshpark:
This will fix action and condition to invoke preprocess_images script


jinevening (approved):
LGTM Thanks

PR #4988, opened 2020-11-11
[one-cmds] add model_format option to cfg

mhs4670go:
This commit adds model_format option to cfg.


seanshpark (approved):
LGTM

PR #4991, opened 2020-11-11
[tests/nnapi]Update unittest to build for android

struss:
This commit update unittest to build for android.

1. Update Path info matching with android (/tmp~ -> /data/local/tmp~) which only apply on android build
2. Add skiplist file based on aarch64 linux file.
3. update Cmake to include boost.


hseok-oh (approved):
LGTM

PR #5003, opened 2020-11-11
[tflchef] Fix type check

seanshpark:
This will fix type check namespace in MaxPoolWithArgMax


mhs4670go (approved):
LGTM

PR #5009, opened 2020-11-11
[one-cmds] update cfg files

mhs4670go:
This commit updates test cfg files for applying converter_version option
that is just been introduced.


seanshpark (approved):
LGTM

PR #4659, opened 2020-10-16
[DRAFT] [onert] Eliminate copies of subgraph call/return

wateret:
- Do not handle a primary graph specially
    - Make model input/outputs `IOTensor` rather than `UserTensor`
    - Also a preparation for NNAPI 1.3 support (Any subgraph can be primary)

ONE-DCO-1.0-Signed-off-by: Hanjoung Lee <hanjoung.lee@samsung.com>

PR #5010, opened 2020-11-11
[onert] Eliminate copies when subgraph call/return

wateret:
- Remove redundant copies during controlflow op execution
    - `If` - All copies are removed
    - `While` - Some copies are still there, temporary tensors are
      allocated/freed during execution
- Do not handle a primary graph specially
    - Make model input/outputs `IOTensor` rather than `UserTensor`
    - Also a preparation for NNAPI 1.3 support (Any subgraph can be
      primary)

Draft : #4659

### DIFF
--- a/runtime/onert/core/src/backend/controlflow/IOTensor.cc
+++ b/runtime/onert/core/src/backend/controlflow/IOTensor.cc
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "IOTensor.h"
+
+#include <assert.h>
+
+namespace onert
+{
+namespace backend
+{
+namespace controlflow
+{
+
+IOTensor::IOTensor(const ir::OperandInfo &info, ir::Layout layout)
+    : IPortableTensor{info}, _orig_info{info}, _orig_layout{layout}
+{
+  setUserTensor(nullptr, 0);
+}
+
+void IOTensor::setTensor(IPortableTensor *tensor)
+{
+  assert(tensor);
+  assert(tensor != this);
+  // TODO Handle when layout was changed
+  assert(tensor->layout() == _orig_layout); // Changing layout is not considered yet
+  _user_tensor.reset();
+  _tensor = tensor;
+}
+
+void IOTensor::setUserTensor(uint8_t *buffer, size_t size)
+{
+  _user_tensor = std::make_unique<UserTensor>(_orig_info, _orig_layout, buffer, size);
+  _tensor = _user_tensor.get();
+}
+
+} // namespace controlflow
+} // namespace backend
+} // namespace onert
--- a/runtime/onert/core/src/backend/controlflow/IOTensor.h
+++ b/runtime/onert/core/src/backend/controlflow/IOTensor.h
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __ONERT_BACKEND_CONTROLFLOW_INDIRECT_TENSOR_H__
+#define __ONERT_BACKEND_CONTROLFLOW_INDIRECT_TENSOR_H__
+
+#include "backend/IPortableTensor.h"
+#include "UserTensor.h"
+
+namespace onert
+{
+namespace backend
+{
+namespace controlflow
+{
+
+/**
+ * @brief Tensor object that indirects to the tensor it is pointing to.
+ *
+ * A model I/O tensor could be two types.
+ *
+ * 1. @c UserTensor, if it is the primary graph
+ * 2. Any other derivative of @c IPortableTensor from another backend, otherwise
+ *
+ * To support these, this object indirects everything to the actual tensor pointer.
+ * Exceptionally if it is UserTensor, this class creates and manages it.
+ */
+class IOTensor : public IPortableTensor
+{
+public:
+  IOTensor(const ir::OperandInfo &info, ir::Layout layout);
+
+public:
+  void setTensor(IPortableTensor *tensor);
+  void setUserTensor(uint8_t *buffer, size_t size);
+  ir::OperandInfo orig_info() const { return _orig_info; }
+  ir::Layout orig_layout() const { return _orig_layout; }
+
+public:
+  uint8_t *buffer() const override { return _tensor->buffer(); }
+  size_t total_size() const override { return _tensor->total_size(); }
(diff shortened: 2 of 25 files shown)


wateret:
Performance test : https://github.com/Samsung/ONE/pull/4659#issuecomment-716945449

wateret:
@hyunsik-yoon Sorry about such a huge diff, but it is hard to separate. 😢

runtime/onert/core/src/backend/controlflow/kernel/WhileLayer.cc
```diff
@@ -61,159 +61,88 @@ void WhileLayer::run()
...
   };
 
+  std::vector<ITensor *> op_inputs(_input_tensors.begin(), _input_tensors.end());
+  std::vector<ITensor *> op_outputs(_output_tensors.begin(), _output_tensors.end());
+  // Copying body inputs to outputs when the loop body is never executed
+  if (!getResultCond(cond_output_tensor.get()))
+  {
+    PermuteLayer copy_body_inputs_to_op_outputs{op_inputs, op_outputs, _external_context};
+    copy_body_inputs_to_op_outputs.run();
+    // return;
```
hyunsik-yoon, line 93: should this line be uncommented?

runtime/onert/core/src/exec/ExecutorBase.cc
```diff
@@ -73,29 +56,29 @@ void ExecutorBase::execute(const std::vector<backend::ITensor *> &src_tensors,
...
-      const auto orig_input_shape = input_tensor->getShape();
-      const auto changed_input_shape =
-          convertShape(src_tensor->getShape(), src_tensor->layout(), input_tensor->layout());
-      if (orig_input_shape != changed_input_shape)
-      {
-        input_tensor->set_dynamic();
-      }
-    }
+    assert(input_tensor != nullptr);
+    input_tensor->set_dynamic();
```
hyunsik-yoon, line 68: Previous code call `set_dynamic` only when "src shape != subgraph input shape"
Shouldn't we need such check here?

runtime/onert/core/src/exec/ExecutorBase.cc
```diff
@@ -111,33 +94,32 @@ void ExecutorBase::execute(const IODescription &desc)
...
-    auto *tensor = dynamic_cast<backend::controlflow::UserTensor *>(_output_tensors[i]);
-    assert(tensor);
-    tensor->set_dynamic(); // It can't be resized but shape could change
+    auto tensor = _output_tensors[i];
+
     if (desc.outputs[i] == nullptr)
       throw std::runtime_error{"Output " + std::to_string(i) + "'s buffer is not set."};
-    tensor->setBuffer(static_cast<uint8_t *>(desc.outputs[i]->buffer), desc.outputs[i]->size);
+    tensor->setUserTensor(static_cast<uint8_t *>(desc.outputs[i]->buffer), desc.outputs[i]->size);
+    tensor->set_dynamic(); // It can't be resized but shape could change
```
hyunsik-yoon, line 122: Could you explain why we need this line all the time?

runtime/onert/core/src/backend/controlflow/IOTensor.h
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * A model I/O tensor could be two types.
+ *
+ * 1. @c UserTensor, if it is the primary graph
+ * 2. Any other derivative of @c IPortableTensor from another backend, otherwise
+ *
+ * To support these, this object indirects everything to the actual tensor pointer.
+ * Exceptionally if it is UserTensor, this class creates and manages it.
+ */
+class IOTensor : public IPortableTensor
```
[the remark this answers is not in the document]
wateret: I'm not sure if I got you right, I think renaming would just resolve the issue. Either `controlflow` -> `builtin`(`builtin::IOTensor`) or the tensor `IOTensor` -> `SubgraphIOTensor`(`controlflow::SubgraphIOTensor`).

> maybe we may need to have controlflow and builtin backend respectively....

I think just renaming `controflow` backend to `builtin` is enough, as mentioned above. Having two different builtin backends(just to have better names) would not be worth it IMHO. 😓 

I think, if `controlflow::IOTensor` sounds confusing, `controlflow::UserTensor` is also confusing. So what I would suggest is **to rename `controlflow` to `builtin`, and that's all**. What do you think about it? 😃

runtime/onert/core/src/backend/controlflow/IOTensor.cc
```diff
@@ -0,0 +1,52 @@
...
+  setUserTensor(nullptr, 0);
+}
+
+void IOTensor::setTensor(IPortableTensor *tensor)
+{
+  assert(tensor);
+  assert(tensor != this);
+  // TODO Handle when layout was changed
+  assert(tensor->layout() == _orig_layout); // Changing layout is not considered yet
+  _user_tensor.reset();
```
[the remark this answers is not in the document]
wateret: Yes it is correct. As I mentioned offline, a subgraph can be called as both primary(UserTensor) or non-primary, after compilation. So for each run they could call any of `setTensor` or `setUserTensor`.

runtime/onert/core/src/exec/ExecutorBase.cc
wateret: It was introduced months ago and I forgot 😅 . I will find out why later. I feel like https://github.com/Samsung/ONE/pull/5010/files/738b9d6a9b056e85e81aff2fe1ba77f2e5e3df07#r522001183 is also relavant.

runtime/onert/core/src/exec/ExecutorBase.cc
wateret: I may have removed it for testing, and forgot. I just put it back.

wateret: Turns out it does not work, need some time to fix!

wateret:
@hyunsik-yoon Please take another look

runtime/onert/core/src/backend/controlflow/IOTensor.h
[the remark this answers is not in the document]
hyunsik-yoon: I agree.

runtime/onert/core/src/exec/ExecutorBase.cc
hyunsik-yoon: It's hard to follow which one is fixed. :crying_cat_face:  I guess this fix is done.

runtime/onert/core/src/exec/ExecutorBase.cc
hyunsik-yoon: Thinking with `WHILE`,
my guess is that... if the all outputs of body_subgraph become dynamic with this line even though some of them are static, probably all inputs of the `WHILE` op will be turned into dynamic, which will make most tensor inside cond and body subgraph dynamic (even though they can be handled in static way)... 
and I am afraid that this is not aligned with the semantics of current dynamic tensor.

wateret:
@hyunsik-yoon I added a commit - https://github.com/Samsung/ONE/pull/5010/commits/f878a1313671a3f106dcab7a5de238ec20bfe00f

runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -278,13 +272,9 @@ ExecutorFactory::createLinearExecutor(std::unique_ptr<compiler::LoweredGraph> lo
...
   runTensorRegistration(lowered_graph.get(), order);
 
-  std::vector<backend::ITensor *> input_tensors;
-  std::vector<backend::ITensor *> output_tensors;
-  if (options.is_primary_subgraph)
-  {
-    input_tensors = initializeModelIOTensors(*lowered_graph, lowered_graph->graph().getInputs());
-    output_tensors = initializeModelIOTensors(*lowered_graph, lowered_graph->graph().getOutputs());
-  }
+  initializeModelIOTensors(
```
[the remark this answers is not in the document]
wateret: Will fix!

runtime/onert/core/src/exec/ExecutorBase.cc
wateret: Anyways, this code was exist already(a few lines above), so this is beyond the scope of this PR.

runtime/onert/core/src/exec/ExecutorBase.cc
wateret: Yes it's back - https://github.com/Samsung/ONE/pull/5010/commits/f878a1313671a3f106dcab7a5de238ec20bfe00f

hyunsik-yoon (dismissed):
LGTM

wateret:

ragmani:

ragmani:

wateret:
@hyunsik-yoon Rebased the commit due to the CI failure.

hyunsik-yoon (dismissed):
LGTM

wateret:
@hyunsik-yoon Sorry, another rebase was done 😅

hyunsik-yoon (approved):
LGTM

Issue #4969, opened 2020-11-10
[onert] Provide Config with nnpackage


seanshpark:
We have a method to configure onert behaviour by setting environment variables.
- `runtime/onert/core/include/util/Config.lst` file

But this does not work for Tizen as it's prohibited.

This functionality may provide a way to control onert per application or per model.
- Providing through adding a metadata file in nnpackage could be a good candidate.

Any sugessions are welcomed.

seanshpark:
Some traces...
- `ONERT_LOG_ENABLE` is accessed before execution at `logging.h`, `static Context &ctx = Context::get();`

seanshpark:
done

PR #5011, opened 2020-11-11
DRAFT: config file in nnpackage

seanshpark:
on-going draft to support config file in nnpackage

ONE-DCO-1.0-ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@samsung.com>

### DIFF
--- a/nnpackage/examples/one_op_in_tflite/metadata/MANIFEST
+++ b/nnpackage/examples/one_op_in_tflite/metadata/MANIFEST
@@ -1,7 +1,8 @@
 {
   "major-version" : "1",
-  "minor-version" : "0",
+  "minor-version" : "1",
   "patch-version" : "0",
+  "configs"     : [ "config.cfg" ],
   "models"      : [ "add.tflite" ],
   "model-types" : [ "tflite" ]
 }
--- a/nnpackage/spec/10_packaging_and_manifest.md
+++ b/nnpackage/spec/10_packaging_and_manifest.md
@@ -18,11 +18,13 @@ For `model` and `custom_op`, see [20_model_and_operators.md](20_model_and_operat
 nnpackage
 ├── custom_op
 ├── metadata
-│   └── MANIFEST
+│   ├── MANIFEST
+│   └── config.cfg
 └── mymodel.model
 ```
 
 - `mymodel.model` is a model file that has computation graph and weights.
+- `config.cfg` is a configuration file that has parameters to configure onert.
 - `metadata` is a directory that contains all metadata including `MANIFEST`.
 - `MANIFEST` is a collection of attributes about this package.
 - `custom_op` is a directory that contains implementation objects.
@@ -61,6 +63,11 @@ For detail, see [semantic versioning 2.0.0](https://semver.org/)
 
 `patch-version` is the patch version of `nnpackage`.
 
+#### configs
+
+`configs` is an array of configuration file names placed in `metadata` folder. This can be empty or
+attribute itself can be omitted. As of now we only support only one item.
+
 #### models
 
 `models` is an array of path to model files, which is relative path from top level directory of this package.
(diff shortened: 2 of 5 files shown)


seanshpark:
For #4969

nnpackage/examples/one_op_in_tflite/metadata/MANIFEST
```diff
@@ -1,7 +1,8 @@
 {
   "major-version" : "1",
-  "minor-version" : "0",
+  "minor-version" : "1",
   "patch-version" : "0",
+  "config"        : "config.cfg",
```
glistening, line 5: @seanshpark What about using a list like `"configs" : [ "config.cfg" ]` for multiple configuration? We may use 1st item by default or explicitly provides the index for others.

nnpackage/spec/10_packaging_and_manifest.md
```diff
@@ -18,11 +18,13 @@ For `model` and `custom_op`, see [20_model_and_operators.md](20_model_and_operat
...
 ├── custom_op
 ├── metadata
-│   └── MANIFEST
+│   ├── MANIFEST
+│   └── config.cfg
 └── mymodel.model
 ```
 
 - `mymodel.model` is a model file that has computation graph and weights.
+- `config.cfg` is a configuratin file that has parameters to configure onert.
```
glistening, line 27:
```suggestion
- `config.cfg` is a configuration file that has parameters to configure onert.
```

nnpackage/spec/10_packaging_and_manifest.md
```diff
@@ -61,6 +63,11 @@ For detail, see [semantic versioning 2.0.0](https://semver.org/)
 
 `patch-version` is the patch version of `nnpackage`.
 
+#### config
+
+`config` is configuration file name placed in `metadata` folder. Value can be empty or attribute
```
glistening, line 68:
```suggestion
`configs` is an array of configuration file names placed in `metadata` folder. Value can be empty or attribute
```

nnpackage/examples/one_op_in_tflite/metadata/MANIFEST
seanshpark: OK, nice to have multiples. Then we need an index to provide... do you any idea how to provide the index ?
maybe another item like `"cfxidx" : "2",` ?

seanshpark: Or is it like one configuration file for one model ?

glistening: @seanshpark the index of "configs" list could be used like we did in "models". We need to choose whether the configuration will be applied to all models in an nnpackage or per-model. If latter, we need to provide 2-dimension array since we allow multiple models in an nnpackage.

glistening: > OK, nice to have multiples. Then we need an index to provide... do you any idea how to provide the index ?
> maybe another item like `"cfxidx" : "2",` ?

Maybe I've misunderstood. If you're talking how the user can select the index, we need to extend or provide a new API.

seanshpark: `config` here is to override environment variables, like `RUY_THREADS=4` or `RUY_THREADS=1`
while looking at `ConfigSource.cc`, when one configure is loaded to provide `getConfigOrDefault()` values, 
I don't know how we can change it before running a model... so per-model is a bit hard to implement righnow for me.

glistening: @seanshpark I think your suggestion is better since it does not require API work. :)

```
"models" : [ "m1.tflite", "m2.circle", "m3.circle", "m4.tflite" ]
"configs" : [ ["m1cfg"], [ ] , ["m3debug", "m3mixed" ] ]
"config-idx" : [ 0, -1, 1, -1 ] 
```
- `config-idx` is the zero-based index from `configs`.
- `-1` in `config-idx` means _"use default configuration"_.

seanshpark: From off-line discussion, as for now, we'll go with array of `configs` and array of `config-idx` with one element for now. extending this will be left for the future.

seanshpark: For `config-idx` it should be always `[ 0 ]` for now, I'll leave this implementation for the future when this is needed.

seanshpark:
all landed

PR #5013, opened 2020-11-11
[one-cmds] Test for TF2.x models

seanshpark:
This will introduce one-import tests for TF2.x saved model and keras model

### DIFF
--- a/compiler/one-cmds/tests/one-import_003.test
+++ b/compiler/one-cmds/tests/one-import_003.test
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+# Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


seanshpark:
For #4268, from draft #4981

compiler/one-cmds/tests/one-import_003.test
```diff
@@ -0,0 +1,42 @@
...
+
+trap trap_err_onexit ERR
+
+configfile="one-import_003.cfg"
+outputfile="test_saved_model.circle"
+
+rm -f ${outputfile}
+
+# run test
+one-import tf -C ${configfile} > /dev/null
```
mhs4670go, line 36: there is no overriding option,,

seanshpark: no need to override. this is just to test saved_model import.

mhs4670go: @seanshpark Oh, I saw Line17 `# positive usage with overriding option`. Maybe not intended?

mhs4670go (approved):
LGTM

PR #5016, opened 2020-11-11
[onert] Use anonymous namespace for static functions

seanshpark:
This will wrap with anonymous namespace for static functions in
nnfw_api_internal.cc


wateret (approved):
LGTM

PR #5017, opened 2020-11-12
[onert/util] introduce external config_source

seanshpark:
This will introduce external config_source that can override default configuration values

### DIFF
--- a/runtime/onert/core/src/util/ConfigSource.cc
+++ b/runtime/onert/core/src/util/ConfigSource.cc
@@ -30,8 +30,10 @@ namespace util
 {
 
 static std::unique_ptr<IConfigSource> _source;
+static std::unique_ptr<IConfigSource> _source_ext;
 
 void config_source(std::unique_ptr<IConfigSource> &&source) { _source = std::move(source); }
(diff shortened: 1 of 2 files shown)


runtime/onert/core/src/util/ConfigSource.cc
```diff
@@ -30,8 +30,10 @@ namespace util
 {
 
 static std::unique_ptr<IConfigSource> _source;
+static std::unique_ptr<IConfigSource> _source_ext;
 
 void config_source(std::unique_ptr<IConfigSource> &&source) { _source = std::move(source); }
+void config_source_ext(std::unique_ptr<IConfigSource> &&source) { _source_ext = std::move(source); }
```
seanshpark, line 36: I'm not sure existing `config_source()` is used from some external projects.
If `config_source()` is not used then we can change this
```cxx
void config_source(std::unique_ptr<IConfigSource> &&source) { _source_ext = std::move(source);
```
and NOT add new `config_source_ext()`

seanshpark:
For #4969, from draft #5011

chunseoklee (approved):

PR #5023, opened 2020-11-12
[onert/compute] Introduce compute/ruy to handle ruy kernels

periannath:
- This commit introduces compute/ruy to handler ruy kernels
  - It has Conv2D ruy kernel that will be used by ruy backend

---

This kernel is tested on draft #4870

For issue #4863

### DIFF
--- a/compute/ruy/CMakeLists.txt
+++ b/compute/ruy/CMakeLists.txt
@@ -1,0 +1,11 @@
+nnfw_find_package(Ruy REQUIRED)
+
+add_library(nnfw_lib_ruy INTERFACE)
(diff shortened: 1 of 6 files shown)


ragmani (dismissed):

periannath:
As discussed offline, I'll move Conv2D ruy kernel into `compute/ruy`. This kernel is only used in ruy backend and it doesn't have to be placed inside `compute/cker`.

periannath:
@Samsung/nnfw I've updated to introduce ruy kernel in `compute/ruy`. PTAL

ragmani (approved):

YongseopKim (approved):
LGTM

PR #5024, opened 2020-11-12
[onert/nnfw_api] Remove duplicate include

seanshpark:
This will remove duplicate include of same header


seanshpark:
For #4969, from draft #5011

hseok-oh (approved):
LGTM

PR #5025, opened 2020-11-12
[onert/api] ConfigSource from nnpkg configs

seanshpark:
This will enable ConfigSource from nnpkg configs file if exist

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -19,6 +19,7 @@
 #include "compiler/Compiler.h"
 #include "util/ConfigSource.h"
 #include "util/Exceptions.h"
+#include "util/logging.h"
 #include "exec/Execution.h"
 #include "circle_loader.h"
 #include "tflite_loader.h"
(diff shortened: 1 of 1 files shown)


seanshpark:
For #4969, from draft #5011

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -95,6 +96,64 @@ NNFW_STATUS getTensorIndexImpl(const onert::ir::Graph &graph, const char *tensor
...
+}
+
+void setConfigKeyValues(const CfgKeyValues &keyValues)
+{
+  auto configsrc = std::make_unique<onert::util::GeneralConfigSource>();
+
+  for (auto it = keyValues.begin(); it != keyValues.end(); ++it)
+  {
+    VERBOSE(NNPKG_CONFIGS) << "(" << it->first << ") = (" << it->second << ")" << std::endl;
+    configsrc->set(it->first, it->second);
```
chunseoklee, line 151: If invalid (key,value) is provided, it will be added without side effect. Am I right ?

chunseoklee (approved):

runtime/onert/api/src/nnfw_api_internal.cc
seanshpark: Yup...

PR #5030, opened 2020-11-12
[model2nnpkg] Provide configs option

seanshpark:
This will revise model2nnpkg to provide configuration as an option

### DIFF
--- a/tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
+++ b/tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
@@ -70,7 +70,7 @@ extension=${modelfile##*.}
 echo "Generating nnpackage "$name" in "$outdir""
 mkdir -p "$outdir"/"$name"/metadata
 
-if [ ! -z "$config_src" ] && [ -s "$config_src" ]; then
+if [ -s "$config_src" ]; then
   config=$(basename "$config_src")
   cp "$config_src" "$outdir/$name/metadata/$config"
 fi
--- a/tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
+++ b/tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
@@ -70,7 +70,7 @@ extension=${modelfile##*.}
 echo "Generating nnpackage "$name" in "$outdir""
 mkdir -p "$outdir"/"$name"/metadata
 
-if [ ! -z $config_src ] && [ -s $config_src ]; then
+if [ ! -z "$config_src" ] && [ -s "$config_src" ]; then
   config=$(basename "$config_src")
   cp $config_src "$outdir"/"$name"/metadata/"$config"
 fi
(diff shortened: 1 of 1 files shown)


seanshpark:
For #4969, from draft #5011

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
```diff
@@ -64,11 +69,18 @@ extension=${modelfile##*.}
 
 echo "Generating nnpackage "$name" in "$outdir""
 mkdir -p "$outdir"/"$name"/metadata
+
+if [ ! -z $config_src ] && [ -s $config_src ]; then
```
glistening, line 73:
```suggestion
if [ ! -z "$config_src" ] && [ -s "$config_src" ]; then
```

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
```diff
@@ -64,11 +69,18 @@ extension=${modelfile##*.}
...
+  config=$(basename "$config_src")
+  cp $config_src "$outdir"/"$name"/metadata/"$config"
```
glistening, line 75:
```suggestion
  cp "$config_src" "$outdir/$name/metadata/$config"
```

glistening (commented):
For arguments of string with space (e.g. "hello world"), changes are required.

seanshpark:
@glistening , thanks!

glistening (approved):
LGTM

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
```diff
@@ -64,11 +69,18 @@ extension=${modelfile##*.}
 
 echo "Generating nnpackage "$name" in "$outdir""
 mkdir -p "$outdir"/"$name"/metadata
+
+if [ ! -z "$config_src" ] && [ -s "$config_src" ]; then
```
mhs4670go, line 73: `[ ! -z "$config_src" ]` condition might be redundant.
```suggestion
if [ -f "$config_src" ]; then
```

glistening: @mhs4670go 

I also thought ` -z ` might be redundant. However, I am not sure it is equivalent.
Also,  your suggestion `-f` allows zero-size config file while `-s` will not allow zero-size config file.

mhs4670go: @glistening Actually, I wanted to check whether it is a directory or not. But just `-s` option seems enough I think. And, I think `-z` option is unnecessary here.
```bash
if [ -s "$config_src" ]; then
```

seanshpark: `$config_src` is optional. it can be null if not given. `-z` is to check if this option is provided and config file is given.

mhs4670go: @seanshpark Right. s option checks what z option does as well.

seanshpark: OK, `-s` seems to work

glistening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5031, opened 2020-11-12
[luci] Add validation for ShapeSignature

llFreetimell:
Parent Issue : #4372

This commit will add validation for `ShapeSignature` in `luci`


compiler/luci/service/src/Validate.cpp
```diff
@@ -119,7 +168,8 @@ bool validate(loco::Graph *g)
   if (!validate_shape_dtype(g))
     return false;
 
-  // TODO add more validation
```
seanshpark, line 122: I think this should be left as-is for future validations

seanshpark (approved):
LGTM

PR #5032, opened 2020-11-12
[luci] Export ShapeSignature only when size is larger than 0

llFreetimell:
Parent Issue : #4372

If rank of `ShapeSignature` is 0, we do not need to export it.


seanshpark (approved):
LGTM

PR #5038, opened 2020-11-12
[onert] Remove backend API IExternalContext

wateret:
Remove `IExternalContext` since it is never used by runtime core.

## I WOULD LIKE TO GET 2+ REVIEWS BEFORE MERGING


runtime/onert/core/include/backend/IExternalContext.h
```diff
@@ -1,34 +0,0 @@
...
-
-namespace onert
-{
-namespace backend
-{
-
-struct IExternalContext
-{
-  virtual ~IExternalContext() = default;
-  virtual void setMaxNumThreads(int) = 0;
```
wateret, line 28: @periannath Is this supposed to be used by onert core?

ragmani (approved):

runtime/onert/core/include/backend/IExternalContext.h
periannath: No, it isn't. It will be only used by backends.

periannath (approved):
LGTM

Issue #5037, opened 2020-11-12
[luci] Implementation of shape signature inference rule


llFreetimell:
# DO NOT WORK THIS FOR A WHILE. IMPORTANT CHANGE WILL BE APPLIED

### Background

This issue is originated from #4372.

### Goal

- Implementation of `ShapeSiganture` inference rule for each `CircleNode`
  - Most of algorithms should be able to be re-used at shape inference algorithm.
- Add testcases for each `ShapeSignature` inference rule
  - You can add already existing code. For example, if there was `Add_000`, you can add `Add_dynamic_000` with `shape_signature`

### How to

1. Select an operation/operations to implement and write your name in the list of table at **assignee**.
    - You can select multiple operations which logic is similar so that PR can be reviewed at once. (ex: ReLU and ReLU6)
    - **Only reserve as much as you can finish in one week. DO NOT BE GREEDY.**
1. Implement each operations as following rules
    - If (operation is virtual)
      - Implement inference rule
    - Else If (operation is circle only)
      - Implement inference rule
      - (If needed, do this first) Add the recipe to `common-artifacts/exclude.lst` if needed.
      - Add testcases to `res/CircleRecipes`
    - Else
      - Implement inference rule
      - (If needed, do this first) Add the recipe to `common-artifacts/exclude.lst` if needed.
      - Add testcases to `res/TensorFlowLiteRecipes`
1. Create PR(s). Then write the PR number in the table at
    - **PR(rule)**, for inference rule implementation
    - **PR(exclude)**, for adding recipe to `common-artifacts/exclude.lst`. If not needed, put :no_entry_sign:.
    - **PR(res)**, for adding recipe to `res/`. **If some recipe is needed to be added to `exclude.lst`, that should be done first**
1. After PR is merged, mark :heavy_check_mark: in the table at **Status**

__NOTE__ You can define a new helper function which can be used in another operation's inference rule.

### Examples (Guide)

- Normal operation : #5073(rule) / #5072(exclude) --> #5079(res)
- Virtual operation : #5039(rule) / exclude, res : Not Needed

llFreetimell:
## Operation Table

| Operation                  | Assignee               | PRs    | Status |
|----------------------------|------------------------|--------|--------|
| Abs |      |      |      |
| Add |      |      |      |
| AddN |      |      |      |
| ArgMax |      |      |      |
| ArgMin |      |      |      |
| AveragePool2D |      |      |      |
| BatchMatMul |      |      |      |
[...]

PR #5039, opened 2020-11-12
[luci] Shape signature inference algorithm for Input,Output

llFreetimell:
Parent Issue : #5037

This commit introduces `ShapeSignatureInferenceAlgorithm` for

- `CircleInput`
- `CircleOutput`

and `ShapeSignatureInferenceHelper`.

### DIFF
--- a/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
+++ b/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
@@ -22,7 +22,7 @@ namespace luci
 namespace ssinf
 {
 
-ShapeSignature signature_of_input(const luci::CircleNode *node, uint32_t index)
+ShapeSignature signature_of_arg(const luci::CircleNode *node, uint32_t index)
 {
   auto circle_input = loco::must_cast<luci::CircleNode *>(node->arg(index));
   return circle_input->shape_signature();
--- a/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
+++ b/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Service/CircleShapeSignatureInferenceHelper.h"
+
+namespace luci
+{
+
+namespace ssinf
+{
+
+ShapeSignature signature_of_input(const luci::CircleNode *node, uint32_t index)
+{
+  auto circle_input = loco::must_cast<luci::CircleNode *>(node->arg(index));
+  return circle_input->shape_signature();
+}
+
+} // namespace ssinf
+
+} // namespace luci
--- a/compiler/luci/service/src/Nodes/CircleOutput.cpp
+++ b/compiler/luci/service/src/Nodes/CircleOutput.cpp
@@ -21,7 +21,7 @@ namespace luci
 
 ShapeSignature ssinf::Algorithm::visit(const luci::CircleOutput *node)
 {
-  return signature_of_input(node, 0);
+  return signature_of_arg(node, 0);
 }
 
 } // namespace luci
--- a/compiler/luci/service/src/Nodes/CircleOutput.cpp
+++ b/compiler/luci/service/src/Nodes/CircleOutput.cpp
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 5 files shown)


compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Service/CircleShapeSignatureInferenceHelper.h"
+
+namespace ShapeSignatureInferenceHelper
+{
+
+luci::ShapeSignature signature_of_Nth_input(const luci::CircleNode *node, uint32_t index)
```
seanshpark, line 22: do we need `_Nth_` in name? why not just `input_signature`  ? or `signature_of_input` ?

llFreetimell: I will change to `signature_of_input` :)

compiler/luci/service/src/Nodes/CircleOutput.cpp
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <luci/Service/CircleShapeSignatureInferenceRule.h>
+
+namespace luci
+{
+
+ShapeSignature ShapeSignatureInferenceAlgorithm::visit(const luci::CircleOutput *node)
```
seanshpark, line 22: Is there any particular reason to put one op each files?

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Service/CircleShapeSignatureInferenceHelper.h"
+
+namespace ShapeSignatureInferenceHelper
```
seanshpark, line 19: This is a very long namespace name and is camel case.
We usually used lowe case, short name.
I don't have any particular perference but this is strange.

compiler/luci/service/src/Nodes/CircleOutput.cpp
llFreetimell: Yes, final goal is to put implementations of `ShapeInferenceAlgorithm`, `TypeInferenceAlgorithm`, `ShapeSignatureInferenceAlgorithm` for each nodes in one file.
Here is [an example](https://github.com/Samsung/ONE/blob/245c0b3c68a7f18e259582126b8f88bfb35fe632/compiler/luci/service/src/Nodes/CircleRelu.cpp) in draft code :)

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
llFreetimell: At first, I was not sure whether I can put these functions to `luci`.
If you don't mind, I can fix as following codes.

```cpp
namespace luci
{

luci::ShapeSignature signature_of_input(const luci::CircleNode *node, uint32_t index);

} // namespace luci
```

compiler/luci/service/src/Nodes/CircleOutput.cpp
seanshpark: But wuy?

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
seanshpark: > At first, I was not sure whether I can put these functions to luci.

codes exist in `luci` folder. I am more curious what may bother you putting in `luci` namespace...

seanshpark:
As we talked offline, 
- let's use namespace `luci` instead of long name
- put each nodes in one file is OK. putting the helpers in one file or in some multiple files is also OK
- `ShapeSignatureInferenceAlgorithm` and `ShapeSignatureInferenceHelper` is a bit long: we may use inner namepsace
   - for example; `namespace luci { namespace ss { class InferAlgo; class Helper; `

llFreetimell:
During trying prototype, I got following thinking.

- `ShapeSignatureInferenceHelper` will not be used in a source code since we decided to use `namespace luci`.
- But how about using `ShapeSignatureInferenceHelper` only in filename, since file name usually includes full names and even more long file names are already used.
- Changing `ShapeSignatureInferenceAlgorithm` to `ShapeSigInferAlgo` or `ShapeSigInferenceLogic`
  - I think that it is not the best idea. But if we use inner namespace such as `namespace ss`, we may use `namespace s` and `namespace t` for shape and dtype considering consistency.
- Changing `CircleShapeSignatureInferenceRule` to `ShapeSigInferRule` or `CircleShapeSigInferRule` or `CircleShapeSigInferenceRule`
  - Similar with above item

llFreetimell:
After #5052 is merged, this PR will be changed and rebased.

compiler/luci/service/src/Nodes/CircleOutput.cpp
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+#include <luci/Service/CircleShapeSignatureInference.h>
+
+namespace luci
+{
+
+ShapeSignature ssinf::Algorithm::visit(const luci::CircleOutput *node)
+{
+  return signature_of_input(node, 0);
```
seanshpark, line 24: Maybe `signature_of_output` ?

seanshpark: Um.... `input` of `signature_of_input` seems confusing for same word `input`
If you intent to use `input` as `argument`, how about rename the function as `signature_of_arg` or `signature_of_argument` ?

llFreetimell: I will change the name to `signature_of_arg` :)

seanshpark (approved):
LGTM thank you!

PR #5040, opened 2020-11-13
[onert] Tidy JSONExecTime names

seanshpark:
This will tidy JSONExecTime method names and some comments

### DIFF
--- a/runtime/onert/core/src/exec/ExecTime.h
+++ b/runtime/onert/core/src/exec/ExecTime.h
@@ -94,7 +94,7 @@ public:
   /**
    * @brief Update metrics file with new data.
    */
-  void uploadOperationsExecTime() const { _json.uploadOperationsExecTime(); }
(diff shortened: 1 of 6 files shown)


runtime/onert/core/src/exec/ExecTime.h
```diff
@@ -94,7 +94,7 @@ class ExecTime
   /**
    * @brief Update metrics file with new data.
    */
-  void uploadOperationsExecTime() const { _json.uploadOperationsExecTime(); }
+  void storeOperationsExecTime() const { _json.storeOperationsExecTime(); }
```
seanshpark, line 97: `upload` gave me some thoughts that `upload to somewhere?` but actually it saves to `_measurement_file` file. we can also choose `save` or `update` if suits better

wateret: I agree.

wateret (approved):
LGTM

PR #5041, opened 2020-11-13
[onert] Use variable for controlflow ID

seanshpark:
This will use variable for controlflow::Config::ID for better readability


wateret (approved):
LGTM

PR #5043, opened 2020-11-13
[luci] Introduce operator== for ShapeSignature

llFreetimell:
Parent Issue : #4372

This commit will introduce `operator==` for `ShapeSignature`

### DIFF
--- a/compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
+++ b/compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
@@ -46,6 +46,8 @@ private:
   std::vector<int32_t> _shape_signature{};
 };
 
+bool operator==(const ShapeSignature &lhs, const ShapeSignature &rhs);
+
 } // namespace luci
 
 #endif // __LUCI_IR_SHAPE_SIGNATURE_H__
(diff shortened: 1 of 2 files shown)


seanshpark (approved):
LGTM

compiler/luci/lang/include/luci/IR/CircleShapeSignature.h
```diff
@@ -46,6 +46,8 @@ class ShapeSignature
   std::vector<int32_t> _shape_signature{};
 };
 
+bool operator==(const ShapeSignature &lhs, const ShapeSignature &rhs);
```
mhs4670go, line 49: (optional) This one looks good as well:-)
```suggestion
...
public:
bool operator==(const ShapeSignature &rhs)
{
  return _shape_signature == rhs._shape_signature;
}

//bool operator==(const ShapeSignature &lhs, const ShapeSignature &rhs)
//{
//  if (lhs.rank() != rhs.rank())
//    return false;
//
//  for (uint32_t i = 0; i < lhs.rank(); ++i)
//    if (lhs.dim(i) != rhs.dim(i))
//      return false;
//
//  return true;
//}
```

mhs4670go (approved):
LGTM

PR #5044, opened 2020-11-13
[onert] Log for manual_scheduler_options

seanshpark:
This will revise LOG of Compiler to show manual_scheduler_options and
also backend_for_all


seanshpark:
Actually prints like
```
[Compiler] manual backend_for_all   :
[Compiler] manual_scheduler_options : BCQFullyConnected=bcq, Permute=controlflow, BCQGather=bcq, If=controlflow, While=controlflow
```
before
```
[Compiler]  manual_scheduler_options : (Too many things to print)
```

wateret (approved):
LGTM

PR #5045, opened 2020-11-13
[onert] Introduce WHENLOGENABLED for log only codes

seanshpark:
This will introduce WHENLOGENABLED macro to wrap only for logging codes
- use for logging ManualScheduler results

### DIFF
--- a/runtime/onert/core/include/util/logging.h
+++ b/runtime/onert/core/include/util/logging.h
@@ -64,4 +64,10 @@ static Context &ctx = Context::get();
   if (::onert::util::logging::ctx.enabled()) \
   std::cout << "[" << __func__ << "] "
 
+#define WHENLOGENABLED(METHOD)               \
+  if (::onert::util::logging::ctx.enabled()) \
+  {                                          \
+    METHOD                                   \
+  }
+
 #endif // __ONERT_UTIL_LOGGING_H__
--- a/runtime/onert/core/include/util/logging.h
+++ b/runtime/onert/core/include/util/logging.h
@@ -66,8 +66,9 @@ static Context &ctx = Context::get();
 
 #define WHEN_LOG_ENABLED(METHOD)             \
   if (::onert::util::logging::ctx.enabled()) \
-  {                                          \
-    METHOD                                   \
-  }
+    do                                       \
+    {                                        \
+      METHOD;                                \
+  } while (0)
 
 #endif // __ONERT_UTIL_LOGGING_H__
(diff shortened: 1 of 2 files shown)


runtime/onert/core/include/util/logging.h
```diff
@@ -64,4 +64,10 @@ static Context &ctx = Context::get();
   if (::onert::util::logging::ctx.enabled()) \
   std::cout << "[" << __func__ << "] "
 
+#define WHENLOGENABLED(METHOD)               \
```
glistening, line 67: @seanshpark What about `WHEN_LOG_ENABLED`? Or `onert` may have some convention for macro name though I don't remember.

runtime/onert/core/include/util/logging.h
```diff
@@ -64,4 +64,10 @@ static Context &ctx = Context::get();
...
+  if (::onert::util::logging::ctx.enabled()) \
+  {                                          \
+    METHOD                                   \
+  }
+
```
glistening, line 72: @seanshpark Perhaps enclosing with `do/while` would be safer.

```c
do { \
  if (::onert::util::logging::ctx.enabled()) { \
    ...
  } \
} while (0);

Comment (seanshpark) [reply runtime/onert/core/include/util/logging.h:72]: I've followed above `VERBOSE` macro. I think using difference implementation would be better in another PR.

Comment (seanshpark) [reply runtime/onert/core/include/util/logging.h:72]: Maybe 
```cxx
if (::onert::util::logging::ctx.enabled()) \
  do {
    METHOD
  } while (0);
```
?

runtime/onert/core/include/util/logging.h
```diff
@@ -64,4 +64,11 @@ static Context &ctx = Context::get();
   if (::onert::util::logging::ctx.enabled()) \
   std::cout << "[" << __func__ << "] "
 
+#define WHEN_LOG_ENABLED(METHOD)             \
+  if (::onert::util::logging::ctx.enabled()) \
+    do                                       \
+    {                                        \
+      METHOD;                                \
+  } while (0)
```
seanshpark, line 72: format tool did this...

runtime/onert/core/include/util/logging.h
glistening: > Maybe
> 
> ```c++
> if (::onert::util::logging::ctx.enabled()) \
> do { METHOD } while (0);
> ```
> 
> ?

I suggested to enclose `if ...` with `do/while` because I would like to avoid cases like:

```c
if (condition)
  WHEN_LOG_ENABLED(statementA);
else
  foo();
```

It is imaginary (unlikely) example. However, I would like to ensure.

> I've followed above `VERBOSE` macro. I think using difference implementation would be better in another PR.

Okay. `VERBOSE` seems to have similar problem. It may be beyond this PR.

glistening (approved):
LGTM

hseok-oh (approved):
LGTM

Issue #4702, opened 2020-10-21
[Test] Android Test Script for runtime


struss:
Currently in `infra/script` there is only build script for android. what about adding test script for android like tizen?

struss:
Currently tizen runtime testing script is `infra/script/tizen_xu4_test.sh`

on `tizen_xu4_test.sh`, this run a some test using `tests/scripts/onert-test` with each back-end, interpreter and mixed back-end.

1. `tests/scripts/onert-test unittest`
    this run a `./Product/out/unittest/nnapi_gtest`
2. `test/onert-test verify-tflite --api=nnapi`
    this run a `./Product/out/bin/nnapi_test`
3. `test/onert-test verify-tflite --api=loader`
    this run a `./Product/out/bin/tflite_loader_test_tool`

And also, nnapi verification test for running
`./Product/out/bin/tflite_run` with `USE_NNAPI=1`

those things are test for android. if any test to recommend, please comment......

Thanks for reading.

struss:
using android build script, all test above are created except for `unittest`

struss:
I think we need to update `CMakeLists.txt` on `tests/nnapi`. we need to add boost on that txt.

hseok-oh:
>I think we need to update CMakeLists.txt on tests/nnapi. we need to add boost on that txt.

tests/nnapi (`nnapi_gtest`) don't uses boost library.

hseok-oh:
>tests/nnapi (nnapi_gtest) doesn't use boost library.

@struss I found that some code use boost library for c++17 feature. `nnapi_gtest` also needs boost library (but header-only).

struss:
@hseok-oh Sorry for update late. then, Is it okay for update `CMakeLists.txt` to include boost on `test/nnapi`?

hseok-oh:
@struss Yes. No problem.

struss:
@hseok-oh A little question about `tests/nnapi`. that tests are built only when compiler uses GNU. but on built for android, it use a clang compiler.
https://github.com/Samsung/ONE/blob/cc458e5f841a8b8c2a3ed8457fc8c12c2730f9a2/infra/nnfw/cmake/CfgOptionFlags.cmake#L18-L23
So, to build this test for android, I think we need to update this. I just thought that updating 
```
CMAKE_COMPILER_IS_GNUCC
```
to accept both `GNU` and `Clang` with below.
```
[...]

PR #5055, opened 2020-11-16
[test/nnfw_api]Fix init to use returned value from dirname

struss:
For #4702 
on Android test, `dirname(path)` return char* of directory and not update path to directory.
So, Change to use returned value.


hseok-oh (approved):
LGTM

periannath (approved):
LGTM

PR #4903, opened 2020-11-03
[onert] DRAFT: Make Chrome Tracing handle multiple subgraphs

hyunsik-yoon:
DRAFT/WIP

Parent issue: #4901

Currently Chrome Tracing works only with one subgraphs.
This will make Chrome Tracing handle multiple subgraphs.

Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>

PR #5057, opened 2020-11-16
[onert] Make ctor of ir::Graph take subgraph index param

hyunsik-yoon:
Make ctor of `ir::Graph` take subgraph index param.

- Parent issue: #4901
- Draft: https://github.com/Samsung/ONE/pull/4903

Background:

To produce right profiling information about multiple subgraphs at execution time, each executor needs to know which _subgraph_ they are executing. The proper place to keep the subgraph index seems to be `ir::Graph`.
The subgraph index in `ir::Graph` will be passed to `EventObservers` and displayed to users.

(If we don't want to keep the index in Graph, we need a map of subgraph index and Graph object, which is created at loading time and passed all the way to execution time, which seems complicated.)

### DIFF
--- a/runtime/onert/core/include/ir/Graph.h
+++ b/runtime/onert/core/include/ir/Graph.h
@@ -52,7 +52,13 @@ private:
   };
 
 public:
-  Graph(void);
+  /**
+   * @brief Construct a new ir::Graph object
+   *
+   * @param subg_index Subgraph index from tflite or circle subgraph
+   *                   Pass 0 if a model has only 1 subgraph. (0 means primary subgraph)
+   */
+  Graph(uint32_t subg_index);
   ~Graph(void);
 
   // Graph Building
@@ -103,6 +109,11 @@ public:
   const std::shared_ptr<Subgraphs> &subgraphs() const { return _subgraphs; }
   std::shared_ptr<Subgraphs> &subgraphs() { return _subgraphs; }
   Layout layout() const { return _layout; }
+  SubgraphIndex subg_index() const
+  {
+    assert(!_subg_index.undefined());
+    return _subg_index;
+  }
 
 private:
   Phase _phase{Phase::BUILDING};
@@ -116,6 +127,9 @@ private:
   std::shared_ptr<Subgraphs> _subgraphs;
   // TFLite and circle's default layout is NHWC;
   Layout _layout{Layout::NHWC};
+
+  // subgraph index of this subgraph in a model
+  SubgraphIndex _subg_index;
 };
 
 } // namespace ir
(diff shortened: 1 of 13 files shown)


runtime/onert/frontend/nnapi/wrapper/ANeuralNetworksModel.cc
```diff
@@ -29,7 +29,8 @@
 ANeuralNetworksModel::ANeuralNetworksModel() noexcept
     : _optional_operands{}, _operand_usages{}, _allowFloat32toFloat16{false}
 {
-  _graph = std::make_shared<onert::ir::Graph>();
+  uint32_t primary_subgraph_index = 0;
+  _graph = std::make_shared<onert::ir::Graph>(primary_subgraph_index);
```
wateret, line 33: In NN API, `ANeuralNetworksModel` is a subgraph. So this would mean all subgraphs made with NN API are going to have index 0. It looks like a dummy value.

runtime/onert/core/include/ir/Graph.h
```diff
@@ -103,6 +109,11 @@ class Graph
   const std::shared_ptr<Subgraphs> &subgraphs() const { return _subgraphs; }
   std::shared_ptr<Subgraphs> &subgraphs() { return _subgraphs; }
   Layout layout() const { return _layout; }
+  SubgraphIndex subg_index() const
```
wateret, line 112: What about just `index()`?

wateret (commented):
> The proper place to keep the subgraph index seems to be ir::Graph.

I agree that it is the most convenient to implement, but I'm not sure if it is the most proper.

- Profiling-only data should not reside in the class itself
- subgraph index is Circle(tflite) specific. NN API does not have such a thing
- `ir::Operand` and `ir::Operation` do not have their indices and it is intended

% FYI Once #5010 is merged, `is_primary_subgraph` flag related stuff is going to be removed as well.

> (If we don't want to keep the index in Graph, we need a map of subgraph index and Graph object, which is created at loading time and passed all the way to execution time, which seems complicated.)

If it complicates the code too much, then we should do it in this way. But I thought it wouldn't as I think that info can be `ChromeTracingObserver`. Please correct me if I'm wrong.

hyunsik-yoon:
For https://github.com/Samsung/ONE/pull/5057#pullrequestreview-531971030,

I am trying to write down requirements:

> subgraph index is Circle(tflite) specific. NN API does not have such a thing

- For Circle(tflite), profiling tool needs to show subgraph index.
- For NN API, there is no such ID. However, each model (subgraph) needs to be distinguished with unique ID, anyways.

> Profiling-only data should not reside in the class itself

- Then we might need some kind of map (ir::Graph to index) somewhere (not in ir::Graph)

I will think about how to achieve this.

hyunsik-yoon:
Instead of putting index into `ir::Graph`, I will create a class where I put graph and index map in, e.g., `util/ProfilingInfo.h`.

I will close this PR after making a pr for the new class.

Issue #4993, opened 2020-11-11
[luci] Introduce ShuffleWeightTo16x1Float32Pass


mhs4670go:
As #4804, col-major 16x1 tiling for fullyconnected gets performance gain
For this, we need to implement it into `circle2circle` pass.

Let's do this!

PR #5068, opened 2020-11-17
[luci] Revise import and export of FullyConnected op

mhs4670go:
This commit revises import and export of FullyConnected op as
WeightFormat has been introduced.

Related : #4993
Draft: #5058

### DIFF
--- a/compiler/luci/import/src/CircleReader.cpp
+++ b/compiler/luci/import/src/CircleReader.cpp
@@ -151,6 +151,22 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode)
   return MirrorPadMode::UNDEFINED;
 }
 
+luci::CircleFullyConnected::WeightsFormat
+luci_weights_format(const circle::FullyConnectedOptionsWeightsFormat weights_format)
+{
+  switch (weights_format)
+  {
+    case circle::FullyConnectedOptionsWeightsFormat_DEFAULT:
+      return luci::CircleFullyConnected::WeightsFormat::DEFAULT;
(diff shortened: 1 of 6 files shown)


compiler/luci/import/src/CircleReader.cpp
```diff
@@ -151,6 +151,22 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode)
...
+  {
+    case circle::FullyConnectedOptionsWeightsFormat_DEFAULT:
+      return luci::CircleFullyConnected::WeightsFormat::DEFAULT;
+    case circle::FullyConnectedOptionsWeightsFormat_SHUFFLED4x16INT8:
+      return luci::CircleFullyConnected::WeightsFormat::SHUFFLED4x16INT8;
+    case circle::FullyConnectedOptionsWeightsFormat_SHUFFLED16x1FLOAT32:
+      return luci::CircleFullyConnected::WeightsFormat::SHUFFLED16x1FLOAT32;
+  }
+  assert(false);
+  return luci::CircleFullyConnected::WeightsFormat::UNDEFINED;
```
seanshpark, line 167: By comment in IR, `UNDEFINED` is not permitted. How about throw an exception?

seanshpark (approved):
LGTM thank you!

compiler/luci/import/src/CircleReader.cpp
seanshpark: From offline disussion;
- this was copied from exsiting other (optional) attributes
- this line itself is to check `set` argument
- it would be safe to throw for release build also to prevent invalid model file
- maybe sometime later we can revise existing codes

PR #5072, opened 2020-11-17
[common-artifacts] Add ReLUN1TO1 to exlucde list

llFreetimell:
Parent Issue : #5037

This commit will add `ReLUN1TO1` to `exclude.lst`


seanshpark (approved):
LGTM

PR #4962, opened 2020-11-09
[Draft][luci] Introduce inter-module optimization

llFreetimell:
Parent Issue : #4961

This draft is for introducing inter-module optimization.
Final goal is applying `FuseBCQPass` to multiple subgraphs.

Issue #4961, opened 2020-11-09
[luci] Support module optimization such as applying BCQ


llFreetimell:
### Description

Current optimizations in `luci/Pass` are only applicable to `loco::graph`, not `luci::Module`.
However, if some model has multiple subgraphs and we want to apply BCQ to the model, BCQ information should be shared by each subgraph.
Therefore, let's support `module` optimization in `luci`!

### Notes

- As design of `FuseBCQPass` is changed, core logic is little bit changed.

### Todo
- [x] Introduce `luci::Pass` #5081 #5090 #5106
- [x] Introduce `luci::PhaseRunner` #5088
- [x] Introduce `ModuleProgressReporter` #5082
- [x] Introduce optimizer for `luci::Module` in `CircleOptimizer` #5114
- [x] Enable `luci::Module` optimization in `circle2circle` #5123
- [x] Update `FuseBCQPass`
    - [x] Update `generate_bcq_output_arrays` #5089
    - [x] Update algorithm of `FuseBCQPass` #5092 #5117 #5128
    - [x] Introduce `FuseBCQPass::run(Module)` #5141
- [x] Update `CircleOptimizer` to use `luci/Service` using `luci::Module` #5129
- [x] Update `CircleOptimizer` to use `ModulePass` instead of `logo::Pass` for some passes #5166

### Draft
- #4962

llFreetimell:
Done!

PR #5081, opened 2020-11-19
[luci] Introduce LuciPass

llFreetimell:
Parent Issue : #4961
Draft code : #4962

Until now, we used `logo::Pass` in optimization passes.
However, it only supports for optimizing `loco::Graph`, not for `luci::Module`.
Furthermore, many related infra is only supporting `loco::Graph`.
`luci::Pass` will enable supporting optimization and infra for `luci::Module`.

### DIFF
--- a/compiler/luci/pass/include/luci/LuciPass.h
+++ b/compiler/luci/pass/include/luci/LuciPass.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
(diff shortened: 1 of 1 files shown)


seanshpark:
Could you please explain why we need luci::Pass? For I could see from this PR is that there is no "why"

llFreetimell:
@seanshpark I missed to write the reason not only in the issue but also in PR :( I will update commit message...!

llFreetimell:
@seanshpark PTAL :)

seanshpark (approved):
LGTM thank you!

PR #5082, opened 2020-11-19
[luci] Introduce progress reporter for Module

llFreetimell:
Parent Issue : #4961
Draft Code : #4962

Current `ProgressReporter` only supports for `loco::Graph`.
However, after `luci::Pass` is introduced, `ProgressReporter` for `luci::Module` is needed
because `luci::Pass` supports optimizations for `luci::Module`.
This commit will introduce `ModuleProgressReporter`, which supports reporting about module progress.


seanshpark:
If you could add a reason why we need this new class, would help the review.

seanshpark (approved):
LGTM thank you!

PR #5091, opened 2020-11-20
[tflitefile_tool] model_parser will use current environment's python

glistening:
It fixes errors like numpy is not found in python venv environment,
when the module is installed in venv only.


hyunsik-yoon (approved):
LGTM

hseok-oh (approved):
LGTM

PR #5092, opened 2020-11-20
[luci] Introduce new functions in FuseBCQPass

llFreetimell:
Parent Issue : #4961 

This commit introduce two new functions to support inter-module BCQ fusing.

- Getter for `alpha` and `packed_binary_code`
  - When `FuseBCQPass` supports fusing between `luci::Module`, same `luci::Const` cannot be used in different subgraph
  - Copying and creating new `luci::Const` will resolve this
- BCQ information Register Function
  - Same information can be shared by different subgraphs.

### DIFF
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -38,11 +38,9 @@ public:
   }
 
 public:
-  bool fuseBCQ(loco::Graph *g)
+  void register_bcq_info(loco::Graph *g)
   {
-
-    const auto output_nodes = loco::output_nodes(g);
-    for (auto node : output_nodes)
+    for (auto node : loco::output_nodes(g))
     {
       auto output_node = loco::must_cast<luci::CircleOutput *>(node);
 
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -435,6 +468,8 @@ bool FuseBCQPass::run(loco::Graph *g)
   const int32_t start_magicnum = -2e9 + 27;
   const int32_t end_magicnum = 2e9 - 27;
 
+  loco::Graph *main_graph = g;
```
seanshpark, line 471: Q) why is `main_graph` introduced? `g` is also used below.

llFreetimell: Another PRs will be created to substitute `run(loco::Graph *g)` to `run(luci::Module *m)`.
In those situation, multiple subgraphs will be appeared and `main_graph` is the main graph which includes BCQ information :)

seanshpark (approved):
LGTM

PR #5021, opened 2020-11-12
[onert] Introduce xnnpack backend

YongseopKim:
Introduce xnnpack backend

ONE-DCO-1.0-Signed-off-by: Yongseop Kim <yons.kim@samsung.com>

PR #5095, opened 2020-11-20
[nnfwapi] Support RUY backend testing in GenModelTest

periannath:
- This commit supports RUY backend testing in GenModelTest

### DIFF
--- a/tests/nnfw_api/src/GenModelTest.h
+++ b/tests/nnfw_api/src/GenModelTest.h
@@ -224,7 +224,7 @@ public:
         _backends.push_back(backend);
       }
 #endif
-      if (backend == "cpu")
+      if (backend == "cpu" || backend == "ruy")
       {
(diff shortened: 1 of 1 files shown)


tests/nnfw_api/src/GenModelTest.h
```diff
@@ -228,6 +228,10 @@ class GenModelTestContext
       {
         _backends.push_back(backend);
       }
+      else if (backend == "ruy")
```
YongseopKim, line 231: `if (backend == "cpu" || if (backend == "ruy"))` looks better. I do like this on #5021

periannath: @YongseopKim Thanks! It looks better. I've updated.

wateret (approved):
LGTM

YongseopKim (approved):
LGTM

PR #5097, opened 2020-11-20
[docs] Write How to Use Specific Backend during Inference

periannath:
- Write "How to Use Specific Backend during Inference" document


wateret (approved):
LGTM

PR #5101, opened 2020-11-20
[onert] Introduce a map of [ir::Graph*, ir::SubgraphIndex]

hyunsik-yoon:
This introduces a map of [`ir::Graph*`, `ir::SubgraphIndex`].
Main purpose of this map is to show subgraph index to user with profiling info.

Parent issue: #4901
Draft: #4903
refer to: https://github.com/Samsung/ONE/pull/5057

### DIFF
--- a/runtime/onert/core/include/util/GraphIndexMap.h
+++ b/runtime/onert/core/include/util/GraphIndexMap.h
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __ONERT_UTIL_GRAPH_INDEX_MAP_H__
+#define __ONERT_UTIL_GRAPH_INDEX_MAP_H__
+
+#include "ir/Graph.h"
+#include "ir/Index.h"
+
+#include <unordered_map>
+
+namespace onert
+{
+namespace util
+{
+
+/**
+ * @brief Class to maintain ir::Graph* and its subgraph index
+ *        Subgraph index comes normally from circle or tflite file.
+ *        When there is no concept of subgraph index like nnapi API, getIndex() always returns 0.
+ */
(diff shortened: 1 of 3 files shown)


runtime/onert/core/include/util/GraphIndexMap.h
```diff
@@ -0,0 +1,83 @@
...
+ *        Subgraph index comes normally from circle or tflite file.
+ *        When there is no concept of subgraph index like nnapi API, getIndex() always returns 0.
+ */
+class GraphIndexMap
+{
+public:
+  GraphIndexMap() = default;
+
+public:
+  static GraphIndexMap &get()
```
wateret, line 41: `GraphIndexMap` should not be a singleton. There may be many sessions at the same time and each session would have duplicated indices.

runtime/onert/core/include/util/GraphIndexMap.h
```diff
@@ -0,0 +1,83 @@
...
+
+    return find == _map.end() ? ir::SubgraphIndex{DEFAULT_SUBG_INDEX} : find->second;
+  }
+
+  /**
+   * @brief Copies [graph, subgraph index] to [graph', subgraph index]
+   *        This is useful when source graph is cloned into another graph. Both graphs will have
+   *        same subgraph index.
+   */
+  void copyIndex(const ir::Graph *src, const ir::Graph *dst)
```
wateret, line 70: (Opinion-based) This is a simple combination of `setIndex` and `getIndex` so I would rather not introduce this function. Just leave it if you prefer this way. 😃

wateret (changes requested):
Please make it not be the singleton.

runtime/onert/core/include/util/GraphIndexMap.h
hyunsik-yoon: @wateret ,
Talking about multiple sessions, I haven't considered that.

Current draft work like the following:

1. put info into `GraphIndexMap` (this PR, singleton)
2. executors will enqueue profiling events
3. flush all profiling events (singleton)

If we consider multiple sessions, the sequence will be

1. info about session, graph and operation
  - 1-1) put these info _per session_, and pass this object to compiler, and then executor 
2. executors will enqueue profiling events
3. flush all profiling events at a right time (_try not to use singleton_)
  - A file per session

I guess you don't prefer singleton due to some architectural issue with singleton.

Please comment about the above sequence. I will fix the draft again after your comment.

wateret: > I guess you don't prefer singleton due to some architectural issue with singleton.

I think it would cause crashes/undefined behavior. What if two sessions exist at the same time? They would have same Subgraph indices. So IMHO we must not do that way.

However as this affects only when profiling(otherwise it is really serious, we have similar issues suffering now reported by nnstreamer tests), it would be OK to have it as a singleton then switch to the second one. If you want work in this order, please add some comments about it. Then I can approve this one :)

hyunsik-yoon: @wateret I will try not to use singleton. :-)

PR #5110, opened 2020-11-23
[Draft] Introduce backend scheduler scripts

periannath:
- This commit introduces backend scheduler scripts
  - Measure execution time of each `Conv2D` layer with TracingObserver
  - Select fastest backend among candidate backends
  - Produce environment variable for fast inference time


periannath:
#5597 is merged.

PR #5116, opened 2020-11-23
[onert] OperationValidator for ArgMax

hseok-oh:
- Enable OperationValidator for ArgMax operation
- Add negative test for ArgMax
- Enable ArgMax test for int64 output

### DIFF
--- a/runtime/onert/core/src/ir/OperationValidator.cc
+++ b/runtime/onert/core/src/ir/OperationValidator.cc
@@ -87,11 +87,13 @@ void OperationValidator::visit(const operation::AddN &node)
 void OperationValidator::visit(const operation::ArgMax &node)
 {
   const auto input_index(node.getInputs().at(operation::ArgMax::Input::INPUT));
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/ir/OperationValidator.cc
```diff
@@ -84,6 +84,18 @@ void OperationValidator::visit(const operation::AddN &node)
   }
 }
 
+void OperationValidator::visit(const operation::ArgMax &node)
+{
+  const auto input_index(node.getInputs().at(operation::ArgMax::Input::INPUT));
```
ragmani, line 89: How about adding checking axis type too?

hseok-oh: I've updated to check axis type. PTAL

hseok-oh:
@ragmani @glistening PTAL

ragmani (approved):
LGTM

glistening (approved):
LGTM

PR #5117, opened 2020-11-23
[luci] Change a method of getting prefix in FuseBCQPass

llFreetimell:
Parent Issue : #4961

Until now, as each BCQ applicable candidate nodes are paired with their BCQ information nodes,
prefix and candidate nodes are paired as `std::map`.
However, because BCQ applicable candidate nodes are no longer paired and BCQ applicated weights are paired instead,
method of getting prefix is changed by comparing each weights values.
This commit will change the method.

### DIFF
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -83,7 +83,7 @@ bool is_fusable_const(luci::CircleConst *before, luci::CircleConst *after, bool 
       // Check if [dim0, dim1, dim2] --> [dim2, dim0 * dim1] or
       //          [dim0, dim1, dim2] --> [dim1 * dim2, dim0]
       if ((after_dim0 == before_dim1 * before_dim2 && after_dim1 == before_dim0) ||
-          (after_dim0 == e && after_dim1 == before_dim0 * before_dim1))
+          (after_dim0 == before_dim2 && after_dim1 == before_dim0 * before_dim1))
       {
         for (uint32_t i = 0; i < after_dim0; ++i)
           for (uint32_t j = 0; j < after_dim1; ++j)
@@ -482,7 +482,7 @@ private:
       return false;
     }
 
-    if (_fusable_op.find(prefix) == _qbits_of_clusters.end())
+    if (_fusable_op.find(prefix) == _fusable_op.end())
     {
       WARN(l) << "fusable_op is not found" << std::endl;
       return false;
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -25,6 +25,85 @@
...
 
+bool is_fusable_const(luci::CircleConst *before, luci::CircleConst *after, bool do_w_x)
+{
+  if (after->dtype() != loco::DataType::FLOAT32)
+    return false;
+
+  if (after->rank() != 2)
+    return false;
+
+  if (after->size<loco::DataType::FLOAT32>() != before->size<loco::DataType::FLOAT32>())
```
mhs4670go, line 36: `before` should be checked if its `dtype` is `FLOAT32` as `after` does.

compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -25,6 +25,85 @@
...
+  auto after_dim0 = after->dim(0).value();
+  auto after_dim1 = after->dim(1).value();
+
+  if (before->rank() == 2)
+  {
+    if (do_w_x)
+    {
+      // Check for [dim0, dim1] --> [dim0, dim1]
+      if (!(after->dim(0) == before->dim(0) && after->dim(1) == before->dim(1)))
+        return false;
```
mhs4670go, line 48: This condition can be placed before this `if` block.

compiler/luci/pass/src/FuseBCQPass.cpp
```diff
@@ -25,6 +25,85 @@
...
+
+      for (uint32_t i = 0; i < after->size<loco::DataType::FLOAT32>(); ++i)
+        if (after->at<loco::DataType::FLOAT32>(i) != before->at<loco::DataType::FLOAT32>(i))
+          return false;
+    }
+    else
+    {
+      // Check for [dim0, dim1] --> [dim1, dim0]
+      if (!(after->dim(0) == before->dim(1) && after->dim(1) == before->dim(0)))
+        return false;
```
mhs4670go, line 58: with this.

compiler/luci/pass/src/FuseBCQPass.cpp
llFreetimell: This check is needed but not here, I will add this validation :)

compiler/luci/pass/src/FuseBCQPass.cpp
llFreetimell: Two condition is different...!

mhs4670go (dismissed):
LGTM

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #5120, opened 2020-11-23
[luci] Pass to propagate quantization parameter

jinevening:
This adds a pass to propagate quantparam of Op from output to input

- Currently only RESHAPE is supported.


compiler/luci/pass/src/PropagateQuantParamPass.cpp
```diff
@@ -0,0 +1,95 @@
...
+  auto input_node = loco::must_cast<luci::CircleNode *>(input);
+  return copy_qparam(node, input_node);
+}
+
+bool PropagateQuantParamPass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    LOGGER(l);
```
mhs4670go, line 78: (trivial) maybe this `LOGGER` macro can move to out of this for block?

compiler/luci/pass/src/PropagateQuantParamPass.cpp
```diff
@@ -0,0 +1,95 @@
...
+      src->quantized_dimension == dst->quantized_dimension)
+    return false;
+
+  dst->scale.assign(src->scale.begin(), src->scale.end());
+  dst->zerop.assign(src->zerop.begin(), src->zerop.end());
+  dst->quantized_dimension = src->quantized_dimension;
+  return true;
+}
+
+bool copy_qparam(luci::CircleNode *src, luci::CircleNode *dst)
```
mhs4670go, line 45: Just curiosity, there is no care for the node that hasn't got quantparam, i.e. newly generated node or sth?

compiler/luci/pass/src/PropagateQuantParamPass.cpp
jinevening: Thanks I'll update it.

compiler/luci/pass/src/PropagateQuantParamPass.cpp
```diff
@@ -0,0 +1,95 @@
...
+    LOGGER(l);
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+    auto opcode = circle_node->opcode();
+    if (opcode != luci::CircleOpcode::RESHAPE)
+      continue;
+
+    INFO(l) << "PropagateQuantParamPass visit node: " << circle_node->name() << std::endl;
+
+    PropagateQuantParam pqp;
+    changed = circle_node->accept(&pqp);
```
mhs4670go, line 87: How about having `accept` method returning `false` by default except the op which has its implementation?

compiler/luci/pass/src/PropagateQuantParamPass.cpp
jinevening: This pass is only for quantized nodes, so I made it skip nodes that do not have quantparams.

```
  auto src_qparam = src->quantparam();
  if (not src_qparam)
    return false;

  auto dst_qparam = dst->quantparam();
  if (not dst_qparam)
    return false;
```

mhs4670go: Oh, I got it. Maybe simple comment about this would be help.

compiler/luci/pass/src/PropagateQuantParamPass.cpp
jinevening: Updated. Thanks.

mhs4670go (approved):
LGTM

compiler/luci/pass/src/PropagateQuantParamPass.test.cpp
```diff
@@ -0,0 +1,116 @@
...
+  EXPECT_EQ(10, g.conv->quantparam()->zerop[2]);
+}
+
+TEST(PropagateQuantParam, wrong_op_neg)
+{
+  SimpleGraph g;
+
+  luci::PropagateQuantParam pqp;
+  auto circle_node = loco::must_cast<luci::CircleNode *>(g.conv);
+  EXPECT_ANY_THROW(circle_node->accept(&pqp));
```
mhs4670go, line 100: `EXPECT_FALSE`:)

jinevening: Oh, Thanks 👍

compiler/luci/pass/src/PropagateQuantParamPassInternal.h
```diff
@@ -0,0 +1,38 @@
...
+#define __LUCI_CIRCLE_PROPAGATE_QUANT_PARAM_PASS_INTERNAL_H__
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleNodeVisitor.h>
+
+namespace luci
+{
+
+//  Visitor to propagate quantization parameters
+struct PropagateQuantParam final : public luci::CircleNodeMutableVisitor<bool>
```
seanshpark, line 27: As I understand, this `PropagateQuantParam` class purpose is to be used (only) in `PropagateQuantParamPass` class
but is exposed with `luci` namespace by this internal header just for the testing.
If so, it would be better not to expose by this header and do testing with `PropagateQuantParamPass` class not with `PropagateQuantParam` class. 

Reason: Exposing this class to `luci` namespace with a header may let others reuse this. There is no information not to. But I don't think this was your intention and purpose of this class. I'm not sure adding a something like a WARNING comment will do...

If you have difficulties chaning the test, please let me know... we may have to find another way...
- (I've recently noticed there already exist two or more Passes working in this way. It would be better to fix them too, but when you have some spare time, with low priority.)

jinevening: > If so, it would be better not to expose by this header and do testing with PropagateQuantParamPass class not with PropagateQuantParam class.

I found this pattern (`XXXInternal.h`) from our legacy code (FuseInstanceNormPass). So I thought the pattern is acceptable. I've changed the test pattern. PTAL

compiler/luci/pass/src/PropagateQuantParamPass.cpp
```diff
@@ -0,0 +1,102 @@
...
+  if (not dst_qparam)
+    return false;
+
+  return copy_qparam(src_qparam, dst_qparam);
+}
+
+//  Visitor to propagate quantization parameters
+struct PropagateQuantParam final : public luci::CircleNodeMutableVisitor<bool>
+{
+  PropagateQuantParam() {}
```
seanshpark, line 62: Can we remove this line? or use like 
```cxx
PropagateQuantParam() = default;
```
?

jinevening, line 62:
```suggestion
  PropagateQuantParam() = default;
```

jinevening: I've updated the code as mentioned.

seanshpark (approved):
LGTM thank you

mhs4670go (approved):
LGTM

PR #5123, opened 2020-11-23
[circle2circle] Enable module optimization

llFreetimell:
Parent Issue : #4961 

Until now, only graph optimizations were possible.
This commit will enable module optimization in `circle2circle`.

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -290,11 +290,14 @@ int entry(int argc, char **argv)
   luci::Importer importer;
   auto module = importer.importModule(circle_model);
 
+  // call luci optimizations for module
+  optimizer.optimize(module.get());
+
   for (size_t idx = 0; idx < module->size(); ++idx)
   {
     auto graph = module->graph(idx);
 
-    // call luci optimizations
+    // call luci optimizations for graph
     optimizer.optimize(graph);
     optimizer.sparsify(graph);
 


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -290,11 +290,14 @@ int entry(int argc, char **argv)
...
+  // call luci optimizations for module
+  optimizer.optimize(module.get());
+
   for (size_t idx = 0; idx < module->size(); ++idx)
   {
     auto graph = module->graph(idx);
 
-    // call luci optimizations
+    // call luci optimizations for graph
     optimizer.optimize(graph);
```
mhs4670go, line 301: This could be removed. It being here is just for safety?

llFreetimell: `optimize(graph)` and `optimize(module)` is different and independent :)

For example, `FuseBCQPass` has some optimization for `module` but do not have for `graph`.
On the other hand, most of other passes has optimizations for `graph` but not for `module`.
Even worse, basic services such as `ShapeInferencePass` should be implemented for both of them.

mhs4670go: Oh, okay. I thought its functionality includes existing `optimize`'s one expecting `graph` as a parameter.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #5074, opened 2020-11-17
[Draft][luci-pass]Add remove Duplicated Transpose

struss:
For #5062 

This commit add remove duplicated tranpose.

ONE-DCO-1.0-Signed-off-by: KiDeuk Bang <rrstrous@nate.com>

Issue #5062, opened 2020-11-16
[Compiler][luci] Introduce new optimization pass, RemoveDuplicateTranspose


struss:
Parent Issue : #4353
***
### What to do in this Pass
Remove Duplicate `Transpose` node on graph.
#### Input Type 1
Duplicate `Transpose` with single input & output
![93728238-fee3ec80-fbf9-11ea-9f1b-98ee8a25706c](https://user-images.githubusercontent.com/2858751/99244107-f77d4000-2844-11eb-93a6-5c52660a5e93.png)
-> In this Case, check permutations of two node. if combination of perm is [0,1,2,3....,n] then remove both.
otherwise, make a single `Transpose` node with combination of perm values.(OPT)
#### Input Type 2(OPT & Question)
Multi Output path on first `Transpose`
-> In this Case, remove second node and create a new `Transpose` node with combination of perm values.
### TODO
- [x] Introduce luci/Pass for above optimization. #5124 
- [x] Introduce this optimization on circle2circle. #5172
### Draft
#5074

seanshpark:
This will sovle for some Ops (for example, pytorch DepthwiseConv2D) conversion where input is NCHW and Op is NHWC

<img src='https://user-images.githubusercontent.com/4616940/99316345-9044a780-28a7-11eb-8df0-f4a9e23566c8.png' width=200px>

struss:
Done! If anything to be done left, please reopen and mention me! thanks for all!

PR #5124, opened 2020-11-23
[luci/pass]Introduce pass/RemoveRedundantTranspose on luci/pass.

struss:
Parent Issue : #5062 
Draft : #5074 
#
This commit introduce pass/RemoveRedundantTranspose on luci/pass.

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/RemoveRedundantTransposePass.h
+++ b/compiler/luci/pass/include/luci/Pass/RemoveRedundantTransposePass.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_REMOVE_REDUNDANT_TRANSPOSE_H__
+#define __LUCI_REMOVE_REDUNDANT_TRANSPOSE_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief fuse or remove subsequent Transpose operators
+ */
+struct RemoveRedundantTransposePass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::REMOVE_REDUNDANT_TRANSPOSE"; }
+
+  bool run(loco::Graph *g) final;
+};
+
+} // namespace luci
+
+#endif // __LUCI_REMOVE_REDUNDANT_TRANSPOSE_H__
--- a/compiler/luci/pass/include/luci/Pass/RemoveRedundantTransposePass.h
+++ b/compiler/luci/pass/include/luci/Pass/RemoveRedundantTransposePass.h
@@ -27,7 +27,7 @@ namespace luci
  */
 struct RemoveRedundantTransposePass final : public logo::Pass
 {
-  const char *name(void) const final { return "luci::REMOVE_REDUNDANT_TRANSPOSE"; }
+  const char *name(void) const final { return "luci::RemoveRedundantTransposePass"; }
 
   bool run(loco::Graph *g) final;
 };
--- a/compiler/luci/pass/src/RemoveRedundantTranspose.cpp
+++ b/compiler/luci/pass/src/RemoveRedundantTranspose.cpp
@@ -1,0 +1,125 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/RemoveRedundantTransposePass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleOpcode.h>
+
+namespace luci
+{
+
+/// @brief Return true if first_prem[second_prem[i]] == i
+bool check_perm(const luci::CircleConst *first_perm, const luci::CircleConst *second_perm)
+{
+  assert(first_perm->rank() == 1);
+  assert(second_perm->rank() == 1);
+  assert(second_perm->size<loco::DataType::S32>() == first_perm->size<loco::DataType::S32>());
+  for (int32_t i = 0; i < static_cast<int32_t>(first_perm->size<loco::DataType::S32>()); i++)
+  {
+    if (first_perm->at<loco::DataType::S32>(second_perm->at<loco::DataType::S32>(i)) != i)
+      return false;
+  }
+  return true;
+}
+
+bool remove_consecutive_transpose_function(luci::CircleNode *node)
+{
+  if (node->opcode() != luci::CircleOpcode::TRANSPOSE)
+    return false;
+
+  auto pred_node = static_cast<luci::CircleNode *>(node->arg(0));
+  if (pred_node->opcode() != luci::CircleOpcode::TRANSPOSE)
+    return false;
+  if (loco::succs(pred_node).size() != 1)
+    return false;
+
+  auto pred_perm = dynamic_cast<luci::CircleConst *>(node->arg(1));
+  if (pred_perm == nullptr)
+    return false;
+
+  auto main_perm = dynamic_cast<luci::CircleConst *>(pred_node->arg(1));
+  if (main_perm == nullptr)
+    return false;
+
+  auto main_node = static_cast<luci::CircleNode *>(pred_node->arg(0));
+  if (check_perm(pred_perm, main_perm))
+  {
+    replace(node).with(main_node);
+    pred_node->drop();
+  }
+  else
+  {
+    auto g = main_perm->graph();
+    auto new_const_node = g->nodes()->create<luci::CircleConst>();
+
+    new_const_node->dtype(loco::DataType::S32);
+    new_const_node->rank(main_perm->rank());
+    uint32_t dim_size = 1;
+    for (uint32_t i = 0; i < new_const_node->rank(); ++i)
+    {
+      new_const_node->dim(i) = main_perm->dim(i);
+      dim_size *= main_perm->dim(i).value();
+    }
+    new_const_node->size<loco::DataType::S32>(dim_size);
+    new_const_node->shape_status(luci::ShapeStatus::VALID);
+    for (uint32_t i = 0; i < main_perm->size<loco::DataType::S32>(); i++)
+    {
+      new_const_node->at<loco::DataType::S32>(i) =
+          pred_perm->at<loco::DataType::S32>(main_perm->at<loco::DataType::S32>(i));
+    }
+    replace(main_perm).with(new_const_node);
+    replace(node).with(pred_node);
+  }
+  node->drop();
+  return true;
+}
+
+/**
+ *  BEFORE
+ *         |
+ *   [CircleInput]    [CircleConst]
+ *           \              /
+ *           [CircleTranspose]  [CircleConst]
+ *                   \               /
+ *                   [CircleTranspose]
+ *                           |
+ *
+ *  AFTER
+ *      <Optional Case>
+ *
+ *          |                 |               |
+ *    [CircleInput]     [CircleConst]         |
+ *           \               /           or   |  [Remove all]
+ *           [CircleTranspose]                |
+ *                   |                        |
+ *
+ */
+bool RemoveRedundantTransposePass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto circle_node = static_cast<luci::CircleNode *>(node);
+    if (remove_consecutive_transpose_function(circle_node))
+    {
+      changed = true;
+      break;
+    }
+  }
+  return changed;
+}
+
+} // namespace luci
--- a/compiler/luci/pass/src/RemoveRedundantTranspose.cpp
+++ b/compiler/luci/pass/src/RemoveRedundantTranspose.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -16,12 +17,11 @@
 #include "luci/Pass/RemoveRedundantTransposePass.h"
 
 #include <luci/IR/CircleNodes.h>
-#include <luci/IR/CircleOpcode.h>
 
-namespace luci
+namespace
 {
 
-/// @brief Return true if first_prem[second_prem[i]] == i
+/// @brief Return true if first_perm[second_perm[i]] == i
 bool check_perm(const luci::CircleConst *first_perm, const luci::CircleConst *second_perm)
 {
   assert(first_perm->rank() == 1);
@@ -37,24 +37,24 @@ bool check_perm(const luci::CircleConst *first_perm, const luci::CircleConst *se
 
 bool remove_consecutive_transpose_function(luci::CircleNode *node)
 {
-  if (node->opcode() != luci::CircleOpcode::TRANSPOSE)
+  auto target_node = dynamic_cast<luci::CircleTranspose *>(node);
+  if (target_node == nullptr)
     return false;
-
-  auto pred_node = static_cast<luci::CircleNode *>(node->arg(0));
-  if (pred_node->opcode() != luci::CircleOpcode::TRANSPOSE)
+  auto pred_node = dynamic_cast<luci::CircleTranspose *>(target_node->a());
+  if (pred_node == nullptr)
     return false;
   if (loco::succs(pred_node).size() != 1)
     return false;
 
-  auto pred_perm = dynamic_cast<luci::CircleConst *>(node->arg(1));
+  auto pred_perm = dynamic_cast<luci::CircleConst *>(target_node->perm());
   if (pred_perm == nullptr)
     return false;
 
-  auto main_perm = dynamic_cast<luci::CircleConst *>(pred_node->arg(1));
+  auto main_perm = dynamic_cast<luci::CircleConst *>(pred_node->perm());
   if (main_perm == nullptr)
     return false;
 
-  auto main_node = static_cast<luci::CircleNode *>(pred_node->arg(0));
+  auto main_node = loco::must_cast<luci::CircleNode *>(pred_node->a());
   if (check_perm(pred_perm, main_perm))
   {
     replace(node).with(main_node);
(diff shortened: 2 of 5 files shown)


compiler/luci/pass/include/luci/Pass/RemoveRedundantTransposePass.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace luci
+{
+
+/**
+ * @brief fuse or remove subsequent Transpose operators
+ */
+struct RemoveRedundantTransposePass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::REMOVE_REDUNDANT_TRANSPOSE"; }
```
jinevening, line 30: For consistency,

```suggestion
  const char *name(void) const final { return "luci::RemoveRedundantTransposePass"; }
```

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -0,0 +1,152 @@
...
+ */
+#include "luci/Pass/RemoveRedundantTransposePass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+#include <vector>
+
+#include <gtest/gtest.h>
+
+void setValue(luci::CircleConst *node, const std::vector<int> &v)
```
jinevening, line 24: Please wrap `setValue` and `create_redundant_transpose` with anonymous namespace.

If you don't, their implementations can go outside of this file.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include "luci/Pass/RemoveRedundantTransposePass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleOpcode.h>
+
+namespace luci
+{
+
+/// @brief Return true if first_prem[second_prem[i]] == i
+bool check_perm(const luci::CircleConst *first_perm, const luci::CircleConst *second_perm)
```
jinevening, line 25: If `check_perm` and `remove_consecutive_transpose_function` are used only in this file, it would be better to wrap them with anonymous namespace (not luci).

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+#include "luci/Pass/RemoveRedundantTransposePass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleOpcode.h>
+
+namespace luci
+{
+
+/// @brief Return true if first_prem[second_prem[i]] == i
```
mhs4670go, line 24:
```suggestion
/// @brief Return true if first_perm[second_perm[i]] == i
```

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
...
+bool check_perm(const luci::CircleConst *first_perm, const luci::CircleConst *second_perm)
+{
+  assert(first_perm->rank() == 1);
+  assert(second_perm->rank() == 1);
+  assert(second_perm->size<loco::DataType::S32>() == first_perm->size<loco::DataType::S32>());
+  for (int32_t i = 0; i < static_cast<int32_t>(first_perm->size<loco::DataType::S32>()); i++)
```
mhs4670go, line 30: This `static_cast` seems redundant. It can be just `uint32_t`.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
...
+ *           [CircleTranspose]                |
+ *                   |                        |
+ *
+ */
+bool RemoveRedundantTransposePass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto circle_node = static_cast<luci::CircleNode *>(node);
```
mhs4670go, line 115: It would be more safe.
```suggestion
    auto circle_node = dynamic_cast<luci::CircleNode *>(node);
```

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
...
+    replace(node).with(pred_node);
+  }
+  node->drop();
+  return true;
+}
+
+/**
+ *  BEFORE
+ *         |
+ *   [CircleInput]    [CircleConst]
```
mhs4670go, line 93: Isn't it more accurate?
```suggestion
 *   [CircleNode]    [CircleConst]
```

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
...
+ *           [CircleTranspose]  [CircleConst]
+ *                   \               /
+ *                   [CircleTranspose]
+ *                           |
+ *
+ *  AFTER
+ *      <Optional Case>
+ *
+ *          |                 |               |
+ *    [CircleInput]     [CircleConst]         |
```
mhs4670go, line 104: ditto

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: If I remove that, the type of `first_perm->size<loco::DataType::S32>()` is `uint32_t`. then comparison of `int32_t` and `uint32_t` gives a error on it.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: Oh I got it. I will change this.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: same as above

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: I will fix it.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: Oh, I will fix it.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: Oh, I will fix it.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
llFreetimell, line 3:
```suggestion

 * Licensed under the Apache License, Version 2.0 (the "License");
```

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
llFreetimell, line 30: Most of index and size in `CircleNode` are `uint32_t` or `size_t`.
So following modification looks better :)

```suggestion
  for (uint32_t i = 0; i < first_perm->size<loco::DataType::S32>(); i++)
```

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
llFreetimell: If `circle_node` is `nullptr`, segmentation fault occurs.
When casting `loco::Node` to `luci::CircleNode`, `loco::must_cast` is used normally.
```suggestion
    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
```

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
...
+  assert(second_perm->size<loco::DataType::S32>() == first_perm->size<loco::DataType::S32>());
+  for (int32_t i = 0; i < static_cast<int32_t>(first_perm->size<loco::DataType::S32>()); i++)
+  {
+    if (first_perm->at<loco::DataType::S32>(second_perm->at<loco::DataType::S32>(i)) != i)
+      return false;
+  }
+  return true;
+}
+
+bool remove_consecutive_transpose_function(luci::CircleNode *node)
```
llFreetimell, line 38: Implementation of this function have several problems for me :(

1. Checking the type of `CircleNode`
    - There are two ways, one is using `node->opcode()` and the other one is `dynamic_cast<luci::Circlexxxx>(node)`
    - Usually, the latter one is used but this function is not consistent
1. Using `node->arg(x)`
    - Some of codes use `node->arg(x)`
    - This is technically not wrong, but readability is really bad
    - For example, if `node` is definitely `CircleTranspose`. then following code can be used
    ```cpp
    auto circle_transpose = dynamic_cast<luci::CircleTranspose *>(node);
    if (circle_transpose == nullptr)
        return false;

    auto pred_node = circle_transpose->a();
    ...
    ```
1. Using `drop()` function
    - You don't need to call this function because dead nodes will be removed automatically.
1. (optional) Applying variable name to comment
    - Line 91 has nice diagrams.
    - What is `pred_node` and what is `main_node` in that diagram? If this is applied in the diagram, it would be wonderful.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
...
+    auto new_const_node = g->nodes()->create<luci::CircleConst>();
+
+    new_const_node->dtype(loco::DataType::S32);
+    new_const_node->rank(main_perm->rank());
+    uint32_t dim_size = 1;
+    for (uint32_t i = 0; i < new_const_node->rank(); ++i)
+    {
+      new_const_node->dim(i) = main_perm->dim(i);
+      dim_size *= main_perm->dim(i).value();
+    }
```
llFreetimell, line 75: You already assumed that rank is always 1 in `check_perm`, so you can write as following codes.

```suggestion
      new_const_node->rank(1);
      new_const_node->dim(0) = main_perm->dim(0);
```

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,125 @@
...
+    new_const_node->size<loco::DataType::S32>(dim_size);
```
llFreetimell, line 76: You already assumed that rank is always 1 in `check_perm`, so you can write as following codes.

```suggestion
      new_const_node->rank(1);
      new_const_node->dim(0) = main_perm->dim(0);
      new_const_node->size<loco::DataType::S32>(main_perm->dim(0).value());
```

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
mhs4670go:
```cxx
for (uint32_t i = 0; i < first_perm->size<loco::DataType::S32>(); i++)
```
This doesn't work?

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
mhs4670go: I think @llFreetimell is right.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: In that case, line below create same error.
```
if (first_perm->at<loco::DataType::S32>(second_perm->at<loco::DataType::S32>(i)) != i)
```
so I think at least one of both would be cast to another type.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: @llFreetimell I have a little question about `drop()`. as said above, I don't have to call `drop()` in all cases?

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
llFreetimell, line 1: Remove this file

struss: Oh, sorry. that must be auto created file from vscode. I will remove that.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
llFreetimell: Yes :)
`logo::RemoveDeadNodeWithQueryPass` will automatically remove them

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,130 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/RemoveRedundantTransposePass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleOpcode.h>
```
llFreetimell, line 20: Not used header

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
mhs4670go: oh, got it.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
jinevening: Currently, `RemoveDeadNodeWithQueryPass` is called after all passes are executed. So, if `node->drop()` is not called here, the connection between the `node` and the predecessor will be alive until `RemoveDeadNodeWithQueryPass`. This can cause unintended results in the passes between this pass and `RemoveDeadNodeWithQueryPass`.

To prevent such cases, how about running `RemoveDeadNodeWithQueryPass` after each pass (I think type/shape inference can be called after each pass as well)?

mhs4670go: @jinevening `replace` changes the user of output, which means the dropped nodes are excluded from the list of `active_node`. Since all passes starts from the `active_node` list, there is no problem I think.

jinevening: @mhs4670go `replace` changes the connections with successors. `drop` removes the connections with predecessors.

If `drop()` is not called, the connections with predecessors are alive. So, the node (dead node) is detected when `loco::succs` is called.

mhs4670go: @jinevening This is what I understand about it.
When the node changes its input like `at(0)->node(node)`, `node` method changes its `Use` list.
Say there is a node having two input and one output.
```
[input1] [input2]
      \      /
    [the_node]        [node_to_replace]
          |
      [output]
```
If we replace `the_node` with `node_to_replace`
```
                            [input1]     [input2] 
                               \          /
    [the_node]              [node_to_replace]
                                    |
                                 [output]
```
- the list of `active_node` : `output`, `node_to_replace`, `input1`, `input2`
- `loco::succs` of `node_to_replace` : `output`
- `loco::pred` of `node_to_replace` : `input1`, `input2`

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -0,0 +1,130 @@
...
+    new_const_node->dim(0) = main_perm->dim(0);
+    new_const_node->size<loco::DataType::S32>(main_perm->dim(0).value());
+    new_const_node->shape_status(luci::ShapeStatus::VALID);
+    for (uint32_t i = 0; i < main_perm->size<loco::DataType::S32>(); i++)
+    {
+      new_const_node->at<loco::DataType::S32>(i) =
+          pred_perm->at<loco::DataType::S32>(main_perm->at<loco::DataType::S32>(i));
+    }
+    replace(main_perm).with(new_const_node);
+    replace(node).with(pred_node);
```
llFreetimell, line 80: `main_perm` can be shared with other operations.
Following codes are safe.

```suggestion
    pred_node->perm(new_const_node);
    replace(node).with(pred_node);
```

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
jinevening: @mhs4670go  The problem is `loco::succs` of `input1` or `input2`.

`loco::succs` of `input1` is not only `node_to_replace` but also `the_node` (dead node). `the_node->drop()` removes the connections between `the_node` and its predecessors (`input1`, `input2`), preventing such cases.

llFreetimell: @jinevening But the point is why we should remove the connection. Eventually `the_node` will be removed at last.
If some optimizations for `the_node` was done, those will be removed too because all the nodes after `the_node` is dead.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: I will remove this.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: Okay I will fix it. Thanks for pointing out!

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: I will apply this. thanks.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: I update with using `loco::must_cast` on this.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: @jinevening @llFreetimell @mhs4670go while find node to optimize on pass, it use a `loco::active_nodes(loco::output_nodes(Graph *g))` to find a target node to optimize. and this return all nodes from `loco::output_nodes(Graph *g)` to it's predecessors. so, if one of active node replaced on one path, another node cannot see this node as a target and only successor case will be shown. so. I think just go without 'drop()' would be okay for it. Thanks for all! if any Idea to raise on, please comment galdly!

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+#include <luci/IR/CircleNodes.h>
+
+#include <vector>
+
+#include <gtest/gtest.h>
+
+namespace
+{
+void setValue(luci::CircleConst *node, const std::vector<int> &v)
```
mhs4670go, line 26: It should be one space after `namepsace` as convention.
```suggestion

void setValue(luci::CircleConst *node, const std::vector<int> &v)
```

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -0,0 +1,155 @@
...
+ *
+ *  AFTER
+ *          |                 |
+ *    [CircleNode]      [CircleConst]
+ *           \               /
+ *           [CircleTranspose]
+ *                   |
+ *
+ */
+
```
mhs4670go, line 74: Maybe this space can be removed.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -0,0 +1,155 @@
...
+  transpose2->a(transpose1);
+  transpose2->perm(perm2_node);
+
+  // output
+  auto output = g->nodes()->create<luci::CircleOutput>();
+  output->from(transpose2);
+  auto graph_output = g->outputs()->create();
+  output->index(graph_output->index());
+}
+} // namespace
```
mhs4670go, line 108:
```suggestion

} // namespace
```

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -0,0 +1,155 @@
...
+  luci::CircleTranspose *transpose_node = nullptr;
+  for (auto node : loco::active_nodes(loco::output_nodes(graph.get())))
+  {
+    auto trans = dynamic_cast<luci::CircleTranspose *>(node);
+    if (not trans)
+      continue;
+    transpose_node = trans;
+    break;
+  }
+  // Just One Transpose node, with updated perm constant.
```
mhs4670go, line 148: (trivial) It would be nice if the first letter of the sentence is uppercase, others also will be. Or lowercase is also good.
```
// create perm1
...
// No Transpose node is in Graph.
```
is should be
```
// Create perm1
...
// No Transpose node is in Graph.
```
or
```
// create perm1
...
// no Transpose node is in Graph.
```
BTW,
```
// Just "o"ne Transpose node, with updated perm constant.
```

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: Okay, I will apply it!

llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5128, opened 2020-11-23
[luci] Update BCQ application algorithm

llFreetimell:
Parent Issue : #4961

This commit will update following things

- Applying BCQ to `Einsum` will be handled within `FullyConnected`
- If `do_w_x` is true, `input` of `FullyConnected` is constant. This will be supported.

### DIFF
--- a/compiler/luci/pass/src/FuseBCQPass.cpp
+++ b/compiler/luci/pass/src/FuseBCQPass.cpp
@@ -200,19 +200,6 @@ public:
         }
       }
 
-      // Einsum is unpacked to FullyConnected, Pack and Reshape
-      if (auto reshape = dynamic_cast<luci::CircleReshape *>(node))
-      {
-        node = dynamic_cast<luci::CircleNode *>(reshape->tensor());
-      }
-      if (auto pack = dynamic_cast<luci::CirclePack *>(node))
-      {
-        if (pack->values_count() == 1 && pack->rank() == 3)
-        {
-          node = dynamic_cast<luci::CircleNode *>(pack->values(0));
-        }
-      }
-
       // Fuse FullyConnected to BCQFullyConnected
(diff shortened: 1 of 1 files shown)


seanshpark:
Q) Can you explain more about `apply` and `handle` meaning in `Applying BCQ to Einsum will be handled within FullyConnected` ?

llFreetimell:
`Einsum` in TensorFlow is converted to `FullyConnected + Pack + Reshape` in TFLite.
As a result, **applying BCQ to the weight of `Einsum`** is same as applying BCQ to the weight of `FullyConnected`.
To do so, we should find which `FullyConnected` has original weights of `Einsum`.
Finding the weight of `FullyConnected` was **handled by following codes.**
```cpp
      // Einsum is unpacked to FullyConnected, Pack and Reshape	
      if (auto reshape = dynamic_cast<luci::CircleReshape *>(node))	
      {	
        node = dynamic_cast<luci::CircleNode *>(reshape->tensor());	
      }	
      if (auto pack = dynamic_cast<luci::CirclePack *>(node))	
      {	
        if (pack->values_count() == 1 && pack->rank() == 3)	
        {	
          node = dynamic_cast<luci::CircleNode *>(pack->values(0));	
        }	
      }
```
However, as the algorithm is changed from finding `FullyConnected` to finding original weight, we do not need to use above codes.
Therefore, from now on, it is handled in the codes for `FullyConnected` as follow.
```cpp
// Fuse FullyConnected to BCQFullyConnected
if (auto fully_connected = dynamic_cast<luci::CircleFullyConnected *>(node))
{
  // Einsum is also handled in this block
}
```

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5138, opened 2020-11-24
[draft] Clang format 8

hseok-oh:
Use clang format 8 and format result
- `BraceWrapping::AfterExternBlock` - true
- `FixNamespaceComments` - false
- `SortUsingDeclarations` - false

HOWTO

1. Install clang-format-8
```
$ sudo apt install clang-format-8
```

2. Run format checker with clang-format-8
```
$ ./nnas format -clang-format clang-format-8
```


wateret:
What about applying this quickly? Sorry for other PRs but it would be better to apply this first.

wateret (approved):

PR #5140, opened 2020-11-24
[nnfwapi] Test : avoid downloaded model tests

wateret:
Use CircleGen based model files(Simple add model) where applicable.


hyunsik-yoon (approved):
LGTM

PR #5143, opened 2020-11-24
[common-artifacts] Add ReduceOps to exclude.lst

llFreetimell:
Parent Issue : #5037

This commit will add following reduce operators to `exclude.lst`

- ReduceAny
- ReduceMax
- ReduceMin
- ReduceProd
- Sum

ONE-DCO-1.0-Signed-off-by: Seok NamKoong seok9311@naver.com


mhs4670go (approved):
LGTM

seanshpark:
I assume this is a temporary change to exclude for working  #5037 and will remove this change later when all things are OK.
If so, please add a note about it in the main comment. If not, please let me know.

llFreetimell:
When we add some recipes to `res/TensorFlowLiteRecipes` for #5037, some of them are not supported by `luci-interpreter`.
Therefore, unless `luci-interpreter` supports those recipes, the exclude list will not be removed.

seanshpark:
I think I've missed something: `dynamic`. You are adding existing Ops, so I thought you where adding them for updates. But in the code, they have `dynamic` suffix which you are adding new recipes.
I also think if you add some hints for this activity that would help the reviewers.

seanshpark (approved):
LGTM

PR #5145, opened 2020-11-24
[luci] ShapeSignature Inference Algorithm for Conv2D

brainer3220:
Parent Issue : #5037

This commit will introduce shape signature inference algorithm for Conv2DTo1

Signed-off-by: brainer <brainer12345@naver.com>

### DIFF
--- a/compiler/luci/service/src/Nodes/CircleConv2D.cpp
+++ b/compiler/luci/service/src/Nodes/CircleConv2D.cpp
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <luci/Service/CircleShapeSignatureInference.h>
+
+namespace luci
+{
+
+ShapeSignature ssinf::Algorithm::visit(const luci::CircleConv2D *node)
+{
+  return signature_of_arg(node, 2);
+}
+
+} // namespace luci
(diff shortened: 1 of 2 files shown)


brainer3220:
@nnfw-bot test format-checker

compiler/luci/service/src/Nodes/CircleConv2D.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace luci
+{
+
+ShapeSignature ssinf::Algorithm::visit(const luci::CircleConv2D *node)
+{
+  return signature_of_arg(node, 0);
+}
+
+} // namespace luci
```
seanshpark, line 27: Please add an additional line here. If you see this little red circle with `-` , it means this line needs one more "ENTER" at the end.

compiler/luci/service/src/Nodes/CircleConv2D.cpp
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+#include <luci/Service/CircleShapeSignatureInference.h>
+
+namespace luci
+{
+
+ShapeSignature ssinf::Algorithm::visit(const luci::CircleConv2D *node)
+{
+  return signature_of_arg(node, 0);
```
seanshpark, line 24: I don't think this will work for `CircleConv2D `
Please refer `infer_conv2d_type()` in `compiler/luci/service/src/CircleShapeInferenceRule.cpp` 
If you need information about how output shape of convolution is, please search with `convolution output shape`

llFreetimell:
Please follow the instruction which is described in https://github.com/Samsung/ONE/issues/5037#issue-741524166.
If you do not write your name in the table, another people may be duplicated with same operation.

compiler/luci/service/src/Nodes/CircleConv2D.cpp
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+#include <luci/Service/CircleShapeSignatureInference.h>
+
+namespace luci
+{
+
+ShapeSignature ssinf::Algorithm::visit(const luci::CircleConv2D *node)
+{
+  return signature_of_arg(node, 2);
```
llFreetimell, line 24: Hmm... as far as I know, shape signature inference algorithm of `CircleConv2D` is not that simple....

llFreetimell:
@brainer3220 
If you add issue number or PR number in the commit, reference notification occurs everytime commit is changed or added.
Therefore we recommend you to follow below ground rules.

- Write commit message as follow
  ```
  [luci] ShapeSignature Inference Algorithm for Conv2D
  
  This commit will introduce shape signature inference algorithm for Conv2DTo1
  
  Signed-off-by: brainer brainer12345@naver.com
  ```

- Insert reference issue number of PR number only in the comment, not in commit message as follow
  ```
  Parent Issue : #5037
  
  This commit will introduce shape signature inference algorithm for Conv2DTo1
  
  Signed-off-by: brainer brainer12345@naver.com
  ```

seanshpark:
@llFreetimell , this is change still valid, besides the algorithm? or would it be better to close?

PR #5152, opened 2020-11-25
[res] Add dynamic tc for ReduceOps

llFreetimell:
Parent issue : #5037

This commit will add dynamic testcases for following operations

- Mean
- ReduceAny
- ReduceMax
- ReduceMin
- ReduceProd
- Sum

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

PR #5159, opened 2020-11-25
[res] Remove redundant ShapeSignature in recipes

llFreetimell:
Parent Issue : #5037
Fix #5152 

---

Some of recipes include redundant `ShapeSignature` in their recipes.

For example, let's see `ReduceAny_dynamic_003`.
Since `reduction_indices` is 1, second dimension is reduced to 1 whatever the
real value of the dimension is.

- When input shape is [2, 2, 4], output shape is [2, 1, 4]
- When input shape is [2, 3, 4], output shape is [2, 1, 4]
- ...

As a result, even the second dimension has unknown dimension,
the output shape is not dynamic!
It means that output tensor may not have shape signature.

This commit will fix those cases.


mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #5161, opened 2020-11-25
[luci] Enable ShapeSignatureInferencePass in luci

llFreetimell:
Parent Issue : #4372

This commit will enable `ShapeSignatureInferencePass` in `luci`.
FYI, operations which shape signature inference algorithm is not implemented,
it returns it's own shape signature.


seanshpark (approved):
LGTM

PR #5163, opened 2020-11-25
[tool/kernel_report] Remove onert base directory from argument

periannath:
- This commit removes onert base directory from argument
  - This script finds onert base directory by using relative path from its absolute path

### DIFF
--- a/tools/kernel_report/kernel_report.py
+++ b/tools/kernel_report/kernel_report.py
@@ -14,8 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 import argparse
+from os.path import dirname, realpath, join
 
 
 class Backend:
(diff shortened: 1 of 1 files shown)


hseok-oh (approved):
LGTM 

This script is outdated.

periannath:
@hseok-oh I'm planning to use this script to find supported operations for newly introduced backends (`ruy` and `xnnpack`). 

> This script is outdated.

Is that mean this script is not working as expected?

hseok-oh:
>Is that mean this script is not working as expected?

This script is introduced to report which tflite(circle) operation is supported by backends. But kernel generator use runtime's IR, and runtime IR is not 1:1 mapping with tflite(circle) operation any more. So it is outdated.

periannath:
@hseok-oh Ah, I dind't know that. Thanks for explaining in detail.

PR #5167, opened 2020-11-25
[bcq-tools] Fix to return correct output arrays

llFreetimell:
This commit will following bugs

- When given output arrays from user is appended, whole string is appended.
  - As-is : ['out1','out2,out3']
  - To-be : ['out1','out2','out3']
- When BCQ information is not found, returning `None` cause runtime error at next step
  - As-is : Return `None`
  - To-be : Return list of user given output arrays


seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5168, opened 2020-11-25
[luci] ShapeSignature Inference Algorithm for reduce operations

llFreetimell:
Parent Issue : #5037

This commit will introduce shape siganture algorithm for following operations

- Mean
- ReduceAny
- ReduceMax
- ReduceMin
- ReduceProd
- Sum

### DIFF
--- a/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
+++ b/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
@@ -52,13 +52,8 @@ ShapeSignature reduced_signature(const loco::Node *node, const loco::Node *indic
   auto circle_indices = loco::must_cast<const luci::CircleNode *>(indices);
   auto input_signature = circle_node->shape_signature();
 
-  // If shape signature is empty, output is static.
-  if (input_signature.rank() == 0)
-    return output_signature;
-
-  auto reduction_indices = dynamic_cast<const luci::CircleConst *>(circle_indices);
-
   // When reduction_indices is not constant
+  auto reduction_indices = dynamic_cast<const luci::CircleConst *>(circle_indices);
   if (reduction_indices == nullptr)
   {
     if (keep_dims)
@@ -78,6 +73,10 @@ ShapeSignature reduced_signature(const loco::Node *node, const loco::Node *indic
 
     return output_signature;
   }
+
+  // If shape signature is empty, output is static.
+  if (input_signature.rank() == 0)
+    return output_signature;
 
   std::vector<int32_t> reduction_values;
   if (reduction_indices->dtype() == loco::DataType::S32)
--- a/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
+++ b/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
@@ -16,11 +16,129 @@
 
 #include "luci/Service/CircleShapeSignatureInferenceHelper.h"
 
+#include <loco.h>
+
+#include <luci/Log.h>
+
+#include <oops/InternalExn.h>
+
+namespace
+{
+
+luci::ShapeSignature clean_signature(const luci::ShapeSignature &signature)
+{
+  // If shape signature has at least one -1, it is not static.
+  for (uint32_t i = 0; i < signature.rank(); ++i)
+    if (signature.dim(i) == -1)
+      return signature;
+
+  // If all dimensions are static, return empty shape signature.
+  return luci::ShapeSignature();
+}
+
+} // namespace
+
 namespace luci
 {
 
 namespace ssinf
 {
+
+ShapeSignature reduced_signature(const loco::Node *node, const loco::Node *indices, bool keep_dims)
+{
+  ShapeSignature output_signature;
+
+  auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
+  auto circle_indices = loco::must_cast<const luci::CircleNode *>(indices);
+  auto input_signature = circle_node->shape_signature();
+
+  // If shape signature is empty, output is static.
+  if (input_signature.rank() == 0)
+    return output_signature;
+
+  auto reduction_indices = dynamic_cast<const luci::CircleConst *>(circle_indices);
+
+  // When reduction_indices is not constant
+  if (reduction_indices == nullptr)
+  {
+    if (keep_dims)
+    {
+      // If keep_dims is true, rank is not changed.
+      output_signature.rank(input_signature.rank());
+      for (uint32_t i = 0; i < output_signature.rank(); ++i)
+        output_signature.dim(i) = -1;
+    }
+    else
+    {
+      // There is no way to inference for this case.
+    }
+
+    return output_signature;
+  }
+
+  std::vector<int32_t> reduction_values;
+  if (reduction_indices->dtype() == loco::DataType::S32)
+  {
+    auto reduction_size = reduction_indices->size<loco::DataType::S32>();
+    for (uint32_t i = 0; i < reduction_size; ++i)
+    {
+      int32_t axis = reduction_indices->at<loco::DataType::S32>(i);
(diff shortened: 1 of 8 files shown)


compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
```diff
@@ -16,12 +16,130 @@
...
+    if (keep_dims)
+    {
+      // If keep_dims is true, rank is not changed.
+      output_signature.rank(input_signature.rank());
+      for (uint32_t i = 0; i < output_signature.rank(); ++i)
+        output_signature.dim(i) = -1;
+    }
+    else
+    {
+      // There is no way to inference for this case.
```
seanshpark, line 73: maybe throw? or is it ok to return empty signature?

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
```diff
@@ -16,12 +16,130 @@
...
+
+} // namespace
+
 namespace luci
 {
 
 namespace ssinf
 {
 
+ShapeSignature reduced_signature(const loco::Node *node, const loco::Node *indices, bool keep_dims)
```
seanshpark, line 47: (this is about the naming convention) below used `signature_of_arg` and this differs from `X_of_Y` type
I propose to revise `signature_of_arg` to `arg_signature` to follow `reduced_signature`
- as `arg_signature` seems somewhat strange, I think `input_arg_signature` would be better to the method to express itself.

What do you think?

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
llFreetimell: Returning empty signature looked better for me..!

When TFLite converter cannot inference shape because of such cases, it returns no shape.
(The example is in https://github.com/Samsung/ONE/blob/master/res/TensorFlowPythonModels/examples/UnknownExpandDims/__init__.py)
And `luci` is supporting `ShapeStatus::NOSHAPE` for those case.

However, as I am not sure that returning empty shape signature is really correct, just throwing is also not bad :)

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
llFreetimell: I think `input_arg_signature` is good idea! Naming is always difficult for me :(

Thanks for your good idea and I will create another PR for applying it :-)

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
seanshpark: as me may not be certain for this case, how about adding `assert` just for debugging?

seanshpark: with a comment 
```cxx
// let's break with debug version just to find this case
``

**Review by seanshpark — DISMISSED** [review]
LGTM

File: compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp [hunk 3]
```diff
@@ -16,12 +16,133 @@
...
+
+      reduction_values.push_back(axis);
+    }
+  }
+  else if (reduction_indices->dtype() == loco::DataType::S64)
+  {
+    auto reduction_size = reduction_indices->size<loco::DataType::S64>();
+    for (uint32_t i = 0; i < reduction_size; ++i)
+    {
+      int32_t axis = static_cast<int32_t>(reduction_indices->at<loco::DataType::S64>(i));
```
mhs4670go, line 103: not `int64_t`?

llFreetimell: This `axis` values are only used for indicating dimension and dimension is 32bit. So it's okay :)

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
```diff
@@ -16,12 +16,133 @@
...
+  // When reduction_indices is not constant
+  if (reduction_indices == nullptr)
+  {
+    if (keep_dims)
+    {
+      // If keep_dims is true, rank is not changed.
+      output_signature.rank(input_signature.rank());
+      for (uint32_t i = 0; i < output_signature.rank(); ++i)
+        output_signature.dim(i) = -1;
+    }
```
mhs4670go, line 70: Maybe this line seems redundant because of below line.
```cxx
  if (keep_dims)
  {
    output_signature.rank(input_signature.rank());
    for (uint32_t i = 0; i < input_signature.rank(); ++i)
      output_signature.dim(i) = input_signature.dim(i);
    for (uint32_t i = 0; i < reduction_values.size(); ++i)
      output_signature.dim(reduction_values.at(i)) = 1;
  }
```

llFreetimell: These two codes looks different...? And this block ends at line 79

mhs4670go: :cold_sweat:

mhs4670go:
@llFreetimell 
BTW, it seems that `ShapeSignatureInference` consists of two part. One is a part that manages `-1`, another is that manages static case, which comes from `ShapeSignature` source code. How about moving its common function to somewhere and call that function for reducing duplicated codes?

llFreetimell:
@mhs4670go Sorry but I couldn't understand your question correctly :(

> One is a part that manages -1, another is that manages static case

AFAIK, the codes in this PR is not divided into such cases.
And usually, shape signature inference cannot be divided into static and dynamic cases because they are so different case by case.

mhs4670go:
@llFreetimell Well, the codes from line81~ in `CircleShapeSignatureInferenceHelper.cpp` are almost same with the codes from line843~ in `CircleShapeInferenceRule.cpp`. But, this kind of refactoring might as well be done later when there is lots of duplicate codes.

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
```diff
@@ -16,12 +16,132 @@
...
+  auto input_signature = circle_node->shape_signature();
+
+  // When reduction_indices is not constant
+  auto reduction_indices = dynamic_cast<const luci::CircleConst *>(circle_indices);
+  if (reduction_indices == nullptr)
+  {
+    if (keep_dims)
+    {
+      // If keep_dims is true, rank is not changed.
+      output_signature.rank(input_signature.rank());
```
mhs4670go, line 62: Is it possible for `output_signature` to have a length of 2 or more?

llFreetimell: `ShapeSignature` can have two kinds of rank.

- Rank 0 : It means that there is not `ShapeSignature`. In short, shape is static.
- Same rank with shape : It means that at least one of shape dimension is unknown and thus rank of `ShapeSignature` will be same with `Shape`.

mhs4670go (approved):
LGTM

seanshpark:
Just curious, how can we test the algorithm is really working for this PR?

mhs4670go:
@seanshpark Right.. Maybe we should create `CircleShapeSignatureInference.test.cpp` or sth. Recently, we've been posting a PR without test. Let's try to make PR with test:)

llFreetimell:
For now, following codes is checking the algorithm.

https://github.com/Samsung/ONE/blob/46eedd09d380f38670e21e3b5511c580dc433ba7/compiler/luci/service/src/Validate.cpp#L171-L172

But I agree that more tests is helpful. I will think about how can we test more better.

seanshpark:
For now, I'll land this PR :)

seanshpark (approved):
LGTM

PR #5179, opened 2020-11-26
[oops] Revise CTOR for type conversion

seanshpark:
This will revise CTORs for int to uint type conversion


llFreetimell (approved):
LGTM

PR #5184, opened 2020-11-26
[onert] Check same data type for AddN operation

hseok-oh:
Check all operand type are same for AddN operation


ragmani (approved):
LGTM

PR #5188, opened 2020-11-26
DRAFT luci use unsigned for height/width

seanshpark:
on-going draft of luci to use unsigned for attributes height/width


seanshpark:
Test done.

PR #5189, opened 2020-11-26
[onert] Introduce TraceCtx that has profiling information per session

hyunsik-yoon:
This introduces `TraceCtx` that has profiling information per session.

Why designed this way:
- Designing an object per session is better for multi-threads where a thread
runs a session

Replaces #5101
Parent issue: #4901
Draft: #4903

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.h
+++ b/runtime/onert/api/src/nnfw_api_internal.h
@@ -158,7 +158,7 @@ private:
   std::shared_ptr<onert::exec::Execution> _execution;
   std::shared_ptr<onert::frontend::custom::KernelRegistry> _kernel_registry;
 
-  std::shared_ptr<onert::util::TracingCtx> _tracing_ctx;
+  std::unique_ptr<onert::util::TracingCtx> _tracing_ctx;
 };
 
 #endif // __API_NNFW_API_INTERNAL_H__
--- a/runtime/onert/api/src/nnfw_api_internal.h
+++ b/runtime/onert/api/src/nnfw_api_internal.h
@@ -21,6 +21,7 @@
 #include "nnfw_experimental.h"
 
 #include <util/GeneralConfigSource.h>
+#include <util/TracingCtx.h>
 
 #include <string>
 #include <memory>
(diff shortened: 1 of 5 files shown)


hseok-oh:

hyunsik-yoon:

runtime/onert/api/src/nnfw_api_internal.h
```diff
@@ -156,6 +157,8 @@ struct nnfw_session
   std::unique_ptr<onert::compiler::Compiler> _compiler;
   std::shared_ptr<onert::exec::Execution> _execution;
   std::shared_ptr<onert::frontend::custom::KernelRegistry> _kernel_registry;
+
+  std::shared_ptr<onert::util::TracingCtx> _tracing_ctx;
```
wateret, line 161: Could you check out https://github.com/Samsung/ONE/pull/4903/files#r531345125 ?

wateret:
I left comments on #4903 .

runtime/onert/core/include/util/TracingCtx.h
```diff
@@ -0,0 +1,70 @@
...
+class TracingCtx
+{
+public:
+  /**
+   * @brief Create and store unique session id managed by this class
+   *        Note that this constructor can be called by multiple sessions running in parallely.
+   */
+  TracingCtx()
+  {
+    std::unique_lock<std::mutex> lock{_session_id_mutex};
```
wateret, line 43: This mutex must be static too. With this way, `next_session_id` is not protected. Each of this class will lock different mutexes, which is useless 😃

wateret (changes requested):
PTAL at the comment (LGTM other that that)

runtime/onert/core/include/util/TracingCtx.h
hyunsik-yoon: wow! wow! wow!!!

Some more info: https://stackoverflow.com/questions/17357909/do-we-need-to-have-static-lock-for-static-member-variable-of-the-class-in-c

hyunsik-yoon: fixed. Thanks!

hyunsik-yoon:
@wateret PTAL, thanks!

wateret (approved):
LGTM

PR #5191, opened 2020-11-26
[luci] Use unsigned for width/height attributes

seanshpark:
This will revise luci width/height of Dilation, Filter and Stride to use
unsigned values.

### DIFF
--- a/compiler/luci/lang/include/luci/IR/AttrDilation.h
+++ b/compiler/luci/lang/include/luci/IR/AttrDilation.h
@@ -18,6 +18,7 @@
 #define __LUCI_IR_ATTRDILATION_H__
 
 #include <stdint.h>
+#include <cassert>
 
 namespace luci
 {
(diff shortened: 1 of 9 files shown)


seanshpark:
This is an update for our NPU compiler

mhs4670go (dismissed):
LGTM

compiler/luci/lang/include/luci/IR/AttrDilation.h
```diff
@@ -27,15 +28,25 @@ class Dilation final
...
-
-  int32_t h() const { return _h; }
-  void h(int32_t h) { _h = h; }
+  uint32_t w() const { return _w; }
+  void w(uint32_t w) { _w = w; }
+  void w(int32_t w)
+  {
+    assert(w >= 0);
+    _w = static_cast<uint32_t>(w);
+  }
```
parjong, line 37: (Optional) Why don't you hide cassert include in .cpp?

parjong (dismissed):

compiler/luci/lang/include/luci/IR/AttrDilation.h
mhs4670go: @parjong Just curiosity, is there any reason of this? Is that for good looking?

parjong: > Just curiosity, is there any reason of this? Is that for good looking?

The goal is to minimize impact on namespace (or symbol table), which eventually minimizes the coupling between components.

The current implementation allows clients (= the code that includes this header) to use assert without including 'cassert' header.

This may introduce implicit coupling between two components, which makes it a bit difficult to update this header later. 

For example, let us assume that we want to remove assert here. Suddenly, build breaks from other components because they uses assert without ``#include <cassert>`` statement.

parjong (approved):
👍

mhs4670go (approved):
LGTM

PR #5194, opened 2020-11-27
[onert] Fix validation check for BatchToSpaceND

hseok-oh:
- Input/output type check
- Crops operand value check in acl backend (non-zero is not supported)
- Crops operand shape check in ShapeValidator


seanshpark (approved):
LGTM

Issue #5121, opened 2020-11-23
[circle-quantizer] Update CWQ algorithm for some Ops (PRelu, InstanceNorm)


jinevening:
Let's support new CWQ algorithms for the following Ops.

## 1. PRelu

Alpha is quantized per-channel. Alpha has a single value per channel, so its fp value is exactly restored from quantized value.

### uint8

Elem | Input | Alpha | Output
--|--|--|--
Dtype | uint8 | uint8 | uint8
Granularity | per-tensor | per-channel | per-tensor

### int16

Elem | Input | Alpha |Output
--|--|--|--
Dtype | int16 | int16 | int16
Granularity | per-tensor | per-channel | per-tensor

## 2. Instance Normalization

gamma and beta are quantized per-channel. Epsilon is saved as-is (fp32). Gamma (and beta) has a single value per channel, so its fp value is exactly restored from quantized value.

### uint8

Elem | Input | Gamma | Beta | Epsilon |Output
--|--|--|--|--|--
Dtype | uint8 | uint8 | uint8 | fp32 | uint8
Granularity | per-tensor | per-channel | per-channel | scalar | per-tensor

### int16

Elem | Input | Gamma | Beta | Epsilon |Output
--|--|--|--|--|--
Dtype | int16 | int16 | int16 | fp32 | int16
Granularity | per-tensor | per-channel | per-channel | scalar | per-tensor

jinevening:
Commit order (based on module name)

DRAFT
luci (quantization algorithm)
common-artifacts (exclude.lst)
res (recipe)
pota-quantization-value-test (test data + test.lst)

jinevening:
All done.

PR #5198, opened 2020-11-27
[luci] Quantization algorithm for InstanceNorm

jinevening:
This adds quantization algorithm for InstanceNorm Op

---
Related to: #5121
Draft PR: #5197

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -86,6 +86,99 @@ void quant_const_values(luci::CircleConst *const_node, float scaling_factor, flo
   }
 }
 
+// Quantize const per channel
+//
+// The last dimension of const is the same as the dimension of channel
+// And the rest of the const dimensions should be 1
+// So, a 'single value' is quantized per channel
+//
+// Quantization spec (f: fp value, q: quantized value)
+//
+// uint8
+//   Positive f: f = f * (q - 0) [q = 1, scale = f, zp = 0]
+//   Negative f: f = (-f) * (q - 1) [q = 0, scale = -f, zp = 1]
+//
+// int16
+//   Positive f: f = f * (q - 0) [q = 1, scale = f, zp = 0]
+//   Negative f: f = (-f) * (q - 0) [q = -1, scale = -f, zp = 0]
+void quant_const_per_channel(CircleConst *node, loco::DataType quant_type)
+{
+  assert(node->dtype() == loco::DataType::FLOAT32);
+  assert(node->rank() > 0);
+
+  for (uint32_t i = 0; i < node->rank() - 1; i++)
+  {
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -86,6 +86,99 @@ void quant_const_values(luci::CircleConst *const_node, float scaling_factor, flo
...
+//   Negative f: f = (-f) * (q - 0) [q = -1, scale = -f, zp = 0]
+void quant_const_per_channel(CircleConst *node, loco::DataType quant_type)
+{
+  assert(node->dtype() == loco::DataType::FLOAT32);
+  assert(node->rank() > 0);
+
+  for (uint32_t i = 0; i < node->rank() - 1; i++)
+  {
+    if (node->dim(i).value() != 1)
+      throw std::runtime_error("Non-channel demension of const node must be 1");
```
seanshpark, line 112: (opional) maybe add a comment why there is throw instead of just return and ignore

seanshpark (approved):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -612,10 +705,31 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
+  if (granularity == QuantizationGranularity::LayerWise)
+  {
+    quant_const(gamma, output_type);
+    quant_const(beta, output_type);
+  }
+  else
+  {
+    quant_const_per_channel(gamma, output_type);
+    quant_const_per_channel(beta, output_type);
+  }
```
mhs4670go, line 725:
```suggestion
  else if (granularity == QuantizationGranularity::ChannelWise)
  {
    quant_const_per_channel(gamma, output_type);
    quant_const_per_channel(beta, output_type);
  }
  else
    throw std::runtime_error("Quantization granularity must be either 'layer' or 'channel'");
```

seanshpark: if `QuantizationGranularity` may provide another type then this change would be good

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: I've added the comment.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: I've updated the code.  Thanks.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

jinevening:
@seanshpark Can you merge this PR? :)

PR #5200, opened 2020-11-27
[onert/cpu] Switch-case for DepthToSpace

hseok-oh:
Use switch-case to call DepthToSpace kernel for each type


seanshpark (approved):
LGTM thank you!

wateret (approved):

PR #5204, opened 2020-11-27
[onert] Adds missing header guard into GenModelTest.h

hyunsik-yoon:
This adds missing header guard into `GenModelTest.h`.

The prefix `__NNFW_API_TEST_` follows the way of other header files in the same directory.


seanshpark (approved):
LGTM

wateret (approved):

PR #5206, opened 2020-11-27
[onert] Show shape information in Chrome Tracing

hyunsik-yoon:
This adds user data field where we can store input shape that eventually
be shown in Chrome Tracing.

During testing a model I am working on, I needed this feature. For example,

- Assume that you have a model with 1000 operations where 20 operation
  is op type _FOO_.
- Some conversion is done to convert _FOO_ to optimized _BAR_ op
  and your converter reduced the total number of operations to 80
  after optimization.
- Now, you want to compare the performance of a specific _FOO_ and _BAR._
  The _FOO_ op has operation index, e.g, 30, but 30th op in optimized model
  is not BAR. (since the total number of operations are reduced).

In such case, some hint should be provided to find the right _BAR._
You can use this input shape as a hint if `FOO's input share == BAR's input shape`.

### DIFF
--- a/runtime/onert/core/src/exec/ExecutionObservers.cc
+++ b/runtime/onert/core/src/exec/ExecutionObservers.cc
@@ -17,12 +17,61 @@
 #include "exec/ExecutionObservers.h"
 
 #include <string>
+#include <sstream>
 
 #include "util/logging.h"
 #include "exec/IExecutor.h"
 #include "misc/polymorphic_downcast.h"
 #include "ir/OpSequence.h"
 #include "util/EventWriter.h"
+
+namespace
+{
+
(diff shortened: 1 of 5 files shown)


hyunsik-yoon:
- `args` is specified in ChromeTracing JSon spec document:
  - https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit#heading=h.q8di1j2nawlp
- result screen
image

hyunsik-yoon:

hyunsik-yoon:

wateret (approved):
LGTM

PR #5221, opened 2020-11-30
[tfldump] Support to dump metadata in Model

llFreetimell:
If we dump `.tflite` file, there was always one dummy buffer with some values.
It was turned out to `metadata` in `Model`.
This commit will enable supporting to dump `metadata` in `Model`.

### DIFF
--- a/compiler/tfldump/src/Dump.cpp
+++ b/compiler/tfldump/src/Dump.cpp
@@ -384,12 +384,15 @@ void dump_model(std::ostream &os, const tflite::Model *model)
   os << std::endl;
 
   // dump metadata
-  os << "metadata : B(index) name" << std::endl;
-  for (uint32_t i = 0; i < metadata->Length(); ++i)
+  if (metadata != nullptr)
   {
-    os << "B(" << metadata->Get(i)->buffer() << ") " << metadata->Get(i)->name()->c_str();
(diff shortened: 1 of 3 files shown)


mhs4670go (dismissed):
LGTM

llFreetimell:
@nnfw-bot test-nncc-release

llFreetimell:
@nnfw-bot test nncc-release

compiler/tfldump/src/Dump.cpp
```diff
@@ -382,6 +383,14 @@ void dump_model(std::ostream &os, const tflite::Model *model)
   }
   os << std::endl;
 
+  // dump metadata
+  os << "metadata : B(index) name" << std::endl;
```
seanshpark, line 387: Maybe surround with `if (metadata != nullptr) {` ?

seanshpark: I've checked with old inception_v3_2018_04_27.tgz which doesn't seem to have metadata; _tfldump_ just exits without any message.

seanshpark (approved):
LGTM

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5242, opened 2020-12-02
[luci] Revise reduced signature algorithm

llFreetimell:
This commit will revise following things for `reduced_siganture` function in `CircleShapeSigantureInferenceHelper.cpp`

- Add more comments for detailed description
- Remove assert for uncertain case and add a log
  - When shape cannot be inferred, returning empty shae in TFLite converter is comfirmed.
- Support when signature of input is empty
  - Even if signauture of input is empty, signature of output can be non-empty.

### DIFF
--- a/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
+++ b/compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
@@ -52,7 +52,6 @@ ShapeSignature reduced_signature(const loco::Node *node, const loco::Node *indic
   ShapeSignature output_signature;
 
   auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
-
   if (circle_node->shape_signature().rank() > 0)
     input_signature = circle_node->shape_signature();
   else
@@ -61,6 +60,13 @@ ShapeSignature reduced_signature(const loco::Node *node, const loco::Node *indic
     for (uint32_t i = 0; i < circle_node->rank(); ++i)
       input_signature.dim(i) = circle_node->dim(i).value();
   }
+
+  // If input rank is 0, it means that one of following case is occurred.
+  // - Input is scalar : result is always scalar
+  // - Input shape signature is not inferenced : cannot infer output shape signauture
+  // Therefore, when input signature rank is 0, always return empty signature.
+  if (input_signature.rank() == 0)
+    return output_signature;
 
   // When reduction_indices is not constant
   auto reduction_indices = dynamic_cast<const luci::CircleConst *>(indices);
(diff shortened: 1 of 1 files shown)


compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
```diff
@@ -46,14 +46,24 @@ namespace ssinf
...
 {
+  LOGGER(l);
+
+  ShapeSignature input_signature;
   ShapeSignature output_signature;
 
   auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
-  auto circle_indices = loco::must_cast<const luci::CircleNode *>(indices);
-  auto input_signature = circle_node->shape_signature();
+
```
seanshpark, line 55: maybe remove this line to follow below `auto reduction_indices = ` line?

seanshpark: or go as-is...

seanshpark (dismissed):
LGTM

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
llFreetimell: You mean following modification?
```suggestion
```
If so, there's no reason to refuse :-)

compiler/luci/service/src/CircleShapeSignatureInferenceHelper.cpp
```diff
@@ -66,15 +75,18 @@ ShapeSignature reduced_signature(const loco::Node *node, const loco::Node *indic
...
     }
 
     return output_signature;
   }
 
-  // If shape signature is empty, output is static.
+  // If input rank is 0, it means that one of following case is occurred.
+  // - Input is scalar : result is always scalar
+  // - Input shape signature is not inferenced : cannot infer output shape signauture
+  // Therefore, when input signature rank is 0, always return empty signature.
```
mhs4670go, line 89: I've got two question about this lines.
1. It's not related with this PR. I think below condition can be moved to above line 65. Because even if either `reduction_indices` is `nullptr` or not, there is no difference for `output_signiture`'s being scalar.
2. Could you let me know about the second one more detail?; `Input shape signature is not inferenced` Because of below lines, this circumstance wouldn't happen I think.
```cxx
else // maybe the case of `input_signature.rank()==0`
  {
    input_signature.rank(circle_node->rank());
    for (uint32_t i = 0; i < circle_node->rank(); ++i)
      input_signature.dim(i) = circle_node->dim(i).value();
  }
```

llFreetimell: 1. Before I created this PR, below condition should be there but for now, I think you are right. I will move it.

1. It can happens. There are some cases in TensorFlow and TFLite that shape cannot be inferred.
For example, let's think about following codes.
```python
inp = tf.compat.v1.placeholder(tf.float32, shape=(2,3,4))
idx = tf.compat.v1.placeholder(tf.int32, shape=(2))
r_mean1 = tf.compat.v1.math.reduce_mean(inp, idx, False)
r_mane2 = tf.compat.v1.math.reduce_mean(r_mean1, [0], True)
```
Because shape of `r_mean1` cannot be inferred, both shape and shape signature of `r_mean1` are `[]`.
In this case, `input_signature` is `[]`, which rank is 0.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #4796, opened 2020-10-27
[Compiler][luci] Refactoring luci/Service for removing dependency with loco


llFreetimell:
Originated Issue : #4726 

---

### Backgrounds

Until now, `CircleNode` has two kinds of shape and dtype.
One is shape/dtype of `loco` as `AnnotedItem` and the other is shape/dtype of `CircleNode`.
However, during `ShapeInferencePass` and `TypeInferencePass`, only shape/dtype of `loco` is updated.
shape/dtype of `CircleNode` is only updated after final exporting step.
These redundancy is not desirable so refactoring to remove these dependency is required.

### Goals

1. Reorder and change phase mode
    - To Keep consistency of shape/dtype whenever some passes change them.
1. Inference shape/dtype only with shape/dtype of circle node, not `loco`.
    - To remove redundant dependency between `luci` and `loco`
1. Support static/dynamic shape inference with `CircleShapeSignatureInferenceHelper`
    - To reduce duplicated code and logic.

### ToDo

- [x] Revise `ShapeSignature` inference related functions #5254
- [x] Reorder `luci::Pass` and change phase mode to `RESTART` #5243
- [x] Introduce `MigrateLegacyShapeDtypePass` to keep consistency. #5247
- [x] Enable `MigrateLegacyShapeDtypePass` and remove `copy_shape_dtype`. #5261 ~~#5269~~ #5276
- [ ] Introduce new shape inference rule and algorithm #5257
- [x] Introduce new dtype inference rule and algorithm #5259
- [ ] Introduce new shape inference pass
- [x] Introduce new dtype inference pass #5279
- [ ] Introduce `CheckCircleRulesPass` for checking correctness of new circle rules
- [ ] Enable new passes in optimizers and related users.
- [ ] Remove loco dependency #5281 #5285 #5286
- [ ] Remove unused codes.
- [ ] Implement new shape/dtype inference algorithm (127 ops)
    - Most of logics will be copied from current algorithm
    - But some of them should be updated not to use `loco` and be applicable for both shape and shape siganture.
    - This may be handled in another issue.
- [ ] Remove old shape/dtype inference service and related codes
- [ ] Remove all of temporary passes after all related works are finished

### Draft

Draft PR : #4850

#### Review points

There are many commits in the draft PR, which are grouped by similar contents. Please refer to each of them.

llFreetimell:
@seanshpark @jinevening @mhs4670go @struss 

I created a draft PR with seven separate commits.
[...]

PR #5243, opened 2020-12-02
[luci] Reorder sequence of passes and change phase mode

llFreetimell:
Parent Issue : #4796 

Until now, consistency between each passes is not kept.
For example, if passA created some new nodes and passB want to use the nodes,
passB should always check whether the shape/dtype of newly created nodes are inferenced or not.
There is more critical example.
If passC modify shape/dtype of some nodes as a result of optimization and passD are to use them,
passD will reference not updated shape/dtype and it may cause unintended errors.

This commit will resolve those issues by following modifications

- Change order the sequence of basic services pass
  - Type/Shape/ShapeSignature inference pass
  - `RemoveDeadNodeWithQueryPass`
- Change phase mode to `RESTART` from `SATURATE`

### DIFF
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -133,6 +133,7 @@ void CircleOptimizer::optimize(luci::Module *m) const
 {
   luci::Phase phase;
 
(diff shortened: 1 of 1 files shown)


mhs4670go (approved):
LGTM

jinevening (approved):
LGTM

PR #5247, opened 2020-12-02
[luci] Introduce MigrateLegacyShapeDtypePass in luci

llFreetimell:
Parent Issue : #4796

Until refactoring for removing dependency between `loco` shape and circle shape is finished,
shape and dtype of `loco` should be copied to circle nodes to keep consistency.
This commit will introduce `CopyLocoItemsToCirclePass` and this pass will do it.
FYI, be careful that when copying `loco` shape to circle node, shape signature of the circle node
is considered, but shape of `loco` is not changed.

When the refactoring is finished, this pass will be removed.

### DIFF
--- a/compiler/luci/pass/src/MigrateLegacyShapeDtypePass.cpp
+++ b/compiler/luci/pass/src/MigrateLegacyShapeDtypePass.cpp
@@ -67,6 +67,9 @@ bool MigrateLegacyShapeDtypePass::run(loco::Graph *g)
     {
       auto loco_shape = loco::shape_get(node).as<loco::TensorShape>();
 
+      assert(circle_node->shape_signature().rank() == 0 ||
+             circle_node->shape_signature().rank() == loco_shape.rank());
+
       // When shape of loco is copied to circle node, ShapeSignature should be applied.
       loco::TensorShape new_shape;
       new_shape.rank(loco_shape.rank());
--- a/compiler/luci/pass/src/MigrateLegacyShapeDtypePass.cpp
+++ b/compiler/luci/pass/src/MigrateLegacyShapeDtypePass.cpp
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#include "luci/Pass/CopyLocoItemsToCirclePass.h"
+#include "luci/Pass/MigrateLegacyShapeDtypePass.h"
 
 #include <loco/Service/ShapeInference.h>
 #include <loco/Service/TypeInference.h>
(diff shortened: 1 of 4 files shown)


seanshpark:
Pass name `CopyLocoItemsToCircle` is somewhat strange...

seanshpark:
How about `MigrateLegacyShapeDTypePass` ?

llFreetimell:
`MigrateLegacyShapeDtypePass` looks elegant, I will use it :)

mhs4670go:
(optional) Umm.. `MigrateCanonicalShapeDtypePass` ..?

compiler/luci/pass/src/MigrateLegacyShapeDtypePass.cpp
```diff
@@ -0,0 +1,109 @@
...
+    {
+      auto loco_shape = loco::shape_get(node).as<loco::TensorShape>();
+
+      // When shape of loco is copied to circle node, ShapeSignature should be applied.
+      loco::TensorShape new_shape;
+      new_shape.rank(loco_shape.rank());
+      for (uint32_t i = 0; i < loco_shape.rank(); ++i)
+      {
+        if (circle_node->shape_signature().rank() > 0 &&
+            circle_node->shape_signature().dim(i) == -1)
```
seanshpark, line 76: I'm not sure for this one but can `circle_node->shape_signature().rank() == 2` and `loco_shape.rank() == 3`  ?
or maybe add `assert(circle_node->shape_signature().rank() == loco_shape.rank())` ?

seanshpark:
When this lands and this Pass is on, please run _circle2circle_ with our internal models...

seanshpark:
(optional) it would be nice if you could add comments about what the intention of the codes are...
after some months or years passes, we may forget them...

compiler/luci/pass/src/MigrateLegacyShapeDtypePass.cpp
llFreetimell: Those error will be checked at following codes.
(Following codes are in master branch)

https://github.com/Samsung/ONE/blob/5a2e89734126996dfbe90d2defc2b1474b623c60/compiler/luci/service/src/Validate.cpp#L121-L127

However, adding `assert` looks good! I will add it :)

---

FYI, I am planning to validate refactoring codes more strictly by following codes.
(It is draft code)
https://github.com/Samsung/ONE/blob/42a5ce43f46e3e22f55b212dcc5d631269381154/compiler/luci/pass/src/CheckCircleRulesPass.cpp#L103-L136

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

PR #5249, opened 2020-12-02
[one-cmds] Fix wrong argument value

llFreetimell:
This commit will fix wrong argument value.
If `converter_version` is used, `--` should be removed


llFreetimell:
This issue will be handled by @mhs4670go :)

So close this PR!

PR #5253, opened 2020-12-02
[onert/test] TFlite model test support all input type

hseok-oh:
- Support all input type
- Implement comparison function for all data type (copy from GenModelTest.h)
- Support thread setting environment variable

---

Part of #5212

### DIFF
--- a/tests/tools/tflite_loader/src/tflite_loader.cc
+++ b/tests/tools/tflite_loader/src/tflite_loader.cc
@@ -46,7 +46,7 @@ const float DIFFERENCE_THRESHOLD = 10e-5;
   }
 
 // Read vector of floats from selected file
-void readData(const string &path, uint8_t *dest)
+void readData(const string &path, std::vector<uint8_t> &dest)
 {
   std::ifstream in(path);
   if (!in.good())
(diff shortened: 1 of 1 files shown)


tests/tools/tflite_loader/src/tflite_loader.cc
```diff
@@ -45,7 +46,7 @@ const float DIFFERENCE_THRESHOLD = 10e-5;
   }
 
 // Read vector of floats from selected file
-std::vector<float> readData(const string &path)
+void readData(const string &path, uint8_t *dest)
```
seanshpark, line 49: (optional) maybe add some assert checkes for `dest` ? caller can call with `nullptr`
also for below codes

seanshpark (approved):
LGTM

seanshpark:
I think there should be another reviewer as I don't have full knowledge of the codes :)

hseok-oh:
@glistening PTAL

hseok-oh:
PTAL

seanshpark (approved):
LGTM

hseok-oh:
I've updated to resolve conflict.

tests/tools/tflite_loader/src/tflite_loader.cc
```diff
@@ -264,31 +318,69 @@ int main(const int argc, char **argv)
...
+        break;
+      case NNFW_TYPE_TENSOR_QUANT8_ASYMM_SIGNED:
+        matched = compareBuffersExact<int8_t>(ref_output.int8, output, out_idx);
+        break;
+      case NNFW_TYPE_TENSOR_INT32:
+        matched = compareBuffersExact<int32_t>(ref_output.i32, output, out_idx);
+        break;
+      case NNFW_TYPE_TENSOR_FLOAT32:
+        // TODO better way for handling FP error?
+        for (uint32_t e = 0; e < num_elems(&ti); e++)
```
glistening, line 352: @hseok-oh (optional) We may compare in the same way of `tflite_run`:

```c++
auto equals = [tolerance](float lhs, float rhs) {
  // NOTE Hybrid approach
  // TODO Allow users to set tolerance for absolute_epsilon_equal
  if (nnfw::misc::fp32::absolute_epsilon_equal(lhs, rhs))
  {
    return true;
  }

  return nnfw::misc::fp32::epsilon_equal(lhs, rhs, tolerance);
};
```

I am not saying `tfflite_run`'s way is best, but would like to use same method.

glistening (approved):
LGTM

tests/tools/tflite_loader/src/tflite_loader.cc
hseok-oh: Thanks. I'll try that later.

PR #5254, opened 2020-12-02
[luci] Revise ShapeSignatureInferenceHelper functions

llFreetimell:
Parent Issue : #4796

This commit will revise following things

- Move `clean_signature` to `namespace luci`
  - It would be better to make it re-usable
- Revise `reduced_signature` to return redundant signature
  - Core logic can be shared by `ShapeInference`
  - Then code duplication would be reduced
- Add more comments for detailed description

### DIFF
--- a/compiler/luci/service/include/luci/Service/CircleShapeSignatureInferenceHelper.h
+++ b/compiler/luci/service/include/luci/Service/CircleShapeSignatureInferenceHelper.h
@@ -26,8 +26,16 @@ namespace luci
 namespace ssinf // Namespace for Shape Signature Inference
 {
 
+// Return empty signature if all of dimensions are known.
+// If one of dimensions is unknown, return signature without change.
+ShapeSignature clean_signature(const luci::ShapeSignature &signature);
+
+// Return reduced input_signature with indices and keep_dims.
+//  - indices : reduction index
+//  - keep_dims : If true, rank is not changed. If false, rank is reduced along indices.
 ShapeSignature reduced_signature(const loco::Node *node, const loco::Node *indices, bool keep_dims);
 
+// Return signature of index-th argument of node.
 ShapeSignature input_arg_signature(const luci::CircleNode *node, uint32_t index);
 
 } // namespace ssinf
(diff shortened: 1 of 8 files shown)


compiler/luci/service/include/luci/Service/CircleShapeSignatureInferenceHelper.h
```diff
@@ -26,8 +26,16 @@ namespace luci
 namespace ssinf // Namespace for Shape Signature Inference
 {
 
+// Return empty signature if all of dimensions are known.
+// If one of dimensions is unknown, return signature without change.
+ShapeSignature clean_signature(const luci::ShapeSignature &signature);
```
seanshpark, line 31: I first though of this to clean(? not sure what the meaning of 'clean' maybe) the signature values... but actually it isn't.

seanshpark: I think we need to use another name than `clean`... let's take some time and think about this...

seanshpark: `normalize_` `validate_` regularize_` `legalize_`
I've seen `legalize` in LLVM that makes things make sense. how about `legalize_signature` ?

llFreetimell: Always thanks for good idea :-)

seanshpark (approved):
LGTM thank you!

mhs4670go:
Ummm.. the legalization's being move into "reduce" function looks better.

llFreetimell:
@mhs4670go 

At first, I wrote the code as you said.
However, illegal(?) siganture should be shared because of `ShapeInference` :(
Please refer to following draft code.

https://github.com/Samsung/ONE/blob/275b27628bf741ff604d2b41045ecdad4e49497c/compiler/luci/service/src/Nodes/CircleReduceProd.cpp

mhs4670go (approved):
LGTM

PR #5257, opened 2020-12-02
[luci] Introduce new circle shape inference

llFreetimell:
Parent Issue : #4796

Until now, `ShapeInference` had some issues.

- Because of dependency with `loco`, circle node has two kinds of shape.
- Current inference algorithm do not support dynamic shape inference.
- Current inference algorithm do not check the status of input nodes

This commit will introduce new circle shape inference in new namespace `sinf`,
which means Shape INFerence, with new Rule, Algorithm and Helper.

### DIFF
--- a/compiler/luci/service/include/luci/Service/CircleShapeInferenceHelper.h
+++ b/compiler/luci/service/include/luci/Service/CircleShapeInferenceHelper.h
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


compiler/luci/service/include/luci/Service/CircleShapeInferenceHelper.h
```diff
@@ -0,0 +1,38 @@
...
+{
+
+namespace sinf // Namespace for Shape Inference
+{
+
+// Return shape of circle node as loco::TensorShape
+loco::TensorShape circle_shape(const luci::CircleNode *node);
+
+} // namespace sinf
+
```
seanshpark, line 35: (optional) here too

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5259, opened 2020-12-02
[luci] Introduce new circle type inference

llFreetimell:
Parent Issue : #4796

Until now, `TypeInference` had some issues.

- Because of dependency with `loco`, circle node has two kinds of type.
- Current inference algorithm do not check the status of input nodes

This commit will introduce new circle type inference in new namespace `tinf`,
which means Type INFerence, with new Rule, Algorithm and Helper.

### DIFF
--- a/compiler/luci/service/include/luci/Service/CircleTypeInference.h
+++ b/compiler/luci/service/include/luci/Service/CircleTypeInference.h
@@ -21,6 +21,10 @@
 
 #include <mio/circle/schema_generated.h>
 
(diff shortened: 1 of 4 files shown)


seanshpark:
typo in commit message `Thape `

llFreetimell:
Updated, PTAL :)

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5265, opened 2020-12-03
[onert] Unify StaticTensorManager impl

wateret:
Make `cpu_common::StaticTensorManager` implementation be same with
`cpu::StaticTensorManager`. It causes changes in `ConstantInitializer`,
`TensorBuilder` and `ExternalTensor`.


wateret:
After this, all cpu-based backends will reuse `cpu_common`'s modules.

YongseopKim (approved):
LGTM

PR #5267, opened 2020-12-03
[onert] Remove redundant const init in cpu backend

wateret:
Registering `ExternalData` is default behavior, so manually adding
custom initializers is redundant(as what all of them do is registering
ExternalData).


YongseopKim (approved):
LGTM

PR #5270, opened 2020-12-03
[onert] Unify StaticTensorManager class

wateret:
Make cpu/ruy/xnnpack reuse `cpu_common::StaticTensorManager`. (They all have identical implementations)


hseok-oh (approved):
LGTM

PR #5279, opened 2020-12-03
[luci] Introduce new type inference pass

llFreetimell:
Parent Issue : #4796 

This commit will introduce new type inference pass to remove dependency with `loco`.
The new type inference pass uses `luci::tinf::Rule`.

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/CircleTypeInferencePass.h
+++ b/compiler/luci/pass/include/luci/Pass/CircleTypeInferencePass.h
@@ -34,7 +34,7 @@ public:
 
 public:
   bool run(luci::Module *m);
-  bool run(loco::Graph *graph);
+  bool run(loco::Graph *g);
 };
 
 } // namespace luci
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/include/luci/Pass/CircleTypeInferencePass.h
```diff
@@ -0,0 +1,42 @@
...
+ * @brief Pass to infer type of circle nodes
+ */
+class CircleTypeInferencePass : public luci::Pass
+{
+public:
+  virtual const char *name(void) const { return "luci::CircleTypeInferencePass"; }
+
+public:
+  bool run(luci::Module *m);
+  bool run(loco::Graph *graph);
```
seanshpark, line 37: I think you can use `loco::Graph *g` here or use `luci::Module *module`. In previouse PR where there was a change but that change was not related with the purpose. Here, you're adding a new class and it would be better to make things similar.

seanshpark: So, it would be nice if you could change to `loco::Graph *g` for that preiouse code.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #5280, opened 2020-12-04
[onert/cpu] Fix BackedContext getTensor bug

hseok-oh:
Skip finding tensor's use set when tensor is defined but used nowhere

---

It will resolve test fail https://github.com/Samsung/ONE/issues/5271#issuecomment-737858216
Test: `res/TensorFlowLiteRecipes/Unpack_003/test.recipe`

### DIFF
--- a/runtime/onert/backend/cpu/BackendContext.cc
+++ b/runtime/onert/backend/cpu/BackendContext.cc
@@ -62,6 +62,8 @@ ITensorRegistry *BackendContext::genTensors(const std::vector<onert::ir::OpSeque
     if (model_io.contains(index))
       continue;
     const auto &obj = graph()->operands().at(index);
+    if (obj.getUses().size() == 0) // Defined by operation, but used nowhere
+      continue;
     const auto frontend_layout = [&]() {
       auto use_op_ind = *obj.getUses().begin(); // FIXME What if it has two or more uses?
       for (auto &operation_info : operation_list())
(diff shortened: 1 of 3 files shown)


hseok-oh:
@wateret PTAL

runtime/onert/backend/cpu/BackendContext.cc
```diff
@@ -62,6 +62,8 @@ ITensorRegistry *BackendContext::genTensors(const std::vector<onert::ir::OpSeque
     if (model_io.contains(index))
       continue;
     const auto &obj = graph()->operands().at(index);
+    if (obj.getUses().size() == 0) // Defined by operation, but used nowhere
+      continue;
```
wateret, line 66:
```suggestion
```

runtime/onert/backend/cpu/BackendContext.cc
```diff
@@ -62,6 +62,8 @@ ITensorRegistry *BackendContext::genTensors(const std::vector<onert::ir::OpSeque
...
     const auto frontend_layout = [&]() {
       auto use_op_ind = *obj.getUses().begin(); // FIXME What if it has two or more uses?
```
wateret, line 68:
```suggestion
      if (obj.getUses().size() == 0)
        return ir::Layout::UNKNOWN;
      auto use_op_ind = *obj.getUses().begin(); // FIXME What if it has two or more uses?
```

wateret (changes requested):
- (Suggestion) What about my code suggestion?
    - I would like it to be registered in this function(rather than the fallback registration in `planTensors`) so moving the getUses size check code
- Please apply these to other cpu-based backends as well

hseok-oh:
@wateret I've updated. PTAL

wateret (approved):
LGTM

PR #5275, opened 2020-12-03
[Draft][luci/pass]Introduce RemoveRedundantReshape on luci/pass

struss:
Parent Issue : #5266 
#
This commit introduce `RemoveRedundantReshape` on luci/pass

ONE-DCO-1.0-Signed-off-by: KiDeuk Bang <rrstrous@nate.com>

Issue #5266, opened 2020-12-03
[Compiler][luci] Introduce new optimization pass, RemoveRedundantReshape.


struss:
Parent Issue : #4353
#
### What to do in this Pass
Remove Redundant Reshape operations into one.
#### Input Type
Redundant Reshape operations.
<img src=https://user-images.githubusercontent.com/24720192/93728716-0e643500-fbfc-11ea-9f27-e44c337ee51c.png width='150' />

### TODO
- [x] Introduce luci/Pass for above optimization. ~#5284~ #5745 
- [x] Add Tensorflow/Res #5800 
- [x] Introduce this optimization on circle2circle. #5804
- [x] Introduce option on one-cmd #5828
- [x] Add `luci-pass-value-test` #5813
- [x] Add `circle2circle-dredd-recipe-test` #5810 

### Reference
https://github.com/tensorflow/tensorflow/blob/30ffc22c792d660042db05f8b0e9244b95ec3d55/tensorflow/core/grappler/optimizers/arithmetic_optimizer.cc#L1929-L1995
A little bit confused but this is optimization in tensorflow. What is Preserving Set?
Also, referenced in `Remove no-effect Reshape`

### Draft
~#5275~ #5793

struss:
I think rename this issue as `BypassRedundantReshape`, as former `Reshape` node will be removed on `RemoveDeadNodeWithQueryPass` if former `Reshape` has only one successor.

struss:
I think previous Draft closed and I force-push on that branch by mistake. so, I will create new Draft PR for this.

struss:
Done! Thanks for all!

PR #5284, opened 2020-12-04
[luci/pass]Introduce RemoveRedundantReshapePass on luci/pass.

struss:
Parent Issue : #5266
Draft : #5275 
#
This commit introduce `RemoveRedundantReshapePass` on luci/pass.

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/RemoveRedundantReshapePass.h
+++ b/compiler/luci/pass/include/luci/Pass/RemoveRedundantReshapePass.h
@@ -23,7 +23,7 @@ namespace luci
 {
 
 /**
- * @brief  Class to Redundant Reshape node into 1 reshape node.
+ * @brief  Class to remove redundant Reshape node into 1 reshape node.
  */
 struct RemoveRedundantReshapePass final : public logo::Pass
 {
--- a/compiler/luci/pass/include/luci/Pass/RemoveRedundantReshapePass.h
+++ b/compiler/luci/pass/include/luci/Pass/RemoveRedundantReshapePass.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_REMOVE_REDUNDANT_RESHAPE_PASS_H__
+#define __LUCI_REMOVE_REDUNDANT_RESHAPE_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to Redundant Reshape node into 1 reshape node.
+ */
+struct RemoveRedundantReshapePass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::RemoveRedundantReshapePass"; }
+
+  bool run(loco::Graph *g) final;
+};
+
+} // namespace luci
+
+#endif // __LUCI_REMOVE_REDUNDANT_RESHAPE_PASS_H__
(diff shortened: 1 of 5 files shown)


compiler/luci/pass/include/luci/Pass/RemoveRedundantReshapePass.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef __LUCI_REMOVE_REDUNDANT_RESHAPE_PASS_H__
+#define __LUCI_REMOVE_REDUNDANT_RESHAPE_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to Redundant Reshape node into 1 reshape node.
```
llFreetimell, line 26:
```suggestion
 * @brief  Class to remove redundant Reshape node into 1 reshape node.
```

compiler/luci/pass/src/RemoveRedundantReshapePass.cpp
```diff
@@ -0,0 +1,66 @@
...
+bool remove_redundant_reshape(luci::CircleNode *node)
+{
+  auto target_node = dynamic_cast<luci::CircleReshape *>(node);
+  if (target_node == nullptr)
+    return false;
+  auto pred_node = dynamic_cast<luci::CircleReshape *>(target_node->tensor());
+  if (pred_node == nullptr)
+    return false;
+  auto value_node = loco::must_cast<luci::CircleNode *>(pred_node->tensor());
+  target_node->tensor(value_node);
```
llFreetimell, line 33:
```suggestion
  target_node->tensor(pred_node->tensor());
```

seanshpark, line 33: This seems to disconnect `pred_node` without any conditions. Can you please explain about how this is possible?

compiler/luci/pass/include/luci/Pass/RemoveRedundantReshapePass.h
struss: Oh, I will fix it.

compiler/luci/pass/src/RemoveRedundantReshapePass.cpp
struss: Oh, I will fix it.

struss: It check subsequent `Reshape` Operation and check `pred_node` doesn't have control inputs. I think I forget to check `pred_node` has control inputs. I will check this and add implementation. sorry for missing.

struss: I find some information about `Reshape` in TFlite kernel. on this [operation](https://github.com/tensorflow/tensorflow/blob/fefcd5704ba4efe873485b981644ce0fb18c2440/tensorflow/lite/kernels/reshape.cc#L47-L50) number of input elements and number of output elements should be same. so, I think this check only reasonable on tensorflow grappler, I think.

seanshpark: You mean, current code does not need additional condition checks? 
Can you add some comments about how `nodeA`-`reshapeB`-`reshapeC` can be just changed to `reshapeC` with an example?
For example, let's say output of `nodeA` is `(4,32)` and `reshapeB` make this `(2,2,32)` and then `reshapeC` makes this `(2,2,2,16)`. overall `(4,32)` willbe converted to `(2,2,2,16)` so `reshapeB` canbe removed.

Another question, with the new network, does this run in our runtime and tflite without any problems?

struss: > Can you add some comments about how `nodeA`-`reshapeB`-`reshapeC` can be just changed to `reshapeC` with an example?

A little bit different. fallow from reference, this will bypass `reshapeB` if another reshape(`reshapeC`) get input as `reshapeB` output.( `NodeA` goes to `ReshapeC` directly)
So, final result will be like below.

image

After changing like that, ReshapeB would be removed if it have no output connection with other active node.

> For example, let's say output of `nodeA` is `(4,32)` and `reshapeB` make this `(2,2,32)` and then `reshapeC` makes this `(2,2,2,16)`. overall `(4,32)` will be converted to `(2,2,2,16)` so `reshapeB` can be removed.

Yes. but, `ReshapeB` will be remain if another node get this as `input Tensor`. 
Something Worried about is the case that reshape node's second input will be dynamic case...

I think only accepting this with both `ReshapeB` and `ReshapeC` have Second input with type `CircleConst`.

> Another question, with the new network, does this run in our runtime and tflite without any problems?

I will check about this more.
Sorry for bothering.

seanshpark:
As we talked a little about this, it would be better to add Passes for real models we encounter.
I think this Pass will benefit the performance but not sure about the conditions and actually this case exist.
If we meet any model that have this, then let's look at this change again.

struss:
Before encountering this case, I will close this issue. Thanks for all!

struss:
To handle [this](https://github.com/Samsung/ONE/issues/5401#issuecomment-748886742) on #5401 , this case also need to be used. So, I re-open this.

PR #5296, opened 2020-12-04
[onert] Make EventWriter work with multiple TracingObservers

hyunsik-yoon:
Parent issue: #4901
Draft: #4903

This makes `EventWriter` work with multiple `TracingObserver`s.

Background:
- One `TracingObserver` is allocated to one executor.
- When a model contains multiple subgraphs, each subgraph is run by
  its executor.
- When multiple `TracingObserver`s are destroyed
  profiling file should be written after destroying the last `TracingObject`.

Note:
- `EventWriter` is implemented as a singleton but guarded with mutex.
- To fully profile a model with multiple subgraphs, more PRs should be pushed to write subgraph index into profiling data.


wateret (approved):
LGTM

PR #5306, opened 2020-12-07
[onert] ExpandDims allows int64 type axis value

hseok-oh:
- Allow int64 axis input
- Shape inference for int64 axis
- Add 2 positive and 2 negative one op test for ExpandDims
- Remove needless axis tensor configuration in kernel generator


seanshpark (approved):
LGTM

PR #5320, opened 2020-12-08
[onert] Split EventWriter.cc into 4 files

hyunsik-yoon:
Split `EventWriter.cc` into 4 files since this file actually contains 4 classes and helper functions for each.

While working on issue: #4901, Draft: #4903

`#include "..."` statements were modified for each cc file
and the rest of code is same as previous `EventWriter.cc`.


wateret (approved):
LGTM

PR #5321, opened 2020-12-08
[onert] Fix `IOTensor::is_dynamic`

wateret:
Without checking `_orig_info` it was not able to reflect the result
from Static Shape Inference. This commit fixes it.

### DIFF
--- a/runtime/onert/core/src/backend/controlflow/IOTensor.h
+++ b/runtime/onert/core/src/backend/controlflow/IOTensor.h
@@ -62,7 +62,10 @@ public:
   ir::DataType data_type() const override { return _tensor->data_type(); }
   float data_scale() const override { return _tensor->data_scale(); }
   int32_t data_offset() const override { return _tensor->data_offset(); }
-  bool is_dynamic() const override { return _is_dynamic || (_tensor && _tensor->is_dynamic()); }
+  bool is_dynamic() const override
+  {
+    return _is_dynamic || _orig_info.isDynamic() || (_tensor && _tensor->is_dynamic());
+  }
   void set_dynamic() override { _is_dynamic = true; }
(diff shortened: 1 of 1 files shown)


hseok-oh:

hyunsik-yoon:
I have a question (beyond the scope of this PR)

Checking dynamic state of `IOTensor` seems complicated. Especially I wonder what `IOTensor._is_dynamic` means.
That may mean the real state is in other tensor since `IOTensor` indirects other tensor... and could we remove the var `IOTensor._is_dynamic`?

Most methods finds the real state from other tensor. Some examples are
```c++
  ir::Layout layout() const override { return _tensor->layout(); }
  ir::DataType data_type() const override { return _tensor->data_type(); }
```
However, in case of `_is_dynamic`, it maintains its own state and I am not sure the meaning of this variable.. and wonder if we can remove this variable.

Looking into `IOTensor` again, it maintains three state:
```
  const ir::OperandInfo _orig_info;
  const ir::Layout _orig_layout;
  bool _is_dynamic{false};
```
I also wonder if we could use `_tensor->info()`, `_tensor->layout()` ?

wateret:
@hyunsik-yoon It is kinda complicated now. That's because when I first introduced it, I directly translated the code to keep the original semantics. And I didn't try to refine the implementation. Sorry about the mess, I will work on it later.

I'll check if these are feasible and if yes, I'll refine the implementation.

- Merging `IOTensor::_is_dynamic` and `_orig_info::_is_dynamic`
- Merging `_orig_info::_is_dynamic` and `_tensor->is_dynamic()`

hyunsik-yoon (approved):
Adding new change (checking `_orig_info`) seems OK. LGTM

PR #5329, opened 2020-12-09
[onert] Fix when subgraph shapes are not updated

wateret:
For `IOTensor`, it must update `_info` as well.
This is actually a workaround.


hyunsik-yoon (approved):
LGTM

PR #5332, opened 2020-12-09
[luci] Fix comment for Reshape 'shape' input

seanshpark:
This will fix comment for Reshape 'shape' input


mhs4670go (approved):
LGTM

PR #5337, opened 2020-12-10
[onert] Unify genTensors impl of cpu backends

wateret:
Unify genTensors implementation of cpu-based backends.
They all use the same logic.


periannath (approved):
LGTM

PR #5342, opened 2020-12-10
[one-cmds] add command option to one-codegen

mhs4670go:
This commit has two changes.
- Don't make a log file
`one-codegen` is a driver that runs specific backend driver given from user. Other `one-cmds` driver generates the log file whose name is from its output file. But we can't determine the name of log file in case of `one-codegen` because we can't know the output file name in advance. Also, since we generates a log file for testing, we don't need the log file here because `backend` test should be tested at `backend` repo:)

- Add command option to configuration file
We can't know what the command is in advance because there are many different backend. So, `command` option will be written by its user.

### DIFF
--- a/compiler/one-cmds/one-codegen
+++ b/compiler/one-cmds/one-codegen
@@ -87,24 +87,19 @@ def main():
     # verify arguments
     _verify_arg(parser, args)
 
-    # get file path to log
+    # make a command to run given backend driver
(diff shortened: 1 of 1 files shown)


llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

chunseoklee (approved):

PR #5346, opened 2020-12-10
[onert] Remove unnamed namespaces in headers

wateret:
Unnamed namespaces in headers are almost always bad since:

> Due to default internal linkage, each translation unit will define its
  own unique instance of members of the unnamed namespace that are
  ODR-used within that translation unit.
> This can cause unexpected results, bloat the resulting executable, or
  inadvertently trigger undefined behavior due to one-definition rule
  (ODR) violations.

See also:

https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL59-CPP.+Do+not+define+an+unnamed+namespace+in+a+header+file
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf21-dont-use-an-unnamed-anonymous-namespace-in-a-header

### DIFF
--- a/runtime/onert/backend/acl_common/AclLinearMemoryManager.h
+++ b/runtime/onert/backend/acl_common/AclLinearMemoryManager.h
@@ -23,7 +23,11 @@
 #include "ir/OperandIndexMap.h"
 #include "util/logging.h"
 
-namespace
+namespace onert
+{
+namespace backend
+{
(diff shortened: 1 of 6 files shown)


hseok-oh (approved):
LGTM

periannath (approved):
LGTM

PR #5348, opened 2020-12-11
[tflchef] Add BidirectionalSequenceLSTM Operation

dr-venkman:
This commit add BidirectionalSequenceLSTM operation on TFlchef.

ONE-DCO-1.0-Signed-off-by: venkat.iyer venkat.iyer@samsung.com


mhs4670go (dismissed):
LGTM

compiler/tflchef/proto/tflchef.proto
```diff
@@ -110,6 +110,15 @@ enum MirrorPadMode {
   SYMMETRIC = 1;
 }
 
+message BidirectionalSequenceLSTMOptions {
+  optional Activation activation = 1 [default = NONE];
+  optional float cell_clip = 2 [default = 0.0];
+  optional float proj_clip = 3 [default = 0.0];
+  optional bool time_major = 4 [default = false];
+  optional bool asymmetric_quantize_inputs = 5 [default = false];
+  optional bool merge_outputs = 6 [default = false];
```
struss, line 119: Is it okay that the order of option is not matching with TFlite `schema.fbs` order?
and Default value is little bit different on `time_major`
(this comment referenced from [here](https://github.com/tensorflow/tensorflow/blob/f3c5b4ea29b1bd536bd70fc332cc1ccdf1b09695/tensorflow/lite/schema/schema.fbs#L637-L654))

mhs4670go: @dr-venkman Seems technically okay. But we might as well follow the order of tflite schema:)

dr-venkman: @struss, @mhs4670go , thanks for the comment. I will be correcting the order in the next commit.

mhs4670go (approved):
LGTM

struss (approved):
LGTM! Thanks!

PR #5395, opened 2020-12-17
[mir] Update clang-format version

hseok-oh:
Use clang-format-8 style for mir


seanshpark:
@binarman , I don't exactly remember that `mir` and related projects are following the format check. But if it's OK with you, we can merge this. If SRR want's to go with own format then it is OK not to land this for me.

binarman (approved):
LGTM
Thank you!

binarman:
@seanshpark 
> I don't exactly remember that mir and related projects are following the format check. But if it's OK with you, we can merge this. If SRR want's to go with own format then it is also OK not to land this for me.

Originally mir user it's own code style, but I think it is ok to unify it with main repo now.
(AFAIK there are no automatic check that could break because of this)

a-zh (approved):
LGTM

ai-moiseev (approved):
LGTM

PR #5363, opened 2020-12-14
[onert] Remove ITensor methods

wateret:
Remove `num_dimensions` and `dimension` as they are a duplication
of `getShape`.

ONE-DCO-1.0-Signed-off-by: Hanjoung Lee <hanjoung.lee@samsung.com>

PR #5400, opened 2020-12-17
[onert] Remove interp::ITensor methods (dups)

wateret:
Remove  interp::ITensor method `num_dimensions` and `dimension` as
those are duplications of `getShape`.

% In the same context with #5363


hseok-oh (approved):
LGTM

PR #5406, opened 2020-12-17
[onert] Independent cpu_common ConstantInitializer

wateret:
Make cpu_common::ConstantInitializer not derive from
`ConstantInitializerBase` which will be removed once
doing the same for `acl_common::ConstantInitializer`.


hseok-oh (approved):
LGTM

Issue #5375, opened 2020-12-15
Compiler FE: add one-build test


mhs4670go:
## What
Let's add more `one-build` test

## Why
There are `one-build` test but it is too insufficient to cover bunch of usecases.

There are seven `one-cmds` drivers and it should be test with variety of combination.

e.g.
one-import-tf -> one-codegen
one-import-tflite -> one-optimize -> one-codgen
one-import-tf -> one-optimize -> one-quantize -> one-codegen
one-import-tf -> one-optimize -> one-quantize -> one-pack

mhs4670go:
@seanshpark 
From https://github.com/Samsung/ONE/pull/5436#issuecomment-749995953, is it okay to revise one tests like below?

- Remove dependency between tests : Some test(e.g. `one-quantize_001.test`) doesn't create test material from scratch, which might occur problem sometimes.
- Create working directory : this is a bit similar with the second one. [create `working_dir` -> pushd `working_dir` -> test -> popd `working_dir`] process will make the tests more clear(?) and repeatable. Test output will be just a log file.

seanshpark:
> Remove dependency between tests : Some test(e.g. one-quantize_001.test) doesn't create test material from scratch, which might occur problem sometimes.

@jinevening , please give your opinion

seanshpark:
> Create working directory : this is a bit similar with the second one. [create working_dir -> pushd working_dir -> test -> popd working_dir] process will make the tests more clear(?) and repeatable. Test output will be just a log file.

Seems OK, but please check when jump to error handler.

jinevening:
> Remove dependency between tests : Some test(e.g. one-quantize_001.test) doesn't create test material from scratch, which might occur problem sometimes.

> @jinevening , please give your opinion
It is OK with me if you revise to use commands, not tests.

I agree with you. It would be better to call `one-import` command instead of running `one-import_001.test`.

mhs4670go:
> using commands is like "we assume this command has no problem".

I will download all the required material in `prepare_test_materials` steps. e.g., `.tflite`

jinevening:
> I will download all the required material in prepare_test_materials steps rather than running other commands. e.g., .tflite

This seems to be a perfect solution if network/storage resources for one-build tests are not too limited.

PR #5436, opened 2020-12-18
[one-cmds] Add one-build_004 test

mhs4670go:
This commit adds one-build_004 test. This test runs `one-import-tf` and
`one-codegen` at once with `one-build`.

Related: #5375 

### DIFF
--- a/compiler/one-cmds/dummy-driver/src/dummy-compile.cpp
+++ b/compiler/one-cmds/dummy-driver/src/dummy-compile.cpp
@@ -14,7 +14,13 @@
  * limitations under the License.
  */
 
-// dummy-compile only tests its interface rather than its functionality.
+/**
+ * dummy-compile only tests its interface rather than its functionality.
+ *
+ * ./dummy-compile -o ${OUTPUT_NAME} ${INPUT_NAME}
+ *
+ * NOTE argv[3](INPUT_NAME) is not used here.
+ */
 
 #include <iostream>
 #include <fstream>
@@ -22,13 +28,12 @@
 
 int main(int argc, char **argv)
 {
-  // Assume cmd from user will be like this.
-  // ./dummy-compile -o ${OUTPUT_NAME} ${INPUT_NAME}
   if (argc != 4)
     return EXIT_FAILURE;
 
   std::string opt_o{"-o"};
   std::string argv_1{argv[1]};
+
   if (opt_o != argv_1)
     return EXIT_FAILURE;
 
--- a/compiler/one-cmds/dummy-driver/src/dummy-compile.cpp
+++ b/compiler/one-cmds/dummy-driver/src/dummy-compile.cpp
@@ -28,12 +28,12 @@ int main(int argc, char **argv)
     return EXIT_FAILURE;
 
   std::string opt_o{"-o"};
-  std::string output_name{argv[1]};
-  if (opt_o != output_name)
+  std::string argv_1{argv[1]};
+  if (opt_o != argv_1)
     return EXIT_FAILURE;
 
-  std::string input_name{argv[2]};
-  std::ofstream outfile(input_name);
+  std::string output_name{argv[2]};
+  std::ofstream outfile(output_name);
 
   outfile << "dummy-compile dummy output!!" << std::endl;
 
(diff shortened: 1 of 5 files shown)


mhs4670go:
@seanshpark In our offline talk, we've decided to install `dummy-codegen` to `bin` folder and exclude it when packaging. But actually on second thoguht, when we just untar `nncc-1.12.0.tar.gz
` and `nncc-test-x.x.x.tar.gz` rather than building it, it would be a problem as well because there would be no `dummy-codegen` in `bin` file.

Speaking of which, how about revising this `one-cmds` test. The list of points to revise are like this.
- Integrate `one-cmds` test to Cmake ctest : it'd solve above issue and we'd can run `one-cmds` test in CI as well.
- Remove dependency between tests : Some test(e.g. `one-quantize_001.test`) doesn't create test material from scratch, which might occur problem sometimes.
- Create working directory : this is a bit similar with the second one. [create `working_dir` -> pushd `working_dir` -> test -> popd `working_dir`] process will make the tests more clear(?) and repeatable. Test output will be just a log file.

It is just my suggestion:) Please feel free to give me your opinion.

seanshpark:
> it would be a problem as well because there would be no dummy-codegen in bin file.

I was thinking to put `dummy-codegen` in `nncc-test-x.x.x.tar.gz`

> Integrate one-cmds test to Cmake ctest

Purpose of this test is with release package. I'm not sure about this with Cmake test

mhs4670go:
@seanshpark 
> Purpose of this test is with release package

got it.

> I was thinking to put `dummy-codegen` in `nncc-test-x.x.x.tar.gz`

Then, I will install `dummy-codegen` to `test` folder. If so, there is no need to exclude it from packaging like `tflchef` does. And, I need to copy `dummy-codegen` to `bin` folder before `one-codegen` test and remove it at the end of the test.

mhs4670go:
@seanshpark PTAL:)

compiler/one-cmds/dummy-driver/src/dummy-compile.cpp
```diff
@@ -0,0 +1,43 @@
...
+  // ./dummy-compile -o ${OUTPUT_NAME} ${INPUT_NAME}
+  if (argc != 4)
+    return EXIT_FAILURE;
+
+  std::string opt_o{"-o"};
+  std::string output_name{argv[1]};
+  if (opt_o != output_name)
+    return EXIT_FAILURE;
+
+  std::string input_name{argv[2]};
```
seanshpark, line 35: please fix this variable name to `output_name`

compiler/one-cmds/dummy-driver/src/dummy-compile.cpp
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * dummy-compile only tests its interface rather than its functionality.
+ *
+ * ./dummy-compile -o ${OUTPUT_NAME} ${INPUT_NAME}
+ *
+ * NOTE argv[3](INPUT_NAME) not used as it is a dummy param.
```
seanshpark, line 22: Maybe `not used as this is a dummy program.`

mhs4670go: @seanshpark Actually, it's kinda funny,, it(argv[3]) is a dummy param("argument" might be more accurate) of dummy program..

seanshpark: Hm... maybe `NOTE argv[3](INPUT_NAME) is not used here`

compiler/one-cmds/tests/one-build_004.test
```diff
@@ -0,0 +1,47 @@
...
+
+trap trap_err_onexit ERR
+
+configfile="one-build_004.cfg"
+outputfile="sample.tvn"
+
+rm -rf ${outputfile}
+
+# copy dummy-compile to bin folder
+cp dummy-compile ../bin/dummy-compile
```
seanshpark, line 36: after this, `../bin/dummy-compile` will exist and if something goes wrong, `../bin/dummy-compile` won't be removed.
maybe add a command to silently remove this file in `trap_err_onexit` ?

seanshpark (approved):
LGTM thank you!

Issue #5401, opened 2020-12-17
Remove redundant TRANSPOSE Ops generated by onnx-tf


jinevening:
## What

Remove redundant TRANSPOSE Ops generated by onnx-tf

## Why

We're going to support ONNX model in ONE via the following steps.

ONNX -> TF (.pb) -> TFLite (.tflite) -> Circle (.circle)

[onnx-tensorflow](https://github.com/onnx/onnx-tensorflow) will be used to convert ONNX file to TF (.pb) file. It generates many Transpose Ops to convert the data layout from NCHW (used by ONNX) to NHWC (used by TF). These Transpose Ops are indeed meaningless in terms of DNN behavior, but introduce a huge overhead.

## How

To remove these Transpose Ops, we'll change NCHW Ops to NHWC Ops and use [RemoveRedundantTransposePass](https://github.com/Samsung/ONE/pull/5172). For example,

Before: NCHW tensor -> **Op (NCHW)** -> NCHW tensor

After: NCHW tensor -> **Transpose (perm:[0,2,3,1]) -> NHWC tensor -> Op (NHWC) -> NHWC tensor -> Transpose (perm:[0,3,1,2])** -> NCHW tensor

`onnx-tensorflow` does this transformation only for special Ops (e.g., Conv, TConv, etc). We're going to do this transformation for all Ops to make `RemoveRedundantTransposePass` remove redundant Transpose Ops.

For example, let's assume the above transformation is done for two subsequent Ops

NCHW tensor -> Transpose (perm:[0,2,3,1]) -> NHWC tensor -> Op (changed to use NHWC) -> NHWC tensor -> **Transpose (perm:[0,3,1,2]) -> NCHW tensor -> Transpose (perm:[0,2,3,1])** -> NHWC tensor -> Op (changed to use NHWC) -> NHWC tensor -> Transpose (perm:[0,3,1,2]) -> NCHW tensor

The bold Transpose Ops can be removed by `RemoveRedundantTransposePass` as they are redundant.

For this, we need a pass (`ConvertNCHWToNHWCPass`) that updates NCHW Ops to NHWC Ops.

## Status

- [x] Convert NCHW to NHWC #5438 @jinevening 
- [x] Remove redundant transpose #5437 @struss 
- [x] Remove redundant split #5456 @struss 
- [x] Remove redundant slice #5416 @struss 
- [x] Convert Transpose to Reshape #5703 @struss 
- [x] Remove redundant Reshape #5275 @struss

[...]

PR #5437, opened 2020-12-18
[luci/pass] Update RemoveRedundantTranspose cover more case.

struss:
For #5401 

This commit update Remove RedundantTranpose to cover more cases.


compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -40,11 +40,10 @@ bool remove_consecutive_transpose_function(luci::CircleNode *node)
   auto target_node = dynamic_cast<luci::CircleTranspose *>(node);
   if (target_node == nullptr)
     return false;
+
```
seanshpark, line 43: let's not add this style change

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -61,9 +60,10 @@ bool remove_consecutive_transpose_function(luci::CircleNode *node)
   }
   else
   {
-    auto g = main_perm->graph();
-    auto new_const_node = g->nodes()->create<luci::CircleConst>();
+    auto g = main_node->graph();
 
+    // Create New Const Node
```
seanshpark, line 65: I don't thing adding comments is aboutr `cover more case`. it would be better to post these changes in another PR.

struss: Okay I will remove this.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: Sorry. I will fix it.

jinevening:
I'm not sure this PR can cover the case in #5401 . 

Let me explain the algorithm. 

**Before** 
```
main -> target (Transpose) -> succ1 (Transpose) -> next1
                           +-> succ2 (Transpose) -> next2
                                ...
```

**After** 
If all successors satisfy `check_perm`,
(i.e., `if check_perm (succ1, target) == true && check_perm(succ2, target) == true ...`)
```
main -> next1
     +-> next2
        ...
```

I think you don't have to handle the case that creates a new Transpose for now (let's leave it as a TODO in the comments).

Plus, can you add a test case after implementation ?

struss:
> I'm not sure this PR can cover the case in #5401 .
> 
> Let me explain the algorithm.
> 
> **Before**
> 
> ```
> main -> target (Transpose) -> succ1 (Transpose) -> next1
>                            +-> succ2 (Transpose) -> next2
>                                 ...
> ```

on previous logic, It only activate on 
```
main -> pred_node (Transpose) -> target_node (Transpose) -> next1
```
with pred_node only goes on target. so, this will never activate on example cause pred_node have 2 succs.

> 
> **After**
> If all successors satisfy `check_perm`,
> (i.e., `if check_perm (succ1, target) == true && check_perm(succ2, target) == true ...`)
> 
> ```
> main -> next1
>      +-> next2
>         ...
> ```
> 
> I think you don't have to handle the case that creates a new Transpose for now (let's leave it as a TODO in the comments).

Current logic update `pred_node`'s perm value and replace `target_node` with `pred_node`.
If I just remove below contents,
```
  if (loco::succs(pred_node).size() != 1)
    return false;
``` 
pred_node with multiple case will result in false value cause first target transpose change `pred_node`'s perm value.

> Plus, can you add a test case after implementation ?

Okay I will add this test.

jinevening:
Oh, I see. The code removes the successors of pred_node one-by-one per iteration. Thanks.

jinevening (dismissed):
LGTM

jinevening:
> Okay I will add this test.

FYI, the test code may run this pass multiple times, so it would be better to use PhaseRunner.

Recently I've made a similar test case (https://github.com/Samsung/ONE/pull/5438/files).

struss:
> > Okay I will add this test.
> 
> FYI, the test code may run this pass multiple times, so it would be better to use PhaseRunner.
> 
> Recently I've made a similar test case (https://github.com/Samsung/ONE/pull/5438/files).

Oh I will reference this. Thanks!

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -154,3 +242,49 @@ TEST(RemoveRedundantTransposePass, remove_consecutive_transpose_function_type2)
...
+  luci::CircleTranspose *transpose_node = nullptr;
+  for (auto node : loco::active_nodes(loco::output_nodes(graph.get())))
+  {
+    auto trans = dynamic_cast<luci::CircleTranspose *>(node);
+    if (not trans)
+      continue;
+    transpose_node = trans;
+    break;
+  }
+  // Just one transpose node, with updated perm constant.
```
jinevening, line 283: If you want to check the number of transpose nodes in the graph, it would be better to fully iterate the for-loop (this code does not check the exact number of transpose nodes).

struss: Oh, I will fix it.

jinevening (dismissed):
LGTM

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -106,6 +115,85 @@ void create_redundunt_transpose(loco::Graph *g, const std::vector<int32_t> &perm
   output->index(graph_output->index());
 }
 
+void create_redundunt_transpose_with_branch(loco::Graph *g, const std::vector<int32_t> &perm1,
```
mhs4670go, line 118: Comment painting(?) of graph would be help. Or is it same with the above graph?

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -74,8 +72,14 @@ bool remove_consecutive_transpose_function(luci::CircleNode *node)
...
-    pred_node->perm(new_const_node);
-    replace(node).with(pred_node);
+
+    // Create New Transpose Node
+    auto new_transpose_node = g->nodes()->create<luci::CircleTranspose>();
+    new_transpose_node->dtype(target_node->dtype());
+    new_transpose_node->a(main_node);
+    new_transpose_node->perm(new_const_node);
+
+    replace(node).with(new_transpose_node);
```
mhs4670go, line 82: @struss Could you explain about this change more detail?
As below comment,
```cxx
*  AFTER
 *      <Optional Case>
 *
 *          |                 |                   |
 *    [CircleNode]      [CircleConst]             |
 *     (main_node)     (new_const_node)           |
 *           \               /           or  [CircleNode]
 *           [CircleTranspose]                (main_node)
 *              (pred_node)                       |
 *                   |                            |
 *
```
This seems to change the way it replace the `pred_node` with `target_node` from just replacement to creation and replacement. Is there any reason of this change?

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: Oh, I will add this case more detail. Thanks for pointing out.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: > @struss Could you explain about this change more detail?
> As below comment,
> 
> ```c++
> *  AFTER
>  *      <Optional Case>
>  *
>  *          |                 |                   |
>  *    [CircleNode]      [CircleConst]             |
>  *     (main_node)     (new_const_node)           |
>  *           \               /           or  [CircleNode]
>  *           [CircleTranspose]                (main_node)
>  *              (pred_node)                       |
>  *                   |                            |
>  *
> ```

Oh, I think I need to change this.

> This seems to change the way it replace the `pred_node` with `target_node` from just replacement to creation and replacement.

that's right.

>Is there any reason of this change?

To handle case that `pred_node` have multiple succs.
previous implementation only activate when `pred_node` has one succ and also check in only one branch.
but, if `pred_node` have multiple succs, it also could be shorten. so, to handle this case, I change logic.
I will update all comment that explain this logic.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -105,6 +114,113 @@ void create_redundunt_transpose(loco::Graph *g, const std::vector<int32_t> &perm
...
+ *          |                |
+ *     [CircleNode]     [CircleNode]
+ *
+ *  AFTER
+ * Type 1
+ *            |
+ *       [CircleNode]
+ *           /  \       Remove all transpose
+ * [CircleNode][CircleNode]
+ * Type2
```
mhs4670go, line 135:
```suggestion
 *  BEFORE
 *            |
 *      [CircleNode]       [CircleConst]
 *               \           /
 *[CircleConst] [CircleTranspose] [CircleConst]
 *        \          / \              /
 *  [CircleTranspose] [CircleTranspose]
 *          |                |
 *     [CircleNode]     [CircleNode]
 *
 *  AFTER
 * Type 1
 *            |
 *       [CircleNode]
 *           /  \       Remove all transpose
 * [CircleNode][CircleNode]
 * Type 2
```

mhs4670go (approved):
LGTM

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -61,9 +59,9 @@ bool remove_consecutive_transpose_function(luci::CircleNode *node)
   }
   else
   {
-    auto g = main_perm->graph();
-    auto new_const_node = g->nodes()->create<luci::CircleConst>();
+    auto g = main_node->graph();
```
seanshpark, line 62: This is change necessary?

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -61,9 +59,9 @@ bool remove_consecutive_transpose_function(luci::CircleNode *node)
...
 
+    auto new_const_node = g->nodes()->create<luci::CircleConst>();
```
seanshpark, line 64: This change seems to be a format change. It would be better to separate this.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -117,15 +232,17 @@ TEST(RemoveRedundantTransposePass, remove_consecutive_transpose_function_type1)
...
   {
     auto trans = dynamic_cast<luci::CircleTranspose *>(node);
     if (not trans)
       continue;
     transpose_node = trans;
-    break;
+    count++;
   }
   // No transpose node is in graph.
+  ASSERT_EQ(0, count);
```
seanshpark, line 245: I think this is not needed for `no transpose node`

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -138,15 +255,67 @@ TEST(RemoveRedundantTransposePass, remove_consecutive_transpose_function_type2)
...
+  for (auto node : loco::active_nodes(loco::output_nodes(graph.get())))
+  {
+    auto trans = dynamic_cast<luci::CircleTranspose *>(node);
+    if (not trans)
+      continue;
+    transpose_node = trans;
+    count++;
+  }
+  // Just one transpose node, with updated perm constant.
+  ASSERT_EQ(1, count);
```
seanshpark, line 268: If we need to check the number of `CircleTranspose` then checking number of `count` is needed.
But here we only check the value is 1. Then I think `ASSERT_NE(nullptr, transpose_node);` is enough.
Is there any other reason for adding `count` that I may not thought of?

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: Oh I will remove this.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: Oh, yes. Just one op, so I also think remove this would be better. thanks.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: oh, I will make as original.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: okay. I make a roll-back to original.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -105,6 +114,112 @@ void create_redundunt_transpose(loco::Graph *g, const std::vector<int32_t> &perm
   auto graph_output = g->outputs()->create();
   output->index(graph_output->index());
 }
+/**
+ *  BEFORE
+ *            |
+ *      [CircleNode]       [CircleConst]
+ *               \           /
+ *[CircleConst] [CircleTranspose] [CircleConst]
```
seanshpark, line 122: Can you add some spaces between `*` and `[CircleConst]` ?

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -105,6 +114,112 @@ void create_redundunt_transpose(loco::Graph *g, const std::vector<int32_t> &perm
...
+ *        \          / \              /
+ *  [CircleTranspose] [CircleTranspose]
+ *          |                |
+ *     [CircleNode]     [CircleNode]
+ *
+ *  AFTER
+ * Type 1
```
seanshpark, line 129: How about change the indentation?
```
 *  AFTER
 *   Type 1
```

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -105,6 +114,112 @@ void create_redundunt_transpose(loco::Graph *g, const std::vector<int32_t> &perm
...
+ *            |
+ *       [CircleNode]
+ *           /  \       Remove all transpose
+ * [CircleNode][CircleNode]
+ * Type 2
```
seanshpark, line 134: Please add an empty line between ` * [CircleNode][CircleNode]` and ` * Type 2`

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -105,6 +114,112 @@ void create_redundunt_transpose(loco::Graph *g, const std::vector<int32_t> &perm
...
+ *                |                 |
+ *          [CircleNode]      [CircleConst]
+ *           (main_node)     (new_const_node)
+ *               / \               /
+ *    [CircleNode] [CircleTranspose]
+ *          |       (new_trans_node)
+ *                         |
+ *                    [CircleNode]
+ *                         |
```
seanshpark, line 143: Let's remove `|` after `[CircleNode]` like the others

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -105,6 +114,112 @@ void create_redundunt_transpose(loco::Graph *g, const std::vector<int32_t> &perm
...
+void run_phase(loco::Graph *g)
+{
+  logo::Phase phase;
+
+  // Default passes.
+  phase.emplace_back(std::make_unique<logo::RemoveDeadNodeWithQueryPass>());
+  phase.emplace_back(std::make_unique<luci::MigrateLegacyShapeDtypePass>());
+  phase.emplace_back(std::make_unique<luci::TypeInferencePass>());
+  phase.emplace_back(std::make_unique<luci::ShapeInferencePass>());
+  phase.emplace_back(std::make_unique<luci::ShapeSignatureInferencePass>());
```
seanshpark, line 215: If above Passes are not mandatory, we can remove them: all or some.
Especially adding `MigrateLegacyShapeDtypePass` may trigger unnecessary future changes.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -123,7 +238,6 @@ TEST(RemoveRedundantTransposePass, remove_consecutive_transpose_function_type1)
     if (not trans)
       continue;
     transpose_node = trans;
-    break;
```
seanshpark, line 126: Is this necessary?

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
seanshpark: Or add `|` to missing ones...

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: Okay I will fix it.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: Okay I will fix it.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: Okay I will fix it.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: Okay I will fix it.

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: okay I make a roll-back

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: oh, I will make as previous test.

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
```diff
@@ -100,13 +104,13 @@ namespace luci
...
- *           [CircleTranspose]                (main_node)
- *              (pred_node)                       |
- *                   |                            |
+ *                 |                 |                    |
+ *           [CircleNode]      [CircleConst]              |
+ *            (main_node)     (new_const_node)            |
+ *                 / \               /           or  [CircleNode]
+ * [CircleTranspose] [CircleTranspose]                (main_node)
+ *    (pred_node)    (new_trans_node)                     |
+ *                          |                             |
```
seanshpark, line 113: minor thing, `|` after `(pred_node)` is removed. is `(pred_node)` disconnected? or if not then let's add `|` after `(pred_node)`

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
```diff
@@ -105,6 +105,95 @@ void create_redundunt_transpose(loco::Graph *g, const std::vector<int32_t> &perm
...
+ *  BEFORE
+ *             |
+ *       [CircleNode]       [CircleConst]
+ *                 \           /
+ *  [CircleConst] [CircleTranspose] [CircleConst]
+ *          \          / \              /
+ *     [CircleTranspose] [CircleTranspose]
+ *            |                |
+ *       [CircleNode]     [CircleNode]
+ *
```
seanshpark, line 118: maybe add `|` after `[CircleNode]` items here and below

compiler/luci/pass/src/RemoveRedundantTranspose.cpp
struss: If `pred_node` have successor more than one(not `target_node` only), then it could connect to that node. otherwise, it will not  connected to any node. so, I couldn't draw any connection line.(just I think)

compiler/luci/pass/src/RemoveRedundantTranspose.test.cpp
struss: I removed as I commented :(
https://github.com/Samsung/ONE/pull/5437#discussion_r549270909

seanshpark: Sorry for the confision... I wanted to make things consistant... There was another comment https://github.com/Samsung/ONE/pull/5437#discussion_r549272751 :)

struss: then I  think remove all `|` under final node would be nice as it could be connect to any other node. what about this idea?

struss: After updating this, I will make a new PR for make a consistency on comment graph.

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

jinevening (approved):

PR #5443, opened 2020-12-21
[onert] Remove ConstantInitializer in cpu_common

wateret:
As cpu constant initialization is simple, we do not need
ConstantInitializer in cpu_common.


hseok-oh (approved):
LGTM

PR #5446, opened 2020-12-21
[onert] Support quantize operator (uint8 to int8)

glistening:
: OperationValidator allows u8 and i8 input and output type.
: add u8-to-i8 kernel
: QuantizeLayer is out of ElementwiseUnary.
: Quanitze CircleGen test is added.

### DIFF
--- a/runtime/onert/core/src/ir/OperationValidator.cc
+++ b/runtime/onert/core/src/ir/OperationValidator.cc
@@ -285,8 +285,10 @@ void OperationValidator::visit(const operation::ElementwiseUnary &node)
   }
   else if (node.param().op_type == operation::ElementwiseUnary::Type::QUANTIZE)
   {
-    OP_REQUIRES(isValidType(input_index, DataType::FLOAT32));
-    OP_REQUIRES(isValidType(output_index, DataType::QUANT_UINT8_ASYMM));
+    OP_REQUIRES(isValidType(
+      input_index, {DataType::FLOAT32, DataType::QUANT_UINT8_ASYMM, DataType::QUANT_INT8_ASYMM}));
+    OP_REQUIRES(
+      isValidType(output_index, {DataType::QUANT_UINT8_ASYMM, DataType::QUANT_INT8_ASYMM}));
   }
   else if (node.param().op_type == operation::ElementwiseUnary::Type::FLOOR)
   {
--- a/tests/nnfw_api/src/one_op_tests/Quantize.cc
+++ b/tests/nnfw_api/src/one_op_tests/Quantize.cc
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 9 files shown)


tests/nnfw_api/src/one_op_tests/Quantize.cc
```diff
@@ -0,0 +1,58 @@
...
+
+#include "GenModelTest.h"
+
+#include <memory>
+
+CircleGen genSimpleQuantizeModel(circle::TensorType from_t, circle::TensorType to_t)
+{
+  CircleGen cgen;
+  int in = cgen.addTensor({{1, 4, 4, 1}, from_t}, 1, 128);
+  int out = cgen.addTensor({{1, 4, 4, 1}, to_t}, 1, 0);
```
ragmani, line 25: How about using changed scale from input?

runtime/onert/core/src/ir/OperationValidator.cc
```diff
@@ -285,8 +285,10 @@ void OperationValidator::visit(const operation::ElementwiseUnary &node)
   }
   else if (node.param().op_type == operation::ElementwiseUnary::Type::QUANTIZE)
   {
-    OP_REQUIRES(isValidType(input_index, DataType::FLOAT32));
-    OP_REQUIRES(isValidType(output_index, DataType::QUANT_UINT8_ASYMM));
+    OP_REQUIRES(isValidType(
+      input_index, {DataType::FLOAT32, DataType::QUANT_UINT8_ASYMM, DataType::QUANT_INT8_ASYMM}));
```
hseok-oh, line 289: TFLite 2.3 allow `int16` input (`int16` -> `int16` or `int8`)

tests/nnfw_api/src/one_op_tests/Quantize.cc
ragmani: If output type is `INT8_ASYMM`, it would be nice to change zero_point to a non-zero number.

runtime/onert/core/src/ir/OperationValidator.cc
glistening: @hseok-oh Thank you for information. But `onert` does not support them yet. I allow the cases that `onert` can currently handle, not tflite's latest specification.

tests/nnfw_api/src/one_op_tests/Quantize.cc
glistening: @ragmani 

> How about using changed scale from input?

Could you explain more what you mean? I think the scale is same `1` after `Quantize` u8 to i8.

glistening: > If output type is `INT8_ASYMM`, it would be nice to change zero_point to a non-zero number.

@ragmani It would take more time since I made the test case by myself. Do you have any testcase that I could use for complex test case?

runtime/onert/core/src/ir/OperationValidator.cc
hseok-oh: I see. It's OK.

tests/nnfw_api/src/one_op_tests/Quantize.cc
ragmani: > Could you explain more what you mean? I think the scale is same 1 after Quantize u8 to i8.

The `Quantize` operation can be change scale value from input to output. So it would be nice to change scale from input to output to for more strict verification.

hseok-oh (dismissed):
LGTM

tests/nnfw_api/src/one_op_tests/Quantize.cc
glistening: > > Could you explain more what you mean? I think the scale is same 1 after Quantize u8 to i8.
> 
> The `Quantize` operation can change scale value from input to output. So it would be nice to change scale from input to output to for more strict verification.

Okay. You want to change output scale to different scale. It took me much more time but consequently it gets better test case.

hseok-oh (approved):

ragmani (approved):
LGTM

PR #5447, opened 2020-12-21
[onert] Remove cpu_common::ConstantInitializer

wateret:
Now that it is moved as a single function `initConsts` in
`BackendContextHelpers.h`, this is no longer used..


hseok-oh (approved):
LGTM

PR #5292, opened 2020-12-04
[onert] ostream overloads for index types

wateret:
Introduce ostream overloads for index types. Each types of indexes
has its own prefix.

Plus, this commit includes usage fixes.

ONE-DCO-1.0-Signed-off-by: Hanjoung Lee <hanjoung.lee@samsung.com>

## 5+ Reviews to merge

PR #5448, opened 2020-12-21
[onert] Change print symbols for ops

wateret:
Change print symbols for operations and operands.
This is LLVM(or MLIR) style.

### DIFF
--- a/runtime/onert/core/include/ir/Index.h
+++ b/runtime/onert/core/include/ir/Index.h
@@ -52,12 +52,12 @@ std::ostream &_index_print_impl(std::ostream &o, const std::string &prefix, Inde
 
 inline std::ostream &operator<<(std::ostream &o, const OperationIndex &i)
 {
-  return _index_print_impl(o, "OPERATION", i);
+  return _index_print_impl(o, "@", i);
 }
 
 inline std::ostream &operator<<(std::ostream &o, const OperandIndex &i)
 {
-  return _index_print_impl(o, "$", i);
+  return _index_print_impl(o, "%", i);
 }
 
 inline std::ostream &operator<<(std::ostream &o, const IOIndex &i)


wateret:
Related PR : #5292

LLVM and MLIR use prefix `@` for functions and `%` for variables. So I suggest doing the same for our repo.

LLVM example

```llvm
define double @bar(double %a) {
entry:
        %calltmp = call double @foo(double %a, double 4.000000e+00)
        %calltmp1 = call double @bar(double 3.133700e+04)
        %addtmp = fadd double %calltmp, %calltmp1
        ret double %addtmp
}
```

MLIR example

```mlir
func @transpose_transpose(%arg0: tensor<*xf64>) -> tensor<*xf64> {
  %0 = toy.transpose(%arg0 : tensor<*xf64>) to tensor<*xf64>
  %1 = toy.transpose(%0 : tensor<*xf64>) to tensor<*xf64>
  toy.return %1 : tensor<*xf64>
}
```

wateret:
Related issue : #5449

glistening:
@wateret 

`LLVM IR` function symbol `@` is followed by function name, and most names look like function names without further explanation. (e.g. `foo`, `bar`, `transpose_transpose`, ...) while our IR is just index. I will forget soon what is operation and operand. If someone are familiar with `LLVM IR` and remember the symbol's meaning, it would be good. However, it is not my case. Though I am not for this change, I don't object this change if others like this change.

wateret:
@glistening It would be also nice if you take a look at #5449 which has more context about this.

glistening:
> @glistening It would be also nice if you take a look at #5449 which has more context about this.

@wateret  I've already read that comment before you wrote this comment.
The issue (#5448) is talking about Graph dumper, however, the change of this PR will change all log printing.
Please let me know if I miss something. :)

wateret:
@glistening 

> `LLVM IR` function symbol `@` is followed by function _**name**_, and most names look like function names without further explanation.

I see this in a different point of view.

- In LLVM, for variables, they also suffix numbers to make them SSA form - they also use numbers
- In LLVM, most inputs(like C++ code) have names for all variables and functions so those are used naturally. However our input, CIRCLE, does not have names for operators, and names for tensors are optional. The reliable ID is their indices. So I think it makes sense to have numbers to represent operation

wateret:
> The issue (#5448) is talking about Graph dumper, however, the change of this PR will change all log printing.
> Please let me know if I miss something. :)

No you don't. I just wanted to say if we are going for #5448 , it would also make sense that reusing the symbols in other places.

glistening:
> @glistening
> 
> I see this in a different point of view.
> 
>     * In LLVM, for variables, they also suffix numbers to make them SSA form - they also use numbers

I didn't said `variables` use names. I talked about `function`.

>     * In LLVM, most inputs(like C++ code) have names for all variables and functions so those are used naturally. However our input, CIRCLE, does not have names for operators, and names for tensors are optional. The reliable ID is their indices. So I think it makes sense to have numbers to represent operation

It seems you misunderstood my point. I didn't want to suggest to use name than index in `onert`.
I am saying it is hard to distinguish operation and operand only with symbol (`@`, `%`) because both are followed by number.

wateret:
@glistening Gotcha.

wateret:
@hseok-oh @dayo09 @hseok-oh PTAL.

If there is one more opinion that is not favor of chaging Operation Index symbol, I'll revert it.

glistening:
> @hseok-oh @dayo09 @hseok-oh PTAL.

Twice @hseok-oh. Perhaps @periannath or @hyunsik-yoon

hseok-oh:
I don't mind changing print symbol or not because its user is developer and it's not difficult to understand for me.

periannath:
I'm agree with this change since `OperationIndex` is widely used in log messages.

periannath (approved):

dayo09 (approved):
I looks good for me, too :-D

wateret:
@glistening @dayo09 @hseok-oh @glistening

Thank you for your opinions! I'll merge it since I got two approvals with no objections.

PR #5452, opened 2020-12-21
[onert] Refine permuteShape implementation

wateret:
- Rename params
- Refine if and assert conditions


ragmani (approved):
LGTM

PR #5453, opened 2020-12-22
[gen_golden] Support int8 quantization random input

glistening:
gen_golden will generate [-127,127] random input for int8 input type.

### DIFF
--- a/tools/nnpackage_tool/gen_golden/gen_golden.py
+++ b/tools/nnpackage_tool/gen_golden/gen_golden.py
@@ -91,6 +91,9 @@ if __name__ == '__main__':
             if this_dtype == tf.uint8:
                 input_values.append(
                     np.random.randint(0, 255, this_shape).astype(np.uint8))
+            if this_dtype == tf.int8:
+                input_values.append(
+                    np.random.randint(-127, 127, this_shape).astype(np.int8))
             elif this_dtype == tf.float32:
                 input_values.append(
                     np.random.random_sample(this_shape).astype(np.float32))
@@ -134,6 +137,9 @@ if __name__ == '__main__':
             if this_dtype == np.uint8:
                 input_values.append(
                     np.random.randint(0, 255, this_shape).astype(np.uint8))
+            if this_dtype == np.int8:
+                input_values.append(
+                    np.random.randint(-127, 127, this_shape).astype(np.int8))
             elif this_dtype == np.float32:
                 input_values.append(
                     np.random.random_sample(this_shape).astype(np.float32))


hseok-oh (approved):

tools/nnpackage_tool/gen_golden/gen_golden.py
```diff
@@ -91,6 +91,9 @@ def usage():
             if this_dtype == tf.uint8:
                 input_values.append(
                     np.random.randint(0, 255, this_shape).astype(np.uint8))
+            if this_dtype == tf.int8:
+                input_values.append(
+                    np.random.randint(-127, 127, this_shape).astype(np.int8))
```
ragmani, line 96: Maybe -128?

glistening: > Maybe -128?

@ragmani No, we don't need to use the full range of int8 as we did in other type. Also, I considered tensorflow lite uses `[-127, 127]` for weight (symmetric) tensor.

ragmani: > i) we don't need to use the full range of int8 as we did in other type.

`ir::DataType::QUANT_INT8_ASYMM` has range as [-128, 127]. I think we support full range of int to verify this type.

> Too tight range may cause out-of-range outputs as it flows several operators.

Other values like `-127` can also cause out-of-range as it flows several operators. I think out-of-range of outputs is responsible for the kernel of operators. Most of kernels of tensorflow lite handle it.

> ii) Also, I've considered tensorflow lite uses [-127, 127] for weight (symmetric) tensor.

Right for symmetric. But tensorflow lite also uses [-128, 127] for asymmetric tensor. How about supporting both type separately?

glistening: > `ir::DataType::QUANT_INT8_ASYMM` has range as [-128, 127]. I think we support full range of int to verify this type.

First, `gen_golden` does not take input type from nnfw IR, its input is `.tflite`. It just handle `int8` type. Valid input range for `int8` depends on model. I don't want to use tighten the random range which sometimes caused unnecessary errors.

glistening: @ragmani I've just found that I am not the only one who would like to narrow the random input range. Here is the code from tensorflow lite's `benchmark_model`.

```c++
case kTfLiteInt32: {
  int low = has_value_range ? low_range : 0;
  int high = has_value_range ? high_range : 99;
  return CreateInputTensorData<int32_t>(
      num_elements, std::uniform_int_distribution<int32_t>(low, high));
}
...
case kTfLiteInt8: {
  int low = has_value_range ? low_range : -127;
  int high = has_value_range ? high_range : 127;
```

ragmani: > First, gen_golden does not take input type from onert IR, its input is .tflite. It just handle int8 type. Valid input range for int8 depends on model. I don't want to use tight random range which sometimes caused unnecessary errors.

I see. 

> `int8` is not just for `ir::DataType::QUANT_INT8_ASYMM` in gen_golden.py. It should work for ir::DataType::QUANT_INT8_SYMM (not exist at this moment and I am not sure it is necessary).

If `gen_golden.py` generates input of `.tflite`, `int8` input can be `QUANT_INT8_ASYMM` in many cases. But as long as it's not easy to distinguish `int8` in `gen_golden.py`, it seems good to do the range to [-127, 127] what you say.

ragmani (approved):
LGTM

PR #5438, opened 2020-12-18
DRAFT: Convert NCHW to NHWC

jinevening:
On going draft to convert NCHW to NHWC.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: #5401

PR #5454, opened 2020-12-22
[luci] Convert NCHW to NHWC for ADD

jinevening:
This introduces a pass to convert NCHW to NHWC for ADD Op.

---
Related to: #5401
Draft PR: #5438

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
@@ -46,6 +46,7 @@ struct ConvertNCHWToNHWCPass final : public logo::Pass
 
   bool run(loco::Graph *g) final;
 
+private:
   std::set<loco::Node *> _target_ops;
   std::map<loco::Node *, DataFormat> _op_status;
 };
--- a/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_CONVERT_NCHW_TO_NHWC_PASS_H__
+#define __LUCI_CONVERT_NCHW_TO_NHWC_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to convert NCHW Ops to NHWC
+ * @details Find operators that use NCHW layout and make them use NHWC.
+ *          Strictly speaking, it is impossible to distinguish whether
+ *          an operator is using NCHW or NHWC without programmers' annotations.
+ *          But we guess the data layout of each operator as much as possible
+ *          based on the assumptions described in the comments.
+ *          Note that this Pass does not change the execution result even
+ *          for the false-positive cases.
+ */
+struct ConvertNCHWToNHWCPass final : public logo::Pass
+{
+  virtual ~ConvertNCHWToNHWCPass() = default;
+
+  enum class DataFormat
+  {
+    NCHW,
+    NHWC
+  };
+
+  const char *name(void) const final { return "luci::ConvertNCHWToNHWCPass"; }
+
+  bool run(loco::Graph *g) final;
+
+  std::set<loco::Node *> _target_ops;
+  std::map<loco::Node *, DataFormat> _op_status;
+};
+
+} // namespace luci
+
+#endif // __LUCI_CONVERT_NCHW_TO_NHWC_PASS_H__
--- a/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
@@ -23,7 +23,8 @@ namespace luci
 {
 
 /**
- * @brief  Class to convert NCHW Ops to NHWC
+ * @brief   Class to convert NCHW Ops to NHWC
+ *
  * @details Find operators that use NCHW layout and make them use NHWC.
  *          Strictly speaking, it is impossible to distinguish whether
  *          an operator is using NCHW or NHWC without programmers' annotations.
--- a/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
@@ -36,19 +36,9 @@ struct ConvertNCHWToNHWCPass final : public logo::Pass
 {
   virtual ~ConvertNCHWToNHWCPass() = default;
 
-  enum class DataFormat
-  {
-    NCHW,
-    NHWC
-  };
-
   const char *name(void) const final { return "luci::ConvertNCHWToNHWCPass"; }
 
   bool run(loco::Graph *g) final;
-
-private:
-  std::set<loco::Node *> _target_ops;
-  std::map<loco::Node *, DataFormat> _op_status;
 };
 
 } // namespace luci
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -25,6 +25,41 @@
 namespace
 {
 
+enum class DataFormat
+{
+  NCHW,
+  NHWC
+};
+
+class DataFormatAnnotation final : public loco::NodeAnnotation
+{
+public:
+  DataFormatAnnotation(const DataFormat &format) : _format{format}
+  {
+    // DO NOTHING
+  }
+
+public:
+  const DataFormat &format(void) const { return _format; }
+
+private:
+  DataFormat _format;
+};
+
+void set_data_format(loco::Node *node, const DataFormat &format)
+{
+  assert(node->annot<DataFormatAnnotation>() == nullptr);
+  node->annot(std::make_unique<DataFormatAnnotation>(format));
+}
+
+DataFormat get_data_format(loco::Node *node)
+{
+  assert(node->annot<DataFormatAnnotation>() != nullptr);
+  return node->annot<DataFormatAnnotation>()->format();
+}
+
+bool has_data_format(loco::Node *node) { return node->annot<DataFormatAnnotation>() != nullptr; }
+
 luci::CircleTranspose *create_4d_transpose(luci::CircleNode *node,
                                            const std::vector<int32_t> indices)
 {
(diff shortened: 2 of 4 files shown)


compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+#include "luci/Pass/ConvertNCHWToNHWCPass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleNodeVisitor.h>
+#include <luci/Log.h>
+
+#include <loco/Service/ShapeInference.h>
+
+#include <iostream>
```
mhs4670go, line 25: maybe for debugging?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+using DataFormat = ConvertNCHWToNHWCPass::DataFormat;
+
+bool ConvertNCHWToNHWCPass::run(loco::Graph *g)
+{
+  LOGGER(l);
+  INFO(l) << "ConvertNCHWToNHWCPass Start" << std::endl;
+
+  // Collect target operators (done once)
+  if (_op_status.empty())
```
mhs4670go, line 227: @jinevening What if the op, that `_op_status` has, is removed or changed? Isn't it possible to derefer null ptr?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: My mistake.. Thanks :)

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: `_op_status` is `std::map`, declared with `ConvertNCHWToNHWCPass`, so  I think it cannot be nullptr.

I found that _op_status is public by default (the pass is `struct`, not `class`). I'll make them `private` not to be modified by others.

mhs4670go: @jinevening Say this node is one of `_op_status` whose name is A and node A is removed by other pass. And then, back to this pass, `_op_status` still has node A that is actually `nullptr`. I might be wrong because I just skimmed this PR. Can't it be happend?

jinevening: > node A is removed by other pass

I think this will not happen because `_op_status` is a private member of `ConvertNCHWToNHWCPass`. There is no way to access `_op_status` outside of this pass (Please tell me if I'm wrong :sweat_smile:).

jinevening: @mhs4670go  I misunderstood the comment. You're right. It is not safe to maintain `_op_status` because nodes can be erased by other passes. I think I can use `loco::NodeAnnotation` to avoid such cases. I'll update the PR soon. Thanks for the review.

jinevening:
@mhs4670go I've updated the code to use Annotation. PTAL

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -0,0 +1,311 @@
...
+      assert(circle_node->rank() == 4);
+      if (circle_node->accept(&converter))
+      {
+        set_data_format(node, DataFormat::NHWC);
+        changed = true;
+        break;
+      }
+      else
+      {
+        throw std::runtime_error("Unsupported pattern. " + circle_node->name() +
```
seanshpark, line 301: If this is illegal network, `throw` is OK. If not, please just `continue`. If you want to detect unknown cases you want to discover later, using `assert` is OK.

jinevening: This case is not an illegal network but a legal network with unsupported patterns. I'll use `continue` instead of `throw`. Thanks.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -0,0 +1,310 @@
...
+         indices[2] * dimension.dim(3).value() + indices[3];
+}
+
+luci::CircleConst *create_NHWC_from_NCHW(luci::CircleConst *constant)
+{
+  assert(constant->rank() == 4);
+
+  // TODO: Support non-float types
+  if (constant->dtype() != loco::DataType::FLOAT32)
+    throw std::runtime_error("ConvertNCHWToNHWCPass only supports FP32 model for now.");
```
seanshpark, line 107: maybe change this to `return nullptr` and handle this in caller. as non float32 can exist then it would be better to skip this case. adding a log would be also nice.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    if (!has_data_format(node))
+    {
+      // Unsupported Op
+      continue;
+    }
+    if (has_data_format(node) && get_data_format(node) == DataFormat::NHWC)
```
mhs4670go, line 283:
```suggestion
    if (get_data_format(node) == DataFormat::NHWC)
```

jinevening: I've updated the code. Thanks.

seanshpark:
Adding `ConvertNCHWToNHWCPass` object as an annotation will live until the object's life.
I am not agaist this usage but this solution may provide a method to other future Passes to use this pattern and may increase memory.
Can we use Pass private data, maybe `std::map`, so that the annotation format data may exist only as local for the Pass?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: The comments (nullptr + log) have been applied. Thanks.

jinevening:
> Can we use Pass private data, maybe std::map, so that the annotation format data may exist only as local for the Pass?

This was my first approach. I changed the implementation because other passes can erase Nodes in the std::map, which may cause dangling pointers. (See @mhs4670go 's [comment](https://github.com/Samsung/ONE/pull/5454#discussion_r547131215) for more details)

> Add: another way is to clear this annotation before exiting Pass.

Data layout information should be kept across multiple runs.

> this solution may provide a method to other future Passes to use this pattern and may increase memory.

I understand your concern. Can I add a comment that warns not to use large data in Annotation?

seanshpark:
> Can I add a comment that warns not to use large data in Annotation?

Yes, that would be nice :)

seanshpark:
> other passes can erase Nodes in the std::map

I don't think this will happend if the `std::map` is used in the life of this Pass.
Usually in a Pass, new nodes are created, link is changed. 
But nodes are not deleted. They are deleted at `RemoveDeadNode` Pass.

> Data layout information should be kept across multiple runs.

This seems strange to me. I'm not expert on optimization passes so can you provide any reference with other compilers doing like this?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -31,6 +31,13 @@ enum class DataFormat
   NHWC
 };
 
+/**
+ * @brief Set annotation for DataFormat (NCHW, NHWC)
+ *
+ * @note DataFormatAnnotation will live longer than this Pass (until the
+ *       annotated loco::Node is erased). So, do not use large data in the
```
seanshpark, line 38: I've not seen where `annotated loco::Node is erased`...Can you share where the code do this?

jinevening:
> I don't think this will happend if the std::map is used in the life of this Pass.
Usually in a Pass, new nodes are created, link is changed.
But nodes are not deleted. They are deleted at RemoveDeadNode Pass.

The problem is that ConvertNCHWToNHWCPass  can run after RemoveDeadNodePass. Let's imagine passes are run as below.

`ConvertNCHWToNHWCPass -> ... (some nodes become dead nodes) -> RemoveDeadNodePass -> ... -> ConvertNCHWToNHWCPass`

In this case, reference of erased loco::Node is not valid in the second run of `ConvertNCHWToNHWCPass`.

> Data layout information should be kept across multiple runs.
> This seems strange to me. I'm not expert on optimization passes so can you provide any reference with other compilers doing like this?

My comment might have confused you. The above comment was not about general optimization passes but just about this pass. "**This pass requires that** data layout information should be kept across multiple runs".

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -0,0 +1,324 @@
...
+      // Both inputs are not constant.
+      // In this case, we cannot distinguish NCHW from NHWC,
+      // so just insert Transpose Ops.
+      auto pre_trans_x = create_pre_transpose(node);
+      pre_trans_x->a(node->x());
+      node->x(pre_trans_x);
+
+      auto pre_trans_y = create_pre_transpose(node);
+      pre_trans_y->a(node->y());
+      node->y(pre_trans_y);
```
mhs4670go, line 243: @jinevening 
You might write this logic because redundant transpose op will be deleted by `RemoveRedundantTranspose` pass.
Say there is this network like this.
```
NHWC - Add(NHWC)  - Conv2D(NHWC) 
```
This network will be like below after this pass
```
NHWC - Transpose - Add(NWCH) - Transpose' - Conv2D(NHWC)
```
This `Transpose` also will be deleted? or it doesn't have to be?

mhs4670go:
@jinevening How about changing `RemoveRedundantTranspose` pass to just method or sth not to run multiple runs? Then, you can all the method at the end of this pass.

seanshpark:
From a small off-line meeting;
- we can utilize data format `DataFormat_CHANNELS_FIRST` of the graph in circle format --> this may remove my concern that user can run this multiple times for the model file, that may result an unintended result.
- let's go with annotation; I now understand that this algorithm needs the life cycle of the annotation data to exist in multiple runs of this Pass
- we may split Phase so that only this Pass may run in that Phase; although this may not be necessary

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: > Say there is this network like this.
NHWC - Add(NHWC)  - Conv2D(NHWC) 

The output of onnx-tf does not produce such a pattern. onnx-tf will create a network for Add-Conv2D as below (Transposes are located before/after Conv2D).

`NCHW - Add(NCHW)  - NCHW - Transpose - NHWC - Conv2D(NHWC)  - NHWC - Transpose - NCHW`

This pass will change the network as below (Transposes are added before/after Add).

`NCHW - Transpose - NHWC - Add(NHWC)  - NHWC - Transpose - NCHW - Transpose - NHWC - Conv2D(NHWC)  - NHWC - Transpose - NCHW`

The Transposes between Add and Conv2D can be removed by `RemoveRedundantTransposePass`.

jinevening:
> How about changing RemoveRedundantTranspose pass to just method or sth not to run multiple runs?Then, you can run the method at the end of this pass.

Can you explain the benefit for the proposed design over the current one?

jinevening:
> we can utilize data format DataFormat_CHANNELS_FIRST of the graph in circle format

This requires changes in import and export. I think it would be better to handle the subject in the separate PR (or issue).

mhs4670go:
@jinevening I thought the reason the pass should be run multiple times is that `RemoveRedundantTranspose` follows this pass. If `RemoveRedundantTranspose` can be called as a library or a method, it can be run at the end of this pass. Then, there is no need to keep any information until the whole phase ends.

jinevening:
@mhs4670go The reason for maintaining data format information is not related to `RemoveRedundantTranspose`. Even though we remove redundant transpose right after NCHW -> NHWC conversion, we still need data format information not to do the conversion again to the same node.

mhs4670go:
@jinevening Oh, I misunderstood. Then, if #5487 is merged, this problem will be resolved?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
mhs4670go: @jinevening Thank  you for your explanation. I just assumed a network that has both of NCHW pattern and NHWC one. But this seems overthinking:)

jinevening:
> Then, if #5487 is merged, this problem will be resolved?

Yes :)

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: I think this question was answered in the offline talk, so I check this conversation resolved.

jinevening:
@seanshpark Can you merge this if there is no more issue ? #5487 may take some time, and there are several PRs related to NCHW->NHWC conversion (https://github.com/Samsung/ONE/pull/5438#issuecomment-749307503) which need this PR. There is a schedule for model delivery, so I'd like to do both jobs in parallel to meet the deadline.

mhs4670go:
@jinevening Just curiosity, #5492 seems to make it possible to distinguish data format from others in subgraphs-level. Will it lower its level by op-level?

compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
```diff
@@ -0,0 +1,46 @@
...
+#define __LUCI_CONVERT_NCHW_TO_NHWC_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to convert NCHW Ops to NHWC
+ * @details Find operators that use NCHW layout and make them use NHWC.
```
seanshpark, line 27: Minor thing, can you please (1) align `Class` and `Find` (2) add an empty line between @brief and @details

seanshpark: Hm... `@brief` is a valid user account... and also `@details`... sorry for sending notifications :)

jinevening:
> @jinevening Will it lower its level by op-level?

No. #5492 sets data format in a subgraph level. Op level data format will be expressed using annotation as this PR does.

seanshpark (approved):
LGTM thank you!

compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
jinevening: > sorry for sending notifications :)

No problem :)

Updated comments. Thanks.

mhs4670go (approved):
LGTM

Issue #5416, opened 2020-12-17
[Compiler][luci] Introduce new optimization pass, RemoveNoEffectSlicePass.


struss:
Parent Issue : #4353
#
### What to do in this Pass
Remove no effect(input and output shape same) `CircleSlice`.
#### Input Type
Slice Op with input output shape same.
<img src=https://user-images.githubusercontent.com/24720192/93737854-27321200-fc1f-11ea-82c0-f2df799c1138.png width='150' />

### TODO
- [x] Introduce luci/Pass for above optimization. #5457
- [x] Introduce this optimization on circle2circle. #5532 
### Draft
#5418

struss:
I think this case must handle a case for input with not defined shape.

struss:
Done! Thanks for all!

PR #5457, opened 2020-12-22
[luci/pass] Introduce RemoveUnnecessarySlicePass.

struss:
For #5416 

This will introduce RemoveUnnecessarySlicePass on luci/pass.

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/RemoveNoEffectSlicePass.h
+++ b/compiler/luci/pass/include/luci/Pass/RemoveNoEffectSlicePass.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_REMOVE_NO_EFFECT_SLICE_PASS_H__
+#define __LUCI_REMOVE_NO_EFFECT_SLICE_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to Remove No Effect(input and output are same) Slice node.
+ */
+struct RemoveNoEffectSlicePass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::RemoveNoEffectSlicePass"; }
+
+  bool run(loco::Graph *g) final;
+};
+
+} // namespace luci
+
+#endif // __LUCI_REMOVE_NO_EFFECT_SLICE_PASS_H__
--- a/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
@@ -23,7 +23,8 @@ namespace
 
 /// @brief Return value in CircleConst.
 /// @details Return value in position on CircleConst with int64 format.
-///          Begin & size must be larger than or equal to -1.
+///          Begin must be larger than or equal to 0.
+///          Size must be larger than or equal to -1.
 ///          Throw an error on Invalid node or invalid node type.
 int64_t value_from_circle_const(const luci::CircleConst *node, uint32_t idx)
 {
--- a/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
@@ -22,18 +22,18 @@ namespace
 {
 
 /// @brief Return value in CircleConst.
-/// @details Return -2 if CircleConst is nullptr or not valid shape, otherwise return value in
-///          position on CircleConst with int64 format.
-///          Begin & size must be larger than or equal to -1, so -2 is invalid value.
+/// @details Return value in position on CircleConst with int64 format.
+///          Begin & size must be larger than or equal to -1.
+///          Throw an error on Invalid node or invalid node type.
 int64_t value_from_circle_const(const luci::CircleConst *node, uint32_t idx)
 {
   if (node == nullptr || node->rank() != 1 || node->dim(0).value() <= idx)
-    return -2;
+    throw std::runtime_error("Invalid constant node on slice");
   if (node->dtype() == loco::DataType::S64)
     return node->at<loco::DataType::S64>(idx);
   else if (node->dtype() == loco::DataType::S32)
     return static_cast<int64_t>(node->at<loco::DataType::S32>(idx));
-  return -2;
+  throw std::runtime_error("Invalid constant type on slice");
 }
 
 bool remove_no_effect_slice(luci::CircleNode *node)
@@ -62,7 +62,7 @@ bool remove_no_effect_slice(luci::CircleNode *node)
       continue;
 
     if (input_node->shape_signature().rank() != 0 && input_node->shape_signature().dim(i) == -1)
-        return false;
+      return false;
 
     if (size_value != static_cast<int64_t>(input_node->dim(i).value()))
       return false;
@@ -86,6 +86,8 @@ namespace luci
  *      |
  *
  *    AFTER
+ *      |
+ * [CircleNode]
  *      |
  * [CircleNode] Remove Slice OP
  *      |
--- a/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
@@ -24,7 +24,7 @@ namespace
 /// @brief Return value in CircleConst.
 /// @details Return -2 if CircleConst is nullptr or not valid shape, otherwise return value in
 ///          position on CircleConst with int64 format.
-///          on this pass, begin & size must be large or equal to -1, so -2 is invalid value.
+///          Begin & size must be larger than or equal to -1, so -2 is invalid value.
 int64_t value_from_circle_const(const luci::CircleConst *node, uint32_t idx)
 {
   if (node == nullptr || node->rank() != 1 || node->dim(0).value() <= idx)
--- a/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
@@ -27,7 +27,7 @@ namespace
 ///          Throw an error on Invalid node or invalid node type.
 int64_t value_from_circle_const(const luci::CircleConst *node, uint32_t idx)
 {
-  if (node == nullptr || node->rank() != 1 || node->dim(0).value() <= idx)
+  if (node->rank() != 1 || node->dim(0).value() <= idx)
     throw std::runtime_error("Invalid constant node on slice");
   if (node->dtype() == loco::DataType::S64)
     return node->at<loco::DataType::S64>(idx);
--- a/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
@@ -41,12 +41,15 @@ bool remove_no_effect_slice(luci::CircleNode *node)
   auto target_node = dynamic_cast<luci::CircleSlice *>(node);
   if (target_node == nullptr)
     return false;
+
   auto begin_const = dynamic_cast<luci::CircleConst *>(target_node->begin());
   if (begin_const == nullptr)
     return false;
+
   auto size_const = dynamic_cast<luci::CircleConst *>(target_node->size());
   if (size_const == nullptr)
     return false;
+
   // Check input output shape.
   auto input_node = loco::must_cast<luci::CircleNode *>(target_node->input());
   for (uint32_t i = 0; i < input_node->rank(); i++)
--- a/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/RemoveNoEffectSlicePass.h"
+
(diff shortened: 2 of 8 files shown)


compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/// @brief Return value in CircleConst.
+/// @details Return -2 if CircleConst is nullptr or not valid shape, otherwise return value in
+///          position on CircleConst with int64 format.
+///          on this pass, begin & size must be large or equal to -1, so -2 is invalid value.
+int64_t value_from_circle_const(luci::CircleConst *node, uint32_t idx)
```
jinevening, line 28:
```suggestion
int64_t value_from_circle_const(const luci::CircleConst *node, uint32_t idx)
```

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/// @brief Return value in CircleConst.
+/// @details Return -2 if CircleConst is nullptr or not valid shape, otherwise return value in
+///          position on CircleConst with int64 format.
+///          on this pass, begin & size must be large or equal to -1, so -2 is invalid value.
```
jinevening, line 27:
```suggestion
///          Begin & size must be larger than or equal to -1, so -2 is invalid value.
```

jinevening (dismissed):

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,112 @@
...
+  auto target_node = dynamic_cast<luci::CircleSlice *>(node);
+  if (target_node == nullptr)
+    return false;
+  auto begin_const = dynamic_cast<luci::CircleConst *>(target_node->begin());
+  if (begin_const == nullptr)
+    return false;
+  auto size_const = dynamic_cast<luci::CircleConst *>(target_node->size());
+  if (size_const == nullptr)
+    return false;
+  // Check input output shape.
```
llFreetimell, line 50:
```suggestion
  auto target_node = dynamic_cast<luci::CircleSlice *>(node);
  if (target_node == nullptr)
    return false;

  auto begin_const = dynamic_cast<luci::CircleConst *>(target_node->begin());
  if (begin_const == nullptr)
    return false;

  auto size_const = dynamic_cast<luci::CircleConst *>(target_node->size());
  if (size_const == nullptr)
    return false;

  // Check input output shape.
```

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+    else
+    {
+      if (input_node->shape_signature().rank() != 0 && input_node->shape_signature().dim(i) == -1)
+        return false;
+    }
+    if (size_value != static_cast<int64_t>(input_node->dim(i).value()))
+      return false;
+  }
+  replace(target_node).with(input_node);
```
llFreetimell, line 74: According to this code, following comment would be right but I think that it is not what you intended...?
```
	 *   BEFORE
 *      |
 * [CircleNode]
 *      |
 * [CircleSlice]
 *      |
 * [CircleNode](with same shape)
 *      |
 *
 *    AFTER
 *      |
 * [CircleNode]
 *      |
 * [CircleNode] Remove Slice OP
 *      |
```

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,115 @@
...
+namespace
+{
+
+/// @brief Return value in CircleConst.
+/// @details Return -2 if CircleConst is nullptr or not valid shape, otherwise return value in
+///          position on CircleConst with int64 format.
+///          Begin & size must be larger than or equal to -1, so -2 is invalid value.
+int64_t value_from_circle_const(const luci::CircleConst *node, uint32_t idx)
+{
+  if (node == nullptr || node->rank() != 1 || node->dim(0).value() <= idx)
```
llFreetimell, line 30: `node == nullptr` will never happen because `size_const` and `begin_const` is already checked by line 42 and line 46.

If some invalid situation is occurred, throwing runtime error would be better because it should not be happened.

llFreetimell: And one more thing, returning -2 seems weird because `value_from_circle_const` implies **the value in the const node**. However -2 is not in the node.

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
struss: I think comment that you draw is right on this. I will update this.

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
struss: Okay I will change to throw error on this case. and also invalid Data type case.

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,115 @@
...
+        return false;
+      size_value = static_cast<int64_t>(input_node->dim(i).value());
+    }
+    else
+    {
+      if (input_node->shape_signature().rank() != 0 && input_node->shape_signature().dim(i) == -1)
+        return false;
+    }
+    if (size_value != static_cast<int64_t>(input_node->dim(i).value()))
+      return false;
```
llFreetimell, line 72:
```suggestion
    if (value_from_circle_const(begin_const, i) != 0)
      return false;

    int64_t size_value = value_from_circle_const(size_const, i);
    if (size_value == -1)
      return true;

    if (input_node->shape_signature().rank() != 0 && input_node->shape_signature().dim(i) == -1)
        return false;

    if (size_value != static_cast<int64_t>(input_node->dim(i).value()))
      return false;
```

struss:
@llFreetimell @jinevening Thanks for pointing out!

llFreetimell:
`Merge branch ...` should not be appeared and the contents is weird...!

struss:
> `Merge branch ...` should not be appeared and the contents is weird...!

Oh I will fix it.

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "luci/Pass/RemoveNoEffectSlicePass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/// @brief Return value in CircleConst.
+/// @details Return value in position on CircleConst with int64 format.
+///          Begin & size must be larger than or equal to -1.
```
llFreetimell, line 26: Begin must be larger or equal  than 0.

struss: Oh, I will fix this comment.

a-zh:
Hello. I came across a similar case for STRIDED_SLICE. Is needed the second Pass or it can be integrated with current Slice?

Model fragment:
strided_slice2

llFreetimell:
@a-zh Hello!
Although I am not the author of this PR, my opinion is that we may go as it is :)
Reasons are...

- There may be more cases that some operations have no effect
- The authors are different so it can be another burden for communication to make one PR by two people

So after the two passes are merged and more cases are investigated, we may be able to integrate them as one pass.

Of course, this is only my opinion :-)

compiler/luci/pass/include/luci/Pass/RemoveNoEffectSlicePass.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to Remove No Effect(input and output are same) Slice node.
+ */
+struct RemoveNoEffectSlicePass final : public logo::Pass
```
seanshpark, line 28: How about using `RemoveRedundantSlicePass` ?

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+#include "luci/Pass/RemoveNoEffectSlicePass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/// @brief Return value in CircleConst.
```
seanshpark, line 24: Let's use `/** *@ */` style as like the others.

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,114 @@
...
+/// @details Return value in position on CircleConst with int64 format.
+///          Begin must be larger than or equal to 0.
+///          Size must be larger than or equal to -1.
+///          Throw an error on Invalid node or invalid node type.
+int64_t value_from_circle_const(const luci::CircleConst *node, uint32_t idx)
+{
+  if (node->rank() != 1 || node->dim(0).value() <= idx)
+    throw std::runtime_error("Invalid constant node on slice");
```
seanshpark, line 32: Q) If this is to throw invalid network, I think it should be checked in import stage.

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+} // namespace
+
+namespace luci
+{
+/**
+ *   BEFORE
+ *      |
+ * [CircleNode]
```
seanshpark, line 82: This diagram seems there is a `BEFORE` node and this is connected to  `[CircleNode]`.
It would be better to use indentation to distinghish this.

compiler/luci/pass/src/RemoveNoEffectSlicePass.test.cpp
```diff
@@ -0,0 +1,121 @@
...
+  slice->begin(begin);
+  slice->size(size);
+
+  // Output Connect.
+  auto output = g->nodes()->create<luci::CircleOutput>();
+  output->from(slice);
+  auto graph_output = g->outputs()->create();
+  output->index(graph_output->index());
+
+  return;
```
seanshpark, line 89: maybe remove this line.

compiler/luci/pass/include/luci/Pass/RemoveNoEffectSlicePass.h
struss: um...what about `RemoveUnnecessarySlicePass`? Redundant have a meaning that have extra thing with same work.

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
struss: Okay. I will remove this shape check and type check below.
I will add a checker on luci/Import

https://github.com/Samsung/ONE/blob/fde0b8aa4d1aa76019b9461370b13e15fc40a9fd/compiler/luci/import/src/Nodes/CircleSlice.cpp#L35

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
struss: Oh, I will fix it.

compiler/luci/pass/src/RemoveNoEffectSlicePass.test.cpp
struss: Okay I will remove the line.

compiler/luci/pass/src/RemoveNoEffectSlicePass.cpp
struss: Okay I will fix it.

compiler/luci/pass/include/luci/Pass/RemoveNoEffectSlicePass.h
seanshpark: OK, `RemoveUnnecessarySlicePass` seems nice.

struss: Okay I will fix it Thanks!

struss:
@a-zh Thanks for suggestion. but currently, I'm creating more no-effect operation removal cases. I also think that creating each case would be better. and also, after adding enough, merge into one could be nice. just I think.

seanshpark:
Please rebase to latest master

compiler/luci/pass/src/RemoveUnnecessarySlicePass.cpp
```diff
@@ -0,0 +1,111 @@
...
+ *    [CircleNode]
+ *
+ * AFTER
+ *
+ *    [CircleNode]
+ *          |
+ *    [CircleNode]
+ *
+ * Slice OP has no effect if,
+ *    1. Static Shape : begin value = 0 and size value = -1 or input dimension
```
seanshpark, line 94: `or input dimension` : not sure what `input dimension` be some condition

seanshpark: above code `if (size_value == -1)` and comment `size value = -1` doesn't match

struss: what about using below?
```
 *    1. Static Shape : begin_const[idx] = 0 && size_const[idx] = -1 or input_dimension[idx]
 *    2. Dynamic Shape : begin_const[idx] = 0 && size_const[idx] = -1
```

seanshpark: How about;
begin_const[idx] = 0 && size_const[idx] = -1 or input_dimension[idx]
-->
begin_const[idx] is 0 AND size_const[idx] is (-1 OR input_dimension[idx])

compiler/luci/pass/src/RemoveUnnecessarySlicePass.test.cpp
```diff
@@ -0,0 +1,104 @@
...
+  slice->input(input);
+  slice->begin(begin);
+  slice->size(size);
+
+  // Output Connect.
+  auto output = g->nodes()->create<luci::CircleOutput>();
+  output->from(slice);
+  auto graph_output = g->outputs()->create();
+  output->index(graph_output->index());
+  return;
```
seanshpark, line 72: `return` can be removed.

struss: okay.I will fix it.

compiler/luci/pass/src/RemoveUnnecessarySlicePass.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/**
+ * @brief   Return value in CircleConst.
+ * @details Return value in position on CircleConst with int64 format.
+ *          Begin must be larger than or equal to 0. Size must be larger
+ *          than or equal to -1. Throw an error on invalid node or invalid
```
seanshpark, line 28: `Throw an error on ~` was removed from this method. Please update this or remove.

struss: I will remove this. Thanks!

seanshpark:
@struss , thanks for this change. One more thing... could you add one or more negative tests for this pass that doesn't remove the Slice when the condition is not met?

seanshpark (approved):
LGTM thanks!

seanshpark:
Optionally in another PR, please add more negative tests for condition checks for the Pass.

struss:
> Optionally in another PR, please add more negative tests for condition checks for the Pass.

Okay, I will add NEG case on other passes I've added.

llFreetimell (approved):
LGTM

mhs4670go (approved):
LGTM

jinevening (approved):

PR #5465, opened 2020-12-22
[tf2tfliteV2] Fix default value of input_shapes

llFreetimell:
When `input_shapes` is not given for `_v2_convert`, input_shapes becomes `None` and
it causes an error at `len(input_shapes)`.
This commit will fix it.


jinevening (approved):
LGTM Thank you.

mhs4670go (approved):
LGTM

PR #5471, opened 2020-12-22
[onert] Remove redundant check in `initConsts`

wateret:
Remove redundant check in `initConsts`, it is already done a few lines
above.


glistening (approved):
LGTM

Issue #5352, opened 2020-12-11
[onert] Refactor BackendContext - graph and assigned operand/operation info


wateret:
Part of #5006 

Currently `BackendContext` gets target operation/operand info like this.

- Full `LoweredGraph`
- Used operations in this backend (`BackendContext::operation_list()`)
- Used operands in this backend (`BackendContext::operand_list()`)

However this is hard for backend implementers and also error-prone. Because it gets too much access to objects(passing the whole graph and whole LowerInfo) which shouldn't have been.

 So I'm thinking about changing it to be like this.

- Partial `LoweredGraph` (the part that is used by this backend)
- External(portable) tensor info
- Partial order of `LoweredGraph`(They are not all connected)

wateret:
> * Partial `LoweredGraph` (the part that is used by this backend)
> * External(portable) tensor info
> * Partial order of `LoweredGraph`(They are not all connected)

Let me visualise these.

Here's a LoweredGraph(backends are assigned). Sub operations are assigned to Green, Adds are assigned to Red. And new BackendContext will get these data:

![image](https://user-images.githubusercontent.com/10582184/102477839-3c90ce00-40a0-11eb-933f-a54372a3d34e.png)

wateret:
It looks like we also need layout info for BackendContext.

Roughly we need `ir::Graph` + `Data` (below) for BackendContext(runtime must pass these to BackendContext)

```cpp
struct Data
{
  set<OperandIndex> external_operands; // Set of external operands
  vector<OpSequence> op_seq_order; // Order of OpSequences
  LayoutInfo layout_info; // Layout info
};
```

wateret:
OpSequence makes it tricky. I am considering #5505 , #5537 is a draft for it.

wateret:
I found one more advantage of doing this. In ACL backends,  they sometimes need to extend dimensions(DimCorrection). And currently it is implemented as manipulating the shared graph. However with this change each graph passed to a backend could be changed and does not affect other backends.

wateret:
Done with #5926 !

PR #5481, opened 2020-12-24
[onert] ObjectManager can specify Index

wateret:
Allow ObjectManager can add an object with specifying an index for it.

### DIFF
--- a/runtime/onert/core/include/util/ObjectManager.h
+++ b/runtime/onert/core/include/util/ObjectManager.h
@@ -36,7 +36,7 @@ namespace util
 template <typename Index, typename Object> class ObjectManager
 {
 public:
-  ObjectManager() : _index_count{0u} {}
+  ObjectManager() : _next_index{0u} {}
 
 public:
   /**
@@ -53,16 +53,22 @@ public:
   }
 
   /**
-   * @brief Put object in the container with a new Index for that
+   * @brief Put object in the container with given index.
+   *
+   * If the index is not valid, the index will be automatically generated.
    *
    * @param[in] object Object to be pushed
-   * @return Created index that is associated to the object
+   * @param[in] index Index associated with the object
+   * @return If @c index is valid, return created index.
+   *         Otherwise, return @c index .
    */
-  Index push(std::unique_ptr<Object> &&object)
+  Index push(std::unique_ptr<Object> &&object, Index index = Index{})
   {
-    auto index = generateIndex();
-    _objects.emplace(index, std::move(object));
-    return index;
+    auto gen_index = generateIndex(index);
+    if (!gen_index.valid())
+      return gen_index;
+    _objects.emplace(gen_index, std::move(object));
+    return gen_index;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


wateret:

wateret:

runtime/onert/core/include/util/ObjectManager.h
```diff
@@ -135,11 +141,33 @@ template <typename Index, typename Object> class ObjectManager
...
+      else
+      {
+        // If the given index exists already, return a non-valid index
+        return Index{};
+      }
+    }
+    else
+    {
+      return Index{_next_index++};
+    }
```
glistening, line 165: @wateret Till now, we've assumed `_next_index` would not reach `UINT32_MAX` because it is always incremented by 1. Now, I've considered what will happen if user specify `UINT32_MAX - 1` as index. At the call,  `_next_index` will be `UINT32_MAX` (= invalid index) by line 153. Then, at the next call of `generateIndex()`, It will reach the lines 162-165 and return invalid index. Furthermore, if the caller does not check the return value, at the next calls, it will return 1, 2, 3, .... which is likely to be used by others.

wateret:

runtime/onert/core/include/util/ObjectManager.h
wateret: I have updated it. As the logic gets a little complicated, I separated the methods that used to get an index as an optional param. Thanks!

runtime/onert/core/include/util/ObjectManager.h
```diff
@@ -36,33 +36,55 @@ namespace util
...
-   * @brief Put object in the container with a new Index for that
+   * @brief Put the object in the container with given index.
+   *
+   * It fails when the given index is already taken or @c index is Undefined.
+   *
+   * @param[in] object Object to be pushed
+   * @param[in] index Index associated with the object
+   * @return @c index if successful, an Undefined index otherwise
+   */
+  Index push(std::unique_ptr<Object> &&object, Index index)
```
glistening, line 66: @wateret Two push(...) work differently. `push() with Index` checks the uniqueness of `index` while `push() w/o Index` checks the validity (whether it is `UNDEFINED` or not) only. I don't think `Index` is not guaranteed not in use for latter case.

I am curious what is the use case of specifying Index? 
- Do you want to re-use some Index which is not used any longer again? Or 
- Do you want to use Index bigger than current `_next_index`? Then why?

wateret: > while `push() w/o Index` checks the validity (whether it is `UNDEFINED` or not) only. I don't think Index is not guaranteed not in use for latter case.

We always keep `_next_index` as (highest index in the container + 1).  So it does not have to check every time. I'll add comment for that. (It is just to make sure not reusing some indices which could have been existed before)

> I am curious what is the use case of specifying Index?

I should've mentioned the background for needing this. 😅 

Here's what I'm going to do #5352 , and in the figure from https://github.com/Samsung/ONE/issues/5352#issuecomment-747361763 I'm changing the backend gets a partial subgraph(only the part for each backend needs). However all the operation/operand indices must be kept, so there should be something like this.

glistening (approved):
LGTM

PR #5483, opened 2020-12-24
[onert] Refactoring : Remove GraphIterator

wateret:
Remove GraphIterator and have a alternative implementation.

Two points that made it overly complicated.

- template based for const/non-const version consistency
    - Makes it hard to read
- interface based to support variety types of iterators
    - But actually we use only one iterator


YongseopKim (approved):
LGTM

PR #5489, opened 2020-12-28
[luci] Check DepthwiseConv input channel match

cgbahk:
This commit checks whether input channel depth matches.

ONE-DCO-1.0-Signed-off-by: Cheongyo Bahk <cg.bahk@gmail.com>

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.cpp
@@ -607,6 +607,7 @@ loco::NodeShape infer_depthwise_conv2d(const luci::CircleDepthwiseConv2D *node)
   assert(ifm_shape.rank() == 4);
   assert(ker_shape.rank() == 4);
   assert(ker_shape.dim(0).value() == 1);
+  assert(ifm_shape.dim(3).value() * node->depthMultiplier() == ker_shape.dim(3).value());
 
   auto os = infer_conv2d_type(node);
 


compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -607,6 +607,7 @@ loco::NodeShape infer_depthwise_conv2d(const luci::CircleDepthwiseConv2D *node)
   assert(ifm_shape.rank() == 4);
   assert(ker_shape.rank() == 4);
   assert(ker_shape.dim(0).value() == 1);
+  assert(ifm_shape.dim(3).value() * node->depthMultiplier() == ker_shape.dim(3).value());
```
cgbahk, line 610: Not sure this is right place to be. Is there stage to shape check for each circle op?

seanshpark: I think `bool CircleDepthwiseConv2DGraphBuilder::validate(const ValidateArgs &args) const` would be the place.

seanshpark: Condition check here is also good.

mhs4670go: > Condition check here is also good.
:+1: 

To cover optimization process, both are needed.

cgbahk: > I think `bool CircleDepthwiseConv2DGraphBuilder::validate(const ValidateArgs &args) const` would be the place.

Thanks, I would do when required :)

seanshpark:
@mhs4670go , we can add the check in import later if it's OK with you. At least this check is needed for our backend compiler.

seanshpark (approved):
LGTM

seanshpark:
FYI, this change will make `LWQDepthwiseConv2DExtra` fail.

mhs4670go (approved):
LGTM

Issue #5487, opened 2020-12-28
[luci] Use DataFormat in circle schema


jinevening:
## What

Support DataFormat in circle schema.

## Why

Currently luci does not correctly read/write DataFormat field in circle schema.

## ToDo

To be added

- [ ] luci/import
- [ ] luci/export
- [ ] Add data_format field in Module #5492
- [ ] circlechef
- [ ] circle recipes
- [ ] ConvertNCHWToNHWCPass
- [ ] tflite2circle
- circledump (already supported)
...
---
Related to: https://github.com/Samsung/ONE/issues/5401

binarman:
AFAIK luci-interpreter does not support channels-first layout also.
Should we add luci-interpreter as an item in this list or track it in separate issue?

jinevening:
> Should we add luci-interpreter as an item in this list or track it in separate issue?

I don't think so. The main purpose of this issue is to identify (and set) the data format of a circle model before/after NCHW->NHWC conversion (#5454). Currently luci-interpreter is used only for quantization, so it may not need NCHW support.

seanshpark:
Before landing #5492, have you verified (by some draft code) the logic you are thinking of?
We may have these situations...
- circle file with out dataformat
- circle file with dataformat channel last
- circle file with dataformat channel first

for each files, user wants to read as channel first using #5401

jinevening:
> Before landing #5492, have you verified (by some draft code) the logic you are thinking of?
We may have these situations...

#5492 only has some unit tests (I thought that's enough because the change is simple).

I'll upload full draft that tests the cases you've mentioned. It may take some time because we need test models (circle recipes) and update in circlechef to generate them (if someone knows a simpler solution for verification, please let me know:slightly_smiling_face:).

seanshpark:
> It may take some time because we need test models

Yes, please take your time :)

jinevening:
> Should we add luci-interpreter as an item in this list or track it in separate issue?

> We may need a guard code that prevents luci-interpreter from running NCHW model. Thanks.

At first glance, I thought that I can easily handle luci-interpreter in this issue by preventing luci-interpreter from running NCHW model. But in fact luci-interpreter can run NCHW model if the model only has "data-format-independent" operators, e.g., element-wise computations such as ADD or MUL. So, we should add guard codes only in the "data-format-dependent" operators (e.g., CONV) to prevent execution when the input of the operator is NCHW.

[...]

PR #5492, opened 2020-12-29
[luci] Introduce data format to Module

jinevening:
This introduces data format to Module.

---
Related to: #5487
Draft PR: #5520

### DIFF
--- a/compiler/luci/lang/include/luci/IR/Module.h
+++ b/compiler/luci/lang/include/luci/IR/Module.h
@@ -18,6 +18,7 @@
 #define __LUCI_MODULE_H__
 
 #include <loco/IR/Graph.h>
+#include "CircleDataFormat.h"
 
 #include <memory>
 #include <vector>
@@ -59,8 +60,19 @@ public:
 
   // TODO provide graph accessor with a name
 
+  /**
+   * @brief set data format of graph
+   */
+  void data_format(const loco::Graph *g, CircleDataFormat data_format);
+
+  /**
+   * @brief get data format of graph
+   */
+  CircleDataFormat data_format(const loco::Graph *g) const;
+
 private:
   std::vector<std::unique_ptr<loco::Graph>> _graphs;
+  std::map<const loco::Graph *, CircleDataFormat> _data_formats;
 };
 
 std::unique_ptr<Module> make_module(void);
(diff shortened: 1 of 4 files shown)


compiler/luci/lang/include/luci/IR/Module.h
```diff
@@ -18,6 +18,7 @@
 #define __LUCI_MODULE_H__
 
 #include <loco/IR/Graph.h>
+#include "CircleDataFormat.h"
```
seanshpark, line 21: I think this should be above `<loco/IR/Graph.h>`

seanshpark:
Before landing this, please check https://github.com/Samsung/ONE/issues/5487#issuecomment-751928531

jinevening:
I'll reopen this when the full draft is ready.

jinevening:
I re-opened this issue as full draft is ready. https://github.com/Samsung/ONE/pull/5520

compiler/luci/lang/include/luci/IR/Module.h
jinevening: @seanshpark I made CircleDataFormat in luci, because data format only exists in circle schema. Also, I don't want to change loco graph (we've talked about it).

seanshpark: I'm talking about the header sequence
```cxx
#include "CircleDataFormat.h"

#include <loco/IR/Graph.h>
```
so that `CircleDataFormat.h` file itself has no problem(self complete) with other headers, as you stated `I made CircleDataFormat in luci` so we include _luci_ first then base _loco_ order.

seanshpark:
@jinevening , is this PR active? if so, can you reply to what I wrote at https://github.com/Samsung/ONE/pull/5492#discussion_r551078504 ?

jinevening:
> is this PR active?

I think this PR should be deactivated until our schema can accept different data formats in one subgraph.

compiler/luci/lang/include/luci/IR/Module.h
jinevening: Sorry for the late reply. The comment is reasonable. But I'd like to close this PR for now as our schema does not support different data formats in one subgraph.

PR #5498, opened 2020-12-29
[luci] Convert NCHW to NHWC (CIRCLEINPUT, CIRCLEOUTPUT)

jinevening:
This converts NCHW to NHWC for CIRCLEINPUT and CIRCLEOUTPUT

---
Related to: #5401
Draft PR: #5438

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -55,6 +55,7 @@ private:
 
 void set_data_format(loco::Node *node, const DataFormat &format)
 {
+  assert(node->annot<DataFormatAnnotation>() == nullptr);
   node->annot(std::make_unique<DataFormatAnnotation>(format));
 }
 
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -70,12 +70,9 @@ bool has_data_format(loco::Node *node) { return node->annot<DataFormatAnnotation
 bool has_dynamic_shape(const loco::Node *node)
 {
   const auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
-  const auto shape_signature = circle_node->shape_signature().as_vector();
-  for (auto ss : shape_signature)
-  {
-    if (ss == -1)
+  for (uint32_t i = 0; i < circle_node->rank(); ++i)
+    if (!circle_node->dim(i).known())
       return true;
-  }
   return false;
 }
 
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -55,7 +55,6 @@ class DataFormatAnnotation final : public loco::NodeAnnotation
 
 void set_data_format(loco::Node *node, const DataFormat &format)
 {
-  assert(node->annot<DataFormatAnnotation>() == nullptr);
```
seanshpark, line 58: This change doesn't seem to be related with the description.

jinevening: I've made a separate PR for this. Thanks.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -212,6 +212,58 @@ class ConvertNCHWToNHWC final : public luci::CircleNodeMutableVisitor<bool>
...
+  bool visit(luci::CircleInput *node)
+  {
+    const auto n = node->dim(0).value();
+    const auto c = node->dim(1).value();
+    const auto h = node->dim(2).value();
+    const auto w = node->dim(3).value();
+
+    node->dim(1).set(h);
+    node->dim(2).set(w);
+    node->dim(3).set(c);
```
llFreetimell, line 224: When considering unknown dimensions, following codes

```suggestion
    const auto n = node->dim(0);
    const auto c = node->dim(1);
    const auto h = node->dim(2);
    const auto w = node->dim(3);

    node->dim(1) = h;
    node->dim(2) = w;
    node->dim(3) = c;
```

 or adding assert would be better.

seanshpark:
Input shape something like (-1, 10), (10, -1) also exists. @jinevening can you check with these?

jinevening:
> @jinevening can you check with these?

Sure, I'll add a check code that prevents converting dynamic-shaped models.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: I've updated the code. Thanks.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -67,6 +67,18 @@ DataFormat get_data_format(loco::Node *node)
...
+bool has_dynamic_shape(const loco::Node *node)
+{
+  const auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
+  const auto shape_signature = circle_node->shape_signature().as_vector();
+  for (auto ss : shape_signature)
+  {
+    if (ss == -1)
+      return true;
+  }
+  return false;
```
llFreetimell, line 79: Because of #5501, shape signature will be deprecated.
Please use following codes :)

```suggestion
  for (uint32_t i = 0;i < circle_node->rank(); ++i)
    if (!circle_node->dim(i).known())
      return true;
  return false;
```

jinevening: Thank you for the instruction :) I've updated the code.

llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5500, opened 2020-12-29
[luci] Fix for wrong assert in ConvertNCHWToNHWCPass.

jinevening:
This removes wrong assert in ConvertNCHWToNHWCPass.


jinevening:
@seanshpark Can you merge this? This is my mistake from #5454. The assert should not be inserted as the data format of node is updated after conversion is done.

seanshpark (approved):
LGTM

seanshpark:
Existing code seems OK but seems `set_xxx` should have been `update_xxx`

PR #5503, opened 2020-12-29
[luci] Fix wrong substitution in ResolveCustomOpBatchMatMulPass

llFreetimell:
Until now, `ResolveCUstomOpBatchMatMulPass` was substituting original
custom operation, not `CircleCustomOut`.
As a result, wrong optimization was done.
This commit will fix it.

### DIFF
--- a/compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.cpp
@@ -42,14 +42,9 @@ bool resolve_custom_op(luci::CircleCustom *cop)
     batch_matmul->adj_x(map["adj_x"].AsBool());
     batch_matmul->adj_y(map["adj_y"].AsBool());
 
-    for (auto s : loco::succs(cop))
-    {
-      if (auto cop_out = dynamic_cast<luci::CircleCustomOut *>(s))
-      {
-        replace(cop_out).with(batch_matmul);
-        changed = true;
-      }
-    }
+    auto customOut = loco::succs(cop);
+    assert(customOut.size() == 1);
+    replace(*customOut.begin()).with(batch_matmul);
   }
 
   return changed;
--- a/compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.cpp
@@ -25,8 +25,6 @@ namespace
 
 bool resolve_custom_op(luci::CircleCustom *cop)
 {
-  bool changed = false;
-
   const std::string custom_code = cop->custom_code();
   const std::vector<uint8_t> custom_options = cop->custom_options();
 
@@ -45,9 +43,11 @@ bool resolve_custom_op(luci::CircleCustom *cop)
     auto customOut = loco::succs(cop);
     assert(customOut.size() == 1);
     replace(*customOut.begin()).with(batch_matmul);
+
+    return true;
   }
 
-  return changed;
+  return false;
 }
 
 } // namespace
--- a/compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.cpp
@@ -25,6 +25,8 @@ namespace
 
 bool resolve_custom_op(luci::CircleCustom *cop)
 {
+  bool changed = false;
+
   const std::string custom_code = cop->custom_code();
   const std::vector<uint8_t> custom_options = cop->custom_options();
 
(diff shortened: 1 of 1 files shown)


llFreetimell:
Original graph of `BatchMatMulV2_001.opt.circle` was as following.

Screenshot from 2020-12-29 19-30-31

mhs4670go (approved):
LGTM

seanshpark:
@llFreetimell , can you add a diagram of the effect of this Pass (in form of BEFORE-AFTER) in the header file?
In this PR is OK and in another PR is also OK. I'd like to understand what this Pass does.
Reason: I don't recall we had any Passes that replaces successor nodes and it seems this is the first one. If you know any, please let me know.

llFreetimell:
> can you add a diagram of the effect of this Pass (in form of BEFORE-AFTER) in the header file?

Adding a diagram is not that difficult :)

> I don't recall we had any Passes that replaces successor nodes and it seems this is the first one.

I think that you are right. However, to fix this problem without successor nodes, another approach should be applied and then opinion of the author of this pass would be important than me :)
@mhs4670go What do you think?

mhs4670go:
@llFreetimell Please see `ResolveCustomOpAddPass`. I think for loop can be removed.

llFreetimell:
@seanshpark @mhs4670go 

I could find below codes.

https://github.com/Samsung/ONE/blob/70e43ca8bde6e7f9d88a79b850f140451013d0aa/compiler/luci/pass/src/ResolveCustomOpAddPass.cpp#L74-L76

Then what seems to be better?

- Remove `succs` both of them : Find `CustomOut` and then resolve the inputs
- Use `succs` here too : Go like above codes

seanshpark:
Im... for our logic, `CircleCustomOut` is to handle multiple `CircleCustom` outputs.
If `CircleCustom` is to be replaced with `CircleBatchMatMul` then this rule breaks.
I don't think existing code and current change reflects our design.

@mhs4670go , can you (as you are have ownership of this module), please consider our design and provide a better solution?

seanshpark:
Had a small talk with @mhs4670go and seems 
```cxx
 auto customOut = loco::succs(addv2); 
 assert(customOut.size() == 1); 
 replace(*customOut.begin()).with(add); 
```
will do.

llFreetimell:
Then I will fix like that!

seanshpark:
```
13:19:01 [ 77%] Generate BatchMatMulV2_000.opt.circle
13:19:01 ERROR: Optimized graph is invalid
13:19:01 make[2]: *** [compiler/common-artifacts/BatchMatMulV2_000.opt.circle] Error 255
13:19:01 compiler/common-artifacts/CMakeFiles/common_artifacts_deps.dir/build.make:2501: recipe for target 'compiler/common-artifacts/BatchMatMulV2_000.opt.circle' failed
13:19:01 make[1]: *** [compiler/common-artifacts/CMakeFiles/common_artifacts_deps.dir/all] Error 2
```

seanshpark:
Maybe (1) disable test, (2) land this change (3) fix the test

compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.cpp
```diff
@@ -40,10 +42,17 @@ bool resolve_custom_op(luci::CircleCustom *cop)
...
+      if (auto cop_out = dynamic_cast<luci::CircleCustomOut *>(s))
+      {
+        replace(cop_out).with(batch_matmul);
+        changed = true;
+      }
+    }
   }
-  return false;
+
+  return changed;
```
mhs4670go, line 50:
```suggestion
  return true;
```

seanshpark: @mhs4670go , `return true` seems to make infinite run. maybe set `changed = true` inside `if` block?

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.cpp
```diff
@@ -40,9 +40,13 @@ bool resolve_custom_op(luci::CircleCustom *cop)
...
     batch_matmul->adj_y(map["adj_y"].AsBool());
 
-    replace(cop).with(batch_matmul);
+    auto customOut = loco::succs(cop);
+    assert(customOut.size() == 1);
+    replace(*customOut.begin()).with(batch_matmul);
+
     return true;
   }
+
```
seanshpark, line 49: This seems unintended format change. @llFreetimell , do you want to revert this?

llFreetimell: I added this empty line for readability..! Shall I remove this?

seanshpark: Usually we separate changes for different purpose :)

PR #5504, opened 2020-12-29
[luci] Fix wrong checking in CIrcleTypeInference

llFreetimell:
`inputs_dtype_ready` should check `dtyps` of input nodes but until now,
original node was only checked.
This commit will fix it.


mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

seanshpark:
Wow, we all missed this mistake :)

PR #5506, opened 2020-12-29
[luci] Change assert to if condition in DepthwiseConv2D validation

mhs4670go:
This commit changes `assert` to `if` in DepthwiseConv2D validation.

To make validation more meaningful, it would be better to use `if` condition instead of `assert`.

FYI, 
```bash
$ cd tensorflow
$ cat tensorflow/tensorflow/core/framework/common_shape_fns.cc
...
Status DepthwiseConv2DNativeShapeImpl(shape_inference::InferenceContext* c,
                                      bool supports_explicit_padding) {
  ShapeHandle input_shape;
  TF_RETURN_IF_ERROR(c->WithRank(c->input(0), 4 /* NOTE HERE */, &input_shape));
  ShapeHandle filter_shape;
  TF_RETURN_IF_ERROR(c->WithRank(c->input(1), 4 /* NOTE HERE */, &filter_shape));
...
```

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleDepthwiseConv2D.cpp
+++ b/compiler/luci/import/src/Nodes/CircleDepthwiseConv2D.cpp
@@ -20,8 +20,6 @@
 
 #include <oops/UserExn.h>
 
-#include <cassert>
-
 namespace luci
 {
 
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM thank you!

cgbahk (approved):
LGTM

Issue #5501, opened 2020-12-29
[Compiler][luci] Refactoring shape/dtype inference


llFreetimell:
Originated from #4372,#4796
Draft PR : #5502, #5756

---

## What?

- Supporting dynamic shape inference with refactoring shape/dtype inference in `luci/Service` like following diagram.
```
   circle file     ---- import --->   luci/IR   ---- export --->   circle file
====================================================================================

   [1, 2, 3, 1]                                                    [1, 2, 3, 1]
     (shape)                        [?, 2, 3, 1]                      (shape)
                        ----->      (TensorShape)    ----->
  [-1, 2, 3, 1]                                                    [-1, 2, 3, 1]
(shape_signature)                                                (shape_signature)

====================================================================================

   [1, 2, 3, 1]                                                    [1, 2, 3, 1]
     (shape)                        [1, 2, 3, 1]                      (shape)
                        ----->      (TensorShape)    ----->
       []                                                                []
(shape_signature)                                                (shape_signature)
```

- Keypoints
  -  type inference : refactoring structure
  - shape inference : refactoring structure + logic upgrade
  - shape_signature inference : Deprecation
  - shape_signature attribute : Deprecation

## Why?

- There are some models which include dynamic shape but current `luci` cannot handle it.
- Current shape/type inference rule file is big so it would be better to implement in each operation file.

## How?

As following codes, `CircleNode` is using `TensorShape`

https://github.com/Samsung/ONE/blob/030371d447c4e222fe9e782f93fc1055344e79ee/compiler/luci/lang/include/luci/IR/CircleNodeDecl.h#L38-L40

and `TensorShape` has `loco::Dimension` which has following methods.

- `unset(void)` : Represent unknown dimension. It matched with `?` in above diagram.
https://github.com/Samsung/ONE/blob/4e4f9ffcc5f149d22123d26a38aed9b31f9d79c4/compiler/loco/include/loco/IR/Dimension.h#L59-L63

- `known(void)` : Judge whether the dimension is known or unknown.
https://github.com/Samsung/ONE/blob/4e4f9ffcc5f149d22123d26a38aed9b31f9d79c4/compiler/loco/include/loco/IR/Dimension.h#L47

By using these methods, we can express both static and dynamic shape.
All we need to do is upgrading shape inference logic which enables dynamic shape inference.
Implementation of shape inference(upgraded version) will be written by each operation together with type inference.

---

### Background History

[...]

PR #5507, opened 2020-12-30
[luci] Change candidates for circle type inference pass

llFreetimell:
Until now, `loco::postorder_traversal` and `loco::output_nodes` were used for getting candidates.
However, some nodes were not inferred even they are not dead nodes.
For example, `While_002` test has following graph.
```
              +-- CircleWhileOut(ofm1) -- CircleOutput
CircleWhile --+
              +-- CircleWhileOut(ofm2)
```
In the above graph, `ofm2` is not used for graph output but it is not dead node because it is essential virtual circle node.
As a result, it is not checked by current traversal and then it becomes to have invalid type.
To prevent this, this commit will change the traversal method.

### DIFF
--- a/compiler/luci/pass/src/CircleTypeInferencePass.cpp
+++ b/compiler/luci/pass/src/CircleTypeInferencePass.cpp
@@ -16,6 +16,7 @@
 
 #include "luci/Pass/CircleTypeInferencePass.h"
 
+#include <luci/IR/DeadNodeQueryService.h>
 #include <luci/Service/CircleTypeInference.h>
 
 #include <loco.h>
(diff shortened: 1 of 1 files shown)


seanshpark:
@llFreetimell , I've posted #5510 for which I'm not sure of current codes. Could you please check that too?

compiler/luci/pass/src/CircleTypeInferencePass.cpp
```diff
@@ -41,7 +41,7 @@ bool CircleTypeInferencePass::run(loco::Graph *g)
   luci::tinf::Rule type_infer_rule;
   bool changed = false;
 
-  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
+  for (auto node : loco::all_nodes(g))
```
seanshpark, line 44: From my past experience `all_nodes` may make visit to unwanted nodes.
As I see that `CircleTypeInferencePass` is not used as of now, we cannot detect any problems our test models.
We can land this change but please be cautious with future changes.

seanshpark (dismissed):
LGTM

compiler/luci/pass/src/CircleTypeInferencePass.cpp
llFreetimell: FYI, `CircleTypeInferencePass` and [`CircleShapeInferencePass`](https://github.com/Samsung/ONE/blob/0c264a717daf9ebac489428b048c6eceead195fa/compiler/luci/pass/src/CircleShapeInferencePass.cpp#L70)will be used soon because of #5501 ..!

llFreetimell:
This PR will be rebased and re-written after #5516 is merged

seanshpark (approved):
LGTM

seanshpark:
@llFreetimell , please check https://github.com/Samsung/ONE/pull/5529/files#r551080345

mhs4670go (approved):
LGTM

PR #4464, opened 2020-09-30
[luci-interpreter] Add support for sub(tract) op

a-zh:
This commit adds `sub` operator in luci-interpreter

ONE-DCO-1.0-Signed-off-by: Anton Zhukov <anton.v.zhukov@gmail.com>

PR #5509, opened 2020-12-30
[onert] Support Add int8 quantization

glistening:
It supports Add int8 quantization.

Also, it removes checking of same type for lhs, rhs for mul
because it is already done in OperationValidator.

### DIFF
--- a/runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
+++ b/runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
@@ -126,9 +126,6 @@ void setAddOrSubQuant8Params(const IPortableTensor *lhs, const IPortableTensor *
   op_params.input1_offset = -lhs->data_zero_point();
   op_params.input2_offset = -rhs->data_zero_point();
   op_params.output_offset = output->data_zero_point();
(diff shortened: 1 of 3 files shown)


runtime/onert/backend/cpu/ops/BinaryArithmeticLayer.cc
```diff
@@ -126,9 +126,6 @@ void setAddOrSubQuant8Params(const IPortableTensor *lhs, const IPortableTensor *
   op_params.input1_offset = -lhs->data_zero_point();
   op_params.input2_offset = -rhs->data_zero_point();
   op_params.output_offset = output->data_zero_point();
-  assert((op_params.input1_offset <= 0) && (op_params.input1_offset >= -255));
-  assert((op_params.input2_offset <= 0) && (op_params.input2_offset >= -255));
-  assert((op_params.output_offset >= 0) && (op_params.output_offset <= 255));
```
glistening, line 131: This assert is not valid for `int8`. Thus, it should be updated. However, I would like to check validity of zero_point in `OperationValidator`. I will do in another PR.

glistening:
#4464

ragmani (approved):
LGTM

hseok-oh (approved):
LGTM

PR #5513, opened 2020-12-30
[luci] Change default shape/type inference algorithm

llFreetimell:
Until now, default inference algorithm for shape and type was returning meaningless things.
By changing the default inference algorithm of shape and type, original inference rule of shape and type will be used,
if new shape/type inference algorithm of an operation are not implemented yet.


seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #5510, opened 2020-12-30
[luci] Problem with postorder_traversal() with dangling virtual nodes


seanshpark:
From PR #5507
```
              +-- CircleWhileOut(ofm1) -- CircleOutput
CircleWhile --+
              +-- CircleWhileOut(ofm2)
```
may have same problem with current type inference logic which uses below code;

```cxx
bool TypeInferenceSession::to(Graph *g) const
{
  bool changed = false;

  for (auto node : postorder_traversal(output_nodes(g)))
  {
    if (_rule->recognize(node->dialect()))
    {
      DataType dtype = DataType::Unknown;

      if (!dtype_known(node) && inputs_dtype_ready(node))
      {
        if (_rule->infer(node, dtype))
        {
          node->annot(stdex::make_unique<DataTypeAnnotation>(dtype));
          changed = true;
        }
      }
    }
  }

  return changed;
}
```

We may need to check this too.

seanshpark:
@llFreetimell , can you check this too?

llFreetimell:
In the above point of view, that code also cannot check `While_002` bug.
However, it works well for now because of following steps.

1. When `CircleWhileOut` is created, `shape` and `dtype` are copied. (circle node dtype : known / loco dtype : unknown)
https://github.com/Samsung/ONE/blob/2c71236a2982874fff2b501f4c8aebbb3d08a45a/compiler/luci/import/src/Nodes/CircleWhile.cpp#L110-L115

1. During `TypeInferencePass`, the `CircleWhileOut` is not checked. (circle node dtype : known / loco dtype : unknown)
https://github.com/Samsung/ONE/blob/2c71236a2982874fff2b501f4c8aebbb3d08a45a/compiler/loco/src/Service/TypeInference.cpp#L66

1. During `MigrateLegacyShapeDtypePass`, there is nothing to copy. (circle node dtype : known / loco dtype : unknown)
https://github.com/Samsung/ONE/blob/2c71236a2982874fff2b501f4c8aebbb3d08a45a/compiler/luci/pass/src/MigrateLegacyShapeDtypePass.cpp#L99-L105

1. When `CircleWhileOut` is exported, dtype of circle node is copied.
(FYI, before last change, `node_dtype()` in `TypeBridge.h` was used but it also used dtype of circle node)
[...]

PR #5516, opened 2020-12-30
[luci] Make DeadNodeQueryService visible in luci/lang

llFreetimell:
Parent Issue : #5510

There are some cases which needs `DeadNodeQueryService` but it was invisible until now.
This commit will make it visible and thus can be used in another modules.


mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #5521, opened 2020-12-31
[luci] Print unknown dimension as question mark

llFreetimell:
Parent Issue : #5501

Until now, unknown dimension was printed as 0.
This commit will fix it.

### DIFF
--- a/compiler/luci/service/src/CircleShapeInference.cpp
+++ b/compiler/luci/service/src/CircleShapeInference.cpp
@@ -46,7 +46,11 @@ std::ostream &operator<<(std::ostream &os, const loco::TensorShape &tensor_shape
   {
     if (r)
       os << ",";
-    os << tensor_shape.dim(r).value();
+
+    if (tensor_shape.dim(r).known())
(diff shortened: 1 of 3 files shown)


jinevening (approved):
LGTM

compiler/luci/service/src/CircleShapeInference.cpp
```diff
@@ -46,7 +46,11 @@ std::ostream &operator<<(std::ostream &os, const loco::TensorShape &tensor_shape
   {
     if (r)
       os << ",";
-    os << tensor_shape.dim(r).value();
+
+    if (tensor_shape.dim(r).known())
```
seanshpark, line 50: (optional) is below possible?
```cxx
os << (tensor_shape.dim(r).known() ? tensor_shape.dim(r).value() : "?");
```

llFreetimell: I tried it but failed because `value()` and `"?"` is different type :(

seanshpark (approved):
LGTM

PR #5522, opened 2020-12-31
[luci] Support unknown dimension expression in ShapeDescription

llFreetimell:
Parent Issue : #5501

Until now, unknown dimension was represented as 0.
This commit will fix it.

### DIFF
--- a/compiler/luci/service/src/ShapeDescription.cpp
+++ b/compiler/luci/service/src/ShapeDescription.cpp
@@ -31,7 +31,7 @@ ShapeDescription to_shape_description(const luci::CircleNode *circle_node)
 
   res._dims.resize(circle_node->rank());
   for (uint32_t i = 0; i < circle_node->rank(); ++i)
-    res._dims.at(i) = circle_node->dim(i).value();
+    res._dims.at(i) = circle_node->dim(i).known() ? circle_node->dim(i).value() : -1;
 
   return res;
 }


jinevening:
It seems that loco expresses unknown dimension as 0.

ONE/compiler/loco/include/loco/IR/Dimension.h

```cpp
 void unset(void) 
 { 
   _kind = Kind::Unknown; 
   _value = 0; 
 } 
```

CircleNode inherits `loco::NodeTrait::TensorShape`, so I think it is natural to use the same value (0).

ONE/compiler/luci/lang/include/luci/IR/CircleNodeDecl.h
```cpp
 struct CircleNode : public loco::Node, 
                     public loco::NodeMixin<loco::NodeTrait::DataType>, 
                     public loco::NodeMixin<loco::NodeTrait::TensorShape> 
```

Is there a reason not to use 0 for unknown dimension?

llFreetimell:
> loco expresses unknown dimension as 0.

Only half is true :)
When the dimension is unknown, `_value` is meaningless. More precisely, when the dimension is unknown, not using `_value` is recommended.
Nevertheless, `_value` must have some value. Since `_value` is `uint32_t`, 0 is selected for the value.

> CircleNode inherits loco::NodeTrait::TensorShape, so I think it is natural to use the same value (0).

Yes, if `CircleNode` has unknown dimension, the dimension would be `Unknown` and `_value` would be 0 because `CircleNode` interits `loco::NodeTrait::TensorShape` :)
This PR is modifying `ShapeDescription` which is used only for exporting circle model, so this PR do not influence to `CircleNode`.

> Is there a reason not to use 0 for unknown dimension?

One of the reason is that `loco` is using 0 as unknown dimension inevitably as I wrote above.
However, another important reason is that there is a case that 0 is used for known dimension.
Some examples are following things.

- When `input` is scalar and `tf.reduce_all(input, ...)` is exported as tflite file, `axis` of `ReduceAll` is [0].
- As I wrote in https://github.com/Samsung/ONE/issues/4793#issuecomment-717151253, using zero dimension in `Concat` is quite useful.

In conclusion, if we express unknown dimension as 0, we cannot distinguish with real zero dimension and unknown dimension.

jinevening:
I see. Thanks for the detailed explanation :)

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

PR #5525, opened 2020-12-31
[luci] Remove ShapeSignauture Inference

llFreetimell:
Parent Issue : #5501

Shape signature is no more needed and thus inference is not needed neither.
This commit will remove `ShapeSignatureInference` and `ShapeSignatureInferencePass`.


seanshpark (dismissed):
LGTM

compiler/luci/service/src/Nodes/CircleSum.cpp
```diff
@@ -14,15 +14,11 @@
...
-
 namespace luci
 {
 
-ShapeSignature ssinf::Algorithm::visit(const luci::CircleSum *node)
-{
-  return legalized_signature(
-    reduced_signature(node->input(), node->reduction_indices(), node->keep_dims()));
-}
+// Shape Inference Algorithm will be added
```
seanshpark, line 20: (optional) we may remove all similar files like this now and add them later when needed.

llFreetimell: I will apply it :)

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5527, opened 2020-12-31
[luci] Refactoring of ConvertNCHWToNHWCPass test

jinevening:
This performs refactoring of ConvertNCHWToNHWCPass test.

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
@@ -102,6 +102,7 @@ public:
 
 class AddGraph final : public SimpleGraph
 {
+protected:
   loco::Node *insertGraphBody(loco::Node *input) override
   {
     add = g.nodes()->create<luci::CircleAdd>();
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
@@ -30,7 +30,7 @@ namespace
 {
 
 /**
- *  Graph with a single Add Op.
+ *  Graph with a single Op (example: Add).
  *
  *  BEFORE
  *  - All Ops including Input/Output are NCHW.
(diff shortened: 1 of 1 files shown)


jinevening:
This makes `SimpleGraph` inheritable, allowing easy implementation of the test graph for each operator.

llFreetimell (approved):
LGTM

seanshpark (dismissed):
LGTM

compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
```diff
@@ -75,19 +76,44 @@ class SimpleGraph
...
+
+public:
+  loco::Graph g;
+  luci::CircleInput *input = nullptr;
+  luci::CircleOutput *output = nullptr;
+};
+
+class AddGraph final : public SimpleGraph
+{
+  loco::Node *insertGraphBody(loco::Node *input) override
```
seanshpark, line 105: default visibility is `private` if I'm correct. `insertGraphBody()`  is `protected` in base class.
is this your intention?

jinevening: insertGraphBody's visibility is protected as AddGraph inherits SimpleGraph in public.

My intention was to prohibit calling `insertGraphBody` in instances. I used protected just because it is a looser condition than private (using private is also ok in this test case).

seanshpark: I mean, scope is different. base `SimpleGraph` is declared as `protected` and `AddGraph` used as default (seems `private`). 
My suggestion is like this to be specific.
```cxx
class AddGraph final : public SimpleGraph
{
protected:
  loco::Node *insertGraphBody(loco::Node *input) override
```

jinevening: Oh, I see. It looks better. I've updated the code. Thanks.

seanshpark (approved):
LGTM thank you!

llFreetimell (approved):
LGTM

PR #5529, opened 2020-12-31
[luci] Introduce CircleShapeInferencePass

llFreetimell:
Parent Issue : #5501

This commit will introduce `CircleShapeInferencePass`, which uses `luci::sinf::Rule`.

### DIFF
--- a/compiler/luci/pass/src/CircleShapeInferencePass.cpp
+++ b/compiler/luci/pass/src/CircleShapeInferencePass.cpp
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/CircleShapeInferencePass.h"
+
+#include <luci/IR/DeadNodeQueryService.h>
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/CircleShapeInferencePass.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  return changed;
+}
+
+bool CircleShapeInferencePass::run(loco::Graph *g)
+{
+  luci::sinf::Rule shape_infer_rule;
+  bool changed = false;
+
+  for (auto node : loco::all_nodes(g))
```
seanshpark, line 70: @llFreetimell , can you please check,
- shape/dtype inference assumes topological order, that is input of the node inference is already done.
- `loco::all_nodes` may not be in topological order : when the graph is modified by some Pass.

llFreetimell: > `loco::all_nodes` may not be in topological order

Yes, it is not topological order.
However, shape inference will be eventually done :)

Before shape inference, inputs' status are checked first. If some of inputs are not ready(which means some of inputs are `ShapeStatus::UNDEFINED`), shape inference will not be done and `false` is returned.
https://github.com/Samsung/ONE/blob/38f825ae5e8d0cf2a90f6528e246699c89988b37/compiler/luci/service/src/CircleShapeInference.cpp#L84-L88
https://github.com/Samsung/ONE/blob/38f825ae5e8d0cf2a90f6528e246699c89988b37/compiler/luci/service/src/CircleShapeInference.cpp#L59-L69

So shape inference only for ready nodes are done and the iteration is finished.
Since phase strategy in `CircleOptimizer` is `RESTART` and type/shape inference pass is placed at the first, next iteration will be started soon.
https://github.com/Samsung/ONE/blob/38f825ae5e8d0cf2a90f6528e246699c89988b37/compiler/luci/pass/src/CircleOptimizer.cpp#L157
https://github.com/Samsung/ONE/blob/38f825ae5e8d0cf2a90f6528e246699c89988b37/compiler/luci/pass/src/CircleOptimizer.cpp#L245
In the next iteration, some nodes which are not ready in first iteration but ready in next iteration will be inferenced.
These steps will be repeated until all nodes are inferenced.

seanshpark (approved):
LGTM

jinevening (approved):
LGTM

PR #5534, opened 2020-12-31
[luci] Update shape/dtype validation in PostImport

llFreetimell:
Parent Issue : #5501

This commit will update shape and dtype validation in `PostImport` as following.

- CircleIf : Consider unknown dimension
- CircleWhile : Consider unknown dimension + Remove redundant checks

### DIFF
--- a/compiler/luci/import/src/PostImport.cpp
+++ b/compiler/luci/import/src/PostImport.cpp
@@ -130,7 +130,10 @@ private:
 namespace
 {
 /**
- * @brief  ValidateNodeProp will validate inter graph connections for each Nodes
+ * @brief  ValidateNodeProp will validate inter graph connections for each Nodes.
+ * @note   In here, only loco::GraphInput and loco::GraphOutput are validated,
+ *         since this class is for checking inter graph connections.
+ *         CircleNodes such as CircleInput and CircleOutput will be validated at later steps.
  */
 class ValidateNodeProp final : public luci::CircleNodeMutableVisitor<void>
 {
--- a/compiler/luci/import/src/PostImport.cpp
+++ b/compiler/luci/import/src/PostImport.cpp
@@ -172,9 +172,19 @@ public:
 
       auto then_graph_output = then_graph_outputs->at(then_out->index());
       auto else_graph_output = else_graph_outputs->at(else_out->index());
-      if (!(*then_graph_output->shape() == *else_graph_output->shape()))
+      if (then_graph_output->shape()->rank() != else_graph_output->shape()->rank())
       {
-        INTERNAL_EXN_V("CircleIf THEN and ELSE Graph Output shape mismatch ", idx);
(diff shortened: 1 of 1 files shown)


compiler/luci/import/src/PostImport.cpp
```diff
@@ -231,18 +241,20 @@ class ValidateNodeProp final : public luci::CircleNodeMutableVisitor<void>
 
       auto cond_graph_input = cond_graph_inputs->at(cond_in->index());
       auto body_graph_input = body_graph_inputs->at(body_in->index());
-      if ((cond_in->rank() != body_in->rank()))
+      if (cond_graph_input->shape()->rank() != body_graph_input->shape()->rank())
```
seanshpark, line 244: Why do you need to change this? `cond_in` and `cond_graph_input` is different.

compiler/luci/import/src/PostImport.cpp
```diff
@@ -231,18 +241,20 @@ class ValidateNodeProp final : public luci::CircleNodeMutableVisitor<void>
...
+        if (cond_graph_input->shape()->dim(i).known() &&
+            body_graph_input->shape()->dim(i).known() &&
+            cond_graph_input->shape()->dim(i).value() != body_graph_input->shape()->dim(i).value())
         {
-          INTERNAL_EXN_V("CircleWhile COND input and BODY input shape mismatch ", idx);
+          INTERNAL_EXN_V("CircleWhile COND input and BODY input dimension mismatch ", idx);
         }
       }
-      if (cond_in->dtype() != body_in->dtype())
+      if (cond_graph_input->dtype() != body_graph_input->dtype())
```
seanshpark, line 257: same here.

compiler/luci/import/src/PostImport.cpp
llFreetimell: AFAIK, `PostImport.cpp` is for checking the graph itself after all subgraphs are imported.
https://github.com/Samsung/ONE/blob/38f825ae5e8d0cf2a90f6528e246699c89988b37/compiler/luci/import/src/Importer.cpp#L289

Therefore, checking shape/dtype of `GraphInputContext` seems correct in `PostImport.cpp`.
Checking shape/dtype of `CircleNode` will be done at shape/type inference and `Validate.cpp`.

In addition, current codes are not consistent because shape/dtype of both `GraphInputContext` and `CircleNode` are used without any criteria.

seanshpark: Validation here is to check the shape/dtype of the nodes and graph; related subgraphs for `IF` and `WHILE`.
I don't actually understand what you wrote, thus I can't disgree or find any problems with it.
If you want to change, please add some comments as you understand and what you are checking and the conditions you think should be.

seanshpark (approved):
LGTM thank you!

jinevening (approved):
LGTM

Issue #5505, opened 2020-12-29
[onert] Removing OpSequence


wateret:
I have been thinking about removing `OpSequence` for months since I don't think it is worth it.

AFAIR, `OpSequence` was first introduced to help backend scheduling(by assigning consecutive operations as the same). However ther was no specific idea(my bad 😢 ). It does not help scheduling yet and it just works as a composite container of operations.

## Roles

- A unit of backend assignment
    - Has a layout property
- A unit of executing graphs(`FunctionSequence` is made from `OpSequence`)

## Pros

- For Parallel Executor, it checks the jobs(`FunctionSequence`) which are ready after every job is done
  and OpSequence reduces this overhead (Could be negligible in many cases though)
- (Maybe?) easy to fuse some operations as OpSequence is a kind of [Basic Block](https://en.wikipedia.org/wiki/Basic_block)

## Cons

- Redundant code and Maintainability issues
    - Makes writing transformation passes harder - need to care about not only operations/operands but also OpSequences
    - Most of things can be done with `Operation` but we have another object to do the same things
    - It is part of `LoweredGraph`, not `Graph` so implementation gets tricky sometimes
    - When execution, `OpSequence`(`FunctionSequence`) is the unit, and it makes harder to implement profilers

As I see we have much more cons than pros, I'm going to write a draft for this, and see what I am missing here. Do you have opinions?

cc @Samsung/nnfw @YongseopKim

wateret:
Done!

PR #5537, opened 2020-12-31
[onert] Remove OpSequence

wateret:
This commit removes `ir::Opsequence` and change the implementation to
be every operation has lower info. This is similar to when
`OPSEQ_MAX_NODE` is 1. Here are details:

- Remove `ir::OpSequence`
- Change `OpSequenceLowerInfo` to `OperationLowerInfo` - every operation
  has LowerInfo since there is no OpSequence
- Introduce `_has_dynamic_tensor_map` as an alternative for OpSequence
- Assume one graph can have only one type of tensor layout.
- Remove Config `OP_SEQ_MAX_NODE`
- FP32 to FP16 conversion is temporarilly disabled and will be
  reimplemented as a graph pass

By doing this, Implementation of many things(especially LoweredGraph
passes) are much easier with (almost) no disadvantage.

RELATED : #5505


wateret:

wateret:
#5557

PR #5538, opened 2021-01-04
[onert] Rename offset to zero_point of TypeInfo and makes CWQ ready

glistening:
: Rename ir::TypeInfo::offset to ir::TypeInfo::zero_point
: Make scale and zeropoint vector to store multiple values.


hseok-oh (approved):
LGTM

PR #5541, opened 2021-01-04
[luci] ConvertNCHWToNHWC Pass for MUL Op

jinevening:
This supports MUL in ConvertNCHWToNHWCPass.

---
Related to: #5401
Draft PR: #5438

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -160,11 +160,12 @@ luci::CircleConst *create_NHWC_from_NCHW(luci::CircleConst *constant)
   return nhwc_const;
 }
 
-// We assume MUL is NCHW if,
-// Input shape: (N, C, H, W)
-// Output shape: (N, C, H, W)
-// 1. Const shape is (1, C, 1, 1)
-// 2. Input, Output, Const have the same C.
+// NOTE Following conditions can be extended later
+//
+// Find MUL with an NCHW pattern described below
+//   - Input (non-constant) shape : [N, C, H, W]
+//   - Input (constant) shape : [1, C, 1, 1]
+//   - Output shape : [N, C, H, W]
 bool is_NCHW_with_const(const luci::CircleMul *node, luci::CircleNode *&pred_node,
                         luci::CircleConst *&multiplier)
 {
@@ -193,10 +194,7 @@ bool is_NCHW_with_const(const luci::CircleMul *node, luci::CircleNode *&pred_nod
 
   for (uint32_t i = 0; i < const_rank; i++)
   {
-    if (i == 1)
-      continue;
-
-    if (multiplier->dim(i).value() != 1)
+    if (i != 1 && multiplier->dim(i).value() != 1)
       return false;
   }
 
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -160,6 +160,56 @@ luci::CircleConst *create_NHWC_from_NCHW(luci::CircleConst *constant)
   return nhwc_const;
 }
 
+// We assume MUL is NCHW if,
+// Input shape: (N, C, H, W)
+// Output shape: (N, C, H, W)
+// 1. Const shape is (1, C, 1, 1)
+// 2. Input, Output, Const have the same C.
+bool is_NCHW_with_const(const luci::CircleMul *node, luci::CircleNode *&pred_node,
+                        luci::CircleConst *&gamma)
+{
+  auto x = dynamic_cast<luci::CircleConst *>(node->x());
+  auto y = dynamic_cast<luci::CircleConst *>(node->y());
+
+  if (x != nullptr && y == nullptr)
+  {
+    pred_node = loco::must_cast<luci::CircleNode *>(node->y());
+    gamma = x;
+  }
+  else if (x == nullptr && y != nullptr)
+  {
+    pred_node = loco::must_cast<luci::CircleNode *>(node->x());
+    gamma = y;
+  }
+  else
+  {
+    // Ignore if MUL does not have a gamma input.
+    return false;
+  }
+
+  const auto const_rank = gamma->rank();
+  if (const_rank != 4)
+    return false;
+
+  for (uint32_t i = 0; i < const_rank; i++)
+  {
+    if (i == 1)
+      continue;
+
+    if (gamma->dim(i).value() != 1)
+      return false;
+  }
+
+  const auto const_cdim = gamma->dim(1);
+  const auto input_cdim = pred_node->dim(1);
+  const auto output_cdim = node->dim(1);
+
+  if (const_cdim == input_cdim && input_cdim == output_cdim)
+    return true;
+  else
+    return false;
+}
+
 // We assume ADD with const input is NCHW if,
 // Input shape: (N, C, H, W)
 // Output shape: (N, C, H, W)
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -160,6 +160,56 @@ luci::CircleConst *create_NHWC_from_NCHW(luci::CircleConst *constant)
   return nhwc_const;
 }
 
+// We assume MUL is NCHW if,
+// Input shape: (N, C, H, W)
+// Output shape: (N, C, H, W)
+// 1. Const shape is (1, C, 1, 1)
+// 2. Input, Output, Const have the same C.
+bool is_NCHW_with_const(const luci::CircleMul *node, luci::CircleNode *&pred_node,
+                        luci::CircleConst *&gamma)
```
seanshpark, line 169: As this is reference to a pointer, how about `luci::CircleConst &*` for `pred_node and `gamma` ?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -316,6 +366,43 @@ class ConvertNCHWToNHWC final : public luci::CircleNodeMutableVisitor<bool>
     post_trans->a(node);
     return true;
   }
+
+  bool visit(luci::CircleMul *node)
+  {
+    LOGGER(l);
+
+    luci::CircleNode *pred_node = nullptr;
+    luci::CircleConst *gamma = nullptr;
```
seanshpark, line 375: Just curious, what use name `gamma`?

seanshpark: In https://en.wikipedia.org/wiki/Multiplication, there's `mutiplicand` and `multiplier`

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: Pointer to reference is not allowed in C/C++.

https://stackoverflow.com/questions/4632762/why-are-pointers-to-a-reference-illegal-in-c

seanshpark: > Pointer to reference is not allowed in C/C++.

I know this and wrote `reference to a pointer`. But is `&*` a pointer to reference?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: The name `gamma` originated from normalization operators.

This PR only supports a special form of MUL which might come from normalization Ops, so I used `gamma` instead of more general terms.

> In https://en.wikipedia.org/wiki/Multiplication, there's mutiplicand and multiplier

multiplier looks ok. I've changed gamma to multiplier.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
```diff
@@ -132,6 +132,38 @@ class AddGraph final : public SimpleGraph
   luci::CircleConst *beta = nullptr;
 };
 
+class MulGraph final : public SimpleGraph
+{
+protected:
+  loco::Node *insertGraphBody(loco::Node *input) override
+  {
+    mul = g.nodes()->create<luci::CircleMul>();
+    gamma = g.nodes()->create<luci::CircleConst>();
```
seanshpark, line 141: (optional) maybe change this too?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: I've learned it too. Thanks :smile:

compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
jinevening: Oh, I forgot this. Thanks :)

seanshpark (dismissed):
LGTM thank you!

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -160,6 +160,56 @@ luci::CircleConst *create_NHWC_from_NCHW(luci::CircleConst *constant)
   return nhwc_const;
 }
 
+// We assume MUL is NCHW if,
+// Input shape: (N, C, H, W)
+// Output shape: (N, C, H, W)
+// 1. Const shape is (1, C, 1, 1)
```
llFreetimell, line 166: What the numbers mean?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -160,6 +160,56 @@ luci::CircleConst *create_NHWC_from_NCHW(luci::CircleConst *constant)
...
+    multiplier = x;
+  }
+  else if (x == nullptr && y != nullptr)
+  {
+    pred_node = loco::must_cast<luci::CircleNode *>(node->x());
+    multiplier = y;
+  }
+  else
+  {
+    // Ignore if MUL does not have a multiplier input.
```
llFreetimell, line 186: What if both `x` and `y` are constant?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -160,6 +160,56 @@ luci::CircleConst *create_NHWC_from_NCHW(luci::CircleConst *constant)
...
+  if (const_rank != 4)
+    return false;
+
+  for (uint32_t i = 0; i < const_rank; i++)
+  {
+    if (i == 1)
+      continue;
+
+    if (multiplier->dim(i).value() != 1)
+      return false;
```
llFreetimell, line 200: (Optional)

```suggestion
    if (i != 1 && multiplier->dim(i).value() != 1)
      return false;
```

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: N: batch size
C: channel size
H: height
W: width

This PR only supports a special form of MUL which has the constant with the shape (1, C, 1, 1).

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: Yes, the case is not covered. If both x and y are constant, it would be better to be handled by constant folding (although we don't have such a logic yet).

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
llFreetimell: Ah, my focus was `1.` and `2.` :)

For me, it is little bit confused because I don't know these are some of conditions or just descriptions.

jinevening: The comment might have confused you. I think it would be better to change it.

How about this?

```cpp
// Find MUL with an NCHW pattern described below
// Input shape: (N, C, H, W)
// Output shape: (N, C, H, W)
// 1. Const shape is (1, C, 1, 1)
// 2. Input, Output, Const have the same C.
// The above condition can be extended later
bool is_NCHW_with_const(const luci::CircleMul *node, luci::CircleNode *&pred_node,
                        luci::CircleConst *&multiplier)
```

llFreetimell: If I understand correctly, how about following comment? :)
```
// NOTE Following conditions can be extended later
//
// Find MUL with an NCHW pattern described below
//   - Input (non-constant) shape : [N, C, H, W]
//   - Input (constant) shape : [1, C, 1, 1]
//   - Output shape : [N, C, H, W]
```

jinevening: Updated. Thanks.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: Updated.

llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

Issue #4664, opened 2020-10-19
[onert] NYI CPU backend kernel - int8


hseok-oh:
NYI cpu backend kernel: uint8 quantization type

- [ ] Add
- [ ] ArgMax
- [ ] ArgMin
- [ ] AvgPool2D
- [ ] BatchMatmul
- [ ] BatchToSpaceND
- [ ] BiDirectionalSequenceLSTM
- [ ] BiDirectionalSequenceRNN
- [ ] Cast
- [ ] Ceil
- [ ] Concat
- [ ] Conv2D
- [ ] DepthToSpace
- [ ] Densify
- [ ] DepthwiseConv2D
- [ ] Dequantize
- [ ] Div
- [ ] EmbeddingLookup
- [ ] Equal
- [ ] Exp
- [ ] ExpandDims
- [ ] Fill
- [ ] FullyConnected
- [ ] Gather
- [ ] GatherNd
- [ ] Greater
- [ ] GreaterEqual
- [ ] HardSwish
- [ ] HashtableLookup
- [ ] L2Normalization
- [ ] LeakyRelu
- [ ] Less
- [ ] LessEqual
- [ ] LocalResponseNormalize
- [ ] LogicalAnd
- [ ] LogicalOr
- [ ] Logistic
- [ ] LogSoftmax
- [ ] LSHProjection
- [ ] MatrixDiag
- [ ] MatrixSetDiag
- [ ] Maximum
- [ ] MaxPool2D
- [ ] Mean
- [ ] Minimum
- [ ] MirrorPad
- [ ] Mul
- [ ] NotEqual
- [ ] OneHot
- [ ] Pack
- [ ] Pad
- [ ] PadV2
- [ ] Pow
- [ ] PReLU
- [ ] Quantize
- [ ] Rank
- [ ] ReduceMax(Max)
- [ ] ReduceMin(Min)
- [ ] ReduceProd
- [ ] ReduceSum(Sum)
- [ ] ReLU
- [ ] RELU_N1_TO_1
- [ ] ReLU6
- [ ] Reshape
- [ ] ResizeBilinear
- [ ] ResizeNearestneighbor
- [ ] RNN
- [ ] ScatterND
- [ ] Select
- [ ] SelectV2
- [ ] Shape
- [ ] Slice
- [ ] Softmax
- [ ] SpaceToBatchND
- [ ] SpaceToDepth
- [ ] SparseToDense
- [ ] Split
- [ ] SplitV
[...]

PR #5544, opened 2021-01-04
[onert] base_loader loads channel-wise quantization params

glistening:
`base_loader` can load array of scales and zero_points.

### DIFF
--- a/runtime/onert/frontend/base_loader/include/base_loader.h
+++ b/runtime/onert/frontend/base_loader/include/base_loader.h
@@ -389,35 +389,36 @@ template <typename LoaderDomain>
 void BaseLoader<LoaderDomain>::loadQuantization(const Tensor *tensor, ir::TypeInfo &typeInfo)
 {
   auto q_params = tensor->quantization();
-  float scale = 0.0;
-  long zero_point = 0;
-  if (q_params != nullptr)
+  if (q_params == nullptr || q_params->scale() == nullptr || q_params->scale()->size() == 0)
   {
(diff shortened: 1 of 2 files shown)


glistening:
Related: #4664

hseok-oh (approved):

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -389,35 +389,36 @@ template <typename LoaderDomain>
...
+  std::vector<float> scales;
+  std::vector<int32_t> zero_points;
+  scales.resize(num_scales);
+  zero_points.resize(num_scales);
+  for (size_t i = 0; i < num_scales; ++i)
+  {
+    scales[i] = q_params->scale()->Get(i);
+    int32_t zero_point = q_params->zero_point()->Get(i);
+    assert(zero_point >= std::numeric_limits<int32_t>::min());
+    assert(zero_point <= std::numeric_limits<int32_t>::max());
```
wateret, line 415: (Out of scope) Wouldn't these checks redundant?  I can't come up with a case that `zero_point` can be less than the min value of int32, and same for the max case.

wateret (approved):

runtime/onert/frontend/base_loader/include/base_loader.h
glistening: @wateret You're right. It is my mistake. It should be stored in `long` first and checked in the way before this PR.

PR #5550, opened 2021-01-05
[luci-interpreter] Remove overflow possibility

llFreetimell:
In theory, if `outer_dims` and `quant_dim_size` is really big, `outer_dims * quant_dim_size` can be overflowed.
This commit will fix it.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Prelu.cpp
+++ b/compiler/luci-interpreter/src/kernels/Prelu.cpp
@@ -197,7 +197,9 @@ void Prelu::evalQuantizedS16() const
     for (int32_t quant_channel = 0; quant_channel < quant_dim_size; ++quant_channel)
     {
       const ChannelQuantMultipliers &neg_mult = _alpha_multipliers[quant_channel];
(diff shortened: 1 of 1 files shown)


wateret:
I'm not familiar with this module, `size_t` is usually 32 bits in a 32-bit machine. Then this would have almost no effect on preventing overflows. (I don't know the compiler is always run on 64-bit machines, though)

seanshpark:
> I don't know the compiler is always run on 64-bit machines, though

We only release x86-64 version. Users may build our compiler frontend with 32 but as our compiler works with TensorFlow and this TensorFlow runs (only) in 64 bit. 32 bit TensorFlow is quite difficult to build.

llFreetimell:
@wateret FYI,

This PR is created because of static warning which alerts something like "Overflowed value(negative) cannot be converted to `size_t`". I am not sure "Overflow will never occurred by this PR", but I am expecting to remove that warning :)

jinevening (approved):

struss (approved):
LGTM

PR #5552, opened 2021-01-05
[onert] disable fp32 to fp16 conversion (temp)

wateret:
Temporarily disable fp32 to fp16 conversion.

Draft : #5537 (it explains why we need this change)


YongseopKim (approved):
LGTM

Issue #5548, opened 2021-01-05
Compiler FE: import onnx files


seanshpark:
Lets support onnx file to circle. 
- [x] prepare simple onnx test file
- [x] check onnx-tensorflow package
- [x] add `one-import-onnx` command
   - [x] add tests for one-import-onnx
   - [x] `one-build` with onnx
- [x] Op conversion verification
   - [x] Conv2D - Conv2D network
- Fix graph input/output name alteration --> #5706

seanshpark:
### prepare simple onnx test file

PyTorch install (v1.7.0)
- from https://pytorch.org/get-started/locally/
```
pip install torch==1.7.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
```

script to create a test onnx file
```py
import torch


class TestModel(torch.nn.Module):
  def __init__(self):
    super(TestModel, self).__init__()

    self.conv1 = torch.nn.Conv2d(3, 16, 3)
    self.relu1 = torch.nn.ReLU()
    self.conv2 = torch.nn.Conv2d(16, 3, 3)
    self.relu2 = torch.nn.ReLU()
    self.pool = torch.nn.MaxPool2d(2)

  def forward(self, input):
    output = self.conv1(input)
    output = self.relu1(output)
    output = self.conv2(output)
    output = self.relu2(output)
    output = self.pool(output)
    return output


model = TestModel()
torch.save(model, "model.pth")

input = torch.randn(1, 3, 224, 224)
torch.onnx.export(model, input, "model.onnx", verbose=True,
                  input_names=["input"],
                  output_names=["output"])
```

seanshpark:
Closing this issue and added #5706 for the issue.

PR #5553, opened 2021-01-05
DRAFT Compiler FE: Enable import ONNX files

seanshpark:
on-going draft to support onnx files

ONE-DCO-1.0-ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@samsung.com>


seanshpark:
For #5548

seanshpark:
@hseok-oh , I need to update `one-prepare-venv` install packages and for this, so could you please tell if does our CI need updates too?

hseok-oh:
@seanshpark We don't need to update CI because CI always call `one-prepare-venv` on package test without caching.

seanshpark:
done for now

PR #5554, opened 2021-01-05
[luci] Make dynamic shape determinant function as public

llFreetimell:
`has_dynamic_shape` in `ConvertNCHWtoNHWCPass` is for determining whether a node has dynamic shape.
Since some of other passes also needs this function, this commit will make it as public.


jinevening (approved):
LGTM

seanshpark (approved):
LGTM

PR #5555, opened 2021-01-05
[onert/cpu] Introduce getBuffer function

hseok-oh:
This commit introduce getBuffer function to get tensor's buffer pointer

### DIFF
--- a/compute/cker/include/cker/operation/Fill.h
+++ b/compute/cker/include/cker/operation/Fill.h
@@ -24,7 +24,8 @@ namespace nnfw
 {
 namespace cker
 {
-template <typename T> inline void Fill(const T value_data, const Shape &output_shape, T output_data)
+template <typename T>
+inline void Fill(const T *value_data, const Shape &output_shape, T *output_data)
 {
   int output_size = output_shape.FlatSize();
   for (int i = 0; i < output_size; i++)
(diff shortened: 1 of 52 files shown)


wateret (approved):
LGTM

glistening (approved):
LGTM.

Beyond this PR. But two ideas popped into my head.

1. `getBuffer()` looks good. Then, we may use `getShape()` than `getTensorShape()`.

2. Why do we define the utility functions out of `IPortableTensor`? Why don't we define those in `IPortableTensor`? For example,

```diff
 switch (_output->data_type())
 {
   case OperandType::FLOAT32:
-    nnfw::cker::Fill<float *>(reinterpret_cast<float *>(_value->buffer()),
-                              getTensorShape(_output),
-                              reinterpret_cast<float *>(_output->buffer()));
+    nnfw::cker::Fill<float *>(_value->buffer<float>(), _output->shape(), _output->buffer<float>());
     break;
   case OperandType::INT32:
     nnfw::cker::Fill<int32_t *>(reinterpret_cast<int32_t *>(_value->buffer()),
```

hseok-oh:
>we may use `getShape()` than `getTensorShape()`.

Agree. I also have same idea.

hseok-oh:
>Why don't we define those in `IPortableTensor`? 

It needs more interface change, `IPortableTensor` and related classes, and `ITensor` if need. I'll leave this as issue.

glistening:
@hseok-oh 

I prefer the style of `after` than `before` since it is more readable for me.

```diff
-    nnfw::cker::Fill<float *>(getBuffer<float>(_value), getTensorShape(_output), getBuffer<float>(_output));
+    nnfw::cker::Fill<float *>(_value->buffer<float>(), _output->shape(), _output->buffer<float>());
```

However, it is just a taste of mine. I will not make PR or issue. Thank you for your opinion.

PR #5557, opened 2021-01-05
[onert] Remove OpSequence

wateret:
This commit removes `ir::Opsequence` and change the implementation to
be every operation has lower info. This is similar to when
`OPSEQ_MAX_NODE` is 1. Here are details:

- Remove `ir::OpSequence`
- Change `OpSequenceLowerInfo` to `OperationLowerInfo` - every operation
  has LowerInfo since there is no OpSequence
- Introduce `_has_dynamic_tensor_map` as an alternative for OpSequence
- NOTE Assumes one graph can have only one type of tensor layout

By doing this, Implementation of many things(especially LoweredGraph
passes) are much easier with (almost) no disadvantage.


wateret:
Sorry about the large diff, but as it is removing a class that is used widely, it is hard to separate.

wateret:
Draft: #5557

hseok-oh (approved):
LGTM

glistening (approved):
LGTM

YongseopKim (approved):
LGTM

PR #5561, opened 2021-01-05
[luci] ConvertNCHWToNHWC Pass for PAD Op

jinevening:
This supports PAD in ConvertNCHWToNHWCPass.

---
Related to: #5401
Draft PR: #5438

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -113,6 +113,43 @@ uint32_t cal_offset(const loco::TensorShape &dimension, const uint32_t *indices)
          indices[2] * dimension.dim(3).value() + indices[3];
 }
 
+luci::CircleConst *create_NHWC_paddings(luci::CircleConst *paddings)
+{
+  // paddings shape is (4,2) (it was checked by is_NCHW)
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -113,6 +113,43 @@ uint32_t cal_offset(const loco::TensorShape &dimension, const uint32_t *indices)
...
+        // get third dimension (H in NCHW)
+        data = paddings->at<loco::DataType::S32>(2 * 2 + i);
+      }
+      else if (dim == 2)
+      {
+        // get fourth dimension (W in NCHW)
+        data = paddings->at<loco::DataType::S32>(3 * 2 + i);
+      }
+
+      nhwc_paddings->at<loco::DataType::S32>(dim * 2 + i) = data;
```
mhs4670go, line 147: It seems that dim 0 and 3 will have data zero. Is it right?

jinevening: Yes, it is intended. Their values must be 0.

I'll add assert that checks the original data is also 0 for clarity.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5562, opened 2021-01-05
[luci] Update shape validation code

llFreetimell:
Parent Issue : #5501

Until now, unknown dimension is not considered at shape validation.
This commit will apply those consideration to shape validation codes.

### DIFF
--- a/compiler/luci/service/src/Validate.cpp
+++ b/compiler/luci/service/src/Validate.cpp
@@ -107,10 +107,27 @@ bool validate_shape_dtype(loco::Graph *g)
     auto go_tensor_shape = graph_out->shape();
     assert(go_tensor_shape);
 
+    // NOTE Even if shape of graph output is [] (which means "shape inference was impossible")
+    //      but shape of CircleNode is not, it can be valid case because shape inference
+    //      algorithm of CircleNode may be upgraded than before. The opposite is possible either.
+    //      If such cases are appeared, following validation code should be fixed.
     bool is_shape_valid = (circle_node->rank() == go_tensor_shape->rank());
     for (uint32_t i = 0; is_shape_valid && i < circle_node->rank(); ++i)
+    {
       if (circle_node->dim(i).value() != go_tensor_shape->dim(i).value())
-        is_shape_valid = false;
+      {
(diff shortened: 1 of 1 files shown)


compiler/luci/service/src/Validate.cpp
```diff
@@ -107,10 +107,27 @@ bool validate_shape_dtype(loco::Graph *g)
...
+    //      but shape of CircleNode is not, it can be valid case because shape inference
+    //      algorithm of CircleNode may be upgraded than before. The opposite is possible either.
+    //      If such cases are appeared, following validation code should be fixed.
     bool is_shape_valid = (circle_node->rank() == go_tensor_shape->rank());
     for (uint32_t i = 0; is_shape_valid && i < circle_node->rank(); ++i)
+    {
       if (circle_node->dim(i).value() != go_tensor_shape->dim(i).value())
-        is_shape_valid = false;
+      {
+        if (!circle_node->dim(i).known() || !go_tensor_shape->dim(i).known())
```
mhs4670go, line 119: Just curiosity, is it normal to check if dim is known before comparing its value?

llFreetimell: Logic has no problem but it may cause unintended confusion, I will fix it :)

compiler/luci/service/src/Validate.cpp
```diff
@@ -114,18 +114,16 @@ bool validate_shape_dtype(loco::Graph *g)
     bool is_shape_valid = (circle_node->rank() == go_tensor_shape->rank());
     for (uint32_t i = 0; is_shape_valid && i < circle_node->rank(); ++i)
     {
+      if (!circle_node->dim(i).known() || !go_tensor_shape->dim(i).known())
+      {
+        // If at least one of two dimensions is unknown,
+        // the unknown dimension can accept any value.
+        INFO(l) << "Unknown dimension is matched with known dimension" << std::endl;
+      }
+
```
seanshpark, line 123: `else if` ?

llFreetimell: Oops...!

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #5564, opened 2021-01-05
[luci] Support unknown dimension in SubstitutePackToReshapePass

llFreetimell:
Parent Issue : #5501

Until now, unknown dimension is not considered in `SubstitutePackToReshapePass`.
However, if `Pack` operation has less or equal than one unknown dimension,
the pass is still applicable.
This commit will support it.


struss (approved):
Thanks! I miss that point

seanshpark (approved):
LGTM

PR #5565, opened 2021-01-05
[luci] Remove redundant constant creation in ResolveCustomOpMatMulPass

llFreetimell:
Parent Issue : #5501

If there is no `bias` in `FullyConnected`, there are two ways to represent it.

- Create constant node and fill it with 0
- Create `CircleOutputExclude` and do nothing

Until now, first way is used but it is inefficient and unknown dimension should be considered.
Therefore this commit will change it to second way, which is more efficient and simple.


seanshpark (dismissed):
LGTM

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5568, opened 2021-01-05
[luci] Remove shape/dtype setting in FusePreActivationBatchNormPass

llFreetimell:
Parent Issue : #5501

Until now, shape setting for `CircleSub` was done in `insert_sub` function.
However, this is redundant and inefficient because of following reasons.

- `ShapeInferencePass` exists exactly for the same purpose and more precise.
- If there are some unknown dimension in `pred`, this is not handled.

Therefore this commit will remove it.


seanshpark (approved):
LGTM

jinevening (approved):
LGTM Thanks

PR #5569, opened 2021-01-05
[one-cmds] Add one-import-onnx to cfg template

seanshpark:
This will add one-import-onnx item as False to config template.


seanshpark:
For #5548, from draft #5553

mhs4670go (approved):
LGTM

PR #5573, opened 2021-01-06
[res] Update rule for Net_InstanceNorm_002

llFreetimell:
As of this commit, `Net_InstanceNorm_002` has three `CircleReshape` nodes.
However, those `CircleReshape` can be removed by optimizations.
If we do not fix current rule, which `CircleReshape` should be 3,
such optimization may cause an error even the optimizatino result is correct.
This commit will update the rule to prevent those situation.

### DIFF
--- a/res/TensorFlowLiteRecipes/Net_InstanceNorm_002/test.rule
+++ b/res/TensorFlowLiteRecipes/Net_InstanceNorm_002/test.rule
@@ -3,6 +3,6 @@
 RULE    "VERIFY_FILE_FORMAT"      $(verify_file_format) '=' 1
 
 RULE    "INSTANCE_NORM_EXIST"     $(op_count INSTANCE_NORM) '=' 1
-RULE    "RESHAPE_EXIST"           $(op_count RESHAPE) '=' 3
+RULE    "RESHAPE_EXIST"           $(op_count RESHAPE) '<=' 3
 RULE    "NO_ADD"                  $(op_count ADD) '=' 0
 RULE    "NO_MUL"                  $(op_count MUL) '=' 0


seanshpark (approved):
LGTM

jinevening:
> However, those CircleReshape can be removed by optimizations.

Why this happens? dredd test will call circle2circle with `--fuse_instnorm` (not with reshape optimizations). Are you going to change fuse_instnorm pass?

llFreetimell:
I only planned to change `FuseInstanceNormPass` to remove following redundant `CircleReshape` creation.
https://github.com/Samsung/ONE/blob/4359af75b2b64c294ba6f900f58a7b45cf345921/compiler/luci/pass/src/FuseInstanceNormPass.cpp#L647-L661

However, #5576 may affect it's dredd test too @.@ I didn't know #5576 is in progress when I created this PR...!

jinevening (approved):
LGTM

mhs4670go:
IIUC, you will change `FuseInstanceNormPass` and it will generate one less reshape. But could you explain more detail about dangerous `dim(3).value()` access? I mean the things make it dangerous

llFreetimell:
> you will change FuseInstanceNormPass and it will generate one less reshape.

two reshape will be removed, one for `const_as_gamma` and the other for `const_as_beta`.

> But could you explain more detail about dangerous `dim(3).value()` access?

Currently, `dim(3).value()` is accessed without considering unknown dimension.
If `dim(3).known()` is `false`, `dim(3).value()` would be 0 and `ifm_channel_depth` becomes 0.
It may cause unintended behavior and I said this is **dangerous** :)

mhs4670go (approved):
LGTM

PR #5574, opened 2021-01-06
[luci] Encode shape with consideration of unknown dimension

llFreetimell:
Parent Issue : #5501

Current shape encoding do not consider unknown dimensions,
but unknown dimensions, which are represented as -1 in ShapeDescription,
should be exported as 1.
This commit will fix it.

### DIFF
--- a/compiler/luci/export/src/CircleTensorExporter.cpp
+++ b/compiler/luci/export/src/CircleTensorExporter.cpp
@@ -237,7 +237,13 @@ flatbuffers::Offset<Vector<int32_t>> encodeShape(FlatBufferBuilder &builder,
                                                  const ShapeDescription &shape)
 {
   assert(shape._rank_known && "unknown number of dimensions is not supported");
-  return builder.CreateVector(shape._dims);
+
+  std::vector<int32_t> encoded_shape;
+  encoded_shape.resize(shape._dims.size());
+  for (uint32_t i = 0; i < shape._dims.size(); ++i)
+    encoded_shape.at(i) = shape._dims.at(i) == -1 ? 1 : shape._dims.at(i);
+
+  return builder.CreateVector(encoded_shape);
 }
 
 flatbuffers::Offset<Vector<int32_t>> encodeShapeSignature(FlatBufferBuilder &builder,


seanshpark:
Is the `-1` written in shape signature?

llFreetimell:
@seanshpark From `CircleNode` :)

When `CircleNode` has unknown dimension, -1 is inserted into `ShapeDescription`.

https://github.com/Samsung/ONE/blob/4359af75b2b64c294ba6f900f58a7b45cf345921/compiler/luci/export/src/CircleTensorExporter.cpp#L117

https://github.com/Samsung/ONE/blob/4359af75b2b64c294ba6f900f58a7b45cf345921/compiler/luci/service/src/ShapeDescription.cpp#L26-L37

seanshpark:
> When CircleNode has unknown dimension, -1 is inserted into ShapeDescription.

This is not exactly what I wanted to check... It's a bit difficult to make a question about this...
It's something like, if there is `-1`, which tells this dim is unknown, if this patch changes the value previously written as `-1` to `1` then how do we know the file shape is unknow or it's really `1`?

llFreetimell:
@seanshpark 

> This is not exactly what I wanted to check... It's a bit difficult to make a question about this...

I misunderstood your question, sorry for this :(

> if there is -1, which tells this dim is unknown, if this patch changes the value previously written as -1 to 1 then how do we know the file shape is unknow or it's really 1?

As I wrote in https://github.com/Samsung/ONE/issues/5501#issue-775808564 as a diagram, `shape_signature` will be exported too :) (Some modification will be applied soon)

If shape of `CircleNode` was [?, 2, 3, 1], first dimension is unknown.
Then `ShapeDescription` of the node would be [-1, 2, 3, 1].
When we export this node, `encodeShape` will return [1, 2, 3, 1] (by this PR) and `encodeShapeSignature` will return [-1, 2, 3, 1].
(Note that `encodeShape` do not change `ShapeDescription`)
As a result,  exported file would have both [1, 2, 3, 1] shape and [-1, 2, 3, 1] shape_signature.

> I'm practicing not to meet off-line for pandemic work home :)

:smiley_cat:

seanshpark:
So it is yes to my first question `Is the -1 written in shape signature?` :)

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5578, opened 2021-01-06
[one-cmds] Revise one-build for import onnx

seanshpark:
This will revise one-build command to support onnx import.


seanshpark:
For #5548, from draft #5553

mhs4670go (approved):
LGTM

PR #5579, opened 2021-01-06
[common-artifacts] Exclude recipes which include dynamic shape

llFreetimell:
Current luci-interpreter do not generate correct test data for dynamic shape.
This commit will exclude resipes which include dynamic shape.

### DIFF
--- a/compiler/common-artifacts/exclude.lst
+++ b/compiler/common-artifacts/exclude.lst
@@ -60,6 +60,9 @@ tcgenerate(MatrixSetDiag_000)
 tcgenerate(MaxPoolWithArgMax_000)
 tcgenerate(MaxPoolWithArgMax_001)
 tcgenerate(MaxPoolWithArgMax_002)
+tcgenerate(Mean_dynamic_000) # luci-interpreter doesn't support dynamic shape
+tcgenerate(Mean_dynamic_001) # luci-interpreter doesn't support dynamic shape
+tcgenerate(Mean_U8_dynamic_000) # luci-interpreter doesn't support dynamic shape
 tcgenerate(NonMaxSuppressionV4_000)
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

seanshpark:
@llFreetimell , by the comment, I thought there was new items to exclude list.
But also, items with comments added about the dynamic shape. This seemed somewhat confusing.

Next time, please sync description and the changes.
As usual, it would be better to add one type of change in one PR

jinevening:
There is a code base for luci-interpreter to support dynamic shape (https://github.com/Samsung/ONE/pull/3051). 

Is this PR for your new changes on dynamic shape?

llFreetimell:
@jinevening 

> There is a code base for luci-interpreter to support dynamic shape (#3051).

Sorry, word selection was bad :(
luci-interpreter supports well but `TestDataGenerator` does not.
As I thought that `TestDataGenerator` is strongly connected with luci-interpreter, I wrote a wrong sentence.
So "Current **luci-interpreter** do not generate correct test data for dynamic shape." must be fixed to "Current **TestDataGenerator** do not generate correct test data for dynamic shape"

> Is this PR for your new changes on dynamic shape?

Yes, but I think that this PR is not needed...

At first, I found that even though current `TestDataGenerator` is incorrect, current tests are all passes by luck. (Really lucky recipes...)
So I planned following steps.

- Disable current dynamic shape recipes
- Fix `TestDataGenerator` to support dynamic shape correctly
- Enable current dynamic shape recipes

However, disabling and enabling current dynamic shape recipes may be redundant since current tests are all passed...
So I will close this PR and just fix `TestDataGenerator` to support dynamic shape correctly!

Thanks for your opinion :)

PR #5580, opened 2021-01-06
[onert] Remove LowerInfo get/setters

wateret:
This commit removes all LowerInfo get/setters except `GraphLowerInfo`.
With those methods, there were two ways of getting or setting
Operand/Operation LowerInfo.

- Get `GraphLowerInfo` and access with this object
- Use get/setter methods (`LoweredGraph::getLowerInfo(...)` overloads)

I felt like the second ones are redundant wrappers, so this commit
removes those in order to force just one way to do things.

Also this commit contains adding some methods to `util::ObjectManager`.

- `set` : Insert or assign
- `getRawPtr` : get the raw pointer of the object


wateret:
I will add some unit tests for new methods ObjectManager, after merging this.

YongseopKim (approved):
LGTM

PR #5582, opened 2021-01-06
[one-cmds] Fix one-prepare-venv for onnx-tf

seanshpark:
This will fix one-prepare-venv for onnx-tf where current variable check is not correct.


seanshpark:
For #5548, from draft #5553

mhs4670go (approved):
LGTM

jinevening (approved):
LGTM

PR #5583, opened 2021-01-06
[luci] Update logic of FuseInstanceNormPass

llFreetimell:
Related Issue : #5501

This commit will update following things in `FuseInstanceNormPass`.

- If some dimension is unknown, `FuseInstanceNormPass` cannot be applied
- Reshape of `CircleConst` creates redundant `CircleReshape`. Thus just reshape the `CircleConst`

### DIFF
--- a/compiler/luci/pass/src/FuseInstanceNormPass.cpp
+++ b/compiler/luci/pass/src/FuseInstanceNormPass.cpp
@@ -454,6 +454,7 @@ bool InstanceNormPattern::matched()
 
     luci::CircleNode *ifm_node = loco::must_cast<luci::CircleNode *>(ifm);
     CHECK_OR_FALSE(ifm_node->rank() == 4);
+    CHECK_OR_FALSE(ifm_node->dim(3).known());
     uint32_t ifm_channel_depth = ifm_node->dim(3).value();
 
     mean_of_ifm = dynamic_cast<luci::CircleMean *>(sub->y());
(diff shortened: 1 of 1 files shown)


mhs4670go:
@cgbahk 
```cxx
/**
 * Instance norm pattern would be fused like following diagram:
 *
 *    [In] --------------------------- CircleInstanceNorm --- [Out]
 *                                     / /
 *    const_as_gamma --- TFLReshape --- /
 *                                     /
 *    const_as_beta ---- TFLReshape ---
 *
 * Note
 *  - 'const_as_gamma' and 'const_as_beta' are from original graph
 *  - Value of 'const_as_epsilon' would be copied to CircleInstanceNorm's attribute
 *  - TFLReshape is added as CircleInstanceNorm only accept 1D tensor
 *  - 'TFLConst --- TFLReshape' is expected to be fused in constant folding for Reshape
 */
```
This comment is really old.. but I have a question.
> TFLReshape is added as CircleInstanceNorm only accept 1D tensor

`const_as_gamma` seems to have 1D shape by itself. Could you let me know why it needs `TFLReshape`?

cgbahk:
> `const_as_gamma` seems to have 1D shape by itself. Could you let me know why it needs `TFLReshape`?

I'm not sure but I guess I was dealing with `1x1x.. xD` constant tensor for gamma and beta.
(nnfw 9103)

mhs4670go:
Hmmm.. 
> gamma : 1-dimensional tensor of size C

The spec says it is 1-D tensor but, as @cgbahk said, sometimes it might be higher dimension.

@llFreetimell Could we remove this reshape op?

llFreetimell:
@mhs4670go 
This PR is for removing `CircleReshape`, not for removing reshaping itself :)
As @cgbahk said, there may be some cases that higher dimension appears.
Nevertheless, we can reshape the higher dimension node just by modifying rank and dimension of `CircleConst` :)

mhs4670go (approved):
LGTM

jinevening (approved):
LGTM

PR #5584, opened 2021-01-06
[circle2circle] Add option for ConvertNCHWToNHWC pass

jinevening:
This adds option for ConvertNCHWToNHWC pass.

---
Related to: #5401


seanshpark:
Adding this option, we may assume that all supported operators are ready and supports the change. Is it ready?
Or if this change is to early enable this feature to find out what is needed more, please add something like "Experimental: " prefix to the description so that we know it may not work correctly for all cases...

jinevening:
> Is it ready?

It is ready only for supported operators. I've added 'Experimental' prefix.

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

PR #5587, opened 2021-01-06
[onert] Refine LoweredGraph ctor

wateret:
- Remove 2 private methods that were called by the constructor only
    - And those are embedded in the constructor
- Remove `operands_lower_info` which was a temporary holder
- Remove some redundant code
- Move the mandatory passes(that were run in the middle) to the end
  (Now all 4 mandatory passes run at the same point)

### DIFF
--- a/runtime/onert/core/include/compiler/LoweredGraph.h
+++ b/runtime/onert/core/include/compiler/LoweredGraph.h
@@ -54,11 +54,7 @@ public:
   }
 
 private:
-  void makeOperationLowerInfo(
-    ir::OperandIndexMap<std::unique_ptr<compiler::OperandLowerInfo>> &operands_lower_info,
-    const compiler::BackendResolver &backend_resolver);
-  void manipulateLowerInfo(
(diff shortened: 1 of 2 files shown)


wateret:

wateret:

runtime/onert/core/include/compiler/LoweredGraph.h
```diff
@@ -54,11 +54,6 @@ class LoweredGraph
   }
 
 private:
-  void makeOperationLowerInfo(
-    ir::OperandIndexMap<std::unique_ptr<compiler::OperandLowerInfo>> &operands_lower_info,
-    const compiler::BackendResolver &backend_resolver);
-  void manipulateLowerInfo(
-    ir::OperandIndexMap<std::unique_ptr<compiler::OperandLowerInfo>> &operands_lower_info);
```
hseok-oh, line 61: Doesn't it make SAM score down ?

wateret: Well, then I will create a method for `LoweredGraph.cc:90-147`. But this time it is going to be just one, no need to be two 😃

hseok-oh (approved):
LGTM

PR #5589, opened 2021-01-07
[onert] Fix acl-neon backend softmax

hseok-oh:
Fix test fail by calling acl-neon backend disableDimCorrection call for softmax operation


wateret (approved):
LGTM

PR #5594, opened 2021-01-07
[luci-interpreter] Set a value for unknown dimension

llFreetimell:
Parent Issue : #5501

Until now, if node dimension is unknown, 0 is set for the dimension but this is incorrect.
This commit will set the value as 1 because -1 in shape_signature is represented as 1 in shape.

### DIFF
--- a/compiler/luci-interpreter/src/loader/GraphLoader.cpp
+++ b/compiler/luci-interpreter/src/loader/GraphLoader.cpp
@@ -30,7 +30,11 @@ template <typename NodeT> Shape getNodeShape(const NodeT *node)
   Shape shape(node->rank());
   for (uint32_t i = 0; i < node->rank(); ++i)
   {
-    shape.dim(i) = node->dim(i).value();
+    // NOTE In theory, unknown dimension can be any value.
+    //      However, we are not sure about the side effect for setting other than 1.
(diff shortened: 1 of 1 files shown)


jinevening:
luci-interpreter copies the shape of circle node to the shape of `Tensor`. Is it necessary to change this principle? Setting hard-coded values looks not good..

If circle node has a correct shape for unknown dimension (not 0, maybe 1?), then there will be no problem.

llFreetimell:
> luci-interpreter copies the shape of circle node to the shape of Tensor. Is it necessary to change this principle?

This PR does not change the principle :-D
Just for fixing wrong value copying...!

> Setting hard-coded values looks not good..

I thought about it a lot because unknown dimension can accept any value.
So if we want to allow users to set the unknown dimension value, it can be best solution but it requires lots of changes :(
Therefore, I decided to set the value as 1 for now...

> If circle node has a correct shape for unknown dimension (not 0, maybe 1?)

Yes, this PR is setting as 1 :)

llFreetimell:
The result of discussion

- It is better to supporting unknown dimension as other expression rather than this PR
- If `getNodeShape` is used for other operations, it may cause confusion
- This is not urgent feature

So close this PR :)

PR #5595, opened 2021-01-07
[luci-interpreter] Introduce output tensor shape getter

llFreetimell:
Parent Issue : #5501

Current `TestDataGenerator` assumes that output tensor shape can be get even before interpreter calculation.
However `TestDataGenerator` sometimes cannot inference the shape of output tensor in advance.
Therefore, we need to get a shape of output tensor shape as interpreter calculation result.
This commit will introduce it.


jinevening (approved):
LGTM

llFreetimell:
Refer to https://github.com/Samsung/ONE/pull/5594#issuecomment-756124418

Close!

Issue #5575, opened 2021-01-06
[Compiler][luci] Introduce new optimization pass, RemoveUnnecessaryReshapePass.


struss:
Parent Issue : #4353
#
### What to do in this Pass
Remove unnecessary(input and output shape same) `CircleReshape`.
#### Input Type
Reshape Op with input output shape same, this means input tensor shape is same with shape Const tensor or new_shape.
![Screenshot from 2020-12-22 12-30-50](https://user-images.githubusercontent.com/24720192/93737911-58124700-fc1f-11ea-9a2e-60f00b5244a2.png)
### TODO
- [x] Introduce luci/Pass for above optimization. #5596
- [x] Introduce this optimization on circle2circle. #5612 
- [x] Update one-cmd optimize option. #5632 
### Draft
#5576

struss:
Done! Thanks for all!

struss:
forget to update `one-cmd` optimize option.

struss:
Sorry! Really done!

PR #5596, opened 2021-01-07
[luci/pass] Introduce RemoveUnnecessaryReshapePass on luci/pass

struss:
For #5575 

This commit introduce a `RemoveUnnecessaryReshapePass` on luci/pass.

### DIFF
--- a/compiler/luci/pass/src/RemoveUnnecessaryReshapePass.cpp
+++ b/compiler/luci/pass/src/RemoveUnnecessaryReshapePass.cpp
@@ -27,43 +27,25 @@ bool remove_no_effect_reshape(luci::CircleNode *node)
   if (target_node == nullptr)
     return false;
 
-  std::vector<int32_t> shape_info;
   auto new_shape = dynamic_cast<luci::CircleConst *>(target_node->shape());
-  // If shape node is not a const node, get updated shape info from option.
   if (new_shape == nullptr)
-  {
-    auto dummy_shape = dynamic_cast<luci::CircleOutputDummy *>(target_node->shape());
-    if (dummy_shape == nullptr)
-      return false;
-
-    // Need to extract new_shape info on target_node->newShape()
-    shape_info.resize(target_node->newShape()->rank());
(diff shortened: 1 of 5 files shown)


compiler/luci/pass/src/RemoveUnnecessaryReshapePass.cpp
```diff
@@ -0,0 +1,93 @@
...
+  std::vector<int32_t> shape_info;
+  auto new_shape = dynamic_cast<luci::CircleConst *>(target_node->shape());
+  // If shape node is not a const node, get updated shape info from option.
+  if (new_shape == nullptr)
+  {
+    auto dummy_shape = dynamic_cast<luci::CircleOutputDummy *>(target_node->shape());
+    if (dummy_shape == nullptr)
+      return false;
+
+    // Need to extract new_shape info on target_node->newShape()
```
mhs4670go, line 39: @struss 
AFAIK, `CircleOutputDummy` is inserted when `Reshape` op doesn't have its option, which means it doesn't have `newShape` attribute as well. This line can be run?

struss: @mhs4670go sorry for my misunderstanding. I think your right. from `luci/import` logic, if second input not comes, shape is created from option. so if `CircleOutputDummy` comes in shape, it have neither shape and reshape option. 
I will fix this matching with `luci/import` logic. Thanks for pointing out!

struss: and also, if shape is `CircleOutputDummy`, then options also no value, follow from `luci/import` (can't extract shape from option)
https://github.com/Samsung/ONE/blob/5b4a868d686fe4d19af0c99921f784de8e027504/compiler/luci/import/src/Nodes/CircleReshape.cpp#L68-L76

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

Issue #5600, opened 2021-01-08
Compiler FE: Add options to preserve input/output shape in ConvertNCHWToNHWCPass


jinevening:
## What

Add options (`--preserve_input_shape`, `--preserve_output_shape`) to `--convert_nchw_to_nhwc` in circle2circle.

## Why

We'd like to provide users with an option to choose to preserve input/output shape of the model.

mhs4670go:
@jinevening It seems input shape and output shape should be preserved to make converting easy. Could you explain about this more detail? I mean I don't know "preserve shape" because nodes always have its shape.

jinevening:
> Could you explain about this more detail? I mean I don't know "preserve shape" because nodes always have its shape.

Well, it seems that my explanation was definitely not enough.

Currently, ConvertNCHWToNHWCPass changes input/output shape of a model under the assumption that the model's input/output is NCHW.

For example,

before: input shape [1, 16, 4, 4] - assumed to be NCHW
after: input shape [1, 4, 4, 16]

This is somewhat dangerous because app developers should change their input data according to the model. So, I'd like to provide an option not to change the input shape.

As to the output, output of some models is not image (e.g., output of face detection). For those models, we should not change the layout of output. That's why I add `--preserve_output_shape`.

jinevening:
All done.

PR #5601, opened 2021-01-08
DRAFT: Options to preserve input/output in ConvertNCHWToNHWC

jinevening:
On going draft to support options to preserve input/output shape in ConvertNCHWToNHWCPass.

---
Related to: #5600


seanshpark:
It would be nice if you could also revisit one-cmds

PR #5602, opened 2021-01-08
[luci] Options to preserve input/output in ConvertNCHWToNHWC

jinevening:
This adds options to preserve input/output shape in ConvertNCHWToNHWCPass.

---
Related to: #5600 
Draft PR: #5601

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
@@ -49,8 +49,8 @@ public:
   bool run(loco::Graph *g) final;
 
 private:
-  bool _preserve_input;
-  bool _preserve_output;
+  bool _preserve_input = false;
+  bool _preserve_output = false;
 };
 
 } // namespace luci
--- a/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
@@ -35,11 +35,22 @@ namespace luci
  */
 struct ConvertNCHWToNHWCPass final : public logo::Pass
 {
+public:
+  ConvertNCHWToNHWCPass(bool preserve_input = false, bool preserve_output = false)
+    : _preserve_input(preserve_input), _preserve_output(preserve_output)
+  {
+    // Do nothing
+  }
+
   virtual ~ConvertNCHWToNHWCPass() = default;
 
   const char *name(void) const final { return "luci::ConvertNCHWToNHWCPass"; }
 
   bool run(loco::Graph *g) final;
+
+private:
+  bool _preserve_input;
+  bool _preserve_output;
 };
 
 } // namespace luci
--- a/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
@@ -36,11 +36,13 @@ namespace luci
 struct ConvertNCHWToNHWCPass final : public logo::Pass
 {
 public:
-  ConvertNCHWToNHWCPass(bool preserve_input = false, bool preserve_output = false)
+  ConvertNCHWToNHWCPass(bool preserve_input, bool preserve_output)
     : _preserve_input(preserve_input), _preserve_output(preserve_output)
   {
     // Do nothing
   }
+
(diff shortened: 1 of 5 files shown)


compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
```diff
@@ -35,11 +35,22 @@ namespace luci
...
+
   virtual ~ConvertNCHWToNHWCPass() = default;
 
   const char *name(void) const final { return "luci::ConvertNCHWToNHWCPass"; }
 
   bool run(loco::Graph *g) final;
+
+private:
+  bool _preserve_input;
+  bool _preserve_output;
```
llFreetimell, line 53:
```suggestion
  bool _preserve_input = false;
  bool _preserve_output = false;
```

or make default constructor `delete` would be better

jinevening: Updated. Thanks.

llFreetimell (dismissed):
LGTM

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -513,7 +513,17 @@ bool ConvertNCHWToNHWCPass::run(loco::Graph *g)
     {
       // List of supported Ops
       case luci::CircleOpcode::CIRCLEINPUT:
+        if (!_preserve_input && !has_data_format(node))
```
seanshpark, line 516: Q) whay also check `has_data_format` here? and if the input node doesn't have data format why set to `NCHW`?

compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
```diff
@@ -35,11 +35,22 @@ namespace luci
  */
 struct ConvertNCHWToNHWCPass final : public logo::Pass
 {
+public:
+  ConvertNCHWToNHWCPass(bool preserve_input = false, bool preserve_output = false)
```
seanshpark, line 39: We usually don't use default argument values. please remove `= false` for `preserve_output`

compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
```diff
@@ -238,7 +238,7 @@ void check_post_trans(loco::Node *node)
   EXPECT_EQ(2, post_trans_perm->at<loco::DataType::S32>(3));
 }
 
-void run_phase(loco::Graph *g)
+void run_phase(loco::Graph *g, bool preserve_input = false, bool preserve_output = false)
```
seanshpark, line 241: ditto

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: ConverNCHWToNHWC works as below.

1. Set data format of target nodes to NCHW (**this should be done once**)
2. Convert the target nodes from NCHW  to NHWC.

The above steps are repeated until no change is detected. To do step 1 once, I check the existence of data_format.

compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
jinevening: Can I ask you why we don't use default arguments?

Default arguments can reduce the changes in the test codes. I can remove them in this PR, but am curious about the reason.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
seanshpark: Thanks! I've forgot -_-;

compiler/luci/pass/include/luci/Pass/ConvertNCHWToNHWCPass.h
seanshpark: Cause of code is implicit and implicit codes may make problems when there are more changes later. this was from when @parjong started.

jinevening: I see. Thanks :)

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM thank you!

PR #5603, opened 2021-01-08
[luci] Consider when rank is not 4

llFreetimell:
According to comment, `pred_node` should be rank 4.
However this is not correctly applied at code.
This commit will fix it.

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -256,7 +256,7 @@ bool is_NCHW_with_const(const luci::CircleMul *node, luci::CircleNode *&pred_nod
   }
 
   const auto const_rank = multiplier->rank();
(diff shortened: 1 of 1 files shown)


jinevening (approved):
LGTM

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -256,7 +256,7 @@ bool is_NCHW_with_const(const luci::CircleMul *node, luci::CircleNode *&pred_nod
   }
 
   const auto const_rank = multiplier->rank();
-  if (const_rank != 4)
+  if (const_rank != 4 || pred_node->rank() != 4)
```
seanshpark, line 259: This is OK but it seems better to check `pred_node->rank() != 4` after above `else` block when `pred_node` is set.

seanshpark (approved):
LGTM thank you!

jinevening (approved):

PR #5606, opened 2021-01-08
[luci] Encode ShapeSignature using shape description

llFreetimell:
Parent Issue : #5501

Until now, shape_signature was encoded using `CircleShapeSignature`.
However, as `TensorShape` of `CircleNode` can represent shape_siganture,
this commit will enable encoding shape_signature using shape description, not `CircleShapeSignature`.

### DIFF
--- a/compiler/luci/export/src/CircleTensorExporter.cpp
+++ b/compiler/luci/export/src/CircleTensorExporter.cpp
@@ -247,12 +247,21 @@ flatbuffers::Offset<Vector<int32_t>> encodeShape(FlatBufferBuilder &builder,
 }
 
 flatbuffers::Offset<Vector<int32_t>> encodeShapeSignature(FlatBufferBuilder &builder,
-                                                          const ShapeSignature &shape_signature)
+                                                          const ShapeDescription &shape)
 {
-  if (shape_signature.rank() == 0)
-    return 0;
+  assert(shape._rank_known && "unknown number of dimensions is not supported");
+  flatbuffers::Offset<Vector<int32_t>> signature_offset;
+  bool has_unknown = false;
 
-  return builder.CreateVector(shape_signature.as_vector());
+  for (uint32_t i = 0; i < shape._dims.size(); ++i)
+    if (shape._dims.at(i) == -1)
+      has_unknown = true;
+
+  // If all of dimensions are known, shape_signature should be empty.
+  if (has_unknown)
+    signature_offset = builder.CreateVector(shape._dims);
+
+  return signature_offset;
 }
 
 flatbuffers::Offset<circle::Buffer> encodeOpBuffer(FlatBufferBuilder &builder)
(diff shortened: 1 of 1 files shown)


compiler/luci/export/src/CircleTensorExporter.cpp
```diff
@@ -247,12 +247,21 @@ flatbuffers::Offset<Vector<int32_t>> encodeShape(FlatBufferBuilder &builder,
 }
 
 flatbuffers::Offset<Vector<int32_t>> encodeShapeSignature(FlatBufferBuilder &builder,
-                                                          const ShapeSignature &shape_signature)
+                                                          const ShapeDescription &shape)
 {
-  if (shape_signature.rank() == 0)
-    return 0;
+  assert(shape._rank_known && "unknown number of dimensions is not supported");
```
seanshpark, line 252: does the caller ensure this condition?
or do we have some test models with unknown rank so that the caller do something?

compiler/luci/export/src/CircleTensorExporter.cpp
```diff
@@ -247,12 +247,21 @@ flatbuffers::Offset<Vector<int32_t>> encodeShape(FlatBufferBuilder &builder,
...
+  flatbuffers::Offset<Vector<int32_t>> signature_offset;
+  bool has_unknown = false;
+
+  for (uint32_t i = 0; i < shape._dims.size(); ++i)
+    if (shape._dims.at(i) == -1)
+      has_unknown = true;
 
-  return builder.CreateVector(shape_signature.as_vector());
+  // If all of dimensions are known, shape_signature should be empty.
+  if (has_unknown)
```
seanshpark, line 261: above comment and the code seems to match in logic but explains in different view. can you match the code and the comment? current is a bit confusing.

compiler/luci/export/src/CircleTensorExporter.cpp
llFreetimell: The only case that `shape._rank_known` is `false` is when `ShapeStatus` of the node is `UNDEFINED` or `NOSHAPE`.

https://github.com/Samsung/ONE/blob/1733eec56afb19d045cd775ab89f21b4feb3340e/compiler/luci/export/src/CircleTensorExporter.cpp#L116-L117

These case are kept when `encodeShapeSignature` is called.

https://github.com/Samsung/ONE/blob/e60ddf7524ddf374632f71d9c59aabca3a6d5ae4/compiler/luci/export/src/CircleTensorExporter.cpp#L457-L461

This feature is already used in `encodeShape`.
I think that https://github.com/Samsung/ONE/issues/5510#issuecomment-752337908 may be the reason :D

seanshpark: Thanks for the explanation, but isn't there possibility that rank maybe unknown? I was wondering for that case.

compiler/luci/export/src/CircleTensorExporter.cpp
seanshpark: (optional)
```cxx
  // shape_signature is only set if one of dimensions is unknown.
  for (uint32_t i = 0; i < shape._dims.size(); ++i)
    if (shape._dims.at(i) == -1)
    {
      signature_offset = builder.CreateVector(shape._dims);
      break;
    }
```

compiler/luci/export/src/CircleTensorExporter.cpp
seanshpark: I think we can go as is. question was just from curiosity.

llFreetimell: Currently, it is impossible that some models have unknown rank.
For now, if shape inference impossible, just [] is returned, which is rank-0.

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

PR #5607, opened 2021-01-08
[luci] Deprecate CircleShapeSignature

llFreetimell:
Parent Issue : #5501

As `TensorShape` can represent dynamic shape without `CircleShapeSignature`,
this commit will deprecate `CircleShapeSignature` using codes.


seanshpark (approved):
LGTM

PR #5609, opened 2021-01-08
[nnfwapi] Test: fix Add model test bug

wateret:
As `nnfw_load_circle_from_buffer` requires the input buffer to outlive
the session, so this change makes `_cbuf` as a member.

### DIFF
--- a/tests/nnfw_api/src/fixtures.h
+++ b/tests/nnfw_api/src/fixtures.h
@@ -100,8 +100,9 @@ protected:
     ValidationTestSessionCreated::SetUp();
     if (PackageNo == NNPackages::ADD)
     {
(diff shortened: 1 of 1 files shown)


wateret:
I found this bug by running a single `ValidationTestAddSessionPrepared.run`(in `nnfw_api_gtest`') test. And it failed. But when running all the tests together it passes since it got lucky...

glistening:
@wateret I've found another `nnfw_load_circle_from_buffer` in line 189. I think we need to fix it also.

hseok-oh (approved):
LGTM

wateret:
@glistening Will fix that too. THanks!

PR #5619, opened 2021-01-11
[onert] Support Conv2D int8 channel-wise quantization

glistening:
: Introduce convQuant8PerChannel
: Add two positive and two negative testcases for int8

### DIFF
--- a/runtime/onert/backend/cpu/ops/ConvolutionLayer.cc
+++ b/runtime/onert/backend/cpu/ops/ConvolutionLayer.cc
@@ -98,6 +98,32 @@ void ConvolutionLayer::convQuant8()
          getBuffer<uint8_t>(_output));
 }
 
+void ConvolutionLayer::convQuant8PerChannel()
+{
+  int32_t output_activation_min = 0;
+  int32_t output_activation_max = 0;
+  CalculateActivationRangeQuantized(_activation, _output, &output_activation_min,
+                                    &output_activation_max);
+
+  nnfw::cker::ConvParams op_params;
+  op_params.input_offset = -_input->data_zero_point();
+  op_params.output_offset = _output->data_zero_point();
+  op_params.stride_height = _strideHeight;
+  op_params.stride_width = _strideWidth;
+  op_params.dilation_height_factor = _dilationHeightFactor;
+  op_params.dilation_width_factor = _dilationWidthFactor;
+  op_params.padding_values.height = _paddingTop;
+  op_params.padding_values.width = _paddingLeft;
+  op_params.quantized_activation_min = output_activation_min;
+  op_params.quantized_activation_max = output_activation_max;
+
+  nnfw::cker::Conv &kernel = *_conv_kernel;
+  kernel(op_params, getShape(_input), reinterpret_cast<const int8_t *>(_input->buffer()),
+         getShape(_kernel), reinterpret_cast<const int8_t *>(_kernel->buffer()), getShape(_bias),
+         reinterpret_cast<const int32_t *>(_bias->buffer()), getShape(_output),
+         reinterpret_cast<int8_t *>(_output->buffer()));
+}
+
 void ConvolutionLayer::configure(const IPortableTensor *input, const IPortableTensor *kernel,
                                  const IPortableTensor *bias, const ir::PaddingType paddingType,
                                  const uint32_t paddingLeft, const uint32_t paddingRight,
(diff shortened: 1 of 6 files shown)


glistening:
Related: #4664

glistening:
@nnfw-bot test x64-builder

runtime/onert/backend/cpu/ops/ConvolutionLayer.cc
```diff
@@ -197,6 +227,15 @@ void ConvolutionLayer::prepare()
     kernel.prepareQuant(getShape(_input), getShape(_kernel), getShape(_output), _strideWidth,
                         _strideHeight, _dilationWidthFactor, _dilationHeightFactor);
   }
+  else if (_input->data_type() == OperandType::QUANT_INT8_ASYMM && _kernel->is_constant() &&
+           !_input->is_dynamic() && !_output->is_dynamic())
+  {
+    kernel.getQuantizedConvolutionMultipliersAndShifts(
+      _input->data_scale(), _output->data_scale(), _kernel->data_scales().data(),
+      _kernel->data_scales().size(), getShape(_kernel).Dims(0));
+    kernel.prepareQuant(getShape(_input), getShape(_kernel), getShape(_output), _strideWidth,
```
periannath, line 236: This function is only necessary when kernel uses `im2col`. It looks like reference kernel doesn't use `im2col`.

runtime/onert/backend/cpu/ops/ConvolutionLayer.cc
```diff
@@ -197,6 +227,15 @@ void ConvolutionLayer::prepare()
     kernel.prepareQuant(getShape(_input), getShape(_kernel), getShape(_output), _strideWidth,
                         _strideHeight, _dilationWidthFactor, _dilationHeightFactor);
   }
+  else if (_input->data_type() == OperandType::QUANT_INT8_ASYMM && _kernel->is_constant() &&
+           !_input->is_dynamic() && !_output->is_dynamic())
+  {
+    kernel.getQuantizedConvolutionMultipliersAndShifts(
```
periannath, line 233: This function will be called when input, kernel, and output tensors are constant. You need to call `getQuantizedConvolutionMultipliersAndShifts` at inference time to get proper quantized parameters for dynamic tensors.

periannath (approved):
LGTM

runtime/onert/backend/cpu/ops/ConvolutionLayer.cc
```diff
@@ -164,6 +190,10 @@ void ConvolutionLayer::run()
   {
     convQuant8();
   }
+  else if (_input->data_type() == OperandType::QUANT_INT8_ASYMM)
```
ragmani, line 193: How about changing data type to `QUANT_INT8_SYMM_PER_CHANNEL`?

glistening: @ragmani `input` is asymmetric.

ragmani: I see.

ragmani (approved):
LGTM

runtime/onert/backend/cpu/ops/ConvolutionLayer.cc
glistening: @ragmani Thank you for pointing out `QUANT_INT8_SYMM_PER_CHANNEL`.

@hseok-oh, @ragmani I think `QUANT_INT8_SYMM_PER_CHANNEL` can (or should) be removed from onert IR.

ragmani: I think the type should not be removed. The weights may have the type with a param called as quantized dimension(channel index).

glistening: > I think the type should not be removed. The weights may have the type with a param called as quantized dimension(channel index).

@ragmani I know the weights may have per-channel quantization params. I think `QUANT_INT8_SYMM_PER_CHANNEL` is introduced from Android NNAPI, which was introduced at API level 29 to represent per-channel quantization since prior type does not contain per-channel scales and zeropoints. However, onert IR can represent per-channel quantization without dedicated type as tflite does. We may create an issue if we need further discussion. It is also related to other types that has `_ASYMM`, `_SYMM`, ....

ragmani: `QUANT_INT8_SYMM_PER_CHANNEL` is used some backends such as ACL. If onert does not have that type, How are you going to get onert to support that type thought the weights may have per-channel quantization params?

glistening: @ragmani I did not considered `ACL` backend. I've found `ARMComputeLibrary` uses `QSYMM8_PER_CHANNEL` type for channel-wise quantization tensor.

PR #5623, opened 2021-01-11
[Res/Recipe] Add recipe for BidirectionalsequenceLSTM

dr-venkman:
This commit adds recipes for `BidirectionalSequenceLSTM`.

ONE-DCO-1.0-Signed-off-by: venkat.iyer venkat.iyer@samsung.com

### DIFF
--- a/compiler/common-artifacts/exclude.lst
+++ b/compiler/common-artifacts/exclude.lst
@@ -5,6 +5,8 @@
 
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
+optimize(BidirectionalSequenceLSTM_000)
+optimize(BidirectionalSequenceLSTM_001)
 optimize(UnidirectionalSequenceLSTM_001) # This recipe contains is_variable Tensor
 
 ## CircleRecipes
@@ -28,6 +30,8 @@ tcgenerate(BatchMatMul_000)
 tcgenerate(BatchMatMulV2_000)
 tcgenerate(BatchMatMulV2_001)
 tcgenerate(BatchToSpaceND_000)
+tcgenerate(BidirectionalSequenceLSTM_000) # runtime and luci-interpreter doesn't support BidirectionalSequenceLSTM op yet
+tcgenerate(BidirectionalSequenceLSTM_001) # runtime and luci-interpreter doesn't support BidirectionalSequenceLSTM op yet
 tcgenerate(Cast_000)
 tcgenerate(Cast_001)
 tcgenerate(Ceil_000)
--- a/res/TensorFlowLiteRecipes/BidirectionalSequenceLSTM_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/BidirectionalSequenceLSTM_000/test.recipe
@@ -1,0 +1,361 @@
+operand {
+  name: "ifm"
+  type: FLOAT32
+  shape { dim: 3 dim: 1 dim: 2 }
+}
+operand {
+  name: "fw_input_to_input_weights"
+  type: FLOAT32
+  shape { dim: 4 dim: 2 }
+}
+operand {
+  name: "fw_input_to_forget_weights"
+  type: FLOAT32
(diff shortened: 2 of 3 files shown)


compiler/common-artifacts/exclude.lst
```diff
@@ -5,6 +5,8 @@
 
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
+optimize(BidirectionalSequenceLSTM_000)
+optimize(BidirectionalSequenceLSTM_001)
```
mhs4670go, line 9: Could you add some comments about why you add  these recipes to `optimize` exclude list?

struss:
Just suggestion : what about adding a testing resources from [here](https://github.com/tensorflow/tensorflow/blob/v2.3.0/tensorflow/lite/kernels/bidirectional_sequence_lstm_test.cc)

compiler/common-artifacts/exclude.lst
dr-venkman: @mhs4670go, I added these lines to exclude list, to ensure that the `./nncc test` works without errors. As an examples, if I do not add as above and test, then I get the following errors:
```
terminate called after throwing an instance of 'oops::UserExn'
  what():  Error: Not supported: BIDIRECTIONAL_SEQUENCE_LSTM
Aborted (core dumped)
compiler/common-artifacts/CMakeFiles/common_artifacts_deps.dir/build.make:3019: recipe for target 'compiler/common-artifacts/BidirectionalSequenceLSTM_000.opt.circle' failed
make[2]: *** [compiler/common-artifacts/BidirectionalSequenceLSTM_000.opt.circle] Error 134
CMakeFiles/Makefile2:18900: recipe for target 'compiler/common-artifacts/CMakeFiles/common_artifacts_deps.dir/all' failed
make[1]: *** [compiler/common-artifacts/CMakeFiles/common_artifacts_deps.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
```

The definition for `BIDIRECTIONAL_SEQUENCE_LSTM` comes later under `luci import` as I understand, so this is a temporary fix to avoid errors as mentioned above. Please let me know if there is a workaround that does not involve exclusion.

dr-venkman:
> Just suggestion : what about adding a testing resources from [here](https://github.com/tensorflow/tensorflow/blob/v2.3.0/tensorflow/lite/kernels/bidirectional_sequence_lstm_test.cc)

@struss, thanks for the link. Do you prefer a new, third recipe, or a modification of one of the two recipes added already?

compiler/common-artifacts/exclude.lst
seanshpark: How about adding _luci_ files first and then `TensorFlowLiteRecipes`?

dr-venkman: @seanshpark , thanks for the suggestion, I will mark this pull request as do-not-merge until all the luci files have been added and merged.

seanshpark:
One file is in `res/TensorFlowLiteRecipes` and one file is in ` compiler/common-artifacts`.
We have a practice to separate PRs for each sub modules `compiler/XXX` or `res/YYY`.
Please split this into 2 PRs.

struss:
> @struss, thanks for the link. Do you prefer a new, third recipe, or a modification of one of the two recipes added already?

I think adding new one would be nice.

dr-venkman:
> One file is in `res/TensorFlowLiteRecipes` and one file is in ` compiler/common-artifacts`.
> We have a practice to separate PRs for each sub modules `compiler/XXX` or `res/YYY`.
> Please split this into 2 PRs.

@seanshpark , agreed. I will do that after posting the luci pull requests.

dr-venkman:
> > @struss, thanks for the link. Do you prefer a new, third recipe, or a modification of one of the two recipes added already?
> 
> I think adding new one would be nice.

@struss, Thanks for clarifying. I will keep in context when revisiting this pull request.

seanshpark:

PR #5621, opened 2021-01-11
DRAFT: Fold SparseToDense with zero-sized indices

jinevening:
On going draft to fold SparseToDense with zero-sized indices.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to : https://github.com/Samsung/ONE/issues/5598

Issue #5598, opened 2021-01-08
Compiler FE: Support constant folding for SparseToDense, AddV2 and Cast


jinevening:
## What

Let's support constant folding for the below operators.

1. SparseToDense @jinevening 
If the size of sparse_indices is 0 and the rest of inputs are constant -> Fold to constant

2. AddV2 @jinevening 
If all inputs are constant -> Fold to constant

3. Cast @jinevening 
If input is constant -> Fold to constant

## Why

onnx-tf generates the above patterns for Reshape Op. They can be removed by constant folding, but tf2tfliteV2 does not handle those patterns.

onnx model
![reshape_onnx](https://user-images.githubusercontent.com/5449554/104256881-726a7c00-54bf-11eb-8453-755e19d1b8ae.png)

tflite model (generated by onnx-tf)
![reshape_circle](https://user-images.githubusercontent.com/5449554/104256886-74ccd600-54bf-11eb-9c2d-5f2fdf963599.png)

jinevening:
I'm working on the first operator (SparseToDense). Any help is welcome.

jinevening:
Interim result. SparseToDense was folded. I'm working on AddV2.
```
circle2circle --fold_sparse_to_dense test.circle test.folded_stod.circle
```

![image](https://user-images.githubusercontent.com/5449554/104431311-1a20a080-55cb-11eb-9093-c920d467159c.png)

seanshpark:
I missed something for #5687

There's a pass `ResolveCustomOpAddPass` and I think it would be better not to work with `CircleCustom` but with `CircleAdd`,
What do you think about this?

seanshpark:
If would could rework on `CircleAddV2` then, (1) undo #5687, (2) add constant folding for `CircleAdd`, ...

jinevening:
> What do you think about this?

CircleCustom may be less stable, so working with CircleAdd looks good.

But for now, I think we had better go with AddV2. In the target pattern, inputs of AddV2 are int64. Add in TFLite does not support int64 (it is not runnable on TFLite interpreter). That's why AddV2 was generated by TFLite converter in the target pattern.

seanshpark:
> inputs of AddV2 are int64

OK, this is a good reason to continue with CustomOp AddV2

> We had better insert some code in ResolveCustomOpAddPass not to convert AddV2 to Add when the inputs are unsupported types.

I'll add an todo issue for this

jinevening:
All done. Final result.

```
[...]

PR #5624, opened 2021-01-11
[luci] Pass to fold SparseToDense with zero-sized indices

jinevening:
This adds a pass to fold SparseToDense with zero-sized indices.

---
Related to : #5598
Draft PR: #5621

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/FoldSparseToDensePass.h
+++ b/compiler/luci/pass/include/luci/Pass/FoldSparseToDensePass.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/compiler/luci/pass/include/luci/Pass/FoldSparseToDensePass.h
+++ b/compiler/luci/pass/include/luci/Pass/FoldSparseToDensePass.h
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_FOLD_SPARSE_TO_DENSE_PASS_H__
+#define __LUCI_FOLD_SPARSE_TO_DENSE_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to fold SparseToDense to a constant tensor
+ *
+ */
+struct FoldSparseToDensePass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::FoldSparseToDensePass"; }
+
+  bool run(loco::Graph *g) final;
+};
+
+} // namespace luci
+
+#endif // __LUCI_FOLD_SPARSE_TO_DENSE_PASS_H__
--- a/compiler/luci/pass/src/FoldSparseToDensePass.cpp
+++ b/compiler/luci/pass/src/FoldSparseToDensePass.cpp
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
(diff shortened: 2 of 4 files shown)


compiler/luci/pass/include/luci/Pass/FoldSparseToDensePass.h
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
```
llFreetimell, line 2:
```suggestion
 * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
```

compiler/luci/pass/src/FoldSparseToDensePass.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/**
+ * Fold to const if
+ *
+ * 1. indices has 0-sized static shape (i.e., output is filled with default value)
```
llFreetimell, line 28: 0-sized static shape means scalar?

compiler/luci/pass/src/FoldSparseToDensePass.cpp
```diff
@@ -0,0 +1,137 @@
...
+ * 2. defalut_value: const scalar
```
llFreetimell, line 29:
```suggestion
 * 2. default_value: const scalar
```

compiler/luci/pass/src/FoldSparseToDensePass.cpp
```diff
@@ -0,0 +1,137 @@
...
+  {
+    if (auto stod = dynamic_cast<luci::CircleSparseToDense *>(node))
+    {
+      if (fold_sparse_to_dense(stod))
+        return true;
+    }
+  }
+
+  return false;
+}
```
llFreetimell, line 135: Is there any problem with following changes?

```suggestion
bool FoldSparseToDensePass::run(loco::Graph *g)
{
  bool changed = false;
  for (auto node : loco::active_nodes(loco::output_nodes(g)))
  {
    if (auto stod = dynamic_cast<luci::CircleSparseToDense *>(node))
    {
      if (fold_sparse_to_dense(stod))
        changed = true;
    }
  }

  return changed;
}
```

compiler/luci/pass/src/FoldSparseToDensePass.cpp
jinevening: The second one. The word is confusing. Do you have a better word?

compiler/luci/pass/src/FoldSparseToDensePass.cpp
jinevening: No problem. I'll update the code.

compiler/luci/pass/src/FoldSparseToDensePass.cpp
seanshpark: maybe add more comments like `(for example, [0])` ?

jinevening: > maybe add more comments like (for example, [0]) ?

Added comments. PTAL

llFreetimell: I was confused because of `output is filled with default value`.
0-sized shape itself is okay for me but that sentence was little bit confused since 0-sized tensor cannot have a value.

llFreetimell: `output` in `output is filled with default value` and indices are different things?

jinevening: > output in output is filled with default value and indices are different things?

Yes. To understand the comment, you have to know the semantics of SparseToDense first.

SpareToDense has four inputs.

`indices`
`output_shape`
`values`
`default_value`

Output tensor is filled in two ways.
1. `indices` in output tensor is filled with `values`.
2. The rest of the tensor is filled with `default_value`.

So, if the size of `indices` is zero, the entire output tensor is filled with `default_value`.

llFreetimell: Now, I understand all clearly, thanks :)

compiler/luci/pass/src/FoldSparseToDensePass.cpp
```diff
@@ -0,0 +1,139 @@
...
+  int32_t indices_size = 1;
+  for (uint32_t i = 0; i < indices->rank(); i++)
+  {
+    if (!indices->dim(i).known())
+      return false;
+    indices_size *= indices->dim(i).value();
+  }
+
+  if (indices_size != 0)
+    return false;
```
llFreetimell, line 51: I think that even some dimensions are unknown, some cases are possible to be folded.
For example, if shape of indices is [?, 0, ?, 1], it is definitely zero sized shape.
So how about following codes?

```suggestion
  bool has_zero = false;
  for (uint32_t i = 0; i < indices->rank(); i++)
  {
    if (indices->dim(i).known() && indices->dim(i).value() == 0)
      has_zero = true;
  }

  if (!has_zero)
    return false;
```

jinevening: Good idea :) I've updated the code. Thanks.

llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

jinevening:
@seanshpark PTAL

Issue #5610, opened 2021-01-11
Compiler FE: Support FakeQuant Op


seanshpark:
We now need to support this `FakeQuant` Op.

- [x]  res/TensorFlowPythonExamples
- [x]  tflchef
- [x]  tflite2circle
- [x]  tfldump, circledump
- [x]  luci.IR
- [x]  luci.Import
- [x]  luci.Export
- [x]  luci.logex
- [x]  luci.Service
- [x] common-artifacts
- [x] res/TensorFlowLiteRecipes
- [x]  luci.tests

And for circlequantizer 
- --> #5672

For #1606

seanshpark:
Done

PR #5625, opened 2021-01-11
DRAFT: Compiler FE: Support FakeQuant Op

seanshpark:
on-going draft to support FakeQuant Op.


seanshpark:
For #5610

seanshpark:
Done

PR #5628, opened 2021-01-11
[one-cmds] Tests to import onnx files

seanshpark:
This will introduce tests for one-import-onnx and one-build with onnx
files.


seanshpark:
For #5548, from draft #5553

mhs4670go (approved):
LGTM

PR #5630, opened 2021-01-12
[one-cmds] Revise one-build section error message

seanshpark:
This will revise one-build section error message.

### DIFF
--- a/compiler/one-cmds/tests/one-build_neg_003.test
+++ b/compiler/one-cmds/tests/one-build_neg_003.test
@@ -21,7 +21,7 @@ filename="${filename_ext%.*}"
 
 trap_err_onexit()
 {
-  if grep -q "section is required in configuraion file" "${filename}.log"; then
+  if grep -q "\[one-build\] section is required in configuraion file" "${filename}.log"; then
     echo "${filename_ext} SUCCESS"
     exit 0
   fi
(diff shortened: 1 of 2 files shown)


seanshpark:
This has effect to solve some problem with my vscode cause of odd number of `'` character.

mhs4670go (approved):
LGTM

compiler/one-cmds/tests/one-build_neg_003.test
```diff
@@ -21,7 +21,7 @@ filename="${filename_ext%.*}"
 
 trap_err_onexit()
 {
-  if grep -q "'one-build' section is required in configuraion file" "${filename}.log"; then
+  if grep -q "section is required in configuraion file" "${filename}.log"; then
```
seanshpark, line 24: using `[]` seems to work as regex pattern match and `[section]` fails to work

mhs4670go: Maybe `\[one-build\]`? Or is it backslash that makes error?

seanshpark: > Maybe `\[one-build\]` ?

I'll try this one

> Or is it backslash that makes error?

Yes `[]` acts as regex pattern so `[one-build]` tries something and fails.

mhs4670go (dismissed):
LGTM

compiler/one-cmds/tests/one-build_neg_003.test
seanshpark: `\[one-build\]` this works :)

mhs4670go (approved):
LGTM

PR #5633, opened 2021-01-12
[luci] Fix order of CircleFill in logex

seanshpark:
This will fix alphabetical order of CircleFill in logex.


mhs4670go (approved):
LGTM

PR #5639, opened 2021-01-12
[luci] Introduce FakeQuant IR

seanshpark:
This will introduce CircleFakeQuant IR for FAKE_QUANT Op.


seanshpark:
For #5610, from draft #5625

compiler/luci/lang/include/luci/IR/Nodes/CircleFakeQuant.h
```diff
@@ -0,0 +1,60 @@
...
+  int32_t num_bits(void) const { return _num_bits; }
+  void num_bits(int32_t num_bits) { _num_bits = num_bits; }
+
+  bool narrow_range(void) const { return _narrow_range; }
+  void narrow_range(bool narrow_range) { _narrow_range = narrow_range; }
+
+private:
+  float _min = 0.0f;
+  float _max = 0.0f;
+  int32_t _num_bits = 0;
```
jinevening, line 54: (optional) how about following the default values in TF?

data| type
-- | --
min | An optional float. Defaults to -6.
max | An optional float. Defaults to 6.
num_bits | An optional int. Defaults to 8.
narrow_range | An optional bool. Defaults to False.

jinevening (approved):

compiler/luci/lang/include/luci/IR/Nodes/CircleFakeQuant.h
seanshpark: Thanks! Current values are from schema generated file (as it's circle file).
I don't think we had default value policy (usually values were 0) but we can decide one.

mhs4670go (approved):
LGTM

PR #5643, opened 2021-01-12
[luci] Workaround for disabling assert in shape inference

llFreetimell:
Parent Issue : #5501

Until now, unknown dimensions are not considered during shape inference.
So if some dimensions are unknown, some of asserts are triggered.
This commit will disable it by setting unknown dimension as 1 as workaround.


seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5644, opened 2021-01-12
[onert] Logging in ConstantInsertionPass

wateret:
Leave some log messages of ConstantInsertionPass.

### DIFF
--- a/runtime/onert/core/src/compiler/pass/ConstantInsertionPass.cc
+++ b/runtime/onert/core/src/compiler/pass/ConstantInsertionPass.cc
@@ -17,8 +17,9 @@
 #include "ConstantInsertionPass.h"
 
 #include "backend/Backend.h"
-#include <ir/Graph.h>
-#include <util/Utils.h>
+#include "ir/Graph.h"
+#include "util/Utils.h"
+#include "util/logging.h"
 
 namespace onert
 {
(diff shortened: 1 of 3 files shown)


wateret:
Example log:

```
[ConstantInsertionPass] New operand %4 added(a copy of %1) for (cpu/NHWC)
[ConstantInsertionPass] Original operand %1 removed - no uses
[ConstantInsertionPass] New operand %5 added(a copy of %2) for (cpu/NHWC)
[ConstantInsertionPass] Original operand %2 removed - no uses
```

seanshpark (approved):
LGTM

runtime/onert/core/src/compiler/pass/ConstantInsertionPass.cc
```diff
@@ -19,6 +19,7 @@
 #include "backend/Backend.h"
 #include <ir/Graph.h>
 #include <util/Utils.h>
+#include "util/logging.h"
```
glistening, line 22: @wateret I have question about coding style rule. Because 

i) line 21 uses `<` and line 22 uses `"` for headers under same folder "util" and 
ii) there is no grouping between `< >` headers and `" "` headers.

Do we have no rule for this two cases? I guess no rules because I cannot find the coding style document and the format checker does not complains. :)

wateret: They all should be `""`, as they are in the same module with this source file. I'll fix those too.
And there is no rules for grouping(AFAIR), but I personally prefer grouping `<>`s and `""`. (But keep the header file for itself at the top)

wateret:
@glistening Updated, PTAL :smiley:

seanshpark (approved):
LGTM

glistening (approved):
LGTM

PR #5651, opened 2021-01-12
[tfldump] Enable FAKE_QUANT

seanshpark:
This will enable tflite file dump of FAKE_QUANT Op.

### DIFF
--- a/compiler/tfldump/src/OpPrinter.cpp
+++ b/compiler/tfldump/src/OpPrinter.cpp
@@ -305,6 +305,24 @@ public:
   }
 };
 
+class FakeQuantPrinter : public OpPrinter
+{
+public:
+  void options(const tflite::Operator *op, std::ostream &os) const override
+  {
+    if (auto *params = op->builtin_options_as_FakeQuantOptions())
+    {
+      os << "    ";
+      os << "Min(" << params->min() << ") ";
+      os << "Max(" << params->max() << ") ";
+      os << "NumBits(" << params->num_bits() << ") ";
+      os << std::boolalpha;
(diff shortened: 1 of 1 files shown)


seanshpark:
For #5610, from draft #5625

compiler/tfldump/src/OpPrinter.cpp
```diff
@@ -305,6 +305,24 @@ class DepthwiseConv2DPrinter : public OpPrinter
...
+public:
+  void options(const tflite::Operator *op, std::ostream &os) const override
+  {
+    if (auto *params = op->builtin_options_as_FakeQuantOptions())
+    {
+      os << "    ";
+      os << "Min(" << params->min() << ") ";
+      os << "Max(" << params->max() << ") ";
+      os << "NumBits(" << params->num_bits() << ") ";
+      os << std::boolalpha;
```
llFreetimell, line 319: AFAIK, unless `std::noboolalpha` should be called at the end of this function, other codes can be affected.

seanshpark: This is what I get
```
O(0:0) FAKE_QUANT
    Min(0) Max(1) NumBits(8) NarrowRange(false)
    I T(0:0) ifm
    O T(0:1) ofm
```
You mean we should turn off with `noboolalpha`?
There already exists several uses if `boolalpha` and seems OK...

seanshpark: Done some test and it would be safe to put `noboolalpha` after boolean dump.

llFreetimell (approved):
LGTM, thanks!

mhs4670go (approved):
LGTM

PR #5653, opened 2021-01-13
[luci] Shape/dtype inference for CircleFakeQuant

seanshpark:
This will enable shape and dtype inference for CircleFakeQuant node.


seanshpark:
For #5610, from draft #5625

llFreetimell (approved):
LGTM

Issue #5642, opened 2021-01-12
[ONERT] OpenCL Tuner for acl_cl backend


intom:
I applied OpenCL Tuner to the acl_cl backend.
OpenCL Tuner finds the optimal local work size for GPU acceleration tuning parameters.
https://developer.arm.com/solutions/machine-learning-on-arm/developer-material/how-to-guides/implement-a-neural-style-transfer-on-android-with-arm-nn-apis/tuning-performance-with-opencl-tuner

I am going to introduce the OpenCL Tuner with Draft PR

Refer to the below test result.
Model : mnasnet
https://tfhub.dev/tensorflow/lite-model/mnasnet_1.0_224/1/default/1
**CL_Tuner** 
```
Package Filename /root/mnasnet_1.0_224_1_default_1/
===================================
MODEL_LOAD   takes 27.815 ms
PREPARE      takes 3220.900 ms
EXECUTE      takes 16.578 ms
- MEAN     :  16.578 ms
- MAX      :  28.282 ms
- MIN      :  14.836 ms
- GEOMEAN  :  16.434 ms
===================================
```

**No CL_Tuner**
```
Package Filename /root/mnasnet_1.0_224_1_default_1/
===================================
MODEL_LOAD   takes 11.660 ms
PREPARE      takes 3142.837 ms
EXECUTE      takes 19.295 ms
- MEAN     :  19.295 ms
- MAX      :  25.660 ms
- MIN      :  17.280 ms
- GEOMEAN  :  19.198 ms
===================================

intom:
cc @ragmani

PR #5657, opened 2021-01-13
OpenCL Tuner for acl_cl backend

intom:
- To find the optimal local work size

related issue : #5642

I have tried to apply OpenCL Tuner to Config of acl_cl backend
But this PR is only Draft.

It should support that external actor (i.e. nnpackage_run, 3rd party app) to set the tuner mode and the path of file with tuned parameter.

I am waiting for your opinion on the appropriate approach.
Give your comment on related issue board.

### DIFF
--- a/runtime/onert/api/include/nnfw.h
+++ b/runtime/onert/api/include/nnfw.h
@@ -23,6 +23,7 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include <string>
 
 #ifdef __cplusplus
 extern "C" {
@@ -146,6 +147,16 @@ typedef enum
   NNFW_INFO_ID_VERSION = 0,
 } NNFW_INFO_ID;
 
+typedef enum
+{
+  NNFW_CLTUNER_MIN = -1,
+  NNFW_CLTUNER_READ,
+  NNFW_CLTUNER_EXHAUSTIVE,
+  NNFW_CLTUNER_NORMAL,
+  NNFW_CLTUNER_RAPID,
+  NNFW_CLTUNER_MAX
+} NNFW_CLTUNER_MODE;
+
 /**
  * @brief Maximum rank expressible with nnfw
  */
@@ -175,6 +186,14 @@ typedef struct nnfw_tensorinfo
    */
   int32_t dims[NNFW_MAX_RANK];
 } nnfw_tensorinfo;
+
+typedef struct nnfw_cltunerinfo
+{
+  bool active;
+  bool update;
+  NNFW_CLTUNER_MODE mode;
+  std::string file_path;
+} nnfw_cltunerinfo;
 
 /**
  * @brief Create a new session instance.
@@ -478,6 +497,19 @@ NNFW_STATUS nnfw_set_available_backends(nnfw_session *session, const char *backe
  * @return @c NNFW_STATUS_NO_ERROR if successful
  */
 NNFW_STATUS nnfw_set_op_backend(nnfw_session *session, const char *op, const char *backend);
+
+/**
+ * @brief     Set the OpenCL Tuner Configuration
+ *
+ * This function should be called before {@link nnfw_prepare} is invoked.
+ *
+ *
+ * @param[in] session session to set OpenCL Tuner Configuration
+ * @param[in] cltuner cltuner to configure OpenCL Tuner Configuration Value
+ *
+ * @return @c NNFW_STATUS_NO_ERROR if successful
+ */
+NNFW_STATUS nnfw_set_tuner(nnfw_session *session, nnfw_cltunerinfo *cltuner);
 
 /**
  * @brief     Retrieve uint32 type of nnfw information for given information ID.
--- a/runtime/onert/api/src/nnfw_api.cc
+++ b/runtime/onert/api/src/nnfw_api.cc
@@ -317,6 +317,20 @@ NNFW_STATUS nnfw_set_op_backend(nnfw_session *session, const char *op, const cha
 }
 
 /*
+ * Set the OpenCL Tuner Configuration
+ *
+ * @param[in] session session to set OpenCL Tuner Configuration
+ * @param[in] cltuner cltuner to configure OpenCL Tuner Configuration Value
+ *
+ * @return NNFW_STATUS_NO_ERROR if successful
+ */
+NNFW_STATUS nnfw_set_tuner(nnfw_session *session, nnfw_cltunerinfo *cltuner)
+{
+  NNFW_RETURN_ERROR_IF_NULL(session);
+  return session->set_tuner(cltuner);
+}
+
+/*
  * Retrieve uint32 type of nnfw information for given information ID.
  *
  * @param[in] session session to be queried on
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -113,6 +113,8 @@ using CfgKeyValues = std::unordered_map<std::string, std::string>;
 
 bool loadConfigure(const std::string cfgfile, CfgKeyValues &keyValues)
 {
+  keyValues.clear();
+
   std::ifstream ifs(cfgfile);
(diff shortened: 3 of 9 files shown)


hseok-oh:
I don't know about OpenCL tuner yet, but it seems that `tuner_file` may different for each `app` and target machine. 

IMHO, one solution is including `tuner_file` in `nnpackage`. 
- `tuner_file`(s)
- `nnpackage`'s config file may have mapping table between `tuner_file` and target machine
  - `nnpackage` may have many `tuner_file`s for many different target machine.

daeinki:
> I don't know details about OpenCL tuner yet, but it seems that `tuner_file` may different for each `app` and target machine.
> 
> IMHO, one solution is including `tuner_file` in `nnpackage`.
> 
> * `tuner_file`(s)
> * `nnpackage`'s config file may have mapping table between `tuner_file` and target machine
>   
>   * `nnpackage` may have many `tuner_file`s for many different target machine.

nnpackage would be one of model packages Platform team has to release to public or product team.
So IMO, I think the tuner_file should be packaged in model package of Platform.

If the tuner_file is packaged in nnpackage then there should be N targets per N nnpackages for each model because a model file is same but many HW could exist.

Thanks,
Inki Dae

chunseoklee:
@daeinki  Is it possible to detect target devices during runtime ?

daeinki:
> @daeinki Is it possible to detect gpu configuration of target devices during runtime ?

Could you give me more details about the GPU configuration and why the GPU configuration is needed?
AFAIK, we could detect machine name though standard ABI - system file - but specific API for GPU information.

chunseoklee:
> If your question is that GPU configuration is need to select a tuned file corresponding to current HW in runtime 

Exactly. I just wonder that runtime can choose proper tuner_file during execution.

daeinki:
> > If your question is that GPU configuration is need to select a tuned file corresponding to current HW in runtime
> 
> Exactly. I just wonder that runtime can choose proper tuner_file during execution.

If ONERT wants to keep the tuner file in nnpackage then ONERT should have platform and HW dependency, which in turn, this will make ONERT complicated. So my opinion is to leave the platform dependency to Platform, and that ONERT focus on only the tuner file provided by Platform.

chunseoklee:
> So my opinion is to leave the platform dependency to Platform, and that ONERT focus on only the tuner file provided by Platform.

I prefer this one if feasible.

intom:
@hseok-oh @juitem Please check this PR

runtime/onert/api/include/nnfw.h
```diff
@@ -479,6 +498,19 @@ NNFW_STATUS nnfw_set_available_backends(nnfw_session *session, const char *backe
...
+ * This function should be called before {@link nnfw_prepare} is invoked.
+ *
+ *
+ * @param[in] session session to set OpenCL Tuner Configuration
+ * @param[in] cltuner cltuner to configure OpenCL Tuner Configuration Value
+ *
+ * @return @c NNFW_STATUS_NO_ERROR if successful
+ */
+NNFW_STATUS nnfw_set_tuner(nnfw_session *session, nnfw_cltunerinfo *cltuner);
+
```
hseok-oh, line 513: IMHO, It's not good solution to add API for backend-specific feature.

daeinki: In case of other open source project, ARMNN, such backend specific information can be delivered by user to internal module using backend option[1]. So user can choose a inference device, CPU or GPU, and ARMNN provides the option which can pass GPU specific data to internal module with the target device. And looks reasonable to me.

Does ONERT provide such interface? If any then could you provide this committer the guide? If not any then I think you could guide a proper way for it rather than just say no.

AFAIK, ONE is an official open source project. So I'd kindly recommend you to give some help to this committer to find a proper solution. That would be one of very important factors for the growth of the community.


[1] https://github.com/ARM-software/armnn/blob/branches/armnn_20_11/include/armnn/IRuntime.hpp#L115

hseok-oh: >In case of other open source project, ARMNN, such backend specific information can be delivered by user to internal module using backend option[1]. So user can choose a inference device, CPU or GPU, and ARMNN provides the option which can pass GPU specific data to internal module with the target device. And looks reasonable to me.

I don't disagree to introduce general-type backend configuration API. My comment means that it looks acl-cl (or OpenCL) backend specific API.

>Does ONERT provide such interface? If any then could you provide this committer the guide? If not any then I think you could guide a proper way for it rather than just say no.

We don't have interface for backend configuration, so I recommand you to use `nnfw_explemental.h` (experimental API) instead of `nnfw.h` as temporary solution. I also don't have any solution to resolve this issue.

@wateret Could you have any idea about API (or interface for backend configuration) design?

wateret: > I don't disagree to introduce general-type backend configuration API. My comment means that it looks acl-cl (or OpenCL) backend specific API.

I agree with this.

> I recommand you to use `nnfw_explemental.h`

And this is what I would recommend too 😄

wateret: Eventually I would prefer introducing a way to pass some general data to backend.

What I just came up is something like below. However I'm not sure about this would be a good solution in many respects.

```c
// Pass data pointer to the backend context of `backend_id`
NNFW_STATUS nnfw_set_backend_data(nnfw_session *session, const char *backend_id, void *data);
```

Is there a good practice when we do this kind of things?

wateret:
One thing I would you like to remind you is that "frontend"(a frontend implementation), "core" and "backend"(a backend implementation) are technically separate libraries. Which means that the "frontend" and "backend" can use "core" but "core" knows nothing about them. And "frontend" and "backend" cannot know each other.

runtime/onert/api/include/nnfw.h
wateret: What I am thinking is:

- The user(who uses NNFW API) knows what should go in `void *data`
- NNFW API Frontend and ONERT Core **knows nothing** but just pass the pointer to the ACL backend
- ACL backend knows the format of `void *data` and it got the pointer from Core, so make use of it

daeinki: > > In case of other open source project, ARMNN, such backend specific information can be delivered by user to internal module using backend option[1]. So user can choose a inference device, CPU or GPU, and ARMNN provides the option which can pass GPU specific data to internal module with the target device. And looks reasonable to me.
> 
> I don't disagree to introduce general-type backend configuration API. My comment means that it looks acl-cl (or OpenCL) backend specific API.
> 
> > Does ONERT provide such interface? If any then could you provide this committer the guide? If not any then I think you could guide a proper way for it rather than just say no.
> 
> We don't have interface for backend configuration, so I recommand you to use `nnfw_explemental.h` (experimental API) instead of `nnfw.h` as temporary solution. I also don't have any solution or idea to resolve this issue.
> 
> @wateret Could you have any idea about API (or interface for backend configuration) design?

@intom It may be a good way to introduce the interface for backend configuration instead of the API if ONERT has no any interface for it and other agree to this way. :) AFAIK, the backend configuration interface can be used for other purpose such as profiling and debugging.

runtime/onert/api/src/nnfw_api.cc
```diff
@@ -316,6 +316,20 @@ NNFW_STATUS nnfw_set_op_backend(nnfw_session *session, const char *op, const cha
...
+ *
+ * @param[in] session session to set OpenCL Tuner Configuration
+ * @param[in] cltuner cltuner to configure OpenCL Tuner Configuration Value
+ *
+ * @return NNFW_STATUS_NO_ERROR if successful
+ */
+NNFW_STATUS nnfw_set_tuner(nnfw_session *session, nnfw_cltunerinfo *cltuner)
+{
+  NNFW_RETURN_ERROR_IF_NULL(session);
+  return session->set_tuner(cltuner);
```
hasw7569, line 330: I think you should also check the tuner pointer.

PR #5660, opened 2021-01-13
[luci] Move Circle Shape/Type Inference Helper to local scope

llFreetimell:
Parent Issue : #5501

Until now, `CircleShapeInferenceHelper.h` and `CircleTypeInferenceHelper.h` were visible in `include`.
However, as they are no more needed to be visible, this commit will make them as invisible.


seanshpark:
(just comment) seems `Move CircleShapeInferenceHelper to local scope` would be better for the title :)

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #5646, opened 2021-01-12
[luci] Revise shape inference with unknown


seanshpark:
`loco::NodeShape infer_reshape(const luci::CircleReshape *node)` has comment 
```cxx
// One of the dimensions can have special value -1, meaning its actual value should be inferred.
```
and the related codes.

We now revisit shape with `unknown()`, so it whould be better to revisit related codes

seanshpark:
From #5643. @llFreetimell , if you are interested could you please take a look? Or if you are already looking at this, that's perfect! :)

seanshpark:
For `Reshape`, where this issue was created for
- `res/TensorFlowLiteRecipes/Reshape_001/test.recipe` has `-1` value and this is not like the `unknown` of shape.
- `meaning its actual value should be inferred.` meaning is somewhat broad but is correct.

seanshpark:
`Slice` checks `-1` value but there is no test recipe. Let's add test recipe for this.

llFreetimell:
> if you are interested could you please take a look? Or if you are already looking at this, that's perfect! :)

Thanks for calling me :D
Fortunately(?), next step of #5501 is implementation of new shape inference algorithm which consider unknown dimension as you said! There are so many operations to re-implement(upgrade) :-D

#5501 is almost finished and only analysis of expected side effect when we `unset()` unknown dimensions during import step is work in progress :)
As soon as #5501 is finished, I will make a new issue for handling this!

seanshpark:
ExpandDims axis may have axis value of -1 but there is no recipe for testing

seanshpark:
Close issue as there no more `-1` to check

PR #5663, opened 2021-01-13
[res] Add Slice_001 recipe

seanshpark:
This will add Slice_001 tflite recipe with negative begin.


seanshpark:
For https://github.com/Samsung/ONE/issues/5646#issuecomment-759132738

llFreetimell (approved):
LGTM

PR #5667, opened 2021-01-13
[luci] Make faster RemoveUnnecessarySplitPass

llFreetimell:
Since `RemoveUnnecessarySplitPass` does not generate new nodes, there is no reason to break after each nodes.
This commit will remove break and make `RemoveUnnecessarySplitPass` more faster.


mhs4670go (approved):
LGTM

struss (approved):
LGTM

Issue #5618, opened 2021-01-11
[Compiler][luci] Introduce FuseBatchNormWithConvPass


llFreetimell:
Draft : #5617 

### What?

Let's support fusing BatchNorm(Mul+Add) with Conv2D.
```
[CircleConv]
      |
    [mul]      --->    [CircleConv]
      |
    [add]
```

### Todo

- [x] Add testcase in `res` #5626
- [x] Introduce `FuseBatchNormWithConvPass` #5627 #5668
- [x] Enable `FuseBatchNormWithConvPass` in `CircleOptimizer` #5658
- [x] Introduce `--fuse_batchnorm_with_conv` option to `circle2circle` #5661
- [x] Introduce `--fuse_batchnorm_with_conv` option to `one-optimize` and `util.py` #5662
- [x] Enroll testcases in `circle2circle-dredd-recipe-test` #5666

seanshpark:
Please add `--fuse_batchnorm_with_conv` option also to `one-cmds`: `one-optionze` and `util.py`

llFreetimell:
Done!

PR #5668, opened 2021-01-13
[luci] Set after new const node is created in FuseBatchNormWithConvPass

llFreetimell:
Parent Issue : #5618

Even shape status is not set to newly created `CircleConst`, it works well.
However, setting shape status is more accurate and thus this commit will fix it.


mhs4670go:
@llFreetimell But it means there wasn't proper validation check that cover these cases. Doesn't it need to be fixed?

seanshpark (approved):
LGTM

llFreetimell:
@mhs4670go Hmm... this is like `assert`.
When we create `CircleConst`, we always set the rank and each dimensions.
It means that even if we do not set shape status, `CircleConst` is always valid.

mhs4670go:
@llFreetimell Of course we know the one we created always has valid shape. I mean we should have found these missing shape status in the first place. Related test might not be enough or there wasn't proper validation check. What should be added may be added later.

mhs4670go (approved):
LGTM

PR #5679, opened 2021-01-13
[luci] Enable tests for ExpandDims_004, Slice_001

seanshpark:
This will enable tests for ExpandDims_004, Slice_001.


seanshpark:
Related with #5646

mhs4670go (approved):
LGTM

Issue #5006, opened 2020-11-11
[onert] Refactor Backend API


wateret:
There are several components that consist of a backend. The components are used during compilation, each has its own role and used in a fixed stage. For details, See [Runtime API](https://github.com/Samsung/ONE/blob/66014afd334f8a618e99aa7196524376fc08542a/docs/runtime/backend-api.md) document.

However we've faced limitation with this approach.

- The compiler workflow(how it uses backend components) looks unnatural, it is somewhat awkwardly specialized for the ACL backends
- We are going to introduce some more backends, but some of them do not fit the current workflow - need modification for backend workflow and API
- It limits backend implementation (e.g. limitation of optimization)

Let me illustrate the compilation workflow.

![BackendAPI (1)](https://user-images.githubusercontent.com/10582184/98778099-2d7d8700-2435-11eb-83d8-040cfd60c0db.png)

Please note:

- The left is as-is, the right is to-be
- BLUEs are the backend components that are used in each stage

**What I'm going to do is to merge 5 backend-dependent compilation tasks(the red box) into one single API.** And I expect it would make us more freely implement backends with a little code duplication.

I'm going to write a draft for this first, to find out if it is better than as-is and what might be obstacles for doing this. I also hope that the unified memory allocator(which @YongseopKim is going to work on) would make the backend API even simpler.

cc: @Samsung/nnfw

wateret:
Issues

- Unified API design : unit of the interface - LoweredGraph vs. OpSequence
    - `(CodeMap, TensorRegistry) ICompiler::codegen(LoweredGraph);`
    - `(Code, TensorRegistry) ICompiler::codegen(OpSequence);`
- One single API seems not possible due to reusing tensors from other backends (migrant tensors)
    - All the tensor objects must be registered for all operands first, then backends can generate kerenels

> What I'm going to do is to merge 5 backend-dependent compilation tasks(the red square) into one single API

Looks like I need to change the plan. "one single API" → "Two API functions"(One for tensor object building and the other for the rest).

YongseopKim:
> Unified API design : unit of the interface - LoweredGraph vs. OpSequence 

if it's possible not to need `LoweredInfo`, just `OpSequence` could be better.

wateret:
#5255 has landed. Now we don't use `ITensorBuilder`, `IConstantInitializer`, `IKernelGenerator` and `IOptimizer`(Not removed yet though).

```cpp
class BackendContext {
  ...
[...]

PR #5680, opened 2021-01-13
[DRAFT] [onert] Refactor BackendContext (except ACL backends)

wateret:
Pass a partial graph when creating a BackendContext. "partial graph"
here means a subset of an original graph that contains
operations/operands used by a backend. Before this commit, each backend
had a reference to the whole graph which could be abused and error-prone.

- Remove `initializeBackendContext` - as a partial graph is passed for
  each graph, we do not need to build used operation/operand lists
  anymore
- Make parameter of `Backend::newContext` simple - pass a single struct
  instead
- Make `BackendContext` methods(`genTensors` and `genKernels`) not have
  any parameters by giving every info to `backend::ContextData`.
- Add some more methods to Graph build partial graphs

More Context : #5006

### DIFF
--- a/runtime/onert/backend/CMakeLists.txt
+++ b/runtime/onert/backend/CMakeLists.txt
@@ -1,8 +1,8 @@
 set(LIB_ONERT_BACKEND_ACL_COMMON onert_backend_acl_common)
 
 add_subdirectory(cpu)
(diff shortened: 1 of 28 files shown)


wateret:
Tracked on #5773

PR #5682, opened 2021-01-13
[luci] Enable unset unknown dimension during import

llFreetimell:
Parent Issue : #5501

Until now, dimensions are always set to known.
This commit will enable unset the unknown dimensions during import step.

### DIFF
--- a/compiler/luci/import/src/CircleReader.cpp
+++ b/compiler/luci/import/src/CircleReader.cpp
@@ -262,6 +262,9 @@ void copy_tensor_attributes(const circle::TensorT &tensor, CircleNode *node)
   node->name(tensor_name(tensor));
   node->dtype(luci_datatype(tensor.type));
 
+  assert(tensor.shape_signature.size() == 0 ||
+         tensor.shape_signature.size() == tensor.shape.size());
+
   std::vector<int32_t> dims = tensor.shape; // in NHWC
   node->rank(dims.size());
   for (uint32_t r = 0; r < dims.size(); ++r)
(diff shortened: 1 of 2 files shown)


seanshpark (approved):
LGTM

compiler/luci/import/src/CircleReader.cpp
```diff
@@ -266,7 +266,10 @@ void copy_tensor_attributes(const circle::TensorT &tensor, CircleNode *node)
   node->rank(dims.size());
   for (uint32_t r = 0; r < dims.size(); ++r)
   {
-    node->dim(r) = loco::Dimension(dims[r]);
+    if (tensor.shape_signature.size() > 0 && tensor.shape_signature.at(r) == -1)
```
mhs4670go, line 269:
```suggestion
    if (tensor.shape_signature.size() == dims.size() && tensor.shape_signature.at(r) == -1)
```
more safe?

compiler/luci/import/src/Importer.cpp
```diff
@@ -101,7 +101,12 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
     const std::vector<int32_t> &input_dims = tensor.shape; // in NHWC
     input_shape->rank(input_dims.size());
     for (uint32_t r = 0; r < input_dims.size(); ++r)
-      input_shape->dim(r) = loco::Dimension(input_dims[r]);
+    {
+      if (tensor.shape_signature.size() > 0 && tensor.shape_signature.at(r) == -1)
```
mhs4670go, line 105: ditto

compiler/luci/import/src/CircleReader.cpp
llFreetimell: Hmm.. if then, it is more safe to add assert because size of shape_signature should be always 0 or `dims.size()`.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #5677, opened 2021-01-13
DRAFT: Constant folding for AddV2

jinevening:
On going draft to support constant folding for AddV2.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to : #5598

PR #5687, opened 2021-01-14
[luci] Pass to fold AddV2 operator

jinevening:
This adds a Pass to fold AddV2 operator.

---
Related to: #5598
Draft PR: #5677

### DIFF
--- a/compiler/luci/pass/src/FoldAddV2Pass.cpp
+++ b/compiler/luci/pass/src/FoldAddV2Pass.cpp
@@ -19,10 +19,11 @@
 #include <luci/IR/CircleNodes.h>
 
 #include <iostream>
+
 namespace
 {
 
-bool same_static_shape(const luci::CircleNode *x, const luci::CircleNode *y)
+bool same_shape(const luci::CircleConst *x, const luci::CircleConst *y)
 {
   if (x->rank() != y->rank())
     return false;
@@ -37,14 +38,11 @@ bool same_static_shape(const luci::CircleNode *x, const luci::CircleNode *y)
 }
 
 /**
- * Fold AddV2 to const if
- *
- * 1. it has two inputs
- * 2. both inputs are const
- *
+ * Fold AddV2 to const if both inputs are const
  **/
 template <loco::DataType T> bool fold_add_v2(luci::CircleCustom *add_v2)
 {
+  // This should hold for AddV2
   if (add_v2->numInputs() != 2)
     return false;
 
@@ -61,7 +59,7 @@ template <loco::DataType T> bool fold_add_v2(luci::CircleCustom *add_v2)
   if (x->dtype() != y->dtype())
     return false;
 
-  if (!same_static_shape(x, y))
+  if (!same_shape(x, y))
     return false;
 
   auto constant = add_v2->graph()->nodes()->create<luci::CircleConst>();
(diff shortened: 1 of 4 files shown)


compiler/luci/pass/src/FoldAddV2Pass.cpp
```diff
@@ -0,0 +1,119 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FoldAddV2Pass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+#include <iostream>
+namespace
```
llFreetimell, line 22: For readability :)

```suggestion

namespace
```

compiler/luci/pass/src/FoldAddV2Pass.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return false;
+
+  for (uint32_t i = 0; i < x->rank(); i++)
+  {
+    if (!(x->dim(i) == y->dim(i)))
+      return false;
+  }
+
+  return true;
+}
```
llFreetimell, line 37:
```suggestion
} // namespace
```

compiler/luci/pass/src/FoldAddV2Pass.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Fold AddV2 to const if
+ *
+ * 1. it has two inputs
```
llFreetimell, line 42: Just curiosity, can `AddV2` have more than two inputs?

compiler/luci/pass/src/FoldAddV2Pass.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+#include "luci/Pass/FoldAddV2Pass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+#include <iostream>
+namespace
+{
+
+bool same_static_shape(const luci::CircleNode *x, const luci::CircleNode *y)
```
llFreetimell, line 25: Can you apply following change?

```suggestion
bool same_static_shape(const luci::CircleConst *x, const luci::CircleConst *y)
```

Since `CircleConst` always has static shape, it seems more proper when considering function name :)

compiler/luci/pass/src/FoldAddV2Pass.cpp
jinevening: This is not the end of namespace :)

compiler/luci/pass/src/FoldAddV2Pass.cpp
jinevening: I don't think so. This is a condition to check custom Op is AddV2. I think it would be better to remove it from the comment. Thanks.

compiler/luci/pass/src/FoldAddV2Pass.cpp
jinevening: > Since CircleConst always has static shape

If so, `same_shape` looks better for `CircleConst` (`static` is unnecessary).

llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

PR #5690, opened 2021-01-14
[circle2circle] Option to fold AddV2 operator

jinevening:
This adds an option to fold AddV2 operator.

---
Related to: #5598
Draft PR: #5677

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -64,7 +64,7 @@ int entry(int argc, char **argv)
     .nargs(0)
     .required(false)
     .default_value(false)
-    .help("This will fold AddV2 operator");
+    .help("This will fold AddV2 operators with constant inputs");
 
   arser.add_argument("--fold_dequantize")
     .nargs(0)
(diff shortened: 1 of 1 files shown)


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -60,6 +60,12 @@ int entry(int argc, char **argv)
   arser.add_argument("--all").nargs(0).required(false).default_value(false).help(
     "Enable all optimize options");
 
+  arser.add_argument("--fold_add_v2")
+    .nargs(0)
+    .required(false)
+    .default_value(false)
+    .help("This will fold AddV2 operator");
```
seanshpark, line 67: (optional) would it be better to express like `This will fold AddV2 operator with both inputs constant` ?

jinevening, line 67:
```suggestion
    .help("This will fold AddV2 operators with constant inputs");
```

jinevening: I've updated the help message. Thanks.

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

seanshpark:
For https://github.com/Samsung/ONE/issues/5598#issuecomment-759932241, I would like to wait for the direction...

Issue #5599, opened 2021-01-08
Compiler FE: Transform minimum-maximum pattern to Relu6


jinevening:
## What

Let's transform max(min(input, 6), 0) to Relu6

![image](https://user-images.githubusercontent.com/5449554/103969788-11cefc80-51aa-11eb-82ea-ba12d223e038.png)

## Why

onnx-tf generates the above pattern for Relu6. We can transform it to a single operator Relu6.
Additionally, this transformation gives a chance for FuseActivationFunctionPass to fuse Relu6 with a preceding operator such as Conv.

seanshpark:
close?

jinevening:
> close?

@mhs4670go did all the jobs for this issue. @mhs4670go  Can you please close this if everything is done?

PR #5692, opened 2021-01-14
[TensorFlowLiteRecipes] Add Net_Maximum_Minimum_000 recipe

mhs4670go:
This commit adds Net_Maximum_Minimum_000 recipe.

FYI, this network will be fused to single relu6 operator.

Related : #5599

### DIFF
--- a/res/TensorFlowLiteRecipes/Net_Maximum_Minimum_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/Net_Maximum_Minimum_000/test.recipe
@@ -1,0 +1,86 @@
+operand {
+  name: "Const"
+  type: FLOAT32
+  shape {
+  }
+  filler {
+    tag: "explicit"
+    arg: "6"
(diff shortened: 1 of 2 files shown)


res/TensorFlowLiteRecipes/Net_Maximum_Minimum_000/test.recipe
```diff
@@ -0,0 +1,86 @@
...
+  is_variable: false
+}
+operand {
+  name: "Hole"
+  type: FLOAT32
+  shape {
+    dim: 1
+    dim: 16
+    dim: 160
+    dim: 160
```
seanshpark, line 36: Is this NCHW form?  Would be better to go with NHWC
And if we run this in our interpreter, would take some times, so it would be better to go with small values like 1,4,4,3

seanshpark (approved):
LGTM thank you!

PR #5693, opened 2021-01-14
[TensorFlowPythonModels] change driver mode bits

mhs4670go:
This commit changes driver(tfpem.py) mode bits - add execute.


seanshpark:
It would not unsafe to do this.

mhs4670go:
@seanshpark 
Actually, `res/TensorFlowPythonExamples/tfpem.py` has 755 permission and so do other scripts. Could you explain why?

seanshpark:
> has 755 permission and so do other scripts

I didn't notice this... 

> Could you explain why?

As I wrote, `It would not unsafe to do this.`.

Here's our choice
- we can +x to this file.
- we can remove -x to all python scripts

I couldn't find who add +x to `res/TensorFlowPythonExamples/tfpem.py` or maybe it was from the first.

Anyway adding +x makes it easy to run.

seanshpark:
Please add shebang...

mhs4670go:
@seanshpark You might mean `It would not safe to do this.`. Actually, I don't know why it is unsafe. I think if it is a library, it is natural to be no execute permission like `one-cmds/utils.py` but a script as driver is not. Anyway, both of two options are good.

seanshpark:
> It would not unsafe to do this..
> It would not safe to do this.

Argh :) "I would be unsafe to do this."... Today I'm a bit strange...

About unsafe, I'll tell you off-line :)

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

PR #5694, opened 2021-01-14
[luci] Test for FoldAddV2 pass

jinevening:
This adds tests for FoldAddV2 pass.

---
Related to: #5598
Draft PR: #5677

### DIFF
--- a/compiler/luci/pass/src/FoldAddV2Pass.test.cpp
+++ b/compiler/luci/pass/src/FoldAddV2Pass.test.cpp
@@ -1,0 +1,121 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

jinevening:
I found that I can improve the test code. Please do not merge for a while :)

jinevening:
@seanshpark Updated the code. PTAL :)

compiler/luci/pass/src/FoldAddV2Pass.test.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+} // namespace
+
+TEST_F(FoldS64AddV2Test, fold_addV2)
+{
+  luci::FoldAddV2Pass pass;
+  while (pass.run(&g))
+    ;
+
+  auto folded_const = dynamic_cast<luci::CircleConst *>(add->y());
```
seanshpark, line 109: Can you point out where `add` came from? it's a bit hard to find this...

jinevening: I understand it is a bit diffcult. It's from `ConstantFoldingTestGraph`, which is the parent class of `FoldAddV2Test`.

seanshpark: Not in this PR but I think it would be better to introduce `class ConstantFoldingAddTestGraph : public ConstantFoldingTestGraph` with only `CircleAdd` member...

seanshpark: And then rename members to have `_` prefix in `protected` scope...

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

mhs4670go (approved):
LGTM

compiler/luci/pass/src/FoldAddV2Pass.test.cpp
jinevening: I've made #5700.

PR #5698, opened 2021-01-14
[DRAFT] Introduce TransformMinMaxToRelu6Pass

mhs4670go:
This commit introduces TransformMinMaxToRelu6Pass for optimization.

Related : #5599 
ONE-DCO-1.0-Signed-off-by: seongwoo <mhs4670go@naver.com>

PR #5699, opened 2021-01-14
[luci] Introduce TransformMinMaxToRelu6Pass

mhs4670go:
This commit introduces TransformMinMaxToRelu6Pass to luci.

Related: #5599 
Draft: #5698 

### DIFF
--- a/compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
+++ b/compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
@@ -117,15 +117,8 @@ bool TransformMinMaxToRelu6Pass::run(loco::Graph *g)
   {
     if (auto maxi = dynamic_cast<luci::CircleMaximum *>(node))
     {
-      switch (maxi->dtype())
-      {
-        case loco::DataType::FLOAT32:
-          if (transform_min_max_pattern<loco::DataType::FLOAT32>(maxi))
-            changed = true;
-          break;
-        default:
-          return false;
-      }
+      if (transform_min_max_pattern<loco::DataType::FLOAT32>(maxi))
+        changed = true;
     }
   }
 
--- a/compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
+++ b/compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
@@ -1,0 +1,135 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/TransformMinMaxToRelu6Pass.h"
+
+#include "helpers/NodeFiller.h"
+#include "helpers/TypeMapper.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+template <loco::DataType DT>
+bool is_scalar_with_value(luci::CircleConst *node, typename loco::DataTypeImpl<DT>::Type val)
+{
+  if (node->dtype() != DT)
+    return false;
+  if (node->rank() != 0)
+    return false;
+  if (node->size<DT>() != 1)
+    return false;
+  if (node->at<DT>(0) != static_cast<typename loco::DataTypeImpl<DT>::Type>(val))
+    return false;
+
+  return true;
+}
+
+/**
+ *  BEFORE
+ *        [CircleNode]
+ *              |
+ *       [CircleMinimum]
+ *              |
+ *       [CircleMaximum]
+ *              |
+ *        [CircleNode]
+ *
+ *  AFTER
+ *
+ *        [CircleNode]
+ *              |
+ *        [CircleRelu6]
+ *              |
+ *        [CircleNode]
+ *
+ *  NOTE Only max(min(input, 6), 0) pattern will be transformed.
+ */
+template <loco::DataType DT> bool transform_min_max_pattern(luci::CircleMaximum *maxi)
+{
+  if (not maxi)
+    return false;
+
+  if (maxi->dtype() != DT)
+    return false;
+
+  luci::CircleConst *maxi_const = nullptr;
+  luci::CircleMinimum *mini = nullptr;
+
+  // There are two ways Maximum takes inputs.
+  // 1. Maximum(x = CircleConst, y = CircleMinimum)
+  // 2. Maximum(x = CircleMinimum, y = CircleConst)
+  if (not luci::fill(&maxi_const, &mini).with_commutative_args_of(maxi))
+    return false;
+
+  // Maximum constant should be scalar whose value is 0.
+  if (not is_scalar_with_value<DT>(maxi_const,
+                                   static_cast<typename loco::DataTypeImpl<DT>::Type>(0)))
+    return false;
+
+  luci::CircleConst *mini_const = nullptr;
+  loco::Node *mini_input = nullptr;
+
+  // There are two ways Miminum takes inputs.
+  // 1. Miminum(x = CircleNode, y = CircleMinimum)
+  // 2. Miminum(x = CircleMinimum, y = CircleNode)
+  if (not luci::fill(&mini_const, &mini_input).with_commutative_args_of(mini))
+    return false;
+
+  // Miminum constant should be scalar whose value is 6.
+  if (not is_scalar_with_value<DT>(mini_const,
+                                   static_cast<typename loco::DataTypeImpl<DT>::Type>(6)))
+    return false;
+
+  // Create Relu6 op
+  auto relu6 = mini->graph()->nodes()->create<luci::CircleRelu6>();
+  relu6->features(mini_input);
+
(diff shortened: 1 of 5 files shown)


seanshpark:
please fix title typo `[tluci]`

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.test.cpp
```diff
@@ -0,0 +1,139 @@
...
+/**
+ *  Optimized graph looks like below.
+ *
+ *  [CircleInput]
+ *        |
+ *  [CircleRelu6]
+ *        |
+ *  [CircleOutput]
+ */
+TEST(ReplaceMulAddWithDepthwiseConv, simple_test)
```
seanshpark, line 114: Could you please use `TEST_F`? (please refer #5694)
And change class members to `protected` score with members having `_` prefix?

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
```diff
@@ -0,0 +1,140 @@
...
+  if (maxi_const == nullptr || mini == nullptr)
+    return false;
+
+  // Maximum constant should be scalar whose value is 0.
+  if (maxi_const->rank() != 0)
+    return false;
+  if (maxi_const->size<loco::DataType::FLOAT32>() != 1)
+    return false;
+  if (maxi_const->at<loco::DataType::FLOAT32>(0) != 0.)
+    return false;
```
seanshpark, line 78: maybe extract this to a function?

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
```diff
@@ -0,0 +1,140 @@
...
+  if (mini_const == nullptr || mini_input == nullptr)
+    return false;
+
+  // Miminum constant should be scalar whose value is 6.
+  if (mini_const->rank() != 0)
+    return false;
+  if (mini_const->size<loco::DataType::FLOAT32>() != 1)
+    return false;
+  if (mini_const->at<loco::DataType::FLOAT32>(0) != 6.)
+    return false;
```
seanshpark, line 108: We can reuse this if extracted to a function

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
```diff
@@ -0,0 +1,140 @@
...
+    mini = dynamic_cast<luci::CircleMinimum *>(maxi->y());
+  }
+  // 2. Maximum(x = CircleMinimum, y = CircleConst)
+  else if (auto lhs = dynamic_cast<luci::CircleMinimum *>(maxi->x()))
+  {
+    maxi_const = dynamic_cast<luci::CircleConst *>(maxi->y());
+    mini = lhs;
+  }
+  else
+    return false;
```
seanshpark, line 67: This block can be reused also

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
```diff
@@ -0,0 +1,119 @@
...
+  // Maximum constant should be scalar whose value is 0.
+  if (not is_scalar_with_value(maxi_const, 0.))
+    return false;
+
+  luci::CircleConst *mini_const = nullptr;
+  loco::Node *mini_input = nullptr;
+
+  // // There are two ways Miminum takes inputs.
+  // // 1. Miminum(x = CircleNode, y = CircleMinimum)
+  // // 2. Miminum(x = CircleMinimum, y = CircleNode)
```
seanshpark, line 81: Unintended two `//` ?

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.test.cpp
```diff
@@ -0,0 +1,135 @@
...
+ *        |
+ *  [CircleRelu6]
+ *        |
+ *  [CircleOutput]
+ */
+TEST_F(MinMaxGraph, simple_test)
+{
+  luci::TransformMinMaxToRelu6Pass pass;
+  auto ret = pass.run(&_g);
+  EXPECT_EQ(true, ret);
```
seanshpark, line 118: Please use `EXPECT_TRUE`

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.test.cpp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+TEST_F(MinMaxGraph, wrong_condition_NEG)
+{
+  _maxi_const->at<loco::DataType::FLOAT32>(0) = 2.;
+
+  luci::TransformMinMaxToRelu6Pass pass;
+  auto ret = pass.run(&_g);
+
+  EXPECT_EQ(false, ret);
```
seanshpark, line 134: Please use `EXPECT_FALSE`

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "luci/Pass/TransformMinMaxToRelu6Pass.h"
+
+#include "helpers/NodeFiller.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+bool is_scalar_with_value(luci::CircleConst *node, float val)
```
jinevening, line 26: How about changing this function to `is_scalar_with_float_value` and adding a check whether node is float?

Current code may raise an error in `node->size<loco::DataType::FLOAT32>()` if the input model is quantized.

mhs4670go:
@seanshpark @jinevening PTAL:)
- EXPECT_EQ -> EXPECT_TRUE, EXPECT_FALSE
- make template to support various dtype

seanshpark:
Does this handle as you expect?
```
 [CircleInput] [CircleConst] [CircleConst] [CircleInput] 
          \         /              |      /
        [CircleMinimum]    [CircleMinimum] 
                 \           /
                [CircleMaximum]
                      |
                [CircleOutput]
```

mhs4670go:
@seanshpark Right
```cxx
// There are two ways Maximum takes inputs.
  // 1. Maximum(x = CircleConst, y = CircleMinimum)
  // 2. Maximum(x = CircleMinimum, y = CircleConst)
  if (not luci::fill(&maxi_const, &mini).with_commutative_args_of(maxi))
    return false;
```
`with_commutative_args_of` returns false except above two cases.

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp
```diff
@@ -0,0 +1,139 @@
...
+    if (auto maxi = dynamic_cast<luci::CircleMaximum *>(node))
+    {
+      switch (maxi->dtype())
+      {
+        case loco::DataType::FLOAT32:
+          if (transform_min_max_pattern<loco::DataType::FLOAT32>(maxi))
+            changed = true;
+          break;
+        case loco::DataType::U8:
+          if (transform_min_max_pattern<loco::DataType::U8>(maxi))
```
jinevening, line 127: u8 data type indicates that the model is quantized.

The value 0 and 6 in the float model may not be always 0 and 6 in the quantized model. They should be dequantized using quantization parameters. I think it would be better to implement FLOAT32 case only.

seanshpark:
@nnfw-bot test nncc-debug

seanshpark:
@nnfw-bot test nncc-release

seanshpark (approved):
LGTM thank you!

jinevening (approved):

seanshpark:
@mhs4670go , I'll wait for the changes from @jinevening suggestion.

compiler/luci/pass/src/TransformMinMaxToRelu6Pass.test.cpp
```diff
@@ -0,0 +1,135 @@
...
+/**
+ *  Optimized graph looks like below.
+ *
+ *  [CircleInput]
+ *        |
+ *  [CircleRelu6]
+ *        |
+ *  [CircleOutput]
+ */
+TEST_F(MinMaxGraph, simple_test)
```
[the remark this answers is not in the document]
mhs4670go: @seanshpark @jinevening 
I followed the suggestion to change `TEST` to `TEST_F` yesterday. But on second thought, it seems natural to use just `TEST` I think. As [document](https://github.com/YOU-i-Labs/googletest/blob/master/googletest/docs/Primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests), `TEST_F` is used to test the fixture we want to test. For example, say if you want to test a `Queue` class, you should write a `QueueTest` having `Queue` as an its attribute and then call `TEST_F(QueueTest, ...)`. 

Likewise, if we want to test `Pass` struct with `TEST_F`, we should write a `PassTest` class, which BTW, looks unnatural because `Pass` structs is more like method(wrapping `run`) rather than a fixture. I mean, current `TEST_F(MinMaxGraph, ...)` or `TEST_F(FoldS64AddV2Test, ...)` looks like it want to test `Graph` rather than its pass.

So, just
```cxx
TEST(..., ...)
{
  MinMaxGraph g;
  luci::TransformMinMaxToRelu6Pass pass;
  auto ret = pass.run(&g._g);
  ...
}
```
is enough I think.

Please give me your opinions.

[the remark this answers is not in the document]
seanshpark: As I know for this PR, you're adding `TransformMinMaxToRelu6Pass` and want to add a test for this class.
For me, `TransformMinMaxToRelu6PassTest` is natural for this point.

> Pass structs is more like method(wrapping run) rather than a fixture.

Pass has a simple structure with one method `run`.  I don't see any problem with this. 
You used word `fixture`, which I have difficulty understanding the meaning of this word in the sentence.

mhs4670go:
@seanshpark @jinevening PTAL:)

seanshpark (approved):
LGTM

jinevening (approved):
LGTM Thank you

PR #5705, opened 2021-01-14
[luci] Introduce substitutive function of loco shape/dtype getter

llFreetimell:
Parent issue : #5501

To deprecate `TypeInferencePass` and `ShapeInferencePass`, following functions should be removed.

- shape_get
- shape_known
- dtype_get
- dtype_known

Until now, replacing them as `rank` and `dim` of `CircleNode` was used.
However, `CircleShapeInferenceRule.cpp` and `CircleTypeInferenceRule.cpp` are using them so much.
As a result, if we do not introduce substitutive functions for them, code diff will be a lot.
Therefore, this commit will introduce substitutive functions for shape and dtype getter of loco.

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceHelper.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceHelper.cpp
@@ -18,6 +18,23 @@
 
 namespace luci
 {
+
+loco::NodeShape shape_get(const loco::Node *node)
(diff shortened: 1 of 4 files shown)


seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5707, opened 2021-01-15
DRAFT: Constant folding for Cast op

jinevening:
On going draft to support constant folding for Cast op.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/5598

PR #5708, opened 2021-01-15
[luci] Introduce a pass to fold Cast op

jinevening:
This introduces a pass to fold Cast op.

---
Related to: #5598
Draft PR: #5707

### DIFF
--- a/compiler/luci/pass/src/FoldCastPass.cpp
+++ b/compiler/luci/pass/src/FoldCastPass.cpp
@@ -21,18 +21,6 @@
 namespace
 {
 
-template <typename FromT, typename ToT> ToT safe_down_cast(FromT from)
-{
-  // Check overflow
-  if (from < std::numeric_limits<ToT>::lowest())
-    return std::numeric_limits<ToT>::lowest();
-
-  if (from > std::numeric_limits<ToT>::max())
-    return std::numeric_limits<ToT>::max();
-
-  return static_cast<ToT>(from);
-}
-
 luci::CircleConst *cast_const(luci::CircleConst *node, loco::DataType from_dtype,
                               loco::DataType to_dtype)
 {
@@ -58,7 +46,7 @@ luci::CircleConst *cast_const(luci::CircleConst *node, loco::DataType from_dtype
       constant->size<loco::DataType::S32>(num_elems);
       for (uint32_t i = 0; i < num_elems; i++)
         constant->at<loco::DataType::S32>(i) =
-          safe_down_cast<int64_t, int32_t>(node->at<loco::DataType::S64>(i));
+          static_cast<int32_t>(node->at<loco::DataType::S64>(i));
       return constant;
     }
     return nullptr;
--- a/compiler/luci/pass/src/FoldCastPass.cpp
+++ b/compiler/luci/pass/src/FoldCastPass.cpp
@@ -33,27 +33,6 @@ template <typename FromT, typename ToT> ToT safe_down_cast(FromT from)
   return static_cast<ToT>(from);
 }
 
-template <loco::DataType FromT, loco::DataType ToT>
-void write_downcasted_const(const luci::CircleConst *from, luci::CircleConst *to)
-{
-  assert(FromT == from->dtype());
-  assert(ToT == to->dtype());
-
-  // Resize destination const
-  const auto size = from->size<FromT>();
-  to->size<ToT>(size);
-
-  // TODO: Support more data types
-  if (FromT == loco::DataType::S64)
-  {
-    if (ToT == loco::DataType::S32)
-    {
-      for (uint32_t i = 0; i < size; i++)
-        to->at<ToT>(i) = safe_down_cast<int64_t, int32_t>(from->at<FromT>(i));
-    }
-  }
-}
-
 luci::CircleConst *cast_const(luci::CircleConst *node, loco::DataType from_dtype,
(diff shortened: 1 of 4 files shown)


compiler/luci/pass/src/FoldCastPass.cpp
```diff
@@ -0,0 +1,129 @@
...
+    constant->dim(i).set(node->dim(i).value());
+
+  constant->shape_status(luci::ShapeStatus::VALID);
+
+  // TODO: Support more data types
+  if (from_dtype == loco::DataType::S64)
+  {
+    if (to_dtype == loco::DataType::S32)
+    {
+      write_downcasted_const<loco::DataType::S64, loco::DataType::S32>(node, constant);
```
seanshpark, line 75: can we inline the codes here?
```cxx
   to->size<ToT>(from->size<FromT>());
   for (uint32_t i = 0; i < size; i++)
        to->at<ToT>(i) = safe_down_cast<int64_t, int32_t>(from->at<FromT>(i));
```
if we can remove either of `if (from_dtype == loco::DataType::S64) if (to_dtype == loco::DataType::S32)` here or in `write_downcasted_const` then I think it's worh separation but don't know for current codes...

jinevening: I've updated the code. It looks more concise. Thanks.

seanshpark (dismissed):
LGTM thank you!

compiler/luci/pass/src/FoldCastPass.cpp
```diff
@@ -0,0 +1,113 @@
...
+    if (to_dtype == loco::DataType::S32)
+    {
+      constant->size<loco::DataType::S32>(num_elems);
+      for (uint32_t i = 0; i < num_elems; i++)
+        constant->at<loco::DataType::S32>(i) =
+          safe_down_cast<int64_t, int32_t>(node->at<loco::DataType::S64>(i));
+      return constant;
+    }
+    return nullptr;
+  }
```
llFreetimell, line 63: If `num_elems` are calculated using `from_dtype`, the number of case would be increased later.
How about following codes?

```suggestion
  uint32_t num_elems = 1;
  for (uint32_t i = 0; i < node->rank(); ++i)
    num_elems *= node->dim(i).value();

  if (to_dtype == loco::DataType::S32)
  {
    constant->size<loco::DataType::S32>(num_elems);
    for (uint32_t i = 0; i < num_elems; i++)
      constant->at<loco::DataType::S32>(i) =
        safe_down_cast<int64_t, int32_t>(node->at<loco::DataType::S64>(i));
    return constant;
  }
```

compiler/luci/pass/src/FoldCastPass.cpp
```diff
@@ -0,0 +1,115 @@
...
+ */
+
+#include "luci/Pass/FoldCastPass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+template <typename FromT, typename ToT> ToT safe_down_cast(FromT from)
```
llFreetimell, line 24: I think this should not be applied.
When I executed following cmds, `safe_down_cast` is not we expected.
```python
import tensorflow as tf

a = tf.constant(21000000000, dtype=tf.int64)
b = tf.cast(a, dtype=tf.int32)
print(b) # tf.Tensor(-453397504, shape=(), dtype=int32)
```

compiler/luci/pass/src/FoldCastPass.cpp
jinevening: It looks good to calculate `num_elems` type-independently. I've updated the code. Thanks.

compiler/luci/pass/src/FoldCastPass.cpp
jinevening: Do you prefer not to check overflow just like TF?

seanshpark: Um... I think it would be better to get some advice from App part guys...

llFreetimell: Of course, if overflow is occurred in some model, most of it might be unintended.

However, even if overflow is unintended result, I think we should not change original graph.
The most important role of compiler frontend is **generating (optimized)circle model which is exactly same with original model.**
i.e., Introducing `safe_down_cast` means that `out1` and `out2` is same but it is not.

```python
import tensorflow as tf

inp = tf.constant(21000000000, dtype=tf.int64)
out1 = tf.cast(a, dtype=tf.int32)

min = tf.minimum(INT32_MAX, inp)
max = tf.maximum(INT32_MIN, min)
out2 = tf.cast(max, dtype=tf.int32)
```

Therefore, for me, `safe_down_cast` seems not proper :)

llFreetimell (approved):
LGTM

compiler/luci/pass/src/FoldCastPass.cpp
jinevening: > I think it would be better to get some advice from App part guys...

This is about how to handle a corner case, which does not happen in our target models, so App part may not be related to this issue. The casted value in our target model is dimension, which hardly exceeds int32 boundary.

jinevening: @llFreetimell Thanks for the comment :) It seems that I was a bit late to remove the comment.

seanshpark (approved):
LGTM

PR #5713, opened 2021-01-15
[luci] Add test for FoldCastPass (S64->S32)

jinevening:
This adds a test for FoldCastPass (S64->S32).

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: #5598
Draft PR: #5707

PR #5716, opened 2021-01-15
[luci/pass] Introduce SubstituteTransposeToReshape on luci/pass

struss:
For #5703 

This commit will introduce `SubstituteTransposeToReshape` on luci/pass.

### DIFF
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -44,7 +44,9 @@
 #include "luci/Pass/SparsifyTensorPass.h"
 #include "luci/Pass/ShuffleWeightTo16x1Float32Pass.h"
 #include "luci/Pass/SubstitutePackToReshapePass.h"
+#include "luci/Pass/SubstituteTransposeToReshapePass.h"
 #include "luci/Pass/TransformMinMaxToRelu6Pass.h"
+
 // TODO add more passes
 
 #include "luci/Pass/ShapeInferencePass.h"
@@ -258,6 +260,10 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   if (_options->query(Options::Algorithm::SubstitutePackToReshape))
   {
     phase.emplace_back(std::make_unique<luci::SubstitutePackToReshapePass>());
+  }
+  if (_options->query(Options::Algorithm::SubstituteTransposeToReshape))
+  {
+    phase.emplace_back(std::make_unique<luci::SubstituteTransposeToReshapePass>());
   }
   if (_options->query(Options::Algorithm::TransformMinMaxToRelu6Pass))
   {
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -46,7 +46,6 @@
 #include "luci/Pass/SubstitutePackToReshapePass.h"
 #include "luci/Pass/SubstituteTransposeToReshapePass.h"
 #include "luci/Pass/TransformMinMaxToRelu6Pass.h"
-
 // TODO add more passes
 
 #include "luci/Pass/ShapeInferencePass.h"
--- a/compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
+++ b/compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
@@ -25,10 +25,10 @@ namespace
  * @brief Convert transpose op in a certain condition to reshape op
  * @details Convert transpose op if it have condition below
  *          1. have a CircleConst perm value.
- *          2. input have a unknown shape less then 2
+ *          2. input have an unknown dimension less then 2
  *          3. the order of shape that except dim value 1 remains same on input and output
- *          eg) input shape  = (1, 2, 3, 4, 5, 1, 1, 1) => (2, 3, 4, 5)
- *              output shape = (2, 1, 3, 1, 1, 4, 5, 1) => (2, 3, 4, 5)
+ *             eg) input shape  = (126, 201, 1, 1) => (126, 201)
+ *                 output shape = (1, 126, 1, 201) => (126, 201)
  */
 bool substitute_transpose_to_reshape(luci::CircleNode *node)
 {
@@ -60,13 +60,13 @@ bool substitute_transpose_to_reshape(luci::CircleNode *node)
   {
     assert(perm_const->at<loco::DataType::S32>(i) >= 0 &&
            perm_const->at<loco::DataType::S32>(i) < static_cast<int32_t>(input_node->rank()));
-    if (input_node->dim(static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i))).known() &&
-        input_node->dim(static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i))).value() == 1)
+    const auto perm_value = static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i));
+    if (input_node->dim(perm_value).known() && input_node->dim(perm_value).value() == 1)
       continue;
     // To check idx values are increasing
-    if (idx > static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i)))
+    if (idx > perm_value)
       return false;
-    idx = static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i));
+    idx = perm_value;
   }
 
   auto new_const_node = node->graph()->nodes()->create<luci::CircleConst>();
--- a/compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
+++ b/compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
@@ -30,19 +30,15 @@ namespace
  *             eg) input shape  = (126, 201, 1, 1) => (126, 201)
  *                 output shape = (1, 126, 1, 201) => (126, 201)
  */
-bool substitute_transpose_to_reshape(luci::CircleNode *node)
+bool substitute_transpose_to_reshape(luci::CircleTranspose *node)
 {
-  auto target_node = dynamic_cast<luci::CircleTranspose *>(node);
-  if (target_node == nullptr)
-    return false;
-
-  auto perm_const = dynamic_cast<luci::CircleConst *>(target_node->perm());
+  auto perm_const = dynamic_cast<luci::CircleConst *>(node->perm());
   if (perm_const == nullptr)
     return false;
 
   assert(perm_const->dtype() == loco::DataType::S32);
 
-  auto input_node = loco::must_cast<luci::CircleNode *>(target_node->a());
+  auto input_node = loco::must_cast<luci::CircleNode *>(node->a());
   if (perm_const->dim(0).value() != input_node->rank())
     return false;
 
@@ -88,7 +84,7 @@ bool substitute_transpose_to_reshape(luci::CircleNode *node)
   new_reshape_node->tensor(input_node);
   new_reshape_node->shape(new_const_node);
 
-  replace(target_node).with(new_reshape_node);
+  replace(node).with(new_reshape_node);
   return true;
 }
 
@@ -120,10 +116,12 @@ bool SubstituteTransposeToReshapePass::run(loco::Graph *g)
   bool changed = false;
   for (auto node : loco::active_nodes(loco::output_nodes(g)))
   {
-    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
-    if (substitute_transpose_to_reshape(circle_node))
+    if (auto circle_node = dynamic_cast<luci::CircleTranspose *>(node))
     {
-      changed = true;
+      if (substitute_transpose_to_reshape(circle_node))
+      {
+        changed = true;
+      }
     }
   }
   return changed;
--- a/compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
+++ b/compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/SubstituteTransposeToReshapePass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/**
+ * @brief Convert transpose op in a certain condition to reshape op
+ * @details Convert transpose op if it have condition below
+ *          1. have a CircleConst perm value.
+ *          2. input have a unknown shape less then 2
+ *          3. the order of shape that except dim value 1 remains same on input and output
+ *          eg) input shape  = (1, 2, 3, 4, 5, 1, 1, 1) => (2, 3, 4, 5)
+ *              output shape = (2, 1, 3, 1, 1, 4, 5, 1) => (2, 3, 4, 5)
+ */
+bool substitute_transpose_to_reshape(luci::CircleNode *node)
+{
+  auto target_node = dynamic_cast<luci::CircleTranspose *>(node);
+  if (target_node == nullptr)
+    return false;
+
+  auto perm_const = dynamic_cast<luci::CircleConst *>(target_node->perm());
+  if (perm_const == nullptr)
+    return false;
+
+  assert(perm_const->dtype() == loco::DataType::S32);
+
+  auto input_node = loco::must_cast<luci::CircleNode *>(target_node->a());
+  if (perm_const->dim(0).value() != input_node->rank())
+    return false;
+  // If input have more than 2 unknown dimension, transpose will not changed.
+  int count = 0;
+  for (uint32_t i = 0; i < input_node->rank(); i++)
+    if (!input_node->dim(i).known())
+      count++;
+  if (count > 1)
+    return false;
+
+  uint32_t idx = 0;
+  for (uint32_t i = 0; i < perm_const->size<loco::DataType::S32>(); i++)
+  {
+    assert(perm_const->at<loco::DataType::S32>(i) >= 0 &&
+           perm_const->at<loco::DataType::S32>(i) < static_cast<int32_t>(input_node->rank()));
+    if (input_node->dim(static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i))).known() &&
+        input_node->dim(static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i))).value() == 1)
+      continue;
+    if (idx > static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i)))
+      return false;
+    idx = static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i));
+  }
+
+  auto new_const_node = node->graph()->nodes()->create<luci::CircleConst>();
+  new_const_node->dtype(loco::DataType::S32);
+  new_const_node->size<loco::DataType::S32>(perm_const->size<loco::DataType::S32>());
+  new_const_node->shape_status(luci::ShapeStatus::VALID);
+  new_const_node->rank(1);
+  new_const_node->dim(0).set(perm_const->size<loco::DataType::S32>());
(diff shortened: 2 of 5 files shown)


compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
```diff
@@ -0,0 +1,129 @@
...
+  auto perm_const = dynamic_cast<luci::CircleConst *>(target_node->perm());
+  if (perm_const == nullptr)
+    return false;
+
+  assert(perm_const->dtype() == loco::DataType::S32);
+
+  auto input_node = loco::must_cast<luci::CircleNode *>(target_node->a());
+  if (perm_const->dim(0).value() != input_node->rank())
+    return false;
+  // If input have more than 2 unknown dimension, transpose will not changed.
```
seanshpark, line 48: `transpose will not changed.` -> `transpose will not change` or `transpose will not  be changed.` or `transpose is not changed.`

seanshpark: Maybe `// We only permit for none or  only one unknown dimension`

struss: Oh, I will fix it.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+  uint32_t idx = 0;
+  for (uint32_t i = 0; i < perm_const->size<loco::DataType::S32>(); i++)
+  {
+    assert(perm_const->at<loco::DataType::S32>(i) >= 0 &&
+           perm_const->at<loco::DataType::S32>(i) < static_cast<int32_t>(input_node->rank()));
+    if (input_node->dim(static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i))).known() &&
+        input_node->dim(static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i))).value() == 1)
+      continue;
+    if (idx > static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i)))
```
seanshpark, line 64: Maybe add `// To check idx values are increasing` ?

struss: That's right. I will add that comment.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
```diff
@@ -0,0 +1,129 @@
...
+      return false;
+    idx = static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i));
+  }
+
+  auto new_const_node = node->graph()->nodes()->create<luci::CircleConst>();
+  new_const_node->dtype(loco::DataType::S32);
+  new_const_node->size<loco::DataType::S32>(perm_const->size<loco::DataType::S32>());
```
seanshpark, line 71: maybe add some variable and use it
```
auto size_items = perm_const->size<loco::DataType::S32>()
```

compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+} // namespace
+
+namespace luci
+{
+
+/**
+ *   BEFORE
+ *
+ * [CircleNode]  [CircleConst]
```
seanshpark, line 100: can you revise indentation like
```
 * BEFORE
 *
 *    [CircleNode]  [CircleConst]
```

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
```diff
@@ -0,0 +1,114 @@
...
+  // Output Connect.
+  auto output = g->nodes()->create<luci::CircleOutput>();
+  output->from(transpose_node);
+  auto graph_output = g->outputs()->create();
+  output->index(graph_output->index());
+}
+
+} // namespace
+
+TEST(SubstituteTransposeToReshapePass, simple_case)
```
seanshpark, line 64: Can you use `TEST_F` macro?  You can refer #5713 .

compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
struss: Okay, I will fix it.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
struss: Oh, I will apply that.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
struss: Okay, I will update after refer that.

jinevening: FYI, this is not constant folding, so `ConstantFoldingTestGraph` used in #5713 may not be directly applicable. There are some useful tutorials about gtest fixture written in Korean. (e.g., https://jakejang.tistory.com/39)

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
```diff
@@ -0,0 +1,123 @@
...
+  luci::SubstituteTransposeToReshapePass pass;
+  while (pass.run(&g))
+    ;
+
+  auto reshape_node = dynamic_cast<luci::CircleReshape *>(output->from());
+  auto transpose_node = dynamic_cast<luci::CircleTranspose *>(output->from());
+  ASSERT_NE(nullptr, reshape_node);
+  ASSERT_EQ(nullptr, transpose_node);
+  auto new_shape = loco::must_cast<luci::CircleConst *>(reshape_node->shape());
+  ASSERT_EQ(1, new_shape->at<loco::DataType::S32>(0));
```
seanshpark, line 102: Can you add some comments about how `{126, 201, 1, 1}` became `{1, 126, 1, 201}` ?

seanshpark: Does this related with the comment ?
```
 *          eg) input shape  = (1, 2, 3, 4, 5, 1, 1, 1) => (2, 3, 4, 5)
 *              output shape = (2, 1, 3, 1, 1, 4, 5, 1) => (2, 3, 4, 5)
```
If not, it would be nice to have some tests for this so that reading the test codes we can understand what this Pass does...

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
```diff
@@ -0,0 +1,123 @@
...
+  ASSERT_EQ(126, new_shape->at<loco::DataType::S32>(1));
+  ASSERT_EQ(1, new_shape->at<loco::DataType::S32>(2));
+  ASSERT_EQ(201, new_shape->at<loco::DataType::S32>(3));
+}
+
+TEST_F(SubstituteTransposeToReshapeTest_NEG, simple_case_NEG)
```
jinevening, line 108:
```suggestion
TEST_F(SubstituteTransposeToReshapeTest, simple_case_NEG)
```

jinevening, line 108: It would be better to use the same test name. One way to implement this is to build the graph in the function, not in the constructor. For example, 

```cpp
TEST_F(SubstituteTransposeToReshapeTest, simple_case)
{
  build_graph({2, 1, 3, 0});
 ...
}
TEST_F(SubstituteTransposeToReshapeTest, simple_case_NEG)
{
  build_graph({2, 0, 3, 1});
 ...
}
```

```suggestion
TEST_F(SubstituteTransposeToReshapeTest, simple_case_NEG)
```

compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/**
+ * @brief Convert transpose op in a certain condition to reshape op
+ * @details Convert transpose op if it have condition below
+ *          1. have a CircleConst perm value.
+ *          2. input have a unknown shape less then 2
```
llFreetimell, line 28:
```suggestion
 *          2. input have an unknown dimensions less then 2
```

compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
```diff
@@ -0,0 +1,132 @@
...
+  {
+    assert(perm_const->at<loco::DataType::S32>(i) >= 0 &&
+           perm_const->at<loco::DataType::S32>(i) < static_cast<int32_t>(input_node->rank()));
+    if (input_node->dim(static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i))).known() &&
+        input_node->dim(static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i))).value() == 1)
+      continue;
+    // To check idx values are increasing
+    if (idx > static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i)))
+      return false;
+    idx = static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i));
```
llFreetimell, line 69: (Optional)

```suggestion
    const auto perm_value = static_cast<uint32_t>(perm_const->at<loco::DataType::S32>(i));
    if (input_node->dim(perm_value).known() && input_node->dim(perm_value).value() == 1)
      continue;
    // To check idx values are increasing
    if (idx > perm_value)
      return false;
    idx = perm_value;
```

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
struss: > Does this related with the comment ?
> 
> ```
>  *          eg) input shape  = (1, 2, 3, 4, 5, 1, 1, 1) => (2, 3, 4, 5)
>  *              output shape = (2, 1, 3, 1, 1, 4, 5, 1) => (2, 3, 4, 5)
> ```
> 
> If not, it would be nice to have some tests for this so that reading the test codes we can understand what this Pass does...

This is a explanation of 3rd condition of this case. on real case, this can't be permit on transpose. I will fix it matching with our example.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
struss: Okay I will fix it.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
struss: > Can you add some comments about how `{126, 201, 1, 1}` became `{1, 126, 1, 201}` ?

okay I will add it.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
```diff
@@ -105,8 +91,15 @@ TEST_F(SubstituteTransposeToReshapeTest, simple_case)
...
-TEST_F(SubstituteTransposeToReshapeTest_NEG, simple_case_NEG)
+TEST_F(SubstituteTransposeToReshapeTest, simple_case_NEG)
 {
+  // Create graph that tranpose input {126, 201, 1, 1} with permutation {2, 1, 3, 0}
+  buildGraph({126, 201, 1, 1}, std::vector<int32_t>({2, 1, 3, 0}));
+  // on this input shape and permutation value, output shape will be {1, 201, 1, 126}
+  // the value in buffer also change as index become reversed
+  // for example, index 1 value in input will value.at[0][1][0][0], output will be
+  // value.at[0][0][0][1] which is same with input value.at[1][0][0][0]. so this transpose cannot
+  // convert to reshape.
```
seanshpark, line 102: it's a bit difficult to understand `value.at[0][1][0][0]` and `value.at[0][0][0][1]` ... maybe add some comments about `[0][1][0][0]` and `[0][0][0][1]` ?

> index 1 value in input will value.at[0][1][0][0]`

Something hard to understand... is this something like `input.at[0][1][0][0]` ?

> which is same with input value.at[1][0][0][0]

how does this work?

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
```diff
@@ -105,8 +91,15 @@ TEST_F(SubstituteTransposeToReshapeTest, simple_case)
   ASSERT_EQ(201, new_shape->at<loco::DataType::S32>(3));
 }
 
-TEST_F(SubstituteTransposeToReshapeTest_NEG, simple_case_NEG)
+TEST_F(SubstituteTransposeToReshapeTest, simple_case_NEG)
```
seanshpark, line 94: how is this test case negative?

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
llFreetimell: Maybe original intention of the comment is following...?
```
  // With this input shape and permutation values, output shape will be {1, 201, 1, 126}.
  // The sequence of values in buffer is also changed as index becomes reversed.
  // For example, output[1][2][3][4] is not input[1][2][3][4] but input[3][2][4][1].
  // So this Transpose op cannot be converted to Reshape op.
```

struss: > it's a bit difficult to understand `value.at[0][1][0][0]` and `value.at[0][0][0][1]` ... maybe add some comments about `[0][1][0][0]` and `[0][0][0][1]` ?
> 

I want to mean a value in  input tensor or output tensor with that position. 

> Something hard to understand... is this something like `input.at[0][1][0][0]` ?

I think those means only input value. but I think this make more confused. so, I will update comments with `input.at` and `output.at`

> how does this work?

On transpose operation, it will change a value shape with permutation value. on this comment, it will only target on this example, so it change shape {126, 201, 1, 1} to {1, 201, 1, 126}. so output.at[0][0][0][1] will be same with input.at[1][0][0][0].

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
struss: um...I think this is a case that will not make any change on. so, I  named as neg.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
seanshpark: I mean, there are two '1' in input shape and output shape, and you used index `[1]` 
If you can you can use something like `{32, 64, 16, 1}` then what shape comes from the output?

Order of `{126, 201, 1, 1} to {1, 201, 1, 126}` is `input - output` and ordef of `output.at[0][0][0][1] will be same with input.at[1][0][0][0]` is `output - input` : this made me a bit confusing. it would be better to explain in same order.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
seanshpark: > I think this is a case that will not make any change on

Then it would be better to add a comment about, something like what we expect but actually doesn't work like that...

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
jinevening: IMHO, the sentence after `for example,` makes things confusing. Actually this test is already an example of the algorithm, so the example in the comment seems redundant.

How about removing the example part as below ? (I slightly modified @llFreetimell  's suggestion)

``` cpp
  // With this input shape and permutation values, output shape will be [1, 201, 1, 126].
  // The order of non-one values is changed (126, 201) -> (201, 126).
  // So this Transpose op cannot be converted to Reshape op.
```

struss: > IMHO, the sentence after `for example,` makes things confusing. Actually this test is already an example of the algorithm, so the example in the comment seems redundant.
> 
> How about removing the example part as below ? (I slightly modified @llFreetimell 's suggestion)
> 
> ```c++
>   // With this input shape and permutation values, output shape will be [1, 201, 1, 126].
>   // The order of non-one values is changed (126, 201) -> (201, 126).
>   // So this Transpose op cannot be converted to Reshape op.
> ```

I think this comment is much more clear and explain what I want to say.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
struss: @seanshpark Umm....little bit confused. on this [comment](https://github.com/Samsung/ONE/pull/5457#issuecomment-752308952), the case that not meet activation condition is negative test. I think this case is also not meet condition intentionally.

seanshpark: > how is this test case negative?

This was a question about how this was a negative test. I wanted to know your idea. But instead of answering, you've removed the `_NEG`. I am also confused about this. I still hope to there is a negative test.

struss: > how is this test case negative?

@seanshpark Sorry for confusing and misunderstanding. I just thought that the question means that the case is not negative. I will update that case as `NEG`, as that case is intentionally make a condition for not changing.

jinevening (approved):
LGTM

llFreetimell (dismissed):
LGTM

compiler/luci/pass/src/SubstituteTransposeToReshapePass.test.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+  auto reshape_node = dynamic_cast<luci::CircleReshape *>(output->from());
+  auto transpose_node = dynamic_cast<luci::CircleTranspose *>(output->from());
+  ASSERT_EQ(nullptr, reshape_node);
+  ASSERT_NE(nullptr, transpose_node);
+  auto perm = loco::must_cast<luci::CircleConst *>(transpose_node->perm());
+  ASSERT_EQ(2, perm->at<loco::DataType::S32>(0));
+  ASSERT_EQ(1, perm->at<loco::DataType::S32>(1));
+  ASSERT_EQ(3, perm->at<loco::DataType::S32>(2));
+  ASSERT_EQ(0, perm->at<loco::DataType::S32>(3));
```
seanshpark, line 113: Thanks for then changes and the comment!

I think checking `perm` is not related with this test. 
As you wrote on the comment, we only need to check `CircleTranspose` stays as it is; `CircleReshape` cannot be found.

And I think `simple_case_NEG` should be renamed to reflect this: 
- how about `faild_to_substitute_NEG`

struss: Oh! thanks! I will fix it.

seanshpark (dismissed):
LGTM thank you!

jinevening (dismissed):

struss:
As this have a conflicts, I will rebase it.

struss:
@nnfw-bot test nncc-release

compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -44,7 +44,9 @@
 #include "luci/Pass/SparsifyTensorPass.h"
 #include "luci/Pass/ShuffleWeightTo16x1Float32Pass.h"
 #include "luci/Pass/SubstitutePackToReshapePass.h"
+#include "luci/Pass/SubstituteTransposeToReshapePass.h"
 #include "luci/Pass/TransformMinMaxToRelu6Pass.h"
+
```
seanshpark, line 49: Please remove this change

compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
```diff
@@ -0,0 +1,132 @@
...
+bool SubstituteTransposeToReshapePass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+    if (substitute_transpose_to_reshape(circle_node))
+    {
+      changed = true;
+    }
```
seanshpark, line 127: can you please apply like https://github.com/Samsung/ONE/pull/5745#discussion_r560796725 here too?

compiler/luci/pass/src/CircleOptimizer.cpp
struss: Okay I will remove it. Sorry.

compiler/luci/pass/src/SubstituteTransposeToReshapePass.cpp
struss: okay I will fix it

jinevening (approved):
LGTM

llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

PR #5721, opened 2021-01-16
[luci/pass] Introduce RemoveUnnecessaryStridedSlicePass

a-zh:
This commit adds pass for unnecessary StridedSlice operator

Draft: https://github.com/Samsung/ONE/pull/5876

Signed-off-by: Anton Zhukov <anton.v.zhukov@gmail.com>

### DIFF
--- a/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
@@ -74,8 +74,7 @@ bool remove_no_effect_strided_slice(luci::CircleStridedSlice *target_node)
    * We check additional attributes on zero after shapes
    * for skipping wrong StridedSlice operator.
    */
-  if (target_node->new_axis_mask() != 0 ||
-      target_node->shrink_axis_mask() != 0)
+  if (target_node->new_axis_mask() != 0 || target_node->shrink_axis_mask() != 0)
     return false;
 
   replace(target_node).with(input_node);
--- a/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
@@ -39,22 +39,19 @@ int64_t value_from_circle_const(const luci::CircleConst *node, uint32_t idx)
 
 bool remove_no_effect_strided_slice(luci::CircleStridedSlice *target_node)
 {
-  auto begin_const = loco::must_cast<luci::CircleConst *>(target_node->begin());
+  auto begin_const = dynamic_cast<luci::CircleConst *>(target_node->begin());
   if (begin_const == nullptr)
     return false;
 
-  auto strides_const = loco::must_cast<luci::CircleConst *>(target_node->strides());
+  auto strides_const = dynamic_cast<luci::CircleConst *>(target_node->strides());
   if (strides_const == nullptr)
     return false;
 
-  auto input_node = loco::must_cast<luci::CircleNode *>(target_node->input());
-  if (input_node == nullptr)
-    return false;
-
-  auto end_const = loco::must_cast<luci::CircleConst *>(target_node->end());
+  auto end_const = dynamic_cast<luci::CircleConst *>(target_node->end());
   if (end_const == nullptr)
     return false;
 
+  auto input_node = loco::must_cast<luci::CircleNode *>(target_node->input());
   for (uint32_t i = 0; i < input_node->rank(); i++)
   {
     if (value_from_circle_const(begin_const, i) != 0)
--- a/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
@@ -56,9 +56,6 @@ bool remove_no_effect_strided_slice(luci::CircleStridedSlice *target_node)
       return false;
 
     int64_t strides_value = value_from_circle_const(strides_const, i);
-    if (strides_value == -1)
-      continue;
-
     if (strides_value != 1)
       return false;
 
@@ -96,8 +93,11 @@ namespace luci
  *    [CircleNode]   [CircleStridedSlice]
  *
  * StridedSlice OP has no effect if,
- *    1. Static Shape : begin_const[idx] is 0 AND strides_const[idx] is (-1 OR input_dimension[idx])
- *    2. Dynamic Shape : begin_const[idx] is 0 AND strides_const[idx] is -1
+ *    1. Static Shape : begin_const[idx] is 0 AND strides_const[idx] is (not 1 OR input_dimension[idx])
+ *    2. Dynamic Shape : begin_const[idx] is 0 AND strides_const[idx] is not 1
+ *
+ * StridedSlice OP has effect if,
+ *    1. begin_const[idx] is 0 AND input_shape[idx] are equal to end_shape[idx]
  */
 bool RemoveUnnecessaryStridedSlicePass::run(loco::Graph *g)
 {
--- a/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
@@ -69,6 +69,15 @@ bool remove_no_effect_strided_slice(luci::CircleStridedSlice *target_node)
     if (!input_node->dim(i).known())
       return false;
   }
+
+  /**
+   * We check additional attributes on zero after shapes
+   * for skipping wrong StridedSlice operator.
+   */
+  if (target_node->new_axis_mask() != 0 ||
+      target_node->shrink_axis_mask() != 0)
+    return false;
+
   replace(target_node).with(input_node);
   return true;
 }
--- a/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
+++ b/compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
@@ -1,0 +1,115 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/RemoveUnnecessaryStridedSlicePass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/**
+ * @brief   Return value in CircleConst.
+ * @details Return value in position on CircleConst with int64 format.
+ *          Begin must be larger than or equal to 0. Size must be larger
+ *          than or equal to -1.
+ */
+int64_t value_from_circle_const(const luci::CircleConst *node, uint32_t idx)
+{
+  assert(node->rank() == 1 && node->dim(0).value() > idx);
+  assert(node->dtype() == loco::DataType::S64 || node->dtype() == loco::DataType::S32);
+
+  if (node->dtype() == loco::DataType::S64)
+    return node->at<loco::DataType::S64>(idx);
+  return static_cast<int64_t>(node->at<loco::DataType::S32>(idx));
+}
+
+bool remove_no_effect_strided_slice(luci::CircleNode *node)
+{
+  auto target_node = dynamic_cast<luci::CircleStridedSlice *>(node);
+  if (target_node == nullptr)
+    return false;
+  
+  auto begin_const = dynamic_cast<luci::CircleConst *>(target_node->begin());
+  if (begin_const == nullptr)
+    return false;
+
+  auto strides_const = dynamic_cast<luci::CircleConst *>(target_node->strides());
+  if (strides_const == nullptr)
+    return false;
+
+  auto input_node = loco::must_cast<luci::CircleNode *>(target_node->input());
+  if (input_node == nullptr)
+    return false;
+
+  for (uint32_t i = 0; i < input_node->rank(); i++)
+  {
+      if (value_from_circle_const(begin_const, i) != 0)
+          return false;
+
+      int64_t strides_value = value_from_circle_const(strides_const, i);
+      if (strides_value == -1)
+          continue;
+
+      if (strides_value != static_cast<int64_t>(input_node->dim(i).value()))
+          return false;
+
+      if (!input_node->dim(i).known())
+          return false;
+  }
+
+  replace(target_node).with(input_node);
+  return true;
+}
+
+} // namespace
+
+namespace luci
+{
+/**
+ * BEFORE
+ *
+ *    [CircleNode]
+ *          |
+ *    [CircleStridedSlice]
+ *          |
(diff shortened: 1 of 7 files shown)


seanshpark (changes requested):
Please split PRs that have different `compiler/modules`

seanshpark:
May I ask which issue is related to this Pass?

a-zh:
> May I ask which issue is related to this Pass? Please ping me from that issue.

Please, find related issue https://github.com/Samsung/ONE/issues/5757

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
```diff
@@ -43,34 +43,45 @@ bool remove_no_effect_strided_slice(luci::CircleNode *node)
   if (target_node == nullptr)
     return false;
 
-  auto begin_const = dynamic_cast<luci::CircleConst *>(target_node->begin());
+  auto begin_const = loco::must_cast<luci::CircleConst *>(target_node->begin());
```
seanshpark: Why is this chage needed?

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
```diff
@@ -43,34 +43,45 @@ bool remove_no_effect_strided_slice(luci::CircleNode *node)
...
 
+  auto end_const = loco::must_cast<luci::CircleConst *>(target_node->end());
+  if (end_const == nullptr)
+    return false;
+
   for (uint32_t i = 0; i < input_node->rank(); i++)
   {
-    if (value_from_circle_const(begin_const, i) != 0)
+    int64_t begin_value = value_from_circle_const(begin_const, i);
+    if (begin_value != 0)
```
seanshpark, line 65: why is this change needed?

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
```diff
@@ -0,0 +1,126 @@
...
+bool RemoveUnnecessaryStridedSlicePass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+    if (remove_no_effect_strided_slice(circle_node))
+    {
+      changed = true;
+    }
```
seanshpark, line 121: It would be better to use like
```cxx
  if (target_node = dynamic_cast<luci::CircleStridedSlice *>(node))
  {
     if (remove_no_effect_strided_slice(target_node))
      changed = true;
  }
```
and remove the same check in `remove_no_effect_strided_slice()`

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
```diff
@@ -0,0 +1,126 @@
...
+  if (input_node == nullptr)
+    return false;
+
+  auto end_const = dynamic_cast<luci::CircleConst *>(target_node->end());
+  if (end_const == nullptr)
+    return false;
+
+  for (uint32_t i = 0; i < input_node->rank(); i++)
+  {
+    int64_t begin_value = value_from_circle_const(begin_const, i);
```
seanshpark, line 64: `begin_value` is only used once to compare so it would be better not to declare,

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
```diff
@@ -0,0 +1,126 @@
...
+ *          |
+ *    [CircleStridedSlice]
+ *          |
+ *    [CircleNode]
+ *
+ * AFTER
+ *
+ *    [CircleNode]
+ *          |
+ *    [CircleNode]
```
seanshpark, line 106: `[CircleStridedSlice]` is not REMOVED. actually it's disconnected after this pass.
```
 *          |
 *    [CircleNode]
 *          |
 *    [CircleNode]   [CircleStridedSlice]
 *          |
```

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.test.cpp
```diff
@@ -0,0 +1,141 @@
...
+  // Output connect
+  auto output = g->nodes()->create<luci::CircleOutput>();
+  output->from(strided_slice);
+  auto graph_output = g->outputs()->create();
+  output->index(graph_output->index());
+}
+
+} // namespace
+
+TEST(RemoveUnnecessaryStridedSlicePass, remove_unnecessary_strided_slice)
```
seanshpark, line 85: Can you use `basic_case` for `remove_unnecessary_strided_slice` ?

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    auto strided_slice = dynamic_cast<luci::CircleStridedSlice *>(node);
+    if (strided_slice == nullptr)
+      continue;
+    strided_slice_node = strided_slice;
+    break;
+  }
+  ASSERT_EQ(nullptr, strided_slice_node);
+}
+
+TEST(RemoveUnnecessaryStridedSlicePass, remove_unnecessary_strided_slice_NEG)
```
seanshpark, line 114: can you use `basic_fail_case_NEG` ?

a-zh: Done

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.test.cpp
a-zh: Done

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
a-zh: Done

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
a-zh: Done

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
a-zh: Done

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
```diff
@@ -0,0 +1,120 @@
...
+  assert(node->dtype() == loco::DataType::S64 || node->dtype() == loco::DataType::S32);
+
+  if (node->dtype() == loco::DataType::S64)
+    return node->at<loco::DataType::S64>(idx);
+  return static_cast<int64_t>(node->at<loco::DataType::S32>(idx));
+}
+
+bool remove_no_effect_strided_slice(luci::CircleStridedSlice *target_node)
+{
+  auto begin_const = loco::must_cast<luci::CircleConst *>(target_node->begin());
```
seanshpark, line 42: `must_cast` will throw if input node type is not the template type. you should use plain dynamic_cast

a-zh: Done

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/**
+ * @brief   Return value in CircleConst.
+ * @details Return value in position on CircleConst with int64 format.
+ *          Begin must be larger than or equal to 0. Size must be larger
```
seanshpark, line 27: Cannot find `Begin` and `Size` in this method.

a-zh: Done

seanshpark:
@nnfw-bot test nncc-release

seanshpark (approved):
LGTM thank you!

struss:
Sorry for late reply. I tested on tensorflow v2.3 and have some question. while I create model like below, values and shapes are not matching with your conditions. 
Screenshot 2021-01-26 at 13 57 25
So, I think it is nice for this pass to add a `luci-pass-value-test` on it. 
for reference, https://github.com/Samsung/ONE/pull/5807

a-zh:
> Sorry for late reply. I tested on tensorflow v2.3 and have some question. while I create model like below, values and shapes are not matching with your conditions.

I see the modified tensor form after the current `strided_slice` operator. The previous shape is (3,2,3) and new shape is (0,2,3). And we got necessary `strided_slice`. And because the pass was skipped.

> So, I think it is nice for this pass to add a `luci-pass-value-test` on your draft.
> for reference, #5807

I will try to add a test case.

a-zh:
> > So, I think it is nice for this pass to add a `luci-pass-value-test` on your draft.
> > for reference, #5807
> 
> I will try to add a test case.

Please, find PRs:
https://github.com/Samsung/ONE/pull/5852
https://github.com/Samsung/ONE/pull/5853

a-zh:
@struss @seanshpark **PTL**

mhs4670go (approved):
LGTM

a-zh:
@struss @seanshpark Please take a look this PR.

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.test.cpp
```diff
@@ -0,0 +1,141 @@
...
+  {
+    strides->at<loco::DataType::S32>(i) = -1;
+  }
+
+  auto end = g->nodes()->create<luci::CircleConst>();
+  end->dtype(loco::DataType::S32);
+  end->size<loco::DataType::S32>(input_shape.size());
+  end->rank(1);
+  end->dim(0).set(input_shape.size());
+  end->shape(input_shape);
```
struss, line 66: I think this need to set value on end(CircleConst) at this.

a-zh: We have `end->shape(input_shape);`
Which one of another initializer are needed?

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.test.cpp
```diff
@@ -0,0 +1,141 @@
...
+  // Strides create
+  auto strides = g->nodes()->create<luci::CircleConst>();
+  strides->dtype(loco::DataType::S32);
+  strides->size<loco::DataType::S32>(input_shape.size());
+  strides->rank(1);
+  strides->dim(0).set(input_shape.size());
+  for (int i = 0; i < input_shape.size(); ++i)
+  {
+    strides->at<loco::DataType::S32>(i) = -1;
+  }
```
seanshpark, line 59:
```suggestion
  auto strides = g->nodes()->create<luci::CircleConst>();
  strides->dtype(loco::DataType::S32);
  strides->rank(1);
  strides->dim(0).set(input_shape.size());
  strides->size<loco::DataType::S32>(input_shape.size());
  for (int i = 0; i < input_shape.size(); ++i)
  {
    strides->at<loco::DataType::S32>(i) = -1;
  }
```

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.test.cpp
seanshpark, line 66:
```suggestion
  end->dtype(loco::DataType::S32);
  end->rank(1);
  end->dim(0).set(input_shape.size());
  end->size<loco::DataType::S32>(input_shape.size());
  for (int i = 0; i < input_shape.size(); ++i)
  {
    strides->at<loco::DataType::S32>(i) = ...;
  }
```
we don't need to set `end->shape(input_shape);` as `rank()` and `dim(..).set(..)` will set the shape.
you need to call ` strides->at<loco::DataType::S32>(i) = ...;` to set values of this const

a-zh: Thanks a lot. Fixed.

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.test.cpp
a-zh: Apply.

mhs4670go (dismissed):
LGTM

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * AFTER
+ *
+ *    [CircleNode]
+ *          |
+ *    [CircleNode]   [CircleStridedSlice]
+ *
+ * StridedSlice OP has no effect if,
+ *    1. Static Shape : begin_const[idx] is 0 AND strides_const[idx] is (-1 OR input_dimension[idx])
+ *    2. Dynamic Shape : begin_const[idx] is 0 AND strides_const[idx] is -1
```
struss, line 100: I think we need to add a comment about `end` value condition. follow from your logic above, end value is same with `(-1 OR input_dimension[idx])`

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
```diff
@@ -0,0 +1,115 @@
...
+  auto input_node = loco::must_cast<luci::CircleNode *>(target_node->input());
+  for (uint32_t i = 0; i < input_node->rank(); i++)
+  {
+    if (value_from_circle_const(begin_const, i) != 0)
+      return false;
+
+    int64_t strides_value = value_from_circle_const(strides_const, i);
+    if (strides_value == -1)
+      continue;
+
```
struss, line 61: this checking logic seems not right for this. on [Tensorflow explain](https://github.com/tensorflow/tensorflow/blob/85c8b2a817f95a3e979ecd1ed95bff1dc1335cff/tensorflow/python/ops/array_ops.py#L1129-L1133), Stride means 
```
  Roughly speaking, this op extracts a slice of size `(end-begin)/stride`
  from the given `input_` tensor. Starting at the location specified by `begin`
  the slice continues by adding `stride` to the index until all dimensions are
  not less than `end`.
  Note that a stride can be negative, which causes a reverse slice.
``` 
so, if `-1` in stride will make a result to reverse of input.

a-zh: OK. Done

compiler/luci/pass/src/RemoveUnnecessaryStridedSlicePass.cpp
a-zh: Done

a-zh:
@struss Please take a look PR

struss:
Different from `Slice`, `StridedSlice` need to check option params. (like shrink_axis_mask).

a-zh:
> Different from `Slice`, `StridedSlice` need to check option params. (like shrink_axis_mask).

I thought about it. But why? The sign for excluding the `StridedSlice` operator is only the equality of the input and output shape? And what test criteria is still required?

struss:
> > Different from `Slice`, `StridedSlice` need to check option params. (like shrink_axis_mask).
> 
> I thought about it. But why? The sign for excluding the `StridedSlice` operator is only the equality of the input and output shape? And what test criteria is still required?

I just think this logic not consider the case with option. if we don't consider this, it could make result not matching with original. so, I think this point need to be updated before something goes wrong.

a-zh:
@struss Please, take a look.

struss (approved):
LGTM! Sorry for late! Thanks!

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #5725, opened 2021-01-18
[luci] Revise temporary shape/dtype getter function

llFreetimell:
Parent Issue : #5501

During refactoring for deprecating original `ShapeInferencePass` and `TypeInferencePass`,
newly created nodes does not have shape and dtype of `CircleNode`.
However, if `luci::dtype_get` and `luci::shape_get` only returns shape and dtype of `CircleNode`,
wrong result would be returned.
Until the refactoring is finished, revise helper function to return correct result.

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceHelper.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceHelper.cpp
@@ -23,6 +23,10 @@ namespace luci
 
 loco::NodeShape shape_get(const loco::Node *node)
 {
+  // NOTE This function is subject to change after refactoring is finished.
+  //      If shape of CircleNode is returned, ShapeInferencePass may cause errors.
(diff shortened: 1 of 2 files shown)


seanshpark:
> Until the refactoring is finished, revise helper function to return correct result.

If this change is for temporary and needs update again, can you please add a comment about it?

llFreetimell:
@seanshpark Some comments are already in header file :D

https://github.com/Samsung/ONE/blob/b1f52ab55eedb7cd7db009112cdaf21c1840fa32/compiler/luci/service/src/CircleShapeInferenceHelper.h#L28-L35

Would it be better to add additional comments in this PR?

seanshpark:
> Would it be better to add additional comments in this PR?

Yes, that would be nice :)

seanshpark (approved):
LGTM thank you!

PR #5726, opened 2021-01-18
[luci] Remove unused headers in ResolveCustomOpMatMulPass

llFreetimell:
This commit will remove unused header files in `ResolveCustomOpMatMulPass`.

### DIFF
--- a/compiler/luci/pass/src/ResolveCustomOpMatMulPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpMatMulPass.cpp
@@ -23,8 +23,6 @@
 
 #include <loco.h>
 #include <oops/InternalExn.h>
-#include <loco/Service/ShapeInference.h>
-#include <loco/Service/TypeInference.h>
 
 namespace
 {


compiler/luci/pass/src/ResolveCustomOpMatMulPass.cpp
```diff
@@ -45,6 +43,8 @@ luci::CircleConst *create_const_node(loco::Graph *g, const loco::DataType dtype,
     size *= shape.at(i);
   }
 
+  node->shape_status(luci::ShapeStatus::VALID);
+
```
seanshpark, line 47:
```suggestion
  node->shape_status(luci::ShapeStatus::VALID);
  
```

seanshpark (dismissed):
LGTM thank you!

seanshpark:
It would be better to revise the title: `[luci] Fix trivial things in ...` doesn't seem to reflect the changes.
There are two different changes. Which would be better to separate, although they are trivial.

seanshpark (approved):
LGTM thank you!

PR #5727, opened 2021-01-18
[luci] Fix invalid type inference test code

llFreetimell:
`minimal_with_CircleRelu` test is for checking whether `relu_node` is inferred correctly.
However, as type of `relu_node` is set at first, the meaning is crahsed with pre-check.
i.e, type of `relu_node` is set to `S32` but `loco::dtype_known(relu_node)` should be `false` at pre-check.
This is definitely strange so this commit will fix it.


compiler/luci/service/src/CircleTypeInferenceRule.test.cpp
```diff
@@ -37,7 +37,6 @@ TEST(CircleTypeInferenceRuleTest, minimal_with_CircleRelu)
 
   // set dtype for nodes; like setting them in import
```
seanshpark, line 38:
```suggestion
  // set dtype for I/O nodes
```

seanshpark (approved):
LGTM thank you!

PR #5730, opened 2021-01-18
[luci] Substitute loco getter functino to luci getter function

llFreetimell:
Parent Issue : #5501

This commit will substitute `loco::shape_get` and `loco::dtype_get` functions to
`luci::shape_get` and `luci::dtype_get` functions.


seanshpark (approved):
LGTM

PR #5735, opened 2021-01-18
[luci] Introduce TypeMapper

mhs4670go:
This commit introduces TypeMapper to luci.

Sometimes, we need to use switch-case even if we use template because `CircleConst` methods needs explicit loco data type. Maybe this `TypeMapper` makes it easy to generalize these functions.

```cxx
template <typename T /* plain type */>
void foo(T v, luci::CircleConst c)
{
  assert(luci::TypeMapper<T>::get() == c->dtype());
  c->at<luci::TypeMapper<T>::get()>(0) = v;
  ...
}

```

### DIFF
--- a/compiler/luci/pass/src/helpers/TypeMapper.h
+++ b/compiler/luci/pass/src/helpers/TypeMapper.h
@@ -39,9 +39,39 @@ template <> struct TypeMapper<uint8_t>
   static constexpr loco::DataType get() { return loco::DataType::U8; }
 };
 
+template <> struct TypeMapper<uint16_t>
+{
+  static constexpr loco::DataType get() { return loco::DataType::U16; }
+};
+
+template <> struct TypeMapper<uint32_t>
+{
+  static constexpr loco::DataType get() { return loco::DataType::U32; }
+};
+
+template <> struct TypeMapper<uint64_t>
+{
+  static constexpr loco::DataType get() { return loco::DataType::U64; }
+};
+
+template <> struct TypeMapper<int8_t>
+{
+  static constexpr loco::DataType get() { return loco::DataType::S8; }
+};
+
 template <> struct TypeMapper<int16_t>
 {
   static constexpr loco::DataType get() { return loco::DataType::S16; }
+};
+
+template <> struct TypeMapper<int32_t>
+{
+  static constexpr loco::DataType get() { return loco::DataType::S32; }
+};
+
+template <> struct TypeMapper<int64_t>
+{
+  static constexpr loco::DataType get() { return loco::DataType::S64; }
 };
 
 } // namespace luci
--- a/compiler/luci/pass/src/helpers/TypeMapper.h
+++ b/compiler/luci/pass/src/helpers/TypeMapper.h
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <loco/IR/DataType.h>
+
+#include <cstdint>
+
+namespace luci
+{
+
+/**
+ * @brief TypeMapper maps between c++ primitive data type and loco::DataType.
+ */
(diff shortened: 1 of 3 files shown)


seanshpark:
Change seems OK. Where can I see applied codes?

mhs4670go:
@seanshpark For example, you may see `ResolveCustomOpMatMulPass.cpp`.
https://github.com/Samsung/ONE/blob/1e68ce706c32716044ac1543a2ec02e84b6a5296/compiler/luci/pass/src/ResolveCustomOpMatMulPass.cpp#L47-L71

This will be like this.
```cxx
...
node->size<luci::TypeMapper<T>::get()>(size);
for (uint32_t i = 0; i < values.size(); ++i)
  node->at<luci::TypeMapper<T>::get()>(i) = values[i];
...
```

seanshpark:
@mhs4670go , so you don't have any draft that uses this change?
If you want to land this change, could you prepare a DRAFT and see for yourself?

mhs4670go:
Draft: #5735

mhs4670go:
Actually, when I see https://github.com/Samsung/ONE/pull/5699#discussion_r558092350, I wanted to generalize that function like this.
```cxx
template <typename T> bool is_scalar_with_value(luci::CircleConst *node, T val)
{
  if (node->dtype() != luci::TypeMapper<T>::get())
    return false;
  if (node->rank() != 0)
    return false;
  if (node->size<luci::TypeMapper<T>::get()>() != 1)
    return false;
  if (node->at<luci::TypeMapper<T>::get()>(0) != static_cast<T>(val))
    return false;

  return true;
}
```

jinevening:
When I implement template functions with `loco::DataType`, I hoped that there exists a way to do conversion between native types and `loco::DataType`.  This is exactly what I wanted 👍

llFreetimell:
@mhs4670go Code itself looks fine to me but just one suggestion! How about adding this to `loco` ? Just for me, it seems okay because it is only related with `loco::DataType`, not `luci` :D

seanshpark:
> How about adding this to loco ? 

@parjong , any comments?

seanshpark:
> How about adding this to loco ?

My thought is that we can move to _loco_ when we are confident; there are several good usages and seems stable.

seanshpark:
@jinevening ,

>  I hoped that there exists a way to do conversion between native types and loco::DataType

This is for `CircleConst` content data conversion? or just `loco::DataType` <--> `circle::TensorType` type conversion?

jinevening:
> Is this CircleConst content data conversion? or just loco::DataType <--> circle::TensorType type conversion?

I think this can be used not only for accessing CircleConst contents but also in any case where type conversion from native type to loco::DataType is necessary. @mhs4670go Am I right?

mhs4670go:
@jinevening You're right. There could be many use cases.
@llFreetimell That would be good. It's up to @parjong :)

seanshpark:
> Draft: #5735

This is 5735

mhs4670go:
@seanshpark It's a mistake.. Draft is #5740

compiler/luci/pass/src/helpers/TypeMapper.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <loco/IR/DataType.h>
+
+#include <cstdint>
+
+namespace luci
+{
+
+/**
+ * @brief TypeMapper maps between plain type and loco::DataType.
```
seanshpark, line 25: maybe `plain type`  -> `c++ primitive data type` ?

jinevening (dismissed):
LGTM

(optional) It would be better to add more data types (at least for int32 and int64) to reduce future works.

compiler/luci/pass/src/helpers/TypeMapper.test.cpp
```diff
@@ -0,0 +1,33 @@
...
+TEST(TypeMapperTest, simple_test)
+{
+  EXPECT_EQ(loco::DataType::FLOAT32, luci::TypeMapper<float>::get());
+  EXPECT_EQ(loco::DataType::U8, luci::TypeMapper<uint8_t>::get());
+  EXPECT_EQ(loco::DataType::S16, luci::TypeMapper<int16_t>::get());
+}
+
+TEST(TypeMapperTest, wrong_condition_NEG)
+{
+  EXPECT_EQ(loco::DataType::Unknown, luci::TypeMapper<double>::get());
```
llFreetimell, line 32: If `TypeMapper<double>` is implemented in the future, this test can be failed.
How about this?
```suggestion
  EXPECT_EQ(loco::DataType::Unknown, luci::TypeMapper<char>::get());
```

mhs4670go: @llFreetimell Actually, I think it is unlikely that `double` will be introduced just as much as `char`. There is FLOAT64 just like there is INT 8. But I'll modify it for more safety.

llFreetimell (approved):
LGTM

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

Issue #5736, opened 2021-01-18
[one-cmds] one-build behavior when no virtual env


cgbahk:
I built and installed compiler 1e68ce70. It gave me `one-build` binary.

I didn't prepare venv yet, and ran `one-build` and it hang:
(Note `<<< sth` is my comment, not actual log)
```
$ ./one-build

Error: Virtual environment not found. Please run 'one-prepare-venv' command.
<<< hang here... command not returned
```

First I thought it was hang. But it turns out that it was waiting for my 'click'!

```
$ ./one-build

Error: Virtual environment not found. Please run 'one-prepare-venv' command.
<<< I clicked some place!
import-im6.q16: not authorized `argparse' @ error/constitute.c/WriteImage/1037.
<<< I clicked some place!
import-im6.q16: not authorized `configparser' @ error/constitute.c/WriteImage/1037.
<<< I clicked some place!
import-im6.q16: not authorized `os' @ error/constitute.c/WriteImage/1037.
<<< I clicked some place!
import-im6.q16: not authorized `subprocess' @ error/constitute.c/WriteImage/1037.
<<< I clicked some place!
import-im6.q16: not authorized `sys' @ error/constitute.c/WriteImage/1037.
<<< I clicked some place!
import-im6.q16: not authorized `_utils' @ error/constitute.c/WriteImage/1037.
build/install/bin/one-build: line 30: syntax error near unexpected token `('
build/install/bin/one-build: line 30: `def _get_parser():'
<<< Then it finally gave me a command prompt

$ echo $?
2
```

Not sure this is expected behavior

Of course it worked okay when I prepared venv before.

llFreetimell:
This issue is reproduced in my environment too :(

seanshpark:
@mhs4670go , can you take a look at this?

llFreetimell:
I found the reason :(
exit command wasn't in the code

llFreetimell:
This issue is resolved by #5737, close!

cgbahk:
Thanks
```
$ build/install/bin/one-build
Error: Virtual environment not found. Please run 'one-prepare-venv' command.

$ echo $?
255
```

PR #5737, opened 2021-01-18
[one-cmds] Add missing exit command

llFreetimell:
Parent Issue : #5736

Exit command was missed in `one-build` command.
This commit will add it.

### DIFF
--- a/compiler/one-cmds/one-build
+++ b/compiler/one-cmds/one-build
@@ -3,6 +3,7 @@
 ''''export PY_PATH=${SCRIPT_PATH}/venv/bin/python                                       # '''
 ''''test -f ${PY_PATH} && exec ${PY_PATH} "$0" "$@"                                     # '''
 ''''echo "Error: Virtual environment not found. Please run 'one-prepare-venv' command." # '''
+''''exit 255                                                                            # '''
 
 # Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
 #


mhs4670go (approved):
LGTM

compiler/one-cmds/one-build
```diff
@@ -3,6 +3,7 @@
 ''''export PY_PATH=${SCRIPT_PATH}/venv/bin/python                                       # '''
 ''''test -f ${PY_PATH} && exec ${PY_PATH} "$0" "$@"                                     # '''
 ''''echo "Error: Virtual environment not found. Please run 'one-prepare-venv' command." # '''
+''''exit 255                                                                            # '''
```
cgbahk, line 6: [Q] What is semantic of `'''`?

llFreetimell: @mhs4670go Can you explain for us...?

mhs4670go: @chbahk It's a little trick to combine bash + python into one file. `''''` is just two `''`, which means nothing in bash. And, after bash calls `PY_PATH`, it will be `''' ... '''`, which means comments in python.

seanshpark:
As @llFreetimell only posted for `one-build` but there exist other commands to update.
- @llFreetimell , you can revise all in this PR
- @mhs4670go, you can revise rest in another PR

seanshpark (approved):
LGTM

llFreetimell:
@seanshpark When I searched other one-cmds, `one-build` is the only one that exit code is missed. Which commands didn't I find?

compiler/one-cmds/one-build
cgbahk: Interesting :thinking:

cgbahk (approved):
LGTM

seanshpark:
> Which commands didn't I find?

I've found that all others have `exit 255` :)

compiler/one-cmds/one-build
cgbahk: So I called https://linux.die.net/man/1/import in bash :smile:

PR #5742, opened 2021-01-19
[luci] Fix assert in Concat shape inference rule

llFreetimell:
If one of dimensions in `axis` are unknown, it can also be accepted.
This commit will fix the assert in shape inference rule of `CircleConcat`.

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.cpp
@@ -549,7 +549,8 @@ loco::NodeShape infer_concatenation(const luci::CircleConcatenation *node)
       if (j == static_cast<uint32_t>(axis))
         output_shape.dim(j) = output_shape.dim(j).value() + input_shape.dim(j).value();
       else
-        assert(output_shape.dim(j) == input_shape.dim(j));
+        assert(!output_shape.dim(j).known() || !input_shape.dim(j).known() ||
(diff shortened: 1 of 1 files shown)


compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -549,7 +549,8 @@ loco::NodeShape infer_concatenation(const luci::CircleConcatenation *node)
       if (j == static_cast<uint32_t>(axis))
         output_shape.dim(j) = output_shape.dim(j).value() + input_shape.dim(j).value();
```
seanshpark, line 550: Then for this line, if we add conditions, would it be like this?
```cxx
assert(output_shape.dim(j).known() && input_shape.dim(j).known());
output_shape.dim(j) = output_shape.dim(j).value() + input_shape.dim(j).value();
```
if so, how about adding it?

seanshpark: It is because, using `known()` is added existing codes which we didn't check but for now, it would be better to check..

llFreetimell: I will add a comment about this :)

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

PR #5743, opened 2021-01-19
[luci] Update candidate nodes of shape/type inference

llFreetimell:
Parent Issue : #5501

Until now, traversal order was not considered with following assumptions.

- Imported circle model is correct
- Pass will create new nodes, not modifying original nodes

However, above assumptions were broken in some cases.
Therefore, this commit will update candidate nodes of shape/type inference
and thus infer shape and type correctly even if above assumptions were broken.

### DIFF
--- a/compiler/luci/pass/src/CircleShapeInferencePass.cpp
+++ b/compiler/luci/pass/src/CircleShapeInferencePass.cpp
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
+#include "helpers/InferenceCandidates.h"
+
 #include "luci/Pass/CircleShapeInferencePass.h"
 
-#include <luci/IR/DeadNodeQueryService.h>
 #include <luci/Service/CircleShapeInference.h>
 
 #include <loco.h>
@@ -67,9 +68,9 @@ bool CircleShapeInferencePass::run(loco::Graph *g)
   luci::sinf::Rule shape_infer_rule;
   bool changed = false;
 
-  for (auto node : loco::all_nodes(g))
-  {
-    if (!node->dialect()->service<DeadNodeQueryServiceImpl>()->isDeadNode(node))
+  // TODO Remove clang-format off
+  // clang-format off
+  for (auto node : inference_candidates(g))
     {
       loco::TensorShape shape;
       auto circle_node = loco::must_cast<luci::CircleNode *>(node);
@@ -85,7 +86,7 @@ bool CircleShapeInferencePass::run(loco::Graph *g)
         changed = true;
       }
     }
-  }
+  // clang-format on
 
   return changed;
 }
--- a/compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
+++ b/compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
@@ -1,0 +1,356 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/CircleShapeInferencePass.h"
+
+#include <loco.h>
+
+#include <luci/IR/CircleNodes.h>
+
+#include <gtest/gtest.h>
+
+/**
+ * This test is for checking when traversal order affect the result of inference.
+ *
+ * Because "transpose1" only changed permutation node from "old_perm" to "new_perm",
+ * shape_status of "transpose1" is still VALID even the shape should be changed.
+ * If "transpose2" is visted first before shape of "transpose1" is updated,
+ * "transpose2" can reference the shape of "relu" which is not updated yet.
+ * Then shape of "transpose2" becomes 3x5x5x1 and it causes an error at "conv2d".
+ *
+ * <Initial graph>
+ *                                                4x1x1x3
+ * [old_perm] ----------+              [filter] ----------+
+ * (0,2,1,3)            |                                 |
+ *                      |               [bias]  ----------+
+ *                      |                                 |
+ *  input  ------> [transpose1] ------> [relu] ------> [conv2d] ------>  output
+ *         1x5x5x3              1x5x5x3        1x5x5x3          1x5x5x4
+ *
+ *
+ * <Right after transformation>
+ *                                                                        4x1x1x3
+ * [new_perm] ----------+-----------------------------------+    [filter] ------+
+ * (3,2,1,0)            |                                   |                   |
+ *                      |                                   |      [bias] ------+
+ *                      |                                   |                   |
+ *  input  ------> [transpose1] ------> [relu] ------> [transpose2] ------> [conv2d] ------>  output
+ *         1x5x5x3              1x5x5x3        1x5x5x3                 ?             1x5x5x4
+ *
+ *
+ * <Expected result>
+ *                                                                        4x1x1x3
+ * [new_perm] ----------+-----------------------------------+    [filter] ------+
+ * (3,2,1,0)            |                                   |                   |
+ *                      |                                   |      [bias] ------+
+ *                      |                                   |                   |
+ *  input  ------> [transpose1] ------> [relu] ------> [transpose2] ------> [conv2d] ------>  output
+ *         1x5x5x3              3x5x5x1        3x5x5x1              1x5x5x3          1x5x5x4
+ *
+ */
+TEST(CircleShapeInferencePassTest, original_node_change)
+{
+  luci::CircleShapeInferencePass pass;
+  auto g = loco::make_graph();
+
+  // Have to be packed into lambda to check throw
+  auto shape_inference_run = [&]() {
+    while (pass.run(g.get()) == true)
+      ;
+  };
+
+  // Create nodes to make relu traversed first
+  auto input = g->nodes()->create<luci::CircleInput>();
+  auto relu = g->nodes()->create<luci::CircleRelu>();
+  auto old_perm = g->nodes()->create<luci::CircleConst>();
+  auto transpose1 = g->nodes()->create<luci::CircleTranspose>();
+  auto filter = g->nodes()->create<luci::CircleConst>();
+  auto bias = g->nodes()->create<luci::CircleConst>();
+  auto conv2d = g->nodes()->create<luci::CircleConv2D>();
+  auto output = g->nodes()->create<luci::CircleOutput>();
+  auto new_perm = g->nodes()->create<luci::CircleConst>();
+  auto transpose2 = g->nodes()->create<luci::CircleTranspose>();
+
+  // Build up initial graph
+  auto graph_input = g->inputs()->create();
+  graph_input->shape({1, 5, 5, 3});
+
+  input->index(graph_input->index());
+  input->shape({1, 5, 5, 3});
+  input->shape_status(luci::ShapeStatus::VALID);
+
+  old_perm->dtype(loco::DataType::S32);
+  old_perm->size<loco::DataType::S32>(4);
+  old_perm->shape({4});
+  old_perm->at<loco::DataType::S32>(0) = 0;
+  old_perm->at<loco::DataType::S32>(1) = 2;
+  old_perm->at<loco::DataType::S32>(2) = 1;
+  old_perm->at<loco::DataType::S32>(3) = 3;
+  old_perm->shape_status(luci::ShapeStatus::VALID);
+
+  transpose1->a(input);
+  transpose1->perm(old_perm);
+
+  relu->features(transpose1);
+
+  filter->dtype(loco::DataType::FLOAT32);
+  filter->size<loco::DataType::FLOAT32>(4 * 1 * 1 * 3);
+  filter->shape({4, 1, 1, 3});
+  filter->shape_status(luci::ShapeStatus::VALID);
+
+  bias->dtype(loco::DataType::FLOAT32);
+  bias->size<loco::DataType::FLOAT32>(4);
+  bias->shape({4});
+  bias->shape_status(luci::ShapeStatus::VALID);
+
+  conv2d->input(relu);
+  conv2d->filter(filter);
+  conv2d->bias(bias);
+  conv2d->padding(luci::Padding::VALID);
+  conv2d->stride()->h(1);
+  conv2d->stride()->w(1);
+  conv2d->dilation()->h(1);
+  conv2d->dilation()->w(1);
+
+  output->from(conv2d);
+  auto graph_output = g->outputs()->create();
+  output->index(graph_output->index());
+  graph_output->shape({1, 5, 5, 4});
+
+  ASSERT_NO_THROW(shape_inference_run());
+
+  // Transform graph
+  new_perm->dtype(loco::DataType::S32);
+  new_perm->size<loco::DataType::S32>(4);
+  new_perm->shape({4});
+  new_perm->at<loco::DataType::S32>(0) = 3;
+  new_perm->at<loco::DataType::S32>(1) = 2;
+  new_perm->at<loco::DataType::S32>(2) = 1;
+  new_perm->at<loco::DataType::S32>(3) = 0;
+  new_perm->shape_status(luci::ShapeStatus::VALID);
+
+  transpose1->perm(new_perm);
+
+  transpose2->a(relu);
+  transpose2->perm(new_perm);
+
+  conv2d->input(transpose2);
+
+  ASSERT_NO_THROW(shape_inference_run());
+
+  // Check result of shape inference is correct
+  ASSERT_EQ(3, transpose1->dim(0).value());
+  ASSERT_EQ(5, transpose1->dim(1).value());
+  ASSERT_EQ(5, transpose1->dim(2).value());
+  ASSERT_EQ(1, transpose1->dim(3).value());
+
+  ASSERT_EQ(3, relu->dim(0).value());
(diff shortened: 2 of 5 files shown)


seanshpark:
> Imported circle model is correct

Not quite exactly... there's a validate function that drops error or warnings...

> Pass will create new nodes, not modifying original nodes

I'm not sure this holds for all Passes and the was designed like this...

seanshpark:
> traversal order was not considered with following assumptions.

Can you describe more about `traversal order` and two items? I don't quite understand the relation.

seanshpark:
I think it would be better to describe what specific problem we have and solution(s) we can think of and what this change suggest.

compiler/luci/pass/src/CircleShapeInferencePass.cpp
```diff
@@ -67,23 +67,41 @@ bool CircleShapeInferencePass::run(loco::Graph *g)
...
+  //      This usually happens to mandatory virtual nodes such as CircleWhileOut, etc.
+  auto candidates = loco::postorder_traversal(loco::output_nodes(g));
   for (auto node : loco::all_nodes(g))
   {
+    // Already inserted in candidates
+    if (std::find(candidates.begin(), candidates.end(), node) != candidates.end())
+      continue;
+
+    // This node is definitely dead node
     if (!node->dialect()->service<DeadNodeQueryServiceImpl>()->isDeadNode(node))
```
mhs4670go, line 85: Maybe?
```suggestion
    if (node->dialect()->service<DeadNodeQueryServiceImpl>()->isDeadNode(node))
```

mhs4670go:
@llFreetimell IIUC, this PR isn't that difference practically. Original codes already do the same thing. Could you explain the main point of the needs to split it into two part?

llFreetimell:
@seanshpark 

> there's a validate function that drops error or warnings...

IIUC, for now, importer only validates the arguments of each operation such as the number of inputs and outputs.
(`CircleNodes` such as `CircleWhile` or `CircleIf` are validated with graph input/output)
My intention of the sentence "Imported circle model is correct" was "shape/type of all nodes in imported model is correct".

> I'm not sure this holds for all Passes and the was designed like this...

FYI, `ConvertNCHWToNHWCPass` are not creating new nodes but modifying original nodes.

> Can you describe more about traversal order and two items?

Let's suppose there is a following graph.
```
                     ---- [node2] ----
input --- [node1] ---+               +---- [node4] --- output
                     ---- [node3] ----
```

In case of `postorder_traversal`, the sequence would be node1 -> node2 -> node3 -> node4 (node2 and node3 can be changed).
  - If shape of node3 will be changed : shape of node4 is always inferred using new shape of node3.
  - If shape of node3 was incorrect when import : shape of node3 will be corrected using node1 and the corrected shape of node3 will be used for inferring shape of node4.

In case of `all_nodes`, the sequence is not considered. As a result traversal sequence is random.
Then what happens if traversal sequence is node4 -> node1 -> node2 -> node3
  - If shape of node3 will be changed : old shape of node3 will be used for inferring shape of node4. This cause incorrect result.
  - If shape of node3 was incorrect when import : incorrect shape of node3 will be used for inferring shape of node4. This cause incorrect result.
These errors are not detected because `shape_status()` is anyway set by `ShapeStatus::VALID`.
That's why setting `shape_status()` as `ShapeStatus::UNDEFINED` is used in `ConvertNCHWToNHWCPass`.

> I think it would be better to describe what specific problem we have and solution(s) we can think of and what this change suggest.

Hmm.. to commit message or to comments?

llFreetimell:
@mhs4670go 

Let's think following graph.
```
                     ---- [node2] ----                     +--- [CircleWhileOut1] -- Output
input --- [node1] ---+               +---- [CircleWhile] ---
                     ---- [node3] ----                     +--- [CircleWhileOut2]
```

Current codes can generate following sequences which may cause some errors.
- CircleWhile -> node1 -> node2 -> node3 -> CircleWhileOut2 -> CircleWhileOut1
- node1 -> node2 -> node3 -> CircleWhileOut2 -> CircleWhileOut1 -> CircleWhile
- etc...

However, if we apply this PR, only following two cases are generated.
- node1 -> node2 -> node3 -> CircleWhile -> CircleWhileOut1 -> CircleWhileOut2
- node1 -> node3 -> node2 -> CircleWhile -> CircleWhileOut1 -> CircleWhileOut2

seanshpark:
> Hmm.. to commit message or to comments?

The commit message, and if possible some comments at the code

seanshpark:
Previous shape inference was using this code
```cxx
bool ShapeInferenceSession::to(Graph *g) const
{
  assert(_rule->support(ShapeInferenceRule::API::V1) && "API v1 is unavailable");

  bool changed = false;

  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
  {
    if (_rule->recognize(node->dialect()))
    {
      loco::NodeShape shape;

      if (!shape_known(node) && inputs_shape_ready(node))
      {
        if (_rule->infer(node, shape))
        {
          node->annot(stdex::make_unique<ShapeAnnotation>(shape));
          changed = true;
        }
      }
    }
  }

  return changed;
}
```

I think you must have referenced this code.

But you created CircleShapeInferencePass.cpp  with `for (auto node : loco::all_nodes(g))`

This was the discussion: https://github.com/Samsung/ONE/pull/5529#discussion_r551080345

And now the code is somewhat strange... Does your new code work correctly? How do we know?

mhs4670go:
@llFreetimell I see your point.

[Caveat] Please understand I'm writing this comment with not enough consideration.

As far as the nodes that is not used for graph output exists, traverse order can't be correct correct because virtual nodes are not always at the end of the graph. And, `all_nodes` returns nodes in specific order that is same with the order we imports the graph in `convert_graph` method. It's quite organized, I mean, `node4 -> node1 -> node2 -> node3` is very unlikely to happen, which might why current compiler runs well.

As I said above, traverse order can't be correct correct because of second loop(could be correct but seems complex). It means we might as well think that traversal order is always random. `all_nodes` is not random random though as I said above. In other words, the result of candidates of this PR is random just as much as `all_nodes`.

Before I answer to questions you said above, please beware agian that `node4 -> node1 -> node2 -> node3` is very unlikely to happen.

> If shape of node3 will be changed : old shape of node3 will be used for inferring shape of node4. This cause incorrect result.

This is responsible for who writes its optimization. I think he/she should finish the optimization with valid shape.

> If shape of node3 was incorrect when import : incorrect shape of node3 will be used for inferring shape of node4. This cause incorrect result.

Actually it is very complicated things.. What if input shape is wrong? how can we know it is wrong or not? Anyway, it should be another topic I think..

llFreetimell:
> I think you must have referenced this code.

Yes I did.

> But you created CircleShapeInferencePass.cpp with for (auto node : loco::all_nodes(g))
> This was the discussion: #5529 (comment)
> And now the code is somewhat strange...

Yes, it is my mistake that I didn't considered following assumptions can be broken :(
- Imported circle model is correct (shape/type of all nodes in imported model is correct)
- Pass will create new nodes, not modifying original nodes

> Does your new code work correctly? How do we know?

I think new code will work but it was same thought when I wrote #5529.
So strictly speaking, there is possibility that this code does not work someday even though I think this code will work :-(
The only sure thing I can do my best is adding these error cases to test codes and not repeating same mistake...

seanshpark:
> Pass will create new nodes, not modifying original nodes

I still don't undestand this. what is the `modification` that breaks node traversal order?

> Until now, traversal order was not considered with following assumptions.
> 
> - Imported circle model is correct
> - Pass will create new nodes, not modifying original nodes

Do you mean `traversal order was considered with following assumptions.` ? I've removed the `NOT`

seanshpark:
> The only sure thing I can do my best is adding these error cases to test codes and not repeating same mistake...

Yes, we may make mistaks all the time. I think this is natural. Test is necessary.
- Please add some tests to for this
- Current diff is somewhat hard to read (but can uderstand reading slowly); I don't have a good idea but hope to get easy to read changes
- Current implementation is somewhat strange and complicated; you gave high level explanation but seems a bit too high with the problem-algorithm you want to apply.

llFreetimell:
@seanshpark 

Please forgive me, I am not good at clear explanation :(
How about following steps?

- Write some tests to show why this PR is created (Maybe CircleShapeInferencePass.test.cpp?)
- Write detailed comments for each test cases what was the problem in the test file.
- Answer for your question using the test cases again.

seanshpark:
> How about following steps?

Your steps seems nice. What I thought about this like
1) Create a test that fails : our problem
2) Create a patch that fixes the problem

How to make this easy? One idea that I got, maybe you may not like it;
1) Post a PR with the problem model (a .test.cpp or a recipe): CI should fail.
2) Post a PR with the problem model and the patch: CI should pass. We have a nice browser UI to navigate.

> Write detailed comments for each test cases what was the problem in the test file.

What is hard is that to much comments and that distracts understanding what the main point is. It would help with short key explanation in one sentence. The test code (or the model) will show the example.
Someone(like me) may say one doesn't understand. Then we can ADD(not EDIT) some detailed, not much explanation or a reference URL that he/she can read.

seanshpark:
It would be better to provide a function that returns `candidate` if this can be used in both shape/dtype.
```cxx
  auto candidates = loco::postorder_traversal(loco::output_nodes(g));
  for (auto node : loco::all_nodes(g))
  {
    // Already inserted in candidates
    if (std::find(candidates.begin(), candidates.end(), node) != candidates.end())
      continue;

    // This node is definitely dead node
    if (node->dialect()->service<DeadNodeQueryServiceImpl>()->isDeadNode(node))
      continue;

    candidates.emplace_back(node);
  }
  return candidates;
```

This code block first seemed strange.
Now with only this block like a method and with some simulation of what this code block does,
and imagination with some explanation comments, seems OK.

llFreetimell:
@seanshpark @mhs4670go 

I created some testcases.
However, if you are still hard to understand this PR, I will divide this PR more small...!

compiler/luci/pass/src/CircleOptimizerUtils.cpp
```diff
@@ -97,4 +98,23 @@ bool has_dynamic_shape(const loco::Node *node)
...
+      continue;
+
+    // This node is not used for both graph output and multiple output operation
+    if (node->dialect()->service<DeadNodeQueryServiceImpl>()->isDeadNode(node))
+      continue;
+
+    result.emplace_back(node);
+  }
+  return result;
+}
```
seanshpark, line 118: I don't think this is the right place for this method.
Please add a brief description of this method.

compiler/luci/pass/src/CircleOptimizerUtils.cpp
```diff
@@ -97,4 +98,23 @@ bool has_dynamic_shape(const loco::Node *node)
   return false;
 }
 
+std::vector<loco::Node *> alive_nodes(loco::Graph *g)
```
seanshpark, line 101: Is `alive` represent this method? Please explain.

compiler/luci/pass/src/CircleShapeInferencePass.cpp
```diff
@@ -67,23 +68,20 @@ bool CircleShapeInferencePass::run(loco::Graph *g)
...
-      auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+    loco::TensorShape shape;
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
 
-      if (shape_infer_rule.infer(circle_node, shape) && !is_same_shape(circle_node, shape))
-      {
-        circle_node->rank(shape.rank());
-        for (uint32_t i = 0; i < shape.rank(); ++i)
-          circle_node->dim(i) = shape.dim(i);
+    if (shape_infer_rule.infer(circle_node, shape) && !is_same_shape(circle_node, shape))
```
seanshpark, line 76: it would be better to use clang format ignore comment to reduce diff

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
```diff
@@ -0,0 +1,347 @@
...
+#include "luci/Pass/CircleShapeInferencePass.h"
+
+#include <loco.h>
+
+#include <luci/IR/CircleNodes.h>
+
+#include <gtest/gtest.h>
+
+/**
+ *  As transpose1 only changed attribute/input instead of creating new Transpose op,
```
seanshpark, line 26: Please start with `"original_node_change" test is about..."` to explain

compiler/luci/pass/src/CircleOptimizerUtils.cpp
llFreetimell: Then I will introduce `luci/IR/Algorithm.h` and define this function there like `loco/IR/Algorithm.h`.
Adding a brief description will be handled there too.

compiler/luci/pass/src/CircleOptimizerUtils.cpp
llFreetimell: I thought that `alive` represent this method because it enumerate all the nodes required to export graph.
If it seems strange, maybe `luci::active_nodes` can be applied?

llFreetimell:
After #5766 is merged, this PR will be rebased.

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
```diff
@@ -0,0 +1,354 @@
...
+#include "luci/Pass/CircleShapeInferencePass.h"
+
+#include <loco.h>
+
+#include <luci/IR/CircleNodes.h>
+
+#include <gtest/gtest.h>
+
+/**
+ * This test is for checking when original shape of node should be changed first.
```
seanshpark, line 26: does `original shape of node` ? mean `shape of input to a node` ? or is it `shape of a node before update` ?

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
```diff
@@ -0,0 +1,354 @@
...
+ *
+ * As transpose1 only changed attribute/input instead of creating new Transpose op,
+ * relu can be trversed first before shape of transpose1 is updated.
```
seanshpark, line 29: `relu` -> `ReLU`, `trversed ` -> `traversed`

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
```diff
@@ -0,0 +1,354 @@
...
+#include <loco.h>
+
+#include <luci/IR/CircleNodes.h>
+
+#include <gtest/gtest.h>
+
+/**
+ * This test is for checking when original shape of node should be changed first.
+ *
+ * As transpose1 only changed attribute/input instead of creating new Transpose op,
```
seanshpark, line 28: `As transpose1 only changed attribute/input` ? a bit strange...

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
```diff
@@ -0,0 +1,354 @@
...
+ * relu can be trversed first before shape of transpose1 is updated.
+ * If then shape of transpose2 becomes 3x5x5x1 and it causes an error at conv2d.
```
seanshpark, line 30: It's a bit hard to understand what the problem is and what the cause is.

compiler/luci/pass/src/CircleShapeInferencePass.cpp
```diff
@@ -67,9 +68,8 @@ bool CircleShapeInferencePass::run(loco::Graph *g)
   luci::sinf::Rule shape_infer_rule;
   bool changed = false;
 
-  for (auto node : loco::all_nodes(g))
-  {
-    if (!node->dialect()->service<DeadNodeQueryServiceImpl>()->isDeadNode(node))
+  // clang-format off
```
seanshpark, line 71: Please add `TODO: fix indentation`

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
llFreetimell: `shape of a node before update` is what I intended.
I will change the whole sentence to represent more specific goal of the test.

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
```diff
@@ -0,0 +1,356 @@
...
+#include <loco.h>
+
+#include <luci/IR/CircleNodes.h>
+
+#include <gtest/gtest.h>
+
+/**
+ * This test is for checking when traversal order affect the result of inference.
+ *
+ * Because "transpose1" only changed permutation node from "old_perm" to "new_perm",
```
seanshpark, line 28: >  Because "transpose1" only changed permutation node from "old_perm" to "new_perm",

This line is understood as "transpose1" will change permutation node by itself
- really?
- there is no when or why

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
```diff
@@ -0,0 +1,356 @@
...
+ * shape_status of "transpose1" is still VALID even the shape should be changed.
+ * If "transpose2" is visted first before shape of "transpose1" is updated,
```
seanshpark, line 30: `visted ` -> `visited ` ?

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
```diff
@@ -0,0 +1,356 @@
...
+#include "luci/Pass/CircleShapeInferencePass.h"
+
+#include <loco.h>
+
+#include <luci/IR/CircleNodes.h>
+
+#include <gtest/gtest.h>
+
+/**
+ * This test is for checking when traversal order affect the result of inference.
```
seanshpark, line 26: > This test is for checking when traversal order affect the result of inference.

Maybe "This test is to check traversal order of shape inference is done in topoligical order." ?

compiler/luci/pass/src/CircleShapeInferencepass.test.cpp
llFreetimell: Oops...? Then how about
`When perm() of "transpose1" is changed from "old_perm" to "new_perm" by some of luci/Pass,` ?

seanshpark: That's better :)

`When perm() of "transpose1" is changed from "old_perm" to "new_perm" by some of luci/Pass in below diagram,`

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

PR #5744, opened 2021-01-19
[luci] Introduce new inference test function

llFreetimell:
Parent Issue : #5501

Until now, `CircleTypeInferenceRule.test.cpp` and `CircleShapeInferenceRule.test.cpp`
were using `loco::apply` function and `loco` related inference rules such as
`CircleShapeInferenceRule` or `CircleTypeInferenceRule`.

This commit will introduce new inference test function which uses new inference rules
such as `luci::sinf::Rule` and `luci::tinf::Rule`.

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
@@ -20,6 +20,7 @@
 
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/CircleDialect.h>
+#include <luci/Service/CircleShapeInference.h>
 
 #include <loco.h>
 #include <loco/IR/CanonicalDialect.h>
@@ -46,6 +47,54 @@ bool shape_pass(loco::Graph *g)
     .bind(luci::CircleDialect::get(), &circle_rule);
 
   return loco::apply(&rules).to(g);
+}
+
+bool is_same_shape(luci::CircleNode *node, loco::TensorShape shape)
+{
+  if (node->shape_status() != luci::ShapeStatus::VALID)
+    return false;
+
+  if (node->rank() != shape.rank())
+    return false;
+
+  for (uint32_t i = 0; i < node->rank(); ++i)
+  {
+    if (node->dim(i).known() != shape.dim(i).known())
+      return false;
+
+    if (node->dim(i).value() != shape.dim(i).value())
+      return false;
+  }
+
+  return true;
+}
+
+// NOTE This function imitates CircleShapeInferencePass but little bit different.
+//      In CircleShapeInferencePass, DeadNodeQueryService is used to get alive nodes,
+//      which are not detected with postorder_traversal.
+//      However, it is not considered in this function because this is just for testing
+//      inference rule itself, not for inference pass.
+bool circle_shape_pass(loco::Graph *g)
+{
+  luci::sinf::Rule shape_infer_rule;
+  bool changed = false;
+
+  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
+  {
+    loco::TensorShape shape;
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+
+    if (shape_infer_rule.infer(circle_node, shape) && !is_same_shape(circle_node, shape))
+    {
+      circle_node->rank(shape.rank());
+      for (uint32_t i = 0; i < shape.rank(); ++i)
+        circle_node->dim(i) = shape.dim(i);
+      circle_node->shape_status(luci::ShapeStatus::VALID);
+      changed = true;
+    }
+  }
+
+  return changed;
 }
 
 } // namespace
--- a/compiler/luci/service/src/CircleTypeInferenceRule.test.cpp
+++ b/compiler/luci/service/src/CircleTypeInferenceRule.test.cpp
@@ -16,6 +16,7 @@
 
 #include "TestGraph.h"
 #include "CircleTypeInferenceHelper.h"
+#include <luci/Service/CircleTypeInference.h>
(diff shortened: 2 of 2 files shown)


compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
```diff
@@ -48,6 +49,54 @@ bool shape_pass(loco::Graph *g)
...
+
+  return true;
+}
+
+// NOTE This function imitates CircleShapeInferencePass but little bit different.
+//      In CircleShapeInferencePass, DeadNodeQueryService is used to get alive nodes,
+//      which are not detected with postorder_traversal.
+//      However, it is not considered in this function because this is just for testing
+//      inference rule itself, not for inference pass.
+bool circle_shape_pass(loco::Graph *g)
```
seanshpark, line 77: Where is this function called from?

compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
```diff
@@ -48,6 +49,54 @@ bool shape_pass(loco::Graph *g)
...
+  {
+    loco::TensorShape shape;
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+
+    if (shape_infer_rule.infer(circle_node, shape) && !is_same_shape(circle_node, shape))
+    {
+      circle_node->rank(shape.rank());
+      for (uint32_t i = 0; i < shape.rank(); ++i)
+        circle_node->dim(i) = shape.dim(i);
+      circle_node->shape_status(luci::ShapeStatus::VALID);
```
seanshpark, line 92: (optional) maybe add this as a function...  maybe not in this PR...
Where... in `luci/IR/CircleNodeShape.h` or `luci/Service/CircleNodeShape.h`
I think `luci/IR/CircleNodeShape.h` seems better. what do you think?

seanshpark: (optional)  `template <> class NodeMixin<NodeTrait::TensorShape>` in compiler/loco/include/loco/IR/NodeMixins.h,
we can add another shape setter like
```cxx
  void shape(std::initializer_list<uint32_t> dims)
  {
    rank(dims.size());

    uint32_t axis = 0;
    for (auto d : dims)
    {
      dim(axis++) = d;
    }
  }

  void shape(const loco::TensorShape &shape) or (const loco::TensorShape *shape)
  {
    ...
  }
```

compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
llFreetimell: This function will substitute `bool shape_pass(loco::Graph *g)` function to use `luci` shape at another PR.

compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
llFreetimell: If there is no problem to add following function at `loco/IR/NodeMixins.h`, it would be good :)

```cpp
  void shape(const loco::TensorShape &shape)
  {
    ...
  }
```

Both introducing `luci/IR/CircleNodeShape.h` and adding following function at `luci/IR/CircleNodeDecl.h` looks good to me!
```cpp
struct CircleNode : public loco::Node,
                    public loco::NodeMixin<loco::NodeTrait::DataType>,
                    public loco::NodeMixin<loco::NodeTrait::TensorShape>
{
  const loco::TensorShape &shape(void) const { return _shape; }
  void shape(const loco::TensorShape &shape) { rank(shape.rank); ... }
...
}
```

compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
seanshpark: If you don't know why I asked,
- new introduced codes are not tested by our CI
- there is no DRAFT link to understand the usage of this code

The code seems OK but the whole usage is unknown to me, so I can not agree with this chage by looking at this PR.

compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
seanshpark: > and adding following function at luci/IR/CircleNodeDecl.h looks good to me!

FYI, there is no `_shape`. `public loco::NodeMixin<loco::NodeTrait::TensorShape>` provides the attributes of `class NodeMixin<NodeTrait::TensorShape>`

compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
llFreetimell: Sorry for bothering you..

I created [Draft PR](https://github.com/Samsung/ONE/pull/5756/commits/c8d3066f0a2440b659579abef7ab6b7bfeb56bf2).
If I apply all of the changes in this PR at once, I thought that code diff would be very large and hard to review.
So I divided the changes as
- Introduce new function
- Enable new function and substitute related functions
I hope it helps your review...

seanshpark:
Thanks for the draft link.
`circle_shape_pass()` is to replace existing `shape_pass()` that was a helper function just to run shape inference.
What I don't agree is you want to introduce different code compared to the real one just for testing.

llFreetimell:
>What I don't agree is you want to introduce different code compared to the real one just for testing. 

To apply the comments, I replaced existing `shape_pass` with current `CircleShapeInferencePass` in the draft(https://github.com/Samsung/ONE/pull/5756/commits/1874cf2f3a5961c2ed743ab1c0a1bb96b0429691)
As a result, this PR is no longer needed.
Close this PR and create another PR...!

PR #5748, opened 2021-01-20
[luci] Rename FuseBatchNormWithTConv files

seanshpark:
This will rename FuseBatchNormWithTConv to FuseBatchNormWithTConvPass to
follow like other file naming convention.


mhs4670go (approved):
LGTM

PR #5749, opened 2021-01-20
[luci] Revise FuseAddWithTConvPass comment

seanshpark:
This will revise FuseAddWithTConvPass comment of before-after model
diagram.


seanshpark:
@jinevening , new diagram is my understanding from the code. If there is something incorrect, please add  some comments.

jinevening (approved):
LGTM

PR #5753, opened 2021-01-20
[DRAFT] Record profiling metadata

mhs4670go:
This darft is to record profiling metadata.

### DIFF
--- a/nnpackage/schema/circle_schema.fbs
+++ b/nnpackage/schema/circle_schema.fbs
@@ -1064,6 +1064,9 @@ table Operator {
   // computation, which may require relevant metadata such as quantization
   // parameters.
   intermediates:[int];
+
+  // For profiling
+  p_index:int;
 }
 
 // The root type, defining a subgraph, which typically represents an entire
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -226,6 +226,12 @@ int entry(int argc, char **argv)
     .default_value(false)
     .help("This will turn off operator validations. May help input model investigation.");
 
+  arser.add_argument("--generate_profiling_info")
+    .nargs(0)
+    .required(false)
+    .default_value(false)
+    .help("This will generates profiling information.");
+
   arser.add_argument("input").nargs(1).type(arser::DataType::STR).help("Input circle model");
   arser.add_argument("output").nargs(1).type(arser::DataType::STR).help("Output circle model");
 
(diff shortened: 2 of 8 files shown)


nnpackage/schema/circle_schema.fbs
```diff
@@ -1064,6 +1064,9 @@ table Operator {
   // computation, which may require relevant metadata such as quantization
   // parameters.
   intermediates:[int];
+
+  // For profiling
+  p_index:int;
```
seanshpark, line 1069: @mhs4670go , can you explain about this?

mhs4670go: @seanshpark
This draft is related tunnel T/F. For profiling, each operator in circle model has its id. That is why `p_index`(profiling_index) is added in schema. But since there is no exact decision yet about spec, I made this PR for experiment.

seanshpark: I know the background. I am asking why you added this item? from what requirement that ended up adding this.

mhs4670go: I come up with these two things to solve this requirement(https://github.com/Samsung/ONE/issues/5794#issuecomment-766536492). And this draft is one of them. It could be different implementation.

seanshpark:
@mhs4670go , 
I'd like to know what your requiremems are. either the requirement was from someone else or made by your self.
Can you describe it at the first comment? (you don't have to update the PR)
I'd like to understand your objective.

mhs4670go:
@seanshpark Please understand not being enough explanation. Because this draft is posted in the middle of discussion. I just wanted to show this draft to @llFreetimell at the time we discussed each other. 

```
TF/tflite/ONNX -> [one-import-x] -> circle1 -> [one-optimize] -> circle2 -> [one-quantize] -> circle3 -> [one-codegen] -> ~
```

Anyway, when I implemented this idea, I just imagined the way the profile data passes from @lemmaa's message in our internal messenger. First, T/F starts from `circle3`. And then if it landed well, we would be able to start from `circle1`. **I focused on making all the operators in the model carry its index.** That is my objective. There could be various solutions to solve this.

seanshpark:
Thanks for the explanation!

> I focused on making all the operators in the model carry its index

So this is what I wanted to ask. is this for reading `circle1` ? or is it for reading `circle3 ` ?

mhs4670go:
@seanshpark `circle3` as of now. It could expanded to `circle1` later because it consumes more time. It's just matter of the time left for us.

seanshpark:
> circle3 as of now. It could expanded to circle1 later because it consumes more time.

So, what is the answer to my question? 

> It's just matter of the time left for us.

I don't know why we should work with `circle1` as of now. I didn't heard of any requirements for this.

mhs4670go:
@seanshpark 
> So, what is the answer to my question?

`circle3`

> I don't know why we should work with circle1 as of now. I didn't heard of any requirements for this.

IIUC, the profiling is to know the number of execution cycle layer by layer. Since `circle3` is modified model by optimization(fusing or sth), the profiling should start from original model, i.e. `circle1`. So, our final goal is to deal with `circle1` and this journey(?) starts from `circle3`.

seanshpark:
For the deadline to make from `circle3` work in the T/F, do you think you have time to struggle with `circle1` ?

mhs4670go:
@seanshpark I think not. But if T/F has to do it, we can try it but would be less stability.

PR #5766, opened 2021-01-21
[luci] Introduce inference_candidates algorithm function

llFreetimell:
Originated from https://github.com/Samsung/ONE/pull/5743#discussion_r561351393

This commit will introduce `inference_candidates` function that enumerates all the nodes
which shape/dtype should be inferenced to export graph.

### DIFF
--- a/compiler/luci/pass/src/helpers/InferenceCandidates.test.cpp
+++ b/compiler/luci/pass/src/helpers/InferenceCandidates.test.cpp
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "InferenceCandidates.h"
+#include "luci/IR/CircleNode.h"
+
+#include <algorithm>
+
+#include <gtest/gtest.h>
+
+namespace
+{
+
+bool contains(const std::vector<loco::Node *> &vec, loco::Node *val)
+{
+  return std::any_of(vec.begin(), vec.end(), [val](loco::Node *node) { return node == val; });
+}
+
+} // namespace
+
+TEST(LuciPassHelpersInferenceCandidates, inference_candidates)
+{
+  auto g = loco::make_graph();
+
+  // Create nodes
+  auto input = g->nodes()->create<luci::CircleInput>();
+  auto split = g->nodes()->create<luci::CircleSplit>();
+  auto split_out1 = g->nodes()->create<luci::CircleSplitOut>();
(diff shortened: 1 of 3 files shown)


compiler/luci/lang/include/luci/IR/Algorithm.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include <loco.h>
+
+#include <vector>
+
+namespace luci
+{
+
+/**
+ * @brief Enumerate all the nodes which shape/dtype should be inferenced to export graph.
```
mhs4670go, line 28:
```suggestion
 * @brief Enumerate all the nodes whose shape/dtype should be inferenced to export graph.
```

mhs4670go (dismissed):
LGTM

seanshpark:
Let file names to express itself with its purpose would be better to understand and commicate: `Algorithm` is good but doesn't express the purpose of the method. 
from you comment, `InferenceCandidates` seems OK.

`inference_candidates` seems to be used from modules that use `luci/service` as you described in the comment.
and doesn't define the IR itself or related. To me this seems a helper function with the graph.
that is, it would better to move into `service`

llFreetimell:
@seanshpark 

Moving to `luci/Service` seems good.

Motivation of the file name `Algorithm.h` was `loco/IR/Algorithm.h` that collects algorithms for `locoIR`.
For now, the file has only `inference_candidates`, but some other functions may be added to this file in the future.
So how about `InferenceHelper.h` or `InferenceUtils.h`?

seanshpark:
> So how about InferenceHelper.h or InferenceUtils.h?

This will make a bunch of unordered methods in this file that may grow to a monster.
I would like to prevent this. 
There was similar PR: https://github.com/Samsung/ONE/pull/5640#pullrequestreview-565966357
Please read it.

llFreetimell:
After reading the comment, I could understand what you said.

Just one thing, if I create this file at `luci/Service`, I think that it is also creating a file with ambiguous purposes because it is not used in Service.
So is it okay to create as `luci/Pass/src/helpers/InferenceCandidates.h` ? `luci/Pass/src/helpers` folder is created recently.

seanshpark:
> So is it okay to create as luci/Pass/src/helpers/InferenceCandidates.h ?

Yes, this seems OK.

seanshpark (approved):
LGTM thank you!

compiler/luci/pass/src/helpers/InferenceCandidates.test.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+
+bool contains(const std::vector<loco::Node *> &vec, loco::Node *val)
+{
+  return std::any_of(vec.begin(), vec.end(), [val](loco::Node *node) { return node == val; });
+}
+
+} // namespace
+
+TEST(LuciPassHelpersInferenceCandidates, inference_candidates)
```
seanshpark, line 34: (In another PR, and if you have some time, try to revise this using `TEST_F`)

compiler/luci/pass/src/helpers/InferenceCandidates.test.cpp
```diff
@@ -0,0 +1,122 @@
...
+  ASSERT_EQ(6, s.size());
+  ASSERT_TRUE(contains(s, input));
+  ASSERT_TRUE(contains(s, split));
+  ASSERT_TRUE(contains(s, split_out1));
+  ASSERT_TRUE(contains(s, split_out2));
+  ASSERT_TRUE(contains(s, split_dim));
+  ASSERT_TRUE(contains(s, output));
+}
+
+TEST(LuciPassHelpersInferenceCandidates, inference_candidates_NEG)
```
seanshpark, line 75: (In another PR, let's try to rename this to reflect what this negative test does...)

mhs4670go (approved):
LGTM

PR #5761, opened 2021-01-21
DRAFT: FuseBatchNormWithDwConvPass

seanshpark:
on-going draft of FuseBatchNormWithDwConvPass and other Pass related
tests

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@samsung.com>

PR #5767, opened 2021-01-21
[luci-pass-value-test] Introduce Pass value test

seanshpark:
This will introduce luci-pass-value-test that will compare reference
values from execution of tflite and circle file that was generated with
certain optimization option.


seanshpark:
This PR is a bit huge but was mostly copied from existing _luci-value-test_  and some modification.
From draft #5761

mhs4670go (approved):
LGTM

PR #5769, opened 2021-01-21
[luci-value-test] Add gitignore

seanshpark:
This will add gitignore file to ignore local only file: test.local.lst


mhs4670go (approved):
LGTM

PR #5770, opened 2021-01-21
[luci] Revise FuseBatchNormWithTConvPass

seanshpark:
This will revise FuseBatchNormWithTConvPass to search pattern from
Add > Mul > TConv and insert new new nodes instead of updating it.


seanshpark:
From draft #5761

mhs4670go (approved):
LGTM

PR #5771, opened 2021-01-21
[luci] Replace loco functions to luci functions in InferenceRule tests

llFreetimell:
Parent Issue : #5501

In `CircleShapeInferenceRule.test.cpp` and `CircleTypeInferenceRule.test.cpp`,
loco functions are used.
This commit will replace those functions to luci functions.

### DIFF
--- a/compiler/luci/service/CMakeLists.txt
+++ b/compiler/luci/service/CMakeLists.txt
@@ -21,5 +21,6 @@ nnas_find_package(GTest REQUIRED)
 
 GTest_AddTest(luci_service_test ${TESTS})
 target_include_directories(luci_service_test PRIVATE src)
+target_link_libraries(luci_service_test luci_pass)
 target_link_libraries(luci_service_test luci_service)
 target_link_libraries(luci_service_test oops)
--- a/compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
@@ -20,12 +20,9 @@
 
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/CircleDialect.h>
+#include <luci/Pass/CircleShapeInferencePass.h>
 
 #include <loco.h>
-#include <loco/IR/CanonicalDialect.h>
-#include <loco/Service/ShapeInference.h>
-#include <loco/Service/CanonicalShapeInferenceRule.h>
-#include <loco/Service/MultiDialectShapeInferenceRule.h>
 
 #include <oops/InternalExn.h>
 
@@ -38,14 +35,8 @@ namespace
 
 bool shape_pass(loco::Graph *g)
 {
-  loco::CanonicalShapeInferenceRule canonical_rule;
-  luci::CircleShapeInferenceRule circle_rule;
-  loco::MultiDialectShapeInferenceRule rules;
-
-  rules.bind(loco::CanonicalDialect::get(), &canonical_rule)
-    .bind(luci::CircleDialect::get(), &circle_rule);
-
-  return loco::apply(&rules).to(g);
+  luci::CircleShapeInferencePass shape_pass;
+  return shape_pass.run(g);
 }
 
 } // namespace
@@ -72,7 +63,7 @@ TEST(CircleShapeInferenceRuleTest, minimal_with_CircleRelu)
   }
 
   // pre-check
-  ASSERT_FALSE(loco::shape_known(relu_node));
+  ASSERT_FALSE(luci::shape_known(relu_node));
 
   // shape inference
   while (shape_pass(graph.graph()) == true)
@@ -80,7 +71,7 @@ TEST(CircleShapeInferenceRuleTest, minimal_with_CircleRelu)
 
   // Verify
   {
-    ASSERT_TRUE(loco::shape_known(relu_node));
+    ASSERT_TRUE(luci::shape_known(relu_node));
     ASSERT_EQ(loco::Domain::Tensor, luci::shape_get(relu_node).domain());
 
     auto shape = luci::shape_get(relu_node).as<loco::TensorShape>();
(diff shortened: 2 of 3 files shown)


compiler/luci/service/CMakeLists.txt
```diff
@@ -21,5 +21,6 @@ nnas_find_package(GTest REQUIRED)
 
 GTest_AddTest(luci_service_test ${TESTS})
 target_include_directories(luci_service_test PRIVATE src)
+target_link_libraries(luci_service_test luci_pass)
```
seanshpark, line 24: sequence of loading in CMakeList.txt is
```
add_subdirectory(service)
add_subdirectory(pass)
```
which when processing this file, `luci_pass` maybe not unavailable.

compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
```diff
@@ -38,14 +35,8 @@ namespace
...
-  loco::CanonicalShapeInferenceRule canonical_rule;
-  luci::CircleShapeInferenceRule circle_rule;
-  loco::MultiDialectShapeInferenceRule rules;
-
-  rules.bind(loco::CanonicalDialect::get(), &canonical_rule)
-    .bind(luci::CircleDialect::get(), &circle_rule);
-
-  return loco::apply(&rules).to(g);
+  luci::CircleShapeInferencePass shape_pass;
+  return shape_pass.run(g);
```
seanshpark, line 39: As we don't use `loco::apply(&rules).to(g);` anymore we have to make same traverseral order.
You may have to copy `CircleShapeInferencePass` inside here to test exact thing as `CircleShapeInferencePass` does: because of depenceny.

This helper was to forget which Pass to test for each cases, one or several.
So, we can remove `shape_pass` and call each pass for each test.

Or, introduce another sub module that builds before pass and service so that both can link.
I can't think of what to name this... I remember I had this trouble when worked on this...

compiler/luci/service/CMakeLists.txt
llFreetimell: I thought that `add_subdirectory` just adds directory to build and CMake will calculates the dependencies after all subdirectories are added... am I wrong..?

seanshpark: Not sure... CI seems to work.. but I sometimes had trouble...

compiler/luci/service/src/CircleShapeInferenceRule.test.cpp
llFreetimell: If both of them is fine, I prefer to call each pass for each test...!
I will apply the former suggestion :)

compiler/luci/service/CMakeLists.txt
seanshpark: You should remove this for safety. It's dependency problem.

llFreetimell: This is needed this because `ShapeInferenceRule.test.cpp` and `TypeInferenceRule.test.cpp` are using `luci/Pass`.
If I remove this, build fail.

FYI, if I used `target_link_libraries(luci_service PUBLIC luci_pass)`, configure step would be failed because cycle is created in the dependency order.

seanshpark: > This is needed this because ShapeInferenceRule.test.cpp and TypeInferenceRule.test.cpp

We have to find another way

llFreetimell: Then the only solution is integrating these files to `luci/Pass/src/CircleShapeInferencePass.test.cpp` and `luci/Pass/src/CircleTypeInferencePass.test.cpp`.

seanshpark: There are tests for each, maybe not all, Pass. That is scope of Pass.
Current existing codes want to test for each Rules(shape algorithms), but somehow began testing with the Pass,
which is strictly speaking, in another module that uses this module.

I think it would be better to do like
- Do Rule testing in as scope of Rule, do not use Pass
- We will do all testing in Pass and there will be no Rule testing

I think you may first disable (not delete) all the tests for doing your refactoring and enable tests one by one, someday...
But this may have risks of "is refactoring safe?"... well.. I think there are several other thests that may cover this...

llFreetimell: I think you are right and I thought about what how should I handle this.
And I found something interesting.

- All of tests in `CircleShapeInferenceRule.test.cpp` and `CircleTypeInferenceRule.test.cpp` are really simple cases. 
- Recipes in `res/TensorFlowLiteRecipes/` include both simple cases and more complicated cases.
- `luci_unit_readtest` and `luci_unit_writetest` are using those recipes for validating inference results.

So is it okay to proceed following steps?

- If simple cases in `CircleShapeInferenceRule.test.cpp` and `CircleTypeInferenceRule.test.cpp` are not included in `res/TensorFlowLiteRecipes`, create new recipes.
- Add those test cases in luci/tests/test.lst`
- Remove `CircleShapeInferenceRule.test.cpp` and `CircleTypeInferenceRule.test.cpp`

seanshpark:
In the title, `Replace loco functions to luci functions in InferenceRule tests`, using `loco functions` and `luci functions` is strange.
What I see is you changed to use `luci::shape_known` instead of `loco::shape_known`.

llFreetimell:
@seanshpark
As this PR is changed a lot compared to initial PR, title seems strange for me too.
However, I am considering to close this PR and create another new PRs for https://github.com/Samsung/ONE/pull/5771#discussion_r562509788 if you are okay :)

seanshpark:
> However, I am considering to close this PR and create another new PRs for #5771 (comment) if you are okay :)

OK, fine with me :)

llFreetimell:
This PR is closed and new PRs will be created for https://github.com/Samsung/ONE/pull/5771#discussion_r562509788.
Close!

PR #5778, opened 2021-01-21
[luci] Quantize bias shared by multiple nodes.

jinevening:
This enables to quantize bias shared by multiple nodes.

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -612,79 +612,81 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<bool>
     if (is_quantized(node))
       return false;
 
-    // Check if this is bias
     auto iwo_list = get_input_weight_output_of_bias(node);
 
+    // TODO: Fix indentation
+    // clang-format off
     for (auto iwo : iwo_list)
     {
-      assert(iwo.size() == 3);
+    assert(iwo.size() == 3);
 
-      auto input = loco::must_cast<luci::CircleNode *>(iwo[0]);
-      auto weight = loco::must_cast<luci::CircleNode *>(iwo[1]);
-      auto output = loco::must_cast<luci::CircleNode *>(iwo[2]);
+    auto input = loco::must_cast<luci::CircleNode *>(iwo[0]);
+    auto weight = loco::must_cast<luci::CircleNode *>(iwo[1]);
+    auto output = loco::must_cast<luci::CircleNode *>(iwo[2]);
 
-      auto const_bias = loco::must_cast<luci::CircleConst *>(node);
-      assert(const_bias->dtype() == loco::DataType::FLOAT32);
+    auto const_bias = loco::must_cast<luci::CircleConst *>(node);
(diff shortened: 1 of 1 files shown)


seanshpark:
@jinevening , can you separate the pass that clones shared bias?
- quantize can have multiple phase.  first phase can do these kind of cloning to make things simple.
- second phase will do the quantization.

jinevening:
@seanshpark I've updated the PR not to clone a FLOAT32 bias. Instead, it directly creates a quantized bias (INT32 or INT64) from the FLOAT32 bias.

This would be the first work for https://github.com/Samsung/ONE/issues/5785.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -503,8 +553,8 @@ struct QuantizeActivation final : public luci::CircleNodeMutableVisitor<bool>
         continue;
 
       // Check if this is bias (bias is quantized later)
```
seanshpark, line 555: Maybe need to change this comment too

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -563,64 +613,77 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<bool>
       return false;
 
     // Check if this is bias
```
seanshpark, line 615: maybe update this comment too

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -563,64 +613,77 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<bool>
...
-    auto iw = get_input_weight_of_bias(node);
-    if (iw.first == nullptr || iw.second == nullptr)
-      return false;
-
-    auto input = loco::must_cast<luci::CircleNode *>(iw.first);
-    auto weight = loco::must_cast<luci::CircleNode *>(iw.second);
+    auto iwo_list = get_input_weight_output_of_bias(node);
 
-    if (granularity == QuantizationGranularity::ChannelWise)
+    for (auto iwo : iwo_list)
```
seanshpark, line 618: From this line to the end, it seems change is not that much but the whole code shows diff.
Can you please use `// clang-format off` to reduce diff and `TODO: fix indentation` or an appropriate word?

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: This comment does not have to be changed.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: I've changed this comment.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: I've added clang-format off and on. It makes the PR much clearer. PTAL :)

seanshpark (approved):
LGTM

PR #5779, opened 2021-01-21
[luci-interpreter] Conv2D fallback logic for std::bad_alloc

jinevening:
This adds a fallback logic to avoid std::bad_alloc.


jinevening:
This is required to quantize a large model such as FST.

seanshpark (approved):
LGTM

PR #5787, opened 2021-01-22
[luci] Introduce FuseBatchNormWithDwConvPass

seanshpark:
This will introduce FuseBatchNormWithDwConvPass that will fuse Mul+Add
into DepthwiseConv2D for sequence of DepthwiseConv2D+Mul+Add
for certain conditions.

### DIFF
--- a/compiler/luci/pass/src/FuseBatchNormWithDwConvPass.cpp
+++ b/compiler/luci/pass/src/FuseBatchNormWithDwConvPass.cpp
@@ -108,7 +108,7 @@ bool fused_batch_norm_with_dwconv(luci::CircleAdd *add)
   if (filter->rank() != 4)
     return false;
 
-  // check attributes of dwcon
+  // check attributes of dwconv
   if (dwconv->fusedActivationFunction() != luci::FusedActFunc::NONE)
     return false;
   if (dwconv->depthMultiplier() < 0) // can this happen?
--- a/compiler/luci/pass/src/FuseBatchNormWithDwConvPass.cpp
+++ b/compiler/luci/pass/src/FuseBatchNormWithDwConvPass.cpp
@@ -1,0 +1,226 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


seanshpark:
From draft #5761

compiler/luci/pass/src/FuseBatchNormWithDwConvPass.cpp
```diff
@@ -0,0 +1,226 @@
...
+    return false;
+
+  // check mul, add attributes
+  if (mul->dtype() != loco::DataType::FLOAT32)
+    return false;
+  if (mul->fusedActivationFunction() != luci::FusedActFunc::NONE)
+    return false;
+  if (add->dtype() != loco::DataType::FLOAT32)
+    return false;
+  if (add->fusedActivationFunction() != luci::FusedActFunc::NONE &&
```
llFreetimell, line 98: I think `RELU` or `TANH` also can be fused to DepthwiseConv2D.
Is there any restriction for this?

compiler/luci/pass/src/FuseBatchNormWithDwConvPass.cpp
```diff
@@ -0,0 +1,226 @@
...
+  // get weight of dwconv
+  auto filter = dynamic_cast<luci::CircleConst *>(dwconv->filter());
+  if (not filter)
+    return false;
+  if (filter->dtype() != loco::DataType::FLOAT32)
+    return false;
+  if (filter->rank() != 4)
+    return false;
+
+  // check attributes of dwcon
```
llFreetimell, line 111:
```suggestion
  // check attributes of dwconv
```

compiler/luci/pass/src/FuseBatchNormWithDwConvPass.cpp
seanshpark: Nope I don't think so. This was copied from existing FuseBatchNormWithConv and FuseBatchNormWithTConv.
We can enable for other activations in another task/PR, as we don't have high needs for other Activations right now.

llFreetimell: If then, adding a comment like `// TODO Support more activation` would be better.

FYI, in case of `FuseBatchNormWithConv`, `FusedActFunc` of `add` is not checked.
(I don't know why `FuseBatchNormWithTConv` introduced such checking)

mhs4670go:
> I don't know why `FuseBatchNormWithTConv` introduced such checking

FYI, I just considered the target model which has `add` operator whose activation function is `RELU6`. This can be expanded, which BTW, is what we can do later:)

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

PR #5798, opened 2021-01-25
[res] Add recipes of shape/type inference rule tests

llFreetimell:
Parent issue : #5809

Current `CircleShape(Type)InferenceRule.test.cpp` have following problems.

- Tests of `test.cpp` are using `Shape(Type)InferencePass` and the passes are using `CircleShape(Type)InferenceRule`
  - Sort of cyclic dependency
- Some of recipes in the tests of `test.cpp` are duplicated with the recipes in `res/`
- `luci_unit_readtest` and `luci_unit_writetest` have same purpose with the tests in `test.cpp`

Thus, integrating the `test.cpp` to `luci_unit_read(write)test` seems better.
For that, this commit will add missing recipes in the `test.cpp` files and make them usable at other tests like `luci_unit_read(write)test`.

### DIFF
--- a/res/TensorFlowLiteRecipes/Add_003/test.recipe
+++ b/res/TensorFlowLiteRecipes/Add_003/test.recipe
@@ -1,0 +1,27 @@
+operand {
+  name: "ifm1"
+  type: FLOAT32
+  shape { dim: 2 dim: 1 dim: 5 }
+}
+operand {
+  name: "ifm2"
+  type: FLOAT32
+  shape { dim: 3 dim: 5 }
+}
+operand {
+  name: "ofm"
+  type: FLOAT32
+  shape { dim: 2 dim: 3 dim: 5 }
+}
+operation {
+  type: "Add"
+  input: "ifm1"
+  input: "ifm2"
+  output: "ofm"
+  add_options {
+    activation: NONE
+  }
+}
+input: "ifm1"
+input: "ifm2"
+output: "ofm"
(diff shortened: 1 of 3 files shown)


seanshpark:
> CircleShapeInferenceRule.test.cpp has some simple test cases.
However, some of them are not in res/TensorFlowLiteRecipes even it can be used for various tests.
This commit will add those missing recipes in CircleShapeInferenceRule.test.cpp to res/TensorFlowLiteRecipes.

By this expression, one cannot understand what the exact change you are introducing.
Change itself seems good but I do not understand why these new recipes are added.
I can guess that these recipes will check shape inference problems that exist or may occur.
But as this change passed the test so we can assume there is no probem in current codes.
So what does this change needed? (I want you to know that this is a thought that it is NOT like "I do not want this change" but I am just curious and others may have same)

What I wish you to do is explain as detailed as much as you can in ONE LINE sentence.
You can use bullets for mutiple recipes, 3 or less for this case.
Think this as a training for you to express a change and make others understand without difficulties.

llFreetimell:
@seanshpark I updated the commit message, can you take another look? :)

seanshpark:
From your comment, I understand as this;
```
This will introduce  recipes that will replace unit tests in CircleShape(Type)InferenceRule.test.cpp
```
Is this correct?

seanshpark:
In Issue, we describe the problem and/or reason of should be done/fixed. And additional flow of thoughts tracing the problem...
In PR we describe the result of what will change from the PR.

I use this way to describe the comment...

llFreetimell:
This PR is not needed for now. Close this PR!

PR #5803, opened 2021-01-25
[luci/export] Add BidirectionalSequenceLSTM operation

dr-venkman:
This commit adds BidirectionalSequenceLSTM operation to luci export.

ONE-DCO-1.0-Signed-off-by: venkat.iyervenkat.iyer@samsung.com

### DIFF
--- a/compiler/luci/export/src/CircleOperationExporter.cpp
+++ b/compiler/luci/export/src/CircleOperationExporter.cpp
@@ -622,6 +622,7 @@ public:
   void visit(luci::CircleAveragePool2D *) final;
   void visit(luci::CircleBatchMatMul *) final;
   void visit(luci::CircleBatchToSpaceND *) final;
+  void visit(luci::CircleBidirectionalSequenceLSTM *) final;
   void visit(luci::CircleCast *) final;
   void visit(luci::CircleCeil *) final;
   void visit(luci::CircleConcatenation *) final;
@@ -735,6 +736,7 @@ public:
   void visit(luci::CircleOutputDummy *) final {}
   void visit(luci::CircleOutputExclude *) final {}
   // Virtual for multiple-outputs
+  void visit(luci::CircleBidirectionalSequenceLSTMOut *) final {}
   void visit(luci::CircleCustomOut *) final {}
   void visit(luci::CircleIfOut *) final {}
   void visit(luci::CircleNonMaxSuppressionV4Out *) final {}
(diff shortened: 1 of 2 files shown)


compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -815,6 +817,48 @@ void OperationExporter::visit(luci::CircleBatchMatMul *node)
                 CreateBatchMatMulOptions(_ctx.builder, node->adj_x(), node->adj_y()).Union());
 }
 
+void OperationExporter::visit(luci::CircleBidirectionalSequenceLSTM *node)
+{
+  auto bidi_lstm_outs = loco::succs(node);
+  assert(int32_t(bidi_lstm_outs.size()) == 2);
```
mhs4670go, line 823: I'm wondering this is always 2.
```cxx
bool visit(luci::CircleBidirectionalSequenceLSTM *node) final
  {
    if (node->merge_outputs())
    {
      store_outputs(node, 1);
    }
    else
    {
      store_outputs(node, 2);
    }
    return true;
  }
```
Because, as this codes, it could be 1.

dr-venkman: > I'm wondering this is always 2.
> 
> ```c++
> bool visit(luci::CircleBidirectionalSequenceLSTM *node) final
>   {
>     if (node->merge_outputs())
>     {
>       store_outputs(node, 1);
>     }
>     else
>     {
>       store_outputs(node, 2);
>     }
>     return true;
>   }
> ```
> 
> Because, as this codes, it could be 1.

You are right, for the currently available recipes, the assert would pass. In general, it should have a range check between 1 to 2. I will reflect this in the subsequent commit. Thanks.

mhs4670go: Got it:) then `format-checker` should be solved to merge this PR.

compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -815,6 +817,48 @@ void OperationExporter::visit(luci::CircleBatchMatMul *node)
                 CreateBatchMatMulOptions(_ctx.builder, node->adj_x(), node->adj_y()).Union());
 }
 
+void OperationExporter::visit(luci::CircleBidirectionalSequenceLSTM *node)
+{
+  auto bidi_lstm_outs = loco::succs(node);
+  assert((int32_t(bidi_lstm_outs.size()) > 0) && (int32_t(bidi_lstm_outs.size()) <= 2));
```
mhs4670go, line 823: This is kind of explicit:)
```suggestion
  assert(bidi_lstm_outs.size() == 0 || bidi_lstm_outs.size() == 2));
```
And, IMHO, when you cast sth, it would be better to use `static_cast` for safety. BTW, why did you cast to `int32_t`? There seems no need to cast I think.

dr-venkman: There is no need to cast, this was more of a copy paste artifact. I will reflect the changes. Thanks.

mhs4670go (approved):
LGTM

struss (approved):
LGTM

Issue #5783, opened 2021-01-22
Compiler FE: Fix fuse_instnorm


seanshpark:
`fuse_instnorm` option value testing fails with core dumped.
See compiler/luci-pass-value-test/test.lst comment.

Let's fix this problem.

seanshpark:
`Aborted (core dumped)` when
```
# Do inference
interpreter.invoke()
```
in `eval_result_verifier.py`

seanshpark:
There are RANK 5 operands...  
- From https://github.com/Samsung/ONE/issues/1804#issuecomment-642699002
- Test with TF 2.3.0

seanshpark:
done

PR #5805, opened 2021-01-25
[res] Revise Net_InstanceNorm_002

seanshpark:
This will revise Net_InstanceNorm_002 from using rank 5 to rank 4.
- rank 5 model fails to run in tflite interpreter
- remove Reshape node that is not needed anymore
- this is to match result from TF2.3.0


seanshpark:
For #5783

mhs4670go (approved):
LGTM

Issue #5777, opened 2021-01-21
[Compiler][luci] Implementation of more efficient luci/pass


llFreetimell:
## What?

Let's re-implement some of passes to be more fast!

## Why?

Some of passes are implemented like following logic, which always break if pass is applied once.
```cpp
bool RemoveRedundantTransposePass::run(loco::Graph *g)
{
  bool changed = false;
  for (auto node : loco::active_nodes(loco::output_nodes(g)))
  {
    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
    if (remove_consecutive_transpose_function(circle_node))
    {
      changed = true;
      break;
    }
  }
  return changed;
}
```
If there are N nodes to apply such kind of passes, we should execute above pass at least N times.
This makes compiler slow for big models.

## How?

Let's enable following implementations.
(Some passes may changed core logic to do so)

```cpp
bool RemoveRedundantTransposePass::run(loco::Graph *g)
{
  bool changed = false;
  for (auto node : loco::active_nodes(loco::output_nodes(g)))
  {
    if (auto transpose = dynamic_cast<luci::CircleTranspose *>(node))
    {
      if (remove_consecutive_transpose_function(transpose))
        changed = true;
    }
  }
  return changed;
}
```

## To-do

Re-implement following passes. If re-implementation is impossible, remove it from following list.

- [x] ConvertNCHWToNHWCPass #5822
- [x] PropagateQuantParamPass #5817
- [x] RemoveRedundantTransposePass #5815
- [x] ReplaceMulAddWithDepthwiseConvPass #5816

seanshpark:
```cxx
    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
    if (remove_consecutive_transpose_function(circle_node))
      changed = true;
```
would be better be
```cxx
    if (specific_node = dynamic_cast<luci::CircleSomeSpecificNode*>(node))
    {
      if (some_specif_function(specific_node ))
        changed = true;
    }
```
You shouldn't use `must_cast` here.

llFreetimell:
That can be handled together :)

llFreetimell:
Done!

PR #5815, opened 2021-01-25
[luci] Revise RemoveRedundantTransposePass

llFreetimell:
Parent Issue : #5777

Current `RemoveRedundantTransposePass` has following problems

- The pass is ended right after the pass is applied to only one node.
- Checking `CircleTranspose` is done twice.

As a result, if there are a lot of nodes to apply this pass, much time is needed.
To make this pass faster, this commit will revise following things.

- Do not break right after this pass is applied for a node. Apply this pass as much as possible at once.
- Filtering `CircleTranspose` first and then execute internal function.

### DIFF
--- a/compiler/luci/pass/src/RemoveRedundantTransposePass.cpp
+++ b/compiler/luci/pass/src/RemoveRedundantTransposePass.cpp
@@ -35,11 +35,8 @@ bool check_perm(const luci::CircleConst *first_perm, const luci::CircleConst *se
   return true;
 }
 
-bool remove_consecutive_transpose_function(luci::CircleNode *node)
+bool remove_consecutive_transpose_function(luci::CircleTranspose *target_node)
 {
-  auto target_node = dynamic_cast<luci::CircleTranspose *>(node);
-  if (target_node == nullptr)
-    return false;
   auto pred_node = dynamic_cast<luci::CircleTranspose *>(target_node->a());
   if (pred_node == nullptr)
     return false;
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/RemoveRedundantTransposePass.cpp
```diff
@@ -118,11 +115,10 @@ bool RemoveRedundantTransposePass::run(loco::Graph *g)
   bool changed = false;
   for (auto node : loco::active_nodes(loco::output_nodes(g)))
   {
-    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
-    if (remove_consecutive_transpose_function(circle_node))
+    if (auto transpose = dynamic_cast<luci::CircleTranspose *>(node))
     {
-      changed = true;
-      break;
```
seanshpark, line 125: @struss , did you have any problem to add a `break` here?

struss: Yes, but I think that `break` make this pass only update once for `RedundantTranspose` case. I think removing this nice.

seanshpark: > Yes, 

What problem did you have?

struss: > > Yes,
> 
> What problem did you have?

Sorry to miss reply. I think removing that break make no problem on it. I first thought only one change would okay, so I added. but think about more deeply, this change could be done multiple. so I think removing that `break` is nice change.

> @struss , did you have any problem to add a `break` here?

Sorry for confusing. to reply on yes. I have no problem on this. Sorry.

seanshpark (approved):
LGTM

struss (approved):
LGTM. Thanks!

PR #5816, opened 2021-01-25
[luci] Revise ReplaceMulAddWithDepthwiseConvPass

llFreetimell:
Parent Issue : #5777

Current `ReplaceMulAddWithDepthwiseConvPass` is ended right after the pass is applied to only one node.
As a result, if there are a lot of nodes to apply this pass, much time is needed.

### DIFF
--- a/compiler/luci/pass/src/ReplaceMulAddWithDepthwiseConvPass.cpp
+++ b/compiler/luci/pass/src/ReplaceMulAddWithDepthwiseConvPass.cpp
@@ -206,14 +206,10 @@ bool ReplaceMulAddWithDepthwiseConvPass::run(loco::Graph *g)
   bool changed = false;
   for (auto node : loco::active_nodes(loco::output_nodes(g)))
   {
-    auto add = dynamic_cast<luci::CircleAdd *>(node);
-    if (not add)
-      continue;
-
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/ReplaceMulAddWithDepthwiseConvPass.cpp
```diff
@@ -206,14 +206,10 @@ bool ReplaceMulAddWithDepthwiseConvPass::run(loco::Graph *g)
...
   {
-    auto add = dynamic_cast<luci::CircleAdd *>(node);
-    if (not add)
-      continue;
-
-    if (replace_mul_add_with_dwconv(add))
+    if (auto add = dynamic_cast<luci::CircleAdd *>(node))
     {
-      changed = true;
-      break;
```
seanshpark, line 216: @jinevening , did you have any problem with this Pass to use `break` here?

jinevening: No problem

jinevening (approved):
LGTM Thank you

seanshpark (approved):
LGTM

PR #5817, opened 2021-01-25
[luci] Revise PropagateQuantParamPass

llFreetimell:
Parent issue : #5777 

Current `PropagateQuantParamPass` is ended right after the pass is applied to only one node.
If there are a lot of nodes to apply this pass, much time is needed.

### DIFF
--- a/compiler/luci/pass/src/PropagateQuantParamPass.cpp
+++ b/compiler/luci/pass/src/PropagateQuantParamPass.cpp
@@ -91,9 +91,8 @@ bool PropagateQuantParamPass::run(loco::Graph *g)
     INFO(l) << "PropagateQuantParamPass visit node: " << circle_node->name() << std::endl;
 
     PropagateQuantParam pqp;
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/PropagateQuantParamPass.cpp
```diff
@@ -91,9 +91,8 @@ bool PropagateQuantParamPass::run(loco::Graph *g)
     INFO(l) << "PropagateQuantParamPass visit node: " << circle_node->name() << std::endl;
 
     PropagateQuantParam pqp;
-    changed = circle_node->accept(&pqp);
-    if (changed)
-      break;
```
seanshpark, line 96: @jinevening , did you have any problem with this Pass to use break here?

jinevening: No problem.

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

Issue #5786, opened 2021-01-22
Compiler FE: Revise resolve_customop_add


seanshpark:
We need to revise `resolve_customop_add` option not to work with `int64` inputs.

Refer: https://github.com/Samsung/ONE/issues/5781#issuecomment-765049842

mhs4670go:
@seanshpark I am not sure but `Const(Int64) - CustomOp(AddV2) - Cast(to int32)` will convert to `Const(Int64) - Add - Cast(to int32)` with `resolve_customop_add`. It is not supposed to do so?

seanshpark:
As-is: doesn not check int64 or int32.
-> Problem: converting to int64 may cause loss
-> To-be: don't convert from CustomOp(AddV2) to CircleAdd

New problem ???: There will be `CustomOp(AddV2)` in circle model.

mhs4670go:
For more detail, since `CircleAdd` input should be `int32`, `int64` shouldn't be its input. If you are not doing this job yet, I'll revise this code.

seanshpark:
> If you are not doing this job yet, I'll revise this code.

That will be nice :)

seanshpark:
From #5819,

@mhs4670go , can you explain a bit why `BroadcastTo` is introduced?

mhs4670go:
@seanshpark https://github.com/Samsung/ONE/blob/721dfa4798086868289d6ddad36992a5098a91b9/compiler/luci/pass/src/ResolveCustomOpAddPass.cpp#L45-L59

Originally `ResolveCustomOpAddPass` was introduced for above pattern. And to make up for insufficient tests, I'll add related network recipe. And to generate test artifacts, `BroadcastTo` should be introduced.

PR #5820, opened 2021-01-25
[common-artifcats] add network having BroadcastTo op to exclude list

mhs4670go:
add networks having BroadcastTo op to exclude list.

FYI, `Net_BroadcastTo_AddV2_000` can be run by `luci-interpreter` because of `resolve_customop_add` pass. But `Net_BroadcastTo_AddV2_001` network can't be optimized by `resolve_customop_add` because `addV2` in the network has input whose dtype is `int64`.

Related : #5786
Draft: #5814


seanshpark (approved):
LGTM

Issue #5794, opened 2021-01-22
[luci] How to describe circle operator's profile info


cgbahk:
## What!

For profiling purpose, each circle node is expected to have _some_ profile info. This is to discuss how it would be described.

### Option 1) Make dedicated profile index for each circle operator
#5753

### Option 2) Use existing operator index

Circle (and tflite) has list of operators:
https://github.com/Samsung/ONE/blob/abb70047faf755bfbd718fbd75afc982953124ea/nnpackage/schema/circle_schema.fbs#L1084-L1085

For example, it will look like

<details>

![image](https://user-images.githubusercontent.com/34672141/105469125-20093680-5cdb-11eb-9b73-be68fbfa3ede.png)
```
$ tfldump inception_v3.tflite
...
Operators: O(subgraph index : operator index) OpCodeName 
    Option(values) ... <-- depending on OpCode
    I T(tensor index) OperandName <-- as input
    O T(tensor index) OperandName <-- as output
O(0:0) CONV_2D 
    Padding(1) Stride.W(2) Stride.H(2) Dilation.W(1) Dilation.H(1) Activation(RELU)
    I T(0:317) input
    I T(0:0) InceptionV3/Conv2d_1a_3x3/weights
    I T(0:5) InceptionV3/InceptionV3/Conv2d_1a_3x3/Conv2D_bias
    O T(0:6) InceptionV3/InceptionV3/Conv2d_1a_3x3/Relu
O(0:1) CONV_2D 
    Padding(1) Stride.W(1) Stride.H(1) Dilation.W(1) Dilation.H(1) Activation(RELU)
    I T(0:6) InceptionV3/InceptionV3/Conv2d_1a_3x3/Relu
    I T(0:1) InceptionV3/Conv2d_2a_3x3/weights
    I T(0:7) InceptionV3/InceptionV3/Conv2d_2a_3x3/Conv2D_bias
    O T(0:8) InceptionV3/InceptionV3/Conv2d_2a_3x3/Relu
O(0:2) CONV_2D 
    Padding(0) Stride.W(1) Stride.H(1) Dilation.W(1) Dilation.H(1) Activation(RELU)
    I T(0:8) InceptionV3/InceptionV3/Conv2d_2a_3x3/Relu
    I T(0:2) InceptionV3/Conv2d_2b_3x3/weights
    I T(0:9) InceptionV3/InceptionV3/Conv2d_2b_3x3/Conv2D_bias
[...]

PR #5821, opened 2021-01-26
[Draft][NO MERGE] Recording Profiling data by annotated item

llFreetimell:
Related Issue : #5794

This draft is for trying to record profiling data by `loco::NodeAnnotation`


llFreetimell:
This is not sophisticate, just for sharing idea :)

compiler/luci/pass/src/FuseBatchNormWithConvPass.cpp
```diff
@@ -195,6 +197,10 @@ bool fused_batch_norm_with_conv(luci::CircleAdd *add)
   fused_conv->dilation()->h(conv->dilation()->h());
   fused_conv->dilation()->w(conv->dilation()->w());
 
+  auto pdata = std::make_unique<luci::ProfilingData>();
+  pdata->origins(add->annot<luci::ProfilingData>()->origins());
```
mhs4670go, line 201: need sum

llFreetimell, line 201: logic is wrong, to be fixed.
just for example...!

compiler/luci/pass/src/FuseBatchNormWithConvPass.cpp
```diff
@@ -195,6 +197,12 @@ bool fused_batch_norm_with_conv(luci::CircleAdd *add)
   fused_conv->dilation()->h(conv->dilation()->h());
   fused_conv->dilation()->w(conv->dilation()->w());
 
+  auto pdata = std::make_unique<luci::ProfilingData>();
+  pdata->add_origins(conv->annot<luci::ProfilingData>()->origins());
+  pdata->add_origins(mul->annot<luci::ProfilingData>()->origins());
+  pdata->add_origins(add->annot<luci::ProfilingData>()->origins());
+  fused_conv->annot(std::move(pdata));
```
llFreetimell, line 204: **NOTE** This file is just an example for ONE. Not related with backend.

cgbahk, line 204: Looks like this PR don't have part to assign `pdata->node_id`. I guess that would be done in export stage?

compiler/luci/lang/include/luci/IR/ProfilingData.h
```diff
@@ -0,0 +1,66 @@
...
+
+  ProfilingData(uint32_t id)
+  {
+    _node_id = static_cast<int32_t>(id);
+    _origins.insert(static_cast<int32_t>(id));
+  }
+
+public:
+  int32_t node_id(void) const { return _node_id; }
+  void node_id(int32_t id) { _node_id = id; }
```
cgbahk, line 49: Actually for me, this is only required :) other can be added later

cgbahk, line 49: Is `node id` same concept with 'operator index'?

compiler/luci/pass/src/FuseBatchNormWithConvPass.cpp
llFreetimell: Yes, operator index(node_id) of newly created nodes can be got only at export step :D
The only thing we can do in here is recording origins :)

compiler/luci/lang/include/luci/IR/ProfilingData.h
llFreetimell: Exactly same for now :)
Sorry for poor naming :(
Do you want to rename like `op_index()`?

compiler/luci/lang/include/luci/IR/ProfilingData.h
```diff
@@ -0,0 +1,66 @@
...
+  {
+    _node_id = id;
+    _origins.insert(id);
+  }
+
+  ProfilingData(uint32_t id)
+  {
+    _node_id = static_cast<int32_t>(id);
+    _origins.insert(static_cast<int32_t>(id));
+  }
```
cgbahk, line 45: (This is not related to backend perspective but...)

These ctors looks strange to me. I guess there is no relation between node id and origins

compiler/luci/lang/include/luci/IR/ProfilingData.h
cgbahk: Not necessarily, it maybe good to have independent name considering future modification.

But I hope comments to be added that this is operator index as of now on main branch :)

cgbahk: Also comment on meaning of negative node id looks required if `int32_t` used.

Alternative design, we may use `uint32_t` for this. I guess this node would have valid unsigned value in all cases?

llFreetimell: I used `int32_t` to represent undefined status because operator index is started from 0.
Not special reason :)

I think we can use `uint32_t` and delete default constructor since node id will not be changed once it is allocated :)

compiler/luci/lang/include/luci/IR/ProfilingData.h
llFreetimell: Sorry for my poor design skill :(
At first, it would be better to include both of them but it would be better to divide `CircleNodeID` and `CircleNodeOrigin`

compiler/luci/pass/src/FuseBatchNormWithConvPass.cpp
```diff
@@ -195,6 +197,9 @@ bool fused_batch_norm_with_conv(luci::CircleAdd *add)
   fused_conv->dilation()->h(conv->dilation()->h());
   fused_conv->dilation()->w(conv->dilation()->w());
 
+  auto origins = std::vector<const luci::CircleNode *>{conv, add, mul};
+  fused_conv->annot(std::make_unique<luci::CircleNodeOrigin>(origins));
+
```
llFreetimell, line 202: **NOTE** This is just an example code for ONE. Not related with backend

compiler/luci/import/src/GraphBuilder.cpp
```diff
@@ -53,6 +55,17 @@ void GraphBuilder::build(const circle::OperatorT &op, GraphBuilderContext *conte
 
   CircleNode *node = build_node(op, input_nodes, context->graph());
 
+  for (uint32_t i = 0; i < operators.size(); ++i)
+  {
+    const circle::OperatorT &ref_op = *operators[i];
+    if (&ref_op == &op)
+    {
+      node->annot(std::make_unique<luci::CircleNodeID>(i));
```
cgbahk, line 63: I guess node ID will be overwritten on export?

llFreetimell: In ONE, yes it will be.
Backend may be different with ONE :)

compiler/luci/lang/include/luci/IR/ProfilingData.h
```diff
@@ -0,0 +1,76 @@
...
+
+public:
+  const std::set<uint32_t> &origins(void) const { return _origins; }
+  void origins(const std::set<uint32_t> &origins) { _origins = origins; } // Do we need this?
+
+private:
+  std::set<uint32_t> _origins;
+};
+
+} // namespace luci
```
cgbahk, line 76: This design is okay to me.

---
 I have one minor concern in long term:
For us, this is familiar to access info from NodeAnnodation. It is actually public and possible.
For external users, accessing info from NodeAnnotation may not be familiar. They may not even know that this is info exists. Say, they might just check `CircleNodeDecl.h` and found nothing about profile API.

This is not be problem in short term. But IMO good to prepare solution to that at least (make a guide? introduce API in circle node when the design becomes stable?)

seanshpark:
For `compiler/luci/lang/include/luci/IR/ProfilingData.h`,
I suggest to put this in `compiler/luci/profile`

compiler/luci/import/src/GraphBuilder.cpp
mhs4670go: @cgbahk To be specific, once operators are imported(get annotation), it won't be overwritten. We will just create new operator list when exports to be optimized(or sth) model.

compiler/luci/lang/include/luci/IR/ProfilingData.h
```diff
@@ -0,0 +1,76 @@
...
+  CircleNodeID() = delete;
+
+  CircleNodeID(uint32_t node_id) : _node_id{node_id}
+  {
+    // Do nothing
+  }
+
+public:
+  uint32_t node_id(void) const { return _node_id; }
+  void node_id(uint32_t id) { _node_id = id; } // Do we need this?
```
mhs4670go, line 42: @llFreetimell I think we can remove this setter. Just comment for letting them know that node id shouldn't be changed.

compiler/luci/lang/include/luci/IR/ProfilingData.h
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * @brief Set annotation for circle node origin
+ */
+class CircleNodeOrigin final : public loco::NodeAnnotation
+{
+public:
+  CircleNodeOrigin() = delete;
+
+  CircleNodeOrigin(uint32_t origin) { _origins.insert(origin); }
```
mhs4670go, line 56:
```suggestion
  CircleNodeOrigin(uint32_t origin) : _origins{origin}
```

compiler/luci/lang/include/luci/IR/ProfilingData.h
```diff
@@ -0,0 +1,76 @@
...
+
+  CircleNodeOrigin(std::vector<const luci::CircleNode *> origin_nodes)
+  {
+    for (auto node : origin_nodes)
+    {
+      assert(node->annot<CircleNodeOrigin>() != nullptr);
+      for (auto origin : node->annot<CircleNodeOrigin>()->origins())
+        _origins.insert(origin);
```
mhs4670go, line 64:
```suggestion
      to_merge = node->annot<CircleNodeOrigin>()->origins();
      _origins.insert(to_merge.begin(), to_merge.end();
```

compiler/luci/lang/include/luci/IR/ProfilingData.h
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+public:
+  const std::set<uint32_t> &origins(void) const { return _origins; }
+  void origins(const std::set<uint32_t> &origins) { _origins = origins; } // Do we need this?
```
mhs4670go, line 70: ditto

compiler/luci/lang/include/luci/IR/ProfilingData.h
llFreetimell: For ONE, this setter may not be needed.
But I am not sure that backend does not need this :(

cgbahk: I think backend will not need this as of now

mhs4670go: Maybe just new annotation would be able to applied. I'll follow your choice.

llFreetimell: If this setter is not used at last, I will remove then :)

llFreetimell:
Related PRs are created for now.
If needed, another draft will be created! Close!

PR #5822, opened 2021-01-26
[luci] Revise ConvertNCHWToNHWCPass

llFreetimell:
Parent Issue : #5777

Current `ConvertNCHWToNHWCPass` is ended right after the pass is applied to only one node.
If there are a lot of nodes to apply this pass, much time is needed.


jinevening (approved):
LGTM Thanks.

PR #5824, opened 2021-01-26
[luci-pass-value-test] Revise comment

seanshpark:
This will revise comment for Net_Preactivation_BN_000 model using fuse_preactivation_batchnorm
as segmentation fault is fixed but there exist value difference for this Pass.


seanshpark:
For #5824

jinevening (approved):
LGTM Thank you

PR #5827, opened 2021-01-26
[luci] Enable convert to NHWC for Neg

seanshpark:
This will enable conversion to NHWC for CircleNeg.


jinevening (approved):
LGTM

Issue #5830, opened 2021-01-26
Compiler FE: ONNX with 2 squeeze


seanshpark:
Let's work on this model with two `Squeeze` with dynamic shape

```py
import torch
import onnx


class TestModel(torch.nn.Module):
  def __init__(self):
    super(TestModel, self).__init__()

  def forward(self, input):
    output = input
    output = torch.squeeze(input, dim=3)
    output = torch.squeeze(output, dim=2)
    return output


model = TestModel()
torch.save(model, "model.pth")

input = torch.randn(1, 16, 1, 1)
dynamic_axes = {'input':[0]}
torch.onnx.export(model, input, "squeeze-2x.onnx", verbose=False,
                  input_names=["input"],
                  output_names=["output"],
                  dynamic_axes=dynamic_axes)
```

seanshpark:
ONNX model
![image](https://user-images.githubusercontent.com/4616940/105816559-ce70ec80-5ff7-11eb-9e53-e0f60a964cf5.png)

seanshpark:
portion of circledump of `squeeze-2x.opt.circle` file
```
Operands: T(subgraph index : tensor index) TYPE (shape) (shape_signature) B(buffer index) OperandName
T(0:0) FLOAT32 (1, 16, 1, 1) (-1, 16, 1, 1) B(0) input
T(0:1) FLOAT32 (1, 16, 1) (-1, 16, 1) B(0) onnx_tf_prefix_Squeeze_0;PartitionedCall/onnx_tf_prefix_Squeeze_0
T(0:2) FLOAT32 (1, 16) (-1, 16) B(0) Identity

Operators: O(subgraph index : operator index) OpCodeName
    Option(values) ... <-- depending on OpCode
    I T(tensor index) OperandName <-- as input
    O T(tensor index) OperandName <-- as output
O(0:0) SQUEEZE
    SqueezeDims(3)
    I T(0:0) input
    O T(0:1) onnx_tf_prefix_Squeeze_0;PartitionedCall/onnx_tf_prefix_Squeeze_0
O(0:1) SQUEEZE
    SqueezeDims(2)
    I T(0:1) onnx_tf_prefix_Squeeze_0;PartitionedCall/onnx_tf_prefix_Squeeze_0
    O T(0:2) Identity

Inputs/Outputs: I(input)/O(output) T(tensor index) OperandName
I T(0:0) input
O T(0:2) Identity
```

seanshpark:
[...]

PR #5833, opened 2021-01-26
DRAFT: Optimize Squeeze sequence

seanshpark:
on-going draft to optimize sequence of Squeeze operators.


seanshpark:
For #5830

seanshpark:
all landed

Issue #5440, opened 2020-12-21
[Compiler] Native code generation


binarman:
Lets investigate code generation techniques in neural network world.

Some NN has subgraphs with lots of relatively small operators, that introduce a lot of overhead on runtime.
There are also cases when some operators can be simplified and fused that can lead to performance improvement.

Some work in this direction is already done in other projects. For example:
- https://github.com/Tiramisu-Compiler/tiramisu
- https://tvm.apache.org/
- http://tensor-compiler.org/docs/machine_learning/index.html
- and many others

Probably we can make use of these technologies to improve inference performance

binarman:
To get some understanding of optimization possibilities I've experimented a bit with small subgraph from [lpcnet](https://github.com/mozilla/LPCNet)
Model has GRU layer that is transformed in set of small operators.

Subgraph from original network I targeted:
![subgraph](https://user-images.githubusercontent.com/3631351/102734386-c1a20e80-4350-11eb-9791-9350c5492c7a.png)

I rewrote this subgraph using primitive tensorflow functions and evaluated it's performance with ONERT.
Then I reimplemented this graph in c++ in three variants:
1. naive implementation (just for reference)
2. naive implementation, but with kernels from ONERT (I assume no loop fusion was applied by compiler, though I did not inspect it carefully)
3. optimized implementation: manually fused and unrolled loops.

Performance results are following (100 000 runs, warm up 100):
1. ONERT: ~3.05 microseconds  (I used modified nnpackage_run to get rid of infrastructure overhead and get better precision)
1. C++ with ONERT kernels: ~2.15 microseconds
1. C++ optimized: ~1.7 microseconds

Optimized model has three clusters of fused layers. It is possible to make only two clusters, but there are not enough registers to hold all needed data, so three-cluster implementation performed better:
![schema_fuse2](https://user-images.githubusercontent.com/3631351/102736304-a5ed3700-4355-11eb-833e-68b2edbd2402.png)

I think these results look promising, though this is just a small part of the network. probably bigger layers, like FC can not offer such improvement.

related files:
[gru_experiment.zip](https://github.com/Samsung/ONE/files/5722115/gru_experiment.zip)

binarman:
I tried to generate code using Tiramisu infrastructure: 
[tiramisu_experiemnts.zip](https://github.com/Samsung/ONE/files/5741747/tiramisu_experiemnts.zip)
due to it's technical limitations I broke graph into different loops: 
[...]

PR #5836, opened 2021-01-26
[DRAFT] Codegen draft

binarman:
This PR reflects current status of POC for issue #5440


compiler/luci-codegen/include/LuciCodegen.h
```diff
@@ -0,0 +1,66 @@
...
+{
+  /***
+   * max size of constant buffer to inline in code in bytes
+   */
+  int max_inline_buffer_threshold = 1024;
+};
+
+class SubgraphContext;
+
+class LuciCodegen
```
seanshpark, line 40: How about `Codegen` ? as it's inside `luci_codegen` namespace.

binarman: Sounds reasonable, I'll rename it

compiler/luci-codegen/src/CodegenKernelBuilder.h
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef NNCC_CODEGEN_KERNEL_BUILDER_H
+#define NNCC_CODEGEN_KERNEL_BUILDER_H
+
+#include "SubgraphContext.h"
+
+namespace luci_codegen
+{
+
+class CodegenKernelBuilder
```
seanshpark, line 25: How about `KernelBuilder` ? as `codegen` is duplicate in namespace and class name...

compiler/luci-interpreter/src/loader/KernelBuilder.cpp
```diff
@@ -74,8 +75,13 @@
 #include "kernels/Transpose.h"
 #include "kernels/TransposeConv.h"
 
+#include "Halide.h"
```
seanshpark, line 78: luci-interpreter is used for quantization tool and this is included in the release package.
if we include also halide, we need to include halide package to our release package,
which I don't know how this is simple or complicated, and there may be concern with version control.
could you please check how this is turn out?

or if you could make this optional so that our tools can be released without halide then that is also OK.

binarman: Yes, I thought about this, and I think I can remove dependency, or at least leave minimal set of struct declarations (there are not many of them, afaik).
I included whole halide for now just because this was faster and easier and I want to get working prototype as soon as possible. =)

If I fail to remove dependency I have two things on my mind:
1) Easy and fast way: make compiled node support optional.
2) Harder and slower way: try to remove/wrap halide related stuff on compiler-side.

P.s. Actually I would like to proceed with 2 case, because this eliminates redundant dependencies entirely.

binarman:
@seanshpark
FYI
I fixed issue you mentioned about luci-interpter dependency from Halide.
I achieved this by defining intermediate layer between Halide generated code and runtime environment.

There are few words about this in README.md.
Next I am going to add rationale to clarify why this interface looks like this.

binarman:
@seanshpark 
I think current code is working and I don't think I will change it significantly.
What do you think if I start cutting it into PR's? (I want to start from documentation)

What I should add for easier review? Maybe tests/documentation/examples?

seanshpark:
@binarman , sorry to miss your last comment -_-;

> What I should add for easier review? Maybe tests/documentation/examples?

this sounds nice :) 

I'll look all the changes roughly and try to add comment what I think (though it may sound strange to you...)

build/run_runner.sh
```diff
@@ -0,0 +1,16 @@
+HALIDE=~/ops/codegen_experiments/halide/Halide/
```
seanshpark, line 1: maybe add `bash` shebang?

seanshpark:
@binarman , thanks for considering moving to root `codegen` :)
although I suggested this, there are build and CI related things to solve and this may not be a simple task to do in short time. 
what I worry is about build burden related with halide and llvm.
if you could make some options: build if exist and build if flag is turned on, that is default is off for the moment, it would be better to put back in to `compiler` folder so that you can consentrate on actual works.

codegen/include/CodegenOptions.h
```diff
@@ -0,0 +1,90 @@
...
+
+#ifndef NNCC_CODEGENOPTIONS_H
+#define NNCC_CODEGENOPTIONS_H
+
+namespace luci_codegen
+{
+
+enum class ArchType
+{
+  Native,
```
seanshpark, line 25: Is `Native` same as build machone architecutre? This seems implecit type and what is the needs or benifits of providing `Native`? As I understand, we're targeting ARM32 or AARCH64...

codegen/include/CodegenOptions.h
```diff
@@ -0,0 +1,90 @@
...
+  int max_inline_buffer_threshold = 1024;
+
+public: // backend options
+
+  /**
+   * If true generate argument correctness checks
+   */
+  bool debug = true;
+
+  Architecture arch;
```
seanshpark, line 75: maybe set default value?

codegen/src/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+target_link_libraries(luci_codegen PRIVATE ${Boost_LIBRARIES})
+
+add_custom_command(OUTPUT ${WRAPPER_LIB}
+        COMMAND python3 ${LUCI_CODEGEN_SUPPORT_DIR}/embed.py ${LUCI_CODEGEN_SUPPORT_DIR}/WrapperLib.h ${CMAKE_CURRENT_BINARY_DIR}/GeneratedWrapperLib.h wrapper_lib_source
+        DEPENDS ${WRAPPER_SOURCES}
+        VERBATIM)
+
+install(TARGETS luci_codegen DESTINATION lib)
+
+# native_compiler EXECUTABLE
```
seanshpark, line 39: `native_compiler` is like the `Native` ArchType? it's a bit ambigious for me... is this x86-64? (as we only work on x86-64 for compiler)

codegen/src/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+# tests
+
+if(ENABLE_TEST)
+    nnas_find_package(GTest REQUIRED)
+    set(TEST_SOURCES KernelBuilder.test.cpp Codegen.test.cpp)
+    GTest_AddTest(luci_codegen_test ${TEST_SOURCES})
+    target_link_libraries(luci_codegen_test luci_codegen luci_import luci_export Halide::Halide)
+endif(ENABLE_TEST)
+
+add_executable(simple_runner simple_runner.cpp)
```
seanshpark, line 61: is `simple_runner` also for tests?

codegen/src/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+target_include_directories(native_compiler PRIVATE ${Boost_INCLUDE_DIRS})
+
+target_link_libraries(native_compiler PRIVATE luci_codegen)
+target_link_libraries(native_compiler PRIVATE luci_import)
+target_link_libraries(native_compiler PRIVATE luci_export)
+target_link_libraries(native_compiler PRIVATE ${Boost_LIBRARIES})
+
+# tests
+
+if(ENABLE_TEST)
```
seanshpark, line 54: we usually used like `if(NOT ENABLE_REST) return()`

build/run_runner.sh
binarman: This script is for fast local debug, I probably should remove it from this branch.

P.s. I even did not commit prerequisites for this script, like model. So essentially it is useless now

binarman:
> what I worry is about build burden related with halide and llvm.

#6191 contains solution (at least temporary) for Halide build.
It downloads compiled Halide from official repository: https://github.com/halide/Halide/releases.
So "build" time mostly depends on network speed: need to download ~100 MB archive.

seanshpark:
> So "build" time mostly depends on network speed: need to download ~100 MB archive.

@hseok-oh , can we use docker image WITH this binary package?

codegen/include/CodegenOptions.h
binarman: > Is Native same as build machine architecutre?

Yes. I want to have this option as a default to simplify debugging/prototyping.

Halide looks stable, so I assume that most of bugs could be found using any target architecture. And debugging on the same machine is easier.

codegen/src/CMakeLists.txt
binarman: "native" refers to machine code.
I want to highlight that this compiler should produce binary code, not byte code or source.

P.s. This is not entirely true (:disappointed: ), because utility generates binary kernels and some **source** code, that wraps Halide structures and provides backend agnostic interfaces.

codegen/src/CMakeLists.txt
binarman: Yes, I am slowly cleaning up this code. I deleted it in #6191

codegen/src/CMakeLists.txt
seanshpark: I see... then as I understand this is not aligned with `Native` ArchType ....
using `native` word here maybe confusing but I can't think of good name right now...

This maybe another scope but is the generated binary wrapped as Custom operator through `CircleCustom` IR?

hseok-oh:
>So "build" time mostly depends on network speed: need to download ~100 MB archive.

Network speed may not be significant issue because `/external` directory uses volume docker as download cache.

codegen/src/CMakeLists.txt
binarman: > This maybe another scope but is the generated binary wrapped as Custom operator through CircleCustom IR?

Yes, generated binary is an implementation of CircleCustom in IR. I'll try to describe it in more detail (I should work on wording in documentation more and fill "Ratinale" section already :smiley: )

CircleCustom node is a "declaration" of operation that should be executed. It stores name of generated kernel.

Along with circle file runtime (ONERT, luci-interpreter, etc.) should be provided with library that contains implementation of generated kernel(s).

Each kernel is accessed using three functions:
- Constructor (of type `ConstructorCompiledFunc`)
- Destructor (of type `DestructorCompiledFunc`)
- Implementation function (of type `CompiledFuncWrapper`)

Constructor and Destructor names are "create_<kernel_name>" and "free_<kernel_name>" for every generated kernel.
Implementation function name is not important.

Wrapper library I mentioned implements these three functions.

More detail

**Constructor**

This function takes input tensor sizes and creates structure of `ConfiguredCompiledFunc` type.
`ConfiguredCompiledFunc` structure contains pointer to implementation function and some context.
You can think of this structure as std::function object, just implemented in C.

This function probably should be called once during preparation phase of runtime. (or after arguments resize)

**Destructor**

This function takes `ConfiguredCompiledFunc` object and free context if any.

**Implementation function**

This function implements kernel. To invoke it you need to pass context and pointers to memory with arguments.

Like this:
```
  ConfiguredCompiledFunc kernel = create_generates_operator(ranks, dims);

  kernel.wrapper(kernel.configuration, pointers_to_buffers);

  free_generates_operator(kernel);
```

P.s.
I decided to introduce contructor/destructor concept, because it opens a lot of interesting opportunities:
1. Some computations could be done in constructor. For example I allocate memory for Halide shape structures and initialize some other fields.
2. Runtime can JIT compile kernel implementation itself! in that case there are no implementation function provided at all.
  This could be useful for autotune approach implemented in TVM,

P.p.s.
I just remembered that we already worked on Custom operations in onert. AFAIK we didn't use it so I forgot about this. Need to check if it is still working.

codegen/luci-codegen/src/Driver.cpp
```diff
@@ -0,0 +1,86 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Codegen.h"
+#include "luci/Importer.h"
+#include "luci/CircleExporter.h"
+#include "luci/CircleFileExpContract.h"
+
+#include <boost/filesystem.hpp>
```
binarman, line 22: @seanshpark 
I noticed that compiler components try to minimize dependencies from `Boost` (and other third party libraries).
If I understand it correctly this is for portability reasons: we do not need to provide libraries with compiled toolchain.

As you can see this code uses Boost filesystem module, I can remove dynamic dependency by building compiler using static linking. Is it ok solution?

Otherwise, do we have some filesystem related submodule in compiler that I can use? (like `arser` instead of `boost::program_options`)

seanshpark: >  I can remove dynamic dependency by building compiler using static linking. Is it ok solution?

This is nice not to release with external libs :) But this may grow bin size but I don't think not that much in this case...

> do we have some filesystem related submodule in compiler that I can use?

You used 
```cxx
  boost::filesystem::path output_package_dir(output_package_path);
  if (!boost::filesystem::exists(output_package_dir))
  {
    boost::filesystem::create_directory(output_package_dir);
  }
  boost::filesystem::path output_model_path = output_package_dir / "model.circle";
```
which can be implemented in standard APIs... but `output_package_dir / "model.circle";` need an `if` line...

So, my thoughts are something like this;
1) If you can and willing to maintain `codegen` root folder, I think it is up to you
2) `maintain` will require responsibility of release tasks also. This may have narrow time slot.
3) as we are not developing cross-platform product, that is, we (currently) release only for Ubuntu, we can provide a small functions that acts like boost APIs, in this case (as I know, boost also covers platform dependent things)
4) when these libs are used several places, we can then make separate module in `codegen/` or in `compiler/`
5) when it's time we can use c++17, then these can be refactored using std libs :)

seanshpark:

PR #5838, opened 2021-01-27
[res] Introduce Net_Squeeze_Squeeze_000 tfl recipe

seanshpark:
This will introduce Net_Squeeze_Squeeze_000 tflite recipe that has two
continued Squeeze operators.

### DIFF
--- a/res/TensorFlowLiteRecipes/Net_Squeeze_Squeeze_000/test.rule
+++ b/res/TensorFlowLiteRecipes/Net_Squeeze_Squeeze_000/test.rule
@@ -1,0 +1,6 @@
+# To check if Squeeze is substituted to Reshape op
+
+RULE    "VERIFY_FILE_FORMAT"      $(verify_file_format) '=' 1
+
+RULE    "SQUEEZE_COUNT"           $(op_count SQUEEZE) '=' 0
+RULE    "RESHAPE_COUNT"           $(op_count RESHAPE) '=' 2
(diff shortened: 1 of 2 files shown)


seanshpark:
For #5830, from draft #5833

res/TensorFlowLiteRecipes/Net_Squeeze_Squeeze_000/test.rule
```diff
@@ -0,0 +1,6 @@
+# To check if Squeeze is substituted to Reshape op
+
+RULE    "VERIFY_FILE_FORMAT"      $(verify_file_format) '=' 1
+
+RULE    "SQUEEZE_COUNT"           $(op_count SQUEEZE) '=' 0
+RULE    "RESHAPE_COUNT"           $(op_count RESHAPE) '=' 2
```
llFreetimell, line 6: For now, since `--all` option was removed, it would not be occurred but just for note.
If `remove_redundant_reshape` option is enabled in the related tests, the number of `Reshape` is 1.

llFreetimell (approved):
LGTM

res/TensorFlowLiteRecipes/Net_Squeeze_Squeeze_000/test.rule
seanshpark: > since --all option was removed,

it's not removed yet :) This maybe revised then.

> If remove_redundant_reshape option is enabled

if there would be any cases, we may revsit this

seanshpark:
@mhs4670go , PTAL

PR #5839, opened 2021-01-27
[luci] Introduce SubstituteSqueezeToReshapePass

seanshpark:
This will introduce SubstituteSqueezeToReshapePass that will substitute
CircleSqueeze to CircleReshape for certain conditions.

### DIFF
--- a/compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
+++ b/compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
@@ -1,0 +1,169 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/SubstituteSqueezeToReshapePass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/**
+ * @brief return TRUE if all dim is known && dim > 0
+ */
+bool can_squeeze_shape(const luci::CircleNode *node)
+{
+  for (uint32_t r = 0; r < node->rank(); ++r)
+  {
+    if (not node->dim(r).known())
+      return false;
+    if (node->dim(r).value() < 1)
+      return false;
+  }
+  return true;
+}
+
+/**
+ * @brief return valid unsigned dim value from 0 ~ rank
+ * @note  dim can be -rank to rank
+ */
+uint32_t valid_unsigned_dim(uint32_t rank, int32_t dim)
+{
+  int32_t irank = static_cast<int32_t>(rank);
+  return dim >= 0 ? static_cast<uint32_t>(dim) : static_cast<uint32_t>(irank + dim);
+}
+
+/**
+ * @brief return TRUE if input dim is 1 for squeeze_dims values
+ */
+bool is_valid_input(const luci::CircleNode *node, const std::vector<int32_t> &squeeze_dims)
+{
+  auto rank = node->rank();
+  for (auto dim : squeeze_dims)
+  {
+    auto udim = valid_unsigned_dim(rank, dim);
+    if (node->dim(udim).value() != 1)
+      return false;
+  }
+  return true;
+}
+
+/**
+ * @brief return shape vector from input
+ */
+std::vector<uint32_t> node_shape(const luci::CircleNode *input)
+{
+  std::vector<uint32_t> shape;
+  uint32_t rank = input->rank();
+  for (uint32_t r = 0; r < rank; ++r)
+    shape.push_back(input->dim(r).value());
+
+  return shape;
+}
+
+/**
+ * @brief return CircleConst ptr with values of new_shape
+ */
+luci::CircleConst *create_shape_const(loco::Graph *graph, const std::vector<uint32_t> &new_shape)
+{
+  uint32_t dim_size = static_cast<uint32_t>(new_shape.size());
+
+  auto shape_const = graph->nodes()->create<luci::CircleConst>();
+
+  // const shape/dtype
+  shape_const->dtype(loco::DataType::S32);
+  shape_const->shape({dim_size});
+  shape_const->shape_status(luci::ShapeStatus::VALID);
+
+  // constant values
+  shape_const->size<loco::DataType::S32>(dim_size);
+  for (uint32_t i = 0; i < new_shape.size(); ++i)
+    shape_const->at<loco::DataType::S32>(i) = new_shape.at(i);
+
+  return shape_const;
+}
+
+bool substitute_squeeze_to_reshape(luci::CircleSqueeze *squeeze)
+{
+  assert(squeeze != nullptr);
+
+  auto input = loco::must_cast<luci::CircleNode *>(squeeze->input());
+  // we need input node shape and all dim should be known
+  if (input->shape_status() != luci::ShapeStatus::VALID)
+    return false;
+  if (not can_squeeze_shape(input))
+    return false;
+
+  // we will use squeeze shape for new shape
+  if (squeeze->shape_status() != luci::ShapeStatus::VALID)
+    return false;
+
+  auto squeeze_dims = squeeze->squeeze_dims();
+  if (not is_valid_input(input, squeeze_dims))
+    return false;
+
+  auto new_shape = node_shape(squeeze);
+  auto graph = squeeze->graph();
+  auto reshape = graph->nodes()->create<luci::CircleReshape>();
+  auto shape_const = create_shape_const(graph, new_shape);
+
+  // graph connection
+  reshape->tensor(input);
+  reshape->shape(shape_const);
+  replace(squeeze).with(reshape);
+
+  return true;
+}
+
+} // namespace
+
+namespace luci
+{
+
+/**
+ * BEFORE
+ *           |
+ *      [CircleNode]
+ *           |
+ *    [CircleSqueeze]
+ *           |
+ *      [CircleNode]
+ *           |
+ *
+ * AFTER
+ *               |
+ *          [CircleNode]  [CircleConst]
+ *             |    \             /
+ *  [CircleSqueeze] [CircleReshape]
+ *                        |
+ *                   [CircleNode]
+ *                        |
+ */
+bool SubstituteSqueezeToReshapePass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    if (auto squeeze = dynamic_cast<luci::CircleSqueeze *>(node))
+    {
+      if (substitute_squeeze_to_reshape(squeeze))
+        changed = true;
+    }
+  }
+  return changed;
+}
+
+} // namespace luci
--- a/compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
+++ b/compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
@@ -37,8 +37,8 @@ bool can_squeeze_shape(const luci::CircleNode *node)
 }
 
 /**
- * @brief return valid unsigned dim value from 0 ~ rank
- * @note  dim can be -rank to rank
+ * @brief return valid unsigned dim value from 0 ~ (rank-1)
+ * @note  dim can be -rank to (rank-1)
  */
 uint32_t valid_unsigned_dim(uint32_t rank, int32_t dim)
 {
@@ -79,18 +79,25 @@ std::vector<uint32_t> node_shape(const luci::CircleNode *input)
  */
 luci::CircleConst *create_shape_const(loco::Graph *graph, const std::vector<uint32_t> &new_shape)
 {
+  // NOTE dim_size can be 0
   uint32_t dim_size = static_cast<uint32_t>(new_shape.size());
 
   auto shape_const = graph->nodes()->create<luci::CircleConst>();
 
   // const shape/dtype
   shape_const->dtype(loco::DataType::S32);
-  shape_const->shape({dim_size});
+  if (dim_size > 0)
+  {
+    shape_const->rank(1);
+    shape_const->dim(0).set(dim_size);
+  }
+  else
+    shape_const->rank(0);
   shape_const->shape_status(luci::ShapeStatus::VALID);
 
   // constant values
   shape_const->size<loco::DataType::S32>(dim_size);
-  for (uint32_t i = 0; i < new_shape.size(); ++i)
+  for (uint32_t i = 0; i < dim_size; ++i)
     shape_const->at<loco::DataType::S32>(i) = new_shape.at(i);
 
   return shape_const;
--- a/compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
+++ b/compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
@@ -120,12 +120,12 @@ bool substitute_squeeze_to_reshape(luci::CircleSqueeze *squeeze)
 
   auto squeeze_dims = squeeze->squeeze_dims();
   if (not is_valid_input(input, squeeze_dims))
-    return false;
+    throw std::runtime_error("Invalid values in squeeze_dims: " + squeeze->name());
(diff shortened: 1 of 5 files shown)


seanshpark:
For #5830, from draft #5833

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+  int dim_1_cnt = 0;
+  for (uint32_t r = 0; r < node->rank(); ++r)
+  {
+    if (not node->dim(r).known())
+      return false;
+    if (node->dim(r).value() < 1)
+      return false;
+    if (node->dim(r).value() == 1)
+      dim_1_cnt++;
```
jinevening, line 37: I think we should check the number of `dim==1` in the `squeeze_dims`.

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  auto input = loco::must_cast<luci::CircleNode *>(squeeze->input());
+  // we need input node shape and all dim should be known
+  if (input->shape_status() != luci::ShapeStatus::VALID)
+    return false;
+  if (not can_squeeze_shape(input))
+    return false;
+
+  auto squeeze_dims = squeeze->squeeze_dims();
+  auto new_shape = new_shape_with_squeeze(input, squeeze_dims);
```
jinevening, line 115: Is there a reason not to use the shape of `squeeze`? Shape inference may have found the new shape.

seanshpark: Thanks, I'll check

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: you mean `squeeze_dims` in `CircleSqueeze` ?

I'm not godd at these but what I understand is that
value of `squeeze_dims` is which dimension to squeeze.
so `dim == 1` check for that would doing "we want to squeeze dim 1"

jinevening: My comment was a bit confusing. What I meant was that we should check the number of dimensions equal to 1 among the dimensions in the `squeeze_dims`.

> here, we are checking if any input dim can be squeezed so, which would be having value 1

I think checking if any input dim can be squeezed is not enough. Let's see the below case.

input shape: [1, 2, 3, 4]
squeeze_dims: [1]
-> This case is not squeeze-able, because `squeeze_dims` in the input shape is 2 (!= 1). So, interpreter (or other runtime) will throw an exception when executing this operator. However, if we substitute this case, the network will run without an error. I thought this can be problematic.

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+bool substitute_squeeze_to_reshape(luci::CircleSqueeze *squeeze)
+{
+  assert(squeeze != nullptr);
+
+  auto input = loco::must_cast<luci::CircleNode *>(squeeze->input());
+  // we need input node shape and all dim should be known
+  if (input->shape_status() != luci::ShapeStatus::VALID)
+    return false;
+  if (not can_squeeze_shape(input))
```
llFreetimell, line 111: Following case cannot be optimized by this condition.
```python
inp = tf.constant(1, shape=[2,3,4,5])
sq = tf.squeeze(inp)

# Expected optimized result : tf.reshape(inp, [2,3,4,5])
# Obtained optimized result : Not optimized
```

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,166 @@
...
+ * @brief Return TRUE if all dim is known && dim > 0 && at least dim 1 exist
+ */
+bool can_squeeze_shape(luci::CircleNode *node)
+{
+  int dim_1_cnt = 0;
+  for (uint32_t r = 0; r < node->rank(); ++r)
+  {
+    if (not node->dim(r).known())
+      return false;
+    if (node->dim(r).value() < 1)
```
llFreetimell, line 34: If `node` shape is [1, 0, 2, 1, 3], `Squeeze` result is [0, 2, 3] which can be squeezed.

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: Um? can you explain more about expected result?

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: Um? you mean we should also enable of dim value 0 ?

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,166 @@
...
+  else
+  {
+    // prepare new_shape from input shape excluding squeeze_dims
+    std::vector<uint32_t> temp_shape;
+    for (uint32_t r = 0; r < input->rank(); ++r)
+      temp_shape.push_back(input->dim(r).value());
+    for (auto dim : squeeze_dims)
+    {
+      assert(temp_shape.at(dim) == 1);
+      temp_shape.at(dim) = 0;
```
llFreetimell, line 69: Input node can include zero dimension such as [1, 0, 2, 3].
If we mark erased dimension as 0, real zero dimension cannot be distinguished.
So how about using -1 instead?

```suggestion
      temp_shape.at(dim) = -1;
```

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+    for (auto ts : temp_shape)
+      if (ts > 0)
+        new_shape.push_back(ts);
+  }
+
+  // TODO check if new_shape is empty
```
llFreetimell, line 76: I think `new_shape` can be empty if result is scalar. :)
(e.g, `Squeeze` of [1, 1] is [])

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,166 @@
...
+ */
+luci::CircleConst *create_shape_const(loco::Graph *graph, const std::vector<uint32_t> &new_shape)
+{
+  uint32_t dim_size = static_cast<uint32_t>(new_shape.size());
+
+  auto shape_const = graph->nodes()->create<luci::CircleConst>();
+
+  // const shape/dtype
+  shape_const->dtype(loco::DataType::S32);
+  shape_const->shape({dim_size});
```
llFreetimell, line 92: If `new_shape` is empty, it means scalar.
However, scalar( [] ) and [0] is different shape.

```suggestion
  shape_const->rank(dim_size);
  if (dim_size > 0)
    shape_const->shape({dim_size});
```

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: this block will be removed as shape for CircleReshape will be from output of CircleSqueeze node.

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: calling `shape_const->shape({dim_size});` will update rank too. I'll revise this.

llFreetimell: Oh, I wrote wrong suggestion code :(

My point was that if `dim_size` is 0, shape would be [0], not [] :)
Sorry for confusion.

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
llFreetimell:
```python
inp = tf.constant(1, shape=[2,3,4,5])
sq = tf.squeeze(inp)
```

If input of `Squeeze` has [2, 3, 4, 5] shape, any dimension will be squeezed.
So the result of `sq` would be same with `inp`.
It means that the `Squeeze` can be substituted by `Reshape` with new_shape [2, 3, 4, 5].

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: we are setting to rank(1) not rank(dim_size) dim(0) value is to be set to dim_size

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: > any dimension will be squeezed.

there is no `1` in input shape. so, I don't think there will be any shape change
If I'm not correct, please let me know.

> Squeeze can be substituted by Reshape with new_shape [2, 3, 4, 5].

OK, now I understand what you mean. For this case, Squeeze will not be doing what Squeeze  supposed to do.
So we can just replace this Reshape and then later it can be removed.
Is your suggestion like this?

llFreetimell: > there is no 1 in input shape. so, I don't think there will be any shape change
> I'm not correct, please let me know.

You are correct :)

> OK, now I understand what you mean. For this case, Squeeze will not be doing what Squeeze is supposed to do.
> Squeeze is doing nothing, like a dummy. So we can just replace this Reshape and then later it can be removed.
> Is your suggestion like this?

Exactly :D

jinevening (approved):
LGTM

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
llFreetimell: I wanted to know why zero dimension is treated as special case.
If this condition is used, following example cannot be optimized.

```python
x = tf.constant(1, shape=[1, 0, 2, 1, 3])
y = tf.squeeze(x)
print(y) # tf.Tensor([], shape=(0, 2, 3), dtype=int32)
```

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
llFreetimell: I wanted to describe following example.

```python
# Current code
x = tf.constant(1, shape=[1, 1])
y = tf.squeeze(x)
print(y) # tf.Tensor(1, shape=(), dtype=int32)

# AS-IS (Error)
asis_optimized_y = tf.reshape(x, [0])

# TO-BE
tobe_optimized_y = tf.reshape(x, [])
```

With above example, `new_shape` parameter in `luci::CircleConst *create_shape_const(loco::Graph *graph, const std::vector<uint32_t> &new_shape)` will be empty.
As a result, `dim_size` will be 0.
If we use `shape_const->shape({dim_size})`, the result would be `asis_optimized_y`.

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,169 @@
...
+      return false;
+    if (node->dim(r).value() < 1)
+      return false;
+  }
+  return true;
+}
+
+/**
+ * @brief return valid unsigned dim value from 0 ~ rank
+ * @note  dim can be -rank to rank
```
llFreetimell, line 41: Maybe...?

```suggestion
 * @brief return valid unsigned dim value from 0 ~ (rank-1)
 * @note  dim can be -rank to (rank-1)
```

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,169 @@
...
+ * @brief return TRUE if input dim is 1 for squeeze_dims values
+ */
+bool is_valid_input(const luci::CircleNode *node, const std::vector<int32_t> &squeeze_dims)
+{
+  auto rank = node->rank();
+  for (auto dim : squeeze_dims)
+  {
+    auto udim = valid_unsigned_dim(rank, dim);
+    if (node->dim(udim).value() != 1)
+      return false;
```
llFreetimell, line 59: Throwing seems better because it is invalid graph.

```python
x = tf.constant(1, shape=[1, 2, 3, 4])
y = tf.squeeze(x, axis=[1]) # Python throws Error
```

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: ah... just wanted to substitute for normal shapes :)
as I know 0 is used for unknown value so I'm not sure replace with Reshape is OK or not for this case...

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,169 @@
...
+    return false;
+
+  auto squeeze_dims = squeeze->squeeze_dims();
+  if (not is_valid_input(input, squeeze_dims))
+    return false;
+
+  auto new_shape = node_shape(squeeze);
+  auto graph = squeeze->graph();
+  auto reshape = graph->nodes()->create<luci::CircleReshape>();
+  auto shape_const = create_shape_const(graph, new_shape);
```
llFreetimell, line 121: Hmm.. `squeeze_dims` is not considered for new shape I think....

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: I think your example needs to be added to the test too

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
llFreetimell: Ah, 0 in `x = tf.constant(1, shape=[1, 0, 2, 1, 3])` is not unknown dimension but real zero dimension :D
Unknown dimension is represented as `None` such as `p = tf.compat.v1.placeholder(shape=[None, 1], dtype=tf.int32)`.

- zero dimension : `dim.known() == true && dim.value() == 0`
- unknown dimension : `dim.known() == false` (dim.value() is 0 but meaningless)

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: actually shape inference throws first :)

jinevening: IMHO, if we're going to regard such a graph as an invalid graph (this may need discussion as well), import would be a better place to throw an exception.

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: you mean, `new` is not appropriate? maybe `reshape_shape` would do?

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
seanshpark: so you mean we only need to check it's known or not...OK :)

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
llFreetimell: Ah, you are right. `reshape_shape` handles it :)

compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+#include "luci/Pass/SubstituteSqueezeToReshapePass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+/**
+ * @brief return TRUE if all dim is known
```
llFreetimell, line 25: There are some cases that this pass can be applied even some of dimensions are unknown.
```python
p = tf.compat.v1.placeholder(shape=[None, 1, 2], dtype=tf.int32)
s = tf.squeeze(p,[0])
print(s) # <tf.Tensor 'Squeeze:0' shape=(1, 2) dtype=int32>
```

However, same logic may be also used for shape inference of `CircleSqueeze`.

So how about going as it is and just adding following comments?
```suggestion
 * @brief return TRUE if all dim is known
 * @note This pass can be applied even some of dimensions are unknown.
         For now, do not consider about it and update logic later.
```

llFreetimell (approved):
LGTM

jinevening (approved):
LGTM

Issue #5785, opened 2021-01-22
Compiler FE: Refactoring quantization codes


jinevening:
## What

Create a new node when quantizing a const node.

## Why

There are cases where a const node is shared by multiple nodes. In-place quantization may cause an error in that case.

jinevening:
Constant tensors in the network can be quantized in various ways. I categorized them in two types.

1. Per-tensor quantization using its own value (Default: in-place quantization)

Find min/max from the values in a tensor and quantize all values in the tensor using the same qparam.

2. Other ways (constant tensor is copied and then quantized)
2.1 Weights: weights can be quantized per channel.
2.2 Bias: bias is quantized using qparams of weights and input tensor
2.3 Instnorm: gamma and beta of Instnorm can be quantized per channel
2.4 PRelu: alpha of PRelu can be quantized per channel
2.5 Concat (propagated qparam): constant inputs of concat can be quantized using the qparam of the concat

The second type of quantization will be done first. It will create a new node and quantize it. After that, the first type of quantization will be done in-place. This will allow the first-type-quantized tensors to be safely shared by multiple nodes.

jinevening:
All done.

PR #5844, opened 2021-01-27
[luci] Copy-based quantization for weights

jinevening:
This supports copy-based quantization for weights.

---
Related to: #5785

### DIFF
--- a/compiler/luci/pass/src/QuantizationUtils.h
+++ b/compiler/luci/pass/src/QuantizationUtils.h
@@ -47,6 +47,40 @@ bool is_weights(CircleNode *node);
 
 bool is_quantized(const CircleNode *node);
 
+// Clone a new const node from an existing node.
+// The new node has the same shape, type, contents, and qparam with 'node'
+template <loco::DataType T> luci::CircleConst *clone_const_from(luci::CircleConst *node)
+{
+  assert(T == node->dtype());
+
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/QuantizationUtils.h
```diff
@@ -47,6 +47,40 @@ bool is_weights(CircleNode *node);
...
+    new_node->dim(i).set(node->dim(i).value());
+
+  const auto size = node->size<T>();
+  new_node->size<T>(size);
+  for (uint32_t i = 0; i < size; i++)
+    new_node->at<T>(i) = node->at<T>(i);
+
+  new_node->shape_status(luci::ShapeStatus::VALID);
+
+  auto qparam = std::make_unique<CircleQuantParam>();
```
seanshpark, line 72: maybe do this and below only when `node->quantparam()` is not null?

seanshpark: or if `node->quantparam()` be always exist then add `assert(node->quantparam() != nullptr);` at the top would be nice.

jinevening: I've added the null check. Thanks.

seanshpark (approved):
LGTM thank you!

PR #5847, opened 2021-01-27
[res] Introduce Net_Squeeze_Squeeze_000 tfl recipe

seanshpark:
This will introduce Net_Squeeze_Squeeze_000 tflite recipe with two
Sequence ops.


seanshpark:
For #5830, from draft #5833

seanshpark:
duplicate

Issue #5845, opened 2021-01-27
Compiler FE: Support concat OP in ConvertNCHWToNHWCPass


jinevening:
## What

Support concat Op in ConvertNCHWToNHWCPass

## Why

We can remove redundant reshape Ops by supporting concat Op in ConvertNCHWToNHWCPass (by the follwing steps).

1. Transpose Ops are added before/after concat Ops
2. The Transpose Ops can be converted to Reshape Ops
3. Redundant/Unncessary Reshapes Ops are removed from the network

seanshpark:
@jinevening , will you be working on this?

jinevening:
> @jinevening , will you be working on this?

Yes, I'll take this issue.

jinevening:
Job is done.

PR #5849, opened 2021-01-27
[luci] Support CONCAT in ConvertNCHWToNHWCPass

jinevening:
This supports CONCAT in CovnertNCHWToNHWCPass.

---
Related to: https://github.com/Samsung/ONE/issues/5845

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -90,16 +90,10 @@ luci::CircleTranspose *create_4d_transpose(luci::CircleNode *node,
 int32_t nchw_axis_to_nhwc(int32_t axis)
 {
   uint32_t pos_axis = axis >= 0 ? static_cast<uint32_t>(axis) : static_cast<uint32_t>(axis + 4);
-  if (pos_axis == 0)
-    return 0;
-  else if (pos_axis == 1)
-    return 3;
-  else if (pos_axis == 2)
-    return 1;
-  else if (pos_axis == 3)
-    return 2;
-  else
+  static const uint32_t to_nhwc[4] = {0, 3, 1, 2};
+  if (pos_axis > 3)
     throw std::runtime_error("Concat axis must be in range [-4, 4)");
+  return to_nhwc[pos_axis];
 }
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -87,6 +87,21 @@ luci::CircleTranspose *create_4d_transpose(luci::CircleNode *node,
...
+  if (pos_axis == 0)
+    return 0;
+  else if (pos_axis == 1)
+    return 3;
+  else if (pos_axis == 2)
+    return 1;
+  else if (pos_axis == 3)
+    return 2;
+  else
+    throw std::runtime_error("Concat axis must be in range [-4, 4)");
```
seanshpark, line 102: How about something like this?
```cxx
uint32_t pos_axis = axis >= 0 ? static_cast<uint32_t>(axis) : static_cast<uint32_t>(axis + 4);
static const uint32_t to_nhwc[4] = {0, 3, 1, 2};
if (pos_axis > 3)
  throw std::runtime_error("Concat axis must be in range [-4, 4)");
return to_nhwc[pos_axis];
```

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -444,6 +459,32 @@ class ConvertNCHWToNHWC final : public luci::CircleNodeMutableVisitor<bool>
     return true;
   }
 
+  bool visit(luci::CircleConcatenation *node)
+  {
+    const auto num_values = node->numValues();
```
seanshpark, line 464: I think we need to check rank of node and if not 4 then return falsel;

jinevening: node rank is checked before the visitor is called.

Related code
```cpp
...
     ConvertNCHWToNHWC converter;
      if (circle_node->rank() != 4)
        continue;

      if (circle_node->accept(&converter))
      {
        set_data_format(node, DataFormat::NHWC);
        changed = true;
      }
...
```

seanshpark (approved):
LGTM thank you!

PR #5851, opened 2021-01-27
[luci] Enable test of Squeeze_001

seanshpark:
This will enable read/write test of SQUEEZE Op. with empty squeeze_dims.


llFreetimell (approved):
LGTM

PR #5858, opened 2021-01-28
[record-minmax] Add requires to luci-interpreter

seanshpark:
This will revise requires.cmake to include luci-interpreter.


seanshpark:
I got some link error and this is the fix.

mhs4670go (approved):
LGTM

jinevening (approved):
LGTM Thank you

PR #5854, opened 2021-01-27
DRAFT: Compiler FE: Swap Reshape-UnaryOp

seanshpark:
on-going draft to swap Reshape-UnaryOp to UnaryOp-Reshape.

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@samsung.com>

Issue #5848, opened 2021-01-27
Compiler FE: Swap of Reshape - UnaryOp from ONNX


seanshpark:
There is a pattern like this, in the middle of transformations from ONNX model.
- `Reshape` - `Neg` - `Reshape`

If we can swpa first `Reshape` and `Neg` then two `Reshape` - `Reshape` can be reduced to one `Reshape` with existing Pass.

seanshpark:
Draft seems to work.

_luci_pass_value_test_ fails... _luci_interpreter_ does not process `CircleNeg`...

seanshpark:
done

PR #5868, opened 2021-01-28
[res] Introduce Net_Reshape_Neg_000 tfl recipe

seanshpark:
This will introduce Net_Reshape_Neg_000 tflite recipe that has
Reshape-Neg operator sequence.


seanshpark:
For #5848, from draft #5854

mhs4670go (approved):
LGTM

PR #5873, opened 2021-01-28
[luci-pass-value-test] Enable Net_Reshape_Neg_000

seanshpark:
This will enable Net_Reshape_Neg_000 test with
forward_reshape_to_unaryop option.


seanshpark:
For #5848, from draft #5854

mhs4670go (approved):
LGTM

PR #5875, opened 2021-01-29
[onert/test] Compare float value using misc library

hseok-oh:
Compare float value in tflite-loader test by using misc::fp32 library (same with nnapi-test)


chunseoklee (approved):

wateret (approved):
LGTM

Issue #5867, opened 2021-01-28
[luci] Relocate clone() of CircleConst to helper


seanshpark:
From https://github.com/Samsung/ONE/pull/5862#discussion_r565893672, I think it would be better to relocate clone() to helper so that we may reuse this.

seanshpark:
done

PR #5877, opened 2021-02-01
DRAFT: [luci] CircleConst clone

seanshpark:
draft to provide clone of CircleConst in luci/service.


seanshpark:
For #5867

seanshpark:
all landed

PR #5878, opened 2021-02-01
[luci] Introduce clone for CircleConst

seanshpark:
This will introduce clone method for CircleConst node.


seanshpark:
For #5867, from draft #5877

jinevening (approved):
LGTM

PR #5880, opened 2021-02-01
[luci] Revise ForwardReshapeToUnaryOpPass use luci::clone

seanshpark:
This will revise ForwardReshapeToUnaryOpPass to use luci::clone for
CircleConst and remove local clone method.


seanshpark:
For #5867, from draft #5877

jinevening (approved):
LGTM

PR #5888, opened 2021-02-01
[luci] Copy-based propagation of concat qparam

jinevening:
This supports copy-based propagation of concat qparam.

---
Related to: #5785


jinevening:
For reviewers: The major change is done in `QuantizeWithMinMax.cpp`. `PropagateConcatenationQparam.test.cpp` was also modified as const inputs of concat are 'copied' and then quantized.

seanshpark (approved):
LGTM

PR #5885, opened 2021-02-01
DRAFT: Compiler FE: Revise --all to --O1

seanshpark:
draft to revise --all option to --O1

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #5780, opened 2021-01-21
Compiler FE: Remove optimization "--all"


seanshpark:
The `--all` option was first introduced to enable all the optimization options but currently it didn't turned out that way.
To add new option to the `all` maybe risky in terms of stability.
As of result, `all` misleads that one may expect it will do all optimizations but actually it doesn't.

I think it would be better to remove this `all` option and provide some more specific group of options as an alias.

I'd like to know if there are any problems before just removing this option.
If there is no comments about a week, I'll go on removing them.

seanshpark:
CC @samsung/ONE

lemmaa:
As a suggestion, what if we provide the `-O1`, `-O2`, and `-O3` options like c compiler? 

As you might expect, it's kind of a group of optimization options. Probably the most stable optimizations are reflected in `-O1`, the slightly challenging ones in `-O2`, and the risky, perhaps current `--all` items are reflected through `-O3`. Of course, the configuration of the options in this group can be changed at any time at our discretion, so users can use it without confusion. The long format of this option would be around `--optimization-level={1,2,3}`. 

In addition, options such as `-O4` or `--experimental` could be added.

seanshpark:
> As a suggestion, what if we provide the -O1, -O2, and -O3 options like c compiler?

Thanks! I'll add another issue for this, as this issue is  just about removing `--all` option.

seanshpark:
Where `--all` is used, for testing
- compiler/common-artifacts/CMakeLists.txt
```cmake
  if(NOT DEFINED NO_OPTIMIZE_${RECIPE})
    # Generate optimized .circle
    add_custom_command(OUTPUT ${OPT_CIRCLE_OUTPUT_PATH}
      COMMAND $<TARGET_FILE:circle2circle> --all ${CIRCLE_OUTPUT_PATH} ${OPT_CIRCLE_OUTPUT_PATH}
      DEPENDS $<TARGET_FILE:circle2circle>  ${CIRCLE_OUTPUT_PATH}
      COMMENT "Generate ${OPT_CIRCLE_FILE}"
    )
    set(OPT_FORMAT ".opt")
    list(APPEND TEST_DEPS ${OPT_CIRCLE_OUTPUT_PATH})
  endif()
```
- docs/compiler/compiler/common-artifacts/CMakeLists.txt
```cmake
  if(NOT DEFINED NO_OPTIMIZE_${RECIPE})
    # Generate optimized .circle
[...]

PR #5889, opened 2021-02-01
[one-cmds] Add option --O1

seanshpark:
This will enable --O1 option to replace --all.


seanshpark:
For #5780, from draft #5885

mhs4670go (approved):
LGTM

PR #5890, opened 2021-02-01
[common-artifacts] Use --O1 for optimize

seanshpark:
This will revise to use --O1 instead of --all for optimization test.


seanshpark:
For #5780, from draft #5885

mhs4670go (approved):
LGTM

PR #5891, opened 2021-02-01
[luci] Quantize const inputs shared by multiple nodes

jinevening:
This supports quantization of const inputs shared by multiple nodes.

---
Related to: #5785


seanshpark (approved):
LGTM

PR #5893, opened 2021-02-01
[one-cmds] Use --O1 for one-optimize tests

seanshpark:
This will revise one-optimize tests to use --O1 for --all.


seanshpark:
For #5780, from draft #5885

mhs4670go (approved):
LGTM

PR #5895, opened 2021-02-01
[one-cmds] Remove --all option in one-optimize

seanshpark:
This will remove not used anymore --all option in one-optimize.


seanshpark:
For #5780, from draft #5885

mhs4670go (approved):
LGTM

PR #5896, opened 2021-02-01
[luci-pass-value-test] Revise README.md

seanshpark:
This will revise README.md file using bullets to fix formats.


mhs4670go (approved):
LGTM

Issue #5886, opened 2021-02-01
Compiler FE: Running record-minmax with random inputs


jinevening:
## What

Run `record-minmax` with random inputs if `--input_data` option is not given.

## Why

To reduce programming burden for testing. We don't need real calibration data for testing.

jinevening:
Done.

PR #5897, opened 2021-02-01
DRAFT: Run record-minmax with random inputs

jinevening:
On going draft to run record-minmax with random inputs if input_data is not given.

---
Related to: https://github.com/Samsung/ONE/issues/5886


jinevening:
Done

Issue #5870, opened 2021-01-28
Compiler FE: ONNX import exec result validation


seanshpark:
It would be safe to prepare a execution result validation method
- ONNX model + ONNX runtime => Run with random input -> result_onnx
- ONNX model --> ... --> optimized circle model + luci_interpreter =>  Run with same random input -> result_circle

Compare two results: result_onnx vs result_circle

Things to considier
- ONNX input should be NCHW, circle input should be NHWC (or we may use preserve shape option)
- There are in-house models that cannot share

seanshpark:
Roughly it's not that complicated to build up this but as there are things to considier, I would like to prepare a local version first and see how this turns out...

seanshpark:
Running onnx with onnx-runtime
- https://pytorch.org/docs/stable/onnx.html
- https://pytorch.org/tutorials/advanced/super_resolution_with_onnxruntime.html

seanshpark:
How to get input/output shape, name from onnx model?
- TBA

seanshpark:
done

PR #5901, opened 2021-02-02
[luci-eval-driver] Introduce luci-eval-driver

seanshpark:
This will introduce luci-eval-driver copied from luci-eval-tester inside luci-value-test.


seanshpark:
This is related with #5870 and as _luci-pass_value_test_ also uses this executable, it is worth promoting to a module.

mhs4670go (approved):
LGTM

PR #5904, opened 2021-02-02
DRAFT: Compiler FE: import onnx validate

seanshpark:
on going draft of import onnx execution result validation


seanshpark:
done

PR #5905, opened 2021-02-02
[luci-pass-value-test] Use luci-eval-driver

seanshpark:
This will revise to use luci-eval-driver for evaluation.


seanshpark:
From draft #5904

mhs4670go (approved):
LGTM

PR #5909, opened 2021-02-02
[compiler-ahub] Revise config

seanshpark:
This will revise config.yml to include luci-eval-driver and
luci-pass-value-test.


seanshpark:
From draft #5904

lemmaa (approved):
LGTM.

PR #5773, opened 2021-01-21
[onert] Refactor backend graph (All backends)

wateret:

PR #5912, opened 2021-02-02
[onert] Add methods for adding operands/operations

wateret:
Add methods for adding operands/operantions with given indices.

These methods are going to be used for building partial graphs.

This makes use of #5481 and also is part of #5773 .


hseok-oh (approved):
LGTM

PR #5916, opened 2021-02-02
[tflchef] Revise the way it cooks custom operator

mhs4670go:
This commit revises the way it cooks custom operator to fix wrong result.

Until now, it found out custom op's opcode index by `tflite::BuiltinOperator_CUSTOM`. But to find out correct one, it need to consider `operation.type()`.

I added `custom_code_vec` attribute to `CookParams`, which is used in `cook_graph`.

Related : #5786
Draft: #5814

### DIFF
--- a/compiler/tflchef/core/src/ModelChef.cpp
+++ b/compiler/tflchef/core/src/ModelChef.cpp
@@ -197,6 +197,7 @@ struct CookParams
   std::vector<flatbuffers::Offset<::tflite::SubGraph>> &subgraph_vec;
   std::unique_ptr<flatbuffers::FlatBufferBuilder> &flatbuffer_builder;
   std::map<tflite::BuiltinOperator, int32_t> &builtin_code_map;
+  std::vector<std::string> &custom_code_vec;
   std::string noname;
 };
 
@@ -209,6 +210,7 @@ template <typename T> void cook_graph(const T &graph, CookParams &cp)
   std::vector<flatbuffers::Offset<::tflite::SubGraph>> &subgraph_vec = cp.subgraph_vec;
   std::unique_ptr<flatbuffers::FlatBufferBuilder> &flatbuffer_builder = cp.flatbuffer_builder;
   std::map<tflite::BuiltinOperator, int32_t> &builtin_code_map = cp.builtin_code_map;
+  std::vector<std::string> &custom_code_vec = cp.custom_code_vec;
 
   // Operand-related
   std::vector<flatbuffers::Offset<::tflite::Tensor>> tensor_vec;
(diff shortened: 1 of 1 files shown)


compiler/tflchef/core/src/ModelChef.cpp
```diff
@@ -480,11 +482,23 @@ template <typename T> void cook_graph(const T &graph, CookParams &cp)
...
+    {
+      opcode_index = std::distance(builtin_code_map.begin(), op_it);
+    }
+    // custom operator
+    else
+    {
+      auto op_it = std::find(custom_code_vec.begin(), custom_code_vec.end(), operation.type());
+      assert(op_it != custom_code_vec.end());
+      opcode_index += builtin_code_map.size();
+      opcode_index += std::distance(custom_code_vec.begin(), op_it);
```
seanshpark, line 500: Um? Can you explain why we add both?

compiler/tflchef/core/src/ModelChef.cpp
```diff
@@ -552,7 +566,7 @@ GeneratedModel cook(const ::tflchef::ModelRecipe &model_recipe)
   std::vector<flatbuffers::Offset<::tflite::Buffer>> buffer_vec;
 
   // Operation-related
-  std::vector<flatbuffers::Offset<::tflite::OperatorCode>> code_vec;
+  std::vector<flatbuffers::Offset<::tflite::OperatorCode>> operator_code_vec;
```
seanshpark, line 569: can you explain why we need to rename this?

compiler/tflchef/core/src/ModelChef.cpp
mhs4670go: > opcode_index is an index into the operator_codes vector.
> operator_codes consists of buildtin_code and custom_code, which is inserted sequentially.

`opcode_index` size is `builtin_code_map` size + `custom_code_vec` size. Since custom codes is preceded by builtin codes, the offset of custom operator's `opcode_index` is `builtin_code_map` size.

compiler/tflchef/core/src/ModelChef.cpp
mhs4670go: @seanshpark It is not necessary but I just wanted to make users easy to differentiate `operator_code_vec` from `custom_code_vec`.

compiler/tflchef/core/src/ModelChef.cpp
seanshpark: OK, thanks for the explanation;
Then is the code guarantee that `builtin_code_map` WILL NOT change when this block executes?

And can you use like this?
```
      opcode_index = builtin_code_map.size();
      opcode_index += std::distance(custom_code_vec.begin(), op_it);
```

compiler/tflchef/core/src/ModelChef.cpp
seanshpark: Um, then maybe in another PR.

compiler/tflchef/core/src/ModelChef.cpp
mhs4670go: @seanshpark I think `const` keyword would help.

seanshpark: > I think const keyword would help.

Um? I don't get what const has go to do with guarantee ...

If this code block executes twice and in between `builtin_code_map` is increased then I code may not work correctly...

mhs4670go: @seanshpark 
`builtin_code_map` and `custom_code_vec` are not changed in `cook_graph` function. Could you explain the situation where they might be increased?

seanshpark: > Could you explain the situation where they might be increased?

I was just concerned.

> builtin_code_map and custom_code_vec are not changed in cook_graph function. 

Then it's OK.

seanshpark (approved):
LGTM thank you!

Issue #5907, opened 2021-02-02
Compiler FE: check fuse_activation_function is working correctly


seanshpark:
`fuse_activation_function` seems not working correctly with value evaluation.

Let's prepare a test recipe and find out the reason.

seanshpark:
Related #5904

seanshpark:
Concatenation can fuse activation. But current _luci-interpreter_ doesn't handle this.

seanshpark:
done for this issue

PR #5917, opened 2021-02-02
[luci-interpreter] Add check for concat activation function

jinevening:
This adds a check if concat has a fused activation function.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Concatenation.test.cpp
+++ b/compiler/luci-interpreter/src/kernels/Concatenation.test.cpp
@@ -174,6 +174,7 @@ TEST(ConcatenationTest, Unsupported_Configure_Type_NEG)
   EXPECT_ANY_THROW(kernel.configure());
 }
 
+// TODO: Remove this test when concat w/ fused_activation is supported
 TEST(ConcatenationTest, With_Fused_Activation_NEG)
 {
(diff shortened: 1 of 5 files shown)


compiler/luci-interpreter/src/kernels/Concatenation.test.cpp
```diff
@@ -159,6 +168,23 @@ TEST(ConcatenationTest, Unsupported_Configure_Type_NEG)
   ConcatenationParams params{};
 
   params.axis = -1;
+  params.activation = luci::FusedActFunc::NONE;
+
+  Concatenation kernel({&input1_tensor, &input2_tensor}, &output_tensor, params);
+  EXPECT_ANY_THROW(kernel.configure());
+}
+
+TEST(ConcatenationTest, With_Fused_Activation_NEG)
```
seanshpark, line 177: Does this test needs to be removed when fused activation works?

jinevening: Yes. At that time, this test should be removed.

seanshpark: Ah, then please add a comment about this...

seanshpark (approved):
LGTM thank you!

seanshpark:
For the record, #5907 is related.

PR #5918, opened 2021-02-02
[luci] Enable ReLU for convert to NHWC

seanshpark:
This will enable ReLU for convert NCHW to NHWC Pass.


jinevening (approved):
LGTM

PR #5919, opened 2021-02-02
[res] Introduce recipe for Conv with Min/Max and ReLU6

seanshpark:
This will introduce two recipes of small network of Conv with Min/Max
and Conv with ReLU6.


seanshpark:
From draft #5904

jinevening (approved):
LGTM

PR #5926, opened 2021-02-02
[onert] Refactor BackendContext

wateret:
Pass a partial graph when creating a BackendContext. "partial graph"
here means a subset of an original graph that contains
operations/operands used by a backend. Before this commit, each backend
had a reference to the whole graph which could be abused and error-prone.

- Remove `initializeBackendContext` - as a partial graph is passed for
  each graph, we do not need to build used operation/operand lists
  anymore
- Introduce `ContextData` which contains all the data for the backend
  - Make parameter of `Backend::newContext` simple - pass a single
    struct `ContextData` instead
  - Make `BackendContext` methods(`genTensors` and `genKernels`) not
    have any parameters, just use `ContextData`.

More Context : #5006 , https://github.com/Samsung/ONE/issues/5352#issuecomment-747361763

### DIFF
--- a/runtime/onert/core/src/compiler/ExecutorFactory.cc
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.cc
@@ -18,6 +18,7 @@
 
 #include <deque>
 #include <functional>
+#include "ir/OperationCloner.h"
 #include "exec/ExecutionObservers.h"
 #include "exec/LinearExecutor.h"
 #include "exec/DataflowExecutor.h"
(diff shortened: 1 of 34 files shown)


wateret:
Sorry for such a huge diff, FYI, all the changes are about switching to **partial graphs**(https://github.com/Samsung/ONE/issues/5352#issuecomment-747361763) and introducing **ContextData**.

hseok-oh (approved):
LGTM

intom (approved):
LGTM

runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -96,31 +97,115 @@ void initializeSubgraphIOTensors(compiler::LoweredGraph &lowered_graph,
...
+    data.graph->finishBuilding();
+    // Handle graph input/outputs or external tensors
+    data.graph->operands().iterate([&](const ir::OperandIndex &ind, const ir::Operand &operand) {
+      if (whole_graph.getInputs().contains(ind) || whole_graph.getOutputs().contains(ind))
+        data.external_operands.add(ind);
+      // Inputs are either "graph input" or "no def op and non-constant"
+      if (whole_graph.getInputs().contains(ind) ||
+          (!operand.getDef().valid() && !operand.isConstant()))
+        // Outputs are either "graph output" or "no uses"
+        data.graph->addInput(ind);
```
intom, line 199: I have a question.
Are the Input/Output of all partial graphs the same as the Input/Output of the whole graph?

wateret: The whole graph's input/outputs are always input/outputs in partial graphs.
Plus, I had a lot of thoughts about it and the conclusion was like below.
Here's a figure as an example.

NewBackendContext (2)

Partial Graph Input

- The whole graph's input (Backend Green %0) OR
- There is no def operation and not a constant (Backend Red %1, Backend Green %2, %3)

Partial Graph Output

- The whole graph's output (Backend Green %5) OR
- There is no use operations (Backend Red %3, Backend Green %1)

I'm not sure this is the best way to define partial graph's I/O 😓 but it works with our current compilation implementation for now.

PR #5927, opened 2021-02-03
[one-cmds] Experimental onnx2circle validation

seanshpark:
This will introduce experimental onnx to circle conversion-optimization
validation tool to check model conversion and optimization is valid
through executing two models and comparing the results.


seanshpark:
For #5870, from draft #5904

mhs4670go (approved):
LGTM

PR #5928, opened 2021-02-03
[luci] Set node id at luci/import

llFreetimell:
Related Issue : #5794 
Draft : #5821 

For profiling, each node sometimes should have unique node id.
This commit will set unique node id using operator index at `luci/import` step.


cgbahk (dismissed):
LGTM

seanshpark:
@llFreetimell , how about revise `void GraphBuilder::build()` to return new `CircleNode *` and set the ID in `convert_graph()` after it gets the node.

llFreetimell:
@seanshpark
Suggestion itself looks good but it may require a lot of change. 
(When I searched briefly, some of operations like `CircleIf`, `CircleSplit`, etc are implemented with `void GraphBuilder::build()`, not with `CircleNode *GraphBuilder::build_node()`)
So how about going as it is and revise it in the near future? :)

seanshpark:
> So how about going as it is and revise it in the near future? :)

OK, there are four. I think it's not a huge change.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

seanshpark:
> (When I searched briefly, some of operations like CircleIf, CircleSplit, etc are implemented with void GraphBuilder::build(), not with CircleNode *GraphBuilder::build_node())

Have you checked with `CircleSplit` and others having it's own `CircleSplitGraphBuilder::build()` ?

seanshpark:
As If, Split and While are not our current interest as of now so this code may work. But strictly speaking this change is incorrect.
You have two(or more?) choices
1. Add this kind of change also to If, Split and While, and then do revision
1. Revise `void GraphBuilder::build()` to `CircleNode *GraphBuilder::build()` and then set NodeID inside `convert_graph()`

llFreetimell:
I will try option 2 first, and there is no big problem, I will go with option 2 :)
Thanks for sharp review :D

llFreetimell:
It is rebased, PTAL :)

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

cgbahk (approved):
LGTM

Issue #5940, opened 2021-02-03
[Compiler][luci] Revise build() function in Graphbuilder


llFreetimell:
### What?

Let's revise `build()` function in `GraphBuilder` to return `CircleNode`.

### Why?

This issue is originated from https://github.com/Samsung/ONE/pull/5928#issuecomment-772235357.
If we apply this change, we can apply additional setting to built operation at importer in the future.

#### As-is

```cpp
virtual void build(const circle::OperatorT &op, GraphBuilderContext *context) const = 0;
```

#### To-be

```cpp
virtual CircleNode *build(const circle::OperatorT &op, GraphBuilderContext *context) const = 0;
```

**NOTE** If `build()` function creates more than one `CircleNode` like `CircleSplit`, `CircleIf` and so on, it returns only one representative node. (For example, if we build for `CircleSplit`, `CircleSplitOut` and `CircleSplit` would be created. Then only `CircleSplit` is returned)

seanshpark:
> if we build for CircleSplit, CircleSplitOut and CircleSplit would be created. Then only CircleSplit is returned

Is this OK for setting node ID?

seanshpark:
And for virtual nodes, do we have any policy for them with assigning node ID ?

llFreetimell:
> if we build for CircleSplit, CircleSplitOut and CircleSplit would be created. Then only CircleSplit is returned
> Is this OK for setting node ID?
> And for virtual nodes, do we have any policy for them with assigning node ID ?

It seems better that virtual nodes do not have node ID because of following reasons

- `CircleNodeID` is operator index which is the sequence of operators in circle file. Virtual nodes are not in there.
- Even if `CIrcleNodeID` is assigned to virtual nodes, it will not be used because the origin or some nodes should be always in the circle file.

So setting node ID with above plan seems okay.

llFreetimell:
Done by #5943 :)

PR #5943, opened 2021-02-03
[luci] Change return type of GraphBuilder::build function

llFreetimell:
Parent Issue : #5940

Until now, return type of `*GraphBuilder::build()` function was `void`.
If we change it to return built operation node, we can set additional attribute at `importer.cpp`.
This commit will change return type of `GraphBuilder::build()` function to `CircleNode *`,
and apply this change to related codes.


seanshpark (approved):
LGTM thank you!

seanshpark:
There are changes more than expeced, but it's OK :)

mhs4670go (approved):
LGTM

PR #5947, opened 2021-02-04
[luci] Add const keyword to parameter of CircleNodeID functions

llFreetimell:
`has_node_id` and `get_node_id` functions have no needs to change `circle_node`.
Therefore adding `const` keyword to them is safe.
This commit will add `const` keyword to parameters of `CircleNodeID` functions.


seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #5931, opened 2021-02-03
Compiler FE: Test coverage is low with compiler.luci.pass.include.luci.Pass


seanshpark:
Current `compiler.luci.pass.include.luci.Pass` header files are not visited in test coverage.
Let's solve this issue.

FYI: daily compiler coverage as of now
```
Packages   94% Files   95% Classes   95% Methods   86% Lines   87% Conditionals   100%
```

seanshpark:
done

PR #5948, opened 2021-02-04
[luci] Increase test coverage

seanshpark:
on-going draft to increase test coverage.


seanshpark:
For #5931

PR #5949, opened 2021-02-04
[luci] Tests for Pass name

seanshpark:
This will introduce luci Pass tests for coverage with name method.


seanshpark:
For #5931, from draft #5948

mhs4670go (approved):
LGTM

mhs4670go:
BTW, I'm just curious about why `logo::Pass` introduced a name method because `name` is not used..``)

seanshpark:
>  why logo::Pass introduced a name method because name is not used..``)

It's used in logging.

jinevening (approved):
LGTM

PR #5954, opened 2021-02-04
[loco] Support more types

binarman:
Add support for U16, U64 and FLOAT64 types


binarman:
My goal was to complete type support, but unfortunately I am not sure what basic type should be used for float16 (probably uint16_t?)

So for now I just completed support of standard types existing in C/C++

seanshpark (approved):
LGTM thank you!

parjong (approved):

parjong:
Just for note. I persoanlly think that this is one of design flaw of _loco_. I SHOULD allow each framework based on _loco_ to have its own type system.. T_T

PR #5955, opened 2021-02-04
[onert] Refine nnfw_api_internal.cc

wateret:
Remove const_cast in `primary_subgraph()`

- Make `primary_subgraph()` return a const pointer
- Replace `primary_subgraph()` use with `_subgraphs->primary()`
  when it is in the right state

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -346,13 +346,6 @@ NNFW_STATUS nnfw_session::prepare()
     return NNFW_STATUS_INVALID_STATE;
   }
 
-  if (!_subgraphs || !primary_subgraph() || primary_subgraph()->isBuildingPhase())
-  {
-    std::cerr << "Error during model prepare : "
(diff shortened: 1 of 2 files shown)


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -346,7 +346,8 @@ NNFW_STATUS nnfw_session::prepare()
     return NNFW_STATUS_INVALID_STATE;
   }
 
-  if (!_subgraphs || !primary_subgraph() || primary_subgraph()->isBuildingPhase())
+  auto primary_subgraph = _subgraphs->primary();
+  if (!_subgraphs || !primary_subgraph || primary_subgraph->isBuildingPhase())
```
hseok-oh, line 350: It checks `_subgraphs` is `nullptr` after `_subgraphs->primary()`

wateret: You're right and I found this redundant as it is all checked from the right above. So I removed the whole checks here.

hseok-oh (approved):
LGTM

PR #5956, opened 2021-02-04
[luci] Refactor size-related methods in Constant OP

binarman:
- Add non-template methods
- Support more types

Depends on #5954 

### DIFF
--- a/compiler/luci/lang/include/luci/IR/Nodes/CircleConst.h
+++ b/compiler/luci/lang/include/luci/IR/Nodes/CircleConst.h
@@ -37,6 +37,9 @@ public:
   CircleConst() = default;
 
 public:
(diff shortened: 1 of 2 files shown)


binarman:
@seanshpark 
I want to generalize Constant node and this is part of this attempt.
I also want to add "raw" data accessor s, that would be useful in case we want to just copy data from or to other places.

What do you think?

seanshpark:
> I want to generalize Constant node and this is part of this attempt.

If this is safe, it's OK to me. But there could be huge amount of small changes...

> I also want to add "raw" data accessor s, that would be useful in case we want to just copy data from or to other places.

Seems reasonable... we didn't have any requirements for this but I think you need one now...

seanshpark:

PR #5961, opened 2021-02-04
DRAFT: More compiler unit tests to increase converage

seanshpark:
on-going draft to increase unit test coverage in compiler frontend.

Signed-off-by: SaeHie Park <saehie.park@samsung.com>

PR #5966, opened 2021-02-05
[luci] Revise export register I/O tensor

seanshpark:
This will revise export implementation to use luci input_node,
output_node methods.


seanshpark:
From draft #5961

mhs4670go (approved):
LGTM

Issue #5960, opened 2021-02-04
Compiler FE: Low test coverage files


seanshpark:
From test coverage report as-of-now [daily-compiler-coverag](http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/181/cobertura/) shows some files need testing.

- [ ] compiler.circle-quantizer.src
- [compiler.luci.pass.src](http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/181/cobertura/compiler_luci_pass_src/)
  - [ ] FoldDequantizePass.cpp
  - [ ] FuseBCQPass.cpp
  - [ ] MakeBatchNormGammaPositivePass.cpp
  - [ ] QuantizationUtils.cpp
  - [ ] QuantizeDequantizeWeightsPass.cpp
  - [ ] QuantizeWithMinMaxPass.cpp
  - [ ] RequantizePass.cpp
  - [ ] SparsifyTensorPass.cpp

There are more but let's add unit tests for these for now.

seanshpark:
@jinevening , @llFreetimell @mhs4670go could you guys please take a look?
I'd like to add unit tests but I think authors would be better to write quality test codes, and I'll be working on other codes.
these are not urgent but need changes...

seanshpark:
How to make a local coverage report

install `lcov`. this is mine
```
$ lcov --version
lcov: LCOV version 1.13
```

Configure and run with coverage enabled
- you may have to clean existing cmake configuration cache manually
```
infra/scripts/configure_compiler_coverage.sh
./nncc build -j6
LUGI_LOG=1 ./nncc test
```
make a report
```
rm -rf coverage
CURRENT_DIR=$(pwd)
NNAS_PROJECT_PATH=$(CURRENT_DIR)/ NNCC_WORKSPACE=$(CURRENT_DIR)/ ./nnas gen-coverage-report compiler
```

open `coverage/html/index.html` in browser.

if this doesn't work, please let me know

seanshpark:
Other files that I'm working 
- [ ] luci/service/src/Nodes/CircleConst.cpp
- [ ] TBA

seanshpark:
close as nonactive for some time

PR #5967, opened 2021-02-05
[luci] Tests for CircleConst IR

seanshpark:
This will add unit tests for CircleConst IR.


seanshpark:
For #5960, from draft #5961

mhs4670go (approved):
LGTM

PR #5968, opened 2021-02-05
[luci] More tests for CircleConst::clone

seanshpark:
This will revise and add more tests for CircleConst::clone method.


seanshpark:
For #5960, from draft #5961

jinevening (approved):
LGTM

PR #5969, opened 2021-02-05
[luci] Tests for to_shape_description

seanshpark:
This will add unit tests for to_shape_description method.

### DIFF
--- a/compiler/luci/service/src/ShapeDescription.test.cpp
+++ b/compiler/luci/service/src/ShapeDescription.test.cpp
@@ -16,7 +16,23 @@
 
 #include "luci/Service/ShapeDescription.h"
 
+#include <luci/IR/CircleNode.h>
+#include <luci/IR/Nodes/CircleConst.h>
+
 #include <gtest/gtest.h>
+
+TEST(ShapeDescriptionTest, CircleNode)
+{
+  // Use CircleConst as CircleNode
+  luci::CircleConst circle_const;
+  circle_const.shape({1, 2, 3, 4});
+
+  auto sd = luci::to_shape_description(&circle_const);
+
+  ASSERT_EQ(4, sd._dims.size());
+  ASSERT_EQ(1, sd._dims.at(0));
+  ASSERT_TRUE(sd._rank_known);
+}
 
 TEST(ShapeDescriptionTest, TensorShape)
 {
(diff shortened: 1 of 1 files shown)


seanshpark:
For #5960, from draft #5961

compiler/luci/service/src/ShapeDescription.test.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+TEST(ShapeDescriptionTest, TensorShape)
+{
+  loco::TensorShape tensor_shape{1, 2, 3, 4};
+  loco::NodeShape node_shape(tensor_shape);
+
+  auto sd = luci::to_shape_description(node_shape);
+
+  ASSERT_EQ(4, sd._dims.size());
+  ASSERT_EQ(1, sd._dims.at(0));
```
llFreetimell, line 29:
```suggestion
  ASSERT_EQ(1, sd._dims.at(0).value());
  ASSERT_EQ(2, sd._dims.at(1).value());
  ASSERT_EQ(3, sd._dims.at(2).value());
  ASSERT_EQ(4, sd._dims.at(3).value());
```

compiler/luci/service/src/ShapeDescription.test.cpp
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Service/ShapeDescription.h"
+
+#include <gtest/gtest.h>
+
+TEST(ShapeDescriptionTest, TensorShape)
```
llFreetimell, line 21: (Optional) As there is one more `to_shape_description` function, how about adding one more test, `TEST(ShapeDescriptionTest, CircleNode)`?

compiler/luci/service/src/ShapeDescription.test.cpp
seanshpark: `sd` is `ShapeDescription` type which is
```cxx
struct ShapeDescription
{
  std::vector<int32_t> _dims;
  bool _rank_known;
};
```
so suggested change fails in compilation.
And, I think just check first value is enough for this test case.

llFreetimell (approved):
LGTM

PR #5970, opened 2021-02-05
[luci] Revise header of CircleExporterImpl

seanshpark:
This will revise header to include used vector and remove stdexcept.


seanshpark:
from draft #5961

mhs4670go (approved):
LGTM

PR #5971, opened 2021-02-05
[WIP] [onert] Revise test tools and libraries

hseok-oh:
Revise test tools and related libraries


hseok-oh:
Outdated. And we will use updated tflite version.

PR #5973, opened 2021-02-05
[onert] Remove Graph state

wateret:
As we have a lot of things that manipulate a graph after finishing
building, so let's remove the graph state.

- Remove graph state (BUILDING/FINISHED)
- Graph now should always be valid
    - No mismatches of edges between operations/operands
    - And so on...
- From now on, operands must be added before operations are added
    - Operands should not contain Def/Use info when it first is added
    - This looks not very good, let's make it better gradually


hseok-oh (approved):
LGTM

PR #5974, opened 2021-02-07
[locop] Add format test for Plain

seanshpark:
This will add FormattedTensorShape test for Plain format.

### DIFF
--- a/compiler/locop/src/FormattedTensorShape.test.cpp
+++ b/compiler/locop/src/FormattedTensorShape.test.cpp
@@ -40,6 +40,7 @@ TEST(FormattedTensorShapeTest, PlainFormat)
 
   tensor_shape->rank(2);
   tensor_shape->dim(0) = 4;
+  tensor_shape->dim(1) = 8;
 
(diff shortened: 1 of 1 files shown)


seanshpark:
For #5960

compiler/locop/src/FormattedTensorShape.test.cpp
```diff
@@ -33,3 +33,15 @@ TEST(FormattedTensorShapeTest, BracketFormat)
 
   SUCCEED();
 }
+
+TEST(FormattedTensorShapeTest, PlainFormat)
+{
+  auto tensor_shape = stdex::make_unique<loco::TensorShape>();
+
+  tensor_shape->rank(2);
+  tensor_shape->dim(0) = 4;
```
mhs4670go, line 42: It doesn't need `tensor_shape->dim(1)`?

seanshpark: Ah, copied from existing code and had a missing code!
Thanks!

mhs4670go (approved):
LGTM

PR #5979, opened 2021-02-08
[tflchef] Revise entry for testing

seanshpark:
This will introduce entry_stream method to make test code easy.


seanshpark:
For #5960, from draft #5961

mhs4670go (approved):
LGTM

PR #5987, opened 2021-02-08
[onert] Re-organize onert test directory

wateret:
The directory structure in the test dir is outdated. This commit makes
it to have the same structure with src dir.


seanshpark (approved):
LGTM

PR #5988, opened 2021-02-08
[oops] Rename test file

seanshpark:
This will rename test file inside src with oops.test.cpp to follow other
test file naming convention.


mhs4670go (approved):
LGTM

PR #5989, opened 2021-02-09
[onert] Remove BackendContext default impl

wateret:
Remove BackendContext default implementations of interfaces.


seanshpark (approved):
LGTM

PR #5990, opened 2021-02-09
[arser] Revise CMakeLists

seanshpark:
This will revise CMakeLists file; use nncc_coverage option and test to
use arser lib for include folder.

### DIFF
--- a/compiler/arser/CMakeLists.txt
+++ b/compiler/arser/CMakeLists.txt
@@ -4,6 +4,7 @@ add_library(arser INTERFACE)
 # It means that a developer who want to link arser just need to add one line.
 # target_link_library(another-users-target arser) 
 target_include_directories(arser INTERFACE include/)
+target_link_libraries(arser INTERFACE nncc_coverage)
 
 if(NOT ENABLE_TEST)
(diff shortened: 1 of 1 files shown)


seanshpark:
For #5960, from draft #5961

compiler/arser/CMakeLists.txt
```diff
@@ -12,4 +13,4 @@ endif(NOT ENABLE_TEST)
 nnas_find_package(GTest REQUIRED)
 set(TESTS "${CMAKE_CURRENT_SOURCE_DIR}/tests/arser.test.cpp")
 GTest_AddTest(arser_test ${TESTS})
-target_include_directories(arser_test PRIVATE include)
+target_link_libraries(arser_test arser)
```
mhs4670go, line 16: I'm not sure but does it have same meaning?

seanshpark: Yup, `arser` has `target_include_directories(arser INTERFACE include/)` and linking `arser` will include to `include/` folder

mhs4670go (approved):
LGTM

PR #5994, opened 2021-02-09
[loco] Add requires.cmake file

seanshpark:
This will add requires.cmake file for loco.


mhs4670go (approved):
LGTM

PR #5997, opened 2021-02-09
[luci] Introduce test for CircleOptimizerUtils

seanshpark:
This will introduce test for CircleOptimizerUtils methods.


seanshpark:
For #5960

jinevening (approved):
LGTM Thank you

PR #5998, opened 2021-02-09
[luci] Introduce test for ModulePhase

seanshpark:
This will introduce unit test for ModulePhase.


seanshpark:
For #5960

llFreetimell (approved):
LGTM

PR #5999, opened 2021-02-09
[tflchef] Fix recipe_path for message in file driver

seanshpark:
This will fix to use recipe_path instead of using argv that somehow got
missed.


mhs4670go (approved):
LGTM

PR #6000, opened 2021-02-09
[circlechef] Enable coverage to circlechef_core

seanshpark:
This will enable coverage build option to circlechef_core as PRIVATE and
revise others to PUBLIC as it's default.
- set as PRIVATE is not to change coverage of depends as this is not released.


seanshpark:
For #5960

mhs4670go (approved):
LGTM

PR #6001, opened 2021-02-09
[circlechef] Enable and add test for core

seanshpark:
This will enable and add Convert unit test for core.


seanshpark:
For #5960

mhs4670go (approved):
LGTM

PR #6006, opened 2021-02-10
[record-minmax] Test for getMovingAverage

seanshpark:
This will introduce simple unit test for getMovingAverage method.


jinevening (approved):
LGTM

PR #6007, opened 2021-02-10
[one-cmds] Fix one-build to accept option

jinevening:
Fix one-build to properly accept a user-given option, not a default value.

### DIFF
--- a/compiler/one-cmds/one-quantize
+++ b/compiler/one-cmds/one-quantize
@@ -71,16 +71,22 @@ def _get_parser():
         type=str,
         help='record mode (supported: percentile/moving_average, default=percentile)')
 
-    # set default values
-    quantization_group.set_defaults(
-        input_dtype='float32',
-        quantized_dtype='uint8',
-        granularity='layer',
-        min_percentile='1.0',
-        max_percentile='99.0',
-        mode='percentile')
-
     return parser
+
+
+def _set_default_values(args):
+    if not _utils._is_valid_attr(args, 'input_dtype'):
+        setattr(args, 'input_dtype', 'float32')
+    if not _utils._is_valid_attr(args, 'quantized_dtype'):
+        setattr(args, 'quantized_dtype', 'uint8')
+    if not _utils._is_valid_attr(args, 'granularity'):
+        setattr(args, 'granularity', 'layer')
+    if not _utils._is_valid_attr(args, 'mode'):
+        setattr(args, 'mode', 'percentile')
+    if not _utils._is_valid_attr(args, 'min_percentile'):
+        setattr(args, 'min_percentile', '1.0')
+    if not _utils._is_valid_attr(args, 'max_percentile'):
+        setattr(args, 'max_percentile', '99.0')
 
 
 def _verify_arg(parser, args):
@@ -223,6 +229,9 @@ def main():
 
     # parse configuration file
     _utils._parse_cfg(args, 'one-quantize')
+
+    # set default values
+    _set_default_values(args)
 
     # verify arguments
     _verify_arg(parser, args)
--- a/compiler/one-cmds/utils.py
+++ b/compiler/one-cmds/utils.py
@@ -51,7 +51,8 @@ def _parse_cfg(args, driver_name):
                 raise AssertionError('configuration file must have \'' + driver_name +
                                      '\' section')
             for key in config[args.section]:
-                setattr(args, key, config[args.section][key])
+                if not _is_valid_attr(args, key):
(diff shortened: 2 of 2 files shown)


seanshpark:
@jinevening , can you describe the problem you had for this change?

jinevening:
> @jinevening , can you describe the problem you had for this change?

I gave a quantization option (graunularity=channel) to one-build, but one-build did not accept the option but used a default value (graunularity=layer).

The reason was the below code. one-build already has a default option (graunularity=layer), so the below if-statement blocks one-build to accept a user-given argument (it prevents overwriting an option).

```python
                if not _is_valid_attr(args, key):
                    setattr(args, key, config[args.section][key])
```

This PR removes the if-statement.

mhs4670go:
@jinevening Actually, it is for the case when user gives the option to command line and configuration file simultaneously. So, user can override the configuration file easily.

How about making `circle-quantizer` have default value? It looks more natural to me.

jinevening:
> Actually, it is for the case when user gives the option to command line and configuration file simultaneously. So, user can override the configuration file easily.

I understand the intention, but am not sure this code is necessary. Isn't there another way to implement such an interface?

> How about making circle-quantizer have default value? It looks more natural to me.

User option is blocked in one-build, before circle-quantizer is called. I'm not sure changing circle-quantizer can solve the problem. Can you update one-build or circle-quantizer if you have a better solution?

mhs4670go:
@jinevening The reason `one-build` blocks that option is because it already have value(default value). So, remove default value from `one-build` interface and set default value to `circle-quantizer` will solve this problem. Could you test it in local with this solution?

jinevening:
> remove default value from one-build interface and set default value to circle-quantizer will solve this problem

I got the idea. I'll try this solution.

mhs4670go:
@jinevening First, `arser` supports default value whose type is string vector.
https://github.com/Samsung/ONE/blob/c95fb7a9d59f04cc7a51adbaa4cb738d15cee0c4/compiler/arser/tests/arser.test.cpp#L288-L292

The thing is that `CircleQuantizer` is not easy to set default value because it gets three argument in once. 

If there's no direct call user of `CircleQuantizer`, how about changing the interface like `one-quantize`? For example, select one out of three   `quantize_dequantize_weights`, `quantize_with_minmax` and `requantize`. Plus, set `input_dtype`, `quantized_dtype`, and `granularity` individually. Then it would be easy to set its default value.

jinevening:
> CircleQuantizer is not easy to set default value because it gets three argument in once.

Yes, this was the place where I stuck.

> If there's no direct call user of CircleQuantizer

It seems that our internal repo is using circle-quantizer for testing. Changing the interface of circle-quantizer may affect CI. 

Can you explain the below comment more? I cannot find how to give cmd options to one-build in how-to-use-one-commands.txt or help message. 

> it is for the case when user gives the option to command line and configuration file simultaneously.

seanshpark:
> It seems that our internal repo is using circle-quantizer for testing. Changing the interface of circle-quantizer may affect CI.

You can disable while working on this and turn on again later when finished.

mhs4670go:
> You can disable while working on this and turn on again later when finished.

:+1:

> Can you explain the below comment more? I cannot find how to give cmd options to one-build in how-to-use-one-commands.txt or help message.

https://github.com/Samsung/ONE/blob/df0dd41b85ab0fe919801b4a614fe6583cf22d2a/compiler/one-cmds/how-to-use-one-commands.txt#L45-L50

> If you pass the option directly to the command line, the option is processed prior to the configuration file.

jinevening:
> You can disable while working on this and turn on again later when finished.

@seanshpark  This may work, but I'd like to avoid turning off NPU compiler CI if possible. And I think we can make it.

> If you pass the option directly to the command line, the option is processed prior to the configuration file.

@mhs4670go There was a misunderstanding. My question was about one-build, so I thought we can pass arguments of each driver via one-build cmdline interface. But that seems impossible.

I found a way to use default values while accepting user-given arguments. It also does not have to change the interface of one-quantize.  PTAL

mhs4670go:
@jinevening 
> This may work, but I'd like to avoid turning off NPU compiler CI if possible. And I think we can make it.

Is it related with NPU compiler CI? Because it refers to specific ONE commit.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

seanshpark:
> avoid turning off NPU compiler CI if possible

I meant ONE CI :) Changes here has no direct relation with NPU Compiler CI.

PR #6008, opened 2021-02-10
[circle-quantizer] Revise exclusive option check

seanshpark:
This will revise exclusive option check to reduce duplicate codes.


jinevening (approved):
LGTM

PR #6011, opened 2021-02-10
[record-minmax] Revise cmake for test source files

seanshpark:
This will revise CMakeLists to gather test source file list.


jinevening (approved):
LGTM

PR #6015, opened 2021-02-10
[luci] Use must_cast in ResolveCustomOpAddPass

llFreetimell:
Until now, if `input` is `nullptr`, this is not handled.
This commit will change `dynamic_cast` to `loco::must_cast` to catch the error.


seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6022, opened 2021-02-15
[luci] Revise ForwardReshapeToUnaryOpPass

seanshpark:
This will revise ForwardReshapeToUnaryOpPass as to extract
local clone_shape method.

### DIFF
--- a/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
+++ b/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
@@ -30,21 +30,28 @@ luci::CircleReshape *as_reshape(loco::Node *node)
   return dynamic_cast<luci::CircleReshape *>(node);
 }
 
+luci::CircleConst *clone_shape(luci::CircleReshape *reshape)
+{
+  const auto shape = dynamic_cast<luci::CircleConst *>(reshape->shape());
+  // only support CircleConst for now
+  if (shape == nullptr)
+    return nullptr;
+
+  auto dtype = shape->dtype();
+  if (dtype != loco::DataType::S32 && dtype != loco::DataType::S64)
+    return nullptr;
+
+  return luci::clone(shape);
+}
+
 bool forward_reshape(luci::CircleReshape *reshape, luci::CircleNeg *neg)
 {
   assert(reshape != nullptr);
   assert(neg != nullptr);
 
-  luci::CircleConst *cloned_shape = nullptr;
-  const auto reshape_shape = dynamic_cast<luci::CircleConst *>(reshape->shape());
-  // only support CircleConst for now
-  if (reshape_shape == nullptr)
+  luci::CircleConst *cloned_shape = clone_shape(reshape);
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
```diff
@@ -30,21 +30,28 @@ luci::CircleReshape *as_reshape(loco::Node *node)
...
+luci::CircleConst *clone_shape(luci::CircleReshape *reshape)
+{
+  const auto shape = dynamic_cast<luci::CircleConst *>(reshape->shape());
+  // only support CircleConst for now
+  if (shape == nullptr)
+    return nullptr;
+
+  auto dtype = shape->dtype();
+  if (dtype != loco::DataType::S32 && dtype != loco::DataType::S64)
+    return nullptr;
```
mhs4670go, line 42: I think this `dtype` checking should be moved to `import` section.

And, as 
https://github.com/tensorflow/tensorflow/blob/5396b5ce655e4667814cbc1dfc06bb86c1a06396/tensorflow/lite/kernels/reshape.cc#L107-L112, `Reshape` in tflite only supports `int32` dtype, which is already being checked in import step.

BTW, one question comes to me. This issue comes from inconsistency between tensorflow and tensorflow lite support dtype. e.g. `add` input support dtype of tensorflow isn't same with the tensorflow lite one(#5786). Then, which one should be picked? If we blocked the condition where add has `int64` input, we wouldn't be able to run `circle` model having `add` with `int64` even though we have its kernel(not tflite kernel, our kernel).

Of course, it is only the case where circle schema doesn't have any idea of dtype. e.g. `Reshape` should support only `int32` dtype unlike `add` input case(schema doesn't have info about its input and output dtype).
```flatbuffers
table ReshapeOptions {
  new_shape:[int]; // int == int32_t
}
```

seanshpark: > I think this dtype checking should be moved to import section.

OK

> Reshape in tflite only supports int32 dtype,

Just assert in import seems good

> which is already being checked in import step.

I couldn't find this. can you point out the code?

> BTW, one question comes to me.  ...

I don't exactly catch your point. Please post an issue for this if you want to discuss more about this.

seanshpark: Anyway this is just a relocation of existing code. I'll remove `dtype` related codes after this lands.

mhs4670go (approved):
LGTM

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
mhs4670go: @seanshpark 
> I couldn't find this. can you point out the code?

More specific, new shape from option is always created with `S32`.

> I don't exactly catch your point. Please post an issue for this if you want to discuss more about this.

Okay.

PR #6026, opened 2021-02-15
[luci] Revise ForwardReshapeToUnaryOpPass copy_shape

seanshpark:
This will revise ForwardReshapeToUnaryOpPass as to extract local
copy_shape method.


mhs4670go (approved):
LGTM

Issue #6025, opened 2021-02-15
[luci] dtype check in ForwardReshapeToUnaryOpPass


seanshpark:
From https://github.com/Samsung/ONE/pull/6022#pullrequestreview-590184307

----

I think this `dtype` checking should be moved to `import` section.

And, as 
https://github.com/tensorflow/tensorflow/blob/5396b5ce655e4667814cbc1dfc06bb86c1a06396/tensorflow/lite/kernels/reshape.cc#L107-L112, `Reshape` in tflite only supports `int32` dtype, which is already being checked in import step.

BTW, one question comes to me. This issue comes from inconsistency between tensorflow and tensorflow lite support dtype. e.g. `add` input support dtype of tensorflow isn't same with the tensorflow lite one(#5786). Then, which one should be picked? If we blocked the condition where add has `int64` input, we wouldn't be able to run `circle` model having `add` with `int64` even though we have its kernel(not tflite kernel, our kernel).

Of course, it is only the case where circle schema doesn't have any idea of dtype. e.g. `Reshape` should support only `int32` dtype unlike `add` input case(schema doesn't have info about its input and output dtype).
```flatbuffers
table ReshapeOptions {
  new_shape:[int]; // int == int32_t
}
```

seanshpark:
What to do for now, we can;
- remove `dtype` check int `ForwardReshapeToUnaryOpPass`
- add `dtype` check in `Reshape` import only for CircleConst

`only for CircleConst` seems not simple for now...

seanshpark:
done for now

PR #6028, opened 2021-02-15
[luci] Check only S32 for clone_shape

seanshpark:
This will revise ForwardReshapeToUnaryOpPass to check only S32 in
clone_shape method for now and add comments about this code.


seanshpark:
For #6025

mhs4670go (approved):
LGTM

PR #6029, opened 2021-02-15
[luci] numOutputs for CircleCustom

seanshpark:
This will revise CircleCustom Op to provide number of outputs.


seanshpark:
This patch is to revise CircleCustom to look like other multi-output IRs.

seanshpark:
Related: #6030

mhs4670go (approved):
LGTM

@seanshpark 
Just curiosity, `loco::succs()` method isn't enough?

seanshpark:
> Just curiosity, loco::succs() method isn't enough?

Using `loco::succs()` is enough. Just to provide explicit method for number of outputs like other multi-output IRs.

PR #6013, opened 2021-02-10
DRAFT Compiler FE: Provide valid names for CircleNodes

seanshpark:
on-going draft to provide valid names for CircleNodes.

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #5913, opened 2021-02-02
Compiler FE: luci: Assign unique node name


seanshpark:
We need to assign a unique name for each nodes in luci Pass.
Currently there are codes that set new node with same old name:
- in `create_empty_const_from()`
```cxx
  // TODO: We don't have any naming convention for quantized nodes yet.
  //       Fix this when we have one.
  new_node->name(node->name());
```

Let's provide a solution to solve this.

seanshpark:
Main problem
- how to assign name for the new node in Pass?
- if we use unique serial number as a suffix, how to provide this?

seanshpark:
Assert failed with name in export
- [ ] `Generate MatMul_000.opt.circle` with `--O1`
- [ ] `Generate Net_BroadcastTo_AddV2_000.opt.circle` with `--O1`
- [ ] `Generate Net_InstanceNorm_001.opt.circle` with `--O1`
- [ ] TBA

seanshpark:
done for now

PR #6032, opened 2021-02-16
[luci] Name for CircleReshape shape constant

seanshpark:
This will set shape node name of CircleReshape in import.


seanshpark:
For #5913, from draft #6013

mhs4670go (approved):
@seanshpark It seems dummy and shape node is gonna have same name. Will it be unique when export?

mhs4670go:
> It seems dummy and shape node is gonna have same name. Will it be unique when export?

I think it will be. One of my concern is whether `name` being unique isn't needed during compilation(conversion, optimization, etc.) or not except export. Actually, as of now, it seems unnecessary concern though:)

PR #6033, opened 2021-02-16
[luci] Names of nodes in Pass test

seanshpark:
This will set names for nodes in Pass test.


seanshpark:
For #5913, from draft #6013

mhs4670go (approved):
LGTM

mhs4670go:
@seanshpark Setting the name is what we should do even if it isn't used in the test?

seanshpark:
> Setting the name is what we should do even if it isn't used in the test?

It's not used (to export or some kind) but new CircleNode that reference this will access the name and set the new node name with that name. Logically it seems strange if it's empty...
I'm adding `assert(name.length() > 0);` in accessing the name in all Passes to know if the node name is valid or not.

jinevening (approved):
LGTM

PR #6036, opened 2021-02-16
[luci] Names for new nodes in ConvertNCHWToNHWCPass

seanshpark:
This will assign names for new nodes in ConvertNCHWToNHWCPass.

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -72,6 +72,9 @@ luci::CircleTranspose *create_4d_transpose(luci::CircleNode *node,
 {
   assert(indices.size() == 4);
 
+  auto name = node->name();
+  assert(name.length() > 0);
+
   auto perm = node->graph()->nodes()->create<luci::CircleConst>();
   perm->dtype(loco::DataType::S32);
   perm->size<loco::DataType::S32>(4);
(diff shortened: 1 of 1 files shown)


seanshpark:
For #5913, from draft #6013

jinevening:
Assertion failed. Circle node can have no name, but the new assert blocks the case.

seanshpark:
> Assertion failed. Circle node can have no name, but the new assert blocks the case.

There seems name setting PRs should be landed before this...

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -80,9 +83,11 @@ luci::CircleTranspose *create_4d_transpose(luci::CircleNode *node,
   for (uint32_t i = 0; i < 4; i++)
     perm->at<loco::DataType::S32>(i) = indices[i];
   perm->shape_status(luci::ShapeStatus::VALID);
+  perm->name(name + "/Transpose/perm");
```
jinevening, line 86: Two transpose nodes are created from one node. pre-transpose/post-transpose.

Both transpose may have the same name in this implementation. How about adding `indices` at the end?

For example, `perm->name(name + "/Transpose/perm_" + to_string(indices));`.

This will generate different names for pre/post Transposes.

jinevening:
> There seems name setting PRs should be landed before this...

AFAIK, name is an optional field in TFLite. Is it different in circle? If not, the assert should be removed to handle the case where the input node has no name.

seanshpark:
> AFAIK, name is an optional field in TFLite.

Maybe for schema? I'm not sure about tflite interpreter itself.

jinevening:
> Maybe for schema? I'm not sure about tflite interpreter itself.

Sorry. I saw that in TF API, not TFLite. https://www.tensorflow.org/api_docs/python/tf/nn/conv2d. I thought that no name (zero-length string) is acceptable. If it is not, the assert will be ok.

seanshpark:
portion of tflite `interpreter_builder.cc`
```cxx
const char* kEmptyTensorName = "";

  // A little helper to get the names of inputs and outputs. Note that they
  // must outlive the subgraph.
  auto get_name = [](const tflite::Tensor* t) -> const char* {
    auto name = t->name();
    if (name) return name->c_str();
    return kEmptyTensorName;
  };
```

in tflite, name is used (only ?) in debugging and I don't know it can be used for any execution/calculation.

for us, as frontend, we're processing circle data and I prefer to keep name as not empty string.
- practically I don't think name string will be used as execution/calculation
- but for debugging, analysis and profiling, permitting empty string can be a pain in the future.

jinevening (approved):
LGTM

PR #6039, opened 2021-02-16
[luci] Names for new nodes in FuseAddWithTConvPass

seanshpark:
This will assign names for new nodes in FuseAddWithTConvPass,
FuseBatchNormWithConvPass, FuseBatchNormWithDwConvPass and
FuseBatchNormWithTConvPass.

### DIFF
--- a/compiler/luci/pass/src/FuseAddWithTConvPass.cpp
+++ b/compiler/luci/pass/src/FuseAddWithTConvPass.cpp
@@ -90,9 +90,12 @@ bool fuse_add_with_tconv(luci::CircleTransposeConv *tconv)
 
   if (add->fusedActivationFunction() == luci::FusedActFunc::RELU6)
   {
+    auto name = addition->name();
+    assert(name.length() > 0);
     // separate relu op from add op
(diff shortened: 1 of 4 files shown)


seanshpark:
For #5913, from draft #6013

compiler/luci/pass/src/FuseAddWithTConvPass.cpp
```diff
@@ -90,9 +90,12 @@ bool fuse_add_with_tconv(luci::CircleTransposeConv *tconv)
 
   if (add->fusedActivationFunction() == luci::FusedActFunc::RELU6)
   {
+    auto name = addition->name();
+    assert(name.length() > 0);
```
jinevening, line 94: As mentioned https://github.com/Samsung/ONE/pull/6036#issuecomment-779575820 , if name is an optional field as TFlite, we should handle nodes with no name.

seanshpark: I don't think we should follow like tflite. it's our format and our decision.

jinevening (approved):
LGTM

Issue #6030, opened 2021-02-15
[luci] CircleCustom export model tensor looks incorrect


seanshpark:
There's `All_000.circle` in which is test model for custom `All` Op.

This is when dump with `circledump`
```
Operands: T(subgraph index : tensor index) TYPE (shape) (shape_signature) B(buffer index) OperandName
T(0:0) FLOAT32 (2, 4) B(1) ifm
T(0:1) FLOAT32 (2) B(2) ofm
T(0:2) INT32 () B(3) All/reduction_indices

Operators: O(subgraph index : operator index) OpCodeName
    Option(values) ... <-- depending on OpCode
    I T(tensor index) OperandName <-- as input
    O T(tensor index) OperandName <-- as output
O(0:0) CUSTOM(All)
    Tidx(2) keep_dims(false)
    I T(0:0) ifm
    I T(0:2) All/reduction_indices
    O T(0:1) ofm

Inputs/Outputs: I(input)/O(output) T(tensor index) OperandName
I T(0:0) ifm
O T(0:1) ofm
```

There's output of `circle2circle`, `All_000.opt.circle` and this looks different in Tensor
```
Operands: T(subgraph index : tensor index) TYPE (shape) (shape_signature) B(buffer index) OperandName
T(0:0) FLOAT32 (2, 4) B(0) ifm
T(0:1) INT32 () B(1) All/reduction_indices
T(0:2) FLOAT32 () B(0) ofm
T(0:3) FLOAT32 (2) B(0) ofm

Operators: O(subgraph index : operator index) OpCodeName
    Option(values) ... <-- depending on OpCode
    I T(tensor index) OperandName <-- as input
    O T(tensor index) OperandName <-- as output
O(0:0) CUSTOM(All)
    Tidx(2) keep_dims(false)
    I T(0:0) ifm
    I T(0:1) All/reduction_indices
    O T(0:3) ofm

Inputs/Outputs: I(input)/O(output) T(tensor index) OperandName
I T(0:0) ifm
O T(0:3) ofm
```

`T(0:2) FLOAT32 () B(0) ofm` seems incorrect output

seanshpark:
This was found while working with #6013

Seems there is no handling of `CircleCustom` in `MultiOutputDetector` class of `compiler/luci/export/src/CircleTensorExporter.cpp`

seanshpark:
Same for
- `CircleNonMaxSuppressionV4`
- `CircleNonMaxSuppressionV5`
[...]

PR #6043, opened 2021-02-16
[luci] Revise MultiOutputDetector in tensor export

seanshpark:
This will revise MultiOutputDetector in tensor exporter for types of
multiple output nodes.


seanshpark:
For #6030

mhs4670go (approved):
LGTM

PR #6045, opened 2021-02-16
[one-cmds] Fix onnx version for prepare venv

seanshpark:
This will fix to use explicit onnx version for prepare venv.


seanshpark:
For master

mhs4670go (approved):
LGTM

PR #6046, opened 2021-02-16
[luci] Assert in export_node for CircleCustom

seanshpark:
This will add assertion check for numOutputs in CircleCustom export.


mhs4670go (approved):
LGTM

seanshpark:
For #5913, from draft #6013

PR #6049, opened 2021-02-16
[onert/lib] Introduce tflite input initializer

hseok-oh:
This commit introduces tflite random input initizlizer and copy input initializer

---

Part of #5971

### DIFF
--- a/runtime/libs/tflite/include/tflite/RandomInputInitializer.h
+++ b/runtime/libs/tflite/include/tflite/RandomInputInitializer.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


runtime/libs/tflite/include/tflite/RandomInputInitializer.h
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
```
seanshpark, line 2: 2021 :)

seanshpark (approved):
LGTM

runtime/libs/tflite/src/RandomTestRunner.cpp
```diff
@@ -36,6 +37,60 @@ namespace tflite
 
 using namespace std::placeholders;
 
+class CopyInputInitializer
```
glistening, line 40: @hseok-oh (optional) What about getting `CopyInputInitializer` out of `RandomTestRunner.cpp` like `RandomInputInitializer`? It can have its own module (`CopyInputInitializer.[h,cpp]`) or could be merged into `RandomInputInitializer` (in this case, the name would be good to be renamed (`RandomInputInitializer` →`InputInitializer`).

glistening (approved):
LGTM

runtime/libs/tflite/src/RandomTestRunner.cpp
hseok-oh: @glistening Different with draft, I didn't move `CopyInitializer` out of `RandomTestRunner.cpp` because it is used in `RandomTestRunner.cpp` only. If it is used by other module, we can move this.

PR #6056, opened 2021-02-17
[foder] Add guard

seanshpark:
This will add header guard for FileLoader.


seanshpark:
From draft #6054

mhs4670go (approved):
LGTM

PR #6057, opened 2021-02-17
[luci] Revise tester cmake for unit testing

seanshpark:
This will revise tester cmake for unit testing to use variable for
tester source files.


seanshpark:
From draft #6054

mhs4670go (approved):
LGTM

Issue #6058, opened 2021-02-17
Compiler FE: Remove test codes from metric score


seanshpark:
Out DuplicateCode analysis (SAM) result shows lots of `*.test.cpp` codes are included in `luci-interpreter`
I think we don't have to include these test files as analysis target.

seanshpark:
@cgbahk , is it possible to do this?
If there is no file filter then we can think of moving the files into `test` folder.

cgbahk:
I guess luci-interpreter may not configured to exclude test by setting
```
-DENABLE_TEST=OFF
```

seanshpark:
@cgbahk thanks!

it turns our that `compiler/luci-interpreter/src/kernels/CMakeLists.txt` doesn't check `ENABLE_TEST` option :)

seanshpark:
This came from #6053

seanshpark:
Done

PR #6059, opened 2021-02-17
[luci-interpreter] Revise test build

seanshpark:
This will revise test to build with ENABLE_TEST is set.


seanshpark:
For #6058

jinevening (approved):
LGTM

Issue #6053, opened 2021-02-17
Compiler FE: Reduce duplicate codes


seanshpark:
Let's reduce some ducplicate codes in our compiler frontend...

seanshpark:
duplicate score: 3.19 --> 3.36

seanshpark:
done for now

seanshpark:
As of now, score is 3.43

PR #6061, opened 2021-02-17
[luci] Validate for number of inputs/outputs

seanshpark:
This will introduce validate method to check number of inputs and single
output.

### DIFF
--- a/compiler/luci/import/include/luci/Import/GraphBuilder.h
+++ b/compiler/luci/import/include/luci/Import/GraphBuilder.h
@@ -33,7 +33,7 @@ class GraphBuilder : public GraphBuilderBase
 public:
   virtual ~GraphBuilder() = default;
 
-  // common validate to check number of inputs and single output
+  // common validate method to check number of inputs and single output
   bool validate(const ValidateArgs &args, size_t i) const
   {
     return (args.op.inputs.size() == i && args.op.outputs.size() == 1);
--- a/compiler/luci/import/include/luci/Import/GraphBuilder.h
+++ b/compiler/luci/import/include/luci/Import/GraphBuilder.h
@@ -34,9 +34,9 @@ public:
   virtual ~GraphBuilder() = default;
 
   // common validate method to check number of inputs and single output
-  bool validate(const ValidateArgs &args, size_t i) const
+  bool validate(const ValidateArgs &args, size_t input_cnt) const
   {
-    return (args.op.inputs.size() == i && args.op.outputs.size() == 1);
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6053, from draft #6054

compiler/luci/import/include/luci/Import/GraphBuilder.h
```diff
@@ -33,6 +33,12 @@ class GraphBuilder : public GraphBuilderBase
 public:
   virtual ~GraphBuilder() = default;
 
+  // common validate to check number of inputs and single output
+  bool validate(const ValidateArgs &args, size_t i) const
```
llFreetimell, line 37: How about `validate_input_cnt` or `validate_common_inout` or `is_single_output_with` or something else?
`validate` seems generous.

seanshpark: Yes, it seems general... 
Behind this is that I like short names and `size_t i` argument will distinguish existing method...

compiler/luci/import/include/luci/Import/GraphBuilder.h
```diff
@@ -33,6 +33,12 @@ class GraphBuilder : public GraphBuilderBase
 public:
   virtual ~GraphBuilder() = default;
 
+  // common validate to check number of inputs and single output
```
llFreetimell, line 36: Maybe :D ?
```suggestion
  // common validation to check number of inputs and single output
```

compiler/luci/import/include/luci/Import/GraphBuilder.h
llFreetimell: I see it :)
Then `bool validate(const ValidateArgs &args, size_t input_cnt) const` seems okay?
I like short names too but if both function name and parameter name iare too short, sometimes it is confused :(
If you don't like this too, let's go as it is :)

compiler/luci/import/include/luci/Import/GraphBuilder.h
seanshpark: Ah, `validate` is the method `validate`... I'll revise this...

compiler/luci/import/include/luci/Import/GraphBuilder.h
seanshpark: `input_cnt` seems OK :)

llFreetimell (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #6062, opened 2021-02-17
[Compiler][luci/Service] Refactoring rule tests


llFreetimell:
Originated from https://github.com/Samsung/ONE/pull/5771#discussion_r562509788, #5809
Draft : #5811

## What?

- Remove dependency with `loco`
- Remove dependency with `luci/Pass`
- Move each tests in `CircleShapeInferenceRule.test.cpp` and `CircleTypeInferenceRule.test.cpp` to `Service/Nodes/Circle*.test.cpp`

## Why?

- Tests of `CircleShapeInferenceRule.test.cpp`/`CircleTypeInferenceRule.test.cpp` and  `ShapeInferencePass`/`TypeInferencePass`are using each other. (Sort of cyclic dependency)

## Todo

- [x] Introduce refactored tests for following ops
  - [x] Add #6068
  - [x] AveragePool2D #6069
  - [x] ExpandDims #6070
  - [x] GatherNd #6071
  - [x] Relu #6063
  - [x] ResizeBilinear #6072
  - [x] ResizeNearestNeighbor #6073
  - [x] Squeeze #6074
  - [x] Transpose #6075
- [x] Remove current tests #6098

llFreetimell:
Done!

PR #5811, opened 2021-01-25
[Draft][luci] Integration of inference rule tests

llFreetimell:
Parent Issue : ~~#5809~~ #6062

This draft is for integrating inference rule tests to other tests.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

PR #6063, opened 2021-02-17
[luci] Introduce refactored tests for Relu in luci/service 

llFreetimell:
Parent Issue : #6062
Draft : #5811

This commit will introduce new tests for `CircleRelu` in `luci/Service`.
The new tests are refactored with current tests in `Circle*Rule.test.cpp`.

### DIFF
--- a/compiler/luci/service/src/Nodes/CircleRelu.test.cpp
+++ b/compiler/luci/service/src/Nodes/CircleRelu.test.cpp
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/Service/CircleShapeInference.h>
+#include <luci/Service/CircleTypeInference.h>
+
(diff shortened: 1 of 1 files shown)


compiler/luci/service/src/Nodes/CircleRelu.test.cpp
```diff
@@ -0,0 +1,58 @@
+/*
```
mhs4670go, line 1: @llFreetimell I think this change is a little bit weird. For now, the file `service/src/Nodes/${NODE_NAME}.cpp` implements "service" function that may be use in passes or sth. And `test.cpp` file tests its `service` function. So, this contents should be moved to `CircleTypeInference.test.cpp` or sth because it tests `luci::tinf::Rule` and so should `luci::sinf::Rule`.

llFreetimell: I am planning to restructure `luci/Service` as following structure.
```
- service/src
  - Nodes
    - CircleRelu.cpp : implements shape/type inference rule for Relu
    - CircleRelu.test.cpp : includes shape/type inference rule tests for Relu
    - ...
  - CircleShapeInference.cpp
  - CircleTypeInference.cpp
  - ...
```
And `CircleShapeInference.cpp` will include only `infer` function like https://github.com/Samsung/ONE/blob/f026107f9c182acb26d25ba0f6dbf4fadce41119/compiler/luci/service/src/CircleShapeInference.cpp (So does `CircleTypeInference.cpp`).

As a result, for me, implementing those tests in `CircleTypeInference.test.cpp` or `CircleShapeInference.test.cpp` looks weird because rules are collected at one file but tests are separated :(

seanshpark:
> Refactoring Relu tests ...

About the title...  I think `Relocate` would be better as this only changes the file location.

seanshpark:
> Refactoring Relu tests ...

As there are multiple changes at once, it was a bit hard to catch the changes.
Actually a new rule test codes are introduce and existing corresponding test code is removed.
So actually reading the change itself seems meaningless...

I think to make things simple, you can split the changes in (1) introduce rule test (2) remove service test.

llFreetimell:
> you can split the changes in (1) introduce rule test (2) remove service test.

I will follow this suggestion, thanks :)

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #6065, opened 2021-02-17
[luci] Revise Tester to use ReadModule

seanshpark:
This will revise ReadTester and WriteTester to use ReadModule.


seanshpark:
For #6053, from draft #6054

mhs4670go (approved):
LGTM

PR #6066, opened 2021-02-17
[luci] Use template for has_same_values

seanshpark:
This will introduce template method has_same_elements() and revise
has_same_values method to use this.


seanshpark:
As we have quantized `U8` models, in next PR, I will add more types.

llFreetimell (approved):
LGTM

PR #6075, opened 2021-02-17
[luci] Introduce refactored tests for Transpose in luci/service

llFreetimell:
Parent Issue : #6062
Draft : #5811

This commit will introduce new tests for `CircleTranspose` in `luci/Service`.
The new tests are refactored with current tests in `Circle*Rule.test.cpp`.

### DIFF
--- a/compiler/luci/service/src/Nodes/CircleTranspose.test.cpp
+++ b/compiler/luci/service/src/Nodes/CircleTranspose.test.cpp
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


mhs4670go:
IMHO, 
1. If type inference related codes that are very similar to shape one are is gong to come in, using fixture(`TEST_F`) would be help to reduce duplicate codes, which makes static analysis score get high.
2. It would be better to implement one positive, one negative test each at once, which is tiresome though:< But it is for the score as well.

mhs4670go:
@seanshpark AFAIK, circle nodes should have its name. Should it be applied to shape and type inference as well? If so, these tests should be changed to set names.

llFreetimell:
> If type inference related codes that are very similar to shape one are is gong to come in, using fixture(TEST_F) would be help to reduce duplicate codes, which makes static analysis score get high.

Yes, this is next step after all the current tests are refactored :)

> It would be better to implement one positive, one negative test each at once, which is tiresome though:< But it is for the score as well.

Good point :) 
However, as most of rules do not throw, I thought that time is needed to think what can be negative test for those rules.
As a result, I pointed to copy&refactor current tests but I agree that adding negative test is good.

> circle nodes should have its name. Should it be applied to shape and type inference as well?

IMHO, as this is small unit tests and only uses inference rule, it can be ignored but it's just my opinion..!

mhs4670go:
@llFreetimell As of now, all Passes checks if op has name or not using assert. If that policy applies to here, `assert` can't be ignored:<

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

seanshpark:
> AFAIK, Should it be applied to shape and type inference as well?

I had a situation: `new created nodes should somehow inherit the name from the node of creation` for the Passes.

For shape and dtype inference Pass,
- shape and dtype inference does not cause new node creation
- shape and dtype inference does not access name

So, logically not giving any name to the test nodes seems reasonable...

PR #6079, opened 2021-02-18
[Draft] Origin annotation and data structure

llFreetimell:
Parent Issue : #6080

This draft is for introducing origin annotation and related data structures.

### DIFF
--- a/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
+++ b/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_PROFILE_CIRCLE_NODE_ORIGIN_H__
+#define __LUCI_PROFILE_CIRCLE_NODE_ORIGIN_H__
+
+#include "CircleNodeID.h"
+
+#include <luci/IR/CircleNode.h>
+
+#include <set>
+
+namespace luci
+{
+
+class CircleNodeOrigin
+{
+public:
+  ~CircleNodeOrigin() = default;
+
+public:
+  struct Source
+  {
+  public:
+    std::string name(void) const { return _name; }
+    void name(const std::string &name) { _name = name; }
+
+    uint32_t id(void) const { return _id; }
+    void id(const uint32_t id) { _id = id; }
+
+  private:
+    std::string _name;
+    uint32_t _id = 0;
+  };
+
+public:
+  virtual std::set<const Source *> sources(void) const = 0;
+};
+
+std::shared_ptr<CircleNodeOrigin> single_origin(uint32_t id, const std::string name);
+
+std::shared_ptr<CircleNodeOrigin>
+composite_origin(std::initializer_list<std::shared_ptr<CircleNodeOrigin>> origins);
+
+std::shared_ptr<CircleNodeOrigin>
+composite_origin(std::vector<std::shared_ptr<CircleNodeOrigin>> origins);
+
+} // namespace luci
+
+namespace luci
+{
+
+bool has_origin(const luci::CircleNode *circle_node);
+
+void add_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin);
+
+const std::shared_ptr<luci::CircleNodeOrigin> get_origin(const luci::CircleNode *circle_node);
+
+} // namespace luci
+
+#endif // __LUCI_PROFILE_CIRCLE_NODE_ORIGIN_H__
--- a/compiler/luci/profile/src/CircleNodeOrigin.cpp
+++ b/compiler/luci/profile/src/CircleNodeOrigin.cpp
@@ -1,0 +1,165 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Profile/CircleNodeOrigin.h"
+
+#include <loco.h>
+
+#include <cassert>
+#include <vector>
+
+namespace
+{
+
+/**
+ * @brief Set annotation for recording origin information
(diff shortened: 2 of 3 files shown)


compiler/luci/profile/include/luci/Profile/CircleOriginData.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef __LUCI_PROFILE_CIRCLE_ORIGIN_DATA_H__
+#define __LUCI_PROFILE_CIRCLE_ORIGIN_DATA_H__
+
+#include <string>
+
+namespace luci
+{
+
+class CircleOriginData
```
seanshpark, line 25: Can we rename this to `CircleOrigin` ?

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
```diff
@@ -0,0 +1,39 @@
...
+namespace luci
+{
+
+using CircleNodeOrigin = std::vector<CircleOriginData>;
+
+bool has_node_origin(const luci::CircleNode *circle_node);
+
+void add_node_origin(luci::CircleNode *circle_node, const CircleOriginData &origin);
+
+const CircleNodeOrigin &get_node_origin(const luci::CircleNode *circle_node);
```
seanshpark, line 35: (just thougts)  we are handling for `CircleNode` and there is an argument `luci::CircleNode *`
so for shorter name, we may use `has_origin`, `add_origin`, `get_origin`.
and as `get_origin` returns a vector of `CircleOriginData`, we may use `get_origins` to denote it returns multiple of origins.

If we use `CircleOrigin` instead of `CircleOriginData` then we may use `CircleOrigins = std::vector<CircleOrigin>;`

compiler/luci/profile/include/luci/Profile/CircleOriginData.h
llFreetimell: At first, I thought `CircleOriginInfo`, `CircleOriginData`, `CircleOrigin` and so on.
But I excluded `CircleOrigin` because the word is very similar with `CircleNodeOrigin` and thus it causes confusion :(

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
llFreetimell: - `has_origin`, `add_origin`, `get_origin` : LGTM
- `has_origins`, `add_origins`, `get_origins` : LGTM. But for this, renaming file to `CircleNodeOrigins` may be needed.

seanshpark: `has_origin` vs `has_origins` :  seems similar, to ask if there is any origin or origins
`add_origin` vs `add_origins` : `add_origin` will add one item. `add_origins` will add multiple items
`get_origin` vs `get_origins` : `get_origin(luci::CircleNode*, int index)` would make sense but not sure this is needed. so for the purpose of origin, we may only need `get_origins`

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
```diff
@@ -0,0 +1,71 @@
...
+    uint32_t id(void) const { return _id; }
+    void id(const uint32_t &id) { _id = id; }
+
+  private:
+    std::string _name;
+    uint32_t _id = 0;
+  };
+
+public:
+  virtual std::set<const OriginSeed *> seeds(void) const = 0;
```
llFreetimell, line 50: There are more candidates...

```cpp
virtual std::set<const OriginSeed *> seeds(void) const = 0;

virtual std::set<const AbstractOriginNode *> nodes(void) const = 0;

virtual std::set<const OriginBlock *> blocks(void) const = 0;
```

seanshpark:
About `OriginSeed`, can you provide some background about why you used the word `seed` ?

llFreetimell:
> About OriginSeed, can you provide some background about why you used the word seed ?

The word `OriginSeed` is not fixed yet :(
Nevertheless, I uploaded the codes to share overall design of `CircleNodeOrigin`.

I am still thinking what is the most proper word for representing it instead of seed...
For example, attribute, property, element, item, block, vertex and so on...
Naming is really difficult....

seanshpark:
> For example, attribute, property, element, item, block, vertex and so on...

I mean, my question is, what is your intention of using the word `seed`?
As you desctibed, you have difficulty choosing word so, if you describe the intention would help find the proper word...

llFreetimell:
I intended to represent **core origin information** of `CircleNode`.

Let's suppose that `nodeA` has following `CircleNodeOrigin`.
```
+----------------------------------------------------------------------+
| CompositeOrigin                                                      |
|                                                                      |
| +------------------+  +--------------------------------------------+ |
| | SingleOrigin     |  | CompositeOrigin                            | |
| |                  |  |                                            | |
| | +--------------+ |  | +------------------+  +------------------+ | |
| | | Seed         | |  | | SingleOrigin     |  | SingleOrigin     | | |
| | |              | |  | |                  |  |                  | | |
| | | id = 0       | |  | | +--------------+ |  | +--------------+ | | |
| | | name = node0 | |  | | | Seed         | |  | | Seed         | | | |
| | +--------------+ |  | | |              | |  | |              | | | |
| +------------------+  | | | id = 1       | |  | | id = 2       | | | |
|                       | | | name = node1 | |  | | name = node2 | | | |
|                       | | +--------------+ |  | +--------------+ | | |
|                       | +------------------+  +------------------+ | |
|                       +--------------------------------------------+ |
+----------------------------------------------------------------------+              
```

Then, following results is what I intended.

```cpp
auto seeds = get_origin(nodeA)->seeds();
for (auto seed : seeds)
{
    std::cout << "Origin ID   : " << seed->id() << std::endl;
    std::cout << "Origin Name : " << seed->name() << std::endl;
}

// Origin ID : 0
// Origin Name : node0
// Origin ID : 1
// Origin Name : node1
// Origin ID : 2
// Origin Name : node2
```

seanshpark:
> // Origin ID   : 0
// Origin Name : node0
// Origin ID   : 1
// Origin Name : node1
// Origin ID   : 2
// Origin Name : node2

So you mean by origin, `nodeA` was added(created) from source of `node0`, `node1` and `node2` ?
That is, where the `nodeA` came from... `originated from`
So for me, instead of using word `seed`, I can think of simple word `from` or `source`
if the word `source` be too generic, we can use `TransformSource` or `TransSource`, `TransSrc`, `TrSrc` as `nodeA` was created from some graph transformations.

The word `origin` itself looks good to me, as it's `ORIGIN`...
```cxx
auto list = get_origin(nodeA)->list();
```
seems reasonable.

llFreetimell:
>  I can think of simple word from or source

`source` seems nice for me :D
Is it okay to go with `sources` and `source` like following example?

```cpp
auto sources = get_origin(nodeA)->sources();
for (auto source : sources)
{
    std::cout << "Origin ID   : " << source->id() << std::endl;
    std::cout << "Origin Name : " << source->name() << std::endl;
}
```

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -0,0 +1,162 @@
...
+private:
+  std::vector<std::shared_ptr<CircleNodeOrigin>> _origins;
+};
+
+} // namespace
+
+namespace luci
+{
+
+std::shared_ptr<CircleNodeOrigin> single_origin(uint32_t id, std::string name)
```
llFreetimell, line 121: @mhs4670go This is changed. Does this affect your draft?

seanshpark: @llFreetimell , can we use like `single_origin(uint32_t id, const std::string &name)` ?
`

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+namespace
+{
+
+class SingleOrigin final : public luci::CircleNodeOrigin
+{
+public:
+  SingleOrigin() = delete;
+
+  SingleOrigin(uint32_t id, std::string name)
```
seanshpark, line 60: can we use like `SingleOrigin(uint32_t id, const std::string &name)` ?

llFreetimell:
Related PR is created, close this!

PR #6081, opened 2021-02-18
[luci] Revise QuantizeDequantizeWeightsPass

seanshpark:
This will revise QuantizeDequantizeWeightsPass as extract
iterate_per_channel method for common multi level iteration loop to
reduce duplicate codes.

### DIFF
--- a/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
+++ b/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
@@ -285,14 +285,14 @@ void asymmetric_wdequant_per_channel(CircleConst *node, std::vector<float> &scal
   uint32_t size = node->size<loco::DataType::U8>();
   std::vector<float> dequantized_values(size);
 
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6053, from #6054. There will be continued PR to reduce similar patterns in this file.

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -246,35 +285,14 @@ void asymmetric_wdequant_per_channel(CircleConst *node, std::vector<float> &scal
   uint32_t size = node->size<loco::DataType::U8>();
   std::vector<float> dequantized_values(size);
 
-  loco::TensorShape dimension;
-  dimension.rank(4);
-  uint32_t indices[4] = {
-    0,
+  auto func = [&](uint32_t *indices, loco::TensorShape &dimension, int channel_dim_index) {
```
jinevening, line 288: (optional) `dequantize` would be a better name than `func`.

jinevening (dismissed):

jinevening (approved):
LGTM Thank you

PR #6084, opened 2021-02-19
[luci] Revise import validate for DE-Ops

seanshpark:
This will revise import validate with input count with single output for
Ops starting with D,E.


seanshpark:
For #6053, from #6054.

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

PR #6096, opened 2021-02-19
[luci] Introduce iterate_per_channel for QuantizeWithMinMax

seanshpark:
This will introduce iterate_per_channel to iterate loop to reduce
duplicate codes in QuantizeWithMinMaxPass.


seanshpark:
For #6053, from #6054,

jinevening (approved):
LGTM

PR #6097, opened 2021-02-19
[luci] revise get_channel_dim_index arg type

seanshpark:
This will revise get_channel_dim_index arg type with int32_t for
specific length.


seanshpark:
From #6054,

jinevening (approved):
LGTM

PR #6099, opened 2021-02-19
[luci] Relocate Relu check

seanshpark:
This will revise to check Add activation is Relu prior to reduce
duplicate codes: is_batchnorm_add method.


seanshpark:
For #6053

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #6080, opened 2021-02-18
[Compiler] Propagation of profiling data


llFreetimell:
## What?

Let's propagate profiling data **from circle to circle**!

## How?

![image](https://user-images.githubusercontent.com/24720192/108467319-5b8e2500-72c8-11eb-80ef-cded7728a0bb.png)

- `metadata:[Metadata];` will be used for storing the origin data in circle file.
- `loco::Annotation` will be used for storing the origin data at `CircleNode`.

## Related Issue/PR

- Draft
  - Origin data structure : #6079 (@llFreetimell)
  - Infra for import/export : #6101 (@llFreetimell)
  - Propagating origin data : #6217 (@mhs4670go)
  - Interface to disable/enable profiling in one-cmd : #6186 (@mhs4670go)
- Issue
  - No related issue for now

/cc @cgbahk

llFreetimell:
## Discussion Items

- What would be proper for identification of **origin** file?
- Someone may not want to include profiling data. Then how can we handle it?
  - Introducing option not to include profiling data is one of the solution. But we still need an answer for good UX.
- What is the intermediate/final results? What should be provided for users?

seanshpark:
> ... of origin file?

whay may be the problems embedding profile data inside `circle` file?

seanshpark:
> But we still need an answer for good UX.

Can you describe more about what `UX` you mentioned?

cgbahk:
Diagram looks a bit strange on what is Node C. I guess you tried to say Node C is fused op for Node A and B, but that was not clear?

cgbahk:
We may also consider the case when input model has no origin file identifier (e.g. tflite). In that case I guess the compiler would set origin file identifier as the input file.

llFreetimell:
@seanshpark 

> whay may be the problems embedding profile data inside circle file?

No problem and that is the plan :) The question is about another thing.
If user ask us "What is the origin file of the profiling data?", then we wanna say "Look at the circle file. The answer is in there.".
Then what should be in the circle file? This is the question what I intended :)

> Can you describe more about what UX you mentioned?

We should give an option such as `--no_profiling_data` not to include profiling data in circle file.
But how can we provide this function to users and make it's usage convenient?
This was what I wanted to say :)

llFreetimell:
@cgbahk 

> Diagram looks a bit strange on what is Node C. I guess you tried to say Node C is fused op for Node A and B, but that was not clear?

[...]

PR #6101, opened 2021-02-19
[Draft] Infra for propagating profiling data

llFreetimell:
Parent Issue : #6080

**NOTE** This draft is based on #6079. If there is something to suggest about `CircleNodeOrigin`(name not fixed), please suggest it at #6079.
**NOTE** There are some `TODO` and these will be implemented soon.

---

This draft is for introducing a new infra to propagate profiling data.


llFreetimell:
### Sorry for reviewrs

Most of variable/function names are not good and they are really confused :(
Those will be fixed soon... sorry for that in advance...
(I wanted to share overall direction of codes. But naming variable/function took much time for me...)

compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -1492,6 +1493,16 @@ void exportNode(loco::Node *node, flatbuffers::FlatBufferBuilder &builder, Seria
     ExportContext ctx{builder, md, gd};
     OperationExporter exporter{ctx};
     circle_node->accept(&exporter);
+    if (has_origins(circle_node))
+    {
+      assert(gd._operators.size() > 0);
+      const auto node_id = gd._operators.size() - 1;
```
mhs4670go, line 1499: `_operators` isn't always increase(virtual node case). It might need to check its size before and after the visitor.

compiler/luci/export/src/CircleExporterImpl.cpp
```diff
@@ -222,19 +222,42 @@ void CircleExporterImpl::exportModule(Module *module)
...
+      CreateMetadata(_builder, _builder.CreateString("origin_mapping"), origin_map_buffer_id);
+    metadata_vec.push_back(origin_map_metadata);
+  }
+  auto metadata = _builder.CreateVector(std::vector<Offset<Metadata>>(metadata_vec));
 
   // This version is taken from comment in fbs
   constexpr uint32_t version = 0;
 
+  // create array of buffers
+  auto buffers = _builder.CreateVector(md._buffers);
```
seanshpark, line 256: metadata and buffers  creation order has changed. is there any particular reason for this?

compiler/luci/export/src/CircleExporterImpl.cpp
```diff
@@ -222,19 +222,42 @@ void CircleExporterImpl::exportModule(Module *module)
...
+
   // Model
   auto model_offset = CreateModel(_builder, version, operator_codes, subgraphs, description,
-                                  buffers, metadata_buffer);
+                                  buffers, 0 /* metadata_buffer */, metadata);
```
seanshpark, line 260: You set `metadata_buffer ` to 0 and added `metadata`. Could you explain about this? 
I'm asking because I'm not good at this part...

compiler/luci/export/src/CircleExporterImpl.cpp
mhs4670go: @seanshpark I guess it is to add `Metadata` to buffers.
```flatbuffers
table Metadata {
  // A human readable string to uniquely identify a Metadata.
  name:string;
  // An index to the buffers table.
  buffer:uint;
}
```
As you can see, `Metadata` stores its data to buffer.

compiler/luci/export/src/CircleExporterImpl.cpp
mhs4670go: @seanshpark 
```flatbuffers
...
// Deprecated, prefer to use metadata field.
  metadata_buffer:[int];
```
`metadata_buffer` seems deprecated. This is why he did the way he does.

compiler/luci/export/src/CircleExporterImpl.cpp
seanshpark: @mhs4670go , isn't this `buffers` for Tensors? below line `buffers, metadata_buffer`  shows they are distinct items.
or maybe I understand incorrectly...

mhs4670go: https://github.com/tensorflow/tensorflow/issues/30387#issuecomment-508681200
> We don't have standardized methods to read from or write to it.

```flatbuffers
table Metadata {
  // A human readable string to uniquely identify a Metadata.
  name:string;
  // An index to the buffers table.
  buffer:uint;
}
```

I couldn't find any document about where the buffer would be used except comments in schema. As comment said, `Metadata` has `buffer` field that points to buffers table.

seanshpark: From `schema.fbs`

```
  // Buffers of the model.
  // Note the 0th entry of this array must be an empty buffer (sentinel).
  // This is a convention so that tensors without a buffer can provide 0 as
  // their buffer.
  buffers:[Buffer];

  // Metadata about the model.  Indirects into the existings buffers list.
  metadata_buffer:[int];
```

mhs4670go:
@llFreetimell It seems you haven't implemented the function yet that enables users to choose whether profile information would be saved.

compiler/luci/export/src/CircleExporterImpl.cpp
llFreetimell: > metadata and buffers creation order has changed. is there any particular reason for this?

This is because metadata should be included in the buffer too.

>  isn't this buffers for Tensors? below line buffers, metadata_buffer shows they are distinct items.
or maybe I understand incorrectly...

Each items are distinct, but there is only one buffer table.
In `circle_schema.fbs`, following items is included.
```
table Model {
  ...

  // Buffers of the model.
  // Note the 0th entry of this array must be an empty buffer (sentinel).
  // This is a convention so that tensors without a buffer can provide 0 as
  // their buffer.
  buffers:[Buffer];

  // Metadata about the model. Indirects into the existings buffers list.
  // Deprecated, prefer to use metadata field.
  metadata_buffer:[int];

  // Metadata about the model.
  metadata:[Metadata];
}

root_type Model;
```
`Model` has only one buffer table(`buffers`) and all the buffer indices in the `Model` should point out one of the indices.
Following example might be helpful.
```
Model
{
  buffers
  {
    Buffer 0 : Empty
    Buffer 1 : Data for Tensor1
    Buffer 2 : Data for Tensor2
    Buffer 3 : Data for metadata1
    Buffer 4 : Data for metadata2
  }

  metadata
  {
    {
      name : "Metadata1"
      buffer : 3
    }
    {
      name : "Metadata2"
      buffer : 4
    }
  }

  subgraphs
  {
    tensors
    {
      Tensor 0 : Do not use buffer
      Tensor 1 : Use buffer 1
      Tensor 2 : Use buffer 2
    }
  }
}
```

llFreetimell:
@mhs4670go 

> It seems you haven't implemented the function yet that enables users to choose whether profile information would be saved.

Yes this is not yet implemented because creating prototype of infra has high priority than user's option :)
That should be implemented too.

compiler/luci/export/src/ExportMetadata.cpp
```diff
@@ -0,0 +1,108 @@
...
+  std::vector<uint8_t> data;
+
+  const auto source_table = metadata.source_table();
+  for (auto &kv : source_table)
+  {
+    const auto id = kv.first;
+    data.emplace_back(0xFF & (id >> 0 * 8));
+    data.emplace_back(0xFF & (id >> 1 * 8));
+    data.emplace_back(0xFF & (id >> 2 * 8));
+    data.emplace_back(0xFF & (id >> 3 * 8));
```
llFreetimell, line 58: (Originally written by @seanshpark )

maybe consider something like this...
```cxx
union {
  uint32_t u32;
  uint8_t u8[4];
} u328_t;

u328_t u328;
u328.u32 = id;
for (...) data.emplace_back(u328.u8[i])
```

seanshpark: if `data.emplace_back(u328.u8)` is possible then this is better :)

llFreetimell: I thunk your intention is to handle these similar codes effectively.
If it is right, I agree with the intention and thus I am thinking to introduce following function.
```cpp
void append(std::vector<uint8_t> &to, uint32_t value)
{
  to.emplace_back(0xFF & (value >> 0 * 8));
  to.emplace_back(0xFF & (value >> 1 * 8));
  to.emplace_back(0xFF & (value >> 2 * 8));
  to.emplace_back(0xFF & (value >> 3 * 8));
}
```

compiler/luci/import/src/ImportMetadata.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    assert(meta.buffer < reader.buffers().size());
+    const std::vector<uint8_t> &buffer = reader.buffers()[meta.buffer]->data;
+
+    // TODO Handle this by enum
+    if (meta.name.compare("op_table") == 0)
+    {
+      node_source_ids_map = metadata_with_op_table(buffer);
+    }
+    else if (meta.name.compare("source_table") == 0)
```
llFreetimell, line 133: (Originally written by @seanshpark)

are these to `op_table` and `source_table` are used for key to find origin data?
- is this format our own or from schema?

if this is from schema, we may consider conflict with tflite file improt.
- It would be better to use one key for origin data instead of two items
- to be safe, it would be better to use some prefix for our keys... this is the first items so I think this is very important
- I suggest to use `ONE_NodeOrigin` as the key

You wrote `// TODO Handle this by enum`... can you explain how you mean by `enum`?

compiler/luci/export/src/ExportMetadata.cpp
seanshpark: > I thunk your intention is to handle these similar codes effectively.

Well, providing this as a function is one thing and this is necessary to reduce duplicate codes.
Actually how to make simple serializing 32 bit to four 8 bit is the main topic I like to solve :)

compiler/luci/import/src/ImportMetadata.cpp
llFreetimell: > are these to `op_table` and `source_table` are used for key to find origin data?
> - is this format our own or from schema?

Yes, this is our own format.
Schema only provides a placeholder for storing the data. (`table Metadata`)

> if this is from schema, we may consider conflict with tflite file improt.
> - It would be better to use one key for origin data instead of two items
> - to be safe, it would be better to use some prefix for our keys... this is the first items so I think this is very important
> - I suggest to use `ONE_NodeOrigin` as the key

For now, tflite only produces `min_runtime_version` metadata so there is no problem.
However, to prevent future conflicts, using some prefix for our metadata seems good.
( `ONE_` looks good to me)
Nevertheless, I think integrating the two table as one is not so good.

- Two table include different information.
- If one table include huge amount of data, it is hard to manage them.
- Backend is also using two kinds of table. There is no obligation to follow exactly same format, but similar format looks better I think.

> You wrote `// TODO Handle this by enum`... can you explain how you mean by `enum`?

This is minor thing :D
For now, I wrote a function like `metadata_with_op_table(buffer)`.
I thought which is better, `metadata_with_op_table(buffer)` or `metadata<OPTABLE>(buffer)`
`OPTABLE` is enum :)
`TODO` in here means this code is not fixed and I am thinking...!

compiler/luci/export/src/ExportMetadata.h
```diff
@@ -0,0 +1,53 @@
...
+{
+
+class ExportMetadata
+{
+public:
+  ExportMetadata() = default;
+
+public:
+  void link_source(uint32_t node_id, uint32_t source_id);
+  void add_source(uint32_t source_id, std::string origin_name);
```
llFreetimell, line 36: (Originally written by @seanshpark)

Can you add a comment or readme file how node_id, source_id and origin_name is converted to MetaData?

llFreetimell: I will add a comment or readme file about them in the near future :D
Maybe right after implementation is almost finished?

compiler/luci/import/src/ImportMetadata.cpp
```diff
@@ -0,0 +1,161 @@
...
+  assert(idx == buffer.size());
+
+  return node_source_ids_map;
+}
+
+/**
+ * 'source_table'
+ * - [ id : uint32_t ][ length : uint32_t ][ data : 'length' Bytes ]
+ */
+std::map<uint32_t, std::string> metadata_with_source_table(const std::vector<uint8_t> &buffer)
```
llFreetimell, line 75: (Originally written by @seanshpark)

how about declare a type (struct or class) for `std::map<uint32_t, std::string>` ?

compiler/luci/import/src/ImportMetadata.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  return source_id_name_map;
+}
+
+} // namespace
+
+namespace luci
+{
+
+ImportMetadata::ImportMetadata(const luci::CircleReader &reader)
```
llFreetimell, line 114: (Originally written by @seanshpark)

please implement body to some separate method NOT in ctor

mhs4670go:
@llFreetimell 
As of now, source table consists of following parts.
[ node_id : uint32_t ][ node's name length : uint32_t ][ data : 'length' Bytes ]

For validation, I think entry number should be inserted at the start of the data like below.

[ entry number : uint32_t ] ... [ node_id : uint32_t ][ node's name length : uint32_t ][ data : 'length' Bytes ]

compiler/luci/import/src/ImportMetadata.cpp
llFreetimell: As this type is used only in `CircleMetadataImporter.cpp`, leaving as it is seems not bad for me :D

llFreetimell:

PR #6102, opened 2021-02-21
[luci] Common for CircleIf shape/dtype inference

seanshpark:
This will extract common code for CircleIf shape/dtype inference as
get_out_graphs method.


seanshpark:
For #6053, from #6054,

seanshpark:
CC @llFreetimell

mhs4670go (approved):
LGTM

PR #6103, opened 2021-02-22
[luci] Introduce BatchNormPatternFinder

seanshpark:
This will introduce common BatchNormPatternFinder with is_batchnorm_add and
is_batchnorm_mul methods for Passes.

### DIFF
--- a/compiler/luci/pass/src/BatchNormPatternFinder.h
+++ b/compiler/luci/pass/src/BatchNormPatternFinder.h
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_PASS_BATCH_NORM_PATTERN_FINDER_H__
+#define __LUCI_PASS_BATCH_NORM_PATTERN_FINDER_H__
+
+#include <luci/IR/CircleNodes.h>
+
+namespace luci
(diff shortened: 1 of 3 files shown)


seanshpark:
For #6053, from #6054

jinevening:
`FindPatternHelper` looks too general to me. How about `BatchNormPatternFinder`? I remember that we do not make a big file with multiple purposes.

compiler/luci/pass/src/BatchNormPatternFinder.h
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_PASS_FIND_PATTERN_HELPER_H__
+#define __LUCI_PASS_FIND_PATTERN_HELPER_H__
```
jinevening, line 18: This should be updated too :)

compiler/luci/pass/src/BatchNormPatternFinder.test.cpp
```diff
@@ -0,0 +1,267 @@
...
+protected:
+  loco::Graph _g;
+  loco::GraphInput *_graph_input = nullptr;
+  loco::GraphOutput *_graph_output = nullptr;
+  luci::CircleInput *_input = nullptr;
+  luci::CircleOutput *_output = nullptr;
+};
+
+/**
+ * @brief Graphlet with Add and Const as beta from BatchNorm
```
jinevening, line 91: How about removing the term Graphlet? graphlet is the term also used in our internal repo, so it may be confusing to be used here. Unlike the graphlet in our internal repo, the graphlet in this file is not inferrable, and this class does not have outlet or ingredients as its members.

seanshpark: `Graph` + `let` which is general term that is not complete Graph but a component of a Graph.
I thought of `SubGraph` but may be confused with Graph with `CircleSub` Op.
Refer https://stackoverflow.com/questions/38586577/why-let-word-added-to-servlet-applet-cmdlet

jinevening: I see.

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6104, opened 2021-02-22
[luci] Revise MakeBatchNormGammaPositivePass

seanshpark:
This will revise MakeBatchNormGammaPositivePass to find pattern
beginning with CircleAdd.

### DIFF
--- a/compiler/luci/pass/src/MakeBatchNormGammaPositivePass.cpp
+++ b/compiler/luci/pass/src/MakeBatchNormGammaPositivePass.cpp
@@ -40,27 +40,39 @@ bool negative_gamma_to_positive(luci::CircleConst *gamma)
 }
 
 // Check if add is batchnorm add
-bool is_batchnorm_add(const luci::CircleAdd *add)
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6053, from #6054

compiler/luci/pass/src/MakeBatchNormGammaPositivePass.cpp
```diff
@@ -40,27 +40,39 @@ bool negative_gamma_to_positive(luci::CircleConst *gamma)
 }
 
 // Check if add is batchnorm add
-bool is_batchnorm_add(const luci::CircleAdd *add)
+bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::CircleConst *&beta)
```
jinevening, line 43: I guess that this will be replaced with the one in `FindPatternHelper.h` (https://github.com/Samsung/ONE/pull/6103). Do you want to merge both and update this file later?

seanshpark: This is to minimize the change. I'll revise this again after #6103 lands.

jinevening (approved):
LGTM

PR #6107, opened 2021-02-22
[record-minmax] Extract common profileData

seanshpark:
This will extract common codes in profileData to a function.


seanshpark:
For #6053, from #6054,

jinevening (approved):
LGTM

PR #6108, opened 2021-02-22
[luci] Extract common ConvertNCHWToNHWCPass

seanshpark:
This will extract common code in ConvertNCHWToNHWCPass with Relu type.

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -369,6 +369,26 @@ bool is_NCHW_with_const(const luci::CircleAdd *node, luci::CircleNode *&pred_nod
     return false;
 }
 
+template <class T> bool convert_unary_features(T *node)
+{
+  const auto pred_node = loco::must_cast<luci::CircleNode *>(node->features());
+  auto pre_trans = create_pre_transpose(node);
+  pre_trans->a(pred_node);
+  node->features(pre_trans);
+
+  // Do shape inference for this node again.
+  // TODO Remove loco::shape_erase()
+  loco::shape_erase(node);
+  node->shape_status(luci::ShapeStatus::UNDEFINED);
+
+  auto post_trans = create_post_transpose(node);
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6053, from #6054,

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -369,6 +369,26 @@ bool is_NCHW_with_const(const luci::CircleAdd *node, luci::CircleNode *&pred_nod
     return false;
 }
 
+template <class T> bool convert_Relu(T *node)
```
jinevening, line 372: How about naming this `convert_single_input_operator` and including Neg too?

`node->features()` in the function body may be replaced with `node->args(0)`.

seanshpark: how about shorter `convert_unary` ?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -369,6 +369,26 @@ bool is_NCHW_with_const(const luci::CircleAdd *node, luci::CircleNode *&pred_nod
...
+{
+  const auto pred_node = loco::must_cast<luci::CircleNode *>(node->features());
+  auto pre_trans = create_pre_transpose(node);
+  pre_trans->a(pred_node);
+  node->features(pre_trans);
```
seanshpark, line 377: we cannot use like `node->args(0) = pre_trans;`

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
seanshpark: for future, I'll update this method as `convert_unary_features`

jinevening: > We cannot use node->args(0) as setter. so cannot merge with Neg

Oh I see. Thank you.

jinevening (approved):
LGTM

PR #6109, opened 2021-02-22
[luci] Add MakeBatchNormGammaPositivePass discription

seanshpark:
This will add simple discription of MakeBatchNormGammaPositivePass.

### DIFF
--- a/compiler/luci/pass/src/MakeBatchNormGammaPositivePass.cpp
+++ b/compiler/luci/pass/src/MakeBatchNormGammaPositivePass.cpp
@@ -70,7 +70,7 @@ namespace luci
 {
 
 /**
- * Make gamma value of Mul-Add(as BatchNorm) to positive
+ * Make negative gamma values of Mul-Add (as BatchNorm) to a small positive value (1e-10)
  *
  *  PATTERN:
  *          |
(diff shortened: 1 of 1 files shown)


seanshpark:
from #6054,

@jinevening , I'm not sure this discription is correct. If not, could you please add a comment?

compiler/luci/pass/src/MakeBatchNormGammaPositivePass.cpp
```diff
@@ -69,6 +69,18 @@ bool make_positive_gamma(luci::CircleAdd *add)
 namespace luci
 {
 
+/**
+ * Make gamma value of Mul-Add(as BatchNorm) to positive
```
jinevening, line 73:
```suggestion
 * Make negative gamma values of Mul-Add (as BatchNorm) to a small positive value (1e-10)
```

jinevening: > @jinevening , I'm not sure this discription is correct. If not, could you please add a comment?

I slightly updated the comment :)

jinevening (approved):
LGTM Thank you

Issue #6055, opened 2021-02-17
Compiler FE: Remove stdex usage


seanshpark:
Let's remove using `stdex` library for (active) frontend codes.

- [x] remove usage of `stdex`
- [x] wait backend to clean `stdex` and rebase to latest
- [x] remove `stdex` module

seanshpark:
done :)

PR #6112, opened 2021-02-22
DRAFT: Remove stdex usage from compiler frontend

seanshpark:
on-going draft to remove stdex from compiler frontend


seanshpark:
For  #6055

PR #6152, opened 2021-02-23
[luci] Tidy RemoveRedundantTransposePass comments

seanshpark:
This will tidy RemoveRedundantTransposePass diagram comments.

### DIFF
--- a/compiler/luci/pass/src/RemoveRedundantTransposePass.cpp
+++ b/compiler/luci/pass/src/RemoveRedundantTransposePass.cpp
@@ -90,30 +90,26 @@ bool remove_consecutive_transpose_function(luci::CircleTranspose *target_node)
 
 namespace luci
 {
+
 /**
  *  BEFORE
  *         |
  *   [CircleNode]     [CircleConst]
- *    (main_node)      (main_perm)
- *         \               /
+ *         |           (pred_perm)
+ *          \              /
  *         [CircleTranspose]  [CircleConst]
- *            (pred_node)      (pred_perm)
+ *            (pred_node)     (target_perm)
  *                 \               /
  *                 [CircleTranspose]
  *                   (target_node)
  *                         |
  *
  *  AFTER
- *      <Optional Case>
- *
- *          |                 |                   |
- *    [CircleNode]      [CircleConst]             |
- *     (main_node)     (new_const_node)           |
- *           \               /           or  [CircleNode]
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/RemoveRedundantTransposePass.cpp
```diff
@@ -90,30 +90,26 @@ bool remove_consecutive_transpose_function(luci::CircleTranspose *target_node)
...
  *  BEFORE
  *         |
  *   [CircleNode]     [CircleConst]
- *    (main_node)      (main_perm)
- *         \               /
+ *         |           (pred_perm)
+ *          \              /
  *         [CircleTranspose]  [CircleConst]
- *            (pred_node)      (pred_perm)
+ *            (pred_node)     (target_perm)
```
struss, line 101: https://github.com/Samsung/ONE/blob/e2014ebf11b6e43c049ad7c46485a1734513e88e/compiler/luci/pass/src/RemoveRedundantTransposePass.cpp#L44-L46
If we change this comment from `perd_perm` to `target_perm`, then this should be changed to matching with comment.

compiler/luci/pass/src/RemoveRedundantTransposePass.cpp
```diff
@@ -90,30 +90,26 @@ bool remove_consecutive_transpose_function(luci::CircleTranspose *target_node)
...
 namespace luci
 {
+
 /**
  *  BEFORE
  *         |
  *   [CircleNode]     [CircleConst]
- *    (main_node)      (main_perm)
- *         \               /
+ *         |           (pred_perm)
```
struss, line 98: https://github.com/Samsung/ONE/blob/e2014ebf11b6e43c049ad7c46485a1734513e88e/compiler/luci/pass/src/RemoveRedundantTransposePass.cpp#L48-L50
If we change this comment from `main_perm`to `pred_perm`, then this should be changed to matching with comment.

compiler/luci/pass/src/RemoveRedundantTransposePass.cpp
seanshpark: I'll post another PR as thic PR is to fix comments.

seanshpark: #6153 posted

struss (approved):
As Commented contents will fixed on #6153 , LGTM!

PR #6154, opened 2021-02-24
[one-cmds] Revise one-prepare-venv with onnx

seanshpark:
This will revise one-prepare-venv to install onnx 1.8.0 for custom
onnx-tensorflow and add a note about this.


mhs4670go (approved):
LGTM

PR #6156, opened 2021-02-25
DRAFT: Common GraphBuilder for multiple outputs

seanshpark:
On going draft to provide common GraphBuilder for multiple outputs.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #6158, opened 2021-02-25
[luci] Set dummy dtype for CircleOutputExclude

seanshpark:
This will set dummy dtype for CircleOutputExclude node at GraphBuilder
and remove for BidirectionalSequenceLSTM and
eUnidirectionalSequenceLSTM.


seanshpark:
Found this while working at #6156

struss (approved):
LGTM!

PR #6159, opened 2021-02-25
[luci] Removed dtype settings for CircleOutputExclude

seanshpark:
This will remove dtype settings for CircleOutputExclude that is now set
at caller GraphBuilder class.

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleTransposeConv.cpp
+++ b/compiler/luci/import/src/Nodes/CircleTransposeConv.cpp
@@ -61,16 +61,15 @@ CircleNode *CircleTransposeConvGraphBuilder::build_node(const circle::OperatorT 
   node->filter(inputs.at(1));
   node->outBackprop(inputs.at(2));
   if (inputs.size() == 3)
-    node->bias(graph->nodes()->create<CircleOutputExclude>());
+  {
+    auto *bias = graph->nodes()->create<CircleOutputExclude>();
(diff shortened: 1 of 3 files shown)


seanshpark:
Related #6158

compiler/luci/import/src/Nodes/CircleTransposeConv.cpp
```diff
@@ -61,16 +61,15 @@ CircleNode *CircleTransposeConvGraphBuilder::build_node(const circle::OperatorT
...
-    node->bias(inputs.at(3));
-
-  if (auto bias = dynamic_cast<luci::CircleOutputExclude *>(node->bias()))
   {
-    // CircleOutputExclude doesn't need a type, but since all nodes must have a type, a dummy type
-    // is inserted.
+    auto *bias = graph->nodes()->create<CircleOutputExclude>();
+    // CircleOutputExclude doesn't need a type, but since all nodes must have a type,
+    // a dummy type is inserted.
     bias->dtype(loco::DataType::FLOAT32);
```
llFreetimell, line 68: According to PR comment, this line should be removed too?

seanshpark: as `CircleOutputExclude` is created here, uline 65, we need to set `dtype` here.

llFreetimell: Ah, I understood :)

llFreetimell (approved):
LGTM

Issue #6155, opened 2021-02-24
Compiler FE: Common method for GraphBuilder::build


seanshpark:
`GraphBuilder::build` methods in `luci/import` for multiple outputs has a similar pattern.
Calling a common method will reduce duplicate codes and lower maintenance cost.
Let's make this common method.

seanshpark:
Buildes to revisit
- CircleCustomGraphBuilder
- CircleIfGraphBuilder
- CircleNonMaxSuppressionV4GraphBuilder
- CircleNonMaxSuppressionV5GraphBuilder
- CircleSplitGraphBuilder
- CircleSplitVGraphBuilder
- CircleTopKV2GraphBuilder
- CircleUniqueGraphBuilder
- CircleUnpackGraphBuilder
- CircleWhileGraphBuilder

seanshpark:
done

PR #6160, opened 2021-02-25
[luci] Introduce GraphBuilderMultiOutput

seanshpark:
This will introduce GraphBuilderMultiOutput as base class for building nodes
with multiple outputs.

### DIFF
--- a/compiler/luci/import/include/luci/Import/GraphBuilderMultiOutput.h
+++ b/compiler/luci/import/include/luci/Import/GraphBuilderMultiOutput.h
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_IMPORT_GRAPH_BUILDER_MULTI_OUTPUT_H__
+#define __LUCI_IMPORT_GRAPH_BUILDER_MULTI_OUTPUT_H__
+
+#include "GraphBuilderContext.h"
+#include "GraphBuilderBase.h"
+
+#include <mio/circle/schema_generated.h>
+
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6155, from draft #6156

compiler/luci/import/include/luci/Import/GraphBuilderMulti.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <mio/circle/schema_generated.h>
+
+namespace luci
+{
+
+/**
+ * @brief Base of general multiple outputs graph builder(e.g., CircleIfGraphBuilder)
+ */
+class GraphBuilderMulti : public GraphBuilderBase
```
llFreetimell, line 31: As `Multi` is a prefix, not noun, it seems little bit weird.
How about `MultipleOutputsGraphBuilder`, `MultiOutputsGraphBuilder` or `MultiOutGraphBuilder`?

compiler/luci/import/include/luci/Import/GraphBuilderMulti.h
```diff
@@ -0,0 +1,67 @@
...
+    GraphBuilderContext *context;
+    const std::vector<CircleNode *> &input_nodes;
+  };
+
+  struct BuildOutArgs
+  {
+    BuildOutArgs(CircleNode *nd, uint32_t n) : node(nd), n(n) {}
+
+    CircleNode *node;
+    uint32_t n;
```
llFreetimell, line 57: At first, I thought that `n` is the number of outputs.
But after I read the codes, I think `n` is the output index of node which has multiple outputs, am I right?

If then, how about `output_index` instead of `n`? :)

compiler/luci/import/include/luci/Import/GraphBuilderMulti.h
seanshpark: I'd like to keep `GraphBuilder` as prefix of this series... so, is `GraphBuilderMultiOutput` OK?

compiler/luci/import/include/luci/Import/GraphBuilderMulti.h
seanshpark: I like short one. Is `index` OK? as class itself has `Out` in the name.

llFreetimell: `index` seems OK :)

p.s. I love short names too but little bit longer name is inevitable when the short name causes confusion I think :)

compiler/luci/import/include/luci/Import/GraphBuilderMulti.h
llFreetimell: `GraphBuilderMultiOutput` looks fine to me :D

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

PR #6162, opened 2021-03-02
[pota-value-test] Support comparison of scalar weights

jinevening:
This supports comparison of scalar weights.

### DIFF
--- a/compiler/pota-quantization-value-test/compare_tensors.py
+++ b/compiler/pota-quantization-value-test/compare_tensors.py
@@ -68,7 +68,7 @@ def compare_quantization(tensor, tensor_name, expect_dir):
     for key in json_load:
         if key == "weights":
             expected_weights = np.array(json_load["weights"])
-            input_weights = tensor["weights"][:]
+            input_weights = tensor["weights"][()]
             abs_tolerance = 1
             # We use higher tolerance for int64 data (bias of int16-quantized model)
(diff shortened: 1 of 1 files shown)


jinevening:
The previous code failed when weights of an Op is a scalar (e.g., SPLIT).

This change enables to fetch the value of a scalar dataset in the h5 file.

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

compiler/pota-quantization-value-test/compare_tensors.py
```diff
@@ -68,7 +68,7 @@ def compare_quantization(tensor, tensor_name, expect_dir):
     for key in json_load:
         if key == "weights":
             expected_weights = np.array(json_load["weights"])
-            input_weights = tensor["weights"][:]
```
llFreetimell, line 71: Just curious, where can I find this grammar reference?

jinevening: You can find it here https://docs.h5py.org/en/stable/high/dataset.html#reading-writing-data.

mhs4670go: @llFreetimell https://docs.h5py.org/en/stable/high/dataset.html#reading-writing-data

PR #6163, opened 2021-03-02
[compiler/ahub] Remove stdex

seanshpark:
This will remove stdex item as not used anymore.


seanshpark:
For #6055

lemmaa (approved):
LGTM!

lemmaa:
So, will we remove `./stdex` and the files below it from the next PR?

seanshpark:
> So, will we remove ./stdex and the files below it from the next PR?

I'll remove `stdex` folder itself after our internal compiler backend is safe without this module.

PR #6165, opened 2021-03-02
[infra/preset] Remove stdex

seanshpark:
This will remove stdex item as it's not used anymore.

ONE-DCO-Signed-off-by: SaeHie Park <saehie.park@gmail.com>


seanshpark:
For #6055

mhs4670go (approved):
LGTM

PR #6167, opened 2021-03-02
DRAFT: Quantization of SPLIT Op

jinevening:
On going draft to support quantization of SPLIT.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: #6161

Issue #6161, opened 2021-02-26
[circle-quantizer] Unsupported SPLIT Op


a-zh:
**How to reproduce:**

Run the command: 
```
$ ./build/arch/circle-quantizer/circle-quantizer --quantize_with_minmax float32 uint8 channel /tmp/1/model.minmax_recorded.circle /tmp/1/model.q.circle
terminate called after throwing an instance of 'std::runtime_error'
  what():  Unsupported Op for const inputs
Aborted
```

Please, find `model.minmax_recorded.circle` in archive: [model.minmax_recorded.zip](https://github.com/Samsung/ONE/files/6049037/model.minmax_recorded.zip)

a-zh:
@hj1-jeong @seanshpark

jinevening:
Hi, I'm hj1-jeong.

circle-quantizer does not support quantization of SPLIT Op yet. I can make it for you next Tuesday~Wednesday (Monday is our national holiday). 

I think SPLIT does not require  a special logic. If you need quantized SPLIT model, you can try to compile circle-quantizer with the below change.

jinevening:
The above code is not tested. Please leave a comment if the above code does not work.

a-zh:
> circle-quantizer does not support quantization of SPLIT Op yet. I can make it for you next Tuesday~Wednesday (Monday is our national holiday).

This would be very helpful to me. Thank you.

a-zh:
> Add two lines in QuantizeWithMinMaxPass.cpp (Line number 955).

This workaround works as expected. But we still need to implement quantizing?

jinevening:
> But we still need to implement quantizing?

The workaround does not work if an input tensor is const. I'll soon upload the PRs for quantizing SPLIT.

jinevening:
@a-zh I close this issue as circle-quantizer now supports SPLIT Op.

PR #6168, opened 2021-03-02
[luci] Quantization of SPLIT Op

jinevening:
This supports quantization of SPLIT Op.

--
Related to: #6161
Draft PR: #6167

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -953,6 +953,14 @@ void quantize_const_inputs(luci::CircleNode *node, loco::DataType output_type)
       }
       break;
 
+    case luci::CircleOpcode::SPLIT:
+      // Only the second input is quantized
+      input_node = node->arg(1);
+      const_node = dynamic_cast<luci::CircleConst *>(input_node);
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -953,6 +953,14 @@ void quantize_const_inputs(luci::CircleNode *node, loco::DataType output_type)
       }
       break;
 
+    case luci::CircleOpcode::SPLIT:
+      // Only the second input is quantized
```
seanshpark, line 957: `CircleSplit` input is second( arg(1) ) one and the first one is `split_dim` from schema definition... If you want to reflect this in the comment, how about `Input is second argument for 'SPLIT' from schema.` ? Or if this is not the case, could you explain a bit more about this comment?

jinevening: @seanshpark  You're right about my intention to quantize the second argument.

I wrote the comment in an operator-independent way, because there can be another operator that falls within the same category, i.e., only the second input is quantized.

jinevening: I've updated the comment PTAL.

seanshpark (approved):
LGTM

PR #6111, opened 2021-02-22
DRAFT: Revise luci pass test graphs

seanshpark:
on-going draft to revise luci pass test graphs for reuse and
maintenance.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #6110, opened 2021-02-22
[luci] Revise pass test graph constructions


seanshpark:
`compier/luci/pass` test codes have graph construction that can be structured to layers
- like `ConstantFoldingAddTestGraph` or 
- like classes in `BatchNormPatternFinder.test.cpp`

Let's revise existing test codes for better maintenance and less effort in adding new ones.

seanshpark:
done

PR #6183, opened 2021-03-03
[luci] Introduce TestInitList

seanshpark:
This will introducet InitList which provides alias of
std::initializer_list.

### DIFF
--- a/compiler/luci/pass/src/test/TestInitList.h
+++ b/compiler/luci/pass/src/test/TestInitList.h
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6110, from draft #6111

mhs4670go (approved):
LGTM

jinevening (approved):
LGTM

compiler/luci/pass/src/test/TestInitList.h
```diff
@@ -0,0 +1,35 @@
...
+#include <initializer_list>
+
+#include <cstdint>
+
+namespace luci
+{
+namespace test
+{
+
+using uilist = std::initializer_list<uint32_t>;
```
llFreetimell, line 29: IIUC, `uilist` is used for `shape` and used only at `luci/pass/src/test/`.
If then, `Shape` or `U32Shape` seems better than `uilist` because

- `uilist` decrease readability compared to `std::initializer_list<uint32_t>`.
- If we want to introduce new alias for `std::initializer_list<uint8_t>`, `uilist` is not unique for uint initializer list.

With this approach, we may be able to use `I32SqueezeDim` for `ilist`.

**p.s.** I love short names and I agree that making variable/function name short is better.
However, if short name is always best, short coding(like code golf) is best approach for writing a code but it is not :(

seanshpark: You got a point. I'll think of better name where it is used.

seanshpark: `ilist` is used for `CircleSqueeze` test and for `set_shape_vector()` for `CircleConst`
Using `I32SqueezeDim` is OK but I'll go with general name for `int32_t`

llFreetimell: My first approach was introducing both `I32Shape` and `I32SqueezeDim` but we may be able to handle it later :)
Thanks!

seanshpark:
I'll post another PR as whole things has changed

PR #6184, opened 2021-03-03
[luci] Introduce Shape related for unit tests

seanshpark:
This will introduce Shape related types and methods for unit testing.

### DIFF
--- a/compiler/luci/pass/src/test/TestShape.test.cpp
+++ b/compiler/luci/pass/src/test/TestShape.test.cpp
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6110, from draft #6111

llFreetimell (approved):
LGTM

compiler/luci/pass/src/test/TestShape.test.cpp
```diff
@@ -0,0 +1,57 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "TestShape.h"
+
+/**
+ * @note This file does not hold any test cases but provides methods for tests
```
mhs4670go, line 20: Then, is it okay to rename this file to `TestShape.cpp`?

jinevening (approved):
LGTM

compiler/luci/pass/src/test/TestShape.test.cpp
seanshpark: `TestShape.cpp` will be added as a source to luci library `luci_pass` that is not used anywhere else, so would be inappropriate.

mhs4670go (approved):
LGTM

PR #6186, opened 2021-03-03
[DRAFT] add profiling option to one-cmds

mhs4670go:
This draft is to add profiling option to one-cmds.

Related: #6080 

### DIFF
--- a/compiler/one-cmds/one-build.template.cfg
+++ b/compiler/one-cmds/one-build.template.cfg
@@ -1,6 +1,3 @@
-[DEFAULT]
-profiling_info=False
-
 [one-build]
 one-import-tf=True
 one-import-tflite=False
@@ -22,6 +19,7 @@ converter_version=v1
 [one-optimize]
 input_path=inception_v3.circle
 output_path=inception_v3.opt.circle
+profiling_data_gen=False
 
 [one-pack]
 input_path=inception_v3.opt.circle
--- a/compiler/one-cmds/one-build.template.cfg
+++ b/compiler/one-cmds/one-build.template.cfg
@@ -1,3 +1,6 @@
+[DEFAULT]
+profiling_info=False
+
 [one-build]
 one-import-tf=True
 one-import-tflite=False
--- a/compiler/one-cmds/how-to-use-one-commands.txt
+++ b/compiler/one-cmds/how-to-use-one-commands.txt
@@ -170,6 +170,7 @@ Current transformation options are
   Note that this pass can change the execution result of the model.
   So, use it only when the impact is known to be acceptable.
 - mute_warnings : This will turn off warning messages.
+- profiling_data_gen : This will turn on profiling data generation.
 - remove_redundant_reshape : This fuses or removes redundant reshape operators.
 - remove_redundant_transpose : This fuses or removes redundant transpose operators.
 - remove_unnecessary_reshape : This removes unnecessary reshape operators.
--- a/compiler/one-cmds/one-optimize
+++ b/compiler/one-cmds/one-optimize
@@ -38,10 +38,7 @@ def _get_parser():
     utility_group = parser.add_argument_group('arguments for utility')
 
     utility_group.add_argument(
-        '-p',
-        '--profiling_info',
-        action='store_true',
-        help='produce profiling information')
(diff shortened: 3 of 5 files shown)


compiler/one-cmds/one-build.template.cfg
```diff
@@ -1,3 +1,6 @@
+[DEFAULT]
```
seanshpark, line 1: This `DEFAULT` section seems somewhat different category compare to `one-build` and others.
Could you elaborate about this?

mhs4670go: `DEFAULT` section is a built-in section of `configparser`. It provides default values for all other sections. Below two config file are exactly the same meaning.
```cfg
[DEFAULT]
profiling_info=True

[one-build]
one-import-tf=True
one-import-tflite=False
one-import-bcq=False
one-optimize=True
one-quantize=False
one-pack=False
one-codegen=False

[one-import-tf]
input_path=inception_v3.pb
input_arrays=input
input_shapes=1,299,299,3
output_arrays=InceptionV3/Predictions/Reshape_1
converter_version=v2

[one-optimize]
input_path=inception_v3.circle
output_path=inception_v3.opt.circle
```

```cfg
[one-build]
one-import-tf=True
one-import-tflite=False
one-import-bcq=False
one-optimize=True
one-quantize=False
one-pack=False
one-codegen=False

[one-import-tf]
input_path=inception_v3.pb
input_arrays=input
input_shapes=1,299,299,3
output_arrays=InceptionV3/Predictions/Reshape_1
converter_version=v2
profiling_info=True // note here

[one-optimize]
input_path=inception_v3.circle
output_path=inception_v3.opt.circle
profiling_info=True // note here
```

seanshpark: Thanks for detail :)
I'm not sure `DEFAULT` is the name for the purpose... need some time to think about this...

These are the questions arose...
- does the key-values exist in `DEFAULT` provide to all other sections? what happens if not?
- can other key-values move here?

mhs4670go: > does the key-values exist in DEFAULT provide to all other sections? what happens if not?

Sorry but I don't understand what the question wants to. `DEFAULT` section DOES provide its options with all other sections like above example.

> can other key-values move here?

As of now, I think logging and warning related options could be here.

seanshpark: > Sorry but I don't understand what the question wants to. DEFAULT section DOES provide its options with all other sections like above example.

I think `all other sections` is dangerous...
While adding an option to each programs, `ALL` will be applied to all sections: which are commands.
We can only test by enabling all commands to support items in `DEFAULT`, if not command without that option may drop unsupported option, which means we cannot test individual commands for the support of that option unless others support this. 
Or  my understanding is incorrect?

seanshpark: Another question: can each sections override `DEFAULT` ?
```
[DEFAULT]
profiling_info=True

[one-optimize]
profiling_info=False
```
I'm not sure this has some meaning but above will let all others has `profiling_info=True` except `one-optimize`

seanshpark: This is something like what I understand at first seeing... although you may thing this is strange,
```
[DEFAULT]
profiling_info_value=True

[one-optimize]
profiling_info=profiling_info_value
```

then when we want to change it
```
[DEFAULT]
profiling_info_value=False

[one-optimize]
profiling_info=profiling_info_value
```

mhs4670go: @seanshpark 
I thought it would be possible for drivers to ignore unsupported options like `--profiling_info`. But, it won't be good a you said.

We can override the values in section at runtime. So, `profiling_info` key in the driver that doesn't need to it can be remove or set `False`.

> Another question: can each sections override DEFAULT ?

Yes. And as I said above, `profiling_info=False` can be moved to internal code rather than configuration file.

And, `DEFAULT` section name could be customized. If `DEFAULT` is a little bit weird, we can change it.

llFreetimell:
@mhs4670go We may add new option to `how-to-use-one-commands.txt` too.

PR #6187, opened 2021-03-03
[luci] Introduce TestFirstNode

seanshpark:
This will introduce TestFirstNode providing first_node method that will
return first node of a type in the graph.


seanshpark:
From #6111

llFreetimell (approved):
LGTM

mhs4670go (approved):
LGTM

I think `target_nodes<T>(g)` could be introduced as well that returns set of T in g.

jinevening (approved):
LGTM

PR #6188, opened 2021-03-03
[luci] Introduce TestIOGraph and components

seanshpark:
This will introduce TestIOGraph and components to declare
any test base Graph with input(s) and output(s).

### DIFF
--- a/compiler/luci/pass/src/test/TestIOGraph.h
+++ b/compiler/luci/pass/src/test/TestIOGraph.h
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


seanshpark:
From #6111

compiler/luci/pass/src/test/TestIOGraph.h
```diff
@@ -0,0 +1,147 @@
...
+#include <luci/IR/CircleNodes.h>
+
+namespace luci
+{
+namespace test
+{
+
+/**
+ * @brief Graphlet with Inputs and loco::Graph for multiple inputs
+ * @note  Every Graph will have Input(s) and Output(s)
```
jinevening, line 31: If this note holds, how about making `TestIOGraphlet`, which can give the number of inputs and outputs at once?

For example,
```cpp
template<unsigned I, unsigned O> class TestIOGraphlet
```

seanshpark: This is good idea :) Current implementation is what is actually used. Adding `TestIOGraphlet` can be done when used, maybe you can later add it.

llFreetimell (approved):
LGTM

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6191, opened 2021-03-03
[luci-codegen] CMake related infrastructure

binarman:
- Halide downloader
- Codegen basic CmakeLists
- Codegen wrapper library infrastructure


binarman:
@seanshpark 
This is cmake infrastructure that I ended up when moved codegen to separate directory.

seanshpark:
@binarman , could you consider this?
I was thinking of `codegen` as root folder for code generators, not only one, as there are lots of modules in `compiler` and `runtime`
So, how about `codegen/luci-codegen` as your root directory (I'm still looking at your draft codes so I understanding may not be correct) ?

seanshpark:
I think it would be better to separate `codegen` files and `infra` files.

binarman:
> So, how about codegen/luci-codegen as your root directory

I'm ok with this =)

I think we need more people to discuss infrastructure, because I am not sure what other modules (and how) we should place in codegen directory. Could you clarify what modules you mean?

seanshpark:
> I think we need more people to discuss infrastructure,

@hseok-oh , could you please take a look?

> I am not sure what other modules (and how) we should place in codegen directory.

At the beginning your work is all that I could think for now. If you have something in your mind, could you please share?

seanshpark:
I think we need to revise CI builder for new `codegen` root folder...

seanshpark:
Please look at https://github.com/Samsung/ONE/pull/5836#issuecomment-790144685

hseok-oh:
@seanshpark 

>how about codegen/luci-codegen as your root directory 

Agree

@binarman 

Please rename `Halide-11.0.1` to `HalideSource-11.0.1`. We are using `Source` suffix for external code downloader. (ex. `TensorFlowGEMMLowpSource-2.3.0`)

seanshpark:
> Please rename Halide-11.0.1 to HalideSource-11.0.1

@binarman , could we use binary package not to build halide from source in CI? just to reduce load to CI.
if possible, we can add another docker image with halide/llvm for codegen.

hseok-oh:
>Please rename `Halide-11.0.1` to `HalideSource-11.0.1`. We are using `Source` suffix for external code downloader. (ex. `TensorFlowGEMMLowpSource-2.3.0`)

I thought `Halide-11.0.1` download source code, but it seems downloading x86 binary. Then we don't need to rename.

binarman:
@hseok-oh 

> it seems downloading x86 binary.

Yes, it is.
To be honest I am not satisfied with current solution, because it limits build to x86 linux systems.
I didn't find simple cross platform solution in cmake how to detect current platform so far, but I hope it is fine for now.

hseok-oh:
>To be honest I am not satisfied with current solution, because it limits build to x86 linux systems.
I didn't find simple cross platform solution in cmake how to detect current platform so far, but I hope it is fine for now.

I also have same concern, and agree with your comment that it is fine for now.

binarman:
> At the beginning your work is all that I could think for now. If you have something in your mind, could you please share?

We already have source generating compilers, like `nnc` and `enco`, but I suppose they are abandoned now, and I like to leave them as-is.

About future directions, we can utilize some other technologies

For now I see two major opportunities:
- TVM-based compiler. TVM implements halide-like IR (it was actually halide some time ago). Major difference from Halide are:
  - pros: It claim better performance of auto scheduled kernels compared to original halide project (At least on the paper)
  - pros: This is already working NN compiler, so no need to write kernels again.
  - cons: This is already working NN compiler, so there could be some difficulties in integration with our pipeline and infrastructure
  - cons: It requires to run code on actual hardware to auto schedule kernels. (though I may be wrong. Maybe @periannath can tell more)

- MLIR-based compiler
   - pros: This is a hot topic and this technology is rapidly growing. It is adopted by TensorFlow project, so I expect it to cover all devices from cpu to specialised NPUs.
   - pros: this is not a stand-alone compiler, but a construction set
   - cons: It is still on early stage of development, so some things are missing and there are a lot of bugs

binarman:
@seanshpark 
Regarding build time and codegen dir:
Then I'll return build to compiler folder and we will see if build time degrade =)

For the reference last build took:
pr-x64-builder-tester : 7 m 10 s
pr-nncc-debug: 34 m
pr-nncc-release: 39 m

binarman:
@hseok-oh 
Unexpected difficulties: cmake installed in build system has version 3.10, but Halide requires 3.16.
On the bright side: I've tried to build it with cmake 3.14.7 (oldest version I can install on my local system) and it succeed, maybe 3.10 will be fine too.

Did you faced this kind of problems before?

seanshpark:
> Then I'll return build to compiler folder and we will see if build time degrade =)

OK, thanks!

hseok-oh:
>Did you faced this kind of problems before?

No. Runtime's cmake version requirement is based on tizen, and tizen was using cmake version 3.9.4 until last year. 

To use cmake 3.16, we need to install cmake manually (download binary and install) on docker image build process by updating dockerfile.

binarman:
@hseok-oh 
> No. Runtime's cmake version requirement is based on tizen, and tizen was using cmake version 3.9.4 until last year.

Thank you for sharing!
I figured out that that it is pretty easy to get around the cmake restriction (see last [commit](https://github.com/Samsung/ONE/pull/6191/commits/cefce8ef4556e74c8324ad04850a31b25a3e4145)).

Now I hit another issue: prebuild Halide uses **glibc 2.29**, but Ubuntu 18.04 uses **2.27**(I believe we use this version in our builds), and produced binaries will be unusable (so no tests will work).

After this I tried to build Halide with glibc 2.27 (and latest cmake from official [releases](https://github.com/Kitware/CMake/releases)) under ubuntu 18.04 and it worked.

So now I have following questions:
1. Can you confirm that we use glibc-2.27 or tell me where can I find docker infrastructure for our builds?
2. Can we put our own compiled Halide in some repo/server/docker image?

P.s. I want to store prebuild package to simplify ONE build. Otherwise I need to download CMake, prebuild LLVM and Halide sources for every build of ONE. I think this adds complexity for nothing.

seanshpark:

seanshpark:

PR #6151, opened 2021-02-23
DRAFT: Rename LuciNodeMixin to CircleNodeMixin

seanshpark:
Draft to rename LuciNodeMixin, LuciNodeTrait to CircleNode~~~.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #6192, opened 2021-03-04
[luci] Use CircleNodeMixin for A,B

seanshpark:
This will revise IRs to use CircleNodeMixin for Nodes A,B.

### DIFF
--- a/compiler/luci/lang/include/luci/IR/Nodes/CircleArgMax.h
+++ b/compiler/luci/lang/include/luci/IR/Nodes/CircleArgMax.h
@@ -20,7 +20,7 @@
 #include "luci/IR/CircleNodeDecl.h"
 #include "luci/IR/CircleOpcode.h"
 
-#include "luci/IR/LuciNodeMixins.h"
+#include "luci/IR/CircleNodeMixins.h"
 
 namespace luci
 {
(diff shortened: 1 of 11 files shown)


seanshpark:
From #6151

compiler/luci/lang/include/luci/IR/Nodes/CircleArgMax.h
```diff
@@ -20,7 +20,7 @@
 #include "luci/IR/CircleNodeDecl.h"
 #include "luci/IR/CircleOpcode.h"
 
-#include "luci/IR/LuciNodeMixins.h"
+#include "luci/IR/CircleNodeMixins.h"
```
llFreetimell, line 23: I think `CircleArgMax` do not use neither `LuciNodeMixins` nor `CircleNodeMixins` @.@
How about just remove this header at another PR together with other same cases?

seanshpark: Most Nodes derive from `FixedArityNode` which is in `CircleNodeMixins.h` as it's also a mixin.

llFreetimell: Oh, I didn't catch it, thanks :D

llFreetimell (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6201, opened 2021-03-04
[onert] Skip permute shape calculation

hseok-oh:
This commit skip permute shape calculation when from and to layout is same.


seanshpark (approved):
LGTM

wateret (approved):
LGTM

PR #6213, opened 2021-03-05
[luci] Revise ResolveCustomOp pass not to use bitwise-or

seanshpark:
This will revise ResolveCustomOpAddPass and
ResolveCustomOpBatchMatMulPass to use simple if and not bitwise-or
operator.


seanshpark:
from #6111

mhs4670go (approved):
LGTM

PR #6218, opened 2021-03-05
[luci] Add graph transformation comment

seanshpark:
This will add before-after graph transformation helper comment for
better understanding what ResolveCustomOpBatchMatMulPass does.


seanshpark:
From #6111

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

PR #6220, opened 2021-03-05
[luci/profile] Introduce CircleNodeOrigin

llFreetimell:
Parent Issue : #6080
Draft : #6079

We sometimes want to know where some nodes are originated from.
For that, this commit introduces `CircleNodeOrigin` to keep recording origin information.

### DIFF
--- a/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
+++ b/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
@@ -50,10 +50,10 @@ public:
 std::shared_ptr<CircleNodeOrigin> single_origin(uint32_t id, const std::string &name);
 
 std::shared_ptr<CircleNodeOrigin>
-composite_origin(const std::initializer_list<std::shared_ptr<CircleNodeOrigin>> &origins);
+composite_origin(const std::initializer_list<std::shared_ptr<CircleNodeOrigin>> origins);
 
 std::shared_ptr<CircleNodeOrigin>
-composite_origin(const std::vector<std::shared_ptr<CircleNodeOrigin>> &origins);
+composite_origin(const std::vector<std::shared_ptr<CircleNodeOrigin>> origins);
 
 } // namespace luci
 
--- a/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
+++ b/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
@@ -53,7 +53,7 @@ std::shared_ptr<CircleNodeOrigin>
 composite_origin(const std::initializer_list<std::shared_ptr<CircleNodeOrigin>> origins);
 
 std::shared_ptr<CircleNodeOrigin>
-composite_origin(const std::vector<std::shared_ptr<CircleNodeOrigin>> origins);
+composite_origin(const std::vector<std::shared_ptr<CircleNodeOrigin>> &origins);
 
 } // namespace luci
 
--- a/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
+++ b/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
@@ -47,13 +47,13 @@ public:
   virtual std::set<const Source *> sources(void) const = 0;
 };
 
-std::shared_ptr<CircleNodeOrigin> single_origin(uint32_t id, const std::string name);
+std::shared_ptr<CircleNodeOrigin> single_origin(uint32_t id, const std::string &name);
 
 std::shared_ptr<CircleNodeOrigin>
-composite_origin(std::initializer_list<std::shared_ptr<CircleNodeOrigin>> origins);
+composite_origin(const std::initializer_list<std::shared_ptr<CircleNodeOrigin>> &origins);
 
 std::shared_ptr<CircleNodeOrigin>
-composite_origin(std::vector<std::shared_ptr<CircleNodeOrigin>> origins);
+composite_origin(const std::vector<std::shared_ptr<CircleNodeOrigin>> &origins);
 
 } // namespace luci
 
@@ -64,7 +64,7 @@ bool has_origin(const luci::CircleNode *circle_node);
 
 void add_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin);
 
-// NOTE When circle_node does not have origin, nullptr will be returned
+// NOTE When circle_node does not have origin, nullptr is returned
 const std::shared_ptr<luci::CircleNodeOrigin> get_origin(const luci::CircleNode *circle_node);
 
 } // namespace luci
--- a/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
+++ b/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


mhs4670go (dismissed):
LGTM

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
```diff
@@ -0,0 +1,72 @@
...
+  private:
+    std::string _name;
+    uint32_t _id = 0;
+  };
+
+public:
+  virtual std::set<const Source *> sources(void) const = 0;
+};
+
+std::shared_ptr<CircleNodeOrigin> single_origin(uint32_t id, const std::string name);
```
seanshpark, line 50: is it possible to use `const std::string &name` ?

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
```diff
@@ -0,0 +1,72 @@
...
+} // namespace luci
+
+namespace luci
+{
+
+bool has_origin(const luci::CircleNode *circle_node);
+
+void add_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin);
+
+// NOTE When circle_node does not have origin, nullptr will be returned
```
seanshpark, line 67: maybe just `nullptr will return`

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+std::shared_ptr<CircleNodeOrigin>
+composite_origin(std::initializer_list<std::shared_ptr<CircleNodeOrigin>> origins)
+{
+  return std::make_shared<CompositeOrigin>(origins);
+}
+
+std::shared_ptr<CircleNodeOrigin>
+composite_origin(std::vector<std::shared_ptr<CircleNodeOrigin>> origins)
```
seanshpark, line 132: Q) there is `composite_origin` with `initializer_list` . why is also `vector` needed?

seanshpark: Q) why did you not use `&origins` ?

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
llFreetimell: I think "be returned" is right :)

- This function **returns** `nullptr`
- `nullptr` **is returned** by this function

seanshpark: How about `NOTE When circle_node does not have origin, this function will return nullptr`

`nullptr will be returned` seems OK but `will` and past format looks odd to me...

compiler/luci/profile/src/CircleNodeOrigin.cpp
llFreetimell: > Q) there is composite_origin with initializer_list . why is also vector needed?

When the number of origins is fixed, we can use `initializer_list` and it makes code simple and easy.
(For example, when we propagate origins in `luci/Pass`)

However, when the number is not fixed, we cannot use `initializer_list` :(
(For example, when we import circle file and annotate in `luci/import`)

By that reason, both of them are introduced.

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
llFreetimell: If then, I will change it to `nullptr is returned`

compiler/luci/profile/src/CircleNodeOrigin.cpp
seanshpark: > However, when the number is not fixed, we cannot use initializer_list :(

Can you show me a draft code for this? It's a bit hard to understand the case.

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+public:
+  virtual std::set<const Source *> sources(void) const = 0;
+};
+
+std::shared_ptr<CircleNodeOrigin> single_origin(uint32_t id, const std::string &name);
+
+std::shared_ptr<CircleNodeOrigin>
+composite_origin(const std::initializer_list<std::shared_ptr<CircleNodeOrigin>> &origins);
```
seanshpark, line 53: I'm not good at this but I usually seen `initializer_list origins` without `&` for `initializer_list`. maybe because the caller usually called with `{A, B, ...}` without declaring a variable: I can assume why but don't know this is a must or somekind...

mhs4670go: FYI, https://stackoverflow.com/a/17803511
> It’s passed by value because it’s cheap.

compiler/luci/profile/src/CircleNodeOrigin.cpp
llFreetimell: Following codes use `composite_origin(std::vector<std::shared_ptr<CircleNodeOrigin>> origins)` at #6101.

https://github.com/Samsung/ONE/blob/9f6a31ee0a1d5d024b82cd37acda3bf741ded779/compiler/luci/import/src/ImportMetadata.cpp#L144-L154

As we do not know the size of `source_ids`, we cannot use `initializer_list` for that case.
However, in case of `luci/Pass`, we can know the exact number of origins.
Please refer to following draft codes in #6217.

https://github.com/Samsung/ONE/blob/fc3b7f724b9aa69c0a78adafa2ac431d56f4b0fb/compiler/luci/pass/src/FuseBatchNormWithConvPass.cpp#L204-L205

mhs4670go (dismissed):
LGTM

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
seanshpark: @llFreetimell , you didn't add any about my comment

llFreetimell: @seanshpark 

I thought that key point of your comment was "What is a difference between passing by reference and passing by value in initializer_list?" and https://stackoverflow.com/a/17803511 became an answer. (Frankly speaking, I didn't know that knowledge, until @mhs4670go commented it)

Sorry for misunderstanding :(

Can you let me know what is your main question between following candidates..?
- Should(must) I use `&` in our codes?
- Why did I used `&` in our codes?
- Something else...

seanshpark: like other codes, it would be better to remove `&`

mhs4670go (dismissed):
LGTM

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -0,0 +1,161 @@
...
+namespace luci
+{
+
+std::shared_ptr<CircleNodeOrigin> single_origin(uint32_t id, const std::string &name)
+{
+  return std::make_shared<SingleOrigin>(id, name);
+}
+
+std::shared_ptr<CircleNodeOrigin>
+composite_origin(const std::initializer_list<std::shared_ptr<CircleNodeOrigin>> origins)
```
seanshpark, line 126: This is OK without `&`

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -0,0 +1,161 @@
...
+{
+  return std::make_shared<CompositeOrigin>(origins);
+}
+
+std::shared_ptr<CircleNodeOrigin>
+composite_origin(const std::vector<std::shared_ptr<CircleNodeOrigin>> origins)
```
seanshpark, line 132: Let's use `&origins` as we'll be calling this method with a variable.

seanshpark:
@llFreetimell , sorry to make us confused for not pointing out about where to put `&` and where not to use. both is correct in syntax but I like to change it for common usage and performace.

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

PR #6229, opened 2021-03-08
[luci/export] Deprecate metadata_buffer in CircleExporter

llFreetimell:
This commit deprecates `metadata_buffer` at `CircleExporterImpl.cpp` with following reasons.

- `metadata_buffer` is deprecated in schema and there was no models which use it.
- Current code generate empty vector, which is exactly same with 0.

### DIFF
--- a/compiler/luci/export/src/CircleExporterImpl.cpp
+++ b/compiler/luci/export/src/CircleExporterImpl.cpp
@@ -161,13 +161,9 @@ void CircleExporterImpl::exportGraph(loco::Graph *graph)
   // create array of buffers
   auto buffers = _builder.CreateVector(md._buffers);
 
-  // empty metadata
-  std::vector<int> metadata_buffer_vec;
-  auto metadata_buffer = _builder.CreateVector(metadata_buffer_vec);
-
   // Model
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

compiler/luci/export/src/CircleExporterImpl.cpp
```diff
@@ -225,16 +221,12 @@ void CircleExporterImpl::exportModule(Module *module)
...
-  std::vector<int> metadata_buffer_vec;
-  auto metadata_buffer = _builder.CreateVector(metadata_buffer_vec);
-
   // This version is taken from comment in fbs
   constexpr uint32_t version = 0;
 
   // Model
   auto model_offset = CreateModel(_builder, version, operator_codes, subgraphs, description,
-                                  buffers, metadata_buffer);
+                                  buffers, 0 /* metadata_buffer */);
```
llFreetimell, line 229: @seanshpark I missed one more place, sorry for bothering you :( ... PTAL

seanshpark (approved):
LGTM

Issue #6226, opened 2021-03-08
Compiler FE: Revise one-codegen interface


mhs4670go:
## What

Let's revise `one-codegen` interface.

### before
```
$ ./one-codegen -b ${BACKEND} ${BACKEND_COMMAND}
```
### after
```
$ ./one-codegen -b ${BACKEND} -- ${BACKEND_COMMAND}
```

## Why

`one-codegen` and backend driver has its own interface. So, if both interface has the same option name, it will be a problem.

To avoid this situation, let's split them by `--`.

PR #6232, opened 2021-03-08
[one-cmds] Revise one-codegen interface

mhs4670go:
This commit revises one-codegen interface.

Related : #6226 


mhs4670go:
@cgbahk If this PR is landed, internal issue will be solved.

compiler/one-cmds/dummy-driver/CMakeLists.txt
```diff
@@ -1,11 +1,18 @@
 # dummy driver for interface test
-set(DUMMY_DRIVER_SRC src/dummy-compile.cpp)
```
seanshpark, line 2: can you please leave this as is and add same kind for `help-compile`?

mhs4670go: @seanshpark You mean this way?
```
set(HELP_DRIVER_SRC src/help-compile.cpp)
```

mhs4670go: @seanshpark I thought that way  but `DUMMY_DRIVER_SRC` means dummy driver source. And `help-compile` is also dummy driver.. If it seems okay, I'll update as you said.

seanshpark: This folder is dummy-driver so I think you are adding help-compile to dummy-driver is somewhat NOT what you may want...

1. rename dummy-driver to something else that can hold both executables
2. add help-compile to another driver

mhs4670go: > rename dummy-driver to something else that can hold both executables

Well, I think the name `dummy-driver` already can holds both executables because `dummy-compile` and `help-compile` are dummies.

> add help-compile to another driver

I don't understand the point. Maybe you asked me to create a new folder whose name is `help-driver` or sth and then add `help-compile` to there?

seanshpark: > Well, I think the name dummy-driver already can holds both executables because dummy-compile and help-compile are dummies.

OK, let's go with this

> I don't understand the point.

My mistake... forget about second one...

compiler/one-cmds/one-codegen
```diff
@@ -68,18 +69,28 @@ def _verify_arg(parser, args):
 
 
 def _parse_arg(parser):
-    args, unknown_args = parser.parse_known_args()
+    codegen_args = []
+    backend_args = []
+    # delete file name
+    del sys.argv[0]
```
seanshpark, line 75: I think instead of delete sys.argv[0] and use it, declare a copy of local argv from [1] and using it would be better.
I don't know how this is implemented in python but seems not comforatable...

seanshpark (approved):
LGTM thank you!

PR #6233, opened 2021-03-08
[circle-quantizer] Add profiling_data_gen option

mhs4670go:
This commit adds profiling_data_gen option to circle-quantizer.

Related: #6080

### DIFF
--- a/compiler/circle-quantizer/src/CircleQuantizer.cpp
+++ b/compiler/circle-quantizer/src/CircleQuantizer.cpp
@@ -102,11 +102,8 @@ int entry(int argc, char **argv)
   arser.add_argument("input").nargs(1).type(arser::DataType::STR).help("Input circle model");
   arser.add_argument("output").nargs(1).type(arser::DataType::STR).help("Output circle model");
 
-  arser.add_argument("--profiling_data_gen")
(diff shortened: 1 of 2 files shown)


compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -98,6 +102,12 @@ int entry(int argc, char **argv)
   arser.add_argument("input").nargs(1).type(arser::DataType::STR).help("Input circle model");
   arser.add_argument("output").nargs(1).type(arser::DataType::STR).help("Output circle model");
 
+  arser.add_argument("--profiling_data_gen")
```
jinevening, line 105: How about this?

```suggestion
  arser.add_argument(pdg)
```

jinevening (approved):
LGTM

compiler/circle-quantizer/src/CircleQuantizer.cpp
mhs4670go: @jinevening Oh, good point:)

jinevening (approved):
LGTM Thank you

seanshpark (approved):
LGTM

PR #6238, opened 2021-03-09
[luci] Propagate origin in optimization pass

mhs4670go:
This commit enables each optimization passes to propagate origins.

The number of files to review is kind of big but it simply uses origin setting API. Nevertheless, I think you will have trouble checking if `add_origin` api is used right way because you should know the entire workflow of its optimization. So, it would be better to take a look at the passes you implemented only.

FYI, only operators has its origin. It means the nodes like constant or bias can be ignored in this review.

Related: #6080 
Draft: #6217 


llFreetimell (dismissed):
LGTM

jinevening (dismissed):
LGTM

compiler/luci/pass/src/FuseAddWithTConvPass.cpp
```diff
@@ -105,6 +107,9 @@ bool fuse_add_with_tconv(luci::CircleTransposeConv *tconv)
     replace(add).with(tconv);
   }
 
+  // set origin
+  luci::add_origin(tconv, luci::get_origin(add));
```
seanshpark, line 111: Q) I thought new nodes are set with origin. `tconv` is existing node. is there a rule for setting origin where I can understand the logic?

seanshpark:
Some use `get_origin` and some uses `composite_origin`, both having one argument.
Is the any descript to understand the rule for this?

compiler/luci/pass/src/FuseAddWithTConvPass.cpp
mhs4670go: @seanshpark Basically, origin propagation is for preserving original nodes' information. If node A and node B are fused into new node C, node C should be inherited node A's and node B's origin. Likewise, if node A is fused to node B(no new node, just node B absorbed node A), then node B should get node A's origin. "preserving original nodes" would be helpful.

mhs4670go:
@seanshpark Oh.. maybe `FuseBCQ.cpp` has kind of different logic as @llFreetimell implemented this file.
```
luci::add_origin(bcq_gather, luci::composite_origin({luci::get_origin(gather)}));
```
could be changed to
```
luci::add_origin(bcq_gather, luci::get_origin(gather));
```
I'll fix this part.
---
```
```
And, FYI, `get_origin` is just a getter and `composite_origin` is a constructor-like method. `add_origin` needs `origin` as a second argument that will be added to given `circle_node`'s(first argument) origin. For second argument, we need to create a origin first. `single_origin` and `composite_origin` exist exactly for that purpose.

compiler/luci/pass/src/FuseAddWithTConvPass.cpp
seanshpark: Thanks for the explanation. It seems reasonable.
And is there any document for readers or guide for writers to follow?

llFreetimell (approved):
LGTM, thanks!

compiler/luci/pass/src/FuseAddWithTConvPass.cpp
seanshpark: > Well, no. Actually, it's a bit hard to find proper place to document.``)

Ah.. how about in `luci/Profile/CircleNodeOrigin.h` file? as this is another task, you can do this later or post an task issue not to forget :)

compiler/luci/pass/src/FuseBatchNormWithConvPass.cpp
```diff
@@ -200,6 +201,8 @@ bool fused_batch_norm_with_conv(luci::CircleAdd *add)
   fused_conv->dilation()->h(conv->dilation()->h());
   fused_conv->dilation()->w(conv->dilation()->w());
   fused_conv->name(name + "/Conv2D");
+  luci::add_origin(fused_conv, luci::composite_origin({luci::get_origin(add), luci::get_origin(mul),
+                                                       luci::get_origin(conv)}));
```
seanshpark, line 205: Q) just a question, is this equivalent? for understanding...
```cxx
luci::add_origin(fused_conv, luci::get_origin(add));
luci::add_origin(fused_conv, luci::get_origin(mul));
luci::add_origin(fused_conv, luci::get_origin(conv));
```

seanshpark (approved):
LGTM

compiler/luci/pass/src/FuseBatchNormWithConvPass.cpp
mhs4670go: @seanshpark Yes:)

jinevening (approved):
LGTM

struss (approved):
LGTM

PR #6240, opened 2021-03-09
[luci/export] Introduce CircleExportMetadata

llFreetimell:
Parent : #6080
Draft : #6101

For now, there is no way to store metadata for export.
This commit introduces `CircleExportMetadata` which can store metadata of circle model.
With this, storing metadata from circle model to `CircleExportMetadata`
and exporting metadata from `CircleExportMetadata` to circle file is possible.

### DIFF
--- a/compiler/luci/export/src/CircleMetadataExporter.cpp
+++ b/compiler/luci/export/src/CircleMetadataExporter.cpp
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


mhs4670go (approved):
LGTM

seanshpark:
(1) `Import` uses name `CircleMetadataImporter` which differs `CircleExportMetadata`.
It would be nice if you could go with same convention.

(2) There seems similar comments in import and export. putting only comments in some header is not common but it would be better to add to one place somehow (header or in readme file) and refer in import and export. as this can be done later, please add an issue for this.

llFreetimell:
@seanshpark 

Following is checked.

- File name - Class name
  - CircleMetadataImporter.h - CircleImportMetadata
  - CIrcleMetadataExporter.h - CircleExportMetadata

I will change both of file name to `CircleImportMetadata.h` and `CircleExportMetadata.h`.

llFreetimell:
> There seems similar comments in import and export. putting only comments in some header is not common but it would be better to add to one place somehow (header or in readme file) and refer in import and export. as this can be done later, please add an issue for this.

I will add a comment or readme file about them soon..!
But I will leave current comments for a moment to make review little bit easier :)

seanshpark (approved):
LGTM

PR #6242, opened 2021-03-09
[luci/export] Rename CircleMetadataExporter to CircleExportMetadata

llFreetimell:
Filename is usually same with class name.
However, current `CircleMetadataExporter.cpp` and `CircleExportMetadata` do not match.
This commit will fix it.


llFreetimell:
This PR is originated from https://github.com/Samsung/ONE/pull/6240#issuecomment-793473486

mhs4670go (approved):
LGTM

PR #6245, opened 2021-03-09
[luci] Disable old inference pass

llFreetimell:
Parent Issue : #5501

This commit will disable `TypeInferencePass` and `ShapeInferencePass` which are not used anymore.


seanshpark (approved):
LGTM

PR #6254, opened 2021-03-10
[luci/profile] Add a logic for has_origin

llFreetimell:
As `get_origin` can return `nullptr`, one more `nullptr` check is needed.
This commit will add a new logic for checking whether origin of annotation is `nullptr` or not.

### DIFF
--- a/compiler/luci/profile/src/CircleNodeOrigin.cpp
+++ b/compiler/luci/profile/src/CircleNodeOrigin.cpp
@@ -141,7 +141,13 @@ namespace luci
 
 bool has_origin(const luci::CircleNode *circle_node)
 {
-  return circle_node->annot<CircleNodeOriginAnnotation>() != nullptr;
+  if (circle_node->annot<CircleNodeOriginAnnotation>() == nullptr)
+    return false;
+
+  if (circle_node->annot<CircleNodeOriginAnnotation>()->origin() == nullptr)
+    return false;
+
+  return true;
 }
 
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -141,7 +141,13 @@ namespace luci
 
 bool has_origin(const luci::CircleNode *circle_node)
 {
-  return circle_node->annot<CircleNodeOriginAnnotation>() != nullptr;
+  if (circle_node->annot<CircleNodeOriginAnnotation>() == nullptr)
+    return false;
+
+  if (circle_node->annot<CircleNodeOriginAnnotation>()->origin() == nullptr)
+    return false;
```
mhs4670go, line 148: When does this condition happen?

llFreetimell: Let's assume that a model has no metadata.
Then `nullptr` will be propagated in `luci/pass`.
In that situation, #6101 has following codes.

https://github.com/Samsung/ONE/blob/d9a735e88c6be93507d17cb08c395fe0580a1f9b/compiler/luci/export/src/CircleOperationExporter.cpp#L1497-L1505

If the logic is not added, `get_origin` returns `nullptr` and it causes SEGFAULT.

llFreetimell:
When I talked with @mhs4670go, this will not happen after #6255 is merged.
However, for safety, adding assert seems good.
I will close this PR and create another PR for that :)

PR #6257, opened 2021-03-10
[luci] Remove deprecated codes

llFreetimell:
Parent Issue : #5501

This commit removes deprecated source codes which are not used.


seanshpark (approved):
LGTM

PR #6258, opened 2021-03-10
[luci/export] Insert origin information to metadata

llFreetimell:
Parent Issue : #6080
Draft : #6101

This commit inserts origin information to `SerializedModelData`, which is used for exporting circle file.


compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -1491,7 +1492,19 @@ void exportNode(loco::Node *node, flatbuffers::FlatBufferBuilder &builder, Seria
...
+
+    const auto ops_size = gd._operators.size();
+
     circle_node->accept(&exporter);
+    if (has_origin(circle_node) && ops_size != gd._operators.size())
+    {
+      const auto node_id = gd._operators.size() - 1;
+      for (auto source : get_origin(circle_node)->sources())
+      {
+        md._metadata.add_source_table(source->id(), source->name());
```
mhs4670go, line 1504: This PR is okay but I think the validation should be added to `add_source_table` because source is not unique(think about fission optimization).
So,
```cxx
public:
  void add_source_table(uint32_t source_id, std::string origin_name)
  {
    // Model with multiple subgraph may have different origin_name
    // even if source_id is same. However, as we do not consider about
    // multiple subgraph in profiling for now, just do not care those cases
    // and support them correctly in the future.
    _source_table.emplace(source_id, origin_name);
  }
```
could be
```cxx
{
  auto ret = _source_table.emplace(source_id, origin_name);
  // duplicate key
  if(!ret.second){
    // assert or runtime error.
  }
}
```

mhs4670go (approved):
LGTM

compiler/luci/export/src/CircleOperationExporter.cpp
llFreetimell: As we talked at offline and my comments, adding validation here causes error because same id may have different name in multiple subgraph .
That's why I wrote `just do not care those cases and support them correctly in the future.`.

For fission optimization, I think it is no problem because same source name is propagated.

seanshpark (approved):
LGTM

PR #6259, opened 2021-03-10
[luci/profile] Add nullptr handling logic

llFreetimell:
If `origin` is `nullptr`, not adding origin seems better.
And for safety, adding an assert seems better.
This commit add them.

### DIFF
--- a/compiler/luci/profile/src/CircleNodeOrigin.cpp
+++ b/compiler/luci/profile/src/CircleNodeOrigin.cpp
@@ -146,6 +146,9 @@ bool has_origin(const luci::CircleNode *circle_node)
 
 void add_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin)
 {
+  if (origin == nullptr)
+    return;
+
   circle_node->annot<CircleNodeOriginAnnotation>(nullptr);
   circle_node->annot(std::make_unique<CircleNodeOriginAnnotation>(origin));
 }
@@ -155,6 +158,7 @@ const std::shared_ptr<luci::CircleNodeOrigin> get_origin(const luci::CircleNode 
   if (!has_origin(circle_node))
     return nullptr;
 
+  assert(circle_node->annot<CircleNodeOriginAnnotation>()->origin() != nullptr);
   return circle_node->annot<CircleNodeOriginAnnotation>()->origin();
 }
 


mhs4670go (approved):
LGTM

llFreetimell:
@mhs4670go I think #6254 is needed because of unit tests.

If circle model passes `luci/import`, #6254 is not needed.
However, as unit tests directly generate circle model, #6254 is needed I think.

mhs4670go:
@llFreetimell 
```cxx
// #6254 
{
  if (circle_node->annot<CircleNodeOriginAnnotation>() == nullptr)
    return false;

  if (circle_node->annot<CircleNodeOriginAnnotation>()->origin() == nullptr)
    return false;

  return true;
}
```

As I said offline, second condition doesn't look natural because `CircleNodeOriginAnnotation` with empty origin shouldn't be existed. And, the reason why we didn't add origin related codes to the test is because we assumes origin is optional. 

Then, I think you should fix the point where `CircleNodeOriginAnnotation` with empty origin is born. How about making `add_origin` doing nothing when nullptr is given?
```cxx
{
  circle_node->annot<CircleNodeOriginAnnotation>(nullptr);
  if(origin == nullptr) return;
  circle_node->annot(std::make_unique<CircleNodeOriginAnnotation>(origin));
}
```

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -146,6 +146,9 @@ bool has_origin(const luci::CircleNode *circle_node)
 
 void add_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin)
 {
+  if (origin == nullptr)
```
seanshpark, line 149: maybe add `assert(circle_node != nullptr)` too

seanshpark (approved):
LGTM

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -146,6 +146,9 @@ bool has_origin(const luci::CircleNode *circle_node)
...
+    return;
+
   circle_node->annot<CircleNodeOriginAnnotation>(nullptr);
```
mhs4670go, line 152: Maybe it would rarely happen but what if `add_origin(circle_node, nullptr)` is called where `circle_node` has `CircleNodeOriginAnnotation`. Then, 
1. remove existing notation and doing nothing.
2. just doing nothing.

Your code has chosen number 2. How about number 2?

seanshpark: I think it's about the `what` or the `purpose` the method does... this is what I think
- `add` will add `origin`, which should not be an empty: thus we should drop an error if this condition does not hold.
- additionaly there will be a `clear` method that just clears the annotation.
- it is up to the caller how to use these APIs.

that is, a method should be clear how it works: we usually use `assert` to check the conditions to be clear.

mhs4670go: @seanshpark Actually, it's a bit complicated. 

When we first designed the origin, we assumed that origin is optional because it is just for profiling. But anyway we thought all the nodes(operators) from the loaded model would have its origin because it would be assigned in the import step. So, all profile-related method assumed the parameter(origin ptr) won't be `nullptr`.

But, unfortunately, each pass test script failed then. Because there's no origin assignment in the pass test. So, we came up with two solutions.
1. add origin assignment step into every single tests. 
2. change the API design to another where the parameter can be `nullptr`.

We've chosen the latter because we thought that nodes with no origin data can be run at the passes because origin is optional. With the latter, it is natural that `add_origin` with `nullptr` is doing nothing. I think API spec as a comment better to let developers know about the parameter being able to `nullptr`.

seanshpark: > I think API spec as a comment better to let developers know about the parameter being able to nullptr.

This seems necessary

llFreetimell: Maybe... I should call the function as `set_origin` :(
If the name is `set_origin`, both `set_origin(node, nullptr)` and `set_origin(node, origin1)` make sense...

llFreetimell:
As #6265 is related too, discussion is needed.
Until then, this PR should not be merged...

llFreetimell:
There were two options to handle #6265.

- Change `add_origin` to `set_origin`
  - Logic of `add_origin` is not changed.
  - Some of `luci/pass` should be fixed.
- Change implementation of `add_origin`
  - Logic is changed from setting origin to adding origin.

Both of options make sense but there can be some cases that original origin should be changed.
For example, if `node1` has `{origin1, origin2}` but it should be changed to `{origin1, origin3}`, second option cannot handle it.
Therefore, first option seems better and I would proceed as following steps.

- Add `nullptr` handling logic with this PR
- Change `add_origin` to `set_origin` (just changing name)
- Fix some of `add_origin` usage in `luci/pass`

seanshpark:
> if node1 has {origin1, origin2} but it should be changed to {origin1, origin3}

Is there actual example for this case?

llFreetimell:
> Is there actual example for this case?

For now, there is no cases for that.
However, I am not sure those cases will no occur in the future :(

seanshpark:
> However, I am not sure those cases will not occur in the future :(

`origin2` is removed from the origin and then `origin3` is added.
- `origin2` is removed from the origin : logically can this happen?
 
I didn't think much of this "Profile : Origin" but I don't think this can happen. As @llFreetimell and @mhs4670go had much more time thinking, how do you guys think? It's not about the API but just about the situation...

llFreetimell:
> origin2 is removed from the origin : logically can this happen?

In theory, it is possible but not sure it will really happens...
Following example shows that is it possible in theory.
```
[original model]
Conv/FusedAct=NONE ---- Relu ---- Conv/FusedAct=NONE
    (origin1)        (origin2)        (origin3)

[After FuseActivationFunctionPass]
Conv/FusedAct=Relu ---- Conv/FusedAct=NONE
(origin1, origin2)           (origin3)

[After new optimization in the future]
Conv/FusedAct=NONE ---- Conv/FusedAct=Relu
    (origin1)           (origin2, origin3)
```
In the above example, origin2 is removed from first Conv.

seanshpark:
> But not sure it will really happen or not...

Hmmmmm.... so it would be better to prepare?

Let's now focus on API... we have these two... and maybe more...
- clear + add
- set

We can talk about pros,cons... but I think sometimes this is kind a preference...
- we can (1) clear, (2) add A, (3), add B....
- set({A,B,...})

I don't know which... @llFreetimell , which do you prefer?

llFreetimell:
For clearance, there are two options.

- Option 1 : Add additional logic in `add_origin`
```cpp
{
  if (origin == nullptr)
    return;

  circle_node->annot<CircleNodeOriginAnnotation>(nullptr);
  auto new_origin = composite_origin({get_origin(circle_node), origin}); // this is added
  circle_node->annot(std::make_unique<CircleNodeOriginAnnotation>(new_origin));
}
```

- Option 2 : Rename `add_origin` to `set_origin`
```cpp
// Rename add_origin
void set_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin)
{
  if (origin == nullptr)
    return;

  circle_node->annot<CircleNodeOriginAnnotation>(nullptr);
  circle_node->annot(std::make_unique<CircleNodeOriginAnnotation>(origin));
}
```

---

At first, I choose option 2 when I considered https://github.com/Samsung/ONE/pull/6259#issuecomment-797190669.
However, we can go with option 1 and if https://github.com/Samsung/ONE/pull/6259#issuecomment-797190669 really happens, we can introduce `set_origin` in the future.

> @llFreetimell , which do you prefer?

Both of them seems okay to me because ... "We can talk about pros,cons... but I think sometimes this is kind a preference..." as you said :)
But if I choose one, I prefer option 1 if we can ignore that special case for now.

llFreetimell:
@mhs4670go @seanshpark 

I will go with option 1!
So this PR can be merged now :)

Additional PRs will be created for option 1 :D

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6263, opened 2021-03-11
[onert/lib] Revise RandomTestRunner compile and run method

hseok-oh:
This commit moves I/O value setting from compile() to run(), and moves NNAPI setting from run() to compile().

---

Part of #5971


wateret (approved):
LGTM

PR #6264, opened 2021-03-11
[onert] Fix deallocation bug

wateret:
Fix deallocation bug that was introduced by #5926 .

### DIFF
--- a/runtime/onert/core/src/exec/DynamicShapeInferer.cc
+++ b/runtime/onert/core/src/exec/DynamicShapeInferer.cc
@@ -48,11 +48,11 @@ void DynamicShapeInferer::handleBinaryArithmeticOp(const ir::Operation &op,
 
     So, only when all inputs are static, we can skip dynamic shape inference.
   */
-  if ((!lhs->is_dynamic()) && (!rhs->is_dynamic()))
-    return;
-
   auto output_idx = op.getOutputs().at(0);
   auto output = _tensor_registry->getITensor(output_idx);
+
+  if ((!lhs->is_dynamic()) && (!rhs->is_dynamic()) && !output->is_dynamic())
+    return;
 
   ir::Shape new_shape = shape_inference::inferEltwiseShape(lhs_shape, rhs_shape);
 
--- a/runtime/onert/core/include/backend/ITensor.h
+++ b/runtime/onert/core/include/backend/ITensor.h
@@ -88,6 +88,12 @@ public:
     throw std::runtime_error("This backend does not support dynamic tensor");
   }
 
+  /// @brief Reset the buffer address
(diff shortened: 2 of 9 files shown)


wateret:
## What's changed

- Do not deallocate in FunctionSequence
- Introduce `ITensor::deallocBuffer` which deallocates the buffer
- Introduce `DeallocFunction` class for deallocation of tensors that are no longer used(only for Linear Executor)
- `IDynamicTensorManager`'s method `planDealloc` and `deallocInput` are no longer used

## TODO

- Rename `resetBuffer` to `???`(deallocBuffer?)
- Rename `LinearDeallocFunction` to `DeallocFunction`
- Remove `OperandIndex` in `DeallocList`
- Remove logging used for debugging

wateret:
cc: @hseok-oh @hyunsik-yoon

runtime/onert/core/src/exec/DynamicShapeInferer.cc
```diff
@@ -48,12 +48,12 @@ void DynamicShapeInferer::handleBinaryArithmeticOp(const ir::Operation &op,
 
     So, only when all inputs are static, we can skip dynamic shape inference.
   */
-  if ((!lhs->is_dynamic()) && (!rhs->is_dynamic()))
-    return;
-
   auto output_idx = op.getOutputs().at(0);
   auto output = _tensor_registry->getITensor(output_idx);
 
+  if ((!lhs->is_dynamic()) && (!rhs->is_dynamic()) && !output->is_dynamic())
```
hyunsik-yoon, line 54: To make this code more readable, could you define a function like this?
```
// in case of output tensor of an op, it is possible that 
// the output became dynamic although it had been static before.
// Once a tensor becomes dynamic, it will lost memory allocated for static.
// Therefore once output is dynamic, it should be treated as dynamic tensor. (memory should be allocated at runtime)
// `previously` means `dynamic` or `static` has been set in previous loop in WHILE of previous call of `nnfw_run()`
bool previously_static(ITensor* op_output) { return ! op_output->is_dynamic(); }

// helper function that check if op's input is static
// Note that input of n'th op has been set to static or dynamic by (n-1)th op.
// That's why it is called `currently_static`
bool currently_static(ITensor* op_input) { return ! op_input->is_dynamic(); }
```

and rewrite the line like the following?

```
if (previously_static(output) && (currently_static(lhs) && currently_static(rhs)))
```

If you have any better idea, it would be awesome.

wateret: I will do it in the actual PR for merge, thanks!

hseok-oh:
#6271 is merged

PR #6265, opened 2021-03-11
[Draft][NO MERGE] Fix errors related with luci/profile

llFreetimell:
This draft is for fixing errors and discussing related with luci/profile.

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -19,6 +19,7 @@
 
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/CircleNodeVisitor.h>
+#include <luci/Profile/CircleNodeOrigin.h>
 #include <luci/Log.h>
 
 #include <loco/Service/ShapeInference.h>
@@ -102,6 +103,7 @@ luci::CircleTranspose *create_4d_transpose(luci::CircleNode *node,
   auto trans = node->graph()->nodes()->create<luci::CircleTranspose>();
   trans->perm(perm);
   trans->name(name + "/Transpose_" + str_indices);
+  luci::add_origin(trans, luci::get_origin(node));
 
   return trans;
 }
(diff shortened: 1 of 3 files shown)


llFreetimell:
More errors can be found. If then, they will be added here.

llFreetimell:
@mhs4670go 

Let's suppose that `nodeA` has `origin1`.
Then `add_origin(nodeA, origin2)` is accepted as following options for now...
- `add_origin(nodeA, origin2)` = origin(nodeA) + origin2 = origin1 + origin2
- `add_origin(nodeA, origin2)` = origin2
I implemented `add_origin` as latter, but some of `luci/pass` are using like former.
We need to unify this...

mhs4670go:
@llFreetimell Well, I thought you did it former. Then, how about changing the name to `set_origin` as you said before?

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -141,13 +141,22 @@ namespace luci
 
 bool has_origin(const luci::CircleNode *circle_node)
 {
-  return circle_node->annot<CircleNodeOriginAnnotation>() != nullptr;
+  if (circle_node->annot<CircleNodeOriginAnnotation>() == nullptr)
+    return false;
+  if (circle_node->annot<CircleNodeOriginAnnotation>()->origin() == nullptr)
+    return false;
```
mhs4670go, line 147: Maybe errors wouldn't occur without this line, would they?

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -141,13 +141,22 @@ namespace luci
...
+  return true;
 }
 
 void add_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin)
 {
+  if (origin == nullptr)
+    return;
+
+  auto new_origin = composite_origin({get_origin(circle_node), origin});
```
mhs4670go, line 156: This is not the way you said in the comments. Is it a former?

compiler/luci/profile/src/CircleNodeOrigin.cpp
llFreetimell: Ah, I added this for another purpose but not yet removed :(
ignore this :)

compiler/luci/profile/src/CircleNodeOrigin.cpp
llFreetimell: This draft is for discussing and recording where to fix. So the code is not fixed :)

llFreetimell:
Related PR will be created soon, close!

PR #6266, opened 2021-03-12
[luci/profile] Synchronize meaning and action of add_origin

llFreetimell:
Currently, `add_origin` is executes like setting origin, not adding.
This is dangerous because mismatch between meaning and actual action cause confusion.
This commit will synchronize meaning and action of `add_origin`.

### DIFF
--- a/compiler/luci/profile/src/CircleNodeOrigin.cpp
+++ b/compiler/luci/profile/src/CircleNodeOrigin.cpp
@@ -146,6 +146,8 @@ bool has_origin(const luci::CircleNode *circle_node)
 
 /**
  * @brief 'origin' is added to the existing origin of circle_node.
+ * @note  If 'origin' is nullptr, nothing is changed.
+ *        For more detail, please refer to CompositeOrigin constructor.
  */
 void add_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin)
 {
--- a/compiler/luci/profile/src/CircleNodeOrigin.cpp
+++ b/compiler/luci/profile/src/CircleNodeOrigin.cpp
@@ -144,13 +144,14 @@ bool has_origin(const luci::CircleNode *circle_node)
   return circle_node->annot<CircleNodeOriginAnnotation>() != nullptr;
 }
 
+/**
+ * @brief 'origin' is added to the existing origin of circle_node.
+ */
 void add_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin)
 {
-  if (origin == nullptr)
-    return;
-
+  auto new_origin = composite_origin({get_origin(circle_node), origin});
   circle_node->annot<CircleNodeOriginAnnotation>(nullptr);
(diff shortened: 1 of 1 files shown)


compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -144,13 +144,18 @@ bool has_origin(const luci::CircleNode *circle_node)
   return circle_node->annot<CircleNodeOriginAnnotation>() != nullptr;
 }
 
+/**
+ * @brief 'origin' is added to the existing origin of circle_node.
+ * @note  If 'origin' is nullptr, nothing changed because there is nothing to be added.
```
seanshpark, line 149: Are you going to leave this like this? If so, I sugguest not to. It would be better to fix caller (maybe later) not to call this methof if there is nothing to add (nullptr).
If not, please change this comment and add a TODO for `if (origin == nullptr)` line.

llFreetimell: Now `origin` can be `nullptr` and it does not cause errors because we changed the logic of `add_origin`.
I missed this point, thanks :)

I will remove the comment and `nullptr` check...!

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -144,13 +144,14 @@ bool has_origin(const luci::CircleNode *circle_node)
...
 
+/**
+ * @brief 'origin' is added to the existing origin of circle_node.
+ */
 void add_origin(luci::CircleNode *circle_node, const std::shared_ptr<CircleNodeOrigin> origin)
 {
-  if (origin == nullptr)
-    return;
-
+  auto new_origin = composite_origin({get_origin(circle_node), origin});
```
seanshpark, line 152: (optional) it would be nice to have an `assert(origin != nullptr)`

llFreetimell: `origin` can be `nullptr` for now :)

```cpp
add_origin(circle_node, nullptr) = get_origin(circle_node)
add_origin(circle_node, originA) = get_origin(circle_node) + originA
```

This is done by following code. `nullptr` is ignored :)
https://github.com/Samsung/ONE/blob/cf3f7ccb945feb43658c8b5c4f15b77a2684feb3/compiler/luci/profile/src/CircleNodeOrigin.cpp#L83-L93

seanshpark (approved):
LGTM

compiler/luci/profile/src/CircleNodeOrigin.cpp
seanshpark: > origin can be nullptr for now :)

Um... from the code in this function (not looking into `CompositeOrigin`) this seems implicit feature.
Could you add an comment that `origin` can be nullptr and it there will be no changes then?

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6268, opened 2021-03-12
[luci/pass] Add and fix origin annotation

llFreetimell:
Currently, some of `luci/pass` are not doing `add_origin` even they should do.
And some of `luci/pass` are doing wrong `add_origin`.
This commit will add and fix them.

### DIFF
--- a/compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
+++ b/compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
@@ -466,10 +466,12 @@ bool fuse_mul_with_conv(luci::CircleMul *mul)
 
       // Update origin
       // TODO need to remove const
-      luci::add_origin(const_cast<luci::CircleConv2D *>(conv), luci::get_origin(mul));
+      luci::add_origin(const_cast<luci::CircleConv2D *>(conv),
+                       luci::get_origin(loco::must_cast<luci::CircleNode *>(s)));
     }
 
     loco::replace(mul).with(pred_node);
+    luci::add_origin(pred_node, get_origin(mul));
     relu->features(pred_node);
 
     mul->drop();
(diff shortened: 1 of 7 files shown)


compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -466,10 +466,12 @@ bool fuse_mul_with_conv(luci::CircleMul *mul)
 
       // Update origin
       // TODO need to remove const
-      luci::add_origin(const_cast<luci::CircleConv2D *>(conv), luci::get_origin(mul));
+      luci::add_origin(const_cast<luci::CircleConv2D *>(conv),
+                       luci::get_origin(loco::must_cast<luci::CircleNode *>(s)));
```
llFreetimell, line 470: @jinevening Can you check this...?

jinevening: I think the previous code is right, because `mul` is fused to `conv`.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
```diff
@@ -466,10 +466,12 @@ bool fuse_mul_with_conv(luci::CircleMul *mul)
...
     }
 
     loco::replace(mul).with(pred_node);
+    luci::add_origin(pred_node, get_origin(mul));
```
jinevening, line 474: I think this should be removed.

compiler/luci/pass/src/FusePreActivationBatchNormPass.cpp
llFreetimell: I thought that ADD and MUL is fused to preceding CONV but it was not.
Thanks for checking!

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6231, opened 2021-03-08
DRAFT: Compiler FE: Support model partition

seanshpark:
on-going draft to support model partition in compiler frontend.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #6230, opened 2021-03-08
Compiler FE: Model partitioning


seanshpark:
As we support several backends and we can think of this in the frontend,
- one of the backend maybe need off-line code generation 
- we can assign specific operator to specific backend
- we can literally split the model into several sub models for the backends

Currently I'm not clear of this task. Let's do some prototyping and see...

seanshpark:
Let's begin with a simple Model
- Source Model: Input - Sin - Cos - Output
- Target Model: (1) Input - Sin - Output , (2) Input - Cos - Output

With this, I could prepare/build-up necessary methods for basic tasks.

seanshpark:
For the first step, lets assume model has only one subgraph with above `Input - Sqrt- Rsqrt- Output` network.
Then we can use `IF` or `WHILE` for multiple subgraphs.

seanshpark:
https://github.com/Samsung/ONE/pull/6231#issuecomment-799037972 commit stores to two circle files like
![image](https://user-images.githubusercontent.com/4616940/111093883-5053af80-857d-11eb-818e-8f017e0825d9.png)

seanshpark:
Let's now try to group graphs with same kind of group(backend).

Cases of source circle model graph 
- graph input: single or multiple
- graph output: single or multiple
- node output: used or unused

seanshpark:
Things to do from 10,000 feet
- [ ] save connection information file
- [ ] check with a model that may produce recursive situation
- [ ] test framework with running the source model vs produced models
- [ ] check for repetitive things (per Op) and find alternative way

seanshpark:
> check with a model that may produce recursive situation

- `Rsqrt` - `Rsqrt` sequnce should have one input and two outputs: one from first `Rsqrt` and another from second `Rsqrt`
- need to fix this

seanshpark:
> todo : Remove unused outputs

Let's think this with `save connection information file` together...

seanshpark:
@hseok-oh wrote https://github.com/Samsung/ONE/pull/6231#issuecomment-800762497

My draft based on partitioned model: `MANIFEST` and `Net_Sqrt_Rsqrt_001.part.config `(json)

`MANIFEST`
```json
{ 
  "major-version" : "1",
  "minor-version" : "1",
  "patch-version" : "0",
[...]

PR #6270, opened 2021-03-15
[luci] opcode_name of partition

seanshpark:
This will introduce first luci parition change with a method
to get string name of OpCode of circle format.

### DIFF
--- a/compiler/luci/partition/src/CircleOpCode.cpp
+++ b/compiler/luci/partition/src/CircleOpCode.cpp
@@ -34,7 +34,7 @@ public:
   BuiltinOperator visit(const CircleRsqrt *) final { return BuiltinOperator_RSQRT; }
   BuiltinOperator visit(const CircleSqrt *) final { return BuiltinOperator_SQRT; }
 
-  // NOTE only buildin operators should be called (NOT virtual nodes)
+  // NOTE only builtin operators should be called (NOT virtual nodes)
 };
 
 } // namespace
--- a/compiler/luci/partition/src/CircleOpCode.cpp
+++ b/compiler/luci/partition/src/CircleOpCode.cpp
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "CircleOpCode.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleNodeVisitor.h>
+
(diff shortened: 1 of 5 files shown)


seanshpark:
For #6230, from draft #6231

compiler/luci/partition/src/CircleOpCode.cpp
```diff
@@ -0,0 +1,53 @@
...
+using namespace circle;
+
+class QueryOpCode final : public CircleNodeVisitor<BuiltinOperator>
+{
+public:
+  // TODO add CircleNodes
+  BuiltinOperator visit(const CircleRsqrt *) final { return BuiltinOperator_RSQRT; }
+  BuiltinOperator visit(const CircleSqrt *) final { return BuiltinOperator_SQRT; }
+
+  // NOTE only buildin operators should be called (NOT virtual nodes)
```
jinevening, line 37:
```suggestion
  // NOTE only builtin operators should be called (NOT virtual nodes)
```

compiler/luci/partition/src/CircleOpCode.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+using namespace luci;
+using namespace circle;
+
+class QueryOpCode final : public CircleNodeVisitor<BuiltinOperator>
+{
+public:
+  // TODO add CircleNodes
+  BuiltinOperator visit(const CircleRsqrt *) final { return BuiltinOperator_RSQRT; }
+  BuiltinOperator visit(const CircleSqrt *) final { return BuiltinOperator_SQRT; }
```
jinevening, line 35: (optional) It would be good to make this not require operator-level implementation.

For example,
```cpp
const std::string toString(luci::CircleOpcode opcode)
{
  static const char *names[] = {
#define CIRCLE_NODE(OPCODE, CIRCLE_CLASS) #CIRCLE_CLASS,
#include <luci/IR/CircleNodes.lst>
#undef CIRCLE_NODE
  };
  // Returned string is the substring of circle class name ("luci::Circle" is sliced out)
  return std::string(names[static_cast<int>(opcode)]).substr(12);
}
```

jinevening (approved):
LGTM

compiler/luci/partition/src/CircleOpCode.cpp
seanshpark: `circle::EnumNameBuiltinOperator()` already implements `toString()` :)

seanshpark: `BuiltinOperator visit(const CircleXXX *)` is added individually for now and later can be refactored to shorter code...

llFreetimell: We already have `luci::CircleOpcode` which can be accessed as `circle_node.opcode()`
Is there any reason not to use it?
I think dependency with `BuiltinOperator_*` should be avoided if possible...!

llFreetimell: In draft, I just found that `luci::opcode_name` is just used for key.
So my point was "Can't we use `luci::CircleOpcode` as key?" :)

seanshpark: Event I've make `luci::CircleOpcode`, forgot about this !
Thanks to remind me. I'll revise this PR :)

seanshpark: I've found that KEY is from the USER as a TEXT STRING... in the `circle-partitioner` code... -_-;;;

llFreetimell (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6271, opened 2021-03-16
[onert] Fix deallocation bug

wateret:
Fix deallocation bug that was introduced by #5926 .

As there was no way to deallocate a dynamic tensor memory of tensors
defined by other backends, dynamic tensor deallocation part needs to
be changed.

- Do not deallocate in FunctionSequence
- Introduce `ITensor::deallocBuffer` which deallocates the buffer
- Introduce `DeallocFunction` class for deallocation of tensors that
  are no longer used(only for Linear Executor)
- `IDynamicTensorManager`'s method `planDealloc` and `deallocInput` are
  no longer used

Plus, it now deallocates all tensors correctly, even between backend
boundaries.

### DIFF
--- a/runtime/onert/core/src/exec/DynamicShapeInferer.cc
+++ b/runtime/onert/core/src/exec/DynamicShapeInferer.cc
@@ -48,11 +48,11 @@ void DynamicShapeInferer::handleBinaryArithmeticOp(const ir::Operation &op,
 
     So, only when all inputs are static, we can skip dynamic shape inference.
   */
-  if ((!lhs->is_dynamic()) && (!rhs->is_dynamic()))
-    return;
-
   auto output_idx = op.getOutputs().at(0);
   auto output = _tensor_registry->getITensor(output_idx);
+
+  if ((previously_static(lhs) && previously_static(rhs)) && currently_static(output))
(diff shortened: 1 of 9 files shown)


wateret:
Refined version of #6264

runtime/onert/core/src/exec/DynamicShapeInferer.cc
```diff
@@ -48,12 +48,12 @@ void DynamicShapeInferer::handleBinaryArithmeticOp(const ir::Operation &op,
 
     So, only when all inputs are static, we can skip dynamic shape inference.
   */
-  if ((!lhs->is_dynamic()) && (!rhs->is_dynamic()))
-    return;
-
   auto output_idx = op.getOutputs().at(0);
   auto output = _tensor_registry->getITensor(output_idx);
 
+  if ((previously_static(lhs) && previously_static(rhs)) && currently_static(output))
```
hyunsik-yoon, line 54:
```suggestion
  if ((currently_static(lhs) && currently_static(rhs)) && previously_static(output))
```

hyunsik-yoon (approved):
LGTM

PR #6277, opened 2021-03-17
[onert] Deref LoweredGraph operand data reference

wateret:
After splitting partial graphs, LoweredGraph should not have reference
of `Data` which prevents unnecessary const data from being freed.

### DIFF
--- a/runtime/onert/core/src/compiler/ExecutorFactory.cc
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.cc
@@ -119,8 +119,7 @@ void initializeSubgraphIOTensors(compiler::LoweredGraph &lowered_graph,
   }
 }
 
-backend::BackendContexts createBackendContexts(const compiler::LoweredGraph &lgraph,
-                                               bool linear_executor)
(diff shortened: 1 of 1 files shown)


runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -138,26 +137,26 @@ backend::BackendContexts createBackendContexts(const compiler::LoweredGraph &lgr
...
+    const auto backend = def_factor.backend();
+    auto &partial_graph = *context_data_map[backend].graph;
+    auto &operand_layouts = context_data_map[backend].operand_layouts;
+    assert(operand_layouts.find(operand_ind) == operand_layouts.end());
+    operand_layouts[operand_ind] = def_factor.layout();
+
+    // Copy the operand and insert it to the partial graph
+    auto new_operand = std::make_unique<ir::Operand>(operand);
+    new_operand->clearDefUse();
+    operand.releaseData(); // Deref data of LoweredGraph
```
hyunsik-yoon, line 155: @wateret mentioned that this line is the actual change of this commit.

hyunsik-yoon (approved):

PR #6278, opened 2021-03-17
[onert] Remove IDynamicTensorManager

wateret:
As we do not use `planDealloc` and `deallocInput` of
`IDynamicTensorManager` any longer, we do not need this class.

cpu backend's `DynamicTensorManger` still exists but does not derive
from `IDynamicTensorManager`.


hseok-oh (approved):
LGTM

hyunsik-yoon (approved):

PR #6283, opened 2021-03-18
[one-cmds] Fix one-codegen bug

mhs4670go:
This commit fixes one-codegen bug.


mhs4670go:
Hmmm.. I missed this bug even though related test(`one-codegen_001.test`) exists. Maybe we need to integrate release test into CI later.

seanshpark:
> Maybe we need to integrate release test into CI later.

one-cmd tests are done daily not per commit :)

seanshpark (approved):
LGTM

PR #6287, opened 2021-03-18
[onert] Remove ParamChecker

hseok-oh:
This commit removes ParamChecker.
ParamChecker was introduced to check parameter's compilability, but it is not used.
And this commit removes checkCompilable() method because it is just check disable_compile flag.


glistening (approved):
LGTM

PR #6308, opened 2021-03-22
[luci] Remove unused loco headers

llFreetimell:
Some of loco related headers are not used but still alive.
This commit will remove them.


seanshpark (approved):
LGTM

PR #6309, opened 2021-03-22
[luci] Do not rename output tensor name

seanshpark:
This will fix not to rename output tensor name.

### DIFF
--- a/compiler/luci/import/src/Importer.cpp
+++ b/compiler/luci/import/src/Importer.cpp
@@ -187,7 +187,8 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
     // set the graph output name and node object
     auto graph_output = graph->outputs()->create();
     std::string tname = luci::tensor_name(tensor);
-    graph_output->name("output_" + tname);
+    assert(tname.length() > 0);
+    graph_output->name(tname);
 
     luci::copy_tensor_attributes(tensor, output_node);
 


seanshpark:
For #6230

mhs4670go (approved):
Looks good to me.

Just curiosity, I tried to found why this PR was needed but failed. Could you explain that a bit? Actually, I have two questions.

1. No extra string is added to input's. It seems there might be a reason that `output_` is added to output's name in the first place when you worked on name-related issue, doesn't it?

2. > I tried to found why this PR was needed but failed. Could you explain that a bit?

seanshpark:
1) When I first wrote this, name was not important and had no policy and named it without any deep thoughts, and just named it like as.

2) adding `output_` suffix alters name and this makes a different name to partitioned graph output.
this is before some result that I'm working:
```json
$ cat Part_Sqrt_Rsqrt_000/Part_Sqrt_Rsqrt_000.json
{
  "source" : {
    "file" : "Part_Sqrt_Rsqrt_000.circle",
    "inputs" : [ "ifm" ],
    "outputs" : [ "output_ofm" ] <---- THIS ONE
  },
  "parts" : [
    {
      "file" : "Part_Sqrt_Rsqrt_000.00001_acl_cl.circle",
      "inputs" : [ "ifm" ],
      "outputs" : [ "sqrt" ]
    },
    {
      "file" : "Part_Sqrt_Rsqrt_000.00002_cpu.circle",
      "inputs" : [ "sqrt" ],
      "outputs" : [ "ofm" ] <-- THIS SHOULD BE SAME AS "source"
    }
  ]
}
```

and this is after
```json
$ cat Part_Sqrt_Rsqrt_000/Part_Sqrt_Rsqrt_000.json
{
  "source" : {
    "file" : "Part_Sqrt_Rsqrt_000.circle",
    "inputs" : [ "ifm" ],
    "outputs" : [ "ofm" ] <--- NOW THIS ONE
  },
  "parts" : [
    {
      "file" : "Part_Sqrt_Rsqrt_000.00001_acl_cl.circle",
      "inputs" : [ "ifm" ],
      "outputs" : [ "sqrt" ]
    },
    {
      "file" : "Part_Sqrt_Rsqrt_000.00002_cpu.circle",
      "inputs" : [ "sqrt" ],
      "outputs" : [ "ofm" ]
    }
  ]
}
```

PR #6311, opened 2021-03-22
[luci] Remove unused loco headers - 2

llFreetimell:
This commit removes unused loco related headres in `luci`.


seanshpark (approved):
LGTM

PR #6319, opened 2021-03-23
[luci] Add helper methods for PGraphs

seanshpark:
This will add some helper methods for PGraphs.

### DIFF
--- a/compiler/luci/partition/src/PartitionIR.cpp
+++ b/compiler/luci/partition/src/PartitionIR.cpp
@@ -75,11 +75,11 @@ std::string PGraphs::group_of(luci::CircleNode *node) const
 
   auto it = node2group.find(node);
   if (it == node2group.end())
+  {
     INFO(l) << "PGraphs::group_of " << node << "(" << node->name() << ") not found" << std::endl;
-  assert(it != node2group.end());
-  auto group = it->second;
-  assert(!group.empty());
-  return group;
+    return "";
+  }
+  return it->second;
 }
 
 const PGraph *PGraphs::pgraph_of(luci::CircleNode *node) const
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6230, from draft #6231

compiler/luci/partition/src/PartitionIR.cpp
```diff
@@ -0,0 +1,101 @@
...
+{
+  assert(node != nullptr);
+
+  LOGGER(l);
+
+  auto it = node2group.find(node);
+  if (it == node2group.end())
+    INFO(l) << "PGraphs::group_of " << node << "(" << node->name() << ") not found" << std::endl;
+  assert(it != node2group.end());
+  auto group = it->second;
```
mhs4670go, line 80: 1. You intended to just be dereferenced even if `it` is `node2group.end()` in release mode?
2. There is no information about `interest` type here. And neither is the logic that returns a default group when `node` is not `interest` type. Does that logic exist at the place where `node2group` is made?

compiler/luci/partition/src/PartitionIR.h
```diff
@@ -65,8 +65,27 @@ struct PGraphs
...
+   * @note  node maybe one of NOT interest type(like CircleInput), there will be
+   *        no pgraph holding it.
+   */
+  const PGraph *pgraph_of(luci::CircleNode *node) const;
 };
 
 } // namespace luci
 
-#endif // __LUCI_PARTITON_IR_H__
+#endif // __LUCI_PARTITION_IR_H__
```
seanshpark, line 91: I'll post another PR or this...

compiler/luci/partition/src/PartitionIR.cpp
seanshpark: You are right. There's seems something missing here.
- we should return empty string if not found
- caller will set to default group if empty
- "interest" is ambigious here. I'll revise the comment

mhs4670go (approved):
LGTM

compiler/luci/partition/src/PartitionIR.h
seanshpark: --> #6321

llFreetimell (approved):

Issue #6333, opened 2021-03-26
[circle-quantizer] Introduce verifier after quantization


jinevening:
## What

Let's introduce a verifier to check the type/granularity of quantized operators after quantization.

## Why

To make our quantizer more reliable.

jinevening:
I close this as #6334 was merged.

PR #6334, opened 2021-03-26
[luci] Introduce verifier for quantized model

jinevening:
Introduce verifier to check the type/granularity of quantized model.

---
Related to: https://github.com/Samsung/ONE/issues/6333


jinevening:
**For reviewers**

This PR has verification codes for the below operators.

Conv2D
DepthwiseConv2D
InstanceNorm
PRelu
TransposeConv
FullyConnected
Add
AveragePool2D
MaxPool2D
Mean
Mul
Relu

compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -378,6 +379,11 @@ void CircleOptimizer::quantize(loco::Graph *g) const
     logo::PhaseRunner<logo::PhaseStrategy::Saturate> phase_runner{g};
     phase_runner.attach(&prog);
     phase_runner.run(phase);
+
+    // Verify the type/granularity of the quantized model
+    luci::QuantizedModelVerifyPass verifier(str_to_dtype(output_dtype),
+                                            str_to_granularity(granularity));
+    verifier.run(g);
```
seanshpark, line 386: result is not used. is this OK?

compiler/luci/pass/src/QuantizedModelVerifyPass.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+  // TODO: Implement more Ops
+};
+
+} // namespace
+
+namespace luci
+{
+
+bool QuantizedModelVerifyPass::run(loco::Graph *g)
```
seanshpark, line 431: (optiona) instead of using `Pass` in suffix, how about `QuantizedModelVerifier` ?
as I understand, this uses `Pass` framework but doesn't work like other transoformation passes.
To be strict, it would be bette not to inherit from `logo::Pass`.

compiler/luci/pass/src/QuantizedModelVerifyPass.cpp
```diff
@@ -0,0 +1,455 @@
...
+    if (_granularity == Granularity::LayerWise)
+    {
+      RETURN_FALSE_UNLESS(is_lwq_const(node->filter()))
+      RETURN_FALSE_UNLESS(is_lwq_const(node->bias()))
+    }
+    else if (_granularity == Granularity::ChannelWise)
+    {
+      RETURN_FALSE_UNLESS(is_cwq_const(node->filter(), 3))
+      RETURN_FALSE_UNLESS(is_cwq_const(node->bias(), rank(node->bias()) - 1))
+    }
```
seanshpark, line 112: (optional) how about spliting `Granularity::LayerWise` and `Granularity::ChannelWise` into to separate visitors?

compiler/luci/pass/src/QuantizedModelVerifyPass.cpp
jinevening: I agree that `Pass` is not necessary here. I'll remove `logo::Pass`. Thanks.

compiler/luci/pass/src/CircleOptimizer.cpp
jinevening: Ah.. verifier did not verify anything :frowning_face: I made verifier throw an exception when the verification fails.

seanshpark:
File exist in `pass` as it started as a pass...
But now it's somekind a service ...
How about moving the files into `luci/service` ?

seanshpark:
And if you provide 4 visitors for verification...
- how may CircleNode types should be supported? if you are planning to support all of them then would it be better to split files now or layer?
- if not putting in one file seems ok for now :)

jinevening:
> How about moving the files into luci/service ?

Moving to `luci/service` requires movement of another file (luci/Pass/QuantizationParameters.h). I've added TODO comment for this.

> how may CircleNode types should be supported? if you are planning to support all of them then would it be better to split files now or layer?

I divided visitors to multiple files. PTAL

compiler/luci/pass/src/helpers/VerifyQuantizedNodeChannelWiseGranularity.h
```diff
@@ -0,0 +1,102 @@
...
+// This macro is undef at the end of the file
+#define RETURN_FALSE_UNLESS(ARG) \
+  if (not(ARG))                  \
+  {                              \
+    return false;                \
+  }
+
+namespace luci
+{
+
```
seanshpark, line 33: could you please remove this line, and also for other files :)

compiler/luci/pass/src/QuantizedModelVerifier.cpp
```diff
@@ -0,0 +1,76 @@
...
+      VerifyQuantizedNodeChannelWiseGranularity vg;
+      if (!circle_node->accept(&vg))
+        throw std::runtime_error("Wrong granularity");
+    }
+  }
+}
+
+} // namespace luci
+
+#undef RETURN_FALSE_UNLESS
```
seanshpark, line 76: I think this is needless

compiler/luci/pass/src/helpers/VerifyQuantizedNodeChannelWiseGranularity.h
```diff
@@ -0,0 +1,101 @@
...
+// This macro is undef at the end of the file
+#define RETURN_FALSE_UNLESS(ARG) \
+  if (not(ARG))                  \
+  {                              \
+    return false;                \
+  }
+
+namespace luci
+{
+namespace verify_quantization
```
seanshpark, line 33: (optional) `verify_quantization` is a bit long. short would be better... 
or, do we need this inner namespace? class begins with `Verify` so I think it's not necessary. some global methos seems better with inner name space...

jinevening: I agree. I moved the methods into the classes and removed the long namespace.

compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
```diff
@@ -0,0 +1,121 @@
...
+
+struct VerifyQuantizedNodeChannelWiseGranularity final : public luci::CircleNodeVisitor<bool>
+{
+private:
+  uint32_t rank(const loco::Node *node)
+  {
+    auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
+    return circle_node->rank();
+  }
+  bool is_cwq_const(const loco::Node *node, uint32_t channel_dim)
```
seanshpark, line 43: just one more little thing, an empty space above this line :)

seanshpark (approved):
LGTM thank you!

PR #6340, opened 2021-03-29
[luci] Fix Partition namespace

seanshpark:
This will remove unnecessary anonymous namespace.


seanshpark:
Current CMakeLists.txt doesn't include partition as a subproject, thus didn't build this module.
I'll post this fix in anothor PR.

mhs4670go (approved):
LGTM

PR #6342, opened 2021-03-29
[luci] Add default_group to PGroups

seanshpark:
This will add default_group member for default group in PGroups.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6345, opened 2021-03-29
[circle-partitioner] Introduce HelperPath

seanshpark:
This will introduce HelperPath with path related helper functions.

### DIFF
--- a/compiler/circle-partitioner/src/HelperPath.cpp
+++ b/compiler/circle-partitioner/src/HelperPath.cpp
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6230, from draft #6231

compiler/circle-partitioner/src/HelperPath.cpp
```diff
@@ -0,0 +1,63 @@
...
+  return base.substr(pos + 1);
+}
+
+std::string make_path(const std::string &base, const std::string &input, uint32_t idx,
+                      const std::string &backend)
+{
+  auto filename_ext = get_filename_ext(input);
+
+  auto pos = filename_ext.find_last_of(".");
+  assert(pos > 0);
```
mhs4670go, line 50: This function assumes length of pos is always positive?

seanshpark: This line seems strange. Thanks for point out. I think some revisition is required...

mhs4670go (approved):
LGTM

PR #6346, opened 2021-03-29
[circle-partitioner] Introduce HelperStrings

seanshpark:
This will introduce HelperStrings with string related helper functions.

### DIFF
--- a/compiler/circle-partitioner/src/HelperStrings.cpp
+++ b/compiler/circle-partitioner/src/HelperStrings.cpp
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6230, from draft #6231

compiler/circle-partitioner/src/HelperStrings.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+bool is_one_of(const std::string item, const std::vector<std::string> items)
+{
+  for (auto &find : items)
+  {
+    if (find == item)
+      return true;
+  }
+  return false;
```
mhs4670go, line 43:
```suggestion
  return std::find(items.begin(), items.end(), item) != items.end();
```

mhs4670go (approved):
LGTM

PR #6347, opened 2021-03-29
[crew] Introduce initial crew library

seanshpark:
This will introduce initial crew library that will provide
config ini/json file import/export functionalities.

### DIFF
--- a/compiler/crew/README.md
+++ b/compiler/crew/README.md
@@ -1,0 +1,13 @@
+# crew
+
+_crew_ is circle partitioning Configuration REader and Writer library.
+
+### Support formats
+
+Current _crew_ supports below format and functionalities.
+- INI read
+- INI write
+- JSON write
+
+_crew_ supports limited portion of JSON and INI formats just enough to access
+circle partition configuration files.
--- a/compiler/crew/CMakeLists.txt
+++ b/compiler/crew/CMakeLists.txt
@@ -1,0 +1,20 @@
+file(GLOB_RECURSE SOURCES "src/*.cpp")
+file(GLOB_RECURSE TESTS "src/*.test.cpp")
+list(REMOVE_ITEM SOURCES ${TESTS})
+
+add_library(crew STATIC ${SOURCES})
+target_include_directories(crew PRIVATE src)
+target_include_directories(crew PUBLIC include)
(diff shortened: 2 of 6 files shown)


seanshpark:
For #6230, from draft #6231

compiler/crew/README.md
```diff
@@ -0,0 +1,13 @@
+# crew
+
+_crew_ is circle partitioning Configuration REader and Writer library.
```
mhs4670go, line 3:
```suggestion
_crew_ is circle partitioning Configuration Reader and Writer library.
```

seanshpark: `RE` is intentionally set capital for `cREw`

seanshpark:
I need to sync with draft: there are some changes...

compiler/crew/src/ConfigIni.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+  assert(data != nullptr);
+  assert(length > 0);
+
+  auto buffer = std::vector<char>();
+  buffer.reserve(length + 1);
+  char *pbuffer = buffer.data();
+  memcpy(pbuffer, data, length);
+  // add null at end to be sure
+  *(pbuffer + length) = 0;
```
mhs4670go, line 39:
```suggestion
  std::vector<char> buffer(data, data + length);
  buffer.push_back('\0');
  char *pbuffer = buffer.data();
```

seanshpark: This doesn compile: `error: no matching function for call to ‘std::vector<char>::vector(const char*&, size_t&)’`

seanshpark:
I've renamed `ConfigIni` to `PConfigIni` as it's for `Partition Config`

compiler/crew/src/ConfigIni.cpp
mhs4670go: @seanshpark Hmmm.. that's weird. It can be compiled in my local pc.

mhs4670go (approved):
LGTM

compiler/crew/src/ConfigIni.cpp
seanshpark: Maybe I have some other problems. If you like, please post a revision after this lands...

PR #6348, opened 2021-03-29
[luci] Introduce PartitionPGroups

seanshpark:
This will introduce PartitionPGroups with produce_pgroups method that
will produce luci::PGroups from luci::Module and luci::PartitionTable.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6349, opened 2021-03-29
[luci] Call produce_pgroups in apply

seanshpark:
This will revise partition apply method to produce pgroups and dump
pgroups object.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6350, opened 2021-03-29
[luci] Introduce PartitionMerge

seanshpark:
This will introduce PartitionMerge that provides merge_pgroups method to
merge PGroup in PGroups with certain conditions.

### DIFF
--- a/compiler/luci/partition/src/PartitionMerge.cpp
+++ b/compiler/luci/partition/src/PartitionMerge.cpp
@@ -1,0 +1,201 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "PartitionMerge.h"
+
+#include <algorithm>
+
+namespace
+{
+
+/**
+ * @brief return true of pgroup_i output is one of the inputs of pgroup
+ */
+bool is_input_of(const luci::PGroup *pgroup_i, const luci::PGroup *pgroup)
+{
+  for (auto *output : pgroup_i->outputs)
+  {
+    for (auto *input : pgroup->inputs)
+    {
+      if (input == output)
+        return true;
+    }
+  }
+  return false;
+}
+
+/**
+ * @brief return true if there is only one input or all the inputs have same group
+ * @note  pgroups is used to find group of pgroup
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6230, from draft #6231

seanshpark:
Current condition of merge still has problems. Still work in progress...

compiler/luci/partition/src/PartitionMerge.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+#include "PartitionMerge.h"
+
+#include <algorithm>
+
+namespace
+{
+
+/**
+ * @brief return true of pgroup_i output is one of the inputs of pgroup
```
mhs4670go, line 25:
```suggestion
 * @brief return true if pgroup_i output is one of the inputs of pgroup
```

compiler/luci/partition/src/PartitionMerge.cpp
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/**
+ * @brief return true if there is only one input or all the inputs have same group
+ * @note  pgroups is used to find group of pgroup
+ */
+bool is_input_same(const luci::PGroup *pgroup, const luci::PGroups *pgroups)
+{
+  assert(pgroups != nullptr);
+  assert(pgroup != nullptr);
```
mhs4670go, line 47: maybe just returning false is better?

seanshpark: usually I add assert for the condition of the method. this assert is to check the condition. the caller should give correct condition.

compiler/luci/partition/src/PartitionMerge.cpp
```diff
@@ -0,0 +1,201 @@
...
+  const luci::PGroup *input_pgroup = nullptr;
+  std::string group;
+  for (auto &input : pgroup->inputs)
+  {
+    auto input_group = pgroups->group_of(input);
+    // NOTE: all the nodes should be registered and return should be valid group.
+    // convert_to_proups() should ensure this.
+    // assert here to find if there is any problem with this.
+    assert(not input_group.empty());
+    if (input_group.empty())
```
mhs4670go, line 58: Maybe this `if` condition would never be called.

seanshpark: As added a comment, this is a guard to find if any problems.
`if` itself is for release build.

compiler/luci/partition/src/PartitionMerge.cpp
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * @brief merge pgroup into pgroup_i
+ * @note  output of pgroup_i should be input of pgroup
+ */
+void merge_into(luci::PGroup *pgroup, luci::PGroup *pgroup_i)
+{
+  for (auto &pnode : pgroup->pnodes)
+  {
+    // update pgroup for this pnode
+    pnode->pgroup = pgroup_i;
```
mhs4670go, line 95: Just curiosity, do they have to have the same group?

seanshpark: No. I the beginning, every pnode(which holds node) reside in it's own pgroup. that is one pnode per one pgroup.
`Merge` is to merge two pgroup if condition is met.

mhs4670go (approved):
LGTM

PR #6353, opened 2021-03-29
[crew] Introduce PConfigJson

seanshpark:
This will introduce PConfigJson for export to json format.

### DIFF
--- a/compiler/crew/src/PConfigJson.cpp
+++ b/compiler/crew/src/PConfigJson.cpp
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


seanshpark:
For #6230, from draft #6231

compiler/crew/src/PConfigJson.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+} // namespace
+
+namespace crew
+{
+
+void JsonExport::indent(void)
+{
+  for (uint32_t i = 0; i < _indent; ++i)
+    _os << "  ";
```
mhs4670go, line 37: You meant to indent two spaces per indent rather than per one spaces? And there seems no way to change `_indent` value.

seanshpark: > You meant to indent two spaces per indent rather than per one spaces?

Yes;

> And there seems no way to change _indent value.

Yes, that is current implementation.

mhs4670go (approved):
LGTM

PR #6355, opened 2021-03-30
[luci] Revise partition merge with inputs

seanshpark:
This will revise partition merge of two pgroup not to add same input.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6356, opened 2021-03-30
[luci] Revise partition check_allocate_partition

seanshpark:
This will revise check_allocate_partition in partition to skip
CircleConst for making PNode and so on.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6358, opened 2021-03-30
[luci] Call merge_groups in partition apply

seanshpark:
This will revise apply to call merge_groups to merge PGroup for some
condition.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6359, opened 2021-03-30
[luci] Introduce cleanup of unused inputs/outputs

seanshpark:
This will introduce remove_unused_inputoutputs method that removes
unused inputs/outputs in all PGroup of PGroups.

### DIFF
--- a/compiler/luci/partition/src/PartitionCleanup.cpp
+++ b/compiler/luci/partition/src/PartitionCleanup.cpp
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6230, from draft #6231

compiler/luci/partition/src/PartitionCleanup.cpp
```diff
@@ -0,0 +1,139 @@
...
+  for (auto &pgroup : pgroups->pgroups)
+  {
+    bool changed;
+    do
+    {
+      changed = false;
+      for (auto it = pgroup->inputs.begin(); it != pgroup->inputs.end(); ++it)
+      {
+        auto input = *it;
+        if (pgroups->pgroup_of(input) == pgroup.get())
```
mhs4670go, line 89: It's a little bit hard to understand. 

`input` is from `pgroup`. So, `pgroups_of(input)` returns `pgroup`. Then, `== pgroup.get()` condition seems always true. Anything wrong?

seanshpark: `pgroup_of(input)` returns `pgroup` pointer of `input` which is `CircleNode`.

here, `input` is input of `pgroup` which may be in another `pgroup`
when `pgroup` is created, there is one `pnode` having `node`
the `input` is the input of this `node`, which will be in another `pgroup`

when two `pgroup` merges, first all the `input` are copied.
and remove not effective `input`.

this line is to check if the `input` is pointing to the merged `pgroup`.
before merge `pgroup` may have been another object.
if `pgroup` of `input` is same as this `pgroup` then we can think that
input is pointing to itself so we can remove this `input` from the `inputs` list.

mhs4670go (approved):
LGTM

PR #6364, opened 2021-03-30
[luci] Produce PModules from PGroups

seanshpark:
This will introduce produce_pmodules method to produce PModules from
PGroups that we can store to circle files.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6369, opened 2021-03-31
[circle-quantizer] Verify the granularity of node and its inputs

jinevening:
This verifies the granularity of the node and its inputs.

### DIFF
--- a/compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
+++ b/compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
@@ -32,9 +32,30 @@ using Granularity = luci::QuantizationGranularity;
 namespace luci
 {
 
+/**
+ * @brief Verify the granualrity of channel-wise quantized node
+ * @details
+ *
+ * Targets to verify
+ * - node's output (i.e., node itself)
+ * - node's inputs
+ */
 struct VerifyQuantizedNodeChannelWiseGranularity final : public luci::CircleNodeVisitor<bool>
 {
 private:
+  bool is_lwq(const loco::Node *node)
+  {
+    auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
+
+    if (circle_node->quantparam()->scale.size() != 1)
+      return false;
+
+    if (circle_node->quantparam()->zerop.size() != 1)
+      return false;
+
+    return true;
+  }
+
   uint32_t rank(const loco::Node *node)
   {
     auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
@@ -60,6 +81,8 @@ private:
 private:
   bool visit(const luci::CircleConv2D *node)
   {
+    RETURN_FALSE_UNLESS(is_lwq(node))
+    RETURN_FALSE_UNLESS(is_lwq(node->input()))
     RETURN_FALSE_UNLESS(is_cwq_const(node->filter(), 0))
     RETURN_FALSE_UNLESS(is_cwq_const(node->bias(), rank(node->bias()) - 1))
     return true;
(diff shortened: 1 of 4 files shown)


jinevening:
**For reviewers**

The output of node itself should be quantized per layer (it should have one scale and one zp). This code checks that.

compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
```diff
@@ -32,9 +32,30 @@ using Granularity = luci::QuantizationGranularity;
...
+ * @details
+ *
+ * Targets to verify
+ * - node's output (i.e., node itself)
+ * - node's inputs
+ */
 struct VerifyQuantizedNodeChannelWiseGranularity final : public luci::CircleNodeVisitor<bool>
 {
 private:
+  bool is_lwq(const loco::Node *node)
```
seanshpark, line 46: We are dealing only with CircleNode in luci. Is there any particular reason to accept loco::Node ?

compiler/luci/pass/src/VerifyQuantizedNodeLayerWiseGranularity.h
```diff
@@ -32,9 +32,30 @@ using Granularity = luci::QuantizationGranularity;
...
+ * @details
+ *
+ * Targets to verify
+ * - node's output (i.e., node itself)
+ * - node's inputs
+ */
 struct VerifyQuantizedNodeLayerWiseGranularity final : public luci::CircleNodeVisitor<bool>
 {
 private:
+  bool is_lwq(const loco::Node *node)
```
seanshpark, line 46: same here

compiler/luci/pass/src/VerifyQuantizedNodeLayerWiseGranularity.h
```diff
@@ -51,33 +72,43 @@ struct VerifyQuantizedNodeLayerWiseGranularity final : public luci::CircleNodeVi
 private:
   bool visit(const luci::CircleConv2D *node)
   {
+    RETURN_FALSE_UNLESS(is_lwq(node))
```
seanshpark, line 75: same here

compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
```diff
@@ -60,33 +81,43 @@ struct VerifyQuantizedNodeChannelWiseGranularity final : public luci::CircleNode
 private:
   bool visit(const luci::CircleConv2D *node)
   {
+    RETURN_FALSE_UNLESS(is_lwq(node))
```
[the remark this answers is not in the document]
jinevening: So far, all operators are checked with is_lwq, but I'm not sure this will be true in the future. More specifically, I'm thinking about boolean type or int type (e.g., argmax).

compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
jinevening: That's because it is more concise.

For example, the current code looks as below.

```cpp
  bool visit(const luci::CircleMul *node)
  {
    RETURN_FALSE_UNLESS(is_lwq(node))
    RETURN_FALSE_UNLESS(is_lwq(node->x()));
    RETURN_FALSE_UNLESS(is_lwq(node->y()));
    return true;
  }
```

If we use `CircleNode` as the argument type, type casting should be inserted to all places where the accessor of CircleNode is called. Like,
 
`RETURN_FALSE_UNLESS(is_lwq(loco::must_cast<const luci::CircleNode>(node->x())));`

seanshpark (approved):
LGTM

PR #6377, opened 2021-04-01
[crew] Introduce read_ini method

seanshpark:
This will introduce read_ini method to read init file and fill PConfig
structure.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6379, opened 2021-04-01
[circle-partitioner] Introduce load_model

seanshpark:
This will introduce load_model method for better readability.


seanshpark:
For #6230, from draft #6231

PR #6382, opened 2021-04-01
[circle-part-driver] Implement load_config

seanshpark:
This will implement load_config that loads config ini file and prepares
models to run list.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6383, opened 2021-04-01
[circle-partitioner] Read partition config ini

seanshpark:
This will enable circle-partitioner to read partition config ini file.

### DIFF
--- a/compiler/circle-partitioner/src/PartitionRead.cpp
+++ b/compiler/circle-partitioner/src/PartitionRead.cpp
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


seanshpark:
For #6230, from draft #6231

compiler/circle-partitioner/src/PartitionRead.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include <luci/Log.h>
+
+#include <stdexcept>
+
+namespace
+{
+
+using namespace partee;
+
+const char *_section_partition = "partition";
```
mhs4670go, line 31: Just curiosity, is there any reason to use `const char *` instead of `std::string`?

mhs4670go (approved):
LGTM

compiler/circle-partitioner/src/PartitionRead.cpp
seanshpark: for simplicity and didn't want to initialize `std::string` in binary load time :)

PR #6386, opened 2021-04-01
[circle-part-driver] Implement load_inputs

seanshpark:
This will implement load_inputs method to load input data from files.

### DIFF
--- a/compiler/circle-part-driver/src/PModelsRunner.h
+++ b/compiler/circle-part-driver/src/PModelsRunner.h
@@ -19,11 +19,16 @@
 
 #include <crew/PConfig.h>
 
+#include <map>
 #include <string>
 #include <vector>
 
 namespace prunner
 {
+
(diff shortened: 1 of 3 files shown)


seanshpark:
For #6230, from draft #6231

compiler/circle-part-driver/src/PModelsRunner.h
```diff
@@ -47,6 +52,7 @@ class PModelsRunner
 private:
   crew::PConfig _pconfig;
   RunModels _models_to_run;
+  Buffers _data_stage;
```
mhs4670go, line 55: maybe..?
```suggestion
  Buffers _data_storage;
```

seanshpark: it's stage, not storatge... :)

mhs4670go (approved):
LGTM

PR #6389, opened 2021-04-01
[crew] Introduce write_json

seanshpark:
This will introduce write_json method to stream out json format from
PConfig object.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

Issue #6367, opened 2021-03-31
[circle-quantizer] Support quantization for more operators


jinevening:
Let's support more operators in circle-quantizer

Q8: q8 quantization logic
Q16: q16 quantization logic
Verifier: Verifies the type/granularity of the quantized Op

### First milestone (4/6)
Operator | Q8 | Q16 | Verifier
-- | -- | -- | --
sigmoid | o | o |  o
softmax | o | o |  o
tanh | o | o | o 
permute (transpose) | o | o | o
reshape | o | o | o
slice | o | o | o 
argmax | o | o | o 

### Second milestone (4/12)
Operator | Q8 | Q16 | Verifier | Assignee
-- | -- | -- | -- | --
floor | o | o | o | @dayo09  
div | o | o | o | @dayo09  
floor_div | o | o | o | @dayo09  
greater | o  | o  | o | @jinevening 
greater_equal | o  | o  | o | @jinevening 
logical_or | o  | o  | o | @jinevening 
not_equal | o | o | o | @jinevening 
concat | o | o | o | @jinevening 
pad | o | o | o | @jinevening 

### Third milestone (4/16)
Operator | Q8 | Q16 | Verifier| Assignee
-- | -- | -- | -- | --
elu  | o | o | o | @dayo09  
rsqrt | o | o | o | @dayo09 
sqrt | o | o | o | @dayo09 
pow | o | o | o |  @dayo09  
resize_bilinear | o | o | o | @dayo09  
resize_nearestneighbor | o | o |  | @dayo09  
yuv_rgb_scale (FC+Reshape) | - | - | - | -
instance_normalize | o | o | o | @jinevening 
local_response_normalize |   |   |  | need discussion
batch_to_space | o  | o  | o | @jinevening 
depth_to_space |  o | o  | o | @jinevening 
space_to_batch | o  | o  | o | @jinevening 
space_to_depth | o  | o  | o | @jinevening 
split | o | o | o | @jinevening 
strided_slice | o  | o  | o | @jinevening 

CC @lemmaa

dayo09:
@jinevening I will start with not_equal 😄

dayo09:
@jinevening It seems compiler/luci/pass/src/test/TestIOGraph.h is staled and moved to compiler/luci/testhelper. https://github.com/Samsung/ONE/pull/6335

May I replace it?

jinevening:
> May I replace it?

Of course :)

seanshpark:
[...]

PR #6390, opened 2021-04-01
[luci] Support quantization of slice

jinevening:
This supports quantization of slice op.

---
Related to: #6367


seanshpark (approved):
LGTM

PR #6392, opened 2021-04-01
[common-artifacts] Enable SquaredDifference related

seanshpark:
This will remove SquaredDifference test from exclude list and also
InstancNorm models.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6395, opened 2021-04-01
DRAFT Relocate CircleNodeClone to service

seanshpark:
Relocate CircleNodeClone to service and rename partition to
CircleNodeConnect.

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #6394, opened 2021-04-01
Compiler FE: CircleNode clone to service


seanshpark:
While working with partitioning, CircleNode cloning was added in `partition` library.
This feature seems like a service of CircleNode, let's relocate this feature.

One thing to care is clone --> graph connection is a different feature but are linked in current `partition` library.

seanshpark:
Related: #6230

seanshpark:
done

PR #6396, opened 2021-04-01
[luci] CircleNodeClone in Service

seanshpark:
This will introduce CircleNodeClone to clone CircleNode in Service


seanshpark:
For #6394, from draft #6395

mhs4670go (approved):
LGTM

PR #6402, opened 2021-04-01
[circle-part-eval] Introduce Net_InstanceNorm part files

seanshpark:
This will introduce Net_InstanceNorm partition files to test with a bit
complicated network of partitioning.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6403, opened 2021-04-02
[luci] Use clone from service and rename

seanshpark:
This has some inevitable mixed changes to refactor CircleNodeClone;
- use CircleNodeClone from luci/Service
- rename to existing CircleNodeClone to CircleNodeConnect
   - as header guard directive is same
- delete unused CircleNodeClone.test.cpp
- add 'deprecated_' prefix to minimize changes


seanshpark:
For #6394, from draft #6395

mhs4670go (approved):
LGTM

Issue #6310, opened 2021-03-22
[one-cmds] Capture segmentation fault


mhs4670go:
## What

Let's capture segmentation faults from the drivers inside `one-cmds`.

## Why

`one-cmds` captures both `stdout` and `stderr`. But when I gave the inputs that causes a seg fault error to the driver, `one-cmds` couldn't catch the seg fault. I think it is because seg faults are printed to its shell instead of program's as seg fault is a kind of signal.

lemmaa:
@mhs4670go , could you give an example of a specific scenario in which seg fault occurs? Along with catching the seg fault, it is also necessary to fix its occurrence scenario.

mhs4670go:
@lemmaa As of now, our backend driver dies when float model is given, and which is already a known issue. Maybe will be fixed soon.

mhs4670go:
#6404 has solved this issue except one limitation. When the program invoked by `one-cmds` occurs a seg fault, `one-cmds` exits with error code without printing what an error is.

I can let users know which error is occurred with using `switch-case` internally but then, it will become platform-dependent because exit codes are different from platform to platform. So, as of now I think we'd better go as it is and it can be enhanced when it should be.

PR #6404, opened 2021-04-02
[one-cmds] missing codes that return exit code

mhs4670go:
This commit adds missing codes that enable them to return exit code.

Related: #6310 


seanshpark (approved):
LGTM

PR #6411, opened 2021-04-05
[circle-quantizer] Support quantization of softmax

jinevening:
This supports quantization of softmax op.

---
Related to: #6367

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -226,36 +226,32 @@ TEST(QuantizedModelVerifierTest, Softmax)
   TEST_WITH_GRAPH(SoftmaxTestGraph, Type::S16, Granularity::ChannelWise);
 }
 
-TEST(QuantizedModelVerifierTest, Softmax_wrong_type_NEG)
+TEST(QuantizedModelVerifierTest, Softmax_wrong_type_U8_LWQ_NEG)
 {
-  {
-    SoftmaxTestGraph g;
-    g.init();
+  SoftmaxTestGraph g;
+  g.init();
 
-    luci::QuantizeWithMinMaxPass pass(Type::FLOAT32, Type::U8, Granularity::LayerWise);
-    pass.run(g.g());
+  luci::QuantizeWithMinMaxPass pass(Type::FLOAT32, Type::U8, Granularity::LayerWise);
+  pass.run(g.g());
 
-    g._softmax->dtype(Type::S16);
+  g._softmax->dtype(Type::S16);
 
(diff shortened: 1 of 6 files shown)


compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -199,6 +219,45 @@ TEST(QuantizedModelVerifierTest, Logistic_wrong_granularity_NEG)
...
+TEST(QuantizedModelVerifierTest, Softmax)
+{
+  TEST_WITH_GRAPH(SoftmaxTestGraph, Type::U8, Granularity::LayerWise);
+  TEST_WITH_GRAPH(SoftmaxTestGraph, Type::U8, Granularity::ChannelWise);
+  TEST_WITH_GRAPH(SoftmaxTestGraph, Type::S16, Granularity::ChannelWise);
+}
+
+TEST(QuantizedModelVerifierTest, Softmax_wrong_type_NEG)
+{
+  {
```
seanshpark, line 231: This brace seems needless :)

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -199,6 +219,45 @@ TEST(QuantizedModelVerifierTest, Logistic_wrong_granularity_NEG)
...
+    g._softmax->dtype(Type::S16);
+
+    luci::QuantizedModelVerifier verifier(Type::U8, Granularity::LayerWise);
+    EXPECT_ANY_THROW(verifier.verify(g.g()));
+  }
+}
+
+TEST(QuantizedModelVerifierTest, Softmax_wrong_granularity_NEG)
+{
+  {
```
seanshpark, line 247: this too

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
jinevening: It is for avoiding name aliasing when adding a new test in the same category (testing wrong type).

Without the brackets, we'll have to use different variable names when adding a new test, for example,

```cpp
// test 1
SoftmaxTestGraph g;
..

// test 2
SoftmaxTestGraph g2;
...
```

seanshpark: (optional) For this case, IMHO, adding one test for each in test case would be better, that is, 
something like, `Softmax_wrong_type_NEG` -> `Softmax_wrong_type_LW_U8_NEG`
I'll wait for your decision :)

jinevening: Dividing the test case looks good. I've updated the code. PTAL

seanshpark (approved):
LGTM thank you!

PR #6412, opened 2021-04-05
[onert] Fix the conditional statement according to the layout in AclT…

YiHyunJin:
…ensorBuilder.h

The if statement and else if statement are the same. Fixed else if statement.


hseok-oh (approved):
LGTM

chunseoklee (approved):

PR #6419, opened 2021-04-05
[circle-partitioner] Fix validate with partitioned modules

seanshpark:
This will fix partitioned modules to validate with results of apply.
- and also add comment for readability.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6421, opened 2021-04-05
[circle-partitioner] Use exit definition

seanshpark:
This will revise to use exit definition instead of numbers.


seanshpark:
For #6230

mhs4670go (approved):
LGTM

PR #6426, opened 2021-04-05
DRAFT: Refactor luci clone

seanshpark:
on-going draft to refactor luci clone.

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #6420, opened 2021-04-05
Compiler FE: Refactor luci clone


seanshpark:
As clone is moved to `luci/service`, let's refactor to support all circle IRs and existing `CircleConst` clone.

seanshpark:
done for now

PR #6427, opened 2021-04-05
[luci] Separate CircleConst clone

seanshpark:
This will separate CircleConst clone to make easy to reuse code with
CircleNodeClone.


seanshpark:
For #6420, from draft #6426

mhs4670go (approved):
LGTM

PR #6429, opened 2021-04-05
[luci] Revise CircleConst clone method

seanshpark:
This will revise CircleConst clone method to use copy_common_attributes
method.


seanshpark:
For #6420, from draft #6426

mhs4670go (approved):
LGTM

Issue #6418, opened 2021-04-05
[arser] compiler error when link with nncc_common


seanshpark:
When using `arser` with `nncc_common`, I get this compilation error

> compiler/arser/include/arser/arser.h:44:18: error: ‘T {anonymous}::lexical_cast(const string&) [with T = bool]’ defined but not used [-Werror=unused-function]
 template <> bool lexical_cast(const std::string &str)
cc1plus: all warnings being treated as errors

seanshpark:
CC @mhs4670go

mhs4670go:
@seanshpark 
> -Wunused-function
Warn whenever a static function is declared but not defined or a non-inline static function is unused. This warning is enabled by -Wall.

This warning is because `lexical_cast` is non-inline static function. Actually, I shouldn't have made it static function because `arser` is header only library. So, I'm gonna move them into `arser::internal` namespace, which will solve this issue.

BTW, I don't think linking `arser` with `nncc_common` is a good idea because `arser` is just an interface library, which means modules who links to `arser` have to link `nncc_common` as well.

seanshpark:
>  I'm gonna move them into arser::internal namespace, which will solve this issue.

Thanks!

> which means modules who links to arser have to link nncc_common as well.

I wanted to use for `nncc_common`  for executable modues that happens to use `arser`  :)

seanshpark:
Build without errors confirmed :)

PR #6430, opened 2021-04-05
[arser] remove unnamed namespace

mhs4670go:
This commit changes unnamed namespace to internal one.

As a header only library, internal functions should be located in `arser::internal` namespace rather than anonymous one.

Related: #6418 


seanshpark (approved):
LGTM thank you!

seanshpark:
This fixes my problem in #6231 :)

PR #6431, opened 2021-04-05
[luci] Relocate CloneNode class

seanshpark:
This will relocate CloneNode class to it's own file and into CircleConst
to reuse methods.


seanshpark:
For #6420, from draft #6426

mhs4670go (approved):
LGTM

PR #6432, opened 2021-04-05
[circle-part-eval] Update eval script with executable

seanshpark:
This will update part_eval_all.sh eval script to be executable.


mhs4670go (approved):
LGTM

PR #6434, opened 2021-04-05
[circle-part-driver] Link with nncc_common

seanshpark:
This will enable circle-part-driver to link with nncc_common.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6435, opened 2021-04-06
[circle-part-eval] Introduce part_eval_one

seanshpark:
This will introduce part_eval_one.py script to run and compare result of
one tflite model and partitioned model(s).


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6436, opened 2021-04-06
[circle-part-eval] Introduce requires.cmake

seanshpark:
This will introduce requires.cmake for circle-part-eval.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6438, opened 2021-04-06
[infra] Update preset for partitioner

seanshpark:
This will update preset to include circle-partitioner and crew library.


seanshpark:
For #6230, from draft #6231

mhs4670go:
@seanshpark 
1. How about creating a new preset? It's been a long time since we've created the preset.
2. Maybe window preset should be checked if it is built successfully. Have you done this?

seanshpark:
> How about creating a new preset? It's been a long time since we've created the preset.

Yes, agree on creating new one :) You may post a PR if you like. After adding new one, we need to update CI release scripts(?) too. I think you know better than me :)

> Have you done this?

Nope :) It would be nice if you could check this, but not mandatory as we don't use this reset...
I just added not to forget changes for the future...

mhs4670go:
@seanshpark Okay, then here's what I am gonna do.
1. Try to create a package in Windows
2. Post a PR that adds a new preset
3. Change default `PRESET` in `infra/packaging/build` (I think this is the only place that should be changed)

seanshpark:
> I think this is the only place that should be changed

@hseok-oh , can you please confirm this?

hseok-oh:
>Change default PRESET in infra/packaging/build (I think this is the only place that should be changed)

>@hseok-oh , can you please confirm this?

Right. CI is using default `PRESET`.

seanshpark:
close as new PRESET posted

PR #6439, opened 2021-04-06
[infra] Revise infra scripts

seanshpark:
This will revise infra scripts to include crew library and
circle-partitioner.
- circle-partitioner is not included into nnpkg_resource as it's not used


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6444, opened 2021-04-06
[circle-part-eval] Add test list

seanshpark:
This will add test list file with items for partition evaluation test.


seanshpark:
For #6230, from draft #6231

mhs4670go (approved):
LGTM

PR #6446, opened 2021-04-06
[common-artifacts] Add intents to cmake

seanshpark:
This will add intentations for readability.


mhs4670go (approved):
LGTM

PR #6447, opened 2021-04-06
[circle-part-eval] Run evaluation

seanshpark:
This will enable evaluation of partition with test list.

### DIFF
--- a/compiler/circle-part-eval/CMakeLists.txt
+++ b/compiler/circle-part-eval/CMakeLists.txt
@@ -6,3 +6,94 @@
 if(NOT ENABLE_TEST)
   return()
 endif(NOT ENABLE_TEST)
+
+get_target_property(ARTIFACTS_BIN_PATH testDataGenerator BINARY_DIR)
+
+unset(RECIPE_LIST)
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6230, from draft #6231

seanshpark:
> 14:32:34 Scanning dependencies of target circle_part_eval_prepare
14:32:34 make[2]: *** No rule to make target 'compiler/common-artifacts/Part_Add_Sub_000.tflite', needed by 'compiler/circle-part-eval/Part_Add_Sub_000/Part_Add_Sub_000.tflite'.  Stop.

@mhs4670go , do you have any idea what might be the problem? my local works OK...

mhs4670go:
@seanshpark As I said in the messenger, I think this error happened because existing design didn't build it with right dependency tree. If `add_dependencies` works, I'll post an issue to discuss about this.

seanshpark:
> add_dependencies(circle_part_eval_prepare common_artifacts_deps)

This was needed :)

mhs4670go (approved):
LGTM

PR #6448, opened 2021-04-06
[common-artifacts] Revise generation dependency

seanshpark:
This will revise common-artifacts input.h5 dependencies to reduce build
actions for each time of project build.

### DIFF
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -228,7 +228,7 @@ foreach(RECIPE IN ITEMS ${RECIPES})
 
   add_custom_command(OUTPUT ${NNPKG_PATH}
     COMMAND ${MODEL2NNPKG} ${MODEL_PATH}
-    DEPENDS ${MODEL2NNPKG} ${MODEL_PATH}
+    DEPENDS ${MODEL2NNPKG}
     COMMENT "Generate ${RECIPE} nnpackage"
   )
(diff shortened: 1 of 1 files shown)


seanshpark:
@mhs4670go , I wanted to make build only changed files.
`MOVE` file made the change of dependency every time so there was build actions in every build although there was no change.
If you had some reason for this, please let me know; maybe we can add a comment about it.

mhs4670go:
@seanshpark 
> If you had some reason for this, please let me know; maybe we can add a comment about it.

Well, it is because I didn't want them to be duplicated. Current outputs are like below.
```bash
$ tree Add_000.opt
Add_000.opt
├── Add_000.opt.circle
└── metadata
    ├── MANIFEST
    └── tc
        ├── expected.h5
        └── input.h5
```
But if this PR is applied,
```bash
$ tree Add_000.opt
Add_000.opt.input.h5 // duplicate
Add_000.opt.expected.h5 // duplicate
Add_000.opt
├── Add_000.opt.circle
└── metadata
    ├── MANIFEST
    └── tc
        ├── expected.h5
        └── input.h5
```

I think I can solve this issue without creating duplicates. If I add 'output directory' option to `testDataGenerator`, there's no need to generate them every time, which will solve this issue. I'll post a related issue and make a PR if you don't mind:)

seanshpark:
> I'll post a related issue and make a PR if you don't mind:)

Thanks, please go ahead :)

seanshpark:
Can close this :)

PR #6450, opened 2021-04-06
[record-minmax] Do not record min/max of bool type

jinevening:
This makes record-minmax not record min/max of bool type.

---
Related to: https://github.com/Samsung/ONE/issues/6367


seanshpark (approved):
LGTM

PR #6456, opened 2021-04-07
[luci] Rename to BATCH_MATMUL

seanshpark:
This will match BATCH_MATMUL like schema file.


seanshpark:
Related with #6230 and #6231

mhs4670go (approved):
LGTM

PR #6460, opened 2021-04-07
[luci] Revise QueryCircleName

seanshpark:
This will revise QueryCircleName to implement nodes with CIRCLE_VNODE.
- PCSTR is removed as it was introduced to shorten the code but not
needed anymore.


seanshpark:
For #6230 , from #6231

mhs4670go (approved):
LGTM

PR #6462, opened 2021-04-07
[circle-partitioner] Link with nncc_common

seanshpark:
This will link with circle-partitioner for strict build.


seanshpark:
For #6230 , from #6231

mhs4670go (approved):
LGTM

PR #6463, opened 2021-04-07
[luci] Introduce validate_module

seanshpark:
This will introduce and invoke validate_module to validate input model before patitioning.


seanshpark:
For #6230 , from #6231

mhs4670go (approved):
LGTM

PR #6465, opened 2021-04-07
[luci] Introduce validate_name for graph

seanshpark:
This will introduce validate_name method for graph to check all nodes
has non-empty name.


seanshpark:
For #6230 , from #6231

mhs4670go (approved):
LGTM

PR #6466, opened 2021-04-07
[circle-partitioner] Validate empty name

seanshpark:
This will call validate_name method to check empty name for all nodes.


seanshpark:
For #6230 , from #6231

mhs4670go (approved):
LGTM

PR #6467, opened 2021-04-07
[luci] Separate ConnectNode for each nodes

seanshpark:
This will introduce ConnectNode visit file for each nodes.


seanshpark:
For #6230 , from #6231

mhs4670go (approved):
LGTM

PR #6469, opened 2021-04-07
[luci] Relocate clone_connect to PartitionPModules

seanshpark:
This will relocate clone_connect method into PartitionPModules as method
itself is now simple.


seanshpark:
For #6230 , from #6231

mhs4670go (approved):
LGTM

PR #6471, opened 2021-04-07
[infra] Revise compiler_modules

seanshpark:
This will revise compiler_modules script to include partitioning tests.

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #6472, opened 2021-04-07
[circle-part-value-test] Rename folder

seanshpark:
This will rename circle-part-eval to circle-part-value-test to match
name for value test project.


seanshpark:
From #6471, found that this is needed...

seanshpark:
CC @hseok-oh

mhs4670go (approved):
LGTM

PR #6473, opened 2021-04-08
[luci] Fix CircleConst service test

seanshpark:
This will rename test to CircleConstTest for CircleConst clone methods test.


seanshpark:
Found this while working on #6230

mhs4670go (approved):
LGTM

PR #6480, opened 2021-04-08
[luci] Revise clone to check FusedActFunc

seanshpark:
This will revise clone methods to check FusedActFunc is undefined.
- it's an error if undefined.


seanshpark:
Related with #6230 and #6231

mhs4670go (approved):
LGTM

PR #6481, opened 2021-04-08
[luci] Add note for clone_node

seanshpark:
This will add a note for clone_node method header.

### DIFF
--- a/compiler/luci/service/include/luci/Service/CircleNodeClone.h
+++ b/compiler/luci/service/include/luci/Service/CircleNodeClone.h
@@ -31,6 +31,7 @@ void copy_common_attributes(const luci::CircleNode *src, luci::CircleNode *dst);
 
 /**
  * @brief Return a new cloned CircleNode object with same attributes value of node to graph.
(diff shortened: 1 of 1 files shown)


seanshpark:
Related with #6230 and #6231

compiler/luci/service/include/luci/Service/CircleNodeClone.h
```diff
@@ -31,6 +31,7 @@ void copy_common_attributes(const luci::CircleNode *src, luci::CircleNode *dst);
 
 /**
  * @brief Return a new cloned CircleNode object with same attributes value of node to graph.
+ * @note  Will return nullptr if clone has failed
```
mhs4670go, line 34: (optional) Maybe there's no need to describe comments with future tense.

mhs4670go (approved):
LGTM

compiler/luci/service/include/luci/Service/CircleNodeClone.h
seanshpark: `will` is used for meaning of intention or design...

PR #6482, opened 2021-04-08
[luci] Extract new_empty_const from test

seanshpark:
This will extract common new_empty_const method from clone tests.


seanshpark:
Related with #6230 and #6231

mhs4670go (approved):
LGTM

PR #6484, opened 2021-04-08
[luci] Check nullptr for Sub clone

seanshpark:
This will add nullptr check for clone node creation of Sub Op.


seanshpark:
Another one that was missed in #6474

mhs4670go (approved):
LGTM

seanshpark:
@mhs4670go , could you please approve again? there was a slight time difference :)

mhs4670go (approved):

PR #6493, opened 2021-04-08
[luci] Add clone_node test for CircleConst

seanshpark:
This will add clone_node method test for CircleConst.


seanshpark:
For #6230 , from draft #6231

mhs4670go (approved):
LGTM

PR #6494, opened 2021-04-08
[luci/pass] Replace TestIOGraph.h

dayo09:
This replaces TestIOGraph.h with testhelper's for unit test files in `compiler/luci/pass`.

ONE-DCO-1.0-Signed-off-by: Dayoung Lee <dayg502@gmail.com>

--------------------

Resolves #6455

PR #6497, opened 2021-04-08
[luci] Revisit testhelper

dayo09:
This commit rewrites init for TestIGraph and TestOGraph.

-----------

Derived from #6494

### DIFF
--- a/compiler/luci/pass/src/RemoveUnnecessarySplitPass.test.cpp
+++ b/compiler/luci/pass/src/RemoveUnnecessarySplitPass.test.cpp
@@ -73,8 +73,8 @@ public:
 public:
   void init()
   {
-    TestIGraphlet::init(g(), {{1}});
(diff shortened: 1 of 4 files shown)


seanshpark:
Code itself looks good!

As I read the changes, there are two different things happen here
1. introduce `init` for `TestIGraphlet` and `TestOGraphlet`
2. revise test codes to use new `init` method

I think it is OK to go as-is but would be better to split in two PRs. Would you like to do so?

dayo09:
@seanshpark 
I would like to. :-D But it's a bit tricky.
As soon as I introduce the init function, they need changes instantly as they are virtual functions. So it looks like as the one change for me.
For sure I'll split them into two if you want. If you don't mind, I want to keep it as it is. 😃

seanshpark (approved):
LGTM

PR #6495, opened 2021-04-08
[luci] Support quantization of LogicalOr

jinevening:
This supports quantization of LogicalOr op.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/6367

PR #6498, opened 2021-04-08
[luci] Revise non-bool type check

jinevening:
This revises non-bool type check to use node type itself.

---
Related to: https://github.com/Samsung/ONE/pull/6495#discussion_r609428735


seanshpark:
> This quantizes non-bool type only.

change itself looks good but this message doesn't seem to explain the change. 
as I understand, this is something like "This revises non-bool type check to use node type itself"...

could you change the message?

dayo09 (approved):
LGTM

seanshpark (approved):
LGTM thank you!

PR #6509, opened 2021-04-09
[luci] Add A Helper To Create Dummy Node

dayo09:
This commit adds a helper function to create dummy node.

-------------------------

For #6367

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -27,6 +27,46 @@ using Granularity = luci::QuantizationGranularity;
 
 namespace
 {
+// helper function to create dummy float const node at the graph
+template<Type T>
+luci::CircleConst *createDummyConst(loco::Graph *g, luci::test::ShapeU32 shape)
+{
+  auto node = g->nodes()->create<luci::CircleConst>();
+  {
+    node->dtype(T);
+    node->shape(shape);
+    node->size<T>(luci::test::num_elements(shape));
+
+    for (int32_t i = 0; i < luci::test::num_elements(shape); i++)
+    {
+      // DESIGN NOTE
+      //
+      // Filling with any random numbers are fine
+      // Q. Should it include minus numbers?
+      switch (T)
+      {
+        case Type::FLOAT32:
+          // Fill with index
+          node->at<T>(i) = static_cast<float>(i);
+          break;
+        case Type::BOOL:
+          // Fill by flip
+          node->at<T>(i) = (i%2) ? true : false;
+          break;
+        case Type::U8:
+          // Fill with index
+          node->at<T>(i) = static_cast<int8_t>(i);
+          break;
+        case Type::S16:
+          // Fill with index
+          node->at<T>(i) = static_cast<uint8_t>(i);
+          break;
+      }
+    }
+  }
+
+  return node;
+}
 
 void insert_scale_zp(luci::CircleNode *node, float scale, int64_t zp)
 {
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -27,6 +27,46 @@ using Granularity = luci::QuantizationGranularity;
 
 namespace
 {
+// helper function to create dummy float const node at the graph
```
seanshpark, line 30: please add a empty line above this

seanshpark: how about using doxygen style
```cxx
/**
  * @brief Helper function to create dummy float const node at the graph
  */
```

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -27,6 +27,46 @@ using Granularity = luci::QuantizationGranularity;
...
+template<Type T>
+luci::CircleConst *createDummyConst(loco::Graph *g, luci::test::ShapeU32 shape)
```
seanshpark, line 32: other methods use snake case; it would be better to follow them: `create_dummy_const`

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -373,6 +406,14 @@ template <class Op> class ComparisonOpTestGraph final : public luci::test::TestI
     EXPECT_ANY_THROW(quantize_and_verify_with_wrong_granularity(&g, type, granularity, node)); \
   } while (0)
 
+// Test Local Helper Function
```
seanshpark, line 409: nit. `Test local helper function` ?

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -27,6 +27,46 @@ using Granularity = luci::QuantizationGranularity;
...
+          // Fill with index
+          node->at<T>(i) = static_cast<float>(i);
+          break;
+        case Type::BOOL:
+          // Fill by flip
+          node->at<T>(i) = (i%2) ? true : false;
+          break;
+        case Type::U8:
+          // Fill with index
+          node->at<T>(i) = static_cast<int8_t>(i);
```
jinevening, line 58:
```suggestion
          node->at<T>(i) = static_cast<uint8_t>(i);
```

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -27,6 +27,46 @@ using Granularity = luci::QuantizationGranularity;
...
+          break;
+        case Type::S16:
+          // Fill with index
+          node->at<T>(i) = static_cast<uint8_t>(i);
```
jinevening, line 62:
```suggestion
          node->at<T>(i) = static_cast<int16_t>(i);
```

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -27,6 +27,46 @@ using Granularity = luci::QuantizationGranularity;
...
+    node->dtype(T);
+    node->shape(shape);
+    node->size<T>(luci::test::num_elements(shape));
+
+    for (int32_t i = 0; i < luci::test::num_elements(shape); i++)
+    {
+      // DESIGN NOTE
+      //
+      // Filling with any random numbers are fine
+      // Q. Should it include minus numbers?
```
jinevening, line 47: I think the current values are ok for now. If we need more complicated tests which require to set specific values, then let's add new API at that time.

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

Issue #6470, opened 2021-04-07
Compiler FE: Test coverage is low


seanshpark:
As of now (Apr 7, 2021 2:27 AM)
```
Packages   97%
Files   98% 
Classes   98% 
Methods   79% 
Lines   82% 
Conditionals   100%
```

Seems cause of latest partition codes...

Lets' raise these values!

seanshpark:
As of April 10, 
```
Packages: 99%   
Files: 99%   
Classes: 99%   
Methods: 80%   
Lines: 87%   
Conditionals: 100% 
```

seanshpark:
Close for now

PR #6510, opened 2021-04-09
[infra] Add circle-tensordump to compiler_modules

seanshpark:
This will add circle-tensordump to compiler_modules for test coverage.
- previous pota-quantization-value-test requires this


seanshpark:
For #6470

PR #6504, opened 2021-04-08
[common-artifacts] Fix generate nnpackage depends

seanshpark:
This will fix cmake depends of generate nnpackage.

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #6515, opened 2021-04-09
[common-artifacts] Prevent cmake from re-running custom command

mhs4670go:
This commit prevents cmake from re-running custom command even though
there's no update.

Related: https://github.com/Samsung/ONE/pull/6504#issuecomment-816348620


seanshpark (approved):
LGTM

Issue #6500, opened 2021-04-08
[luci-interpreter] Add batch_to_space and space_to_batch kernels


jinevening:
## What

Let's add `batch_to_space` and `space_to_batch` kernels

## Why

To support quantization of those Ops. Related to: https://github.com/Samsung/ONE/issues/6367

struss:
Done!

PR #6527, opened 2021-04-09
[luci-interpreter] Introduce BatchToSpaceND kernel

struss:
For #6500

This commit will introduce `BatchToSpaceND` kernel to enable evaluation `CircleBatchToSpaceND`.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/BatchToSpaceND.cpp
+++ b/compiler/luci-interpreter/src/kernels/BatchToSpaceND.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2019 The TensorFlow Authors. All Rights Reserved.
  *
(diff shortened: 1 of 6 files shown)


compiler/luci-interpreter/src/kernels/BatchToSpaceND.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <stdexcept>
+
+namespace luci_interpreter
+{
+
+namespace kernels
+{
+
+const int kInputMinDimensionNum = 3;
+const int kInputMaxDimensionNum = 4;
```
jinevening, line 32: It would be better to place these constants inside anonymous namespace.

struss: Oh, I will fix it.

jinevening (approved):
LGTM Thank you

seanshpark (approved):
LGTM

PR #6533, opened 2021-04-09
[souschef] Introduce INT16 support

binarman:
This commit introduces generators for int16 type.

related issue #6576


compiler/souschef/src/Gaussian.cpp
```diff
@@ -50,22 +50,26 @@ std::vector<uint8_t> GaussianFloat32DataChef::generate(int32_t count) const
   return res;
 }
 
-std::vector<uint8_t> GaussianInt32DataChef::generate(int32_t count) const
+template <typename T> std::vector<uint8_t> GenerateGaussian(int32_t count, float mean, float stddev)
```
seanshpark, line 53: could you please split this change (maybe this change first before adding int16) to another PR?
this seems a bit different context.

binarman: Do you mean split template introduction from actual support?

compiler/souschef/src/Gaussian.cpp
```diff
@@ -50,22 +50,26 @@ std::vector<uint8_t> GaussianFloat32DataChef::generate(int32_t count) const
...
   std::vector<uint8_t> res;
 
+  constexpr float max_cap = std::numeric_limits<T>::max();
+  constexpr float min_cap = std::numeric_limits<T>::min();
   for (uint32_t n = 0; n < count; ++n)
   {
-    auto const value = static_cast<int32_t>(dist(rand));
+    float raw_value = dist(rand);
+    const float capped_value = std::max(min_cap, std::min(max_cap, raw_value));
+    auto const value = static_cast<T>(capped_value);
```
binarman, line 69: Or do you have concerns about this changes?

compiler/souschef/src/Gaussian.cpp
seanshpark: Yup :)

compiler/souschef/src/Gaussian.cpp
seanshpark: Nope :)
code looks good as I imagine as new method intruduction of `template <typename T> std::vector<uint8_t> GenerateGaussian(int32_t count, float mean, float stddev)`

binarman: Ok! I've created #6538

It differs from this commit: it actually unifies all generators, including floating point one.

binarman: I want to merge this PR after 6538

binarman:
I need to investigate. It works on local machine

binarman:
FYI @seanshpark 
Problem is in implicit dependency between souschef and circlechef.
In current design I can not add anything in `compiler/souschef/include/souschef/DataChef.def` without adding it in schema in both tflchef and circlechef.

So I created #6593 =)

seanshpark (approved):
LGTM thank you!

Issue #4433, opened 2020-09-25
[arser] Support short option


mhs4670go:
It would be helpful for users if arser supports short option.

Let's support this!

mhs4670go:
help message should print short option.

mhs4670go:
Test for help message is needed.=)

PR #6535, opened 2021-04-09
[arser] Support short option

mhs4670go:
This commit enables arser to use short option.

Related : #4433

### DIFF
--- a/compiler/arser/include/arser/arser.h
+++ b/compiler/arser/include/arser/arser.h
@@ -261,6 +261,34 @@ public:
     return *_arg_map[arg_name];
   }
 
+  Argument &add_argument(const std::vector<std::string> &args)
+  {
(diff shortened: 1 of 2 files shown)


compiler/arser/include/arser/arser.h
```diff
@@ -261,6 +261,34 @@ class Arser
     return *_arg_map[arg_name];
   }
 
+  Argument &add_argument(const std::vector<std::string> &args)
+  {
+    _optional_arg_vec.emplace_back(args.at(0));
+    for (const auto &arg : args)
```
seanshpark, line 267: this `arg` seems same level as `arg_name` from existing `add_argument`
how about using the same variable name for better understanding for new `add_argument` ?

seanshpark:
Nice to have this new feature !

seanshpark (approved):
LGTM thank you!

PR #6538, opened 2021-04-09
[souschef] Unify gaussian generators

binarman:
This commit unifies float, int32_t and uint8_t gaussian generators.

### DIFF
--- a/compiler/souschef/src/Gaussian.cpp
+++ b/compiler/souschef/src/Gaussian.cpp
@@ -26,76 +26,48 @@
 namespace souschef
 {
 
-std::vector<uint8_t> GaussianFloat32DataChef::generate(int32_t count) const
+template <typename T>
+static std::vector<uint8_t> generate_gaussian(int32_t count, float mean, float stddev)
 {
   // TODO Support seed value override
   auto seed = std::chrono::system_clock::now().time_since_epoch().count();
 
   std::minstd_rand rand{static_cast<std::minstd_rand::result_type>(seed)};
-  std::normal_distribution<float> dist{_mean, _stddev};
+  std::normal_distribution<float> dist{mean, stddev};
 
   std::vector<uint8_t> res;
 
+  constexpr float max_cap = std::numeric_limits<T>::max();
+  constexpr float min_cap = std::numeric_limits<T>::min();
   for (uint32_t n = 0; n < count; ++n)
   {
-    auto const value = dist(rand);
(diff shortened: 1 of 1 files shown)


seanshpark:
@binarman , I think from #6533 , `as new method intruduction of template <typename T> std::vector<uint8_t> GenerateGaussian(int32_t count, float mean, float stddev)` wasn't enough to express what I wished...

For upcoming PRs, I would like you to do in some order to reduce misunderstanding and for simple and clear code understanding.
Each PR would be better to have single purpose: add, change, remove, for single subject.
And it is OK to have large lines of changes if the purpose is still single. 

Also it would be better to add an issue (for this PR, I think it is not necessary) to describe reason and background. And full draft code if there are multiple subject of changes. (#6533 seems to show the full code so no draft is needed for this case)

For this PR, I wish you could post in separate PRs:
- introduce new template method 
- revise existing codes to use new method
- add new codes that uses new types
- remove unused code if any

binarman:
@seanshpark 
> For this PR, I wish you could post in separate PRs:

Hmm, in case I prefer to create issue, because number or required PRs growing too fast (first I thought that 2 will be enough)
Progress will be tracked in #6576

P.s.
> revise existing codes to use new method
> remove unused code if any

Probably this two PRs are actually one, because during revise old code will be removed.

seanshpark:
> Probably this two PRs are actually one, because during revise old code will be removed.

Strictly speaking, there are 3 distinct actions: add new method, change existing code to use new method, remove old method.

binarman:
> Strictly speaking, there are 3 distinct actions: add new method, change existing code to use new method, remove old method.

Ok! I was confused, because previously this rule was applied only to interfaces, but this PR affects only implementation in one file.

seanshpark:
As there are time difference and applying too strict changes will make delay in progress.
I hope not to apply too many changes of purpose in one PR.

seanshpark (approved):
LGTM

seanshpark:
If you like, you can merge this or get another approval :)

binarman:
> As there are time difference and applying too strict changes will make delay in progress.

This is not so an urgent task. I just want to update infrastructure =)

I've already cut changes (and refactored them a little), so i would like to proceed in other PRs.
I am going to close this one after changes are merged, (consider this as a draft I guess)

binarman:
Closing, everything from this PR is merged

PR #6558, opened 2021-04-12
[luci] Rewrite QuantizedModelVerifier Test

dayo09:
This commit rewrites some parts of QuantizedModelVerifier.test.cpp

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -373,14 +373,8 @@ public:
   void init(void)
   {
     TestIOGraph::init({16}, {32});
-    _param = g()->nodes()->create<luci::CircleConst>();
-    {
-      _param->dtype(Type::FLOAT32);
-      _param->shape({16});
-      _param->size<Type::FLOAT32>(16);
-      for (int16_t i = 0; i < 16; i++)
-        _param->at<Type::FLOAT32>(i) = static_cast<float>(i);
-    }
+    
+    _param = create_dummy_const<Type::FLOAT32>(g(), {16});
     _concat = g()->nodes()->create<luci::CircleConcatenation>(2);
     {
       _concat->values(0, input());
@@ -418,6 +412,9 @@ public:
     set_minmax_to_non_const(g(), -1, 1);
   }
 
+  loco::Node *x() { return _op->x(); }
+  loco::Node *y() { return _op->y(); }
+
 public:
   Op *_op = nullptr;
   luci::CircleConst *_y = nullptr;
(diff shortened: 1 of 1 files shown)


seanshpark:
Please separate changes that have different purposes.

dayo09:
@seanshpark 

I can seperate that `(1)utilization of create_dummy_const` from `(2)x(), y() getter implementation`.
Can I just separate commits or do I need to make separate PRs - in that case, what title do you recommend for each? (Sorry for asking for detailed instructions, this is because I don't have a clue and I want to know your intention 😸  )

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -373,14 +373,8 @@ class ConcatenationTestGraph final : public luci::test::TestIOGraph
...
-    _param = g()->nodes()->create<luci::CircleConst>();
-    {
-      _param->dtype(Type::FLOAT32);
-      _param->shape({16});
-      _param->size<Type::FLOAT32>(16);
-      for (int16_t i = 0; i < 16; i++)
-        _param->at<Type::FLOAT32>(i) = static_cast<float>(i);
-    }
+    
+    _param = create_dummy_const<Type::FLOAT32>(g(), {16});
```
seanshpark, line 377: This is revist to use `create_dummy_const`

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -418,6 +412,9 @@ template <class Op> class ComparisonOpTestGraph final : public luci::test::TestI
     set_minmax_to_non_const(g(), -1, 1);
   }
 
+  loco::Node *x() { return _op->x(); }
```
seanshpark, line 415: this is introduction of new accessor(?) `x()` and `y()`

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -860,18 +852,18 @@ TEST(QuantizedModelVerifierTest, GreaterEqual_wrong_type_NEG)
 TEST(QuantizedModelVerifierTest, GreaterEqual_wrong_granularity_NEG)
 {
   TEST_WITH_WRONG_GRANULARITY_TARGET(ComparisonOpTestGraph<luci::CircleGreaterEqual>, Type::U8,
-                                     Granularity::LayerWise, g._op->x());
+                                     Granularity::LayerWise, g.x());
```
seanshpark, line 855: these seem to be to test new `x()` and `y()`

seanshpark:
> this is because I don't have a clue and I want to know your intention 

I've added comments to the code: if you need more clue, please add more questions :)

seanshpark:
So what I could say is that we can separate to two PRs.

dayo09:
@seanshpark I mean, won't they have the same PR title: `Revisit QuantizedModelVerifier Test`?

seanshpark:
> I mean, won't they have the same PR title: Revisit QuantizedModelVerifier Test?

Yes it will. So it would be better to rename the title to reflect the changes.

seanshpark:
> Am I able to put specific reasons like implement x,y getter onto the title?

Yes something like, `Introduce accessor to x, y to test graph`

PR #6563, opened 2021-04-12
[luci] Add test attributes of AvgPool2D clone

seanshpark:
This will add CircleAveragePool2D clone test with attributes value
testing.


mhs4670go (approved):
LGTM

PR #6564, opened 2021-04-12
[luci] Set valid values of clone negative test

seanshpark:
This will set valid values for items that should be valid
for correct negative testing of invalid attributes.


mhs4670go (approved):
LGTM

PR #6565, opened 2021-04-12
[luci] Introduce node accessors to QuantizedModelVerifier test graphs

dayo09:
This commit introduces node accessors to TestGraph classes.

----

From #6558

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -418,6 +418,9 @@ public:
     set_minmax_to_non_const(g(), -1, 1);
   }
 
+  const loco::Node *x(void) const { return _op->x(); }
+  const loco::Node *y(void) const { return _op->y(); }
+
 public:
   Op *_op = nullptr;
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -452,6 +452,9 @@ template <class Op> class BinaryLogicalOpTestGraph final : public luci::test::Te
     set_minmax_to_non_const(g(), -1, 1);
   }
 
+  loco::Node *x() { return _op->x(); }
+  loco::Node *y() { return _op->y(); }
```
seanshpark, line 456: if usage is readonly, 
```cxx
  const loco::Node *x(void) const { return _op->x(); }
  const loco::Node *y(void) const { return _op->y(); }
```

seanshpark:
build error --> I think we may have to go with
```cxx
  loco::Node *x(void) const { return _op->x(); }
  loco::Node *y(void) const { return _op->y(); }
```

jinevening (approved):
LGTM

seanshpark (approved):
LGTM thank you!

PR #6567, opened 2021-04-12
[record-minmax] Skip nan values during calibration

jinevening:
This skips nan values during calibration.

### DIFF
--- a/compiler/record-minmax/src/MinMaxObserver.cpp
+++ b/compiler/record-minmax/src/MinMaxObserver.cpp
@@ -18,6 +18,8 @@
 
 #include <luci/IR/CircleOpcode.h>
 
+#include <math.h>
+
 using DataType = luci_interpreter::DataType;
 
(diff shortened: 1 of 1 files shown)


compiler/record-minmax/src/MinMaxObserver.cpp
```diff
@@ -73,9 +75,27 @@ void MinMaxObserver::postTensorWrite(const luci::CircleNode *node,
...
+
+    if (number > max)
+      max = number;
+
+    if (number < min)
+      min = number;
+  }
+
+  if (all_nan)
+    throw std::runtime_error("All values are NAN (Not-A-Number)");
```
seanshpark, line 98: https://en.wikipedia.org/wiki/NaN

how about `All values are NaN (Not a Number)` ?

seanshpark (approved):
LGTM thank you!

PR #6526, opened 2021-04-09
DRAFT: Support bool type input in record-minmax

jinevening:
This supports bool type input in record-minmax.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/6367

PR #6569, opened 2021-04-12
[record-minmax] Skip recording bool type tensor

jinevening:
This skips recording bool type tensor.

---

Related to: #6367
Draft PR: #6526


seanshpark (approved):
LGTM

Issue #6537, opened 2021-04-09
[luci] revise CircleUnique _idx_out_type accessor


seanshpark:
Current `luci::CircleUnique` has
```cxx
  loco::DataType idx_out_type(void) const { return _idx_out_type; }
  void output_type(loco::DataType ot) { _idx_out_type = ot; }
```
which is
1) have different accessor names `idx_out_type` vs `output_type`
2) different from TF `out_idx`

let's rename the accessor to ~`out_idx`~
--> found out that current rule is to follow tflite schema attributes: `idx_out_type`

seanshpark:
done

PR #6570, opened 2021-04-12
[luci] Introduce idx_out_type for CircleUnique IR

seanshpark:
This will add idx_out_type() accessor with a comment in CircleUnique IR
to relace existing idx_out_type.


seanshpark:
For #6537

mhs4670go (approved):
LGTM

PR #6572, opened 2021-04-12
[arser] Apply short option

mhs4670go:
This commit applies short option to help message and `exit_with` method.

Related: #4433 

### DIFF
--- a/compiler/arser/include/arser/arser.h
+++ b/compiler/arser/include/arser/arser.h
@@ -295,7 +295,7 @@ public:
     _program_name.erase(0, _program_name.find_last_of("/\\") + 1);
     if (argc >= 2)
     {
-      if (!std::strcmp(argv[1], "--help"))
+      if (!std::strcmp(argv[1], "--help") || !std::strcmp(argv[1], "-h"))
       {
         std::cout << *this;
         std::exit(0);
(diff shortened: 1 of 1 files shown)


compiler/arser/include/arser/arser.h
```diff
@@ -305,7 +305,7 @@ class Arser
         for (const auto &arg : _arg_map)
         {
           const auto &func = arg.second->_func;
-          if (func && !std::strcmp(argv[1], arg.second->_name.c_str()))
+          if (func && !std::strcmp(argv[1], arg.first.c_str()))
```
seanshpark, line 308: How does this change be related with short option?

mhs4670go: @seanshpark Very implementation detail.. 

arser internally has argument map with key as `std::string` given by user in `add_argument` and with value as `Argument` class. And, `Argument` has only one name attribute(`_name`, i.e. `arg.second->_name`) which is the first parameter of `add_argument`. Before the short option, comparing `argv`(from user) with `arg.second->_name` was enough because there was only one option name. But, after the short option is merged, the contents of the argument map has been changed like below.

```bash
// before short option
$ cat _arg_map<std::string, Argument *>
- "--help", help_Argument
- "--version", version_Argument
- "--my_option", my_option_Argument
...
// after short option
- "--help", help_Argument
- "-h", help_Argument
- "--version", version_Argument
- "-v", version_Argument
- "--my_option", my_option_Argument
...
```

So, what is compared to `argv` must be `arg.first` rather than `arg.second->_name`.

seanshpark (approved):
LGTM

PR #6574, opened 2021-04-12
[luci] Use CircleUnique idx_out_type

seanshpark:
This will revise to use idx_out_type() of CircleUnique.


mhs4670go (approved):
LGTM

seanshpark:
For #6537

PR #6579, opened 2021-04-13
[luci] Support importing quantized input for RSQRT

jinevening:
This supports importing quantized input for RSQRT.


seanshpark:
change itself looks good but I'm not sure how the change itself has any relation with `RSQRT`
maybe it would be better to change title/comment about this change.

seanshpark (approved):
LGTM

PR #6588, opened 2021-04-13
[luci] Fix wrong assert in compute_sym_scale_zp

jinevening:
This fixes wrong assert in compute_sym_scale_zp.


seanshpark (approved):
LGTM

PR #6589, opened 2021-04-13
[arser] Make arser print short option properly

mhs4670go:
This commit makes arser print short option properly.

Even though #6535,  #6572 have been merged, arser didn't print the short option properly because they were just about an interface-related changes. From this PR, I expect arser will be able to fully support a short option.

I think it is a bit hard to review the changes but I will try to explain it as much as I can.

Here are changes that I think is important.
- `Argument` class has introduced `_short_name`, `_long_name`, `_names`.
  - `_name` has been divided into `_short_name` and `_long_name` because `arser` got to support multiple names.
  - `_names` vector includes `_short_name` and `_long_name`. This variable is for pretty printing.
- Validation of arguments given by users.
  - Multiple names requires more thorough validation than it used to.
- Comments for developers.

Related: #4433 

### DIFF
--- a/compiler/arser/include/arser/arser.h
+++ b/compiler/arser/include/arser/arser.h
@@ -29,6 +29,8 @@
 
 #include <cstring>
 
+#include <cassert>
+
 namespace arser
 {
 namespace internal
@@ -56,6 +58,31 @@ template <typename T> inline std::string to_string(const T value) { return std::
 template <> inline std::string to_string(const char *value) { return std::string(value); }
 
 template <> inline std::string to_string(const bool value) { return value ? "true" : "false"; }
+
+/**
+ * @brief Returns the string with the leading dash removed.
+ *
+ * If there is no dash, it returns as it is.
+ */
+std::string remove_dash(const std::string &str)
+{
+  std::string ret{str};
+  auto pos = ret.find_first_not_of('-');
+  if (pos == std::string::npos)
+    return ret;
+  return ret.substr(pos);
+}
+
+/**
+ * @brief Returns the string that created by concatenating the elements of a vector with commas.
+ */
+std::string make_comma_concatenated(const std::vector<std::string> &vec)
+{
+  std::ostringstream oss;
+  std::copy(vec.begin(), std::prev(vec.end()), std::ostream_iterator<std::string>(oss, ", "));
+  oss << vec.back();
+  return oss.str();
+}
 
 } // namespace internal
 } // namespace arser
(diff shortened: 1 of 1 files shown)


compiler/arser/include/arser/arser.h
```diff
@@ -57,6 +59,29 @@ template <> inline std::string to_string(const char *value) { return std::string
...
+std::string remove_dash(const std::string &str)
+{
+  std::string ret{str};
+  auto pos = ret.find_first_not_of('-');
+  if (pos == std::string::npos)
+    return ret;
+  return ret.substr(pos);
+}
+
+/// @brief Returns the string that created by concatenating the elements of a vector with commas.
```
seanshpark, line 76: Please use one style `/**` or `///`

compiler/arser/include/arser/arser.h
```diff
@@ -468,21 +580,21 @@ std::ostream &operator<<(std::ostream &stream, const Arser &parser)
...
-      std::string arg_name = arg._name.substr(2);
+      std::string arg_name = arser::internal::remove_dash(arg._long_name);
       std::for_each(arg_name.begin(), arg_name.end(),
                     [&stream](const char &c) { stream << static_cast<char>(::toupper(c)); });
     }
     stream << "]"
            << " ";
   }
-  // positional arguement
+  // positional argument
```
seanshpark, line 594: this seems not related with the change

compiler/arser/include/arser/arser.h
mhs4670go: @seanshpark Oh, I thought `///` is for single line and `/**` for larger line. Should it be fixed?

seanshpark: > Oh, I thought /// is for single line and /** for larger line.

Actually we didn't have any strict guide for this but usually we used `/**` style, even for one line.

> Should it be fixed?

I hope so...

compiler/arser/include/arser/arser.h
```diff
@@ -119,10 +146,41 @@ enum class DataType
 
 class Arser;
 
+/**
+ * Argument
+ *   ├── positional argument
+ *   └── optioanl argument  [ dash at the beginning of the string ]
+ *       ├── long option    [ two or more dashes ]
```
seanshpark, line 153: (Q) does `---` or `----` work?

compiler/arser/include/arser/arser.h
```diff
@@ -119,10 +146,41 @@ enum class DataType
...
+ *       └── short option   [ one dash ]
```
seanshpark, line 154: Q) is `-hello` valid?

compiler/arser/include/arser/arser.h
```diff
@@ -119,10 +146,41 @@ enum class DataType
...
+ *
+ * Argument has two types - positional argument, optional argument.
+ *
+ * The way to distinguish the two types is whether there is a dash('-') at the beginning of the
+ * string.
+ *
+ * And, optional argument has two types as well - long option, short option, which is distinguished
```
seanshpark, line 161: What I assumed is, long option with full name, like `--help`and short option with one dash and one character like `-h`
from this comment, I has some questions like above...

seanshpark:
Code itself is a bit hard to understand... I only can do what I can...
Spliting the change will help bit it's an option.

compiler/arser/include/arser/arser.h
mhs4670go: @seanshpark Yes. It works well. Which one will be long or short is intentionally not be described because that is so verbose.

compiler/arser/include/arser/arser.h
mhs4670go: ditto

compiler/arser/include/arser/arser.h
mhs4670go: ditto. That's the basic assumption. Even though user gives `----h` or etc, it will work.

mhs4670go:
@seanshpark 
> Code itself is a bit hard to understand... I only can do what I can...
Spliting the change will help bit it's an option.

Well, I guess the reason why it is hard to understand is because the place to review is a part of printing. For pretty printing, it needs to be complicated. What I come up with is to merge this PR after #6592 is merged. 
> Add help message test that outputs of `./driver -h`.

At least, you will be able to see the print message, which will be basis for reviewing this PR.

seanshpark (approved):
LGTM

seanshpark:
> At least, you will be able to see the print message, which will be basis for reviewing this PR.

I think we can go with this :)

PR #6595, opened 2021-04-13
[luci] Remove unused accessor in CircleUnique

seanshpark:
This will remove unused accessor output_type() in CircleUnique IR.


seanshpark:
For #6537

mhs4670go (approved):
LGTM

PR #6600, opened 2021-04-14
[ONE] Revert android build runtime release shell script.

Hyunjun85:
- In previously, To adapt armcompute library v21.02,
  The 'EXT_ACL_FOLDER' was commented out.
  If not, It tried to build with v20.05 and showed some errors.
  Because the binary of library was not update on CI.

ONE-DCO-1.0-Signed-off-by: <Hyunjun Kim> <hyunjun2.kim@samsung.com>


chunseoklee:
@Hyunjun85 We do not need this since acl build artifact cached on server. Moreover, we have to change acl version external CI script to support this PR.

PR #6605, opened 2021-04-14
[luci] Verify quantization of Split

jinevening:
This verifies quantization of Split Op.

---
Related to: #6367

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -270,7 +270,7 @@ public:
 class SplitTestGraph final : public SimpleTestGraph
 {
 public:
-  void init(void)
+  void init(void) override
   {
     // Split has the same
     TestIOGraph::init({32}, {32});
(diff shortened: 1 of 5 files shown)


compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -267,6 +267,29 @@ template <Type indexT> class SliceTestGraph final : public SimpleTestGraph
   luci::CircleConst *_size = nullptr;
 };
 
+class SplitTestGraph final : public SimpleTestGraph
+{
+public:
+  void init(void)
```
seanshpark, line 273: please add `override` here :)

jinevening: Oops.. Thank you!

dayo09 (approved):
LGTM :-D

jinevening:
Split has virtual successors (SplitOut), so it is quantized a bit differently from normal operators.

In fp32 model,
fp32 -> Split -> fp32 -> SplitOut -> fp32

In u8 model,
u8 -> Split -> fp32 -> SplitOut -> u8

So the output of SplitOut should be checked.

seanshpark (approved):
LGTM

dayo09 (approved):
LGTM

PR #6612, opened 2021-04-14
[luci] Rename variable for TestOGraphlet

seanshpark:
This will rename variable name in TestOGraphlet.


mhs4670go (approved):
LGTM

Issue #6576, opened 2021-04-12
[tflchef] [circlechef] Support Q16 tensors


binarman:
## Goal

Support INT16 data type in tflchef tool.

## Rationale

We need Q16 tests generator. In particular for internal part of the project.

## Progress

- [x] Support INT16 datatype in tflchef #6534
- [x] Support INT16 datatype in circlechef #6593
- [x] Add unified gaussian generator for sousechef to reduce duplication of code in generators (this and next item #6538) #6577
- [x] Use unified generator instead of duplicated code #6578 
- [x] Support INT16 data generators in sousechef (for data generators) #6533
- [x] Add smoke test for tflchef #6615
- [x] Add smoke test for circlechef #6614

binarman:
+cc @seanshpark

binarman:
Everything is merged, closing.
@seanshpark Thank you!

PR #6614, opened 2021-04-14
[circlechef] Introduce INT16 test

binarman:
This commit adds test for INT16 tensors in recipe.

issue #6576


seanshpark (approved):
LGTM

Issue #6592, opened 2021-04-13
[arser] Revise the test


mhs4670go:
## What

Let's revise the test.

## Why

Existing test works well but it is kind of hard for developers to read the test. Testing code is an introduction to a newbie as well as old(?) developers. Running and reading the test codes are often the best thing they can do to start when they have to work on it. So, I want it to be more readable for everybody:)

Here's what I am gonna do.
- Divide `arser.test.cpp` into several files according to its functionality.
- Add missing negative cases.
- Add help message test that outputs of `./driver -h`.

PR #6617, opened 2021-04-15
[arser] Make arser follow ODR

mhs4670go:
This commit makes arser follow the One Definition Rule.

There are three changes to accomplish this.
- define header guard
- make non-template function inline
- move non-template function related `Arser` to inside its method (`operator<<`)

Related: #6592 


seanshpark (approved):
LGTM

PR #6618, opened 2021-04-15
[luci] Support quantization of ResizeBilinear

dayo09:
This commit supports quantization and its unittest of ResizeBilinear operation.

ONE-DCO-1.0-Signed-off-by: Dayoung Lee <dayoung.lee@samsung.com>

------------

For #6367

PR #6622, opened 2021-04-16
[luci] Add create_const helper to QuantizedModelVerifier test

dayo09:
This commit adds create_const helper function to create const node with value

-----

From #6618

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -72,7 +72,6 @@ template <Type T> luci::CircleConst *create_dummy_const(loco::Graph *g, luci::te
 
 /**
  * @brief A helper function to create const node with value
- *
  */
 template <Type DT, typename T>
 luci::CircleConst *create_const(loco::Graph *g, luci::test::ShapeU32 shape,
@@ -888,16 +887,6 @@ TEST(QuantizedModelVerifierTest, LocalCreateConst)
   {
     EXPECT_EQ(node->at<Type::FLOAT32>(index++), val);
   }
-}
-
-TEST(QuantizedModelVerifierTest, LocalCreateConst_NEG)
-{
-  loco::Graph g;
-  std::initializer_list<float> values = {0.1, 0, -5, 100};
-  luci::CircleConst *node = create_const<Type::FLOAT32, float>(&g, {2, 2}, values);
-
-  // access out of bound
-  EXPECT_DEATH(node->at<Type::FLOAT32>(4), "Assertion `n < size<DT>()");
 }
 
 TEST(QuantizedModelVerifierTest, InstanceNorm)
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -70,6 +70,32 @@ template <Type T> luci::CircleConst *create_dummy_const(loco::Graph *g, luci::te
   return node;
 }
 
+/**
+ * @brief A helper function to create const node with value
+ *
+ */
+template <Type DT, typename T>
+luci::CircleConst *create_const(loco::Graph *g, luci::test::ShapeU32 shape,
+                                std::initializer_list<T> values)
+{
+  auto node = g->nodes()->create<luci::CircleConst>();
+  {
+    node->dtype(DT);
+    node->shape(shape);
+    node->size<DT>(luci::test::num_elements(shape));
+
+    assert(values.size() == node->size<DT>());
+
+    uint32_t index = 0;
+    for (auto val : values)
+    {
+      node->at<DT>(index++) = static_cast<T>(val);
+    }
+  }
+
+  return node;
+}
+
 void insert_scale_zp(luci::CircleNode *node, float scale, int64_t zp)
 {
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -851,6 +877,29 @@ TEST(QuantizedModelVerifierTest, LocalCreateDummyConst)
...
+}
+
+TEST(QuantizedModelVerifierTest, LocalCreateConst_NEG)
+{
+  loco::Graph g;
+  std::initializer_list<float> values = {0.1, 0, -5, 100};
+  luci::CircleConst *node = create_const<Type::FLOAT32, float>(&g, {2, 2}, values);
+
+  // access out of bound
+  EXPECT_DEATH(node->at<Type::FLOAT32>(4), "Assertion `n < size<DT>()");
```
dayo09, line 900: I guess this don't work with release test, because assert doens't work.

dayo09:
@nnfw-bot test nncc-debug
@nnfw-bot test nncc-release
CI Network Error..

seanshpark:
@nnfw-bot test nncc-debug
@nnfw-bot test nncc-release

dayo09:
### EXPECT_DEATH regex

EXPECT_DEATH(statement, regex) gets regex parameter and expects the statement to be death by a message of which contains regex.
It seems that `'` is regarded as a special character. (but even with escape character `\` doesn't work somehow.)


```
TEST(QuantizedModelVerifierTest, LocalCreateConst_NEG)
{
  ...
  EXPECT_DEATH(node->at<Type::FLOAT32>(4), "Assertion `n < size<DT>()' failed."); //FAIL
}
TEST(QuantizedModelVerifierTest, LocalCreateConst_NEG2)
{
  ...
  EXPECT_DEATH(node->at<Type::FLOAT32>(4), "Assertion `n < size<DT>()"); // SUCCESS
}
```

Full target error message:
```
[  DEATH   ] luci_pass_test: /nfs/git/ONE/compiler/luci/lang/src/Nodes/CircleConst.cpp:48: typename loco::DataTypeImpl<DT>::Type& luci::CircleConst::at(uint32_t) [with loco::DataType DT = (loco::DataType)10; typename loco::DataTypeImpl<DT>::Type = float; uint32_t = unsigned int]: Assertion `n < size<DT>()' failed.
```

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -851,6 +877,29 @@ TEST(QuantizedModelVerifierTest, LocalCreateDummyConst)
   EXPECT_NO_THROW(create_dummy_const<Type::FLOAT32>(&g, {32, 32}));
 }
 
+TEST(QuantizedModelVerifierTest, LocalCreateConst)
```
seanshpark, line 880: `QuantizedModelVerifierTest` Test is used for IR(Op) tests. to distinguish this, how about revise?
- use anoter name, maybe `QuantizedModelVerifierHelperTest`
- use `local_create_const`

dayo09: > use anoter name, maybe QuantizedModelVerifierHelperTest

I like this idea :-D

> use local_create_const

It's using a seperate namespace in order to bound the scope. Does it still need `local_` prefix?

seanshpark: > Does it still need local_ prefix?

I don't think so. :)

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -70,6 +70,32 @@ template <Type T> luci::CircleConst *create_dummy_const(loco::Graph *g, luci::te
   return node;
 }
 
+/**
+ * @brief A helper function to create const node with value
+ *
```
seanshpark, line 75: maybe remove this empty line?

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
seanshpark: yes, assert check is a bit difficult with gtest for this reason.

seanshpark (approved):
LGTM

mhs4670go:
@dayo09 FYI, c++ compiler have specific [escape sequences rule](https://en.cppreference.com/w/cpp/language/escape). With that, you can make a proper regex that works well in c++.

In short, here's what we should do.
1. `()`(group capturing), `'`, `.`(Matches any character) should be escaped 
```
"Assertion `n < size<DT>\\(\\)\' failed\\."
```
Just `\(` can't be applied because there is no such escape sequence in c++. So, `\\(` can be transformed into `\(`, which is our goal. And so can `\\.`. 

2. > At any rate, I am removing that EXPECT_DEATH checker as it needs more lines to be working at release build check.

AFAIK, there's `EXPECT_DEBUG_DEATH`. It won't need more lines I think.

seanshpark:
`EXPECT_DEBUG_DEATH` is good for testing Debug but this will make difference in release vs debug.
When we had this kind of problem last year, I recall as it will be not good to make differences for testing with build type.

dayo09:
@mhs4670go Thanks for the information! :-D

Issue #6624, opened 2021-04-16
Compiler FE: Tidy luci circle IRs


seanshpark:
While working on partitioning, I've found that IR class has some differences from each other
- some have constructors, although it's default
- some have redundant comments
- some have attribute init in ctor, some have in default values
- some have attribute init with `= 0`, some use `{0}`
- mose have `public:` between input and attribute accessors, some don't

seanshpark:
There is no one correct answerwhich to use but it's a styly that I want to make not several but one.
- remove constructor just `() = default`
- remove redundant comments
- set default values using `{...}` style in attribute declaration not in constructor.
- put `public:` between input / attributes accessor

seanshpark:
`CircleReverseSequence` uses `int` for `seq_axis`, `batch_axis` --> ...

seanshpark:
Seems to be done for now

PR #6625, opened 2021-04-16
[luci] Tidy IR with default ctor

seanshpark:
This will tidy IR with default ctor.


seanshpark:
For #6624

mhs4670go (approved):
LGTM

PR #6597, opened 2021-04-14
DRAFT: [luci] Connect node for partition

seanshpark:
on-going draft to implement and test node connection in partition.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #6596, opened 2021-04-14
Compiler FE: Connect nodes partitioned graph


seanshpark:
Lets' implement all the nodes for _luci/partition_ `clone_connect()` method.

Implementation itself maybe not that difficult but may require larger amount of codes to preare the environment
- we need source graph to clone
- we need to make clone of the graph: inputs of the target nde to clone and the node itself
- we need to validate clone is OK

seanshpark:
This is sub task of #6230

seanshpark:
done for now :)

PR #6629, opened 2021-04-18
[luci] Relocate clone_connect method

seanshpark:
This will relocate clone_connect method so that this method can be
called from other places.


seanshpark:
For #6596, from draft #6597

mhs4670go (approved):
LGTM

PR #6630, opened 2021-04-19
[luci] Introduce TestIsOGraph for connection test

seanshpark:
This will introduce TestIsOGraph class for connection test providing
arbitrary N input with template.

### DIFF
--- a/compiler/luci/partition/src/ConnectNode.test.h
+++ b/compiler/luci/partition/src/ConnectNode.test.h
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6596, from draft #6597

compiler/luci/partition/src/ConnectNode.test.h
```diff
@@ -0,0 +1,53 @@
...
+
+namespace luci
+{
+namespace test
+{
+
+template <unsigned N> class TestIsOGraph : public TestIsGraphlet<N>, public TestOGraphlet
+{
+public:
+  TestIsOGraph() = default;
```
mhs4670go, line 37: Removing the explicit default constructor is only applied to IR(#6624)?

seanshpark: Yup :)

seanshpark: For some tests graphs there will be empty class if this default ctor is removed. this is OK in syntax but may look a bit difficult to read the code.

mhs4670go (approved):
LGTM

PR #6631, opened 2021-04-19
[luci] Revise to use explicit int32 type

seanshpark:
This will revise to use explicit int32 type for attributes.


seanshpark:
For #6624

mhs4670go (approved):
LGTM

PR #6634, opened 2021-04-19
DRAFT: [luci] Connect node for partition (for merge)

seanshpark:
on-going draft to implement and test node connection in partition.

Signed-off-by: SaeHie Park saehie.park@gmail.com

PR #6636, opened 2021-04-19
[luci] Add num_inputs to TestIsGraphlet

seanshpark:
This will add num_inputs method to access number of inputs of TestIsGraphlet.


seanshpark:
For #6596, from draft #6634

mhs4670go (approved):
LGTM

PR #6638, opened 2021-04-19
[luci] Introduce node graphlet create helper classes

seanshpark:
This will introduce two helper classes NodeIsGraphletT and
NodeIsGraphletT that will create node graphlet that will provide one
input or arbitrary inputs.

### DIFF
--- a/compiler/luci/partition/src/ConnectNode.test.h
+++ b/compiler/luci/partition/src/ConnectNode.test.h
@@ -47,6 +47,38 @@ public:
   }
 };
 
+template <class T> class NodeGraphletT
+{
+public:
+  virtual void init(loco::Graph *g)
+  {
+    _node = g->nodes()->create<T>();
+    _node->dtype(loco::DataType::S32);
+    _node->name("node");
+  }
+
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6596, from draft #6634

compiler/luci/partition/src/ConnectNode.test.h
```diff
@@ -47,6 +47,38 @@ template <unsigned N> class TestIsOGraph : public TestIsGraphlet<N>, public Test
...
+    _node = g->nodes()->create<T>();
+    _node->dtype(loco::DataType::S32);
+    _node->name("node");
+  }
+
+  T *node(void) const { return _node; }
+
+protected:
+  T *_node{nullptr};
+};
```
mhs4670go, line 64: I think this could be substituted by `NodeIsGraphletT<1>`.

seanshpark: Yes, you are right. But for reading the code, without `<1>` would be better...

mhs4670go (approved):
LGTM

PR #6639, opened 2021-04-19
[luci] Revise partition connect for nodes Add and others

seanshpark:
This will revise partition connect method for nodes Add, Div, Mul and
Sub. Purpose of this revisit is to reduce class LOC of ConnectNode as it
will grow more than 500 lines.


seanshpark:
For #6596, from draft #6634

mhs4670go (approved):
LGTM

seanshpark:
I think this needs #6640

mhs4670go (approved):

PR #6640, opened 2021-04-19
[luci] Change public to find_connect method

seanshpark:
This will change scope public to find_connect method of ConnectNode
class.


seanshpark:
For #6596, from draft #6634

mhs4670go (approved):
LGTM

PR #6641, opened 2021-04-19
[luci] Introduce ConnectionTestHelper class

seanshpark:
This will introduce ConnectionTestHelper class to make cloned node
connection test easy.


seanshpark:
For #6596, from draft #6634

mhs4670go (approved):
LGTM

PR #6644, opened 2021-04-19
[Compiler FE] Bump version to 1.16.0

seanshpark:
This will bump compiler frontend version to 1.16.0


mhs4670go (approved):
LGTM

PR #6646, opened 2021-04-19
[luci] prepare_inputs_miss for negative test

seanshpark:
This will introduce prepare_inputs_miss method in ConnectionTestHelper
for negative test case.

### DIFF
--- a/compiler/luci/partition/src/ConnectNode.test.h
+++ b/compiler/luci/partition/src/ConnectNode.test.h
@@ -105,6 +105,23 @@ public:
     }
   }
 
+  /**
+   * @note prepare_inputs_miss is for negative testing
+   */
+  template <unsigned N> void prepare_inputs_miss(TestIsOGraph<N> *isograph)
+  {
+    assert(N == isograph->num_inputs());
+
+    for (uint32_t i = 0; i < N; ++i)
+    {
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6596, from draft #6634

mhs4670go:
@nnfw-bot test nncc-release

compiler/luci/partition/src/ConnectNode.test.h
```diff
@@ -105,6 +105,23 @@ class ConnectionTestHelper
...
+
+    for (uint32_t i = 0; i < N; ++i)
+    {
+      auto *input = _graph_clone->nodes()->create<luci::CircleInput>();
+      luci::copy_common_attributes(isograph->input(i), input);
+      if (i != 0)
+        _clonectx.emplace(isograph->input(i), input);
+      _inputs.push_back(input);
+    }
+  }
```
mhs4670go, line 123: I wonder there's an any difference between normal one. Above inputs are invalid?

seanshpark: Yup. we remove first input so the connection to input SHOULD fail. this is the purpose.

mhs4670go: Oh, I missed that part. Now I got it:)

mhs4670go (approved):
LGTM

PR #6649, opened 2021-04-20
[luci] prepare_inputs for TestIOGraph

seanshpark:
This will introduce prepare_inputs and prepare_inputs_miss method for
TestIOGraph.
- although it's single input, has 's' to make test codes simple


seanshpark:
For #6596, from draft #6634

mhs4670go (approved):
LGTM

PR #6650, opened 2021-04-20
[luci] Tests of partition connection for Mean and others

seanshpark:
This will introduce unit tests of partiton connection for nodes Mean,
Pow, Rsqrt, Sqrt and SquaredDifference.


seanshpark:
For #6596, from draft #6634

mhs4670go (approved):
LGTM

PR #6637, opened 2021-04-19
[luci] Replace Argmax quantize verification test

dayo09:
This commit replaces argmax quantize verification tests.
Specifically, it tests with wrong type not with dimension but with input node, with more test variant.
And it tests with wrong granularity with more test variant.
The removed wrong type test with dimension node will be revived after the implementation of proper test macros.

ONE-DCO-1.0-Signed-off-by: Dayoung Lee <dayg502@gmail.com>

----

For #6635

PR #6652, opened 2021-04-20
[luci] Add a wrong type target test to QuantizedModelVerifier test

dayo09:
This commit adds a wrong type test function with a specified target node.

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -114,6 +114,7 @@ void quantize_and_verify(loco::Graph *g, Type quantized_dtype, Granularity granu
 
 // Helper function to reduce duplicate test codes
 // Assumption: g->output()->from() is the target node
+// TODO: Remove the function and use below function to explicit target node
 void quantize_and_verify_with_wrong_type(luci::test::TestIOGraph *g, Type quantized_dtype,
                                          Granularity granularity, Type wrong_dtype)
 {
@@ -142,6 +143,7 @@ void quantize_and_verify_with_wrong_type(luci::test::TestIOGraph *g, Type quanti
 
 // Helper function to reduce duplicate test codes
 // Assumption: g->output()->from() is the target node
+// TODO: Remove the function and use below function to explicit target node
 void quantize_and_verify_with_wrong_granularity(luci::test::TestIOGraph *g, Type quantized_dtype,
                                                 Granularity granularity)
 {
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -1144,7 +1172,23 @@ TEST(QuantizedModelVerifierTest, ArgMax_wrong_dimension_type_NEG)
   SUCCEED();
 }
 
-// TODO : Add a negative test which sets wrong dtype on dimension node
+TEST(QuantizedModelVerifierTest, ArgMax_wrong_dimension_type_NEG)
```
dayo09, line 1175: This test code ensures that the above test function works well.

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -114,6 +114,7 @@ void quantize_and_verify(loco::Graph *g, Type quantized_dtype, Granularity granu
 
 // Helper function to reduce duplicate test codes
 // Assumption: g->output()->from() is the target node
+// TODO: Remove the function and use below function to explicit target node
```
dayo09, line 117: The previous helper functions assumed the target node as g->output->from() node, but now that we have both speficied-target test helper, we can replace the function to explicit the target node.

seanshpark:
@dayo09 , there are mutliple purpose of changes and this makes me hard to understand the code. please split the changes as short as possible you can.
and it would be better not to use explanations like `with a specified target node.` that I dont have any idea what this `specific` means.

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
dayo09: @jinevening PTAL :-D

seanshpark:
This PR is about `output->from()` to `dimension()` of ArgMax if I understand the changes.
Can you explain why this change is needed?

dayo09:
@seanshpark Previously it was dimension() test. But as I refactor argmax test code https://github.com/Samsung/ONE/pull/6637/files, it was impossible to keep so I have replaced them with TODO command.

```
(line 1147) // TODO : Add a negative test which sets wrong dtype on dimension node
```

Now, this code implements that TODO.

seanshpark:
> Previously it was dimension() test. But as I refactor argmax test code

It's a bit hard to know why we need to change from `out->from()` to `dimension()`.
As I understand #6637 was a change to previous one kind of test to using macro for multple kinds of test.
Do we need to roleback it? Or are you planning to add test for `out->from()` too?

seanshpark:
>  it was impossible to keep

what was impossible to keep? and why? I think we need to solve this problem.
why I am asking is that I don't fully understand you changes and @jinevening is absent to look the codes with domain knowledge.

dayo09:
@seanshpark 
It has already merged. Let me repeat the commit message in the pr.

> Specifically, it tests with wrong type not with dimension but with input node, with more test variant.
...
> The removed wrong type test with dimension node will be revived after the implementation of proper test macros.
```

1. At first, @jinevening made a test with `dimension` node
2. At #6637, I have removed the test and replaced with TODO comment.
   And I added the test with `input` node but mistakenly used the test name `ArgMax_wrong_dimension_type_NEG`
3. At #6655, I have fixed the mistaken name.
4. At this draft, it revives `dimension` node test.

seanshpark:
@dayo09 , thanks for the explanation. Please explain next time prior to doing some fixes to reduce confusion.

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
jinevening: IMHO, I think we can keep this function to avoid changing test cases.

Instead, we can implement this function with your function (with the `target`) to reduce duplicate codes.

dayo09:
Rebased :-D

jinevening (approved):
LGTM

dayo09:
All the prs from this draft are merged. :-D

PR #6653, opened 2021-04-20
[luci] Add a helper function to QuantizedModelVerifier test

dayo09:
This commit adds a helper function to test a quantized graph with wrong type.
Previously there was a function with the same name that tests verifier with wrongly typed output's from node.
This function tests verifier with wrongly typed user-given node.

-----

From #6652

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -127,6 +127,19 @@ void quantize_and_verify_with_wrong_type(luci::test::TestIOGraph *g, Type quanti
   verifier.verify(g->g());
 }
 
+void quantize_and_verify_with_wrong_type(luci::test::TestIOGraph *g, Type quantized_dtype,
+                                         Granularity granularity, Type wrong_dtype,
+                                         luci::CircleNode *target)
+{
+  luci::QuantizeWithMinMaxPass pass(Type::FLOAT32, quantized_dtype, granularity);
+  pass.run(g->g());
+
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -127,6 +127,19 @@ void quantize_and_verify_with_wrong_type(luci::test::TestIOGraph *g, Type quanti
   verifier.verify(g->g());
 }
 
+void quantize_and_verify_with_wrong_type(luci::test::TestIOGraph *g, Type quantized_dtype,
```
seanshpark, line 130: (optional) as this is another version of `quantize_and_verify_with_wrong_type` with explict `target` node as input parameter, it would be better to add a comment like
```
// Helper function to reduce duplicate test codes with a target as input parameter
```
as existing had a comment but this new one didn't

dayo09: I had thought about that, but I didn't put the line, as this verifier file is a test.cpp file. If a function is declared in a test file they are already a helper function, and looks redundant to tell :-D

dayo09: Also `reduce duplication` sounds kinda function's basic role, so...

seanshpark: > already a helper function, and looks redundant to tell 

Sound reasonable :)

seanshpark (approved):
LGTM

Issue #6616, opened 2021-04-15
Compiler FE: Check onnx-tensorflow 1.8


seanshpark:
Latest onnx-tensoflow 1.8  version was released. 
We need to check with our in-house models are converted without any problems.

We will upgrade to latest TF version but can we use with our current TF 2.3?
- Yes, seems working with TF 2.3.0: but need more model conversion testing.

seanshpark:
Set (1) = TF2.3.0 + onnx-tf 1.7.0, (2) = TF2.4.1 + onnx-tf 1.8.0
- using onnx 1.9.0, which is latest, has problem converting our in-house models.

ReLU6 conversion has changed
- (1) = Minimum + Maximum
- (2) = Minimum + ReLU

seanshpark:
onnx-tf(==1.8.0) + TF(==2.4.1) has some conversion issue with PB-->TFLite.

TODO: find the pattern (graphlet) and dig the reason.

seanshpark:
Close this issue for now. Proceed to next step with new issues.

PR #6706, opened 2021-04-27
DRAFT Validate with onnx-tf 1.8.0

seanshpark:
on-going draft to validate in-house models compilation with onnx-tf 1.8.0


seanshpark:
For #6616

seanshpark:
done for now

PR #6710, opened 2021-04-27
[luci] Clone empty for OutputDummy and OutputExclude

seanshpark:
This will add empty connection for CircleOutputDummy and
CircleOutputExclude. These two has no inputs so doesn't
do any actions.


seanshpark:
For #6596, from draft #6634

mhs4670go (approved):
LGTM

PR #6713, opened 2021-04-27
[luci] Comment of clone/connect

seanshpark:
This will add comment of clone/connect for CircleInput and CircleOutput.


seanshpark:
For #6230, #6596

mhs4670go (approved):
LGTM

PR #6715, opened 2021-04-29
[one-cmds] Revise one-prepare-venv with PIP_OPTIONS

seanshpark:
This will revise one-prepare-venv script with common PIP_OPTIONS for pip
options.


seanshpark:
This is to prepare package version upgrade to be easy.

mhs4670go (approved):
LGTM

PR #6718, opened 2021-04-29
[one-cmds] Introduce version variable to one-prepare-venv

seanshpark:
This will introduce version variable to interested packages in
one-prepare-venv. And with version information URL.


seanshpark:
This is to prepare package version upgrade to be easy.

mhs4670go (approved):
LGTM

Issue #6707, opened 2021-04-27
Compiler FE: Partition model with popular public models


seanshpark:
Let's check model partition with IV3 and other popular models

From this experience we may get more ideas with model partitioning.
- such as, partition before/after particular op

and find unknown problems from this task

---
Sites of public models
- https://tfhub.dev/

seanshpark:
For #6230

seanshpark:
How to test the partition...
- Just a sctipt to do some manual partitioning
- A compiler sub module run as a test

As a first step, let's find out what are needed to do this.

seanshpark:
Found working on top of one-cmds testing is one of (many??) easy approach.

seanshpark:
Models check 
- https://tfhub.dev/captain-pool/lite-model/esrgan-tf2/1
- https://tfhub.dev/google/faster_rcnn/openimages_v4/inception_resnet_v2/1

seanshpark:
CC @mhs4670go

mhs4670go:
@seanshpark Saved model related [API](https://www.tensorflow.org/api_docs/python/tf/lite/TFLiteConverter#from_saved_model) - `from_saved_model` has three parameters., i.e. `saved_model_dir`, `signature_keys`, `tags`.

As of now, `tf2tfliteV2` supports limited interface when it comes to saved model. It means that users can set `saved_model_dir` but can't set `signature_keys`, `tags`. They are just set to default value.

Most of saved models might have default tags, i.e. `SERVING`. But above `inception_resnet_v2` model doesn't.
> All tags in the tag set must be present. (default set(SERVING))

```bash
$ saved_model_cli show --dir ${model_folder} --all

MetaGraphDef with tag-set: '' contains the following SignatureDefs:

signature_def['default']:
  The given SavedModel SignatureDef contains the following input(s):
    inputs['images'] tensor_info:
        dtype: DT_FLOAT
        shape: (1, -1, -1, 3)
        name: hub_input/image_tensor:0
  The given SavedModel SignatureDef contains the following output(s):
    outputs['detection_boxes'] tensor_info:
        dtype: DT_FLOAT
        shape: (-1, 4)
        name: hub_input/strided_slice:0
    outputs['detection_class_entities'] tensor_info:
        dtype: DT_STRING
        shape: (-1, 1)
        name: hub_input/index_to_string_Lookup:0
    outputs['detection_class_labels'] tensor_info:
        dtype: DT_INT64
        shape: (-1, 1)
[...]

PR #6719, opened 2021-04-29
DRAFT: Partition with public models

seanshpark:
on-going draft to run partition with popular public models.


seanshpark:
For  #6707

compiler/one-cmds/tests/partition_esrgan-add.test-x
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+# Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
```
YiHyunJin, line 3: I think.... it is 2021
`2020 Samsung Electronics Co ` >> `2021 Samsung Electronics Co`

Issue #6661, opened 2021-04-23
[luci-interpreter] Add PadV2 Kernel


ai-moiseev:
## What

Add PadV2 Op

## Why

Need to quantize models with PadV2 Op.

## Steps

* [x] luci-interpreter #6662 
* [x] circle-quantizer #6721

## Quantization result

* Float: [padv2.f32.circle.zip](https://github.com/Samsung/ONE/files/6413934/padv2.f32.circle.zip)
* Q8: [padv2.q8.circle.zip](https://github.com/Samsung/ONE/files/6413939/padv2.q8.circle.zip)

seanshpark:
Can you please add some expected value tests?

seanshpark:
@jinevening , can we add tests for individual Ops? or we need to prepare some infra for this?

jinevening:
> @jinevening , can we add tests for individual Ops? or we do need to prepare some infra first?

We have `pota-quantization-value-test` for testing end-to-end process for quantization.

IMHO, uploading the quantization result (fp32 model and q8 model) to this issue may be enough for PadV2.

ai-moiseev:
@jinevening I've attached Q8 results to the issue body.

seanshpark:
Purpose of the test is to include tests for new codes introduce for test coverage :)
- https://github.com/Samsung/ONE/pull/6721#issuecomment-831612669

But I guess I was not specific for the reason :)

seanshpark:
@ai-moiseev , can we close this issue?

ai-moiseev:
@seanshpark Sorry for the delay, closing now :)

PR #6544, opened 2021-04-12
[luci] Verify quantization of Rsqrt

dayo09:
This commit verifies quantization of rsqrt operation.

ONE-DCO-1.0-Signed-off-by: Dayoung Lee <dayg502@gmail.com>

-----
For #6367

PR #6721, opened 2021-04-29
[circle-quantizer] Support quantization of PadV2

ai-moiseev:
This commit adds support for quantization of PadV2.

Requires #6725 
For #6661 

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -981,21 +981,10 @@ void quantize_const_inputs(luci::CircleNode *node, loco::DataType output_type)
       break;
 
     case luci::CircleOpcode::PADV2:
-      // Only the third input is quantized
+      // First and third constant inputs are quantized
       // Second input should not be quantized (e.g., paddings)
-      // Quant params propagated from output range to the input
-      input_node = node->arg(2);
-      const_node = dynamic_cast<luci::CircleConst *>(input_node);
-      if (const_node != nullptr && !is_quantized(const_node))
-      {
-        auto input = dynamic_cast<luci::CircleNode *>(node->arg(0));
-        assert(input->quantparam()->scale.size() == 1); // input scale's layer-wise
-
-        luci::quant_const_values(const_node, node->quantparam()->scale.at(0),
-                                 node->quantparam()->zerop.at(0), output_type);
-        luci::overwrite_quantparam(node, const_node);
-        luci::overwrite_quantparam(node, input);
-      }
+      // Quant params are propagated from output range to the non-constant input
+      propagate_pad_v2_quantparam(loco::must_cast<CirclePadV2 *>(node), output_type);
       break;
 
     default:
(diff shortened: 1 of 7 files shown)


seanshpark:
(optional) As you wrote in the message, there are two changes in this PR. It would be better to split them in separate PR.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -980,6 +980,24 @@ void quantize_const_inputs(luci::CircleNode *node, loco::DataType output_type)
         quant_const(const_node, output_type);
       break;
 
+    case luci::CircleOpcode::PADV2:
+      // Only the third input is quantized
+      // Second input should not be quantized (e.g., paddings)
+      // Quant params propagated from output range to the input
+      input_node = node->arg(2);
```
jinevening, line 987: I think the quantization logic needs to be updated as below.

1. We apply `overwrite_quantparam` only to (1) cloned const node or (2) feature map with one successor, because the target node can be used by another node.
(PTAL at `propagate_concat_qparam`)

2. `input` can be a constant tensor. If so, its value should be quantized as well.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -980,6 +980,24 @@ void quantize_const_inputs(luci::CircleNode *node, loco::DataType output_type)
         quant_const(const_node, output_type);
       break;
 
+    case luci::CircleOpcode::PADV2:
+      // Only the third input is quantized
```
jinevening, line 984: First input can be quantized if it is a constant tensor.

jinevening:
I'm also concerned about subsequent PadV2 Ops. For the case, qparam must be propagated from the output to the input.

It may hardly occur, so it would be possible to skip that case (not propagate qparam) as we did in concat.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
ai-moiseev: I've updated logic to follow these rules, now it supports constant input and checks number of successors in case of non-constant node.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
ai-moiseev: Updated this commentary.

ai-moiseev:
@seanshpark I've separated method refactor into its own PR, PTAL: #6725 
@jinevening I've added check for subsequent PadV2 ops pattern, so quantparams not propagated in that case now.

jinevening:
LGTM Thank you.

Please rebase after #6725 is merged.

jinevening:
One more thing.

Could you add type/granularity tests? This is for test coverage. https://github.com/Samsung/ONE/pull/6544/files may help you.

ai-moiseev:
> One more thing.
> 
> Could you add type/granularity tests? This is for test coverage. https://github.com/Samsung/ONE/pull/6544/files may help you.

Sure, done :) Thank you for an example, I was wondering where to add tests.

jinevening (approved):
LGTM Thank you

PR #6724, opened 2021-04-30
[one-cmds] Add model_format option to cfg template

mhs4670go:
Tihs commit adds model_format option to cfg template.


seanshpark (approved):
LGTM thank you!

Issue #6726, opened 2021-05-03
Compiler FE: Install header files


mhs4670go:
## What

Let's install header files!

## Why

To make `one-compiler-dev` debian package, all the header files that are used by `one-compiler` binaries should be installed.

## Lists
- [x] loco
- [x] luci
- [x] luci-interpreter

Related: #6717

PR #6727, opened 2021-05-03
[loco] Install header files

mhs4670go:
This commit makes loco header files installed to `include` dir.

Related: #6726 

### DIFF
--- a/compiler/loco/CMakeLists.txt
+++ b/compiler/loco/CMakeLists.txt
@@ -14,6 +14,8 @@ target_link_libraries(loco PRIVATE nncc_common)
 target_link_libraries(loco PUBLIC nncc_coverage)
 # Q. HOW TO MAKE DEV PACKAGE(?)
 install(TARGETS loco DESTINATION lib)
+install(DIRECTORY include/ DESTINATION include
+        FILES_MATCHING PATTERN "*.h")
 
 if(NOT ENABLE_TEST)
   return()


compiler/loco/CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ target_link_libraries(loco PRIVATE nncc_common)
 target_link_libraries(loco PUBLIC nncc_coverage)
 # Q. HOW TO MAKE DEV PACKAGE(?)
 install(TARGETS loco DESTINATION lib)
+# Install header files
```
seanshpark, line 17: I think this comment is same as the below code itself is can be omitted :)

seanshpark (approved):
LGTM thank you!

compiler/loco/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ target_link_libraries(loco PRIVATE nncc_common)
 target_link_libraries(loco PUBLIC nncc_coverage)
 # Q. HOW TO MAKE DEV PACKAGE(?)
 install(TARGETS loco DESTINATION lib)
+install(DIRECTORY include/ DESTINATION include
+        FILES_MATCHING PATTERN "*.h")
```
parjong, line 18: What is this for?

parjong: Are you going to introduce developer package?

parjong: Oh, CMake 3.5 support this pattern ;-) 
- [Reference document](https://cmake.org/cmake/help/v3.5/command/install.html#installing-files).

mhs4670go: @parjong 
> Are you going to introduce developer package?

Right. This is for `one-compiler-dev` debian package:)

parjong (approved):

PR #6728, opened 2021-05-03
[luci] Install header files

mhs4670go:
This commit makes luci header files installed to `include` dir.

Related: #6726


compiler/luci/env/CMakeLists.txt
```diff
@@ -6,6 +6,9 @@ add_library(luci_env SHARED ${SOURCES})
 target_include_directories(luci_env PUBLIC include)
 target_link_libraries(luci_env PRIVATE nncc_common)
 install(TARGETS luci_env DESTINATION lib)
+# Install header files
```
seanshpark, line 9: I think this comment is same as the below code itself is can be omitted :)

seanshpark (approved):
LGTM thank you!

Issue #6730, opened 2021-05-03
Compiler FE: Support STRING type tensor


seanshpark:
There seems STRING type tensor exist in TensorFlow lite files.
Let's at least import without error for these files.

seanshpark:
From https://github.com/Samsung/ONE/issues/6707#issuecomment-831591443

seanshpark:
```
$ circledump faster_rcnn_openimages_v4_inception_resnet_v2_1.circle | grep STRING
T(0:19) STRING () B(20) std.constant18
T(0:1312) STRING () B(1313) StatefulPartitionedCall_1:3
T(0:1313) STRING () B(1314) StatefulPartitionedCall_1:1
T(11:0) STRING (604) B(1387) std.constant565
T(11:1) STRING (604) B(1388) std.constant566
```

seanshpark:
Not sure this can be solved in one SPRINT...

- [ ] understand how to read/write STRING tensor
- [ ] support STRING in tflchef
   - [ ] prepare a recipe
   - [ ] enable STRING type in tflchef
- [ ] enable in luci
   - [ ] identify codes to enable
   - [ ] TBD.

seanshpark:
python script to generate tflite file with STRING type

```py
# run in TF2.3.0
import tensorflow as tf
from tensorflow.python.ops import lookup_ops

tf.compat.v1.disable_eager_execution()
sess = tf.compat.v1.Session()

lhs_ = tf.compat.v1.placeholder(dtype=tf.string, shape=(), name="Hole")
rhs_ = tf.constant("ABC", name="Hole")
op_ = tf.add(lhs_, rhs_)

converter = tf.compat.v1.lite.TFLiteConverter.from_session(sess, [lhs_], [op_])
converter.allow_custom_ops = True
tflite_model = converter.convert()
tflite_file = './const_string.tflite'
open(tflite_file, "wb").write(tflite_model)
```

seanshpark:
How does TF generate Buffer?

https://github.com/tensorflow/tensorflow/blob/14b13fce8c351fb9efc32b0dcad1804fd2d03e0b/tensorflow/lite/toco/tflite/types.cc#L24-L37
```cxx
DataBuffer::FlatBufferOffset CopyStringToBuffer(
    const Array& array, flatbuffers::FlatBufferBuilder* builder) {
  const auto& src_data = array.GetBuffer<ArrayDataType::kString>().data;
[...]

PR #6745, opened 2021-05-07
DRAFT: Compiler FE support STRING tensor type

seanshpark:
On going draft to support STRING tensor type.


seanshpark:
For #6730

seanshpark:
all landed

PR #6747, opened 2021-05-07
[luci-interpreter] Support more fused activation functions

jinevening:
This enables to support more fused activation functions.


seanshpark:
I don't think this is correct changes for activations.
Please do not remove for `default` and implement for each cases.

seanshpark (approved):
LGTM thank you!

Issue #6742, opened 2021-05-06
Compiler FE: Import 'faster_rcnn_openimages' model


seanshpark:
Let's check import of public model in TensorFlow hub `faster_rcnn/openimages_v4/inception_resnet_v2`

Model: https://tfhub.dev/google/faster_rcnn/openimages_v4/inception_resnet_v2/1

seanshpark:
From https://github.com/Samsung/ONE/issues/6707#issuecomment-831591443
Depends on #6730

seanshpark:
There is a CustomOp with 0 input...

seanshpark:
This too needs revise
> compiler/luci/import/src/GraphBuilderMultiOutput.cpp:63: virtual luci::CircleNode* luci::GraphBuilderMultiOutput::build(const ci
rcle::OperatorT&, luci::GraphBuilderContext*) const: Assertion `output_count > 0' failed.
Aborted (core dumped)

Hum?
> compiler/luci/service/src/CircleShapeInferenceRule.cpp:558: loco::NodeShape {anonymous}::infer_concatenation(const luci::CircleConcatenation*): Assertion `!output_shape.dim(j).known() || !input_shape.dim(j).known() || output_shape.dim(j) == input_shape.dim(j)' failed.
Aborted (core dumped)

seanshpark:
Hum?
> compiler/luci/service/src/CircleShapeInferenceRule.cpp:558: loco::NodeShape {anonymous}::infer_concatenation(const luci::CircleConcatenation*): Assertion `!output_shape.dim(j).known() || !input_shape.dim(j).known() || output_shape.dim(j) == input_shape.dim(j)' failed.
Aborted (core dumped)

![image](https://user-images.githubusercontent.com/4616940/117403669-81c87600-af43-11eb-8e3a-42fc77f0c65b.png)

seanshpark:
```
[luci] CircleConv2D ShapeInf ifm(4) ker(4) output(1,1,1,192) tfl.conv_2d86
[luci] CircleConv2D ShapeInf ifm(4) ker(4) output(1,1,1,128) tfl.conv_2d87
[luci] CircleConv2D ShapeInf ifm(4) ker(4) output(1,1,4294967285,160) tfl.conv_2d88
```
strange...

_circleump_ values
```
T(0:686) FLOAT32 (1, 1, 1, 128) (-1, -1, -1, 128) B(687) tfl.conv_2d87
[...]

PR #6748, opened 2021-05-07
[luci] Add assert on invalid shape

seanshpark:
This will add assert on invalid shape for Conv2D and Pool2D types.


seanshpark:
Related: https://github.com/Samsung/ONE/issues/6742#issuecomment-834112587

llFreetimell (approved):
LGTM

PR #6750, opened 2021-05-09
[foder] Do not delete move members

mhs4670go:
This commit removes `delete` keyword of move members.

I just wanted the class to be not copyable. Turns out that move members shouldn't be deleted at any conditions.

For more details, please refer to https://stackoverflow.com/a/38820178/8741151.


seanshpark (approved):
LGTM

Issue #6746, opened 2021-05-07
[chef] Support STRING tensor type


seanshpark:
Let's support STRING tensor type in _souschef_, _circlechef_ and _tflchef_

seanshpark:
Splitting #6730 issue for a SPRINT

seanshpark:
done

PR #6752, opened 2021-05-09
[souschef] to_number for string

seanshpark:
This will introduce to_number for string type.


seanshpark:
For #6746, from draft #6745

mhs4670go (approved):
LGTM

PR #6753, opened 2021-05-09
[circlechef] Relocate DataChef.def

seanshpark:
This will relocate DataChef.def back to circhechef as to reduce
change dependency and it's not used in souschef.


seanshpark:
Related: #6746, draft #6745

mhs4670go (approved):
LGTM

Issue #6717, opened 2021-04-29
Compiler FE: create debian package


mhs4670go:
## What

Let's create a debian package.

## Why

It's gonna make it easy to distribute our compiler. 

For the record, I'll just make a debian folder that is needed for building debian package rather than using  [launchpad](https://launchpad.net/) ppa. It could be used but there would be a lot of works to do because launchpad build farm does not have internet access(please see [this answer](https://answers.launchpad.net/launchpad/+question/260227) for more detail).

Related: #6627 

### reference

https://www.debian.org/doc/manuals/debmake-doc/index.en.html
https://www.debian.org/doc/manuals/maint-guide/index.en.html

mhs4670go:
To build a debian package, I should write bunch of scripts. You could see [this page](https://www.debian.org/doc/manuals/maint-guide/dreq.en.html) for more detail.

**Caveat**. You should install [pre-requisite package](https://www.debian.org/doc/manuals/debmake-doc/ch03.en.html) before following below steps.

### Create debian package
`debmake` is a program to make the `debian` source package.
```bash
$ wget https://github.com/Samsung/ONE/archive/refs/tags/1.15.0.tar.gz
$ tar -xf 1.15.0.tar.gz
$ cd ONE-1.15.0
$ ls
compiler  CONTRIBUTORS  docs   LICENSE  Makefile.template  nncc  nnpackage  README.md  runtime  tools
compute   COPYRIGHT     infra  logo     nnas               nnfw  packaging  res        tests

# create debian folder
$ debmake -t
$ tree debian -F
debian
├── changelog
├── compat
├── control
├── copyright
├── one-core.install
├── one.install
├── patches/
│   └── series
├── README.Debian
├── rules*
├── source/
│   ├── format
│   └── local-options
└── watch
```

### Write required files

```bash
$ cat debian/control

Source: one
Section: devel
Priority: extra
Maintainer: Neural Network Acceleration Solution Developers <nnfw@samsung.com>
Build-Depends: cmake, debhelper (>=9)
Standards-Version: 3.9.8
Homepage: https://github.com/Samsung/ONE

Package: one-compiler
Architecture: any
Multi-Arch: foreign
Depends: one-compiler-dev, ${misc:Depends}, ${shlibs:Depends}
Description: auto-generated package by debmake
 This Debian binary package was auto-generated by the
 debmake(1) command provided by the debmake package.

Package: one-compiler-dev
Architecture: any
Multi-Arch: same
[...]

PR #6755, opened 2021-05-10
Introduce debian folder

mhs4670go:
This commit introduces debian folder for generating debian package.

Related: #6717 


debian/copyright
```diff
@@ -0,0 +1,3 @@
+Files: *
+License: Proprietary
+Copyright (c) <2018> <Samsung Electronics Co.,Ltd.>
```
seanshpark, line 3: 2018 is tart of _nnfw_ ?

debian/control
```diff
@@ -0,0 +1,25 @@
...
+Package: one-compiler
+Architecture: any
+Multi-Arch: foreign
+Depends: ${misc:Depends}, ${shlibs:Depends}
+Description: on-device neural engine compiler
+
+Package: one-compiler-dev
+Architecture: any
+Multi-Arch: same
+Depends: ${shlibs:Depends}, ${misc:Depends}
```
seanshpark, line 18: Can `one-compiler-dev` work without `one-compiler` ? If not, we need to add depend on `one-compiler`

debian/postinst
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html
+# Boradly speaking, the `postinst` is called after a package is unpacked.
+
+set -e
+
+/opt/one_compiler/bin/one-prepare-venv
```
seanshpark, line 8: If user can change installation path to somewhere else, is `/opt/one_compiler/bin/one-prepare-venv` available?

debian/copyright
mhs4670go: @seanshpark Right.

debian/control
mhs4670go: @seanshpark I think `one-compiler-dev` should work without `one-compiler`. I'll add `lib` to its install list.

debian/postinst
mhs4670go: @seanshpark Umm I thought user can't change the installation path. Is it possible?

debian/control
seanshpark: > I'll add lib to its install list.

Does this mean, both `one-compiler` and `one-compiler-dev` will install `lib` ?

debian/postinst
seanshpark: > Is it possible?

I don't know either...
But with some searching, it seems not possible :)

debian/control
seanshpark: > Yes. That's what I meant to.

I think this may make conflict on `lib`.

mhs4670go: > I think `one-compiler-dev` should have a dependency on `one-compiler`.

Is it for preventing `one-compiler-dev` from having duplicate `lib` list?

mhs4670go:
[one-compiler.tar.gz](https://github.com/Samsung/ONE/files/6449902/one-compiler.tar.gz)
This is the package from this PR.

debian/control
```diff
@@ -0,0 +1,25 @@
...
+Maintainer: Neural Network Acceleration Solution Developers <nnfw@samsung.com>
+Build-Depends: cmake, debhelper (>=9), dh-python, python3-all
+Standards-Version: 3.9.8
+Homepage: https://github.com/Samsung/ONE
+
+Package: one-compiler
+Architecture: amd64
+Multi-Arch: foreign
+Depends: ${misc:Depends}, ${shlibs:Depends}
+Description: on-device neural engine compiler
```
seanshpark, line 13: `On-device Neural Engine compiler` ?

seanshpark: maybe `Description: ... compiler package` ? as below two ends with `package`

seanshpark (approved):
LGTM thank you!

chunseoklee (approved):

debian/one-compiler-dev.install
```diff
@@ -0,0 +1,5 @@
+# {FILES_TO_INSTALL} {DEST_DIR}
+# bin
+usr/bin/circledump opt/one_compiler/bin/
```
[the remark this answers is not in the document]
seanshpark: `/usr/local/share/one` :)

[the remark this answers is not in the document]
jyoungyun: My mistake, Thank you :)

PR #6762, opened 2021-05-10
[circlechef] Add STRING TensorType in proto

seanshpark:
This will introduce STRING TensorType in proto file.


seanshpark:
Related: #6746, draft #6745

PR #6765, opened 2021-05-11
[tflchef] Enable STRING TensorType

seanshpark:
This will enable STRING TensorType in tflchef.


seanshpark:
For #6746, from draft #6745

mhs4670go (approved):
LGTM

PR #6766, opened 2021-05-11
[circlechef] Enable STRING TensorType

seanshpark:
This will enable STRING TensorType in circlechef.


compiler/circlechef/tests/string_tensor/test.recipe
```diff
@@ -0,0 +1,31 @@
...
+    arg: "Hello"
+  }
+}
+operand {
+  name: "ofm"
+  type: STRING
+  shape { }
+}
+operation {
+  type: "BatchMatMul"
```
mhs4670go, line 21: Is it possible for `BatchMatMul` to run with string type?

seanshpark: Actually not possible :) This test is just to test creation of _circle_ model from _recipe_ file.
FYI, files in `circlechef/tests` are not executed.

mhs4670go (approved):
LGTM

PR #6775, opened 2021-05-11
[common-artifacts] Exclude for STRING TensorType

seanshpark:
This will add exclude items for upcomming STRING TensorType support.


seanshpark:
For #6730, from draft #6745

PR #6776, opened 2021-05-11
[res] Add_STR recipes for luci testing

seanshpark:
This will add introduce Add_STR_000 and Add_STR_001 recipes for luci
import/export testing with STRING TensorType.
- "Add" Op with STRING is not supported in any current backends


seanshpark:
For #6730, from draft #6745

mhs4670go (approved):
LGTM

PR #6778, opened 2021-05-11
[luci] Remove static function keyword

seanshpark:
This will remove static function keyword and move to anonymous
namespace.


mhs4670go (approved):
LGTM

PR #6780, opened 2021-05-11
[luci] Enable export for STRING TensorType Const

seanshpark:
This will enable export of STRING TensorType of CircleConst.

### DIFF
--- a/compiler/luci/export/src/CircleTensorExporter.cpp
+++ b/compiler/luci/export/src/CircleTensorExporter.cpp
@@ -28,6 +28,8 @@
 #include <loco/IR/DataTypeTraits.h>
 #include <oops/InternalExn.h>
 
+#include <string.h>
+
 using namespace circle;
 using namespace flatbuffers;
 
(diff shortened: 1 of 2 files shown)


seanshpark:
FYI: Format of buffer
https://github.com/Samsung/ONE/blob/dda39274f3fa06ee5ddc384926d69f133e27423f/compiler/souschef/src/Explicit.cpp#L25-L33

compiler/luci/export/src/CircleTensorExporter.cpp
```diff
@@ -352,6 +354,50 @@ flatbuffers::Offset<circle::Buffer> encodeOpBufferByDType(FlatBufferBuilder &bui
...
+  }
+
+  auto *data = reinterpret_cast<uint8_t *>(i32d);
+  for (uint32_t i = 0; i < count; ++i)
+  {
+    int32_t length = offsets[i + 1] - offsets[i];
+    auto &value = c->at<loco::DataType::STRING>(i);
+    memcpy(data, value.c_str(), length);
+    data += length;
+  }
```
mhs4670go, line 395: I'm not sure but it can be like this:)
```suggestion
  auto *data = reinterpret_cast<uint8_t *>(i32d + offset);

  *i32d++ = count;
  *i32d++ = start;
  for (uint32_t i = 0; i < count; ++i)
  {
    auto &value = c->at<loco::DataType::STRING>(i);
    offset += value.length();
    *i32d++ = offset;
    memcpy(data, value.c_str(), value.size());
    data += value.length();
  }
```

compiler/luci/export/src/CircleTensorExporter.cpp
```diff
@@ -352,6 +354,50 @@ flatbuffers::Offset<circle::Buffer> encodeOpBufferByDType(FlatBufferBuilder &bui
   return CreateBuffer(builder, array_offset);
 }
 
+template <>
```
mhs4670go, line 357: I think memory structure comment would be helpful. Otherwise, below `count + 2` things would look like magic number.

compiler/luci/export/src/CircleTensorExporter.cpp
seanshpark: May work but would better to apply in another PR.

mhs4670go (approved):
LGTM

PR #6781, opened 2021-05-11
[luci] Enable import of STRING TensorType Const

seanshpark:
This will enable import of STRING TensorType CircleConst.


mhs4670go (approved):
LGTM

PR #6783, opened 2021-05-12
[luci] Assert check in shape inference

seanshpark:
This will add assert check of dimensions used in shape inference of conv, pool.

ONE-DCO-1.0-ONE-DCO-1.0-ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>


seanshpark:
Thought would be nice to have this while working on #6742

mhs4670go (approved):
LGTM

PR #6785, opened 2021-05-12
[luci] Enable test for STRING TensorType

seanshpark:
This will enable read/write test for STRING TensorType constant.


seanshpark:
For #6730, from draft #6745

mhs4670go (approved):
LGTM

PR #6786, opened 2021-05-12
[common-artifacts] Remove exclude item for STRING

seanshpark:
This will remove exclude item for STRING TensorType of circle2circle
optimization test.


mhs4670go (approved):
LGTM

PR #6789, opened 2021-05-12
[preset] Include luci-eval-driver for setup

seanshpark:
This will revise latest preset to include luci-eval-driver tool for
setup to run circle model.


mhs4670go (approved):
LGTM

Issue #6773, opened 2021-05-11
Compiler FE: Debian package installation directory


mhs4670go:
> Yesterday we discussed about the directory name in Release TF meeting. And it was concluded that `/usr/local/share/one` would be better for these files. Please update the file location in the next patch. :)

_Originally posted by @jyoungyun in https://github.com/Samsung/ONE/pull/6755#discussion_r629799844_

Actually, when we build the debian package that installs binaries in `/usr/local/share/one`, an error happened.

```bash
$ cd ONE
$ debuild -us -uc -b
...
	install -p -m0644 debian/copyright debian/one-compiler-test/usr/share/doc/one-compiler-test/copyright
   dh_installchangelogs
	install -p -m0644 debian/changelog debian/one-compiler/usr/share/doc/one-compiler/changelog
	install -p -m0644 debian/changelog debian/one-compiler-dev/usr/share/doc/one-compiler-dev/changelog
	install -p -m0644 debian/changelog debian/one-compiler-test/usr/share/doc/one-compiler-test/changelog
   dh_perl
   dh_usrlocal
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_profile.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_interpreter.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_partition.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_export.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libloco.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_env.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_import.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_pass.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_log.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_logex.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_lang.so is not a directory
dh_usrlocal: debian/one-compiler/usr/local/share/one/lib/libluci_service.so is not a directory
	rmdir debian/one-compiler/usr/local/share/one/lib
[...]

PR #6790, opened 2021-05-12
[rawdata2hdf5] Add readlink command

mhs4670go:
This commit adds readlink command when one sciprt finds script path,
which resolves symbolic link.

Related: #6773 (comment)


seanshpark (approved):
LGTM thank you!

PR #6795, opened 2021-05-12
[one-cmds] Upgrade venv onnx-tf to 1.8.0

seanshpark:
This will upgrade venv onnx-tf to 1.8.0 with a minor patch for
DepthwiseConv2D conversion.


mhs4670go (approved):
LGTM

Issue #6714, opened 2021-04-29
[luci] Transformation of Min+ReLU to ReLU6


seanshpark:
Let's add transformation of Min(6) + ReLU to ReLU6.

seanshpark:
From #6616

seanshpark:
done

PR #6797, opened 2021-05-13
DRAFT: Compiler FE: Support Min+ReLU to ReLU6

seanshpark:
draft to support Min+ReLU to ReLU6 transformation.


seanshpark:
For #6714

seanshpark:
all landed

Issue #6801, opened 2021-05-13
Compiler FE: Remove FakeQuant Pass


seanshpark:
Let's introduce a Pass to remove `FakeQuant` Op for experimental purposes.

seanshpark:
Background: #6800

seanshpark:
done

PR #6805, opened 2021-05-13
[luci] Introduce RemoveFakeQuantPass

seanshpark:
This will introduce RemoveFakeQuantPass that will remove all FakeQuant
Ops in the graph.

### DIFF
--- a/compiler/luci/pass/src/RemoveFakeQuantPass.cpp
+++ b/compiler/luci/pass/src/RemoveFakeQuantPass.cpp
@@ -23,10 +23,10 @@ namespace
 
 bool remove_fake_quant(luci::CircleFakeQuant *fakequant)
 {
-  if (fakequant == nullptr)
-    return false;
+  assert(fakequant != nullptr);
 
   auto input_node = loco::must_cast<luci::CircleNode *>(fakequant->inputs());
+
   replace(fakequant).with(input_node);
 
   return true;
--- a/compiler/luci/pass/src/RemoveFakeQuantPass.cpp
+++ b/compiler/luci/pass/src/RemoveFakeQuantPass.cpp
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


seanshpark:
For #6801, from draft #6802

compiler/luci/pass/src/RemoveFakeQuantPass.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+bool remove_fake_quant(luci::CircleFakeQuant *fakequant)
+{
+  if (fakequant == nullptr)
+    return false;
```
jinevening, line 27: This if-statement seems unnecessary. Maybe just `assert(fakequant != nullptr)`?

compiler/luci/pass/src/RemoveFakeQuantPass.test.cpp
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
```
jinevening, line 2:
```suggestion
 * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
```

jinevening (dismissed):
LGTM Thank you

compiler/luci/pass/src/RemoveFakeQuantPass.cpp
```diff
@@ -0,0 +1,69 @@
...
+ * CircleFakeQuant OP will be removed from the output graph
+ */
+bool RemoveFakeQuantPass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto target_node = dynamic_cast<luci::CircleFakeQuant *>(node);
+    if (target_node != nullptr)
+      if (remove_fake_quant(target_node))
```
mhs4670go, line 63: There's no case where `remove_fake_quant` returns false. How about remove `if (target_node != nullptr)` line and make it return false when `target_node == nullptr`?

seanshpark: I'll revise this :) Thanks!

seanshpark: Or would it be better to make `void remove_fake_quant()`....

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6806, opened 2021-05-13
[circle2circle] Enable transform_min_relu_to_relu6

seanshpark:
This will enable transform_min_relu_to_relu6 option.


seanshpark:
For #6714, from draft #6797

compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -256,6 +256,12 @@ int entry(int argc, char **argv)
     .default_value(false)
     .help("Transform Minimum-Maximum pattern to Relu6 operator");
```
seanshpark, line 257: I'll revise this to `Minimum(6)-Maximum(0)` in next PR.

mhs4670go (approved):
LGTM

Issue #6804, opened 2021-05-13
[circle-quantizer] Hard code scale of Floor/FloorDiv/FloorMod/Ceil


jinevening:
## What

Let's set the scale of Floor/FloorDiv/FloorMod/Ceil to a positive integer (`scale < 1 ? 1 : std::round(scale)`).

## Why

The output of Floor/FloorDiv/FloorMod/Ceil must be an integer. For that, scale should be an integer.

This may increase a quantization error (= |original value - dequantized value|) if scale is larger than 1. But we think it is more important to keep the output as integer.

PR #6808, opened 2021-05-13
[luci] Quantize Ops with an integer output

jinevening:
This sets scale to integer for Ops with integer output.

---
Related to: https://github.com/Samsung/ONE/issues/6804


seanshpark (approved):
LGTM

Issue #6793, opened 2021-05-12
Compiler FE: Proper way of debian packaging with python


mhs4670go:
## What

Let's find a better way to do debian-packaging of python things.

## Why

Recently, I've been working on #6717. Our one compiler binaries consist of c++ and python programs. Packaging c++ binaries seems "relatively" okay(but not that okay =_=) but packaging python ones are a bit tricky.

Debian supports [maintainer script](https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html) that will be run by the package management system for you when your package is installed, upgraded or removed.
> Broadly speaking the `preinst` is called before (a particular version of) a package is unpacked, and the `postinst` afterwards; the `prerm` before (a version of) a package is removed and the `postrm` afterwards.

As of now, `postinst` is called when our one-compiler deb package is installed where `one-prepare-venv` is called that installs all of python things that is needed to compiler binaries. Well, it kinda looks fine and is working out okay. But, technically, this method isn't recommended.

https://www.debian.org/doc/manuals/maint-guide/dother.en.html#maintscripts
> As a novice maintainer, you should avoid any manual editing of maintainer scripts because they are problematic.

https://askubuntu.com/questions/327543/how-can-a-debian-package-install-python-modules-from-pypi
> Downloading non-debian packages on-the-fly when installing a binary (.deb) is a serious security risk, definitely a no-no.

---

## Candidates

These are what I've found so far. If you have a better idea, feel free to post a comment about it.

### Use `postinst` script [chosen]

This is the way we currently use. After talking to @seanshpark, we've decided to go this way now until we find a better way.

### Use user's python environment

One of the reason that makes packaging hard is python virtual environment. Our `one-cmds` needs virtual environment(venv) and venv should be located in the same directory with the place `one-cmds` is located in, which can be automatically made it by running `one-prepare-venv` script. Also, users don't need to activate their virtual environment because `one-cmds` finds the venv path by itself.

To make things easier, we can let users create their virtual environment by themselves and activate it whenever they want to run `one-cmds`. Or, they can just install required python packages on their system's python environment. Then, they wouldn't need to activate the environment when they want to run.

### Use `stdeb` or `Pybuild`

Simple python package can be packaged by those tools. But there are so many packages to build and, especially, tensorflow can't be packaged in debian build system. So, this method is almost impossible.

### Use docker

[...]

PR #6809, opened 2021-05-13

mhs4670go:
This commit let users use venv python even though users run with sudo.

Related: https://github.com/Samsung/ONE/issues/6793#issuecomment-840306866

### DIFF
--- a/compiler/one-cmds/one-prepare-venv
+++ b/compiler/one-cmds/one-prepare-venv
@@ -19,6 +19,7 @@ set -e
 DRIVER_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 VENV_ACTIVATE=${DRIVER_PATH}/venv/bin/activate
+VENV_PYTHON=${DRIVER_PATH}/venv/bin/python
 
 function error_no_ensurepip ()
 {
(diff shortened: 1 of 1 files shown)


seanshpark:
Please update this too
```
PY_SITE_PACKAGES=$(python -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')
```

compiler/one-cmds/one-prepare-venv
```diff
@@ -61,30 +61,30 @@ PIP_OPTIONS="${PIP_TIMEOUT} ${PIP_TRUSTED_HOST}"
...
+  ${VENV_PYTHON} -m pip ${PIP_OPTIONS} install onnx==${VER_ONNX} ${EXT_ONNX_TF_WHL}
 else
-  python -m pip ${PIP_OPTIONS} install onnx==${VER_ONNX} onnx-tf==${VER_ONNX_TF}
+  ${VENV_PYTHON} -m pip ${PIP_OPTIONS} install onnx==${VER_ONNX} onnx-tf==${VER_ONNX_TF}
 fi
 
 # Create python symoblic link
 rm -f ${DRIVER_PATH}/python
-ln -s venv/bin/python ${DRIVER_PATH}/python
+ln -s ${VENV_PYTHON} ${DRIVER_PATH}/python
```
seanshpark, line 82: I'd like to see how this is... with `ls -l`, is it like this? if with full path, can you leave it as-is ?
```
python -> venv/bin/python*
```

seanshpark (approved):
LGTM thank you!

PR #6811, opened 2021-05-13
[res] Introduce Net_Conv_Min_Relu_000 recipe

seanshpark:
This will introduce Net_Conv_Min_Relu_000 recipe to check transformation
of Min(6)+ReLU to ReLU6.


seanshpark:
For #6714, from draft #6797

mhs4670go (approved):
LGTM

PR #6817, opened 2021-05-14
[circle2circle] Enable remove_fakequant

seanshpark:
This will enable remove_fakequant transformation option.


seanshpark:
For #6801, from draft #6802

mhs4670go (approved):
LGTM

PR #6831, opened 2021-05-17
[luci-eval-driver] Change install dir

mhs4670go:
This commit changes install dir from bin to test.


compiler/luci-eval-driver/CMakeLists.txt
```diff
@@ -11,4 +11,4 @@ target_link_libraries(luci_eval_driver PRIVATE luci_lang)
 target_link_libraries(luci_eval_driver PRIVATE luci_interpreter)
 target_link_libraries(luci_eval_driver PRIVATE safemain)
 
-install(TARGETS luci_eval_driver DESTINATION bin)
+install(TARGETS luci_eval_driver DESTINATION test)
```
seanshpark, line 14: Can you leave this as `bin` but to include to `test` package?

Issue #6832, opened 2021-05-17
Compiler FE: Turn off I/O for convert_nchw_to_nhwc option


seanshpark:
Currnet `convert_nchw_to_nhwc` converts input and output to NHWC also.
As this option is from ONNX and this conversion requires input data to be converted also which seems not correct for the user.

Let's revise `nchw_to_nhwc_preserve_input_shape` to `nchw_to_nhwc_input_shape` and 
`nchw_to_nhwc_preserve_output_shape` to `nchw_to_nhwc_output_shape`
so that default feature preserves as NCHW for input and output.

seanshpark:
From https://github.com/Samsung/ONE/issues/6828#issuecomment-841967126

seanshpark:
I'm not sure `nchw_to_nhwc_input_shape` and `nchw_to_nhwc_output_shape` is appropriate ...
It is OK to provide better name for this.

seanshpark:
done

PR #6840, opened 2021-05-18
[luci] Introduce NCHW to NHWC convert sub option enums

seanshpark:
This will introduce two sub option enums for ConvertNCHWToNHWC to
convert also with I/O that will replace 'preserve' sub options.

### DIFF
--- a/compiler/luci/pass/include/luci/CircleOptimizer.h
+++ b/compiler/luci/pass/include/luci/CircleOptimizer.h
@@ -87,8 +87,11 @@ public:
       Sparsify_block_map,
 
       // convert NCHW to NHWC
+      // TODO remove preserve options
       NCHW_to_NHWC_preserve_input_shape,
       NCHW_to_NHWC_preserve_output_shape,
+      NCHW_to_NHWC_input_shape,
+      NCHW_to_NHWC_output_shape,
     };
 
     virtual ~Options() = default;


seanshpark:
For #6832

jinevening:
Do we have to add new options?

Why not change the default value of `--nchw_to_nhwc_preserve_input_shape` to true in `Circle2Circle.cpp`?

seanshpark:
> Do we have to add new options?
> Why not change the default value of --nchw_to_nhwc_preserve_input_shape to true in Circle2Circle.cpp?

That can be an option.
This point is from how the user will think for this feature.
As of default, compiler will convert nodes from NCHW to NHWC but not inputs nor outputs.
From the view, "provide an option to make I/O conversion" will directly translates to 
"NCHW_to_NHWC_input_shape" and "NCHW_to_NHWC_output_shape"

But foir "..preserve..." option, user has to think twice for the option.

And with compilation option, we have 
```
circle2circle --nchw_to_nhwc_preserve_input_shape
```
to turn it off, it would be something like (not sure this is the way...)
```
circle2circle --nchw_to_nhwc_preserve_input_shape 0
```
or in `cfg` file
```
nchw_to_nhwc_preserve_input_shape=False
```
which is a twist...

So, for user interface simplicity, IMHO it seems better with `nchw_to_nhwc_input_shape`

jinevening:
I see. Thanks for the detailed explanation.

jinevening (approved):
LGTM

PR #6845, opened 2021-05-18
[debian] Revise postrm script

mhs4670go:
This commit revises postrm script.

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -2,4 +2,17 @@
 
 set -e
 
-rm -rf /usr/share/one/
+case "$1" in
+  remove|purge)
+    rm -rf /usr/share/one/
+  ;;
+  upgrade)
+    # DO NOTHING
+  ;;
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

mhs4670go:
@seanshpark After this PR is landed, Upgrade procedure'd work out okay.

jyoungyun (approved):
LGTM

mhs4670go:
FYI, [this document](https://wiki.debian.org/MaintainerScripts) shows how the maintainer scripts work.

When upgrading the package, here's the step of it.
- old.prerm upgrade
- new.preinst upgrade
- Files are unpacked
- old.postrm upgrade # here
- Old files are deleted
- new.postinst upgrade

Before this PR, `postrm` script always removes `/usr/share/one/` folder, which causes the unpacked files to be removed. So, I'll change this not to remove the folder when `upgrade` argument is given to `postrm`.

PR #6847, opened 2021-05-18
[luci] Remove nchw_to_nhwc preserve options

seanshpark:
This will remove not used anymore nchw_to_nhwc preserve options.


seanshpark:
For #6832

mhs4670go (approved):
LGTM

Issue #6842, opened 2021-05-18
[one-cmds] Fix arser bool with 'False'


seanshpark:
```cxx
if (arser.get<bool>("--someoption"))
```
doesn't seem to work (return `False`) when cfg file has
```
someoption=False
```

that is `if (arser.get<bool>("--option"))` seems to be always true.

seanshpark:
Found this while working with #6832

seanshpark:
CC @mhs4670go

mhs4670go:
@seanshpark I think that `someoption` has 0 nargs. It means, there is no need to give value to that option.

```bash
$ ./driver --someoption // arser.get<bool>("--someoption") == true
$ ./driver                      // arser.get<bool>("--someoption") == false
```

mhs4670go:
Hmm.. To make this more intuitive, the point where cfg is evaluated should be changed. I'm gonna check that out.

seanshpark:
Fixed! thank you!

PR #6849, opened 2021-05-18
[one-cmds] Revise the way one-optimize receive options

mhs4670go:
This commit revises the way one-optimize receive options.

Related: #6842 

### DIFF
--- a/compiler/one-cmds/utils.py
+++ b/compiler/one-cmds/utils.py
@@ -174,10 +174,12 @@ def _make_circle2circle_cmd(args, driver_path, input_path, output_path):
     # profiling
     if _is_valid_attr(args, 'generate_profile_data'):
         cmd.append('--generate_profile_data')
-    # optimization pass
+    # optimization pass(only true/false options)
+    # TODO support options whose number of arguments is more than zero
     for opt in _CONSTANT.OPTIMIZATION_OPTS:
-        if _is_valid_attr(args, opt[0]):
-            cmd.append('--' + opt[0])
+        if _is_valid_attr(args, opt[0]) and type(getattr(args, opt[0])) is str:
+            if not getattr(args, opt[0]).lower() in ['false', '0', 'n']:
+                cmd.append('--' + opt[0])
 
     return cmd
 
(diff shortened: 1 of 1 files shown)


compiler/one-cmds/utils.py
```diff
@@ -80,11 +80,10 @@ class _CONSTANT:
...
     # version
-    parser.add_argument(
-        '-v',
-        '--version',
-        action='store_true',
-        help='show program\'s version number and exit')
+    parser.add_argument('-v',
+                        '--version',
+                        action='store_true',
+                        help='show program\'s version number and exit')
```
mhs4670go, line 86: This is because of format checker.. I don't know why..

seanshpark (dismissed):
LGTM thank you!

compiler/one-cmds/utils.py
```diff
@@ -174,10 +174,17 @@ def _make_circle2circle_cmd(args, driver_path, input_path, output_path):
     # profiling
     if _is_valid_attr(args, 'generate_profile_data'):
         cmd.append('--generate_profile_data')
-    # optimization pass
+    # optimization pass(only true/false options)
+    # TODO support options whose number of arguments is more than zero
     for opt in _CONSTANT.OPTIMIZATION_OPTS:
         if _is_valid_attr(args, opt[0]):
-            cmd.append('--' + opt[0])
+            # ./driver --opt[0]
```
mhs4670go, line 181: This is for command line interface

compiler/one-cmds/utils.py
```diff
@@ -174,10 +174,17 @@ def _make_circle2circle_cmd(args, driver_path, input_path, output_path):
...
+            if type(getattr(args, opt[0])) is bool:
+                cmd.append('--' + opt[0])
+            # ./driver --opt[0] true/false
```
mhs4670go, line 184: This is for config file interface

mhs4670go:
@seanshpark PTAL:)

compiler/one-cmds/utils.py
seanshpark: `./driver --opt[0] true/false` seems a bit confusing as not sure this means like `./driver --SomeOption True` or to understand as `SomeOption` can be boolean and `false` can be given.
How about this?
```
# This condition check is for config file interface, usually would be
#    SomeOption=True
# but user can write as follows while development
#    SomeOption=False
# instead of removing SomeOption option
```

seanshpark (approved):
LGTM thank you!

Issue #6854, opened 2021-05-18
[one-cmds] Make parser case-sensitive


mhs4670go:
## What

Let's make configparser preserve the case.

## Why

Actually, the keys in sections of configparser are case-insensitive. But I've missed `--O1` case. I think it would be better to make the configparser preserve the case rather than changing `O1` to `o1`.

seanshpark:
@mhs4670go , we now have `2021...` in `Projects`. Please use them instead of old `2020 ... `

mhs4670go:
@seanshpark Oh, sorry. I didn't know 2021 exists as it always show me `Recent` tap. I'll keep it in mind.

PR #6855, opened 2021-05-18
[one-cmds] Make configparser preserve the case

mhs4670go:
This commit makes configparser preserve the case.

Related: #6854 


seanshpark (approved):
LGTM

seanshpark:
https://docs.python.org/ko/3.6/library/configparser.html

---

You don’t need to subclass the parser to use this method, you can also set it on an instance, to a function that takes a string argument and returns a string. Setting it to `str`, for example, would make option names case sensitive:

```
cfgparser = ConfigParser()
cfgparser.optionxform = str
```

PR #6859, opened 2021-05-20
[tfldump] Restore ostream state

mhs4670go:
This commit restores ostream state of boolplpha.


seanshpark (approved):
LGTM

PR #6862, opened 2021-05-20
Update NN_Runtime TCM configuration to exclude base TC set in anlysis

lemmaa:
- Exclude base TC set from TCM in order to bypass the limitation of not
  considering base TC count while generating analysis report.

Signed-off-by: Sung-Jae Lee <sj925.lee@samsung.com>


mhs4670go (approved):
LGTM

PR #6877, opened 2021-05-20
[debian] Change distribution value

mhs4670go:
This commit changes distribution value to stable one.


jyoungyun (approved):
LGTM

PR #6898, opened 2021-05-28
[circle-quantizer] Manually set scale/zp of special Ops

jinevening:
This manually sets scale/zp of special Ops.


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -525,6 +525,58 @@ void set_bias(luci::CircleNode *node, luci::CircleConst *bias)
                              "fully-connected layer have bias");
 }
 
+void set_act_qparam(luci::CircleNode *node, float scale, int64_t zp)
+{
+  auto qparam = node->quantparam();
```
seanshpark, line 530: no need to check `qparam` nullptr or not?

seanshpark (approved):
LGTM thank you!

PR #6901, opened 2021-05-28
[infra] Retry if external source download fails

llFreetimell:
This commit enable retrying when external source download is failed.


llFreetimell:
### For reviewers

This PR is created because some of external sources are frequently failed recently.
Although this PR does not solve all the download failure issue, I expect that possibility of external source download error would be much decreased.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #6903, opened 2021-05-31
[luci] Quantize const input before bias quantization

jinevening:
This quantizes const input before bias is quantized.


seanshpark (approved):
LGTM

Issue #5695, opened 2021-01-14
Compiler FE: Support QUANTIZE Op


seanshpark:
Let's support `QUANTIZE` Op: from #1606
- This is set to `DEQUATIZE` Op
- `fake_quant_with_min_max_~` will drop `QUANTIZE-DEQUANTIZE` set
   - old version with experimental option drops `FAKE_QUANT` Op

Old issue: #2499

seanshpark:
Now, let's check with out in-house model has any problems...

seanshpark:
circle2circle passes as there is no special actions performed.
but `record-minmax` stops with `ERROR: Unsupported operator.`

PR #6916, opened 2021-06-02
DRAFT Support QUANTIZE Op

seanshpark:
on-going draft to support QUANTIZE Op for circle2circle.


seanshpark:
For #5695

seanshpark:
all landed

PR #6920, opened 2021-06-02
[res] Introduce Quantize recipe

seanshpark:
This will introduce Quantize recipe for testing Quantize Op.


seanshpark:
For #5695, from draft #6916

mhs4670go (approved):
LGTM

PR #6922, opened 2021-06-02
[luci] Enable logex of CircleQuantize

seanshpark:
This will enable logex to dump CircleQuantize.


seanshpark:
For #5695, from draft #6916

mhs4670go (approved):
LGTM

PR #6926, opened 2021-06-02
[luci] Reviese shape/dtype inference for CircleQuantize

seanshpark:
This will revise shape and dtype inference for CircleQuantize.


seanshpark:
For #5695, from draft #6916

PR #6927, opened 2021-06-02
[luci] Connection of CircleQuantize for partition

seanshpark:
This will enable connection of CircleQuantize for model partition.


seanshpark:
For #5695, from draft #6916

mhs4670go (approved):
LGTM

Issue #6931, opened 2021-06-03
Compiler FE: Remove Quantize-Dequantize sequence


seanshpark:
Let's provide a pass to remove `Quantize-Dequantize` Op sequence for experimental usage
- as our quantization algorithm is not like that of TensorFlow lite, we may not needed these ops.

This is like #6801 task.

seanshpark:
done

PR #6932, opened 2021-06-03
DRAFT: Introduce remove_quantdequant

seanshpark:
on-going draft ot support remove_quantdequant option.


seanshpark:
For #6931

seanshpark:
all landed

Issue #5799, opened 2021-01-25
[discussion] High throughput support by automatic model-partition & multithreading 


chunseoklee:
It is planning to support high throughput inference by  auto-partitioning model and executing them parallel in a session.
This issue will track the progress w.r.t the goal. 
Any suggestion is welcome.

## Goal 

Provide users to achieve the high throughput inference

## Todo

- Given (AOT) partitioned models, run the models in a session(note that @intom is working on this)
   + define a proper feature to support this on nnpkg
- Given a single model, partition this model and execute it in parallel in a session.

chunseoklee:
cc @intom

seanshpark:
> Given (AOT) partitioned models, 

Just curious, who or what gives this partitioned models?

chunseoklee:
> > Given (AOT) partitioned models,
> 
> Just curious, who or what gives this partitioned models?

For now, we will use manually made model. However, it is better to provide a tool. for this if needed.

intom:
### Asynchronous Run & Graph Partition


### Requirements
 - Optimal Graph Partitioning
 - Build Partial Graph based on the predefined partitioning map
 - Asynchronous Run for each partial graph


### Objectives
 - Enhance processing throughput for streaming inputs
   : Partitioning graph according to the purpose
   : Supporting for Asynchronous Run and Pipelining

 
###  Policy
- Create the map file for partitioning graph with script
- The predefined file should be in the model package 
--> The above policy is only one of candidates and currently local implementation. Give your opinions. 

![image](https://user-images.githubusercontent.com/41365503/117231758-d217d900-ae5a-11eb-873e-ab5561d28a50.png)


![image](https://user-images.githubusercontent.com/41365503/117231838-fe335a00-ae5a-11eb-81dc-d9e191dd4d44.png)

PR #6937, opened 2021-06-03
[onert] Add mutex for multi-threading

intom:
related issue : #5799

A segmentation fault occurs when the quantized mobilenet model is executed using different threads.

Add mutex to use gemmlowp in different threads.

Signed-off-by: sungho22.lee <sungho22.lee@samsung.com>

### DIFF
--- a/compute/cker/include/cker/operation/optimized/Conv.h
+++ b/compute/cker/include/cker/operation/optimized/Conv.h
@@ -42,6 +42,8 @@ namespace cker
 namespace optimized
 {
 
+std::mutex _gemmlowp_mutex;
+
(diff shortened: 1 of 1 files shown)


chunseoklee:
@intom It looks like working. But it will block the other thread's run while running in parallel ?

intom:
> @intom It looks like working. But it will block the other thread's run while running in parallel ?

right, but it is only the lock per operation function. 
I think multiple contexts to avoid blocking are inefficient currently.

chunseoklee (approved):

hseok-oh (approved):

Issue #6938, opened 2021-06-03
[cone-cmds] option to drop tflite file in output folder


seanshpark:
While debugging the compilation with _one-build_, it would be nice if we could access intermediate files,
which are generated in system temp folder and removed after used.

Tools that generate intermediate files
- _one-import-bcq_ --> maybe later
- _one-import-onnx_
- _one-import-tf_

seanshpark:
done

PR #6941, opened 2021-06-03
DRAFT one-cmds save intermediate files

seanshpark:
On-going draft to enable save intermediate files while model conversion.


compiler/one-cmds/one-import-onnx
```diff
@@ -61,6 +61,14 @@ def _get_parser():
     tf2tfliteV2_group.add_argument('--model_format', default='saved_model')
     tf2tfliteV2_group.add_argument('--converter_version', default='v2')
 
+    # save intermediate file(s)
+    parser.add_argument(
+        '--save_intermediate',
+        action='store_const',
+        dest='save_intermediate',
+        const='--save_intermediate',
+        help='Save intermediate files to output folder')
```
mhs4670go, line 70: If it just stores boolean value, then it can be like below.
```cxx
converter_version.add_argument(
        "--save_intermediate", action="store_true", help="Save intermediate files to output folder")
```

seanshpark:
For #6938

compiler/one-cmds/one-import-onnx
seanshpark: Ah... got this error
```
TypeError: __init__() missing 1 required positional argument: 'const'
```

mhs4670go: @seanshpark Apparently,, that error occurs when `action='store_const'`. Could you try one more time with `action='store_true'`?

seanshpark: This works :)

seanshpark:
all necessary changes landed

PR #6946, opened 2021-06-03
[one-cmds] Fix remove log file

seanshpark:
This will fix remove log file with correct name.
- also remove rm command when there is no log generated


seanshpark:
Found this while working on draft #6941

mhs4670go (approved):
LGTM

Issue #6948, opened 2021-06-04
Compiler FE: Reduce LoC SAM violation


seanshpark:
Let's reduce violation values of SAM LoC values.

As of 2021-06-02:

No. | Metric | Path | Package | Class | LineNo | Value | ClassType | Violation
-- | -- | -- | -- | -- | -- | -- | -- | --
1 | LOC | compiler/luci/service/src/CircleCloneNode.h | &nbsp; | CloneNode | 27 | 948 | CLASS | Y
2 | LOC | compiler/luci-interpreter/src/loader/KernelBuilder.h | &nbsp; | KernelBuilder | 32 | 755 | CLASS | Y
3 | LOC | compiler/luci/export/src/CircleOperationExporter.cpp | &nbsp; | OperationExporter | 609 | 748 | CLASS | Y
4 | LOC | compiler/luci/logex/src/FormattedGraph.cpp | &nbsp; | CircleNodeSummaryBuilder | 194 | 699 | CLASS | Y

seanshpark:
These are huge because of using visitor pattern with more than 100 circle IR nodes.

seanshpark:
AHUB report before change
```
4.08 | 4 | 0.85 | 5.22 %
```

seanshpark:
https://github.com/Samsung/ONE/pull/6949/commits/91e14c6fd6af942aeb1ae918e76ab697b92fd97e

```
4.16 | 4 | 0.85 | 5.14 %
```
Violation value: 902

seanshpark:
With #6949 , https://github.com/Samsung/ONE/commit/3ef78ba1fdeab6ad7475a3eba5929526e8c6f8a3

```
4.43 | 3 | 0.64 | 3.65 %
```
Not listed violation item

PR #6949, opened 2021-06-04
DRAFT Refactor CloneNode to reduce LoC

seanshpark:
on-going draft to reduce LoC violation value of CloneNode class.


seanshpark:
For #6948

PR #6951, opened 2021-06-04
[Infra/docker] Upgrade git version

hseok-oh:
Let's upgrade git version on bionic docker.

Github checkout action requires v2.18, but ubuntu bionic's default git version is v2.17.1.


seanshpark (approved):
LGTM

Issue #6952, opened 2021-06-04
[Compiler FE] Preserve initial nodes list in source_table for profiling


llFreetimell:
### What?

Let's export `source_table` metadata with following criteria
- If `source_table` metadata is originally included in circle model, export it.
- Otherwise, export `source_table` with initially imported nodes list.

### Why?

Currently, `source_table` is updated every moment when circle graph is exported.
(For example, if there were `node1` and `node2` but `node` is removed by optimization, only `node2` is exported to `source_table`).
However, for profiling, `source_table` with all of initial nodes is more meaningful.

### Todo

- [x] Introduce `source_table` in `luci::Module` #6962
- [x] Introduce getter of `source_table` in `luci::ImportMetadata` #6963
- [x] Initialize `Module::source_table` #6965
- [x] Export `Module::source_table` instead of current new `source_table`. #6966 
- [x] Deprecated unused functions #6967 

---

Draft : #6953

/cc @ejjeong

llFreetimell:
With `Net_Squeeze_Squeeze_000.circle` (in `res/TensorFlowLiteRecipes/Net_Squeeze_Squeeze_000/`).

### Command

```bash
circle2circle \
--generate_profile_data \
--substitute_squeeze_to_reshape \
--remove_redundant_reshape \
build/compiler/common-artifacts/Net_Squeeze_Squeeze_000.circle \
Net_Squeeze_Squeeze_000.opt.circle
```

### Before

```bash
$ circledump Net_Squeeze_Squeeze_000.opt.circle
...
metadata : B(index) name
B(2) ONE_source_table
    [node_id : node_name]
    [1 : ofm]
B(3) ONE_op_table
    [node_id : origin_ids]
    [0 : 1]
```

### After

```
$ circledump Net_Squeeze_Squeeze_000.opt.circle
...
metadata : B(index) name
B(2) ONE_source_table
    [node_id : node_name]
    [0 : t1]
    [1 : ofm]
B(3) ONE_op_table
    [node_id : origin_ids]
    [0 : 1]
```

llFreetimell:
Done!

PR #6953, opened 2021-06-04
[Draft][luci] Preserve originally imported profiling data

llFreetimell:
Parent Issue : #6952 

This draft is for preserving originally imported profiling data, `source_table`.

### DIFF
--- a/compiler/luci/import/src/Importer.cpp
+++ b/compiler/luci/import/src/Importer.cpp
@@ -315,6 +315,35 @@ std::unique_ptr<Module> Importer::importModule(const circle::Model *model) const
 
   post_import_graph(module.get(), reader);
 
+  // Initialize 'source_table'
+  auto circle_metadata = std::make_unique<luci::CircleImportMetadata>(reader);
+  if (circle_metadata->source_table().size() > 0)
+  {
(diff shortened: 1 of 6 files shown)


compiler/luci/import/src/Importer.cpp
```diff
@@ -315,6 +315,35 @@ std::unique_ptr<Module> Importer::importModule(const circle::Model *model) const
 
   post_import_graph(module.get(), reader);
 
+  // Initialize 'source_table'
```
mhs4670go, line 318: I think the validation is missing. These codes are inserted after origins are already created and there could be differences between the node-ids that are created when they are imported and the one that are from circle model's source table that exists in binary itself.

There are two situations. One is when circle model doesn't have the source table(`circle_metadata->source_table().size() == 0`). And, the second is the one who has it(`circle_metadata->source_table().size() > 0`). Former has no problem because node-ids are generated at the time of the importing model. But, IIUC, latter has a problem. There is no guarantee that node-id that is generated above and the one from `source_table` are same because previous model might have been optimized as what you said.

So, if circle model has `source_table`, these should be guaranteed.
- the number of subgraphs is 1.
- set node id to the one from `source_table` not from its importing order.

llFreetimell:

PR #6957, opened 2021-06-07
Group CloneNode to small subclass for GHIJ

seanshpark:
This will group CloneNode to smaller subclass for GHIJ.


seanshpark:
For #6948, from draft #6949

mhs4670go (approved):
LGTM

PR #6964, opened 2021-06-08
[one-cmds] Prevent reverse patch for onnx

seanshpark:
This will prevent apply patch for onnx when patch is already applied.


mhs4670go (approved):
LGTM

Issue #6970, opened 2021-06-09
[luci] Quantize small weights to zero


Bronnikoff:
### What?

Let's quantize small weight values to 0.

### Why?

Small weights values produces small scales for weights and, therefore, small scales for biases with large quantized value:
https://github.com/Samsung/ONE/blob/2a3e35b1d3c01a850546a6067e1359313e929c5a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp#L377-L381

Rounding errors may be significant for values that are too low!

For example, let's take a look at MobileNetV2, which could be obtained using:

```
$ wget https://storage.googleapis.com/download.tensorflow.org/models/tflite_11_05_08/mobilenet_v2_1.0_224.tgz
$ tar xvzf mobilenet_v2_1.0_224.tgz
```

Filter №3  of first Conv2d layer of this model has low weights values:
![image](https://user-images.githubusercontent.com/43731397/121377377-f6c80a80-c94a-11eb-90b3-ac7a7c39af5c.png)
And bias equal to `-1.2156529426574707`.

After 16bit quantiztion quantized bias value equal to `INT64_MIN` with low scale:  `7.617614820112793e-28`, produced by low scales of input and weights. Dequantized value is very different from the original value:

```python
>>> 7.617614820112793e-28 * -9223372036854775808
-7.026009551935886e-09
```

One possible solution is quantize small values to 0 with scale equal to 1.0 (or may be some another value?).

Bronnikoff:
With changes from #6971 small weights quantizes to 0 with scale `1.0`. Bias value for example from https://github.com/Samsung/ONE/issues/6970#issue-916328038 equal to `-7967` with scale `0.0001525924017187208`. 
Dequantized value:
```python
>>> 0.0001525924017187208 * -7967
-1.2157036644930486
```

binarman:
+cc @jinevening

jinevening:
We've faced a similar issue in uint8 quantization.

We decided to use a minimum value for scale to prevent overflow, i.e., scale is set to `1e-5` if it is less than `1e-5`.
(`1e-5` was determined by some internal experiments.)

See `compute_asym_scale_zp`
```cpp
  // protect scale from being very low due to overflow
  if (scale < 1e-5)
  {
    scale = 1e-5;
    nudged_zero_point = fp32_to_uint8_cast(std::round(qmin_double - rmin / scale));
  }
```
[...]

PR #6971, opened 2021-06-09
[luci] Quantize low weights to 16-bit zero

Bronnikoff:
This commit adds 16-bit quantization of low weights values to zero in QuantizeDequantizeWeightsPass.

-----------------

For: #6970

### DIFF
--- a/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
+++ b/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
@@ -124,6 +124,14 @@ void sym_wquant_per_channel(CircleConst *node, std::vector<float> &min, std::vec
 
   auto quantize = [&](uint32_t *indices, loco::TensorShape &dimension, int channel_dim_index) {
     int channel_idx = indices[channel_dim_index];
+
+    // Small scale factor produced by small weights, which can be quantized to 0 with scale 1.0f
+    // TODO find better way to fix very low scales
+    if (scaling_factor[channel_idx] < std::numeric_limits<float>::epsilon())
+    {
+      scaling_factor[channel_idx] = 1.0f;
+    }
+
     const float scaling_factor_inv = 1.0 / scaling_factor[channel_idx];
     auto data = node->at<loco::DataType::FLOAT32>(cal_offset(dimension, indices));
     data = data < nudged_min[channel_idx] ? nudged_min[channel_idx] : data;


compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -124,6 +124,13 @@ void sym_wquant_per_channel(CircleConst *node, std::vector<float> &min, std::vec
 
   auto quantize = [&](uint32_t *indices, loco::TensorShape &dimension, int channel_dim_index) {
     int channel_idx = indices[channel_dim_index];
+
+    // Small scale factor produced by small weights, which can be quantized to 0 with scale 1.0f
+    if (scaling_factor[channel_idx] < std::numeric_limits<float>::epsilon())
+    {
+      scaling_factor[channel_idx] = 1.0f;
```
Bronnikoff, line 131: Scale equal to 1.0 with small weight values produce zero in code bellow:

https://github.com/Samsung/ONE/blob/898ca6995b5b2b46fa074a8f098f6725327a6f39/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp#L138

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -124,6 +124,13 @@ void sym_wquant_per_channel(CircleConst *node, std::vector<float> &min, std::vec
 
   auto quantize = [&](uint32_t *indices, loco::TensorShape &dimension, int channel_dim_index) {
     int channel_idx = indices[channel_dim_index];
+
+    // Small scale factor produced by small weights, which can be quantized to 0 with scale 1.0f
+    if (scaling_factor[channel_idx] < std::numeric_limits<float>::epsilon())
```
Bronnikoff, line 129: I'm not sure it's best condition for rounding to 0. May be better to use value lower than `epsilon()`?

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
AShedko: `scaling_factor_inv == 1.0f` for `scaling_factor[channel_idx] < std::numeric_limits<float>::epsilon()`.
So `data * 1.0f` is `data`. Is this what you had in mind?

AShedko (approved):

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
binarman: I also find this design confusing.
If you want to quantize small small weights to 0, it makes sense to do it below, like
```c++
    if (scaling_factor < epsilon)
    {
      quantized_values[cal_offset(dimension, indices)] =
        static_cast<int32_t>(std::round(data * scaling_factor_inv));
    }
```

Bronnikoff: > So `data * 1.0f` is `data`. Is this what you had in mind?

Yes, I mean `data * scale = data` in cases of very low scales, but very low scales can be produced only by low weights. Therefore `round(inv_scale * data) == round(data) == 0`. =)

Bronnikoff: > I have concerns that messing with scales could lead to some unexpected side effects.

Low scales of weights produces very low scales of biases and quantized biases may have large quantization error (details: https://github.com/Samsung/ONE/issues/6970#issue-916328038)
This a reason to change scale here =)

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
Bronnikoff: I added TODO comment. I think better to find better design latter!

Bronnikoff:
Close because another solution found: #7063

PR #6972, opened 2021-06-10
Fix coverity 1229862 : uinit variable in PoolParams

chunseoklee:
Add initilization for activation.

### DIFF
--- a/compute/cker/include/cker/Types.h
+++ b/compute/cker/include/cker/Types.h
@@ -80,8 +80,6 @@ enum class BroadcastableOpCategory : uint8_t
 
 struct PoolParams
 {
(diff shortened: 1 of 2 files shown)


chunseoklee:
@hseok-oh Please share your opinion since the patch seems weird for me. activation is not used anywhere in ONE. How about removing this from PoolParams.

hseok-oh:
@chunseoklee 

>How about removing this from PoolParams.

Agree. We don't use this field now, and tflite v2.3 also don't use this field.

chunseoklee:
close since this PR is based on samsung/ONE's test branch, not on chunseoklee's local branch

Issue #6974, opened 2021-06-10
[circle2circle] Change output node


seanshpark:
Some in-house models have `CAST` at the end of the model, which our _luci-interpreter_ doesn't support yet.

To go around this situation,
if we could change the output node to the one before the `CAST`,
then we could at least test the model with the sub network.

And with this new feature, we can test the model without cutting the model with other tools.

mhs4670go:
@seanshpark Codes look fine to me. I just wonder it can be introduced to `circle2circle` which uses for circle graph optimization. IMHO, this might be introduced to another tool that modify the graph.

seanshpark:
>  I just wonder it can be introduced to circle2circle which uses for circle graph optimization.

it will be introduced in `circle2circle`, you can see in DRAFT code.

seanshpark:
Done for now

PR #6976, opened 2021-06-10
DRAFT [luci] change output nodes

seanshpark:
on-going draft to change output nodes by name

ONE-DCO-1.0-ONE-DCO-1.0-ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>


seanshpark:
For #6974

seanshpark:
all landed

PR #6977, opened 2021-06-10
[luci] Verify bias only when FullyConnected has it

llFreetimell:
Since `bias` is optional, sometimes `CircleFullyConnected` does not have `bias`.
However, current verification code did not consider it.
This commit will fix it.


jinevening (approved):
LGTM Thank you

seanshpark (approved):
LGTM

PR #6980, opened 2021-06-14
[luci] Introduce change_outputs

seanshpark:
This will introduce change_outputs experimental method of luci to
change output nodes by node name.
- only work for first subgraph (main graph)


seanshpark:
For #6974, from draft #6976

mhs4670go (approved):
LGTM

Issue #6981, opened 2021-06-14
Compiler FE: Reduce LoC SAM violation for CircleNodeSummaryBuilder


seanshpark:
Let's reduce SAM violation value for `CircleNodeSummaryBuilder`

Current (as of now)

No. | Metric | Path | Package | Class | LineNo | Value | ClassType | Violation
-- | -- | -- | -- | -- | -- | -- | -- | --
3 | LOC | compiler/luci/logex/src/FormattedGraph.cpp | &nbsp; | CircleNodeSummaryBuilder | 194 | 699 | CLASS | Y

seanshpark:
No. | Metric | Path | Package | Class | LineNo | Value | ClassType | Violation
-- | -- | -- | -- | -- | -- | -- | -- | --
3 | LOC | compiler/luci/logex/src/FormattedGraph.cpp | &nbsp; | CircleNodeSummaryBuilder | 194 | 613 | CLASS | Y

Extracting CircleNodes starting with A,B,C (https://github.com/Samsung/ONE/pull/6982/commits/1534ff66d4a72371d5eb91fdb5a53ef830ef0713) gave 699 -> 613
Seems working.

seanshpark:
Full draft doesn't show `FormattedGraph.cpp`

PR #6982, opened 2021-06-14
DRAFT: Reduce Class LoC of FormattedGraph

seanshpark:
On-going draft to reduce class LoC of FormattedGraph.


seanshpark:
For #6981

seanshpark:
all landed

PR #6984, opened 2021-06-14
[circle2circle] Support change_outputs option

seanshpark:
This will add change_outputs option to change outputs of the model.


seanshpark:
For #6974, from draft #6976

mhs4670go (approved):
LGTM

PR #6989, opened 2021-06-14
[luci] Fix CloneNode typo in enum CN

seanshpark:
This will fix CloneNode typo in enum class CN as WXYZ as V is declared two times.


seanshpark:
Found this while working on #6982

mhs4670go (approved):
LGTM

PR #6990, opened 2021-06-14
[luci] Extract to SummaryBuilderLet with group ABC

seanshpark:
group ABC CircleNodes to reduce class LoC violation.


seanshpark:
For #6981, from draft #6982

mhs4670go (approved):
LGTM

PR #6991, opened 2021-06-14
[luci] Extract to SummaryBuilderLet with group DEF

seanshpark:
group DEF CircleNodes to reduce class LoC violation.


seanshpark:
For #6981, from draft #6982

mhs4670go:
@nnfw-bot test nncc-release

mhs4670go (approved):
LGTM

PR #6998, opened 2021-06-15
DRAFT: Reduce class LoC of OperationExporter

seanshpark:
on-going draft to reduce class LoC of OperationExporter.


seanshpark:
For #6999

PR #7001, opened 2021-06-15
[luci] Extract to SummaryBuilderLet with group KLMN

seanshpark:
group KLMN CircleNodes to reduce class LoC violation.

### DIFF
--- a/compiler/luci/logex/src/FormattedGraph.cpp
+++ b/compiler/luci/logex/src/FormattedGraph.cpp
@@ -241,6 +241,7 @@ private:
   IMPLEMENT(luci::CircleGreaterEqual)
   IMPLEMENT(luci::CircleIf)
 #endif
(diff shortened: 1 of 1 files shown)


compiler/luci/logex/src/FormattedGraph.cpp
```diff
@@ -2110,6 +2150,8 @@ bool NodeSummaryBuilder::build(const loco::Node *node, locop::NodeSummary &s) co
 
   BUILD_GRP(ABC);
   BUILD_GRP(DEF);
+  BUILD_GRP(GHIJ);
```
seanshpark, line 2153: Forgot this in previous PR -_-;

seanshpark:
For #6981, from draft #6982

mhs4670go (approved):
LGTM

Issue #6999, opened 2021-06-15
Compiler FE: Reduce class LoC for CircleOperationExporter.cpp


seanshpark:
Let's reduce class LoC violation value for CircleOperationExporter.cpp file.

No. | Metric | Path | Package | Class | LineNo | Value | ClassType | Violation
-- | -- | -- | -- | -- | -- | -- | -- | --
2 | LOC | compiler/luci/export/src/CircleOperationExporter.cpp | &nbsp; | OperationExporter | 609 | 748 | CLASS | Y

seanshpark:
done

PR #7010, opened 2021-06-15
[luci] Extract ExportHelper

seanshpark:
This will extract ExportHelper as separate class from OperationExporter.


seanshpark:
For #6999, from draft #6998

mhs4670go (approved):
LGTM

PR #7008, opened 2021-06-15
[luci/profile] Add negative origin test

cgbahk:
This commit adds negative test for origin: the case when adding null
origin.

ONE-DCO-1.0-Signed-off-by: Cheongyo Bahk <cg.bahk@gmail.com>

---

Testing with CI as I have no local build env :sob:

PR #7013, opened 2021-06-15
[luci/profile] Update logic of has_origin

llFreetimell:
Originated from #7008

Until now, `has_origin` returned `true` when `sources()` is not empty.
However, empty `sources()` actually means that there are no origin information.
So this commit will update `has_origin` to return `false` when `sources()` is empty.

### DIFF
--- a/compiler/luci/profile/src/CircleNodeOrigin.test.cpp
+++ b/compiler/luci/profile/src/CircleNodeOrigin.test.cpp
@@ -106,3 +106,27 @@ TEST(LuciCircleNodeOrigin, composite_origin_empty_ctor_NEG)
 {
   ASSERT_ANY_THROW(luci::composite_origin({}));
 }
+
+TEST(LuciCircleNodeOrigin, add_null_origin_NEG)
+{
+  auto g = loco::make_graph();
+  auto add = g->nodes()->create<luci::CircleAdd>();
+
+  ASSERT_FALSE(has_origin(add));
+
+  add_origin(add, nullptr);
+
+  ASSERT_FALSE(has_origin(add));
+}
+
+TEST(LuciCircleNodeOrigin, add_empty_origin_NEG)
+{
+  auto g = loco::make_graph();
+  auto add = g->nodes()->create<luci::CircleAdd>();
+
+  ASSERT_FALSE(has_origin(add));
+
+  add_origin(add, luci::composite_origin({nullptr, nullptr}));
+
+  ASSERT_FALSE(has_origin(add));
+}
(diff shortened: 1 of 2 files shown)


compiler/luci/profile/src/CircleNodeOrigin.test.cpp
```diff
@@ -106,3 +106,15 @@ TEST(LuciCircleNodeOrigin, composite_origin_empty_ctor_NEG)
...
 }
+
+TEST(LuciCircleNodeOrigin, add_null_origin_NEG)
+{
+  auto g = loco::make_graph();
+  auto add = g->nodes()->create<luci::CircleAdd>();
+
+  ASSERT_FALSE(has_origin(add));
+
+  add_origin(add, nullptr);
```
mhs4670go, line 117: Hmmm.. looks a little bit weird. How about letting `add_origin` do nothing when second parameter is `nullptr`?

llFreetimell: It can be done but main purpose of this PR is because of following case.

```cpp
TEST(LuciCircleNodeOrigin, add_null_origin_2_NEG)
{
  auto g = loco::make_graph();
  auto add = g->nodes()->create<luci::CircleAdd>();
  auto sub = g->nodes()->create<luci::CircleSub>();

  ASSERT_FALSE(has_origin(add));

  add_origin(add, get_origin(sub)); // get_origin(sub) returns nullptr

  ASSERT_FALSE(has_origin(add));
}
```

llFreetimell: It was originated from following case and only for the case, your comment make sense.

```cpp
  auto g = loco::make_graph();
  auto add = g->nodes()->create<luci::CircleAdd>();

  // Sometimes sub has origin, but sometimes does not have
  add_origin(add, get_origin(sub));
```

However, we have following codes in `luci/pass`.
```cpp
// TransformMinReluToRelu6Pass.cpp#96
luci::add_origin(relu6, luci::composite_origin({luci::get_origin(relu), luci::get_origin(mini)}));
```
Even if `relu` and `mini` do not have origin, `luci::composite_origin({luci::get_origin(relu), luci::get_origin(mini)})` is not `nullptr`. The returned value is CompositeOrigin with empty sources.

FYI, please refer to 9323 issue of internal repo :)

BTW, I should add one more testcase...
```cpp
TEST(LuciCircleNodeOrigin, add_empty_origin_NEG)
{
  auto g = loco::make_graph();
  auto add = g->nodes()->create<luci::CircleAdd>();

  ASSERT_FALSE(has_origin(add));

  add_origin(add, luci::composite_origin({nullptr, nullptr}));

  ASSERT_FALSE(has_origin(add));
}
```

cgbahk: I have no opinion on detail design. I'm okay as long as
- `add_null_origin_NEG`(in this PR) test passed - must
- `add_empty_origin_NEG`(in above comment) test passed - optional, though looks natural to me

llFreetimell: Then let's go simple...!

I will update as
- `composite_origin(nullptr)` returns `nullptr`

Then,
- If `has_origin(node)` is `true`, `sources()` is not empty always.
- If `has_origin(node)` is `false`, node has any origin.

cgbahk (approved):
LGTM

mhs4670go (approved):
LGTM

PR #7014, opened 2021-06-15
[luci-interpreter] Sort KernelBuilder members in order

seanshpark:
This will sortKernelBuilder members in alphabetical order.


jinevening (approved):
LGTM

Issue #6993, opened 2021-06-15
Compiler FE: Reduce class LoC for KernelBuilder.h


seanshpark:
Let's reduce class LoC violation value for `KernelBuilder.h` file.

No. | Metric | Path | Package | Class | LineNo | Value | ClassType | Violation
-- | -- | -- | -- | -- | -- | -- | -- | --
1 | LOC | compiler/luci-interpreter/src/loader/KernelBuilder.h | &nbsp; | KernelBuilder | 32 | 755 | CLASS | Y

seanshpark:
done

PR #7015, opened 2021-06-15
DRAFT: Reduce class LoC of KernelBuilder

seanshpark:
on-going draft to reduce class LoC of KernelBuilder.


seanshpark:
For #6993

seanshpark:
all landed

PR #7022, opened 2021-06-16
[luci-interpreter] Extract KernelBuilderLet for group ABC

seanshpark:
This will extract KernelBuilderLet for circle IR group ABC to reduce
class LoC violation score.

ONE-DCO-1.0-Signed-off-by: SaeHie Park saehie.park@gmail.com

### DIFF
--- a/compiler/luci-interpreter/src/loader/KernelBuilder.cpp
+++ b/compiler/luci-interpreter/src/loader/KernelBuilder.cpp
@@ -106,12 +106,60 @@ std::vector<const loco::Node *> collectOutputNodes(const luci::CircleNode *node)
 namespace luci_interpreter
 {
 
-std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleNode *)
+// TODO move to anonymous namespace
+enum class KB
 {
+  ABC,
+};
+
+#define DECLARE_VISIT(CLASS) std::unique_ptr<Kernel> visit(const luci::CLASS *) override
+
+template <KB kb> class KernelBuilderLet;
+
+template <>
+class KernelBuilderLet<KB::ABC> : public luci::CircleNodeVisitor<std::unique_ptr<Kernel>>,
+                                  public KernelBuilderHelper
+{
+public:
+  KernelBuilderLet(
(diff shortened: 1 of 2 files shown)


seanshpark:
For #6993, from draft #7015

jinevening:
Does our SE tool really need this change? The code looks very unnatural to me..

compiler/luci-interpreter/src/loader/KernelBuilder.cpp
```diff
@@ -106,12 +106,60 @@ std::vector<const loco::Node *> collectOutputNodes(const luci::CircleNode *node)
...
+  DECLARE_VISIT(CircleAdd);
+  DECLARE_VISIT(CircleArgMax);
+  DECLARE_VISIT(CircleAveragePool2D);
+  DECLARE_VISIT(CircleBatchToSpaceND);
+  DECLARE_VISIT(CircleConcatenation);
+  DECLARE_VISIT(CircleConst);
+  DECLARE_VISIT(CircleConv2D);
+};
+
+#undef DECLARE
```
jinevening, line 144:
```suggestion
#undef DECLARE_VISIT
```

compiler/luci-interpreter/src/loader/KernelBuilder.cpp
```diff
@@ -106,12 +106,60 @@ std::vector<const loco::Node *> collectOutputNodes(const luci::CircleNode *node)
...
+
+std::unique_ptr<Kernel> KernelBuilder::visit(const luci::CircleNode *node)
+{
+#define VISIT_KB(GRP)                                                          \
```
jinevening, line 148: There is no #undef for `VISIT_KB`.

seanshpark:
> Does our SE tool really need this change? The code looks very unnatural to me..

Not desperately. I want to remove violation scores from our analysis report.
There are other items to remove besides LoC but didn't know how to until recent idea.
I've talked with @parjong about it and said this approach is not a hack but one of the solution.
I'm also not a fan on this approach but at least there are less change to existing codes and structual looking for me
and mostly removes from the violation item....

jinevening (approved):
LGTM

PR #7030, opened 2021-06-16
[luci-interpreter] Extract KernelBuilderLet for group OPQR

seanshpark:
This will extract KernelBuilderLet for circle IR group OPQR to
reduce class LoC violation score.


seanshpark:
For #6993, from draft #7015

jinevening (approved):
LGTM

Issue #7029, opened 2021-06-16
[luci] Fix SquaredDifference import to support Q8/Q16


seanshpark:
We need to import SquaredDifference with quantized 8/16 bit.

Current implementation:
```cxx
  switch (tensor->type)
  {
    case circle::TensorType_FLOAT16:
    case circle::TensorType_FLOAT32:
    case circle::TensorType_FLOAT64:
    case circle::TensorType_INT32:
    case circle::TensorType_INT64:
    case circle::TensorType_COMPLEX64:
      break;
    // TODO support bfloat16, complex128
    default:
      return false;
  }
```

seanshpark:
Simple test: run in `build/debug/compiler/common-artifacts`

`SquaredDifference_000-q8.cfg` :
```
[one-build]
one-import-tf=False
one-import-tflite=False
one-import-bcq=False
one-import-onnx=False
one-optimize=False
one-quantize=True
one-pack=False
one-codegen=False

[one-quantize]
input_path=./SquaredDifference_000.circle
output_path=./SquaredDifference_000.q8.circle
```
```
one-build -C SquaredDifference_000-q8.cfg
```

```
../circle2circle/circle2circle SquaredDifference_000.q8.circle SquaredDifference_000.q8.2.circle
```
->
> terminate called after throwing an instance of 'oops::UserExn'
  what():  Error: Invalid operator: SQUARED_DIFFERENCE
Aborted (core dumped)

seanshpark:
Patch landed

PR #7031, opened 2021-06-16
[luci] Supprt import U8/S16 for SquaredDifference

seanshpark:
This will fix import of SquaredDifference to support U8/S16 tensor types.


seanshpark:
For #7029

jinevening (approved):
LGTM

PR #7039, opened 2021-06-17
DRAFT Fuse InstNorm from TF 1.15.3

seanshpark:
on-going draft to support fusion of instance norm pattern from TF 1.15.3


seanshpark:
For #7039

seanshpark:
done

Issue #7032, opened 2021-06-16
[luci] Support InstanceNorm fusion from TF 1.15.3


seanshpark:
Let's support fusion of InstanceNorm from TF 1.15.3

<image src='https://user-images.githubusercontent.com/4616940/122181037-f1773c80-cec3-11eb-897c-d77753fbfbb1.png' width=300/>

seanshpark:
Use `tf.contrib.layers.instance_norm`

seanshpark:
Prepare TF 1.15.3 virtual env

```
virtualenv --system-site-packages -p python3 ./tf153venv
source ./tf153venv/bin/activate
python -m pip install --upgrade tensorflow==1.15.3
```

seanshpark:
Reuse `res/TensorFlowPythonExamples/instance_norm` in this repo...

make a copy `instance_norm_2` with DIM4 shape
```
in_ = tf.compat.v1.placeholder(dtype=tf.float32, shape=(1, 1, 1, 32), name="Hole")
norm_ = tf.contrib.layers.instance_norm(in_)
```

```
python tfpem.py instance_norm_2

freeze_graph --input_graph instance_norm_2.pbtxt \
--input_binary=false \
--input_checkpoint=./ckpt/instance_norm_2.ckpt \
--output_node_names=InstanceNorm/instancenorm/add_1 \
--output_graph instance_norm_2_fr.pbtxt

one-build -C instance_norm_2_fr.cfg
```

`instance_norm_2_fr.cfg`
```
[one-build]
one-import-tf=True
one-import-tflite=False
one-import-bcq=False
one-import-onnx=False
one-optimize=False
one-quantize=False
one-pack=False
one-codegen=False

[one-import-tf]
input_path=./instance_norm_2_fr.pbtxt
output_path=./instance_norm_2_fr.circle
input_arrays=Hole
input_shapes=1,1,1,32
output_arrays=InstanceNorm/instancenorm/add_1
converter_version=v2
save_intermediate=True
```

seanshpark:
recipe file with a little touch
```
operand {
  name: "Hole"
  type: FLOAT32
  shape {
    dim: 1
    dim: 1
    dim: 1
    dim: 32
  }
  quant {
    quantized_dimension: 0
  }
  is_variable: false
}
operand {
  name: "InstanceNorm/beta"
  type: FLOAT32
  shape {
    dim: 32
  }
  filler {
    tag: "constant"
    arg: "0"
  }
  quant {
[...]

PR #7041, opened 2021-06-17
[res] Fix comment for Net_InstanceNorm rule files

seanshpark:
This will fix comment for Net_InstanceNorm rule files.


seanshpark:
Found this while working on #7032

mhs4670go (approved):
LGTM

PR #7042, opened 2021-06-17
[luci-interpreter] Revise KernelBuilder

seanshpark:
This will revise KernelBuild class not to inherit from Visitor as not
needed anymore and rename visit method to build.


seanshpark:
Tidy from result of #6993

seanshpark:
We may just use a single method
```cxx
std::unique_ptr<Kernel> build_kernel(const luci::CircleNode *node, ..... );
```
but we can revise like this later...

jinevening (approved):
LGTM

PR #7043, opened 2021-06-17
[tools] Remove tflite2circle.sh and fuse_instance_norm.js

glistening:
It removes tflite2circle.sh which converts using 'flatc` internally.
It was once used before tflite2circle in c++ was introduced.
However, it will not be maintained and seems outdated.

Also, it removes fuse_instance_norm.js, it was also replaced with c++
implementation.


mhs4670go (approved):
LGTM

hseok-oh (approved):

PR #7046, opened 2021-06-17
[luci] Common IMPLEMENT macro in logex

seanshpark:
This will commonize IMPLEMENT macro and add undef.


mhs4670go (approved):
LGTM

PR #7047, opened 2021-06-17
[luci] Add undef for CNVISIT_GRP

seanshpark:
This will add undef for CNVISIT_GRP in CircleCloneNode.


mhs4670go (approved):
LGTM

PR #7048, opened 2021-06-17
[luci] Tidy OperationExporter class

seanshpark:
This will tidy OperationExporter class as
- remove inherit from visitor as needless now
- rename visit() to export_node()
- tidy unused visit() codes
- add undef of VISIT_OE macro


mhs4670go (approved):
LGTM

PR #7054, opened 2021-06-21
[one-cmds] Flush commands

seanshpark:
This will add flush write execution to show log immediatly for long
execution commands, such as record-minmax.


mhs4670go (approved):
LGTM

PR #7055, opened 2021-06-21
[record-minmax] Log for every record

seanshpark:
This will revise log for every record in profileData method.


jinevening:
For benchmark models, we typically use a thousand of records. Printing all of them may be unncessary.

How about logging each record only when the number of records is less than 10?

jinevening (approved):
I get your intention after offline talk. Thanks :)

seanshpark:
For the record;
- running in our internal CircleCI makes some time.
- plan to revise logging: #7059
- will tune interval (100 --> 1) in future, maybe 10?

PR #7057, opened 2021-06-21
DRAFT: [luci] Fix FormattedGraph

seanshpark:
on-going draft to fix luci logex FormattedGraph

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #7061, opened 2021-06-22
[luci] Add missing IRs for logex SummaryBuilder

seanshpark:
This will add missing IRs for logex SummaryBuilder.


seanshpark:
From draft #7057

mhs4670go (approved):
LGTM

Issue #7056, opened 2021-06-21
Compiler FE: Test coverage drop with logoex/FormattedGraph.cpp


seanshpark:
Test coverage has dropped with compiler/luci/logex/src/FormattedGraph.cpp file after refactoring.

seanshpark:
Related change: https://github.com/Samsung/ONE/pull/6990/files

seanshpark:
Did not visit `BUILD_GRP(DEF);` and after in `bool NodeSummaryBuilder::build(const loco::Node *node, locop::NodeSummary &s) const`.

Assume eaten by default handler in `CircleNodeSummaryBuilderBase`
```cxx
#define CIRCLE_NODE(OPCODE, CLASS) \
  virtual bool summary(const CLASS *, locop::NodeSummary &) const { return false; }
#define CIRCLE_VNODE CIRCLE_NODE
#include <luci/IR/CircleNodes.lst>
#undef CIRCLE_VNODE
#undef CIRCLE_NODE
```

seanshpark:
There are unimplemented visit for some IRs as there was no problem with it.
Another problem as there is default implementation.

I think it is time to change this like other codes with IR visitor pattern: don't handle in base class.

seanshpark:
These fail with default handler is disabled
```
31/44 Test #31: luci_unit_readtest ...............***Failed   13.42 sec
FAILED
- BatchMatMulV2_000
- BatchMatMulV2_001
- FullyConnected_002
- MatMul_000
- Reshape_003
- TransposeConv_000
- UnidirectionalSequenceLSTM_001
- Net_Dangle_001
- BCQFullyConnected_001
- L2Pool2D
- Equal
```

seanshpark:
Fixed

seanshpark:
Now:  Methods   84% Lines   87%

PR #7062, opened 2021-06-22
[luci] Revise base build method to return false

seanshpark:
This will revise CircleNodeSummaryBuilderBase::build method just to
return false
- to make all IRs to be implemented
- to make current SummaryBuilderLet to work properly


seanshpark:
For #7056, from draft #7057

mhs4670go (approved):
LGTM

PR #7064, opened 2021-06-22
[model_partition] Add scripts for runtime-driven, heuristic-based model partitioning

dr-venkman:
Add scripts for Graph partitioning. The algorithm is a heuristic that mixes topological sorting, followed by a variant of the Fiduccia Mattheyses algorithm (see https://en.wikipedia.org/wiki/Fiduccia%E2%80%93Mattheyses_algorithm).

Note that the data for the algorithm comes from the TFlite model and a chrome trace output from running nnpackage_run.

The partitions approximate a MinMax split of execution times, such that they can be pipelined.

Signed-off-by: venkat.iyer venkat.iyer@samsung.com

### DIFF
--- a/tools/model_partition_tool/Graph.py
+++ b/tools/model_partition_tool/Graph.py
@@ -1,0 +1,457 @@
+#! /usr/bin/python
+import graph_analysis
+import json
+import logging
+import runtime_stats
+import os
+import sys
+import numpy as np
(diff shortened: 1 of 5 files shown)


dr-venkman:
@chunseoklee, PTAL.

chunseoklee (approved):

Issue #7067, opened 2021-06-23
Compiler FE: InstanceNorm from TF1.15.x tflite


seanshpark:
Let's investivate `InstanceNormalization` from `keras_contrib.layers` with TensorFlow 2.0.0 alpha0

Background
- CartoonGan works only with this TF 2.0.0 a0 version
- InstanceNormalization is used
- InstanceNormalization decomposed network of `tflite` model looks somewhat differenent than 2.3.0 or 2.5.0

Purpose
- Support this pattern to our InstNorm fusion

Investigation result
- Use `InstanceNormalization` source from `keras_contrib`
   -  `keras_contrib` will request above TensorFlow 2.2.0, which this is not our targert
- Use TensorFlow 1.15.3
   -  This version will produce somewhat different InstanceNormalization tflite which I am interested in.

seanshpark:
Virtual env setup

```
virtualenv --system-site-packages -p python3 tf20a0venv
source tf20a0venv/bin/activate
```
```
python -m pip install tensorflow==2.0.0-alpha0
python -m pip install git+https://www.github.com/keras-team/keras-contrib.git
```

seanshpark:
Virtual env setup
```
virtualenv --system-site-packages -p python3 tf1153venv
source tf1153venv/bin/activate
```
```
python -m pip install tensorflow==1.15.3
python -m pip install h5py==2.10.0
```

seanshpark:
Test sorce

```py
import sys
import tensorflow as tf

from tensorflow.keras.layers import Layer, InputSpec
from tensorflow.keras import initializers, regularizers, constraints
from tensorflow.keras import backend as K

class InstanceNormalization(Layer):
```
> copy-paste `InstanceNormalization` class from [keras_contrib/instancenormalization.py](https://github.com/keras-team/keras-contrib/blob/master/keras_contrib/layers/normalization/instancenormalization.py)
```py
# Our test model
inputs = tf.keras.Input(shape=(16, 16, 3))
y = tf.keras.layers.ZeroPadding2D(padding=(3, 3))(inputs)
y = InstanceNormalization(axis=-1, epsilon=1e-9)(y)
y = tf.keras.layers.Activation("relu")(y)
keras_model = tf.keras.Model(inputs=inputs, outputs=y)

# save to h5 model
keras_model.save("InstNorm_keras.h5")

# convert to tflite model
converter = tf.lite.TFLiteConverter.from_keras_model_file("InstNorm_keras.h5",
[...]

PR #7071, opened 2021-06-23
DRAFT: Support InstanceNorm from tensorflow.keras

seanshpark:
on-going draft to support InstanceNorm fusion from tensorflow.keras v1.15.3


seanshpark:
For #7067

seanshpark:
done

PR #7074, opened 2021-06-24
[luci] Relocate CHECK_OR_FALSE outside of method

seanshpark:
This will relocate CHECK_OR_FALSE macro outside of method.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7075, opened 2021-06-24
[luci] Extract is_add_input_mul_const for FuseInstNorm

seanshpark:
This will extract is_add_input_mul_const() method FuseInstanceNormPass
to check input of Add is Mul/Const.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7077, opened 2021-06-27
[luci] Extract InstNorm match for version 2

seanshpark:
This will extract InstanceNorm pattern match for version 2 to separate
method.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7078, opened 2021-06-27
[luci] Add version comment for InstNorm pattern

seanshpark:
This will add version comment to InstNorm pattern.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7079, opened 2021-06-27
[luci] Fix FuseInstNorm final pattern comment

seanshpark:
This will fix FuseInstNorm final pattern comment that was copied from
moco project.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7080, opened 2021-06-27
[luci] Fix name for fused InstNorm node

seanshpark:
This will fix name for fused InstNorm node to be unique for multiple
instances.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7082, opened 2021-06-28
[Draft][No merge] ReplaceSubWithAddPass

llFreetimell:
Parent Issue : #7081

This draft is to introduce `ReplaceSubWithAddPass`.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

Issue #7081, opened 2021-06-28
[luci] Introduce ReplaceSubToAddPass


llFreetimell:
## What?

Let's replace `CircleSub` to `CircleAdd`

```
<before>
[Input] ------+
              +-- [Sub] --
[Const] ------+

<after>
[Input] ------+
              +-- [Add] --
[Const_neg] --+
```

## Why?

In some environment, `Sub` is not supported.
To support `Sub`, we can use `Add`!

### Cuation

This pass is experimental and can be temporary solution.
When this pass is determined as non useful pass, it can be deprecated in the future.

llFreetimell:
Done

PR #7084, opened 2021-06-28
[luci] Introduce ReplaceSubWithAddPass

llFreetimell:
Parent Issue : #7081 
Draft : #7082 

This commit introduces `ReplaceSubWithAddPass` in `luci`.

### DIFF
--- a/compiler/luci/pass/src/ReplaceSubWithAddPass.cpp
+++ b/compiler/luci/pass/src/ReplaceSubWithAddPass.cpp
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


compiler/luci/pass/src/ReplaceSubWithAddPass.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/Service/Nodes/CircleConst.h>
+
+namespace
+{
+
+bool replace_sub_with_const_rhs(luci::CircleSub *sub)
+{
+  auto const_rhs = dynamic_cast<luci::CircleConst *>(sub->y());
```
mhs4670go, line 27: Only y position is supported?

llFreetimell: For now, yes :)
Only `CircleSub(x, y) whose y is constant` supported

seanshpark: X - Y = X + (-Y), where Y is Const
but when X is Const,  X - Y != (-X) + Y
Am I right?

mhs4670go: It really is :)

mhs4670go (approved):
LGTM

compiler/luci/pass/src/ReplaceSubWithAddPass.cpp
llFreetimell: Hmm... the equation `X - Y != (-X) + Y` is always correct regardless of "X is const?" so the answer is yes...

seanshpark (approved):
LGTM

PR #7085, opened 2021-06-28
[luci] Extract InstNorm match for version 0,1

seanshpark:
This will extract InstanceNorm pattern match for version 0 and 1 to its
own method.


mhs4670go (approved):
LGTM

seanshpark:
For history, For #7067 , from draft #7071

PR #7088, opened 2021-06-28
[luci] Introduce FuseInstanceNorm class

seanshpark:
This will inteoduce FuseInstanceNorm class with apply() method that
will do actual replace sub-network with on InstanceNorm node.
- This change will apply with Version_0 pattern.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7093, opened 2021-06-28
[luci] Use FuseInstanceNorm directly

seanshpark:
This will revise to use FuseInstanceNorm directly and remove
not used anymore fuse_instance_norm() method.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7094, opened 2021-06-28
[res] Introduce Net_InstanceNorm_004

seanshpark:
This will introduce Net_InstanceNorm_004 recipe that is like the model
from TensorFlow 1.15.3


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7096, opened 2021-06-28
[luci] InstanceNormPattern::match for Version_3

seanshpark:
This will instroduce InstanceNormPattern::match method for Version_3.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7097, opened 2021-06-28
[luci] Extract fuse_instance_norm with Add and hint

seanshpark:
This will revise FuseInstanceNormPass::run() method as extract
fuse_instance_norm() method with target CircleAdd node and
PatternVersion hint so that finding Add node and pattern matching
is separated.
- Version_Unknown is introduced as hint to let fuse_instance_norm method manage the pattern.


seanshpark:
For #7067 , from draft #7071

seanshpark:
@nnfw-bot test nncc-debug

mhs4670go (approved):
LGTM

PR #7100, opened 2021-06-29
[luci] Remove shape check for V3 InstNorm pattern

seanshpark:
This will revise shape check for Version 3 InstanceNormalization pattern
from version 0 where real models don't match.


mhs4670go (approved):
LGTM

seanshpark:
For #7067

Issue #7111, opened 2021-06-29
Resolve Svace/Coverity Defect(21/06/29)


chunseoklee:
Due to PR #7003, several defects are introduced. Let's resolve them. 

svace link : ~~https://analysishub.sec.samsung.net/service/analyses/641305~~ (resolved at 07/05/2021)
coverity link : ~~https://analysishub.sec.samsung.net/service/analyses/641529~~

chunseoklee:
@YiHyunJin PTAL

YiHyunJin:
|  |
-- | -- |
DeviceInfo.h:130 | UNINIT.CTOR.MANY |
Tensor.cc:511 | NO_CAST.INTEGER_OVERFLOW |
StorageTypeUtil.cc:40 | NO_CAST.INTEGER_OVERFLOW |
GpuObject.h:154 | UNINIT.CTOR |
GpuObject.h:151 | UNINIT.CTOR |
ClKernel.h:50 | UNINIT.CTOR |
add | O |

YiHyunJin:
|  |
-- | -- |
DeviceInfo.h:130 | UNINIT.CTOR.MANY |
Tensor.cc:511 | NO_CAST.INTEGER_OVERFLOW |
StorageTypeUtil.cc:40 | NO_CAST.INTEGER_OVERFLOW |
GpuObject.h:154 | UNINIT.CTOR |
GpuObject.h:151 | UNINIT.CTOR |
ClKernel.h:50 | UNINIT.CTOR |
Buffer.h:64 | UNINIT.CTOR |
Tensor.cc:513 |  NO_CAST.INTEGER_OVERFLOW |
Shape.h:391 |  INVARIANT_RESULT.OP_ZERO |
Shape.h:418 | NO_EFFECT |
StorageTypeUtil.cc:40 | SIGNED_TO_BIGGER_UNSIGNED | 
Shape.h:391 | NO_EFFECT |

YiHyunJin:
StorageTypeUtil.cc:40

`const uint64_t flt4_size = 4 * (descriptor.data_type == DataType::FLOAT32 ? 4 : 2);`

so flt4_size is only 8 or 16.

chunseoklee:
New svace result : https://analysishub.sec.samsung.net/service/analyses/655979

@YiHyunJin PTAL 
1 (relevant) remaining defect : [link](https://analysishub.sec.samsung.net/dm/tizen/sb2/main/review#PRJID=501&WGID=464229)

[...]

PR #7113, opened 2021-06-29
Reslove Svace/Covertiy Defect

YiHyunJin:
uint32_t is always greater than or equal to 0. So the '0 <= i' conditional statement is not needed.

### DIFF
--- a/runtime/onert/backend/gpu_cl/open_cl/Arguments.cc
+++ b/runtime/onert/backend/gpu_cl/open_cl/Arguments.cc
@@ -165,6 +165,8 @@ std::string GetImageModifier(AccessType access)
       return "__write_only";
     case AccessType::READ_WRITE:
       return "__read_write";
+    default:
+      throw std::runtime_error("Invalid AccessType");
   }
   return "";
 }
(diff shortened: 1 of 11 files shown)


runtime/onert/backend/gpu_cl/open_cl/Arguments.cc
```diff
@@ -165,6 +165,8 @@ std::string GetImageModifier(AccessType access)
       return "__write_only";
     case AccessType::READ_WRITE:
       return "__read_write";
+    default:
+      return "";
```
seanshpark, line 169: (optional) maybe throw?

runtime/onert/backend/gpu_cl/open_cl/ClMemory.cc
```diff
@@ -34,6 +34,8 @@ cl_mem_flags ToClMemFlags(AccessType access_type)
       return CL_MEM_WRITE_ONLY;
     case AccessType::READ_WRITE:
       return CL_MEM_READ_WRITE;
+    default:
+      return CL_MEM_READ_ONLY; // unreachable
```
seanshpark, line 38: (optional) maybe throw here too;
I'm not sure this is realistic but if someone adds another type in `AccessType` then this may cause a problem if not added.

chunseoklee:
@YiHyunJin You try to resolve all defects of #7111 on this PR ?

YiHyunJin:
@chunseoklee  yes... I think I've done it all.

seanshpark (approved):
LGTM thank you!

hasw7569 (approved):
LGTM

chunseoklee (approved):
LTGM. @YiHyunJin Please note that there will be new defect even after this PR.

PR #7114, opened 2021-06-29
[res] Introduce Net_InstanceNorm_005 recipe

seanshpark:
This will introduce Net_InstanceNorm_005 recipe to testing pattern
Version_4 of fuse InstanceNorm.


seanshpark:
For #7067 , from draft #7071

mhs4670go (approved):
LGTM

PR #7118, opened 2021-06-29
[luci] Tidy FuseInstanceNormPass run without version

seanshpark:
This will tidy FuseInstanceNormPass::run method without version as it
can be moved inside fuse_instance_norm method.


mhs4670go (approved):
LGTM

PR #7121, opened 2021-06-30
[luci] Revise fuse InstNorm pass Version_0 to 1

seanshpark:
This will revise Version_0 to Version_1 to make numberings smooth.


mhs4670go (approved):
LGTM

Issue #7070, opened 2021-06-23
Compiler FE: Installation failed behind firewall with one-prepare-venv


seanshpark:
While running `one-prepare-venv` inside corporate behind firewall, certificate may have problems

This message is from installing debian package;

> Setting up one-compiler (1.16.1) ...
> ...
> WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852)'),)': /packages/cc/f6/b2996d65fb72200b1455a776d38915760b097ad1fa8c1c177a8cccbff07a/tensorflow_cpu-2.3.0-cp36-cp36m-manylinux2010_x86_64.whl
ERROR: Could not install packages due to an EnvironmentError: HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Max retries exceeded with url: /packages/cc/f6/b2996d65fb72200b1455a776d38915760b097ad1fa8c1c177a8cccbff07a/tensorflow_cpu-2.3.0-cp36-cp36m-manylinux2010_x86_64.whl (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852)'),))

seanshpark:
Something like this made it pass with `one-prepare-venv`
```
PIP_TRUSTED_HOST="--trusted-host pypi.org "
PIP_TRUSTED_HOST+="--trusted-host files.pythonhost.org "
PIP_TRUSTED_HOST+="--trusted-host download.pytorch.org "
PIP_TRUSTED_HOST+="--cert SomePrivateCetificate.crt "
```

seanshpark:
What is needed;
- `"--cert SomePrivateCetificate.crt "` be passed to `apt install triv2-toolchain triv2-toolchain-test`

seanshpark:
Passing options
- pass by argument: hum....
- use environment variable: seems possible

seanshpark:
patch landed.

wait for client has checked this...

seanshpark:
Variable : `ONE_PREPVENV_PIP_OPTION`

seanshpark:
resolved

PR #7122, opened 2021-06-30
[one-cmds] Enable additional pip install options

seanshpark:
This will enable to provide additional pip install options.


seanshpark:
For #7070

mhs4670go (approved):
LGTM

Issue #7125, opened 2021-07-01
Compiler FE: Check ONNX 1.9.0


seanshpark:
ONNX latest version is 1.9.0 (since April) and our `one-prepare-venv` is installing with 1.8.0

Let's check with latest version and if no problem, upgrade to it.

seanshpark:
https://github.com/onnx/onnx/releases/tag/v1.9.0

seanshpark:
ONNX-TF now upgraded to 1.9.0

PR #7126, opened 2021-07-02
[one-cmds] Remove comment about old version of onnx-tf

seanshpark:
This will remove comment about old version of onnx-tf issue.


seanshpark:
Related #7125

mhs4670go (approved):
LGTM

Issue #7130, opened 2021-07-02
Compiler FE: FuseInstNorm having one element for beta, gamma


seanshpark:
Our InstanceNorm model from TF 1.15.3 has beta and gamma constant as having only one element.

This is OK for normal calculation, as it would be broadcasted, but it would be nice for our backend
that currently requires element be number of channel, not a single value.

seanshpark:
Related #7067

seanshpark:
Two cases are needed
- where gamma and beta constant exist but has only one element
- where gamma and beta doesn't exist (assumed to be removed for gamma = 1.0, beta = 0.0)

seanshpark:
done

PR #7131, opened 2021-07-02
[luci] Fix InstNorm beta, gamma const shape

seanshpark:
This will fix fuse InstNorm beta, gamma with 0.0, 1.0 be vector.


seanshpark:
Second case of https://github.com/Samsung/ONE/issues/7130#issuecomment-872749143

seanshpark:
While working on the first case, it would be better to introduce "PostFusion" step for this to covert both cases.

seanshpark:
For #7130

PR #7133, opened 2021-07-02
DRAFT Fix FuseInstanceNorm beta, gamma shape

seanshpark:
on-going draft to fix InstanceNorm beta and gamma shape.


seanshpark:
For #7130

seanshpark:
all landed

PR #7134, opened 2021-07-02
[luci] Introduce FuseInstNorm post_fusion

seanshpark:
This will introduce post_fusion method in FuseInstanceNormPass to
provide post processing after InstanceNormalization fusion.


seanshpark:
For #7130, from draft #7133

mhs4670go (approved):

PR #7136, opened 2021-07-02
[res] Introduce Net_InstanceNorm_006

seanshpark:
This will introduce Net_InstanceNorm_006 recipe that has scala gamma and beta.

### DIFF
--- a/res/TensorFlowLiteRecipes/Net_InstanceNorm_006/test.recipe
+++ b/res/TensorFlowLiteRecipes/Net_InstanceNorm_006/test.recipe
@@ -1,0 +1,283 @@
+# InstanceNorm network with one element for gamma, beta
+
+operand {
+  name: "input_1"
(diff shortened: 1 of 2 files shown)


seanshpark:
For #7130, from draft #7133

res/TensorFlowLiteRecipes/Net_InstanceNorm_006/test.recipe
```diff
@@ -0,0 +1,283 @@
+# same as Net_InstanceNorm_004 but having one element for gamma, beta
```
mhs4670go, line 1: (optinal) I think this file better be described without any other files' explanation becoz it would become its dependency. If 004 is changed, this file should be changed either. But seems rarely happen:)

mhs4670go (approved):

mhs4670go (approved):
LGTM

Issue #7124, opened 2021-07-01
[onert] Mixed test fail


hseok-oh:
`aarch64` debug build, mixed test log

```
20:09:35  [ RUN      ] GenModelTest.OneOp_Add_VarToConst
20:09:35  terminate called after throwing an instance of 'std::logic_error'
20:09:35    what():  basic_string::_M_construct null not valid
20:09:35  /jenkins/workspace/nnfw/master/push-cross-aarch64-runtime-test-debug/Product/out/test/command/unittest: line 80:  2389 Aborted                 (core dumped) $TEST_BIN $(get_gtest_option)
20:09:35  Product/out/unittest_standalone/nnfw_api_gtest failed... return code: 134
20:09:35  ============================================
20:09:35  Finishing set 4: Product/out/unittest_standalone/nnfw_api_gtest...
20:09:35  ============================================
```

Same on `armv7l` debug/release.

Fail after #7119

chunseoklee:
on local test, found that reverting https://github.com/Samsung/ONE/commit/e93bcb128a564a7b7224d7756e2bca9ba7bf94ad fixed failure.

@YiHyunJin PTAL

chunseoklee:
@intom Here is a log for fail on nnfw_api_gtest:

```
~/d/ONE ⬢  8.10.0 29.276s
➜  ssh xu4 ONERT_LOG_ENABLE=0 /home/odroid/Product/out/unittest_standalone/nnfw_api_gtest --gtest_filter="*OneOp_Add_VarToConst"
bash: warning: setlocale: LC_ALL: cannot change locale (ko_KR.UTF-8)
Note: Google Test filter = *OneOp_Add_VarToConst
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from GenModelTest
[ RUN      ] GenModelTest.OneOp_Add_VarToConst
/home/twoflower/dev/ONE/tests/nnfw_api/src/GenModelTest.h:319: Failure
      Expected: (nnfw_prepare(_so.session))
      Which is: 1
To be equal to: NNFW_STATUS_NO_ERROR
      Which is: 0
[  FAILED  ] GenModelTest.OneOp_Add_VarToConst (177 ms)
[----------] 1 test from GenModelTest (177 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (177 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[...]

PR #7137, opened 2021-07-04
[onert] Modify storage type

YiHyunJin:
Modified to use less memory.
Currently, when reading and writing images, it is reading and writing in multiples of 16 bytes.
There is still a memory issue, not a 16byte multiple in cpu memory.


YiHyunJin:
related : https://github.com/Samsung/ONE/issues/7124#issuecomment-872900136

PR #7139, opened 2021-07-05
[onert] Fix precision to F32

YiHyunJin:
F16 is not implemented yet, so it is fixed as F32.


YiHyunJin:
related : #7124

PR #7140, opened 2021-07-05
[luci] Add match beta, gamma channel for FuseInstanceNorm

seanshpark:
This will enable match beta, gamma channel for FuseInstanceNorm as post
process.


seanshpark:
For #7130, from draft #7133

mhs4670go (approved):
LGTM

Issue #7142, opened 2021-07-05
[luci] Fix multiple '; Mem = ' messages in log


seanshpark:
Recent changes in log `SummaryBuilderLet` class shows redudant message
```
; Mem = 0x55c16641e900
; Mem = 0x55c16641e900
; Mem = 0x55c16641e900
...
```

Let's fix not to show multiple `; Mem` messages

seanshpark:
done

PR #7143, opened 2021-07-05
[luci] Fix to print Mem comment only once

seanshpark:
This will revise to log for Mem value only if summary method of the CLASS is handled.
This is to prevent multiple Mem logs where dynamic_cast itself is
executed by multiple times as ther are multiple 'SummaryBuilderLet' classes.


seanshpark:
For #7142

mhs4670go (approved):
LGTM

Issue #7147, opened 2021-07-05
[luci] Introduce FoldNegPass


llFreetimell:
## What?

Let's introduce constant folding pass for `CircleNeg`

### Todo

- [ ] Introduce `FoldNegPass` #7148
- [ ] Add `FoldNegPass` in `circle2circle`

llFreetimell:
Let's reopen this when it is needed

PR #7148, opened 2021-07-05
[luci] Introduce FoldNegPass

llFreetimell:
Parent Issue : #7147

This commit introduces `FoldNegPass` in `luci`.

### DIFF
--- a/compiler/luci/pass/src/FoldNegPass.cpp
+++ b/compiler/luci/pass/src/FoldNegPass.cpp
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FoldNegPass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+bool fold_neg(luci::CircleNeg *neg)
+{
+  // Check neg has const input
+  auto const_x = dynamic_cast<luci::CircleConst *>(neg->x());
+  if (not const_x)
+    return false;
+
+  // If quantparam exists, fold by scale
+  if (auto qparam = const_x->quantparam())
+  {
+    for (uint32_t i = 0; i < qparam->scale.size(); ++i)
+      qparam->scale.at(i) *= -1.0;
+  }
+  else
+  {
+    switch (const_x->dtype())
+    {
+      case loco::DataType::FLOAT32:
+        for (uint32_t i = 0; i < const_x->size<loco::DataType::FLOAT32>(); ++i)
+          const_x->at<loco::DataType::FLOAT32>(i) *= -1.0;
+        break;
+      default:
+        return false;
+    }
+  }
+
+  loco::replace(neg).with(const_x);
+  return true;
+}
+
+} // namespace
+
+namespace luci
+{
+
+/**
+ * Constant Folding for Neg Op
+ **/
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/src/FoldNegPass.cpp
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+#include "luci/Pass/FoldNegPass.h"
+
+#include <luci/IR/CircleNodes.h>
+
+namespace
+{
+
+bool fold_neg(luci::CircleNeg *neg)
```
seanshpark, line 24: Could you please add a diagram explaining what this pass does?

compiler/luci/pass/src/FoldNegPass.test.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    _neg->x(_x);
+
+    _neg->name("neg");
+    _x->name("x");
+  }
+
+  loco::Node *createFoldedPattern() override { return _neg; }
+
+  void set_qparam()
```
seanshpark, line 53: Why is this needed?

compiler/luci/pass/src/FoldNegPass.cpp
```diff
@@ -0,0 +1,77 @@
...
+      case loco::DataType::FLOAT32:
+        for (uint32_t i = 0; i < const_x->size<loco::DataType::FLOAT32>(); ++i)
+          const_x->at<loco::DataType::FLOAT32>(i) *= -1.0;
+        break;
+      default:
+        return false;
+    }
+  }
+
+  loco::replace(neg).with(const_x);
```
seanshpark, line 50: if `const_x` is used as input to another node, this will damage the graph.
usuallly we leave the `const_x` node, create a clone of it and set the values and use it.

compiler/luci/pass/src/FoldNegPass.cpp
```diff
@@ -0,0 +1,77 @@
...
+  // Check neg has const input
+  auto const_x = dynamic_cast<luci::CircleConst *>(neg->x());
+  if (not const_x)
+    return false;
+
+  // If quantparam exists, fold by scale
+  if (auto qparam = const_x->quantparam())
+  {
+    for (uint32_t i = 0; i < qparam->scale.size(); ++i)
+      qparam->scale.at(i) *= -1.0;
```
seanshpark, line 35: there are some models that have `quantparam` by tflite converter, but set with meaningless values.
for that case, does it make correct folding?

seanshpark:
@llFreetimell , please prepare a draft PR and do this
(1) add a recipe with the network and add/run `circle2circle-dredd-recipe-test`
(2) and this is optional, add/run `luci-pass-value-test`: I'm not not sure of `quantparam` change with _luci-interpreter_

compiler/luci/pass/src/FoldNegPass.cpp
llFreetimell: Oh, I missed this, thanks :)

compiler/luci/pass/src/FoldNegPass.cpp
llFreetimell: If `CircleConst` has `quantparam`, following equation can be calculated.

```
f = (q - zeropoint) * scale
-f = (q - zeropoint) * (-scale)
```

So I think it is not related with the meaningless values.

llFreetimell:
> please prepare a draft PR

I thought that #7147 is really simple and this is almost all of to-do lists but I will do it.


> (1) add a recipe with the network and add/run circle2circle-dredd-recipe-test
> (2) and this is optional, but highly recommended, add/run luci-pass-value-test

I am not object to these but just curious, why most of other passes did not have this test until now...?

compiler/luci/pass/src/FoldNegPass.cpp
jinevening, line 35: Negative scale seems risky to me. In the normally quantized model, zp will be changed after Neg is applied.

Do you have to fold Neg for quantized model? If not, IMHO it would be better to support only FP32 in this pass.

seanshpark: @jinevening pointed out what I was worried in mind. 
This code doesn't check the dtype so would also run with Float32. 
What I wanted to point out is that float models also have `quantparam`, is this what you have in mind too?

seanshpark:
> I am not object to these but just curious, why most of other passes did not have this test until now...?

I think they are simple or easly predictable or just lazy not to make it.
What made me ask of the test is about the `quantparam` usage which is not sure for me.

llFreetimell:
Let's reopen this when it is needed

Issue #7149, opened 2021-07-06
Compiler FE: CSV to vector library


seanshpark:
Let's introduce a library that provides CSV to vector functions.

seanshpark:
While working on #6723 for some recfacrting, we need to extract `csv_to_vector()` to some library as this is used `luci/pass` too.

seanshpark:
With a little discussion with @parjong , `pepper-csv2vec` or `csv2vec` can be a good candidate.

seanshpark:
I'll work this inside #6231 as to see how it works.

seanshpark:
TODO: wait for our internal analysis server reports...

seanshpark:
![image](https://user-images.githubusercontent.com/4616940/124832216-47b53800-dfb7-11eb-9e06-81024cf67382.png)

seanshpark:
done

PR #7154, opened 2021-07-06
[pepper-csv2vec] Introduce csv2vec

seanshpark:
This will introduce pepper module csv2vec that will return vector from
CSV format text and some related helper methods.


seanshpark:
For #7149, from #6231
- This is to gather similar methods in luci and some command tools.

mhs4670go (approved):
LGTM

Issue #7152, opened 2021-07-06
Compiler FE: Remove redundant error message


mhs4670go:
## What

Let's remove `ERROR` keyword from `std::runtime_error`.

## Why

Some codes that use `std::runtime_error` have message including `ERROR` string like below.
```cxx
throw std::runtime_error("ERROR: Failed to verify circle '" + input_model_path + "'");
```

But, `runtime_error` function itself already let us know that it is kind of error function. So, `ERROR` string can be removed.

Plus, `safemain` module in our compiler already add a string `ERROR` to error message.
```cxx
int main(int argc, char **argv)
{
  try
  {
    return entry(argc, argv);
  }
  catch (const std::exception &e)
  {
    std::cerr << "ERROR: " << e.what() << std::endl;
  }

  return 255;
}
```

PR #7156, opened 2021-07-06
[onnxkit] Remove redundant "ERROR" string

mhs4670go:
This commit removes redundant ERROR string from error message.

Related: #7152


seanshpark (approved):
LGTM

Issue #6723, opened 2021-04-30
Compiler FE: partition model with specific nodes


seanshpark:
Let's provide partitioning model into several models with specific nodes.
- specific nodes can be providing a list nodes name

First, we need to develop some ideas in meaningful way of partitioning.

seanshpark:
For #6230

seanshpark:
Todo...
- [ ] extend PartitionTable by operator name
- [ ] ...

seanshpark:
seems done

PR #7159, opened 2021-07-06
[res] Introduce Part_Add_Sub_001

seanshpark:
This will introduce Part_Add_Sub_001 that has long and special
characters in node name.


seanshpark:
For #6723, from draft #6231

mhs4670go (approved):
LGTM

PR #7158, opened 2021-07-06
DRAFT: Support raw data in record-minmax

jinevening:
On going draft to support raw data in record-minmax.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/7132

Issue #7132, opened 2021-07-02
[record-minmax] Accept raw data as a representative dataset


jinevening:
## What

Let's enable record-minmax to accept raw data as a representative dataset.

## Why

For someone who is not familiar with hdf5.

CC @lemmaa

jinevening:
`record-minmax --input_data <input_data>`

If <input_data> is an .h5 file (~~.h5), record-minmax will run as before.

Else, <input_data> is assumed to be a text file which specifies the representative data. The text file should contain absolute file path per line. The pointed file should be a binary file containing one representative input, ready to be consumed by the input circle model without any modification (i.e., pre-processed raw data).

jinevening:
Future work may be to support some pre-precessing functionalities.

seanshpark:
> Else, <input_data> is assumed to be a text file which specifies the representative data.

This makes codes complicated. I don't think this is a good approach. It would be better to introduce new option.

jinevening:
I'm afraid adding a new option may confuse users. For example, if there are two options (e.g., --input_data (we should keep this option for backward compatibility), --raw_input_list), users can misuse the option (using with a wrong file type or using both options, etc).

seanshpark:
> using with a wrong file type 

All the options have to solve this problem. Even you go with one option: h5 or not.

> using both options

We also have this cases. I think this is natural. We have this situation on _one-import_tf_ where input model be one of saved model, keras or just graph_def.

1) to support by option
- `--input_data` or  `--raw_input_list` or ...
- `--input_data` for input file and addition option `--data-type`  (default is `h5` and there are others, `list` for text format list)
- we only check for the type, if failed it is a failure.  codes and testing will be simple.

2) automatic type detection
- we must first try with existing supported type. how? using extension can be dangerous.
- we may try with error handling. h5 throws. this is ok. for text of list.. well...
- when we try to support other types... it happens to be a text with a list but format is different...
- code itself should handle all the cases with `if, elseif, elseif,....`

anyway, there can be lots of bad cases that we can think of and lots of good things we only wany to think of..
there can be many complicated and error prone for also (1) but here we want to make it simple for both use interface and code.

jinevening:
@seanshpark  I understand your concern.

Then, there are two options.
1. --input_data, --raw_input_list, ..
2.  --input_data, --input_data_format (input_data_format can be `h5`/`hdf5` (default) and `list`)

I'd like to go with the second option because it is simpler.

[...]

PR #7160, opened 2021-07-06
[record-minmax] Support calibration using raw data

jinevening:
This supports record-minmax to accept raw data.

---
Related to: #7132
Draft PR: #7158

### DIFF
--- a/compiler/record-minmax/src/RecordMinMax.cpp
+++ b/compiler/record-minmax/src/RecordMinMax.cpp
@@ -38,12 +38,14 @@ using DataType = luci_interpreter::DataType;
 namespace
 {
 
-void readDataFromFile(const std::string &filename, char *data, size_t data_size)
+void readDataFromFile(const std::string &filename, std::vector<char> &data, size_t data_size)
 {
+  assert(data.size() == data_size); // FIX_CALLER_UNLESS
+
   std::ifstream fs(filename, std::ifstream::binary);
   if (fs.fail())
     throw std::runtime_error("Cannot open file \"" + filename + "\".\n");
-  if (fs.read(data, data_size).fail())
+  if (fs.read(data.data(), data_size).fail())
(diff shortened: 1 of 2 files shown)


compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -152,6 +161,50 @@ void RecordMinMax::initialize(const std::string &input_model_path)
...
+
+    uint32_t offset = 0;
+    for (auto input : input_nodes)
+    {
+      const auto *input_node = loco::must_cast<const luci::CircleInput *>(input);
+      std::vector<char> input_data(getTensorSize(input_node));
+      readDataFromFile(record, input_data.data() + offset, input_data.size());
+      _interpreter->writeInputTensor(input_node, input_data.data() + offset, input_data.size());
+
+      offset += input_data.size();
```
seanshpark, line 192: This seems somewhat strange... offset accumulates and then read to `input_data.data() + offset`...
If there are 10 inputs, offset may go out of bounds.

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -38,6 +38,15 @@ using DataType = luci_interpreter::DataType;
 namespace
 {
 
+void readDataFromFile(const std::string &filename, char *data, size_t data_size)
+{
```
seanshpark, line 42: it would be nice to have some validation check
- is data valid?
- is data_size smaller than the file size?

compiler/record-minmax/src/RecordMinMax.cpp
jinevening: Ah, that was wrong implementation.. Thanks for pointing it out. I've updated the logic.

jinevening:
@seanshpark I've updated the comment. PTAL

seanshpark (approved):
LGTM thank you!

PR #7162, opened 2021-07-06
[luci] Revise PartitionTable

seanshpark:
This will revise PartitionTable to have type for comply and add byopnames;


seanshpark:
For #6723, from draft #6231

mhs4670go (approved):
LGTM

Issue #7164, opened 2021-07-06
[circle-partitioner] Relocate methods


seanshpark:
Some local methods canbe reused if they are relocated.
Let's move them from _circle-partitioner_ to appropriate places.

seanshpark:
While working on #6723 ...

seanshpark:
done

seanshpark:
done

PR #7165, opened 2021-07-06
[luci] Add PartitionTable dump

seanshpark:
This will add PartitionTable class dump into luci::partition.


seanshpark:
For #7164 , from draft #6231

mhs4670go (approved):
LGTM

PR #7166, opened 2021-07-06
[circle-partitioner] Separate section reading

seanshpark:
This will revise parse_table method to read main '[partition]' section
first then read other sections.
- to prevent default group settings when orders are inverted


seanshpark:
For #6723, from draft #6231

mhs4670go (approved):
LGTM

I think `crew::Sections` can be map instead of vector.

PR #7170, opened 2021-07-06
[record-minmax] Add input_data_format option to record-minmax

jinevening:
This adds input_data_format option to record-minmax.

---
Related to: #7132
Draft PR: #7158

### DIFF
--- a/compiler/record-minmax/driver/Driver.cpp
+++ b/compiler/record-minmax/driver/Driver.cpp
@@ -76,6 +76,11 @@ int entry(const int argc, char **argv)
     .type(arser::DataType::STR)
     .help("Record mode. percentile (default) or moving_average");
 
+  arser.add_argument("--input_data_format")
+    .nargs(1)
+    .type(arser::DataType::STR)
+    .help("Input data format. h5/hdf5 (default) or list/filelist");
+
   arser.add_argument("--generate_profile_data")
     .nargs(0)
     .required(false)
@@ -102,6 +107,7 @@ int entry(const int argc, char **argv)
   std::string mode("percentile");
   float min_percentile = 1.0;
   float max_percentile = 99.0;
+  std::string input_data_format("h5");
 
   if (arser["--min_percentile"])
     min_percentile = arser.get<float>("--min_percentile");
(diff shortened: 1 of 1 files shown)


compiler/record-minmax/driver/Driver.cpp
```diff
@@ -127,8 +136,24 @@ int entry(const int argc, char **argv)
...
     auto input_data_path = arser.get<std::string>("--input_data");
 
-    // Profile min/max while executing the given input data
-    rmm.profileData(mode, input_data_path, min_percentile, max_percentile);
+    if (input_data_format == "h5" || input_data_format == "hdf5")
+    {
+      // Profile min/max while executing the H5 data
+      rmm.profileData(mode, input_data_path, min_percentile, max_percentile);
+    }
+    // input_data is a text file where the absolute path of each record (a representative data) is
```
seanshpark, line 144: > absolute path

Actual code just reads the file so I thin relative to CWD would also work.
If your intention is to use absolute path, them we need to check this.
If not, how about changing this to

> file where the file path of each record (a representative data) is

seanshpark: > each record

Is this each line of the text file? using word `record` is hard to understand the unit or something...

compiler/record-minmax/driver/Driver.cpp
```diff
@@ -127,8 +136,24 @@ int entry(const int argc, char **argv)
...
+    // written line-by-line. Each record should contain all inputs of a model, concatenated in the
```
seanshpark, line 145: `record` here gives better understanding but still, it's a bit hard to imagine

```
input data is a text file having file path in each line.
each data file is composed of multiple sets of input data of the model,
concatenated  in same order with the input index of the model

[input 1 data][input 2 data]...[input n data] | [input 1 data][input 2 data]...[input n data] ...
|<-----------   set 1    ------------------>|   |<-----------   set 2    ------------------>|
```

jinevening: According to the picture, a data file should contain a single set (composed of all inputs of a model).

I've added a picture for easier understading. PTAL.

seanshpark (approved):
LGTM thank you!

PR #7173, opened 2021-07-06
[circle-partitioner] Read comply and enable group by name

seanshpark:
This will revise to read section by comply and enable to set group by opname.

### DIFF
--- a/compiler/circle-partitioner/src/PartitionRead.cpp
+++ b/compiler/circle-partitioner/src/PartitionRead.cpp
@@ -30,14 +30,22 @@ using namespace partee;
 
 const char *_section_partition = "partition";
 const char *_section_OPCODE = "OPCODE";
+const char *_section_OPNAME = "OPNAME";
+
+const char *_comply_opcode = "opcode";
+const char *_comply_opname = "opname";
 
 const char *_key_backends = "backends";
 const char *_key_default = "default";
+const char *_key_comply = "comply";
 const char *_key_underscore = "_";
 
 luci::PartitionTable parse_table(const crew::Sections &sections)
 {
   luci::PartitionTable table;
+
+  // default comply as OPCODE
+  table.comply = luci::PartitionTable::COMPLY::OPCODE;
 
   // read main '[partition]' first
   for (auto &section : sections)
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6723, from draft #6231

compiler/circle-partitioner/src/PartitionRead.cpp
```diff
@@ -69,13 +92,33 @@ luci::PartitionTable parse_table(const crew::Sections &sections)
       for (auto &item : items)
       {
         if (item.first == _key_underscore)
-          table.default_group = item.second;
+        {
+          if (table.comply == luci::PartitionTable::COMPLY::OPCODE)
+            table.default_group = item.second;
+        }
```
seanshpark, line 98: This change is to prevent setting default group from wrong section (not the section of comply)

compiler/circle-partitioner/src/PartitionRead.cpp
```diff
@@ -69,13 +92,33 @@ luci::PartitionTable parse_table(const crew::Sections &sections)
       for (auto &item : items)
       {
         if (item.first == _key_underscore)
```
mhs4670go, line 94: When the item.first have just "_" string?

seanshpark: Maybe DRAFT code https://github.com/Samsung/ONE/pull/6231/files#diff-8d11f27aebeac12e3f55cb8df554d505279af923c78791b7b3132415590f7625 will help?

compiler/circle-partitioner/src/PartitionRead.cpp
```diff
@@ -69,13 +92,33 @@ luci::PartitionTable parse_table(const crew::Sections &sections)
...
+        {
+          if (table.comply == luci::PartitionTable::COMPLY::OPNAME)
+            table.default_group = item.second;
+        }
+        else
+        {
+          table.byopnames.emplace(item.first, item.second);
+        }
+      }
+    }
```
mhs4670go, line 121: (optional) Just for future revision, this is what I think it could be considered.
1. As I said in another PR, if `Sections` was map, it would be easy to implement these kind of logic.
2. I think there is only one section for each comply. So, like `find_section` function would help. Then, duplicate codes can be removed.
```cxx
..
// read other sections
section = find_section(sections, table.comply) // or, sections[table.comply] if map.
auto &items = section.items;
..
```
Or, just inserting `break` at each end of `if` condition in `for` loop would help.

mhs4670go (approved):
LGTM

compiler/circle-partitioner/src/PartitionRead.cpp
seanshpark: Hope to see the draft if you have some time to prepare :)

PR #7176, opened 2021-07-07
[luci] Revise produce_pgroups with comply

seanshpark:
This will revise produce_pgroups to set group by comply type given:
OPCODE or OPNAME.


seanshpark:
For #6723, from draft #6231

mhs4670go (approved):
LGTM

PR #7177, opened 2021-07-07
[luci] Introduce Module validate method

seanshpark:
This will introduce validate method for Module.


seanshpark:
Fpr #7164, from draft #6231
- this is copied from compiler/circle-partitioner/src/CirclePartitioner.cpp and replace in near future

mhs4670go (approved):
LGTM

PR #7178, opened 2021-07-07
[circle-partition] Enable value test with comply OPNAME

seanshpark:
This will enable basic value test for comply OPNAME with simple Add-Sub model.


seanshpark:
For #6723, from draft #6231

mhs4670go (approved):
LGTM

PR #7180, opened 2021-07-07
[circle-partitioner] Update readme about using '_'

seanshpark:
This will update readme about using '_' to override default group.


seanshpark:
From draft #6231

mhs4670go (approved):
LGTM

PR #7187, opened 2021-07-08
Remove memory_management and related files

chunseoklee:
- remove memory management related files
- remove InferenceContext.cc, which is not used on test code.


chunseoklee:
@YiHyunJin PTAL

YiHyunJin (approved):

PR #7188, opened 2021-07-08
[onert] Resolve svace for UNINIT and RESOURCE_LEAK 

YiHyunJin:
Initialize uninitialized variables.
Implemented UnloadOpenCL.

### DIFF
--- a/runtime/onert/backend/gpu_cl/open_cl/OpenclWrapper.cc
+++ b/runtime/onert/backend/gpu_cl/open_cl/OpenclWrapper.cc
@@ -63,7 +63,7 @@ void LoadOpenCLFunctions(HMODULE libopencl);
 void LoadOpenCLFunctions(void *libopencl, bool use_wrapper);
 #endif
 
-absl::Status LoadOpenCL()
+absl::Status LoadOpenCL(void *libopencl)
 {
 #ifdef __WINDOWS__
   HMODULE libopencl = LoadLibraryA("OpenCL.dll");
@@ -79,7 +79,7 @@ absl::Status LoadOpenCL()
       absl::StrCat("Can not open OpenCL library on this device, error code - ", error_code));
   }
 #else
-  void *libopencl = dlopen("libOpenCL.so", RTLD_NOW | RTLD_LOCAL);
+  libopencl = dlopen("libOpenCL.so", RTLD_NOW | RTLD_LOCAL);
   if (libopencl)
   {
     LoadOpenCLFunctions(libopencl, false);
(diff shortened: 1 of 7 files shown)


YiHyunJin:
Related : https://github.com/Samsung/ONE/issues/7111

runtime/onert/backend/gpu_cl/open_cl/OpenclWrapper.cc
```diff
@@ -108,23 +108,34 @@ absl::Status LoadOpenCL()
 #endif
 }
 
+void UnloadOpenCL(void *libopencl)
+{
+  if (libopencl)
+  {
+    dlclose(libopencl);
```
seanshpark, line 115: have you tested this code?

chunseoklee (dismissed):

chunseoklee:
wrong approve by mistake

runtime/onert/backend/gpu_cl/open_cl/OpenclWrapper.cc
YiHyunJin: TC worked... So i uploaded a commit.
But when I check, it doesn't come in.
So.. i will check more.. Thanks!!

runtime/onert/backend/gpu_cl/open_cl/OpenclWrapper.cc
```diff
@@ -79,7 +79,7 @@ absl::Status LoadOpenCL()
       absl::StrCat("Can not open OpenCL library on this device, error code - ", error_code));
   }
 #else
-  void *libopencl = dlopen("libOpenCL.so", RTLD_NOW | RTLD_LOCAL);
+  libopencl = dlopen("libOpenCL.so", RTLD_NOW | RTLD_LOCAL);
```
seanshpark, line 82:
```cxx
absl::Status LoadOpenCL(void **libopencl)
...
*libopencl = dlopen("libOpenCL.so", RTLD_NOW | RTLD_LOCAL);
```
maybe try like this

YiHyunJin:
```
FORTIFY: pthread_mutex_lock called on a destroyed mutex (0x703a1762a8)
Aborted
```

I debugging it now...

runtime/onert/backend/gpu_cl/open_cl/OpenclWrapper.cc
```diff
@@ -108,23 +108,33 @@ absl::Status LoadOpenCL()
...
+}
+
 #ifdef __WINDOWS__
 void LoadOpenCLFunctions(HMODULE libopencl)
 {
 #else
-void LoadOpenCLFunctions(void *libopencl, bool use_wrapper)
-{
 #ifdef __ANDROID__
+void LoadOpenCLFunctions(void **libopencl, bool use_wrapper)
```
seanshpark, line 124: maybe leave ANDROID code  as-is...

YiHyunJin: I modified this because "use_wrapper" is not used when not on Android.

runtime/onert/backend/gpu_cl/open_cl/OpenclWrapper.cc
```diff
@@ -108,23 +108,33 @@ absl::Status LoadOpenCL()
 #endif
 }
 
+void UnloadOpenCL(void **libopencl)
+{
+  if (*libopencl)
+  {
+    dlclose(*libopencl);
```
seanshpark, line 115: (I think I lost what I added here...anyway...)
as `_handle` is not updated, we can use `void UnloadOpenCL(void *libopencl)`
and this makes here without two `**` and caller without `&` which looks better to me.

seanshpark (approved):
LGTM thank you!

chunseoklee (approved):

Issue #7192, opened 2021-07-09
[circle-quantizer] Use same quantparam for input/output of Split


jinevening:
## What

Let's make input/output of Split have the same quantparam

PR #7193, opened 2021-07-09
[circle-quantizer] Use the same qparam for input/output of Split

jinevening:
This sets the same qparam for input/output of Split.

---
Related to: https://github.com/Samsung/ONE/issues/7192


seanshpark (approved):
LGTM

PR #7208, opened 2021-07-12
[circle-quantizer] Quantize fused Tanh using pre-determined values.

jinevening:
This quantizes fused Tanh with special values.


seanshpark (approved):
LGTM

seanshpark:
This PR itself looks OK but the whole codes inside is getting more unorganized(?) to me.
- visitor and if (...) is mixed
- vistor with <true> but return is not used
- ...

It would be better to refactor somehow as I think you will be continue changing this file...

jinevening:
> the whole codes inside is getting more unorganized(?) to me.

I agree. The code is getting complex as new (unexpected) cases are added.. I'll think about refactoring.

Issue #7209, opened 2021-07-12
[circle-partitioner] Assign CircleConst to user node group


seanshpark:
Let's assign group of `CircleConst` to follow the user(s) node.

If there are multiple user nodes, let's make a clone for each group.

seanshpark:
This replaces #7185 issue.

seanshpark:
Test recipe to prepare
- Net_InstanceNorm_003.003.part for main
- Const node used by two nodes: cpu / acl_cl
- Const node used by three ndoes: cpu / acl_cl / cpu
- Const node used by three ndoes: cpu / acl_cl / acl_cl

seanshpark:
seems working

PR #7212, opened 2021-07-13
[res] Introduce Part_Add_Sub_002 recipe

seanshpark:
This will introduce Part_Add_Sub_002 recipe that will have shared
constant input to Add and Sub nodes.


seanshpark:
For #7209, from draft #6231

mhs4670go (approved):
LGTM

PR #7214, opened 2021-07-13
[circle-quantizer] Verify optional bias in FC layer.

jinevening:
This verifies optional bias in FC layer.

### DIFF
--- a/compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
+++ b/compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
@@ -180,6 +180,7 @@ private:
     RETURN_FALSE_UNLESS(is_lwq(node->input()))
     RETURN_FALSE_UNLESS(is_cwq_const(node->weights(), 0))
     luci::CircleConst *bias = dynamic_cast<luci::CircleConst *>(node->bias());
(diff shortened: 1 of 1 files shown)


jinevening:
@seanshpark  I've seen this pattern when converting GRU onnx model to quantized circle.

seanshpark:
> I've seen this pattern when converting GRU onnx model to quantized circle.

You mean, there is no bias or bias is not Constant?
Maybe add some comment about this change in the source...

@llFreetimell , can you please share your experience with such kind?

llFreetimell:
> there is no bias or bias is not Constant?
> can you please share your experience with such kind?

I am not sure whether bias can be non-constant or not.
However, since bias is optional, no bias is possible.
This is from long time ago and it is already applied in our many source codes :)
Following is one of them.

https://github.com/Samsung/ONE/blob/1f38057354541d2467da7eae6f28b5b3aa0fa788/compiler/luci/import/src/Nodes/CircleFullyConnected.cpp#L40

jinevening:
> You mean, there is no bias or bias is not Constant?

What I've seen was bias with CircleOutputExclude type.

jinevening:
@nnfw-bot test nncc-debug

seanshpark (approved):
LGTM thank you!

Issue #7186, opened 2021-07-08
Compiler FE: circle-partitioner validation methods


seanshpark:
To validate _circle-partitioner_ is working as expected

Step 1) count output circle files and compare with expected
Step 2) TBD --> X

Todo
- [x] provide a validation method that can be used in cmake test process
- [ ] enable the test in `circle-partitioner-test` --> X
- [x] enable the test in `circle-part-value-test`

seanshpark:
> count output circle files and compare with expected

`.conn.ini` file has `[models]` section that has total models
or in `conn.json` file with `parts` key value array size is total models

seanshpark:
done

PR #7217, opened 2021-07-14
[circle-part-value-test] Save expected count

seanshpark:
This will revise cmake to save expected paritioned models count of partitioning.


seanshpark:
For #7186, from draft #6231

mhs4670go (approved):
LGTM

PR #7220, opened 2021-07-14
[circle-partitioner] Test check number of output models

seanshpark:
This will revise circle-part-value-test to check also number of output
partitioned models generated with expected number.

### DIFF
--- a/compiler/circle-part-value-test/part_eval_one.py
+++ b/compiler/circle-part-value-test/part_eval_one.py
@@ -4,6 +4,7 @@ import tensorflow as tf
 import subprocess
 import argparse
 import traceback
+import json
 
 #
 # This script compares the execution result of TFLite interpreter and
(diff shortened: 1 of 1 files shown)


seanshpark:
For #7186, from draft #6231

compiler/circle-part-value-test/part_eval_one.py
```diff
@@ -23,6 +24,32 @@
...
 circle_model = args.name + ".circle"
 partition_conn_ini = args.name + ".conn.ini"
+partition_conn_json = args.name + ".conn.json"
+expected_count = args.name + ".excnt"
+
+# Check expected count of models from partitioning
+try:
+    with open(expected_count, "r") as expected_count_file:
+        expected_count_line = expected_count_file.readline()
+        expected_count_file.close()
```
mhs4670go, line 34:
```suggestion
```
This line can be removed because `with` statement does the same thing.

compiler/circle-part-value-test/part_eval_one.py
```diff
@@ -23,6 +24,32 @@
...
+        with open(partition_conn_json) as json_file:
+            json_data = json.load(json_file)
+            parts_value = json_data["parts"]
+            if len(parts_value) != expected_count_line:
+                print("Partitioned model count differs from expected:",
+                      expected_count_line)
+                quit(255)
+
+            print("Partitioned model count expected: ", expected_count_line)
+            json_file.close()
```
mhs4670go, line 47: ditto
```suggestion
```

compiler/circle-part-value-test/part_eval_one.py
seanshpark: Thanks!

compiler/circle-part-value-test/part_eval_one.py
```diff
@@ -23,6 +24,30 @@
...
+
+    expected_count_line = int(expected_count_line)
+    if expected_count_line:
+        with open(partition_conn_json) as json_file:
+            json_data = json.load(json_file)
+            parts_value = json_data["parts"]
+            if len(parts_value) != expected_count_line:
+                print("Partitioned model count differs from expected:",
+                      expected_count_line)
+                quit(255)
```
mhs4670go, line 43: How about using internal Exception class?
```suggestion
                raise ValueError("Partitioned model count differs from expected:",
                      expected_count_line)
```
Or `AssertionError` could be used.

seanshpark: Ah, just want to exit with 255 code. This may work but it is tested so I want to go as-is.

mhs4670go (approved):
LGTM

PR #7221, opened 2021-07-14
[circle-partitioner] More tests for OPNAME

seanshpark:
This will add more tests for circle-partitioner for comply=opname.


seanshpark:
For #6723, from draft #6231

mhs4670go (approved):
LGTM

Issue #6708, opened 2021-04-27
Compiler FE: Partition enable with IF Op


seanshpark:
Lets' enable partitioning with IF and WHILE Op.

These two has linked sub-graph to clone; which can be a bit more complicated than other ops.

seanshpark:
For #6230

seanshpark:
Start with `IF`
- from a simplest one `IF` model, let's see how the partitioned result graph should look like.

seanshpark:
Refresh for `IF` node
- `THEN` and `ELSE` branch in file is index of subgraph --> order of subgraph is important
- there is `post_import_graph()` step in `import` that will make a link FROM index TO  graph pointer

seanshpark:
In the steps of partitioning, preparation for `IF` is now ready in draft.
Last big thing left is to attach subgraphs to `IF` node: in `produce_pmodules()` method.

seanshpark:
Test of `circle-partition`
```
cd circle-partitioner-test

LUCI_LOG=1 \
../circle-partitioner/circle_partitioner \
Net_InstanceNorm_003.part \
Net_InstanceNorm_003.circle \
Net_InstanceNorm_003
```

Test of `circle-part-value-test`
```
cd circle-part-value-test/Net_If_Add_Sub_000.001

LUCI_LOG=1 \
../../circle-part-driver/circle_part_driver \
Net_If_Add_Sub_000.001.conn.ini \
3 \
Net_If_Add_Sub_000.001.circle.input \
Net_If_Add_Sub_000.001.circle.output
```

seanshpark:
Testing another IF graph inside IF-THEN graph...

_luci-interpreter_ has this condition.
```cxx
graph->getInputTensors().size() == getInputTensors().size() - 1
```

seanshpark:
> graph->getInputTensors().size() == getInputTensors().size() - 1
> 
> This condition doesn't seem to be correct.

@jinevening , do you know why this condition is needed? or if you don't know, can you give your opinion on this condition?

seanshpark:
`circle-part-value-test` with `Part_If_Add_Sub_001` graph + `Part_If_Add_Sub_001.001.part`
--> _luci_interpreter_ needs more updates

> RuntimeError: tensorflow/lite/kernels/if.cc:77 num_inputs != subgraph->inputs().size() (2 != 3)Node number 2 (IF) failed to prepare.

jinevening:
[...]

PR #7223, opened 2021-07-15
DRAFT: partition IF

seanshpark:
on-going draft to partition IF node.


seanshpark:
For #6708

seanshpark:
all necessary files landed

Issue #7224, opened 2021-07-15
[common-artifacts] Check enable If_000


seanshpark:
Let's check if we can enable `If_000` for `tcgenerate(If_000)`

seanshpark:
From #7222

seanshpark:
> Generate build/debug/compiler/common-artifacts/If_000.opt/metadata/tc/input.h5 and build/debug/compiler/common-artifacts/If_000.opt/metadata/tc/expected.h5
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Invalid data size.
Aborted (core dumped)

seanshpark:
In compiler/common-artifacts/src/TestDataGenerator.cpp

```diff
@@ -166,7 +178,8 @@ int entry(int argc, char **argv)

   std::random_device rd; // used to obtain a seed for the random number engine
   uint32_t input_index = 0;
-  for (uint32_t g = 0; g < circle_model->subgraphs()->size(); g++)
+  // for (uint32_t g = 0; g < circle_model->subgraphs()->size(); g++)
+  uint32_t g = 0;
   {
     const auto input_nodes = loco::input_no
```

seems to make `If_000` and `IF_001` pass...

mhs4670go:
@seanshpark IIUC, since subgraphs except first one don't receive inputs from user, above change seems okay:)

seanshpark:
`If` test is now enabled.

PR #7225, opened 2021-07-15
[common-artifacts] Support bool for random data

seanshpark:
This will enable to support bool type for random data.

### DIFF
--- a/compiler/common-artifacts/src/TestDataGenerator.cpp
+++ b/compiler/common-artifacts/src/TestDataGenerator.cpp
@@ -65,6 +65,15 @@ template <typename T> void geneate_random_data(std::mt19937 &gen, void *data, ui
   }
 }
 
+template <> void geneate_random_data<bool>(std::mt19937 &gen, void *data, uint32_t size)
+{
+  std::normal_distribution<float> distrib(0, 2); // mean(0), stddev(2)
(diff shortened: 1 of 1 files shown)


seanshpark:
For #7224, from draft #7223

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -65,6 +65,15 @@ template <typename T> void geneate_random_data(std::mt19937 &gen, void *data, ui
...
 
+template <> void geneate_random_data<bool>(std::mt19937 &gen, void *data, uint32_t size)
+{
+  std::normal_distribution<float> distrib(0, 2); // mean(0), stddev(2)
+  for (uint32_t i = 0; i < size; i++)
+  {
+    static_cast<bool *>(data)[i] = distrib(gen) >= 0 ? true : false;
+  }
+}
+
```
mhs4670go, line 76: I'm not sure but is it necessary? I think template covers this bool case.

seanshpark: as the random value rang would be -2 ~ 2, mostly would be not 0, that is mostly would be true...
logically it is OK but I wanted to give equal possibility...

mhs4670go (approved):
LGTM

PR #7226, opened 2021-07-15
[common-artifacts] Data generation only for main graph

seanshpark:
This will revise data generation only for main graph as other subgraphs
are invoked from main graph.


seanshpark:
For #7224, from draft #7223

mhs4670go (approved):
LGTM

PR #7228, opened 2021-07-15
[common-artifacts] Enable IF tcgenerate

seanshpark:
This will enable tfgenerate for IF models.


seanshpark:
For #7224

mhs4670go (approved):
LGTM

Issue #7229, opened 2021-07-15
[luci] MaxPoolWithArgmax resolving pass


Bronnikoff:
### What? 

Let's introduce resolving pass to lower MaxPoolWithArgmax custom op to subgraph with currently supported Circle node's.

### Why?

Circle schema and luci doesn't has `MaxPoolWithArgmax`. So we can't quantize and execute models with this custom op now.

Bronnikoff:
Models with MaxPoolWithArgmax can be obtained from TF models with tf.nn.max_pool_with_argmax operation using conversion `tf --> tflite --> circle`.

Bronnikoff:
DRAFT: #7236

binarman:
This issue depends on #7306

binarman:
Everything is merged, closing

Issue #6912, opened 2021-06-01
[luci] Support CAST quantization


seanshpark:
We need to support for `CAST` Op for quantization
- our in-house model has this Op.
- cast from int64 input to float32.

First, we need to decide how quantization of `CAST` should work.

seanshpark:
Related: #6911

seanshpark:
resolved

PR #7236, opened 2021-07-16
[DRAFT] Introduce MaxPoolWithArgmax resolving pass

Bronnikoff:
This commit adds resolving pass for MaxPoolWithArgmax op.

-----------------

For: #7229

### DIFF
--- a/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
@@ -231,7 +231,7 @@ void fill_coords_addition(luci::Padding padding, const luci::Stride &stride,
     for (int32_t x_o = 0, x_i = start_y; x_o < output_width; ++x_o, x_i += step_x)
     {
       auto const output_idx = y_o * output_width + x_o;
-      auto const input_idx = y_i * input_width + x_i;
+      auto const input_idx = y_i * static_cast<int32_t>(input_width) + x_i;
 
       cords->at<DT>(output_idx) = input_idx;
     }
--- a/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
@@ -257,19 +257,353 @@ luci::CircleConst *create_coords_addition(loco::Graph *graph, luci::Padding padd
   return cords;
 }
 
+luci::CircleNode *get_custom_output(const luci::CircleCustom *cop, int32_t idx)
+{
+  auto const outputs = loco::succs(cop);
+  assert(outputs.size() == 2);
+
+  auto output = loco::must_cast<luci::CircleCustomOut *>(*outputs.begin());
+  if (output->index() != idx)
+  {
+    output = loco::must_cast<luci::CircleCustomOut *>(*outputs.rbegin());
+  }
+
+  return output;
+}
+
+luci::CircleNode *max_pool_branch(luci::Padding padding, const luci::Stride &stride,
+                                  const luci::Filter filter, luci::CircleCustom *cop)
+{
+  auto graph = cop->graph();
+  auto input = cop->inputs(0);
+
+  auto origin = luci::get_origin(cop);
+  auto name = cop->name() + "/Argmax";
+
+  // Create MaxPool
+  auto maxpool = none_act_func(graph->nodes()->create<luci::CircleMaxPool2D>());
+  {
+    init_name_and_origin(maxpool, name + "/MaxPool2D", origin);
+
+    set_stride(maxpool, stride);
+    set_filter(maxpool, filter);
+    maxpool->padding(padding);
+
+    maxpool->value(input);
+  }
+
+  return maxpool;
+}
+
+luci::CircleNode *window_flattened_coord(const std::string &name, luci::Padding padding,
+                                         const luci::Stride &stride, const luci::Filter filter,
+                                         uint32_t output_height, uint32_t output_width,
+                                         luci::CircleNode *input)
+{
+  auto const graph = input->graph();
+  auto const origin = luci::get_origin(input);
+
+  auto const depth_dimension = 3;
+
+  // Create Conv2D
+  auto conv = none_act_func(graph->nodes()->create<luci::CircleConv2D>());
+  {
+    init_name_and_origin(conv, name + "/Conv2D", origin);
+
+    // Padding, Stride and kernel size equal to MaxPool's
+    set_stride(conv, stride);
+    conv->padding(padding);
+
+    // depth of kernel is equal to square size
+    auto const kh = filter.h();
+    auto const kw = filter.w();
+    auto const kd = kh * kw;
+
+    // use zero bias
+    auto bias = create_zero_bias(graph, kd);
+    init_name_and_origin(bias, conv->name() + "/Bias", origin);
+
+    // create filter
+    // TODO make shared
+    auto weights = create_conv_filter(conv, kh, kw, kd);
+    init_name_and_origin(weights, conv->name() + "/Weights", origin);
+
+    conv->bias(bias);
+    conv->filter(weights);
+    conv->input(input);
+  }
+
+  // Create ArgMax
+  auto argmax = graph->nodes()->create<luci::CircleArgMax>();
+  {
+    init_name_and_origin(argmax, name + "/ArgMax", origin);
+
+    argmax->output_type(loco::DataType::S32);
+
+    // Create argmax_dim
+    auto argmax_dim = create_scalar<loco::DataType::S32>(graph, depth_dimension);
+    init_name_and_origin(argmax_dim, argmax->name() + "/Dimension", origin);
+
+    argmax->dimension(argmax_dim);
+    argmax->input(conv);
+  }
+
+  // Create Reshape to 4-rank back, because argmax decrease rank of tensor by 1
+  auto reshape = graph->nodes()->create<luci::CircleReshape>();
+  {
+    init_name_and_origin(reshape, name + "/Reshape", origin);
+
+    auto shape = create_shape_tensor(graph, {1, output_height, output_width, 1});
+    init_name_and_origin(shape, reshape->name() + "/Shape", origin);
+
+    reshape->tensor(argmax);
+    reshape->shape(shape);
+  }
+
+  // Create Cast to use float32 instead int32
+  auto argmax_cast = create_cast(reshape, loco::DataType::S32, loco::DataType::FLOAT32);
+  init_name_and_origin(argmax_cast, argmax->name() + "/Cast", origin);
+
+  return argmax_cast;
+}
+
+luci::CircleNode *window_y_coord(const std::string &name, float filter_width,
+                                 luci::CircleNode *flattened)
+{
+  auto const graph = flattened->graph();
+  auto const origin = luci::get_origin(flattened);
+
+  auto div = none_act_func(graph->nodes()->create<luci::CircleMul>());
+  {
+    init_name_and_origin(div, name + "/Div", origin);
+
+    auto divider = create_scalar<loco::DataType::FLOAT32>(graph, 1.0f / filter_width);
+    init_name_and_origin(divider, div->name() + "/Divider", origin);
+
+    div->x(flattened);
+    div->y(divider);
+  }
+
+  auto floor = graph->nodes()->create<luci::CircleFloor>();
+  {
+    init_name_and_origin(floor, name + "/Floor", origin);
+    floor->x(div);
+  }
+
+  return floor;
+}
+
+luci::CircleNode *window_x_coord(const std::string &name, float filter_width,
+                                 luci::CircleNode *flattened, luci::CircleNode *y_coord)
+{
+  auto const graph = flattened->graph();
+  auto const origin = luci::get_origin(flattened);
+
+  auto mod = none_act_func(graph->nodes()->create<luci::CircleAdd>());
+  {
+    init_name_and_origin(mod, name + "/Mod", origin);
+
+    auto neg = graph->nodes()->create<luci::CircleNeg>();
+    {
+      init_name_and_origin(neg, mod->name() + "/Neg", origin);
+
+      auto mul = none_act_func(graph->nodes()->create<luci::CircleMul>());
+      {
+        init_name_and_origin(mul, neg->name() + "/Neg", origin);
+
+        auto multipler = create_scalar<loco::DataType::FLOAT32>(graph, filter_width);
+        init_name_and_origin(multipler, mul->name() + "/Multipler", origin);
+
+        mul->x(y_coord);
+        mul->y(multipler);
+      }
+
+      neg->x(mul);
+    }
+
+    mod->x(flattened);
+    mod->y(neg);
+  }
+
+  return mod;
+}
+
+luci::CircleNode *plane_flattened_coord(const std::string &name, uint32_t input_width,
+                                        luci::CircleNode *y_coord, luci::CircleNode *x_coord,
+                                        luci::CircleNode *corners)
+{
+  auto const graph = corners->graph();
+  auto const origin = luci::get_origin(corners);
+
+  auto add = none_act_func(graph->nodes()->create<luci::CircleAdd>());
+  {
+    init_name_and_origin(add, name + "/Add", origin);
+
+    auto addition = none_act_func(graph->nodes()->create<luci::CircleAdd>());
+    {
+      init_name_and_origin(addition, add->name() + "/Add", origin);
+
+      auto y_addition = none_act_func(graph->nodes()->create<luci::CircleMul>());
+      {
+        init_name_and_origin(y_addition, addition->name() + "/Mul", origin);
+
+        auto width_scalar = create_scalar<loco::DataType::FLOAT32>(graph, input_width);
+        init_name_and_origin(width_scalar, y_addition->name() + "/Const", origin);
+
+        y_addition->x(y_coord);
+        y_addition->y(width_scalar);
+      }
+
+      addition->x(x_coord);
+      addition->y(y_addition);
+    }
+
+    add->x(addition);
+    add->y(corners);
+  }
+
+  return add;
+}
+
+luci::CircleNode *volume_flattened_coords(const std::string &name, uint32_t channel,
+                                          uint32_t input_depth, luci::CircleNode *plane)
+{
+  auto const graph = plane->graph();
+  auto const origin = luci::get_origin(plane);
+
+  // Create Mul
+  auto mul = none_act_func(graph->nodes()->create<luci::CircleMul>());
+  {
+    init_name_and_origin(mul, name + "/Mul", origin);
+
+    auto depth_scalar = create_scalar<loco::DataType::FLOAT32>(graph, input_depth);
+    init_name_and_origin(depth_scalar, mul->name() + "/Const", origin);
+
+    mul->x(plane);
+    mul->y(depth_scalar);
+  }
+
+  luci::CircleNode *volume = mul;
+
+  // Add channel number to output
+  if (channel > 0)
+  {
+    // Create Add
+    auto add_ch = none_act_func(graph->nodes()->create<luci::CircleAdd>());
+    init_name_and_origin(add_ch, name + "/Add_Channel", origin);
+
+    auto channel_scalar = create_scalar<loco::DataType::FLOAT32>(graph, channel);
+    init_name_and_origin(channel_scalar, add_ch->name() + "/Const", origin);
+
+    add_ch->x(mul);
+    add_ch->y(channel_scalar);
+
+    volume = add_ch;
+  }
+
+  return volume;
+}
+
+luci::CircleNode *argmax_branch(luci::Padding padding, const luci::Stride &stride,
+                                const luci::Filter filter, luci::CircleCustom *cop)
+{
+  auto graph = cop->graph();
+  auto input = loco::must_cast<luci::CircleNode *>(cop->inputs(0));
+  auto output = get_custom_output(cop, 1);
+
+  auto const depth_dimension = 3;
+  auto const input_depth = input->dim(depth_dimension).value();
+  auto const input_height = input->dim(1).value();
+  auto const input_width = input->dim(2).value();
+
+  assert(output->rank() == 4);
+  auto const output_height = output->dim(1).value();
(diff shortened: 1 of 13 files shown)


seanshpark:
Please add `circle2circle-dredd-recipe-test`

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,670 @@
...
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto cop = dynamic_cast<luci::CircleCustom *>(node);
+    if (not cop)
+      continue;
+
+    if (cop->custom_code() != "MaxPoolWithArgmax")
+      continue;
+
+    changed |= resolve_max_pool_with_argmax(cop);
```
seanshpark, line 664: Please revise this like others. (Please do not use `|=` )

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,670 @@
...
+
+  // replace old node with new subgraph
+  loco::replace(cop).with(output_cast);
+
+  return true;
+}
+
+} // namespace
+
+namespace luci
```
seanshpark, line 649: Please add a graph diagram explaining before - after.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,670 @@
...
+
+  return cords;
+}
+
+bool resolve_max_pool_with_argmax(luci::CircleCustom *cop)
+{
+#define CHECK_OR_FALSE(condition) \
+  if (not(condition))             \
+    return false;
+#define CHECK_OR_THROW(condition, message) \
```
seanshpark, line 220: Please do not use throw if condition is not met. Just reutrn false.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,670 @@
...
+  return luci::Padding::UNDEFINED;
+}
+
+template <loco::DataType DT> void fill_conv_weights(luci::CircleConst *weights)
+{
+  assert(weights->rank() == 4);
+
+  auto const kn = weights->dim(0).value();
+  auto const kw = weights->dim(1).value();
+  auto const kh = weights->dim(2).value();
```
seanshpark, line 74: If this is NHWC and N is 0, I assume `kw` is for width, then is kw accessing dim(1) correct?

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.test.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "luci/Pass/ResolveCustomOpMaxPoolWithArgmaxPass.h"
+
+#include <gtest/gtest.h>
+
+TEST(ResolveCustomOpMaxPoolWithArgmaxPassTest, name)
+{
+  luci::ResolveCustomOpMaxPoolWithArgmaxPass pass;
+  auto const name = pass.name();
+  ASSERT_NE(nullptr, name);
+}
```
seanshpark, line 26: I think we need some value evaluation test before and after transformation.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,774 @@
...
+      break;
+    default:
+      // TODO support other data types
+      INTERNAL_EXN("Datatype is not supported, yet!");
+  }
+
+  return cords;
+}
+
+bool resolve_max_pool_with_argmax(luci::CircleCustom *cop)
```
seanshpark, line 290: this method is very long.. it would be better to split to several functions.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,774 @@
...
+      // TODO support other data types
+      INTERNAL_EXN("Datatype is not supported, yet!");
+  }
+
+  return weights;
+}
+
+template <loco::DataType DT> void fill_zero_bias(luci::CircleConst *bias)
+{
+  assert(bias->rank() == 1);
```
seanshpark, line 159: how about adding an empty line after `assert()` like above function

Bronnikoff:
## Pass description

Let me please explain introduced by this draft pass! We have operation `MaxPoolWithArgmax` with 2 outputs: output №0: `output` and output №1: `argmax`. 

image

image


### Output №0

Output №0 is MaxPool2D of input (from [documentation](https://www.tensorflow.org/api_docs/python/tf/nn/max_pool_with_argmax)). Let's create `CircleMaxPool2D` instead of this branch:

https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L362-L382
https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L742

### Output №1

This is a much harder part of this pass.  Output №1 (`argmax`) is `S64` tensor of corresponding to output №0 flattened indexes of max values from input tensor. So, we need to decompose this branch using exist circle nodes.

From [documentation](https://www.tensorflow.org/api_docs/python/tf/nn/max_pool_with_argmax):
image

Let's define this indexes by following steps:
1. For each output of maxpool define corresponding indexes (x, y) in sliding window: y in [0, `kh`], x in [0, `kw`], where `kh` - kernel height, `kw` - kernel width.
2. Transform window indexes to plane indexes using position (x', y') of left top corner of  window in input tensor:
`plane_idx = ((y' * width) + x') + y*width + x`
Example (white cross - position of max value for window, black point - position of top left corner):
image
3. Transform plane coords to volume coords by `volume_idx = plane_idx * depth + ch`, where `ch` - number of channel.

#### Zero step

Let's split input by depth and search indexes of max values for each plane separately. 
https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L385

When idxs is found, we can glue result back:
https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L407

#### First step

To find idx of max value in window, we can transform 2D window to 1D using Conv2D with count of kernels: `kw * kh`. 

https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L407
https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L479

For example, for 2x2 kernel convolution filter will have following values:

```
|1 0| |0 1| |0 0| |0 0|
|0 0| |0 0| |1 0| |0 1|
```
After this we can apply `CirceArgmax` by depth dimension to output of Conv2D to obtain flattened window idxs.  

https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L491

So, `y = floor(wind_idx / kw)`, `x = wind_idx - kw * y`. This operations simulated by subgraph:
image

#### Second step:

To compute indexes `((y' * width) + x')` of left top corners introduced function `create_coords_addition()`, which calls `fill_coords_addition()`.
 
https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L257
https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L217

`fill_coords_addition()` just iterate over input tensor as pooling operation and writes coord of corner to constant tensor.

`plane_idx = ((y' * width) + x') + y*width + x` implemented as Mul with Add operations.


#### Third step

`volume_idx = plane_idx * depth + ch` equation also computes as Mul with Add operations.

Bronnikoff:
This resolving pass is not elegant =( 
It would be nice to hear suggestions about this pass, if somebody could help!

Bronnikoff:
Negative side of this implementation is `CircleCast`'s! AFAIK *luci-interpreter* not supports this operator now, but it's required to satisfy datatype of output =(

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
Bronnikoff: I will try to make this function shorter =)

Bronnikoff:
> Negative side of this implementation is `CircleCast`'s! AFAIK _luci-interpreter_ not supports this operator now, but it's required to satisfy datatype of output =(

Related: #7261

seanshpark:
> Negative side of this implementation is CircleCast

I've seen you are working also to suppoart `CircleCast` in _luci-interpreter_.
And I assume you are working on this Op to support our NPU backend.
Then our existing issue is to support quantization for `Cast` Op. Do you have any plans for this?

seanshpark:
> This resolving pass is not elegant =(

I've also no idea and don't fully understand current code but...
What I think is, our first code may not not be that elegant but we can always make things better when we have good idea.

seanshpark:
@meejeong , can you please look in to this transformation and give any comments with the viewpoint of correctness?

FYI @Bronnikoff , we usually discuss with or ask for guides from @meejeong for transformations and quantization...

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
binarman: @Bronnikoff
You can make functions for operations and leave only high level structure.
Something like this:
``` c++
luci::CircleMaxPool2D *max_pool = make_maxpool(input);
std::vector<luci::CircleSplitOut> channel_split = make_channel_split(input);
for (int i = 0 ; i<input_depth; ++i)
{
    luci::CircleConv2D *conv = make_conv(channel_split[i]);
    etc...
}
luci::CircleConcat channels_concat = make_concat(argmax_results);
```

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,774 @@
...
+    split->dtype(input->dtype());
+    copy_shape(input, split);
+    split->num_split(int32_t(input_depth));
+
+    split->split_dim(split_dim);
+    split->input(input);
+  }
+
+  // Create Concatenation
+  auto concat = graph->nodes()->create<luci::CircleConcatenation>(input_depth);
```
binarman, line 407: Maybe move this to the bottom of the function?

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,774 @@
...
+      // shape of split out equal to shape of split except split_dim, which equal to 1
+      copy_shape(split, split_out);
+      split_out->dim(depth_dimension).set(1);
+
+      split_out->input(split);
+    }
+
+    // Define idx of max element in Window:
+    // Create Conv2D
+    auto conv = graph->nodes()->create<luci::CircleConv2D>();
```
binarman, line 447: This conv implements a special "space_to_depth" operation with arbitrary block width and height, you can use this analogy to describe this convolution purpose.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
Bronnikoff: I created this node here in order to set `Concatenation` inputs (outputs of the branches) at the end of loop body:

https://github.com/Samsung/ONE/blob/ff94f1985fd76d0e4ca731d593644b3c2249c96d/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp#L722

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,668 @@
...
+
+template <typename NodeT> NodeT *none_act_func(NodeT *node)
+{
+  node->fusedActivationFunction(luci::FusedActFunc::NONE);
+  return node;
+}
+
+template <loco::DataType DT> luci::CircleCast *create_cast(luci::CircleNode *input)
+{
+  auto cast = input->graph()->nodes()->create<luci::CircleCast>();
```
llFreetimell, line 99: If we use `CircleCast`, can we handle this at internal backend @.@?

seanshpark: Good question. And `circle-quantizer` too...

Bronnikoff:
> Then our existing issue is to support quantization for `Cast` Op (#6912). Do you have any plans for this?

Unfortunately, I can propose only one strategy to quantize `Cast` for integer output dtypes: record min-max values during execution and fail quantization, if `max - min > 255` (for Q8), otherwise, quantize this value with `scale = 1.0f`. 

This approach has some negative sides:
- We have no guarantee that min-max values during execution will satisfy condition: `max - min > 255`. Cost of error for integer datatypes, which usually represents indexes in tensors, is higher than for floats.
- In case of using `Cast` for `MaxPoolWithArgmax` resolving we will get failing of quantization in cases, when input will contain more than 255 elements. 

But, I have no idea how to represent `MaxPoolWithArgmax` with another way, yet =(

seanshpark:
@jinevening , PTAL about `Cast` quantization...

jinevening:
> @jinevening , PTAL about Cast quantization...

I don't have a plan to support Cast as of now. I'm not sure if it is even possible. In addition, ArgMax is not quantized as well.

Bronnikoff:
MaxPoolWithArgmax is not supported by `tflite`, so we need to verify this pass using `tf`. For this I made following scripts:

- *invoker.py* which creates simple model with `tf.nn.max_pool_with_argmax`, executes this model with random data, store input/output of model to file and saves model to tflite.

<details> 
<summary> invoker.py </summary>

```python
import tensorflow as tf
import numpy as np

with tf.Graph().as_default() as graph:
    # create model
    input_shape = (1, 4, 4, 32)
    ksize = [1, 2, 2, 1]
    stride = [1, 1, 1, 1]
    input = tf.compat.v1.placeholder(tf.float32, shape = input_shape)
    output, argmax = tf.nn.max_pool_with_argmax(input, ksize, stride, 'VALID')

    # save to tflite
    with tf.compat.v1.Session() as sess:
        sess.run(tf.compat.v1.global_variables_initializer())
        converter = tf.compat.v1.lite.TFLiteConverter.from_session(sess, [input], [output, argmax])
        converter.allow_custom_ops = True
        content = converter.convert()
        with open("custom_model.tflite", 'wb') as binf:
            binf.write(content)

    # invoke and save input ans output output
    with tf.compat.v1.Session() as sess:
        in0 = np.random.rand(*input_shape).astype(np.float32)
        out0, out1 = sess.run([output, argmax], {input: in0})
        #save output in npy format
        np.save("output0.npy", out0)
        np.save("output1.npy", out1)
        # save input in both formats
        np.save("input0.npy", in0)
        with open("input.0", 'wb') as binf:
            binf.write(in0.tobytes('C'))
```

</details>

- *check_invoker.sh*, which calls *invoker.py*, make transformation of dumped tflite model to circle model with resolved node and calls *check_invoker.py* to verify outputs of model.

<details> 
<summary> check_invoker.sh </summary>

```bash
#!/bin/bash

python3 invoker.py &> /dev/null 
./build/compiler/tflite2circle/tflite2circle custom_model.tflite custom_model.circle
./build/compiler/circle2circle/circle2circle --resolve_customop_max_pool_with_argmax custom_model.circle opt.circle
./build/compiler/luci-eval-driver/luci_eval_driver opt.circle 1 input. output.
python3 check_invoker.py
```

</details> 

- *check_invoker.py* makes verification of `tf` and `luci-interpreter` outputs.

<details> 
<summary> check_invoker.py </summary>

```python
import numpy as np

tfl_out0 = np.load('output0.npy')
tfl_out1 = np.load('output1.npy')

with open('output.0', 'rb') as binf:
    cir_out0 = np.frombuffer(binf.read(), dtype=tfl_out0.dtype).reshape(tfl_out0.shape)

with open('output.1', 'rb') as binf:
    cir_out1 = np.frombuffer(binf.read(), dtype=tfl_out1.dtype).reshape(tfl_out1.shape)

print("output0 equality is:", np.array_equal(tfl_out0, cir_out0))
print("output1 equality is:", np.array_equal(tfl_out1, cir_out1))
print("MAE of output0:", np.abs((tfl_out0 - cir_out0)).mean())
print("MAE of output1:", np.abs((tfl_out1 - cir_out1)).mean())
```

</details> 


TODO: make `circle2circle-dredd-recipe-test` and test provide test with other configurations (paddings, strides, input sizes).

> Please add `circle2circle-dredd-recipe-test`

res/TensorFlowLiteRecipes/MaxPoolWithArgmax_002/test.recipe
```diff
@@ -1,20 +1,20 @@
...
   name: "ofm"
   type: FLOAT32
-  shape { dim: 1 dim: 8 dim: 8 dim: 1 }
+  shape { dim: 1 dim: 8 dim: 8 dim: 2 }
 }
 operand {
   name: "argmax"
   type: INT64
-  shape { dim: 1 dim: 8 dim: 8 dim: 1 }
+  shape { dim: 1 dim: 8 dim: 8 dim: 2 }
```
binarman, line 14: I changed number of channels to diversify this test.
Otherwise all three tests had one common set of check rules.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.test.cpp
binarman: @Bronnikoff mentioned [here](https://github.com/Samsung/ONE/pull/7236#issuecomment-885763769) tflite can not run this operation, so we can not add dredd value test for this operator.

Do we have Tensorflow (not tflite) based tests in this repo?

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
binarman: @llFreetimeII
> If we use CircleCast, can we handle this at internal backend @.@?

We can do it! (but with some restrictions =/ )

Currently I am working on it.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,746 @@
...
+/**
+ * BEFORE
+ *                 |
+ *            [CircleNode]
+ *                 |
+ *     [CUSTOM(MaxPoolWithArgmax]
+ *         |              |
+ *  [MaxPool output]  [Argmax output]
+ *
+ * AFTER
```
llFreetimell, line 687: Wow..... O.O

llFreetimell:
@binarman 

I was trying to understand the overall logic but other stuffs make me busy T.T
I will still try to review this draft, but if time is not enough to wait my review, you can proceed without my review and then I will ask later if there is some question :)

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
binarman: Yeah, this is the problem with this PR, it is hard to review such complex computations =_=

binarman:
@seanshpark 
I unfortunately, I do not know how to make this PR easier to review.
But I've run several tests using @Bronnikoff scripts and fixed some bugs, I suppose now this code works as intended.

Configurations I've tried:
```
input_shape = (1, 3, 3, 1)
ksize = [1, 3, 3, 1]
stride = [1, 1, 1, 1]

input_shape = (1, 3, 3, 1)
ksize = [1, 3, 3, 1]
stride = [1, 2, 2, 1]

input_shape = (1, 10, 10, 1)
ksize = [1, 2, 3, 1]
stride = [1, 3, 2, 1]

input_shape = (1, 64, 64, 32)
ksize = [1, 5, 5, 1]
stride = [1, 2, 2, 1]
```
every configuration with both SAME and VALID paddings.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,814 @@
...
+    int32_t left_pad = full_w_pad / 2;
+    int32_t right_pad = full_w_pad - left_pad;
+    int32_t top_pad = full_h_pad / 2;
+    int32_t bottom_pad = full_h_pad - top_pad;
+    auto padding_const = create_padding_const(graph, left_pad, right_pad, top_pad, bottom_pad);
+    init_name_and_origin(padding_const, name + "/Pad_shape", origin);
+    pad->paddings(padding_const);
+
+    auto padding_value =
+      create_scalar<loco::DataType::FLOAT32, float>(graph, std::numeric_limits<float>::lowest());
```
binarman, line 370: @jinevening 
I want to use lowest floating point value to "simulate" max pool with padding bahavior.
These values could be harmful for quantization, how do you feel about filtering such values in record-minmax?

binarman:
Everything is merged, closing.

@seanshpark @llFreetimell Thank you!

Issue #6302, opened 2021-03-19
[luci-interpreter] Integrate build configurator of luci interpreter for cortex-m4 microcontrollers


SlavikMIPT:
In continuation of https://github.com/Samsung/ONE/issues/5080

## Goal
Integrate build configurator of luci interpreter for cortex-m4 microconrollers into the mainstream.

## Rationale
During investigation I found that luci interpreter as-is does not fit into flash memory of MCU.

In offline discussion with SRR members we came up with three solutions:
- Static set of kernels: like in tflite micro, code contains fixed subset of tflite kernels.
- Dynamic set of kernels: compile all kernels in form of dynamic library, store this library in large external storage (for example SD card) and load them from interpreter on demand.
- Build time configurable set of kernels: during luci interpreter build time we can choose which kernels to use.

**Static set**

pros:
- This is the easiest approach to implement

cons:
- Can not run arbitrary NNs, because this approach restricts set of available kernels, even if they are available in "big" luci-interpreter. It requires some engineering effort from user to add more kernels
- Could be inefficient in terms of used memory if target NN does not use all kernels presented in interpreter

**Dynamic set**

pros:
- Can run arbitrary NN
- Can update working NN at any time, even after MCU is flashed, for example download NN from internet

cons:
- Additional hardware is needed (like SD card for storing all code and structures)
- Need a lot of engineering effort to implement dynamic code loading on target MCU

**Build time configurable set**

There are ways to configure build:
- By set of target layers
- By set of target models

pros:
- No need for additional hardware to support arbitrary NNs
- Saves resources, because this solution uses only needed code
- Can apply additional optimizations on kernels/runtime based on target models

cons:
- Can not update NN that is flashed on device
- Need to rebuild interpreter for every target model/set of layers

## Conclusion

For now we decided to try **Build time configurable** approach, because:
- "Static" approach is not flexible and effective
- "Dynamic" approach is flexible and effective, but hard to implement and requires additional hardware.
- "Build time" approach is flexible, and provides more optimization opportunities

## Plan

1. Add support of arm-none-eabi toolchain to cmake build script(to load third-party dependencies such as tensorflow, mbed-os e.t.c)
2. Write a python script which configures libraries and generates code (supporting required kernels to run the .tflite/.circle models provided in input parameters) which is ready to build using mbed-cli.
3. Run python script from cmake

SlavikMIPT:
[...]

PR #7237, opened 2021-07-16
[DRAFT] [luci-interpreter] Introduce PAL in luci-interpreter

binarman:
This commit introduces basic HAL to hide system-specific
code, like pthreads.

related issue: #6302


binarman:
+cc @SlavikMIPT @Bronnikoff

seanshpark:
As I know, `HAL: Hardware Adaptation Layer.`
Is this about hardware? I understand this is about kernels...

binarman:
@seanshpark 

> As I know, HAL: Hardware Adaptation Layer.
Is this about hardware? I understand this is about kernels...

Yes, HAL stands for "Hardware Abstraction/Adaptation Layer".

However this PR is not about hardware, but **platform** (Linux, Windows, bare metal, etc.)
I did not use PAL term, because I did not know that it exists :slightly_smiling_face:

Me, @SlavikMIPT and @Bronnikoff had a discussion how to enable luci-interpreter on MCUs(#6302) 

This PR demonstrates a way to specialize luci-interpreter code depending on operating system(or bare metal).

seanshpark:
> However this PR is not about hardware, but platform (Linux, Windows, bare metal, etc.)
I did not use PAL term, because I did not know that it exists 

`PAL` looks OK :)

PR #7241, opened 2021-07-19
[luci-interpreter] Support load BOOL tensor type

seanshpark:
This will enable loading BOOL tensor type.


seanshpark:
For #6708, from draft #7223

jinevening (approved):
LGTM

PR #7242, opened 2021-07-19
[luci] Reorder logex add order in cmake

seanshpark:
This will reorder logex add order in cmake so that other luci modules
can use logex.


seanshpark:
From draft #7223

mhs4670go (approved):
LGTM

PR #7243, opened 2021-07-19
[luci] Revise Module::add for graph

seanshpark:
This will revise Module::add method to return index of added graph.


seanshpark:
For #6708, from draft #7223

seanshpark:
Drop this and add another method (not change interace) for safety...

PR #7244, opened 2021-07-19
[common-artifacts] Revise input/expected message

seanshpark:
This will revise cmake message for input.h5/expcedted.h5 generation.


seanshpark:
To reduce message length as current code shows full path.

seanshpark:
From draft #7223

mhs4670go (approved):
LGTM

PR #7246, opened 2021-07-19
[luci] Introduce TestIsOsGraph in partition

seanshpark:
This will introduce NodeIsOsGraphletT and TestIsOsGraph to support
arbitrary number of inputs and outputs.


seanshpark:
For #6708, from draft #7223

mhs4670go (approved):
LGTM

PR #7247, opened 2021-07-19
[souschef] Revise bool to_number

seanshpark:
This will revise to_number for bool type as atoi will fail if not a number.

### DIFF
--- a/compiler/souschef/src/LexicalCast.cpp
+++ b/compiler/souschef/src/LexicalCast.cpp
@@ -47,9 +47,11 @@ template <> uint8_t to_number(const std::string &s)
 }
 template <> bool to_number(const std::string &s)
 {
-  if (std::stoi(s) || s == "T" || s == "t" || s == "TRUE" || s == "true")
+  if (s == "T" || s == "t" || s == "TRUE" || s == "true" || s == "1")
     return true;
-  return false;
+  if (s == "F" || s == "f" || s == "FALSE" || s == "false" || s == "0")
+    return false;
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6708, from draft #7223

Will throw without this patch...
> terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi
Aborted (core dumped)

compiler/souschef/src/LexicalCast.cpp
```diff
@@ -47,7 +48,11 @@ template <> uint8_t to_number(const std::string &s)
...
 template <> bool to_number(const std::string &s)
 {
-  if (std::stoi(s) || s == "T" || s == "t" || s == "TRUE" || s == "true")
+  if (s == "T" || s == "t" || s == "TRUE" || s == "true")
+    return true;
+  if (s == "F" || s == "f" || s == "FALSE" || s == "false")
+    return false;
+  if (std::stoi(s))
     return true;
   return false;
```
mhs4670go, line 57:
```suggestion
  if (s == "F" || s == "f" || s == "FALSE" || s == "false" || s == "0")
    return false;
  return true;
```
How about this one?

seanshpark: Anything else would be true, which seems a bit strange... like `faLse`

seanshpark: Anyway... I would also like to make this simple but works good...

mhs4670go (approved):
LGTM

PR #7248, opened 2021-07-19
[luci] Use GroupKey for std::string

seanshpark:
This will introduce GroupKey for group key string instead of std::string.


seanshpark:
From draft #7223

mhs4670go (approved):
LGTM

PR #7252, opened 2021-07-19
[luci] Add num_outputs for TestOsGraphlet

seanshpark:
This will add num_outputs attribute for TestOsGraphlet.


seanshpark:
From draft #7223

mhs4670go (approved):
LGTM

PR #7253, opened 2021-07-19
[tflchef] Test for explicit filler for boolean

seanshpark:
This will introduce test recipe for explicit filler for boolean type.


seanshpark:
From draft #7223

mhs4670go (approved):
LGTM

Issue #7230, opened 2021-07-15
Use pip cache while one-compiler install


cgbahk:
While installing `one-compiler` with apt, I found several occurances of these line in install log.
```
The directory '/home/chbahk/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
The directory '/home/chbahk/.cache/pip' or its parent directory is not owned by the current user and caching wheels has been disabled. check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
```

It seems failed to use pip cache. As I understand, `one-compiler` install many big python packages like TF/torch. It would be much faster if we can use pip cache. (I felt like pip download time was more than 70% of all install time)

Why it fails to use pip cache is not clear. My best guess is that install user(i.e. root) has no write permission to the cache directory. I may fix permission setting locally (not tested though), but  it would be better if we find global fix for this :smile:

Log says to try `sudo -H`, but I'm not sure this is applicable or resolve this.

cgbahk:
cc @seanshpark @mhs4670go

seanshpark:
#7218 is related and patch #7219 has landed. please check with nightly build debian package.
I've pinged you in our internal setup repo.

cgbahk:
I've installed `one-compiler 1.16.1-21071519` from nightly, but the symptom was same.

Progress bar means it did download, not using cache

<details>

```
$ sudo apt update && sudo apt install -y one-compiler

Hit:1 XXXXXXXXXXXXXXXXXXXXX bionic InRelease
Hit:2 XXXXXXXXXXXXXXXXXXXXX bionic-dev InRelease
Get:3 XXXXXXXXXXXXXXXXXXXXX bionic-nightly InRelease [4,357 B]
Get:6 XXXXXXXXXXXXXXXXXXXXX bionic-nightly/universe amd64 Packages [22.3 kB]
(Some removed)
Fetched 91.1 kB in 1s (61.7 kB/s)
Reading package lists... Done
Building dependency tree
Reading state information... Done
26 packages can be upgraded. Run 'apt list --upgradable' to see them.
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following packages were automatically installed and are no longer required:
  python3-click python3-colorama
Use 'sudo apt autoremove' to remove them.
The following NEW packages will be installed:
  one-compiler
[...]

PR #7255, opened 2021-07-19
[infra] Revise postinst

mhs4670go:
This commit revies postinst script.

Related: #7230 

### DIFF
--- a/infra/debian/compiler/postinst
+++ b/infra/debian/compiler/postinst
@@ -5,4 +5,8 @@
 
 set -e
 
-/usr/share/one/bin/one-prepare-venv
+# This script is invoked as root except environmental variables,
+# which causes invalid permission problem.
+# e.g. When `pip` installs user packages, it proceeds based on $HOME.
+# To proper installation, $HOME should be root.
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

infra/debian/compiler/postinst
```diff
@@ -5,4 +5,8 @@
 
 set -e
 
-/usr/share/one/bin/one-prepare-venv
+# This script is invoked as root except environmental variables,
+# which causes invalid permission problem.
+# e.g. When `pip` installs user packages, it proceeds based on $HOME.
+# To proper installation, $HOME should be root.
+su - $(whoami) -c '/usr/share/one/bin/one-prepare-venv' # $(whoami) = root
```
cgbahk, line 12:
```
$ man su
...
       -, -l, --login
           Provide an environment similar to what the user would expect had the user logged in
           directly.

           When - is used, it must be specified before any username. For portability it is recommended
           to use it as last option, before any username. The other forms (-l and --login) do not have
           this restriction.
```

cgbahk (approved):
LGTM, thx :)

Issue #7256, opened 2021-07-19
Compiler FE: Invalid buffer index after one-optimize


hseok-oh:
When I tested `mobilenet` tflite model with `fuse_batchnorm_with_tconv` option, most tensor's buffer index changed to special purpose buffer `0`.

#### `.cfg` file for `one-build`

```
[one-build]
one-import-tf=False
one-import-tflite=True
one-import-bcq=False
one-import-onnx=False
one-optimize=True
one-quantize=True
one-pack=False
one-codegen=False

[one-import-tflite]
input_path=./mobilenet_v1_1.0_128.tflite
output_path=./mobilenet_v1_1.0_128.circle

[one-optimize]
input_path=./mobilenet_v1_1.0_128.circle
output_path=./mobilenet_v1_1.0_128.opt.circle
fuse_batchnorm_with_tconv=True

[one-quantize]
input_path=./mobilenet_v1_1.0_128.opt.circle
output_path=./mobilenet_v1_1.0_128.opt.q8.circle
```

#### Parsing optimized model

```
$ python3 ./tools/tflitefile_tool/model_parser.py mobilenet_v1_1.0_128.opt.circle
```

```
[#0 b'' (MAIN)]

#0 b'' (MAIN) input tensors: [0]
        Tensor    0 : buffer Empty |  Empty | FLOAT32 | Memory 192.0K | Shape [1, 128, 128, 3] (b'input')
#0 b'' (MAIN) output tensors: [88]
        Tensor   88 : buffer Empty |  Empty | FLOAT32 | Memory 3.9K   | Shape [1, 1001] (b'MobilenetV1/Predictions/Reshape_1')

Operator 0: CONV_2D
        Fused Activation: RELU6
        Input Tensors[0, 1, 2]
                Tensor    0 : buffer Empty |  Empty | FLOAT32 | Memory 192.0K | Shape [1, 128, 128, 3] (b'input')
                Tensor    1 : buffer 1     | Filled | FLOAT32 | Memory 3.4K   | Shape [32, 3, 3, 3] (b'MobilenetV1/Conv2d_0/weights')
                Tensor    2 : buffer 2     | Filled | FLOAT32 | Memory 128.0B | Shape [32] (b'MobilenetV1/MobilenetV1/Conv2d_0/convolution_bias')
        Output Tensors[3]
[...]

PR #7257, opened 2021-07-19
[luci] Fix wrong buffer id allocation

llFreetimell:
Parent Issue : #7256

Buffer index 0 is used for special purpose.
Therefore, we should not use it without special purpose.
This commit will fix it.


hseok-oh (approved):
LGTM

Issue #7261, opened 2021-07-19
[luci-interpreter] Support Cast


Bronnikoff:
### What? 

Let's support Cast in luci-interpreter!

### Why?

Currently, if we try to execute model with CircleCast:

![image](https://user-images.githubusercontent.com/43731397/126226830-a23e5282-0dbb-45c0-bb7e-5a8a7d7b52a6.png)

we fail due to unsupported operator:

```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Unsupported operator: 9 Cast
```

Bronnikoff:
DRAFT: #7262

Bronnikoff:
Currently, Cast op supported in *luci-interpreter*:
- Builder of kernel with unit test: #7302
- Kernel with unit test: #7295

Let's check luci-interpreter in `master` with introduced changes using tflite model and corresponding circle model!

Model:

![image](https://user-images.githubusercontent.com/43731397/126806803-c860410c-44c1-4686-bb5e-6d7e7167f16c.png)

<details> 
<summary> Check: </summary>

```
$ ./build/compiler/tflite2circle/tflite2circle model.tflite model.circle
$ python3 
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> input_data = 2000 * np.random.rand(1, 500, 500, 32).astype(np.float32) - 1000
>>> with open("input.0", 'wb') as wf:
...     wf.write(input_data.tobytes('C'))
... 
32000000
>>> with open("model.tflite", 'rb') as rb:
...     interp = lite.Interpreter(model_content=rb.read())
... 
>>> interp.allocate_tensors()
>>> interp.set_tensor(0, input_data)
>>> interp.invoke()
>>> output_data = interp.get_tensor(1)
>>> with open("tflite_output.0", 'wb') as wf:
...     wf.write(output_data.tobytes('C'))
... 
32000000
>>> 
$ ./build/compiler/luci-eval-driver/luci_eval_driver model.circle 1 input. circle_output.
$ python3
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> circle_output_path = "circle_output.0"
>>> tflite_output_path = "tflite_output.0"
>>> 
[...]

PR #7262, opened 2021-07-19
 [DRAFT][luci-interpreter] Introduce Cast

Bronnikoff:
This draft introduces Cast to luci-interpreter.

---------------

For: #7261

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Cast.cpp
+++ b/compiler/luci-interpreter/src/kernels/Cast.cpp
@@ -1,0 +1,149 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


compiler/luci-interpreter/src/kernels/Cast.cpp
```diff
@@ -0,0 +1,149 @@
...
+      cast_from_pointer_to_tensor(getTensorData<int64_t>(in_tensor), out_tensor);
+      break;
+    case loco::DataType::FLOAT32:
+      cast_from_pointer_to_tensor(getTensorData<float>(in_tensor), out_tensor);
+      break;
+    case loco::DataType::BOOL:
+      cast_from_pointer_to_tensor(getTensorData<bool>(in_tensor), out_tensor);
+      break;
+    default:
+      throw std::runtime_error("Unsupported input type.");
```
binarman, line 112: Did you take this from tflite kernels?
If so, better to add tensorflow copyright in boilerplace.

binarman: Another option is to write translation from loco types to tflite types and reuse tflite code.

Bronnikoff: > Did you take this from tflite kernels?

I used https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/cast.cc for kernel implementation.

> If so, better to add tensorflow copyright in boilerplace.

Done =)

Bronnikoff:
Close, because changes merged and issue closed.

PR #7264, opened 2021-07-19
[luci/partition] Revise produce_pmodules

seanshpark:
This will revise produce_pmodules to add graph to module before calling
build_graph method.

### DIFF
--- a/compiler/luci/partition/src/PartitionPModules.cpp
+++ b/compiler/luci/partition/src/PartitionPModules.cpp
@@ -184,16 +184,20 @@ luci::PartedModules produce_pmodules(const luci::PGroups *pgroups)
     pm.module = std::make_unique<luci::Module>();
     pm.group = pgroup->group;
 
+    // the main graph for this module
     auto graph = loco::make_graph();
+    auto graph_ptr = graph.get();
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6708, from draft #7223

compiler/luci/partition/src/PartitionPModules.cpp
```diff
@@ -184,16 +184,20 @@ luci::PartedModules produce_pmodules(const luci::PGroups *pgroups)
...
     auto graph_name = make_name(pgroup.get());
     graph->name(graph_name);
 
+    // Add main graph so that other subgraphs can be added inside build_graph
+    pm.module->add(std::move(graph));
+
     INFO(l) << "--- Partition Graph build----------------------";
     INFO(l) << "--- name: " << graph_name;
-    build_graph(graph.get(), pgroup.get());
+    build_graph(graph_ptr, pgroup.get());
```
seanshpark, line 199: we can't use `graph` as it's moved to `pm.module`, so we use pinter `graph_ptr`

mhs4670go (approved):
LGTM

PR #7265, opened 2021-07-20
[luci/service] Add graph dump to validate Module

seanshpark:
This will add graph dump log to validate for Module method to help debugging.


seanshpark:
From draft #7223

mhs4670go (approved):
LGTM

PR #7266, opened 2021-07-20
[res] Introduce Part_If_Add_Sub_001

seanshpark:
This will introduce Part_If_Add_Sub_001 recipe that has IF inside IF node.


seanshpark:
For #6708, from draft #7223

seanshpark:
IF `cond` is undefined... -_-;;;

image

seanshpark:
Now 
image

mhs4670go (approved):
LGTM

PR #7268, opened 2021-07-20
[luci/partition] Introduce clone_ifnode_subgraphs

seanshpark:
This will introduce dummy clone_ifnode_subgraphs method that will clone
subgraphs of If node and revise build_graph method to call it.

### DIFF
--- a/compiler/luci/partition/src/PartitionPModules.cpp
+++ b/compiler/luci/partition/src/PartitionPModules.cpp
@@ -75,10 +75,21 @@ void add_graph_output(loco::Graph *graph, luci::CircleOutput *output_node)
   graph_output->shape(std::move(output_shape));
 }
 
+void clone_ifnode_subgraphs(luci::PartedModule &pm, const luci::CircleIf *if_node,
+                            const luci::CloneContext &clonectx)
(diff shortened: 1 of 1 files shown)


seanshpark:
For #6708, from draft #7223

compiler/luci/partition/src/PartitionPModules.cpp
```diff
@@ -75,10 +75,21 @@ void add_graph_output(loco::Graph *graph, luci::CircleOutput *output_node)
   graph_output->shape(std::move(output_shape));
 }
 
+void clone_ifnode_subgraphs(luci::PartedModule &pm, const luci::CircleIf *if_node,
+                            luci::CloneContext &clonectx)
```
seanshpark, line 79: `const luci::CloneContext &clonectx` seems better to read the code.

seanshpark:
Will be revised after #7269 lands

mhs4670go (dismissed):
LGTM

seanshpark:
@mhs4670go , this was rebased, PTAL again...

mhs4670go (approved):

PR #7269, opened 2021-07-20
[luci/partition] Introduce const_iterator of CloneContext

seanshpark:
This will introduce const_iterator of CloneContext.
- for better understading of used codes.


seanshpark:
For #6708, from draft #7223

mhs4670go (approved):
LGTM

PR #7275, opened 2021-07-20
[circle-part-driver] Remove check for single graph

seanshpark:
This will remove check for single graph as to support multiple subgraphs.


seanshpark:
For #6708, from draft #7223

mhs4670go (approved):
LGTM

PR #7277, opened 2021-07-20
[luci/partition] Enable recursive clone of subgraphs

seanshpark:
This will enable recursive clone of subgraphs for CircleIf then and else.


seanshpark:
For #6708, from draft #7223

PR #7279, opened 2021-07-20
[circle-part-value-test] Add test for If

seanshpark:
This will enable test for If node with subgraphs and another If inside subgraph.


seanshpark:
For #6708, from draft #7223

mhs4670go (approved):
LGTM

PR #7287, opened 2021-07-21
[circle-quantizer] Fix bug in fused TANH quantization

jinevening:
This fixes a bug in fused TANH quantization.


seanshpark (approved):

seanshpark:
@jinevening , it's a bit hard to understand the reason and this patch. As mentioned before, it would be better to make the code better readable, like separate the codes all in `CircleNode` to own handler.

jinevening:
> it's a bit hard to understand the reason and this patch. As mentioned before, it would be better to make the code better readable, like separate the codes all in CircleNode to own handler.

This code is just a simple bug (I don't why I inserted continue there). There is no change in quantization logic.

Actually refactoring is quite complicated.. But I'm thinking about that.

PR #7290, opened 2021-07-21
[luci-interpreter] Add Cast Params

Bronnikoff:
This commit adds Cast's kernel params.

----------------

For: #7261
Draft: #7262

### DIFF
--- a/compiler/luci-interpreter/src/core/KernelParams.h
+++ b/compiler/luci-interpreter/src/core/KernelParams.h
@@ -43,6 +43,12 @@ struct ArgMaxParams
   DataType output_type;
 };
 
+struct CastParams
+{
+  DataType in_data_type;
+  DataType out_data_type;
+};
+
 struct ConcatenationParams
 {
   int axis;


Bronnikoff:
@seanshpark @binarman PTAL =)

compiler/luci-interpreter/src/core/KernelParams.h
```diff
@@ -43,6 +43,12 @@ struct ArgMaxParams
   DataType output_type;
 };
 
+struct CastParams
+{
+  DataType in_data_type;
+  DataType out_data_type;
```
seanshpark, line 49: above `ArgMaxParams` uses just `output_type` not `output_data_type`.
I think it would be better to match naming convention.
On the other hand, how about using `dtype` instead of `type` ?
- `in_dtype`, `out_dtype` or `input_dtype` or `output_dtype`
- fix for `ArgMaxParams` later in another PR.

Bronnikoff: Hmm, I thought I should follow naming convention of luci and use same names of params for kernels, as in corresplonding Circle node:

https://github.com/Samsung/ONE/blob/302627812b713393cce9b4e904862784e2e3c197/compiler/luci/lang/include/luci/IR/Nodes/CircleArgMax.h#L41-L42

https://github.com/Samsung/ONE/blob/302627812b713393cce9b4e904862784e2e3c197/compiler/luci/lang/include/luci/IR/Nodes/CircleCast.h#L38-L42

But if this is not required, I would like to not use CastParams and close this PR, because input/output dtype could be extracted from nodes properties. `node->dtype()` must be provided, when `in_data_type` and `out_data_type` is optional params and may be equal to `Unknown`.

I did PR of kernel without usage of params: #7290, could you please takse a look?

seanshpark: > I did PR of kernel without usage of params: #7290,

this is #7290...

seanshpark: If it was the draft code
```cxx
void Cast::configure()
{
  LUCI_INTERPRETER_CHECK(input()->element_type() == _params.in_data_type);
  LUCI_INTERPRETER_CHECK(_params.out_data_type == loco::DataType::Unknown ||
                         output()->element_type() == _params.out_data_type);

  LUCI_INTERPRETER_CHECK(input()->element_type() != loco::DataType::Unknown);
  LUCI_INTERPRETER_CHECK(output()->element_type() != loco::DataType::Unknown);

  const Shape &shape = input()->shape();
  output()->resize(shape);
}
```
`CastParams` can be removed and not used.

Bronnikoff: > > I did PR of kernel without usage of params: #7290,
> 
> this is #7290...

Ahh, sorry for mistake, I meant #7295.

Bronnikoff:
Let's close! Reason: https://github.com/Samsung/ONE/pull/7290#discussion_r674516234, https://github.com/Samsung/ONE/pull/7290#discussion_r674528256

PR #7284, opened 2021-07-21
[luci] Add UNPACK in QuantizeWithMinMaxPass

hyunsik-yoon:
This makes QuantizeWithMinMaxPass process UNPACK op.

Signed-off-by: Hyun Sik Yoon <hyunsik.yoon@samsung.com>

PR #7293, opened 2021-07-22
[luci] Add Null Check

hyunsik-yoon:
This adds null check into code related to quantization.

From https://github.com/Samsung/ONE/pull/7284#discussion_r674459810, I tried to add similar checks to other operations.

Signed-off-by: Hyun Sik Yoon <hyunsik.yoon@samsung.com>

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -581,6 +581,7 @@ struct QuantizeSpecialActivation final : public luci::CircleNodeMutableVisitor<v
   {
     auto input = loco::must_cast<luci::CircleNode *>(node->input());
     auto i_qparam = input->quantparam();
+    assert(i_qparam);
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -581,6 +581,7 @@ struct QuantizeSpecialActivation final : public luci::CircleNodeMutableVisitor<v
   {
     auto input = loco::must_cast<luci::CircleNode *>(node->input());
     auto i_qparam = input->quantparam();
+    assert(i_qparam);
```
mhs4670go, line 584: (optional) @jinevening I'm just worrying about dereferencing null pointer problem that might be checked by static analysis tool either. Someday, it would be better to make these kind of functions return bool or sth and replace `assert` with return statement.

mhs4670go (approved):
LGTM

jinevening (approved):
LGTM

Issue #7297, opened 2021-07-22
[one-cmds] Don't show python traceback to end user by default


glistening:
Currently, `one-build` shows traceback when it gets error.

For example, running with non-existing config file shows

```
$ one-build -C one-build_00.cfg
Traceback (most recent call last):
  File "/usr/bin/one-build", line 141, in <module>
    main()
  File "/usr/bin/one-build", line 111, in main
    config = _parse_cfg(args)
  File "/usr/bin/one-build", line 74, in _parse_cfg
    raise FileNotFoundError('Not found given configuration file')
FileNotFoundError: Not found given configuration file
```

What about suppress traceback by default in end-user tool?

```
$ one-build -C one-build_00.cfg
FileNotFoundError: Not found given configuration file
```

Error message itself could be updated. "FileNotFoundError" and "Not found given ..." seems superfluous.

It may be good the traceback is shown when the user passes options (e.g. `--verbose` or `--verbose=0`).

glistening:
_Originally posted by @mhs4670go in_ https://github.com/Samsung/ONE/pull/7300#discussion_r675320889


> @glistening I saw your point but the thing is even though we suppress `one-build`, the traceback could be printed because `one-build` executes other `one-cmds` tools as subprocess, which worn off the effectiveness of `sys.tracebacklimit = 0`. I just concerned about this. 
```bash
$ ./one-build -C one-build.template.cfg
2021-07-23 14:48:12.492792: I tensorflow/core/platform/cpu_feature_guard.cc:142] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN)to use the following CPU instructions in performance-critical operations:  AVX2 FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2021-07-23 14:48:12.514823: I tensorflow/core/platform/profile_utils/cpu_utils.cc:104] CPU Frequency: 3600000000 Hz
2021-07-23 14:48:12.515146: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x457a820 initialized for platform Host (this does not guarantee that XLA will be used). Devices:
2021-07-23 14:48:12.515157: I tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): Host, Default Version
Traceback (most recent call last):
  File "/home/seongwoo/test/bin/tf2tfliteV2.py", line 257, in <module>
    main()
[...]

PR #7300, opened 2021-07-23
[one-cmds] Suppress trackback on error from one-build

glistening:
one-build is an end user tool.
It would be good to suppress traceback by default.

Related: #7297

### DIFF
--- a/compiler/one-cmds/one-build
+++ b/compiler/one-cmds/one-build
@@ -27,6 +27,8 @@ import sys
 
 import utils as _utils
 
+# TODO Find better way to suppress trackback on error
+# This suppression is applied only to `one-build`
 sys.tracebacklimit = 0
 
 
--- a/compiler/one-cmds/one-build
+++ b/compiler/one-cmds/one-build
@@ -27,6 +27,8 @@ import sys
 
 import utils as _utils
 
+sys.tracebacklimit = 0
+
 
 def _get_parser():
     parser = argparse.ArgumentParser(


compiler/one-cmds/one-build
```diff
@@ -27,6 +27,8 @@ import sys
 
 import utils as _utils
 
+sys.tracebacklimit = 0
```
mhs4670go, line 30: Could you add TODO comment?
```suggestion
# TODO Find better way to suppress trackback on error
# This suppression is applied only to `one-build`
sys.tracebacklimit = 0
```

glistening: @mhs4670go I intentionally added this line in `one-build` only. Because some tools will be exposed and others won't.
Thus, I have no idea for better ( ? ) solution. But if you have a good idea it could be noted as TODO.

mhs4670go: @glistening I saw your point but the thing is even though we suppress `one-build`, the traceback could be printed because `one-build` executes other `one-cmds` tools as subprocess, which worn off the effectiveness of `sys.tracebacklimit = 0`. I just concerned about this. 
```bash
$ ./one-build -C one-build.template.cfg
2021-07-23 14:48:12.492792: I tensorflow/core/platform/cpu_feature_guard.cc:142] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN)to use the following CPU instructions in performance-critical operations:  AVX2 FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2021-07-23 14:48:12.514823: I tensorflow/core/platform/profile_utils/cpu_utils.cc:104] CPU Frequency: 3600000000 Hz
2021-07-23 14:48:12.515146: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x457a820 initialized for platform Host (this does not guarantee that XLA will be used). Devices:
2021-07-23 14:48:12.515157: I tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): Host, Default Version
  File "/home/seongwoo/test/bin/tf2tfliteV2.py", line 257, in <module>
    main()
  File "/home/seongwoo/test/bin/tf2tfliteV2.py", line 253, in main
    _convert(flags[0])
  File "/home/seongwoo/test/bin/tf2tfliteV2.py", line 234, in _convert
    _v1_convert(flags)
  File "/home/seongwoo/test/bin/tf2tfliteV2.py", line 162, in _v1_convert
    _parse_array(flags.output_arrays), input_shapes)
  File "/home/seongwoo/test/bin/venv/lib/python3.6/site-packages/tensorflow/lite/python/lite.py", line 1822, in from_frozen_graph
    raise IOError("File '{0}' does not exist.".format(graph_def_file))
OSError: File '/path/to/inception_v3.pb' does not exist.
```

mhs4670go (approved):
LGTM

compiler/one-cmds/one-build
glistening: Ah, I've just got your point. `one-build` does not calls actual implementation but spawn new process with `subprocess.Popen`. Thanks. I will consider this case.

glistening:
@mhs4670go Could you please let me know the current policy for landing `one-cmds` changes? 2 approvals? If so, who is the appropriate person to review this issue? Github only suggests you. Or 1 approval is enough?

mhs4670go:
There's no specific policy. 1 approvals is enough I think.

glistening:
@Samsung/nncc_committers PTAL

PR #7292, opened 2021-07-22
[ONE/one-cmds] Fix wrong path in test guide

glistening:
It fixes wrong path `test` to `tests` in test guide.

Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>

PR #7301, opened 2021-07-23
[one-cmds] Add README for tests

mhs4670go:
This commit adds README.md to `tests/` for developers.

Related: #7292

### DIFF
--- a/compiler/one-cmds/tests/README.txt
+++ b/compiler/one-cmds/tests/README.txt
@@ -1,6 +1,8 @@
 one-cmds testing
 ================
 
+NOTE. `one-compiler` should be installed before testing.
+
 Run 'runtestall.sh' program to test ONE command line programs, all at once.
 
 Steps:


glistening:
From https://github.com/Samsung/ONE/pull/7292#issuecomment-885399103, the most helpful section was:

```
# cmake install
$ NNAS_BUILD_PREFIX=build/release ./nnas create-package --preset 20210706 --prefix build/install
```

If you're not talking these line because it is already in [how-to-build-compiler.md](https://github.com/Samsung/ONE/blob/master/docs/howto/how-to-build-compiler.md#build-for-debug-and-release-separately).


(In fact, the command line is somewhat different.

```
$ NNCC_WORKSPACE=build/release ./nncc configure -DCMAKE_BUILD_TYPE=Release
$ ./nncc build
```

It provides `--preset`, `--prefix`, `NNAS_BUILD_PREFIX`. I guess the former is more close to actual command before releasing.

`README.md` doesn't give much information.

With current information only, it may be good to rename `README.txt`, rather than introducing yet-another `README.md`.

The only additional information is

mhs4670go:
@glistening This is the command that I've written in that comment. And, the different points like `NNAS_BUILD_PREFIX=build/release` or `--preset` aren't needed because they are set to default value.
```bash
$ cat how-to-build-compiler.md
..
Above steps will build all the modules in the compiler folder. There are modules that are currently
not active. To build only as of now active modules of the compiler, we provide a preset of modules
to build with below command:

$ ./nnas create-package --prefix $HOME/.local
```

What I wanted to let them know who read `README.md` is that `This test assumes that one compiler is already installed.`, which was confusing you. I thought `README.txt` is for anyone who has one-compier-test installed.

glistening:
> What I wanted to let them know who read `README.md` is that `This test assumes that one compiler is already installed.`, which was confusing you. I thought `README.txt` is for anyone who has one-compier-test installed.

First, it would be good to specify `install` in _"This test assumes that one compiler is already **installed**."_ includes `CMake install`. When I see the word `install`, it looks like I can test only with prebuilt and released package (either *.deb or *.tar in release page). 

Second,  `README.md` and `README.txt` is almost same contents.
Both requires installation before test. I think it can be written in a document.

Third, the test script can be used in source tree, the only reason I cannot use it from source tree was `tests/...` or `test/...`. I still personally think the assumption is too strict. There is no reason to get installed to run tests. However, I honor the author's choice.

mhs4670go:
@glistening 

> Third, the test script can be used in source tree, the only reason I cannot use it from source tree was `tests/...` or `test/...`. I still personally think the assumption is too strict. There is no reason to get installed to run tests. However, I honor the author's choice. 

Please see this comments. https://github.com/Samsung/ONE/pull/7292#issuecomment-884841757 It is not easy to test in the source tree because `runtestall.sh` doesn't exist until installation. `README.txt` is only for anyone who has one-compier-test installed.

But for developers as you said, I add a comment to `README.txt` rather than creating `README.md`.

glistening:
> @glistening
> 
> > Third, the test script can be used in source tree, the only reason I cannot use it from source tree was `tests/...` or `test/...`. I still personally think the assumption is too strict. There is no reason to get installed to run tests. However, I honor the author's choice.
> 
> Please see this comments. [#7292 (comment)](https://github.com/Samsung/ONE/pull/7292#issuecomment-884841757) It is not easy to test in the source tree because `runtestall.sh` doesn't exist until installation. `README.txt` is only for anyone who has one-compier-test installed.
> 
> But for developers as you said, I add a comment to `README.txt` rather than creating `README.md`.

`one-cmds` developers are highly likely already have the environment in their devel envirnment. I thought you don't mention how to build in this `Readme.md` in the same reason.

compiler/one-cmds/tests/README.txt
```diff
@@ -1,6 +1,8 @@
 one-cmds testing
 ================
 
+NOTE. `one-compiler` should be installed before testing.
+
```
glistening, line 5: @mhs4670go This document does not work without installing `one-compiler-test-*.deb` or `nncc-test-*.tar`. It is the reason I've made #7292. As I said several times, I've installed `one-compiler`, but it does not work.

glistening:
@mhs4670go The only additional information from this PR

> NOTE. `README.txt` is for end users.

disappeared after your update because `README.md` is removed.

I think this information should be somewhere.

Option 1. Write in `README.txt`.
Option 2. Add `README.md` for developer.

```
NOTE. `README.txt` is for end users.
We guide to install `one-compiler` and `one-compiler-test` before testing.
Even you can run `one-cmds` tests in your development environment, please install before test.
Here, `install` includes `cmake install`. Please find other materials how to build ...
```

I still don't think it is good solution.

I cannot suggest better document because of
- our policy (install before test)
- inconsistent folder name (`tests` under source tree, `test` under released and build tree)

In addition, it seems to be assumed that `README.md` is for developer and `README.txt` is for end user. I am not sure about it, too.

I think it would be better to keep current document to save you and my time. We've spend much time than expected.
I hope someone who are going to contribute `one-cmds` find information from our issue and PR comments.

mhs4670go:
@glistening Thank you for your explanation and I agree on it. It could be better if more developers are contributing this.

PR #7304, opened 2021-07-23
[Draft] Introduce luci-micro

SlavikMIPT:
This commit adds dependencies configuration for luci-interpreter buid on microcontrollers


seanshpark:
@SlavikMIPT , it seems you are running our CI very frequently and you always have failed. What is the problem?

binarman:
@seanshpark @lemma

> it seems you are running our CI very frequently and you always have failed. What is the problem?

This is WIP draft, I think we can disable testing for this PR until it is stable.
Thank you for your attention!

compiler/luci-micro/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
+
```
binarman, line 1: I was able to build interpreter using this configuration, but it need some refactoring for proper integration within rest of the build.

This is how I see use cases for this CMake:
- Automated build for x86 micro profile for easy integration with CI and automated tests.
- Automated build for arm architecture to ckeck if interpreter build without errors.
- Manually triggered build of test program (build with mbed os). This could be useful for testing/benchmarking purposes.

To achieve integration I propose:
- Remove required from this file, it is needed for mbed-os build only (mbed-os has this required inside already)
- Use existing targets, instead of including dependency directories manually
- Use existing externals infrastructure instead of downloading dependencies in this file.
- Write our own cmake toolchain file for mbed-os instead of patching existing one.
- Remove mbed-os patch, if we use our own toolchain file it is not needed (other changes look redundant in case we don't have spaces in paths).
- Optionally. Add Dockerfile for manual ARM build, because mbed os requires new cmake and python, not available on CI and many developer machines.

binarman:
@nnfw-bot test pr-nncc-debug

binarman:
@nnfw-bot test pr-nncc-release

Issue #7308, opened 2021-07-25
Compiler FE: Support more Ops for NCHW to NHWC


seanshpark:
There's an ONNX model to import that has
- `CONV` - `CONV` - `INSTNORM` - `LEAKYRELU` - `CONV` - `CONV`
- 2'nd and 6'th converts to `DepthWiseConv2D`

As `InstNorm` explodes to seveal Ops, we need to support NCHW to NHWC for
- `SquaredDifference`
- `Rsqrt`
- `Sub`
- `Add` with scalar beta
- `Mul` with both NOT Constant

seanshpark:
CC @jinevening

seanshpark:
done

PR #7309, opened 2021-07-26
DRAFT Support more NCHW to NHWC and Fuse InstanceNorm TF1.15.3

seanshpark:
This will support more Ops for NCHW to NHWC conversion.


seanshpark:
For #7308

seanshpark:
all landed

PR #7310, opened 2021-07-26
[luci] Revise NCHWToNHWCPass Add to support scalar beta

seanshpark:
This will revise NCHWToNHWCPass to support scalar beta to CircleAdd node.


seanshpark:
For #7308, from draft #7309

jinevening:
Code look ok (there may be a better structure with simpler control flow, but it will require a large change..).

Can you update the comment of `is_NCHW_with_const` to reflect this change?

seanshpark:
> Can you update the comment of is_NCHW_with_const to reflect this change?

Maybe `TODO: simplify support for multiple ranks` ?

jinevening:
I meant the comment in the function declaration.

```cpp
// We assume ADD with const input is NCHW if,
// Input shape: (N, C, H, W)
// Output shape: (N, C, H, W)
// 1. Const shape is (1, C, 1, 1)
// 2. Input, Output, Const have the same C.
```

1 should be changed to something like
`// 1. Const shape is (1, C, 1, 1) or a scalar (1)`

jinevening (approved):
LGTM Thank you

PR #7316, opened 2021-07-26
[luci] Revise NCHWToNHWCPass Mul for not Const

seanshpark:
This will revise NCHWToNHWCPass to support Mul with both not Const.


seanshpark:
For #7308, from draft #7309

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -593,9 +593,13 @@ class ConvertNCHWToNHWC final : public luci::CircleNodeMutableVisitor<bool>
     }
     else if (multiplier == nullptr)
     {
```
jinevening, line 595: Can add checks if node->x() and y() are rank 4?

The transpose will be valid only for rank 4 inputs.

jinevening (approved):
LGTM

Issue #7299, opened 2021-07-23
Let's design new driver onecc


jyoungyun:
# onecc

### Installation
By one-compiler package

### How to work?

Default command line interface:

```
onecc -C [cfg file]
```
```
onecc {import | optimize | quantize | pack | codegen | profile(?) | inference(?)} [options]
```
```
onecc --help
usage: onecc [-h] [--version] [-v] [-C CONFIG]
                 [import] [optimize] [quantize] [pack] [codegen]

command line tool to run ONE drivers in customized order

optional arguments:
  -h, --help            show this help message and exit
  --version             show program version number and exit
  -v, --verbose       output additional information to stdout or stderr
  -q, --quiet           suppress stdout (?)
  -C CONFIG, --config CONFIG
                        run with configuation file

sub commands:
  import            Convert given model to circle
  optimize              Optimize circle model
  quantize              Quantize circle model
  pack                  Package circle and metadata into nnpackage
  codegen               Code generation tool
```

**onecc import [options] {tf | tflite | bcq}**
* Run `one-import` tool
* supported type: tf, tflite, bcq
```
one-import [-h] [-C CONFIG] [-v] driver

command line tool to convert various format to circle

positional arguments:
  driver                driver name to run (supported: tf, tflite, bcq)
```

**onecc optimize [options]**
* Run `one-optimize` tool
```
one-optimize [-h] [-v] [-C CONFIG] [-p]
                    [--change_outputs CHANGE_OUTPUTS] [-i INPUT_PATH]
                    [-o OUTPUT_PATH] [--O1] [--convert_nchw_to_nhwc]
                    [--nchw_to_nhwc_input_shape] [--nchw_to_nhwc_output_shape]
                    [--fold_add_v2] [--fold_cast] [--fold_dequantize]
                    [--fold_sparse_to_dense] [--forward_reshape_to_unaryop]
                    [--fuse_add_with_tconv] [--fuse_batchnorm_with_conv]
                    [--fuse_batchnorm_with_dwconv]
                    [--fuse_batchnorm_with_tconv] [--fuse_bcq]
                    [--fuse_preactivation_batchnorm]
                    [--make_batchnorm_gamma_positive]
                    [--fuse_activation_function] [--fuse_instnorm]
                    [--replace_cw_mul_add_with_depthwise_conv]
                    [--remove_fakequant] [--remove_quantdequant]
                    [--remove_redundant_reshape]
                    [--remove_redundant_transpose]
[...]

PR #7318, opened 2021-07-26
[one-import] Support onnx as input parameter

jyoungyun:
This commit adds onnx model case.
It invokes the one-import-onnx model when the parameter is onnx.
It contains the test case for onnx param.

Related issue: #7299

### DIFF
--- a/compiler/one-cmds/one-import
+++ b/compiler/one-cmds/one-import
@@ -37,7 +37,8 @@ def _get_parser():
 
     # driver
     parser.add_argument(
-        'driver', type=str, help='driver name to run (supported: tf, tflite, bcq)')
+        'driver', type=str, help='driver name to run (supported: tf, tflite,' \
+        ' bcq, onnx)')
 
     # version
     dir_path = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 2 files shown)


compiler/one-cmds/tests/one-import_005.test
```diff
@@ -31,7 +33,7 @@ outputfile="test_onnx_model.circle"
 rm -f ${outputfile}
 
 # run test
-one-build -C ${configfile} > ${filename}.log 2>&1
+one-import onnx -C ${configfile} > ${filename}.log 2>&1
```
seanshpark, line 36: it would be better not to change this but add new test

jyoungyun: okay I will add new test case.

seanshpark (approved):
LGTM thank you!

compiler/one-cmds/tests/one-import_005.test
seanshpark: sorry for my mistake... current change is OK.

mhs4670go (approved):
LGTM

Issue #7314, opened 2021-07-26
Compiler FE: Support more patterns in ConvertNCHWToNHWC


jinevening:
## What

Let's support more patterns in ConvertNCHWToNHWC

1. PadV2
![image](https://user-images.githubusercontent.com/5449554/126932602-a82ba5f4-3acc-48ac-991e-cff95e8c7458.png)

2. TConv with output padding
![image](https://user-images.githubusercontent.com/5449554/126932571-0b173449-8b59-42bd-9486-5bd8d5edabc8.png)

jinevening:
All done.

PR #7321, opened 2021-07-26
[luci] Add annotation for NHWC operators.

jinevening:
This adds annotation for NHWC operators.
The annotated operators are not converted.

---
Related to: #7314


seanshpark (approved):
LGTM

seanshpark:
Just curious, related task didn't mention about this so was there any problems that this patch will fix?

jinevening:
> Just curious, related task didn't mention about this so was there any problems that this patch will fix?

This PR resolves the second item of #7314 by handling multiple ops inside pre-Transpose and post-Transpose.

PR #7324, opened 2021-07-26
[luci] Support Sub in NCHWToNHWCPass

seanshpark:
This will revise to support Sub node in NCHWToNHWCPass.

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -466,6 +466,45 @@ bool is_NCHW_with_const(const luci::CircleAdd *node, luci::CircleNode *&pred_nod
     return false;
 }
 
+// We assume SUB with const input is NCHW if,
+// Input shape: (N, C, H, W)
+// Output shape: (N, C, H, W)
+// 1. Const shape is (1, C, 1, 1)
+// 2. Input, Output, Const have the same C.
+bool is_NCHW_with_const(const luci::CircleSub *node, const luci::CircleNode *pred_node,
+                        const luci::CircleConst *subtract)
+{
+  assert(pred_node != nullptr);
+  assert(subtract != nullptr);
+
+  if (pred_node->rank() != 4)
+    return false;
+
+  const auto const_rank = subtract->rank();
+  if (const_rank != 4)
+    return false;
+
+  // Check the shape is (1, C, 1, 1)
+  for (uint32_t i = 0; i < const_rank; i++)
+  {
+    if (i == 1)
+      continue;
+
+    if (subtract->dim(i).value() != 1)
+      return false;
+  }
+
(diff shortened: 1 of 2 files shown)


seanshpark:
For #7308, from draft #7309

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -823,6 +905,73 @@ class ConvertNCHWToNHWC final : public luci::CircleNodeMutableVisitor<bool>
...
+      auto nhwc_const = create_NHWC_from_NCHW(subtract);
+      if (nhwc_const == nullptr)
+        return false;
+
+      node->x(pre_trans);
+      node->y(nhwc_const);
+    }
+    else if (subtract == nullptr)
+    {
+      // Both inputs are not constant.
```
jinevening, line 942: Is this true? It seems that both `is_NCHW_with_const_x` and `is_NCHW_with_const_y` can return false even when one of x/y is const.

seanshpark: Hum... it was suspicious to me too... I'll revise this in another way...

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -823,6 +905,77 @@ class ConvertNCHWToNHWC final : public luci::CircleNodeMutableVisitor<bool>
...
+      if (nhwc_const == nullptr)
+        return false;
+
+      node->x(pre_trans);
+      node->y(nhwc_const);
+    }
+    else
+    {
+      auto const_x = dynamic_cast<luci::CircleConst *>(node->x());
+      auto const_y = dynamic_cast<luci::CircleConst *>(node->y());
```
jinevening, line 943: This pattern appears three times in this PR. Actually, this information is used in all three cases.

I think the if statements can be restructured as follows.

```cpp
auto const_x = dynamic_cast<luci::CircleConst *>(node->x());
auto const_y = dynamic_cast<luci::CircleConst *>(node->y());

if (const_x == nullptr && const_y != nullptr)
{
  // Check NCHW for const_y
  // Insert pre-transpose
}
else if (const_x == nullptr && const_y != nullptr)
{
  // Check NCHW for const_x
  // Insert pre-transpose

}
else if (const_x == nullptr && const_y != nullptr)
{
  as-is
}
```

`post_NCHW_with_const` may be used for checking NCHW.

seanshpark: Thanks for the suggestion!

seanshpark:
@jinevening , now the code looks better! :)

jinevening (approved):
LGTM Thank you

seanshpark:
@nnfw-bot test nncc-debug

seanshpark:
@nnfw-bot test nncc-release

PR #7330, opened 2021-07-26
[res] Introduce Net_InstanceNorm_007

seanshpark:
This will introduce Net_InstanceNorm_007 recipe.


seanshpark:
For #7032, from draft #7309

mhs4670go (approved):
LGTM

Issue #7329, opened 2021-07-26
Compiler FE: Support CircleMean with keep_dims = False in ConvertNCHWToNHWC


kvochko:
### What

Let's support CircleMean with keep_dims = false in ConvertNCHWToNHWC pass!

### Why

This is required by some models.

kvochko:
All done

PR #7334, opened 2021-07-27
[luci] Support NHWC conversion for Mean with keep_dims=false

kvochko:
This commit adds support for NCHW-NHWC conversion of CircleMean
operation when keep_dims is set to false.

ONE-DCO-1.0-Signed-off-by: Andrey Kvochko <a.kvochko@samsung.com>

------

Related: #7329

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -98,6 +98,46 @@ bool check_4d_transpose(loco::Node *node, const std::vector<int32_t> indices)
   return true;
 }
 
+luci::CircleTranspose *create_4d_transpose(luci::CircleNode *node,
+                                           const std::vector<int32_t> indices)
+{
+  assert(indices.size() == 4);
+
+  auto name = node->name();
+  assert(name.length() > 0);
+
+  auto perm = node->graph()->nodes()->create<luci::CircleConst>();
+  perm->dtype(loco::DataType::S32);
+  perm->size<loco::DataType::S32>(4);
+  perm->rank(1);
+  perm->dim(0) = 4;
+  for (uint32_t i = 0; i < 4; i++)
+    perm->at<loco::DataType::S32>(i) = indices[i];
+  perm->shape_status(luci::ShapeStatus::VALID);
+
+  auto make_string = [](const std::vector<int32_t> &nums) {
+    std::string str;
+    for (auto num : nums)
+    {
+      if (str.length() > 0)
+        str += ".";
+      str += std::to_string(num);
+    }
+    return str;
+  };
+
+  auto str_indices = make_string(indices);
+
+  perm->name(name + "/Transpose_" + str_indices + "/perm");
+
+  auto trans = node->graph()->nodes()->create<luci::CircleTranspose>();
+  trans->perm(perm);
+  trans->name(name + "/Transpose_" + str_indices);
+  luci::add_origin(trans, luci::get_origin(node));
+
+  return trans;
+}
+
 luci::CircleTranspose *create_Nd_transpose(luci::CircleNode *node,
                                            const std::vector<int32_t> indices)
 {
@@ -147,17 +187,12 @@ int32_t nchw_axis_to_nhwc(int32_t axis)
 
 luci::CircleTranspose *create_post_transpose(luci::CircleNode *node)
 {
-  return create_Nd_transpose(node, {0, 3, 1, 2});
+  return create_4d_transpose(node, {0, 3, 1, 2});
 }
 
 luci::CircleTranspose *create_pre_transpose(luci::CircleNode *node)
 {
-  return create_Nd_transpose(node, {0, 2, 3, 1});
-}
-
-luci::CircleTranspose *create_post_transpose_3d(luci::CircleNode *node)
-{
-  return create_Nd_transpose(node, {0, 2, 1});
+  return create_4d_transpose(node, {0, 2, 3, 1});
 }
 
 bool is_post_transpose(loco::Node *node) { return check_4d_transpose(node, {0, 3, 1, 2}); }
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -1047,7 +1070,11 @@ bool ConvertNCHWToNHWCPass::run(loco::Graph *g)
       ConvertNCHWToNHWC converter;
       auto circle_node = loco::must_cast<luci::CircleNode *>(node);
       if (circle_node->rank() != 4)
-        continue;
+      {
+        // Q(kvochko) What will happen if we remove rank-4 check?
```
binarman, line 1074: I suppose it could generate wrong IR for some operations. (insert 4d transpose for non 4d tensors)

kvochko: I guess so. I'm only a bit concerned about this workaround, because we are starting to allow non-4d nodes in this pass. If more such nodes are introduced, this condition will grow and it could turn out that it's better to check assert that the rank of a node's input is 4, rather than of the node itself.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -98,20 +98,18 @@ bool check_4d_transpose(loco::Node *node, const std::vector<int32_t> indices)
   return true;
 }
 
-luci::CircleTranspose *create_4d_transpose(luci::CircleNode *node,
+luci::CircleTranspose *create_Nd_transpose(luci::CircleNode *node,
```
seanshpark, line 101: This change has internal relation but not direct. I recommand to separate this change.
And it would be better to leave `create_4d_transpose` as is and add new `create_Nd_transpose` .

seanshpark:
@jinevening , PTAL

compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
```diff
@@ -696,6 +701,30 @@ TEST(ConvertNCHWToNHWC, Mean)
   EXPECT_EQ(2, new_rindices->at<loco::DataType::S32>(1));
 }
 
+TEST(ConvertNCHWToNHWC, Mean_keep_dims_false)
```
seanshpark, line 704: Could you please add negative test that will fail with non 3d?
And maybe if possible, more for rejected conditions.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -149,12 +147,17 @@ int32_t nchw_axis_to_nhwc(int32_t axis)
...
 
 luci::CircleTranspose *create_pre_transpose(luci::CircleNode *node)
 {
-  return create_4d_transpose(node, {0, 2, 3, 1});
+  return create_Nd_transpose(node, {0, 2, 3, 1});
+}
+
+luci::CircleTranspose *create_post_transpose_3d(luci::CircleNode *node)
+{
+  return create_Nd_transpose(node, {0, 2, 1});
```
jinevening, line 160: Does {0, 2, 1} work for all cases?

Let's assume reduction_indices = 0 (N).

For NCHW (before transformation), the output will be CHW
For NHWC, the output will be HWC, so 3d transpose should be {2, 0, 1}.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: Is it possible to transfrom NCHW -> NHWC in a non-rank 4 node?

For example, how can you identify the original layout of rank 3, 2, or 5?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
kvochko: Ah, I completely missed the possibility that N axis can be reduced, let me rethink this PR

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
kvochko: I separated out `create_Nd_transpose`, but, if possible, can I keep it in this PR? This way, we introduce a helper method and use it in the same pull request.

jinevening (dismissed):
LGTM

binarman (approved):
LGTM

kvochko:
@seanshpark could you please take a look?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
```diff
@@ -696,6 +707,120 @@ TEST(ConvertNCHWToNHWC, Mean)
...
+    EXPECT_EQ(tc.nhwc_ind.size(), new_rindices->dim(0).value());
+    EXPECT_EQ(tc.nhwc_ind.size(), new_rindices->size<loco::DataType::S32>());
+    for (uint32_t i = 0; i < tc.nhwc_ind.size(); ++i)
+    {
+      EXPECT_EQ(tc.nhwc_ind[i], new_rindices->at<loco::DataType::S32>(i));
+    }
+  }
+}
+
+TEST(ConvertNCHWToNHWC, ConvertNCHWToNHWC_Mean_keep_dims_false_N)
```
seanshpark, line 767: please use `_NEG` suffix for negative test

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -675,15 +713,59 @@ class ConvertNCHWToNHWC final : public luci::CircleNodeMutableVisitor<bool>
...
+    uint32_t num_reduced_indices = nhwc_rindices->size<loco::DataType::S32>();
+
+    for (uint32_t ri = 0; ri < num_reduced_indices; ++ri)
+    {
+      reduced_dims_nhwc[nhwc_rindices->at<loco::DataType::S32>(ri)] = true;
+    }
+
+    // if channel dimension has been reduced, we don't need a transpose
+    if (reduced_dims_nhwc[3])
+      return true;
```
seanshpark, line 749: @jinevening , Q) return true here will mark this node as `NHWC`, is this correct?

jinevening: I think it is ok.

The output is not rank 4, but the behavior fits our intention (for removing redundant Transpose).

seanshpark (approved):
LGTM thank you!

jinevening (approved):

binarman (approved):

Issue #7288, opened 2021-07-21
[luci-interpreter] Enable MCU configurable build


binarman:
This task is generalization of #6302

## Goal

Enable MCU build of luci-interpreter along with the regular version in ONE.

## Rationale

To run NN on MCUs we need lightweights runtime, luci-interpreter seems suitable for this task.

## Details

During offline discussions we found several challenges:
1. Make MCU build developer friendly. For now build of luci-micro requires a lot of manual actions, like toolchain downloading, project configuration, etc.
2. Make it possible to build luci-interpreter for different platforms in ONE repo without interference between them: enabling of one platform **should not degrade features/performance** of other supported platforms. Currently if we want to build interpreter for bare metal we need to switch to slower kernels on all platforms.
3. Require as little RAM as possible for interpreter itself. For now there are no convenient mechanism to build subset of all available kernels (described as a **Build time configurable** approach mentioned above).

To achieve goals above we came up with several steps:

- [x] Refactor `KernelBuilder`, to simplify it's configuration with set of supported operators during build for different platforms. (Current design contains multiple `KernelBuilders` and make it hard to configure subset of used kernels.
  - One possible solution is to revive old "large" `KernelBuilder`, put it's visit methods into different files and generate declaration of `KernelBuilder` during CMake stage. (need to check if SAM is ok with this solution)
  - Another possible solution is to make `KernelBuilder` dynamic and fill opcode->builder mapping in runtime. (I suppose SAM score will be ok, but need to check readability of such code)
- [x] Enable partial interpreter build from list of operators.
  - Introduce list of supported operations for each target platform and implement infrastructure to build only selected kernels from this list.
- [x] Implement Hardware Abstraction Layer (maybe "Platform Abstraction Layer", we need to decide terminology yet) to abstract from available OS resources, for example threads (unavailable on bare metal).
  - Implement Abstract layer to specialize different luci-interpreter builds variants without interference between them. See draft #7237 for example of such AL.
- [x] Build integration of luci-interpteter for MCU into main build system.
  - Add wrappers around OS dependent components in various `CMakeList.txt`, to conditionally enable/disable these builds for different platforms.
  - Automate download and configuration of target MCU toolchain.
  - Integrate MCU specific build into main build sequence.

**Upd**. We have all part in form of drafts, but it is hard to merge them like this, because they have a lot of dependencies between them. Instead we will break large DRAFTs in smaller PRs, plan TBD below.
  
**Related PRs**

[...]

PR #7298, opened 2021-07-22
[DRAFT] [luci-interpreter] Introduce KernelBuilderRegistry

binarman:
This commit introduces KernelBuilderRegistry that is configured in runtime.

related issue: #7288

ONE-DCO-1.0-Signed-off-by: Alexander Efimov <a.efimov@samsung.com>

PR #7289, opened 2021-07-21
[DRAFT][luci-interpreter] Add optional kernel build

m-bronnikov:
This commit adds optional build of kernels in luci-interpreter.

ONE-DCO-1.0-Signed-off-by: Maksim Bronnikov <max120199@gmail.com>

--------------------

For: #7288

PR #7337, opened 2021-07-27
[luci] Remove luci namespace from operators list

binarman:
This commit removes lusi:: namespace from CircleNodes.lst.

For #7288


binarman:
This is related to luci-interpreter KernelBuilder refactoring.
I have two possible variants for now: #7298 (this PR is part of this) and #7338

I am still trying to invent best solution, because both existing have some disadvantages.

seanshpark:
@binarman , I can't catch why this change is needed, can you please explain a bit?
Change it self seems OK.

binarman:
> @binarman , I can't catch why this change is needed, can you please explain a bit?
Change it self seems OK.

This PR is definitely confusing =)
This is why I worked on different solutions yesterday.

Back to your question:

**Prerequisites**
This change is part of #7298.

in #7288 (luci-interpreter for MCU) we want to make luci-interpreter build-time configurable (compile only needed kernels, instead of all collection).
For this we need:
1) Select only needed source files in cmake.
2) Generate KernelBuilder with only needed visit methods.

@Bronnikoff proposed to use configuration file with list of needed operators (see https://github.com/Samsung/ONE/pull/7289/files#diff-aeca1300daf2df3fa6dc67ca4b218d7e7978dbe0a138ae4e5f83c6fea5d4ddbd)
This file should be included in both CMake (select files to compile) and KernelBuilder visitor.

**Problem**
Original solution removed KernelBuilder refactoring you did some time ago (by introducing smaller "KernelBuilderLets") this dropped SAM score.

**My solution**
To fit both analysis and configurability I proposed to replace traditional visitor with builder registry that maps opcode to builder function (different implementations in #7298 (this PR is part of this), #7338 and #7339).
So we first fill builder registry with pairs (opcode, builder_function)
when we want to generate kernel we gets builder by opcode and calls it.

**FINALLY, THE ANSWER**
Original configuration file contains only kernel name, but BuilderRegistry from my solution needs both **opcode** and builder name to be filled, I used `luci/IR/CircleNodes.lst` to get opcode information from kernel name.
Unfortunately `luci::` prefix gets in the way =(

P.s. I hope we will use different solution (one of #7338 and #7339), so I'll close this PR.

binarman:
oops, forgot to add mention
@seanshpark

binarman:
This draft is implemented and merged in other PR. closing

Issue #7306, opened 2021-07-24
Compiler FE: Fix typo in MaxPoolWithArgmax 


Bronnikoff:
### What?

Let's change `MaxPoolWithArgMax` to `MaxPoolWithArgmax` in all modules.

### Why? 

Currently, in *compiler/tflchef* and *res/TensorflowLiteRecipes* we use `MaxPoolWithArgMax` as name for castom op, but Tensorflow provides node with name [MaxPoolWithArgmax](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/max-pool-with-argmax) and conversion of corresponding `tf` models to `tflite` produces castom node with name `MaxPoolWithArgmax`. 

![image](https://user-images.githubusercontent.com/43731397/126882491-a9f6f0bd-c44a-48be-b994-6e125cc00520.png)

This misleading makes harder to provide `circle2circle-dredd-recipe-test` for #7236.

- [x] Support both variants of tests in tflchef #7344
- [x] Disable new tests in common-artifacts #7345
- [x] Fix resources #7390 
- [x] Apply changes in tflchef #7407
- [x] Cleanup common-artifacts (remove old names) #7406

Bronnikoff:
Draft: #7307

binarman:
everything is merged, closing

PR #7344, opened 2021-07-28
[tflchef] Support MaxPoolWithArgmax opcode

binarman:
This commit adds support of MaxPoolWithArgmax in addition with MaxPoolWithArgMax.

Related issue: #7306


compiler/tflchef/core/src/CustomOp/MaxPoolWithArgMax.cpp
```diff
@@ -29,7 +29,7 @@ MaxPoolWithArgMaxChef::custom_value(flatbuffers::FlatBufferBuilder &fbb) const
 {
   auto &operation = (*_operation);
 
-  assert(operation.type() == "MaxPoolWithArgMax");
+  assert(operation.type() == "MaxPoolWithArgMax" || operation.type() == "MaxPoolWithArgmax");
```
AShedko, line 32: Not a typo, I'm guessing?

AShedko (approved):
LGTM, assuming this isn't a typo

compiler/tflchef/core/src/CustomOp/MaxPoolWithArgMax.cpp
binarman: you are right, this the idea of this PR.

Currently we have wrong spelling in both tflchef and recipes.
This PR enables correct behavior and preserves old one.

Why not just rename everything?
Because I can not change tflchef and recipes in one PR and can not disable tests without deletion of recipes.
Separate changes in tflchef or recipes will break dredd tests (and only way to disable them is to delete corresponding recipes, I don't want to do this)

seanshpark: I assume you'll remove `MaxPoolWithArgMax` later when all has changed.

seanshpark (approved):
LGTM

Issue #7346, opened 2021-07-29
[onert/lib] Remove ported NNAPI delegator from tflite library


hseok-oh:
### What?

We have ported NNAPI delegator in `runtime/libs/tflite/port`. Let's remove this extension.

### Why?

When we introduced ported NNAPI delegator, we don't have our API to load TFLite model for our runtime and support NNAPI only. But now, we have our API to load tflite (nnpackage) model and and we can validate runtime's TFLite model inference with TensorFlow Lite by `tflite_comparator` tool. 

Maintaining `runtime/libs/tflite/port` makes hard to update tflite version for runtime tools, and we didn't maintain this for a while. And we don't need to maintain `NeuralNetworksEx.h` any more.

hseok-oh:
cc/ @chunseoklee

chunseoklee:
@hseok-oh Then, we will remove all parts depending on nnapi delegator ?

hseok-oh:
@chunseoklee Maybe we will remove all codes under `runtime/libs/tflite/port`, and update depending code to use tflite directly.

hseok-oh:
`runtime/libs/tflite/port` is removed.

PR #7347, opened 2021-07-29
Draft: Remove runtime/libs/tflite/port

hseok-oh:
Remove TFLite port: NNAPI Delegator & Op Resolver

---

Draft for #7346


chunseoklee:
LGTM.

PR #7349, opened 2021-07-30
[circle-quantizer] Propagate qparam of Pack

jinevening:
This propagates qparam of Pack.

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1054,7 +1054,7 @@ void propagate_pack_quantparam(luci::CirclePack *pack, loco::DataType quant_type
   {
     auto node = loco::must_cast<luci::CircleNode *>(pack->arg(i));
 
-    // Skip if this input is CONCAT Op
+    // Skip if this input is PACK Op
     if (node->opcode() == luci::CircleOpcode::PACK)
       continue;
 
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1024,6 +1024,74 @@ private:
   bool visit(luci::CircleNode *) { return false; }
 };
 
+/** EXAMPLE
+ *
+ * BEFORE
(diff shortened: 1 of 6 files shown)


jinevening:
@seanshpark This is necessary for supporting PACK in our backend.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1024,6 +1024,74 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
+  assert(pack->quantparam() != nullptr);
+
+  const auto num_inputs = pack->values_count();
+
+  for (uint32_t i = 0; i < num_inputs; i++)
+  {
+    auto node = loco::must_cast<luci::CircleNode *>(pack->arg(i));
+
+    // Skip if this input is CONCAT Op
+    if (node->opcode() == luci::CircleOpcode::PACK)
```
seanshpark, line 1058: Comment and `opcode` differs...

seanshpark: (optional) it would be nice to have a test case for this.

jinevening: Ah, I've updated the comment.

I also added a test case for this behavior.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1024,6 +1024,74 @@ struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<bool>
...
+ *                     (qparam2)
+ *
+ *  AFTER
+ *
+ *         [CircleNode]        [CircleConst]   [CircleConst] <- Dead node
+ *           (qparam2)           (qparam2)         (FP32)
+ *                   \            /
+ *                    \          /
+ *                    [CirclePack]
+ *                     (qparam2)
```
seanshpark, line 1045: (optional) there is `qparam1` and `qparam2` which I could not find about this in this PR.
it would help understand if there is some notes about this.

seanshpark (approved):
LGTM thank you!

Issue #7350, opened 2021-07-30
Compiler FE: Turn on proper options when converting NCHW to NHWC


jinevening:
## What

Let's automatically turn on proper options when converting NCHW to NHWC

## Why

It is difficult for users to find and turn on the options.

List of enabled options
```cpp
    // Options necessary to remove redundant Ops (Transpose/Slice/Reshape/Split) generated by
    // onnx-tf Related to https://github.com/Samsung/ONE/issues/5401
    options->enable(Algorithms::ForwardReshapeToUnaryOp);
    options->enable(Algorithms::RemoveRedundantTranspose);
    options->enable(Algorithms::RemoveRedundantReshape);
    options->enable(Algorithms::RemoveUnnecessaryReshape);
    options->enable(Algorithms::RemoveUnnecessarySlice);
    options->enable(Algorithms::RemoveUnnecessarySplit);
    options->enable(Algorithms::SubstituteTransposeToReshape);
    options->enable(Algorithms::SubstituteSqueezeToReshape);

    // Constant folding of Ops created when onnx-tf converts Reshape Op
    // Related to https://github.com/Samsung/ONE/issues/5598
    options->enable(Algorithms::FoldSparseToDense);
    options->enable(Algorithms::FoldAddV2);
    options->enable(Algorithms::FoldCast);
```

seanshpark:
CC @lemmaa

seanshpark:
@jinevening , I'm not confident of enabling all related options for NCHW to NHWC at this time.
It would be better to have some test and validation to be sure.
Just turning on the option and see what happens can be a way but may bring dangerous situation.
If you can take the chance I'll let you do it.

seanshpark:
@jinevening , ended milestone is set to this task (and the PR). please move to another milestone or remove it.

jinevening:
I removed the milestone, because this task is postponed.

PR #7351, opened 2021-07-30
[circle2circle] Enable proper options when converting NCHW to NHWC

jinevening:
This automatically enables proper options when convertin NCHW to NHWC.

---
Related to #7350


seanshpark:
It would be better to introduce another option something like `--O2`, instead of enabling other options.
- something like `--enable_nchw2nhwc`

jinevening:
This PR is outdated.

PR #7356, opened 2021-08-01
[bcq-tools] Suppress trackback on error from generate_bcq_metadata.py

glistening:
It suppresses traceback from generate_bcq_metadata.py.


mhs4670go (approved):
LGTM

PR #7357, opened 2021-08-01
[tf2tfliteV2] Suppress trackback on error from tf2tfliteV2

glistening:
It suppresses traceback from tf2tfliteV2.py.


mhs4670go (approved):
LGTM

Issue #7280, opened 2021-07-20
Compiler FE: Partition enable with WHILE Op


seanshpark:
As for next of IF op, lets' enable partitioning with WHILE Op.

seanshpark:
Next of #6708

seanshpark:
Prepare a test recipe: WHILE inside WHILE subgraph

seanshpark:
Done

PR #7358, opened 2021-08-02
DRAFT: Compile FE partition WHILE

seanshpark:
on-going draft to enable WHILE in compiler frontend graph partition.

### DIFF
--- a/compiler/circle-partitioner-test/parts/Part_While_000.part
+++ b/compiler/circle-partitioner-test/parts/Part_While_000.part
@@ -1,0 +1,7 @@
+[partition]
+backends=cpu,acl_cl
+default=cpu
+comply=opcode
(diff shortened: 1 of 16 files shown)


seanshpark:
For #7280

seanshpark:
`Part_While_000` `main` graph `ofm` shape is known but shape status is `UNDEFINED`
--> `assert(circle_node->shape_status() != luci::ShapeStatus::UNDEFINED);` fails in `bool validate_shape_dtype(loco::Graph *g)`

image

seanshpark:
Seems necessary codes to partition WHILE has landed.

Now let's enable common-artifacts test with `While_00?`.

seanshpark:
done for now

PR #7359, opened 2021-08-02
[one-cmds] Add verbose option to utils

jyoungyun:
This commit allows to add verbose option to all one-xxx tools.

Related issue: #7299

### DIFF
--- a/compiler/one-cmds/utils.py
+++ b/compiler/one-cmds/utils.py
@@ -89,6 +89,15 @@ def _add_default_arg(parser):
         action='store_true',
         help='show program\'s version number and exit')
 
+    # verbose
+    parser.add_argument(
+        '-V',
+        '--verbose',
+        action='store_const',
+        default='0',
+        const='1',
+        help='output additional information to stdout or stderr')
+
     # configuration file
     parser.add_argument('-C', '--config', type=str, help='run with configuation file')
     # section name that you want to run in configuration file


compiler/tf2tfliteV2/tf2tfliteV2.py
```diff
@@ -17,7 +17,24 @@
...
 import tensorflow as tf
 import argparse
+import os
 import sys
+# NOTE
+# TF_CPP_MIN_LOG_LEVEL
+#   0 : INFO + WARNING + ERROR + FATAL
+#   1 : WARNING + ERROR + FATAL
+#   2 : ERROR + FATAL
+#   3 : FETAL
```
glistening, line 27:
```suggestion
#   3 : FATAL
```

compiler/one-cmds/utils.py
```diff
@@ -89,6 +89,15 @@ def _add_default_arg(parser):
         action='store_true',
         help='show program\'s version number and exit')
 
+    # verbose
+    parser.add_argument(
+        '-V',
+        '--verbose',
+        action='store_const',
+        default='0',
```
mhs4670go, line 97: Isn't it `2` for normal case?

compiler/one-cmds/one-import-tf
```diff
@@ -123,11 +123,18 @@ def _verify_arg(parser, args):
         parser.error('the following arguments are required: ' + ' '.join(missing))
 
 
+def _set_verbose_level(level):
+    """set verbose environment"""
+    os.environ['ONE_IMPORT_TF_VERBOSE'] = str(level)
```
mhs4670go, line 128: How about using argument option instead of environment variable? I mean, let `one-import-tf` receive verbose option.

compiler/one-cmds/utils.py
mhs4670go: Plus, how about using `count` action?
```python
parser.add_argument('-V', '--verbose', action='count', default=0)
```
Then, seems better argument parsing.
`-V` -> count 1
`-VV` -> count 2
`-VVV` -> count 3
`` -> count 0
`-V -V` -> count 2
..

glistening: > Isn't it `2` for normal case?

Good point.  We need to decide.

Q. Is the meaning of verbose level same to `TF_CPP_MIN_LOG_LEVEL`? (0: All, ..., 3: Fatal)

glistening: > Plus, how about using `count` action?
> 
> ```python
> parser.add_argument('-V', '--verbose', action='count', default=0)
> ```
> 
> Then, seems better argument parsing.
> `-V` -> count 1
> `-VV` -> count 2
> `-VVV` -> count 3
> ` ` -> count 0
> `-V -V` -> count 2
> ..

I would like to provide consistent style for our tool's arguments. Most tools are written in python, however, if we have some tool written in C++, the tool is not likely to provide the handling of `-VVV` and so on.

mhs4670go: @glistening 
> Q. Is the meaning of verbose level same to `TF_CPP_MIN_LOG_LEVEL`? (0: All, ..., 3: Fatal)

Oh, I misunderstood the codes. Current change looks okay to me.

> I would like to provide consistent style for our tool's arguments. Most tools are written in python, however, if we have some tool written in C++, the tool is not likely to provide the handling of `-VVV` and so on.

Well, okay. I just thought current tools for users is `one-cmds` and it only consists of python scripts. Both ways looks good to me.

compiler/one-cmds/one-import-tf
jyoungyun: In this PR, `one-import-tf` takes a verbose option as argument parameters. The environment is for tf2tfliteV2 tool.

compiler/one-cmds/utils.py
jyoungyun: > Plus, how about using count action?

This is a good suggestion. However our verbose level has not yet been precisely defined, so I want to expand it later. Currently it supports to turn verbose option on and off. Also, additional consideration is needed on how to handle arguments in our multiple type of tools.

compiler/one-cmds/one-import-tf
mhs4670go: Oh, verbose option is added to `_add_default_arg`. But then, verbose option will be added to all the `one-cmds` tools because they invoke `_add_default_arg`. And they don't have anything that take care of that option. Maybe it might as well be added to this file as of now.

> The environment is for tf2tfliteV2

Then, how about adding verbose option to `tf2tfliteV2` as well?

compiler/tf2tfliteV2/tf2tfliteV2.py
```diff
@@ -16,19 +16,32 @@
...
+# TODO Find better way to suppress trackback on error
+# tracebacklimit
+#   The default is 1000.
+#   When set to 0 or less, all traceback information is suppressed
+if os.getenv('ONE_IMPORT_TF_VERBOSE') == '1':
+    os.environ['TF_CPP_MIN_LOG_LEVEL'] = '0'
+    sys.tracebacklimit = 1000
+else:
+    os.environ['TF_CPP_MIN_LOG_LEVEL'] = '2'
+    sys.tracebacklimit = 0
```
mhs4670go, line 37: As of now, we have a convention that only one module needs to be changed per PR. This change should be posted to another PR.

jyoungyun: Okay. I will create another PR for this.

jyoungyun:
PTAL

mhs4670go (approved):
LGTM

jyoungyun:
I will make a separated PR for `one-import-tf`.

glistening (approved):
LGTM

seanshpark (approved):
LGTM

PR #7360, opened 2021-08-02
[one-cmds] Add one-profile tool

jyoungyun:
This commit adds one-profile.
It will execute profile of backend model file.

Related issue: #7299

### DIFF
--- a/compiler/one-cmds/one-profile
+++ b/compiler/one-cmds/one-profile
@@ -1,0 +1,171 @@
+#!/usr/bin/env bash
+''''export SCRIPT_PATH="$(cd "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")" && pwd)" # '''
+''''export PY_PATH=${SCRIPT_PATH}/venv/bin/python                                       # '''
+''''test -f ${PY_PATH} && exec ${PY_PATH} "$0" "$@"                                     # '''
+''''echo "Error: Virtual environment not found. Please run 'one-prepare-venv' command." # '''
+''''exit 255                                                                            # '''
+
+# Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import copy
+import glob
+import itertools
+import ntpath
+import os
+import subprocess
+import sys
+import tempfile
+
+import utils as _utils
+
+# TODO Find better way to suppress trackback on error
+sys.tracebacklimit = 0
+
+
+def _get_backends_list():
+    """
+    [one hierarchy]
+    one
+    ├── backends
+    ├── bin
+    ├── doc
+    ├── include
+    ├── lib
+    └── test
+
+    The list where `one-profile` finds its backends
+    - `bin` folder where `one-profile` exists
+    - `backends` folder
+
+    NOTE If there are backends of the same name in different places,
+     the closer to the top in the list, the higher the priority.
+    """
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    backend_set = set()
+
+    # bin folder
+    files = [f for f in glob.glob(dir_path + '/*-profile')]
+    # backends folder
+    files += [
+        f for f in glob.glob(dir_path + '/../backends/**/*-profile', recursive=True)
+    ]
+    # TODO find backends in `$PATH`
+
+    backends_list = []
+    for cand in files:
+        base = ntpath.basename(cand)
+        if not base in backend_set and os.path.isfile(cand) and os.access(cand, os.X_OK):
+            backend_set.add(base)
+            backends_list.append(cand)
+
+    return backends_list
+
+
+def _get_parser(backends_list):
+    profile_usage = 'one-profile [-h] [-v] [-C CONFIG] [-b BACKEND] [--] [COMMANDS FOR BACKEND]'
+    parser = argparse.ArgumentParser(
+        description='command line tool for profile backend model', usage=profile_usage)
+
+    _utils._add_default_arg(parser)
+
+    # get backend list in the directory
+    backends_name = [ntpath.basename(f) for f in backends_list]
+    if not backends_name:
+        backends_name_message = '(There is no available backend drivers)'
+    else:
+        backends_name_message = '(available backend drivers: ' + ', '.join(
+            backends_name) + ')'
+    backend_help_message = 'backend name to use ' + backends_name_message
+    parser.add_argument('-b', '--backend', type=str, help=backend_help_message)
+
+    return parser
+
+
+def _verify_arg(parser, args):
+    """verify given arguments"""
+    # check if required arguments is given
+    missing = []
+    if not _utils._is_valid_attr(args, 'backend'):
+        missing.append('-b/--backend')
+    if len(missing):
+        parser.error('the following arguments are required: ' + ' '.join(missing))
+
+
+def _parse_arg(parser):
+    profile_args = []
+    backend_args = []
+    unknown_args = []
+    argv = copy.deepcopy(sys.argv)
+    # delete file name
+    del argv[0]
+    # split by '--'
+    args = [list(y) for x, y in itertools.groupby(argv, lambda z: z == '--') if not x]
+    # one-profile has two interfaces
+    # 1. one-profile [-h] [-v] [-C CONFIG] [-b BACKEND] [COMMANDS FOR BACKEND]
+    if len(args) == 1:
+        profile_args = args[0]
+        profile_args, unknown_args = parser.parse_known_args(profile_args)
+    # 2. one-profile [-h] [-v] [-C CONFIG] [-b BACKEND] -- [COMMANDS FOR BACKEND]
+    if len(args) == 2:
+        profile_args = args[0]
+        backend_args = args[1]
+        profile_args = parser.parse_args(profile_args)
+    # print version
+    if len(args) and profile_args.version:
+        _utils._print_version_and_exit(__file__)
+
+    return profile_args, backend_args, unknown_args
+
+
+def main():
+    # get backend list
+    backends_list = _get_backends_list()
+
+    # parse arguments
+    parser = _get_parser(backends_list)
+    args, backend_args, unknown_args = _parse_arg(parser)
+
+    # parse configuration file
+    _utils._parse_cfg(args, 'one-profile')
+
+    # verify arguments
+    _verify_arg(parser, args)
+
+    # make a command to run given backend driver
+    profile_path = None
+    backend_base = getattr(args, 'backend') + '-profile'
+    for cand in backends_list:
+        if ntpath.basename(cand) == backend_base:
+            profile_path = cand
+    if not profile_path:
+        raise FileNotFoundError(backend_base + ' not found')
+    profile_cmd = [profile_path] + backend_args + unknown_args
+    if _utils._is_valid_attr(args, 'command'):
+        profile_cmd += getattr(args, 'command').split()
+
+    # run backend driver
+    with subprocess.Popen(
+            profile_cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+            bufsize=1) as p:
+        for line in p.stdout:
+            sys.stdout.buffer.write(line)
+            sys.stdout.buffer.flush()
+    if p.returncode != 0:
+        sys.exit(p.returncode)
+
+
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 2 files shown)


jyoungyun:
I have to remove the configuration option in `one-profile` tool. I will update PR soon.

compiler/one-cmds/one-profile
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env bash
```
jyoungyun, line 1: This file is copied from `one-codegen`.

compiler/one-cmds/one-profile
```diff
@@ -0,0 +1,168 @@
...
+    argv = copy.deepcopy(sys.argv)
+    # delete file name
+    del argv[0]
+    # split by '--'
+    args = [list(y) for x, y in itertools.groupby(argv, lambda z: z == '--') if not x]
+    # one-profile has two interfaces
+    # 1. one-profile [-h] [-v] [-C CONFIG] [-b BACKEND] [COMMANDS FOR BACKEND]
+    if len(args) == 1:
+        profile_args = args[0]
+        profile_args, unknown_args = parser.parse_known_args(profile_args)
```
mhs4670go, line 117: Actually, the reason that `one-codegen` has two interfaces is for backward compatibility. First interface is followed by second one. If there is specific reason, it would be better to use only second one.

compiler/one-cmds/utils.py
```diff
@@ -81,18 +81,19 @@ class _CONSTANT:
...
 
-    # configuration file
-    parser.add_argument('-C', '--config', type=str, help='run with configuation file')
-    # section name that you want to run in configuration file
-    parser.add_argument('-S', '--section', type=str, help=argparse.SUPPRESS)
+    if support_config is True:
+        # configuration file
+        parser.add_argument('-C', '--config', type=str, help='run with configuation file')
+        # section name that you want to run in configuration file
+        parser.add_argument('-S', '--section', type=str, help=argparse.SUPPRESS)
```
mhs4670go, line 96: I don't know about details of not supporting config file. But isn't it needed?
```yaml
[one-profile]
backend=dummy
command=sample.tvn
```
Hmmm.. on second thought, since profile is not the process of compiling, it seems not be needed.

seanshpark:
(optional) using `[one-profile]` in the title looks like you are working on `compiler/one-profile` module.
but this is actually `one-cmds` command. it would be nice if you rename the title.

compiler/one-cmds/utils.py
```diff
@@ -81,18 +81,19 @@ class _CONSTANT:
 _CONSTANT = _CONSTANT()
 
 
-def _add_default_arg(parser):
+def _add_default_arg(parser, support_config=True):
```
seanshpark, line 84: I've little understanding about this but seems not related with adding `one-profile` tool. (`one-profile` tool may use this feature)
if so, please separate this change to another PR.

compiler/one-cmds/utils.py
glistening: @mhs4670go For limit of time (due to Aug 5), we've decided to existing features only. Configuration for `one-profile` will be the next step after completing skeleton of `onecc` within Aug 5.

compiler/one-cmds/one-profile
```diff
@@ -0,0 +1,168 @@
...
+    ├── include
+    ├── lib
+    └── test
+
+    The list where `one-profile` finds its backends
+    - `bin` folder where `one-profile` exists
+    - `backends` folder
+
+    NOTE If there are backends of the same name in different places,
+     the closer to the top in the list, the higher the priority.
```
seanshpark, line 51: `closer to the top in the list` : what would be the sort order?

compiler/one-cmds/one-profile
```diff
@@ -0,0 +1,168 @@
...
+            backend_set.add(base)
+            backends_list.append(cand)
+
+    return backends_list
+
+
+def _get_parser(backends_list):
+    profile_usage = 'one-profile [-h] [-v] [-b BACKEND] [--] [COMMANDS FOR BACKEND]'
+    parser = argparse.ArgumentParser(
+        description='command line tool for profile backend model', usage=profile_usage)
```
seanshpark, line 77: (optional) `usage` is given as a separate string variable. why not `description` too?

compiler/one-cmds/utils.py
seanshpark: `_add_default_arg` seems to be called from all `one-xxx` tools. if you make this default as `True` all the commands will have `--config` and `--section` options, which I don't see other tools are ready to use this.

I'm my understanding is incorrect, please let me know.

compiler/one-cmds/one-profile
glistening: At first, I wondered how same name of backends can exist. I didn't know `bin` folder is also searched. Then, `NOTE` seeming aims for giving priority between `bin` and `backends/*`. If so, I think we can make the sentence clear.

Also, I guess backends are not searched recursively for both `bin` and `backends`. Am I right?

jyoungyun: > Also, I guess backends are not searched recursively for both bin and backends. Am I right?

Yes, right. `one-profile` searches the *-profile file in `./bin` first, then in `backends/*`. There is no recursive search case.

I can't understand the expression the same name of backends.
I think it means that if there are multiple backends in backends directory, their priority is in alphabetical order.

I will update it.

compiler/one-cmds/one-profile
jyoungyun: The description is a brief description of what the program does and how it works. In help messages, the description is displayed between the command-line usage string and the help messages for the various arguments.

The usage is automatically created by ArgumentParser. But it can't support `[--] [COMMANDS FOR BACKEND]` term so we overrides it. A separating string variable is follow by ArgumentParser's usage rule.

```
# ./one-profile --help
usage: one-profile [-h] [-v] [-b BACKEND] [--] [COMMANDS FOR BACKEND] <-Usage

command line tool for profile backend model <- Description
```

seanshpark: I think you misunderstood my suggestion.
```py
    profile_usage = 'one-profile [-h] [-v] [-b BACKEND] [--] [COMMANDS FOR BACKEND]'
    profile_description = 'command line tool for profile backend model'
    parser = argparse.ArgumentParser(description=profile_description, usage=profile_usage)
```
This is what I ment. `parser = argparse.ArgumentParser(` line is split in two lines which is little hard to read than one line.

compiler/one-cmds/one-profile
jyoungyun: Okay, I want to keep the same usage of codegen. But the current codegen usage has some inconveniences such as having to call `one-codegen -b xxx -- --help` to see the backend's help message.

I will find out if there's a better way.

compiler/one-cmds/one-profile
```diff
@@ -0,0 +1,168 @@
...
+        base = ntpath.basename(cand)
+        if not base in backend_set and os.path.isfile(cand) and os.access(cand, os.X_OK):
+            backend_set.add(base)
+            backends_list.append(cand)
+
+    return backends_list
+
+
+def _get_parser(backends_list):
+    profile_usage = 'one-profile [-h] [-v] [-b BACKEND] [--] [COMMANDS FOR BACKEND]'
```
lemmaa, line 75: I don't know why you omit the `-C` option here. This is a nice feature to refer to the `[one-profile]` section of the `*.cfg` file. In the case of `one-profile`, this function can be more useful because it is used at the last step in the workflow. This is because the profile can be repeatedly executed multiple times for the same compiled model with a simple `one-profile -C model.cfg` command.

compiler/one-cmds/utils.py
jyoungyun: Okay I see. :) I will make it as another PR.

lemmaa: This is probably an attempt to remove the `-C` option. I think it's an unnecessary part, as I commented before.

compiler/one-cmds/one-profile
jyoungyun: Ah! I misunderstood your comments. Now I've got it. :)

compiler/one-cmds/one-profile
jyoungyun: Um, then the profile of backend module should take the cfg file. I will talk about this with @glistening.

jyoungyun:
I updated patch.

The `one-profile` supports both argument parameters and configuration file to execute backend's profile tool.
And `one-profile` file is originally copied from `one-codegen`. But it removes the `--` in argument. (refer to https://github.com/Samsung/ONE/blob/16392623087aae698a7747c5809856cd1b22acab/compiler/one-cmds/one-codegen#L78)

```
# ../bin/one-profile --help
usage: one-profile [-h] [-v] [-C CONFIG] [-b BACKEND] [COMMANDS FOR BACKEND]

command line tool for profile backend model

optional arguments:
  -h, --help            show this help message and exit
  -v, --version         show program's version number and exit
  -C CONFIG, --config CONFIG
                        run with configuation file
  -b BACKEND, --backend BACKEND
                        backend name to use (available backend drivers: dummy,
                        one, tv2)
```

```
# ../bin/one-profile -b tv2 --help
USAGE: /opt/test/build/install/bin/../backends/trinity/bin/tv2-profile [OPTIONS] ...

SUPPORTED OPTIONS:
  -h --help
  --save-cycle-as-env STRING
  --save-chrome-trace STRING
```

seanshpark:
> But it removes the -- in argument

There was an issue for using `--` for backend options. @cgbahk , is this OK for profile?

jyoungyun:
> There was an issue for using -- for backend options. @cgbahk , is this OK for profile?

This tool is newly introduced so I want to remove the meaningless `--`. And if other tools are ready, I want to change the codegen tool in the same way. What do you think about this @cgbahk ?

cgbahk:
History on `--` of `one-codegen` is from internal 6170.

If it is not very hard, why don't we have `--`?
- IMHO, it is natural
- Also, there would be name conflict between option (reason found in https://github.com/Samsung/ONE/issues/6226)

jyoungyun:
Test result:
```
one-profile_001.test SUCCESS
one-profile_002.test SUCCESS
one-profile_003.test SUCCESS
one-profile_004.test SUCCESS
one-profile_neg_001.test SUCCESS
ALL TESTS PASSED!
```

jyoungyun:
In #6226 issue:
> `one-codegen` and backend driver has its own interface. So, if both interface has the same option name, it will be a problem.
> 
> To avoid this situation, let's split them by `--`.

I found a way to avoid the conflict options between `one-codegen` and backend driver with @mhs4670go.
So we do not need to use the `--` option.

> * IMHO, it is natural

And I think it's more natural to have `--` option.

seanshpark:
`gdb` invokes other programs and it uses `--` to send options for that program.
For me, `--` was strange at first but when I got the idea, it was way natural to split options for `gdb` and the debug target.

cgbahk:
> I found a way to avoid the conflict options between `one-codegen` and backend driver with @mhs4670go.

Sounds nice :)

I don't mind about this issue, https://github.com/Samsung/ONE/pull/7360#issuecomment-892501961 was just my thought.
I cannot anticipate any UX issue by now, we can fix them later if we find them later.
(Please merge this PR without my approval)

jyoungyun:
> I cannot anticipate any UX issue by now, we can fix them later when we find them.

To avoid the unexpected issue of codegen, I can support two options in codegen the same as now.
However I want to take only one interface in `one-profile` because it is newly introduced.

jyoungyun:
PTAL

lemmaa:
I think we shouldn't remove `--` for the following reasons.

1. Processing if the same option exists between the main command and subsequent commands. even in the future.
2. UX consistency with other main commands

We have 3 groups of one-cmds.

1. Workflow manager: `one-build`
  In this case, only the `-C` option is processed. This is because there is no function to do by itself other than calling other cmds in order.
2. Working alone: ​​`one-import-*`, `one-optimize`, `one-quantize`
  In this case, it has the `-C` option and its own options. However, there is no problem because it does not call other subsequent commands.
3. Calling other commands: `one-import`, `one-codegen`, `one-profile`
 In this case, along with case 2, the option for the subsequent command to be called must be handled.
 
In the case of 3, the most natural way is to pass all of the received options to the subsequent command except for those that can be processed. I prefer this way. :)

A possible problem with this is when the same options are used between the main command and the subsequent commands. How to solve it?

1. Manage options well to avoid such conflicts between main and subcommands. Special treatment is given to some equally used options. That is, in the case of the `-h` option, the help of the main command is displayed as well as the help of subsequent commands. Same to `-v` or so. It's not easy, but it's possible. And more natural and easy to use for users. Even in this case, I think that number 2 mentioned below must be provided.

2. Provides a way to explicitly separate options between main commands and subcommands. Usually `--` in Linux tools.

So, '--' is necessary. Even if we don't need it right now, we may need it in a subsequent command we'll make in the future. :)

----

Extra explanation for `-C` .

In the original design, the options recorded in the `*.cfg` file received through `-C` had lower priority than the options entered directly on the command-line. That is, if the same option exists in the `*.cfg` file and command-line, the command-line option takes precedence. That is, the command-line option overrides the `*.cfg` option.

jyoungyun:
I simply copied the `one-codegen` to `one-profile`.

```
# ./test/runtestall.sh
one-profile_001.test SUCCESS
one-profile_002.test SUCCESS
one-profile_003.test SUCCESS
one-profile_004.test SUCCESS
one-profile_neg_001.test SUCCESS
ALL TESTS PASSED!
```

PTAL.

mhs4670go (approved):
LGTM

glistening (approved):
LGTM

lemmaa (approved):
LGTM

jyoungyun:
@seanshpark Please merge this PR. Thank you.

PR #7365, opened 2021-08-03
[luci] Disable annotate NHWC Ops

seanshpark:
This will disable recent annotate NHWC Ops that seems to have a problem
with fusing Reshape-Add-Mul-Pad-Reshape pattern.


seanshpark:
Related: #7321

jinevening:
Could you explain what is the problem?

If we remove the code, the pattern (pre-Transose + [chain of Ops] + post-Transpose) will not be correctly converted.

seanshpark:
I'll prepare a test recipe (model) for this case...

seanshpark:
I think this can be closed as issue itself is resovled.

PR #7367, opened 2021-08-03
DRAFT add regresstion test for Conv-Add-Mul

seanshpark:
regression test for Conv-Add-Mul model multiple transformations.


seanshpark:
CC @jinevening

seanshpark:
When the fix lands, I'll remove temporary fix `FIX_ME_ANNOTATE_NHWC_OPS` and check this passes.

seanshpark:
Close as all landed

PR #7370, opened 2021-08-04
[common-artifacts] Exclude Part_While

seanshpark:
This will revise to exclude Part_While_000 and Part_While_001 while code
lands and will be removed afterwards.


seanshpark:
For #7280, from draft #7358

mhs4670go (approved):
LGTM

PR #7372, opened 2021-08-04
DRAFT: Extend Convert NCHW to NHWC

jinevening:
On going draft to extend ConvertNCHWToNHWC pass.

---
Related to: https://github.com/Samsung/ONE/pull/7365

### DIFF
--- a/compiler/circle2circle-dredd-recipe-test/test.lst
+++ b/compiler/circle2circle-dredd-recipe-test/test.lst
@@ -46,3 +46,16 @@ Add(StridedSlice_003 PASS substitute_strided_slice_to_reshape)
 ## CIRCLE RECIPE
 
 Add(CircleBatchMatMul_000)
+
+# REGRESSION test
+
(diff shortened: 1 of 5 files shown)


jinevening:
This draft has the following changes.

1. Run `ConvertNCHWToNHWCPass` before other passes (This was not the direct reason of the recent failure, but I'd like to make this change to avoid future problem)
2. Consider pre/post-Reshape as well as pre/post-Transpose (`onnx-tf` sometimes creates Reshape instead of Transpose)
3. Support more complex patterns when annotating NHWC operators (please see the comment in `ConvertNCHWToNHWCPass::run`).

I checked the target model was compiled well. And I checked the regression test passed (https://github.com/Samsung/ONE/pull/7367).

@seanshpark I'll start to upload separate PRs if you confirm that this draft is ok.

seanshpark:
Looks OK :)

jinevening:
Close as all done.

PR #7374, opened 2021-08-04
[luci] Skip WhileOut in validate_shape_dtype

seanshpark:
This will skip WhileOut Op in validate_shape_dtype method.

### DIFF
--- a/compiler/luci/service/src/Validate.cpp
+++ b/compiler/luci/service/src/Validate.cpp
@@ -102,6 +102,9 @@ bool validate_shape_dtype(loco::Graph *g)
     // Shape and dtype validation for CiecleOutputExclude is not needed
     if (dynamic_cast<luci::CircleOutputExclude *>(circle_node))
       continue;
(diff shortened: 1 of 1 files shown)


seanshpark:
For #7280, from draft #7358

llFreetimell:
IMHO, excluding validation for `CircleWhileOut` seems not good :(
The purpose of `validate_shape_dtype` is validating whether all nodes have correct shape and dtype.
Only one exception, because `CircleOutputExclude` is for excluding output, validation is not needed for the node.
However, `CircleWhileOut` is not those kind of node, it should be always valid.

Because of these reason, `validate(graph)` function is only called when the graph is ready for validation.
So if the graph is not yet ready for the validation, not calling `validate` function seems reasonable instead of excluding specific nodes.

seanshpark:
@llFreetimell , thanks for the comment, I'll check if it is OK with shape/dtype inference before validation.

seanshpark:
It's OK :) This change is not needed.

PR #7375, opened 2021-08-04
[res] Introduce Part_While recipes

seanshpark:
This will introduce Part_While_000 and Part_While_001 recipes.

### DIFF
--- a/res/TensorFlowLiteRecipes/Part_While_000/test.readme
+++ b/res/TensorFlowLiteRecipes/Part_While_000/test.readme
@@ -1,0 +1,4 @@
+test.readme of Part_While_000
+
+MAXIMUM and MINIMUM Op exist to make random input to 0
+to make this model loop from 0 to 10.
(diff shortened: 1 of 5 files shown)


seanshpark:
For #7280, from draft #7358

res/TensorFlowLiteRecipes/Part_While_000/test.readme
```diff
@@ -0,0 +1,4 @@
+test.readme of Part_While_000
+
+MAXIMUM and MINIMUM Op exist to make random input to 0
+to make this model loop from 0 to 10.
```
mhs4670go, line 4: I'm not sure it is intended?
```suggestion
MAXIMUM and MINIMUM Op exist to make random input
to make this model loop from 0 to 10.
```

seanshpark: > I'm not sure it is intended?

Yup :) Very large random value as input +1 will not change so, WHILE will loop in infinite...
Limitation is to prevent that.

mhs4670go (approved):
LGTM

PR #7381, opened 2021-08-04
[circle-partitioner] Call shape/dtype inference

seanshpark:
This will revise circle-partitioner to call shape/dtype inference
before calling graph validation.


seanshpark:
For #7280, from draft #7358

mhs4670go (approved):
LGTM

PR #7384, opened 2021-08-04
[luci] Enable partition connect for While

seanshpark:
This will enable partition connect for CircleWhile and CircleWhileOut nodes.


seanshpark:
For #7280, from draft #7358

mhs4670go (approved):
LGTM

PR #7387, opened 2021-08-04
[circle-quantized] Improve quant verification for Cast and Reshape

binarman:
This commit adds Cast Quantization checks and improves Reshape.

Related issue: #7229

### DIFF
--- a/compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
+++ b/compiler/luci/pass/src/VerifyQuantizedNodeChannelWiseGranularity.h
@@ -267,8 +267,12 @@ private:
 
   bool visit(const luci::CircleReshape *node)
   {
-    RETURN_FALSE_UNLESS(is_lwq(node))
-    RETURN_FALSE_UNLESS(is_lwq(node->tensor()));
+    auto input = loco::must_cast<const luci::CircleNode *>(node->tensor());
+    bool input_quantized = input->quantparam() != nullptr;
+    bool node_quantized = node->quantparam() != nullptr;
+    RETURN_FALSE_UNLESS(input_quantized == node_quantized);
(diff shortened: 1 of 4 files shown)


binarman:
See https://github.com/Samsung/ONE/pull/7388#issuecomment-892968227 for target graph.

seanshpark:
@binarman , @jinevening , have we decided how to quantize CAST Op?

jinevening:
> have we decided how to quantize CAST Op?

I think this is the first try to quantize CAST.

@binarman It seems that your approach is to quantize CAST only when the output is fp32 (please correct if I'm wrong). It would be better to add the comment somewhere.

I think the properties (in_data_type, out_data_type) of CAST should be updated when quantized.

And could you check if this change does not cause any problem in our in-house models?

binarman:
@jinevening 
> It seems that your approach is to quantize CAST only when the output is fp32 (please correct if I'm wrong). It would be better to add the comment somewhere.

Yes, I want to quantize only <some type> -> fp32 case and leave cases like fp32 -> <int32/64> to resolve in backends, because at this point I I do not have information about supported types there. (some may have full support of integer operations, some may not)
This approach mimics approach used for Argmax.

binarman:
@jinevening
> I think the properties (in_data_type, out_data_type) of CAST should be updated when quantized.

Added #7405 with changes in quantization pass. In this PR I've added appropriate checks. (tested them manually on small model with Cast operations)

jinevening (approved):
LGTM

binarman:
@jinevening 
I forgot to add a dependency(and nomerge label) after review fixes.
Merged this PR bit early, #7405 should be first =(

After review fixes Cast checks rely on cast attributes that are fixed only in #7405.

P.s. I've checked our benchmark models and found that this PR + #7405 has no impact, but this PR alone drops three models (not sure about importance of these models, I've just tests all tflite models I found :slightly_smiling_face: )

PR #7388, opened 2021-08-04
[record-mimax] Add more exceptions for quantization

binarman:
This commit adds type check exceptions for Reshape and Cast operations.

Related issue: #7229

### DIFF
--- a/compiler/record-minmax/src/MinMaxObserver.cpp
+++ b/compiler/record-minmax/src/MinMaxObserver.cpp
@@ -55,15 +55,17 @@ void MinMaxObserver::postTensorWrite(const luci::CircleNode *node,
   // Only support recording of float32 values
   if (tensor->element_type() != DataType::FLOAT32)
   {
-    // Argmax and Cast are exceptions we can process in backends
-    // Output of arg_max is the index of the largest value across axes of a tensor
-    // Cast converts float to int
-    // these should not be quantized
+    // Exceptions that should be processed in backends
     switch (node->opcode())
     {
       case luci::CircleOpcode::ARG_MAX:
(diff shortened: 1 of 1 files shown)


binarman:
This is needed to quantize graph like this:
image

compiler/record-minmax/src/MinMaxObserver.cpp
```diff
@@ -61,7 +54,21 @@ void MinMaxObserver::postTensorWrite(const luci::CircleNode *node,
 
   // Only support recording of float32 values
   if (tensor->element_type() != DataType::FLOAT32)
-    throw std::runtime_error("Tensor's data type is not float");
+  {
+    // Argmax and Cast are exceptions we can process in backends
+    // Output of arg_max is the index of the largest value across axes of a tensor
+    // Cast converts float to int
```
jinevening, line 60: Cast is a more general operator. I think the comment can be revised (or removed?).

In fact, there are three Ops (ArgMax, Cast, Reshape) for now, and I think more Ops will be added to this list. I think explanation for each operator is not necessary.

seanshpark: Comments seems good information to me :)
Maybe move to below each `case luci::CircleOpCode::XXX` line ?

binarman: > Cast is a more general operator. I think the comment can be revised (or removed?).

You are right, when I wrote this I had just "int->float" and "float->int" cases in mind =_=

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

Issue #7385, opened 2021-08-04
[luci-interpreter] Extract to function for While kernel


seanshpark:
By @binarman , from https://github.com/Samsung/ONE/pull/7371#pullrequestreview-722111592

> I think it would be better to move this code to function.
> Tensors data move is used 4 times in this kernel

seanshpark:
done

PR #7391, opened 2021-08-05
[luci-interpreter] Revise While kernel with copy method

seanshpark:
This will extract tensor copy to a method a remove similar repeating codes.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/While.cpp
+++ b/compiler/luci-interpreter/src/kernels/While.cpp
@@ -25,6 +25,32 @@ namespace luci_interpreter
 namespace kernels
 {
 
+namespace
+{
+
+void copy(const std::vector<const Tensor *> &src, const std::vector<Tensor *> &dst)
+{
+  for (size_t i = 0; i < src.size(); ++i)
+  {
+    LUCI_INTERPRETER_CHECK(dst[i]->element_type() == src[i]->element_type());
+    dst[i]->resize(src[i]->shape());
+
+    const int32_t num_elements = src[i]->shape().num_elements();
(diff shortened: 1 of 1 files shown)


seanshpark:
For #7385

compiler/luci-interpreter/src/kernels/While.cpp
```diff
@@ -52,16 +73,7 @@ void While::execute() const
   const auto &cond_inputs = _cond_graph->getInputTensors();
   const auto &cond_outputs = _cond_graph->getOutputTensors();
 
-  // Copy initial kernel inputs to condition graph inputs.
```
seanshpark, line 55: Removed comments as the function call itself express what it does

compiler/luci-interpreter/src/kernels/While.cpp
```diff
@@ -44,6 +44,27 @@ void While::configure()
   LUCI_INTERPRETER_CHECK(cond_outputs[0]->element_type() == DataType::BOOL);
 }
 
+void While::copy(const std::vector<const Tensor *> &src, const std::vector<Tensor *> &dst) const
```
binarman, line 47: Optional:
These methods could be simple static functions, they do not use any fields/methods of While.
In that case there are no need to change `While.h`.

binarman (dismissed):
LGTM

compiler/luci-interpreter/src/kernels/While.cpp
seanshpark: I'll follow your suggestion :)

binarman (approved):

PR #7393, opened 2021-08-05
[luci-value-test] Enable Part_While tests

seanshpark:
This will enable value testing of Part_While models for WHILE with float32.


mhs4670go (approved):
LGTM

PR #7394, opened 2021-08-05
DRAFT add regression test for ONNX models

seanshpark:
regression test for Conv-BN-Relu6 model with multiple transformations from ONNX.


seanshpark:
All landed.

PR #7395, opened 2021-08-05
PoC: nnpackage specification for pre-built model binary 

hseok-oh:
This is PoC for nnpackage including pre-built binary model for specific backend.

Example
- Use pre-compiled dynmaic library (.so)
- Add info json file: model type, I/O shape & type
  - Loader will use this to make fake graph (one fake op with I/Os)
  - Backend will use this to load and execute binary


hseok-oh:
cc/ @chunseoklee

nnpackage/examples/README.md
```diff
@@ -1,5 +1,12 @@
 # NNPackage example
 
+## Package version 1.2.0
+
+### one_op_in_bin
+
+- Model file: pre-compiled dynamic library (TBD)
+- Only one `ADD` operation (TBD)
```
chunseoklee, line 8: Since it is pre-compiled binary, we do not need to specify binary's internal operations ? IMHO, this binary will be loaded into Graph of single operation.

hseok-oh: Right. Internal operation is not important. It's just explain example model.

hseok-oh:
```suggestion
- Compiled model has only one `ADD` operation (TBD)
```

nnpackage/spec/10_packaging_and_manifest.md
```diff
@@ -79,10 +81,11 @@ The first element from the array will be the default model to be executed.
...
 
-| name   | description            |
-|--------|------------------------|
-| tflite | tensorflow lite schema |
-| circle | nnpackage schema       |
+| name   | description              |
+|--------|--------------------------|
+| tflite | tensorflow lite schema   |
+| circle | nnpackage schema         |
+| bin    | backend dependent binary |
```
glistening, line 88: Maybe it is not yet for review. However, please let me leave comment since I will forget to do later. : )
`tflite` and `circle` are also binary in term that it is not human readable. Thus, I prefer to use more specific names. Maybe `prebuilt` (as you wrote in title) or `backend-dependent` (as you described in table) would be better.

PR #7399, opened 2021-08-05
[circle-partitioner] Enable value test for Part_While

seanshpark:
This will enable circle-partitioner value test for Part_While_000 and
Part_While_001 models.


seanshpark:
For #7280, from draft #7358

mhs4670go (approved):
LGTM

PR #7404, opened 2021-08-05
[res] Regression test recipe with Conv+BN

seanshpark:
This will add regression test recipe with Conv+BN.


seanshpark:
From draft #7367

jinevening (approved):
LGTM

PR #7405, opened 2021-08-05
[luci] Fix Cast attributes after quantization

binarman:
This commit adjusts Cast attributes after quantization.

Related issue: #7229

Requested by https://github.com/Samsung/ONE/pull/7387#issuecomment-893116766

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -734,6 +734,9 @@ struct QuantizeActivation final : public luci::CircleNodeMutableVisitor<bool>
       {
         auto *cast = loco::must_cast<luci::CircleCast *>(circle_node);
         auto *cast_input = loco::must_cast<luci::CircleNode *>(cast->x());
+
+        // make sure that cast_input is already quantized
+        assert(cast_input->dtype() != loco::DataType::FLOAT32);
         cast->in_data_type(cast_input->dtype());
         cast->out_data_type(cast->dtype());
       }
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -729,6 +729,14 @@ struct QuantizeActivation final : public luci::CircleNodeMutableVisitor<bool>
         circle_node->quantparam()->scale.push_back(scaling_factor);
         circle_node->quantparam()->zerop.push_back(zp);
       }
+      // Fix special attributes
```
binarman, line 732: @jinevening 
I have a question about design of this Visitor:
Why do we quantize node inputs instead of node itself?

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -729,6 +729,14 @@ struct QuantizeActivation final : public luci::CircleNodeMutableVisitor<bool>
...
+      if (circle_node->opcode() == luci::CircleOpcode::CAST)
+      {
+        auto *cast = loco::must_cast<luci::CircleCast *>(circle_node);
+        auto *cast_input = loco::must_cast<luci::CircleNode *>(cast->x());
+        cast->in_data_type(cast_input->dtype());
```
jinevening, line 737: For cast (fp32->fp32), if `cast_input` is quantized later than `cast`, `cast_input->dtype()` may be fp32, which is invalid.

Cast (fp32->fp32) seems unnatural, so it would be good to add an assert `assert(cast_input->dtype() != float32)`.

binarman: > if cast_input is quantized later than cast

Currently this code is called in loop over topologically sorted nodes, so described situation should be impossible. (I assume that sorting algorithm is correct).

If this visitor should be agnostic to the order, I'll move this code to separate loop and leave a comment that it should be always sorted.

P.s. I'll add assert for fp32->fp32. It is indeed suspicious case.

jinevening (approved):
LGTM

PR #7414, opened 2021-08-06
[common-artifacts] Enable Part_While

seanshpark:
This will enable generation of Part_While in common-artifacts.


mhs4670go (approved):
LGTM

Issue #7341, opened 2021-07-28
[one-cmds] Print what driver causes the error when it is invoked indirectly


glistening:
When error occurs during user's execution of `one-build`, it does not show what driver caused the error.

For example,

```
$ ./one-build -C one-build.template.cfg
OSError: File '/path/to/inception_v3.pb' does not exist.
```

The error occurred during `one-import-tf`. However, the user can't know from error message.

```
$ ./one-build -C one-build.template.cfg
one-import-tf: OSError: File '/path/to/inception_v3.pb' does not exist.
```
would be better.

glistening:
@jyoungyun Do we need to modify `one-build` ? As I heard, we will deprecate `one-build` with `onecc -C`.

jyoungyun:
No, I don't seem to need it. But this issue seems to be valid with `onecc` too.

glistening:
> No, I don't seem to need it. But this issue seems to be valid with `onecc` too.

Yes, I consider this issue is mainly about `onecc`, not `one-build` (to be deprecated).

PR #7418, opened 2021-08-06
[onecc] Print driver name in error message

glistening:
When user run `onecc -C`, several steps are done by configuration.
In this case, it would be better to print driver name that the error
occurred in.

### DIFF
--- a/compiler/one-cmds/onecc
+++ b/compiler/one-cmds/onecc
@@ -52,7 +52,7 @@ def _call_driver(driver_name, options):
     with subprocess.Popen(
             cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, bufsize=1) as p:
         for line in p.stdout:
-            sys.stdout.buffer.write(line)
+            sys.stdout.buffer.write(f"{driver_name}: ".encode() + line)
             sys.stdout.buffer.flush()
     if p.returncode != 0:
         sys.exit(p.returncode)


glistening:
Related: #7341 

#### BEFORE
```
$ bin/onecc -C bin/onecc.template.cfg 
OSError: File '/path/to/inception_v3.pb' does not exist.
```
#### AFTER
```
$ bin/onecc -C bin/onecc.template.cfg 
one-import-tf: OSError: File '/path/to/inception_v3.pb' does not exist.
```

jinevening (approved):
LGTM

jyoungyun (approved):
LGTM

lemmaa:
Wouldn't it be nice to let `one-import-tf` display its own name?

compiler/one-cmds/onecc
```diff
@@ -52,7 +52,7 @@ def _call_driver(driver_name, options):
     with subprocess.Popen(
             cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, bufsize=1) as p:
         for line in p.stdout:
-            sys.stdout.buffer.write(line)
+            sys.stdout.buffer.write(f"{driver_name}: ".encode() + line)
```
lemmaa, line 55: I have two questions.

Although irrelevant to this PR, is there any special reason for connecting stderr to stdout on line 52?

Regarding this PR, isn't it correct that the error message is displayed in the module where the error message occurred?

mhs4670go: > Although irrelevant to this PR, is there any special reason for connecting stderr to stdout on line 52?

That is a recommend way to implement `subprocess` in [document](https://docs.python.org/3/library/subprocess.html).

> If capture_output is true, stdout and stderr will be captured. When used, the internal Popen object is automatically created with stdout=PIPE and stderr=PIPE. The stdout and stderr arguments may not be supplied at the same time as capture_output. If you wish to capture and combine both streams into one, use stdout=PIPE and stderr=STDOUT instead of capture_output.

> isn't it correct that the error message is displayed in the module where the error message occurred?

I think it is doing as you said. Current `one-cmds` tools spit out their outputs. That outputs should be piped(`stdout=subprocess.PIPE`) to be displayed.

glistening: @lemmaa Thank you for review. For 1st question, @jyoungyun got answer no special reason..., and she will follow up this issue. For 2nd question, I think it depends on. Anyway, I will continue to create PR or issue to help users identify where the error comes.

lemmaa: > ... If you wish to capture and combine both streams into one ...

@mhs4670go , Is there any reason for it? If not, it is natural to bypass stderr of called process to stderr of caller process. 

> I think it is doing as you said. Current `one-cmds` tools spit out their outputs. That outputs should be piped(`stdout=subprocess.PIPE`) to be displayed.

My point in here is `onecc` prepends `driver-name` to the error message of called process. :)

mhs4670go: @lemmaa  
> Is there any reason for it?

Both way looks okay to me. The reason why I chose the way I did it is because I wanted to keep the log sequences.

Say, the error logs are like this sequence.

```python3
print("First", file=sys.stderr)
print("Second", file=sys.stdout)
print("third", file=sys.stderr)
```

Then, current implementation prints like this.
```
First
Second
Third
```

But, if I split the pipe(`stdout=subprocess.PIPE`, `stderr=subprocess.PIPE`), I'm going to have to catch them separately. And, it will print like this.
```
Second # print stdout first
First
Thrid
```

Issue #7422, opened 2021-08-06
[one-cmds] Each module displays its name for its messages


glistening:
From #7418, `onecc` prints the driver name of callee (`one-xxx`).

However, @lemmaa suggested

> Wouldn't it be nice to let `one-import-tf` display its own name?

from https://github.com/Samsung/ONE/pull/7418#issuecomment-894039739

Maybe, it is not applicable for all modules since several modules (e.g. `tflite_convert`, `onnx_tf` and some future tools) are out of our control.

glistening:
@lemmaa As I understand, you want each module to display its own name for only (self) message. 

That is, suppose that user invoke `onecc -C ...`, then

```
onecc -> one-import-tf -> tf2tfliteV2 -> tflite_converter
```

`one-import-tf` will put `one-import-tf: ` for its own message, (not for child's message).
For message from `tf2tfliteV2`, it will have `tf2tfliteV2: `.
For message from `tflite_converter`, it will have no prefix.

You want to identify the message generator as much as possible by letting each module to add prefix.

Please correct me if my understanding is wrong.

glistening:
@lemmaa I am modifying sources as you've requested. It requires several changes around several sources from several repositories.

Please find #7423, which is a bit of changes.

I am trying `circle-quantizer`. However, I am in trouble because `circle-quantizer`'s error message are generated all around the source. Several runtime_exceptions are thrown from several places. Some of them (e.g, `foder::FileLoader`) are caught in `SafeMain`, unfortunately, which is library used throughout a lot of tools within this repo.

from `safemain/SafeMain.cpp`

```c++
int main(int argc, char **argv)
{
  try
  {
    return entry(argc, argv);
  }
  catch (const std::exception &e)
  {
    std::cerr << "ERROR: " << e.what() << std::endl;
  }

  return 255;
}
```

Changing existing callee (existing modules)'s behavior requires much labor and hard to maintain, and may break something.

@lemmaa As I understand, you already know this limitation and want to apply this change for some applicable ones (perhaps python implementation) only. Could you confirm me to whether continue this change or not?

lemmaa:
Right

Under our rule, we can guess that unnamed logs are theirs. Of course, if such a relationship is repeated over 2 levels, it will be difficult to distinguish. But this is also true in other cases.

In that case, if it is absolutely necessary, you can make the module caller print it out.

The subtle differences between the two are:

1. In the case that the error occurred module displays the name of the error module by default.
[...]

PR #7423, opened 2021-08-06
[one-cmds,tf2tfliteV2] Print module name for error messages

glistening:
Instead of printing driver in `onecc`, each module print its name by its own.

It is a draft for #7422

### DIFF
--- a/compiler/circle-quantizer/src/CircleQuantizer.cpp
+++ b/compiler/circle-quantizer/src/CircleQuantizer.cpp
@@ -111,8 +111,8 @@ int entry(int argc, char **argv)
   }
   catch (const std::runtime_error &err)
   {
-    std::cout << err.what() << std::endl;
-    std::cout << arser;
+    std::cerr << "circle-quantizer: " << err.what() << std::endl;
+    std::cerr << arser;
     return 255;
   }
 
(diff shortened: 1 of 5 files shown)


glistening:
Now, `one-import-tf` and `tf2tfliteV2.py` will print its name even when it is invoked directly.

### BEFORE
```
$ bin/one-import-tf -C bin/one-import-tf.cfg 
OSError: File './../compiler/one-cmds/tests/inception_v3.pb' does not exist.
```

### AFTER
```
$ bin/one-import-tf -C bin/one-import-tf.cfg 
tf2tfliteV2.py: OSError: File './../compiler/one-cmds/tests/inception_v3.pb' does not exist.
```

lemmaa (approved):
LGTM. Thanks!

compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -180,7 +180,16 @@ int entry(int argc, char **argv)
...
+  std::vector<char> model_data;
+  try
+  {
+    model_data = file_loader.load();
+  }
+  catch (const std::runtime_error &err)
+  {
+    std::cerr << "circle-quantizer: ERROR: " << err.what() << std::endl;
+    return 255;
+  }
```
glistening, line 192: Let's catch exception in app itself as much as possible to customize error message.

jyoungyun (approved):
LGTM

glistening:
Thank you for reviews. I updated this PR in #7427, and split it into several ones.

PR #7427, opened 2021-08-09
[onexxx,tf2tfliteV2] Print module name for error messages (hybrid)

glistening:
Most user tools in Python (`onecc` and `one-xxx`) will print its name by
its own. Nested modules from external or in C++ get the name printed by
Python caller.

Another draft for #7422

### DIFF
--- a/compiler/tf2tfliteV2/tf2tfliteV2.py
+++ b/compiler/tf2tfliteV2/tf2tfliteV2.py
@@ -261,4 +261,9 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except Exception as e:
+        prog_name = os.path.basename(__file__)
+        print(f"{prog_name}: {type(e).__name__}: " + str(e))
+        sys.exit(-1)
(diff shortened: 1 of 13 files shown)


compiler/tf2tfliteV2/tf2tfliteV2.py
```diff
@@ -261,4 +261,9 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except Exception as e:
+        prog_name = os.path.basename(__file__)
+        print(f"{prog_name}: {type(e).__name__}: " + str(e))
+        sys.exit(-1)
```
mhs4670go, line 269: How about exit with 255 like others?

mhs4670go: Or just exit with its code using `e.code`.

mhs4670go: Sorry =ㅁ= Seems there is no `e.code`.

PR #7435, opened 2021-08-10
[onert][gpu_cl] Add Resize operation. 

Hyunjun85:
1. Supports Resize operation.
2. Add testcases for it.

Signed-off-by: H. Kim hyunjun2.kim@samsung.com
Signed-off-by: hansolu hansolu.lee@samsung.com


runtime/onert/backend/gpu_cl/ClFunction.h
```diff
@@ -18,6 +18,8 @@
 #define __ONERT_GPU_CL_CL_FUNCTION_H__
```
juitem, line 18: on other header files on runtime backend includes directory name on define.
how about apply it on this file too?

Hyunjun85: Modified.

PR #7386, opened 2021-08-04
[infra] Add man page for onecc

jyoungyun:
This commit adds man page for onecc tool.

Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #7442, opened 2021-08-10
[tflite2circle,circle2circle] Write errors to std::cerr

glistening:
Errors are supposed to be written to stderr, not to stdout.

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -374,7 +374,7 @@ int entry(int argc, char **argv)
   }
   catch (const std::runtime_error &err)
   {
-    std::cout << err.what() << std::endl;
+    std::cerr << err.what() << std::endl;
     std::cout << arser;
     return 255;
   }
--- a/compiler/tflite2circle/driver/Driver.cpp
+++ b/compiler/tflite2circle/driver/Driver.cpp
@@ -55,7 +55,7 @@ int entry(int argc, char **argv)
   }
   catch (const std::runtime_error &err)
   {
-    std::cout << err.what() << std::endl;
+    std::cerr << err.what() << std::endl;
     std::cout << arser;
     return 255;
   }


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -374,7 +374,7 @@ int entry(int argc, char **argv)
   }
   catch (const std::runtime_error &err)
   {
-    std::cout << err.what() << std::endl;
+    std::cerr << err.what() << std::endl;
     std::cout << arser;
```
glistening, line 378: @jinevening, @jyoungyun 
```c++
std::cout << arser;
```
corresponds to help message after error occurred. Should it be printed to cout? or cerr?

If we consider the usage as a part of error message, the second `cout` should be `cerr` also.

glistening:
@nnfw-bot test nncc-release

compiler/circle2circle/src/Circle2Circle.cpp
jyoungyun: When I tested in my environment, docker put the additional message to the stderr.

```
$ docker abc
docker: 'abc' is not a docker command.
See 'docker --help'
$ docker abc 2>./abc
# no message
```

refer to https://www.middlewareinventory.com/blog/linux-stdout-stderr-bash/

compiler/tflite2circle/driver/Driver.cpp
```diff
@@ -55,7 +55,7 @@ int entry(int argc, char **argv)
   }
   catch (const std::runtime_error &err)
   {
-    std::cout << err.what() << std::endl;
+    std::cerr << err.what() << std::endl;
```
seanshpark, line 58: Please separate `compiler/*` modules in different PRs

glistening: @seanshpark Yes, I know. It seems implicit rule in this repo. But may I ask why? For who? For what? I personally does not agree the rule. If reviews are pending, and if the changes are conceptually same thing, I believe putting in one PR makes sense.

compiler/circle2circle/src/Circle2Circle.cpp
glistening: @jyoungyun Thank you for providing `git` and `docker` behaviors. I will update and split this PR.

compiler/tflite2circle/driver/Driver.cpp
seanshpark: >  But may I ask why? For who? For what? I personally does not agree the rule

This was made by @parjong from the first place when this repo was made(?) to co-work with SRR members not to mixup codes with different projects (SR leading, SRR leading) and as a general rule, make PRs easy to review. This is my understanding what I know. But you may ask to @parjong.

If you don't want to follow this rule, then take the maintainer authority and define your own.

glistening: > > But may I ask why? For who? For what? I personally does not agree the rule
> 
> This was made by @parjong from the first place when this repo was made(?) to co-work with SRR members not to mixup codes with different projects (SR leading, SRR leading) and as a general rule, make PRs easy to review. This is my understanding what I know. But you may ask to @parjong.
> 
> If you don't want to follow this rule, then take the maintainer authority and define your own.

Everyone needs not to be maintainer. Someone may give his/her opinion. : )

seanshpark: > Then, I don't think this case has same problem or issues.

The rule was and still applied without exceptions.
If you want to break this rule, you can go ahead as a maintainer.
As I followed the basic rule from the start, breaking this rule means breaking the decision what build up till now.
For that I cannot go on as a maintainer thus I will leave ONE development.

CC @lemmaa

glistening: >> Then, I don't think this case has same problem or issues.
>
> The rule was and still applied without exceptions.

I am saying why this case is different that this rule is valid.

- There is no one who working for `one-cmds` except for UX T/F members (@jyoungyun and @jinevening)
- No one from SRR work for this task, and 
- Even you said 5 days ago,

> And as I am not following any activity in current UX related task, so please do not add me as a reviewer,
I don't know the backgrounds thus I can not decide go or no-go.
from https://github.com/Samsung/ONE/pull/7386#issuecomment-894016388

This rule works as a blocker against progress to make things done in given schedule.

> If you want to break this rule, you can go ahead as a maintainer.

I don't want and can't break rule because I am not maintainer.
I am just suggesting my opinion, which works well for onert code, which I had mainly worked for.

compiler/circle2circle/src/Circle2Circle.cpp
jinevening: I prefer the latter. The usage message without prefix looks better in the viewpoint of users.

jyoungyun: I think it would be better if all logs put to stderr, but usage prints without the module name.
But if the implement becomes more complicated, the second example (std::cout << arser) looks better.

compiler/tflite2circle/driver/Driver.cpp
seanshpark: > I don't want and can't break rule because I am not maintainer.
I am just suggesting my opinion, which works well for onert code, which I had mainly worked for.

These two conflict.
You don't want to break the rules cause you are not maintainer.
You want to break the rule by means of suggestion through me, which I cannot.
So I suggestest you to become a maintainer and build up your own rule with the `compile/*` but you don't want to.


> There is no one who working for one-cmds except for UX T/F members (@jyoungyun and @jinevening)
No one from SRR work for this task, and
Even you said 5 days ago,

These are true. But I don't understand why these can be a reason to mix `compiler/*` modules is OK.
For the worst case, I could say mixing modules makes review hard and still request to split them.

> This rule works as a blocker against progress to make things done in given schedule. (for trivial PRs like this).

You can split this PR into two and that will comply the rule and I don't know how that is a blocker.
The problem I see here is that review itself is SLOW,

compiler/circle2circle/src/Circle2Circle.cpp
glistening: @jyoungyun Please give your opinion. If you would like to follow `docker` way, I will modify this PR.
But I think it is just difference of preference and my previous way may be more similar to `docker`  in perspective of user.

Here are  further example runs from `docker` and `git`

```
$ docker version 142
"docker version" accepts no arguments.
See 'docker version --help'.

Usage:  docker version [OPTIONS]

Show the Docker version information
```

```
 $ docker run alkdsajflkdsajf
Unable to find image 'alkdsajflkdsajf:latest' locally
docker: Error response from daemon: pull access denied for alkdsajflkdsajf, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
```

```
$ git hello world
git: 'hello' is not a git command. See 'git --help'.

The most similar command is
        shell

$ git log akljladksjfasd
fatal: ambiguous argument 'akljladksjfasd': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

They usually don't prepend its name to all lines. Rather, it prepends only for 1st or selected lines only.

compiler/tflite2circle/driver/Driver.cpp
lemmaa: Personally, I don't like that rule. (I'm not a robot, so I prefer to operate with the appropriate flexibility at a level that can convince each other. ^_^)

However, it is also true that such rules have been maintained here. If several people feel uncomfortable, I think it is necessary to discuss and revise it together. But until then, it's also correct to follow the old rules. So, perhaps it is better to leave this as a `Draft PR`, and after sufficiently discussing and modifying it, split it into several PRs and reflect it.

jyoungyun (approved):
LGTM

glistening:
It was split to #7458 and #7459.

PR #7443, opened 2021-08-10
[luci][loco] Resolve return type conflict in arity(void) methods

SlavikMIPT:
This commit resolves return type conflicts in arity(void) methods which trigger a compiler error when using arm-none-eabi-gcc.

### DIFF
--- a/compiler/loco/include/loco/IR/NodeMixins.h
+++ b/compiler/loco/include/loco/IR/NodeMixins.h
@@ -83,7 +83,7 @@ private:
   std::vector<Dimension> _dims;
 };
 
-template <unsigned N> struct FixedArity
+template <uint32_t N> struct FixedArity
 {
   template <typename Base> class Mixin : public virtual Base
   {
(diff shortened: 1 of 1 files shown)


compiler/luci/lang/include/luci/IR/CircleNodeMixins.h
```diff
@@ -67,7 +67,7 @@ template <> class CircleNodeMixin<CircleNodeTrait::Bias>
  *
  * TODO Deprecated this class, and use loco::FixedArity instead
  */
-template <unsigned N, typename Base> class FixedArityNode : public Base
+template <uint32_t N, typename Base> class FixedArityNode : public Base
```
seanshpark, line 70: We have a rule not to mix files in different `compiler/*`
But I don't know splitting to separate PRs will work with compilation.

SlavikMIPT: > We have a rule not to mix files in different `compiler/*`
> But I don't know splitting to separate PRs will work with compilation.

It will, ok - I will split this pr

SlavikMIPT:
@parjong PTAL

binarman (approved):
LGTM

binarman:
removed Parjong, because these changes looks trivial and do not change neither behavior, nor architecture.

Issue #7438, opened 2021-08-10
[luci] More to NHWC (Logistics and Mul with scalar)


seanshpark:
Let's support more Ops in NHWC conversion
- `Logistics`
- `Mul` with (x * scalar constant)

Add
- `Sub` with (scalar constant - y)

seanshpark:
done

PR #7448, opened 2021-08-11
DRAFT ConvertToNHWC for Logistic and Mul w/scalar

seanshpark:
on-going draft to convert to NHWC for Logistic and Mul with scalar const.

ADD: and `Sub` with scalar


seanshpark:
For #7438

seanshpark:
done

PR #7449, opened 2021-08-11
remove tflite's open_cl files from coverage test

chunseoklee:
This is a workdaround to resolve line coverage problem.
gpu_cl's open_cl(depth1) folder contains tflite's opencl files. This
commit will remove those files from coverage test.


hseok-oh (approved):
We need to split this ported code as separated static library with `-fPIC` option and link with `gpu_cl` backend. 

Then we can manage ported part individually and remove coverage without handling exceptional case.

PR #7452, opened 2021-08-11
[luci] Revise toNHWC Mul to support scalar Const

seanshpark:
This will revise convert toNHWC of Mul to support scalar Constant.


seanshpark:
For #7438, from draft #7448

jinevening (approved):
LGTM

PR #7454, opened 2021-08-11
[luci] Support Logistic in convert to NHWC

seanshpark:
This will enable to support Logistic Op in convert to NHWC.


seanshpark:
For #7438, from draft #7448

jinevening (approved):
LGTM

PR #7455, opened 2021-08-11
[luci] Support quantized model with Exp

seanshpark:
This will enable import of Exp with quantized data type.

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleExp.cpp
+++ b/compiler/luci/import/src/Nodes/CircleExp.cpp
@@ -38,6 +38,10 @@ bool CircleExpGraphBuilder::validate(const ValidateArgs &args) const
     case circle::TensorType_FLOAT32:
     case circle::TensorType_FLOAT64:
       break;
+    // Additional support for quantized tensors
+    case circle::TensorType_UINT8:
+    case circle::TensorType_INT16:
+      break;
     // TODO support TensorType_COMPLEX64, complex128, bfloat16
     default:
       return false;


seanshpark:
This is for our internal quantized model with `Exp`.

llFreetimell:
Currently, all `GraphBuilder::validate` function are following TensorFlow rules AFAIK.
However, according to https://www.tensorflow.org/api_docs/python/tf/math/exp, `Exp` does not accept `uint8` and `int16`.
If we merge this PR, I think we begin to break the rule.
Is it okay to go as it is? I cannot decide whether is it fine or not...

seanshpark:
> If we merge this PR, I think we begin to break the rule.

This rule is to provide a minimul validation to check circle is valid or not from operator level.
When we first began, we had no experience and no guides and just to make our own.
From discussion and from short work experience, we found a formal document, written, TensorFlow guides that we : SR and SRR could share and follow without asking every time how it should be.
Throught the devepment, there was difference of level of implementation. Some followed strictly and some just added a TODO.
Now we have some expereince that, at least for me, these validation rules does not provide a good validation method.
There wasn't any broken models that some of input types was invalid.
There was an one invalid model case where the model file itself a LFS link file.

And we have another needs to support NPU that has different types.

> Is it okay to go as it is? 

If you have any better idea (like say to our customer don't use that Exp Op) please let me know. I'll drop this and consider it.

llFreetimell (approved):

PR #7457, opened 2021-08-11
[luci] Revise toNHWC to use convert_unary_x

seanshpark:
This will revise toNHWC for Ops with one x() to use convert_unary_x method.


jinevening (approved):
LGTM 👍

PR #7463, opened 2021-08-11
[luci] Resolve return type conflict in arity(void) methods

SlavikMIPT:
This commit resolves return type conflict in arity(void) methods which trigger a compiler error when using arm-none-eabi-gcc.


seanshpark:
Code itself looks OK. I'll land this after `loco` part has landed.

seanshpark:
one minor thing, it would be better to remove `[loco]` from title prefix `[luci][loco]`

lucenticus (approved):
LGTM

seanshpark (approved):
LGTM

PR #7471, opened 2021-08-12
[luci] Introduce MaxPoolWithArgmax resolving pass

binarman:
This commit adds resolving pass for MaxPoolWithArgmax op.

related issue: #7229

### DIFF
--- a/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
@@ -757,30 +757,34 @@ namespace luci
  *                    [CircleNode]
  *                    /          \
  *       [Split over channels]  [MaxPool2D]
- *             /   |   \              \
- *       [Conv2D] ...  ...      [MaxPool output]
+ *         /       |      \              \
+ *   [Requantize] ...     ...      [MaxPool output]
+ *         |
+ *      [PadV2]
+ *         |
+ *      [Conv2D]
+ *         |
+ *      [ArgMax]
+ *         |
+ *    [Reshape to 4d]
+ *         |
+ *   [Cast to fp32]
+ *    /        |
+ *   |  [Mul 1/<window width>]
+ *   |                \
+ *   |              [Floor]
+ *   |              /     \
+ *   | [Mul window width] |
+ *   \       /           /
+ *    \   [Neg] [Mul input width]
+ *     \   /    /
+ *     [Add]   /
+ *         \  /
+ *        [Add]
+ *          |
+ *     [Add const]
  *           |
- *       [ArgMax]
- *           |
- *     [Reshape to 4d]
- *           |
- *     [Cast to fp32]
- *      /        |
- *     |  [Mul 1/<window width>]
- *     |                \
- *     |              [Floor]
- *     |              /     \
- *     | [Mul window width] |
- *     \       /           /
- *      \   [Neg] [Mul input width]
- *       \   /    /
- *       [Add]   /
- *           \  /
- *          [Add]
- *            |
- *       [Add const]
- *            |
- *   [Mul number of channels]
+ * [Mul number of channels]
  *             \
  * [Optional Add with channels id]   ...  ...
  *                            \      |     /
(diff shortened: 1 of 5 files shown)


kvochko (approved):
LGTM

seanshpark:
@mhs4670go , PTAL

binarman:
@seanshpark  @mhs4670go FYI
This is basic pass implementation, it works with floating point type, but has some precision problems after quantization:
- floating point can represent integer numbers accurately, so integer computations using fp32 are correct.
- quantized numbers are guarantee to represent accurately only zero, but not other values.
- this pass use mul + floor (to emulate FloorDiv, it is easier to support in backend) and cast (fp->integer) operators. These operators fails to perform correct computations when applied on "integer" quantized numbers. For example integer 2 could be quantized to 1.996 after that `floor` can decrease it to 1.0 which is undesirable. 

To improve it I am going to make another PR that introduces adjustment coefficients.

Draft with these adjustments: #7481 ([one](https://github.com/Samsung/ONE/pull/7481/files#diff-12b562f1cad3615f227db91860a50b75bd907dd91a2411832c79846a42213937R275-R279), [two](https://github.com/Samsung/ONE/pull/7481/files#diff-12b562f1cad3615f227db91860a50b75bd907dd91a2411832c79846a42213937R463-R470))

seanshpark:
CC @jinevening

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -0,0 +1,829 @@
...
+
+namespace luci
+{
+
+/**
+ * BEFORE
+ *                 |
+ *            [CircleNode]
+ *                 |
+ *     [CUSTOM(MaxPoolWithArgmax]
```
mhs4670go, line 762:
```suggestion
 *     [CUSTOM(MaxPoolWithArgmax)]
```

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.test.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "luci/Pass/ResolveCustomOpMaxPoolWithArgmaxPass.h"
+
+#include <gtest/gtest.h>
+
+TEST(ResolveCustomOpMaxPoolWithArgmaxPassTest, name)
+{
+  luci::ResolveCustomOpMaxPoolWithArgmaxPass pass;
+  auto const name = pass.name();
+  ASSERT_NE(nullptr, name);
+}
```
mhs4670go, line 26: (optional) Testing with real graph would be good like `ResolveCustomOpBatchMatMulPass.test`.

mhs4670go: But seems bothering you to check above `after` graph..

mhs4670go: (optional) And, it would be good if current pass passes `circle2circle-dredd-recipe-test`. Value tests are more correct than code review with eyes..

mhs4670go (approved):
LGTM

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
mhs4670go: Since this change is trivial, it could be another PR(after merging this PR)

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.test.cpp
binarman: @mhs4670go 
I plan to add dredd tests to check pass works, but not value test.

It is hard to perform automated value test for now, because dredd tests get reference data using tflite model, but MaxPoolWithArgmax is a custom operation and available only in "big" TensorFlow.

To perform value tests I see two ways:
- Use [select operators](https://www.tensorflow.org/lite/guide/ops_select) in testing system
- Introduce new test suite that will compare circle execution with original tensor flow model (AFAIK there are no such test suite for now)

mhs4670go: @binarman Ah, you're right. I misunderstood for a moment. Thank you for detail:)

PR #7475, opened 2021-08-12
[record-minmax] Reject extremely small numbers from statistics

binarman:
This commit adds filter that rejects numbers that are too small.
Such numbers are produced by MaxPoolWithArgmax resolving pass.

related issue: #7229

### DIFF
--- a/compiler/record-minmax/src/MinMaxObserver.cpp
+++ b/compiler/record-minmax/src/MinMaxObserver.cpp
@@ -86,6 +86,10 @@ void MinMaxObserver::postTensorWrite(const luci::CircleNode *node,
     if (isnan(number))
       continue;
 
+    // TODO use metadata hints to detect such cases
+    if (number == std::numeric_limits<float>::lowest())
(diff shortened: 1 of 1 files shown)


compiler/record-minmax/src/MinMaxObserver.cpp
```diff
@@ -86,6 +86,10 @@ void MinMaxObserver::postTensorWrite(const luci::CircleNode *node,
     if (isnan(number))
       continue;
 
+    // TODO use metadata hints to detect such cases
```
binarman, line 89: @jinevening 
This is what I was talking about in https://github.com/Samsung/ONE/pull/7236/files#r687530696

What do you think about adding hints in circle metadata to control such hacks?

jinevening: It looks ok. The lowest value of float may not appear in normal execution.

jinevening (approved):
LGTM

PR #7488, opened 2021-08-17
[one-cmds] Use store_true in one-cmd utils

jyoungyun:
This commit changes the action way to use store_true instead of store_const.
The store_true means if the option is specified, assign the value True to verbose.
Not specifying it implies False.
The actual operation is the same as before.


glistening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #7491, opened 2021-08-17
[infra] Fix one-codegen man page

mhs4670go:
This commit fixes one-codegen man page.

ONE-DCO-1.0-Signed-off-by: seongwoo <mhs4670go@naver.com>

PR #7493, opened 2021-08-17
[infra] Remove unnecessary comment in one-codegen man page

jyoungyun:
This commit removes the unnecessary comment in one-codegen man page.


mhs4670go:
duplicate:) #7491

PR #7496, opened 2021-08-17
[tf2tfliteV2] Redirect the error message

mhs4670go:
This commit redirects the error message to std.err.


glistening (approved):
LGTM. The same code in `utils.py` needs to be updated, too. Thanks!

Issue #7400, opened 2021-08-05
Support pre-built binary on runtime


chunseoklee:
### TODO

- [ ] Define pre-built binary format on nnpackage(#7395)
- [ ] Define operation for PBB(pre-built binary)
- [ ] Loader for PBB
- [ ] TBD

chunseoklee:
Here is a picture which depicts the whole flow during model conversion from pb/tflite to nnpkg with pre-built(e.g. NPU) binary :  
<img src=https://user-images.githubusercontent.com/4862887/133561973-c5c34080-fa3d-4fb8-80cb-691054c6d287.png height=300px>

Here is TODOs on compiler side, IMHO.

- Define nnpkg format for partitioned models with pre-built binaries(This is both for runtime and compiler)
    + Consider the nnpkg containing both circle model and pre-built binary
- Toolchain(in one-cmd) for generating the multi-model nnpkg
    + This tool will compile pre-built binary, generate pkg information(e.g. meta info about prebuilt bin), and make an nnpkg.

PR #7501, opened 2021-08-17
[Draft] Prebuilt binary operation in runtime 

chunseoklee:
This draft is a part of https://github.com/Samsung/ONE/issues/7400.

> Define operation for PBB(pre-built binary)
>  Loader for PBB

`Bulk` is used for fake operation name, here. Now, `Prebuilt` or `Bin` sounds better.


runtime/onert/core/include/ir/operation/Bulk.h
```diff
@@ -0,0 +1,54 @@
...
+#include "ir/Operation.h"
+
+namespace onert
+{
+namespace ir
+{
+namespace operation
+{
+
+class Bulk : public Operation
```
hseok-oh, line 29: IR `Bulk` looks good to me.

PR #7510, opened 2021-08-17
[tflite2circle] Add verbose option

jyoungyun:
This commit adds verbose option.

### DIFF
--- a/compiler/tflite2circle/driver/Driver.cpp
+++ b/compiler/tflite2circle/driver/Driver.cpp
@@ -43,6 +43,12 @@ int entry(int argc, char **argv)
     .help("Show version information and exit")
     .exit_with(print_version);
 
+  arser.add_argument("-V", "--verbose")
+    .nargs(0)
+    .required(false)
+    .default_value(false)
+    .help("output additional information to stdout or stderr");
+
   arser.add_argument("tflite")
     .nargs(1)
     .type(arser::DataType::STR)


glistening (approved):
LGTM

jyoungyun:
@nnfw-bot test nncc-release

mhs4670go:
@jyoungyun As you said in offline, there is no contents to verbose. What it is for?

jyoungyun:
@mhs4670go 
Yes, there is no log data in this tool. But I want to keep the same usability between our tools. All other tools in this project, there is a verbose option. What do you think about it?

mhs4670go:
@jyoungyun Technically speaking, verbose option is only needed to `one-cmds`(python script). Because c++ tools can be verbosed just by setting `LUCI_LOG` environment. Anyway, if it looks fine to you, this can be merged.

mhs4670go (approved):
LGTM

jyoungyun:
> Technically speaking, verbose option is only needed to one-cmds(python script). Because c++ tools can be verbosed just by setting LUCI_LOG environment. Anyway, if it looks fine to you, this can be merged.

Yes, right. But if users want to use the native tool directly, they probably don't know the LOG TAG like `LUCI_LOG`. Because there is a `--verbose` option, the user will be able to use it more conveniently.

Thank you!

PR #7516, opened 2021-08-18
[luci] Requantize tensor after floor op in ResolveCustomOpMaxPoolWithArgmax pass

binarman:
This commit adds requantization node in ResolveCustomOpMaxPoolWithArgmax
to force circle-quantizer produce tensors with real scales after floor instead of 1.0.

related issue #7229

### DIFF
--- a/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
@@ -446,6 +446,43 @@ luci::CircleNode *window_flattened_coord(const std::string &name, luci::Padding 
   return argmax_cast;
 }
 
+// Creates dummy operation after floor to force circle-quantizer requantize tensor with scale << 1
+// Dealing with values of extremely different scales
+// in following binary operations hurts backend precision.
+luci::CircleNode *create_post_floor_requantize_node(luci::CircleFloor *floor)
+{
+  auto graph = floor->graph();
+  auto const origin = luci::get_origin(floor);
+  auto name = floor->name();
+
+  auto requantizer = none_act_func(graph->nodes()->create<luci::CircleDepthwiseConv2D>());
+  init_name_and_origin(requantizer, name + "/Requantizer", origin);
+
+  requantizer->input(floor);
+
+  auto requantizer_filter = create_scalar<loco::DataType::FLOAT32>(graph, 1.0f);
+  init_name_and_origin(requantizer_filter, name + "/Requantizer/filter", origin);
+  requantizer_filter->rank(4);
+  for (uint32_t i = 0; i < 4; ++i)
+  {
+    requantizer_filter->dim(i) = 1;
+  }
+  requantizer->filter(requantizer_filter);
+
+  auto requantizer_bias = create_zero_bias(graph, 1);
+  init_name_and_origin(requantizer_bias, name + "/Requantizer/bias", origin);
+  requantizer->bias(requantizer_bias);
+
+  requantizer->padding(luci::Padding::VALID);
+  requantizer->stride()->w(1);
+  requantizer->stride()->h(1);
+  requantizer->depthMultiplier(1);
+  requantizer->dilation()->w(1);
+  requantizer->dilation()->h(1);
+
+  return requantizer;
+}
+
 luci::CircleNode *window_y_coord(const std::string &name, float filter_width,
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -446,6 +446,43 @@ luci::CircleNode *window_flattened_coord(const std::string &name, luci::Padding
...
+// Creates dummy operation after floor to force circle-quantizer requantize tensor with scale << 1
+// Dealing with values of extremely different scales
+// in following binary operations hurts backend precision.
+luci::CircleNode *create_post_floor_requantize_node(luci::CircleFloor *floor)
+{
+  auto graph = floor->graph();
+  auto const origin = luci::get_origin(floor);
+  auto name = floor->name();
+
+  auto requantizer = none_act_func(graph->nodes()->create<luci::CircleDepthwiseConv2D>());
```
binarman, line 458: `depthwise conv` was chosen in favor of other operations to prevent it's optimization (elimination of fuse)

ai-moiseev (approved):
LGTM

binarman:
@seanshpark
This PR is needed for August release along with #7517, could you take a look at them today?

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
```diff
@@ -784,6 +823,8 @@ namespace luci
  *   |  [Mul 1/<window width>]
  *   |                \
  *   |              [Floor]
+ *   |                 |
+ *   |            [Requantize]
```
seanshpark, line 827: Code seems to be `CircleDepthwiseConv2D` but `Requantize` here. Is  `Requantize` a meaning of a purpose?
May this?
```
[DepthwiseConv2D] for requantize
```

binarman: Yes, this is DepthwiseConv2D, but it serves requantization (aka rescale) purpose.

I've added `[Requantize]` just for simplicity, because it is not really matter what nop operation to use (unless it is not wiped out by optimizations)

If you think this matters I propose to merge this change, plus https://github.com/Samsung/ONE/pull/7517/files#r691675340 in batch in new PR.
What do you think?

seanshpark: (1) Using `Requantize` may make confusion understanding and also only `DepthwiseConv2D` may make hard to understand the purpose of inserted op. Like existing nodes, `[DepthwiseConv2D for requantize]` seems OK.

(2) Personnally I do not like to mix multiple contexts in one PR. It makes understanding difficult.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
seanshpark: Maybe adding a note about why `DepthwiseConv2D` acts as dequantizer.

binarman: Is [this](https://github.com/Samsung/ONE/pull/7516/files#diff-12b562f1cad3615f227db91860a50b75bd907dd91a2411832c79846a42213937R449-R451) comment sufficient?

binarman:
@seanshpark FYI
I do not know when @ai-moiseev would be able to check this PR again, so I would like to merge it without his approval.

compiler/luci/pass/src/ResolveCustomOpMaxPoolWithArgmaxPass.cpp
seanshpark: > Is this comment sufficient?

> // Creates dummy operation after floor to force circle-quantizer requantize tensor with scale << 1
// Dealing with values of extremely different scales
// in following binary operations hurts backend precision.

Explanation itself looks good. But I could not catch this `dummy` is the `CircleDepthwiseConv2D`.
I think explicitly adding a note would be nice.

jinevening (approved):
LGTM

seanshpark (dismissed):

seanshpark (approved):
LGTM thank you!

binarman:
@jinevening PTAL

jinevening (approved):

Issue #7521, opened 2021-08-18
LUCI_LOG disabled


seanshpark:
```cxx
  if (arser.get<bool>("--verbose"))
    setenv("LUCI_LOG", "100", true);
  else
    setenv("LUCI_LOG", "0", true);
```

Now that setting `LUCI_LOG=1` in command line is DISBLED!!!!!!!!

PLEASE DO NOT ALTER WHAT WAS WORKING TO DIFFERENT BEHAVIOUR!!!!!

seanshpark:
CC @lemmaa 

@jyoungyun PLEASE FIX THIS!!!!!

seanshpark:
`setenv("LUCI_LOG", "100", true);` IS BAD USAGE FOR `LUCI_LOG`

jyoungyun:
I was guided by @mhs4670go to use `100`. I will talk to him.

lemmaa:
> I was guided by @mhs4670go to use `100`. I will talk to him.

Apart from that, it seems that it was necessary to first check whether a value is set in LUCI_LOG. :)

jyoungyun:
When I checked this value in this project, test already use 100 and @mhs4670go suggests me to use 100 too.
https://github.com/Samsung/ONE/blob/8241bc459ee82b2727b285852562eb750f37f378/compiler/luci/tests/readverify.sh#L11
And when I tested this value as 1, 2, 3 and 100, the 100 is that most detailed log output. So I chose it.

lemmaa:
Another problem seems to be the lack of documentation about the currently implemented feature and usage, and the lack of design discussion about applying `--vesbose`. (Of course, in both cases, maybe it's just something I didn't find.)

It is also our mistake that occurred due to the urgent schedule.

I think it will be a good lesson for us. :)

seanshpark:
(1) The problem I think is
```cxx
  else
    setenv("LUCI_LOG", "0", true);
```
This should not be added. This will eatup given from command line.

(2) `setenv("LUCI_LOG", "100", true); ` is bad usage
I mean, this is not wrong, setting to `100` maybe a way to work as `verbose`
The `BAD` in my viewpoint is we can use (or add if not present) by an interface
not by calling `setenv` cause the code that reads `LUCI_LOG` is by `getenv` or something.

I understand there is short time slot to implement things talked in UX T/F so happend.

jyoungyun:
There are some wrong points in my code. :(

> (1) The problem I think is
> ```cxx
[...]

PR #7526, opened 2021-08-19
[record-minmax] Change the verbose behavior

jyoungyun:
This commit changes the verbose behavior.
The LUCI_LOG value set by the user should have higher priority
than the value of --verbose option.

Related issue: #7521

### DIFF
--- a/compiler/record-minmax/driver/Driver.cpp
+++ b/compiler/record-minmax/driver/Driver.cpp
@@ -105,9 +105,11 @@ int entry(const int argc, char **argv)
   }
 
   if (arser.get<bool>("--verbose"))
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

jyoungyun:
In this code, if user sets the LUCI_LOG as zero and passes the --verbose option to the our tool, it does not print luci logs.

```
$ LUCI_LOG=0 ../bin/onecc --verbose -C onecc_003.cfg
# Print tensorflow logs only
# Not print luci log
```

mhs4670go (approved):
LGTM

glistening (approved):
LGTM

PR #7527, opened 2021-08-19
[circle2circle] Change the verbose behavior

jyoungyun:
This commit changes the verbose behavior.
The LUCI_LOG value set by the user should have higher priority
than the value of --verbose option.

Related issue: #7521

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -399,10 +399,11 @@ int entry(int argc, char **argv)
   }
 
   if (arser.get<bool>("--verbose"))
-    setenv("LUCI_LOG", "100", true);
-  else
-    setenv("LUCI_LOG", "0", true);
-
+  {
+    // The third parameter of setenv means REPLACE.
+    // If REPLACE is zero, it does not overwrite an existing value.
+    setenv("LUCI_LOG", "100", 0);
+  }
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

jyoungyun:
In this code, if user sets the LUCI_LOG as zero and passes the --verbose option to the our tool, it does not print luci logs.

```
$ LUCI_LOG=0 ../bin/onecc --verbose -C onecc_003.cfg
# Print tensorflow logs only
# Not print luci log
```

seanshpark:
> In this code, if user sets the LUCI_LOG as zero and passes the --verbose option to the our tool, it does not print luci logs.

This seems a tricky case. Currently `LUCI_LOG` has higher priority than `--verbose`.
I'm not sure which taking higher priority is better for the user.

mhs4670go (approved):
LGTM

jyoungyun:
> This seems a tricky case. Currently LUCI_LOG has higher priority than --verbose.
> I'm not sure which taking higher priority is better for the user.

I think the user who knows the `LUCI_TAG` is most likely an internal developer. Therefore, if the person gives the value of another environment variable, there might be a specific purpose, so I decided that it would be better to give his environment value a higher priority.

If there is a missed point in my comment, please give me advice.

glistening (approved):
LGTM

PR #7478, opened 2021-08-13
DRAFT: Convert PadV2 to Pad

hyunsik-yoon:
This is a draft to convert PadV2 with negative padding value to
Pad (of which padding value is 0).

for #7477

Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>

Issue #7477, opened 2021-08-13
Compiler FE: Convert PadV2 to Pad


hyunsik-yoon:
### what

Let's convert PadV2 (of which padding value is negative number) to Pad (padding value is `0`).

### why

While converting pytorch to onnx, we found that
sometimes PadV2 with very huge negative number is added before MaxPool.

The output of MaxPool works well with F32 type 
but when we quantize such PadV2, huge negative number could cause unexpected issue.

In certain condition, we could change such negative number to `0`.

hyunsik-yoon:
all merged.

PR #7528, opened 2021-08-19
[luci] Check if MaxPool2D is Only-Successor

hyunsik-yoon:
This adds code to check if `MaxPool2D` is only-successor of `PadV2` in `SubstitutePadV2ToPadPass.cpp`.

For #7477
Draft #7478

(I will add test code in next PR.)

### DIFF
--- a/compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
+++ b/compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
@@ -285,11 +285,11 @@ bool used_by_maxpool_only(luci::CircleNode *node, Paddings &paddings)
 }
 
 // Check condition C3), C4) and C6)
-bool used_by_maxpool_only(luci::CirclePadV2 *node)
+bool used_by_maxpool_only(luci::CirclePadV2 *pad_v2)
 {
-  Paddings paddings(loco::must_cast<luci::CircleConst *>(node->paddings()));
+  Paddings paddings(loco::must_cast<luci::CircleConst *>(pad_v2->paddings()));
 
-  return used_by_maxpool_only(node, paddings);
+  return used_by_maxpool_only(pad_v2, paddings);
 }
 
 loco::Node *build_pad_from(luci::CirclePadV2 *pad_v2)
--- a/compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
+++ b/compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
@@ -83,6 +83,94 @@
 namespace
 {
 
+struct Paddings
+{
+  struct Pad
+  {
+    int32_t front;
+    int32_t end;
+  };
+  /**
+   * @brief Store paddings position information.
+   * @details _padding_pos[k] stores Pad object at axis k
+   */
+  std::vector<Pad> _padding_pos;
+
+  Paddings(luci::CircleConst *paddings)
+  {
+    assert(paddings->dtype() == loco::DataType::S32);
+    assert(paddings->rank() == 2);
+    assert(paddings->dim(1).value() == 2);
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
```diff
@@ -159,16 +259,39 @@ bool has_all_positive_values(luci::CircleConst *node)
...
+    paddings.apply(transpose);
+    return used_by_maxpool_only(transpose, paddings);
+  }
+  // Support more 'reshaping op' later
 
   return false;
 }
 
+// Check condition C3), C4) and C6)
+bool used_by_maxpool_only(luci::CirclePadV2 *node)
```
seanshpark, line 288: (optional) below function used `pad_v2` and `node` here. maybe go with one name?

seanshpark (dismissed):
LGTM

compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
hyunsik-yoon: `node` was renamed to `pad_v2`. Thanks.

compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
```diff
@@ -83,6 +83,94 @@
...
+
+  /**
+   * @brief Check if this padding area is covered by filter
+   *
+   * @note This is to check condition C4).
+   *       _padding_pos should store values according to NHWC.
+   */
+  bool smaller_than(int32_t filter_h, int32_t filter_w)
+  {
+    assert(_padding_pos.size() == 4);
```
seanshpark, line 122: If there can be `_padding_pos.size() != 4` for any model, 
it would be better to `return false` if not met the condition
- debug build will stop, which will break the testing/development
- release will just go and may raise access violation.

if this is to break on purpose to check while developing, it is OK.

compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
```diff
@@ -83,6 +83,94 @@
...
+   *   This method keeps track of such change of padding position.
+   */
+  void apply(luci::CircleTranspose *transpose)
+  {
+    assert(transpose);
+    luci::CircleConst *perm = loco::must_cast<luci::CircleConst *>(transpose->perm());
+
+    assert(perm && perm->dtype() == loco::DataType::S32);
+    assert(_padding_pos.size() == 4);
+    assert(perm->size<loco::DataType::S32>() == 4);
```
seanshpark, line 159: Here too.. if these conditions can be false for normal models, it would be better to just return false.
If these MUST be satisfied, it is OK.

seanshpark (approved):

compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
```diff
@@ -93,6 +93,8 @@ struct Paddings
   /**
    * @brief Store paddings position information.
    * @details _padding_pos[k] stores Pad object at axis k
+   *
+   * @note  Paddings must be for rank 4 tensor
```
hyunsik-yoon, line 97: Now, class `Paddings` support tensor with rank-4 only and `assert` exists only in the constructor.

compiler/luci/pass/src/SubstitutePadV2ToPadPass.cpp
hyunsik-yoon: That is a very good recommendation. I fixed code.

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

PR #7559, opened 2021-08-20
Update runtime debian pkg version to 1.17.0

chunseoklee:
Use 1.17.0 in runtime debian package


mhs4670go (approved):
LGTM

PR #7569, opened 2021-08-24
[one-cmds] utils._run will same code regardless of logfile presence

glistening:
Previously, utils._run uses different style of stdout/stderr capturing
for logfile = None and not. Now, both case use select version.


glistening:
When `logfile` is absent, the `stdout` and `stderr` output of subprocess are separated.

For example, when original native output is:
```
STDERR error message
STDOUT help message
STDOUT help message
STDOUT help message
```

the previous implementation showed in my environment:
```
STDOUT help message
STDOUT help message
STDOUT help message
STDERR error message
```

With `select` solution, it shows better output though no one may guarantee the order of mixed output from stdout and stderr.
```
STDERR error message
STDOUT help message
STDOUT help message
STDOUT help message
```

compiler/one-cmds/utils.py
```diff
@@ -243,33 +243,26 @@ def _run(cmd, err_prefix=None, logfile=None):
...
+                    continue
+                if x == p.stdout:
+                    out = sys.stdout
+                if x == p.stderr:
+                    out = sys.stderr
+                    if err_prefix:
+                        line = f"{err_prefix}: ".encode() + line
+                out.buffer.write(line)
+                out.buffer.flush()
+                if logfile == None:
```
mhs4670go, line 265: Maybe?
```suggestion
                if logfile != None:
```

glistening: 😨 ... I've updated.

mhs4670go (approved):
LGTM

jyoungyun (approved):
LGTM

glistening:
@Samsung/nncc_committers PTAL

glistening:
@seanshpark Could you please merge this?

Issue #7505, opened 2021-08-17
[circle2circle] Constant folding for DepthwiseConv2D


ai-moiseev:
## What

Implement folding optimization to simplify DepthwiseConv2D nodes with constant inputs.

## Why

There's need to optimize models with such a pattern.

## Steps

* [x] DepthwiseConv2D constant folding pass #7576 
* [x] Optimization option in `circle2circle` #7597 
* [x] Option in `one-cmd` #7613 
* [x] Unit tests #7576 

Draft: #7570

seanshpark:
Also need to add option to `one-cmds`

llFreetimell:
When trying to build `luci` only, following header file causes build error.

https://github.com/Samsung/ONE/blob/072c4c0377bdbc600a035a3faa1d6a389fa7c906/compiler/luci/pass/src/FoldDepthwiseConv2DPass.cpp#L19

Cannot find the header file... :(

PR #7583, opened 2021-08-27
[mio-tflite] Add header interfaces for TF and GEMMLowp

ai-moiseev:
This commit adds TensorFlow include directories to mio_tflite_inc target with INTERFACE scope to allow usage of TF headers with `target_link_libraries(<target> mio_tflite_inc)`.

For #7505 
Draft #7582 


seanshpark (approved):
LGTM thank you!

PR #7588, opened 2021-08-30
[luci] Support const input quantization for Exp

ai-moiseev:
This commit adds support for const input quantization for Exp op.


seanshpark (approved):
LGTM

jinevening (approved):

Issue #7602, opened 2021-09-03
[circle2circle] Broadcast removal for contstant inputs


ai-moiseev:
## What

Implement transformation pass to expand constant node inputs following broadcasting rules.

## Why

There's need to transform models with such a pattern.

## Steps

* [x] RemoveBroadcast transformation pass #7603 
* [x] Option in `circle2circle` #7646
* [x] Option in `one-cmd` #7675 
* [x] Unit tests

seanshpark:
@ai-moiseev , could you please share which models needs this and the reason?

ai-moiseev:
@seanshpark This pattern appears in one of the internal models, I've emailed you link to it...

seanshpark:
Thanks for the mail :)

For the record; see compiler backend issue 9567

seanshpark:
close as done

PR #7603, opened 2021-09-03
[luci] Introduce ExpandBroadcastConst pass

ai-moiseev:
This commit introduces ExpandBroadcastConst transformation pass for expanding of constant inputs in case of possible broadcast.

For #7602 


compiler/luci/pass/src/RemoveBroadcastPass.cpp
```diff
@@ -0,0 +1,167 @@
...
+bool RemoveBroadcastPass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto const_node = dynamic_cast<luci::CircleConst *>(node);
+    if (const_node == nullptr)
+      continue;
+
+    changed = changed || expand_broadcast_const(const_node);
```
seanshpark, line 162: It would be better to follow other Pass codes.

compiler/luci/pass/src/RemoveBroadcastPass.cpp
```diff
@@ -0,0 +1,167 @@
...
+ *    [CircleInput] [CircleConst (H x W x D)]
+ *               |     |
+ *             [CircleAdd]
+ */
+bool expand_broadcast_const(luci::CircleConst *node)
+{
+  if (node->dtype() != loco::DataType::FLOAT32)
+    return false; // Unsupported data type
+
+  for (auto successor : loco::succs(node))
```
seanshpark, line 127: Is there a reason to use `loco::succs()` for node?

seanshpark: If this constant is used multiple times (there are models like this), would this pass work as expected?

compiler/luci/pass/src/RemoveBroadcastPass.cpp
ai-moiseev: Done, replaced with branching.

compiler/luci/pass/src/RemoveBroadcastPass.cpp
ai-moiseev: The reason here to use `loco::succs()` is to check which successors require broadcasting, as constant node can be connected with both types (and may have different broadcast expansions). In case of multiple usages it should work correctly, but may create extra copies of expanded constant.

compiler/luci/pass/include/luci/CircleOptimizer.h
```diff
@@ -67,6 +67,7 @@ class CircleOptimizer final
       SubstitutePadV2ToPad,
       SubstituteSqueezeToReshape,
       ConvertNCHWToNHWC,
+      RemoveBroadcast,
```
seanshpark, line 70: As this is about `Const` broadcast, can we use `RemoveConstBroadcast`?
Um... purpose is to remove broadcasting in successor, actual thing is to expand Const to match shape...
`RemoveConstBroadcast` itself seems not quite fit...
@jinevening , can you propose a good name for this?

compiler/luci/pass/src/RemoveBroadcastPass.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+template <typename N> bool expand_node_input(luci::CircleConst *node, luci::CircleNode *successor)
+{
+  static_assert(std::is_base_of<luci::CircleNode, N>::value,
+                "Successor node should have CircleNode base");
+
+  auto successor_node = loco::must_cast<N *>(successor);
+  luci::CircleConst *expanded_const;
+
```
seanshpark, line 82: I'd like to remove two lines code that hurts readability.
Can you declare x and y and use it?
```cxx
auto successor_x = loco::must_cast<luci::CircleNode *>(successor_node->x());
auto successor_y = loco::must_cast<luci::CircleNode *>(successor_node->y());
```

compiler/luci/pass/src/RemoveBroadcastPass.cpp
seanshpark: Could you add a test case or at least a TODO comment for multiple successors for the Const?

compiler/luci/pass/src/RemoveBroadcastPass.cpp
ai-moiseev: Sure, done, more readable and laconic now :)

compiler/luci/pass/src/RemoveBroadcastPass.cpp
ai-moiseev: Done, added multi-successor test case.

compiler/luci/pass/include/luci/CircleOptimizer.h
jinevening: > can you propose a good name for this?

I think `RemoveConstBroadcast` or `ExpandBroadcastConst` is a better name.

Plus, I think that it would be better to implement this transformation in the backend.

jinevening: The transformation code looks ok to me.

seanshpark: yes model size may grow... as this is enabled by option switch we can provide in frontend:)

seanshpark (dismissed):

compiler/luci/pass/include/luci/CircleOptimizer.h
seanshpark: @ai-moiseev if possible could you consider the pass name modification?

jinevening (approved):

compiler/luci/pass/include/luci/CircleOptimizer.h
ai-moiseev: Done, I've taken ExpandBroadcastConst variant, since no removal takes place here, but rather expansion.

jinevening (approved):
LGTM

seanshpark (approved):
LGTM thank you!

PR #7605, opened 2021-09-03
[DRAFT] [luci-interpreter] Move kernel builders to separate files

binarman:
This PR moves kernel builders to separate files
to enable partial build (subset of kernels) of luci-interpreter.
This could be used for MCUs builds to reduce memory consumption.

related issue: #7288
    

binarman:
This PR splits large `KernelBuilder.cpp` file in multiple files, to enable conditional build for MCUs.

I've tried to preserve git history, but I don't know how to do it without copying enormous amount of code(idea is to merge copy of whole `KernelBuilder.cpp` file for every kernel builer, then remove redundant code from each file).

I don't like the burden this approach introduction of large redundant code So I want to merge this as-is, destroying history

So I want to merge this in several steps:
- [] Move set of `KERNEL_REGISTER_FILE` to  `KernelsToBuild.lst` to interpreter root `CMakeLists.txt`, so I can use it for both kernels and loaders libraries (now it is used only for kernels).
- [] Move `collectOutputNodes` function to `KernelBuilderHelper.h` to use in splitted builder files.
- [] Add separate `.cpp` and `.h` files for every kernel builder
- [] Switch to code from introduced files:
   - remove splitted code of builder functions from `KernelBuilder.cpp`
   - update included headers in `KernelBuilder.cpp`
   - enable build of splitted files in `CMakeLists.txt`

binarman:
A few words about how this RP was created

- Cut actual kernel builder functions from cpp to separate file
- Create "Nodes" subdirectory
- run https://github.com/Samsung/ONE/pull/7605/files#diff-367c6a06df7f6d442b5aa81c938ce48246c3bdf87a4c4076bf2080125320a01c script 
- fix redundant lines script generated (it's "disposable" script, so I didn't bother to write actual code parsing, and it makes some mistakes)
- Rewrite other code to make it work with splitted functions

binarman:
everything is merged, closing.

Issue #7609, opened 2021-09-06
[luci] Relax constraints of ConvertNCHWToNHWC for Add, Mul, Sub


jinevening:
## What

Let's relax the shape constraints of ConvertNCHWToNHWC for Add, Mul, Sub.

For Add, we apply NCHW -> NHWC conversion only when the below condition holds.

```
// We assume ADD with const input is NCHW if,
// Input shape: (N, C, H, W)
// Output shape: (N, C, H, W)
// 1. Const shape is (1, C, 1, 1) or a scalar (1)
// 2. Input, Output, Const have the same C.
bool is_NCHW_with_const(const luci::CircleAdd *node, luci::CircleNode *&pred_node,
                        luci::CircleConst *&beta)
```

I'd like to relax the condition to support the case when Const shape is (N, C, H, W).

PR #7611, opened 2021-09-06
[luci] Support NHWC const for binary Ops in ConvertNCHWToNHWC

jinevening:
This supports NHWC const for binary Ops (Add, Mul, Sub) in CovertNCHWToNHWC.

---
Related to: https://github.com/Samsung/ONE/issues/7609

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -28,6 +28,22 @@
 namespace
 {
 
+bool check_shape(const luci::CircleNode *node, const std::vector<loco::Dimension> &shape)
+{
+  if (not node)
+    return false;
+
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -28,6 +28,22 @@
 namespace
 {
 
+bool check_shape(const luci::CircleNode *node, const std::vector<loco::Dimension> &shape)
```
seanshpark, line 31: (optional) about the name... `is_shape_same` or `is_same_shape` or `is_shape_equal`

jinevening: I'll go with `is_same_shape` :)

seanshpark (approved):
LGTM thank you!

PR #7613, opened 2021-09-06
[one-cmds] Add Fold DepthwiseConv2D optimization option

ai-moiseev:
This commit adds `fold_depthwise_conv_2d` optimization option to `one-build` config.

For #7505 

### DIFF
--- a/compiler/one-cmds/how-to-use-one-commands.txt
+++ b/compiler/one-cmds/how-to-use-one-commands.txt
@@ -152,7 +152,7 @@ Current transformation options are
 - disable_validation : This will turn off operator validations.
 - fold_add_v2 : This removes AddV2 operation which can be folded
 - fold_cast : This removes Cast operation which can be folded
-- fold_depthwise_conv_2d : This folds DepthwiseConv2D operation which can be folded
+- fold_dwconv : This folds Depthwise Convolution operation which can be folded
 - fold_dequantize : This removes Dequantize operation which can be folded
 - fold_sparse_to_dense : This removes SparseToDense operation which can be folded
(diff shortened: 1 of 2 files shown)


binarman:
There are also `how-to-use-one-commands.txt` file nearby. Did you add this command there?

seanshpark:
@ai-moiseev , @binarman , code itself is OK
But while looking at the name and comparing with other pass options,
- we used `Conv2D` as just `conv`
- `DepthwiseConv2D` as just `dwconv`

As to match with other options, (although existing may seem odd), what do you think about chanigng (including existing codes) option name to `fold_dwconv` ? I apologize not pointing this out earlier...

ai-moiseev:
@seanshpark I've created renaming PR for circle2circle, PTAL #7617
It seems it doesn't break any tests... when it will be merged, I'll rebase this PR.

compiler/one-cmds/how-to-use-one-commands.txt
```diff
@@ -152,6 +152,7 @@ Current transformation options are
 - disable_validation : This will turn off operator validations.
 - fold_add_v2 : This removes AddV2 operation which can be folded
 - fold_cast : This removes Cast operation which can be folded
+- fold_dwconv : This folds Depthwise Convolution operation which can be folded
```
seanshpark, line 155: (optional) maybe you may want to swap with below item?

seanshpark (approved):
LGTM thank you!

compiler/one-cmds/how-to-use-one-commands.txt
ai-moiseev: Sorry, I've read this comment a bit late... Here's PR which fixes sorting order: #7619 :)

PR #7617, opened 2021-09-07
[circle2circle] Rename DConv fold option as 'fold_dwconv'

ai-moiseev:
This commit renames 'fold_depthwise_conv_2d' option in circle2circle into 'fold_dwconv' to be more consistent with other option names.

For #7505 


seanshpark (approved):
LGTM

Issue #7604, opened 2021-09-03
[luci-interpreter] Remove constants duplication in RAM


binarman:
### Problem

To execute model using luci-interpreter we need  to:
1. Read circle file contents
2. Parse buffer to flatbuffers representation
3. Convert flatbuffers to luci IR
4. Convert luci IR to interpreter Runtime representation

Phase 3 and 4 copies constants, so effective RAM consumption is doubled.

This is a big problem on MCU, where RAM is a very limited resource.

#### Phase 3 solutions

**Idea 1**

Introduce utility that splits circle model in two files: network topology and weights.

Network topology file is a normal Circle file, but with empty buffers. This file could be read by normal luci importer.

**Idea 2**

Introduce specialized luci importer that takes buffer with circle model and produces two separate entities:
- luci graph with empty constants data (but correct shapes)
- Weights with mapping to original luci constants

#### Phase 4 solutions

**Idea 1**

Simply assign data pointer of runtime tensor to external data
DRAFT ~~#7572~~ #7625 (edited by @m-bronnikov)

**Idea 2**

Transform `CircleConst` nodes to real kernels to introduce different behaviors:
- simple copy of weright data (for example to bring data from slow to fast memory region)
- prefetch data (if DMA is available this coult bring some level of parallelism)
- access original constant (to save as much RAM as possible)

m-bronnikov:
Idea 2 of step 3 was implemented in local branch(https://github.com/m-bronnikov/ONE/tree/move_constants), but I got failure with this because source data buffer of tensor destroyed after importing. We need to find another approach for importer or choose Idea1.

m-bronnikov:
Let's track progress for step 3 via #8042!

seanshpark:
requirement is now for `onert-micro`

PR #7408, opened 2021-08-05
[Draft][luci-interpreter] Introduce MemoryManager

BalyshevArtem:
Introduce tensor memory manager class.
Role or this manager is to assign and free tensor memory.

issue #7355

PR #7625, opened 2021-09-08
[DRAFT] Reuse luci IR constants in luci-interpreter

binarman:
This PR introduces reuse contents of `CircleConst` in luci-interpreter.
That significantly reduces memory consumption of luci-interpreter based tools.

Related issue #7604 
This PR is based on #7408 DRAFT


binarman:
Let's try different approach, to remove all duplications, instead of one in this draft.
Closing, since approach is outdated.

Issue #7616, opened 2021-09-07
Compiler FE: Upgrade to TF 2.6.0


seanshpark:
Let's upgrade TF 2.3.0 to TF 2.6.0 for `tflite2circle` and related modules.

After this task is finished, 
- enable to use `SignatureDef` to fix index changes

seanshpark:
Previois task: https://github.com/Samsung/ONE/issues/7592#issuecomment-914066903

llFreetimell:
To correctly support TFLite 2.6.0, following things should be considered and updated too.

- https://github.com/Samsung/ONE/issues/5688#issuecomment-759911509
- TFLite related modules in #6269

llFreetimell:
By the way, can I ask you why mio-tflite260 is introduced?
At first, I thought following steps.

- Introduce `mio-tflite260` which uses TF 2.6.0
- Update tflite schame related functions in each module by using `mio-tflite260`
  - For example, https://github.com/Samsung/ONE/blob/f82912563af1da092bd8718f7bf0124d46c809d1/compiler/tfl-inspect/src/Reader.cpp#L27-L28
- After tflite schema changes are applied to all of modules, deprecate `mio-tflite`
-  Change `mio-tflite260` to `mio-tflite`

But current workflows are quite different and thus changes across different modules are inevitable I think.
So I was curious why `mio-tflite260` is introduced.

seanshpark:
> By the way, can I ask you why mio-tflite260 is introduced?

Your thoughts are similar what I'm working on. `mio-tflite` may exist if there exist modules use it.

> current workflows are quite different 

What exactly are different?

llFreetimell:
> mio-tflite may exist if there exist modules use it.

As TFLite keeps backward compatibility, I think we don't need both 2.3.0 and 2.6.0 schema.
Only with 2.6.0 schema, we can support them all.

> What exactly are different?

For now, `mio-tflite260` uses 2.3.0 TF.
But if we update it to 2.6.0 later, all of current modules which are using `mio-tflite260` must change their functions to read tflite file.
So current workflow is same as "Change `mio-tflite` to use 2.6.0 and change all the functions in related modules at the same time"

seanshpark:
>  I think we don't need both 2.3.0 and 2.6.0 schema.

I will let 2.3.0 exist if needed. I think this is the main part of difference.

> But if we update it to 2.6.0 later,

Yes... 

>  must change their functions to read tflite file.

[...]

PR #7593, opened 2021-08-31
DRAFT ugrade to TF 2.6.0

seanshpark:
on-going changes to upgrade to TF 2.6.0

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #7642, opened 2021-09-09
[circle/partition] Use TF2.6.0 virt-env for value test

seanshpark:
This will revise circle-part-value-test to use TF2.6.0 virt-environment.


seanshpark:
For #7616, from draft #7593

mhs4670go (approved):
LGTM

PR #7645, opened 2021-09-09
[mio-tflite260] Introduce mio-tflite260

seanshpark:
This will introduce mio-tflite260 for providing 2.6.0 version.
As modules using mio-tflite needs to be ugpraded at once,
actual version will be changed to 2.6.0 when all dependent modules
uses mio-tflite260.


seanshpark:
For #7616, from draft #7593

mhs4670go (approved):
LGTM

PR #7587, opened 2021-08-30
[luci-interpreter] Introduce MemoryManager Interface and SimpleMemoryManager

BalyshevArtem:
This commit introduces memory manager interface and simple memory manager realization. This is first step of merging changes from pull request #7408 

issue #7355

ONE-DCO-1.0-Signed-off-by: Artem Balyshev a.balyshev@partner.samsung.com

Issue #7355, opened 2021-07-30
[luci-interpreter] Move tensor memory management to separate class


binarman:
### Goal

Introduce tensor memory manager class.
Role or this manager is to assign and free tensor memory.

Proposed interface looks like this:
```c++
class MManager
{
public:
  virtual void allocate_memory(luci_interpreter Tensor *tensor);
  virtual void release_memory(luci_interpreter Tensor *tensor);
};
```

Manager is passed to interpreter at it's(Interpreter) construction time.
For example:
``` c++
  AdvancedMManager mm;
  luci_interpreter::Interpreter interpreter(module.get(), mm);
```

`allocate_memory` and `release_memory` methods should be called in `RuntimeGraph::TensorAllocPlan::allocate(size_t)` and `RuntimeGraph::TensorAllocPlan::deallocate(size_t)` methods instead of `Tensor::allocate`/`Tensor::deallocate`.

### Rationale

Currently we can not control memory allocations, proposed design could enable static memory planning (for specific model) and different strategies specific for neural networks.

binarman:
+cc @BalyshevArtem

BalyshevArtem:
Result of tracking allocations and releases memory (with help of overloading operators new[] and delete[]) in inception_v3, mobilenet_v1_224 and mobilenet_v2_224.
Attach two variants for each network of tracing new[] / delete[] with memoryManager and without:

[mobilenet_v2_224_without_mm.txt](https://github.com/Samsung/ONE/files/7098199/mobilenet_v2_224_without_mm.txt)
[mobilenet_v2_224_with_mm.txt](https://github.com/Samsung/ONE/files/7098200/mobilenet_v2_224_with_mm.txt)
[mobilenet_v1_224_without_mm.txt](https://github.com/Samsung/ONE/files/7098201/mobilenet_v1_224_without_mm.txt)
[mobilenet_v1_224_with_mm.txt](https://github.com/Samsung/ONE/files/7098202/mobilenet_v1_224_with_mm.txt)
[inception_v3_without_mm.txt](https://github.com/Samsung/ONE/files/7098203/inception_v3_without_mm.txt)
[inception_v3_with_mm.txt](https://github.com/Samsung/ONE/files/7098204/inception_v3_with_mm.txt)

binarman:
Everything is merged, closing!

PR #7664, opened 2021-09-10
[luci-interpreter] Extend Interpreter interface with MemoryManager configuration

BalyshevArtem:
This commit extends Interpreter, RuntimeGraph, GraphLoader and ModuleLoader interfaces with MemoryManager configuration. (add it to parameter). This is second step of merging changes from pull request #7408. First step pr #7587

issue #7355

### DIFF
--- a/compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
@@ -22,6 +22,7 @@
 #include <luci/IR/Nodes/CircleInput.h>
 #include <luci/IR/Nodes/CircleOutput.h>
 
+#include "luci_interpreter/MemoryManager.h"
 #include <luci/IR/Module.h>
 
 #include <memory>
@@ -49,7 +50,7 @@ public:
 class Interpreter
 {
 public:
-  explicit Interpreter(const luci::Module *module);
+  explicit Interpreter(const luci::Module *module, IMemoryManager *memory_manager = nullptr);
 
   ~Interpreter();
 
@@ -64,7 +65,11 @@ public:
   const Tensor *getTensor(const loco::Node *node) { return _node_to_tensor[node]; }
 
 private:
+  // _default_memory_manager should be before _runtime_module due to
+  // the order of deletion in the destructor
+  std::unique_ptr<IMemoryManager> _default_memory_manager = nullptr;
   std::unique_ptr<class RuntimeModule> _runtime_module;
+  IMemoryManager *_memory_manager = nullptr;
 
   // Observer functionality support.
   std::unique_ptr<struct RuntimeToIR> _runtime_to_ir;
--- a/compiler/luci-interpreter/src/Interpreter.cpp
+++ b/compiler/luci-interpreter/src/Interpreter.cpp
@@ -77,7 +77,6 @@ Interpreter::Interpreter(const luci::Module *module,
   _event_notifier = std::make_unique<EventNotifierImpl>(*_runtime_to_ir, _observers);
   _runtime_module = std::make_unique<RuntimeModule>(_event_notifier.get());
 
-
   if (memory_manager == nullptr)
   {
     _default_memory_manager = std::make_unique<SimpleMemoryManager>();
@@ -88,7 +87,8 @@ Interpreter::Interpreter(const luci::Module *module,
     _memory_manager = memory_manager;
   }
 
-  ModuleLoader loader(module, _runtime_module.get(), *_runtime_to_ir, _node_to_tensor, _memory_manager);
+  ModuleLoader loader(module, _runtime_module.get(), *_runtime_to_ir, _node_to_tensor,
+                      _memory_manager);
   loader.load();
 }
 
--- a/compiler/luci-interpreter/src/Interpreter.cpp
+++ b/compiler/luci-interpreter/src/Interpreter.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "luci_interpreter/Interpreter.h"
+#include "luci_interpreter/SimpleMemoryManager.h"
 
 #include "loader/ModuleLoader.h"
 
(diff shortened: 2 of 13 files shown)


compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
```diff
@@ -64,7 +65,11 @@ class Interpreter
   const Tensor *getTensor(const loco::Node *node) { return _node_to_tensor[node]; }
 
 private:
+  //_default_memory_manager should be before _runtime_module due to
```
m-bronnikov, line 68: nitpick.

```suggestion
  // _default_memory_manager should be before _runtime_module due to
```

compiler/luci-interpreter/src/Interpreter.cpp
```diff
@@ -90,7 +104,9 @@ void Interpreter::writeInputTensor(const luci::CircleInput *input_node, const vo
     throw std::runtime_error("Cannot find tensor for input node named \"" + name + "\".");
   }
   if (data != nullptr)
+  {
     tensor->writeData(data, data_size);
+  }
```
m-bronnikov, line 109: This changes is redundant

compiler/luci-interpreter/src/core/RuntimeGraph.cpp
```diff
@@ -29,15 +29,22 @@ class RuntimeGraph::TensorAllocPlan
   std::vector<std::vector<Tensor *>> _alloc_plan;
   std::vector<std::vector<Tensor *>> _dealloc_plan;
   bool _valid = false;
+  IMemoryManager *_memory_manager;
```
m-bronnikov, line 32: Q. may be const?

compiler/luci-interpreter/src/core/RuntimeGraph.h
```diff
@@ -51,6 +52,7 @@ class RuntimeGraph
   void execute() const;
 
 private:
+  IMemoryManager *_memory_manager;
```
m-bronnikov, line 55: Same question about `const` =)

compiler/luci-interpreter/src/loader/GraphLoader.h
```diff
@@ -42,6 +44,7 @@ class GraphLoader
   const loco::Graph *_graph;
   RuntimeGraph *_runtime_graph;
   RuntimeToIR &_runtime_to_ir;
+  IMemoryManager *_memory_manager;
```
m-bronnikov, line 47: Ditto)

compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
```diff
@@ -64,7 +65,11 @@ class Interpreter
...
+  // the order of deletion in the destructor
+  std::unique_ptr<IMemoryManager> _default_memory_manager = nullptr;
```
m-bronnikov, line 70: (optional) I think it would be nice to make `_memory_manager` as `unique_ptr` and remove `_default_memory_manager` member, but this is requires a lot of changes in this and next PR's, so, I'm okay with current design =)

m-bronnikov (commented):
Itself looks good =)

compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
BalyshevArtem: The main idea why we have two memory_managers is that we want that some tools (like EvalDriver) could have own memory manager and interpreter can use it. But if tool hasn't memory manager interpreter creates default_memory_manager and uses it.

compiler/luci-interpreter/src/core/RuntimeGraph.cpp
BalyshevArtem: I think that in the future we may have some kind of memory manager that can change, for example, collects some statistics. And that is why it can be better if we remain this _memory_manager as not constant ptr.

compiler/luci-interpreter/include/luci_interpreter/Interpreter.h
binarman: Let me add some details:
We want to have several variants of interpreter interfaces:
- Old simple interface, when you create interpreter and good to go
- New customizable interface which can be used if you want to adjust behavior for specific use case

To reduce code duplication while adding new options (like custom memory manager in this PR) we can implement customizable implementation and define default values for options.

Problem with memory manager is following:
- If simplified interface is used, we need to create memory manager inside interpreter, so luci-interpreter owns memory manager and should delete it.
- If full interface is used, memory manager is created outside of interpreter and could be shared between multiple interpreters. luci-interpreter does not own this memory manager and should not delete it.

`_default_memory_manager` is needed in case we use simplified interface and need to remove MM with luci-interpreter.

m-bronnikov: Ah, I got it! =)

But I imagined following desing:
1. If we use `Interpreter` and would like to provide own memory manager, we need to create manger wrapped to `unique_ptr` like:
```cpp
auto manager = std::make_unique<CustomManager>();
auto const *manager_ptr = manager.get(); // if we need call some const managers methods later
```
2. Then,  **move** unique_ptr to Interpreter's constructor (give ownership to manager). 
```cpp
auto interpreter = std::make_unique<Interpreter>(module, std::move(manager));
```
3. Constructor of Interpreter expects rvalue, `_memory_manager` is unique and manger set up will be something like:
```cpp
  if (memory_manager == nullptr)
  {
    _memory_manager = std::make_unique<SimpleMemoryManager>();
  }
  else
  {
    _memory_manager = std::move(memory_manager);
  }
```

m-bronnikov: But this is a not requirement, just a note. Current design is good to me too!

> Ah, I got it! =)
> 
> But I imagined following desing:

m-bronnikov (approved):
LGTM

compiler/luci-interpreter/src/core/RuntimeGraph.cpp
```diff
@@ -29,15 +29,22 @@ class RuntimeGraph::TensorAllocPlan
...
   void invalidate() { _valid = false; }
   bool isValid() const { return _valid; }
   void build(const RuntimeGraph &graph);
   void allocate(size_t kernel_index) const;
   void deallocate(size_t kernel_index) const;
 };
 
+RuntimeGraph::TensorAllocPlan::TensorAllocPlan(IMemoryManager *memory_manager)
+  : _memory_manager(memory_manager)
+{
```
binarman, line 45: Note:
In such cases we insert comment, like
```suggestion
{
  // Do nothing
```

binarman (approved):
LGTM

PR #7650, opened 2021-09-09
[luci-interpreter] Introduce PAL kernel wrappers

binarman:
This commit introduces wrappers for computation code to abstract it
from platform specific features, like system libraries.

related issue: #7288
draft: #7336 

ONE-DCO-1.0-Signed-off-by: Artem Balyshev <a.balyshev@partner.samsung.com>

PR #7666, opened 2021-09-10
[luci-interpreter] Add kernel test with pointwise conv2d

binarman:
This PR adds kernel unittest for conv2d pointwise operation.

Found that this case is not covered by unit tests during debug of #7650.


m-bronnikov (approved):
LGTM

PR #7668, opened 2021-09-13
[one-cmds] Enable finding backend command from system path

jinevening:
This enables to find backend command from system path.

### DIFF
--- a/compiler/one-cmds/one-codegen
+++ b/compiler/one-cmds/one-codegen
@@ -28,6 +28,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import shutil
 
 import utils as _utils
 
@@ -49,6 +50,7 @@ def _get_backends_list():
     The list where `one-codegen` finds its backends
     - `bin` folder where `one-codegen` exists
     - `backends` folder
+    - System path
 
     NOTE If there are backends of the same name in different places,
      the closer to the top in the list, the higher the priority.
@@ -150,6 +152,10 @@ def main():
     for cand in backends_list:
         if ntpath.basename(cand) == backend_base:
             codegen_path = cand
+    if not codegen_path:
+        # Find backend from system path
+        codegen_path = shutil.which(backend_base)
+
     if not codegen_path:
         raise FileNotFoundError(backend_base + ' not found')
     codegen_cmd = [codegen_path] + backend_args + unknown_args


jinevening:
Currently, backend should be placed in (1) `bin` folder where `one-codegen` exists or  (2) `backends` folder

For convenience, I'd like to enable finding backend command from the system path.

seanshpark (approved):
LGTM

seanshpark:
> For convenience, I'd like to enable finding backend command from the system path.

@jinevening , this will change(add) characteristic of backend command search algorithm
it would be better describe here about the reason for this change, not just `convenience`
but actual problem or something like.

jinevening:
> it would be better describe here about the reason for this change, not just convenience
but actual problem or something like.

When we make a release package of ONE, `one-codegen` is not linked with any backend. So, we should copy (or make a symbolic link of) the binary of the backend into the corresponding directories before using `one-codegen`. This change allows `one-codegen` to find the backend if the backend is available in the system path. This will reduce the step to use `one-codegen`.

Note that this change is backward-compatible, because the search on the system path is only conducted if the backend is not found in the `bin` directory or `backends` directory.

seanshpark:
> Note that this change is backward-compatible, because the search on the system path is only conducted if the backend is not found in the bin directory or backends directory.

I this just that I would like to leave records for this change as it's directly related with UX.
So there is no issues or problems with this change? 
This is what I can imagine.
- A may install (by manual?) a backend to some folder and add the PATH
- B may install (by manual?) a backend to some folder but forgot to add to the PATH

If you have'nt then it would be nice if you add something about this in the documents.

glistening:
I also think it would be good if it is written somewhere in ONE_GUIDE. I thought there was already an explanation for rule for finding backend driver. As I remember, at least it was there in some issue or PR. @jyoungyun Do you remember the location?

jyoungyun:
@glistening Do you mean here? We have discussed this code before.
https://github.com/Samsung/ONE/blob/c2a31b05d86250d2708cf1fa87a9a1615fd401ad/compiler/one-cmds/one-codegen#L39-L54

And `one-profile` has the same rule with `one-codegen`. If this changes is applied, `one-profile` must be modified too.

jyoungyun (approved):
LGTM

mhs4670go (approved):
LGTM

PR #7673, opened 2021-09-13
[luci] Fix unreachable loop in ExpandBroadcastConstPass

llFreetimell:
From second loop, current code is not reachable
because the function is always returned at first iteration.
This commit will fix it.

### DIFF
--- a/compiler/luci/pass/src/ExpandBroadcastConstPass.cpp
+++ b/compiler/luci/pass/src/ExpandBroadcastConstPass.cpp
@@ -133,13 +133,19 @@ bool expand_broadcast_const(luci::CircleConst *node)
     switch (circle_successor->opcode())
     {
       case luci::CircleOpcode::ADD:
-        changed = changed || expand_node_input<luci::CircleAdd>(node, circle_successor);
+        if (expand_node_input<luci::CircleAdd>(node, circle_successor))
+          changed = true;
+        break;
       case luci::CircleOpcode::MUL:
-        changed = changed || expand_node_input<luci::CircleMul>(node, circle_successor);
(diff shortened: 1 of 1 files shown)


ai-moiseev:
In initial pass PR there was a request to follow the style of other passes here: https://github.com/Samsung/ONE/pull/7603#discussion_r702479369... Maybe we can just remove `return false` at the end of the function? :)

llFreetimell:
Using `if()` instead of `= ||` is fine.
But I want to ask which one was your intention :)
Only applying to first `succ` of the node or applying to all of `succ` of the node?

ai-moiseev:
@llFreetimell Oh, it seems it expands only the first successor input, the intention was to expand all successors. Thanks for noticing this mistake :)

compiler/luci/pass/src/ExpandBroadcastConstPass.cpp
```diff
@@ -125,22 +125,25 @@ bool expand_broadcast_const(luci::CircleConst *node)
...
+  bool changed = false;
+
   for (auto successor : loco::succs(node))
   {
     auto const circle_successor = loco::must_cast<luci::CircleNode *>(successor);
     switch (circle_successor->opcode())
     {
       case luci::CircleOpcode::ADD:
-        return expand_node_input<luci::CircleAdd>(node, circle_successor);
+        changed = changed || expand_node_input<luci::CircleAdd>(node, circle_successor);
```
ai-moiseev, line 136: Missing `break` in cases?

ai-moiseev (approved):
LGTM👍

Issue #7677, opened 2021-09-13
[luci-interpreter] Support simple CNN int8 model inference


binarman:
### What?

Need to support several int8 quantized operations in luci-interpreter:
- [x] Conv2d #7680 
- [x] DepthwiseConv2d #7679
- [x] AveragePool2d #7687 
- [x] Reshape (nothing to do, already supported)
- [x] Sorfmax #7711
- [x] FullyConnected #7690

Additional changes:
- [x] Load const  #7678

### Why?

Currently tflite is moved from uint8 to int8 quantization format, so recent tflite models use int8 data type.

To compare tflite and luci interpreter performance properly we need to support this quantization schema in luci interpreter.

This issue provides list of most common operations used in CNNs. (mobilenet, and some others)

binarman:
Turned out we still need to do some adjustments to softmax

binarman:
I've tried to run person_detect model from tflite micro benchmark and it did not work yet

binarman:
checked model from https://github.com/Samsung/ONE/issues/7598#issuecomment-916753214, it works!
closing

PR #7679, opened 2021-09-13
[luci-interpreter] Support int8 depthwise conv2d

binarman:
This PR enables int8 quantizer depthwise conv2d.

Related issue #7677 

### DIFF
--- a/compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
+++ b/compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
@@ -62,7 +62,7 @@ void DepthwiseConv2D::configure()
   else if (input()->element_type() == DataType::S8 && filter()->element_type() == DataType::S8)
   {
     LUCI_INTERPRETER_CHECK(filter()->shape().num_dims() == 4);
-    LUCI_INTERPRETER_CHECK(filter()->shape().dim(3) == filter()->scales().size());
+    LUCI_INTERPRETER_CHECK(static_cast<uint32_t>(filter()->shape().dim(3)) == filter()->scales().size());
     for (auto zerop : filter()->zero_points())
     {
       LUCI_INTERPRETER_CHECK(zerop == 0);
--- a/compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
+++ b/compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
@@ -62,7 +62,8 @@ void DepthwiseConv2D::configure()
   else if (input()->element_type() == DataType::S8 && filter()->element_type() == DataType::S8)
   {
     LUCI_INTERPRETER_CHECK(filter()->shape().num_dims() == 4);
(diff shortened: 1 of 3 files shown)


compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
```diff
@@ -58,6 +59,17 @@ void DepthwiseConv2D::configure()
   {
     LUCI_INTERPRETER_CHECK(bias() == nullptr || bias()->element_type() == DataType::S32);
   }
+  else if (input()->element_type() == DataType::S8 && filter()->element_type() == DataType::S8)
+  {
+    LUCI_INTERPRETER_CHECK(filter()->shape().num_dims() == 4);
```
BalyshevArtem, line 64: This LUCI_INTERPRETER_CHECK may be redundant since there is the same check below in the code:
```
const Shape &filter_shape = filter()->shape();
LUCI_INTERPRETER_CHECK(input_shape.num_dims() == 4 && filter_shape.num_dims() == 4);
```

compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
```diff
@@ -58,6 +59,17 @@ void DepthwiseConv2D::configure()
...
+    LUCI_INTERPRETER_CHECK(filter()->shape().dim(3) == filter()->scales().size());
+    for (auto zerop : filter()->zero_points())
+    {
+      LUCI_INTERPRETER_CHECK(zerop == 0);
+    }
+    LUCI_INTERPRETER_CHECK(filter()->shape().dim(0) == 1);
```
BalyshevArtem, line 70: There is also the same check below in the code:
`LUCI_INTERPRETER_CHECK(filter_shape.dim(0) == 1);`

compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
binarman: I'll leave this check instead, because I think it is better to report the problem early

compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
binarman:
```suggestion
```

compiler/luci-interpreter/src/kernels/DepthwiseConv2D.cpp
```diff
@@ -123,6 +135,10 @@ void DepthwiseConv2D::execute() const
         evalQuantizedPerChannel();
       }
       break;
+    case DataType::S8:
+      LUCI_INTERPRETER_CHECK(filter()->shape().num_dims() == 4);
```
binarman, line 139:
```suggestion
```

BalyshevArtem (dismissed):
LGTM

m-bronnikov (approved):
LGTM

BalyshevArtem (approved):

Issue #7682, opened 2021-09-14
[circle-quantizer] Support manual writing of quantization parameters


jinevening:
## What

Let's enable a new feature that manually writes quantization parameters (scale, zerop) to specified tensors.

Target tensors and scale/zerop are specified as `--force_quantparam <tensor_name> <scale> <zerop>`. Multiple tensors can be updated by repeating the interface. 

For example,

input model: input.circle
output model: output.circle 

Write scale = 1.0, zerop = 0 to the tensor named Input
```
circle-quantizer --force_quantparam Input 1.0 0 input.circle output.circle
```

Write scale = 1.0, zerop = 0 to the tensor named Input and scale = 2.0, zerop = 1 to the tensor named Input2
```
circle-quantizer --force_quantparam Input 1.0 0 --force_quantparam Input2 2.0 1 input.circle output.circle
```

CC @lemmaa

jinevening:
This feature will only support writing per-tensor qparams (not per-channel).

seanshpark:
@jinevening , tensorname may include `;` character... have to tested with this?

jinevening:
> @jinevening , tensorname may include ; character... have to tested with this?

I used `;` as a delimiter in the draft, but now `tensor_name` is saved in a vector, so it is ok for `tensor_name` to have `;`.

jinevening:
All done.

PR #7685, opened 2021-09-14
[luci] Add ForceQuantParam pass to luci

jinevening:
This adds ForceQuantParam pass to luci.

---
Related to: #7682
Draft PR: #7683

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/ForceQuantParamPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ForceQuantParamPass.h
@@ -21,10 +21,6 @@
 
 #include <logo/Pass.h>
 
-using TensorVector = std::vector<std::string>;
-using ScaleVector = std::vector<float>;
-using ZPVector = std::vector<int64_t>;
-
 namespace luci
 {
 
@@ -33,6 +29,11 @@ namespace luci
  */
 class ForceQuantParamPass : public logo::Pass
 {
+public:
+  using TensorVector = std::vector<std::string>;
+  using ScaleVector = std::vector<float>;
+  using ZPVector = std::vector<int64_t>;
+
 public:
   ForceQuantParamPass(TensorVector &tensors, ScaleVector &scales, ZPVector &zerops)
     : _tensors{tensors}, _scales{scales}, _zerops{zerops}
--- a/compiler/luci/pass/include/luci/Pass/ForceQuantParamPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ForceQuantParamPass.h
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/include/luci/Pass/ForceQuantParamPass.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef __LUCI_FORCE_QUANT_PARAM_PASS_H__
+#define __LUCI_FORCE_QUANT_PARAM_PASS_H__
+
+#include <loco.h>
+
+#include <logo/Pass.h>
+
+using TensorVector = std::vector<std::string>;
+using ScaleVector = std::vector<float>;
+using ZPVector = std::vector<int64_t>;
```
seanshpark, line 26: These names, especially `TensorVector` is generic that may slip in files that include this.
How about using in another namespace,... maybe `luci::pass` ?

compiler/luci/pass/src/ForceQuantParamPass.cpp
```diff
@@ -0,0 +1,77 @@
...
+    _zerops.erase(_zerops.begin() + index);
+  }
+
+  if (_tensors.size() > 0)
+  {
+    std::string msg;
+    for (auto const &t : _tensors)
+      msg += "Tensor does not exist: " + t + ".\n";
+    msg += "Please check tensor name.\n";
+    throw std::runtime_error(msg);
```
seanshpark, line 70: This will stop the optimization pass.
But it would be better to stop if the target tensor is not found...

Maybe provide this Pass as a separate command line tool?

seanshpark: > Maybe provide this Pass as a separate command line tool?

This may not intergrate with one-build/oncc... -_-;;;

seanshpark: Please leave a note to explain why this Pass has to throw.
All other Pass never has thrown exception.
If possible we can print unfiltered log message if this is not that critical.

jinevening: > Please leave a note to explain why this Pass has to throw.
> All other Pass never has thrown exception.
> If possible we can print unfiltered log message if this is not that critical.

It seems that there are two options to handle the case when a user gives a wrong tensor name.
(1) Throw an exception and do not produce output model.
(2) Update only tensors with correct names and produce output model (+ Print logs for wrong tensor names).

I think the first option is better, because it is easier to detect whether a user gives a wrong input. For the second option, it will be difficult to detect the wrong input if a user misses the log (or the program is run on CI machine), .

This pass is different from a normal optimization pass. This pass has its own functionality, like QuantizeWithMinMaxPass or ConvetNCHWToNHWCPass, which throws an exception.

> This may not intergrate with one-build/oncc...

This pass will be integrated with `one-quantize` (the codes are in #7683). The usage is as below.

```
one-quantize --force_quantparam --tensor_name Input --scale 1.0 --zero_point 3 -i input.circle -o output.circle
```

compiler/luci/pass/include/luci/Pass/ForceQuantParamPass.h
jinevening: I've moved them inside `ForceQuantParamPass`, because they are used only in this pass.

seanshpark (approved):
LGTM

PR #7687, opened 2021-09-14
[luci-interpreter] Support int8 AveragePool2D

BalyshevArtem:
This pr enables int8 AveragePool2D operation.
Related issue: #7677

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Utils.cpp
+++ b/compiler/luci-interpreter/src/kernels/Utils.cpp
@@ -91,7 +91,7 @@ static void calculateActivationRangeQuantizedImpl(Activation activation, int32_t
 void calculateActivationRangeQuantized(Activation activation, const Tensor *output,
                                        int32_t *activation_min, int32_t *activation_max)
 {
-  // For now, assume that signed type implies signed symmetric quantization.
+  assert(output->zero_points().size() == 1);
   int32_t qmin{};
   int32_t qmax{};
   switch (output->element_type())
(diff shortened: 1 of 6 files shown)


binarman (dismissed):
LGTM

binarman (approved):

binarman:
@nnfw-bot test nncc-release

compiler/luci-interpreter/src/kernels/Utils.cpp
```diff
@@ -91,7 +91,7 @@ static void calculateActivationRangeQuantizedImpl(Activation activation, int32_t
 void calculateActivationRangeQuantized(Activation activation, const Tensor *output,
                                        int32_t *activation_min, int32_t *activation_max)
 {
-  // For now, assume that signed type implies signed symmetric quantization.
+  assert(output->zero_points().size() == 1);
```
m-bronnikov, line 94: As I understand, this assert comes from `output->zero_point()` method and we adds this assert here, because removes `assert(output->zero_point() == 0)` below.

m-bronnikov (approved):
LGTM

compiler/luci-interpreter/src/kernels/Utils.cpp
binarman: Kind of:
- removed assert checked that there are one zero point AND it is equal to 0.
- now we check that there are only one zero point (and doing it for all kinds of activations, because this is how all existing kernels working) :slightly_smiling_face:

PR #7690, opened 2021-09-14
[luci-interpreter] Support int8 FullyConnected

BalyshevArtem:
This pr enables int8 FullyConnected operation.

Related issue: #7677

### DIFF
--- a/compiler/luci-interpreter/src/kernels/FullyConnected.cpp
+++ b/compiler/luci-interpreter/src/kernels/FullyConnected.cpp
@@ -19,6 +19,7 @@
 #include "kernels/Utils.h"
 
 #include <tensorflow/lite/kernels/internal/reference/fully_connected.h>
+#include <tensorflow/lite/kernels/internal/reference/integer_ops/fully_connected.h>
 
 #include <stdexcept>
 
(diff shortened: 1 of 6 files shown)


binarman:
Linking problems, maybe try to rebase on master/older revision

SlavikMIPT:
@nnfw-bot test nncc-release

compiler/luci-interpreter/src/kernels/FullyConnected.cpp
```diff
@@ -77,6 +84,9 @@ void FullyConnected::execute() const
     case DataType::U8:
       evalQuantized();
```
m-bronnikov, line 85: How about make `evalQuantized()` as template function for types with assymetric quantization like `int8` and `uint8`? =)

If so, we could reduce code duplication and use this method as:
```cpp
switch (input()->element_type())
  {
    case DataType::U8:
      evalQuantized<uint8_t>();
      break;
    case DataType::S8:
      evalQuantized<int8_t>();
      break;
    case DataType::FLOAT32:
      evalFloat();
      break;
```

m-bronnikov (approved):
LGTM

compiler/luci-interpreter/src/kernels/FullyConnected.cpp
binarman: This technique could be used for other operations, like Conv, DepthwiseConv, etc. Let's consider this refactoring after enabling minimal set (from #7677) of operations.

compiler/luci-interpreter/src/kernels/FullyConnected.test.cpp
```diff
@@ -48,6 +48,39 @@ void Check(std::initializer_list<int32_t> input_shape, std::initializer_list<int
...
+template <>
+void Check<int8_t>(std::initializer_list<int32_t> input_shape,
+                   std::initializer_list<int32_t> weights_shape,
+                   std::initializer_list<int32_t> bias_shape,
+                   std::initializer_list<int32_t> output_shape,
+                   std::initializer_list<float> input_data,
+                   std::initializer_list<float> weights_data,
+                   std::initializer_list<float> bias_data, std::initializer_list<float> output_data)
+{
+  const float quantized_tolerance = getTolerance(-127, 128, 255);
```
binarman, line 60: nitpick =)
minimum/maximum for int8 type is -128/127, not -127/128

binarman (approved):
LGTM

PR #7697, opened 2021-09-15
[luci] Set origin info for constants during import

ai-moiseev:
This commit adds origin information based on node name during constant value import.


seanshpark (dismissed):
LGTM

llFreetimell:
@ai-moiseev 

Maybe you created this PR because you thought that not adding origin to constant nodes is mistake, right? :)
But it was intentional :D

As we are treating buffer values as `CircleConst`, it seems that `Const` is also one of operation kinds.
However, if you look at schema file, they are not operation actually @.@
And the constant values are not "executed", just "used".

So IMHO, this PR is not needed :)

seanshpark:
> So IMHO, this PR is not needed :)

Maybe add a comment no origin is intentional...

ai-moiseev:
I've had an idea that every node that can contain a name can contain an origin too, including constants (to be able to follow model changes such as constant folding during debugging). But origin info can be retrieved as fallback during debug output on some other level probably...

I've changed this PR to contain only note commentary. :)

seanshpark (approved):
LGTM

PR #7703, opened 2021-09-16
[tflchef] Use bracket for flatbuffers include

seanshpark:
This will revise to use bracket for flatbuffers include.


mhs4670go (approved):
LGTM

Issue #7700, opened 2021-09-16
Compiler FE: Add version 1.10 to flatbuffers


seanshpark:
Let's revise flatbuffers cmake target name and others to have ver 1.10 suffix.

While upgrading to TF2.6.0 flatbuffers 1.10 and 1.12 should co-exist.
And some modules may still need to use 1.10.

seanshpark:
For #7616, #7688

seanshpark:
done

PR #7705, opened 2021-09-16
[enco/tflite] Revise to use flatbuffers-1.10

seanshpark:
This will revise to use with version flatbuffres-1.10.


seanshpark:
For #7700, from draft #7593

compiler/enco/frontend/tflite/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-nnas_find_package(FlatBuffers QUIET)
+nnas_find_package(FlatBuffers EXACT 1.10 QUIET)
```
seanshpark, line 1: Just to check if this is possible now...

seanshpark: Seems OK but let's remove this for now.

mhs4670go (approved):
LGTM

Issue #7688, opened 2021-09-14
[luci] libluci_pass fails to link with libflatbuffers.a


seanshpark:
This was found and seemed fixed but appeared again while working on upgrade to TF2.6.0

Build: Release version, local build, Ubuntu 18.04

> /usr/bin/ld: ../../../overlay/lib/libflatbuffers.a(util.cpp.o): relocation R_X86_64_PC32 against symbol `_ZTVSt9basic_iosIcSt11char_traitsIcEE@@GLIBCXX_3.4' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
compiler/luci/pass/CMakeFiles/luci_pass.dir/build.make:1000: recipe for target 'compiler/luci/pass/libluci_pass.so' failed

Situation
- TF2.6.0 related uses FlatBuffers 1.12
- Existing uses FlatBuffers 1.10
- Debug version is OK
- CI build both debug/release seems OK

seanshpark:
This is another topic...
- current build only permits only one libflatbuffers.a and flatc binary in `build/overlay` folder.

seanshpark:
Related change: #7672

seanshpark:
> current build permits only one libflatbuffers.a and flatc binary in build/overlay folder.

--> need to check if this is the cause; we may need to make 1.10 and 1.12 co-exist...

seanshpark:
Close as error seems gone

PR #7706, opened 2021-09-16
[luci/pass] Add link to flatbuffers-1.10

seanshpark:
This will explictly add link to flatbuffers-1.10.


seanshpark:
From draft #7593

mhs4670go:
@seanshpark luci/pass doesn't have the link in the first place. Does it become to be needed?

seanshpark:
> luci/pass doesn't have the link in the first place. Does it become to be needed?

flatbuffers files was used in some Pass (you can see them previous PR)
but this was implictly OK as with `mio-circle`.
I had some trouble with while upgrading to TF2.6.0: #7688
To overcome this problem (I don't fully understand this) I had to explictly give flatbuffers version.

mhs4670go (approved):
LGTM

PR #7707, opened 2021-09-16
[pota-value-test] Require mio-circle module

seanshpark:
This will revise to require mio-circle module.


seanshpark:
From draft #7593

mhs4670go (approved):
LGTM

PR #7708, opened 2021-09-16
[tfldump] Revise to use flatbuffers-1.10

seanshpark:
This will revise to use with version flatbuffres-1.10.


seanshpark:
For #7700, from draft #7593

mhs4670go (approved):
LGTM

PR #7696, opened 2021-09-15
[luci-interpreter] Adjust tests to new memory manager

BalyshevArtem:
This pr adjusts kernels tests to use memory manager and adds TestMemoryManager for this. This is third step of merging changes from pull request #7408. First step pr #7587, second pr #7664.

issue #7355

ONE-DCO-1.0-Signed-off-by: Artem Balyshev <a.balyshev@partner.samsung.com>

PR #7716, opened 2021-09-17
[luci-interpreter] Switch interpreter tensor allocations to mem_manager

BalyshevArtem:
This pr switches interpreter tensor allocations to memory manager and fixes FullyConnected.test and AveragePool2D.test. This is fourth step of merging changes from pull request #7408. First step pr #7587, second pr #7664, third pr #7696.

issue #7355


compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
```diff
@@ -118,15 +115,16 @@ class Tensor
 
   template <typename T> const T *data() const
   {
-    assert(_data_allocated);
-    return reinterpret_cast<const T *>(_data.get());
+    static_assert(std::is_same<uint8_t, char>::value or
```
binarman, line 118: @SlavikMIPT 
If I recall it correctly, you wanted this check.
Q) Isn't it enough to check if `uint8_t == unsigned char` here?

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
```diff
@@ -118,15 +115,16 @@ class Tensor
...
+                  std::is_same<uint8_t, unsigned char>::value);
+    return reinterpret_cast<const T *>(_data);
   }
 
   template <typename T> T *data()
   {
-    if (!_data_allocated)
-      allocate();
-    return reinterpret_cast<T *>(_data.get());
+    static_assert(std::is_same<uint8_t, char>::value or
```
binarman, line 125: ditto

compiler/luci-interpreter/src/core/Tensor.cpp
```diff
@@ -68,10 +53,6 @@ void Tensor::writeData(const void *data_ptr, size_t data_size)
   std::memcpy(data<void>(), data_ptr, data_size);
 }
 
-void Tensor::resize(const Shape &new_shape)
```
binarman, line 71: Q) is data allocated at this point?
Maybe add an assert?

BalyshevArtem: As I understand, at this method data can be not allocated. And what assert can be good at this method?

binarman: I meant something like this:
``` c++
assert(!_data_allocated);
```

this is a proposal, not requirement (I think we need to check this locally first :slightly_smiling_face: )

m-bronnikov: Q. What happens if at this point data is allocated and new shape is more than old? I affraid that user of `Tensor` may expect that tensor contain buffer with size `shape.num_elements() * sizeof(datatype)`  but tensor not has enought space.

BalyshevArtem: But sometimes we can call this method, when data already allocated. For example in kernel Concatenation test. This assert will fail at this test.

> I meant something like this:
> 
> ```c++
> assert(!_data_allocated);
> ```
> 
> this is a proposal, not requirement (I think we need to check this locally first 🙂 )

BalyshevArtem: In this situation user should call allocate_memory method. And after that old buffer will release and then memory manager will allocate new buffer for the new shape.

> Q. What happens if at this point data is allocated and new shape is more than old? I affraid that user of `Tensor` may expect that tensor contain buffer with size `shape.num_elements() * sizeof(datatype)` but tensor not has enought space.

compiler/luci-interpreter/src/kernels/While.cpp
```diff
@@ -48,6 +48,12 @@ void copy(const std::vector<Tensor *> &src, const std::vector<Tensor *> &dst)
     const_src.push_back(t);
   copy(const_src, dst);
 }
+// TODO: Think about how allocate memory for output in main graph
```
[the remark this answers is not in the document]
BalyshevArtem: Done

m-bronnikov (approved):
LGTM

compiler/luci-interpreter/src/core/Tensor.cpp
m-bronnikov: (optional) It would be nice to have comment with statement why this method is safe =)

binarman (approved):
LGTM

PR #7718, opened 2021-09-20
[luci] Remove name length assert

m-bronnikov:
This commit removes unnecessary length `assert` in SubstituteSqueezeToReshape pass.

### DIFF
--- a/compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
+++ b/compiler/luci/pass/src/SubstituteSqueezeToReshapePass.cpp
@@ -136,7 +136,6 @@ bool substitute_squeeze_to_reshape(luci::CircleSqueeze *squeeze)
     throw std::runtime_error("Invalid values in squeeze_dims: " + squeeze->name());
 
   auto name = squeeze->name();
-  assert(name.length() > 0);
 
   auto reshape_shape = node_shape(squeeze);
   auto graph = squeeze->graph();


m-bronnikov:
I would like to remove this assert because if `name` of Squeeze is not configured this pass still valid.  For real models this assert doesn' t trigger, but makes testing inconvenient.

seanshpark:
@m-bronnikov , it would be better to give the node a name. there are tools that depend on name of a node.

m-bronnikov:
> @m-bronnikov , it would be better to give the node a name. there are tools that depend on name of a node.

m-bronnikov:
Sorry, closed by accident =(

m-bronnikov:
> @m-bronnikov , it would be better to give the node a name. there are tools that depend on name of a node.

I agree that it is better to provide a name for the operation, but the name existance is the necessary condition for the pass to work correctly?

seanshpark:
> but the name existance is the necessary condition for the pass to work correctly?

I don't think it is.

But there are codes for the new node to assign a name, which uses the name.
```cxx
  auto reshape = graph->nodes()->create<luci::CircleReshape>();
  auto shape_const = create_shape_const(graph, reshape_shape);
  copy_quantparam(reshape, squeeze);
  reshape->name(name + "/Reshape");
  luci::add_origin(reshape, luci::get_origin(squeeze));
  shape_const->name(name + "/Reshape/shape");
```
So, it is not necessary but I think it's better to have a check.

I think it would be better to check why this node has not a name and fix that.

seanshpark:
```
auto shape_const = create_shape_const(graph, reshape_shape);
```
this `create_shape_const()` creates a Const node that does not have a name...

I think it would be better to fix this node to have a name.

seanshpark:
FYI, there are assert checks for node names in luci import/export...

m-bronnikov:
Thank you for clarifying! So, let's close.

Issue #7571, opened 2021-08-24
[luci-interpreter] Design and implement Buddy memory manager


binarman:
### Goal

Implement [Buddy memory manager](https://en.wikipedia.org/wiki/Buddy_memory_allocation) for luci-interpreter using interfaces introduced in #7355

### Rationale

If compiled for MCU, luci-interpreter requires third party implementation of `new/delete` and `malloc/free` functions.
Buddy memory manager provides basic memory management for such cases.

binarman:
memory manager is merged, thank you!

PR #7720, opened 2021-09-21
[luci-interpreter] Introduce BuddyMemoryManager

BalyshevArtem:
This pr adds BuddyMemoryManager to luci_interpreter. This manager is implementation of buddy memory allocator(https://en.wikipedia.org/wiki/Buddy_memory_allocation).

Related issue: #7571

### DIFF
--- a/compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
@@ -33,9 +33,10 @@ private:
   struct Block
   {
     Block *next_free;
-    Block *self;
     bool is_free;
     uint32_t size;
+    // debug field
+    Block *self;
   };
 
   Block *_start_block;
@@ -43,7 +44,7 @@ private:
   uint32_t _size;
   Block *_free_blocks[32]{};
 
-  static int32_t powOf2(uint32_t val)
+  static int32_t lowerLog2(uint32_t val)
   {
     int32_t i = 0;
     while (val >>= 1)
@@ -61,7 +62,7 @@ private:
     _free_blocks[l] = block;
   }
 
-  void removeFromBlocks(Block *block, int32_t l)
+  void removeFromBlocks(const Block *block, int32_t l)
   {
     if (!block)
       return;
@@ -77,18 +78,22 @@ private:
     while (tmp)
     {
       if (tmp->next_free == block)
+      {
         tmp->next_free = block->next_free;
+        return;
+      }
 
       tmp = tmp->next_free;
     }
   }
 
-  Block *divideBlock(Block *block, int32_t l)
+  void divideBlock(Block *block, int32_t l)
   {
     int32_t size = ((block->size + sizeof(Block)) / 2) - sizeof(Block);
 
     removeFromBlocks(block, l);
 
+    // there is no need to add to the free_blocks list here
     block->is_free = true;
     block->size = size;
     block->self = block;
@@ -100,18 +105,16 @@ private:
     buddy->self = buddy;
 
     addToBlocks(buddy, l - 1);
-
-    return block;
   }
 
   Block *mergeBlock(Block *block)
   {
     Block *buddy;
 
-    int32_t l = powOf2(block->size + sizeof(Block));
+    const int32_t l = lowerLog2(block->size + sizeof(Block));
 
-    int64_t address = ((uint8_t *)block - (uint8_t *)_start_block);
-    buddy = (Block *)((address ^= (1 << l)) + (size_t)_start_block);
+    const int64_t address = ((uint8_t *)block - (uint8_t *)_start_block);
+    buddy = (Block *)((address ^ (1 << l)) + (uint8_t *)_start_block);
 
     if (!buddy->is_free || buddy->size != block->size)
       return nullptr;
--- a/compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
@@ -1,0 +1,141 @@
+/* Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci_interpreter/MemoryManager.h"
+
+#ifndef LUCI_INTERPRETER_BUDDY_MEMORY_MANAGER_H
+#define LUCI_INTERPRETER_BUDDY_MEMORY_MANAGER_H
+
+namespace luci_interpreter
+{
+
+class BuddyMemoryManager : public IMemoryManager
+{
+public:
+  BuddyMemoryManager(uint8_t *memory_start, int32_t memSize);
+
+  void allocate_memory(luci_interpreter::Tensor &tensor) final;
+  void release_memory(luci_interpreter::Tensor &tensor) final;
+
+private:
+  struct Block
+  {
+    Block *next_free;
+    Block *self;
+    bool is_free;
+    uint32_t size;
+  };
+
+  Block *_start_block;
+  int32_t _num_blocks;
+  uint32_t _size;
+  Block *_free_blocks[32]{};
+
+  static int32_t powOf2(uint32_t val)
+  {
+    int32_t i = 0;
+    while (val >>= 1)
+      i++;
+
+    return i;
+  }
+
+  void addToBlocks(Block *block, int32_t l)
+  {
+    if (!block)
+      return;
+
+    block->next_free = _free_blocks[l];
+    _free_blocks[l] = block;
+  }
+
+  void removeFromBlocks(Block *block, int32_t l)
+  {
+    if (!block)
+      return;
+
+    Block *tmp = _free_blocks[l];
+
+    if (block == tmp)
+    {
+      _free_blocks[l] = block->next_free;
+      return;
+    }
+
+    while (tmp)
+    {
+      if (tmp->next_free == block)
+        tmp->next_free = block->next_free;
+
+      tmp = tmp->next_free;
+    }
+  }
+
+  Block *divideBlock(Block *block, int32_t l)
+  {
+    int32_t size = ((block->size + sizeof(Block)) / 2) - sizeof(Block);
+
+    removeFromBlocks(block, l);
+
+    block->is_free = true;
+    block->size = size;
+    block->self = block;
+
+    Block *buddy;
+    buddy = (Block *)((uint8_t *)block + sizeof(Block) + size);
+    buddy->is_free = true;
+    buddy->size = size;
+    buddy->self = buddy;
+
+    addToBlocks(buddy, l - 1);
+
+    return block;
+  }
+
+  Block *mergeBlock(Block *block)
+  {
+    Block *buddy;
+
+    int32_t l = powOf2(block->size + sizeof(Block));
+
+    int64_t address = ((uint8_t *)block - (uint8_t *)_start_block);
+    buddy = (Block *)((address ^= (1 << l)) + (size_t)_start_block);
+
+    if (!buddy->is_free || buddy->size != block->size)
+      return nullptr;
+
+    if (block > buddy)
+    {
+      Block *x = block;
+      block = buddy;
+      buddy = x;
+    }
+
+    removeFromBlocks(block, l);
+    removeFromBlocks(buddy, l);
+
+    block->size = block->size * 2 + sizeof(Block);
+    block->is_free = true;
+    block->self = block;
+
+    addToBlocks(block, l + 1);
+
+    return block;
+  }
+};
+
+} // namespace luci_interpreter
+
+#endif // LUCI_INTERPRETER_BUDDY_MEMORY_MANAGER_H
--- a/compiler/luci-interpreter/src/BuddyMemoryManager.cpp
+++ b/compiler/luci-interpreter/src/BuddyMemoryManager.cpp
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
```diff
@@ -0,0 +1,141 @@
...
+    bool is_free;
+    uint32_t size;
+  };
+
+  Block *_start_block;
+  int32_t _num_blocks;
+  uint32_t _size;
+  Block *_free_blocks[32]{};
+
+  static int32_t powOf2(uint32_t val)
```
m-bronnikov, line 46: Better to name this function `upper_log2` because this func returns `(val - 1) & val == 0 ? int(log2(val)) : int(log2(val)) + 1`.

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  Block *mergeBlock(Block *block)
+  {
+    Block *buddy;
+
+    int32_t l = powOf2(block->size + sizeof(Block));
+
+    int64_t address = ((uint8_t *)block - (uint8_t *)_start_block);
+    buddy = (Block *)((address ^= (1 << l)) + (size_t)_start_block);
```
m-bronnikov, line 114: As I see, we do not need to change `address` variable:

```suggestion
    buddy = (Block *)((address ^ (1 << l)) + (size_t)_start_block);
```

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
```diff
@@ -0,0 +1,141 @@
...
+    return block;
+  }
+
+  Block *mergeBlock(Block *block)
+  {
+    Block *buddy;
+
+    int32_t l = powOf2(block->size + sizeof(Block));
+
+    int64_t address = ((uint8_t *)block - (uint8_t *)_start_block);
```
m-bronnikov, line 113: May be `const` =)

m-bronnikov: `l` valiable ditto.

m-bronnikov:
Science `Block` - helper struct for algorithm implementation, how about moving this struct with `divideBlock`, `removeFromBlocks`, `addToBlocks`, `mergeBlock` and `powOf2` functions to anonimus namespace of *BuddyMemoryManager.cpp* file.

compiler/luci-interpreter/src/CMakeLists.txt
```diff
@@ -25,14 +25,22 @@ message(STATUS "LUCI INTERPTER INITALIZED")
...
+        "${LUCI_INTERPRETER_INCLUDE_DIR}/luci_interpreter/TestMemoryManager.h" TestMemoryManager.cpp
+        "${LUCI_INTERPRETER_INCLUDE_DIR}/luci_interpreter/BuddyMemoryManager.h" BuddyMemoryManager.cpp)
 
 if (NOT LUCI_INTERPRETER_STATIC)
   add_library(${LUCI_INTERPRETER_BINARY} SHARED ${SOURCES})
 else ()
   add_library(${LUCI_INTERPRETER_BINARY} STATIC ${SOURCES})
 endif ()
 
+nnas_find_package(GTest REQUIRED)
```
binarman, line 37: Could you move this to the bottom of CMakeLists.txt and add condition?
``` c++
if(NOT ENABLE_TEST)
  return()
endif(NOT ENABLE_TEST)
```

compiler/luci-interpreter/src/BuddyMemoryManager.test.cpp
```diff
@@ -0,0 +1,69 @@
...
+namespace luci_interpreter
+{
+namespace
+{
+
+using namespace testing;
+
+TEST(BuddyMemoryManager, BuddyMemoryManager)
+{
+  uint8_t *memPool = new uint8_t[200];
```
binarman, line 29: 1. Use lower case with underscore in local variable names
2. Use unique_ptr or simple array, current implementation will leak:
``` c++
  auto mem_pool = std::make_unique<uint8_t[]>(200);
```
``` c++
  uint8_t mem_pool[200];
```
P.s. first option is preferred if you want to allocate large memory pools.

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
binarman: @m-bronnikov interesting article, by the way =)

BalyshevArtem: Done

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
BalyshevArtem: Done

compiler/luci-interpreter/src/CMakeLists.txt
BalyshevArtem: Done

compiler/luci-interpreter/src/BuddyMemoryManager.test.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <gtest/gtest.h>
+
+namespace luci_interpreter
+{
+namespace
+{
+
+using namespace testing;
+
+TEST(BuddyMemoryManager, BuddyMemoryManager)
```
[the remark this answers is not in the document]
BalyshevArtem: Done

compiler/luci-interpreter/src/BuddyMemoryManager.test.cpp
BalyshevArtem: Done

BalyshevArtem:
> **(optional)** Science `Block` - helper struct for algorithm implementation, how about moving this struct with `divideBlock`, `removeFromBlocks`, `addToBlocks`, `mergeBlock` and `powOf2` functions from class definition to anonimus namespace of _BuddyMemoryManager.cpp_ file.

this will be difficult to do since I use class variables in these functions and they are part of the class definition.

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
binarman: > upper_log2

If I understand what this function is doing, it is **lower**_log2

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
```diff
@@ -0,0 +1,141 @@
...
+  BuddyMemoryManager(uint8_t *memory_start, int32_t memSize);
+
+  void allocate_memory(luci_interpreter::Tensor &tensor) final;
+  void release_memory(luci_interpreter::Tensor &tensor) final;
+
+private:
+  struct Block
+  {
+    Block *next_free;
+    Block *self;
```
binarman, line 36: If I understand it right, this field is just for debug.
Could you add a comment?

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
BalyshevArtem: > > upper_log2
> 
> If I understand what this function is doing, it is more look like **lower**_log2

Yes, you are right, thank you!

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
```diff
@@ -0,0 +1,141 @@
...
+    if (block == tmp)
+    {
+      _free_blocks[l] = block->next_free;
+      return;
+    }
+
+    while (tmp)
+    {
+      if (tmp->next_free == block)
+        tmp->next_free = block->next_free;
```
binarman, line 80: Maybe return right after remove?
If I understand algorithm, there is no need to continue search after we found the block

compiler/luci-interpreter/src/BuddyMemoryManager.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+namespace luci_interpreter
+{
+
+BuddyMemoryManager::BuddyMemoryManager(uint8_t *memory_start, int32_t memSize)
+{
+  int32_t p = powOf2(memSize);
+  memSize = 1 << p;
+
+  _start_block = (Block *)memory_start;
```
binarman, line 27: better to use c++ style casts, they have well defined use cases, so they are easier to read and understand (except cases with multiple casts in a row =_=).
```suggestion
  _start_block = reinterpret_cast<Block *>(memory_start);
```

compiler/luci-interpreter/src/BuddyMemoryManager.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+  for (auto &_free_block : _free_blocks)
+    _free_block = nullptr;
+
+  addToBlocks(_start_block, p);
+}
+
+void BuddyMemoryManager::allocate_memory(luci_interpreter::Tensor &tensor)
+{
+  tensor.set_data_buffer(nullptr);
```
binarman, line 42: redundant?

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
```diff
@@ -0,0 +1,141 @@
...
+
+  Block *divideBlock(Block *block, int32_t l)
+  {
+    int32_t size = ((block->size + sizeof(Block)) / 2) - sizeof(Block);
+
+    removeFromBlocks(block, l);
+
+    block->is_free = true;
+    block->size = size;
+    block->self = block;
```
binarman, line 94: do we need to add block to list of free blocks?

binarman: I suppose, no, so I propose to add a comment to highlight this.
Also, do maybe remove `block->is_free = true;` line?

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
BalyshevArtem: > Also, maybe replace `Block *_start_block;` with `uint8_t *buffer;`?
> I noticed that Block type property of _start_block is actually unused in algorithm, plus this will remove several type conversions,

Property of _start_block is used in BuddyMemoryManager initializer, when we init _start_block and then pass it to addToBlocks method.

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
BalyshevArtem: Yes, we shouldn't add block to list of free blocks. And in my understanding, we should leave the line `block->is_free = true;` because (how I understand algorithm) after calling `removeFromBlocks(block, l);` method current block should be marked as free block.

compiler/luci-interpreter/src/BuddyMemoryManager.cpp
BalyshevArtem: done

compiler/luci-interpreter/src/BuddyMemoryManager.cpp
BalyshevArtem: done

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
BalyshevArtem: done

binarman (approved):
LGTM

binarman:
@BalyshevArtem 
> "how about moving this struct with divideBlock, removeFromBlocks, addToBlocks, mergeBlock and powOf2 functions from class definition to anonimus namespace of BuddyMemoryManager.cpp file."

I think we can do it, let's see at this in another PR =)

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
m-bronnikov: It looks like `block` is already free before `divideBlock` function call because we can divide only free blocks. Also, `removeFromBlocks` function not changes `is_free` status of block, so, this setting looks redundant. But, personally I would like to stay it here because it increase code readability.

compiler/luci-interpreter/src/BuddyMemoryManager.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  addToBlocks(_start_block, p);
+}
+
+void BuddyMemoryManager::allocate_memory(luci_interpreter::Tensor &tensor)
+{
+  const size_t element_size = getDataTypeSize(tensor.element_type());
+  const int32_t num_elements = tensor.shape().num_elements();
+  auto size = num_elements * element_size;
+  int32_t l = lowerLog2(size + sizeof(Block)) + 1;
```
m-bronnikov, line 45: How about:
```cpp
auto footprint = size + sizeof(Block));
auto l = (footprint & (footprint - 1)) == 0 : lowerLog2(footprint) : lowerLog2(footprint) + 1;
```

As I understand you add `+1` for cases when footprint is not power of 2 and required memory is `2^upperLog2(footprint)`, but in cases when footprint is pow of 2 this addition is `x2` overheap.

BalyshevArtem: Done

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
```diff
@@ -0,0 +1,146 @@
...
+      {
+        tmp->next_free = block->next_free;
+        return;
+      }
+
+      tmp = tmp->next_free;
+    }
+  }
+
+  Block *divideBlock(Block *block, int32_t l)
```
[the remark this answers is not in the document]
BalyshevArtem: Done

compiler/luci-interpreter/include/luci_interpreter/BuddyMemoryManager.h
```diff
@@ -0,0 +1,146 @@
...
+  void addToBlocks(Block *block, int32_t l)
+  {
+    if (!block)
+      return;
+
+    block->next_free = _free_blocks[l];
+    _free_blocks[l] = block;
+  }
+
+  void removeFromBlocks(Block *block, int32_t l)
```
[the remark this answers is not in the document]
BalyshevArtem: done

m-bronnikov (approved):
LGTM

BalyshevArtem:
@nnfw-bot test nncc-debug

binarman (approved):

compiler/luci-interpreter/src/BuddyMemoryManager.cpp
```diff
@@ -0,0 +1,93 @@
...
+  auto size = num_elements * element_size;
+  auto footprint = size + sizeof(Block);
+  auto l = (footprint & (footprint - 1)) == 0
+             ? lowerLog2(footprint)
+             : lowerLog2(footprint) + 1; // check footprint is pow_of_2
+
+  while (!_free_blocks[l] && l < 32)
+    l++;
+
+  assert(l < 32);
```
mhs4670go, line 53: @BalyshevArtem Hi, I don't know this part well but static analysis tool worried about the possibility `l` == 32 case. can `l` never be 32?

BalyshevArtem: @mhs4670go , Hi! When l is 32, it means that a piece of memory of 8 GB has been requested and we do not expect that someone will ever ask for so much, so we can assume that this is a limitation of our memory manager. It would be a good idea to add an assert to the constructor of this memory manager, thank you!

PR #7726, opened 2021-09-23
[one-cmds] Support force_quantparam in one-quantize and onecc

jinevening:
This supports force_quantparam in one-quantize and onecc.

---
Related to: #7682
Draft PR: #7683


compiler/one-cmds/one-quantize
```diff
@@ -88,6 +88,17 @@ def _get_parser():
...
+    # arguments for force_quantparam
+    parser.add_argument(
+        '--force_quantparam',
+        action='store_true',
+        help='write quantparam to the specified tensor')
+    parser.add_argument(
+        '--tensor_name', type=str, action='append', help='tensor name (string)')
+    parser.add_argument('--scale', type=float, action='append', help='scale (float)')
+    parser.add_argument(
+        '--zero_point', type=int, action='append', help='zero point (int)')
```
seanshpark, line 100: Q) if we only set one of three sub options (or two or all three), without `force_quantparam`, how does this work?
Is there any error or warning message?

From this question, I though of giving `--force_quantparam` with argument `name,scale,zero_point` in CSV format...

Which would be better?

jinevening: > if we only set one of three sub options (or two or all three), without force_quantparam, how does this work?
Is there any error or warning message?

The sub options (scale, tensor_name, zero_point) will be ignored if force_quantparam is not enabled. If the warning messages are needed, I'll add them in a separate PR (it would be better to implement in circle-quantizer).

> From this question, I though of giving --force_quantparam with argument name,scale,zero_point in CSV format...

That could be an option, but the current use case of force_quantparam is just updating the qparam of an input tensor. There is only a single (or maybe a few) target tensor at least for now, so I made the tool simply accept input tensors in command line, not a separate file.

I think we can extend force_quantparam to use a separate input file when the feature is needed (e.g., updating qparams of many tensors).

seanshpark: > updating qparams of many tensors

Ah, I thought just for one tensor, `name,scale,zero_point` :)
Anyway this will require revision of whole codes in this PR... -_-;
But if you go with separate options for `name`, `zero_point` and `zero_point` the changing this to one `name,scale,zero_point` argument won't be good idea...

jinevening: Ah, I misunderstood your intention.

> Anyway this will require revision of whole codes in this PR... -_-;

Indeed.. Plus, I'm not sure `tensor_name` does not have a comma.

seanshpark: > Indeed.. 

OK, let's go with current implementation..

> Plus, I'm not sure tensor_name does not have a comma.

Me too, I've seen `/`, `;` but not sure `,` -_-; ... Maybe surround whole string with `"`" :)

seanshpark (approved):
LGTM

compiler/one-cmds/one-quantize
```diff
@@ -247,8 +309,12 @@ def main():
...
     _verify_arg(parser, args)
 
-    # quantize
-    _quantize(args)
+    if _utils._is_valid_attr(args, 'force_quantparam'):
+        # write quantization parameters
+        _write_qparam(args)
+    else:
+        # quantize
+        _quantize(args)
```
mhs4670go, line 317: How about insert this change into `_quntize` method?
```python
def _quantize(args):
    # get file path to log
    dir_path = os.path.dirname(os.path.realpath(__file__))
    logfile_path = os.path.realpath(args.output_path) + '.log'

	if _utils._is_valid_attr(args, 'force_quantparam'):
        # write quantization parameters
        _write_qparam(args)
		return
```

jinevening: Updated.

compiler/one-cmds/one-quantize
```diff
@@ -233,6 +263,43 @@ def _quantize(args):
         _utils._run(circle_quantizer_cmd, err_prefix="circle_quantizer", logfile=f)
 
 
+def _write_qparam(args):
+    # get file path to log
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    logfile_path = os.path.realpath(args.output_path) + '.log'
```
mhs4670go, line 269: Maybe?
```suggestion
```

jinevening: `dir_path` and `logfile_path` are used below.

mhs4670go: Ah, sorry.
```python
def _quantize(args):
    if _utils._is_valid_attr(args, 'force_quantparam'):
        # write quantization parameters
        _write_qparam(args)
        return

    # get file path to log
    dir_path = os.path.dirname(os.path.realpath(__file__))
    logfile_path = os.path.realpath(args.output_path) + '.log'
```
If condition can be hoisted.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #7736, opened 2021-09-23
[luci-interpreter] Add assert in BuddyMemoryManager

BalyshevArtem:
This pr adds an assert to the buddy memory manager constructor.

### DIFF
--- a/compiler/luci-interpreter/src/BuddyMemoryManager.cpp
+++ b/compiler/luci-interpreter/src/BuddyMemoryManager.cpp
@@ -22,6 +22,9 @@ namespace luci_interpreter
 BuddyMemoryManager::BuddyMemoryManager(uint8_t *memory_start, int32_t memSize)
 {
   int32_t p = lowerLog2(memSize);
+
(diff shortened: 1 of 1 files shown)


compiler/luci-interpreter/src/BuddyMemoryManager.cpp
```diff
@@ -22,6 +22,9 @@ namespace luci_interpreter
 BuddyMemoryManager::BuddyMemoryManager(uint8_t *memory_start, int32_t memSize)
 {
   int32_t p = lowerLog2(memSize);
+
+  // We assume that the requested size of memory does not exceed 8 GB
```
mhs4670go, line 26: Isn't it 4GB? 

2^32 == 4GB

BalyshevArtem: Yes, you are right! Thank you!

mhs4670go (approved):
LGTM

binarman (approved):
LGTM

PR #7737, opened 2021-09-23
[luci-interpreter] Return vector by reference

m-bronnikov:
This commit reduce data copying with return vector by reference.

### DIFF
--- a/compiler/luci-interpreter/src/core/Kernel.h
+++ b/compiler/luci-interpreter/src/core/Kernel.h
@@ -36,8 +36,8 @@ protected:
 public:
   virtual ~Kernel() = default;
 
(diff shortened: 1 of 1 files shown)


m-bronnikov:
Return of vector by value may produce negative performance effect in places like this:

https://github.com/Samsung/ONE/blob/4fff17f198f1cd8dc240a19b3169012832330477/compiler/luci-interpreter/src/kernels/If.cpp#L64

binarman:
> Return of vector by value may produce negative performance effect in places like this

I think typical number of inputs/outputs are 1 =)

binarman (approved):
LGTM

PR #7740, opened 2021-09-23
[DRAFT][WIP][luci-interpreter] Alloc memory only for module inputs

m-bronnikov:
This commit allocs memory only for inputs of main graph of runtime module.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/If.cpp
+++ b/compiler/luci-interpreter/src/kernels/If.cpp
@@ -60,34 +60,35 @@ void If::execute() const
   const auto &graph_inputs = active_graph->getInputTensors();
   const auto &graph_outputs = active_graph->getOutputTensors();
 
-  // Copy kernel inputs to active graph inputs.
+  // share kernel inputs to active graph inputs.
   for (size_t i = 0; i < getInputTensors().size() - 1; ++i)
   {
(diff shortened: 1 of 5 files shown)


compiler/luci-interpreter/src/kernels/If.cpp
```diff
@@ -60,34 +60,35 @@ void If::execute() const
...
-    // TODO: Think about how allocate memory for output in main graph
-    active_graph->configureAllocations(output(i));
 
-    const int32_t num_elements = output(i)->shape().num_elements();
-    const std::size_t element_size = getDataTypeSize(output(i)->element_type());
-    std::memcpy(output(i)->data<void>(), graph_outputs[i]->data<void>(),
-                num_elements * element_size);
+    // Note: usage of subgraph output's buffer as operation output's memory buffer is safe since
+    // runtime graph not release input and output tensors itself
+    output(i)->set_data_buffer(graph_outputs[i]->data<uint8_t>());
```
binarman, line 84: Could you run interpreter with leak sanitizer?
I think memory can leak at this point.

binarman: I was thinking about managing graph outputs memory **outside** of it's graph.

If I understand right, you proposed opposite thing: inner graph manages it's output, and output graph reuse it

m-bronnikov:
Let's close as declined.

PR #7744, opened 2021-09-23
[res] Introduce tflite 2.6.0 schema

seanshpark:
This will introduce TFlite 2.6.0 schema and revise download list.


mhs4670go (approved):
LGTM

Issue #7724, opened 2021-09-23
[luci] Fuse Add with FullyConnected


jinevening:
## What

Let's fuse Add (from batchnorm) with a preceding FullyConnected.

jinevening:
![image](https://user-images.githubusercontent.com/5449554/134607059-08319968-724e-49dd-b2c0-c50429ec5633.png)

Add can be fused to the bias of FC.

jinevening:
All done.

PR #7746, opened 2021-09-24
DRAFT: Fuse Add with FullyConnected

jinevening:
On going draft to fuse Add with FullyConnected.

---
Related to: https://github.com/Samsung/ONE/issues/7724

### DIFF
--- a/compiler/luci/pass/src/FuseAddWithFullyConnectedPass.cpp
+++ b/compiler/luci/pass/src/FuseAddWithFullyConnectedPass.cpp
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


compiler/luci/pass/src/FuseAddWithFullyConnectedPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+ */
+bool fuse_add_with_fc(luci::CircleFullyConnected *fc)
+{
+  if (not fc)
+    return false;
+
+  if (fc->dtype() != loco::DataType::FLOAT32)
+    return false;
+
+  if (fc->fusedActivationFunction() != luci::FusedActFunc::NONE)
```
seanshpark, line 49: (Q) if `fc` and `add` have same activation, can we do the fusion?

jinevening: I don't think so. If fc has an activation function, the semantic will change.

PR #7750, opened 2021-09-24
[tfldump] Dump signature_defs

seanshpark:
This will enable dump of signature_defs records.


seanshpark:
This will dump something like this
```
SignatureDef
S(0) serving_default, key(serve), sub_graph(0)
    I T(0) input0: 0
    O T(0) out1: 705
    O T(1) out2: 710
    O T(2) out3: 714
```

mhs4670go (approved):
LGTM

PR #7755, opened 2021-09-24
[luci-interpreter] Fix int vs int32_t mismatch errors for arm build

binarman:
This PR fixes conversion erorrs caused by int != int32_t on arm.

Related issue: #7288


binarman:
I removed some files from PAL that are not that easy to enable for MCU build due to integer mismatches.

seanshpark:
@nnfw-bot test nncc-debug

binarman:
More info about what is going on:

`tensorflow/lite/kernels/internal/reference/reference_ops.h` and `tensorflow/lite/kernels/internal/optimized/optimized_ops.h` headers include other headers with kernel implementations (like `tensorflow/lite/kernels/internal/reference/conv.h`). But also they include additional code that is not cross-platform.

So I changed general headers with operation specific, where it was possible.

SlavikMIPT (approved):
LGTM

compiler/luci-interpreter/src/kernels/Equal.h
```diff
@@ -42,9 +42,9 @@ class Equal : public Kernel
 
 private:
   int32_t _x_multiplier = 0;
-  int32_t _x_shift = 0;
+  int _x_shift = 0;
   int32_t _y_multiplier = 0;
-  int32_t _y_shift = 0;
+  int _y_shift = 0;
```
m-bronnikov, line 47: Why we need this changes? Datatype should be less specific?

m-bronnikov (approved):

compiler/luci-interpreter/src/kernels/Equal.h
binarman: This is because of it's usage in quantizeMultiplierSmallerThanOneExp function:
it takes `int&` and it is incompatible with `int32_t &`

Another option is to rewrite this function, but I looked at tensorflow implementation and found that they use simple int, so I choose it =)

PR #7759, opened 2021-09-24
[luci] Configurable library type

binarman:
This PR introduces option to select
shared/static type of generated library.

This is a cosmetic change, to suppress cmake warning in case we build with compiler that can not produce shared libraries.
I ran into this when compiled luci-interpreter for MCU (and used `arm-none-eabi-gcc`) #7288 

CMake still produces all libraries, it just forces them to have STATIC type.
This change is not mandatory, I can close it if we do not want this complication.

### DIFF
--- a/compiler/luci/CMakeLists.txt
+++ b/compiler/luci/CMakeLists.txt
@@ -1,3 +1,8 @@
+# Some targets do not support dynamic linking: MCU, TrustZone applications, etc.
+# STATIC_LUCI option allows us to compile luci and luci related components safely
+# and suppress various cmake warnings.
+#
+# Currently this feature is used for luci-interpreter MCU builds.
 if (STATIC_LUCI)
   set(LIBRARY_TYPE "STATIC")
 else()
--- a/compiler/luci/CMakeLists.txt
+++ b/compiler/luci/CMakeLists.txt
@@ -1,3 +1,9 @@
+if (STATIC_LUCI)
+  set(LIBRARY_TYPE "STATIC")
+else()
+  set(LIBRARY_TYPE "SHARED")
+endif()
+
 add_subdirectory(env)
 add_subdirectory(log)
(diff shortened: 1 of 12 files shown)


compiler/luci/CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
+if (STATIC_LUCI)
```
seanshpark, line 1: I want to request this `STATIC_LUCI` be explictly defined somewhere like `infra/nncc/CMakeLists.txt`
and do not change characteristics before-after introducing this, and certainly not change characteristics of our backend.
And also a short comment here why we added this.

infra/nncc/CMakeLists.txt
```diff
@@ -130,6 +130,11 @@ option(ENABLE_STRICT_BUILD "Treat warning as error" OFF)
 # Check our ProtobufConfig.cmake for its usage.
 option(USE_PROTOBUF_LEGACY_IMPORT "Use legacy MODULE mode import rather than CONFIG mode" OFF)
 
+# This option might be turned ON for MCU builds of luci related components.
+# It specify which library type to use for build:
+# if set ON - luci libraries are static, otherwise - shared.
+option(STATIC_LUCI "Build luci as a static libraries" OFF)
+
```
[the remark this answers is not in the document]
binarman: moved it to #7767

compiler/luci/CMakeLists.txt
seanshpark: > I want to do same thing with several other libraries (for example loco)

_loco_ and other modules that are used in NPU backend needs review from @parjong

seanshpark (approved):
LGTM

binarman:
@SlavikMIPT FYI

SlavikMIPT (approved):
LGTM

PR #7757, opened 2021-09-24
[luci-interpreter] fix pal related cmakes

binarman:
This PR moves platform specific sources and headers to PAL
and adds warning message.

related issue: #7288

ONE-DCO-1.0-Signed-off-by: Alexander Efimov <a.efimov@samsung.com>

PR #7764, opened 2021-09-27
[luci-micro] Introduce basic infrastructure for arm build

binarman:
This PR introduces basic infrastructure to build interpreter for arm-based MCUs.

related issue: #7288 


binarman:
This includes changes from #7757, this PR would not build without it =(

seanshpark:
`luci-interpreter` and `luci-micro` files are mixed. please split them.

seanshpark:
> This includes changes from #7757, this PR would not build without it =(

This is critical dependency problem. Please solve.

seanshpark:
> This is critical dependency problem. Please solve.

@binarman , I think this is not a critical dependency problem but just luci-mico depends on luci-interpreter. (Am I right?)

If so, it would be better to not include luci-interpreter codes, mark as `PR/NO_MERGE` and wait  #7757 to land and leave build as break.  If you are uncomfortable with build fail, please wait it lands and then send PR.
I think you will rebase after #7757, wait for the build to finish.

Why I am asking is that we don't mix module codes but this PR does and there was an alert on my head.
(This really confused me at first)

What I see is that you can get the  review with build success (without #7757) 
but anyway you will have to rebase and get review again.

compiler/luci-micro/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+  DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/standalone/CMakeLists.txt"
+  VERBATIM
+)
+
+add_custom_target(luci_interpreter_micro_arm_cmake DEPENDS "${MICRO_ARM_BUILD_DEPENDENCY}")
+
+set(MICRO_ARM_BINARY "${MICRO_ARM_BUILD_DIR}/compiler/luci-interpreter/src/libluci_interpreter.a")
+
+add_custom_command(
+        OUTPUT "${MICRO_ARM_BINARY}"
```
seanshpark, line 50: two `add_custom_command` here and above uses different indentation. maybe use same.

compiler/luci-micro/README.md
```diff
@@ -0,0 +1,55 @@
...
+```
+$ sudo apt-get install gcc-arm-none-eabi
+```
+
+**cmake build**
+
+``` bash
+$ cd <path to ONE>
+$ mkdir build
+$ cmake ../infra/nncc
```
seanshpark, line 31: maybe `cd build` is missed?

seanshpark (approved):
LGTM

compiler/luci-micro/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+$ sudo apt-get install gcc-arm-none-eabi
+```
+
+**cmake build**
+
+``` bash
+$ cd <path to ONE>
+$ mkdir build
+# cd build
```
seanshpark, line 31: Above uses `$` but `#` prompt here. maybe use same ?

binarman: oof, I am extremely careless, sorry =_=

Let me fix it in following PR's, I am going to modify (add build options and maybe targets) this instruction anyway =_=

SlavikMIPT (approved):
LGTM

PR #7772, opened 2021-09-29
[DRAFT] Introduce Circle-Memory-Planner tool

BalyshevArtem:
This pr adds circle-memory-planner tool and all necessary code to CircleImportMetadata and CircleExportMetadata. This tool accepts a circle file as input. Then MemoryPlanner works and the found information is recorded in the circle file metadata.

issue #7522
first draft #7573

ONE-DCO-1.0-Signed-off-by: Artem Balyshev a.balyshev@partner.samsung.com

Issue #7522, opened 2021-08-18
[luci-interpreter] Design and implement static memory planner and manager


binarman:
This is continuation of #7355 that should use introduced memory manager interface.

## Goal

Introduce at least one static memory planner and manager suitable for MCU usage.

High level schema of interpreter setup:
![interpreter_initialization](https://user-images.githubusercontent.com/3631351/129979890-e6c6120c-cbaf-4d1c-9eb5-3e1c2671bcc0.png)

- Static memory manager contains memory pool that it can assign to tensors on request.
- Static memory manager do not owns memory buffer, so it is possible to use static array for this purpose.

High level interpreter algorithm:

- Interpreter seeks for "execution order" metadata in model
  - if it is found - execute kernels in given order
  - if it is **not** found - execute using standard post order
- According allocation plan interpreter asks memory manager to allocate memory for tensor.
  - static memory manager seeks into metadata stored in tensor and assigns actual memory address to a given tensor.

## Rationale

- Baremetal MCU has no OS, so we need to manage memory manually.
- Even if we have OS (for example, MBED OS), it's memory subsystem is not optimal for neural networks domain.
  - Only dynamic allocations available (no way to plan ahead of time)
  - Poor utilization of different memory types (for example TCM on ARM-m7)

binarman:
+cc @BalyshevArtem

BalyshevArtem:
**Brief explanation about circle-memory-plan tool**

This tool accepts a `circle file` as input and also returns a `circle file`. The output circle file contains metadata about the information that the memory planner found. This information is a vector of `uint32_t` values - `memory plan vector`. 
Each node in the graph has its own `memory plan vector`.
This memory plan vector for the current node stores:
1) execution position of current node - **required**
2) offsets from the beginning of the allocated memory buffer:
     a) right after execution position there is offset of current node in buffer - **required**
     b) then offsets for temporary tensors that can be produced by the current node -  **not required**

In order to record and read this metadata, we use CircleImportMetadata and CircleExportMetadata. And for this purpose we use `std::map<uint32_t, std::vector<uint32_t>> _memory_plan_table` which for each node with key `ID` contains `memory plan vector`

binarman:
Functionality is merged, now we need to cover following code with tests:
- [ ] luci-interpreter parts that uses static memory manager
- [ ] circle-execution-plan tool

seanshpark:
[...]

PR #7773, opened 2021-09-29
[luci] Add Memory Plan metadata

BalyshevArtem:
This pr adds memory_plan metadata to luci importer and exporter.
It helps write memory plan information found by circle-memory-plan tool to circle file and import this information from circle file. And then use it in static memory manager.
This is first step of merging from #7772.

issue #7522


compiler/luci/profile/include/luci/Profile/CircleNodeMemoryPlan.h
```diff
@@ -0,0 +1,57 @@
+/*
```
binarman, line 1: not sure this file is profile related.
Need to find more appropriate place

compiler/luci/export/src/SerializedData.h
```diff
@@ -63,13 +63,23 @@ class CircleExportMetadata
...
   const std::vector<uint8_t> encoded_op_table(void);
+  const std::vector<uint8_t> encoded_memory_plan_table(void);
 
 private:
   std::map<uint32_t, std::string> _source_table;
   std::map<uint32_t, std::set<uint32_t>> _op_table;
+  // _memory_plan_table stores for node with node_id order of execution, and offsets:
+  // first go execution order, then offset of current node
+  // and then offsets of temporary nodes if needed
+  std::map<uint32_t, std::vector<uint32_t>> _memory_plan_table;
```
seanshpark, line 82: As wrote in the draft, it would be bette to use one term `plan` or `table`

seanshpark: As this memory_plan_table is widely used in several modules it would be using an alias as
```cxx
using MemoryPlanTable = std::map<uint32_t, std::vector<uint32_t>>;
```
or 
```cxx
using MemoryPlanTable_t = std::map<uint32_t, std::vector<uint32_t>>;
```

seanshpark:
It would be better to split this change in each purpose to separate PRs.

compiler/luci/profile/include/luci/Profile/CircleNodeMemoryPlan.h
seanshpark: > Not sure this file is profile related.

Agree :)

seanshpark:
Where can I get some brief explanation about memory plan (table) ?
There are files in `profile` folder that would be better to relocate it 
and I want suggest a better place, but need some understanding.

BalyshevArtem:
> Where can I get some brief explanation about memory plan (table) ? There are files in `profile` folder that would be better to relocate it and I want suggest a better place, but need some understanding.

@seanshpark
I tried to give a brief explanation in this issue comment: https://github.com/Samsung/ONE/issues/7522#issuecomment-931128956

compiler/luci/env/include/luci/UserSettings.h
```diff
@@ -33,6 +33,7 @@ struct UserSettings
     MuteWarnings,
     DisableValidation,
     ProfilingDataGen,
+    MemoryPlanGen,
```
binarman, line 36: I have some concerns about "Memory" in this name:
We plan memory AND execution order at the same time...

So we can name it "ExecutionPlanGen", but this sounds too generic for me...

compiler/luci/profile/include/luci/Profile/CircleNodeMemoryPlan.h
binarman: @seanshpark 
What do you think about separate directory, something like "ExecutionPlan"?

Different approach I see is:
- introduce "plugin" interfaces in luci importer and exporter to injecting external code
- place plan related code in luci-interpreter/somewhere else

compiler/luci/lang/include/luci/IR/Module.h
```diff
@@ -65,6 +65,13 @@ class Module final
 
   const std::map<uint32_t, std::string> &source_table(void) const { return _source_table; }
 
+  void memory_plan(const std::map<uint32_t, std::vector<uint32_t>> &memory_plan)
```
binarman, line 68: I think code in this file is redundant, could you remove it?

compiler/luci/import/src/Importer.cpp
```diff
@@ -315,8 +315,14 @@ std::unique_ptr<Module> Importer::importModule(const circle::Model *model) const
...
   post_import_graph(module.get(), reader);
 
-  // Initialize 'source_table'
+  // Initialize 'memory_plan_table'
   auto circle_metadata = std::make_unique<luci::CircleImportMetadata>(reader);
+  if (circle_metadata->memory_plan_table().size() > 0)
+  {
+    module->memory_plan(circle_metadata->memory_plan_table());
+  }
+
```
BalyshevArtem, line 324: @binarman,  As I understand code from compiler/luci/lang/include/luci/IR/Module.h should be removed 
> I think code in this file is redundant, could you remove it?

that is why code from this file also should be removed, right? 


But how will we pass the information to the module?

compiler/luci/env/include/luci/UserSettings.h
BalyshevArtem: Maybe smth like:
`InterpreterPlanGen` ?

compiler/luci/profile/include/luci/Profile/CircleNodeMemoryPlan.h
seanshpark: I think first one looks better for now. Second one may take some time but maybe better for future extensions...
Does 'MemoryPlan' need profile stuff? I didn't look in to the code so don't know this question is valid...
If `profile` and memory planning are not related then can we think of adding something like `luci/plan` ?
This `plan` or `planning` can provide memory and execution planning. 
Or any better word is welcome for purpose of execution optimization. (as I understand this to provide information for better execution performance)

BalyshevArtem: As I understand `profile` and memory plan are not related. And that is why, yes, i think adding `luci/plan` can be good idea. @binarman what do you think about it?

compiler/luci/import/src/Importer.cpp
seanshpark: > But how will we pass the information to the module?

Just by reading this, (not know other things discuessed), we can introduce new header for this where both modules can access. maybe in new `luci/plan` folder for planning stuff.

compiler/luci/profile/include/luci/Profile/CircleNodeMemoryPlan.h
binarman: > Does 'MemoryPlan' need profile stuff?

> As I understand profile and memory plan are not related

For now I don't see how these two could be related =)

I think "plan" is good =)

@seanshpark FYI
> as I understand this to provide information for better execution performance

This should reduce **initialization** time and **memory consumption** of luci-interpreter (maybe other runtimes, if they will to use this annotations)

seanshpark: FYI when `luci_plan` gets to exist, we need to register to `infra/debian/compiler/one-compiler.links`

@mhs4670go , is there any other place to register new file?

compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -22,6 +22,8 @@
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/CircleNodeVisitor.h>
 #include <luci/Profile/CircleNodeOrigin.h>
+#include <luci/Plan/CircleNodeExecutionPlan.h>
+
```
seanshpark, line 26: maybe redundant emtpy space?

seanshpark:
I think it would be better to split to small PRs, as usual; there are multiple context of changes and that I don't know still need discussions and some are simple that can be landed without any difficulties.

compiler/luci/profile/include/luci/Profile/CircleNodeMemoryPlan.h
mhs4670go: @seanshpark Just registering it to `infra/debian/compiler/one-compiler.links` is enough.

PR #7775, opened 2021-09-30
[luci] Propagate qparam of TConv op

jinevening:
This propagates qparam of TConv op.


seanshpark (approved):
LGTM

Issue #7596, opened 2021-09-01
[Circle-OpSelector] Let's design Circle-OpSeletor Project


Paran-Yu:
Hi! We are team COSMOS from SSAFY(Samsung SW Academy For Youth).
We will make `Circle-OpSelector` referring to issue #6648.
This issue regards our plan on developing the `Circle-OpSelector`. We are open to all kinds of feedbacks, so please feel free to offer your opinion.

### 1. Select from location numbers
(EDIT "1-3 5" -> "1-3,5", --op_ids -> --by_id)
```bash
./circle-opselector --by_id "1-3,5" input.circle output.circle
```

Then, circle-opselctor creates a model consisting of subgraphs that only contain nodes in the given op id list(`1-3` means *from 1 to 3*). In the case above, nodes whose indices are { 1, 2, 3, 5 } will be selected.

### 2. Select from node names

(EDIT "Add_1-Sub_1 Concat_2" -> "Add_1,Sub_1,Concat_2", --op_names -> --by_name)
```bash
./circle-opselector --by_name "Add_1,Sub_1,Concat_2" input.circle output.circle
```

~~Then, circle-opselctor creates a model consisting of subgraphs that only contain nodes in the given op name list. In the case above, both nodes whose names are {  Add_1, Sub_1, Concat_2 } and nodes between Add_1 and Sub_1 will be selected.~~
(EDIT)
Then, circle-opselctor creates a model consisting of subgraphs that only contain nodes in the given op name list. In the case above, both nodes whose names are {  Add_1, Sub_1, Concat_2 } will be selected.

### 1. CLI
- [x]  Plan Options (#7596)
- [x]  Create Draft (#7600)
- [x]  Option Parsing
    - [x] --by_id, --by_name(#7777 )
    - [x]  --select, --deselect

### 2. Test
- [ ]  Prepare Test Set(`tflite` file) for testing
- [ ]  Define selecting circle model process by shape of circle mode and command
- [ ]  Compare result of `select_operator.py(select tflite)` and `opselector(select circle)`
    - [ ]  Select tflite and circle model(Converting .tflite to .circle)
    - [ ]  Create golden value by selected **tflite** model
    - [ ]  Create nnpackage by selected **circle** model
    - [ ]  Compare two models using `nnpkg-test`
    - [ ]  Create a shell script file to automate the above process
    - [ ]  Manage cases which can't make golden value (#7630)
- [ ]  Test automation from random cases

### 3. Core
[...]

PR #7789, opened 2021-10-05
[DRAFT][circle-opselector] Create Core

dongyooncho:
> Related #7596, #7600 and #7777

~This PR is the Core code of circle-opselector.~
This PR introduce entire codes of selecting nodes methods.
It can build new graphs from user selection.

`select_nodes` proceeds in two steps.
Frist, it check if the selected nodes are all connected and decide which node to use for input and output.
And this step checks if there is a `while` or `if` node.

Next, it build a new graph with this information.
When the user selects the `while` or `if` node, subgraph is copied along with it.

I am sorry for creating huge PR.

ONE-DCO-1.0-Signed-off-by: dongyooncho <dycho96@gmail.com>


compiler/circle-opselector/driver/Driver.cpp
```diff
@@ -265,7 +266,13 @@ int entry(int argc, char **argv)
     std::cerr << "ERROR: No operator selected" << std::endl;
     exit(EXIT_FAILURE);
   }
-  // TODO implement node selections
+  // Select nodes.
+  // To avoid error, I declare model_data in Driver.cpp
+  // Otherwise 'CircleReader' throw an error when use select_subgraph.
```
seanshpark, line 271:
```suggestion
  // model_data is used to prevent error when calling select_subgraph()
```

seanshpark: This is strange...

compiler/circle-opselector/src/ModuleIO.cpp
```diff
@@ -27,7 +27,7 @@
 namespace opselector
 {
 
-std::unique_ptr<luci::Module> getModule(std::string &input_path)
+std::vector<char> getModelData(std::string &input_path)
```
seanshpark, line 30: This PR seems to have multiple purpose of changes.
It would be better to split them. This one changed function prototype.

seanshpark:
> Create Core

Please be more specific. This is a bit broad.

compiler/circle-opselector/src/ModuleIO.cpp
```diff
@@ -41,6 +41,14 @@ std::unique_ptr<luci::Module> getModule(std::string &input_path)
     exit(EXIT_FAILURE);
   }
 
+  return model_data;
+}
+
+std::unique_ptr<luci::Module> getModule(std::string &input_path)
```
seanshpark, line 47: This introduce new method

compiler/circle-opselector/src/OpSelector.cpp
```diff
@@ -0,0 +1,396 @@
...
+{
+  const auto &operators = _reader.operators();
+
+  std::set<uint32_t> selected_input_tensors;
+  std::set<uint32_t> selected_output_tensors;
+
+  std::vector<const circle::OperatorT *> selected_operators;
+
+  // enroll all output nodes.
+  for (auto &op : operators)
```
seanshpark, line 55: please use `{`, `}` even for just one line

compiler/circle-opselector/src/OpSelector.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace luci
+{
+void post_import_graph(luci::Module *module, const luci::CircleReader &reader);
+}
+
+namespace opselector
+{
+
+class OpSelector
```
seanshpark, line 34: I think it would be better to revise this PR to introduce this `class OpSelector` and change th comment explaining what it doesn.

compiler/circle-opselector/src/OpSelector.h
```diff
@@ -0,0 +1,63 @@
...
+private:
+  void build_cache_outputs(luci::GraphBuilderContext &gb_context);
+  void create_graph_inputs(luci::GraphBuilderContext &gb_context, uint32_t input);
+  void create_circle_const(luci::GraphBuilderContext &gb_context);
+  void import_operators(luci::GraphBuilderContext &gb_context);
+  void create_graph_outputs(luci::GraphBuilderContext &gb_context, uint32_t output);
+
+private:
+  luci::CircleReader _reader;
+  const circle::Model *_src_model;
```
seanshpark, line 57:
```suggestion
  const circle::Model *_src_model = nullptr;
```

compiler/circle-opselector/src/OpSelector.cpp
```diff
@@ -0,0 +1,396 @@
...
+  {
+    uint32_t node_id = luci::get_node_id(cnode);
+    selected_operators.push_back(operators[node_id].get()); // put selected nodes in vector.
+
+    if (cnode->name().find("while") != std::string::npos ||
+        cnode->name().find("if") != std::string::npos) // if has while of if node,
+    {
+      _has_subgraph = true; // A flag indicating whether to copy the subgraph or not,
+    }
+  }
```
seanshpark, line 73: I'm not sure of this. can there be multiple subgraphs without `IF` or `WHILE` node?

compiler/circle-opselector/src/OpSelector.cpp
```diff
@@ -0,0 +1,396 @@
...
+#include <iostream>
+#include <map>
+
+#define MAIN_SUBGRAPH 0
+
+// TODO: Re-implement to use just module.
+namespace opselector
+{
+
+void OpSelector::check_connected(std::vector<const luci::CircleNode *> &selected_nodes,
```
seanshpark, line 42: Please add some doxygen format comments about this method for better understanding

compiler/circle-opselector/src/OpSelector.cpp
```diff
@@ -0,0 +1,396 @@
...
+  {
+    std::unique_ptr<loco::Graph> graph = loco::make_graph(); // create new empty graph
+
+    assert(_reader.select_subgraph(g)); // select subgraph. usually, 0 is main
+    graph->name(_reader.name());        // set name of graph (if it has one graph, name is empty.)
+
+    auto nodefinder = std::make_unique<luci::IndexNodeFinder>();
+    auto tensoroutputs = std::make_unique<luci::IndexTensorOutputs>();
+
+    luci::GraphBuilderContext gb_context(graph.get(), &_reader, nodefinder.get(),
```
seanshpark, line 346: Can you please explain why you need to use `GraphBuilderContext` ?

There's `Importer::importModule` what will read the circle file and return `Module`

glistening:
@dongyooncho I totally agree with @seanshpark. Please split this PR into several ones as @seanshpark suggested. It is okay and preferred for each PR to have a few lines of changes.

compiler/circle-opselector/src/OpSelector.cpp
dongyooncho: No. Without that, no subgraph.
I used it to check if it has `WHILE` or `IF` node.

I didn't know another way to check if a node is `WHILE` or `IF`.
So I checked if the string contains `WHILE` or `IF`.
And if so, I judged that there is a sub-item.

compiler/circle-opselector/src/OpSelector.cpp
dongyooncho: > Can you please explain why you need to use GraphBuilderContext ?

I use this because I create a new graph with only selected nodes to create a 'circle' file with selected nodes.

At first, I thought of two ways. The first was to use only module to perform Propagation of profiling data from the original module to the new module, and the second was to create a new graph by selecting only the desired data directly from the model data.
I tried it for quite a long time as the first method, but as you said, it was so difficult to understand the whole logic. So I chose the second way that was a little easier.

After the development was over, @mhs4670go informed me of the related module, so I was changing the structure, but the deadline was just around the corner, so I uploaded the existing code.

dongyooncho:
I close this PR and create new smaller PR.

compiler/circle-opselector/src/OpSelector.cpp
glistening: @dongyooncho I don't think adding doxygen comments will do. Better way is to rename the method name. I also expect `check_...` will return `true`/`false` or some enumeration and so on. But it returns nothing.

seanshpark:

PR #7792, opened 2021-10-06
[circle-opselector] Introduce getModelData

dongyooncho:
This PR introduce new method of ModuleIO.
It can get model data to vector from circle file.

It return model data to vector from circle file, and this data is used to prevent error when calling `CircleReader.selecting_subgraph()`

ONE-DCO-1.0-Signed-off-by: dongyooncho <dycho96@gmail.com>

### DIFF
--- a/compiler/circle-opselector/src/ModuleIO.cpp
+++ b/compiler/circle-opselector/src/ModuleIO.cpp
@@ -27,7 +27,8 @@
 namespace opselector
 {
 
-std::unique_ptr<luci::Module> getModule(std::string &input_path)
+// return circle model data.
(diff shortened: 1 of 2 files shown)


seanshpark:
@dongyooncho with comments in https://github.com/Samsung/ONE/pull/7789#discussion_r722743696 , are you going to continue work with your implementation of using parts of luci/import and your implementation?
I can't accept casuse of ` which will be a burden for maintenance.`

dongyooncho:
@seanshpark 
I've checked!
I'll close this PR and reopen #7789 to expose the entire code.

PR #7793, opened 2021-10-06
[tfl-inspect] Use mio-tflite260 and support v3a format

seanshpark:
This will revise to use mio-tflite26- and support v3a for format for
extended builtin_code.


mhs4670go (approved):
LGTM

seanshpark:
For #7616, from draft #7593

PR #7794, opened 2021-10-06
[tflite2circle] Use mio-tflite260 for v3a format

seanshpark:
This will revise to use mio-tflite260 to support v3a format.


mhs4670go (approved):
LGTM

PR #7765, opened 2021-09-28
 [Draft] Circle-OpSelector test code(new)

parkeunsang:
In prior draft code(#7728),  I think the overall structure of test should be changed. So I created new draft PR.
1. We will use to check the circle-opselector using `luci-interpreter`(or better methods) to avoid using **runtime environment**
2. To do this, We decided to use ctest by `CMakeLists.txt`. 
3. And to test the validity of selected circle files, various test set should be needed. So we will make these using `recipe`.

PR #7797, opened 2021-10-06
[res] add tflite recipes for circle-opselector-test

parkeunsang:
Related PR about `circle-opselector-test` is #7765
The file name's format is {origin_recipe_name}_Selected_00
ex) Part_Sqrt_Rsqrt_002_Selected_01 consists of partial nodes from Part_Sqrt_Rsqrt_002.


res/TensorFlowLiteRecipes/Part_Sqrt_Rsqrt_002_Selected_01/test.recipe
```diff
@@ -0,0 +1,37 @@
+operand {
```
seanshpark, line 1: `Part_xxx` recipes are for circle-partitioner so please use other prefix for opselector like `OpSel_xxx`.

seanshpark:
There is no strict limit of lines of recipe files. But each recipe has some reason of testing.
If you have added some with no particular reason of testing, it would be better to shrink to test that feature or condition of opselecor.

To be more specific, please write down the reason of needs, that is the condition or model's network configuration you want to test. If the the recipe has redundant nodes for that particular test, then you can remove them.
If you share what you wrote, that will help me review this PR.

seanshpark:
@parkeunsang , please make a mew branch from current master and add changes and post a PR.
I don't think current commit message will be clean cause of lots of commits and commit messages.

parkeunsang:
@seanshpark 
I understood. I will send a PR again after excluding unnecessary parts.

PR #7800, opened 2021-10-06
[luci/pass] Use mio_tflite260 and flatbuffers-1.12

seanshpark:
This will revise to use mio_tflite260 and flatbuffers-1.12 as this
version is used in TFlite 2.6.0.


mhs4670go (approved):
LGTM

PR #7802, opened 2021-10-07
[res] add tflite recipes for circle-opselector-test

parkeunsang:
Related PR about circle-opselector-test is #7765 
The file name's format is OpSel_{origin_recipe_name}_00x
ex) OpSel_Part_Sqrt_Rsqrt_002_001 consists of partial nodes from Part_Sqrt_Rsqrt_002.
+Other test cases will be submitted(creating new recipes take time so much...)
ONE-DCO-1.0-Signed-off-by: parkeunsang <dislive@naver.com>

### DIFF
--- a/res/TensorFlowLiteRecipes/OpSel_Part_Sqrt_Rsqrt_002_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/OpSel_Part_Sqrt_Rsqrt_002_000/test.recipe
@@ -1,0 +1,37 @@
+operand {
+  name: "sqrt"
+  type: FLOAT32
+  shape { dim: 1 dim: 3 dim: 3 dim: 2 }
+}
+operand {
+  name: "rsqrt"
+  type: FLOAT32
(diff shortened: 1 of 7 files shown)


glistening:
@parkeunsang Please add more description in your commit. It has no words.

```
[res] add tflite recipes for circle-opselector-test

ONE-DCO-1.0-Signed-off-by: parkeunsang <dislive@naver.com>
```

It will fix the error of `nobody-checker`.

glistening:
> > I think the origin recipe name should be included in selected recipe name
> 
> You don't have to. Thiis only makes names longer and someone can asume there can be some relation with `OpSel` and `Part` and as this is not true, may give confusion. You can start serial number from 000 or 001 as you like.

@parkeunsang I also +1 for not including origin in test file. What about leave comment in `.recipe` for specifying origin.

You can write comment in `.recipe` like the following:

```
# Test for ...
# Copied from ...

```

seanshpark:

PR #7814, opened 2021-10-13
[tflite2circle] Relocate verify tflite model

seanshpark:
This will relocate verify tflite model into TFLModel class.


mhs4670go:
@seanshpark I don't know the reason this PR is needed but I guess loaded data needs to be verified before `flatbuffers` read it. IIUC, could I post another PR that covers this PR with `foder` library?

seanshpark:
>  I don't know the reason this PR is needed

Just to make resposibility of tflite verify to TFLModel class as it looks like it.

> I guess loaded data needs to be verified before flatbuffers read it.

Not sure :)

> could I post another PR that covers this PR with foder library?

Um... I think foder  don't have to deal with flatbuffer files.

PR #7823, opened 2021-10-14
[luci-interpreter] Fix wrong short circuit

llFreetimell:
Array should be accessed after index is checked first.
This commit will fix it.

---

### 1 approval is enough


llFreetimell:
@BalyshevArtem FYI

PR #7838, opened 2021-10-15
[vconone] Fix compiler version

mhs4670go:
This commit fixes compiler version to be printed.


mhs4670go:
@seanshpark PTAL:)

PR #7846, opened 2021-10-18
[arser] Add NEG suffix

mhs4670go:
This commit adds NEG suffix to the tests that was supposed to be
negative.


mhs4670go:
@seanshpark PTAL:)

PR #7849, opened 2021-10-18
[tflite2circle] Revise to use member _fb

seanshpark:
This will revise to use member _fb for FlatBufferBuilder parameter.


seanshpark:
@nnfw-bot test nncc-release

PR #7857, opened 2021-10-19
[luci-micro-importer] Initial commit

m-bronnikov:
This commit initializes luci-micro-importer by copying existance *luci/importer*.

----------------------------

For: #7604 (phase 3, idea 2)
Draft: #7821


m-bronnikov:
## For reviewer

### Why new folder?

This importer should be a part of *luci-micro* project, but I not sure it's possible for now add this importer to build system of *luci-micro*. Please, correct me if I'm wrong. 

### What changed?

This commit copies *compiler/luci/importer* to *compiler/luci-micro-importer*. All files remained unchanged except:

- *CMakeLists.txt* is chaged. `luci_importer` target changed to `luci_micro_importer`.
- added *requires.cmake* file with dependency to *luci*.
- Changed *README.md*

seanshpark:
> This importer should be a part of luci-micro project, but I not sure it's possible for now add this importer to build system of luci-micro. Please, correct me if I'm wrong.

There is no explanation what's the difference of `luci_importer` and `luci_micro_importer`

seanshpark:
I disagree on landing this
- this change is too huge
- I don't know the reason of making a clone
- who is going to follow luci/importer changes to luci-micro-importer ?

binarman:
@seanshpark 
> There is no explanation what's the difference of luci_importer and luci_micro_importer

luci_micro_importer main goal is to minimize number of constant data copies:
- current luci_importer unpacks flatbuffers structures (one copy) and creates luci::CircleConstant node (second copy)
- luci_importer_micro do not unpack flatbuffers and creates custom node which reference original buffer instead of copying it:

See draft #7821 Importer.cpp:122
```
// Create CircleConst nodes for constant tensors.
  {
    luci::CircleNode *const_node = nullptr;
    if (copy_consts)
      const_node = luci::create_circleconst(&gb_context, i);
    else
      const_node = luci::create_circle_remote_const(&gb_context, i);
```
and https://github.com/Samsung/ONE/pull/7821/files#diff-1a70356211a01d6a2ebb3c8846622fd070fafaf2e3c1de66d7ccd2b77a5ae58a

seanshpark:
@binarman , thanks for the long comments !

> Constant copies are undesirable in environments, like MCU, because of small RAM size.

Yes, this is critical

> But since this importer is used only on MCUs it will not affect existing tools.

But still this code will exist in this repo and may need maintenance.
luci-interpreter depends on this code and our system depends on luci-interpreter.
And someday this module will be very important.

> If you are ok with new node type (CircleRemoteConst) in luci IR, these changes could be applied to original importer. What do you think?

I didn't look into the codes or thought of this, but seems nice to have this feature, especially when we think of embeeded systems.

> Another way that came to my mind is to extend GraphBuilderSource interface with method creating constant from flatbuffers.
In that case nothing new is needed in luci IR (But we still need to refactor luci Import to eliminate unpack overhead)

This too seems nice. I'm not sure "how to implement" this right now but, I think we can get a nice approach.

m-bronnikov:
Close because refactoring of exist luci/importer is started!

PR #7895, opened 2021-10-28
[luci-interpreter] Support DepthwiseConv2dS8 CMSIS-NN optimized kernel

SlavikMIPT:
This commit introduces DepthwiseConv2dS8 CMSIS-NN optimized kernel support.

### DIFF
--- a/compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
+++ b/compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef LUCI_INTERPRETER_PAL_DEPTHWISECONV2D_H
+#define LUCI_INTERPRETER_PAL_DEPTHWISECONV2D_H
+
+#include <tensorflow/lite/kernels/internal/reference/depthwiseconv_float.h>
+#include <tensorflow/lite/kernels/internal/reference/depthwiseconv_uint8.h>
+#include <tensorflow/lite/kernels/internal/reference/integer_ops/depthwise_conv.h>
+#include <arm_nnfunctions.h>
+
+namespace luci_interpreter_pal
+{
+template <typename T>
+static inline void
+DepthwiseConvPerChannel(const tflite::DepthwiseParams &params, const int32_t *output_multiplier,
+                        const int32_t *output_shift, const tflite::RuntimeShape &input_shape,
+                        const T *input_data, const tflite::RuntimeShape &filter_shape,
+                        const T *filter_data, const tflite::RuntimeShape &bias_shape,
+                        const int32_t *bias_data, const tflite::RuntimeShape &output_shape,
+                        T *output_data)
+{
+  {
+    // MARK: At this moment this operation is not supported
+    assert(false && "DepthwiseConvPerChannel NYI");
+    (void)params;
+    (void)output_multiplier;
+    (void)output_shift;
+    (void)input_shape;
+    (void)output_data;
+    (void)input_data;
+    (void)filter_shape;
+    (void)filter_data;
+    (void)bias_shape;
+    (void)bias_data;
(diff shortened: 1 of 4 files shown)


binarman:
@nnfw-bot test nncc-release

compiler/luci-interpreter/pal/linux/PALDepthwiseConv2d.h
```diff
@@ -0,0 +1,68 @@
...
+static inline void
+DepthwiseConvPerChannel(const tflite::DepthwiseParams &params, const int32_t *output_multiplier,
+                        const int32_t *output_shift, const tflite::RuntimeShape &input_shape,
+                        const T *input_data, const tflite::RuntimeShape &filter_shape,
+                        const T *filter_data, const tflite::RuntimeShape &bias_shape,
+                        const int32_t *bias_data, const tflite::RuntimeShape &output_shape,
+                        T *output_data)
+{
+  {
+    // MARK: At this moment this operation doesn't support
```
binarman, line 36: Maybe
```suggestion
    // MARK: At this moment this operation is not supported
```

compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
```diff
@@ -0,0 +1,137 @@
...
+    bias_dims.c = output_depth;
+
+    cmsis_nn_dims output_dims;
+    output_dims.n = batch_size;
+    output_dims.h = output_shape.Dims(1);
+    output_dims.w = output_shape.Dims(2);
+    output_dims.c = output_depth;
+
+    cmsis_nn_context ctx;
+    ctx.buf = nullptr;
```
binarman, line 115: I assume this is scratchpad, like im2col in normal convolution?
@BalyshevArtem FYI

binarman (dismissed):
LGTM, but this PR introduces interesting detail: some versions of depthwise requires memory for scratchpad.

compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
```diff
@@ -0,0 +1,137 @@
...
+    ctx.size = 0;
+
+    const int32_t buf_size = arm_depthwise_conv_wrapper_s8_get_buffer_size(
+      &dw_conv_params, &input_dims, &filter_dims, &output_dims);
+
+    ctx.buf = malloc(buf_size);
+    ctx.size = buf_size;
```
m-bronnikov, line 122: 1. Why we can't to define `buf_size` first and then create cmsis_nn_context?
2. Could we avoid usage of malloc and use `std::make_unique<int8_t[]>(buf_size);`?
3. If `malloc` is neccessary, please, add some assert.
4. It's Okay to allocate memory without introduced memory allocator?

```suggestion
    const int32_t buf_size = arm_depthwise_conv_wrapper_s8_get_buffer_size(
      &dw_conv_params, &input_dims, &filter_dims, &output_dims);
    auto buffer = std::make_unique<int8_t[]>(buf_size);
    
    cmsis_nn_context ctx;
    {
      ctx.buf = buffer.get();
      assert(ctx.buf != nullptr);
      ctx.size = buf_size;
    }
```

m-bronnikov: Comment above related to other PR's with kernels.

binarman: @SlavikMIPT ping?
This blocks another PR: #7987

binarman:
@SlavikMIPT  FYI
It better to add commits on top of existing ones if you do not need to rebase or do something like this.
It is easier to see what changes, github even provides interface to see diff after your lest review.

binarman (dismissed):

compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
binarman: @m-bronnikov 
> It's Okay to allocate memory without introduced memory allocator?

It is not the best way, but I think it is separate problem, because it requires refactoring of existing scratchpad related logic, BalyshevArtem is working on this problem.

compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
```diff
@@ -0,0 +1,143 @@
...
+
+    quant_params.multiplier = &output_multiplier;
+    quant_params.shift = &output_shift;
+
+    TFLITE_DCHECK_LE(dw_conv_params.activation.min, dw_conv_params.activation.max);
+    const int batch_size = tflite::MatchingDim(input_shape, 0, output_shape, 0);
+    const int output_depth = tflite::MatchingDim(filter_shape, 3, output_shape, 3);
+    if (bias_data)
+    {
+      TFLITE_DCHECK_EQ(bias_shape.FlatSize(), output_depth);
```
m-bronnikov, line 92: (optional) Personally, I think better to use `assert`'s instead tensorflow specific check's.

m-bronnikov (approved):
LGTM

binarman (approved):
LGTM

m-bronnikov (approved):

Issue #7886, opened 2021-10-26
[luci/importer] Support importing without models UnPack


m-bronnikov:
### What?

Current importer's implementation uses flatbuffer's `ModelT` obtained with UnPack of the `Model`:

https://github.com/Samsung/ONE/blob/2c0ab2d0ec92487138a281604b08c7c4f92a514c/compiler/luci/import/src/CircleReader.cpp#L322-L332

Let's import `luci::Module` from `circle::Model`directly!

### Why?

Internally, `UnPack()` provides `circle::ModelT` via full copying of the model:

```cpp
inline ModelT *Model::UnPack(const flatbuffers::resolver_function_t *_resolver) const {
  auto _o = new ModelT();
  UnPackTo(_o, _resolver);
  return _o;
}

inline void Model::UnPackTo(ModelT *_o, const flatbuffers::resolver_function_t *_resolver) const {
  (void)_o;
  (void)_resolver;
  { auto _e = version(); _o->version = _e; };
  { auto _e = operator_codes(); if (_e) { _o->operator_codes.resize(_e->size()); for (flatbuffers::uoffset_t _i = 0; _i < _e->size(); _i++) { _o->operator_codes[_i] = std::unique_ptr<OperatorCodeT>(_e->Get(_i)->UnPack(_resolver)); } } };
  { auto _e = subgraphs(); if (_e) { _o->subgraphs.resize(_e->size()); for (flatbuffers::uoffset_t _i = 0; _i < _e->size(); _i++) { _o->subgraphs[_i] = std::unique_ptr<SubGraphT>(_e->Get(_i)->UnPack(_resolver)); } } };
  { auto _e = description(); if (_e) _o->description = _e->str(); };
  { auto _e = buffers(); if (_e) { _o->buffers.resize(_e->size()); for (flatbuffers::uoffset_t _i = 0; _i < _e->size(); _i++) { _o->buffers[_i] = std::unique_ptr<BufferT>(_e->Get(_i)->UnPack(_resolver)); } } };
  { auto _e = metadata_buffer(); if (_e) { _o->metadata_buffer.resize(_e->size()); for (flatbuffers::uoffset_t _i = 0; _i < _e->size(); _i++) { _o->metadata_buffer[_i] = _e->Get(_i); } } };
[...]

PR #7901, opened 2021-10-29
[draft][luci/import] Use direct access API

m-bronnikov:
This draft changes UnPack API usage to direct access from model buffer.

-----------

Draft for: #7886

### DIFF
--- a/compiler/luci/import/include/luci/Import/CircleReader.h
+++ b/compiler/luci/import/include/luci/Import/CircleReader.h
@@ -35,19 +35,10 @@
 namespace luci
 {
 
-bool is_valid(const circle::OperatorCodeT &opcode);
 bool is_valid(const circle::OperatorCode *opcode);
-
-bool is_custom(const circle::OperatorCodeT &opcode);
 bool is_custom(const circle::OperatorCode *opcode);
-
-std::string opcode_name(const circle::OperatorCodeT &opcode);
 std::string opcode_name(const circle::OperatorCode *opcode);
-
-const char *tensor_name(const circle::TensorT &tensor);
 const char *tensor_name(const circle::Tensor *tensor);
-
-const circle::QuantizationParametersT *tensor_quantization(const circle::TensorT &tensor);
 const circle::QuantizationParameters *tensor_quantization(const circle::Tensor *tensor);
 
 loco::DataType luci_datatype(circle::TensorType type);
@@ -57,12 +48,9 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode);
 luci::CircleFullyConnected::WeightsFormat
 luci_weights_format(const circle::FullyConnectedOptionsWeightsFormat weights_format);
 std::unique_ptr<CircleQuantParam>
-luci_quantparam(const circle::QuantizationParametersT *quantization);
-std::unique_ptr<CircleQuantParam>
 luci_quantparam(const circle::QuantizationParameters *quantization);
 
 /// @brief Copy common tensor attributes such as name, type, etc. to node.
-void copy_tensor_attributes(const circle::TensorT &tensor, CircleNode *node);
 void copy_tensor_attributes(const circle::Tensor *tensor, CircleNode *node);
 
 /**
@@ -101,54 +89,28 @@ template <typename T> VectorWrapper<T> wrap(const flatbuffers::Vector<T> *vec)
  */
 class CircleReader
 {
-private: // unpack API
-  using CircleBuffers_t = std::vector<std::unique_ptr<circle::BufferT>>;
-  using CircleTensors_t = std::vector<std::unique_ptr<circle::TensorT>>;
-  using CircleOperators_t = std::vector<std::unique_ptr<circle::OperatorT>>;
-  using CircleOperatorCodes_t = std::vector<std::unique_ptr<circle::OperatorCodeT>>;
-  using CircleMetadata_t = std::vector<std::unique_ptr<circle::MetadataT>>;
-
-private: // direct API
+private:
   using CircleBuffers = VectorWrapper<flatbuffers::Offset<circle::Buffer>>;
   using CircleTensors = VectorWrapper<flatbuffers::Offset<circle::Tensor>>;
   using CircleOperators = VectorWrapper<flatbuffers::Offset<circle::Operator>>;
   using CircleOperatorCodes = VectorWrapper<flatbuffers::Offset<circle::OperatorCode>>;
   using CircleMetadataSet = VectorWrapper<flatbuffers::Offset<circle::Metadata>>;
 
-  using CircleSubGraphsPtr_t = flatbuffers::Vector<flatbuffers::Offset<circle::SubGraph>>;
-  using CircleTensorsPtr_t = flatbuffers::Vector<flatbuffers::Offset<circle::Tensor>>;
-
 public:
   CircleReader() = default;
 
-public: // unpack API
-  const CircleOperatorCodes_t &opcodes() const { return _model->operator_codes; }
-  const CircleBuffers_t &buffers() const { return _model->buffers; }
-  const CircleTensors_t &tensors() const { return _current_subgraph->tensors; }
-  const CircleOperators_t &operators() const { return _current_subgraph->operators; }
-  const std::vector<int32_t> &inputs() const { return _current_subgraph->inputs; }
-  const std::vector<int32_t> &outputs() const { return _current_subgraph->outputs; }
-  const std::string &name() const { return _current_subgraph->name; }
-  const circle::DataFormat &data_format() const { return _current_subgraph->data_format; }
-  const CircleMetadata_t &metadata() const { return _model->metadata; }
+public:
+  CircleOperatorCodes opcodes() const { return wrap(_model->operator_codes()); }
+  CircleBuffers buffers() const { return wrap(_model->buffers()); }
+  CircleTensors tensors() const { return wrap(_current_subgraph->tensors()); }
+  CircleOperators operators() const { return wrap(_current_subgraph->operators()); }
+  VectorWrapper<int32_t> inputs() const { return wrap(_current_subgraph->inputs()); }
+  VectorWrapper<int32_t> outputs() const { return wrap(_current_subgraph->outputs()); }
+  std::string name() const { return _current_subgraph->name()->str(); }
+  circle::DataFormat data_format() const { return _current_subgraph->data_format(); }
+  CircleMetadataSet metadata() const { return wrap(_model->metadata()); }
 
-  const CircleTensorsPtr_t *tensors_ptr() const { return _native_tensors; }
-
-  uint32_t num_subgraph() const { return _model->subgraphs.size(); }
-
-  circle::BuiltinOperator builtin_code(const circle::OperatorT &op) const;
-  std::string opcode_name(const circle::OperatorT &op) const;
-
-public: // direct API
-  CircleOperatorCodes native_opcodes() const { return wrap(_native_model->operator_codes()); }
-  CircleBuffers native_buffers() const { return wrap(_native_model->buffers()); }
-  CircleTensors native_tensors() const { return wrap(_native_subgraph->tensors()); }
-  CircleOperators native_operators() const { return wrap(_native_subgraph->operators()); }
-  VectorWrapper<int32_t> native_inputs() const { return wrap(_native_subgraph->inputs()); }
-  VectorWrapper<int32_t> native_outputs() const { return wrap(_native_subgraph->outputs()); }
-  std::string native_name() const { return _native_subgraph->name()->str(); }
-  circle::DataFormat native_data_format() const { return _native_subgraph->data_format(); }
-  CircleMetadataSet native_metadata() const { return wrap(_native_model->metadata()); }
+  uint32_t num_subgraph() const { return wrap(_model->subgraphs()).size(); }
 
   circle::BuiltinOperator builtin_code(const circle::Operator *op) const;
   std::string opcode_name(const circle::Operator *op) const;
@@ -158,12 +120,8 @@ public:
   bool select_subgraph(uint32_t subgraph);
 
 private:
-  std::unique_ptr<const circle::ModelT> _model;
-  const circle::SubGraphT *_current_subgraph{nullptr};
-
-  const circle::Model *_native_model{nullptr};
-  const circle::SubGraph *_native_subgraph{nullptr};
-  const CircleTensorsPtr_t *_native_tensors{nullptr};
+  const circle::Model *_model{nullptr};
+  const circle::SubGraph *_current_subgraph{nullptr};
 };
 
 } // namespace luci
--- a/compiler/luci/import/include/luci/Import/CircleReader.h
+++ b/compiler/luci/import/include/luci/Import/CircleReader.h
@@ -36,10 +36,19 @@ namespace luci
 {
 
 bool is_valid(const circle::OperatorCodeT &opcode);
+bool is_valid(const circle::OperatorCode *opcode);
+
 bool is_custom(const circle::OperatorCodeT &opcode);
+bool is_custom(const circle::OperatorCode *opcode);
+
 std::string opcode_name(const circle::OperatorCodeT &opcode);
+std::string opcode_name(const circle::OperatorCode *opcode);
+
 const char *tensor_name(const circle::TensorT &tensor);
+const char *tensor_name(const circle::Tensor *tensor);
+
 const circle::QuantizationParametersT *tensor_quantization(const circle::TensorT &tensor);
+const circle::QuantizationParameters *tensor_quantization(const circle::Tensor *tensor);
 
 loco::DataType luci_datatype(circle::TensorType type);
 FusedActFunc luci_actfunc(const circle::ActivationFunctionType type);
@@ -49,21 +58,62 @@ luci::CircleFullyConnected::WeightsFormat
 luci_weights_format(const circle::FullyConnectedOptionsWeightsFormat weights_format);
 std::unique_ptr<CircleQuantParam>
 luci_quantparam(const circle::QuantizationParametersT *quantization);
+std::unique_ptr<CircleQuantParam>
+luci_quantparam(const circle::QuantizationParameters *quantization);
 
 /// @brief Copy common tensor attributes such as name, type, etc. to node.
 void copy_tensor_attributes(const circle::TensorT &tensor, CircleNode *node);
+void copy_tensor_attributes(const circle::Tensor *tensor, CircleNode *node);
+
+/**
+ * @brief Wrapper to use flatbuffers::Vector pointer as std::vector entity
+ */
+template <typename T> class VectorWrapper
+{
+public:
+  explicit VectorWrapper(const flatbuffers::Vector<T> *ptr);
+
+  const T *data() const;
+  uint32_t size() const;
+
+  using iterator = typename flatbuffers::Vector<T>::const_iterator;
+  iterator begin() const;
+  iterator end() const;
+
+  using value_type = typename flatbuffers::Vector<T>::return_type;
+  value_type at(uint32_t i) const;
+  value_type operator[](uint32_t i) const;
+
+  bool null() const;
+  bool empty() const;
+
+private:
+  const flatbuffers::Vector<T> *_vector;
+};
+
+template <typename T> VectorWrapper<T> wrap(const flatbuffers::Vector<T> *vec)
+{
+  return VectorWrapper<T>(vec);
+}
 
 /**
  * @brief Loads Circle file and provides helpers to access attributes
  */
 class CircleReader
 {
-private:
+private: // unpack API
   using CircleBuffers_t = std::vector<std::unique_ptr<circle::BufferT>>;
   using CircleTensors_t = std::vector<std::unique_ptr<circle::TensorT>>;
   using CircleOperators_t = std::vector<std::unique_ptr<circle::OperatorT>>;
   using CircleOperatorCodes_t = std::vector<std::unique_ptr<circle::OperatorCodeT>>;
   using CircleMetadata_t = std::vector<std::unique_ptr<circle::MetadataT>>;
+
+private: // direct API
+  using CircleBuffers = VectorWrapper<flatbuffers::Offset<circle::Buffer>>;
+  using CircleTensors = VectorWrapper<flatbuffers::Offset<circle::Tensor>>;
+  using CircleOperators = VectorWrapper<flatbuffers::Offset<circle::Operator>>;
+  using CircleOperatorCodes = VectorWrapper<flatbuffers::Offset<circle::OperatorCode>>;
+  using CircleMetadataSet = VectorWrapper<flatbuffers::Offset<circle::Metadata>>;
 
   using CircleSubGraphsPtr_t = flatbuffers::Vector<flatbuffers::Offset<circle::SubGraph>>;
   using CircleTensorsPtr_t = flatbuffers::Vector<flatbuffers::Offset<circle::Tensor>>;
@@ -71,7 +121,7 @@ private:
 public:
   CircleReader() = default;
 
-public:
+public: // unpack API
   const CircleOperatorCodes_t &opcodes() const { return _model->operator_codes; }
   const CircleBuffers_t &buffers() const { return _model->buffers; }
   const CircleTensors_t &tensors() const { return _current_subgraph->tensors; }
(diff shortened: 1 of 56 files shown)


compiler/luci/import/include/luci/Import/CircleReader.h
```diff
@@ -35,11 +35,11 @@
 namespace luci
 {
 
-bool is_valid(const circle::OperatorCodeT &opcode);
-bool is_custom(const circle::OperatorCodeT &opcode);
-std::string opcode_name(const circle::OperatorCodeT &opcode);
-const char *tensor_name(const circle::TensorT &tensor);
-const circle::QuantizationParametersT *tensor_quantization(const circle::TensorT &tensor);
+bool is_valid(const circle::OperatorCode *opcode);
```
seanshpark, line 38: could you please explain why this is needed?

compiler/luci/import/include/luci/Import/CircleReader.h
```diff
@@ -48,57 +48,80 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode);
...
+luci_quantparam(const circle::QuantizationParameters *quantization);
 
 /// @brief Copy common tensor attributes such as name, type, etc. to node.
-void copy_tensor_attributes(const circle::TensorT &tensor, CircleNode *node);
+void copy_tensor_attributes(const circle::Tensor *tensor, CircleNode *node);
+
+/**
+ * @brief Wrapper to use flatbuffers::Vector pointer as std::vector entity
+ */
+template <typename T> class VectorWrapper
```
seanshpark, line 59: this too about `VectorWrapper`

compiler/luci/import/include/luci/Import/CircleReader.h
```diff
@@ -48,57 +48,80 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode);
...
-  const CircleOperatorCodes_t &opcodes() const { return _model->operator_codes; }
-  const CircleBuffers_t &buffers() const { return _model->buffers; }
-  const CircleTensors_t &tensors() const { return _current_subgraph->tensors; }
-  const CircleOperators_t &operators() const { return _current_subgraph->operators; }
-  const std::vector<int32_t> &inputs() const { return _current_subgraph->inputs; }
-  const std::vector<int32_t> &outputs() const { return _current_subgraph->outputs; }
-  const std::string &name() const { return _current_subgraph->name; }
-  const circle::DataFormat &data_format() const { return _current_subgraph->data_format; }
-  const CircleMetadata_t &metadata() const { return _model->metadata; }
+  CircleOperatorCodes opcodes() const { return wrap(_model->operator_codes()); }
```
seanshpark, line 103: Does this return a copy of `CircleOperatorCodes` ? If so, please avoid this.

compiler/luci/import/src/CircleReader.cpp
```diff
@@ -257,33 +276,46 @@ std::unique_ptr<SparsityParam> luci_sparsityparam(const circle::SparsityParamete
...
+  return luci_sparsityparam(&sparsity);
+}
+
+void copy_tensor_attributes(const circle::Tensor *tensor, CircleNode *node)
+{
+  assert(tensor != nullptr);
+
   node->name(tensor_name(tensor));
-  node->dtype(luci_datatype(tensor.type));
+  node->dtype(luci_datatype(tensor->type()));
```
seanshpark, line 294: this too

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: If you ask about removing functions from old API, I mark this changes as optional. I guess reviewer will check introduced changes by commits to better understanding, but if this changes makes confuse, I could remove them from draft.

If you ask about functions duplication, I made copies of functions which works with pointers from direct access API instead UnPacked since I would like to minimize usage of UnPacked API with this draft without changes in logic of importer. 

-------------

If I haven't answered your question, could you please clarify it? =)

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: I made this struct to minimize code changes. Let me explain the reason with some simple example! 
Currently, we use some methods of reader (e.g. `inputs()`), which returns references to `std::vector`, which created by model's `UnPack`.

https://github.com/Samsung/ONE/blob/393385e4a29627b7143d509866b4452bbba0de9f/compiler/luci/import/src/Importer.cpp#L76

But usage direct access API return **pointer** to vector, which could be equal to `nullptr`. So, before using of range based for loop we need make each time following actions:
1. Check result is not `nullptr`
2. Deref pointer.  
3. replace all `at(n)` usages to `Get()` usages.

In order to minimize changes, I made safe `std::vector`-like wrapper, which helps to avoid many `nullptr` checks and code replacements.

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: This returns copy of `CircleOperatorCodes`, but `CircleOperatorCodes` is just a alias for `VectorWrapper<>`, which stores only one pointer to `flatbuffers::Vector` from model's buffer. So, weight of this struct is size of pointer. 

```cpp
  using CircleOperatorCodes = VectorWrapper<flatbuffers::Offset<circle::OperatorCode>>;
```

compiler/luci/import/src/CircleReader.cpp
m-bronnikov: Unfortunately, I don't know how to avoid  this changes (using getters of direct structures instead attributes of unpacked). This changes with asserts on `nullptr` represents the largest part of this pull request =(

compiler/luci/import/include/luci/Import/CircleReader.h
seanshpark: Sorry for ambigious question..
You've change from using reference to pointer. I guess this is not to use `Unpack`..

compiler/luci/import/include/luci/Import/CircleReader.h
```diff
@@ -48,57 +48,80 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode);
...
+  using iterator = typename flatbuffers::Vector<T>::const_iterator;
+  iterator begin() const;
+  iterator end() const;
+
+  using value_type = typename flatbuffers::Vector<T>::return_type;
+  value_type at(uint32_t i) const;
+  value_type operator[](uint32_t i) const;
+
+  bool is_null() const;
+  bool empty() const;
```
seanshpark, line 76: as you used `is_null` to get is null or not, could you change `empty` to `is_empty` if this is to check empty or not?

compiler/luci/import/include/luci/Import/CircleReader.h
seanshpark: OK thanks, I got it.

seanshpark:
@m-bronnikov , 
(1) you posted with small changes in this draft PR, sorry for not reading one-by-one, but is' hard to do that for me.
(2) only the last commit has the green check, I don't think you wanted to show me each commit won't break the build

What I wanted was 
(1) you progressive change won't break the build
(2) to see and discuess/agree on the changes (not for all IRs but with some IRs revised as you want)
(3) without the changes that are not related what you want; those can be put to another issue/PR

compiler/luci/import/include/luci/Import/CircleReader.h
```diff
@@ -48,57 +48,80 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode);
...
-  const std::string &name() const { return _current_subgraph->name; }
-  const circle::DataFormat &data_format() const { return _current_subgraph->data_format; }
-  const CircleMetadata_t &metadata() const { return _model->metadata; }
+  CircleOperatorCodes opcodes() const { return wrap(_model->operator_codes()); }
+  CircleBuffers buffers() const { return wrap(_model->buffers()); }
+  CircleTensors tensors() const { return wrap(_current_subgraph->tensors()); }
+  CircleOperators operators() const { return wrap(_current_subgraph->operators()); }
+  VectorWrapper<int32_t> inputs() const { return wrap(_current_subgraph->inputs()); }
+  VectorWrapper<int32_t> outputs() const { return wrap(_current_subgraph->outputs()); }
+  const char *name() const { return _current_subgraph->name()->c_str(); }
```
seanshpark, line 109: what is the reason to use `char *` and not `std::string` ?

compiler/luci/import/include/luci/Import/CircleReader.h
```diff
@@ -48,57 +48,80 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode);
...
+  circle::DataFormat data_format() const { return _current_subgraph->data_format(); }
+  CircleMetadataSet metadata() const { return wrap(_model->metadata()); }
 
-  const CircleTensorsPtr_t *tensors_ptr() const { return _tensors_ptr; }
+  circle::BuiltinOperator builtin_code(const circle::Operator *op) const;
```
seanshpark, line 113: why is `const CircleTensorsPtr_t *tensors_ptr() const { return _tensors_ptr; }` gone?

compiler/luci/import/src/CircleImportMetadata.cpp
```diff
@@ -21,8 +21,10 @@
 namespace
 {
 
-uint32_t read_u32(const std::vector<uint8_t> &buffer, uint32_t idx)
+template <typename VectorType> uint32_t read_u32(const VectorType &buffer, uint32_t idx)
```
seanshpark, line 24: `typename VectorType` : `VectorType` makes confusion reading the code. I was searching `VectorType` at first sight.
 It would be better to use `VECTORTYPE`

seanshpark: This seems one of the changes that don't fit in with `Use direct access API` or is it?

compiler/luci/import/src/CircleReader.cpp
```diff
@@ -257,33 +276,46 @@ std::unique_ptr<SparsityParam> luci_sparsityparam(const circle::SparsityParamete
   return sparsityparam;
 }
 
-void copy_tensor_attributes(const circle::TensorT &tensor, CircleNode *node)
+std::unique_ptr<SparsityParam> luci_sparsityparam(const circle::SparsityParameters *sparparam)
```
seanshpark, line 279: why is this added?

compiler/luci/import/src/CircleReader.cpp
```diff
@@ -257,33 +276,46 @@ std::unique_ptr<SparsityParam> luci_sparsityparam(const circle::SparsityParamete
...
+  auto const tensor_shape = wrap(tensor->shape());
+  assert(tensor_shape_signature.is_null() || tensor_shape_signature.size() == tensor_shape.size());
 
-  std::vector<int32_t> dims = tensor.shape; // in NHWC
+  auto const dims = tensor_shape; // in NHWC
   node->rank(dims.size());
   for (uint32_t r = 0; r < dims.size(); ++r)
   {
-    if (tensor.shape_signature.size() > 0 && tensor.shape_signature.at(r) == -1)
+    if (not tensor_shape_signature.is_null() && tensor_shape_signature.at(r) == -1)
```
seanshpark, line 304: do we have to change `tensor.shape_signature.size() > 0` to `(not tensor_shape_signature.is_null()` ?

compiler/luci/import/src/CircleReader.cpp
```diff
@@ -257,33 +276,46 @@ std::unique_ptr<SparsityParam> luci_sparsityparam(const circle::SparsityParamete
...
       node->dim(r).unset();
     else
       node->dim(r).set(dims[r]);
   }
 
-  const auto *quantization = tensor.quantization.get();
+  auto const quantization = tensor->quantization();
```
seanshpark, line 310: `const auto *quantization` was to explicitly express it's a pointer and still `tensor->quantization();` seems too.
if this is so, please remove these kind of changes.

compiler/luci/import/src/GraphBuilder.cpp
```diff
@@ -29,10 +29,9 @@ CircleNode *GraphBuilder::build(const circle::OperatorT &op, GraphBuilderContext
 
   const std::vector<int32_t> &inputs = op.inputs;
   const std::vector<int32_t> &outputs = op.outputs;
-  const auto &tensors = context->reader()->tensors();
-  const auto &opcodes = context->reader()->opcodes();
-  auto tensors_ptr = context->reader()->tensors_ptr();
-  assert(tensors_ptr != nullptr);
+  const auto opcodes = context->reader()->opcodes();
+  const auto tensors = context->reader()->tensors();
```
seanshpark, line 33: do you have to change to order two variableS?

compiler/luci/import/src/Nodes/CircleConst.cpp
```diff
@@ -30,10 +30,12 @@
 namespace
 {
 
-std::ostream &operator<<(std::ostream &os, const std::vector<int32_t> &vect)
+template <typename VectorType>
+std::enable_if_t<std::is_same<typename VectorType::value_type, int32_t>::value, std::ostream &>
+operator<<(std::ostream &os, const VectorType &vect)
```
seanshpark, line 35: could you explain why use `enable_if_t` ?

compiler/luci/import/src/Nodes/CircleConst.cpp
```diff
@@ -181,12 +186,12 @@ CircleConst *create_circleconst(GraphBuilderContext *context, int32_t tensor_ind
         break;
 
       case loco::DataType::STRING:
-        copy_data<loco::DataType::STRING>(buffer, num_elements, const_node);
+        copy_string(buffer, num_elements, const_node);
```
seanshpark, line 189: I don't think this is related with `Use direct access API`, or is it?

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: I made `empty()` method because `std::vector<>` has `empty()` method too. How about rename `is_null()` to `null()` or something like this instead?

m-bronnikov:
> @m-bronnikov , (1) you posted with small changes in this draft PR, sorry for not reading one-by-one, but is' hard to do that for me. (2) only the last commit has the green check, I don't think you wanted to show me each commit won't break the build
> 
> What I wanted was (1) you progressive change won't break the build (2) to see and discuess/agree on the changes (not for all IRs but with some IRs revised as you want) (3) without the changes that are not related what you want; those can be put to another issue/PR

@seanshpark Thank you for your review!  Changes, splitted by commits here should not break build (I checked this locally for each commit).

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: I removed this method as unused.

Before this chages was 2 different methods: 
1. `tensors()` which returns reference to `std::vector` of UnPacked tensors (`TensorT`).
2. `tensors_ptr()` which returns pointer to `flatbuffers::Vector` of direct Tensors (`Tensor *`)

This changes replaces `tensors()` method to provide direct Tesnors, so, IMHO `tensors_ptr` looks unused and confused.

compiler/luci/import/src/CircleImportMetadata.cpp
m-bronnikov: It's necessary to change  this function to work with `VectorWrapper` instead `std::vector`. I see 3 ways to do that:

1. Just change type of argument (`std::vector<uint8>` -> `VectorWrapper<uin8_t>`). 
2. Add copy of this function with `VectorWrapper` as argument. There is a lot of new lines will be added =(
3. Make this function as template.

I chose №3, but I think we could take approach №1 too, if this option looks better for rerview.

compiler/luci/import/src/CircleReader.cpp
m-bronnikov: Sparsity param needed above:

https://github.com/Samsung/ONE/blob/eda824dab4f2e547760de161e9f935177796f33c/compiler/luci/import/src/CircleReader.cpp#L318-L321

`copy_tensor_attributes` removed because this function unused.

compiler/luci/import/src/CircleReader.cpp
m-bronnikov: Ah, I think this change is really looks unrelated, I will reduce this, sorry.

--------------

**FYI** The reason why I changes logic here: this condition checks that `shape_signature` must have value `-1` if signature exist. In case signature not exist `tensor->shape_signature == nullptr`, but with `UnPacked` API we don't have ability to check `nullptr`, because works with `std::vector` reference instead pointer. I think with direct access structure check of `nullptr` looks more natural. 

Anyway, I think better to introduce this changes later as unrelated for current draft.

compiler/luci/import/src/GraphBuilder.cpp
m-bronnikov: Yes, I will reduce this kind of changes too.

compiler/luci/import/src/Nodes/CircleConst.cpp
m-bronnikov: I changed this function to template with same rationale as in https://github.com/Samsung/ONE/pull/7901/files#r740778845. 

But this definition without `enable_if`  redefines `operator<<` definition for many other entities, e.g. `std::string`. So, I used `enable_if_t` here. 

--------

I think in this case better to just changes type of operator's argument, because this change looks confusing. Let's introduce tenplate impl later if needed, Thank you.

compiler/luci/import/src/Nodes/CircleConst.cpp
m-bronnikov: Yes, let me reduce this change too.

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: The reason was:
1. not make copy of std::string at each call of this function.
2. build not breaks with `const char *` =)

I can change this method to return std::string.

compiler/luci/import/include/luci/Import/CircleReader.h
seanshpark: `null()` seems OK

compiler/luci/import/include/luci/Import/CircleReader.h
seanshpark: Thanks for sharing your thoughts.
(1) `const char *` is fast, aligns with current revision. but may have to check nullptr and may imply dangers
(2) `std::string` is slower than `const char *` but is safe to use
As our whole system uses lots of `std::string` for safety... I think I give +1 to `std::string`.

compiler/luci/import/src/CircleImportMetadata.cpp
seanshpark: OK thanks for the explanation. So to use `VectorWrapper` which I understand as good helper, this change is also a good helper. I think we can go as is like the draft code here.

seanshpark:
@m-bronnikov , thanks for the replies :)

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: Done.

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: Done.

compiler/luci/import/src/CircleImportMetadata.cpp
m-bronnikov: > `typename VectorType` : `VectorType` makes confusion reading the code. I was searching `VectorType` at first sight.
> It would be better to use `VECTORTYPE`

Done.

compiler/luci/import/src/CircleReader.cpp
m-bronnikov: Returned to `> 0` check.

compiler/luci/import/src/CircleReader.cpp
m-bronnikov: Done.

compiler/luci/import/src/GraphBuilder.cpp
m-bronnikov: Source order is returned.

compiler/luci/import/src/Nodes/CircleConst.cpp
m-bronnikov: Made as not template solution.

compiler/luci/import/src/Nodes/CircleConst.cpp
m-bronnikov: Done.

m-bronnikov:
@seanshpark I reduced unrelated changes in draft. Could we start to make PRs?

compiler/luci/import/include/luci/Import/CircleReader.h
```diff
@@ -48,57 +48,80 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode);
...
 private:
-  using CircleBuffers_t = std::vector<std::unique_ptr<circle::BufferT>>;
-  using CircleTensors_t = std::vector<std::unique_ptr<circle::TensorT>>;
-  using CircleOperators_t = std::vector<std::unique_ptr<circle::OperatorT>>;
-  using CircleOperatorCodes_t = std::vector<std::unique_ptr<circle::OperatorCodeT>>;
-  using CircleMetadata_t = std::vector<std::unique_ptr<circle::MetadataT>>;
-
-  using CircleSubGraphsPtr_t = flatbuffers::Vector<flatbuffers::Offset<circle::SubGraph>>;
-  using CircleTensorsPtr_t = flatbuffers::Vector<flatbuffers::Offset<circle::Tensor>>;
+  using CircleBuffers = VectorWrapper<flatbuffers::Offset<circle::Buffer>>;
```
seanshpark, line 93: `CircleBuffers_t` has change to `CircleBuffers`, is there any reason to do so?

compiler/luci/import/include/luci/Import/CircleReader.h
```diff
@@ -48,57 +48,80 @@ MirrorPadMode luci_mirrorpad_mode(const circle::MirrorPadMode mode);
...
+  CircleOperators operators() const { return wrap(_current_subgraph->operators()); }
+  VectorWrapper<int32_t> inputs() const { return wrap(_current_subgraph->inputs()); }
+  VectorWrapper<int32_t> outputs() const { return wrap(_current_subgraph->outputs()); }
+  std::string name() const { return _current_subgraph->name()->str(); }
+  circle::DataFormat data_format() const { return _current_subgraph->data_format(); }
+  CircleMetadataSet metadata() const { return wrap(_model->metadata()); }
 
-  const CircleTensorsPtr_t *tensors_ptr() const { return _tensors_ptr; }
+  circle::BuiltinOperator builtin_code(const circle::Operator *op) const;
+  std::string opcode_name(const circle::Operator *op) const;
```
seanshpark, line 114: this line has change in order. please put where it was.

compiler/luci/import/src/CircleReader.cpp
```diff
@@ -292,24 +325,29 @@ void copy_tensor_attributes(const circle::TensorT &tensor, CircleNode *node)
...
+  assert(op != nullptr);
+
+  const auto op_codes = opcodes();
+  uint32_t index = op->opcode_index();
   assert(index < op_codes.size());
-  const circle::OperatorCodeT &opcode = *op_codes[index];
+  const auto opcode = op_codes[index];
 
-  if (!is_valid(opcode))
+  if (not is_valid(opcode))
```
seanshpark, line 350: please put these kind of changes to other title

compiler/luci/import/src/GraphBuilder.cpp
```diff
@@ -60,16 +59,19 @@ CircleNode *GraphBuilder::build(const circle::OperatorT &op, GraphBuilderContext
   // Set up node parameters.
   assert(outputs.size() == 1);
   {
-    const circle::TensorT &output_tensor = *tensors[outputs[0]];
+    const auto output_tensor = tensors[outputs[0]];
+    assert(output_tensor != nullptr);
+
```
seanshpark, line 64: here and below, there are lots of this kind of adding lines.
I don't think there is one correct way but it would be better not to insert these on this title.
they enhance readability and I personally like these but like others, the seem to distract reading changes.

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: Ah, there is no reason. First commit of draft introduces `CircleBuffers` alias, but `CircleBuffers_t` is removed only in last commit. I can add another commit (or update latest) that will add `_t` suffix.

compiler/luci/import/src/GraphBuilder.cpp
m-bronnikov: Okay, I will reduce this kind of changes too.

compiler/luci/import/src/CircleReader.cpp
m-bronnikov: Sorry, I missed this during rebase.

compiler/luci/import/include/luci/Import/CircleReader.h
m-bronnikov: Q. If I understand correct `_t` means *type*. Could you please clarify I'm right?

seanshpark: > Q. If I understand correct _t means type. Could you please clarify I'm right?

Yes, it's for `type`. This _t was used to figure out quickly it's a type (not struct nor class) but there exist lots other types don't use _t. I think it is OK to move them now...

m-bronnikov:
Rebase with changes! Changed:

1. Removed some empty lines around `assert`s.
2. `not` -> `!` in two `if`
3. Returned source order of methods in CircleReader.

Before: 
image

As is:
image

Also, this draft rebased to fresh master in order to ensure current changes is valid.

Note: I use rebase with `push --force` to not break structure of commits, which I would like to split by PRs.

compiler/luci/import/src/CircleReader.cpp
```diff
@@ -292,22 +325,27 @@ void copy_tensor_attributes(const circle::TensorT &tensor, CircleNode *node)
   }
 }
 
-circle::BuiltinOperator CircleReader::builtin_code(const circle::OperatorT &op) const
+circle::BuiltinOperator CircleReader::builtin_code(const circle::Operator *op) const
 {
-  const auto &op_codes = opcodes();
-  uint32_t index = op.opcode_index;
+  assert(op != nullptr);
+
```
m-bronnikov, line 331: I guess this kind of empty lines not confuses reviewer.

m-bronnikov: In other cases I tried to reduce amount of empty lines near `assert`s.

seanshpark:
@m-bronnikov , thanks for the change :) Please send PRs.
There may be comments when reviewers do some more detailed review.

m-bronnikov:
This draft finally published by PR's, let's close.

Issue #7831, opened 2021-10-14
[luci/plan] Add unit tests


seanshpark:
We need to add unit tests for `luci/plan`. Also 1:1 negative tests, that is one ore more negative test(s) for one positive test.

- [x] Tests for luci/plan #7883 
- [x] Tests for luci/import #7903 
- [x] Tests for luci/export #7884

seanshpark:
I think we can close this issue. Please reopen if necessary.

PR #7903, opened 2021-10-29
[luci] Test for import execution plan

binarman:
This PR adds unit test for execution plan import.

Related issue: #7831 

### DIFF
--- a/compiler/luci/import/CMakeLists.txt
+++ b/compiler/luci/import/CMakeLists.txt
@@ -36,6 +36,3 @@ target_link_libraries(luci_import_test oops)
 target_link_libraries(luci_import_test luci_plan)
 target_link_libraries(luci_import_test luci_lang)
 target_link_libraries(luci_import_test mio_circle)
-
-nnas_find_package(GTest REQUIRED)
-
--- a/compiler/luci/import/CMakeLists.txt
+++ b/compiler/luci/import/CMakeLists.txt
@@ -33,3 +33,9 @@ GTest_AddTest(luci_import_test ${TESTS})
 target_include_directories(luci_import_test PRIVATE src)
 target_link_libraries(luci_import_test luci_import)
 target_link_libraries(luci_import_test oops)
+target_link_libraries(luci_import_test luci_plan)
+target_link_libraries(luci_import_test luci_lang)
+target_link_libraries(luci_import_test mio_circle)
+
+nnas_find_package(GTest REQUIRED)
+
--- a/compiler/luci/import/src/Importer.test.cpp
+++ b/compiler/luci/import/src/Importer.test.cpp
@@ -16,13 +16,289 @@
 
 #include "luci/Importer.h"
 
-#include <loco.h>
+#include <luci/IR/CircleNode.h>
+#include <luci/Plan/CircleNodeExecutionPlan.h>
 
 #include <gtest/gtest.h>
+#include <mio/circle/schema_generated.h>
+#include <flatbuffers/flatbuffers.h>
 
 TEST(TensorFlowLiteImport, Dummy)
 {
   luci::Importer import;
 
   SUCCEED();
+}
+
+// helpers for flatbuffers
+namespace
+{
+
+struct BasicCircleModel
+{
+  std::unique_ptr<circle::ModelT> model;
+
+  BasicCircleModel()
+  {
+    model = std::make_unique<circle::ModelT>();
+    model->buffers.emplace_back(new circle::BufferT);
+    model->description = "nnpackage";
+    model->version = 0;
+  }
+
+  uint32_t add_subgraph()
+  {
+    model->subgraphs.emplace_back(new circle::SubGraphT);
+    model->subgraphs.back()->name = "";
+    model->subgraphs.back()->data_format = circle::DataFormat_CHANNELS_LAST;
+    return model->subgraphs.size() - 1;
+  }
+
+  void add_subgraph_inputs(uint32_t subgraph_id, const std::vector<uint32_t> &inputs)
+  {
+    model->subgraphs[subgraph_id]->inputs.assign(inputs.begin(), inputs.end());
+  }
+
+  void add_subgraph_outputs(uint32_t subgraph_id, const std::vector<uint32_t> &outputs)
+  {
+    model->subgraphs[subgraph_id]->outputs.assign(outputs.begin(), outputs.end());
+  }
+
+  uint32_t add_builtin_opcode(circle::BuiltinOperator opcode)
+  {
+    uint32_t id = model->operator_codes.size();
+    model->operator_codes.emplace_back(new circle::OperatorCodeT);
+    model->operator_codes[id]->builtin_code = opcode;
+    model->operator_codes[id]->version = 1;
+    return id;
+  }
+
+  uint32_t add_buffer()
+  {
+    model->buffers.emplace_back(new circle::BufferT);
+    return model->buffers.size() - 1;
+  }
+
+  uint32_t add_float_tensor(uint32_t graph_id, const std::vector<int32_t> &shape,
+                            uint32_t buffer_id)
+  {
+    auto &graph = model->subgraphs[graph_id];
+    uint32_t idx = graph->tensors.size();
+    graph->tensors.emplace_back(new circle::TensorT);
+    graph->tensors[idx]->shape = shape;
+    graph->tensors[idx]->type = circle::TensorType_FLOAT32;
+    graph->tensors[idx]->buffer = buffer_id;
+    graph->tensors[idx]->name = std::to_string(idx);
+    graph->tensors[idx]->quantization.reset(nullptr);
+    graph->tensors[idx]->is_variable = false;
+    graph->tensors[idx]->sparsity.reset(nullptr);
+    (void)graph->tensors[idx]->shape_signature;
+    return idx;
+  }
+
+  uint32_t add_builtin_operator(uint32_t graph_id, uint32_t opcode_id,
+                                const std::vector<uint32_t> &inputs,
+                                const std::vector<uint32_t> &outputs)
+  {
+    auto &graph = model->subgraphs[graph_id];
+    auto idx = graph->operators.size();
+    graph->operators.emplace_back(new circle::OperatorT);
+    graph->operators[idx]->opcode_index = opcode_id;
+    graph->operators[idx]->inputs.assign(inputs.begin(), inputs.end());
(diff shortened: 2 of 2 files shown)


compiler/luci/import/src/Importer.test.cpp
```diff
@@ -16,13 +16,289 @@
...
+        break;
+      }
+    }
+  }
+}
+
+/**
+ * This test checks that model with incomplete execution plan is successfully imported
+ */
+TEST(TensorFlowLiteImporter, DISABLED_incomplete_plan_NEG)
```
binarman, line 236: This test is questionable.
The problem with it is following:
- Exporter theoretically can produce models with incomplete execution plan (no data for some operations)
- Importer throws on attempt to import such models.

From my point of view this is inconsistency of interfaces, we need to decide, should we:
1. Throw in exporter in case of incomplete plan?
2. Successfully import models with some data missing?

seanshpark: > Throw in exporter in case of incomplete plan?

I disagree on this. As I understand `plan` is optional, and I don't know how to use this `plan`

binarman: > I disagree on this. As I understand plan is optional, and I don't know how to use this plan

You are right, it is optional.

There are three cases:
- All nodes has plan metadata
- All nodes does not have plan metadata
- Some nodes has, some does not

First two cases are fine, third one is inconsistent:
- Exporter can produce such circle model
- Importer can not parse it =(

I just thought that there are third solution:
- discard all plan annotation in case some of the them are missing

binarman: I think serialize plan partially (2 case from original post) is the best solution.

- Exception case. After applying planning tool we can not do any transformations. Otherwise we need to explicitly remove plan annotations. It is not clear for me who should do this cleaning (this is also hard to describe to user :slightly_smiling_face: )
- Implicit discarding. This is also hard do describe and IMHO a little confusing

binarman: Lets merge this PR, then fix execution plan importing and enable this test.
Related issue #7907

seanshpark (dismissed):
LGTM

BalyshevArtem (approved):
LGTM

binarman:
@m-bronnikov ping?

compiler/luci/import/src/Importer.test.cpp
```diff
@@ -16,13 +16,289 @@
...
+  void add_plan_entry(uint32_t plan_buffer_id, uint32_t execution_order,
+                      const std::vector<uint32_t> &offsets)
+  {
+    auto &buffer = model->buffers[plan_buffer_id]->data;
+    auto old_size = buffer.size();
+    assert(old_size % 4 == 0);
+    assert(old_size > 0);
+
+    // Allocate space for new entry
+    // +4 from entry id, +4 from entry size, +4 from execution order + offsets.size() * 4 for
```
m-bronnikov, line 136: May be?

```suggestion
    // +4 for entry id, +4 for entry size, +4 for execution order + offsets.size() * 4 for
```

compiler/luci/import/src/Importer.test.cpp
```diff
@@ -16,13 +16,289 @@
...
+    // offsets
+    buffer.resize(old_size + 12 + offsets.size() * 4);
+    uint32_t *number_of_entries_ptr = reinterpret_cast<uint32_t *>(buffer.data());
+    *number_of_entries_ptr += 1;
+
+    uint32_t *entry_data_ptr = reinterpret_cast<uint32_t *>(buffer.data() + old_size);
+
+    entry_data_ptr[0] = *number_of_entries_ptr - 1; // entry id
+    entry_data_ptr[1] = 1 + offsets.size();         // entry size
+    entry_data_ptr[2] = execution_order;            // entry size
```
m-bronnikov, line 146: May be?

```suggestion
    entry_data_ptr[2] = execution_order;            // execution order
```

compiler/luci/import/src/Importer.test.cpp
```diff
@@ -16,13 +16,289 @@
...
+    // +4 from entry id, +4 from entry size, +4 from execution order + offsets.size() * 4 for
+    // offsets
+    buffer.resize(old_size + 12 + offsets.size() * 4);
+    uint32_t *number_of_entries_ptr = reinterpret_cast<uint32_t *>(buffer.data());
+    *number_of_entries_ptr += 1;
+
+    uint32_t *entry_data_ptr = reinterpret_cast<uint32_t *>(buffer.data() + old_size);
+
+    entry_data_ptr[0] = *number_of_entries_ptr - 1; // entry id
+    entry_data_ptr[1] = 1 + offsets.size();         // entry size
```
m-bronnikov, line 145: Q. Could you please explain why +1?

compiler/luci/import/src/Importer.test.cpp
```diff
@@ -16,13 +16,289 @@
...
+        break;
+      }
+      case luci::CircleOpcode::CIRCLEOUTPUT:
+      {
+        case luci::CircleOpcode::RELU:
+        {
+          ASSERT_FALSE(luci::has_execution_plan(node));
+          break;
+        }
+      }
```
m-bronnikov, line 275: (optional) How about?

```suggestion
      case luci::CircleOpcode::RELU:
      {
        ASSERT_FALSE(luci::has_execution_plan(node));
        break;
      }
```

compiler/luci/import/src/Importer.test.cpp
```diff
@@ -16,13 +16,289 @@
...
+    return idx;
+  }
+
+  uint32_t add_plan_metadata(uint32_t buffer_id)
+  {
+    uint32_t idx = model->metadata.size();
+    model->metadata.emplace_back(new circle::MetadataT);
+    model->metadata[idx]->name = "ONE_execution_plan_table";
+    model->metadata[idx]->buffer = buffer_id;
+    model->buffers[buffer_id]->data.resize(4);
```
m-bronnikov, line 123: (optional) How about set counter as 0 explicitly to increase code readability? 

```suggestion
    model->buffers[buffer_id]->data.resize(4);
    uint32_t &counter = *reinterpret_cast<uint32_t *>(model->buffers[buffer_id]->data.data());
    counter = 0;
```

compiler/luci/import/src/Importer.test.cpp
m-bronnikov: I think this braces some redundant.

compiler/luci/import/src/Importer.test.cpp
```diff
@@ -16,13 +16,289 @@
...
+    uint32_t idx = graph->tensors.size();
+    graph->tensors.emplace_back(new circle::TensorT);
+    graph->tensors[idx]->shape = shape;
+    graph->tensors[idx]->type = circle::TensorType_FLOAT32;
+    graph->tensors[idx]->buffer = buffer_id;
+    graph->tensors[idx]->name = std::to_string(idx);
+    graph->tensors[idx]->quantization.reset(nullptr);
+    graph->tensors[idx]->is_variable = false;
+    graph->tensors[idx]->sparsity.reset(nullptr);
+    (void)graph->tensors[idx]->shape_signature;
```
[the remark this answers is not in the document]
m-bronnikov: IIUC this lines added to increase review experience and denotes attributes that exist in added structure but unused for this test.

compiler/luci/import/src/Importer.test.cpp
m-bronnikov: I think from `from` means that something like:

```
[  ... | entry id | +4 butes from entry id | ... ]
```

Please correct if I'm wrong.

compiler/luci/import/src/Importer.test.cpp
binarman: yep, looks like I messed up with my keyboard and editor =)

compiler/luci/import/src/Importer.test.cpp
binarman: This value contains size of `uint32_t` values in entry:
- 1 value is `execution_order`
- rest of the values are `offsets`

+1 is for `execution_order`

compiler/luci/import/src/Importer.test.cpp
binarman: When I wrote this I had in mind something like this:
"4 bytes come from id, 4 bytes come from entry size, etc."

I think `for` is better, and maybe it is a good idea to remove "+"

compiler/luci/import/src/Importer.test.cpp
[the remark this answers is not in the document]
binarman: > unused for this test.

Yes

> denotes attributes that exist

Yes =)

> increase review experience

Hmm, I did not think about this =)

It is for theoretical future tests(if we need them).
I found it annoying to look into schema all the time, so decided to add all fields at once.

compiler/luci/import/src/Importer.test.cpp
```diff
@@ -16,13 +16,289 @@
...
+{
+
+struct BasicCircleModel
+{
+  std::unique_ptr<circle::ModelT> model;
+
+  BasicCircleModel()
+  {
+    model = std::make_unique<circle::ModelT>();
+    model->buffers.emplace_back(new circle::BufferT);
```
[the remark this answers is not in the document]
binarman: I investigated this a little and found interesting thing.

`model->buffers.emplace_back(new circle::BufferT);` is generally unsafe, it can leak memory:
if relocation for larger buffer fails, exception is thrown. In this case memory allocated by `new` is lost.

Therefore your fears are reasonable!

It is unlikely case, but still interesting  =)

compiler/luci/import/CMakeLists.txt
```diff
@@ -33,3 +33,9 @@ GTest_AddTest(luci_import_test ${TESTS})
 target_include_directories(luci_import_test PRIVATE src)
 target_link_libraries(luci_import_test luci_import)
 target_link_libraries(luci_import_test oops)
+target_link_libraries(luci_import_test luci_plan)
+target_link_libraries(luci_import_test luci_lang)
+target_link_libraries(luci_import_test mio_circle)
+
+nnas_find_package(GTest REQUIRED)
```
seanshpark, line 40: maybe redundant line?

binarman: Yes, thank you!
fixed

seanshpark (approved):
LGTM

m-bronnikov (approved):
LGTM

BalyshevArtem (approved):
LGTM

PR #7914, opened 2021-11-02
DRAFT check Conv2D without bias in luci

seanshpark:
on-going draft to check Conv2D without bias in luci.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #7913, opened 2021-11-02
[luci] Conv2D without bias


seanshpark:
Current tflite test materials using `tflchef` does not have `Conv2D` with no bias; empty bias.

There seems to some problem we need to investigate.

seanshpark:
This began from #7911

seanshpark:
fixed

PR #7923, opened 2021-11-03
[res] Introduce Conv2D_005 recipe

seanshpark:
This will introduce Conv2D_005 recipe that having no bias for Conv2D.


seanshpark:
For #7913, from draft #7914

chunseoklee (approved):

mhs4670go (approved):
LGTM

PR #7879, opened 2021-10-22
[WIP] [DRAFT] [one-cmds] Add pytorch frontend

binarman:
This is attempt to support pytorch format in one-cmds using various converters.

ONE-DCO-1.0-Signed-off-by: Alexander Efimov <a.efimov@samsung.com>

PR #7924, opened 2021-11-03
[one-cmds] Auto detect drivers

binarman:
This PR introduces mechanism for auto detecting available drivers.

This is "auto detect" feature I was talking in https://github.com/Samsung/ONE/pull/7879#issuecomment-951845839

### DIFF
--- a/compiler/one-cmds/one-build
+++ b/compiler/one-cmds/one-build
@@ -21,6 +21,7 @@
 
 import argparse
 import configparser
+import importlib
 import os
 import subprocess
 import sys
@@ -75,19 +76,6 @@ def _verify_arg(parser, args):
             parser.error('Invalid optimization option')
 
 
-def _get_driver_name(driver_name):
-    return {
-        'one-import-bcq': 'one-import-bcq',
-        'one-import-tf': 'one-import-tf',
-        'one-import-tflite': 'one-import-tflite',
-        'one-import-onnx': 'one-import-onnx',
-        'one-optimize': 'one-optimize',
-        'one-quantize': 'one-quantize',
-        'one-pack': 'one-pack',
-        'one-codegen': 'one-codegen'
-    }[driver_name]
-
-
 def _parse_cfg(args):
     config = configparser.ConfigParser()
     config.optionxform = str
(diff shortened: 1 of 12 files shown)


binarman:
The Idea is following:
- Each driver provides function `get_driver_spec`, which return specification of this driver( Currently this is name of related config section and type of the driver).
- `one-build` and `onecc` tools tries to import every file in current directory and call `get_driver_spec` function. If call succeeds, file is considered a driver and handled according it's specification.

seanshpark:
@jyoungyun , I've added you as a reviewer but if you feel uncomfortable, you can remove.

binarman:
I run tests, but they are slow =(
Currently only 12 finished

I think I'll proceed tomorrow

binarman:
Removed NO MERGE, because all tests finished with success =)

mhs4670go:
@binarman I should have asked this more earlier. Why is this change necessary? I mean the relationship between "auto detecting" and "introducing experimental feature". Isn't it enough to just add `one-import-pytorch` to driver list and suppress related help message?

binarman:
@mhs4670go 
> relationship between "auto detecting" and "introducing experimental feature"

My idea is following:
If there are no explicit references to drivers, we can deliver arbitrary subset of drivers to customers without changing code of one-cmd tools or writing additional logic for suppressing messages, etc.

> Why is this change necessary?

This change is not necessary, indeed =)
Same behavior can be achieved with "suppress related help message" and related stuff as well.

Reasons I choose this approach over checking drivers are:
- (this is kind of my personal preferences :slightly_smiling_face: ) I don't like when tools references some "excluded" features 
- I think this approach produces simpler code without complex error checking logic, like "if this is particular driver, do this"

mhs4670go:
@binarman Thank you for your explanation:)

> If there are no explicit references to drivers, we can deliver arbitrary subset of drivers to customers without changing code of one-cmd tools or writing additional logic for suppressing messages, etc.

Actually, when I designed this part, I intentionally forced drivers to need explicit references. But either way looks good to me. Let's merge this and see how things are going to be:)

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

binarman:
@mhs4670go 
> Actually, when I designed this part, I intentionally forced drivers to need explicit references. But either way looks good to me. Let's merge this and see how things are going to be:)

Thank you!

This is not critical part, since I need it for experimental stuff
Feel free to revert it if it breaks something :slightly_smiling_face:

compiler/one-cmds/one-build
```diff
@@ -154,25 +161,26 @@ def main():
...
-    for d in drivers:
-        if _is_available_driver(config, d):
-            section_to_run.append(d)
+    # get drivers to run
+    driver_to_run = []
+    for driver_path in drivers:
+        driver_spec = drivers[driver_path]
+        driver_name = driver_spec[0]
+        if _is_available_driver(config, driver_name):
+            driver_to_run.append(driver_path)
```
m-bronnikov, line 178: @binarman As @mhs4670go noticed in https://github.com/Samsung/ONE/issues/7962#issuecomment-965983975, drivers can be launched in wrong order. In case of *one-build_001.test* `one-optimize` is first element in `driver_to_run` and `one-import-tf` is second.

I think sorting of `driver_to_run` in right order could fix this PR.

binarman: Now I noticed this thing too, now I understand why we needed explicit driver list =)

Issue #7928, opened 2021-11-04
[circle-partitioner] Can't partition Unpack-Tanh model


chunseoklee:
### Issue

circle-partitioner can't partition model with Unpack-Tanh sequence

### Reproduce Step 

recipe :  #7927 

partition file 

```
[partition]
backends=cpu,npu
default=npu
comply=opcode

[OPCODE]
UNPACK=cpu
```

Note that it doe snot happen if no partition occur. Also it succeed with https://github.com/Samsung/ONE/blob/master/res/TensorFlowLiteRecipes/Unpack_001/test.recipe (single unpack model)

seanshpark:
Existing `Net_UnpackAdd_001` seems to fail. I'll work on with this model.

seanshpark:
Multi-output nodes like `Unpack` was not handled in current partitioner.
That is, the feature was not developed yet. I need some time to think how to implement this...

seanshpark:
@chunseoklee , are there any other nodes that have multiple outputs like UNPACK?

seanshpark:
@chunseoklee , could you please check with #7929 ?

chunseoklee:
(Maybe)split operation ?

seanshpark:
> (Maybe)split operation ?

OK, I'll work on this Op too

seanshpark:
SPLIT added to #7929

chunseoklee:
> @chunseoklee , could you please check with #7929 ?

This patch seems to resolve the multi-output issue of Unpack.(But, during partition, new error occurs. I will post new issue after analysis)

seanshpark:
> This patch seems to resolve the multi-output issue of Unpac

If the draft code for Unpack and others work, do you think it is OK to post PRs for them?

chunseoklee:
> > This patch seems to resolve the multi-output issue of Unpac
> 
> If the draft code for Unpack and others work, do you think it is OK to post PRs for them?

Yes. 

> during partition, new error occurs. I will post new issue after analysis

Still WIP(hard to figure out).

chunseoklee:
closed by resolved

PR #7929, opened 2021-11-04
DRAFT partition unpack op

seanshpark:
on-going draft to enable multi-output op like unpack.


seanshpark:
For #7928

seanshpark:
done

PR #7930, opened 2021-11-05
[tflchef] Add SignatureDef in schema

seanshpark:
This will add SignatureDef and TensorMap in schema to tflite with
SignatureDef.

### DIFF
--- a/compiler/tflchef/proto/tflchef.proto
+++ b/compiler/tflchef/proto/tflchef.proto
@@ -647,6 +647,22 @@ message Operation {
   // use the number not listed in the above reserve list
 }
 
+message TensorMap {
+  optional string name = 4;
+  // use tensor to find tensor index from name or use tensor_index with index.
+  // either one should exist.
+  optional string tensor = 5;
+  optional uint32 tensor_index = 6;
+}
+
+message SignatureDef {
+  repeated TensorMap inputs = 4;
+  repeated TensorMap outputs = 5;
+  optional string method_name = 6;
+  optional string key = 10;
+  optional uint32 subgraph_index = 12;
+}
+
 // For additional subgraphs
 message Graph {
   repeated Operand operand = 1;
@@ -664,4 +680,5 @@ message ModelRecipe {
   optional string name = 5;
   optional uint32 version = 6 [default = 1];
   repeated Graph graph = 7;
+  repeated SignatureDef signature_def = 8;
 }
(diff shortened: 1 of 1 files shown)


seanshpark:
From #7593

compiler/tflchef/proto/tflchef.proto
```diff
@@ -647,6 +647,22 @@ message Operation {
   // use the number not listed in the above reserve list
 }
 
+message TensorMap {
+  optional string name = 4;
+  // use tensor to find tensor index from name or use tensor_index with index.
+  // either one should exist.
+  optional string tensor = 5;
+  optional uint32 tensor_index = 6;
```
mhs4670go, line 655: Could you give me some example about this? I don't know the difference between `tensor` and `tensor_index`. Because `TensorMap` is a map from an alias name of tensor to tensor index in the graph.
```flatbuffers
// 2.6.0 tflite schema
table TensorMap {
  // Represents the alias to use for this tensor.
  name:string;

  // The actual tensor index in the primary graph, that 'name' corresponds to.
  tensor_index:uint;
}

Comment (seanshpark) [reply compiler/tflchef/proto/tflchef.proto:655]: Background: while authoring `recipe` file,
- there is no sequence(index) number of the tensor (operand)
- we can count the from the beginning, but if the file gets huge, counting will become annoying thing todo

`tensor` is a string to used to give the operand name for the output. this is what we usually use in recipe file
- https://github.com/Samsung/ONE/pull/7593/files#diff-c9567d08ead481592c2070896ff9aaac1c71a81aed65839ba0f28172ef1c57c7
`tensor_index` is when we can easly get the index number.
- https://github.com/Samsung/ONE/pull/7593/files#diff-55f8e5d0b4e5c234960ebc56bdedef895d84cbf7fcfb26d3d8de6e7329612cad 

note that output name may not be same as operand name.

Comment (mhs4670go) [reply compiler/tflchef/proto/tflchef.proto:655]: @seanshpark Thank you for you explanation. Draft's example helped me to understand this. I think the reason why I didn't get it is because `name` of the above comments and `TensorMap`'s `name` are not the same. IIUC, maybe another comment could be added like below.
```
// `tensor` refers to the name of `Operand` whose index is `tensor_index`.
optional string tensor = 5;
```

seanshpark: Thanks, I'll revise comment to reflect your thoughts

mhs4670go (approved):
Thank you:) LGTM

PR #7931, opened 2021-11-05
[tflchef] Revise cook_graph to return symbol_table

seanshpark:
This will revise cook_graph method to return symbol_table that will
be used to find tensor index from name for SignatureDef.


seanshpark:
From #7593

mhs4670go (approved):
LGTM

PR #7933, opened 2021-11-07
[tflchef] Cook SignatureDef from recipe

seanshpark:
This will cook tflite SignatureDef from recipe.


seanshpark:
From #7593

mhs4670go (approved):
LGTM

PR #7938, opened 2021-11-08
[luci/import] Add and rename CircleReader members

m-bronnikov:
This commit adds `_native_subgraph` and renames `_model_ptr` to `_native_model` in order to use them similar to `_model` and `_current_subgraph` members for direct API support.

### DIFF
--- a/compiler/luci/import/src/CircleReader.cpp
+++ b/compiler/luci/import/src/CircleReader.cpp
@@ -326,7 +326,7 @@ bool CircleReader::parse(const circle::Model *model)
   _model.reset(model->UnPack());
 
   // for direct pointer access
-  _model_ptr = model;
+  _native_model = model;
 
   return true;
 }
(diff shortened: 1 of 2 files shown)


compiler/luci/import/src/CircleReader.cpp
```diff
@@ -342,10 +342,13 @@ bool CircleReader::select_subgraph(uint32_t sgindex)
   _current_subgraph = _model->subgraphs[sgindex].get();
 
   // for direct pointer access
-  auto subgraphs = _model_ptr->subgraphs();
-  const circle::SubGraph *subgraph = (*subgraphs)[sgindex];
+  auto subgraphs = _native_model->subgraphs();
+  assert(subgraphs != nullptr);
```
m-bronnikov, line 346: Also, this PR adds some `assert`s since `subgraphs` can be `nullptr`.

seanshpark (approved):
LGTM

PR #7948, opened 2021-11-09
[luci] Revise eval_verifier

seanshpark:
This will revise luci_eval_verifier python script to use variables.
- this is to prepare next revision to support SignatureDef


seanshpark:
From #7593

mhs4670go (approved):
LGTM

PR #7956, opened 2021-11-09
[luci/import] Use direct-access tensors

m-bronnikov:
This commit replaces tensors() to native_tensors() in Importer, ValidateHelpers and GraphBuilders base classes.

------------------

For: #7886
Draft: #7901

### DIFF
--- a/compiler/luci/import/src/Importer.cpp
+++ b/compiler/luci/import/src/Importer.cpp
@@ -51,10 +51,9 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
   luci::GraphBuilderContext gb_context(graph, &reader, nodefinder.get(), tensoroutputs.get());
 
   const auto operators = reader.native_operators();
-  const auto &tensors = reader.tensors();
-  auto tensors_ptr = reader.tensors_ptr();
-  assert(tensors_ptr != nullptr);
+  const auto tensors = reader.native_tensors();
   auto circle_metadata = std::make_unique<luci::CircleImportMetadata>(reader);
+  assert(!tensors.null());
 
   // build a cache to identify if a tensor is output of an operator
   // if this is set, we should not create a CircleConst for this tensor
@@ -78,10 +77,11 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
   {
     auto input_node = graph->nodes()->create<luci::CircleInput>();
     assert(input_node != nullptr);
-    const circle::TensorT &tensor = *tensors[input];
+    const auto tensor = tensors[input];
+    assert(tensor != nullptr);
 
     luci::copy_tensor_attributes(tensor, input_node);
-    if (tensors_ptr->Get(input)->shape() == nullptr)
+    if (tensor->shape() == nullptr)
       input_node->shape_status(luci::ShapeStatus::NOSHAPE);
     else
       input_node->shape_status(luci::ShapeStatus::VALID);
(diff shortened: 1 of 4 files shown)


m-bronnikov:
Since `tensors()` --> `native_tensors()` replacement is huge amount of changes I would like to split it to 4 (or more if needed) PR's. This is a first PR of this series. 

But I totally failed to come up with informative headers for these requests. So, I would like to hear any suggestion for titles. =)

compiler/luci/import/src/Importer.cpp
```diff
@@ -51,10 +51,9 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
   luci::GraphBuilderContext gb_context(graph, &reader, nodefinder.get(), tensoroutputs.get());
 
   const auto operators = reader.native_operators();
-  const auto &tensors = reader.tensors();
-  auto tensors_ptr = reader.tensors_ptr();
-  assert(tensors_ptr != nullptr);
+  const auto tensors = reader.native_tensors();
   auto circle_metadata = std::make_unique<luci::CircleImportMetadata>(reader);
+  assert(!tensors.null());
```
seanshpark, line 56: maybe put this after `const auto tensors = reader.native_tensors();` line

compiler/luci/import/src/Importer.cpp
```diff
@@ -102,16 +102,18 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
...
-           tensor.shape_signature.size() == tensor.shape.size());
+    const auto tensor_shape_signature = luci::wrap(tensor->shape_signature());
+    const auto tensor_shape = luci::wrap(tensor->shape());
+    assert(tensor_shape_signature.size() == 0 ||
+           tensor_shape_signature.size() == tensor_shape.size());
 
     // Shape of GraphInput
     auto input_shape = std::make_unique<loco::TensorShape>();
-    const std::vector<int32_t> &input_dims = tensor.shape; // in NHWC
+    const auto input_dims = tensor_shape; // in NHWC
```
seanshpark, line 112: will this make a copy? if so, how about using `const auto &input_dims` ?

compiler/luci/import/src/Importer.cpp
```diff
@@ -203,24 +206,26 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
...
-           tensor.shape_signature.size() == tensor.shape.size());
+    const auto tensor_shape_signature = luci::wrap(tensor->shape_signature());
+    const auto tensor_shape = luci::wrap(tensor->shape());
+    assert(tensor_shape_signature.size() == 0 ||
+           tensor_shape_signature.size() == tensor_shape.size());
 
     // Shape of Output
     auto output_shape = std::make_unique<loco::TensorShape>();
-    const std::vector<int32_t> &output_dims = tensor.shape; // in NHWC
+    const auto &output_dims = tensor_shape; // in NHWC
```
seanshpark, line 216: (here you used `&`)

seanshpark (commented):
.

compiler/luci/import/src/Importer.cpp
m-bronnikov: This line makes a copy of `VectorWrapper` struct, but I would like to use `&` here. Thank you!

```suggestion
    const auto &input_dims = tensor_shape; // in NHWC
```

compiler/luci/import/src/Importer.cpp
m-bronnikov: Done.

seanshpark (approved):
LGTM thank you!

PR #7957, opened 2021-11-09
[luci/import] Use direct-access tensors from Abs to If

m-bronnikov:
This commit replaces tensors() to native_tensors() in all builders from CircleAbs to CircleIf.

-------------------------

For: #7886
Draft: #7901

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleConst.cpp
+++ b/compiler/luci/import/src/Nodes/CircleConst.cpp
@@ -30,10 +30,10 @@
 namespace
 {
 
-std::ostream &operator<<(std::ostream &os, const std::vector<int32_t> &vect)
+std::ostream &operator<<(std::ostream &os, const luci::VectorWrapper<int32_t> &vect)
 {
   uint32_t seq = 0;
-  for (auto &v : vect)
+  for (const auto &v : vect)
   {
     if (seq)
       os << ", ";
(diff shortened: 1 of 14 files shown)


m-bronnikov:
Since `tensors()` --> `native_tensors()` replacement is huge amount of changes I would like to split it to 4 (or more if needed) PR's. This is a first PR of this series. 

But I totally failed to come up with informative headers for these PRs. So, I would like to hear any suggestion for titles. =)

_Originally posted by @m-bronnikov in https://github.com/Samsung/ONE/pull/7956#issuecomment-964573935_

-------------

This is a second PR of this series.

seanshpark:
> So, I would like to hear any suggestion for titles. =)

Title is OK, I also couldn't get better title...

compiler/luci/import/src/Nodes/CircleConst.cpp
```diff
@@ -30,10 +30,10 @@
 namespace
 {
 
-std::ostream &operator<<(std::ostream &os, const std::vector<int32_t> &vect)
+std::ostream &operator<<(std::ostream &os, const luci::VectorWrapper<int32_t> &vect)
 {
   uint32_t seq = 0;
-  for (auto &v : vect)
+  for (const auto &v : vect)
```
seanshpark, line 36: what happens when we don't add `const` ?
I'd like to know if this is necessary for using direct access.

seanshpark:
> I would like to split it to 4 (or more if needed)

From reviwer point of view, more than 50 lines of changes will degrade concentration on reviewing... for my case -_-;

compiler/luci/import/src/Nodes/CircleConst.cpp
m-bronnikov: Unfortunately, compilation error occurs:

```cpp
/ONE/compiler/luci/import/src/Nodes/CircleConst.cpp: In function ‘std::ostream& {anonymous}::operator<<(std::ostream&, const luci::VectorWrapper<int>&)’:
/ONE/compiler/luci/import/src/Nodes/CircleConst.cpp:36:18: error: cannot bind non-const lvalue reference of type ‘int&’ to an rvalue of type ‘int’
   for (auto &v : vect)
```

m-bronnikov:
> > I would like to split it to 4 (or more if needed)
> 
> From reviwer point of view, more than 50 lines of changes will degrade concentration on reviewing... for my case -_-;

Let me please push PR's with `<= 50` lines of code starting with the next PR. Sorry for the inconvenience.

seanshpark (approved):
LGTM thank you!

PR #7958, opened 2021-11-10
[luci/partition] Revise IsVirtualNode for multiple outputs

seanshpark:
This will revise IsVirtualNode class to check multiple output node types.


seanshpark:
For #7928, from draft #7929

seanshpark:
maybe have to rebase this after some other codes land.

seanshpark:
moved into #7961

PR #7959, opened 2021-11-10
[luci/part] Introduce append for produce_pgroups

seanshpark:
This will introduce append method extracted fom produce_pgroups for reuse.

### DIFF
--- a/compiler/luci/partition/src/PartitionPGroups.cpp
+++ b/compiler/luci/partition/src/PartitionPGroups.cpp
@@ -60,6 +60,42 @@ bool check_allocate_partition(const luci::CircleNode *node)
 
 } // namespace
 
+namespace
+{
+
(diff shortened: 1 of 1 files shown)


seanshpark:
For #7928, from draft #7929

compiler/luci/partition/src/PartitionPGroups.cpp
```diff
@@ -120,6 +156,8 @@ std::unique_ptr<luci::PGroups> produce_pgroups(const luci::Module *source,
       INFO(l) << "Op: " << node->name() << ": " << opcodename << ", " << node << ", " << group
               << std::endl;
 
+      append(node, pgroups.get(), group, idx);
+#if 0
```
chunseoklee, line 160: Why dont you remove this in the PR ?

seanshpark: this is to reduce changes :)

chunseoklee (approved):

PR #7961, opened 2021-11-10
[luci/partition] Revise pgroup for multioutput virt ndoes

seanshpark:
This will revise creating pgroup for multioutput virtual nodes to follow
group of the owner node.

### DIFF
--- a/compiler/luci/partition/src/PartitionPGroups.cpp
+++ b/compiler/luci/partition/src/PartitionPGroups.cpp
@@ -35,6 +35,17 @@ class IsVirtualNode final : public luci::CircleNodeVisitor<bool>
 public:
   bool visit(const luci::CircleInput *) final { return true; }
   bool visit(const luci::CircleOutput *) final { return true; }
+  // For multiple outputs
+  bool visit(const luci::CircleCustomOut *) final { return true; }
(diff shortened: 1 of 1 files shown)


seanshpark:
For #7928, from draft #7929

seanshpark:
Needs #7958 change too...

compiler/luci/partition/src/PartitionPGroups.cpp
```diff
@@ -195,6 +255,22 @@ std::unique_ptr<luci::PGroups> produce_pgroups(const luci::Module *source,
     }
   }
 
+  // handle for virtual nodes like multiple outputs
+  // these nodes should follow group of the input
```
mhs4670go, line 259: I'm not sure about what "follow" means. "follow group of the input" means the nodes are appended to the group of input?

seanshpark: this is intended to explain that thiese nodes should be grouped to the same group of the input (owner).
maybe use `these nodes should be assigned to same group of the input` ?

chunseoklee (approved):

mhs4670go (approved):
LGTM

PR #7966, opened 2021-11-10
[res] Add Part_Split_Add_000 recipe

seanshpark:
This will add Part_Split_Add_000 tflite recipe for partitioning of
multiple output nodes.


seanshpark:
For #7928, from draft #7929

mhs4670go (approved):
LGTM

PR #7969, opened 2021-11-10
[luci/partition] Remove unused in produce_pgroups

seanshpark:
This will remove unused codes in produce_pgroups method.


seanshpark:
For #7928, from draft #7929

chunseoklee (approved):

PR #7971, opened 2021-11-10
[circle-partitioner] Value tests of multi-output nodes

seanshpark:
This will enable circle-partitioner value testing of multi-output node types.


seanshpark:
For #7928, from draft #7929

seanshpark:
maybe should rebase after #7948 lands

chunseoklee (approved):

PR #7973, opened 2021-11-10
[luci] Support SignatureDef in eval verifier

seanshpark:
This will revise luci_eval_verifier to read SignatureDef and use
output tensor order according to the SignatureDef's outputs value.


seanshpark:
From #7593

mhs4670go (approved):
LGTM

PR #7937, opened 2021-11-08
[luci/import] Add VectorWrapper helper struct

m-bronnikov:
This commit adds VectorWrapper struct which helps to use `flatbuffers::Vector<>` from direct-access API.

ONE-DCO-1.0-Signed-off-by: Maksim Bronnikov max120199@gmail.com

---------------------

For: #7886
Draft: #7901

PR #7974, opened 2021-11-10
[hermes] Introduce severity based colored console reporter

SlavikMIPT:
This commit introduces severity based colored console reporter.

### DIFF
--- a/compiler/hermes-std/src/ConsoleReporter.cpp
+++ b/compiler/hermes-std/src/ConsoleReporter.cpp
@@ -17,15 +17,67 @@
 #include "hermes/ConsoleReporter.h"
 
 #include <iostream>
+#include <cstdlib>
+#include <string>
 
 namespace hermes
 {
 
+static constexpr const char *kTermColorRedTextCode = "\033[0;31m";
+static constexpr const char *kTermColorGreenTextCode = "\033[0;32m";
+static constexpr const char *kTermColorOrangeTextCode = "\033[0;33m";
+static constexpr const char *kTermColorBlueTextCode = "\033[0;34m";
+static constexpr const char *kTermColorMagentaTextCode = "\033[0;35m";
+static constexpr const char *kTermColorCyanTextCode = "\033[0;36m";
+static constexpr const char *kTermColorWhiteTextCode = "\033[0;37m";
+
+static constexpr const char *kTermBoldTextCode = "\033[1m";
+static constexpr const char *kTermUnderlineTextCode = "\033[4m";
+static constexpr const char *kTermInverseTextCode = "\033[7m";
+static constexpr const char *kTermBoldOffTextCode = "\033[21m";
+static constexpr const char *kTermUnderlineOffTextCode = "\033[24m";
+static constexpr const char *kTermInverseOffTextCode = "\033[27m";
+
+static constexpr const char *kTermColorResetAllCode = "\033[0m";
+
 void ConsoleReporter::notify(const hermes::Message *m)
 {
+  const char *env_color_p = std::getenv("ONE_HERMES_COLOR");
+  if (env_color_p)
+  {
+    auto env_color_str = std::string(env_color_p);
+    if ((env_color_str == "1") or (env_color_str == "ON"))
+      _is_colored = true;
+  }
+
+  if (_is_colored)
+  {
+    switch (m->get_severity())
+    {
+      case FATAL:
+        std::cout << kTermColorRedTextCode << kTermBoldTextCode << kTermUnderlineTextCode;
+        break;
+      case ERROR:
+        std::cout << kTermColorRedTextCode;
+        break;
+      case WARN:
+        std::cout << kTermColorOrangeTextCode;
+        break;
+      case INFO:
+        std::cout << kTermColorGreenTextCode;
+        break;
+      case VERBOSE:
+        std::cout << kTermColorResetAllCode;
+        break;
+    };
+  }
   for (uint32_t n = 0; n < m->text()->lines(); ++n)
   {
     std::cout << m->text()->line(n) << std::endl;
+  }
+  if (_is_colored)
+  {
+    std::cout << kTermColorResetAllCode;
   }
 }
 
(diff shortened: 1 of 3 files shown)


SlavikMIPT:
image

compiler/hermes/include/hermes/core/Message.h
```diff
@@ -21,6 +21,7 @@
 #include <sstream>
 #include <string>
 #include <vector>
+#include "Severity.h"
```
binarman, line 24: I think this is considered as a separate component ("hermes" vs "hermes-std" ), could you move this in separate PR?

binarman:
@SlavikMIPT 
This PR is based on broken master, rebase should fix tests

compiler/hermes-std/src/ConsoleReporter.cpp
```diff
@@ -20,13 +20,48 @@
 
 namespace hermes
 {
+static constexpr const char *kTermColorRedTextCode = "\033[0;31m";
```
seanshpark, line 23: please put an empty line before this line

compiler/hermes-std/src/ConsoleReporter.test.cpp
```diff
@@ -43,3 +43,83 @@ TEST(ConsoleReporterTest, notify)
 
   ASSERT_NO_THROW(r.notify(&m));
 }
+
+TEST(ConsoleReporterTest, notify_fatal)
```
seanshpark, line 47: Please add same amout of negative tests, name that end with `_NEG`.
Refer https://github.com/Samsung/ONE/pull/7937#issuecomment-962942396 why.

compiler/hermes/include/hermes/core/Message.h
seanshpark: Please put this line above `#include <memory>` with an empty line separation.

seanshpark (commented):
.

compiler/hermes-std/src/ConsoleReporter.cpp
```diff
@@ -20,13 +20,48 @@
...
+static constexpr const char *kTermInverseOffTextCode = "\033[27m";
+
+static constexpr const char *kTermColorResetAllCode = "\033[0m";
 
 void ConsoleReporter::notify(const hermes::Message *m)
 {
+  switch (m->get_severity())
+  {
+    case FATAL:
+      std::cout << kTermColorRedTextCode << kTermBoldTextCode << kTermUnderlineTextCode;
```
seanshpark, line 45: If these eacape character are to be show, these may affect tools that may use log captures.
Pleae provide an option to turn this on, where default is off.
- a method, that can be set from the caller of hermes
- environment variable, such as `ONE_HERMES_COLOR=1`
- and maybe more...

compiler/hermes-std/src/ConsoleReporter.cpp
seanshpark, line 23: Please put an empty line before this line

seanshpark, line 23: Please put an empty line before this line

seanshpark, line 23: Please put an empty line before this line

seanshpark, line 23: Please put an empty line before this line

compiler/hermes-std/src/ConsoleReporter.cpp
seanshpark, line 45: If these eacape character are to be shown, these may affect tools that may use log captures.
Pleae provide an option to turn this on, where default is off.
- a method, that can be set from the caller of hermes
- environment variable, such as ONE_HERMES_COLOR=1
- and maybe more...

compiler/hermes/include/hermes/core/Message.h
```diff
@@ -57,11 +57,17 @@ class Message final
   Message() = default;
 
 public:
-  void text(std::unique_ptr<MessageText> &&text) { _text = std::move(text); }
+  void text(std::unique_ptr<MessageText> &&text, SeverityCategory severity = VERBOSE)
```
seanshpark, line 60: Please do not use = VERBOSE. We don't use implicit parameter settings.

compiler/hermes/include/hermes/core/Message.h
SlavikMIPT: > I think this is considered as a separate component ("hermes" vs "hermes-std" ), could you move this in separate PR?

https://github.com/Samsung/ONE/pull/8002

seanshpark (approved):
LGTM thank you!

compiler/hermes-std/src/ConsoleReporter.cpp
```diff
@@ -17,16 +17,68 @@
...
+
 void ConsoleReporter::notify(const hermes::Message *m)
 {
+  const char *env_color_p = std::getenv("ONE_HERMES_COLOR");
+  if (env_color_p)
+  {
+    auto env_color_str = std::string(env_color_p);
+    if ((env_color_str == "1") or (env_color_str == "ON"))
+      _is_colored = true;
+  }
```
seanshpark, line 51: (optional) (A) this routine is simple and I think it won't take that much time but will run in everly lines of message.
you can (1) add a TODO like "optimize this for speed" or (2) revise this code to check if not checked before
and (B) when turned on, there is no check to turn it off. I'm not sure when once turned one, we need to check for off but for safetly in usage case, we may need it. for this case, we may have to skip (A) case.

Anyway we can go as-is or just add some TODOs...
this library maybe used in our TVN backend and if something goes wrong, there will be some reports.

binarman (approved):
LGTM

compiler/hermes-std/src/ConsoleReporter.cpp
binarman: I've tested if this code and found that million calls takes ~20 milliseconds on my hardware.
This should not be a big problem, at least for now =)

jj-crypto (approved):

Issue #7939, opened 2021-11-08
[one-cmds] Investigate alternative ways to import PyTorch model


binarman:
This issue is dedicated to experiments on importing PyTorch model in different formats.

PyTorch has some problems with models serialization:
- PyTorch is extremely tied to python, so it is hard to support import in other languages
- PyTorch manifest "flexibility", so there are no convenient way to store tensor shapes of types in model file
- It has very loose serialization rules, users can even invent their own ways to store models and all these still could be names as ".pt" or ".pth" model

Lest see serialization techniques described in PyTorch documentation:
- Serialization of model weights. Also referenced as "eager model" in some documents. Requires python source code for model deserialization. Has `.pt/.pth` file extension. https://pytorch.org/tutorials/beginner/basics/saveloadrun_tutorial.html
- Serialization of checkpoints. Same as basic eager method, plus learning attributes. Looks like user can name fields of serialized data as he please. Has `.pt/.pth` file extension. https://pytorch.org/tutorials/recipes/recipes/saving_and_loading_a_general_checkpoint.html
- Serialization of "entire model". Same as basic eager method, plus other attributes and fields of the model. Despite it's name requires python source. Has `.pt/.pth` file extension. https://github.com/pytorch/tutorials/blob/master/beginner_source/saving_loading_models.py#L195
- TorchScript. This technique stores topology and weights in one file. Has `.pt/.pth` file extension. https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html?highlight=torchscript
- Package. This methods packs arbitrary resources in zip file. Virtually it is just a convenient wrapper around zip, but for some reason it is presented as a way to deploy models. Is not updates for some time. Has `.pt/.pth` file extension. https://pytorch.org/docs/stable/package.html
- Model archive for Torch Serve. Another way to pack models, but it has well-defined interface, can wrap both eager (.py + weights) and TorchScript models. Has `.mar` file extension. https://github.com/pytorch/serve#store-a-model
- ONNX exporter (one-cmd already supports it) Has `.onnx` file extension

binarman:
Related PRs: #7924 #7903

binarman:
My plan on related changes:

- Add more examples in PyTorch resources to cover all required cases:
  - [ ] TBD
- Add one-cmd tests using examples from PyTorch examples
- Support PyTorch import in one-cmd tools
  - [ ] Implement import driver
  - [ ] Enable driver in  one-cmd tool (onecc one-build)

[...]

PR #7975, opened 2021-11-10
[res] PyTorch examples for spatial block rearrangement operations

binarman:
This PR adds examples:
- BatchToSpaceND
- SpaceToBatchND
- PixelShuffle (DepthToSpace)
- SpaceToDepth

Related issue:  #7939 

### DIFF
--- a/res/PyTorchExamples/examples/BatchToSpaceND/__init__.py
+++ b/res/PyTorchExamples/examples/BatchToSpaceND/__init__.py
@@ -1,0 +1,49 @@
+import torch
+import torch.nn as nn
+import numpy as np
+
+
+# model equivalent to tensorflow batch_to_space, but with channels first layout
+class net_BatchToSpaceND(nn.Module):
+    def __init__(self, block_shape, crop):
+        super().__init__()
+        self.block_shape = block_shape
+        self.crop = crop
+
+    def forward(self, input):
+        # Prepare attributes
+        input_shape = list(map(int, list(input.shape)))
+        block_shape = self.block_shape
+        crop = self.crop
+
+        # number of spatial dimensions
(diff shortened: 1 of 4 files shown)


res/PyTorchExamples/examples/BatchToSpaceND/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+        # Actual model starts here
+        unfolded_input = input.reshape(unfolded_shape)
+        permuted = torch.permute(unfolded_input, permute_dims)
+        full_output = permuted.reshape(fold_shape)
+        # crop output tensor
+        crop_output = full_output
+        for i in range(m):
+            crop_size = sum(crop[i])
+            crop_output = crop_output.narrow(i + n, crop[i][0], fold_shape[i + n] - crop_size)
+        return crop_output
```
binarman, line 40: @seanshpark 
Are examples like this appropriate for `PyTorchExamples` suite?

seanshpark: > Are examples like this appropriate for PyTorchExamples suite?

Well... can't say cause don't know much about PyTorch... :)
Purpose of files in `res` are used like
- unit test and intrgration test materials like tflite recipes
- just to share and for manual testing, that are not used anywhere as source file

> This example contains multiple operations, but looks like existing examples contains only single op per model.

It's not a strict rule but usually we added one Op for that particular example.
I don't know much about how PyTorch works, so I'm not sure this is valid or not.
Multiple Ops can exist if they are needed for that particular Op to operate.
If not we usually named with `Net_` prefix.

seanshpark (approved):
LGTM

PR #7976, opened 2021-11-11
[res] Add SignatureDef_MultiOut_001

seanshpark:
This will introduce SignatureDef_MultiOut_001 that has different output
names in SignatureDef record.


seanshpark:
From #7593

mhs4670go (approved):
LGTM

PR #7980, opened 2021-11-11
[circle-partition] Revise value test with variables

seanshpark:
This will revise part_eval_one using variables.
This is to apply tflite SignatureDef better.


mhs4670go (approved):
LGTM

seanshpark:
From #7593

PR #7981, opened 2021-11-11
[luci-pass] Value test with SignatureDef

seanshpark:
This will enable value test with SignatureDef with multiple outputs.


mhs4670go (approved):
LGTM

seanshpark:
From #7593

PR #7983, opened 2021-11-11
[circle-partition] Support SignatureDef in value test

seanshpark:
This will revise part_eval_one to read tflite SignatureDef and
use output tensor order according to the SignatureDef's outputs value.


seanshpark:
From #7593

seanshpark:
Duplicate of  #7982

PR #7984, opened 2021-11-11
[circle-partition] Value test with SignatureDef

seanshpark:
This will enable value test with SignatureDef with multiple outputs.

### DIFF
--- a/compiler/circle-part-value-test/parts/SignatureDef_MultiOut_000.part
+++ b/compiler/circle-part-value-test/parts/SignatureDef_MultiOut_000.part
@@ -1,0 +1,7 @@
+[partition]
+backends=cpu,acl_cl
+default=cpu
+comply=opcode
+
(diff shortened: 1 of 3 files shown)


compiler/circle-part-value-test/parts/SignatureDef_MultiOut_000.part
```diff
@@ -0,0 +1,7 @@
+[partition]
+backends=cpu,acl_cl
+default=cpu
+comply=opcode
+
+[OPCODE]
+MAXIMUM=acl_cl
```
seanshpark, line 7: `MAXIMUM` is just random OP and doesn't matter cause this test is to validate `part_eval_one.py` is working OK with SignatureDef

seanshpark:
From #7593

mhs4670go (approved):
LGTM

PR #7940, opened 2021-11-09
[luci] Support import of partially defined execution plan

binarman:
This PR enables import of graphs with with execution plan defined for
subset of operations.

related issue: #7831
for comment: https://github.com/Samsung/ONE/pull/7903#discussion_r740456803
    
ONE-DCO-1.0-Signed-off-by: Alexander Efimov <a.efimov@samsung.com>

PR #7986, opened 2021-11-11
[luci] Rename Import testsuite

binarman:
This PR changes suite name from TensorFlowLiteImport to CircleImport.

For https://github.com/Samsung/ONE/pull/7940#discussion_r746061209

### DIFF
--- a/compiler/luci/import/src/Importer.test.cpp
+++ b/compiler/luci/import/src/Importer.test.cpp
@@ -23,7 +23,7 @@
 #include <mio/circle/schema_generated.h>
 #include <flatbuffers/flatbuffers.h>
 
-TEST(TensorFlowLiteImport, Dummy)
+TEST(CircleImport, Dummy)
 {
   luci::Importer import;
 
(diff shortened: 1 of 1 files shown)


BalyshevArtem (approved):

seanshpark (approved):
LGTM

seanshpark:
@binarman , if you have copied (portion of) from tflite test codes, maybe we need to add TensorFlow copyright notice.

binarman:
@seanshpark 
I did not copy any of tflite code for these test, but I can check if there are some similar sources =)

binarman:
@seanshpark FYI
I have looked in TF sources.

Сonclusion: I consider our code is **not** look-alike TF sources =)

P.s.
There are some tests that create objects in C++, but they do not use any helping framework, like I did all indexes and structures are hardcoded.
 For example: https://github.com/tensorflow/tensorflow/blob/r2.7/tensorflow/lite/tools/optimize/model_utils_test.cc#L42
or https://github.com/tensorflow/tensorflow/blob/r2.7/tensorflow/lite/tools/optimize/modify_model_interface_test.cc#L126

Issue #7906, opened 2021-11-01
[luci-interpreter] Generalize temporary buffer handling in conv, depthwise_conv and avg_pool2d operations


binarman:
### What?

Support temporary buffer of different sizes for different target platforms.

### Why?

CMSIS NN requires temporary buffers for conv2d and depthwise_conv2d operations.

- luci-interpreter creates im2col for Conv2d operation, but cmsis-nn buffer is significantly smaller, os use of ordinary im2col buffer would be a waste of RAM.
- luci-interpreter does not allocate any temporary buffers for depthwise_conv2d.

### Plan

Need to:
- [ ] Implement and use PAL interface for temporary buffers allocation for Conv2d #7987
- [ ] Implement and use PAL interface for temporary buffers allocation for Depthwise_Conv2d #8044 
- [ ] Implement and use PAL interface for temporary buffers allocation for AvgPool2d #8047 
- [ ] Support different planning modes in execution planer #8048

binarman:
PRs exposed this problem: #7895 #7890

binarman:
+cc @BalyshevArtem @SlavikMIPT

seanshpark:
seems applied now

PR #7987, opened 2021-11-11
[luci-interpreter] New PAL interface for Conv2d temporary buffers

BalyshevArtem:
Implement and use PAL interface for temporary buffers allocation for Conv2d.

This is first task from issue #7906 

### DIFF
--- a/compiler/luci-interpreter/pal/cmsisnn/PALConv2d.h
+++ b/compiler/luci-interpreter/pal/cmsisnn/PALConv2d.h
@@ -28,11 +28,11 @@ static inline void Conv(const tflite::ConvParams &params, const tflite::RuntimeS
                         const float *input_data, const tflite::RuntimeShape &filter_shape,
                         const float *filter_data, const tflite::RuntimeShape &bias_shape,
                         const float *bias_data, const tflite::RuntimeShape &output_shape,
-                        float *output_data, const tflite::RuntimeShape &im2col_shape,
-                        float *im2col_data)
+                        float *output_data, const tflite::RuntimeShape &scratchpad_shape,
+                        float *scratchpad_data)
 {
-  (void)im2col_shape;
-  (void)im2col_data;
+  (void)scratchpad_shape;
+  (void)scratchpad_data;
   tflite::reference_ops::Conv(params, input_shape, input_data, filter_shape, filter_data,
                               bias_shape, bias_data, output_shape, output_data,
                               tflite::RuntimeShape(), nullptr);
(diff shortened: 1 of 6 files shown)


compiler/luci-interpreter/src/kernels/Conv2D.cpp
```diff
@@ -118,9 +118,17 @@ void Conv2D::configure()
...
+    params.padding_values.height = _padding_height;
+    params.padding_values.width = _padding_width;
+    params.stride_height = _params.stride_height;
+    params.stride_width = _params.stride_width;
+    params.dilation_height_factor = _params.dilation_height_factor;
+    params.dilation_width_factor = _params.dilation_width_factor;
+
+    Shape im2col_shape = luci_interpreter_pal::TemporaryTensorShape(
+      params, getTensorShape(input()), getTensorShape(filter()), getTensorShape(output()));
+
```
binarman, line 131: Maybe rename `im2col` to `scratchpad` and move this logic to PAL?
Rationale is following:
- We do not need this buffer for `mcu` PAL at all
- This is not `im2col` tensor in `cmsis` PAL, it uses more memory efficient approach in MCU kernel

BalyshevArtem: Done

BalyshevArtem:
@binarman PTAL)

compiler/luci-interpreter/pal/cmsisnn/PALConv2d.h
```diff
@@ -140,6 +131,64 @@ static inline void ConvPerChannel(const tflite::ConvParams &params, const int32_
   }
 }
 
+static inline void SetScratchpadTensor(luci_interpreter::Tensor *scratchpad,
```
binarman, line 134: Optional. I think Set**up**ScratchpadTensor suits this method better.

compiler/luci-interpreter/pal/cmsisnn/PALConv2d.h
```diff
@@ -42,33 +42,29 @@ static inline void Conv(const tflite::ConvParams &params, const tflite::RuntimeS
...
-  (void)im2col_data;
-
-  cmsis_nn_conv_params conv_params;
-  conv_params.dilation.h = params.dilation_height_factor;
-  conv_params.dilation.w = params.dilation_width_factor;
-
-  if (conv_params.dilation.h == 1 && conv_params.dilation.w == 1)
+  if (scratchpad_data)
   {
+    cmsis_nn_conv_params conv_params;
```
binarman, line 65: optional.
Add asserts just in case:
```suggestion
    assert(conv_params.dilation.h == 1);
    assert(conv_params.dilation.w == 1);
    cmsis_nn_conv_params conv_params;
```

binarman: Why I mentioned this:
I've looked in cmsis sources and found no usages of dilation fields, so this kernel will not work properly with dilation !=1 :slightly_frowning_face:

binarman (approved):
LGTM, see optional comments

binarman:
Another suggestion: maybe add `Conv2d` in PR name?
like 
`[luci-interpreter] New PAL interface for Conv2d temporary buffers`

binarman (approved):

binarman:
@SlavikMIPT 
Noticed that previous version of this patch was broken for cmsis PAL, but was not detected by CI tests.
Take this into account when building benchmarks

jj-crypto (approved):

m-bronnikov (approved):

PR #8012, opened 2021-11-16
[luci/part] Extract group_from_partition method

seanshpark:
This will extract group_from_partition method in produce_pgroups.


seanshpark:
From #8004

chunseoklee (approved):

jj-crypto (approved):

PR #8015, opened 2021-11-16
[one-cmds] Auto detect available import drivers

binarman:
This PR introduces auto detect of import drivers in one cmd tools.

For #7939 

### DIFF
--- a/compiler/one-cmds/one-build
+++ b/compiler/one-cmds/one-build
@@ -154,25 +154,30 @@ def main():
     config = _parse_cfg(args)
 
     # verify configuration file
-    drivers = [
-        'one-import-tf', 'one-import-tflite', 'one-import-bcq', 'one-import-onnx',
-        'one-optimize', 'one-quantize', 'one-pack', 'one-codegen'
+    import_drivers_dict = _utils._detect_one_import_drivers()
+    other_drivers = [
+        'one-optimize', 'one-quantize', 'one-pack', 'one-codegen', 'one-profile'
     ]
-    _verify_cfg(drivers, config)
+    _verify_cfg(import_drivers_dict, config)
 
     # verify optimization option file
     _verify_opt(args)
 
     # get sections to run
     section_to_run = []
-    for d in drivers:
+    for d in list(import_drivers_dict) + other_drivers:
         if _is_available_driver(config, d):
             section_to_run.append(d)
 
     # run
     dir_path = os.path.dirname(os.path.realpath(__file__))
     for section in section_to_run:
-        driver_path = os.path.join(dir_path, _get_driver_name(section))
+        if section in import_drivers_dict:
+            # we already has driver name in dict
+            driver_name = import_drivers_dict[section]
+        else:
+            driver_name = _get_driver_name(section)
+        driver_path = os.path.join(dir_path, driver_name)
         cmd = [driver_path, '--config', getattr(args, 'config'), '--section', section]
         if section == 'one-optimize' and _utils._is_valid_attr(args, 'O'):
             cmd += ['-O', getattr(args, 'O')]
--- a/compiler/one-cmds/one-build
+++ b/compiler/one-cmds/one-build
@@ -155,7 +155,7 @@ def main():
 
     # verify configuration file
     import_drivers_dict = _utils._detect_one_import_drivers()
-    other_drivers = [
+    transform_drivers = [
         'one-optimize', 'one-quantize', 'one-pack', 'one-codegen', 'one-profile'
     ]
     _verify_cfg(import_drivers_dict, config)
@@ -165,7 +165,7 @@ def main():
 
     # get sections to run
     section_to_run = []
-    for d in list(import_drivers_dict) + other_drivers:
+    for d in list(import_drivers_dict) + transform_drivers:
         if _is_available_driver(config, d):
             section_to_run.append(d)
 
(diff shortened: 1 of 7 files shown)


binarman:
Local one-cmd tests passed

binarman:
@mhs4670go
After previous fail, I decided to search only for import drivers.
It should be safe, because one-cmd tools check that only one is used.

Most important function in this PR is `_detect_one_import_drivers` from utils.

chunseoklee:
@binarman Note that @mhs4670go is out of office for 3 weeks

compiler/one-cmds/utils.py
```diff
@@ -368,3 +369,32 @@ def _get_optimization_list(get_name=False):
...
+    for module_name in os.listdir(bin_dir):
+        full_path = os.path.join(bin_dir, module_name)
+        if not os.path.isfile(full_path):
+            continue
+        if module_name.find("one-import-") != 0:
+            continue
+        module_loader = importlib.machinery.SourceFileLoader(module_name, full_path)
+        module_spec = importlib.util.spec_from_loader(module_name, module_loader)
+        module = importlib.util.module_from_spec(module_spec)
+        try:
```
mhs4670go, line 393: I'm not familiar with `importlib` library. So, I'm not sure but `try` statement is necessary? I just saw the example codes at https://docs.python.org/3/library/importlib.html#checking-if-a-module-can-be-imported and there's no that statement. 
Also, if it is necessary, why `except` statement just runs `pass`?

compiler/one-cmds/utils.py
```diff
@@ -368,3 +369,32 @@ def _get_optimization_list(get_name=False):
         opt_list = [_remove_suffix(s, '.cfg') for s in opt_list]
 
     return opt_list
+
+
+def _detect_one_import_drivers():
+    """Looks for available import drivers
```
mhs4670go, line 375: I think it'd be good if the comments let us know that it finds import drivers from "current directory".

mhs4670go:
@binarman Still in progress?? If you have a due date, it looks okay to merge and revise it later because what I pointed out is not that critical.

compiler/one-cmds/utils.py
binarman: I am not very experienced with `importlib` too, but this is the only way to import **arbitrary** named module I found :slightly_smiling_face:.

Problem with methods described in documentation is following: they works with ordinary named python files only (should end with "*.py").

One-cmds tools have names `one-import-tflite`, `one-import-onnx`, etc. and examples from documentation do not recognize these files as a python modules =(

Back to your question about try:
Everything above this line just initialize some structures, but do not access actual files.
Only `exec_module` call do the actual job and raise an exception in case chosen file is not a python module.

binarman: about `pass`

I can not guarantee if there will be no other files starting with `one-import-` (maybe even unrelated to one-cmds) in current directory (especially after installation), so I decided to simply ignore import failures...

binarman:
@mhs4670go 

> Still in progress?? If you have a due date, it looks okay to merge and revise it later because what I pointed out is not that critical.

At first I thought it is a blocker, but then found out that I can finish my targets without this particular PR,
so last week I did not pay much attention to this, sorry =_=

Now this is a "good-to-have" feature =)

compiler/one-cmds/one-build
```diff
@@ -154,25 +154,30 @@ def main():
     config = _parse_cfg(args)
 
     # verify configuration file
-    drivers = [
-        'one-import-tf', 'one-import-tflite', 'one-import-bcq', 'one-import-onnx',
-        'one-optimize', 'one-quantize', 'one-pack', 'one-codegen'
+    import_drivers_dict = _utils._detect_one_import_drivers()
+    other_drivers = [
```
seanshpark, line 158: (optioanl) maybe `transform_drivers` ?

compiler/one-cmds/onecc
```diff
@@ -170,22 +162,26 @@ def main():
     config = _parse_cfg(args)
 
     # verify configuration file
-    drivers = [
-        'one-import-tf', 'one-import-tflite', 'one-import-bcq', 'one-import-onnx',
+    import_drivers_dict = _utils._detect_one_import_drivers()
+    other_drivers = [
```
seanshpark, line 166: ditto

seanshpark:
@nnfw-bot test nncc-release

mhs4670go:
@nnfw-bot test nncc-release

compiler/one-cmds/utils.py
mhs4670go: @binarman Thank you for your explanation:) I understood `module_from_spec` method works with `*.py` as you said.

compiler/one-cmds/one-build
binarman: Does this imply that `_detect_one_import_drivers` should be `_detect_transform_drivers` ?

seanshpark: Um? I think `import_drivers_dict = _utils._detect_one_import_drivers()` is OK for import drivers. I think `transform_drivers` would be better for `other_drivers`. If you ment provide`_detect_transform_drivers` method for transform drivers, this is not what I thought.

binarman: Uh, I messed it up...
Thank you for clarification =)

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM!!

PR #8020, opened 2021-11-17
[luci/import] Remove native_ prefix from CircleReader methods

m-bronnikov:
This commit removes native_ prefix from other CircleReader's methods.

------------------

For: #7886
Draft: #7901


seanshpark (approved):
LGTM

PR #8024, opened 2021-11-18
[luci/part] Partition merge skip for same output

seanshpark:
This will revise merge PGroup case to skip if using nodes are in
differnt groups.


seanshpark:
From #8004, fix for https://github.com/Samsung/ONE/issues/6230#issuecomment-969656809

chunseoklee (approved):

PR #8025, opened 2021-11-18
[luci] Remove unnecessary check in PropagateQuantParam

jinevening:
This removes unnecessary successor check.


jinevening:
To reviewer:

The removed check was used when PropagateQuantParam works in a backward direction. Now the pass works in a fowrard direction, so we don't need the check.

seanshpark (approved):
LGTM

PR #8027, opened 2021-11-18
[luci/part] Fix is_input_same to check group value

seanshpark:
This will fix is_input_same method to check group value not the group ptr.


seanshpark:
From #8004

chunseoklee (approved):

PR #8028, opened 2021-11-18
[luci/part] Revise for OutputExclude

seanshpark:
This will revise condition for CircleOutputExclude node as this node
type is treated in preparing PGroup objects.


seanshpark:
From #8004

chunseoklee (approved):

Issue #7834, opened 2021-10-15
[luci] Revise 'LIBRARY_TYPE' to 'LUCI_LIBRARY_TYPE'


seanshpark:
How about renaming `LIBRARY_TYPE` to `LUCI_LIBRARY_TYPE` as this seems to set luci library type: shared or static.
If this is used for other modules, I think it would be better to introduce another for that too.

seanshpark:
Related #7830

binarman:
> could you please consider this change? If you don't agree, could you share your thoughts ?

As 7830 showed, this change will improve interface robustness, when integrating luci in other build systems, so I don't have any arguments against this =)

seanshpark:
done :)

PR #8031, opened 2021-11-19
[hermes-std] Add NNCC_LIBRARY_NO_PIC option to CMakeLists.txt

SlavikMIPT:
This commit adds NNCC_LIBRARY_NO_PIC option to have an opportunity to disable POSITION_INDEPENDENT_CODE property.

ONE-DCO-1.0-Signed-off-by: Vyacheslav Bazhenov slavikmipt@gmail.com


m-bronnikov (dismissed):
LGTM

binarman (commented):
LGTM, but let me call for @seanshpark, just in case

This change is needed for MCU build, to eliminate redundant indirect calls and simplify build process (need custom linking scripts like [here](https://mcuoneclipse.com/2021/06/05/position-independent-code-with-gcc-for-arm-cortex-m/) for PIC code on MCU)

@seanshpark Are you ok with this changes?

P.s. There are tons of similar changes in other components, but they all similar: https://github.com/Samsung/ONE/issues/7288#issuecomment-973911884

seanshpark:
@binarman , where is `HERMES_STD_NO_PIC` defined?

seanshpark:
> @seanshpark Are you ok with this changes?

Please share why we need this for MCUs. I need to understand the reason.

seanshpark:
Q) Do we have to define/add `---_NO_PIC` for each and every modules?

seanshpark:
@binarman , please check with our NPU backend compiler is OK with these definitions.
Previous one that was added in several weeks ago had some problem.

SlavikMIPT:
> > @seanshpark Are you ok with this changes?
> 
> Please share why we need this for MCUs. I need to understand the reason.

There are troubles with arm-none-eabi linker with PIC option enabled - the program is build, but it fails in runtime when trying to use components which have PIC option

binarman:
> where is HERMES_STD_NO_PIC defined?

For now there are no such places, but we want to use this option for separate MCU builds of interpreter only.
This flags should not affect x86 builds of ONE toolchain.

> Q) Do we have to define/add ---_NO_PIC for each and every modules?

This should be done to modules used in interpreter only.
@SlavikMIPT created PR for all such components and listed them in the https://github.com/Samsung/ONE/issues/7288#issuecomment-973911884.

> Please check with our NPU backend compiler is OK with these definitions.

Yes, good idea. I'll check to today =)

binarman:
@seanshpark 
> Please check with our NPU backend compiler is OK with these definitions.

This and other related PRs do not break backend build and tests

seanshpark:
>>  Q) Do we have to define/add ---_NO_PIC for each and every modules?
> 
> This should be done to modules used in interpreter only.

If one definition is OK then can we go with only one definition?
That is, do we have cases like some libs use PIC and some don't?

>>  where is HERMES_STD_NO_PIC defined?
> 
> For now there are no such places, but we want to use this option for separate MCU builds of interpreter only.
> This flags should not affect x86 builds of ONE toolchain.

Could you please explictly defined these, maybe in `infra/nncc/CMakeLists.txt` and add comments ?
- If someone finds this definition is used and cannot find a clue about it in the code then there will be questions.
- github blame will lead to the PRs but only this PR has discussions.

binarman:
@seanshpark 

FYI, @SlavikMIPT is on vacation, so I'll advance this task while he is absent.

> If one definition is OK then can we go with only one definition?

I think yes, one definition is OK.

>That is, do we have cases like some libs use PIC and some don't?

I can imagine only some strange case, like kernel updates over network.
I don't think we will need something like this in several years =)

binarman:
Before I rename variables, I want to clarify if this is the best solution,
previously we decided to create separate variables in similar situation: https://github.com/Samsung/ONE/pull/7759/files#r716265985

seanshpark:
> previously we decided to create separate variables in similar situation

I think the comment is about this question was

> I want to request this STATIC_LUCI be explictly defined somewhere like infra/nncc/CMakeLists.txt

Which is like https://github.com/Samsung/ONE/pull/8031#issuecomment-974914066
> @binarman , where is HERMES_STD_NO_PIC defined?

seanshpark:
> previously we decided to create separate variables in similar situation:

Yes, this thought also exist for this case too. So I asked if it's possible to use one...
Because, for static luci, there was only one definition.
For this, there are lots of them, and if we explictly defined NO_PIC, then we have to change all of them to turn it ON.
I don't think this is horrible thing but a choice of management for switching.
I prefer one definition.
@binarman , if you have another story for this, please share and I will rethink to declare individual defines.

binarman:
> if you have another story for this, please share and I will rethink to declare individual defines.

I mean #7834, just messed up with links =(

Actually I prefer to use one definition too, this variable is used in separate build and should not affect normal build for PC.

binarman:
Created PR with general option in infra/nncc #8071

binarman:
CMake variable is renamed to comply with #8071

seanshpark (approved):
LGTM

binarman (approved):

binarman:
+cc @parjong 
I am not sure if you are still interested in these components, so just for your information.

There are more similar PRs: https://github.com/Samsung/ONE/issues/7288#issuecomment-973911884

m-bronnikov (approved):
LGTM

Issue #8042, opened 2021-11-19
[luci] Remove constants duplication from circle to luci


m-bronnikov:
### What?

Let's avoid constants duplication between circle and luci IR. 

### Why?

Constant duplication is significant overheap for low-end devices, especially for MCU. 

### How?

It's possible to use constants directly from model's buffer!

-----------------

This issue is part of #7604.

m-bronnikov:
I described in draft #8043 what changes I would like to introduce. This approach leads to extension of luci IR and Importer interface, so, I would like to discuss this changes first.

Also, approach implemented in #8043 requires that model's buffer must be alive longer than imported IR. I don't know how to declare this restriction in the best way and draft not gives answer to this question =(

So, I would like to hear any suggestions for this issue and any comments about draft #8043.

-----------------------

+cc @binarman @seanshpark @BalyshevArtem

seanshpark:
There are features in _luci_ where CircleConst are duplicated when there are identical contents.
Have you check with these kind of models?

m-bronnikov:
> There are features in _luci_ where CircleConst are duplicated when there are identical contents. Have you checked with these kind of models?

No, I didn't check, I'll make it soon, thank you.

m-bronnikov:
Issue is resolved. Value test for feature: #8344. Let's close!

PR #8043, opened 2021-11-19
[WIP][DRAFT][luci] Support importing without constant copying

m-bronnikov:
This draft adds support of avoiding constants copying in model import.

----------------

For: #8042


m-bronnikov:
For reviewer: Please note that chages in `luci-eval-driver` (last commit) introduced here for demonstration of added feature usage and for coverage by some tests, which dependent to `luci-eval-driver`. I don't plan to include this changes to PR's. With this draft all test (unit + https://github.com/Samsung/ONE/blob/master/compiler/one-cmds/tests/README.txt) passed well locally.

compiler/luci/import/include/luci/Importer.h
```diff
@@ -36,7 +36,13 @@ class Importer final
   Importer();
 
 public:
-  explicit Importer(const GraphBuilderSource *source) : _source{source}
+  Importer(const GraphBuilderSource *source, bool copy_consts)
+    : _source(source), _copy_consts(copy_consts)
```
seanshpark, line 40: Please introduce new CTOR and do not remove existing CTOR.

compiler/luci/import/src/Importer.cpp
```diff
@@ -41,7 +42,7 @@ namespace
 {
 
 void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &reader,
-                   loco::Graph *graph)
+                   loco::Graph *graph, bool copy_consts)
```
seanshpark, line 45: Can you please leave existing `convert_graph` and add a new one with `copy_consts` ?

compiler/luci/import/src/Importer.cpp
```diff
@@ -125,7 +126,14 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
   // NOTE Origin is intentionally not provided for constants.
   for (uint32_t i = 0; i < tensors.size(); ++i)
   {
-    luci::CircleConst *const_node = luci::create_circleconst(&gb_context, i);
+    luci::CircleNode *const_node = nullptr;
+
+    // create copy of const data if data is 'String' or copying required by user
```
seanshpark, line 131: Please share why `String` is treated separately.

compiler/luci/import/src/Nodes/CircleRemoteConst.cpp
```diff
@@ -0,0 +1,98 @@
...
+  }
+  return os;
+}
+
+} // namespace
+
+namespace luci
+{
+
+CircleRemoteConst *create_circle_remote_const(GraphBuilderContext *context, int32_t tensor_index)
```
seanshpark, line 47: I think we need better name than `create_circle_remote_const`

compiler/luci/lang/include/luci/IR/Nodes/CircleRemoteConst.h
```diff
@@ -0,0 +1,63 @@
...
+namespace luci
+{
+
+/**
+ * @brief Class for remote access to provided tensor data
+ *
+ * @note  CircleRemoteConst has access to provided data and provides reading access to user.
+ * Node not owns given data, this is only interface to use data without memory allocation.
+ */
+class CircleRemoteConst final
```
seanshpark, line 36: Please share why to introduce `CircleRemoteConst` ?

m-bronnikov:
`check-format` fails for this PR on CI, but `./nnas format` with `clang-format-8` passes well locally. Personally, I don't know a reason, may be due to added *Note.md* file?

seanshpark:
> check-format fails for this PR on CI, but ./nnas format with clang-format-8 passes well locally. Personally, I don't know a reason, may be due to added Note.md file?

@hseok-oh , I cannot find format checker result... PTAL

seanshpark:
Please try to add additional new line to avoid this red circle
image

seanshpark:
> For reviewer: Please note that chages in luci-eval-driver (last commit) introduced here for demonstration of added feature usage and for coverage by some tests, which dependent to luci-eval-driver. I don't plan to include this changes to PR's. With this draft all test (unit + https://github.com/Samsung/ONE/blob/master/compiler/one-cmds/tests/README.txt) passed well locally.

I don't understand this. Why do you mention about the test in this PR?

compiler/luci/lang/include/luci/IR/Nodes/CircleRemoteConst.h
m-bronnikov: To reduce memory usage by constants of model we can use constants directly from model's buffer. But `CircleConst` internally uses own `std::vector` for data and makes data copying during `import`/`export`. So, `CircleRemoteConst` is analogue of `CircleConst`, which operates data pointer instead own storage. 

----------------

**FYI** This design has one negative side: `CircleRemoteConst` not guarantees that provided data is valid. Following usage scenario is prohibited:

1. Create model's buffer.
2. Make import and bind data pointer to `CircleRemoteConst` during import.
3. Free model's buffer (**this action makes data given to `CircleRemoteConst` invalid**)
4. Use imported model. `CircleRemoteConst` will provide memory access to corrupted data.

m-bronnikov:
> > For reviewer: Please note that chages in luci-eval-driver (last commit) introduced here for demonstration of added feature usage and for coverage by some tests, which dependent to luci-eval-driver. I don't plan to include this changes to PR's. With this draft all test (unit + https://github.com/Samsung/ONE/blob/master/compiler/one-cmds/tests/README.txt) passed well locally.
> 
> I don't understand this. Why do you mention about the test in this PR?

Ahh, this PR helps to avoid copying of constants by adding new parameter to `luci::Importer`. Some tests (e.g. `luci_value_test`) uses `luci-eval-driver` module, which uses `luci::Importer` for model importing. So, I changed `luci-eval-driver` in this draft to proof support done correct, but I don't plan to introduce this changes in `master`.

hseok-oh:
https://github.com/Samsung/ONE/suites/4425603744/artifacts/117089614

```diff
diff --git a/compiler/luci/import/include/luci/Import/Nodes/CircleRemoteConst.h b/compiler/luci/import/include/luci/Import/Nodes/CircleRemoteConst.h
index 7cae14a..daf0eaa 100644
--- a/compiler/luci/import/include/luci/Import/Nodes/CircleRemoteConst.h
+++ b/compiler/luci/import/include/luci/Import/Nodes/CircleRemoteConst.h
@@ -32,4 +32,4 @@ CircleRemoteConst *create_circle_remote_const(GraphBuilderContext *context, int3
 
 } // namespace luci
 
-#endif // __LUCI_IMPORT_OP_CIRCLE_REMOTE_CONST_H__
\ No newline at end of file
+#endif // __LUCI_IMPORT_OP_CIRCLE_REMOTE_CONST_H__
diff --git a/compiler/luci/import/src/Nodes/CircleRemoteConst.cpp b/compiler/luci/import/src/Nodes/CircleRemoteConst.cpp
index 3efa4da..7550787 100644
--- a/compiler/luci/import/src/Nodes/CircleRemoteConst.cpp
+++ b/compiler/luci/import/src/Nodes/CircleRemoteConst.cpp
@@ -95,4 +95,4 @@ CircleRemoteConst *create_circle_remote_const(GraphBuilderContext *context, int3
   return const_node;
 }
 
-} // namespace luci
\ No newline at end of file
+} // namespace luci
diff --git a/compiler/luci/lang/src/Nodes/CircleRemoteConst.cpp b/compiler/luci/lang/src/Nodes/CircleRemoteConst.cpp
index ae4f7d4..12ee484 100644
--- a/compiler/luci/lang/src/Nodes/CircleRemoteConst.cpp
+++ b/compiler/luci/lang/src/Nodes/CircleRemoteConst.cpp
@@ -68,4 +68,4 @@ INSTANTIATE(loco::DataType::BOOL);
 
 #undef INSTANTIATE
 
-} // namespace luci
\ No newline at end of file
+} // namespace luci
```

compiler/luci/import/include/luci/Import/Nodes/CircleRemoteConst.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+namespace luci
+{
+
+CircleRemoteConst *create_circle_remote_const(GraphBuilderContext *context, int32_t tensor_index);
+
+} // namespace luci
+
+#endif // __LUCI_IMPORT_OP_CIRCLE_REMOTE_CONST_H__
```
hseok-oh, line 35: Format fail: EOL

compiler/luci/import/src/Nodes/CircleRemoteConst.cpp
```diff
@@ -0,0 +1,98 @@
...
+          << const_dims << std::endl;
+  if (num_elements > 0)
+  {
+    const_node->bind_buffer(buffer.data(), buffer.size());
+  }
+
+  return const_node;
+}
+
+} // namespace luci
```
hseok-oh, line 98: Format fail: EOL

compiler/luci/lang/src/Nodes/CircleRemoteConst.cpp
```diff
@@ -0,0 +1,71 @@
...
+INSTANTIATE(loco::DataType::S32);
+INSTANTIATE(loco::DataType::S16);
+INSTANTIATE(loco::DataType::S8);
+INSTANTIATE(loco::DataType::FLOAT32);
+INSTANTIATE(loco::DataType::U8);
+INSTANTIATE(loco::DataType::BOOL);
+
+#undef INSTANTIATE
+
+} // namespace luci
```
hseok-oh, line 71: Format fail: EOL

compiler/luci/import/src/Importer.cpp
m-bronnikov: `create_circleconst` handles Strings separately because format of data is differ:

```cpp
template <>
void copy_data<loco::DataType::STRING>(const VectorWrapper<uint8_t> &raw_data,
                                       uint32_t num_elements, CircleConst *const_node)
{
  assert(const_node->sparsityparam() == nullptr);

  const auto *data = reinterpret_cast<const char *>(raw_data.data());
  const auto *i32d = reinterpret_cast<const int32_t *>(raw_data.data());

  // de-serialize string data
  //   int32_t count
  //   int32_t offsets[count + 1]
  //   string  values[count]
  assert(static_cast<uint32_t>(*i32d) == num_elements);
  i32d++; // skip count

  std::vector<int32_t> offsets;
  offsets.push_back(*i32d++);
  for (uint32_t i = 0; i < num_elements; ++i)
  {
    offsets.push_back(*i32d++);
  }
  assert(offsets.size() == num_elements + 1);

  const_node->size<loco::DataType::STRING>(num_elements);
  for (uint32_t i = 0; i < num_elements; ++i)
  {
    int32_t start = offsets[i];
    int32_t next = offsets[i + 1];

    std::string value(data + start, next - start);
    const_node->at<loco::DataType::STRING>(i) = value;
  }
}
```

So, I think better to make copy for Strings.

compiler/luci/lang/src/Nodes/CircleRemoteConst.test.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+  luci::CircleRemoteConst node;
+  node.dtype(loco::DataType::S32);
+
+  const int32_t scalar = 7;
+  node.bind_buffer(reinterpret_cast<const uint8_t *>(&scalar), sizeof(int32_t));
+
+  ASSERT_EQ(loco::DataType::S32, node.dtype());
+  ASSERT_EQ(node.scalar<loco::DataType::S32>(), scalar);
+}
```
m-bronnikov, line 57: TODO add negative tests, e.g. for prohibited memory access.

seanshpark:
> Some tests (e.g. luci_value_test) uses luci-eval-driver module, which uses luci::Importer for model importing. So, I changed luci-eval-driver in this draft to proof support done correct, but I don't plan to introduce this changes in master.

Can you please explain how current change proves?

compiler/luci-eval-driver/src/EvalDriver.cpp
```diff
@@ -90,8 +90,20 @@ int entry(int argc, char **argv)
...
+  // define model data
+  std::ifstream fs(filename, std::ifstream::binary);
+  if (fs.fail())
+  {
+    throw std::runtime_error("Cannot open model file \"" + std::string(filename) + "\".\n");
+  }
+
+  // model buffer (must be alive longer than importer and interpreter)
+  std::vector<char> model_data((std::istreambuf_iterator<char>(fs)),
+                               std::istreambuf_iterator<char>());
```
seanshpark, line 102: (1) This seems very important condition. How do we let users of luci-import provide this?
(2) `std::vector<char` looks like make a copy content from the file and this seems misleading of not uising constant copy. actully I am confused for this.

m-bronnikov: > (1) This seems very important condition. How do we let users of luci-import provide this? (2) `std::vector<char` looks like make a copy content from the file and this seems misleading of not uising constant copy. actully I am confused for this.

1. I agree with you and this condition confuses me too. I wrote about this in https://github.com/Samsung/ONE/pull/8043#discussion_r754029148 and in https://github.com/Samsung/ONE/issues/8042#issuecomment-974442010. I think we need to enable constant copying by default (as done in draft) and notify user via comments in Importer.h about this. Or introduce another approach or close this draft.
2. This behavior is not confuses me because usage scenario of interpreter on embedded system differs from usage on more powerfull devices. Usually, embedded systems works without file system, so, loading of model to array may be done in speciall .cpp file (e.g. https://github.com/tensorflow/tflite-micro/blob/fdc88436a1928b2314caf5e1445fdf16ccd83d68/tensorflow/lite/micro/examples/hello_world/model.cc) before compilation.

compiler/luci/lang/include/luci/IR/Nodes/CircleRemoteConst.h
seanshpark: Thanks for the explanation. As above comment `Remote` seems ambigious.

As you wrote
```
3. Free model's buffer (this action makes data given to CircleRemoteConst invalid)
```
this can be critical for usage scenario. import routine should have self awared feature not to work for this case.

compiler/luci-eval-driver/src/EvalDriver.cpp
seanshpark: >  I think we need to enable constant copying by default

This is the current behaviour. I would like not to name this to `copy` and name some for the new one.
You've used `Remote` to express the difference of existing CircleConst. Something like, but not `Remote`.

> Or introduce another approach or close this draft.

For safety, I prefer another approach if possible, but have to see...

> This behavior is not confuses me

The import thing is I am confused.

> Usually, embedded systems works without file system

OK, thanks for this sample code. But I cannot image using `vector` can be connected to the const byte buffer. You have not provided any clue.

m-bronnikov:
> > Some tests (e.g. luci_value_test) uses luci-eval-driver module, which uses luci::Importer for model importing. So, I changed luci-eval-driver in this draft to proof support done correct, but I don't plan to introduce this changes in master.
> 
> Can you please explain how current change proves?

I found mistake in changes, which I pushed first and repaired it now. To proof we need change not only `luci_eval_driver` but `luci_interpreter` too.

If something done wrong, we obtain following result with `./nncc test`:

```
93/96 Test #93: luci_value_test ............................***Failed   87.50 sec
FAILED
- ArgMax_002
- ArgMax_003
- ArgMax_U8_002
- ArgMax_U8_003
- Conv2D_000
- Conv2D_001
- Conv2D_002
- Conv2D_U8_000
- DepthwiseConv2D_000
- DepthwiseConv2D_U8_000
- DepthwiseConv2D_001
- Mean_001
- Pad_000
- Pad_U8_000
- Reshape_001
- Reshape_002
- Slice_000
- StridedSlice_000
- StridedSlice_001
- StridedSlice_002
- Transpose_000
- Part_While_000
- Part_While_001

      Start 94: circle2circle_dredd_recipe_test
94/96 Test #94: circle2circle_dredd_recipe_test ............   Passed    2.12 sec
      Start 95: circle_part_value_test
95/96 Test #95: circle_part_value_test .....................   Passed   51.00 sec
      Start 96: luci_pass_value_test
96/96 Test #96: luci_pass_value_test .......................***Failed   32.21 sec
FAILED
- Net_Conv_Add_Mul_000
- Net_Conv_Add_Mul_000
- Net_Conv_Add_Mul_001
- Net_Conv_Add_Mul_002
- Net_Conv_Min_Max_000
- Net_Conv_Min_Relu_000
- Net_Conv_Relu6_000
- Net_DwConv_BN_000
- Net_DwConv_BN_001
- Net_Reshape_Neg_000
- Net_Reshape_Reshape_000
- Net_Squeeze_Squeeze_000
- Net_TConv_Add_000
- Net_TConv_Add_001
- Net_TConv_Add_002
- Net_TConv_BN_000
- Net_TConv_BN_001
- Net_TConv_BN_002
- Net_InstanceNorm_001
- Net_InstanceNorm_002
- Net_InstanceNorm_003
- Net_StridedSlice_StridedSlice_000


98% tests passed, 2 tests failed out of 96

Total Test time (real) = 231.69 sec

The following tests FAILED:
	 93 - luci_value_test (Failed)
	 96 - luci_pass_value_test (Failed)
Errors while running CTest
```

Currently, all test passed well:

```
100% tests passed, 0 tests failed out of 96

Total Test time (real) = 226.81 sec
```

compiler/luci/lang/include/luci/IR/Nodes/CircleRemoteConst.h
m-bronnikov: > As above comment Remote seems ambigious.

@seanshpark What do you think about `CircleReferenceConst` instead `CircleRemoteConst` or may be `CircleConstProvider`?

seanshpark: Um...  It would be better to explain why `Reference` or `Provider` as I can't find a clue why you used these words.
Please consider that I cannot deeply think on this topic as I am concentrating on other tasks.

m-bronnikov:
## Summarize of latest changes

Let me please summarize what changed in this draft since last review!

### Find new name for introduced note

https://github.com/Samsung/ONE/pull/8043#discussion_r753867495 and https://github.com/Samsung/ONE/pull/8043#discussion_r754122102 requested renaming of `CircleRemoteConst` to something else. I choosen `CircleConstReference` in rebased draft because introduced node similar to `CircleConst` node without copyng (by reference). If this name looks ambigious I'm happy to hear any suggestion about name.

### Refactor some code

- https://github.com/Samsung/ONE/pull/8043#discussion_r753867255 requested to introduce copy of `convert_graph` instead changing existing one. Done in this draft!
- https://github.com/Samsung/ONE/pull/8043#discussion_r753867039 requested moving of added CTOR below of existing one. Done in this draft!

Please correct me if I'm forgot to add something here.

### Change desing 

https://github.com/Samsung/ONE/pull/8043#discussion_r754071171 talks about negative side of previos design: destroying of data buffer makes all CircleRemoteConst nodes invalid. This behavior makes usage of `luci` very fragile. I changed this design to another. Let me share details here!

When i thought about how to introduce new feature to importer I found 2 basic requirement:
1. This feature should be optional and disabled by default for backward capability.
2. This feature should be safe (last approach violated this condition)

So, I applied following changes:

1. Add new feature to provide model's buffer wrapped to `shared_ptr` into Importer: 
https://github.com/Samsung/ONE/blob/85734939a9c8a5d04761b09de5783ccd259cb6c5/compiler/luci/import/include/luci/Importer.h#L44-L47
https://github.com/Samsung/ONE/blob/85734939a9c8a5d04761b09de5783ccd259cb6c5/compiler/luci/import/include/luci/Importer.h#L55
`nullptr` shared_ptr means buffer is not provided and `CircleConst` nodes with data copying is using. Otherwise, Importer uses `CircleConstReference` nodes when possible.
2. To use luci module safety we need to ensure models buffer alive during usage. I extended luci::Module with shared_ptr to buffer data too:
https://github.com/Samsung/ONE/blob/cd70a93330488762b1053989898d6ae34707e828/compiler/luci/lang/include/luci/IR/Module.h#L86-L94

**Why shared_ptr?**

1. shared_ptr alive while alive at least one entity, which points to data. So, passing of `shared_ptr` to `luci::Module` via importer makes usage of `CircleConstReference` nodes safe.
2. shared_ptr has argument - custom deleter. By modifying deleter we can deprecate memory release in cases, when model provided as const pointer (like in embedded systems).

### Some proof of correct

In prev version of draft I changed luci_eval_driver to proof changes applied well. This kind of changes looks confusing and unrelated https://github.com/Samsung/ONE/pull/8043#issuecomment-975304714. In new version I introduced handwritten test as new module (model-import-verifier) with necessary changes to luci-interpreter, As I mentioned above https://github.com/Samsung/ONE/pull/8043#issuecomment-975244378, I don't plan introduce this changes to `master`.

Introduced test works with created using following script `Conv2D` model

<details>
<summary> script to create model </summary>

```python
import tensorflow.compat.v1 as tf
from tensorflow.compat.v1 import nn
import numpy as np

def conv_kwargs(input_depth, filters_count=1, window_size=1, stride_size=1):
    return {
        'filter': np.random.randn(window_size, window_size, input_depth, filters_count).astype(np.float32),
        'padding': 'SAME',
        'strides': [1, stride_size, stride_size, 1]
    }

def conv2d(input_layer, **kwargs):
    bias = np.random.randn(kwargs['filter'].shape[3]).astype(np.float32)
    conv = nn.conv2d(input_layer, **kwargs)
    return nn.bias_add(conv, bias)


with tf.Graph().as_default() as graph:
    # create model
    input_shape = (1, 5, 5, 32)
    input_layer = tf.placeholder(dtype=tf.float32, shape=input_shape)
    output_layer = conv2d(input_layer, **conv_kwargs(input_depth=32, filters_count=64, window_size=2))

    # save to tflite
    with tf.Session() as sess:
        sess.run(tf.global_variables_initializer())
        converter = tf.lite.TFLiteConverter.from_session(sess, [input_layer], [output_layer])
        converter.allow_custom_ops = True
        content = converter.convert()
        with open("conv2d.tflite", 'wb') as binf:
```

</details>

Model: 

image


In this test I used 2 mode of importing (with and without constant copying) and 2 kinds of models representation: in file (as did in luci-eval-driver) and in constant array (as for embedded systems).

https://github.com/Samsung/ONE/blob/5da8e882d4ba925d6f3951e4f3beb5e89da859d3/compiler/model-import-verifier/src/EvalDriver.cpp#L125-L163

If all output values for same input after model's execution are equal, test passed:

https://github.com/Samsung/ONE/blob/5da8e882d4ba925d6f3951e4f3beb5e89da859d3/compiler/model-import-verifier/src/EvalDriver.cpp#L165-L182


--------------------

+cc @seanshpark

compiler/model-import-verifier/src/sample_model.cpp
```diff
@@ -0,0 +1,2124 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "sample_model.h"
+
+const uint8_t conv2d_circle[] = {
```
m-bronnikov, line 18: Note: usually, model's for embedded systems stored in this format.

compiler/model-import-verifier/src/EvalDriver.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+  auto const pointer_to_pass = copy_const ? nullptr : model_data;
+  return luci::Importer(pointer_to_pass).importModule(circle::GetModel(model_data.get()));
+}
+
+std::unique_ptr<luci::Module> import_model_constant_buffer(const uint8_t *buffer, bool copy_const)
+{
+  // wrap constant buffer to shared ptr with deprecated deletion
+  auto const nothing_to_do_deleter = [](uint8_t *) { /* Do nothing */ };
+  std::shared_ptr<uint8_t[]> model_data(const_cast<uint8_t *>(buffer), nothing_to_do_deleter);
```
m-bronnikov, line 62: Wrap to `shared_ptr` and use empty deleter for models provided as const pointer to buffer.

compiler/model-import-verifier/src/EvalDriver.cpp
```diff
@@ -0,0 +1,184 @@
...
+  // define model's size
+  uint32_t model_size = 0;
+  {
+    fs.seekg(0, fs.end);
+    model_size = fs.tellg();
+    fs.seekg(0, fs.beg);
+  }
+
+  // allocate memory and provide memory management to shared ptr
+  std::shared_ptr<uint8_t[]> model_data(new uint8_t[model_size]);
```
m-bronnikov, line 51: For importing from file use default deleter and delegate memory management to `shared_ptr`.

seanshpark:
1. `CircleConstReference`: Your reasoning has points. But still I'm not convinced. And cannot provide better one as I have not enough time for this topic. What I understand of this whole thing is to use file (or constant buffer for embedded devices) as source of network, including weight constants.
One popped up while typing this. `CircleConstStatic` as for it's from static buffer.

2. What I now distinguish is, using map() from original Tensorflow lite recommandartion to use weight constantas without coping and the one we're working for embedded device are different.
For embedded device, there is no file open or close but just passing buffer pointer.
For this, does using `shared_ptr` makes sense? For providing an API, I think it is, but for your purpose, I'm not sure.
Anyway we have to have some API...
In frontend, we didn't use `shared_ptr` in the first place as when @parjong designed, `shared_ptr` can give  convenience but also ambigious of memory management. For me, I've little preference of using it or not. But I'd like to follow his thoughts not to use ` shared_ptr` if we can.
For this I'm not sure using `shared_ptr` is appropriate or not...

compiler/luci/import/include/luci/Importer.h
```diff
@@ -36,7 +36,12 @@ class Importer final
   Importer();
 
 public:
-  explicit Importer(const GraphBuilderSource *source) : _source{source}
+  explicit Importer(const GraphBuilderSource *source) : _source(source)
```
seanshpark, line 39: Do you have to change this?

compiler/luci/import/src/Importer.cpp
```diff
@@ -126,6 +127,204 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
...
+  // NOTE Origin is intentionally not provided for constants.
+  for (uint32_t i = 0; i < tensors.size(); ++i)
+  {
+    luci::CircleNode *const_node = nullptr;
+
+    // create copy of const data if data is 'String' or copying requested by caller
+    if (copy_consts || luci::luci_datatype(tensors[i]->type()) == loco::DataType::STRING)
+      const_node = luci::create_circleconst(&gb_context, i);
+    else
+      const_node = luci::create_circle_const_ref(&gb_context, i);
```
seanshpark, line 326: withour `ref` was `create_circleconst` and this is `create_circle_const` + `_ref` which, I hope you can follow as `create_circleconst_ref` for the name.

compiler/luci/import/src/Importer.cpp
```diff
@@ -274,7 +473,8 @@ std::unique_ptr<loco::Graph> Importer::import(const circle::Model *model) const
     return nullptr;
 
   // Convert circle::Model to loco::Graph
-  convert_graph(*source_ptr, reader, graph.get());
+  // not copy constants if model's buffer provided
+  convert_graph(*source_ptr, reader, graph.get(), _model_buffer == nullptr);
```
seanshpark, line 477: I cann't check existing `convert_graph` is not used anymore from this change. is it not used anymore?

compiler/luci/lang/include/luci/IR/Nodes/CircleConstReference.h
```diff
@@ -0,0 +1,63 @@
...
+
+private:
+  struct RemoteBuffer
+  {
+    const uint8_t *data = nullptr;
+    uint32_t size = 0;
+  };
+
+private:
+  RemoteBuffer _buffer;
```
seanshpark, line 58: I think this needs renaming

compiler/model-import-verifier/src/sample_model.cpp
```diff
@@ -0,0 +1,2124 @@
...
+  0x14, 0x00, 0x10, 0x00, 0x00, 0x00, 0x0c, 0x00, 0x08, 0x00, 0x04, 0x00, 0x0e, 0x00, 0x00, 0x00,
+  0x14, 0x00, 0x00, 0x00, 0x14, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x1c, 0x00, 0x00, 0x00,
+  0x04, 0x00, 0x04, 0x00, 0x04, 0x00, 0x00, 0x00, 0x0b, 0x00, 0x00, 0x00, 0x50, 0x6c, 0x61, 0x63,
+  0x65, 0x68, 0x6f, 0x6c, 0x64, 0x65, 0x72, 0x00, 0x04, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00,
+  0x05, 0x00, 0x00, 0x00, 0x05, 0x00, 0x00, 0x00, 0x20, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00,
+  0x0c, 0x00, 0x00, 0x00, 0x00, 0x00, 0x06, 0x00, 0x08, 0x00, 0x07, 0x00, 0x06, 0x00, 0x00, 0x00,
+  0x00, 0x00, 0x00, 0x03, 0x11, 0x00, 0x00, 0x00, 0x4f, 0x4e, 0x45, 0x2d, 0x74, 0x66, 0x6c, 0x69,
+  0x74, 0x65, 0x32, 0x63, 0x69, 0x72, 0x63, 0x6c, 0x65, 0x00, 0x00, 0x00};
+
+const uint32_t conv2d_circle_len = 33660;
```
seanshpark, line 2124: For proof you can use just small 2x2x... weight kernel. I don't think this huge file for the network is a good idea.

seanshpark:
What keeps me itchy is that mixing methods in the Import class for the conventional and for embedded is not a good idea.
They have differnt environment, constraints, policy and philosophy.
What do you think of this? We came a long path but can you rethink of another path?

compiler/luci/import/include/luci/Importer.h
m-bronnikov: No, it's just a strange diff =)

I moved introduced ctor below of existing one (without any modifying) in this changes.

compiler/luci/import/src/Importer.cpp
m-bronnikov: Yes, it not used anymore =)

Checked with removing of old `convert_graph`: project build finishes successfully.

seanshpark: OK, then we can change existing one to your new one.

binarman:
@seanshpark 
I have yet another idea how to specialize importer.

What do you think about extending `GraphBuilderSource` interface?

As for now:
``` c++
struct GraphBuilderSource
{
  virtual const GraphBuilderBase *lookup(const circle::BuiltinOperator &op) const = 0;
};
```

What I propose:
``` c++
struct GraphBuilderSource
{
  virtual const GraphBuilderBase *lookup(const circle::BuiltinOperator &op) const = 0;
  virtual const TensorBuilderBase *tensor_builder() const = 0;
};
```

Where `TensorBuilderBase` provides methods for creation of graph inputs, graph outputs and constant nodes:
``` c++
class TensorBuilderBase
{
  virtual CircleNode *build_input(const circle::TensorT &t, GraphBuilderContext *context) const = 0;
  virtual CircleNode *build_output(const circle::TensorT &t, GraphBuilderContext *context) const = 0;
  virtual CircleNode *build_constant(const circle::TensorT &t, GraphBuilderContext *context) const = 0;
};
```

Using this extension we can eliminate changes in main Importer interface.
All specific logic could be moved to a graph builder.

seanshpark:
> Using this extension we can eliminate changes in main Importer interface.

I'm not worried about chaning Importer itself.
Main concern is about using Importer with embedded methods and using mixed of regular(copy) and reference.
Currenrly there is no protection to avoid this.
What I could think of is to separate interface and then find something somehow later.
For embedded systems. use new Importer.

Current draft code does not cover all the tools that use Importer. And I don't think check all of them is a good idea.

> All specific logic could be moved to a graph builder.

Sorry but I can't catch the idea/benifit of this. I assume this will result lots of changes, which I don't prefer.

m-bronnikov:
@seanshpark Sorry for a long response from last review. I would like to show another path, which requires less changes in luci/import. This approach based on  @binarman's idea  https://github.com/Samsung/ONE/pull/8043#issuecomment-982959618 to delegate constant nodes importing GraphBuilderSource to import constant nodes.  PTAL https://github.com/Samsung/ONE/pull/8190#issuecomment-996675761

P.S. I would like to show this approach via another draft #8190 since I don't want to lose the current changes.

m-bronnikov:
Close because another solution introduced.

PR #8044, opened 2021-11-22
[luci-interpreter] New PAL interface for DepthwiseConv2D temporary buffers

BalyshevArtem:
Implement and use PAL interface for temporary buffers allocation for DepthwiseConv2D.

This is the second task from issue #7906

### DIFF
--- a/compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
+++ b/compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
@@ -31,7 +31,8 @@ DepthwiseConvPerChannel(const tflite::DepthwiseParams &params, const int32_t *ou
                         const T *input_data, const tflite::RuntimeShape &filter_shape,
                         const T *filter_data, const tflite::RuntimeShape &bias_shape,
                         const int32_t *bias_data, const tflite::RuntimeShape &output_shape,
-                        T *output_data)
+                        T *output_data, const tflite::RuntimeShape &scratchpad_shape,
+                        T *scratchpad_data)
 {
   {
     // MARK: At this moment this operation is not supported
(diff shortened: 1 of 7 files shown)


compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
```diff
@@ -116,15 +121,10 @@ DepthwiseConvPerChannel<int8_t>(const tflite::DepthwiseParams &params,
...
-
-    auto buffer = std::make_unique<int8_t[]>(buf_size);
-    assert(buffer != nullptr);
-
     cmsis_nn_context ctx;
-    ctx.buf = buffer.get();
-    ctx.size = buf_size;
+    ctx.buf = scratchpad_data;
+    ctx.size = scratchpad_shape.Dims(0);
+    ;
```
m-bronnikov, line 127: Redundant?

m-bronnikov (commented):
I

compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
BalyshevArtem: ye, sorry! remove it

m-bronnikov (approved):

compiler/luci-interpreter/pal/cmsisnn/PALDepthwiseConv2d.h
```diff
@@ -48,24 +49,28 @@ DepthwiseConvPerChannel(const tflite::DepthwiseParams &params, const int32_t *ou
...
 }
 
 template <>
-inline void
-DepthwiseConvPerChannel<int8_t>(const tflite::DepthwiseParams &params,
-                                const int32_t *output_multiplier, const int32_t *output_shift,
-                                const tflite::RuntimeShape &input_shape, const int8_t *input_data,
-                                const tflite::RuntimeShape &filter_shape, const int8_t *filter_data,
-                                const tflite::RuntimeShape &bias_shape, const int32_t *bias_data,
-                                const tflite::RuntimeShape &output_shape, int8_t *output_data)
```
binarman, line 61: Why did you change formatting?

binarman: Nevermind, looks like this is due to auto format

binarman (approved):
LGTM

compiler/luci-interpreter/src/kernels/DepthwiseConv2D.test.cpp
```diff
@@ -166,6 +172,7 @@ TEST_F(DepthwiseConv2DTest, SInt16)
   Tensor bias_tensor =
     makeInputTensor<DataType::S64>(bias_shape, 0.25 * 0.2, 0, bias_data, _memory_manager.get());
   Tensor output_tensor = makeOutputTensor(DataType::S16, 0.5, 0);
+  Tensor scratchpad(DataType::S64, Shape({}), {}, "");
```
binarman, line 175: Note: this type looks different from other places.

binarman:
Just one another thought about scratchpad tensors:
I noticed that we should decide data type on creation, but this could be PAL defined detail...
Currently Tensor interface do not provide knob to adjust tensor type after creation, maybe we need one?

BalyshevArtem:
> Just one another thought about scratchpad tensors: I noticed that we should decide data type on creation, but this could be PAL defined detail... Currently Tensor interface do not provide knob to adjust tensor type after creation, maybe we need one?
@binarman,
I'm not sure I got you right. Could you explain more detailed, when it can be useful?
As I understand the type of scratchpad tensor depends on the type of input tensor, and I don't really understand in which cases we want to change this.

binarman:
@BalyshevArtem 
I've added comment regarding this topic in another PR: https://github.com/Samsung/ONE/pull/8047#discussion_r758394641

The thing is: scratchpad is platform specific, so it is hard to choose data type during model loading tensor, because PAL is not used on model loading pass.

PR #8047, opened 2021-11-23
[luci-interpreter] New PAL interface for AvgPool2D temporary buffers

BalyshevArtem:
Implement and use PAL interface for temporary buffers allocation for AvgPool2D .

This is the third task from issue #7906

### DIFF
--- a/compiler/luci-interpreter/src/loader/nodes/AveragePool2D.cpp
+++ b/compiler/luci-interpreter/src/loader/nodes/AveragePool2D.cpp
@@ -17,6 +17,7 @@
 #include "Builders.h"
 
 #include "kernels/AveragePool2D.h"
+#include <luci/Plan/CircleNodeExecutionPlan.h>
 
 namespace luci_interpreter
 {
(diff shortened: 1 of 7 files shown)


m-bronnikov (approved):

compiler/luci-interpreter/src/loader/nodes/AveragePool2D.cpp
```diff
@@ -40,7 +41,25 @@ std::unique_ptr<Kernel> build_kernel_CircleAveragePool2D(const luci::CircleNode
   params.stride_width = node->stride()->w();
   params.activation = node->fusedActivationFunction();
 
-  return std::make_unique<kernels::AveragePool2D>(input, output, params);
+  auto scratchpad =
+    std::make_unique<Tensor>(input->element_type(), Shape({}), AffineQuantization{}, "");
```
binarman, line 45: Let's set data type as `INT8` and let PAL cast it if needed.

Data type of this tensor is platform specific.
So, what we actually need to do is:
- option 1) introduce another PAL function for construction of scratchpad tensor in loader
- option 2) introduce new tensor method to change it's type

I don't think these interface additions are actually improving something that we could use in future, so, lets stick to simple 1-byte type.

Also, add comment here, something like:
```
// It is unknown what data will be stored in scratchpad tensor,
// using SINT8 as a most general option
```

binarman: After offline discussion with @BalyshevArtem we decided to merge this PR first, then apply changes, like #8069

binarman (approved):
LGTM

PR #8049, opened 2021-11-24
[circle-part] Update readme

seanshpark:
This will update readme with example and work todo.


seanshpark:
From #6231

seanshpark:
@chunseoklee , if you think this addition is not necessary, I can drop this.

chunseoklee (approved):

PR #8063, opened 2021-11-29
[one-cmds] Add ONNX tests

binarman:
This PR adds tests that uses onnx examples from res/PyTorchExamples.

related issue: #7939

### DIFF
--- a/compiler/one-cmds/tests/CMakeLists.txt
+++ b/compiler/one-cmds/tests/CMakeLists.txt
@@ -73,3 +73,5 @@ install(FILES ${PREPROCESS_IMAGES_PY}
 
 install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/README.txt
         DESTINATION test)
+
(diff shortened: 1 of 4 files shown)


binarman:
I did not include these tests in any automatic testing system, because:
- some of them are failing (need to decide what to do with them first)
- ONNX is not something we are focusing a lot right now, so I think we don't need to run them frequently

seanshpark:
@binarman , can you please describe what you want to achieve from this tesT?

binarman:
> can you please describe what you want to achieve from this test?

I want to get get list of successfully converted ONNX models.
This is a side product of PyTorch tests development from #8061.
I think it is good to have them to compare different ways to export PyTorch models.

P.s. I don't want to include them in nightly, at least for now

seanshpark (approved):
LGTM

lucenticus (approved):
LGTM

PR #8068, opened 2021-11-30
DRAFT: check support TF 2.7.0

seanshpark:
on-going draft to check support TF 2.7.0.


seanshpark:
Support TF2.8.0 in progress

PR #8069, opened 2021-11-30
[luci-interpreter] Let PAL cast temporary tensors for Conv2d

BalyshevArtem:
This commit lets change data type for temporary tensors for Conv2d in PAL interface using new tensor method.

according to https://github.com/Samsung/ONE/pull/8047#discussion_r758394641 comment

### DIFF
--- a/compiler/luci-interpreter/pal/cmsisnn/PALConv2d.h
+++ b/compiler/luci-interpreter/pal/cmsisnn/PALConv2d.h
@@ -136,7 +136,7 @@ static inline void ConvPerChannel(const tflite::ConvParams &params, const int32_
 }
 
 static inline void SetupScratchpadTensor(luci_interpreter::Tensor *scratchpad,
-                                         const luci_interpreter::DataType &data_type,
+                                         const luci_interpreter::DataType &input_data_type,
                                          const tflite::ConvParams &params,
                                          const tflite::RuntimeShape &input_shape,
                                          const tflite::RuntimeShape &filter_shape,
@@ -148,6 +148,7 @@ static inline void SetupScratchpadTensor(luci_interpreter::Tensor *scratchpad,
 
   if (conv_params.dilation.h == 1 && conv_params.dilation.w == 1)
   {
+    assert(input_data_type == loco::DataType::S8);
     const int32_t batches = tflite::MatchingDim(input_shape, 0, output_shape, 0);
     const int32_t input_depth = tflite::MatchingDim(input_shape, 3, filter_shape, 3);
     const int32_t output_depth = tflite::MatchingDim(filter_shape, 0, output_shape, 3);
--- a/compiler/luci-interpreter/pal/linux/PALConv2d.h
+++ b/compiler/luci-interpreter/pal/linux/PALConv2d.h
@@ -29,10 +29,20 @@ static inline void Conv(const tflite::ConvParams &params, const tflite::RuntimeS
                         float *output_data, const tflite::RuntimeShape &scratchpad_shape,
                         float *scratchpad_data)
 {
+  (void)scratchpad_shape;
   if (scratchpad_data)
   {
+    const int32_t batches = tflite::MatchingDim(input_shape, 0, output_shape, 0);
+    const int32_t input_depth = tflite::MatchingDim(input_shape, 3, filter_shape, 3);
+    const int32_t output_height = output_shape.Dims(1);
+    const int32_t output_width = output_shape.Dims(2);
+    const int32_t filter_height = filter_shape.Dims(1);
+    const int32_t filter_width = filter_shape.Dims(2);
+    tflite::RuntimeShape im2col_shape{batches, output_height, output_width,
+                                      input_depth * filter_height * filter_width};
+
     tflite::optimized_ops::Conv(params, input_shape, input_data, filter_shape, filter_data,
-                                bias_shape, bias_data, output_shape, output_data, scratchpad_shape,
+                                bias_shape, bias_data, output_shape, output_data, im2col_shape,
                                 scratchpad_data);
   }
   else
@@ -75,7 +85,7 @@ static inline void ConvPerChannel(const tflite::ConvParams &params, const int32_
 }
 
 static inline void SetupScratchpadTensor(luci_interpreter::Tensor *scratchpad,
-                                         const luci_interpreter::DataType &data_type,
+                                         const luci_interpreter::DataType &input_data_type,
                                          const tflite::ConvParams &params,
                                          const tflite::RuntimeShape &input_shape,
                                          const tflite::RuntimeShape &filter_shape,
@@ -90,7 +100,7 @@ static inline void SetupScratchpadTensor(luci_interpreter::Tensor *scratchpad,
     params.dilation_height_factor != 1 || params.dilation_width_factor != 1;
   const bool need_non_dilated_scratchpad = params.stride_height != 1 || params.stride_width != 1 ||
                                            filter_height != 1 || filter_width != 1;
-  auto _need_scratchpad = data_type != luci_interpreter::DataType::S16 &&
+  auto _need_scratchpad = input_data_type != luci_interpreter::DataType::S16 &&
                           (need_dilated_scratchpad || need_non_dilated_scratchpad);
 
   if (_need_scratchpad)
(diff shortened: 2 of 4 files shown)


compiler/luci-interpreter/src/loader/nodes/Conv2D.cpp
```diff
@@ -35,8 +35,9 @@ std::unique_ptr<Kernel> build_kernel_CircleConv2D(const luci::CircleNode *circle
   const Tensor *bias = helper.getOptionalInputTensor(node->bias());
   Tensor *output = helper.getOutputTensor(node);
 
-  auto scratchpad =
-    std::make_unique<Tensor>(input->element_type(), Shape({}), AffineQuantization{}, "");
+  // It is unknown what data will be stored in scratchpad tensor,
+  // using SINT8 as a most general option
+  auto scratchpad = std::make_unique<Tensor>(DataType::S8, Shape({}), AffineQuantization{}, "");
```
m-bronnikov, line 40: `loco` provides special `DataType` value for case of unknown dtype.

```suggestion
  auto scratchpad = std::make_unique<Tensor>(DataType::Unknown, Shape({}), AffineQuantization{}, "");
```

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
```diff
@@ -162,6 +162,8 @@ class Tensor
 
   void set_offset(int32_t offset) { _offset = offset; }
 
+  void change_data_type(DataType element_type) { _element_type = element_type; }
```
m-bronnikov, line 165: (optional) Personally, I think better to name this method `set_data_type` in accordance with other methods like `set_observable` or `set_data_buffer`.

compiler/luci-interpreter/src/loader/nodes/Conv2D.cpp
```diff
@@ -35,8 +35,10 @@ std::unique_ptr<Kernel> build_kernel_CircleConv2D(const luci::CircleNode *circle
   const Tensor *bias = helper.getOptionalInputTensor(node->bias());
   Tensor *output = helper.getOutputTensor(node);
 
+  // It is unknown what data will be stored in scratchpad tensor,
+  // using SINT8 as a most general option
```
m-bronnikov, line 39: Comment is redundant and invalid for now.

m-bronnikov: Better to remove =)

compiler/luci-interpreter/src/loader/nodes/Conv2D.cpp
binarman: @m-bronnikov 
I still think it is better to stick to UINT8 and do not change it after tensor is created.

This scratch pad is used by PAL for different purposes and sometimes we can not even predict what is the actual data type is required and will be stored.
For example:
- [here](https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/NN/Source/ConvolutionFunctions/arm_convolve_s8.c#L136) it is used as 8 bit integer array
- [in the same function](https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/NN/Source/ConvolutionFunctions/arm_convolve_s8.c#L225) it is used as a 16 bit integer array

binarman: I propose to present `scratchpad` to code outside of PAL is a simple blob of data, without any explicit semantics.

I think UINT8 data type and size in bytes looks the most appropriate option to represent scratchpad to "outer world".

compiler/luci-interpreter/src/loader/nodes/Conv2D.cpp
binarman: see https://github.com/Samsung/ONE/pull/8069#discussion_r759210782

compiler/luci-interpreter/src/loader/nodes/Conv2D.cpp
m-bronnikov: `UINT8` or `SINT8` https://github.com/Samsung/ONE/pull/8047#discussion_r758394641? =)

I like `loco::DataType::Unknown` because, as you mentioned above in comment line and in https://github.com/Samsung/ONE/pull/8069#discussion_r759210782, we don't really know which datatype will be used by implementation in PAL internally and `Unknown` looks more natural from reviewer point of view. 

However, I don't see any key points here and no matter for me what value will be setted by default. So, I agree with any approach.

compiler/luci-interpreter/pal/cmsisnn/PALConv2d.h
```diff
@@ -136,12 +136,13 @@ static inline void ConvPerChannel(const tflite::ConvParams &params, const int32_
...
 
 static inline void SetupScratchpadTensor(luci_interpreter::Tensor *scratchpad,
-                                         const luci_interpreter::DataType &data_type,
+                                         const luci_interpreter::DataType &input_data_type,
                                          const tflite::ConvParams &params,
                                          const tflite::RuntimeShape &input_shape,
                                          const tflite::RuntimeShape &filter_shape,
                                          const tflite::RuntimeShape &output_shape)
 {
+  (void)input_data_type;
```
binarman, line 145: optional: you can add `assert(input_data_type == loco::DataType::S8);`

binarman (approved):
LGTM

compiler/luci-interpreter/pal/linux/PALConv2d.h
```diff
@@ -100,8 +110,10 @@ static inline void SetupScratchpadTensor(luci_interpreter::Tensor *scratchpad,
     const int32_t output_height = output_shape.Dims(1);
     const int32_t output_width = output_shape.Dims(2);
 
-    luci_interpreter::Shape scratchpad_shape{batches, output_height, output_width,
-                                             input_depth * filter_height * filter_width};
+    auto data_type_size = (int32_t)luci_interpreter::getDataTypeSize(input_data_type);
```
m-bronnikov, line 113: (optional) Better to use `static_cast`

m-bronnikov (dismissed):

compiler/luci-interpreter/pal/cmsisnn/PALConv2d.h
BalyshevArtem: Done

compiler/luci-interpreter/pal/linux/PALConv2d.h
BalyshevArtem: Done

m-bronnikov (approved):
LGTM

m-bronnikov (approved):

compiler/luci-interpreter/pal/linux/PALConv2d.h
```diff
@@ -29,10 +29,20 @@ static inline void Conv(const tflite::ConvParams &params, const tflite::RuntimeS
...
   if (scratchpad_data)
   {
+    const int32_t batches = tflite::MatchingDim(input_shape, 0, output_shape, 0);
+    const int32_t input_depth = tflite::MatchingDim(input_shape, 3, filter_shape, 3);
+    const int32_t output_height = output_shape.Dims(1);
+    const int32_t output_width = output_shape.Dims(2);
+    const int32_t filter_height = filter_shape.Dims(1);
+    const int32_t filter_width = filter_shape.Dims(2);
+    tflite::RuntimeShape im2col_shape{batches, output_height, output_width,
+                                      input_depth * filter_height * filter_width};
```
m-bronnikov, line 42: Q. Why we set `im2col` shape here when `scratchpad_shape` has same value?

m-bronnikov: Ahh, I got it, nevermind.

compiler/luci-interpreter/src/loader/nodes/Conv2D.cpp
binarman: @m-bronnikov 
I prefer UINT8 (not a big different from SINT8, but internally it is used uint8, so I changes my mind from comment in 8047 :slightly_smiling_face: ) over Unknown, because Unknown semantically does not have "one element" size.
So we have kind of ambiguity if we use Unknown.

When we are using UINT8 we can see scratchpad as a simple buffer of bytes, which has it's size encoded in tensor shape.

binarman (approved):

m-bronnikov (approved):

PR #8051, opened 2021-11-25
[Draft][Compiler] Introcuce utils for reading TFLite files

llFreetimell:
This draft is for introducing utils for reading TFLite files.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

---

Parent Issue : #8050

Issue #8050, opened 2021-11-25
[Compiler] Introduce utils for reading TFLite file


llFreetimell:
### What?

Let's introduce utils for reading TFLite files.
(e.g. Getting builtin code, checking custom, etc)

### Why?

Currently, we have some difficulties such as following.

- Some modules have duplicated codes to read TFLite file
- When TFLite schema is updated, we should update all of the codes
- When we need to fix something, we should modify all of the codes

By introducing common utility functions by wrapping TFLite flatbuffer functions, we may expect following things.

- Reduce duplicated codes
- When can provide utility functions according to each version of TFLite
- When TFLite schema is updated or when we need to fix something, we may modify codes only once.

### PRs

- [x] Introduce helper functions
  - #8082 
  - #8091
- [x] Replace helper functions in `tflchef`
  - #8095
- [x] Replace helper functions in `tfldump`
  - #8093
- [x] Replace helper functions in `tfl-inspect`
  - #8092
- [x] Replace helper functions in `tflite2circle`
  - #8094

#### Draft

- #8051

llFreetimell:
This will be rebased/updated after TF 2.7.0 is applied

seanshpark:
> This will be rebased/updated after TF 2.7.0 is applied

We will go with TF 2.6.0 for a while. Please ask @lemmaa for background.
And TF2.7.0 cannot run on Ubuntu 18.04 by default. This requires >= python 3.7.

llFreetimell:
> We will go with TF 2.6.0 for a while

Oh, thanks for notification. Then I shall proceed with current draft :)

lemmaa:
> We will go with TF 2.6.0 for a while. Please ask @lemmaa for background.

In a nutshell, just as we followed ACLs one version back, TensorFlow support intends to do the same. :)

llFreetimell:
Done!

PR #8082, opened 2021-12-02
[mio-tflite260] Introduce utility functions for reading TFLite file

llFreetimell:
By introducing utility functions for reading TFLite file, we may expect following things.
- Reduce duplicated codes
- Reduce code modifications When TFLite schema is updated

---

Parent Issue :#8050
Draft : #8051 

NOTE Implementation of each function is copied from original codes, not modified :)

### DIFF
--- a/compiler/mio-tflite260/CMakeLists.txt
+++ b/compiler/mio-tflite260/CMakeLists.txt
@@ -47,3 +47,10 @@ endif(NOT TensorFlowGEMMLowpSource_FOUND)
 add_library(mio_tflite260_inc INTERFACE)
 target_include_directories(mio_tflite260_inc SYSTEM INTERFACE "${TensorFlowSource_DIR}")
 target_include_directories(mio_tflite260_inc SYSTEM INTERFACE "${TensorFlowGEMMLowpSource_DIR}")
+
+file(GLOB_RECURSE SOURCES "src/*.cpp")
+
+add_library(mio_tflite260_helper STATIC ${SOURCES})
+target_include_directories(mio_tflite260_helper PRIVATE src)
+target_include_directories(mio_tflite260_helper PUBLIC include)
+target_link_libraries(mio_tflite260_helper mio_tflite260)
--- a/compiler/mio-tflite260/include/mio_tflite260/Helper.h
+++ b/compiler/mio-tflite260/include/mio_tflite260/Helper.h
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __MIO_TFLITE260_HELPER_H__
+#define __MIO_TFLITE260_HELPER_H__
+
+#include <mio/tflite/schema_generated.h>
+
+namespace mio
+{
+
+tflite::BuiltinOperator builtin_code_neutral(const tflite::OperatorCode *opcode);
+bool is_valid(const tflite::OperatorCode *opcode);
+bool is_custom(const tflite::OperatorCode *opcode);
+std::string opcode_name(const tflite::OperatorCode *opcode);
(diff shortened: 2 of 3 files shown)


compiler/mio-tflite260/CMakeLists.txt
```diff
@@ -47,3 +47,10 @@ endif(NOT TensorFlowGEMMLowpSource_FOUND)
 add_library(mio_tflite260_inc INTERFACE)
 target_include_directories(mio_tflite260_inc SYSTEM INTERFACE "${TensorFlowSource_DIR}")
 target_include_directories(mio_tflite260_inc SYSTEM INTERFACE "${TensorFlowGEMMLowpSource_DIR}")
+
+file(GLOB_RECURSE SOURCES "src/*.cpp")
+
+add_library(mio_tflite260_util ${LUCI_LIBRARY_TYPE} ${SOURCES})
```
seanshpark, line 53: (1) mio_tflite260_util is not luci
(2) assume mio_tflite260_util is shared, you may need to check with debian package generation.
(3) using name util was long time ago rejected by @parjong as for broad meaning. instead helper was used or you can provide another better one.

compiler/mio-tflite260/include/mio_tflite260/Util.h
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#ifndef __MIO_TFLITE260_UTIL_H__
+#define __MIO_TFLITE260_UTIL_H__
+
+#include <mio/tflite/schema_generated.h>
+
+namespace mio
+{
+namespace tfl
```
seanshpark, line 24: other libraries used `tflite` not `tfl`

compiler/mio-tflite260/CMakeLists.txt
llFreetimell: > (1) mio_tflite260_util is not luci

:+1:

> (2) assume mio_tflite260_util is shared, you may need to check with debian package generation.

I didn't catch this thanks :)
I changed it as static!

> (3) using name util was long time ago rejected by @parjong as for broad meaning. instead helper was used or you can provide another better one.

I will use `helper` instead :)

compiler/mio-tflite260/include/mio_tflite260/Util.h
llFreetimell: I tried to modify this as `tflite` but because `schema_generated.h` is already using `namespace tflite`, I failed :(
So I will remove `namespace tfl` and just only leave it as `mio` :)

seanshpark: > is already using namespace tflite, I failed :(

For me, just `mio` is also ok.
I', not sure what may be the problem, but just to check the failure, can you try with `::mio::tflite` from the caller?

llFreetimell: Did you mean following trial?

```cpp
namespace mio
{

tflite::BuiltinOperator builtin_code_neutral(const tflite::OperatorCode *opcode);
// ...

} // namespace mio

mio::tflite::builtin_code_neutral(opcode); // COMPILE FAILED
mio::builtin_code_neutral(opcode); // COMPILE & RUNNING SUCCESS
```

---

FYI, original trial was following :(
```cpp
namespace mio
{
namespace tflite
{

tflite::BuiltinOperator builtin_code_neutral(const tflite::OperatorCode *opcode);
// ...

} // namespace tflite
} // namespace mio
```

```sh
Helper.h:27:9: error: ‘BuiltinOperator’ in namespace ‘mio::tflite’ does not name a type
 tflite::BuiltinOperator builtin_code_neutral(const tflite::OperatorCode *opcode);
         ^~~~~~~~~~~~~~~
...
```
`tflite::BuiltinOperator` is interpreted as `mio::tflite::BuiltinOperator`

seanshpark: > It failed because tflite::BuiltinOperator is interpreted as mio::tflite::BuiltinOperator

```cxx
namespace mio
{
namespace tflite
{

::tflite::BuiltinOperator builtin_code_neutral(const tflite::OperatorCode *opcode);
// ...

} // namespace tflite
} 
```
maybe try with `::tflite::BuiltinOperator` instead of `tflite::BuiltinOperator`

anyway, using `mio` instrad of `tflite` looks OK.

one thing, if you are to revise `mio-circle` too, then `mio::tfltie` and `mio::circle` would look better.

seanshpark (approved):
LGTM

seanshpark:
OK to merge?

llFreetimell:
> OK to merge? 

Yes please :)

compiler/mio-tflite260/include/mio_tflite260/Util.h
llFreetimell: > maybe try with ::tflite::BuiltinOperator instead of tflite::BuiltinOperator

I think it would work but not sure how much should I change those things :)
So as you advised, when I try to revise `mio-circle` later, I will think together this!

PR #8083, opened 2021-12-02
[luci-interpreter] Remove redundant field

binarman:
This PR removes _memory_manager field, because it is unused.


BalyshevArtem (approved):
LGTM

Issue #7535, opened 2021-08-19
[luci-interpreter] Add tests for MirrorPad


SlavikMIPT:
## Goal
Introduce tests for MirrorPad
## Rationale
Provide test coverage for all kernels, there is no MirrorPad.test.cpp in compiler/luci-interpreter/src/kernels at this moment

ai-moiseev:
#8085 introduces value tests for MirrorPad (`REFLECT` and `SYMMETRIC`).

seanshpark:
close as done

Issue #6856, opened 2021-05-19
[luci-interpreter] Support MirrorPad kernel


ai-moiseev:
## What

Support for MirrorPad Op

## Why

Need to quantize models with MirrorPad Op.

## Steps

#6852 adds initial support which is only enough to support single-op models and models with MirrorPad near the output. Complete inference support is needed.

* [x] luci-interpreter #8085
* [x] circle-quantizer #6853
* [x] coverage tests #6871

PR #8085, opened 2021-12-02
[luci-interpreter] Implement MirrorPad kernel

ai-moiseev:
This commit adds valid implementation of MirrorPad kernel based on reference PadV2 implementation.

Issue: #6856 
Related: #7535 

### DIFF
--- a/compiler/luci-interpreter/src/kernels/MirrorPad.cpp
+++ b/compiler/luci-interpreter/src/kernels/MirrorPad.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2019 The TensorFlow Authors. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,8 +18,6 @@
 #include "kernels/MirrorPad.h"
 
 #include "kernels/Utils.h"
-
-#include <tensorflow/lite/kernels/internal/reference/pad.h>
 
 namespace luci_interpreter
 {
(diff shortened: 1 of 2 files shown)


compiler/luci-interpreter/src/kernels/MirrorPad.cpp
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
ai-moiseev, line 3: @binarman PTAL. Is it enough as attribution for Apache 2.0 code?

binarman: Yes, it should be ok

compiler/luci-interpreter/src/kernels/MirrorPad.cpp
```diff
@@ -59,50 +58,113 @@ void MirrorPad::configure()
...
+  const T *in_ptr = input_data;
+  T *out_ptr = output_data;
+
+  for (int32_t b = 0; b < output_b; ++b)
+  {
+    for (int32_t h = 0; h < output_h; ++h)
+    {
+      for (int32_t w = 0; w < output_w; ++w)
+      {
+        for (int32_t d = 0; d < output_d; ++d)
```
binarman, line 139: If input shape is not 4d this nested loop will not execute even once.
Do we need complex logic with with index computation above?

Maybe add simple check that shape is 4d and we are good to go?

compiler/luci-interpreter/src/kernels/MirrorPad.cpp
binarman: Could you point where this code is located in TF sources?

I can not find it, so I am not sure we need this line.

compiler/luci-interpreter/src/kernels/MirrorPad.cpp
ai-moiseev: Thanks!
You're probably right, limiting support only to 4d would simplify implementation a bit... but the fix for lower dimension is kinda simple, I've also added lower dimension count testcase. If you think support for <4d is not needed, I'll remove it :)

compiler/luci-interpreter/src/kernels/MirrorPad.cpp
ai-moiseev: It's in [``tensorflow/lite/kernels/internal/reference/pad.h``](https://github.com/tensorflow/tensorflow/blob/919f693420e35d00c8d0a42100837ae3718f7927/tensorflow/lite/kernels/internal/reference/pad.h#L38), modified in extended (it supported up to 5d, I've limited support to 4).

binarman (approved):
LGTM

m-bronnikov (approved):
LGTM

PR #8091, opened 2021-12-06
[mio-tflite260] Add mio::tflite namespace

llFreetimell:
In case of `mio-circle` is updated, it would be better to introduce `mio::tflite` namespace.
For this, `tflite::*` for `schema_generated.h` is changed to `::tflite::*`.

---

Parent Issue :#8050
Draft : #8051
Inspired by, and thanks to https://github.com/Samsung/ONE/pull/8082#discussion_r762662258


seanshpark (approved):
LGTM

PR #8097, opened 2021-12-06
[Draft][Compiler] Introcuce helper functions for reading circle files

llFreetimell:
This draft is for introducing helper functions for reading circle files.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

---

Parent Issue : #8096

Issue #8096, opened 2021-12-06
[Compiler] Introduce utils for reading circle file


llFreetimell:
### What?

Let's introduce utils for reading circle files.
(e.g. Checking custom, etc)

### Why?

Currently, we have some difficulties such as following.

- Some modules have duplicated codes to read circle file
- When we need to fix something related to circle, we should modify all of the codes
- When circle schema is updated just like TFLite, we should update all of the codes

By introducing common utility functions by wrapping circle flatbuffer functions, we may expect following things.

- Reduce duplicated codes
- When can provide utility functions according to each version of circle
- When circle schema is updated or when we need to fix something, we may modify codes only once.

### Related PRs

- [x] Introduce helper functions to `mio-circle`
  - #8098
- [x] Replace helper functions in `circlechef`
  - #8104
- [x] Replace helper functions in `circledump`
  - #8105
- [x] Replace helper functions in `circle-inspect`
  - #8106
- [x] Replace helper functions in `circle-tensordump`
  - #8103

#### Draft

- #8097

llFreetimell:
Done!

PR #8098, opened 2021-12-06
[mio-circle] Introduce utility functions for reading circle file

llFreetimell:
By introducing utility functions for reading circle file, we may expect following things.
- Reduce duplicated codes
- Reduce code modifications When circle schema is updated

---

Parent issue : #8096 
Draft : #8097


seanshpark (approved):
LGTM

Issue #8089, opened 2021-12-03
[circle-execution-plan] Add execution order estimator


BalyshevArtem:
## Goal
Add execution order estimator to ExecutionPlan. This estimator iterates through a certain set of possible traversals of the graph and selects the best option that requires the least amount of memory for intermediate tensors among all the considered traversals.

## Why
Now ExecutionPlan provides only default order of execution of graph nodes. But there are situations when we can find such an execution order in which the necessary memory for the intermediate tensors of the network will decrease. This can help save memory during network execution, which is very necessary for devices with very small RAM, such as microcontrollers.

**Example**

To illustrate this, figure below shows an example of a computation graph. Evaluating operators as numbered 1 through to 7 will result in peak memory usage of 5216, coming from operator №3 (input and output buffers + the output of operator №1 that is held back for operator №4). However, fully evaluating the rightmost branch first (execution order 1, 4, 6, 2, 3, 5, 7), would result in peak memory usage of 4960, coming from operator №2 (input and output buffers + output of operator №6 that is held back for operator №7).

![imgonline-com-ua-Resize-lTp79MdjPLxW](https://user-images.githubusercontent.com/43214667/144586731-9c25db8f-4f9d-43bd-ba04-08576a98470f.jpg)

BalyshevArtem:
+cc @binarman

BalyshevArtem:
You can find more information about this approach in https://arxiv.org/pdf/1910.05110.pdf

PR #8099, opened 2021-12-06
[Draft][circle-execution-plan] Add execution order estimator

BalyshevArtem:
This commit adds execution order estimator to ExecutionPlanner.

issue #8089 


compiler/circle-execution-plan/src/ExecutionOrderEstimator.cpp
```diff
@@ -0,0 +1,249 @@
...
+    case luci::CircleOpcode::CIRCLEWHILEOUT:
+      return false;
+    default:
+      return true;
+  }
+}
+
+bool NextSet(std::vector<uint32_t> &vector)
+{
+  int n = vector.size();
```
binarman, line 65: In case we try to iterate over empty vector
```suggestion
  int n = vector.size();
  if (n == 0)
    return false;
```

compiler/circle-execution-plan/src/ExecutionOrderEstimator.cpp
```diff
@@ -0,0 +1,249 @@
...
+    case luci::CircleOpcode::CIRCLESPLITOUT:
+    case luci::CircleOpcode::CIRCLEUNPACKOUT:
+    case luci::CircleOpcode::CIRCLEWHILEOUT:
+      return false;
+    default:
+      return true;
+  }
+}
+
+bool NextSet(std::vector<uint32_t> &vector)
```
binarman, line 63: I think this is better name for this function.

I also suggest to add some comment on what is the expected input and what is the expected output.

For example, I am not sure it will work correctly with vectors containing duplicated values.

```suggestion
bool nextPermutation(std::vector<uint32_t> &vector)
```

compiler/circle-execution-plan/src/ExecutionOrderEstimator.cpp
```diff
@@ -0,0 +1,249 @@
...
+      j--;
+
+    if (j == -1)
+      return false;
+
+    int k = n - 1;
+    while (vector[j] >= vector[k])
+      k--;
+
+    std::iter_swap(vector.begin() + j, vector.begin() + k);
```
binarman, line 80: You can use simple
```suggestion
    std::swap(vector[j], vector[k]);
```

compiler/circle-execution-plan/src/ExecutionOrderEstimator.cpp
```diff
@@ -0,0 +1,249 @@
...
+    int l = j + 1, r = n - 1;
+
+    while (l < r)
+      std::iter_swap(vector.begin() + (l++), vector.begin() + (r--));
```
binarman, line 84: ditto

compiler/circle-execution-plan/src/ExecutionOrderEstimator.cpp
```diff
@@ -0,0 +1,249 @@
...
+      return true;
+  }
+}
+
+bool NextSet(std::vector<uint32_t> &vector)
+{
+  int n = vector.size();
+  int j;
+  do
+  {
```
binarman, line 68: This loop is useless, because `j = n - 2` and then it always decreased, so continue condition `j > n - 1` is always false.

compiler/circle-execution-plan/src/ExecutionOrderEstimator.cpp
```diff
@@ -0,0 +1,249 @@
...
+      // The following nodes denote outputs of multiple-output nodes.
+    case luci::CircleOpcode::CIRCLEIFOUT:
+    case luci::CircleOpcode::CIRCLESPLITOUT:
+    case luci::CircleOpcode::CIRCLEUNPACKOUT:
+    case luci::CircleOpcode::CIRCLEWHILEOUT:
+      return false;
+    default:
+      return true;
+  }
+}
```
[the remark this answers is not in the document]
BalyshevArtem: done

compiler/circle-execution-plan/src/ExecutionOrderEstimator.cpp
BalyshevArtem: done

compiler/circle-execution-plan/src/ExecutionPlanner.cpp
```diff
@@ -116,9 +82,75 @@ uint32_t compute_im2col_size(const luci::CircleConv2D *conv)
...
+        if (_required_size < best_required_size)
+        {
+          best_required_size = _required_size;
+          best_execution_order = _ordered_nodes;
+        }
+
+        _ordered_nodes.clear();
+        _alloc_node_inform_vector.clear();
+        _alloc_node.clear();
+        _dealloc_node.clear();
```
binarman, line 126: optional: Move this to separate method, something like "clean_state", so we can call it from other places to reset planning state during some search algorithm.

Alternative way is to move memory planning stuff (including `make_execution_plan` method) into separate class.

compiler/circle-execution-plan/src/ExecutionOrderEstimator.cpp
```diff
@@ -0,0 +1,235 @@
...
+  const std::vector<NodeOrderInform> &node_order_inform_vector)
+{
+  std::vector<loco::Node *> res;
+
+  std::set<loco::Node *> visited_nodes;
+  std::stack<Frame> frames;
+
+  auto visited = [&visited_nodes](loco::Node *node) {
+    return visited_nodes.find(node) != visited_nodes.end();
+  };
```
binarman, line 176: Maybe move this in separate function?
Also could make it MCST style (check + add in set of visited) =D

compiler/circle-execution-plan/src/ExecutionOrderEstimator.cpp
```diff
@@ -0,0 +1,235 @@
...
+            combination.emplace_back(s);
+          it->all_combinations.emplace_back(combination);
+        } while (nextPermutation(it->executable_arg_idx));
+      }
+      frames.pop();
+    }
+  }
+}
+
+std::vector<loco::Node *> ExecutionOrderEstimator::get_current_order(
```
binarman, line 166: Q) I've just read the rest of code and noticed that it not producing sequences like this:
order

Did I understand this algorithm right?

BalyshevArtem: No, the sequence for the graph you mentioned above could be: 
- 1, 2, 4, 3, 5, 6
- 1, 3, 5, 2, 4, 6
We only estimate the order of branches

binarman: Then, I want to take my words back =)
We need to name this search algorithm different. "Exhaustive search" implies that it checks all possible sequences

binarman:
Overall, I think this code is fine, but need to add more comments near main functions.

I also think we need to break this code in smaller more structured functions, but this could be done later, after we cover it with some tests.

PR #8109, opened 2021-12-07
[circle-execution-plan] Add execution plan for multi-output nodes

BalyshevArtem:
This pr adds to ExecutionPlanner possibility to find execution plan for nodes with multi-output nodes. Now ExecutionPlanner can find offsets for nodes with multi-output nodes.

ONE-DCO-1.0-Signed-off-by: Artem Balyshev a.balyshev@partner.samsung.com

PR #8110, opened 2021-12-07
[circle-execution-plan] Add tests to ExecutionPlanner for multi-outputs nodes

BalyshevArtem:
This commit adds to ExecutionPlanner tests for multi-outputs nodes.

Merge it after #8109 

### DIFF
--- a/compiler/circle-execution-plan/src/ExecutionPlanner.test.cpp
+++ b/compiler/circle-execution-plan/src/ExecutionPlanner.test.cpp
@@ -1,0 +1,175 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <loco.h>
+#include "ExecutionPlanner.h"
+
+#include <luci/IR/CircleNodes.h>
+
+#include <gtest/gtest.h>
+
+TEST(ExecutionPlannerTest, op_with_multiple_outputs)
+{
+  auto g = loco::make_graph();
+
+  // Create nodes
+  auto input = g->nodes()->create<luci::CircleInput>();
+  auto split = g->nodes()->create<luci::CircleSplit>();
+  auto split_out1 = g->nodes()->create<luci::CircleSplitOut>();
+  auto split_out2 = g->nodes()->create<luci::CircleSplitOut>();
+  auto split_dim = g->nodes()->create<luci::CircleConst>();
+
+  // Build up initial graph
+  auto graph_input1 = g->inputs()->create();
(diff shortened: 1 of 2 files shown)


compiler/circle-execution-plan/src/ExecutionPlanner.test.cpp
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <loco.h>
+#include "ExecutionPlanner.h"
```
binarman, line 18: nitpick:
- maybe swap these includes?
- do you need to include loco.h in the first place?

compiler/circle-execution-plan/src/ExecutionPlanner.test.cpp
```diff
@@ -0,0 +1,175 @@
...
+  split_out2->shape({1, 3});
+  split_out2->dtype(split->dtype());
+  split_out2->name("split_out2");
+
+  auto add = g->nodes()->create<luci::CircleAdd>();
+  add->name("add");
+  add->x(split_out1);
+  add->y(split_out2);
+  add->dtype(split_out1->dtype());
+  add->shape({1, 3});
```
binarman, line 75: Note: if you are going to make larger test graphs, maybe it is easier to set only input shaped and dtypes, then run type and shape inference.

compiler/circle-execution-plan/src/ExecutionPlanner.test.cpp
```diff
@@ -0,0 +1,175 @@
...
+  // Create ExecutionPlanner
+  circle_planner::ExecutionPlanner execution_planner(g.get());
+  execution_planner.make_execution_plan();
+
+  // Check result
+  auto output1_exec_plan = luci::get_execution_plan(output1);
+  auto output2_exec_plan = luci::get_execution_plan(output2);
+
+  ASSERT_TRUE(output1_exec_plan.offsets().front() != output2_exec_plan.offsets().front());
+  ASSERT_TRUE(output1_exec_plan.offsets().front() + output1->dim(1).value() <=
```
binarman, line 171: I don't get it. What does this check looking for?

P.s. this is also confusing: `offset` is measured in bytes, `dim(1).value` in number of elements

BalyshevArtem: Yes, this is typo. I actually would like to check that offsets of the graph output nodes do not intersect. I will rewrite it

BalyshevArtem:
@binarman,
I think, that it is better to close this pr now, in order to wait for all the pr associated with CircleExecutionPlan (#8128 #8099) to  merge and then write tests covering the entire ExecutionPlanner.
What do you think about this?

binarman:
> I think, that it is better to close this pr now, in order to wait for all the pr associated with CircleExecutionPlan 

Do you plan to change planner interfaces dramatically?

If not, I don't see why we have to postpone this particular PR

compiler/circle-execution-plan/src/ExecutionPlanner.test.cpp
```diff
@@ -0,0 +1,175 @@
...
+              output2_exec_plan.offsets().front() + output2->dim(1).value());
+
+  SUCCEED();
+}
```
seanshpark, line 175: Please add same or more negative tests with `_NEG` ending at the name.
Our SE requires to provide more negative tests and currently numbers are at boundary.

@binarman , please check tests codes for this in `circle-execition-plan` and `luci-interpreter` tests which I don't fully review.

Issue #7627, opened 2021-09-08
[luci] provide public copy_quantparam()


seanshpark:
As `copy_quantparam()` is used multiple times, let's provide this as public method, maybe in luci/lang or luci/service

And revise codes to use public method.

seanshpark:
Like in #7623

seanshpark:
done

PR #8112, opened 2021-12-08
DRAFT: copy_quantparam

seanshpark:
on-going draft to provide copy_quantparam in luci/lang.


seanshpark:
For #7627

PR #8116, opened 2021-12-08
[one-cmds] Legalize onnx models

binarman:
This PR introduces onnx legalizer that replaces
unsupported operations with equivalent subgraphs.

related issue: #7939 

### DIFF
--- a/compiler/one-cmds/onnx_legalizer.py
+++ b/compiler/one-cmds/onnx_legalizer.py
@@ -1,0 +1,802 @@
+#!/usr/bin/python3
+import onnx
+import onnx.numpy_helper
+import sys
+import numpy as np
+import re
+
+# Transform onnx model to make it compilable with our toolchain
+# List of transformations:
+# - Replace RNN operation with unrolled subgraph
+# - Replace LSTM operation with unrolled subgraph
+
+
+def reverse_str(s):
+    return ''.join(reversed(s))
+
+
+def parse_tensor_name(name):
+    rev = reverse_str(name)
+    m = re.match('(\d*)(.*)', rev)
+    if m.groups()[0] != '':
+        return (reverse_str(m.groups()[1]), int(reverse_str(m.groups()[0])))
+    else:
+        return (reverse_str(m.groups()[1]), 0)
+
+
+class ModelTransformerHelper:
+    def __init__(self, model):
+        self._model = model
+        self._nodes_to_delete = []
+        self._insert_id = 0
+        # each tensor has name containing base name and unique number. for example:
+        # "abc_123": "abs_" - base name, "123" - unique number
+        # if no number in name, consider it is equal to "0"
+
+        # mapping from base names to largest given number
+        self._base_name_idx = {}
+        # gather name information for existing tensors
+        for node in model.graph.node:
+            for t in list(node.input) + list(node.output):
+                base_name, number = parse_tensor_name(t)
+                if base_name in self._base_name_idx:
+                    self._base_name_idx[base_name] = max(self._base_name_idx[base_name],
+                                                         number)
+                else:
+                    self._base_name_idx[base_name] = number
+
+    def make_tensor_with_base_name(self, base_name):
+        if base_name in self._base_name_idx:
+            self._base_name_idx[base_name] += 1
+            return base_name + str(self._base_name_idx[base_name])
+        else:
+            self._base_name_idx[base_name] = 0
+            return base_name + '0'
+
+    def make_node(self, opcode, inputs, outputs, *p_args, **k_args):
+        """
+    Create node and it in graph.
+ 
+    Parameters:
+        opcode (str): opcode name of desired operation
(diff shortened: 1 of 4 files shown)


binarman:
Idea of this utility is to replace problematic onnx operations with something that we can convert.
For example, RNN:

Before | After
--- | ---
image | image

binarman:
This kind of transformation allows one-cmds to convert RNN and LSTM onnx/pytorch nodes.

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -0,0 +1,749 @@
...
+import re
+
+# Transform onnx model to make it compilable with our toolchain
+# List of transformations:
+# - Replace RNN operation with unrolled subgraph
+# - Replace LSTM operation with unrolled subgraph
+
+
+def reverse_str(s):
+    return ''.join(reversed(s))
```
mhs4670go, line 15: I think you already know about this. `s[::-1]` can do the same. But, what you implemented is more readable 👍

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -0,0 +1,749 @@
...
+
+
+def parse_tensor_name(name):
```
mhs4670go, line 18: It has nothing to do with this PR. Just like C++ private attribute, I've written methods whose name starts with "_". But, it turns out that underscore seems unnecessary and doesn't look good to me. So, I'm gonna post a PR that removes them. What is your opinion about this?

compiler/one-cmds/onnx_legalizer.py
binarman: actually, I did not know that it is applicable to strings =)

compiler/one-cmds/onnx_legalizer.py
binarman: > underscore seems unnecessary and doesn't look good to me

As far as I know, underscore in function names affects `imports * from module` expressions.
Sooo, maybe I need to add underscores to some functions here =)

P.s. I personally think that this code needs some refactoring, but deadlines are near, so published it as-is =_=
Sorry, if it is hard to read...

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -0,0 +1,749 @@
+#!/usr/bin/python3
+import onnx
```
seanshpark, line 2: maybe add copyright notice?

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -0,0 +1,749 @@
...
+    return ''.join(reversed(s))
+
+
+def parse_tensor_name(name):
+    rev = reverse_str(name)
+    m = re.match('(\d*)(.*)', rev)
+    if m.groups()[0] != '':
+        return (reverse_str(m.groups()[1]), int(reverse_str(m.groups()[0])))
+    else:
+        return (reverse_str(m.groups()[1]), 0)
```
mhs4670go, line 24: @binarman How about just do like this rather than reversing strings?
```python
m = re.findall('\D+|\d*', 'abc_123')
```

Is there any tricky case?

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -0,0 +1,749 @@
...
+        if base_name in self._base_name_idx:
+            self._base_name_idx[base_name] += 1
+            return base_name + str(self._base_name_idx[base_name])
+        else:
+            self._base_name_idx[base_name] = 0
+            return base_name + '0'
+
+    def make_node(self, opcode, inputs, outputs, *p_args, **k_args):
+        """
+    Create node and it in graph.
```
mhs4670go, line 58: `Create node and it(?) in graph` Maybe some mistake??

mhs4670go:
> I personally think that this code needs some refactoring, but deadlines are near, so published it as-is =_=
Sorry, if it is hard to read...

As you said, it is a bit hard to read but looks kinda good to me. It could be refactored later:)

compiler/one-cmds/onnx_legalizer.py
binarman: Name in onnx is an arbitrary string, so "123" and "abc_321_cba_123" are valid.
This base name and id convention is my attempt to make new tensor names without risk of name collision.

Your approach do the trick, but requires some additional logic to cut only last number anyway.

If you think it is better option, I can replace reverse string approach with findall. reverse is just a first solution that came to my mind =)

compiler/one-cmds/onnx_legalizer.py
binarman: yes, should be "Create node and insert it in graph."

binarman:
@seanshpark @mhs4670go 
Ping.
Do you have more comments?

mhs4670go (approved):
Thank you!! LGTM:)

seanshpark (approved):
LGTM

PR #8121, opened 2021-12-08
[luci/IR] Expose copy_quantparam method

seanshpark:
This will expose copy_quantparam as IR method.


seanshpark:
For #7627, from draft #8112

jinevening (approved):
LGTM

PR #8122, opened 2021-12-08
[luci/pass] Use copy_quantparam

seanshpark:
This will restore using copy_quantparam method in pass.


seanshpark:
For #7627, from draft #8112

jinevening (approved):
LGTM

PR #8128, opened 2021-12-08
[Draft][circle-execution-plan] Add Pal to ExecutionPlanner

BalyshevArtem:
This pr adds pal interface for different platforms to correctly calculation of the size of the scratchpad.

### DIFF
--- a/compiler/circle-execution-plan/CMakeLists.txt
+++ b/compiler/circle-execution-plan/CMakeLists.txt
@@ -1,16 +1,3 @@
-set(SOURCES
-        src/CircleExecutionPlan.cpp
-        src/ExecutionPlanner.cpp
-        src/ExecutionPlanner.h
-        )
+set(CIRCLE_EXECUTION_PLAN_PAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/pal")
 
-add_executable(circle_execution_plan "${SOURCES}")
-target_link_libraries(circle_execution_plan foder)
-target_link_libraries(circle_execution_plan safemain)
-target_link_libraries(circle_execution_plan luci_env)
-target_link_libraries(circle_execution_plan luci_import)
-target_link_libraries(circle_execution_plan luci_export)
-target_link_libraries(circle_execution_plan luci_plan)
-target_link_libraries(circle_execution_plan arser)
-
-install(TARGETS circle_execution_plan DESTINATION bin)
+add_subdirectory(src)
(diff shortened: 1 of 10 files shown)


binarman:
I think this could be done by option, not PAL.
We used PAL in luci-interpreter, because:
1. It is impossible to actually build some pieces of code for all platforms (for example pthreads for MCU or ARM intrinsics for x86)
2. We wanted to save build time and runtime resources.

compiler/circle-execution-plan/src/CircleExecutionPlan.cpp
```diff
@@ -35,6 +35,18 @@ int entry(int argc, char **argv)
...
+    .type(arser::DataType::STR)
+    .required(false)
+    .default_value("linux")
+    .help("Platform name: linux mcu cmsisnn");
+  arser.add_argument("--use_dsp")
+    .nargs(1)
+    .type(arser::DataType::BOOL)
+    .required(false)
+    .default_value(false)
+    .help("Plan with or without dsp");
```
binarman, line 49: I think it worth to mention that it applied to cmsis only for now.

compiler/circle-execution-plan/src/CircleExecutionPlan.cpp
```diff
@@ -82,7 +115,7 @@ int entry(int argc, char **argv)
   auto module = importer.importModule(circle_model);
 
   // Do main job
-  circle_planner::ExecutionPlanner execution_planner(module->graph());
+  circle_planner::ExecutionPlanner execution_planner(module->graph(), platform_type, use_dsp);
```
binarman, line 118: What about make "TargetPlatform" structure/class and add `platform_type` and `use_dsp` there?
I think at some point we will need even more options, so this function definition will grow...

binarman (commented):
Overall looks good to me

compiler/circle-execution-plan/src/CircleExecutionPlan.cpp
BalyshevArtem: Done

compiler/circle-execution-plan/CMakeLists.txt
```diff
@@ -1,16 +1,3 @@
-set(SOURCES
```
binarman, line 1: Maybe merge this and cmake file from src dir?

BalyshevArtem: Did I understand you correctly, that you are suggesting to use only this (`compiler/circle-execution-plan/CMakeLists.txt`) cmakeList and don't create new `compiler/circle-execution-plan/src/CMakeLists.txt`?

BalyshevArtem:
everything has been added, close this draft

PR #8129, opened 2021-12-08
[luci/svc] Use copy_quantparam in copy_common_attriburtes

seanshpark:
This will revise copy_common_attributes method to use copy_quantparam method.


seanshpark:
For #7627, from draft #8112

jinevening (approved):
LGTM

PR #8133, opened 2021-12-09
[luci/pass] Fix src,dst of copy_quantparam

seanshpark:
This will fix src, dst argument with calling copy_quantparam method.


seanshpark:
To fix wrong change of #8122
- source, destination was swapped for new `copy_quantparam`

jinevening (approved):
LGTM

Issue #8079, opened 2021-12-02
[circle-quantizer] Support copy of qparam between tensors in circle-quantizer


jinevening:
## What

Let's support copy qparam feature in circle-quantizer

## Why

> @jinevening , although the simplest way, this method was not preferred because it affects all MaxPool operations of the model. Let's think about a little more. :)

Before I make this issue, I've checked that this option only affects the tensor what we want in our target model. Other MaxPool Ops appear after `Conv2D+Relu` Ops. Relu before MaxPool makes MaxPool Ops have the same input/output qparam, so this option does not affect those MaxPools.

I agree that copying qparam is a more powerful (general) option. My suggestion is to support TF-style MaxPool for our target model (for ease of integration with our GUI tool and better UX), and also support the copy feature for future usage.

_Originally posted by @jinevening in https://github.com/Samsung/ONE/issues/8055#issuecomment-981376936_

jinevening:
I'm thinking about the interface as follows.

```
circle-quantizer --copy_qparam A B C D
```

```
one-quantize --copy_qparam --src_tensor_name A --dst_tensor_name B --input_path C --output_path D
```

A: source tensor name
B: destination tensor name
C: input path
D: output path

Qparam of A is copied to B.

jinevening:
This work would be related to https://github.com/Samsung/ONE/pull/8115.

seanshpark:
> This work would be related to #8115.

You can add or use `copy_quantparam()`, if you don't want to wait my task to finish.
I'll revise new codes later.

jinevening:
All done. CC @lemmaa

PR #8134, opened 2021-12-09
[circle-quantizer] Support copy qparam option

jinevening:
This supports copy qparam option.

---
Related to: #8079
Draft PR: #8123

### DIFF
--- a/compiler/circle-quantizer/src/CircleQuantizer.cpp
+++ b/compiler/circle-quantizer/src/CircleQuantizer.cpp
@@ -65,6 +65,7 @@ int entry(int argc, char **argv)
   const std::string qwmm = "--quantize_with_minmax";
   const std::string rq = "--requantize";
   const std::string fq = "--force_quantparam";
+  const std::string cq = "--copy_quantparam";
 
   const std::string tf_maxpool = "--TF-style_maxpool";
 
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -125,6 +126,15 @@ int entry(int argc, char **argv)
           "Three arguments required: tensor_name(string), "
           "scale(float) zero_point(int)");
 
+  arser.add_argument(cq)
+    .nargs(2)
+    .type(arser::DataType::STR_VEC)
+    .required(false)
+    .accumulated(true)
+    .help("Copy quantization parameter from a tensor to another tensor."
```
seanshpark, line 134: Do we usually have several tensors to apply this or can there be like hundreds of tensors?
If there are lots of tensor pairs, we can provide an external text file with form of 'a tensor, another tensor' 
```
t1,t11
t2,t21,
...
tn,tn1
```

jinevening: Currently our use case is copying the qparam of a single tensor. I think the number of target tensors will not be too large, because fort that case it would be better to requantize the model.

I agree that the text file interface is useful for designating many tensors. Let's revisit the interface when we receive a new request from our customer.

PR #8144, opened 2021-12-09
[one-cmds] Revise version fix of pip, setuptools

seanshpark:
This will revise one-cmds-prepare to use latest version of pip and setuptools.


mhs4670go (approved):
LGTM

PR #8154, opened 2021-12-10
[DRAFT] [luci-interpreter] Improve While operation

binarman:
Fix memory allocations in while operation, add support for some arithmetic and comparision operations

### DIFF
--- a/compiler/luci-interpreter/src/core/RuntimeGraph.cpp
+++ b/compiler/luci-interpreter/src/core/RuntimeGraph.cpp
@@ -138,6 +138,8 @@ void RuntimeGraph::setOutputTensors(const std::vector<Tensor *> &output_tensors)
 
 void RuntimeGraph::configureAllocations(Tensor *tensor)
 {
+  if (tensor->is_data_allocated())
+    _memory_manager->release_memory(*tensor);
   _memory_manager->allocate_memory(*tensor);
 }
 
--- a/compiler/luci-interpreter/src/core/RuntimeGraph.cpp
+++ b/compiler/luci-interpreter/src/core/RuntimeGraph.cpp
@@ -138,8 +138,6 @@ void RuntimeGraph::setOutputTensors(const std::vector<Tensor *> &output_tensors)
 
 void RuntimeGraph::configureAllocations(Tensor *tensor)
 {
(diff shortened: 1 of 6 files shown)


compiler/luci-interpreter/src/kernels/Add.cpp
```diff
@@ -54,6 +54,12 @@ void Add::execute() const
     case DataType::FLOAT32:
       evalFloat();
       break;
+    case DataType::S64:
+      evalInteger<int64_t>();
+      break;
+    case DataType::S32:
+      evalInteger<int32_t>();
+      break;
```
seanshpark, line 62: 👍🏻

compiler/luci-interpreter/src/kernels/Equal.cpp
```diff
@@ -49,6 +49,19 @@ void Equal::execute() const
...
+      evalInteger<int64_t>();
+      break;
+    case DataType::S32:
+      evalInteger<int32_t>();
+      break;
+    case DataType::U64:
+      evalInteger<uint64_t>();
+      break;
+    case DataType::U32:
+      evalInteger<int32_t>();
```
seanshpark, line 62: maybe `uint32_t`

binarman: I skipped this case when made #8157
I'll add it as a separate change

binarman:
@seanshpark FYI explanation what this PR is doing
In short: these changes are needed to properly allocate buffers for while input/output tensors.

**Before we introduced memory managers**

Actual memory allocations were done automatically and everything worked.

**After we introduced memory managers**

Allocations should be done manually by calling to memory allocator.

We need to **reshape** tensor first, then **allocate** memory for it.
But in current master interpreter tries to **allocate** first, then **reshape** in `while` kernel...

P.s. `While` is a special operations, memory for most other kernels are managed by `RuntimeGraph` automatically.

BalyshevArtem (approved):
LGTM

seanshpark:
@binarman , thanks for adding explanation!
Couldn't fully understand the changes but I assume you wrote it correctly :)

seanshpark:

Issue #8155, opened 2021-12-11
[luci-interpreter] Support integer arithmetics and comparisons


binarman:
## Problem
Sometimes we need integer computations, for example for shape manipulations or complex operations, like LSTM, MaxPoolWithArgmax, etc.

## Goal

- [x] Add #8156 
- [x] Sub #8179 
- [x] Mul #8176 
- [x] Div #8180
- [x] Equal #8157
- [x] NotEqual #8158 
- [x] Less #8160 
- [x] Greated #8166 
- [x] LessEqual #8159 
- [x] GreaterEqual #8167 

It is low priority, but IMHO, good to have

binarman:
Everything is merged!
Closing

PR #8157, opened 2021-12-11
[luci-interpreter] Support Integer Equal

binarman:
This PR supports S32 and S64 Equal comparison.

Related issue: #8155 


m-bronnikov (approved):
LGTM

m-bronnikov:
Note: I think we could use TYPED_TEST for `int32_t, int64_t, float`  to reduce test code amount. But better to introduce this changes in separated PR to simplify review process =)

BalyshevArtem (approved):
LGTM

Issue #8065, opened 2021-11-29
[circle-quantizer] Support OneHot's quantization


m-bronnikov:
### What?

Let's support qunatization of OneHot and qunatize following model:

![image](https://user-images.githubusercontent.com/43731397/143946965-84453edc-209e-4fba-a0d4-8da9a8536af2.png)

<details>
<summary> Code to obtain model </summary>

```python
import tensorflow as tf
import numpy as np

depth=100

with tf.Graph().as_default() as graph:
    # create model
    input_shape = (4, )
    ifm = tf.compat.v1.placeholder(tf.int32, shape = input_shape)
    ofm = tf.one_hot(ifm, depth=depth)

    # save to tflite
    with tf.compat.v1.Session() as sess:
        sess.run(tf.compat.v1.global_variables_initializer())
        converter = tf.compat.v1.lite.TFLiteConverter.from_session(sess, [ifm], [ofm])
        converter.allow_custom_ops = True
        content = converter.convert()
        with open("one_hot.tflite", 'wb') as binf:
            binf.write(content)
```

</details>

### Why?

To increase models coverage which could be quantized using *circle-quantizer*.

m-bronnikov:
To support qunatization of OneHot in *circle-quantizer* we need to resolve #8064 first.

m-bronnikov:
+cc @jinevening

m-bronnikov:
@jinevening I would like to ask your suggestion!

OneHot op expects int32 non-constant input. How we could quantize this input?
AFAIK `record-minmax` expects float32 input and `circle-quantizer` also not allowed to handle these dtype of input. =(

jinevening:
> OneHot op expects int32 (or int64) non-constant input. How we could quantize this input?

Such an input is an index, so it should not be quantized. For example, we do not quantize ArgMax as its output is index.

> AFAIK record-minmax expects float32 input and circle-quantizer also not allowed to handle these dtype of input. =(

As to record-minmax, we may have to support non-float type input. I think MinMaxObserver should be changed as below.

MinMaxObserver.cpp
```cpp
  // Only support recording of float32 values
  if (tensor->element_type() != DataType::FLOAT32)
  {
    // Exceptions that should be processed in backends
    switch (node->opcode())
    {
      case luci::CircleOpcode::ARG_MAX:
        // Output of arg_max is the index of the largest value across axes of a tensor.
        // It always has integer type.
      case luci::CircleOpcode::CAST:
[...]

PR #8171, opened 2021-12-14
[draft][circle-quantizer] Support OneHot quantization

m-bronnikov:
This draft enables quantization of models with OneHot ops.

----------------

For: #8065
+cc @jinevening

### DIFF
--- a/compiler/record-minmax/src/MinMaxObserver.cpp
+++ b/compiler/record-minmax/src/MinMaxObserver.cpp
@@ -51,6 +51,16 @@ void MinMaxObserver::postTensorWrite(const luci::CircleNode *node,
     // Bool type tensor is not quantized
     return;
   }
+  if (node->dtype() == DataType::S32)
+  {
+    // Integer type tensor is not quantized
+    return;
+  }
+  if (node->dtype() == DataType::S64)
+  {
+    // Integer type tensor is not quantized
+    return;
+  }
 
   // Only support recording of float32 values
   if (tensor->element_type() != DataType::FLOAT32)
--- a/Note.md
+++ b/Note.md
@@ -1,0 +1,4 @@
+Let's support quantization for OneHot
+
+At first we need to support input with type int32.
+At second enable quantization with int32 type input.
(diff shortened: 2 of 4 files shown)


m-bronnikov:
This draft resolves #8065 (Model from "What?" quantizes well).

compiler/record-minmax/src/MinMaxObserver.cpp
```diff
@@ -51,6 +51,16 @@ void MinMaxObserver::postTensorWrite(const luci::CircleNode *node,
...
+  if (node->dtype() == DataType::S32)
+  {
+    // Integer type tensor is not quantized
+    return;
+  }
+  if (node->dtype() == DataType::S64)
+  {
+    // Integer type tensor is not quantized
+    return;
+  }
```
binarman, line 63: I see that this code is already merged, just want to ask how this code is interacts with code below?
```
if (tensor->element_type() != DataType::FLOAT32)
```

m-bronnikov: Personally, I think it's OK to skip quantization for all integer tensors and just return, when observer detects node with integer type. But as I understand now, this code makes `switch`/`case` statement bellow redundant in most cases. So, I think better to ask some advice about correctness of this approach.

@jinevening Could you please take a look? =)

binarman: > But as I understand now, this code makes switch/case statement bellow redundant in most cases

Yes, this is what attracted my attention =)

jinevening: record-minmax records min/max of fp32 feature maps. AFAIK, int32 and int64 are not used for feature maps, so I approved the change (@binarman please tell me if I missed something).

It seems that the case for ARG_MAX becomes redundant in the below switch statement. @m-bronnikov could you remove the case?

m-bronnikov:
I think this draft could be closed as already merged.

compiler/record-minmax/src/MinMaxObserver.cpp
binarman: > int32 and int64 are not used for feature maps

So far I saw integer operations only for indexing, so you are right.

PR #8172, opened 2021-12-14
[record-minmax] Add integer generator of random data

m-bronnikov:
This commit adds helper to generate integer random input data.

ONE-DCO-1.0-Signed-off-by: Maksim Bronnikov max120199@gmail.com

-------------------------

For: #8065
Draft: #8171

### DIFF
--- a/compiler/record-minmax/src/RecordMinMax.cpp
+++ b/compiler/record-minmax/src/RecordMinMax.cpp
@@ -62,10 +62,10 @@ std::vector<uint8_t> genRandomBoolData(std::mt19937 &gen, uint32_t num_elements)
   return input_data;
 }
 
-template <typename T> std::vector<T> genRandomIntData(std::mt19937 &gen, uint32_t num_elements)
+template <typename T>
+std::vector<T> genRandomIntData(std::mt19937 &gen, uint32_t num_elements, T min, T max)
 {
-  std::uniform_int_distribution<T> dist(std::numeric_limits<T>::min(),
-                                        std::numeric_limits<T>::max());
+  std::uniform_int_distribution<T> dist(min, max);
   std::vector<T> input_data(num_elements);
 
   // Write random data
--- a/compiler/record-minmax/src/RecordMinMax.cpp
+++ b/compiler/record-minmax/src/RecordMinMax.cpp
@@ -62,6 +62,21 @@ std::vector<uint8_t> genRandomBoolData(std::mt19937 &gen, uint32_t num_elements)
   return input_data;
 }
 
+template <typename T> std::vector<T> genRandomIntData(std::mt19937 &gen, uint32_t num_elements)
+{
+  std::uniform_int_distribution<T> dist(std::numeric_limits<T>::min(),
+                                        std::numeric_limits<T>::max());
+  std::vector<T> input_data(num_elements);
+
+  // Write random data
+  {
+    auto const generator = [&gen, &dist]() { return dist(gen); };
+    std::generate(begin(input_data), end(input_data), generator);
+  }
+
+  return input_data;
+}
+
 /**
  * @brief  getTensorSize will return size in bytes
  */


compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -388,28 +403,45 @@ void RecordMinMax::profileDataWithRandomInputs(const std::string &mode, float mi
...
+        }
+        case DataType::S64:
+        {
+          auto input_data = genRandomIntData<int64_t>(gen, num_elements);
+          _interpreter->writeInputTensor(input_node, input_data.data(),
+                                         input_data.size() * sizeof(int64_t));
+          break;
+        }
+        default:
+          throw std::runtime_error("Unsupported input type, yet!");
```
m-bronnikov, line 444: For reviewer: note this PR introduces throw of `runtime_error()` in case of unsupported input datatype and removes assert, which checks datatypes (see above).

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -388,28 +403,45 @@ void RecordMinMax::profileDataWithRandomInputs(const std::string &mode, float mi
       if (num_elements == 0)
         throw std::runtime_error("Only support non-zero sized inputs");
 
-      // TODO Support more input data types
```
m-bronnikov, line 391: Hmm, I think better to leave this comment.

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -388,28 +403,45 @@ void RecordMinMax::profileDataWithRandomInputs(const std::string &mode, float mi
...
-      assert(input_node->dtype() == loco::DataType::FLOAT32 ||
-             input_node->dtype() == loco::DataType::BOOL);
-
-      if (input_node->dtype() == DataType::FLOAT32)
+      switch (input_node->dtype())
```
seanshpark, line 406: this and below lines dones' seem to be related with the title/explanation.
please put these changes to another PR.

seanshpark:
This PR title starts with `[luci-interpreter] ` where the file is in `record-minmax`
Please correct this.

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -62,6 +62,21 @@ std::vector<uint8_t> genRandomBoolData(std::mt19937 &gen, uint32_t num_elements)
   return input_data;
 }
 
+template <typename T> std::vector<T> genRandomIntData(std::mt19937 &gen, uint32_t num_elements)
+{
+  std::uniform_int_distribution<T> dist(std::numeric_limits<T>::min(),
+                                        std::numeric_limits<T>::max());
```
jinevening, line 68: This code can generate extremely large values. Is it intended?

I don't know how the integer values will be used (maybe index?), but I think it would be better to restrict the range of generated values.

m-bronnikov: I thought about it, but I have not come up with the best way to do it (which range should be?), and for the OneHot operation, the input data (indices) can be any. So, I decided to use any integer values for input for this PR.

Do you have any suggestion to range of random input data? =)

m-bronnikov: @jinevening What do you think about [0, 127] range, which valid for any integer datatype?

jinevening: >  for the OneHot operation, the input data (indices) can be any.

For OneHot Op, I guess that the depth (range of indices) would be much smaller than the whole integer range. For example, for one-hot with depth = 1000, this random generator will make most of the output off-value.

How about using the range -100~100 (or -1000~1000)? It'll be possible to change the range if necessary. I think it would be better to take the range (min, max) as the arguments of this function.

m-bronnikov: > I think it would be better to take the range (min, max) as the arguments of this function.

I think this is a good approach.

jinevening: > What do you think about [0, 127] range, which valid for any integer datatype?

This looks also good to me :)

m-bronnikov: > > I think it would be better to take the range (min, max) as the arguments of this function.
> 
> I think this is a good approach.

I introduced 2 additional argument: min and max for generation helper. PTAL. =)

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

PR #8173, opened 2021-12-14
[record-minmax] Skip minmax recording for integer tensors

m-bronnikov:
This commit introduces skip of min/max recording for integer tensors.

-------------

For: #8065
Draft: #8171


m-bronnikov:
In https://github.com/Samsung/ONE/issues/8065#issuecomment-984172591 proposed to skip min/max recording in case of CircleInput with `dtype != FLOAT32`, but I think better to skip min/max recording for any operation with integer output as more general solution.

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

PR #8174, opened 2021-12-14
[luci/pass] Enable OneHot quantization

m-bronnikov:
This commit enables quantization of OneHot op.

-------------------

For: #8065
Draft: #8171

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1130,8 +1130,8 @@ void propagate_pack_quantparam(luci::CirclePack *pack, loco::DataType quant_type
  *
  *  AFTER
  *
- *      [CircleNode] [CircleConst] [CircleConst] [CircleNode]
- *          (S32)        (S32)     (U8 qparam2)  (U8 qparam2)
+ *      [CircleNode] [CircleConst] [CircleConst] [CircleNode]      [CircleConst] <- Dead node
+ *          (S32)        (S32)     (U8 qparam2)  (U8 qparam2)         (FP32)
  *              \          \           /           /
  *               \          \        /            /
  *                \          \     /            /
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1138,7 +1138,7 @@ void propagate_pack_quantparam(luci::CirclePack *pack, loco::DataType quant_type
  *                 -------[CircleOneHot]-------
  *                         (U8 qparam2)
  *
- * NOTE Quantization parameter of CirclePack (qparam2) is propagated to on_value/off_value.
+ * NOTE Quantization parameter of CircleOneHot (qparam2) is propagated to on_value/off_value.
  */
 void propagate_one_hot_quantparam(luci::CircleOneHot *one_hot, loco::DataType quant_type)
 {
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1114,6 +1114,86 @@ void propagate_pack_quantparam(luci::CirclePack *pack, loco::DataType quant_type
   }
 }
 
+/** EXAMPLE
+ *
+ *
+ *
+ * BEFORE
+ *
+ *      [CircleNode] [CircleConst] [CircleConst] [CircleNode]
+ *          (S32)        (S32)        (FP32)     (U8 qparam1)
+ *              \          \           /            /
+ *               \          \        /            /
+ *                \          \     /            /
+ *                 -------[CircleOneHot]-------
+ *                         (U8 qparam2)
+ *
+ *  AFTER
+ *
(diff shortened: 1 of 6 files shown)


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1114,6 +1114,86 @@ void propagate_pack_quantparam(luci::CirclePack *pack, loco::DataType quant_type
...
+      luci::CircleConst *const_node = loco::must_cast<luci::CircleConst *>(node);
+      if (is_quantized(const_node))
+        return;
+
+      if (const_node->dtype() != loco::DataType::FLOAT32)
+        throw std::runtime_error("Unsupported data type for constant input of OneHot Op");
+
+      const auto qparam = one_hot->quantparam();
+      if (qparam == nullptr)
+        throw std::runtime_error("quantparam of PadV2 is not found during propagation");
```
jinevening, line 1165:
```suggestion
        throw std::runtime_error("quantparam of OneHot is not found during propagation");
```

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1114,6 +1114,86 @@ void propagate_pack_quantparam(luci::CirclePack *pack, loco::DataType quant_type
...
+ *                 -------[CircleOneHot]-------
+ *                         (U8 qparam2)
+ *
+ * NOTE Quantization parameter of CircleOneHot (qparam2) is propagated to on_value/off_value.
+ */
+void propagate_one_hot_quantparam(luci::CircleOneHot *one_hot, loco::DataType quant_type)
+{
+  assert(one_hot->quantparam() != nullptr);
+
+  // Propagate quantization paramters from output to inputs,
```
jinevening, line 1147:
```suggestion
  // Propagate quantization parameters from output to inputs,
```

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -795,6 +795,53 @@ class ConcatenationTestGraph final : public SimpleTestGraph
...
 };
 
+template <Type indexT> class OneHotTestGraph final : public SimpleTestGraph
+{
+public:
+  void init(void) override
+  {
+    TestIOGraph::init({32}, {32, 10});
+    {
+      // input dtype is float by default, but OneHot should have indexType (s32/s64)
```
m-bronnikov, line 805:
```suggestion
      // input dtype is float by default, but OneHot's input should have indexType (s32/s64)
```

seanshpark:
Please use `[luci/pass]` or just `[luci]` instead of `[circle-quantizer]`.

The meaning of `[xxx]` used in `compiler` is the module name.
You don't have to follow this scheme but this helps reviewers to catch the module and understand the change better.
If you use the different name, this leads mis understanding which I don't think you intend to do.

m-bronnikov:
> Please use `[luci/pass]` or just `[luci]` instead of `[circle-quantizer]`.

Rebased with another module in title.

jinevening (approved):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1114,6 +1114,86 @@ void propagate_pack_quantparam(luci::CirclePack *pack, loco::DataType quant_type
...
+
+    // Quantize constant values
+    if (node->opcode() == luci::CircleOpcode::CIRCLECONST)
+    {
+      luci::CircleConst *const_node = loco::must_cast<luci::CircleConst *>(node);
+      if (is_quantized(const_node))
+        return;
+
+      if (const_node->dtype() != loco::DataType::FLOAT32)
+        throw std::runtime_error("Unsupported data type for constant input of OneHot Op");
```
seanshpark, line 1161: @m-bronnikov , @jinevening , Q) can there be other types than F32 as normal type?
if this(`const_node->dtype() == loco::DataType::FLOAT32`) is a MUST then throwing seems OK but if there can be other types, then it would be better to just return

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
seanshpark: ditto for (1) just return or (2) throw

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
m-bronnikov: @seanshpark Please note this lambda is a bit modififed version of lambda from `propagate_pad_v2_quantparam`, which contains following lines:

https://github.com/Samsung/ONE/blob/0fb144c4d75526b7ffe6e2befa11905dab236659/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp#L1433-L1434

Also, this lambda calls only for `on_value` and `off_value` nodes, which must have same type, as output node, which must have qparam, and, therefore `on_value`/`off_value` must has FP32 type or be already quantized. If node quantized, this lambda makes `return` (see above).

seanshpark: Thanks for the explanation!

seanshpark (approved):
LGTM thank you!

PR #8175, opened 2021-12-14
[record-minmax] Allow random integer inputs

m-bronnikov:
This commit allows integer inputs in `profileDataWithRandomInputs`.

--------------------

For: #8065
Draft: #8171

### DIFF
--- a/compiler/record-minmax/src/RecordMinMax.cpp
+++ b/compiler/record-minmax/src/RecordMinMax.cpp
@@ -405,7 +405,9 @@ void RecordMinMax::profileDataWithRandomInputs(const std::string &mode, float mi
 
       // TODO Support more input data types
       assert(input_node->dtype() == loco::DataType::FLOAT32 ||
-             input_node->dtype() == loco::DataType::BOOL);
+             input_node->dtype() == loco::DataType::BOOL ||
+             input_node->dtype() == loco::DataType::S32 ||
+             input_node->dtype() == loco::DataType::S64);
 
       if (input_node->dtype() == DataType::FLOAT32)
       {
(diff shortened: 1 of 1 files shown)


compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -404,27 +404,45 @@ void RecordMinMax::profileDataWithRandomInputs(const std::string &mode, float mi
...
+        }
+        case DataType::S64:
+        {
+          auto input_data = genRandomIntData<int64_t>(gen, num_elements, -100, 100);
+          _interpreter->writeInputTensor(input_node, input_data.data(),
+                                         input_data.size() * sizeof(int64_t));
+          break;
+        }
+        default:
+          throw std::runtime_error("Unsupported input type, yet!");
```
m-bronnikov, line 445: For reviewer: note this PR introduces throw of `runtime_error()` in case of unsupported input datatype and removes `assert`, which checks datatypes (see above).

seanshpark:
@m-bronnikov , you have (1) changed if-else to switch-case (2) add to support other types.
(a) you did not explain about (1)
(b) there are two different context of changes in one PR, which we don't prefer.

It would be better to reflect these and I'd prefer to split changes to separate PRs.

m-bronnikov:
> @m-bronnikov , you have (1) changed if-else to switch-case (2) add to support other types. (a) you did not explain about (1) (b) there are two different context of changes in one PR, which we don't prefer.
> 
> It would be better to reflect these and I'd prefer to split changes to separate PRs.

Part (1) removed from this PR for now. PTAL.

seanshpark (approved):
LGMT thank you!

jinevening (approved):
LGTM

Issue #8169, opened 2021-12-14
[one-cmds] Reivse utils.py


mhs4670go:
## What

Let's revise `utils.py` of `one-cmds`.

## Why

As `one-cmds` gets matured, the size of `utils.py` is getting bigger. Now, `utils` is too ambiguous to know its purpose. So, I think it'd be better to split it into more specific ones for maintainability and readability.

This is what i think it is going to look like.

```bash
# AS-IS
$ tree
.
├── circle2circle
├── circledump
..
├── generate_bcq_metadata.py
├── generate_bcq_output_arrays.py
├── luci_eval_driver
..
├── tflite2circle
├── utils.py
└── venv/
```

```bash
# TO-BE
$ tree
.
├── circle2circle
├── circledump
..
├── python/
    ├── generate_bcq_metadata.py
    ├── generate_bcq_output_arrays.py
    ├── make_command.py
    ├── parse.py
    └── ..

```

---
Actually, other files could be moved to another single directory as well later. Then, `one-cmds` drivers would be remained at parent directory.

```bash
# TO-BE
$ tree
.
├── one-build
├── one-build.template.cfg
├── onecc
├── onecc.template.cfg
├── one-codegen
├── one-import
├── one-import-bcq
├── one-import-onnx
├── one-import-tf
├── one-import-tflite
├── one-optimize
├── one-pack
├── one-prepare-venv
├── one-profile
├── one-quantize
├── one-version
├── python/
│   ├── generate_bcq_metadata.py
│   ├── generate_bcq_output_arrays.py
│   ├── onnx_legalizer.py
│   ├── tf2tfliteV2.py
│   └── ..
├── scripts/
│   └── model2nnpkg.sh
├── tools/
│   ├── circle2circle
│   ├── circledump
│   ├── circle_partitioner
│   ├── circle-quantizer
│   ├── circle-tensordump
│   ├── luci_eval_driver
│   ├── rawdata2hdf5
│   ├── record-minmax
│   ├── tf2nnpkg
│   ├── tflchef
│   ├── tflchef-file
│   ├── tflchef-reverse
│   └── tflite2circle
└── venv/
```

seanshpark:
```
├── python/
    ├── generate_bcq_metadata.py
```
why is the folder name `python` ?

mhs4670go:
@seanshpark It can be better name. I just wrote it because there will be only python files. Just `utils` is came up with as a directory name.

seanshpark:
[...]

PR #8182, opened 2021-12-15
[one-cmds] Add parameter to autodetect drivers

binarman:
This PR adds search path parameter to _detect_one_import_drivers.

Related issue: #8169.


binarman:
As `util.py` could be moved in the directory different from `one-import-*` files, I decided that it would be good to parameterize detecting function from most outer script =)

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #8197, opened 2021-12-20
[luci-interpreter] Move common activation related code to utils

binarman:
This PR moves computation of min/max activation values from arithmetic kernels
to Utils.h for float, S32 and S64 types.

Simplify code added in #8155 

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Add.cpp
+++ b/compiler/luci-interpreter/src/kernels/Add.cpp
@@ -76,13 +76,8 @@ void Add::execute() const
 
 void Add::evalFloat() const
 {
-  float activation_min{};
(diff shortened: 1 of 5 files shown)


binarman:
@m-bronnikov  FYI
I've also tried to merge integer types into one TYPED_TEST after your comment, but found that I need either create separate testsuite or rewrite other tests (they use fixture class, which becomes typed) :slightly_frowning_face: 

So after all I decided to leave these tests as-is =)

binarman:
@nnfw-bot test nncc-debug

m-bronnikov (approved):
LGTM

kvochko (approved):
LGTM

Issue #8198, opened 2021-12-21
[luci/export] Refactoring Circle Operation Exporter


llFreetimell:
### What?

Currently, we applied following workarounds to avoid large LoC.

- Separating `visit` functions with alphabet grouping
https://github.com/Samsung/ONE/blob/638a1e64b417a6ff9fe243b402e95eac2b449b2e/compiler/luci/export/src/CircleOperationExporter.cpp#L956-L964
- Just moving implementation to another function
https://github.com/Samsung/ONE/blob/638a1e64b417a6ff9fe243b402e95eac2b449b2e/compiler/luci/export/src/CircleOperationExporter.cpp#L982

Let's try to remove the workarounds!

### How?

I will apply following things
- Remove duplicated code
- Integrate separated operations
- Create separate Visitor classes

Then I expect following result :)

#### Before

![before](https://user-images.githubusercontent.com/24720192/146870634-38c94305-be3e-43fb-b691-aeea6b01e906.png)

#### After

![after](https://user-images.githubusercontent.com/24720192/146870668-237ce114-a126-4cba-a110-6e9ee78c6144.png)

---

#### Related PRs

- Draft : #8199
- [x] Introduce BuiltinTypesMappingRule
  - #8205
- [x] Introduce BuiltinOptionsExtractor
  - #8209
- [x] Util functions update
  - #8206
  - #8211
- [x] Introduce OperationExporterRule
  - #8218
- [x] Apply refactored class
  - #8220
- [x] Remove unused codes
  - #8225

seanshpark:
@llFreetimell , please share LoC value (URL would be better through our internal github) from our static analyzer of your draft :)

llFreetimell:
@seanshpark Please refer to internal PR#13154, LoC value is same with current :D

seanshpark:
FYI #8203 was for new Op. I am curious of how new Ops can be added without and problems for latest refactorings of mine and yours :)

llFreetimell:
> I am curious of how new Ops can be added without any problems for latest refactorings of mine and yours

Plus my additional (but not that critical) concern is.. LoC score would be 4.99 not 5.0 after `CircleSVDF` is added ^.T

seanshpark:
> LoC score would be 4.99 not 5.0 after CircleSVDF is added ^.T

Well, this was my concern as all the lines for all the nodes in a class adds up and that sometimes exceeded the tipping point ... -_-; I think this was why I splitted to multiple classes.

[...]

PR #8199, opened 2021-12-21
[DRAFT][luci/export] Refactoring OperationExporter

llFreetimell:
This draft is for refactoring Circle Operation Exporter.

---

Parent Issue : #8198


seanshpark:
(optional) as you introduced `compiler/luci/export/src/CircleOps.lst` file,
I assume incremental changes for each luci sub-modules can land with out breaking CI.
Can you confirm this?

compiler/luci/export/src/CircleBuiltinTypesExtractor.h
```diff
@@ -0,0 +1,527 @@
...
+#include <luci/IR/CircleNode.h>
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleNodeVisitor.h>
+
+#include <flatbuffers/flexbuffers.h>
+
+namespace luci
+{
+
+// NOTE Return value 0 means that the operation has no option
```
seanshpark, line 31: (optioal)
```cxx
#define BUILTIN_OPTION_NONE 0
...
flatbuffers::Offset<void> visit(luci::CircleCeil *) { return BUILTIN_OPTION_NONE ; }
```

compiler/luci/export/src/CircleBuiltinTypesMappingRule.h
```diff
@@ -0,0 +1,77 @@
...
+  static BuiltinOperatorMappingRule &get()
+  {
+    static BuiltinOperatorMappingRule instance;
+    return instance;
+  }
+
+public:
+#define CIRCLE_NODE(CIRCLE_CLASS, OP, OPTION) \
+  circle::BuiltinOperator visit(const CIRCLE_CLASS *) final { return circle::OP; }
+#define CIRCLE_VNODE(CIRCLE_CLASS)
```
seanshpark, line 45: (optional) adding a comment may help understand the code
```cxx
// Virtual nodes do not need export
#define CIRCLE_VNODE(CIRCLE_CLASS)
```

compiler/luci/export/src/CircleBuiltinTypesMappingRule.h
```diff
@@ -0,0 +1,77 @@
...
+  static BuiltinOptionsMappingRule &get()
+  {
+    static BuiltinOptionsMappingRule instance;
+    return instance;
+  }
+
+public:
+#define CIRCLE_NODE(CIRCLE_CLASS, OP, OPTION) \
+  circle::BuiltinOptions visit(const CIRCLE_CLASS *) final { return circle::OPTION; }
+#define CIRCLE_VNODE(CIRCLE_CLASS)
```
seanshpark, line 69: ditto

compiler/luci/export/src/CircleOps.lst
```diff
@@ -0,0 +1,149 @@
+#ifndef CIRCLE_NODE
```
seanshpark, line 1: Q) was there any particular reason to introduce this list file not using existing on in lang?

compiler/luci/export/src/SerializedData.h
```diff
@@ -131,8 +132,8 @@ struct SerializedModelData final
    * @param builtin_code
    * @return idx of opcode in table of opcodes (see schema)
    */
-  uint32_t registerBuiltinOpcode(circle::BuiltinOperator builtin_code, const int32_t op_version);
-  uint32_t registerCustomOpcode(const std::string &custom_op);
+  uint32_t registerBuiltinOpcode(circle::BuiltinOperator builtin_code, const int32_t op_version,
+                                 const std::string &custom_code);
```
seanshpark, line 136: Q) please share why remove `registerCustomOpcode` and merge into `registerBuiltinOpcode`

llFreetimell:
> as you introduced compiler/luci/export/src/CircleOps.lst file,
> I assume incremental changes for each luci sub-modules can land with out breaking CI.

I didn't catch your point T.T
`compiler/luci/export/src/CircleOps.lst` is only used in `luci/export` and not used in other modules.
So I thought that it would not affect other luci sub-modules.
What am I misunderstanding...?

compiler/luci/export/src/CircleOps.lst
llFreetimell: In this draft, `luci/export` needed relation between CircleNode and BuiltinOption/BuiltinOperator.
(e.g. CircleAbs <--> BuiltinOperator_ABS, BuiltinOptions_AbsOptions)

However, `luci/lang/CircleNodes.lst` has only relation between CircleNode and OpCode
(e.g. CircleAbs <--> ABS)

As a result, I had two options, adding BuiltinOptions to `luci/lang/CircleNodes.lst` and creating new `luci/export/CircleOps.lst`.
I selected latter option because
- `luci/lang/CircleNodes.lst` is originally for OpCode of circle, not for BuiltinTypes.
- `luci/lang` is a module about CircleNodes itself, but relation with BuiltinTypes are not about CircleNode.
- If I add to `luci/lang`, I should update it even when circle schema is updated but CircleNode is not needed to be updated.

seanshpark:
> So I thought that it would not affect other luci sub-modules.

Hum... maybe I was overthinking 😭.

compiler/luci/export/src/SerializedData.h
llFreetimell: It was because of following reason :)

- To reduce duplicated codes
  - Only the content of  `custom_code` is different for both functions
- Both functions store same structure
  - Both CustomOp and NormalOp should store same data(i.e. `builtin_code`, `op_version`, `custom_code`)
  - CustomOp is not special case, just NormalOp has empty string for `custom_code`

compiler/luci/export/src/CircleBuiltinTypesExtractor.h
llFreetimell: I will apply this by const private member, not define :D

seanshpark:
@llFreetimell , diff codes are a bit hard to read.
Can you explain how the change is for one Op?

compiler/luci/export/src/CircleOps.lst
seanshpark: OK thanks, got it :)

llFreetimell:
> diff codes are a bit hard to read.

Sorry for confusion... I will separate the codes as several commits for better readability...!

> Can you explain how the change is for one Op?

I wrote following description with additional comments.
But it may not enough for easy understanding....
I will think little bit more if there is a good idea for better understanding... T.T

#### CircleAdd

This operation has multiple inputs, single outputs (Most simple operation)
Let's assume following simple example operation!

```
[Tensor1] --+--    [CircleAdd]    -- [Tensor3]
[Tensor2] --+   (fusedAct = ReLU)
```

The operation is exported as following STEPs :)

```cpp
// CircleOperationExporter.cpp
void exportNodes(loco::Graph *g, flatbuffers::FlatBufferBuilder &builder, SerializedModelData &md,
                 SerializedGraphData &gd)
{
    // ...

    // STEP 1. Call OperationExporterRule
    OperationExporterRule exporter_rule{ctx};
    circle_node->accept(&exporter_rule);

    // ...
}

// CircleOperationExporterRule.cpp
void OperationExporterRule::visit(luci::CircleNode *node)
{
  // STEP 2. Get input vectors (Same with current code)
  //
  // inputs_vec = {Tensor1, Tensor2}
  std::vector<int32_t> inputs_vec;
  for (uint32_t i = 0; i < node->arity(); ++i)
    inputs_vec.push_back(luci::get_tensor_index(node->arg(i)));

  // STEP 3. Extract output vector using OutputVectorExtractor
  //
  // outputs_vec = {Tensor3}
  OutputVectorExtractor outputs_vec_extractor;
  auto outputs_vec = node->accept(&outputs_vec_extractor);

  // STEP 4. Extract BuiltinOptions using BuiltinOptionsExtractor
  //
  // options_offset = 
  // flatbuffers::Offset<void>
  //   (AddOptions{fused_activation_function=ActivationFunctionType_RELU})
  luci::BuiltinOptionsExtractor builtin_options_extractor(_ctx.builder);
  auto options_offset = node->accept(&builtin_options_extractor);

  // STEP 5. Register Builtin Opcode using two functions.
  //         (They are implented with CircleBuiltinTypesMappingRule.h)
  //
  // circle_builtin_operator(node) = BuiltinOperator_ADD
  // circle_custom_code(node) = ""
  auto op_idx = _ctx.md.registerBuiltinOpcode(circle_builtin_operator(node), node->op_version(),
                                              circle_custom_code(node));

  // STEP 6. (Same with current code)
  auto inputs = _ctx.builder.CreateVector(inputs_vec);
  auto outputs = _ctx.builder.CreateVector(outputs_vec);

  // STEP 7. Get BuiltinOptions
  //
  // circle_builtin_options(node) = BuiltinOptions_AddOptions
  auto builtin_options = circle_builtin_options(node);

  // STEP 8. Get custom options
  //         If node is not CircleCustom, null offset(0) is returned.
  //
  // custom_options = flatbuffers::Offset<flatbuffers::Vector<uint8_t>>(0)
  auto custom_options = circle_custom_options(_ctx.builder, node);

  // STEP 9. (Same with current code)
  auto op_offset = circle::CreateOperator(_ctx.builder, op_idx, inputs, outputs, builtin_options,
                                          options_offset, custom_options);
  _ctx.gd._operators.push_back(op_offset);

  // FINISH!
}
```

seanshpark:
Thanks for the explanation!

llFreetimell:
Related PRs are all created, close :)

Issue #8148, opened 2021-12-09
[one-cmds] Refactor onnx_legalizer


binarman:
## Problem

onnx_legalizer.py code is hard to understand, need to improve it's readability.

## What to do

- [x] replace general `transformer.make_node` method with specialized methods, like `make_gemm` #8202
- [x] add '_' to hide internal functions #8419
- [x] add more comments to main functions #8405
- [ ] add helper methods to change output tensor in operations (need them to remove reset of `make_node` calls)
- [x] Remove unused variables #8419

seanshpark:
Title typo?

seanshpark:
close as done

PR #8202, opened 2021-12-22
[one-cmds] Add dedicated builder methods for various operations

binarman:
This PR adds dedicated methods to create operations of particular type.
This change should improve readability and remove some code duplication.

Related issue: #8148

### DIFF
--- a/compiler/one-cmds/onnx_legalizer.py
+++ b/compiler/one-cmds/onnx_legalizer.py
@@ -99,7 +99,8 @@ class ModelTransformerHelper:
             list: list of output tensor names
 
         '''
-        return self.make_node('Split', [input], len(split_sizes), axis = axis, split = split_sizes)
+        return self.make_node(
+            'Split', [input], len(split_sizes), axis=axis, split=split_sizes)
 
     def make_concat(self, inputs, axis):
         '''Create Concat operation and insert it in graph.
@@ -112,7 +113,7 @@ class ModelTransformerHelper:
             str:: output tensor name
 
         '''
-        return self.make_node('Concat', inputs, 1, axis = axis)[0]
+        return self.make_node('Concat', inputs, 1, axis=axis)[0]
 
     def make_squeeze(self, input, axes):
         '''Create Squeeze operation and insert it in graph.
@@ -125,7 +126,7 @@ class ModelTransformerHelper:
             str: output tensor name
 
         '''
-        return self.make_node('Squeeze', [input], 1, axes = axes)[0]
+        return self.make_node('Squeeze', [input], 1, axes=axes)[0]
 
     def make_unsqueeze(self, input, axes):
         '''Create Unsqueeze operation and insert it in graph.
@@ -138,9 +139,9 @@ class ModelTransformerHelper:
             str: output tensor name
 
         '''
-        return self.make_node('Unsqueeze', [input], 1, axes = axes)[0]
+        return self.make_node('Unsqueeze', [input], 1, axes=axes)[0]
 
-    def make_gemm(self, A, B, C, trans_a = False, trans_b = False):
+    def make_gemm(self, A, B, C, trans_a=False, trans_b=False):
         '''Create Gemm operation and insert it in graph.
 
         Result tensor contains A*B + C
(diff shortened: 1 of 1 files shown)


binarman:
To check correctness of generated models I used onnxruntime-based scripts from #7879:
- [compiler/one-cmds/compare.py](https://github.com/Samsung/ONE/pull/7879/files#diff-0007475344f19c4c34ac29b035e38e6e4a76d70908119b2dee6487b72d2384b8)
- [compiler/one-cmds/testall.sh](https://github.com/Samsung/ONE/pull/7879/files#diff-bd6d11aea69729963251f216de6246615d3a79d76f4fcb95242c90b8eb82afe9)

binarman:
There are still some uses of "general" make_node methods, I plan to remove them later.

It requires to more infrastructure, because these operations should write to already existing tensors, I did not manage to design convenient interface to do this yet.

binarman:
@nnfw-bot test nncc-release

seanshpark:
@nnfw-bot test nncc-release

seanshpark (approved):
LGTM

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -87,11 +87,138 @@ def make_node(self, opcode, inputs, outputs, *p_args, **k_args):
...
+
+    def make_concat(self, inputs, axis):
+        '''Create Concat operation and insert it in graph.
+
+        Args:
+            inputs (list): list of tensors names to concat
+            axis (int): axis number to concat
+
+        Returns:
+            str:: output tensor name
```
mhs4670go, line 113:
```suggestion
            str: output tensor name
```

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -195,20 +320,20 @@ def transform_unidirectional_RNN(transformer, original_node, x, tensor_infos, ac
     outputs = original_node.output
     if direction == 'reverse':
         x.reverse()
-    w = transformer.make_node('Squeeze', [inputs[1]], 1, axes=[0])
-    r = transformer.make_node('Squeeze', [inputs[2]], 1, axes=[0])
+    w = transformer.make_squeeze(inputs[1], axes=[0])
```
mhs4670go, line 323: `make_sqeuuze` returns `self.make_node('Squeeze', [input], 1, axes=axes)[0]` and original code was make_node('Squeeze', [inputs[1]], 1, axes=[0])`, which doesn't have `[0]` at the end of it. Is it okay?

seanshpark: `[0]` : return is list :)

mhs4670go: @seanshpark Umm.. I don't get it. You mean, type of `self.make_node('Squeeze', [input], 1, axes=axes)[0]` is a list?

Original type of `w`(`w = transformer.make_node('Squeeze', [inputs[1]], 1, axes=[0])`) is a list of string or int. And,  Changed one is just a string as the method spec said.

seanshpark: Hum... I thought `self.make_node('Squeeze', [input], 1, axes=axes)` returned list and `[0]` is to access the item... not sure my understanding is correct -_-;

lucenticus (approved):

compiler/one-cmds/onnx_legalizer.py
seanshpark: @binarman , ping ?

binarman: @mhs4670go
Sorry for the long delay.

**First thing: detailed description of what `make_node` method is doing.**

`make_node` is a generic method that should be capable to create nodes with multiple outputs.
In order to do this, it return array of strings(names of output tensors), instead of simple string.

In fact, this method works in two modes, depending on what is the type of `output` method:
- **int**: `output` contains number of outputs operation produces and method will create unique names automatically
- **list of strings**: `output` contains names of already existing tensors, that should be used as-is.

Second mode is used to attach created operations to original model.


**What changed**

Old code is mostly uses arrays of tensor names, new code mostly uses single tensor name for every argument.

Regarding this particular place:

- In old code `make_node('Squeeze')` created array, which is passed to `generate_one_direction_RNN` and then to other `make_node` methods.
- In this PR all values in this chain of calls are single string instead of array.

P.s.
Thanks to you comment I realized that documentation for `generate_one_direction_RNN` and `generate_one_direction_LSTM` is outdated =)

mhs4670go (approved):
LGTM

PR #8208, opened 2021-12-23
[luci/pass] Fix check_4d_reshape method

seanshpark:
This will fix check_4d_reshape to check input rank is 4.


seanshpark:
sometimes input share rank is 2, like `[1, 256]` and reshape it to `[1, 1, 1, 256]` where expanding rank is implicit(?). (not sure my understanding is correct)

jinevening (approved):
LGTM Thank you.

Issue #8213, opened 2021-12-24
[one-cmds] import onnx_legalizer difference


seanshpark:
Current local setup has `onnx_legalizer.py` which will enable `import onnx_legalizer`,
where package setup version doesn't as currently this file isn't included.

From this difference I had a problem of difference, where local setup compiled without problem,
but my test machine installed with debian package failed.

seanshpark:
What I think as for now
(1) use environment variable to enable import onnx_legalizer (for temporary)
(2) add `one-import-onnx` option to enable execute `onnx_legalizer.legalize(onnx_model)`

Another thought;
For `onnx_legalizer.legalize()`, this method is doing RNN/LSTM unrolling as I understand.
If this method is for unrolling, how about chaning method name to `unroll_rnn` ?
If not, I think we need to split actions and call by option from `one-import-onnx`

cc @binarman

seanshpark:
Posted #8215

binarman:
@seanshpark 
> If this method is for unrolling, how about changing method name to unroll_rnn ?

I imagined this method as a top level interface to "legalize" onnx model so it could be processed by later stages, like `onnx->tflite`, `tflite2circle` and `circle2circle` without errors.

For now it only capable of transforming `RNN` and `LSTM` layers, because this is the only problems I faced so far. :slightly_smiling_face: 

I implemented unrolling, instead of general while loop, because it is relative simple and reliable method:
- debugging code inside while loop could be painful
- I did not see NN models where it is needed

binarman:
To make it clear.
Main purpose of `onnx_legalizer` is to **fix problematic ONNX models** that we can not convert for some reason:
- errors in third party converters
- unsupported complex operations
- etc.

If these problems are resolved (for example by updating some converters/frameworks) we can simply remove/disable code in `onnx_legalize.py`

seanshpark:
I think issue from not much understanding is resolved and adding option is in progress.

PR #8215, opened 2021-12-24
[one-cmds] Add option unroll_rnn for onnx import

seanshpark:
This will add unroll_rnn option for onnx import.


seanshpark:
For #8213,

@binarman , what do you think of this?

seanshpark:
@binarman ping

seanshpark:
@binarman ping

binarman:
@seanshpark 
Sorry for the long delay =_=
I was on annual vacation then we had long new year holidays =)

As I described in https://github.com/Samsung/ONE/issues/8213#issuecomment-1009247072 , unrolling itself is not a main goal of "onnx_legalizer", and unrolling is sort of implementation details...

Do you want to control this particular unrolling behavior or do you just want to have enable/disable "legalizer"?

- In first case I I think it is better state that this is "developer" or "experimental" option, and people should not rely on it
- In second case, I would like to remove mention of "unrolling" from opt description

seanshpark:
> Do you want to control this particular unrolling behavior or do you just want to have enable/disable "legalizer"?

Yes for particullar unrolling behavior.
Reason)
- I assume someday this module will be changed to release from experimental
- There will be more than unrolling and each functionality should change default behavior before and after introduing this module into release
- To do this each function should be provided as a switch
- It would be better provide one switch per function for use case

seanshpark:
I've modified to reflect `one switch per function`

seanshpark:
@binarman , PTAL

seanshpark:
changed a bit more from 83b7c8d54f379349f0b471e68a7be53112506132 to fc6386f2eb96537245cb0e7f06ef3fdb289ac9bb
- remove using `utils` in `onnx_legalize`
- introduce `unroll_lstm` option

binarman (approved):
LGTM, Thank you!

binarman:
> - For this, each function should be provided as a switch, and off as default
> - It would be better to provide one switch per function for better use case

Ok, sounds reasonable =)

seanshpark:
@binarman , thanks for review, I'll post a PR :)

mhs4670go (approved):
LGTM

PR #8218, opened 2021-12-24
[luci/export] Introduce CircleOperationExporterRule

llFreetimell:
This commit introduces `CircleOperationExporterRule`,
which includes exporting rule for all CircleNode.

---

Parent Issue : #8198
Draft : #8199

### DIFF
--- a/compiler/luci/export/src/CircleOperationExporterRule.cpp
+++ b/compiler/luci/export/src/CircleOperationExporterRule.cpp
@@ -1,0 +1,277 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


compiler/luci/export/src/CircleOperationExporterRule.cpp
```diff
@@ -0,0 +1,277 @@
...
+    std::vector<int32_t> outputs_vec{luci::get_tensor_index(node)};
+    return outputs_vec;
+  }
+
+  std::vector<int32_t> visit(luci::CircleBidirectionalSequenceLSTM *node) final
+  {
+    auto bidi_lstm_outs = loco::succs(node);
+    assert((bidi_lstm_outs.size() == 1) || (bidi_lstm_outs.size() == 2));
+
+    std::vector<int32_t> outputs_vec(bidi_lstm_outs.size());
```
llFreetimell, line 51: Current code assumes that `CircleBidirectionalSequenceLSTM` has always two outputs.

https://github.com/Samsung/ONE/blob/09ba4eb8bc4e3a76a46bf1da0fddf99947f69242/compiler/luci/export/src/CircleOperationExporter.cpp#L1020

However, according to above assert(Line 49), it can have one output.
This code fixed this :)

seanshpark (approved):
LGTM

PR #8220, opened 2021-12-24
[luci/export] Revise exportNodes function

llFreetimell:
This commit revise `exportNodes` function with following updates.
- Integrate `exportNode` into `exportNodes`
- Substitute `OperationExporter` with `OperationExporterRule`

---

Parent Issue : #8198
Draft : #8199

### DIFF
--- a/compiler/luci/export/src/CircleOperationExporter.cpp
+++ b/compiler/luci/export/src/CircleOperationExporter.cpp
@@ -32,6 +32,8 @@
 
 #include <flatbuffers/flexbuffers.h>
 
+// TODO Remove unused code
+// NOTE This is for better code diff visualization
 #if 0
 using namespace flatbuffers;
 using namespace circle;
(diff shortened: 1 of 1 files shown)


compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -1724,17 +1726,53 @@ void exportNode(loco::Node *node, flatbuffers::FlatBufferBuilder &builder, Seria
...
 
-void exportNodes(loco::Graph *g, FlatBufferBuilder &builder, SerializedModelData &md,
+void exportNodes(loco::Graph *g, flatbuffers::FlatBufferBuilder &builder, SerializedModelData &md,
                  SerializedGraphData &gd)
 {
   uint32_t node_position = 0;
   for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
   {
-    exportNode(node, builder, md, gd, node_position);
+    // exportNode(node, builder, md, gd, node_position);
```
llFreetimell, line 1740: This line will be removed in the next PR

compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -1724,17 +1726,53 @@ void exportNode(loco::Node *node, flatbuffers::FlatBufferBuilder &builder, Seria
...
+    if (auto circle_node = dynamic_cast<luci::CircleNode *>(node))
```
seanshpark, line 1741: as all nodes should be circle nodes, we may use `must_cast`

compiler/luci/export/src/CircleOperationExporter.cpp
```diff
@@ -31,6 +32,7 @@
 
 #include <flatbuffers/flexbuffers.h>
 
+#if 0
```
seanshpark, line 35: Plan to remove in next PR? anyway, maybe add a comment to delete this block?

compiler/luci/export/src/CircleOperationExporter.cpp
llFreetimell: I just copied original code without change and following `else` statement is found.

```cpp
    else
    {
      INTERNAL_EXN("Node with unsupported dialect found");
    }
```

Would it be better to change to `must_cast` and remove if-else statement? :D
(I think using `must_cast` is much pretty :D)

compiler/luci/export/src/CircleOperationExporter.cpp
llFreetimell: I am planning to remove in next PR but it would be better to add a comment :)

compiler/luci/export/src/CircleOperationExporter.cpp
seanshpark: > Would it be better to change to must_cast and remove if-else statement? :D

Yes, you can go as-is for this PR and change later or apply in this PR.

compiler/luci/export/src/CircleOperationExporter.cpp
seanshpark: OK thanks :)

seanshpark (approved):
LGTM thank you!

Issue #8239, opened 2022-01-04
[mio-tflite260] Add testcase and update implementation for helper functions


llFreetimell:
### What?

Current implementation of helper functions are little bit different with Tensorflow codes.
Let's update implementation and add testcases for testing the helper functions!

### Reference

- How to read BuiltinCode
  - https://github.com/tensorflow/community/blob/2c8c1d727a5a46472b24c6adbc0fb17f376f8ae3/rfcs/20200901-tflite-builtin-code-extension.md#how-to-read-builtin_code
  - https://github.com/tensorflow/tensorflow/commit/cf28969fb2ea72c9738cd3acdddc0b69fd7dff5e
  - https://github.com/tensorflow/tensorflow/blob/7d12007d7800d3714a02e05059f3ea602d1aec78/tensorflow/lite/schema/schema_utils.cc
- How to create BuiltinCode 
  - https://github.com/tensorflow/tensorflow/blob/1ab788fa8d08430be239ab970980b891ad7af494/tensorflow/lite/schema/schema_conversion_utils.cc#L46-L50
- Special case which do not follow above rule
  - https://github.com/tensorflow/tflite-micro/blob/25da8245417b834ced53df722d6a0f517eb425b4/tensorflow/lite/micro/test_helpers.cc#L154

---

#### Related PRs

- Draft : #8240
- [x] Update implementation
  - #8244 
- [x] Introduce test case and related codes
  - #8246

PR #8240, opened 2022-01-04
[DRAFT][NO MERGE] Add testcases to mio-tflite260 helpers

llFreetimell:
This draft is for adding testcases to `mio-tflite260` helper functions.

---

Parent Issue : #8239

### DIFF
--- a/compiler/mio-tflite260/src/Helper.cpp
+++ b/compiler/mio-tflite260/src/Helper.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2020 The TensorFlow Authors. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,22 +24,31 @@ namespace mio
 namespace tflite
 {
 
-// This will provide v3/v3a format neutral BuiltinOperator
+/**
+ * This will provide v3/v3a format neutral BuiltinOperator
+ *
+ * This function referenced
+ * https://github.com/tensorflow/tensorflow/blob/7d12007d7800d3714a02e05059f3ea602d1aec78/tensorflow/lite/schema/schema_utils.cc
+ */
 ::tflite::BuiltinOperator builtin_code_neutral(const ::tflite::OperatorCode *opcode)
 {
   assert(opcode != nullptr);
-  int8_t dp_code = opcode->deprecated_builtin_code();
-  // 127 is max of int8_t which is upper bound of v3 builtin_code
-  // NOTE TensorFlow uses 'BuiltinOperator_PLACEHOLDER_FOR_GREATER_OP_CODES' for 127
-  if (dp_code < 127 && dp_code >= 0)
-    return ::tflite::BuiltinOperator(dp_code);
-  return opcode->builtin_code();
+  return std::max(opcode->builtin_code(),
+                  static_cast<::tflite::BuiltinOperator>(opcode->deprecated_builtin_code()));
 }
 
 bool is_valid(const ::tflite::OperatorCode *opcode)
 {
-  ::tflite::BuiltinOperator code = builtin_code_neutral(opcode);
-  return (::tflite::BuiltinOperator_MIN <= code && code <= ::tflite::BuiltinOperator_MAX);
+  const int8_t deprecated_builtin_code = opcode->deprecated_builtin_code();
+  if (!(0 <= deprecated_builtin_code && deprecated_builtin_code <= 127))
+    return false;
+
+  const ::tflite::BuiltinOperator builtin_code = opcode->builtin_code();
+  if (!(::tflite::BuiltinOperator_MIN <= builtin_code &&
+        builtin_code <= ::tflite::BuiltinOperator_MAX))
+    return false;
+
+  return true;
 }
 
 bool is_custom(const ::tflite::OperatorCode *opcode)
--- a/compiler/mio-tflite260/src/Helper.test.cpp
+++ b/compiler/mio-tflite260/src/Helper.test.cpp
@@ -1,0 +1,159 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "mio_tflite260/Helper.h"
+
+#include <flatbuffers/flatbuffers.h>
+#include <gtest/gtest.h>
+
+#include <vector>
+
+class mio_tflite260_helper_test : public ::testing::Test
+{
+protected:
+  void initialization_finish(void)
+  {
+    _fbb.Finish(tflite::CreateModelDirect(_fbb, 0, &_opcodes_vec));
+  }
+
+protected:
+  void add_operator_code(int8_t deprecated_builtin_code, const char *custom_code,
+                         tflite::BuiltinOperator builtin_code)
+  {
+    _opcodes_vec.push_back(tflite::CreateOperatorCodeDirect(
+      _fbb, deprecated_builtin_code, custom_code, 1 /* version */, builtin_code));
+  }
+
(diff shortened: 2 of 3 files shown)


compiler/mio-tflite260/src/Helper.cpp
```diff
@@ -23,22 +24,52 @@ namespace mio
...
 {
   assert(opcode != nullptr);
-  int8_t dp_code = opcode->deprecated_builtin_code();
-  // 127 is max of int8_t which is upper bound of v3 builtin_code
-  // NOTE TensorFlow uses 'BuiltinOperator_PLACEHOLDER_FOR_GREATER_OP_CODES' for 127
-  if (dp_code < 127 && dp_code >= 0)
-    return ::tflite::BuiltinOperator(dp_code);
-  return opcode->builtin_code();
+  return std::max(opcode->builtin_code(),
+                  static_cast<::tflite::BuiltinOperator>(opcode->deprecated_builtin_code()));
```
seanshpark, line 39: please do not change. I intended to code as is.

compiler/mio-tflite260/src/Helper.cpp
```diff
@@ -23,22 +24,52 @@ namespace mio
...
+ *
+ * According to
+ * https://github.com/tensorflow/tensorflow/blob/1ab788fa8d08430be239ab970980b891ad7af494/tensorflow/lite/schema/schema_conversion_utils.cc#L46-L50,
+ * 127 is stored.
+ *
+ * According to schema.fbs of TFLite,
+ * any information is found.
+ *
+ * Until it becomes clear, do not validate about it.
+ */
```
seanshpark, line 60: can you please elaborate this?

compiler/mio-tflite260/src/Helper.cpp
seanshpark: is existing code wrong?

llFreetimell: [Tensorflow guide](https://github.com/tensorflow/community/blob/2c8c1d727a5a46472b24c6adbc0fb17f376f8ae3/rfcs/20200901-tflite-builtin-code-extension.md#how-to-read-builtin_code) and [schema comment](https://github.com/tensorflow/tensorflow/blob/7d12007d7800d3714a02e05059f3ea602d1aec78/tensorflow/lite/schema/schema_v3a.fbs#L987-L989) are saying that `builtin_code()` should be considered first, not `deprecated_builtin_code()`.
And then detailed implementation is updated by https://github.com/tensorflow/tensorflow/commit/cf28969fb2ea72c9738cd3acdddc0b69fd7dff5e because of some issue.
Could you let me what was your intention? :)

---

FYI, current code return incorrect builtin code for [following example](https://github.com/tensorflow/tflite-micro/blob/25da8245417b834ced53df722d6a0f517eb425b4/tensorflow/lite/micro/test_helpers.cc#L154).
```
deprecated_builtin_code = 0
builtin_code = 128 // CUMSUM
```

seanshpark:
@llFreetimell , can you please check this?
(1) using TensorFlow 2.6 create a model with one Op
(2) Ops are ADD, CONV2D and CUMSUM
(3) print the values of `deprecated_builtin_code` and `builtin_code`

compiler/mio-tflite260/src/Helper.cpp
llFreetimell: I agree it's little bit dirty but I had no idea for long URL T.T
Except long URL, which point seems to be elaborated? :D

seanshpark: I can't understand the point of explanation. especially what do you mean by `Until it becomes clear, do not validate about it.` ?

compiler/mio-tflite260/src/Helper.cpp
```diff
@@ -23,22 +24,52 @@ namespace mio
...
+ * i.e., which value is stored when builtin_code() is larger than
+ * BuiltinOperator_PLACEHOLDER_FOR_GREATER_OP_CODES?
+ *
+ * According to
+ * https://github.com/tensorflow/tflite-micro/blob/25da8245417b834ced53df722d6a0f517eb425b4/tensorflow/lite/micro/test_helpers.cc#L154,
+ * 0 is stored.
+ *
+ * According to
+ * https://github.com/tensorflow/tensorflow/blob/1ab788fa8d08430be239ab970980b891ad7af494/tensorflow/lite/schema/schema_conversion_utils.cc#L46-L50,
+ * 127 is stored.
```
seanshpark, line 54:
```cxx
  int8_t deprecated_builtin_code =
      static_cast<int8_t>(BuiltinOperator_PLACEHOLDER_FOR_GREATER_OP_CODES);
  if (builtin_code < BuiltinOperator_PLACEHOLDER_FOR_GREATER_OP_CODES) {
    deprecated_builtin_code = static_cast<int8_t>(builtin_code);
  }
```
The code is like this from your URL.

(1) Adding a URL helps understanding, but need to jump to the code that is not convenient
(2) I think it's not 127 but `deprecated_builtin_code` set to `builtin_code` if it's below or same as 127.

The second URL (this) is from a conversion tool...
If the values are different, I'll follow the main code not this tool.
I would assume there is no big deal in difference or someday this tool may be revised from some problem.

compiler/mio-tflite260/src/Helper.test.cpp
```diff
@@ -0,0 +1,160 @@
...
+class mio_tflite260_helper_test : public ::testing::Test
+{
+protected:
+  void initialization_finish(void)
+  {
+    _fbb.Finish(tflite::CreateModelDirect(_fbb, 0, &_opcodes_vec));
+  }
+
+protected:
+  void add_operator_code(int8_t deprecated_builtin_code, const char *custom_code, int32_t version,
```
seanshpark, line 33: all the users of this method calls with `version` as `1`.
if you don't have any plan to check with version as of now (let the future fix this if needed), we may set version to 1 and remove this argument.

llFreetimell:
> can you please check this?
> (1) using TensorFlow 2.6 create a tflite model with one Op
> (2) Ops are ADD, CONV2D and CUMSUM
> (3) print the values of `deprecated_builtin_code` and `builtin_code`

Result from `tfldump` with TF 2.6 :)

```
===================================================================
Model version: 3
 # sub graphs: 1

Operator Codes: [order] OpCodeName (OpCode Enum)
[0] ADD (code: 0, dep_code: 0, version: 1)
...
```
```
===================================================================
Model version: 3
 # sub graphs: 1

Operator Codes: [order] OpCodeName (OpCode Enum)
[0] SUB (code: 41, dep_code: 41, version: 1)
...
```
```
===================================================================
Model version: 3
 # sub graphs: 1

Operator Codes: [order] OpCodeName (OpCode Enum)
[0] CUMSUM (code: 128, dep_code: 127, version: 1)
...
```

compiler/mio-tflite260/src/Helper.cpp
llFreetimell: > (2) I think it's not 127 but deprecated_builtin_code set to builtin_code if it's below or same as 127.

I only focused on "opcode which is larger than 127" cases :D
For "opcode which is smaller than 127" cases, you are correct :)

That's why I wrote "i.e., which value is stored when builtin_code() is larger than BuiltinOperator_PLACEHOLDER_FOR_GREATER_OP_CODES?" in comment

---
>  someday this tool may be revised from some problem.

If we only care about [Second URL](https://github.com/tensorflow/tensorflow/blob/1ab788fa8d08430be239ab970980b891ad7af494/tensorflow/lite/schema/schema_conversion_utils.cc#L46-L50), I agree that there is no problem for now.
Then shall we go as-is? :D

compiler/mio-tflite260/src/Helper.cpp
llFreetimell: I see it :D
Then it would be better to remove this comment and then add it later if related codes are written :)

compiler/mio-tflite260/src/Helper.cpp
seanshpark: > Then shall we go as-is? :D

Um? didn't quite catch this. Cabn you update this PR (the comment) so that it is clear?

llFreetimell: Ah, I mean "if you don't want to change the implementation now, I will not change" :)

seanshpark: I think `builtin_code_neutral` should be changed as in this PR as current implementation seems to have a problem.

I think your `is_valid` code is OK.
Existing one depends on `builtin_code_neutral` which current PR does't and looks more clear.

About the comment, I think it would be better to make simple and clear. (I assumed they were added for the reviewers not for landing... but just a guess)

llFreetimell: > About the comment, I think it would be better to make simple and clear. (I assumed they were added for the reviewers not for landing... but just a guess)

Half was for review, half was for landing.
But I think it would be better to remove the comments because it looks like... Too Much Information =)
Instead, I will leave reference links to #8239!

seanshpark:
Looks Good :)

llFreetimell:
All related PRs are created, close!

Issue #8241, opened 2022-01-04
[Frontend] Support Gather


seanshpark:
Our internal model has `Gather` Op which _luci-interpreter_ doesn't support yet.
Let's enable this Op.

There is abnormal termination in `record-minmax`, in circle quantization step.

seanshpark:
Seems done

PR #8245, opened 2022-01-04
DRAFT Enable Gather

seanshpark:
on-going draft to enable Gather in quantization.


seanshpark:
For #8241

seanshpark:
Current implementation has some problems and there can be other methods of specifing input values;
#8245 is added after a small offline talk.

seanshpark:
done

PR #8248, opened 2022-01-05
[common-artifacts] Support sequential_int option

seanshpark:
This will introduce sequential_int option for int32/int64 data type to
prepare sequential input data for Ops that may have problem with random
data.

### DIFF
--- a/compiler/common-artifacts/src/TestDataGenerator.cpp
+++ b/compiler/common-artifacts/src/TestDataGenerator.cpp
@@ -74,6 +74,14 @@ template <> void geneate_random_data<bool>(std::mt19937 &gen, void *data, uint32
   }
 }
 
+template <typename T> void geneate_sequencial_data(void *data, uint32_t size)
+{
+  for (uint32_t i = 0; i < size; i++)
+  {
+    static_cast<T *>(data)[i] = static_cast<T>(i);
+  }
+}
+
 void fill_random_data(void *data, uint32_t size, loco::DataType dtype, uint32_t seed)
 {
   std::mt19937 gen(seed); // standard mersenne_twister_engine seeded with rd()
@@ -100,6 +108,21 @@ void fill_random_data(void *data, uint32_t size, loco::DataType dtype, uint32_t 
   }
 }
 
+void fill_sequencial_data(void *data, uint32_t size, loco::DataType dtype)
+{
+  switch (dtype)
+  {
+    case loco::DataType::S32:
+      geneate_sequencial_data<int32_t>(data, size);
+      break;
+    case loco::DataType::S64:
+      geneate_sequencial_data<int64_t>(data, size);
+      break;
+    default:
+      throw std::runtime_error("NYI data type.");
+  }
+}
+
 } // namespace
 
 int entry(int argc, char **argv)
@@ -120,6 +143,10 @@ int entry(int argc, char **argv)
     .required(false)
     .nargs(0)
     .help("Put a fixed seed into the random number generator");
+  arser.add_argument("--sequential_int")
+    .required(false)
+    .nargs(0)
+    .help("Use sequential generation for int32/int64 inputs");
 
   try
   {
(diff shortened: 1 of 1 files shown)


seanshpark:
From draft #8245

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -217,7 +244,10 @@ int entry(int argc, char **argv)
       std::vector<int8_t> data(byte_size);
 
       // generate random data
-      if (arser["--fixed_seed"])
+      if (arser["--sequential_int"] && (input_node->dtype() == loco::DataType::S32 ||
+                                        input_node->dtype() == loco::DataType::S64))
```
llFreetimell, line 248: It seems that type checking is already done inside of `fill_sequencial_data`.
And if `--sequential_int` is given but `input_node->dtype()` is not `S32`/`S64`, it is not thrown, just do nothing.
So how about following? :D

```suggestion
      if (arser["--sequential_int"])
```

seanshpark: There will be multiple nodes in the graph and each node may have different dtype.
If we remove dtype checking, it'll call `fill_sequencial_data` for FLOAT32 also, which is incorrect.

seanshpark: `--sequential_int` is applied to model, not per node.

llFreetimell: > If we remove dtype checking, it'll call fill_sequencial_data for FLOAT32 also, which is incorrect.

Yes it's incorrect and thus it will be thrown inside of `fill_sequencial_data`.

```cpp
// This code is copy&pasted from this PR

void fill_sequencial_data(void *data, uint32_t size, loco::DataType dtype)
{
  switch (dtype)
  {
    case loco::DataType::S32:
      geneate_sequencial_data<int32_t>(data, size);
      break;
    case loco::DataType::S64:
      geneate_sequencial_data<int64_t>(data, size);
      break;
    default:
      throw std::runtime_error("NYI data type.");
  }
}
```

So current code is doing duplicated type checking I think :)

seanshpark: Yes, but throwing will stop the process which I don't think is good thing to do.
This (always) happens
```
[ 68%] Generate input.h5 and expected.h5 in Gather_000.opt/metadata/tc
terminate called after throwing an instance of 'std::runtime_error'
  what():  NYI data type.
Aborted (core dumped)
compiler/common-artifacts/CMakeFiles/common_artifacts_deps.dir/build.make:5578: recipe for target 'compiler/common-artifacts/Gather_000.opt/metadata/tc/input.h5' failed
make[3]: *** [compiler/common-artifacts/Gather_000.opt/metadata/tc/input.h5] Error 134
make[3]: *** Deleting file 'compiler/common-artifacts/Gather_000.opt/metadata/tc/input.h5'
```

seanshpark: This is because first input of `Gather` of `Gather_000` recipe is dtype FLOAT32...

llFreetimell: > throwing will stop the process which I don't think is good thing to do.

If so, how about removing following codes in `fill_sequencial_data`  and add some comments?

```cpp
default:
  throw std::runtime_error("NYI data type.");
```

For me, it looks like dummy code because... function has error checking, but current code is avoiding it intentionally @.@

seanshpark: You got a point...
But current code is like
- `fill_sequencial_data` provides a function to fill with data, that happens to be S32 and S64
- this function is implementa as is and other types may not be fill and
- instead of just returning, throw will notifiy to the call(programmer) to implement or don't use such type
- this is view point `fill_sequencial_data` itself
- the caller in this coe is to call when with optioin `sequential_int` and with condition only for S32 and S64
- this is because for other types we should call for random data generation

There exist two different independent things that happens to be having same condition `S32 && S64`.

seanshpark: As wrote in #8249 , I'm thinking to revise to another method to prepare the input data file as you are holding the landing of this PR.

llFreetimell: I see it, thanks for the explanation :)

llFreetimell (approved):
LGTM

mhs4670go (approved):
LGTM

seanshpark:
Thanks for the review...
I've prepared another method to specify input node name and range,
which I think is better than this one.

PR #8250, opened 2022-01-05
[res] Fix Gather_000 recipe

seanshpark:
This will remove graph input for indices in Gather_000 recipe
as indices is constant buffer.


seanshpark:
From draft #8245

seanshpark:
`filler { tag: "explicit" arg: "1" arg: "2" }` will make a Buffer which becomes `CircleConst` with data.

llFreetimell (approved):
LGTM, thank you :)

PR #8253, opened 2022-01-05
[common-artifacts] Introduce input_range

seanshpark:
This will introduce '--input_range' option to specify min,max range of
values for the random input for particular input.

### DIFF
--- a/compiler/common-artifacts/src/TestDataGenerator.cpp
+++ b/compiler/common-artifacts/src/TestDataGenerator.cpp
@@ -27,6 +27,9 @@
 #include <memory>
 #include <random>
 #include <string>
+#include <vector>
+#include <cassert>
+#include <cstdlib>
 
 namespace
 {
@@ -74,6 +77,20 @@ template <> void geneate_random_data<bool>(std::mt19937 &gen, void *data, uint32
   }
 }
 
+template <typename T>
+void generate_random_range(void *data, uint32_t size, int32_t range_min, int32_t range_max)
+{
+  assert(range_min <= range_max);
+
+  for (uint32_t i = 0; i < size; i++)
+  {
+    // +1 will make value of [range_min, range_max]
+    int32_t range = range_max - range_min + 1;
+    int32_t value = (rand() % range) + range_min;
+    static_cast<T *>(data)[i] = static_cast<T>(value);
+  }
+}
+
 void fill_random_data(void *data, uint32_t size, loco::DataType dtype, uint32_t seed)
 {
   std::mt19937 gen(seed); // standard mersenne_twister_engine seeded with rd()
@@ -100,6 +117,22 @@ void fill_random_data(void *data, uint32_t size, loco::DataType dtype, uint32_t 
   }
 }
 
+void fill_random_range(void *data, uint32_t size, loco::DataType dtype, int32_t range_min,
+                       int32_t range_max)
+{
+  switch (dtype)
+  {
+    case loco::DataType::S32:
+      generate_random_range<int32_t>(data, size, range_min, range_max);
+      break;
+    case loco::DataType::S64:
+      generate_random_range<int64_t>(data, size, range_min, range_max);
+      break;
+    default:
+      throw std::runtime_error("NYI data type.");
+  }
+}
+
 } // namespace
 
 int entry(int argc, char **argv)
@@ -120,6 +153,11 @@ int entry(int argc, char **argv)
     .required(false)
     .nargs(0)
     .help("Put a fixed seed into the random number generator");
+  arser.add_argument("--input_range")
+    .required(false)
+    .nargs(3)
+    .type(arser::DataType::STR_VEC)
+    .help("Set random number range [min max] for the input as 'name min max'");
 
   try
   {
@@ -176,6 +214,24 @@ int entry(int argc, char **argv)
   std::unique_ptr<H5::Group> output_value_group =
     std::make_unique<H5::Group>(output_file.createGroup("value"));
 
+  std::string range_name;
+  int32_t range_min = 0;
+  int32_t range_max = 0;
+  bool range_check = false;
+  bool range_input_found = false;
+  if (arser["--input_range"])
+  {
+    // NOTE limitation: we can only set one input range
+    // TODO expand this for multiple inputs
+    std::vector<std::string> values = arser.get<std::vector<std::string>>("--input_range");
+    assert(values.size() == 3);
+    range_name = values.at(0);
+    // TODO add check for valid numbers
+    range_min = std::atoi(values.at(1).c_str());
+    range_max = std::atoi(values.at(2).c_str());
+    range_check = true;
+  }
+
   std::random_device rd; // used to obtain a seed for the random number engine
   uint32_t input_index = 0;
   // TODO remove indentation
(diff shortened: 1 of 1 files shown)


seanshpark:
From #8245 to replace #8248

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -216,8 +252,21 @@ int entry(int argc, char **argv)
       auto byte_size = dtype_size * data_size;
       std::vector<int8_t> data(byte_size);
 
+      std::string range_name;
+      int32_t range_min = 0;
+      int32_t range_max = 0;
+      if (arser["--input_range"])
+      {
+        std::vector<std::string> values = arser.get<std::vector<std::string>>("--input_range");
```
mhs4670go, line 260: (optional)
```suggestion
        std::vector<std::string> values = arser.get<std::vector<std::string>>("--input_range");
        assert(values.size() == 3);
```

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -216,8 +252,21 @@ int entry(int argc, char **argv)
...
+        range_name = values.at(0);
+        range_min = std::atoi(values.at(1).c_str());
+        range_max = std::atoi(values.at(2).c_str());
+      }
+
       // generate random data
-      if (arser["--fixed_seed"])
+      if (range_name == input_node->name())
```
mhs4670go, line 267: (optional) If given name is not exist, printing it as a warning would be helpful.

mhs4670go (dismissed):
LGTM

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -74,6 +76,19 @@ template <> void geneate_random_data<bool>(std::mt19937 &gen, void *data, uint32
...
 }
 
+template <typename T>
+void geneate_random_range(void *data, uint32_t size, int32_t range_min, int32_t range_max)
+{
+  for (uint32_t i = 0; i < size; i++)
+  {
+    // +1 will make value of [range_min, range_max]
+    int32_t range = range_max - range_min + 1;
+    int32_t value = (rand() % range) + range_min;
```
llFreetimell, line 86: Not a problem but just for your information :D

---

Range of `rand()` function is [0, RAND_MAX].
One interesting thing is that RAND_MAX is different between Windows and Linux.

- Windows RAND_MAX =0x7fff = 32767 = 2^15-1
- Linux RAND_MAX = 0x7fffffff = 2147483647 = 2^31-1

So, when I call `generate_random_range(data, size, 0, 40000)` in Windows, I would get following random values.
- Expectation : [0, 40000]
- Actual : [0,32768]

As actual range is included in expectation range, it is not incorrect, just different with our expectation :D

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -74,6 +76,19 @@ template <> void geneate_random_data<bool>(std::mt19937 &gen, void *data, uint32
...
+    static_cast<T *>(data)[i] = static_cast<T>(value);
+  }
+  std::cout << std::endl;
```
llFreetimell, line 89: This is intended code? @.@

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -74,6 +76,19 @@ template <> void geneate_random_data<bool>(std::mt19937 &gen, void *data, uint32
   }
 }
 
+template <typename T>
+void geneate_random_range(void *data, uint32_t size, int32_t range_min, int32_t range_max)
```
llFreetimell, line 80:
```suggestion
void generate_random_range(void *data, uint32_t size, int32_t range_min, int32_t range_max)
```

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -176,6 +212,23 @@ int entry(int argc, char **argv)
...
+  bool range_check = false;
+  bool range_input_found = false;
+  if (arser["--input_range"])
+  {
+    // NOTE limitation: we can only set one input range
+    // TODO expand this for multiple inputs
+    std::vector<std::string> values = arser.get<std::vector<std::string>>("--input_range");
+    assert(values.size() == 3);
+    range_name = values.at(0);
+    range_min = std::atoi(values.at(1).c_str());
```
llFreetimell, line 227: According to https://www.cplusplus.com/reference/cstdlib/atoi/, `atoi` never throws even the string is not a number format.
For example, `std::atoi("1a3")` in Ubuntu returns `1`.
So how about adding an assert like following?

```cpp
bool is_num(std::string num_string)
{
  for (auto ch : num_string)
    if (std::isdigit(ch) == false)
      return false;
  return true;
}

assert(is_num(values.at(1)));
assert(is_num(values.at(2)));
```

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -74,6 +76,19 @@ template <> void geneate_random_data<bool>(std::mt19937 &gen, void *data, uint32
   }
 }
 
+template <typename T>
+void geneate_random_range(void *data, uint32_t size, int32_t range_min, int32_t range_max)
+{
```
llFreetimell, line 81: (Optional) just for safety? :D

```suggestion
{
  assert(range_min <= range_max);
```

compiler/common-artifacts/src/TestDataGenerator.cpp
mhs4670go: @llFreetimell I think throwing an error or returning failure(-1) looks good as well because that kind of value like `1a3` could be given from user, which isn't a bit proper for `assert`.

compiler/common-artifacts/src/TestDataGenerator.cpp
mhs4670go: Taking care of `range_min` and `range_max` should exist instead of `assert` in `random_range*` functions. It could do nothing or throwing an error or something.

compiler/common-artifacts/src/TestDataGenerator.cpp
```diff
@@ -120,6 +151,11 @@ int entry(int argc, char **argv)
     .required(false)
     .nargs(0)
     .help("Put a fixed seed into the random number generator");
+  arser.add_argument("--input_range")
+    .required(false)
+    .nargs(3)
+    .type(arser::DataType::STR_VEC)
+    .help("Set random number range for the input as 'name min max'");
```
mhs4670go, line 158: This could be a bit more helpful. Whether it is closed interval or open one.
```suggestion
    .help("Set random number range [min max] for the input as 'name min max'");
```

compiler/common-artifacts/src/TestDataGenerator.cpp
seanshpark: Ah, for debuging, not removed -_-;
Thanks!

compiler/common-artifacts/src/TestDataGenerator.cpp
seanshpark: I think assert is good to have :)

compiler/common-artifacts/src/TestDataGenerator.cpp
seanshpark: I'll leave as TODO

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

PR #8254, opened 2022-01-05
[res] Introduce Gather_001 recipe

seanshpark:
This will introduce Gather_001 recipe that has both inputs of Gather
op as graph input with axis 3.


seanshpark:
From draft #8245

mhs4670go (approved):
LGTM

PR #8255, opened 2022-01-06
[luci-interpreter] Enable Gather

seanshpark:
This will enable Gather Op for float32 in luci-interpreter.


seanshpark:
For #8241, from draft #8245

seanshpark:
CC @binarman

compiler/luci-interpreter/src/loader/nodes/Gather.cpp
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
```
jinevening, line 2:
```suggestion
 * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
```

compiler/luci-interpreter/src/kernels/Gather.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+#include <tensorflow/lite/kernels/internal/common.h>
+
+#include <stdexcept>
+#include <cassert>
+
+// NOTE code is copied here cause "ruy::profiler::ScopeLabel label("Gather");" exist
+// but "#include "ruy/profiler/instrumentation.h"" line is absent and compile fails.
+// TensorFlow doesn't seem to use tflite::reference_ops::Gather() alone.
+// TODO include header file and remove reference_ops block after this is fixed
```
jinevening, line 29: I guess that using `<tensorflow/lite/kernels/internal/optimized/optimized_ops.h>` will work well.

Plus, I think that a header should be created under `pal/linux` directory, and `KernelsToBuild.lst` needs to be updated. Because luci-interpreter is not used for desktop linux only.

I think `Split.cpp` may be helpful to see the pattern.

seanshpark: > I guess that using <tensorflow/lite/kernels/internal/optimized/optimized_ops.h> will work well.

I'll leave using optimized code to later work.

> Plus, I think that a header should be created under pal/linux directory

Some kernels moved to PAL layer and some still lives in `kernels` folder. I think not all kernels have migrated to PAL layer as of now. For this, I want to leave this task to future work.

jinevening (approved):

seanshpark:
I'll wait a day landing this PR for @binarman to review.

seanshpark:
@jinevening , thanks for reviewing :)

yunjayh (approved):
LGTM

compiler/luci-interpreter/src/kernels/Gather.cpp
binarman: @jinevening
> I think that a header should be created under pal/linux directory

As @seanshpark said, not all kernels have headers in `pal`.

`pal` is used to abstract kernel from platform details, like multi threading, special instructions, temporary buffer sizes(for example im2col), etc.
If kernel is written in plain c/c++ and there are no platform specific stuff, there are no reason to add this kernel in pal.

Also, there could be some operations supported only on a subset of platforms.
In that case these operations should be added in part of `pal/<platform>/KernelsToBuild.lst` files and everything still works =)

P.s. it is not a written rule, but I think it is good to have a fullest set of supported operations for `linux` platform. Other platforms are optional and support could be added later if needed.

PR #8256, opened 2022-01-06
[common-artifacts] Additional options for testDataGenerator

seanshpark:
This will enable to provide additional options for testDataGenerator
command per model.

### DIFF
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -140,6 +140,13 @@ endmacro()
 
 include("exclude.lst")
 
+# TODO revise using variadic arguments
+macro(tcgenerate_option NAME OPTION ARG1 ARG2 ARG3)
+  set(TCGEN_OPT_${NAME} ${OPTION} ${ARG1} ${ARG2} ${ARG3})
+endmacro()
+
+include("options.lst")
+
 foreach(RECIPE IN ITEMS ${RECIPES})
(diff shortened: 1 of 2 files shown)


seanshpark:
From draft #8245

compiler/common-artifacts/CMakeLists.txt
```diff
@@ -274,13 +281,19 @@ foreach(RECIPE IN ITEMS ${RECIPES})
     )
     list(APPEND TEST_DEPS ${TC_DIRECTORY})
 
+    set(ADDITIONAL_OPTIONS )
```
mhs4670go, line 284: Is it intended?
```suggestion
    set(ADDITIONAL_OPTIONS)
```

seanshpark: Ah, yes... Just to set `ADDITIONAL_OPTIONS` as empty string.

seanshpark: maybe add a note?

seanshpark: added a comment;

mhs4670go: (optinal) Maybe empty string can be created like below.
```cmake
set(ADDITIONAL_OPTIONS "")
Or
set(ADDITIONAL_OPTIONS)
```

mhs4670go (approved):
LGTM

Issue #8259, opened 2022-01-07
[luci/pass] Constant folding of Slice, Concat and Squeeze


seanshpark:
There's a sub network with `Slice` having input data constant.
And then there's a `Concat` and then `Squeeze` that will be constant input afterwards.
Lets' do some foldings.

![image](https://user-images.githubusercontent.com/4616940/148474734-ad0cd4bc-fa16-4793-805a-ccfb91d4d278.png)

seanshpark:
(not related but...) as this issue number is 8259 which reminds me of https://en.wikipedia.org/wiki/Intel_8259 -_-;;;

ai-moiseev:
If no one works on this yet, I would like to implement Slice folding part :)

seanshpark:
> If no one works on this yet, I would like to implement Slice folding part :)

Thanks for your interest! Actually I am working on this; #8263

seanshpark:
Found that this pattern is not needed. I was looking at ONNX model and thought it was of tflite.
I think I need some rest... -_-;;;

PR #8263, opened 2022-01-07
DRAFT: Constant folding for Slice, Concat, Squeeze

seanshpark:
on-going draft to enable luci/pass constant folding for Slice, Concat
and Squeeze.


seanshpark:
For #8259

seanshpark:
This is not needed anymore.

PR #8265, opened 2022-01-10
[luci] Support const input quantization of Gather

jyoungyun:
This commit adds support for const input quantization for Gather op.


llFreetimell (approved):
LGTM

seanshpark:
@jyoungyun , this change is OK but you've conflict with my current task of enabling Gather.
See #8245.

jinevening:
The change looks good to me.

In addition to this change, we'll need to add Gather Op to `PropagateQuantParamPass`. The ofm of Gather is a subsample of ifm, so it would be better to make the qparam of ifm same with that of ofm.

jyoungyun:
> @jyoungyun , this change is OK but you've conflict with my current task of enabling Gather.

I saw your changes. Your patches quantize all input params including indices. Do we need to quantize indices?

seanshpark:
> Your patches quantize all input params including indices.

Um? which code are you mentioning?

jyoungyun:
This code is duplicated with #8268. So I close it.

PR #8266, opened 2022-01-10
[circle-execution-plan] Add command line interface and TargetPlatform

BalyshevArtem:
This pr adds command line interface + TargetPlatform structure and it's usage in ExecutionPlanner. This is first part of changes for adding pal to circle-execution-planner. 

draft: #8128 

### DIFF
--- a/compiler/circle-execution-plan/pal/TargetPlatform.h
+++ b/compiler/circle-execution-plan/pal/TargetPlatform.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/compiler/circle-execution-plan/pal/TargetPlatform.h
+++ b/compiler/circle-execution-plan/pal/TargetPlatform.h
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CIRCLE_EXECUTION_PLAN_TARGET_PLATFORM_H
+#define CIRCLE_EXECUTION_PLAN_TARGET_PLATFORM_H
+
+namespace circle_planner
+{
+
+enum SupportedPlatformType
+{
+  LINUX,
+  MCU,
+  CMSISNN
+};
+
+struct TargetPlatform
+{
+  SupportedPlatformType platform_type;
+  bool use_dsp;
+};
+
+} // namespace circle_planner
+
+#endif // CIRCLE_EXECUTION_PLAN_TARGET_PLATFORM_H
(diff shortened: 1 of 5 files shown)


compiler/circle-execution-plan/pal/TargetPlatform.h
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
```
m-bronnikov, line 2: Happy New Year! (Please, use 2022) =)

compiler/circle-execution-plan/src/CircleExecutionPlan.cpp
```diff
@@ -49,6 +61,33 @@ int entry(int argc, char **argv)
 
   std::string input_path = arser.get<std::string>("input");
   std::string output_path = arser.get<std::string>("output");
+  std::string platform_name = arser.get<std::string>("--platform");
+  bool use_dsp = arser.get<bool>("--use_dsp");
```
m-bronnikov, line 65: (optional) Personally, I prefer use `const` as much as possible because it helps to reviewer not expect variable changes bellow of definition. Hovewer this comment is optional because some variables above is not `const` too.

compiler/circle-execution-plan/src/ExecutionPlanner.cpp
```diff
@@ -358,34 +359,42 @@ void ExecutionPlanner::create_alloc_node_inform_vector(bool null_consts, bool nu
...
-        temp_alloc.last_node = i + 1;
-        temp_alloc.node_num = i;
-        temp_alloc.is_temp = true;
-
-        _alloc_node_inform_vector.push_back(temp_alloc);
-        _alloc_node.push_back(i);
-        _dealloc_node.push_back(i);
+        auto conv = loco::must_cast<const luci::CircleConv2D *>(circle_node);
+        scratchpad_size = compute_im2col_size(
+          conv); // TODO: rewrite it with ScratchpadHelper method when it will be added
```
m-bronnikov, line 370: (optional) As far as I know implicit convection is to use own line for comments like this.

m-bronnikov: But I'm not sure =)

compiler/circle-execution-plan/src/ExecutionPlanner.cpp
```diff
@@ -358,34 +359,42 @@ void ExecutionPlanner::create_alloc_node_inform_vector(bool null_consts, bool nu
...
-        temp_alloc.first_node = i - 1;
-        temp_alloc.last_node = i + 1;
-        temp_alloc.node_num = i;
-        temp_alloc.is_temp = true;
-
-        _alloc_node_inform_vector.push_back(temp_alloc);
-        _alloc_node.push_back(i);
-        _dealloc_node.push_back(i);
+        auto conv = loco::must_cast<const luci::CircleConv2D *>(circle_node);
+        scratchpad_size = compute_im2col_size(
```
m-bronnikov, line 369: Since `im2col` uses only with `CircleConv` I think better to handle `null_im2col` here:

```suggestion
        scratchpad_size = null_im2col ? 0 : compute_im2col_size(
```

And then remove redundant `if (null_im2col)` statement bellow.

BalyshevArtem: Done

compiler/circle-execution-plan/src/ExecutionPlanner.h
```diff
@@ -19,6 +19,7 @@
 
 #include <luci/IR/Module.h>
 #include <luci/Plan/CircleNodeExecutionPlan.h>
+#include "TargetPlatform.h"
```
binarman, line 22: let's add headers this way:
```suggestion
#include "TargetPlatform.h"
#include <luci/IR/Module.h>
#include <luci/Plan/CircleNodeExecutionPlan.h>
```

Headers more related to current component should be placed higher, like:
``` c++
#include "header from this component"
#include <header from related component>
#include <header from third party project>
#include <system header>
```

This approach helps to detect issues with incomplete headers. For example cases where local header uses std library, but do not include it properly.

binarman (approved):
overall LGTM, but fix include order, please

compiler/circle-execution-plan/src/ExecutionPlanner.h
BalyshevArtem: Done

m-bronnikov (approved):
LGTM

binarman (approved):

PR #8267, opened 2022-01-10
[common-artifacts] Add node name not empty check

seanshpark:
This will add node name is not empty assert check for all nodes.


seanshpark:
from draft #8245

seanshpark:
@nnfw-bot test nncc-release

mhs4670go (approved):
LGTM

PR #8268, opened 2022-01-10
[luci/pass] Quantize const for Gather

seanshpark:
This will enable quantization for Gather Op.

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1227,6 +1227,7 @@ void quantize_const_inputs(luci::CircleNode *node, loco::DataType output_type)
     case luci::CircleOpcode::ARG_MAX:
     case luci::CircleOpcode::ARG_MIN:
     case luci::CircleOpcode::BATCH_TO_SPACE_ND:
+    case luci::CircleOpcode::GATHER:
     case luci::CircleOpcode::LOCAL_RESPONSE_NORMALIZATION:
     case luci::CircleOpcode::MEAN:
     case luci::CircleOpcode::MIRROR_PAD:
--- a/compiler/luci/pass/src/PropagateQuantParamPass.cpp
+++ b/compiler/luci/pass/src/PropagateQuantParamPass.cpp
@@ -63,6 +63,12 @@ struct PropagateQuantParam final : public luci::CircleNodeMutableVisitor<bool>
 
   bool visit(luci::CircleNode *) { return false; }
 
+  bool visit(luci::CircleGather *node)
+  {
+    auto input_node = loco::must_cast<luci::CircleNode *>(node->params());
+    return copy_qparam(input_node, node);
+  }
+
   bool visit(luci::CircleReshape *node)
   {
     auto input_node = loco::must_cast<luci::CircleNode *>(node->tensor());


seanshpark:
For #8241, from draft #8245

jinevening:
#8265 is the same with this PR, so my review is also the same :) 

---
From https://github.com/Samsung/ONE/pull/8265#issuecomment-1009495954

The change looks good to me.

In addition to this change, we'll need to add Gather Op to PropagateQuantParamPass. The ofm of Gather is a subsample of ifm, so it would be better to make the qparam of ifm same with that of ofm.

jinevening (dismissed):

seanshpark:
> In addition to this change, we'll need to add Gather Op to PropagateQuantParamPass. The ofm of Gather is a subsample of ifm, so it would be better to make the qparam of ifm same with that of ofm.

I'll try to add this in this PR

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1264,6 +1264,7 @@ void quantize_const_inputs(luci::CircleNode *node, loco::DataType output_type)
     case luci::CircleOpcode::EXP:
     case luci::CircleOpcode::FLOOR:
     case luci::CircleOpcode::FLOOR_DIV:
+    case luci::CircleOpcode::GATHER:
```
jyoungyun, line 1267: Here is all const input quantization case.
If we don't need to quantize indices, you can add gather to above case. (Line 1250 case)

llFreetimell, line 1267: If `Gather` is added to this section, all the inputs are used to quantize AFAIK.

```cpp
      // Quantize all const inputs using their values
      for (uint32_t i = 0; i < arity; i++)
      {
        input_node = node->arg(i);
        const_node = dynamic_cast<luci::CircleConst *>(input_node);
        if (const_node != nullptr && !is_quantized(const_node))
          quant_const(const_node, output_type);
      }
```

But, since `Gather` has integer type for `gather->indices()`, I think it would be thrown by following assert.

https://github.com/Samsung/ONE/blob/89f95897ab4864e8bce8b6acdfe4c22a1db9308f/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp#L303

llFreetimell: How about adding `CircleGather` to `VerifyQuantizedNodeGranularity`?
Not exactly but maybe such like following example?

```cpp
  bool visit(const luci::CircleGather *node)
  {
    RETURN_FALSE_UNLESS(is_lwq(node));
    RETURN_FALSE_UNLESS(is_cwq_const(node->params()));
    return true;
  }
```

llFreetimell:
How about adding `CircleGather` to `VerifyQuantizedNodeGranularity`? :D
Not exactly but maybe such like following example?

```cpp
  bool visit(const luci::CircleGather *node)
  {
    RETURN_FALSE_UNLESS(is_lwq(node));
    RETURN_FALSE_UNLESS(is_cwq_const(node->params()));
    RETURN_FALSE_UNLESS(is_lwq(node->indices()));
    return true;
  }
```

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
seanshpark: @jyoungyun , @llFreetimell , there is `if (const_node != nullptr && !is_quantized(const_node))` for each input.. that will skip for `indices()` as I understand. Did I miss something?
```cxx

bool is_quantized(const CircleNode *node)
{
  return node->quantparam() != nullptr &&
         (node->dtype() == loco::DataType::U8 ||  // activation, weight (uint8 quant)
          node->dtype() == loco::DataType::S16 || // activation, weight (int16 quant)
          node->dtype() == loco::DataType::S32 || // bias (uint8 quant)
          node->dtype() == loco::DataType::S64);  // bias (int16 quant)
}
```

seanshpark: Ah, got it! thanks!

seanshpark: I've moved the `case` line that checks only for arg(0)

jinevening: You've moved Greater, not Gather.

seanshpark:
> After I saw your draft code, I couldn't find CircleGather in VerifyQuantizedNodeGranularity.

I think you can add the change in another PR.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
llFreetimell: I thought that if `indices()` is `CircleConst` with `S32`,

- `const_node != nullptr` : true
- `node->quantparam()` != nullptr : false

So I thought that `assert` would throw... T.T
But anyway this is resolved now..! :D

jinevening: > Did I miss something?

At this point, Const input of Gather is not yet quantized (quantparam() == nullptr, so `is_quantized` returns false).

BTW, Sorry for the wrong review. I thought that this PR is the same with #8265 , but it wasn't.

seanshpark: 😭

llFreetimell (approved):

jinevening (approved):
LGTM Thank you

jyoungyun (approved):
LGTM Thank you!!!

PR #8269, opened 2022-01-10
[common-artifacts] Enable Gather test

seanshpark:
This will enable Gather Op test.


seanshpark:
For #8241, from draft #8245

seanshpark:
@mhs4670go , PTAL

mhs4670go (approved):
LGTM

compiler/common-artifacts/options.lst
```diff
@@ -2,4 +2,5 @@
 
 #[[ tcgenerate_option : add additional option(s) for generation ]]
 
-# TODO add options
+# make valid 'indices' input value
+tcgenerate_option(Gather_001 --input_range indices 0 3)
```
mhs4670go, line 6: `Gather_000` isn't needed?

seanshpark: 000 has constant incices with explicit values.
```
operand {
  name: "indices"
  type: INT32
  shape { dim: 2 }
  filler { tag: "explicit" arg: "1" arg: "2" }
}
```

PR #8271, opened 2022-01-12
[circle-execution-plan] Add ScratchpadHelper and it's usage

BalyshevArtem:
This pr adds ScratchpadHelpers for linux, mcu and cmsisnn platforms and it's usage.

draft: #8128

### DIFF
--- a/compiler/circle-execution-plan/pal/IScratchpadHelper.h
+++ b/compiler/circle-execution-plan/pal/IScratchpadHelper.h
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CIRCLE_EXECUTION_PLAN_ISRCRATCHPAD_HELPER_H
+#define CIRCLE_EXECUTION_PLAN_ISRCRATCHPAD_HELPER_H
+
+namespace circle_planner
+{
+
+class IScratchpadHelper
+{
+public:
+  virtual uint32_t ComputeScratchpadSizeConv2d(const luci::CircleConv2D *conv) = 0;
+
+  virtual ~IScratchpadHelper() = default;
+};
+
+} // namespace circle_planner
+
+#endif // CIRCLE_EXECUTION_PLAN_ISRCRATCHPAD_HELPER_H
--- a/compiler/circle-execution-plan/pal/IScratchpadHelper.h
+++ b/compiler/circle-execution-plan/pal/IScratchpadHelper.h
@@ -17,6 +17,9 @@
 #ifndef CIRCLE_EXECUTION_PLAN_ISRCRATCHPAD_HELPER_H
 #define CIRCLE_EXECUTION_PLAN_ISRCRATCHPAD_HELPER_H
 
+#include <luci/IR/Nodes/CircleConv2D.h>
+#include <cstdint>
+
 namespace circle_planner
 {
 
--- a/compiler/circle-execution-plan/pal/ScratchpadHelperCMSISNN.h
+++ b/compiler/circle-execution-plan/pal/ScratchpadHelperCMSISNN.h
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CIRCLE_EXECUTION_PLAN_SCRATCHPAD_HELPER_CMSISNN_H
+#define CIRCLE_EXECUTION_PLAN_SCRATCHPAD_HELPER_CMSISNN_H
+
+#include "IScratchpadHelper.h"
+
+namespace circle_planner
+{
+
+namespace
+{
+
+inline int32_t computePadding(int32_t stride, int32_t dilation_rate, int32_t in_size,
+                              int32_t filter_size, int32_t out_size)
+{
+  const int32_t effective_filter_size = (filter_size - 1) * dilation_rate + 1;
+  const int32_t padding = ((out_size - 1) * stride + effective_filter_size - in_size) / 2;
+  return padding > 0 ? padding : 0;
+}
+
+} // namespace
+
+class ScratchpadHelperCMSISNN : public IScratchpadHelper
+{
+public:
+  explicit ScratchpadHelperCMSISNN(bool use_dsp) : _use_dsp(use_dsp)
+  {
+    // Do nothing
+  }
+
+  uint32_t ComputeScratchpadSizeConv2d(const luci::CircleConv2D *conv) final
+  {
+    const auto dilation_height_factor = static_cast<int32_t>(conv->dilation()->h());
+    const auto dilation_width_factor = static_cast<int32_t>(conv->dilation()->w());
+
+    auto conv_input = loco::must_cast<luci::CircleNode *>(conv->input());
+    auto filter = loco::must_cast<luci::CircleNode *>(conv->filter());
+
+    if (dilation_width_factor != 1 or dilation_height_factor != 1 or
+        conv_input->dtype() != loco::DataType::S8)
+      return 0;
+
+    const auto input_depth = static_cast<int32_t>(conv_input->dim(3).value());
+
+    const auto input_height = static_cast<int32_t>(conv_input->dim(1).value());
+    const auto input_width = static_cast<int32_t>(conv_input->dim(2).value());
+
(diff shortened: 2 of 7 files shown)


compiler/circle-execution-plan/src/ExecutionPlanner.h
```diff
@@ -20,6 +20,10 @@
 #include "TargetPlatform.h"
 #include <luci/IR/Module.h>
 #include <luci/Plan/CircleNodeExecutionPlan.h>
+#include "IScratchpadHelper.h"
+#include "ScratchpadHelperLinux.h"
+#include "ScratchpadHelperMCU.h"
+#include "ScratchpadHelperCMSISNN.h"
```
BalyshevArtem, line 26: for reviewers:
I tried to place these includes 
```c++
#include "IScratchpadHelper.h"
#include "ScratchpadHelperLinux.h"
#include "ScratchpadHelperMCU.h"
#include "ScratchpadHelperCMSISNN.h"
```
 higher than includes with `<...>`, but in this case, the compiler stops seeing what is (for example) `uint32_t` or `luci::CircleConv2D` and smth else.
I found that this is the most simple variant for me to solve this problem, but if there are another solutions, I will be glad to add them.

compiler/circle-execution-plan/pal/ScratchpadHelperMCU.h
```diff
@@ -0,0 +1,37 @@
...
+#include "IScratchpadHelper.h"
+
+namespace circle_planner
+{
+
+class ScratchpadHelperMCU : public IScratchpadHelper
+{
+public:
+  uint32_t ComputeScratchpadSizeConv2d(const luci::CircleConv2D *conv) final
+  {
```
binarman, line 29: just in case we add some checkers...
```suggestion
  {
    (void)conv;
```

compiler/circle-execution-plan/src/ExecutionPlanner.h
binarman: This is exactly why we try to place headers as high as possible =)
This error means that `IScratchpadHelper` and related headers are not "self-sufficient", i.e. they use something that is not defined/included in them.

For this particular error with `uint32_t` you need to add `#include <cstdint>` in `IScratchpadHelper.h`

I'll check if there are other problems in few minutes

compiler/circle-execution-plan/pal/IScratchpadHelper.h
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CIRCLE_EXECUTION_PLAN_ISRCRATCHPAD_HELPER_H
+#define CIRCLE_EXECUTION_PLAN_ISRCRATCHPAD_HELPER_H
+
```
binarman, line 19: This helps to include headers in proper order
```suggestion

#include <luci/IR/Nodes/CircleConv2D.h>
#include <cstdint>

```

compiler/circle-execution-plan/pal/ScratchpadHelperCMSISNN.h
```diff
@@ -0,0 +1,98 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CIRCLE_EXECUTION_PLAN_SCRATCHPAD_HELPER_CMSISNN_H
+#define CIRCLE_EXECUTION_PLAN_SCRATCHPAD_HELPER_CMSISNN_H
+
+#include "IScratchpadHelper.h"
```
binarman, line 20: For reordering headers
```suggestion
#include "IScratchpadHelper.h"
#include <cassert>
```

compiler/circle-execution-plan/pal/ScratchpadHelperLinux.h
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CIRCLE_EXECUTION_PLAN_SCRATCHPAD_HELPER_LINUX_H
+#define CIRCLE_EXECUTION_PLAN_SCRATCHPAD_HELPER_LINUX_H
+
+#include "IScratchpadHelper.h"
```
binarman, line 20: For reordering headers
```suggestion
#include "IScratchpadHelper.h"
#include <loco/IR/DataTypeTraits.h>
```

compiler/circle-execution-plan/pal/ScratchpadHelperMCU.h
seanshpark: maybe this?
```diff
- uint32_t ComputeScratchpadSizeConv2d(const luci::CircleConv2D *conv) final
+ uint32_t ComputeScratchpadSizeConv2d(const luci::CircleConv2D *) final
```

compiler/circle-execution-plan/pal/ScratchpadHelperCMSISNN.h
```diff
@@ -0,0 +1,103 @@
...
+
+namespace
+{
+
+inline int32_t computePadding(int32_t stride, int32_t dilation_rate, int32_t in_size,
+                              int32_t filter_size, int32_t out_size)
+{
+  const int32_t effective_filter_size = (filter_size - 1) * dilation_rate + 1;
+  const int32_t padding = ((out_size - 1) * stride + effective_filter_size - in_size) / 2;
+  return padding > 0 ? padding : 0;
```
m-bronnikov, line 34: I think better to leave comment about which padding computes by this code:

Example: in case of `stride=1`, `in_size=out_size` ('SAME' padding), `filter_size=2`, `dilation_rate=1` this function will return `0`, but padding of bottom and right sides exist (and equal to `1`). So, I think better to leave comment that this helper computes correct padding for `left`/`top` side.

compiler/circle-execution-plan/pal/ScratchpadHelperCMSISNN.h
```diff
@@ -0,0 +1,103 @@
...
+    const auto padding_height = computePadding(stride_height, dilation_height_factor, input_height,
+                                               filter_height, output_height);
+    const auto padding_width =
+      computePadding(stride_width, dilation_width_factor, input_width, filter_width, output_width);
+
+    if ((padding_width == 0) && (padding_height == 0) && (input_depth % 4 == 0) &&
+        (stride_width == 1) && (stride_height == 1) && (filter_width == 1) && (filter_height == 1))
+    {
+      return 0;
+    }
```
m-bronnikov, line 87: (optional) I know this code just moved, but there is not clear to reviewer (e.g. to me) why scratchpad is not used if this condition is satisfied. So, could you please leave some comment here?

m-bronnikov: Optional because it may be better to do this as separated PR.

compiler/circle-execution-plan/pal/ScratchpadHelperCMSISNN.h
```diff
@@ -0,0 +1,103 @@
...
+
+    assert(conv_input->quantparam()->zerop.size() == 1);
+    assert(conv->quantparam()->zerop.size() == 1);
+
+    const auto padding_height = computePadding(stride_height, dilation_height_factor, input_height,
+                                               filter_height, output_height);
+    const auto padding_width =
+      computePadding(stride_width, dilation_width_factor, input_width, filter_width, output_width);
+
+    if ((padding_width == 0) && (padding_height == 0) && (input_depth % 4 == 0) &&
```
m-bronnikov, line 83: Q. As I commented above (helper for padding), there is exist case when `top_padding = 0`, `bottom_padding=1`, `left_padding=0`, `right_padding=1` and `computePadding` helper will set `padding_width` and `padding_height` to `0`. Does this behavior is correct for this check?

binarman: I've looked at cmsis code and looks like it uses only left-top paddings, so this code is fine.

compiler/circle-execution-plan/pal/ScratchpadHelperCMSISNN.h
binarman: This logic is based on https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/NN/Source/ConvolutionFunctions/arm_convolve_wrapper_s8.c#L114

``` c++
    if ((conv_params->padding.w == 0) && (conv_params->padding.h == 0) && (input_dims->c % 4 == 0) &&
        (conv_params->stride.w == 1) && (conv_params->stride.h == 1) && (filter_dims->w == 1) &&
        (filter_dims->h == 1) && (conv_params->dilation.w == 1 && conv_params->dilation.h == 1))
    {
        // in implementation of our target cmsis version this function returns 0
        return arm_convolve_1x1_s8_fast_get_buffer_size(input_dims);
    }
    else if ((output_dims->h == 1) && (input_dims->h == 1) && (filter_dims->h == 1) && (output_dims->w % 4 == 0) &&
             (input_dims->n == 1) && (conv_params->dilation.w == 1 && conv_params->dilation.h == 1))
    {
        return arm_convolve_1_x_n_s8_get_buffer_size(input_dims, filter_dims);
    }
    else
    {
        return arm_convolve_s8_get_buffer_size(input_dims, filter_dims);
    }
```

binarman: addition: dilation is not supported in current cmsis nn kernels =_=

compiler/circle-execution-plan/pal/ScratchpadHelperCMSISNN.h
m-bronnikov: > I've looked at cmsis code and looks like it uses only left-top paddings, so this code is fine.

Probably, you are right. Code for padding computation: 
https://github.com/ARM-software/CMSIS_5/blob/1cdd0e824db7340b543a7806e9a73244e73c675e/CMSIS/NN/Tests/UnitTest/generate_test_data.py#L339-L355

m-bronnikov (approved):
Code looks good for me.

seanshpark (approved):
LGTM

binarman (approved):
LGTM

Issue #8273, opened 2022-01-13
[luci] Fix SLICE shape inference with non-constant inputs


seanshpark:
SLICE Op may have non-constant inputs but current shape inference of CircleSlice only supports for constant inputs.
Let's fix this problem.

seanshpark:
From https://github.com/Samsung/ONE/issues/8217#issuecomment-1001882497

PR #8219, opened 2021-12-24
DRAFT import pytorch/onnx LSTM

seanshpark:
on-going draft to import pytorch/onnx LSTM Op.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #8217, opened 2021-12-24
[one-cmds] pythorch (or ONNX) LSTM op import fails


seanshpark:
With current version of other packages, import LSTM op of ONNX model fails (from pytorch)

Our pytorch example also fails and we can use this example for tracing.

```
$ cd res/PyTorchExamples
$ python3 ptem.py LSTM
```
> PyTorch version= 1.10.0+cpu
ONNX version= 1.10.1
ONNX-TF version= 1.9.0
TF version= 2.6.0
...
> loc("lstm_bias_lstm_0"): error: is not immutable, try removing mutable variables in your model since mutable variables are currently not supported through this converter

seanshpark:
```
$ cd res/PyTorchExamples
$ python3 ptem.py LSTM
```
> loc("lstm_bias_lstm_0"): error: is not immutable, try removing mutable variables in your model since mutable variables are currently not supported through this converter

seanshpark:
LSTM.onnx

![image](https://user-images.githubusercontent.com/4616940/147321857-5c7806d5-9433-4241-bc31-c33061af99d4.png)

seanshpark:
Seems `converter.experimental_enable_resource_variables = True` will convert to tflite without error.

seanshpark:
`tflite2circle` fails: `tflite::BuiltinOperator_CALL_ONCE`

seanshpark:
These need attention. @lemmaa , do you have any comments for these?
```
[0] CALL_ONCE (code: 129, dep_code: 127, version: 1)
[1] CUSTOM(VarHandleOp) (code: 32, dep_code: 32, version: 1)
[3] WHILE (code: 119, dep_code: 119, version: 1)
[5] CUSTOM(AssignVariableOp) (code: 32, dep_code: 32, version: 1)
[11] CUSTOM(ReadVariableOp) (code: 32, dep_code: 32, version: 1)
```

seanshpark:
Or maybe try follow `try removing mutable variables` in `onnx-tf` 😭

seanshpark:
Current LSTM.tflite 
![image](https://user-images.githubusercontent.com/4616940/147328925-34382eb3-8948-4e00-85bd-47781d550822.png)

seanshpark:
Export to TF (2.6) model warnings
[...]

Issue #5683, opened 2021-01-13
[Compiler][luci] Upgrade logic of shape/type inference algorithm


llFreetimell:
## What?

Let's re-implement(upgrade) shape and type inference algorithm for following things!

- Shape inference
  - Add testcases which include dynamic shape
  - Support shape inference with unknown dimensions
  - Reduce code duplication by grouping  similar logic as a helper function
- Type inference
  - Reduce code duplication by grouping similar logic as a helper function

## How

1. Select an operation/operations to implement and write your name in the list of table at **assignee**.
    - You can select multiple operations which logic is similar so that PR can be reviewed at once. (ex: ReLU and ReLU6)
    - **Only reserve as much as you can finish in one week. DO NOT BE GREEDY.**
1. Implement each operations as following rules
    - Shape Inference
        - Write new recipes with dynamic shape, which are enough to cover most of algorithm codes. (For code coverage)
        - Add the recipes to `common-artifacts/exclude.lst`. (TestDataGenerator not supports yet)
        - Add the recipes to `res/TensorFlowLiteRecipes/` or `res/CircleRecipes`.
        - Upgrade shape inference algorithm with considering unknown dimension. (Group similar algorithms as helper function)
    - Type Inference
        - Move original type inference algorithm. (Group similar algorithms as helper function)
1. Create PR(s). Then write the PR number in the table at
    - **PR(exclude)**, for adding recipe to `common-artifacts/exclude.lst`. If not needed, write :no_entry_sign: ans skip it.
    - **PR(res)**, for adding recipe to `res/`. If not needed, write :no_entry_sign: ans skip it.
    - **PR(rule)**, for inference algorithm implementation
1. After PR is merged, write :heavy_check_mark: at **Status** in the table.

### Examples (Guide)

TBD

### Background

This issue is originated from #5501. Please refer to the [comment in that issue](https://github.com/Samsung/ONE/issues/5501#issue-775808564).

llFreetimell:
### Operation Table (A-C)

:no_entry_sign: : Not needed
:heavy_check_mark: : Done

| Operation                  | Assignee               | PR(exlucde) | PR(res) | PR(algorithm) | Status |
|----------------------------|------------------------|----------|-------------|------|--------|
| Abs |      |      |      |      |      |
| Add |      |      |      |      |      |
| AddN |      |      |      |      |      |
| ArgMax |      |      |      |      |      |
| ArgMin |      |      |      |      |      |
[...]

PR #8274, opened 2022-01-13
[luci] Revise shape inference for CircleSlice

seanshpark:
This will revise shape inference for CircleSlice Op to skip non-constant inputs.

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.cpp
@@ -1312,8 +1312,13 @@ loco::NodeShape infer_slice(const luci::CircleSlice *node)
 
   auto input_shape = luci::shape_get(node->input()).as<loco::TensorShape>();
 
-  auto const_begin = loco::must_cast<luci::CircleConst *>(node->begin());
-  auto const_size = loco::must_cast<luci::CircleConst *>(node->size());
+  auto const_begin = dynamic_cast<luci::CircleConst *>(node->begin());
+  auto const_size = dynamic_cast<luci::CircleConst *>(node->size());
+  if (const_begin == nullptr || const_size == nullptr)
+  {
+    // NOTE shape depends on values of begin and size
+    return use_own(node);
+  }
 
   loco::TensorShape output_shape;
   std::vector<int64_t> vect_begin; // to hold both S32/S64, we use int64_t


seanshpark:
For #8273, from draft #8219

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -1312,8 +1312,13 @@ loco::NodeShape infer_slice(const luci::CircleSlice *node)
...
   auto input_shape = luci::shape_get(node->input()).as<loco::TensorShape>();
 
-  auto const_begin = loco::must_cast<luci::CircleConst *>(node->begin());
-  auto const_size = loco::must_cast<luci::CircleConst *>(node->size());
+  auto const_begin = dynamic_cast<luci::CircleConst *>(node->begin());
+  auto const_size = dynamic_cast<luci::CircleConst *>(node->size());
+  if (const_begin == nullptr || const_size == nullptr)
+  {
+    // NOTE shape depends on values of begin and size
+    return use_own(node);
```
llFreetimell, line 1320: This case seems to be dynamic shape.

If we don't know `begin()` or `size()`, we cannot decide each dimension values.
However, there is one thing we can decide, which is rank :D

For example, let's assume following situation.
- `input()`, `begin()`, `size()` are not `CircleConst`
- shape of  `input()` : [1, 2, 3, 4]
- shape of  `begin()` : [4]
- shape of `size()` : [4]
Then, shape of `CircleSlice` would be [?, ?, ?, ?].
This can be represented as following NodeShape.
```cpp
  loco::TensorShape output_shape;
  output_shape.rank(4);
  output_shape.dim(0).unset();
  output_shape.dim(1).unset();
  output_shape.dim(2).unset();
  output_shape.dim(3).unset();
```

In conclusion, current PR can be updated as following.
```suggestion
    loco::TensorShape output_shape;
    output_shape.rank(input_shape.rank());
    for (uint32_t i = 0;i < output_shape.rank(); ++i)
      output_shape.dim(i).unset();
    return output_shape;
```

llFreetimell:
Would you add testcase for dynamic shape inference in `luci/service/src/Nodes/CircleSlice.test.cpp`?
Following is one of the example!

```cpp
#include "luci/Service/CircleShapeInference.h"

TEST(ShapeInferenceTest, Slice_dynamic)
{
  luci::sinf::Algorithm shape_inference_algo;

  auto g = loco::make_graph();
  auto slice_input = g->nodes()->create<luci::CircleRelu>();
  auto slice_begin = g->nodes()->create<luci::CircleRelu>();
  auto slice_size = g->nodes()->create<luci::CircleRelu>();
  auto slice = g->nodes()->create<luci::CircleSlice>();

  slice_input->shape({1,2,3,4});
  slice_input->shape_status(luci::ShapeStatus::VALID);
  slice_begin->shape({4});
  slice_begin->shape_status(luci::ShapeStatus::VALID);
  slice_size->shape({4});
  slice_size->shape_status(luci::ShapeStatus::VALID);

  slice->input(slice_input);
  slice->begin(slice_begin);
  slice->size(slice_size);

  auto output_shape = slice->accept(&shape_inference_algo);

  ASSERT_FALSE(output_shape.dim(0).known());
  ASSERT_FALSE(output_shape.dim(1).known());
  ASSERT_FALSE(output_shape.dim(2).known());
  ASSERT_FALSE(output_shape.dim(3).known());
}
```

compiler/luci/service/src/CircleShapeInferenceRule.cpp
llFreetimell: FYI, this is why `Slice` in [initial model](https://github.com/Samsung/ONE/issues/8217#issuecomment-1011694606) has [1, 1, 1] :D
If we dump this with `tfldump`, `shape_signature` may have `[-1, -1, -1]`!

seanshpark:
```
    for (uint32_t i = 0;i < output_shape.rank(); ++i)
      output_shape.dim(i).unset();
    return output_shape;
```
This will remove current shape info read from the file.
I don't think we have any other nodes do this. Or is there?
Using ` return use_own(node);` was to preserve what was from the file.
Can you explain more about why unset dim values?

seanshpark: > If we dump this with tfldump, shape_signature may have [-1, -1, -1]!

I'll check this

seanshpark: Your are right!
```
T(2:35) FLOAT32 (1, 1, 1) (-1, -1, -1) B(67) LSTM_2df696de/rnn/while/TensorArrayV2Write/TensorListSetItem6
T(2:38) FLOAT32 (1, 1, 1) (-1, -1, -1) B(70) LSTM_2df696de/rnn/while/TensorArrayV2Write/TensorListSetItem9
```

llFreetimell: > I don't think we have any other nodes do this. Or is there?

I had a plan with #5683 but nothing done yet because it was not that urgent and no real use case...

> Can you explain more about why unset dim values?

`unset()` makes the dimension `unknown`.
i.e., `output_shape.dim(0).unset()` means `[?]`
https://github.com/Samsung/ONE/blob/ca224b4a329382bf146c06b2c312881f238cdfd4/compiler/loco/include/loco/IR/Dimension.h#L59-L63

seanshpark: There's a problem: `Optimized graph is invalid` is dropped.
output node shape vs graph output shape differes: `[1,?,?] vs [5,3,20]`

seanshpark:
> There's a problem: Optimized graph is invalid is dropped.
output node shape vs graph output shape differes: [1,?,?] vs [5,3,20]

@llFreetimell , can you guess what may be the problem of this?
(1) As we don't eval the network, we can't get calcuate correct shape
(2) some nodes needs fix
(3) ?

There's a comment in `own_shape`
```cxx
loco::TensorShape own_shape(const luci::CircleNode *node)
{
  loco::TensorShape shape;
  shape.rank(node->rank());
  for (uint32_t r = 0; r < node->rank(); ++r)
  {
    // Shape inference rules in this file did not consider unknown dimension.
    // If some node has unknown dimension, 0 is inserted and wrong shape
    // inference was done as a result.
    // To fix this, new shape inference algorithm is being implemented.
    // Until new inference algorithm is fully implemented, unknown dimension
    // would be represented as 1 along with TFLite expression.
    shape.dim(r) = node->dim(r).known() ? node->dim(r).value() : 1;
  }
  return shape;
}
```

is this comment still valid? that is do we need more changes?

compiler/luci/service/src/CircleShapeInferenceRule.cpp
llFreetimell: Looking the graph but I cannot understand why `concat( [?,?,?], [?,?,?], [1,3,20] ) = [5, 3, 20]`.... -_-;;

I will look more the model with https://github.com/Samsung/ONE/issues/8217#issuecomment-1011860364...

llFreetimell:
> can you guess what may be the problem of this?

First we should investigate why [5, 3, 20] is inferred...
- If the original network is something wrong, we should update the incorrect shape to our inferred shape.
  - https://github.com/Samsung/ONE/blob/ca224b4a329382bf146c06b2c312881f238cdfd4/compiler/luci/pass/src/CircleShapeInferencePass.test.cpp#L187
  - When operation has wrong shape, it is updated by our shape. But if graph output has wrong shape, nothing done for now.
- If the original network is correct using other hints, we may add the hints.

> is this comment still valid? that is do we need more changes?

This is still valid..!

- Importing dynamic shape : Done
- Handling dynamic shape : Not done
  - Let's assume there is a shape [1, 2, ?, 4]
  - Before importing dynamic shape : [1, 2, 1, 4]
  - After importing dynamic shape : [1, 2, ?, 4] 
  - If we do not add `shape.dim(r) = node->dim(r).known() ? node->dim(r).value() : 1;`, `dim(3).value()` returns 0, not 1.
  - It breaks some of shape inference logic

seanshpark:
> Handling dynamic shape : Not done 

OK got it. Recurrent may needs this.

llFreetimell:
I am looking the model, but before that, there is one way to get a model quickly.
Using `--disable_validation`....! 
(You may know already but just for information... T.T)

seanshpark:
> Using --disable_validation....!

Yup :) I wasn't in a hurry for the model, well anyway with this option got this; which is also strange.

image

seanshpark:
I think it would be better to continue with new shape inference. closing :)

PR #8275, opened 2022-01-13
[luci] Revise shape inference for CircleConcatenation

seanshpark:
This will revise shape inference for CircleConcatenation to drop
warnings and use own shape if the dim conditions don't match.

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.cpp
@@ -564,8 +564,19 @@ loco::NodeShape infer_concatenation(const luci::CircleConcatenation *node)
         output_shape.dim(j) = output_shape.dim(j).value() + input_shape.dim(j).value();
       }
       else
-        assert(!output_shape.dim(j).known() || !input_shape.dim(j).known() ||
-               output_shape.dim(j) == input_shape.dim(j));
+      {
+        if (output_shape.dim(j).known() && input_shape.dim(j).known() &&
+            output_shape.dim(j).value() != input_shape.dim(j).value())
+        {
+          LOGGER(l);
+          WARN(l) << "WARNING: Concat inputs have different dim values at axis " << j << ": "
+                  << output_shape.dim(j).value() << " vs " << input_shape.dim(j).value()
+                  << std::endl;
+          // NOTE if the input is determined at runtime, we cannot determine without running it
+          //      LSTM body has Slice-Concat sequence where Slice input is non-constant.
+          return use_own(node);
+        }
+      }
     }
   }
 


seanshpark:
For case of https://github.com/Samsung/ONE/issues/8217#issuecomment-1011694606

seanshpark:
@llFreetimell , I'm not fully confident of this change. Please share your opinion.

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -564,8 +564,19 @@ loco::NodeShape infer_concatenation(const luci::CircleConcatenation *node)
...
+        if (output_shape.dim(j).known() && input_shape.dim(j).known() &&
+            output_shape.dim(j).value() != input_shape.dim(j).value())
+        {
+          LOGGER(l);
+          WARN(l) << "WARNING: Concat inputs have different dim values at axis " << j << ": "
+                  << output_shape.dim(j).value() << " vs " << input_shape.dim(j).value()
+                  << std::endl;
+          // NOTE if the input is determined at runtime, we cannot determine without running it
+          //      LSTM body has Slice-Concat sequence where Slice input is non-constant.
+          return use_own(node);
```
llFreetimell, line 577: If this were some kind of workaround, it may be no problem and I could agree.
However, this breaks the most important rule for `CircleConcatenation`, which is "Dimension other than axis should be same".
And actually, we should have negative TC for this assert... T.T

---

I think this is related with #8274.
If #8274 accepts dynamic shape inference rule, I think this PR can be resolved because current assert is considering about dynamic shape :)

seanshpark: > some kind of workaround

As of now, yes.. I'm not fully aware of what you were doing (or finished?) with shape inference, so I'm not confident final form of share inference.

OK, I'll wait for #8274 conclusion.

llFreetimell: > As of now, yes..

For me, this was not workaround because... T.T
- Workaround
  - **(before)Correct thing -> (after)Correct thing**
  - (before)Incorrect thing -> (after)Consider this as correct for now
- This PR
  - **(before)Correct thing -> (after)Incorrect thing**
  - (before)Incorrect thing -> (after)Consider this as correct for now

If `CircleConcatenation`(axis=1) has two inputs with `[1, 2, 3, 4]` and `[1, 3, 2, 4]`,
it should be thrown but this PR makes this not to throw.

---

> I'm not fully aware of what you were doing (or finished?) with shape inference

It's totally my fault sorry :(
- Removing `loco::shape` and use `luci::shape` : Finished
- Updating shape inference logic with dynamic shape : NYI (#5683)
- Structure refactoring : TBD

So there is no problem to modify shape inference logic :)
My comment was only about the logic itself :D

seanshpark: > it should be thrown but this PR makes this not to throw.

Yes, this is a problem. Currently I'm not sure how to fix this problem.

seanshpark: From https://github.com/Samsung/ONE/pull/8274#issuecomment-1011872673, we may need to implement dynamic shape inference for all other nodes? Do you expect we can resolve this issue from it?

seanshpark:
I think we need to continue on new shape inference. closing :)

PR #8277, opened 2022-01-14
[luci/import] Introduce NodeBuilder

m-bronnikov:
This commit introduces NodeBuilder class for nodes which not handled by common builders.

--------------

For: #8042
Draft: #8190

### DIFF
--- a/compiler/luci/import/src/GraphBuilderRegistry.cpp
+++ b/compiler/luci/import/src/GraphBuilderRegistry.cpp
@@ -162,12 +162,12 @@ GraphBuilderRegistry::GraphBuilderRegistry()
   // BuiltinOperator_HARD_SWISH = 117,
   // BuiltinOperator_DENSIFY = 124,
 
-  // Register builders for virtual nodes which not handles in builders registered above.
-#define CIRCLE_VNODE(OPCODE, CLASS) add(std::make_unique<CLASS>())
+  // Register builders for nodes which not handles in builders registered above.
+#define CIRCLE_NODE(CLASS) add(std::make_unique<CLASS>())
 
   // TODO add ops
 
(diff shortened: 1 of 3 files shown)


compiler/luci/import/src/GraphBuilderRegistry.cpp
```diff
@@ -161,6 +161,13 @@ GraphBuilderRegistry::GraphBuilderRegistry()
   // BuiltinOperator_ARG_MAX = 56,
   // BuiltinOperator_HARD_SWISH = 117,
   // BuiltinOperator_DENSIFY = 124,
+
+  // Register builders for virtual nodes which not handles in builders registered above.
+#define CIRCLE_VNODE(OPCODE, CLASS) add(std::make_unique<CLASS>())
```
seanshpark, line 166: Q) I assume `V` is from `Virtual` like other macros. But do we need to name it here too?

seanshpark: > Register builders for virtual nodes which not handles in builders registered above.

This comment may need revision in the future... For now I'm not sure how to...
- Because we cannot generally express as like, but I think in certain places for certain nodes.

I can think of this; 
```cxx
  #define CIRCLE_NODE(CLASS) add(std::make_unique<CLASS>())
```
Removed `V` as actually virtual or not doesn't seem to be important
Removed OPCODE as it's not used.
- This file uses `CIRCLE_NODE` but we do not use `circlenodes.lst` here.
- Maybe using `CIRCLE_NODE` in this file gives confusion with `circlenodes.lst`

seanshpark (approved):
LGTM thank you!

binarman (approved):
LGTM

PR #8281, opened 2022-01-16
[luci/lang] Sort nodes for multi-output

seanshpark:
This will sort node in header/lst files in alphabetical order.


mhs4670go (commented):
LGTM

mhs4670go (approved):

PR #8279, opened 2022-01-14
[DRAFT] Support SVDF Op

seanshpark:
on-going draft to support SVDF Op in compiler frontend.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #8231, opened 2021-12-28
[Draft] Enable SVDF op

BalyshevArtem:
This draft enables SVDF op. 

Related issue #8203

ONE-DCO-1.0-Signed-off-by: Artem Balyshev a.balyshev@partner.samsung.com

Issue #8203, opened 2021-12-22
Enable SVDF Op


BalyshevArtem:
## What?
Let's support svdf operation.
## Why?
One of the benchmarks for tflite micro is the following model: 

[model.zip](https://github.com/Samsung/ONE/files/7762412/model.zip)

![keyword_model](https://user-images.githubusercontent.com/43214667/147089001-6e5dc3d7-c091-44e1-89fe-971f25a4f58b.png)

At this moment we cannot run this model on the `luci-interpreter`. One of the problems is the presence of an SVDF operation that needs to be supported. 
To understand better you can see this issue #7598 and this comment https://github.com/Samsung/ONE/issues/7598#issuecomment-916384853

## How?
First of all we need support this parts:
- [x] tflite2circle #8204
- [x] common-artifacts add item to exclude new recipes #8282
- [x] res/TensorFlowLiteRecipes #8285
- [x] tflchef #8284
- [x] tfldump and circledump #8286 #8287
- [x] luci/lang  #8291
- [x] luci/import #8292 
- [x] luci/service
- [ ] luci/pass
- [x] luci/log, luci/logex
- [x] luci/tests addread
- [x] luci/exports
- [x] luci/tests addwrite
- [x] luci-interpreter
- [x] luci-interpreter/pal
- [x] luci/partition
- [ ] common-artifacts remove items added

BalyshevArtem:
+cc @binarman

seanshpark:
Please prepare a full working draft PR for this new Op to verify it's working
- perpare tflchef recipe(s) for "testing" in `TensorFlowLiteRecipes` like other Ops.
- revise `compiler/tflchef` to produce tflite file from recipe file.

`compiler/common-artifacts` will fail if any module may fail, when you add the recipe file(s)

llFreetimell:
@BalyshevArtem 

I am refactoring `luci/export` with #8198.
It may affect your (draft)code for `luci/export`.
I will keep an eye on this issue and let you know if there are something to be updated :)
Meanwhile, I will finish #8198 ASAP!

binarman:
It is hard to manipulate model from top comment, because it does not contain tensor names and model description string.
This is fixed version of this model: [kws.zip](https://github.com/Samsung/ONE/files/7774989/kws.zip) (kws stands for KeyWord Spotting, it is model from google to recognize "ok, google" phrase)

<details>

<summary>related files</summary>

[...]

PR #8282, opened 2022-01-17
[common-artifacts] Exclude SVDF

seanshpark:
This will exclude tests for SVDF Op.

### DIFF
--- a/compiler/common-artifacts/exclude.lst
+++ b/compiler/common-artifacts/exclude.lst
@@ -1,11 +1,13 @@
 #[[ circlize : Exclude from tflite-to-circle conversion(tflite2circle) ]]
 ## TensorFlowLiteRecipes
+circlize(SVDF_000)
 
 ## CircleRecipes
 
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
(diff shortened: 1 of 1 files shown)


seanshpark:
For  #8203

mhs4670go (approved):
LGTM

BalyshevArtem (approved):
LGTM

BalyshevArtem:
Q:
According to #8279 there is also `res/TensorFlowLiteRecipes/Part_Add_SVDF_000/test.recipe` and according to https://github.com/Samsung/ONE/pull/8231#discussion_r782575151 I will add `SVDF_001/test.recipe`  - that is why should we add them to this PR?

seanshpark:
> According to #8279 there is also res/TensorFlowLiteRecipes/Part_Add_SVDF_000/test.recipe

Thanks for point this!
I think I forgot this one. But I think it's OK as `Part_Add_SVDF_000` will be used when partition codes. I'll post in another PR.

> according to #8231 (comment) I will add SVDF_001/test.recipe - that is why should we add them to this PR?

Yes, after this lands, please post `SVDF_001` recipe PR :)

PR #8284, opened 2022-01-17
[tflchef] SVDF support for network generation

BalyshevArtem:
Add network generation support for SVDF operator

issue: #8203 

ONE-DCO-1.0-Signed-off-by: Artem Balyshev a.balyshev@partner.samsung.com

PR #8285, opened 2022-01-17
[recipe] SVDF add network recipe

BalyshevArtem:
Add SVDF recipe used for testing

issue: #8203 

### DIFF
--- a/res/TensorFlowLiteRecipes/SVDF_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/SVDF_000/test.recipe
@@ -1,0 +1,62 @@
+operand {
+  name: "ifm"
+  type: FLOAT32
+  shape { dim: 1 dim: 16 }
+}
+operand {
+  name: "weight_feature"
+  type: FLOAT32
+  shape { dim: 64 dim: 16 }
+  filler {
+    tag: "gaussian"
+    arg: "0.0"
+    arg: "1.0"
+  }
+}
+operand {
+  name: "weight_time"
+  type: FLOAT32
+  shape { dim: 64 dim: 8 }
+  filler {
+    tag: "gaussian"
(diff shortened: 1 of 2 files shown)


binarman (dismissed):
LGTM

seanshpark (dismissed):
LGTM thank you!

seanshpark:
#8288 landed. please rebase.

compiler/tflchef/proto/tflchef.proto
```diff
@@ -366,6 +366,12 @@ message SquaredDifferenceOptions {
   // None
 }
 
+message SVDFOptions {
+  optional int32 rank = 1 [default = 0];
+  optional Activation activation = 2 [default = NONE];
+  optional bool asymmetric_quantize_inputs = 3 [default = false];
+}
+
```
BalyshevArtem, line 374: For reviewers:
need to add this `SVDFOptions` since the `test.recipe` files use it.
Maybe it is necessary to add this in a separate PR?

seanshpark: > Maybe it is necessary to add this in a separate PR?

Yes, that would be nice and rebase this PR after that lands :)

seanshpark: Or if you like, you can land all tflchef codes with the proto.
My memories are not sure which one we should land first; tflchef or recipes...

BalyshevArtem: Yes, I think tflchef should land first, so this pr will be landed after landing this one #8284

seanshpark (approved):
LGTM

seanshpark:
optional) to make test faster, please reduce dimension size of I/O of each operands as possible.

binarman (approved):

PR #8291, opened 2022-01-19
[luci/lang] SVDF support for IR

BalyshevArtem:
Add IR support for SVDF operator

issue: #8203 

ONE-DCO-1.0-Signed-off-by: Artem Balyshev a.balyshev@partner.samsung.com

PR #8292, opened 2022-01-19
[luci/import] SVDF support for import

BalyshevArtem:
Add import support for SVDF operator.

issue: #8203 

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleSVDF.cpp
+++ b/compiler/luci/import/src/Nodes/CircleSVDF.cpp
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Import/Nodes/CircleSVDF.h"
+
+#include <luci/IR/Nodes/CircleSVDF.h>
+
+#include <loco.h>
+
+namespace luci
+{
+
+bool CircleSVDFBuilder::validate(const ValidateArgs &args) const
+{
+  const auto &inputs = args.op.inputs;
+  if (!(inputs.size() == 4 || inputs.size() == 5))
+    return false;
+
+  return true;
+}
+
+CircleNode *CircleSVDFBuilder::build_node(const circle::OperatorT &op,
+                                          const std::vector<CircleNode *> &inputs,
+                                          loco::Graph *graph) const
+{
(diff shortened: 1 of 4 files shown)


BalyshevArtem:
This PR can be merged only when #8291 will be landed

seanshpark (approved):
LGTM

seanshpark:
@BalyshevArtem , this change is not from your draft but is from mine, which I don't think you should post as a PR; this is what I meant "settle down the conflicts." in https://github.com/Samsung/ONE/issues/8203#issuecomment-1012939932.

For this I'll let you land this code, but please don't afterwards.

compiler/luci/import/src/Nodes/CircleSVDF.cpp
```diff
@@ -0,0 +1,67 @@
...
+    bias->dtype(inputs.at(0)->dtype());
+    node->bias(bias);
+
+    node->input_activation_state(inputs.at(3));
+  }
+  else
+  {
+    node->bias(inputs.at(3));
+    node->input_activation_state(inputs.at(4));
+  }
```
m-bronnikov, line 57: IIUC there is only 2 possible cases: `inputs.size() == 4` or `inputs.size() == 5`. May be add some assert to input size?

m-bronnikov: I think we could simplify this expression a bit and use `node->input_activation_state(inputs.back());`.

m-bronnikov: My suggestion to this part of code:

```suggestion
  const auto create_output_exclude = [graph]()
  {
    auto *bias = graph->nodes()->create<CircleOutputExclude>();
    // CircleOutputExclude doesn't need a type, but since all nodes must have a type,
    // a dummy type is inserted.
    bias->dtype(loco::DataType::FLOAT32);
    return bias;
  }

  assert(input.size() == 4 || input.size() == 5);
  const auto *bias = (inputs.size() == 4) ? create_output_exclude() : inputs.at(3); 
  node->bias(bias);
  node->input_activation_state(inputs.back());
```

m-bronnikov: Please find my comments above as **optional**. I'm agree with current changes.

m-bronnikov (approved):

compiler/luci/import/src/Nodes/CircleSVDF.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <loco.h>
+
+namespace luci
+{
+
+bool CircleSVDFBuilder::validate(const ValidateArgs &args) const
+{
+  const auto &inputs = args.op.inputs;
+  if (!(inputs.size() == 4 || inputs.size() == 5))
+    return false;
```
seanshpark, line 30: > IIUC there is only 2 possible cases: inputs.size() == 4 or inputs.size() == 5. May be add some assert to input size?

This is where input size is checked.

compiler/luci/import/src/Nodes/CircleSVDF.cpp
seanshpark: (1) I prefer not using lambda

(2) `validate` method exist cause not to do `assert(input.size() == 4 || input.size() == 5);` in this method.

m-bronnikov: So, let's leave changes as it is.

m-bronnikov:
LGTM

PR #8295, opened 2022-01-20
[DRAFT] Support scalar BatchNormPatternFinder

llFreetimell:
This draft is for supporting BatchNormPatternFinder.

---

Requested by @meejeong

### DIFF
--- a/compiler/luci/pass/src/BatchNormPatternFinder.cpp
+++ b/compiler/luci/pass/src/BatchNormPatternFinder.cpp
@@ -44,10 +44,10 @@ bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::C
     return false;
   }
 
-  if (constant->rank() != 1)
+  if (constant->rank() > 1)
     return false;
 
-  auto channel_dim = constant->dim(0);
+  auto channel_dim = constant->rank() == 1 ? constant->dim(0) : 1;
   // Assumption: Layout is channel-last
   if (!(channel_dim == add->dim(add->rank() - 1)))
     return false;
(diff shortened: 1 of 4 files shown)


llFreetimell:
@jinevening Could you check whether I can merge this or not? :D (Not urgent!)

compiler/luci/pass/src/BatchNormPatternFinder.cpp
```diff
@@ -44,10 +44,10 @@ bool is_batchnorm_add(const luci::CircleAdd *add, luci::CircleMul *&mul, luci::C
...
   }
 
-  if (constant->rank() != 1)
+  if (constant->rank() > 1)
     return false;
 
-  auto channel_dim = constant->dim(0);
+  auto channel_dim = constant->rank() == 1 ? constant->dim(0) : 1;
   // Assumption: Layout is channel-last
   if (!(channel_dim == add->dim(add->rank() - 1)))
```
jinevening, line 52: I think acceptable batchnorm const shapes are
1. rank 1 const whose channel size is the same with `add`
2. scalar const

This code may not handle the case when `add` has non-one channel.

llFreetimell: Oops, I missed that... thanks :D
"Handling scalar" is okay? :)

jinevening: I think it's okay. 

> Requested by @meejeong

I don't know what is the exact request, but the idea looks feasible to me.

llFreetimell: Ah, the request was "supporting scalar for `ReplaceMulAddWithDepthwiseConvPass`" :)
Current `ReplaceMulAddWithDepthwiseConvPass` only accepts rank-1 constant value but scalar can be accepted actually :D

llFreetimell:
Re-open when this feature is really needed, close.

PR #8299, opened 2022-01-20
[mio-tflite260] Replace constant expression

llFreetimell:
Static analyzer says that `deprecated_builtin_code <= 127` is always true
because maximum value of `int8_t` is 127.


seanshpark (approved):
LGTM thank you!

PR #8304, opened 2022-01-21
[luci/svc] Shape and dtype inference of CircleVariable

seanshpark:
This will enable shape and dtype inference of CircleVariable Node.


seanshpark:
For #8203 , from draft #8279

llFreetimell (approved):

PR #8305, opened 2022-01-21
[luci/import] Enable CircleVariable

seanshpark:
This will enable import of variable tensor as CircleVariable Node.

### DIFF
--- a/compiler/luci/import/include/luci/Import/Nodes/CircleVariable.h
+++ b/compiler/luci/import/include/luci/Import/Nodes/CircleVariable.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_IMPORT_OP_CIRCLE_VARIABLE_H__
+#define __LUCI_IMPORT_OP_CIRCLE_VARIABLE_H__
+
+#include "luci/Import/GraphBuilderContext.h"
+
+#include <luci/IR/Nodes/CircleVariable.h>
+
+/*
+ * @note  Circle does not have node for variable tensor
+ *        Methods here provide helper that creates CircleVariable from
+ *        Tensor having is_variable true value.
+ */
+
+namespace luci
+{
+
+CircleVariable *create_circlevariable(GraphBuilderContext *context, int32_t tensor_index);
+
+} // namespace luci
+
+#endif // __LUCI_IMPORT_OP_CIRCLE_VARIABLE_H__
--- a/compiler/luci/import/src/Importer.cpp
+++ b/compiler/luci/import/src/Importer.cpp
@@ -23,6 +23,7 @@
 #include "luci/Import/GraphBuilderRegistry.h"
 #include "luci/Import/CircleReader.h"
 #include "luci/Import/Nodes/CircleConst.h"
+#include "luci/Import/Nodes/CircleVariable.h"
 
 #include <luci/IR/Module.h>
 #include <luci/IR/CircleNodes.h>
@@ -132,6 +133,15 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
     auto *const_node = const_builder->build(i, &gb_context);
     if (const_node != nullptr)
       nodefinder->enroll(i, const_node);
+  }
+
+  // Create CircleVariable nodes for variable tensors
+  // TODO Add Origin if needed, skip for now
+  for (uint32_t i = 0; i < tensors.size(); ++i)
+  {
+    luci::CircleVariable *variable_node = luci::create_circlevariable(&gb_context, i);
+    if (variable_node != nullptr)
+      nodefinder->enroll(i, variable_node);
   }
 
   // Import the operators.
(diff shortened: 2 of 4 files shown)


seanshpark:
For #8203 , from draft #8279

compiler/luci/import/include/luci/Import/Nodes/CircleVariable.h
```diff
@@ -0,0 +1,37 @@
...
+/*
+ * @note  Circle does not have node for variable tensor
+ *        Methods here provide helper that creates CircleVariable from
+ *        Tensor having is_variable true value.
+ */
+
+namespace luci
+{
+
+CircleVariable *create_circlevariable(GraphBuilderContext *context, int32_t tensor_index);
```
llFreetimell, line 33: Can I ask you why `CircleVariable` not used `TypedNodeBuilder<NodeBuilderType::BUFFER>` contrast to `CircleConst`?
`CircleVariable` seems quite similar with `CircleConst` except only `is_variable`.
Therefore, for me, it would be better to use similar design with `CircleConst` for better understanding and readibility.
But I want to know if you intended something else :)

compiler/luci/import/src/Importer.cpp
```diff
@@ -134,6 +135,15 @@ void convert_graph(const luci::GraphBuilderSource &source, luci::CircleReader &r
...
   }
 
+  // Create CircleVariable nodes for variable tensors
+  // TODO Add Origin if needed, skip for now
+  for (uint32_t i = 0; i < tensors.size(); ++i)
+  {
+    luci::CircleVariable *variable_node = luci::create_circlevariable(&gb_context, i);
+    if (variable_node != nullptr)
+      nodefinder->enroll(i, variable_node);
+  }
```
llFreetimell, line 145: If we apply above comment, I think following can be applied?

```suggestion
  for (uint32_t i = 0; i < tensors.size(); ++i)
  {
    if (auto const_node = const_builder->build(i, &gb_context))
      nodefinder->enroll(i, const_node);
    else if (auto variable_node = variable_builder->build(&gb_context, i))
      nodefinder->enroll(i, variable_node);
  }
```

compiler/luci/import/include/luci/Import/Nodes/CircleVariable.h
seanshpark: That is a choice what I didn't select it for now; there is no need as of now.
If someone has a requiremment then he/she can do it.

compiler/luci/import/src/Importer.cpp
seanshpark: > If we apply above comment,

which comment?

> I think following can be applied?

Yes I think we can but I don't want to mix with existing code for now. You may revise with remove loop for CircleConst later.

compiler/luci/import/include/luci/Import/Nodes/CircleVariable.h
llFreetimell: Hmm...  okay.
FYI, since all other nodes in `luci/Import/Nodes/*.h` use class which is not function, this would break the unification.

compiler/luci/import/src/Importer.cpp
llFreetimell: I recommend this not only because of better readability but also because it can prevent fragment of tensor handling logic thus we may prevent unexpected handling error.
But okay, it's your choice :)

seanshpark: > it can prevent fragment of tensor handling logic

Can you explain more about this?

compiler/luci/import/include/luci/Import/Nodes/CircleVariable.h
seanshpark: > use class which is not function, this would break the unification.

Um? Did we have requirements to provide `unification` ? I designed this but I don't remember putting this in here.

llFreetimell: My questions was because of following thoughts

- When we add a new CircleNode, we followed the design of existing nodes (kind of copy&paste)
- When we add `CircleVariable`, design is different with the existing nodes

So I wondered why they are different, not objected...!

compiler/luci/import/src/Importer.cpp
llFreetimell: It's not critical thing, just the difference of style :)

---

- If we need to change something about `tensors`, we should fix each fragmented blocks(block 1 & block 2)
- If logic of `const_builder` or `variable_builder` is changed, node can be created by both of them unintentionally.
- If other codes are inserted between the fragmented blocks, reading codes becomes little bit harder

```cpp
  // block 1
  {
    auto *const_node = const_builder->build(i, &gb_context);
    if (const_node != nullptr)
      nodefinder->enroll(i, const_node);
  }

  // ...

  // block 2
  {
    luci::CircleVariable *variable_node = luci::create_circlevariable(&gb_context, i);
    if (variable_node != nullptr)
      nodefinder->enroll(i, variable_node);
  }
```

compiler/luci/import/include/luci/Import/Nodes/CircleVariable.h
seanshpark: > When we add CircleVariable, design is different with the existing nodes

CircleVariable is to represent circle ensors(or from tflite tensor) having true with `is_variable`. As like CircleConst, which was introduced from the first place, to represent tensors with initial values. (CircleVariable doesn't have initial values, from the BUFFER, as of now).
Preparing(or initializing) CircleVariable followed procedures from CircleConst, which was first created with a method (not a class). Recently it was changed using a class like the Op nodes in purpose of #8042 (if you didn't notice).
As I wrote currently there is no need to prepare CircleVariable through a builder.
So I wrote it using plain method, which follow the initial implementation (copy-paste).

compiler/luci/import/src/Importer.cpp
seanshpark: > If logic of const_builder or variable_builder is changed, node can be created by both of them unintentionally.

You've pointed an important thing which current CircleConst doesn't check.
I think CircleConst should not be created when `variable` is false.

seanshpark: We can still merge two loops for creating CircleNode from tensor as you pointed.
Please post a PR after this lands if you like.

mhs4670go (approved):
LGTM

PR #8314, opened 2022-01-21
[luci/part] Revise check for virtual input nodes

seanshpark:
This will introduce IsVirtualInputNode visitor to check virtual input
node CircleOutputExclude to prepare more nodes.

### DIFF
--- a/compiler/luci/partition/src/PartitionPGroups.cpp
+++ b/compiler/luci/partition/src/PartitionPGroups.cpp
@@ -118,6 +118,17 @@ std::string group_from_partition(const luci::CircleNode *node,
   return group;
 }
 
+class IsVirtualInputNode final : public luci::CircleNodeVisitor<bool>
+{
+public:
+  // TODO check CircleOutputDummy
+  bool visit(const luci::CircleOutputExclude *) final { return true; }
+  // TODO check CircleVariable
+
+  // default is false
+  bool visit(const luci::CircleNode *) final { return false; }
+};
+
 class IsMultiOutputNode final : public luci::CircleNodeVisitor<bool>
 {
 public:
@@ -148,11 +159,12 @@ void append(luci::CircleNode *node, luci::PGroups *pgroups, const std::string &g
 
   pgroup->pnodes.push_back(std::move(pnode));
 
+  IsVirtualInputNode queryvi;
   // Set input of PGroup
   for (uint32_t in = 0; in < node->arity(); ++in)
   {
     auto input = loco::must_cast<luci::CircleNode *>(node->arg(in));
-    if (dynamic_cast<const luci::CircleOutputExclude *>(input) != nullptr)
+    if (input->accept(&queryvi))
     {
       auto pnode = std::make_unique<luci::PNode>();
       pnode->node = input;


seanshpark:
For #8203 , from draft #8279

compiler/luci/partition/src/PartitionPGroups.cpp
```diff
@@ -118,6 +118,17 @@ std::string group_from_partition(const luci::CircleNode *node,
...
+class IsVirtualInputNode final : public luci::CircleNodeVisitor<bool>
+{
+public:
+  // TODO check CircleOutputDummy
+  bool visit(const luci::CircleOutputExclude *) final { return true; }
+  // TODO check CircleVariable
+
+  // default is false
+  bool visit(const luci::CircleNode *) final { return false; }
+};
```
mhs4670go, line 130: Just curiosity, is there any reason this class is needed even though there is a class that checks if it is a virtual node?

seanshpark: Yes, this class will only check for virtual nodes that are used for input of some node.

seanshpark: Currently only three; CircleOutputExclude, CircleVariable and CircleOutputDummy are the ones.

mhs4670go: > this class will only check for virtual nodes that are used for input of some node.

@seanshpark Yes but the reason why I asked this is because I thought `input->accept(&IsVirtualNode)` and `input->accept(&queryvi)` are same things as it is already an "input".

As you said, `IsVirtualInputNode` checks both of virtual nodes and input. But since we already know that the callers(visitable nodes) are input, checking if it is input or not seems unnecessary.

seanshpark: This code checks if the `input` node can have input of it's own.
The `input` is NOT the node.
That's is we're NOT checking the `node` is the type of I'm interested.

seanshpark: This change does not change it's behaviour.
```
if (dynamic_cast<const luci::CircleOutputExclude *>(input) != nullptr)
```
-->
```
if (input->accept(&queryvi))
```

`input->accept(&IsVirtualNode) and input->accept(&queryvi)` if defintly NOT THE SAME.

```cxx
class IsVirtualNode final : public luci::CircleNodeVisitor<bool>
{
public:
  bool visit(const luci::CircleInput *) final { return true; }
  bool visit(const luci::CircleOutput *) final { return true; }
  // For multiple outputs
  bool visit(const luci::CircleCustomOut *) final { return true; }
  bool visit(const luci::CircleIfOut *) final { return true; }
  bool visit(const luci::CircleNonMaxSuppressionV4Out *) final { return true; }
  bool visit(const luci::CircleNonMaxSuppressionV5Out *) final { return true; }
  bool visit(const luci::CircleSplitOut *) final { return true; }
  bool visit(const luci::CircleSplitVOut *) final { return true; }
  bool visit(const luci::CircleTopKV2Out *) final { return true; }
  bool visit(const luci::CircleUniqueOut *) final { return true; }
  bool visit(const luci::CircleUnpackOut *) final { return true; }
  bool visit(const luci::CircleWhileOut *) final { return true; }
  // For inputs not used
  bool visit(const luci::CircleOutputDummy *) final { return true; }
  bool visit(const luci::CircleOutputExclude *) final { return true; }
  bool visit(const luci::CircleVariable *) final { return true; }
  // TODO add all virtual nodes

  // default is false
  bool visit(const luci::CircleNode *) final { return false; }
};
```

seanshpark: How is same as
```cxx
class IsVirtualInputNode final : public luci::CircleNodeVisitor<bool>
{
public:
  // TODO check CircleOutputDummy
  bool visit(const luci::CircleOutputExclude *) final { return true; }
  bool visit(const luci::CircleVariable *) final { return true; }

  // default is false
  bool visit(const luci::CircleNode *) final { return false; }
};
```
?

mhs4670go: @seanshpark Thank you for detail:) I think `input` word confused me. `input` could be two things.
- input(arg) of circle node
- `CircleInput` node (graph input, operation)

And I used to think `input` of `IsVirtualInputNode` had first meaning. 

(optional) `CanbeGraphInput` or something might good to reduce misunderstanding. But, naming is always hard to me,,

mhs4670go (approved):
LGTM

PR #8315, opened 2022-01-21
[luci-interpreter] Support import without constant copying

m-bronnikov:
This commit adds support of importing without constant copying.

ONE-DCO-1.0-Signed-off-by: Maksim Bronnikov <max120199@gmail.com>

-------------------------

For: #8042
Draft: #8190

PR #8320, opened 2022-01-21
[luci-interpreter] Add CircleReferencingConst import

m-bronnikov:
This commit adds support of constants import via `CircleReferencingConst` custom node.

--------------------------

For: #8042
Draft: #8190
Second part of: #8315

### DIFF
--- a/compiler/luci-interpreter/src/import/Nodes/CircleReferencingConst.cpp
+++ b/compiler/luci-interpreter/src/import/Nodes/CircleReferencingConst.cpp
@@ -16,8 +16,6 @@
 
 #include "CircleReferencingConst.h"
 
-#include <ostream>
-#include <string>
 #include <vector>
 
 namespace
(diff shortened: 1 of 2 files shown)


compiler/luci-interpreter/src/import/Nodes/CircleReferencingConst.cpp
```diff
@@ -0,0 +1,115 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "CircleReferencingConst.h"
+
+#include <ostream>
+#include <string>
```
BalyshevArtem, line 20: (optional)
Do you actually use it?

BalyshevArtem (approved):
LGTM, thank you

compiler/luci-interpreter/src/import/Nodes/CircleReferencingConst.cpp
seanshpark: I think would be better to remove unused header file(s)

m-bronnikov:
```suggestion
```

BalyshevArtem (approved):

seanshpark (approved):
LGTM thank you!

PR #8321, opened 2022-01-21
[luci-interpreter/loader] Support CircleReferencingConst load

m-bronnikov:
This commit adds support of CircleReferencingConst custom node to loader.

------------------------

For: #8042
Draft: #8190

### DIFF
--- a/compiler/luci-interpreter/src/loader/GraphLoader.cpp
+++ b/compiler/luci-interpreter/src/loader/GraphLoader.cpp
@@ -121,9 +121,11 @@ bool isExecutableNode(const luci::CircleNode *node)
     {
       auto const custom_node = loco::must_cast<const luci::CircleCustom *>(node);
 
-      // TODO handle more Custom ops here
+      // TODO handle more non-executable Custom ops here
       if (custom_node->custom_code() == "CircleReferencingConst")
         return false;
+
+      return true;
     }
     default:
       return true;
--- a/compiler/luci-interpreter/src/loader/GraphLoader.cpp
+++ b/compiler/luci-interpreter/src/loader/GraphLoader.cpp
@@ -73,6 +73,25 @@ const void *getNodeData(const luci::CircleConst *node, size_t *data_size)
   }
 }
 
+const void *getNodeData(const luci::CircleCustom *node, size_t *data_size)
+{
+  if (node->custom_code() != "CircleReferencingConst")
+    return nullptr;
+
+  // helper struct which describes data loaded to custom_options of CircleReferencingConst node
+  struct ConstDataReference
+  {
+    const uint8_t *data = nullptr;
+    uint32_t size = 0;
+  };
+
+  const auto &custom_options = node->custom_options();
+  const auto &const_data_ref = *reinterpret_cast<const ConstDataReference *>(custom_options.data());
+
+  *data_size = const_data_ref.size;
+  return const_data_ref.data;
+}
+
 bool isExecutableNode(const luci::CircleNode *node)
 {
   switch (node->opcode())
@@ -96,6 +115,15 @@ bool isExecutableNode(const luci::CircleNode *node)
     case luci::CircleOpcode::CIRCLEVARIABLE:
     case luci::CircleOpcode::CIRCLEWHILEOUT:
       return false;
+    // Custom nodes may be executable and non-executable
+    case luci::CircleOpcode::CUSTOM:
+    {
+      auto const custom_node = loco::must_cast<const luci::CircleCustom *>(node);
+
+      // TODO handle more Custom ops here
+      if (custom_node->custom_code() == "CircleReferencingConst")
+        return false;
+    }
     default:
       return true;
   }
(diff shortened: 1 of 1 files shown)


m-bronnikov:
This PR conflicts a bit with #8318. Let's wait until #8318 is merged and rebase this PR after that.

seanshpark:
@m-bronnikov , #8318 landed, please rebase.

compiler/luci-interpreter/src/loader/GraphLoader.cpp
```diff
@@ -88,7 +107,19 @@ bool isExecutableNode(const luci::CircleNode *node)
...
+    // Custom nodes may be executable and non-executable
+    case luci::CircleOpcode::CUSTOM:
+    {
+      auto const custom_node = loco::must_cast<const luci::CircleCustom *>(node);
+
+      // TODO handle more custom ops here
+      if (custom_node->custom_code() == "CircleReferencingConst")
+        return false;
+
+      throw std::runtime_error("Unknown Custom Node, yet.");
```
seanshpark, line 121: If we throw here, maybe also check and throw inside `isTensorProducingNode()` for consistancy?

m-bronnikov: Hmm, this suggestion makes sense, but I don't like idea of adding more checks for supported `CircleCustom` cases. So, I think it would be better to add another helper: `isValidCustomNode` and check Custom node is valid only once in thi s helper.

m-bronnikov: And therefore remove this throw here.

m-bronnikov: > So, I think it would be better to add another helper: `isValidCustomNode` and check Custom node is valid only once in thi s helper.

I added this check with name "isSupportedCustomNode".

compiler/luci-interpreter/src/loader/GraphLoader.cpp
```diff
@@ -73,6 +73,25 @@ const void *getNodeData(const luci::CircleConst *node, size_t *data_size)
...
+{
+  if (node->custom_code() != "CircleReferencingConst")
+    return nullptr;
+
+  // helper struct which describes data loaded to custom_options of CircleReferencingConst node
+  struct ConstDataReference
+  {
+    const uint8_t *data = nullptr;
+    uint32_t size = 0;
+  };
```
seanshpark, line 86: (optional) I assume this struct exist also in other file. So I think it would be better to include from header. Or add a `// TODO move to header`

compiler/luci-interpreter/src/loader/GraphLoader.cpp
```diff
@@ -162,6 +204,10 @@ void GraphLoader::loadTensors()
     {
       shape = getNodeShape(variable_node);
     }
+    else if (const auto *custom_out_node = dynamic_cast<const luci::CircleCustomOut *>(node))
+    {
+      shape = getNodeShape(custom_out_node);
+    }
```
m-bronnikov, line 210: Since this place has not small (4 for now) number of cases with `dynamic_cast` usage I think better to rewrite this place with `switch(opcode)-case` construction in order increase performance and reduce code duplication.

m-bronnikov: P.S. In another PR.

+cc @binarman

compiler/luci-interpreter/src/loader/GraphLoader.cpp
m-bronnikov: I'd prefer second option for now.

seanshpark (approved):
LGTM

seanshpark:
Maybe another reviewer like @binarman ?

m-bronnikov:
TODO comment added.

seanshpark (dismissed):
LGTM thank you!

m-bronnikov:
> Maybe another reviewer like @binarman ?

It would be nice. I will add @BalyshevArtem and @binarman  to rewievers. I think a review of one of them will be enough.

compiler/luci-interpreter/src/loader/GraphLoader.cpp
```diff
@@ -96,6 +116,15 @@ bool isExecutableNode(const luci::CircleNode *node)
...
       return false;
+    // Custom nodes may be executable and non-executable
+    case luci::CircleOpcode::CUSTOM:
+    {
+      auto const custom_node = loco::must_cast<const luci::CircleCustom *>(node);
+
+      // TODO handle more Custom ops here
+      if (custom_node->custom_code() == "CircleReferencingConst")
+        return false;
+    }
```
m-bronnikov, line 127: CI fails due to implicit fallthrough here =(
Let's make explicit!

m-bronnikov: done!

seanshpark (approved):
LGTM

BalyshevArtem (approved):
LGTM, thank you!

m-bronnikov:
@seanshpark I think 2 review should be enough =)

PR #8298, opened 2022-01-20
[DRAFT][NO MERGE] Refactoring VerifyQuantizedNode*Type

llFreetimell:
This draft if for refactoring `VerifyQUantizedNote*Type`.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

---

Parent Issue : #8297

Issue #8297, opened 2022-01-20
[luci/pass] Refactoring VerifyQuantizedNode*Type


llFreetimell:
### What?

There are `VerifyQuantizedNodeU8Type` and `VerifyQuantizedNodeS16Type` in `luci/pass`.
Let's refactor following things!

- Reduce duplicated codes
- Reduce if-else statement
- Increase cohesion of `QuantizedNodeType` related codes.

### How?

- Apply strategy pattern by introducing `VerifyQuantizedNodeType` as base Strategy interface class
  - Introduce `create` function which creates `VerifyQuantizedNodeType` according to `datatype`
- Introduce `VerifyQuantizedNodeTypeNormalStrategy` as concrete Strategy class with template
  - Grouping similar `visit` functions to here
  - Different `visit` functions are implemented in derived class
  - Introduce helper function to reduce similar codes

---

#### Related PRs

- Draft : #8298
- [x] Introduce `VerifyQuantizedNodeType`
  - [x] Declaration of base and derived class #8333
  - [x] Declaration of helper function #8351
- [x] Apply `VerifyQuantizedNodeType` #8350
- [x] Remove unused codes #8378

llFreetimell:
Done!

PR #8333, opened 2022-01-25
[luci/pass] Introduce VerifyQuantizedNodeType

llFreetimell:
This commit introduces `VerifyQuantizedNodeType` which includes both
`VerifyQuantizedNodeU8Type` and `VerifyQuantizedNodeS16Type`.

---

Parent Issue : #8297
Draft : #8298

### DIFF
--- a/compiler/luci/pass/src/VerifyQuantizedNodeType.h
+++ b/compiler/luci/pass/src/VerifyQuantizedNodeType.h
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_VERIFY_QUANTIZED_NODE_TYPE_H__
+#define __LUCI_VERIFY_QUANTIZED_NODE_TYPE_H__
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleNodeVisitor.h>
+
+namespace luci
+{
+
+/**
+ * @brief Verify the data type of quantized node
+ * @details
+ *
+ * Targets to verify
+ * - node's output (i.e., node itself)
+ * - node's inputs
+ */
+class VerifyQuantizedNodeType
+{
+public:
+  static std::shared_ptr<VerifyQuantizedNodeType> create(loco::DataType dtype);
+
+public:
+  virtual bool verify(luci::CircleNode *node) = 0;
+};
+
+/**
+ * @brief Verify using quantization type of a node and bias
+ *
+ * @tparam Qtype Quantization type for a node (e.g. Q8, Q16, ...)
+ * @tparam Btype Bias quantization type (e.g. For Q8, S32 is used)
+ */
+template <loco::DataType Qtype, loco::DataType Btype>
+class VerifyQuantizedNodeTypeBase : public luci::CircleNodeVisitor<bool>,
+                                    public VerifyQuantizedNodeType
+{
+public:
+  bool verify(luci::CircleNode *node) { return node->accept(this); }
+
+protected:
+  bool has_type(const loco::Node *node, loco::DataType dtype)
+  {
+    auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
+    return circle_node->dtype() == dtype;
+  }
+
+private:
+  bool visit(const luci::CircleAdd *node);
+  bool visit(const luci::CircleArgMax *node);
+  bool visit(const luci::CircleAveragePool2D *node);
+  bool visit(const luci::CircleBatchToSpaceND *node);
+  bool visit(const luci::CircleCast *node);
+  bool visit(const luci::CircleConv2D *node);
+  bool visit(const luci::CircleConcatenation *node);
+  bool visit(const luci::CircleDepthToSpace *node);
+  bool visit(const luci::CircleDepthwiseConv2D *node);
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/VerifyQuantizedNodeType.h
```diff
@@ -0,0 +1,618 @@
...
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_VERIFY_QUANTIZED_NODE_TYPE_H__
+#define __LUCI_VERIFY_QUANTIZED_NODE_TYPE_H__
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleNodeVisitor.h>
+
+#include <cmath>
```
jinevening, line 23: Is this necessary?

compiler/luci/pass/src/VerifyQuantizedNodeType.h
```diff
@@ -0,0 +1,618 @@
...
+    RETURN_FALSE_UNLESS(node->quantparam()->zerop[0] == 0);
+    return true;
+  }
+};
+
+} // namespace luci
+
+#undef RETURN_FALSE_UNLESS
+
+#endif // __LUCI_VERIFY_QUNTIZED_NODE_TYPE_H__
```
jinevening, line 618:
```suggestion
#endif // __LUCI_VERIFY_QUANTIZED_NODE_TYPE_H__
```

compiler/luci/pass/src/VerifyQuantizedNodeType.h
llFreetimell: I copied it from original code and I found that `CircleFloorDiv` and `CircleFloor` need it! :D

```cpp
RETURN_FALSE_UNLESS(std::roundf(node->quantparam()->scale[0]) == node->quantparam()->scale[0]);
```

compiler/luci/pass/src/VerifyQuantizedNodeType.h
```diff
@@ -0,0 +1,618 @@
...
+class VerifyQuantizedNodeType
+{
+public:
+  static std::shared_ptr<VerifyQuantizedNodeType> create(loco::DataType dtype);
+
+public:
+  virtual bool verify(luci::CircleNode *node) = 0;
+};
+
+template <loco::DataType Qtype, loco::DataType Btype>
```
jinevening, line 52: It would be better to add comments about what is Qtype and Btype.

jinevening (approved):
LGTM

llFreetimell:
@jinevening I fixed typo errors T.T... PTAL

seanshpark:
Q) implementation is in header not in cpp file. is there some benefit for this?
For me, readability is not good; which node is handled is not identified at once

And I think it would be better to sort node items in alphabetical order.

llFreetimell:
> Q) implementation is in header not in cpp file. is there some benefit for this?

I just tried to keep original codes as similar as possible, not for benefits :D
I will move implementations to cpp files!

> it would be better to sort node items in alphabetical order.

👍

jinevening (approved):

compiler/luci/pass/src/VerifyQuantizedNodeType.h
```diff
@@ -0,0 +1,141 @@
...
+  bool visit(const luci::CircleSpaceToBatchND *node);
+  bool visit(const luci::CircleSpaceToDepth *node);
+  bool visit(const luci::CircleSplit *node);
+  bool visit(const luci::CircleSplitOut *node);
+  bool visit(const luci::CircleSplitV *node);
+  bool visit(const luci::CircleSplitVOut *node);
+  bool visit(const luci::CircleSqrt *node);
+  bool visit(const luci::CircleStridedSlice *node);
+  virtual bool visit(const luci::CircleTanh *node) = 0;
+  bool visit(const luci::CircleTranspose *node);
```
seanshpark, line 111: (optional) just optional; mixing pure virtual and implementation is OK but seems a little odd.
how about, (1) change `VerifyQuantizedNodeTypeBase` to `VerifyQuantizedNodeType` with all pure virtuals (2) introduce `VerifyQuantizedNodeTypeBase` with type neutral implementations only ?

seanshpark (approved):
LGTM

compiler/luci/pass/src/VerifyQuantizedNodeType.h
llFreetimell: Thanks for your opinion and I agree that mixing them is not that "pretty" :)

But one of my original intention was to reduce duplicated codes.

#### Before
```cpp
bool visit(const luci::CircleTranspose *node)
{
  RETURN_FALSE_UNLESS(has_type(node, loco::DataType::U8))
  RETURN_FALSE_UNLESS(has_type(node->a(), loco::DataType::U8))
  RETURN_FALSE_UNLESS(has_type(node->perm(), loco::DataType::S32))
  return true;
}

bool visit(const luci::CircleTranspose *node)
{
  RETURN_FALSE_UNLESS(has_type(node, loco::DataType::S16))
  RETURN_FALSE_UNLESS(has_type(node->a(), loco::DataType::S16))
  RETURN_FALSE_UNLESS(has_type(node->perm(), loco::DataType::S32))
  return true;
}
```

#### After
```cpp
bool visit(const luci::CircleTranspose *node)
{
  RETURN_FALSE_UNLESS(has_type(node, Qtype))
  RETURN_FALSE_UNLESS(has_type(node->a(), Qtype))
  RETURN_FALSE_UNLESS(has_type(node->perm(), loco::DataType::S32))
  return true;
}
```

Unfortunately, three operation(Tanh, Softmax, Logistic) cannot be reduced as one function so I implemented as current code :(

---

The reason why I created `VerifyQuantizedNodeType` as sort of interface is for preparing when new kinds of quantization is introduced :)

```cpp
class VerifyNewQuantizationMethod : public VerifyQuantizedNodeType
{
//...
}
```

seanshpark: Um... there seems some misunderstanding with my explanation.
I'll post a PR after this lands.

seanshpark:
I missed something.

`CircleNodeVisitorBase` already declared all the nodes as pure virtual
```
#define CIRCLE_NODE(OPCODE, CIRCLE_CLASS) virtual T visit(const CIRCLE_CLASS *) = 0;
```

adding vi

compiler/luci/pass/src/VerifyQuantizedNodeType.h
seanshpark: Found out that removing pure virtuall will do

PR #8338, opened 2022-01-26
[common-artifacts] Include SVDF tests

seanshpark:
This will remove SVDF test items from exclude list to turn on tests.


seanshpark:
For #8203 , from draft #8279

mhs4670go (approved):
LGTM

PR #8340, opened 2022-01-26
[luci/value-test] Enable CircleSVDF

seanshpark:
This will enable value test for CircleSVDF node.


seanshpark:
For #8203 , from draft #8279

seanshpark:
@BalyshevArtem , could you please check luci-interpreter SVDF kernel implementation?
This change will enable luci value test; comparing random input execution results of tensorflow lite vs luci-interpreter. if some goes wrong, this means we need to check luci-interpreter kernel.
I'v noticed there was some changes in the kernel code of your first draft and current master.

mhs4670go (approved):
LGTM

mhs4670go:
@nnfw-bot test nncc-debug

mhs4670go:
@nnfw-bot test nncc-release

mhs4670go:
Oh, I missed https://github.com/Samsung/ONE/pull/8340#issuecomment-1021851744.

BalyshevArtem:
I made a model (from `SVDF_000`) using a tflchef tool. Then I made random data of the right shape. After that, I ran this data using tflite and luci interpreter and made a diff of the resulting data:
```
   array([[ 0.0000000e+00, -1.4012985e-45,  0.0000000e+00,  0.0000000e+00,
            0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00,                                                          
            0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00,                                                         
           -3.5762787e-07,  0.0000000e+00, -5.9604645e-08,  0.0000000e+00,                                                          
            0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00,                                                          
            2.9802322e-08,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00,                                                         
            0.0000000e+00,  2.3841858e-07,  0.0000000e+00,  0.0000000e+00,                                                         
            0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00,                                                          
            0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00,                                                          
            0.0000000e+00, -1.7881393e-07,  0.0000000e+00,  0.0000000e+00,                                                          
            0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00,                                                         
            0.0000000e+00, -1.4012985e-45,  0.0000000e+00,  0.0000000e+00,                                                          
            0.0000000e+00,  0.0000000e+00, -5.9604645e-08,  0.0000000e+00,                                                          
            0.0000000e+00, -2.3841858e-07,  0.0000000e+00,  0.0000000e+00,                                                         
            0.0000000e+00,  0.0000000e+00,  2.8025969e-45,  0.0000000e+00,                                                          
            0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  0.0000000e+00]],                                                     
            dtype=float32) 
```
There is very small difference between them. Actually in kernel  `FloatTest` SVDF test uses tolerance=10e-5 just like the test in tflite.
Is there a way to add tolerance=10e-5 to this test? Or then we can try changing float to int in SVDF_000 and SVDF_001

seanshpark:
> Is there a way to add tolerance=10e-5 to this test? 

`luci_eval_verifier.py` has this code for float32, which tolerance is 1e-5.
maybe increase to 1e-4 ?
```
        elif output_details["dtype"] == np.float32:
            if np.allclose(
                    luci_output_data, intp_output_data, rtol=1.e-5, atol=1.e-5) == False:
                raise SystemExit("Execution result of " + tflite_model +
                                 " does not match with " + circle_model)
```

seanshpark:
> 18:43:02  98/101 Test  #98: luci_value_test ............................***Failed  195.58 sec
18:46:17 FAILED
18:46:17 - SVDF_000
18:46:17 - SVDF_001

seanshpark:
@BalyshevArtem , could you please check again using `luci_eval_verifier.py`?

> SVDF test uses tolerance=10e-5 just like the test in tflite.

As failed also with `10e-5` (or `1e-4`), I'll do some local test tomorrow.

seanshpark:
local test log shows this for `SVDF_001`
> RuntimeError: tensorflow/lite/kernels/svdf.cc:83 node->inputs->size != 5 (4 != 5)Node number 0 (SVDF) failed to prepare.

seanshpark:
I think we need to disable SVDF_001 for the moment...

seanshpark:
@BalyshevArtem , can you please share where I can get `SVDF test uses tolerance=10e-5 just like the test in tflite.` reference ?
If I have to change our code, at least I want to know this.

BalyshevArtem:
> @BalyshevArtem , can you please share where I can get `SVDF test uses tolerance=10e-5 just like the test in tflite.` reference ? If I have to change our code, at least I want to know this.

https://github.com/tensorflow/tensorflow/blob/3038529326066c4eb5cb0453d7f82d929551a24e/tensorflow/lite/kernels/svdf_test.cc#L238

```c++
class SVDFOpTest : public ::testing::TestWithParam<bool> {
 protected:
  void VerifyGoldens(float golden_input[], float golden_output[],
                     int golden_size, BaseSVDFOpModel* svdf,
                     float tolerance = 1e-5) {
    const int svdf_num_batches = svdf->num_batches();
    const int svdf_input_size = svdf->input_size();
    const int svdf_num_units = svdf->num_units();
    const int input_sequence_size =
        golden_size / sizeof(float) / (svdf_input_size * svdf_num_batches);
    // Going over each input batch, setting the input tensor, invoking the SVDF
    // op and checking the output with the expected golden values.
    for (int i = 0; i < input_sequence_size; i++) {
      float* batch_start =
          golden_input + i * svdf_input_size * svdf_num_batches;
      float* batch_end = batch_start + svdf_input_size * svdf_num_batches;
      svdf->SetInput(0, batch_start, batch_end);

      svdf->Invoke();

      const float* golden_start =
          golden_output + i * svdf_num_units * svdf_num_batches;
      const float* golden_end =
          golden_start + svdf_num_units * svdf_num_batches;
      std::vector<float> expected;
      expected.insert(expected.end(), golden_start, golden_end);

      EXPECT_THAT(svdf->GetOutput(),
                  ElementsAreArray(ArrayFloatNear(expected, tolerance)));
    }
  }
};
```

BalyshevArtem:
I also found this problem with `SVDF_001` and as I understand there is no problem with `SVDF_000` and we can leave the old value `1e-5`

seanshpark:
In the test, there is `/*tolerance=*/0.007175` and `/*tolerance=*/0.004285` which is `7.175e-3`.
Depending on the random input value, test result maybe out of tolerance.
I think this is the case when I first seen this test failure only with `SVDF_000`.

I'm not sure to change the tolerance cause of this Op.
I think I need to add a method something like`addevalex(SVDF_000, 7e-3)`

seanshpark:
close as #8359 posted

PR #8341, opened 2022-01-26
[luci/pass] Revise VerifyQuantizedNodeTypeBase

seanshpark:
This will revise VerifyQuantizedNodeTypeBase class to remove pure virtual.

### DIFF
--- a/compiler/luci/pass/src/VerifyQuantizedNodeType.h
+++ b/compiler/luci/pass/src/VerifyQuantizedNodeType.h
@@ -80,7 +80,6 @@ private:
   bool visit(const luci::CircleInstanceNorm *node);
   bool visit(const luci::CircleLocalResponseNormalization *node);
   bool visit(const luci::CircleLogicalOr *node);
-  virtual bool visit(const luci::CircleLogistic *node) = 0;
   bool visit(const luci::CircleMaxPool2D *node);
   bool visit(const luci::CircleMean *node);
   bool visit(const luci::CircleMirrorPad *node);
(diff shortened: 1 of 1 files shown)


seanshpark:
https://github.com/Samsung/ONE/pull/8333#issuecomment-1021810821

jinevening (approved):
LGTM

compiler/luci/pass/src/VerifyQuantizedNodeType.h
```diff
@@ -107,12 +105,17 @@ class VerifyQuantizedNodeTypeBase : public luci::CircleNodeVisitor<bool>,
...
   bool visit(const luci::CircleTranspose *node);
   bool visit(const luci::CircleTransposeConv *node);
   bool visit(const luci::CircleUnpack *node);
   bool visit(const luci::CircleUnpackOut *node);
 
+  // NOTE below nodes has differnent implementation for Qtype/Btype and
+  //      implementations exist in VerifyQuantizedNodeU8Type, VerifyQuantizedNodeS16Type
+  // bool visit(const luci::CircleLogistic *node);
+  // bool visit(const luci::CircleSoftmax *node);
+  // bool visit(const luci::CircleTanh *node);
```
llFreetimell, line 117: I agree with this change, following comment is just FYI after seeing https://github.com/Samsung/ONE/pull/8333#issuecomment-1021810821 :)

---

At first, I used `virtual bool visit(...) = 0;` to prevent mistake such like "An operation is implemented in `VerifyQuantizedNodeU8Type` but not implemented in `VerifyQuantizedNodeS16Type`".
It was because of following code.

```cpp
  bool visit(const luci::CircleNode *) { return true; }
```

Since `VerifyQuantizedNodeTypeBase` will be used only as base class, designating some of operations as pure virtual looked okay for me :)

llFreetimell (approved):
LGTM, thank you

compiler/luci/pass/src/VerifyQuantizedNodeType.h
seanshpark: > "An operation is implemented in VerifyQuantizedNodeU8Type but not implemented in VerifyQuantizedNodeS16Type".
It was because of following code.

You got a point. This may give some problem in the future. Let's fix it when this happens.

PR #8352, opened 2022-01-27
[debian] Add preserve environment option

mhs4670go:
This commit adds -p option to postinst script that preserves current environment.

This change fixes a bug where current user env variables are not passed properly.

### DIFF
--- a/infra/debian/compiler/postinst
+++ b/infra/debian/compiler/postinst
@@ -9,4 +9,4 @@ set -e
 # which causes invalid permission problem.
 # e.g. When `pip` installs user packages, it proceeds based on $HOME.
 # To proper installation, $HOME should be root.
-su - $(whoami) -c '/usr/share/one/bin/one-prepare-venv' # $(whoami) = root
+su - $(whoami) -p -c '/usr/share/one/bin/one-prepare-venv' # $(whoami) = root


seanshpark (approved):
LGTM thank you!

infra/debian/compiler/postinst
```diff
@@ -9,4 +9,4 @@ set -e
 # which causes invalid permission problem.
 # e.g. When `pip` installs user packages, it proceeds based on $HOME.
 # To proper installation, $HOME should be root.
-su - $(whoami) -c '/usr/share/one/bin/one-prepare-venv' # $(whoami) = root
+su - $(whoami) -p -c '/usr/share/one/bin/one-prepare-venv' # $(whoami) = root
```
hseok-oh, line 12: Conflict with above comment ?

>This script is invoked as root except environmental variables,
which causes invalid permission problem.

seanshpark: `su`
```
  -m, -p,
  --preserve-environment        do not reset environment variables, and
                                keep the same shell
```
As I know, `-p` is to pass environment variable that currently don't.

seanshpark:
#7070 is related

infra/debian/compiler/postinst
mhs4670go: @hseok-oh Well, this script(`postinst`) is invoked as root. But, the environment variables are from the user that executed the script. For example, `$HOME` would refer to `/home/user/` and after introducing `su` command, it points to the root.

But, there was a problem. We have to pass the user's env variables to the root because of pip related settings(proxy, etc.) or something. So, `-p` is introduced.

No `su` command : root privilege + user env var
With 'su' command : root privilege + root env var
With 'su -p' command : root privilege + root env var + user env var

hseok-oh: I see. Thanks.

hseok-oh (approved):
LGTM

PR #8353, opened 2022-01-27
[one-cmds] Introduce constant.py

mhs4670go:
This commit introduces constant.py.

Related: #8169 


seanshpark (approved):
LGTM

Issue #8354, opened 2022-01-27
[circle-execution-plan] Support for calculating the size of scratchpad tensors


BalyshevArtem:
### What?
Need to support calculation of scratchpad tensor sizes for some operations:
- [x] DepthwiseConv2d #8364
- [x] AveragePool2d  #8364
- [x] SVDF #8364
- [x] BatchMatMulOp #8368

### Why?
These operations have scratchpad tensors, whose size must be calculated and recorded in the meta-information of the node in order to be able to use the static memory manager for these operations.

BalyshevArtem:
According to #8335 add to list also BatchMatMul op.

BalyshevArtem:
Everything is merged, close this issue

PR #8335, opened 2022-01-25
[luci-interpreter] Support BatchMatMul Op

jinevening:
This supports BatchMatMul Operator in luci-interpreter.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Draft PR: #8334 
Related to: #8325

PR #8355, opened 2022-01-27
[circle-execution-plan] Support scratchpad size for DepthwiseConv2d

BalyshevArtem:
This pr adds support calculation for DepthwiseConv2d scratchpad tensors for circle execution planner.

issue #8354 

### DIFF
--- a/compiler/circle-execution-plan/src/ExecutionPlanner.cpp
+++ b/compiler/circle-execution-plan/src/ExecutionPlanner.cpp
@@ -304,17 +304,27 @@ void ExecutionPlanner::create_alloc_node_inform_vector(bool null_consts, bool nu
 
     // Scratchpad If needed
     uint32_t scratchpad_size = 0;
-    switch (circle_node->opcode())
+    if (!null_scratchpad)
     {
-      case luci::CircleOpcode::CONV_2D:
+      switch (circle_node->opcode())
       {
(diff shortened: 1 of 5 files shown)


seanshpark:
I don't know much about execution planer. I think guys who are involved would be better :)

compiler/circle-execution-plan/src/ExecutionPlanner.cpp
```diff
@@ -308,11 +308,21 @@ void ExecutionPlanner::create_alloc_node_inform_vector(bool null_consts, bool nu
...
         scratchpad_size =
           null_scratchpad ? 0 : _scratchpad_helper->ComputeScratchpadSizeConv2d(conv);
         break;
       }
+      case luci::CircleOpcode::DEPTHWISE_CONV_2D:
+      {
+        const auto depthwise_conv =
+          loco::must_cast<const luci::CircleDepthwiseConv2D *>(circle_node);
+        scratchpad_size =
+          null_scratchpad
```
binarman, line 321: hmmm, maybe move this condition outside of switch?
I suppose it will apply to every operation in condition

binarman: Also, FYI: #8335

BalyshevArtem: Done

m-bronnikov (approved):
LGTM

Issue #8357, opened 2022-01-28
[luci/logex] Refactoring NodeSummaryBuilder


llFreetimell:
#### Removing LoC Workarounds

Currently, we applied following workarounds to avoid large LoC.

- Separating visit functions with alphabet grouping
https://github.com/Samsung/ONE/blob/3265df39970d76db5433815b8ba68a8180532645/compiler/luci/logex/src/FormattedGraph.cpp#L2248-L2256
- Just moving implementation to another function
https://github.com/Samsung/ONE/blob/3265df39970d76db5433815b8ba68a8180532645/compiler/luci/logex/src/FormattedGraph.cpp#L1468

Let's try to remove the workarounds!

#### Trivial updates

- Add `CircleBidirectionalSequenceLSTMOut`
- Remove unused codes
- Remove dependency with `pepper-str`

### How?

I will apply following things

- Strategy pattern for creating NodeSummaryBuilder
- Template method pattern for building summary
- Remove duplicated code with class inheritance

Then I expect following result :)

#### Before

![before](https://user-images.githubusercontent.com/24720192/151469394-12bdf759-af6b-4f9b-90aa-c755384db4c7.png)

#### After

![after](https://user-images.githubusercontent.com/24720192/151469406-666a1a34-968c-4442-8140-26e39fbfb0be.png)

---

#### Related PR/Issue

- Draft : #8358

#### Todo

- [x] Introduce helper functions as separate file #8391
- [x] Introduce `CircleNodeSummaryBuilder`
  - [x] #8403
  - [x] #8565
- [x] Implement `Circle*SummaryBuilder` for each CircleNode
  - [x] ABC #8483 
  - [x] DEF #8490 
  - [x] GHIJ #8514
  - [x] KLMN #8518
  - [x] OPQR #8533 
  - [x] STUV #8544
  - [x] WXYZ #8564
  - [x] CIRC, VIRT #8571
- [x] Replace `SummaryBuilderLet` with `CircleNodeSummaryBuilder` #8469
- [x] Remove unused codes #8576

llFreetimell:
Finally Done!

PR #8358, opened 2022-01-28
[DRAFT][luci/logex] Refactoring luci NodeSummaryBuilder

llFreetimell:
This draft is for refactoring luci NodeSummaryBuilder.

---

Parent Issue : #8357


seanshpark:
@llFreetimell , if there is no particular reason, could you revise implementation in .h to .cpp file?

compiler/luci/logex/src/CircleNodeSummaryBuilder.cpp
```diff
@@ -0,0 +1,178 @@
...
+std::shared_ptr<CircleNodeSummaryBuilder>
+CircleNodeSummaryBuilder::create_builder(const luci::CircleNode *node)
+{
+  switch (node->opcode())
+  {
+#define CIRCLE_NODE(OPCODE, CLASS)    \
+  case luci::CircleOpcode::OPCODE:    \
+  {                                   \
+    return std::make_shared<CLASS>(); \
+  }
```
seanshpark, line 34:
```suggestion
#define CIRCLE_NODE(OPCODE, CLASS)    \
  case luci::CircleOpcode::OPCODE: return std::make_shared<CLASS>();
```
I'm not sure how line count works but this seems to reduce line count of the function. If possible could you check this too?

llFreetimell:
> if there is no particular reason, could you revise implementation in .h to .cpp file?

Not special reason but just I agonized which one would be better for readability, all in .h or separating declaration and implementation :)
If separating seems better, I will move it :D

compiler/luci/logex/src/CircleNodeSummaryBuilder.cpp
llFreetimell: This is counted as one line :)
So LoC of `CircleNodeSummaryBuilder` is 192!

compiler/luci/logex/src/CircleNodeSummaryBuilder.h
```diff
@@ -0,0 +1,104 @@
...
+  virtual void build_attributes(const luci::CircleNode *node, const locop::SymbolTable *tbl,
+                                locop::NodeSummary &s)
+  {
+    // Do nothing for default
+  }
+
+  virtual void update_status(locop::NodeSummary &s) { s.state(locop::NodeDesc::State::Complete); }
+
+private:
+  std::shared_ptr<CircleNodeSummaryBuilder> create_builder(const luci::CircleNode *node);
```
seanshpark, line 99: Can we use `unique_ptr` here? (I'm not good at using `unique_ptr` vs `shared_ptr`)

compiler/luci/logex/src/CircleNodeSummaryBuilder.h
```diff
@@ -0,0 +1,104 @@
...
+      return false;
+
+    auto ptr_to_str = [](const void *ptr) {
+      std::stringstream ss;
+      ss << ptr;
+      return ss.str();
+    };
+
+    auto circle_node = loco::must_cast<const luci::CircleNode *>(node);
+    if (const auto builder = create_builder(circle_node))
```
seanshpark, line 49: I'm not sure I understand this correctly but this seems to create `CircleNodeSummaryBuilder` object every time a node has to print summary of the graph.
Could you please check how may times this is called for a Inception V3 model (any version) with `LUCI_LOG=100` set ?

compiler/luci/logex/src/CircleNodeSummaryBuilder.h
llFreetimell: Yeap, we can use `unique_ptr` :)
I will replace with it

compiler/luci/logex/src/CircleNodeSummaryBuilder.cpp
seanshpark: both are 192 same ?

compiler/luci/logex/src/CircleNodeSummaryBuilder.h
llFreetimell: 319 (nodes in IV3) * 12 (printing function called) = 3828 creation and I think this is not a big deal :D

---

FYI, current code is creating 4272 times T.T
It's because
- `SummaryBuilderLet<SB::GRP>(_tbl).build(node, s))` creates `SummaryBuilderLet<SB::GRP>`
- According to the type of node, `SummaryBuilderLet<SB::GRP>` can be created with maximum 9 trials.

https://github.com/Samsung/ONE/blob/4c9870d3b781fe1f422e082d4bdee3b1a1638d0e/compiler/luci/logex/src/FormattedGraph.cpp#L2241-L2256

seanshpark: I'd like to understand if current code do object allocation.
New `create_builder` will create an object `3828 ` times while running(circle2circle I guess).

> current code is creating 4272 times

To compare with `create_builder` which code is using allocation?

compiler/luci/logex/src/CircleNodeSummaryBuilder.cpp
llFreetimell: The number of `CIRCLE_NODE(...)` is 133 and each `CIRCLE_NODE(...)` is counted as 1.
As a result, LoC of `CircleNodeSummaryBuilder` is 192 because `CircleNodeSummaryBuilder` includes other member functions such as `validate(const luci::CircleNode *node)`, `update_status(locop::NodeSummary &s)`, etc :)

compiler/luci/logex/src/CircleNodeSummaryBuilder.h
llFreetimell: `SummaryBuilderLet<SB::GRP>(_tbl).build(node, s))` creates `SummaryBuilderLet<SB::GRP>`.
AFAIK, unless the class isn't Singleton(or static), it creates new class everytime.
So following two codes are same.

```cpp
   do
   {
     if (SummaryBuilderLet<SB::GRP>(_tbl).build(node, s))
       return true;
   } while (false)
```

```cpp
   do
   {
     auto builder = SummaryBuilderLet<SB::GRP>(_tbl);
     if (builder.build(node, s))
       return true;
   } while (false)
```

Am I misunderstand something...?

compiler/luci/logex/src/CircleNodeSummaryBuilder.cpp
seanshpark: I mean
```
#define CIRCLE_NODE(OPCODE, CLASS)    \
  case luci::CircleOpcode::OPCODE:    \
  {                                   \
    return std::make_shared<CLASS>(); \
  }
```
vs
```
#define CIRCLE_NODE(OPCODE, CLASS)    \
  case luci::CircleOpcode::OPCODE: return std::make_shared<CLASS>();
```

llFreetimell: Ah, both would be same.
But I used former one because concluding with semicolon is not recommended for macro.
(https://wiki.sei.cmu.edu/confluence/display/c/PRE11-C.+Do+not+conclude+macro+definitions+with+a+semicolon)

And maybe that's why following codes are wrapped with do-while statement :)
https://github.com/Samsung/ONE/blob/4c9870d3b781fe1f422e082d4bdee3b1a1638d0e/compiler/luci/logex/src/FormattedGraph.cpp#L2241-L2246

compiler/luci/logex/src/CircleNodeSummaryBuilder.h
seanshpark: > Am I misunderstand something...?

I'm also not sure... One thing to check is either
`SummaryBuilderLet<SB::GRP>(_tbl).build(node, s))` or `auto builder = SummaryBuilderLet<SB::GRP>(_tbl);`, where the allocation happens.
`make_unique` will use heap, but not sure about `SummaryBuilderLet<SB::GRP>(_tbl);`
If this is created in the heap, it will be the same allocation. Number is decreased so this is good. If this is created in the stack, memory management will be less heavy.

Anyway, as this is used when the LOG is enabled for debugging, concern is not critical.

BTW, is it possible not to use make_unique/make_shared?

compiler/luci/logex/src/CircleNodeSummaryBuilder.cpp
seanshpark: OK, thanks for checking!

> semicolon is not recommended for macro.

We can use like this for this case.
```cxx
#define CIRCLE_NODE(OPCODE, CLASS)    \
  case luci::CircleOpcode::OPCODE: return std::make_shared<CLASS>()
```

I think I missed about `\` that this is actually make in a single line.

compiler/luci/logex/src/CircleNodeSummaryBuilder.h
llFreetimell: > make_unique will use heap, but not sure about SummaryBuilderLet<SB::GRP>(_tbl);

It would be created on stack :)

```cpp
auto on_heap = new SummaryBuilderLet<SB::GRP>(_tbl); // SummaryBuilderLet<SB::GRP>*
auto on_stack = SummaryBuilderLet<SB::GRP>(_tbl); // SummaryBuilderLet<SB::GRP>
```

> BTW, is it possible not to use make_unique/make_shared?

Unfortunately, not using pointer is impossible T.T
One way to avoid using `make_unique` is just using raw pointer.
However, if we use raw pointer, memory leak can be occurred :(

seanshpark: OK, let's go as-is. When/if this becomes issue, let's revise this later.

compiler/luci/logex/src/CircleNodeSummaryBuilders.h
```diff
@@ -0,0 +1,820 @@
...
+  std::vector<std::string> get_input_names(const luci::CircleNode *)
+  {
+    return {"input", "reduction_indices"};
+  }
+
+  void build_attributes(const luci::CircleNode *node, locop::NodeSummary &s)
+  {
+    auto mean = loco::must_cast<const REDUCER_NODE *>(node);
+    s.args().append("keep_dims", to_str(mean->keep_dims()));
+  }
```
llFreetimell, line 67: These functions cannot be moved to cpp because these are member functions of template class.

compiler/luci/logex/src/CircleNodeSummaryBuilders.cpp
```diff
@@ -0,0 +1,1014 @@
...
+std::vector<std::string> CircleWhileSummaryBuilder::get_input_names(const luci::CircleNode *node)
+{
+  auto circle_while = loco::must_cast<const luci::CircleWhile *>(node);
+
+  auto input_names = std::vector<std::string>();
+  input_names.push_back("cond");
+  for (uint32_t i = 0; i < circle_while->input_count(); ++i)
+    input_names.push_back("input");
+
+  return input_names;
```
llFreetimell, line 990:
```suggestion
  return std::vector<std::string>(circle_while->input_count(), "input");
```

llFreetimell:

PR #8359, opened 2022-01-28
[luci/test] Enable SVDF test with tolerance

seanshpark:
This will introduce testing with tolerance and enable SVDF with it.


seanshpark:
For #8203 , from draft #8279

mhs4670go (approved):
LGTM

PR #8363, opened 2022-01-28
[one-cmds] Fix wrong target

mhs4670go:
This commit fixes wrong target for for loop.


seanshpark (approved):
LGTM

PR #8380, opened 2022-02-07
[one-cmds] Introduce make_cmd.py

mhs4670go:
This commit introduces make_cmd.py to one-cmds.

Related: https://github.com/Samsung/ONE/issues/8169


compiler/one-cmds/one-import-bcq
```diff
@@ -25,6 +25,7 @@ import subprocess
 import sys
 import tempfile
 
+import python.make_cmd as _make_cmd
```
seanshpark, line 28: I think `python.make_cmd` came from using folder name `python`.
Which looks a bit strange as the this is implemented in python.
What I could think of is like `onelib` or `cmdslib`; not using generic word that can confuse code readers.
What do you think?

mhs4670go: @seanshpark Good idea. It'd really confuse people. How about merge this and change all at once?

seanshpark: OK

seanshpark (approved):
LGTM

PR #8384, opened 2022-02-08
[luci/svc] Tidy shape/dtype inference header

seanshpark:
This will tidy shape/dtype inference header orders and remove unused.


llFreetimell (approved):
LGTM, thanks!

PR #8386, opened 2022-02-08
[record-minmax] Generate positive random int

jinevening:
This generates positive random int for indices.


jinevening:
Some Ops (ex: Gather) accepts positive indices only. Generating positive int would be helpful to easily quantize those models.

seanshpark (approved):
LGTM

seanshpark:
Q) for other nodes that work with negative values, is it OK with testing?

jinevening:
Random int generation in record-minmax was introduced quite recently. AFAIK, there is no test that uses negative int yet.

In the long term, it would be better to distinguish (1) inputs that accept only positive int (indices) from (2) inputs that accept both negative/positive int (ex: axis) for wide range of testing.

m-bronnikov (approved):
LGTM

PR #8387, opened 2022-02-08
[luci/pass] Add comment for RemoveUnnecessaryReshapePass

seanshpark:
This will add before/after graph comment for RemoveUnnecessaryReshapePass.


seanshpark:
Related: #5596

mhs4670go (approved):
LGTM

Issue #8385, opened 2022-02-08
[circle-quantizer] Fix bias quantization


jinevening:
## What

Fix bias quantization
- Reorder quantization steps
- Add tests to prevent such cases

## Why

Currently, input activation can be overwritten after bias is quantized. This causes mismatch between input scale and bias scale. To prevent this, we should reorder quantization steps.

Current quantization flow

Quantize activation -> Quantize weights -> Quantize bias -> Propagate concat param (backward) -> Quantize const input -> Set input/output dtype -> Propagate activation qparam (forward)

Changed flow

Quantize activation -> Propagate concat param (backward) -> Quantize const input -> Propagate activation qparam (forward) -> Quantize weights -> Quantize bias -> Set input/output dtype

Activation qparam should not be changed after bias is quantized.

jinevening:
All done. Refactoring (https://github.com/Samsung/ONE/pull/8388#issuecomment-1032214650) will be done after #8393 is done.

seanshpark:
@jinevening , what about `TF_style_maxpool getter/setter` ?

jinevening:
As mentioned in https://github.com/Samsung/ONE/pull/8388#issuecomment-1032214650, I think that it would be better to introduce `Context` struct, which contains all arguments including `TF_style_maxpool`.

Introducing `Context` is not directly related to this issue, so I plan to do it after #8393 is done.

seanshpark:
> Please tell me if you have a different view.

I also think `Context` is good so don't have different view.

> so I plan to do it after https://github.com/Samsung/ONE/issues/8393 is done.

OK :) thanks!

PR #8382, opened 2022-02-07
DRAFT: Reorder quantization steps

jinevening:
On going draft to reorder quantization steps for correct bias quantization.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related issue: https://github.com/Samsung/ONE/issues/8385

PR #8388, opened 2022-02-08
[luci] Fix bias scale quantization

jinevening:
This reorders quantization steps to ensure input scale is not changed after bias is quantized.

---
Related to: #8385
Draft PR: https://github.com/Samsung/ONE/pull/8382

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
+++ b/compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
@@ -56,6 +56,10 @@ public:
 public:
   bool run(loco::Graph *graph);
 
+public:
+  void TF_style_maxpool(bool val) { _TF_style_maxpool = val; }
+  bool TF_style_maxpool() const { return _TF_style_maxpool; }
+
 private:
   void set_input_type(loco::Graph *graph) const;
   void set_output_type(loco::Graph *graph) const;
@@ -66,6 +70,7 @@ private:
   QuantizationGranularity _granularity;
   loco::DataType _input_type;
   loco::DataType _output_type;
+  bool _TF_style_maxpool = false;
 };
 
 } // namespace luci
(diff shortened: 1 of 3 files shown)


seanshpark:
Change seems to have multiple purposes...
can we split them? or these should be in one for not to break the test?

jinevening:
I think unit tests will not be failed even when the PR is split, because this reordering is for more complex models such as image captioning. It would be possible to split PRs to do reordering one-by-one, but I'm not sure intermediate states (master branch) can correctly quantize complex models.

seanshpark:
OK, then can you remove introducing new `TF_style_maxpool` and check?

jinevening:
I've reverted TF_style_maxpool getter/setter. I'll add them after this PR is landed.

jinevening:
In another thought, I think it would be better to introduce `Context` struct so that we can easily add new attributes when doing quantization.

compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
```diff
@@ -45,9 +45,10 @@ class QuantizeWithMinMaxPass : public logo::Pass
 public:
   QuantizeWithMinMaxPass(loco::DataType input_model_dtype, loco::DataType output_model_dtype,
                          QuantizationGranularity granularity, loco::DataType input_type,
-                         loco::DataType output_type)
+                         loco::DataType output_type, bool TF_style_maxpool)
```
seanshpark, line 48: `TF_style_maxpool` is added but I counld search 'maxpool` in the issue and in the change message. Is this parameter necessary for reordering quantization steps?
Why I asked before and now is that from previous change I see (1) this introduction and (2) reorder of steps, which it would be better to split.

jinevening: Ah, I get the point.

Before this change, `TF_style_maxpool` was used by `PropagateQuantParamPass`, which is invoked after `QuantizeWithMinMaxPass` is called. This PR makes `PropagateQuantParamPass` be called inside `QuantizeWithMinMaxPass`, so the variable is moved together.

So, `TF_style_maxpool` is moved, not newly added.

seanshpark: OK got it :)

seanshpark:
@git-bot-sec help

seanshpark:
@nnfw-bot help

seanshpark:
@nnfw-bot test nncc-debug

seanshpark:
@nnfw-bot test nncc-release

seanshpark (approved):
LGTM

PR #8389, opened 2022-02-08
DRAFT: onnx conv2d with different padding

seanshpark:
on-going draft to import onnx conv2d with different padding


seanshpark:
`Conv2d-pad/__init__.py` will produce ONNX file with one `Conv` with pads `1, 0, 1, 0`

When converted to NHWC circle, there seems unnecessary `Transpose-Reshape` pair before `Conv2D`

image

seanshpark:
Forgot to add `substitute_transpose_to_reshape` -_-;

PR #8391, opened 2022-02-08
[luci/logex] Introduce CircleNode SummaryBuilder Helper functions

llFreetimell:
This commit introduces `CircleNodeSummaryBuilderHelper`,
which includes helper functions for `CircleNodeSummaryBuilder`.

---

Parent Issue : #8357 
Draft : #8358

### DIFF
--- a/compiler/luci/logex/src/CircleNodeSummaryBuilderHelper.h
+++ b/compiler/luci/logex/src/CircleNodeSummaryBuilderHelper.h
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_LOGEX_CIRCLE_NODE_SUMMARY_BUILDER_HELPER__
+#define __LUCI_LOGEX_CIRCLE_NODE_SUMMARY_BUILDER_HELPER__
+
+#include <luci/IR/AttrFusedActFunc.h>
+#include <luci/IR/AttrPadding.h>
+#include <luci/IR/AttrMirrorPadMode.h>
+#include <luci/IR/AttrStride.h>
+#include <luci/IR/AttrFilter.h>
+#include <luci/IR/CircleOpcode.h>
+#include <loco/IR/DataType.h>
+
+#include <string>
+
+namespace luci
+{
+
+std::string to_str(bool value);
+std::string to_str(loco::DataType type);
+std::string to_str(luci::FusedActFunc fused);
+std::string to_str(luci::Padding padding);
+std::string to_str(luci::MirrorPadMode mode);
+std::string to_str(const luci::Stride *stride);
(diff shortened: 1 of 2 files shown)


compiler/luci/logex/src/CircleNodeSummaryBuilderHelper.h
```diff
@@ -0,0 +1,45 @@
...
+namespace luci
+{
+
+std::string to_str(bool value);
+std::string to_str(loco::DataType type);
+std::string to_str(luci::FusedActFunc fused);
+std::string to_str(luci::Padding padding);
+std::string to_str(luci::MirrorPadMode mode);
+std::string to_str(const luci::Stride *stride);
+std::string to_str(const luci::Filter *filter);
```
seanshpark, line 39: These are generic not only for logging but also for other message purposes.
That is, these may later (not sure as of now) produce duplicates in link time.
Thus (1) add inner namespace such as `log` to prevent this or (2) we can move these to common module (maybe need to add new one)

compiler/luci/logex/src/CircleNodeSummaryBuilderHelper.h
```diff
@@ -0,0 +1,45 @@
...
+
+std::string circle_opname(luci::CircleOpcode opcode);
```
seanshpark, line 41: (maybe this too)

compiler/luci/logex/src/CircleNodeSummaryBuilderHelper.h
llFreetimell: > That is, these may later (not sure as of now) produce duplicates in link time.

I copied this functions from current codes.

https://github.com/Samsung/ONE/blob/6b94dbb780d8514ff8d876ac3705ae02de535cb9/compiler/luci/logex/src/FormattedGraph.cpp#L84

https://github.com/Samsung/ONE/blob/6b94dbb780d8514ff8d876ac3705ae02de535cb9/compiler/luci/logex/src/FormattedGraph.cpp#L105

Therefore, it is only used here and ans as a result duplication would not appear for now.
But for preventing future duplication, I can apply following steps :)

- Change namespace to `log` for now
- And then move each `to_str` functions to `luci/IR/*`
  - `to_str(luci::FusedActFunc fused)` --> `luci/IR/AttrFusedActFunc.h`
  - `to_str(luci::Padding padding)` --> `luci/IR/AttrPadding.h`
  - ...

compiler/luci/logex/src/CircleNodeSummaryBuilderHelper.h
llFreetimell: This will not be used in other modules because this function returns `circle.ADD`, `circle.ABS`, and so on.
So this function is hard to be moved to other modules(or create new module) because `circle.*` is only used here and these are unofficial names for each CircleNode :)

compiler/luci/logex/src/CircleNodeSummaryBuilderHelper.h
seanshpark: > I copied this functions from current codes.

Current is anonymous namespace; it's local.

> Change namespace to log for now

Just to be sure, is it `luci::log` or just `log` ?

> And then move each to_str functions to luci/IR/*

I'm not sure putting in `IR` is good.

compiler/luci/logex/src/CircleNodeSummaryBuilderHelper.h
seanshpark: > This will not be used in other modules

OK. Then why move to separate helper? we can just leave in where it was.

llFreetimell: > we can just leave in where it was.

I think separating may be good but if leaving is much better, I will leave them as anonymous namespace in `CircleNodeSummaryBuilders.cpp` :)
If so, https://github.com/Samsung/ONE/pull/8391#discussion_r802146620 can be resolved too :D

llFreetimell:
Let's use these functions in anonymous namespace!
As a result, this PR is no more needed. Close!

Issue #8396, opened 2022-02-08
[one-cmds] one-optimize failure


seanshpark:
Daily release failed
> File "/usr/share/one/bin/one-optimize", line 28, in
import python.make_cmd as _make_cmd
ModuleNotFoundError: No module named 'python.make_cmd'

seanshpark:
@mhs4670go , PTAL

seanshpark:
#8380 seems related

seanshpark:
closeing as solved

PR #8402, opened 2022-02-09
[debian] Add make_cmd.py to install list

mhs4670go:
This commit adds make_cmd.py to install list.

Related: #8396 


seanshpark (approved):
LGTM

PR #8403, opened 2022-02-09
[luci/logex] Introduce CircleNodeSummaryBuilder

llFreetimell:
This commit introduces `CircleNodeSummaryBuilder`,
which builds summary for `CircleNode`s.

---

Parent Issue ; #8357 
Draft : #8358


seanshpark:
You are adding new files. This prevents history tracking. Is this really necessary ?

llFreetimell:
Original files are not deleted and only new files are added.
So current history will not be deleted and they can be tracked in original files :)

mhs4670go (approved):
LGTM

Issue #8406, opened 2022-02-10
[luci] Create a new const when quantizing const inputs


jinevening:
## What 

Let's create new const nodes when quantizing const inputs

## Why

Currently cont inputs are in-place quantized. (in `quantize_const_inputs`)

It was for ease of implementation, and did not cause any problem before quant steps are reordered (https://github.com/Samsung/ONE/issues/8385).

But now it becomes a problem when the const input is used both as an activation and a bias.

seanshpark:
> But now it becomes a problem when the const input is used both as an activation and a bias.

You mean, when constant is shared, quantization should be appliced differntly?

jinevening:
> You mean, when constant is shared, quantization should be applied for each case?

Exactly :)

For example, let's assume there is a zero constant used as the input of ADD and the bias of CONV.

They should be quantized in a different way, so we need to create separate consts.

jinevening:
All done.

PR #8410, opened 2022-02-10
DRAFT: Create new node for quantized const inputs

jinevening:
On going draft to create new nodes for quantized const inputs.

---
Related to: https://github.com/Samsung/ONE/issues/8406

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1197,145 +1197,185 @@ void propagate_one_hot_quantparam(luci::CircleOneHot *one_hot, loco::DataType qu
   quant_input(&CircleOneHot::off_value, &CircleOneHot::off_value);
 }
 
-/**
- * @brief Quantize const input tensors using min/max of const values
- */
-void quantize_const_inputs(luci::CircleNode *node, loco::DataType output_type)
+// Quantize constant input activation of a node
+// The input of a node is quantized if it is
+// 1. Constant (instance of CircleConst*)
+// 2. Activation (other inputs e.g., weights, bias, axis, etc should not be quantized here)
+struct QuantizeConstInputActivation final : public luci::CircleNodeMutableVisitor<void>
 {
-  auto opcode = node->opcode();
-  auto arity = node->arity();
+  QuantizeConstInputActivation(loco::DataType output_type) : _output_type(output_type) {}
 
-  loco::Node *input_node{nullptr};
-  luci::CircleConst *const_node{nullptr};
+private:
+  loco::DataType _output_type;
 
-  switch (opcode)
-  {
-    case luci::CircleOpcode::CONV_2D:
-    case luci::CircleOpcode::DEPTHWISE_CONV_2D:
-    case luci::CircleOpcode::FULLY_CONNECTED:
-    case luci::CircleOpcode::INSTANCE_NORM:
-    case luci::CircleOpcode::PRELU:
-    case luci::CircleOpcode::TRANSPOSE_CONV:
-      // Handled in QuantizeWeights and QuantizeBias
-      break;
+// Skip NODE
+#define SKIP(NODE) \
+  void visit(NODE *) {}
 
-    case luci::CircleOpcode::CONCATENATION:
-      // Handled in propagate_concat_quantparam
-      break;
-
-    case luci::CircleOpcode::LOGICAL_OR:
-      // Inputs of logical Ops are bool, thus not quantized
-      break;
-
-    case luci::CircleOpcode::ARG_MAX:
-    case luci::CircleOpcode::ARG_MIN:
-    case luci::CircleOpcode::BATCH_TO_SPACE_ND:
-    case luci::CircleOpcode::GATHER:
-    case luci::CircleOpcode::LOCAL_RESPONSE_NORMALIZATION:
-    case luci::CircleOpcode::MEAN:
-    case luci::CircleOpcode::MIRROR_PAD:
-    case luci::CircleOpcode::PAD:
-    case luci::CircleOpcode::REDUCE_ANY:
-    case luci::CircleOpcode::REDUCE_PROD:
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1197,145 +1197,185 @@ void propagate_one_hot_quantparam(luci::CircleOneHot *one_hot, loco::DataType qu
...
+private:
+  loco::DataType _output_type;
+
+// Skip NODE
+#define SKIP(NODE) \
+  void visit(NODE *) {}
+
+// INPUT_NAME is the only activation of NODE
+#define QUANTIZE_SINGLE_CONST_INPUT(NODE, INPUT_NAME)           \
+  void visit(NODE *node)                                        \
```
seanshpark, line 1217: using macro will expand, why not use function?

jinevening: That's because we need the name of each node's input in `visit` function.

For example, for ArgMax, we need the string `input` to call `node->input()`.  For Elu, we need the string `features` to call `node->features()`.

seanshpark: Ah, there was similare case in shape inference
```cxx
template <class CIRCLENODE> loco::NodeShape use_inputs(const CIRCLENODE *node)
{
  auto inputs_shape = luci::shape_get(node->inputs()).template as<loco::TensorShape>();
  return loco::NodeShape{inputs_shape};
}
```
likewise can something this work?
```cxx
template <class CIRCLENODE> void use_input(CIRCLENODE *node)
{
  auto input = node->input();
  auto const_node = dynamic_cast<luci::CircleConst *>(input);
  if (const_node && !is_quantized(const_node))
  {
    auto new_const = luci::clone(const_node);
    quant_const(new_const, _output_type);
    node->input(new_const);
  }
}
```

jinevening: Actually, that was my first attempt.

But I realized that I should make templates for different input names. For example,

For nodes with `input`
```
template <class CIRCLENODE> void use_input(CIRCLENODE *node)
{
  auto input = node->input();
  auto const_node = dynamic_cast<luci::CircleConst *>(input);
  if (const_node && !is_quantized(const_node))
  {
    auto new_const = luci::clone(const_node);
    quant_const(new_const, _output_type);
    node->input(new_const);
  }
}
```

For nodes with `features`
```
template <class CIRCLENODE> void use_features(CIRCLENODE *node)
{
  auto input = node->features();
  auto const_node = dynamic_cast<luci::CircleConst *>(input);
  if (const_node && !is_quantized(const_node))
  {
    auto new_const = luci::clone(const_node);
    quant_const(new_const, _output_type);
    node->features(new_const);
  }
}
```
and so on.

Nodes have various names for their inputs, e.g., input, features, x, logits, params, .. so I'd like to avoid writing templates for all of them.

seanshpark: Actually there are multiple input so I suggested to write multiple template codes...

But then, can this work?
```cxx
template <class CIRCLENODE, TYPE INPUT> void use_input(CIRCLENODE *node)
{
  auto input = node->INPUT();
  ...
  node->INPUT(new_const);
}
```

jinevening: That will not work as INPUT is not a typename.

If we have a way to set args of `loco::Node`, we may reduce duplicate codes, but I guess that is a more complex decision.

seanshpark: Um.. I'll check how this can be possible (maybe not)...

Anyway I would like to reduce codes from macro expansion.

So, what I could think next is 
(1) for each input method name, we declare `void visit_input(NODE *node)`, `void visit_x(NODE *node)` and so on for input, x, feature, params, logits, tensor, ...
(2) define macro `void visit_INPUT(NODE *node)` with body as you defined the macro
(3) using the macro, we provide implements of `visit_input`, `visit_x` and so on

I'll try this in shape inference

seanshpark: Can you try something this?
```cxx
#define DEFINE_VISIT_INPUT(INPUT_NAME) \
template <class CIRCLENODE> void visit_##INPUT_NAME(const CIRCLENODE *node) \
{                                                             \
  auto input = node->INPUT_NAME();                            \
  auto const_node = dynamic_cast<luci::CircleConst *>(input); \
  if (const_node && !is_quantized(const_node))                \
  {                                                           \
    auto new_const = luci::clone(const_node);                 \
    quant_const(new_const, _output_type);                     \
    node->INPUT_NAME(new_const);                              \
  }                                                           \
}

DEFINE_VISIT_INPUT(input);
DEFINE_VISIT_INPUT(x);
DEFINE_VISIT_INPUT(logits);
...
```

jinevening: > Anyway I would like to reduce codes from macro expansion.

Is it possible to reduce code size by using template rather than macro? Template will also increase code size. It will generate code for each class during compilation.

> Can you try something this?

I can try your approach, but the compiled binary size will be almost the same, while user code becomes longer.

seanshpark: > Is it possible to reduce code size by using template rather than macro? Template will also increase code size. It will generate code for each class during compilation.

Ah, yes, you are right. I forgot this when started just from tought of using a function.

If you like you can try with template + macro or go with current macro.

jinevening: https://github.com/Samsung/ONE/pull/8417 is refactoring of existing template-based codes, but now I'm writing new codes. So I'd like to go with the macro. PTAL at https://github.com/Samsung/ONE/pull/8414

PR #8412, opened 2022-02-11
[luci] Update quantized model verifier test

jinevening:
This makes the test reload variables for copied const nodes.

---
Related to: #8406
Draft PR: https://github.com/Samsung/ONE/pull/8410

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
@@ -127,19 +127,6 @@ void quantize_and_verify_with_wrong_type(luci::test::TestIOGraph *g, Type quanti
   verifier.verify(g->g());
 }
 
-void quantize_and_verify_with_wrong_type(luci::test::TestIOGraph *g, Type quantized_dtype,
-                                         Granularity granularity, Type wrong_dtype,
-                                         luci::CircleNode *target)
-{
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -1197,25 +1171,34 @@ class AddTestGraph final : public SimpleTestGraph
...
-      quantize_and_verify_with_wrong_type(&g, type, granularity, wrong_dtype, node)); \
+#define TEST_WITH_WRONG_TYPE_TARGET(graph, type, granularity, wrong_dtype, target) \
+  do                                                                               \
+  {                                                                                \
+    graph g;                                                                       \
+    g.init();                                                                      \
+    auto node = loco::must_cast<luci::CircleNode *>(target);                       \
+    luci::QuantizeWithMinMaxPass pass(Type::FLOAT32, type, granularity);           \
+    pass.run(g.g());                                                               \
+    auto after_node = loco::must_cast<luci::CircleNode *>(target);                 \
```
jinevening, line 1182: For reviewers: This is a key change in this PR.

This PR is necessary for #8406.

`target` is an expression, e.g., `g.y()`, not just a variable. We should evaluate `target` after quantization, because `node` is not valid here as it is replaced with a new const node.

PR #8413, opened 2022-02-11
[debian] Modify .install

mhs4670go:
This commit modifies one-compiler.install for applying previous relocation of python lib.


seanshpark (approved):
LGTM

PR #8417, opened 2022-02-11
[luci/svc] Use macro for similar methods in shape inference

seanshpark:
This will replace similar method implementations with macro.


jinevening (approved):
LGTM

lemmaa:
@seanshpark , ah, that's my mistake. There was a problem setting the permissions of the new group. Solved! :)

PR #8419, opened 2022-02-11
[one-cmds] Cleanup onnx_legalizer.py code

binarman:
This PR:
- marks internal methods using underscore
- fixes stand-alone script mode (used as a stand-alone tool, not library)
- simplifies interface of some functions
- removes unused variables

Related issue: #8148 

### DIFF
--- a/compiler/one-cmds/onnx_legalizer.py
+++ b/compiler/one-cmds/onnx_legalizer.py
@@ -45,11 +45,11 @@ class LegalizeOptions:
     unroll_lstm = False
 
 
-def reverse_str(s):
+def _reverse_str(s):
     return ''.join(reversed(s))
 
 
-def parse_tensor_name(name):
+def _parse_tensor_name(name):
     """Splits tensor name to base part and serial number
 
     Most of tensor names have following format: "tensor_123".
@@ -64,15 +64,15 @@ def parse_tensor_name(name):
     Returns:
         tuple of str, int: base name and serial number of tensor
     """
-    rev = reverse_str(name)
+    rev = _reverse_str(name)
     m = re.match('(\d*)(.*)', rev)
     if m.groups()[0] != '':
-        return (reverse_str(m.groups()[1]), int(reverse_str(m.groups()[0])))
+        return (_reverse_str(m.groups()[1]), int(_reverse_str(m.groups()[0])))
     else:
-        return (reverse_str(m.groups()[1]), 0)
+        return (_reverse_str(m.groups()[1]), 0)
 
 
-class ModelTransformerHelper:
+class _ModelTransformerHelper:
     """Helper for onnx model transformation
 
     This helper is used for convenient operation replacement in onnx model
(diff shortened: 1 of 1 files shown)


compiler/one-cmds/onnx_legalizer.py
```diff
@@ -1040,8 +1052,14 @@ def legalize(model, options):
...
+              '\n'
+              '    In stand-alone utility mode this tool provides basic funtionality\n'
+              '    If you want to have more control over applied transformations, use this legalizer as a library')
         exit(1)
+    options = LegalizeOptions()
+    options.unroll_lstm = True
+    options.unroll_rnn = True
     model = onnx.load(sys.argv[1])
-    legalize(model)
+    legalize(model, options)
```
binarman, line 1064: Added options here, to make it possible run this functionality as an stand-alone program

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -292,37 +292,37 @@ def delete_marked_nodes(self):
             self._model.graph.node.remove(node)
 
 
-class Info:
+class _TensorInfo:
```
binarman, line 295: rename class, to make it is more specific

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -577,6 +578,9 @@ def legalize_RNN(transformer, tensor_infos, node):
         if attr.name == 'layout':
             layout = attr.i
 
+    if len(activation_alpha) > 0 or len(activation_beta) > 0:
```
binarman, line 581: Added yet another check

compiler/one-cmds/onnx_legalizer.py
```diff
@@ -967,6 +976,9 @@ def legalize_LSTM(transformer, tensor_infos, node):
...
+        raise NotImplementedError('Unsupported parameters for LSTM activations')
```
binarman, line 980: Added yet another check

binarman:
For reviewers:
I commented interesting places, other changes are method renaming (`abc` -> `_abc`) and unused variables cleanup.

compiler/one-cmds/onnx_legalizer.py
mhs4670go: @binarman (optional) Isn't it inconvenient to use python script as both of library and stand-alone? I mean, it would have better maintainability to split into separate file - a file that has method as a library and another one for main. Seems to me that this script is too big to me.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM thank you!

compiler/one-cmds/onnx_legalizer.py
binarman: @mhs4670go 
Hmm, I am using script as a stand-alone tool simply for debug/experiment purposes, so maybe it is a good idea to remove or move it in separate file :thinking: 

About breaking this file in parts - it is probably a good idea, if we plan to extend it =)

PR #8423, opened 2022-02-14
[tflchef] Remove nnkit dependency for tests

seanshpark:
This will remove nnkit dependency check for running tests.


mhs4670go (approved):
LGTM

PR #8407, opened 2022-02-10
DRAFT: Compiler FE ARM32 build

seanshpark:
on-going draft to cross build for ARM32

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #8424, opened 2022-02-14
[tflchef] Use mio_tflite260

seanshpark:
This will revise to use mio_tflite260_validate as tflchef is using
mio_tflite260.


seanshpark:
from draft #8407

Issue #8395, opened 2022-02-08
[one-cmds] add onnx_legalizer into debian package


seanshpark:
Discussion about adding `onnx_legalizer.py` to debian package release.

seanshpark:
@binarman , today in our office we had some discussion about onnx_legalizer.py script to include in debian release package. And I also heard you will be busy leading another project :)
There will be some tasks to do before adding to release, like adding some tests and some tidy in the script where some variables are not used... and so on...

I'd like to ask if you can continue on doing what you planned to do or
if I could continue the whatever you planned and the release jobs.
I'd be glad if you can share your thougts and also plans.

binarman:
@seanshpark 
For now I was planning to refactor and document code properly, so it will be more "maintainable", and other people could change it with ease.

I am not sure about my log-term contribution plan at this moment, but I think I can participate in this release, 
What kind of tests should be implemented?

seanshpark:
> I think I can participate in this release,

Thank you!

> What kind of tests should be implemented?

Like the tests in `one-cmds/tests` that can check it won't abnormally exit.
And if possible, add correctness tests with actually running the model with random inputs and compare outputs before and after.

seanshpark:
Q) does existing onnx_leaglizer depend on onnx or onnx-tf versions? if so, could you add some comments on the file?

binarman:
> does existing onnx_leaglizer depend on onnx or onnx-tf versions

This question is kind of tricky, because currently onnx_legalize can operate on **models** with "ai.onnx Opset version" (see [this table](https://github.com/onnx/onnx/blob/main/docs/Versioning.md#released-versions)) <=12, but **imported python onnx module** could have any version (to be precise with versions 1.0+, It will crash with pre release versions)

binarman:
Recurrent models for testing

Compiled them using [PyTorch res module](https://github.com/Samsung/ONE/tree/e4777b0f01999b519adb26b261a4018c056dc85f/res/PyTorchExamples) using `ptem.py` utility

[test_onnx_recurrent_models.zip](https://github.com/Samsung/ONE/files/8067909/test_onnx_recurrent_models.zip)

seanshpark:
Closing as currently related tasks are under way.

binarman:
some simple negative tests with models with "variable" input length
[neg_test_onnx_recurrent_models.zip](https://github.com/Samsung/ONE/files/8137183/neg_test_onnx_recurrent_models.zip)

[...]

PR #8429, opened 2022-02-15
[DRAFT] [one-cmds] Add tests for onnx_legalizer.py

binarman:
This PR adds tests to check onnx_legalizer correctly modifies.

Related issue: #8395


binarman:
+cc @seanshpark @mhs4670go 

I want to break this PR in three:
- python scripts
- infrastructure (cmake, prepare env, etc.)
- actual test files

compiler/one-cmds/one-prepare-venv
```diff
@@ -35,6 +35,7 @@ fi
 
 VER_TENSORFLOW=2.6.0
 VER_ONNX=1.10.1
+VER_ONNXRUNTIME=1.10.0
```
seanshpark, line 38: Is this available in both Ubuntu 18.04 and 20.04 ?

seanshpark:
@mhs4670go , can you point out which files should be added to debian include list?

mhs4670go:
> @mhs4670go , can you point out which files should be added to debian include list?

I think there's nothing to be added.

compiler/one-cmds/one-prepare-venv
binarman: Checked, everything works.

P.s. to satisfy my curiosity, tried to install it without venv: 20.04 - ok, 18.04 - fail, highest available version is 1.3.0

binarman:
@seanshpark 
I have a question: do we need negative tests for one-cmds? In particular for this tool

seanshpark:
> I have a question: do we need negative tests for one-cmds? In particular for this tool

SE has a guide to provide negative tests for unit tests. (not sure about UI commands)
I think it would be nice to have them. Personally not sure as I didn't think much...
I would say, if you feel safe to have some, would be better to add cases that you already know and you can take your time.

seanshpark:

Issue #8443, opened 2022-02-17
[common-artifacts] Ubuntu 20.04 no tensorflow==1.13.2 


seanshpark:
On Ubuntu 20.04, I got this error while building

`No matching distribution found for tensorflow==1.13.2`

Maybe it's time to deprecate TensorFlow 1.X

seanshpark:
ping @lemmaa

lemmaa:
@seanshpark , Sorry for the late reply due to vacation. It seems to be a natural decision to stop supporting TF1.13.2 in Ubuntu 20.04 or later if it is maintained in the 18.04 version.

ragmani:
For more information, python3.8 can install only tensorflow 2.2 or later via using pip and python 3.8 is installed by default on Ubuntu 20.04.
https://www.tensorflow.org/install/pip#system-requirements

seanshpark:
done

PR #8444, opened 2022-02-17
DRAFT [common-artifacts] deprecate TF1.13.2

seanshpark:
draft to remove TF1.13.2 virtual environment and check


seanshpark:
For #8443

PR #8401, opened 2022-02-09
[DRAFT] Add new attribute and update shape inference for FullyConnected

llFreetimell:
This draft is for adding new attribute `keep_num_dims` and updating
shape inference algorithm for `CircleFullyConnected`.

ONE-DCO-1.0-Signed-off-by: Seok NamKoong <seok9311@naver.com>

---

Parent Issue : #8400

Issue #8400, opened 2022-02-09
[Compiler] Add keep_num_dims attribute of FullyConnected


llFreetimell:
### What?

Current shape inference logic for FullyConnected has following issue
- Attribute `keep_num_dims` is not considered
- Some recipes violated ["Input rank should be 2 or 3"](https://github.com/tensorflow/tensorflow/blob/ea33c1e7a25d8025e8ee405ad8ab7be261798d76/tensorflow/lite/kernels/fully_connected.cc#L194) assertion
    - `res/TensorFlowLiteRecipes/FullyConnected_001`
    - `res/TensorFlowLiteRecipes/FullyConnected_002`
    - `res/TensorFlowLiteRecipes/YUV_TO_RGB_000`
    - `res/TensorFlowLiteRecipes/YUV_TO_RGB_U8_000`

Let's resolve above problems!
(Dynamic shape inference will not be handled this time)

#### Related PR/Issue

- Draft : #8401

#### Todo

- [ ] ~~Fix TensorFlowLiteRecipes which violate assertions~~
- [x] Add `keep_num_dims` to `tflchef` #8447
- [x] Add `keep_num_dims` to `luci/IR/Nodes/CircleFullyConnected` #8408
- [x] Add `keep_num_dims` to `tflite2circle` #8445
- [x] Add `keep_num_dims` to `circledump` #8446
- [x] Add `keep_num_dims` to `luci-interpreter` #8449
- [x] Add TensorFlowLiteRecipes for `FullyConnected` with `keep_num_dims` #8476
- [x] Import `keep_num_dims` in `luci/import` #8450
- [x] Import `keep_num_dims` in `luci/export` #8451
- [x] Update shape inference logic for `CircleFullyConnected` #8467

seanshpark:
@llFreetimell , please check if any recipe exist with `keep_num_dims`, if not, please add the recipe.

llFreetimell:
All done, close.

PR #8449, opened 2022-02-18
[luci-interpreter] Add keep_num_dims to FullyConnected

llFreetimell:
This commit adds `keep_num_dims` to `FullyConnected`.

---

Parent issue : #8400 
Draft : #8401

### DIFF
--- a/compiler/luci-interpreter/src/core/KernelParams.h
+++ b/compiler/luci-interpreter/src/core/KernelParams.h
@@ -89,6 +89,7 @@ struct DivParams
 struct FullyConnectedParams
 {
   Activation activation;
+  bool keep_num_dims = false;
 };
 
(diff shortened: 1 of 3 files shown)


jinevening (approved):
LGTM

compiler/luci-interpreter/src/core/KernelParams.h
```diff
@@ -89,6 +89,7 @@ struct DivParams
 struct FullyConnectedParams
 {
   Activation activation;
+  bool keep_num_dims = false;
```
jinevening, line 92: (not for this PR) For this file, our convention is not to set initial values. But I think it would be better to set them for SAM score.

binarman (approved):
LGTM

PR #8452, opened 2022-02-18
[luci] Add comments for QuantizeWithMinMax

jinevening:
This adds comments for how QuantizeWithMinMax works.

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1700,7 +1700,7 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
  *
  * Why quantization sequence was determined as above?
  * - Activation and weights should be quantized before bias (1->2->3). Input/Output
- *   type can be updated at the end (4->5).
+ *   dtype can be updated at the end (4->5).
  * - During activation quantization,
  *   - Backward propagation is performed earlier than forward propagation. This allows
  *     backward-propagated qpram to be overwritten during forward propagation.
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1691,8 +1691,8 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
  *   - Quantize using recorded min/max (QuantizeActivation)
  *   - Propagate qparam of concat backward for optimization (propagate_concat_quantparam)
  *   - Quantize const inputs + propagate qparam backward (QuantizeConstInputActivation)
- *   - Propagate qparam forward (PropagateQuantParamPass)
  *   - Quantize using pre-defined values (QuantizeSpecialActivation)
+ *   - Propagate qparam forward (PropagateQuantParamPass)
  * 2. Quantize Weights
  * 3. Quantize Bias
  * 4. Set input type
@@ -1706,8 +1706,8 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
  *     backward-propagated qpram to be overwritten during forward propagation.
  *     We made this decision as Ops for forward propagation (reshape, transpose, ..)
  *     are more common than backward propagation. TODO Check this decision is safe.
- *   - QuantizeSpecialActivation is called at last to make sure that the pre-defined
- *     qparam values are not updated anymore
+ *   - QuantizeSpecialActivation is called before forward propagation to make sure that
+ *     the pre-defined qparam values are propagated.
  */
 bool QuantizeWithMinMaxPass::run(loco::Graph *g)
 {
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -1665,6 +1665,39 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
   }
 }
 
+/**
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1665,6 +1665,39 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
...
+ *   - Quantize using recorded min/max (QuantizeActivation)
+ *   - Propagate qparam of concat backward for optimization (propagate_concat_quantparam)
+ *   - Quantize const inputs + propagate qparam backward (QuantizeConstInputActivation)
+ *   - Propagate qparam forward (PropagateQuantParamPass)
+ *   - Quantize using pre-defined values (QuantizeSpecialActivation)
+ * 2. Quantize Weights
+ * 3. Quantize Bias
+ * 4. Set input type
+ * 5. Set output type
+ */
```
seanshpark, line 1700: (optional) Recently there was sequence change in quantization steps. If possible adding a note why the steps are as-is will help understanding current steps are as like. In the future if someone may want to change the step for some reason and the note may help preventing if it is in some wrong direction.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1665,6 +1665,39 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
...
+ * - Weights: Const tensors of specific Ops (Conv, FC, ...)
+ * - Bias: Const tensors of specific Ops (Conv, FC, ...)
+ * - Others: padding value, one_hot value, axis, ..
+ *
+ * Activation is quantized in different ways
+ * 1. For non-constant activation, quantize using recorded min/max
+ * 2. For constant activation, quantize using min/max of its value
+ * 3. For some Ops (ex: pad_v2), output qparam is used as input qparam (backward propagation)
+ * 4. For some Ops (ex: reshape), input qparam is used as output qparam (forward propagation)
+ * 5. For some Ops (ex: tanh), output qparam has pre-defined values
```
llFreetimell, line 1682: Just for curiosity!

In following example, which one should be quantized first? Reshape or PadV2?
```
PadV2 --- activation --- Reshape
```

jinevening: In our implementation, PadV2 -> Reshape.

1. PadV2's ofm is quantized in `QuantizeActivation` (using recorded min/max values)
2. qparam of PadV2's ofm is propagated to its input (`propagate_pad_v2_quantparam`) - backward prop
3. qparam of Reshape's ifm is propagated to its output (`PropagateQuantParamPass`) - forward prop

Reshape -> PadV2 would be also ok. as there is no conflict here.

A problematic situation would be Reshape - PadV2. The ofm of reshape (= ifm of PadV2) is quantized as below.
1. Quantize using recorded min/max in `QuantizeActivation`
2. Overwrite qparam of PadV2's ofm - backward prop
3. Overwrite qparam of Reshape's ifm - forward prop

Since we perform backward prop and then forward prop, the result of backward prop (step 2) can be overwritten by forward propagation (step 3).

I'm not sure the order (backward prop ->forward prop) is always safe for all Ops (that's why I did not explicitly write the order in the code). But for now, Ops for forward prop are much more common than Ops for backward prop, so I decided to do backward prop first.

llFreetimell (approved):
LGTM

seanshpark (dismissed):
LGTM thank you!

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1665,6 +1665,50 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
...
+ * 1. Quantize Activation
+ *   - Quantize using recorded min/max (QuantizeActivation)
+ *   - Propagate qparam of concat backward for optimization (propagate_concat_quantparam)
+ *   - Quantize const inputs + propagate qparam backward (QuantizeConstInputActivation)
+ *   - Quantize using pre-defined values (QuantizeSpecialActivation)
+ *   - Propagate qparam forward (PropagateQuantParamPass)
+ * 2. Quantize Weights
+ * 3. Quantize Bias
+ * 4. Set input type
+ * 5. Set output type
```
seanshpark, line 1699: I think `type` here would be S16 or U8. How about use `dtype` or `data type` ?

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -1665,6 +1665,50 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
...
+ *   - Quantize using pre-defined values (QuantizeSpecialActivation)
+ *   - Propagate qparam forward (PropagateQuantParamPass)
+ * 2. Quantize Weights
+ * 3. Quantize Bias
+ * 4. Set input dtype
+ * 5. Set output dtype
+ *
+ * Why quantization sequence was determined as above?
+ * - Activation and weights should be quantized before bias (1->2->3). Input/Output
+ *   type can be updated at the end (4->5).
```
seanshpark, line 1703: maybe `type` here too?

jinevening, line 1703:
```suggestion
 *   dtype can be updated at the end (4->5).
```

seanshpark (approved):
LGTM thank you!

jinevening:
@seanshpark Could you merge this?

llFreetimell (approved):
LGTM

Issue #8454, opened 2022-02-18
[quantizer] Create Context for QuantizeWithMinMax


jinevening:
## What

Let's make `Context` struct for QuantizeWithMinMax

## Why

> Please tell me if you have a different view.

I also think `Context` is good so don't have different view.

> so I plan to do it after https://github.com/Samsung/ONE/issues/8393 is done.

OK :) thanks!

_Originally posted by @seanshpark in https://github.com/Samsung/ONE/issues/8385#issuecomment-1033253609_

## TODO

- [x] Make Context
- [x] Remove existing member variables

PR #8455, opened 2022-02-18
[luci] Introduce Context to QuantizeWithMinMaxPass

jinevening:
This introduces Context to QuantizeWithMinMaxPass.

---
Related to: https://github.com/Samsung/ONE/issues/8454

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
+++ b/compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
@@ -31,6 +31,17 @@ namespace luci
  */
 class QuantizeWithMinMaxPass : public logo::Pass
 {
+public:
+  struct Context
+  {
+    loco::DataType input_model_dtype;
+    loco::DataType output_model_dtype;
+    QuantizationGranularity granularity;
+    loco::DataType input_type;
+    loco::DataType output_type;
+    bool TF_style_maxpool = false;
+  };
+
   // For backward-compatibility
   // TODO Remove this constructor
 public:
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
```diff
@@ -31,6 +31,17 @@ namespace luci
  */
 class QuantizeWithMinMaxPass : public logo::Pass
 {
+public:
+  struct Context
+  {
+    loco::DataType input_model_dtype;
```
seanshpark, line 37: (optional) OK with no default?

compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
```diff
@@ -62,6 +77,7 @@ class QuantizeWithMinMaxPass : public logo::Pass
   void set_output_type(loco::Graph *graph) const;
 
 private:
+  std::unique_ptr<Context> _ctx;
```
seanshpark, line 80: Q) maybe just `Context _ctx;`? need to use unique ptr?

compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
jinevening: AFAIK, we do not give initial value to loco::DataType. Unknown is a default value of loco::DataType.

compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
jinevening: I chose to use `unique_ptr` to avoid copy overhead.

For now, copy is not burdensome. But it can be if we pass information of many layers (ex: precision of each layer).

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

Issue #8399, opened 2022-02-09
[circle-quantizer] Quantization configuration file


jinevening:
## What

Let's design & implement quantization configuration file for mixed-precision quantization

## Attributes
- `default_quantization_dtype` : `uint8` or `int16`
- `layers` : Array of layers that will have quantization schemes different from `default_quantization_dtype`
  - Each layer will have following attributes
    - `name`
    - `precision` : `uint8` or `int16`
    - `granularity` : `layer` or `channel`

## TODO
- [x] infra: Add jsoncpp library
- [x] luci: Introduce Context to QuantizeWithMinMax and QuantizeDequantizeWeights
- [x] luci: Introduce LayerInfo to QuantizeWithMinMax and QuantizeDequantizeWeights
- [x] circle-quantizer: Parse quant config file using jsoncpp
- ~one-quantize: Update cmdline interface to accept quant config file (this will be done after mixed-precision is actually supported)~

Related to: https://github.com/Samsung/ONE/issues/8398

jinevening:
Copied from #8398 

@seanshpark : Have you decided which format to use? and which library?

@jinevening : JSON looks good to me due to the following reasons
- Widely used, MIT licensed parser: jsoncpp (https://github.com/open-source-parsers/jsoncpp)
- More concise than XML
- Allow array

seanshpark:
> Widely used, MIT licensed parser: jsoncpp 

@llFreetimell , can you please proceed internal license checking?

llFreetimell:
`jsoncpp` itself can be used.
Detailed license verification is available when actual source code is written.

jinevening:
Done.

PR #8465, opened 2022-02-21
[luci] Introduce Context to QuantizeDequantizeWeights

jinevening:
This introduces Context struct to QuantizedDequantizeWeights.

---
Related to: https://github.com/Samsung/ONE/issues/8399
Draft PR: https://github.com/Samsung/ONE/pull/8460

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/QuantizeDequantizeWeightsPass.h
+++ b/compiler/luci/pass/include/luci/Pass/QuantizeDequantizeWeightsPass.h
@@ -34,9 +34,9 @@ class QuantizeDequantizeWeightsPass : public logo::Pass
 public:
   struct Context
   {
-    loco::DataType input_model_dtype;
-    loco::DataType output_model_dtype;
-    QuantizationGranularity granularity;
+    loco::DataType input_model_dtype = loco::DataType::Unknown;
+    loco::DataType output_model_dtype = loco::DataType::Unknown;
+    QuantizationGranularity granularity = QuantizationGranularity::ChannelWise;
   };
 
 public:
--- a/compiler/luci/pass/include/luci/Pass/QuantizeDequantizeWeightsPass.h
+++ b/compiler/luci/pass/include/luci/Pass/QuantizeDequantizeWeightsPass.h
@@ -32,12 +32,29 @@ namespace luci
 class QuantizeDequantizeWeightsPass : public logo::Pass
 {
 public:
-  QuantizeDequantizeWeightsPass(loco::DataType input_model_dtype, loco::DataType output_model_dtype,
-                                QuantizationGranularity granularity)
-    : _input_model_dtype{input_model_dtype}, _output_model_dtype{output_model_dtype}, _granularity{
-                                                                                        granularity}
+  struct Context
+  {
+    loco::DataType input_model_dtype;
+    loco::DataType output_model_dtype;
+    QuantizationGranularity granularity;
+  };
+
+public:
+  QuantizeDequantizeWeightsPass(std::unique_ptr<Context> &&ctx) : _ctx{std::move(ctx)}
   {
     // DO NOTHING
+  }
+
+public:
+  QuantizeDequantizeWeightsPass(loco::DataType input_model_dtype, loco::DataType output_model_dtype,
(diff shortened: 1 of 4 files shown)


hyunsik-yoon:
I am not sure about the context about this change since draft (https://github.com/Samsung/ONE/pull/8460) does not seem to have this change. Could you explain a bit about the reason of this change?

jinevening:
@hyunsik-yoon It was my mistake. I forgot to update the draft PR. I'll update it.

> Could you explain a bit about the reason of this change?

We're going to support mixed-precision, and the precision information of layers should be passed to both `QuantizeDequantizeWeightsPass` (fake quantization) and `QuantizeWithMinMaxPass` (quantization), because their behaviors are coupled.

For example, if we fake quantize with u8, then we should quantize with u8 (no s16 fake quant -> u8 quant).

compiler/luci/pass/include/luci/Pass/QuantizeDequantizeWeightsPass.h
```diff
@@ -31,23 +31,38 @@ namespace luci
  */
 class QuantizeDequantizeWeightsPass : public logo::Pass
 {
+public:
+  struct Context
+  {
+    loco::DataType input_model_dtype = loco::DataType::Unknown;
+    loco::DataType output_model_dtype = loco::DataType::Unknown;
+    QuantizationGranularity granularity = QuantizationGranularity::ChannelWise;
```
jinevening, line 39: I've initialized variables to avoid #8468.

hyunsik-yoon, line 39: Can we create `QuantizationGranularity::Unknown` and initialized `granularity` with `unknown` just like `input_model_dtype` and `output_model_dtype`? 
Initializing to channel by dfault is good to some NPU but may not be good to some other mechanisms which ONE needs to support (in future, which I don't know now 😸).

jinevening: I totally agree with you. I've posted  #8472 for this issue :)

compiler/luci/pass/include/luci/Pass/QuantizeDequantizeWeightsPass.h
```diff
@@ -31,23 +31,38 @@ namespace luci
...
-                                                                                        granularity}
   {
-    // DO NOTHING
+    _ctx = std::make_unique<Context>();
+    {
+      _ctx->input_model_dtype = input_model_dtype;
+      _ctx->output_model_dtype = output_model_dtype;
+      _ctx->granularity = granularity;
+    }
   }
```
mhs4670go, line 58: How about use initializer list?
```cxx
QuantizeDequantizeWeightsPass(loco::DataType input_model_dtype, loco::DataType output_model_dtype,
                                QuantizationGranularity granularity) : _ctx{std::make_unique<Context>(Context{input_model_dtype, output_model_dtype, granularity})}
{
  // DO NOTHING
}
```
Or
```cxx
// need constructor of `Context`
QuantizeDequantizeWeightsPass(loco::DataType input_model_dtype, loco::DataType output_model_dtype,
                                QuantizationGranularity granularity) : _ctx{std::make_unique<Context>(input_model_dtype, output_model_dtype, granularity)}
{
  // DO NOTHING
}
```

seanshpark: I like @jinevening 's code as it's easy to read :)

hyunsik-yoon (approved):
LGTM, Thanks.

compiler/luci/pass/include/luci/Pass/QuantizeDequantizeWeightsPass.h
jinevening: In addition to @seanshpark 's comment, I'll add a new variable to `Context`, which may change the constructor signature. So, I'd like to keep the current style :)

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #8468, opened 2022-02-21
[luci/pass] Initialize Context member variables

llFreetimell:
This commit initializes `Context` member variables in `QuantizeWithMinMaxPass`.

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
+++ b/compiler/luci/pass/include/luci/Pass/QuantizeWithMinMaxPass.h
@@ -34,11 +34,11 @@ class QuantizeWithMinMaxPass : public logo::Pass
 public:
   struct Context
   {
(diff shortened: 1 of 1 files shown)


llFreetimell:
This PR is generated because of static analysis :)

jinevening:
Does our SE tool complain about uninitialized enum? I didn't know that..

AFAIK, there exist many uninitialized enum variables in our source code. We may need refactoring for them.

llFreetimell:
> Does our SE tool complain about uninitialized enum? I didn't know that..

I don't know exact reason but the tool sometimes complain, sometimes does not...
This time, it complained :(

> AFAIK, there exist many uninitialized enum variables in our source code. We may need refactoring for them.

I agree that there are uninitialized enum variables but the tool didn't complain yet @.@

jinevening (approved):
LGTM Thank you

PR #8470, opened 2022-02-21
[luci] Make cloned weights during fake quantization

jinevening:
This makes weights be cloned when fake-quantized.

### DIFF
--- a/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
+++ b/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
@@ -19,6 +19,7 @@
 
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/CircleNodeVisitor.h>
+#include <luci/Service/Nodes/CircleConst.h>
 #include <luci/Log.h>
 #include <loco/IR/TensorShape.h>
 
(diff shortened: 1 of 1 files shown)


jinevening:
Reason of this PR:

Until now, weights have been updated in-place when fake-quantized. However, this can cause a problem in the mixed-precision scenario.

Let's consider a case where the same weights are shared by different Ops (Conv1, Conv2). If Conv1 is u8-quantized and Conv2 is s16-quantized, then the weights should be fake-quantized separately. For this, we have to clone weights before do fake quantization.

I've separated visitors to use the setter of arguments for each `CircleNode`. Ex: `filter` for `CircleConv2D` and `weights` for `CircleFullyConnected`.

seanshpark:
it's a bit hard to read the changes, please split this PR or add comments to understand better and why the changes are in one PR.

jinevening:
> it's a bit hard to read the changes, please split this PR or add comments to understand better and why the changes are in one PR.

I understand. It looks very difficult to me either. I'll use this PR as a draft PR and make new smaller PRs.

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -263,88 +264,166 @@ struct QuantizeDequantizeWeights final : public luci::CircleNodeMutableVisitor<b
...
+    INFO(l) << "QuantizeDequantizeWeights visit node: " << node->name() << std::endl;
+
+    // TODO Check if we can replace dynamic_cast with loco::must_cast
+    auto weights = dynamic_cast<luci::CircleConst *>(node->filter());
+    if (not weights)
+      return false;
+
+    // Skip if this is already quantized
+    if (is_quantized(weights))
+      return false;
```
llFreetimell, line 361: How about creating separate private function like below? It seems to be duplicated for each operations :D

```cpp
bool check_weights(luci::CircleConst *weights)
{
    // TODO Check if we can replace dynamic_cast with loco::must_cast
    if (not weights)
      return false;

    // Skip if this is already quantized
    if (is_quantized(weights))
      return false;
    
    return true;
}
```

If so, we may be able to reduce duplicated codes as following :)

```suggestion
    if (!check_weights(node->filter()))
      return false;
```

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -263,88 +264,166 @@ struct QuantizeDequantizeWeights final : public luci::CircleNodeMutableVisitor<b
   loco::DataType output_type;
   QuantizationGranularity granularity;
 
-  // Quantize and dequantize input tensors of each node
-  bool visit(luci::CircleNode *node)
+private:
+  // Fake quantize weights (Only u8 quantization is supported)
```
llFreetimell, line 268: For safety, how about adding assert like following in `fake_quantize_lwq`? :D

```cpp
assert(output_type == loco::DataType::U8);
```

mhs4670go, line 268: I'm not sure but can it be checked in advance if it is for u8 or s16 or something?

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -263,88 +264,166 @@ struct QuantizeDequantizeWeights final : public luci::CircleNodeMutableVisitor<b
...
-      // Check if this is already quantized
-      if (is_quantized(circle_node))
-        continue;
+private:
+  // Fake quantize weights (u8/s16 quantization are supported)
+  void fake_quantize_cwq(luci::CircleConst *weights) const
+  {
+    // Find min/max per channel
+    std::vector<float> min;
+    std::vector<float> max;
```
llFreetimell, line 302: Unlike line 271, this is not initialized with `std::numeric_limits<float>::max()` and `std::numeric_limits<float>::lowest()`.

```cpp
    float min = std::numeric_limits<float>::max();
    float max = std::numeric_limits<float>::lowest();
```

And when I looked into `cal_minmax_per_channel` function, I couldn't find the initialization with `max()` and `lowest()` too :(
Is there no need to initialize with those values or did I missed something?

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
jinevening: Yes, it is checked before this pass is called (more specifically, in the circle-quantizer).

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
jinevening: This PR just extracts the logic from the existing source code, but the code seems dangerous.

I guess that 0 is set when min/max (std::vector) is resized. We can get the same result (quantized values) with the implicit initialization, but it would be better to specify the initial value (max(), lowest()) when we resize min/max vectors. Nice catch!

jinevening:
All done.

PR #8474, opened 2022-02-21
[luci] Apply fake quantization to visitor target

jinevening:
This applies fake quantization to visitor target, not args.

---
Related to: https://github.com/Samsung/ONE/pull/8470

### DIFF
--- a/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
+++ b/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
@@ -369,21 +369,15 @@ private:
     assert(output_type == loco::DataType::U8 || output_type == loco::DataType::S16);
     LOGGER(l);
     INFO(l) << "QuantizeDequantizeWeights visit node: " << node->name() << std::endl;
-    auto arity = node->arity();
-    for (uint32_t i = 0; i < arity; i++)
-    {
-      auto input_node = node->arg(i);
-      auto circle_node = loco::must_cast<luci::CircleNode *>(input_node);
 
-      if (not is_quantizable(input_node))
-        continue;
+    if (not is_quantizable(node))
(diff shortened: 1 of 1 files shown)


jinevening:
For reviewers:

Previously, fake quantization was applied to `arg` of visitor target (`node`).

This PR applies it directly to the visitor target (`node`). The change will not cause any difference in the final quantized values.

seanshpark (approved):
LGTM

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -369,21 +369,15 @@ struct QuantizeDequantizeWeights final : public luci::CircleNodeMutableVisitor<b
...
-    auto arity = node->arity();
-    for (uint32_t i = 0; i < arity; i++)
-    {
-      auto input_node = node->arg(i);
-      auto circle_node = loco::must_cast<luci::CircleNode *>(input_node);
 
-      if (not is_quantizable(input_node))
-        continue;
+    if (not is_quantizable(node))
+      return false;
```
hyunsik-yoon, line 374: not a scope of this PR but
all return statements in this method is `return false` and there is no `return true`.
Would `void visit(...)` be right signature?

seanshpark:
@nnfw-bot test nncc-release

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
jinevening: > Would void visit(...) be right signature?

Yes. It will be better to use void visitor, as we don't use return values (same for `QuantizeWithMinMaxPass`).

hyunsik-yoon (approved):

Issue #8473, opened 2022-02-21
[one-cmds] one-prepare-venv fails with pip install


seanshpark:
Sometimes in our internal CI, downloading whl file fails while running `one-prepare-venv`

> ConnectionResetError: [Errno 104] Connection reset by peer

Let's provide a way to use alternative path to download.

seanshpark:
done

PR #8475, opened 2022-02-22
[one-cmds] Revise one-prepare-venv to split onnx install

seanshpark:
This will revise one-prepare-venv to split onnx and onnx-tf install
in separate command.


seanshpark:
Related #8473

mhs4670go (approved):
LGTM

PR #8479, opened 2022-02-22
[infra] Introduce macro nnas_find_package_folder

seanshpark:
This will introduce CMake macro nnas_find_package_folder() that will
find package in infra/cmake and also provided folder.


seanshpark:
From draft #8407

NOTE: cmake with higher versions (Ubuntu 20.04, Tizen) will fail with recursive call when `find_package` is used for some packages. This will prevent that issue.

hyunsik-yoon (approved):

chunseoklee (approved):

hseok-oh (approved):

Issue #8397, opened 2022-02-08
[infra] CMake duplicate function from FlatBuffers-1.10 and FlatBuffers-1.12


seanshpark:
In `infra/cmake/packages`, `FlatBuffers-1.10` and `FlatBuffers-1.12` both provide `FlatBuffers_Generate` and `FlatBuffers_Target`. 
`FlatBuffers_Target` has difference in `target_link_libraries(${TGT} PUBLIC flatbuffers-1.10)` vs `target_link_libraries(${TGT} PUBLIC flatbuffers-1.12)` which may give problems.

seanshpark:
This was found while working on #8379.
Upgrading circle schema is needed; but still version mismatch is not good.
Maybe rename `FlatBuffers_Target` with version like `FlatBuffers_1_12_Target` ? (this seems a hack)

seanshpark:
#8499 may solve this issue

seanshpark:
this issue is resolved by using one FB 2.0

PR #8481, opened 2022-02-22
[Compiler FE] build for armv7l on tizen obs

chunseoklee:


hseok-oh:
@chunseoklee @seanshpark Why don't we use `flatbuffers` latest version (`2.0.0`) for all compiler and runtime modules? Tizen is using `2.0.0` to build both `tensorflow-1.13.1` and `tensorflow-2.7.0`.

chunseoklee:
> @chunseoklee @seanshpark Why don't we use flatbuffers latest version (2.0.0) for all compiler and runtime modules? Tizen is using 2.0.0 to build both tensorflow-1.13.1 and tensorflow-2.7.0

I am not against this. 
compiler side has no problem ?

seanshpark:
> I am not against this.
compiler side has no problem ?

I am also OK. Just didn't do any tests :)

seanshpark:
> Why don't we use flatbuffers latest version (2.0.0) for all compiler and runtime modules? 

If NPU compiler is also OK then we can try with 2.0.0 :)

seanshpark:
Related: #8397

ejjeong:
@chunseoklee Here's the list we discussed offline. (cc/ @hseok-oh)

```
angkor
hermes
hermes-std
loco
locomotiv
locop
logo
logo-core
luci-env
luci-import
luci-lang
luci-log
luci-logex
luci-pass
luci-plan
luci-profile
luci-service
mio-circle
mio-tflite260
oops
pepper-csv2vec
pepper-str
pepper-strcast
pp
```

PR #8483, opened 2022-02-22
[luci/logex] Replace SummaryBuilderLet::ABC

llFreetimell:
This commit replaces `SummaryBuilderLet::ABC` with `CircleNodeSummaryBuilders`.

---

Parent Issue : #8357 
Draft : #8358

### DIFF
--- a/compiler/luci/logex/src/CircleNodeSummaryBuilders.cpp
+++ b/compiler/luci/logex/src/CircleNodeSummaryBuilders.cpp
@@ -1,0 +1,404 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


compiler/luci/logex/src/CircleNodeSummaryBuilders.cpp
```diff
@@ -0,0 +1,404 @@
...
+namespace luci
+{
+
+std::vector<std::string> CircleNodeWithXSummaryBuilder::get_input_names(const luci::CircleNode *)
+{
+  return {"x"};
+}
+
+std::vector<std::string>
+CircleNodeWithINPUTSummaryBuilder::get_input_names(const luci::CircleNode *)
```
mhs4670go, line 123: Is it intentional?
```suggestion
CircleNodeWithInputSummaryBuilder::get_input_names(const luci::CircleNode *)
```

mhs4670go (approved):
LGTM

compiler/luci/logex/src/CircleNodeSummaryBuilders.cpp
llFreetimell: I wrote it as upper case to represent `input` of `node->input()`.
But if it looks weird, I can apply your suggestion :D

mhs4670go: Both way looks okay to me:)

llFreetimell: Then I will go as it is :D

PR #8485, opened 2022-02-22
[luci] Add layer parameter for quantization

jinevening:
This adds layer parameter for quantization.

---
Related to: https://github.com/Samsung/ONE/issues/8399
Draft PR: https://github.com/Samsung/ONE/pull/8460

### DIFF
--- a/compiler/luci/pass/src/CircleQuantizer.cpp
+++ b/compiler/luci/pass/src/CircleQuantizer.cpp
@@ -78,7 +78,7 @@ private:
   std::vector<Algorithm> _algorithms;
   std::map<AlgorithmParameters, const std::string> _algorithm_params;
   std::map<AlgorithmParameters, std::vector<std::string>> _multiple_params;
-  std::map<AlgorithmParameters, std::vector<std::shared_ptr<LayerParam>>> _multiple_layer_params;
+  std::map<AlgorithmParameters, std::vector<std::shared_ptr<LayerParam>>> _layer_params;
 };
 
 void QuantizeOptionsImpl::enable(Algorithm algo) { _algorithms.push_back(algo); }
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/src/CircleQuantizer.cpp
```diff
@@ -69,12 +70,15 @@ class QuantizeOptionsImpl final : public luci::CircleQuantizer::Options
...
   std::vector<std::string> params(AlgorithmParameters) const final;
+  void layer_params(AlgorithmParameters, std::vector<std::shared_ptr<LayerParam>> &) final;
+  std::vector<std::shared_ptr<LayerParam>> layer_params(AlgorithmParameters) const final;
   bool query(Algorithm) final;
 
 private:
   std::vector<Algorithm> _algorithms;
   std::map<AlgorithmParameters, const std::string> _algorithm_params;
   std::map<AlgorithmParameters, std::vector<std::string>> _multiple_params;
+  std::map<AlgorithmParameters, std::vector<std::shared_ptr<LayerParam>>> _multiple_layer_params;
```
seanshpark, line 81: is this need to be `shared_ptr` ?

mhs4670go, line 81: (optional) I'm not sure but just `_layer_params` without `multiple` looks okay because param"s" already means "multiple".

mhs4670go (approved):
LGTM

seanshpark:
Is this O(N) or O(N^2) or anything else?

compiler/luci/pass/src/CircleQuantizer.cpp
jinevening: I think there can be two other options.

1. Use `LayerParam`
  - This will cause copy whenever we access the parameter by `layer_params`. Copy overhead would be acceptable for now, but I just chose to avoid copy because LayerParam can be extended in the future.

2. Use `LayerParam*`
  - I didn't choose this option, because I don't want to track the lifetime of LayerParam.

seanshpark: Sorry for not asking what I wanted to ask. ONE usually use `unique_ptr`. this uses `shared_ptr`. reason is `unique_ptr` is lighter than `shared_ptr`. so I asked why use `shared_ptr`.

jinevening: Ah, I misunderstood your comment :)

> that is, is this shared among other objects that own this?

It is not shared for now, but I'm not sure it will be in the future. That is the reason I used `shared_ptr` instead of `unique_ptr`. Performance difference would be negligible.

jinevening:
> Is this O(N) or O(N^2) or anything else?

This PR just passes parameters and checks if the parameters are valid. So the algorithm would be O(N) (if N is the number of specified layers)

compiler/luci/pass/src/CircleQuantizer.cpp
seanshpark: OK :)

seanshpark (approved):
LGTM

compiler/luci/pass/src/CircleQuantizer.cpp
jinevening: _layer_params looks better. I've updated the name. Thanks.

mhs4670go (approved):
LGTM

Issue #8379, opened 2022-02-07
[Compile FE] Build ARM32 target 


seanshpark:
Let's enable build compiler frontend for ARM32 target.
- first module to enable is luci
- refer runtime compilation for ARM32
- to enable running inside Tizen

---

Why ARM32 not AArch64?
- Our Tizen target device runs in ARM32.
- From my point of view, ARM32 compilation test is to validate correct execution in ARM32.

seanshpark:
TODO
- [ ] prepare ARM32 rootfs
- [ ] cmake configure
   - [ ] build dependency libraries to ARM32
- [ ] ...

seanshpark:
Issue: flatc native is required to compile fbs file(s)
Issue: protoc, js_embed native is required to compile proto file(s) and generate `well_known_types_js`

seanshpark:
Issue
> compiler/luci/service/include/luci/Service/CircleTypeInference.h:22:10: fatal error: mio/circle/schema_generated.h: No such file or directory
 #include <mio/circle/schema_generated.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

seanshpark:
For protobuf, `find_package(protobuf EXACT 3.5.2 QUIET)` fails in cross compile.

seanshpark:
Run tests in RPi4
```
sudo apt-get install software-properties-common
sudo apt-get install build-essential cmake scons git
```

seanshpark:
HDF5 uses `TRY_RUN` which doesn't seem to work with cross compile...
Need to find some workarounds.

seanshpark:
> HDF5_CONFIG_DIR-NOTFOUND/hdf5-config.cmake

--> install has problem --> fixed
--> still `find_path` doesn't work for cross build...

seanshpark:
`record-minmax` requires `luci-interpreter`
- need to check kernel codes that use x86 specific codes

seanshpark:
tests are disable as gtest isn't found in cross compile.
- at least I want to run unit tests to run in target device
- later we can enable this in CI with target devices

seanshpark:
How to enable `common-artifacts` ?
- issue: python virtual envionment with TensorFlow 2.6 is needed

Maybe ?
- host drops all reference data
- target executes luci-interpreter and compare with reference data

seanshpark:
[...]

PR #8487, opened 2022-02-22
[infra] Fix order of stamp generation for ExternalBuild_CMake

seanshpark:
This will fix order of stamp generation after external cmake build is success.

### DIFF
--- a/infra/cmake/modules/ExternalBuildTools.cmake
+++ b/infra/cmake/modules/ExternalBuildTools.cmake
@@ -45,8 +45,6 @@ function(ExternalBuild_CMake)
   file(MAKE_DIRECTORY ${ARG_BUILD_DIR})
   file(MAKE_DIRECTORY ${ARG_INSTALL_DIR})
 
-  file(WRITE "${BUILD_STAMP_PATH}" "${PKG_IDENTIFIER}")
-
   execute_process(COMMAND ${CMAKE_COMMAND}
                             -G "${CMAKE_GENERATOR}"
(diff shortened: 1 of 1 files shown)


seanshpark:
For #8379, from draft #8407

This is to make BUILD_STAMP after build is success, where it was created before the build.

hyunsik-yoon (approved):

infra/cmake/modules/ExternalBuildTools.cmake
```diff
@@ -45,8 +45,6 @@ function(ExternalBuild_CMake)
   file(MAKE_DIRECTORY ${ARG_BUILD_DIR})
   file(MAKE_DIRECTORY ${ARG_INSTALL_DIR})
 
-  file(WRITE "${BUILD_STAMP_PATH}" "${PKG_IDENTIFIER}")
```
hseok-oh, line 48: It was designed to not retry build when failed. 

https://github.com/Samsung/ONE/blob/master/infra/cmake/modules/ExternalBuildTools.cmake#L27-L33

Please remove above comment.

seanshpark: OK, it was by design...
I had a problem while make a cross build of external project and had to remove the stamp to make it retry the build.
So I think by design is like while development, the developer should manually clean the stamp.
Once the development is done, external build should not break in normal cases.
If it does break, then there is some critical situation and should skip retry as it will always fail...

PR #8494, opened 2022-02-22
[souschef] Close parenthesis in CMake

seanshpark:
This will close missing parenthesis in CMakeLists.txt message.


mhs4670go (approved):
LGTM

PR #7215, opened 2021-07-13
[circle-quantizer] Propagate quantparam forward direction

jinevening:
This propagates quantparam in a forward direction.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

PR #8495, opened 2022-02-23
[luci] Rename Propagate QParam Pass

jinevening:
This adds direction of propagation (Forward) to pass name.

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/PropagateQParamForwardPass.h
+++ b/compiler/luci/pass/include/luci/Pass/PropagateQParamForwardPass.h
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#ifndef __LUCI_PROPAGATE_QUANT_PARAM_PASS_H__
-#define __LUCI_PROPAGATE_QUANT_PARAM_PASS_H__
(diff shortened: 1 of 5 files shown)


jinevening:
For reviewers:

I renamed the pass because backward propagation pass will be introduced.

llFreetimell (dismissed):
LGTM

hyunsik-yoon:
Can I ask a question?
Since qparam is propagated from output to input, 'backward' (not `forward`) sounds better to me. Maybe I misunderstand the term `forward` and `backward`. Could you explain a bit?

```
/**
 * @brief  Class to propagate quantization parameters of an operator's output to input
 */
struct PropagateQParamForwardPass final : public logo::Pass
```

jinevening:
> Could you explain a bit?

Ah, the comment is wrong. The direction of the pass was changed in #7215, but the comment was not updated. I've updated the comment too. Thanks.

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

mhs4670go (approved):
LGTM

PR #8498, opened 2022-02-23
[common-artifacts] Remove prepare of venv of TF2.3.0

seanshpark:
This will remove prepare of python venv of TF 2.3.0 as it's not used anymore.


mhs4670go (approved):
LGTM

jinevening (approved):
LGTM

Issue #8497, opened 2022-02-23
[quantizer] Introduce a pass for backward propagation of qparam


jinevening:
## What

Let's introduce a pass for backward propagation of qparam

## Why

Currently, backward propagation of qparam is done in different places.

1. propagate_concat_ quantparam
2. QuantizeConstInputActivation

Let's make a new pass for backward propagation of qparam
1. for well-orgnaized code
2. to support consecutive propagation

TODO
- [ ] Introduce PropagateQParamBackwardPass
- [ ] Move (and modify) existing backward propagation codes to PropagateQParamBackwardPass

jinevening:
Currenlty, backward propagation of qparam is done for the following Ops.

Concat : ofm -> values (ifms)
Pack : ofm -> values (ifms)
OneHot : ofm -> on_value, off_value
PadV2 : ofm -> ifm, constant_values (PadV2 has a special case where ifm qparam is propagated to ofm and constant_values)

PR #8502, opened 2022-02-23
[luci] Introduce PropagateQParamBackward Pass

jinevening:
This introduces a pass to propagate qparam backward.

---
Related to: https://github.com/Samsung/ONE/issues/8497
Draft PR: #8500

### DIFF
--- a/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
+++ b/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/PropagateQParamBackwardPass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/CircleNodeVisitor.h>
+#include <luci/Log.h>
+
+namespace
+{
+
+// Visitor to propagate quantization parameters backwards
+struct PropagateQParamBackward final : public luci::CircleNodeMutableVisitor<void>
+{
+  PropagateQParamBackward(loco::DataType output) : _output_type(output) {}
+
+private:
+  loco::DataType _output_type;
+
+  void visit(luci::CircleNode *) {}
+};
+
+} // namespace
+
+namespace luci
+{
+
+bool PropagateQParamBackwardPass::run(loco::Graph *g)
+{
+  LOGGER(l);
+
+  // We use post-order traversal as qparam is propagated backward
+  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
+  {
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+    INFO(l) << "PropagateQParamBackwardPass visit node: " << circle_node->name() << std::endl;
+
+    PropagateQParamBackward pqb(_output_model_dtype);
+    circle_node->accept(&pqb);
+  }
+
+  return false;
+}
+
+} // namespace luci
--- a/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
+++ b/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
@@ -43,8 +43,10 @@ bool PropagateQParamBackwardPass::run(loco::Graph *g)
 {
   LOGGER(l);
 
-  // We use post-order traversal as qparam is propagated backward
-  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
+  // We use reverse post-order traversal as qparam is propagated backward
+  auto nodes = loco::postorder_traversal(loco::output_nodes(g));
+  std::reverse(nodes.begin(), nodes.end());
+  for (auto node : nodes)
   {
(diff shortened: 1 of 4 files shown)


seanshpark (dismissed):
LGTM

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
```diff
@@ -0,0 +1,59 @@
...
+  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
+  {
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+    INFO(l) << "PropagateQParamBackwardPass visit node: " << circle_node->name() << std::endl;
+
+    PropagateQParamBackward pqb(_output_model_dtype);
+    circle_node->accept(&pqb);
+  }
+
+  return false;
```
llFreetimell, line 56: After seeing full draft, this pass seems to return always `false`.
I think it is maybe because visitor class return `void` and thus there is nothing to check, am I right?

Then how about changing return type of visitor class and functions to `bool` and then using it to determine whether there are something changed or not?
If this is hard to apply, we may go as it is :D

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
```diff
@@ -0,0 +1,59 @@
...
+} // namespace
+
+namespace luci
+{
+
+bool PropagateQParamBackwardPass::run(loco::Graph *g)
+{
+  LOGGER(l);
+
+  // We use post-order traversal as qparam is propagated backward
```
llFreetimell, line 46: If `qparam` is propagated backward, I think `qparam` is propagated with `op3 --> op2 --> op1` sequence, am I right?
```
[op1] --- [op2] --- [op3] -- output
```

If so, I think pre-order traversal is correct... isn't it?

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
jinevening: You're right :) The return value of this pass is not used anywhere, so the pass just returns false.

In the following PRs, I'll just move existing codes in QuantizeWithMinMaxPass to the PropagateQParamBackwardPass. For example, `propagate_concat_quantparam`, which returns void. We can modify the existing codes, but it does not give actual benefit because the return value is not used.

Another option would be implementing `PropagateQParamBackward` without inheriting `logo::Pass`, which looks better to me. I used `logo::Pass` for consistency with `PropagateQParamForwardPass`, but I'm very open to change the decision.

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
jinevening: Ah, you're right. `loco::postorder_traversal` is in fact post-order traversal of args, which is the opposite of data flow direction. Thanks.

jinevening: > If so, I think pre-order traversal is correct... isn't it?

Hmm.. actually, reverse post-order traversal (topological order) would be right. Pre-ordering makes overwrite of qparam when there is a diamond-shaped dataflow. (See https://en.wikipedia.org/wiki/Depth-first_search#Vertex_orderings for more details)

llFreetimell: I think both pre-order and reverse(post-order) is fine when diamond-shaped dataflow :)
But, for diamond-shaped flow, both of them would be problem..!

```
[op1] -+-- [op2] --+- [op3] --- output
       +-- [op4] --+
```

For above example graph, `reverse(loco::postorder_traversal)` would return `reverse(op1, op2, op4, op3) = (op3, op4, op2, op1)`.
Then `op2` overwrites qparam which is propagated from `op4` :(

In this case, what is intended behavior?

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
llFreetimell: > We can modify the existing codes, but it does not give actual benefit because the return value is not used.

I agree with it :)

> implementing PropagateQParamBackward without inheriting logo::Pass

It looks better to me either :D
But for now, we may go as it is and then discuss later how to refactoring this case!
If another similar passes are created, it would be more helpful to think which solution would be better :)
Could you leave comments like following example?

```
  // This pass is only run once, so return false
  // TODO Refactoring not to return meaningless value
```

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
jinevening: > In this case, what is intended behavior?

It depends on whether `op1` is const or not
1. If `op1` is const : `op1` is cloned and quantized for each Op (`op2` and `op4`)
2. If `op1` is not const : qparam of `op2` and `op4` is not propagated to `op1`, because `op1` has multiple successors.

> both pre-order and reverse(post-order) is fine when diamond-shaped dataflow :)

You're right. I was worried about the mismatch of qparam in the overwritten op (`op1`'s ofm and its ifms), but the source op (`op4`) also has the same problem.

Both traversals (pre-order and reverse post-order) would be ok, because qparam will not be propagated if `op1` is not const. In this case, I'll go with reverse post-order for ease of implementation.

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
llFreetimell: Not for now but maybe good solution :D
We can introduce new phase such as `logo::PhaseStrategy::OnlyOnce` for these cases :)

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
llFreetimell: > I was worried about the mismatch of qparam in the overwritten op (op1's ofm and its ifms), but the source op (op4) would also have the same problem.

Then it would be better to add some validation/assertion code to check it later :)

> I'll go with reverse post-order for ease of implementation.

I agree with this because pre-order traversal is not implemented in `loco`.
So using `postorder_traversal` seems much safer and better :D

llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

PR #8506, opened 2022-02-23
[luci] Note for tests CMake target

seanshpark:
This will add a note for tests CMake target as a reminder for cross build.


seanshpark:
For #8379, from draft #8407

hyunsik-yoon (approved):

PR #8509, opened 2022-02-24
[luci] Move propagation of concat qparam code

jinevening:
This moves propagate_concat_quantparam to PropagateQParamBackwardPass.

---
Related to: https://github.com/Samsung/ONE/issues/8497
Draft PR: https://github.com/Samsung/ONE/pull/8500

### DIFF
--- a/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
+++ b/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
@@ -15,10 +15,200 @@
  */
 
 #include "luci/Pass/PropagateQParamBackwardPass.h"
+#include "QuantizationUtils.h"
 
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/CircleNodeVisitor.h>
+#include <luci/Service/Nodes/CircleConst.h>
 #include <luci/Log.h>
+
+#include <cmath>
+
+namespace
+{
+
+void quant_const_values(luci::CircleConst *const_node, float scaling_factor, float zerop,
+                        loco::DataType quant_type)
+{
+  uint32_t size = const_node->size<loco::DataType::FLOAT32>();
+
+  const float scaling_factor_inv = 1.0 / scaling_factor;
+  std::vector<int32_t> quantized_values(size);
+  for (uint32_t i = 0; i < size; ++i)
+  {
+    auto data = static_cast<double>(const_node->at<loco::DataType::FLOAT32>(i));
+    double quantized_float = std::round(data * scaling_factor_inv) + zerop;
+    constexpr auto int_max = static_cast<double>(std::numeric_limits<int32_t>::max());
+    constexpr auto int_min = static_cast<double>(std::numeric_limits<int32_t>::min());
+    quantized_float = std::min(int_max, std::max(int_min, quantized_float));
+
+    quantized_values[i] = static_cast<int32_t>(quantized_float);
+  }
+
+  switch (quant_type)
+  {
+    case loco::DataType::U8:
+      const_node->dtype(loco::DataType::U8);      // change the type of tensor
+      const_node->size<loco::DataType::U8>(size); // resize tensor
+      for (uint32_t i = 0; i < size; ++i)
+        const_node->at<loco::DataType::U8>(i) = std::min(255, std::max(0, quantized_values[i]));
+      break;
+    case loco::DataType::S16:
+      assert(zerop == 0);
+      const_node->dtype(loco::DataType::S16);      // change the type of tensor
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
```diff
@@ -15,11 +15,201 @@
  */
 
 #include "luci/Pass/PropagateQParamBackwardPass.h"
+#include "QuantizationUtils.h"
```
seanshpark, line 18: I think this be above "luci/Pass/PropagateQParamBackwardPass.h" as this header is in same folder, more closer to this file.

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
```diff
@@ -15,11 +15,201 @@
...
+
+namespace
+{
+
+void quant_const_values(luci::CircleConst *const_node, float scaling_factor, float zerop,
+                        loco::DataType quant_type)
+{
+  uint32_t size = const_node->size<loco::DataType::FLOAT32>();
+
+  const float scaling_factor_inv = 1.0 / scaling_factor;
```
seanshpark, line 35: (optional) I PR looks relocation so maybe not related with this PR but is `scaling_factor` checked with zero value?

jinevening: Hmm.. it is not checked anywhere. We may have to throw an exception in that case.

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
jinevening: I put `PropagateQParamBackwardPass.h` first because this file is for the implementation of `PropagateQParamBackwardPass`.

I've made cpp files in this way for many passes.

QuantizeDequantizeWeightsPass
QuantizeWithMinMaxPass
FusePreactivationBatchNormPass
MakeBatchNormGammaPositivePass
and so on

If we're going to make rules for including headers, I think it would be better to do it in a separate issue.

seanshpark: I don't think there can be a strict rule.
First I thought "QuantizationUtils.h" is local header so would be better to include first
but this source is implementation of PropagateQParamBackwardPass, so "PropagateQParamBackwardPass.h" should be first to include.

seanshpark (approved):
LGTM

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
```diff
@@ -15,11 +15,201 @@
...
+  std::vector<int32_t> quantized_values(size);
+  for (uint32_t i = 0; i < size; ++i)
+  {
+    auto data = static_cast<double>(const_node->at<loco::DataType::FLOAT32>(i));
+    double quantized_float = std::round(data * scaling_factor_inv) + zerop;
```
llFreetimell, line 40: Unlike other codes, this variables are declared as `double`, not `float`.
Furthermore, type of `quantized_float` is `double` but variable name includes `float`.
Calculation may not have problem but it seems inconsistent.
How about changing variable name or variable type? :)

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
```diff
@@ -15,11 +15,201 @@
...
+    constexpr auto int_max = static_cast<double>(std::numeric_limits<int32_t>::max());
+    constexpr auto int_min = static_cast<double>(std::numeric_limits<int32_t>::min());
```
llFreetimell, line 42: Not mandatory, but how about applying following change? :D

```suggestion
    constexpr auto int_max_as_double = static_cast<double>(std::numeric_limits<int32_t>::max());
    constexpr auto int_min_as_double = static_cast<double>(std::numeric_limits<int32_t>::min());
```

Type of `int_max` and `int_min` seems `int` but they are actually `double` :)

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
jinevening: This code was copied from `QuantizeWithMinMaxPass`.

It'll be better to update the variable name in the separate PR.

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
jinevening: Same with https://github.com/Samsung/ONE/pull/8509#discussion_r813497064. It'll be good to make a PR for renaming.

llFreetimell (approved):
LGTM

PR #8510, opened 2022-02-24
[common-artifacts] Use variable for target files

seanshpark:
This will revise to use variable for target files.

### DIFF
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -126,6 +126,12 @@ macro(tcgenerate_option NAME OPTION ARG1 ARG2 ARG3)
   set(TCGEN_OPT_${NAME} ${OPTION} ${ARG1} ${ARG2} ${ARG3})
 endmacro()
 
+set(CIRCLECHEF_FILE_PATH $<TARGET_FILE:circlechef-file>)
+set(TFLCHEF_FILE_PATH $<TARGET_FILE:tflchef-file>)
+set(TFLITE2CIRCLE_PATH $<TARGET_FILE:tflite2circle>)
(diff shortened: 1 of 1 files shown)


seanshpark:
For #8379, from draft #8407

mhs4670go (approved):
LGTM. But IMHO, actually, current codes are readable and short enough to me.

seanshpark:
This is to prepare override target executables for cross build.
For cross build, we need to run host executable.
In next PR, the variable will changee for cross build.

mhs4670go:
> This is to prepare override target executables for cross build.
For cross build, we need to run host executable.
In next PR, the variable will change for cross build.

I thought cross-build means our preset for package is going to be built at cross platform. And our preset doesn't have `common-artifacts` for build list. Is there any reason for building `common-artifacts`?

seanshpark:
> Is there any reason for building common-artifacts?

To validate value testing of `luci` + `luci-interpreter` in ARM32 target is the main reason.
For this, I used `luci-value-test` and this uses files from `common-artifacts`

But as I could't find ARM32 TensorFlow and lite, I could not use the tflite interpreter as generating reference output data.
So I tought of (1) prepare reference input/output data with host build (2) use those files as reference data to compare the output of ARM32 luci-interpreter.

Current draft has modified versions of test scripts that runs in ARM32 target device.

seanshpark:
Hum... actually common-artifacts generated files doesn't seem to be used cross build testing...

I think it was because current dependency link from "requires.cmake" fixes the dependency.

seanshpark:
I'll check if we can skip whole `common-artifacts` for cross build.
That is #8511 can be revised.

seanshpark:
This change is not needed. We can skip this module for ARM32 cross build.

PR #8511, opened 2022-02-24
[common-artifacts] Skip prepare venv for cross build

seanshpark:
This will revise to skip preparation of venv for cross build.

### DIFF
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -1,3 +1,6 @@
+if(NOT CMAKE_CROSSCOMPILING)
+# TODO fix indentation
+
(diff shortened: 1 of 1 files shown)


seanshpark:
For #8379, from draft #8407

mhs4670go (approved):
LGTM
(optional) It would be better to comment the reason why it has to be skipped.

seanshpark:
> It would be better to comment the reason why it has to be skipped.

Ah, main reason is that 
(1) I'm not sure preparing host venv for target build is needed
(2) TensorFlow for ARM32 doesn't seem to exist.

seanshpark:
This change is not needed. We can skip this module for ARM32 cross build.

PR #8517, opened 2022-02-24
[luci] Apply activation quantization to visitor target

jinevening:
This applies activation quantization to visitor target, not args.


jinevening:
For reviewers:

Previously, activation quantization was applied to arg of visitor target (node).

This PR applies it directly to the visitor target (node). The change will not cause any difference in the final quantized values.

NOTE: Hide whitespace option may help to review.

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

PR #8515, opened 2022-02-24
[luci] Move backward propagation of qparam codes

jinevening:
This moves backward propagation of qparam codes for PadV2, Pack, OneHot.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/8497
Draft PR: https://github.com/Samsung/ONE/pull/8500

PR #7476, opened 2021-08-12
[circle-quantizer] Infer Pad quantization from input in special case

binarman:
If Pad is a product of MaxPoolWithArgmax operator
do not fit constant value into quantization range.
Quantize whole pad using input quantization parameters.

related issue: #7229

ONE-DCO-1.0-Signed-off-by: Alexander Efimov <a.efimov@samsung.com>

PR #8522, opened 2022-02-25
[luci] Tidy PropagateQParamBackwardPass.cpp file

jinevening:
This PR tidies PropagateQParamBackwardPass.cpp file.

---
Related to comments in https://github.com/Samsung/ONE/pull/8515 and https://github.com/Samsung/ONE/pull/8509

### DIFF
--- a/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
+++ b/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
@@ -37,12 +37,12 @@ void quant_const_values(luci::CircleConst *const_node, float scaling_factor, flo
   for (uint32_t i = 0; i < size; ++i)
   {
     auto data = static_cast<double>(const_node->at<loco::DataType::FLOAT32>(i));
-    double quantized_float = std::round(data * scaling_factor_inv) + zerop;
-    constexpr auto int_max = static_cast<double>(std::numeric_limits<int32_t>::max());
-    constexpr auto int_min = static_cast<double>(std::numeric_limits<int32_t>::min());
-    quantized_float = std::min(int_max, std::max(int_min, quantized_float));
+    double quantized_data = std::round(data * scaling_factor_inv) + zerop;
+    constexpr double int_max = static_cast<double>(std::numeric_limits<int32_t>::max());
+    constexpr double int_min = static_cast<double>(std::numeric_limits<int32_t>::min());
+    quantized_data = std::min(int_max, std::max(int_min, quantized_data));
 
-    quantized_values[i] = static_cast<int32_t>(quantized_float);
+    quantized_values[i] = static_cast<int32_t>(quantized_data);
   }
 
   switch (quant_type)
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
```diff
@@ -37,12 +37,12 @@ void quant_const_values(luci::CircleConst *const_node, float scaling_factor, flo
   for (uint32_t i = 0; i < size; ++i)
   {
     auto data = static_cast<double>(const_node->at<loco::DataType::FLOAT32>(i));
-    double quantized_float = std::round(data * scaling_factor_inv) + zerop;
-    constexpr auto int_max = static_cast<double>(std::numeric_limits<int32_t>::max());
-    constexpr auto int_min = static_cast<double>(std::numeric_limits<int32_t>::min());
-    quantized_float = std::min(int_max, std::max(int_min, quantized_float));
+    double quantized_data = std::round(data * scaling_factor_inv) + zerop;
```
llFreetimell, line 40: Making variable name consistent looks good :)
But I have one question!

`scaling_factor` and `sacling_factor_inv` are `float`.
However, `data` is `double` even `const_node` has `FLOAT32` type.
Is there any special reason to use different type for both `float`?

---
FYI, calculation itself has no problem because cpp reference is saying that ["for the binary operators (except shifts), if the promoted operands have different types, additional set of implicit conversions is applied, known as usual arithmetic conversions with the goal to produce the common type"](https://en.cppreference.com/w/cpp/language/operator_arithmetic)


i.e. `std::round(data * scaling_factor_inv) + zerop;` is calculated like following steps :D
```cpp
double quantized_data = std::round(data * scaling_factor_inv) + zerop;

double tmp1 = data * (double)scaling_factor_inv;
double tmp2 = std::round(tmp1);
double tmp3 = tmp2 + (double)zerop;
quantized_data = tmp3;
```

jinevening: > Is there any special reason to use different type for both float?

AFAIK, @binarman implemented this to handle a value smaller than the lowest value of float. (https://github.com/Samsung/ONE/pull/7476)

llFreetimell: > AFAIK, @binarman implemented this to handle a value smaller than the lowest value of float. (#7476)

Ah, if so this question may not be directly related with this PR :)
Thanks!

llFreetimell: @binarman 

Could you give me the answer for above question regardless of whether this PR is merged or not? :D

llFreetimell (approved):
LGTM

compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
binarman: Sorry for the delayed answer.

Idea behind this choice was precision: float has only 24 bytes to store mantissa, but we want to quantize to 32 bit int here. As @jinevening mentioned, this means that least significant bits will be lost if `data*scale` is large.

Double will work better in some cases, like:
```
data = 1.0
scale = 2^30
zero point = 5

double: 1073741829 
float:  1073741824
```

P.s. I am not sure if `double` will make a difference with current quantizations, I was not able to find proper example, that will propagate error through all `min(max())` clipping.

Issue #8525, opened 2022-02-25
[luci-interpreter] Support Dequantize op


jinevening:
## What

Let's support Dequantize operator

## Why

Dequantize op is necessary for dynamic fake quantization.

BalyshevArtem:
Can I work on this issue?

jinevening:
Of course! :)

jinevening:
After #8527 is merged, it'll be possible to remove models with Dequantize Op from `exclude.lst` in common-artifacts.

PR #8527, opened 2022-02-27
[luci-interpreter] Support Dequantize op in luci-interpreter

BalyshevArtem:
This commit adds support for Dequantize op in luci-interpreter.

for issue: #8525 


compiler/luci-interpreter/src/kernels/Dequantize.cpp
```diff
@@ -0,0 +1,77 @@
...
+Dequantize::Dequantize(const Tensor *input, Tensor *output) : Kernel({input}, {output}) {}
+
+void Dequantize::configure()
+{
+  LUCI_INTERPRETER_CHECK(input()->element_type() == loco::DataType::S8 ||
+                         input()->element_type() == loco::DataType::U8 ||
+                         input()->element_type() == loco::DataType::S16);
+
+  if (input()->element_type() == loco::DataType::S16)
+    LUCI_INTERPRETER_CHECK(input()->zero_point() == 0);
```
binarman, line 35: Could you also add check that input tensor is quantized "per-tensor", not per-channel.
This is unlikely scenario, but just in case.

binarman, line 35: Could you also add check that input tensor is quantized "per-tensor", not per-channel.
This is unlikely scenario, but just in case.

like this: 
```suggestion
  LUCI_INTERPRETER_CHECK(filter()->scales().size() == 1);

  if (input()->element_type() == loco::DataType::S16)
    LUCI_INTERPRETER_CHECK(input()->zero_point() == 0);
```

BalyshevArtem: Done :)

binarman (approved):
LGTM

Issue #8480, opened 2022-02-22
[Compiler FE] Build armv7l target on tizen gbs


chunseoklee:
parent : #8379 (draft: #8407)

Let try to build and run Compiler FE on tizen obs build system !

chunseoklee:
Error on https://github.com/Samsung/ONE/pull/8481/commits/bd2bb66c3fcb81f09ffd3aa1a518bfb667bc7a79

> /home/abuild/rpmbuild/BUILD/nnfw-1.20.0/compiler/mio-tflite260/CMakeLists.txt:13 (nnas_find_package)

mio needs tensorflow-2.6. Thus tf-2.6 source should be included into packaging folder. 

<details>
```
[  120s] -- !!! FlatBuffers_FOUND
[  120s] -- Download TENSORFLOW from https://github.com/tensorflow/tensorflow/archive/v2.6.0.tar.gz
[  120s] -- (Trial Count : 0)
[  120s] CMake Warning at /home/abuild/rpmbuild/BUILD/nnfw-1.20.0/infra/cmake/modules/ExternalSourceTools.cmake:66 (message):
[  120s]   error: downloading
[  120s]   'https://github.com/tensorflow/tensorflow/archive/v2.6.0.tar.gz' failed
[  120s]
[  120s]                 status_code: 6
[  120s]                 status_string: "Couldn't resolve host name"
[  120s]                 log: getaddrinfo(3) failed for github.com:443
[  120s]
[  120s]   Could not resolve host: github.com
[  120s]
[  120s]   Closing connection 0
[  120s]
[  120s] Call Stack (most recent call first):
[  120s]   /home/abuild/rpmbuild/BUILD/nnfw-1.20.0/infra/cmake/packages/TensorFlowSource-2.6.0/TensorFlowSourceConfig.cmake:12 (ExternalSource_Download)
[  120s]   /home/abuild/rpmbuild/BUILD/nnfw-1.20.0/infra/cmake/packages/TensorFlowSource-2.6.0/TensorFlowSourceConfig.cmake:18 (_TensorFlowSource_import)
[  120s]   CMakeLists.txt:43 (find_package)
[  120s]   /home/abuild/rpmbuild/BUILD/nnfw-1.20.0/compiler/mio-tflite260/CMakeLists.txt:13 (nnas_find_package)
[  120s]
[  120s]
```
</details>

chunseoklee:
[...]

PR #8532, opened 2022-03-02
[luci/pass] Fix pass test with logging

seanshpark:
This will fix pass test with logging when LUCI_LOG is given.


seanshpark:
For https://github.com/Samsung/ONE/issues/8480#issuecomment-1056027106

jinevening (approved):
LGTM Thank you

I think we may need a check to confirm that there is no node with `FusedActFunc::UNDEFINED`. Maybe in `luci::validate`.

seanshpark:
> I think we may need a check to confirm that there is no node with FusedActFunc::UNDEFINED. Maybe in luci::validate.

Actually I wanted to permit `UNDEFINED` for unit testing. But `Pass` happens to dump graph logs...

PR #8535, opened 2022-03-02
[luci/value-test] Evaluate scripts with reference

seanshpark:
This will introduce evaluation scripts with data from reference folder.
- these scripts are used to test for cross build binaries


hyunsik-yoon (approved):

mhs4670go:
@seanshpark They doesn't need to be installed?

mhs4670go (approved):
LGTM

PR #8545, opened 2022-03-02
[luci] Remove duplicate codes in QuantizeBias

jinevening:
This removes duplicate codes in QuantizeBias.

---
Related to: https://github.com/Samsung/ONE/pull/8540#discussion_r817402140

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -701,7 +701,7 @@ private:
       bias = dynamic_cast<luci::CircleConst *>(b);
     }
 
-    // Return true if bias is quantizable
+    // Return true if bias can be quantized with valid input an weights
     operator bool()
     {
       if (bias == nullptr || is_quantized(bias))
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -692,30 +692,25 @@ private:
   {
     CircleNode *input = nullptr;
     CircleNode *weights = nullptr;
-    CircleNode *bias = nullptr;
+    CircleConst *bias = nullptr;
 
-    IWB() {}
+    IWB(loco::Node *i, loco::Node *w, loco::Node *b)
+    {
+      input = dynamic_cast<luci::CircleNode *>(i);
+      weights = dynamic_cast<luci::CircleNode *>(w);
+      bias = dynamic_cast<luci::CircleConst *>(b);
+    }
 
-    IWB(CircleNode *i, CircleNode *w, CircleNode *b) : input(i), weights(w), bias(b) {}
-
-    operator bool() { return input && weights && bias; }
+    // Return true if bias is quantizable
+    operator bool()
+    {
+      if (bias == nullptr || is_quantized(bias))
+        return false;
+      if (input == nullptr || weights == nullptr)
+        return false;
+      return true;
+    }
   };
-
-private:
-  IWB check_iwb(loco::Node *input, loco::Node *weight, loco::Node *bias)
-  {
-    auto b = dynamic_cast<luci::CircleConst *>(bias);
-    if (not b)
-      return IWB();
-
-    if (is_quantized(b))
-      return IWB();
-
-    auto i = loco::must_cast<luci::CircleNode *>(input);
-    auto w = loco::must_cast<luci::CircleNode *>(weight);
-
-    return IWB(i, w, b);
-  }
 
 private:
   // Return a quantized bias node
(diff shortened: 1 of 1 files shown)


jinevening:
@llFreetimell  #8540 was merged before I apply this change :)

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -687,6 +687,36 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<void>
...
+
+    if (is_quantized(b))
+      return IWB();
+
+    auto i = loco::must_cast<luci::CircleNode *>(input);
+    auto w = loco::must_cast<luci::CircleNode *>(weight);
+
+    return IWB(i, w, b);
+  }
+
```
llFreetimell, line 719:
```suggestion
```

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -687,6 +687,36 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<void>
...
+  {
+    CircleNode *input = nullptr;
+    CircleNode *weights = nullptr;
+    CircleNode *bias = nullptr;
+
+    IWB() {}
+
+    IWB(CircleNode *i, CircleNode *w, CircleNode *b) : input(i), weights(w), bias(b) {}
+
+    operator bool() { return input && weights && bias; }
```
llFreetimell, line 701:
```suggestion
    IWB(loco::Node *i, loco::Node *w, loco::Node *b)
    {
      input = loco::must_cast<luci::CircleNode *>(i);
      weights = loco::must_cast<luci::CircleNode *>(w);
      bias = dynamic_cast<luci::CircleConst *>(input);
    }

    // Return if bias is quantizable or not
    operator bool()
    {
      if (bias == nullptr || is_quantized(bias))
        return false;
      if (input == nullptr || weight == nullptr)
        return false;
    }
```

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -771,71 +801,47 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<void>
...
-      return;
-
-    if (is_quantized(bias))
-      return;
-
-    auto i = loco::must_cast<luci::CircleNode *>(node->input());
-    auto w = loco::must_cast<luci::CircleNode *>(node->filter());
-    auto new_bias = quantized_bias(i, w, bias);
-    node->bias(new_bias);
+    if (auto iwb = check_iwb(node->input(), node->filter(), node->bias()))
```
llFreetimell, line 804:
```suggestion
    if (auto iwb = IWB(node->input(), node->filter(), node->bias()))
```

llFreetimell (commented):
Thanks for applying my [overall idea](https://github.com/Samsung/ONE/pull/8540/files#r817401647) :D
Just one more thing! if we modify member functions of `struct IWB` like my comments, the code would be more readable I think :)

But it's totally my thoughts!
So if you don't like this idea, you can ignore and go as it is :D

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -687,6 +687,31 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<void>
...
+  struct IWB
+  {
+    CircleNode *input = nullptr;
+    CircleNode *weights = nullptr;
+    CircleConst *bias = nullptr;
+
+    IWB(loco::Node *i, loco::Node *w, loco::Node *b)
+    {
+      input = dynamic_cast<luci::CircleNode *>(i);
+      weights = dynamic_cast<luci::CircleNode *>(w);
```
jinevening, line 700: AFAIK, we have a convention that does not throw an exception in constructor. So I used dynamic_cast instead of loco::must.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -687,6 +687,31 @@ struct QuantizeBias final : public luci::CircleNodeMutableVisitor<void>
...
+      bias = dynamic_cast<luci::CircleConst *>(b);
+    }
+
+    // Return true if bias is quantizable
```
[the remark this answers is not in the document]
llFreetimell: Then how about `Return true if node with input, weights, bias can be quantized` ?
Implementation is using `input` and `weights` to determine this :)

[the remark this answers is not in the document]
seanshpark: > Then how about Return true if node with input, weights, bias can be quantized ?

What I understand is only bias is quantized, if I understand correctly...
Um... maybe `Return true if bias can be quantized with valid input and weights` ?

seanshpark (approved):
LGTM thank you!

llFreetimell (approved):
LGTM

PR #8546, opened 2022-03-02
[luci/value-test] Add executable flags to scripts

seanshpark:
This will revise scripts to have executable flag for test execution.


mhs4670go (approved):
LGTM

PR #8550, opened 2022-03-02
[luci/value-test] Revise test in target

seanshpark:
This will revise test as for host and for target device.


seanshpark:
For #8379, from draft #8407

hyunsik-yoon (approved):

mhs4670go (approved):
LGTM

Issue #8524, opened 2022-02-25
[quantizer] Refactoring QuantizeWithMinMaxPass


jinevening:
## What

Let's extract the below codes in QuantizeWithMinMaxPass to separate files
- QuantizeActivation
- QuantizeConstInputActivation
- QuantizeSpecialActivation
- QuantizeWeights
- QuantizeBias

## Why

Those codes can be used in another pass.

jinevening:
All done.

PR #8553, opened 2022-03-03
[luci] Extract visitors to quantize activation

jinevening:
This extracts visitors to quantize activation from QuantizeWithMinMax.cpp.
- QuantizeActivation
- QuantizeConstInputActivation
- QuantizeSpecialActivation

---
Related to: https://github.com/Samsung/ONE/issues/8524


jinevening:
I force-pushed just to update the commit message more properly.

compiler/luci/pass/src/QuantizeActivation.cpp
```diff
@@ -0,0 +1,408 @@
...
+{
+  assert(node->dtype() == loco::DataType::FLOAT32);
+
+  float min = std::numeric_limits<float>::max();
+  float max = std::numeric_limits<float>::lowest();
+  for (uint32_t i = 0; i < node->size<loco::DataType::FLOAT32>(); i++)
+  {
+    auto data = node->at<loco::DataType::FLOAT32>(i);
+    min = data < min ? data : min;
+    max = data > max ? data : max;
```
seanshpark, line 40: (optional) maybe use `std::min()`, `std::max()` ?

seanshpark (approved):
LGTM

compiler/luci/pass/src/QuantizeActivation.cpp
seanshpark: ah, maybe need to include `<algorithm>` too :)

seanshpark (approved):
LGTM thank you

llFreetimell (approved):
LGTM

Issue #8512, opened 2022-02-24
[quantizer] How to express mixed-precision model?


jinevening:
## What

Let's discuss how to express mixed-precision model in the circle model.

## Considerations
1. We should be able to express the below cases.
- Conv (u8) -> Requantize (u8->s16) -> Conv (s16)
- ~Hybrid Conv (ifm: u8, weights: u8, bias: s32, **ofm: s16**) -> Conv (s16)~ This will be considered later

2. Qparam can be propagated through Quantize Op (forward, backward).

## How?

Mixed-precision model has `default_quantization_dtype`, and layers can have dtype different from the default type.

We can express an Op with a dtype different from `default_quantization_dtype` by inserting Quantize Op before ifm and after ofm of the Op.

![image](https://user-images.githubusercontent.com/5449554/155442547-3c78d424-b894-4695-9416-df950a17f67f.png)
Left: Example model
Middle: Default type is uint8 and Conv is int16.
Right: If the first Op and Conv are int16, Quantize Ops between them will be removed.

In this approach, we'll have to
- quantize non-default type Ops
- insert Quantize Ops before/after activations (non-const)
- remove redundant Quantize Ops (ex: Quantize(u8->s16) -> Quantize(s16->u8)). #8526
- propagate qparam through Quantize Ops properly
- handle constant activations
- update verifier

For Hybrid conv, we can express it in the circle model, or we can give the role to decide whether to use Hybrid Conv or not to the backend compiler (or runtime).

Related to : mixed-precision quantization (#8398)

CC @Samsung/one_quantizer

jinevening:
Now quantize Ops are inserted for mixed-precision model.
(Quant_Conv_Mul_Add_000.q.circle generated in `circle-quantizer-dredd-recipe-test`)

Default quantization dtype: U8
Conv2D dtype: S16

![image](https://user-images.githubusercontent.com/5449554/157780225-612b9615-7859-42e8-b3a8-508f32d013a3.png)

PR #8555, opened 2022-03-03
[luci] Use layer info during quantization

jinevening:
This uses layer info during quantization.

---

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -900,7 +900,7 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
 
 LayerInfoMap QuantizeWithMinMaxPass::create_layer_info_map(loco::Graph *g) const
 {
-  auto info_by_name = LayerInfoMap();
+  LayerInfoMap info_by_name;
 
   for (auto &&info : _ctx->layers_info)
   {
@@ -911,6 +911,12 @@ LayerInfoMap QuantizeWithMinMaxPass::create_layer_info_map(loco::Graph *g) const
       auto cnode = loco::must_cast<luci::CircleNode *>(node);
       if (cnode->name() == name)
       {
+        if (info_by_name.find(name) != info_by_name.end())
+        {
+          throw std::runtime_error("Duplicate layer name " + name +
+                                   ". Check layer names in the quantization configuration file.");
+        }
+
         info_by_name[name] = &info;
         found = true;
         break;
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -33,6 +33,8 @@
 #include <iostream>
 #include <cmath>
 
+using LayerInfoMap = std::unordered_map<std::string, luci::LayerInfo *>;
+
 namespace
 {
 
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -896,6 +898,35 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
...
+    }
+
+    if (not found)
+      throw std::runtime_error("No such layer named " + name +
+                               ". Check layer names in the quantization configuration file.");
+  }
+
+  assert(info_by_name.size() == _ctx->layers_info.size()); // FIX_ME_UNLESS
+
+  return info_by_name;
```
seanshpark, line 927: I'm not sure how big size memory would be for return but how about
```cxx
bool QuantizeWithMinMaxPass::create_layer_info_map(loco::Graph *g, LayerInfoMap &info_by_name) const
```
return false if not found or true if OK.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -896,6 +898,35 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
...
+  for (auto &&info : _ctx->layers_info)
+  {
+    auto name = info.name;
+    bool found = false;
+    for (auto node : loco::active_nodes(loco::output_nodes(g)))
+    {
+      auto cnode = loco::must_cast<luci::CircleNode *>(node);
+      if (cnode->name() == name)
+      {
+        info_by_name[name] = &info;
```
seanshpark, line 914: circle graph assumes `name` should be unique. (there is validation but duplicate can splip in)
Q) is there any assumptions in `create_layer_info_map` about the name as this returns `map` ?

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: AFAIK, std::map (and other std containers) will not be copied due to copy elision.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: Hmm.. There is no assumption in `_ctx->layers_info`. I should add a logic to handle duplicate names. Thanks!

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -896,6 +898,35 @@ void QuantizeWithMinMaxPass::set_output_type(loco::Graph *g) const
...
+        break;
+      }
+    }
+
+    if (not found)
+      throw std::runtime_error("No such layer named " + name +
+                               ". Check layer names in the quantization configuration file.");
+  }
+
+  assert(info_by_name.size() == _ctx->layers_info.size()); // FIX_ME_UNLESS
```
llFreetimell, line 925: Current logic itself has no problem :)
But if we apply following codes, we may be able to expect :D
- Better readability
- Better performance : O(n^2) --> O(n)

```suggestion
  for (auto &&info : _ctx->layers_info)
    info_by_name[info.name] = &info;

  for (auto node : loco::active_nodes(loco::output_nodes(g)))
  {
    auto cnode = loco::must_cast<luci::CircleNode *>(node);
    if (info_by_name.find(cnode->name()) == info_by_name.end())
    {
      throw std::runtime_error("No such layer named " + cnode->name() +
                               ". Check layer names in the quantization configuration file.");
    }
  }
```

jinevening: Looks much better!

llFreetimell (approved):
LGTM

seanshpark (dismissed):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: Hmm.. this logic causes an error. It throws an exception whenever a graph has a node not in `info_by_name`, so I roll backed the code.

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

PR #8558, opened 2022-03-03
[luci] Pass layer info to QuantizedModelVerifier

jinevening:
This passes layer info to QuantizedModelVerifier.

---

### DIFF
--- a/compiler/luci/pass/src/QuantizedModelVerifier.cpp
+++ b/compiler/luci/pass/src/QuantizedModelVerifier.cpp
@@ -22,13 +22,74 @@
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/CircleNodeVisitor.h>
 
+using LayerInfoMap = std::unordered_map<std::string, luci::LayerInfo *>;
+
 namespace luci
 {
+
+LayerInfoMap QuantizedModelVerifier::create_layer_info_map(loco::Graph *g) const
+{
+  LayerInfoMap info_by_name;
+
+  for (auto &&info : _ctx->layers_info)
+  {
+    auto name = info.name;
+    bool found = false;
+    for (auto node : loco::active_nodes(loco::output_nodes(g)))
+    {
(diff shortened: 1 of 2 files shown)


seanshpark (approved):
LGTM

seanshpark:
Q) how are the new codes tested? code coverage may drop if not tested in CI...

compiler/luci/pass/src/QuantizedModelVerifier.cpp
```diff
@@ -22,14 +22,75 @@
 #include <luci/IR/CircleNodes.h>
 #include <luci/IR/CircleNodeVisitor.h>
 
+using LayerInfoMap = std::unordered_map<std::string, luci::LayerInfo *>;
+
 namespace luci
 {
 
+LayerInfoMap QuantizedModelVerifier::create_layer_info_map(loco::Graph *g) const
```
llFreetimell, line 30: This function can be improved just like https://github.com/Samsung/ONE/pull/8555#discussion_r818311347 :)

jinevening: As mentioned https://github.com/Samsung/ONE/pull/8555#discussion_r818339492, the code raises an error.

llFreetimell:
I think almost all codes in this PR are duplicated with #8555.
We may extract these duplicated codes as separate class/files later :)
(Recently reducing duplicated code becomes important T.T)

compiler/luci/pass/src/QuantizedModelVerifier.cpp
llFreetimell: Ah, I missed the comment sorry :(
I thought that all nodes in the graph should be appeared in `info_by_name`, but all names in `info_by_name` should be appeared in the graph T.T
I will post another PR to make it better later :)
Thanks for pointing out!

llFreetimell (approved):
LGTM

PR #8559, opened 2022-03-03
[infra] Makefile for ARM32 build of compiler

seanshpark:
This will introduce a Makefile for ARM32 cross build of some compiler modules.

### DIFF
--- a/infra/nncc/Makefile.arm32
+++ b/infra/nncc/Makefile.arm32
@@ -1,0 +1,144 @@
+#
+# NOTE this is provided as experimental Makefile to ARM32 cross building
+#      some modules of compiler.
+#
+
+BUILD_TYPE?=Debug
+
+CURRENT_DIR=$(shell pwd)
+BUILDFOLDER=build
+ARM32_FOLDER=arm32
+ROOTFS_ARM?=$(CURRENT_DIR)/tools/cross/rootfs/arm
+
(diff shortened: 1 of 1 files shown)


seanshpark:
For #8379, from draft #8407

infra/nncc/Makefile.arm32
```diff
@@ -0,0 +1,144 @@
...
+		-DCMAKE_TOOLCHAIN_FILE=$(ARM32_TOOLCHAIN_FILE) \
+		-DCMAKE_INSTALL_PREFIX="$(ARM32_INSTALL_FOLDER)" \
+		-DENABLE_TEST=ON
+
+
+#
+# builds
+#
+int_build_arm32_host:
+	NNCC_WORKSPACE=$(BUILD_ARM32_HOST) ./nncc build -j1
```
hyunsik-yoon, line 113: (optional) It would be great if we can change `-j1` to something else later. :-) 
   (but this is an experimental Makefile as mentioned in L2, so `-j1` seems OK for now.)

seanshpark: (backgrdound) `j1` was added as I had some terrible compiler segmentation fault -_-;;;
anyway we can add option for this from the user :)
you can send a PR for this if you like.

chunseoklee (approved):

PR #8568, opened 2022-03-04
[res] Revise Sqrt_000 recipe

seanshpark:
This will revise Sqrt_000 recipe to have input as input not constant.

### DIFF
--- a/res/TensorFlowLiteRecipes/Sqrt_000/test.recipe
+++ b/res/TensorFlowLiteRecipes/Sqrt_000/test.recipe
@@ -2,7 +2,6 @@ operand {
   name: "ifm"
   type: FLOAT32
   shape { dim: 1 dim: 3 dim: 3 dim: 2 }
-  filler { tag: "constant" arg: "3.5" }
 }
 operand {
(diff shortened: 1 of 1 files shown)


hyunsik-yoon (approved):

mhs4670go:
@seanshpark I read https://github.com/Samsung/ONE/pull/8568#issuecomment-1058820163 but I still don't know about the reason of this PR. Is it non-input that causes a test failure at arm32?

seanshpark:
> but I still don't know about the reason of this PR

Sorry didn't describe the problem. I made ARM32 luci-value-test to have an input file. this was from simple query to get number of inputs to feed. I didn't think that there will be a model without a model. I thought in real world, no input model would be none-sense. But there was one model in our recipe, `Sqrt_000` with constant input. 

I could fix luci-value-test to handle with no input or fix like this PR.
As I wrote, I thought it would be better to fix this recipe as it makes more sense in view point of a model and little change compare to fix luci-value-test.

mhs4670go (approved):
LGTM

PR #8573, opened 2022-03-04
[recipe] Add network with redundant quantize

BalyshevArtem:
This commit adds network with redundant quantize operation to recipes.

for issue: #8526

ONE-DCO-1.0-Signed-off-by: Artem Balyshev a.balyshev@partner.samsung.com

PR #8575, opened 2022-03-04
[common-artifacts] Support int16 type

mhs4670go:
This commit supports int16 data type.

Related: https://github.com/Samsung/ONE/pull/8573#issuecomment-1059150836

### DIFF
--- a/compiler/common-artifacts/src/TestDataGenerator.cpp
+++ b/compiler/common-artifacts/src/TestDataGenerator.cpp
@@ -46,6 +46,8 @@ H5::PredType hdf5_dtype_cast(const loco::DataType loco_dtype)
   {
     case loco::DataType::U8:
       return H5::PredType::NATIVE_UINT8;
+    case loco::DataType::S16:
+      return H5::PredType::NATIVE_INT16;
     case loco::DataType::S32:
       return H5::PredType::NATIVE_INT32;
     case loco::DataType::S64:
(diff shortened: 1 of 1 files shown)


BalyshevArtem (approved):
LGTM

seanshpark:
@mhs4670go , have you checked python scripts too?

mhs4670go:
@seanshpark I haven't. I thought it was okay because all tests were completed successfully. But as you said, I should have checked it in advance. 

Actually, value test can't be done whose expected data to be compared is from tensorflow interpreter. Because tensorflow doesn't support Quantize op whose input is u8 and output is int16;`luci-interpreter` does. Related codes are like below.
```python
} else {
    // Requantize use case.
    if (input->type == kTfLiteInt16) {
      TF_LITE_ENSURE(context, output->type == kTfLiteInt8 ||
                                  output->type == kTfLiteInt16 ||
                                  output->type == kTfLiteInt32);
    } else {
      TF_LITE_ENSURE(context,
                     input->type == kTfLiteInt8 || input->type == kTfLiteUInt8);
      TF_LITE_ENSURE(
          context, output->type == kTfLiteUInt8 || output->type == kTfLiteInt8); # here
    }   
```

So, AFAIK, there is no script that should be checked more.

jinevening (approved):
LGTM

seanshpark:
@mhs4670go this PR itself is OK but other modules like `luci-value-test` don't support S16.
Could you search and revise them too?

mhs4670go:
@seanshpark Okay. I'm gonna post related PR:)

seanshpark (approved):
LGTM

seanshpark:
@mhs4670go , and if we don't have any S16 recipes, could you add one or two for any Op?

PR #8579, opened 2022-03-07
[luci] Insert Quantize Op to mixed-precision point

jinevening:
This inserts Quantize Ops to mixed-precision points.

---

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -97,6 +97,201 @@ luci::CircleQuantize *create_quantize_op(luci::CircleNode *node, loco::DataType 
 namespace luci
 {
 
+namespace
+{
+
+/**
+ * Insert Quantize operator for mixed-precision quantization
+ * 1. Before input feature map (only for non-const)
+ * 2. After output feature map
+ *
+ * For example, if default_dtype = U8 and op_dtype = S16,
+ * 1. Quantize Op for U8->S16 is inserted before ifm
+ * 2. Quantize Op for S16->U8 is inserted after ofm
+ *
+ * Why not insert Quantize Op for const ifm?
+ * We quantize const tensor at once to preserve precision.
+ * For example, if default dtype = U8, op_dtype = S16, and op is CONV2D,
+ * We directly quantize weights to 16 bits, not 8->16 bits.
+ */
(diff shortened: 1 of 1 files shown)


seanshpark:
Q) when we enable new Op, how do we know we need to add to this new codes?

jinevening:
We can know that as default visitor throws an exception for NYI operators.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -97,6 +97,201 @@ luci::CircleQuantize *create_quantize_op(luci::CircleNode *node, loco::DataType
...
+  }
+
+  // Skip output layer
+  void visit(luci::CircleOutput *) {}
+
+  // Ops that receive a single activation as an input
+  INSERT_QUANTIZE_TO_UNARY_OP(luci::CircleArgMax, input)
+  INSERT_QUANTIZE_TO_UNARY_OP(luci::CircleArgMin, input)
+  INSERT_QUANTIZE_TO_UNARY_OP(luci::CircleBatchToSpaceND, input)
+  INSERT_QUANTIZE_TO_UNARY_OP(luci::CircleConv2D, input)
```
seanshpark, line 191: This assumes other inputs should be CircleConst? no need to do check?

jinevening: Quantizer only accepts const weights and bias.

If the input model has non-const weights or bias, quantization will fail, i.e., exception will be thrown during quantization or verification.

jinevening: It would be better to check conditions for constant inputs of each operator somewhere (maybe at the beginning of quantization).

seanshpark: Will you add the check in this PR or in another?

jinevening: I think that should be implemented as a new pass or visitor. I'll check it in another PR.

seanshpark (approved):
LGTM

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: I've made https://github.com/Samsung/ONE/issues/8580.

llFreetimell (approved):
LGTM

PR #8584, opened 2022-03-07
[luci-pass-value-test] Support int16 type

mhs4670go:
This commit supports int16 data type.


seanshpark (approved):
LGTM

PR #8589, opened 2022-03-07
[luci] Add test for quantization with layer info

jinevening:
This adds test for quantization with layer info.

---
Related to: https://github.com/Samsung/ONE/pull/8558#issuecomment-1057792092

### DIFF
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -244,6 +244,7 @@ private:
   INSERT_QUANTIZE_TO_BINARY_OP(luci::CircleMinimum, x, y)
   INSERT_QUANTIZE_TO_BINARY_OP(luci::CircleMul, x, y)
   INSERT_QUANTIZE_TO_BINARY_OP(luci::CircleNotEqual, x, y)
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -244,6 +244,7 @@ struct InsertQuantizeOp final : public luci::CircleNodeMutableVisitor<void>
   INSERT_QUANTIZE_TO_BINARY_OP(luci::CircleMinimum, x, y)
   INSERT_QUANTIZE_TO_BINARY_OP(luci::CircleMul, x, y)
   INSERT_QUANTIZE_TO_BINARY_OP(luci::CircleNotEqual, x, y)
+  INSERT_QUANTIZE_TO_BINARY_OP(luci::CircleOneHot, on_value, off_value)
```
jinevening, line 247: This code is required to add test for OneHot op.

seanshpark:
There's a `Type` type that uses general word `Type` which is confusing.
I assume this is data type and also I may have agreed as this has landed.
But I highly recommend to change this not `Type` but something else.

jinevening:
`Type` is alias of `loco::DataType` in this file. Renaming will be done in the next PR.

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

PR #8613, opened 2022-03-11
[luci] Extract layer info map code

jinevening:
This extracts code for layer info map to reduce duplication.


llFreetimell (approved):
LGTM

seanshpark (approved):
LGTM

PR #8616, opened 2022-03-11
[luci] Initialize pointers with nullptr

jinevening:
This initializes pointers with nullptr.


seanshpark (approved):
LGTM

Issue #8628, opened 2022-03-15
[Compiler FE] Can remove locomotiv from ARM32 build?


seanshpark:
Investigate if we can remove _locomotiv_ from build for ARM32.
- _locomotiv_ is used for constant folding pass in _logo_
- we may arrange files or introduce _logo_xxx_ module to disconnect dependency

seanshpark:
For #8379

seanshpark:
choice 1) introduce _logo-ex_ for passes that use _locomotiv_
choice 2) move files in _logo_ to _logo-core_ that are used in _luci_

I think choice 1 looks better for responsibility

seanshpark:
draft code shows that we can unlink _locomotiv_ from ARM32 build.

seanshpark:
CC @llFreetimell

seanshpark:
CC @ejjeong

seanshpark:
done

PR #8631, opened 2022-03-15
DRAFT: introduce logo-ex

seanshpark:
on-going draft to introduce logo-ex for passes with locomotiv


seanshpark:
For #8628

seanshpark:
done

PR #8637, opened 2022-03-15
DRAFT rename to circle-partitioner

seanshpark:
draft to rename to use circle-partitioner


seanshpark:
This is to follow naming convention of commands.

seanshpark:
done

PR #8639, opened 2022-03-15
[circle-partitioner] Rename tool

seanshpark:
This will add tool as circle-partitioner to follow other command naming
conventions. Old name will be removed after changing other codes.


seanshpark:
From draft #8637

hseok-oh (approved):

PR #8595, opened 2022-03-08
[luci] Disable mixed-precision for improper Ops

jinevening:
This disables mixed-precision quantization of improper Ops.
- Logical Ops
- Index Ops
- Ops with multiple outputs

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

PR #8640, opened 2022-03-15
[luci] Exclude output from layer info map.

jinevening:
This excludes output layer from layer info map.

### DIFF
--- a/compiler/luci/pass/src/helpers/LayerInfoMap.cpp
+++ b/compiler/luci/pass/src/helpers/LayerInfoMap.cpp
@@ -34,6 +34,9 @@ LayerInfoMap layer_info_map(loco::Graph *g, std::vector<LayerInfo> &layers_info)
     for (auto node : loco::active_nodes(loco::output_nodes(g)))
     {
       auto cnode = loco::must_cast<luci::CircleNode *>(node);
+      if (cnode->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
+        continue;
+
       if (cnode->name() == name)
       {
         if (info_by_name.find(name) != info_by_name.end())


seanshpark:
Q) could you share background of this change?

compiler/luci/pass/src/helpers/LayerInfoMap.cpp
```diff
@@ -34,6 +34,9 @@ LayerInfoMap layer_info_map(loco::Graph *g, std::vector<LayerInfo> &layers_info)
     for (auto node : loco::active_nodes(loco::output_nodes(g)))
     {
       auto cnode = loco::must_cast<luci::CircleNode *>(node);
+      if (cnode->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
```
jinevening, line 37: To reviewers:

Output layer has the same name with the last layer, so the exception for duplicate layer name is thrown if `layers_info` has the last layer. This PR resolves the problem.

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

compiler/luci/pass/src/helpers/LayerInfoMap.cpp
seanshpark: > so the exception for duplicate layer name is thrown

Not sure but, maybe check with multiple output nodes like `CircleSplit`

jinevening: For now, Ops with multiple outputs are disabled (https://github.com/Samsung/ONE/pull/8595). Supporting Ops w/ multiple outputs can be a future task.

PR #8653, opened 2022-03-15
[tfldump] Remove link with flatbuffers

seanshpark:
This will remove direct link with flatbuffers as mio-tflite will do.


seanshpark:
To fix recent push-nncc-debug CI failure reports.

hseok-oh (approved):
LGTM

Issue #8398, opened 2022-02-09
[one-quantize] Support mixed-precision quantization


jinevening:
## What

Let's support manual mixed-precision quantization. 'Manual' indicates that quantization configuration file is manually written by users.

![image](https://user-images.githubusercontent.com/5449554/153107541-65946412-efbd-43dc-b52c-dadfd0c51b62.png)

- [x] Quantization configuration file #8399
- [x] Update quantizer
  - [x] FakeQuant
  - [x] ~RecordMinMax~ Nothing to update
  - [x] QuantizeWithMinMax
  - [x] one-quantize

NOTE: In this issue, we'll assume that the output model will not contain any fp32 layer.

seanshpark:
> Quantization configuration file

Have you decided which format to use?

jinevening:
> Have you decided which format to use? and which library?

JSON looks good to me due to the following reasons
    - Widely used, MIT licensed parser: jsoncpp (https://github.com/open-source-parsers/jsoncpp)
    - More concise than XML
    - Allow array

I've made https://github.com/Samsung/ONE/issues/8399 for quantization configuration file. Let's discuss there :)

jinevening:
Manual mixed-precision quantization is now working.

PR #8655, opened 2022-03-16
[one-cmds] Support mixed-precision quantization

jinevening:
This adds an option for mixed-precision quantization and its test.

--
Related to: https://github.com/Samsung/ONE/issues/8398

### DIFF
--- a/compiler/one-cmds/one-quantize
+++ b/compiler/one-cmds/one-quantize
@@ -125,6 +125,12 @@ def _get_parser():
         help=
         "Force MaxPool Op to have the same input/output quantparams. NOTE: This option can degrade accuracy of some models.)"
     )
+    quantization_group.add_argument(
+        '--quant_config',
+        type=str,
+        help=
+        "Path to the quantization configuration file.)"
+    )
 
     # arguments for force_quantparam option
     force_quantparam_group = parser.add_argument_group(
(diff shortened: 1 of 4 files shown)


compiler/one-cmds/one-quantize
```diff
@@ -125,6 +125,12 @@ def _get_parser():
         help=
         "Force MaxPool Op to have the same input/output quantparams. NOTE: This option can degrade accuracy of some models.)"
     )
+    quantization_group.add_argument(
+        '--quant_config',
+        type=str,
+        help=
+        "Path to the quantization configuration file.)"
```
seanshpark, line 132: Q) end of help has `)`... others too. is there any particular reason?

compiler/one-cmds/tests/one-quantize_009.qconf.json
```diff
@@ -0,0 +1,43 @@
+// Quantization configuration file
```
seanshpark, line 1: (just comment; not sure adding // comment in json is standard. I think the json lib that we use may handle this. if we change the lib for some reason, this maybe a problem. you can see this is in read background)

compiler/one-cmds/one-quantize
jinevening: I think they are typos. Thanks for finding.

seanshpark: I think fix in anther PR will do.

seanshpark (approved):
LGTM

compiler/one-cmds/tests/one-quantize_009.qconf.json
jinevening: I've added comment for readability although it is not json standard. But in another thought, it would be better to follow the standard, so that we can parse this file in another platform (ex: Python).

seanshpark (approved):
LGTM

llFreetimell (approved):
LGTM

PR #8659, opened 2022-03-16
[cir-par-val-test] Use circle-partitioner

seanshpark:
This will revise to use circle-partitioner tool with hyphen.


seanshpark:
From draft #8637

hseok-oh (approved):

PR #8665, opened 2022-03-17
[moco-tf] Revise test to include logo_ex

seanshpark:
This will revise moco_tf_frontend_test to include logo_ex.


seanshpark:
For #8628, from draft #8631 (this should have been in #8663)

PR #8667, opened 2022-03-17
[moco-tf] Require logo-ex

seanshpark:
This will revise to require logo-ex module.


seanshpark:
For #8628, from draft #8631 (this should have been in #8663)

yunjayh (approved):
LGTM 😅

Issue #8664, opened 2022-03-17
Revise circle schema


seanshpark:
Let's revise circle schema base from TF 2.7.0, schma_v3b.fbs.

seanshpark:
Related #6216

seanshpark:
Background
- revise common schema may break compile/runtime build

Preliminary plan
- introduce schema copy in res for compiler
- revise _mio-circle_ to use the copy
- (runtime can start upgrading)
- add new (based on v3b) schema as v0.4 and introduce _mio-circle04_
- revise modules to use _mio-circle04_ one by one
- after runtime finishes revision, change _mio-circle04_ to use common schema

hseok-oh:
Schema draft: #8442

seanshpark:
Modules to work
- [ ] add mio-circle04
- [ ] circledump
- [ ] circlechef
- [ ] tflite2circle
- [ ] luci import
- [ ] luci export
- [ ] ...

seanshpark:
@chunseoklee , @hseok , new schema has now SignatureDef field in the format (currently not migrated from tflite but will work on this soon). Do you think we need this? If so, can you share any benefits you can think of?

chunseoklee:
> @chunseoklee , @hseok-oh , new schema has now SignatureDef field in the format (currently not migrated from tflite but will work on this soon). Do you think we need this? If so, can you share any benefits you can think of?

1. I am not used to SignatureDef field(seems like custom IO description)
2. Not planned to process it on onert side. But, we may need in the future.

seanshpark:
@chunseoklee , OK. I think it would be better to take some time and understand `SignatureDef`.

In short, my short understanding of one of the feature is that, this provides input and output node order.
Previous version was the tensor (operand) order itself was the input/output order.
Now, SignatureDef provides the order and the tensor (operand) may be in unordered sequence.
I also don't fully understand the purpose of `SignatureDef`.

chunseoklee:
>  OK. I think it would be better to take some time and understand SignatureDef.

I agree. I will take a look at sugnaturedef from TFL's viewpoint.

At a glance, 

1. In TFL interpreter, signature itself is experimental
2. In schema, signaturedef is optional.

It is strange that SignatureDef affects conversion process even although optional. Anyway, I will take a look

seanshpark:
close as inactive for long time

PR #8668, opened 2022-03-17
DRAFT revise compiler circle schema

seanshpark:
on-going draft to revise compiler circle schema based on v3b


seanshpark:
For #8664

seanshpark:
build and test is OK.
need to revise tflite2circle to migrate signaturedefs correctly.

PR #8669, opened 2022-03-17
[luci] Allow concatenating indices in quantized model

jinevening:
This allows to concatenate indices in quantized model.

### DIFF
--- a/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
+++ b/compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
@@ -466,6 +466,10 @@ bool PropagateQParamBackwardPass::run(loco::Graph *g)
     auto circle_node = loco::must_cast<luci::CircleNode *>(node);
     INFO(l) << "PropagateQParamBackwardPass visit node: " << circle_node->name() << std::endl;
 
+    // We can't propagate non-existent qparam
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/src/PropagateQParamBackwardPass.cpp
```diff
@@ -466,6 +466,9 @@ bool PropagateQParamBackwardPass::run(loco::Graph *g)
     auto circle_node = loco::must_cast<luci::CircleNode *>(node);
     INFO(l) << "PropagateQParamBackwardPass visit node: " << circle_node->name() << std::endl;
 
+    if (circle_node->quantparam() == nullptr)
+      continue;
```
jinevening, line 470: This code is necessary as Bool/S32/S64 type Ops do not have quantparam at this point.

seanshpark: (optional) maybe add a NOTE or TODO about this?

seanshpark (approved):
LGTM

jinevening:
The purpose of PR is to quantize the below model (concatenation of int64 tensors).

image

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #8671, opened 2022-03-17
[luci] Support rank4 add/mul in batchnorm finder

jinevening:
This supports rank4 add/mul in batchnorm finder.


seanshpark (approved):
LGTM

PR #8677, opened 2022-03-18
[res] tflite schema 2.7.0 and 2.8.0

seanshpark:
This will add tflite schema of version 2.7.0 and 2.8.0


seanshpark:
These files are just for reference.

hseok-oh (approved):

PR #8678, opened 2022-03-18
[exo] Revise to use circle schema from res folder

seanshpark:
This will revise to use circle schema from res folder so that upgrading
will not break exo.


seanshpark:
For #8664, from draft #8668

hseok-oh (approved):
LGTM

yunjayh (approved):
LGTM

PR #8681, opened 2022-03-18
[mio-circle] Revise to use circle schema from res folder

seanshpark:
This will revise to use circle schema from res folder so that upgrading
will not change version of mio-circle.


seanshpark:
For #8664, from draft #8668

hseok-oh (approved):

yunjayh (approved):
LGTM

PR #8697, opened 2022-03-22
[common-artifacts] Remove mio-circle

seanshpark:
This will remove using mio-circle as it is not directly used but through luci.


seanshpark:
from draft #8668

mhs4670go (approved):
LGTM

hseok-oh (approved):

PR #8700, opened 2022-03-22
[luci/svc] Remove mio-circle

seanshpark:
This will remove link with mio-circle as it is not used.


seanshpark:
related #8668

yunjayh (approved):

hseok-oh (approved):

PR #8703, opened 2022-03-22
[circle-verify] Show skip message

seanshpark:
This will show skip message if mio_circle04 target is not found.


seanshpark:
To check #8586

hseok-oh (approved):
LGTM

PR #8715, opened 2022-03-23
[circle-quantizer] Remove link with mio-circle

seanshpark:
This will remove link with mio-circle as it is not used directly.


seanshpark:
@jinevening ping

jinevening (approved):
Ah, sorry I missed it.

LGTM Thank you.

PR #8673, opened 2022-03-17
[Draft][quantizer] Support mixed-precision in Ops with multiple outputs

BalyshevArtem:
This commit adds support for mixed-precision in Ops with multiple outputs.

for issue #8654 

ONE-DCO-1.0-Signed-off-by: Artem Balyshev a.balyshev@partner.samsung.com

Issue #8654, opened 2022-03-16
[quantizer] Support mixed-precision in Ops with multiple outputs


jinevening:
## What

Let's support mixed-precision quantization in Ops with multiple outputs.

## Why

For now, Ops with multiple outputs are disabled (https://github.com/Samsung/ONE/pull/8595). Supporting Ops w/ multiple outputs can be a future task.

_Originally posted by @jinevening in https://github.com/Samsung/ONE/pull/8640#discussion_r827526608_

jinevening:
There is a naming issue in multi-out Ops.

For example, for Split Op with num_split = 2, three circle nodes are generated.
1. CircleSplit (name: "split0") <- name is copied from the first SplitOut
2. CircleSplitOut (index: 0, name: "split0")
3. CircleSplitOut (index: 1, name: "split1")

If quantization config file designates precision of "split0", Quantize Ops will be inserted before/after of both 1 and 2, which is not intended.

BalyshevArtem:
I would like to work on this issue, can I? :)

jinevening:
> I would like to work on this issue, can I? :)

Sure :)

jinevening:
I'm closing this issue as @BalyshevArtem has finished this job :)

PR #8717, opened 2022-03-23
[luci] Support mixed-precision in Ops with multiple outputs

BalyshevArtem:
This commit adds support for mixed-precision in Ops with multiple outputs.

draft: #8673
issue: #8654

### DIFF
--- a/compiler/luci/pass/src/helpers/LayerInfoMap.cpp
+++ b/compiler/luci/pass/src/helpers/LayerInfoMap.cpp
@@ -22,6 +22,154 @@
 
 namespace luci
 {
+namespace
+{
+
+bool is_multiple_output_node(const luci::CircleNode *node)
+{
+  switch (node->opcode())
+  {
+    // The following nodes have multiple outputs. Output tensors are not produced by themselves but
+    // by the corresponding *Out nodes.
+    case luci::CircleOpcode::SPLIT:
+    case luci::CircleOpcode::SPLIT_V:
+    case luci::CircleOpcode::TOPK_V2:
+    case luci::CircleOpcode::UNIQUE:
+    case luci::CircleOpcode::UNPACK:
+      return true;
+    // TODO: Support ops
+    case luci::CircleOpcode::BIDIRECTIONAL_SEQUENCE_LSTM:
+    case luci::CircleOpcode::CUSTOM:
+    case luci::CircleOpcode::IF:
+    case luci::CircleOpcode::NON_MAX_SUPPRESSION_V4:
+    case luci::CircleOpcode::NON_MAX_SUPPRESSION_V5:
+    case luci::CircleOpcode::WHILE:
+      throw std::runtime_error("Unsupported op now");
+    default:
+      return false;
+  }
+}
+
+const luci::CircleNode *get_multi_output_node(const luci::CircleNode *node)
+{
+  if (is_multiple_output_node(node))
+    return node;
+
+  switch (node->opcode())
+  {
+    // The following nodes denote outputs of multiple-output nodes.
+    case luci::CircleOpcode::CIRCLESPLITOUT:
+    {
+      const auto split_out = loco::must_cast<const CircleSplitOut *>(node);
+      return loco::must_cast<luci::CircleNode *>(split_out->input());
+    }
+    case luci::CircleOpcode::CIRCLESPLITVOUT:
+    {
+      const auto splitv_out = loco::must_cast<const CircleSplitVOut *>(node);
+      return loco::must_cast<luci::CircleNode *>(splitv_out->input());
+    }
+    case luci::CircleOpcode::CIRCLETOPKV2OUT:
+    {
+      const auto top_kv2_out = loco::must_cast<const CircleTopKV2Out *>(node);
+      return loco::must_cast<luci::CircleNode *>(top_kv2_out->input());
+    }
+    case luci::CircleOpcode::CIRCLEUNIQUEOUT:
+    {
+      const auto unique_out = loco::must_cast<const CircleUniqueOut *>(node);
+      return loco::must_cast<luci::CircleNode *>(unique_out->input());
+    }
+    case luci::CircleOpcode::CIRCLEUNPACKOUT:
+    {
+      const auto unpack_out = loco::must_cast<const CircleUnpackOut *>(node);
+      return loco::must_cast<luci::CircleNode *>(unpack_out->input());
+    }
+    // TODO: Support these ops
+    case luci::CircleOpcode::CIRCLEBIDIRECTIONAL_SEQUENCE_LSTM_OUT:
+    case luci::CircleOpcode::CIRCLECUSTOMOUT:
+    case luci::CircleOpcode::CIRCLEIFOUT:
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/src/helpers/LayerInfoMap.test.cpp
```diff
@@ -43,6 +43,61 @@ class SoftmaxTestGraph : public luci::test::TestIOGraph
...
+      _split_dim->shape({1});
+      _split_dim->name("split_dim");
+    }
+
+    _split = g()->nodes()->create<luci::CircleSplit>();
+    {
+      _split->input(input());
+      _split->num_split(2);
+      _split->split_dim(_split_dim);
+      _split->name("split");
```
jinevening, line 67: Could you make a test that CircleSplit has the same name with the first CircleSplitOut ? It is the actual behavior of `luci/import`.

```suggestion
      _split->name("split0");
```

compiler/luci/pass/src/helpers/LayerInfoMap.test.cpp
```diff
@@ -65,6 +120,54 @@ TEST(LayerInfoMapTest, simple_test)
...
+    info.granularity = luci::QuantizationGranularity::ChannelWise;
+  }
+  std::vector<luci::LayerInfo> v;
+  v.emplace_back(info);
+  auto map = luci::layer_info_map(g.g(), v);
+
+  EXPECT_EQ(map.size(), 3);
+  EXPECT_EQ("split0", map["split"]->name);
+  EXPECT_EQ("split0", map["split0"]->name);
+  EXPECT_EQ("split0", map["split1"]->name);
```
jinevening, line 141: Hmm.. This behavior seems a bit weird. Could you make the behavior as below? I think that does not change the final quantized model.

```suggestion
  EXPECT_EQ("split", map["split"]->name);
  EXPECT_EQ("split0", map["split0"]->name);
  EXPECT_EQ("split1", map["split1"]->name);
```

BalyshevArtem: I agree with you, and I tried to solve this, but I found difficulties with it.
Actually, `map["some_name"]` returns `luci::LayerInfo *` and if initially there was no structure for the "some_name" (which is output node from mulitple output node in this example), then I see two solutions that will not lead to big changes. 

First, we add to the `map` for each out node name the current `LayerInfo` (which can contain name one of the output node, for example `split1`) and then we do not add anything to the original `std::vector<LayerInfo>` (this variant is current implementation).

The second way is to create a `LayerInfo` structure for each out node name and add to the `std::vector<LayerInfo>`, but at the same time I found that pointers can invalidate in the map because it keeps pointers to `luci::LayerInfo`.

Another way may be to create a structure for each output node (for multi output nodes) at the moment when the vector is created: https://github.com/Samsung/ONE/blob/76e9a6c1ef6498abde430ae279c9c0feda104daf/compiler/luci/pass/src/CircleQuantizer.cpp#L335

BalyshevArtem: The third option will add a search and check that the current name does not belong to multiple output node (or its output node), and so on for each name.

jinevening: Ah, it seems that `LayerInfoMap` contains inappropriate dtype. I didn't expect that new `LayerInfo` (not included in the `std::vector<LayerInfo>`) will be added.

I think it would be better to update `LayerInfoMap` to contain `LayerInfo`, instead of `LayerInfo *`. Then it'll be possible to add new `LayerInfo`s to the map.

Could you proceed after that change?

jinevening: I've made #8726 PTAL

compiler/luci/pass/src/helpers/LayerInfoMap.cpp
```diff
@@ -22,6 +22,154 @@
...
+
+    if (it == layers_info.end())
+    {
+      info_by_name[name] = {name, layer_info.dtype, layer_info.granularity};
+    }
+    else
+    {
+      info_by_name[name] = *it;
+    }
+  }
```
jinevening, line 145: Is it ok to replace the code with the below one line?

```suggestion
  info_by_name[name] = {name, layer_info.dtype, layer_info.granularity};
```

compiler/luci/pass/src/helpers/LayerInfoMap.cpp
```diff
@@ -22,6 +22,154 @@
...
+      if (it == layers_info.end())
+      {
+        info_by_name[succs_circle_node_name] = {succs_circle_node_name, layer_info.dtype,
+                                                layer_info.granularity};
+      }
+      else
+      {
+        info_by_name[succs_circle_node_name] = *it;
+      }
+    }
```
jinevening, line 168: Similarly, is it possible to simplify the code?

```suggestion
    info_by_name[succs_circle_node_name] = {succs_circle_node_name, layer_info.dtype,
                                                layer_info.granularity};
```

compiler/luci/pass/src/helpers/LayerInfoMap.cpp
```diff
@@ -56,8 +212,6 @@ LayerInfoMap layer_info_map(loco::Graph *g, std::vector<LayerInfo> &layers_info)
                                ". Check layer names in the quantization configuration file.");
   }
 
-  assert(info_by_name.size() == layers_info.size()); // FIX_ME_UNLESS
```
jinevening, line 59: Could you add asserts for `info_by_name` or leave TODO messages as below?

```cpp
// TODO Check all names in layers_info exist in the info_by_name
// TODO Check names in info_by_name but not in layers_info are from virtual outputs
```

compiler/luci/pass/src/helpers/LayerInfoMap.cpp
BalyshevArtem: Oh, yes, you are right, it is much better, thank you :)

jinevening (approved):
LGTM Thank you

seanshpark (dismissed):

jinevening:
@BalyshevArtem  Please fix the test failure.

```
17:08:54 /opt/jenkins_slave/workspace/nnfw/master/pr-nncc-debug/compiler/luci/pass/src/helpers/LayerInfoMap.cpp: In function ‘void luci::{anonymous}::add_multi_output_node(luci::LayerInfoMap&, luci::LayerInfo&, const luci::CircleNode*, std::vector<luci::LayerInfo>&)’:
17:08:55 /opt/jenkins_slave/workspace/nnfw/master/pr-nncc-debug/compiler/luci/pass/src/helpers/LayerInfoMap.cpp:105:82: error: unused parameter ‘layers_info’ [-Werror=unused-parameter]
17:08:55                             const luci::CircleNode *node, std::vector<LayerInfo> &layers_info)
17:08:55                                                           ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
```

BalyshevArtem:
@nnfw-bot test nncc-release

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

Issue #8735, opened 2022-03-25
[quantizer] Extract h5 importer


jinevening:
## What

Let's extract HDF5Importer from record-minmax and make it a separate module.

## Why 

Other modules (ex: circle-eval-diff https://github.com/Samsung/ONE/issues/8634) can use HDF5Importer. This can reduce code duplication.

jinevening:
TODO

- [ ] Create h5-importer module
- [ ] Update record-minmax

jinevening:
Done.

PR #8737, opened 2022-03-25
[dio-hdf5] Introduce h5-importer module

jinevening:
This introduces a module to help loading h5 files.

---
Related to: https://github.com/Samsung/ONE/issues/8735
Draft PR: https://github.com/Samsung/ONE/pull/8736


jinevening:
To reviewers:

Most of the code were copied from `HDF5Importer.h` and `HDF5Importer.cpp` in `record-minmax` module.

compiler/h5-importer/include/h5_importer/HDF5Importer.h
```diff
@@ -0,0 +1,78 @@
...
+{
+public:
+  explicit HDF5Importer(const std::string &path);
+
+public:
+  /**
+   * @brief importGroup has to be called before readTensor is called
+   *        Otherwise, readTensor will throw an exception
+   */
+  void importGroup(const std::string &group) { _group = _file.openGroup(group); }
```
jinevening, line 48: I slightly modified the interface of this method to allow the importer can load any group.

Previously, the group name was hard-coded for use by `record-minmax` as below.

```cpp
  void importGroup() { _group = _file.openGroup("value"); }
```

seanshpark:
please add some unit tests.

compiler/h5-importer/src/HDF5Importer.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "h5_importer/HDF5Importer.h"
+
+#include <H5Cpp.h>
+
+#include <string>
+#include <cassert>
+#include <stdexcept>
+
+using Shape = std::vector<loco::Dimension>;
+using DataType = loco::DataType;
```
jinevening, line 26: I also changed the type of `Shape` and `DataType` to remove dependency from luci-interpreter.

Previously, Shape and DataType used struct in luci-interpreter.
```
using Shape = luci_interpreter::Shape;
using DataType = luci_interpreter::DataType;
```

HDF5Importer will work in the same way even after this change.

seanshpark:
name of the module is `h5-importer` which implies this may be the unique h5 format file importer.
why we use names like loco, luci, moco is that we can have multiple versions of modules for some feature.
please provide a name for this reason.

our `record-minmax` should be renamed for this reason too.

jinevening:
@seanshpark I see. Can I use the name `hoof`? Just picked a word starting with `h` from dictionary.

seanshpark:
> Can I use the name hoof? Just picked a word starting with h from dictionary.
> Naming is difficult

Yes... naming is soo difficult... `hoof` looks nice :)
Last weekend, I tought of something. We have modules with `mio-` that came from `model I/O`.
As this is for data, how about `dio-hdf5` or `dio-h5` ?

jinevening:
dio-hdf5 looks very good. Thanks.

compiler/h5-importer/README.md
```diff
@@ -0,0 +1,29 @@
+# h5-importer
+
+_h5_importer_ is a library to help loading hdf5 files used by ONE quantizer.
```
mhs4670go, line 3:
```suggestion
_h5_importer_ is a library to help loading hdf5 files.
```
I think this module can be imported to other modules as well.

compiler/h5-importer/include/h5_importer/HDF5Importer.h
```diff
@@ -0,0 +1,78 @@
...
+namespace h5_importer
+{
+
+// HDF5Importer reads an input data saved in the hdf5 file in the given path
+// The hierarchy of the hdf5 file is as follows.
+// Group "/"
+//  > Group <group_name>
+//    > Group <data_idx>
+//      > Dataset <input_idx>
+// data_idx : index of the data (dataset file can contain multiple datas)
```
mhs4670go, line 35:
```suggestion
// data_idx : index of the data (dataset file can contain multiple data)
```

compiler/dio-hdf5/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+  return()
+endif(NOT HDF5_FOUND)
+
+file(GLOB_RECURSE SOURCES "src/*.cpp")
+file(GLOB_RECURSE TESTS "src/*.test.cpp")
+list(REMOVE_ITEM SOURCES ${TESTS})
+
+add_library(dio_hdf5 SHARED ${SOURCES})
+target_include_directories(dio_hdf5 PUBLIC include)
+target_include_directories(dio_hdf5 PRIVATE ${HDF5_INCLUDE_DIRS})
```
seanshpark, line 14: As `HDF5Importer.h` has `#include <H5Cpp.h>` it would be better to use `PUBLIC`

compiler/dio-hdf5/include/dio_hdf5/HDF5Importer.h
```diff
@@ -0,0 +1,78 @@
...
+#ifndef __DIO_HDF5_H__
+#define __DIO_HDF5_H__
+
+#include <H5Cpp.h>
+
+#include <loco.h>
+
+#include <stdexcept>
+
+namespace dio_hdf5
```
seanshpark, line 26: (optional) if we follow `mio`, this can be changed to `namespace dio { namespace hdf5 {` as `dio::hdf5`

compiler/dio-hdf5/CMakeLists.txt
mhs4670go: I'm not sure but IIUC, the `PRIVATE` keyword applies to the module that link `dio-hdf5` library. So, if we don't want to make that module use `HDF5_INCLUDE_DIRS`, it might as well set it to `PRIVATE`.

compiler/dio-hdf5/README.md
```diff
@@ -0,0 +1,29 @@
+# dio-hdf5
```
mhs4670go, line 1: Would be better let us know what is `dio`.

compiler/dio-hdf5/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+target_link_libraries(dio_hdf5 ${HDF5_CXX_LIBRARIES})
+target_link_libraries(dio_hdf5 loco)
```
seanshpark, line 16: we can set explict `PUBLIC` or `PRIVATE` for link. I think it would be `PUBLIC`.

compiler/dio-hdf5/CMakeLists.txt
seanshpark: As I know, when we include `HDF5Importer.h` with `PRIVATE` then user module will not see `${HDF5_INCLUDE_DIRS}` thus compile will fail. If this is not the case then we can use `PRIVATE` or we should use `PUBLIC`

jinevening: Whether it is PUBLIC or PRIVATE, `HDF5Importer.h` can use `H5Cpp.h`.

I set this as `PRIVATE`, because I wanted to expose only `HDF5Importer` to other modules. If we make dio-hdf5 `PUBLIC`, other modules will be possible to access H5Cpp.h when dio-hdf5 is linked.

compiler/dio-hdf5/src/HDF5Importer.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "dio_hdf5/HDF5Importer.h"
+
+#include <H5Cpp.h>
+
+#include <string>
+#include <cassert>
+#include <stdexcept>
+
+using Shape = std::vector<loco::Dimension>;
+using DataType = loco::DataType;
```
mhs4670go, line 26: I'm not sure but is it okay to place this `using` statement on here? I mean on outside of no namespace?

compiler/dio-hdf5/CMakeLists.txt
jinevening: I think PRIVATE would be better, as we don't want to expose loco interface when linking dio_hdf5.

compiler/dio-hdf5/src/HDF5Importer.cpp
jinevening: Yes :) Actually it is a recommended way by @parjong.

compiler/dio-hdf5/src/HDF5Importer.cpp
```diff
@@ -0,0 +1,167 @@
...
+  {
+    return DataType::FLOAT32;
+  }
+  if (h5_type == H5::PredType::STD_I32BE || h5_type == H5::PredType::STD_I32LE)
+  {
+    return DataType::S32;
+  }
+  if (h5_type == H5::PredType::STD_I64BE || h5_type == H5::PredType::STD_I64LE)
+  {
+    return DataType::S64;
```
mhs4670go, line 60: (optional) If `switch` statement is used instead, it would increase the score of static analysis tool and readability a bit:)

compiler/dio-hdf5/src/HDF5Importer.cpp
```diff
@@ -0,0 +1,167 @@
...
+    case DataType::FLOAT32:
+      readTensorData(tensor, static_cast<float *>(buffer));
+      break;
+    case DataType::S32:
+      readTensorData(tensor, static_cast<int32_t *>(buffer));
+      break;
+    case DataType::S64:
+      readTensorData(tensor, static_cast<int64_t *>(buffer));
+      break;
+    case DataType::BOOL:
```
mhs4670go, line 159: Is it bool rather than U8?

compiler/dio-hdf5/CMakeLists.txt
seanshpark: Found out with @jinevening that `H5Cpp.h` is installed in `overlay` folder and other libs provide this. If we make a small project without others then compile fails with `H5Cpp.h` not found...

jinevening: In offline talk with @seanshpark , we've confirmed that @seanshpark is right.

`HDF5Importer.h` is PUBLIC, and it uses `H5Cpp.h`, so we have to set PUBLIC for HDF5 libraries.

What makes things confusing was that HDF5 library with PRIVATE keyword is also compilable. That's because `overlay` (where HDF5 is installed) directory is included by another module (such as luci_import - mio-circle). Technically, that should not be compiled.

Thanks for lesson! @seanshpark

compiler/dio-hdf5/src/HDF5Importer.cpp
jinevening: U8 is not yet implemented.

I implemented BOOL first to support logical Ops.

compiler/dio-hdf5/CMakeLists.txt
jinevening: loco is used by HDF5Importer.h, so it should be public.

compiler/dio-hdf5/src/HDF5Importer.cpp
jinevening: h5_type is not an integer (enum), so It can't be used with switch.

compiler/dio-hdf5/src/HDF5Importer.cpp
mhs4670go: Oh, I thought `static_cast<uint8_t *>(buffer)` should be `static_cast<bool *>(buffer)`. Am I wrong?

compiler/dio-hdf5/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+        FILES_MATCHING PATTERN "*.h")
+
+if(NOT ENABLE_TEST)
+  return()
+endif(NOT ENABLE_TEST)
+
+nnas_find_package(GTest REQUIRED)
+
+GTest_AddTest(dio_hdf5_test ${TESTS})
+target_include_directories(dio_hdf5_test PRIVATE include)
```
seanshpark, line 29: (optional) as `dio_hdf5_test` is executable and usually this won't be used for other modules, we can ommit `PRIVATE`

compiler/dio-hdf5/include/dio_hdf5/HDF5Importer.h
```diff
@@ -0,0 +1,81 @@
...
+// Ex: the j'th input of the i'th data of group 'value' can be accessed by "/value/i/j"
+class HDF5Importer final
+{
+public:
+  explicit HDF5Importer(const std::string &path);
+
+public:
+  /**
+   * @brief importGroup has to be called before readTensor is called
+   *        Otherwise, readTensor will throw an exception
```
seanshpark, line 48: (optional) how about using `@note` as the comment seems to be a note. I'm not good at this but `@brief` was used as explaing something about the class/method/... itself.

compiler/dio-hdf5/include/dio_hdf5/HDF5Importer.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+#ifndef __DIO_HDF5_H__
+#define __DIO_HDF5_H__
+
+#include <H5Cpp.h>
+
+#include <loco.h>
+
+#include <stdexcept>
```
seanshpark, line 24: (optional) including string, vector would be nice (I think loco includes these so there is no error)

compiler/dio-hdf5/src/HDF5Importer.cpp
seanshpark: (optional) how about adding a comment about `h5_type is not an enum so we cannot use switch-case` ?

seanshpark (approved):
LGTM

compiler/dio-hdf5/src/HDF5Importer.cpp
jinevening: You're not wrong. Making a wrapper for bool would be better. I've used u8 for simplicity (bool values are saved as an integer in hdf5 files).

I think that is out of the scope of this PR (most code in HDF5Importer.cpp are copied from record-minmax).

compiler/dio-hdf5/CMakeLists.txt
jinevening: It is a required keword. We cannot omit it.

hseok-oh:
@nnfw-bot test nncc-debug
@nnfw-bot test nncc-release

seanshpark (approved):
LGTM thank you!

hseok-oh:
@nnfw-bot test nncc-debug
@nnfw-bot test nncc-release

mhs4670go (approved):
LGTM

PR #8741, opened 2022-03-29
DRAFT: Introduce circle-eval-diff

jinevening:
On going draft to introduce circle-eval-diff.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/8634

Issue #8634, opened 2022-03-15
[circle-eval-diff] Introduce a module to compare results of inference


jinevening:
**Update: I've renamed `luci-eval-diff` to `circle-eval-diff`.**

## What

Let's introduce a new module (`circle-eval-diff`) that compares inference results of two circle models.

![image](https://user-images.githubusercontent.com/5449554/158323377-65de12c9-a470-4b2a-bfd1-156729114e90.png)

`circle-eval-diff` will be useful to evaluate the accuracy of quantized model. We may use the module as follows.

```
./circle-eval-diff --first_model <circle> --second_model <circle> --first_input_data <h5/directory> --second_input_data <h5/directory> --metric <metric>
```

NOTES
- It would be possible to implement circle-eval-diff in different ways (C++ or Python)
- Input models can have different dtypes (e.g., A is fp32 and B is uint8). Comparison code should recognize the dtype of output.
- Possible metrics would be MAE, MSE, PEIR ..

CC @Samsung/one_quantizer

jinevening:
Format of `input_data` would be `h5` and `directory`, which is the same with that of `record-minmax`.

seanshpark:
Q) can we use random data if `<input_data>` is not given?

jinevening:
> Q) can we use random data if <input_data> is not given?

I think that would be a convenient option (although not very accurate way to evaluate accuracy).

jinevening:
I've renamed the tool to `circle-eval-diff`.

jinevening:
HDF5 format of input_data

```
/first
/first/<data_num>/<input_num>
/second
/second/<data_num>/<input_num>
```

jinevening:
Now draft has merged.

## Test results

Same model, same data (MAE = 0)
```
$ build/debug/compiler/circle-eval-diff/circle-eval-diff --first_model /home/hjj/TRIV2_Model_Zoo/VD-MODIFIED/AI_ZOOM/centernet_1216/centernet_1216.circle --second_model /home/hjj/TRIV2_Model_Zoo/VD-MODIFIED/AI_ZOOM/centernet_1216/centernet_1216.circle --first_input_data /home/hjj/TRIV2_Data_Zoo/VD/centernet_1216.input.1.h5 --second_input_data /home/hjj/TRIV2_Data_Zoo/VD/centernet_1216.input.1.h5
Evaluating 0'th data
Evaluation finished. Number of data: 1
Mean Absolute Error (MAE)
[...]

PR #8742, opened 2022-03-29
[circle-eval-diff] Introduce circle-eval-diff

jinevening:
This first introduces circle-eval-diff module.

---
Related to: #8634
Draft PR: https://github.com/Samsung/ONE/pull/8741

### DIFF
--- a/compiler/circle-eval-diff/README.md
+++ b/compiler/circle-eval-diff/README.md
@@ -2,6 +2,11 @@
 
 _circle-eval-diff_ compares inference results of two circle models.
 
+## Use cases
+
+1. _circle-eval-diff_ can be used to evaluate reconstruction errors of quantized models.
+2. _circle-eval-diff_ can be used to verify optimization (or any kind of value-preserving conversion) is safe.
+
 ## Usage
 
 Run circle-eval-diff with the following arguments.
--- a/compiler/circle-eval-diff/README.md
+++ b/compiler/circle-eval-diff/README.md
@@ -1,0 +1,41 @@
+# circle-eval-diff
+
+_circle-eval-diff_ compares inference results of two circle models.
+
+## Usage
+
+Run circle-eval-diff with the following arguments.
+
+--first_input_model: first model to compare (.circle).
+--second_input_model: second model to compare (.circle).
+--first_input_data: input data for the first model (.h5, directory). Random data will be used if this argument is not given.
+--second_input_data: input data for the second model (.h5, directory). Random data will be used if this argument is not given.
+--input_data_format: input data format (h5 (default), directory).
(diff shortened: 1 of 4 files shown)


compiler/circle-eval-diff/README.md
```diff
@@ -0,0 +1,41 @@
+# circle-eval-diff
+
+_circle-eval-diff_ compares inference results of two circle models.
+
```
seanshpark, line 4: maybe add some typical example of this tool?
I assume this will be used before-after quantization of a cirle model which first and second model data types are different, thus input data files have different data types according.

seanshpark (approved):
LGTM thank you!

llFreetimell (approved):
LGTM

compiler/circle-eval-diff/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Usage
+
+Run circle-eval-diff with the following arguments.
+
+--first_input_model: first model to compare (.circle).
+--second_input_model: second model to compare (.circle).
+--first_input_data: input data for the first model (.h5, directory). Random data will be used if this argument is not given.
+--second_input_data: input data for the second model (.h5, directory). Random data will be used if this argument is not given.
+--input_data_format: input data format (h5 (default), directory).
+--metric: metric to compare inference results (MAE (default), etc).
```
mhs4670go, line 19: They needs extra new line because it prints in one line rather than multiple lines.

compiler/circle-eval-diff/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+```
+$ ./circle-eval-diff
+  --first_input_model <first_input_model>
+  --second_input_model <second_input_model>
+  --first_input_data <first_input_data>
+  --second_input_data <second_input_data>
```
mhs4670go, line 26: I thought first model and second one have same inputs. But, it seems not, doesn't it?

mhs4670go:
Thank you for introducing this module:) I'm working on `circle-opselector` and this module can be used for a test!

compiler/circle-eval-diff/README.md
jinevening: Ah, right. I forgot that. Thanks!

compiler/circle-eval-diff/README.md
jinevening: No, it doesn't. Models' input signature can be different (fp32, u8), so input_data can be different too.

Of course, you can give the same input_data if two models have the same input signature.

mhs4670go (approved):
LGTM

llFreetimell (approved):
LGTM

seanshpark (approved):

PR #8744, opened 2022-03-29
[record-minmax] Extract check code for input dimension.

jinevening:
This extracts check code for input dimension.

### DIFF
--- a/compiler/record-minmax/src/RecordMinMax.cpp
+++ b/compiler/record-minmax/src/RecordMinMax.cpp
@@ -39,6 +39,35 @@ using DataType = luci_interpreter::DataType;
 namespace
 {
 
+uint32_t numElements(const luci::CircleNode *node)
+{
+  uint32_t num_elements = 1;
+  for (uint32_t i = 0; i < node->rank(); i++)
+    num_elements *= node->dim(i).value();
+
+  assert(num_elements > 0); // FIX_CALLER_UNLESS
+
+  return num_elements;
+}
+
+// Throw exception if input has one of the following conditions.
+// 1. Have unknown dimension
+// 2. Number of elements is 0
+void checkInputDimension(const luci::CircleInput *input)
+{
+  uint32_t num_elements = 1;
+  for (uint32_t i = 0; i < input->rank(); i++)
+  {
+    if (!input->dim(i).known())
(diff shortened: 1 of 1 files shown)


jinevening:
This PR is for using `checkInputDimension` in other functions.

seanshpark (approved):
LGTM

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -39,6 +39,35 @@ using DataType = luci_interpreter::DataType;
...
+  for (uint32_t i = 0; i < input->rank(); i++)
+  {
+    if (!input->dim(i).known())
+      throw std::runtime_error(input->name() + " has unknown dimension");
+
+    num_elements *= input->dim(i).value();
+  }
+
+  if (num_elements == 0)
+    throw std::runtime_error(input->name() + " is a zero-sized input");
```
llFreetimell, line 68: I think calculating the number of elements is duplicated :)
How about following?

```suggestion
  for (uint32_t i = 0; i < input->rank(); i++)
    if (!input->dim(i).known())
      throw std::runtime_error(input->name() + " has unknown dimension");

  if (numElements(input) == 0)
    throw std::runtime_error(input->name() + " is a zero-sized input");
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -39,6 +39,35 @@ using DataType = luci_interpreter::DataType;
 namespace
 {
 
+uint32_t numElements(const luci::CircleNode *node)
+{
+  uint32_t num_elements = 1;
+  for (uint32_t i = 0; i < node->rank(); i++)
+    num_elements *= node->dim(i).value();
+
+  assert(num_elements > 0); // FIX_CALLER_UNLESS
```
llFreetimell, line 48: I think removing this assert would be better.
For example, if we want to check additional validation for zero-sized tensor, `numElements` should return 0.
However, with this assert, those validation could not be done :)

llFreetimell: And meaning of `numElements` if "the number of elements", `numElements(zero-sized node)` should return 0 I think :D

llFreetimell (approved):
LGTM

seanshpark (approved):

PR #8746, opened 2022-03-29
[luci] Allow more dtypes in Square.

jinevening:
This allows more dtypes in Square.

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleSquare.cpp
+++ b/compiler/luci/import/src/Nodes/CircleSquare.cpp
@@ -29,14 +29,13 @@ bool CircleSquareGraphBuilder::validate(const ValidateArgs &args) const
     return false;
 
   const auto &inputs = args.op.inputs;
-  // Must be one of the following types
-  // bfloat16, half (float16), float32, float64, complex64, complex128
-  // Currently, circle supports float16, float32, complex64
   const auto tensors = args.reader.tensors();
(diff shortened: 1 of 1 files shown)


jinevening:
This allows to import Square with u8/s16 type, which is our quantization scheme.

compiler/luci/import/src/Nodes/CircleSquare.cpp
```diff
@@ -29,14 +29,13 @@ bool CircleSquareGraphBuilder::validate(const ValidateArgs &args) const
     return false;
 
   const auto &inputs = args.op.inputs;
-  // Must be one of the following types
-  // bfloat16, half (float16), float32, float64, complex64, complex128
-  // Currently, circle supports float16, float32, complex64
```
seanshpark, line 34: Q) why remove this comment?

jinevening: Because this PR makes the code do not match with the comment.

The comment does not match with the behavior of tf operator as well.

seanshpark: Thanks for the reply :) 
Current comment has changed to `A Tensor. Must be one of the following types: bfloat16, half, float32, float64, int8, int16, int32, int64, uint8, uint16, uint32, uint64, complex64, complex128`. As like other ops, agree remove this comment.

seanshpark (approved):
LGTM

PR #8749, opened 2022-03-29
[circle-eval-diff] Introduce Tensor data structure

jinevening:
This introduces data structure for input/output tensor.

---
Related to: https://github.com/Samsung/ONE/issues/8634
Draft PR: https://github.com/Samsung/ONE/pull/8741

### DIFF
--- a/compiler/circle-eval-diff/CMakeLists.txt
+++ b/compiler/circle-eval-diff/CMakeLists.txt
@@ -16,6 +16,8 @@ if(NOT ENABLE_TEST)
   return()
 endif(NOT ENABLE_TEST)
 
+# circle-eval-diff is executable, so we do not link it to the test.
+# Instead, we use TEST_SOURCES to specify sources uesd for tests.
 set(TEST_SOURCES
     "src/Tensor.cpp")
 
--- a/compiler/circle-eval-diff/CMakeLists.txt
+++ b/compiler/circle-eval-diff/CMakeLists.txt
@@ -3,8 +3,6 @@ set(DRIVER "driver/Driver.cpp")
 file(GLOB_RECURSE SOURCES "src/*.cpp")
 file(GLOB_RECURSE TESTS "src/*.test.cpp")
 list(REMOVE_ITEM SOURCES ${TESTS})
-set(TEST_SOURCES
-    "src/Tensor.cpp")
 
 add_executable(circle-eval-diff ${DRIVER} ${SOURCES})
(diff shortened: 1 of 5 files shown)


compiler/circle-eval-diff/CMakeLists.txt
```diff
@@ -1,10 +1,25 @@
 set(DRIVER "driver/Driver.cpp")
 
 file(GLOB_RECURSE SOURCES "src/*.cpp")
+file(GLOB_RECURSE TESTS "src/*.test.cpp")
+list(REMOVE_ITEM SOURCES ${TESTS})
+set(TEST_SOURCES
```
seanshpark, line 6: this line can be moved after `ndif(NOT ENABLE_TEST)` line.

compiler/circle-eval-diff/CMakeLists.txt
```diff
@@ -1,10 +1,25 @@
...
 target_link_libraries(circle-eval-diff safemain)
+target_link_libraries(circle-eval-diff loco)
 target_link_libraries(circle-eval-diff vconone)
+
+if(NOT ENABLE_TEST)
+  return()
+endif(NOT ENABLE_TEST)
+
+nnas_find_package(GTest REQUIRED)
+GTest_AddTest(circle_eval_diff_test ${TESTS} ${TEST_SOURCES})
```
seanshpark, line 22: Aboud adding `${TEST_SOURCES}`
- `circle-eval-diff` is executable so adding link to this exec looks strange
- so adding the source through `${TEST_SOURCES}` looks unavoidable.

Bu we can think this as optional;
- create `add_library(eval_diff_helper STATIC ${SOURCES})`
- add `target_link_libraries(circle-eval-diff eval_diff_helper)`
- and add `target_link_libraries(circle_eval_diff_test eval_diff_helper)` too

jinevening: I've thought about that option, but I decided not to choose it.

`{SOURCES}` have all cpp files, which may require arser, safemain, vconone, which are not used by `circle_eval_diff_test`. To remove those dependencies, we'll have to select source files, similarly to the current way.

seanshpark: Ah, OK then can you add some comments of this background facts?

seanshpark (approved):
LGTM thank you!

llFreetimell (approved):
LGTM

PR #8752, opened 2022-03-29
[infra] Additional option for ARM32 makefile

seanshpark:
This will revise Makefile.arm32 to accept additional
NNCC_CFG_OPTION_EXTRA option from caller.


hyunsik-yoon (approved):

infra/nncc/Makefile.arm32
```diff
@@ -9,6 +9,7 @@ CURRENT_DIR=$(shell pwd)
 BUILDFOLDER=build
 ARM32_FOLDER=arm32
 ROOTFS_ARM?=$(CURRENT_DIR)/tools/cross/rootfs/arm
+NNCC_CFG_OPTION_ADD?=
```
chunseoklee, line 12: I am not sure the usage of the varible, but how about name NNCC_CFG_OPTION_EXTRA or something if this is for a sort of extra option ?

seanshpark: > I am not sure the usage of the varible,

This is to provide `additional` option from the user(caller) of this Makefile.arm32.

like
```
NNCC_CFG_OPTION_ADD=-DCMAKE_DO_THIS=THAT make -f infra/nncc/Makefile.arm32 all
```

`NNCC_CFG_OPTION_EXTRA` also looks OK. If @hyunsik-yoon also agrees I'll change it.

chunseoklee: @hyunsik-yoon PTAL

hyunsik-yoon: `NNCC_CFG_OPTION_EXTRA` seems better to me. I thought that `ADD` meant adding some more cfg options from users. 😹

seanshpark: > adding some more cfg options from users

Yes, it's adding more cfg options... anyway I'll revise this :)

chunseoklee (approved):

hyunsik-yoon (approved):
LGTM

PR #8766, opened 2022-03-31
[infra/nncc] Enable NEON with option

seanshpark:
This will enable NEON compile option with BUILD_ARM32_NEON option.


seanshpark:
For #8379 from draft #8578

chunseoklee (approved):

PR #8774, opened 2022-04-01
DRAFT Compiler FE use python3.8 + TF2.8.0

seanshpark:
on-going draft to upgrade to use python3.8 + TF2.8.0


seanshpark:
done

Issue #8772, opened 2022-04-01
Compiler FE: deprecate common_artifacts venv_1_13_2


seanshpark:
As we can use `venv_2_6_0` for existing modules using `venv_1_13_2` we can upgrade and deprecate `venv_1_13_2`

seanshpark:
From #8767

seanshpark:
@mhs4670go , any comments?

mhs4670go:
@seanshpark I'm not sure but I removed `venv_1_13_2` and tested it. It passed successfully. I think it is okay to do that.

seanshpark:
done

PR #8779, opened 2022-04-01
[tf2tfliteV2] Revise to use venv_2_6_0

seanshpark:
This will revise test environment to use venv with TF2.6.0.


seanshpark:
For #8772, from draft #8774

mhs4670go (approved):
LGTM

PR #8785, opened 2022-04-04
[common-artifacts] Remove venv_1_13_2

seanshpark:
This will remove not used anymore venv_1_13_2.


seanshpark:
For #8772, from draft #8774

mhs4670go (approved):
LGTM

PR #8787, opened 2022-04-04
[common-artifacts] Remove version for pip, setuptools

seanshpark:
This will remove to install specific pip, setuptools as the issue is not
reproduced anymore.


seanshpark:
from draft #8774

mhs4670go (approved):
LGTM

Issue #8732, opened 2022-03-24
Compiler FE: Upgrade packages


seanshpark:
To use latest packages with model import, we have to upgrade our execution environments
- [x] python 3.8 (or higher for using Tensorflow 2.7.0 or higher) for Ubuntu 18.04

And the packages
- [x] TensorFlow 2.8.0
- [x] ONNX-TF 1.10.0
- [x] ONNX 1.11.0
- [x] ONNERUNTIME 1.11.0

And some module(s)
- [x] introduce `mio-tflite280` and revise to use this

seanshpark:
From https://github.com/Samsung/ONE/issues/7942#issuecomment-1056682307

seanshpark:
Current onnx-tensorflow version of 1.10.0
```
tensorflow==2.8.0
onnx==1.10.2
```

seanshpark:
## Upgrade to use python 3.8

### For Ubuntu 18.04

- explicit install of python3.8 is required
- need to use `python3.8`
- `common-artifacts` venv preparation needs update
- `infra/docker/bionic/Dockerfile` needs update
- CI needs update with Docker image
- ...

seanshpark:
local PRESET build test fails a lot -_-;

```
one-build_003.test FAILED
one-build_006.test FAILED
one-build_007.test FAILED
one-build_008.test FAILED
one-build_009.test FAILED
one-build_011.test FAILED
one-build_012.test FAILED
one-build_013.test FAILED
one-build_neg_006.test FAILED
one-import-onnx_001.test FAILED
one-import_005.test FAILED
one-import_006.test FAILED
one-quantize_001.test FAILED
one-quantize_003.test FAILED
one-quantize_004.test FAILED
one-quantize_neg_004.test FAILED
one-quantize_neg_009.test FAILED
one-quantize_neg_010.test FAILED
one-quantize_neg_011.test FAILED
one-quantize_neg_012.test FAILED
one-quantize_neg_014.test FAILED
onecc_003.test FAILED
onecc_006.test FAILED
onecc_007.test FAILED
onecc_008.test FAILED
onecc_009.test FAILED
onecc_011.test FAILED
onecc_012.test FAILED
onecc_016.test FAILED
onecc_018.test FAILED
onecc_019.test FAILED
onecc_neg_006.test FAILED
rawdata2hdf5_001.test FAILED
33 TESTS FAILED
```

seanshpark:
```
[...]

PR #8796, opened 2022-04-04
[common-artifacts] Use python 3.8

seanshpark:
This revise to explictly use python3.8
- this is a preparation to use TensorFlow 2.8.0

### DIFF
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -1,14 +1,14 @@
 #[[ Generate common python virtual enviornment ]]
-find_package(PythonInterp 3 QUIET)
-find_package(PythonLibs 3 QUIET)
+find_package(PythonInterp 3.8 QUIET)
+find_package(PythonLibs 3.8 QUIET)
 
 if(NOT ${PYTHONINTERP_FOUND})
   message(STATUS "Build common-artifacts: FALSE (Python3 is missing)")
   return()
 endif()
 
-if(${PYTHON_VERSION_MINOR} LESS 3)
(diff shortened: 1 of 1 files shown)


seanshpark:
For #8732, from draft #8774

seanshpark:
CC @hseok-oh

compiler/common-artifacts/CMakeLists.txt
```diff
@@ -1,14 +1,14 @@
 #[[ Generate common python virtual enviornment ]]
-find_package(PythonInterp 3 QUIET)
-find_package(PythonLibs 3 QUIET)
+find_package(PythonInterp 3.8 QUIET)
+find_package(PythonLibs 3.8 QUIET)
```
seanshpark, line 3: @mhs4670go , `PythonLibs` seems not used, can we remove this? (I'll remove in another PR if so) If used, could you please point me where?

mhs4670go: @seanshpark As [document](https://cmake.org/cmake/help/v3.10/module/FindPythonLibs.html) said, `PythonLibs` finds python libraries. I don't know what the libraries exactly are but I think it is `libpython*.so` things. Maybe default python libraries. I needs some test if whether it is really being used by removing above line and building it. Please let me take some time.

FYI, `PythonLibs` has been [deprecated](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html) since cmake 3.12. We'll need to update it later.

seanshpark: OK, let's go with this for a while :)

mhs4670go (approved):
LGTM

Issue #8788, opened 2022-04-04
[onert] Backend for Trix 


chunseoklee:
Let's implement trix backend on onert.

Here is a example PR when we introduce ruy backend : https://github.com/Samsung/ONE/pull/5067

chunseoklee:
cc @glistening, @chunseoklee, @dayo09

jyoungyun:
### References

* [Backend API guideline](https://github.com/Samsung/ONE/blob/master/docs/runtime/backend-api.md)

### ToDo
[x] Introduce TRIX Engine(#8788)
[ ] Add "Trix" directory for trix backend
...
???

chunseoklee:
Closed by done

PR #8805, opened 2022-04-05
[Draft] Add TRIX backend

jyoungyun:
This commit adds TRIX backend for ONERT.

/cc @hseok-oh @dayo09 @chunseoklee


runtime/onert/backend/trix/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+set(LIB_ONERT_BACKEND_TRIX onert_backend_trix)
+
+nnfw_find_package(TRIXEngine EXACT 2.5.0 REQUIRED)
```
hseok-oh, line 3:
```suggestion
nnfw_find_package(TRIXEngine EXACT 2.5.0 QUIET)
if(NOT TRIXEngine_FOUND)
  return()
endif(NOT TRIXEngine_FOUND)
```

hseok-oh: It will resolve build fail on ubuntu.

runtime/onert/backend/trix/Backend.h
```diff
@@ -0,0 +1,63 @@
...
+class Backend : public ::onert::backend::Backend
+{
+public:
+  Backend() : _config{std::make_shared<Config>()} {}
+
+  std::shared_ptr<IConfig> config() const override { return _config; }
+
+  std::unique_ptr<onert::backend::BackendContext> newContext(ContextData &&data) const override
+  {
+    auto custom_kernel_builder = data.custom_kernel_builder;
```
ragmani, line 44: This is to support the custom operation, and the custom operation is supported by the cpu backend.  So this line is not needed in this backend.

runtime/onert/backend/trix/Config.cc
```diff
@@ -0,0 +1,32 @@
...
+namespace onert
+{
+namespace backend
+{
+namespace trix
+{
+
+bool Config::initialize() { return true; }
+
+ir::Layout Config::supportLayout(const ir::Operation &, ir::Layout) { return ir::Layout::NHWC; }
```
chunseoklee, line 28: @hseok-oh I heard that there are several layouts on trix, which is I am not familiar with.

runtime/onert/backend/trix/Config.h
```diff
@@ -0,0 +1,48 @@
...
+{
+
+class Config : public IConfig
+{
+public:
+  std::string id() override { return "trix"; }
+  bool initialize() override;
+  ir::Layout supportLayout(const ir::Operation &node, ir::Layout frontend_layout) override;
+  bool supportPermutation() override { return true; }
+  bool supportDynamicTensor() override { return true; }
```
ragmani, line 38:
```suggestion
  bool supportDynamicTensor() override { return false; }
```

ragmani (commented):
You do not need to worry about dynamic tensor(shape) when adding trix backend because trix backend does not need to support dynamic shape.

runtime/onert/backend/trix/KernelGenerator.cc
```diff
@@ -0,0 +1,88 @@
...
+      assert(portable_tensor->layout() == ir::Layout::NHWC);
+    }
+
+    auto tensor = _tensor_reg->getNativeTensor(ind);
+    if (tensor)
+    {
+      tensor->increase_ref();
+    }
+  }
+  return ret;
```
ragmani, line 83:
```suggestion
  auto ret = std::make_unique<exec::FunctionSequence>();
  ret->enableDynamicShapeInferer(false);

  const auto &op = _graph.operations().at(ind);
  op.accept(*this);
  ret->append(releaseFunction());
  return ret;
```

runtime/onert/backend/trix/KernelGenerator.cc
```diff
@@ -53,6 +56,29 @@ std::unique_ptr<exec::FunctionSequence> KernelGenerator::generate(ir::OperationI
...
+  for (auto *ofm_idx : node.getOutputs())
+    output_tensors.emplace_back(_tensor_reg->getPortableTensor(ofm_index));
+
+  std::vector<const IPortableTensor *> input_tensors;
+  for (auto *ifm_idx : node.getInputs())
+    input_tensors.emplace_back(_tensor_reg->getPortableTensor(ifm_idx));
+
+  // parameters
+  const auto binary_path = node.param().binary_path;
+  const auto info_path = node.param().info_path;
```
chunseoklee, line 73: info_path is removed on final PR( #8816 )

runtime/onert/backend/trix/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+  return()
+endif(NOT TRIXEngine_FOUND)
+
+file(GLOB_RECURSE SOURCES "*.cc")
+
+add_library(${LIB_ONERT_BACKEND_TRIX} SHARED ${SOURCES})
+
+target_link_libraries(${LIB_ONERT_BACKEND_TRIX} PRIVATE onert_core)
+target_link_libraries(${LIB_ONERT_BACKEND_TRIX} PRIVATE nnfw_common)
+target_link_libraries(${LIB_ONERT_BACKEND_TRIX} PRIVATE nnfw_coverage)
```
hseok-oh, line 14:
```suggestion
target_link_libraries(${LIB_ONERT_BACKEND_TRIX} PRIVATE onert_core)
target_link_libraries(${LIB_ONERT_BACKEND_TRIX} PRIVATE npu_engine)
target_link_libraries(${LIB_ONERT_BACKEND_TRIX} PRIVATE nnfw_common)
target_link_libraries(${LIB_ONERT_BACKEND_TRIX} PRIVATE nnfw_coverage)
```

We need to rename `npu_engine` to `trix_engine` (defined in `infra/nnfw/cmake/packages/TRIXEngineConfig.cmake`)

runtime/onert/backend/trix/KernelGenerator.cc
jyoungyun: I will rebase again.

runtime/onert/backend/trix/DevContext.h
```diff
@@ -0,0 +1,125 @@
...
+{
+namespace trix
+{
+
+class DevContext
+{
+public:
+  DevContext()
+  {
+    auto device_count = getNPUdeviceByType(NPUCOND_TRIV2_CONN_SOCIP);
```
chunseoklee, line 34: maybe getnumNPUdeviceByType ?

runtime/onert/backend/trix/DevContext.h
```diff
@@ -0,0 +1,125 @@
...
+      case ir::DataType::INT64:
+        return DATA_TYPE_INT64;
+      default:
+        throw std::runtime_error("Not support data type");
+    }
+  }
+
+private:
+  // NPU device handle
+  npudev_h _dev_handle;
```
chunseoklee, line 118: @jyoungyun AFAIK, npudev_h correspond to one core for TRIX, right ? 

(just for note : I hardly imagine how to handle "multicore" after M1 :) )

runtime/onert/backend/trix/CMakeLists.txt
chunseoklee: IN my local gbs, this does not resolve the issue "libnpuhost.h not found". I have resolved it by `+target_include_directories(${LIB_ONERT_BACKEND_TRIX} PRIVATE ${TRIXEngine_INCLUDE_DIR})` :(

runtime/onert/backend/trix/DevContext.h
glistening: @jyoungyun Please use `getnumNPUdeviceByType` as @chunseoklee mentioned. : )

jyoungyun: > maybe getnumNPUdeviceByType ?
> @jyoungyun Please use getnumNPUdeviceByType as @chunseoklee mentioned. : )

Yes, that's what I was trying to exactly. Thank you!

runtime/onert/backend/trix/DevContext.h
jyoungyun: Yes, right. This `_dev_handle` handles one core. So if we support multicore, we have to expand it. Do we need to consider multicore case now?

chunseoklee: Nope. I mean that I am afraid of the future :)

chunseoklee:
FYI: with https://github.com/chunseoklee/ONE-1/tree/pr8805 (commit: 64980cb3b), gbs build done.

runtime/onert/backend/trix/ops/BulkLayer.cc
```diff
@@ -0,0 +1,120 @@
...
+  tensors_data_info in_info;
+  tensors_data_info out_info;
+  _dev_context->setDataInfo<const IPortableTensor>(&in_info, _inputs);
+  _dev_context->setDataInfo<IPortableTensor>(&out_info, _outputs);
+
+  input_buffers input_buf;
+  output_buffers output_buf;
+  _dev_context->setBuffer<const IPortableTensor>(&input_buf, _inputs);
+  _dev_context->setBuffer<IPortableTensor>(&output_buf, _outputs);
+
```
chunseoklee, line 91: These could be moved to bulklayer::configure unless the address is not changed during execution ?

jyoungyun: As I know, the input and output memory are not allocated at the configure time, aren't it? And @ragmani says we need to consider overhead according to switch memory address when we do not use `IPortableTensor`.

ragmani: FYI, the overhead due to switching memory occurs when input/output of a whole model (all partitioned models) is used or generated in `trix` backend. If the overhead is not big we can move this function out of the function "bulklayer::run()". And I think it doesn't have to be at the time when `bulklayer::configure()` is called.

hseok-oh:
#8788 completed

PR #8808, opened 2022-04-06
[Draft] Reduce duplicated code

hseok-oh:
- Reduce duplicated code

TODO
- acl-cl / acl-neon / gpu-cl
- cpu / xnnpack / ruy

### DIFF
--- a/runtime/onert/core/src/compiler/Compiler.cc
+++ b/runtime/onert/core/src/compiler/Compiler.cc
@@ -86,6 +86,31 @@ void verboseOptions(compiler::CompilerOptions &options)
                     << std::noboolalpha;
 }
 
+void setBackendOptionPerOP(compiler::ManualSchedulerOptions &ms_options, const ir::Subgraphs &subgs,
+                           const std::string &str)
+{
+  // Index to Backend
+  // TODO Support multiple subgraphs for manual scheduling
+  auto key_val_list = nnfw::misc::split(str, ';');
+  for (const auto &key_val_str : key_val_list)
+  {
+    if (key_val_str.empty())
+    {
+      continue;
+    }
+
+    auto key_val = nnfw::misc::split(key_val_str, '=');
+    const auto &key_str = key_val.at(0);
+    const auto &val = key_val.at(1);
+    auto key = static_cast<uint32_t>(std::stoi(key_str));
+
+    subgs.at(ir::SubgraphIndex{0})
+      ->operations()
+      .at(ir::OperationIndex{key}); // Check if exist, or this wil throw
+    ms_options.index_to_backend.emplace(ir::OperationIndex{key}, val);
+  }
+}
+
 } // namespace
 
 namespace onert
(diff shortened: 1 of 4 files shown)


hseok-oh:
DC score on this PR: 3.96

hseok-oh:
DC score on this PR: 4.01
- Move xnnpack backend to contrib

chunseoklee:
> DC score on this PR: 4.01
> - Move xnnpack backend to contrib

It seems that DC requires build(I thought DC computation is done by text processing)

runtime/onert/backend/cl_common/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+file(GLOB SOURCES "*.cc")
+
+add_library(${LIB_ONERT_BACKEND_CL_COMMON} INTERFACE ${SOURCES})
```
ragmani, line 3: I think naming this as cl_common is strange since there is also `acl neon`.

ragmani (dismissed):
LGTM (but it's sad that we have to merge even these duplicated codes like this to increase dc score. )

hseok-oh:
DC score: 4.01 -> 4.02

hseok-oh:
Reduce ExecutorFactory duplication, Remove CalculateActivationRange duplication: 3.97 -> 3.98

hseok-oh:
ConstantInitializer: 3.98 -> 3.98

hseok-oh:
Many `KernelGenerator.cc`'s duplications come from basic variable definition.

ex) 2nd largest duplication:
`runtime/onert/backend/acl_cl/KernelGenerator.cc` (L1256-1309)
`runtime/onert/backend/acl_cl/KernelGenerator.cc` (L535-589)

```cpp
    input_tensor->handle(), k, values_tensor->handle(), indices_tensor->handle());

  _return_fn = asAclFunction(std::move(fn));
}

void KernelGenerator::visit(const ir::operation::Gather &node)
{
  const auto ofm_index{node.getOutputs().at(0)};

  const auto ifm_index{node.getInputs().at(ir::operation::Gather::Input::INPUT)};
  const auto indices_index{node.getInputs().at(ir::operation::Gather::Input::INDICES)};

  const auto ifm_rank = _ctx.at(ifm_index).shape().rank();
  const auto axis_raw = node.param().axis;
  const auto axis_value = (axis_raw < 0 ? (ifm_rank + axis_raw) : axis_raw);
  const int axis = ::onert::backend::acl_common::ToARMComputeAxis(ifm_rank, axis_value).value();

  auto ofm_tensor = _tensor_reg->getAclTensor(ofm_index);
  auto ifm_tensor = _tensor_reg->getAclTensor(ifm_index);
  auto indices_tensor = _tensor_reg->getAclTensor(indices_index);

  // NOTE The frontend layout and backend layout must be the same for this operation.
  //      If not the same, we have to add a stage(?) to perform permutation of output tensor. It
  //      is not not efficient even if it works well. If so, it would be better to set the
  //      layout of these backend tensors to the same layout.
  //      There is also one thing we have to think about. This operation depends on the layout of
  //      a model. For example, if a model in NHWC has this operation as output rank == 4, indices
  //      rank == 2 and axis == 2, this operation should work as the axis W and C, but the axis W
  //      and C are not sequential in NCHW. So the backend in NCHW cannot handle this case.
  const auto backend_layout = ofm_tensor->layout();
  UNUSED_RELEASE(backend_layout);
  assert(backend_layout == ifm_tensor->layout());
  assert(backend_layout == indices_tensor->layout());
  assert(ifm_rank < 4 || _current_layout == backend_layout);

  // input is n-D, indices k-D, output is (n + k - 1)-D
  size_t n = ifm_rank;
  assert(n == ifm_tensor->num_dimensions());
  size_t k = _ctx.at(indices_index).shape().rank();
  assert(k == indices_tensor->num_dimensions());

  // Disable applied dim_correction
  if (n != ifm_tensor->info()->num_dimensions())
  {
    // This means that high dimension's value is 1 and ifm tensor is applied dim_correction
    acl_common::disableDimCorrection(ifm_tensor);
  }
  if (k != indices_tensor->info()->num_dimensions())
  {
    // This means that high dimension's value is 1 and indices tensor is applied dim_correction
    acl_common::disableDimCorrection(indices_tensor);
  }

  auto fn = acl_common::generateLayer<arm_compute::CLGatherEx>(
```

To reduce these duplication, we need to change various backend class definition by introducing common interface between similar backend (ex. `acl-cl` and `acl-neon` / `cpu` and `ruy`). Temporary solution is to using template (like `runtime/onert/backend/acl_common/AclKernelGen.h`), but it may make hard to reading code.

hseok-oh:
Exclude `KernelGenerator.cc`: 3.98 -> 4.24 (refactoring item: 1051 -> 674)

chunseoklee:
> Exclude `KernelGenerator.cc`: 3.98 -> 4.24 (refactoring item: 1051 -> 674)

This is just to check how much KernelGenerator.cc affects DC score , right ?

hseok-oh:
>This is just to check how much KernelGenerator.cc affects DC score , right ?

Right.

hseok-oh:
DC score: 4.03

Issue #8762, opened 2022-03-31
[dio-hdf5/arm32] Unit test fails


seanshpark:
Unit test fails inside RPi3/ARM32 for _dio-hdf5_

```
(build/arm32.debug/compiler/dio-hdf5)

./dio_hdf5_test
HDF5-DIAG: Error detected in HDF5 (1.8.16) thread 0:
  #000: /../externals/HDF5/src/H5T.c line 4508 in H5T_path_find(): unable to initialize conversion function
    major: Datatype
    minor: Unable to initialize object
...
```

I'm not sure current cross build of HDF5 is correct or not; as I don't think any execution of using HDF5 codes was tested previously in ARM32 device.

seanshpark:
ARM32 native build/test is OK, which should be.

seanshpark:
Searching...
- error shown when building universal in macos.

seanshpark:
fixed

PR #8810, opened 2022-04-06
[infra] Fix HDF5 ARM32 cross build

seanshpark:
This will fix HDF5 ARM32 cross build.
- HDTinit.c is added from ARM32 native build
- which cannot be generated from cross build host


seanshpark:
For #8762

glistening (approved):
LGTM :+1:  ( I had to handle similar problems during cross-compiling hdf5 for android. )

chunseoklee (approved):

PR #8816, opened 2022-04-07
Add 'Bulk' operation for pre-built(e.g. npu) binary

chunseoklee:
- Bulk operation load and execute binary object on special devices


hseok-oh:
@nnfw-bot test all

chunseoklee:
cc @jyoungyun

hseok-oh (approved):
LGTM

runtime/onert/core/include/ir/operation/Bulk.h
```diff
@@ -0,0 +1,54 @@
...
+namespace operation
+{
+
+class Bulk : public Operation
+{
+
+  struct Param
+  {
+    std::string binary_path;
+    std::string info_path;
```
glistening, line 35: Does It imply that 

1) information about prebuilt will come from nnpkg `configs` ?
2) validation will not be done in loader, but in backend ?

jyoungyun: (Q) What does the `info_path` mean?

jyoungyun (dismissed):
LGTM

runtime/onert/core/include/ir/operation/Bulk.h
chunseoklee: info_path is reserved for future use for now. It is not mandatory.

glistening: @chunseoklee

About my Q1: I understood. What about removing this at this moment and adding when it become necessary.

About my Q2: Not answered yet. I guess frontend does validation, backend will get every information by itself from `binary_path`. Is it right?

glistening:
@jyoungyun I think `info_path` means config path for prebuilt binary, which can be specified in nnpackage metadata `configs` field.

runtime/onert/core/include/ir/operation/Bulk.h
chunseoklee: > What about removing this at this moment and adding later.

OK.  i will update PR

> validation will not be done in loader, but in backend ?

I prefer loader. 

> From this PR, backend will get every information by itself from binary_path. Is it right?
For details, I failed to figure out from this PR. If it is not clear, it would be better not to merge this PR yet.

I do not get the point. Could you please describe more ?

hseok-oh:
I don't know exact usage of `info_path`, but when I saw draft, I thought it is optional field to path backend directly.

chunseoklee:
I have removed info_path from param, PTAL

hseok-oh (approved):

jyoungyun (approved):
LGTM

runtime/onert/core/src/ir/operation/Bulk.cc
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ir/operation/Bulk.h"
+
+#include <cassert>
```
glistening, line 19: I think this header is not necessary.

glistening (approved):
https://github.com/Samsung/ONE/pull/8816#discussion_r845701428 is optional. I am okay to let this PR merged.

runtime/onert/core/src/ir/operation/Bulk.cc
hseok-oh: There are similar code on many IRs. Let's remove at once later.

PR #8822, opened 2022-04-07
[pota-quan-val-test] Use venv_2_8_0

seanshpark:
This will revise to venv_2_8_0 having TensorFlow 2.8.0 and other latest packages.


seanshpark:
For #8732, from draft #8774

hseok-oh (approved):

PR #8837, opened 2022-04-11
DRAFT: introduce mio-tflite280

seanshpark:
on-going draft to introduce mio-tflite280.


seanshpark:
For  #8732

seanshpark:
`method_name` needs to be changed to `signature_key` where this exist in `res` and also in `compiler`
- any one of change will break the build
- we need to (1) disable the test (2) change (3) re-enable the test

seanshpark:
done

PR #8839, opened 2022-04-11
[debian] Add python3.8 dependency

mhs4670go:
This commit adds python3.8 dependency to debian packageing.


seanshpark (approved):
LGTM thank you!

Issue #8833, opened 2022-04-11
[mio-tflite280] For using TensorFlow 2.8.0


seanshpark:
Lets introduce mio-tflite280 using TensorFlow 2.8.0.

seanshpark:
From #8732

seanshpark:
compared to TF2.6.0, schema changes are about 6 more operators
```diff
975c993
< inline const BuiltinOperator (&EnumValuesBuiltinOperator())[145] {
---
> inline const BuiltinOperator (&EnumValuesBuiltinOperator())[151] {
1121c1139,1145
<     BuiltinOperator_ASSIGN_VARIABLE
---
>     BuiltinOperator_ASSIGN_VARIABLE,
>     BuiltinOperator_BROADCAST_ARGS,
>     BuiltinOperator_RANDOM_STANDARD_NORMAL,
>     BuiltinOperator_BUCKETIZE,
>     BuiltinOperator_RANDOM_UNIFORM,
>     BuiltinOperator_MULTINOMIAL,
>     BuiltinOperator_GELU

```

seanshpark:
done

PR #8842, opened 2022-04-11
[mio-tflite280] Introdice mio-tflite for TF2.8.0

seanshpark:
This will introduce mio-tflite280 to follow schema of TensorFlow 2.8.0.


seanshpark:
For #8833, from draft #8837

mhs4670go (approved):
LGTM

PR #8806, opened 2022-04-06
[common-artifacts] Revise exclude for ExpandDims

seanshpark:
This will revise exclude list for ExpandDims Op.

ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #8843, opened 2022-04-11
[luci-value-test] Enable ExpandDims

seanshpark:
This will enable value test for possible ExpandDims.


seanshpark:
To follow #8806

mhs4670go (approved):
LGTM

PR #8846, opened 2022-04-12
[common-artifacts] Enable tests

seanshpark:
This will enable tests(remove from exclude list) that are working as of now.


mhs4670go (approved):
LGTM

PR #8847, opened 2022-04-12
[luci-value-test] Enable tests

seanshpark:
This will enable tests that are working as of now.


mhs4670go (approved):
LGTM

PR #8849, opened 2022-04-12
[infra] Introduce cmake for TensorFlow2.8.0

seanshpark:
This will introduce two cmake for TensorFlow 2.8.0 source.


seanshpark:
For #8833, from draft #8837

seanshpark:
CC @mhs4670go , other PRs depend on this...

PR #8851, opened 2022-04-12
[infra] Revise preset 20220323

seanshpark:
This will revise preset 20220323 to include mio-tflite280.


seanshpark:
For #8833, from draft #8837

mhs4670go (approved):
LGTM

PR #8853, opened 2022-04-12
[tfldump] Revise to use mio-tflite280

seanshpark:
This will revise to use mio-tflite280.


compiler/tfldump/src/Dump.cpp
```diff
@@ -410,9 +410,8 @@ void dump_model(std::ostream &os, const tflite::Model *model)
     for (uint32_t i = 0; i < signaturedefs->Length(); ++i)
     {
       auto sign_i = signaturedefs->Get(i);
-      os << "S(" << i << ") method_name(" << sign_i->method_name()->c_str() << "), key("
-         << sign_i->key()->c_str() << "), sub_graph(" << sign_i->subgraph_index() << ")"
-         << std::endl;
+      os << "S(" << i << ") signature_key(" << sign_i->signature_key()->c_str() << "), sub_graph("
```
seanshpark, line 413: `key` is obsolete and `method_name` is renamed to `signature_key`

seanshpark:
For #8833, from draft #8837

mhs4670go (approved):
LGTM

PR #8857, opened 2022-04-12
[tflchef] Revise to use mio-tflite280

seanshpark:
This will revise to use mio-tflite280.


seanshpark:
may have to disable models in `res`...

mhs4670go (approved):
LGTM

Issue #8838, opened 2022-04-11
[luci-interpreter] critical error: Variable 'i - 1' is used as index


seanshpark:
Code: `void ExpandDims::configure()`

> Error: The index can have value -1 that is out of range, as indicated by preceding conditional expression at ExpandDims.cpp:68.

```cxx
output_shape.dim(i) = input_shape.dim(i - 1);
```

seanshpark:
This is from our internal code analysis tool.

seanshpark:
@BalyshevArtem PTAL

BalyshevArtem:
Actually, I don't see situations when it could happen. If `input_shape.dim(i - 1) ==  -1 `, then i = 0, then according to code
```c++
  for (int32_t i = 0; i < output_shape.num_dims(); ++i)
  {
    if (i < axis_value)
    {
      output_shape.dim(i) = input_shape.dim(i);
    }
    else if (i == axis_value)
    {
      output_shape.dim(i) = 1;
    }
    else
    {
      output_shape.dim(i) = input_shape.dim(i - 1);
    }
  }
```
`axis_value` should be < 0 and such situations is checked with:
```c++
LUCI_INTERPRETER_CHECK(axis_value <= input_shape.num_dims() and axis_value >= 0);
```  

But of course I could add additional check for this line `(i - 1) >= 0`, if necessary :)

seanshpark:
> But of course I could add additional check for this line (i - 1) >= 0, if necessary :)

Our analysis tool is not that intelligent but must fix this.

seanshpark:
As patch has landed, I'll check this when analysis server runs again.

seanshpark:
Issue is resolved :)

PR #8858, opened 2022-04-12
[luci-interpreter] Add additional checking for ExpandDims op

BalyshevArtem:
This commit adds additional checking in configure for ExpandDims op.

for issue: #8838


seanshpark (approved):
LGTM thank you!

PR #8859, opened 2022-04-12
[res] Temporay disable signature_def

seanshpark:
This will temporary disable signature_def for build break.


seanshpark:
This change will be rolled-back after `tflchef` is revised.

mhs4670go (approved):
LGTM

PR #8861, opened 2022-04-13
[res] Re-enable signature_def test

seanshpark:
This will re-enable signature_def test from tflite scehma v3a.


seanshpark:
This is to re-enable that was temporary disabled with #8859

mhs4670go (approved):
LGTM

PR #8870, opened 2022-04-13
DRAFT CFE reduce duplicate code

seanshpark:
on-going draft to reduce duplicate code score.

### DIFF
--- a/compiler/logo/src/Passes/SimplifyDomainConversionPass.cpp
+++ b/compiler/logo/src/Passes/SimplifyDomainConversionPass.cpp
@@ -112,6 +112,22 @@ void set_input_null(loco::Node *node)
 namespace logo
 {
 
+#define CHECK_GET_PERM(DOMAIN, encode_node, decode_node)                       \
+  auto encoder = encode_node->encoder();                                       \
+  assert(encoder != nullptr);                                                  \
+                                                                               \
+  auto decoder = decode_node->decoder();                                       \
+  assert(decoder != nullptr);                                                  \
+                                                                               \
+  /* NOTE Work only for permuting codec */                                     \
+  auto perm_decoder = dynamic_cast<const PermutingDecoder<DOMAIN> *>(decoder); \
+  auto perm_encoder = dynamic_cast<const PermutingEncoder<DOMAIN> *>(encoder); \
+                                                                               \
+  if (perm_encoder == nullptr || perm_decoder == nullptr)                      \
+  {                                                                            \
+    return;                                                                    \
+  }
+
 bool SimplifyDomainConversionPass::run(loco::Graph *g)
 {
   // TODO Introduce and Use "Pattern Match"
@@ -129,20 +145,7 @@ bool SimplifyDomainConversionPass::run(loco::Graph *g)
       }
       assert(decode_node->input() != nullptr);
 
-      auto encoder = encode_node->encoder();
-      assert(encoder != nullptr);
-
-      auto decoder = decode_node->decoder();
-      assert(decoder != nullptr);
-
-      // NOTE Work only for permuting codec
-      auto perm_decoder = dynamic_cast<const PermutingDecoder<Domain::Feature> *>(decoder);
-      auto perm_encoder = dynamic_cast<const PermutingEncoder<Domain::Feature> *>(encoder);
-
-      if (perm_encoder == nullptr || perm_decoder == nullptr)
-      {
-        return;
-      }
+      CHECK_GET_PERM(Domain::Feature, encode_node, decode_node);
 
       if (equal(perm_encoder->perm(), perm_decoder->perm()))
       {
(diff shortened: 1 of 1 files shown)


seanshpark:
Score before: 3.11

.ahub/sam/exclude.txt
```diff
@@ -5,6 +5,37 @@
 # Eigen
 /ONE/compiler/nnc/backends/soft_backend/code_snippets/eigen.def
 
+# Frontend obsolete
+/ONE/compiler/caffe2circle
+/ONE/compiler/mio-tflite260
+
+# Frontend test codes
```
llFreetimell, line 12: For test codes, #8881 might be better solution :)

seanshpark: Thanks, you gave me a good idea :)

seanshpark:
With `--notest`, score is now `3.49` (ID:1272941)

seanshpark:
draft score is 3.99 (ID:1274202), total number of BlockKey: 109

.ahub/sam/exclude.txt
```diff
@@ -5,6 +5,17 @@
 # Eigen
 /ONE/compiler/nnc/backends/soft_backend/code_snippets/eigen.def
 
+# Frontend tools for testing and debugging
```
llFreetimell, line 8: AFAIK, if these tools are not needed for release, they should be removed from preset, not here.

seanshpark: > they should be removed from preset, not here.

`preset` items are used also for coverage test which needs testing.

llFreetimell: What is the range of coverage test?

In case of static analysis, the range is "source codes which are needed to create release package".
If the range of coverage test is "all source codes in the repository", the script for coverage test should be changed IMHO.

seanshpark: > AFAIK, `preset` items are used also for coverage test which needs testing.

@hseok-oh hold me `coverage` list is in another file --> I need to revise this file... -_-;
From my long time ago memory, test modiules in `preset` list are added as some modules need them while build configure without checking target existance. we may need to revise them later if we are to remove them from `preset` list.

compiler/arser/include/arser/arser.h
```diff
@@ -737,6 +728,34 @@ template <typename T> T Arser::get(const std::string &arg_name)
...
+      .required(false)
+      .default_value(false)
+      .help("output additional information to stdout or stderr");
+  }
+
+  static void add_string(Arser &arser, const std::string &name, const std::string &help)
+  {
+    arser.add_argument(name).nargs(1).type(arser::DataType::STR).help(help);
+  }
+};
```
mhs4670go, line 757: I think `add_version` and `add_verbose` are very good methods to have. But `add_string` might a bit confuse a reader because of its name and `add_argument` is kinda intuitive expression to me. 

How about just omit `nargs` and `type` method? As of now, `nargs` has default value "1" and `type` can have `arser::DataType::STR` default value with a single PR. Then,
```cxx
arser.add_argument("--by_name")
    .nargs(1)
    .type(arser::DataType::STR)
    .help("Input operation name to select nodes.");
```
could be
```cxx
arser.add_argument("--by_name")
    .help("Input operation name to select nodes.");
```

seanshpark: > type can have arser::DataType::STR default value with a single PR.

Can you post this as a PR?

mhs4670go: @seanshpark Sure:)

.ahub/sam/exclude.txt
seanshpark: > What is the source code range of coverage test?

`infra/scripts/compiler_modules.sh` provides modules for release package + test modules for test these modules

seanshpark: I think some of these can be removed

seanshpark:
draft score is 3.62 (ID:1275814), total number of BlockKey: 543
- too much difference compared to ID:1274202, strange...
- ID:1274202 don't have any luci-interpreter items, which is wrong

seanshpark:
Adding interpreter kernel codes as it may hurt performance.
Adding loco IR header files
Adding luci import header files

seanshpark:
draft running (score is x.y) (ID:1276327), total number of BlockKey: z

seanshpark:
Added some common methods

seanshpark:
Some tools for `-dev`, `-test` packages was accidently remove from the list.
Will revise the list again.

seanshpark:
Master score is 3.92 (ID:1281296), total number of BlockKey: 348

seanshpark:
Score drop was caused by adding `tflchef`, `circledump` and `circle-tensordump`
which are included in `-test` and `-dev` release package.

PR #8873, opened 2022-04-13
[onert] Add BulkLayer on TRIX backend

jyoungyun:
This commit adds BulkLayer on TRIX backend.
It uses trix-engine library to handle TRIX device.

draft pr: #8805

### DIFF
--- a/runtime/onert/backend/trix/DevContext.h
+++ b/runtime/onert/backend/trix/DevContext.h
@@ -55,6 +55,54 @@ public:
 
   npudev_h getDev() { return _dev_handle; }
 
+  template <typename T> void setDataInfo(tensors_data_info *info, std::vector<T *> &tensors)
+  {
+    info->num_info = static_cast<uint32_t>(tensors.size());
+
+    for (uint32_t idx = 0; idx < info->num_info; ++idx)
+    {
+      info->info[idx].layout = convertDataLayout(tensors[idx]->layout());
+      info->info[idx].type = convertDataType(tensors[idx]->data_type());
+    }
+  }
+
+  template <typename T> void setBuffer(generic_buffers *buf, std::vector<T *> &tensors)
+  {
+    buf->num_buffers = static_cast<uint32_t>(tensors.size());
+
+    for (uint32_t idx = 0; idx < buf->num_buffers; ++idx)
+    {
+      buf->bufs[idx].addr = tensors[idx]->buffer();
+      buf->bufs[idx].size = static_cast<uint64_t>(tensors[idx]->total_size());
+      buf->bufs[idx].type = BUFFER_MAPPED;
+    }
+  }
+
+private:
+  data_layout convertDataLayout(const ir::Layout layout)
+  {
+    switch (layout)
+    {
+      case ir::Layout::NCHW:
+        return DATA_LAYOUT_NCHW;
+      case ir::Layout::NHWC:
+        return DATA_LAYOUT_NHWC;
+      default:
+        throw std::runtime_error("Unknown Layout");
+    }
+  }
+
+  data_type convertDataType(const ir::DataType type)
+  {
+    switch (type)
+    {
+      case ir::DataType::QUANT_UINT8_ASYMM:
+        return DATA_TYPE_QASYMM8;
+      default:
+        throw std::runtime_error("Unsupported data type");
+    }
+  }
+
 private:
   // NPU device handle
   // TODO Support multicore npu device
--- a/runtime/onert/backend/trix/KernelGenerator.cc
+++ b/runtime/onert/backend/trix/KernelGenerator.cc
@@ -16,6 +16,8 @@
 
 #include "KernelGenerator.h"
 
+#include "ops/BulkLayer.h"
+
 #include <backend/Backend.h>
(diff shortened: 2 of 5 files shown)


runtime/onert/backend/trix/DevContext.h
```diff
@@ -55,6 +55,64 @@ class DevContext
...
+        return DATA_LAYOUT_NHWC;
+      default:
+        throw std::runtime_error("Unknown Layout");
+    }
+  }
+
+  data_type convertDataType(const ir::DataType type)
+  {
+    switch (type)
+    {
```
jyoungyun, line 98: I need to consider another case more later(?).

jyoungyun: Is it okay? :) /cc @hseok-oh

hseok-oh: I prefer to support type conversion which can test, and remain NYI others.

runtime/onert/backend/trix/DevContext.h
```diff
@@ -55,6 +55,64 @@ class DevContext
 
   npudev_h getDev() { return _dev_handle; }
 
+  template <typename T> void setDataInfo(tensors_data_info *info, std::vector<T *> &tensors)
```
ragmani, line 58: I think there is no reason to use `function template`

runtime/onert/backend/trix/DevContext.h
```diff
@@ -55,6 +55,64 @@ class DevContext
...
+    info->num_info = static_cast<uint32_t>(tensors.size());
+
+    for (uint32_t idx = 0; idx < info->num_info; ++idx)
+    {
+      info->info[idx].layout = convertDataLayout(tensors[idx]->layout());
+      info->info[idx].type = convertDataType(tensors[idx]->data_type());
+    }
+  }
+
+  template <typename T> void setBuffer(generic_buffers *buf, std::vector<T *> &tensors)
```
ragmani, line 69: Here too

runtime/onert/backend/trix/DevContext.h
jyoungyun: This function supports both case of `const IPortableTensor` as input and `IPortableTensor` as output. So I use template function.

runtime/onert/backend/trix/DevContext.h
jyoungyun: ditto. :)

runtime/onert/backend/trix/DevContext.h
ragmani: I think that a function `void setDataInfo(tensors_data_info *info, const std::vector<const IPortableTensor *> &tensors)` supports both cases. Is there a problem?

jyoungyun: I don't think that const Variable will be compatible with Variable........ Am I wrong..?

```
/opt/test/runtime/onert/backend/trix/ops/BulkLayer.cc:85:40: error: cannot convert 'std::vector<onert::backend::IPortableTensor*>' to 'const std::vector<const onert::backend::IPortableTensor*>&'
   85 |   _dev_context->setDataInfo(&out_info, _outputs);
      |                                        ^~~~~~~~
      |                                        |
      |                                        std::vector<onert::backend::IPortableTensor*>
In file included from /opt/test/runtime/onert/backend/trix/ops/BulkLayer.h:21,
                 from /opt/test/runtime/onert/backend/trix/ops/BulkLayer.cc:17:
/opt/test/runtime/onert/backend/trix/ops/../DevContext.h:58:89: note:   initializing argument 2 of 'void onert::backend::trix::DevContext::setDataInfo(tensors_data_info*, const std::vector<const onert::backend::IPortableTensor*>&)'
   58 |   void setDataInfo(tensors_data_info *info, const std::vector<const IPortableTensor *> &tensors)
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
make[3]: *** [runtime/onert/backend/trix/CMakeFiles/onert_backend_trix.dir/build.make:102: runtime/onert/backend/trix/CMakeFiles/onert_backend_trix.dir/ops/BulkLayer.cc.o] Error 1
```

ragmani: I did't know they are different types. How about using a way (template <class const_iter>
void f(const_iter b, const_iter e) {} ) in [link](https://code-examples.net/en/q/2013e4). It's not madatory, so you don't have to.

jyoungyun: I've viewed the link, but the `setDataInfo` function requires a total size of the tensor vector. If I pass the const_iterator of the vector, I have to assign the total tensor size outside the `setDataInfo` function. It is not a pretty code I think. Why don't we just keep it?

jyoungyun:
PTAL

runtime/onert/backend/trix/DevContext.h
ragmani: OK. The suggested way doesn't look good either.

ragmani (approved):
LGTM

hseok-oh (approved):

PR #8875, opened 2022-04-13
[onert] Make Param as public member on Bulk operation

jyoungyun:
This commit makes Param as public member on Bulk operation.
The Param should be public to access elsewhere.

Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #8876, opened 2022-04-14
[onert] Make Bulk param public

glistening:
Bulk operation's param must be public to be accessable from frontend
like other operation's param.


glistening:
I've found #8875.

PR #8881, opened 2022-04-14
[infra] Disable tests when creating package

llFreetimell:
This commit disable tests when package is created.

---

This PR aims to 
- Disable tests which are not needed when package is created
- Make more precise static analysis

### DIFF
--- a/infra/packaging/preset/20220323
+++ b/infra/packaging/preset/20220323
@@ -39,6 +39,7 @@ function preset_configure()
 
   # TODO Use "nncc configure" and "nncc build"
   cmake \
+    -DENABLE_TEST=OFF \
     -DCMAKE_INSTALL_PREFIX="${NNCC_INSTALL_PREFIX}" \
     -DCMAKE_BUILD_TYPE=release \
     -DBUILD_WHITELIST=$(join_by ";" "${REQUIRED_UNITS[@]}") \


llFreetimell:
For detailed score updates, please refer to 1269561 analysis.

mhs4670go:
@llFreetimell `one-cmds` test will be disabled with this PR that is our release package test.
```cmake
if(NOT ENABLE_TEST)
  return()
endif(NOT ENABLE_TEST)

add_subdirectory(dummy-driver)
add_subdirectory(tests)
add_subdirectory(validate-onnx2circle)
```

I think if this pr should be merged, above if statement should be removed first.

llFreetimell:
@mhs4670go 

Thanks for pointing out, I didn't know that :)
I thought that this PR is valid because windows preset is already disabling the tests as below..!

https://github.com/Samsung/ONE/blob/cf976a753b60bf29a5efdc0e6301e1bcab90dbd9/infra/packaging/preset/20220323_windows#L40

If above windows preset is wrong or it has special reason to disable the tests, I will close this PR :)

mhs4670go:
@llFreetimell Actually, that is because gtest can't be built in Windows. And, FYI, we doesn't support Windows build as of now. That file is just a legacy thing.

infra/packaging/preset/20220323
```diff
@@ -39,6 +39,7 @@ function preset_configure()
 
   # TODO Use "nncc configure" and "nncc build"
   cmake \
+    -DENABLE_TEST=OFF \
```
seanshpark, line 42: @mhs4670go , does our one-compiler-test (?) package related with this?

seanshpark: There was some discussion.
So, is this PR valid?
For me I'm not sure we can disable for PRESET.

seanshpark:
@hseok-oh , do you know how our DC analysis work? does it depend on cmake or build or is it just static code analysis based on folders?

hseok-oh:
>do you know how our DC analysis work? does it depend on cmake or build or is it just static code analysis based on folders?

DC analysis uses build command. IMHO, build command may be used to find actually used code files.

llFreetimell:
> IMHO, build command may be used to find actually used code files.

Yes, `./nnas create-package` command is used for now.
Criteria of "used code files" is interpreted as "files which are needed to create release package".
As `one-cmds-tests` are released, codes which are need to produce them should be considered at analysis.

---

BTW, I have understood that `ENABLE_TEST` is just used for GTest, not `one-cmds-test`.
If `one-cmds-tests` should be always tested because they are included in release package, I think `ENABLE_TEST` condition should be removed from them  IMHO :)

seanshpark:
@llFreetimell , I'm currently testing with my draft to exclude all tests for analysis with external option from anaylsis server.

infra/packaging/preset/20220323
mhs4670go: https://github.com/Samsung/ONE/blob/f1716af6aa38ef4f8d95f4cb4b899229d050236a/compiler/one-cmds/CMakeLists.txt#L120-L125

I think one-cmds tests won't build with this PR. Removing if condition could be an option because it doesn't have ctest.

llFreetimell:
Not tracked anymore. Close.

PR #8893, opened 2022-04-15
[infra] Remove unused mio-tflite260 from ARM32

seanshpark:
This will remove mio-tflite260 from ARM32 build list.


chunseoklee (approved):

mhs4670go (approved):
LGTM

PR #8899, opened 2022-04-15
[enco] Revise test with ENABLE_TEST

seanshpark:
This will revise skip test with ENABLE_TEST.


seanshpark:
From draft #8870 but not related with DC score

mhs4670go (approved):
LGTM

PR #8900, opened 2022-04-15
[luci] Enable skip test

seanshpark:
This will revise to skip test when ENABLE_TEST is not defined.


seanshpark:
From draft #8870

mhs4670go (approved):
LGTM

PR #8904, opened 2022-04-15
[onert/cpu] Fix bool type comparision

hseok-oh:
This commit removes bool type comparison except Equal and NotEqual.
Other operations can't compare bool type.


hseok-oh:
NNAPI requires this feature

PR #8927, opened 2022-04-15
[DRAFT] Non-Const weight in fully connected

AShedko:
This DRAFT shows the changes needed to facilitate support for non-const input to FC.
`res/TensorFlowLiteRecipes/FullyConnected_006/test.recipe`
can be used as a way to check that this pass works.

This change allows circle2circle to replace Fully Connected with Batch Mat Mul when its weight is non-constant.

ONE-DCO-1.0-Signed-off-by: a.shedko <a.shedko@samsung.com>


seanshpark:
@AShedko , thanks for the draft
- dredd test is a test to run `replace_non_const_fc_with_batch_matmul`
- you can edit `circle2circle-dredd-recipe-test` module `test.lst`

seanshpark:
I forgot to tell you one more thing.
If you look at `Net_BroadcastTo_AddV2_000` recipe, there is a `test.rule` file for the dredd test
We can check if the optimization pass works with (1) operator is removed as expected, (2) (new) operator exist as expected.
So I think it would be better to add new recipe with `test.rule` file for this Pass.

seanshpark:
@AShedko , thank you, code looks good and test is OK :)

compiler/luci/pass/src/ReplaceNonConstFCWithBatchMatMulPass.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/Profile/CircleNodeOrigin.h>
+#include <luci/Pass/ReplaceNonConstFCWithBatchMatMulPass.h>
+
+namespace
+{
+
+/**
+ *  Replace channel-wise Mul/Add with DepthwiseConv2D
```
seanshpark, line 25: This comment seems to need update

seanshpark:
Ah, sorry, I think it would be better with one more change.
- `circle2circle-dredd-recipe-test` is to check transformation works correctly
- `luci-pass-value-test` checks before-after values of transformation

File : `compiler/luci-pass-value-test/test.lst`

Add 
```diff
 addeval(Net_StridedSlice_StridedSlice_000 remove_unnecessary_strided_slice)
+addeval(FullyConnected_007 replace_non_const_fc_with_batch_matmul)

 # test SignatureDef, with any optimization
```

AShedko:
All parts in DRAFT were merged

PR #8933, opened 2022-04-17
[infra] Revise preset 20220323

seanshpark:
This will revise preset 20220323 to remove unused modules.


mhs4670go (approved):
LGTM

PR #8938, opened 2022-04-18
[logo] Relocate encoder

seanshpark:
This will relocate encoder to match like others for preparation to
reduce duplicate codes.


mhs4670go (approved):
LGTM

PR #8941, opened 2022-04-18
[luci/import] Introduce importVerifyModule

seanshpark:
This will introduce importVerifyModule from file path, import with verify.


seanshpark:
from draft #8870, to reduce duplicate codes.

mhs4670go (approved):
LGTM

seanshpark:
I think it would be better to introduce `ImportEx` class...

seanshpark:
@mhs4670go , PTAL again

mhs4670go (approved):
LGTM

Anyway, what does `Ex` mean?

seanshpark:
> Anyway, what does Ex mean?

Like `Extended`.
googled and got https://stackoverflow.com/questions/20613172/what-does-suffix-ex-mean-in-function-name

PR #8942, opened 2022-04-18
[arser] Set type default value

mhs4670go:
This commit sets type default value.

### DIFF
--- a/compiler/arser/include/arser/arser.h
+++ b/compiler/arser/include/arser/arser.h
@@ -303,7 +303,7 @@ private:
   std::string _long_name;
   std::string _short_name;
   std::vector<std::string> _names;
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM thank you!

seanshpark:
Q) can I tidy codes that setting default values? I think below can be removed for all user.
```
    .nargs(1)
    .type(arser::DataType::STR)
    .required(false)
```

mhs4670go:
> Q) can I tidy codes that setting default values? I think below can be removed for all user.
> 
> ```
>     .nargs(1)
>     .type(arser::DataType::STR)
>     .required(false)
> ```

Right:)

PR #8944, opened 2022-04-18
[luci] Add require of circlechef

seanshpark:
This will revise requires list with circlechef that is used in tests.


mhs4670go (approved):
LGTM

PR #8948, opened 2022-04-19
[arser] Extract common code print_usage_arg

seanshpark:
This will extract common code print_usage_argto reduce duplicate codes.


seanshpark:
From draft #8870

mhs4670go (approved):
LGTM

PR #8951, opened 2022-04-19
[arser] Helper for version and verbose

seanshpark:
This will introduce Helper class for common version and verbose option.


seanshpark:
From draft #8870

mhs4670go (approved):
LGTM

PR #8953, opened 2022-04-19
[circle2circle] Revise using ImporterEx

seanshpark:
This will revise to use ImporterEx to reduce ducplicate codes.


seanshpark:
from draft #8870

mhs4670go (approved):
LGTM

PR #8959, opened 2022-04-19
[circle-opselector] Revise arser usage

seanshpark:
This will revise arser usage to reduce duplicate codes
- default settings can be removed
- use helper for version


seanshpark:
from draft #8870

mhs4670go (approved):
LGTM

Issue #8935, opened 2022-04-18
[onert] Increase function coverage


hseok-oh:
### What?

Let's increases function coverage

### Why?

Function coverage is 78%. We need function coverage over 80%.

From https://github.com/Samsung/ONE/issues/8775#issuecomment-1098956890

- Hit/Total: 4672/5920 - 78%
  - Need more 120 hit
  - Major miss directory: `compute/cker/include/cker/operation`
    - 370/604 61.3% (line coverage: 3105/3582 86.7%)
    - `Comparison.h`
      - Line coverage 100% (79/79), Function coverage 47.6% (80/168)
      - Maybe `template` 's type issue

hseok-oh:
Today function coverage `compute/cker/include/cker/operation/Comparison.h`

47.6% (80/168) -> 56% (94/168) by #8901

We need more comparision OP tests like #8901

- [ ] `NotEqual`
- [ ] `Less`
- [ ] `LessEqual`
- [ ] `Greater`
- [ ] `GreaterEqual`

hseok-oh:
cc/ @glistening

glistening:
@hseok-oh https://github.com/Samsung/ONE/pull/8901#issuecomment-1099901074

chunseoklee:
cc @jyoungyun since this issue related to TEM

glistening:
I will do `Less`, `LessEqual` ,... If anyone is interested in this task, please start from bottom line (e.g. `GreaterEqual`), and let me know.

chunseoklee:
At the time of writing, we have 78.4% function coverage (4,719/6,019).
To achieve 80% function coverage, 4816(6019*0.8)-4719 = 97 more functions should be covered.

glistening:
I will add more test in `tests/nnfw_api/src/one_op_tests/Cast.cc`.

glistening:
As of 1:38 PM in KST, fn coverage is 4752/6019 (79%). Let's see after #8954 and `Cast`.

glistening:
We may copy `compiler/angkor/src/ADT/tensor/IndexEnumerator.test.cpp`  to cover some of `runtime.libs.misc.include.misc.tensor`. (e.g. `IndexEnumerator.h` (_**0 of 20**_) and `Shape.h` (**_0 of 5_**)). This is basically same structure to make use of tensorflow lite's tensor structure.

chunseoklee:
[...]

PR #8971, opened 2022-04-19
[onert] Add tests for IndexEnumerator

glistening:
It adds test for IndexEnumerator based on
compiler/angor/**/IndexEnumerator.test.cpp.

Related: #8935


glistening:
On x86_64

```
$ Product/x86_64-linux.debug/out/unittest_standalone/misc_index_enumerator_test
Running main() from /home/gyu/z/ONE/externals/GTEST/googletest/src/gtest_main.cc
[==========] Running 1 test from 1 test suite.
[----------] 1 test from MiscIndexEnumeratorTest
[ RUN      ] MiscIndexEnumeratorTest.iterate_full_range
[       OK ] MiscIndexEnumeratorTest.iterate_full_range (0 ms)
[----------] 1 test from MiscIndexEnumeratorTest (0 ms total)

[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 1 test.
```

chunseoklee:
Can we add a negative test for TCM ?

glistening:
I took a look at `IndexEnumerator`. I have no idea what would be a negative test. It doesn't have much room to add.

runtime/libs/misc/CMakeLists.txt
```diff
@@ -8,4 +10,23 @@ target_link_libraries(nnfw_lib_misc PRIVATE nnfw_common)
...
+endif(NOT ENABLE_TEST)
+
+add_executable(misc_index_enumerator_test src/IndexEnumerator.test.cpp)
+target_link_libraries(misc_index_enumerator_test PRIVATE nnfw_lib_misc)
+
+nnfw_find_package(GTest)
+if(NOT GTest_FOUND)
+    message(STATUS "GTest not avaialble. Skipping lib/misc test build")
+    return()
+endif(NOT GTest_FOUND)
```
hseok-oh, line 27: If `GTEST_FOUND` is false, `ENABLE_TEST` become false automatically.
So we don't need this check.

https://github.com/Samsung/ONE/blob/master/infra/nnfw/CMakeLists.txt#L85-L91

hseok-oh:
I suggested to use `*.test.cpp` for unittest. But we cannot filter these naming on code analysis tool.

https://github.com/Samsung/ONE/blob/master/.ahub/sam/exclude.txt

So please use `test/` directory and add exclude list.

glistening:
Hmm.. it is just for cover at most 9 functions. But it seems there are many work to be done to add unit test.
- Negative Test
- Source Relocation for other score

Let me pend this PR. It already took more time than I expected.

hseok-oh:
> I suggested to use `*.test.cpp` for unittest. But we cannot filter these naming on code analysis tool. (it may makes down DC score)
> 
> https://github.com/Samsung/ONE/blob/master/.ahub/sam/exclude.txt
> 
> So please use `test/` directory and add exclude list.

@glistening 
I found out we don't build test code (gtest) on code analysis, so test code doesn't effect to analysis score. 
Sorry for confuse.

glistening:
> @glistening I found out we don't build test code (gtest) on code analysis, so test code doesn't effect to analysis score. Sorry for confuse.

@hseok-oh Never mind. Anyway, it is now under its own test folder as you can see this PR. I will keep this structure. : )

@chunseoklee I've added negative test. Please take a look after all tests pass.

.ahub/sam/exclude.txt
```diff
@@ -24,6 +24,7 @@
 
 # Test codes
 /ONE/tests
+/ONE/runtime/libs/misc/test
```
glistening, line 27: @hseok-oh I dont need to add this according to https://github.com/Samsung/ONE/pull/8971#issuecomment-1103293004. Is it right? Then, I will remove this line.

hseok-oh: Right. Instead, we need to update `.ahub/tcchecker-tca/config.yaml`.

.ahub/tcchecker-tca/config.yaml
```diff
@@ -17,6 +17,7 @@ test:
       - /runtime/onert/test/ir
       - /runtime/onert/test/util
       - /tests/nnfw_api/src
+      - /runtime/libs/misc/test
```
glistening, line 20: @hseok-oh I think it is the only line to be added, but please take a look. It is first time for me to edit this file.

hseok-oh: :+1:

hseok-oh (approved):
LGTM

chunseoklee (approved):

Issue #8889, opened 2022-04-14
[circle-quantizer] QoL improvement: Warnings for poorly representable weights


AShedko:
## What?

It would be nice to issue a warning when a weight (this includes any learned parameters, such as bias values, mean and variance in batch norm, etc.) is poorly represented with a given quantization scheme.

## Why?

Warnings could help debug accuracy problems in compiled models and shorten the time needed to find a bug or issue.

## How?

We can check the range of orders of magnitude of values in a tensor and if it is significantly larger than the number of bits in a given quantization, issue a warning.

## Example

Consider the following pseudocode
```py
A = np.array([300000,30001,0.8e-10])  # Example weight
r = np.log2(A.max())-np.log2(A.min()) # = 51.73581201891895 range of binary orders of magnitude
thresh = 1.5 # Threshold to decrease the number of false-positive warnings
nbits = 16 # for int16, 8 for uint8
if (r > nbits * thresh):
    print("""Weight A is not faithfully represented in the quantized model,
             the resulting model's accuracy may suffer""")
```

This can be implemented as a generic checker which is run for all nodes with `CircleConst` type.

seanshpark:
CC @jinevening

seanshpark:
maybe done? close as inactive for long time

PR #8973, opened 2022-04-19
[luci/DRAFT] Issue warning when quantized weight is poorly representable

AShedko:
This Draft shows an approach to dumping warnings for poorly representable (given quantization params) nodes.
See #8889.

ONE-DCO-1.0-Signed-off-by: a.shedko <a.shedko@samsung.com>

### DIFF
--- a/compiler/circle-quantizer/src/CircleQuantizer.cpp
+++ b/compiler/circle-quantizer/src/CircleQuantizer.cpp
@@ -189,6 +189,13 @@ int entry(int argc, char **argv)
     .required(false)
     .help("Output type of quantized model (uint8 or int16)");
 
+  arser.add_argument("--warn_weights")
+    .nargs(0)
+    .required(false)
+    .default_value(true)
+    .help(
+      "Issue warnings if the quantized model would be inaccurate due to unrepresentable weights");
+
   arser.add_argument(cfg)
     .nargs(1)
     .type(arser::DataType::STR)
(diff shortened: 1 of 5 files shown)


compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -410,8 +417,14 @@ int entry(int argc, char **argv)
...
-    // quantize the graph
-    quantizer.quantize(graph);
+    if (arser["--warn_weights"])
+    {
+      quantizer.check(graph);
+    }
+    else
+    { // quantize the graph
+      quantizer.quantize(graph);
+    }
```
seanshpark, line 427: when `warn_weights` is used, only check is done?
I'd prefer not use use `else`

seanshpark: or if your intension is to just "check" then it would be better to describe it in the help message

AShedko: My intention with this version is to run the weights check only once.

seanshpark: Could you describe more about what `only once` means?
I don't understand why do `quantizer.check` and then do save it to a file.

AShedko: I only want to check constants, not activations

AShedko:
Another way to implement this was selected

PR #8974, opened 2022-04-19
[luci] Issue warnings when weights cannot be accurately represented

AShedko:
This PR logs warnings when weights cannot be accurately represented.
Another way to implement #8889

### DIFF
--- a/compiler/luci/pass/src/QuantizationUtils.cpp
+++ b/compiler/luci/pass/src/QuantizationUtils.cpp
@@ -395,4 +395,74 @@ void quant_const(luci::CircleConst *node, loco::DataType quant_type)
   node->quantparam(std::move(quantparam));
 }
 
+namespace
+{
+
+// TODO move this to a more global helper file
+int nbits(loco::DataType dt) noexcept
+{
+  switch (dt)
+  {
+    case loco::DataType::S8:
+    case loco::DataType::U8:
+      return 8;
+    case loco::DataType::S16:
+    case loco::DataType::U16:
+    case loco::DataType::FLOAT16:
+      return 16;
+    case loco::DataType::S32:
+    case loco::DataType::U32:
+    case loco::DataType::FLOAT32:
+      return 32;
+    case loco::DataType::S64:
+      return 64;
+    default:
+      return 64; // a safe large default
+  }
+}
+
+// TODO Check if the metric is valid
+// Returns true if [min,max] is poorly representable
+bool range_check(float min, float max, loco::DataType dtype)
+{
+  float thresh = 1.5f;
+  return log2f(max) - log2f(min) > nbits(dtype) * thresh;
+}
+
+bool warn_scale_zp(float scale, int64_t zp, luci::CircleNode *n)
+{
+  float min, max;
+  // estimate min/max
+  switch (n->dtype())
+  {
+    case loco::DataType::U8:
+      min = scale * (0 - zp);
+      max = scale * (255 - zp);
+      break;
+    case loco::DataType::S16:
+      min = scale * (-32767);
+      max = scale * (32767);
+      break;
+    default:
+      return false;
+  }
+  return range_check(min, max, n->dtype());
+}
+
+} // namespace
+
+void warn_accuracy_with_range(luci::CircleNode *n)
+{
+  LOGGER(l);
+  auto qp = n->quantparam();
+  auto k = qp->zerop.size();
+  for (uint32_t i = 0; i < k; i++)
+  {
+    if (warn_scale_zp(qp->scale[i], qp->zerop[i], n))
+      WARN(l) << "Quantization of " << i << "-th channel of " << n->name()
+              << "'s quantization may cause accuracy issues" << std::endl;
+    ;
+  }
+}
+
 } // namespace luci
--- a/compiler/luci/pass/src/QuantizationUtils.h
+++ b/compiler/luci/pass/src/QuantizationUtils.h
@@ -78,6 +78,10 @@ void set_int_scale(luci::CircleNode *node);
 // Quantize const tensor using its min/max values
 void quant_const(luci::CircleConst *node, loco::DataType quant_type);
 
+// Check that a node is quantized without significant loss of precision;
+// Emits warnings to log with WARN
+void warn_accuracy_with_range(luci::CircleNode *n);
+
 } // namespace luci
 
 #endif // __LUCI_QUANTIZATION_UTILS_H__
--- a/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
+++ b/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
@@ -636,6 +636,7 @@ bool QuantizeWithMinMaxPass::run(loco::Graph *g)
     auto circle_node = loco::must_cast<luci::CircleNode *>(node);
     if (auto qparam = circle_node->quantparam())
     {
+      warn_accuracy_with_range(circle_node);
       qparam->min.clear();
       qparam->max.clear();
     }


seanshpark:
> compiler/luci/pass/src/QuantizeBias.cpp:221:27: error: comparison of integer expressions of different signedness: ‘int32_t’ {aka ‘int’} and ‘std::vector<float>::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]
03:10:22      for (int32_t i = 0; i < qp->min.size(); i++)
03:10:22                          ~~^~~~~~~~~~~~~~~~

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -395,4 +399,36 @@ void quant_const(luci::CircleConst *node, loco::DataType quant_type)
...
+      return 32;
+    case loco::DataType::S64:
+      return 64;
+    default:
+      return 64; // a safe large default
+  }
+}
+
+} // namespace
+bool range_check(float min, float max, loco::DataType dtype)
```
seanshpark, line 428: please add empty line between these to lines

seanshpark:
If the codes are necessary in release build, its OK, if not, it would be nice if only work in debug mode.
one way I can think of is to put it inside predefined function/macro only valid in debug mode like `assert()`.

AShedko:
> If the codes are necessary in release build, its OK,
I would like warnings to appear in release mode, because we deploy the release mode binaries.

I could have used QuantizeValidationPass, but by the time it is invoked, min/max info is already cleared for some `CircleConst`s

compiler/luci/pass/src/QuantizeWeights.cpp
```diff
@@ -253,6 +255,12 @@ void QuantizeWeights::quantize_weights(luci::CircleConst *weights)
       return;
     }
 
+    for (uint32_t i = 0; i < quantparam->min.size(); i++)
+      if (range_check(quantparam->min[i], quantparam->max[i], output_type))
+        WARN(l) << weights->name()
+                << " cannot be represented accurately according to chosen quantization; " << i
+                << "-th channel's range is too great" << std::endl;
```
seanshpark, line 262: Can you please open/close with curly brace for readability altough it's a single line?

seanshpark: And others too

compiler/luci/pass/src/QuantizeBias.cpp
```diff
@@ -215,6 +217,15 @@ CircleConst *QuantizeBias::quantized_bias(CircleNode *input, const CircleNode *w
       throw std::runtime_error("Unsupported quantization type.");
     }
 
+    auto qp = weight->quantparam();
```
jinevening, line 220: I think your intention is to check qparam of bias?

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -395,4 +399,38 @@ void quant_const(luci::CircleConst *node, loco::DataType quant_type)
...
+      return 64; // a safe large default
+  }
+}
+
+} // namespace
+
+bool range_check(float min, float max, loco::DataType dtype)
+{
+  float thresh = 1.2f;
+  return log2f(max) - log2f(min) > nbits(dtype) * thresh;
```
jinevening, line 433: Do you have any reference for this metric?

`min/max` is changed or ignored according to quantization schemes. So, how about using `scale` rather than `min/max`? Scale is the result of quantization algorithms, thus independent of quantization schemes. In addition, I think it would be possible to make warnings when quantizing activation.

If you'll make warnings only for const nodes, I think that quantization error (e.g., mean percentage error of `|original weights - quantized/dequantized weights|`) is a better proxy to evaluate accuracy.

compiler/luci/pass/src/QuantizeBias.cpp
AShedko: Yes

compiler/luci/pass/src/QuantizationUtils.cpp
AShedko: It's what I used to find a bug with a customer model. 
`mean percentage error of |original weights - quantized/dequantized weights|` is a good metric, but it's more expensive to compute (O(N) vs O(1)).
There's a tradeoff here: how often do we want to warn the user? The more numerous warnings are more easily ignored, because they are perceived to have lower signal to noise ratio.

compiler/luci/pass/src/QuantizeBias.cpp
jinevening: Then you'll need to update `weight` to `new_bias`

compiler/luci/pass/src/QuantizeBias.cpp
```diff
@@ -235,6 +246,10 @@ CircleConst *QuantizeBias::quantized_bias(CircleNode *input, const CircleNode *w
     assert(weight_q->scale.size() == 1); // Only support per-layer quant
     auto weight_scale = weight_q->scale[0];
 
+    if (range_check(weight_q->min[0], weight_q->max[0], loco::DataType::S32))
+      WARN(l) << weight->name()
+              << " cannot be represented accurately according to chosen quantization" << std::endl;
```
jinevening, line 251: This checks the range of weights, not bias.

compiler/luci/pass/src/QuantizationUtils.cpp
jinevening: I see. Then could you add TODO comment like,

'TODO Check if the metric is proper'

jinevening:
@AShedko Can you move the codes to one place? Codes are located in many places, so it'll be difficult to change the metric later.

For example, make a function `warn_accuracy_with_range` and do all things in the function.

The function may be invoked right before min/max values are removed. FYI, min/max values are removed at the end of QuantizeWithMinMaxPass (in the below code).

QuantizeWithMinMax.cpp
```cpp
  // Remove min/max values
  {
    if (auto qparam = circle_node->quantparam())
    {
      qparam->min.clear();
      qparam->max.clear();
    }
  }
```

compiler/luci/pass/src/QuantizeBias.cpp
```diff
@@ -215,6 +233,7 @@ CircleConst *QuantizeBias::quantized_bias(CircleNode *input, const CircleNode *w
       throw std::runtime_error("Unsupported quantization type.");
     }
 
+    //
```
jinevening, line 236: Unnecessary change

AShedko:
`min` and `max` are also removed in `QuantizeWeights.cpp`:
https://github.com/Samsung/ONE/blob/0c22aacc5395190d626c5d3a3392ca4b1a881f52/compiler/luci/pass/src/QuantizeWeights.cpp#L283-L284

jinevening:
Hmm.. how about restoring min/max from scale/zp instead of using saved values ?

For example,

For uint8,  `min = scale * (0 - zp)`, `max = scale * (255 - zp)`
For int16, `min = scale * (-32767)`, `max = scale * (32767)`

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -395,4 +395,81 @@ void quant_const(luci::CircleConst *node, loco::DataType quant_type)
...
+  auto qp = n->quantparam();
+  auto k = qp->zerop.size();
+  bool acc = false;
+  for (uint32_t i = 0; i < k; i++)
+  {
+    WARN(l) << i << "-th channel";
+    acc = acc || warn_scale_zp(qp->scale[i], qp->zerop[i], n);
+  }
+  if (acc)
+    WARN(l) << n->name() << "'s quantization may cause accuracy issues" << std::endl;
```
jinevening, line 472: It would be better to move this code inside loop. E.g.,

```
for ..
{
  if (warn_scale_zp(...))
    WARN(l) "Quantization of " << i << "-th channel of " << n->name() << " may cause accuracy issues" << std::endl; 
}
```

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -395,4 +395,81 @@ void quant_const(luci::CircleConst *node, loco::DataType quant_type)
...
+// TODO Check if the metric is valid
+bool range_check(float min, float max, loco::DataType dtype)
+{
+  float thresh = 1.2f;
+  return log2f(max) - log2f(min) > nbits(dtype) * thresh;
+}
+
+bool warn_scale_zp(float scale, int64_t zp, luci::CircleNode *n)
+{
+  LOGGER(l);
```
jinevening, line 433: Logging inside this function seems unnecessary.

```suggestion
```

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -395,4 +395,81 @@ void quant_const(luci::CircleConst *node, loco::DataType quant_type)
...
+      min = scale * (-32767);
+      max = scale * (32767);
+      break;
+    default:
+      return false;
+      break;
+  }
+  if (range_check(min, max, n->dtype()))
+  {
+    WARN(l) << " cannot be represented accurately according to chosen quantization" << std::endl;
```
jinevening, line 452:
```suggestion
```

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -395,4 +395,81 @@ void quant_const(luci::CircleConst *node, loco::DataType quant_type)
...
+
+void warn_accuracy_with_range(luci::CircleNode *n)
+{
+  LOGGER(l);
+  auto qp = n->quantparam();
+  auto k = qp->zerop.size();
+  bool acc = false;
+  for (uint32_t i = 0; i < k; i++)
+  {
+    WARN(l) << i << "-th channel";
```
jinevening, line 468:
```suggestion
```

jinevening:
Please fix the format issue.

AShedko:
@jinevening, PTAL?

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

PR #8983, opened 2022-04-20
[luci/pass] Extract common code from FuseInstNorm

seanshpark:
This will extract common code from pattern v1 and v5 of FuseInstanceNormPass.


seanshpark:
from draft  #8870

mhs4670go (approved):
LGTM

PR #8984, opened 2022-04-20
[luci/svc] Common code for ArgMax, ArgMin

seanshpark:
This will introduce common code for shape infercnce of CircleArgMax and
CircleArgMin to reduce duplicate codes.


seanshpark:
from draft #8870

mhs4670go (approved):
LGTM

PR #8989, opened 2022-04-20
[luci/pass] Extract more common code from FuseInstNorm

seanshpark:
This will extract common code from pattern v3 and v4 of
FuseInstanceNormPass.


seanshpark:
from draft #8870

mhs4670go (approved):
LGTM

PR #8994, opened 2022-04-21
[tflchef] Relocate tflite Op headers

seanshpark:
This will relocate tflite Op headers to separate from source files.


seanshpark:
from draft #8870

mhs4670go (approved):
LGTM

PR #8996, opened 2022-04-21
[luci-inpt] Add missing arity check in loader

seanshpark:
This will add missing arity assert check in loader.


jinevening (approved):
LGTM

BalyshevArtem (approved):

PR #9001, opened 2022-04-21
[circledump] Guard buffers

seanshpark:
This will revise to guard buffer pointers.


seanshpark:
This came from `circle-tensordump`

mhs4670go (approved):
LGTM

PR #9002, opened 2022-04-21
[circledump] Use mio::circle::builtin_code_neutral

seanshpark:
This will revise to use mio::circle::builtin_code_neutral for correct
builtin_code.


seanshpark:
This was missed somehow when upgrade to mio_circle04 was done.

mhs4670go (approved):
LGTM

PR #9003, opened 2022-04-21
[circledump] Remove unused files

seanshpark:
This will remove unused files that is now done with foder.


mhs4670go (approved):
LGTM

PR #9011, opened 2022-04-21
[tflchef] Fix fill_tensor index

seanshpark:
This will fix to use input tensor index not the input index.


mhs4670go (approved):
LGTM

PR #9017, opened 2022-04-22
[tflchef] Introduce fill_two_inputs helper

seanshpark:
This will introduce fill_two_inputs method as common filler helper.


seanshpark:
from draft #8870

mhs4670go (approved):
LGTM

PR #9019, opened 2022-04-22
[tflchef] Revise using fill_two_inputs

seanshpark:
This will revise using fill_two_inputs to reduce duplicate codes.


seanshpark:
from draft #8870

mhs4670go (approved):
LGTM

Issue #8997, opened 2022-04-21
Refactor: circledump and circle-tensordump


seanshpark:
_circledump_ and _circletensordump_ have common looking codes, where our analysis tool warns with ducplicate codes.

let's refactor this... what can be possible;

(1) make a common library `libcircledump` as a seperate module
(2) make a common library `libcircledump` inside  `circledump` and let `circle-tensordump` use it
(3) merge `circle-tensordump` into `circle-dump` but with two executables

I think start with (2) seems nice, then goto (1) later.

seanshpark:
related #8870

CC @mhs4670go

mhs4670go:
@seanshpark Well, seems to me that there are not that things to be a common,, =ㅁ= If they can be, what you think is a good option I think.

seanshpark:
> seems to me that there are not that things to be a common

Reader.cpp and Reader.cpp have common things :)

seanshpark:
I think can introduce `mio::circle::Reader` in `mio-circle04` for this.

seanshpark:
`circle-inspect` also has `Reader`

seanshpark:
done

PR #9020, opened 2022-04-22
[circledump] Use mio Reader

seanshpark:
This will revise to use mio::circle::Reader helper class.


seanshpark:
for #8997, from draft  #8870

mhs4670go (approved):
LGTM

PR #9025, opened 2022-04-22
[res] Add Fully Connected test with non-const weight

AShedko:
This PR adds Fully Connected test with non-const weight.
full draft: #8927 


seanshpark (approved):
LGTM

PR #9026, opened 2022-04-22
[circle2circle] Add ReplaceNonConstFCWithBatchMatMul option

AShedko:
This PR adds ReplaceNonConstFCWithBatchMatMul option to circle2circle.
The pass allows us to compile FC with non-const weight to Batch MatMul.

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -250,7 +250,7 @@ int entry(int argc, char **argv)
   arser.add_argument("--replace_non_const_fc_with_batch_matmul")
     .nargs(0)
     .default_value(false)
-    .help("Replace Fully Connected with Batch Mat Mul when its weight is non-constant");
+    .help("Replace Fully Connected with BatchMatMul when its weight is non-constant");
 
   arser.add_argument("--substitute_pack_to_reshape")
     .nargs(0)
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -250,7 +250,7 @@ int entry(int argc, char **argv)
   arser.add_argument("--replace_non_const_fc_with_batch_matmul")
     .nargs(0)
     .default_value(false)
-    .help("Replace Fully Connected with BatchMatMul when its weight is non-constant");
+    .help("Replace FullyConnected with BatchMatMul when its weight is non-constant");
 
   arser.add_argument("--substitute_pack_to_reshape")
     .nargs(0)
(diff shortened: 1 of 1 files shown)


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -247,6 +247,11 @@ int entry(int argc, char **argv)
     .help("This will convert weight format of FullyConnected to SHUFFLED16x1FLOAT32. Note that "
           "it only converts weights whose row is a multiple of 16");
 
+  arser.add_argument("--replace_non_const_fc_with_batch_matmul")
+    .nargs(0)
+    .default_value(false)
+    .help("Replace Fully Connected with Batch Mat Mul when its weight is non-constant");
```
seanshpark, line 253: maybe use `BatchMatMul` ? without spaces?

compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -247,6 +247,11 @@ int entry(int argc, char **argv)
     .help("This will convert weight format of FullyConnected to SHUFFLED16x1FLOAT32. Note that "
           "it only converts weights whose row is a multiple of 16");
 
+  arser.add_argument("--replace_non_const_fc_with_batch_matmul")
```
seanshpark, line 250: Q) which would be correct? `fc_with_batch_matmul` or `fc_with_batchmatmul` ?

AShedko: In comments it is usually spelled with a space

AShedko:
```suggestion
  arser.add_argument("--replace_non_const_fc_with_batchmatmul")
```

compiler/circle2circle/src/Circle2Circle.cpp
AShedko:
```suggestion
    .help("Replace Fully Connected with Batch MatMul when its weight is non-constant");
```
Here the CamelCase improves readability

compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -247,6 +247,11 @@ int entry(int argc, char **argv)
...
+    .nargs(0)
+    .default_value(false)
+    .help("Replace Fully Connected with BatchMatMul when its weight is non-constant");
```
seanshpark, line 253: maybe `Fully Connected` -> `FullyConnected` as it's operator type?
I didn't see this when I previously looked...

AShedko:
```suggestion
    .help("Replace FullyConnected with BatchMatMul when its weight is non-constant");
```

seanshpark (approved):
LGTM thank you!

PR #9028, opened 2022-04-22
[luci-pass-value-test] Add test for replace NCFC with Batch MatMul

AShedko:
This PR adds a test for fully-connected layer with non-const weight.


AShedko:
Merge after #9026

seanshpark (approved):
LGTM

Issue #9043, opened 2022-04-25
[luci-eval-driver] major defect detected


seanshpark:
Our analysis tool reports;

> Return value of a function 'circle::GetModel' is dereferenced at EvalDriver.cpp:59 without checking for null, but it is usually checked for this function (6/7).

analysis ID: 1290146

chunseoklee:
Note that it is not urgent one for 1.20.0 release.

seanshpark:
close as change landed

PR #9044, opened 2022-04-25
[luci-eval-driver] Fix major defect

seanshpark:
This will fix major defect of not checking nullptr with calling existing
importVerifyModule method.


seanshpark:
for #9043

mhs4670go (approved):
LGTM

Issue #8731, opened 2022-03-24
[quantizer] Fully fake-quantized model


jinevening:
## What

Let's make a fully fake-quantized model.

## Why

To evaluate a quantized model, we need to run the fp32 model and the quantized model, and compare their results.

However, it is difficult to implement and verify kernels for various quantization schemes.

To tackle this issue, we can use fully fake-quantized model, which can mimic the inference of quantized model with fp32 kernels.

For example,

Input (ofm: u8) -> Conv2D (ifm: u8, weights: u8, bias: s32, ofm: u8) -> Output (ofm: u8) can be approximated to

Input (ofm: fp32) -> **Quantize (ofm: u8) -> Dequantize (ofm: fp32)** -> Conv2D (ifm: fp32, weights: fp32, bias: fp32, ofm: fp32) -> **Quantize (ofm: u8) -> Dequantize (ofm: fp32)** -> Output (ofm: fp32)

NOTE: The word "fully" is used to distinguish from a "partially" fake-quantized model only whose weights are quantized and dequantized before calibration.

@Samsung/one_quantizer

jinevening:
## How to implement?

`--fake_quantization` option will be added to `circle-quantizer`. Input will be a quantized model, and the output will be a fake-quantized model, where Quantize/Dequantize Ops are properly inserted.

The behavior can be implemented with the following passes.

1. FakeQuantizationPass : Insert Quantize/Dequantize Ops to proper places to simulate inference of the quantized model.
2. FoldDequantizePass : Fold Const->Dequantize patterns

## FakeQuantizationPass

For each op (circle node),

Inputs
- ifm: Do nothing (this becomes fp32 by other Ops)
- weights, bias, quantized const: Insert dequantize Op
- integer inputs (axis, index): Do nothing

Output
- ofm: If ofm is not fp32 (i.e., is u8 or s16), change dtype to fp32 and insert Q-DQ after ofm. (Q has the same qparam with ofm)

For virtual output node of multi-output Ops (ex: SplitOut),

Input (single input)
- ifm: Do nothing

Output
- ofm: If ofm is not fp32 (i.e., is u8 or s16), change dtype to fp32 and insert Q-DQ after ofm. (Q has the same qparam with ofm)

NOTE Nodes for multi-out Ops (ex: CircleSplit) have fp32 type and single input, so nothing will happen for them

## FoldDequantizePass

[...]

PR #9062, opened 2022-04-27
[luci/pass] Support more dtype for FoldDequantizePass

llFreetimell:
This commit is for supporting more dtype for `FoldDequatizePass`.

---

Parent Issue : #8731

### DIFF
--- a/compiler/luci/pass/src/FoldDequantizePass.cpp
+++ b/compiler/luci/pass/src/FoldDequantizePass.cpp
@@ -90,27 +90,32 @@ luci::CircleConst *dequantized_const_node(luci::CircleConst *const_node)
     {
       case loco::DataType::S8:
         new_const_node->at<loco::DataType::FLOAT32>(i) =
-          (float)(const_node->at<loco::DataType::S8>(i) - const_node->quantparam()->zerop.at(qd)) *
+          static_cast<float>(const_node->at<loco::DataType::S8>(i) -
+                             const_node->quantparam()->zerop.at(qd)) *
           const_node->quantparam()->scale.at(qd);
         break;
       case loco::DataType::S16:
         new_const_node->at<loco::DataType::FLOAT32>(i) =
-          (float)(const_node->at<loco::DataType::S16>(i) - const_node->quantparam()->zerop.at(qd)) *
+          static_cast<float>(const_node->at<loco::DataType::S16>(i) -
+                             const_node->quantparam()->zerop.at(qd)) *
           const_node->quantparam()->scale.at(qd);
         break;
       case loco::DataType::S32:
         new_const_node->at<loco::DataType::FLOAT32>(i) =
-          (float)(const_node->at<loco::DataType::S32>(i) - const_node->quantparam()->zerop.at(qd)) *
+          static_cast<float>(const_node->at<loco::DataType::S32>(i) -
+                             const_node->quantparam()->zerop.at(qd)) *
           const_node->quantparam()->scale.at(qd);
         break;
       case loco::DataType::S64:
         new_const_node->at<loco::DataType::FLOAT32>(i) =
-          (float)(const_node->at<loco::DataType::S64>(i) - const_node->quantparam()->zerop.at(qd)) *
+          static_cast<float>(const_node->at<loco::DataType::S64>(i) -
+                             const_node->quantparam()->zerop.at(qd)) *
           const_node->quantparam()->scale.at(qd);
         break;
       case loco::DataType::U8:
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/FoldDequantizePass.cpp
```diff
@@ -74,31 +80,41 @@ luci::CircleConst *dequantized_const_node(luci::CircleConst *const_node)
...
+          const_node->quantparam()->scale.at(qd);
+        break;
+      case loco::DataType::S32:
+        new_const_node->at<loco::DataType::FLOAT32>(i) =
+          (float)(const_node->at<loco::DataType::S32>(i) - const_node->quantparam()->zerop.at(qd)) *
+          const_node->quantparam()->scale.at(qd);
+        break;
+      case loco::DataType::S64:
+        new_const_node->at<loco::DataType::FLOAT32>(i) =
+          (float)(const_node->at<loco::DataType::S64>(i) - const_node->quantparam()->zerop.at(qd)) *
```
mhs4670go, line 108: You should use `static_cast` as this is safer way in c++. And, is it safe to convert s64 to float? It becomes to have narrower bits.

llFreetimell: > You should use static_cast as this is safer way in c++

I will apply this :)

> is it safe to convert s64 to float? It becomes to have narrower bits.

"Safe" in your saying is about some disaster like "2^63-1(int64_t) --> static_cast --> 2^32(float)"?
If so, it is safe because the range of `float` is wider than `int64_t` :)
```
Float Range : [-3.40282e+38, 3.40282e+38]
int64_t Range : [-9223372036854775808, 9223372036854775807]
```

Of course, there are inevitable error caused by conversion between integer and float.
But your worrying may not occur :D

jinevening:
The code looks good to me.

(optional) I'd prefer to extract the switch statement out of loop for performance.

before
```
for()
{
  switch()
...
}
```

after
```
switch()
{
for() // change some codes to a function to reduce duplicate code
...
}

File: compiler/luci/pass/src/FoldDequantizePass.test.cpp [hunk 2]
```diff
@@ -15,12 +15,187 @@
...
 #include <gtest/gtest.h>
 
+namespace
+{
+
+template <loco::DataType DT>
+class FoldDequantizeTest : public luci::ConstantFoldingAddTestGraph, public ::testing::Test
+{
+public:
+  S32FoldDequantizeTest() : luci::ConstantFoldingAddTestGraph({2, 2, 2}, DT) {}
```
jinevening, line 29:
```suggestion
  FoldDequantizeTest() : luci::ConstantFoldingAddTestGraph({2, 2, 2}, DT) {}
```

compiler/luci/pass/src/FoldDequantizePass.cpp
mhs4670go: Well, floating point system is not like one to one mapping. Since 23bit(float mantissa) can be one to one mapping, bigger integer may lose precision. But, seems to me that the data are lower than 2^23. So, it will be okay.

llFreetimell:
@jinevening 

> (optional) I'd prefer to extract the switch statement out of loop for performance.

As there are only five `case`, the performance seems not that different I think :)

And already existing code was something like you suggested :)
But I changed it as current PR to reduce duplicated codes because I was hard to find out a function to reduce codes :(
So if you don't mind, can I go as it is? :D

compiler/luci/pass/src/FoldDequantizePass.cpp
llFreetimell: > floating point system is not like one to one mapping.

This is what I thought as "inevitable error caused by conversion between integer and float" :)
So even we use `double` instead of `float`, the problem itself is not removed T.T
The most important thing is... because of `new_const_node->at<loco::DataType::FLOAT32>`, the final result is always float even we use `double` ^.T

jinevening:
> can I go as it is?

It was an optional request. You can go as is.

jinevening (approved):
LGTM Thank you

compiler/luci/pass/src/FoldDequantizePass.cpp
mhs4670go: Yes. Actually, it doesn't needed to be perfectly resolved because integer bigger than 2^32 may not come in.  It's better it prints some warning if it happens to come in:)

mhs4670go (approved):
LGTM

PR #9068, opened 2022-04-28
[infra] Revisit runtime build compiler version check

hseok-oh:
This commit moves GNU compiler version check to root cmake
and remove redundant version checking on runtime build.

If NEON2SSE is used and GNU compiler version is under 9,
tensorflow lite build uses release build option (-O3).

### DIFF
--- a/infra/nnfw/cmake/packages/TensorFlowLite-1.13.1/TensorFlowLite/CMakeLists.txt
+++ b/infra/nnfw/cmake/packages/TensorFlowLite-1.13.1/TensorFlowLite/CMakeLists.txt
@@ -52,6 +52,14 @@ target_compile_definitions(tensorflow-lite PUBLIC "GEMMLOWP_ALLOW_SLOW_SCALAR_FA
 set_property(TARGET tensorflow-lite PROPERTY POSITION_INDEPENDENT_CODE ON)
 target_link_libraries(tensorflow-lite eigen-tf-1.13.1 flatbuffers::flatbuffers ${LIB_PTHREAD} dl)
 
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_BUILD_TYPE_LC STREQUAL "debug")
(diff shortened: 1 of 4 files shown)


infra/nnfw/cmake/packages/TensorFlowLite-1.13.1/TensorFlowLite/CMakeLists.txt
```diff
@@ -52,6 +52,14 @@ target_compile_definitions(tensorflow-lite PUBLIC "GEMMLOWP_ALLOW_SLOW_SCALAR_FA
 set_property(TARGET tensorflow-lite PROPERTY POSITION_INDEPENDENT_CODE ON)
 target_link_libraries(tensorflow-lite eigen-tf-1.13.1 flatbuffers::flatbuffers ${LIB_PTHREAD} dl)
 
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_BUILD_TYPE_LC STREQUAL "debug")
+  if(NEON2SSESource_FOUND AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9)
+    message(WARNING "GNU Compiler version under 9 is not supporting NEON2SSE debug build. "
+                    "Compiler will use O3 (release) build option as a workaround.")
+    target_compile_options(tensorflow-lite PRIVATE -O3 -DNDEBUG)
+  endif()
+endif()
```
glistening, line 61: (optional) I usually don't take a look warning especially during building since there is a lot of messages. I am not sure it is good choice to change build mode to release silently. It is just a personal preference.

glistening (approved):
LGTM

infra/nnfw/cmake/packages/TensorFlowLite-1.13.1/TensorFlowLite/CMakeLists.txt
hseok-oh: >I am not sure it is good choice to change build mode to release silently.

Agree

>But it may be okay since it just do release build only for tensorflow lite.

It will use release for tensorflow lite only for x64 native build if gcc version < 9.
I hope we can find better way.

PR #9069, opened 2022-04-28
[luci] Support more Ops for fake quantization

jinevening:
This supports more Ops for fake quantization.


llFreetimell (approved):
LGTM

PR #9073, opened 2022-04-29
[circle-partitioner] Remove old binary

seanshpark:
This will remove creating binary with old name.


seanshpark:
@mhs4670go PTAL

mhs4670go (approved):
LGTM

PR #9083, opened 2022-05-02
[luci] Support more ops in fake quantization.

jinevening:
This supports more ops in fake quantization.


mhs4670go (approved):
LGTM

PR #9088, opened 2022-05-02
[circle-eval-diff] Support MPEIR metric

jinevening:
This supports Mean PEIR metric.

### DIFF
--- a/compiler/circle-eval-diff/driver/Driver.cpp
+++ b/compiler/circle-eval-diff/driver/Driver.cpp
@@ -71,6 +71,11 @@ int entry(const int argc, char **argv)
     .default_value(false)
     .help("Print Mean Absolute PercentageError");
 
+  arser.add_argument("--print_mpeir")
+    .nargs(0)
+    .default_value(false)
+    .help("Print Mean Peak Error to Interval Ratio");
+
   arser.add_argument("--input_data_format")
     .default_value("h5")
     .help("Input data format. h5/hdf5 (default) or directory");
@@ -114,6 +119,10 @@ int entry(const int argc, char **argv)
   if (arser["--print_mape"] and arser.get<bool>("--print_mape"))
   {
     metrics.emplace_back(Metric::MAPE);
+  }
+  if (arser["--print_mpeir"] and arser.get<bool>("--print_mpeir"))
+  {
+    metrics.emplace_back(Metric::MPEIR);
   }
 
   input_data_format = arser.get<std::string>("--input_data_format");
--- a/compiler/circle-eval-diff/src/MetricPrinter.cpp
+++ b/compiler/circle-eval-diff/src/MetricPrinter.cpp
@@ -285,6 +285,111 @@ void MAPEPrinter::dump(std::ostream &os) const
   }
 }
 
+// TODO Remove duplicate codes with MAEPrinter
+void MPEIRPrinter::init(const luci::Module *first, const luci::Module *second)
+{
+  THROW_UNLESS(first != nullptr, "Invalid module.");
+  THROW_UNLESS(second != nullptr, "Invalid module.");
+
+  const auto first_output = loco::output_nodes(first->graph());
+  const auto second_output = loco::output_nodes(second->graph());
+
+  assert(first_output.size() == second_output.size()); // FIX_CALLER_UNLESS
+
+  for (uint32_t i = 0; i < first_output.size(); i++)
+  {
+    const auto first_node = loco::must_cast<luci::CircleOutput *>(first_output[i]);
+    const auto second_node = loco::must_cast<luci::CircleOutput *>(second_output[i]);
+    assert(same_shape(first_node, second_node)); // FIX_CALLER_UNLESS
+
+    // Create places to store intermediate results
+    _intermediate.emplace_back(0.0);
+
+    // Save output names for logging
+    _output_names.emplace_back(first_node->name());
+  }
+}
+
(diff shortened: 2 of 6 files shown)


jinevening:
@mhs4670go FYI, I posted this PR as MPEIR is an important metric.

compiler/circle-eval-diff/driver/Driver.cpp
```diff
@@ -71,6 +71,11 @@ int entry(const int argc, char **argv)
     .default_value(false)
     .help("Print Mean Absolute PercentageError");
 
+  arser.add_argument("--print_mpeir")
```
mhs4670go, line 74: (optional) How about use capital letter on abbreviations? like `--print_MPEIR` It looks like lower case doesn't refer to what's abbreviated to.

jinevening: Existing options (mae and mape) are not capital letters.

It would be better to discuss this issue in another PR (or issue).

compiler/circle-eval-diff/src/MetricPrinter.cpp
```diff
@@ -285,6 +285,111 @@ void MAPEPrinter::dump(std::ostream &os) const
...
+}
+
+// Accumulate PEIR (Peak Error to Interval Ratio)
+// PEIR = max(|a - b|) / (max(a) - min(a))
+// PEIR >= 0 (lower is better)
+void MPEIRPrinter::accum_peir(uint32_t output_idx, const std::shared_ptr<Tensor> &a,
+                              const std::shared_ptr<Tensor> &b)
+{
+  assert(a->dtype() == loco::DataType::FLOAT32 and
+         b->dtype() == loco::DataType::FLOAT32); // FIX_CALLER_UNLESS
```
mhs4670go, line 320: This assumption looks critical. Doesn't it need to throw an error if it failed?

jinevening: `a` and `b` are casted to fp32 before `accum_peir` is called.

```cpp
void MPEIRPrinter::accumulate(const std::vector<std::shared_ptr<Tensor>> &first,
                              const std::vector<std::shared_ptr<Tensor>> &second)
{
  assert(first.size() == second.size());        // FIX_CALLER_UNLESS
  assert(first.size() == _intermediate.size()); // FIX_CALLER_UNLESS

  for (uint32_t output_idx = 0; output_idx < _intermediate.size(); output_idx++)
  {
    const auto first_output = first[output_idx];
    const auto second_output = second[output_idx];

    // Cast data to fp32 and then compute absolute error
    const auto fp32_first_output = fp32(first_output);
    const auto fp32_second_output = fp32(second_output);

    accum_peir(output_idx, fp32_first_output, fp32_second_output);
  }

  _num_data++;
}
```

mhs4670go: I'm not sure but casting data means they could be non-fp32 type? If not, it checks if fp32 or not rather than casting.

jinevening: The casting was introduced for comparison between different dtypes. But it seems that such a case rarely happens. It would be better to disable the comparison between different dtypes.

jinevening: #9090 will track the issue.

mhs4670go: Oh, I thought what you said means `fp32` function would be deprecated. Are you going to merge this PR without change?

Actually, I also thought two outputs are already fp32 types. So, it doesn't need to type casting. And, if any of them isn't fp32 type, we should throw an error, shouldn't we?

jinevening: > Actually, I also thought two outputs are already fp32 types. So, it doesn't need to type casting. And, if any of them isn't fp32 type, we should throw an error, shouldn't we?

Then circle-eval-diff can be used only for fp32 models. I'm considering support of more types, e.g., u8 or s16 models.

Exception will be thrown if two outputs have different dtypes/shapes.

jinevening: > Oh, I thought what you said means fp32 function would be deprecated. Are you going to merge this PR without change?

fp32 function is for ease of computing metrics. I've updated comment.

jinevening: > casting data means they could be non-fp32 type?

Ah, I didn't answer this question. The answer is yes. It would be good to support non-fp models e.g., u8 VS u8 or s16 VS s16

mhs4670go: Thank you for your explanation:)

'''
// Assumption
// first: the result of fp32 model
// second: the result of fake-quantized model
'''

Actually, this comment confused a bit. I thought two model outputs were fp32 type only as comment said. But seems it is just one of examples of all cases.

mhs4670go (approved):
LGTM

compiler/circle-eval-diff/src/MetricPrinter.cpp
jinevening: Right. The comment for MAPE is a bit confusing. So the comment for MPEIR (in this PR) is slightly improved.

```
// Assumption (when testing the accuracy of quantized model)
// first: the result of fp32 model
// second: the result of fake-quantized model
```

PR #9093, opened 2022-05-02
[onert] Change getBuiltin() type

hseok-oh:
This commit changes BackendManager::getBuiltin() type to backend::Backend.
Builtin backend is not public header, so we cannot use backend::builtin::Backend type on public header "BackendManager.h".

This commit includes public/private header include fix and reorder.

### DIFF
--- a/runtime/onert/core/src/compiler/ExecutorFactory.cc
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.cc
@@ -16,23 +16,22 @@
 
 #include "ExecutorFactory.h"
 
-#include "backend/builtin/Config.h"
-#include "backend/builtin/KernelGenerator.h"
-#include "backend/builtin/TensorBuilder.h"
-#include "backend/builtin/UserTensor.h"
-#include "backend/IPortableTensor.h"
-#include "compiler/BackendManager.h"
-#include "compiler/BackendManager.h"
-#include "compiler/ExecutionBuilder.h"
-#include "compiler/Linear.h"
-#include "dumper/text/GraphDumper.h"
-#include "exec/DataflowExecutor.h"
-#include "exec/ExecTime.h"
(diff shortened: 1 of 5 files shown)


glistening (approved):
LGTM. 

IIUC, 

> Builtin backend is not public header, so we cannot use backend::builtin::Backend type on public header "BackendManager.h".

I think you mean we _should not_ use, by _cannot use_.

chunseoklee:
@ragmani PTAL

runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -16,23 +16,22 @@
...
+#include "../backend/builtin/BackendContext.h"
+#include "../backend/builtin/Config.h"
+#include "../backend/builtin/UserTensor.h"
+#include "../dumper/text/GraphDumper.h"
+#include "../exec/DataflowExecutor.h"
+#include "../exec/ExecTime.h"
+#include "../exec/ExecutionObservers.h"
+#include "../exec/LinearExecutor.h"
+#include "../exec/ParallelExecutor.h"
+#include "../ir/OperationCloner.h"
```
ragmani, line 29: I think `ExecutorFactory.cc` is likely to change its path. Are you going to modify these relative paths?

ragmani (approved):
LGTM

runtime/onert/core/src/compiler/ExecutorFactory.cc
hseok-oh: I change this because they are private header. They may be moved from `include/` to `/src`.

If we remove below workaround, we can meet build fail. To resolve build fail, we may need more similar fix on other files.

https://github.com/Samsung/ONE/blob/f25c3e23ac84365c03c26afcba63bf0640031e55/runtime/onert/core/CMakeLists.txt#L10

Issue #8570, opened 2022-03-04
[circle-partitioner] Include in one-cmds (onecc)


seanshpark:
There are some more features to develop this year(2022) like partition by node name,
I think it is time to include `circle-partitioner` in `onecc` command.

seanshpark:
Had a short off-line meeting with @chunseoklee and @hyunsik-yoon today.
First question is about when to partithion the model?
- current procedure of `onecc` is like import -> optimize -> quntize -> package

What we agreeded as of now is to run partition after quantize and before package like
- import -> optimize -> quantize -> partition -> package

I'd like to open this topic for a while before actually implementhing this.
Any thoughts are welcome :)

Note: we can discussion about flexible procedure in another issue.

seanshpark:
I'll work on this topic as `import -> optimize -> quantize -> partition -> package`

seanshpark:
In the `README.md` of _circle-partitioner_ , 
```
Why does input files be placed in `work` folder too?
- this is still work in progress condition
- use cases are still ambigious
- original `input` model file can be used by the backend, so `.conn` file links it as `source`
- to make things simple for the backend, it would be better not to use relative path for the files
```

Now, it's time to fix this...

seanshpark:
@mhs4670go , can positional argument (the last one) be optional?

lemmaa:
> I'll work on this topic as `import -> optimize -> quantize -> partition -> package`

@seanshpark , oops, I was totally missing out on this issue. I'm sorry. :(
One point I would like to make is that partitions can happen anywhere.

1. `partition -> optimize -> quantize`
    - I'm not sure what the advantages of this are clearly. But I think it's a possible scenario.
2. `optimize -> partition -> quantize`
   - Required when targeting backends that use different quantization schemes. For example, when using CPU/GPU and NPU at the same time
3. `optimize -> quantize -> partition`
   - May be useful for multiple backends using the same quantization scheme. For example multi-core .

Perhaps @chunseoklee first considered case 3. It is expected that the case 2 will be next. :)

seanshpark:
>  oops, I was totally missing out on this issue. I'm sorry. :(

Case 3 was consisdered first, cause when performing quantize after the partition,
we can't provide reference data (real data) for quantization (as of now).

mhs4670go:
> @mhs4670go , can positional argument (the last one) be optional?

No,, it can't.

seanshpark:
[...]

PR #9096, opened 2022-05-03
[infra] Add one-partition to debian install

seanshpark:
This will revise debian to add one-partition in install.


seanshpark:
for #8570

mhs4670go (approved):
LGTM

PR #9074, opened 2022-04-29
DRAFT one-partition

seanshpark:
on-going draft to provide one-partition tool.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #9097, opened 2022-05-03
[one-cmds] Include one-partition

seanshpark:
This will include one-partition into onecc, one-build and build list.


seanshpark:
for #8570, from draft #9074

mhs4670go (approved):
LGTM

Issue #9063, opened 2022-04-27
[onert] Rename and move unittest code file


hseok-oh:
### What?

Let's change unittest (gtest) code file name in `runtime/`, `compute/`, and `test/nnfw_api`, then move into `src/` from `test/`.

### Why?

By using file name to`*.test.cpp` (or `*.test.cc`), we can move them (except `nnfw_api` test) into `src/` to place same position with implementation. It will help to maintain unittest when implementation file and/or function is changed.

Additionally, we can reduce overhead to find test file and function on TCM tools.

- [x] Rename unittest file 
- [x] Update TCM configuration
- Move unittest into `src/`
  - [x] `compute/cker`
  - [x] `runtime/libs/misc`
  - [x] `runtime/libs/ndarray`
  - [x] `runtime/onert/test`

hseok-oh:
All done

PR #9101, opened 2022-05-04
[onert] Move exec and interp unittest

hseok-oh:
This commit moves exec and interp unittest into "src/".

---

Related issue: https://github.com/Samsung/ONE/issues/9063


runtime/onert/core/src/exec/Execution.test.cc
```diff
@@ -14,15 +14,16 @@
  * limitations under the License.
  */
 
-#include <gtest/gtest.h>
-#include <thread>
+#include "exec/Execution.h"
```
glistening, line 17: (optional) Following may be more consistent to others.
```suggestion
#include "Execution.h"
```

glistening (approved):
LGTM

chunseoklee (approved):

Issue #8723, opened 2022-03-23
[one-cmds] Provide visualization facility for memory allocation info


ai-moiseev:
## What?

There's a need for visualization utility for model developers, as stated in [ONE-vscode issue](https://github.com/Samsung/ONE-vscode/issues/309#issue-1160993830) _(Memory measurement in compilation environment and visualization of measurement result)_. This feature requires support in ONE and codegen backends. It also requires unified documented format for backends to emit.

## Why?

There's need in for model development and deployment visualization utilities (with `ONE-vscode` plugin support) for allocation tracking in memory-constrained environments.

### Data format draft

There's already CSV-based format is in use internally, but it was discussed that for more generic usage it would make more sense to use JSON-based format containing all requiring metadata and versioning info.

Proposed allocation data file example:
```json
{
  "schema_version": 1,
  "model_name": "model.circle",
  "codegen_ts": 1648028141,
  
  "memory_segments": [
    {
      "segment": "SEGMENT1",
      "memory_allocations": [
        { "front_addr": 0, "back_addr": 256, "alive_from": 5, "alive_till": 120, "origin": "Node1" }
      ]
    }
  ]
}
```

### Visualization utility

Proposed solution is to introduce `one-cmd` utility for visualization tasks (e.g. `one-visualizer`) with dependency on `matplotlib` and ability to convert memory allocation info JSON files into SVGs and JPEGs. It can be used not only for memory allocations, but for other visualization tasks as well in the future.

### Links

* [ONE-vscode Performance & Memory Management issue](https://github.com/Samsung/ONE-vscode/issues/309)
* [ONE-vscode UX support issue](https://github.com/Samsung/ONE-vscode/issues/395)

seanshpark:
Q) why does this tool should be developed in `ONE`, not in`ONE-vscode` ?

ai-moiseev:
@seanshpark This tool already existed and was in active use in one of backends as a CLI utility. The effort here is to provide cross-backend variant of this utility which has its own value and which can be used even without VSCode plugin (and after that the aim is to provide a GUI for it in VSCode...)

seanshpark:
I actully don't understand what you are going to do. 
Can you please describe what you want to do? add a new tool? what does that tool do? are you going to modify exist codes? etc...

ai-moiseev:
[...]

PR #9104, opened 2022-05-05
[one-cmds] Add 'save-allocations' codegen option

ai-moiseev:
This commit adds 'save-allocations' optional codegen argument which passes this option into codegen backends with allocation trace export support.

Issue: #8723 
ONE-vscode issue: https://github.com/Samsung/ONE-vscode/issues/395 (viewer implementation for allocation trace files)

NOTE: During discussion in #8723 it was stated that 'visualization' solution for memory allocation is outside of ONE's scope, so visualizer implementation task was moved into ONE-vscode (active PR: https://github.com/Samsung/ONE-vscode/pull/483).

Support in ONE is currently limited to `save-allocations` option support (which can be either passed as an argument or config entry) which is implemented in this PR.

### DIFF
--- a/compiler/one-cmds/one-codegen
+++ b/compiler/one-cmds/one-codegen
@@ -90,6 +90,8 @@ def _get_parser(backends_list):
             backends_name) + ')'
     backend_help_message = 'backend name to use ' + backends_name_message
     parser.add_argument('-b', '--backend', type=str, help=backend_help_message)
+    parser.add_argument('--save-allocations', action='store_true',
+        help='save memory allocations trace for supported backends')
 
     return parser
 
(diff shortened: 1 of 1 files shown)


seanshpark:
@ai-moiseev as you can see in the help, format is `one-codegen [-h] [-v] [-C CONFIG] [-b BACKEND] [--] [COMMANDS FOR BACKEND]` where backend options are added after `--` and these are unknwon for and also don't treat in `one-codegen`. I don't think adding new option to here follows this.

seanshpark:
> Although it would be nice to have a documentation for options which may be used by more than one backend. 

I have no idea as of now.

ai-moiseev:
Let's close this for now, then. Maybe in a future there will be a place to document common backend options.

PR #9108, opened 2022-05-09
[circle-eval-diff] Dump output with prefix

jinevening:
This adds a new option to dump output as a file with prefix.

### DIFF
--- a/compiler/circle-eval-diff/src/ModuleEvalDiff.cpp
+++ b/compiler/circle-eval-diff/src/ModuleEvalDiff.cpp
@@ -24,6 +24,7 @@
 #include <stdexcept>
 #include <iostream>
 #include <cassert>
+#include <fstream>
 
 using Tensor = circle_eval_diff::Tensor;
 using DataType = loco::DataType;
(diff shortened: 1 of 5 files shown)


compiler/circle-eval-diff/src/ModuleEvalDiff.cpp
```diff
@@ -200,6 +213,22 @@ void H5InputEvalDiff::evalDiff(const std::string &first_input_data_path,
...
+                          std::to_string(i),
+                        (char *)(out->buffer()), out->byte_size());
+      }
+      for (uint32_t i = 0; i < second_output.size(); i++)
+      {
+        auto out = second_output[i];
+        writeDataToFile(output_prefix + "." + std::to_string(data_idx) + ".second.output" +
+                          std::to_string(i),
+                        (char *)(out->buffer()), out->byte_size());
+      }
```
jinevening, line 231: Just for note. This may change when https://github.com/Samsung/ONE/pull/8830 is landed.

llFreetimell (approved):
LGTM

Issue #8829, opened 2022-04-08
[circle-eval-diff] Redesign the class hierarchy


mhs4670go:
## What 

Let's redesign the class hierarchy of `circle-eval-diff`.

## Why

I have a plan that extends input data format - random input, directory input.

But as of now, `CircleEvalDiff` has too many role, which violates SRP. It makes it difficult to extends new features.

So, I'm gonna redesign the class hierarchy for better extensibility.

The part that loads input data files in the class will be separated and made into a new class.

PR #9109, opened 2022-05-09
[circle-eval-diff] Introduce file data loader

mhs4670go:
This commit introduces file data loader to the module.

Draft: #8830
Related: #8829 

### DIFF
--- a/compiler/circle-eval-diff/CMakeLists.txt
+++ b/compiler/circle-eval-diff/CMakeLists.txt
@@ -6,6 +6,7 @@ list(REMOVE_ITEM SOURCES ${TESTS})
 
 add_executable(circle-eval-diff ${DRIVER} ${SOURCES})
 target_include_directories(circle-eval-diff PRIVATE include)
+target_include_directories(circle-eval-diff PRIVATE src)
 
 target_link_libraries(circle-eval-diff arser)
 target_link_libraries(circle-eval-diff safemain)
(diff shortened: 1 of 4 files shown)


compiler/circle-eval-diff/src/FileDataLoader.h
```diff
@@ -0,0 +1,83 @@
...
+  virtual ~FileDataLoader() = default;
+
+public:
+  virtual uint32_t size(void) const = 0;
+
+public:
+  virtual Data get(uint32_t data_idx) const = 0;
+};
+
+class HDF5Loader : public FileDataLoader
```
jinevening, line 63:
```suggestion
class HDF5Loader final : public FileDataLoader
```

compiler/circle-eval-diff/src/DataSet.h
```diff
@@ -0,0 +1,33 @@
...
+#define __CIRCLE_EVAL_DIFF_DATASET_H__
+
+#include <string>
+#include <vector>
+
+namespace circle_eval_diff
+{
+
+// Data for each input of a single model
+using Input = std::vector<int8_t>;
```
jinevening, line 27: Is there a reason not to use `Tensor`? `Tensor` has `byte_size` method, which can reduce duplicate codes with `byteSizeOf`.

compiler/circle-eval-diff/src/DataSet.h
```diff
@@ -0,0 +1,33 @@
...
+// A set of data for one inference of a single model
+using Data = std::vector<Input>;
```
jinevening, line 29: Hmm.. It seems that the file name `Dataset.h` does not match well with the content (Data is a vector, not a set). How about `InputData.h` or `TensorVector.h`?

Or, I think it would be better to remove this file and simply replace `Data` with `std::vector<Tensor>`.

compiler/circle-eval-diff/src/DataSet.h
mhs4670go: oh, you're right. I didn't consider it. I will change this soon.

compiler/circle-eval-diff/src/FileDataLoader.h
```diff
@@ -0,0 +1,83 @@
...
+{
+
+using DataType = loco::DataType;
+using Shape = std::vector<loco::Dimension>;
+
+void verifyTypeShape(const luci::CircleInput *input_node, const DataType &dtype,
+                     const Shape &shape);
+uint32_t byteSizeOf(const luci::CircleNode *node);
+
+} // namespace circle_eval_diff
```
jinevening, line 39: I guess that you've exposed these methods for testing. If so, `byteSizeOf` is unnecessary to be exposed.

And could you remove `using` keyword in the header file?

compiler/circle-eval-diff/src/FileDataLoader.h
```diff
@@ -0,0 +1,83 @@
...
+{
+
+enum class InputFormat
+{
+  Undefined, // For debugging
+  H5,
+  // TODO Implement Random, Directory
+};
+
+class FileDataLoader
```
jinevening, line 51: Could you rename `FileDataLoader` ? We may have different types of loaders, e.g., random generation or directory.

How about `InputDataLoader`?

jinevening (approved):
LGTM Thank you

mhs4670go:
Please merge this PR:)

PR #9110, opened 2022-05-09
[circle-eval-diff] Add creating empty tensor method

mhs4670go:
This commit adds createEmptyTensor method that literally creates empty tensor from existing CircleNode.

This function is from `compiler/circle-eval-diff/src/CircleEvalDiff.cpp` that will be removed soon.

Draft: https://github.com/Samsung/ONE/pull/8830
Related: https://github.com/Samsung/ONE/issues/8829


jinevening (approved):
LGTM

PR #9115, opened 2022-05-13
[one-cmds] Fix one-parititon

seanshpark:
This will fix getattr with work_path.


mhs4670go (approved):
LGTM

PR #9118, opened 2022-05-13
DRAFT: Find ActivationQType using input

jinevening:
On going draft to find ActivationQType using input for some Ops.

This draft  has the following changes
1. Define ActivationQType for pre-defined values for each Op (This is for reducing duplicate codes)
2. Change input of `make_predefined_qparam`
3. Find ActivationQType by recursively looking at the input for some Ops


jinevening:
This draft allows pre-defined qparams to be created directly from ActivationQType, reducing duplicate codes.

Issue #9130, opened 2022-05-17
[onecc] Annoying warning messages


jinevening:
## What

onecc prints warning messages about buffer size, which makes it hard to read logs.

```
$ ~/ONE/build/release.install/bin/onecc -C mv1-mixed.cfg 
/usr/lib/python3.8/subprocess.py:844: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  self.stdout = io.open(c2pread, 'rb', bufsize)
/usr/lib/python3.8/subprocess.py:849: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  self.stderr = io.open(errread, 'rb', bufsize)
/usr/lib/python3.8/subprocess.py:844: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  self.stdout = io.open(c2pread, 'rb', bufsize)
/usr/lib/python3.8/subprocess.py:849: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  self.stderr = io.open(errread, 'rb', bufsize)
Recording 0'th data
Recording 1'th data
Recording 2'th data
Recording 3'th data
...
```

glistening:
@jinevening I am sorry for writing my opinion after merging. But it occurred to me that I've never met this warnings before. 

Is there any difference in `mv1-mixed.cfg `? It seems call `record-minmax`, and it uses binary mode differently with most other tools. If so, I am not sure it is good idea to change all behaviors for some cases.

jinevening:
> It seems call record-minmax, and it uses binary mode differently with most other tools.

The warning messages were printed in other tools as well (ex: one-import-tflite, tflite2circle, ...). I printed cmd before `_run` function is invoked.

```
hjj@hjj-ubuntu:~/tmp$ ~/ONE/build/release.install/bin/onecc -C mv1-q8.cfg
['/home/hjj/ONE/build/release.install/bin/one-import-tflite', '--config', 'mv1-q8.cfg', '--section', 'one-import-tflite']
/usr/lib/python3.8/subprocess.py:844: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  self.stdout = io.open(c2pread, 'rb', bufsize)
[...]

PR #9131, opened 2022-05-17
[one-cmds] Suppress warning message about line buffer

jinevening:
This suppresses warning message about line buffer.

---
Related to: https://github.com/Samsung/ONE/issues/9130

### DIFF
--- a/compiler/one-cmds/utils.py
+++ b/compiler/one-cmds/utils.py
@@ -154,7 +154,7 @@ def _run(cmd, err_prefix=None, logfile=None):
         logfile: file stream to which both of stdout and stderr lines will be written
     """
     with subprocess.Popen(
-            cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, bufsize=1) as p:
+            cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE) as p:
         import select
         inputs = set([p.stdout, p.stderr])
         while inputs:


jinevening:
This change will make 


```
bufsize will be supplied as the corresponding argument to the [open()](https://docs.python.org/dev/library/functions.html#open) function when creating the stdin/stdout/stderr pipe file objects:

0 means unbuffered (read and write are one system call and can return short)

1 means line buffered (only usable if universal_newlines=True i.e., in a text mode)

any other positive value means use a buffer of approximately that size

negative bufsize (the default) means the system default of io.DEFAULT_BUFFER_SIZE will be used.

Changed in version 3.3.1: bufsize now defaults to -1 to enable buffering by default to match the behavior that most code expects. In versions prior to Python 3.2.4 and 3.3.1 it incorrectly defaulted to 0 which was unbuffered and allowed short reads. This was unintentional and did not match the behavior of Python 2 as most code expected.
```
From https://docs.python.org/ko/dev/library/subprocess.html#popen-constructor

glistening (approved):
LGTM. However, someone else would be better person to review this change. I just factor out the repeated code (e.g. `one-build` https://github.com/Samsung/ONE/blame/22989ff7b06f1f330290999354292ebb1dfafd9b/compiler/one-cmds/one-build#L135-L138, ... ) as function in `utils.py`, I don't have idea why someone explicitly passed `bufsize=1`. 

Anyway, if it removes annoying warnings, I am okay. : )

mhs4670go:
Well, actually, I don't remember the reason. Maybe it was for instant printing or flushing of buffer. Is there any difference between log of original and the one from this PR? Maybe passing all `one-cmds` tests will do. If there's no problem, it seems okay for this PR to be merged.

mhs4670go (approved):
LGTM

PR #9134, opened 2022-05-18
[onert] Remove core private include

hseok-oh:
This commit removes private type target_include_directories in core.
It prevents to include private header in public header.

This commit includes header include reordering and removing redundant include.


glistening (approved):
LGTM

PR #9141, opened 2022-05-19
DRAFT: Support evaluation of quantized model

jinevening:
This supports evaluation of quantized model.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/9124

Issue #9124, opened 2022-05-16
[one-quantize] Evaluate quantized model


jinevening:
## What

Let's implement an option to evaluate quantized model

![image](https://user-images.githubusercontent.com/5449554/168549057-659c9322-93d0-4eaf-91fc-24070c78f04b.png)

I'm thinking about the interface as follows.

```
[one-quantize]
input_path=test.circle
output_path=test.q.circle
input_data=representative.h5
test_data=test.h5
```

If `test_data` option is given, one-quantize will run the data with the fp32 model and the quantized model, and print the evaluation results (e.g., MAE (Mean Absolute Error), MPEIR (Mean Peak Error to Interval Ratio)).

CC @Samsung/one_quantizer

seanshpark:
FYI using dark theme doesn't show the lines or text in the image.

jinevening:
I've added white background :)

jinevening:
Interface will be updated as below. (from https://github.com/Samsung/ONE/pull/9143#issuecomment-1132282943)

---

The current interface (using `test_data`) is simple but limited. For example, we may not give `test_data` to use random data for simple testing. The current interface cannot support such a scenario.

I'd like to update the interface as below.

```
[one-quantize]
input_path=..
output_path=..
quantized_dtype=..
evaluate_result=True
test_data=..
print_mpeir=True
print_top1_match=True
print_top5_match=True
```

Evaluation is conducted when `evaluate_result` is True. And more detailed control would be possible for test_data or metrics.

seanshpark:
@jinevening , just to be sure; is current change you are doing is backward compatible?

jinevening:
> is current change you are doing is backward compatible?

I think so. In the viewpoint of interface, I'm not changing the existing interface but just adding new options.

In the viewpoint of modules, `circle-eval-diff` uses modules already used by other tools (record-minmax, circle-quantizer).

jinevening:
Done.

PR #9143, opened 2022-05-19
[one-cmds] Support evaluation of quantized model

jinevening:
This supports evaluation of quantized model.
1. test_data option is added to one-quantize.
2. If test_data is given, circle-eval-diff evaluates inference results
   between fake-quantized model and original fp32 model.

---
Related to: https://github.com/Samsung/ONE/issues/9124
Draft PR: https://github.com/Samsung/ONE/pull/9141


seanshpark (dismissed):
LGTM

compiler/one-cmds/one-quantize
```diff
@@ -130,6 +130,12 @@ def _get_parser():
         help=
         "Path to the quantization configuration file."
     )
+    quantization_group.add_argument(
+        '--test_data',
+        type=str,
+        help=
+        "Path to the test data. Evaluation result will be printed if test data is given."
```
mhs4670go, line 137: (trivail)
```suggestion
        "Data path to test. Evaluation result will be printed if test data is given."
```

compiler/one-cmds/one-quantize
```diff
@@ -360,6 +366,42 @@ def _quantize(args):
...
+            circle_eval_diff_cmd.append('--second_input_data')
+            circle_eval_diff_cmd.append(getattr(args, 'test_data'))
+            if _utils._is_valid_attr(args, 'input_data_format'):
+                circle_eval_diff_cmd.append('--input_data_format')
+                circle_eval_diff_cmd.append(getattr(args, 'input_data_format'))
+            circle_eval_diff_cmd.append('--print_mae')
+            circle_eval_diff_cmd.append('--print_mpeir')
+
+            f.write((' '.join(circle_eval_diff_cmd) + '\n').encode())
+            _utils._run(circle_eval_diff_cmd, err_prefix="circle_eval_diff", logfile=f)
```
mhs4670go, line 404: Seems `quantize` method will become longer and longer. How about make them several functions later?

mhs4670go: Above two comments are optional.

compiler/one-cmds/one-quantize
jinevening: "Path to XX" is the style we've used in other options. And I'm not sure"Data path" is a widely used term.

jinevening:
I put NO_MERGE label to change the interface.

The current interface (using `test_data`) is simple but limited. For example, we may not give `test_data` to use random data for simple testing. The current interface cannot support such a scenario.

I'd like to update the interface as below.

```
[one-quantize]
input_path=..
output_path=..
quantized_dtype=..
evaluate_result=True
test_data=..
print_mpeir=True
print_top1_match=True
print_top5_match=True
```

Evaluation is conducted when `evaluate_result` is True. And more detailed control would be possible for test_data or metrics.

I'll remove the NO_MERGE label after updating the interface.

compiler/one-cmds/utils.py
```diff
@@ -61,6 +61,42 @@ def is_accumulated_arg(arg, driver):
 def _is_valid_attr(args, attr):
     return hasattr(args, attr) and getattr(args, attr)
 
+class Command:
```
jinevening, line 64: I made this class to easily setup options.

compiler/one-cmds/one-quantize
```diff
@@ -130,6 +131,36 @@ def _get_parser():
         help=
         "Path to the quantization configuration file."
     )
+    quantization_group.add_argument(
+        '--evaluate_result',
```
jinevening, line 135: This option is added to turn on evaluation.

compiler/one-cmds/one-quantize
```diff
@@ -360,6 +366,42 @@ def _quantize(args):
...
+            tmp_fake_quant_model = os.path.join(
+                tmpdir,
+                os.path.splitext(os.path.basename(args.input_path))[0]) + '.fake_quant.circle'
+
+            # do fake quantization
+            fake_quantize_cmd = [circle_quantizer_path]
+            if _utils._is_valid_attr(args, 'verbose'):
+                fake_quantize_cmd.append('--verbose')
+            fake_quantize_cmd.append('--fake_quantize')
+            fake_quantize_cmd.append(getattr(args, 'output_path'))
```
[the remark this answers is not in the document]
jinevening: I've made `Command` class to easily add option only when attr is valid. PTAL

seanshpark (approved):
LGTM

compiler/one-cmds/one-quantize
jinevening: > Also, adding test about this change would be helpful.

I've added tests.

compiler/one-cmds/one-quantize
```diff
@@ -360,6 +391,33 @@ def _quantize(args):
         # run circle-quantizer
         _utils._run(circle_quantizer_cmd, err_prefix="circle_quantizer", logfile=f)
 
+        # evaluate
+        if _utils._is_valid_attr(args, 'evaluate_result'):
+            circle_eval_diff_path = os.path.join(dir_path, 'circle-eval-diff')
+            quant_model = ""
+            if _utils._is_valid_attr(args, 'output_path'):
+              quant_model =getattr(args, 'output_path')
```
mhs4670go, line 399:
```suggestion
              quant_model = getattr(args, 'output_path')
```

compiler/one-cmds/one-quantize
```diff
@@ -360,6 +391,33 @@ def _quantize(args):
         # run circle-quantizer
         _utils._run(circle_quantizer_cmd, err_prefix="circle_quantizer", logfile=f)
 
+        # evaluate
+        if _utils._is_valid_attr(args, 'evaluate_result'):
+            circle_eval_diff_path = os.path.join(dir_path, 'circle-eval-diff')
+            quant_model = ""
+            if _utils._is_valid_attr(args, 'output_path'):
+              quant_model = getattr(args, 'output_path')
```
mhs4670go, line 399:
```suggestion
                quant_model = getattr(args, 'output_path')
```

compiler/one-cmds/one-quantize
```diff
@@ -360,6 +391,33 @@ def _quantize(args):
...
+            circle_eval_diff_cmd = Command(circle_eval_diff_path, args, f)
+            circle_eval_diff_cmd.add_option_with_valid_args('--first_model', ['input_path']) \
+                .add_option_with_values('--second_model', [tmp_fake_quant_model]) \
+                .add_option_with_valid_args('--first_input_data', ['test_data']) \
+                .add_option_with_valid_args('--second_input_data', ['test_data']) \
+                .add_option_with_valid_args('--input_data_format', ['input_data_format']) \
+                .add_noarg_option_if_valid_arg('--print_mae', 'print_mae') \
+                .add_noarg_option_if_valid_arg('--print_mape', 'print_mape') \
+                .add_noarg_option_if_valid_arg('--print_mpeir', 'print_mpeir') \
+                .run()
```
mhs4670go, line 420: Hmmm.. I think I understand why you implement `Command` class and it's a very good idea. But, IMHO, I think the name of methods are a bit verbose and seems that they even don't need to be called to me because `args` already has the contents for the command.

Can I revisit this class later? Maybe the changes are gonna be like this.
```suggestion
            circle_eval_diff_cmd = Command(Command::Driver::circle-eval-diff, args, log_file=f)
            circle_eval_diff_cmd.run()
```

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

compiler/one-cmds/one-quantize
jinevening: Looks good to me. Hide details into Command class is a good idea.

jinevening: > Can I revisit this class later?

Of course :)

>  seems that they even don't need to be called to me because args already has the contents for the command.

FYI, `tmp_fake_quant_model` is not in args.

PR #9147, opened 2022-05-19
[circle-eval-diff] Introduce Top-K Match Printer

jinevening:
This introduces Top-K Match Printer.

---
Related to: https://github.com/Samsung/ONE/issues/9124
Draft PR: https://github.com/Samsung/ONE/pull/9141

### DIFF
--- a/compiler/circle-eval-diff/src/MetricPrinter.cpp
+++ b/compiler/circle-eval-diff/src/MetricPrinter.cpp
@@ -30,6 +30,16 @@ using Tensor = circle_eval_diff::Tensor;
 namespace
 {
 
+uint32_t num_elems(const luci::CircleNode *node)
+{
+  uint32_t res = 1;
+
+  for (uint32_t i = 0; i < node->rank(); i++)
+    res *= node->dim(i).value();
+
+  return res;
+}
+
 template <typename T> bool same_shape(const T a, const T b)
 {
   if (a->rank() != b->rank())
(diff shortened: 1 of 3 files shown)


compiler/circle-eval-diff/src/MetricPrinter.cpp
```diff
@@ -412,6 +422,174 @@ void MPEIRPrinter::dump(std::ostream &os) const
...
+         b->dtype() == loco::DataType::FLOAT32); // FIX_CALLER_UNLESS
+  assert(same_shape(a.get(), b.get()));          // FIX_CALLER_UNLESS
+  assert(output_idx < _intermediate.size());     // FIX_CALLER_UNLESS
+
+  // Find Top-k largest elements
+  // This implementation is a variant of "Method 2 (Use temporary array)" in
+  // https://www.geeksforgeeks.org/k-largestor-smallest-elements-in-an-array/
+  // We sort top-k elements by value and index to ensure that the element with an earlier
+  // index comes first if multiple elements have the same value.
+  auto find_topk = [this](const std::shared_ptr<Tensor> &tensor) {
```
llFreetimell, line 479: Can I suggest alternative implementation for this?
It's not because of code quality, but because of the source code origin and copyright...!
It seems that there are no strict issue but it would be much safer to use our own code :D

jinevening: > it would be much safer to use our own code

Hmm.. I did not copy any code from the reference but wrote the whole code by myself.

You're familiar with license, so I'll follow your opinion. If you think it is safer to use different implementation, please share yours when ready.

llFreetimell: Ah, I misunderstood that some of codes are copied from it :(
Sorry, please ignore my comment...!

llFreetimell (approved):
LGTM

PR #9122, opened 2022-05-16
[Draft] Introduce one-infer to one-cmds

yunjayh:
This commit introduces `one-infer` to `one-cmds` which enables inference tool
of backend driver, and the other runtimes, including tflite(included) and onnx(not yet)

Signed-off-by: yunjay-hong <yunjay.hong@samsung.com>

---

for: #8970

Issue #8970, opened 2022-04-19
[compiler/one-cmds] Introduce one-infer


yunjayh:
### What

Let's introduce `one-infer` to onecc as a inferring wrapper binary.

### Why

ONE-vscode is planed to support End-to-end value validation test. `one-infer` will execute the backend models' inference tool as `one-profile` does. 

There was a previous work at #7299 

- [x] Create inference wrapper binary (https://github.com/Samsung/ONE/pull/7419)
  - [x]  Add one-inference man page
  - [x]  Add test cases
  - [x]  Add debian package
- [ ] Introduce `tflite-infer`, `onnx-infer` for each runtime

yunjayh:
#### [WIP] Let's define data spec for each data format (e.g. `npy` and `h5`). 

Let's assume there is a `model.tflite`, which takes two input tensors (their shapes are (1,5,5,3) and (2,2) for each), and the 1000 sample images which is used for input. After data-preprocessing stage, user got 1000 pairs of input data. 

- For npy
```sh
$ tree ./input_data
input_data
├── input_data.0.0.npy
├── input_data.0.1.npy
├── input_data.1.0.npy
├── input_data.1.1.npy
├── ...
├── input_data.999.0.npy
└── input_data.999.1.npy

$ tflite-infer model.tflite --input-spec npy --load-input-from input_data ...
running....
```

- For h5
```sh
$ some-cmd-show-h5-hierarchy input_data.h5
# GROUP "/"
# ㄴGROUP "input"
#   ㄴGROUP "0"
#     ㄴDATASET "0"
#       ㄴDATA ... [shape (1,5,5,3)]
#     ㄴDATASET "1"
#       ㄴDATA ... [shape (2,2)]
#   ㄴGROUP "1"
#     ㄴDATASET "0"
#       ㄴDATA ... [shape (1,5,5,3)]
#     ㄴDATASET "1"
#       ㄴDATA ... [shape (2,2)]
#   ...
#   ㄴGROUP "999"
#     ㄴDATASET "0"
#       ㄴDATA ... [shape (1,5,5,3)]
#     ㄴDATASET "1"
#       ㄴDATA ... [shape (2,2)]

$ tflite-infer model.tflite --input-spec h5 --load-input-from input_data ...
running....
```

yunjayh:
#### [WIP] How about extracting data generator?

[...]

PR #9158, opened 2022-05-23
[one-cmds] Introduce `one-infer` to `one-cmds`

yunjayh:
This commit introduces `one-infer` which executes backend inference tool.

---

for: #8970 
draft: #9122

### DIFF
--- a/compiler/one-cmds/one-infer
+++ b/compiler/one-cmds/one-infer
@@ -34,10 +34,6 @@ sys.tracebacklimit = 0
 
 
 def _get_backends_list():
-    # This approach is too much backend-depended, 
-    # since if the backend driver doesn't use {backend}-infer as its driver,
-    # then `one-infer` cannot run the driver. 
-    # TODO: introduce new layer which maps backend infer driver to one-infer
     """
     [one hierarchy]
     one
@@ -76,12 +72,55 @@ def _get_backends_list():
     return backends_list
 
 
+def _search_backend_driver(driver):
+    """
+    [one hierarchy]
+    one
+    ├── backends
+    ├── bin
+    ├── doc
+    ├── include
+    ├── lib
+    └── test
+
+    The list where `one-infer` finds its backend driver
+    - `bin` folder where `one-infer` exists
+    - `backends/**/bin/` folder
+
+    NOTE If there are drivers of the same name in different places,
+     the closer to the top in the list, the higher the priority.
+    """
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+
+    # CASE 0: check drvier name validation
+    if not driver:
+        return None
+
+    # CASE 1: one/bin/{driver} is found
+    driver_path = dir_path + '/' + driver
+    if os.path.isfile(driver_path) and os.access(driver_path, os.X_OK):
+        return driver_path
+
+    # CASE 2: one/backends/**/bin/{driver} is found
+    for driver_path in glob.glob(dir_path + '/../backends/**/bin/' + driver, recursive=True):
+        print(driver_path)
+        if os.path.isfile(driver_path) and os.access(driver_path, os.X_OK):
+            return driver_path
+
+    # CASE 3: {driver} is found in nowhere
+    return None
+
+
 def _get_parser(backends_list):
-    infer_usage = 'one-infer [-h] [-v] [-C CONFIG] [-b BACKEND] [--] [COMMANDS FOR BACKEND]'
+    infer_usage = 'one-infer [-h] [-v] [-C CONFIG] [-d DRIVER] [-b BACKEND] [--] [COMMANDS FOR BACKEND DRIVER]'
     parser = argparse.ArgumentParser(
         description='command line tool to infer model', usage=infer_usage)
 
     _utils._add_default_arg(parser)
+
+    # TODO: add tflite/onnx-infer driver to helper message when it is implemented
+    driver_help_message = 'backend inference driver name to execute'
+    parser.add_argument('-d', '--driver', type=str, help=driver_help_message)
 
     # get backend list in the directory
     backends_name = [ntpath.basename(f) for f in backends_list]
@@ -100,8 +139,8 @@ def _verify_arg(parser, args):
     """verify given arguments"""
     # check if required arguments is given
     missing = []
-    if not _utils._is_valid_attr(args, 'backend'):
-        missing.append('-b/--backend')
+    if not _utils._is_valid_attr(args, 'backend') and not _utils._is_valid_attr(args, 'driver'):
+        missing.append('-d/--driver or -b/--backend')
     if len(missing):
         parser.error('the following arguments are required: ' + ' '.join(missing))
 
@@ -109,27 +148,45 @@ def _verify_arg(parser, args):
 def _parse_arg(parser):
     infer_args = []
     backend_args = []
-    unknown_args = []
     argv = copy.deepcopy(sys.argv)
     # delete file name
     del argv[0]
     # split by '--'
     args = [list(y) for x, y in itertools.groupby(argv, lambda z: z == '--') if not x]
-    # one-infer has two interfaces
-    # 1. one-infer [-h] [-v] [-C CONFIG] [-b BACKEND] [COMMANDS FOR BACKEND]
-    if len(args) == 1:
+    
+    # one-infer [-h] [-v] [-C CONFIG] [-d DRIVER] [-b BACKEND] -- [COMMANDS FOR BACKEND DRIVER]
+    if len(args):
         infer_args = args[0]
-        infer_args, unknown_args = parser.parse_known_args(infer_args)
-    # 2. one-infer [-h] [-v] [-C CONFIG] [-b BACKEND] -- [COMMANDS FOR BACKEND]
-    if len(args) == 2:
-        infer_args = args[0]
-        backend_args = args[1]
         infer_args = parser.parse_args(infer_args)
+        backend_args = backend_args if len(args) < 2 else args[1]
     # print version
     if len(args) and infer_args.version:
         _utils._print_version_and_exit(__file__)
 
-    return infer_args, backend_args, unknown_args
+    return infer_args, backend_args
+
+
+def _get_executable(args, backends_list):
+    executable = None
+    missing = []
+
+    driver = getattr(args, 'driver')
+    if driver:
+        executable = _search_backend_driver(driver)
+        if executable:
+                return executable
+        missing.append(driver)
+
+    if getattr(args, 'backend'):
+        backend_base = getattr(args, 'backend') + '-infer'
+        for cand in backends_list:
+            if ntpath.basename(cand) == backend_base:
+                executable = cand
+                return executable
+        missing.append(backend_base)
+
+    if not executable:
+        raise FileNotFoundError(' or '.join(missing) + ' not found')
 
 
 def main():
@@ -138,7 +195,7 @@ def main():
 
     # parse arguments
     parser = _get_parser(backends_list)
-    args, backend_args, unknown_args = _parse_arg(parser)
+    args, backend_args = _parse_arg(parser)
 
     # parse configuration file
     _utils._parse_cfg(args, 'one-infer')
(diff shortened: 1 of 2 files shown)


compiler/one-cmds/one-infer
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env bash
```
yunjayh, line 1: This code is mostly copied from `one-profile`.

yunjayh:
I manually added reviewers whom I could find from reviewer lists in `one-cmds`-related-PRs. If it's inappropriate, please disassign yourself..

yunjayh:
The tests are simply copied from `one-profile` to `one-infer`.

```console
$ ./test/runtestall.sh 
one-infer_001.test SUCCESS
one-infer_002.test SUCCESS
one-infer_003.test SUCCESS
one-infer_004.test SUCCESS
one-infer_neg_001.test SUCCESS
ALL TESTS PASSED!
```

seanshpark (approved):
LGTM

compiler/one-cmds/one-infer
```diff
@@ -0,0 +1,166 @@
...
+import sys
+
+import utils as _utils
+
+# TODO Find better way to suppress trackback on error
+sys.tracebacklimit = 0
+
+
+def _get_backends_list():
+    # This approach is too much backend-depended,
```
mhs4670go, line 37: I'm not sure but comments sounds like to me that this approach should follow backend rules. But it is the exact opposite.
```suggestion
    # This approach forces the backend drivers to have a restricted name.
```

mhs4670go: Anyway, is it natural to do so? Drivers for inference seems hard to have restricted name. How about just add `-d, --driver` option where we can pass exact driver path to `one-infer`?

compiler/one-cmds/one-infer
```diff
@@ -0,0 +1,166 @@
...
+    # delete file name
+    del argv[0]
+    # split by '--'
+    args = [list(y) for x, y in itertools.groupby(argv, lambda z: z == '--') if not x]
+    # one-infer has two interfaces
+    # 1. one-infer [-h] [-v] [-C CONFIG] [-b BACKEND] [COMMANDS FOR BACKEND]
+    if len(args) == 1:
+        infer_args = args[0]
+        infer_args, unknown_args = parser.parse_known_args(infer_args)
+    # 2. one-infer [-h] [-v] [-C CONFIG] [-b BACKEND] -- [COMMANDS FOR BACKEND]
```
mhs4670go, line 123: We don't have to cover both cases. Please just allow second one. The reason other drivers do that is for backward compatibility.

compiler/one-cmds/one-infer
```diff
@@ -0,0 +1,166 @@
...
+    infer_cmd = [infer_path] + backend_args + unknown_args
+    if _utils._is_valid_attr(args, 'command'):
+        infer_cmd += getattr(args, 'command').split()
+
+    # run backend driver
+    _utils._run(infer_cmd, err_prefix=backend_base)
+
+
+if __name__ == '__main__':
+    main()
```
mhs4670go, line 166:
```suggestion
    _utils._safemain(main, __file__)
```

mhs4670go: This is for better error handling.

compiler/one-cmds/one-infer
yunjayh: As a wrapper which enables backend inference driver, it sounds better to specify its name. 

Here are some limitation if we apply the change.

1. driver candidates cannot be given to user. 
For now, `one-infer` searches `*-infer` from `one/bin/` and `one/backends/**/bin/* and shows them as backend driver candidates. `-d/--driver` option cannot suggest the same option. 

2. Cannot sure the driver argument is **inference tool**
Since `one-infer` works as a wrapper command for `one-cmds`, any driver can be passed and executed in `one-infer` stage. But, IMHO it cannot be solved in this situation. (just for note)

I'll apply the changes and revise some help messages.

mhs4670go: Hmmm.. on second thought, the original codes is necessary because `one-codegen` already does that. 

(optional) Maybe just adding `--driver` option will do the trick. I mean we can use both of `-b` and `-d` option. Then,

> driver candidates cannot be given to user.

Just printing `-infer` driver list is enough.

> Cannot sure the driver argument is **inference tool**

We don't need to know about this. It would cause an error.

yunjayh: > Hmmm.. on second thought, the original codes is necessary because `one-codegen` already does that.
> (optional) Maybe just adding `--driver` option will do the trick. I mean we can use both of `-b` and `-d` option.

`one-codegen` does what? `-b/--backend` option? Yes, for consistency, it'd be better to remaining `-b` option.
However if there exists both `-b` and `-d` options, both of the options should be **optional** argument, not required. It's a little bit weird for me. Does it look fine?

mhs4670go: > Does it look fine?

Seems fine to me as of now because other tools already have `-b` option.

But, there could be some drivers that we cannot control including its name. So, we should use `-d` option later. Or, we could make `-b` option the one that receive backend driver path.

@seanshpark What do you think of this?

seanshpark: > What do you think of this?

I wasn't following the context so had a short chat with @yunjayh :)
If both `-b` and `-d` exist then we can choose with `priority`; give `-d` higher so if `-d` is given we choose the driver, if not then choose `-b` option. both absent will be an error.

yunjayh:
```console
$ ./test/runtestall.sh 
one-infer_001.test SUCCESS
one-infer_002.test SUCCESS
one-infer_003.test SUCCESS
one-infer_004.test SUCCESS
one-infer_005.test SUCCESS
one-infer_006.test SUCCESS
one-infer_007.test SUCCESS
one-infer_neg_001.test SUCCESS
one-infer_neg_002.test SUCCESS
one-infer_neg_003.test SUCCESS
one-infer_neg_004.test SUCCESS
ALL TESTS PASSED!
```

You can find the tests at #9122 bc07d7ed9b524065d065f439c396ba4fb28ba162

compiler/one-cmds/one-infer
yunjayh: Now, it's much more added from one-profile. I'll collapse this comment

compiler/one-cmds/one-infer
```diff
@@ -76,13 +72,56 @@ def _get_backends_list():
...
+    - `backends/**/bin/` folder
+
+    NOTE If there are drivers of the same name in different places,
+     the closer to the top in the list, the higher the priority.
+    """
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+
+    # CASE 0: check drvier name validation
+    if not driver:
+        return None
```
mhs4670go, line 97: When does this condition happen?

compiler/one-cmds/one-infer
```diff
@@ -76,13 +72,56 @@ def _get_backends_list():
...
+
+    # CASE 1: one/bin/{driver} is found
+    driver_path = dir_path + '/' + driver
+    if os.path.isfile(driver_path) and os.access(driver_path, os.X_OK):
+        return driver_path
+
+    # CASE 2: one/backends/**/bin/{driver} is found
+    for driver_path in glob.glob(dir_path + '/../backends/**/bin/' + driver, recursive=True):
+        print(driver_path)
```
mhs4670go, line 106: Maybe for debugging?

compiler/one-cmds/one-infer
mhs4670go: Umm.. I think we don't have to receive both of them simultaneously. Just making them exclusive is enough.

compiler/one-cmds/one-infer
yunjayh: From this source, never happened.. While designing the function structures, I tried to keep each function deals with every condition it can meet. And that makes this situation. 
Is it dropping the code quality to check the validity or threat twice or more dropping code quality?

mhs4670go: yeah, unnecessary condition'd better be removed. Actually, https://github.com/Samsung/ONE/pull/9158#discussion_r882327750 comment will change the overall codes.

compiler/one-cmds/one-infer
yunjayh: I have no preference between mutually exclusive approach and priority approach. Making them exclusive can make user behavior more clear, but defining priority level may give a chance to user for driver replacement.
How about to print the executable binary's name which is used in `one-infer` if both options are set?

mhs4670go: @yunjayh Seems good:) Maybe using warning would be good, either.

yunjayh: > this is tricky. as you wrote, if the `-d` is invalid we can search for `-b`. or as `-d` is invalid we can raise error and stop, or show error message and continue.
for the user, although `-d` is invalid and the tool continues, user may not notice there is an error. and this may confuse users. (or some users may think the tool is clever)
for me, I would go with raise error and stop.

_Originally posted by @seanshpark in https://github.com/Samsung/ONE/pull/9175#discussion_r883139663_


Let's go with exclusive relation between two options! 

`-d` | `-b` | log
-- | -- | :-- 
o | o | `Error: -d and -b options are mutually exclusive. Please use only one of them`
o | x | \-
x | o | \-
x | x | `Error: Following arguments are required. {-d/--driver \| -b/--backend}, ...`

yunjayh:
Tests are #9175 
```console
$ ./test/runtestall.sh 
one-infer_001.test SUCCESS
one-infer_002.test SUCCESS
one-infer_003.test SUCCESS
one-infer_004.test SUCCESS
one-infer_005.test SUCCESS
one-infer_neg_001.test SUCCESS
one-infer_neg_002.test SUCCESS
one-infer_neg_003.test SUCCESS
one-infer_neg_004.test SUCCESS
ALL TESTS PASSED!
```

seanshpark (dismissed):
LGTM

compiler/one-cmds/one-infer
```diff
@@ -0,0 +1,215 @@
...
+    if getattr(args, 'backend'):
+        backend_base = getattr(args, 'backend') + '-infer'
+        for cand in backends_list:
+            if ntpath.basename(cand) == backend_base:
+                executable = cand
+                return executable
+        missing.append(backend_base)
+
+    if not executable:
+        raise FileNotFoundError(' or '.join(missing) + ' not found')
```
mhs4670go, line 187:
```suggestion
    executable = None

    if _utils._is_valid_attr(args, 'driver'):
        executable = _search_backend_driver(getattr(args, 'driver'))

    if _utils._is_valid_attr(args, 'backend'):
        backend_base = getattr(args, 'backend') + '-infer'
        for cand in backends_list:
            if ntpath.basename(cand) == backend_base:
                executable = cand
                break

    if not executable:
        raise FileNotFoundError(' or '.join(missing) + ' not found')
```

1. AFAIK, just calling `getattr` may cause an error if `args` doesn't have `driver` attribute. Is it okay?
2. They've become mutually exclusive option. So, `missing` variable seems not needed anymore.

compiler/one-cmds/one-infer
```diff
@@ -0,0 +1,215 @@
...
+def _get_backends_list():
+    """
+    [one hierarchy]
+    one
+    ├── backends
+    ├── bin
+    ├── doc
+    ├── include
+    ├── lib
+    └── test
```
mhs4670go, line 45:
```suggestion
    ├── include
    ├── lib
    ├── optimization
    └── test
```

compiler/one-cmds/one-infer
```diff
@@ -0,0 +1,215 @@
...
+
+def _search_backend_driver(driver):
+    """
+    [one hierarchy]
+    one
+    ├── backends
+    ├── bin
+    ├── doc
+    ├── include
+    ├── lib
```
mhs4670go, line 83: ditto

compiler/one-cmds/one-infer
seanshpark: @mhs4670go , if `driver` has higher priority (as I know) is your suggestion work as expected? or maybe my understanding is incorrect?

mhs4670go: @seanshpark No, above suggestion is based on the codes where both options are mutually exclusive.

yunjayh: I've applied the suggested changes :D

mhs4670go (approved):
LGTM

yunjayh:
@seanshpark Could you please take a look? :smile:

PR #9165, opened 2022-05-24
[circle-eval-diff] Check output signature in init

jinevening:
This checks output signature in init to reduce duplicate checks.

### DIFF
--- a/compiler/circle-eval-diff/src/CircleEvalDiff.cpp
+++ b/compiler/circle-eval-diff/src/CircleEvalDiff.cpp
@@ -27,6 +27,22 @@
 namespace
 {
 
+template <typename T> bool same_shape(const T a, const T b)
+{
+  if (a->rank() != b->rank())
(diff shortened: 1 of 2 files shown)


compiler/circle-eval-diff/src/CircleEvalDiff.cpp
```diff
@@ -27,6 +27,22 @@
 namespace
 {
 
+template <typename T> bool same_shape(const T a, const T b)
```
llFreetimell, line 30: Can I ask you why template is needed?
With this PR, i think all of the parameters are `luci::CircleNode` :D

jinevening: I just copied this code from another file. It is not necessary to be a template. Thanks for finding.

I've updated the code.

llFreetimell (approved):
LGTM

Issue #9155, opened 2022-05-23
[nnpackage] Support multiple model packaging on model2nnpkg.sh


hseok-oh:
### What?

Let's support multiple model packaging on `model2nnpkg.sh` script as experimental feature.

### Why?

Currently, `model2nnpkg.sh` allows one model packaging only. But package spec allow multiple model. (It is independent with partitioned model)

We cannot use multiple model package on runtime yet, so let's support this feature as experimental.

hseok-oh:
### How to get multiple model?

Option 1) All remain arguments are model files
```
./model2nnpkg.sh -o out model1.circle model2.circle
```

Option 2) Introduce new option for model files (ex. `-m`)
- If only one argument remain, last argument is model file (backward compatibility)

Option 2-1) All argument between option `-m` and new option are model files
```
./model2nnpkg.sh -m model1.circle model2.circle -o out
```
Option 2-2) Use option `-m` for each model file
```
./model2nnpkg.sh -m model1.circle -m model2.circle -o out 
```

### Package name

Option 1) Use 1st model file name if package name argument (option `-p`) is not passed 
Option 2) Always require package name argument if multiple model file

hseok-oh:
PR #9167: 
- Model input: select option 1 - all remain arguments are model files
- Name: select option 1 - use 1st model file name if package name argument (option `-p`) is not passed

glistening:
As I mentioned in https://github.com/Samsung/ONE/pull/9167#issuecomment-1135435386, I don't have clear idea and even I am not sure it is good to extend model2nnpkg.sh for multiple models support. When I first wrote `model2nnpkg.sh`, it is a tool for just a single model. I am thinking yet another tool for packaging multiple models and connection information, and nnpkg input/output information into a nnpkg. It may get the input not from command line, but from a kind of configurations or the output from other tool (e.g. `circle-partitioner`). But I would like to decide after getting our assumption and requirements and so on. I hope it can be more specific after on-coming discussions (including offline).

glistening:
See https://github.com/Samsung/ONE/pull/9167#discussion_r881066114 and so on for discussion and conclusion about how to get multiple configs.

For short,
```
$ model2nnpkg -c c1 c2 c3 -m m1 m2 m3
```

PR #9167, opened 2022-05-24
[tools/model2nnpkg] Package multiple model

hseok-oh:
This commit updates model2nnpkg.sh to package multiple models.

---

Related issue: #9155

### DIFF
--- a/tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
+++ b/tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
@@ -5,26 +5,33 @@ set -eu
 progname=$(basename "${BASH_SOURCE[0]}")
 outdir="."
 name=""
-config=""
-config_src=""
+configs_src=()
+models_src=()
+configs_str=""
 models_str=""
 types_str=""
 
 usage() {
-  echo "Usage: $progname [options] modelfile..."
+  echo "Usage: $progname [options]"
   echo "Convert modelfile (tflite, circle or tvn) to nnpackage."
   echo ""
   echo "Options:"
   echo "    -h   show this help"
   echo "    -o   set nnpackage output directory (default=$outdir)"
-  echo "    -p   set nnpackage output name (default=[modelfile name])"
-  echo "    -c   provide configuration file"
+  echo "    -p   set nnpackage output name (default=[1st modelfile name])"
+  echo "    -c   provide configuration files"
+  echo "         pass . to skip providing configuration for matching model"
+  echo "    -m   provide model files"
+  echo ""
+  echo "         (Will be deprecated: if there is one remain parameter, that is model file)"
   echo ""
   echo "Examples:"
-  echo "    $progname add.tflite                  => create nnpackage 'add' in $outdir/"
-  echo "    $progname -o out add.tflite           => create nnpackage 'add' in out/"
-  echo "    $progname -o out -p addpkg add.tflite => create nnpackage 'addpkg' in out/"
-  echo "    $progname -c add.cfg add.tflite       => create nnpackage 'add' with add.cfg"
+  echo "    $progname -m add.tflite                  => create nnpackage 'add' in $outdir/"
+  echo "    $progname -o out -m add.tflite           => create nnpackage 'add' in out/"
+  echo "    $progname -o out -p addpkg -m add.tflite => create nnpackage 'addpkg' in out/"
+  echo "    $progname -c add.cfg -m add.tflite       => create nnpackage 'add' with add.cfg"
+  echo ""
+  echo "(Will be deprecated: if there is one remain parameter, that is model file)"
   exit 1
 }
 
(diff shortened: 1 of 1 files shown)


glistening:
I still don't have big picture about multiple models. For example, I have no idea how the connection information (which is not merely connection information, according to our discussion under going it may contains nnpkg inputs and outputs information also) will be added in nnpkg.

glistening:
From offline meeting, I heard that:

- connection information will be given by file perhaps with another option (e.g. `-c`)
- ONE FE tools usually don't get params from cfg except for `one-build` or `one-cc`.

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
```diff
@@ -7,6 +7,8 @@ outdir="."
 name=""
 config=""
 config_src=""
+models_str=""
+types_str=""
 
 usage() {
   echo "Usage: $progname [options] modelfile"
```
glistening, line 14: What about updating usage also? e.g. modelfile1 ... modelfileN

glistening: Or 
```suggestion
  echo "Usage: $progname [options] modelfile..."
```

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
```diff
@@ -43,29 +45,42 @@ done
...
+
+  extension=${modelfile##*.}
+
+  if [[ "$models_str" == "" ]]; then
+    models_str="\"$modelfile\""
+    types_str="\"$extension\""
+  else
+    models_str="$models_str, \"$modelfile\""
+    types_str="$types_str, \"$extension\""
+  fi
```
glistening, line 77: (optional) If you prefer _if_-less version like me,
```suggestion

  models_str="$models_str$delim \"$modelfile\""
  types_str="$types_str$delim \"${modelfile##*.}\""
  delim=","
i
```
Also, you need `delim=""` before entering _for_-loop.

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
```diff
@@ -81,8 +96,12 @@ cat > "$outdir"/"$name"/metadata/MANIFEST <<-EOF
   "minor-version" : "2",
   "patch-version" : "0",
   "configs"     : [ "$config" ],
-  "models"      : [ "$modelfile" ],
-  "model-types" : [ "$extension" ]
+  "models"      : [ $models_str ],
```
glistening, line 99: It is a comment about `configs`, not `models`. I cannot point out `configs` line.
How can we set multiple configs? If we can pass multiple models, we also need to provide a way to pass multiple configs. There may be several options:

```
Option 1: 

$ model2nnpkg -c c1 c2 c3 -m m1 m2 m3

We need to update parsing logic. It requires an option before model files.

Option 2: 

$ model2nnpkg --start-config c1 c2 c3 --end-config m1 m2 m3

model2nnpkg does not support long option name.

Option 3:

$ model2nnpkg -c c1:c2:c3 m1 m2 m3

Option 4:

$ model2nnpkg -c c1:c2:c3 -m m1:m2:m3
```

glistening:
(out of scope) In addition, I would like to remove `.sh` extension from `model2nnpkg.sh`. When I wrote this tool, it was just an internal tool. But as I understand, it is a part of release. We don't need to expose the implementation language. `model2nnpkg` would be enough.

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
hseok-oh: @glistening My preference: 1 > 4 > 3 > 2

glistening: I also like option 1. At firs,t I thought it will break the usage, but we may add this option without breaking previous behavior for most cases (e.g. one model input only).

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
```diff
@@ -43,29 +45,37 @@ done
...
-if [ $# -ne 1 ]; then
-  >&2 echo "error: wrong argument (no argument or too many arguments)."
+if [ $# -eq 0 ]; then
+  >&2 echo "error: wrong argument (no model argument)."
   >&2 echo "For help, type $progname -h"
   exit 1
 fi
 
-modelfile=$(basename "$1")
+for modelpath in "$@"
```
glistening, line 54:
```suggestion
delim=""
for modelpath in "$@"
```

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
glistening: I am okay for adding `-m` and `-c` options later in another PR. Let's merge this after making pr-checker happy. : )

glistening:
@hseok-oh I tested with the following command, and got an error.
```
$ tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh -m inception_v3.tflite model.circle
tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh: line 53: 4: unbound variable
tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh: line 53: 4: unbound variable
tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh: line 66: [: missing `]'
error: wrong argument (no model argument).
For help, type model2nnpkg.sh -h
```

glistening:
```
$ tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh -m inception_v3.tflite model.circle
model2nnpkg.sh: Generating nnpackage inception_v3 in .

$ tree inception_v3
inception_v3
├── inception_v3.tflite
├── metadata
│   └── MANIFEST
└── model.circle

$ cat inception_v3/metadata/MANIFEST
{
  "major-version" : "1",
  "minor-version" : "2",
  "patch-version" : "0",
  "configs"     : [  ],
  "models"      : [ "inception_v3.tflite", "model.circle" ],
  "model-types" : [ "tflite", "circle" ]
}
```
```
$ touch c1 c2
$ tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh -m inception_v3.tflite model.circle -c c1 c2
model2nnpkg.sh: Generating nnpackage inception_v3 in .

$ tree inception_v3
inception_v3
├── inception_v3.tflite
├── metadata
│   ├── MANIFEST
│   ├── c1
│   └── c2
└── model.circle

$ cat inception_v3/metadata/MANIFEST
{
  "major-version" : "1",
  "minor-version" : "2",
  "patch-version" : "0",
  "configs"     : [ "c1", "c2" ],
  "models"      : [ "inception_v3.tflite", "model.circle" ],
  "model-types" : [ "tflite", "circle" ]
}
```
```
$ tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh inception_v3.tflite
model2nnpkg.sh: Generating nnpackage inception_v3 in .

$ tree inception_v3
inception_v3
├── inception_v3.tflite
└── metadata
    └── MANIFEST

$ cat inception_v3/metadata/MANIFEST
{
  "major-version" : "1",
  "minor-version" : "2",
  "patch-version" : "0",
  "configs"     : [  ],
  "models"      : [ "inception_v3.tflite" ],
  "model-types" : [ "tflite" ]
}
```

glistening (approved):
LGTM

hseok-oh:
One unexpected result case: one "-c" parameter, but no "-m" option

```
$ tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh -c c1 model.circle
```

hseok-oh:
I've removed echo for debugging.

tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh
hseok-oh: Ive updated

glistening:
> One unexpected result case: one "-c" parameter, but no "-m" option
> 
> ```
> $ tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh -c c1 model.circle
> tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh: line 115: models_src[0]: unbound variable
> ```

I would interpret this as two configs and no model at all, but it seems someone may consider this as one config and one model without `-m` option. I think it would be better to give warning or error for this usage.

One more unusual case, consider someone wants to two models (m1, m2) and one config (c2). How it can be done?
I thought 
```
$ model2nnpkg -m m1 m2 -c "" c2
```
may be a way. But current implementation does silently like the below:

```
$ tools/nnpackage_tool/model2nnpkg/model2nnpkg.sh -m inception_v3.tflite model.circle -c "" c2
model2nnpkg.sh: Generating nnpackage inception_v3 in .

$ cat inception_v3/metadata/MANIFEST
{
  "major-version" : "1",
  "minor-version" : "2",
  "patch-version" : "0",
  "configs"     : [ "c2" ],
  "models"      : [ "inception_v3.tflite", "model.circle" ],
  "model-types" : [ "tflite", "circle" ]
}
```

I am not saying that let's handle all possible unusual input combinations, but would like to limit the expected inputs.

For example,

- the number of configs should be same to the number of models or zero. (we could lessen the limitation later).
- provide `-m` explicitly if the user uses `-c` option
...

hseok-oh:
>One unexpected result case: one "-c" parameter, but no "-m" option

I found solution for this

>But current implementation does silently like the below:

One option is to define skip mark (ex. `.` => `model2nnpkg -m m1 m2 -c . c2`)

glistening:
> One option is to define skip mark (ex. `.` => `model2nnpkg -m m1 m2 -c . c2`)

Okay. It could be another solution. My suggestion was not to allow this because there is not seeming requirements at this moment, and skip characters may make user it complicated than needed. 

Anyway, I am okay both for 1) skip with `.` and 2) not-defining at this moment.

hseok-oh:
>Anyway, I am okay both for 1) skip with . and 2) not-supporting at this moment.

OK. Then don't support this feature now and think later if we need.

glistening (approved):
LGTM 👍

PR #9173, opened 2022-05-25
[luci] Support Logistic in ForwardReshapeToUnaryOpPass

jinevening:
This supports Logistic in ForwardReshapeToUnaryOpPass.


seanshpark:
What is the background of this change?

jinevening:
> What is the background of this change?

The below pattern appears in our in-house model after ConvertNCHWToNHWC.

```
Reshape->Logistic->Reshape-> ...
```

We can remove the first Reshape if we forward Reshape.

```
Logistic->Reshape->Reshape->..
(Apply RemoveRedundantReshapePass)
Logistic->Reshape->..
```

seanshpark (approved):
LGTM

PR #9189, opened 2022-05-27
[infra] Fix build script bug

hseok-oh:
This commit resolve build fail by mounting on docker container on docker version >= 20
When "TENSORFLOW_PREFIX" and "ONNXRUNTIME_PREFIX" has no value, "-d" test return true.

### DIFF
--- a/infra/scripts/docker_build_nncc.sh
+++ b/infra/scripts/docker_build_nncc.sh
@@ -27,13 +27,13 @@ else
 fi
 
 # prepare tensorflow
-if [ -d $TENSORFLOW_PREFIX ]; then
+if [ -n "$TENSORFLOW_PREFIX" ]; then
   DOCKER_OPTS+=" -v $TENSORFLOW_PREFIX:/opt/tensorflow"
   CONFIG_OPTIONS+=" -DTENSORFLOW_PREFIX=/opt/tensorflow"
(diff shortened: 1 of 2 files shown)


infra/scripts/docker_build_nncc.sh
```diff
@@ -27,13 +27,13 @@ else
 fi
 
 # prepare tensorflow
-if [ -d $TENSORFLOW_PREFIX ]; then
+if [ -n "$TENSORFLOW_PREFIX" ]; then
```
glistening, line 30: I think the original author wants to check the directory exists also. Maybe we need two condition for empty value and directory check.

seanshpark (approved):
LGTM

infra/scripts/docker_build_nncc.sh
mhs4670go: I think so. How about just add double quotes?
```bash
if [ -d "$TENSORFLOW_PREFIX" ];
```

hseok-oh: I thought same things, but I removes directory check because we can mount docker volume (not directory).

mhs4670go (approved):
LGTM

glistening (approved):
LGTM

Issue #9195, opened 2022-05-29
[circle-eval-diff] Analysis warning


seanshpark:
> Return value of a function 'circle::GetModel' is dereferenced at CircleEvalDiff.cpp:63 without checking for null, but it is usually checked for this function (6/7).

seanshpark:
```cxx
auto module = luci::Importer().importModule(circle::GetModel(model_data.data()));
```

@jinevening , PTAL

seanshpark:
will close this issue after next report is OK

seanshpark:
fixed

PR #9196, opened 2022-05-29
[circle-eval-diff] Do null check for circle

jinevening:
This performs null check for circle model.

---
Related to: https://github.com/Samsung/ONE/issues/9195


seanshpark (approved):
LGTM

PR #9193, opened 2022-05-27
DRAFT circle-operator

seanshpark:
on-going draft to introduce circle-operator tool.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #9163, opened 2022-05-24
[circle-operator] Support dump of names for operators


seanshpark:
There is a need to aquire operator names, actually output tensor name of each operators, for one-partition tool to work in ONE-vscode repo.
- implement this in ONE-vscode
   - need to utilize external source already existing but not sure will work and need to find out how-to.
   - done a little code picking... but I don't think this is good...
- implement this in circle-tensordump
   - sure this will work

We may need to read circle files in ONE-vscode repo someday for other reasons.
I'll leave this as technical-debt.

seanshpark:
For https://github.com/Samsung/ONE-vscode/issues/641

seanshpark:
current `circle-tensordump` is release as `-dev` package item... hm...

seanshpark:
maybe better to introduce `circle-opdump`.

seanshpark:
Go with (2) as (3) needs all the feature testing and (4) is strange.

seanshpark:
looks done

PR #9199, opened 2022-05-30
[circle-operator] Introduce driver

seanshpark:
This will introduce circle-operator driver and empty dump run method.

### DIFF
--- a/compiler/circle-operator/driver/Driver.cpp
+++ b/compiler/circle-operator/driver/Driver.cpp
@@ -1,0 +1,117 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Dump.h"
+
+#include <arser/arser.h>
+#include <foder/FileLoader.h>
+#include <fstream>
+
+#include <functional>
+#include <iostream>
+#include <map>
+#include <memory>
+#include <vector>
+#include <string>
+
+#include <signal.h>
+
+void handle_segfault(int signal, siginfo_t *si, void *arg)
+{
+  std::cerr << "ERROR: Failed to load file" << std::endl;
+  exit(255);
+}
+
+int entry(int argc, char **argv)
+{
+  // TODO add option to dump for all sub-graphs
+  arser::Arser arser{
+    "circle-operator allows users to retrieve operator information from a Circle model file"};
+  arser.add_argument("--name").nargs(0).help("Dump operators name in circle file");
+  arser.add_argument("--code").nargs(0).help("Dump operators code in circle file");
+  arser.add_argument("--output_path").help("Save output to file (default output is console)");
+  arser.add_argument("circle").help("Circle file to dump");
+
+  try
+  {
+    arser.parse(argc, argv);
+  }
+  catch (const std::runtime_error &err)
+  {
+    std::cerr << err.what() << std::endl;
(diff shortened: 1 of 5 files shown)


seanshpark:
for #9163, from draft #9193

compiler/circle-operator/src/Dump.cpp
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (c) 2019 Samsung Electronics Co., Ltd. All Rights Reserved
```
mhs4670go, line 2:
```suggestion
 * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
```

compiler/circle-operator/src/Dump.h
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2019 Samsung Electronics Co., Ltd. All Rights Reserved
```
mhs4670go, line 2: ditto

compiler/circle-operator/src/Dump.h
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __DUMP_H__
+#define __DUMP_H__
```
mhs4670go, line 18: Maybe?
```suggestion
#ifndef __CIROPS_DUMP_H__
#define __CIROPS_DUMP_H__
```

compiler/circle-operator/driver/Driver.cpp
```diff
@@ -0,0 +1,117 @@
...
+    oStream = &oFstream;
+  }
+
+  // hook segment fault
+  struct sigaction sa;
+  memset(&sa, 0, sizeof(struct sigaction));
+  sigemptyset(&sa.sa_mask);
+  sa.sa_sigaction = handle_segfault;
+  sa.sa_flags = SA_SIGINFO;
+  sigaction(SIGSEGV, &sa, NULL);
```
mhs4670go, line 88: What does these codes do? I mean when this segment fault happens?

compiler/circle-operator/src/Dump.h
seanshpark: it's executable so uniqueness is not needed.

compiler/circle-operator/driver/Driver.cpp
```diff
@@ -0,0 +1,117 @@
...
+  catch (const std::runtime_error &err)
+  {
+    std::cerr << err.what() << std::endl;
+    std::cerr << arser;
+    return 255;
+  }
+
+  cirops::DumpOption option;
+  option.names = !!arser["--name"];
+  option.codes = !!arser["--code"];
```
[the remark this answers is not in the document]
seanshpark: for double NOT, I wasn't sure we just use `option.names = arser["--name"];`.

compiler/circle-operator/driver/Driver.cpp
seanshpark: Yup. `handle_segfault()` will be called. This is to not show `Segmentation fault` message by the system.

compiler/circle-operator/driver/Driver.cpp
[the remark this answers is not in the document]
seanshpark: we can :) `bool operator[](const std::string &arg_name)`

compiler/circle-operator/src/Dump.h
```diff
@@ -0,0 +1,45 @@
...
+#include <ostream>
+
+namespace cirops
+{
+
+struct DumpOption
+{
+  bool names = false;
+  bool codes = false;
+  bool all_graphs = false;
```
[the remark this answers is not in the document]
seanshpark: Why you ask?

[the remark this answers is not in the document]
mhs4670go: I meant to say it wasn't needed. But looks optional:)

mhs4670go (approved):
LGTM

compiler/circle-operator/src/Dump.h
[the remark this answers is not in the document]
seanshpark: Ah, just to be sure as `all_graphs` is used but not set not cleared.

Issue #9200, opened 2022-05-30
[compiler] protobuf error


seanshpark:
> File "/usr/share/one/bin/venv/lib/python3.8/site-packages/google/protobuf/descriptor.py", line 560, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
TypeError: Descriptors cannot not be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
> 1. Downgrade the protobuf package to 3.20.x or lower.
> 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

Our internal CI fails with this issue

seanshpark:
https://github.com/protocolbuffers/protobuf/issues 10051

Local tested and this works
(1) download grade protobuf python package to `protobuf==3.20.1`
(2) `export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python`

seanshpark:
We can use `venv/bin/postactivate`

seanshpark:
`protobuf==3.20.1` is enough.

hseok-oh:
I found same issue on `common-artifacts` when I try clean build.

PR #9203, opened 2022-05-30
[one-cmds] Fix for protobuf issue

seanshpark:
This will add temporary fix for protobuf issue that requires old version
and environment set.

### DIFF
--- a/compiler/one-cmds/one-prepare-venv
+++ b/compiler/one-cmds/one-prepare-venv
@@ -84,3 +84,7 @@ if [ -n "${EXT_ONNX_TF_WHL}" ]; then
 else
   ${VENV_PYTHON} -m pip ${PIP_OPTIONS} install onnx-tf==${VER_ONNX_TF}
 fi
+
+# NOTE refer https://github.com/protocolbuffers/protobuf/issues/10051
+# TODO remove this when issue is resolved
+${VENV_PYTHON} -m pip install --upgrade protobuf==3.20.1


seanshpark:
for #9200

seanshpark:
I'll check with using `venv/bin/postactivate`

lemmaa:
Looks good to me PR itself.

By the way, how about adding our own `.prescript` and `.postscript` modules to solve this kind of problem?

```
#!/bin/bash
source ${SCRIPT_PATH}/.prescript
.
.
.
source ${SCRIPT_PATH}/.postscript
```

hseok-oh (approved):

seanshpark:
> By the way, how about adding our own .prescript and .postscript modules to solve this kind of problem?

That I think is refactoring and it would be better to do in anther PR.

mhs4670go:
How about follow first one of the warning?

I mean, adding `install protobuf==3.20.0` to `one-prepare-venv`. Because it is more simple.

hseok-oh:
Can we use `tensorflow==2.8.2` ?

https://github.com/tensorflow/tensorflow/releases/tag/v2.8.2

>Add an upper bound for `protobuf` in `setup.py` since `protobuf` after version 3.20 is currently incompatible with TensorFlow

seanshpark:
> Can we use tensorflow==2.8.2

Not sure, our onnx-tf is based on 2.8.0.
We need some tests that may take time and I don't have time for this at this moment.

seanshpark:
> How about follow first one of the warning?
> I mean, adding install protobuf==3.20.0 to one-prepare-venv. Because it is more simple.

I don't understand this.
Please let me know if something is wrong.

lemmaa (approved):

compiler/one-cmds/one-prepare-venv
```diff
@@ -84,3 +84,9 @@ if [ -n "${EXT_ONNX_TF_WHL}" ]; then
 else
   ${VENV_PYTHON} -m pip ${PIP_OPTIONS} install onnx-tf==${VER_ONNX_TF}
 fi
+
+# NOTE refer https://github.com/protocolbuffers/protobuf/issues/10051
+# TODO remove this when issue is resolved
+python3.8 -m pip install --upgrade protobuf==3.20.1
```
seanshpark, line 90: I think I need to revise this...

seanshpark:
previous my local test was incorrect. we only need to limit protobuf version to make it work.

hseok-oh (approved):

mhs4670go (approved):
LGTM

lemmaa (approved):
LGTM

Issue #9194, opened 2022-05-27
[one-cmds] Add docs/one-partition.1


seanshpark:
`one-partition.1` document is missing in `infra/debian/compiler/docs` folder.
Let's add this file.

seanshpark:
@mhs4670go , is this file generated by some tool? or hand-written?
I'v little knowledge about debian documents; how do I compile this file for preview?

mhs4670go:
You can use `help2man`.

```bash
$ help2man build/compiler/tflite2circle/tflite2circle 
.\" DO NOT MODIFY THIS FILE!  It was generated by help2man 1.47.6.
.TH TFLITE2CIRCLE "1" "May 2022" "tflite2circle version 1.21.0" "User Commands"
.SH NAME
tflite2circle \- manual page for tflite2circle version 1.21.0
.SH SYNOPSIS
.B tflite2circle
[\fI\,-h\/\fR] [\fI\,--version\/\fR] [\fI\,-V\/\fR] \fI\,tflite circle\/\fR
.SH DESCRIPTION
What tflite2circle does: tflite2circle is a Tensorflow lite to circle model converter
.PP
[Positional argument]
tflite          Source tflite file path to convert
circle          Target circle file path
.PP
[Optional argument]
\fB\-h\fR, \fB\-\-help\fR      Show help message and exit
\fB\-\-version\fR       Show version information and exit
\fB\-V\fR, \fB\-\-verbose\fR   output additional information to stdout or stderr
.SH COPYRIGHT
Copyright \(co 2020\-2021 Samsung Electronics Co., Ltd. All Rights Reserved
Licensed under the Apache License, Version 2.0
https://github.com/Samsung/ONE
.SH "SEE ALSO"
The full documentation for
.B tflite2circle
is maintained as a Texinfo manual.  If the
.B info
and
.B tflite2circle
programs are properly installed at your site, the command
.IP
.B info tflite2circle
.PP
should give you access to the complete manual.

```

seanshpark:
done

PR #9204, opened 2022-05-30
[Draft][one-cmds] Introduce `tflite-infer` for `one-infer`

yunjayh:
This commit introduces `tlifte-infer` tool which helps to infer
tflite model and get the result data. It can be used by `one-infer`
in `one-cmds`.

### DIFF
--- a/NOTE.md
+++ b/NOTE.md
@@ -1,0 +1,95 @@
+### Introduce `tflite-infer` tool
+
+It's used to infer tflite model and get the result data.
+`tflite-infer` code itself is included in [#9122](https://github.com/Samsung/ONE/pull/9122) too. This pr includes tests, manpage, cmake codes.
+
+
+### `tflite-infer` I/O file strategy
+
+Working on `tflite-infer`, options for loading and dumping I/O data are the most difficult part to design. Here are the restrictions and requirements that I thinking about.
+
+1. Each I/O data needs to support every `npy`, `h5` and `bin` format. 
+2. After dumping randomly generated data as a file, it should be able to be loaded without any additional process. 
+3. Dumping data should not mess up the working directory with many independent files.
+4. (Optional) Data file should be expandable for multiple data inference, for example, infer multiple images in one `tflite-infer` command.
+
+Here are two command line examples which show the changes between before and after `tflite-infer` called. 
+
+<details><summary>Case of loading npy input data</summary>
+
+```console
+$ tree .
+.
+├── model.tflite
+└── simple_data
+    ├── simple_data.input.0.npy
+    └── simple_data.input.1.npy
+
+$ tflite-infer --loadable model.tflite \
+    --input-spec  npy:simple_data \
+    --dump-input-npy  simple_new \
+    --dump-input-h5   simple_new \
+    --dump-output-npy simple_new \
+    --dump-output-h5  simple_new
+
+$ tree .
+.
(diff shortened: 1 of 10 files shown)


NOTE.md
```diff
@@ -0,0 +1,20 @@
...
+### test design description
+
+Category | Test # | Description
+:--: | :--: | :--
+positive test | P-01 | -h printing test
+positive test |	P-02	|-l --input-spec npy:\<filename> execution & output npy dump value test
+positive test |	P-03	|-l --input-spec h5:\<filename> execution & output h5 dump value test
+positive test |	P-04	|-l --input-spec any input npy&h5 shape/dtype test
+positive test|	P-05	|-l --input-spec non-zero input npy&h5 value, shape/dtype test
+positive test|	P-06	|-l --input-spec positive input npy&h5 value, shape/dtype test
```
yunjayh, line 15: Current on-going local draft's approach:

Use simple add tflite model (which is already quantized to int8) ,pre-generated random input file(both npy and h5) and golden output data file. Those will be prepared while `prepare_test_materials` stage. 
Save the `tflite-infer`'s output data and compare the md5sum hash with golden.

```
Simple add model:
LHS ---\
        +--- ADD --- OUT
RHS ---/
```

infra/debian/compiler/docs/tflite-infer.1
```diff
@@ -0,0 +1,32 @@
+.TH TFLITE-INFER "1" "August 2022" "tflite-profile version 1.17.0" "User Commands"
```
seanshpark, line 1: Please refer https://github.com/Samsung/ONE/issues/9194#issuecomment-1139494662 to generate this file. 

@mhs4670go , can you share the base rule of needs for this file?
that is, do we need doc file for tflite-infer ?

yunjayh: This manpage is working on draft. It's not ready at all, but I'll be really grateful if you give me the base rule or something! :heart:

mhs4670go: There's no base rule. Just posting a comment that shows what the man page looks like when we type `man tflite-infer` in the bash would be helpful to review this file.

seanshpark: @mhs4670go , thanks for the comment.
If the user wants to see the help them the user is curious about the command tool while using.
If the tool itself is not direct command for the user than we can think of providing the help can be optional (nice to have).
@yunjayh , is this `tflite-infer` a tool that a user uses normally? I think it would be a tool for ONE-vscode.

yunjayh: > @yunjayh , is this `tflite-infer` a tool that a user uses normally? I think it would be a tool for ONE-vscode

It's not expected to use direct by user in cli command. However, as `one-infer` being a part of `one-cmds`, I thought the users will be curious about the `tflite-infer`'s command line usage.

seanshpark: > as one-infer being a part of one-cmds

I think adding help for `one-infer` is enough as `one-infer` exist for the user.
- current document files are outdated: version and release year. we need to update them.
- adding `tflite-infer` is OK but this will increae one more file to maintain

But adding the file does not harm much so keeping this is OK.

seanshpark:

PR #9217, opened 2022-05-31
[crew] Fix negative test as NEG

seanshpark:
This will fix negative test with NEG suffix.


seanshpark:
found while working on #9216

mhs4670go (approved):
LGTM

Issue #9212, opened 2022-05-31
[circle-partitioner] semicolon in partition key


seanshpark:
While working with ONE-vscode partition editor, javascript ini parser has unexpected behaviour with `;` in key. (unexpected for me but I think this is by the specification)

For example
`aaa;bbb=ccc` will parse as `aaa=ccc` where `;bbb` is ignored.
to set key as `aaa;bbb` then we have to surround it with quotation or use esacpe sequence
- `'aaa;bbb'=ccc`
- `aaa\;bbb=ccc`

TODO
- [ ] revise `crew::read_ini` library to support quotation(single/double)
- [x] and escape sequence

seanshpark:
CC @chunseoklee

seanshpark:
looks done

PR #9218, opened 2022-05-31
[crew] Fix to read esc + semicolon

seanshpark:
This will fix to read escape with semicolon sequence.


seanshpark:
for #9212, from draft #9216

mhs4670go (approved):
LGTM

mhs4670go:
Just curiosity, since semicolon is a comment in ini format, it better not to use this as a string. But should it be used?

PR #9222, opened 2022-05-31
[circle-operator] Introduce README

seanshpark:
This will introduce README file for circle-operator.


seanshpark:
for #9163, from draft #9193

PR #9227, opened 2022-06-06
[circle-operator-test] Test for circle-operator

seanshpark:
This will introduce integration test for circle-operator tool.


mhs4670go (approved):
LGTM

PR #9228, opened 2022-06-06
[infra] Note of compiler_modules

seanshpark:
This will add note of compiler_modules.sh file for better understanding.


hseok-oh (approved):
LGTM

PR #9230, opened 2022-06-07
[infra] Introduce one-partition.1

seanshpark:
This will add missing one-partition.1 in deian docs folder.


seanshpark:
for #9194

mhs4670go (approved):
LGTM

PR #9236, opened 2022-06-07
[infra] Add circle-operator to debian install

seanshpark:
This will revise debian install list to include circle-operator tool.


seanshpark:
from draft #9193

mhs4670go (approved):
LGTM

PR #9238, opened 2022-06-07
[one-cmds] Add man page of one-infer

yunjayh:
This commit adds man page for one-infer tool.

Signed-off-by: yunjay-hong <yunjay.hong@samsung.com>

---

The man page will be shown as following image. 

image

---

for: https://github.com/Samsung/ONE/issues/8970
draft: https://github.com/Samsung/ONE/pull/9122

PR #9239, opened 2022-06-07
[infra] Update compiler manpages

seanshpark:
This will update compiler manpages file to include one-partition.


seanshpark:
Noticed from #9238 that I need this change :)

mhs4670go (approved):
LGTM

Issue #9240, opened 2022-06-07
[one-cmds] TensorFlow Probability error


seanshpark:
Error while loading TensorFlow for `one-import`

> ImportError: This version of TensorFlow Probability requires TensorFlow version >= 2.9; Detected an installation of version 2.8.0. Please upgrade TensorFlow to proceed.

seanshpark:
There was protobuf issue and now tensorflow_probability ...

seanshpark:
https://github.com/tensorflow/probability/releases

> 4 Hours ago, TensorFlow Probability 0.17.0

hseok-oh:
@seanshpark Can we resolve issue https://github.com/Samsung/ONE/issues/9154#issuecomment-1149280718 by this?

seanshpark:
> Can we resolve issue https://github.com/Samsung/ONE/issues/9154#issuecomment-1149280718 by this?

I'm not sure.
9154 error is strange. I could not find download failure but there seems to be problem with test material...

seanshpark:
Patch seems working.

PR #9241, opened 2022-06-07
[one-cmds] Fix tensorflow_probability version

seanshpark:
This will fix tensorflow_probability to match TF2.8.0.


seanshpark:
for  #9240

hseok-oh (approved):

PR #9242, opened 2022-06-08
[one-cmds] Give qconfig when quantize_dequantize

jinevening:
This gives qconfig file when doing quantize_dequantize.


jinevening:
For an MPQ setting, qconfig should be considered in the QuantizeDequantizeWeights stage.

mhs4670go (approved):
LGTM

Issue #9246, opened 2022-06-09
[onert] Map bulk operation to trix backend on compile phase


hseok-oh:
### What?

Let's map bulk operation to trix backend on operation mapping table as workaround. (like bcq operation)

### Why?

We don't have feature to pass configuration for each model. So we cannot pass backend mapping for each partitioned model yet.

We need to map bulk operation to trix backend without configuration until we ready to support passing configuration to runtime compiler for each partitioned model.

chunseoklee:
closed by done

PR #9247, opened 2022-06-09
[onert] Map Bulk operation to trix backend

hseok-oh:
Let's use workaround to map bulk operation to trix backend.

---

Related issue: #9246


glistening (approved):
LGTM

PR #9251, opened 2022-06-09
[one-cmds] Use default dtype/granularity in qconfig

jinevening:
This uses default dtype/granularity in qconfig file.

### DIFF
--- a/compiler/one-cmds/one-quantize
+++ b/compiler/one-cmds/one-quantize
@@ -23,6 +23,7 @@ import argparse
 import os
 import sys
 import tempfile
+import json
 
 import utils as _utils
 from utils import Command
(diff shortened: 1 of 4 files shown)


jinevening:
This PR enables to omit `quantized_dtype` and `granularity` option when using `one-quantize`, if qconfig file has `default_quantization_dtype` and `default_granularity`.

compiler/one-cmds/one-quantize
```diff
@@ -218,9 +219,29 @@ def _set_default_values(args):
...
-        setattr(args, 'quantized_dtype', 'uint8')
+        if _utils._is_valid_attr(args, 'quant_config'):
+            # Get quantized_dtype from qconfig file
+            try:
+                with open(getattr(args, 'quant_config')) as f:
+                    qconf = json.load(f)
+                    if 'default_quantization_dtype' in qconf:
+                        setattr(args, 'quantized_dtype', qconf['default_quantization_dtype'])
+            except json.decoder.JSONDecodeError:
+                print('Failed to decode ' + getattr(args, 'quant_config') + '. Please check it is a json file.')
```
jinevening, line 230: This exception is tested by one-quantize_neg_020.test

compiler/one-cmds/tests/one-quantize_013.test
```diff
@@ -0,0 +1,48 @@
...
+
+rm -rf ${outputfile}
+
+# run test without input data
+# quantized_dtype and granularity are not given here
+one-quantize \
+--input_dtype float32 \
+--quant_config one-quantize_013.qconf.json \
+--input_path ${inputfile} \
+--output_path ${outputfile} > /dev/null 2>&1
```
jinevening, line 42: This is what this PR enables.

compiler/one-cmds/one-quantize
seanshpark: what if json file is valid but `default_quantization_dtype` doesn't existing in json ?

jinevening: Then, it will use default quantization_dtype, i.e., uint8, as before.

seanshpark: But `setattr(args, 'quantized_dtype', 'uint8')` is not called... or is there somewhere else?

jinevening: It is called if both quantized_dtype and quant_config are not given. The code is right below the comment (line 232) :)

seanshpark: I think if something like this case
```json
{
    "default_quantization_dtype" : "uint8",
    "layers" : [
        {
            "names" : ["InceptionV3..., ..."],
            "dtype" : "int16",
            "granularity" : "channel"
        }
    ]
}
```
or
```json
{
    "default_quantization_dtype" : "uint8",
    "default_granularities" : "channel",
    "layers" : [
        {
            "names" : ["InceptionV3..., ..."],
            "dtype" : "int16",
            "granularity" : "channel"
        }
    ]
}
```

jinevening: Ah.. Sorry, I misunderstood the comment. I've updated the PR. Thanks!

jinevening:
@seanshpark  I simplified the code as you suggested :) PTAL

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM thank you!

PR #9255, opened 2022-06-10
[one-cmds] Fix `one-infer` man page

yunjayh:
This commit fixes `one-infer` man page to include
enough information about the command itself.

---

Left is master and right is this pr's

image

I think one approval is enough :)


seanshpark (approved):
LGTM

Issue #9224, opened 2022-06-02
[circle-eval-diff] Support random and directory input


mhs4670go:
## What

Let's support random and directory input.

## Why

As of now, `circle-eval-diff` supports only hdf5 input data format. But sometimes, user want to run it without data set. And, directory input is needed as well.

seanshpark:
> But sometimes, user want to run it without data set. 

Q) how is the test done without data set?

mhs4670go:
@seanshpark Well, I thought test can be done because it is about comparing outputs of two model - output diff. But, after offline meeting with @jinevening, random input seems not to be needed because it is a bit useless because the random data has absolutely nothing to do with the data needed for quantization.

PR #9261, opened 2022-06-14
[circle-eval-diff] Support directory input

mhs4670go:
This commit makes `circle-eval-diff` support directory input.

Related test is gonna be posted after this pr is merged.
Related: #9224 

### DIFF
--- a/compiler/circle-eval-diff/src/InputDataLoader.h
+++ b/compiler/circle-eval-diff/src/InputDataLoader.h
@@ -86,7 +86,7 @@ public:
 
 private:
   const std::vector<loco::Node *> _input_nodes;
-  std::vector<Data> _data_set;
+  std::vector<std::string> _data_paths;
 };
 
 std::unique_ptr<InputDataLoader> makeDataLoader(const std::string &file_path,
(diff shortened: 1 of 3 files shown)


compiler/circle-eval-diff/src/InputDataLoader.h
```diff
@@ -73,6 +74,21 @@ class HDF5Loader final : public InputDataLoader
...
+public:
+  DirectoryLoader(const std::string &dir_path, const std::vector<loco::Node *> &input_nodes);
+
+public:
+  uint32_t size(void) const final;
+  Data get(uint32_t data_idx) const final;
+
+private:
+  const std::vector<loco::Node *> _input_nodes;
+  std::vector<Data> _data_set;
```
jinevening, line 89: It seems that all data are loaded at once. Could you change this to load data one by one whenever each is used? AFAIK, the current interface (`size`, `get`) was introduced for that purpose. One possible implementation would be as follows.

1. Save file names of the data in constructor. (for example, in `std::vector<std::string> _data_files`)
2. When `get` is called, open the file for the given index and read data.

jinevening (approved):
LGTM Thank you

PR #9266, opened 2022-06-15
Compile and Execute ModelGraph 

chunseoklee:


hseok-oh:
We are supporting simple type multimodel now.

PR #9271, opened 2022-06-17
[onert/api] Create Compiler in prepare(), not in load...()

glistening:
Currently, Compiler is created immediately in load...().
Let's create Compiler just before compile in prepare().

Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>

Background: https://github.com/Samsung/ONE/pull/9244/files#r899671545

PR #9277, opened 2022-06-17
[onert] Get CompilerOptions out of Compiler

glistening:
CompilerOptions outlives Compiler.
CompilerOptions are prepared before Compiler::compile()
  - nnfw_set_available_backecnds(), nnfw_set/get_config().

Related: https://github.com/Samsung/ONE/pull/9271#issuecomment-1158387870

### DIFF
--- a/runtime/onert/core/include/compiler/Compiler.h
+++ b/runtime/onert/core/include/compiler/Compiler.h
@@ -50,8 +50,17 @@ struct PartialGraphOptions
   std::unordered_map<ir::OperationIndex, ir::SubgraphIndex> index_to_graph;
 };
 
-struct CompilerOptions
+class CompilerOptions
 {
+public:
+  CompilerOptions(const ir::Subgraphs &subgs) { fetchCompilerOptionsFromGlobalConfig(subgs); }
+
+private:
+  // Set default values for CompilerOptions
+  // All these default values should not be fetched from Env, when we stop supporting Android NNAPI.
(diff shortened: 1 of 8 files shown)


chunseoklee (approved):

runtime/onert/core/include/compiler/Compiler.h
```diff
@@ -133,7 +142,7 @@ class Compiler
   // have to add allocate non-constant tensor memory of executors in execution time when each
   // subgraph is called.
   State _state;
-  CompilerOptions _options;
+  CompilerOptions &_options;
```
ragmani, line 145: Why did you change this to a reference member?

glistening: @ragmani `nnfw_set_available_backends()`, `nnfw_set_config()` and `nnfw_get_config()` reads and modifies the compiler option of `nnfw_session`. `CompilerOptions` should be a reference to `nnfw_session`'s `CompilerOptions`.

ragmani: How about change it a pointer member instead?

glistening: I used reference because
1) I don't want null check which is necessary for pointer.
2) Using pointer type requires more changes.

Why do you prefer pointer?

ragmani: It is nothing. Ignore it.

ragmani (approved):

PR #9279, opened 2022-06-17
[draft][onert/api] Remove _compiler from session

glistening:
`_compiler` is only used during compile().
Let's kick off `_compiler` out of nnfw_session.

Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>

### Dear Reviewers
It adds more changes on top of #9277 and #9271.
You only need to read last two commits.

PR #9285, opened 2022-06-20
[onert/api] Don't use compiler from set_backends_per_operation 

glistening:
set_backends_per_operation changes CompilerOptions.
Except for this, It don't need compiler at all.

### Reviewer

It is a part of #9279, which will remove `_compiler` from `nnfw_session`.
It has one additional commit on top of #9271. Please see the last commit only.


glistening:
It is ready for review, but it has `No merge` labels because it is waiting for #9271 merging.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1218,9 +1218,14 @@ NNFW_STATUS nnfw_session::output_tensorindex(const char *tensorname, uint32_t *i
...
-  }
-  _compiler->set_backend_from_str(backend_settings);
+
+  if (!isStateModelLoaded())
+    return NNFW_STATUS_INVALID_STATE;
+
+  // Backend for all
+  auto &ms_options = _coptions->manual_scheduler_options;
+  onert::compiler::setBackendMap(ms_options, *_subgraphs, std::string{backend_settings});
+
```
glistening, line 1229: Copied from `Compiler::set_backend_from_str`.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1218,9 +1218,14 @@ NNFW_STATUS nnfw_session::output_tensorindex(const char *tensorname, uint32_t *i
 NNFW_STATUS nnfw_session::set_backends_per_operation(const char *backend_settings)
 {
   if (backend_settings == NULL)
-  {
     return NNFW_STATUS_ERROR;
-  }
-  _compiler->set_backend_from_str(backend_settings);
+
+  if (!isStateModelLoaded())
+    return NNFW_STATUS_INVALID_STATE;
```
glistening, line 1224: `set_backends_per_operation` must be in loaded state like `set_available_backends()`.

runtime/onert/api/src/nnfw_api_internal.h
```diff
@@ -149,6 +149,10 @@ struct nnfw_session
   NNFW_STATUS register_custom_operation(const std::string &id, nnfw_custom_eval eval_func);
   NNFW_STATUS input_tensorindex(const char *tensorname, uint32_t *index);
   NNFW_STATUS output_tensorindex(const char *tensorname, uint32_t *index);
+  /**
+   * @brief   Set backends from string-encoded mappings from operation index to backend type (cpu,
+   * acl_cl)
+   */
```
glistening, line 155: Copied from `Compiler.h`.

runtime/onert/core/src/compiler/Compiler.cc
```diff
@@ -82,7 +82,13 @@ void verboseOptions(compiler::CompilerOptions &options)
   VERBOSE(Compiler) << "fp16_enable              : " << options.fp16_enable << std::endl
                     << std::noboolalpha;
 }
+} // namespace
 
+namespace onert
+{
+
+namespace compiler
+{
```
glistening, line 91: Now, `setBackendMap` is in namespace onert::compiler to be used in `nnfw_api_internal.cc`.

runtime/onert/core/include/compiler/Compiler.h
```diff
@@ -78,6 +78,9 @@ class CompilerOptions
   util::TracingCtx *tracing_ctx; //< Profiling information
 };
 
+// TODO: Make this as ManualSchedulerOptions member function
+void setBackendMap(compiler::ManualSchedulerOptions &ms_options, const ir::Subgraphs &subgs,
+                   const std::string &str);
```
glistening, line 83: It was in unnamed namespace, but it is exposed for `nnfw_api_internal.cc.`

runtime/onert/api/src/nnfw_api_internal.h
```diff
@@ -149,6 +149,10 @@ struct nnfw_session
   NNFW_STATUS register_custom_operation(const std::string &id, nnfw_custom_eval eval_func);
   NNFW_STATUS input_tensorindex(const char *tensorname, uint32_t *index);
   NNFW_STATUS output_tensorindex(const char *tensorname, uint32_t *index);
+  /**
+   * @brief   Set backends from string-encoded mappings from operation index to backend type (cpu,
+   * acl_cl)
```
hseok-oh, line 154:
```suggestion
   * @brief   Set backends from string-encoded mappings from operation index to backend type 
   *          (cpu, acl_cl)
```

runtime/onert/core/include/compiler/Compiler.h
```diff
@@ -40,6 +40,10 @@ enum class State
 
 struct ManualSchedulerOptions
 {
+public:
+  void setBackendMap(const ir::Subgraphs &subgs, const std::string &str);
```
hseok-oh, line 44: I don't want to break field-only struct `ManualSchdulerOptions`. Do you have any reason to define here?

runtime/onert/api/src/nnfw_api_internal.h
```diff
@@ -149,6 +149,10 @@ struct nnfw_session
   NNFW_STATUS register_custom_operation(const std::string &id, nnfw_custom_eval eval_func);
   NNFW_STATUS input_tensorindex(const char *tensorname, uint32_t *index);
   NNFW_STATUS output_tensorindex(const char *tensorname, uint32_t *index);
+  /**
+   * @brief   Set backends from string-encoded mappings from operation index to backend type
```
hseok-oh, line 153:
```suggestion
   * @brief   Set backends from string-encoded mappings from operation index to backend type
```

runtime/onert/core/include/compiler/Compiler.h
glistening: I moved because it looks more readable. I also have question. Why do you want to keep `ManualSchedulerOptions` as field-only? It already has several c++ members, it is not a POD. It is already ordinary c++ class except that it has `public` access by default.

runtime/onert/api/src/nnfw_api_internal.h
glistening: @hseok-oh format-check failed after committing your suggestion. Let me revise the description. I want to repeated `from`.

glistening: Ah, it is your 2nd suggestion to fix. The problem was the trailing space. I see.

hseok-oh (approved):

runtime/onert/core/include/compiler/Compiler.h
hseok-oh: Not special reason. In my memory, we decided to use struct as field-only, but now it is broken on many place. Ignore my comment.

chunseoklee (approved):

PR #9288, opened 2022-06-20
[mio-tflite260] Update readme with obsolete

seanshpark:
This will update readme with note with this module is obsolete.


seanshpark:
@mhs4670go , PTAL

mhs4670go (approved):
LGTM

PR #9292, opened 2022-06-21
[onert] Add IR for multiple models nnpackage

glistening:
It is another approach of #9244.

#9244 has the following path:

`ModelGraph` -> `Models` -> `Model` -> `Subgraphs` -> `Graph`

I would like to reduce the depth by merging `Model` and `Subgraphs`.

`ModelGraph` -> `Models` -> (`Model`/ `Subgraphs`) -> `Graph`

To minimize diff, at first, I keep `Subgraphs` in 1st commit of this PR.

That is, two new members about model input and output is now in `Subgraphs`.

But I would like to rename `Subgraphs` to `Model` because:

- `tflite` and `circle` schema also uses `Model` for model itself.
- our loader API `loadModel()` returns `Subgraphs` while other loader API returns the same type. It would be consistent to return `Model`
- `Subgraphs` has a collection of `Graph`, which also doesn't look good.

It is 2nd commit of this PR.

Related: #9206

PR #9293, opened 2022-06-21
[onert] Rename subgraphs to model

glistening:
It renames `Subgraphs` to `Model`.

Related: https://github.com/Samsung/ONE/pull/9292#issue-1277842096


runtime/onert/core/include/ir/Graph.h
```diff
@@ -134,10 +134,10 @@ class Graph
   Operands &operands() { return _operands; } // TODO Remove this non-const accessor
   const Operations &operations() const { return _operations; }
   Operations &operations() { return _operations; }
-  const std::shared_ptr<Subgraphs> &subgraphs() const { return _subgraphs; }
-  std::shared_ptr<Subgraphs> &subgraphs() { return _subgraphs; }
+  const std::shared_ptr<Model> &subgraphs() const { return _model; }
```
ragmani, line 137: How about renaming this function name too?

runtime/onert/frontend/nnapi/wrapper/ANeuralNetworksCompilation.h
```diff
@@ -40,7 +40,7 @@ struct ANeuralNetworksCompilation
   }
 
 private:
-  std::shared_ptr<onert::ir::Subgraphs> _subgraphs;
+  std::shared_ptr<onert::ir::Model> _subgraphs;
```
ragmani, line 43: Here too.

glistening:
I added two commits. one for @ragmani's suggestion, another has more changes about missing ones.

glistening:
@Samsung/one_onert Rebased. PTAL.

runtime/onert/core/include/ir/Graph.h
```diff
@@ -172,13 +172,13 @@ class Graph
   OperandIndexSequence _outputs;
   std::unordered_map<std::string, IOIndex> _name_to_input;
   std::unordered_map<std::string, IOIndex> _name_to_output;
-  // Child subgraphs
-  std::shared_ptr<Subgraphs> _subgraphs;
+  // model for child subgraphs
+  std::shared_ptr<Model> _model;
```
hseok-oh, line 176: Hmm. IMHO, this field may not used any more. It may be used before `SubGraphs`(`Model` in this PR) is introduced. 
@ragmani Do you know about this?

glistening: @hseok-oh Thanks for comment. I also have same question about this field. That said, what about removing this line in another PR when it turned out unnecessary? It is beyond this PR (mechanical renaming).

hseok-oh: Right. And I confirmed that cpu backend and interpreter test are passed when I removed this field and related code.

ragmani: I can't remember exactly. If there is another way to find child subgraphs without this field and this field is not used anymore, you can remove this field.

ragmani (approved):
LGTM

runtime/onert/core/include/ir/Graph.h
hseok-oh: Let's check on #9308

Issue #9281, opened 2022-06-20
[onert] Move tracing context into executor map


hseok-oh:
### What?

Let's move ownership tracing context (`TracingCtx`) into executor map (`ExecutorMap`) instead of session (`nnfw_session`)

### Why?

Now tracing context is generated when model is loaded, and it is owned by session. But tracing context is compilation result, and it is used on executor. So it is nature to move tracing context's ownership to compilation result: executor map.

Draft: #9280

hseok-oh:
Introducing `TracingCtx` Issue: https://github.com/Samsung/ONE/issues/4901

hseok-oh:
Done

PR #9294, opened 2022-06-21
[onert] Remove TracingCtx in CompilerOption

hseok-oh:
This commit removes TracingCtx in CompilerOption.

---

Related issue: https://github.com/Samsung/ONE/issues/9281
Draft: https://github.com/Samsung/ONE/pull/9280

### DIFF
--- a/runtime/onert/core/src/compiler/ExecutorFactory.h
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.h
@@ -38,6 +38,7 @@ public:
 
 public:
   exec::IExecutor *create(std::unique_ptr<compiler::LoweredGraph> lowered_graph,
+                          const util::TracingCtx *tracing_ctx,
                           const compiler::CompilerOptions &options,
                           const std::shared_ptr<exec::ExecutorMap> &executor_map);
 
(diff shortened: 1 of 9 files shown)


runtime/onert/core/src/compiler/ExecutorFactory.h
```diff
@@ -38,6 +38,7 @@ class ExecutorFactory
 
 public:
   exec::IExecutor *create(std::unique_ptr<compiler::LoweredGraph> lowered_graph,
+                          const util::TracingCtx *tracing_ctx,
```
glistening, line 41: (beyond this PR) I wondered why `IExecutor` requires `CompilerOptions`. It seems that the only required member is `options.executor`. I think it is an option for Executor, not Compiler. : )

glistening (approved):
LGTM

runtime/onert/core/src/compiler/ExecutorFactory.h
hseok-oh: `IExecutor` doesn't requires `CompilerOptions`, but `IExecutor` factory requires because it is part of compilation. Factory uses to check executor type, tracing/profiling setting.

glistening: @hseok-oh You're right. I think `ExecutorFactory` requires only `options.executor` (string for executor name like "LINEAR", ...) We may replace `CompilerOptions` to `const string exectur_name`.

Issue #9290, opened 2022-06-21
[one-cmds] Introduce data post processing option to `one-infer`


yunjayh:
### WHAT 

Let's introduce post-process stage by option in `one-infer`.

### WHY

To support *value-test* between two different `one-infer` results, the inference result data should be well formatted.
The standard format of data was decided to use `h5` because `h5` is a type which can contain multiple data with specific hierarchy and some other reasons. (There is a discussion in [this link](https://github.com/Samsung/ONE/issues/9248))

On the other hand, `one-infer` cannot guarantee various backend drivers emits well defined format data. 

So, `one-infer` needs to solve that problem with `post-process` option which run data conversion after driver execution.

There is previous discussion about **why post processing?** in #9282.

### HOW

Brief example usage will be like this.

```console
$ ls
model.npu aaa.0.npy aaa.1.npy

# Note that [commands] run model.npu with aaa.0.npy and aaa.1.npy as input data 
# and emit output data as bbb.0.npy ...
# ./npy2h5.py can be any script which converts the npu-infer's result 
# to preset standard format of h5
$ one-infer -d npu-infer --post-process ./npy2h5.py -- [commands]
Inference complete!

$ ls
model.npu aaa.0.npy aaa.1.npy bbb.0.npy bbb.1.npy ccc.h5
```

IMHO `--post-process ./npy2h5.py` can be replaced to `--post-process "./npy2h5.py aaa.0.npy aaa.1.npy bbb.0.npy ccc.h5"` depends on how `npy2h5.py` has been implemented. 

Any questions and suggestions are gratefully welcome!

jinevening:
The idea looks good to me. Could you show the example of `npy2h5.py`? I'm curious about the format of the post-process script.

yunjayh:
> The idea looks good to me. Could you show the example of `npy2h5.py`? I'm curious about the format of the post-process script.

Of course. When I'm ready, I'll make and add it to my draft and ping you. It will be great to having one sample post-process script. :smile: Thank you!

seanshpark:
close as inactive for long time

PR #9295, opened 2022-06-21
[DRAFT][one-cmds] Introduce post-process option to `one-infer`

yunjayh:
This commit adds `post-process` option to `one-infer` which can be executed
after the backend driver runs.

---

- [x] Add tests
- [x] Fix man page

---

for #9290


compiler/one-cmds/one-infer
```diff
@@ -206,6 +209,16 @@ def main():
     # run backend driver
     _utils._run(infer_cmd, err_prefix=ntpath.basename(driver_path))
 
+    # run post process script if it's given
+    if _utils._is_valid_attr(args, 'post_process'):
+        # Currently, it is only supported to use one-venv.
+        # TODO: support local python env or specify python path
```
mhs4670go, line 215: Since `one-cmds` script runs under the virtual env, I think it is hard to call local python. If you find the way, it would be good. If not, these seems to be possible solutions.

1. Get `requirement.txt` from user needed for running given script. And, install them with `pip` by `subprocess`. Maybe it needs to be done with `sudo` privilege. 
2. Get python executable path from user when the `one-cmds` is being installed. 
3. List the libraries frequently used for data processing and install them in advance.

Third option looks good to me as of now.

yunjayh: > Since `one-cmds` script runs under the virtual env, I think it is hard to call local python.

I thought so, but when I tried to run `subprocess.run(['python', 'some_python_script.py'])` in venv, the local python is executed for the subprocess. It seems to run a subprocess with a new bash. 
The problem is how to make user specify the python path. 

Anyway, I prefer the third option, too. :smile:

infra/debian/compiler/docs/one-infer.1
```diff
@@ -1,11 +1,11 @@
-.TH ONE-INFER "1" "June 2022" "one-profile version 1.21.0" "User Commands"
```
yunjayh, line 1: I regenerated man page by `help2man` tool. There were few errors before, and it will be fixed from this patch

yunjayh:
I think it'd be better to describe why this option is needed to `one-infer` at some place, for example, inside the help message. 

And how can user recognize the standard format of I/O data is h5 with some pre-defined hierarchy? (Do they need to know?)

PR #9296, opened 2022-06-21
DRAFT: simplify circle2circle options

seanshpark:
simplify circle2circle options with a method.


seanshpark:
done long time ago...

PR #9310, opened 2022-06-22
[circle-eval-diff] Catch h5 exception

mhs4670go:
This commit catches h5 exception. This is for static analysis tool.


mhs4670go:
Static analysis tool still warns that `makeDataLoader` could throw H5 related exception but it isn't being caught. But this code does that. I think it is false positive case.

jinevening (approved):
LGTM

PR #9311, opened 2022-06-22
[infra] Change using pthread option

hseok-oh:
Instead of using old convention - linking pthread (`-lpthread`), let's use compiler build option (`-pthread`) by using cmake FindThread.


hseok-oh:
Difference between `-lpthread` and `-pthread`
https://stackoverflow.com/questions/23250863/difference-between-pthread-and-lpthread-while-compiling

glistening (approved):
LGTM

PR #9308, opened 2022-06-22
[onert] Remove model reference in graph

hseok-oh:
This commit removes model reference field in graph.

Signed-off-by: Hyeongseok Oh <hseok82.oh@samsung.com>

PR #9312, opened 2022-06-22
[onert] Use copy of graph in interpreter

hseok-oh:
This commit updates interpreter to use copy of graph, not original graph reference.
Sometimes interpreter executor cannot refer original graph.

---

This PR is previous commit for #9308


glistening (approved):
LGTM

PR #9320, opened 2022-06-23
[circle-eval-diff] Use try-catch to handle H5 related errors

llFreetimell:
This commit adds try-catch statements to handle H5 related errors.

---

Expanded from #9310

### DIFF
--- a/compiler/circle-eval-diff/src/InputDataLoader.cpp
+++ b/compiler/circle-eval-diff/src/InputDataLoader.cpp
@@ -96,10 +96,18 @@ namespace circle_eval_diff
 HDF5Loader::HDF5Loader(const std::string &file_path, const std::vector<loco::Node *> &input_nodes)
   : _input_nodes{input_nodes}
 {
-  using HDF5Importer = dio::hdf5::HDF5Importer;
+  try
+  {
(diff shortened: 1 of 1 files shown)


compiler/circle-eval-diff/src/InputDataLoader.cpp
```diff
@@ -201,18 +217,26 @@ std::unique_ptr<InputDataLoader> makeDataLoader(const std::string &file_path,
...
+  try
   {
-    case InputFormat::H5:
+    switch (format)
     {
-      return std::make_unique<HDF5Loader>(file_path, input_nodes);
+      case InputFormat::H5:
+      {
+        return std::make_unique<HDF5Loader>(file_path, input_nodes);
+      }
```
mhs4670go, line 227: FYI, try-catch statement should be smaller because below `DirectoryLoader` is not related to h5 things.
```suggestion
      case InputFormat::H5:
      {
        try{
          return std::make_unique<HDF5Loader>(file_path, input_nodes);
          }
          catch(const H5::Exception &e)
          {
            H5::Exception::printErrorStack();
            throw std::runtime_error("HDF5 error occurred.");
          }
      }
```

llFreetimell:
Close this because it is proved that this PR has no effect :(

PR #9326, opened 2022-06-23
[one-cmds] Introduce `post-process` option to `one-infer`

yunjayh:
This commit adds `post-process` option to `one-infer` which will execute
given command after running backend driver inference.

---

draft: #9295

### DIFF
--- a/compiler/one-cmds/one-infer
+++ b/compiler/one-cmds/one-infer
@@ -109,7 +109,7 @@ def _search_backend_driver(driver):
 
 
 def _get_parser(backends_list):
-    infer_usage = 'one-infer [-h] [-v] [-C CONFIG] [-d DRIVER] [-b BACKEND] [--] [COMMANDS FOR BACKEND DRIVER]'
+    infer_usage = 'one-infer [-h] [-v] [-C CONFIG] [-d DRIVER] [-b BACKEND] [--post-process POST_PROCESS] [--] [COMMANDS FOR BACKEND DRIVER]'
     parser = argparse.ArgumentParser(
         description='command line tool to infer model', usage=infer_usage)
 
@@ -128,6 +128,9 @@ def _get_parser(backends_list):
             backends_name) + ')'
     backend_help_message = 'backend name to use ' + backends_name_message
     parser.add_argument('-b', '--backend', type=str, help=backend_help_message)
+
+    post_process_help_message = 'post processing script to convert I/O data to standard format'
+    parser.add_argument('--post-process', type=str, help=post_process_help_message)
 
     return parser
 
@@ -156,7 +159,7 @@ def _parse_arg(parser):
     # split by '--'
     args = [list(y) for x, y in itertools.groupby(argv, lambda z: z == '--') if not x]
 
-    # one-infer [-h] [-v] [-C CONFIG] [-d DRIVER] [-b BACKEND] -- [COMMANDS FOR BACKEND DRIVER]
+    # one-infer [-h] [-v] [-C CONFIG] [-d DRIVER] [-b BACKEND] [--post-process POST_PROCESS] -- [COMMANDS FOR BACKEND DRIVER]
     if len(args):
         infer_args = args[0]
         infer_args = parser.parse_args(infer_args)
@@ -207,6 +210,14 @@ def main():
 
     # run backend driver
     _utils._run(infer_cmd, err_prefix=ntpath.basename(driver_path))
+
+    # run post process script if it's given
+    if _utils._is_valid_attr(args, 'post_process'):
+        # NOTE: the given python script will be executed by venv of ONE
+        python_path = sys.executable
+        post_process_command = [python_path] + getattr(args,
+                                                       'post_process').strip().split(' ')
+        _utils._run(post_process_command, err_prefix='one-infer')
 
 
 if __name__ == '__main__':
--- a/compiler/one-cmds/tests/one-infer_neg_005.test
+++ b/compiler/one-cmds/tests/one-infer_neg_005.test
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+# Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# one-infer with invalid post process script
+
+filename_ext="$(basename -- $0)"
+filename="${filename_ext%.*}"
+
+trap_err_onexit()
+{
+  return_code=$?
+  if grep -q "dummy-infer dummy output!!!" "${filename}.log"; then
+    # Case of succeed of inference driver but error after it
+    if [ "$return_code" -ne "0" ]; then
+      echo "${filename_ext} SUCCESS"
+      exit 0
+    fi
+  fi
+
+  echo "${filename_ext} FAILED"
+  rm -rf ../bin/dummy-infer
+  exit 255
+}
+
+trap trap_err_onexit ERR
+
+inputfile="sample.tvn"
+
+if [[ ! -s "${inputfile}" ]]; then
+  touch ${inputfile}
+fi
+
+# copy dummy-infer to bin folder
+cp dummy-infer ../bin/dummy-infer
+
+# run test
+one-infer -b dummy --post-process "./one-infer-test-post-process.py" -- ${inputfile} > ${filename}.log 2>&1
+
+rm -rf ../bin/dummy-infer
+echo "${filename_ext} FAILED"
+exit 255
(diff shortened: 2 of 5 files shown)


compiler/one-cmds/one-infer
```diff
@@ -129,6 +129,9 @@ def _get_parser(backends_list):
     backend_help_message = 'backend name to use ' + backends_name_message
     parser.add_argument('-b', '--backend', type=str, help=backend_help_message)
 
+    post_process_help_message = 'python script and arguments which will be executed after backend driver'
```
yunjayh, line 132: Is this approach better for user? 

```suggestion
    post_process_help_message = 'post processing script to convert I/O data to standard format'
```

infra/debian/compiler/docs/one-infer.1
```diff
@@ -1,13 +1,11 @@
 .TH ONE-INFER "1" "June 2022" "one-infer version 1.21.0" "User Commands"
 .SH NAME
-\fBone-infer\fR \- infer model file
+one-infer \- manual page for one-infer version 1.21.0
```
seanshpark, line 3: plz separate this in another PR

compiler/one-cmds/one-infer
```diff
@@ -109,7 +109,7 @@ def _search_backend_driver(driver):
 
 
 def _get_parser(backends_list):
-    infer_usage = 'one-infer [-h] [-v] [-C CONFIG] [-d DRIVER] [-b BACKEND] [--] [COMMANDS FOR BACKEND DRIVER]'
+    infer_usage = 'one-infer [-h] [-v] [-C CONFIG] [-d DRIVER | -b BACKEND] [--post-process PYTHON SCRIPT AND ARGUMENT] [--] [COMMANDS FOR BACKEND DRIVER]'
```
seanshpark, line 112: `[-d DRIVER | -b BACKEND]` change seems not related with the title of this PR

compiler/one-cmds/one-infer
```diff
@@ -206,6 +209,16 @@ def main():
     # run backend driver
     _utils._run(infer_cmd, err_prefix=ntpath.basename(driver_path))
 
+    # run post process script if it's given
+    if _utils._is_valid_attr(args, 'post_process'):
+        # Currently, it is only supported to use one-venv.
```
seanshpark, line 214: what is `one-venv` ? if you meant virtual environment then I think this comment is needless as all the commands can only run in virtual environment.

compiler/one-cmds/one-infer
```diff
@@ -206,6 +209,16 @@ def main():
...
+        # TODO: support local python env or specify python path
```
seanshpark, line 215: I'm not good at python venv but technically is this possible?

compiler/one-cmds/tests/one-infer_006.py
```diff
@@ -0,0 +1,14 @@
+# This script gets one argument and print it
+
+import sys
+from pathlib import Path
```
seanshpark, line 4: I think this file is excluded in format checker.
@yunjayh , can you also look at format checker, if not a huge burden. if not I'll check.

compiler/one-cmds/tests/one-infer_006.py
```diff
@@ -0,0 +1,14 @@
...
+
+def main():
+    if len(sys.argv) < 2:
+        filepath = Path(sys.argv[0])
+        sys.exit("Usage: " + filepath.name + " [Word to print]")
+    word = sys.argv[1]
+    print(word)
+
+if __name__ == '__main__':
+    main()
```
seanshpark, line 14: optional) how about return with value, or exit with a value so that caller can check with it for success or failure.

compiler/one-cmds/one-infer
yunjayh: https://github.com/Samsung/ONE/blob/75a30387f36ebbceafe509e755e97687e82d8c95/compiler/one-cmds/one-infer#L216-L217
Yes if the `python_path` is changed to user's python path, it can be run by local python or another venv that user specified. (Even if it is set to `python3`, local python binary will be run)

compiler/one-cmds/tests/one-infer_006.py
yunjayh: I'm a little bit confused about this comment. Which part do you want to return with a value? 
- Adding a process exit code to check process' succeed? Or 
- add return value to `main` function to check main function's validity?

seanshpark: Well, as I'm not good at python, what I thought was like this called from bash or other process
```
python3 one-infer_006.py
ret = $?
echo $ret
# check ret value
```
Could you please how process exit code or return value of main shows?

yunjayh: > Could you please how process exit code or return value of main shows?

I'm sorry I couldn't catch what you saying.. :sob:

Current `one-infer_006.py` prints the first given argument if it successfully executed. 
To check whether a process which executes `one-infer_006.py` as its subprocess, we can check the printed line includes the first argument. (As you know, it's how the `one-infer_006.test` works). I'm not sure how to make test codes for both options you mentioned, exit code or return value.

yunjayh: I got your intention through the off-line talk. :wink:
Since `python3 some_python_script` will emit exit code as zero if the script ended successfully, and non-zero for else, it would be better to check its exit code for both valid and invalid post processing script.

`one-infer -d dummy-infer --post-process "[valid-command]" -- [infer command]`'s exit code will be zero.
`one-infer -d dummy-infer --post-process "[invalid-command]" -- [infer command]`'s exit code will be non-zero.

One positive test and one negative test will be added. :smile:

compiler/one-cmds/one-infer
```diff
@@ -205,6 +208,15 @@ def main():
...
     _utils._run(infer_cmd, err_prefix=ntpath.basename(driver_path))
+    
+    # run post process script if it's given
+    if _utils._is_valid_attr(args, 'post_process'):
+        # NOTE: the given python script will be executed by venv of ONE
+        python_path = sys.executable
+        post_process_command = [python_path] + getattr(args, 'post_process').strip().split(' ')
+        print(f'one-infer: [post-process] {post_process_command[1]} is started', flush=True)
+        _utils._run(post_process_command, err_prefix='one-infer')
+        print(f'one-infer: [post-process] {post_process_command[1]} is done', flush=True)
```
yunjayh, line 219: Since `_utils._run` flushes stdout buffer after writing it, the log has been messed up. So I put `flush=True` flag.

compiler/one-cmds/tests/one-infer_neg_005.test
```diff
@@ -0,0 +1,53 @@
...
+
+if [[ ! -s "${inputfile}" ]]; then
+  touch ${inputfile}
+fi
+
+# copy dummy-infer to bin folder
+cp dummy-infer ../bin/dummy-infer
+
+# run test
+one-infer -b dummy --post-process "./one-infer_006.py" -- ${inputfile} > ${filename}.log 2>&1
```
yunjayh, line 49: Here is a problem. If this line emits error and exit with non-zero exit code, then `trap trap_err_onexit ERR` catches it and get into `trap_err_onexit` function. Then I lose the exit code of `one-infer` command. I'm not sure the current test approach is proper way. Could you give me a comment? @seanshpark

compiler/one-cmds/one-infer
```diff
@@ -205,6 +208,15 @@ def main():
 
     # run backend driver
     _utils._run(infer_cmd, err_prefix=ntpath.basename(driver_path))
+    
+    # run post process script if it's given
+    if _utils._is_valid_attr(args, 'post_process'):
+        # NOTE: the given python script will be executed by venv of ONE
+        python_path = sys.executable
+        post_process_command = [python_path] + getattr(args, 'post_process').strip().split(' ')
+        print(f'one-infer: [post-process] {post_process_command[1]} is started', flush=True)
```
seanshpark, line 217: I think `print` can be omitted if not really necessary.

seanshpark: `is started` is somewhat strange. maybe `has started` or `will start`, ...
or just `begin`, `start`

compiler/one-cmds/tests/one-infer_006.py
```diff
@@ -0,0 +1,14 @@
+# This script gets one argument and print it
```
seanshpark, line 1: how about renaming this as `one-infer-test-post-process.py` ?

yunjayh: That's great 👍

compiler/one-cmds/one-infer
yunjayh: I'll remove the `print`s.

yunjayh:
I applied the new format checker script and changed few minor formats. PTAL

seanshpark (approved):
LGTM thank you!

compiler/one-cmds/one-infer
```diff
@@ -208,6 +211,14 @@ def main():
     # run backend driver
     _utils._run(infer_cmd, err_prefix=ntpath.basename(driver_path))
 
+    # run post process script if it's given
+    if _utils._is_valid_attr(args, 'post_process'):
+        # NOTE: the given python script will be executed by venv of ONE
+        python_path = sys.executable
+        post_process_command = [python_path] + getattr(args,
+                                                       'post_process').strip().split(' ')
```
mhs4670go, line 219:
```suggestion
        post_process_command = [python_path] + getattr(args, 'post_process').strip().split()
```
(trivial) `split` has default parameter - space

mhs4670go (approved):
LGTM

compiler/one-cmds/one-infer
yunjayh: I'd like to specify the token. :) thank you for pointing it. :smile:

PR #9245, opened 2022-06-08
Draft: Support compiler build on Ubuntu 20.04

hseok-oh:
Try to build compiler on ubuntu 20.04.

Signed-off-by: Hyeongseok Oh <hseok82.oh@samsung.com>

PR #9329, opened 2022-06-24
[coco] Remove return with std::move

hseok-oh:
This commit removes return with std::move(...) for local unique pointer.
It needs on old compiler version, but we don't need this any more because we are using c++14 feature.

---

Draft: #9245


seanshpark (approved):
LGTM thank you!

PR #9331, opened 2022-06-24
[luci] Remove return with std::move

hseok-oh:
This removes return with std::move(...) for local unique pointer.
It needs on old compiler version, but we don't use old compiler any more.

---

Same with https://github.com/Samsung/ONE/pull/9329
Draft: https://github.com/Samsung/ONE/pull/9245


seanshpark (approved):
LGTM thank you!

jinevening (approved):
LGTM

Issue #9328, opened 2022-06-24
[onert] Revisit CompilerOptions for multiple models nnpkg


glistening:
- There is a single `CompilerOptions` in `nnfw_session`.
- `CompilerOptions` constructor requires `Model`.

It was okay till now because there is only 1 Compiler and 1 Model in 1 nnfw_session.

But now things changed.

- nnpackage may have several models.
- there may be config for each model, not only global config.

Related: #9206, #9254

glistening:
If `CompilerOptions` will be still an option for `nnfw_session` (not for `Model`) in the possiblity of multiple models.

I think `CompilerOptions` should not take `Model` in constructor, it must take `NNPkg`.

I investigated why `CompilerOptions` requires `Model`. It is used in 

```
void CompilerOptions::fetchCompilerOptionsFromGlobalConfig(const ir::Model &model)
{
...
    auto map_str = util::getConfigString(util::config::OP_BACKEND_MAP);
    ms_options.setBackendMap(model, map_str);
```

```
void ManualSchedulerOptions::setBackendMap(const ir::Model &model, const std::string &str)
{
...
    auto key = static_cast<uint32_t>(std::stoi(key_str));

    model.at(ir::SubgraphIndex{0})
      ->operations()
      .at(ir::OperationIndex{key}); // Check if exist, or this wil throw
```
It is used to check whether the operation index exists or not in 1st model (hardcoded).

(To be continued)

glistening:
`ManualSchedulerOptions::setBackendMap` is used in two ways.

1) From GlobalConfig
2) From `nnfw_set_backends_per_operation`, which is declared in `nnfw_experimental.h`

How will we handle them in possibility of multiple models nnpkg?

Option 1: `setBackendMap` works only when # of model = 1
- These values are ignored or warned when it is set for multiple model nnpkg

Option 2: `setBackendMap` applied to 1st model (as it does silently in hardcoded code).
- It applies the given backed map only for 1st model.
- Its value has "`OP_BACKEND_MAP=\"0=acl_cl;1=cpu;2=acl_cl\"`"

I would choose option 1.

I don't consider in this comment how it will work when `configs` is set for each model.

We may need another revisit to support multiple configs.

glistening:
@hseok-oh, @chunseoklee

Q. Will we create Compiler for each model? Or 1 Compiler will handle several models?

chunseoklee:
> I think 1).

[...]

PR #9334, opened 2022-06-24
[draft][onert] Revisit CompilerOptions and ManualSchedulerOptions

glistening:
There has been only 1 `CompilerOptions` in `nnfw_session`. But there may be multiple `CompilerOptions` for multiple models.

Related #9328

It simplifies `CompilerOptions` by:

- Introducing `fromGlobalConfig` to show explicitly the values come from GlobalConfig
- Removing `Model` param from CompilerOptions constructor and ManualSchedulerOptions::setBackend.

### DIFF
--- a/runtime/onert/core/src/compiler/Compiler.cc
+++ b/runtime/onert/core/src/compiler/Compiler.cc
@@ -90,7 +90,7 @@ namespace onert
 
 namespace compiler
 {
-void ManualSchedulerOptions::setBackendMap(const ir::Model &model, const std::string &str)
+void ManualSchedulerOptions::setBackendMap(const std::string &str)
 {
   // TODO Support multiple subgraphs for manual scheduling
   auto key_val_list = nnfw::misc::split(str, ';');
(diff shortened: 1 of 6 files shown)


runtime/onert/core/src/compiler/Compiler.cc
```diff
@@ -105,27 +105,31 @@ void ManualSchedulerOptions::setBackendMap(const ir::Model &model, const std::st
...
     const auto &val = key_val.at(1);
     auto key = static_cast<uint32_t>(std::stoi(key_str));
-
+#if 0
+    // Let setBackemdMap to parse str and put key, value pair to map.
+    // Check is done during execution.
     model.at(ir::SubgraphIndex{0})
       ->operations()
       .at(ir::OperationIndex{key}); // Check if exist, or this wil throw
+#endif
```
glistening, line 114: It is the only place that uses `model` to check the operation index exists or not.
The same check is done in `ManualScheduler::schedule()`.

https://github.com/Samsung/ONE/blob/a0fbb99ee34ae112e97b2f1a5d2ba165f2f3908c/runtime/onert/core/src/compiler/ManualScheduler.cc#L90

PR #9335, opened 2022-06-26
[onert/api] create CompilerOptions in try block

glistening:
CompilerOptions construction may throw exceptions (bad_alloc, out_of_range, ...).
Let's move all load statements into try block.

Related: #9328, https://github.com/Samsung/ONE/pull/9334/files#r905917895

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -220,14 +220,14 @@ NNFW_STATUS nnfw_session::load_circle_from_buffer(uint8_t *buffer, size_t size)
   try
   {
     _model = onert::circle_loader::loadModel(buffer, size);
+    _coptions = onert::compiler::CompilerOptions::fromGlobalConfig();
+    _state = State::MODEL_LOADED;
   }
   catch (const std::exception &e)
   {
     std::cerr << "Error during model loading : " << e.what() << std::endl;
     return NNFW_STATUS_ERROR;
   }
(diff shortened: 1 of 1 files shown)


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -220,14 +220,14 @@ NNFW_STATUS nnfw_session::load_circle_from_buffer(uint8_t *buffer, size_t size)
   try
   {
     _model = onert::circle_loader::loadModel(buffer, size);
+    _coptions = std::make_unique<onert::compiler::CompilerOptions>(*_model);
```
hseok-oh, line 223: It's strange that model loading is failed by compile option. IMO, we need to move validation check.

glistening: Yes, it is a part of https://github.com/Samsung/ONE/pull/9334/files#r905917895 I mentioned in this PR main comment.

glistening: Anyway, I think it is better to enclose whole load statements for other possible exceptions.

glistening:
@hseok-oh, @Samsung/one_onert I've rebased after #9337 and #9338. PTAL

runtime/onert/api/src/nnfw_api_internal.cc
hseok-oh: Anyway I agree with you that we need to move this into try block. I'll think about this later.

hseok-oh (approved):
LGTM

chunseoklee (approved):

PR #9337, opened 2022-06-27
[onert] Rename CompilerOptions(Model&) to fromGlobalConfig(Model&)

glistening:
CompilerOptions(Model&) looks like it creates CompilerOptions using
Model. But it may be misleading. CompilerOptions is created from GlobalConfig values.
The parameter(Model&) is only used during ManualScheduler option validation.
`Model` is not stored in CompilerOptions at all also.

I would like to make it clear CompilerOptions are created using
GlobalConfig.

Draft: #9334


ragmani (approved):
LGTM

hseok-oh (approved):
LGTM

PR #9338, opened 2022-06-27
[onert] Remove model from ManualSchedulerOptions::setBackendMap

glistening:
It removes model parameter from ManualSchedulerOptions::setBackendMap
since model is only used to check the existence of operation index,
and same check is done during execution also.
    
After removing parameter from ManualSchedulerOptions, at last,
CompilerOptions::fromGlobalConfig() can kick off model parameter.
    
### For reviewers
Please see 2nd commit only, the 1st commit is to be merged in #9337.
#9337 has to be landed first.

Related: #9328

### DIFF
--- a/runtime/onert/core/src/compiler/Compiler.cc
+++ b/runtime/onert/core/src/compiler/Compiler.cc
@@ -90,7 +90,7 @@ namespace onert
 
 namespace compiler
 {
-void ManualSchedulerOptions::setBackendMap(const ir::Model &model, const std::string &str)
+void ManualSchedulerOptions::setBackendMap(const std::string &str)
 {
   // TODO Support multiple subgraphs for manual scheduling
   auto key_val_list = nnfw::misc::split(str, ';');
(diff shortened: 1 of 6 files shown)


runtime/onert/core/src/compiler/Compiler.cc
```diff
@@ -105,27 +105,24 @@ void ManualSchedulerOptions::setBackendMap(const ir::Model &model, const std::st
     const auto &key_str = key_val.at(0);
     const auto &val = key_val.at(1);
     auto key = static_cast<uint32_t>(std::stoi(key_str));
-
-    model.at(ir::SubgraphIndex{0})
-      ->operations()
-      .at(ir::OperationIndex{key}); // Check if exist, or this wil throw
```
glistening, line 111: Please see https://github.com/Samsung/ONE/pull/9334#discussion_r905917895. The same check is done before `schedule`.

hseok-oh:
Please rebase to resolve conflict

glistening:
@hseok-oh, @Samsung/one_onert PTAL. I've rebased.

hseok-oh (approved):
LGTM

chunseoklee (approved):

Issue #9336, opened 2022-06-27
[circle-operator] Support both code and name option


seanshpark:
Let's revise `circle-operator` tool to support both code and name at the same time.

seanshpark:
for https://github.com/Samsung/ONE-vscode/issues/822

seanshpark:
Operator code does not have `,` in text but name can have any.
We can use `,` to extract code and name.

seanshpark:
done

PR #9339, opened 2022-06-27
[circle-operator] Revise multi output tensor

seanshpark:
This will revise multi output tensor to dump first output.
- this is to fix as node is single object
- add todo for if there is other cases


seanshpark:
related  #9336

mhs4670go:
> this is to fix as node is single object

```cxx
for (const auto output : output_tensors)
```
This for loop seems enough but this case doesn't work with current codes?

seanshpark:
> This for loop seems enough but this case doesn't work with current codes?

When there are multiple outputs, there will be multiple lines of tensor names for one node.
Which I don't think this will correctly work with one-vscode partition editor gui.

mhs4670go (approved):
LGTM

PR #9340, opened 2022-06-27
[onert] Refine DotDumper

ragmani:
This commit refines DotDumper.
  - Seperate the method `dump()` of `DotDumper`
  - Remove class members that might cause the problems with c++ lifetime

### DIFF
--- a/runtime/onert/core/src/dumper/dot/DotDumper.cc
+++ b/runtime/onert/core/src/dumper/dot/DotDumper.cc
@@ -19,6 +19,7 @@
 
 #include "DotDumper.h"
 #include "DotBuilder.h"
+#include "ir/OperandIndexMap.h"
 #include "ir/OperationIndexMap.h"
 #include "backend/Backend.h"
 #include "backend/IConfig.h"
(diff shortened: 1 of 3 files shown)


ragmani:
I checked whether `DotDumper` works well after this commit with a mobilenet model, and it works well.

runtime/onert/core/src/dumper/dot/DotDumper.cc
```diff
@@ -31,97 +32,77 @@ namespace dumper
...
-    }
+    bool showing_cond =
+      level == DotDumper::Level::ALL
+        ? true
+        : !object.isConstant() || (graph.getInputs() + graph.getOutputs()).contains(index);
     if (showing_cond)
     {
+      // This set is not used anywhere.
+      // TODO: Remove this set
       shown_operand_set.add(index);
```
ragmani, line 76: I don't know why this variable exists. Anyway I will remove this variable because it is not used anywhere.

runtime/onert/core/src/dumper/dot/DotDumper.h
```diff
@@ -38,27 +38,28 @@ class DotDumper
...
+   * @brief Dump graph information to dot file as tag name if "GRAPH_DOT_DUMP" is set
+   *
+   * @param[in] graph  The graph that would be used to get operations and operands
+   * @param[in] tag    The name of dot file that would be created
+   * @return N/A
+   */
+  void dump(const ir::Graph &graph, const std::string &tag);
+
+  /**
+   * @brief Dump lowed graph information to dot file as tag name if "GRAPH_DOT_DUMP" is set
```
glistening, line 54:
```suggestion
   * @brief Dump lowered graph information to dot file as tag name if "GRAPH_DOT_DUMP" is set
```

glistening:
@ragmani Personally I like your change to get explicit parameter rather than keeping reference as a member. It is hard to figure out when `DotDumper` is used as lambda, and it gets reference to graph.

After this change, `DotDumper` seems to have only 1 member `level`. Then, what about making `DotDumper::dump` static class function?

ragmani:
> After this change, `DotDumper` seems to have only 1 member `level`. Then, what about making `DotDumper::dump` static class function?

Do you mean to move the member `level` to a parameter of `DotDumper::dump`?

glistening:
> Do you mean to move the member `level` to a parameter of `DotDumper::dump`?

At first, I thought. But I found that `level` comes from `_options`. Then, it would be good to keep it in `DotDumper` though it is the only one. Current change looks good. : )

runtime/onert/core/src/dumper/dot/DotDumper.h
ragmani: I fixed it as you mentioned. Thanks.

glistening (approved):
LGTM. (Frankly, I don't know what is change from your last update (after my approval) : )

ragmani:
Sorry for my force pushing. The last update is just to fix a typo as you mentioned :)

PR #9244, opened 2022-06-08
[onert] Add IR structures for multiple model nnpkg

glistening:
It adds Model, Models, ModelOperandIndex, and ModelOperandIndexSequence.

Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>

Related: #9206 

### For Reviewers
It is draft PR for supporting multiple model nnpkg, based on the draft in #6274.

The difference to #6274:
- This PR does not cover api_internal.
- This PR will change `ModelGraphs` since there were changes after #6274.

- 1st commit in this PR is exactly copy-and-paste of #6274 except for copyright year change 2021 → 2022.

PR #9341, opened 2022-06-27
[onert/api] Get file extension by searching ., not by fixed length

glistening:
load_model_from_modelfile() assumes that its model file will end with
either `.circle` or `.tflite`. It get its extension by getting last 7
characters. But it would be more generatl to find extension by last dot.

Related: https://github.com/Samsung/ONE/pull/9244#discussion_r896447982

I tested this change by running `tflite_comparator 1.tvn`.
Sure, it only succeeds to run `1.tvn`, and fails whille trying `1.tvn` with tflite interpreter. : )
But I confirmed this change works by seeing trix emulator's run log.


hseok-oh (approved):
LGTM

ragmani (approved):
LGTM

PR #9342, opened 2022-06-27
[onert] Revisit session creation

hseok-oh:
This commit revisits session creation.
It introduces factory method for `nnfw_session`. Factory method handles different exception type and return different status.
To prevent using constructor directly, `nnfw_session`'s constructor is changed to private.
To initialize on session creation, `_coption` and `_kernel_registry` initialization is moved into factory method.

---

Related issue: https://github.com/Samsung/ONE/pull/9335#discussion_r906899854

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -210,15 +210,10 @@ NNFW_STATUS nnfw_session::create(nnfw_session **session)
     return NNFW_STATUS_UNEXPECTED_NULL;
 
   // Create session
-  try
+  *session = new (std::nothrow) nnfw_session();
+  if (*session == nullptr)
   {
-    *session = new nnfw_session();
-  }
-  catch (const std::exception &e)
-  {
-    std::cerr << "Error during session creation : " << e.what() << std::endl;
-    *session = nullptr;
-
+    std::cerr << "Error during session creation" << std::endl;
     return NNFW_STATUS_OUT_OF_MEMORY;
   }
 
@@ -231,6 +226,7 @@ NNFW_STATUS nnfw_session::create(nnfw_session **session)
   catch (const std::exception &e)
   {
     std::cerr << "Error during session initialization : " << e.what() << std::endl;
+    delete *session;
     *session = nullptr;
 
     return NNFW_STATUS_ERROR;
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -199,9 +199,44 @@ void fillTensorInfo(nnfw_tensorinfo *ti, const onert::ir::Shape &shape,
 
 nnfw_session::nnfw_session()
   : _model{nullptr}, _coptions{nullptr}, _compiler_artifact{nullptr}, _execution{nullptr},
-    _kernel_registry{std::make_shared<onert::api::CustomKernelRegistry>()}
+    _kernel_registry{nullptr}
 {
   // DO NOTHING
+}
+
+NNFW_STATUS nnfw_session::create(nnfw_session **session)
+{
+  if (session == nullptr)
+    return NNFW_STATUS_UNEXPECTED_NULL;
+
+  // Create session
+  try
+  {
+    *session = new nnfw_session();
+  }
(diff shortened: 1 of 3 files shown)


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -220,7 +255,6 @@ NNFW_STATUS nnfw_session::load_circle_from_buffer(uint8_t *buffer, size_t size)
   try
   {
     _model = onert::circle_loader::loadModel(buffer, size);
-    _coptions = onert::compiler::CompilerOptions::fromGlobalConfig();
```
glistening, line 223: Good! It is no longer needs to be here. `model is gone. : )

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -199,11 +199,46 @@ void fillTensorInfo(nnfw_tensorinfo *ti, const onert::ir::Shape &shape,
...
 
+NNFW_STATUS nnfw_session::create(nnfw_session **session)
+{
+  if (session == nullptr)
+    return NNFW_STATUS_UNEXPECTED_NULL;
+
+  // Create session
+  try
+  {
+    *session = new nnfw_session();
```
glistening, line 215: What about using smart pointer and set `*session` at the end of this function. There may be memory leak if exception is thrown between line 228-229.

glistening (dismissed):
LGTM

ragmani (dismissed):
LGTM

runtime/onert/api/src/nnfw_api_internal.cc
hseok-oh: We need free on catch blocks. But it's hard to use `unique_ptr` because we can't find proper owner of new session's `unique_ptr`.

glistening: I am not saying returning `unique_ptr`, but suggesting use `unique_ptr` as local variable in this funciton, and return raw pointer from this `unique_ptr`. Anyway, I am okay.

glistening (approved):
LGTM. I have to update this code as I will change `_coptions` to be a collection of `CompilerOptions` for #9328. Also, I was thinking of creating a local `unique_ptr`, then return its raw pointer using `release()`.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -353,7 +382,6 @@ NNFW_STATUS nnfw_session::load_model_from_nnpackage(const char *package_dir)
       return NNFW_STATUS_ERROR;
     }
     _model->primary_subgraph()->bindKernelBuilder(_kernel_registry->getBuilder());
-    _coptions = onert::compiler::CompilerOptions::fromGlobalConfig();
```
glistening, line 356: Just note.

- `_coptions` will be a collection (i.e. `std::vector<CompilerOptions>`) for the possibility of multiple model of nnpkg. 
- Two internal `load...` APIs are used for one model test only. Thus, moving this into `nnfw_create` is okay.
- Besides, `load_model_from_nnpackage` may have multiple models, which requires multiple `CompilerOptions`, and it is determined after load. It will be here again.

runtime/onert/api/src/nnfw_api_internal.cc
hseok-oh: >return raw pointer from this `unique_ptr`

My comment means that after returning raw pointer, allocated session by local `unique_ptr` will be deleted, so returned raw pointer will be invalidate.

glistening: @hseok-oh Please find my comment below (I wrote yesterday for approval). I was thinking of `release()`, which doesn't free the pointer.

glistening:
@hseok-oh  Just for your information, `_coptions` creation will be restored in `load_...` for multiple models. Please find the 3rd commits from #9292.

hseok-oh:
@ragmani PTAL

ragmani (approved):

runtime/onert/api/src/nnfw_api_internal.cc
hseok-oh: I understood. Thanks.

PR #9343, opened 2022-06-27
[onert] Refine ShapeValidator

ragmani:
This commit refines ShapeValidator.
  - Remove unnecessary members
  - Add deleted copy & move constructors to avoid copying or moving of ShapeValidator

### DIFF
--- a/runtime/onert/core/src/compiler/ShapeValidator.cc
+++ b/runtime/onert/core/src/compiler/ShapeValidator.cc
@@ -34,73 +34,72 @@ namespace onert
 namespace compiler
 {
 
-ShapeValidator::ShapeValidator(const ir::Graph &graph)
-  : _graph{graph}, _ctx{graph.operands()}, _current_layout{ir::Layout::UNKNOWN}
-{
-}
+ShapeValidator::ShapeValidator(const ir::Graph &graph) : _graph{graph} {}
 
 void ShapeValidator::checkUnaryOp(const ir::Operation &node)
 {
+  const auto &operands = _graph.operands();
   const auto output_index{node.getOutputs().at(0)};
   const auto input_index{node.getInputs().at(0)};
 
-  if (_ctx.at(output_index).info().isDynamic())
+  if (operands.at(output_index).info().isDynamic())
     return;
 
   // Check if I/O shapes match
-  OP_REQUIRES(_ctx.at(output_index).shape() == _ctx.at(input_index).shape());
+  OP_REQUIRES(operands.at(output_index).shape() == operands.at(input_index).shape());
 }
 
 void ShapeValidator::operator()()
 {
-  _current_layout = _graph.layout();
-
   _graph.operations().iterate(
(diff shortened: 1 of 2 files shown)


runtime/onert/core/src/compiler/ShapeValidator.h
```diff
@@ -39,8 +39,13 @@ class ShapeValidator : public ir::OperationVisitor
 public:
   ShapeValidator(void) = delete;
   ShapeValidator(const ir::Graph &graph);
+  ShapeValidator(const ShapeValidator &) = delete;
+  ShapeValidator(ShapeValidator &&) = delete;
+  ~ShapeValidator() = default;
 
 public:
+  ShapeValidator &operator=(const ShapeValidator &) = delete;
```
ragmani, line 47: An implicitly-declared copy assignment operator for this class is defined as `deleted` because of the member `const ir::Graph &_graph;` (const or reference member), but I explicitly declared `deleted copy assignment operator` because of `The Rule of Five`.

chunseoklee:
@ragmani Aside from PR content, should we use "ONE-DCO-1.0-Signed-off-by: Chunseok Lee <chunseok.lee@samsung.com>
" in signing ?

glistening:
@chunseoklee A few days ago, I found other members use `ONE-DCO-...`, instead of `Signed-off...`. Then, I found the [latest guide](https://github.com/Samsung/ONE/wiki/ONE-Developer%27s-Certificate-of-Origin#using-the-signed-off-by-process) says the former is recommended format. Thus, I updated my signing.

ragmani:
I changed the signed-off message.

runtime/onert/core/src/compiler/ShapeValidator.cc
```diff
@@ -34,82 +34,82 @@ namespace onert
...
     return;
 
   const auto ifm_index{node.getInputs().at(ir::operation::BatchToSpaceND::Input::INPUT)};
   const auto block_size_index{
     node.getInputs().at(ir::operation::BatchToSpaceND::Input::BLOCK_SIZE)};
 
-  const auto frontend_layout = _current_layout;
-  const auto input_shape = _ctx.at(ifm_index).shape().asFeature(frontend_layout);
-  const auto output_shape = _ctx.at(ofm_index).shape().asFeature(frontend_layout);
+  const auto frontend_layout = _graph.layout();
```
glistening, line 85: @ragmani Before this change, `_current_layout` is the copy of the value at the `ShapeValidator` creation time. After this change, it will get the value from `_graph`. May I assume there has been no change `_graph.layout()` ? If it is true, your change looks good.

ragmani: Right. `_graph.layout()` have to be not changed since it was set when models was loaded. `_graph.layout()` is the layout information of frontend. The layout information for backend is stored in `LoweredGraph`.

glistening (approved):
LGTM

PR #9344, opened 2022-06-28
[CirOpTest] remove test for both options

seanshpark:
This will remove test for both options as both options will be valid.


seanshpark:
for #9336 , from draft #9345

mhs4670go (approved):
LGTM

PR #9348, opened 2022-06-28
[circle-operator] Enable both name code possible

seanshpark:
This will revise to enable both name and code option possible.


seanshpark:
for #9336, from draft #9345

mhs4670go (approved):
LGTM

PR #9358, opened 2022-06-28
[circle-operator] Test for both code and name

seanshpark:
This will add test for both code and name options provided.


seanshpark:
for #9336, from draft #9345

mhs4670go (approved):
LGTM

PR #9361, opened 2022-06-28
[one-cmds] Add torch URL message

seanshpark:
This will add torch install URL message for debugging for redirecting server.
- this will be removed after tracing is finished.


mhs4670go (approved):
LGTM

Issue #9206, opened 2022-05-30
[onert] Support multiple model nnpkg (loader, ir)


glistening:
Let's support multiple models nnpkg.

- [x] Create a sample multiple model nnpkg
- [x] IR
- [x] Loader
- [x] nnfw_session 

This issue will focus on **frontend** (loader and IR building) of onert.
**Backend** of onert (e.g. executor) will be handled in other issue by other members.
_~Perhaps onert **API** will be covered in another issue by discussing together.~_ ←I have to cover API to check the feasibility of extended IR.

## Related
- #6274 : last year draft by @hseok-oh

glistening:
## Create a sample multiple model nnpkg

I created the prerequisite models using mobilenet v1 uint8 model.

Here is [mv1.mm.zip](https://github.com/Samsung/ONE/files/8850805/mv1.mm.zip):
- mv1.01.circle  ( Conv2D - DepthwiseConv2D ) <sub> <I>I cannot upload screenshot. </I></sub>
  - model by selecting first two ops from mobilnet v1 q8 model
  - `python tools/tflitefile_tool/select_operator.py mobilenet_v1.220207.quantized.random.channel.uint8.circle <( echo 0-1 ) 
mv1.01.tflite`
  - `build/release/nncc/compiler/tflite2circle/tflite2circle mv1.12.{tflite,circle}`
- mv1.0.circle ( Conv2D )
  - model by selecting first op from mv1.01.circle
- mv1.1.circle ( DepthwiseConv2D )
  - model by selecting second op from mv1.01.circle
- mv1.1.tvn ( I don't upload this file. )
  - model by compiling mv1.1.circle

I will add more (e.g. `MANIFEST`) later.

glistening:
I am trying to complete `MANIFEST`.

### A. Combination of #9129 and #9126

#### MANIFEST

Provide a separate json for connection via _**model-connect**_
```
{
  "major-version" : "1",
  "minor-version" : "2",
  "patch-version" : "0",
  "configs"       : [  ],
  "models"        : [ "mv1.0.circle", "mv1.1.tvn" ],
  "model-types"   : [ "circle", "tvn" ],
  "model-connect" : "conn.json"
}
```
#### conn.json
Use _**source**_ and _**conn**_ fields.
```
{
  "source" : {
    "inputs" : [ "0:0:0" ],
    "outputs" : [ "0:0:1"]
  },
  "conn" : [
[...]

PR #9373, opened 2022-07-01
[onert/api] nnfw_session may have multiple CompilerOptions

glistening:
nnfw_session has a container of CompilerOptions because it may have
multiple CompilerOptions if nnpkg has multiple models.

Draft: https://github.com/Samsung/ONE/pull/9292
Related: https://github.com/Samsung/ONE/issues/9206

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -212,7 +212,7 @@ std::unique_ptr<onert::ir::Model> loadModel(const std::string filename,
 } // namespace
 
 nnfw_session::nnfw_session()
-  : _model{nullptr}, _coptions{nullptr}, _compiler_artifact{nullptr}, _execution{nullptr},
+  : _model{nullptr}, _coptions{}, _compiler_artifact{nullptr}, _execution{nullptr},
     _kernel_registry{nullptr}
 {
   // DO NOTHING
@@ -235,7 +235,6 @@ NNFW_STATUS nnfw_session::create(nnfw_session **session)
   try
   {
     (*session)->_kernel_registry = std::make_shared<onert::api::CustomKernelRegistry>();
-    (*session)->_coptions = onert::compiler::CompilerOptions::fromGlobalConfig();
   }
   catch (const std::exception &e)
   {
@@ -265,6 +264,7 @@ NNFW_STATUS nnfw_session::load_circle_from_buffer(uint8_t *buffer, size_t size)
   try
   {
     _model = onert::circle_loader::loadModel(buffer, size);
+    _coptions.push_back(onert::compiler::CompilerOptions::fromGlobalConfig());
     _state = State::MODEL_LOADED;
   }
   catch (const std::exception &e)
@@ -300,6 +300,7 @@ NNFW_STATUS nnfw_session::load_model_from_modelfile(const char *model_file_path)
     _model = loadModel(filename, model_type);
     if (_model == nullptr)
       return NNFW_STATUS_ERROR;
+    _coptions.push_back(onert::compiler::CompilerOptions::fromGlobalConfig());
     _state = State::MODEL_LOADED;
   }
   catch (const std::exception &e)
@@ -368,6 +369,7 @@ NNFW_STATUS nnfw_session::load_model_from_nnpackage(const char *package_dir)
     if (_model == nullptr)
       return NNFW_STATUS_ERROR;
     _model->primary_subgraph()->bindKernelBuilder(_kernel_registry->getBuilder());
+    _coptions.push_back(onert::compiler::CompilerOptions::fromGlobalConfig());
     _state = State::MODEL_LOADED;
   }
   catch (const std::exception &e)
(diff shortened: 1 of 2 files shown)


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -974,7 +976,7 @@ NNFW_STATUS nnfw_session::set_available_backends(const char *backends)
     if (null_terminating(backends, MAX_BACKEND_NAME_LENGTH) == false)
       return NNFW_STATUS_ERROR;
 
-    auto &options = *_coptions;
+    auto &options = *_coptions[0];
```
glistening, line 979: Currently `set_available_backends` assumes there is only one model. We need to extend this API.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1008,7 +1010,7 @@ NNFW_STATUS nnfw_session::set_op_backend(const char *op, const char *backend)
       return NNFW_STATUS_ERROR;
     }
 
-    auto &opcode_to_backend = _coptions->manual_scheduler_options.opcode_to_backend;
+    auto &opcode_to_backend = _coptions[0]->manual_scheduler_options.opcode_to_backend;
```
glistening, line 1013: `set_op_backend` needs to be extended for multiple models.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1097,7 +1099,7 @@ NNFW_STATUS nnfw_session::get_config(const char *key, char *value, size_t value_
   if (!key || !value)
     return NNFW_STATUS_UNEXPECTED_NULL;
 
-  auto &options = *_coptions;
+  auto &options = *_coptions[0];
```
glistening, line 1102: `get_config`

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1027,7 +1029,7 @@ NNFW_STATUS nnfw_session::set_config(const char *key, const char *value)
...
```
glistening, line 1032: `set_config`

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1233,7 +1235,7 @@ NNFW_STATUS nnfw_session::set_backends_per_operation(const char *backend_setting
     return NNFW_STATUS_INVALID_STATE;
 
   // Backend for all
-  auto &ms_options = _coptions->manual_scheduler_options;
+  auto &ms_options = _coptions[0]->manual_scheduler_options;
```
glistening, line 1238: `set_backends_per_operation`

runtime/onert/api/src/nnfw_api_internal.cc
hseok-oh: Or we can use `set_available_backends` to set same backend list for all models, and introduce API for each model (ex. `set_available_backends_for_model(backends, model_idx)`

chunseoklee: Also note that set_available_backends are used in nnstreamer : https://github.com/nnstreamer/nnstreamer/blob/173e999ebb6866b08c71913e1402b3335b7d0e84/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c#L243

glistening: As I wrote in https://github.com/Samsung/ONE/issues/9328#issuecomment-1165166436, I was thinking of keeping the API, but have 1 additional parameter with default value = 0. Anyway, let's discuss in another issue or PR.

glistening:
@hseok-oh, @chunseoklee This PR is about the internal implementation of existing API, not about changing public API itself. Let's discuss more in another PR. Please consider my comments as a note for reviewers for reminder for future.

chunseoklee (approved):

runtime/onert/api/src/nnfw_api_internal.cc
hseok-oh: My comment is also thinking of keeping current API. I thought current `set_available_backend` is used when user want to use specific backend only for package. (and `set_op_backend` same). And I think this PR is proper place to discuss about this.

glistening: @hseok-oh We may continue to discuss more in #9328 where @chunseoklee and I discussed 1 week ago. Please leave your opinion or idea there. This is PR about supporting multiple model w/o breaking existing API and behavior. The API design for getting/setting config for multiple model is another beast in my mind. Currently we don't provide such API at all.

hseok-oh: >The API design for getting/setting config for multiple model is another beast in my mind.

Agree. My suggestion about introducing new API can discuss there.

My comment is we need to decide in this PR about `set_available_backend` and `set_op_backend`'s result if user keep current implementation.

glistening:
@hseok-oh, @chunseoklee I've gathered related comments about API in https://github.com/Samsung/ONE/issues/9328#issuecomment-1172039496. I would like to hear your suggestion and information in place before adding new API or extending the meaning of public APIs.

runtime/onert/api/src/nnfw_api_internal.cc
hseok-oh: By offline discussion, I agree to set model index 0 as default for consistency with other APIs. If we need to support package-level backend setting and op-per-backend setting, we can introduce new API later.

hseok-oh (approved):
LGTM

PR #9378, opened 2022-07-01
[onert/ir] Add fields for model connection and nnpkg IO in NNPkg

glistening:
pkg-inputs, pkg-outputs and model_edges are added.

### DIFF
--- a/runtime/onert/core/include/ir/NNPkg.h
+++ b/runtime/onert/core/include/ir/NNPkg.h
@@ -111,7 +111,7 @@ public:
    * @param[in] index Index of pkg_input to be returned
    * @return OperandDesc at index
    */
-  OperandDesc input(uint32_t index) const { return _pkg_inputs[index]; }
+  const OperandDesc &input(uint32_t index) const { return _pkg_inputs[index]; }
   /**
    * @brief Get pkg_input at index
    *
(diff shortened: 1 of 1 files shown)


runtime/onert/core/include/ir/NNPkg.h
```diff
@@ -70,9 +105,67 @@ class NNPkg
...
+   */
+  void addInput(const OperandDesc &input) { _pkg_inputs.push_back(input); }
+
+  /**
+   * @brief Get pkg_output at index
+   *
+   * @param[in] index Index of pkg_output to be returned
+   * @return OperandDesc at index
+   */
+  OperandDesc output(uint32_t index) const { return _pkg_outputs[index]; }
```
ragmani, line 135:
```suggestion
  const OperandDesc &output(uint32_t index) const { return _pkg_outputs[index]; }
```

glistening: @ragmani I returned copy of `OperandDesc` because I met a compilation error during making this change, and `OperandDesc` is 3 uint32_t, its copy overhead would be small. But I checked there is no compilation error with your suggestion. I've updated. (including `input()`).

ragmani (approved):
LGTM(+0.5), but I don't konw about ModelEdge

glistening:
@hseok-oh, @chunseoklee PTAL

chunseoklee (approved):
Although I hardly imagine how to query, `_model_edges` contains full relevant information in it. LGTM.

hseok-oh:
I think we can add unittest to show how to use. Anyway looks good.

hseok-oh (approved):

glistening:
I also told @chunseoklee last week offline. It may be hard to use to query. It is likely to be updated during backend implementation (compilation and execution).

PR #9382, opened 2022-07-04
Draft: Support compiler test on Ubuntu 20.04

hseok-oh:
Try to test compiler on ubuntu 20.04

### DIFF
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -38,11 +38,14 @@ if(PYTHON_EXECUTABLE MATCHES python3.8)
   set(PYTHON_OVERLAY python3.8)
 endif()
 
+# NOTE refer https://github.com/protocolbuffers/protobuf/issues/10051
+# TODO remove protobuf==3.20.1 when issue is resolved
 add_custom_command(
   OUTPUT ${REQUIREMENTS_OVERLAY_PATH_TF_2_6_0}
   COMMAND ${CMAKE_COMMAND} -E remove -f ${REQUIREMENTS_OVERLAY_PATH_TF_2_6_0}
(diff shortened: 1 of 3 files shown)


compiler/mir/src/Graph.cpp
```diff
@@ -44,9 +44,14 @@ std::vector<Operation *> getSortedNodes(Graph *graph)
   std::deque<Operation *> ready_nodes;
   std::unordered_map<Operation *, std::size_t> num_visited_input_edges;
 
+  for (Operation *op : graph->getInputs())
+  {
+    ready_nodes.push_back(op);
+  }
```
hseok-oh, line 50: I added this because below `getNodes()` return `unordered_set`, and below searching changes input order and test fail on ubuntu 22.04.  (ex. `compiler/mir2loco/src/mir2loco.test.cpp`'s `Add_Test`)

It may related with different std lib implementation.

@seanshpark Is it acceptable update?

compiler/common-artifacts/CMakeLists.txt
```diff
@@ -38,11 +38,14 @@ if(PYTHON_EXECUTABLE MATCHES python3.8)
   set(PYTHON_OVERLAY python3.8)
 endif()
 
+# NOTE refer https://github.com/protocolbuffers/protobuf/issues/10051
+# TODO remove protobuf==3.20.1 when issue is resolved
```
hseok-oh, line 42: https://github.com/Samsung/ONE/issues/9200#issuecomment-1173385092

compiler/mir/src/Graph.cpp
seanshpark: > Is it acceptable update?

`mir` project has not been updated more than a year... I think it is OK
@binarman PTAL

hseok-oh:
All merged

PR #9383, opened 2022-07-04
[onert] Refine StaticShapeinferer

ragmani:
This commit refines StaticShapeInferer.
  - Change StaticShapeInferer so that a StaticShapeInferer is assigned to each subgraph
  - Add observers to call StaticShapeInferers recursively

### DIFF
--- a/runtime/onert/core/include/compiler/StaticShapeInferer.h
+++ b/runtime/onert/core/include/compiler/StaticShapeInferer.h
@@ -28,6 +28,36 @@ namespace onert
 {
 namespace compiler
 {
+/**
+ * @brief Class that observe and update operands.
+ */
+class OperandObserver
+{
+public:
+  /**
+   * @brief Constructor of OperandObserver
+   *
+   * @param operands Operands to be updated
+   */
+  OperandObserver(const std::vector<ir::Operand *> &operands) : _operands{operands} {}
+  /**
+   * @brief Destructor of OperandObserver
+   */
+  virtual ~OperandObserver() = default;
+
+public:
+  /**
+   * @brief Update Shape and some OperandInfo of operands
+   *
+   * @param operands Operands to be updated
+   * @param unpredictable Whether runtime can predict shapes of operands in compilation time
+   */
+  void updateShapes(const std::vector<ir::OperandInfo> &changed_operands_info,
+                    bool unpredictable = false);
+
+private:
+  std::vector<ir::Operand *> _operands;
+};
 
 /**
  * @brief Class to infer shape before running kernels. It does the following:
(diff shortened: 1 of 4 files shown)


chunseoklee:
> Failed in branch mixed

@ragmani 
Not sure that the fail is due to this PR ...

chunseoklee:

runtime/onert/core/include/compiler/StaticShapeInferer.h
```diff
@@ -38,32 +86,42 @@ namespace compiler
...
-   * @param op Operation
-   * @return @c true if op's input or output has any dynamic tensor; @c false otherwise.
    */
-  bool infer(const ir::Operation &op);
+  void update(void);
 
   void dump();
 
 private:
-  void inferSubgraph(ir::SubgraphIndex subg_ind);
```
ragmani, line 65: `bool infer(const ir::Operation &op)` and `void inferSubgraph(ir::SubgraphIndex subg_ind)` are replaced with `void update(void)`

runtime/onert/core/include/compiler/StaticShapeInferer.h
```diff
@@ -128,12 +186,11 @@ class StaticShapeInferer : public ir::OperationVisitor
   void handleSimpleUnaryOp(const ir::Operation &op, const ir::OperandIndex input_idx);
 
 private:
-  const std::unordered_map<ir::SubgraphIndex, std::unique_ptr<compiler::LoweredGraph>>
-    &_lowered_subgs;
-  // _operands and _operations can be changed by controlflow operation
-  ir::Operands &_operands;     // operands of current subgraph
-  ir::Operations &_operations; // operations of current subgraph
```
ragmani, line 135: The above 2 members are replaced with `compiler::LoweredGraph *_lowered_subg`

runtime/onert/core/include/compiler/StaticShapeInferer.h
```diff
@@ -128,12 +186,11 @@ class StaticShapeInferer : public ir::OperationVisitor
...
-  bool _return_has_dynamic_tensor;
+  compiler::LoweredGraph *_lowered_subg;
+  std::unordered_map<ir::SubgraphIndex, std::unique_ptr<OperandObserver>>
+    _subg_input_observers;                                       // child subg input
+  std::unique_ptr<OperandObserver> _controlflow_output_observer; // parent controlflow op output
```
ragmani, line 192: These observer are introduced to remove the dependency of `LoweredGraph` map.

runtime/onert/core/include/compiler/StaticShapeInferer.h
```diff
@@ -28,6 +28,36 @@ namespace onert
 {
 namespace compiler
 {
+/**
+ * @brief Class that observe and update operands.
+ */
+class OperandObserver
```
hseok-oh, line 34: I thought it uses observer pattern from naming, but it seems operand set wrapper for update (operand info updater?), not observer.

tests/nnfw_api/src/one_op_tests/While.test.cc
```diff
@@ -265,6 +265,9 @@ TEST_F(GenModelTest, neg_while_wrong_dtype)
   tc.expectFailRun();
```
hseok-oh, line 265: If test fail on compile phase, we can remove `tc` variable and `tc.expectFailRun()` call.

runtime/onert/core/include/compiler/StaticShapeInferer.h
```diff
@@ -38,32 +68,42 @@ namespace compiler
...
   /**
-   * @brief Infer shape of operands beloning to ops and set the output shape.
+   * @brief Infer shape of operands belonging to ops and set the output shape.
    *        If output shape cannot be known without running op, mark it so that it can be allocated
    *        when running kernel.
-   * @param op Operation
-   * @return @c true if op's input or output has any dynamic tensor; @c false otherwise.
    */
-  bool infer(const ir::Operation &op);
+  void update(void);
```
hseok-oh, line 100: Why do you change name `infer` to `update`?

runtime/onert/core/include/compiler/StaticShapeInferer.h
ragmani: > I thought it uses observer pattern from naming, but it seems operand set wrapper for update (operand info updater?), not observer.

Right, it's not strictly observer pattern. But it performs a role similar to observer in observer pattern. `StaticShapeInferer` lets `OperandObserver`s register with a `StaticShapeInferer` and notifies registered `OperandObserver`s to update operand info. The difference from observer pattern is that it does not provide a observer interface and handles only `OperandObserver`.

tests/nnfw_api/src/one_op_tests/While.test.cc
ragmani: OK. I will remove it.

runtime/onert/core/include/compiler/StaticShapeInferer.h
ragmani: The function has parts to update. I thought that infer parts could be conveyed their meaning by class name. If it is strange, I will restore the name as it was.

runtime/onert/core/include/compiler/StaticShapeInferer.h
ragmani: > I reviewed StaticShaperInferer.h and part of Compiler.cc today. It takes too long time to understand and review. I'll check remain part tomorrow.

OK. This is not urgent. Thanks for your review.

runtime/onert/core/include/compiler/StaticShapeInferer.h
hseok-oh: >I thought that infer parts could be conveyed their meaning by class name. If it is strange, I will restore the name as it was.

IMHO, it's a little bit strange because this function is only function to execute static inferer on class user's view.

runtime/onert/core/include/compiler/StaticShapeInferer.h
hseok-oh: >But it performs a role similar to observer in observer pattern. StaticShapeInferer lets OperandObservers register with a StaticShapeInferer and notifies registered OperandObservers to update operand info. The difference from observer pattern is that it does not provide a observer interface and handles only OperandObserver.

I understood. Thanks. It's OK.

hseok-oh (commented):
Looks good except my comment about naming and test.

runtime/onert/core/include/compiler/StaticShapeInferer.h
ragmani: OK. I will restore the name.

ragmani:
> Looks good except my comment about naming and test.

@hseok-oh 
I modified them as your comment. Thanks.

hseok-oh (approved):
LGTM

PR #9384, opened 2022-07-04
[infra] Install CfgRunner.py

mhs4670go:
This commit installs CfgRunner.py to onelib.


seanshpark (approved):
LGTM

PR #9397, opened 2022-07-08
[luci-interpreter] Improve error message for custom op

jinevening:
This improves error message for unsupported custom op.


seanshpark (approved):
LGTM

Issue #9396, opened 2022-07-08
[one-import-onnx] Support _experimental_disable_batchmatmul_unfold


seanshpark:
Some ONNX model needs `converter._experimental_disable_batchmatmul_unfold = True` option
when converting TF to tflite model while running _one-import-onnx_

Let's find a way to support this.

seanshpark:
as-is
```py
    converter.allow_custom_ops = True
    converter.experimental_new_converter = True

    converter.target_spec.supported_ops = [tf.lite.OpsSet.TFLITE_BUILTINS]
```

to-be
```py
    converter.allow_custom_ops = True
    converter.experimental_new_converter = True
    if ...blabla.exp_disable_batchmatmul_unfold:
        converter._experimental_disable_batchmatmul_unfold = True

    converter.target_spec.supported_ops = [tf.lite.OpsSet.TFLITE_BUILTINS]
```

seanshpark:
and to `one-import-onnx`

```py
parser.add_argument('--exp_disable_batchmatmul_unfold` action='store_true', help='Experimental disable BatMatMul unfold')
...

if _utils._is_valid_attr(args, 'exp_disable_batchmatmul_unfold'):
    tf2tfliteV2_group.add_argument('--exp_disable_batchmatmul_unfold')
```

seanshpark:
done

PR #9398, opened 2022-07-08
[one-cmds] Experimental disable_batchmatmul_unfold option

seanshpark:
This will add experimental disable_batchmatmul_unfold option to
tf2tfliteV2 and one-import-onnx to support ONNX models having BatMatMul Ops.

### DIFF
--- a/compiler/one-cmds/one-import-onnx
+++ b/compiler/one-cmds/one-import-onnx
@@ -86,6 +86,12 @@ def _get_parser():
         action='store_true',
         help='Save intermediate files to output folder')
 
+    # experimental options
+    parser.add_argument(
+        '--disable_batchmatmul_unfold',
+        action='store_true',
+        help='Experimental disable BatMatMul unfold')
+
     return parser
 
 
--- a/compiler/one-cmds/one-import-onnx
+++ b/compiler/one-cmds/one-import-onnx
@@ -90,7 +90,7 @@ def _get_parser():
     parser.add_argument(
         '--disable_batchmatmul_unfold',
         action='store_true',
-        help='Experimental disable BatMatMul unfold')
+        help='Experimental disable BatchMatMul unfold')
 
     return parser
 
--- a/compiler/one-cmds/one-import-onnx
+++ b/compiler/one-cmds/one-import-onnx
@@ -88,7 +88,7 @@ def _get_parser():
 
     # experimental options
     parser.add_argument(
-        '--disable_batchmatmul_unfold',
+        '--experimental_disable_batchmatmul_unfold',
         action='store_true',
         help='Experimental disable BatchMatMul unfold')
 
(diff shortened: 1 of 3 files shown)


seanshpark:
for #9396

seanshpark:
@ejjeong , please check with this patch and give some feedbacks.

mhs4670go (dismissed):
LGTM

compiler/tf2tfliteV2/tf2tfliteV2.py
```diff
@@ -110,6 +110,12 @@ def _get_parser():
         type=str,
         help="Names of the output arrays, comma-separated.")
 
+    # experimental options
+    parser.add_argument(
+        "--disable_batchmatmul_unfold",
+        action="store_true",
+        help="Experimental disable BatMatMul unfold")
```
ejjeong, line 117: How about:
```suggestion
        help="Experimental disable BatchMatMul unfold")
```

compiler/one-cmds/one-import-onnx
```diff
@@ -86,6 +86,12 @@ def _get_parser():
         action='store_true',
         help='Save intermediate files to output folder')
 
+    # experimental options
+    parser.add_argument(
+        '--disable_batchmatmul_unfold',
+        action='store_true',
+        help='Experimental disable BatMatMul unfold')
```
ejjeong, line 93: How about:
```suggestion
        help='Experimental disable BatchMatMul unfold')
```

compiler/one-cmds/one-import-onnx
```diff
@@ -86,6 +86,12 @@ def _get_parser():
         action='store_true',
         help='Save intermediate files to output folder')
 
+    # experimental options
+    parser.add_argument(
+        '--disable_batchmatmul_unfold',
```
ejjeong, line 91: I have a minor concern about the `onecc`-side option name "disable_batchmatmul_unfold".

The tflite converter option `_experimental_disable_batchmatmul_unfold` is explicitly marked as "experimental".

If we cannot maintain this option after tflite converter decides to stop supporting this feature,
we could name the option as "experimental" as well.

---

But please note that this is an optional suggestion. 
I think the final decision should be made after considering the consistency with other option names.

ejjeong (commented):
The behavior looks good to me!
But I have some optional suggestions regarding option naming and help messages.
(I'm okay if this PR is merged as-is.)

compiler/one-cmds/one-import-onnx
ejjeong, line 91: Question: shouldn't we update the option name here as well?
```suggestion
        '--experimental_disable_batchmatmul_unfold',
```

ejjeong (approved):
LGTM, thank you! 
(I hope there's a place to add a test for this feature to avoid issues like https://github.com/Samsung/ONE/pull/9398#discussion_r917566384)

seanshpark:
> (I hope there's a place to add a test for this feature to avoid issues like

We can add a test in `one-cmds/tests`. I need a small test material and some works to do.

seanshpark:
> And FYI) @mhs4670go will be out of the office for two weeks.

thanks for the comment :)

PR #9386, opened 2022-07-04
[npud] Add npud main

jyoungyun:
This commit adds an initial commit of npud.
It defines BUILD_NPUD and builds npud code when this option is on.

ONE-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

draft pr: #9267

PR #9402, opened 2022-07-11
[doc] Update how-to-build-runtime.md

ragmani:
This commit updates how-to-build-runtime.md.
  - Add installing libglib2.0-dev for x64 build

### DIFF
--- a/docs/howto/how-to-build-runtime.md
+++ b/docs/howto/how-to-build-runtime.md
@@ -8,11 +8,12 @@ If you are going to build this project, the following modules must be installed 
 
 - CMake
 - Boost C++ libraries
+- GLib library of C routines
 
 In the Ubuntu, you can easily install it with the following command.
 
 ```
-$ sudo apt-get install cmake libboost-all-dev
+$ sudo apt-get install cmake libboost-all-dev libglib2.0-dev
 ```
 
 If your linux system does not have the basic development configuration, you will need to install more packages. A list of all packages needed to configure the development environment can be found in https://github.com/Samsung/ONE/blob/master/infra/docker/bionic/Dockerfile.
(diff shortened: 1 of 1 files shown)


ragmani:
#9386 requires the installation of `libglib2.0-dev` package to build onert.

docs/howto/how-to-build-runtime.md
```diff
@@ -12,7 +12,7 @@ If you are going to build this project, the following modules must be installed
 In the Ubuntu, you can easily install it with the following command.
 
 ```
-$ sudo apt-get install cmake libboost-all-dev
+$ sudo apt-get install cmake libboost-all-dev libglib2.0-dev
```
hseok-oh, line 15: We need to update above list (line 9~10) amd below `apt` command (line 23~38)

hseok-oh: If we install simulator, we may use `npud` on x64.

cc/ @jyoungyun

ragmani:
> IMO, it's better to make separate part for system service because it is optional depend on platform yet.

OK, I will close this PR. And I have no plan to write a separate part for system service because I don't know about it well.

https://github.com/Samsung/ONE/blob/debeb771d15c6d3efe1f5449ffc0164457919e73/infra/nnfw/cmake/CfgOptionFlags.cmake#L34
https://github.com/Samsung/ONE/blob/debeb771d15c6d3efe1f5449ffc0164457919e73/runtime/service/npud/CMakeLists.txt#L1-L7

By the way, `npud` is built as default now. It looks like `npud` is not optional.

glistening:
> By the way, npud is built as default now. It looks like npud is not optional.

I think it would be good default off. `npd` does nothing with `onert` at this moment.

Currently, everyone who want to use `onert` have to install `glib` and build `npud`, which is optional at this moment.

cc @Samsung/one_onert, @jyoungyun

jyoungyun:
I agree with this point. Currently, npu daemon is not ready and not activated yet. So we can change it to false by default.

Issue #9367, opened 2022-06-30
[circle2circle] Deprecate -O1 option


seanshpark:
There exist group option in _circle2circle_ `-O1` which was once `--all`
When `--all` was added, it's intention was to include all optimization options
so that user don't have to add individual options.
But it didn't go well and was renamed as `-O1`.
Still this is not maintained as well and group option is now able to be provided in _one-cmd_ level.

Thus I'd like to deprecate this option as there is no known usage case as of now and it would be better to remove this before there are serious use cases.

seanshpark:
related: #5780

seanshpark:
Please add comments about this.
I'll wait for a week an start doing this.

lemmaa:
> ... But it didn't go well and was renamed as `-O1`. ...

@seanshpark , can you explain in a little more detail where it did not go well? Or is there a link for reference about this history?  Thank you in advance!

seanshpark:
> can you explain in a little more detail where it did not go well? 

We didn't care about `-O1` and I have no known usage of this option.
We have added many optimization options since but never managed to add some to this group including me.

> Or is there a link for reference about this history? 

As for me, have none.

seanshpark:
CC @hyunsik-yoon

seanshpark:
done

PR #9406, opened 2022-07-11
DRAFT deprecate --O1

seanshpark:
on-going draft to deprecate --O1 option in circle2circle.


seanshpark:
for  #9367

seanshpark:
done

PR #9407, opened 2022-07-11
[onert] Change unnecessary copies to use references

ragmani:
This commit changes unnecessary copies to use references.


hseok-oh (approved):
LGTM. 

BTW, `Fp32ToFp16Converter.cc/h` was disabled by #5552

PR #9413, opened 2022-07-11
[one-cmds] Do not use --O1 for one-optimize test

seanshpark:
This will revise not to use --O1 option and instead use --resolve_customop_add
for one-optimize test.


seanshpark:
for #9367 from #9406

yunjayh (approved):
LGTM

PR #9414, opened 2022-07-11
[cmn-arts] Do not use --O1

seanshpark:
This will revise not to use --O1 option and instead use --resolve_customop_add
for generating optimized circle with circle2circle.


seanshpark:
for #9367 from #9406

yunjayh (approved):
LGTM

PR #9423, opened 2022-07-13
[luci] Resolve custom Ops before nchw-to-nhwc conversion

jinevening:
This resolves custom Ops before nchw-to-nhwc conversion.


seanshpark (approved):
LGTM

PR #9425, opened 2022-07-13
[one-cmds] Fix PIP_OPTIONS

seanshpark:
This will fix PIP_OPTIONS to include also pypi.python.org for trusted-host.
and fix patch for protobuf==3.20.1 to use PIP_OPTIONS too.


yunjayh (approved):
LGTM

PR #9428, opened 2022-07-13
[circle2circle] Deprecate --O1 option

seanshpark:
This will deprecate --O1 option.


jinevening (approved):
LGTM

seanshpark:
for #9367 from #9406

PR #9429, opened 2022-07-13
[common-artifact] Make it possible to build since cmake version 3.12

ragmani:
This commit makes it possible to build since cmake version 3.12.
  - Replace FindPythonlibs(deprecated since 3.12) with FindPython3 in cmake version 3.12 or later to use deprecated variables.

### DIFF
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -2,6 +2,12 @@
 find_package(PythonInterp 3.8 QUIET)
 find_package(PythonLibs 3.8 QUIET)
 
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12")
+  find_package(Python3 COMPONENTS Interpreter Development QUIET)
+  set(PYTHON_EXECUTABLE Python3_EXECUTABLE)
+  set(PYTHON_VERSION_MINOR Python3_VERSION_MINOR)
+endif()
+
 if(NOT ${PYTHONINTERP_FOUND})
   message(STATUS "Build common-artifacts: FALSE (Python3 is missing)")
   return()


compiler/common-artifacts/CMakeLists.txt
```diff
@@ -2,6 +2,12 @@
 find_package(PythonInterp 3.8 QUIET)
 find_package(PythonLibs 3.8 QUIET)
```
ragmani, line 3: Deprecated since version 3.12.
https://cmake.org/cmake/help/latest/module/FindPythonLibs.html

seanshpark: Are you going to remove above these two lines? if so, please add a TODO.
If you have a full change, please share it with draft PR.

ragmani: Unfortunately, FindPython3 was introduced in version 3.12. So, to support python variables, these two lines should be kept.
https://cmake.org/cmake/help/latest/module/FindPython3.html

ragmani:
> @ragmani , can you please share the issue that explains the background of this change?
Are you planning to change to cmake 3.12?

There is no background. I'm using a higher version of cmake (ubuntu 22.04), and I just want to build `onecc` in my local environment.

compiler/common-artifacts/CMakeLists.txt
seanshpark: @ragmani , you've commented to these two lines
```
find_package(PythonInterp 3.8 QUIET)
find_package(PythonLibs 3.8 QUIET)
```
with `Deprecated since version 3.12.` . 
I don't understand what the relation between these.
Why did you mention `Deprecated ` ?

seanshpark:
> I'm using a higher version of cmake (ubuntu 22.04), and I just want to build onecc in my local environment.

Please fire an issue about your activity.
As you know, our primary platform is 18.04 (and also 20.04 is tested too).
I'm not sure I can accept for your changes for `build onecc in my local environment` without any issue registered.

compiler/common-artifacts/CMakeLists.txt
```diff
@@ -2,6 +2,12 @@
...
 
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12")
+  find_package(Python3 COMPONENTS Interpreter Development QUIET)
+  set(PYTHON_EXECUTABLE Python3_EXECUTABLE)
+  set(PYTHON_VERSION_MINOR Python3_VERSION_MINOR)
+endif()
```
seanshpark, line 9: I'm not sure adding patche with version check is appropriate.
What I think, but also not confident so we can discuss in another issue about this topic.
- raise an issue to support 22.04 as this year is 2022.
- upgrade default cmake version to meet this
- and so on...

compiler/common-artifacts/CMakeLists.txt
ragmani: `FindPythonInterp` and `PythonLibs` are deprecated since the version 3.12 of `cmake`. So variables of these modules like `PYTHON_EXECUTABLE` or `PYTHON_VERSION_MINOR` are empty since the version 3.12 of `cmake`.
`FindPython3` is introduced since the version 3.12 of `cmake`.  We should use `FindPython3` module to get the informatio of variables since the version 3.12 of `cmake`.

seanshpark:
This is from my environment
```
$ cmake --version
cmake version 3.16.3
```

```diff
diff --git a/compiler/common-artifacts/CMakeLists.txt b/compiler/common-artifacts/CMakeLists.txt
index f4acc5846..8bce04da4 100644
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -2,6 +2,12 @@
 find_package(PythonInterp 3.8 QUIET)
 find_package(PythonLibs 3.8 QUIET)

+message(STATUS "!!!! PYTHONINTERP_FOUND=${PYTHONINTERP_FOUND}")
+message(STATUS "!!!! PYTHON_VERSION_MINOR=${PYTHON_VERSION_MINOR}")
+message(STATUS "!!!! PYTHON_VERSION_MAJOR=${PYTHON_VERSION_MAJOR}")
+message(STATUS "!!!! PYTHON_EXECUTABLE=${PYTHON_EXECUTABLE}")
+message(STATUS "!!!! PYTHON_VERSION_MINOR=${PYTHON_VERSION_MINOR}")
+
 if(NOT ${PYTHONINTERP_FOUND})
   message(STATUS "Build common-artifacts: FALSE (Python3 is missing)")
   return()
```
result
```
-- Configure COMMON-ARTIFACTS
-- !!!! PYTHONINTERP_FOUND=TRUE
-- !!!! PYTHON_VERSION_MINOR=8
-- !!!! PYTHON_VERSION_MAJOR=3
-- !!!! PYTHON_EXECUTABLE=/home/maxwell/.local/bin/python3.8
-- !!!! PYTHON_VERSION_MINOR=8
```

Can you describe what your problem is?
There is no explanation of the problem except cmake 3.12 does not support something but there is actual problem explained.

compiler/common-artifacts/CMakeLists.txt
ragmani: OK, I will create an issue about this topic.
I think it's a bit early to upgrade default cmake version. I will upload it in a direction supports ubuntu 22.04 as well.

seanshpark:
> Thanks for your help. Could I change 3.8 to 3 as the patch above? I'm not sure it works well in other environments.

Glad you found the problem :)
I'm afraid we have to go with `python3.8`, not `python3` for now.
- We support Ubuntu 18.04, which default python is 3.6
- We use TensorFlow2.8 which doesn't exist for python 3.6 but exist for python 3.8
- So we made our one-compiler depend on python 3.8 so that python 3.8 is also installed
- But as it may be dangerous to change 3.8 as default, we explictly use python 3.8

compiler/common-artifacts/CMakeLists.txt
ragmani: I created that issue at #9432

hseok-oh:
I'm using cmake 3.23 on ubuntu 20.04, and python 3.8 (default) is installed.
I tried to find python version 3.7, not 3.8. And it finds 3.8 as I expected. 

```
hseok82:/home/nfs/git/ONE$ git diff
diff --git a/compiler/common-artifacts/CMakeLists.txt b/compiler/common-artifacts/CMakeLists.txt
index f4acc5846..f1e044c83 100644
--- a/compiler/common-artifacts/CMakeLists.txt
+++ b/compiler/common-artifacts/CMakeLists.txt
@@ -1,6 +1,12 @@
 #[[ Generate common python virtual enviornment ]]
-find_package(PythonInterp 3.8 QUIET)
-find_package(PythonLibs 3.8 QUIET)
+find_package(PythonInterp 3.7 QUIET)
+find_package(PythonLibs 3.7 QUIET)
+
+message(STATUS "!!!! PYTHONINTERP_FOUND=${PYTHONINTERP_FOUND}")
+message(STATUS "!!!! PYTHON_VERSION_MINOR=${PYTHON_VERSION_MINOR}")
+message(STATUS "!!!! PYTHON_VERSION_MAJOR=${PYTHON_VERSION_MAJOR}")
+message(STATUS "!!!! PYTHON_EXECUTABLE=${PYTHON_EXECUTABLE}")
+message(STATUS "!!!! PYTHON_VERSION_MINOR=${PYTHON_VERSION_MINOR}")
 
 if(NOT ${PYTHONINTERP_FOUND})
   message(STATUS "Build common-artifacts: FALSE (Python3 is missing)")
hseok82:/home/nfs/git/ONE$ cmake --version
cmake version 3.23.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).

hseok82:/home/nfs/git/ONE$ ./nncc configure

.....

-- !!!! PYTHONINTERP_FOUND=TRUE
-- !!!! PYTHON_VERSION_MINOR=8
-- !!!! PYTHON_VERSION_MAJOR=3
-- !!!! PYTHON_EXECUTABLE=/usr/bin/python3.8
-- !!!! PYTHON_VERSION_MINOR=8

......

```

I'll try on ubuntu 22.04.

hseok-oh:
I've tried on my ubuntu 22.04 environment (cmake 3.22.1, python 3.10.4), and didn't meet fail.

```
-- !!!! PYTHONINTERP_FOUND=TRUE
-- !!!! PYTHON_VERSION_MINOR=10
-- !!!! PYTHON_VERSION_MAJOR=3
-- !!!! PYTHON_EXECUTABLE=/usr/bin/python3
-- !!!! PYTHON_VERSION_MINOR=10
```

ragmani:
I've tried on an ubuntu 22.04 docker container, and I didn't meet that fail either.
```
-- Configure COMMON-ARTIFACTS
-- !!!! PYTHONINTERP_FOUND=TRUE
-- !!!! PYTHON_VERSION_MINOR=10
-- !!!! PYTHON_VERSION_MAJOR=3
-- !!!! PYTHON_EXECUTABLE=/usr/bin/python3
-- !!!! PYTHON_VERSION_MINOR=10
-- Configure COMMON-ARTIFACTS - Done
```
I'm confused as to why only my local machine doesn't work.
Anyway, this problem was limited to an individual environment. So I close this issue.

ragmani:
Try it after installing build-essential
```bash
sudo apt install build-essential
```

PR #9430, opened 2022-07-13
[one-cmds] Fix one-import-onnx_002.test failure

seanshpark:
This will fix one-import-onnx_002.test to use circle-operator.
- circle-inspect tool is not a installed tool thus cannot use here


seanshpark:
to fix #9427

jinevening (approved):

PR #9445, opened 2022-07-17
DRAFT Fix StridedSlice with -1 shape

seanshpark:
on-going draft to fix StridedSlice with -1 shape.


res/TensorFlowLiteRecipes/StridedSlice_004/test.recipe
```diff
@@ -0,0 +1,50 @@
+#
+# Failed case from https://github.com/Samsung/ONE/issues/9439
+#
+operand {
+  name: "Placeholder"
+  type: FLOAT32
+  shape { dim: 1 dim: 16 dim: 16 dim: 32 }
+  quant { quantized_dimension: 0 }
```
jinevening, line 8: The target model is fp32 model, so `quant` field is not necessary for all operands.

seanshpark: `for all operands.`

Can you confirm for `INT32` too?
- add `quant` for UINT8 and INT16 only
- if and only if `quantized_dimension` is not zero

And what about `scale` and `zero_point` ?

jinevening: > can you confirm for INT32 too?
> And what about scale and zero_point ?

For the target model, quantization parameters (`quant`) is not necessary for INT32 tensors. scale and zero_point are not necessary as well.

As to https://github.com/Samsung/ONE/issues/9446 (not directly related to the strided slice model in this PR), 

I think tflchef-reverse should generate quantization parameters (`quant`) based on the tflite model, i.e., `quantization` field in the `Tensor` (from tflite schema)

```
  quantization:QuantizationParameters;  // Optional.
```

We may not need complex rules for generating quantization parameters.

seanshpark:
@jinevening , can you please check with current draft?

jinevening:
I checked that `circle2circle` works well for the target model on this DRAFT. 👍

seanshpark:
done

PR #9454, opened 2022-07-19
[luci] Fix QuantizeDequantizeWeights condition

jinevening:
This fixes condition to apply QuantizeDequantizeWeights.

### DIFF
--- a/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
+++ b/compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
@@ -353,15 +353,15 @@ private:
 private:
   // Check if
   // 1. node is const
-  // 2. node was not quantized
+  // 2. node's dtype is float32
   bool is_quantizable(loco::Node *node)
   {
     auto const_node = dynamic_cast<luci::CircleConst *>(node);
     if (not const_node)
       return false;
 
-    // Skip if this is already quantized
-    if (is_quantized(const_node))
+    // Skip if this is not float32
+    if (const_node->dtype() != loco::DataType::FLOAT32)
       return false;
 
     return true;


compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -353,15 +353,15 @@ struct QuantizeDequantizeWeights final : public luci::CircleNodeMutableVisitor<v
...
-  // 2. node was not quantized
+  // 2. node's dtype is float32
   bool is_quantizable(loco::Node *node)
   {
     auto const_node = dynamic_cast<luci::CircleConst *>(node);
     if (not const_node)
       return false;
 
-    // Skip if this is already quantized
-    if (is_quantized(const_node))
```
jinevening, line 364:
```cpp
bool is_quantized(const CircleNode *node)
{
  return node->quantparam() != nullptr &&
         (node->dtype() == loco::DataType::U8 ||  // activation, weight (uint8 quant)
          node->dtype() == loco::DataType::S16 || // activation, weight (int16 quant)
          node->dtype() == loco::DataType::S32 || // bias (uint8 quant)
          node->dtype() == loco::DataType::S64);  // bias (int16 quant)
}
```

Note that `is_quantized` checks if quantparam exists. quantparam is removed before QuantizeDequantizeWeigthsPass is invoked, so `is_quantized` always returns false.

This PR fixes this by not checking the existence of quantparam.

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -353,15 +353,15 @@ struct QuantizeDequantizeWeights final : public luci::CircleNodeMutableVisitor<v
...
+    // Skip if this is not float32
```
seanshpark, line 363: there are float64 and float16... some models are float16.

jinevening: You're right. But currently we only support quantization of float32. And I think we can remove the float16 tensors in the target model by some pre-processing :)

jinevening: No. float16/64 can enter here. Then `is_quantizable` will return false, and nothing will happen.

Currently, the model will be rejected by `record-minmax` here (https://github.com/Samsung/ONE/pull/9457/files).

Issue #9439, opened 2022-07-15
[luci] Fail to infer shape of StridedSlice


jinevening:
## What

Shape inference failed in StridedSlice Op with negative strides.

```
$ circle2circle model.circle model.circle
circle2circle: /home/hjj/ONE/compiler/luci/service/src/ShapeInfer_StridedSlice.cpp:272: loco::TensorShape
 luci::infer_output_shape(const luci::CircleStridedSlice*): Assertion `dim_shape > 0' failed.
Aborted (core dumped)
```

[model.zip](https://github.com/Samsung/ONE/files/9119548/model.zip)

CC @seanshpark

seanshpark:
Thank you for this test model :)
I think fixing this may take some time... I think I need some study with stridedslice again -_-;

seanshpark:
@jinevening , if you have `.recipe` file for this model, can you please share it?

jinevening:
I made the model using tfgex, so I don't have a recipe. I left my comment in #9445 :)

seanshpark:
@jinevening , do you have `.tflite` file of this model?

jinevening:
> do you have .tflite file of this model?

Here is the tflite of this model. [model_tflite.zip](https://github.com/Samsung/ONE/files/9129961/model_tflite.zip)

> for test sample model.circle, is expected output shape (1, 16, 16, 32) ?

Yes. It seems that the output should have the same shape, but contents of the second dimension should be reversed.

seanshpark:
> Yes. It seems that the output should have the same shape, but contents of the second dimension should be reversed.

value test seems OK; that is, luci-interpreter is working OK :)

jinevening:
I checked that the target model is successfully imported/exported. @seanshpark Can I close this issue?

PR #9455, opened 2022-07-19
[luci/svc] Tidy shape inference for StridedSlice

seanshpark:
This will tidy unused codes of shape inference for StridedSlice.


seanshpark:
for #9439 , from #9445

jinevening (approved):
LGTM

PR #9457, opened 2022-07-19
[record-minmax] Improve error message

jinevening:
This prints node name if tensor dtype is not float.


seanshpark (approved):
LGTM

Issue #9464, opened 2022-07-20
[luci/pass] Revise FuseBatchNormWithTConvPass where TConv has bias


seanshpark:
Let's enable `FuseBatchNormWithTConvPass` to work where TConv has bias.

seanshpark:
done

PR #9465, opened 2022-07-20
DRAFT: FuseBatchNormWithTConvPass with bias

seanshpark:
on-going draft to support FuseBatchNormWithTConvPass with bias.


seanshpark:
for #9464

seanshpark:
done

PR #9469, opened 2022-07-20
[luci/pass] Revise FuseBatchNormWithTConvPass

seanshpark:
This will revise FuseBatchNormWithTConvPass to support fusion when bias exist.

### DIFF
--- a/compiler/luci/pass/src/FuseBatchNormWithTConvPass.cpp
+++ b/compiler/luci/pass/src/FuseBatchNormWithTConvPass.cpp
@@ -29,7 +29,7 @@ namespace
  *  NOTE TF's BatchNormalization is converted to Mul and Add.
  *
  *  BEFORE
- *                     |   [CircleOutputExclude]
+ *                     |   [CircleConst]/[CircleOutputExclude]
  *                     |   / [CircleConst]
  *                     |  / /
  *     [CircleTransposeConv]  [CircleConst]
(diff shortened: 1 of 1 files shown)


seanshpark:
for #9464 , from  #9465

compiler/luci/pass/src/FuseBatchNormWithTConvPass.cpp
```diff
@@ -96,10 +96,27 @@ bool fused_batch_norm_with_tconv(luci::CircleAdd *add)
     return false;
 
   auto filter_out_chn = filter->dim(0).value();
-  if (filter_out_chn != scale->dim(0).value())
-    return false;
-  if (filter_out_chn != shift->dim(0).value())
-    return false;
+  // allow scale/shift and bias shape of [N], [1,N], [N,1], [1,1,N], ...
```
jinevening, line 99: Since BN works "channel-wise", the last dimension should be N. So, [N, 1] is not allowed.

The last dimension should be N, and the rest of dimensions should be 1. So, the shape of scale and shift should be [N] or [1, 1, 1, N] (scale and shift are limited to rank 4 in the code).

seanshpark: Thanks! I'll update this PR :)

jinevening (approved):
LGTM!

PR #9470, opened 2022-07-20
[res] Tidy Net_TConv recipes

seanshpark:
This will tidy unnecessary fields in Net_TConv_* recipes.


jinevening (approved):
LGTM

Issue #9450, opened 2022-07-18
Implement one-init


hyunsik-yoon:
Let's implement one-init.
#9369 describes analysis work for this implementation.

- [x] write a draft (#9421)
- [x] test if cfg file from the draft work OK
  - If cfg file be used to compile models (except some unusual models), make PRs and make one-init to be included into the release.
- [ ] Add more PRs incrementally by enhancing one-init with some exceptional models

Note: 
- With cfg file fron one-init, onecc should produce backend code after codegen. 
- "backend code after codegen should be optimized" would be convered in next step.
    - why? some param cannot be automatically entered (e.g., quantization input data path)
    - for the first step, we will put some comment into cfg file that, e.g., entering input data path would increase accuracy.

glistening:
@hyunsik-yoon Could you please point out the issues to get background about `one-init`?

As I roughly understand, `one-init` is:
- created for one-vscode
- related to `onecc`
- not related to `onert`

IIUC, people may be confused what for `one-init`. 
- Is it the first mandatory step to build or use something in one?
- Is it required to run before run `onert`?

Maybe `onecc-init` ?

hyunsik-yoon:
@glistening,
Thanks for the good questions. It's kinda hard to track the history. 😢 

> Could you please point out the issues to get background about one-init?

- one-init is tool in ONE project (not one-vscode). Of course, one-init can be used by other projects such as one-vscode and one-vscode will have plan to use one-init.
- what is one-init?
  - To compile a given model, some people need to spend time to create the first cfg file. Normally people who does not understand one tools, they have difficulty in creating the first cfg file.
  - To make this easier, one-init generates an initial cfg file for a model.

> - Is it the first mandatory step to build or use something in one?
> - Is it required to run before run `onert`?

It is an optional step to create initial cfg easier. People who can create a cfg file for a given model by themselves don't need to use one-init.

> Maybe `onecc-init` ?

Like other onecc tools, IMHO we could follow the naming convention in ONE project:
- `one-init` (cmd name)
  - there are many other tools with name `one-...`. (`$ ls /usr/share/one/bin/one-*`)
- `one-init` also could be invoked with `onecc init ...` like others in `EXAMPLES` section in `$ man onecc`

hyunsik-yoon:
I added two more lines in action list of https://github.com/Samsung/ONE/issues/9450#issue-1307418648:

>   - If cfg file be used to compile models (except some unusual models), make PRs and make one-init to be included into the release.
[...]

PR #9421, opened 2022-07-12
DRAFT: Introduce one-init

hyunsik-yoon:
This is draft work to introduce one-init.

for #9450
after #9369

ONE-DCO-1.0-Signed-off-by: Hyun Sik Yoon <hyunsik.yoon.1024@gmail.com>

PR #9259, opened 2022-06-14
draft for onecc init

YongseopKim:
Issue: https://github.com/Samsung/ONE/issues/9153

## How to use?

```
$ onecc init
$ ls
init.cfg # copied by onecc.template.cfg
```

```
$ onecc init model.tflite
$ ls
model.cfg # copied by onecc.template.cfg and fixed
```

PR #9475, opened 2022-07-21
[one-cmds] Add args list for one-init

hyunsik-yoon:
This adds args list for one-init.
- Currently tflite and onnx models are supported.
- `-C config_file` and `-S section` are not args for one-init. So, they are excluded.

parent issue: https://github.com/Samsung/ONE/issues/9450
draft: https://github.com/Samsung/ONE/pull/9421

### DIFF
--- a/compiler/one-cmds/one-init
+++ b/compiler/one-cmds/one-init
@@ -71,25 +71,36 @@ def _get_backends_list():
 
     return backends_list
 
+
 # TODO Add support for TF graphdef and bcq
 def _get_parser(backends_list):
-    init_usage = ('one-init [-h] [-v] [-V] '
-                  '[-i INPUT_PATH] '
-                  '[-o OUTPUT_PATH] '
-                  '[-m MODEL_TYPE] '
-                  '[-b BACKEND] '
-                  # args for onnx model
-                  '[--input_img_layout LAYOUT] '
-                  '[--output_img_layout LAYOUT] '
-                  # args for backend driver
-                  '[--] [COMMANDS FOR BACKEND DRIVER]')
+    init_usage = (
+        'one-init [-h] [-v] [-V] '
+        '[-i INPUT_PATH] '
+        '[-o OUTPUT_PATH] '
+        '[-m MODEL_TYPE] '
+        '[-b BACKEND] '
+        # args for onnx model
+        '[--convert_nchw_to_nhwc] '
+        '[--nchw_to_nhwc_input_shape] '
+        '[--nchw_to_nhwc_output_shape] '
+        # args for backend driver
+        '[--] [COMMANDS FOR BACKEND DRIVER]')
+    """
+    NOTE
+    layout options for onnx model could be difficult to users.
+    In one-init, we could consider easier args for the the above three:
+    For example, we could have another option, e.g., --input_img_layout LAYOUT
+      - When LAYOUT is NHWC, apply 'nchw_to_nhwc_input_shape=True' into cfg
+      - When LAYOUT is NCHW, apply 'nchw_to_nhwc_input_shape=False' into cfg
+    """
 
     parser = argparse.ArgumentParser(
         description='Command line tool to generate initial cfg file. '
-        'Currently tflite and onnx models are supported.',
+        'Currently tflite and onnx models are supported',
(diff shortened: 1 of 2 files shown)


compiler/one-cmds/one-init
```diff
@@ -71,14 +71,51 @@ def _get_backends_list():
...
-
+# TODO Add support for TF graphdef and bcq
 def _get_parser(backends_list):
+    init_usage = ('one-init [-h] [-v] [-V] '
+                  '[-i INPUT_PATH] '
+                  '[-o OUTPUT_PATH] '
+                  '[-m MODEL_TYPE] '
+                  '[-b BACKEND] '
+                  # args for onnx model
+                  '[--input_img_layout LAYOUT] '
```
seanshpark, line 82: is there particular reason to use `img` in input layout?

compiler/one-cmds/one-init
```diff
@@ -71,14 +71,51 @@ def _get_backends_list():
...
+        '-i', '--input_path', type=str, help='full filepath of the input model file')
+    parser.add_argument(
+        '-o', '--output_path', type=str, help='full filepath of the output cfg file')
+    parser.add_argument(
+        '-m',
+        '--model_type',
+        type=str,
+        help=('type of input model: "onnx", "tflite". '
+              'If the file extension passed to --input_path is '
+              '".tflite" or ".onnx", this arg can be omitted.'))
```
seanshpark, line 104: Q) this seems optional and if model is tflite or onnx and this can be omitted, why need this option?

compiler/one-cmds/one-init
```diff
@@ -71,14 +71,51 @@ def _get_backends_list():
...
+
+    onnx_group = parser.add_argument_group('arguments when model type is onnx')
+    onnx_group.add_argument(
+        '--input_img_layout',
+        type=str,
+        help=
+        'input data layout when input data format is image. Allowed list is "NHWC", "NCHW".'
```
seanshpark, line 111: ONNX models usually accept NCHW models. if `--input_img_layout` is `NCHW` then do users need to give `--input_img_layout NCHW` ?

compiler/one-cmds/utils.py
```diff
@@ -26,7 +26,7 @@
 import onelib.constant as _constant
 
 
-def _add_default_arg(parser):
+def _add_default_arg(parser, include_config_file=True):
```
seanshpark, line 29: I think we need to get review from @mhs4670go

seanshpark: I've searched function with default value and there already exist.
I think it's OK... but it would be better to get review from @mhs4670go :)

compiler/one-cmds/one-init
hyunsik-yoon: This option is from this comment: https://github.com/Samsung/ONE/pull/9259/files#r897520623

compiler/one-cmds/one-init
hyunsik-yoon: The reason behind this arg is:
- When input is not image (not 4D tensor), we don't need `--convert_nchw_to_nhwc` option even in onnx model. So I added this optional arg to check if input is 4D image or not.
- This arg is used to determine `--convert_nchw_to_nhwc` , `--nchw_to_nhwc_input_shape` options, which I think the are difficult to understand from the view point of users. I think that `--input_img_layout NHWC` seems easier to understand.

After discussion with @seanshpark 
1. We will add traditional args `--convert_nchw_to_nhwc` , `--nchw_to_nhwc_input_shape`, `--nchw_to_nhwc_output_shape` to keep consistency with one-optimize.
2. We can add `--img_layout NCHW,NCHW` arg too. But let's do 1. first.

compiler/one-cmds/utils.py
hyunsik-yoon: After discussion with @seanshpark,
we could introduce, e.g., `def _add_default_arg_no_CS(..)` function. Later we can merge `_add_default_arg()` and `def _add_default_arg_no_CS(..)`.

seanshpark (approved):
LGTM

jinevening (approved):
LGTM

Issue #9468, opened 2022-07-20
[luci-micro] Split ordinary luci-interpreter and luci-micro interpreter


BalyshevArtem:
## Why
For further development of luci-micro, it is necessary to create its own luci-interpreter due to the differences between the current and the interpreter for luci-micro have become too noticeable. 

Current luci-interpreter is designed for collecting runtime data and convenience of developers (integration with luci IR).
This requires some additional code and data structures that are not needed for MCU inference. So, we discovered three main problems for MCU build:

- dependency from luci IR and luci importer. Currently, luci-interpreter is very much tied to luci-importer, which creates an intermediate representation that significantly spends resources on the MCU.

- internal structures could be simplified to reduce RAM consumption. Currently, luci-interpreter during its configure and execution create large structures that also produce a lot of overhead for memory consumption.

- some kernels in luci-interpreter could be rewrite to reduce memory consumption.

So, during rewriting luci-interpreter to resolve these problems we found that our "new" interpreter is very different from the original one. We decided that it is better to split "new" luci-interpereter" from "old" one.

## How

There is small plan for actions to create our own luci-micro interpreter:

- [x] Copy Interpreter class and luci-ir -> runtime IR converters from compiler/luci-interpreter to compiler/luci-micro #9478
- [x] Refactor copied code to remove luci IR dependency #9651 
- [x] Remove unnecessary dependencies  #9698
- [x] Rewrite luci-interpreter-micro runtime and kernels #9700
- [x] Add helper to generate kernels builder list #9702 #9703
- [x] Changes for kernels to reduce memory consumption #9705
- [x] Add schema generation to luci-micro #9758 
- [x] Remove C++ unnecessary dependencies from micro interpreter #9788 
- [x] Move from /compiler dir to new dir, update Cmake #9830

BalyshevArtem:
I think, everything done, let's close issue

PR #9478, opened 2022-07-21
[luci-micro] Copy luci-interpreter to luci-micro

BalyshevArtem:
This PR copies luci-interpreter to luci-micro.

issue: #9468


binarman (approved):
LGTM

binarman:
FYI @seanshpark

This is simple copy, nothing is changed compared to original luci-interpreter.

This code is not connected to main cmake infrastructure yet, so it should not drop any SW metrics.

PR #9482, opened 2022-07-22
[tflite2circle] Convert old tflite with opcode

seanshpark:
This will fix to convert old tflite with opcode being 0.


seanshpark:
from #9480

Issue #9479, opened 2022-07-22
Compiler FE: limited FLOAT16 support


seanshpark:
Some of our internal models have FLOAT16 tensors followed by Dequantize.
This can be folded to FLOAT32 tensors.
Let's support for this case.

Issues
- need to enable _loco_
- need to enable for _tflchef_ for test materials
- and maybe more

Plan
- support this in `CircleConst - CircleDequantize` sequence with `fold_dequantize` option

seanshpark:
FLOAT16 is 16bit value but there is no type for this.
As we are to convert to FLOAT32 in `FoldDequantizePass`, we can use https://github.com/Maratyszcza/FP16 library where TensorFlow uses this too.

seanshpark:
done

PR #9483, opened 2022-07-22
[tflite2circle] Tidy MaximumMinimumOptions

seanshpark:
This will tidy MaximumMinimumOptions with unused variable.


seanshpark:
from #9480

jinevening (approved):
LGTM

It seems that this PR is not related to #9480 ?

seanshpark:
> It seems that this PR is not related to #9480 ?

Not related to task of #9480, which is #9479, but found it while working on it :)

PR #9484, opened 2022-07-22
[luci/logex] Revise CircleConst

seanshpark:
This will revise CircleConstSummaryBuilder to show dtype and shape for
CircleConst node.


seanshpark:
from #9480

PR #9494, opened 2022-07-22
[one-cmds] Adds early version of _generate() into one-init

hyunsik-yoon:
This adds early version of _generate() into one-init.
- introduce functions to construct cfg file
- introduce class handling cfg where comment can be stored.

Why comment is important?
cfg file will be used by human. Therefore we'd better leave human-readable description inside cfg for them to fix cfg
in easier way.

For example, we can make cfg like the following:

```
[one-optimize]
; onnx only option
convert_nchw_to_nhwc = True
; add more option below

```

parent issue: https://github.com/Samsung/ONE/issues/9450
draft: https://github.com/Samsung/ONE/pull/9421

### DIFF
--- a/compiler/one-cmds/one-init
+++ b/compiler/one-cmds/one-init
@@ -196,6 +196,7 @@ def _get_executable(args, backends_list):
         raise FileNotFoundError(backend_base + ' not found')
 
 
+# TODO Support workflow format (https://github.com/Samsung/ONE/pull/9354)
 def _generate():
     # generate cfg file
     config = CommentableConfigParser()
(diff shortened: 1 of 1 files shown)


compiler/one-cmds/one-init
```diff
@@ -177,6 +196,51 @@ def _get_executable(args, backends_list):
...
+    def _gen_import():
+        pass  # NYI
+
+    def _gen_optimize():
+        pass  # NYI
+
+    def _gen_quantize():
+        pass  # NYI
+
+    def _gen_codegen():
```
seanshpark, line 223: CC @mhs4670go 
@hyunsik-yoon , #9354 is WIP and these may not be compatible with it.
please add a TODO comment to revise this later.

hyunsik-yoon: Thanks. TODO was added.

seanshpark (approved):
LGTM

Issue #9458, opened 2022-07-19
[quantizer] Support ReduceMax


jinevening:
## What

Let's support quantization of ReduceMax Op

TODOs
- Support in record-minmax
  - luci-interpreter
- Support in circle-quantizer
  - QuantizeWithMinMax
  - QuantizedNodeVerifier 
- Support in fake quantization
  - ConvertToFakeQuantizedModel

jinevening:
Done.

PR #9495, opened 2022-07-24
[luci] Support fake quantization of ReduceMax Op

jinevening:
This supports fake quantization of ReduceMax.

---
Related to: https://github.com/Samsung/ONE/issues/9458


seanshpark (approved):
LGTM

PR #9497, opened 2022-07-24
[souschef] Use numeric_limits::lowest for min cap

jinevening:
This uses numeric_limits::lowest for min cap.


jinevening:
std::numeric_limits<float>::min() returns the smallest positive value that float can express. So, negative values are clipped, which is not intended by random generator.

seanshpark:
@binarman PTAL

seanshpark (approved):
LGTM thank you!

hseok-oh:
@nnfw-bot test nncc-debug nncc-release

seanshpark:
I'll wait for a day for review from @binarman and if no response, I'll land.

binarman (approved):
LGTM, thank you!

PR #9498, opened 2022-07-25
[tflchef] Support limited FLOAT16

seanshpark:
This will support limited FLOAT16 with tensor.


seanshpark:
for #9479 , from #9480

jinevening (approved):
LGTM

PR #9504, opened 2022-07-25
[common-artifacts] Revise circle2circle option

seanshpark:
This will revise circle2circle option for optimization tests to include fold_dequantize.


seanshpark:
for #9479 , from #9480

jinevening:
Can I ask why we need this change? Export of FLOAT16 is not supported?

seanshpark:
> Can I ask why we need this change? Export of FLOAT16 is not supported?

`Net_Dequantize_Add_*` models have float16 tensors, which I intentionally didn't export this.
Reason is that our OneRT and TRIX don't support float16 so it would be an error.
(I think it would be better to add some comments to luci/export about this.)

So, what I could do for now is that, use `fold_dequantize` option and remove `float16` in `circle2circle` testing.
This PR is for this purpose.

seanshpark:
I've posted #9506

jinevening:
Thanks for the explanation.

jinevening (approved):
LGTM

PR #9506, opened 2022-07-25
[luci/export] Comment for FLOAT16

seanshpark:
This will add a comment why FLOAT16 tensor is not exported.


seanshpark:
related #9479

jinevening (approved):
LGTM

Issue #9304, opened 2022-06-22
[npud] Introduce initial npud code


jyoungyun:
## What

Introduce initial npud code

Related issue: #9258
Draft PR: #9267

## To do

- [x] Implement event loop core (#9493)
- [x] Implement logger (#9509)
- [x] Implement signal handler (#9507)
- [x] Define HAL interface (#9971)
- [x] Implement Device plugins
   - [x] trix (#10012)
- [ ] ~~Define User APIs~~
- [x] Implement DBus handler (#9929)
   - [x] DBus test (#9952)
- [x] Add DBus configuration file (#10058)

jyoungyun:
I will make a new issue for `Define User APIs`.

jyoungyun:
No more progress.

PR #9509, opened 2022-07-26
[npud] Add Logging class and related config files

jyoungyun:
This commit adds Logging class and related config files.
It turns the log on and off using `NPUD_LOG_ENABLE` environment.

Related issue: #9304

### DIFF
--- a/packaging/nnfw.spec
+++ b/packaging/nnfw.spec
@@ -33,6 +33,7 @@ Source2002: nnfw-plugin.pc.in
 %{!?coverage_build: %define coverage_build  0}
 %{!?test_build:     %define test_build      0}
 %{!?extra_option:   %define extra_option    %{nil}}
+%{!?config_support: %define config_support  1}
 
 %if %{coverage_build} == 1
 # Coverage test requires debug build runtime
@@ -150,12 +151,21 @@ NPU daemon for optimal management of NPU hardware
 %define option_test -DENABLE_TEST=ON -DENVVAR_ONERT_CONFIG=ON
 %endif # test_build
 
+# Set option for configuration
+%define option_config %{nil}
+%if %{config_support} == 1
+%if %{npud_build} == 1
+# ENVVAR_NPUD_CONFIG: Use environment variable for npud configuration and debug
+%define option_config -DENVVAR_NPUD_CONFIG=ON
+%endif # npud_build
+%endif # config_support
+
 %if %{coverage_build} == 1
 %define option_coverage -DENABLE_COVERAGE=ON
 %endif # coverage_build
 
 %define build_options -DCMAKE_BUILD_TYPE=%{build_type} -DTARGET_ARCH=%{target_arch} -DTARGET_OS=tizen -DBUILD_MINIMAL_SAMPLE=ON \\\
-        %{option_test} %{option_coverage} %{extra_option}
+        %{option_test} %{option_coverage} %{option_config} %{extra_option}
 
 %prep
 %setup -q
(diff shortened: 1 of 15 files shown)


hseok-oh:
Tizen security check doesn't allow to use environment variable. So runtime tizen build disabled this feature.

https://github.com/Samsung/ONE/blob/c5f3f873364b93ec59b7b63566c1ea857fb68b46/infra/nnfw/cmake/options/options_armv7l-tizen.cmake#L9-L11

Runtime still uses environment variable because NNAPI doesn't have way to config runtime without environment variable. IMHO, we need to find better way to config `npud`.

jyoungyun:
/cc @hseok-oh 
What about this changes? https://github.com/Samsung/ONE/pull/9509/commits/2acb05f1eedd2722b0df403f39aa606d90644435

packaging/nnfw.spec
```diff
@@ -150,6 +150,13 @@ NPU daemon for optimal management of NPU hardware
 %define option_test -DENABLE_TEST=ON -DENVVAR_ONERT_CONFIG=ON
 %endif # test_build
 
+%if %{build_type} == Debug
+%if %{npud_build} == 1
+# ENVVAR_NPUD_CONFIG: Use environment variable for npud configuration and debug
+%define extra_option ${extra_option} -DENVVAR_NPUD_CONFIG=ON
```
hseok-oh, line 156: Please define different option variable. (ex. `option_debug`?)
BTW, Don't we need to use logging on release build?

jyoungyun: I added `option_config` define. Could you see this patch? https://github.com/Samsung/ONE/commit/02f9c350b9903cc5a7edbed4dab5e9d461e609a6

```
# Set option for configuration
%define option_config %{nil}
%if %{config_support} == 1
%if %{npud_build} == 1
# ENVVAR_NPUD_CONFIG: Use environment variable for npud configuration and debug
%define option_config -DENVVAR_NPUD_CONFIG=ON
%endif # npud_build
%endif # config_support
```

hseok-oh (approved):
This PR has some minor issue, but conceptually LGTM.
I'll check minor issue after merging this PR.
- Code duplication
- Useless code: copied from runtime but some used for nnfw API

runtime/service/npud/core/Server.cc
```diff
@@ -33,7 +33,7 @@ Server::Server() noexcept
 
 void Server::run(void)
 {
-  std::cout << "Starting Server" << std::endl;
+  VERBOSE(Server) << "Starting Server\n";
```
ragmani, line 36: I have a question. Why did you change `std::endl` to `\n`? Was there a problem?

jyoungyun: No, there is no problem. I just follow different code rules.
One runtime uses both of `"blah blah ~~\n"` and `"blah bla~~" << std::endl`.

jyoungyun:
/cc @hseok-oh 

> I'll check minor issue after merging this PR.
> - Code duplication
> - Useless code: copied from runtime but some used for nnfw API

Thank you!

runtime/service/npud/core/Server.cc
ragmani: I was wondering if there was a problem. I'm happy because we would have to change the other codes like this in `onert` if there was a problem. But I think that `std::endl` is better than `\n` because `std::endl` flushes the stream. Anyway this change doesn't matter.

ragmani (approved):
LGTM

PR #9510, opened 2022-07-26
[luci] Fold scalar with Dequantize Op

jinevening:
This enables to fold scalar with Dequantize Op.


llFreetimell (approved):
LGTM

PR #9514, opened 2022-07-26
[DRAFT] Implement one-init

hyunsik-yoon:
This is yet another one-init, which works like the following step:

- generate cfg file for import, optimize, quantization
- ask backend-init to update and save cfg file
- add some more comment


hyunsik-yoon:
This is a modified version of #9421.

compiler/one-cmds/dummy-driver/src/dummy-init.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * ... json output
+ *
+ */
+
+#include <iostream>
+
+const char *output =
+  R"(
+{
+  \"workflow\": [\"one-import\", \"one-optimize\", \"one-quantize\", \"one-codegen\"],
```
seanshpark, line 31: I've searched for this `R("...")`, raw string and looks good feature.
And I think we don't need to use `\"` but just `"` seems OK.

compiler/one-cmds/dummy-init
```diff
@@ -0,0 +1,14 @@
+#/bin/bash
+
+cat << EOF
+{
+    "workflow": ["one-import", "one-optimize", "one-quantize", "one-codegen"],
+    "one-optmize": {
```
seanshpark, line 6: typo `optmize`

compiler/one-cmds/one-init
```diff
@@ -26,12 +26,77 @@ import itertools
...
 import sys
+import configparser
 
 import configparser
 import utils as _utils
 
 # TODO Find better way to suppress trackback on error
 sys.tracebacklimit = 0
+'''
+NOTE
```
seanshpark, line 37: how about `NOTE about backend specific init`?

yunjayh:
I'm going to inherit his work about `one-init`. Let me close and open a new draft and prs. (I don't have write permission to his repositories.) 

Thank you for your effort. :smile: :+1:

Issue #9520, opened 2022-08-03
[one-import-onnx] fix_io_order initializers as inputs


seanshpark:
For some ONNX models, there is a strange case, don't know this is usual or not, where (all) initializers are also registered as inputs.

Let's handle for this case.

jinevening:
From https://github.com/onnx/onnx/blob/master/docs/IR.md,

`
When an initializer has the same name as a graph input, it specifies a default value for that input. When an initializer has a name different from all graph inputs, it specifies a constant value. The order of the list is unspecified.
`

initializer can have the same name with input. If we change the name of an input, we should update the name of the corresponding initializer.

So, I think we should update initializer too, rather than not update input.

seanshpark:
> So, it would be better to update initializer too, rather than not update input.

If the initializer name is changed, kernel of Conv2D for instance becomes an input not a Constant.
This is strange. Maybe there is something I missed...

seanshpark:
The point of this change is not to change initializer names. Do you think this is a problem?

jinevening:
Hmm.. I'm worried about the below case. Please correct me if I missed something.

Before,
Inputs : ['a', 'b']
Initializer = ['a']

After #9521 is applied,
Inputs : ['a', '0001_b']
Initializer = ['a']

After alphabetical sorting, input order will be ['0001_b', 'a'], which is different from the original order.

seanshpark:
I think I found what the problem of existing code.
- current code changes all inputs to something like `NAME` to `nnnn_NAME`
- but did not change `model.graph.initializer`
- so there was no matching `input` to the `initializer`
- and the model became strange like all constants became inputs

To fix this
(1) rename the `initializer` items too
(2) or don't rename the `initializer` items

I think (2) is also valid because order of input is preserved and name of initializers are preserved. input as initializer is referenced by name so connection itself in the graph is not harmed.
(1) is not intended so not necessary.

jinevening:
> input stays in ['a', '0001_b'] as the change don't reorder the graph.

The change only renames the inputs, but the inputs may be reordered after it passes through onnx-tf.

seanshpark:
>  after it passes through onnx-tf.

Yes, it seens so when TF graph is made...
But I don't know order of initialize and input should be considered together.
In TF, initializers will be Buffers and these don't exist with graph inputs.

jinevening:
> In TF, initializers will be Buffers(?) and these don't exist with graph inputs.

[...]

PR #9521, opened 2022-08-03
[one-cmds] Fix one-import-onnx with fix_io_order

seanshpark:
This will fix one-import-onnx with fix_io_order option
to ignore initializers as inputs.


seanshpark:
for #9520

jinevening:
Could you take a look at https://github.com/Samsung/ONE/issues/9520#issuecomment-1203627651 ?

I think it would be better to update initializer too, rather than not update input.

jinevening (approved):
LGTM

PR #9523, opened 2022-08-04
[one-cmds] Add fake quantization option to one-quantize

jinevening:
This adds fake quantization option to one-quantize.

### DIFF
--- a/compiler/one-cmds/one-quantize
+++ b/compiler/one-cmds/one-quantize
@@ -203,6 +203,14 @@ def _get_parser():
     copy_quantparam_group.add_argument(
         '--dst_tensor_name', type=str, action='append', help='tensor name (string)')
 
+    # arguments for fake_quant option
+    fake_quant_group = parser.add_argument_group('arguments for fake_quantize option')
+
+    fake_quant_group.add_argument(
+        '--fake_quantize',
+        action='store_true',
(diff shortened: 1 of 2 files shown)


jinevening:
To reviewers: This PR allows users (ex: python interface) to use fake quantization feature.

llFreetimell (approved):
LGTM

compiler/one-cmds/one-quantize
```diff
@@ -512,6 +525,27 @@ def _copy_qparam(args):
...
+    logfile_path = os.path.realpath(args.output_path) + '.log'
+
+    with open(logfile_path, 'wb') as f:
+        # get driver path
+        circle_quantizer_path = os.path.join(dir_path, 'circle-quantizer')
+        q_model = ""
+        if _utils._is_valid_attr(args, 'input_path'):
+            q_model = getattr(args, 'input_path')
+        if _utils._is_valid_attr(args, 'output_path'):
+            fq_model = getattr(args, 'output_path')
```
seanshpark, line 540: I think we don't have to call `if _utils._is_valid_attr():` for input_path nor output_path as these are checked at `_verify_arg()`

there are more codes that checks again in this file and would be, in another PR, OK to remove them.

llFreetimell (approved):
LGTM

Issue #9488, opened 2022-07-22
Compiler FE: Support Densify operator


seanshpark:
One of our internal model uses `Densify` operator.
Let's enable this.

Info
- https://www.tensorflow.org/guide/sparse_tensor
- ...

seanshpark:
We need to Add `CircleDensify` Op.
- [ ] tflite2circle
- [ ] luci/lang
- [ ] luci/import
- [ ] luci/service
- [ ] luci/partition
- [ ] luci/logex
- [ ] luci/export
- [ ] res recipes
- [ ] ...

seanshpark:
Related information
- TACO format (Tensor Algebra COmpiler) from http://tensor-compiler.org/files/kjolstad-oopsla17-tensor-compiler.pdf
   - https://github.com/tensor-compiler/taco
   - https://youtu.be/Kffbzf9etLE
- https://stackoverflow.com/questions/63719424/understanding-of-sparsityparameters-in-tensorflow-lite-schema

seanshpark:
About sparse tensor in _tflchef_;
- authoring sparse tensor by hand is difficult
- we can borrow codes from TF to generate sparse tensor from dense tensor and vise versa
- ...

seanshpark:
done and left over task in another issue.

PR #9524, opened 2022-08-04
DRAFT Support DENSIFY

seanshpark:
on-going draft to support DENSIFY Op.


seanshpark:
for #9488

seanshpark:
done for now

PR #9532, opened 2022-08-08
[infra/gbs] Remove gbs.conf internal repo

hseok-oh:
This commit removes internal repo definition in gbs.conf.
We will find different way for package sync when version up.

This commit includes clean up gbs.conf file.


hseok-oh:
For security reason, we need to block to access file outside CI infra.

chunseoklee (approved):

PR #9543, opened 2022-08-09
[luci/pass] Extract FP16

seanshpark:
This will extract external FP16 to separate header.


seanshpark:
related #9488, from draft #9524

compiler/luci/pass/src/helpers/FP16.h
```diff
@@ -0,0 +1,202 @@
+/*
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved
```
jinevening, line 2:
```suggestion
 * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
```

jinevening:
Not for this PR, but it would be good to make fp16 as a separate module to reduce duplicate codes with souschef.

jinevening (approved):
LGTM Thank you

Issue #9526, opened 2022-08-05
[one-import-onnx] Rename option fix_io_order


seanshpark:
`fix_io_order` option fixes ONNX models I/O order when it's converted to TensorFlow.
But actully it doesn't fix but keeps the ONNX I/O order to our circle model.
Let's rename this option name with `keep_io_order`

; suggested by @lemmaa

seanshpark:
Steps;
- add `keep_io_order` to act same as existing `fix_io_order`
- update internal CI
- remove `fix_io_order`

jinevening:
How about turning on `keep_io_order` by default? It seems natural to me.

seanshpark:
> How about turning on keep_io_order by default? It seems natural to me.

I understand `by default` is to apply this in `one-import-onnx` without option.
This option modifies input/output names input onnx model.
Technically I don't see there can be any problem with the modification.
But as like all other options, do I don't want to apply this by default as it changes something.
It's up to the user to apply or not. The user can be end user or the user of `one-import-onnx` script.

seanshpark:
done

PR #9546, opened 2022-08-09
[one-import-onnx] Deprecate fix_io_order

seanshpark:
This will remove fix_io_order that was replaced with keep_io_order.


seanshpark:
for #9526

jinevening (approved):
LGTM

PR #9549, opened 2022-08-09
[tflchef] Sparsity format converter

seanshpark:
This will introduce FormatConverter to convert sparse vs dense tensors.


seanshpark:
for #9488, from draft #9524

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

FYI, this is very similar one with `compiler/luci/pass/src/Sparsifier.h`. It can be one file later.

PR #9556, opened 2022-08-09
[luci/pass] FoldDensifyPass

seanshpark:
This will introduce FoldDensifyPass that will convert spase tensor with
Densify Op to densor tensor.

### DIFF
--- a/compiler/luci/pass/src/FoldDensifyPass.cpp
+++ b/compiler/luci/pass/src/FoldDensifyPass.cpp
@@ -1,0 +1,182 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FoldDensifyPass.h"
+#include "helpers/SparsityFormatConverter.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/Profile/CircleNodeOrigin.h>
+
+#include <cassert>
+#include <vector>
+
+#include "helpers/FP16.h"
+
+namespace
+{
+
+bool is_foldable_const(luci::CircleConst *node)
+{
+  if (node->sparsityparam() == nullptr)
+    return false;
+
+  if (node->dtype() == loco::DataType::FLOAT32)
+    return true;
+  if (node->dtype() == loco::DataType::FLOAT16)
+    return true;
+
+  return false;
+}
+
+luci::CircleConst *densified_const_node(luci::CircleConst *const_node)
+{
+  assert(const_node->sparsityparam());
+
+  auto name = const_node->name();
+  assert(name.length() > 0);
+  auto g = const_node->graph();
+  auto new_const_node = g->nodes()->create<luci::CircleConst>();
+
+  new_const_node->dtype(const_node->dtype());
+  new_const_node->rank(const_node->rank());
+
+  uint32_t dim_size = 1;
+  std::vector<int> dense_shape;
+  for (uint32_t i = 0; i < new_const_node->rank(); ++i)
+  {
+    assert(const_node->dim(i).known());
+    new_const_node->dim(i) = const_node->dim(i);
+
+    uint32_t value = const_node->dim(i).value();
+    dim_size *= value;
+    dense_shape.emplace_back(static_cast<int32_t>(value));
+  }
+
+  if (const_node->dtype() == loco::DataType::FLOAT32)
+    new_const_node->size<loco::DataType::FLOAT32>(dim_size);
+  else
+  {
+    assert(const_node->dtype() == loco::DataType::FLOAT16);
+    new_const_node->size<loco::DataType::FLOAT16>(dim_size);
+  }
+
+  new_const_node->shape_status(luci::ShapeStatus::VALID);
+  new_const_node->name(name + "_DS");
+
+  if (const_node->dtype() == loco::DataType::FLOAT32)
+  {
+    auto const_items = const_node->size<loco::DataType::FLOAT32>();
+    auto f_data = std::make_unique<float[]>(const_items);
+    for (size_t i = 0; i < const_items; ++i)
+      f_data[i] = const_node->at<loco::DataType::FLOAT32>(i);
+
+    sparsity::TfLiteSparsity sp = to_tflite_sparsity(const_node->sparsityparam());
+    sparsity::FormatConverter<float> converter(dense_shape, sp);
+    converter.SparseToDense(f_data.get());
+    const auto &data_dense = converter.GetData();
+    assert(data_dense.size() == dim_size);
+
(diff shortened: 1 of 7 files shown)


seanshpark:
for #9488, from draft #9524

compiler/luci/pass/src/FoldDensifyPass.cpp
```diff
@@ -0,0 +1,193 @@
...
+  {
+    assert(const_node->dtype() == loco::DataType::FLOAT16);
+
+    auto const_items = const_node->size<loco::DataType::FLOAT16>();
+    auto f_data = std::make_unique<uint16_t[]>(const_items);
+    for (size_t i = 0; i < const_items; ++i)
+    {
+      f_data[i] = const_node->at<loco::DataType::FLOAT16>(i);
+      float vv = fp16::fp16_ieee_to_fp32_value(f_data[i]);
+      printf("sparse %lu %f\r\n", i, vv);
```
jinevening, line 110: Better to use our logging API?

jinevening: If const_items is large, e.g., a million, log message will be very long.

compiler/luci/pass/src/FoldDensifyPass.cpp
```diff
@@ -0,0 +1,193 @@
...
+    sparsity::TfLiteSparsity sp = to_tflite_sparsity(const_node->sparsityparam());
+    sparsity::FormatConverter<uint16_t> converter(dense_shape, sp);
+    converter.SparseToDense(f_data.get());
+    const auto &data_dense = converter.GetData();
+    assert(data_dense.size() == dim_size);
+    for (uint32_t i = 0; i < dim_size; ++i)
+    {
+      new_const_node->at<loco::DataType::FLOAT16>(i) = data_dense[i];
+      float vv = fp16::fp16_ieee_to_fp32_value(data_dense[i]);
+      printf("dense %u %f\r\n", i, vv);
```
jinevening, line 123: ditto

compiler/luci/pass/src/FoldDensifyPass.cpp
```diff
@@ -0,0 +1,193 @@
...
+  auto const_input = dynamic_cast<luci::CircleConst *>(densify->input());
+  if (not const_input)
+    return false;
+
+  if (not is_foldable_const(const_input))
+    return false;
+
+  auto dense_const = densified_const_node(const_input);
+  if (not dense_const)
+    return false;
```
jinevening, line 146: It seems that `densified_const_node` does not return nullptr. So this check can be replaced to `assert`.

compiler/luci/pass/src/FoldDensifyPass.test.cpp
```diff
@@ -0,0 +1,158 @@
...
+  {
+    // fill 4x4 of
+    //  [[1 0 0 0]
+    //   [0 2 0 0]
+    //   [0 0 3 0]
+    //   [0 0 0 4]]
+
+    // values of 1.0, 2.0, 3.0, 4.0
+    uint32_t udata[] = {0x3f800000, 0x40000000, 0x40400000, 0x40800000};
+    float *fdata = reinterpret_cast<float *>(udata);
```
jinevening, line 72: Why do you use `uint32_t` instead of `float`? Does this change raise error?

```suggestion
    float fdata[] = {1.0, 2.0, 3.0, 4.0};
```

compiler/luci/pass/src/helpers/SparsityFormatConverter.cpp
```diff
@@ -0,0 +1,312 @@
...
+{
+  data_.resize(dense_size_);
+  std::fill(data_.begin(), data_.end(), T(0));
+
+  int total_rank = traversal_order_.size();
+  int src_data_ptr = 0;
+  std::vector<int> indices(total_rank);
+  Populate(src_data, indices, 0, 0, &src_data_ptr, data_.data());
+
+  return true;
```
jinevening, line 210: This function always returns true?

compiler/luci/pass/src/helpers/SparsityFormatConverter.cpp
```diff
@@ -0,0 +1,312 @@
...
+{
+
+sparsity::TfLiteDimensionType to_tflite_sparsity(luci::DimensionType dt)
+{
+  switch (dt)
+  {
+    case luci::DimensionType::DENSE:
+      return sparsity::TfLiteDimensionType::kTfLiteDimDense;
+    case luci::DimensionType::SPARSE_CSR:
+      return sparsity::TfLiteDimensionType::kTfLiteDimSparseCSR;
```
jinevening, line 230: It would be better to throw an exception in the default case to track a potential error.

compiler/luci/pass/src/helpers/SparsityFormatConverter.h
```diff
@@ -0,0 +1,129 @@
...
+{
+  kTfLiteDimDense = 0,
+  kTfLiteDimSparseCSR,
+} TfLiteDimensionType;
+
+// Fixed size list of integers. Used for dimensions and inputs/outputs tensor
+// indices
+typedef struct TfLiteIntArray
+{
+  int size;
```
jinevening, line 43: Our SE tool may complain about uninitialized member variables.

compiler/luci/pass/src/helpers/SparsityFormatConverter.h
```diff
@@ -0,0 +1,129 @@
...
+  int data[];
+} TfLiteIntArray;
+
+// Metadata to encode each dimension in a sparse tensor.
+typedef struct TfLiteDimensionMetadata
+{
+  TfLiteDimensionType format;
+  int dense_size;
+  TfLiteIntArray *array_segments;
+  TfLiteIntArray *array_indices;
```
jinevening, line 53: ditto

compiler/luci/pass/src/helpers/SparsityFormatConverter.h
```diff
@@ -0,0 +1,129 @@
...
+} TfLiteDimensionMetadata;
+
+// Parameters used to encode a sparse tensor. For detailed explanation of each
+// field please refer to lite/schema/schema.fbs.
+typedef struct TfLiteSparsity
+{
+  TfLiteIntArray *traversal_order;
+  TfLiteIntArray *block_map;
+  TfLiteDimensionMetadata *dim_metadata;
```
jinevening, line 62: ditto

compiler/luci/pass/src/FoldDensifyPass.test.cpp
seanshpark: Using `float fdata[] = {1.0, 2.0, 3.0, 4.0};` is also OK.
I intentionally used uint32_t to use byte values.

compiler/luci/pass/src/helpers/SparsityFormatConverter.cpp
seanshpark: Yes, this is as-is from tensorflow.

compiler/luci/pass/src/helpers/SparsityFormatConverter.cpp
seanshpark: This is as-is from tensorflow.

compiler/luci/pass/src/helpers/SparsityFormatConverter.h
seanshpark: This is as-is from tensorflow.
I'll fix this when SE tool complains.

compiler/luci/pass/src/helpers/SparsityFormatConverter.h
seanshpark: This is as-is from tensorflow.
I'll fix this when SE tool complains.

compiler/luci/pass/src/helpers/SparsityFormatConverter.h
seanshpark: This is as-is from tensorflow.
I'll fix this when SE tool complains.

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #9557, opened 2022-08-09
[common-artifacts] Exclude for Densify

seanshpark:
This will add exclude recipes for Densify operator.

### DIFF
--- a/compiler/common-artifacts/exclude.lst
+++ b/compiler/common-artifacts/exclude.lst
@@ -32,6 +32,7 @@ tcgenerate(BroadcastTo_000) # luci-interpreter doesn't support custom operator
 tcgenerate(Ceil_000)
 tcgenerate(Conv2D_003) # runtime doesn't support dilation
 tcgenerate(Cos_000)
+tcgenerate(Densify_000) # luci-interpreter doesn't support
 tcgenerate(DepthwiseConv2D_001) # runtime doesn't support dilation
 tcgenerate(DepthwiseConv2D_003) # runtime doesn't support dilation
 tcgenerate(DepthwiseConv2D_U8_001)  # luci-interpreter doesn't support channel-wise quantization yet
@@ -67,6 +68,8 @@ tcgenerate(Neg_000)
 tcgenerate(Net_BroadcastTo_AddV2_001) # luci-interpreter doesn't support custom operator
 tcgenerate(Net_Conv_FakeQuant_000) # luci-interpreter doesn't support FakeQuant yet
 tcgenerate(Net_Dangle_001)
+tcgenerate(Net_Densify_Add_000) # luci-interpreter doesn't support Densify yet
+tcgenerate(Net_Densify_Dequantize_Add_000) # luci-interpreter doesn't support Densify/Dequantize yet
 tcgenerate(Net_Gather_SparseToDense_AddV2_000) # luci-interpreter doesn't support custom operator
 tcgenerate(Net_ZeroDim_001) # luci-interpreter doesn't support zero dim
 tcgenerate(OneHot_000)


seanshpark:
for #9488, from draft #9524

compiler/common-artifacts/exclude.lst
```diff
@@ -6,6 +6,7 @@
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
 optimize(UnidirectionalSequenceLSTM_001) # This recipe contains is_variable Tensor
+optimize(Net_Densify_Dequantize_Add_000) # This recipe contains Dequantize
```
jinevening, line 9: Maybe?

```suggestion
optimize(Net_Densify_Dequantize_Add_000) # This recipe contains Densify
```

compiler/common-artifacts/exclude.lst
```diff
@@ -6,6 +6,7 @@
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
 optimize(UnidirectionalSequenceLSTM_001) # This recipe contains is_variable Tensor
+optimize(Net_Densify_Dequantize_Add_000) # Export of Dequantize which intentionally not suppoerted yet
```
jinevening, line 9: AFAIK, export of dequantize is supported. Export of 'float16' is not supported. Please correct me if I'm wrong.

seanshpark: Ah, you're right! I think I need some coffee -_-;

seanshpark: in `common-artifacts, `*.opt.circle` is generated with `--resolve_customop_add --fold_dequantize` options with `circle2circle` in `CMakeLists.txt` script.
- `--resolve_customop_add` was added when I removed `-O1` option
- `--fold_dequantize` was added because I had no solution as of now to test models having `Float16 - Dequantize` model.

with this, `Net_Dequantize_Add_000` with `FLOAT16` is OK.
thus, that is, reason of `FLOAT16` is not the main reason for excluding `Net_Densify_Dequantize_Add_000`.

The first comment I wrote about `Densify` is that I didn't add `--fold-densify`.

As of now, I'm not sure which is the solution of `Net_Densify_Dequantize_Add_000` test in `common-artifacts`.

First I'll rollback to the message with `Densify`.
If you have any suggestion with this please advise
(1) we can add also `--fold-densify` in `circle2circle` test and remove this exclusion
(2) go as like this PR

jinevening: Thanks for the explanation.

To sum up, for `Net_Dequantize_Add_000`, Float16 + Dequantize is folded to Float32, so it is ok.

For `Net_Densify_Dequantize_Add_000`, Float16 + Densify + Dequantize is not folded, so it generates error. I thought that the reason is export of Float16, but it could be seen as folding of Float16 + Densify.

Maybe adding the advice (`turn on --fold_densify to enable this recipe`) would be enough.

seanshpark: OK, let's go with `--fold_densify` too

mhs4670go: Hmmm.. 

> Maybe adding the advice (`turn on --fold_densify to enable this recipe`) would be enough.

For the time being, what @jinevening said above can be a temporal solution.

And, I think it would be better to make it possible to manipulate the options for `circle2circle`. There are lots of canidate that we should consider though. 

I'm thinking of using `options.lst`. We could introduce `optimize_option` macro.

```bash
$ cat compiler/common-artifacts/options.lst

..
optimize_option(Net_Densify_Dequantize_Add_000 --fold_densify)
```

seanshpark: OK, I'll add `--fold_densify` for temporary :)

mhs4670go (approved):
LGTM

jinevening (approved):
LGTM Thank you

PR #9566, opened 2022-08-10
[tflchef] Test recipe of make sparse float16

seanshpark:
This will introduce test recipe of make sparse with float16.


seanshpark:
for #9488, from draft #9524

mhs4670go (approved):
LGTM

PR #9567, opened 2022-08-10
[luci-pass-value-test] Enable multiple options

seanshpark:
This will enable to test with multiple optimization options.


seanshpark:
for #9488, from draft #9524

mhs4670go (approved):
LGTM

PR #9573, opened 2022-08-12
[luci/pass] use delete []

seanshpark:
This will revise to use delete [] to make analysis tool happy.


mhs4670go (dismissed):
LGTM

mhs4670go (approved):

PR #9574, opened 2022-08-12
[luci] Check if nullptr after dynamic casting

mhs4670go:
This commit checks if casted ptr is nullptr.

This is for resolving static analysis tool issue.


seanshpark (approved):
LGTM

compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
```diff
@@ -1999,6 +1999,7 @@ TEST(ConvertNCHWToNHWC, SplitV)
 
   // Check axis
   auto axis = dynamic_cast<luci::CircleConst *>(g.splitv()->split_dim());
+  EXPECT_NE(nullptr, axis);
```
seanshpark, line 2002: Q) have you tested with this change?

mhs4670go: Yes, I have:)

Issue #9548, opened 2022-08-09
Compiler FE: Provide common module for FP16


seanshpark:
> Not for this PR, but it would be good to make fp16 as a separate module to reduce duplicate codes with souschef.

by @jinevening from https://github.com/Samsung/ONE/pull/9543#issuecomment-1208845124

seanshpark:
We already have `infra/cmake/packages/Fp16SourceConfig.cmake` :)

seanshpark:
What was used in souschef was `0a92994d729ff76a58f692d3028ca1b64b145d91`
infra/cmake/packages/Fp16SourceConfig.cmake uses `3c54eacb74f6f5e39077300c5564156c424d77ba`
which has 2 commit changes but code itself has no difference.

seanshpark:
done

PR #9578, opened 2022-08-16
DRAFT use FP16 from infra

seanshpark:
on-going draft to revise to use FP16 from infra/cmake/packages.


seanshpark:
for #9548

seanshpark:
done

PR #9584, opened 2022-08-16
[souschef] Remove not used FP16.h

seanshpark:
This will remove not used anymore FP16.h header file.


seanshpark:
for #9548, from draft #9578

jinevening (approved):
LGTM

PR #9595, opened 2022-08-18
[onnx-tools] Install tools

seanshpark:
This will revise to install tools with install step.


seanshpark:
FYI; tools are not included in package release. currently it is for developers only.

mhs4670go (approved):
LGTM

PR #9596, opened 2022-08-18
[luci-interpreter] Add reference_ops.h to PAL

SlavikMIPT:
This commit adds reference_ops.h to PAL (with fixes to support builds for cortex mcus).
- fixed errors like `"no matching function for call to 'min(const int&, long int)"`

needed for https://github.com/Samsung/ONE/issues/9468
ONE-DCO-1.0-Signed-off-by: Vyacheslav Bazhenov <v.bazhenov@partner.samsung.com>

### DIFF
--- a/compiler/luci-interpreter/pal/cmsisnn/PALreference_ops.h
+++ b/compiler/luci-interpreter/pal/cmsisnn/PALreference_ops.h
@@ -1,0 +1,1568 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


compiler/luci-interpreter/pal/cmsisnn/PALreference_ops.h
```diff
@@ -0,0 +1,1568 @@
+/*
```
AShedko, line 1: You may want to use a relative symlink to avoid code duplication (git allows it, but IDK if the repo policy does)

AShedko (approved):
LGTM otherwise

compiler/luci-interpreter/pal/cmsisnn/PALreference_ops.h
AShedko: NVM, the file contents are different.

binarman (approved):
LGTM, Could you clarify what is changed to fix MCU build in more detail?
Like "fixed assignment of long to int", "fixed comparison of signed and unsigned integers", "removed compiler specific attributes", etc.

seanshpark:
(optional) for new `PALreference_ops.h` files, we can move 
`Copyright 2017 The TensorFlow Authors. All Rights Reserved.` line to below `* Copyright (c) 2022 Samsung ...` and remove rest duplicate comments as license are the same.

PR #9598, opened 2022-08-19
[luci] Perform fuse_add_with_fc earlier than ConvertNCHWToNHWC

jinevening:
This performs fuse_add_with_fc earlier than ConvertNCHWToNHWC.


seanshpark:
code itself looks OK but it would be nice if you could add some comments behind this change.

seanshpark (approved):
LGTM thank you!

PR #9599, opened 2022-08-19
[one-cmds] Fix OptionBuilder

jinevening:
This enables to use correct optimization options.

### DIFF
--- a/compiler/one-cmds/onelib/OptionBuilder.py
+++ b/compiler/one-cmds/onelib/OptionBuilder.py
@@ -60,7 +60,7 @@ class OptionBuilder:
                 options.extend(['--' + k])
                 continue
             for opt in CONSTANT.OPTIMIZATION_OPTS:
-                if k == opt and v == "True":
+                if k == opt[0] and v == "True":
                     options.extend(['--' + k])
                     break
         return options


compiler/one-cmds/onelib/OptionBuilder.py
```diff
@@ -60,7 +60,7 @@ def _build_optimize(self, commands):
                 options.extend(['--' + k])
                 continue
             for opt in CONSTANT.OPTIMIZATION_OPTS:
-                if k == opt and v == "True":
+                if k == opt[0] and v == "True":
```
seanshpark, line 63: (this is not related with this PR but... @mhs4670go , iv `v` is not related with this for loop, can we hoist `v == 'True'` outside of this loop? in another PR)

seanshpark:
`opt` and `opt[0]` looks very different, but does python work as same?
how did it work?

jinevening:
Previously, optimization options were not applied at all. So this is a bug fix.

seanshpark:
> Previously, optimization options were not applied at all. So this is a bug fix.

Wow... this is a serious thing...
was the usage like
```
onecc --optiona --optionb ...
```
but `optiona` nor `optionb` was applied for optimize ?

compiler/one-cmds/onelib/OptionBuilder.py
mhs4670go: Sorry. I don't get it. Could you explain what you said with simple example?

mhs4670go (approved):
LGTM

mhs4670go:
This is a bug of workflows features. I think related tests should be added.

compiler/one-cmds/onelib/OptionBuilder.py
seanshpark: as `v` is not related any variable inside the loop,
```py
if v == "True":
  for opt in CONSTANT.OPTIMIZATION_OPTS:
    if k == opt[0]:
      options.extend(['--' + k])
      break
```

seanshpark (approved):
LGTM

compiler/one-cmds/onelib/OptionBuilder.py
mhs4670go: We can:) I will post a PR soon.

PR #9603, opened 2022-08-19
[one-cmds] Add workflow test

mhs4670go:
This commit adds workflow test where it checks if optimize options are applied well.


jinevening (approved):
LGTM

BTW, this test is not for optimization options, but for utility option.

seanshpark (approved):
LGTM

PR #9619, opened 2022-08-24
[one-cmd] Add section, input_path, output_path for one-init

hyunsik-yoon:
This commit enables one-init to add sections and
input_path and output_path of each section.

for #9450
draft #9514

### DIFF
--- a/compiler/one-cmds/one-init
+++ b/compiler/one-cmds/one-init
@@ -365,7 +365,6 @@ def main():
     # verify arguments
     _verify_arg(parser, args)
 
-    #TODO generate cfg file
     model_type = _get_model_type(parser, args)
     inout_path = InputOutputPath(args.input_path)
     _generate(args, model_type, inout_path)
--- a/compiler/one-cmds/one-init
+++ b/compiler/one-cmds/one-init
@@ -34,6 +34,38 @@ import utils as _utils
 sys.tracebacklimit = 0
 
 
+class InputOutputPath:
+    '''
+    Class that remembers input circle file and output circle file of section k,
+
+    After calling enter_new_section(),
+    output path in section k will be used as input path of section k+1
+    '''
+
+    def __init__(self, initial_input_path: str):
+        self._first_step = True
+        self._input_path = initial_input_path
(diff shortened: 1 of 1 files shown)


compiler/one-cmds/one-init
```diff
@@ -275,6 +365,11 @@ def main():
     # verify arguments
     _verify_arg(parser, args)
 
+    #TODO generate cfg file
```
yunjayh, line 368: AFAIU, `_generate()` writes cfg file. What does `TODO` in the comment mean?

hyunsik-yoon: Thanks. I removed!

compiler/one-cmds/one-init
```diff
@@ -235,16 +284,40 @@ def _generate():
         pass  # NYI
 
     def _gen_import():
-        pass  # NYI
+        section = f'one-import-{model_type}'
```
seanshpark, line 287: Q) where is `model_type` from?

seanshpark: Ah, `_gen_import` is inner function of `_generate`... I'm not good with python and learned now about this :)

compiler/one-cmds/one-init
```diff
@@ -264,6 +337,23 @@ def _generate():
...
+        _, ext = os.path.splitext(args.input_path)
+
+        # ext would be, e.g., '.tflite' or '.onnx'.
+        # Note: when args.input_path does not have an extension, e.g., '/home/foo'
+        # ext after os.path.splitext() is '' and ''[1:] is still ''.
+        ext = ext[1:]
+        if ext in ["tflite", "onnx"]:
+            return ext
+
+    parser.error(f'the following argument is required: --input_path')
```
seanshpark, line 354: `f'...'` is used but seems no variable used for format.

seanshpark: if `ext` is not one of `tflite` nor `onnx` this error will be shown even if `--input_path` is given, which seems logical error.

yunjayh:
Substituted to #9813.

PR #9620, opened 2022-08-24
[one-cmd] Remove useless code from one-init

hyunsik-yoon:
This removes useless code:
1. remove `_add_from_backend_cfg()`
- Previously, there were two cfg output files:
  One was output of one-init, and another was output of backend-init.
  In this approach, after bckend-init creates its own cfg,
  one-init reads the cfg file of backend-init and write them into the output of one-init. --- (A)
  `_add_from_backend_cfg()` was created for (A) and it is now useless.

  Now, the changed way works like this:
  output cfg of one-init is used as input of backend-init,
  and backend-init generates the final output cfg.

2. remove `_gen_profile()`
  Let's not handle this at the current version.
  Comment at L242 already mentioned this.

3. handling 'command' arg was copied from other file
   and not used in one-init.

for https://github.com/Samsung/ONE/issues/9450
draft https://github.com/Samsung/ONE/pull/9514

### DIFF
--- a/compiler/one-cmds/one-init
+++ b/compiler/one-cmds/one-init
@@ -226,14 +226,6 @@ def _generate():
     def _add_onecc_sections():
         pass  # NYI
 
-    def _add_from_backend_cfg():
-        '''
-        This copies key & value in sections of backend cfg into one cfg.
-        Call this after setting default values for circle since there might be values overrided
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

Issue #9609, opened 2022-08-22
[onert] NYI CPU backend kernel - uint8, per-channel


glistening:
- [x] CONV2D
- [x] DEPTHWISE_CONV2D

`onert` cannot run per-channel quantized mobilenet v1 circle.

It fails at the 1st op due to lack of kernel.

```
$ ./build/install/bin/circledump 0c/0.circle
Dump: 0c/0.circle

===================================================================
Model version: 0
 # sub graphs: 1

Operator Codes: [order] OpCodeName (OpCode Enum)
[0] CONV_2D (code: 3, version: 1)

Buffers: B(index) (length) values, if any
B(0) (0)
B(1) (0)
B(2) (864) 0xe0 0x7c 0x40 0xfa 0x8b 0x4e 0xff 0x87 0x40 0xa4 0x25 0xa 0xbf 0x4f 0x1e 0x8e ...
B(3) (128) 0x73 0x4d 0xd1 0xff 0xab 0x75 00 00 0x15 0xf8 0xff 0xff 0x92 0x7e 0xef 0xff ...
B(4) (0)

-------------------------------------------------------------------
Sub-Graph: #0 (noname)

Data Format:
CHANNEL_LAST (NHWC for 2d, NDHWC for 3d data)

Operands: T(subgraph index : tensor index) TYPE (shape) (shape_signature) B(buffer index) OperandName
T(0:0) UINT8 (1, 224, 224, 3) B(1) input
    Quantization: scale(0.0392156) zeropt(128) quantized_dimension(0)

T(0:1) UINT8 (32, 3, 3, 3) B(2) MobilenetV1/Conv2d_0/weights
    Quantization: scale((32) 1.03063e-05, 0.00192652, 0.00438276, 1e-05 ... )
                  zeropt((32) 166, 189, 71, 123 ... )
                  quantized_dimension(0)

T(0:2) INT32 (32) B(3) MobilenetV1/MobilenetV1/Conv2d_0/Conv2D_bias
    Quantization: scale((32) 4.04168e-07, 7.55498e-05, 0.000171873, 3.92156e-07 ... )
                  zeropt((32) 0, 0, 0, 0 ... )
                  quantized_dimension(0)

T(0:3) UINT8 (1, 112, 112, 32) B(4) MobilenetV1/MobilenetV1/Conv2d_0/Relu6
    Quantization: scale(0.0235294) zeropt(0) quantized_dimension(0)

[...]

PR #9633, opened 2022-08-29
[nnpkg_tools/gen_golden] Use same range for float random

glistening:
It changse the default range for float random input.
Previosly, [0, 1) to [-5, 5), which is the same to RecordMinMax.

Related: https://github.com/Samsung/ONE/issues/9609#issuecomment-1227983715

The range for random input depends on each model. As I remember, some model expected the input [0, 1.0).
In spite of this fact, I am trying to use the same one to `RecordMinMax` which is the tool we use to quantize tflite float to circle because I am guessing it is more likely to meet a value test error, and it may be hard to find the reason who is not familiar with our internal tools.

### DIFF
--- a/tools/nnpackage_tool/gen_golden/gen_golden.py
+++ b/tools/nnpackage_tool/gen_golden/gen_golden.py
@@ -96,7 +96,7 @@ if __name__ == '__main__':
                     np.random.randint(-127, 127, this_shape).astype(np.int8))
             elif this_dtype == tf.float32:
                 input_values.append(
-                    np.random.random_sample(this_shape).astype(np.float32))
+                    (10 * np.random.random_sample(this_shape) - 5).astype(np.float32))
             elif this_dtype == tf.bool:
                 # generate random integer from [0, 2)
                 input_values.append(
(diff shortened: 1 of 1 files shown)


chunseoklee (approved):

hseok-oh (approved):

glistening:
@jinevening PTAL

tools/nnpackage_tool/gen_golden/gen_golden.py
```diff
@@ -96,7 +96,7 @@ def usage():
                     np.random.randint(-127, 127, this_shape).astype(np.int8))
             elif this_dtype == tf.float32:
                 input_values.append(
-                    np.random.random_sample(this_shape).astype(np.float32))
+                    5 * np.random.random_sample(this_shape).astype(np.float32))
```
jinevening, line 99: https://numpy.org/doc/stable/reference/random/generated/numpy.random.random_sample.html

```
Return random floats in the half-open interval [0.0, 1.0).
```
Maybe `10 * np.random.random_sample(this_shape) - 5` ?

glistening:
@jinevening, @hseok-oh, @chunseoklee PTAL.

jinevening (approved):
LGTM

chunseoklee (approved):

PR #9644, opened 2022-08-30
[onert] Renames methods in dconv to more proper names

glistening:
It is the first step for introducing dconv per-channel uint8 kernel.

### DIFF
--- a/runtime/onert/backend/cpu/ops/DepthwiseConvolutionLayer.cc
+++ b/runtime/onert/backend/cpu/ops/DepthwiseConvolutionLayer.cc
@@ -49,7 +49,7 @@ void DepthwiseConvolutionLayer::convFloat32()
     getBuffer<float>(_output), _external_context->ruy_context());
 }
 
-void DepthwiseConvolutionLayer::convQuant8()
+void DepthwiseConvolutionLayer::convQ8uPerTensor()
 {
   int32_t output_activation_min = 0;
   int32_t output_activation_max = 0;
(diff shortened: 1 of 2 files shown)


runtime/onert/backend/cpu/ops/DepthwiseConvolutionLayer.cc
```diff
@@ -84,11 +84,11 @@ void DepthwiseConvolutionLayer::convQuant8()
     getBuffer<uint8_t>(_output), _external_context->ruy_context());
 }
 
-void DepthwiseConvolutionLayer::convQuant8PerChannel()
+void DepthwiseConvolutionLayer::convQ8i()
```
chunseoklee, line 87: PerChannel sounds proper for me. What does suffix `i` mean here ?

glistening: 1) I would like to distinguish `uint8` and `int8`. `i` means `int8`. `u` means `uint8`.
2) `convQuant8PerChannel` is `int8` kernel which can handle both per-tensor and per-channel. Meanwhile, `uint8` kernel uses two different kernels for per-tensor and per-channel respectively.

runtime/onert/backend/cpu/ops/DepthwiseConvolutionLayer.cc
```diff
@@ -169,11 +169,11 @@ void DepthwiseConvolutionLayer::run()
   }
   else if (_input->data_type() == OperandType::QUANT_UINT8_ASYMM)
   {
-    convQuant8();
+    convQ8uPerTensor();
```
glistening, line 172: In next PR, there will be `if` statement to check per-channel or per-tensor as I did for `conv2d`.

chunseoklee (approved):

hseok-oh (approved):

PR #9656, opened 2022-09-01
[onert] Use non-negative zeropoint for conv2d uint8 output

glistening:
It changes the zeropoint for output of uint8 per-channel conv testcase
from negative to [0, 255].


hseok-oh (approved):
LGTM

PR #9635, opened 2022-08-29
[Draft] Introduce `circle-interpreter`

yunjayh:
This commit introduces `circle-interpreter` as driver of `luci-interpreter`.

Signed-off-by: yunjay-hong <yunjay.hong@samsung.com>

---

for #9634

Issue #9634, opened 2022-08-29
Introduce `circle-interpreter` as driver of `luci-interpreter`


yunjayh:
### What

Let's publish `circle-interpreter` as driver of luci-interpreter. 

### Why

Currently, `luci_eval_driver` is published in `one-compiler-test` package which is mostly used for test. Here are few reasons to introduce a new binary, `circle-interpreter` for package. 

  * `luci-eval-driver` has `num_inputs` argument, which can be derived from circle model. Let's reduce the end-user's effort.
  * `luci`, which is a internal module name, is not proper to exposed to end-user. 
  * some internal tests are using `luci-eval-driver`. For theirs compatibility, it needs not to deprecate or fix arguments.

yunjayh:
TODO: 

> Is this behavior (succeed for larger size input data) intended?

IMHO, it would be better to generate an error (ex: throw an exception) in `circle-interpreter` when the size of input data mismatches with the size of the model input. There may be a way to get the size of input file in cpp code.

So, it would be better to remove this test, because error message may not be the same with the current one.

_Originally posted by @jinevening in https://github.com/Samsung/ONE/pull/9658#discussion_r972605270_

seanshpark:
@yunjayh , I think we need to add circle-interpreter and test to `infra/scripts/compiler_modules.sh`
CC @hseok-oh

yunjayh:
I read the script and related scripts, and here is a question about it. 

`compiler_modules.sh` is only referenced in `infra/scripts/configure_compiler_coverage.sh`. However I cannot find `configure_compiler_coverage.sh` in this repository anymore. Could you please share what and why the script is needed?

seanshpark:
ping @hseok-oh

seanshpark:
https://github.com/Samsung/ONE/blob/master/infra/scripts/configure_compiler_coverage.sh

yunjayh:
It's been confirmed that nightly released package successfully included `circle-interpreter`. Thank you for your kind help and reviews, everyone.

```text
$ onecc infer -d circle-interpreter -- -h
What circle-interpreter does: Interpreter driver for circle models

Usage: ./circle-interpreter [-h] [--version] model_path input_prefix output_prefix 

[Positional argument]
model_path   	Circle model filepath
input_prefix 	Input data filepath for circle model. n-th input data is rea
             	d from ${input_prefix}n, for example, Add.circle.input0, Add
             	.circle.input1
[...]

PR #9658, opened 2022-09-01
[circle-interpreter-test] Add circle-interpreter-test

yunjayh:
This commit adds `circle-interpreter-test` to test `circle-interpreter`'s functionality.

---

for: #9634 
draft: #9635


compiler/circle-interpreter-test/eval_driver.sh
```diff
@@ -0,0 +1,66 @@
...
+  PASSED_TAG="${TEST_RESULT_FILE}.passed"
+  rm -f "${PASSED_TAG}"
+
+  cat > "${TEST_RESULT_FILE}.log" <(
+    exec 2>&1
+    set -ex
+
+    source "${VIRTUALENV}/bin/activate"
+
+    "${VIRTUALENV}/bin/python" "${VERIFY_SCRIPT_PATH}" \
```
seanshpark, line 40: whre is `VERIFY_SCRIPT_PATH` ?

compiler/circle-interpreter-test/README.md
```diff
@@ -0,0 +1,16 @@
+# circle-interpreter-test
+
+`circle-interpreter-test` executes `circle-interpreter` and validates its results with tflite model's output.
```
seanshpark, line 3: IMHO, this test would be better to check valid/invalid input arguments.

compiler/circle-interpreter-test/eval_result_verifier.py
```diff
@@ -0,0 +1,109 @@
...
+                raise SystemExit("Execution result of " + tflite_model +
+                                 " does not match with " + circle_model)
+        elif output_details["dtype"] == np.int32:
+            if np.allclose(luci_output_data, intp_output_data, rtol=0, atol=0) == False:
+                raise SystemExit("Execution result of " + tflite_model +
+                                 " does not match with " + circle_model)
+        elif output_details["dtype"] == np.int16:
+            if np.allclose(luci_output_data, intp_output_data, rtol=0, atol=0) == False:
+                raise SystemExit("Execution result of " + tflite_model +
+                                 " does not match with " + circle_model)
```
jinevening, line 102: How about fusing some branches to reduce duplicate codes?

compiler/circle-interpreter-test/requires.cmake
```diff
@@ -0,0 +1,4 @@
+require("common-artifacts")
+require("luci-interpreter")
+require("safemain")
```
jinevening, line 3: Seems unnecessary

compiler/circle-interpreter-test/README.md
yunjayh: I tried to check the input arguments' validity through `one-cmds` test, via `one-infer`. Do you think it's better to check it here, and do something else in `one-infer` test? (I have no idea what to do for `one-infer` test if verifying arguments are done here.)

compiler/circle-interpreter-test/eval_result_verifier.py
```diff
@@ -0,0 +1,109 @@
...
+
+    interpreter.set_tensor(input_details["index"], input_data)
+    input_data.tofile(circle_model + ".input" + str(i))
+
+# Do inference
+interpreter.invoke()
+
+# Execute circle interpreter.
+subprocess.run(
+    [driver, circle_model, circle_model + ".input", circle_model + ".output"], check=True)
```
jinevening, line 66: Hmm.. I guess that this test will save outputs under the `common-artifacts` directory. Am I right? If so, it can conflict with other tests.

How about making this script receive `circle_model` and `output_prefix` separately, so that it can save outputs under `circle-interpreter-test` directory? You can find similar codes in `luci-pass-value-test`.

compiler/circle-interpreter-test/eval_result_verifier.py
yunjayh: I divided the dtypes to two groups, `int_type` and `float_type` to distinguish their tolerance value.

compiler/circle-interpreter-test/eval_result_verifier.py
yunjayh: Great point to fix. Thank you for pointing out!

compiler/circle-interpreter-test/README.md
seanshpark: If so, how about adding some comments about argument validity test?

yunjayh: I added it to README :)

compiler/circle-interpreter-test/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+macro(addeval NAME)
+  list(APPEND CIRCLE_INTERPRETER_TESTS ${NAME})
+endmacro(addeval)
+
+# Read "test.lst"
+include("test.lst")
+# Read "test.local.lst" if exists
+include("test.local.lst" OPTIONAL)
+
+add_custom_target(circle_interpreter_test_files ALL DEPENDS ${TEST_DEPS})
```
mhs4670go, line 19: This is not used because this method is for custom command.

compiler/circle-interpreter-test/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+add_dependencies(circle_interpreter_test_files common_artifacts_deps)
```
mhs4670go, line 20: IIUC, this is not needed as well. Because what it does is already being done by `requires.cmake`.

yunjayh:
Yes I am. I'm trying to add test with ctest, but it takes time since I have a little knowledge about both cmake, ctest and bash scripts. 

I was trying to add  command line tests including argument verification through ctest. I think it would be done in a day.. 😓

seanshpark:
> I think it would be done in a day.. 

Please take your time :)

compiler/circle-interpreter-test/README.md
```diff
@@ -0,0 +1,18 @@
+# circle-interpreter-test
```
yunjayh, line 1: README should be revised

compiler/circle-interpreter-test/run_command.sh
```diff
@@ -0,0 +1,68 @@
...
+# ./run_command.sh <path/to/intp_dir>
+#                  <TEST NAME #1> <TEST TYPE #1> <CMD TO TEST #1> \
+#                  <TEST NAME #2> <TEST TYPE #2> <CMD TO TEST #2> ...
+#
+# intp_dir    : path to circle-interpreter
+# TEST NAME   : identifier for the test which can be printed if it fails
+# TEST TYPE   : type of each test cases. it should be one of `POS` or `NEG`
+# CMD TO TEST : a command including driver and its arguments 
+#                 (ex. sample.circle sample.input sample.output)
+
```
yunjayh, line 16: Add NOTE about `WORK_DIR` envvar

compiler/circle-interpreter-test/test.lst
```diff
@@ -0,0 +1,26 @@
...
+#
+# NOTE: Test can use enviornment variable `WORK_DIR` for model and data
+#       `WORK_DIR` is set as path to `common-artifacts` directory
+#
+
+# Print help message
+testcmd(POS_001 POS -h)
+
+# Invoke circle-interpreter with proper arguments
+testcmd(POS_002 POS ${WORK_DIR}/Conv2D_000.circle ${WORK_DIR}/Conv2D_000.circle.input /tmp/Conv2D_000.output)
```
jinevening, line 17: output file will be left in `/tmp` directory. It would be better to remove it.

yunjayh: Do you have any great way to do preprocess (which could prepare test materials) and postprocess (such as cleaning by-products)? IMO it can be solved with introducing another macro which removes given files.

jinevening: I don't have "great" idea, but let me share :)

I guess that output file is generated only when (1) circle-interpreter runs with three arguments (2) and returns 0.

So, if you can parse `TESTCMD` here (https://github.com/Samsung/ONE/pull/9658/files#diff-364961265dc34e246ea0cf4c85fdb1e7b6c3a1aafbad633724322cc872d2ad58R45), it will be possible to get the path to the output file. Then run `rm ${OUTPUT_FILE}`.

Something like below

```
if [[ $? -eq 0 ]]; then
  if [[ $TESTTYPE == "POS" ]]; then
    if length of ${TESTCMD} == 3
      OUTPUT_FILE = ${TESTCMD}[2]
      rm ${OUTPUT_FILE}
```

yunjayh: Thank you for your idea! :smile: I'll think more how to solve it.

yunjayh: FYI, I'm considering to adapt gtest to check the output files' existence and it's validation. I'm looking at `circle-operator-test` script, and will try to make code asap. 🙃

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
```diff
@@ -0,0 +1,265 @@
+/*
```
yunjayh, line 1: This code is mostly sourced from `circle-operator-test/src/circle-operator.test.cpp`.

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
```diff
@@ -0,0 +1,265 @@
...
+  {
+    FAIL();
+    return;
+  }
+
+  std::string model = _artifacts_path + "/Conv2D_000.circle";
+  std::string input_prefix = _artifacts_path + "/Conv2D_000.circle.input";
+  std::string output_prefix = "/tmp/Conv2D_000.circle.output";
+  std::string generated_output = output_prefix + "0";
+  std::remove(generated_output.c_str());
```
yunjayh, line 152: > output file will be left in `/tmp` directory. It would be better to remove it.
from https://github.com/Samsung/ONE/pull/9658#discussion_r970435174_

Before the test begins, clean the file name which will be generated. How about this approach? @jinevening

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
```diff
@@ -0,0 +1,263 @@
...
+    FAIL();
+    return;
+  }
+
+  const auto pos = _result.find("Cannot open file");
+  ASSERT_NE(std::string::npos, pos);
+}
+
+// Case of given input data has less number of values than model's input
+TEST_F(circle_interpreter_test, input_data_mismatch_NEG)
```
yunjayh, line 239: I worried how `circle-interpreter` will work if the given input data's number of values mismatch with input tensor size. 

1) If circle model's input size is smaller than the input data's size, execution succeed.
2) If circle model's input size is larger than the input data's size, execution failed with following message.
ERROR: Failed to read data from file "${INPUT_FILE_PATH}".

So, I added this negative test. Is this behavior (succeed for larger size input data) intended? 
/cc @jinevening @seanshpark

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
```diff
@@ -0,0 +1,263 @@
...
+  {
+    std::cerr << "ARTIFACTS_PATH not found" << std::endl;
+    return false;
+  }
+  _artifacts_path = path;
+
+  path = std::getenv("CIRCLE_INTERPRETER_PATH");
+  if (path == nullptr)
+  {
+    std::cerr << "ARTIFACTS_BIN_PATH not found" << std::endl;
```
jinevening, line 52:
```suggestion
    std::cerr << "CIRCLE_INTERPRETER_PATH not found" << std::endl;
```

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
jinevening: Does `std::remove` remove the generated output file?

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
jinevening: > Is this behavior (succeed for larger size input data) intended?

IMHO, it would be better to generate an error (ex: throw an exception) in `circle-interpreter` when the size of input data mismatches with the size of the model input. There may be a way to get the size of input file in cpp code.

So, it would be better to remove this test, because error message may not be the same with the current one.

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
yunjayh: Yes, it removes `generated_output` file if it exists before the test.

IMO, generated file under `/tmp` would not affect to others. (And also, I implemented this way since `circle-operator-test` used same way.)

jinevening (dismissed):
LGTM

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
```diff
@@ -0,0 +1,236 @@
...
+{
+  std::ifstream f1(file1.c_str(), std::ifstream::in | std::ifstream::binary);
+  std::ifstream f2(file2.c_str(), std::ifstream::in | std::ifstream::binary);
+
+  if (!f1.is_open() || !f2.is_open())
+  {
+    return false;
+  }
+
+  char *buffer1 = new char[1024]();
```
seanshpark, line 100: where did 1024 come from?

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
```diff
@@ -0,0 +1,236 @@
...
+  char *buffer2 = new char[1024]();
+
+  do
+  {
+    f1.read(buffer1, 1024);
+    f2.read(buffer2, 1024);
```
seanshpark, line 106: what if read fails?

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+    if (std::memcmp(buffer1, buffer2, 1024) != 0)
```
seanshpark, line 108: buffer may have garbage when allocated. it would be better to revise to (1) clear buffers before reading or (2) compare only for file size.

compiler/circle-interpreter-test/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+`circle-interpreter-test` provides tests of _circle-interpreter_ tool is working as expected.
+
+Current tests includes
+- input arguments test is working as expected
+- output data is written as expected
+- printing help message as well
+- validation of arguments and error message
```
mhs4670go, line 9: (optional)
```suggestion
`circle-interpreter-test` checks if _circle-interpreter_ is working as expected.

Current tests includes
- input arguments test
- output data test
- printing help message test
- validation of arguments and error message test
```

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
```diff
@@ -0,0 +1,236 @@
...
+    return false;
+  }
+  _circle_interpreter_path = path;
+
+  return true;
+}
+
+bool circle_interpreter_test::run(const std::string &command)
+{
+  std::vector<char> buffer(260);
```
mhs4670go, line 62: What does this `260` mean?

yunjayh: Honestly I don't have idea, but I just referred it from `circle-operator.test.cpp`.

seanshpark: Good question. 260 came from Windows platform `_MAX_PATH` but not related with this particular usecase. Just assuming one line should be within this length.

yunjayh: FYI, [linux kernel](https://github.com/torvalds/linux/blob/v5.11/drivers/tty/n_tty.c#L1681) says maximum line length cannot exceed 4095 characters.

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
yunjayh: It's predefined size to read. I'll change it to fixed variable for better reading experience.

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
yunjayh: IMHO, `if (!f1.is_open() || !f2.is_open())` line guarantees that each files are exist. And, `f.read(buffer, size)` will return zero if it cannot read anymore (or an empty file).

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
mhs4670go: Thank you for information. `Note` command or creating a variable named `LINE_MAX_LEN` or something would be helpful.

yunjayh:
Current approach has a problem that `${ARTIFACTS_PATH}/model.circle.input0` is generated from `luci-value-test`. 

I have two questions.

(1) Does anybody know the reason to remain the `*.input` and `*.output` file after `luci-value-test`? Is it used by somewhere after its test? 
(2) `circle-interpreter-test` might test it works or not. IMHO verifying the output's validity wouldn't be needed. How do you think about excluding the comparing file between generated file and expected file? 

Thank you for reviewing :)
/cc @seanshpark @jinevening @mhs4670go

seanshpark:
> Current approach has a problem that ${ARTIFACTS_PATH}/model.circle.input0 is generated from luci-value-test.

I can't catch what the problem is.

> (1) Does anybody know the reason to remain the *.input and *.output file after luci-value-test?

As I know of, to let users/developers look at the artifacts or use for any reason. No particular reason to delete them

> How do you think about excluding the comparing file between generated file and expected file?

If the generated files are to be valid, at least minimum test is needed.
What is the difficult issue for you to do the testing?

yunjayh:
**tl; dr**
First, how about keep this test use input and output data which are generated from `luci-value-test`? 
And, IMHO it would be need to discuss about having dependencies between the tests. 

---

@seanshpark and I had offline discussion and let me recap what I concerned. 

<details>
Currently, the tflite and circle models are generated in `common-artifacts/` after build time. but `*.input*` and `*.output*` files are generated while `luci-value-test`. When I designed `circle-interpreter-test`, I didn't know that input/output files are generated from `luci-value-test`
So now, I am worried about having dependencies between two different tests, e.g. `luci-value-test` and `circle-interpreter-test`. 

I found that a [source](https://www.erlang.org/doc/apps/common_test/dependencies_chapter.html) says **it is strongly recommended to not create dependencies between test cases**.

However, few tests already has dependency with `luci-value-test` to use the input and output data. (`circle-part-value-test` `luci-pass-value-test` ..)

We can (1) keep the dependencies between the tests, or (2) break the dependencies and generate input/output data from tflite model using tflite-interpreter for every tests.

Would it be better to make an issue about this?

</details>

seanshpark:
> Would it be better to make an issue about this?

I think it's up to you :)

> (1) keep the dependencies between the tests vs (2) break the dependencies

this is tricky. `circle-part-value-test` `luci-pass-value-test` have base of `luc-value-test` direct or indirect. -_-;

I think this is also up to you. for (1), we will have more technical-debt but will make you things go faster compare to (2). you can later revise to (2) when you have some time. going to (2) directly will take some time, not sure how much for you.

compiler/circle-interpreter-test/src/circle-interpreter.test.cpp
jinevening: If you're going to update this pattern, better to use `std::vector` like the above code. So you don't have to think about `delete`.

yunjayh:
It has been a long time that this work started. Let's keep this test has dependency with `luci-value-test` but mark as todo to resolve it. :smile:

yunjayh:
Recent changes

- Fix README
- Use buffer for file read
- Use `READSIZE` as 4096 for both output line read and file read block size.
- Fix typo

jinevening (approved):
LGTM

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

PR #9666, opened 2022-09-02
StaticShapeInferer for Bulk

chunseoklee:
Note that Bulk's Static Shape Inferer assumes :
- new input shape has new batch size if it is different from model's
original input shape. Other dimension is not allowed to be changed.
- All new input shapes have the same batch. Thus, it is sufficient to
check 1st input's batch dim.

Signed-off-by: Chunseok Lee <chunseok.lee@samsung.com>

PR #9657, opened 2022-09-01
Batch Execution by passing implicit shapeinference of Bulk op

chunseoklee:

PR #9668, opened 2022-09-02
Bulk layer's batch execution

chunseoklee:
Implement Bulk's batch execution. It supports parallel execution with 2
trix cores.

Signed-off-by: Chunseok Lee <chunseok.lee@samsung.com>

### DIFF
--- a/runtime/onert/backend/trix/ops/BulkLayer.cc
+++ b/runtime/onert/backend/trix/ops/BulkLayer.cc
@@ -117,30 +117,40 @@ void BulkLayer::run()
   _dev_context->setDataInfo<const IPortableTensor>(&in_info, _inputs);
   _dev_context->setDataInfo<IPortableTensor>(&out_info, _outputs);
 
-  input_buffers input_buf[2];
-  output_buffers output_buf[2];
+  std::vector<input_buffers> input_buf;
+  std::vector<output_buffers> output_buf;
+  input_buf.resize(_dev_context->getDevSize());
+  output_buf.resize(_dev_context->getDevSize());
 
+  std::vector<std::future<void>> f(_dev_context->getDevSize());
+
+  const int num_cores = _dev_context->getDevSize();
   if (is_batch_execution)
   {
     // TODO: Support for general number of cores(>2)
     // Here we assume that 2 trix cores
-    for (int i = 0; i < (batch_size); i = i + 2)
+    for (int i = 0; i < (batch_size); i = i + num_cores)
     {
-      _dev_context->setBuffer<const IPortableTensor>(&input_buf[0], _inputs, batch_size, i);
-      _dev_context->setBuffer<IPortableTensor>(&output_buf[0], _outputs, batch_size, i);
-
-      _dev_context->setBuffer<const IPortableTensor>(&input_buf[1], _inputs, batch_size, i + 1);
-      _dev_context->setBuffer<IPortableTensor>(&output_buf[1], _outputs, batch_size, i + 1);
-
-      auto f0 = std::async(std::launch::async, &single_job, _dev_context->getDev(0), req_id[0],
-                           &input_buf[0], &in_info, &output_buf[0], &out_info);
-      if (i + 1 < batch_size) // ignore last job if batch_size is odd number
+      for (int core = 0; core < num_cores; core++)
       {
-        auto f1 = std::async(std::launch::async, &single_job, _dev_context->getDev(1), req_id[1],
-                             &input_buf[1], &in_info, &output_buf[1], &out_info);
-        f1.wait();
+        _dev_context->setBuffer<const IPortableTensor>(&input_buf[core], _inputs, batch_size,
+                                                       i + core);
+        _dev_context->setBuffer<IPortableTensor>(&output_buf[core], _outputs, batch_size, i + core);
       }
-      f0.wait();
+      for (int core = 0; core < num_cores; core++)
+      {
+
+        if (i + core < batch_size)
+        {
+          f[core] =
+            std::async(std::launch::async, &single_job, _dev_context->getDev(core), req_id[core],
+                       &input_buf[core], &in_info, &output_buf[core], &out_info);
+        }
+      }
+      for (int core = 0; core < num_cores; core++)
+      {
+        f[core].wait();
+      }
     }
   }
   else
--- a/runtime/onert/backend/trix/ops/BulkLayer.cc
+++ b/runtime/onert/backend/trix/ops/BulkLayer.cc
@@ -18,6 +18,7 @@
 #include <util/logging.h>
 
 #include <libnpuhost.h>
+#include <future>
 
 namespace onert
 {
@@ -49,24 +50,57 @@ void BulkLayer::configure(const std::vector<const IPortableTensor *> &inputs,
     throw std::runtime_error("Unable to extract the model metadata");
   }
 
+  _model_id.resize(_dev_context->getDevSize());
+
   generic_buffer model_file;
   model_file.type = BUFFER_FILE;
   model_file.filepath = binary_path.c_str();
   model_file.size = _meta->size;
 
-  if (registerNPUmodel(dev_context->getDev(), &model_file, &_model_id) < 0)
+  if (registerNPUmodel(dev_context->getDev(0), &model_file, &_model_id[0]) < 0)
+  {
+    throw std::runtime_error("Failed to register npu model");
+  }
+  if (registerNPUmodel(dev_context->getDev(1), &model_file, &_model_id[1]) < 0)
   {
     throw std::runtime_error("Failed to register npu model");
   }
 }
 
+void single_job(npudev_h dev, int req_id, input_buffers *input_buf, tensors_data_info *in_info,
+                output_buffers *output_buf, tensors_data_info *out_info)
+{
+  if (setNPU_requestData(dev, req_id, input_buf, in_info, output_buf, out_info))
+  {
+    throw std::runtime_error("Unable to create NPU request for red_id (" + std::to_string(req_id) +
+                             ")");
+  }
+
+  if (submitNPU_request(dev, req_id))
+  {
+    throw std::runtime_error("Unable to submit NPU request with req id (" + std::to_string(req_id) +
+                             ")");
+  }
+}
+
 void BulkLayer::run()
 {
-  int req_id;
-  if (createNPU_request(_dev_context->getDev(), _model_id, &req_id))
+  // TODO: Remove too many assumption
+  // We assume user wants batch execurion if user's input size is larger than model's input size
+  int user_input_batch = (_inputs[0]->get_info().shape()).dim(0);
+  int model_input_batch = _meta->input_seg_dims[0][0];
+  int batch_size = user_input_batch / model_input_batch;
+  bool is_batch_execution = (batch_size != 1 ? true : false);
+
+  std::vector<int> req_id(_dev_context->getDevSize());
+
+  for (int i = 0; i < _dev_context->getDevSize(); i++)
   {
-    throw std::runtime_error("Unable to create NPU request with model id (" +
-                             std::to_string(_model_id) + ")");
+    if (createNPU_request(_dev_context->getDev(i), _model_id[i], &req_id[i]))
+    {
+      throw std::runtime_error("Unable to create NPU request with model id (" +
+                               std::to_string(_model_id[i]) + ")");
+    }
   }
 
   if (_meta->input_seg_num != _inputs.size())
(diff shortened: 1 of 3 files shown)


chunseoklee:
This is a last commit of #9657. 

wait for #9666 landed.

runtime/onert/backend/trix/ops/BulkLayer.cc
```diff
@@ -49,24 +50,57 @@ void BulkLayer::configure(const std::vector<const IPortableTensor *> &inputs,
...
   model_file.type = BUFFER_FILE;
   model_file.filepath = binary_path.c_str();
   model_file.size = _meta->size;
 
-  if (registerNPUmodel(dev_context->getDev(), &model_file, &_model_id) < 0)
+  if (registerNPUmodel(dev_context->getDev(0), &model_file, &_model_id[0]) < 0)
+  {
+    throw std::runtime_error("Failed to register npu model");
+  }
+  if (registerNPUmodel(dev_context->getDev(1), &model_file, &_model_id[1]) < 0)
```
glistening, line 64: It seems to assume device number is 2. It would be better to use iterate with number_of_devices. As I understand, it is `2` for both x86 and arm. It is also preferred to introduce constant instead of magic number `2`.

glistening (commented):
@chunseoklee 
- @jyoungyun may be more proper person to review. However, I started to review for her absence.
- I did not finish my review, but I leave some comments.

runtime/onert/backend/trix/ops/BulkLayer.cc
```diff
@@ -49,24 +50,57 @@ void BulkLayer::configure(const std::vector<const IPortableTensor *> &inputs,
...
+                             ")");
+  }
+}
+
 void BulkLayer::run()
 {
-  int req_id;
-  if (createNPU_request(_dev_context->getDev(), _model_id, &req_id))
+  // TODO: Remove too many assumption
+  // We assume user wants batch execurion if user's input size is larger than model's input size
```
glistening, line 89:
```suggestion
  // We assume user wants batch execution if user's input size is larger than model's input size
```

Personally, I would like to use stricter condition for assuming batch execution.
- Not just user's input size > model's input size
- But user's input size is multiples of model's input size

runtime/onert/backend/trix/ops/BulkLayer.cc
```diff
@@ -49,24 +50,57 @@ void BulkLayer::configure(const std::vector<const IPortableTensor *> &inputs,
...
+  int user_input_batch = (_inputs[0]->get_info().shape()).dim(0);
+  int model_input_batch = _meta->input_seg_dims[0][0];
+  int batch_size = user_input_batch / model_input_batch;
+  bool is_batch_execution = (batch_size != 1 ? true : false);
```
glistening, line 93: @chunseoklee Can `is_batch_execution` be determined in `configure` stage, not in `run`? I think you're thinking dynamic shape. But I am not sure multiple batch would work for dynamic shape. For example, AFAIK, trix is not ready for dynamic shape.

runtime/onert/backend/trix/ops/BulkLayer.cc
```diff
@@ -84,28 +118,48 @@ void BulkLayer::run()
   _dev_context->setDataInfo<const IPortableTensor>(&in_info, _inputs);
   _dev_context->setDataInfo<IPortableTensor>(&out_info, _outputs);
 
-  input_buffers input_buf;
-  output_buffers output_buf;
-  _dev_context->setBuffer<const IPortableTensor>(&input_buf, _inputs);
-  _dev_context->setBuffer<IPortableTensor>(&output_buf, _outputs);
+  input_buffers input_buf[2];
+  output_buffers output_buf[2];
```
glistening, line 122: I would like to see some name instead of magic number `2` here.

runtime/onert/backend/trix/ops/BulkLayer.cc
```diff
@@ -84,28 +118,48 @@ void BulkLayer::run()
...
-
-  if (submitNPU_request(_dev_context->getDev(), req_id))
+  else
   {
-    throw std::runtime_error("Unable to submit NPU request with req id (" + std::to_string(req_id) +
-                             ")");
+    _dev_context->setBuffer<const IPortableTensor>(&input_buf[0], _inputs, batch_size, 0);
+    _dev_context->setBuffer<IPortableTensor>(&output_buf[0], _outputs, batch_size, 0);
+
+    single_job(_dev_context->getDev(0), req_id[0], &input_buf[0], &in_info, &output_buf[0],
```
glistening, line 152: @chunseoklee `single_job` give me the impression it is going to be run by `thread` or `std::async`. Do we need a thread or `future` for non-batch execution?

runtime/onert/backend/trix/ops/BulkLayer.cc
```diff
@@ -49,24 +50,57 @@ void BulkLayer::configure(const std::vector<const IPortableTensor *> &inputs,
...
+  int batch_size = user_input_batch / model_input_batch;
+  bool is_batch_execution = (batch_size != 1 ? true : false);
+
+  std::vector<int> req_id(_dev_context->getDevSize());
+
+  for (int i = 0; i < _dev_context->getDevSize(); i++)
   {
-    throw std::runtime_error("Unable to create NPU request with model id (" +
-                             std::to_string(_model_id) + ")");
+    if (createNPU_request(_dev_context->getDev(i), _model_id[i], &req_id[i]))
```
glistening, line 99: @chunseoklee I think `getDevSize()` is always 2. Thus, for a single batch, it always call `createNPU_request` twice (= getDevSize()). It means even for a single batch, there are always two calls. Is it right?

runtime/onert/backend/trix/ops/BulkLayer.cc
chunseoklee: This calculation works for both static and dynamic shape cases even though we do not have dynamic shape inference for Bulk.

runtime/onert/backend/trix/ops/BulkLayer.cc
chunseoklee: Yes. it is always called twice. Thus, _dev_context->getDev(2) never used in the case of non-batch execution.

runtime/onert/backend/trix/ops/BulkLayer.cc
glistening: I also found it is just a typical function, not for thread start func and async after finishing to read your code. Now I am okay,

runtime/onert/backend/trix/ops/BulkLayer.cc
glistening: Then, I prefer to have different path for single batch and multiple batch ealier though I don't know how much overhead or impaction of `createNPU_request`. But I am okay for this code.

runtime/onert/backend/trix/ops/BulkLayer.cc
glistening: Then, I prefer to move this code in `configure` and use `is_batch_execution` to have single batch and multiple match different (its own way) for readability and removing redundant code. But it is just my preference. I am okay for current code.

glistening:
I added @jyoungyun as optional reviewer to give her a notification. It would be helpful if we can get her review.

runtime/onert/backend/trix/ops/BulkLayer.cc
```diff
@@ -84,28 +117,48 @@ void BulkLayer::run()
...
+      auto f0 = std::async(std::launch::async, &single_job, _dev_context->getDev(0), req_id[0],
+                           &input_buf[0], &in_info, &output_buf[0], &out_info);
+      if (i + 1 < batch_size) // ignore last job if batch_size is odd number
+      {
+        auto f1 = std::async(std::launch::async, &single_job, _dev_context->getDev(1), req_id[1],
+                             &input_buf[1], &in_info, &output_buf[1], &out_info);
+        f1.wait();
+      }
+      f0.wait();
+    }
```
jyoungyun, line 144: (Q) Does the `std::async` function replace what `single_job` function does?

glistening: I don't 100% understand the meaning of _replace_. The main thread uses `std::async` to run `single_job` asynchronously. Internally, `std::async` will create worker thread and run `single_job` though its working depends on its implementation.

chunseoklee:
confired that https://github.com/Samsung/ONE/pull/9668/commits/d922cc421dc5b4b6131f476ffa34cae0f4d8d827 works fine on real device:

```
tv@tv ~/chunseoklee> sdb2 shell BACKENDS=trix /usr/bin/nnfw-test/Product/out/bin/nnpackage_run /opt/trix_batch/mv1.q8 -r 3 --shape_prepare="[0,[9,224,224,3]]"
Package Filename /opt/trix_batch/mv1.q8
MODEL_LOAD   takes 1.520 ms
PREPARE      takes 84.548 ms
EXECUTE      takes 47.910 ms
- MEAN     :  47.910 ms
- MAX      :  64.654 ms
- MIN      :  38.423 ms
- GEOMEAN  :  46.568 ms
]0;fish  /home/tv/chunseoklee⏎                                                                                                                                                                                     tv@tv ~/chunseoklee> sdb2 shell BACKENDS=trix /usr/bin/nnfw-test/Product/out/bin/nnpackage_run /opt/trix_batch/mv1.q8 -r 3 --shape_prepare="[0,[12,224,224,3]]"
Package Filename /opt/trix_batch/mv1.q8
MODEL_LOAD   takes 2.376 ms
PREPARE      takes 68.032 ms
EXECUTE      takes 55.078 ms
- MEAN     :  55.078 ms
- MAX      :  82.703 ms
- MIN      :  40.954 ms
- GEOMEAN  :  52.026 ms
]0;fish  /home/tv/chunseoklee⏎                                                                                                                                                                                     tv@tv ~/chunseoklee> sdb2 shell BACKENDS=trix /usr/bin/nnfw-test/Product/out/bin/nnpackage_run /opt/trix_batch/mv1.q8 -r 3 
Package Filename /opt/trix_batch/mv1.q8
MODEL_LOAD   takes 1.829 ms
PREPARE      takes 73.171 ms
EXECUTE      takes 5.117 ms
- MEAN     :  5.117 ms
- MAX      :  6.836 ms
- MIN      :  3.587 ms
- GEOMEAN  :  4.944 ms
```

runtime/onert/backend/trix/ops/BulkLayer.cc
chunseoklee: I remove the usage of magic number 2

runtime/onert/backend/trix/ops/BulkLayer.cc
glistening: (optional) `is_batch_execution` determined by both `batch_size` and `_dev_context->getDevSize()`. Also, it can be determined in `configure`, not each `run` since we don't have plan to support dynamic shape yet.

runtime/onert/backend/trix/ops/BulkLayer.cc
```diff
@@ -84,28 +117,58 @@ void BulkLayer::run()
...
   _dev_context->setDataInfo<IPortableTensor>(&out_info, _outputs);
 
-  input_buffers input_buf;
-  output_buffers output_buf;
-  _dev_context->setBuffer<const IPortableTensor>(&input_buf, _inputs);
-  _dev_context->setBuffer<IPortableTensor>(&output_buf, _outputs);
+  std::vector<input_buffers> input_buf;
+  std::vector<output_buffers> output_buf;
+  input_buf.resize(_dev_context->getDevSize());
+  output_buf.resize(_dev_context->getDevSize());
```
glistening, line 123: (optional) I prefer to call `resize` only for batch execution. Heap allocation has overhead and may cause defragmentation.

runtime/onert/backend/trix/ops/BulkLayer.h
```diff
@@ -50,7 +50,7 @@ class BulkLayer : public ::onert::exec::IFunction
   std::vector<const IPortableTensor *> _inputs;
   std::vector<IPortableTensor *> _outputs;
 
-  uint32_t _model_id;
+  std::vector<uint32_t> _model_id;
```
glistening, line 53: Q. Do we need to make multiple `_model_id` for identical models? Is it now allowed to use same `_model_id` for same model (`.tvn`)?

glistening (approved):
I would like to update this code as I suggested, but handle this later, not in this PR.

runtime/onert/backend/trix/ops/BulkLayer.h
chunseoklee: AFAIK, each core creates unique model_id for each request.

glistening: No. It seems to make unique `req_id` for each request (via `createNPU_request`). `model_id` is for each model (via `registerNPUmodel`).

jyoungyun (approved):
LGTM

runtime/onert/backend/trix/ops/BulkLayer.h
glistening: @jyoungyun Do you have any information about https://github.com/Samsung/ONE/pull/9668#discussion_r963219378?

PR #9678, opened 2022-09-06
[circle-quantizer] Remove unused argparse

glistening:
argparse is not used anywhere and misleading.
It removes argparse and related comment.


mhs4670go (approved):
LGTM

jinevening (approved):
LGTM Thank you!

(TMI) I was curious why it was not detected as an unused variable during compilation. I found that g++ only detects unused variables for built-in types.

PR #9695, opened 2022-09-08
[infra/runtime] Clean up makefile

hseok-oh:
This commit cleans up "Makefile.template" used on runtime CI infra.
- Change default target: build -> install
- Remove unused environment variable setting on CI infra
- Rename target for test suite and packaging
- Hide coverage build setting
  - Set coverage build when target is coverage test suite
- Remove stamp file for build step

### DIFF
--- a/Makefile.template
+++ b/Makefile.template
@@ -5,7 +5,6 @@ CROSS_BUILD?=0
 HOST_OS?=linux
 TARGET_OS?=linux
 COVERAGE_BUILD?=0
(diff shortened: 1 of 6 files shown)


glistening:
@hseok-oh I like this change. But I would like to know what is the changed behavior after removing timestamp. As I understand, we've used them for reduce duplicated configure, download, and so on.

glistening (approved):
LGTM

chunseoklee (approved):

PR #9707, opened 2022-09-15
DRAFT tidy one-cmds tests

seanshpark:
on-going draft to tidy one-cmds tests.


seanshpark:
Found out that there are unnecessary preparation of `inception_v3.circle`, wrong name for log deletion and others.

seanshpark:
done

PR #9709, opened 2022-09-15
[one-cmds-test] Fix remove log file

seanshpark:
This will fix remove log file as
- add remove log file where not done
- use rm -f without r option
- correct log filename


seanshpark:
from draft #9707

jinevening (approved):
LGTM

Issue #9712, opened 2022-09-15
[ONE-Dockerizing] Introduce ONE Dockerizing


holawan:
### What

Let’s introduce **`ONE Dockerizing`**.

### Why

Currently, `onecc` is only supported on Ubuntu 18.04 and 20.04 (not offially).

Therefore, it is not easy for people to use `onecc` in other environments.

Our `ONE-Dockerizing` will use `Docker` to make `onecc` easy for users to use in other environments.
The ultimate goal will be to use `onecc` with Docker on Windows, Mac, etc., but we are currently working on Linux, and our goal is to do `ONE-Dockerizing` on all Linux environments.

#### System Scenario
1. Receive commands entered by the user.
2. Verify that `onecc-docker` can be run in the user's environment (existence of docker, installation of python3, etc.)
3. Run Docker Container
    - When Docker Image is not present
        - Create the docker image
        - Run Docker Container
    - When Docker Image is present
    - Run the docker container by specifying the volume as '/home:/home' and by specifying the location where the current user is running as WORKDIR.
        - `docker run --rm --name {container_name} -v /home:/home --workdir {user_pwd} {docker_image} {user_command}`
4. Create `circle files` and `nnnpackage` using the `.cfg` and TensorFlow models shared as volumes inside the docker container and provide them to the user.

#### User Scenario

- Create a `*.cfg` and TensorFlow Model file(`.pb`, `.tfilte` ..)  containing the option you want to perform related to the "onecc"

- The user runs the `onecc-docker`. (including the flag desired by the user)
  - `onecc-docker -C *.cfg `

- The user will get the desired result. Below is an example.
  - version
  - circlefile,nnpackages
  - help message

### TODO

We will first run Docker-related tasks to run "onecc-docker". Because this requires a Dockerfile to create a Docker Image, we will create it. 
Second, we're going to write an "onecc-docker" shell script and run Docker on the system. 
Finally, we would like to provide Debian Package with "onecc-docker".

I have summarized and will update what I need for this work below.

- [x] Create Dockerfile
    - [ ] Make Docker container run successful with commands entered by the user
        - [x] `-I,` `-C`, `-O`,..
- [x] Create onec-docker shell script
    - [ ] Check for `docker` installation script
    - [ ] Check for `python3` installation script
    - [ ] Check for `onecc` installation script
    - [x] run docker if docker image exists, and create a script to build if not.
- [x] Create `onecc-docker` Debian Package

related issue : #8232

[...]

Issue #8232, opened 2021-12-29
[one-cmds] Introduce `onecc-docker` binary and `one-compiler-docker` package


jyoungyun:
### What

Let's consider how to use `one-cmds` tools easily for more people.

For now, `one-cmds` tools only support Ubuntu 18.04 and 20.04(not officially).
So people in other environments can't use our tools unless they upgrade the OS (or install Ubuntu OS).

IMO, in this case, docker can be a good solution.

[Docker](https://www.docker.com/) is an open source containerization platform. It supports developer to provide applications into containers. And docker is already available on a variety of Linux platforms, macOS and Windows 10 through Docker Desktop. So if we provide `one-compiler` tool using docker, we can support many types of environment at once.

### How

**Dockerfile**
   - Based on Ubuntu 18.04 or 20.04
   - Install `one-compiler`
   - Set `onecc` as entrypoint

This dockerfile generates a docker image for `one-compiler` on Ubuntu 18.04 or 20.04.
User can run a process in isolated container using this docker image with several `onecc` parameters.

**`onecc-docker` tool**
   - Run docker container with `onecc` parameters and several docker run params

This `onecc-docker` tool runs docker container using above docker image. It sets user and group ids the same as those on host PC, mounts `/tmp` and `${HOME}` directory, adds `HOME` environment, and changes working directory to current directory.

**`one-compiler-docker` package**
   - files
      - `Dockerfile`
      - `onecc-docker` files
   - install
      - `Dockerfile` to `/usr/share/one/docker/`
      - `onecc-docker` to `/usr/share/one/bin/`
   - postinst
      - Install symbolic link from `/usr/share/one/bin/onecc-docker` to `/usr/bin/onecc` with priority 0 using [update-alternatives](https://linux.die.net/man/8/update-alternatives) tool
   - prerm
      - Remove symbolic link

**`one-compiler` package**
   - postinst
      - Install symbolic link from `/usr/share/one/bin/onecc` to `/usr/bin/onecc` with priority 1 using [update-alternatives](https://linux.die.net/man/8/update-alternatives) tool
        (1 is higher priority than 0)
   - prerm
      - Remove symbolic link

### Todo
- [ ] dockerfile apt-source and version management

mhs4670go:
I have a few question. 
1. It seems that users should build a docker image with `Dockerfile` when they start from scratch. Is it impossible to let users to pull the pre-built image from docker hub or somewhere?
[...]

PR #9714, opened 2022-09-15
[Draft][ONE-Dockerizing] Introduce onecc dockerizing

Seunghui98:
This commit introduces Dockerizing for onecc.
Regarding [#8232](https://github.com/Samsung/ONE/issues/8232) issue, we are going to do Dockerizing so that more users can easily use onecc.
I wrote onecc version selection code part for dockerizing with python and bash shell script. in the draft code for now.
However, if DraftCode is checked, we plan to first post an implementation code in PR with the latest version of onecc dockerizing as the default operation.

Related Issue : [#9712](https://github.com/Samsung/ONE/issues/9712)

## Work Flow
- If the selected version of onecc is pre-installed, the input onecc-cmd is executed. 
- And if the selected onecc version is not installed, check if there is a docker image of the selected version. 
   - If the selected version of onecc docker image is not present, the selected onecc docker image is built. 
   - The selected version of the onecc docker container is then run.

## How to test

I currently proceeded with Dockerizing in the following way.

### summary

1. Import TensorFlow Model for Dockerizing and write Configure File.
2. Write a Dockerfile to create a Docker image and run the container.
    - When `onecc` is finished running, the container is terminated and deleted.
3. Check that circle file and nnpackage have been created.

### Import Tensorflow model for testing

First, I created the path '/home/test/' for Linux (wsl) and downloaded the model file to test. After downloading [inception_v3](https://storage.googleapis.com/download.tensorflow.org/models/tflite/model_zoo/upload_20180427/inception_v3_2018_04_27.tgz), the Configure file was created.

```
# /home/test/
/home/test$ wget https://storage.googleapis.com/download.tensorflow.org/models/tflite/model_zoo/upload_20180427/inception_v3_2018_04_27.tgz

/home/test$ tar -xvf inception_v3_2018_04_27.tgz

/home/test$ vi onecc.template.cfg

/home/test$ tree
.
├── inception_v3.pb
├── inception_v3.tflite
├── inception_v3_2018_04_27.tgz
├── labels.txt
└── onecc.template.cfg
```

- onecc.template.cfg

    ```
    [onecc]
    one-import-tf=True
    one-import-tflite=False
    one-import-bcq=False
    one-import-onnx=False
    one-optimize=True
    one-quantize=False
    one-parition=False
    one-pack=True
    one-codegen=False
    
    [one-import-tf]
    input_path=inception_v3.pb
    input_arrays=input
    input_shapes=1,299,299,3
    output_arrays=InceptionV3/Predictions/Reshape_1
    converter_version=v1
    model_format=graph_def
    
    [one-optimize]
    input_path=inception_v3.circle
    output_path=inception_v3.opt.circle
    generate_profile_data=False
    
    [one-pack]
    input_path=inception_v3.opt.circle
    output_path=inception_v3_pack
    ```

### run onecc-docker.sh
```
$ cd ONE/compiler/onecc-docker/
$ bash ./onecc-docker.sh
``` 

   
### Execution Results 

This is my execution result. The circle file and nnpackage have been created well, and I don't think there is any problem with my method so far. 

Docker container was also successfully removed. 

- Test results
```
/home/test$ tree
.
├── inception_v3.circle
├── inception_v3.circle.log
├── inception_v3.opt.circle
├── inception_v3.opt.circle.log
├── inception_v3.pb
├── inception_v3.tflite
├── inception_v3_2018_04_27.tgz
├── inception_v3_pack
│   └── inception_v3.opt
│       ├── inception_v3.opt.circle
│       └── metadata
│           └── MANIFEST
├── inception_v3_pack.log
├── labels.txt
└── onecc.template.cfg

3 directories, 12 files
```
/cc [@Samsung/ootpg](https://github.com/orgs/Samsung/teams/ootpg)
Signed-off-by: seunghui-lee <hithere1012@naver.com>

### DIFF
--- a/compiler/onecc-docker/Dockerfile
+++ b/compiler/onecc-docker/Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:18.04 
+
+RUN apt-get update 
+
+RUN apt-get install -y wget 
+
+RUN wget https://github.com/Samsung/ONE/releases/download/1.21.0/one-compiler_1.21.0_amd64.deb 
+
+RUN apt install -y ./one-compiler_1.21.0_amd64.deb 
+
+WORKDIR /home/ 
+
+RUN mkdir test 
+
+WORKDIR /home/test 
+
+CMD ["onecc", "-C", "onecc.template.cfg"]
--- a/compiler/onecc-docker/onecc-docker.py
+++ b/compiler/onecc-docker/onecc-docker.py
@@ -1,0 +1,64 @@
+import sys
+import subprocess
+from distutils.version import StrictVersion
+import json
+
+def _is_image(version) : 
+    cmd = ['docker', 'images', f'onecc:{version}']
+    p = subprocess.Popen(cmd,  stdout=subprocess.PIPE, stderr=subprocess.STDOUT) 
+    out, err = p.communicate()
+    dic = {}
+
+    decoded_out = [i.split() for i in out.decode().split("\n") if len(i)]
+
+    if len(decoded_out) <2 :
+        return False
+
+    dic[decoded_out[0][0]] = decoded_out[1][0]
+    dic[decoded_out[0][1]] = decoded_out[1][1]
+
+    tag_name = cmd[2].split(':')
+
+    if dic['REPOSITORY'] == tag_name[0] and dic['TAG'] == tag_name[1]  :
+        return True 
+    else :
+        return False
+
+
+def main():
+    installed_onecc_version = sys.argv[1]
+    versions_str = subprocess.check_output("curl -s https://api.github.com/repos/Samsung/ONE/tags", shell=True ,encoding='utf-8')
+    versions_json = json.loads(versions_str)
+    versions = []
+    for i in versions_json:
+        versions.append(i["name"])
+
+    versions.sort(key=StrictVersion)
+    chosen_version = ""
+    while 1:
+        for i in range(18, len(versions)):
+            print(versions[i])
+        chosen_version = input("choose version: ")
+        flag = False
+        for i in range(18, len(versions)):
+            if chosen_version == versions[i]:
+                flag = True
+                break
+        if(flag):
+            break
+        else:
+            print("choose right version")
+    if chosen_version == installed_onecc_version:
+        print("onecc is installed in computer")
+        onecc_cmd = ["onecc"] + sys.argv[2:]
+        subprocess.call(onecc_cmd)
+        exit(0)
+
+    if not _is_image(chosen_version):
+        build_cmd = ["docker", "build", "-t", f"onecc:{chosen_version}", "."]
+        subprocess.call(build_cmd)
+    run_cmd = ["docker", "run", "--rm", "--name", "onecc", "-v", "/home/:/home/", f"onecc:{chosen_version}"]
+    subprocess.call(run_cmd)
+
+if __name__ == "__main__":
+	main()
--- a/compiler/onecc-docker/onecc-docker.sh
+++ b/compiler/onecc-docker/onecc-docker.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+if ! command -v docker > /dev/null
+        then
+        echo "docker must be installed"
+        exit 0
+fi
+
+if ! command -v python3 > /dev/null
+        then
+        echo "python3 must be installed"
+        exit 0
+fi
+
+if command -v onecc > /dev/null 
+        then
+        version=( $(onecc --version) )
+        exec "python3" "onecc-docker.py" "${version[2]}" "$@" 
+        exit 0
+else
+        exec "python3" "onecc-docker.py" "0" "$@" 
+        exit 0
+fi
+


jyoungyun:
There is nothing in the commit message of this PR except the header. You should put more contents in the commit message. :)
https://github.com/Samsung/ONE/pull/9714/commits

And if you have a lot to say about PR as it is now, it would be better to use the issue.
The content of the PR is not re-written so long except for the content in the commit itself. :)

compiler/onecc-docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:18.04 
+
+RUN apt-get update
```
jyoungyun, line 3: When writing a `Dockerfile`, we have to minimize the number of layers. The `RUN`, `COPY`, `ADD` create layers. If you execute only one command at one layer, each execution result is saved as the cache. In this case, if you execute `apt-get update`, all updated files will create a cache. Rather than implementing it like this, it is more efficient to implement it by executing `apt-get update` on one `RUN` command, installing the package you actually want to install, and removing files that have been used unnecessarily.

related guideline : https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

compiler/onecc-docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:18.04 
+
+RUN apt-get update 
+
+RUN apt-get install -y wget 
+
+RUN wget https://github.com/Samsung/ONE/releases/download/1.21.0/one-compiler_1.21.0_amd64.deb 
+
+RUN apt install -y ./one-compiler_1.21.0_amd64.deb
```
jyoungyun, line 9: I think `apt-get` would be better to use in bash script file. Because the `apt` is designed to Linux user and does not have a stable CLI interface.

compiler/onecc-docker/onecc-docker.py
```diff
@@ -0,0 +1,64 @@
...
+        exit(0)
+
+    if not _is_image(chosen_version):
+        build_cmd = ["docker", "build", "-t", f"onecc:{chosen_version}", "."]
+        subprocess.call(build_cmd)
+    run_cmd = ["docker", "run", "--rm", "--name", "onecc", "-v", "/home/:/home/", f"onecc:{chosen_version}"]
+    subprocess.call(run_cmd)
+
+if __name__ == "__main__":
+	main()
```
jyoungyun, line 64: Need new line :)

compiler/onecc-docker/onecc-docker.sh
```diff
@@ -0,0 +1,24 @@
...
+if ! command -v python3 > /dev/null
+        then
+        echo "python3 must be installed"
+        exit 0
+fi
+
+if command -v onecc > /dev/null 
+        then
+        version=( $(onecc --version) )
+        exec "python3" "onecc-docker.py" "${version[2]}" "$@"
```
jyoungyun, line 18: (Q) Isn't it possible to receive the `onecc` version in `onecc-docker.py` file?

morumoru: I thought that onecc --version command should be checked when onecc is installed in ubuntu system
Do you mean that just excute python3 onecc-docker.py $@ command and check installation of onecc in onecc-docker.py?

compiler/onecc-docker/Dockerfile
holawan: @jyoungyun
Do you mean to use the apt-get option rather than the wget option? Or do you mean to change the apt of the Debian package installation part?

I wonder if this is in line with your intentions
```
FROM ubuntu:18.04 

RUN apt-get update && apt-get install -y wget

RUN wget https://github.com/Samsung/ONE/releases/download/1.21.0/one-compiler_1.21.0_amd64.deb 
RUN apt-get install -y ./one-compiler_1.21.0_amd64.deb 
```

compiler/onecc-docker/onecc-docker.sh
jyoungyun: Yes. I think that you can get the status whether the `onecc` is installed in `onecc-docker.py` file.

compiler/onecc-docker/Dockerfile
jyoungyun: AFAIK, `apt` is not appropriate to use in bash script. :) I suggest to use `apt-get` instead of `apt`.

compiler/onecc-docker/onecc-docker.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
mhs4670go, line 1: Why do we need this bash script? Is it for checking if `docker` and `python` exist? We can skip them if you are going to create debian package by having a dependencies. Then, debian tools automatically checks what you want to do.

Please refer to `onecc` codes. (https://github.com/Samsung/ONE/blob/330f9a37cc520c4a5afc91858a8cc2276bb7559a/compiler/one-cmds/onecc#L1)

compiler/onecc-docker/onecc-docker.py
```diff
@@ -0,0 +1,64 @@
+import sys
```
mhs4670go, line 1: Format checking before commiting the codes would make an easier review.

You can check it by a command `./nnas format` or `./nnas format --diff-only`.

compiler/onecc-docker/onecc-docker.sh
morumoru: Oh I read that onecc used shell script on top of python code so it can call itself by python
I'll combine shell script and python code later!

I didn't know well about debian package so I didn't consider using dependencies to check python and docker
I'll fix my code according to your advice! Thank you!

compiler/onecc-docker/onecc-docker.py
Seunghui98: Oh, thank you for the code review. I'm not sure what you mean by part `./nnas format` or part `./nnas format --diff-only`. Is there anything I can refer to?

mhs4670go: I mean `/ONE/nnas`:)

compiler/onecc-docker/onecc-docker.sh
jyoungyun: But we don't use the virtual environment in python script. So we don't need to set the `PY_PATH` I think.

compiler/onecc-docker/onecc-docker.py
```diff
@@ -0,0 +1,64 @@
...
+
+    if dic['REPOSITORY'] == tag_name[0] and dic['TAG'] == tag_name[1]  :
+        return True 
+    else :
+        return False
+
+
+def main():
+    installed_onecc_version = sys.argv[1]
+    versions_str = subprocess.check_output("curl -s https://api.github.com/repos/Samsung/ONE/tags", shell=True ,encoding='utf-8')
```
mhs4670go, line 30: How about use python library? We don't have to run bash command here.
```suggestion
import requests

response = requests.get("https://api.github.com/repos/Samsung/ONE/tags")
```

morumoru: I'll use requests instead of subprocess! Thank you

mhs4670go: Please consider the case when the response has a problem. I'm not sure but you can refer to `response.status`, `response.reason` information.

compiler/onecc-docker/onecc-docker.py
```diff
@@ -0,0 +1,64 @@
...
+    versions = []
+    for i in versions_json:
+        versions.append(i["name"])
+
+    versions.sort(key=StrictVersion)
+    chosen_version = ""
+    while 1:
+        for i in range(18, len(versions)):
+            print(versions[i])
+        chosen_version = input("choose version: ")
```
mhs4670go, line 41: I think we shouldn't force users to select the version every time. It means that a default version will do the trick. If users don't specify the version, we can assume they want to run `one-compiler` with latest version.

compiler/onecc-docker/onecc-docker.py
```diff
@@ -0,0 +1,64 @@
...
+    versions_str = subprocess.check_output("curl -s https://api.github.com/repos/Samsung/ONE/tags", shell=True ,encoding='utf-8')
+    versions_json = json.loads(versions_str)
+    versions = []
+    for i in versions_json:
+        versions.append(i["name"])
+
+    versions.sort(key=StrictVersion)
+    chosen_version = ""
+    while 1:
+        for i in range(18, len(versions)):
```
mhs4670go, line 39: Where is this `18` coming from? This is what people called "magic number". For better readability, you should create a variable for them.

compiler/onecc-docker/onecc-docker.py
```diff
@@ -0,0 +1,64 @@
+import sys
+import subprocess
+from distutils.version import StrictVersion
+import json
+
+def _is_image(version) : 
+    cmd = ['docker', 'images', f'onecc:{version}']
+    p = subprocess.Popen(cmd,  stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
mhs4670go, line 8: You can use the function `_run` from `utils` library.
https://github.com/Samsung/ONE/blob/330f9a37cc520c4a5afc91858a8cc2276bb7559a/compiler/one-cmds/utils.py#L204-L211

mhs4670go: This is invalid answers because `onecc-docker` doesn't have a dependency on `onecc`.

compiler/onecc-docker/onecc-docker.sh
mhs4670go: I used to thought `onecc-docekr` were placed at `/usr/share/one/`. But it seems not.

jyoungyun:
This PR is a draft code. We are not going to reflect this code, but we are going to split it into new PR and upload it again.
Based on the comments given by others, please upload a new PR for building a dockerfile and running a docker container as the first step. :)

/cc @Samsung/ootpg_docker

compiler/onecc-docker/Dockerfile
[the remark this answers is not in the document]
holawan: Oh! Thank you. I'll let you know after applying your advice!

PR #9716, opened 2022-09-15
Draft: Enable executor generation of multiple model with controlflow

hseok-oh:
Introduce executor index [ci skip]

### DIFF
--- a/runtime/onert/core/include/exec/Execution.h
+++ b/runtime/onert/core/include/exec/Execution.h
@@ -243,11 +243,8 @@ public:
   void sholudStop();
 
 private:
-  const std::unique_ptr<IExecutor> &primary_executor() const
-  {
-    return _executors->at(ir::SubgraphIndex{0});
-  };
-  std::unique_ptr<IExecutor> &primary_executor() { return _executors->at(ir::SubgraphIndex{0}); };
+  const IExecutor *primary_executor() const { return _executors->at(ir::ExecutorIndex{0}); };
+  IExecutor *primary_executor() { return _executors->at(ir::ExecutorIndex{0}); };
 
 private:
   const std::shared_ptr<Executors> _executors;
--- a/runtime/onert/core/include/exec/Execution.h
+++ b/runtime/onert/core/include/exec/Execution.h
@@ -243,8 +243,8 @@ public:
   void sholudStop();
 
 private:
-  const IExecutor *primary_executor() const { return _executors->at(ExecutorIndex{0}); };
-  IExecutor *primary_executor() { return _executors->at(ExecutorIndex{0}); };
+  const IExecutor *primary_executor() const { return _executors->primary_executor(); }
+  IExecutor *primary_executor() { return _executors->primary_executor(); }
 
 private:
   const std::shared_ptr<Executors> _executors;
--- a/runtime/onert/core/include/exec/Execution.h
+++ b/runtime/onert/core/include/exec/Execution.h
@@ -243,8 +243,8 @@ public:
   void sholudStop();
 
 private:
-  const IExecutor *primary_executor() const { return _executors->at(ir::ExecutorIndex{0}); };
-  IExecutor *primary_executor() { return _executors->at(ir::ExecutorIndex{0}); };
+  const IExecutor *primary_executor() const { return _executors->at(ExecutorIndex{0}); };
+  IExecutor *primary_executor() { return _executors->at(ExecutorIndex{0}); };
 
 private:
   const std::shared_ptr<Executors> _executors;
--- a/runtime/onert/core/include/exec/Executors.h
+++ b/runtime/onert/core/include/exec/Executors.h
@@ -19,11 +19,15 @@
 
 #include "IExecutor.h"
 #include "ir/NNPkg.h"
+#include "util/Index.h"
 
 namespace onert
 {
 namespace exec
 {
+
+struct ExecutorIndexTag;
+using ExecutorIndex = ::onert::util::Index<uint32_t, ExecutorIndexTag>;
 
 /**
  * @brief Class to gather executors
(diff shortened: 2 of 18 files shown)


runtime/onert/core/include/ir/Index.h
```diff
@@ -41,6 +41,10 @@ using SubgraphIndex = ::onert::util::Index<uint32_t, SubgraphIndexTag>;
 struct ModelIndexTag;
 using ModelIndex = ::onert::util::Index<uint32_t, ModelIndexTag>;
 
+// TODO Move to exec/???.h
```
glistening, line 44: Maybe exec/Executors.h ?

hseok-oh: Maybe

runtime/onert/core/include/exec/Executors.h
```diff
@@ -57,12 +62,22 @@ class Executors
...
+  }
+
 private:
-  // TODO Use Executor index
-  //      Changing index will effect if/while compile and kernel implementation
-  std::unordered_map<ir::SubgraphIndex, std::unique_ptr<IExecutor>> _executors;
+  std::unordered_map<ExecutorIndex, std::unique_ptr<IExecutor>> _executors;
   // NOTE _model_edges may use different struct type for executor implementation
   std::unique_ptr<ir::ModelEdges> _model_edges;
+  std::unordered_map<ExecutorIndex, std::pair<ir::ModelIndex, ir::SubgraphIndex>> _index_map;
```
glistening, line 79: @hseok-oh What about inverting key and value?
```cc
std::unordered_map<std::pair<ir::ModelIndex, ir::SubgraphIndex>, ExecutorIndex> _index_map;
```

runtime/onert/core/include/exec/Executors.h
```diff
@@ -57,12 +62,22 @@ class Executors
...
+      return ExecutorIndex{_next_index++};
+    else
+      return ExecutorIndex{};
+  }
+
 private:
-  // TODO Use Executor index
-  //      Changing index will effect if/while compile and kernel implementation
-  std::unordered_map<ir::SubgraphIndex, std::unique_ptr<IExecutor>> _executors;
+  std::unordered_map<ExecutorIndex, std::unique_ptr<IExecutor>> _executors;
```
glistening, line 76: @hseok-oh Why don't we just use `std::vector`. We need ordering.

runtime/onert/core/include/exec/Executors.h
hseok-oh: If we use index pair as key, we need to introduce hash function for index pair like index's hash function. I don't know it's good solution.

https://github.com/Samsung/ONE/blob/888235ed09309347e1d4455419f65f23545eb4a3/runtime/onert/core/include/util/Index.h#L151-L157

glistening: @hseok-oh Yes, defining a hash function for two basically integers is not complicated. Do you have any concern what I am missing? I think it is basically 64 bit integer to 32 bit integer hash map.

hseok-oh: No. You mean hash function like below?

```cpp
template <> struct hash<std::pair<::onert::ir::ModelIndex, ::onert::ir::SubgraphIndex>>
{
  size_t
  operator()(const std::pair<::onert::ir::ModelIndex, ::onert::ir::SubgraphIndex> &pair) const
    noexcept
  {
    return hash<uint64_t>()((uint64_t(pair.first.value()) << 32) | uint64_t(pair.second.value()));
  }
};
```

Then, we may use this pair on `_executors` map directly, and remove `_index_map`, then don't need to introduce `ExecutorIndex`. I'll try this.

glistening: @hseok-oh Yes, I guess we are on same page. I left more comment in https://github.com/Samsung/ONE/pull/9740#pullrequestreview-1114734308.

glistening:
@hseok-oh Thank you for update. It (= without `ExecutorIndex`) is more readable at least for me. I can review on `Executors.*`. ( I am not familiar with `IF`, `WHILE`, ..., and I have to read code. )

runtime/onert/core/include/exec/Executors.h
```diff
@@ -19,6 +19,22 @@
...
+namespace std
+{
+
+template <> struct hash<std::pair<::onert::ir::ModelIndex, ::onert::ir::SubgraphIndex>>
+{
+  size_t
+  operator()(const std::pair<::onert::ir::ModelIndex, ::onert::ir::SubgraphIndex> &pair) const
+    noexcept
+  {
+    return hash<uint64_t>()((uint64_t(pair.first.value()) << 32) | uint64_t(pair.second.value()));
```
glistening, line 33: (beyond this pr) As I mentioned in https://github.com/Samsung/ONE/pull/9716#discussion_r975996904, we may reduce bit width i32 to i16 for both `ModelIndex` and `SubgraphIndex`, which are unlikely to be beyond 2^<sup>15</sup>.

hseok-oh:
I've updated to use index pair and remove `ExecutorIndex`. I'll close PR #9740 and make new PR.

runtime/onert/core/include/exec/Executors.h
hseok-oh: Basically right. But we need to think about tflite and circle schema, and NNAPI spec. They are using 32bit for subgraph index. (we can reduce model index bit)

runtime/onert/core/include/exec/Execution.h
```diff
@@ -243,11 +243,8 @@ class Execution
   void sholudStop();
 
 private:
-  const std::unique_ptr<IExecutor> &primary_executor() const
-  {
-    return _executors->at(ir::SubgraphIndex{0});
-  };
-  std::unique_ptr<IExecutor> &primary_executor() { return _executors->at(ir::SubgraphIndex{0}); };
+  const IExecutor *primary_executor() const { return _executors->primary_executor(); }
+  IExecutor *primary_executor() { return _executors->primary_executor(); }
```
hseok-oh, line 247: I found `primary_executor()` is used to get input/output size. But by multi model, primary executor's input/output is not always same with package input/output. We need to fix this. (nnfwapi and may return invalid input/output size on multi model)

hseok-oh:
All PRs are merged

PR #9717, opened 2022-09-15
[one-cmds-test] Use variables

seanshpark:
This will revise tests to use variables.


seanshpark:
from draft #9707

jinevening (approved):
LGTM!

PR #9720, opened 2022-09-16
[one-cmds-test] Rename output circle use for input

seanshpark:
This will rename output circle name that is used as input for other tests.


seanshpark:
for https://github.com/Samsung/ONE/issues/9710#issuecomment-1248761588

seanshpark:
The one that gives other failure is `one-import_neg_006.test` which generates wrong tflite file.
This test that was introduced in TF2.3.0 had failed in TF but now doesn't exit with error but produces unexpected tflite file.
Currently this test itself is treated as SKIPPED.

We can go on with fixing only this test but as existing tests are used as copy source so there can be similar cases in the future.
This change is to prevent that situation.

jinevening (approved):
LGTM

Issue #9719, opened 2022-09-15
[luci] Extend Maximum Op in ConvertNCHWToNHWC


jinevening:
## What

Let's support Maximum Op whose inputs are both non-const in ConvertNCHWToNHWC

## Why

Our potential target model requires the pattern.

jinevening:
Done.

PR #9729, opened 2022-09-19
[luci] Support non-const inputs of Maximum in ConvertNCHWToNHWC

jinevening:
This supports non-const inputs of Maximum Op in ConvertNCHWToNHWC.

---
Related to: https://github.com/Samsung/ONE/issues/9719

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -503,6 +503,14 @@ bool is_NCHW(const luci::CirclePadV2 *node)
   return true;
 }
 
+bool is_const(const loco::Node *node)
+{
+  if (not dynamic_cast<const luci::CircleConst *>(node))
+    return false;
+
+  return true;
+}
+
 // NOTE Following conditions can be extended later
 // NOTE Used for Maximum, Miminum as ReLU/ReLU6
 //
@@ -869,6 +877,16 @@ class ConvertNCHWToNHWC final : public luci::CircleNodeMutableVisitor<bool>
       auto pre_trans = create_pre_transpose(node);
       pre_trans->a(pred_node);
       node->x(pre_trans);
+    }
+    else if ((not is_const(node->x())) and (not is_const(node->y())))
+    {
+      auto pre_trans_x = create_pre_transpose(node);
+      pre_trans_x->a(node->x());
+      node->x(pre_trans_x);
+
+      auto pre_trans_y = create_pre_transpose(node);
+      pre_trans_y->a(node->y());
+      node->y(pre_trans_y);
     }
     else
     {
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
@@ -530,6 +530,27 @@ public:
   luci::CircleConst *limit = nullptr;
 };
 
+class MaximumNonConstGraph final : public SimpleGraph
+{
+protected:
+  loco::Node *insertGraphBody(loco::Node *input) override
+  {
+    max = g.nodes()->create<luci::CircleMaximum>();
+    max->dtype(loco::DataType::FLOAT32);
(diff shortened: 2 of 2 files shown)


seanshpark (approved):
LGTM

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -870,6 +878,16 @@ class ConvertNCHWToNHWC final : public luci::CircleNodeMutableVisitor<bool>
       pre_trans->a(pred_node);
       node->x(pre_trans);
     }
+    else if ((not is_const(node->x())) and (not is_const(node->y())))
```
mhs4670go, line 881: Current codes look like this.

```cxx
    if (is_NCHW_with_s_const<luci::CircleMaximum>(node, pred_node, comp_constant))
    {
      auto pre_trans = create_pre_transpose(node);
      pre_trans->a(pred_node);
      node->x(pre_trans);
    }
    else if ((not is_const(node->x())) and (not is_const(node->y())))
    {
      auto pre_trans_x = create_pre_transpose(node);
      pre_trans_x->a(node->x());
      node->x(pre_trans_x);

      auto pre_trans_y = create_pre_transpose(node);
      pre_trans_y->a(node->y());
      node->y(pre_trans_y);
    }
    else
    {
      // TODO support other cases
      return false;
    }
```

Seems considering only y const case is missing. How about this logic instead?
```cxx
    if (is_const(node->x())
    {
      auto pre_trans = create_pre_transpose(node);
      pre_trans->a(pred_node);
      node->x(pre_trans);
    }
    if (is_const(node->y())
    {
      auto pre_trans_y = create_pre_transpose(node);
      pre_trans_y->a(node->y());
      node->y(pre_trans_y);
    }
```

Of course you should also revisit `is_NCHW_with_s_const` method.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
```diff
@@ -530,6 +530,27 @@ class MaximumGraph final : public SimpleGraph
...
+{
+protected:
+  loco::Node *insertGraphBody(loco::Node *input) override
+  {
+    max = g.nodes()->create<luci::CircleMaximum>();
+    max->dtype(loco::DataType::FLOAT32);
+    max->shape({1, 16, 4, 4});
+
+    max->x(input);
+    max->y(input);
```
mhs4670go, line 543: How about create two input graph for this?

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
jinevening: > Seems considering only y const case is missing

The case is handled by `is_NCHW_with_s_const`.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
jinevening: That could be another test. The graph with a single input can test what this PR intends to implement.

compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
mhs4670go: Ah, I mean the case y isn't const case. It always call just `x` method. Is it right?

mhs4670go: (optional) It'd be better to rename `is_NCHW_with_s_const` methdo to `set_NCHW_with_s_const` because "is" prefix doesn't look like it is going to setting something.

jinevening: You're right. node->y() should be called if y is non-const. Existing code should be updated. Thanks for finding.

jinevening: Let me fix it in another PR :)

mhs4670go (approved):
LGTM

PR #9735, opened 2022-09-19
[luci] Set correct inputs for Maximum/Minimum in ConvertNCHWToNHWC

jinevening:
This sets correct inputs for Maximum/Minimum Ops in ConvertNCHWToNHWC.

---
Related to: https://github.com/Samsung/ONE/pull/9729#discussion_r973897656

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
@@ -511,44 +511,20 @@ bool is_const(const loco::Node *node)
   return true;
 }
 
-// NOTE Following conditions can be extended later
-// NOTE Used for Maximum, Miminum as ReLU/ReLU6
-//
-// Find T with an NCHW pattern described below
-//   - Input (non-constant) shape : [N, C, H, W]
-//   - Input (constant) shape : [1] or []
-//   - Output shape : [N, C, H, W]
-template <class T>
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp
```diff
@@ -511,44 +511,20 @@ bool is_const(const loco::Node *node)
...
+  if (not const_node)
     return false;
-  }
 
-  if (pred_node->rank() != 4)
+  const auto const_rank = const_node->rank();
+  // shape of scalar
+  // 1. rank = 0
+  // 2. rank = 1, dimension = 1
+  if (const_rank != 0 and (const_rank == 1 && const_node->dim(0).value() == 1))
```
mhs4670go, line 524: Maybe?
```suggestion
  if (const_rank != 0 and (const_rank != 1 || const_node->dim(0).value() != 1))
```
(optional)
IMHO, check if true condition seems easier to read. I mean original code looks better.

mhs4670go: And even though condition is wrong, all the tests are failed. Related test might be needed.

jinevening: You're right. I've updated conditions and added tests. PTAL

Issue #9721, opened 2022-09-16
[ONE-Dockerizing] shell script related to onecc-docker


Seunghui98:
## What

To implement onecc-docker, we created a shell script for creating a docker image file, building an image, and executing a container.

## Why

Regarding [#8232](https://github.com/Samsung/ONE/issues/8232) issue, we are going to do Dockerizing so that more users can easily use onecc
Related Issue : [#9714](https://github.com/Samsung/ONE/pull/9714), #9712  

### Work flow
- Get the version of onecc installed on the Host PC
- Get version for latest onecc
   - Output exception message when request fails
- If the latest onecc is installed on the Host PC are the same, the input onecc-cmd is executed. 
- If the latest one is not installed, onecc-docker containers runs.
   - If one image does not exist, generates onecc-docker image file through write_dockerfile() and build onecc-docker image.
   - Run the onecc-docker container.

### How to test
I currently proceeded with Dockerizing in the following way.
Since onecc-docker was mounted as `/home/`, we tested it to create a circle file in `/home/test`.

#### summary

1. Import TensorFlow Model for Dockerizing and write Configure File.
2. Write a Dockerfile to create a Docker image and run the container.
    - When `onecc` is finished running, the container is terminated and deleted.
3. Check that circle file and nnpackage have been created.

#### Import Tensorflow model for testing

First, I created the path '/home/test/' for Linux (wsl) and downloaded the model file to test. After downloading [inception_v3](https://storage.googleapis.com/download.tensorflow.org/models/tflite/model_zoo/upload_20180427/inception_v3_2018_04_27.tgz), the Configure file was created.

```
# /home/test/
/home/test$ wget https://storage.googleapis.com/download.tensorflow.org/models/tflite/model_zoo/upload_20180427/inception_v3_2018_04_27.tgz

/home/test$ tar -xvf inception_v3_2018_04_27.tgz

/home/test$ vi onecc.template.cfg

/home/test$ tree
.
├── inception_v3.pb
├── inception_v3.tflite
├── inception_v3_2018_04_27.tgz
├── labels.txt
└── onecc.template.cfg
```

- onecc.template.cfg

    ```
    [onecc]
    one-import-tf=True
    one-import-tflite=False
    one-import-bcq=False
    one-import-onnx=False
    one-optimize=True
    one-quantize=False
    one-parition=False
    one-pack=True
    one-codegen=False
    
    [one-import-tf]
    input_path=inception_v3.pb
    output_path=inception_v3.circle
[...]

PR #9737, opened 2022-09-19
[ONE-Dockerizing] Scripts that can run onecc using docker containers

holawan:
> 1st PR

This PR is a script that builds images and drives containers to run onecc with docker containers

ONE-DCO-1.0.-Signed-off-by: Dongwan Kim <asdf134652@gmail.com>
> 2nd PR

This PR is an improved script to reflect the review

- How to get version
- Functionalize duplicate use options
- Check Dockerfile Existence
- Run docker container

>3rd PR 

This PR has added error handling logic

> 4th PR 

This PR was created to add README to "onecc-docker".

>5th PR

This PR has added an API request excess exception handling code

parent issue: #9712 #9721 
draft: #9722 

/cc [@Samsung/ootpg](https://github.com/orgs/Samsung/teams/ootpg)


seanshpark:
@jyoungyun , I think `compiler/one-docker` folder python files isn't checked with format checker.

mhs4670go:
Please add empty line to the place between the commit body and the sign-off. You can see other commits with `git log` command.

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+        return True 
+    else :
+        return False
+
+
+
+def main():
+    work_dir = os.getcwd()
+    response = requests.get("https://api.github.com/repos/Samsung/ONE/tags")
+    if  response.status_code != 200 :
```
mhs4670go, line 51: How about remove the magic number for better readability?
```suggestion
    OK_RESPOSE = 200
    if  response.status_code != OK_RESPOSE:
```

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+        print("Error was occured while downloading ONE release tags")
+        print("status_code: ", response.status_code)
+        print("Error_reason: " + response.reason)
+        exit(1)
```
mhs4670go, line 55: Requests library already has exception handling features. Since I don't run this feature with this script, the codes should be tested before applying this code snippet.

```suggestion
    TAG_URL = "https://api.github.com/repos/Samsung/ONE/tags"
    try:
        r = requests.get(TAG_URL)
    except requests.exceptions.RequestException as e:
        raise SystemExit(e)
```

### Reference
https://stackoverflow.com/questions/16511337/correct-way-to-try-except-using-python-requests-module
https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions

compiler/onecc-docker/onecc-docker
mhs4670go: This review has become outdated with https://github.com/Samsung/ONE/pull/9737#discussion_r974094032.

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+    
+    versions_str = response.content
+    versions_json = json.loads(versions_str)
+    recent_version = versions_json[0]["name"]
```
mhs4670go, line 59: github api always returns latest version of release at the first of the content?

How about use https://docs.github.com/en/rest/releases/releases#get-the-latest-release?

It means,
```python
TAG_URL = "https://api.github.com/repos/Samsung/ONE/releases/latest" # NOTE HERE
try:
    response = requests.get(TAG_URL)
except requests.exceptions.RequestException as e:
    raise SystemExit(e)
```

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+    if dic['REPOSITORY'] == tag_name[0] and dic['TAG'] == tag_name[1]  :
+        return True 
+    else :
+        return False
+
+
+
+def main():
+    work_dir = os.getcwd()
+    response = requests.get("https://api.github.com/repos/Samsung/ONE/tags")
```
mhs4670go, line 50: For reviewers, this is the example content of response.
```json
{
  "url": "https://api.github.com/repos/octocat/Hello-World/releases/1",
  "html_url": "https://github.com/octocat/Hello-World/releases/v1.0.0",
  "assets_url": "https://api.github.com/repos/octocat/Hello-World/releases/1/assets",
  "upload_url": "https://uploads.github.com/repos/octocat/Hello-World/releases/1/assets{?name,label}",
  "tarball_url": "https://api.github.com/repos/octocat/Hello-World/tarball/v1.0.0",
  "zipball_url": "https://api.github.com/repos/octocat/Hello-World/zipball/v1.0.0",
  "discussion_url": "https://github.com/octocat/Hello-World/discussions/90",
  "id": 1,
  "node_id": "MDc6UmVsZWFzZTE=",
  "tag_name": "v1.0.0",
  "target_commitish": "master",
  "name": "v1.0.0",
  "body": "Description of the release",
  "draft": false,
  "prerelease": false,
  "created_at": "2013-02-27T19:35:32Z",
  "published_at": "2013-02-27T19:35:32Z",
  "author": {
    "login": "octocat",
    "id": 1,
    "node_id": "MDQ6VXNlcjE=",
    "avatar_url": "https://github.com/images/error/octocat_happy.gif",
    "gravatar_id": "",
    "url": "https://api.github.com/users/octocat",
    "html_url": "https://github.com/octocat",
    "followers_url": "https://api.github.com/users/octocat/followers",
    "following_url": "https://api.github.com/users/octocat/following{/other_user}",
    "gists_url": "https://api.github.com/users/octocat/gists{/gist_id}",
    "starred_url": "https://api.github.com/users/octocat/starred{/owner}{/repo}",
    "subscriptions_url": "https://api.github.com/users/octocat/subscriptions",
    "organizations_url": "https://api.github.com/users/octocat/orgs",
    "repos_url": "https://api.github.com/users/octocat/repos",
    "events_url": "https://api.github.com/users/octocat/events{/privacy}",
    "received_events_url": "https://api.github.com/users/octocat/received_events",
    "type": "User",
    "site_admin": false
  },
  "assets": [
    {
      "url": "https://api.github.com/repos/octocat/Hello-World/releases/assets/1",
      "browser_download_url": "https://github.com/octocat/Hello-World/releases/download/v1.0.0/example.zip",
      "id": 1,
      "node_id": "MDEyOlJlbGVhc2VBc3NldDE=",
      "name": "example.zip",
      "label": "short description",
      "state": "uploaded",
      "content_type": "application/zip",
      "size": 1024,
      "download_count": 42,
      "created_at": "2013-02-27T19:35:32Z",
      "updated_at": "2013-02-27T19:35:32Z",
      "uploader": {
        "login": "octocat",
        "id": 1,
        "node_id": "MDQ6VXNlcjE=",
        "avatar_url": "https://github.com/images/error/octocat_happy.gif",
        "gravatar_id": "",
        "url": "https://api.github.com/users/octocat",
        "html_url": "https://github.com/octocat",
        "followers_url": "https://api.github.com/users/octocat/followers",
        "following_url": "https://api.github.com/users/octocat/following{/other_user}",
        "gists_url": "https://api.github.com/users/octocat/gists{/gist_id}",
        "starred_url": "https://api.github.com/users/octocat/starred{/owner}{/repo}",
        "subscriptions_url": "https://api.github.com/users/octocat/subscriptions",
        "organizations_url": "https://api.github.com/users/octocat/orgs",
        "repos_url": "https://api.github.com/users/octocat/repos",
        "events_url": "https://api.github.com/users/octocat/events{/privacy}",
        "received_events_url": "https://api.github.com/users/octocat/received_events",
        "type": "User",
        "site_admin": false
      }
    }
  ]
}
```

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+
+    image_name = f"onecc:{recent_version}"
+    build_arg = f"VERSION={recent_version}"
+    if not _is_image(recent_version):
+        build_cmd = ["docker", "build", "-t", image_name, "--build-arg", build_arg, work_dir]
+        subprocess.call(build_cmd)
+
+    contianer_name = f"onecc_{recent_version.replace('.','_')}"
+    user_cmd = ''.join(sys.argv[1:])
+    run_cmd = ["docker", "run", "--rm", "--name", contianer_name, "-v", "${HOME}:${HOME}", "-e", "HOME=${HOME}", "-w", work_dir, image_name, user_cmd]
```
mhs4670go, line 69: @jyoungyun I'm not sure about mounting host's HOME directory. How about receive volume path from user or use temporary dir if path is not given?

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+
+import sys
+import subprocess
+import json
+import os
+import requests
+
+
+def _is_image(version) : 
+    cmd = ['docker', 'images', f'onecc:{version}']
```
mhs4670go, line 26: Using `-q` option will do the trick:)

```bash
$ docker images -q onecc:1.21.0
d243dfsfeid # image id
$ docker images -q onecc:wrong_version
# return nothing
```
Or you can also use `docker inspect`.
```bash
$ docker inspect --type=image onecc:1.21.0
```

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+# limitations under the License.
+
+import sys
+import subprocess
+import json
+import os
+import requests
+
+
+def _is_image(version) :
```
mhs4670go, line 25:
```suggestion
def _image_exists(name) : # or _image_exists(tag) 
```

1. `_is_image` sounds like it checks if the thing is image or not. How about use `exist` verb?
2. Receiving `name` directly rather than version seems more intuitive and would reduce the duplication of information.

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+    build_arg = f"VERSION={recent_version}"
+    if not _is_image(recent_version):
+        build_cmd = ["docker", "build", "-t", image_name, "--build-arg", build_arg, work_dir]
+        subprocess.call(build_cmd)
+
+    contianer_name = f"onecc_{recent_version.replace('.','_')}"
+    user_cmd = ''.join(sys.argv[1:])
+    run_cmd = ["docker", "run", "--rm", "--name", contianer_name, "-v", "${HOME}:${HOME}", "-e", "HOME=${HOME}", "-w", work_dir, image_name, user_cmd]
+    cmd = ' '.join(run_cmd)
+    subprocess.run(cmd,shell=True)
```
mhs4670go, line 71: In this `onecc-docker`, there are three method call when the subprocess is invoked - `subprocess.run`, `subprocess.call`, `Popen`. Is there a reason to use three different functions?

How about make a method that runs a command?
```python
def run_subprocess(cmd):
    ret = subprocess.run(cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    # do something or return the results
    # error handling might be needed
```

holawan:
> How about write `README.md` for users?

Yes, I think README.md is necessary. Can I add it to this PR?
I'm sorry I'm not used to it yet.

compiler/onecc-docker/onecc-docker
mhs4670go: @holawan Even if we add the features about selecting version, I think we would use two URL later - latest and tags(or releases). Because we can't assume tags or releases api returns versions in the specific order.  

I mean that we should use `latest` URL in the first place.

mhs4670go:
> Can I add it to this PR?

compiler/onecc-docker/onecc-docker
holawan: > I'm not sure about mounting host's HOME directory


The reason why we mounted `home` is because we considered the following cases.

**User command** 
```
~/onecc-docker/test$ onecc-docker -C ../onecc.template.cfg 
```
**onecc.template.cfg** 
```
...
[one-import-tflite]
# mandatory
; tflite file
input_path=../inception_v3.tflite
; circle file
...
```
**file structure** 
```
~/onecc-docker$ tree5
.
├── Dockerfile
├── inception_v3.pb
├── inception_v3.tflite
├── onecc-docker
├── onecc.template.cfg
├── onecc_028.workflow.json
└── test
    ├── inception_v3.circle
    ├── inception_v3.circle.log
    ├── inception_v3.opt.circle
    ├── inception_v3.opt.circle.log
    ├── inception_v3_pack
    │   └── inception_v3.opt
    │       ├── inception_v3.opt.circle
    │       └── metadata
    │           └── MANIFEST
    ├── inception_v3_pack.log
    └── onecc-docker

4 directories, 14 files
```

As in the example above, I thought that the user did not know where to receive the input or where to expose the output.

> How about receive volume path from user or use temporary dir if path is not given?

Also, it is hard to understand your ungiven path. Can you explain in detail?

I don't know if that's the right answer, I'm sorry.

compiler/onecc-docker/onecc-docker
holawan: @mhs4670go 
Thank you for the detailed explanation. I'll check again tomorrow!

holawan:
@mhs4670go , @jyoungyun 
There seems to be a lot of corrections in our code. 
May I continue to talk in this PR and improve the code?

compiler/onecc-docker/onecc-docker
mhs4670go: > The previous plan seems to cause you a lot of confusion.
Would it be better to exclude those parts?

Actually, I think whether the previous plan exists or not doesn't matter. Even if the plan existed, the function name still would have seemed to check if given version of image exists or not.

I don't know what you wrote in the first place where you used to consider various versions but even considering multiple versions, the explanation in the above my comment would not be changed. Just a method listing the release versions might be needed later.

In short, you don't have to exclude those parts because we should always consider extensibility. I'm reviewing the codes with that mind:) 

Please let me know if you have any comments!

mhs4670go:
> May I continue to talk in this PR and improve the code?

Yes. Splitting the PR will split the information, which makes it difficult to find the history later.

compiler/onecc-docker/onecc-docker
mhs4670go: > Are you saying that I should simply check the image in the way you mentioned?

Strictly speaking, no. You can use another way. I just recommend it for simple implementation. Even if you are considering the case various version comes in, the method `_image_exists` still seems to need a simple check. Otherwise, you should rename the method or make another method for SRP(Single Responsibility Principle).

compiler/onecc-docker/onecc-docker
mhs4670go: > Also, it is hard to understand your ungiven path. Can you explain in detail?

I thought below option. But temporary directory things turns out to be invalid. I used `pwd` instead.
```bash
$ onecc-docker -C in.cfg -w my_dir # -w (--working_dir)
# docker run .. -v my_dir:my_dir
$ onecc-docker -C in.cfg # working diectory not given
# docker run .. -v $(pwd):$(pwd)
```

> The reason why we mounted `home` is because we considered the following cases.

Well, I understood what you were trying to say. But I'm not sure it is safe or not. We need a discussion about this.

@jyoungyun Is it okay with mounting users' HOME directory?

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+        exit(1)
+    
+    versions_str = response.content
+    versions_json = json.loads(versions_str)
+    recent_version = versions_json[0]["name"]
+
+    image_name = f"onecc:{recent_version}"
+    build_arg = f"VERSION={recent_version}"
+    if not _is_image(recent_version):
+        build_cmd = ["docker", "build", "-t", image_name, "--build-arg", build_arg, work_dir]
```
jyoungyun, line 64: Since the two files are not in the same location in the future, the location of the `Dockerfile` must be specified.

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+    tag_name = cmd[2].split(':')
+
+    if dic['REPOSITORY'] == tag_name[0] and dic['TAG'] == tag_name[1]  :
+        return True 
+    else :
+        return False
+
+
+
+def main():
```
jyoungyun, line 48: Unfortunately, I can't use the github rest api without authorization due to in-house network issue. So I suggest to support request header so that the users who need it can use it.

```
headers["Authorization"] = "token {}".format(/*token*/)
response = requests.get("https://api.github.com/repos/Samsung/ONE/tags", headers=headers)
```

https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting
https://docs.github.com/en/rest/overview/resources-in-the-rest-api#oauth2-token-sent-in-a-header

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+''''exec "python3" "$0" "$@" #'''
```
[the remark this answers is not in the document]
seanshpark: Q) I'm curious about why use `exec "python3"` instead of `#!/usr/bin/env python3` ?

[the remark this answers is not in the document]
mhs4670go: @jyoungyun You mean `python3 onecc-docker`?

But with python3 shebang, it could run with below command.
```bash
$ ./onecc-docker
```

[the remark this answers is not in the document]
seanshpark: > $ ./onecc-docker

this file needs +x attrite set but currently not. is it normal to turn it on? or off for safetly?

jyoungyun:
> I think `compiler/one-docker` folder python files isn't checked with format checker.

I made #9742 PR.

@holawan  Could you rebase this PR after merging #9742 PR. :)

compiler/onecc-docker/onecc-docker
morumoru: I'll test this code for exeception handling! Thank you for advice

compiler/onecc-docker/onecc-docker
Seunghui98: @jyoungyun 
You said that `Dockerfile` would exist in a specific location, where do you think it is?

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,74 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import subprocess
+import json
+import os
+import requests
+
+
```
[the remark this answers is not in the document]
morumoru: You mean that make -h option on onecc-docker with argparse library?
I think that -h option is duplicated with onecc's -h option.
Should we make new flag for getting onecc docker's arguments?

compiler/onecc-docker/onecc-docker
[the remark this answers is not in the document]
mhs4670go: @seanshpark We don't need to set it until installation. `one-cmds` tools have +x attribute after installation.

compiler/onecc-docker/onecc-docker
holawan: > Could you add user and group id options so that the result generated in the docker image can be the same user as the host?


@jyoungyun I understand what you are saying, but there are difficulties in the implementation process. Do you mean to modify the docker file to add information about uid and gid?
Here's what I did, but I don't think this is right.

```
docker run --rm --name {container_name} -u 1000 -v ${HOME}:${HOME} -e HOME=${HOME} -w ${PWD} onecc:1.21.0 

$ docker exec {container_name} id
uid=1000 gid=0(root) groups=0(root)
```

jyoungyun: You can use `id -u` for user id and `id -g` for group id. :)

jyoungyun: https://medium.com/redbubble/running-a-docker-container-as-a-non-root-user-7d2e00f8ee15

This link will help you :)

holawan: I checked it well. Thank you!

compiler/onecc-docker/onecc-docker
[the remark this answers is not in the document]
jyoungyun: I think `onecc-docker` can take own argument. For example, if the user want to pass a personal access token to the request headers, `onecc-docker` should take a token value from argument. In view of this case, this script should implement a separate argparser.

compiler/onecc-docker/onecc-docker
Seunghui98: @jyoungyun 
If I make a Debian package and run onecc-docker, there is a little problem with referring to Dockerfile, so would it be good to make an issue about this?

jyoungyun: > You said that `Dockerfile` would exist in a specific location, where do you think it is?

The `onecc-docker` script will be located in `/usr/bin` so that the user can execute it directly.
And the `Dockerfile` will be located in `/usr/share/one/docker`.

> If I make a Debian package and run onecc-docker, there is a little problem with referring to `Dockerfile`, so would it be good to make an issue about this? 

Sure. :)
But I don't think the code related to debian is related to this PR.
I want this PR to be merged first regardless of debian related things.

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,75 @@
...
+        _run(build_cmd)
+
+    contianer_name = f"onecc_{recent_version.replace('.','_')}"
+    user_cmd = ''.join(sys.argv[1:])
+
+    run_cmd = [
+        "docker", "run", "--rm", "-u", "$(id -u):$(id -g)"
+        "--name", contianer_name, "-v", "${HOME}:${HOME}", "-e", "HOME=${HOME}", "-w",
+        "${PWD}", image_name, user_cmd
+    ]
```
holawan, line 66:
```suggestion
    run_cmd = [
        "docker", "run", "--rm", "-u", "$(id -u):$(id -g)",
        "--name", contianer_name, "-v", "${HOME}:${HOME}", "-e", "HOME=${HOME}", "-w",
        "${PWD}", image_name, user_cmd
    ]
```

I found a comma missing from the list late.
I'll revise it soon and commit it again!

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,83 @@
...
+    image_name = f"onecc:{recent_version}"
+    build_arg = f"VERSION={recent_version}"
+
+    if not _image_exists(image_name):
+        build_cmd = [
+            "docker", "build", "-t", image_name, "--build-arg", build_arg, script_path
+        ]
+        print('build Docker image ...')
+        _run(build_cmd)
+        print('Dockerfile successfully built.')
```
holawan, line 66: @jyoungyun , when the user does not have a docker image, the build time takes about 10 minutes. Is it necessary to print out the flow of the docker being built?

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
+# onecc-docker
+
+`onecc-docker` supports `one-cmd`, which is currently only supported by ubuntu 18.04, in more OS.
```
mhs4670go, line 3:
```suggestion
`onecc-docker` supports `one-cmds`, which is currently only supported on ubuntu 18.04, in more OS.
```

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+
+
+## Description
+
+For now, `one-cmds` tools only support Ubuntu 18.04 and 20.04(not officially).
+So people in other environments can't use our tools unless they upgrade the OS (or install Ubuntu OS).
+
+Therefore, we developed `onecc` that runs using a docker so that users can use `one-cmds` more easily. This is `onecc-docker.`
```
mhs4670go, line 12:
```suggestion
Therefore, we developed `onecc` that runs using a docker so that users can use `one-cmds` more easily. This is _onecc-docker_.
```

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,84 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import subprocess
+import json
+import os
+import requests
+import argparse
```
mhs4670go, line 22: This `argparse` has been added for https://github.com/Samsung/ONE/pull/9737/files#r974949214. Is it going to be used?

holawan: We haven't started the task on "argparse" yet.
I am trying to use it according to @jyoungyun's  opinion.
I think I can give you a detailed answer after the meeting tomorrow.

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+
+
+## Requirements
+
+- Linux OS
+
+- Docker
+
+    - `onecc-docker` requires the current `user ID` to be included in the `docker group` because it requires the docker-related commands to be executed without `sudo` privileges.
```
jyoungyun, line 24:
```suggestion
    - Requires root privileges.
          - `onecc-docker` requires the current `user ID` to be included in the `docker group` because it requires the docker-related commands to be executed without `sudo` privileges.
          - See "[Post-installation steps for Linux](https://docs.docker.com/engine/install/linux-postinstall/)".
```

compiler/onecc-docker/onecc-docker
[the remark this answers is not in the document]
jyoungyun: If `onecc` is already installed when running `onecc-docker`, shows `onecc` help message and if `onecc` is not installed, shows `onecc-docker` help message. What about this solution?

First of all, let's put the PR without considering `-h` option handling.

compiler/onecc-docker/onecc-docker
jyoungyun: Could you add token argument for in-house developers?

compiler/onecc-docker/onecc-docker
jyoungyun: Yes, I think it is necessary.

compiler/onecc-docker/onecc-docker
```diff
@@ -0,0 +1,84 @@
...
+
+def main():
+    script_path = os.path.dirname(os.path.realpath(__file__))
+
+    TAG_URL = "https://api.github.com/repos/Samsung/ONE/releases/latest"
+    try:
+        response = requests.get(TAG_URL)
+        response.raise_for_status()
+    except requests.exceptions.RequestException as e:
+        raise SystemExit(e)
```
jyoungyun, line 52: Could you user-friendly print an error message?

`SystemExit('blah blah blah')`

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+- To prevent repetitive image generation, we verify that there is an `onecc` image of the latest `one` release version.
+- If not, build the docker image using the docker file.
+
+### Run Docker Container
+
+- When the above processes are successfully completed, we drive the docker container with the command the user wants. `onecc` runs on the user's current path and returns the result to the command line.
+
+
+
+## Examples
```
jyoungyun, line 55: What about separating this content into a separate file like `tutorial.md`?

compiler/onecc-docker/onecc-docker
holawan: yes let's try that.

compiler/onecc-docker/onecc-docker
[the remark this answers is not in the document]
holawan: > If onecc is already installed when running onecc-docker, shows onecc help message and if onecc is not installed, shows onecc-docker help message. What about this solution?

Yes, after finalizing existing developments, it would be a good idea to consider them!

>First of all, let's put the PR without considering -h option handling.

Okay.

holawan:
@jyoungyun , we will proceed with code refactoring by collecting your opinions. If our code has been fixed and you think it's okay to merge, can you please tell me?

compiler/onecc-docker/onecc-docker
morumoru: When error occured in requests for example "API rate limit exceeded for 211.187.252.208. (But here\'s the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details." message is printed to bash with 403 error code. (Git api send this message from server)
Should I print more information except this?

@jyoungyun

compiler/onecc-docker/README.md
seanshpark: Q) what is `in more OS.` ?

compiler/onecc-docker/README.md
seanshpark: > we developed `onecc` that runs using a docker

I understand this as `onecc` is already running in docker

```suggestion
To overcome this limitation, we provide _onecc-docker_ that runs using a docker so that users can use `one-cmds` more widely.
```

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
+# onecc-docker
+
+`onecc-docker` supports `one-cmd`, which is currently only supported by ubuntu 18.04, in more OS.
+
+
+
+## Description
+
+For now, `one-cmds` tools only support Ubuntu 18.04 and 20.04(not officially).
+So people in other environments can't use our tools unless they upgrade the OS (or install Ubuntu OS).
```
seanshpark, line 10: > they upgrade the OS

Actually, users may use recent Ubutuntu 22.04. Thus it is likely `they downgrade the OS`

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+Therefore, we developed `onecc` that runs using a docker so that users can use `one-cmds` more easily. This is `onecc-docker.`
+
+> Currently, `onecc-docker` is only supported in Linux environments.
```
seanshpark, line 14: Why is ther a `>` ?

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+
+
+## Requirements
+
+- Linux OS
```
seanshpark, line 20: (optional) how about `Any Linux distribution` ?

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Structure
+
+### Summary 
+
+`onecc-docker` creates a docker image with the latest release version of `ONE`, runs `onecc` inside the docker container after the container is driven to deliver the desired result to the user.
+
+### To check the latest version of ONE
+
+- We use github api to get the latest release version of ONE. Create a docker file using the imported version.
```
seanshpark, line 42: Maybe `github api` -> `GitHub API`

seanshpark: I'm not sure about this line is really necessary.

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Check Docker Image Presence and Build Image
+
+- To prevent repetitive image generation, we verify that there is an `onecc` image of the latest `one` release version.
+- If not, build the docker image using the docker file.
```
seanshpark, line 47: `build` -> `rebuild` ?

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Run Docker Container
+
+- When the above processes are successfully completed, we drive the docker container with the command the user wants. `onecc` runs on the user's current path and returns the result to the command line.
```
seanshpark, line 51: `processes` -> `steps` ?

Maybe descript steps in numbe bullets than using `###` style?

While reading here `## Structure` seems better with `### Steps`
I was expecting structure of this module or structure of how the docker works with _one-cmds_ but looks like procedure of codes are working.

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+
+
+## Examples
+
+### Preparations
+
+We prepare the Tensorflow file, configure file, and workflow for the test.
```
seanshpark, line 59:
```suggestion
Prepare source files like below;
```

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,263 @@
...
+├── inception_v3.pb
+├── inception_v3.tflite
+├── onecc-docker
+├── onecc.template.cfg
+└── onecc.workflow.json
+
+0 directories, 6 files
+```
+
+#### `-h` ,`--help`
```
seanshpark, line 74: maybe still WIP?

compiler/onecc-docker/onecc-docker
jyoungyun: Yes. I think the message is not friendly to the users. :) 
You can see some error messages in https://github.com/Samsung/ONE/blob/master/compiler/one-cmds/utils.py file.

jyoungyun:
@holawan 

> we will proceed with code refactoring by collecting your opinions. If our code has been fixed and you think it's okay to merge, can you please tell me?

Of course. If there is nothing more to modify, I will let you know. :)

compiler/onecc-docker/onecc-docker
morumoru: Thank you for advice! I'll check utils.py.

compiler/onecc-docker/README.md
holawan: Do you mean help message only for `onecc-docker`? This is under discussion with @jyoungyun .

compiler/onecc-docker/README.md
holawan: That must be clear. I'll reflect on it.

compiler/onecc-docker/README.md
jyoungyun: Let's discuss after the code supporting the input arguments is shared. :)

compiler/onecc-docker/README.md
holawan: In my opinion, we build the docker image only once (unless the user intentionally deletes the image), what we repeat is considered to be the creation and deletion of the container. 
@jyoungyun  
What do you think?

compiler/onecc-docker/README.md
holawan: If you mean line 41, I'll remove it. Thank you!

compiler/onecc-docker/README.md
holawan: That sounds better!

compiler/onecc-docker/README.md
holawan: When I usually organize documents, I used BlockQuote for further explanation. I checked other documents, and it doesn't seem common. I'll fix it!

compiler/onecc-docker/README.md
holawan: > Actually, users may use recent Ubutuntu 22.04. Thus it is likely they downgrade the OS

How about changing it like this?

Therefore, it is difficult for people in different environments to use our tools without using ubuntu18:04.

compiler/onecc-docker/README.md
jyoungyun: > `build` -> `rebuild` ?

@seanshpark I can't catch your intention. :( Could you explain more?
The docker image is built only once there is no image with the same name. So I think `build` is more suitable for this step...

compiler/onecc-docker/README.md
seanshpark: No... it's line 42, explanation seems internal implementation(?)
Maybe I couldn't catch the whole story how the explanation is going.
I think this is related with using word `structure`

compiler/onecc-docker/README.md
seanshpark: Well, this maybe because I'm not a good Docker user/developer.
What I see is a Docker image. from the explanation, when the `latest` is changed, it sounds like Docker image is built again. So from the though of `again`, it looks like `rebuild`.
Technically it looks like `build` not `rebuild` so, we can go with `build`

compiler/onecc-docker/onecc-docker
morumoru: > You can use `id -u` for user id and `id -g` for group id. :)

I am using virtualbox to test our script and using mounted folder with Window and Linux.
Ubuntu user who want to accpet mounted folder need to belong to vboxsf group.
but when using only id -u and id -g option on docker , like docker run -u $(id -u):$(id -g), vboxsf is not inserted into docker user's group so onecc-docker can't access mounted folder

@jyoungyun 
Shoud I insert all groups to docker user's group to care the situation like this or don't care about it and just use docker run -u $(id -u):$(id -g)?

compiler/onecc-docker/README.md
jyoungyun: How about using a different name instead of `structure`?
I don't think the title and content match.

compiler/onecc-docker/onecc-docker
morumoru: > You can use `id -u` for user id and `id -g` for group id. :)

I am using virtualbox to test our script and using mounted folder with Window and Ubuntu.
Ubuntu user who want to accpet mounted folder need to belong to vboxsf group.
but when using only id -u and id -g option on docker , like docker run -u $(id -u):$(id -g), vboxsf is not inserted into docker user's group so onecc-docker can't access mounted folder

@jyoungyun
Shoud I insert all current user's groups to docker user's group to care the situation like this or don't care about it and just use docker run -u $(id -u):$(id -g)?

compiler/onecc-docker/README.md
holawan: > Q) what is in more OS. ?

More OSs can provide it.

compiler/onecc-docker/onecc-docker
jyoungyun: It is cautious to put all user groups. Even if it's a little inconvenient for you to test, what about keeping the existing code?
And in your case, can you change the user id of the mounted folder to ubuntu user?

morumoru: Ok I'll keep existing code! Thank you!

compiler/onecc-docker/README.md
holawan: This means that `onecc-docker` supports running `onecc` in environments other than ubuntu:18.04.

seanshpark: How about
- _onecc-docker_ can expand `one-cmds` in more platforms than just Ubuntu 18.04/20.04.
- or
- We can use `one-cmds` in more platorms than just Ubuntu 18.04/20.04 through _onecc-docker_.

compiler/onecc-docker/README.md
```diff
@@ -0,0 +1,267 @@
...
+            - `onecc-docker` requires the current `user ID` to be included in the `docker group` because it requires the docker-related commands to be executed without `sudo` privileges.
+                  - See "[Post-installation steps for Linux](https://docs.docker.com/engine/install/linux-postinstall/)"
+        
+        ```
+        sudo usermod -aG docker ${USER}
+        ```
+
+- Python 3.8
+
+
```
jyoungyun, line 34:
```suggestion
## Note

_onecc-docker_ is currently in incubation stage.

The onecc-docker debian package should be created with one-compiler debian package when ONE
compiler project builds. To this end, it is correct to configure the onecc-docker debian codes in
./infra/debian/compiler directory. However, we are currently working on the code, so we will
temporarily implement it in this location.

TODO: Merge this debian directory into ./infra/debian/compiler code.
```

PR #9739, opened 2022-09-19
[dalgona] Introduce Dalgona class

jinevening:
This introduces Dalgona class, which invokes luci-interpreter.

---
Related to: https://github.com/Samsung/ONE/pull/3501

### DIFF
--- a/compiler/dalgona/src/Dalgona.cpp
+++ b/compiler/dalgona/src/Dalgona.cpp
@@ -1,0 +1,214 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Dalgona.h"
+
+#include <luci/Importer.h>
+#include <foder/FileLoader.h>
+#include <dio_hdf5/HDF5Importer.h>
+
+#include <iostream>
+#include <random>
+
+using Shape = std::vector<loco::Dimension>;
+using DataType = loco::DataType;
+
+namespace
+{
+
+uint32_t numElements(const luci::CircleNode *node)
+{
+  uint32_t num_elements = 1;
+  for (uint32_t i = 0; i < node->rank(); i++)
+    num_elements *= node->dim(i).value();
+
+  return num_elements;
+}
+
+// Return tensor's size in bytes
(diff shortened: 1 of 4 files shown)


jinevening:
For reviewers:

Most of the codes are the same with `record-minmax`. A major difference will be Observer of Interpreter.

seanshpark (approved):
LGTM

compiler/dalgona/src/Dalgona.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+using Shape = std::vector<loco::Dimension>;
+using DataType = loco::DataType;
+
+namespace
+{
+
+uint32_t numElements(const luci::CircleNode *node)
+{
+  uint32_t num_elements = 1;
```
mhs4670go, line 34: How about add `assert` or throw an error for the case the `node` is `nullptr`?

compiler/dalgona/src/Dalgona.cpp
```diff
@@ -0,0 +1,214 @@
...
+  for (uint32_t i = 0; i < node->rank(); i++)
+    num_elements *= node->dim(i).value();
+
+  return num_elements;
+}
+
+// Return tensor's size in bytes
+template <typename NodeT> size_t getByteSize(const NodeT *node)
```
mhs4670go, line 42: ditto.

compiler/dalgona/src/Dalgona.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+  uint32_t dtype_size = loco::size(node->dtype());
+  return dtype_size * numElements(node);
+}
+
+// Throw exception if input has one of the following conditions.
+// 1. Have unknown dimension
+// 2. Number of elements is 0
+void checkInputDimension(const luci::CircleInput *input)
```
mhs4670go, line 51: ditto

compiler/dalgona/src/Dalgona.cpp
```diff
@@ -0,0 +1,214 @@
...
+          importer.readTensor(record_idx, input_idx, &dtype, &shape, input_data.data());
+
+          // Check the type and the shape of the input data is valid
+          verifyTypeShape(input_node, dtype, shape);
+        }
+        else
+        {
+          // Skip type/shape check for raw data
+          importer.readTensor(record_idx, input_idx, input_data.data());
+        }
```
mhs4670go, line 159: How about use positive condition for better readability?
```suggestion
        if (is_raw_data)
        {
          // Skip type/shape check for raw data
          importer.readTensor(record_idx, input_idx, input_data.data());
        }
        else
        {
          DataType dtype;
          Shape shape;
          importer.readTensor(record_idx, input_idx, &dtype, &shape, input_data.data());

          // Check the type and the shape of the input data is valid
          verifyTypeShape(input_node, dtype, shape);
        }
```

mhs4670go:
(suggestion for future) Current design forces to write each methods according to the format of input data file - HDF5(`runAnalysisWithH5Input`), RANDOM(`runAnalysisWithRandomInput`). If we split apart the section of loading data, it would have better extensibility like `circle-eval-diff` does.

mhs4670go:
(optional) It would be better to add unittest for methods in unnamed namespace later - `numElements`, `getByteSize`, `checkInputDimension`, `verifyTypeShape`.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #9713, opened 2022-09-15
[onert] Enable multi-model package compilation with controlflow


hseok-oh:
### What?

Let's introduce `ExecutorIndex` for `Executor`.

### Why?

Now `Executors` uses `SubgraphIndex` to maintain `Executor`'s `unordered_map`. If we use multi-model package compilation and model includes controlflow operation (`if`, `while`), we cannot use `SubgraphIndex` for `Executor` because subgraph index will be duplicated on different subgraph.

hseok-oh:
Draft: #9716

hseok-oh:
To enable compilation of multiple model with controlflow, we decide to use `ModelIndex`-`SubgraphIndex` pair as key on `Executor` map.

hseok-oh:
Done

PR #9740, opened 2022-09-19
[onert] Introduce ExecutorIndex

hseok-oh:
This commit introduces ExecutorIndex in Executors.h and use as key on Executor's map. 
To maintain mapping between model's subgraph index and executor index, introduce index map. 
Index map will use model index with subgraph index later.

---

Related issue: #9713
Draft: https://github.com/Samsung/ONE/pull/9716

### DIFF
--- a/runtime/onert/core/include/exec/Executors.h
+++ b/runtime/onert/core/include/exec/Executors.h
@@ -19,11 +19,15 @@
 
 #include "IExecutor.h"
 #include "ir/NNPkg.h"
+#include "util/Index.h"
 
 namespace onert
 {
 namespace exec
 {
+
+struct ExecutorIndexTag;
+using ExecutorIndex = ::onert::util::Index<uint32_t, ExecutorIndexTag>;
 
 /**
  * @brief Class to gather executors
@@ -36,13 +40,11 @@ public:
   Executors(const Executors &) = delete;
   Executors(Executors &&) = default;
 
-  // TODO Use Executor index
-  void emplace(ir::SubgraphIndex idx, std::unique_ptr<IExecutor> exec)
-  {
-    _executors.emplace(idx, std::move(exec));
-  }
+  void emplace(const ir::SubgraphIndex &idx, std::unique_ptr<IExecutor> exec);
 
-  IExecutor *at(ir::SubgraphIndex idx) const { return _executors.at(idx).get(); }
+  IExecutor *at(const ExecutorIndex &idx) const { return _executors.at(idx).get(); }
+
+  IExecutor *at(const ir::SubgraphIndex &idx) const;
 
   uint32_t inputSize() const;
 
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/exec/Executors.cc
```diff
@@ -41,13 +64,13 @@ const ir::OperandInfo Executors::inputInfo(const ir::IOIndex &index)
     // TODO handle general case
     const auto desc = _model_edges->pkg_inputs[index.value()];
     const auto model_idx = std::get<0>(desc);
-    const auto executor_idx = ir::SubgraphIndex{model_idx.value()};
+    const auto executor_idx = ExecutorIndex{model_idx.value()};
```
glistening, line 67: @hseok-oh Does it assume executor order is same to model order? That is, executor 1 comes from model 1's subgraph 1. executor 2 comes from model 2's subgraph, ... ?

runtime/onert/core/include/exec/Executors.h
```diff
@@ -36,13 +40,11 @@ class Executors
...
   Executors(Executors &&) = default;
 
-  // TODO Use Executor index
-  void emplace(ir::SubgraphIndex idx, std::unique_ptr<IExecutor> exec)
-  {
-    _executors.emplace(idx, std::move(exec));
-  }
+  void emplace(const ir::SubgraphIndex &idx, std::unique_ptr<IExecutor> exec);
+
+  IExecutor *at(const ExecutorIndex &idx) const { return _executors.at(idx).get(); }
```
glistening, line 45: @hseok-oh I thought more. Is it possible not to expose `ExecutorIndex` at all? 

**Interface**

`(ModelIndex, SubgraphIndex)`  -> `IExecutor *`
```suggestion
  IExecutor *at(std::pair<ir::ModelIndex, ir::SubgraphIndex>) const;
```

runtime/onert/core/include/exec/Executors.h
```diff
@@ -57,12 +59,27 @@ class Executors
...
+    else
+      return ExecutorIndex{};
+  }
+
 private:
-  // TODO Use Executor index
-  //      Changing index will effect if/while compile and kernel implementation
-  std::unordered_map<ir::SubgraphIndex, std::unique_ptr<IExecutor>> _executors;
+  // NOTE _executors[0] is primary executor
+  std::unordered_map<ExecutorIndex, std::unique_ptr<IExecutor>> _executors;
```
glistening, line 77:
```suggestion
  std::unordered_map<std::pair<ir::ModelIndex, ir::SubgraphIndex>, std::unique_ptr<IExecutor>> _executors;
```

runtime/onert/core/include/exec/Executors.h
glistening: Or you may use two parameters instead of `std::pair`.

hseok-oh:
Draft is updated and `ExecutorIndex` will not be used

PR #9728, opened 2022-09-19
[infra] Install Python dev libraries

jinevening:
This installs python dev libraries on CI workers.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/pull/3501

PR #9743, opened 2022-09-20
[dalgona] Add Python hooks

jinevening:
This adds python hooks to Dalgona.

---
Related to: https://github.com/Samsung/ONE/pull/3501

### DIFF
--- a/compiler/dalgona/CMakeLists.txt
+++ b/compiler/dalgona/CMakeLists.txt
@@ -1,10 +1,33 @@
+find_package(PythonInterp 3.8 QUIET)
+find_package(PythonLibs 3.8 QUIET)
+
+if(NOT ${PYTHONINTERP_FOUND})
+  message(STATUS "Build dalgona: FAILED (Python3 is missing)")
+  return()
+endif()
+
+if(${PYTHON_VERSION_MINOR} LESS 8)
+  message(STATUS "Build dalgona: FAILED (Install Python version higher than or equal to 3.8)")
+  return()
+endif()
+
+nnas_find_package(Pybind11)
+if(NOT Pybind11_FOUND)
+  message(STATUS "Build dalgona: FAILED (Pybind11 is missing)")
+  return()
+endif(NOT Pybind11_FOUND)
+
 set(DRIVER "driver/Driver.cpp")
 
 file(GLOB_RECURSE SOURCES "src/*.cpp")
 
 add_executable(dalgona ${DRIVER} ${SOURCES})
 target_include_directories(dalgona PRIVATE include)
+target_include_directories(dalgona PRIVATE ${PYTHON_INCLUDE_DIRS})
+target_include_directories(dalgona PRIVATE ${Pybind11_INCLUDE_DIRS})
 
+target_link_libraries(dalgona INTERFACE pybind11::embed)
+target_link_libraries(dalgona PRIVATE ${PYTHON_LIBRARIES})
 target_link_libraries(dalgona PRIVATE arser)
 target_link_libraries(dalgona PRIVATE safemain)
 target_link_libraries(dalgona PRIVATE foder)
--- a/compiler/dalgona/include/PythonHooks.h
+++ b/compiler/dalgona/include/PythonHooks.h
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __DALGONA_PYTHON_HOOKS_H__
+#define __DALGONA_PYTHON_HOOKS_H__
+
+#include <loco/IR/Graph.h>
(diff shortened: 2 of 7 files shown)


compiler/dalgona/CMakeLists.txt
```diff
@@ -1,10 +1,33 @@
+find_package(PythonInterp 3.8 QUIET)
+find_package(PythonLibs 3.8 QUIET)
+
+if(NOT ${PYTHONINTERP_FOUND})
+  message(STATUS "Build dalgona: FAILED (Python3 is missing)")
+  return()
+endif()
+
+if(${PYTHON_VERSION_MINOR} LESS 8)
```
seanshpark, line 9: Our development platform starts from Ubuntu 18.04 which has python 3.6 as default.
This requires to install 3.8 as default or somehow use it with alternatives.
Can we start from 3.6?

jinevening: The main reason I used 3.8 is to match with our `venv` environment, which is used for tests.

I guess that lowering the bar into 3.6 would be ok. I'll test it in the full draft PR to make sure it works in the CI worker.

seanshpark: > The main reason I used 3.8 is to match with our venv environment, which is used for tests.

The viewpoint of mine is that the `venv` is execution environment and this check is of build environment. Can build with 3.6 and run with 3.8 be possible and valid?

jinevening: > Can build with 3.6 and run with 3.8 be possible and valid?

I've locally tested it, but it was impossible. Python library (ex: libpython3.6m.so.1.0 or libpython3.8.so.1.0) linked during the build time should match with execution environment.

`common-artifacts` already requires python 3.8 (even for Ubuntu 18.04 environment), so I think it would be better to use 3.8 in `dalgona` too.

compiler/dalgona/CMakeLists.txt
```diff
@@ -1,10 +1,33 @@
+find_package(PythonInterp 3.8 QUIET)
+find_package(PythonLibs 3.8 QUIET)
```
seanshpark, line 2: We need to update documents, docker, test environment to include these two packages if not already installed. I hope you to update them :)

seanshpark (approved):
LGTM

compiler/dalgona/CMakeLists.txt
jinevening: Python 3.8 was already used by common-artifacts. From CMakeLists.txt in common-artifacts

```
find_package(PythonInterp 3.8 QUIET)
find_package(PythonLibs 3.8 QUIET)

if(NOT ${PYTHONINTERP_FOUND})
  message(STATUS "Build common-artifacts: FALSE (Python3 is missing)")
  return()
endif()

if(${PYTHON_VERSION_MINOR} LESS 8)
  message(STATUS "Build common-artifacts: FALSE (You need to install Python version higher than 3.8)")
  return()
endif()
```

Actually I copy-and-pasted those codes :)

So I guess that the change should be limited to dalgona. I uploaded PR to update README (https://github.com/Samsung/ONE/pull/9744). Docker file for dalgona was already updated (https://github.com/Samsung/ONE/pull/9728). Tests for dalgona will be based on Python 3.8.

seanshpark: Ah, thanks! I mis-read the package in common-artifacts :)

compiler/dalgona/src/Dalgona.cpp
```diff
@@ -114,14 +119,17 @@ void Dalgona::initialize(const std::string &input_model_path)
...
+
+  _interpreter->attachObserver(_hooks.get());
 }
 
 void Dalgona::runAnalysisWithH5Input(const std::string &input_data_path,
                                      const std::string &analysis_path,
                                      const std::string &analysis_args)
 {
-  // TODO Import analysis code
+  py::object scope = py::module::import("__main__").attr("__dict__");
```
mhs4670go, line 131: I've just seen the pybind document at https://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html#eval.

The example codes look like below.
```cxx
// Evaluate in scope of main module
py::object scope = py::module_::import("__main__").attr("__dict__");
```
Where `module` keyword here has extra underscore.

Is there any difference between `py::module` and `py::module_`?

compiler/dalgona/include/PythonHooks.h
```diff
@@ -0,0 +1,62 @@
...
+class PythonHooks : public luci_interpreter::ExecutionObserver
+{
+public:
+  PythonHooks(luci_interpreter::Interpreter *interpreter) : _interpreter(interpreter)
+  {
+    // Do nothing
+  }
+
+  // Called when the analysis starts
+  void importAnalysis(const std::string &analysis_path, py::object &globals,
```
mhs4670go, line 41: Looks like `globals` parameter always give the value `py::module::import("__main__").attr("__dict__")`, which menas it can be initailized in constructor and removed from `importAnalysis` spec. Is there any other case?

compiler/dalgona/src/Dalgona.cpp
jinevening: Thanks for careful reading! I didn't even know there was module_.

I've found the history from pybind11 doc.
image

https://github.com/pybind/pybind11/pull/2489

module_ was introduced to avoid conflict with module keyword in cpp. In dalgona, module is used with namespace, so such a conflict will not occur :)

compiler/dalgona/include/PythonHooks.h
jinevening: I defined `globals` object outside of importAnalysis to keep the object alive while analysis is running.

mhs4670go: Ah, I mean that `globals` can be `PythonHooks`'s attribute. Anyways looks good.

mhs4670go (commented):
(optional) It would be better to write some tests for 'utils' lib later.

mhs4670go (approved):
LGTM

PR #9762, opened 2022-09-22
[onert] Introduce entryExecutor

hseok-oh:
This commit introduce entryExecutor to abstract 1st running executor.
Method executeEntries is renamed to executeModels to avoid confusion.

---

Draft: #9716

### DIFF
--- a/runtime/onert/core/include/exec/Execution.h
+++ b/runtime/onert/core/include/exec/Execution.h
@@ -53,9 +53,9 @@ public:
    * @brief   Returns primary graph object
    * @return  Graph object
    */
-  const ir::Graph &primary_subgraph() const { return primary_executor()->graph(); }
+  const ir::Graph &primary_subgraph() const { return entryExecutor()->graph(); }
 
-  const ir::Graph &primary_parentgraph() const { return primary_executor()->parent_graph(); }
+  const ir::Graph &primary_parentgraph() const { return entryExecutor()->parent_graph(); }
   /**
    * @brief     Change input shape
    * @param[in] index   Input index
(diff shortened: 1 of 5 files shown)


glistening (approved):
LGTM

glistening:
@hseok-oh 

**Q.** What is the primary executor in case of multiple models, especially there are multiple inputs which feed into different models?

Related to https://github.com/Samsung/ONE/pull/9716#discussion_r977173435

hseok-oh:
> @hseok-oh What is the primary executor in case of multiple models, especially there are multiple inputs which feed into different models? [#9716 (comment)](https://github.com/Samsung/ONE/pull/9716#discussion_r977173435)

I think multiple models' primary executor is executed executor at first. (It may be better to rename `entryExecutor`)
I assume that if package has multiple models, 1st model has no depedency with other models (1st model's inputs are all comes from outside, not from other models). Then 1st model's 1st subgraph executor can become prime executor.

glistening:
> I think multiple models' primary executor is executed executor at first. (It may be better to rename `entryExecutor`) I assume that if package has multiple models, 1st model has no depedency with other models (1st model's inputs are all comes from outside, not from other models). Then 1st model's 1st subgraph executor can become prime executor.

Okay. If there are multiple independent executors, the executor of the model which preceeds others in nnpackage manifest's `models` would be primaryExecutor. Also, `entryExecutor` seems more intuiteve than `primaryExecutor`.

hseok-oh:
Renamed to `entryExecutor`

glistening (approved):
LGTM

chunseoklee (approved):

PR #9768, opened 2022-09-22
[Draft][npud] Introduce D-Bus IPC system

jyoungyun:
/cc @chunseoklee @hseok-oh


runtime/service/npud/core/org.tizen.npud.xml
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
```
chunseoklee, line 1: org.tizen.* files are for tizen's platform service, right ?
Is there any documentation for those spec ?

jyoungyun: This spec is came from dbus guideline. This spec is not related to Tizen, but I think we should think about the service name, `org.tizen.npud`.
https://dbus.freedesktop.org/doc/dbus-api-design.html

hseok-oh:
@nnfw-bot test all

jyoungyun:
### How to test

Run `npud`
```bash
$ NPUD_LOG_ENABLE=1 DEVICE_MODULE_PATH=${PWD}/Product/x86_64-linux.debug/out/lib ./Product/x86_64-linux.debug/out/bin/npud
```

Send dbus message
```bash
$ dbus-send \
  --system \
  --print-reply \
  --dest=org.tizen.npud \
  /org/tizen/npud \
  org.tizen.npud.core.device_get_available_list
```

Method lists: https://github.com/Samsung/ONE/blob/57a041278b91833f132a8a7f7d26f5fb7f1d1b51/runtime/service/npud/org.tizen.npud.xml

jyoungyun:
The DBUS IPC implementation merged. This commit has many changes related to the DBus method. However, this is a big difference from the main code, so I will upload it to the new draft code.

PR #9788, opened 2022-09-23
[luci-micro] Remove exceptions

BalyshevArtem:
This PR removes exceptions from luci-micro to reduce binary file size.

issue  #9468


compiler/luci-micro/helpers/GenerateKernelsListHelper.cpp
```diff
@@ -157,7 +157,7 @@ std::vector<char> loadFile(const std::string &path)
   if (!file.good())
   {
     std::string errmsg = "Failed to open file: " + path;
-    throw std::runtime_error(errmsg.c_str());
+    assert(false && errmsg.c_str());
```
binarman, line 160: Could you rewrite these lines, like it is done below?
`assert` will not print error message as intended.

binarman (approved):
LGTM, but it seems that this PR introduces two separate things:
- fixes build of circle schema generated files
- removes exceptions

Could you move cmake changes in separate PR?

BalyshevArtem:
> Could you move cmake changes in separate PR?

Done, see #9827

binarman (approved):

PR #9791, opened 2022-09-26
[onert] Use ModelIndex on executor map

hseok-oh:
This commit updates executor map in Executors to use ModelIndex and SubgraphIndex pair as key.

---

Related issue: https://github.com/Samsung/ONE/issues/9713
Draft: https://github.com/Samsung/ONE/pull/9716

### DIFF
--- a/runtime/onert/core/include/exec/Executors.h
+++ b/runtime/onert/core/include/exec/Executors.h
@@ -20,6 +20,21 @@
 #include "IExecutor.h"
 #include "ir/NNPkg.h"
 
+namespace std
+{
+
+template <> struct hash<std::pair<::onert::ir::ModelIndex, ::onert::ir::SubgraphIndex>>
+{
+  size_t
+  operator()(const std::pair<::onert::ir::ModelIndex, ::onert::ir::SubgraphIndex> &pair) const
+    noexcept
+  {
+    return hash<uint64_t>()((uint64_t(pair.first.value()) << 32) | uint64_t(pair.second.value()));
+  }
+};
+
+} // namespace std
+
 namespace onert
 {
 namespace exec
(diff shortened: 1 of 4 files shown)


runtime/onert/core/include/exec/Executors.h
```diff
@@ -20,6 +20,21 @@
...
+namespace std
+{
+
+template <> struct hash<std::pair<::onert::ir::ModelIndex, ::onert::ir::SubgraphIndex>>
+{
+  size_t
+  operator()(const std::pair<::onert::ir::ModelIndex, ::onert::ir::SubgraphIndex> &pair) const
+    noexcept
+  {
+    return hash<uint64_t>()((uint64_t(pair.first.value()) << 32) | uint64_t(pair.second.value()));
```
glistening, line 32: (optional) I didn't consider arm32 when I suggested 64bit hash. We may use 32bit version. ( But I don't think it matters. )

```suggestion
    return hash<uint32_t>()((pair.first.value() << 16) | pair.second.value());
```

runtime/onert/core/include/exec/Executors.h
```diff
@@ -37,14 +52,16 @@ class Executors
...
+
+  // Temporary method for builtin backend migration
+  // TODO Remove this method
+  IExecutor *at(ir::SubgraphIndex idx) const { return at(ir::ModelIndex{0}, idx); }
 
-  IExecutor *at(ir::SubgraphIndex idx) const { return _executors.at(idx).get(); }
+  IExecutor *at(const ir::ModelIndex &model_index, const ir::SubgraphIndex &subg_index) const;
 
-  IExecutor *entryExecutor() const { return at(ir::SubgraphIndex{0}); }
+  IExecutor *entryExecutor() const { return at(ir::ModelIndex{0}, ir::SubgraphIndex{0}); }
```
glistening, line 64: @hseok-oh I didn't fully understand and notice your assumption you wrote in https://github.com/Samsung/ONE/pull/9762#issuecomment-1254543579. At first, I understood {model, subg} of pkg_input[0] would be entryExecutor while you assume that the subgraph {model_idx, subg_idx} = {0, 0} always has pkg input. We may need to update nnpkg specification and add a validation check whether pkg_input contains {0, 0}. I am okay about this assumption.

glistening (approved):
LGTM

runtime/onert/core/include/exec/Executors.h
hseok-oh: I'll make PR to change `ModelIndex` and `SubgraphIndex` value type to `uint16_t` later.

ragmani: How about use XOR operator?
```suggestion
    auto hash1 = hash<uint32_t>{}(pair.first.value());
    auto hash2 = hash<uint32_t>{}(pair.second.value());
 
    if (hash1 != hash2) {
      return hash1 ^ hash2;             
    }
         
    // If hash1 == hash2, their XOR is zero.
    return hash1;
```

ragmani (approved):

runtime/onert/core/include/exec/Executors.h
hseok-oh: Let's update this when we change value type.

hseok-oh: Hmm. Is it safe to use hash value shift-xor?

ragmani: If both(hash1 and shifted hash2) are not the same, it is safe.

ragmani: And If hash2 >= 2^16, it is not safe?

PR #9803, opened 2022-09-27
[luci] Quantize fp32 nodes only

jinevening:
This quantizes fp32 nodes only.

### DIFF
--- a/compiler/luci/pass/src/QuantizeActivation.cpp
+++ b/compiler/luci/pass/src/QuantizeActivation.cpp
@@ -44,12 +44,8 @@ void QuantizeActivation::visit(luci::CircleNode *node)
   LOGGER(l);
   INFO(l) << "QuantizeActivation visit node: " << node->name() << std::endl;
 
-  // Check if this is already quantized
-  if (is_quantized(node))
-    return;
-
-  // Check if this is bool type (bool type is not quantized)
-  if (node->dtype() == loco::DataType::BOOL)
+  // Check if node is fp32
+  if (not is_fp32(node))
     return;
 
   // Check if this is const (const activation is handled by QuantizeConstInputActivation)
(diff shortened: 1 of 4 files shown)


compiler/luci/pass/src/QuantizeActivation.cpp
```diff
@@ -185,7 +181,7 @@ void QuantizeConstInputActivation::visit(luci::CircleNode *node)
   {                                                             \
     auto input = node->INPUT_NAME();                            \
     auto const_node = dynamic_cast<luci::CircleConst *>(input); \
-    if (const_node && !is_quantized(const_node))                \
+    if (const_node && is_fp32(const_node))                      \
```
jinevening, line 184: To reviewers:

Previous codes attempted to quantize non-fp32 nodes, which can raise an error. This PR resolves it.

seanshpark (approved):

compiler/luci/pass/src/QuantizedModelVerifier.test.cpp
```diff
@@ -104,10 +106,36 @@ void insert_scale_zp(luci::CircleNode *node, float scale, int64_t zp)
   qparam->zerop.push_back(zp);
 }
 
+void run_phase(loco::Graph *g, Type quantized_dtype, Granularity granularity)
```
jinevening, line 109: Test codes did not correctly set dtypes of nodes, so I use `Phase` including `CircleTypeInferencePass` for setting them at once.

Previously, only `QuantizeWithMinMaxPass` was invoked.

seanshpark: > Test codes did not correctly set dtypes of nodes

Q) plan to fix test codes?
Q) if fix test codes, can we remove this?
If either is not yes, can you add a NOTE about this situation?

jinevening: Ah, this PR applies fix :)

PR #4324, opened 2020-09-17
[onert] add transposed flag into Conv kernel

dayo09:
- add explicit transposed flag
- not to use _kernel->data if the data is transposed and freed
- to solve assert in #4319

ONE-DCO-1.0-Signed-off-by: dayo09 <dayg502@gmail.com>

PR #9808, opened 2022-09-28
[onert] Remove unncessary basic::Tensor's reference count functions

glistening:
No one uses reset_ref() and num_references().

Related: #4319, #4324


chunseoklee:
AFAIK, they are used in internal BCQ backend.

glistening:
I confirmed it is used in BCQ.

Issue #9750, opened 2022-09-21
[ONE-Dockerizing] debian package related to onecc-docker


Seunghui98:
## What
We created a onecc-docker debian package.

## Why
We have to make it a Debian package to run the onecc-docker.

## How
1. Structure
```
$ tree
.
├── DEBIAN
│   ├── control
│   ├── install
│   └── preinst
└── usr
    ├── bin
    │   └── onecc-docker
    └── share
        └── one
            └── docker
                ├── Dockerfile
                └── copyright
```
- `usr/bin/onecc-docker`, `usr/share/one/docker/Dockerfile`
  - Related issues : [#9738](https://github.com/Samsung/ONE/issues/9738)
  - Parent issues : [#9712](https://github.com/Samsung/ONE/issues/9712)
  - Related PR : [#9737](https://github.com/Samsung/ONE/pull/9737)
- The Dockerfile path has been set so that onecc-docker file can use Dockerfile in `usr/share/one/docker`

2. Test
1) Create onecc-docker debian package
```
$ sudo dpkg -b onecc-docker_1.21.0_amd64
```

2) Install the onecc-docker debian package
```
$ sudo apt-get install -y ./onecc-docker_1.21.0_amd64.deb
```

3) The result of executing a simple onecc command to see if onecc-docker works well
```
$ onecc-docker --version
build Docker image ...
Dockerfile successfully built.
onecc version 1.21.0
Copyright (c) 2020-2022 Samsung Electronics Co., Ltd. All Rights Reserved
Licensed under the Apache License, Version 2.0
https://github.com/Samsung/ONE
```

/cc @Samsung/ootpg_docker

seanshpark:
the main comment has similar contents with the link and the parent links.
Q) why add a new issue? I mean why do you put so much efforts to write this issue?

Seunghui98:
@seanshpark 
I uploaded an issue to include the related content of draft code of onecc-docker debian package. Would selling similar issues on one topic cause confusion for others?

seanshpark:
> Would selling similar issues on one topic cause confusion for others?

I think think it's about the confusion with multiple issues posted.
But as I wrote, IMHO, (1) it seems you are putting much effort for writing this issue. Is it really necessary? 
(2) there are lots of subjects in the main comment. what is the main point of this issue?
(3) I think for explaining the background and benefits and the task just one issue is OK. but I don't think posting all issues like this is good.

jyoungyun:
@Seunghui98

[...]

PR #9815, opened 2022-09-29
[onecc-docker] onecc-docker permission error fix

Seunghui98:
This PR corrects a permission error when using onecc-docker after installing onecc-docker debian package.

ONE-DCO-1.0.-Signed-off-by: seunghui-lee hithere1012@naver.com

Related Issue : [#9750]

### DIFF
--- a/compiler/onecc-docker/debian/rules
+++ b/compiler/onecc-docker/debian/rules
@@ -3,3 +3,6 @@
 %:
 	dh $@
 
+override_dh_fixperms:
+	dh_fixperms
+	chmod +x debian/onecc-docker/usr/share/one/bin/onecc-docker


compiler/onecc-docker/debian/postinst
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html
+# Boradly speaking, the `postinst` is called after a package is unpacked.
+
+set -e
+
+sudo chmod +x /usr/share/one/bin/onecc-docker
```
jyoungyun, line 8: Um, I suggest to use `dh_fixperms` function in rules file.
https://manpages.debian.org/testing/debhelper/dh_fixperms.1.en.html

ex)
```
override_dh_fixperms:                                                                               
  dh_fixperms                                                                                       
  chmod +x <files>
```

Seunghui98: @jyoungyun 
Oh, thank you so much for the code review. Thanks to you, I learned that it is possible to set permissions with debhelper. I did a simple action test after modifying the code, and I checked that the authority was well granted.

compiler/onecc-docker/debian/rules
```diff
@@ -2,4 +2,8 @@
 
 %:
 	dh $@
-
+	
+overrid_dh_fixperms:
```
jyoungyun, line 6:
```suggestion
override_dh_fixperms:
```

Is the `onecc-docker` file in `/usr/share/one/bin/` when this function is executed?

Seunghui98: @jyoungyun 

Oh, thank you for your advice. As I looked it up, the permission adjustment to a occurs when the package is installed, so I set it as path `debian/onecc-docker/usr/share/one/bin/onecc-docker`. The path of onecc-docker file was set considering the entry path of the Debian package. I tested it after that, but I give permission to onecc-docker well, so can you check if the code is okay?

jyoungyun (approved):
LGTM :)

mhs4670go (approved):
LGTM

PR #9818, opened 2022-09-29
[dalgona] Add template for analysis script

jinevening:
This adds template for analysis script.

---
Draft PR: https://github.com/Samsung/ONE/pull/3501

### DIFF
--- a/compiler/dalgona/analysis/AnalysisTemplate.py
+++ b/compiler/dalgona/analysis/AnalysisTemplate.py
@@ -1,0 +1,125 @@
+# Copyright 2022 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


compiler/dalgona/analysis/AnalysisTemplate.py
```diff
@@ -0,0 +1,125 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""
+A template for anlaysis code.
+This template shows how to access the information of each operator inside hooks.
+Users can write their own hooks by modifying this file.
```
seanshpark, line 18: I assume this file will be installed... inside `/usr/share/one/bin/...`
That is will need sudo to edit.
Q) Where is this file going to be installed?

seanshpark: Ah, maybe "Users can write their own hooks by copy and modifing this file" ?

jinevening: This file is like a document (manual), not an executed code. So it will not be installed in the `bin` directory.

Only one executable `dalgona` will be installed in `/usr/share/one/bin/`.

seanshpark (approved):
LGTM

seanshpark:
Oh no... -_-;

seanshpark:
@jinevening , @mhs4670go , sorry to landed with not review from @mhs4670go 😭

jinevening:
@mhs4670go is generous, so it would be ok :)  @mhs4670go Please tell me if there is something to correct.

PR #9667, opened 2022-09-02
Support batch-parallel execution of synchronously called trix models.

ragmani:

Issue #9207, opened 2022-05-30
Multicore execution Milestone1 


chunseoklee:
- Milestone1 : Run Batch request in parallel manner via direct call to trix-engine(~Tizen M2, Aug 30th)

## What we need ?
- User scenario
  + User will use nnfw api to inference 1 batch (e.g. mobilenet's tvn binary) model as usual
  + onert core will do tvn BATCH execution if tvn is 1-batch model and user's input shape is multi-batch
    - Q. how to determine multi-batch input w.r.t. nnfw api ?
      + A. nnfw_set_input_tensorinfo(session, input_index, ti);
  + Note that CPU backend supports batched inference(though each kernel in the model should support batched inference)
- Unified API of NPU Daemon
  + Support multi NPU ?

## How batch execution works when running tflite model in onert ?
- Suppose that the models's input shape is [1,244,244,3]
- Then, user will request an inference with input shape of [4,244,244,3] as follows

``` cpp
  nnfw_session *session = nullptr;
  nnfw_create_session(&session);

  // Loading nnpackage
  nnfw_load_model_from_file(session, path_to_nnpkg);

  // input shape is modified
  nnfw_tensorinfo ti;
  ti.rank = 4;
  ti.dims = {4,244,244,3}; // syntax abuse
  nnfw_set_input_tensorinfo(session, input_index, ti);

  // compile model
  nnfw_prepare(session);

  // Prepare input. Here we just allocate dummy input arrays.
  std::vector<float> input;
  nnfw_input_tensorinfo(session, 0, &ti); // get first input's info
  uint32_t input_elements = num_elems(&ti);
  input.resize(input_elements);
  // TODO: Please add initialization for your input.
  nnfw_set_input(session, 0, ti.dtype, input.data(), sizeof(float) * input_elements);

  // Prepare output
  std::vector<float> output;
  nnfw_output_tensorinfo(session, 0, &ti); // get first output's info
  uint32_t output_elements = num_elems(&ti);
  output.resize(output_elements);
  nnfw_set_output(session, 0, ti.dtype, output.data(), sizeof(float) * output_elements);

  // Do inference
  nnfw_run(session);
```

chunseoklee:
@ragmani FYI, 

how to check dynamic shape infer on nnpackage_run: 

```
[...]

PR #9820, opened 2022-09-29
[onert] Add handling of errors that occur within threads of `BatchThreadPool`

ragmani:
For issue https://github.com/Samsung/ONE/issues/9207
Draft PR https://github.com/Samsung/ONE/pull/9667

This commit adds handling of errors that occur within threads of `BatchThreadPool`.

### DIFF
--- a/runtime/onert/backend/trix/DevContext.cc
+++ b/runtime/onert/backend/trix/DevContext.cc
@@ -18,6 +18,8 @@
 
 #include "Convert.h"
 
+#include <stdexcept>
+
 namespace onert
 {
 namespace backend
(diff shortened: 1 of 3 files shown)


runtime/onert/backend/trix/DevContext.cc
```diff
@@ -137,60 +139,64 @@ void DevContext::requestRun(ModelID model_id, input_buffers *input_bufs, tensors
...
+
+          try
+          {
+            // dev_num is the same as the thread number in _batch_thread_pool
+            this->runOneBatch(dev_num, model_id, &in_batch_buffers, in_info, &out_batch_buffers,
+                              out_info);
+          }
+          catch (...)
+          {
+            _eptr = std::current_exception();
```
jyoungyun, line 192: (Q) Are you trying to see only the last exception that occurred?

runtime/onert/backend/trix/DevContext.cc
```diff
@@ -137,60 +139,64 @@ void DevContext::requestRun(ModelID model_id, input_buffers *input_bufs, tensors
...
+          }
+
           return batch_num;
         },
-        model_id, in_batch_buffers, &in_infos[batch_num], out_batch_buffers, &out_infos[batch_num],
+        model_id, input_bufs, output_bufs,
+        // model_id, in_batch_buffers, &in_infos[batch_num], out_batch_buffers,
+        // &out_infos[batch_num],
```
chunseoklee, line 199:
```suggestion
```

runtime/onert/backend/trix/DevContext.cc
```diff
@@ -225,30 +238,57 @@ void DevContext::runOneBatch(uint32_t dev_num, ModelID model_id, input_buffers *
...
+        setNPU_requestData(dev_handle, req_id, input_bufs, in_info, output_bufs, out_info))
   {
     removeNPU_request(dev_handle, req_id);
     throw std::runtime_error("Unable to create NPU request for model id (" +
-                             std::to_string(model_id_at_device) + ")");
+                             std::to_string(model_id_at_device) + ")" +
+                             " error code : " + std::to_string(error_code));
+  }
+
+  // NOTE submitNPU_request is not thread-safe(?). It is rarely hanging(unresponsive).
```
chunseoklee, line 258: (optional) need to ask to trix-engine developers

runtime/onert/backend/trix/DevContext.cc
```diff
@@ -225,30 +238,57 @@ void DevContext::runOneBatch(uint32_t dev_num, ModelID model_id, input_buffers *
...
+  //      Ultimately, to solve this problem, we have to either use other thread-safe API or
+  //      change submitNPU_request to be thread-safe, but both works take time.
+  //      As a workaround, let's allow hanging thread.
+  // TODO Change submitNPU_request to be thread-safe or replaced with other thread-safe API
+  std::packaged_task<int(npudev_h, int)> task(submitNPU_request);
```
chunseoklee, line 263: Q) this approach incurs two threads per 1 batch, which results in 2X number of threads compared to previous approach ? Then, it may affects performance ?

runtime/onert/backend/trix/DevContext.cc
ragmani: I only saw an explanation that it was a current exception. I think that the current exception is the same as the last exception is is a try-catch block.

runtime/onert/backend/trix/DevContext.cc
ragmani: Oops, Thanks for your pointing out. I will fix it.

runtime/onert/backend/trix/DevContext.cc
ragmani: My answer is yes. I'm not sure it is not thread-safe because I don't know the internal code, but when I tried to wrap this block with a mutex block, it worked without any problem.

runtime/onert/backend/trix/DevContext.cc
ragmani: Yes. The codes here are not intended to improve performance. They can slightly weaken performance. The codes here are just to run `submutNPU_request` on a new thread, not the current thread because of avoiding hanging on main thread.

jyoungyun (approved):
LGTM

runtime/onert/backend/trix/DevContext.cc
chunseoklee: Ok. Then, Let's check significant difference on performance.

ragmani: There is a difference, 10% ~ 20%.
It's not difference on slightly weaken performance :(
```bash
# BACKENDS=trix /usr/bin/nnfw-test/Product/out/unittest/nnpackage_run /opt/workspace/manual/mv1.q8 --shape_prepare="[0,[10,224,224,3]]" --load:raw /opt/workspace/manual/mv1.q8/input_all.tv2b --output_sizes="[0,20020]" --dump:raw /opt/workspace/manual/mv1.q8/result_parallel.tv2b -w 2 -r 100
Package Filename /opt/workspace/manual/mv1.q8
/opt/workspace/manual/mv1.q8/result_parallel.tv2b.0 is generated.
MODEL_LOAD   takes 1.540 ms
PREPARE      takes 56.435 ms
EXECUTE      takes 35.038 ms
- MEAN     :  35.038 ms
- MAX      :  63.506 ms
- MIN      :  25.003 ms
- GEOMEAN  :  34.330 ms
```

chunseoklee (approved):
Merge as it is for now. will update if better solution is found.

ragmani:
> Merge as it is for now. will update if better solution is found.

@jyoungyun Could you review this again?

hseok-oh:
@jyoungyun If there is no respond, I'll merge this PR.

jyoungyun (approved):
LGTM (I'm sorry for late response...)

PR #9767, opened 2022-09-22
[Draft][npud] Implement test framework

jyoungyun:
/cc @chunseoklee @hseok-oh

PR #9828, opened 2022-10-04
[npud] Introduce test framework

jyoungyun:
This commit introduces test framework on npud.
It tests Server and Signal classes.

draft: #9767

### DIFF
--- a/runtime/service/npud/tests/core/Server.test.cc
+++ b/runtime/service/npud/tests/core/Server.test.cc
@@ -1,0 +1,75 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "core/Server.h"
+
+#include <gtest/gtest.h>
+#include <thread>
+
+namespace
+{
+using namespace npud;
+using namespace core;
+
+//
+// ServerTest setup/teardown
+//
+class ServerTest : public ::testing::Test
+{
+protected:
+  static void runTask()
+  {
+    auto &server = Server::instance();
+    server.run();
+  }
+
+  void SetUp() override
+  {
+    std::thread child = std::thread(runTask);
+    child.detach();
+    auto &server = Server::instance();
+    while (server.isRunning() != true)
+    {
+    }
+  }
+
+  void TearDown() override
+  {
+    auto &server = Server::instance();
+    if (server.isRunning())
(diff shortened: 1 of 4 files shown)


runtime/service/npud/tests/core/Server.test.cc
```diff
@@ -0,0 +1,75 @@
...
+protected:
+  static void runTask()
+  {
+    auto &server = Server::instance();
+    server.run();
+  }
+
+  void SetUp() override
+  {
+    std::thread child = std::thread(runTask);
```
glistening, line 41: @jyoungyun Why do you create thread for every test case?

chunseoklee:
Q.  No problem in server(due to singleton) when two test suites run in parallel ? (I am not sure that two test suites run at the same time.)

runtime/service/npud/tests/core/Server.test.cc
jyoungyun: > First, I don't know how `Server` will be used in usual scenario and what is your expectation for each API.

The intention of Server test case is to check whether the server is running or not.

> I think it can be done without thread.

If not implementing using threads, the test main function is stuck because the Server holds the process until an event occurs.

jyoungyun:
> Q.  No problem in server(due to singleton) when two test suites run in parallel ? (I am not sure that two test suites run at the same time.)  

As you told, this test case should be executed in serialization. Because the server was implemented in Singleton, this is only one instance, which is why the test case did not work well. Does ONE Runtime's gtest run in parallel ?

chunseoklee:
>Does ONE Runtime's gtest run in parallel ?

I cannot find any documentation about gtest's parallelism. But, https://github.com/google/gtest-parallel implies that vanilla gtest runs in sequencial ?

runtime/service/npud/tests/core/Server.test.cc
glistening: @jyoungyun Frankly speaking, I still don't understand why it needs thread. Is it necessary just for making a test? Maybe I need to read a whole code. Please proceed without my approval.

jyoungyun: Yes, this thread just needs for test environment. To test server functions, I need to ask some commands to server from client. In this test case, the child thread is a server and the main thread is a client. I'm not sure this implement is right approach for server tests. However, when searching for various data, I wasn't find a general way to test Server, so it is implemented so that it can be tested using gtest. If there is a better way, I will revise it later. :)

jyoungyun:
> I cannot find any documentation about gtest's parallelism. But, https://github.com/google/gtest-parallel implies that vanilla gtest runs in sequencial ? 

I think so. When we need to consider the parallel test, I will update these test cases. :)

hseok-oh (approved):

runtime/service/npud/tests/core/Server.test.cc
glistening: > I wasn't find a general way to test Server, 

As I understand, npud will use d-bus and glib mainloop.

Then, I don't think testcase cover the actual case.

First, server and client will be in different processes. Client is not likely to use `Server::isRunning` (How client can get the instance out of different process?) I would add timeout source and add the codes (isRunning, raise signal, ...).

Second, I guess you will revisit testcase after d-bus related code is added based on how server and client will actually work.

chunseoklee (approved):

glistening:
If anyone is waiting for my approval, please find https://github.com/Samsung/ONE/pull/9828#discussion_r987569371.

PR #9831, opened 2022-10-05
[onert] Change SubgraphIndex value type

glistening:
It changes type for value from uint32 to uint16.
    
tflite and circle schema define Subgraphs as following:
```
  subgraphs:[SubGraph];
```
It can have 2^32 Subgraphs.
But onert assumes the number of subgraph does not exceed 2^16.
We will remove the assumption when we encounter such case.

### DIFF
--- a/runtime/onert/core/src/compiler/Compiler.cc
+++ b/runtime/onert/core/src/compiler/Compiler.cc
@@ -287,7 +287,7 @@ bool Compiler::buildPartialGraph(uint32_t num_graphs)
 
   auto partialgraphs = std::make_shared<ir::Model>();
 
-  for (uint32_t idx = 0; idx < num_graphs; idx++)
+  for (uint16_t idx = 0; idx < num_graphs; idx++)
   {
     auto partialgraph = std::make_unique<ir::Graph>();
     partialgraphs->push(ir::SubgraphIndex{idx}, std::move(partialgraph));
@@ -353,7 +353,7 @@ bool Compiler::buildPartialGraph(uint32_t num_graphs)
       assert(new_operation_index == operation_index);
     });
 
-  for (uint32_t idx = 0; idx < partial_graph->subgraphs_count(); idx++)
+  for (uint16_t idx = 0; idx < partial_graph->subgraphs_count(); idx++)
   {
     auto partition = partial_graph->at(ir::SubgraphIndex{idx});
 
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/compiler/Compiler.cc
```diff
@@ -664,23 +664,21 @@ std::vector<std::shared_ptr<CompilerArtifact>> Compiler::compile(const char *pac
...
-    }
-  }
-  else
-  {
-    throw std::runtime_error("There is no partition map file");
+        ir::SubgraphIndex(subg_idx);
   }
-
-  if (!buildPartialGraph(num_graphs))
+  uint32_t num_partial_graphs = np.asUInt();
```
glistening, line 680: I revisit some code around here. For example,

- Previous `num_graphs` is renamed to `num_partial_graphs` because partial graph and subgraph (from model file) is different concept and it makes it more readable.
- It is relocated since there is no one to use `num_graphs`.

But I will remove this change soon. It is not the scope of this PR.

hseok-oh (dismissed):
LGTM

ragmani (approved):
LGTM

glistening:
@ragmani and @hseok-oh I updated commit comment (No code change): 

2<sup>32</sup> - 1 → 2<sup>32</sup>.

hseok-oh (approved):

glistening:
Just for record

circle_schema_generated.h:
```cc
  const flatbuffers::Vector<flatbuffers::Offset<circle::SubGraph>> *subgraphs() const
  {
    return GetPointer<const flatbuffers::Vector<flatbuffers::Offset<circle::SubGraph>> *>(
      VT_SUBGRAPHS);
  }
```
flatbuffers/flatbuffers.h
```cc
template<typename T> class Vector {
 public:
  typedef VectorIterator<T, typename IndirectHelper<T>::mutable_return_type>
      iterator;
  typedef VectorIterator<T, typename IndirectHelper<T>::return_type>
      const_iterator;

  uoffset_t size() const { return EndianScalar(length_); }
```
flatbuffers/base.h
```cc
/// @cond FLATBUFFERS_INTERNAL
// Our default offset / size type, 32bit on purpose on 64bit systems.
// Also, using a consistent offset type maintains compatibility of serialized
// offset values between 32bit and 64bit systems.
typedef uint32_t uoffset_t;
```

ragmani (approved):

runtime/onert/core/src/compiler/Compiler.cc
```diff
@@ -671,8 +671,13 @@ std::vector<std::shared_ptr<CompilerArtifact>> Compiler::compile(const char *pac
     num_graphs = np.asUInt();
     for (uint32_t i = 0; i < (uint32_t)map.size(); ++i)
     {
-      options.partial_graph_options.index_to_graph[ir::OperationIndex{i}] =
-        ir::SubgraphIndex{map[i].asUInt()};
+      auto subg_idx = map[i].asUInt();
+      if (subg_idx > ir::SubgraphIndex::max())
+        throw std::runtime_error("The number of partition cannot exceed " +
+                                 std::to_string(ir::SubgraphIndex::max() + 1));
+      else
```
glistening, line 678: Just for note

`else` is not necessary. But I would like to make sure to make static analysis tool happy, which complains about the possibility of overflow.

PR #9832, opened 2022-10-05
[infra] Add dalgona-test compiler module

jinevening:
This adds dalgona-test to compiler module.


seanshpark (approved):
LGTM

PR #9840, opened 2022-10-07
[onert] Support tracing context for multiple models

glistening:
It support tracing for multiple models.
Origin is introduced to encapsulate the origin information. It is currently ir::SubgrphIndex.

### DIFF
--- a/runtime/onert/core/include/util/TracingCtx.h
+++ b/runtime/onert/core/include/util/TracingCtx.h
@@ -29,6 +29,8 @@ namespace onert
 namespace util
 {
 
+using Origin = ir::SubgraphIndex;
+
 /**
  * @brief Class to maintain information about profiling per session
  */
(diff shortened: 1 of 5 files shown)


glistening:
To support multiple model tracing, we need to update not only tracing context, but also:

### 1. ExecutionObservers interface
- runtime/onert/core/src/exec/ExecutionObservers.h
```cc
/// @brief Invoked just before model (not individual operation) execution begins
virtual void handleSubgraphBegin(ir::SubgraphIndex) { return; }

virtual void handleJobBegin(IExecutor *, ir::SubgraphIndex, ir::OperationIndex,
                            const backend::Backend *) = 0;
virtual void handleJobEnd(IExecutor *, ir::SubgraphIndex, ir::OperationIndex,
                          const backend::Backend *) = 0;

/// @brief Invoked just after model (not individual operation) execution ends
virtual void handleSubgraphEnd(ir::SubgraphIndex) { return; }
```

### 2. Update existing ExecutionObserver implementations
- Not only `TracingObserver`, but also `ProfileObserver`)
- runtime/onert/core/src/exec/ExecutionObservers.cc

### 3. `TracingObserver` specific implementations
- runtime/onert/core/src/util/EventCollector.h
- runtime/onert/core/src/util/EventRecorder.h

glistening:
### Another option

Just add new interfaces, not modify existing interface.

### 1. ExecutionObservers interface
- runtime/onert/core/src/exec/ExecutionObservers.h
```cc
virtual void handleSubgraphBegin(ir::ModelIndex, ir::SubgraphIndex) { return; }

virtual void handleJobBegin(IExecutor *, ir::ModelIndex, ir::SubgraphIndex, 
                 ir::OperationIndex, const backend::Backend *) = 0;
virtual void handleJobEnd(IExecutor *, ir::ModelIndex, 
                ir::SubgraphIndex, ir::OperationIndex, const backend::Backend *) = 0;

/// @brief Invoked just after model (not individual operation) execution ends
virtual void handleSubgraphEnd(ir::ModelIndex, ir::SubgraphIndex) { return; }
```
### 2. Update existing ExecutionObserver implementations
- Only `TracingObserver`
- runtime/onert/core/src/exec/ExecutionObservers.cc

### 3. `TracingObserver` specific implementations
- runtime/onert/core/src/util/EventCollector.h
- runtime/onert/core/src/util/EventRecorder.h

It may reduce the size of work. But it introduce more difference between each config.
It is far from what I want.

-----
It is not the main goal we have to do at this moment. I will find other approach.

PR #9841, opened 2022-10-07
[dalgona] Cast before performing arithmetic

jinevening:
This performs cast before doing arithmetic to avoid svace defect.


seanshpark:
This is to fix internal analysis issue ID 1621042

seanshpark (approved):
LGTM thank you!

Issue #9838, opened 2022-10-06
[Compiler Frontend] ARM32 build in CI


seanshpark:
I would like to enable compiler frontend ARM32 build in CI.
Purpose is not to break compiler frontend ARM32 build.
If the compile time is similar with existing debug/release then it would be better to put in PR builds.
If not, daily build is acceptable.

seanshpark:
@hseok-oh , can you please guide me what to prepare for this?

seanshpark:
Things behind this...
- some packages are getting older and we may need to upgrade them but also have pathes for ARM32 build
- some code changes may break ARM32 build

hseok-oh:
Compiler frontend build & test in CI is not using any script in `infra/scripts`. CI is using `nncc` command directly on jenkins script. You can check debug test implementation on internal CI repo's two files `scripts/nncc-config.sh` and `scripts/nncc-debug.sh`.

One simple way is using new script in `infra/scripts`. Below is `infra/scripts/docker_build_cross_arm_runtime_release.sh` with comment out for runtime build specific setting. Please make new build script base on below example with filling compiler frontend arm32 specific environment setting and build command. Then I'll decide to use new script or use command directly.

```bash
#!/bin/bash

[[ "${BASH_SOURCE[0]}" != "${0}" ]] && echo "Please don't source ${BASH_SOURCE[0]}, execute it" && return

CURRENT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
ROOT_PATH="$CURRENT_PATH/../../"

# Please enable using rootfs if compiler frontend arm32 build need rootfs
# prepare rootfs
#if [ -z "$ROOTFS_DIR" ] || [ ! -d $ROOTFS_DIR ]; then
#  echo "It will use default rootfs path"
#else
#  DOCKER_VOLUMES+=" -v $ROOTFS_DIR:/opt/rootfs"
#  DOCKER_ENV_VARS+=" -e ROOTFS_DIR=/opt/rootfs"
#fi

# mount volume (or directory) for externals
if [ -n "$EXTERNAL_VOLUME" ]; then
  DOCKER_VOLUMES+=" -v $EXTERNAL_VOLUME:/externals"
  DOCKER_ENV_VARS+=" -e EXTERNAL_VOLUME=/externals"
else
  echo "It will use default external path"
fi

# Please set environment variable if need (below is example for arm runtime build)
[...]

PR #9851, opened 2022-10-11
DRAFT check arm32 build

seanshpark:
draft to check arm32 build with PR.


seanshpark:
check #9838

Issue #9694, opened 2022-09-08
[quantizer] Visualize layer-wise qerror


jinevening:
## What

Let's visualize quantization error in the quantized model

## Why

To observe which quantized layer harms accuracy.

## How

Input: fp32 circle, q circle, test data

Steps
1. Fake quantize q circle
2. Dump intermediate FMs of fp32 model with test data
3. Dump intermediate FMs of fq model with test data
4. Compare intermediate FMs and compute qerror
5. Generate DOT graph

![image](https://user-images.githubusercontent.com/5449554/189035647-15c47c53-f239-4e35-9ed4-630e644e16f3.png)

Nodes with higher qerror will be more red.

## Sample (MobileNet V2 full uint8)

![image](https://user-images.githubusercontent.com/5449554/189035823-02a76555-858e-433f-8244-a48c1b10c036.png)

[...]

PR #9852, opened 2022-10-11
DRAFT visq viwer with nodejs

seanshpark:
on-going draft to test visq viewer with nodejs.


seanshpark:
for #9694

seanshpark:
replaced by #9862

PR #9858, opened 2022-10-12
[test/onert] Introduce onert_run

hseok-oh:
This commit introduces onert_run for simple onert test.

It enables to measure onert performance by model file (tflite, circle) without package.


tests/tools/onert_run/onert_run.cc
```diff
@@ -0,0 +1,315 @@
...
+    default:
+      throw std::runtime_error{"Invalid tensor type"};
+  }
+}
+
+static const char *default_backend_cand = "cpu";
+
+} // namespace
+
+int main(const int argc, char **argv)
```
hseok-oh, line 101: Almost same with `tflite_run` except input data setting. Input data setting is comes from `tflite_comparator`'s nnfw API call.

PR #9809, opened 2022-09-28
Draft: Support no branch multimodels nnpackage

hseok-oh:
Support multimodel NN package with no branch

ONE-DCO-1.0-Signed-off-by: Hyeongseok Oh <hseok82.oh@samsung.com>

Related issue: #9655

Issue #9655, opened 2022-09-01
[onert] Support no branch multimodels nnpackage


hseok-oh:
Support multimodel NN package with no branch 

```
inputs -> model 1 -> model 2 -> ...  -> model N-1 -> model N -> outputs
```

input(s) -> model 1 input(s)
model 1 output(s) -> model 2 input(s)
...
model N-1 output(s) -> model N input(s)
model N output(s) -> output(s)

_Originally posted by @hseok-oh in https://github.com/Samsung/ONE/issues/9646#issuecomment-1233678141_

hseok-oh:
Not scheduled issue yet

hseok-oh:
Example nnpackage

[three_tflites.tar.gz](https://github.com/Samsung/ONE/files/9762251/three_tflites.tar.gz)


metadata
```
{
  "major-version" : "1",
  "minor-version" : "3",
  "patch-version" : "0",
  "configs"       : [  ],
  "models"        : [ "mv1.0.tflite", "mv1.1.tflite", "mv1.2.tflite" ],
  "model-types"   : [ "tflite", "tflite", "tflite" ],
  "pkg-inputs"    : [ "0:0:0" ],
  "pkg-outputs"   : [ "2:0:0" ],
  "model-connect" : [ { "from" : "0:0:0", "to" : [ "1:0:0" ] }, { "from" : "1:0:0", "to" : [ "2:0:0" ] } ]
}
```

hseok-oh:
More test case 

[three_tflites2.tar.gz](https://github.com/Samsung/ONE/files/9796564/three_tflites2.tar.gz)

metadata
```
  "major-version" : "1",
  "minor-version" : "3",
  "patch-version" : "0",
  "configs"       : [  ],
  "models"        : [ "inception_v3.14to22.tflite", "inception_v3.23to32.tflite", "inception_v3.33to38.tflite" ],
  "model-types"   : [ "tflite", "tflite", "tflite" ],
  "pkg-inputs"    : [ "0:0:0", "0:0:1", "0:0:2", "0:0:3" ],
  "pkg-outputs"   : [ "2:0:0", "2:0:1", "2:0:2" ],
  "model-connect" : [ { "from" : "0:0:0", "to" : [ "1:0:0" ] }, { "from" : "0:0:1", "to" : [ "1:0:1" ] }, 
[...]

PR #9859, opened 2022-10-12
[onert] Support no branch multimodels execution

hseok-oh:
This commit updates Executors to support no branch multimodels execution.

---

Related issue: #9655
Draft: #9809

### DIFF
--- a/runtime/onert/core/include/exec/Executors.h
+++ b/runtime/onert/core/include/exec/Executors.h
@@ -72,7 +72,7 @@ public:
 private:
   void checkSupportedMultimodel() const;
   void executeModels(const IODescription &desc);
-  uint16_t getModels() const;
+  uint16_t modelCount() const;
 
 private:
   std::unordered_map<std::pair<ir::ModelIndex, ir::SubgraphIndex>, std::unique_ptr<IExecutor>>
--- a/runtime/onert/core/include/exec/Executors.h
+++ b/runtime/onert/core/include/exec/Executors.h
@@ -72,6 +72,7 @@ public:
 private:
   void checkSupportedMultimodel() const;
   void executeModels(const IODescription &desc);
+  uint16_t getModels() const;
 
 private:
   std::unordered_map<std::pair<ir::ModelIndex, ir::SubgraphIndex>, std::unique_ptr<IExecutor>>
--- a/runtime/onert/core/src/exec/Executors.cc
+++ b/runtime/onert/core/src/exec/Executors.cc
@@ -85,48 +85,57 @@ void Executors::execute(const IODescription &desc)
   entryExecutor()->execute(desc);
 }
 
+// Allow below case only
+//  input(s) -> 1st model -> 2nd model -> ... -> (n-1)th model -> (n)th model -> output(s)
 void Executors::checkSupportedMultimodel() const
 {
-  // Assumption
-  // Models
-  //   size: n
-  //   sizeof(model(a).outputs) == sizeof(model(a+1).inputs)
   auto const model_count = modelCount();
 
+  // If package includes no-connection model, model_count is less than real model count in package.
+  // Then this method will throw exception based on model index
+  //  1st model: input assuption
+  //  (n)th model: output assuption
+  //  Otherwise: edges assuption
+
+  // Assumption: executor input/output size
+  //  Models size: n
+  //  sizeof(model(x).outputs) == sizeof(model(x+1).inputs) if 0 < x < n-2
+  //  sizeof(model(0).inputs) == sizeof(package.inputs)
+  //  sizeof(moddel(n-1).outputs) == sizeof(package.outputs)
   for (auto &pair : _executors)
   {
     auto const &model_index = pair.first.first;
     auto const &subg_index = pair.first.second;
 
-    // Inner subgraph
+    // Skip inner subgraph
     if (subg_index != ir::SubgraphIndex{0})
       continue;
 
+    auto executor = at(model_index, subg_index);
+
     // Last model's output
-    auto const next_index = ir::ModelIndex{static_cast<uint16_t>(model_index.value() + 1)};
-    if (_executors.find(std::make_pair(next_index, ir::SubgraphIndex{0})) == _executors.end())
+    if (model_index.value() == model_count - 1)
     {
-      auto executor = at(model_index, subg_index);
       if (executor->graph().getOutputs().size() != _model_edges->pkg_outputs.size())
         throw std::runtime_error{"NYI: Unsupported model edge pattern"};
 
       continue;
     }
 
-    auto executor_from = at(model_index, subg_index);
-    auto executor_to = at(next_index, subg_index);
-
-    if (executor_from->graph().getOutputs().size() != executor_to->graph().getInputs().size())
-      throw std::runtime_error{"NYI: Multi model execution for this package is not supported yet"};
-
     // 1st model's input
     if ((model_index == ir::ModelIndex{0}) &&
-        (executor_from->graph().getInputs().size() != _model_edges->pkg_inputs.size()))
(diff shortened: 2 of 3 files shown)


hseok-oh:
Example https://github.com/Samsung/ONE/issues/9655#issuecomment-1275674110 passed

```
$ ./Product/out/test/onert-test nnpkg-test -i /home/nfs/model/nnpkg/part three_tflites
[  Run  ] /home/nfs/model/nnpkg/part/three_tflites       Pass
[Compare] /home/nfs/model/nnpkg/part/three_tflites       Pass
```

ragmani (dismissed):
LGTM

runtime/onert/core/include/exec/Executors.h
```diff
@@ -72,6 +72,7 @@ class Executors
 private:
   void checkSupportedMultimodel() const;
   void executeModels(const IODescription &desc);
+  uint16_t getModels() const;
```
glistening, line 75: (optional) What about using `modelSize()`, `modelCount()`, ...? I thought it would get the model collection.

hseok-oh: I've renamed to `modelCount()`.

ragmani (dismissed):

hseok-oh:
@nnfw-bot test tizen-gbs-builder

runtime/onert/core/src/exec/Executors.cc
```diff
@@ -87,21 +87,58 @@ void Executors::execute(const IODescription &desc)
...
-  // Assume 2 executors only
-  // Assume that each model may have only one subgraph
-  // TODO Support general case
-  if (_executors.size() != 2)
-    throw std::runtime_error{"NYI: Multi model execution for this package is not supported yet"};
+  // Assumption
+  // Models
+  //   size: n
+  //   sizeof(model(a).outputs) == sizeof(model(a+1).inputs)
+  auto const model_count = modelCount();
```
glistening, line 94: (Just note) I wondered why `model_count` is required when iterating with `for :`. I've found `model_count` is used at the last for-loop.

runtime/onert/core/src/exec/Executors.cc
```diff
@@ -87,21 +87,58 @@ void Executors::execute(const IODescription &desc)
...
+  {
+    auto const &model_index = pair.first.first;
+    auto const &subg_index = pair.first.second;
+
+    // Inner subgraph
+    if (subg_index != ir::SubgraphIndex{0})
+      continue;
+
+    // Last model's output
+    auto const next_index = ir::ModelIndex{static_cast<uint16_t>(model_index.value() + 1)};
```
glistening, line 106: (optional) It would be better hide the internals (like static_cast<uint16_t> ...). I tried to introduce `operator+` in my local. I will make a PR to get your opinion. If it looks dangerous by implicit application, we may introduce `nextIndex()`.

runtime/onert/core/src/exec/Executors.cc
```diff
@@ -87,21 +87,58 @@ void Executors::execute(const IODescription &desc)
...
 
-  // Assume all edges are 0:0:x -> 1:0:x
+  for (auto &pair : _executors)
+  {
+    auto const &model_index = pair.first.first;
+    auto const &subg_index = pair.first.second;
+
+    // Inner subgraph
+    if (subg_index != ir::SubgraphIndex{0})
+      continue;
```
glistening, line 103: It allows multiple subgraph model though we don't support it yet. Is it right?

runtime/onert/core/src/exec/Executors.cc
glistening: I don't understand the comment "Last model's output`. Do you mean next model?

runtime/onert/core/src/exec/Executors.cc
```diff
@@ -87,21 +87,58 @@ void Executors::execute(const IODescription &desc)
...
-  if (_executors.size() != 2)
-    throw std::runtime_error{"NYI: Multi model execution for this package is not supported yet"};
+  // Assumption
+  // Models
+  //   size: n
+  //   sizeof(model(a).outputs) == sizeof(model(a+1).inputs)
+  auto const model_count = modelCount();
 
-  // Assume all edges are 0:0:x -> 1:0:x
+  for (auto &pair : _executors)
```
glistening, line 96: @hseok-oh If there is no executor for model_index _a_ in [0, model_count), this loop seems not detect error. If you expects all models are connected sequentially, what about just iterate with integer index?

runtime/onert/core/src/exec/Executors.cc
```diff
@@ -87,21 +87,58 @@ void Executors::execute(const IODescription &desc)
...
+
+    // 1st model's input
+    if ((model_index == ir::ModelIndex{0}) &&
+        (executor_from->graph().getInputs().size() != _model_edges->pkg_inputs.size()))
+      throw std::runtime_error{"NYI: Unsupported model edge pattern"};
+  }
+
+  // Edges
+  //   a:0:z -> a+1:0:z  (0 < a < n-2)
   for (auto edge : _model_edges->edges)
```
glistening, line 130: Again, apparently it doesn't check there is no missing edge from [1, n-1] .

glistening:
I talked with @hseok-oh offline. To make clear, what is allowed and not allowed.

I got the answers:

Q. Are all models from `MANIFEST` strongly connected?

No. There may be partitions. For example, {model 0, model 1}, { model 2 }. That is model 2 is not connected.

Q. Are all models compiled regardless of connection?

Yes, current implementation compile all models by iterating till model_count, not by traversing model-connects.

Q. Execution ordering should be same to nnpkg manifest models ordering?


@hseok-oh Please check it is right, and give answer for last question. I forgot the conclusion.

hseok-oh:
I've updated usage of `modelCount()` in `checkSupportedMultimodel()`, and added comments.

runtime/onert/core/src/exec/Executors.cc
```diff
@@ -85,26 +85,73 @@ void Executors::execute(const IODescription &desc)
...
-  if (_executors.size() != 2)
-    throw std::runtime_error{"NYI: Multi model execution for this package is not supported yet"};
+  auto const model_count = modelCount();
 
-  // Assume all edges are 0:0:x -> 1:0:x
+  // If package includes no-connection model, model_count is less than real model count in package.
+  // Then this method will throw exception based on model index
+  //  1st model: input assuption
+  //  (n)th model: output assuption
+  //  Otherwise: edges assuption
```
glistening, line 98:
```suggestion
  //  1st model: input assumption
  //  (n)th model: output assumption
  //  Otherwise: edges assumption
```

glistening (approved):
LGTM

hseok-oh:
Answer of https://github.com/Samsung/ONE/pull/9859#issuecomment-1278365150:

I don't understand exactly meaning of **allow** (specification? or current implementation?).

>Q1. Are all models from MANIFEST strongly connected?

We don't have any restriction for connection on package specification.
On this PR, `checkSupportedMultimodel()` will throw exception if there is no-connection model by checking condition.

> Q2. Are all models compiled regardless of connection?

Currently, yes.

>Q3. Execution ordering should be same to nnpkg manifest models ordering?

We don't have any restriction for this.
On this PR, `checkSupportedMultimodel()` will confirm this case because its check condition confirm dependency based on model index order, and `checkSupportedMultimodel()` will executes based on model index order.

>Q4. Is Executors::model_count same to nnpackage MANIFEST models count ?

Please refer updated `Executors::modelCount()` comment.

glistening:
> I don't understand exactly meaning of **allow**. (specification? or current implementation?)

current implementation

glistening (approved):
LGTM

ragmani (approved):

PR #9860, opened 2022-10-12
[onert] nnfw_session::create covers more allocation failures

glistening:
Previously, nnfw_session::create checks only from new nnfw_session(). Now, it covers CustomKernelRegistry and more.

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -208,29 +208,23 @@ NNFW_STATUS nnfw_session::create(nnfw_session **session)
 {
   if (session == nullptr)
     return NNFW_STATUS_UNEXPECTED_NULL;
-
-  // Create session
-  *session = new (std::nothrow) nnfw_session();
-  if (*session == nullptr)
+  try
+  {
+    auto new_session = std::unique_ptr<nnfw_session>(new nnfw_session());
+    new_session->_kernel_registry = std::make_shared<onert::api::CustomKernelRegistry>();
+    *session = new_session.release();
+  }
+  catch (const std::bad_alloc &e)
   {
     std::cerr << "Error during session creation" << std::endl;
     return NNFW_STATUS_OUT_OF_MEMORY;
   }
-
-  // Initialize fields
-  try
-  {
-    (*session)->_kernel_registry = std::make_shared<onert::api::CustomKernelRegistry>();
-  }
   catch (const std::exception &e)
   {
     std::cerr << "Error during session initialization : " << e.what() << std::endl;
-    delete *session;
-    *session = nullptr;
-
+    *session = nullptr; // Set nullptr to keep the old behavior
     return NNFW_STATUS_ERROR;
   }
-
   return NNFW_STATUS_NO_ERROR;
 }
 
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -217,12 +217,13 @@ NNFW_STATUS nnfw_session::create(nnfw_session **session)
   catch (const std::bad_alloc &e)
   {
     std::cerr << "Error during session creation" << std::endl;
+    *session = nullptr; // Set nullptr on error to keep the old behavior
     return NNFW_STATUS_OUT_OF_MEMORY;
   }
   catch (const std::exception &e)
   {
     std::cerr << "Error during session initialization : " << e.what() << std::endl;
-    *session = nullptr; // Set nullptr to keep the old behavior
+    *session = nullptr; // Set nullptr on error to keep the old behavior
     return NNFW_STATUS_ERROR;
   }
   return NNFW_STATUS_NO_ERROR;


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -208,29 +208,24 @@ NNFW_STATUS nnfw_session::create(nnfw_session **session)
...
+    *session = nullptr; // Set nullptr on error to keep the old behavior
+    return NNFW_STATUS_OUT_OF_MEMORY;
   }
   catch (const std::exception &e)
   {
     std::cerr << "Error during session initialization : " << e.what() << std::endl;
-    delete *session;
-    *session = nullptr;
-
+    *session = nullptr; // Set nullptr on error to keep the old behavior
```
ragmani, line 226: I have two questions.
1. Is there any guarantee that the old behavior was `nullptr`? If so, it is necessary to add a code checking `nullptr` before allocating.
2. Is there any case where `*session` was assigned in this block?

And There is a case where `new_session` is not deleted if not called `new_session.release()`. It's memory leak.

ragmani (approved):

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -208,29 +208,24 @@ NNFW_STATUS nnfw_session::create(nnfw_session **session)
...
     return NNFW_STATUS_UNEXPECTED_NULL;
-
-  // Create session
-  *session = new (std::nothrow) nnfw_session();
-  if (*session == nullptr)
+  try
   {
-    std::cerr << "Error during session creation" << std::endl;
-    return NNFW_STATUS_OUT_OF_MEMORY;
+    auto new_session = std::unique_ptr<nnfw_session>(new nnfw_session());
```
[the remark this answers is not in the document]
glistening: See the below. `make_unique` cannot access the private constructor.
```
[ 83%] Building CXX object runtime/onert/api/CMakeFiles/nnfw-dev.dir/src/nnfw_api_internal.cc.o
[ 83%] Built target test_onert_frontend_nnapi
[ 90%] Built target onert_backend_cpu
In file included from /usr/include/c++/9/memory:80,
                 from /home/gyu/z/ONE/runtime/onert/core/include/ir/Model.h:20,
                 from /home/gyu/z/ONE/runtime/onert/core/include/ir/Graph.h:23,
                 from /home/gyu/z/ONE/runtime/onert/core/include/util/TracingCtx.h:20,
                 from /home/gyu/z/ONE/runtime/onert/api/src/nnfw_api_internal.h:23,
                 from /home/gyu/z/ONE/runtime/onert/api/src/nnfw_api_internal.cc:17:
/usr/include/c++/9/bits/unique_ptr.h: In instantiation of ‘typename std::_MakeUniq<_Tp>::__single_object std::make_unique(_Args&& ...) [with _Tp = nnfw_session; _Args = {}; typename std::_MakeUniq<_Tp>::__single_object = std::unique_ptr<nnfw_session, std::default_delete<nnfw_session> >]’:
/home/gyu/z/ONE/runtime/onert/api/src/nnfw_api_internal.cc:213:55:   required from here
/usr/include/c++/9/bits/unique_ptr.h:857:30: error: ‘nnfw_session::nnfw_session()’ is private within this context
  857 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
      |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/gyu/z/ONE/runtime/onert/api/src/nnfw_api_internal.cc:200:1: note: declared private here
  200 | nnfw_session::nnfw_session()
      | ^~~~~~~~~~~~
runtime/onert/api/CMakeFiles/nnfw-dev.dir/build.make:120: recipe for target 'runtime/onert/api/CMakeFiles/nnfw-dev.dir/src/nnfw_api_internal.cc.o' failed
```

runtime/onert/api/src/nnfw_api_internal.cc
glistening: 3. `session` is out variable (pointer to pointer to session). 

> I have two questions.
> 
> 1. Is there any guarantee that the old behavior was `nullptr`?

The old implementations set `session` to `nullptr` in case of out-of-memory error. New code also set to `nullptr` in case of out-of-memory error to keep this old behavior.

> If so, it is necessary to add a code checking `nullptr` before allocating.

I don't understand what you mean by checking `nullptr`. Do you mean the 1st null check for output variable's address? Or do you mean the 2nd null check?

> 2. Is there any case where `*session` was assigned in this block?

I don't understand what you mean by this block. It sets the passed value in `*session`. It may be have some value prior to `nnfw_session::create.

> And there is a case where `new_session` is not deleted if not called `new_session.release()`. 

Again, I don't understand what you mean. `nnfw_session.release()` is to get the ownership of `nnfw_session` allocation.

> It's memory leak.

Could you explain more? From `session` or `_kernel_registry`? When does it happen?

hseok-oh (approved):

runtime/onert/api/src/nnfw_api_internal.cc
ragmani: > The old implementation sets session (output parameter) to nullptr in case of out-of-memory. The content in *session will be overwritten to nullptr. New code also sets to nullptr in case of out-of-memory error to keep this old behavior.

The old implementation also overwrites `*session` to `nullptr`. ㄲegardless of the fact, do you mean the old behavior is that `*session ` was `nullptr` before called this function? If so, I thought it would be better to add an exception handling with checking `*session != nullptr`. 

> I don't understand what you mean by checking nullptr. Do you mean the 1st null check for output variable's address? Or do you mean the 2nd null check?

I mean the `*session` null check.

> Could you please explain more? From session or _kernel_registry? When does it happen?

Sorry, I missed `unique_ptr`. Please ignore about memory leak.

> I don't understand what you mean by this block. It sets the passed value in *session. It may be have some value prior to nnfw_session::create.

My question was related to memory leak. Please ignore my 2nd question.

glistening: > The old implementation also overwrites `*session` to `nullptr`. 

Yes, it is exactly what I mean by "to keep the old behavior". I don't change the old behavior.

> Regardless of the fact, do you mean the old behavior is that `*session ` was `nullptr` before called this function? 

No. Old behavior what I mean is setting `*session` to `nullptr` on out-of-memory.

PR #9863, opened 2022-10-14
[locomotiv] Add nullptr exception handling

llFreetimell:
Some of nullptr exception handlings are missed.
This commit will add them.

Signed-off-by: Seok NamKoong <seok9311@naver.com>

---

This PR is because of static analysis warnings.

### DIFF
--- a/compiler/locomotiv/src/Node/TensorBroadcast.cpp
+++ b/compiler/locomotiv/src/Node/TensorBroadcast.cpp
@@ -42,6 +42,10 @@ using namespace locomotiv;
 void execute_node(loco::TensorBroadcast *tensor_broadcast)
 {
   auto input_data = annot_data(tensor_broadcast->input());
+  if (input_data == nullptr)
+  {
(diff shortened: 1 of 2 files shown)


compiler/locomotiv/src/Node/TensorBroadcast.cpp
```diff
@@ -42,6 +42,10 @@ using namespace locomotiv;
 void execute_node(loco::TensorBroadcast *tensor_broadcast)
 {
   auto input_data = annot_data(tensor_broadcast->input());
+  if (input_data == nullptr)
+  {
+    throw std::runtime_error("Any annotation is found in TensorBroadcast");
```
seanshpark, line 47: Q) what is `Any` ?

llFreetimell: As I thought that `annot_data` returns annotated data from node, `input_data == nullptr` means that there are no annotations..!
Could you recommend other error msg? :)

seanshpark (approved):
LGTM thank you!

Issue #9872, opened 2022-10-18
[one-optmize] Introduce ForwardTransposeOpPass


jinevening:
## What

Let's introduce `ForwardTranseposeOpPass`.

## Why

This moves Transpose Ops to the rear part, finding a chance to optimize redundant Transpose Ops (similar to `ForwardReshapeToUnaryOpPass`).

Before
![image](https://user-images.githubusercontent.com/5449554/196355195-ab70d53b-7374-4e65-b2cb-842c8354a53a.png)

After
![image](https://user-images.githubusercontent.com/5449554/196355046-d916b286-ae65-47a8-90ed-7630d67f2f5c.png)

jinevening:
This reduces the latency of one of our target models.

6.2 ms -> 5.6 ms

PR #9873, opened 2022-10-18
DRAFT: Introduce ForwardTransposeOp pass

jinevening:
On going draft to introduce ForwardTransposeOpPass.

---
Related to: https://github.com/Samsung/ONE/issues/9872

### DIFF
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -26,6 +26,7 @@
 #include "luci/Pass/FoldGatherPass.h"
 #include "luci/Pass/FoldSparseToDensePass.h"
 #include "luci/Pass/ForwardReshapeToUnaryOpPass.h"
+#include "luci/Pass/ForwardTransposeOpPass.h"
 #include "luci/Pass/FuseActivationFunctionPass.h"
 #include "luci/Pass/FuseAddWithFullyConnectedPass.h"
 #include "luci/Pass/FuseAddWithTConvPass.h"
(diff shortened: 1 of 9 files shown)


compiler/luci/pass/src/ForwardTransposeOpPass.test.cpp
```diff
@@ -0,0 +1,161 @@
...
+public:
+  void run_pass(void) { run_phase(_graph.g()); }
+
+protected:
+  ForwardTransposeToAddGraph _graph;
+};
+
+} // namespace
+
+TEST_F(ForwardTransposeToAddGraphTest, forward_add)
```
seanshpark, line 132: can we have negative tests too?

jinevening: I'll add negative tests when I upload PRs.

compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -26,6 +26,7 @@
 #include "luci/Pass/FoldGatherPass.h"
 #include "luci/Pass/FoldSparseToDensePass.h"
 #include "luci/Pass/ForwardReshapeToUnaryOpPass.h"
+#include "luci/Pass/ForwardTransposeOpPass.h"
```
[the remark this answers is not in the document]
jinevening: I think that it is not necessary to split the pass according to arity.

If we split the pass, users will have to turn on multiple options that basically do the same thing (forwarding Transpose).

ForwardTransposeToUnary..
ForwardTransposeToBinary..
ForwardTransposeToMultiInput...

We can split the passes internally, but I think it would be better to expose a single option to users.

jinevening:
- [ ] luci
- [ ] res
- [ ] circle-quantizer-dredd-test

jinevening:
Done.

PR #9868, opened 2022-10-17
DRAFT: Introduce visq module

jinevening:
On going draft to introduce visq module.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/9694

PR #9876, opened 2022-10-18
[visq] Introduce visq module

jinevening:
This introduces visq for visualizing quantization error.

---
Related to: https://github.com/Samsung/ONE/issues/9694
Draft PR: #9868

### DIFF
--- a/compiler/visq/visq
+++ b/compiler/visq/visq
@@ -1,0 +1,206 @@
+#!/usr/bin/env bash
+''''export SCRIPT_PATH="$(cd "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")" && pwd)" # '''
+''''export PY_PATH=${SCRIPT_PATH}/venv/bin/python                                       # '''
+''''test -f ${PY_PATH} && exec ${PY_PATH} "$0" "$@"                                     # '''
+''''echo "Error: Virtual environment not found. Please run 'one-prepare-venv' command." # '''
+'''exit 255                                                                            # '''
+
+# Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import subprocess
+import tempfile
+import json
+import os
+import math
+
+from shutil import copyfile
+from pathlib import Path
+
+from visqlib.Palette import YLORRD9Palette
+from visqlib.QErrorComputer import MPEIRComputer
+from visqlib.Util import valid_attr, pretty_float
+
+
+def _get_parser():
+    parser = argparse.ArgumentParser(
+        description='Command line tool to visualize layer-wise quantization errors')
+    parser.add_argument(
+        "-f",
+        "--fp32_circle",
+        type=str,
+        help="Path to the fp32 circle model.",
+        required=True)
+    parser.add_argument(
+        "-q",
+        "--q_circle",
+        type=str,
+        help="Path to the quantized circle model.",
+        required=True)
+    parser.add_argument(
+        "-d",
+        "--data",
+        type=str,
+        help=
+        "Path to the data used for inference. Random data will be used if this option is not given.",
+        required=False)
+    parser.add_argument(
+        "--mpeir_output",
+        type=str,
+        help="Path to the output dot file (qerror metric = mPEIR).",
+        required=False)
+    return parser
+
+
+def _verify_args(args):
+    """Verify the given arguments"""
+
+    valid_outputs = ['mpeir_output', 'mse_output', 'mpesr_output']
+
+    # Check if at least one output option is given
+    num_outputs = 0
+    for output_name in valid_outputs:
+        if valid_attr(args, output_name):
+            num_outputs += 1
+
+    if num_outputs == 0:
+        raise RuntimeError("At least one output should be given.")
+
+
+def _run_dalgona(model, data, analysis, save_dir):
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    dalgona_path = os.path.join(dir_path, 'dalgona')
+    cmd = [dalgona_path]
+    cmd += ['--input_model', str(model)]
+    cmd += ['--analysis', str(analysis)]
+    if data != None:
+        cmd += ['--input_data', str(data)]
+    cmd += ['--analysis_args', str(save_dir)]
+
+    try:
+        subprocess.run(cmd, capture_output=True, check=True)
+    except subprocess.CalledProcessError as e:
+        print('Error raised while running the below command')
+        print(' '.join(cmd))
+        print(e.output)
+        raise
+
+
+def _fake_quantize(input_model, output_model):
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    circle_quantizer_path = os.path.join(dir_path, 'circle-quantizer')
+    cmd = [circle_quantizer_path]
+    cmd += ['--fake_quantize']
+    cmd += [str(input_model)]
+    cmd += [str(output_model)]
+
+    try:
+        subprocess.run(cmd, capture_output=True, check=True)
+    except subprocess.CalledProcessError as e:
+        print('Error raised while running the below command')
+        print(' '.join(cmd))
+        print(e.output)
+        raise
+
+
+# Recursively visit items and check if there is Infinity or NaN
+def _check_float(item):
+    if isinstance(item, dict):
+        for v in item.values():
+            _check_float(v)
+    if isinstance(item, list):
+        for v in item:
+            _check_float(v)
+    if isinstance(item, float):
+        if item == -float('inf') or item == float('inf'):
+            raise RuntimeError('Infinite value detected. Value must be float')
+        if math.isnan(item):
+            raise RuntimeError('NaN value detected. Value must be float')
+
+
+def _build_json(**kwargs):
+    # model: string
+    # metric: string
+    # colorscheme: list ['b': begin, 'e': end, 'c':color]
+    # error: dict {tensor_name:error}
+    assert 'metric' in kwargs
+    assert 'model' in kwargs
+    assert 'colorscheme' in kwargs
+    assert 'error' in kwargs
+
+    meta = {}
+    meta["model"] = Path(kwargs['model']).name
+    meta["metric"] = kwargs['metric']
+    meta["colorscheme"] = pretty_float(kwargs['colorscheme'])
+    result = {}
+    result["meta"] = meta
+    # Why list? To support multiple subgraphs
+    result["error"] = [pretty_float(kwargs['error'])]
+
+    # Invariants
+    _check_float(meta["colorscheme"])
+    _check_float(result["error"])
+    return result
+
+
+def main():
+    # parse arguments
+    parser = _get_parser()
+    args = parser.parse_args()
+    _verify_args(args)
+
+    fp32_model = args.fp32_circle
+    q_model = args.q_circle
+    data = None
+    if valid_attr(args, 'data'):
+        data = args.data
+
+    curr_dir = Path(__file__).parent.resolve()
+    dump_fp32_py = curr_dir / 'visqlib' / 'DumpFP32FM.py'
+    dump_fq_py = curr_dir / 'visqlib' / 'DumpFakeQuantFM.py'
+
+    with tempfile.TemporaryDirectory() as model_dir, \
+         tempfile.TemporaryDirectory() as fp32_dir, \
+         tempfile.TemporaryDirectory() as fq_dir:
+        fq_model = model_dir + '/fq_model.circle'
+
+        # Step 1. Fake quantize q circle
+        _fake_quantize(q_model, fq_model)
+
+        # Step 2. Run dalgona to dump intermediate FMs in fp32 model
+        _run_dalgona(fp32_model, data, dump_fp32_py, fp32_dir)
+
+        # Copy list of dumped tensors
+        copyfile(fp32_dir + '/tensors.txt', fq_dir + '/tensors.txt')
+
+        # Step 3. Run dalgona to dump intermediate FMs in fq model
+        _run_dalgona(fq_model, data, dump_fq_py, fq_dir)
+
+        # Step 4. Read results and compute qerror
+        if args.mpeir_output:
+            computer = MPEIRComputer(fp32_dir, fq_dir)
+            qerror_map = computer.run()
+            palette = YLORRD9Palette(min=0.0, max=1.0)
+            result = _build_json(
+                metric='MPEIR',
+                model=fp32_model,
+                colorscheme=palette.colorscheme(),
+                error=qerror_map)
+            with open(args.mpeir_output, "w") as f:
+                json.dump(result, f)
+
+
+if __name__ == '__main__':
+    main()
--- a/compiler/visq/visq
+++ b/compiler/visq/visq
@@ -25,6 +25,7 @@ import tempfile
 import json
 import os
 import math
+import sys
 
 from shutil import copyfile
 from pathlib import Path
@@ -59,7 +60,7 @@ def _get_parser():
     parser.add_argument(
         "--mpeir_output",
         type=str,
-        help="Path to the output dot file (qerror metric = mPEIR).",
+        help="Path to the output json file (qerror metric = MPEIR).",
         required=False)
     return parser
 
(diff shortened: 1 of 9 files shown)


compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+        "-d",
+        "--data",
+        type=str,
+        help=
+        "Path to the data used for inference. Random data will be used if this option is not given.",
+        required=False)
+    parser.add_argument(
+        "--mpeir_output",
+        type=str,
+        help="Path to the output dot file (qerror metric = mPEIR).",
```
seanshpark, line 62: README shows it's `*.vsiq.json` but here `dot file`

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+        required=False)
+    return parser
```
seanshpark, line 64: no argument for `--mse_output` ?

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+
+
+def _verify_args(args):
+    """Verify the given arguments"""
+
+    valid_outputs = ['mpeir_output', 'mse_output', 'mpesr_output']
```
seanshpark, line 70: Q) README doesn't show about `mpesr_output`. and also parser.

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+    curr_dir = Path(__file__).parent.resolve()
+    dump_fp32_py = curr_dir / 'visqlib' / 'DumpFP32FM.py'
+    dump_fq_py = curr_dir / 'visqlib' / 'DumpFakeQuantFM.py'
+
+    with tempfile.TemporaryDirectory() as model_dir, \
+         tempfile.TemporaryDirectory() as fp32_dir, \
+         tempfile.TemporaryDirectory() as fq_dir:
+        fq_model = model_dir + '/fq_model.circle'
+
+        # Step 1. Fake quantize q circle
```
seanshpark, line 179: Q) `q` here is `quantized` ?
And about `circle` here, other comments use `model`, so IIUC, how about `Fake quantize quantized circle model` ?

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+        _fake_quantize(q_model, fq_model)
+
+        # Step 2. Run dalgona to dump intermediate FMs in fp32 model
```
seanshpark, line 182: ... `FP32 model` ?

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+        _run_dalgona(fp32_model, data, dump_fp32_py, fp32_dir)
+
+        # Copy list of dumped tensors
+        copyfile(fp32_dir + '/tensors.txt', fq_dir + '/tensors.txt')
```
seanshpark, line 186: Q) where does `tensors.txt` name come from?

compiler/visq/visq
seanshpark: draft PR has all but adding all types for `valid_outputs` makes confusion.
maybe add only valid items for this PR...

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+                metric='MPEIR',
+                model=fp32_model,
+                colorscheme=palette.colorscheme(),
+                error=qerror_map)
+            with open(args.mpeir_output, "w") as f:
+                json.dump(result, f)
+
+
+if __name__ == '__main__':
+    main()
```
mhs4670go, line 206: Could make print better log like `one-cmds` tools do.
```suggestion
    try:
        main()
    except Exception as e:
        prog_name = os.path.basename(mainpath)
        print(f"{prog_name}: {type(e).__name__}: " + str(e), file=sys.stderr)
        sys.exit(255)
```

compiler/visq/visqlib/DumpFP32FM.py
```diff
@@ -0,0 +1,54 @@
...
+# Script that dumps FM of fp32 model
+# NOTE This script runs on dalgona
+
+import numpy as np
+
+from pathlib import Path
+from Util import to_filename
+
+
+# Dump fp32 model's intermediate FM data and their names
```
seanshpark, line 24: maybe `fp32` -> `FP32` ?

compiler/visq/visqlib/DumpFakeQuantFM.py
```diff
@@ -0,0 +1,65 @@
...
+# NOTE This script runs on dalgona
+
+import numpy as np
+
+from pathlib import Path
+from Util import to_filename
+
+# Fake-quantized Op has the postfix of fq_postfix
+# TODO Remove coupling with fake quantization codes
+fq_postfix = '_FQ_Quantize_FQ_Dequantize'
```
seanshpark, line 25: (optional) for shorter name, `_Fake_Qu_DeQu` ?

compiler/visq/visq
seanshpark: Q) what is `FM` ?

compiler/visq/visq
mhs4670go: +1

would be better to comment or append readme for the abbreviations.

For the reviewers, 

MAE: Mean Absolute Error
MSE: Mean Squared Error
MAPE: Mean Absolute Percentage Error
MPEIR: Mean Peak Error to Interval Ratio = max(|a - b|) / (max(a) - min(a))

compiler/visq/visqlib/QErrorComputer.py
```diff
@@ -0,0 +1,96 @@
...
+                fq_data = np.load(fq_data_path)
+
+                diff = np.absolute(fp32_data - fq_data).reshape(-1)
+
+                fp32_min = np.min(fp32_data.reshape(-1))
+                fp32_max = np.max(fp32_data.reshape(-1))
+
+                # Peak Error-to-Interval Ratio (PEIR)
+                #
+                # DESIGN NOTE: PEIR is an analogue of PSNR
```
seanshpark, line 78: (optioanl) maybe provide also for `PSNR` meaning?

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+        cmd += ['--input_data', str(data)]
+    cmd += ['--analysis_args', str(save_dir)]
+
+    try:
+        subprocess.run(cmd, capture_output=True, check=True)
+    except subprocess.CalledProcessError as e:
+        print('Error raised while running the below command')
+        print(' '.join(cmd))
+        print(e.output)
+        raise
```
mhs4670go, line 98: This can be another function.

mhs4670go: Why is `capture_output` is used? If you want to capture the output, you would initialize a variable for output.
```python
captured_output = subprocess.run(cmd, capture_output=True, check=True)
```

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+    args = parser.parse_args()
+    _verify_args(args)
+
+    fp32_model = args.fp32_circle
+    q_model = args.q_circle
+    data = None
+    if valid_attr(args, 'data'):
+        data = args.data
+
+    curr_dir = Path(__file__).parent.resolve()
```
mhs4670go, line 170: Is there any reason there are two ways to get current directory?
```python
dir_path = os.path.dirname(os.path.realpath(__file__)) # L83
```

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+    num_outputs = 0
+    for output_name in valid_outputs:
+        if valid_attr(args, output_name):
+            num_outputs += 1
+
+    if num_outputs == 0:
+        raise RuntimeError("At least one output should be given.")
+
+
+def _run_dalgona(model, data, analysis, save_dir):
```
mhs4670go, line 82: (optional) Would be better to know about exact spec of function if there are type annotations for parameters.

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+    q_model = args.q_circle
+    data = None
+    if valid_attr(args, 'data'):
+        data = args.data
+
+    curr_dir = Path(__file__).parent.resolve()
+    dump_fp32_py = curr_dir / 'visqlib' / 'DumpFP32FM.py'
+    dump_fq_py = curr_dir / 'visqlib' / 'DumpFakeQuantFM.py'
+
+    with tempfile.TemporaryDirectory() as model_dir, \
```
mhs4670go, line 174: For reviewers, `tempfile.TemporaryDirectory` returns `string` type path with `with` statement.
```python
def __enter__(self):
    return self.name
```

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+        for v in item:
+            _check_float(v)
+    if isinstance(item, float):
+        if item == -float('inf') or item == float('inf'):
+            raise RuntimeError('Infinite value detected. Value must be float')
+        if math.isnan(item):
+            raise RuntimeError('NaN value detected. Value must be float')
+
+
+def _build_json(**kwargs):
```
mhs4670go, line 133: I think if the number of arguments are not arbitrary, it would be better to use named arguments. Explicit is better than implicit. And, we'd be able to use type annotation.
```suggestion
def _build_json(model: str, metric: str, colorscheme: list[str], error: dict):
```

compiler/visq/visqlib/Util.py
```diff
@@ -0,0 +1,37 @@
...
+    return tensor_name.replace('/', '_')[-255:]
+
+
+# Check if attr is valid
+def valid_attr(args, attr):
+    return hasattr(args, attr) and getattr(args, attr)
+
+
+# Recursively visit items and round floats with ndigits=4
+def pretty_float(item):
```
mhs4670go, line 30: How about make `ndigits` a dedicated parameter for flexibility?
```suggestion
# Recursively visit items and round floats with ndigits=4
def pretty_float(item, ndigit):
```

compiler/visq/visq
seanshpark: (optional) maybe make this a class and use setters to set attributes and provide `build()` like method.

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+    try:
+        subprocess.run(cmd, capture_output=True, check=True)
+    except subprocess.CalledProcessError as e:
+        print('Error raised while running the below command')
+        print(' '.join(cmd))
+        print(e.output)
+        raise
+
+
+def _fake_quantize(input_model, output_model):
```
mhs4670go, line 101: (optional) We could make a dedicated command builder like the one you wrote for `one-quantize`.

```python
class CommandBuilder:
  def __init__(self, driver):
    # ..

  def add_optional(self, name: str, *, N: int, values: list):
    # ..

  def build(self):
    # ..

# ..

cb = ComandBuilder('circle-quantizer')
cb.add_optional('fake_quantize', 2, [input_model, output_model])
subprocess_run(cb.build())
```

compiler/visq/visq
```diff
@@ -0,0 +1,206 @@
...
+        subprocess.run(cmd, capture_output=True, check=True)
+    except subprocess.CalledProcessError as e:
+        print('Error raised while running the below command')
+        print(' '.join(cmd))
+        print(e.output)
+        raise
+
+
+# Recursively visit items and check if there is Infinity or NaN
+def _check_float(item):
```
mhs4670go, line 119: (optional) Would be better to write tests for these functions and `visqlib`'s classes and functions.

compiler/visq/visqlib/DumpFP32FM.py
```diff
@@ -0,0 +1,54 @@
...
+#
+# Before
+# self._dir/
+#
+# After
+# self._dir/
+#  tensors.txt
+#  <TENSOR_NAME>.npy
+# NOTE TENSOR_NAME is transformed by to_filename
+class DumpFP32FM(object):
```
mhs4670go, line 34: Just curiosity, what is this `object` keyword for?

mhs4670go: (optional) And if there is an interface for dalgona, it would be helpful to write a dalgona analysis script.

mhs4670go: (optional2) It is not related this PR but we can introduce an option that receives python script and the class name to analyze. Then, users manages multiple simple classes by just a one python script.

For example,
```bash
$ cat DumpFM.py
class DumpFP32FM(object):
..
class DumpFakeQuantFM(object):
..


```

compiler/visq/visqlib/Palette.py
```diff
@@ -0,0 +1,73 @@
...
+
+            self._slots.append([lower_bound, upper_bound])
+
+        # Invariant
+        assert len(self._slots) == num_colors
+
+
+# Palette for ylorrd9 colorscheme
+class YLORRD9Palette(UniformPalette):
+    def __init__(self, **kwargs):
```
mhs4670go, line 61: ditto

compiler/visq/visq
jinevening: I planned to add mse_output in the next PR :)

compiler/visq/visq
jinevening: > Q) README doesn't show about mpesr_output. and also parser.

I planned to upload mpeir only for easier reivew, so I should have removed mse and mpesr here. Let me update the PR. Thanks!

compiler/visq/visq
jinevening: FM is abbreviation of feature map.

compiler/visq/visq
jinevening: No. Copied from multiple sources. I'll use one way.

compiler/visq/visq
jinevening: I'll remove the option.

compiler/visq/visq
jinevening: I agree that type annotations improve readability, but python codes in ONE are not using type annotations at all. Why don't we change the style in a separate issue?

compiler/visq/visq
jinevening: I have some unittests in draft. We may add more tests in the future :)

compiler/visq/visqlib/DumpFP32FM.py
jinevening: > Just curiosity, what is this object keyword for?

`object` is unnecessary. Thanks for finding.

> (optional) And if there is an interface for dalgona, it would be helpful to write a dalgona analysis script.

DumpFP32FM.py and DumpFakeQuantFM,py are dalgona analysis scripts.

> (optional2) It is not related this PR but we can introduce an option that receives python script and the class name to analyze. Then, users manages multiple simple classes by just a one python script.

Dalgona analysis should have the class with the name same with the file name. Ex: DumpFP32FM class in DumpFP32FM.py.

compiler/visq/visqlib/DumpFakeQuantFM.py
```diff
@@ -0,0 +1,65 @@
...
+    def StartAnalysis(self, args):
+        self._dir = Path(args)
+        self._num_data = 0
+        with open(self._dir / 'tensors.txt') as f:
+            self._target_tensors = set([line.rstrip() for line in f])
+
+    def EndNetworkExecution(self, outputs: list):
+        self._num_data += 1
+
+    def DefaultOpPost(self, name, opcode, inputs, output):
```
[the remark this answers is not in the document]
jinevening: Dequantize Op is not implemented in dalgona, so I used DefaultOpPost. It would be possible to replace it with DequantizePost when it is ready.

compiler/visq/visq
mhs4670go: @jinevening I got it. Actually I think there is no problem with that codes. But if you can do the same thing with below codes and let readers know about your intent.

```python
subprocess.run(cmd, stdout=DEVNULL, stderr=STDOUT, check=True)
```

compiler/visq/visqlib/DumpFP32FM.py
mhs4670go: @jinevening 
> DumpFP32FM.py and DumpFakeQuantFM,py are dalgona analysis scripts.

Right. I think if we can force dalgona script writers to implement the things that should be implemented by interface.

For example,
```python
class DumpFP32FM(IDalgona):
  # if `StartAnalysis` is missing, throw an error
```

> Dalgona analysis should have the class with the name same with the file name. Ex: DumpFP32FM class in DumpFP32FM.py.

Right. I just wanted to check if we could consider changing the policy where one file can have multiple analyzing classes. But current policy looks good as well.

jinevening: > I think if we can force dalgona script writers to implement the things that should be implemented by interface.

At least for now, I think `StartAnalysis` (or any other method) does not have to be implemented. We can think of such an interface when necessary.

> I just wanted to check if we could consider changing the policy where one file can have multiple analyzing classes. But current policy looks good as well.

Hmm.. This may change the interface of `dalgona`. I'm not sure we can do this easily. Let's think about it later :)

compiler/visq/visqlib/DumpFakeQuantFM.py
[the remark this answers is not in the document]
seanshpark: maybe add a TODO or NOTE comment?

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

compiler/visq/visqlib/DumpFakeQuantFM.py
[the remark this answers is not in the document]
jinevening: I've added TODO comment :)

```python
    # TODO Use DequantizePost when dalgona supports it
```

PR #9888, opened 2022-10-20
[visq] Add MSE metric

jinevening:
This adds Mean Squared Error metric.

---
Related to: https://github.com/Samsung/ONE/issues/9694
Draft PR: https://github.com/Samsung/ONE/pull/9868

### DIFF
--- a/compiler/visq/visq
+++ b/compiler/visq/visq
@@ -209,7 +209,7 @@ def main():
             palette = YLORRD9Palette(qerror_min=q_min, qerror_max=q_max)
             result = _build_json(
                 metric='MSE',
-                model=fp32_model,
+                model=Path(fp32_model).name,
                 colorscheme=palette.colorscheme(),
                 error=qerror_map)
             with open(args.mse_output, "w") as f:
--- a/compiler/visq/visq
+++ b/compiler/visq/visq
@@ -31,7 +31,7 @@ from shutil import copyfile
 from pathlib import Path
 
 from visqlib.Palette import YLORRD9Palette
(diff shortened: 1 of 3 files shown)


compiler/visq/visq
```diff
@@ -198,6 +203,18 @@ def main():
             with open(args.mpeir_output, "w") as f:
                 json.dump(result, f)
 
+        if args.mse_output:
+            computer = MSEComputer(fp32_dir, fq_dir)
+            qerror_map, q_min, q_max = computer.run()
+            palette = YLORRD9Palette(qerror_min=q_min, qerror_max=q_max)
+            result = _build_json(
+                metric='MSE',
+                model=fp32_model,
```
mhs4670go, line 212: it is not `Path(fp32_model).stem`?

seanshpark: viewer assumes model path is relative to this json or absolute path is also ok.

mhs4670go: I got it but above `mpeir_output` case uses `Path(fp32_model).stem`. Is there any difference?

jinevening: Ah, I misunderstood the spec of meta section. Actually it should be `Path(..).name`. Ex: "test.circle"

Path(fp32_model).name -> "test.name"
Path(fp32_model).stem -> "test"

stem will be updated in the next PR. Thanks for finding!

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #9894, opened 2022-10-24
[res] Revise TF examples to output folder

seanshpark:
This will revise TensorFlowPythonExamples to generate to output folder and ignore from git.


mhs4670go (approved):
LGTM

Issue #9895, opened 2022-10-24
[CFE] Import TensorFlow RNN/LSTM/GRU


seanshpark:
Let's findout how to import TensorFlow LSTM operator.

reference:
- https://www.tensorflow.org/api_docs/python/tf/keras/layers/LSTM
- https://www.tensorflow.org/api_docs/python/tf/keras/layers/Bidirectional
- https://www.tensorflow.org/lite/models/convert/rnn

seanshpark:
related #8217

seanshpark:
With https://github.com/Samsung/ONE/blob/6a8a5e3a7247a246271286ada127a2eba6f8c4ef/res/TensorFlowPythonExamples/examples/lstm/__init__.py

```py
        # tflite export for experiments
        converter = tf.lite.TFLiteConverter.from_keras_model(m.model)
        converter.allow_custom_ops = True
        converter.experimental_new_converter = True
        converter.target_spec.supported_ops = [tf.lite.OpsSet.TFLITE_BUILTINS]

        tflite_model = converter.convert()
```

gives
> error: failed to legalize operation 'tf.TensorListReserve' that was explicitly marked illegal
error: Lowering tensor list ops is failed. Please consider using Select TF ops and disabling `_experimental_lower_tensor_list_ops` flag in the TFLite converter object. For example, converter.target_spec.supported_ops = [tf.lite.OpsSet.TFLITE_BUILTINS, tf.lite.OpsSet.SELECT_TF_OPS]\n converter._experimental_lower_tensor_list_ops = False

seanshpark:
With `converter._experimental_lower_tensor_list_ops = False`,

> The following operation(s) need TFLite custom op implementation(s):
Custom ops: TensorListFromTensor, TensorListGetItem, TensorListReserve, TensorListSetItem, TensorListStack
Details:
        tf.TensorListFromTensor(tensor<4x?x4xf32>, tensor<2xi32>) -> (tensor<!tf_type.variant<tensor<?x4xf32>>>) : {device = ""}
        tf.TensorListGetItem(tensor<!tf_type.variant<tensor<?x4xf32>>>, tensor<i32>, tensor<2xi32>) -> (tensor<?x4xf32>) : {device = ""}
        tf.TensorListReserve(tensor<2xi32>, tensor<i32>) -> (tensor<!tf_type.variant<tensor<?x2xf32>>>) : {device = ""}
[...]

PR #9899, opened 2022-10-24
DRAFT TF LSTM

seanshpark:
on-going draft to work with LSTM.


seanshpark:
for  #9895

seanshpark:
done

PR #9903, opened 2022-10-24
[luci] Introduce ForwardTransposeOpPass

jinevening:
This introduces Forward Transpose Op Pass.

---
Related to: https://github.com/Samsung/ONE/issues/9872
Draft PR: #9873

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/ForwardTransposeOpPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ForwardTransposeOpPass.h
@@ -23,7 +23,7 @@ namespace luci
 {
 
 /**
- * @brief  Class to Forward send Reshape after UnaryOp.
+ * @brief  Class to Forward Transpose Ops for further optimization.
  */
 struct ForwardTransposeOpPass final : public logo::Pass
 {
--- a/compiler/luci/pass/include/luci/Pass/ForwardTransposeOpPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ForwardTransposeOpPass.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


compiler/luci/pass/include/luci/Pass/ForwardTransposeOpPass.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef __LUCI_FORWARD_TRANSPOSE_OP_PASS_H__
+#define __LUCI_FORWARD_TRANSPOSE_OP_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to Forward send Reshape after UnaryOp.
```
seanshpark, line 26: maybe this message needs revision?

jinevening: Ah.. I missed it. Thanks!

compiler/luci/pass/src/ForwardTransposeOpPass.test.cpp
```diff
@@ -0,0 +1,173 @@
...
+TEST_F(ForwardTransposeToAddGraphTest, forward_transpose_NEG)
+{
+  _graph.init({1, 64, 51, 1}, {0, 3, 2, 1});
+
+  // Remove add
+  _graph.output()->from(_graph.transpose());
+
+  luci::ForwardTransposeOpPass pass;
+  EXPECT_FALSE(pass.run(_graph.g()));
+}
```
seanshpark, line 173: can you add more tests with (1) where add->x() is transpose, (2) op is mul ?
reason: to maintain test coverage, which we need to maintain for SE requirements.

seanshpark: I think #9905 can cover (1)

jinevening: You're right. I'll add more tests to improve coverage

seanshpark (approved):
LGTM thank you!

PR #9908, opened 2022-10-24
[res] Remove pbtxt option for TensorFlowPythonExamples

seanshpark:
This will remove pbtxt mode option for TensorFlowPythonExamples.


seanshpark:
from draft #9899 
- keras model will be generated if possible without this option in following PRs

mhs4670go (approved):
LGTM

PR #9912, opened 2022-10-24
[res] Fix Path is not defined

seanshpark:
This will fix Path is not defined error in TensorFlowPythonExamples.


mhs4670go (approved):
LGTM

PR #9913, opened 2022-10-24
[res] Save keras and tflite models

seanshpark:
This will revise TensorFlowPythonExamples to save keras and tflite model if possible.


seanshpark:
for #9895 , from #9899

mhs4670go (approved):
LGTM

PR #9916, opened 2022-10-24
[onecc] Change the way removing whitespace in Makefile.arm32

ragmani:
This commit changes the way removing whispace in Makefile.arm32.


ragmani:
This way is from https://web.mit.edu/gnu/doc/html/make_8.html

seanshpark (approved):
LGTM

PR #9917, opened 2022-10-24
[res] Introduce LSTM_unroll for TF examples

seanshpark:
This will introduce LSTM_unroll in TensorFlowPythonExamples
- LSTM with unroll
- use known(1) batch size


seanshpark:
for #9895 , from #9899

mhs4670go (approved):
LGTM

PR #9927, opened 2022-10-25
[res] GRU_unroll TF example

seanshpark:
This will introduce GRU_unroll TF python example.


seanshpark:
for #9895

mhs4670go (approved):
LGTM

PR #9929, opened 2022-10-26
[npud] Introduce DBus and dbus introspection xml

jyoungyun:
This commit add DBus class and dbus introspection xml file.
The introspection xml file is converted to be c code using
gdbus-codegen file.


jyoungyun:
### How to run

**Server**
```bash
$ NPUD_LOG_ENABLE=1 ./Product/x86_64-linux.debug/out/bin/npud
[      DBus      ] Starting dbus service
[      main      ] Starting npud
[     Server     ] Starting Server
[      DBus      ] org.tizen.npud on bus acquired
[      DBus      ] org.tizen.npud on name acquired
```

**Client**
```bash
$ dbus-send \
    --system \
    --print-reply \
    --dest=org.tizen.npud \
    /org/tizen/npud \
    org.tizen.npud.core.device_get_available_list
```

hseok-oh:
```
17:00:26  [----------] 6 tests from SignalTest
17:00:26  [ RUN      ] SignalTest.raise_SIGTERM
17:00:26  terminate called after throwing an instance of 'std::runtime_error'
17:00:26    what():  Mainloop is not running
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  terminate called recursively
17:00:26  [       OK ] SignalTest.raise_SIGTERM (1 ms)
17:00:26  /jenkins/aws/workspace/nnfw/master/pr-cross-aarch64-runtime-test-pipeline-release/Product/out/test/command/unittest: line 80: 13575 Aborted                 (core dumped) $TEST_BIN $(get_gtest_option)
```

jyoungyun:

hseok-oh:
@nnfw-bot test cross-aarch64-runtime-test-debug

hseok-oh (approved):
LGTM

seanshpark (approved):
LGTM

runtime/service/npud/core/DBus.h
```diff
@@ -0,0 +1,56 @@
...
+  static void on_bus_acquired(GDBusConnection *conn, const gchar *name, gpointer user_data);
+  static void on_name_acquired(GDBusConnection *conn, const gchar *name, gpointer user_data);
+  static void on_name_lost(GDBusConnection *conn, const gchar *name, gpointer user_data);
+
+  static gboolean on_handle_device_get_available_list(NpudCore *object,
+                                                      GDBusMethodInvocation *invocation);
+
+private:
+  guint _gdbus_id;
+  static std::atomic_bool _isReady;
```
ragmani, line 50: Just my curious question.
Why should this member be static?

ragmani:
I don't mind if this pr merges. The above questions are just my curiosity.

runtime/service/npud/core/DBus.h
jyoungyun: Because this member uses in static function. It is used in `DBus::on_bus_acquired` static function.

runtime/service/npud/core/DBus.cc
```diff
@@ -0,0 +1,89 @@
...
+  iface->handle_device_get_available_list = &on_handle_device_get_available_list;
+
+  if (!g_dbus_interface_skeleton_export(G_DBUS_INTERFACE_SKELETON(core), conn, "/org/tizen/npud",
+                                        &error))
+  {
+    VERBOSE(DBus) << "Failed to export skeleton, Server will stop." << std::endl;
+    Server::instance().stop();
+  }
+
+  _isReady.exchange(true);
```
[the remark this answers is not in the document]
jyoungyun: The `exchange` command is a read-modify-write operation. It means that this command performs read-modify-write with one command. It prevents race conditions in multi-threaded application.

https://stackoverflow.com/questions/61391912/why-do-people-use-atomic-exchange-instead-of-store

runtime/service/npud/core/DBus.h
ragmani: My question was meant to include that static function. The function is used only in constructor of `class DBus`, so I don't understand why the function and the member should be static.

jyoungyun: DBus callback should be static function. So the member variable should be static too.
reference: http://www.manpagez.com/html/gio/gio-2.28.6/gio-Owning-Bus-Names.php#GBusAcquiredCallback

runtime/service/npud/core/DBus.cc
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "DBus.h"
+#include "Server.h"
```
ragmani, line 18: It's "Circular Dependency". I think it should be resolved.

runtime/service/npud/core/DBus.h
ragmani: I found why the callback function should be static. It's because of this pointer of DBus instance.
In other words, the static function is the type `void (*)(GDBusConnection *conn, const gchar *name, gpointer user_data)`, but `void(DBus::*)(GDBusConnection *conn, const gchar *name, gpointer user_data)` if the function is not static.

I'm sorry to bother you, but as static was added, the implementation of `DBus` was strange, so I asked. I thought It's likely that you may have considered the case where `DBus` instances are created multiple times. Anyway LGTM enough for now as is.
And it seems that `_isReady` doesn't have to be `std::atomic_bool` if there is no case where `_isReady` is set more than once.

Issue #9931, opened 2022-10-27
[luci] Fix CircleUnidirectionalSequenceLSTM


seanshpark:
While working with `CircleUnidirectionalSequenceLSTM` in luci-interpreter, I've found that
- `CircleUnidirectionalSequenceLSTM` has wrong order in import.
- `activation_state` renamed to `output_state` in tflite

let's revise this.

seanshpark:
related #9899

seanshpark:
done

PR #9932, opened 2022-10-27
[luci/lang] Introduce output_state for LSTM

seanshpark:
This will introduce output_state attribute in CircleUnidirectionalSequenceLSTM to replace activation_state attribute.


seanshpark:
for #9931

mhs4670go (approved):
LGTM

PR #9933, opened 2022-10-27
[luci/import] Fix import LSTM

seanshpark:
This will fix CircleUnidirectionalSequenceLSTM import with order of inputs.


seanshpark:
for #9931

mhs4670go (approved):
LGTM

PR #9937, opened 2022-10-27
[luci] Use LSTM output_state

seanshpark:
This will revise to use output_state instead of activation_state.


mhs4670go (approved):
LGTM

PR #9939, opened 2022-10-27
[luci-interpreter] Relocate getTfLiteActivation

seanshpark:
This will relocate get_tflite_activation method as getTfLiteActivation from SVDF kernel to Utils. Method name is changed to match naming convention of Utils.


seanshpark:
for https://github.com/Samsung/ONE/issues/9895#issuecomment-1290152184 , from  #9899

BalyshevArtem (approved):
LGTM

PR #9943, opened 2022-10-27
[infra/onert-micro] Add armv8-m generic options

BalyshevArtem:
This PR adds options file for armv8-m architecture.


binarman (approved):
LGTM

Issue #9646, opened 2022-08-30
[onert] Support simply partitioned nnpackage


hseok-oh:
Simply partitioned multimodel NN package

- Package has two model
- Package's all inputs = 1st model's all inputs (same order)
- Package's all outputs = 2nd model's all outputs (same order)
- 1st model's all outputs = 2nd model's all inputs (same order)

_Originally posted by @hseok-oh in https://github.com/Samsung/ONE/issues/9610#issuecomment-1230007372_

hseok-oh:
I made test package from inception

MANIFEST
```
{
  "major-version" : "1",
  "minor-version" : "3",
  "patch-version" : "0",
  "configs"       : [  ],
  "models"        : [ "inception_v3.14to22.tflite", "inception_v3.23to32.tflite" ],
  "model-types"   : [ "tflite", "tflite" ],
  "pkg-inputs"    : [ "0:0:0", "0:0:1", "0:0:2", "0:0:3" ],
  "pkg-outputs"   : [ "1:0:0", "1:0:1", "1:0:2", "1:0:3" ],
  "model-connect" : [ { "from" : "0:0:0", "to" : [ "1:0:0" ] }, { "from" : "0:0:1", "to" : [ "1:0:1" ] }, { "from" : "0:0:2", "to" : [ "1:0:2" ] }, { "from" : "0:0:3", "to" : [ "1:0:3" ] } ]
}

```

hseok-oh:
Example package:
[two_tflites2.tar.gz](https://github.com/Samsung/ONE/files/9458406/two_tflites2.tar.gz)

hseok-oh:
Next step: one-way models package

```
inputs -> model 1 -> model 2 -> ...  -> model N-1 -> model N -> outputs
```

all inputs -> model 1 inputs
all model 1 outputs -> model 2 inputs
...
all model N-1 outputs -> model N inputs
model N outputs -> outputs

hseok-oh:
Done

PR #9951, opened 2022-10-28
[onert] Allow using I/O buffer for all models

hseok-oh:
This commit allows using I/O buffers for all models on multimodel package. 
Intermediate executor output buffer should be used for next model executor.

### DIFF
--- a/runtime/onert/core/src/exec/Executors.cc
+++ b/runtime/onert/core/src/exec/Executors.cc
@@ -97,99 +97,64 @@ void Executors::checkSupportedMultimodel() const
   //  (n)th model: output assumption
   //  Otherwise: edges assumption
 
-  // Assumption: executor input/output size
-  //  Models size: n
-  //  sizeof(model(x).outputs) == sizeof(model(x+1).inputs) if 0 < x < n-2
-  //  sizeof(model(0).inputs) == sizeof(package.inputs)
-  //  sizeof(moddel(n-1).outputs) == sizeof(package.outputs)
-  for (auto &pair : _executors)
-  {
-    auto const &model_index = pair.first.first;
-    auto const &subg_index = pair.first.second;
-
-    // Skip inner subgraph
-    if (subg_index != ir::SubgraphIndex{0})
-      continue;
-
-    auto executor = at(model_index, subg_index);
-
-    // Last model's output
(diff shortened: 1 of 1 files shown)


hseok-oh:
@chunseoklee Could you check this PR works on example model?

chunseoklee:
> @chunseoklee Could you check this PR works on example model?

I have to make a new nnpkg w/o need of type conversion(uint8->f32). Current target model consists of 1) trix model and 2) f32 circle model. I will update the result after creating new nnpkg. 

Anyway, it works w/o error with current model though it is weird that the size mismatch between trix's output and  circle's input does not occur any segfault.

hseok-oh:
I confirmed that previous test packages (https://github.com/Samsung/ONE/issues/9646#issuecomment-1232445122, https://github.com/Samsung/ONE/issues/9655#issuecomment-1280216480) are passed.

chunseoklee:
>  I will update the result after creating new nnpkg.

Execution passed. Need time to create verification value

runtime/onert/core/src/exec/Executors.cc
```diff
@@ -97,99 +97,64 @@ void Executors::checkSupportedMultimodel() const
...
-      if (std::find_if(edges.begin(), edges.end(), check_from_edge) == edges.end())
-      {
-        throw std::runtime_error{"NYI: Unsupported model edge pattern"};
-      }
-    }
-  }
-
   // Assumption: edges
-  //   x:0:z -> x+1:0:z  (0 < x < n-2)
+  //   x:0:z -> x+1:0:z'
```
glistening, line 101:
```suggestion
  //   x:0:z -> x+1:0:z
```

glistening (approved):
LGTM

runtime/onert/core/src/exec/Executors.cc
hseok-oh: `from.IOIndex` and `to.IOIndex` can be different on this commit.

jyoungyun (approved):

chunseoklee:
> Need time to create verification value

Value matched with nnpkg of two F32 circle models.

chunseoklee (approved):

Issue #8170, opened 2021-12-14
[circle-opselector] Implement main logic


mhs4670go:
## What

Let's implement `circle-opselector`'s main logic.

## Why

There is a draft(#7600) that implements overall `circle-opselector`'. However, the implementer stopped the implementation for some reason. Moreover, it was a flatbuffers-aware(?) implementation, which a bit overlaps with the other module. We already have `luci` module that addresses conversion from flatbuffers contents to circle IR. 

So, let's implement `circle-opselector`'s main logic with this. 

Related: #7596

PR #9954, opened 2022-10-28
[circle-opselector] Implement circle-opselector main logic

mhs4670go:
This commit is to implement circle-opselector main logic.

Related: #8170 


mhs4670go:
The value test will be added soon.

compiler/circle-opselector/driver/Driver.cpp
```diff
@@ -255,10 +365,17 @@ int entry(int argc, char **argv)
...
+  auto new_module = std::make_unique<luci::Module>();
+  // make graph with selected nodes
+  auto new_graph = ::make_graph(selected_nodes);
+  // TODO set proper graph name
+
+  new_module->add(std::move(new_graph));
 
   // Export to output Circle file
-  assert(opselector::exportModule(module.get(), output_path));
+  assert(opselector::exportModule(new_module.get(), output_path));
```
glistening, line 378: (beyond this PR) It would be better to show explicit error if there is a problem while exporting module. Otherwise, the user will try something more with wrong circle. It could be done in another PR.

compiler/circle-opselector/driver/Driver.cpp
```diff
@@ -153,6 +155,107 @@ std::vector<std::string> split_name_input(const std::string &str)
...
+      if (ctx_it != ctx.end())
+        continue;
+      // the node isn't graph input if it is an other node's input
+      if (std::find(nodes.begin(), nodes.end(), arg) != nodes.end())
+        continue;
+      auto circle_const = dynamic_cast<luci::CircleConst *>(arg);
+      if (circle_const == nullptr)
+      {
+        // circle input
+        auto circle_input = graph->nodes()->create<luci::CircleInput>();
```
seanshpark, line 187: how is `NOT CircleConst` sure `IS CircleInput` ?

compiler/circle-opselector/driver/Driver.cpp
```diff
@@ -153,6 +155,107 @@ std::vector<std::string> split_name_input(const std::string &str)
...
+      auto input_node = dynamic_cast<luci::CircleNode *>(arg);
+      auto ctx_it = ctx.find(input_node);
+      // check if the node already has been cloned
+      if (ctx_it != ctx.end())
+        continue;
+      // the node isn't graph input if it is an other node's input
+      if (std::find(nodes.begin(), nodes.end(), arg) != nodes.end())
+        continue;
+      auto circle_const = dynamic_cast<luci::CircleConst *>(arg);
+      if (circle_const == nullptr)
```
seanshpark, line 184: IMHO it would better use like `if (circle_const != nullptr) { ... }` for `CircleConst`

compiler/circle-opselector/driver/Driver.cpp
```diff
@@ -153,6 +155,107 @@ std::vector<std::string> split_name_input(const std::string &str)
...
+        ctx.emplace(circle_const, clone);
+      }
+    }
+  }
+  // set graph output
+  for (const auto &n : nodes)
+  {
+    auto outputs = loco::succs(n);
+    for (const auto &o : outputs)
+    {
```
seanshpark, line 221: I don't understand this block. Can you please explain what your intention is?

compiler/circle-opselector/driver/Driver.cpp
seanshpark: Sorry, incorrectly read the code -_-;

mhs4670go: @seanshpark Let me explain overall logic.

Parameter `nodes` are selected nodes from user that will be another new graph.

1. First, I cloned each nodes.
```cxx
// clone nodes
  for (const auto &n : nodes)
  {
    auto clone = luci::clone_node(n, graph.get());
    ctx.emplace(n, clone);
  }
```
2. Next, list candidates that could be a graph input.

We may check below conditions to know who is graph input.
- Graph input is a virtual node, which means it wasn't cloned at first step.
```cxx
if (ctx_it != ctx.end())
        continue;
```
- Graph input is not others' input.
```cxx
if (std::find(nodes.begin(), nodes.end(), arg) != nodes.end())
        continue;
```

3. All the nodes that went through above conditions can be a graph input now except one thing; const.

For example, say the node is `CircleConv2D`. it has three inputs - `input`, `filter`, `bias`.

And, only `input` should be a graph input. `filter` and `bias` will be just `CircleConst` nodes.

compiler/circle-opselector/driver/Driver.cpp
mhs4670go: @seanshpark This block finds the graph output like above finding graph input logic does.

The output node is not others' output.
```cxx
if (std::find(nodes.begin(), nodes.end(), o) != nodes.end())
        continue;
```

All the nodes that went through above condition will be graph outputs.

compiler/circle-opselector/driver/Driver.cpp
mhs4670go: @glistening Actually, there is explicit condition in the function.
```cxx
bool exportModule(luci::Module *module, std::string &output_path)
{
  luci::CircleExporter exporter;

  luci::CircleFileExpContract contract(module, output_path);

  if (!exporter.invoke(&contract))
  {
    std::cerr << "ERROR: Failed to export '" << output_path << "'" << std::endl; // NOTE HERE
    return false;
  }

  return true;
}
```

Strictly speaking, `assert` is useless. I'll remove it.

seanshpark (approved):
LGTM

Issue #9968, opened 2022-11-02
[loco] unit test segmentation fault 


seanshpark:
Runing `loco_test` as-is is OK, but running single test fails

```
cd build/compiler/loco

./loco_test --gtest_filter="*consturctor_with_param_node*"

Note: Google Test filter = *consturctor_with_param_node*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from GraphTest
[ RUN      ] GraphTest.consturctor_with_param_node
[1]    1556778 segmentation fault  ./loco_test --gtest_filter="*consturctor_with_param_node*"
```

code
```cxx
  ASSERT_NO_THROW(g->nodes()->destroy(test_node));
  ASSERT_THROW(g->nodes()->destroy(test_node), std::invalid_argument);  <-- THIS
```

Also this may fail too
```
./loco_test --gtest_filter="*create_and_destroy_node*"
```

seanshpark:
Found from https://github.com/Samsung/ONE/issues/9962#issuecomment-1297947575
- Ubuntu 22.04 fails with `./loco_test` so cannot pass test / CI
- Ubuntu 18.04/22.04 is success with `./loco_test` but fails with running single test

seanshpark:
What is happening

&nbsp; | ./loco_test | create_and_destroy_node | consturctor_with_param_node
--- | --- | --- | ---
U1804 | OK | Segmentation fault | Segmentation fault
U2204 | Segmentation fault | OK | Segmentation fault

where `create_and_destroy_node ` is
```
./loco_test --gtest_filter=*create_and_destroy_node
```
`consturctor_with_param_node` is
```
./loco_test --gtest_filter=*consturctor_with_param_node
```

seanshpark:
Anyway, as `destroy(node)` will delete node object, using already deleted object seems unsafe.

mhs4670go:
```cxx
g->nodes()->destroy(pull);
g->nodes()->destroy(pull); // seg fault
```
but
```cxx
g->nodes()->destroy(pull);
g->nodes()->destroy(nullptr);
g->nodes()->destroy(pull); // end successfully
```
..???

seanshpark:
> ..???

o_O;;;

mhs4670go:
After doing some research, the segfault occurs when the `destroy` method is called.
```cxx
void destroy(Node *node)
  {
    if (!ObjectPool<Node>::erase(node))
    {
[...]

PR #9969, opened 2022-11-02
[loco] Fix test not to double destroy with node

seanshpark:
This will fix unit test not to double destory with a single node object.

### DIFF
--- a/compiler/loco/src/IR/Graph.test.cpp
+++ b/compiler/loco/src/IR/Graph.test.cpp
@@ -79,6 +79,18 @@ TEST(GraphTest, create_and_destroy_node)
   auto pull = g->nodes()->create<loco::Pull>();
 
   ASSERT_NO_THROW(g->nodes()->destroy(pull));
+}
+
+TEST(GraphTest, DISABLED_create_and_destroy_node_again)
+{
+  auto g = loco::make_graph();
+
+  auto pull = g->nodes()->create<loco::Pull>();
+
+  ASSERT_NO_THROW(g->nodes()->destroy(pull));
+  // NOTE calling destroy again raises Segmentation fault
+  //      refer https://github.com/Samsung/ONE/issues/9968
+  // TODO fix this
   ASSERT_THROW(g->nodes()->destroy(pull), std::invalid_argument);
 }
 
@@ -139,7 +151,6 @@ TEST(GraphTest, consturctor_with_param_node)
   ASSERT_FLOAT_EQ(test_node->f(), 11.11);
 
   ASSERT_NO_THROW(g->nodes()->destroy(test_node));
-  ASSERT_THROW(g->nodes()->destroy(test_node), std::invalid_argument);
 }
 
 TEST(GraphTest, getters_over_const_instance)


compiler/loco/src/IR/Graph.test.cpp
```diff
@@ -139,7 +143,6 @@ TEST(GraphTest, consturctor_with_param_node)
   ASSERT_FLOAT_EQ(test_node->f(), 11.11);
 
   ASSERT_NO_THROW(g->nodes()->destroy(test_node));
-  ASSERT_THROW(g->nodes()->destroy(test_node), std::invalid_argument);
```
seanshpark, line 142: no need to check this here again

seanshpark:
for #9968

compiler/loco/src/IR/Graph.test.cpp
```diff
@@ -79,7 +79,11 @@ TEST(GraphTest, create_and_destroy_node)
   auto pull = g->nodes()->create<loco::Pull>();
 
   ASSERT_NO_THROW(g->nodes()->destroy(pull));
-  ASSERT_THROW(g->nodes()->destroy(pull), std::invalid_argument);
```
parjong, line 82: ``nodes()`` returns NodeContext

https://github.com/Samsung/ONE/blob/cd3ae1e46dd947e320d4bced110effa4879b1dbe/compiler/loco/include/loco/IR/Graph.h#L218

NodeContext is an alias of NodePool:
https://github.com/Samsung/ONE/blob/cd3ae1e46dd947e320d4bced110effa4879b1dbe/compiler/loco/include/loco/IR/Graph.h#L170

``NodePool::destroy`` internally invokes ``ObjectPool::erase``:
https://github.com/Samsung/ONE/blob/cd3ae1e46dd947e320d4bced110effa4879b1dbe/compiler/loco/include/loco/IR/NodePool.h#L44-L50

``ObjectPool::erase`` internally checks the existence of node..
https://github.com/Samsung/ONE/blob/cd3ae1e46dd947e320d4bced110effa4879b1dbe/compiler/loco/include/loco/ADT/ObjectPool.h#L52-L69

compiler/loco/src/IR/Graph.test.cpp
```diff
@@ -79,7 +79,11 @@ TEST(GraphTest, create_and_destroy_node)
...
+  // NOTE calling g->nodes()->destroy(pull) again will raise segmentation fault
+
+  auto g2 = loco::make_graph();
+  auto pull2 = g2->nodes()->create<loco::Pull>();
+  ASSERT_THROW(g->nodes()->destroy(pull2), std::invalid_argument);
 }
```
parjong, line 87: Hmm 🤔  the logic itself is written with the consideration of double free. So, this test tries to test that. If this line cause segmentation fault. It would be better to disable this test temporary by prepending ``DISABLED_`` to test name instead of modifying the test itself.

seanshpark: OK, thanks for the comment.
I'll split this test as (1) just call destroy once (2) add another test with double destroy but disabled.

parjong (approved):

Issue #9957, opened 2022-10-31
[onert/cpu] ReduceSum uint8 test fail on CPU backend.


hseok-oh:
### What?

After updating test driver to use TFLite 2.8, CPU backend test fails. 
We need to resolve this issue.

_Originally posted by @hseok-oh in https://github.com/Samsung/ONE/pull/9956#discussion_r1008990188_

hseok-oh:
I found out `acl-cl` and `acl-neon` backend didn't run ReduceSum uint8 test. Maybe there was a bug on old CPU kernel.

hseok-oh:
Small diff on test

```
$ ./Product/out/bin/tflite_comparator Product/out/test/models/cache/reduce_sum_uint8.tflite 
[Execution] Stage start!
[Execution] Model is deserialized!
[Execution] Model compiled!
[Execution] Input data is defined!
[Execution] Done!
[Comparison] Stage start!
INFO: Initialized TensorFlow Lite runtime.
[Comparison] TFLite run done!
Output #0, Element Index : 0, ref: G, act: H (diff: 1)
Output #0, Element Index : 2, ref: �, act: � (diff: 1)
Output #0, Element Index : 4, ref: �, act: � (diff: 1)
Output #0, Element Index : 5, ref: , act:  (diff: 1)
Output #0, Element Index : 7, ref: �, act: � (diff: 1)
Output #0, Element Index : 8, ref: 3, act: 4 (diff: 1)
Output #0, Element Index : 9, ref: r, act: s (diff: 1)
Output #0, Element Index : 10, ref: �, act: � (diff: 1)
Output #0, Element Index : 11, ref: B, act: C (diff: 1)
Output #0, Element Index : 12, ref: �, act: � (diff: 1)
Output #0, Element Index : 13, ref: K, act: L (diff: 1)
Output #0, Element Index : 14, ref: F, act: G (diff: 1)
Output #0, Element Index : 15, ref: �, act: � (diff: 1)
Output #0, Element Index : 17, ref: �, act: � (diff: 1)
Output #0, Element Index : 18, ref: �, act:  (diff: 255)
Output #0, Element Index : 19, ref: R, act: S (diff: 1)
Output #0, Element Index : 20, ref: �, act: � (diff: 1)
[Comparison] outputs is not equal!
[Comparison] Done!
```

hseok-oh:
Resolved

PR #9976, opened 2022-11-03
[compute/cker] Fix Reduce kernel

hseok-oh:
This commit resolves quantized ReduceSum and ReduceSum kernel  bug.

---

Related issue: #9957

### DIFF
--- a/compute/cker/include/cker/operation/Reduce.h
+++ b/compute/cker/include/cker/operation/Reduce.h
@@ -312,12 +312,12 @@ public:
     }
 
     // Calculate mean by dividing output_data by num of aggregated element.
-    U num_elements_in_axis = 1;
+    size_t num_elements_in_axis = 1;
     for (int idx = 0; idx < num_resolved_axis; ++idx)
     {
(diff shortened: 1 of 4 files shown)


glistening (approved):
LGTM

compute/cker/include/cker/operation/Reduce.h
```diff
@@ -330,7 +330,7 @@ class Reduce
       if (compute_sum)
       {
         // TODO(b/116341117): Eliminate float and do this completely in 8bit.
-        const float bias = -input_zero_point * scale * num_elements_in_axis + 0.5f;
```
ragmani, line 333: Just curious.
Why was there `+ 0.5f` in the old code? `bias` is used only for `std::round`, not `std::floor`. It's strange.

ragmani (approved):
LGTM

compute/cker/include/cker/operation/Reduce.h
hseok-oh: I also don't know reason. TFLite 1.13 uses `+0.5f`, and it is removed now.

PR #9979, opened 2022-11-03
[common-artifacts] Enable UnidirectionalSequenceLSTM_001

seanshpark:
This will enable test for UnidirectionalSequenceLSTM_001 as it is now possible to run UnidirectionalSequenceLSTM Op.
- UnidirectionalSequenceLSTM_000 is not enabled this is just for creation of this Op


seanshpark:
for #9895 from #9899

mhs4670go (approved):
LGTM

PR #9987, opened 2022-11-07
[cmn-arts] Note for python find_package

seanshpark:
This will add notes about find_package python with platform versions.


seanshpark:
cc @mhs4670go

jinevening (approved):
LGTM

PR #9990, opened 2022-11-07
[npud] Add Core Class

jyoungyun:
This commit adds Core class.
DBus method requests to the core and the core returns the request execution result.

### DIFF
--- a/runtime/service/npud/core/Server.h
+++ b/runtime/service/npud/core/Server.h
@@ -44,7 +44,7 @@ public:
     return server;
   }
 
-  static std::shared_ptr<Core> &core(void) { return instance()._core; }
+  static std::shared_ptr<Core> core(void) { return instance()._core; }
 
 private:
   Server() noexcept;
--- a/runtime/service/npud/core/Server.h
+++ b/runtime/service/npud/core/Server.h
@@ -44,7 +44,7 @@ public:
     return server;
   }
 
-  static std::shared_ptr<Core> core(void) { return instance()._core; }
+  static const Core &core(void) { return *instance()._core.get(); }
 
 private:
   Server() noexcept;
@@ -53,7 +53,7 @@ private:
 
   std::unique_ptr<GMainLoop, void (*)(GMainLoop *)> _mainloop;
   std::unique_ptr<Signal> _signal;
-  std::shared_ptr<Core> _core;
+  std::unique_ptr<Core> _core;
   std::unique_ptr<DBus> _dbus;
 };
 
--- a/runtime/service/npud/core/Server.h
+++ b/runtime/service/npud/core/Server.h
@@ -44,7 +44,7 @@ public:
     return server;
   }
 
-  static const Core &core(void) { return *instance()._core.get(); }
+  const Core &core(void) { return *_core.get(); }
 
 private:
   Server() noexcept;
--- a/runtime/service/npud/core/Core.cc
+++ b/runtime/service/npud/core/Core.cc
@@ -14,50 +14,24 @@
  * limitations under the License.
  */
 
-#ifndef __ONE_SERVICE_NPUD_CORE_SERVER_H__
-#define __ONE_SERVICE_NPUD_CORE_SERVER_H__
-
-#include "Signal.h"
 #include "Core.h"
-#include "DBus.h"
-
-#include <glib.h>
-#include <memory>
-#include <atomic>
 
 namespace npud
 {
 namespace core
 {
 
-class Server
+void Core::init()
 {
-public:
-  void run(void);
-  void stop(void);
+  // TODO Implement details
+}
 
-  bool isRunning() { return _isRunning.load(); }
+void Core::deinit()
+{
+  // TODO Implement details
+}
 
-  static Server &instance(void)
-  {
-    static Server server;
-    return server;
-  }
-
-  static std::shared_ptr<Core> &core(void) { return instance()._core; }
-
-private:
-  Server() noexcept;
-
-  static std::atomic_bool _isRunning;
-
-  std::unique_ptr<GMainLoop, void (*)(GMainLoop *)> _mainloop;
-  std::unique_ptr<Signal> _signal;
-  std::shared_ptr<Core> _core;
-  std::unique_ptr<DBus> _dbus;
-};
+int Core::getAvailableDeviceList(std::vector<std::string> &list) { return 0; }
 
 } // namespace core
 } // namespace npud
-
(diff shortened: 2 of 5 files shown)


hseok-oh (approved):

runtime/service/npud/core/Server.h
```diff
@@ -43,13 +44,16 @@ class Server
     return server;
   }
 
+  static std::shared_ptr<Core> &core(void) { return instance()._core; }
```
ragmani, line 47: Returning `shared_ptr` by reference has the risk of deleting `Core` at the improper time because the reference count does not increase. 
And I don't think this function has to be static. So I suggest as below.
```suggestion
  std::shared_ptr<Core> core(void) { return _core; }
```

runtime/service/npud/core/Server.h
```diff
@@ -43,13 +44,16 @@ class Server
...
+
 private:
   Server() noexcept;
```
ragmani, line 51:
```suggestion
        Server(const Server &);
        void operator=(const Server &);
```
or
```suggestion
public:
        Server(const Server &) = delete;
        void operator=(const Server &) = delete;

```

runtime/service/npud/core/Server.h
hseok-oh: BTW, which class will require this public method?

jyoungyun: If this function is not a static function, when using this Core variable, we should use it like `Server::instance().core()`. I thought it was too long, so I used a static function to use it like this `Server::core()`.

runtime/service/npud/core/Server.h
jyoungyun: This change is not related to this PR. I will make a new PR for this.

runtime/service/npud/core/Server.h
ragmani: It's risky if this function is called without calling `instance()` because there is no instance of `Server`.

runtime/service/npud/core/Server.h
ragmani: OK. But I think it would be better to add disabling copy constructor and operator when using singleton pattern.

runtime/service/npud/core/Server.h
hseok-oh: I found usage on draft.
Basically, I agree with @ragmani. Aditionally, we can use constant reference of `_core` on this case IMO

```suggestion
  const Core &core(void) { return _core; }
```

jyoungyun: > It's risky if this function is called without calling `instance()` because there is no instance of `Server`.

I don't understand. There is no case to call `core()` function without Server instance because the `core` function already calls the `instance()` function inside.

jyoungyun: > if we don't allow to share ownership.

Yes, I agree with your comments. There is no need to share ownership of Core. I will update it.

ragmani: @jyoungyun 
I missed `instance()._core`. It's not risky. If you want to use this function as static, it looks fine. But returning `share_ptr` by reference has the risk explained at https://github.com/Samsung/ONE/pull/9990#discussion_r1015292113. 

> we can use constant reference of _core on this case IMO if we don't allow to share ownership.

I worried about removing ownership of `_core` from the outside, not sharing owner. But returning const reference can also disable removing ownership of `_core`. I think It's a good way.
Additionally, we can use const pointer of `_core` if we don't allow to sharing ownership and removing ownership.
To summarize,
it would be better to use returning reference if we allow only sharing ownership or returning const reference, 
or use returning reference if we allow only sharing ownership.

jyoungyun: I understood the ownership related `Core` object. However, I don't agree with the reason why the `Core()` function should not be used as a static function.

runtime/service/npud/core/Server.h
jyoungyun: Related PR: #9995

runtime/service/npud/core/Server.h
jyoungyun: @ragmani @hseok-oh Please see the latest patch.

ragmani: > However, I don't agree with the reason why the core() function should not be used as a static function.

Personally, I tend to avoid using static member methods. There are some inconveniences.
1. They lose the advantages of class like polymorphic, abstraction and inheritance.
2. It is difficult to deal with static methods because static methods belong to the class and not a particular instance. If a class has static member method and non-static member attributes, the class has two responsibilities(be used without creating an instance, and be used after an instance is created). It means that static member methods can break the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle). 

In here, I was confused because static member method(`core()`) uses non-static member attribute(`_core`). Static methods can be called without creating an instance, but non-static attributes exist only within an instance. This is not only ambiguous which instance the member attribute came from, but it can also be a dangerous call if the class is not a singleton pattern.
I still prefer to use the method as non-static here. Nevertheless, I think it would be fine to use it here if you are going to not create multiple instances of `class Server` and not take the advantages of class.

ragmani (approved):
As described in https://github.com/Samsung/ONE/pull/9990#discussion_r1016218461, I prefer not to use static, but the current code is also LGTM.

runtime/service/npud/core/Server.h
jyoungyun: @ragmani 

> Static methods can be called without creating an instance, but non-static attributes exist only within an instance.

The `core()` static function calls the `instance()` static function inside it. So we don't need to worry that this function is called without creating an instance..

> This is not only ambiguous which instance the member attribute came from

I agree with this point.

> but it can also be a dangerous call if the class is not a singleton pattern.
>  if you are going to not create multiple instances of `class Server`

Is it possible to create a multiple `Server` in this program?


I'm sorry, but I still don't understand what your concern is. :( However I agree that the static function may be ambiguous. I will update according to your suggestion. :)

ragmani: > Is it possible to create a multiple `Server` in this program?

If `Server` keep using the singleton pattern, it probably won't.

ragmani:
@nnfw-bot test tizen-gbs-builder

ragmani (approved):

jyoungyun:
@nnfw-bot test tizen-gbs-builder

hseok-oh (approved):
LGTM

hseok-oh:
@nnfw-bot test tizen-gbs-builder

PR #9997, opened 2022-11-08
[visq] Add tests for different fp32/fq data

jinevening:
This adds tests for different fp32/fq data.


jinevening:
CC @mhs4670go

seanshpark (approved):
LGTM

PR #9998, opened 2022-11-09
[one-cmds] Fix trusted host name

seanshpark:
This will fix trusted host name for files.pythonhosted.org


mhs4670go (approved):
LGTM

PR #10005, opened 2022-11-09
[onert] Remove compiler state variable

hseok-oh:
This commit removes compiler state variable setting on Compiler. It is used on NNAPI only, and it can be maintained on frontend.


ragmani (approved):
LGTM

glistening (approved):
LGTM

PR #10008, opened 2022-11-09
[Draft][circle2circle] Add RemoveDuplicateConstPass

BalyshevArtem:
This PR adds pass to remove duplicate const nodes from model.

for issue:  #10009


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -108,6 +108,7 @@ int entry(int argc, char **argv)
              "when the impact is known to be acceptable.");
   add_switch(arser, "--fuse_preactivation_batchnorm",
              "This will fuse BatchNorm operators of pre-activations to Convolution operator");
+  add_switch(arser, "--remove_duplicate_const", "This will remove all duplicate constant nodes");
```
seanshpark, line 111: please split `circle2circle` codes and `luci` codes in separate PRs.

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+        // Remove from next checking
+        cur_pair.second[i] = nullptr;
+
+        changed = true;
+      }
+    }
+  }
+  if (changed)
+    printf("ggg\n");
```
seanshpark, line 126: this seems unintentionally added

seanshpark:
it would be nice to have a link to some issue for this PR.

seanshpark:
https://github.com/Samsung/ONE/blob/master/compiler/luci/export/src/CircleTensorExporter.cpp#L559-L563 was introduced to remove CircleConst having same values. please take a look.
it would be nice if you can add comments in the issue(if you have one or add a new one) for your condition.

BalyshevArtem:
> https://github.com/Samsung/ONE/blob/master/compiler/luci/export/src/CircleTensorExporter.cpp#L559-L563 was introduced to remove CircleConst having same values. please take a look. it would be nice if you can add comments in the issue(if you have one or add a new one) for your condition.

As I understand  https://github.com/Samsung/ONE/blob/master/compiler/luci/export/src/CircleTensorExporter.cpp#L559-L563 creates new tensor but with the same buffer, and in the process of importing the model, new constant nodes are still created for each constant tensor, where the data from the buffer is copied, and during execution we will have duplicate nodes. 

Idea of this pass to remove from model such duplicate nodes, and propagate unique constan node for all places, where originals  were used. And during execution in will be used only one unique node.

seanshpark:
> P.S. I created small example model with duplicate constant:
example_model.tar.gz

to be clear, can you please explain detail of this example?

anyway, with `circledump` tool, I could see:
```
Buffers: B(index) (length) values, if any
B(0) (0)
B(1) (0)
B(2) (0)
B(3) (0)
B(4) (4) 0xff 0xff 0xff 0xff
B(5) (4) 00 00 00 00
B(6) (0)
B(7) (0)
B(8) (48) 0x3 00 00 00 00 00 00 00 0x4 00 00 00 0x41 0x64 0x64 00 ...
B(9) (4) 00 00 00 00

Operands: T(subgraph index : tensor index) TYPE (shape) (shape_signature) B(buffer index) (variable) OperandName
T(0:0) UINT8 (1, 32, 32, 1) B(1) Placeholder
    Quantization: scale(0.039203) zeropt(128) quantized_dimension(0)

T(0:1) UINT8 (1, 32, 32, 1) B(2) Placeholder_1
    Quantization: scale(0.0391837) zeropt(128) quantized_dimension(0)

T(0:2) UINT8 (1, 32, 32, 1) B(3) Add
    Quantization: scale(0.0774637) zeropt(127) quantized_dimension(0)

T(0:3) UINT8 (1, 2, 2, 1) B(4) Const
    Quantization: scale(0.00392157) zeropt(0) quantized_dimension(0)

T(0:4) INT32 (1) B(5) Conv2D_bias
    Quantization: scale(0.000303779) zeropt(0) quantized_dimension(0)

T(0:5) UINT8 (1, 31, 31, 1) B(6) Conv2D
    Quantization: scale(0.222277) zeropt(128) quantized_dimension(0)

T(0:6) UINT8 (1, 2, 2, 1) B(4) Const_6
    Quantization: scale(0.00392157) zeropt(0) quantized_dimension(0)

T(0:7) INT32 (1) B(5) Conv2D_1_bias
    Quantization: scale(0.000871676) zeropt(0) quantized_dimension(0)

T(0:8) UINT8 (1, 30, 30, 1) B(7) Conv2D_1
    Quantization: scale(0.765644) zeropt(127) quantized_dimension(0)
```

`B(5)` and `B(9)` looks same but I can't find the tensor using `B(9)`.
I think there exist some other issue with this circle file that created it.

Can you please share the original file, like a `.tflchef` or `.tflite` ?

seanshpark:
> ew constant nodes are still created for each constant tensor, where the data from the buffer is copied, and during execution we will have duplicate nodes.

Ah, I see... purpose of removing same content tensors was to reduce file size.
You are to change the model itself to reduce in-memory size of the model.

I think there is some difference of using circle IR here.
- (1) _luci_ circle IR is designed to make things easy for manipulating model graph
- (2) it was NOT designed for memory efficency

I assume background of this PR is for memory efficiency of `onert-micro`.
I think introducing `RemoveDuplicateConstPass` itself seems OK but not sure with relation with other Passes and for backends as I can't imagine as of now.

IMHO, I think it would be better to customize _luci_ in `onert-micro` for this purpose in the long run.

BalyshevArtem:
> I assume background of this PR is for memory efficiency of `onert-micro`

Yes :)



> Can you please share the original file, like a `.tflchef` or `.tflite`

[example_copy.tfgex.txt](https://github.com/Samsung/ONE/files/9979116/example_copy.tfgex.txt)

seanshpark:
I don't think `example_copy` model is anywhere used in this repo.

BalyshevArtem:
> I don't think `example_copy` model is anywhere used in this repo.

This model is my example with duplicate constants to experiment with this pass.

Also with our experiments with tilling for onert-micro I faced with models with duplicate weights, so this pass can be helpful for such situations, as it can save memory.

seanshpark:
> This model is my example with duplicate constants to experiment with this pass.

I don't think you understand what I am saying. I can catch the idea but this repo has no relation with this file `.tfgex`.

seanshpark:
DO NOT USE FILES FROM INTERNAL BACKEND REPO!!!

seanshpark:
I've DELETED the comments about the file.

BalyshevArtem:
> I've DELETED the comments about the file.

[model.tflite.txt](https://github.com/Samsung/ONE/files/9979621/model.tflite.txt)

res/TensorFlowLiteRecipes/Net_Duplicate_Weights_000/test.rule
```diff
@@ -0,0 +1,5 @@
+# To check if RemoveDuplicateConstPass removes all duplicate consts
+
+RULE "VERIFY_FILE_FORMAT" $(verify_file_format) '=' 1
+
+RULE "CIRCLE_CONST_COUNT" $(op_count CONST) '=' 2
```
BalyshevArtem, line 5: @seanshpark,

I have added rule for circle2circle dredd test, but  i don't know what rule I should use here, could you please help, how I can count CONST node, because `RULE "CIRCLE_CONST_COUNT" $(op_count CONST) '=' 2` doesn't work, due to CONST is not an operation, as I understand

seanshpark: dredd test uses `circle-inspect` tool to print operators but as CircleConst is not an operator, it doesn't show it.
I'm not sure which may be the best solution but I can think is to add an option to `circle-inspect` to print `CONST` for each `CircleCons`.
We need to revise
- `circle-inspect` to add option something like `--constants`
- `dredd-rule-lib/rule-lib.sh` to add something like `const_count`

If you like, you can change them or wait for someone to do it.
I'll post an issue for this.

BalyshevArtem: > dredd test uses `circle-inspect` tool to print operators but as CircleConst is not an operator, it doesn't show it. I'm not sure which may be the best solution but I can think is to add an option to `circle-inspect` to print `CONST` for each `CircleCons`. We need to revise
> 
>     * `circle-inspect` to add option something like `--constants`
> 
>     * `dredd-rule-lib/rule-lib.sh` to add something like `const_count`

Done

binarman:
@seanshpark 
Sorry for long reply

My opinion is following: this pass could be useful for both onert-micro runtime and for original onert.
- this optimization could reduce memory consumption, which is essential for micro runtime
- this optimization could improve data locality, slightly improving performance of original onert (we didn't test it though)

binarman:
> IMHO, I think it would be better to customize luci in onert-micro for this purpose in the long run.

Could you clarify, what do you mean by this?
Do you propose to put source code of this pass into onert-micro directory?

seanshpark:
>> IMHO, I think it would be better to customize luci in onert-micro for this purpose in the long run.
>
> Could you clarify, what do you mean by this?
Do you propose to put source code of this pass into onert-micro directory

@binarman , circle IR itself is not memory efficient structure, as it was designed for compiler with better for modification of the graph, not for any kind of runtime such as onert-micro. So in the long-run, I thought of slim down the _luci_ codes copied into onert-micro.

BalyshevArtem:
everything is merged, close it

Issue #10009, opened 2022-11-09
[circle2circle] Create pass to remove duplicate consant nodes


BalyshevArtem:
### What

Let's create circle2circle pass to remove duplicate constant nodes. This pass will delete all similar constant nodes and propagate the only one for all places, where original one was used. 

### Why

There are some models that have constant nodes, which values is the same, so they are duplicate. Deleting such nodes can save memory, both the runtime memory and the size of circle file, which is very important for systems with a small amount of memory, for example MCU.  

### Note

https://github.com/Samsung/ONE/pull/10008#issuecomment-1308875134

seanshpark:
For better maintenance of this Pass, this is not mandatory but highly recommended,
it would be nice to have some `circle2circle-dredd-recipe-test`.
could you please add them (recipe and others) to the draft?

BalyshevArtem:
done

PR #10010, opened 2022-11-09
[luci/pass] Add RemoveDuplicateConstPass to luci/pass

BalyshevArtem:
This PR adds pass to remove duplicate consts node from model to luci.

for issue: https://github.com/Samsung/ONE/issues/10009
draft: #10008

### DIFF
--- a/compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
+++ b/compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
@@ -126,7 +126,7 @@ bool RemoveDuplicateConstPass::remove_duplicate_const()
             is_equal = is_equal_consts<loco::DataType::U8>(reference_const, cur_const);
             break;
           default:
-            throw std::runtime_error("Not supported type\n");
+            continue;
         }
 
         if (not is_equal)
@@ -166,8 +166,28 @@ void RemoveDuplicateConstPass::add_to_map(luci::CircleConst *const_node)
   }
 }
 
-/*
+/**
  * Remove duplicate Const nodes.
+ *
+ * BEFORE
+ *    [CircleNode]   [CircleConst]
+ *          |        /
+ *          |      /
+ *    [CircleNode]    [CircleConst]
+ *          |        /
+ *          |      /
+ *    [CircleNode]
+ *
+ * AFTER
+ *
+ *    [CircleNode]   [CircleConst]
+ *          |        /     /
+ *          |      /     /
+ *    [CircleNode]     /
+ *          |        /
+ *          |      /
+ *    [CircleNode]
+ *
  */
 bool RemoveDuplicateConstPass::run(loco::Graph *g)
 {
--- a/compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
+++ b/compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
@@ -21,24 +21,29 @@
 namespace
 {
 
-template <loco::DataType DT> bool is_equal_consts(luci::CircleConst *left, luci::CircleConst *right)
+bool compare_quant_params(luci::CircleConst *left, luci::CircleConst *right)
 {
-  const auto left_rank = left->rank();
-  const auto right_rank = right->rank();
-
   const auto left_quant_param = left->quantparam();
   const auto right_quant_param = right->quantparam();
 
   if (left_quant_param != nullptr and right_quant_param != nullptr)
   {
     if (left_quant_param->scale != right_quant_param->scale or
-        left_quant_param->quantized_dimension != right_quant_param->quantized_dimension)
+        left_quant_param->quantized_dimension != right_quant_param->quantized_dimension or
+        left_quant_param->zerop != right_quant_param->zerop or
+        left_quant_param->min != right_quant_param->min or
+        left_quant_param->max != right_quant_param->max)
+    {
       return false;
-
-    if ((left->dtype() == loco::DataType::U8 or left->dtype() == loco::DataType::S8) and
-        left_quant_param->zerop != right_quant_param->zerop)
-      return false;
+    }
   }
+  return true;
+}
+
+bool compare_dim_values(luci::CircleConst *left, luci::CircleConst *right)
+{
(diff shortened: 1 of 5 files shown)


seanshpark:
I think we need some unit test for this Pass.

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
```diff
@@ -0,0 +1,186 @@
...
+  const auto left_rank = left->rank();
+  const auto right_rank = right->rank();
+
+  const auto left_quant_param = left->quantparam();
+  const auto right_quant_param = right->quantparam();
+
+  if (left_quant_param != nullptr and right_quant_param != nullptr)
+  {
+    if (left_quant_param->scale != right_quant_param->scale or
+        left_quant_param->quantized_dimension != right_quant_param->quantized_dimension)
```
jinevening, line 35: CircleQuantParam has more fields, e.g., min, max, zerop. It would be better to make a function to compare CircleQuantParam.

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
```diff
@@ -0,0 +1,186 @@
...
+      return false;
+
+    if ((left->dtype() == loco::DataType::U8 or left->dtype() == loco::DataType::S8) and
+        left_quant_param->zerop != right_quant_param->zerop)
```
jinevening, line 39: I think we can just compare zerop without consideration of dtype.

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  if (left_rank != right_rank)
+    return false;
+
+  for (uint32_t i = 0; i < left_rank; ++i)
+  {
+    if (left->dim(i).value() != right->dim(i).value())
+      return false;
+  }
```
jinevening, line 50: It would be possible to extract this as a function to reduce duplicate code from template.

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
```diff
@@ -0,0 +1,186 @@
...
+            is_equal = is_equal_consts<loco::DataType::S16>(reference_const, cur_const);
+            break;
+          case loco::DataType::S8:
+            is_equal = is_equal_consts<loco::DataType::S8>(reference_const, cur_const);
+            break;
+          case loco::DataType::U8:
+            is_equal = is_equal_consts<loco::DataType::U8>(reference_const, cur_const);
+            break;
+          default:
+            is_equal = false;
```
jinevening, line 110: IIUC, this case should not happen.

How about initializing `is_equal` with false, and throwing an exception in this line?

BalyshevArtem: Done

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
```diff
@@ -21,24 +21,29 @@
...
     if (left_quant_param->scale != right_quant_param->scale or
-        left_quant_param->quantized_dimension != right_quant_param->quantized_dimension)
-      return false;
-
-    if ((left->dtype() == loco::DataType::U8 or left->dtype() == loco::DataType::S8) and
-        left_quant_param->zerop != right_quant_param->zerop)
+        left_quant_param->quantized_dimension != right_quant_param->quantized_dimension or
+        left_quant_param->zerop != right_quant_param->zerop or
+        left_quant_param->min != right_quant_param->min or
+        left_quant_param->max != right_quant_param->max)
```
jinevening, line 35: It would be better to check quantparam is the same, and then return true.

This function will return true even if left has quantparam while right does not have quantparam.

BalyshevArtem: > This function will return true even if left has quantparam while right does not have quantparam.
 Fixed it, thanks

jinevening (dismissed):
LGTM

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
```diff
@@ -0,0 +1,205 @@
...
+            is_equal = is_equal_consts<loco::DataType::S16>(reference_const, cur_const);
+            break;
+          case loco::DataType::S8:
+            is_equal = is_equal_consts<loco::DataType::S8>(reference_const, cur_const);
+            break;
+          case loco::DataType::U8:
+            is_equal = is_equal_consts<loco::DataType::U8>(reference_const, cur_const);
+            break;
+          default:
+            throw std::runtime_error("Not supported type\n");
```
seanshpark, line 129: Q) if not supported dtype exist, is it appropriate to stop with this? or can we just stkip this constant?

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
```diff
@@ -0,0 +1,205 @@
...
+  else
+  {
+    _sum_to_const.at(sum).push_back(const_node);
+  }
+}
+
+/*
+ * Remove duplicate Const nodes.
+ */
+bool RemoveDuplicateConstPass::run(loco::Graph *g)
```
seanshpark, line 172: Like other Pass, it would be nice to have some text graph that explains the key of the Pass.

seanshpark:
> I think we need some unit test for this Pass.

no tests?

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
BalyshevArtem: Done

compiler/luci/pass/src/RemoveDuplicateConstPass.cpp
BalyshevArtem: Done

BalyshevArtem:
> no tests?

done

seanshpark (approved):
LGTM

PR #9861, opened 2022-10-12
[Draft][onert-micro / circle-execution-plan] Enable static memory manager for onert-micro

BalyshevArtem:
This PR enables circle-execution-plan and static memory manager for onert-micro.

for issue: #9849 

ONE-DCO-1.0-Signed-off-by: Artem Balyshev a.balyshev@partner.samsung.com

Issue #9849, opened 2022-10-10
[onert-micro] Add support for StaticMemoryManager


BalyshevArtem:
### What

Let's add support for a static memory manager for `onert-micro`. Currently, support for the static manager in `onert-micro` is disabled since the `circle-execution-planner` is linked to the luci graph view.

### Why

For some models, using a static manager results in a reduction in memory consumption.

### How

There are three main steps:

- [ ] Rewrite `circle-execution-planner`
- [ ] Enable `StaticMemoryManager` in `onert-micro`
- [ ] Change Interpreter interface in `oert-micro`

PR #10011, opened 2022-11-09
[onert-micro] Enable StaticMemoryManager in onert-micro

BalyshevArtem:
This PR enables StaticMemoryManager in onert-micro.

for issue: https://github.com/Samsung/ONE/issues/9849

from draft:  #9861


onert-micro/luci-interpreter/include/luci_interpreter/Interpreter.h
```diff
@@ -30,14 +30,17 @@ namespace luci_interpreter
 class Interpreter
 {
 public:
-  explicit Interpreter(const char *model_data_raw);
+  explicit Interpreter(const char *model_data_raw, bool allocate_input = true);
```
SlavikMIPT, line 33:
```suggestion
  explicit Interpreter(const char *model_data_raw);
  explicit Interpreter(const char *model_data_raw, bool allocate_input);
```

onert-micro/luci-interpreter/include/luci_interpreter/Interpreter.h
```diff
@@ -30,14 +30,17 @@ namespace luci_interpreter
...
 
-  explicit Interpreter(const char *model_data_raw, IMemoryManager *memory_manager);
+  explicit Interpreter(const char *model_data_raw, IMemoryManager *memory_manager,
+                       bool allocate_input = true);
```
SlavikMIPT, line 36:
```suggestion
  explicit Interpreter(const char *model_data_raw, IMemoryManager *memory_manager);
  explicit Interpreter(const char *model_data_raw, IMemoryManager *memory_manager,
                       bool allocate_input);
```

SlavikMIPT (changes requested):
LGTM, but it is not desirable to use default arguments in the constructor

onert-micro/luci-interpreter/include/luci_interpreter/Interpreter.h
BalyshevArtem: Done

onert-micro/luci-interpreter/src/core/RuntimeGraph.h
```diff
@@ -29,42 +29,33 @@ namespace luci_interpreter
 
 class RuntimeModule;
 
-class RuntimeGraph
+class BaseRuntimeGraph
```
SlavikMIPT, line 32:
```suggestion
class IBaseRuntimeGraph
```
since this class has become abstract, I suggest using an "I" prefix for readability

SlavikMIPT (changes requested):

onert-micro/luci-interpreter/src/core/RuntimeGraph.h
BalyshevArtem: Done

SlavikMIPT (approved):
LGTM, thank you

onert-micro/luci-interpreter/src/core/RuntimeGraph.cpp
```diff
@@ -89,83 +143,113 @@ void RuntimeGraph::deallocate(size_t kernel_index) const
...
 }
 
-void RuntimeGraph::addKernel(std::unique_ptr<Kernel> &&kernel)
+StaticRuntimeGraph::~StaticRuntimeGraph()
 {
-  assert(kernel != nullptr);
-  _kernels.push_back(std::move(kernel));
-  _is_valid = false;
+  // Release intermediate computing buffer.
+  const auto static_memory_manager = dynamic_cast<StaticMemoryManager *>(_memory_manager);
```
binarman, line 188: Do we use dynamic casts in other parts of interpreter?
I think it is a good idea to omit such RTTI if possible

onert-micro/luci-interpreter/src/core/RuntimeModule.h
```diff
@@ -32,9 +32,12 @@ class RuntimeModule
 public:
   RuntimeModule() = default;
 
-  RuntimeGraph *addGraph(IMemoryManager *memory_manager)
+  IBaseRuntimeGraph *addGraph(IMemoryManager *memory_manager)
   {
-    _graphs.push_back(std::make_unique<RuntimeGraph>(this, memory_manager));
+    if (memory_manager->is_static_manager())
```
binarman, line 37: Looking at this code I am thinking maybe it is a good idea to make memory manager internal and do not expose as part of user API.

I mean, just leave it as a boolean option in Interpreter contructor

binarman (approved):
We discussed this PR offline with @BalyshevArtem 

The concern I have is following:
- Memory manager was designed to be defined externally to the interpreter, interpreter should not know implementation details of the manager
- This PR leaks some internal details of the manager into interpreter, so in this PR abstraction is not complete.

This raised a question: Do we want to maintain memory manager as an abstract interface ?

We have two options:
- restore abstraction of memory manager: need to store more information in tensors (what is the type of the tensor: input/output/intermediate/etc.)
- reject memory manager abstraction and leave with options in the constructor of the interpreter (like "use dynamic memory manager"/"use static memory manager")

After some discussion, we decided that second way is favorable:
- it is easier to implement
- probability, that someone will need to implement their own memory manager is too small.

Therefore, LGTM, let's merge this as-is and let to replace memory manager interface with constructor options

PR #9989, opened 2022-11-07
[draft][npud] Add Trix backend

jyoungyun:
This commit is a draft PR.
It contains the detail implementation of Trix Backend.

ONE-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #10028, opened 2022-11-16
[npud] Add ContextManager class

jyoungyun:
This commit adds ContextManager to manange Context in npu daemon. It is used by Core class.

Draft PR: https://github.com/Samsung/ONE/pull/9989

### DIFF
--- a/runtime/service/npud/core/ContextManager.cc
+++ b/runtime/service/npud/core/ContextManager.cc
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ContextManager.h"
+
+#include <algorithm>
+#include <util/Logging.h>
+
+namespace npud
+{
+namespace core
+{
+
+ContextManager::ContextManager() noexcept {}
+
+ContextManager::~ContextManager() noexcept { _contexts.clear(); }
+
+void ContextManager::newContext(NpuContext *npuContext, ContextID *contextId)
+{
+  auto context = std::make_unique<Context>();
+  // TODO Consider the possibility of reusing the same address.
+  context->contextId = reinterpret_cast<ContextID>(context.get());
+  context->npuContext = npuContext;
+  *contextId = context->contextId;
+  _contexts.emplace_back(std::move(context));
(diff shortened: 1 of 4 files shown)


runtime/service/npud/core/ContextManager.cc
```diff
@@ -0,0 +1,92 @@
...
+  // TODO Consider the possibility of reusing the same address.
+  context->contextId = reinterpret_cast<ContextID>(context.get());
+  context->npuContext = npuContext;
+  *contextId = context->contextId;
+  _contexts.emplace_back(std::move(context));
+
+  this->listContexts();
+}
+
+void ContextManager::deleteContext(ContextID contextId)
```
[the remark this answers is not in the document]
jyoungyun: @ragmani 

This function deletes the given context Id. What does it mean to delete from?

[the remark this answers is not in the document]
ragmani: `_contexts.erase(iter, _contexts.end());` removes the elements in the range [`iter`, `_contexts.end()`).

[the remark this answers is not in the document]
jyoungyun: The `remove_if` shifts all elements forward except for the item to remove and returns the location of the object to be erased.  And then it erases the location to the end.

This operation just deletes the given item.
So I don't understand why I have to put `from`.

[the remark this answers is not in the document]
ragmani: Sorry, it's my misunderstood. I didn't know about `remove_if`.

runtime/service/npud/core/ContextManager.cc
```diff
@@ -0,0 +1,92 @@
...
+{
+  const auto iter =
+    std::remove_if(_contexts.begin(), _contexts.end(),
+                   [&](std::unique_ptr<Context> &c) { return c->contextId == contextId; });
+  if (iter == _contexts.end())
+  {
+    return;
+  }
```
ragmani, line 51: Is this if block necessary?

jyoungyun: Yes. If there is no item in the list, it just returns.

ragmani: Just FYI, the following codes are safe if there is no item in the list.
But, there is no problem if this block exists as it is.

jyoungyun: Thank you for information. However, I want to return it explicitly. Can I keep this code as it is?

runtime/service/npud/core/ContextManager.cc
```diff
@@ -0,0 +1,92 @@
...
+{
+
+ContextManager::ContextManager() noexcept {}
+
+ContextManager::~ContextManager() noexcept { _contexts.clear(); }
+
+void ContextManager::newContext(NpuContext *npuContext, ContextID *contextId)
+{
+  auto context = std::make_unique<Context>();
+  // TODO Consider the possibility of reusing the same address.
```
ragmani, line 34: I'm not sure about it but I think it depends how you allocate and deallocate `Context` instances.
If there are deallocated heap memory for those instances, there is a possibility of reusing the same address. But, it seems that there is no duplicated addresses in the vector if those deallocated instances were removed from the vector.

ragmani (approved):
LGTM

hseok-oh (approved):

PR #10024, opened 2022-11-16
DRAFT fix UnidirSeqLSTM value test

seanshpark:
on going draft to fix UnidirectionalSequenceLSTM value test.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #10030, opened 2022-11-16
[luci-interpreter] Tidy unnecessary pragma

seanshpark:
This will tidy unnecessary pragma that was unintentionally added.


seanshpark:
from  #10024

mhs4670go (approved):
LGTM

PR #10031, opened 2022-11-16
[luci-interpreter] Remove SetupScratchpadTensor

seanshpark:
This will remove using SetupScratchpadTensor and inline codes.
- to reduce difficulties tracing codes
- we can extract them to pal when necessary


seanshpark:
from draft #10024

mhs4670go (approved):
LGTM

PR #10032, opened 2022-11-16
[luci-interpreter] Clear also sp_scratch_buffer

seanshpark:
This will clear also sp_scratch_buffer values to zero.


seanshpark:
from draft #10024, this is to prevent wrong values.

mhs4670go (approved):
LGTM

Issue #10025, opened 2022-11-16
[luci] fix value-test for UnidirectionalSequenceLSTM


seanshpark:
I have some how missed adding `UnidirectionalSequenceLSTM` for value testing.
- `UnidirectionalSequenceLSTM_002` was not added to list
- `UnidirectionalSequenceLSTM_002` fails when added

PR #10033, opened 2022-11-16
[luci-interpreter] Fix argument order of UnidirSeqLSTM

seanshpark:
This will fix argument order of creating UnidirectionalSequenceLSTM kernel.


seanshpark:
for  #10025, from draft #10024

mhs4670go (approved):
LGTM

PR #10037, opened 2022-11-16
[luci/logex] Add float/int to_str

seanshpark:
This will add to_str() method for float and int32_t to fix calling wrong bool type.


seanshpark:
found this while working on  #10025

mhs4670go (approved):
LGTM

Issue #10002, opened 2022-11-09
[onert] Support Backend aware Quantization


ragmani:
# What
Let's support quantization/dequantization between heterogeneous backends in onert

# Why
1. Supports CPU/TRIX heterogeneous computing with clrcle models that cannot be compiled to be used in TRIX backend.
2. Heterogeneous computing of models that require more accuracy
3. any other cases?

# how far
- Scope of support
  1. Support between multiple models : Unconditional support
  2. Support between all other backends : if necessary

- Supported data types
  - FLOAT <-> QASYYM8 : Unconditional support
  - FLOAT <-> QSYMM8: If necessary (It seems to be used only in hybrid weight, so it's ambiguous if this case should be supported)
  - FLOAT <-> QASYYM8_PER_CHANNERL : if necessary
  - FLOAT <-> QSYMM16 : Unconditional support
  - FLOAT <-> QASYMM16 : if necessary
  - Others: if necessary

# How to
Discussion at ...

# TODO
TBD

chunseoklee:
> Support between multiple models : Unconditional support

"Unconditional" means model without control flow ?  or else ? 

> FLOAT <-> QASYYM8 : Unconditional support

What does "Unconditonal" mean here ?

ragmani:
@chunseoklee 

> "Unconditional" means model without control flow ? or else ?

My intention was supporting multiple models regardless of control flow.

> What does "Unconditonal" mean here ?

My intention was essential. I changed it in https://github.com/Samsung/ONE/issues/10002#issue-1441527173.

ragmani:
I tried to add the new builtin operation to subgraph's output corresponding to `from` of edge. However, I figured out that this way is  difficult to support dynamic tensor in our current `onert`. So, I'm trying to change the way to adding the new built-in operation to the subgraph's input operand corresponding to `to` of edge.

ragmani:
>  I'm trying to change the way to adding the new built-in operation to the subgraph's input operand corresponding to to of edge.

I discard the plan to add the new built-in operation. It is going to be replaced with #10296. For detail, https://github.com/Samsung/ONE/issues/10003#issuecomment-1373117394.

PR #10041, opened 2022-11-16
[draft][onert] Add ModelEdge operation

ragmani:
For #10002

This operation will support functions as follows:
      - Support copying memory of tensors between different models
      - Support permutes tensors between different models
      - Support type conversion(quant/dequant) of tensors between different models


ragmani:
@chunseoklee 
> where "quan/dequant" take places ? We will implment it on PermuteLayer ?

Yes, I have a plan to implement it inside `PermuteLayer`.

ragmani:
Discarded this draft.

Issue #10014, opened 2022-11-13
[CFE] dredd test support count for CircleConst


seanshpark:
dredd test uses `circle-inspect` tool to print operators but as CircleConst is not an operator, it doesn't show it.
I'm not sure which may be the best solution but I can think is to add an option to `circle-inspect` to print `CONST` for each `CircleCons`.
We need to revise
- `circle-inspect` to add option something like `--constants`
- `dredd-rule-lib/rule-lib.sh` to add something like `const_count`

If you like, you can change them or wait for someone to do it.
I'll post an issue for this.

_Originally posted by @seanshpark in https://github.com/Samsung/ONE/pull/10008#discussion_r1020968813_

BalyshevArtem:
draft:  #10008

seanshpark:
done :)
thank you @BalyshevArtem !

PR #10044, opened 2022-11-16
[dredd-rule-lib] Add rule to check constant tensor count

BalyshevArtem:
This PR adds rule to check constant tensor count.

for issue: https://github.com/Samsung/ONE/issues/10014
draft: https://github.com/Samsung/ONE/pull/10008


seanshpark (approved):
LGTM

Issue #9940, opened 2022-10-27
[CFE] Unroll UnidirectionalSequenceLSTM for trix


seanshpark:
Let's provide unrolling `UnidirectionalSequenceLSTM` Op for compilation in Trix.
- to match how LSTM op unrolling in ONNX, we may provide this option in _one-import-tf_ or _one-import-tflite_
- as for `UnidirectionalSequenceLSTM`, it will be better to unroll in _circle2circle_, technically
- _one-import-tf_ or _one-import-tflite_ can pass this flag to _circle2circle_

seanshpark:
related #9895

seanshpark:
> one-import-tf or one-import-tflite can pass this flag to circle2circle

How? 

@mhs4670go , any comments about `how-to` ?

mhs4670go:
For others' information, I summarize the offline talk with @seanshpark.

_one-import-onnx_ has following workflow when it comes to processing LSTM operator.

.onnx -> (onnx_legalizer) -> .onnx -> (onnx_tf) -> ..

Current legalizer only support _.onnx_ inputs and accordingly only _one-import-onnx_ has an option that unrolls LSTM operators.

And now, this issue is trying to make _one-import-tf*_ have the same functionality.

We could introduce the option that unrolls LSTM op into _circle2circle_, which means that option would be introduced at _one-optimize_. But, when it comes to unified interface with existing _one-import-onnx_, _one-import-tf*_ not having legalizing option seems not natural.

So, here is the thing.

1. Introduce the option into _one-import-tf*_ but strictly speaking, that option does nothing. It's just for aligning the interface with _one-import-onnx_.
2. Introduce the option into _circle2circle_(_one-optimize_) and turn it on by default. It always unrolls the LSTM operator. Then, both workflows'd result in the same meaning;unrolling the LSTM op.
- _one-import-onnx_ -> _one-optimize_
- _one-import-tf*_ -> _one-optimize_

seanshpark:
How to validate?
- use tflite model in https://github.com/Samsung/ONE/issues/9895#issuecomment-1289768739
- get input-output values
- unroll in circle2circle
- get input-output values of circle
- compare tflite (with LSTM) vs circle (unrolled)

seanshpark:
Unroll pattern?
- assume input is [C,X,Y]
- transpose input with perm [1,0,2], iput as [X,C,Y] -- (1)
- unpack with axis 0, number of X -- (1)
[...]

PR #10051, opened 2022-11-17
[luci/pass] UnrollLSTM first Transpose

seanshpark:
This will add first Transpose when timesteps is larger than 1.


seanshpark:
for  #9940 , from draft #9899

mhs4670go (approved):
LGTM

PR #10055, opened 2022-11-17
[luci/pass] UnrollLSTM unpack

seanshpark:
This will add first unpack with count of timesteps.

### DIFF
--- a/compiler/luci/pass/src/UnrollUnidirectionalSequenceLSTMPass.cpp
+++ b/compiler/luci/pass/src/UnrollUnidirectionalSequenceLSTMPass.cpp
@@ -123,6 +123,7 @@ struct UnrollLSTM
 {
   luci::CircleConst *transpose_perm(void);
   luci::CircleTranspose *first_transpose(luci::CircleNode *input);
+  std::vector<luci::CircleUnpackOut *> input_unpacks(luci::CircleNode *input);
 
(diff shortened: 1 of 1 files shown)


seanshpark:
for  #9940 , from draft #9899

compiler/luci/pass/src/UnrollUnidirectionalSequenceLSTMPass.cpp
```diff
@@ -162,6 +163,30 @@ luci::CircleTranspose *UnrollLSTM::first_transpose(luci::CircleNode *input)
   return transpose;
 }
 
+std::vector<luci::CircleUnpackOut *> UnrollLSTM::input_unpacks(luci::CircleNode *input)
```
mhs4670go, line 166: Parameter null checking seems needed. And so does below `unroll_lstm` method.

seanshpark: I'll add null check in another PR.

seanshpark: I think null check for input only in `unroll_lstm` will do.

seanshpark: null check inside `input_unpacks` for safety seems nice.

mhs4670go (approved):
LGTM

PR #10056, opened 2022-11-17
[luci/pass] UnrollLSTM MatMul

seanshpark:
This will add first MatMul(FullyConnected) with merged weights of LSTM.


seanshpark:
for  #9940 , from draft #9899

PR #10063, opened 2022-11-18
[npud] Implement register and unregister model in Trix backend

jyoungyun:
This commit registers a given model. If the same model is already registered, it shares the model information data.
Each NpuContext has own model list.

Draft PR: #9989

### DIFF
--- a/runtime/service/npud/backend/trix/TrixBackend.cc
+++ b/runtime/service/npud/backend/trix/TrixBackend.cc
@@ -116,14 +116,76 @@ NpuStatus TrixBackend::destroyBuffer(NpuContext *ctx, GenericBuffer *buffer)
 NpuStatus TrixBackend::registerModel(NpuContext *ctx, const std::string &modelPath,
                                      ModelID *modelId)
 {
-  // TODO Implement details
-  return NPU_STATUS_ERROR_NOT_SUPPORTED;
+  if (ctx == nullptr)
+  {
+    return NPU_STATUS_ERROR_INVALID_ARGUMENT;
+  }
+
+  ModelID id;
+  auto iter = std::find_if(
+    _dev->models.begin(), _dev->models.end(), [&](const std::weak_ptr<NpuModelInfo> &p) {
+      auto info = p.lock();
+      return info ? (info->core == ctx->defaultCore && info->path == modelPath) : false;
+    });
+  // Already registered model.
+  if (iter != _dev->models.end())
+  {
+    auto info = iter->lock();
+    id = info->id;
+
+    auto mIter = ctx->models.find(id);
+    if (mIter == ctx->models.end())
+    {
+      ctx->models.insert(std::make_pair(id, info));
+    }
+  }
+  else
+  {
+    auto meta = getNPUmodel_metadata(modelPath.c_str(), false);
+    if (meta == nullptr)
+    {
+      return NPU_STATUS_ERROR_OPERATION_FAILED;
+    }
+
+    generic_buffer fileInfo;
+    fileInfo.type = BUFFER_FILE;
+    fileInfo.filepath = modelPath.c_str();
+    fileInfo.size = meta->size;
+
+    npudev_h handle = _dev->handles.at(ctx->defaultCore);
+    if (registerNPUmodel(handle, &fileInfo, &id) < 0)
+    {
+      return NPU_STATUS_ERROR_OPERATION_FAILED;
+    }
+
+    auto info = std::shared_ptr<NpuModelInfo>(new NpuModelInfo({id, modelPath, ctx->defaultCore}));
+    ctx->models.insert(std::make_pair(id, info));
+    _dev->models.emplace_back(info);
+  }
+
+  *modelId = id;
+  return NPU_STATUS_SUCCESS;
 }
 
 NpuStatus TrixBackend::unregisterModel(NpuContext *ctx, ModelID modelId)
 {
-  // TODO Implement details
-  return NPU_STATUS_ERROR_NOT_SUPPORTED;
+  if (ctx == nullptr)
+  {
+    return NPU_STATUS_ERROR_INVALID_ARGUMENT;
+  }
+
+  auto iter = ctx->models.find(modelId);
+  if (iter == ctx->models.end())
+  {
+    return NPU_STATUS_ERROR_INVALID_MODEL;
+  }
+  iter->second.reset();
+  ctx->models.erase(iter);
+
+  auto mIter = std::remove_if(_dev->models.begin(), _dev->models.end(),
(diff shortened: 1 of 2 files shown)


runtime/service/npud/backend/trix/TrixBackend.cc
```diff
@@ -116,14 +116,83 @@ NpuStatus TrixBackend::destroyBuffer(NpuContext *ctx, GenericBuffer *buffer)
...
+                           [&](const std::weak_ptr<NpuModelInfo> &p) {
+                             auto info = p.lock();
+                             if (info)
+                             {
+                               return info->core == ctx->defaultCore && info->path == modelPath;
+                             }
+                             else
+                             {
+                               return false;
+                             }
```
hseok-oh, line 135:
```suggestion
                             return info ? (info->core == ctx->defaultCore && info->path == modelPath) : false;
```

runtime/service/npud/backend/trix/TrixBackend.h
```diff
@@ -36,6 +36,7 @@ using Handle = void *;
 struct TrixDevice
 {
   std::vector<Handle> handles;
+  std::vector<std::weak_ptr<NpuModelInfo>> models;
```
hseok-oh, line 39: Any reason to use `weak_ptr` ?

jyoungyun: The `_dev` has all model files. When a specific context requests model registration, if `_dev` does not have this model, it registers the model and push the model data to `_dev`. And it shares the model data with a specific context as weak_ptr.

If it use `shared_ptr`, the model file will not be destroyed because `_dev` always has a model.

The reason for using `weak_ptr` is that if all contexts unregister model, it recognizes it and deletes the model data in `_dev`.

hseok-oh: I see. IMO, you need to add more comment on header files. It's not easy to understand concept of fields or methods.

hseok-oh (approved):
LGTM

runtime/service/npud/backend/trix/TrixBackend.cc
```diff
@@ -116,14 +116,76 @@ NpuStatus TrixBackend::destroyBuffer(NpuContext *ctx, GenericBuffer *buffer)
...
+  auto iter = std::find_if(_dev->models.begin(), _dev->models.end(),
+                           [&](const std::weak_ptr<NpuModelInfo> &p) {
+                             auto info = p.lock();
+                             return info ? (info->core == ctx->defaultCore && info->path == modelPath) : false;
+                           });
+  // Already registered model.
+  if (iter != _dev->models.end())
+  {
+    auto info = iter->lock();
+    id = info->id;
```
ragmani, line 134: I guess there should be some code that checks if `info` is nullptr. AFAIK `weak_ptr` does not count reference count, so there is a potential risk of referencing nullptr.

jyoungyun: If `iter` does not reach `_dev->models.end()`, it means that it has found the item it want to find. So I didn't do the null check here.

hseok-oh (approved):

runtime/service/npud/backend/trix/TrixBackend.cc
ragmani: First of all, I want to make one thing clear. There may be no problems in the code in your implementation. But I'm afraid of any potential risk occurred by the code being modified by someone later in any other codes.
Anyway, I found that there is a null check when looking for an item. But it is ambiguous. If item is nullptr in your implementation, it may return false even if the algorithm(`info->core == ctx->defaultCore && info->path == modelPath`) you implemented matches.

jyoungyun: The `found_if` function finds an item using `UnaryPredicate` function by iterating each element.  In `UnaryPredicate` function, it accesses all item and returns to false if it is nullptr. The iterator then becomes the last item on the list. Therefore, there is no case that `info` becomes nullptr becasue `found_if` has already checked.

Should I update this code?

ragmani: The fact that a `info` is `nullptr` seems to mean the model of the `info` has been registered, but has been released from `models` of `NpuContext`. In other words, items in `models` of `TrixDevice` can be `nullptr` regardless of `find_if` if an item in `models` of `NpuContext` is released outside this function.
What I'm worried about is whether there is a problem if `find_if` cannot find the `info` that was `(info->core == ctx->defaultCore && info->path == modelPath)`, but is nullptr now.
It's too hard to describe in English :cry:

runtime/service/npud/backend/trix/TrixBackend.cc
```diff
@@ -116,14 +116,76 @@ NpuStatus TrixBackend::destroyBuffer(NpuContext *ctx, GenericBuffer *buffer)
...
+    fileInfo.filepath = modelPath.c_str();
+    fileInfo.size = meta->size;
+
+    npudev_h handle = _dev->handles.at(ctx->defaultCore);
+    if (registerNPUmodel(handle, &fileInfo, &id) < 0)
+    {
+      return NPU_STATUS_ERROR_OPERATION_FAILED;
+    }
+
+    auto info = std::shared_ptr<NpuModelInfo>(new NpuModelInfo({id, modelPath, ctx->defaultCore}));
```
ragmani, line 161: (optional)
```suggestion
    auto info = std::shared_ptr<NpuModelInfo>(new NpuModelInfo{id, modelPath, ctx->defaultCore});
```

runtime/service/npud/backend/trix/TrixBackend.cc
```diff
@@ -116,14 +116,76 @@ NpuStatus TrixBackend::destroyBuffer(NpuContext *ctx, GenericBuffer *buffer)
...
+  {
+    return NPU_STATUS_ERROR_INVALID_ARGUMENT;
+  }
+
+  auto iter = ctx->models.find(modelId);
+  if (iter == ctx->models.end())
+  {
+    return NPU_STATUS_ERROR_INVALID_MODEL;
+  }
+  iter->second.reset();
```
ragmani, line 182: (Optional) Maybe there is no problem if this line is removed.

runtime/service/npud/backend/trix/TrixBackend.cc
ragmani: @jyoungyun told me offline that there is no problem in the codes afterwards even if  `find_if` returns `false` by `nullptr` of `info`. The following codes has already been considered the case that `info` is `nullptr` and `info->core == ctx->defaultCore && info->path == modelPath` is `true`.

ragmani (approved):

PR #10070, opened 2022-11-21
[luci/pass] assert for unroll_lstm

seanshpark:
This will add assert for lstm object.


mhs4670go (approved):
LGTM

PR #10071, opened 2022-11-21
[luci] Add guard code to FuseAddWithFullyConnectedPass

jinevening:
This prevents Add from being fused with FC with non-const bias.


seanshpark (approved):
LGTM

Issue #10050, opened 2022-11-17
CFE fix ARM cross build on U20.04


seanshpark:
CI fails to cross build on Ubuntu20.04.

> 17:30:10 /home/jenkins/jenkins_agent/workspace/nnfw/master/push-cross-compiler-debug/externals/TENSORFLOW-2.6.0-RUY/ruy/pack_arm.cc: In function 'void ruy::Pack8bitColMajorForNeon4Cols(const ruy::PackParams8bit&)':
17:30:10 /home/jenkins/jenkins_agent/workspace/nnfw/master/push-cross-compiler-debug/externals/TENSORFLOW-2.6.0-RUY/ruy/pack_arm.cc:469:72: error: 'asm' operand has impossible constraints
17:30:10   469 |         "q4", "q5", "q6", "q7", "q8", "q9", "q10", "q11", "q12", "q13");
17:30:10       |                                                                        ^
17:30:10 [ 36%] Generating TopKV2_001.gen.recipe
17:30:10 make[3]: Leaving directory '/home/jenkins/jenkins_agent/workspace/nnfw/master/push-cross-compiler-debug/build/arm32.debug'
17:30:10 make[3]: *** [compiler/luci-interpreter/src/kernels/CMakeFiles/luci_interpreter_linux_pal.dir/build.make:232: compiler/luci-interpreter/src/kernels/CMakeFiles/luci_interpreter_linux_pal.dir/home/jenkins/jenkins_agent/workspace/nnfw/master/push-cross-compiler-debug/externals/TENSORFLOW-2.6.0-RUY/ruy/pack_arm.cc.o] Error 1
17:30:10 make[2]: *** [CMakeFiles/Makefile2:5331: compiler/luci-interpreter/src/kernels/CMakeFiles/luci_interpreter_linux_pal.dir/all] Error 2

lets fix this

seanshpark:
from https://github.com/Samsung/ONE/issues/10020#issuecomment-1316059344

seanshpark:
using docker with https://github.sec.samsung.net/seanshpark/Dockerfiles/tree/office/ONE-focal

seanshpark:
Same error with
- changed some compile options
- Use `ruy` from TensorFlow 2.8.0

chunseoklee:
> Same error with
> - changed some compile options
> - Use `ruy` from TensorFlow 2.8.0
> 

Error occurs with both changes ?

seanshpark:
[...]

PR #10078, opened 2022-11-21
[infra/nncc] Disable NEON

seanshpark:
This will disable NEON not to use RUY with ARM32 cross build.


seanshpark:
for #10050

PR #10080, opened 2022-11-21
[luci-value-test] UnidirSeqLSTM

seanshpark:
This will add value tests of UnidirectionalSequenceLSTM operator.


seanshpark:
for  #9940 , from draft #9899

PR #10081, opened 2022-11-21
DRAFT: Support constant folding of FC

jinevening:
On going draft to support constat folding of FC.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/10066

Issue #10066, opened 2022-11-18
[one-optimize] Constant folding for FullyConnected


jinevening:
## What

Let's implement constant folding for FC layer.

jinevening:
`FoldDepthwiseConv2DPass` would be helpful for this taks.

seanshpark:
@jinevening , I also have needs of kernel codes for shape inference of luci service.

How about adding another module like `luci-kernels` or `luci-comptute` to provide computation codes for multiple purposes?
- constant folding in pass
- shape inference in service
- kernel codes for luci-interpreter
- ...

jinevening:
Look good to me. #10081 does not harm code structure, so can we go as-is and introduce the new module after this task is done? This task is for our customer's model support.

jinevening:
Done.

seanshpark:
@jinevening , could you please add a test in `luci_pass_value_test` ?

jinevening:
I'll add a test.

PR #10082, opened 2022-11-21
[mio-tflite280] Support ruy library

jinevening:
This supports ruy in mio-tflite280.

---
Related to: https://github.com/Samsung/ONE/issues/10066
Draft PR: #10081

### DIFF
--- a/compiler/mio-tflite280/CMakeLists.txt
+++ b/compiler/mio-tflite280/CMakeLists.txt
@@ -44,9 +44,16 @@ if(NOT TensorFlowGEMMLowpSource_FOUND)
   return()
 endif(NOT TensorFlowGEMMLowpSource_FOUND)
 
+nnas_find_package(TensorFlowRuySource EXACT 2.8.0 QUIET)
+
+if(NOT TensorFlowRuySource_FOUND)
+  return()
+endif(NOT TensorFlowRuySource_FOUND)
+
 add_library(mio_tflite280_inc INTERFACE)
 target_include_directories(mio_tflite280_inc SYSTEM INTERFACE "${TensorFlowSource_DIR}")
 target_include_directories(mio_tflite280_inc SYSTEM INTERFACE "${TensorFlowGEMMLowpSource_DIR}")
+target_include_directories(mio_tflite280_inc SYSTEM INTERFACE "${TensorFlowRuySource_DIR}")
 
 file(GLOB_RECURSE SOURCES "src/*.cpp")
 file(GLOB_RECURSE TESTS "src/*.test.cpp")


seanshpark:
@jinevening , have you checked this in Ubuntu 20.04 and Ubuntu 22.04?

seanshpark:
from #10020 , I've disabled using RUY in ARM cross build.
I don't think this change is for x86-64 host.
What target is this for?

jinevening:
The change is for one-optimize, so the main target is x86 (I have no idea on-device compiler will have this code).

If ruy is not allowed, I'll find another kernel (e.g., legacy kernel using Eigen).

seanshpark:
> so the main target is x86 

Ah, this is for x86-64... I think disabling for ARM cross build has nothing to do with this.

compiler/mio-tflite280/CMakeLists.txt
```diff
@@ -44,9 +44,16 @@ if(NOT TensorFlowGEMMLowpSource_FOUND)
...
+nnas_find_package(TensorFlowRuySource EXACT 2.8.0 QUIET)
+
+if(NOT TensorFlowRuySource_FOUND)
+  return()
+endif(NOT TensorFlowRuySource_FOUND)
+
 add_library(mio_tflite280_inc INTERFACE)
 target_include_directories(mio_tflite280_inc SYSTEM INTERFACE "${TensorFlowSource_DIR}")
 target_include_directories(mio_tflite280_inc SYSTEM INTERFACE "${TensorFlowGEMMLowpSource_DIR}")
+target_include_directories(mio_tflite280_inc SYSTEM INTERFACE "${TensorFlowRuySource_DIR}")
```
seanshpark, line 56: Could you please explain why you added RUY header includes to `mio-tflite` ?
I could guess but this module is for IO of tflite files not kernels.

jinevening: I found that `FoldDepthwiseConv2DPass` in `luci_pass` is using `mio-tflite` which includes `GEMMLowpSource`.

Similarly, I thought that `FoldFullyConnectedPass` can use `mio-tflite` which includes `Ruy`.

seanshpark: I'm not sure but I think GEMMLowpSource was added from dependency to use tflite source codes.

> Similarly, I thought that FoldFullyConnectedPass can use mio-tflite which includes Ruy.

Yes, logically this is OK.
The thing that makes me hesitate is that `mio-tflite` is getting complicated.
I'll post an issue to separate things.

seanshpark (approved):
LGTM

PR #10089, opened 2022-11-22
[luci] Add FoldFullyConnectedPass

jinevening:
This adds FoldFullyConnectedPass to luci/pass.

---
Related to: https://github.com/Samsung/ONE/issues/10066
Draft PR: https://github.com/Samsung/ONE/pull/10081

### DIFF
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -23,6 +23,7 @@
 #include "luci/Pass/FoldDensifyPass.h"
 #include "luci/Pass/FoldDepthwiseConv2DPass.h"
 #include "luci/Pass/FoldDequantizePass.h"
+#include "luci/Pass/FoldFullyConnectedPass.h"
 #include "luci/Pass/FoldGatherPass.h"
 #include "luci/Pass/FoldSparseToDensePass.h"
 #include "luci/Pass/ForwardReshapeToUnaryOpPass.h"
@@ -300,6 +301,10 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   if (_options->query(Options::Algorithm::FoldDequantize))
   {
     phase.emplace_back(std::make_unique<luci::FoldDequantizePass>());
+  }
+  if (_options->query(Options::Algorithm::FoldFullyConnected))
+  {
+    phase.emplace_back(std::make_unique<luci::FoldFullyConnectedPass>());
   }
   if (_options->query(Options::Algorithm::FoldGather))
   {
--- a/compiler/luci/pass/src/FoldFullyConnectedPass.cpp
+++ b/compiler/luci/pass/src/FoldFullyConnectedPass.cpp
@@ -1,0 +1,205 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FoldFullyConnectedPass.h"
+
+#include <tensorflow/lite/kernels/internal/reference/fully_connected.h>
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/AttrFusedActFunc.h>
+
+#include <luci/Log.h>
+
+#include <limits> // std::numeric_limits
+
+namespace
+{
+
+bool set_kernel_parameters(tflite::FullyConnectedParams *params, luci::CircleFullyConnected *node)
+{
+  switch (node->fusedActivationFunction())
+  {
+    case luci::FusedActFunc::NONE:
+    case luci::FusedActFunc::TANH:
+      params->float_activation_min = std::numeric_limits<float>::lowest();
+      params->float_activation_max = std::numeric_limits<float>::max();
+      break;
+    case luci::FusedActFunc::RELU:
+      params->float_activation_min = 0;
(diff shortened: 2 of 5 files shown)


compiler/luci/pass/src/FoldFullyConnectedPass.cpp
```diff
@@ -0,0 +1,205 @@
...
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto fc = dynamic_cast<CircleFullyConnected *>(node);
+
+    if (fc == nullptr)
+      continue;
+
+    switch (fc->dtype())
+    {
+      case loco::DataType::FLOAT32:
```
seanshpark, line 192: Q) current `fold_fully_connected()` only supports FLOAT32. are you planning to support for other types?
Anyway, as for current implementation switch-case doesn't seem to be needed.

jinevening: I don't have any plan for another type yet :)

I've removed switch-case.

compiler/luci/pass/src/FoldFullyConnectedPass.cpp
```diff
@@ -0,0 +1,201 @@
...
+  {
+    auto fc = dynamic_cast<CircleFullyConnected *>(node);
+
+    if (fc == nullptr)
+      continue;
+
+    if (fc->dtype() != loco::DataType::FLOAT32)
+      continue;
+
+    changed = fold_fully_connected(fc);
```
seanshpark, line 193:
```suggestion
    if (fold_fully_connected(fc))
        changed = true;
```
or `changed` can be updated to `false`

mhs4670go, line 193: Is it okay with just `=`?
```suggestion
    changed |= fold_fully_connected(fc);
```

compiler/luci/pass/src/FoldFullyConnectedPass.cpp
```diff
@@ -0,0 +1,201 @@
...
+ *
+ *    [CircleConst] [CircleConst]
+ *               |   |
+ *       [CircleFullyConnected]
+ *
+ *    AFTER
+ *
+ *           [CircleConst]
+ */
+bool fold_fully_connected(luci::CircleFullyConnected *node)
```
mhs4670go, line 84: `node` should be checked if it is `nullptr` or not.

compiler/luci/pass/src/FoldFullyConnectedPass.cpp
jinevening: Thanks for finding!

compiler/luci/pass/src/FoldFullyConnectedPass.cpp
mhs4670go: Well, turns out that it is not necessary because below logic already checks it.

(optional) I think function itself should check the parameter because we can't expect the behaviors of uers.
So you can move below logic into inside the function. The function call overhead'd come in though.

jinevening: I added the check to avoid complain from our SE tool.

compiler/luci/pass/src/FoldFullyConnectedPass.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+#include <luci/Log.h>
+
+#include <limits> // std::numeric_limits
+
+namespace
+{
+
+bool set_kernel_parameters(tflite::FullyConnectedParams *params, luci::CircleFullyConnected *node)
+{
```
mhs4670go, line 32: These parameters should be checked, too.

compiler/luci/pass/src/FoldFullyConnectedPass.cpp
```diff
@@ -0,0 +1,204 @@
...
+    tflite::RuntimeShape runtime_shape(node->rank());
+    for (uint32_t i = 0; i < node->rank(); ++i)
+      runtime_shape.SetDim(i, node->dim(i).value());
+    return runtime_shape;
+  };
+
+  auto tensor_data = [](luci::CircleConst *node) -> float * {
+    if (node == nullptr)
+      return nullptr;
+
```
mhs4670go, line 161: Maybe the size check is needed.
```cxx
if(node->size<loco::DataType::FLOAT32>() != 1)
  return nullptr;

return &node->at<loco::DataType::FLOAT32>(0);
```

compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -301,6 +302,10 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   {
     phase.emplace_back(std::make_unique<luci::FoldDequantizePass>());
   }
+  if (_options->query(Options::Algorithm::FoldFullyConnected))
+  {
+    phase.emplace_back(std::make_unique<luci::FoldFullyConnectedPass>());
+  }
```
mhs4670go, line 308: (optional) Just note for future. Options are getting bigger and code lines is also bigger accordingly. So, we may introduce a dedicated constant container of `Options::Algorithm` enum class.

It would be like this.
```cxx
std::unique_ptr<logo::Pass> make_pass(const Options::Algorithm &algo)
{
  switch(algo)
  {
    case Options::Algorithm::FoldFullyConnected:
      return std::make_unique<luci::FoldFullyConnectedPass>();
    // ..
  }
}

for(const auto &algo: ALGORITHM_BOOKS)
{
  if (_options->query(algo))
    phase.emplace_back(make_pass(algo));
}
```

compiler/luci/pass/src/FoldFullyConnectedPass.cpp
jinevening: The returned value is the pointer of the vector, not the value of the first element. So, the size can be non-one.

compiler/luci/pass/src/CircleOptimizer.cpp
jinevening: Good suggestion!

compiler/luci/pass/src/FoldFullyConnectedPass.cpp
mhs4670go: Ah, I see. Then, I just worried about the case `size == 0` because of accessing `at(0)`. But vector `at` method will throw an error.

mhs4670go (approved):
LGTM. Thanks:)

PR #10090, opened 2022-11-22
[infra/nncc] Relocate tree-vectorize

seanshpark:
This will move '-ftree-vectorize' option into NEON block.


seanshpark:
As `tree-vectorize` uses SIMD vectorize option should move into NEON flag.

chunseoklee (approved):
for reference :  https://developer.arm.com/documentation/dht0002/a/Introducing-NEON/Developing-for-NEON/Automatic-vectorization

Issue #10085, opened 2022-11-21
CFE enable RUY for ARM cross build


seanshpark:
This feature is disabled cause of compiler error.
Let's enable this when the problem is fixed.

This issue is added not to forget the task.

seanshpark:
related #10050

chunseoklee:
FYI, armv7l-linux build is successful with:

```
(tf-2.8) ~/d/ONE> /usr/bin/arm-linux-gnueabihf-g++ -v
Using built-in specs.
COLLECT_GCC=/usr/bin/arm-linux-gnueabihf-g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc-cross/arm-linux-gnueabihf/9/lto-wrapper
Target: arm-linux-gnueabihf
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 9.4.0-1ubuntu1~20.04.1' --with-bugurl=file:///usr/share/doc/gcc-9/README.Bugs --enable-languages=c,ada,c++,go,d,fortran,objc,obj-c++,gm2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-9 --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-libitm --disable-libquadmath --disable-libquadmath-support --enable-plugin --enable-default-pie --with-system-zlib --without-target-system-zlib --enable-libpth-m2 --enable-multiarch --enable-multilib --disable-sjlj-exceptions --with-arch=armv7-a --with-fpu=vfpv3-d16 --with-float=hard --with-mode=thumb --disable-werror --enable-multilib --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=arm-linux-gnueabihf --program-prefix=arm-linux-gnueabihf- --includedir=/usr/arm-linux-gnueabihf/include
Thread model: posix
gcc version 9.4.0 (Ubuntu 9.4.0-1ubuntu1~20.04.1)
                                                                                                                                                   [ 0s | 11월 22 10:21오전 ]
(tf-2.8) ~/d/ONE> git rev-parse --short HEAD
76ddbe340
 ```

chunseoklee:
[...]

PR #10096, opened 2022-11-22
[infra/nncc] Revise config_armv7l-linux.cmake

seanshpark:
This will revise config_armv7l-linux.cmake to fix ruy build.


seanshpark:
for #10085

chunseoklee (approved):

PR #10099, opened 2022-11-22
[infra/nncc] Revise Makefile.arm32 NEON ON

seanshpark:
This will revise Makefile.arm32 with NEON build ON again.


seanshpark:
for #10085

chunseoklee (approved):

PR #10026, opened 2022-11-16
[npud] Add context_create and context_destroy dbus methods

jyoungyun:
This commit adds context_create and context_destroy dbus methods.

Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

Draft PR: https://github.com/Samsung/ONE/pull/9989

PR #10104, opened 2022-11-22
[npud] Add request_create and request_destroy dbus methods

jyoungyun:
This commit adds request_create and request_destroy dbus methods.

Draft Pr: #9989

### DIFF
--- a/runtime/service/npud/core/DBus.cc
+++ b/runtime/service/npud/core/DBus.cc
@@ -57,6 +57,8 @@ void DBus::on_bus_acquired(GDBusConnection *conn, const gchar *name, gpointer us
   iface->handle_context_destroy = &on_handle_context_destroy;
   iface->handle_network_create = &on_handle_network_create;
   iface->handle_network_destroy = &on_handle_network_destroy;
+  iface->handle_request_create = &on_handle_request_create;
+  iface->handle_request_destroy = &on_handle_request_destroy;
 
   if (!g_dbus_interface_skeleton_export(G_DBUS_INTERFACE_SKELETON(core), conn, "/org/tizen/npud",
                                         &error))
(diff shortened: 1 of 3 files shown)


hseok-oh (approved):

runtime/service/npud/core/DBus.cc
```diff
@@ -132,5 +134,28 @@ gboolean DBus::on_handle_network_destroy(NpudCore *object, GDBusMethodInvocation
...
+gboolean DBus::on_handle_request_create(NpudCore *object, GDBusMethodInvocation *invocation,
+                                        guint64 arg_ctx, guint arg_nw_handle)
+{
+  VERBOSE(DBus) << "on_handle_request_create with " << arg_ctx << ", " << arg_nw_handle
+                << std::endl;
+  RequestID requestID = 0;
+  int ret = -1;
+  // TODO Invoke core function.
+  npud_core_complete_request_create(object, invocation, guint(requestID), ret);
+  return TRUE;
```
ragmani, line 146: I have a curious question. Why is `ret` not returend?

ragmani (approved):

runtime/service/npud/core/DBus.cc
jyoungyun: > The `return TRUE` is from `gdbus` guideline. I wanted to find the document about this, but I couldn't. 
> Dbus regards that the `ret` value of `npud_core_complete_context_create` is the return value of function. If it has an error, we need to use [g_dbus_method_invocation_return_error](https://www.freedesktop.org/software/gstreamer-sdk/data/docs/latest/gio/GDBusMethodInvocation.html#g-dbus-method-invocation-return-error) function.

from https://github.com/Samsung/ONE/pull/10026#discussion_r1023512856

PR #10107, opened 2022-11-22
[npud] Do not handle SIGSEGV signal

jyoungyun:
This commit removes the SIGSEGV signal handler.
When program segfaults, the Linux kernel will sometimes write a core dump to disk. To use this operation, it does not handle SIGSEGV signal.


hseok-oh (approved):

glistening (approved):
LGTM

Issue #10098, opened 2022-11-22
[one-quantize] Adjust minimum scale in int16 quantization


jinevening:
## What

Let's adjust minimum scale in int16 quantization

## Why

Currently, the minimum scale of int16 quant is 1e-9. This can cause value overflow in some cases.

Overflow can happen according to the model (more specifically, scale_in, scale_weights, and bias value), so it would be helpful to add an option to control scales.

jinevening:
This issue is handled by backend.

PR #10111, opened 2022-11-23
[luci] Update minimum scale of int16 quantization

jinevening:
This updates minimum scale of int16 quantization.


jinevening:
This can reduce overflow of bias in our benchmark models.

We may have a smarter way than hard-coding of minimum scale. Let's discuss in https://github.com/Samsung/ONE/issues/10098

seanshpark (approved):
LGTM

PR #9273, opened 2022-06-17
[one-cmds] Add one-quantize test

mhs4670go:
This commit adds `one-quantize` test. This test tests if `circle-eval-diff` supports directory input.

Related: #9224 
ONE-DCO-1.0-Signed-off-by: seongwoo <mhs4670go@naver.com>

PR #10112, opened 2022-11-23
[circle-eval-diff] Fix input data size

jinevening:
This fixes byte size of directory input.

### DIFF
--- a/compiler/circle-eval-diff/src/InputDataLoader.cpp
+++ b/compiler/circle-eval-diff/src/InputDataLoader.cpp
@@ -55,7 +55,7 @@ std::vector<size_t> getEachByteSizeOf(const std::vector<loco::Node *> &nodes)
   for (const auto node : nodes)
   {
     const auto input_node = loco::must_cast<const luci::CircleInput *>(node);
-    size_t element_size = 1;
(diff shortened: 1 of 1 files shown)


jinevening:
We'll need tests for this kind of bug 😭

mhs4670go:
Could you change the codes like below?
```cxx
std::vector<size_t> getEachByteSizeOf(const std::vector<loco::Node *> &nodes)
{
  std::vector<size_t> vec;
  for (const auto node : nodes)
  {
    const auto input_node = loco::must_cast<const luci::CircleInput *>(node);
    size_t element_size = 1;

    for (uint32_t index = 0; index < input_node->rank(); index++)
    {
      element_size *= input_node->dim(index).value();
    }
    vec.push_back(element_size * loco::size(input_node->dtype())); // note here
  }
  return vec;
}
```

mhs4670go:
> 

Well, yeah.. Even though I've added https://github.com/Samsung/ONE/pull/9273, it seems not enough.

jinevening:
@mhs4670go I make a const variable for `loco::size(input_node->dtype())` for readability. PTAL :)

mhs4670go (approved):
LGTM

Issue #10088, opened 2022-11-22
CFE module to provide kernel(compute) codes


seanshpark:
This issue is added to discuss to provide(separate) kernel(compute) codes to reduce module responsibility and to provide common module for it's purpose.

From https://github.com/Samsung/ONE/pull/10082#discussion_r1028644584 and https://github.com/Samsung/ONE/issues/10066#issuecomment-1322766629

seanshpark:
Current usage
- compute codes for luci/pass
- compute codes for luci shape inference
- maybe compute codes for luci-interpreter

seanshpark:
as we have `compute` for `onert`, I'd like to go with `luci-compute` for the name

@lemmaa , any comments?

seanshpark:
First step;
- relocate include headers (e.g. mio-tflit280)

seanshpark:
Resume to this old issue :)

seanshpark:
Current 'luci/pass` that directly use tflite kernels
- `compiler/luci/pass/src/FoldDepthwiseConv2DPass.cpp`
- `compiler/luci/pass/src/FoldFullyConnectedPass.cpp`

seanshpark:
`luci-interpreter` kernel has `pal` layer for multiple implementations...
- it's getting dizzy to think also for this
- so I'm just leaving task to unify with `luci-interpreter` to next step

seanshpark:
Items of luci-compute came up while extracting compute codes from luci/pass
- add unit tests (#12150)
- introduce `Tensor` class with dtype like luci-interpreter (https://github.com/Samsung/ONE/pull/12155#discussion_r1410160095)

seanshpark:
done

PR #10121, opened 2022-11-24
DRAFT luci-compute

seanshpark:
on-going draft to relocate computation related to luci-compute.


seanshpark:
for #10088

seanshpark:
done

PR #10122, opened 2022-11-25
[luci-compute] Introduce luci-compute

seanshpark:
This will introduce luci-compute for computation kernels.


jinevening:
How about renaming the module to `tf-compute` or move `luci-compute` under luci?

seanshpark:
I have plan to use this in _luci-interpreter_ so it would be appropriate to put this in the compile folder. This module won't use ONLY tensorflow kernels but also other computation codes if necessary so I don't want to use `tf` as the prefix.

jinevening:
Thanks for the explanation. I was just curious why it has prefix `luci`, but not placed under `luci`.

jinevening (approved):

mhs4670go (approved):
LGTM

PR #10128, opened 2022-11-25
[infra] Use 20221125 preset in debian

seanshpark:
This will revise debian rules file to use 20221125 preset.


seanshpark:

mhs4670go (approved):
LGTM

PR #10129, opened 2022-11-25
[infra] Revise Makefile.arm32

seanshpark:
This will revise nncc Makefile.arm32 to include luci-compute.


seanshpark:

mhs4670go (approved):
LGTM

PR #10132, opened 2022-11-25
[DRAFT][record-minmax] Add parallel recording for h5 input data format 

BalyshevArtem:
This commit adds parallel recording for h5 input data format to record-minmax tool via OpenMP tool.

for issue #10133


BalyshevArtem:
results see in  https://github.com/Samsung/ONE/issues/10133#issuecomment-1327288302

mhs4670go:
Seems very good feature to me:)

Could I ask a few questions?

1. Do you have any test plan? Since `record-minmax` tool has something to do with the computation, unit testing or integration testing matter.
2. Is it okay with the other platform like ARM? I'm not sure but if we have to support ARM platform someday, it becomes matter. Like whether it can be built on ARM or not. If not, we should consider possibility of turning off the OpenMP option.

compiler/record-minmax/driver/Driver.cpp
```diff
@@ -50,6 +50,10 @@ int entry(const int argc, char **argv)
     .type(arser::DataType::FLOAT)
     .help("Record n'th percentile of min");
 
+  arser.add_argument("--parallel_record")
```
mhs4670go, line 53: (optional) Just `--parallel` seems good option since this driver already "record"s min/max.
```suggestion
  arser.add_argument("--parallel")
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -373,32 +402,108 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+
+        // TODO: Input data is copied twice (file -> buffer (input_data) -> interpreter inputs)
+        //       We can reduce the copy by directly writing data from file to interpreter inputs
+        const auto &cur_input_data = vector_input_data[record_idx][input_idx];
+        _vector_interpreters[current_thread_num]->writeInputTensor(
+          input_node, cur_input_data.data(), cur_input_data.size());
+      }
+      _vector_interpreters[current_thread_num]->interpret();
+    }
+  }
```
mhs4670go, line 488: This would make it have better readability.
```suggestion
  } // #pragma omp parallel
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -373,32 +402,108 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+  assert(_vector_interpreters.size() == threads_size);
+  assert(_vector_observers.size() == threads_size);
+  const auto vector_input_data = importH5Data(input_data_path);
+  const auto num_records = vector_input_data.size();
+  const auto input_nodes = loco::input_nodes(_module->graph());
+  const auto num_inputs = input_nodes.size();
+
+#pragma omp parallel
+  {
+#pragma omp for
```
mhs4670go, line 472: FYI, seems just `omp parallel for` does the same thing.
```suggestion
#pragma omp parallel for
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -209,6 +215,28 @@ void RecordMinMax::initialize(const std::string &input_model_path)
...
+{
+  initializeCircleModule(input_model_path);
+
+  // Create and initialize interpreters and observers
+  const auto threads_size = omp_get_max_threads();
+
+  _vector_interpreters.resize(threads_size);
+  _vector_observers.resize(threads_size);
+
+  for (int t = 0; t < threads_size; ++t)
```
mhs4670go, line 228: How about use `uint32_t` instead of `int` for safety? 
```suggestion
  for (uint_32_t t = 0; t < threads_size; ++t)
```
This can be applied to below for loops as well.

compiler/record-minmax/include/RecordMinMax.h
```diff
@@ -51,9 +56,17 @@ class RecordMinMax
...
+
+  std::vector<std::vector<std::vector<char>>> importH5Data(const std::string &input_data_path);
+
   std::unique_ptr<luci::Module> _module;
   std::unique_ptr<luci_interpreter::Interpreter> _interpreter;
   std::unique_ptr<MinMaxObserver> _observer;
+
+  // Used in parallel execution
+  std::vector<std::unique_ptr<luci_interpreter::Interpreter>> _vector_interpreters;
+  std::vector<std::unique_ptr<MinMaxObserver>> _vector_observers;
```
mhs4670go, line 69: How about name them without `vector` prefix? The data type already let us know the information.
```suggestion
  std::vector<std::unique_ptr<luci_interpreter::Interpreter>> _interpreters;
  std::vector<std::unique_ptr<MinMaxObserver>> _observers;
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -209,6 +215,28 @@ void RecordMinMax::initialize(const std::string &input_model_path)
   _interpreter->attachObserver(_observer.get());
 }
 
+void RecordMinMax::initialize_with_parallel_record(const std::string &input_model_path)
```
mhs4670go, line 218: How about not introduce this function? And we can only use introducing vector containers like `_interpreters` and `observers` instead of using existing `_interpreter` and `_oberver`. Then, codes will be look like below.
```cxx
// We could introduce dedicated context variable instead
void RecordMinMax::initialize(const std::string &input_model_path, bool in_parallel) 
// void RecordMinMax::initialize(const RMMConext ctx)
{
  initializeCircleModule(input_model_path);

  // Create and initialize interpreters and observers
  const auto threads_size = in_parallel ? omp_get_max_threads() : 1;

  _interpreters.resize(threads_size);
  _observers.resize(threads_size);

  for (uint32_t thread_idx = 0; thread_idx < threads_size; ++thread_idx)
  {
    auto interpreter = std::make_unique<luci_interpreter::Interpreter>(_module.get());
    auto observer = std::make_unique<MinMaxObserver>();

    interpreter->attachObserver(observer.get());

    _observers[thread_idx] = std::move(observer);
    _interpreters[thread_idx] = std::move(interpreter);
  }
}
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -373,32 +402,108 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+  std::cout << "Recording finished. Number of recorded data: " << num_records << std::endl;
+
+  update_quantparam(_observer.get(), mode, min_percentile, max_percentile);
+}
+
+void RecordMinMax::profileData_with_parallel_record(const std::string &mode,
+                                                    const std::string &input_data_path,
+                                                    float min_percentile, float max_percentile)
+{
+  const auto threads_size = omp_get_max_threads();
```
mhs4670go, line 462: (optional) `thread_size` can be an attribute of the `RecordMinMax` class.

compiler/record-minmax/src/RecordMinMax.cpp
BalyshevArtem: done

compiler/record-minmax/src/RecordMinMax.cpp
BalyshevArtem: done

BalyshevArtem:
> 1. Do you have any test plan? Since `record-minmax` tool has something to do with the computation, unit testing or integration testing matter

added pota_parallel_record_minmax_test in pota-quantization-value-test

compiler/pota-quantization-value-test/CMakeLists.txt
```diff
@@ -92,6 +92,14 @@ add_test(
           ${QUANTIZATION_VALUE_TEST_WITH_PARAM}
 )
 
+add_test(
```
seanshpark, line 95: Please split `pota-` and `record-` modules.
@binarman , I am keep saying this to SRR members. Why do you not listen to this?

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -335,11 +349,11 @@ void RecordMinMax::profileRawData(const std::string &mode, const std::string &in
 
   std::cout << "Recording finished. Number of recorded data: " << num_records << std::endl;
 
-  update_quantparam(_observer.get(), mode, min_percentile, max_percentile);
+  update_quantparam(_observers[0].get(), mode, min_percentile, max_percentile);
 }
 
-void RecordMinMax::profileData(const std::string &mode, const std::string &input_data_path,
-                               float min_percentile, float max_percentile)
+std::vector<std::vector<std::vector<char>>>
```
seanshpark, line 355: Please split this PR by purpose
- revise to prepare to multiple instance
- use openmp for running in multiple instance

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -355,13 +369,14 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+    std::vector<std::vector<std::vector<char>>> vector_input_data(num_records);
+
+    // Read inputs to vector_input_data
+    for (int i = 0; i < num_records; ++i)
     {
-      if (num_inputs != importer.numInputs(record_idx))
+      if (num_inputs != importer.numInputs(i))
         throw std::runtime_error("Wrong number of inputs.");
 
-      std::cout << "Recording " << record_idx << "'th data" << std::endl;
```
seanshpark, line 363: this log was added for CI not to stop for long running with no response from the program.
I'm not sure current CI requires this but I don't want this log to be removed.

As I wrote in previous comment, this PR is quite hard to read from multiple purpose of changes and especially removal of the log. Ad I am not willing to approve this by this.

compiler/record-minmax/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ target_link_libraries(record-minmax dio_hdf5)
 target_link_libraries(record-minmax vconone)
 target_link_libraries(record-minmax nncc_coverage)
 
+if (_OPENMP)
+  find_package(OpenMP REQUIRED)
```
seanshpark, line 19: Please add this to see in CI if OpenMP is really enabled.
```
message(STATUS "record-minmax with OpenMP")
```

compiler/record-minmax/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ target_link_libraries(record-minmax dio_hdf5)
 target_link_libraries(record-minmax vconone)
 target_link_libraries(record-minmax nncc_coverage)
 
+if (_OPENMP)
```
seanshpark, line 18: where is `_OPENMP` defined?

compiler/record-minmax/driver/Driver.cpp
```diff
@@ -50,6 +50,8 @@ int entry(const int argc, char **argv)
     .type(arser::DataType::FLOAT)
     .help("Record n'th percentile of min");
 
+  arser.add_argument("--parallel").type(arser::DataType::BOOL).help("Use parallel recording");
```
jinevening, line 53:
```suggestion
  arser.add_argument("--parallel").type(arser::DataType::BOOL).help("Use parallel recording (default: false)");
```

compiler/record-minmax/include/RecordMinMax.h
```diff
@@ -39,6 +49,9 @@ class RecordMinMax
   void profileData(const std::string &mode, const std::string &input_data_path,
                    float min_percentile, float max_percentile);
 
+  void profileData_with_parallel_record(const std::string &mode, const std::string &input_data_path,
```
jinevening, line 52: Please use consistent naming.

```suggestion
  void profileDataInParallel(const std::string &mode, const std::string &input_data_path,
```

compiler/record-minmax/include/RecordMinMax.h
```diff
@@ -51,9 +64,17 @@ class RecordMinMax
...
 private:
+  void initializeCircleModule(const std::string &input_model_path);
+
+  std::vector<std::vector<std::vector<char>>> importH5Data(const std::string &input_data_path);
+
   std::unique_ptr<luci::Module> _module;
-  std::unique_ptr<luci_interpreter::Interpreter> _interpreter;
-  std::unique_ptr<MinMaxObserver> _observer;
+
+  // Used in parallel execution
```
jinevening, line 73:
```suggestion
  // Multiple interpreters are used for parallel execution
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -373,34 +388,106 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
     throw std::runtime_error("HDF5 error occurred.");
   }
+}
+
+void RecordMinMax::profileData(const std::string &mode, const std::string &input_data_path,
+                               float min_percentile, float max_percentile)
+{
+  const auto input_nodes = loco::input_nodes(_module->graph());
+  const auto num_inputs = input_nodes.size();
+  const auto vector_input_data = importH5Data(input_data_path);
```
jinevening, line 419: This PR loads all input data on the memory at once, which is vulnerable to OOM (out of memory).

Can you make it more memory-efficiently? E.g., processing data one by one as the previous code did.

jinevening: At least, please keep the previous behavior of the single-thread recording.

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -373,34 +388,106 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+#pragma omp parallel for
+  for (uint32_t record_idx = 0; record_idx < num_records; record_idx++)
+  {
+    const auto current_thread_num = omp_get_thread_num();
+    for (int32_t input_idx = 0; input_idx < num_inputs; input_idx++)
+    {
+      const auto *input_node = loco::must_cast<const luci::CircleInput *>(input_nodes[input_idx]);
+
+      // TODO: Input data is copied twice (file -> buffer (input_data) -> interpreter inputs)
+      //       We can reduce the copy by directly writing data from file to interpreter inputs
```
jinevening, line 464: The comments can be removed.

```suggestion
      // TODO: Input data is copied twice (file -> buffer (input_data) -> interpreter inputs)
      //       We can reduce the copy by directly writing data from file to interpreter inputs
```

compiler/record-minmax/include/MinMaxObserver.h
```diff
@@ -43,6 +43,15 @@ class MinMaxMap
     vectors.max_vector.push_back(max);
   }
 
+  void recordMinMaxVector(const luci::CircleNode *node, const MinMaxVectors &minmax_vector)
```
jinevening, line 46:
```suggestion
  void appendMinMaxVector(const luci::CircleNode *node, const MinMaxVectors &minmax_vector)
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -373,34 +388,106 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+
+  for (const auto &obs : _observers)
+  {
+    const auto cur_minmax_map = obs->minMaxData()->getMap();
+    for (auto &iter : *cur_minmax_map)
+    {
+      const auto node = iter.first;
+      const auto &minmax = iter.second;
+
+      main_min_max_map->recordMinMaxVector(node, minmax);
```
jinevening, line 484:
```suggestion
      main_min_max_map->appendMinMaxVector(node, minmax);
```

compiler/record-minmax/driver/Driver.cpp
```diff
@@ -113,17 +119,35 @@ int entry(const int argc, char **argv)
...
+
     if (input_data_format == "h5" || input_data_format == "hdf5")
     {
       // Profile min/max while executing the H5 data
-      rmm.profileData(mode, input_data_path, min_percentile, max_percentile);
+      if (not parallel_record)
+        rmm.profileData(mode, input_data_path, min_percentile, max_percentile);
+      else
+      {
+        std::cout << "Using parallel recording" << std::endl;
```
jinevening, line 148: This log will be printed even when `_OPENMP` is not defined but a user gives `--parallel true`.

I think warning message should be printed at that case, e.g., `"Parallel recording is not available in this platform. A single thread is used for recording."`

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -373,34 +388,106 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+  const auto input_nodes = loco::input_nodes(_module->graph());
+  const auto num_inputs = input_nodes.size();
+  const auto vector_input_data = importH5Data(input_data_path);
+  const auto num_records = vector_input_data.size();
+
+  for (int32_t record_idx = 0; record_idx < num_records; record_idx++)
+  {
+    std::cout << "Recording " << record_idx << "'th data" << std::endl;
+
+    for (int32_t input_idx = 0; input_idx < num_inputs; input_idx++)
```
BalyshevArtem, line 426: `std::cout << "Recording " << record_idx << "'th data" << std::endl;`
for https://github.com/Samsung/ONE/pull/10132#discussion_r1034141358 comment

compiler/record-minmax/src/RecordMinMax.cpp
BalyshevArtem: > At least, please keep the previous behavior of the single-thread recording.

Done

compiler/record-minmax-thread-safety-test/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+# Read "test.local.lst" if exists
+include("test.local.lst" OPTIONAL)
+
+unset(TEST_DEPS)
+
+get_target_property(ARTIFACTS_BIN_PATH testDataGenerator BINARY_DIR)
+
+# Add thread sanitizer to record-minmax
+target_compile_options(record-minmax PUBLIC -fsanitize=thread)
+target_link_options(record-minmax PUBLIC -fsanitize=thread)
```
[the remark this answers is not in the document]
jinevening: Our cmake version is lower than 3.13, where "target_link_options" is introduced.

I didn't test, but please try with `target_link_libraries`.

```
target_link_libraries (record-minmax -fsanitize=thread)
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -400,7 +493,67 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+    }
+    else
+    {
+      threads.emplace_back(interpret_batch_records, records_batch * t, num_records,
+                           _interpreters[t].get(), vector_input_data, input_nodes);
+    }
+  }
+
+  for (int i = 0; i < _threads_size; ++i)
+    threads.at(i).join();
```
jinevening, line 534: Multi-threading part.

I guess that `interpret_batch_records` can be defined in this function to reduce parameters. For example,

```cpp
auto interpret_batch_records = [&vector_input_data, &input_nodes, &batch_size] (Interpreter *interp, uint32_t batch_index) {
...
};

```

compiler/record-minmax/include/RecordMinMax.h
```diff
@@ -23,14 +23,19 @@
...
 
 namespace record_minmax
 {
 
 class RecordMinMax
 {
 public:
-  explicit RecordMinMax() = default;
+  explicit RecordMinMax(bool use_parallel)
+    : _threads_size(use_parallel ? std::thread::hardware_concurrency() : 1)
```
jinevening, line 35: Better to make the number of threads configurable in the command line option. Perhaps it would be better to replace `--parallel` with `--num_threads <int>` (default: 1).

`std::thread::hardware_concurrency()` can return 0, and it will cause problem in release mode. In addition, I'm not sure the function works as expected in CI machine.

compiler/record-minmax-thread-safety-test/CMakeLists.txt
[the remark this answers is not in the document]
BalyshevArtem: Another problem 

```
Attempt to add link library "-fsanitize=thread" to target "record-minmax"
  which is not built in this directory.

  This is allowed only when policy CMP0079 is set to NEW.
```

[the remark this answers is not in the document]
jinevening: Can you move the code to Cmake file in record-minmax?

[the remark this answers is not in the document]
BalyshevArtem: Done

compiler/record-minmax/src/RecordMinMax.cpp
BalyshevArtem: Done

compiler/record-minmax/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ target_link_libraries(record-minmax luci_interpreter)
 target_link_libraries(record-minmax dio_hdf5)
 target_link_libraries(record-minmax vconone)
 target_link_libraries(record-minmax nncc_coverage)
+target_compile_options(record-minmax PUBLIC -fsanitize=thread)
+target_link_libraries(record-minmax -fsanitize=thread)
```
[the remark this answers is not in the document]
BalyshevArtem: Done

compiler/record-minmax/CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ if(NOT ENABLE_TEST)
   return()
 endif(NOT ENABLE_TEST)
 
+target_compile_options(record-minmax PUBLIC -fsanitize=thread)
+target_link_libraries(record-minmax -fsanitize=thread)
+
```
jinevening, line 26: Ah, this will make record-minmax with sanitizer for all other cases. Could you make a separate executable for thread  test? For example,
 
```suggestion
# record-minmax-for-thread-test is used in record-minmax-thread-safety-test
add_executable(record-minmax-for-thread-test ${DRIVER} ${SOURCES})
target_include_directories(record-minmax-for-thread-test PRIVATE include)
target_link_libraries(record-minmax-for-thread-test arser)
...
target_compile_options(record-minmax-for-thread-test PUBLIC -fsanitize=thread)
target_link_libraries(record-minmax-for-thread-test -fsanitize=thread)
```

You may use `record-minmax-for-thread-test` in `record-minmax-thread-safety-test`.

jinevening: I guess this will resolve CI failure.

BTW, it seems that thread sanitizer found an issue in luci-interpereter. I made https://github.com/Samsung/ONE/issues/10188 for that issue.

BalyshevArtem: Done

BalyshevArtem:
@jinevening,
Can I split this draft into PRs? or I need to add something else?

BalyshevArtem:
All done close it

PR #10135, opened 2022-11-28
[npud] Implement create and destroy request in core class

jyoungyun:
The Core function invokes the dev manager to create and destroy request. The npu daemon will do the job with a given model id using backend library.

Draft PR: #9989

### DIFF
--- a/runtime/service/npud/core/Core.cc
+++ b/runtime/service/npud/core/Core.cc
@@ -118,5 +118,51 @@ int Core::destroyNetwork(ContextID contextId, ModelID modelId) const
   return 0;
 }
 
+int Core::createRequest(ContextID contextId, ModelID modelId, RequestID *requestId) const
+{
+  VERBOSE(Core) << "createRequest with " << contextId << ", " << modelId << std::endl;
+  NpuContext *npuContext = _contextManager->getNpuContext(contextId);
+  if (!npuContext)
+  {
+    VERBOSE(Core) << "Invalid context id" << std::endl;
+    // TODO Define CoreStatus
+    return 1;
+  }
+
+  RequestID id;
+  int ret = _devManager->createRequest(npuContext, modelId, &id);
+  if (ret != NPU_STATUS_SUCCESS)
+  {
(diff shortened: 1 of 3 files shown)


jyoungyun:

runtime/service/npud/core/Core.cc
```diff
@@ -118,5 +118,51 @@ int Core::destroyNetwork(ContextID contextId, ModelID modelId) const
   return 0;
 }
 
+int Core::createRequest(ContextID contextId, ModelID modelId, RequestID *requestId) const
+{
+  VERBOSE(Core) << "createRequest with " << contextId << ", " << modelId << std::endl;
```
glistening, line 123: I think `Core` is too broad. There is much modules in `ONE`. But it may be okay since `npud` is a separate process. It is not likely to interleave log messages from `onert`.

hseok-oh (approved):
LGTM.

Please check @glistening 's comment.

runtime/service/npud/core/Core.cc
jyoungyun: I couldn't catch your concerns.. Do you think this verbose log tag should be defined as a function/feature name? I have defined a verbose log tag by associating them with class name.. I will think more about what you said.

jyoungyun: > What core? `onert` core? `dbus` core? 

It is `npud` core. As far as I imagine,,,, `executor`, `scheduler`, etc. will be added to this `core` class and the `core` will operate all of this. I will find more specific name for this. :)

glistening: I thought it would be better to avoid using same name in log tag. But I am not sure that `npud` log messages and `onert` log messages will not be interleaved, I am unsure because I still don't understand what `npud` is for. Is it for managing `npu` (scheduling) ? (I think it is not the role of runtime. It would be better to locate out of `runtime` like `onert-micro`. Or is `npud` a kind of runtime? (Then, I think the name is misnomer).

glistening (approved):
LGTM

Issue #10018, opened 2022-11-15
[luci] Multiple output type nodes validation


seanshpark:
We have several circle operators having virtual operators for multiple outputs; like CircleUnpack, CircleSplit.

While working on unrolling LSTM, found out that there was no validation that CirleUnpackOut must follow CircleUnpack.

Let's add some checks for these kind.

seanshpark:
from #9940

seanshpark:
done

PR #10138, opened 2022-11-29
DRAFT luci graph validate for multi-outs

seanshpark:
on-going draft to validate multiple output nodes.


seanshpark:
for #10018

seanshpark:
done

PR #10141, opened 2022-11-29
[infra] Revise default preset

seanshpark:
This will revise default preset to 20221125.


seanshpark:
for #10088, to fix  #10013

hseok-oh (approved):
LGTM

PR #10143, opened 2022-11-29
[luci/svc] Implement validate_multi_outs

seanshpark:
This will implement validate_multi_outs method to check multiple output IRs has valid successor type.


seanshpark:
for  #10018 , from draft #10138

jinevening (approved):
LGTM

Issue #10144, opened 2022-11-29
[luci/partition] Fix static analysis warning


seanshpark:
This warning just came up from our internal static analysis tool; but to code itself has landed long time ago. Not sure this is really a warning.

> Return value of a function 'loco::NamedEntity::name[abi:cxx11]' is dereferenced at PartitionPModules.cpp:95 without checking for null, but it is usually checked for this function

seanshpark:
This warning was gone before the patch landed.

PR #10145, opened 2022-11-29
[luci/partition] Use local variable for clone_graph

seanshpark:
This will revise to use local variables in clone_graph method, hope to fix static analysis warnings.


seanshpark:
#10144

not sure this change will fix the warning. have to see...

jinevening (approved):
LGTM

Issue #9977, opened 2022-11-03
[one-quantize] Improve interface for quantization dataset of multi-input models


jinevening:
## What

Let's discuss interface for quantization dataset of multi-input models.

## Why

Currently, multi-input data should be concatenated for `directory` (or `list`) format. This may not be intuitive to users.

jinevening:
To use separate files for multiple inputs, we may need "**naming convention**".

Under the specified directory, data can be saved as below.

`<data_index>.<input_index>`

`data_index`: Index of data
`input_index`: Index of input

For example, if there are three data for the model with two inputs, data will be saved as below.

```
/0.0
/0.1
/1.0
/1.1
/2.0
/2.1
```

NOTE data_index can be replaced with a string unique for each data. For example, `fish.0`, `fish.1`, `cow.0`, `cow.1` ...

For backward compatibility, we'll need a logic to fallback to the single-input format unless data is saved in the multi-input-supported format.

staddy:
Test data:
[img_files_enumerated.zip](https://github.com/Samsung/ONE/files/10154290/img_files_enumerated.zip)
[add_299_299_3.zip](https://github.com/Samsung/ONE/files/10154291/add_299_299_3.zip)

staddy:
I would like to work on this issue.

jinevening:
Previous interface.

- Data for multiple inputs have to be concatenated in the correct order. https://github.com/Samsung/ONE/blob/master/compiler/record-minmax/driver/Driver.cpp#L129-L135

```cpp
    // Each data file is composed of inputs of a model, concatenated in
    // the same order with the input index of the model
    //
    // For example, for a model with n inputs, the contents of each data
    // file can be visualized as below
    // [input 1][input 2]...[input n]
    // |start............end of file|
```

Proposed interface

- Data for multiple inputs are separate files, but they should have pre-defined naming format (https://github.com/Samsung/ONE/issues/9977#issuecomment-1316620779).

`<data_name>.<input_idx>`

@seanshpark proposed a datalist format.

datalist.txt
```
car.data, box.data // first and second inputs
dog.data cat.data
...
```

Data list format also looks good to me, but our client complained that it is difficult to use the list format in the docker environment.

@seanshpark @lemmaa Could you give any suggestion?

seanshpark:
>  but our client complained

As you told me, `client` is SDK developer team. I hope they can solve the issue :)
Maybe we need some conditions such as the list and the data files should have same parent folder.

[...]

PR #10164, opened 2022-12-05
[record-minmax] New format for multi-input models input data

staddy:
This PR adds support for new format for multi-input models input data.
See related issue https://github.com/Samsung/ONE/issues/9977 for details.

ONE-DCO-1.0-Signed-off-by: Stanislav Ponkrashov <s.ponkrashov@samsung.com>

### DIFF
--- a/compiler/one-cmds/tests/prepare_test_materials.sh
+++ b/compiler/one-cmds/tests/prepare_test_materials.sh
@@ -57,6 +57,24 @@ if [ ! -d "raw_files" ] || [ ! -s "datalist.txt" ]; then
     ../bin/venv/bin/python preprocess_images.py
 fi
 
+if [[ ! -s "img_files_enumerated" ]]; then
+    rm -rf img_files_enumerated.zip
+    wget https://github.com/Samsung/ONE/files/10154290/img_files_enumerated.zip
+    unzip img_files_enumerated.zip
+    # https://github.com/Samsung/ONE/issues/9977#issuecomment-1337294959
+fi
+
+if [[ ! -s "add_299_299_3.circle" ]]; then
+    rm -rf add_299_299_3.zip
+    wget https://github.com/Samsung/ONE/files/10154291/add_299_299_3.zip
+    unzip add_299_299_3.zip
+    # https://github.com/Samsung/ONE/issues/9977#issuecomment-1337294959
+fi
+
+if [ ! -d "raw_files_enumerated" ] || [ ! -s "datalist_enumerated.txt" ]; then
+    ../bin/venv/bin/python preprocess_images.py --input_dir img_files_enumerated --output_dir raw_files_enumerated --list_file datalist_enumerated.txt --output_extension ''
+fi
+
 if [[ ! -s "inception_v3_test_data.h5" ]]; then
   ../bin/venv/bin/python ../bin/rawdata2hdf5 \
   --data_list datalist.txt \
--- a/compiler/record-minmax/src/RecordMinMax.cpp
+++ b/compiler/record-minmax/src/RecordMinMax.cpp
@@ -32,6 +32,7 @@
 #include <stdexcept>
 #include <iostream>
 #include <random>
+#include <unordered_set>
 
 using Shape = std::vector<loco::Dimension>;
 using DataType = loco::DataType;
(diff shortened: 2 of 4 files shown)


compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -32,6 +32,7 @@
 #include <stdexcept>
 #include <iostream>
 #include <random>
+#include <unordered_set>
```
seanshpark, line 35: please split record-minmax and one-cmds in separate PRs

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -164,6 +165,50 @@ void update_quantparam(record_minmax::MinMaxObserver *observer, const std::strin
   }
 }
 
+std::unordered_set<std::string> list_regular_files(DIR *directory)
+{
+  std::unordered_set<std::string> regular_files;
```
seanshpark, line 170: add assert check for `directory`

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -164,6 +165,50 @@ void update_quantparam(record_minmax::MinMaxObserver *observer, const std::strin
   }
 }
 
+std::unordered_set<std::string> list_regular_files(DIR *directory)
```
seanshpark, line 168: any particular reason to must use `unordered_set` ?

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -164,6 +165,50 @@ void update_quantparam(record_minmax::MinMaxObserver *observer, const std::strin
...
+    for (i = 0; i < count; ++i)
+      if (files.find(without_index + std::to_string(i)) == files.end())
+        break;
+    if (i >= count)
+      enumerated.emplace(std::move(without_index));
+    else
+      excluded.emplace(std::move(without_index));
+  }
+  return enumerated;
+}
```
seanshpark, line 211: please add empty line after this

compiler/one-cmds/tests/prepare_test_materials.sh
```diff
@@ -57,6 +57,24 @@ if [ ! -d "raw_files" ] || [ ! -s "datalist.txt" ]; then
...
+
+if [[ ! -s "add_299_299_3.circle" ]]; then
+    rm -rf add_299_299_3.zip
+    wget https://github.com/Samsung/ONE/files/10154291/add_299_299_3.zip
+    unzip add_299_299_3.zip
+    # https://github.com/Samsung/ONE/issues/9977#issuecomment-1337294959
+fi
+
+if [ ! -d "raw_files_enumerated" ] || [ ! -s "datalist_enumerated.txt" ]; then
+    ../bin/venv/bin/python preprocess_images.py --input_dir img_files_enumerated --output_dir raw_files_enumerated --list_file datalist_enumerated.txt --output_extension ''
```
seanshpark, line 75: please split this line so that comes in to 100 columns

seanshpark:
@jinevening , I don't fully understand about quantization with files in the folder.
does this change existing feature or does this add a new feature for multiple files?

seanshpark:
> See related issue https://github.com/Samsung/ONE/issues/9977 for details.

Please remove this from commit message.

seanshpark:
> The new implemented behavior is:

@staddy , don't get me wrong...
I am asking to @jinevening to understand about this situation:
- users expect something from previous usage experience
- this feature changes default behavior
- output is different
- users get confused or panic

compiler/record-minmax/src/RecordMinMax.cpp
staddy: As the function will be used uniquely in this file it seems reasonable to return a set, because we need a set anyway to implement readable and fast search of file names. std::unordered_set seemingly fits the task best, but the performance boost compared to std::set is insufficient (if there is any), so I can use std::set instead for the sake of readability.

jinevening:
> I am asking to @jinevening to understand about this situation:
users expect something from previous usage experience
this feature changes default behavior
output is different
users get confused or panic

@seanshpark Thanks for notice. I agree with your concern. This feature has to be implemented while guaranteeing backward compatibility, as mentioned in https://github.com/Samsung/ONE/issues/9977#issuecomment-1316620779.

> For backward compatibility, we'll need a logic to fallback to the single-input format unless data is saved in the multi-input-supported format.

This PR works as follows.

- Find multi-input format files -> Record multi-input data -> Read rest of the files -> Record the rest of data

This way affects the previous behavior and allows mixed use of file formats.

It would be better to split the process.

1. Check number of inputs of circle model
2. Check files are saved in a pre-defined format (let's say canonical format)
3. If canonical format,
  -  Record with the canonical format
4. Else (if not)
  - Record with the previous format

---

BTW, we first need consensus for the interface in https://github.com/Samsung/ONE/issues/9977#issuecomment-1316620779. @staddy Could you proceed this work after we (@seanshpark @lemmaa ) all agree to the interface?

staddy:
> @staddy Could you proceed this work after we (@seanshpark @lemmaa ) all agree to the interface?

Sure

seanshpark:

Issue #9961, opened 2022-10-31
[circle-opselector] Revisit Driver.cpp


mhs4670go:
## What

Let's revisit `Driver.cpp`.

## Why

Current `Driver.cpp` is too complicated to read easily. So, let's revisit it by making functions and dedicated classes.

PR #10165, opened 2022-12-05
[circle-opselector] Revisit Driver

mhs4670go:
This commit revists Driver codes.

I introduced a dedicated class for op-selecting instead of functions. 

If you look at how `Driver.cpp` has changed, you should know the impact of the class.

Related: #9961 


compiler/circle-opselector/driver/Driver.cpp
```diff
@@ -37,6 +38,7 @@ void print_version(void)
   std::cout << vconone::get_copyright() << std::endl;
 }
 
+#if 0
```
seanshpark, line 41: (optional) add `TODO remove this block`

compiler/circle-opselector/src/OpSelector.test.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    output()->from(_dconv);
+  }
+
+  std::unique_ptr<loco::Graph> graph(void) { return std::move(_g); }
+};
+
+} // namespace
+
+TEST(OpSelectorTest, select_by_name)
```
seanshpark, line 107: negative tests too please

compiler/circle-opselector/src/SelectType.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
```
seanshpark, line 2: we don't need to change this for existing file

seanshpark: maybe add and the delete old files in another PR.

seanshpark (approved):
LGTM thank you!

PR #10169, opened 2022-12-07
[onert/test] Update md5 check curl option

hseok-oh:
This commit updates curl option for md5sum check to use same with download.


glistening (approved):
LGTM

Issue #10154, opened 2022-12-02
Introduce BisectionSolver for mixed precision quantization.


stamalakhov:
### What

Introduce light-weight fast methods for finding quasi-optimal MP quantization. So let's start with `bisectioning` method.
### Why
We may need to build mixed precision quantized model in acceptable time aiming to achieve `admissible accuracy` with `reduced execution time`.

jinevening:
Lightweight MPQ will definitely improve the usability of our MPQ feature :) Could you explain the `quasi-optimal` aspect of `bisection` method?

stamalakhov:
> Lightweight MPQ will definitely improve the usability of our MPQ feature :) Could you explain the `quasi-optimal` aspect of `bisection` method?

`Bisection` assumes:
1) The `accuracy` can be parameterized with just a single parameter, e.g., its `depth` (`distance` from inputs nodes)
2) The `accuracy` is monotone with the neural network `depth`.
In this case `bisection` can provide solution in ~`ln(N) * T(inference)` (`N`- is maximal NN depth, `T(inference)` - time of full inference).

In general, these assumptions do not hold. Since NNs can be very heterogeneous they cannot be treated as one-dimensional structures. Moreover, initial training can be done using strong augmentation, which helps a lot to deal with quantization errors, so `accuracy` may be not monotone along NN's depth as well.
Summing up:
1) (speed) `bisection` - is fast enough
2) (`quasi-optimal` or `suboptimal`) but it may turn more Q8 layers into Q16 than needed to meet the `accuracy` requirement

stamalakhov:
@jinevening
Do we need to implement calling of circle-mpqsolver via onecc interface or direct calling is ok?
I mean for UI implementation in ONE-vscode it would be convenient to use onecc.

jinevening:
> Do we need to implement calling of circle-mpqsolver via onecc interface or direct calling is ok?

For now, direct calling is ok. `circle-mpqsolver` may be integrated with `onecc` when it is ready (working in more general scenarios). 

Do you have a plan to integrate circle-mpqsolver into ONE-vscode?  Could you describe your plan? Or it would be good to have a meeting about that item :)

stamalakhov:
> For now, direct calling is ok. `circle-mpqsolver` may be integrated with `onecc` when it is ready (working in more general scenarios).

Ok.



[...]

PR #10170, opened 2022-12-07
[DRAFT] [mpqsolver] Introduce bisection method

stamalakhov:
This commit introduces bisection method for the task of automatic mixed precision quantization.
Related to: #10154

The cornerstone of the currently implemented method is:
```
the whole network is assumed to be split into two adjacent parts (Q8 part and Q16 part) 
```
Algorithm proceeds as follows:
1) Determine which part is Q16 and which is Q8 (Q16 at _input_related_nodes_ or Q16 at _output_related_nodes_)
2) Compute 'depth' parameterization for input model (which is supposed to be '_recorded_')
3) Use '_bisectioning_' of 'depth' parameter to find _Accuracy(final_depth) == Target_

'_Mobile_net_v1_' output for 100 images:
```
int16_accuracy 0.000108095
int8_accuracy 0.000296601
target accuracy 0.000202348
290.08 vs 25.9079
16 : 0.000161125
8 : 0.00018704
4 : 0.000204438
6 : 0.000187962
5 : 0.000194962
4.5 : 0.000204438
Found optimal configuration at 5 depth.
Time elapsed is 7.3167 minutes.
```
'_Mobile_net_v2_' output for 100 images:
```
int16_accuracy 0.0161741
int8_accuracy 0.297486
target accuracy 0.128699
491.042 vs 50.6277
35 : 0.0927452
17.5 : 0.114215
8.75 : 0.138438
13.125 : 0.116353
10.9375 : 0.125074
9.84375 : 0.134745
10.3906 : 0.125074
Found optimal configuration at 10.3906 depth.
Time elapsed is 4.7333 minutes.
```
'_cartoon_' output for 100 images:
```
int16_accuracy 0.0072662
int8_accuracy 0.0943325
target accuracy 0.0420927
109.214 vs 178.843
31.5 : 0.0174134
47.25 : 0.0220028
55.125 : 0.0352334
59.0625 : 0.094363
57.0938 : 0.0675179
56.1094 : 0.0675179
55.6172 : 0.0352334
Found optimal configuration at 55.6172 depth.
Time elapsed is 16.583 minutes.
```
Note:
There is a lot of room for improvements, e.g. OpenMP parallelization in inference part.

Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

### DIFF
--- a/compiler/circle-mpqsolver/README.md
+++ b/compiler/circle-mpqsolver/README.md
@@ -1,0 +1,3 @@
+# circle-mpqsolver
+
+_circle-mpqsolver_ provides light-weight methods for finding Mixed Precision Quantization model.
--- a/compiler/circle-mpqsolver/src/BisectionSolver.cpp
+++ b/compiler/circle-mpqsolver/src/BisectionSolver.cpp
@@ -1,0 +1,287 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "BisectionSolver.h"
+#include "ErrorApproximator.h"
+#include "ModuleCloner.h"
+#include "Evaluator.h"
+
+#include <cmath>
+#include <iostream>
+
+using namespace mpqsolver;
+
+using NodeDepthType = std::map<luci::CircleNode *, float>;
+
+namespace
+{
+
+int compute_depth(const luci::Module *module, NodeDepthType &nodes_depth, float &min_depth,
+                  float &max_depth)
+{
+  if (module == nullptr)
+    return EXIT_FAILURE;
+
+  if (module->size() != 1)
+    return EXIT_FAILURE;
+
+  auto graph = module->graph(0);
+  if (!graph)
+    return EXIT_FAILURE;
+
+  auto nodes = graph->nodes();
+  uint32_t nodes_size = nodes->size();
+  std::set<std::string> input_names;
+  {
+    auto inp_nodes = graph->inputs();
+    for (uint32_t i = 0; i < inp_nodes->size(); ++i)
+    {
+      auto inp_node = inp_nodes->at(i);
+      auto inp_name = inp_node->name();
+      input_names.insert(inp_name);
+    }
+  }
+
+  // initializing
+  std::vector<luci::CircleNode *> to_process;
+  std::map<std::string, float> named_depth;
+  {
+    auto inputs = loco::input_nodes(graph);
+    for (auto &node : inputs)
+    {
+      auto cnode = loco::must_cast<luci::CircleNode *>(node);
+      to_process.emplace_back(cnode);
+      nodes_depth[cnode] = 0.f;
+      named_depth[cnode->name()] = 0.f;
+    }
+  }
+
+  // enumerating
+  while (!to_process.empty())
+  {
+    auto cur_node = to_process.back();
+    to_process.pop_back();
+    auto iter = nodes_depth.find(cur_node);
+    if (iter == nodes_depth.end())
+    {
+      return EXIT_FAILURE; // unexpected
+    }
+    float cur_depth = iter->second + 1;
+    // processing children
+    auto children = loco::succs(cur_node);
+    for (auto &child : children)
+    {
+      auto cichild = loco::must_cast<luci::CircleNode *>(child);
+      auto node_depth = nodes_depth.find(cichild);
+      if (node_depth == nodes_depth.end() || node_depth->second < cur_depth)
+      {
+        // initialize depth
+        nodes_depth[cichild] = cur_depth;
+        to_process.push_back(cichild);
+        named_depth[cichild->name()] = cur_depth;
+      }
+    }
+  }
+
+  auto minmax = std::minmax_element(
+    nodes_depth.begin(), nodes_depth.end(),
+    [=](const std::pair<luci::CircleNode *, float> &el1,
+        const std::pair<luci::CircleNode *, float> &el2) { return el1.second < el2.second; });
+
+  min_depth = minmax.first->second;
+  max_depth = minmax.second->second;
+
+  return EXIT_SUCCESS;
+}
+
+bool error_at_input_is_larger_than_at_output(const NodeDepthType &nodes_depth, float cut_depth)
+{
+  float error_at_input = 0;
+  float error_at_output = 0;
+  for (auto &iter : nodes_depth)
+  {
+    float cur_error = ErrorApproximator::approximate(iter.first);
+    if (iter.second < cut_depth)
+    {
+      error_at_input += cur_error;
+    }
+    else
+    {
+      error_at_output += cur_error;
+    }
+  }
+
+  if (error_at_input > error_at_output)
+  {
+    std::cerr << "Q16 will be set at input due to ";
+  }
+  else
+  {
+    std::cerr << "Q8 will be set at input due to ";
+  }
+  std::cerr << error_at_input << " error at input vs ";
+  std::cerr << error_at_output << " error at output." << std::endl;
+
+  return error_at_input > error_at_output;
+}
+
+class BisectionOptionsImpl final : public BisectionSolver::Options
+{
+public:
+  virtual void enable(Q16AtInput) final;
+  virtual bool query(Q16AtInput) final;
+
+private:
+  Q16AtInput _q16AtInput = Q16AtInput::Auto;
+};
+
+void BisectionOptionsImpl::enable(Q16AtInput q16AtInput) { _q16AtInput = q16AtInput; }
+bool BisectionOptionsImpl::query(Q16AtInput q16AtInput) { return _q16AtInput == q16AtInput; }
+
+} // namespace
+
+BisectionSolver::BisectionSolver(const std::string &input_data_path, float qerror_ratio)
+  : _input_data_path(input_data_path), _qerror_ratio(qerror_ratio), _qerror(0.f)
+{
+}
+
+BisectionSolver::Options *BisectionSolver::options(void)
+{
+  if (_options == nullptr)
+  {
+    _options = std::make_unique<BisectionOptionsImpl>();
+  }
+
+  return _options.get();
+}
+
+std::unique_ptr<luci::Module> BisectionSolver::run(const luci::Module *in_module)
+{
+  float min_depth = 0.f;
+  float max_depth = 0.f;
+  NodeDepthType nodes_depth;
+  if (compute_depth(in_module, nodes_depth, min_depth, max_depth) != EXIT_SUCCESS)
+  {
+    std::cerr << "Invalid graph for bisectioning" << std::endl;
+    return nullptr;
+  }
+
+  DatasetEvaluator evaluator(in_module, _input_data_path);
+  // const auto &ref_output = compute_outputs(module.get(), _input_data_path);
+  LayerParams layer_params;
+  float int16_qerror = evaluator.evaluate("int16", layer_params);
+  std::cerr << "int16_quantization_error " << int16_qerror << std::endl;
+
+  float int8_qerror = evaluator.evaluate("uint8", layer_params);
+  std::cerr << "int8_quantization_error " << int8_qerror << std::endl;
+
+  _qerror = int16_qerror + _qerror_ratio * std::fabs(int8_qerror - int16_qerror);
+  std::cerr << "target quantization error " << _qerror << std::endl;
+
+  int last_depth = -1;
+  float best_depth = -1;
+  LayerParams best_params;
+  auto graph = in_module->graph(0);
+  auto active_nodes = loco::active_nodes(loco::output_nodes(graph));
+  // input and output nodes are not valid for quantization, so let's remove them
+  for (auto node : loco::input_nodes(graph))
+  {
+    active_nodes.erase(node);
+  }
+  for (auto node : loco::output_nodes(graph))
+  {
+    active_nodes.erase(node);
+  }
+
+  // let's decide whether nodes at input are more suspectible to be quantized into Q16, than at
+  // output
+  bool int16_at_input = true;
+  if (auto option = options())
+  {
+    if (option->query(Options::Q16AtInput::Auto))
+    {
+      int16_at_input =
+        error_at_input_is_larger_than_at_output(nodes_depth, 0.5f * (max_depth + min_depth));
+    }
+    else if (option->query(Options::Q16AtInput::True))
+    {
+      int16_at_input = true;
+    }
+    else if (option->query(Options::Q16AtInput::False))
+    {
+      int16_at_input = false;
+    }
+  }
+
+  while (true)
+  {
+    float cut_depth = 0.5f * (min_depth + max_depth);
+
+    if (last_depth == int(cut_depth))
+    {
+      break;
+    }
+    last_depth = int(cut_depth);
+
+    auto nodes = graph->nodes();
+    LayerParams layer_params;
+    for (auto &node : active_nodes)
+    {
+      auto cur_node = loco::must_cast<luci::CircleNode *>(node);
+      auto iter = nodes_depth.find(cur_node);
+      if (iter == nodes_depth.end())
+      {
+        continue;
+      }
+
+      float depth = iter->second;
+      if ((depth <= cut_depth && int16_at_input) || (depth >= cut_depth && !int16_at_input))
+      {
+        auto layer_param = std::make_shared<LayerParam>();
+        {
+          layer_param->name = cur_node->name();
+          layer_param->dtype = "int16";
+          layer_param->granularity = "channel";
+        }
+
+        layer_params.emplace_back(layer_param);
+      }
+    }
+
+    float cur_accuracy = evaluator.evaluate("uint8", layer_params);
+    std::cerr << cut_depth << " : " << cur_accuracy << std::endl;
+
+    if (cur_accuracy < _qerror)
+    {
+      int16_at_input ? (max_depth = cut_depth) : (min_depth = cut_depth);
+      best_params = layer_params;
+      best_depth = cut_depth;
+    }
+    else
+    {
+      int16_at_input ? (min_depth = cut_depth) : (max_depth = cut_depth);
+    }
+  }
+
+  if (best_params.empty())
+  {
+    std::cerr << "Failed to optimal configuration" << std::endl;
+    return nullptr;
+  }
+
+  std::cerr << "Found optimal configuration at " << best_depth << " depth." << std::endl;
+  return evaluator.quantize("uint8", best_params);
+}
--- a/compiler/circle-mpqsolver/src/BisectionSolver.h
+++ b/compiler/circle-mpqsolver/src/BisectionSolver.h
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __MPQSOLVER_BISECTION_SOLVER_H__
+#define __MPQSOLVER_BISECTION_SOLVER_H__
+
+#include <luci/IR/Module.h>
+
+#include <memory>
+#include <string>
+
+namespace mpqsolver
+{
+
+class BisectionSolver final
+{
+
+public:
+  struct Options
+  {
+    enum class Q16AtInput
+    {
+      Auto,
+      True,
+      False,
+    };
+    virtual ~Options() = default;
+
+    virtual void enable(Q16AtInput) = 0;
+    virtual bool query(Q16AtInput) = 0;
+  };
+
+public:
+  Options *options(void);
+
+public:
+  BisectionSolver(const std::string &input_data_path, float qerror_ratio);
+  ~BisectionSolver() = default;
+
+  std::unique_ptr<luci::Module> run(const luci::Module *in_module);
+
+private:
+  std::string _input_data_path;
+  float _qerror = 0.f;       // quantization error
+  float _qerror_ratio = 0.f; // quantization error ratio
+  std::unique_ptr<Options> _options;
+};
+
+} // namespace mpqsolver
+
+#endif //__MPQSOLVER_BISECTION_SOLVER_H__
--- a/compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
+++ b/compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
@@ -1,0 +1,149 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "BisectionSolver.h"
+
+#include <arser/arser.h>
+#include <vconone/vconone.h>
+
+#include <luci/ImporterEx.h>
+#include <luci/Importer.h>
+#include <luci/CircleExporter.h>
+#include <luci/CircleFileExpContract.h>
+
+#include <iostream>
+#include <iomanip>
+#include <chrono>
+
+void print_version(void)
+{
+  std::cout << "circle-mpqsolver version " << vconone::get_string() << std::endl;
+  std::cout << vconone::get_copyright() << std::endl;
+}
+
+int entry(int argc, char **argv)
+{
+  const std::string bisection_str = "--bisection";
+
+  arser::Arser arser("circle-mpqsolver provides circle_model mixed precision quantization");
+
+  arser::Helper::add_version(arser, print_version);
+  arser::Helper::add_verbose(arser);
+
+  arser.add_argument("--data").required(true).help(".h5 file with test data");
+  arser.add_argument("--qerror_ratio")
+    .type(arser::DataType::FLOAT)
+    .default_value(0.5f)
+    .help("quantization error ratio");
+
+  arser.add_argument(bisection_str)
+    .nargs(1)
+    .default_value("auto")
+    .type(arser::DataType::STR)
+    .help("Bisection method Q16 at input nodes"
+          "Default is 'auto'."
+          "Single optional argument: whether Q16 qunatization should be at input nodes "
+          "'auto', 'true', 'false'.");
+
+  arser.add_argument("--input_model")
+    .required(true)
+    .help("Input float model with min max initialized");
+
+  arser.add_argument("--output_model").required(true).help("Output quantized model");
+
+  try
+  {
+    arser.parse(argc, argv);
+  }
+  catch (const std::runtime_error &err)
+  {
+    std::cerr << err.what() << std::endl;
+    std::cout << arser;
+    return EXIT_FAILURE;
+  }
+
+  if (arser.get<bool>("--verbose"))
+  {
+    // The third parameter of setenv means REPLACE.
+    // If REPLACE is zero, it does not overwrite an existing value.
+    setenv("LUCI_LOG", "100", 0);
+  }
+
+  auto data_path = arser.get<std::string>("--data");
+  auto input_model_path = arser.get<std::string>("--input_model");
+  auto output_model_path = arser.get<std::string>("--output_model");
+
+  float qerror_ratio = arser.get<float>("--qerror_ratio");
+
+  auto start = std::chrono::high_resolution_clock::now();
+
+  // Load input model
+  luci::ImporterEx importerex;
+  auto module = importerex.importVerifyModule(input_model_path);
+  if (module.get() == nullptr)
+  {
+    std::cerr << "Failed to load " << input_model_path << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  // optimize
+  mpqsolver::BisectionSolver solver(data_path, qerror_ratio);
+  {
+    auto value = arser.get<std::string>(bisection_str);
+    if (value == "auto")
+    {
+      solver.options()->enable(mpqsolver::BisectionSolver::Options::Q16AtInput::Auto);
+    }
+    else if (value == "true")
+    {
+      solver.options()->enable(mpqsolver::BisectionSolver::Options::Q16AtInput::True);
+    }
+    else if (value == "false")
+    {
+      solver.options()->enable(mpqsolver::BisectionSolver::Options::Q16AtInput::False);
+    }
+    else
+    {
+      std::cerr << "Unrecognized option for bisection algortithm" << input_model_path << std::endl;
+      return EXIT_FAILURE;
+    }
+  }
+
+  auto optimized = solver.run(module.get());
+  if (optimized == nullptr)
+  {
+    std::cerr << "Failed to build mixed precision model" << input_model_path << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  // save optimized
+  {
+    luci::CircleExporter exporter;
+    luci::CircleFileExpContract contract(optimized.get(), output_model_path);
+    if (!exporter.invoke(&contract))
+    {
+      std::cerr << "Failed to build mixed precision model" << input_model_path << std::endl;
+      return EXIT_FAILURE;
+    }
+  }
+
+  auto duration = std::chrono::duration_cast<std::chrono::seconds>(
+    std::chrono::high_resolution_clock::now() - start);
+  std::cerr << "Time elapsed is " << std::setprecision(5) << duration.count() / 60.f << " minutes."
+            << std::endl;
+
+  return EXIT_SUCCESS;
+}
--- a/compiler/circle-mpqsolver/src/ErrorApproximator.cpp
+++ b/compiler/circle-mpqsolver/src/ErrorApproximator.cpp
@@ -1,0 +1,352 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ErrorApproximator.h"
+
+#include <cmath>
+#include <limits>
+#include <vector>
+#include <functional>
+#include <luci/IR/Nodes/CircleConst.h>
+#include <luci/IR/Nodes/CircleConv2D.h>
+#include <luci/IR/Nodes/CircleTransposeConv.h>
+#include <luci/IR/Nodes/CircleFullyConnected.h>
+#include <luci/IR/Nodes/CircleDepthwiseConv2D.h>
+
+using namespace mpqsolver;
+
+namespace
+{
+
+using namespace luci;
+using IterFunc = std::function<void(uint32_t *, loco::TensorShape &, int32_t)>;
+
+inline bool has_min_max(const CircleNode *node)
+{
+  return node->quantparam() && !node->quantparam()->min.empty() && !node->quantparam()->max.empty();
+}
+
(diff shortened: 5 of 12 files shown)


compiler/circle-mpqsolver/src/CircleMPSolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+  arser::Arser arser("circle-mpsolver provides circle_model mixed precision quantization");
+
+  arser::Helper::add_version(arser, print_version);
+  arser::Helper::add_verbose(arser);
+
+  arser.add_argument("--data").required(true).help(".h5 file with test data");
+  arser.add_argument("--accuracy_ratio")
+    .type(arser::DataType::FLOAT)
+    .default_value<float>(0.5f)
+    .help("accuracy ratio");
```
jinevening, line 43: This seems an intuitive interface to receive metric goals. But accuracy is a mis-leading term.. We may need to find a better name.

compiler/circle-mpqsolver/src/utils/ErrorApproximator.cpp
```diff
@@ -0,0 +1,334 @@
...
+    int channel_idx = indices[channel_dim_index];
+    auto data = node->at<loco::DataType::FLOAT32>(cal_offset(dimension, indices));
+    values[channel_idx].push_back(data);
+  };
+
+  iterate_per_channel(node, channel_dim_index, cal_minmax);
+}
+} // namespace
+
+float ErrorApproximator::Approximate(CircleNode *node)
```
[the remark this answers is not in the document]
stamalakhov: > Can you explain what this function does?

It tries to estimate maximal layer error introduced by quantization, to recognize where the most sensitive to quantization layers are (closer to input nodes or to output nodes).

>It would be better to add comments.

Ok.

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+  //  arser.add_argument("--method")
+  //    .type(arser::DataType::STR_VEC)
+  //    .help("Currently only bisection is supported");
```
jinevening, line 47: You may receive arguments by command line, e.g.,

```
circle-mpqsolver --data <data> --bisection <option_1> ... <option_n> --input_model <model> --output_model <model>
```

Or you may receive configuration from json file (perhaps compatible with quantization configuration file). This may be  useful if circle-mpqsolver allows user's tuning, e.g., enforcing q16 to specific layers.

jinevening:
Overall, draft looks ok to me. You may need some unit tests when you upload non-draft PRs. Also, some integration tests (e.g., running bisection solver in the circle model) are necessary as well.

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+  arser.add_argument("--input_model")
+    .required(true)
+    .help("Input float model with min max initialized");
+
+  arser.add_argument("--output_model").required(true).help("Output quantized model");
+
```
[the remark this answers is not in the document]
stamalakhov: Ok

compiler/circle-mpqsolver/src/utils/ErrorApproximator.cpp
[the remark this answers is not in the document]
jinevening: > layer error

Detailed equation to compute the layer error would be helpful. For example, mse or mae, etc.

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
stamalakhov: Ok.

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
jinevening: Ah, I mean the term `accuracy` is mis-leading. `accuracy` is typically used to measure task loss, which `mpqsolver` does not know.  It may be replaced with `quantization_error`, `qerror` or something else.

stamalakhov: Ahh... Understood. `quantization_error` seems to be quite reasonable.

compiler/circle-mpqsolver/src/Methods/BisectionSolver.h
```diff
@@ -0,0 +1,46 @@
...
+#define __MPQSOLVER_BISECTION_SOLVER_H__
+
+#include <luci/IR/Module.h>
+
+#include <memory>
+#include <string>
+
+namespace mpqsolver
+{
+class BisectionSolver final
```
mhs4670go, line 27: (trival) one empty space is usually added.
```suggestion
{

class BisectionSolver final
```

compiler/circle-mpqsolver/src/Methods/BisectionSolver.h
```diff
@@ -0,0 +1,46 @@
...
+{
+public:
+  BisectionSolver(const std::string &input_data_path, float accuracy_ratio);
+  ~BisectionSolver() = default;
+
+  int run(const std::string &input_model);
+  void save(const std::string &output_model_path);
+
+private:
```
mhs4670go, line 36:
```suggestion
```

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
```diff
@@ -0,0 +1,94 @@
+/*
```
mhs4670go, line 1: Maybe the file name should be `CircleMPQSolver.cpp`?

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+void print_version(void)
+{
+  std::cout << "circle-quantizer version " << vconone::get_string() << std::endl;
+  std::cout << vconone::get_copyright() << std::endl;
+}
+
+int entry(int argc, char **argv)
+{
+  arser::Arser arser("circle-mpsolver provides circle_model mixed precision quantization");
```
mhs4670go, line 34:
```suggestion
  arser::Arser arser("circle-mpqsolver provides circle_model mixed precision quantization");
```

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+  arser::Helper::add_version(arser, print_version);
+  arser::Helper::add_verbose(arser);
+
+  arser.add_argument("--data").required(true).help(".h5 file with test data");
+  arser.add_argument("--accuracy_ratio")
+    .type(arser::DataType::FLOAT)
+    .default_value<float>(0.5f)
```
mhs4670go, line 42: FYI, it would work without `<float>`.
```suggestion
    .default_value(0.5f)
```

compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp
```diff
@@ -0,0 +1,291 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "BisectionSolver.h"
+#include "../utils/ErrorApproximator.h"
+#include "../utils/ModuleCloner.h"
+#include "../utils/Evaluator.h"
```
mhs4670go, line 20: It is not recommended to use `..` when including headers for maintainability. How about add below to `CMakeLists.txt`?
```cmake
target_include_directories(circle-mpqsolver PRIVATE src)
```
```suggestion
#include "utils/ErrorApproximator.h"
#include "utils/ModuleCloner.h"
#include "utils/Evaluator.h"
```

compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp
```diff
@@ -0,0 +1,291 @@
...
+{
+}
+
+int BisectionSolver::run(const std::string &input_model)
+{
+  // Load model from the file
+  luci::ImporterEx importerex;
+  auto module = importerex.importVerifyModule(input_model);
+  if (module.get() == nullptr)
+    return EXIT_FAILURE;
```
mhs4670go, line 167: (optional) It seems better to move the code that loads the model outside of this function for SRP.

compiler/circle-mpqsolver/README.md
```diff
@@ -0,0 +1,3 @@
+# circle-mpqsolver
+
+_circle-mpqsolver_ provides light-weight methods for finding mixed precision quantization model.
```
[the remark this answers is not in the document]
mhs4670go: (optional) Also, it would be better to add the explanation about what Bisection method means and what it does.

compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp
```diff
@@ -0,0 +1,291 @@
...
+      auto inp_name = inp_node->name();
+      input_names.insert(inp_name);
+    }
+  }
+
+  std::set<luci::CircleNode *> inputs;
+  std::map<luci::CircleNode *, std::vector<luci::CircleNode *>> inputs_to_outputs;
+  for (uint32_t node_index = 0; node_index < nodes_size; ++node_index)
+  {
+    auto node = loco::must_cast<luci::CircleNode *>(nodes->at(node_index));
```
mhs4670go, line 65: You can use `loco` library.
```suggestion
  for (auto node : loco::all_nodes(graph))
  {
    auto cnode = loco::must_cast<const luci::CircleNode *>(node);
```
And, `loco::input_nodes` could be used as well.

compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp
```diff
@@ -0,0 +1,291 @@
...
+    for (uint32_t inp_index = 0; inp_index < arity; ++inp_index)
+    {
+      auto arg = loco::must_cast<luci::CircleNode *>(node->arg(inp_index));
+      inputs_to_outputs[arg].push_back(node);
+    }
+  }
+
+  // initializing
+  std::vector<luci::CircleNode *> to_process{inputs.begin(), inputs.end()};
+  std::map<std::string, float> named_depth;
```
mhs4670go, line 81: What is this `named_depth` for?

compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp
```diff
@@ -0,0 +1,291 @@
...
+  // Export to output Circle file
+  luci::CircleExporter exporter;
+
+  luci::CircleFileExpContract contract(_module.get(), output_model_path);
+
+  if (!exporter.invoke(&contract))
+  {
+    throw std::runtime_error("Failed to export '" + output_model_path + "'");
+  }
+}
```
mhs4670go, line 291: This function shouldn't be here for SRP as well.

compiler/circle-mpqsolver/src/utils/ModuleCloner.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+  std::unique_ptr<luci::Module> new_module = luci::make_module();
+  for (size_t idx = 0; idx < module->size(); ++idx)
+  {
+    auto graph = module->graph(idx);
+    auto new_graph = clone_graph(graph);
+    if (!graph)
+    {
+      return nullptr;
+    }
```
mhs4670go, line 167: Checking first would reduce the costs.
```suggestion
    if (!graph)
    {
      return nullptr;
    }
    auto new_graph = clone_graph(graph);
    
```

mhs4670go: Well, if `clone_graph` returns `nullptr`, both way have almost same costs.

compiler/circle-mpqsolver/src/utils/ModuleCloner.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <luci/IR/Module.h>
+
+namespace mpqsolver
+{
+
+class ModuleCloner final
+{
+public:
+  static std::unique_ptr<luci::Module> Clone(const luci::Module *module);
```
mhs4670go, line 27: Method names start with lowercase.
```suggestion
  static std::unique_ptr<luci::Module> clone(const luci::Module *module);
```

compiler/circle-mpqsolver/src/utils/Evaluator.h
```diff
@@ -0,0 +1,52 @@
...
+  DatasetEvaluator(luci::Module *ref_module, std::string &h5file);
+  ~DatasetEvaluator() = default;
+  float Evaluate(const std::string &def_quant, LayerParams &layer_params);
+  std::unique_ptr<luci::Module> Quantize(const std::string &def_quant, LayerParams &layer_params);
+
+private:
+  float evaluate(const luci::Module *module);
+
+private:
+  luci::Module *_ref_module;
```
mhs4670go, line 45: `const` keyword seems safer.
```suggestion
  const luci::Module *_ref_module;
```

compiler/circle-mpqsolver/src/utils/ModuleCloner.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    add_graph_input(graph_clone, input_clone);
+    clonectx.emplace(input_org, input_clone);
+  }
+
+  // clone nodes
+  auto nodes = graph_org->nodes();
+  for (uint32_t n = 0; n < nodes->size(); ++n)
+  {
+    auto node = nodes->at(n);
```
mhs4670go, line 104:
```suggestion
  auto nodes = graph_org->nodes();
  for (const auto &node : loco::all_nodes(graph_org))
  {
    auto cnode = loco::must_cast<const luci::CircleNode *>(node);
```

mhs4670go: Or you could use set operations to remove `if` conditions.
```c++
auto outputs = loco::output_nodes(graph_org);
auto in_out = loco::input_nodes(graph_org);
in_out.insert(outputs.begin(), outputs.end());

auto nodes = loco::all_nodes(graph_org);
std::set nodes_without_in_out;
std::set_difference(nodes.begin(), nodes.end(), 
                   in_out.begin(), in_out.end(), 
           std::inserter(nodes_without_in_out, nodes_without_in_out.begin()));
```

mhs4670go:
FYI, when you add a unittest, the number of the positive test and the negative test should be similar.

compiler/circle-mpqsolver/src/utils/ErrorApproximator.cpp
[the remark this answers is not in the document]
jinevening: Well, It seems that the error is aggregate of each layer's upper bound of L1 quantization error. Am I right?

In the non-draft code, the equation will be included in the comments.

compiler/circle-mpqsolver/src/utils/ErrorApproximator.cpp
```diff
@@ -0,0 +1,334 @@
...
+{
+  auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+  auto name = circle_node->name();
+  CircleConst *weights = nullptr;
+  uint32_t volume = 0;
+  std::vector<uint32_t> cir_shape;
+  if (getShape(circle_node, cir_shape) && cir_shape.size() == 4)
```
jinevening, line 241: It would be better to find a way to compute errors of non-rank 4 layers.

compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp
```diff
@@ -0,0 +1,291 @@
+/*
```
seanshpark, line 1: about the folder name; `Methods` starts with capital, but `utils` doesn't.
it would be better to not mix them; maybe be use lower case.

compiler/circle-mpqsolver/src/utils/ErrorApproximator.cpp
```diff
@@ -0,0 +1,334 @@
+/*
```
seanshpark, line 1: Q) do we need to put these files in `utils` folder? why not just put them in `src` folder?

compiler/circle-mpqsolver/src/utils/ErrorApproximator.cpp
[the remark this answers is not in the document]
stamalakhov: > Well, It seems that the error is aggregate of each layer's upper bound of L1 quantization error. Am I right?

Yes. You are.

>In the non-draft code, the equation will be included in the comments.

Ok.

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
stamalakhov: > Maybe the file name should be `CircleMPQSolver.cpp`

Ahhh... Ok.

compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp
stamalakhov: It's for debug. May be it should be removed in the nonDRAFT PR.

compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp
stamalakhov: > about the folder name; `Methods` starts with capital, but `utils` doesn't. it would be better to not mix them; maybe be use lower case.

Ahhh... Sorry. Will be fixed.

compiler/circle-mpqsolver/src/utils/ErrorApproximator.cpp
stamalakhov: > Q) do we need to put these files in `utils` folder? why not just put them in `src` folder?

Ok. Seems like 'methods' folder should also be removed.

compiler/circle-mpqsolver/README.md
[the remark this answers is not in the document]
stamalakhov: > (optional) Also, it would be better to add the explanation about what Bisection method means and what it does.

Ok.

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
stamalakhov: Thank you.

compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp
stamalakhov: Ok.

compiler/circle-mpqsolver/src/Methods/BisectionSolver.h
```diff
@@ -0,0 +1,46 @@
+/*
```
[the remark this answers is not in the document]
stamalakhov: I'm going to remove 'utils' and 'methods' folder

compiler/circle-mpqsolver/src/utils/Evaluator.h
seanshpark: better with initial value
```
const luci::Module *_ref_module = nullptr;
```

stamalakhov: > better with initial value
> 
> ```
> const luci::Module *_ref_module = nullptr;
> ```

Ok.

mhs4670go:
```bash
15:08:36 /opt/jenkins_agent/workspace/nnfw/master/pr-nncc-debug/compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp:18:10: fatal error: utils/ErrorApproximator.h: No such file or directory
15:08:36  #include "utils/ErrorApproximator.h"
15:08:36           ^~~~~~~~~~~~~~~~~~~~~~~~~~~
15:08:36 compilation terminated.
```

stamalakhov:
> ```shell
> 15:08:36 /opt/jenkins_agent/workspace/nnfw/master/pr-nncc-debug/compiler/circle-mpqsolver/src/Methods/BisectionSolver.cpp:18:10: fatal error: utils/ErrorApproximator.h: No such file or directory
> 15:08:36  #include "utils/ErrorApproximator.h"
> 15:08:36           ^~~~~~~~~~~~~~~~~~~~~~~~~~~
> 15:08:36 compilation terminated.
> ```

Thank you. Currently i'm fixing this and other remarks. New version will be pushed as soon as possible.

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
stamalakhov: `quantization_error` is rather lengthy :( , let's try `qerror`  instead.

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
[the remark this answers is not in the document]
stamalakhov: Ahhh... Seems like we can add some tracing later.

compiler/circle-mpqsolver/src/CircleMPSolver.cpp
stamalakhov: For now introduced int16_at_input option for --bisection, so we can manually set where most sensitive to quantization layers are. But json option seems like quite usable, i need some more information, what values it should include.

compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
```diff
@@ -0,0 +1,149 @@
...
+    .default_value(0.5f)
+    .help("quantization error ratio");
+
+  arser.add_argument(bisection_str)
+    .nargs(1)
+    .type(arser::DataType::STR_VEC)
+    .help("Bisection method Q16 at input nodes"
+          "Default is 'auto'."
+          "Single optional argument: whether Q16 should be at input nodes "
+          "'auto', 'input', 'output'.");
```
mhs4670go, line 58: 1. Seems to me that below codes check if the number of arguments is 1. But its type is set to `STR_VEC`. Maybe `STR`?
2. it says `Default is auto`. Then, `.default_value('auto')` might be needed.
3. Actually, it is very hard to know what values can be given -  'auto', 'true', 'false'.

compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
```diff
@@ -0,0 +1,149 @@
...
+  arser::Arser arser("circle-mpqsolver provides circle_model mixed precision quantization");
+
+  arser::Helper::add_version(arser, print_version);
+  arser::Helper::add_verbose(arser);
+
+  arser.add_argument("--data").required(true).help(".h5 file with test data");
+  arser.add_argument("--qerror_ratio")
+    .type(arser::DataType::FLOAT)
+    .default_value(0.5f)
+    .help("quantization error ratio");
```
mhs4670go, line 50: This help message make me think "What is quantization error ratio?". IIUC, the ratio is (quant error of full-Q16)/(quant error of full-Q8 model). Right?

compiler/circle-mpqsolver/src/BisectionSolver.h
```diff
@@ -0,0 +1,64 @@
...
+
+public:
+  struct Options
+  {
+    enum class Q16AtInput
+    {
+      Q16AtInput_Auto,
+      Q16AtInput_True,
+      Q16AtInput_False
+    };
```
mhs4670go, line 39: Prefixes are duplicates. How about this one?
```suggestion
    enum class Q16AtInput
    {
      Auto,
      True,
      False,
    };
```

compiler/circle-mpqsolver/src/BisectionSolver.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+using NodeDepthType = std::map<luci::CircleNode *, float>;
+
+namespace
+{
+
+int compute_depth(const luci::Module *module, NodeDepthType &nodes_depth, float &min_depth,
+                  float &max_depth)
+{
+  if (module->size() != 1)
```
mhs4670go, line 35:
```suggestion
{
  if (module == nullptr)
    return EXIT_FAILURE;

  if (module->size() != 1)
```

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
```diff
@@ -0,0 +1,341 @@
...
+}
+} // namespace
+
+/**
+ * How Approximate works?
+ *
+ * Currently it works just for convolution layers, but may be generalized for other types as well.
+ * See discussion at https://github.com/Samsung/ONE/pull/10170#discussion_r1042246598
+ * Convolution is just a matrix multiplication.
+ * While quantizing we intoduce quantization error into convolution operand (activations) as well as
```
mhs4670go, line 240:
```suggestion
 * While quantizing we introduce quantization error into convolution operand (activations) as well as
```

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
```diff
@@ -0,0 +1,341 @@
...
+    weights = loco::must_cast<luci::CircleConst *>(convWeights->filter());
+  }
+  else if (auto convWeights = dynamic_cast<luci::CircleDepthwiseConv2D *>(circle_node))
+  {
+    weights = loco::must_cast<luci::CircleConst *>(convWeights->filter());
+  }
+  else if (auto convWeights = dynamic_cast<luci::CircleTransposeConv *>(circle_node))
+  {
+    weights = loco::must_cast<luci::CircleConst *>(convWeights->filter());
+  }
```
mhs4670go, line 272: Maybe?
```suggestion
  if (auto conv = dynamic_cast<luci::CircleConv2D *>(circle_node))
  {
    weights = loco::must_cast<luci::CircleConst *>(conv->filter());
  }
  else if (auto dconv = dynamic_cast<luci::CircleDepthwiseConv2D *>(circle_node))
  {
    weights = loco::must_cast<luci::CircleConst *>(dconv->filter());
  }
  else if (auto tconv = dynamic_cast<luci::CircleTransposeConv *>(circle_node))
  {
    weights = loco::must_cast<luci::CircleConst *>(tconv->filter());
  }
```

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+  if (weights != nullptr)
+  {
+    float ci_min = 0.f;
+    float ci_max = 0.f;
+
+    int n_arity = circle_node->arity();
+    for (int i = 0; i < n_arity; i++)
+    {
+      auto parent_node = loco::must_cast<luci::CircleNode *>(circle_node->arg(i));
```
mhs4670go, line 282: You can use `loco::preds` method.
```suggestion
    auto preds = loco::preds(n);
    for (const auto &pred : preds)
    {
      auto parent_node = loco::must_cast<const luci::CircleNode *>(pred);
```

compiler/circle-mpqsolver/src/BisectionSolver.h
stamalakhov: Thank you. Seems quite reasonable!

compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
stamalakhov: Yes. `_quantization_error = Q16_quantization_error + _qerror_ratio * (Q8_quantization_error - Q16_qunatization_error)`

compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
stamalakhov: Ahhh... Ok. It was hardcoded, but in the wrong place, I'll fix it. Thank you.

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
stamalakhov: Ahhh... Ok. Thank you!

compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
stamalakhov: > 1. Seems to me that below codes check if the number of arguments is 1. But its type is set to `STR_VEC`. Maybe `STR`?

There may be other options, like .json file with frozen layers. But for now single option is just enough. Let it be STR. All other options may be set at .json file later as soon as there is need in them.

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
```diff
@@ -0,0 +1,341 @@
...
+    {
+      shape[i] = circle_node->dim(i).value();
+    }
+    return true;
+  }
+
+  return false;
+}
+
+static void get_values_per_channel(CircleConst *node, std::vector<std::vector<float>> &values,
```
mhs4670go, line 209: This function is not used.

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
```diff
@@ -0,0 +1,318 @@
...
+ * so quantization error of matrix multiplication is A * q_err(W) + W * q_err(A).
+ * Estimating its largest value we get
+ * ~ number_of_additions * (A_max * (W_max - W_min) / 255 + W_max * (A_max - A_min) / 255)
+ * The following code tries to get total error for quantizing convolution node into Q8.
+ * It's just an heuristic (Metric sensitivity depends highly on derivatives as well).
+ */
+float ErrorApproximator::approximate(CircleNode *node)
+{
+  auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+  auto name = circle_node->name();
```
mhs4670go, line 229: `name` is not used, too.
```suggestion
```

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
```diff
@@ -0,0 +1,318 @@
...
+    // total weight quantization error across all channels
+    // so maximal error of quantization is ~ (max_value - min_value) / 255
+    // omitting 255 term we get that maximal error of quantization is just its range
+    float sum_err = 0.f;
+    for (auto cur_err : ranges)
+    {
+      sum_err += cur_err;
+    }
+
+    // maximal quantization error introduced by quantizimh weights
```
mhs4670go, line 305: Or "introduced by weight quantization"?
```suggestion
    // maximal quantization error introduced by quantizing weights
```

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
```diff
@@ -0,0 +1,318 @@
...
+ * W * q_err(A), q_err - quantization error, W - weight matrix, A - activations from previous layer,
+ * so quantization error of matrix multiplication is A * q_err(W) + W * q_err(A).
+ * Estimating its largest value we get
+ * ~ number_of_additions * (A_max * (W_max - W_min) / 255 + W_max * (A_max - A_min) / 255)
+ * The following code tries to get total error for quantizing convolution node into Q8.
+ * It's just an heuristic (Metric sensitivity depends highly on derivatives as well).
+ */
+float ErrorApproximator::approximate(CircleNode *node)
+{
+  auto circle_node = loco::must_cast<luci::CircleNode *>(node);
```
mhs4670go, line 228:
```suggestion
```

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
```diff
@@ -0,0 +1,318 @@
...
+        }
+      }
+    }
+
+    // channel-size min, max
+    std::vector<float> min_values;
+    std::vector<float> max_values;
+    {
+      int32_t channel_dim_index = 0;
+      cal_minmax_per_channel(weights, min_values, max_values, channel_dim_index);
```
mhs4670go, line 276: (optional)
1. Seems to me that `channel_dim_index` doesn't have to be a parameter of `cal_minmax_per_channel` method.
2. How about use convolution nodes having `opcode` instead of `weights`. Current codes have to check its origin(conv or dconv, etc) everytime.
3. `get_channel_dim_index` can be revisited. Simply getting only channel dim index.

With above consideration, codes would look like below.
```cxx
static uint32_t get_channel_dim_index(CircleNode *node)
{
  uint32_t index = 0;
  auto opcode = node->opcode();
  switch (opcode)
  {
    case luci::CircleOpcode::CONV_2D:
    case luci::CircleOpcode::TRANSPOSE_CONV:
    case luci::CircleOpcode::FULLY_CONNECTED:
    {
      index = 0;
    }
    case luci::CircleOpcode::DEPTHWISE_CONV_2D:
    {
      index = 3;
    }
    default:
      throw std::runtime_error("Failed to find channel index in " + node->name());
  }
  
  return index;
}

static bool set_weight_dim(CircleNode *node, loco::TensorShape dimension)
{
  auto opcode = node->opcode();
  switch (opcode)
  {
    case luci::CircleOpcode::CONV_2D:
    case luci::CircleOpcode::TRANSPOSE_CONV:
    case luci::CircleOpcode::DEPTHWISE_CONV_2D:
    {
      assert(node->rank() == 4);
      dimension.rank(node->rank());
      dimension.dim(0).set(node->dim(0).value());
      dimension.dim(1).set(node->dim(1).value());
      dimension.dim(2).set(node->dim(2).value());
      dimension.dim(3).set(node->dim(3).value());
    }
    case luci::CircleOpcode::FULLY_CONNECTED:
    {
      assert(node->rank() == 2);
      dimension.rank(4);
      dimension.dim(0).set(node->dim(0).value());
      dimension.dim(1).set(1); // Set FC layer like CONV
      dimension.dim(2).set(1);
      dimension.dim(3).set(node->dim(1).value());
    }
    default:
      return false;
  }

  return true;
}

static loco::Node *get_weight(CircleNode *node)
{
  loco::Node *weight = nullptr;
  auto opcode = node->opcode();
  switch (opcode)
  {
    case luci::CircleOpcode::CONV_2D:
    {
      auto conv = loco::must_cast<CircleConv2D *>(node);
      weight = conv->filter();
    }
    case luci::CircleOpcode::DEPTHWISE_CONV_2D:
    {
      auto dconv = loco::must_cast<CircleDepthwiseConv2D *>(node);
      weight = dconv->filter();
    }
    case luci::CircleOpcode::TRANSPOSE_CONV:
    {
      auto tconv = loco::must_cast<CircleTransposeConv *>(node);
      weight = tconv->filter();
    }
    case luci::CircleOpcode::FULLY_CONNECTED:
    {
      auto fc = loco::must_cast<CircleFullyConnected *>(node);
      weight = fc->weights();
    }
    default:
    {
      break;
    }
  }
  return weight;
}

static void iterate_per_channel(CircleNode *node, IterFunc func)
{
  loco::TensorShape dimension;
  set_weight_dim(node, dimension);
  uint32_t indices[4] = {
    0,
  };

  auto channel_dim_index = get_channel_dim_index(node);

  for (indices[0] = 0; indices[0] < dimension.dim(0).value(); indices[0]++)
  {
    for (indices[1] = 0; indices[1] < dimension.dim(1).value(); indices[1]++)
    {
      for (indices[2] = 0; indices[2] < dimension.dim(2).value(); indices[2]++)
      {
        for (indices[3] = 0; indices[3] < dimension.dim(3).value(); indices[3]++)
        {
          func(indices, dimension, channel_dim_index);
        }
      }
    }
  }
}

static void cal_minmax_per_channel(CircleNode *node, std::vector<float> &min,
                                   std::vector<float> &max)
{
  loco::TensorShape dimension;
  set_weight_dim(node, dimension);

  auto channel_dim_index = get_channel_dim_index(node);
  auto size = dimension.dim(channel_dim_index).value();

  std::vector<bool> has_min_max_value(size, false);
  min.resize(size);
  max.resize(size);

  CircleConst *weight = loco::must_cast<CircleConst *>(get_weight(node));
  auto cal_minmax = [&](uint32_t *indices, loco::TensorShape &dimension, int channel_dim_index) {
    int channel_idx = indices[channel_dim_index];
    auto data = weight->at<loco::DataType::FLOAT32>(cal_offset(dimension, indices));
    if (has_min_max_value[channel_idx])
    {
      min[channel_idx] = data < min[channel_idx] ? data : min[channel_idx];
      max[channel_idx] = data > max[channel_idx] ? data : max[channel_idx];
    }
    else
    {
      min[channel_idx] = data;
      max[channel_idx] = data;
      has_min_max_value[channel_idx] = true;
    }
  };

  iterate_per_channel(node, cal_minmax);
}
```

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
stamalakhov: IMHO circle_node is heavily used below. It is operand for getShape and castings. Without `circle_node` the file won't be compiled.

stamalakhov: Ahhh... You're right. One of them should be removed.

stamalakhov: Fixed.

compiler/circle-mpqsolver/src/ErrorApproximator.cpp
stamalakhov: I've reworked the whole `ErrorEpproximator.cpp` module according to @mhs4670go suggestions. I've discovered some bugs in computations. Seems like they are fixed for now. Still it correctly predicts that input nodes of `mobilenet_v1` should be quantized to Q16, and for `cartoon` model predicts the opposite (Q16 should be at output).

mhs4670go:
@stamalakhov 

Looks good to me. You had a really hard time!

I think it's time for you to post a PR:) And, when you post a PR, below are things that I think you can consider.

1. Adding unit test. As I said earlier, you need to take the ratio of the positive test and the negative test equally. Negative test name ends with `_NEG` suffix like below.
https://github.com/Samsung/ONE/blob/cc66a2776171ac6f44422c6903ee5545b373091c/compiler/circle-eval-diff/src/InputDataLoader.test.cpp#L45

2. (optiaonl) Adding explanation about bisection method. Like what is this, how it works, etc. Actually, even though I've reviewed the codes, I still don't fully get it..

3. (optional) Adding integration test. Well, actually, I think we can't do the value test because there is no label. But we can test if the artifacts are generated as expected. Or, if you have a good idea, share your ideas:)

stamalakhov:
> @stamalakhov
> 
> Looks good to me. You had a really hard time!
> 
> I think it's time for you to post a PR:) And, when you post a PR, below are things that I think you can consider.
> 
>     1. Adding unit test. As I said earlier, you need to take the ratio of the positive test and the negative test equally. Negative test name ends with `_NEG` suffix like below.
>        https://github.com/Samsung/ONE/blob/cc66a2776171ac6f44422c6903ee5545b373091c/compiler/circle-eval-diff/src/InputDataLoader.test.cpp#L45
> 
>     2. (optiaonl) Adding explanation about bisection method. Like what is this, how it works, etc. Actually, even though I've reviewed the codes, I still don't fully get it..
> 
>     3. (optional) Adding integration test. Well, actually, I think we can't do the value test because there is no label. But we can test if the artifacts are generated as expected. Or, if you have a good idea, share your ideas:)

Ok.

compiler/circle-mpqsolver/src/utils/ErrorApproximator.cpp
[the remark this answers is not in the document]
stamalakhov: Let's be more precise for estimation of convolutions. Please correct me, if i am wrong.
image
[convolution.pdf](https://github.com/Samsung/ONE/files/10379502/convolution.pdf)

Issue #3963, opened 2020-08-25
Compiler FE: how to handle the type of CircleOutputExclude


mhs4670go:
This issue is about handling the type of `CircleOutputExclude` IR.

The current implementation is as follows.

```cxx
$ cat compiler/luci/import/src/GraphBuilder.cpp
...
std::vector<CircleNode *> input_nodes;
  for (const int32_t input_tensor_index : inputs)
  {
    if (input_tensor_index >= 0)
    {
      auto input = context->nodefinder()->node(input_tensor_index);
      if (input == nullptr)
        INFO(l) << "[luci] Warning: input node is null " << input_tensor_index << std::endl;
      input_nodes.push_back(input);
    }
    else
    {
      // If there is no tensor, insert CircleOutputExclude.
      input_nodes.push_back(context->graph()->nodes()->create<luci::CircleOutputExclude>()); // NOTE HERE
    }
  }
...
```

`CircleOutputExclude` doesn't need a type, but since all nodes must have a type, a dummy type is inserted.

Because of this, when we build a node that has tensor whose tensor index is -1, we've got to set dummy dtype.

```cxx
$ cat compiler/luci/import/src/Nodes/CircleFullyConnected.cpp
...
// TODO Find and move to appropriate place for setting optional input
  if (auto bias = dynamic_cast<luci::CircleOutputExclude *>(node->bias()))
  {
    // bias is not used for type inference, but node itself should have a type
    bias->dtype(loco::DataType::FLOAT32);

    // bias is not used for shape inference
  }
...
```

I don't think it's efficient. It would be better to set dtype as soon as we create `CircleOutputExclude` node.
```cxx
auto node = context->graph()->nodes()->create<luci::CircleOutputExclude>();
node->dtype(loco::DataType::FLOAT32);
input_nodes.push_back(node);
```

@llFreetimell Please give me some your opinion.

llFreetimell:
At first, I had had two ideas.

- Set `dtype` when `CircleOutputExclude` is created (as you mentioned)
- Introducing `loco::DataType::NONE`

However, both of them was rejected at that time :(
I am glad for your opinion but other people's opinion seems important I think :)

seanshpark:
> However, both of them was rejected at that time :(

Do you remember the reason it was reject?

llFreetimell:
> Set dtype when CircleOutputExclude is created

[...]

PR #10173, opened 2022-12-08
[luci/pass] Revise bias for ReplaceNonConstFCWithBatchMatMulPass

seanshpark:
This will revise ReplaceNonConstFCWithBatchMatMulPass for bias
- bias can be empty with CircleOutputExclude type
- only check dtype when type is CircleConst

### DIFF
--- a/compiler/luci/pass/src/ReplaceNonConstFCWithBatchMatMulPass.cpp
+++ b/compiler/luci/pass/src/ReplaceNonConstFCWithBatchMatMulPass.cpp
@@ -91,7 +91,6 @@ bool replace_fc_with_matmul(luci::CircleFullyConnected *fc)
 {
   luci::CircleNode *x = nullptr;
   luci::CircleNode *y = nullptr;
-  luci::CircleNode *b = nullptr;
   luci::CircleTranspose *ty = nullptr;
   luci::CircleTranspose *tx = nullptr;
   bool adj_x = false;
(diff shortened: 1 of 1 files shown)


seanshpark:
related https://github.com/Samsung/ONE/issues/3963#issuecomment-1341755697

compiler/luci/pass/src/ReplaceNonConstFCWithBatchMatMulPass.cpp
```diff
@@ -142,6 +138,8 @@ bool replace_fc_with_matmul(luci::CircleFullyConnected *fc)
   luci::add_origin(matmul, luci::get_origin(fc));
 
   auto all_zero = [](const luci::CircleConst *c) {
+    if (c->dtype() != loco::DataType::FLOAT32)
+      return false;
```
jinevening, line 142: Adding this check to `all_zero` seems not appropriate.

How about simply rejecting non-float const bias? For example, in line 123,

```cpp
auto bc = dynamic_cast<luci::CircleConst *>(fc->bias());
if (bc && bc->dtype() != loco::DataType::FLOAT32)
  return false;
```
This may simplify the codes around line 153.

seanshpark: Yes, thanks for suggestion :)

seanshpark: as we need to handle `bc == nullptr` case, won't be exactly like this...

compiler/luci/pass/src/ReplaceNonConstFCWithBatchMatMulPass.cpp
```diff
@@ -150,12 +146,16 @@ bool replace_fc_with_matmul(luci::CircleFullyConnected *fc)
     return ac;
   };
 
-  auto bc = dynamic_cast<luci::CircleConst *>(b);
-  if ((nullptr != bc) && !all_zero(bc))
+  auto bc = dynamic_cast<luci::CircleConst *>(fc->bias());
+  // NOTE bias can be empty as CircleOutputExclude type
+  // NOTE we can only handle bias as FLOAT32 type as of now
+  if (nullptr != bc && bc->dtype() != loco::DataType::FLOAT32)
+    return false;
```
jinevening, line 153: CircleBatchMatMul is already created at this point (which changes the graph). 

It would be better to move this code before line 130.

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #10176, opened 2022-12-08
[cir2cir-dredd-recipe-test] Revise README

seanshpark:
This will revise README file for text be inside 100 columns.


mhs4670go (approved):
LGTM

Issue #10185, opened 2022-12-13
[circle-quantizer] Support ABS Op quantization 


jinevening:
## What

Let's support quantization for ABS Op.

- [x] luci-interpreter kernel https://github.com/Samsung/ONE/pull/10186
- [x] QuantizeConstInputActivation #10195
- [x] InsertQuantizeOp (for mpq) #10195
- [x] Fake quantization #10255

jinevening:
Done. @BalyshevArtem Thanks for your help!

PR #10186, opened 2022-12-13
[luci-interpreter] Introduce Abs kernel

BalyshevArtem:
This commit adds Abs kernel to luci-interpreter.

for issue: #10185


AShedko (dismissed):
LGTM

compiler/luci-interpreter/src/kernels/Abs.cpp
```diff
@@ -0,0 +1,73 @@
...
+      break;
+    case DataType::S8:
+      eval<int8_t>();
+      break;
+    case DataType::U8:
+      eval<uint8_t>();
+      break;
+    case DataType::S16:
+      eval<int16_t>();
+      break;
```
jinevening, line 53: There are some issues in integer types.

- s32: Size of `int` differs across platforms.
- u8, s8, s16: Quantized value. These types should be handled differently.

Please remove them and only leave FLOAT32. FLOAT32 is enough for quantization.

compiler/luci-interpreter/src/kernels/Abs.cpp
```diff
@@ -0,0 +1,73 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "kernels/Abs.h"
+
+#include "kernels/Utils.h"
+
```
jinevening, line 20:
```suggestion
#include <cmath> // abs for float

```

compiler/luci-interpreter/src/kernels/Abs.cpp
BalyshevArtem: done

jinevening (approved):
LGTM

AShedko (approved):

PR #10189, opened 2022-12-14
[tools] Add an argument `--store-metadata` to `select_operator.py`

ragmani:
Add an argument `--store-metadata` to `select_operator.py` that allows metadata to be stored.

### DIFF
--- a/tools/tflitefile_tool/select_operator.py
+++ b/tools/tflitefile_tool/select_operator.py
@@ -23,6 +23,7 @@ import tflite.SubGraph
 import tflite.BuiltinOptions
 import argparse
 import pkg_resources
+import json
 
 
 # On flatbuffers 2.0, EndVector doesn't require length argument any more.
@@ -1142,10 +1143,10 @@ def GenerateSubgraphs(args, new_builder, sample_model, operator_list, new_input_
     # The selected subgraph will be primary subgraph of the model to be created newly
     selected_subgraph = sample_model.Subgraphs(args.subgraph)
 
-    # k: old subg index, v: new subg index
+    # k: orginal subg index, v: new subg index
     # new subg index is sequential in used_subgraphs_dic
     for k, v in used_subgraphs_dic.items():
-        print("Append subgraphs, old index : ", k, ", new index : ", v)
+        print("Append subgraphs, orginal index : ", k, ", new index : ", v)
         if k == args.subgraph:
             assert v == 0
             new_subgraph = GenerateSubgraph(new_builder, selected_subgraph, operator_list,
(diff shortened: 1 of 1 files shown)


ragmani:
Related to https://github.sec.samsung.net/STAR/nnfw/issues/12627#issuecomment-1800581

tools/tflitefile_tool/select_operator.py
```diff
@@ -1253,6 +1254,28 @@ def GenerateModel(args, new_builder, sample_model, operator_list, new_input_tens
     return tflite.Model.ModelEnd(new_builder)
 
 
+def StoreMetadata(path, used_tensors, new_inputs, new_outputs):
```
glistening, line 1257: @ragmani Are you going to add more information in `metadata`?
At this moment, it contains only map original indices to new indices for input/output.
It would be helpful to use more descriptive name if you're going to add more properties in metadata.

tools/tflitefile_tool/select_operator.py
```diff
@@ -1253,6 +1254,28 @@ def GenerateModel(args, new_builder, sample_model, operator_list, new_input_tens
...
+    output_old_idx = []
+    output_new_idx = []
+    for input_tensor_idx in new_inputs:
+        input_old_idx.append(input_tensor_idx)
+        input_new_idx.append(used_tensors[input_tensor_idx])
+    for output_tensor_idx in new_outputs:
+        output_old_idx.append(output_tensor_idx)
+        output_new_idx.append(used_tensors[output_tensor_idx])
+
+    metadata["inputs"] = {"old": input_old_idx, "new": input_new_idx}
```
glistening, line 1272: (optional) (about naming)

1. I thought input_old_idx is just a number. But it is a list which could contain multiple ones. Could you help me by renaming `idx` to `indices` or `indexes`?

2. I prefer `org` to `old`. I found there are two more `old` in this source. I found it was also added by you. : )

ragmani: > 1. I thought `input_old_idx` would be just a number. But it is a list which could contain multiple ones. Could you help me by renaming `idx` to `indices` or `indexes`?

OK, I will.

> 2. I prefer `org` to `old`. I found there are two more `old` in this source. I found it was also added by you. : )

Thanks, `org` looks good.  I will change all `old` to `org`.

tools/tflitefile_tool/select_operator.py
ragmani: I have no plan to add mode information in `metadata` yet.
How about changing the name to `IOInfo`?

hseok-oh (approved):

ragmani:
I added operand index, but I think I should have added io index. I will fix it.

glistening (approved):
LGTM. Thanks. It would be good if we add description about this option in https://github.com/Samsung/ONE/tree/master/tools/tflitefile_tool#readme.

PR #10206, opened 2022-12-20
[luci/svc] Revise OutputExclude dtype inference

seanshpark:
This will revise CircleOutputExclude dtype interference to be FLOAT32 when Unknown.


seanshpark:
from #10205

mhs4670go (approved):
LGTM

Issue #10207, opened 2022-12-20
[luci] FuseBatchNormWithTConvPass with ReLU


seanshpark:
Current `FuseBatchNormWithTConvPass` doesn't work with `Add` with `ReLU`.
Let's make this work.

seanshpark:
done

PR #10208, opened 2022-12-20
DRAFT FuseBatchNormWithTConvPass with ReLU

seanshpark:
on-going draft to support ReLU in FuseBatchNormWithTConvPass.


seanshpark:
for #10207

PR #10210, opened 2022-12-20
[res] Introduce Net_TConv_BN_005

seanshpark:
This will introduce Net_TConv_BN_005 recipe for TConv + BN fusion with ReLU.


jinevening (approved):
LGTM

PR #10212, opened 2022-12-20
[luci] Assert check for Activation None

seanshpark:
This will add assert check in FuseBatchNormWithTConvPass for Activation is None.


seanshpark:
from #10208

jinevening (approved):
LGTM

(optional) Perhaps switch statement would be a good choice.

seanshpark:
> (optional) Perhaps switch statement would be a good choice.

I'll revise to switch in another PR :)

PR #10214, opened 2022-12-20
[luci-pass-value-test] Add Net_TConv_BN_005

seanshpark:
This will enable Net_TConv_BN_005 test with ReLU.


seanshpark:
for  #10207 , from  #10208

PR #10215, opened 2022-12-20
[luci/pass] Revise FuseBatchNormWithTConvPass with switch

seanshpark:
This will revise FuseBatchNormWithTConvPass with switch instead of if-else.


seanshpark:
from https://github.com/Samsung/ONE/pull/10212#pullrequestreview-1223924816

seanshpark:
later, we can extract similar codes for ReLU and ReLU6 to a function.

jinevening (approved):
LGTM!

PR #10191, opened 2022-12-14
 [mpqsolver] Introduce bisection method

stamalakhov:
This extended draft introduces bisection method and some unit-tests for the task of automatic mixed precision quantization.

Previous draft is #10170
Related to: https://github.com/Samsung/ONE/issues/10154

Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

PR #10216, opened 2022-12-20
[mpqsolver] Introduce DepthParameterizer

stamalakhov:
This commit introduces DepthParameterizer for the task of automatic mixed precision quantization.

Draft: https://github.com/Samsung/ONE/pull/10191
Related to: https://github.com/Samsung/ONE/issues/10154

Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

### DIFF
--- a/compiler/circle-mpqsolver/src/bisection/DepthParameterizer.cpp
+++ b/compiler/circle-mpqsolver/src/bisection/DepthParameterizer.cpp
@@ -1,0 +1,95 @@
+/*
+ * Copyright (c) 2022 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "DepthParameterizer.h"
+
+namespace mpqsolver
+{
+namespace bisection
+{
+
+/**
+ * @brief compute maximal distance from graph inputs to graph nodes along with min/max values of
+ * distance and return status of computation (Assumes graph has no cycles)
+ */
+ParameterizerResult compute_depth(const luci::Module *module, NodeDepthType &nodes_depth,
+                                  float &min_depth, float &max_depth)
+{
+  if (module == nullptr)
+    return ParameterizerResult::FAILURE;
+
+  if (module->size() != 1)
+    return ParameterizerResult::FAILURE;
+
+  auto graph = module->graph(0);
+  if (!graph)
+    return ParameterizerResult::FAILURE;
+
+  // initializing
+  std::vector<luci::CircleNode *> to_process;
+  std::map<std::string, float> named_depth;
+  {
+    auto inputs = loco::input_nodes(graph);
+    for (auto &node : inputs)
+    {
+      auto cnode = loco::must_cast<luci::CircleNode *>(node);
+      to_process.emplace_back(cnode);
+      nodes_depth[cnode] = 0.f;
+      named_depth[cnode->name()] = 0.f;
+    }
+  }
+
+  // enumerating
+  while (!to_process.empty())
+  {
+    auto cur_node = to_process.back();
+    to_process.pop_back();
+    auto iter = nodes_depth.find(cur_node);
+    if (iter == nodes_depth.end())
+    {
+      return ParameterizerResult::FAILURE; // unexpected
+    }
+    float cur_depth = iter->second + 1;
+    // processing children
+    auto children = loco::succs(cur_node);
+    for (auto &child : children)
+    {
+      auto cichild = loco::must_cast<luci::CircleNode *>(child);
+      auto node_depth = nodes_depth.find(cichild);
+      if (node_depth == nodes_depth.end() || node_depth->second < cur_depth)
+      {
+        // initialize depth
+        nodes_depth[cichild] = cur_depth;
+        to_process.push_back(cichild);
+        named_depth[cichild->name()] = cur_depth;
+      }
+    }
+  }
+
+  // compute min/max of depth parameter
+  auto minmax = std::minmax_element(
+    nodes_depth.begin(), nodes_depth.end(),
+    [=](const std::pair<luci::CircleNode *, float> &el1,
+        const std::pair<luci::CircleNode *, float> &el2) { return el1.second < el2.second; });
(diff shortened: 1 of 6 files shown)


compiler/circle-mpqsolver/src/DepthParameterizer.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <luci/IR/Module.h>
+#include <luci/IR/CircleNodeDecl.h>
+
+namespace mpqsolver
+{
+
+using NodeDepthType = std::map<luci::CircleNode *, float>;
+
+int compute_depth(const luci::Module *module, NodeDepthType &nodes_depth, float &min_depth,
```
jinevening, line 28: Could you write comments about this function?

stamalakhov: Ahh... sorry. I'll write.

compiler/circle-mpqsolver/src/DepthParameterizer.h
```diff
@@ -0,0 +1,33 @@
...
+
+#ifndef __MPQSOLVER_DEPTH_PARAMETERIZER_H__
+#define __MPQSOLVER_DEPTH_PARAMETERIZER_H__
+
+#include <luci/IR/Module.h>
+#include <luci/IR/CircleNodeDecl.h>
+
+namespace mpqsolver
+{
+
```
jinevening, line 25: DepthParameterizer seems to be only related to BisectionSolver. Could you make a directory (e.g., `src/bisection`) and move these files there?

Also, adding one more namespace would be good for avoiding potential name conflict. For example, 

```
namespace mpqsolver
{
namespace bisection
{
...
```

stamalakhov: ok. I'll do.

compiler/circle-mpqsolver/src/DepthParameterizer.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "DepthParameterizer.h"
+
+namespace mpqsolver
+{
+
+int compute_depth(const luci::Module *module, NodeDepthType &nodes_depth, float &min_depth,
+                  float &max_depth)
+{
+  if (module == nullptr)
+    return EXIT_FAILURE;
```
jinevening, line 26: Better to define enum class for the return code.

stamalakhov: ok.

stamalakhov: fixed

compiler/circle-mpqsolver/src/bisection/DepthParameterizer.cpp
```diff
@@ -0,0 +1,108 @@
...
+  if (module->size() != 1)
+    return ParameterizerResult::FAILURE;
+
+  auto graph = module->graph(0);
+  if (!graph)
+    return ParameterizerResult::FAILURE;
+
+  auto nodes = graph->nodes();
+  uint32_t nodes_size = nodes->size();
+  std::set<std::string> input_names;
```
mhs4670go, line 43: I think `input_names` is not used. Maybe inserted for debugging purpose?

stamalakhov: Yes. It was done for debugging.I'll remove them. Thank you.

compiler/circle-mpqsolver/src/bisection/TestHelper.h
```diff
@@ -0,0 +1,79 @@
+/*
```
seanshpark, line 1: DRAFT shows that this file is shared, but has moved into `bisection` folder.
Is this OK?

compiler/circle-mpqsolver/src/bisection/DepthParameterizer.test.cpp
```diff
@@ -0,0 +1,95 @@
...
+  g.init();
+  auto conv = g._conv;
+  auto input = g._input;
+  auto output = g._output;
+
+  g.transfer_to(m.get());
+
+  mpqsolver::bisection::NodeDepthType nodes_depth;
+  float min_depth = std::numeric_limits<float>().max();  // should be  zero
+  float max_depth = -std::numeric_limits<float>().max(); // should be 2 (at output)
```
seanshpark, line 76: comment `should be zero` and `should be 2` seems confusing.
are these values be after calling `compute_depth()` ? if so, it would be better to move these comments.

compiler/circle-mpqsolver/src/bisection/TestHelper.h
stamalakhov: Currently the only implemented method is bisection. So for now it's ok.

compiler/circle-mpqsolver/src/bisection/DepthParameterizer.test.cpp
seanshpark: or as below `EXPECT_TRUE(max_depth == 2 && min_depth == 0);` is stated comment may be not needed.

stamalakhov: > comment `should be zero` and `should be 2` seems confusing. are these values be after calling `compute_depth()` ? if so, it would be better to move these comments.

Yes. It depicts their values, after calling compute_depth.

stamalakhov: I'll remove them.

compiler/circle-mpqsolver/src/bisection/TestHelper.h
seanshpark: > So for now it's ok.

If you plan to move this file, it would be better to put it where it should by as you have the full draft.

stamalakhov: > If you plan to move this file, it would be better to put it where it should be as you have the full draft.

I see. I expect all modules needed for `bisection` algorithm will placed into `bisection` folder, so unit-tests will also be placed into `bisection` folder. Someday later the situation may change, then we can make some refactoring.

compiler/circle-mpqsolver/src/bisection/DepthParameterizer.test.cpp
stamalakhov: Fixed.

compiler/circle-mpqsolver/src/bisection/DepthParameterizer.cpp
stamalakhov: Fixed

compiler/circle-mpqsolver/src/bisection/TestHelper.h
seanshpark: > then we can make some refactoring.

OK got it

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

jinevening (approved):
LGTM

PR #10217, opened 2022-12-20
[luci/pass] Extract replace_with_relu

seanshpark:
This will extract common codes replace_with_relu() from FuseBatchNormWithTConvPass.


jinevening (approved):
LGTM

PR #10218, opened 2022-12-20
[luci] Tidy OutputExclude dtype settings

seanshpark:
This will tidy dummy dtype settings for OutputExclude nodes.


mhs4670go (approved):
LGTM

PR #10146, opened 2022-11-29
[record-minmax] Revise to prepare to multiple instance

BalyshevArtem:
This pr revises record-minmax to prepare to using multiple instance of luci-interpreters and observers.

for issue https://github.com/Samsung/ONE/issues/10133
draft: #10132
first step according to https://github.com/Samsung/ONE/pull/10132#discussion_r1034138863

ONE-DCO-1.0-Signed-off-by: Artem Balyshev <a.balyshev@samsung.com>

PR #10223, opened 2022-12-21
[record-minmax] Use nncc_common

jinevening:
This updates code to use nncc_common.

---
Related to: https://github.com/Samsung/ONE/pull/10146#issuecomment-1360557336

### DIFF
--- a/compiler/record-minmax/include/RecordFunction.h
+++ b/compiler/record-minmax/include/RecordFunction.h
@@ -69,13 +69,13 @@ float getMovingAverage(const std::vector<float> &vector, const float alpha,
   assert(alpha >= 0.0 && alpha <= 1.0);
   assert(batch_size > 0);
 
-  auto getBatchMinOrMax = [&](int start_index) {
-    assert(start_index >= 0 && start_index < vector.size());
+  auto getBatchMinOrMax = [&](uint32_t start_index) {
+    assert(start_index < vector.size());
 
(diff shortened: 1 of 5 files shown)


compiler/record-minmax/include/RecordFunction.h
```diff
@@ -69,13 +69,13 @@ float getMovingAverage(const std::vector<float> &vector, const float alpha,
   assert(alpha >= 0.0 && alpha <= 1.0);
   assert(batch_size > 0);
 
-  auto getBatchMinOrMax = [&](int start_index) {
-    assert(start_index >= 0 && start_index < vector.size());
+  auto getBatchMinOrMax = [&](uint32_t start_index) {
+    assert(start_index < vector.size());
```
jinevening, line 73: I removed `start_index >= 0`, which is redundant for uint32_t type.

seanshpark (approved):
LGTM

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -247,7 +247,7 @@ void RecordMinMax::profileRawDataDirectory(const std::string &mode,
     total_input_size += getTensorSize(input_node);
   }
 
-  while (entry = readdir(dp))
+  while ((entry = readdir(dp)))
```
jinevening, line 250: To avoid warning to add parentheses when using assignment as a boolean.

Issue #10133, opened 2022-11-25
[record-minmax] Add parallel recording


BalyshevArtem:
### What
 
Let's add parallelism to record-minmax with help of OpenMP.

### Why

To speed up recording

### How

Let's add parallel recording for different input data format with plan:
- [x] add `h5` input_data_format #10132
- [ ] add `list` input_data_format
- [ ] add `directory` input_data_format
- [ ] add for random inputs

BalyshevArtem:
with #10132 for [mobilenet_v1_1.0_224.circle.txt](https://github.com/Samsung/ONE/files/10091094/mobilenet_v1_1.0_224.circle.txt) and for random `h5` file with 50 num records I had the following result (host has 8 threads)(using `./time`):
- without parallel recording: 
```
real	8m54,355s
user	8m54,286s
sys	0m0,048s
```

- with parallel recording: 
```
real	1m30,052s
user	10m8,754s
sys	0m0,671s
```

jinevening:
Looks very good feature! Before landing this feature, we need some tests.

1. Option to turn off OpenMP for supporting other platforms e.g., Arm. @mhs4670go already left a comment. PTAL https://github.com/Samsung/ONE/pull/10132#issuecomment-1327465549.

2. Value check. Min/max values recorded in sequential/parallel settings must be the same. The easiest way to add a test would be to add `pota_parallel_record_minmax_test` in `pota-quantization-value-test`.

FYI, I ran `pota-quantization-value-test` with `parallel_record` (by modifing `test_record_minmax.sh`), and it failed.

```
104/111 Test  #98: pota_record_minmax_test ....................***Failed    5.19 sec
-- Found RECORD-MINMAX: /home/hjj/ONE/build/release/compiler/record-minmax/record-minmax
-- Found CIRCLE_TENSORDUMP: /home/hjj/ONE/build/release/compiler/circle-tensordump/circle-tensordump
-- Found workdir: /home/hjj/ONE/build/release/compiler/common-artifacts
~/ONE/build/release/compiler/common-artifacts ~/ONE/build/release/compiler/pota-quantization-value-test
~/ONE/build/release/compiler/pota-quantization-value-test
FAILED
- Add_002.layer.uint8
- Add_002.channel.int16
...
```

BalyshevArtem:
```
[...]

PR #10228, opened 2022-12-21
[record-minmax] Add parallel recording

BalyshevArtem:
This PR adds parallel recording to record-minmax.

for issue https://github.com/Samsung/ONE/issues/10133
from draft: https://github.com/Samsung/ONE/pull/10132


compiler/record-minmax/driver/Driver.cpp
```diff
@@ -97,6 +101,12 @@ int entry(const int argc, char **argv)
   if (arser["--min_percentile"])
     min_percentile = arser.get<float>("--min_percentile");
 
+  if (arser["--num_threads"])
+    num_threads = arser.get<int>("--num_threads");
+
+  if (num_threads < 1)
+    throw std::runtime_error("The number of threads must be greater than zero\n");
```
seanshpark, line 108: optional) not sure we need `\n`

compiler/record-minmax/driver/Driver.cpp
```diff
@@ -114,17 +124,35 @@ int entry(const int argc, char **argv)
...
+
     if (input_data_format == "h5" || input_data_format == "hdf5")
     {
       // Profile min/max while executing the H5 data
-      rmm.profileData(mode, input_data_path, min_percentile, max_percentile);
+      if (num_threads == 1)
+        rmm.profileData(mode, input_data_path, min_percentile, max_percentile);
+      else
+      {
+        std::cout << "Using parallel recording" << std::endl;
```
seanshpark, line 153: there's `add_verbose()` used. maybe show this message only when some level is given.

compiler/record-minmax/include/RecordMinMax.h
```diff
@@ -58,6 +61,8 @@ class RecordMinMax
   luci_interpreter::Interpreter *getInterpreter() const { return _interpreters[0].get(); }
   MinMaxObserver *getObserver() const { return _observers[0].get(); }
 
+  std::vector<std::vector<std::vector<char>>> importH5Data(const std::string &input_data_path);
```
seanshpark, line 64: the `vector` is not friendly to read.
is there any particular reason to use so?

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -348,6 +348,65 @@ void RecordMinMax::profileRawData(const std::string &mode, const std::string &in
   update_quantparam(getObserver(), mode, min_percentile, max_percentile);
 }
 
+std::vector<std::vector<std::vector<char>>>
+RecordMinMax::importH5Data(const std::string &input_data_path)
```
seanshpark, line 352: Q) will this read all the files in the path? what if there are lots of files that exceeds memory?

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -413,6 +472,81 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+                                         float max_percentile)
+{
+  assert(_interpreters.size() == _threads_size);
+  assert(_observers.size() == _threads_size);
+  const auto vector_input_data = importH5Data(input_data_path);
+  const auto num_records = vector_input_data.size();
+  const auto input_nodes = loco::input_nodes(_module->graph());
+
+  // Start parallel part
+  std::cout << _threads_size << " concurrent threads are supported.\n";
```
seanshpark, line 486: optional) maybe use verbose option

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -413,6 +472,81 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+
+  if (num_records < _threads_size)
+  {
+    _threads_size = num_records;
+  }
```
seanshpark, line 491: as I understand, this method runs and exits the problem and this code is OK.
but I think using local variable for this would be better for future changes.
```suggestion
  auto run_threads = num_records < _threads_size ? num_records : _threads_size 
```

compiler/record-minmax/driver/Driver.cpp
```diff
@@ -114,17 +124,35 @@ int entry(const int argc, char **argv)
 
   RecordMinMax rmm(num_threads);
 
+  // TODO: support parallel record for profile with random data
+  if (num_threads > 1 and not arser["--input_data"])
+  {
+    throw std::runtime_error("Parallel recording is used only for h5 now\n");
```
jinevening, line 130:
```suggestion
    throw std::runtime_error("Input data must be given for parallel recording");
```

compiler/record-minmax/src/RecordMinMax.cpp
jinevening: @BalyshevArtem It would be better to catch `std::bad_alloc` to handle the case.

```
  catch (const std::bad_alloc &e)
  {
    throw std::runtime_error("Out of memory during h5 data load.");
  }
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -413,6 +472,81 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+  uint32_t records_batch = static_cast<int>(num_records / _threads_size);
```
jinevening, line 492:
```suggestion
  uint32_t records_batch = static_cast<uint32_t>(num_records / _threads_size);
```

compiler/record-minmax/src/RecordMinMax.cpp
jinevening: @BalyshevArtem , I and @seanshpark had offline talk. We're concerned about excessive memory consumption. Could you add a code to check the size of input data and reject if the whole data size is too large ? We think 1 GB of the max byte size is acceptable.

Reducing the memory usage of parallel recording would be a future development item, which is necessary to enable parallel recording in our public toolchain (one-quantize).

compiler/record-minmax/include/RecordMinMax.h
BalyshevArtem: Actually no, just convenience.
And then what is better to use in this case?

compiler/record-minmax/src/RecordMinMax.cpp
BalyshevArtem: Done

compiler/record-minmax/src/RecordMinMax.cpp
BalyshevArtem: Done

compiler/record-minmax/include/RecordMinMax.h
seanshpark: > And then what is better to use in this case?

Not using three vectors.

Why not use unique_ptr to memory for inner most vector<char> ? 
or refer recent PR uses with `using` alias; https://github.com/Samsung/ONE/pull/10234/files#diff-e4360f1df0c1003cc097406de284d98870516a540ebdd4b19181e1a4d25852a3

compiler/record-minmax/driver/Driver.cpp
```diff
@@ -114,17 +127,35 @@ int entry(const int argc, char **argv)
...
   rmm.initialize(input_model_path);
 
   if (arser["--input_data"])
   {
     auto input_data_path = arser.get<std::string>("--input_data");
 
+    // TODO: support parallel record from file and dir input data format
+    if (num_threads > 1 and not(input_data_format == "h5") and not(input_data_format == "hdf5"))
+    {
+      throw std::runtime_error("Parallel recording is used only for h5 now\n");
```
seanshpark, line 146: need `\n` here too?

compiler/record-minmax/driver/Driver.cpp
```diff
@@ -20,6 +20,7 @@
 #include <vconone/vconone.h>
 
 #include <luci/UserSettings.h>
+#include <luci/Log.h>
```
seanshpark, line 23: This module uses _luci_ but itself is _record-minmax_ , thus I think seeing LUCI prefix may confuse the users reading log.
Defining LOG will take some time so please add a TODO
```suggestion
// TODO declare own log signature of record-minmax
#include <luci/Log.h>
```

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -38,6 +39,16 @@ using DataType = loco::DataType;
 
 namespace
 {
+// Max h5 file size for parallel recording in bytes = 1 GB
```
seanshpark, line 42: please add an empty line above this comment

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -413,6 +483,98 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+                                         float max_percentile)
+{
+  LOGGER(l);
+
+  assert(_interpreters.size() == _threads_size);
+  assert(_observers.size() == _threads_size);
+
+  const long h5_file_size = getH5FileSize(input_data_path);
+
+  std::cout << "File size = " << h5_file_size << std::endl;
```
seanshpark, line 497: maybe for debugging?

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -413,6 +483,98 @@ void RecordMinMax::profileData(const std::string &mode, const std::string &input
...
+
+  if (h5_file_size > h5_max_size_bytes)
+    throw std::runtime_error("H5 file size is too large");
```
jinevening, line 500:
```suggestion
    throw std::runtime_error("H5 file size is too large for parallel recording");
```

compiler/record-minmax/src/RecordMinMax.cpp
BalyshevArtem: Yes, forgot to delete it, thank you!

compiler/record-minmax/src/RecordMinMax.cpp
BalyshevArtem: Done

compiler/record-minmax/driver/Driver.cpp
BalyshevArtem: Done

jinevening (dismissed):
LGTM

SlavikMIPT (dismissed):
LGTM

compiler/record-minmax/include/RecordMinMax.h
seanshpark, line 64: plz do not use 3D vector but use logical meaning of it, with `using` alias

seanshpark (changes requested):

compiler/record-minmax/include/RecordMinMax.h
jinevening: I think the below aliasing is ok.

std::vector<char> -> Buffer
std::vector<Buffer> -> InputData
std::vector<InputData> -> InputDataset

BalyshevArtem: Done

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -348,6 +360,64 @@ void RecordMinMax::profileRawData(const std::string &mode, const std::string &in
...
+    {
+      if (num_inputs != static_cast<uint32_t>(importer.numInputs(i)))
+        throw std::runtime_error("Wrong number of inputs.");
+
+      for (uint32_t input_idx = 0; input_idx < num_inputs; input_idx++)
+      {
+        const auto *input_node = loco::must_cast<const luci::CircleInput *>(input_nodes[input_idx]);
+        assert(input_node->index() == input_idx);
+        checkInputDimension(input_node);
+        std::vector<char> input_data(getTensorSize(input_node));
```
seanshpark, line 392: As @jinevening commented, can you use `std::vector<char>` as a meaning for this?

And revise `using BufferInputData = std::vector<std::vector<std::vector<char>>>;`
too.

seanshpark (changes requested):

seanshpark:
Please read the comment https://github.com/Samsung/ONE/pull/10228#discussion_r1055915792 and use something like this.
```
using Buffer = std::vector<char>;
using Output = std::vector<Buffer>;
using WholeOutput = std::vector<Output>;
```

compiler/record-minmax/src/RecordMinMax.cpp
BalyshevArtem: Done

seanshpark (approved):
LGTM

SlavikMIPT (approved):
LGTM, thank you

jinevening (approved):
LGTM Thank you

PR #10229, opened 2022-12-21
[npud] Set request data in Trix backend

jyoungyun:
This commit support to set request data in Trix backend.


runtime/service/npud/backend/trix/TrixBackend.cc
```diff
@@ -307,8 +307,105 @@ NpuStatus TrixBackend::setRequestData(NpuContext *ctx, RequestID requestId, Inpu
...
+  {
+    return NPU_STATUS_ERROR_INVALID_ARGUMENT;
+  }
+
+  auto &req = _dev->requests.at(requestId);
+  auto miter = std::find(ctx->models.begin(), ctx->models.end(), req->modelId);
+  if (miter == ctx->models.end())
+  {
+    return NPU_STATUS_ERROR_INVALID_MODEL;
+  }
```
hseok-oh, line 328: Can't we use try-catch from `std::map::at(...)`'s `std::out_of_range` exception?

runtime/service/npud/backend/trix/TrixBackend.cc
```diff
@@ -307,8 +307,105 @@ NpuStatus TrixBackend::setRequestData(NpuContext *ctx, RequestID requestId, Inpu
...
+  auto &minfo = _dev->models.at(req->modelId);
+  if (minfo->meta->input_seg_num != inputBufs->numBuffers ||
+      minfo->meta->output_seg_num != outputBufs->numBuffers)
+  {
+    return NPU_STATUS_ERROR_INVALID_DATA;
+  }
+
+  auto &iter = _dev->requests.at(requestId);
+  auto &inInfos = iter->inInfos;
+  auto &outInfos = iter->outInfos;
```
hseok-oh, line 340: `_dev->requests.at(requestId)` duplication

```suggestion
  auto &inInfos = req->inInfos;
  auto &outInfos = req->outInfos;
```

runtime/service/npud/backend/trix/TrixBackend.cc
jyoungyun: Um, before calling `at` function, it already checks whether the `requestId` is included in ctx requests array. Still, should I use `try-catch` phrase?

jyoungyun: Oh, do you mention it because of my TODO comment?

hseok-oh: >Um, before calling `at` function, it already checks whether the `requestId` is included in ctx requests array. Still, should I use `try-catch` phrase?

My comment means that we can remove `std::find(...)` checking if we use exception.

jyoungyun: @hseok-oh 
Um, I can't find a way to declare the reference variable of the unique_ptr item outside the `try-catch` block. I don't want to use `get()` function of unique_ptr... Could you suggest an elegant way to do this?

```
  TrixRequestInfo *req;
  try
  {
    req = _dev->requests.at(requestId).get();
  }
  catch(const std::exception &e)
  {
    return NPU_STATUS_ERROR_INVALID_ARGUMENT;
  }
```

hseok-oh: I also don't have any idea for elegant implementation. :)

jyoungyun:
PTAL

jyoungyun:
PTAL. Thanks in advance. :)

hseok-oh (approved):

Issue #9947, opened 2022-10-28
[CFE] Reduce Duplicate Code 


seanshpark:
Our recent DC score is 4.04 and seems near the edge.
Let's make this score higher.

seanshpark:
Top 10

No. | Metric | BlockKey | Path | LineStart | LineEnd | LineCount | Token | Violation
-- | -- | -- | -- | -- | -- | -- | -- | --
1 | DC | 1 | compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp | 1099 | 1175 | 77 | 351 | Y
2 | DC | 1 | compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp | 889 | 964 | 76 | 351 | Y
3 | DC | 1 | compiler/luci/pass/src/ConvertNCHWToNHWCPass.cpp | 1175 | 1250 | 76 | 351 | Y
4 | DC | 3 | compiler/luci/pass/src/ForwardTransposeOpPass.cpp | 164 | 219 | 56 | 286 | Y
5 | DC | 3 | compiler/luci/pass/src/ForwardTransposeOpPass.cpp | 221 | 276 | 56 | 286 | Y
6 | DC | 67 | compiler/logo/src/Passes/ReorderDecodePass.cpp | 83 | 131 | 49 | 102 | Y
7 | DC | 67 | compiler/logo/src/Passes/ReorderDecodePass.cpp | 184 | 231 | 48 | 102 | Y
8 | DC | 2 | compiler/circle-eval-diff/src/MetricPrinter.cpp | 196 | 234 | 39 | 306 | Y
9 | DC | 47 | compiler/luci/pass/src/TransformMinMaxToRelu6Pass.cpp | 25 | 63 | 39 | 115 | Y
10 | DC | 47 | compiler/luci/pass/src/TransformMinReluToRelu6Pass.cpp | 25 | 63 | 39 | 115 | Y

jinevening:
Ah, I wrote many of them. I'll take a look. Thanks for sharing :)

seanshpark:
This is not urent, so please take your time :)

seanshpark:
I'll later show you how to get full report at the office (today is my WFH)

jinevening:
@mhs4670go showed interests in this task :)

jinevening:
@mhs4670go ping

seanshpark:
close as inactive for long time

PR #10250, opened 2022-12-27
[luci/pass] Reduce duplicates of ForwardTransposeOpPass

mhs4670go:
This commit reduces duplicates of ForwardTransposeOpPass.

Related : #9947 

### DIFF
--- a/compiler/luci/pass/src/ForwardTransposeOpPass.test.cpp
+++ b/compiler/luci/pass/src/ForwardTransposeOpPass.test.cpp
@@ -119,6 +119,24 @@ public:
   }
 };
 
+class ForwardTransposeToAddInvalidGraph : public TestIOGraph, public TransposeAddGraphlet
+{
+public:
+  void init(const ShapeU32 shape_in, const ShapeU32 shape_out)
+  {
+    TestIOGraph::init(shape_in, shape_out);
+    TransposeAddGraphlet::init(g(), shape_in, shape_out);
+
+    // connect network
+    _transpose->a(input());
+    _transpose->perm(_perm);
+    _binary->x(_transpose);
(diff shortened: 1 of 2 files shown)


jinevening (approved):
LGTM

compiler/luci/pass/src/ForwardTransposeOpPass.test.cpp
```diff
@@ -308,6 +335,14 @@ TEST_F(ForwardTransposeToAddGraphTest, forward_transpose_add_NEG)
...
 }
 
+TEST_F(ForwardTransposeToAddGraphNegTest, forward_transpose_add_non_const_NEG)
+{
+  _graph.init({1, 64, 51, 1}, {0, 3, 2, 1});
+
+  luci::ForwardTransposeOpPass pass;
+  EXPECT_FALSE(pass.run(_graph.g()));
+}
+
```
jinevening, line 345: Why is this test added in this PR?

mhs4670go: Well, for the case where binary op has no const operaters. Is this case going to be covered later?

jinevening: Adding a new test seems not related to reducing duplicate codes.

Is this for increasing the portion of negative test?

mhs4670go: Ah, sort of. The test is included to prevent minor bugs that may occur during refactoring.

jinevening: I see.

seanshpark (approved):

PR #10004, opened 2022-11-09
Draft: ICompiler and IExecutors (WIP)

hseok-oh:
Draft for `ICompiler` and `IExecutors`

Related issue: #10000

PR #10265, opened 2022-12-29
[onert] Introduce MultiModelCompiler

hseok-oh:
This commit introduces MultiModelCompiler to separate single and multi model nnpackage compilation.

---

Draft: #10004

### DIFF
--- a/runtime/onert/core/src/compiler/Compiler.cc
+++ b/runtime/onert/core/src/compiler/Compiler.cc
@@ -42,40 +42,25 @@ namespace onert
 namespace compiler
 {
 
-Compiler::Compiler(const std::shared_ptr<ir::Model> &model, CompilerOptions &copt)
-  : _nnpkg{std::make_shared<ir::NNPkg>(model)}, _voptions{&copt}
-{
-  // DO NOTHING
-}
-
-Compiler::Compiler(const std::shared_ptr<ir::NNPkg> &nnpkg,
-                   std::vector<std::unique_ptr<CompilerOptions>> &copts)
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/compiler/Compiler.cc
```diff
@@ -106,22 +91,12 @@ std::shared_ptr<CompilerArtifact> Compiler::compile(void)
...
-      throw std::runtime_error{"NYI: Disable compilation for multi model is not supported yet"};
-
-    auto executors = std::make_shared<exec::Executors>();
-
-    _nnpkg->primary_model()->iterate([&](const ir::SubgraphIndex &index, ir::Graph &subg) {
-      executors->emplace(ir::ModelIndex{0}, index, std::make_unique<interp::InterpExecutor>(subg));
-    });
-    return std::make_shared<CompilerArtifact>(executors, nullptr);
-  }
+    throw std::runtime_error{"NYI: Disable compilation for multi model is not supported yet"};
```
glistening, line 94: @hseok-oh Is it enough to check `_voptions[0]` only? If not, we may introduce option iteration loop to check `disable_compile` and `he_profiling_mode` in the loop.

hseok-oh: I'll move option check into for loop.

glistening (approved):
LGTM.

`Compiler` is a class for a single model compiler (typical test) while `MoltiModelCompiler` is for multiple models (special case).

jyoungyun (approved):
LGTM

PR #10266, opened 2022-12-29
[luci-evel-driver] Tidy dependency

seanshpark:
This will tidy dependency and remove unnecessary headers.

### DIFF
--- a/compiler/luci-eval-driver/src/EvalDriver.cpp
+++ b/compiler/luci-eval-driver/src/EvalDriver.cpp
@@ -16,11 +16,10 @@
 
 #include <luci/ImporterEx.h>
 #include <luci_interpreter/Interpreter.h>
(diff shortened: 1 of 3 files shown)


compiler/luci-eval-driver/src/EvalDriver.cpp
```diff
@@ -16,11 +16,10 @@
 
 #include <luci/ImporterEx.h>
 #include <luci_interpreter/Interpreter.h>
-#include <luci/CircleExporter.h>
-#include <luci/CircleFileExpContract.h>
 
 #include <cstdlib>
 #include <fstream>
+#include <iostream>
```
seanshpark, line 22: added as `std::cout` is used

mhs4670go (approved):
LGTM

PR #10269, opened 2023-01-02
[circle-part] Tidy value test CMake

seanshpark:
This will remove dependency from folder as it doesn't work correctly.


seanshpark:
@mhs4670go , before this change, `command` was executed in every build. could you double check if this happens to you too and with this change, it's fixed?

mhs4670go (approved):

mhs4670go:
@seanshpark It works:)

PR #10272, opened 2023-01-02
[common-artifacts] Tidy CMake

seanshpark:
This will remove dependency from folder as it doesn't work correctly.


mhs4670go (approved):
LGTM

PR #10274, opened 2023-01-03
[one-cmds] Install python model2nnpkg

hseok-oh:
Let's install python model2nnpkg script instead of deprecated bash script for one-pack.


seanshpark (approved):
LGTM thank you!

PR #10290, opened 2023-01-04
[onert] Support non-linear multi-model

ragmani:
This commit makes onert support non-linear multi-model.

---

For #10002 
Draft #10041


ragmani:
This commit does not support non-linear multi-model. I'll remake this as another PR.

PR #10296, opened 2023-01-05
[onert] Enhance supporting multi-model

ragmani:
This commit makes onert support multi-model based on always `model2 > model1` if edges `model1 -> model2`.
  - Use IOTensor instead of IODesc for inputs/outputs executors
  - Use sharing buffers instead of moving buffers between executors.

---

For #10002 
Draft #10041

### DIFF
--- a/runtime/onert/core/src/exec/Executors.cc
+++ b/runtime/onert/core/src/exec/Executors.cc
@@ -195,8 +195,18 @@ void Executors::executeModels(const IODescription &desc)
   //      and all inputs/outputs become IPortableTensor at compile stage.
   //      This allows user's buffers to be set to inputs/outputs of executors.
   // Shared buffers and tensors between executors
-  std::vector<std::unique_ptr<uint8_t[]>> edge_bufs;
-  std::vector<std::unique_ptr<backend::builtin::IOTensor>> edge_tensor;
+  std::vector<std::unique_ptr<backend::builtin::IOTensor>> edge_tensors;
+  std::unordered_map<ir::IODesc, std::unique_ptr<uint8_t[]>> internal_bufs;
+
+  // Reference count for each `from` of edges to reduce peak memory
+  std::unordered_map<ir::IODesc, int32_t> internal_tensor_ref_counts(_model_edges->edges.size());
+  for (const auto &edge : _model_edges->edges)
+  {
+    if (internal_tensor_ref_counts.find(edge.from) == internal_tensor_ref_counts.end())
+      internal_tensor_ref_counts[edge.from] = 1;
+    else
+      internal_tensor_ref_counts[edge.from]++;
+  }
 
   // Execute each model
   // NOTE May be better to use vector instead of unordered_map for _executors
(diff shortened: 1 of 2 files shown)


ragmani:
```bash
$ cat iv3/metadata/MANIFEST 
{
  "major-version": "1",
  "minor-version": "2",
  "patch-version": "0",
  "configs": [
    ""
  ],
  "models": [
    "i1.tflite",
    "i2.tflite",
    "i3.tflite",
    "i4.tflite",
    "i5.tflite"
  ],
  "model-types": [
    "tflite",
    "tflite",
    "tflite",
    "tflite",
    "tflite"
  ],
  "pkg-inputs": [
    "0:0:0"
  ],
  "pkg-outputs": [
    "4:0:0"
  ],
  "model-connect": [
    {
      "from": "0:0:0",
      "to": [
        "1:0:0"
      ]
    },
    {
      "from": "0:0:1",
      "to": [
        "1:0:1"
      ]
    },
    {
      "from": "0:0:2",
      "to": [
        "1:0:2"
      ]
    },
    {
      "from": "1:0:0",
      "to": [
        "2:0:0",
        "3:0:0"
      ]
    },
    {
      "from": "2:0:0",
      "to": [
        "4:0:0"
      ]
    },
    {
      "from": "2:0:1",
      "to": [
        "4:0:1"
      ]
    },
    {
      "from": "3:0:0",
      "to": [
        "4:0:2"
      ]
    },
    {
      "from": "3:0:1",
      "to": [
        "4:0:3"
      ]
    }
  ]
}
$ BACKENDS=cpu Product/x86_64-linux.release/out/bin/onert_run iv3 --load iv3/input.h5 --dump iv3/result.h5 
Package Filename iv3
0:0:0 -> 1:0:0
0:0:1 -> 1:0:1
0:0:2 -> 1:0:2
1:0:0 -> 2:0:0
1:0:0 -> 3:0:0
2:0:0 -> 4:0:0
2:0:1 -> 4:0:1
3:0:0 -> 4:0:2
3:0:1 -> 4:0:3
MODEL_LOAD   takes 41.317 ms
PREPARE      takes 75.240 ms
EXECUTE      takes 70.666 ms
- MEAN     :  70.666 ms
- MAX      :  70.666 ms
- MIN      :  70.666 ms
- GEOMEAN  :  70.666 ms

$ h5diff -v --delta=0.00001 iv3/result.h5 iv3/expected.h5 

file1     file2
---------------------------------------
    x      x    /              
           x    /name          
    x      x    /value         
    x      x    /value/0       

group  : </> and </>
0 differences found
group  : </value> and </value>
0 differences found
dataset: </value/0> and </value/0>
0 differences found
```

ragmani:
@nnfw-bot test x64-builder"

hseok-oh:
@nnfw-bot test x64-builder

hseok-oh:
@nnfw-bot test x64-builder

hseok-oh:
@nnfw-bot test all

hseok-oh:
Q1. What happen if model's output is used as other model's input and pkg's output at once?
Q2. Isn't it waste memory if pkg uses 3 models and 1st model's output is used as 2nd model's input?

ragmani:
If an input or output of executor is one of inputs/outputs of pkg, the `IOTensor` will be set buffer of user(pkg).
If an input or output of executor is not inputs/outputs of pkg, the `IOTensor` will be set internal buffer.

> Q1. What happen if model's output is used as other model's input and pkg's output at once?

All will use user buffer. 

> Q2. Isn't it waste memory if pkg uses 3 models and 1st model's output is used as 2nd model's input?

It's not. Currently, executor's inputs/outputs are always `IPortableTensor` and have to be set buffers externally regardless of backend. Only necessary buffers of internal tensors are created.

hseok-oh:
> > Q1. What happen if model's output is used as other model's input and pkg's output at once?
> 
> All will use user buffer.

:+1:

runtime/onert/core/src/exec/Executors.cc
```diff
@@ -255,24 +256,25 @@ void Executors::executeModels(const IODescription &desc)
...
+        outputs_inter[i] = pkgs_outputs[output_pkg_index].get();
+      }
+      else
+      {
+        assert(edge_tensor.size() == edge_bufs.size());
+        const auto temp_index = edge_tensor.size();
+        edge_bufs.emplace_back(std::make_unique<uint8_t[]>(info.total_size()));
+        edge_tensor.emplace_back(std::make_unique<backend::builtin::IOTensor>(info, layout));
+        edge_tensor.at(temp_index)->setUserTensor(edge_bufs[temp_index].get(), info.total_size());
+        outputs_inter[i] = edge_tensor[temp_index].get();
```
hseok-oh, line 273: > > Q2. Isn't it waste memory if pkg uses 3 models and 1st model's output is used as 2nd model's input?
> 
> It's not. Currently, executor's inputs/outputs are always `IPortableTensor` and have to be set buffers externally regardless of backend. Only necessary buffers of internal tensors are created.

Then on example, 1st model's output buffer is allocated here, and when runtime executes 3rd model, it doesn't free useless 1st model's output tensor buffer. Then It may increase peak memory.

ragmani: > Then on example, 1st model's output buffer is allocated here, and when runtime executes 3rd model, it doesn't free useless 1st model's output tensor buffer. Then It may increase peak memory.

Oh, you are right. peak memory may increase. It is more necessary to count the number of edges for each `from` and to free memory when the count becomes 0 after executing the executor.

hseok-oh (approved):
LGTM

PR #10318, opened 2023-01-11
[onert] Change the time measurement method

jyoungyun:
This commit uses `clock_gettime` instead of `gettimeofday`. POSIX.1-2008 marks gettimeofday() as obsolete,
and recommending the use of `clock_gettime` instead. https://man7.org/linux/man-pages/man2/settimeofday.2.html

### DIFF
--- a/runtime/libs/benchmark/src/Phases.cpp
+++ b/runtime/libs/benchmark/src/Phases.cpp
@@ -22,16 +22,16 @@
 #include <cassert>
 #include <chrono>
 #include <iostream>
-#include <sys/time.h>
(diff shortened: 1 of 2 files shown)


hseok-oh (approved):
LGTM

hseok-oh (approved):

glistening (approved):
LGTM

- We need to use same time functions to compare execution time between tflite and onert.
- I confirmed TensorFlow lite 1.14.0 used `gettimeofday`, latest version (v2.11.0) uses `clock_gettime`.

PR #10329, opened 2023-01-12
[Draft] Type-aware quantization

ragmani:
TBD


ragmani:
I try not to optimize type-aware quantization for non-portable tensors because it is less necessary compared to difficulty.

PR #10334, opened 2023-01-13
[one-cmds] Get right optimization list

mhs4670go:
This commit fixes the function to get the opt list correctly.

Related: https://github.com/Samsung/ONE/issues/10323#issuecomment-1381113666


seanshpark (approved):
LGTM

PR #10337, opened 2023-01-16
[one-cmds] Use no-verbose option for materials download

hseok-oh:
This commit updates `prepare_test_materials.sh` script to use `wget`'s `-nv` option when script tries to download materials.
This is introduced for easy error check on CI.

### DIFF
--- a/compiler/one-cmds/tests/prepare_test_materials.sh
+++ b/compiler/one-cmds/tests/prepare_test_materials.sh
@@ -21,34 +21,34 @@ pushd $SCRIPT_PATH > /dev/null
 
 if [[ ! -s "inception_v3.pb" ]]; then
     rm -rf inception_v3_2018_04_27.tgz
-    wget https://storage.googleapis.com/download.tensorflow.org/models/tflite/model_zoo/upload_20180427/inception_v3_2018_04_27.tgz
+    wget -nv https://storage.googleapis.com/download.tensorflow.org/models/tflite/model_zoo/upload_20180427/inception_v3_2018_04_27.tgz
     tar zxvf inception_v3_2018_04_27.tgz
 fi
 
 if [[ ! -s "while_3.pbtxt" ]]; then
     rm -rf while_3.zip
-    wget https://github.com/Samsung/ONE/files/5095630/while_3.zip
+    wget -nv https://github.com/Samsung/ONE/files/5095630/while_3.zip
     unzip while_3.zip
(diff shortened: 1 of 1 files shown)


hseok-oh:
@nnfw-bot test nnpackage-test

mhs4670go (dismissed):
LGTM

seanshpark:
@hseok-oh , why do you want to change to curl?

hseok-oh:
@seanshpark I want to use silent for easy error check on CI. I choose `curl` because its silent mode shows error message by using `-S` option if download fails, but `wget`'s silent mode does not.

seanshpark:
@hseok-oh , OK I got it.

But have you tested this behind proxy server without proxy exception?

seanshpark:
>  silent mode shows error message by using -S option if download fails

I would appreciate if you can add some comment for this in this change.

seanshpark:
In another way, can you add an option (default use wget) to use `curl` for downloading? and let CI use `curl` ?

Test with SE is done with `wget` with proxy settings and I don't want to change this for SE.
And personally I want to keep `wget` cause of our proxy settings and don't want to do another time loss of getting curl to work.

hseok-oh:
> @hseok-oh , OK I got it.
> 
> But have you tested this behind proxy server without proxy exception?
> 
> Test with SE is done with `wget` with proxy settings and I don't want to change this for SE. And personally I want to keep `wget` cause of our proxy settings and don't want to do another time loss of getting curl to work.

I don't know SE's setting. But `wget` and `curl` are support envrionment variable setting `http_proxy`, `https_proxy` to use proxy. (I've tested it on server room's server)

I checked `wget`'s supporting options again. I found option `-nv` (no-verbose) option as alternative. It seems to show less download log than default. If you want to use `wget`, we may use this option.

seanshpark:
> If you want to use wget, we may use this option.

it would be nice to use wget with this option :)

hseok-oh:
> > If you want to use wget, we may use this option.
> 
> it would be nice to use wget with this option :)

OK. I'll update.

hseok-oh:
@nnfw-bot test nnpackage-test

hseok-oh:
@seanshpark @mhs4670go I've updated. PTAL

mhs4670go (approved):

seanshpark (approved):
LGTM thank you!

Issue #10344, opened 2023-01-17
[one-cmds] Generate log file for test


mhs4670go:
## What 

Let's generates the log files for each test.

## Why

As of now, some tests don't generate the log of its test.

https://github.com/Samsung/ONE/blob/a49eb91fe77e305068575afb65eea3bedf5d9503/compiler/one-cmds/tests/onecc_001.test#L34

But, it forces us to run again without `/dev/null` part if the test failed. And even though the test on the CI failed, we can't know about the exact reason.

So, let's make them generate the log file like below.

https://github.com/Samsung/ONE/blob/a49eb91fe77e305068575afb65eea3bedf5d9503/compiler/one-cmds/tests/onecc_030.test#L41

@seanshpark What's your take on this? Is it okay to proceed this task?

seanshpark:
> it forces us to run again without /dev/null part if the test failed

yes, this is current problem :(

> Is it okay to proceed this task?

I think creating logs for all the tests looks OK :)

PR #10347, opened 2023-01-17
[one-cmds] Revisit one-cmds test

mhs4670go:
This commit revisits one-cmds test as follows.
- make it generate log file
- remove previous output and log file before the test

Related: #10344 


seanshpark (approved):
LGTM

Issue #10359, opened 2023-01-19
[luci] Fuse PReLU from ONNX model


seanshpark:
PReLU operator from ONNX is exploded to a small graph that would be better be fused to `CirclePRelu`.
Let's add a pass to do so.

`res/PyTorchExamples/examples/PReLU` becomes like this when converted to circle.

![image](https://user-images.githubusercontent.com/4616940/213358046-01c5eacf-ea71-41d8-824a-a6323c306914.png)

seanshpark:
from internal repo 22 issue 177

seanshpark:
When there's a Conv2d as input of `PReLU`, network has `Reshape` that we should consider.


```py
# model
class net_Conv2dPReLU(nn.Module):
    def __init__(self):
        super().__init__()
        self.op1 = nn.Conv2d(1, 1, 1)
        self.op2 = nn.PReLU()

    def forward(self, input):
        return self.op2(self.op1(input))

_model_ = net_Conv2dPReLU()
```

becomes this in .circle
![image](https://user-images.githubusercontent.com/4616940/213359922-74f889d9-fb83-40ac-b1e8-07c358887163.png)

seanshpark:
Found out that our model uses torch `Conv1d` which input is 3D not 4D.

seanshpark:
With draft and below cfg
```ini
[onecc]
one-import-onnx=True
one-optimize=True

[one-import-onnx]
input_path=PReLUwConv2d.onnx
output_path=PReLUwConv2d.circle

[one-optimize]
input_path=PReLUwConv2d.circle
output_path=PReLUwConv2d.opt.circle

convert_nchw_to_nhwc=True
substitute_transpose_to_reshape=True 
remove_redundant_reshape=True
remove_unnecessary_reshape=True
forward_reshape_to_unaryop=True
```
gives 
![image](https://user-images.githubusercontent.com/4616940/213382985-5dd19275-71ec-477e-9a25-336b90f7addf.png)

put `Conv2D` as input, `Reshape` as output, we can use this graph as basic form to `PReLU`

seanshpark:
First `Mul` is slop for minus input, and second `Mul` is always `0.5` to half the input that was doubled by `X - ABS(X)`; when `X > 0`, this will be 0, `X < 0` then `2 x X`

seanshpark:
Waiting fro the model from our customer for validation...

seanshpark:
Now working.

PR #10361, opened 2023-01-19
DRAFT import ONNX PReLU

seanshpark:
on-going draft to import ONNX PReLU.


seanshpark:
for #10359

seanshpark:
done

PR #10362, opened 2023-01-20
[res] extract onnx_model_path

seanshpark:
This will extract onnx_model_path from ptem.py as common path for onnx model.


seanshpark:
related #10359 from #10361

jinevening (approved):
LGTM

PR #10363, opened 2023-01-20
[res] PyTorchExamples of PReLU with Conv

seanshpark:
This will introduce PyTorchExamples of PReLU with Conv.


seanshpark:
for #10359 from #10361

jinevening (approved):
LGTM

BTW, our customer said that the model will be changed to use Conv2d..

seanshpark:
> BTW, our customer said that the model will be changed to use Conv2d..

Ah, yes, I'll check only with Conv2D with luci :)

PR #10365, opened 2023-01-20
[luci/pass] NodeFiller with_args_of helper

seanshpark:
This will add with_args_of helper method without commutative for Sub/Div.


seanshpark:
for #10359 from #10361

jinevening (approved):
LGTM

PR #10368, opened 2023-01-20
[res] Add shape inference for PyTorchExamples

seanshpark:
This will add shape inference to show shapes in PyTorchExamples result ONNX files.


seanshpark:
related #10359 from #10361

jinevening (approved):
LGTM

PR #10369, opened 2023-01-20
[visq-unittest] Remove redundant log

jinevening:
This removes redundant print in the test code.


seanshpark (approved):
LGTM

PR #10370, opened 2023-01-20
[luci/pass] Implement FusePReluPass

seanshpark:
This will implement FusePReluPass to fuse certain subgraph to CirclePRelu Op.

### DIFF
--- a/compiler/luci/pass/src/FusePReluPass.cpp
+++ b/compiler/luci/pass/src/FusePReluPass.cpp
@@ -64,9 +64,6 @@ public:
     _add_ofm = candidate;
   }
 
-private:
-  bool match();
-
 public:
   bool matched();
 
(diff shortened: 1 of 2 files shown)


seanshpark:
for #10359 from #10361

compiler/luci/pass/src/FusePReluPass.cpp
```diff
@@ -15,17 +15,173 @@
...
+  luci::CircleAbs *_abs = nullptr;
+  luci::CircleSub *_sub = nullptr;
+  luci::CircleMul *_mul_alpha = nullptr;
+  luci::CircleMul *_mul_half = nullptr;
+  luci::CircleAdd *_add_ofm = nullptr;
+  luci::CircleConst *_const_alpha = nullptr;
+  luci::CircleConst *_const_half = nullptr;
+
+private:
+  bool _matched = false;
```
jinevening, line 85: Is this variable necessary?

compiler/luci/pass/src/FusePReluPass.cpp
```diff
@@ -15,17 +15,173 @@
...
+};
+
+#define CHECK_OR_FALSE(condition) \
+  if (not(condition))             \
+    return false;
+
+bool PReluPattern::match()
```
jinevening, line 92: CircleMul, CircleSub, CircleAdd can have fused act function.

They have to be None.

compiler/luci/pass/src/FusePReluPass.cpp
seanshpark: This variable came from InstNormPass :)
I'll remove this.

jinevening (approved):
LGTM

PR #10372, opened 2023-01-20
[luci/pass] Enable FusePReluPass

seanshpark:
This will enable optimizer with FusePReluPass.


seanshpark:
for #10359 from #10361

jinevening (approved):
LGTM

PR #10377, opened 2023-01-27
[circle2circle] Add fuse_prelu option

seanshpark:
This will add fuse_prelu option to fuse some sort of subgraph to CirclePRelu Op.

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -110,7 +110,7 @@ int entry(int argc, char **argv)
              "when the impact is known to be acceptable.");
   add_switch(arser, "--fuse_preactivation_batchnorm",
              "This will fuse BatchNorm operators of pre-activations to Convolution operator");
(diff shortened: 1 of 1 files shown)


seanshpark:
for #10359 from #10361

compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -110,6 +110,7 @@ int entry(int argc, char **argv)
              "when the impact is known to be acceptable.");
   add_switch(arser, "--fuse_preactivation_batchnorm",
              "This will fuse BatchNorm operators of pre-activations to Convolution operator");
+  add_switch(arser, "--fuse_prelu", "This will fuse sub-graph to PReLU that is equivalent");
```
jinevening, line 113: sub-graph might be a little bit confusing. It is used in tflite or circle schema. How about using the same words with `--fuse_instnorm`? They behave similarly.

```suggestion
  add_switch(arser, "--fuse_prelu", "This will fuse operators to PReLU operator");
```

jinevening (approved):
LGTM

PR #9973, opened 2022-11-02
DRAFT build/test on Ubuntu22.04

seanshpark:
on-going draft to build/test on Ubuntu22.04

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #9962, opened 2022-10-31
[CFE] Ubuntu and Python upgrade


seanshpark:
Out default platform Ubuntu 18.04 will end support Apr 2023, less than a year left.
And we now have Ubuntu 22.04, it's time to support new platform.
But also support old 18.04 for a while...

Let's see how versions exist;

Ubuntu | default | support upto| with deadsnakes/ppa | tentative
--- | --- | --- | ---- | ---
Ubuntu 18.04 | 3.6 | 3.8 | 3.11 | 3.8 (as-is)
Ubuntu 20.04 | 3.8 | 3.9 | 3.12 | 3.8 (as-is)
Ubuntu 22.04 | 3.10 | 3.11 | 3.12 | 3.10 (default)

- latest python version is 3.11

Purpose of this task
- decide python version (or versions)
- use single python version if possible (currently 3.8)
   - Ubuntu officially doesn't support all version --> test not confirmed
- to reduce code complexity
   - --> 22.04 use default 3.10
   - --> 18.04 / 20.04 go as-is
   - from draft, management is possible

seanshpark:
Reference
- https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa

seanshpark:
From Ubuntu vs Python version table...
- we can't go with one python version, unless we use deadsnakes/ppa

When using different python versions...
- cmake codes needs modification to support different versions by different platform
- some scripts may also need modification

seanshpark:
### Issues with build/test in Ubuntu 22.04

- TBA...

seanshpark:
### Issues with build/test in Ubuntu 22.04 : `one-prepare-venv` 실행

`python3.8` not found --> let's use `python3` for the moment... 

I'm behind corporate firewall and using proxy

> Could not fetch URL https://pypi.org/simple/tensorflow-cpu/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/tensorflow-cpu/ (Caused by SSLError(SSLZeroReturnError(6, 'TLS/SSL connection has been closed (EOF) (_ssl.c:997)'))) - skipping

--> ?

mhs4670go:
Just for others information, 
> ### Issues with build/test in Ubuntu 22.04 : `one-prepare-venv` run

We need to write dedicated venv script for 22.04 named `one-prepare-venv_2204`(or `one-prepare-venv-jammy`..?).

[...]

PR #10383, opened 2023-01-30
[one-cmds] Revise package versions

seanshpark:
This will revise package versions for Ubuntu22.04 venv.


seanshpark:
for #9962 , from draft #9973

mhs4670go (approved):
LGTM

PR #10391, opened 2023-01-31
[cmn-arts] Prepare venv for jammy

seanshpark:
This will prepare additional venv for jammy with TF2.10.1.


seanshpark:
for #9962 , from draft #9973

mhs4670go (approved):
LGTM

PR #10394, opened 2023-01-31
[one-cmds] Install one-prepare-venv per platform

seanshpark:
This will revise to install one-prepare-venv per platform.
- one-prepare-venv.u2204 is for jammy


seanshpark:
for #9962 , from draft #9973

Issue #10396, opened 2023-01-31
[Compiler] PR check with Ubuntu22.04


seanshpark:
Let's add additional PR check for Ubuntu22.04.
Ubuntu18.04 and Ubuntu20.04 uses same version of python and TensorFlow.
But for is different for Ubuntu22.04, so CI check seems necessary for our PRs.

seanshpark:
@hseok-oh , could you please enable for Ubuntu22.04(jammy) build for PRs ?
I think only debug seems OK for the mean time.
And just run as existing test on Ubuntu22.04 docker image would do.

seanshpark:
for #9962

seanshpark:
I've prepared #10395 and #10397 and will land the codes without the PR checks of U22.04 as this is not prepared yet. You can check these two changes for testing.
After PR check is ready, I'll work on with remaining changes.

hseok-oh:
Build failed 

```
17:59:51  /opt/jenkins_agent/workspace/nnfw/master/pr-jenkins-test/compiler/nnc/passes/optimizations/ConstantFoldTranspose.cpp:65:21: error: loop variable 'match' creates a copy from type 'const std::pair<mir::Operation*, mir::Operation*>' [-Werror=range-loop-construct]
17:59:51     65 |     for (const auto match : matches)
17:59:51        |                     ^~~~~
17:59:51  /opt/jenkins_agent/workspace/nnfw/master/pr-jenkins-test/compiler/nnc/passes/optimizations/ConstantFoldTranspose.cpp:65:21: note: use reference type to prevent copying
17:59:51     65 |     for (const auto match : matches)
17:59:51        |                     ^~~~~
17:59:51        |                     &
17:59:52  cc1plus: all warnings being treated as errors
```

seanshpark:
my local
```
$ g++ --version
g++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
anyway, I'll post a patch :)

hseok-oh:
>can you share g++ version of CI ?

```
$ docker run -it --rm nnfw/one-devtools:jammy gcc --version
gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
[...]

PR #10398, opened 2023-01-31
[nnc/passes] Make compiler happy

seanshpark:
This will fix compiler error with range-loop-construct.


seanshpark:
for https://github.com/Samsung/ONE/issues/10396#issuecomment-1410014096

mhs4670go (approved):
LGTM

PR #10410, opened 2023-02-01
[onert-micro] Introduce UnidirectionalSequenceLSTM kernel

SlavikMIPT:
This commit introduces UnidirectionalSequenceLSTM floating point kernel.


onert-micro/luci-interpreter/src/kernels/UnidirectionalSequenceLSTM.cpp
```diff
@@ -0,0 +1,884 @@
...
+namespace kernels
+{
+namespace lstm
+{
+namespace
+{
+
+using namespace tflite;
+
+void UpdateLstmCellFloat(int n_batch, int n_cell, float *cell_state, const float *input_gate,
```
BalyshevArtem, line 35: Optional (maybe for future refactoring): How about all functions from this `lstm` namespace move to another helper file (smth like FloatLSTMEvalKernel), due to it is very difficult to read this file, and now it is only for float and when we will add integer (and maybe hybrid part) it will be very huge file.
But now it is okay leave it like this, but add todo note please

BalyshevArtem: Or maybe create PAL for mcu and move these functions to this file.

onert-micro/luci-interpreter/src/kernels/UnidirectionalSequenceLSTM.cpp
```diff
@@ -0,0 +1,884 @@
...
+{
+namespace kernels
+{
+namespace lstm
+{
+namespace
+{
+
+using namespace tflite;
+
```
SlavikMIPT, line 34:
```suggestion

// TODO: move this functions to another helper file (PAL, templates ?)
```

BalyshevArtem (approved):
LGTM

PR #10421, opened 2023-02-02
[onert] Fix using after forward

hseok-oh:
This commit resolves warning: using after std::forward.


hseok-oh:
I confirmed internally it resolves warning.

PR #10435, opened 2023-02-07
[circle-opselector] Set proper output

mhs4670go:
This commit fixes the bug to make it set valid output.

### DIFF
--- a/compiler/circle-opselector/src/OpSelector.cpp
+++ b/compiler/circle-opselector/src/OpSelector.cpp
@@ -138,35 +138,41 @@ std::unique_ptr<loco::Graph> make_graph(const std::vector<const luci::CircleNode
   for (const auto &n : nodes)
   {
     auto outputs = loco::succs(n);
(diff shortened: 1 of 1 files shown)


seanshpark:
@mhs4670go , can you please add a comment about the actual problem or a link to some issue?

mhs4670go:
> @mhs4670go , can you please add a comment about the actual problem or a link to some issue?

In a nutshell, there was a problem where the output shapes were set.

This bug had not been found because the circle exporter runs shape inference before graph export.

seanshpark (approved):
LGTM

seanshpark:
for the record;
- if the node was used as an input to multiple nodes, output was created that much but actually only one for the node was necessary
- there was a typo in variable name that node was set instead of setting the shape
  - fixed code: `output_shape->dim(i).set(circle_output->dim(i).value());`

mhs4670go:

seanshpark:

PR #10437, opened 2023-02-07
[visq] Fix single quote pair

seanshpark:
This will fix single quite pair count that seems a typo.


mhs4670go:
```bash
#!/usr/bin/env bash
''''command                        # '''
''''exec ${PYTHON_PATH} "$0" "$@"  # '''
```

FYI, above script runs as follows.
1. run this script by bash.
2. `''''` means `''` + `''` and `''` means empty string. So, `''''command` is same with just `command`.
3. run this script by python.
4. `'''' # '''` is ignored since it is docstring in the python.

jinevening (approved):
LGTM Thank you

seanshpark:

PR #10444, opened 2023-02-07
[luci-value-test] Temporary disable YUV_TO_RGB_U8_000

seanshpark:
This will temporary disable YUV_TO_RGB_U8_000 as sometimes fails in jammy and this makes CI unhappy.


mhs4670go (approved):
LGTM

Issue #10438, opened 2023-02-07
[luci_value_210_test] sometimes fails with YUV_TO_RGB_U8_000


seanshpark:
Sometimes fails with this test

```
18:10:59  114/128 Test #114: luci_value_210_test ........................***Failed  170.14 sec
18:10:59  FAILED
18:10:59  - YUV_TO_RGB_U8_000
18:10:59  
```

seanshpark:
from #10437

seanshpark:
local test (Ubuntu22.04) also sometimes fail with `YUV_TO_RGB_U8_000`

seanshpark:
maybe rounding error ?
```
   [ 31  73 168]
```
vs
```
   [ 31  72 168]
```

seanshpark:
wait for a week if problem occurs again.

seanshpark:
not reproduced

PR #10445, opened 2023-02-08
[luci-value-test] Revise YUV_TO_RGB_U8_000 to use tolerance

seanshpark:
This will revise YUV_TO_RGB_U8_000 to test with tolerance
- revise value compare with tolerance for integer too


seanshpark:
for #10438

mhs4670go (approved):
LGTM

PR #10449, opened 2023-02-08
[res] Add FullyConnected_009 recipe

mhs4670go:
This commit adds FullyConnected_009 recipe.

The graph looks like below.

image

And, the recipe is introduced to fix the bug of _circle-opselector_.

As of now, the selected graph by _circle-opselector_ has a bug where the output shape is wrong.

image

Related: #10450 


seanshpark (approved):
LGTM

PR #10451, opened 2023-02-08
[luci-value-test] Dump diff when error

seanshpark:
This will revise verify script to dump tensors when there is error for better debugging.


seanshpark:
related #10438

mhs4670go (approved):
LGTM

Issue #10450, opened 2023-02-08
[circle-opselector] Generate wrong output shape graph


mhs4670go:
## What

There is a bug where the output shape of the graph has the wrong values.

### _res/TensorFlowLiteRecipes/FullyConnected_009/test.recipe_

![image](https://user-images.githubusercontent.com/24587354/217441786-6eb8d359-a09b-4f9a-b0fa-685bb8a0bd12.png)

```bash
$ build/compiler/circle-opselector/circle-opselector --by_id 0 \
    build/compiler/common-artifacts/FullyConnected_009.circle selected.circle
```

### _selected.circle_

![image](https://user-images.githubusercontent.com/24587354/217442170-68fd63b2-95e1-40c3-9cc0-e2bcefa31aa0.png)

PR #10452, opened 2023-02-08
[luci] Set keep_num_dims of CircleFullyConnected when cloning

mhs4670go:
This commit sets keep_num_dims attribute of CircleFullyConnected when cloning the node.

Related: #10450 


seanshpark (approved):
LGTM thank you!

PR #10454, opened 2023-02-08
[onert] Use forwarding reference in range-based for loop

ragmani:
This commit changes range-based for loops to use forwarding reference(universal reference).


ragmani:
It is safe, and in fact, preferable in generic code, to use deduction to forwarding reference. (https://en.cppreference.com/w/cpp/language/range-for#:~:text=It%20is%20safe%2C%20and%20in%20fact%2C%20preferable%20in%20generic%20code%2C%20to%20use%20deduction%20to%20forwarding%20reference%2C)

And I found a case where `auto&` doesn't work if you cannot bind a non-const lvalue to a rvalue. Please see https://stackoverflow.com/questions/29320575/whats-the-difference-between-and-in-a-range-based-for-loop

hseok-oh (approved):
LGTM

PR #10455, opened 2023-02-08
[onert] Revisit `Executors`

ragmani:
This commit revisits `Executors`
  - Separate initializing EdgeTensors and Layers related to type-aware quantization
  - Remove unnecessary implementation
  - Fix creating type-aware quantization layer
  - Fix io tensors of nnpkg for type-aware quantization to manage inputs and outputs separately.

ONE-DCO-1.0-Signed-off-by: ragmani <ragmani0216@gmail.com>

---
For https://github.com/Samsung/ONE/issues/10002
Draft https://github.com/Samsung/ONE/pull/10329

PR #10461, opened 2023-02-13
[onert] Seperate creating things for type-aware quant of nnpkg io

ragmani:
This commit seperates what `Executors` creates for type-aware quantization of nnpkg io.

---
For https://github.com/Samsung/ONE/issues/10002
Draft https://github.com/Samsung/ONE/pull/10329

### DIFF
--- a/runtime/onert/core/src/exec/Executors.cc
+++ b/runtime/onert/core/src/exec/Executors.cc
@@ -18,6 +18,43 @@
 
 #include "../backend/builtin/IOTensor.h"
 
+namespace
+{
+
+using namespace onert;
+
+int32_t find_input_index(const std::vector<ir::IODesc> &pkg_inputs,
+                         const ir::ModelIndex &model_index, const ir::SubgraphIndex &subg_index,
+                         const ir::IOIndex &io_index)
+{
+  for (size_t i = 0; i < pkg_inputs.size(); i++)
+  {
+    auto &input_desc = pkg_inputs[i];
+    if ((std::get<ir::ModelIndex>(input_desc) == model_index) &&
+        (std::get<ir::SubgraphIndex>(input_desc) == subg_index) &&
+        (std::get<ir::IOIndex>(input_desc) == io_index))
+      return static_cast<int32_t>(i);
+  }
+  return -1;
+}
+
+int32_t find_output_index(const std::vector<ir::IODesc> &pkg_outputs,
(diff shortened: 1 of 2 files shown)


ragmani:
This PR is a part of https://github.com/Samsung/ONE/pull/10455

hseok-oh:
Conceptually, `Executors` and each `Executor` is expected as read-only on `execute()` operation because it is result of compilation (like binary file). But after introducing type-aware quantization, this concept is broken. And it became thread-unsafe. If we can do it statically, we need to move this to compile phase. Otherwise, we need to introduce execution context and use context instance for each execution (inference).

Already concept is broken, I'll approve this PR. But we need to revive concept again later.

hseok-oh (approved):

ragmani:
> Conceptually, `Executors` and each `Executor` is expected as read-only on `execute()` operation because it is result of compilation (like binary file). But after introducing type-aware quantization, this concept is broken. And it became thread-unsafe. If we can do it statically, we need to move this to compile phase. Otherwise, we need to introduce execution context and use context instance for each execution (inference).

I guess there is no problem to move edge tensors for type-aware quantization between executors to compile phase. But, for   tensors for nnpkg io, there is a thing we need to consider.
- `onert` doesn't know type and information of nnpkg io that user may trying to change in compile phase.

The issues I think related to this consideration are as follows:
1. `onert` cannot change the type and information of created tensors now. How to apply the type and information that user wants to change to tensors of nnpkg io? 
2. The tensors and layers for type-aware quantization of nnpkg io need to be created if user want to change types of nnpkg io. In compile phase, Changing types is not known. How should `onert` deal with those tensors and layers in `Executors`?

And I have a question.
`onert` has already supported dynamic tensor. So user can change buffer and its size in execution phase. Doesn't it let `Executor` becomes thread-unsafe?

hseok-oh:
>And I have a question.
`onert` has already supported dynamic tensor. So user can change buffer and its size in execution phase. Doesn't it let `Executor` becomes thread-unsafe?

It is using workaround by mutex.

https://github.com/Samsung/ONE/blob/21252a03fe7f483bd775df53d645c0047f2a7d9a/runtime/onert/core/src/exec/ExecutorBase.cc#L50-L57

https://github.com/Samsung/ONE/blob/21252a03fe7f483bd775df53d645c0047f2a7d9a/runtime/onert/core/src/exec/ExecutorBase.cc#L100-L106

`Executors` can uses same workaround for thread-safe.

Issue #10381, opened 2023-01-27
[one-optimize] Make a basic option sets for safe optimization (O1)


jinevening:
## What

~Let's make pre-defined option sets for optimization.~

~We may group some options by level of optimizations, e.g., `-O`, `-O2`, or by source models, e.g., `--source onnx`, `--source tflite`.~

~Or we can use both. `--source onnx -O`.~

Updated 2023/02/17: Let's make a basic option sets for safe optimizations (O1), which do not change the execution result of models and finish within a reasonable time.

## Why

To make our customers easily select proper options. Note that our customers directly requested this feature.

We have discussed this topic for a long time, and it's time to proceed :)

Previous discussions
- https://github.com/Samsung/ONE/issues/5784
- https://github.com/Samsung/ONE/issues/7350
- #9191
- #9369
- ...

CC @seanshpark @mhs4670go

mhs4670go:
Please refer to https://github.com/Samsung/ONE/issues/7513. As of now we have the infra that makes it possible to provide pre-defined options. Maybe -OONNX, -OTFL option could be one of candidates.

jinevening:
@mhs4670go As discussed offline, #7513 supports functionality of `onecc` (previously `one-build`) to import another config file using CLI. I hope we can set default options inside the cfg file. For example,

```
[one-optimize]
input_path=...
output_path=...
source=onnx
```

jinevening:
I'm considering pre-defined option sets for the below cases.

By the level of optimization,
`O1`: This option sets will include most optimization options which do not change the result of the model.
- We'll need to measure the compile time when applying `O1` to our target models. If it is too time-consuming, we may introduce `O2` and move some options from `O1` to `O2`.

By the source of model (`source` attribute),

1. `onnx`: `O1` + `convert_nchw_to_nhwc` related options
2. `tf`: `O1`
3. `tflite`: `O1`

@mhs4670go @seanshpark  @lemmaa  Could you give your opinions?

seanshpark:
> Could you give your opinions?

(background: we cannot be perfect in the first place)
I think we can go with your proposal.
The problem: when something is wrong with some particular model but works OK with most...
Somehow we have to make a change, but what happens to most models working OK?

jinevening:
> Somehow we have to make a change, but what happens to most models working OK?

I'm considering option sets that work for all of the current commercial models (I guess that is our best for now). If a new model causes an issue, let's resolve that at that time :)

[...]

PR #10463, opened 2023-02-14
DRAFT: Add O1 option

jinevening:
On going draft to support O1 option.

---
Related to: https://github.com/Samsung/ONE/issues/10381


compiler/one-cmds/onelib/constant.py
```diff
@@ -17,8 +17,65 @@
...
+        'substitute_transpose_to_reshape',
+        'forward_reshape_to_unaryop',
+        'forward_transpose_op',
+        'replace_non_const_fc_with_batch_matmul',  # For quantization
+    )
+
     OPTIMIZATION_OPTS = (
         # (OPTION_NAME, HELP_MESSAGE)
+        ('O1',
+         'A basic set of optimization options. List of options are as follows. ' + str(O1)
```
jinevening, line 77: Help message.

```
arguments for optimization:
  -i INPUT_PATH, --input_path INPUT_PATH
                        full filepath of the input file
  -o OUTPUT_PATH, --output_path OUTPUT_PATH
                        full filepath of the output file
  --O1                  A basic set of optimization options. List of options
                        are as follows. ('fold_add_v2', 'fold_cast',
                        'fold_densify', 'fold_dequantize', 'fold_dwconv',
                        'fold_fully_connected', 'fold_gather',
                        'fold_sparse_to_dense', 'fuse_add_with_tconv',
                        'fuse_add_with_fully_connected',
                        'fuse_batchnorm_with_conv',
                        'fuse_batchnorm_with_dwconv',
                        'fuse_batchnorm_with_tconv',
                        'fuse_activation_function', 'fuse_instnorm',
                        'fuse_prelu', 'fuse_mean_with_mean',
                        'fuse_transpose_with_mean',
                        'transform_min_max_to_relu6',
                        'transform_min_relu_to_relu6',
                        'remove_redundant_reshape',
                        'remove_redundant_transpose',
                        'remove_unnecessary_reshape',
                        'remove_unnecessary_slice',
                        'remove_unnecessary_strided_slice',
                        'remove_unnecessary_split', 'resolve_customop_add',
                        'resolve_customop_batchmatmul',
                        'resolve_customop_matmul',
                        'resolve_customop_max_pool_with_argmax',
                        'resolve_customop_splitv',
                        'substitute_pack_to_reshape',
                        'substitute_padv2_to_pad',
                        'substitute_splitv_to_split',
                        'substitute_squeeze_to_reshape',
                        'substitute_strided_slice_to_reshape',
                        'substitute_transpose_to_reshape',
                        'forward_reshape_to_unaryop', 'forward_transpose_op',
                        'replace_non_const_fc_with_batch_matmul')
...
```

seanshpark:
About the name `O1`... 
- Will there be `O2`, `O3` ... ?
- What about for `ONNX` ? or for `TRIX` ? or some new source or new NPU device ?
- I was hoping to have some consistency with `-Oxxxx` for these.

seanshpark:
@jinevening , what is the most point do you want to get from the review?

jinevening:
> Will there be O2, O3 ... ?
> What about for ONNX ? or for TRIX ? or some new source or new NPU device ?

I think O1 is enough. O2 and O3 are not necessary at least for now. The name was just for leaving a room for extension. Of course it can be changed.
(FYI, the name of the basic option sets (`O1`) was mentioned here https://github.com/Samsung/ONE/issues/10381#issuecomment-1413312577.)

O1 is not dependent on source frameworks or backends. It is just a set of basic optimization options. For onnx model, I think users can give `--O1` and other options (ex: `convert_nchw_to_nhwc`) if necessary. FYI, we found that `convert_nchw_to_nhwc` is not beneficial for recent transformer models.

> -OBASIC : basic optimization
-OONNX : for ONNX
-OTRIX : for TRIX
-OSPEED : for latency performance (something like all the fuse stuff not in BASIC)
-OSIZE : for smaller model size but may hurt correctness (sparsify + compress + ...)

About the naming, it is a bit confusing to me what comes next to `-O`. `BASIC`, `SPEED`, and `SIZE` are optimization policies, `ONNX` is a source framework, `TRIX` is a backend. I think it is better to use `O` just for optimization policies, while making a new option for another category, for example, --source onnx, --backend trix.

jinevening:
> what is the most important or uncertain point do you want to get from the review?

Any comment would be helpful. It can be about the naming (as you suggested), individual options, implementation, etc. Beyond them, the main purpose of the review is to share the status with frontend compiler team :)

seanshpark:
> I think it is better to use O just for optimization policies

This means, there is only `-O1` of `-O` for now? (maybe there can be `-O2` and so on later)

> for example, --source onnx, --backend trix.

This is new category. As of now I don't know how to accept this.

jinevening:
> This means, there is only -O1 of -O for now? (maybe there can be -O2 and so on later)

Yes (except that the option is `--O1`, not `-O1`). The name can be changed to `--OBASIC` as you suggested. I'm ok with both names.

> This is new category. As of now I don't know how to accept this.

I understand. Me too it is a bit difficult to make option sets for specific source frameworks or backends (it depends on models..). I'd like to narrow down the task to make a default options set for safe optimizations, which does not change the execution results and is good in general.

compiler/one-cmds/onelib/constant.py
mhs4670go: I think we don't have to list the options that will be applied in the help message, which seems too verbose. Maybe there are two options. The one is just not to show the detail list and the other is to show them on the man page only.

jinevening: > the other is to show them on the man page only.

This looks good to me. I'll update manpage.

Issue #10425, opened 2023-02-03
[CFE] Revisit docker script for nncc build


mhs4670go:
## What

Let's revisit the script where _nncc_ is built inside the docker container.

Related: https://github.com/Samsung/ONE/issues/10389#issuecomment-1414896045

## Why

https://github.com/Samsung/ONE/blob/325f167873ada19434950d97291071f64501c519/infra/scripts/docker_build_nncc.sh#L74-L75

In the script, _debuild_ is run with `--buildinfo-option=-ubuild --changes-option=-ubuild` options.

```bash
# --buildinfo-option=-ubuild
-uupload-files-dir
              Look for the files to be uploaded in upload-files-dir rather than ..  (dpkg-genbuildinfo needs to find these files so that it
              can include their sizes and checksums in the .buildinfo file).

# --changes-option=-ubuild
-uupload-files-dir
              Look  for  the files to be uploaded in upload-files-dir rather than ..  (dpkg-genchanges needs to find these files so that it
              can include their sizes and checksums in the .changes file).
```

These options were introduced for workaround.

_debuild_ creates its artifacts on the parent directory. But `./nncc docker-run` mounts only _ONE_ project directory, which is why current workaround is needed. The workaround makes the _debuild_ change the artifacts' path inside the project directory.

But, default _debuild_ behavior of above options has changed a bit from ubuntu 22.04. And it is seems hard to make it do the same behavior with the one from previous ubuntu version. Every time new version comes out, the behavior could change again, which would be a big costs. It'd better not to use the options.

So, let's use just `docker run` command instead of `./nncc docker-run` one. And, just not to use the options.

hseok-oh:
I have plan to remove docker_xxx.sh build script, and use docker container as jenkins build agent directly. Runtime build scripts are already removed, and compiler also uses this feature. Remain is only compiler packaging script. I'm testing to resolve permission issue, but it needs some time to resolve issue because jenkins docker agent has difficult issue to use root user.

mhs4670go:
@hseok-oh Oh, it would be very helpful! Then, should I just put this issue on hold? Or just go to proceed?

hseok-oh:
> @hseok-oh Oh, it would be very helpful! Then, should I just put this issue on hold? Or just go to proceed?

If this is not emergency issue, please wait. IMHO, it may takes 1~2 week to resolve permission issue.

hseok-oh:
[...]

PR #10466, opened 2023-02-15
[debian] Revisit rules

mhs4670go:
This commit revisits rules to change the destdir.

This change follows the CI change. At first, `destdir` should be given because default `destdir` - parent directory of working directory - cannot be accessed by docker container for permission issue. But for now, the restriction has been resolved so we don't have to override `destdir` anymore.

NOTE. This commit should be merged at the same time with the [CI change](https://github.sec.samsung.net/RS7-RuntimeNTools/nnfw_ci/pull/595).

Related: https://github.com/Samsung/ONE/issues/10425#issuecomment-1415340488


hseok-oh (approved):
LGTM

PR #10471, opened 2023-02-17
[record-minmax] Set unknown batch to one

jinevening:
This sets unknown batch size to one.

### DIFF
--- a/compiler/record-minmax/src/RecordMinMax.cpp
+++ b/compiler/record-minmax/src/RecordMinMax.cpp
@@ -215,6 +215,18 @@ void RecordMinMax::initialize(const std::string &input_model_path)
     throw std::runtime_error("Failed to load '" + input_model_path + "'");
   }
 
+  // Set unknown batch to 1
+  auto input_nodes = loco::input_nodes(_module->graph());
+  for (auto input : input_nodes)
+  {
+    auto input_node = loco::must_cast<luci::CircleInput *>(input);
+
+    // If input is rank 4, we assume the input is NHWC or NCHW
+    // If batch size (N) is unknown, we assume a single batch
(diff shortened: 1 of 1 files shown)


jinevening:
To reviewers:

This will allow quantization of models with unknown batch. I made this feature because models with unknown batch sizes keep coming from our customers.

@seanshpark I checked that mobilenet edgetpu model is quantized well with this PR.

compiler/record-minmax/src/RecordMinMax.cpp
```diff
@@ -215,6 +215,18 @@ void RecordMinMax::initialize(const std::string &input_model_path)
...
+  // Set unknown batch to 1
+  auto input_nodes = loco::input_nodes(_module->graph());
+  for (auto input : input_nodes)
+  {
+    auto input_node = loco::must_cast<luci::CircleInput *>(input);
+
+    // If input is rank 4, we assume the input is NHWC or NCHW
+    // If batch size (N) is unknown, we assume a single batch
+    if (input_node->rank() == 4 and !input_node->dim(0).known())
+      input_node->dim(0).set(1);
```
seanshpark, line 227: @jinevening , I disagree on this patch. We cannot just change the model like this.

seanshpark: Alternative way that I can think is provide another tool, not in record-minmax, that will fix input shape with batch unknown to batch 1.

jinevening: Hmm.. I see. It would be better to give an option to users.

> provide another too

Maybe `one-optimize`?

seanshpark: > Maybe one-optimize?

Yes, maybe... with `--fix-unknown-batch` or something like it.

PR #10472, opened 2023-02-17
[one-cmds] Add one-optimize test for O1

jinevening:
This adds one-optimize test for O1.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/10381
Draft PR: https://github.com/Samsung/ONE/pull/10463

PR #10473, opened 2023-02-17
[one-cmds] Add onecc tests for O1

jinevening:
This adds onecc tests for O1.

---
Related to: https://github.com/Samsung/ONE/issues/10381
Draft PR: https://github.com/Samsung/ONE/pull/10463


compiler/one-cmds/tests/onecc_044.test
```diff
@@ -0,0 +1,53 @@
...
+
+# run test
+onecc -C ${configfile} > ${filename}.log 2>&1
+
+if [[ ! -s "${outputfile}" ]]; then
+  trap_err_onexit
+fi
+
+# Check convert_nchw_to_nhwc is included
+if ! grep -q "convert_nchw_to_nhwc" <<< $(cat ${outputfile}.log); then
```
seanshpark, line 44: Q) what is `<<<` . I've seen one or two but not three.

compiler/one-cmds/tests/onecc_044.test
```diff
@@ -0,0 +1,53 @@
...
+  trap_err_onexit
+fi
+
+# Check fold_add_v2 is excluded
+if grep -q "fold_add_v2" <<< $(cat ${outputfile}.log); then
```
mhs4670go, line 49: Maybe this would work, too.
```suggestion
if ! grep -q "convert_nchw_to_nhwc" ${outputfile}.log; then
  trap_err_onexit
fi

# Check fold_add_v2 is excluded
if grep -q "fold_add_v2" ${outputfile}.log; then
```

compiler/one-cmds/tests/onecc_043.test
```diff
@@ -0,0 +1,43 @@
...
+rm -f ${filename}.log
+rm -rf ${outputfile}
+
+# run test
+onecc -C ${configfile} > ${filename}.log 2>&1
+
+if [[ ! -s "${outputfile}" ]]; then
+  trap_err_onexit
+fi
+
```
mhs4670go, line 42: Maybe we need to check the O1 option. #10472 can apply this codes as well. 

```suggestion
declare -a O1_OPTS=(
  # Constant folding
  "fold_add_v2" "fold_cast"
  "fold_densify" "fold_dequantize"
  "fold_dwconv" "fold_fully_connected"
  "fold_gather" "fold_sparse_to_dense"

  # Operator fusion
  "fuse_add_with_tconv" "fuse_add_with_fully_connected"
  "fuse_batchnorm_with_conv" "fuse_batchnorm_with_dwconv"
  "fuse_batchnorm_with_tconv" "fuse_activation_function"
  "fuse_instnorm" "fuse_prelu"
  "fuse_mean_with_mean" "fuse_transpose_with_mean"
  "transform_min_max_to_relu6" "transform_min_relu_to_relu6"

  # Remove redundant operators
  "remove_redundant_reshape" "remove_redundant_transpose"
  "remove_unnecessary_reshape" "remove_unnecessary_slice"
  "remove_unnecessary_strided_slice" "remove_unnecessary_split"

  # Canonicalization
  "resolve_customop_add" "resolve_customop_batchmatmul"
  "resolve_customop_matmul" "resolve_customop_max_pool_with_argmax"
  "resolve_customop_splitv" "substitute_pack_to_reshape"
  "substitute_padv2_to_pad" "substitute_splitv_to_split"
  "substitute_squeeze_to_reshape" "substitute_strided_slice_to_reshape"
  "substitute_transpose_to_reshape" "forward_reshape_to_unaryop"
  "forward_transpose_op" "replace_non_const_fc_with_batch_matmul"
)

declare -a NO_O1_OPTS=(
  "convert_nchw_to_nhwc" "expand_broadcast_const"
  "nchw_to_nhwc_input_shape" "nchw_to_nhwc_output_shape"
  "fuse_bcq" "fuse_preactivation_batchnorm"
  "make_batchnorm_gamma_positive" "replace_cw_mul_add_with_depthwise_conv"
  "remove_fakequant" "remove_quantdequant"
  "remove_redundant_quantize" "replace_sub_with_add"
  "shuffle_weight_to_16x1float32" "unroll_unidirseqlstm"
)


for opt in "${O1_OPTS[@]}"
do
  if ! grep -q ${opt} ${outputfile}.log; then
    trap_err_onexit
  fi
done

for no_opt in "${NO_O1_OPTS[@]}"
do
  if grep -q ${no_opt} ${outputfile}.log; then
    trap_err_onexit
  fi
done
```

compiler/one-cmds/tests/onecc_044.test
mhs4670go: FYI, this is [here-string](https://tldp.org/LDP/abs/html/x17837.html). In a nutshell, it will be an input of the command.

compiler/one-cmds/tests/onecc_043.test
seanshpark: > Maybe we need to check the O1 option.

Can you explain more about this?

compiler/one-cmds/tests/onecc_044.test
jinevening: `<<<` is used to pass a string.

https://unix.stackexchange.com/a/80368

compiler/one-cmds/tests/onecc_043.test
jinevening: Hmm.. this code has to be updated if the list of options in `O1` is changed.

mhs4670go: > > Maybe we need to check the O1 option.
> 
> Can you explain more about this?

This test will be passed even though O1 option turns off. So, I think we need to check if the option turns off well.

mhs4670go: > Hmm.. this code has to be updated whenever the list of options in `O1` are changed, which is too burdensome.
> 
> We may find another way (make a single file that maintains `O1` list) to implement this check.

I thought so at first. Hmm.. but checking it seems necessary.

jinevening: > I thought so at first. Hmm.. but checking it seems necessary.

I agree with this. I'll make another PR for that.

seanshpark: > So, I think we need to check if the option turns off well.

If this test is for what @mhs4670go explained, I think it would be better to check in some unit(?) test not in this end-user one-cmds test.

jinevening: @seanshpark I guess that what @mhs4670go wanted to say is, we have to check the individual options of `O1` are turned on (except `fold_add_v2=False` in the cfg file).

And if we hard-code the list of optimization options in each test file, we will have to update the test files whenever `O1` is updated. To address this, I made https://github.com/Samsung/ONE/pull/10474, which makes a single source of optimization list for `O1` (= `O1.txt`). The file will be copied to test directory and used by this test :)

I'll rebase this PR after #10474 is landed.

jinevening:
I close this as https://github.com/Samsung/ONE/pull/10530 will be landed.

PR #10474, opened 2023-02-17
[one-cmds] Use separate file for O1 option

jinevening:
This uses a separate file for O1 option list.

---
Related to: https://github.com/Samsung/ONE/pull/10473#discussion_r1109357181


compiler/one-cmds/CMakeLists.txt
```diff
@@ -151,3 +152,21 @@ endif(NOT ENABLE_TEST)
 add_subdirectory(dummy-driver)
 add_subdirectory(tests)
 add_subdirectory(validate-onnx2circle)
+
+# Copy O1.txt file to test directory
```
jinevening, line 156: `one-cmds` tests (https://github.com/Samsung/ONE/pull/10473) will use O1.txt.

This makes a single source of O1 list, making it easier to test default option sets.

compiler/one-cmds/onelib/constant.py
```diff
@@ -20,56 +20,8 @@ class CONSTANT:
...
-        'substitute_splitv_to_split',
-        'substitute_squeeze_to_reshape',
-        'substitute_strided_slice_to_reshape',
-        'substitute_transpose_to_reshape',
-        'forward_reshape_to_unaryop',
-        'forward_transpose_op',
-        'replace_non_const_fc_with_batch_matmul',  # For quantization
-    )
+    with open('O1.txt', 'r') as f:
+        O1 = f.read().split()
```
mhs4670go, line 24: How about add same comments into `O1.text`? Then, below codes would be helpful.
```python
O1=list()
with open('O1.txt', 'r') as f:
    for line in f:
        # Ignore the comments
        if not line.startswith('#'):
            O.append(line.rstrip())

# Make it immutable
O1=tuple(O1)
```

jinevening: That will make parsing codes in tests complicated.. but let's do it :)

mhs4670go (approved):
LGTM

compiler/one-cmds/CMakeLists.txt
```diff
@@ -101,7 +101,8 @@ set(ONE_PYTHON_FILES constant.py
                      OptionBuilder.py
                      TopologicalSortHelper.py
                      WorkflowRunner.py
-                     utils.py)
+                     utils.py
+                     O1.txt)
```
seanshpark, line 105: 1) this is not python file. please move to so somewhere else add new copy sections.

2) please do not use `.txt` file. this file format is text but not the content. maybe `.list` or `.options`.

seanshpark, line 105: @mhs4670go , when you first designed group option, didn't you provide a way to define a file  with options like this?

mhs4670go: @seanshpark Yes. You can refer below files. But two features are a bit different.
- https://github.com/Samsung/ONE/blob/4aef2205cf05f80aa0eb0c2b83e793aa9b7ee22e/compiler/one-cmds/tests/onecc_024.test
- https://github.com/Samsung/ONE/blob/4aef2205cf05f80aa0eb0c2b83e793aa9b7ee22e/compiler/one-cmds/tests/OONECC_024.cfg

But, there are some different things. Say what I supported is Feature1(F1) and this one is F2.
- F1 is for _onecc_ interface. Like `onecc -C one.cfg -O1 -ONNX`. It also customizes other tools' options like _one-import-tf_, _one-quantize_ and etc. 
- F2 is for _one-optimize_ feature and just only for optimization option.
- F1, the group option can be created by user as well as by us. All the users have to do is create O{option_name}.cfg under _optimization_ folder.
- F2, the group option is only created by us. It is why this PR hard codes the O1.txt.

seanshpark: Why not use `O1` as described for F1? Is there a particular reason not to use this way?

jinevening: Well, there are several reasons.

1. F1 is only supported in cmdline for now. I wanted to enable the feature in cfg file.
2. It is difficult to set the priority of options. I want to make individual options have higher priority than default option. For example, if `O1=True` and `fold_dequantize=False` in cfg file, I want to turn off `fold_dequantize`. Current implementation of F1 overwrites `O1` (where `fold_dequantize=True`) even if `fold_dequantize=False` is given in the cfg file.

seanshpark: My concern is that there is two F1, F2 ways of providing group options, that are slightly differently in some aspects.
I'd like these to be provided in one way.
IIUC, group option can be off in F2, there is no particular way of doing this in F1.
@mhs4670go , can you revise F1 work as such?

mhs4670go: > I'm not sure how to control priority of multiple cfg files in F1.

Actually F1 accumulates the options that each multiple cfg files turns on because there is no way to turn off the options in the command line.

Hmm.. two features may cause a confusion. How about only use the F1 feature instead?
1. Introduce _opt_ or _optimization_ key to _onecc_ section
```
[onecc]
one-import-tf=True
# ..
optimization=O1
```
2. Decide priorities of multiple cfg files. I think making the cfg file given by C option have a highest priority is enough. 
3. Add and install O1.cfg file to the `/usr/hsare/one/optimization/`.

compiler/one-cmds/onelib/constant.py
jinevening: This would work `readarray -t O1_list <<< $(grep -v -e "^$" -e "^#.*" O1.options)`.

compiler/one-cmds/CMakeLists.txt
mhs4670go: > can you revise F1 work as such?

I'll post a draft about this soon.

jinevening: > Actually F1 accumulates the options that each multiple cfg files turns on because there is no way to turn off the options in the command line.

I understand why it was implemented that way. But if a user uses a cfg file, option can be turned off.

---

If we're going to keep F1 feature, the following things may have to be resolved.

1. cfg interface. F1 does not support cfg interface. AFAIK, F1 allows sections other than `one-optimize`, so `optimization=O1` seems too optimization-specific. `include` or something else would be more appropriate. BTW, are we going to allow include multiple cfg files? For example, `include=1.cfg, include=2.cfg` ...

2. cfg files can conflict with each other (as mentioned above, e.g., `O1=True`, `fold_dequantize=False`). It is necessary to define user-friendly rules to resolve the conflict.

@mhs4670go Could you address the above questions? Or it may be possible to drop F1 feature, because there is no active user of F1 until now.

seanshpark: > Or it may be possible to drop F1 feature, because there is no active user of F1 until now.

actually I expected this `O1` would be the first user -_-;

jinevening: > I expected this O1 would be the first user

I examined the code, and found that F1 is not enough for implementing what I want (#10381). I made F2 to implement #10381 as simply as possible. It is possible to fix F1 and drop F2. It is up to you. I said because F1 still seems like an experimental feature.

I'll wait for @mhs4670go to address the above issues :)

mhs4670go: @jinevening 
### 1. Naming issue

It is about naming issue. The reason why I use a "optimization" is as follows.

The "optimization" here refers to the optimizations performed throughout the overall compilation process. 

For example, say there is a pass in the _import_ step where BatchMatMul is unfolded into smaller ops like bunch of FullyConnected. But some users may want to turn off the pass because not unfolding the BatchMatMul would decrease the latency, which can be done by giving an option named "Disable_unfold_BatchMatMul`. That option should be placed under the _import_ tools.

Likewise, some optimization should be considered throughout the compilation process not only inside the _one-optimize_. What I call "optimization" here has more general meaning.

Anyway, this seems just a matter of personal preference. If you still think it'd cause a confusion, `include" looks good to me.

### 2. Priority issue

Current policy about option priority is simple.
```
The option from command line > The one from configuration file
```

And, I think it would be good to stick with this policy for simplicity. Let's think of "-O" feature as the option from the command line.

> For example, if `O1=True` and `fold_dequantize=False` in cfg file, I want to turn off `fold_dequantize`. 

So, I think if `O1=True`(`fold_dequantize=True`) and `fold_dequantize=False`, the `fold_dequantize` should be True. If user want to turn off that option and turn on O1 simultaneously, all user should do is unfold the O1 option.

Please feel free to let me know about your opinion.

jinevening: > It is about naming issue. 

My comment might have confused you. It is not about naming issue. It is about "supporting default options in cfg file" (not cmdline). Naming is a minor issue.

My request is to define behavior (especially about priority between default opt and other opts) of F1 in cfg file and implement the feature, so that users (or developers) can use F1 feature.

> The option from command line > The one from configuration file

I don't object to this policy.

> O1=True(fold_dequantize=True) and fold_dequantize=False, the fold_dequantize should be True.

This is not about the above policy (cmdline VS cfg). It is about "priority between default option and other options". Please note that in my scenario default option is given by cfg file, because most of our customers use cfg files (command would be fixed (`onecc -C XX.cfg`)). Anyway, this is the point where we have different views.

```ini
[one-optimize]
O1=True  (or include=O1 in [onecc] section)
fold_dequantize=False
```

The issue is whether to ignore `fold_dequantize=False` or not. @seanshpark @lemmaa Could you give your opinions?

seanshpark: > The issue is whether to ignore fold_dequantize=False or not

1/ we can decide to use the order of options. 

```
[one-optimize]
O1=True  (or include=O1 in [onecc] section)
fold_dequantize=False
```
--> `fold_dequantize=False`

```
[one-optimize]
fold_dequantize=False
O1=True  (or include=O1 in [onecc] section)
```
--> `fold_dequantize=True`

2/ we can decide priority to `group option` < `individual option`
so for above two cases, it would be `fold_dequantize=False`

personally I prefer 2/
- group option is to set multiple options
- for some case like finding better results, it would be useful when user wants to turn off just one or two options.
- technically, order in `.cfg`, that is in `INI` format, is not defined for individual items, thus not sure management of order can be strict.

mhs4670go: Okay. I'll write a draft based on the follows.
1. group option < individual option
2. Name it `include`
```ini
[onecc]
include=O1
```

SlavikMIPT (changes requested):

SlavikMIPT (approved):
LGTM

jinevening:
Let's close this PR as https://github.com/Samsung/ONE/pull/10530 will be landed.

PR #10475, opened 2023-02-20
[onert-micro] Refactor onert-micro

BalyshevArtem:
This pr refactors onert-micro to use lower runtime memory and have smaller binary file size.
Also removes all unused files. 


onert-micro/CMakeLists.txt
```diff
@@ -56,6 +56,25 @@ else()
     return()
 endif()
 
+if (USE_STATIC_ALLOC)
+    # TODO: enable it
+    message(STATUS "FAILED ONERT-MICRO is not support Static Memory Manager now")
```
chunseoklee, line 61: Q. Static MM here means TensorArena in tflite micro ?

BalyshevArtem: I think it is similar to it. Static MM has buffer and know how to allocate intermediate tensors during execution. It works only with intermediate tensors.

SlavikMIPT (approved):
LGTM

chunseoklee (approved):

PR #10478, opened 2023-02-22
[DRAFT] Revisit group option

mhs4670go:
This draft is to revisit group option.

- Introduce `include` key to _onecc_ section that receives optimization group option
- Adjust the priorities among options : individual options > group options
- Introduce `optimization/O1.cfg`

Related: https://github.com/Samsung/ONE/pull/10474#discussion_r1113763693


compiler/one-cmds/onelib/utils.py
```diff
@@ -104,44 +106,36 @@ def parse_cfg_and_overwrite(config_path, section, args):
...
+
+    for key in parser[section_to_parse]:
+        if is_accumulated_arg(key, section_to_parse):
             if not is_valid_attr(args, key):
-                setattr(args, key, config[section_to_run][key])
+                setattr(args, key, [parser[section_to_parse][key]])
+            else:
+                getattr(args, key).append(parser[section_to_parse][key])
+            continue
+        if not overwrite and hasattr(args, key):
```
jinevening, line 136: At this point, `hasattr(args, key)` always returns true (even though it is not set by user). That's because when Python argparser parses arguments with `action='store_true'`, they save attributes name and default value = False.

So, this makes all optimization options in O1.cfg be not applied.

In the test `onecc_044`, `test_onnx_model.onecc_044.opt.cirlce.log` is printed as below.

```
~/ONE/build/release.install/bin/circle2circle test_onnx_model.circle test_onnx_model.onecc_044.opt.circle --convert_nchw_to_nhwc
```

O1 options are not applied. The test is passed because it does not check individual options, but the actual behavior is wrong.

jinevening: A simple workaround would be to use `hasattr(args, key) and getattr(args,key)`, but it adds an assumption that arguments must be defined with `action=store_true. Arguments with `action=store_false` will not work..

mhs4670go:
@jinevening I've revisited the logic and tests according to your comments. Could you please take another look?

compiler/one-cmds/one-optimize
```diff
@@ -137,22 +137,47 @@ def _parse_opt(args):
...
+        # group option do not overwrite existing args
+        oneutils.parse_cfg(config_path, 'one-optimize', args)
 
 
+# There are several cases to receive the optimization options:
+#  - Indivisual option
+#    1. From command line
+#    2. From cfg file
+#  - Group option
+#    3. From command line
```
jinevening, line 149:
```suggestion
#    3. From command line
# 
# NOTE Group option from cfg file (`include` in `[onecc]` section) is passed as a command line argument
```

compiler/one-cmds/one-optimize
```diff
@@ -137,22 +137,47 @@ def _parse_opt(args):
...
+#
+# Their priority is as follows, since each option can be given simultaneously:
+#  1. Indivisual option from command line
+#  2. Indivisual option from cfg file
+#  3. Group option from command line
```
jinevening, line 154: This looks reasonable to me :)

compiler/one-cmds/one-codegen
```diff
@@ -113,6 +113,8 @@ def _parse_arg(parser):
     del argv[0]
     # split by '--'
     args = [list(y) for x, y in itertools.groupby(argv, lambda z: z == '--') if not x]
+    if len(args) == 0:
+        codegen_args = parser.parse_args(codegen_args)
```
jinevening, line 117: What is this change for? There is a similar change in one-infer and one-profile too.

compiler/one-cmds/tests/onecc_024.test
```diff
@@ -70,7 +70,7 @@ LUCI_LOG=5 onecc -C ${configfile} -OONECC_024 > ${filename}.log 2>&1
 
 clean_envir
 
-if grep -q "MakeBatchNormGammaPositivePass" "${filename}.log"; then
+if ! grep -q "MakeBatchNormGammaPositivePass" "${filename}.log"; then
```
jinevening, line 73: This change is necessary because now group options have lower priority than the options in user-given cfg file.

OONECC_024.cfg (group option)
```
make_batchnorm_gamma_positive=True
```

onecc_024.cfg (user option)
```
make_batchnorm_gamma_positive=False
```

The result is now False.

jinevening:
@mhs4670go  Could you merge the changes related to group options (`include`)? I can make the following PRs for O1.

compiler/one-cmds/one-codegen
mhs4670go: This is added for a technical reason. 

```diff
-    oneutils.parse_cfg(args, 'one-codegen')
+    oneutils.parse_cfg(args.config, 'one-codegen', args)
```

Before above diff, `args` can be a list or the obejct. But, after above diff, `args` always must be the returned object from argument parser. Which is why I changed `codegen_args`''s type from a list to the object when `args`'s length is 0.

Issue #10484, opened 2023-03-02
[luci] Forward Transpose for Unary Op; Abs


seanshpark:
We have a pattern `Transpose - Abs - Transpose` which can be just `Abs`.
- With `Transpose - Abs - Transpose`
- --> `Abs - Transpose - Transpose`
- `Transpose - Transpose` can be removed from another pass.

Let's support this pass.

seanshpark:
info: from 22 repo 177 issue

seanshpark:
`circledump` of `Net_Transpose_Abs_000.pass.circle` in `luci-pass-value-test`
```
O(0:0) ABS
    I T(0:0) ifm
    O T(0:1) ofm
O(0:1) TRANSPOSE
    I T(0:1) ofm
    I T(0:2) perm_C
    O T(0:3) transpose_C

Inputs/Outputs: I(input)/O(output) T(tensor index) OperandName
I T(0:0) ifm
O T(0:3) transpose_C
```
where source file `Net_Transpose_Abs_000.circle` is
```
O(0:0) TRANSPOSE
    I T(0:0) ifm
    I T(0:1) perm
    O T(0:2) transpose
O(0:1) ABS
    I T(0:2) transpose
    O T(0:3) ofm

Inputs/Outputs: I(input)/O(output) T(tensor index) OperandName
I T(0:0) ifm
O T(0:3) ofm
```

seanshpark:
done

PR #10485, opened 2023-03-02
DRAFT foward transpose with unary op

seanshpark:
on-going draft to support foward transpose with unary op.


seanshpark:
for #10484

seanshpark:
done

PR #10488, opened 2023-03-03
[luci/pass] Revise ForwardTransposeOpPass with Abs

seanshpark:
This will revise ForwardTransposeOpPass to support Unary Abs Op.


seanshpark:
for #10484 from draft #10485

jinevening (approved):
LGTM

PR #10491, opened 2023-03-07
[luci/pass] Test for ForwardTransposeOpPass Unary ABS

seanshpark:
This will add unit tests for ForwardTransposeOpPass Unary ABS Op.

### DIFF
--- a/compiler/luci/pass/src/ForwardTransposeOpPass.test.cpp
+++ b/compiler/luci/pass/src/ForwardTransposeOpPass.test.cpp
@@ -353,3 +353,172 @@ TEST_F(ForwardTransposeToMulGraphTest, forward_transpose_mul_NEG)
   luci::ForwardTransposeOpPass pass;
   EXPECT_FALSE(pass.run(_graph.g()));
 }
+
+// Unary
+
+namespace
+{
+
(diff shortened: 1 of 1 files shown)


seanshpark:
for #10484 from draft #10485

compiler/luci/pass/src/ForwardTransposeOpPass.test.cpp
```diff
@@ -353,3 +353,186 @@ TEST_F(ForwardTransposeToMulGraphTest, forward_transpose_mul_NEG)
...
+    _unary->x(_relu);
+
+    output()->from(_unary);
+  }
+
+protected:
+  luci::CircleRelu *_relu = nullptr;
+};
+
+void run_phase_unary(loco::Graph *g)
```
jinevening, line 464: Do you have a reason not to use existing `run_phase`?

seanshpark: Added `run_phase_unary()` for difference for testing but seems there was none.
I think I forgot to remove this -_-;

jinevening (approved):
LGTM

PR #10492, opened 2023-03-07
[luci/pass] Add value test for forward_transpose_op

seanshpark:
This will add value test for forward_transpose_op with Net_Transpose_Abs_000.


jinevening (approved):
LGTM

PR #10432, opened 2023-02-06
[circle-mpqsolver] Replace 'ErrorApproximator' with 'VISQErrorApproximator'

stamalakhov:
This PR replaces 'ErrorApproximator' with 'VISQErrorApproximator'

This PR intoduces:
1) `total absolute error` and `batch_mode` into `visq` component.
2) it replaces `ErrorApproximator` in `circle-mpqsolver` with `VISQErrorApproximator`
3) introduces `circle-mpqsolver` into packaging.

FYI:
for mobile_net_v1 and 200 images it produces the following output:
```
Full int16 model quantization error 9.84288e-05
Full uint8 model quantization error 0.000275709
Target quantization error 0.000187069
Q16 will be set at input due to 
152628 error at input vs 
50324.6 error at output.
16 : 0.000148376
8 : 0.000175282
4 : 0.000189274
6 : 0.000176954
5 : 0.000181785
4 : 0.000189274
Found the best configuration at 5 depth.
Elapsed Time: 19.533 minutes.
```
Output for CARTOON and 200 images:
```
Full int16 model quantization error 0.0037992
Full uint8 model quantization error 0.0617312
Target quantization error 0.0327652
Q8 will be set at input due to 
1.25213e+06 error at input vs 
2.38788e+06 error at output.
31 : 0.0119042
47 : 0.0154902
55 : 0.0219636
59 : 0.0614759
57 : 0.0448715
56 : 0.0244399
Found the best configuration at 56 depth.
Elapsed Time: 40.15 minutes.

```
Issue: #10393
ONE-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

Issue #10393, opened 2023-01-31
[circle-mpqsolver] Replace `ErrorApproximator` with `visq`


stamalakhov:
### What

Replace `ErrorApproximator` with `visq`.
### Why
Because `ErrorApproximator` is not general enough to provide error estimations for all types of nodes, we need to replace it with `visq`.

stamalakhov:
@jinevening 
Introduction of total error computer to `visq` component produced admissible results:
![image](https://user-images.githubusercontent.com/112689352/216008239-98ec06c9-ebbe-4dd9-9e84-c8451d918a2d.png)
But it is very disk intensive, actually for MobileNetV1 it took 30 Gb, and for CARTOON over 60 Gb (100 images only). So i don't have a chance to run it on my ssd, just using hdd.
May be for mpq_solver  we could estimate total error by inside mpq_solver using ExecutionObserver ?

jinevening:
> May be inside mpq_solver we could estimate total error using ExecutionObserver and directly calling Interpreter with hooks attached?

I think that will significantly increase program complexity. Maybe you can make a separate module (or extend `visq`) to generate the information you want.

As to the disk usage of `visq`, I agree that is definitely an issue. One alternative may be to use partial dataset.

You can make partial h5 file using `h5copy`.

jinevening:
> Or may be we can change visq component, so that it process quantization errors pairwise (float vs fake_quantized) on each sample sequentially?

This can also be a good alternative. Currenlty `visq` invokes `dalgona` with full data set. For per-sample analysis, `visq` will have to invoke `dalgona` per sample.

stamalakhov:
> > Or may be we can change visq component, so that it process quantization errors pairwise (float vs fake_quantized) on each sample sequentially?
> 
> This can also be a good alternative. Currenlty `visq` invokes `dalgona` with full data set. For per-sample analysis, `visq` will have to invoke `dalgona` per sample.

Ok. I'll try this one.

stamalakhov:
Seems like we can close this one, as for now bisection can be used with `visq` parameter for `auto` mode.

PR #10500, opened 2023-03-09
[visq] Introduce 'batch' mode

stamalakhov:
This commit introduces 'batch' mode for `visq` component.

Draft: #10432
Related to: #10393

### DIFF
--- a/compiler/visq/visq
+++ b/compiler/visq/visq
@@ -27,6 +27,9 @@ import os
 import math
 import sys
 
+import h5py as h5
+import numpy as np
+
 from shutil import copyfile
 from pathlib import Path
 
@@ -70,6 +73,13 @@ def _get_parser():
         required=False)
     parser.add_argument(
         "--dump_dot_graph", action="store_true", help="Dump dot graph.", required=False)
+    parser.add_argument(
+        "-b",
+        "--batch_size",
+        type=int,
+        help="Batch size to process large datasets.",
+        required=False)
+
     return parser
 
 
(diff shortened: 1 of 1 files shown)


compiler/visq/visq
```diff
@@ -107,6 +118,64 @@ def _run_dalgona(model, data, analysis, save_dir):
         raise
 
 
+def save_inputs(inputs, inputs_path):
```
jinevening, line 121:
```suggestion
# Generate h5 file that contains a dataset of a single batch
# This is for batch execution of visq
def gen_batch_h5(inputs_data, inputs_path):
```

compiler/visq/visq
```diff
@@ -107,6 +118,64 @@ def _run_dalgona(model, data, analysis, save_dir):
...
+    for i in range(len(inputs)):
+        sample = group.create_group(str(i))
+        for j in range(len(inputs[i])):
+            sample.create_dataset(str(j), data=inputs[i][j])
+
+    h5_file.close()
+    return output_path
+
+
+def advance_on_data(models, data_batch_path, computers):
```
jinevening, line 137: Can you split models into fp32_model and fq_model?
```suggestion
def advance_on_data(models, data_batch_path, computers):
```

compiler/visq/visq
```diff
@@ -107,6 +118,64 @@ def _run_dalgona(model, data, analysis, save_dir):
...
+            computers[metric_key][0].advance_on(fp32_dir, fq_dir)
+
+
+def _run_batch(models, data, computers, batch_size):
+    with tempfile.TemporaryDirectory() as inputs_dir:
+        with h5.File(data, 'r') as f:
+            dataset = f['value']
+
+            inputs = []
+            for index, data_index in enumerate(dataset):
```
jinevening, line 161: index is not used anywhere. You don't have to use enumerate.

compiler/visq/visq
```diff
@@ -107,6 +118,64 @@ def _run_dalgona(model, data, analysis, save_dir):
...
+                inputs.append(cur_inputs)
+                if len(inputs) >= batch_size:
+                    input_path = save_inputs(inputs, inputs_dir)
+                    advance_on_data(models, input_path, computers)
+                    inputs = []
+
+            if len(inputs) > 0:
+                input_path = save_inputs(inputs, inputs_dir)
+                advance_on_data(models, input_path, computers)
+                inputs = []
```
jinevening, line 176: Seems unnecessary
```suggestion
```

jinevening:
Please fix the format error.

stamalakhov:
> Please fix the format error.

Sorry for delay. Was out of office.

jinevening (dismissed):
LGTM

compiler/visq/visq
```diff
@@ -107,6 +117,66 @@ def _run_dalgona(model, data, analysis, save_dir):
...
+def advance_on_data(fp32_model, fq_model, data, computers):
+
+    curr_dir = Path(__file__).parent.resolve()
+    dump_fp32_py = curr_dir / 'visqlib' / 'DumpFP32FM.py'
+    dump_fq_py = curr_dir / 'visqlib' / 'DumpFakeQuantFM.py'
+
+    with tempfile.TemporaryDirectory() as fp32_dir, \
+         tempfile.TemporaryDirectory() as fq_dir:
+
+        #print("advance on data + {fp_dir} + {fq_dir}".format(fp_dir = fp32_dir, fq_dir = fq_dir))
```
seanshpark, line 148: Q) what is the need for this comment?

compiler/visq/visq
```diff
@@ -172,19 +242,48 @@ def _save_dot(circle_path: str, dot_path: str, metric: str, colors: list, qerror
...
+                          batch_size):
 
-    fp32_model = args.fp32_circle
-    q_model = args.q_circle
-    data = None
-    if valid_attr(args, 'data'):
-        data = args.data
-    dump_dot_graph = args.dump_dot_graph
+    with tempfile.TemporaryDirectory() as model_dir:
+        fq_model = model_dir + '/fq_model.circle'
```
seanshpark, line 249: is there a reason to fix this name?

compiler/visq/visq
stamalakhov: Sorry. No. We don't need it. I'll remove it.

compiler/visq/visq
stamalakhov: It stands for `fake_quantize_model`. As it's an internal it should be created in temporary directory to be removed as soon as possible.  IMHO there is no other reason to call it `fq_model` other than ease of use and for the full batch it's also called the same.

seanshpark: OK, it is temporary file.

seanshpark: maybe use `TemporaryFile()` not a fixed name?

stamalakhov: Yes, IMHO `TemporaryFile()` can also be used, but then IMHO we should also change it for `run_on_data`. IMHO this is not needful for this PR. May be could deal with later?

seanshpark: OK got it

seanshpark (approved):
LGTM

jinevening (approved):

PR #10511, opened 2023-03-14
Fix build error for luci_interpreter_micro_arm Target

chunseoklee:
reference : https://gitlab.kitware.com/cmake/cmake/-/issues/19693


BalyshevArtem (dismissed):
LGTM

BalyshevArtem:
Maybe only need to add more words to commit message

PR #10516, opened 2023-03-15
[one-cmds] Introduce export_constant.py

mhs4670go:
This commit introduces a script that exports CONSTANT value with given format.

Draft: #10478 
Related: https://github.com/Samsung/ONE/pull/10474#discussion_r1113763693

### DIFF
--- a/compiler/one-cmds/onelib/export_constant.py
+++ b/compiler/one-cmds/onelib/export_constant.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


compiler/one-cmds/onelib/export_constant.py
```diff
@@ -0,0 +1,74 @@
...
+    parser = argparse.ArgumentParser(
+        description='Export CONSTANT value with given file format')
+    parser.add_argument(
+        '-c', '--constant', type=str, required=True, help='Constant name to export')
+    parser.add_argument(
+        '-f',
+        '--format',
+        type=str,
+        required=True,
+        choices=['cfg', 'txt'],
```
jinevening, line 33: How about `optcfg` or `opt_cfg`?

```suggestion
        choices=['opt_cfg', 'txt'],
```

The current implementation uses a fixed section name [one-optimize].

mhs4670go: Hmm.. I think `opt_cfg` is also hard to know exporting cfg has only [one-optimize] section. I'll append the related help message to the description.

jinevening (approved):
LGTM

PR #10518, opened 2023-03-15
DRAFT: Support fine-grained input/output_type quantization

jinevening:
On going draft to support fine-grained input/output_type quantization.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/10493

Issue #10493, opened 2023-03-07
[one-quantize] Support fine-grained input/output quantization dtype


jinevening:
## What

Let's allow users to set dtype of each input/output

## Why

Current interface only allows to set the same dtype for all inputs (or all outputs).

From `one-quantize -h`
```
  --input_type INPUT_TYPE
                        data type of inputs of quantized model (supported:
                        uint8, int16, float32, default=quantized_dtype).
                        QUANTIZE Op will be inserted at the beginning of the
                        quantized model if input_type is different from
                        quantized_dtype.
  --output_type OUTPUT_TYPE
                        data type of outputs of quantized model (supported:
                        uint8, int16, float32, default=quantized_dtype).
                        QUANTIZE Op will be inserted at the end of the
                        quantized model if output_type is different from
                        quantized_dtype.
```

This works for most of simple CNN models. However, complex NN models (ex: RNN, transformer) will require a more fine-grained setting for input/output, especially with mixed-precision quantization.

jinevening:
Interface will be like as below.

For a model with three inputs and three outputs,

cmdline
```
onecc quantize --input_type uint8;int16;uint8 --output_type int16;uint8;int16
```

cfg
```
[one-quantize]
...
input_type=uint8;int16;uint8
output_type=int16;uint8;int16
```

The order of inputs/outputs must be set correctly.

seanshpark:
> onecc quantize --input_type uint8;int16;uint8 --output_type int16;uint8;int16

`;` is used to separate command

```
$ a;b;c
-bash: a: command not found
-bash: b: command not found
-bash: c: command not found
```

do not use this.

jinevening:
Thanks. I'll use `,` as a delimiter.

jinevening:
All done.

PR #10521, opened 2023-03-15
[luci] Canonicalize user-given input/output_type

jinevening:
This parses input/output_type and canonicalize them.

---
Related to: https://github.com/Samsung/ONE/issues/10493
Draft PR: https://github.com/Samsung/ONE/pull/10518

### DIFF
--- a/compiler/luci/pass/src/CircleQuantizer.cpp
+++ b/compiler/luci/pass/src/CircleQuantizer.cpp
@@ -113,14 +113,15 @@ void canonicalize_input_type(loco::Graph *g, std::vector<loco::DataType> &input_
           input_type[i] != loco::DataType::S16)
       {
         throw std::runtime_error("Input dtype of " + input->name() +
-                                 " is invalid. Model's input is not float32.");
+                                 " is invalid. For float32 input, the input dtype after "
+                                 "quantization must be one of uint8, int16, or float32.");
       }
     }
   }
 }
 
-// This function updates user-given input_type to match with the input signature of graph
-// If user gives only one input_type, it will be expanded to the number of graph inputs
+// This function updates user-given output_type to match with the output signature of graph
+// If user gives only one output_type, it will be expanded to the number of graph outputs
 // NOTE This function is almost same with canonicalize_input_type, but it is written as a
 // separate function for more precise error messaging.
 // TODO Find a way to reduce duplicate codes
@@ -186,7 +187,8 @@ void canonicalize_output_type(loco::Graph *g, std::vector<loco::DataType> &outpu
           output_type[i] != loco::DataType::S16)
       {
         throw std::runtime_error("Output dtype of " + output->name() +
-                                 " is invalid. Model's output is not float32.");
+                                 " is invalid. For float32 output, the output dtype after "
+                                 "quantization must be one of uint8, int16, or float32.");
       }
     }
   }
--- a/compiler/luci/pass/src/CircleQuantizer.cpp
+++ b/compiler/luci/pass/src/CircleQuantizer.cpp
@@ -54,7 +54,9 @@ using LayerParam = luci::CircleQuantizer::Options::LayerParam;
 // If user gives only one input_type, it will be expanded to the number of graph inputs
 void canonicalize_input_type(loco::Graph *g, std::vector<loco::DataType> &input_type)
 {
-  assert(g); // FIX_CALLER_UNLESS
+  if (g == nullptr)
+    return;
+
   const auto inputs = g->inputs();
 
   assert(inputs); // FIX_CALLER_UNLESS
@@ -127,7 +129,9 @@ void canonicalize_input_type(loco::Graph *g, std::vector<loco::DataType> &input_
 // TODO Find a way to reduce duplicate codes
 void canonicalize_output_type(loco::Graph *g, std::vector<loco::DataType> &output_type)
 {
-  assert(g); // FIX_CALLER_UNLESS
+  if (g == nullptr)
+    return;
+
   const auto outputs = g->outputs();
 
   assert(outputs); // FIX_CALLER_UNLESS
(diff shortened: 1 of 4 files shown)


compiler/luci/pass/src/CircleQuantizer.cpp
```diff
@@ -268,6 +412,9 @@ void CircleQuantizer::quantize(loco::Graph *g) const
     if (output_type.empty())
       output_type = output_model_dtype;
 
+    auto input_type_vec = parse_comma_separated_str(input_type);
+    auto output_type_vec = parse_comma_separated_str(output_type);
```
jinevening, line 416: Now user input is given as a comma-separated string.

compiler/luci/pass/src/CircleQuantizer.cpp
```diff
@@ -49,6 +49,148 @@ namespace
 using namespace luci;
 using LayerParam = luci::CircleQuantizer::Options::LayerParam;
 
+// This function updates user-given input_type to match with the input signature of graph
+// If user gives only one input_type, it will be expanded to the number of graph inputs
```
jinevening, line 53: This explains why canonicalization is introduced.

Previously, users only gave one input/output_type. For backward-compatibility, we still allow users to give one input/output_type. For models with multiple inputs/outputs, we expand the one input/output_type to match with the number of inputs/outputs of a graph. I'd call the process canonicalization.

compiler/luci/pass/src/CircleQuantizer.cpp
```diff
@@ -323,8 +483,9 @@ void CircleQuantizer::quantize(loco::Graph *g) const
       ctx->input_model_dtype = str_to_dtype(input_model_dtype);
       ctx->output_model_dtype = str_to_dtype(output_model_dtype);
       ctx->granularity = str_to_granularity(granularity);
-      ctx->input_type = str_to_dtype(input_type);
-      ctx->output_type = str_to_dtype(output_type);
+      // TODO Change input/output_type to vector
```
jinevening, line 486: This TODO will be resolved in the following PR.

compiler/luci/pass/src/helpers/Strings.h
```diff
@@ -36,6 +36,10 @@ std::string to_lower_case(std::string);
 
 loco::DataType str_to_dtype(const std::string &);
 
+std::vector<std::string> parse_comma_separated_str(const std::string &);
```
mhs4670go, line 39: We could use existing library.

https://github.com/Samsung/ONE/blob/9ff5081c91f4fd9155f38beda649d0424a8b5a7a/compiler/pepper-csv2vec/include/pepper/csv2vec.h#L26

jinevening: Thanks for notice. I'll update source code to reduce duplicate code.

I'm a bit surprised that there is an individual module just for this purpose.

compiler/luci/pass/src/CircleQuantizer.cpp
```diff
@@ -49,6 +50,148 @@ namespace
 using namespace luci;
 using LayerParam = luci::CircleQuantizer::Options::LayerParam;
 
+// This function updates user-given input_type to match with the input signature of graph
+// If user gives only one input_type, it will be expanded to the number of graph inputs
+void canonicalize_input_type(loco::Graph *g, std::vector<loco::DataType> &input_type)
+{
+  assert(g); // FIX_CALLER_UNLESS
```
mhs4670go, line 57: Sometimes our static analysis tool complains about this assert on release build. Maybe we can do nothing for this case.

```suggestion
  if (g == nullptr)
    return;
    
  // ..
  
  if (inputs == nullptr)
    return;
```

compiler/luci/pass/src/CircleQuantizer.cpp
```diff
@@ -49,6 +50,148 @@ namespace
...
+          input_type[i] != loco::DataType::S16)
+      {
+        throw std::runtime_error("Input dtype of " + input->name() +
+                                 " is invalid. Model's input is not float32.");
+      }
+    }
+  }
+}
+
+// This function updates user-given input_type to match with the input signature of graph
```
mhs4670go, line 122: Maybe `input_type` -> `output_type`?

compiler/luci/pass/src/CircleQuantizer.cpp
```diff
@@ -49,6 +50,148 @@ namespace
...
+          " is invalid. It has to be the same with the model's input dtype.");
+    }
+    else
+    {
+      // C2
+      if (input_type[i] != loco::DataType::FLOAT32 and input_type[i] != loco::DataType::U8 and
+          input_type[i] != loco::DataType::S16)
+      {
+        throw std::runtime_error("Input dtype of " + input->name() +
+                                 " is invalid. Model's input is not float32.");
```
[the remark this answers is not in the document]
seanshpark: I think current message is OK.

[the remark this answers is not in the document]
mhs4670go: IIUC, 
`Model's input is not float32.` -> `Model's input is float32.`. Am I wrong? So, "Input dtype must be one of uint8, int16, float32" looks okay.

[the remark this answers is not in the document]
seanshpark: Um...
Here, Model's input is float32. Message `Model's input is not float32.` is somewhat confusing, `input` can be understood to (1) Model or (2) User given.

IIUC, message should be `Input must be one of uint8, int16, float32` without `Model's`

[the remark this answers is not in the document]
jinevening: @mhs4670go  @seanshpark  Thanks for finding. The error message looks confusing.

The error message is printed when user-given input is not one of {uint8 or int16 or float32} when model's input is float32.

I'll update the error message as below.

`For float32 input, the input type after quantization must be one of uint8, int16, or float32.`

jinevening:
Ping @mhs4670go @seanshpark

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #10523, opened 2023-03-16
[luci/log] Tidy disabled code

seanshpark:
This will tidy disabled code with directives.


mhs4670go (approved):
LGTM

PR #10539, opened 2023-03-23
[circle-mpqsolver] Implement visq parsing

stamalakhov:
This commit implements visq parsing.

Draft: https://github.com/Samsung/ONE/pull/10432
Related: https://github.com/Samsung/ONE/issues/10393

### DIFF
--- a/compiler/circle-mpqsolver/CMakeLists.txt
+++ b/compiler/circle-mpqsolver/CMakeLists.txt
@@ -1,9 +1,17 @@
+nnas_find_package(Jsoncpp)
+if(NOT Jsoncpp_FOUND)
+    message(STATUS "Build circle-mpqsolver: FAILED (missing jsoncpp)")
+    return()
+endif(NOT Jsoncpp_FOUND)
+
 file(GLOB_RECURSE SOURCES "src/*.cpp")
 file(GLOB_RECURSE TESTS "src/*.test.cpp")
 list(REMOVE_ITEM SOURCES ${TESTS})
 
 add_executable(circle-mpqsolver "${SOURCES}")
 target_include_directories(circle-mpqsolver PRIVATE src)
+target_include_directories(circle-mpqsolver PRIVATE ${Jsoncpp_INCLUDE_DIRS})
+target_link_libraries(circle-mpqsolver ${Jsoncpp_STATIC_LIB})
 target_link_libraries(circle-mpqsolver arser)
 target_link_libraries(circle-mpqsolver vconone)
 target_link_libraries(circle-mpqsolver safemain)
@@ -28,10 +36,14 @@ endif(NOT ENABLE_TEST)
 set(TEST_SOURCES
     "src/bisection/DepthParameterizer.cpp"
     "src/bisection/Quantizer.cpp"
-    "src/bisection/ErrorApproximator.cpp")
+    "src/bisection/ErrorApproximator.cpp"
+    "src/bisection/VISQErrorApproximator.cpp"
+)
 
 nnas_find_package(GTest REQUIRED)
 GTest_AddTest(circle_mpqsolver_test ${TESTS} ${TEST_SOURCES})
+target_include_directories(circle_mpqsolver_test PRIVATE ${Jsoncpp_INCLUDE_DIRS})
+target_link_libraries(circle_mpqsolver_test ${Jsoncpp_STATIC_LIB})
 target_link_libraries(circle_mpqsolver_test luci_lang)
 target_link_libraries(circle_mpqsolver_test luci_service)
 target_link_libraries(circle_mpqsolver_test luci_pass)
--- a/compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
+++ b/compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
@@ -17,6 +17,7 @@
 #include "VISQErrorApproximator.h"
 
 #include <fstream>
+#include <json.h>
 
 using namespace mpqsolver::bisection;
 
@@ -30,9 +31,28 @@ void VISQErrorApproximator::init(const std::string &visq_data_path)
   }
 }
 
-bool VISQErrorApproximator::init(std::istream &)
+bool VISQErrorApproximator::init(std::istream &visq_data)
 {
-  // TODO
+  Json::Reader reader;
+  Json::Value completeJsonData;
+  if (!reader.parse(visq_data, completeJsonData))
+  {
+    return false;
+  }
+
+  if (!completeJsonData.isMember("error"))
+  {
+    return false;
+  }
+
+  auto layers = completeJsonData["error"][0];
+  auto names = layers.getMemberNames();
+  for (auto name : names)
+  {
+    auto value = layers[name].asFloat();
+    _layer_errors[name] = value;
+  }
+
   return true;
 }
 
--- a/compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.test.cpp
+++ b/compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.test.cpp
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "VISQErrorApproximator.h"
+
+#include <json.h>
+#include <fstream>
+#include <gtest/gtest.h>
+
+namespace
+{
+
+void writeDataToFile(const std::string &path, const std::string &data)
+{
+  std::ofstream file;
+  file.open(path);
+  file << data;
+  file.close();
+}
+
+} // namespace
+
+TEST(CircleMPQSolverVISQErrorApproximatorTest, verifyResultsTest)
+{
+  static std::string errors_key = "error";
+  static std::string layer_key = "layer_0";
+  static float layer_error = 0.5f;
+  // trivial json with a single layer
+  Json::Value error_data;
+  Json::Value layer_data;
+  layer_data[layer_key] = layer_error;
+  error_data[errors_key].append(layer_data);
+
+  Json::StreamWriterBuilder builder;
+  auto data = Json::writeString(builder, error_data);
+  std::string path = std::tmpnam(nullptr);
+  writeDataToFile(path, data);
+
+  mpqsolver::bisection::VISQErrorApproximator approximator;
+  EXPECT_NO_THROW(approximator.init(path));
+  EXPECT_FLOAT_EQ(approximator.approximate(layer_key), layer_error);
+  unlink(path.c_str());
+}
+
+TEST(CircleMPQSolverVISQErrorApproximatorTest, verifyResultsTest_NEG)
+{
+  Json::Value error_data;
+  // just an empty json
+  Json::StreamWriterBuilder builder;
+  auto data = Json::writeString(builder, error_data);
+  std::string path = std::tmpnam(nullptr);
+  writeDataToFile(path, data);
+
+  mpqsolver::bisection::VISQErrorApproximator approximator;
+  EXPECT_THROW(approximator.init(path), std::exception);
+  unlink(path.c_str());
+}


compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
```diff
@@ -30,9 +31,28 @@ void VISQErrorApproximator::init(const std::string &visq_data_path)
...
 
-bool VISQErrorApproximator::init(std::istream &)
+bool VISQErrorApproximator::init(std::istream &visq_stream)
 {
-  // TODO
+  Json::Reader reader;
+  Json::Value completeJsonData;
+  if (!reader.parse(visq_stream, completeJsonData))
+  {
+    return false;
```
jinevening, line 40: How is the return value of this function? I think it is necessary to throw an exception if this function returns false.

stamalakhov: Ahhh. Sorry. Ok.

stamalakhov: done

compiler/circle-mpqsolver/CMakeLists.txt
```diff
@@ -28,10 +36,13 @@ endif(NOT ENABLE_TEST)
 set(TEST_SOURCES
     "src/bisection/DepthParameterizer.cpp"
     "src/bisection/Quantizer.cpp"
-    "src/bisection/ErrorApproximator.cpp")
+    "src/bisection/ErrorApproximator.cpp"
+    "src/bisection/VISQErrorApproximator.cpp")
```
seanshpark, line 40:
```suggestion
    "src/bisection/VISQErrorApproximator.cpp"
)
```

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
```diff
@@ -17,23 +17,45 @@
...
-bool VISQErrorApproximator::init(std::istream &)
+void VISQErrorApproximator::init(std::istream &visq_stream)
 {
-  // TODO
-  return true;
+  Json::Reader reader;
+  Json::Value completeJsonData;
+  if (!reader.parse(visq_stream, completeJsonData))
+  {
+    throw std::runtime_error("Invalid visq stream");
```
seanshpark, line 44: this message isn't used as catch in `init(const std::string &visq_data_path)
{` ignores this.

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
```diff
@@ -17,23 +17,45 @@
...
+    init(file);
+  }
+  catch (std::runtime_error &)
   {
     throw std::runtime_error("Invalid visq file " + visq_data_path);
   }
 }
 
-bool VISQErrorApproximator::init(std::istream &)
+void VISQErrorApproximator::init(std::istream &visq_stream)
```
seanshpark, line 38: I don't understand why change bool to void in this change.

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
```diff
@@ -17,23 +17,45 @@
...
+  Json::Reader reader;
+  Json::Value completeJsonData;
+  if (!reader.parse(visq_stream, completeJsonData))
+  {
+    throw std::runtime_error("Invalid visq stream");
+  }
+
+  if (!completeJsonData.isMember("error"))
+  {
+    throw std::runtime_error("No error section in visq stream");
```
seanshpark, line 49: optional) 
```suggestion
    throw std::runtime_error("No 'error' section in visq stream");
```

seanshpark: Q) what does `section` mean?

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
```diff
@@ -17,23 +17,45 @@
...
   // read file
   std::ifstream file(visq_data_path);
-  if (!init(file))
+  try
+  {
+    init(file);
+  }
+  catch (std::runtime_error &)
   {
     throw std::runtime_error("Invalid visq file " + visq_data_path);
```
jinevening, line 34: If you're going to change `init(std::istream)` to throw exceptions (for better error messaging), it would be better to remove the try-catch statement.

stamalakhov: ok

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
stamalakhov: > I don't understand why change bool to void in this change.

Now it throws to show the reason of failure.

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
stamalakhov: > Q) what does `section` mean?

`section` stands for serialized compound error info layerwise. but it can be 'array' or just `info`.

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
seanshpark: commit message is `This commit implements visq parsing.` which I don't know the relation with changing this method prototype. If you want to change this, split to another PR.

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.h
```diff
@@ -45,13 +47,15 @@ class VISQErrorApproximator final
...
+   * @brief initiliaze by visq_data (throws on invalid input)
    */
-  bool init(std::istream &visq_data);
+  void init(std::istream &visq_data);
 
 private:
   std::string _visq_data_path;
   std::map<std::string, float> _layer_errors;
+
+  friend CircleMPQSolverVISQErrorApproximatorTest;
```
seanshpark, line 58: Adding this ~Test class here looks odd.
Do you know any practice doing so?

stamalakhov: @seanshpark 
It's added to the class for unitests. For 'ONE` repository `friend` are added to the top of the class. Am i supposed to add it to the top?

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.cpp
stamalakhov: ok. 
@jinevening i'll post another PR with corrected exceptions.

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.h
seanshpark: I know the usage but this class or any other class doesn't have any relation with `CircleMPQSolverVISQErrorApproximatorTest`. 
`CircleMPQSolverVISQErrorApproximatorTest` is only for testing and testing is not included in any production.

And I don't agree on usging `~Test` as a friend of a class.

stamalakhov: Are tests supposed to use `temporary` files?

seanshpark: > Will it be ok if tests will use temporary files?

Our tests uses lots of file based testing. It is OK to use files.
I assume you choose to use memory (stream) testing.
Was there any assumption behind this? like does it consume huge space? or long time?

Alternative thing I could think of is to use some name like `ErrorApproximatorAccessor` class, not having `Test`, but as I asked above, I'd like to know any know practice like this.

stamalakhov: > Alternative thing I could think of is to use some name like `ErrorApproximatorAccessor` class, not having `Test`, but as I asked above, I'd like to know any know practice like this.

No. I've not seen this in `ONE`.
I'd rename it to `VISQErrorApproximatorAccessor`.

stamalakhov: > Was there any assumption behind this?

Yes. Like it's slower than memory testing.

seanshpark: > No. I've not seen this in ONE.

Have you seen this in any other projects? or is this your invention?

stamalakhov: I've seen it here:
https://stackoverflow.com/questions/47354280/what-is-the-best-way-of-testing-private-methods-with-googletest

seanshpark: > I recommend avoiding option 1 and 2 when you can, as it typically causes "test duplication", and as a consequence, many more tests than necessary will break when requirements change. Use them as a last resort. Option 1 and 2 are the fastest ways to "test private methods" for here and now (as in the fastest to implement), but they'll really hurt productivity in the long run.

which recommends not using `friend` class for testing.

option 1/ use with file testing with some comments as you worry about performance.
option 2/ use `protected` for `init(stream)` and inherit this class in Test.

jinevening (approved):
LGTM

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.test.cpp
```diff
@@ -0,0 +1,70 @@
...
+  static float layer_error = 0.5f;
+  // trivial json with a single layer
+  Json::Value error_data;
+  Json::Value layer_data;
+  layer_data[layer_key] = layer_error;
+  error_data[errors_key].append(layer_data);
+
+  Json::StreamWriterBuilder builder;
+  auto data = Json::writeString(builder, error_data);
+  std::string path = "/tmp/mpqsolver-valid-visq-test.visq.json";
```
seanshpark, line 51: plz do not use fixed file name. get temp file from API.

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.test.cpp
```diff
@@ -0,0 +1,83 @@
...
+  // trivial json with a single layer
+  Json::Value error_data;
+  Json::Value layer_data;
+  layer_data[layer_key] = layer_error;
+  error_data[errors_key].append(layer_data);
+
+  Json::StreamWriterBuilder builder;
+  auto data = Json::writeString(builder, error_data);
+  char path[] = "/tmp/VISQErrorApproximator-TEST-XXXXXX";
+  makeTemporaryFile(path);
```
seanshpark, line 61: use system API to get temporary file

stamalakhov: @seanshpark 
Unfortunately i failed to query the name (which can be transferred to constructor of `VISQErrorAppriximator`) of temporary file generated in this way. `tmpnam` is not recommended to use though.

stamalakhov: @seanshpark 
Ahh. Sorry. Didn't saw a method to extract name of tempfile, i'll try. Thank you.

compiler/circle-mpqsolver/src/bisection/VISQErrorApproximator.test.cpp
```diff
@@ -0,0 +1,82 @@
...
+  file.open(path);
+  file << data;
+  file.close();
+}
+
+std::string makeTemporaryFile()
+{
+  FILE *file = std::tmpfile();
+  char fname[FILENAME_MAX], link[FILENAME_MAX] = {0};
+  sprintf(fname, "/proc/self/fd/%d", fileno(file));
```
seanshpark, line 39: sorry for recommend `tmpfile()`
I couldn't catch your intention of temporary file.
Maybe https://en.cppreference.com/w/cpp/io/c/tmpnam would be better usage for this.
Can you please revise this once more?

Why?
- using system `/tmp/~~~` string looks system-ish
- `/proc/self/fd` also looks system-ish

stamalakhov: Just FYI
https://stackoverflow.com/questions/21295974/are-there-any-reasons-to-avoid-using-tmpnam-to-get-a-name-for-a-temporary-file,
IMHO seems like `tempnam` is not recommended for usage.
also from https://man7.org/linux/man-pages/man3/tmpnam.3.html
```
Note: avoid using these functions; use [mkstemp(3)](https://man7.org/linux/man-pages/man3/mkstemp.3.html) or [tmpfile(3)](https://man7.org/linux/man-pages/man3/tmpfile.3.html)
       instead.
```
May be return to `mkstemp`?

seanshpark: > May be return to mkstemp?

it requires system-ish fixed path to temporary filename. I want to avoid using this.

stamalakhov: ok

stamalakhov: @seanshpark 
Sorry for use of '/tmp` .
For `mkstemp` `tmp` may be removed , so that the name can be resolved to 'VISQERRORApproximator-TEST-XXXXXX`.
`tmp` was added for clarity that it's temporary file.

seanshpark (approved):
LGTM thank you!

SlavikMIPT (approved):
LGTM

jinevening (approved):
LGTM

Issue #10420, opened 2023-02-02
[CFE] Install circle-opselector on one-compiler-dev package


mhs4670go:
## What

Let's consider including _circle-opselector_ module into one of the _one-compiler-dev_ package.

## Why

Sometimes it would helpful for us to be able to slice the circle model when we handle the model.

This is the dependency list of the _circle-opselector_. There's no extra module dependency for _one-compiler-dev_.
https://github.com/Samsung/ONE/blob/5b90f2d0c84fed6e4c9fe5322d6d815fa6946804/compiler/circle-opselector/requires.cmake#L1-L6

seanshpark:
please too `infra/scripts/compiler_modules.sh` too

PR #10541, opened 2023-03-23
[preset] Add circle-opselector to preset 20221125

mhs4670go:
This commit adds _circle-opselector_ to preset 20221125.

Related: #10420 


seanshpark (approved):
LGTM

lemmaa:
@mhs4670go , FYI, if preset settings are changed after release, inconvenience will occur in reproducing the corresponding release. It seems that this part also needs to be considered in the release process. 

Alternatively, it is also a way to make the operation of the preset settings different. That is, presets are never modified, so that new presets are always added when necessary.

mhs4670go:
@lemmaa Yeah, it's a good point. Having a preset for release seems a good idea.

PR #10566, opened 2023-03-29
[res] Remove verbose from PyTorchExamples

seanshpark:
This will revise PyTorchExamples genegation script without verbose information.


seanshpark:
`verobse` gives addition file information and larger file size, which is useless for our case.

mhs4670go (approved):
LGTM

PR #10525, opened 2023-03-16
DRAFT: Introduce SRMSE metric to visq

jinevening:
On going draft to support SRMSE in visq.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/10524

Issue #10524, opened 2023-03-16
[visq] Check SRMSE metric 


jinevening:
## What

Let's add SRMSE (Scaled Root Mean-Squared Error) metric to properly represent a quantization error.

Limitations of current metrics
- MPEIR (Peak error / (interval + epsilon))
  - Only cares the "peak error" for each data, and the rest of the values are discarded. This metric is highly dominated by outliers.
  - For a tensor with a short Interval, the metric is distorted (by epsilon).
- MSE (Mean squared error)
  - Difficult to interpret (scale is not considered)
- TAE (Total absolute error)
  - Difficult to interpret (scale is not considered)

Why SRMSE?
- SRMSE = RMSE / scale  (from [TF quantization debugger](https://www.tensorflow.org/lite/performance/quantization_debugger#step_3_data_analysis))
- Scaling of the error is good for interpretation.
- Among several error metrics (MAE, MSE, RMSE), RMSE is chosen, because it puts more weights on outliers (MAE is discarded), and provides better interpretation than MSE.
- How to interpret?
  - SRMSE value can be a zero or a positive number. The lower is the better (let's check 1/sqrt(12) [here](https://discuss.tensorflow.org/t/question-about-quantization-debugger-metric/15467)).
  - For weights, the average qerror is scale / 4 (max error = scale /2) assuming that the actual min/max is used and the data distribution is random. So, SRMSE would typically be in [0, 0.5]
  - For activation, it is difficult to estimate expected qerror, so we may need some data in real models.

jinevening:
> (let's check 1/sqrt(12) [here](https://discuss.tensorflow.org/t/question-about-quantization-debugger-metric/15467)).

TF assumes that the model is well-quantized if quantization errors have a uniform distribution within a scale. This makes expected RMSE be scale/sqrt(12).

jinevening:
SRMSE is now supported in visq.

PR #10575, opened 2023-03-29
[visq] Introduce SRMSE metric

jinevening:
This introduces a new metric SRMSE.

---
Related to: https://github.com/Samsung/ONE/pull/10524
Draft PR: https://github.com/Samsung/ONE/pull/10525

### DIFF
--- a/compiler/visq/visqlib/QErrorComputer.py
+++ b/compiler/visq/visqlib/QErrorComputer.py
@@ -15,6 +15,7 @@
 import os
 import glob
 import numpy as np
+import json
 
 from pathlib import Path
 from visqlib.Util import to_filename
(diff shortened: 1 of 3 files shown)


compiler/visq/visqlib/QErrorComputer.py
```diff
@@ -198,3 +199,46 @@ def get_final_result(self):
...
+            for (fp32_data_path, fq_data_path) in data_path:
+                fp32_data = np.load(fp32_data_path)
+                fq_data = np.load(fq_data_path)
+
+                MSE = np.square(fp32_data - fq_data).mean()
+
+                if tensor_name in self.qerror_map:
+                    self.qerror_map[tensor_name] += MSE
+                else:
+                    self.qerror_map[tensor_name] = MSE
```
mhs4670go, line 225: If `qerror_map` only contains float value. It would be helpful.

```suggestion
qerror_map = defaultdict(float)

self.qerror_map[tensor_name] += MSE

```

jinevening: Thanks! It looks much more concise.

jinevening: There are other codes using `dict`. I'll handle this comment in another PR.

compiler/visq/visqlib/QErrorComputer.py
```diff
@@ -198,3 +199,46 @@ def get_final_result(self):
...
+
+    def get_final_result(self):
+        with open(self.scale_file) as f:
```
seanshpark, line 228: do need to do file error check?

jinevening: I guess that file error means that file's existence or contents (including format).

1. Existence
- scale.txt is created by `DumpFakeQuantFM.py` which is always called before metric computation. So we can make sure that the file exists.

2. Contents
- scale.txt is created only by `DumpFakeQuantFM.py`, so the format is fixed (json).

Or do you mean other kinds of errors?

seanshpark: Well... there are unexpected things that happen :)
We can go as-is and fix when bad things happen.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #10583, opened 2023-03-30
[visq] Use defaultdict

jinevening:
This uses defaultdict instead of dict for simplicity.


mhs4670go (approved):
LGTM

Issue #10537, opened 2023-03-22
[Q] Compile into multiple backend binaries with one cfg file


jyoungyun:
## What

Currently, the cfg file provides backend option for `codegen`, `profile`, and `infer` tools. The `onecc` executes each backend tool according to the backend option.

If I want to compile one cfg file to multiple backends binray, how can I do?

As far as I know, I need to create three different cfg files that contain each backend name.
I don't think this is efficient.

Is there any solution that can compile to multiple backend binaries using one cfg file?

If the `onecc` frontend tool supports the `backend` option, it can be compiled into multiple backend binaries with one cfg file.
If possible, could you consider supporting the `backend` option in `onecc` frontend tool?

lemmaa:
This seems like a good example of what we're experiencing with actually supporting multiple backends. 👍

mhs4670go:
Well, in a nutshell, you can do what you want to do with workflows feature rather than the cfg feature. You can see the examples in [here](https://github.com/Samsung/ONE/tree/3222491722fc813ffa72697ba21b8e841ca01faa/compiler/one-cmds/tests).

cfg feature is designed to call each _one-cmds_ tools only once. And, IMHO, I'd have multiple cfg file for multiple backends because it''d be more convenient for me. However, each person's taste is different, so workflows feature was introduced to support various use cases.

seanshpark:
@jyoungyun , this is feature needed in `ONE-vscode` ?

seanshpark:
One idea that pop-uped

as-is
```
[code-gen]
backend=trix
command=-O compild.trix .....
```
to-be
```
[code-gen]
backend=trix
command=-O compild.trix .....
backends=trix,trix-2,trix-3
trix=-O compild.trix .....
trix-2=-O compild.trix-2 .....
trix-3=-O compild.trix-3 .....
```
same key `trix` in `backends` will be ignored
```
[code-gen]
backend=trix
command=-O compild.trix .....
backends=trix-2,trix-3
trix-2=-O compild.trix-2 .....
trix-3=-O compild.trix-3 .....
```
-> we will have 3 outputs, trix, trix-2, trix-3
```
[code-gen]
backends=trix-2,trix-3
trix-2=-O compild.trix-2 .....
trix-3=-O compild.trix-3 .....
```
-> we will have2 outputs, trix-2, trix-3


any comments?

jinevening:
@seanshpark I think introducing `backends` is a simple and good solution.

[...]

PR #10591, opened 2023-04-01
[one-cmds] Compile multiple backends with a single cfg file

mhs4670go:
This commit introduces 'backends' option into one-codegen section.

Related: #10537 

### DIFF
--- a/compiler/one-cmds/one-codegen
+++ b/compiler/one-cmds/one-codegen
@@ -97,42 +97,35 @@ def _get_parser(backends_list):
 
 def _verify_arg(parser, args, cfg_args):
     """verify given arguments"""
-    if oneutils.is_valid_attr(cfg_args, 'backend') and oneutils.is_valid_attr(
-            cfg_args, 'backends') and oneutils.is_valid_attr(cfg_args, 'command'):
+    cmd_backend_exist = oneutils.is_valid_attr(args, 'backend')
+    cfg_backend_exist = oneutils.is_valid_attr(cfg_args, 'backend')
+    cfg_backends_exist = oneutils.is_valid_attr(cfg_args, 'backends')
+    if cfg_backend_exist and cfg_backends_exist:
         parser.error(
             '\'backend\' option and \'backends\' option cannot be used simultaneously.')
 
     # Check if given backend from command line exists in the configuration file
-    if oneutils.is_valid_attr(args, 'backend') and oneutils.is_valid_attr(
-            cfg_args, 'backend'):
+    if cmd_backend_exist and cfg_backend_exist:
         if args.backend != cfg_args.backend:
             parser.error('Not found the command of given backend')
 
     # check if required arguments is given
     missing = []
-    if not oneutils.is_valid_attr(args, 'backend') and not oneutils.is_valid_attr(
-            args, 'backends') and not oneutils.is_valid_attr(
-                cfg_args, 'backend') and not oneutils.is_valid_attr(cfg_args, 'backends'):
+    if not cmd_backend_exist and not cfg_backend_exist and not cfg_backends_exist:
         missing.append('-b/--backend')
     if len(missing):
         parser.error('the following arguments are required: ' + ' '.join(missing))
 
-    if oneutils.is_valid_attr(cfg_args, 'backends'):
-        backends = getattr(cfg_args, 'backends').split(',')
+    if cfg_backends_exist:
+        cfg_backends = getattr(cfg_args, 'backends').split(',')
         # check if commands of given backends exist
-        for b in backends:
+        for b in cfg_backends:
             if not oneutils.is_valid_attr(cfg_args, b):
                 parser.error('Not found the command for ' + b)
 
         # Check if given backend from command line exists in the configuration file
-        if oneutils.is_valid_attr(args, 'backend') and oneutils.is_valid_attr(
-                cfg_args, 'backends'):
-            exists = False
-            for b in backends:
-                if b == args.backend:
-                    exists = True
-                    break
-            if not exists:
+        if cmd_backend_exist:
+            if args.backend not in cfg_backends:
                 parser.error('Not found the command of given backend')
 
 
@@ -164,22 +157,6 @@ def _parse_arg(parser):
     return codegen_args, backend_args, unknown_args
 
 
-'''
-one-codegen defines its behavior for below cases.
-
-[1] one-codegen -h
-[2] one-codegen -v
-[3] one-codegen -C ${cfg} (backend, command key in cfg)
-[4] one-codegen -C ${cfg} (backends key in cfg)
-[5] one-codegen -b ${backend} ${command}
-[6] one-codegen -b ${backend} -- ${command}
-[7] one-codegen -b ${backend} -C {cfg} (backend, command key in cfg)
-[8] one-codegen -b ${backend} -C {cfg} (backends key in cfg)
-
-All other cases are not allowed or an undefined behavior.
-'''
-
-
 def main():
     # get backend list
     backends_list = _get_backends_list()
@@ -197,7 +174,21 @@ def main():
 
     # verify arguments
     _verify_arg(parser, args, cfg_args)
+    '''
+    one-codegen defines its behavior for below cases.
 
+    [1] one-codegen -h
+    [2] one-codegen -v
+    [3] one-codegen -C ${cfg} (backend, command key in cfg)
+    [4] one-codegen -C ${cfg} (backends key in cfg)
+    [5] one-codegen -b ${backend} ${command}
+    [6] one-codegen -b ${backend} -- ${command}
+    [7] one-codegen -b ${backend} -C {cfg} (backend, command key in cfg)
+    [8] one-codegen -b ${backend} -C {cfg} (backends key in cfg) (Only 'backend' is invoked, 
+         even though cfg file has multiple backends)
+
+    All other cases are not allowed or an undefined behavior.
+    '''
     if oneutils.is_valid_attr(args, 'config'):
         assert (not backend_args or not unknown_args)
         # [7], [8]
--- a/compiler/one-cmds/one-codegen
+++ b/compiler/one-cmds/one-codegen
@@ -27,6 +27,7 @@ import ntpath
 import os
 import sys
 import shutil
+from types import SimpleNamespace
 
 import onelib.utils as oneutils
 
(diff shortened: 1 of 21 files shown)


mhs4670go:
If it is difficult to review the codes, focusing on the commands in the tests would be helpful.

compiler/one-cmds/one-codegen
```diff
@@ -94,15 +95,46 @@ def _get_parser(backends_list):
     return parser
 
 
-def _verify_arg(parser, args):
+def _verify_arg(parser, args, cfg_args):
     """verify given arguments"""
+    if oneutils.is_valid_attr(cfg_args, 'backend') and oneutils.is_valid_attr(
+            cfg_args, 'backends') and oneutils.is_valid_attr(cfg_args, 'command'):
+        parser.error(
+            '\'backend\' option and \'backends\' option cannot be used simultaneously.')
```
mhs4670go, line 103: I make it throw an error when the user uses 'backends' and 'backend' key simultaneously in cfg file for simplicity. Otherwise, the logic would be too complicated.. (maybe already it is complicated 😢)

compiler/one-cmds/one-codegen
```diff
@@ -94,15 +95,46 @@ def _get_parser(backends_list):
...
+    # Check if given backend from command line exists in the configuration file
+    if oneutils.is_valid_attr(args, 'backend') and oneutils.is_valid_attr(
+            cfg_args, 'backend'):
+        if args.backend != cfg_args.backend:
+            parser.error('Not found the command of given backend')
+
     # check if required arguments is given
     missing = []
-    if not oneutils.is_valid_attr(args, 'backend'):
+    if not oneutils.is_valid_attr(args, 'backend') and not oneutils.is_valid_attr(
```
seanshpark, line 113: (optional) give `oneutils.is_valid_attr(args, 'backend')` others a flag. use the flags as to some human readable, meaningful function or variable. use this so that it is more understandable.

compiler/one-cmds/one-codegen
```diff
@@ -141,29 +189,56 @@ def main():
...
-        raise FileNotFoundError(backend_base + ' not found')
-    codegen_cmd = [codegen_path] + backend_args + unknown_args
-    if oneutils.is_valid_attr(args, 'command'):
-        codegen_cmd += getattr(args, 'command').split()
-
-    # run backend driver
-    oneutils.run(codegen_cmd, err_prefix=backend_base)
+    _verify_arg(parser, args, cfg_args)
+
+    if oneutils.is_valid_attr(args, 'config'):
```
seanshpark, line 201: maybe add case comment above this line?

compiler/one-cmds/tests/onecc_046.test
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
```
seanshpark, line 3: 2023 ?

compiler/one-cmds/tests/onecc_047.test
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+# Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
```
seanshpark, line 3: 2023 ?

compiler/one-cmds/tests/onecc_048.test
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+# Copyright (c) 2021 Samsung Electronics Co., Ltd. All Rights Reserved
```
seanshpark, line 3: 2023 ?

compiler/one-cmds/one-codegen
```diff
@@ -94,15 +95,46 @@ def _get_parser(backends_list):
...
+
+        # Check if given backend from command line exists in the configuration file
+        if oneutils.is_valid_attr(args, 'backend') and oneutils.is_valid_attr(
+                cfg_args, 'backends'):
+            exists = False
+            for b in backends:
+                if b == args.backend:
+                    exists = True
+                    break
+            if not exists:
```
jinevening, line 135: It may be simplified as below.

```suggestion
            if args.backend not in backends:
```

(optional) better to rename backends to cfg_backends, to distinguish it from cmd_backend.

compiler/one-cmds/onelib/CfgRunner.py
```diff
@@ -109,6 +114,8 @@ def run(self, working_dir, verbose=False):
                 options += ['-O', self.opt]
             if verbose:
                 options.append('--verbose')
+            if section == 'one-codegen' and self.backend:
```
jinevening, line 117: Not for this PR, I think it would be better to split CfgRunner (maybe implement a different version for each driver), because it is doing too many jobs..

compiler/one-cmds/tests/onecc_046.test
```diff
@@ -0,0 +1,57 @@
...
+  echo "${filename_ext} FAILED"
+  rm -rf ../bin/dummy-compile
+  exit 255
+}
+
+trap trap_err_onexit ERR
+
+configfile="onecc_046.cfg"
+outputfile="sample.tvn"
+outputfile2="sample2.tvn"
```
jinevening, line 33: It would be better to use unique names to avoid conflicts with other tests.

```suggestion
outputfile="onecc_046.tvn"
outputfile2="onecc_046.2.tvn"
```

compiler/one-cmds/tests/onecc_046.cfg
```diff
@@ -0,0 +1,7 @@
+[onecc]
+one-codegen=True
+
+[one-codegen]
+backends=dummy,dummyV2
+dummy=-o sample.tvn inception_v3.onecc_046.circle
+dummyV2=-O sample2.tvn inception_v3.onecc_046.circle
```
jinevening, line 7:
```suggestion
dummy=-o onecc_046.tvn inception_v3.onecc_046.circle
dummyV2=-O onecc_046.2.tvn inception_v3.onecc_046.circle
```

compiler/one-cmds/one-codegen
```diff
@@ -132,6 +164,22 @@ def _parse_arg(parser):
...
+one-codegen defines its behavior for below cases.
+
+[1] one-codegen -h
+[2] one-codegen -v
+[3] one-codegen -C ${cfg} (backend, command key in cfg)
+[4] one-codegen -C ${cfg} (backends key in cfg)
+[5] one-codegen -b ${backend} ${command}
+[6] one-codegen -b ${backend} -- ${command}
+[7] one-codegen -b ${backend} -C {cfg} (backend, command key in cfg)
+[8] one-codegen -b ${backend} -C {cfg} (backends key in cfg)
```
jinevening, line 177: When I was reviewing other codes, I found that [8] has the following behavior.

```suggestion
[8] one-codegen -b ${backend} -C {cfg} (backends key in cfg)
    NOTE Only 'backend' is invoked, even though cfg file has multiple backends.
```

It would be better to add the comment, because it may not be intuitive for some users (or developers). My personal preference is to block using both `-b` and `-C`, but let's go as-is if you think it's good. I leave this comment just want to be careful, because UI is difficult to change after it is released..

compiler/one-cmds/one-codegen
jinevening: Reducing the number of cases looks good to me.

compiler/one-cmds/one-codegen
mhs4670go: Yeah, I'll add a comment. Actually, [8] is the main request from https://github.com/Samsung/ONE/issues/10537..

mhs4670go:
@seanshpark @jinevening 
I've changed the codes according to the reviews. PTAL:)

compiler/one-cmds/one-codegen
```diff
@@ -94,15 +95,40 @@ def _get_parser(backends_list):
     return parser
 
 
-def _verify_arg(parser, args):
+def _verify_arg(parser, args, cfg_args):
     """verify given arguments"""
+    cmd_backend_exist = oneutils.is_valid_attr(args, 'backend')
+    cfg_backend_exist = oneutils.is_valid_attr(cfg_args, 'backend')
+    cfg_backends_exist = oneutils.is_valid_attr(cfg_args, 'backends')
+    if cfg_backend_exist and cfg_backends_exist and oneutils.is_valid_attr(
```
jyoungyun, line 103: (Q) Can't we use `backend` and `backends` options at the same time in cfg?

compiler/one-cmds/one-codegen
```diff
@@ -94,15 +95,40 @@ def _get_parser(backends_list):
...
 
+    if cfg_backends_exist:
+        cfg_backends = getattr(cfg_args, 'backends').split(',')
+        # check if commands of given backends exist
+        for b in cfg_backends:
+            if not oneutils.is_valid_attr(cfg_args, b):
+                parser.error('Not found the command for ' + b)
+
+        # Check if given backend from command line exists in the configuration file
+        if cmd_backend_exist and cfg_backends_exist:
```
jyoungyun, line 128:
```suggestion
        if cmd_backend_exist:
```

You have already checked this term.

compiler/one-cmds/one-codegen
```diff
@@ -141,29 +167,70 @@ def main():
...
-    # run backend driver
-    oneutils.run(codegen_cmd, err_prefix=backend_base)
+    _verify_arg(parser, args, cfg_args)
+    '''
+    one-codegen defines its behavior for below cases.
+
+    [1] one-codegen -h
+    [2] one-codegen -v
+    [3] one-codegen -C ${cfg} (backend, command key in cfg)
+    [4] one-codegen -C ${cfg} (backends key in cfg)
```
jyoungyun, line 184: (Q) Is there no `command` key in cfg? Then what command does it use?

jyoungyun: Ah, I understood. Each backend key has its own command.

compiler/one-cmds/one-codegen
mhs4670go: Yeah, I intentionally made it impossible to make the implementation simple. Otherwise, it'd too complicated. I think there is no need to use them at the same time.

jyoungyun: (Q) Is there no case to share command key between multiple backends?

compiler/one-cmds/one-codegen
jyoungyun: (Added)

Now we support multiple backends. If the `-b` and `-C` options cannot be used together, each backend must have multiple cfg files. So, I proposed to introduce a `--backend` option with a `-C` option.

compiler/one-cmds/one-codegen
mhs4670go: Which case do you think it better be supported? As I said offline, I think we don't need to support such cases.

seanshpark (approved):
LGTM

mhs4670go:
@jinevening @jyoungyun PTAL:)

compiler/one-cmds/one-codegen
```diff
@@ -94,15 +95,40 @@ def _get_parser(backends_list):
...
 
 
-def _verify_arg(parser, args):
+def _verify_arg(parser, args, cfg_args):
     """verify given arguments"""
+    cmd_backend_exist = oneutils.is_valid_attr(args, 'backend')
+    cfg_backend_exist = oneutils.is_valid_attr(cfg_args, 'backend')
+    cfg_backends_exist = oneutils.is_valid_attr(cfg_args, 'backends')
+    if cfg_backend_exist and cfg_backends_exist and oneutils.is_valid_attr(
+            cfg_args, 'command'):
```
jinevening, line 104: Maybe this would be correct? IIUC, this if statement checks the use of `backend` and `backends` in cfg file, without consideration of command line.

```suggestion
    if cfg_backend_exist and cfg_backends_exist:
```

compiler/one-cmds/tests/onecc_neg_028.cfg may be related to this change.

compiler/one-cmds/tests/onecc_neg_031.workflow.json
```diff
@@ -0,0 +1,29 @@
...
+    "codegen_wf": {
+        "steps": [
+            "import_tf",
+            "codegen"
+        ],
+        "import_tf": {
+            "one-cmd": "one-import-tf",
+            "commands": {
+                "input_path": "inception_v3.pb",
+                "output_path": "inception_v3.onecc_031.circle",
```
jinevening, line 14: To avoid conflict with onecc_031 test.

```suggestion
                "output_path": "inception_v3.onecc_neg_031.circle",
```

compiler/one-cmds/tests/onecc_neg_031.workflow.json
```diff
@@ -0,0 +1,29 @@
...
+                "input_shapes": "1,299,299,3",
+                "output_arrays": "InceptionV3/Predictions/Reshape_1",
+                "converter_version": "v2"
+            }
+        },
+        "codegen": {
+            "one-cmd": "one-codegen",
+            "commands": {
+                "backend": "dummy",
+                "command": "-o sample.tvn inception_v3.onecc_031.circle"
```
jinevening, line 25:
```suggestion
                "command": "-o sample.tvn inception_v3.onecc_neg_031.circle"
```

compiler/one-cmds/one-codegen
mhs4670go: Well, both seems correct because `backend` and `command` are kinda a set. But I think your suggestion covers broader negative cases.

jinevening: Ah, I misunderstood the code. This is about `command` in cfg file, not about `cmd_backend_exist`.. Sorry for confusion.

jinevening (approved):
LGTM

mhs4670go:
@seanshpark @jinevening @jyoungyun PTAL:)

jyoungyun (approved):
LGTM

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

PR #10596, opened 2023-04-05
[luci] Reorder optimization passes

jinevening:
This performs forwarding after removing redundant ops.

---

Previous ordering
`ForwardReshapeToUnaryOp -> ForwardTransposeOp -> RemoveRedundantReshape -> RemoveRedundantTranspose -> SubstituteTransposeToReshape`

This caused a problem in the below case.

Step 0 (original): Reshape -> Transpose -> Add
Step 1 (after ForwardTransposeOp): Reshape -> Add -> Transpose
Step 2 (after SubstituteTransposeToReshape): **Reshape -> Add -> Reshape**

If we reorder optimization passes as below (move Forward passes to the end),
`RemoveRedundantReshape -> RemoveRedundantTranspose -> SubstituteTransposeToReshape -> ForwardReshapeToUnaryOp -> ForwardTransposeOp`

Step 0 (original): Reshape -> Transpose -> Add
Step 1 (after SubstituteTransposeToReshape): Reshape -> Reshape -> Add
Step 1 (after RemoveRedundantReshape): **Add**

### DIFF
--- a/compiler/luci/pass/src/CircleOptimizer.cpp
+++ b/compiler/luci/pass/src/CircleOptimizer.cpp
@@ -319,14 +319,6 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   {
     phase.emplace_back(std::make_unique<luci::FoldSparseToDensePass>());
   }
-  if (_options->query(Options::Algorithm::ForwardReshapeToUnaryOp))
-  {
-    phase.emplace_back(std::make_unique<luci::ForwardReshapeToUnaryOpPass>());
-  }
-  if (_options->query(Options::Algorithm::ForwardTransposeOp))
-  {
-    phase.emplace_back(std::make_unique<luci::ForwardTransposeOpPass>());
-  }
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/CircleOptimizer.cpp
```diff
@@ -432,6 +424,15 @@ void CircleOptimizer::optimize(loco::Graph *g) const
   {
     phase.emplace_back(std::make_unique<luci::UnrollUnidirectionalSequenceLSTMPass>());
   }
+  // Forward Reshape/Transpose after trying to remove redundant Reshape/Transpose
```
seanshpark, line 427: As this PR #10596 explains detail in main comment, how about adding
`refer https://github.com/Samsung/ONE/pull/10596` comment too?

jinevening: I've updated the comment as you suggested.

seanshpark (approved):
LGTM

Issue #10531, opened 2023-03-17
[one-quantize] Support Sum operator


jinevening:
## What

Let's support quantization of Sum operator

## Why

Our target model (GroupNorm) has the Op.

TODO
- luci-interpreter kernel
- QuantizeWithMinMax
- QuantizedModelVerifier
- FakeQuantize

jinevening:
Done.

PR #10601, opened 2023-04-06
[luci] Support Sum in fake quantization

jinevening:
This supports Sum Op in fake quantization.

---
Related to: https://github.com/Samsung/ONE/issues/10531


mhs4670go (approved):
LGTM

Issue #10604, opened 2023-04-06
[onert] Make it possible for `onert` to store minmax


glistening:
It is just a feature addition to `onert`.
It is not supposed to replace original record-minmax from Compile FE.

glistening:
At first, I introduced new execution event system — ExecutionEvent{Emitter,Listener}.
But based on @hseok-oh suggestion, I will reuse the existing execution event system — ExecutionObserve{r,e}.

PR #10606, opened 2023-04-06
[onert] Define minmax path in Config.lst and CompilerOptions

glistening:
It adds an option to enable recordminmax.
It introduces minmax_filepath in CompilerOptions.
It checks recordminmax can be enabled under given configuration.

Related: #10604

1st commit from Inner Repo's Draft PR

### Example Run
```
# It checks executor is linear executor or not.
$ EXECUTOR=DataFlow MINMAX_FILEPATH=minmax.h5 Product/x86_64-linux.debug/out/bin/onert_run pow/
Package Filename pow/
Error during model prepare : Recording minmax works only with Linear executor
```


hseok-oh (approved):
LGTM

chunseoklee (approved):

PR #10607, opened 2023-04-06
[onert] Introduce ExecutionEvent{Emitter,Listener}

glistening:
Let's introduce ExecutionListener/Emitter for recording minmax.
They are named as Listener/Emitter because ExecutorObserve{r,e} are in use.

Related: https://github.com/Samsung/ONE/issues/10604

2nd commit from Inner Repo's Draft PR

### DIFF
--- a/runtime/onert/core/src/exec/ExecutionEventEmitter.cc
+++ b/runtime/onert/core/src/exec/ExecutionEventEmitter.cc
@@ -27,11 +27,12 @@ void ExecutionEventEmitter::addListener(std::unique_ptr<IExecutionEventListener>
 }
 
 void ExecutionEventEmitter::notifyExecuteOpEnd(ir::SubgraphIndex s, ir::OperationIndex o,
-                                               ir::OpCode opcode, backend::ITensor *tensor)
+                                               ir::OpCode opcode, backend::ITensor **tensors,
+                                               uint32_t ntensors)
 {
   for (auto &l : _listeners)
   {
-    l->handleExecuteOpEnd(s, o, opcode, tensor);
+    l->handleExecuteOpEnd(s, o, opcode, tensors, ntensors);
   }
 }
 
--- a/runtime/onert/core/src/exec/ExecutionEventEmitter.cc
+++ b/runtime/onert/core/src/exec/ExecutionEventEmitter.cc
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


glistening:
Fix `onert-android-release` build.
It does not happen in my android ndk version. Maybe some version of ndk is more strict.

hseok-oh:
It looks these classes will be used for minmax dump. Then why you use operation index, not tensor index?

glistening:
ExecutionEvent{Event,Listener} is not for only minmax-recording.

I would like to provide general execution event hook system like existing `ExecutionObservee`.

```cc
  void notifyJobBegin(IExecutor *executor, ir::SubgraphIndex subg_ind, ir::OperationIndex op_ind,
                      const backend::Backend *backend);
  void notifyJobEnd(IExecutor *executor, ir::SubgraphIndex subg_ind, ir::OperationIndex op_ind,
                    const backend::Backend *backend);
```

But new one can read the output Tensor's value.

```cc
  virtual void notifyExecuteOpEnd(ir::SubgraphIndex, ir::OperationIndex, ir::OpCode,
                                  backend::ITensor *) = 0;
```

From your comment, I've found that the signature must be changed for multiple outputs.

```cc
  virtual void notifyExecuteOpEnd(ir::SubgraphIndex, ir::OperationIndex, ir::OpCode,
                                  backend::ITensor **, int num_outputs) = 0;
```

hseok-oh:
>From your comment, I've found that the signature must be changed for multiple outputs.

Then, can't we call new notifier for each output tensor on Executor?

`ExecutionObserver` and `ExecutionObservee` were introduced to hook operator (or graph) execution start and end. IMHO, for minmax recording, we need to introduce new observer and observee to hook tensor update.

glistening:
@hseok-oh 

> we need to introduce new observer and observee to hook tensor update. (or add notify function to observee and handle function to observer interface)

ExecutionEvent{Emitter,Listener} are new observer and observee to hook tensor update. Tensors are updated after op end.

Could you please let me know what you mean? If possible, in code level?

hseok-oh:
At first, I want to know why you want to dump operation index, not tensor index. I'm not against your implementation. I just want to know your implentation intent. Is it essential requirement to use minmax data? I didn't show code level because this question is not cleared.

glistening:
For getting minmax only (especially for percentile), it may be better to get only tensor index. But when I introduce new ExecutionEvent, I thought other applications also. For example, dumping the operator name (and op idx) and the output tensor values. It may help to debug incorrect values.

We may introduce tensor-idx based event system later for some reason. For example, we may need other approach to support `while`/`if`.

----

I will update this PR to change the signature as I wrote in https://github.com/Samsung/ONE/pull/10607#issuecomment-1499893984.

hseok-oh (approved):

PR #10611, opened 2023-04-07
[luci] Fix bug in ConvertNCHWToNHWCPass

jinevening:
This clones a constant when expanding it to rank 4.


jinevening:
To reviewer:

Previous code did in-place update, which causes a problem when the updated node is used by multiple operators.

This PR copies the target node before update, so that the change is only applied to the specific operator.

seanshpark (approved):
LGTM

seanshpark:
Is it possible to add a recipe for regression test with dredd test?

mhs4670go (approved):
LGTM

PR #10618, opened 2023-04-10
[circle-opselector] Include original output

jinevening:
This includes original output to the generated model.

### DIFF
--- a/compiler/circle-opselector/src/OpSelector.cpp
+++ b/compiler/circle-opselector/src/OpSelector.cpp
@@ -94,8 +94,7 @@ public:
   bool visit(const luci::CircleNode *) final { return false; }
 };
 
-std::unique_ptr<loco::Graph> make_graph(const std::vector<const luci::CircleNode *> nodes,
-                                        loco::Graph *g)
+std::unique_ptr<loco::Graph> make_graph(const std::vector<const luci::CircleNode *> nodes)
 {
   auto graph = loco::make_graph();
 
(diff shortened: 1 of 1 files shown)


compiler/circle-opselector/src/OpSelector.cpp
```diff
@@ -169,8 +173,19 @@ std::unique_ptr<loco::Graph> make_graph(const std::vector<const luci::CircleNode
...
+      if (std::find(original_outputs.begin(), original_outputs.end(), o) != original_outputs.end())
+      {
+        originalOutput = true;
+        break;
+      }
+    }
+
     // the node isn't graph output if it is an other node's output
-    if (beingUsed)
+    if (beingUsed and not originalOutput)
```
jinevening, line 188: Previous code does not create output if it is used by another node. This PR fixes the behavior.

compiler/circle-opselector/src/OpSelector.cpp
```diff
@@ -156,6 +157,9 @@ std::unique_ptr<loco::Graph> make_graph(const std::vector<const luci::CircleNode
       }
     }
   }
+
+  const auto original_outputs = loco::output_nodes(g);
```
mhs4670go, line 161: The nodes already have its graph.
```suggestion
  const auto original_outputs = loco::output_nodes(const_cast<loco::Graph *>(nodes.at(0)->graph());
```

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #10622, opened 2023-04-12
[onert] Remove BackendContexts from ExecutorBase

glistening:
BackendContexts is defined but not used.


glistening:
`BackendContexts` seems to be defined to make sure `BackendContexts` is not destroyed while `Executor` is alive though `ExecutorBase` itself does not access the field from .

PR #10642, opened 2023-04-14
[onert] Add MinMaxRecorder Observer

glistening:
It adds MinMaxRecorder as ExecutionObserver.

Related: #10604

### DIFF
--- a/runtime/onert/core/src/compiler/ExecutorFactory.cc
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.cc
@@ -25,7 +25,9 @@
 #include "../exec/ExecTime.h"
 #include "../exec/ExecutionObservers.h"
 #include "../exec/LinearExecutor.h"
+#ifdef MINMAX_H5DUMPER
 #include "../exec/MinMaxRecorder.h"
+#endif
 #include "../exec/ParallelExecutor.h"
 #include "../ir/OperationCloner.h"
 
@@ -433,11 +435,6 @@ exec::IExecutor *ExecutorFactory::createLinearExecutor(
 
   auto code_map = builder.releaseCodeMap();
 
-  auto mmr = options.minmax_filepath.empty()
-               ? nullptr
-               : std::make_unique<exec::MinMaxRecorder>(options.minmax_filepath,
-                                                        lowered_graph->graph(), backend_contexts);
-
   auto exec = new exec::LinearExecutor{std::move(lowered_graph),
                                        std::move(backend_contexts),
                                        tensor_regs,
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -432,6 +433,11 @@ exec::IExecutor *ExecutorFactory::createLinearExecutor(
 
   auto code_map = builder.releaseCodeMap();
 
+  auto mmr = options.minmax_filepath.empty()
+               ? nullptr
+               : std::make_unique<exec::MinMaxRecorder>(options.minmax_filepath,
+                                                        lowered_graph->graph(), backend_contexts);
```
glistening, line 439: @hseok-oh

- It saves reference to `backend_contexts`, which is going to `std::move`'d.
  - It would be better to borrow `backend_contexts` from `LinearExecutor` after `std::move`.
- It assumes `LinearExecutor` outlives `MinMaxRecorder`. 

I will add a commit on top of previous code.

runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -25,6 +25,7 @@
 #include "../exec/ExecTime.h"
 #include "../exec/ExecutionObservers.h"
 #include "../exec/LinearExecutor.h"
+#include "../exec/MinMaxRecorder.h"
```
glistening, line 28: Android and Tizen (lack of HDF5) cannot use `MinMaxRecorder`

runtime/onert/core/src/compiler/ExecutorFactory.cc
glistening: As I guessed, previous code gets segmentation fault because it accesses to std::move'd objects. I added fix commit.

glistening:
```
$ MINMAX_FILEPATH=iv3.h5 Product/x86_64-linux.debug/out/bin/onert_run \
--load build/install/dataset.h5.0 \
inception_v3_.circle

Model Filename inception_v3_.circle
MODEL_LOAD   takes 49.872 ms
PREPARE      takes 170.779 ms
EXECUTE      takes 6381.484 ms
- MEAN     :  6381.484 ms
- MAX      :  6381.484 ms
- MIN      :  6381.484 ms
- GEOMEAN  :  6381.484 ms

$ h5dump iv3.h5 | head -n 20
HDF5 "iv3.h5" {
GROUP "/" {
   GROUP "value" {
      GROUP "0" {
         GROUP "0" {
            GROUP "0" {
               DATASET "0" {
                  DATATYPE  H5T_IEEE_F32BE
                  DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
                  DATA {
                  (0): 0, 11.1378
                  }
               }
               DATASET "1" {
                  DATATYPE  H5T_IEEE_F32BE
                  DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
                  DATA {
                  (0): 0, 15.2343
                  }
               }
```

hseok-oh (approved):

Issue #10661, opened 2023-04-17
[luci-interpreter] Support Select


seanshpark:
Let's support `Select` from ONNX model with `Tensor.Expand`
`onnx.Where` is `circle.Select`

![image](https://user-images.githubusercontent.com/4616940/232631170-4134682d-a255-47e2-a4a5-d492a2fc4056.png)

seanshpark:
done

PR #10663, opened 2023-04-18
[luci-interpreter] Enable CircleSelect

seanshpark:
This will enable CircleSelect Op in luci-interpreter.


seanshpark:
for #10661 from  #10662

compiler/luci-interpreter/src/kernels/Select.h
```diff
@@ -0,0 +1,55 @@
...
+
+  void configure() override;
+  void execute() const override;
+
+private:
+  void evalFloat() const;
+
+private:
+  // for SelectV2
+  bool _requires_broadcast = false;
```
jinevening, line 46: It seems that `SelectV2` is a different operator from `Select`?

compiler/luci-interpreter/src/kernels/Select.cpp
```diff
@@ -0,0 +1,100 @@
...
+  if (_has_low_rank_input_condition)
+  {
+    tflite::reference_ops::RankOneSelect(condition_shape, condition_data, t_shape, t_data, e_shape,
+                                         e_data, output_shape, output_data);
+  }
+  else if (_requires_broadcast)
+  {
+    // TODO call for broadcast kernel
+    assert(false);
+  }
```
jinevening, line 91: Is there a reason not to use BroadcastSelect5DSlow?

compiler/luci-interpreter/src/kernels/Select.h
seanshpark: I assume yes from kernel source.

seanshpark: CircleSelect and CircleSelectV2 both exist.

compiler/luci-interpreter/src/kernels/Select.cpp
seanshpark: `_requires_broadcast` is added for SelectV2 which is not supported now.
would it be better to remove this for now?

jinevening: Maybe because of TF version?

compiler/luci-interpreter/src/kernels/Select.h
jinevening: For future reference, `SelectV2` has to be implemented with `Select.h` and `Select.cpp`.

jinevening (approved):
LGTM

PR #10673, opened 2023-04-18
[pota-quant-value-test] Remove unnecessary dependency

jinevening:
This removes unnecessary mio-circle04 dependency.


compiler/pota-quantization-value-test/requires.cmake
```diff
@@ -2,5 +2,5 @@ require("record-minmax")
 require("circle-quantizer")
 require("circle-tensordump")
 require("common-artifacts")
-require("mio-circle04")
+require("mio-circle05")
```
seanshpark, line 5: I think we can remove depends to `mico-circleXX` for this module if there is no direct dependency from any source.

seanshpark (approved):
LGTM thank you!

PR #10676, opened 2023-04-18
[pics] Add requires.cmake file

jinevening:
This adds requires cmake file.

---
Related to: https://github.com/Samsung/ONE/pull/10672#issuecomment-1512516362


seanshpark (approved):
LGTM

PR #10677, opened 2023-04-18
[luci-value-test] Enable Select

seanshpark:
This will enable Select Op value test.


seanshpark:
for #10661

jinevening (approved):
LGTM

PR #10678, opened 2023-04-18
[visq] Fix typos

stamalakhov:
This commit fixes some typos in visq.

I'm trying to update my draft on visq_runner for one-vscode.

### DIFF
--- a/compiler/visq/visqlib/QErrorComputer.py
+++ b/compiler/visq/visqlib/QErrorComputer.py
@@ -211,10 +211,15 @@ class TAEComputer(QErrorComputer):  #total absolute error
 class SRMSEComputer(QErrorComputer):
     def __init__(self, fp32_dir, fq_dir):
         super().__init__(fp32_dir, fq_dir)
-        self.scale_file = Path(fq_dir) / 'scales.txt'
+        if fq_dir != None:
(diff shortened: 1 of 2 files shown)


compiler/visq/visqlib/QErrorComputer.py
```diff
@@ -211,10 +211,15 @@ def run(self):
 class SRMSEComputer(QErrorComputer):
     def __init__(self, fp32_dir, fq_dir):
         super().__init__(fp32_dir, fq_dir)
-        self.scale_file = Path(fq_dir) / 'scales.txt'
+        if fq_dir != None:
```
jinevening, line 214: Is there a case when `fq_dir` is None?

stamalakhov: > Is there a case when `fq_dir` is None?

Yes.
See https://github.com/Samsung/ONE/blob/16dcb2f481ca650c6a4a49817d751ad84f3e039f/compiler/visq/visq#L364.

jinevening: Ah.. right. I think we have to change the constructor signature. Thanks!

jinevening (approved):
LGTM Thank you

PR #10681, opened 2023-04-18
[infra] Include mio-circle05 to Makefile.arm32

seanshpark:
This will include mio-circle05 to Makefile.arm32.


jinevening (approved):
LGTM Thank you

Issue #10687, opened 2023-04-19
[luci] Suppprt S64 for CircleReshape


seanshpark:
`CircleReshape` second input can be S32/S64. Mostly it's S32 but some ONNX produces S64.
Let's revise related codes to support this.

seanshpark:
done

PR #10693, opened 2023-04-19
[luci-interpreter] Enable S64 shape for Reshape

seanshpark:
This will enable S64 dtype from shape.

### DIFF
--- a/compiler/luci-interpreter/src/kernels/Reshape.cpp
+++ b/compiler/luci-interpreter/src/kernels/Reshape.cpp
@@ -17,6 +17,8 @@
 
 #include "kernels/Reshape.h"
 
+#include "kernels/Utils.h"
+
 #include <cassert>
 #include <cstring>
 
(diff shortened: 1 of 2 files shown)


seanshpark:
for #10687 from  #10688

compiler/luci-interpreter/src/kernels/Reshape.cpp
```diff
@@ -28,12 +30,26 @@ namespace kernels
...
+    {
+      shape.dim(i) = shape_data[i];
+    }
+  }
+  else if (tensor->element_type() == DataType::S64)
+  {
+    const auto *shape_data = tensor->data<int64_t>();
+    for (int i = 0; i < tensor->shape().num_elements(); ++i)
+    {
+      shape.dim(i) = static_cast<int32_t>(shape_data[i]);
```
jinevening, line 47: Our SE may complain about this downcasting? It would be better to check overflow.

seanshpark: > Our SE may complain about this downcasting?

there are already lots of codes like this in interpreter.

jinevening (approved):

PR #10694, opened 2023-04-19
[luci/import] Reshape S64 for Reshape shape

seanshpark:
This will enable to support S65 for Resape shape input.

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleReshape.cpp
+++ b/compiler/luci/import/src/Nodes/CircleReshape.cpp
@@ -30,7 +30,7 @@ bool CircleReshapeGraphBuilder::validate(const ValidateArgs &args) const
   if (args.op.outputs.size() != 1)
     return false;
 
-  // for two inputs, check if type is S32
+  // for two inputs, check if type is S32 or S64
   if (args.op.inputs.size() == 2)
   {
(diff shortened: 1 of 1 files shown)


compiler/luci/import/src/Nodes/CircleReshape.cpp
```diff
@@ -30,17 +30,16 @@ bool CircleReshapeGraphBuilder::validate(const ValidateArgs &args) const
...
+  // for two inputs, check if type is S32 or S64
   if (args.op.inputs.size() == 2)
   {
     const auto &inputs = args.op.inputs;
     const auto tensors = args.reader.tensors();
     const auto tensor_in = tensors.at(inputs.at(1));
     assert(tensor_in != nullptr);
 
-    // NOTE fix this if there is any other case
-    // TensorFlow lite and circle only supports S32
```
seanshpark, line 42: existing comment `only supports S32` is for attribute not second input. so strictly speaking, this comment is incorrect.

seanshpark:
for #10687 from  #10688

jinevening (approved):
LGTM

It would be better to modify commit message. S65 -> S64

seanshpark:
code landed with `S64` :)

PR #10655, opened 2023-04-14
[circle-mpqsolver] Implement intermediate dumping

stamalakhov:
This commit implements intermediate dumping.
It produces following files:

image

for every iteration it provides `mpq` configuration(`*.mpq.json`) and quantized circle model (`*.mpq.circle`) 
along with base errors and errors at every iteration (`errors.mpq.txt`)

Sample output is attached:
[sample_output.zip](https://github.com/Samsung/ONE/files/11233315/sample_output.zip)

Related: #10634
ONE-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

Issue #10634, opened 2023-04-13
[circle-mpqsolver] Implement saving intermediate results


stamalakhov:
### What

Implement saving intermediate results mentioned in https://github.com/Samsung/ONE/issues/10612#issuecomment-1501540011
### Why

Because users of circle-mpqsolver need detailed information about bisection runs.

stamalakhov:
done

PR #10703, opened 2023-04-20
[circle-mpqsolver] Introduce QuantizerHook

stamalakhov:
This commit introduces QuantizerHook and uses in Quantizer.

Its correctness is tested in #10655

Draft: #10655
Related: #10634 

### DIFF
--- a/compiler/circle-mpqsolver/src/core/Quantizer.h
+++ b/compiler/circle-mpqsolver/src/core/Quantizer.h
@@ -32,10 +32,24 @@ namespace core
 using LayerParam = luci::CircleQuantizer::Options::LayerParam;
 using LayerParams = std::vector<std::shared_ptr<LayerParam>>;
 
+struct QuantizerHook
+{
+  /**
+   * @brief called on successfull quantization
+   * @param module quantized module
+   */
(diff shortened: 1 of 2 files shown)


compiler/circle-mpqsolver/src/core/Quantizer.h
```diff
@@ -53,6 +67,7 @@ class Quantizer
 private:
   std::string _input_dtype = "uint8";
   std::string _output_dtype = "uint8";
+  const QuantizerHook *_hook = nullptr;
```
jinevening, line 70: _hook is const but updated by set_hook?

stamalakhov: Yes. AFAIK it means that all hooks' functions should also be const.
In case i've used:
```
const QuantizerHook *const _hook = nullptr;
```
compiler would give an error in attempt to set `_hook` to another callback.
@jinevening
Should i change it to non-const?

jinevening: Ah, I misunderstood. You don't have to change it :)

jinevening (approved):
LGTM

Issue #10635, opened 2023-04-13
[onert-micro] Enable kernel unit tests


BalyshevArtem:
### What

Let's introduce and implement unit tests for kernels part.

### Why

To cover the code with tests

### How

Let's add tests for currently supported kernels one by one:

- [x] Add  #10609
- [x] FullyConnected #10696
- [x] Conv2D  #10818
- [x] Logistic #10697 
- [x] Gather #10829
- [x] ExpandDims #10698
- [x] Reshape #10699
- [x] Less #10700
- [x] Mul #10719
- [x] MaxPool2D #10837
- [x] Concatenation https://github.com/Samsung/ONE/pull/10838
- [x] Slice https://github.com/Samsung/ONE/pull/10836
- [x] Sub #10724
- [x] Split #10839
- [x] Tanh https://github.com/Samsung/ONE/pull/10830
- [x] Softmax
- [x] While  #10849
- [x] UnidirectionalSequenceLSTM https://github.com/Samsung/ONE/pull/10831

Let's add NEG tests for kernels:

- [x] Add  #10841
- [x] FullyConnected #10841
- [x] Conv2D #10841
- [x] Logistic #10844
- [x] Gather #10844
- [x] ExpandDims  #10840
- [x] Reshape  #10840
- [x] Less   #10840
- [x] Mul #10842
- [x] MaxPool2D #10842
- [x] Concatenation #10842
- [x] Slice in progress
- [x] Sub in progress
- [x] Split in progress
- [x] Tanh in progress
- [x] Softmax
- [x] While
- [x] UnidirectionalSequenceLSTM in progress

BalyshevArtem:
https://github.com/Samsung/ONE/pull/10836#issuecomment-1568092126

PR #10719, opened 2023-04-24
[onert-micro] Add Mul unit test

BalyshevArtem:
This pr adds Mul unit test

for issue: https://github.com/Samsung/ONE/issues/10635


onert-micro/luci-interpreter/src/kernels/Mul.cpp
```diff
@@ -180,6 +180,7 @@ void execute_kernel_CircleMul(const circle::Operator *cur_op, BaseRuntimeGraph *
     }
     break;
 #ifndef DIS_QUANT
+      // TODO: check quantize Mul
```
BalyshevArtem, line 183: results were incorrect, need to find the issue and then can add tests for quantize version

SlavikMIPT (approved):
LGTM

Issue #10659, opened 2023-04-17
[luci-interpreter] Support FloorMod operator


ragmani:
## What

Let's support FloorMod operator

## Why
When converting Gather operator from onnx to circle, FloorMod operator is added to prevent index values out of data bounds.
- onnx
![Screenshot from 2023-04-17 15-07-34](https://user-images.githubusercontent.com/18050611/232423021-66f50bd7-8cdf-45a9-a892-e101556a5e9b.png)

- circle
![Screenshot from 2023-04-17 15-08-04](https://user-images.githubusercontent.com/18050611/232423030-18d67934-55b2-4dad-ab80-56b470c5954e.png)

TODO
- luci-interpreter kernel

seanshpark:
After luci-interpreter change lands, please revise `compiler/luci-value-test/test.lst` to run model value test :)

ragmani:
Done.

PR #10726, opened 2023-04-26
[luci-value-test] Enable FloorMod

ragmani:
This commit enables FloorMod Op value test

related to #10659


seanshpark (approved):
LGTM thank you!

PR #10732, opened 2023-04-28
[onert] Fix potential resource leak from trix DevContext

glistening:
There is potential memory leak. It fixes the case.

### DIFF
--- a/runtime/onert/backend/trix/DevContext.cc
+++ b/runtime/onert/backend/trix/DevContext.cc
@@ -71,6 +71,11 @@ DevContext::~DevContext()
 
 ModelID DevContext::registerModel(const std::string &model_file_path)
 {
+  if (_dev_handles.size() == 0)
+  {
+    throw std::runtime_error("No npu device is available");
+  }
+
   auto meta = getNPUmodel_metadata(model_file_path.c_str(), false);
 
   if (meta == nullptr)
@@ -91,6 +96,8 @@ ModelID DevContext::registerModel(const std::string &model_file_path)
     uint32_t model_id_at_device;
     if (registerNPUmodel(_dev_handles.at(dev_num), &file_info, &model_id_at_device) < 0)
     {
+      if (meta != nullptr)
+        free(meta);
       throw std::runtime_error("Failed to register npu model");
     }
 
@@ -102,6 +109,7 @@ ModelID DevContext::registerModel(const std::string &model_file_path)
     else
     {
       _meta_map[model_id_at_device] = _meta_map[model_id];
+      meta = nullptr;
     }
 
     _model_ids[model_id].resize(dev_num + 1);
--- a/runtime/onert/backend/trix/DevContext.cc
+++ b/runtime/onert/backend/trix/DevContext.cc
@@ -76,7 +76,8 @@ ModelID DevContext::registerModel(const std::string &model_file_path)
     throw std::runtime_error("No npu device is available");
   }
 
-  auto meta = getNPUmodel_metadata(model_file_path.c_str(), false);
+  std::unique_ptr<npubin_meta, decltype(&free)> meta(
+    getNPUmodel_metadata(model_file_path.c_str(), false), free);
 
   if (meta == nullptr)
   {
@@ -96,20 +97,17 @@ ModelID DevContext::registerModel(const std::string &model_file_path)
     uint32_t model_id_at_device;
     if (registerNPUmodel(_dev_handles.at(dev_num), &file_info, &model_id_at_device) < 0)
     {
-      if (meta != nullptr)
-        free(meta);
       throw std::runtime_error("Failed to register npu model");
     }
 
     if (dev_num == 0)
     {
       model_id = model_id_at_device;
-      _meta_map[model_id_at_device] = std::shared_ptr<npubin_meta>(meta);
+      _meta_map[model_id_at_device] = std::shared_ptr<npubin_meta>(std::move(meta));
     }
     else
     {
       _meta_map[model_id_at_device] = _meta_map[model_id];
-      meta = nullptr;
     }
 
     _model_ids[model_id].resize(dev_num + 1);


hseok-oh:
Maybe we need below patch, not this one

```diff
diff --git a/runtime/onert/backend/trix/DevContext.cc b/runtime/onert/backend/trix/DevContext.cc
index 059514878..1e02f7478 100644
--- a/runtime/onert/backend/trix/DevContext.cc
+++ b/runtime/onert/backend/trix/DevContext.cc
@@ -91,17 +91,19 @@ ModelID DevContext::registerModel(const std::string &model_file_path)
     uint32_t model_id_at_device;
     if (registerNPUmodel(_dev_handles.at(dev_num), &file_info, &model_id_at_device) < 0)
     {
+      free(meta);
       throw std::runtime_error("Failed to register npu model");
     }
 
     if (dev_num == 0)
     {
       model_id = model_id_at_device;
-      _meta_map[model_id_at_device] = std::shared_ptr<npubin_meta>(meta);
+      _meta_map[model_id_at_device] = std::shared_ptr<npubin_meta>(meta, std::free);
     }
     else
     {
       _meta_map[model_id_at_device] = _meta_map[model_id];
+      free(meta);
     }
 
     _model_ids[model_id].resize(dev_num + 1);
```

glistening:
@hseok-oh 

I think we don't need 1st free before `throw std::runtime_error("Failed to register npu model");` because the program will be destroyed at this case as I guess.


> +      _meta_map[model_id_at_device] = std::shared_ptr<npubin_meta>(meta, std::free);

I think it is `_meta_map`'s role to free `meta`. `_meta_map` should destroy `meta`.

>       _meta_map[model_id_at_device] = _meta_map[model_id];
> +      free(meta);

I am not sure it is okay to free here since `_meta_map` may use the pointer that is going to be free'd.

ragmani:
maybe?
```patch
diff --git a/runtime/onert/backend/trix/DevContext.cc b/runtime/onert/backend/trix/DevContext.cc
index 059514878..74f179fef 100644
--- a/runtime/onert/backend/trix/DevContext.cc
+++ b/runtime/onert/backend/trix/DevContext.cc
@@ -71,6 +71,11 @@ DevContext::~DevContext()
 
 ModelID DevContext::registerModel(const std::string &model_file_path)
 {
+  if (_dev_handles.size() == 0)
+  {
+    throw std::runtime_error("There is no npu device available")
+  }
+
   auto meta = getNPUmodel_metadata(model_file_path.c_str(), false);
 
   if (meta == nullptr)
@@ -91,6 +96,8 @@ ModelID DevContext::registerModel(const std::string &model_file_path)
     uint32_t model_id_at_device;
     if (registerNPUmodel(_dev_handles.at(dev_num), &file_info, &model_id_at_device) < 0)
     {
+      if (meta != nullptr)
+        free(meta);
       throw std::runtime_error("Failed to register npu model");
     }
 
@@ -98,6 +105,7 @@ ModelID DevContext::registerModel(const std::string &model_file_path)
     {
       model_id = model_id_at_device;
       _meta_map[model_id_at_device] = std::shared_ptr<npubin_meta>(meta);
+      meta == nullptr;
     }
     else
     {
```

hseok-oh (dismissed):
LGTM

hseok-oh:
I understood this on messanger communication.

glistening:
As @hseok-oh told,

The case of `_dev_handles.size() == 0` is handled from DevContext constructor.

https://github.com/Samsung/ONE/blob/c763867500fe3d80bfd1ef834990d34a81640d17/runtime/onert/backend/trix/DevContext.cc#L32-L49

Thus, the change of this PR seems optional to be more safe, but not necessary.

ragmani:
I think we should add this part as well. `registerNPUmodel()` may fail even if `_dev_handles.size() != 0`.
```patch
diff --git a/runtime/onert/backend/trix/DevContext.cc b/runtime/onert/backend/trix/DevContext.cc
index 059514878..74f179fef 100644
--- a/runtime/onert/backend/trix/DevContext.cc
+++ b/runtime/onert/backend/trix/DevContext.cc
@@ -91,6 +96,8 @@ ModelID DevContext::registerModel(const std::string &model_file_path)
     uint32_t model_id_at_device;
     if (registerNPUmodel(_dev_handles.at(dev_num), &file_info, &model_id_at_device) < 0)
     {
+      if (meta != nullptr)
+        free(meta);
       throw std::runtime_error("Failed to register npu model");
     }
 
@@ -98,6 +105,7 @@ ModelID DevContext::registerModel(const std::string &model_file_path)
     {
       model_id = model_id_at_device;
       _meta_map[model_id_at_device] = std::shared_ptr<npubin_meta>(meta);
+      meta == nullptr;
     }
     else
     {
```

glistening:
@ragmani As I wrote in https://github.com/Samsung/ONE/pull/10732#issuecomment-1526890272, it is not nexessary because the program will exit in this case. But I agree it is more safe way to make static analyzer happy.

But as I understand by our morning talk, we decide not to update for this alarm.

ragmani:
@glistening 
Just `throw std::runtime_error("Failed to register npu model");`. I didn't understand why throwing an exception guarantees that the program terminates.

glistening:
> @glistening Just `throw std::runtime_error("Failed to register npu model");`. I didn't understand why throwing an exception guarantees that the program terminates.

I don't said throwing an exception guarantees the the program terminates. Rather I told I agree freeing is the safer way. I meant the program will exit no one is likely to catch runtime_error. Our current code is very likely not to catch this runtime_error.

Anyway, I don't want to put more time here, I will update code though it looks somewhat dirty. It would be better to use smart pointer in some way.

glistening:

glistening:

glistening:
@ragmani, @hseok-oh PTAL 1st commit is safer version, 2nd commit is smart pointer version. I prefer 2nd, but if there is error, please let me know.

ragmani (approved):
LGTM
The start pointer version looks good.

runtime/onert/backend/trix/DevContext.cc
```diff
@@ -71,7 +71,13 @@ DevContext::~DevContext()
...
 ModelID DevContext::registerModel(const std::string &model_file_path)
 {
-  auto meta = getNPUmodel_metadata(model_file_path.c_str(), false);
+  if (_dev_handles.size() == 0)
+  {
+    throw std::runtime_error("No npu device is available");
+  }
+
+  std::unique_ptr<npubin_meta, decltype(&free)> meta(
+    getNPUmodel_metadata(model_file_path.c_str(), false), free);
```
ragmani, line 80: Is it OK if the type is `decltype(&free)` and the argument value is `free`?

ragmani, line 80: Is it OK if the type is `decltype(&free)` and the argument value is `free`?
Both are prefixed with `&` in [the cpp example](https://en.cppreference.com/w/cpp/memory/unique_ptr).
```c++
using unique_file_t = std::unique_ptr<std::FILE, decltype(&close_file)>;
        unique_file_t fp(std::fopen("demo.txt", "r"), &close_file);
```

hseok-oh (approved):
LGTM

PR #10733, opened 2023-04-28
[onert] Throw exception for uint32 type of array_segment and array_index

glistening:
Currently, for sparse tensor type, it uses uint16 regardless of the type of array_segments_type and array_indices_type. Thus, it would be safe to throw exception for uint32 type.


glistening:
@Samsung/one_onert PTAL

ragmani (approved):
LGTM

Issue #10692, opened 2023-04-19
Compiler FE: Support GeLU operator


jinevening:
## What

Let's support GeLU operator

TODO
- [x] tflite2circle
- [x] tfldump
- [x] circledump
- [x] res recipes
- [x] tflchef
- [x] luci/lang
- [x] luci/import
- [x] luci/service
- [x] luci/partition
- [x] luci/logex
- [x] luci/export
- [x] luci/tests
- [x] luci-interpreter
- [x] circle2circle
- [x] quantizer (verifier, fake quantization, const input)
- [x] luci-value-test
- [x] one-cmds (fuse_gelu)

seanshpark:
please apply `compiler/luci/tests/test.lst` too

jinevening:
All done.

PR #10764, opened 2023-05-08
[luci] Support mixed-precision quantization of Gelu

jinevening:
This supports mixed-precision quantization of Gelu.

---
Related to: https://github.com/Samsung/ONE/issues/10692


seanshpark (approved):
LGTM

Issue #10717, opened 2023-04-24
[one-cmds] extract common _get_backends_list


seanshpark:
duplicate `_get_backends_list` method existin `one-init`, `one-codegen` and `one-infer`

let's extract duplicates to common file.

seanshpark:
done

PR #10766, opened 2023-05-09
[one-cmds] Introduce onelib/backends

seanshpark:
This will introduce onelib/backends for common one commands backend list.


seanshpark:
for #10717 from #10723

mhs4670go (approved):
LGTM

PR #10767, opened 2023-05-10
[infra/debian] Revise one-compiler.install

seanshpark:
This will add backends.py file to one-compiler.install list.


seanshpark:
for #10717 from #10723

mhs4670go (approved):
LGTM

Issue #10777, opened 2023-05-11
[one-cmds/U22] one-import-onnx fails to execute


seanshpark:
>  This version of TensorFlow Probability requires TensorFlow version >= 2.12; Detected an installation of version 2.10.1. Please upgrade TensorFlow to proceed.

We need to fix version for `tensorflow_probability` and maybe `tensorflow_addons` too.

seanshpark:
similar issue: #9240

seanshpark:
TF 2.10.1 -> Nov 17 2022
- tensorflow_probability : 0.18.0 @ Nov Sep 13 2022
- tensorflow_addons  : 0.18.0 @ Nov Sep 21 2022

seanshpark:
problem fixed

PR #10778, opened 2023-05-11
[one-cmds] Tidy unused codes for backends list

seanshpark:
This will tidy unused codes for backends list.


seanshpark:
for #10777

seanshpark:
for #10717 from #10723

mhs4670go (approved):
LGTM

PR #10780, opened 2023-05-12
[luci/svc] Fix analysis warnings

seanshpark:
This will fix analysis warnings with types.


mhs4670go (approved):
LGTM

PR #10781, opened 2023-05-12
[luci-eval-driver] Fix for analysis report

seanshpark:
This will fix for analysis report warning.


seanshpark:
> Integer value 'num_inputs' obtained from untrusted source at EvalDriver.cpp:76 without checking its bounds is used as a loop bound at EvalDriver.cpp:97.

mhs4670go (approved):
LGTM

PR #10783, opened 2023-05-12
Fix type for MCU Pack Kernel

chunseoklee:
- Use int8_t in mcu Pack Kernel


BalyshevArtem (approved):
LGTM

PR #10791, opened 2023-05-16
[onert] Choose memory planner

ragmani:
This commit make basic::TensorBuilder choose memory planner.

### DIFF
--- a/runtime/onert/core/include/backend/basic/StaticTensorManager.h
+++ b/runtime/onert/core/include/backend/basic/StaticTensorManager.h
@@ -38,6 +38,8 @@ class StaticTensorManager
 public:
   StaticTensorManager(const std::shared_ptr<TensorRegistry> &reg,
                       DynamicTensorManager *dynamic_tensor_manager);
+  StaticTensorManager(const std::shared_ptr<TensorRegistry> &reg, const std::string planner_id,
+                      DynamicTensorManager *dynamic_tensor_manager);
   virtual ~StaticTensorManager() = default;
 
   void allocateNonconsts(void);
--- a/runtime/onert/core/include/backend/basic/TensorBuilder.h
+++ b/runtime/onert/core/include/backend/basic/TensorBuilder.h
@@ -38,6 +38,7 @@ class TensorBuilder
 {
 public:
   TensorBuilder(const std::shared_ptr<TensorRegistry> &tensor_reg);
+  TensorBuilder(const std::shared_ptr<TensorRegistry> &tensor_reg, const std::string planner_id);
 
   /**
    * @brief     Register tensor information to allocate on CPU backend
--- a/runtime/onert/core/src/backend/basic/StaticTensorManager.cc
+++ b/runtime/onert/core/src/backend/basic/StaticTensorManager.cc
@@ -35,6 +35,15 @@ StaticTensorManager::StaticTensorManager(const std::shared_ptr<TensorRegistry> &
   // DO NOTHING
 }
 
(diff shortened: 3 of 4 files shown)


ragmani:

jyoungyun (approved):
Are you going to allocate all trainer-related memory using Dump Manager?
It seems to make changes to the existing behaviour of Dump manager. If there is no needs to use `Dump` memory planner, it is okay to modify it and use it for training feature. However, if we need to keep the `Dump` memory planner, it would be good to introduce a new Memory planner.

Current code also looks good to me. :)

ragmani:
@jyoungyun 
At the beginning of our trainer development, our plan was to keep the memory of all tensors alive. So I'm going to allocate all trainer-related memory using `Dump` Manager at the beginning. And I don't have detailed plans for memory management yet. 
If you have any plan about it, you can introduce a new memory planner and apply it.

glistening:
> @jyoungyun At the beginning of our trainer development, our plan was to keep the memory of all tensors alive.

To keep the memory of _all_ tensors alive, it may be simpler like `constant`:

https://github.com/Samsung/ONE/blob/def64d6ea2b313dfdcf7c42298e8e538f453ed05/runtime/onert/core/include/backend/basic/BackendContextHelpers.h#L81-L85

where `constant` is 

https://github.com/Samsung/ONE/blob/def64d6ea2b313dfdcf7c42298e8e538f453ed05/runtime/onert/core/include/backend/basic/BackendContextHelpers.h#L60-L61

glistening:
@ragmani concerns my suggestion changes code under `backend/basic/*`. Although we are developing experimental training feature in this repo, it should not break other backends.

I agree. (Though I guess it will not break, I did not check it works.)

Then, I suggested to why not simply change 

https://github.com/Samsung/ONE/blob/48b5a631aa05cb6dd06e70c9c0381978615e7ead/runtime/onert/core/include/util/Config.lst#L27

`CPU_MEMORY_PLANNER` to `Bump` when `training` backend is chosen.

I would like to change minimally. I am not sure the changes is good way and have enough time to review since I don't understand (or remember) several parts.

I will follow the maintainer or contributor (@hseok-oh, @chunseoklee )'s opinion.

glistening:
I would like to change minimally. I am not sure the changes is good way and have enough time to review since I don't understand (or remember) several parts.

I will follow the maintainer or contributor (@hseok-oh, @chunseoklee and so on)'s opinion.

ragmani:
I would like to fix the memory planner for `train` backend as `Bump` at the beginning.
Currently, the backends(cpu, ruy, trix, ...) using `basic::TensorBuilder` support all cpu memory planners, but `train` backend does not support planners except for `Bump` yet. I think it's good to be able to give them a choice with `CONFIG' after having several memory planners to choose from for using the `train` backend.
And these changes are also minimal.

glistening:
> I would like to fix the memory planner for `train` backend as `Bump` at the beginning. Currently, the backends(cpu, ruy, trix, ...) using `basic::TensorBuilder` support all cpu memory planners, but `train` backend does not support planners except for `Bump` yet. I think it's good to be able to give them a choice with `CONFIG' after having several memory planners to choose from for using the `train` backend. And these changes are also minimal.

I think differently. For example, as you told me, your change introduces several functions thought each body is short. It would give worse score for function test coverage without adding unit test for this. But setting existing backend by changing config does not introduce any functions. I think my suggestion is smaller change.

glistening (approved):
LGTM. @ragmani  told me added functions will be covered by any tests which uses training backend.

glistening:
@chunseok or any committer, please merge this change. It has 2 approvals.

Issue #10773, opened 2023-05-11
[onecc-docker] Fix Dockerfile to match one-compiler release


Seunghui98:
## What
- Since the name of one-compiler release asset has been changed, so onecc-docker Dockerfile needs to be modified.
- What needs to change : [Dockerfile](https://github.com/Samsung/ONE/blob/bf2bea8f6d3a104f0892ba5aa583736c659189cb/compiler/onecc-docker/docker/Dockerfile#LL22C52-L22C52)
```
RUN apt-get update && apt-get install -qqy --no-install-recommends \ 
    wget \
    ca-certificates \ 
    && wget https://github.com/Samsung/ONE/releases/download/${VERSION}/one-compiler_${VERSION}_amd64.deb \ 
    && apt-get install -y ./one-compiler_${VERSION}_amd64.deb \ 
    && rm -rf /var/lib/apt/lists/* 
```

### Related issue:
 [#ONE-vscode 1553](https://github.com/Samsung/ONE-vscode/issues/1553)
- This is an issue that also appears in `ONE-vscode` related to this problem.

PR #10807, opened 2023-05-18
[onecc-docker] Change the release name of one-compiler in docker file

Seunghui98:
This commit changes the release name of one-compiler in docker file. 
The name will be changed to match the url of one-compiler release name asset.

ONE-DCO-1.0.-Signed-off-by: Seunghui Lee <shsh1004.lee@samsung.com>

Related Issue : [#10773]

### DIFF
--- a/compiler/onecc-docker/docker/Dockerfile
+++ b/compiler/onecc-docker/docker/Dockerfile
@@ -19,8 +19,8 @@ ARG VERSION
 RUN apt-get update && apt-get install -qqy --no-install-recommends \ 
     wget \
     ca-certificates \ 
-    && wget https://github.com/Samsung/ONE/releases/download/${VERSION}/one-compiler_${VERSION}_amd64.deb \ 
-    && apt-get install -y ./one-compiler_${VERSION}_amd64.deb \ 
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

seanshpark:
`.deb` package naming with `version` is from 1.22.0.
if someone wants to build 1.21.0 or lower then this change will fail the build.
is it possible to check version? ( @jyoungyun , do we need to do this version check?)

jyoungyun:
> is it possible to check version? ( @jyoungyun , do we need to do this version check?)

Yes, we need to do version check. I will implement this soon. This PR focuses on renaming the debian package name. This is because name matching does not work in this script after ONE supports multiple ubuntu OS, Bionic and Focal. So I want to fix this first and modify it so that we can get ONE-Compiler pacakge flexibly in the future.

I asked @Seunghui98 to simply change the name string.

It would be nice to create a new issue to solve version check issue.

jyoungyun (approved):
LGTM

This changing does not consider many cases. I'm planning to develop this script as a separate issue due to the lack of design in the first place.

This script changes the wrong name so that it can operate normally in the upcoming release.

Design issue related to this script will be discussed in #10811 issue.

seanshpark (approved):
LGTM

Issue #10802, opened 2023-05-18
Checklist for ONE 1.23.0 Release


mhs4670go:
### Release Action

- [x] Release Note #10810 
- [x] Update debian package changelog ( before branch out ) #10809 
- After Branch out  
  - [x] `master` branch version up #10812 #10813 
    - [x] tizen spec file
    - [x] nnfw_version.h
    - [x] document version in `/docs/conf.py`
    - [x] onert aar version
    - [x] vconone version

### Release Package

- Compiler
  - [x] x64-linux (tar.gz) for Ubuntu 18.04
     - [x] test with public models
     - [x]  test with in-hose models
  - [x] x64-linux (tar.gz) for Ubuntu 20.04
     - [x] test with public models
     - [x] test with in-hose models

### Release Criteria

- Compiler : TC line coverage (>80%), TCM(New Negitive / New Positive > 1), Svace & Coverity (critical=0), SAM (>4.0) 
  - [x] TEM coverage link : http://npuci.mooo.com:8080/job/nnfw/job/release_1.23/job/daily-compiler-coverage/
  - [x] TCM link : https://analysishub.sec.samsung.net/service/analyses/2083414
  - [x] Svace & SAM link : https://analysishub.sec.samsung.net/service/analyses/2083411
  - [x] Coverity link : https://analysishub.sec.samsung.net/service/analyses/2083413
  - [x] License Verification : verified

### Miscellaneous

- [x] nncc preset update if needed
- Security 
  - Compiler
    - [x] Svace_security : https://analysishub.sec.samsung.net/service/analyses/2083412

seanshpark:
Release Note for 1.23.0
- TBA

mhs4670go:
- Support multiple backends compilation with a single configuration file
- Upgrade Circle schema into 0.5

PR #10813, opened 2023-05-19
Update runtime version to 1.24.0

mhs4670go:
This commit updates runtime version into 1.24.0.

Related: #10802


glistening:
@mhs4670go You released one-compiler 1.23 a few days ago, and it changes `runtime` version to 1.24.0. May I ask why? As I understand, there are not much changes after 1.22.1 release.

PR #10814, opened 2023-05-19
[luci] Change types of StridedSlice fields

stamalakhov:
This PR changes types of StridedSlice fields to their 32-bit counterparts to avoid potential flow issues.


compiler/luci/service/src/ShapeInfer_StridedSlice.cpp
```diff
@@ -45,21 +45,23 @@ const loco::DataType S32 = loco::DataType::S32;
...
+  uint32 stop_indices_count = 0;
+  int32 stop_indices[kMaxDim];
+  uint32 strides_count = 0;
+  int32 strides[kMaxDim];
+
+  int32 begin_mask = 0;
+  int32 ellipsis_mask = 0;
+  int32 end_mask = 0;
+  int32 new_axis_mask = 0;
+  int32 shrink_axis_mask = 0;
```
seanshpark, line 64: these are from reference code. why do you want to change thee?

stamalakhov: @seanshpark 
because  in e.g. https://github.com/Samsung/ONE/blob/a13d56cffc2704c703b059db4c136f7d149fd2a7/compiler/luci/service/src/ShapeInfer_StridedSlice.cpp#L331 there is an assignment of 32-bit integer to 16-bit integer, so if an integer will be greater than 2^(16)/2=32768, there will be a flow issue.

stamalakhov: @seanshpark
Please see relevant part of tensorflow:
https://github.com/tensorflow/tensorflow/blob/388d952114e59a1aeda440ed4737b29f8b7c6e8a/tensorflow/lite/kernels/internal/types.h#LL972C1-L985C3.

jinevening: Could you sync the types with TF? e.g., counts are int8 and masks are uint16 in TF. It would be better to use the same type and insert overflow checks in proper places. Our basic principle is to sync with TF for TF-existing Ops.

stamalakhov: > Could you sync the types with TF?

Ahh. Ok. I see. Thank you.

seanshpark: I would suggest not to change existing codes from TF.
Current reference code is from TF 2.6.0 (or maybe TF 2.8.0)
If you must want to change them for any problems with any models, then refer TF 2.10.x codes.

stamalakhov: For 2.10.1 tensorflow `*indices` and `strides` of `StridedSliceParams` struct are changed to 32bits. Please see `types.h` attached.
[types.h.txt](https://github.com/Samsung/ONE/files/11550817/types.h.txt)

stamalakhov: done

stamalakhov: please see
https://github.com/tensorflow/tensorflow/commit/79605069321520bd8af318eef92b71070dcc8961
commit fixing `indices` to 32 bits.

jinevening (approved):
LGTM

seanshpark:
reference TF code at https://github.com/Samsung/ONE/pull/10814#discussion_r1203575694

seanshpark (approved):
LGTM thank you!

PR #10815, opened 2023-05-22
[onecc-docker] Add no-check-certificate option

jyoungyun:
This commit uses `--no-check-certificate` option when it downloads one-compiler package from GitHub Release page.


jyoungyun:
When onecc-docker tries to build Docker image using one-compiler package, it may fail depending on the user environment related to security settings. This option helps prevent certificate errors.

seanshpark (approved):
LGTM

Issue #10702, opened 2023-04-20
CFE: Add interface to expand one-import-onnx to other tools


seanshpark:
As an alternative to `onnx-tf` project I'm working on an internal project to import ONNX models.
I would like to provide a feature in `onecc` or in `one-import-onnx` to call this tool.
This tool will simply convert `.onnx` file to `.circle` file with two positional arguments.

This issue for discussion and to record flow of thoughts about this topic.

seanshpark:
from #10581

seanshpark:
What to decide
- [ ] where to put the tool so that `one-import-onnx` can understand there exist a tool?
- [ ] how to override to use this tool? always? or by option?
- [ ] ...

seanshpark:
> where to put the tool

try with `/usr/share/one/imports`

> how to override to use this tool? always? or by option?

...

seanshpark:
some imagination...

```
$ one-import-onnx --ext-list
Available ONNX import tools:
- Default (active)

(install external tool)

$ one-import-onnx --ext-list
Available ONNX import tools:
- AwesomeImportONNX
- Default (active)

$ one-import-onnx --ext-set AwesomeImportONNX
ONNX import tool changed to AwesomeImportONNX 

$ one-import-onnx --ext-list
Available ONNX import tools:
- AwesomeImportONNX (active)
- Default
```

seanshpark:
external tool calling parameters
- DRIVER ONNX CIRCLE [options ...]
- `options` can be ignored if not support but must not show error
- current options:
   - `unroll_rnn`
   - `unroll_lstm`
   - `keep_io_order`
   - `save_intermediate`
   - `experimental_disable_batchmatmul_unfold`

example)
files:
`one/import/`
- `onnx-AwesomeImportONNX.inf` (inf file)
- `AwesomeImportONNX.py` (python script to drive AwesomeImportONNX and match parameters
- `AwesomeImportONNX`

`onnx-AwesomeImportONNX.inf`
```ini
[one-import-onnx]
tool=AwesomeImportONNX.py
```

`AwesomeImportONNX.py`
```py
main():
   ...
   cmd = ['AwesomeImportONNX', file_onnx, file_circle, `--experimental_disable_batchmatmul_unfold`]
   subprocess.run(cmd, check=True) ...
   ...
```

seanshpark:
@lemmaa , can you give some feedback about above usage?

lemmaa:
[...]

PR #10827, opened 2023-05-26
[one-cmds] Support one-import-onnx-ext

seanshpark:
This will revise one-import-onnx to support running extension.


seanshpark:
for  #10702 from  #10765

mhs4670go (approved):
LGTM

PR #10832, opened 2023-05-29
Ignore missing ARM Compiler

chunseoklee:
- Do not exit with missing arm compiler for x64 configure


chunseoklee:
@BalyshevArtem Is there any reason to exit configuration without arm compiler ?

BalyshevArtem:
> @BalyshevArtem Is there any reason to exit configuration without arm compiler ?

for x64 configure (for eval-driver and google tests) we can skip arm compiler

BalyshevArtem (approved):
LGTM

PR #10835, opened 2023-05-30
[Draft][onert] Make core support training

ragmani:

### DIFF
--- a/runtime/onert/backend/train/BackendContext.cc
+++ b/runtime/onert/backend/train/BackendContext.cc
@@ -28,7 +28,7 @@ namespace backend
 namespace train
 {
 
-ITensorRegistry *BackendContext::genTensors() { return basic::genTensors(*this); }
+ITensorRegistry *BackendContext::genTensors() { return basic::genTensors(*this, _tensor_builder); }
 
 ITensorRegistry *BackendContext::genTrainingTensors()
 {
@@ -36,14 +36,14 @@ ITensorRegistry *BackendContext::genTrainingTensors()
 
   // TODO Generate training-related tensors except for gradient
 
-  return grad_tensor_registry.get();
+  return _grad_tensor_registry.get();
 }
 
 void BackendContext::genGradTensors()
 {
   const ir::train::TrainableGraph &tgraph = *trainable_graph();
   auto tensor_builder = _grad_tensor_builder;
-  auto tensor_reg = grad_tensor_registry;
+  auto tensor_reg = _grad_tensor_registry;
 
   tgraph.operands().iterate([&](const ir::OperandIndex &ind, const ir::Operand &) {
     if (external_operands().contains(ind))
@@ -69,26 +69,28 @@ void BackendContext::genGradTensors()
 
 FunctionMap BackendContext::genKernels()
 {
-  FunctionMap ret;
+  train::FunctionMap ret;
 
-  for (auto op_ind : _data.op_order)
-  {
-    auto fn_seq = kernel_gen->generate(op_ind);
-    ret.emplace_back(op_ind, std::move(fn_seq));
-  }
+  // TODO Generate TrainableSeqeunce
+  // for (auto op_ind : _tdata->op_order)
+  // {
+  //   auto fn_seq = kernel_gen->generate(op_ind);
+  //   ret.emplace_back(op_ind, std::move(fn_seq));
+  // }
 
   basic::initConsts(*this);
 
   // NOTE For memory optimization, we want to free some operand data
-  const_cast<ir::Graph &>(*_data.graph)
+  const_cast<ir::train::TrainableGraph &>(*_tdata->tgraph)
     .operands()
     .iterate([&](const ir::OperandIndex &, ir::Operand &obj) { obj.releaseData(); });
 
-  for (auto &&it : ret)
-  {
-    auto &fn_seq = it.second;
-    fn_seq->iterate([&](exec::IFunction &ifunc) { ifunc.prepare(); });
-  }
+  // TODO Enable
+  // for (auto &&it : ret)
+  // {
+  //   auto &fn_seq = it.second;
+  //   fn_seq->iterate([&](exec::IFunction &ifunc) { ifunc.prepare(); });
+  // }
 
   return ret;
 }
--- a/runtime/onert/backend/train/BackendContext.cc
+++ b/runtime/onert/backend/train/BackendContext.cc
@@ -30,6 +30,43 @@ namespace train
 
 ITensorRegistry *BackendContext::genTensors() { return basic::genTensors(*this); }
 
+ITensorRegistry *BackendContext::genTrainingTensors()
+{
+  genGradTensors();
+
+  // TODO Generate training-related tensors except for gradient
+
+  return grad_tensor_registry.get();
+}
+
+void BackendContext::genGradTensors()
+{
+  const ir::train::TrainableGraph &tgraph = *trainable_graph();
+  auto tensor_builder = _grad_tensor_builder;
+  auto tensor_reg = grad_tensor_registry;
+
+  tgraph.operands().iterate([&](const ir::OperandIndex &ind, const ir::Operand &) {
+    if (external_operands().contains(ind))
+      return;
+    // NOTE Assuming there is no layout changes (Always assume NHWC or UNKNOWN)
+    assert(tgraph.layout() != ir::Layout::NCHW);
+
+    // TODO Register TensorInfo that has gradient's shape
+    // ir::OperandInfo backend_info{obj.shape(), obj.typeInfo(), obj.info().memAllocType(),
+    //                              obj.isConstant()};
+    // tensor_builder->registerTensorInfo(ind, backend_info, ir::Layout::NHWC);
+  });
+
+  // TODO Plan tensor builds to reduce peak memory usage
+  tgraph.operands().iterate([&](const ir::OperandIndex &ind, const ir::Operand &) {
+    if (tensor_builder->isRegistered(ind))
+      tensor_builder->notifyFirstUse(ind);
+  });
+
+  // TODO Allocate tensors
+  // tensor_builder->allocate();
+}
+
 FunctionMap BackendContext::genKernels()
 {
   FunctionMap ret;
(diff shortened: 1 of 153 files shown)


runtime/onert/core/include/exec/ITrainableFunction.h
```diff
@@ -17,12 +17,15 @@
 #ifndef __ONERT_EXEC_I_TRAINABLE_FUNCTION_H__
 #define __ONERT_EXEC_I_TRAINABLE_FUNCTION_H__
 
+#include "exec/IFunction.h"
```
jyoungyun, line 20:
```suggestion
```

runtime/onert/core/include/exec/ITrainableFunction.h
```diff
@@ -17,12 +17,15 @@
...
+
 namespace onert
 {
 namespace exec
 {
 
-class ITrainableFunction
+// TODO Move in train folder
+class ITrainableFunction : public IFunction
```
jyoungyun, line 28:
```suggestion
class ITrainableFunction
```

runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -561,8 +565,23 @@ exec::IExecutor *ExecutorFactory::createTrainableExecutor(
...
-  backend::BackendContexts backend_contexts =
-    createBackendContexts(*lowered_graph, options.executor == "Linear");
+  backend::BackendContexts backend_contexts = createBackendContexts(*lowered_graph, true);
+
+  // Replace BackendContext with TrainbleBackendContext
+  // TODO Create context only once instead of replacing
+  for (auto &&pair : backend_contexts)
+  {
+    const auto backend = pair.first;
+    if (!backend->config()->supportTraining())
```
jyoungyun, line 575: Why do you check `supportTraining()` function?
I assume it is a training mode that `createTrainableExecutor` is called. So we don't need to check training mode here.
If it is not a training mode, this function should not be called.
What kinds of cases do you consider?

runtime/onert/core/src/compiler/TensorRegistries.h
```diff
@@ -52,6 +55,15 @@ class TensorRegistries
       {
         _tensor_regs.insert(tensor_reg);
       }
+
+      if (e.first->config()->supportTraining())
```
jyoungyun, line 59: I suggest to define another TensorRegistries constructor for TrainableExecutor.

runtime/onert/core/src/compiler/train/LoweredTrainableGraph.cc
```diff
@@ -71,13 +71,26 @@ void LoweredTrainableGraph::lowerGraph(const CompilerOptions &options)
...
+    [&](const ir::OperationIndex &op_ind, const ir::Operation &) {
+      const auto backend = backend_resolver->getBackend(op_ind);
+
+      // TODO Remove hardcoded
+      if (!(backend->config()->supportTraining()) &&
+          backend->config()->id() != backend::builtin::Config::ID)
+      {
+        throw std::runtime_error(backend->config()->id() + "backend does not support training");
+      }
+    });
```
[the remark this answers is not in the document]
jyoungyun: Now, I think the only way to enter the training mode is to call the `prepare_train()` function like [this](https://github.com/Samsung/ONE/pull/10825/files#diff-c62b86f62bfb800409860fb5a02724a75abbd4d95c6ce2f2833bb3353caf5288).
I'm not sure if it supports all cases for training mode. And I don't know which is better, `Config` or `API`.

runtime/onert/core/include/backend/IConfig.h
```diff
@@ -69,6 +69,8 @@ struct IConfig
   virtual bool supportPermutation() = 0;
   virtual bool supportDynamicTensor() = 0;
   virtual bool supportFP16() = 0;
+  // TODO Find a way to check whether the backend supports training at compile time
+  virtual bool supportTraining() { return false; }
```
jyoungyun, line 73: Now, I think the only way to enter the training mode is to call the `prepare_train()` function like [this](https://github.com/Samsung/ONE/pull/10825/files#diff-c62b86f62bfb800409860fb5a02724a75abbd4d95c6ce2f2833bb3353caf5288).
I'm not sure if it supports all cases for training mode. And I don't know which is better, `Config` or `API`.

runtime/onert/core/src/compiler/train/LoweredTrainableGraph.cc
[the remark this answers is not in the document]
ragmani: Yes, `ManualScheduler` schedules which backend to be used, depending on user's configuration.
If you forcefully change train backend to be used as discussed offline, we cannot need this exception handling.

runtime/onert/core/src/compiler/ExecutorFactory.cc
ragmani: There is no code that checks the training mode yet in the code implemented in this draft. This is just defensive programming. This will probably be removed after introducing training mode.

runtime/onert/core/src/compiler/TensorRegistries.h
ragmani: `BackendContexts` may contain backends such as the`builtin` backend that do not support training. I don't think it can be solved with the ways you suggested.

runtime/onert/core/include/backend/IConfig.h
ragmani: This is introduced to check whether a backend supports training. Unfortunately, we cannot check it just by checking whether a session has the intention to train.
Whether or not a backend supports training is determined in the code, but since this method is checked at runtime, I want to find a better way.

runtime/onert/core/src/compiler/ExecutorFactory.cc
ragmani: `BackendContexts` may contain backends such as the builtin backend that do not support training. So we need a way to check whether a backend in `BackendContexts` supports training.

runtime/onert/core/include/backend/IConfig.h
jyoungyun: I introduced the `prepare_train()` function. It can help to determine whether this session is trainable or not. How about it?

runtime/onert/core/src/compiler/ExecutorFactory.cc
jyoungyun: But this function is `createTrainableExecutor`. So it will be called when the session is trainable.
Let's talk about it next week. :)

ragmani: > But this function is createTrainableExecutor. So it will be called when the session is trainable.

The local variable of this function, `backend_contexts`, has `buildin` backend. `builtin` backend is a backend that one core has and is now appended by `ManualScheduler`.
I think it's good to keep the `builtin` backend in training as well.

> Let's talk about it next week. :)

Sure :)

runtime/onert/core/src/exec/train/TrainableExecutors.h
```diff
@@ -0,0 +1,75 @@
...
+   */
+  ~TrainableExecutors() = default;
+
+public:
+  void emplace(const ir::ModelIndex &model_index, const ir::SubgraphIndex &subg_index,
+               std::unique_ptr<IExecutor> exec) override;
+
+  TrainableExecutor *at(const ir::ModelIndex &model_index,
+                const ir::SubgraphIndex &subg_index) const override;
+
```
jyoungyun, line 55:
```suggestion

  TrainableExecutor *entryExecutor() const { return at(ir::ModelIndex{0}, ir::SubgraphIndex{0}); }
 
```

ragmani: I added it here.

runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
```diff
@@ -0,0 +1,73 @@
...
+KernelGenerator::KernelGenerator(const ir::train::TrainableGraph &tgraph,
+                                 const std::shared_ptr<TensorRegistry> &tensor_reg,
+                                 const std::shared_ptr<TensorRegistry> &grad_tensor_reg,
+                                 const std::shared_ptr<ExternalContext> &external_context)
+  : KernelGeneratorBase{tgraph}, _tensor_reg{tensor_reg}, _grad_tensor_reg{grad_tensor_reg},
+    _external_context(external_context)
+{
+}
+
+std::unique_ptr<exec::ITrainableFunction> KernelGenerator::generate(ir::OperationIndex ind)
```
ragmani, line 39: You can change `ITrainableFunction` to `TrainableSequence`.

runtime/onert/core/src/backend/builtin/train/BackendContext.cc
```diff
@@ -0,0 +1,95 @@
...
+FunctionMap BackendContext::genKernels()
+{
+  FunctionMap ret;
+
+  // TODO Enable
+  // for (auto op_ind : _data.op_order)
+  // {
+  //   auto tn_seq = kernel_gen->generate(op_ind);
+  //   ret.emplace_back(op_ind, std::move(tn_seq));
+  // }
```
ragmani, line 79: You can enable commented code here if `generate` returns `TrainableSequence`.

runtime/onert/backend/train/BackendContext.cc
```diff
@@ -30,6 +30,43 @@ namespace train
...
+      tensor_builder->notifyFirstUse(ind);
+  });
+
+  // TODO Allocate tensors
+  // tensor_builder->allocate();
+}
+
 FunctionMap BackendContext::genKernels()
 {
   FunctionMap ret;
```
ragmani, line 72: You can also change this to have only `TrainableSequence`, then `TrainableExecutor` will have only `TrainableSequence`.

ragmani (commented):
@jyoungyun PTAL at my commnets.

runtime/onert/core/include/ir/train/ITrainableOperation.h
```diff
@@ -0,0 +1,49 @@
...
+{
+
+struct OperationVisitor;
+
+namespace train
+{
+
+struct TrainableOperationVisitor;
+
+class ITrainableOperation : virtual public IOperation
```
ragmani, line 34: https://en.wikipedia.org/wiki/Virtual_inheritance

ragmani:
Done.

PR #10846, opened 2023-05-31
[one-cmds] Introduce dummy-onnx-ext

seanshpark:
This will introduce dummy-onnx-ext for testing one-import-onnx extension.


seanshpark:
for  #10702 , from #10845

mhs4670go (approved):
LGTM

PR #10857, opened 2023-06-02
[one-cmds] Fix minor bug in test

mhs4670go:
This commit removes dummy driver that should be removed.


seanshpark (approved):
LGTM

Issue #10796, opened 2023-05-17
[onert] Let's implement train backend operations


jyoungyun:
## What

Let's implement train backend operations.

## Why

The first goal of train backend is to train `MNIST` model.
So the related operations should be implemented.

## Tasks
- [ ] Introduce ITrainableFunction. #10795
- [ ] Add layers with ITrainableFunction.

jyoungyun:
## Target model: MNIST CNN model from Kaggle

### Model summary

```
Model: "sequential_1"
_________________________________________________________________
 Layer (type)                Output Shape              Param #   
=================================================================
 conv2d_4 (Conv2D)           (None, 28, 28, 32)        832       
                                                                 
 conv2d_5 (Conv2D)           (None, 28, 28, 32)        25632     
                                                                 
 max_pooling2d_2 (MaxPooling  (None, 14, 14, 32)       0         
 2D)                                                             
                                                                 
 dropout_3 (Dropout)         (None, 14, 14, 32)        0         
                                                                 
 conv2d_6 (Conv2D)           (None, 14, 14, 64)        18496     
                                                                 
 conv2d_7 (Conv2D)           (None, 14, 14, 64)        36928     
                                                                 
 max_pooling2d_3 (MaxPooling  (None, 7, 7, 64)         0         
 2D)                                                             
                                                                 
 dropout_4 (Dropout)         (None, 7, 7, 64)          0         
                                                                 
 flatten_1 (Flatten)         (None, 3136)              0         
                                                                 
 dense_2 (Dense)             (None, 128)               401536    
...
Total params: 484,714
Trainable params: 484,714
Non-trainable params: 0
_________________________________________________________________
```

### Train property

- total epoch : 5
- batch size : 32
- loss function : CategoricalCrossentropy
- optimizer : Adam
   - learning rate: 1.0e-4
   - beta_1 : 0.9
   - beta_2 : 0.999
   - epsilon : 1.0e-7

PR #10860, opened 2023-06-08
[onert] Implement max pool layer in train backend

jyoungyun:
This commit implements max pool layer in train backend. The forward pass calls cpu pool layer function.

Related issue: #10796


glistening (approved):
LGTM

Issue #10869, opened 2023-06-15
[one-cmds] unroll_lstm doesn't work with one-import-onnx extension


seanshpark:
`unroll_lstm` option doesn't seem to work with calling one-import-onnx extension...

let's find the reason and fix this.

seanshpark:
when calling extension, input onnx file path is used, not the legalized file.
legalized model is not saved to file, but the loaded `onnx_model` is used...

seanshpark:
done

PR #10870, opened 2023-06-15
[one-cmds] Fix input path for one-import-onnx-ext

seanshpark:
This will fix input model path for calling one-import-onnx-ext.


seanshpark:
for #10869

mhs4670go (approved):
LGTM

Issue #10871, opened 2023-06-15
[one-quantize] Propagate qparam backward in Reshape, Transpose


jinevening:
## What

Let's propagate qparam backward in Reshape and Transpose Op.

## Why

To reduce rescale.

PR #10872, opened 2023-06-15
[luci] Propagate QParam backward across more Ops

jinevening:
This propagates QParam backward across Reshape/Transpose.

---
Related to: #10871

### DIFF
--- a/compiler/luci/pass/src/PropagateQParamBackwardPass.test.cpp
+++ b/compiler/luci/pass/src/PropagateQParamBackwardPass.test.cpp
@@ -129,6 +129,119 @@ public:
   CircleOutput *output = nullptr;
 };
 
+/**
+ *  BEFORE
+ *
+ *        [Input]
+ *           |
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/PropagateQParamBackwardPass.test.cpp
```diff
@@ -165,3 +278,33 @@ TEST(PropagateQParamBackwardPassTest, subsequent_propagation)
   EXPECT_EQ(3.0, graph.input->quantparam()->scale[0]);
   EXPECT_EQ(3, graph.input->quantparam()->zerop[0]);
 }
+
+TEST(PropagateQParamBackwardPassTest, reshape)
```
jinevening, line 282: Added one positive/negative test for reshape. I didn't add tests for transpose, because qparam propagation works in the exactly same way for both Ops.

mhs4670go (approved):
LGTM

PR #10875, opened 2023-06-15
[onert] Replace object of Operations with IOperation

ragmani:
This commit replaces object of Operations with IOperation.

### DIFF
--- a/runtime/onert/backend/acl_cl/Config.cc
+++ b/runtime/onert/backend/acl_cl/Config.cc
@@ -47,7 +47,7 @@ bool Config::initialize()
   return true;
 }
 
-ir::Layout Config::supportLayout(const ir::Operation &, ir::Layout frontend_layout)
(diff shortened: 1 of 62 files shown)


glistening:
@ragmani This PR is huge (62 files changes). Could you please split this PR into smaller ones? For example, you already introduced `IOperation` class.

I guess it may be split into:
- Replace Operation with IOperation (not-related part of operations)
- Replace Operation with IOperation (operations related part)

Also, it is out of this PR. Is it possible to introduce `IOperation.h` so that `IOperation` can have its own header?

ragmani:
> This PR is huge (62 files changes). Could you please split this PR into smaller ones? For example, you already introduced IOperation class.

Unfortunately, this PR cannot be split. All parts are related to replacing Operation with IOperation.

> Is it possible to introduce IOperation.h so that IOperation can have its own header?

OK I will create a PR for it.

glistening (approved):
LGTM

hseok-oh (approved):

PR #10886, opened 2023-06-19
[luci/pass] Show node name for unknown node

seanshpark:
This will revise runtime_error message to show node name for better debugging.


jinevening (approved):
LGTM Thank you.

PR #10893, opened 2023-06-20
[onert] Encapsulate NNPkg

ragmani:
This commit prevents non-const operands from being directly modified externally.


hseok-oh (approved):
LGTM

PR #10895, opened 2023-06-20
[onert] Apply IGraph to Model

ragmani:
This commit applies IGraph to Model.
  - Apply IGraph to Model
  - Add a method to check whether the model has only one type of Graph

### DIFF
--- a/runtime/onert/core/src/compiler/MultiModelCompiler.cc
+++ b/runtime/onert/core/src/compiler/MultiModelCompiler.cc
@@ -30,6 +30,7 @@
 #include "compiler/StaticShapeInferer.h"
 
 #include <misc/string_helpers.h>
+#include <misc/polymorphic_downcast.h>
 
 namespace onert
 {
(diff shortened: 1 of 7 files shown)


ragmani:
waiting for #10896

runtime/onert/core/src/compiler/MultiModelCompiler.cc
```diff
@@ -77,7 +78,15 @@ std::shared_ptr<CompilerArtifact> MultiModelCompiler::compile(void)
...
-    _nnpkg->model(ir::ModelIndex{i})->iterate([&](const ir::SubgraphIndex &, ir::Graph &subg) {
+    if (!_nnpkg->model(ir::ModelIndex{i})->hasOnly<ir::Graph>())
+      throw std::runtime_error("MultiModelCompiler can only compile models for inference.");
+  }
+
+  for (uint16_t i = 0; i < model_count; i++)
+  {
+    _nnpkg->model(ir::ModelIndex{i})->iterate([&](const ir::SubgraphIndex &, ir::IGraph &graph) {
+      auto &subg = nnfw::misc::polymorphic_downcast<ir::Graph &>(graph);
+
```
glistening, line 89: We can live without this change. MultipleModelCompiler is created because it is about to compile multiple models.

glistening:
@ragmani I would like to remove `dynamic...` and `polymorphic...` which is not likely to happen because it is already checked. But if you would like to keep the check, I don't oppose.

jyoungyun (approved):
LGTM

PR #10903, opened 2023-06-21
[onert] Introduce CodeMap for training

ragmani:
This commit introduces CodeMap for training.

### DIFF
--- a/runtime/onert/core/include/compiler/train/TrainableCodeMap.h
+++ b/runtime/onert/core/include/compiler/train/TrainableCodeMap.h
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


runtime/onert/core/include/compiler/train/CodeMap.h
```diff
@@ -0,0 +1,54 @@
...
+#include "ir/train/ITrainableOperation.h"
+
+namespace onert
+{
+namespace compiler
+{
+namespace train
+{
+
+struct CodeAndInfo
```
glistening, line 32: @ragmani (out of curiosity) You used `Train...` prefix for most cases. But `CodeAndInfo` and `CodeMap` don't have `Train...` prefix. I know it is not mandatory since they are defined in `train` namespace. But I am just curious.

ragmani: I didn't append the prefix because I implemented it without thinking deeply about naming.
I will append the prefix here as well to make it easier to find when modifying later.

SlavikMIPT (approved):
LGTM

glistening:
@ragmani Sorry, changing class name is not enough. The filename would be better to be renamed to `TrainableCodeMap.h`, and it leads to rename header guard also... As I wrote, I was just curious. I don't object the previous names - `train::CodeMap`.

ragmani:
@glistening 
>  The filename would be better to be renamed to TrainableCodeMap.h, and it requires to rename header guard also...

Sorry, it's my mistake. I will fix it.

> As I wrote, I was just curious. I don't object the previous names - train::CodeMap.

`train::CodeMap` also looks better to me. But it would be better to proceed in the direction of adding the prefix, and then reorganize the naming with the namespace later.

glistening (approved):
LGTM

ragmani:
> train::CodeMap also looks better to me. But it would be better to proceed in the direction of adding the prefix, and then reorganize the naming with the namespace later.

@glistening 
Thinking about it again, it's easy to find files related to training in the train directory, so I think it's okay to proceed in the direction of not appending the prefix from now.
Do you think it would be better to remove the prefix that has already been merged?

glistening:
@ragmani I would like to continue to merge your current draft w/o renaming. Let's rename later.

jyoungyun (approved):
LGTM

PR #10920, opened 2023-06-22
[onert] Create trainable executors

ragmani:
This commit make ExecutorFactory create trainable executors.

### DIFF
--- a/runtime/onert/core/include/util/Set.h
+++ b/runtime/onert/core/include/util/Set.h
@@ -53,6 +53,16 @@ public:
 
 public:
   /**
+   * @brief copy assignment operator
+   */
+  Set<Element> &operator=(const Set<Element> &) = default;
+  /**
+   * @brief move assignment operator
+   */
+  Set<Element> &operator=(Set<Element> &&) = default;
+
+public:
+  /**
    * @brief Add a given element to the set
    *
    * @param e Element added
--- a/runtime/onert/core/src/compiler/ExecutorFactory.cc
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.cc
@@ -39,6 +39,14 @@
 #include <functional>
 #include <memory>
 
+#ifdef ONERT_TRAIN
+#include "../backend/builtin/train/BackendContext.h"
+#include "../exec/train/TrainableExecutor.h"
+
+#include <backend/train/TrainableBackendContext.h>
+#include <backend/train/ITrainableBackend.h>
+#endif // ONERT_TRAIN
+
 namespace onert
 {
 namespace
@@ -89,6 +97,7 @@ private:
   DeallocList _dealloc_list;
 };
 
+// TODO Unify initializeSubgraphIOTensors
 void initializeSubgraphIOTensors(compiler::ILoweredGraph &lowered_graph,
                                  const backend::BackendContexts &backend_contexts,
                                  const ir::OperandIndexSequence &indices)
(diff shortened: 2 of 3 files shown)


runtime/onert/core/src/compiler/ExecutorFactory.cc
```diff
@@ -39,6 +39,13 @@
 #include <functional>
 #include <memory>
 
+// Headers For training
```
jyoungyun, line 42: Do we need `ENABLE_ONERT_TRAIN` flag here?

ragmani: Yes, I will add it.

jyoungyun (approved):
LGTM

glistening (approved):

runtime/onert/core/include/util/Set.h
```diff
@@ -51,6 +51,16 @@ template <typename Element> class Set
...
+public:
+  /**
+   * @brief copy assignment operator
+   */
+  Set<Element> &operator=(const Set<Element> &) = default;
+  /**
+   * @brief move assignment operator
+   */
+  Set<Element> &operator=(Set<Element> &&) = default;
+
```
glistening, line 63: With this change, `util/Set` becomes assignable. I built w/o this change. Then, I've found:
```
/z/ONE/runtime/onert/core/src/compiler/ExecutorFactory.cc: In static member function ‘static onert::exec::IExecutor* onert::compiler::ExecutorFactory::createTrainableExecutor(std::unique_ptr<onert::compiler::train::LoweredTrainableGraph>, const std::shared_ptr<onert::exec::IExecutors>&, const onert::compiler::ExecutorFactoryArgs&)’:
/ONE/runtime/onert/core/src/compiler/ExecutorFactory.cc:688:63: error: use of deleted function ‘onert::util::Set<onert::util::Index<unsigned int, onert::ir::OperandIndexTag> >& onert::util::Set<onert::util::Index<unsigned int, onert::ir::OperandIndexTag> >::operator=(const onert::util::Set<onert::util::Index<unsigned int, onert::ir::OperandIndexTag> >&)’
  688 |     tdata.external_operands = std::move(data.external_operands);
      |                                                               ^
...
/ONE/runtime/onert/core/include/util/Set.h:38:35: note: ‘onert::util::Set<onert::util::Index<unsigned int, onert::ir::OperandIndexTag> >& onert::util::Set<onert::util::Index<unsigned int, onert::ir::OperandIndexTag> >::operator=(const onert::util::Set<onert::util::Index<unsigned int, onert::ir::OperandIndexTag> >&)’ is implicitly declared as deleted because ‘onert::util::Set<onert::util::Index<unsigned int, onert::ir::OperandIndexTag> >’ declares a move constructor or move assignment operator
   38 | template <typename Element> class Set
      |                                   ^~~
```

`ExecutorFactory.cc:688` seems the reason you've add this change.

I confirmed that adding copy assignment operator is enough to build 
```
  688 |     tdata.external_operands = std::move(data.external_operands);
```

Q1. is it okay to remove the move assignment operator?

Q2. (out of curiosity) Why copy assignment operator is required to support `std::move()`?

ragmani: > I confirmed that adding copy assignment operator is enough to build
```
  688 |     tdata.external_operands = std::move(data.external_operands);
```

It's not related to copy assignment operator.
`std::move` makes `data.external_operands` to rvalue.

> Q1. is it okay to remove the move assignment operator?

Why is it to be removed? I hope it's not a meaningless mention that lines are getting longer.

> Q2. (out of curiosity) Why copy assignment operator is required to support `std::move()`?

It is introduced to follow the Rule of Five in c++.

glistening: @ragmani Thank you for explanation. But as I wrote, line 688 seems require copy assignment, not move assignment. It was my 2nd question. Anyway, `util/Set` already has copy constructor and move constructor. Thus, two assignment operators needs to be added. I understood.

ragmani: >> But as I wrote, line 688 seems require copy assignment, not move assignment.

Right, `data` is a const object. So it requires copy assignment.

glistening (approved):

PR #10921, opened 2023-06-23
[luci-interpreter] Enable Tanh activation

seanshpark:
This will enable Tanh activation for calculateActivationRange method.


seanshpark:
one of our internal model has `Add-Tanh` sequence that can be fused.

seanshpark:
@nnfw-bot test nncc-release

jinevening:
I'm not sure that execution result is correct. We may need unittest for the case.

PR #10864, opened 2023-06-12
[onert] Introduce TrainableGraph

ragmani:
This commit introduces TrainableGraph that has trainable operations.

ONE-DCO-1.0-Signed-off-by: ragmani <ragmani0216@gmail.com>

PR #10927, opened 2023-06-27
[onert] Access gradient tensors in builtin KernelGenerator

ragmani:
This commit make `builtin`'s KernelGenerator can access tensors for gradient.

### DIFF
--- a/runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
+++ b/runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
@@ -32,7 +32,7 @@ KernelGenerator::KernelGenerator(const ir::train::TrainableGraph &tgraph,
                                  const std::shared_ptr<TensorRegistry> &grad_tensor_reg,
                                  const std::shared_ptr<ExternalContext> &external_context)
   : KernelGeneratorBase{tgraph}, _tensor_reg{tensor_reg}, _grad_tensor_reg{grad_tensor_reg},
-    _tensor_registries{}, _grad_tensor_registries{}, _external_context{external_context}
+    _external_context(external_context)
 {
 }
 
--- a/runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
+++ b/runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
@@ -32,7 +32,7 @@ KernelGenerator::KernelGenerator(const ir::train::TrainableGraph &tgraph,
                                  const std::shared_ptr<TensorRegistry> &grad_tensor_reg,
                                  const std::shared_ptr<ExternalContext> &external_context)
   : KernelGeneratorBase{tgraph}, _tensor_reg{tensor_reg}, _grad_tensor_reg{grad_tensor_reg},
-    _external_context(external_context)
+    _tensor_registries{}, _grad_tensor_registries{}, _external_context{external_context}
 {
 }
 
@@ -71,6 +71,14 @@ backend::ITensor *KernelGenerator::getTensor(const ir::OperandIndex &index)
 {
   // get Tensor from all tensor registries (for Permute op)
   auto ret = _tensor_registries.getITensor(index);
+  assert(ret != nullptr);
+  return ret;
+}
+
+backend::ITensor *KernelGenerator::getGradTensor(const ir::OperandIndex &index)
+{
+  // get gradient Tensor from all tensor registries (for Permute op)
+  auto ret = _grad_tensor_registries.getITensor(index);
   assert(ret != nullptr);
   return ret;
 }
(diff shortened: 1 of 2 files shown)


ragmani:

runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
```diff
@@ -32,7 +32,7 @@ KernelGenerator::KernelGenerator(const ir::train::TrainableGraph &tgraph,
                                  const std::shared_ptr<TensorRegistry> &grad_tensor_reg,
                                  const std::shared_ptr<ExternalContext> &external_context)
   : KernelGeneratorBase{tgraph}, _tensor_reg{tensor_reg}, _grad_tensor_reg{grad_tensor_reg},
-    _external_context(external_context)
+    _tensor_registries{}, _grad_tensor_registries{}, _external_context{external_context}
```
glistening, line 35: As I understand, the change of this line is not necessary. Compiler will call it for us.

SlavikMIPT (dismissed):
LGTM

runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
ragmani: I think the changes are code without any problem as I mentioned at https://github.com/Samsung/ONE/pull/10864#discussion_r1230374562.
I don't know what the problem is, but I'll change it as your opinion.

glistening: The problem is the unnecessary changes are added it requires more cost. maintenance, review, compilation time, and so on.

glistening (approved):
LGTM

runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
ragmani: `_grad_tensor_registries` is introduced in this PR. So I was trying to add initializing it. And I just added more missing initialization. Why is it unnecessary changes?

> maintenance cost

Why is there maintenance cost? I think it's more intuitive to initialize members explicitly.

> review time

This change is just just initialization...

> compilation time

Could you explain on why it requires more compilation time? I want to know about it.

> network traffic and so on.

This change is trivial to network traffic. Just some words...

glistening: @ragmani 


> Why is there maintenance cost? I think it's more intuitive to initialize members explicitly.

Everything I mentioned above is maintenance cost.

> > review time
> 
> This change is just just initialization...

Reviewer (like me) have to read the line to make sure there is nothing special is done, same thing (default behavior) that would be done automatically.

> > compilation time
> 
> Could you explain on why it requires more compilation time? I want to know about it.

Not only compiler, but also any static analysis tools need to read and parse the line.

> > network traffic and so on.
> 
> This change is trivial to network traffic. Just some words...

Yes, it is what I've said.


In fact, I don't understand why I need to explain this.
I guess you need to explain me why the redundant code needs to be reviewed and added in this repo w/o adding or modifying the default behavior.

ragmani: >>> review time
>>
>> This change is just just initialization...
>
> Reviewer (like me) have to read the line to make sure there is nothing special is done, same thing (default behavior) that would be done automatically.

Reviewers will immediately notice that it is initialized by default. Rather, not initializing newly introduced members may further confuse reviewers.

>>> compilation time
>>
>> Could you explain on why it requires more compilation time? I want to know about it.
>
> Not only compiler, but also any static analysis tools need to read and parse the line.

I wonder why explicitly initializing members requires more compile time and the time of any static analysis tools. It seems a lower evidence that their time increases because they read a little more line. Compilers aren't just a reader. It is unlikely that compilers spend a lot of time reading the code. Rather, explicit initialization may be more helpful to the compiler. It makes no sense to reading a little more time as a problem.
And why should we care about the time of static analysis tools? We just have to consider it's okay if there are no problems when performing them. It makes no sense to regard it as a problem.

>>> network traffic and so on.
>>
>> This change is trivial to network traffic. Just some words...
>
> Yes, it is what I've said.

Why is that a trivial traffic increases by a little code a problem? Do you really think it is a problem? So images and models shouldn't be push into the repo? It makes no sense to regard it as a problem.

> In fact, I don't understand why I need to explain this.
> I guess you need to explain me why the redundant code needs to be reviewed and added in this repo, which does not add or modify the default behavior.

I explained it at https://github.com/Samsung/ONE/pull/10864#discussion_r1230374562 and https://github.com/Samsung/ONE/pull/10927#discussion_r1244634282. And I don't think it's the redundant code.
You reviewed as if there were problems, so I want to know what the problem is. But I couldn't find any problem from what you mentioned.

PR #10931, opened 2023-06-28
[onert] Create TrainingCompiler

ragmani:
This commit makes CompilerFactory can create TrainingCompiler.

ONE-DCO-1.0-Signed-off-by: ragmani <ragmani0216@gmail.com>

PR #10932, opened 2023-06-28
[onert] Implement train_prepare nnfw api

jyoungyun:
This commit implements train_prepare nnfw api details.
It creates TrainingCompiler using the given train information.

waiting #10931

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -1144,12 +1144,32 @@ NNFW_STATUS nnfw_session::train_prepare(const nnfw_train_info *info)
     return NNFW_STATUS_INVALID_STATE;
   }
 
-  // Use default info if info is null, otherwise use given info
-  (void)info;
+  nnfw_train_info tinfo;
+  if (info != nullptr)
+  {
+    tinfo = *info;
+  }
 
-  // NYI
-  // _state = State::PREPARED_TRAINING;
-  return NNFW_STATUS_ERROR;
+  onert::compiler::train::TrainingInfo training_info;
+  training_info.setBatchSize(tinfo.batch_size);
+  // TODO Set Loss function
+
+  try
+  {
+    auto compiler =
+      onert::compiler::CompilerFactory::get().create(_nnpkg, _coptions, &training_info);
+    _nnpkg.reset();
+    _compiler_artifact = compiler->compile();
+    _execution = std::make_unique<onert::exec::Execution>(_compiler_artifact->_executors);
+  }
+  catch (const std::exception &e)
+  {
+    std::cerr << "Error during nnfw_session::train_prepare : " << e.what() << std::endl;
+    return NNFW_STATUS_ERROR;
+  }
+
+  _state = State::PREPARED_TRAINING;
+  return NNFW_STATUS_NO_ERROR;
 }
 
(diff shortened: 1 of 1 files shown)


ragmani (approved):
LGTM

hseok-oh:
You means nnfw API, not NNAPI. Right?

jyoungyun:
I updated commit message. The approval is not erased and remains the same :)

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1144,12 +1144,32 @@ NNFW_STATUS nnfw_session::train_prepare(const nnfw_train_info *info)
     return NNFW_STATUS_INVALID_STATE;
   }
 
-  // Use default info if info is null, otherwise use given info
-  (void)info;
+  if (info == nullptr)
+  {
+    std::cerr << "Error during model prepare training: info is null" << std::endl;
+    return NNFW_STATUS_UNEXPECTED_NULL;
+  }
```
hseok-oh, line 1151: Please update `nnfw_train_prepare`'s comment in `nnfw_experimental.h` if you want to return `NNFW_STATUS_UNEXPECTED_NULL` when `info` is `nullptr`.

https://github.com/Samsung/ONE/blob/f03dd297f7d235b722f7151b0e9ca248744d4f47/runtime/onert/api/include/nnfw_experimental.h#L196-L200

jyoungyun: I will update the `nnfw_experimental.h` file. If we support nullptr information, we need to keep the trainingInfo in session object. However, it is only necessary if it is training. Since we do not fully support the training function yet, it would be good to support it by putting it into the session when the train function implementation is fully included in the ONERT in the future.

ragmani (approved):

glistening (approved):
LGTM.

@hseok-oh PTAL

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1144,12 +1144,32 @@ NNFW_STATUS nnfw_session::train_prepare(const nnfw_train_info *info)
...
-  // _state = State::PREPARED_TRAINING;
-  return NNFW_STATUS_ERROR;
+  onert::compiler::train::TrainingInfo training_info;
+  training_info.setBatchSize(tinfo.batch_size);
+  // TODO Set Loss function
+
+  try
+  {
+    auto compiler =
+      onert::compiler::CompilerFactory::get().create(_nnpkg, _coptions, &training_info);
```
hseok-oh, line 1160: Now it's OK, but it's a little bit dangerous. 
We should confirm that `training_info` will not be used in `_compiler_artifact` (`_compiler_artifact->_executors`) because `training_info` will be free after this function is finished.

hseok-oh (approved):
LGTM

runtime/onert/api/src/nnfw_api_internal.cc
ragmani: How about passing by reference?

ragmani: How about passing `training_info` by const reference into `create()`?

ragmani: If you want to use `TrainingInfo` elsewhere other than `TrainingCompiler`, it's nice to have it in a session.
If not, it would be good for `TrainingCompiler` to have a copied `TrainingInfo` by passing it by const reference.

PR #10936, opened 2023-06-28
[onert] Fix missing override methods in TrainableGraph

ragmani:
This commit fixes missing override methods in TrainableGraph.

### DIFF
--- a/runtime/onert/core/src/ir/train/TrainableGraph.cc
+++ b/runtime/onert/core/src/ir/train/TrainableGraph.cc
@@ -64,6 +64,11 @@ IOIndex TrainableGraph::getOutputIndex(const std::string &name) const
   return _graph.getOutputIndex(name);
 }
 
+void TrainableGraph::changeShape(const OperandIndex &index, const ir::Shape &new_shape)
+{
+  _graph.changeShape(index, new_shape);
+}
+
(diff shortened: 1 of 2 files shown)


jyoungyun (approved):
LGTM

runtime/onert/core/src/ir/train/TrainableGraph.cc
```diff
@@ -64,6 +64,11 @@ IOIndex TrainableGraph::getOutputIndex(const std::string &name) const
   return _graph.getOutputIndex(name);
 }
 
+void TrainableGraph::changeShape(const OperandIndex &index, const ir::Shape &new_shape)
+{
+  _graph.changeShape(index, new_shape);
+}
+
```
glistening, line 71: @ragmani I think it is not the scope of this PR. Please let me know if it needs to be included.

ragmani: It is the scope of this PR. I forgot to add the definition of this virtual method in `TrainableGraph`.

glistening: I thought this PR fixes the missing `override` **keyword** in TrainableGraph. But I found the title is "Fix missing override **methods** in TrainableGraph". `changeShape` is the missing method. :)

glistening (approved):

Issue #10941, opened 2023-06-29
[visq] Use tensor identifier


jinevening:
## What

Let's introduce tensor identifier, which is a unique number for each tensor.

## Why

`visq` saves a tensor's value as a file whose name is the tensor's name, e.g., `<tensor_name>.npy`. If a tensor name is longer than max length of file name in Ubuntu (255 characters), OSError is raised.

`
ERROR: OSError: [Errno 36] File name too long: '/tmp/0/l_batch_normalization_FusedBatchNormV3_model_1_model_depthwise_conv2d_3_depthwise_model_1_model_conv2d_Conv2D1;depthwise_3;Conv2D;model_1_model_batch_normalization_FusedBatchNormV3_model_1_model_depthwise_conv2d_3_depthwise_model_1_model_conv2d_Conv2D1_y1.npy'
`

jinevening:
#10944 is landed. Now tensor data is written as <tensor_id>.npy.

PR #10944, opened 2023-06-29
[visq] Use tensor ID as a filename

jinevening:
This uses tensor ID as a filename to avoid too long filename error.

---
Related to: https://github.com/Samsung/ONE/issues/10941


seanshpark (approved):
LGTM

seanshpark (approved):
LGTM

PR #10945, opened 2023-06-29
[onert] Introduce StaticGradientShapeInfer

ragmani:
This commit introduces StaticGradientShapeInfer that infer shapes of tensors for gradient.

### DIFF
--- a/runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
+++ b/runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "StaticGradientShapeInferer.h"
+#include "util/ShapeInference.h"
+#include "util/logging.h"
+
+#include <misc/polymorphic_downcast.h>
+
+#include <sstream>
+#include <stdexcept>
+
+namespace onert
+{
+namespace compiler
+{
+namespace train
+{
+
+void StaticGradientShapeInferer::infer()
+{
+  // It is not determined to iterate in reverse order.
+  auto sorted_ops = _lowered_subg->graph().topolSortOperations();
+  for (auto it = sorted_ops.rbegin(); it != sorted_ops.rend(); ++it)
+  {
+    const auto op_idx = *it;
+    const auto &op = _lowered_subg->trainable_graph().operation(op_idx);
+    if (checkDynamicInput(op))
+    {
+      std::stringstream msg;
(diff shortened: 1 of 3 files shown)


ragmani:
wait for #11015

ragmani:

runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
```diff
@@ -0,0 +1,109 @@
...
+{
+
+void StaticGradientShapeInferer::infer()
+{
+  // It is not determined to iterate in reverse order.
+  auto sorted_ops = _lowered_subg->graph().topolSortOperations();
+  for (auto it = sorted_ops.rbegin(); it != sorted_ops.rend(); ++it)
+  {
+    const auto op_idx = *it;
+    const auto &op = _lowered_subg->trainable_graph().operation(op_idx);
```
glistening, line 40: @ragmani I guess `trainable_graph()` is the only change with existing StaticShapeInferer. Am I right?

runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
```diff
@@ -0,0 +1,109 @@
...
+      return true;
+    }
+  }
+
+  return false;
+}
+
+void StaticGradientShapeInferer::visit(const ir::train::operation::Conv2D &)
+{
+  // NYI
```
glistening, line 74: Do we need to make a duplication here?

runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
ragmani: I didn't understand what you said.

runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
ragmani: What duplication do you mean?

runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
glistening: - I guess `StaticGradientShapeInferer.cc` is almost identical to `StaticShapeInferer.cc`.
- Most code will be just a copy of `StaticGradientShapeInferer`.
- I guess the only change would be `StaticGradientShapeInferer` iterates `trainable_graph()` in reverse order.

Am I right?

runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
glistening: By duplication, I mean `// NYI` will be filled with the copy of `void StaticShapeInferer::visit(const ir::operation::Conv2D &op)`. Am I right?

runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
ragmani: No, shapes of gradient tensors are different from tensors for inference. Most code will be different from `StaticShapeInferer`.

runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
ragmani: No.

runtime/onert/core/src/compiler/train/StaticGradientShapeInferer.cc
glistening: > No, shapes of gradient tensors are different from tensors for inference. Most code will be different from `StaticShapeInferer`.

Thank you. I would like to get this information. I will see the difference from forthcoming PRs.

glistening (approved):
LGTM

jyoungyun (approved):
LGTM

Issue #10933, opened 2023-06-28
[onert-micro] Compare Operation list to TFLM


chunseoklee:
latest status, see below :  https://github.com/Samsung/ONE/issues/10933#issuecomment-2630037244

<details>

| tflite-micro        | onert-micro |  remark     |
|   --------------------------| -----------| --------|
| Abs                           |  ✅　#10946       | 　  @BalyshevArtem     |
| Add                           |  ✅                     | 　                                  |
| AddN                        | ✅　#10947    | 　 @BalyshevArtem          |
| ArgMax                    | ✅　#10948    | 　  @BalyshevArtem         |
| ArgMin                     |✅ 　 #10948    | 　  @BalyshevArtem        |
| AssignVariable         | 　 🛑                | 　   NOT SUPPORTED BY CIRCLE       |
| AveragePool2D        | ✅ #10961   | 　   @BalyshevArtem       |
| BatchToSpaceNd     | 　✅    #11620               | 　 @BalyshevArtem      |
| BroadcastArgs         | 　     🛑            | 　 NOT SUPPORTED BY CIRCLE          |
| BroadcastTo             | 　 ✅  #11633                | 　 @BalyshevArtem          |
| CallOnce                  | 　    🛑             | 　 NOT SUPPORTED BY CIRCLE          |
| Cast                         | 　 ✅    #11632              | 　 @BalyshevArtem            |
| Ceil                          | 　 ✅   #11627              | 　 @BalyshevArtem          |
| CircularBuffer          | 　     🛑            | 　   NOT SUPPORTED BY CIRCLE        |
| Concatenation         | ✅                  | 　          |
| Conv2D                   | ✅                  | 　          |
| Cos                         | 　✅    #11673               | 　   @SlavikMIPT        |
| CumSum                | 　    🛑               | 　    NOT SUPPORTED BY CIRCLE       |
| DepthToSpace       | 　   ✅     #11622             | 　  @BalyshevArtem         |
| DepthwiseConv2D | 　✅     #11556             | 　     @BalyshevArtem      |
| Dequantize             | 　✅    #11619                | 　 @BalyshevArtem          |
| DetectionPostProcess |  　🛑             | 　   NOT SUPPORTED BY CIRCLE        |
| Div                        | ✅　#10962        | 　 @BalyshevArtem             |
| Elu                        | ✅　 #10963       | 　 @BalyshevArtem            |
| Equal                    | ✅　 #10978        | 　   @BalyshevArtem          |
[...]

PR #10946, opened 2023-06-29
[onert-micro] Support Abs kernel

BalyshevArtem:
This PR adds supporting of Abs kernel.

for issue: #10933


onert-micro/luci-interpreter/src/loader/GraphLoader.cpp
```diff
@@ -25,6 +25,7 @@ bool isInplaceOperation(const circle::BuiltinOperator &op)
 {
   switch (op)
   {
+    case circle::BuiltinOperator_ABS:
     case circle::BuiltinOperator_LOGISTIC:
```
BalyshevArtem, line 29: for inplace optimizations

chunseoklee (approved):

Issue #8739, opened 2022-03-29
[Compiler FE] Convert dynamic batch to single batch


jinevening:
## What

Let's implement an option to convert a model with a dynamic batch size to a model with a single batch.

For example,

```
onecc optimize --dynamic_batch_to_single_batch -i dynamic.circle -o single.circle
```

## Why

The option will

1. enable quantization of the model
2. allow the model to be run on backends that do not support dynamic shapes

seanshpark:
`dynamic batch size`:  where shape of input is like `(-1, 224, 244, 3)` where -1 is unknown(dynamic) batch size.

jinevening:
More specifically, `shape_signature` is like (-1, 224, 224, 3). The below case is possible

shape: (1, 224, 224, 3)
shape_signature: (-1, 224, 224, 3)

jinevening:
Hmm.. there exist consistent needs for this feature from our customers.. I think it's time to support it.

jinevening:
This enables quantization of edgetpu MV1 model in MLPerf.

https://github.com/Samsung/ONE/pull/10974#issuecomment-1619517461

jinevening:
All done.

PR #10974, opened 2023-07-03
DRAFT: Convert dynamic batch to single batch

jinevening:
This adds an option to convert dynamic batch to single batch.

---
Related to: https://github.com/Samsung/ONE/issues/8739

### DIFF
--- a/compiler/circle2circle/src/Circle2Circle.cpp
+++ b/compiler/circle2circle/src/Circle2Circle.cpp
@@ -16,6 +16,7 @@
 
 #include <luci/ImporterEx.h>
 #include <luci/CircleOptimizer.h>
+#include <luci/DynamicBatchToSingleBatch.h>
 #include <luci/Service/ChangeOutputs.h>
 #include <luci/Service/Validate.h>
 #include <luci/CircleExporter.h>
@@ -172,6 +173,8 @@ int entry(int argc, char **argv)
   add_switch(arser, "--disable_validation",
              "This will turn off operator validations. May help input model investigation.");
   add_switch(arser, "--generate_profile_data", "This will turn on profiling data generation.");
+  add_switch(arser, "--dynamic_batch_to_single_batch",
+             "Convert dynamic batch size (first dimension) of inputs to 1.");
 
   arser.add_argument("--change_outputs")
     .help("Experimental: Change first subgraph output nodes to CSV names");
@@ -371,11 +374,38 @@ int entry(int argc, char **argv)
     csv_tokenize(csv_nodes, new_outputs);
   }
 
+  bool dynamic_batch_to_single_batch = false;
+  if (arser.get<bool>("--dynamic_batch_to_single_batch"))
+  {
+    dynamic_batch_to_single_batch = true;
+  }
+
   // Import from input Circle file
   luci::ImporterEx importerex;
   auto module = importerex.importVerifyModule(input_path);
   if (module.get() == nullptr)
     return EXIT_FAILURE;
+
+  // Convert dynamic batch to single batch
+  // Why here? It has to be done before 'optimize', because most optimization
+  // passes are written based on static shapes
+  if (dynamic_batch_to_single_batch)
+  {
+    luci::dynamic_batch_to_single_batch(module.get());
+
+    if (!luci::validate_shape(module.get()))
+    {
+      if (settings->get(luci::UserSettings::Key::DisableValidation))
+        std::cerr << "WARNING: Invalid module after converting dynamic batch to single batch"
+                  << std::endl;
+      else
+      {
+        std::cerr << "ERROR: Invalid module after converting dynamic batch to single batch"
+                  << std::endl;
+        return 255;
+      }
+    }
+  }
 
   if (change_outputs)
   {
--- a/compiler/luci/pass/src/DynamicBatchToSingleBatchPass.cpp
+++ b/compiler/luci/pass/src/DynamicBatchToSingleBatchPass.cpp
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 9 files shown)


compiler/circle2circle/src/Circle2Circle.cpp
```diff
@@ -371,12 +374,39 @@ int entry(int argc, char **argv)
...
     return EXIT_FAILURE;
 
+  // Convert dynamic batch to single batch
+  // Why here? It has to be done before 'optimize', because most optimization
+  // passes are written based on static shapes
+  if (dynamic_batch_to_single_batch)
+  {
+    luci::dynamic_batch_to_single_batch(module.get());
+
+    if (!luci::validate(module.get()))
```
seanshpark, line 396: We may have to extract shape validation to another method, validate is used for other features too.

compiler/luci/pass/src/DynamicBatchToSingleBatchPass.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    if (input_node->rank() == 0)
+      continue;
+
+    // Skip if batch dimension is known
+    if (input_node->dim(BATCH_DIM).known())
+      continue;
+
+    // 'set' will make the dimension known
+    input_node->dim(BATCH_DIM).set(1);
```
seanshpark, line 48: Is this really valid thing to do?

seanshpark:
Please add all the test models you have in mind.

seanshpark:
@jinevening , as you may know, this won't fix recent model problem from SRUK as it has unknown shape like `(-1, 12, -1, 64)`

jinevening:
> as you may know, this won't fix recent model problem from SRUK as it has unknown shape like (-1, 12, -1, 64)

The strange unknown shape (-1, 12, -1, 64) comes from unknown batch, so theoretically, we can fix the issue by setting the static batch size. But we may need much more constant folding passes to handle the model.

This PR is mainly for edgetpu MV1 model in MLPerf. https://github.com/mlcommons/mobile_models/blob/main/v0_7/tflite/mobilenet_edgetpu_224_1.0_float.tflite

compiler/circle2circle/src/Circle2Circle.cpp
jinevening: Looks reasonable. I'll only test shape here.

compiler/luci/pass/src/DynamicBatchToSingleBatchPass.cpp
jinevening: Could you explain your question a little bit more ? This will make dynamic batch to single batch, which is friendly to some backends.

seanshpark: I have no knowledge about doing this so the question is as wrote. We had no previous communication about this. It would be helpful if you can share any knowledge with doing like this.
If not, it is also OK, and for this, with various test models with unknown dimension would validate and help understanding what is supported and what is not. As I know, current test models don't have unknown dimension cases.

seanshpark:
> This PR is mainly for edgetpu MV1 model in MLPerf. https://github.com/mlcommons/mobile_models/blob/main/v0_7/tflite/mobilenet_edgetpu_224_1.0_float.tflite

This model didn't show any unknown dim with `tfldump` . Can you verify the model?

compiler/luci/pass/src/DynamicBatchToSingleBatchPass.cpp
seanshpark: For actual numbers, is it OK to do like this with rank 1 and 2?
For I can think is , some rank 2 model would be HW, some would have other meanings. I'm not sure unknown H would be meaningful and it is OK to fix -1 it to 1.

But, as this is compilation option, users can decide it is valid to apply this option or not.
But from previous experience, I cannot say most users do.

jinevening:
Sorry. The model in `v1_1` has a dynamic batch.

https://github.com/mlcommons/mobile_models/blob/main/v1_1/tflite/mobilenet_edgetpu_224_1.0_float.tflite

```
Operands: T(subgraph index : tensor index) TYPE (shape) (shape_signature) B(buffer index) (variable) OperandName
T(0:0) FLOAT32 (1, 224, 224, 3) (-1, 224, 224, 3) B(1) input
```

compiler/luci/pass/src/DynamicBatchToSingleBatchPass.cpp
jinevening: Let me explain this PR a bit more. This option will simply change the size of the first dimension to 1, if its size is unknown (= dynamic). I decided to use the "first" dimension, because most models are NHWC or NCHW. Of course some models have different layouts, e.g., HW as you mentioned. In that case, users should not use this option. That is why this option is not included in O1.

jinevening: @seanshpark  As discussed offline, I'll throw an exception if input has unknown dimension, but its rank is not four.

jinevening:
Now it is possible to quantize edgetpu MV1 model in MLPerf (https://github.com/mlcommons/mobile_models/blob/main/v1_1/tflite/mobilenet_edgetpu_224_1.0_float.tflite) with the below cfg file.

```ini
[onecc]
one-import-tflite=True
one-optimize=True
one-quantize=True
include=O1

[one-import-tflite]
input_path=./mobilenet_edgetpu_224_1.0_float.tflite
output_path=./mobilenet_edgetpu_224_1.0_float.circle

[one-optimize]
input_path=./mobilenet_edgetpu_224_1.0_float.circle
output_path=./mobilenet_edgetpu_224_1.0_float.opt.circle
dynamic_batch_to_single_batch=True

[one-quantize]
input_path=./mobilenet_edgetpu_224_1.0_float.opt.circle
output_path=./mobilenet_edgetpu_224_1.0_float.opt.q8.circle
granularity=channel
quantized_dtype=uint8
```

Previously, it failed in `record-minmax`, because batch size is unknown.

jinevening:
commit plan
- [ ] luci pass
- [ ] validate_shape
- [ ] luci
- [ ] circle2circle
- [ ] one-cmds

PR #10982, opened 2023-07-04
[luci] Add DynamicBatchToSingleBatchPass

jinevening:
This PR adds DynamicBatchToSingleBatch pass.

---
Related to: https://github.com/Samsung/ONE/issues/8739
Draft PR: #10974

### DIFF
--- a/compiler/luci/pass/src/DynamicBatchToSingleBatchPass.cpp
+++ b/compiler/luci/pass/src/DynamicBatchToSingleBatchPass.cpp
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/src/DynamicBatchToSingleBatchPass.cpp
```diff
@@ -0,0 +1,78 @@
...
+    // Skip if batch dimension is known
+    if (input_node->dim(BATCH_DIM).known())
+      continue;
+
+    if (input_node->rank() != 4)
+    {
+      // Limit use only for rank 4 inputs (for NHWC and NCHW)
+      // TODO Enable this if necessary
+      throw std::runtime_error("First dimension of input is unknown, but its rank is not 4.");
+    }
```
jinevening, line 52: @seanshpark Now this pass throws an exception for non-rank 4 cases.

seanshpark: Can you please add for another `rank_1_NEG` test with rank 1 input and expect throw?

seanshpark: later when we are sure rank 1 works then we can remove this new test.
maybe a note about this would help future task.

compiler/luci/pass/src/DynamicBatchToSingleBatchPass.test.cpp
```diff
@@ -0,0 +1,126 @@
...
+  {
+    input->index(0);
+    input->shape({1, 5, 5, 3});
+  }
+
+  EXPECT_FALSE(pass.run(g.get()));
+}
+
+// Remove this test if we support rank 1 in this pass
+TEST(DynamicBatchToSingleBatchPassTest, rank1_NEG)
```
jinevening, line 100: @seanshpark  I've added rank1 test. PTAL :)

seanshpark (approved):
LGTM thank you!

PR #10901, opened 2023-06-21
[Draft][tools] Introduce generate_datafiles for onert_train

jyoungyun:
This commit introduces generate_datafile.py script to create multiple input and expected data files for onert_train. It includes some modification to onert_run tool to save the input data file.

ONE-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

Related PR: #10900

PR #9016, opened 2022-04-22
[nnpackage_run] Support multiple input for load:raw option

glistening:
nnpackage_run can accept multiple inputs.

--load:raw in

will get input raw data from in.0, in.1, in.2,..., in.{N-1}
where N is number of inputs, determined by nnfw info query api.

Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>

### Related Issue
#9012

### For Reviewers

```
$ tree add.q8
add.q8
├── add.q8.circle
├── add.q8.expected.0
├── add.q8.in.0
├── add.q8.in.1
├── add.q8.tvn
└── metadata
    └── MANIFEST

$ BACKENDS="trix;cpu" ./Product/x86_64-linux.debug/out/bin/nnpackage_run \
--load:raw add.q8/add.q8.in \
--dump:raw add.q8/add.q8.out \
add.q8

( ... )

add.q8/add.q8.out.0 is generated.
===================================
MODEL_LOAD   takes 0.959 ms
PREPARE      takes 6.088 ms
EXECUTE      takes 331.659 ms
- MEAN     :  331.659 ms
- MAX      :  331.659 ms
- MIN      :  331.659 ms
- GEOMEAN  :  331.659 ms
===================================

$ diff add.q8/add.q8.{expected,out}.0 
$ echo $?
0
```

PR #10988, opened 2023-07-04
[tools/onert_train] Add input and expected datafile arguments

jyoungyun:
This commit adds input and expected datafile arguments and data_length. The onert_train will train a given model using input and expected datafile.

Draft PR: #10900

### DIFF
--- a/tests/tools/onert_train/src/args.cc
+++ b/tests/tools/onert_train/src/args.cc
@@ -150,6 +150,20 @@ void Args::Initialize(void)
     }
   };
 
+  auto process_load_raw_inputfile = [&](const std::string &input_filename) {
+    _load_raw_input_filename = input_filename;
+
+    std::cerr << "Model Input Filename " << _load_raw_input_filename << std::endl;
+    checkModelfile(_load_raw_input_filename);
+  };
+
+  auto process_load_raw_expectedfile = [&](const std::string &expected_filename) {
+    _load_raw_expected_filename = expected_filename;
+
+    std::cerr << "Model Expected Filename " << _load_raw_expected_filename << std::endl;
+    checkModelfile(_load_raw_expected_filename);
+  };
+
   auto process_output_sizes = [&](const std::string &output_sizes_json_str) {
     Json::Value root;
     Json::Reader reader;
@@ -184,6 +198,17 @@ void Args::Initialize(void)
     ("nnpackage", po::value<std::string>()->notifier(process_nnpackage), "NN Package file(directory) name")
     ("modelfile", po::value<std::string>()->notifier(process_modelfile), "NN Model filename")
     ("path", po::value<std::string>()->notifier(process_path), "NN Package or NN Modelfile path")
+    ("data_length", po::value<int>()->default_value(-1)->notifier([&](const auto &v) { _data_length = v; }), "Data length number")
+    ("load_input:raw", po::value<std::string>()->notifier(process_load_raw_inputfile),
+         "NN Model Raw Input data file\n"
+         "The datafile must have data for each input number.\n"
+         "If there are 3 inputs, the data of input0 must exist as much as data_length, "
+         "and the data for input1 and input2 must be held sequentially as data_length.\n"
+    )
+    ("load_expected:raw", po::value<std::string>()->notifier(process_load_raw_expectedfile),
+         "NN Model Raw Expected data file\n"
+         "(Same data policy with load_input:raw)\n"
+    )
     ("epoch", po::value<int>()->default_value(5)->notifier([&](const auto &v) { _epoch = v; }), "Epoch number (default: 5)")
     ("batch_size", po::value<int>()->default_value(32)->notifier([&](const auto &v) { _batch_size = v; }), "Batch size (default: 32)")
     ("learning_rate", po::value<float>()->default_value(1.0e-4)->notifier([&](const auto &v) { _learning_rate = v; }), "Learning rate (default: 1.0e-4)")
--- a/tests/tools/onert_train/src/args.h
+++ b/tests/tools/onert_train/src/args.h
@@ -49,6 +49,9 @@ public:
   const std::string &getPackageFilename(void) const { return _package_filename; }
   const std::string &getModelFilename(void) const { return _model_filename; }
   const bool useSingleModel(void) const { return _use_single_model; }
+  const int getDataLength(void) const { return _data_length; }
+  const std::string &getLoadRawInputFilename(void) const { return _load_raw_input_filename; }
+  const std::string &getLoadRawExpectedFilename(void) const { return _load_raw_expected_filename; }
   const int getEpoch(void) const { return _epoch; }
   const int getBatchSize(void) const { return _batch_size; }
   const float getLearningRate(void) const { return _learning_rate; }
@@ -69,6 +72,9 @@ private:
   std::string _package_filename;
   std::string _model_filename;
   bool _use_single_model = false;
+  int _data_length;
+  std::string _load_raw_input_filename;
+  std::string _load_raw_expected_filename;
   int _epoch;
   int _batch_size;
   float _learning_rate;


tests/tools/onert_train/src/args.cc
```diff
@@ -184,6 +198,17 @@ void Args::Initialize(void)
...
+    ("input_datafile", po::value<std::string>()->notifier(process_input_datafile),
+         "NN Model Input data file\n"
+         "The datafile must have data for each input number.\n"
+         "If there are 3 inputs, the data of input0 must exist as much as data_length, "
+         "and the data for input1 and input2 must be held sequentially as data_length.\n"
+    )
+    ("expected_datafile", po::value<std::string>()->notifier(process_expected_datafile),
+         "NN Model Expected data file\n"
+         "(Same data policy with input_datafile)\n"
+    )
```
glistening, line 211: @jyoungyun I don't understand the input and expected data file format. I guess it may need more information. For your information, `onert_run` used hdf5 format for input and output file. `--dump` and `--load` are the options.

jyoungyun: `onert_run` takes both file formats hdf5 and raw format. And this code is related to raw format data files. Besides I haven't considered hdf5 datafile format yet. The `--dump` option in `onert_run` is for output data. However, this code is for loading input and expected data file, so I can't use `--dump` here.

I can change the argument name to `--load_input:raw` and `--load_expected:raw`.
Do you want to change the argument name like this?

glistening: @glistening Yes, I forgot to mention raw format. ( I wrote the support for both hdf5 and raw. ) Why don't you use --load:raw? I've found https://github.com/Samsung/ONE/pull/9016.

glistening: @jyoungyun I am not sure the meaning of `raw`. It does not alter the format, it just read and write as-it-is. But where will you get the expected format?

jyoungyun: > But where will you get the expected raw values?

I got the raw value from `onert_run` with `--dump:raw` argument and small changes to dump raw input data.
The meaning of `raw` is the same with `onert_run`.

glistening: Ah... I thought the explanation

>   "The datafile must have data for each input number.\n"
>   "If there are 3 inputs, the data of input0 must exist as much as data_length, "
>   "and the data for input1 and input2 must be held sequentially as data_length.\n"

is about file format.

But it seems about multiple input files.

@jyoungyun What does `data_length` mean? Input count?

If so, how can I pass the multiple input files using `load_input:raw`?
I don't understand how to use this parameter.

Could you please show me an example usage?

jyoungyun: > @jyoungyun What does data_length mean? Input count?

`data_length` means how many input data in `load_input:raw` file.

You can use this argument like below:

```
$ onert_train --modelfile a.circle --data_length 100 --load_input:raw a.in.bin --load_expected:raw a.ex.bin
```

jyoungyun: > @jyoungyun In above case, the input data file names should be:
> a.in.bin1, a.in.bin2, ..., a.in.bin100 ?
> 
> Am I right?

No, in the above case, `a.in.bin` and `a.ex.bin` contain 100 counts data. So we need `data_length`.
The one datafile contains both data as much as data_length and input counts.

glistening: Ah, is it for batch_size? Then, how multiple inputs are handled?

jyoungyun: It is not a batch_size. The data_length will be greater than batch_size.
And I described how to handle multiple inputs in this PR.

```
         "The datafile must have data for each input number.\n"
         "If there are 3 inputs, the data of input0 must exist as much as data_length, "
         "and the data for input1 and input2 must be held sequentially as data_length.\n"
```

If you have 3 inputs in a model and the data size is 100, your data file will be configured like this:

```
in0_0 in0_1 in0_2 ... in0_(data_length-1) in1_0 in1_1 in1_2 ... in1_(data_length-1) in2_0 in2_1 in2_2 ... in2_(data_length-1)
```

glistening: I could not understand how to use from your PR's argument description. Now, with more explanation especially with the example above, I understood. 

1. But argument description lacks of the information of `_#` for each input.

2. Do we need to specify in0_0 in0_1 ... in0_{data_length_1) ?
We may choose to try in0_0 to in0_{data_length_1} automatically if `data_length` is passed by user.

I would like to get other's opinion.

glistening: > No, in the above case, `a.in.bin` and `a.ex.bin` contain 100 counts data. So we need `data_length`. The one datafile contains both data as much as data_length and input counts.

Then, as I understand, `data_length` is about dataset. `data_length` means the count of training examples.

jyoungyun: > you mean that the user needs to prepare those files, not to type those in command line.

Yes, the data file contains the data with the above format.

> Then, as I understand, data_length is about dataset. data_length means the count of training examples.

Yes, it is what I want to say. :)

jyoungyun: @glistening I made a draft PR for generating data file. This readme file will help you understand the format of datafile.
https://github.com/Samsung/ONE/pull/10901/files#diff-bbfbb7c2736d349f694e20894520282f5478b13cdf499e02f20b2e890e6d37b8

glistening: Thank you for your draft. It is the same to what I've understood before your draft PR.

At first, I was confused it is about file format. But, as I understand, it does not introduce new format, it uses the existing `raw` format. It defines input/expected file name convention for dataset.

tests/tools/onert_train/src/args.cc
```diff
@@ -184,6 +198,17 @@ void Args::Initialize(void)
     ("nnpackage", po::value<std::string>()->notifier(process_nnpackage), "NN Package file(directory) name")
     ("modelfile", po::value<std::string>()->notifier(process_modelfile), "NN Model filename")
     ("path", po::value<std::string>()->notifier(process_path), "NN Package or NN Modelfile path")
+    ("data_length", po::value<int>()->default_value(-1)->notifier([&](const auto &v) { _data_length = v; }), "Data length number")
```
glistening, line 201: (optional) I prefer `dataset_len`.

tests/tools/onert_train/src/args.cc
```diff
@@ -184,6 +198,17 @@ void Args::Initialize(void)
...
+    ("load_input:raw", po::value<std::string>()->notifier(process_load_raw_inputfile),
+         "NN Model Raw Input data file\n"
+         "The datafile must have data for each input number.\n"
+         "If there are 3 inputs, the data of input0 must exist as much as data_length, "
+         "and the data for input1 and input2 must be held sequentially as data_length.\n"
```
glistening, line 206: From this description, I would prepare input0, input1, ... input{data_length-1}.
But your comment seems to require input_0, input_1, ... input_{data_length-1}.
Also, if the model gets multiple inputs input0_0, input0_1, ...input_{data_length-1}, ... That is, the model input index comes first, and training example index follows with `_`. Let's update the explanation if necessary in next PR.

glistening (approved):
LGTM

jyoungyun:
PTAL @Samsung/one_onert

ragmani (approved):

glistening:
From https://github.com/Samsung/ONE/pull/11029, I've realized I misunderstood. @jyoungyun suggests to use one file for dataset. Then, I think we need to review this approach again.

glistening:
I think the explanation for command line argument needs to be updated.

>  "The datafile must have data for each input number.\n"

What do you mean for each input number?
- model input count?
- training input example count?

>  "If there are 3 inputs, the data of input0 must exist as much as data_length, "

What do you mean by `input0` here?
- 0th input of model?
- 1st training example's input?

>  "and the data for input1 and input2 must be held sequentially as data_length.\n"

What do you mean by `input` and `input2` here?

jyoungyun:
> >  "The datafile must have data for each input number.\n"
> 
> What do you mean for each input number?
> - model input count?
> - training input example count?

Model input count.

> training input example count is equal with `data_length`.
> 
> >  "If there are 3 inputs, the data of input0 must exist as much as data_length, "
> 
> What do you mean by `input0` here?
> - 0th input of model?
> - 1st training example's input?

0th input of model.

PR #10991, opened 2023-07-04
[onert] Convert Permute op to trainable operation

ragmani:
This commit applies converting Permute operation to trainable Permute operation.

### DIFF
--- a/runtime/onert/core/src/compiler/train/LoweredTrainableGraph.cc
+++ b/runtime/onert/core/src/compiler/train/LoweredTrainableGraph.cc
@@ -26,8 +26,7 @@
 #include "../../backend/builtin/Config.h"
 #include "../../dumper/text/GraphDumper.h"
 #include "../../ir/verifier/Verifier.h"
-// TODO Enable after merging TrainableOperationConverter
-// #include "TrainableOperationConverter.h"
+#include "TrainableOperationConverter.h"
 
 #include "backend/Backend.h"
 #include "backend/train/ITrainableBackend.h"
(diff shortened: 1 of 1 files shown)


jyoungyun (approved):
LGTM

(out of curiosity)
Why does Permute Op change to a trainable operation?
Should all operators be changed to trainable operator?

ragmani:
@jyoungyun 
> Why does Permute Op change to a trainable operation?

Kernel generators for training generate trainable operations to `TrainableFnSequence`. The Permute op is no exception either.

> Should all operators be changed to trainable operator?

Yes, unless we change kernel generators.

hseok-oh (approved):
LGTM

runtime/onert/core/src/compiler/train/LoweredTrainableGraph.cc
```diff
@@ -105,19 +104,18 @@ void LoweredTrainableGraph::lowerGraph(const CompilerOptions &options)
...
-  //       assert(gen_index == index);
-  //     }
-  //   });
+  // TODO Move converting Permute op into PermutationInsertionPass
+  auto op_converter = TrainableOperationConverter{_trainable_graph, nullptr};
+  _trainable_graph.operations().iterate(
+    [&](const onert::ir::OperationIndex &index, const onert::ir::IOperation &op) {
+      if (op.opcode() == ir::OpCode::Permute)
+      {
+        auto top = op_converter(index);
```
glistening, line 113: I thought `top` as it is, the opposite of bottom. : )

op_converter` coverts op to trainable operation, and push it in a stack, and returns the top index. But I realized that `top` means `trainable operation`.

glistening (approved):
LGTM

runtime/onert/core/src/compiler/train/LoweredTrainableGraph.cc
ragmani: How about changing to convert operation instead of index? changing the name to `trainable_op`?

glistening: I hope I am the only one who was confused by `top`. To make sure, `trainable_op` looks good. `t_op`, `tr_op`, ... are all good.

ragmani: OK, I will change that. What is your opinion on changing to convert operation instread of index?

ragmani: I create a new RR, #10996 for that.

PR #10993, opened 2023-07-05
[onert] Remove _has_info variable from TrainingInfo

jyoungyun:
This commit removes the has_info variable from TrainingInfo.
The default value of trainnig_info in CompilerFactory.create() is nullptr.
Therefore, if create() is not called with TrainingInfo, it is the same with shouldTrain() == false.
Conversely, if create() is called with TrainingInfo, it is the same with shouldTrain() == true.


glistening (approved):
LGTM

ragmani (approved):
LGTM

PR #11007, opened 2023-07-06
[onert] Introduce TrainableTensor

ragmani:
This commit introduce TrainableTensor into train backend.

ONE-DCO-1.0-Signed-off-by: ragmani <ragmani0216@gmail.com>

PR #11013, opened 2023-07-06
[onert] Separate NativeTensor of TensorRegistry in train backend

ragmani:
This commit separates NativeTensor of TensorRegistry into NonConstTensor and TrainableTensor in train backend.

### DIFF
--- a/runtime/onert/backend/train/TensorRegistry.h
+++ b/runtime/onert/backend/train/TensorRegistry.h
@@ -40,7 +40,13 @@ public:
     return PortableTensorRegistryTemplate<Tensor>::getITensor(ind);
   }
 
-  ITensor *getNativeITensor(const ir::OperandIndex &ind) override { return getNativeTensor(ind); }
+  ITensor *getNativeITensor(const ir::OperandIndex &ind) override
+  {
+    ITensor *tensor = getTrainableTensor(ind);
+    if (tensor == nullptr)
+      tensor = getNonConstTensor(ind);
+    return tensor;
+  }
 
   IPortableTensor *getPortableTensor(const ir::OperandIndex &ind)
   {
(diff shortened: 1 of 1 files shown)


glistening:
@ragmani As I understand, native tensor is about whether the tensor is created from its backend or other backend. I don't have idea how and why this concept is related to trainable tensor. Could you please explain the background or design behind this PR?

ragmani:
@glistening 
`TrainableTensor` will be introduced at #11007. And within train backend, we can handle `Tensor` for non constants and `TrainableTensor` for constants(like weights) separately.
Also, outside of train backend we can get native tensors by using `getNativeITensor()`.

runtime/onert/backend/train/TensorRegistry.h
```diff
@@ -82,7 +85,7 @@ class TensorRegistry : public PortableTensorRegistryTemplate<Tensor>
     return PortableTensorRegistryTemplate<Tensor>::setMigrantTensor(ind, tensor);
   }
 
-  void setNativeTensor(const ir::OperandIndex &ind, std::unique_ptr<Tensor> tensor)
+  void setNonConstTensor(const ir::OperandIndex &ind, std::unique_ptr<Tensor> tensor)
```
glistening, line 88: Please update exception message at line 94.

Also, before setting non-const tensor, it checks `_trainable`.
Does every non-const tensor is in _trainable?
I guess there has been non-const tensors before trainable tensor is introduced.

ragmani: > Please update exception message at line 94.

OK, I will change "native" to "non const"

> Also, before setting non-const tensor, it checks _trainable.
> Does every non-const tensor is in _trainable?

No, It checks whether the index is valid. Every const tensor is in `_trainable`.

glistening (approved):

PR #11011, opened 2023-07-06
[onert] Remove copy from auto keyword

hseok-oh:
This commit removes compiler and analysis warning from copying on auto keyword by adding &.

ONE-DCO-1.0-Signed-off-by: Hyeongseok Oh <hseok82.oh@samsung.com>

PR #11017, opened 2023-07-07
[onert] Use forwarding on range based loop

hseok-oh:
This commit update range based loop to use forwarding.

---

Related issue: https://github.com/Samsung/ONE/pull/11011#discussion_r1254201106


ragmani (approved):
LGTM

jyoungyun (approved):
LGTM

glistening:
After searching, I've found the notation `auto&&` is necessary when the element is going to be modified, and do no harm for reading cases.

Personally I don't think it is necessary change. In most cases, we just read the elements and the sequence element seems index.

I think it is a taste of preference. Please let me know if I am missing something.

Anyway, I will merge this.

ragmani:
If we just want to read elements in the range, I recommend using `const auto &`. 
If we want to modify elements in the range, I recommend using `auto &&`. It's a forwarding reference and safe even if the code is moved in generic code.

PR #11016, opened 2023-07-06
[DRAFT][luci/pass] Support a new pattern in FuseGeluPass

BalyshevArtem:
This PR supports a new pattern in FuseGeluPass.

for issue:  #11002

ONE-DCO-1.0-Signed-off-by: Artem Balyshev <a.balyshev@samsung.com>

Issue #11002, opened 2023-07-06
[one-optimize] Support a new pattern in FuseGeluPass


jinevening:
## What

Let's add a new pattern in FuseGeluPass.

tflite file: [GELU.zip](https://github.com/Samsung/ONE/files/11962886/GELU.zip)

## Why

Currently FuseGeluPass can handle a pattern generated when Gelu is imported from onnx model.

Old TF generates a different pattern to perform Gelu. Let's support it.

seanshpark:
![image](https://github.com/Samsung/ONE/assets/4616940/232b6f7e-50b0-44f9-9340-374735c2f73a)

Ops
```
[0] MUL (code: 18, dep_code: 18, version: 1)
[1] CUSTOM(Erf) (code: 32, dep_code: 32, version: 1)
[2] ADD (code: 0, dep_code: 0, version: 1)
```

seanshpark:
If anyone interested doing this, please take it by self assign :)
If not taken in a time, I'll do it :)

seanshpark:
From the class `GeluPattern` ... what I can think of 
1/ Add another class `GeluPattern2` to match pattern of this issue
2/ Add `GeluPattern::matched2()` to match pattern of this issue
3/ Inside `GeluPattern::matched()`, call internal `match_type1()`, `match_type2()` ...
4/ ...

seanshpark:
@BalyshevArtem , thanks for taking this issue!

@jinevening , I've found that we don't have luci-pass-value test for fuse GELU. was there any issue when you added the pass ?

jinevening:
There is an issue. Erf Op (included in Gelu pattern) cannot be executed on the current venv (w/ TF 2.8.0).

seanshpark:
As we cannot run value test, let's check if we can do some `circle2circle-dredd-recipe-test`

seanshpark:
done

PR #11020, opened 2023-07-07
[DRAFT][luci/pass] Refactor FuseGeluPass

BalyshevArtem:
This PR refactors FuseGeluPass to add more patterns in future.

for issue:  #11002
from draft: #11016

ONE-DCO-1.0-Signed-off-by: Artem Balyshev <a.balyshev@samsung.com>

PR #11024, opened 2023-07-10
[luci/pass] Introduce GeluPatternBase

BalyshevArtem:
This PR introduces GeluPatternBase class.

for issue: https://github.com/Samsung/ONE/issues/11002
from draft: https://github.com/Samsung/ONE/pull/11016

### DIFF
--- a/compiler/luci/pass/src/FuseGeluPass.cpp
+++ b/compiler/luci/pass/src/FuseGeluPass.cpp
@@ -43,9 +43,6 @@ public:
 public:
   virtual bool matched() = 0;
 
-protected:
-  bool matched_base();
-
 public:
   luci::CircleNode *_ifm = nullptr;
   luci::CircleMul *_mul_sqrt = nullptr;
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/FuseGeluPass.cpp
```diff
@@ -62,10 +89,10 @@ bool same(float a, float b) { return fabs(a - b) < 1e-5; }
  *          [Out]
  *
  */
-class GeluPattern final
+class GeluPattern final : public GeluPatternBase
 {
 public:
-  GeluPattern(luci::CircleMul *candidate)
+  GeluPattern(luci::CircleMul *candidate) : GeluPatternBase(candidate)
   {
```
BalyshevArtem, line 96: just to avoid `define but not used` error

seanshpark:
it would be better to update #11016 and use this as draft and drop #11020

compiler/luci/pass/src/FuseGeluPass.cpp
```diff
@@ -91,6 +118,51 @@ class GeluPattern final
   if (not(condition))             \
     return false;
 
+bool GeluPatternBase::matched_base()
```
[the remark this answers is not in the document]
BalyshevArtem: okay, I will remove this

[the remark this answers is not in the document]
BalyshevArtem: Done

BalyshevArtem:
> it would be better to update #11016 and use this as draft and drop #11020

I'm not sure I understood you correctly. If I understand correctly, do you want me to update one of these drafts with the content from this PR? I would like the overall picture to remain visible, from where the current PR stands out, and therefore leave them, but if it will bettter to remove them, then I will.

seanshpark:
> do you want me to update one of these drafts

I mean, would be better to use (link in the message) first draft not #11020

compiler/luci/pass/src/FuseGeluPass.cpp
seanshpark: can you check this as the method is now removed?

BalyshevArtem: I checked, without this CI crashes with an error `define but not used`

seanshpark (approved):
LGTM thank you!

PR #11030, opened 2023-07-10
[onert] Introduce train function in Execution

jyoungyun:
This commit introduces train function in Execution. This function will be called from nnfw_api_internal apis.

### DIFF
--- a/runtime/onert/core/src/exec/Execution.cc
+++ b/runtime/onert/core/src/exec/Execution.cc
@@ -16,6 +16,8 @@
 
 #include "exec/Execution.h"
 
+#include "train/TrainableExecutors.h"
+
 #include "util/logging.h"
 
 namespace onert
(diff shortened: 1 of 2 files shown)


runtime/onert/core/src/exec/Execution.cc
```diff
@@ -151,6 +153,24 @@ void Execution::waitFinish()
 
 bool Execution::isFinished(void) const { return finished; }
 
+#ifdef ONERT_TRAIN
+void Execution::train()
+{
+  const auto execs = dynamic_cast<exec::train::TrainableExecutors *>(_executors.get());
```
ragmani, line 159:
```suggestion
  auto execs = dynamic_cast<exec::train::TrainableExecutors *>(_executors.get());
```

jyoungyun: (Out of curiosity)
Why should `const` keyword to be removed?

ragmani: Can't `execs` be updated inside `train()`? 
I'm not sure why there is no compilation error. If you changes `const auto` to `const TrainableExecutors *`, you will get a compilation error.

ragmani (approved):
LGTM

hseok-oh (approved):
LGTM

PR #11044, opened 2023-07-11
[onert] Refine TrainableTensor

ragmani:
This commit refines TrainableTensor.
  - Introduce derivative tensor into TrainableTensor
  - Remove applyGradient

### DIFF
--- a/runtime/onert/backend/train/Tensor.h
+++ b/runtime/onert/backend/train/Tensor.h
@@ -50,16 +50,21 @@ public:
   TrainableTensor() = delete;
 
 public:
-  TrainableTensor(const ir::OperandInfo &info, const ir::Layout layout) : Tensor{info, layout}
+  TrainableTensor(const ir::OperandInfo &info, const ir::Layout layout)
+    : Tensor{info, layout}, _gradient{info, layout}
   {
     // DO NOTHING
   }
 
 public:
-  void applyGradient(const IPortableTensor &grad_tensor, double lr) override;
+  IPortableTensor &weightTensor() override { return *this; }
+  IPortableTensor &gradTensor() override { return _gradient; }
 
 public:
   void fillBuffer(const std::shared_ptr<ir::Data> &data);
+
+private:
+  Tensor _gradient;
 };
 
 } // namespace train
--- a/runtime/onert/core/include/backend/train/ITrainableTensor.h
+++ b/runtime/onert/core/include/backend/train/ITrainableTensor.h
@@ -35,13 +35,9 @@ class ITrainableTensor
 public:
   virtual ~ITrainableTensor() = default;
 
-  /**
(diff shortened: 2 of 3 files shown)


runtime/onert/core/include/backend/train/ITrainableTensor.h
```diff
@@ -35,13 +35,9 @@ class ITrainableTensor
...
-  /**
-   * @brief Apply gradient to this trainable tensor
-   *
-   * @param grad_tensor A gradient tensor to be used for updating this trainable tensor
-   * @param lr          Learning rate
-   */
-  virtual void applyGradient(const IPortableTensor &grad_tensor, double lr) = 0;
+  // TODO Encapsulation
+  virtual IPortableTensor &weightTensor() = 0;
+  virtual IPortableTensor &derivTensor() = 0;
```
ragmani, line 40: I did this because I still don't fully understand. I'll refactor it later.

hseok-oh (approved):

runtime/onert/backend/train/Tensor.h
```diff
@@ -50,16 +50,21 @@ class TrainableTensor : public Tensor, public train::ITrainableTensor
...
-  TrainableTensor(const ir::OperandInfo &info, const ir::Layout layout) : Tensor{info, layout}
+  TrainableTensor(const ir::OperandInfo &info, const ir::Layout layout)
+    : Tensor{info, layout}, _gradient{info, layout}
   {
     // DO NOTHING
   }
 
 public:
-  void applyGradient(const IPortableTensor &grad_tensor, double lr) override;
+  IPortableTensor &weightTensor() override { return *this; }
```
glistening, line 60: (optional) Not every trainable tensor is weight tensor. Some may be `bias` tensor. Also, do we need this method? TrainableTensor itself is `Tensor`. `grad()` or `gradTensor()` is enough.

glistening (approved):

runtime/onert/backend/train/Tensor.h
ragmani: > Not every trainable tensor is weight tensor. Some may be bias tensor.

Right. It may not be a weight, but may be a bias. I couldn't find a suitable name. Could you recommend?

> Also, do we need this method? TrainableTensor itself is Tensor. grad() or gradTensor() is enough.

Tensors that can be trained need tensors for gradients to be updated. So I put a weight tensor and a gradient tensor together in `TrainableTensor` to make it easier to access them without using index in kernels and optimizers.

glistening: I don't suggest to use another name. Rather, I suggested not to introduce new method since `TrainableTensor` itself is a `Tensor`, which inherits `Tensor`. Anyway, I don't oppose this PR.

PR #11053, opened 2023-07-12
[oops] Add requires.cmake

hseok-oh:
It introduces requires.cmake because "oops" links "pepper-str".


seanshpark (approved):
LGTM thank you!

PR #11060, opened 2023-07-13
[luci/pass] Rename GeluPattern to GeluPattern1

BalyshevArtem:
This PR renames GeluPattern to GeluPattern1.

for issue: https://github.com/Samsung/ONE/issues/11002
from draft: https://github.com/Samsung/ONE/pull/11016

### DIFF
--- a/compiler/luci/pass/src/FuseGeluPass.cpp
+++ b/compiler/luci/pass/src/FuseGeluPass.cpp
@@ -86,10 +86,10 @@ public:
  *          [Out]
  *
  */
-class GeluPattern final : public GeluPatternBase
+class GeluPattern1 final : public GeluPatternBase
 {
 public:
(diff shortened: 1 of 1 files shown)


BalyshevArtem:
@nnfw-bot test nncc-release

BalyshevArtem:
> what's the relation with #11062 ?

there is only renaming. After renaming will be merged then we can add inherit  details from https://github.com/Samsung/ONE/pull/11062 (mark matched as override and add override destructo).
Of course we can just merge #11062. I am just adhere to  plan  from https://github.com/Samsung/ONE/pull/11020#issuecomment-1627830310 comment :)

seanshpark:
> there is only renaming. After renaming ...

Sorry to make some direction without full understanding of your changes.
11062 has similar changes so this makes code reading confusing.
It would be better not to do like this way.

PR #11061, opened 2023-07-13
[onert] Use const & for saving training info in TrainingCompiler

jyoungyun:
This commit uses const & keyword for saving training info in TrainingCompiler.


runtime/onert/core/src/compiler/train/TrainingCompiler.h
```diff
@@ -73,7 +73,7 @@ class TrainingCompiler : public ICompiler
 private:
   std::shared_ptr<ir::Model> _model;
   CompilerOptions *_options;
-  const TrainingInfo *_training_info;
+  const TrainingInfo &_training_info;
```
ragmani, line 76: If you want to copy `TrainableInfo`
```suggestion
  const TrainingInfo _training_info;
```

ragmani (approved):
LGTM

hseok-oh (approved):
LGTM

PR #11063, opened 2023-07-13
[tools/onert_train] Add loadata_2 test case

jyoungyun:
This commit adds loaddata_2 test case.
This test case is test for verifying Float32 input and expected data.


jyoungyun:
The tensorinfo concept has changed. So I need to fix codes.

jyoungyun:
I will do this later.

PR #11075, opened 2023-07-14
[onert] Introduce SGD optimizer

ragmani:
This commit introduces SGD optimizer.

### DIFF
--- a/runtime/onert/core/src/exec/train/optimizer/SGD.cc
+++ b/runtime/onert/core/src/exec/train/optimizer/SGD.cc
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


jyoungyun:

runtime/onert/core/src/exec/train/optimizer/SGD.cc
```diff
@@ -0,0 +1,66 @@
...
+  {
+    throw std::runtime_error("SGD: Invalid gradient tensor");
+  }
+
+  switch (grad_tensor.data_type())
+  {
+    case ir::DataType::FLOAT32:
+      elementwise<float>(shape, grad_tensor, trainable_tensor,
+                         [&](float dst, float src) -> float { return dst + src * lr; });
+      break;
```
jyoungyun, line 57: (out of curiosity)
Can this lambda function be changed according to data type?

jyoungyun (approved):
LGTM

ragmani:

runtime/onert/core/src/exec/train/optimizer/SGD.cc
ragmani: Yes, unless the data type is a quantization type.

glistening:
I expected `SGD` will be the simplest optimizer.
But this codes has more like `momentum`, `nesterov`.

After searching more, I realized that it is about `SGD` with momentum (more general version). `SGD` may be set by using your default value:

```cc
double momentum{0.0};
bool nesterov{false};
```

glistening (approved):
LGTM

PR #11035, opened 2023-07-11
[Draft] Implement cker train

jyoungyun:

PR #11077, opened 2023-07-17
[onert] Add getInputTotalSize in execution

jyoungyun:
This commit adds getInputTotalSize in execution.
This function will be used in train_set_input function of nnfw internal api.

Draft: #11035


glistening (approved):
LGTM

(optional) `getInputTotalSize()` may be useful for inference also, and I would like to reduce #ifdef ONERT_TRAIN... . Thus, it may be okay to remove #ifdef ONERT_TRAIN...

ragmani (approved):
LGTM

PR #11083, opened 2023-07-17
[onert] Introduce unified TensorBuilder into train backend

ragmani:
This commit introduces unified TensorBuilder into train backend.

### DIFF
--- a/runtime/onert/backend/train/TensorBuilder.cc
+++ b/runtime/onert/backend/train/TensorBuilder.cc
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "TensorBuilder.h"
+
+#include "Tensor.h"
+
+namespace onert
+{
+namespace backend
+{
+namespace train
+{
+
+TensorBuilder::TensorBuilder(const std::shared_ptr<TensorRegistry> &tensor_reg,
+                             const std::string planner_id)
+  : _tensor_reg{tensor_reg}, _tensor_mgr{new TensorManager(tensor_reg, planner_id)}
+{
+  /* empty */
+}
+
+void TensorBuilder::registerTensorInfo(const ir::OperandIndex &index, const ir::OperandInfo &info,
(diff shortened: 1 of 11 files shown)


runtime/onert/backend/train/TensorBuilder.cc
```diff
@@ -0,0 +1,134 @@
...
+  //      the corresponding operand is constant.
+  if (_as_constants[index])
+  {
+    auto tensor = std::make_unique<TrainableTensor>(info, layout);
+    _tensor_reg->setTrainableTensor(index, std::move(tensor));
+  }
+  else
+  {
+    auto tensor = std::make_unique<Tensor>(info, layout);
+    _tensor_reg->setNonConstTensor(index, std::move(tensor));
```
jyoungyun, line 55: Could you explain about `setTrainableTensor` and `setNonConstTensor`?

runtime/onert/backend/train/TensorBuilder.cc
```diff
@@ -0,0 +1,134 @@
...
+{
+  // TODO Support momory plan
+  if (_as_constants[index])
+  {
+    _tensor_mgr->claimGradientPlan(index);
+  }
+  else
+  {
+    _tensor_mgr->claimDerivativePlan(index);
+  }
```
jyoungyun, line 111: What is the meaning of `_as_constants`?
Is it related to differentiability?

runtime/onert/backend/train/TensorBuilder.cc
ragmani: They are just setters. They append tensors to TensorRegistry for each tensor type.

runtime/onert/backend/train/TensorBuilder.cc
ragmani: > What is the meaning of _as_constants?

`_as_constants` contains whether each operand is a constant. Since there is no way to distinguish trainable operands(like weights) yet, `_as_constants` is used to distinguish trainable operands.

> Is it related to differentiability?

No, rather than that,  `_as_constants` was used to distinguish between derivative tensors that will be backpropagated and gradient tensors that will be used to update trainable tensors here.
Also we may need derivative tensors of the corresponding trainable tensors. If so, we have to modify this change like this:
```suggestion
  if (_as_constants[index])
  {
    _tensor_mgr->claimGradientPlan(index);
  }
  _tensor_mgr->claimDerivativePlan(index);
```
But I'm not sure if those derivative tensors is necessary.

runtime/onert/backend/train/TensorBuilder.cc
jyoungyun: Ah, I want to know what's difference between TrainableTensor and NonConstTensor. :)

jyoungyun: As I understood, TrainableTensor has weight or bias to be trained. NonConstTensor will be used for forward and backward function and has a differential value during backwarding.

jyoungyun (approved):
LGTM

runtime/onert/backend/train/TensorBuilder.cc
ragmani: Mostly right, but `NonConstTensor` does not have a differential value. It has only a result value of inference and can be used to calculate differential values.

PR #11087, opened 2023-07-17
[onert] Implement train Permute kernel

ragmani:
This commit implements train Permute kernel.

### DIFF
--- a/runtime/onert/core/src/backend/builtin/train/kernel/PermuteLayer.cc
+++ b/runtime/onert/core/src/backend/builtin/train/kernel/PermuteLayer.cc
@@ -31,16 +31,37 @@ namespace kernel
 
 PermuteLayer::PermuteLayer(const std::vector<ITensor *> &src_tensors,
                            const std::vector<ITensor *> &dst_tensors,
+                           const std::vector<ITensor *> &input_deriv_tensors,
+                           const std::vector<ITensor *> &output_deriv_tensors,
                            const std::shared_ptr<ExternalContext> &external_context)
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/backend/builtin/train/kernel/PermuteLayer.cc
```diff
@@ -31,16 +31,37 @@ namespace kernel
...
                            const std::shared_ptr<ExternalContext> &external_context)
-  : builtin::kernel::PermuteLayer{src_tensors, dst_tensors, external_context}
+  : builtin::kernel::PermuteLayer{src_tensors, dst_tensors, external_context},
+    _input_deriv_tensors{input_deriv_tensors}, _output_deriv_tensors{output_deriv_tensors}
 {
+  assert(input_deriv_tensors.size() == output_deriv_tensors.size());
+  assert(input_deriv_tensors.size() == src_tensors.size());
+}
+
+void PermuteLayer::optimize()
```
jyoungyun, line 44: Is this optimize() function called from train optimizer?

ragmani: No, this is for `PermuteLayer` only.

jyoungyun (approved):
LGTM

glistening:
Here is my understanding.

In inference, `Permute` layer is inserted for reconciling different backends and layouts.
( I don't know why `Permute` layer is not eliminated because `traininig` backend uses only cpu and there is only one layout. )

Anyway, in training, especially in this PR, @ragmani told me that `Permute` layer just forwards the derivatives to the next layer.

PR #11089, opened 2023-07-17
[circle-opselector] No string copy from auto type deduction

mhs4670go:
This commit uses a reference not to copy string.

This makes a static analysis tool happy.


seanshpark:
@nnfw-bot test nncc-debug

seanshpark (approved):
LGTM

PR #11105, opened 2023-07-18
[onert] Introduce MEAN_SQUARED_ERROR to Loss operation

jyoungyun:
This commit introduces MEAN_SQUARED_ERROR to Loss operation. Since CATEGORICAL_CROSSENTROPY may not operate according to the model structure, the default type of loss op has also been changed to MEAN_SQUARED_ERROR.

Draft : #11035

### DIFF
--- a/runtime/onert/core/include/ir/operation/Loss.h
+++ b/runtime/onert/core/include/ir/operation/Loss.h
@@ -41,6 +41,7 @@ public:
   //      from the object of this class, we have to consider whether to change this enum class.
   enum class Type
   {
+    MEAN_SQUARED_ERROR,
     CATEGORICAL_CROSSENTROPY
   };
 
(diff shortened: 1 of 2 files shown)


ragmani (approved):
LGTM

runtime/onert/core/include/ir/operation/Loss.h
```diff
@@ -41,14 +41,15 @@ class Loss : public Operation
   //      from the object of this class, we have to consider whether to change this enum class.
   enum class Type
   {
+    MEAN_SQUARED_ERROR,
     CATEGORICAL_CROSSENTROPY
```
glistening, line 45: @jyoungyun (optional) I would like to add new enumerator at the end. Also, it would be better to align to circle+ schema (draft).
```suggestion
    CATEGORICAL_CROSSENTROPY = 1,
    MEAN_SQUARED_ERROR = 2,
```
But we may change it after implementation is almost done. It is under implementation, no one is likely to use it since it is not working yet. : )

glistening (approved):

PR #11106, opened 2023-07-18
[onert] Implement LossInsertionPass

jyoungyun:
This commit implements LossInsertionPass.
This function runs in the middle of constructing the trainable graph. Loss uses the original graph output as y_pred input. And it creates y_true input in the same shape and type as y_pred input. The output is Scalar tensor by default.

Draft: #11035


ragmani (approved):
LGTM

glistening (approved):

PR #11108, opened 2023-07-18
[onert] Implement Loss visitor function in StaticShapeInferer

jyoungyun:
This commit implements Loss visitor function in StaticShapeInferer.
The Y_TRUE input should be the same shape as Y_PRED.

Draft: #11035

### DIFF
--- a/runtime/onert/core/src/compiler/StaticShapeInferer.cc
+++ b/runtime/onert/core/src/compiler/StaticShapeInferer.cc
@@ -605,6 +605,23 @@ void StaticShapeInferer::visit(const ir::operation::L2Normalization &op)
   handleSimpleUnaryOp(op, op.getInputs().at(ir::operation::L2Normalization::Input::INPUT));
 }
 
+void StaticShapeInferer::visit(const ir::operation::Loss &op)
+{
+  auto &operands = _lowered_subg->graph().operands();
+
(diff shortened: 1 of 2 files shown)


glistening (approved):

runtime/onert/core/src/compiler/StaticShapeInferer.cc
```diff
@@ -605,6 +605,23 @@ void StaticShapeInferer::visit(const ir::operation::L2Normalization &op)
...
+  const auto &y_pred = operands.at(y_pred_idx);
+
+  const auto y_true_idx{op.getInputs().at(ir::operation::Loss::Input::Y_TRUE)};
+  auto &y_true = operands.at(y_true_idx);
+
+  // TODO Consider SparseCategoricalCrossentropy case
+
+  y_true.info().shape(y_pred.info().shape());
+
+  // TODO Consider output shape in case of reduction option
```
ragmani, line 622: Why don't you set the output shape in case it's not a reduction option?

jyoungyun: We don't need to change the loss output even if the batch size changes.
The output of loss is always a scalar tensor, regardless of batch_size.
However, some types of reduction option use different shape output size.
I will consider it later.

ragmani (approved):
LGTM

PR #11112, opened 2023-07-19
[tools/onert_train] Do not change input and expected tensorinfo

jyoungyun:
This commit prevents input and expected tensorinfo from changing. If the last parameter of set_input and set_expected function is not nullptr, the runtime changes the tensorinfo. This tool uses nullptr because it does not change tensorinfo.

Draft : #11035

Related code: https://github.com/Samsung/ONE/blob/276bd85d67bd274b55c767612669a24eb1ca43da/runtime/onert/api/src/nnfw_api_internal.cc#L1246-L1251

### DIFF
--- a/tests/tools/onert_train/src/onert_train.cc
+++ b/tests/tools/onert_train/src/onert_train.cc
@@ -193,15 +193,13 @@ int main(const int argc, char **argv)
         // prepare input
         for (uint32_t i = 0; i < num_inputs; ++i)
         {
-          NNPR_ENSURE_STATUS(
-            nnfw_train_set_input(session, i, input_data[i].data(), &input_infos[i]));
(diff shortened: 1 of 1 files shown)


jyoungyun:

ragmani:
> This commit prevents input and expected tensorinfo from changing. If the last parameter of set_input and set_expected function is not nullptr, the runtime changes the tensorinfo. This tool uses nullptr because it does not change tensorinfo.

How about adding throwing an exception in `nnfw_session::train_set_input()` in case sizes of `tensorinfo` and input of graph are different?

jyoungyun:
> How about adding throwing an exception in nnfw_session::train_set_input() in case sizes of tensorinfo and input of graph are different?

I will apply your suggestion and close this issue. Thank you for your suggestion. :)

PR #11114, opened 2023-07-19
[tools/onert_train] Do not multiply batch_size to tensor info

jyoungyun:
The tensor info of trainable graph changes the tensor info according to the batch_size during `train_prepare` function.
The onert_train tool does not need to consider the batch_size for tensorinfo. This commit therefore does not multiply batch_size to the tensor info.

~~waiting : #11112~~
Draft : #11035
Related code:
https://github.com/Samsung/ONE/blob/276bd85d67bd274b55c767612669a24eb1ca43da/runtime/onert/api/src/nnfw_api_internal.cc#L1165-L1174


hseok-oh (approved):

ragmani (approved):
LGTM

PR #11124, opened 2023-07-19
[onert] Exclude copies of derivative inputs of model

ragmani:
This commit make PermuteLayer exclude permutation of derivative inputs of model.

### DIFF
--- a/runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
+++ b/runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
@@ -59,8 +59,16 @@ void KernelGenerator::visit(const ir::train::operation::Permute &node)
   std::vector<ITensor *> output_tensors{getTensor(output_index)};
   std::vector<ITensor *> input_tensors{getTensor(input_index)};
 
-  std::vector<ITensor *> output_deriv_tensors{getDerivativeTensor(output_index)};
-  std::vector<ITensor *> input_deriv_tensors{getDerivativeTensor(input_index)};
+  std::vector<ITensor *> output_deriv_tensors;
+  std::vector<ITensor *> input_deriv_tensors;
+  // NOTE The derivative tensors corresponding to inputs of model are nullptr
+  if (auto input_deriv_tensor = getDerivativeTensor(input_index))
+  {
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/backend/builtin/train/KernelGenerator.cc
```diff
@@ -59,8 +59,16 @@ void KernelGenerator::visit(const ir::train::operation::Permute &node)
   std::vector<ITensor *> output_tensors{getTensor(output_index)};
   std::vector<ITensor *> input_tensors{getTensor(input_index)};
 
-  std::vector<ITensor *> output_deriv_tensors{getDerivativeTensor(output_index)};
-  std::vector<ITensor *> input_deriv_tensors{getDerivativeTensor(input_index)};
+  std::vector<ITensor *> output_deriv_tensors;
+  std::vector<ITensor *> input_deriv_tensors;
+  // NOTE The derivative tensors corresponding to inputs of model are nullptr
+  if (auto input_deriv_tensor = getDerivativeTensor(input_index))
```
glistening, line 65: We use c++14, and "if-statement-with-init" is introduced from c++17.

ragmani: This is not using "if-statement-with-initializer"

Please refer to https://en.cppreference.com/w/cpp/language/if
This is included in "condition" in the if syntax below.
```
attr ﻿(optional) if constexpr(optional) ( init-statement ﻿(optional) condition ) statement-true
```
Here is an explanation of what "condition" is.
"[declaration](https://en.cppreference.com/w/cpp/language/declarations) of a single non-array variable with a brace-or-equals [initializer](https://en.cppreference.com/w/cpp/language/initialization)."

Anyway, I think you don't have to worry about using c++17 syntax. If we use c++17 syntax like "if-statement-with-initializer", the compiler will generate an error like below.
```
error: init-statement in selection statements only available with ‘-std=c++17’ or ‘-std=gnu++17’ [-Werror]
```

glistening (approved):
LGTM

hseok-oh (approved):

PR #11126, opened 2023-07-19
[onert] Compare tensor size in train_set_input

jyoungyun:
This commit compares tensor size in train_set_input to verify that the tensor info has changed. If the tensor info is changed, it is not supported in this case.

draft: #11035

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -195,6 +195,29 @@ std::unique_ptr<onert::ir::Model> loadModel(const std::string filename,
   return std::unique_ptr<onert::ir::Model>(nullptr);
 }
 
+#ifdef ONERT_TRAIN
+uint64_t getBufSize(const nnfw_tensorinfo *info)
+{
+  static int elmsize[] = {
+    sizeof(float),   /* NNFW_TYPE_TENSOR_FLOAT32 = 0 */
+    sizeof(int),     /* NNFW_TYPE_TENSOR_INT32 = 1 */
+    sizeof(uint8_t), /* NNFW_TYPE_TENSOR_QUANT8_ASYMM = 2 */
+    sizeof(bool),    /* NNFW_TYPE_TENSOR_BOOL = 3 */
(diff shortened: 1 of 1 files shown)


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -195,6 +195,27 @@ std::unique_ptr<onert::ir::Model> loadModel(const std::string filename,
...
+  };
+
+  uint64_t n = 1;
+  for (int32_t i = 0; i < info->rank; ++i)
+  {
+    assert(info->dims[i] >= 0);
+    n *= info->dims[i];
+  }
+  return elmsize[info->dtype] * n;
+}
```
ragmani, line 218:
```suggestion
#ifdef ONERT_TRAIN
uint64_t getBufSize(const nnfw_tensorinfo *info)
{
  static int elmsize[] = {
    sizeof(float),   /* NNFW_TYPE_TENSOR_FLOAT32 = 0 */
    sizeof(int),     /* NNFW_TYPE_TENSOR_INT32 = 1 */
    sizeof(uint8_t), /* NNFW_TYPE_TENSOR_QUANT8_ASYMM = 2 */
    sizeof(bool),    /* NNFW_TYPE_TENSOR_BOOL = 3 */
    sizeof(uint8_t), /* NNFW_TYPE_TENSOR_UINT8 = 4 */
    sizeof(int64_t), /* NNFW_TYPE_TENSOR_INT64 = 5 */
    sizeof(int8_t),  /* NNFW_TYPE_TENSOR_QUANT8_ASYMM_SIGNED = 6 */
    sizeof(int16_t), /* NNFW_TYPE_TENSOR_QUANT16_SYMM_SIGNED = 7 */
  };

  uint64_t n = 1;
  for (int32_t i = 0; i < info->rank; ++i)
  {
    assert(info->dims[i] >= 0);
    n *= info->dims[i];
  }
  return elmsize[info->dtype] * n;
}
```
#endif // ONERT_TRAIN

or 

Move this function into `#ifdef ONERT_TRAIN` block

ragmani (approved):
LGTM

hseok-oh (approved):

PR #11131, opened 2023-07-19
DRAFT dredd test for FuseGeluPass

seanshpark:
on-going draft to do some dredd test for FuseGeluPass.

 ONE-DCO-1.0-Signed-off-by: SaeHie Park <saehie.park@gmail.com>

Issue #11130, opened 2023-07-19
[CFE] dredd test for FuseGeluPass


seanshpark:
from https://github.com/Samsung/ONE/issues/11002#issuecomment-1641790101

As we cannot run value test, let's check if we can do some `circle2circle-dredd-recipe-test`

Reference model from #11002 : [GELU.zip](https://github.com/Samsung/ONE/files/11962886/GELU.zip)

seanshpark:
steps
- [ ] support `Custom(X)` flexbuffer dump, for debugging
- [ ] produce sub graph model of Gelu with tflchef
- [ ] ...

seanshpark:
In Tensorflow, `tensorflow/core/ops/math_ops.cc`
```cxx
#define UNARY_REAL()                              \
  Input("x: T")                                   \
      .Output("y: T")                             \
      .Attr("T: {bfloat16, half, float, double}") \
      .SetShapeFn(shape_inference::UnchangedShape)

REGISTER_OP("Erf").UNARY_REAL();
```

seanshpark:
current tfldump supports `CustomOp` dump ... but not much for this model
```
O(0:2) CUSTOM(Erf)
    T(0)
    I T(0:5) Gelu/truediv;Gelu/Cast/x1
    O T(0:6) Gelu/Erf
```
`T(0)` is the content of `Custom Erf`

seanshpark:
draft is now ready

seanshpark:
![image](https://github.com/Samsung/ONE/assets/4616940/ab43b096-01e7-4c6b-bceb-3cd0bf2243c2)

seanshpark:
Net_Gelu_000.tflite | Net_Gelu_000.opt.circle
--- | ----
![image](https://github.com/Samsung/ONE/assets/4616940/dbe6d675-8582-436d-aebf-82099d803782) | ![image](https://github.com/Samsung/ONE/assets/4616940/d33a8956-4a41-42d3-a381-8694fb31ad55)

seanshpark:
done

PR #11138, opened 2023-07-20
[common-artifacts] Exclude Net_Gelu for tcgenerate

seanshpark:
This will exclude tcgenerate for upcommiong Net_Gelu models.


seanshpark:
for #11130 , from #11131

mhs4670go (approved):
LGTM

PR #11050, opened 2023-07-12
[runtime] Introduce makefile target for luci build

hseok-oh:
This commit introduces makefile target for luci build to use runtime. It supports native build only.

ONE-DCO-1.0-Signed-off-by: Hyeongseok Oh <hseok82.oh@samsung.com>

PR #11140, opened 2023-07-20
[loco] Install CanonicalNodes.lst file

hseok-oh:
Let's install CanonicalNodes.lst file.

"CanonicalNodes.lst" is used in installed loco header files. 
So if we include installed loco header file, it can make build fail.


hseok-oh:
I found this when I try to build luci for runtime #11050

mhs4670go (approved):
LGTM

PR #11178, opened 2023-07-27
[onert-micro] disable build workflow

chunseoklee:
- This is tentative quick fix to avoid CI fail
- Will be enabled soon


BalyshevArtem (approved):
LGTM

Issue #10930, opened 2023-06-28
Compiler FE: Enable HardSwish Op


seanshpark:
Lets' enable HARD_SWISH Op

- [x] tflchef
- [x] res/TensorFlowLiteRecipes
- [x] tfldump (nothing to do)
- [x] tflite2circle (nothing to do)
- [x] circledump (nothing to do)
- [x] luci.lang
- [x] luci.import
- [x] luci.export
- [x] luci.logex
- [x] luci.service
- [x] luci.partition
- [x] luci.tests
- [x] luci_interpreter
- [x] luci-value-test
- [x] circle-part-value-test (optional)

2'nd step with decomposition
- [x] luci.pass to decompose
- [x] circle2circle
- [x] circle2circle-dredd-recipe-test
- [x] luci-pass-value-test

Additional work
- [x] one-cmds

`luci.pass to decompose`
- to run this Op in TRIX, we need to decompose this to primitive ops like Mul, Add, Rel, Minumum.

seanshpark:
tflite code has this comment in `tensorflow/lite/toco/graph_transformations/identify_hardswish.cc`
```
// This transformation rule tries to identify the HardSwish structure generated
// by tensorflow.
// The formula of hardswish is:
// f(x) = x * relu6((x+3))/6
//
// We look for the following tensorflow subgraph:
// x * tf.nn.relu6(x + np.float32(3)) * np.float32(1. / 6.)
```
- we can assume TF generates decomposed network and tflite converter converts to 1 HardSwish Op.

https://github.com/tensorflow/tensorflow/blob/f0b8b9a1457bcc8d83716b2034d3d6501693c7d0/tensorflow/compiler/mlir/lite/transforms/optimize_patterns.td#L276
- this shows there are 5 patterns if TF

jinevening:
> to run this Op in TRIX, we need to decompose this to primitive ops like Mul, Add, Rel, Minumum.

If the transformation is only needed for TRIX, why don't we implement the decomposition pass in TRIX?

seanshpark:
@shs-park , can you please check done items in this top message?

shs-park:
@seanshpark, Sure =)

shs-park:
Does it also make sense to implement decomposition in ONE and TRIX, both?

seanshpark:
> Does it also make sense to implement decomposition in ONE and TRIX, both?

ONE-compiler and TRIX backends is independent so both having decomposition is logically OK but can be waste of resource.

[...]

PR #11181, opened 2023-07-28
[luci/import] Support HardSwish operation

shs-park:
Support HardSwish operation in luci/import

ONE-DCO-1.0-Signed-off-by: Seungho Henry Park [shs.park@samsung.com](mailto:shs.park@samsung.com)

---

Related to: #10930 
Draft PR: #10985

### DIFF
--- a/compiler/luci/import/src/Nodes/CircleHardSwish.cpp
+++ b/compiler/luci/import/src/Nodes/CircleHardSwish.cpp
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


seanshpark (dismissed):
LGTM

compiler/luci/import/src/Nodes/CircleHardSwish.cpp
```diff
@@ -0,0 +1,69 @@
...
+    case circle::TensorType_FLOAT64:
+      break;
+    case circle::TensorType_FLOAT32:
+      break;
+    case circle::TensorType_INT16:
+      break;
+    case circle::TensorType_UINT8:
+      break;
+    default:
+      return false;
```
seanshpark, line 49: @jinevening , FYI
We may have imports that reject INT8 type. I think it would be good to check other Ops for recent re-quantize feature.

jinevening: I'll check. Thanks!

jinevening: @shs-park Do you have a reason not to support INT8?

seanshpark: @shs-park , if there is no particular reason, we may remove type restriction like Relu.

seanshpark (approved):
LGTM thank you!

jinevening (approved):
LGTM

PR #11177, opened 2023-07-27
[Draft] Fix build warnings over various modules

batcheu:
As result of build with Clang compiler, there were several build warnings that GCC might missed.
Some of them are obvious bugs which can occur 'memory leaks', 'undefined behavior', and others are related with code readability which can prevent a human error.

PR #11190, opened 2023-07-31
[mir] Replace deprecated macro with recommended

batcheu:
The macro 'INSTANTIATE_TEST_CASE_P' from GTest is deprecated.
It is recommended to use 'INSTANTIATE_TEST_SUITE_P' instead.

---
FYI, you can check the history of renaming of the macro in below link.

* Issue from GoogleTest
https://github.com/google/googletest/issues/1085

* Summary of the above issue
https://github.com/Samsung/ONE/pull/11190#issuecomment-1661274707

Related: #11177

### DIFF
--- a/compiler/mir/unittests/ShapeRange.cpp
+++ b/compiler/mir/unittests/ShapeRange.cpp
@@ -56,7 +56,7 @@ TEST_P(ShapeIteratorTest, ElementCount)
 std::vector<ParamType> test_data{ParamType{6, 1, 2, 3}, ParamType{16, 2, 2, 4},
                                  ParamType{1, 1, 1, 1, 1, 1}, ParamType{5, 5, 1, 1, 1, 1, 1}};
 
-INSTANTIATE_TEST_CASE_P(SimpleInput, ShapeIteratorTest, ::testing::ValuesIn(test_data));
+INSTANTIATE_TEST_SUITE_P(SimpleInput, ShapeIteratorTest, ::testing::ValuesIn(test_data));
 
 TEST(ShapeRange, Contains)
 {


seanshpark:
@batcheu link is too long to read. why do we have to change this?

seanshpark:
http://google.github.io/googletest/reference/testing.html
- this explains about `INSTANTIATE_TEST_SUITE_P`
- there is no explain about `INSTANTIATE_TEST_CASE_P`
- cannot find about rename but can assume name has changed from `CASE` to `SUITE`

seanshpark:
@batcheu , commit message has
```
The macro 'INSTANTIATE_TEST_CASE_P' from GTest is deprecated.
It is recommended to use 'INSTANTIATE_TEST_SUITE_P' instead.
```
I prefer use this sentence as is in the first commit.
Current message gives indirect message and asks reviewers to read heavy document which I don't think is a good for review.

batcheu:
> @batcheu , commit message has
> 
> ```
> The macro 'INSTANTIATE_TEST_CASE_P' from GTest is deprecated.
> It is recommended to use 'INSTANTIATE_TEST_SUITE_P' instead.
> ```
> 
> which expresses directly what this change wants to do.
> 
> I prefer use this sentence as is in the first comment message. Current message gives indirect message and asks reviewers to read heavy document which I don't think is a good for review.

Sorry, I didn't notice that the comment in cover isn't clear as much as commit message.
I'll write description more clearer and readable in next PR :) 

Thanks for your review and comment 👍

batcheu:
> http://google.github.io/googletest/reference/testing.html
> 

FYI, Iet me summarize the above changes.
The maco 'INSTANTIATE_TEST_CASE_P' was used to define set of tests.
But, generally the word 'test case' represents a single test which check the certain behavior if it works like as expected with preconditions and some inputs.
So, they decided to use the ward 'tese suite' instead of 'test case' for the set of tests.
The change was applied into upstream in May, 2017, and the previous macro was marked as deprecated since then.
Both of macros provide same functionality at the moment, but sooner or later the depcrecated macro will be **deprecated** :)

seanshpark:
@batcheu , I think you've missed my comment at https://github.com/Samsung/ONE/pull/11190#issuecomment-1658015339 . Can you please update the first comment to your commit message?

batcheu:
> @batcheu , I think you've missed my comment at [#11190 (comment)](https://github.com/Samsung/ONE/pull/11190#issuecomment-1658015339) . Can you please update the first comment to your commit message?

Sorry for late response 😢 
I updated the cover letter as you advised.

seanshpark (approved):
LGTM thank you!

PR #11200, opened 2023-08-01
[onert-micro] Fix build error for arm

chunseoklee:
- fix error: invalid conversion from 'const int*' to 'const int32_t*'


BalyshevArtem (approved):
LGTM, thank you!

PR #11207, opened 2023-08-03
[cker] Introduce a kernel that computes gradient of relu

ragmani:
This commit introduces a kernel that computes gradient of relu.
    

hseok-oh (approved):
Calculation looks good. 
Is it correct function naming?

ragmani:
> @jyoungyun @glistening Is ReLUGrad correct function naming on training implementation convention? I'm not good at this.

It seems so. I followed the name of raw ops in tensorflow. Rather, the previously decided name "derivative" is not appropriate. Actually, it is computed gradients back-propagated.

jyoungyun (approved):
LGTM

glistening (approved):
LGTM

PR #11208, opened 2023-08-03
[onert] Allow to use multiple GradientAppliers

ragmani:
This commit allows to use multiple GradientAppliers in a TrainableFnSequence.

### DIFF
--- a/runtime/onert/backend/train/KernelGenerator.cc
+++ b/runtime/onert/backend/train/KernelGenerator.cc
@@ -74,6 +74,15 @@ ops::PoolType convertPoolType(ir::operation::Pool2D::PoolType type_ir)
       throw std::runtime_error("train KernelGenerator : Not supported operation yet");
   }
 }
+
+std::unique_ptr<ops::GradientApplier>
+generateGradientApplier(const std::shared_ptr<exec::train::optimizer::Optimizer> optimizer,
+                        const IPortableTensor *gradient, ITrainableTensor *trainable)
+{
+  auto update_fn = std::make_unique<ops::GradientApplier>();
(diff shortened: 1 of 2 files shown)


glistening:
@ragmani Could you please let me know when multiple gradient appliers are required?

ragmani:
@ragmani Could you please let me know when multiple gradient appliers are required?

This change is to able to update multiple trainable tensors in one operation. FC and Conv2D have two trainable tensors (weights, bias).

runtime/onert/backend/train/KernelGenerator.cc
```diff
@@ -85,8 +94,9 @@ std::unique_ptr<exec::train::TrainableFnSequence> KernelGenerator::generate(ir::
   assert(_return_fn);
   ret->append(std::move(_return_fn));
 
-  if (_update_fn)
-    ret->append(std::move(_update_fn));
+  for (auto &&update_fn : _update_funcs)
+    ret->append(std::move(update_fn));
+  _update_funcs.clear();
```
jyoungyun, line 99: Why do you reset the `_update_funcs`? Do you consider the case of repeating backwarding several times?

jyoungyun (approved):
LGTM

runtime/onert/backend/train/KernelGenerator.cc
ragmani: It's my mistake. It is not necessary.

ragmani: This method is called for every node to be generated a `TrainableFnSequence`. So `_update_funcs` must be empty before this method called for the next node.

glistening (approved):
LGTM

PR #11179, opened 2023-07-27
[one-cmds] Integrate 'circle-mpqsolver' with 'onecc'

stamalakhov:
This commit implements integration of 'circle-mpqsolver' with 'onecc'.

for sample [one-quantize] section
```
[one-quantize]
input_path=/home/stanislav/repos/WorkSpaces/ampq/mobilenet_v1_1.0_224.opt.circle
output_path=/home/stanislav/repos/WorkSpaces/ampq/model.q_opt.circle
input_data=/mnt/storage/datasets/ms_coco/dataset_MB_v1_10.h5
ampq=True
verbose=True
save_intermediate=True
ampq_qerror_ratio=0.1
ampq_algorithm=bisection
bisection_type=auto
```
it produced the following log for `circle-mpqsolver`:

```
Recording 0'th data
Recording 1'th data
Recording 2'th data
Recording 3'th data
Recording 4'th data
Recording 5'th data
Recording 6'th data
Recording 7'th data
Recording 8'th data
Recording 9'th data
Recording finished. Number of recorded data: 10
>> Searching mixed precision configuration 
model:/home/stanislav/repos/WorkSpaces/ampq/mobilenet_v1_1.0_224.opt.minmax_recorded.circle
dataset: /mnt/storage/datasets/ms_coco/dataset_MB_v1_10.h5
input dtype: uint8
output dtype: uint8
algorithm: bisection (auto)
qerror metric: MAE
target qerror ratio: 0.1

>> Computing baseline qerrors
Full int16 model qerror: 9.27323e-05
Full uint8 model qerror: 0.000254375
Target qerror: 0.000108897

>> Running bisection(auto) algorithm
Qerror of front half: 147884
Qerror of rear half: 45008.1
Front part will be Q16, while the rear will be Q8

Looking for the optimal configuration in [0 , 32] depth segment
Qerror at depth 16 is 0.000130434 > target qerror (0.000108897)
Looking for the optimal configuration in [16 , 32] depth segment
Qerror at depth 24 is 0.000107807 < target qerror (0.000108897)
Looking for the optimal configuration in [16 , 24] depth segment
Qerror at depth 20 is 0.000124079 > target qerror (0.000108897)
Looking for the optimal configuration in [20 , 24] depth segment
Qerror at depth 22 is 0.000130697 > target qerror (0.000108897)
Looking for the optimal configuration in [22 , 24] depth segment
Qerror at depth 23 is 0.000113433 > target qerror (0.000108897)
Found the best configuration at depth 24
[...]

Issue #11146, opened 2023-07-21
[circle-mpqsolver] Implement onecc integration


stamalakhov:
### What

Implement integration of `circle-mpqsolver` with `onecc`

### Why

Because users of `circle-mpqsolver` need to run it using `onecc`.

stamalakhov:
@jinevening 
For a user of onecc is it supposed to be a subsection in `one-quantize` section?
Or it should be separate section of `onecc` configuration?
if it would be a separate section how about `one-mpqsolve` name for it?

jinevening:
I think It would be better to make ampq feature as an option of `one-quantize`. If we make a separate section`one-mpqsolve`, the output would be the quantized circle, which is the same with `one-quantize`. It would be desirable to keep the same interface to generate the quantized circle.

Many options in circle-mpqsolver are compatible with existing options in one-quantize.

circle-mpqsolver options
```
-h, --help              Show help message and exit
--version               Show version information and exit
-V, --verbose           output additional information to stdout or stderr
--data                  Path to the test data
--data_format           Test data format (default: h5)
--qerror_ratio          quantization error ratio ([0, 1])
--bisection             Single optional argument for bisection method. Whether input
                         node should be quantized to Q16: 'auto', 'true', 'false'.
--input_model           Input float model with min max initialized
--input_dtype           Data type of quantized model's inputs (default: uint8)
--output_dtype          Data type of quantized model's outputs (default: uint8)
--output_model          Output quantized model
--visq_file             *.visq.json file with quantization errors
--save_intermediate     path to save intermediate results
```

So for example,

```ini
[one-quantize]
input_path=fp32 (min/max recorded model may be generated internally)
output_path=shared
input_data=shared
input_data_format=shared
input_type=shared
output_type=shared
save_intermediate=shared
...
ampq=True
ampq_qerror_ratio=
ampq_algorithm=bisection_auto
ampq_bisection_visq=(required for bisection_auto)
...
```

`ampq_` prefix can be used to designate ampq options. If you have another idea, please suggest :)

stamalakhov:
> I think It would be better to make ampq feature as an option of `one-quantize`. If we make a separate section`one-mpqsolve`, the output would be the quantized circle, which is the same with `one-quantize`. It would be desirable to keep the same interface to generate the quantized circle.

Ok. I see. Thank you.

stamalakhov:
Locally tested that `master` branch it's possible now to run one-cc configuration files with `ampq` option enabled.

PR #11209, opened 2023-08-03
[one-cmds] Revisit 'run'

stamalakhov:
This commit implements running of 'one-cmd' command with a user set environment.

Draft: https://github.com/Samsung/ONE/pull/11179
Related: https://github.com/Samsung/ONE/issues/11146

### DIFF
--- a/compiler/one-cmds/onelib/utils.py
+++ b/compiler/one-cmds/onelib/utils.py
@@ -134,7 +134,7 @@ def safemain(main, mainpath):
         sys.exit(255)
 
 
-def run(cmd, err_prefix=None, logfile=None):
+def run(cmd, err_prefix=None, logfile=None, env=None):
     """Execute command in subprocess
 
     Args:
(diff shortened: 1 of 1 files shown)


compiler/one-cmds/onelib/utils.py
```diff
@@ -134,15 +134,16 @@ def safemain(main, mainpath):
         sys.exit(255)
 
 
-def run(cmd, err_prefix=None, logfile=None):
+def run(cmd, err_prefix=None, logfile=None, env=None):
```
jinevening, line 137: This change is to turn on logs for circle_mpqsolver.

Other tools control logs using `--verbose` option. Do you have any reason for directly passing env variables?

stamalakhov: > This change is to turn on logs for circle_mpqsolver.
> 
> Other tools control logs using `--verbose` option. Do you have any reason for directly passing env variables?

@jinevening 
Yes. It's just for logging. Would it be better to implement logging by `verbose` option?

jinevening: Yes. It would be better to keep the interface consistent.

stamalakhov: > Yes. It would be better to keep the interface consistent.

Ok. Thank you.

PR #11219, opened 2023-08-07
[onert] Use trainable graph as a class member variable in lowered trainable graph

jyoungyun:
This commit uses trainable graph as a class member variable instead of reference in lowered trainable graph.

### DIFF
--- a/runtime/onert/core/include/compiler/train/LoweredTrainableGraph.h
+++ b/runtime/onert/core/include/compiler/train/LoweredTrainableGraph.h
@@ -68,7 +68,7 @@ private:
    *  @note   It uses copy of graph, not reference.
    *          It allows the original graph can be compiled multiple times.
    */
-  ir::train::TrainableGraph &_trainable_graph;
+  ir::train::TrainableGraph _trainable_graph;
   std::shared_ptr<ir::OperationIndexMap<int64_t>> _indexed_ranks;
   compiler::GraphLowerInfo _lower_info_map;
 };


ragmani (approved):
LGTM

runtime/onert/core/include/compiler/train/LoweredTrainableGraph.h
```diff
@@ -68,7 +68,7 @@ class LoweredTrainableGraph : public ILoweredGraph
    *  @note   It uses copy of graph, not reference.
    *          It allows the original graph can be compiled multiple times.
    */
-  ir::train::TrainableGraph &_trainable_graph;
+  ir::train::TrainableGraph _trainable_graph;
```
glistening, line 71: @jyoungyun or @ragmani 

I considered #11219 is for fixing a bug. But this change and the comment above says it is for allowing multiple time compilation. When do we need to compile multiple times for a given TrainableGraph?

jyoungyun: Both commits are required. The `_trainable_graph` disappeared after Executor was initialized. (I don't know the exact time.) Therefore, the reference value cannot be used after compilation step is finished. The deep copy are useful when this graph is copied to the TrainableExecutor.

glistening: @jyoungyun Thank you for explanation. But my curiosity is mainly about _"can be compiled multiple times"_ from the comment.

```cc
 @note   It uses copy of graph, not reference.
         It allows the original graph can be compiled multiple times.
```

glistening (approved):

PR #11223, opened 2023-08-09
[onert] Remove assertion to perform run() multiple times

jyoungyun:
This commit removes assertion to perform run() multiple times.
It will help to measure the same phase repeatedly.

Draft: #11035

### DIFF
--- a/runtime/libs/benchmark/src/Phases.cpp
+++ b/runtime/libs/benchmark/src/Phases.cpp
@@ -104,8 +104,14 @@ void Phases::run(const std::string &tag, const PhaseFunc &exec, const PhaseFunc 
     return;
   }
 
-  assert(_phases.find(tag) == _phases.end());
-  _phases.emplace(tag, phase);
+  // TODO Support to store multiple phase data in one tag
+  // onert_train can run training function multiple times.
+  // So, we need to store multiple phase data in one tag.
+  // But, now, it only stores the first phase data.
+  if (_phases.find(tag) == _phases.end())
+  {
+    _phases.emplace(tag, phase);
+  }
 }
 
 } // namespace benchmark


glistening (approved):

runtime/libs/benchmark/src/Phases.cpp
```diff
@@ -104,8 +104,14 @@ void Phases::run(const std::string &tag, const PhaseFunc &exec, const PhaseFunc
     return;
   }
 
-  assert(_phases.find(tag) == _phases.end());
-  _phases.emplace(tag, phase);
+  // TODO Support to store multiple phase data in one tag
+  // onert_train can run training function multiple times.
+  // So, we need to store multiple phase data in one tag.
+  // But, now, it only stores the first phase data.
```
hseok-oh, line 110: Could you show example? I cannot found example on draft. 
Can't we use `loop_num` for training multiple times?

jyoungyun: This code is already merged in master branch.

https://github.com/Samsung/ONE/blob/33aa8abe140c5c1656a7ad78b3cb4cc97f716200/tests/tools/onert_train/src/onert_train.cc#L226-L228

In the epoch loop, the same `EXECUTE` tag is called several times. However, the existing code made an assertion because the same tag has already been generated. I think there will be an error even if I use `loop num`.

hseok-oh: If you need input/expected setting before run and get loss after run for each execution, we can introduce `pre` argument to `Phases::run` function to pass function to call before run. Then we can use `pre` and `post` to handle input/expected and loss. 

Otherwise, we can allow to monitor each phase multiple times and update data if memory usage is larger than before.

`Phases` class is introduce to monitor runtime's memory usage on each phase. IMO, this PR breaks concept of `Phase` class.

jyoungyun: I understood what you want. The `EXECUTE` should be executed only once. However, `onert_train` does not fit in the existing benchmark phases because training must be executed several times depending on the epoch, batch_size and data_length. I will change the behavior of `onert_train` instead of modifying the Phases concept. Thank you for your comment.

jyoungyun: I made a PR to fix the same issue, but the way is different. #11239

jyoungyun:
Closed!

PR #11220, opened 2023-08-07
[DRAFT] Support HardSwish decomposing pass

shs-park:
This PR is on going draft to support decomposing the HardSwish operation in luci/pass.

ONE-DCO-1.0-Signed-off-by: Seungho Henry Park <shs.park@samsung.com>

---

Related to: https://github.com/Samsung/ONE/issues/10930

PR #11234, opened 2023-08-10
[res] Support HardSwish decomposing pass

shs-park:
Add .recipe file and .rule file to test

Those files will be used in luci-pass-value-test and circle2circle-dredd-recipe-test

ONE-DCO-1.0-Signed-off-by: Seungho Henry Park <shs.park@samsung.com>

--- 

Related to: #10930 
Draft PR: #11220


res/TensorFlowLiteRecipes/Net_Decompose_HardSwish_000/test.recipe
```diff
@@ -0,0 +1,17 @@
+operand {
```
seanshpark, line 1: `Net_` prefix recipes are used for a (small) network models, say, with two or more operations.
IMHO, we can just use `HardSwish_001` for decomposition testing.

shs-park: Thanks for pointing out the meaning of `Net_` prefix.
I will change the name =)

shs-park: Applied in 58c7640

seanshpark (approved):
LGTM thank you!

PR #11235, opened 2023-08-10
[luci] Add check before QuantizeSpecialActivation

jinevening:
This checks if node is a quantization target.


seanshpark (approved):
LGTM

PR #11256, opened 2023-08-16
[luci] Remove zeropoint from symmetric quantization

glistening:
It removes zeropoint parameters from symmetric quantization.

### DIFF
--- a/compiler/luci/pass/src/QuantizationUtils.cpp
+++ b/compiler/luci/pass/src/QuantizationUtils.cpp
@@ -73,14 +73,14 @@ void asymmetric_wquant_with_minmax_per_layer(CircleConst *node, float min, float
 }
 
 void symmetric_wquant_with_minmax_per_layer(CircleConst *node, float min, float max,
-                                            float &scaling_factor, int64_t &zp, float &nudged_min,
+                                            float &scaling_factor, float &nudged_min,
                                             float &nudged_max)
 {
   const int32_t kMaxScale = std::numeric_limits<int16_t>::max();
   const int32_t kMinScale = -kMaxScale;
 
   uint32_t size = node->size<loco::DataType::FLOAT32>();
-  compute_sym_scale_zp(min, max, scaling_factor, zp, nudged_min, nudged_max);
+  compute_sym_scale(min, max, scaling_factor, nudged_min, nudged_max);
   const float scaling_factor_inv = 1.0 / scaling_factor;
   std::vector<int32_t> quantized_values(size);
   for (uint32_t i = 0; i < size; ++i)
@@ -101,8 +101,8 @@ void symmetric_wquant_with_minmax_per_layer(CircleConst *node, float min, float 
   }
 }
 
-void compute_sym_scale_zp(float min, float max, float &scaling_factor, int64_t &zp,
-                          float &nudged_min, float &nudged_max)
+void compute_sym_scale(float min, float max, float &scaling_factor, float &nudged_min,
+                       float &nudged_max)
 {
   assert(min <= max);
 
@@ -129,7 +129,6 @@ void compute_sym_scale_zp(float min, float max, float &scaling_factor, int64_t &
   if (scaling_factor < 1e-8)
     scaling_factor = 1e-8;
 
-  zp = 0;
   nudged_min = static_cast<float>(qmin_double * scaling_factor);
   nudged_max = static_cast<float>(qmax_double * scaling_factor);
 }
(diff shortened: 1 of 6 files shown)


compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -129,7 +129,6 @@ void compute_sym_scale_zp(float min, float max, float &scaling_factor, int64_t &
   if (scaling_factor < 1e-8)
     scaling_factor = 1e-8;
 
-  zp = 0;
```
glistening, line 132: - Always `zp = 0` for sym
- `zp` is already allocated (_and initialized as zero_) by `compute_sym_scale_zp` caller.

jinevening: We may need to check caller must set zp to 0.

compiler/luci/pass/src/QuantizationUtils.cpp
```diff
@@ -424,7 +423,7 @@ void quant_const(luci::CircleConst *node, loco::DataType quant_type)
                                               nudged_max);
       break;
     case loco::DataType::S16:
-      symmetric_wquant_with_minmax_per_layer(node, min, max, scaling_factor, zp, nudged_min,
+      symmetric_wquant_with_minmax_per_layer(node, min, max, scaling_factor, nudged_min,
```
jinevening, line 426: checked `int64_t zp{0};`

compiler/luci/pass/src/QuantizeActivation.cpp
```diff
@@ -78,7 +78,7 @@ void QuantizeActivation::visit(luci::CircleNode *node)
     }
     else
     {
-      compute_sym_scale_zp(min, max, scaling_factor, zp, nudged_min, nudged_max);
+      compute_sym_scale(min, max, scaling_factor, nudged_min, nudged_max);
```
jinevening, line 81: `int64_t zp{0};`

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
```diff
@@ -322,7 +322,7 @@ struct QuantizeDequantizeWeights final : public luci::CircleNodeMutableVisitor<v
     }
     else
     {
-      sym_wquant_per_channel(weights, min, max, scaling_factor, zp, nudged_min, nudged_max);
+      sym_wquant_per_channel(weights, min, max, scaling_factor, nudged_min, nudged_max);
```
jinevening, line 325: `zp` is constructed with `std::vector<int64_t> zp(min.size());`.
 
It may be better to update to `std::vector<int64_t> zp(min.size(), 0);` for clarity ?

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -101,7 +101,7 @@ luci::CircleQuantize *create_quantize_op(luci::CircleNode *node, loco::DataType
   else
   {
     assert(out_type == loco::DataType::S16);
-    compute_sym_scale_zp(min, max, scaling_factor, zp, nudged_min, nudged_max);
+    compute_sym_scale(min, max, scaling_factor, nudged_min, nudged_max);
```
jinevening, line 104: int64_t zp{0};

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -434,7 +434,7 @@ void QuantizeWithMinMaxPass::set_input_type(loco::Graph *g) const
       else
       {
         assert(user_given_dtype == loco::DataType::S16);
-        compute_sym_scale_zp(min, max, scaling_factor, zp, nudged_min, nudged_max);
+        compute_sym_scale(min, max, scaling_factor, nudged_min, nudged_max);
```
jinevening, line 437: int64_t zp{0};

compiler/luci/pass/src/QuantizationUtils.cpp
glistening: I already checked existing code already initialized to zero since vector constructor with size initialize as zero.

compiler/luci/pass/src/QuantizeDequantizeWeightsPass.cpp
glistening: As I understand, constructor with size will initialize the member with default value. I don't think the 2nd parmeter is necessary. I personally this form is succinct and enough. But if you prefer, I will add it.

jinevening (approved):
LGTM

hseok-oh (approved):
LGTM

Issue #11202, opened 2023-08-01
[record-minmax] Build fail on ubuntu 20.04


hseok-oh:
Its build fails since #11195

```
/usr/bin/ld: cannot find libtsan_preinit.o: No such file or directory
collect2: error: ld returned 1 exit status
make[2]: *** [compiler/record-minmax/CMakeFiles/record-minmax-for-thread-test.dir/build.make:151: compiler/record-minmax/record-minmax-for-thread-test] Error 1
make[1]: *** [CMakeFiles/Makefile2:19560: compiler/record-minmax/CMakeFiles/record-minmax-for-thread-test.dir/all] Error 2
```

It fails only on 20.04. (not on 18.04, 22.04)

hseok-oh:
It seems that this fail is independent with PR #11195. 
On more old commit, it fails. 

Maybe it is related with ubuntu package update.

seanshpark:
@jinevening , can you guess what may be the problem?
log has `record-minmax-for-thread-test`

seanshpark:
test itself was introduced from #10307

hseok-oh:
I checked nncc build by using docker image that is build yesterday, and it passed...

seanshpark:
Local build on my U20.04 Docker is also OK.

jinevening:
> can you guess what may be the problem?

I guess two reasons.

`record-minmax-for-thread-test` uses `record-minmax` compiled with thread sanitizer (compied with `-fsanitize=thread`). I guess two reasons.

1. `libtsan` is not installed. It is usually installed with gcc, but I guess somehow it is missing. It would be worthwhile to try the below command.

`sudo apt install -y libtsan0`

2. `record-minmax-for-thread-test` has to be turned on in 64 bit machine. If the host machine is not 64 bit, our cmake script may have a bug.

@hseok-oh Could you give more details about the build environment?

hseok-oh:
`libtsan` is installed on local PC

```
$ dpkg -l | grep libtsan
ii  libtsan0:amd64                                10.5.0-1ubuntu1~20.04                                            amd64        ThreadSanitizer -- a Valgrind-based detector of data races (runtime)
ii  libtsan0-arm64-cross                          10.5.0-1ubuntu1~20.04cross1                                      all          ThreadSanitizer -- a Valgrind-based detector of data races (runtime)
```

Docker's `libtsan` version

```
$ docker run -it --rm nnfw/one-devtools:focal-backup /bin/bash
[...]

PR #11268, opened 2023-08-17
Disable record-minmax-thread-safety-test

jinevening:
This PR disables record-minmax-thread-safety test to avoid gcc package bug.

---
Related to: https://github.com/Samsung/ONE/issues/11202

### DIFF
--- a/compiler/record-minmax-thread-safety-test/CMakeLists.txt
+++ b/compiler/record-minmax-thread-safety-test/CMakeLists.txt
@@ -2,6 +2,11 @@ if(NOT ENABLE_TEST)
     return()
 endif(NOT ENABLE_TEST)
 
+# Disable the test if record-minmax-for-thread-test does not exist
+if (NOT TARGET record-minmax-for-thread-test)
+    return()
+endif(NOT TARGET record-minmax-for-thread-test)
+
 # Build record-minmax-for-thread-test if target arch is 64bit
 # Thread sanitizer is only available on 64bit machine
 # (https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual#supported-platforms)
--- a/compiler/record-minmax/CMakeLists.txt
+++ b/compiler/record-minmax/CMakeLists.txt
@@ -23,30 +23,35 @@ if(NOT ENABLE_TEST)
   return()
 endif(NOT ENABLE_TEST)
 
+### 
+### record-minmax-for-thread-test is disabled, because gcc package has a bug.
+### (https://bugs.launchpad.net/ubuntu/+source/gcc-10/+bug/2029910)
+### Let's enable the target after the bug is fixed.
+###
 # Build record-minmax-for-thread-test if target arch is 64bit
 # Thread sanitizer is only available on 64bit machine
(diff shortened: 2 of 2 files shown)


compiler/record-minmax/CMakeLists.txt
```diff
@@ -23,30 +23,35 @@ if(NOT ENABLE_TEST)
   return()
 endif(NOT ENABLE_TEST)
 
+###
```
seanshpark, line 26: plz split this to another PR

compiler/record-minmax-thread-safety-test/CMakeLists.txt
```diff
@@ -2,6 +2,11 @@ if(NOT ENABLE_TEST)
     return()
 endif(NOT ENABLE_TEST)
 
+# Disable the test if record-minmax-for-thread-test does not exist
+if (NOT TARGET record-minmax-for-thread-test)
+    return()
```
seanshpark, line 7: optional)
can you add a message to show that `record-minmax-for-thread-test` is temporary disabled?

compiler/record-minmax/CMakeLists.txt
```diff
@@ -23,30 +23,35 @@ if(NOT ENABLE_TEST)
...
-  target_link_libraries(record-minmax-for-thread-test luci_interpreter)
-  target_link_libraries(record-minmax-for-thread-test dio_hdf5)
-  target_link_libraries(record-minmax-for-thread-test vconone)
-  target_link_libraries(record-minmax-for-thread-test nncc_coverage)
-  target_link_libraries(record-minmax-for-thread-test luci_log)
-
-  target_compile_options(record-minmax-for-thread-test PUBLIC -fsanitize=thread)
-  target_link_libraries(record-minmax-for-thread-test -fsanitize=thread)
-endif("${CMAKE_SIZEOF_VOID_P}" STREQUAL "8")
+# if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "8")
```
seanshpark, line 34: Instead of adding `#` to this whole block,
can you revise this something like
```
if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "8" AND False)
```

hseok-oh: We can use `CMAKE_CXX_COMPILER_VERSION` to disable under gcc-10.

```cmake
if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "8" AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10)
  add_executable(record-minmax-for-thread-test ${DRIVER} ${SOURCES})
....

endif()

jinevening: @hseok-oh Thanks for the suggestion. I'd like to go with @seanshpark 's way, because we know `gcc-10` is ok, but not tested `gcc-11,12,13`. I think it would be safer to simply block the test for a while.

jinevening:
Done.

Issue #11255, opened 2023-08-16
[CFE] Upgrade TensorFlow to 2.12.x


seanshpark:
Let's upgrade TensorFlow to 2.12.1
(Latest is 2.13.0 as of adding this issue)

To do
- [x] Test common-artifacts with TF 2.12.1 (test run on Docker)
   - [x] U18.04
   - [x] U20.04
   - [x] U22.04
- [x] Test one-cmds with TF 2.12.1 (test run on VM, installed + update)
   - [x] U18.04
   - [x] U20.04
   - [x] U22.04
- [x] Update codes
   - [x] add venv-2.12.1 in common-artifacts
   - [ ] revise tests to use venv-2.12.1
   - [x] remove venv-2.8.0
   - [x] remove jammy venv-2.10.1
   - [x] update one-cmds to use 2.12.1

Upgrading source to use TF2.12.1 can be done in next step
- current value test is OK with luci-interpreter

seanshpark:
Check version status for Ubuntu 18.04:
- TBA

seanshpark:
U20.04 Testing

```
VER_TENSORFLOW=2.12.1
VER_ONNX=1.14.0
VER_ONNXRUNTIME=1.15.0
VER_ONNX_TF=1.10.0
VER_PYDOT=1.4.2
...
${VENV_PYTHON} -m pip ${PIP_OPTIONS} install Pillow
${VENV_PYTHON} -m pip ${PIP_OPTIONS} install tensorflow_probability==0.16.0
${VENV_PYTHON} -m pip ${PIP_OPTIONS} install tensorflow_addons==0.20.0
...

--> failed in `prepare_test_materials.sh`
> Archive:  reshape_matmul.zip
  inflating: reshape_matmul.onnx
...
> one-import-onnx: StagingError: in user code:
> IndexError: tuple index out of range


and failed in `one-build` testing
> one-build: AttributeError: module 'importlib' has no attribute 'machinery'

**Comment by seanshpark** [comment]
@mhs4670go , do you know to fix `module 'importlib' has no attribute 'machinery'` error?

**Comment by mhs4670go** [comment]
> @mhs4670go , do you know to fix `module 'importlib' has no attribute 'machinery'` error?

I'm not sure why this happened suddenly but adding below line into `compiler/one-cmds/onelib/utils.py` resolves this problem.

```python
import importlib.machinery
```

seanshpark:
also need this too
```py
import importlib.util
```

[...]

PR #11277, opened 2023-08-17
[one-cmds] Add prefix to I/O names for import onnx

seanshpark:
This will revise one-import-onnx to add alphabet prefix in remap IO names.


seanshpark:
for https://github.com/Samsung/ONE/issues/11255#issuecomment-1681989521

jinevening (approved):
LGTM

PR #11278, opened 2023-08-18
DRAFT Upgrade TF 2.12.1

seanshpark:
on-going draft to upgrade TF 2.12.1


seanshpark:
for #11255

seanshpark:
done

PR #11281, opened 2023-08-20
[one-cmds] Use fixed I/O name for reshape_matmul.onnx

seanshpark:
This will revise to use fixed I/O name without special character for reshape_matmul.onnx model as new TF doesn't allow.


seanshpark:
existing model with TF2.12.1

> WARNING:absl:l.1 is not a valid tf.function parameter name. Sanitizing to l_1.
WARNING:absl:r.1 is not a valid tf.function parameter name. Sanitizing to r_1.

which gives error of

> IndexError: tuple index out of range

New model is OK

jinevening (approved):
LGTM

PR #11285, opened 2023-08-21
[one-cmds] Update input name for reshape_matmul

seanshpark:
This will update input name for using recent updated reshape_matmul. Using T(0:0), T(0:1) is to make independent from naming convention of TF upgrade.


seanshpark:
for https://github.com/Samsung/ONE/issues/11255#issuecomment-1685410862 and https://github.com/Samsung/ONE/issues/11284#issuecomment-1685635465

jinevening (approved):
LGTM Thank you

PR #11297, opened 2023-08-21
[CFE] cmake flag for bionic

seanshpark:
This will provide cmake flag for bionic.


seanshpark:
for #11255 , from draft #11278

seanshpark:
@jinevening ping

PR #11216, opened 2023-08-06
[circle-mpqsolver] Fix'verbose' mode

stamalakhov:
This commit places 'logger' right after possible initialization of 'LUCI_LOG` variable to initialize 'logger' properly.

Draft: https://github.com/Samsung/ONE/pull/11179
Related: https://github.com/Samsung/ONE/issues/11146

ONE-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

PR #11299, opened 2023-08-21
[circle-mpqsolver] Revisit logging

stamalakhov:
This commit replaces zero level logging with always printed messages.

It implements suggestion from https://github.com/Samsung/ONE/pull/11216#discussion_r1285390062.

Running `circle-mpqsolver` without 'LUCI_LOG' being defined produced the following output:
```
>> Searching mixed precision configuration 
model:/home/stanislav/repos/WorkSpaces/mobile_net_v1/model.recorded.circle
dataset: /mnt/storage/datasets/ms_coco/dataset_MB_v1_3.h5
input dtype: uint8
output dtype: uint8
algorithm: bisection (auto)
qerror metric: MAE
target qerror ratio: 0.1

>> Computing baseline qerrors
Full int16 model qerror: 7.68947e-05
Full uint8 model qerror: 0.000288324
Target qerror: 9.80376e-05

>> Running bisection(auto) algorithm
Qerror of front half: 152628
Qerror of rear half: 50324.6
Front part will be Q16, while the rear will be Q8

Looking for the optimal configuration in [0 , 32] depth segment
Qerror at depth 16 is 9.03983e-05 < target qerror (9.80376e-05)
Looking for the optimal configuration in [0 , 16] depth segment
Qerror at depth 8 is 0.000123055 > target qerror (9.80376e-05)
Looking for the optimal configuration in [8 , 16] depth segment
Qerror at depth 12 is 0.000103797 > target qerror (9.80376e-05)
Looking for the optimal configuration in [12 , 16] depth segment
Qerror at depth 14 is 0.000102298 > target qerror (9.80376e-05)
Looking for the optimal configuration in [14 , 16] depth segment
Qerror at depth 15 is 9.56676e-05 < target qerror (9.80376e-05)
Looking for the optimal configuration in [14 , 15] depth segment
Qerror at depth 14 is 0.000102298 > target qerror (9.80376e-05)
Found the best configuration at depth 15
Saving output model to /home/stanislav/repos/WorkSpaces/mobile_net_v1/model.q_opt.circle
```

Draft: https://github.com/Samsung/ONE/pull/11179
Related: https://github.com/Samsung/ONE/issues/11146

### DIFF
--- a/compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
+++ b/compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
@@ -18,9 +18,9 @@
 #include <vconone/vconone.h>
 #include <luci/CircleExporter.h>
 #include <luci/CircleFileExpContract.h>
-#include <luci/Log.h>
 
 #include "bisection/BisectionSolver.h"
+#include <core/SolverOutput.h>
 
 #include <iostream>
 #include <iomanip>
@@ -114,8 +114,6 @@ int entry(int argc, char **argv)
     setenv("LUCI_LOG", "100", 0);
   }
 
-  LOGGER(l);
-
   auto data_path = arser.get<std::string>("--data");
   auto input_model_path = arser.get<std::string>("--input_model");
   auto output_model_path = arser.get<std::string>("--output_model");
@@ -129,11 +127,11 @@ int entry(int argc, char **argv)
     return EXIT_FAILURE;
   }
 
-  VERBOSE(l, 0) << ">> Searching mixed precision configuration " << std::endl
-                << "model:" << input_model_path << std::endl
-                << "dataset: " << data_path << std::endl
-                << "input dtype: " << input_dtype << std::endl
-                << "output dtype: " << output_dtype << std::endl;
+  SolverOutput::get() << ">> Searching mixed precision configuration \n"
+                      << "model:" << input_model_path << "\n"
+                      << "dataset: " << data_path << "\n"
+                      << "input dtype: " << input_dtype << "\n"
+                      << "output dtype: " << output_dtype << "\n";
 
   if (arser[bisection_str])
   {
@@ -145,7 +143,7 @@ int entry(int argc, char **argv)
       auto value = arser.get<std::string>(bisection_str);
       if (value == "auto")
       {
-        VERBOSE(l, 0) << "algorithm: bisection (auto)" << std::endl;
+        SolverOutput::get() << "algorithm: bisection (auto)\n";
         if (!handleAutoAlgorithm(arser, solver))
         {
           return EXIT_FAILURE;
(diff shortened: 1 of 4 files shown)


seanshpark:
@jinevening , I'm against this change. LOG was introduced NOT to use direct printing to somehow filter the message.
Why do we have to change like this?

jinevening:
>  I'm against this change. LOG was introduced NOT to use direct printing with somehow filter the message.
Why do we have to change like this?

The logs modified in this PR are always-printed. `record-minmax` also uses `std::cout` for always printed logs.

```cpp
    std::cout << "Recording " << num_records << "'th data" << std::endl;
```

jinevening:
After offline discussion with @seanshpark we reached agreements at some points.

1. Need to give a method for users to control messages (whether to print them or not). Could you make a class for log like below?

```cpp
class Logger
{
  Logger(bool turn_on);

  static void print(const std::string &msg) {
    if (not _turn_on)
      return;
    std::cout << msg << std::endl;
  }

  static bool _turn_on = true;
};

Logger::print("message")
```

2. Need to make a single point of modification. Please implement the above class and method in the separate files (Logger.h, Logger.cpp)

@stamalakhov Sorry for confusion.

seanshpark:
Additionally it would be nice to have `<<` operator something like...
```cxx
Logger &operator<<(Logger &logger, const std::string &message)
{
  logger.print(message);
  return logger;
}

```

seanshpark:
As I understand, printed results is one of the purpose of this tool,
IMHO, class name looks important so it would be better to use something like `SomethingResults` or some kind instead of just `Logger` to express itself for the purpose.

stamalakhov:
Sorry if i misunderstood something.

compiler/circle-mpqsolver/src/core/SolverOutput.cpp
```diff
@@ -0,0 +1,45 @@
...
+  {
+    std::cout << value;
+  }
+
+  return *this;
+}
+
+void SolverOutput::TurnOn(bool on) { _turn_on = on; }
+
+SolverOutput sout;
```
seanshpark, line 45: I don't think static object is a good thing.
Please use singleton pattern.

stamalakhov: > I don't think static object is a good thing. Please use singleton pattern.

ok

stamalakhov: done

compiler/circle-mpqsolver/src/core/SolverOutput.h
```diff
@@ -0,0 +1,58 @@
...
+   */
+  const SolverOutput &operator<<(float value) const;
+
+  /**
+   * @brief turn on/off actual output
+   */
+  static void TurnOn(bool on);
+
+private:
+  static bool _turn_on;
```
seanshpark, line 55: You can remove `static`

stamalakhov: done

compiler/circle-mpqsolver/src/CircleMPQSolver.cpp
```diff
@@ -129,11 +127,12 @@ int entry(int argc, char **argv)
     return EXIT_FAILURE;
   }
 
-  VERBOSE(l, 0) << ">> Searching mixed precision configuration " << std::endl
-                << "model:" << input_model_path << std::endl
-                << "dataset: " << data_path << std::endl
-                << "input dtype: " << input_dtype << std::endl
-                << "output dtype: " << output_dtype << std::endl;
+  SolverOutput::get().TurnOn(true);
```
seanshpark, line 130: as default is ON, this line looks not needed

stamalakhov: done

seanshpark (approved):
LGTM thank you!

jinevening (approved):

PR #11300, opened 2023-08-21
[luci/svc] Max() with unsigned zero will always be ignored

batcheu:
Using std::max() with unsigned type value and unsigned zero(0u) always return a value greater than 0. 
It fixes the use of std::max() to work as intended.

### DIFF
--- a/compiler/luci/service/src/ShapeInfer_StridedSlice.cpp
+++ b/compiler/luci/service/src/ShapeInfer_StridedSlice.cpp
@@ -305,7 +305,7 @@ StridedSliceParams BuildStridedSliceParams(StridedSliceContext *op_context)
     if ((1 << i) & effective_ellipsis_mask)
     {
       // If ellipsis_mask, set the begin_mask and end_mask at that index.
-      added_ellipsis = std::max(0u, i - ellipsis_start_idx);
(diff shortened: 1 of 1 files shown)


seanshpark:
> It fixes the use of std::max() to work as intended.

Can you explain more about the `intended` thing?

This code block is from TensorFlow and found that lots of variables are `int` type (which would be 64bit) but somehow I used `uint32_t` type.
I think whole method should be update someday.

batcheu:
> > It fixes the use of std::max() to work as intended.
> 
> Can you explain more about the `intended` thing?
> 
> This code block is from TensorFlow and found that lots of variables are `int` type (which would be 64bit) but somehow I used `uint32_t` type. I think whole method should be update someday.

It seems that it tries to set 'added_ellipsis' with value difference between 'i' and 'ellipsis_start_idx' unless it is less then zero, otherwise zero is to be set.
However, the current code will set 'added_ellipsis' with unsigned value of minus (e.g. -1 for 4,294,967,295) value if the value difference is less than zero.

If the whole code will be updated sooner, I'll discard the changes and close this PR.

batcheu:
The issue #11301 will handle this.

PR #11302, opened 2023-08-21
Fix cmake warnings.

lemmaa:
- Match case of module names and function call parameters.
- To remove warnings from cmake version 3.27.2.

ONE-DCO-1.0-Signed-off-by: Sung-Jae Lee <sj925.lee@samsung.com>

### DIFF
--- a/infra/cmake/packages/ProtobufConfig.cmake
+++ b/infra/cmake/packages/ProtobufConfig.cmake
@@ -24,12 +24,12 @@ endfunction(_Protobuf_module_import)
 function(_Protobuf_import)
   # Let's use find_package here not to export unnecessary definitions
   # NOTE Here we use "exact" match to avoid possible infinite loop
-  find_package(protobuf EXACT 3.5.2 QUIET)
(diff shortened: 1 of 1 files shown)


seanshpark:
Lots of 
```
[2023-08-21T18:11:37.186Z] CMake Error at /opt/jenkins_agent/workspace/nnfw/master/pr-nncc-debug/compiler/common-artifacts/CMakeLists.txt:301 (add_custom_command):
[2023-08-21T18:11:37.186Z]   Error evaluating generator expression:
[2023-08-21T18:11:37.186Z]     $<TARGET_FILE:tflchef-file>
[2023-08-21T18:11:37.186Z]   No target "tflchef-file"
```

seanshpark:
Strange... still lots of `No target "tflchef-file"`
`

seanshpark:
Maybe rebase to current master?

lemmaa:
I have confirmed that this fix is problematic. It's better to keep the current 'warning' state until we find a better way.

PR #11314, opened 2023-08-22
[record-minmax] Fix cmake warning

seanshpark:
This will fix cmake warning about if block mis-matching.


jinevening (approved):
LGTM Thank you

PR #11316, opened 2023-08-22
[record-minmax-thread-safety-test] Tidy test on Jammy

seanshpark:
This will tidy test on Jammy venv.


jinevening (approved):
LGTM

PR #11321, opened 2023-08-22
[onert] Use EXPECT gtest interface in cker Loss unittest

jyoungyun:
This commit uses EXPECT gtest interface in cker Loss unittest. ASSERT interface fails fast and aborts the current function. So if the previous test case fails, other function cannot be tested. However, EXPECT interface continues after the failure and reports errors after all tests have been finished. This cker unittest has multiple test cases for one function. Therefore, EXPECT is more useful for performing multipel test cases.

Draft: #11035


hseok-oh (approved):
LGTM

PR #11322, opened 2023-08-22
[vconone] Update copyright year

seanshpark:
This will update copyright year to 2023.
Seems late but better than not fixing.


jinevening (approved):
LGTM

PR #11323, opened 2023-08-22
[onert] Implement shape inferer of Loss Op

jyoungyun:
This commit implements shape inferer of Loss op.
This change is related to the Reduction case. Loss output has a shape equal to the input batch_size. The loss kernel calculates the loss value for each batch data. And the get_loss function returns the average of each batch loss.

Draft: #11035


hseok-oh:
@jyoungyun Please resolve build failure

jyoungyun:
I need to consider other case of Reduction. So I will fix it and make a new PR for this issue later.

PR #11326, opened 2023-08-22
[common-artifacts] Revise AArch64 to use TF2.12.1

seanshpark:
This will revise for AArch64 to use TF2.12.1 and upgrade related packages.


seanshpark:
for #11255 , from draft  #11278

seanshpark:
@lemmaa , PTAL
- Ubuntu 22.04 seems working and will test with Ubuntu 20.04 when ready
- If possible please check with Ubuntu 20.04

seanshpark:
@lemmaa PTAL

lemmaa:
@seanshpark , for now, the custom command for `REQUIREMENTS_OVERLAY_PATH_TF_2_8_0` has been removed, causing issues with aarch64 builds. However, we will be removing that dependency soon, so you can merge it as is.

LGTM.

lemmaa (approved):

PR #11328, opened 2023-08-22

seanshpark:
This will revise one-prepare-venv as
- separate for U18.04 as it uses python3.8
- merge for U20.04 and U22.04 as contents are same


seanshpark:
for #11255 , from draft  #11278

jinevening (approved):
LGTM

Issue #11329, opened 2023-08-23
[onnx-import-onnx] Fix '...is not a valid tf.function parameter name. Sanitizing to...'


seanshpark:
This occurs after upgrade to TF2.12.1.
With error `IndexError: tuple index out of range`

from #11255

seanshpark:
This happens when I/O name has special(?) characters: `.`, ...

seanshpark:
Patch in #11278

seanshpark:
done

PR #11330, opened 2023-08-23
[one-cmds] Refactor _remap_io_names

seanshpark:
This will refactor _remap_io_names method with new TidyIONames class.

### DIFF
--- a/compiler/one-cmds/one-import-onnx
+++ b/compiler/one-cmds/one-import-onnx
@@ -41,6 +41,67 @@ import onelib.utils as oneutils
 sys.tracebacklimit = 0
 
 
+# Class to rename input/output to prevent issues while import ONNX models
+class TidyIONames:
+    def __init__(self, onnx_model):
+        self.input_nodes = []
+        self.output_nodes = []
+        self.remap_inputs = []
+        self.remap_outputs = []
+        self.initializers = []
+        self.onnx_model = onnx_model
+        # some models may have initializers as inputs. ignore them.
+        for initializer in onnx_model.graph.initializer:
+            self.initializers.append(initializer.name)
+
+    def order(self):
+        for idx in range(0, len(self.onnx_model.graph.input)):
+            name = self.onnx_model.graph.input[idx].name
+            if not name in self.initializers:
+                self.input_nodes.append(name)
+                self.remap_inputs.append('i_' + format(idx + 1, '04d') + '_' + name)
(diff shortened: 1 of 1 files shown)


seanshpark:
for #11329 , from draft #11278

seanshpark:
ping @ejjeong

hseok-oh:
@nnfw-bot test all

compiler/one-cmds/one-import-onnx
```diff
@@ -41,6 +41,75 @@ import onelib.utils as oneutils
...
+        for idx in range(0, len(self.onnx_model.graph.output)):
+            name = self.onnx_model.graph.output[idx].name
+            self.output_nodes.append(name)
+            self.remap_outputs.append('o_' + format(idx + 1, '04d') + '_' + name)
+        # change names for graph input
+        for i in range(len(self.onnx_model.graph.input)):
+            if self.onnx_model.graph.input[i].name in self.input_nodes:
+                to_rename = self.onnx_model.graph.input[i].name
+                idx = self.input_nodes.index(to_rename)
+                self.onnx_model.graph.input[i].name = self.remap_inputs[idx]
```
ejjeong, line 74:
```suggestion
```
Question: Is there any specific reason that this block is included twice? :)
(Inside `order` and `update`)

seanshpark: Good catch! looks like copy-paste error :)

hseok-oh:
@nnfw-bot test all

compiler/one-cmds/one-import-onnx
```diff
@@ -41,6 +41,69 @@ import onelib.utils as oneutils
...
 
 
+# Class to rename input/output to prevent issues while import ONNX models
+class TidyIONames:
+    input_nodes = []
+    output_nodes = []
+    remap_inputs = []
+    remap_outputs = []
+    initializers = []
+    onnx_model = None
```
ejjeong, line 51: Ah, another comment:
Is there any reason to use `class variables` instead of `instance variable` here?
These members will be shared by all instances of the class.

From https://docs.python.org/3.8/tutorial/classes.html#class-and-instance-variables

> Generally speaking, instance variables are for data unique to each instance and class variables are for attributes and methods shared by all instances of the class:

seanshpark: Ah, should use class variables :)
Thanks for pointing this !

ejjeong (approved):
LGTM, thank you 😄

PR #11331, opened 2023-08-23
[one-cmds] Introduce _sanitize_io_names step

seanshpark:
This will introduce _sanitize_io_names step in import onnx to filter unacceptable characters in I/O names from onnx import models.


seanshpark:
for #11329 , from draft #11278

ejjeong (commented):
~Hmm I may be doing something wrong, but I'm still getting errors from the internal model 🤔~
~I'll share more details in the internal repo.~

Nevermind :) It works well for the few (not all) models I tested.

ejjeong (approved):
LGTM

I haven't read all the code, but the behavior looks fine.

PR #11332, opened 2023-08-23
[one-cmds] Revise to use python3 for one-prepare-venv

seanshpark:
This will revise one-prepare-venv to use python3.
- one-prepare-venv is for U20.04 and U22.04


seanshpark:
for U18.04, `one-prepare-venv.u1804` is used with `python3.8`

PR #11333, opened 2023-08-23
[one-cmds] Remove one-prepare-venv.u2204

seanshpark:
This will remove not used anymore one-prepare-venv.u2204.


seanshpark:
for  #11255

PR #11337, opened 2023-08-23
[one-cmds] Revise one-prepare-venv for AArch64

seanshpark:
This will revise one-prepare-venv for AArch64
- remove Ubuntu version fron name
- update TF and related packages versions


seanshpark:
for  #11255 , from draft  #11278

lemmaa (approved):
LGTM!

I will follow up with a fix to install `tensorflow_addons`. :)

PR #11338, opened 2023-08-23
[pota-quantization-value-test] Handle weights-only in CMakeLists.txt

glistening:
Now, pota-quantization-value-test can handle weights-only test.


jinevening (approved):
LGTM

PR #11343, opened 2023-08-24
[luci/pass] Fix build warnings

batcheu:
This PR fixes buid warnings from luci/pass module.

---
### Descriptions

The build warnigs were obtained while it compiles the one compiler with Clang 14.0.
The more information about each fixes are in below.

| Commits | Description | Severity(*) |
|---|---|---|
|[[luci/pass] Use std::abs() for float type instead of abs()](https://github.com/Samsung/ONE/commit/b97844c19e987bb5f3baf2544b1d136b4ff31e62) | * The abs() in libc only gets integer parameters.<br> * It should uses std::abs() for float type. | High |
|[[luci/pass] Initialize std::initializer_list with static variable](https://github.com/Samsung/ONE/commit/e8336cfaf203b0b99c018e34329556a31fdd2660)| * Using std::initializer_list with brace-init-list via default member intializer is not safe. <br> * The underlying array will be released as contructor is returning. <br>  * As a workaround, it tries to use a static array to extend life time of underying array. <br>* (Reference) <br> http://eel.is/c++draft/dcl.init.list#6 | High |
|[[luci/pass] Add 'template' keyword](https://github.com/Samsung/ONE/commit/5e9f3c4f8e8c74c5e41eb5042de8d32f081b6a3a)| * Because of the C++ grammar ambiguity, '<' token can be interpreted both of the start of template argument list or a less than operator. <br> * And also, if the type of base class is dependent (e.g. template), then the compiler can't know about the member function. <br> * (Reference) <br> http://www.aerialmantis.co.uk/blog/2017/03/17/template-keywords/ | Low |
|[[luci/pass] Missing default handle for switch statement](https://github.com/Samsung/ONE/commit/36a3dfffbedef483be4f21a277d6233d46d49990)| * The default case is not mandatory in switch statement, but it is necessary to handle the unxexpected cases. | Low |

(*) The severity level is based on personal opinion. Higher level is more harmful :(

### DIFF
--- a/compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
+++ b/compiler/luci/pass/src/ConvertNCHWToNHWCPass.test.cpp
@@ -535,6 +535,8 @@ public:
   luci::CircleMaximum *max = nullptr;
 };
 
+static constexpr std::initializer_list<uint32_t> kDefaultShape = {1, 16, 1, 1};
+
 class MeanGraph final : public SimpleGraph
 {
 protected:
@@ -577,7 +579,7 @@ public:
 private:
   bool _keep_dims = true;
   std::vector<int32_t> _axes = {2, 3};
-  std::initializer_list<uint32_t> _shape = {1, 16, 1, 1};
+  std::initializer_list<uint32_t> _shape = kDefaultShape;
 };
 
 class MinimumGraph final : public SimpleGraph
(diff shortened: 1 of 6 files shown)


seanshpark:
@batcheu , we usually squash merge a PR but this PR looks four separate changes in one PR.
It would be better to separate each commits to a PR...

batcheu:
> @batcheu , we usually squash merge a PR but this PR looks four separate changes in one PR. It would be better to separate each commits to a PR...

Oh sorry, I misunderstood the your offline comment.
I'll update it with separated PRs ! Thanks :)

PR #11345, opened 2023-08-24
[luci/pass] Initialize std::initializer_list with static variable

batcheu:
Initialize the std::initializer_list with brace-init-list via default member intializer is not safe since the underlying array has temporary lifetime bound to constructor.

---

Related with #11343


seanshpark (approved):
LGTM

PR #11346, opened 2023-08-24
[luci/pass] Use std::abs() for float type instead of abs()

batcheu:
The abs() will typecasts float type into integer implictly.
Better to use std::abs() for float type variable.

---

Related with #11343


seanshpark (approved):
LGTM

Issue #11355, opened 2023-08-25
[one-import-onnx] Fix "... is not a valid tf.function parameter name. Sanitizing to..."


seanshpark:
Some models has just number in names and these are also need sanitizing.

previous issue: #11329

seanshpark:
waiting for internal CI release & test results...

seanshpark:
fixed

PR #11356, opened 2023-08-25
[one-cmds] Fix import onnx I/O names with numeric

seanshpark:
This will fix one-import-onnx I/O names starting with numeric.

### DIFF
--- a/compiler/one-cmds/one-import-onnx
+++ b/compiler/one-cmds/one-import-onnx
@@ -70,17 +70,21 @@ class TidyIONames:
         for idx in range(0, len(self.onnx_model.graph.input)):
             name = self.onnx_model.graph.input[idx].name
             if not name in self.initializers:
-                if '.' in name or ':' in name:
+                if '.' in name or ':' in name or name[:1].isdigit():
                     self.input_nodes.append(name)
                     name_alt = name.replace('.', '_')
                     name_alt = name_alt.replace(':', '_')
+                    if name_alt[:1].isdigit():
+                        name_alt = 'a_' + name_alt
                     self.remap_inputs.append(name_alt)
         for idx in range(0, len(self.onnx_model.graph.output)):
             name = self.onnx_model.graph.output[idx].name
-            if '.' in name or ':' in name:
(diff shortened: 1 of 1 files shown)


seanshpark:
for #11355

compiler/one-cmds/one-import-onnx
```diff
@@ -75,13 +75,19 @@ class TidyIONames:
...
         for idx in range(0, len(self.onnx_model.graph.output)):
             name = self.onnx_model.graph.output[idx].name
             if '.' in name or ':' in name:
                 self.output_nodes.append(name)
                 name_alt = name.replace('.', '_')
                 name_alt = name_alt.replace(':', '_')
                 self.remap_outputs.append(name_alt)
+            if name[:1].isdigit():
+                self.output_nodes.append(name)
+                self.remap_outputs.append('a_' + name)
```
jinevening, line 90: What if both conditions are met? For example, if output name is "3.", self.output_nodes will be ["3_", "a_3."]. I guess that it should be ["a_3_"]

seanshpark: Good point. I think two condition blocks won't work as expected for `3.`...
`self.output_nodes` would have to `3.` items and two different remap outputs...
I'll merge two conditions.

jinevening: Actually, `replace` can be always applied, so if-statement seems unnecessary.

```python
              name_alt = name.replace('.','_').replace(':','_')
              if name[:1].isdigit():
                  name_alt = 'a_' + name_alt
              self.output_nodes.append(name)
              self.remap_outputs.append(name_alt)
```

jinevening (approved):
LGTM with minor comment https://github.com/Samsung/ONE/pull/11356#discussion_r1305025260

seanshpark:
> so if-statement seems unnecessary.

Ah, yes, I'll revise this in another PR.

seanshpark:
> > so if-statement seems unnecessary.
> Ah, yes, I'll revise this in another PR.

logically `if` condition is not needed but that will fill remap array with all the I/O nodes,
which intentionally I don't want to do... :)

PR #11357, opened 2023-08-25
[luci/logex] Add missing parentheses

batcheu:
The insertion operator ('<<') has higher precedence than conditional
operator ('?'). It adds a parentheses to keep the order of operators.

### DIFF
--- a/compiler/luci/logex/src/CircleNodeSummaryBuilder.cpp
+++ b/compiler/luci/logex/src/CircleNodeSummaryBuilder.cpp
@@ -68,7 +68,7 @@ bool CircleNodeSummaryBuilder::build(const loco::Node *node, const locop::Symbol
     {
       if (i)
         ss << ",";
-      ss << node->dim(i).known() ? node->dim(i).value() : -1;
+      ss << (node->dim(i).known() ? node->dim(i).value() : -1);
     }
     ss << ">";
     return ss.str();


compiler/luci/logex/src/CircleNodeSummaryBuilder.cpp
```diff
@@ -68,7 +68,7 @@ bool CircleNodeSummaryBuilder::build(const loco::Node *node, const locop::Symbol
     {
       if (i)
         ss << ",";
-      ss << node->dim(i).known() ? node->dim(i).value() : -1;
+      ss << (node->dim(i).known() ? node->dim(i).value() : -1);
```
seanshpark, line 71: this is not shift op... 
what warning/error do we get from which compiler?

batcheu: Oh you're right. I'll recheck the warning. :cry:

batcheu: The full warning message as like following
```
warning: operator '?:' has lower precedence than '<<'; '<<' will be evaluated first [-Wparentheses]
```

The word 'shift' was mistakenly chosen, so I update the commit comment correctly.
Thanks.

seanshpark (approved):
LGTM

PR #11358, opened 2023-08-25
[luci/import] Add a virtual destructor for Base class

batcheu:
The NodeBuilderBase abstract class has a non-virtual destructor implicitly. If a virtual destructor is not defined in the base class, a derived class's destructor won't be called when it's deleted with the pointer to its base class type.


seanshpark:
> derived class's destructor won't be called when it's deleted with the pointer to its base class type.

base class doesn't do anything but you added with `= default` . what's the difference between before / after?

batcheu:
> 

If a base class doesn't have virtual destructor, derived class's destructor won't be called when it is deleted with the pointer of base class type. Well-shown example can be checked in the below article.
https://www.geeksforgeeks.org/virtual-destructor/

seanshpark:
> derived class's destructor won't be called

Ah, yes... :)

seanshpark (approved):
LGTM thank you!

batcheu:
Thanks for rapid review !
:+1:

PR #11359, opened 2023-08-25
[circledump] Add a virtual destructor for Base class

batcheu:
The MetadataPrinter abstract class has a non-virtual destructor implicitly. If a virtual destructor is not defined in the base class, a derived class's destructor won't be called when it's deleted with the pointer to its base class type.


seanshpark (approved):
LGTM

PR #11360, opened 2023-08-25
[circle-eval-diff] Add missing parentheses

batcheu:
The compiler warns using the result of an assignment as a condition without parentheses. Let's add additional parentheses to make sure the condition is intended.


seanshpark:
what's the compiler version?

batcheu:
> what's the compiler version?

I tested with clang++ 14.0

seanshpark (approved):
LGTM

PR #11366, opened 2023-08-28
[one-cmds] Use different output name

jinevening:
This makes tests use different output name to avoid name conflicts.


stamalakhov (approved):
LGTM. Thank you.

Issue #11301, opened 2023-08-21
[luci/service] Update BuildStridedSliceParams method


seanshpark:
Need to update `BuildStridedSliceParams()` in `ShapeInfer_StridedSlice.cpp` file.

From https://github.com/Samsung/ONE/pull/11300#issuecomment-1686043808

code from `tensorflow/lite/kernels/strided_slice.cc`

seanshpark:
done

PR #11369, opened 2023-08-29
DRAFT Fix type in shapeinfer_stridedslice

seanshpark:
on-going draft to fix type in shapeinfer_stridedslice.


seanshpark:
for  #11301

PR #11370, opened 2023-08-29
[luci/svc] Remove type alias

seanshpark:
This will remove redundant type alias in ShapeInfer_StridedSlice.


seanshpark:
`int8`, `int16` and `int32` types are still used in original source code but looks clear with types from `stdint`.

seanshpark:
@jinevening , PTAL

jinevening (approved):
LGTM

PR #11375, opened 2023-08-29
[luci/svc] Revise ShapeInfer_StridedSlice with int64

seanshpark:
This will revise ShapeInfer_StridedSlice to follow original source with int64.

### DIFF
--- a/compiler/luci/service/src/ShapeInfer_StridedSlice.cpp
+++ b/compiler/luci/service/src/ShapeInfer_StridedSlice.cpp
@@ -92,7 +92,7 @@ struct StridedSliceContext
 
   // Equivalent input shape after adding axis according to new_axis_mask.
   loco::TensorShape effective_input_shape;
-  uint32_t input_dims = 0;
+  int64_t input_dims = 0;
 };
 
 // Use until std::clamp() is available from C++17.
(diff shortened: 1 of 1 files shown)


seanshpark:
for  #11301 , from #11369

`uint32_t` was used to match `Dimension` in shape, but would be better to follow original source
and then use `uint32_t` in final step.

seanshpark:
@jinevening PTAL

SlavikMIPT (dismissed):
LGTM

compiler/luci/service/src/ShapeInfer_StridedSlice.cpp
```diff
@@ -433,7 +433,7 @@ loco::TensorShape infer_output_shape(const CircleStridedSlice *node)
   for (uint32_t idx = 0; idx < shape_size; ++idx)
   {
     // reverse copy
-    output_shape.dim(idx) = output_shape_vector.at(shape_size - 1u - idx);
+    output_shape.dim(idx) = static_cast<uint32_t>(output_shape_vector.at(shape_size - 1u - idx));
```
jinevening, line 436: Does our SE tool accepts this pattern? I think downcasting would need range check?

seanshpark:
@jinevening , @SlavikMIPT PTAL

jinevening (approved):
LGTM

seanshpark:
@SlavikMIPT PTAL

seanshpark:
Landing without @SlavikMIPT second approval; first approval was OK.

PR #11387, opened 2023-08-30
DRAFT circle schema 06

seanshpark:
on-going draft to support new tflite/circle schema.


seanshpark:
for #11390

seanshpark:
can not update branch ... change is too big that our gateway rejects update.

seanshpark:
done

PR #11392, opened 2023-08-30
[circledump] Rename variable

seanshpark:
This will rename variable to follow like other ops.


jinevening (approved):
LGTM

PR #11402, opened 2023-08-30
[tools] Introduce tf dataset converter

ragmani:
This commit introduces a tool, tf dataset converter, which converts tensorflow datasets to datasets for `onert_train`. In this commit only mnist datasets were tested.

### DIFF
--- a/tools/generate_datafile/tf_dataset_converter/requirements.txt
+++ b/tools/generate_datafile/tf_dataset_converter/requirements.txt
@@ -1,0 +1,8 @@
+argparse
+numpy
+pathlib
+# Please upgrade pip version before installing these requerements.
+# pip 20.2 and earlier doesn’t have true dependency resolution.
+# Refer to https://pip.pypa.io/en/latest/user_guide/#requirements-files
+tensorflow==2.8.2  # This version specifies the upper bound for protobuf
+tensorflow_datasets==4.7.0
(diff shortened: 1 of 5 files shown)


tools/generate_datafile/tf_dataset_converter/requirements.txt
```diff
@@ -0,0 +1,8 @@
+argparse
+numpy
+pathlib
+# Please upgrade pip version before installing these requerements.
+# pip 20.2 and earlier doesn’t have true dependency resolution.
+# Refer to https://pip.pypa.io/en/latest/user_guide/#requirements-files
+tensorflow==2.8.2  # This version specifies the upper bound for protobuf
```
glistening, line 7: @ragmani Does this tool require this specific version of tensorflow? I guess it could be >= 2.8.2

ragmani: It's complicated to explain why a specific version was used. protobuf issues, tensorflow versions, python versions, the dependency resolution of pip, ...
So, it's convenient to fix python version as 3.8 and use the fixed the version to one of tensorflow versions provided by that python version.

### References
- Breaking change in the protobuf : https://protobuf.dev/news/2022-05-06/
- TensorFlow version with the protobuf issue fixed : https://github.com/tensorflow/tensorflow/releases/tag/v2.8.2
- pip 20.2 and earlier [doesn’t have true dependency resolution](https://github.com/pypa/pip/issues/988)

glistening (approved):
LGTM

hseok-oh (approved):

Issue #11390, opened 2023-08-30
[CFE] circle/tflite schema upgrade


seanshpark:
Let's upgrade circle/tflite schema.

from  #11377 
related #11381

seanshpark:
Tasks
- [ ] add mio-tflite2121
- [ ] add mio0-circle06
- [ ] revise to use mio-tflite2121
- [ ] revise to use mio-circle06
- [ ] enable TransposeConv activation
  - [ ] tflchef
  - [ ] tfldump
  - [ ] circledump
  - [ ] luci import/export
  - [ ] luci-interpreter
- [ ] luci/pass to separate activation from TransposeConv
   - [ ] new pass
   - [ ] circle2circle

seanshpark:
draft is prepared... but our internal model still fails with execution result validation. -_-;

seanshpark:
done for now

PR #11423, opened 2023-08-31
[mio] Introduce mio-circle06

seanshpark:
This will introduce mio-circle06 to provide circle version 0.6.


seanshpark:
for  #11390 , from #11387

jinevening (approved):
LGTM

Issue #11269, opened 2023-08-17
[record-minmax] Implement moving average per image


jinevening:
## What

Let's implement moving average per image.

```
            min_val = min_val + self.averaging_constant * (min_val_cur - min_val)
            max_val = max_val + self.averaging_constant * (max_val_cur - max_val)
```

averaging_constant = 0.01 (default)

## Why

To test if the algorithm is effective for ViT model.

jinevening:
We can implement this by adding two hyperparameters (`moving_avg_batch`, `moving_avg_const`) to the existing `moving_average` mode option.

```bash
onecc quantize ... \
  --mode moving_average
  --moving_avg_batch 1
  --moving_avg_const 0.01
```

jinevening:
Done.

PR #11429, opened 2023-09-01
[one-cmds] Add moving_average parameters

jinevening:
This adds parameters for moving_average in one-quantize.

---
Related to: https://github.com/Samsung/ONE/issues/11269
Draft PR: https://github.com/Samsung/ONE/pull/11409

### DIFF
--- a/compiler/one-cmds/one-quantize
+++ b/compiler/one-cmds/one-quantize
@@ -121,6 +121,18 @@ def _get_parser():
         'maximum percentile (0.0~100.0, default=99.0). Algorithm parameter for calibration. This is valid when calibration algorithm is percentile.'
     )
     quantization_group.add_argument(
+        '--moving_avg_batch',
+        type=str,
+        help=
+        'batch size of moving average (default=16). This is valid when calibration algorithm is moving_average.'
+    )
+    quantization_group.add_argument(
+        '--moving_avg_const',
+        type=str,
+        help=
+        'hyperparameter (C) to compute moving average (default=0.1). Update equation: avg <- avg + C * (curr_batch_avg - avg). This is valid when calibration algorithm is moving_average.'
+    )
+    quantization_group.add_argument(
         '--mode',
         type=str,
         help=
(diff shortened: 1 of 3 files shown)


compiler/one-cmds/one-quantize
```diff
@@ -345,6 +361,31 @@ def _verify_arg(parser, args):
...
+            try:
+                moving_avg_batch = int(getattr(args, 'moving_avg_batch'))
+            except ValueError:
+                parser.error('moving_avg_batch must be integer')
+            try:
+                moving_avg_const = float(getattr(args, 'moving_avg_const'))
+            except ValueError:
+                parser.error('moving_avg_const must be float')
+        else:
+            parser.error('Unsupported mode')
```
jinevening, line 387: This check was added for negative test.

seanshpark (approved):
LGTM

compiler/one-cmds/one-quantize
```diff
@@ -345,6 +361,31 @@ def _verify_arg(parser, args):
...
+            except ValueError:
+                parser.error('min_percentile must be float')
+            try:
+                max_percentile = float(getattr(args, 'max_percentile'))
+            except ValueError:
+                parser.error('max_percentile must be float')
+        elif getattr(args, 'mode') == 'moving_average':
+            # Check dtype
+            try:
+                moving_avg_batch = int(getattr(args, 'moving_avg_batch'))
```
glistening, line 379: `int` may not raise error for float number. You may use `isinstance(..., int)`.

compiler/one-cmds/one-quantize
```diff
@@ -345,6 +361,31 @@ def _verify_arg(parser, args):
...
+            except ValueError:
+                parser.error('moving_avg_batch must be integer')
+            try:
+                moving_avg_const = float(getattr(args, 'moving_avg_const'))
```
glistening, line 383: Also, I am not sure `float` will raise error when `moving_avg_const` is integer.

compiler/one-cmds/one-quantize
jinevening: The type of argument is string, so `int` raises an exception.

```
>>> int('0.1')
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '0.1'
```

You might mean

```
>>> int(0.1)
0
```

compiler/one-cmds/one-quantize
jinevening: Integer value is ok for `moving_avg_const`.

compiler/one-cmds/one-quantize
jinevening: The negative test case in this PR tests if the exception is raised.

compiler/one-cmds/one-quantize
jinevening: Hmm.. there is no check for `moving_avg_const` has to be [0,1]. Let me add the check in `record-minmax` in the next PR.

compiler/one-cmds/one-quantize
glistening: Ah, the argument type is `string`! Thank you for explanation.

compiler/one-cmds/one-quantize
glistening: I see. By `float`, any number is acceptable (including `int`). But it is okay. You use `float` to check whether it is number.

glistening (approved):
LGTM

PR #11432, opened 2023-09-01
[mio] Introduce mio-tflite12121

seanshpark:
This will introduce mio-tflite1212 for tflite version 2.12.1.


seanshpark:
for  #11390 , from #11387

jinevening (approved):
LGTM

nit) the title may need to be changed 12121 -> 2121.

We may need to introduce a separator ex: 2121 -> 2.12.1

seanshpark:
> nit) the title may need to be changed 12121 -> 2121.

thanks! I'll update this when landing

> We may need to introduce a separator ex: 2121 -> 2.12.1

This was cause existing ones had 260 or 280 ... -_-;;;;

PR #11440, opened 2023-09-01
[infra] Add mio-circle06 and mio-tflite2120

seanshpark:
This will add new mio-circle06 and mio-tflite2120 to modules list.


seanshpark:
for #11390

this change is not to break Debian package production.

seanshpark:

jinevening (approved):
LGTM

PR #11445, opened 2023-09-01
[onert] Don't call prepare explicitly in ConvolutionLayer

glistening:
prepare() is supposed to be executed once before run(). But prepare() is overriden method of IFunction::prepare, which is called automatically by KernelGenerator.
Thus, prepare() seems redundant.

### DIFF
--- a/runtime/onert/backend/cpu/ops/ConvolutionLayer.cc
+++ b/runtime/onert/backend/cpu/ops/ConvolutionLayer.cc
@@ -34,7 +34,7 @@ ConvolutionLayer::ConvolutionLayer()
     _paddingType(ir::PaddingType::EXPLICIT), _paddingLeft(0), _paddingTop(0), _paddingRight(0),
     _paddingBottom(0), _strideWidth(0), _strideHeight(0), _dilationWidthFactor(1),
     _dilationHeightFactor(1), _activation(ir::Activation::NONE),
-    _conv_kernel(new nnfw::cker::Conv()), _prepare(false)
+    _conv_kernel(new nnfw::cker::Conv())
 {
   // DO NOTHING
 }
@@ -178,8 +178,6 @@ void ConvolutionLayer::configure(const IPortableTensor *input, const IPortableTe
 
 void ConvolutionLayer::run()
 {
-  prepare();
-
   if (_input->is_dynamic() || _kernel->is_dynamic())
   {
     const auto ifm_shape = _input->getShape().asFeature(_input->layout());
@@ -233,9 +231,6 @@ void ConvolutionLayer::run()
 
 void ConvolutionLayer::prepare()
 {
-  if (_prepare)
-    return;
-
   nnfw::cker::Conv &kernel = *_conv_kernel;
   if (_input->data_type() == OperandType::FLOAT32 && _kernel->is_constant())
   {
@@ -284,7 +279,6 @@ void ConvolutionLayer::prepare()
       throw std::runtime_error{"Conv2D: Int8 dynamic weight is not supported"};
     }
   }
-  _prepare = true;
 }
 
 } // namespace ops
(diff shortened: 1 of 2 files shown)


glistening:
`ConvolutionLayer::prepare()` is called during `BackendContext::genKernels()`.

https://github.com/Samsung/ONE/blob/bd82374c4f6ef070d5697bc82c580013383b90f8/runtime/onert/backend/cpu/BackendContext.cc#L56

Thus, `_prepare` is always `true`, when `prepare()` is called in `ConvolutionLayer::run()`.

I don't know why this code is added. for historical reason, or by mistake?

I tried to find history, but this code is there about 4 years ago, and I cannot find history.

glistening:
```
[2023-09-01T07:35:14.776Z] [  258s] /home/abuild/rpmbuild/BUILD/nnfw-1.25.0/compiler/luci/service/src/ShapeInfer_StridedSlice.cpp: In function '{anonymous}::StridedSliceParams {anonymous}::BuildStridedSliceParams({anonymous}::StridedSliceContext*)':
[2023-09-01T07:35:14.776Z] [  258s] /home/abuild/rpmbuild/BUILD/nnfw-1.25.0/compiler/luci/service/src/ShapeInfer_StridedSlice.cpp:304:59: error: no matching function for call to 'max(long int, int64_t)'
[2023-09-01T07:35:14.776Z] [  258s]   304 |       added_ellipsis = std::max(0l, i - ellipsis_start_idx);
```

ragmani:
>I don't know why this code is added. for historical reason, or by mistake?

Actually, I don't know why either.

ragmani (approved):
LGTM

hseok-oh:
>I don't know why this code is added. for historical reason, or by mistake?

PR: https://github.com/Samsung/ONE/pull/3072
Issue: https://github.com/Samsung/ONE/issues/3053

@glistening Could you check issue is still remained?

glistening:
@hseok-oh Thank you for finding some history.

1. #3053 introduced `prepare()` but it is not the only changes, I suspect the other part of #3053, which solved the regression issue.

2. This PR is equivalent as I described in commit body. I checked `prepare()` is always called from `genKernel()`. Thus, `prepare` from `run()` is not executed at all.

3. I don't figure out what do you want to check.
Running `mobilenet_v1_1.0_224_quant.tflite` before and after this PR?
If so, #3072 (about 3 years ago) issue used `NNAPI=1`.
Do I need to run with `NNAPI=1`?

I wrote this PR to remove seemingly meaningless code, which makes me hard to understand.
It is not critical part for #11457.

Let's keep it open w/o further time putting.

glistening:
In addition, keeping `prepare()` in `run()` may be helpful for dynamic tensor, though current code seems not to execute `prepare()` in `run()` execution.

glistening:
Just for recording

`DepthwiseConvoultionLayer` is differently implemented.

It does not have `prepare()`. 
It has `prepareQ8i()`,`prepareQ8uPerChannel`, ..., which are called from `configure()`.

<sub> <b> runtime/onert/backend/cpu/BackendContext.cc </b> </sub>
```cc
FunctionMap BackendContext::genKernels()
{
  FunctionMap ret;

  for (auto &&op_ind : _data.op_order)
  {
    // configure() is called here
    auto fn_seq = kernel_gen->generate(op_ind);
    ret.emplace_back(op_ind, std::move(fn_seq));
  }

  basic::initConsts(*this);

  // NOTE For memory optimization, we want to free some operand data
  const_cast<ir::Graph &>(*_data.graph)
    .operands()
    .iterate([&](const ir::OperandIndex &, ir::Operand &obj) { obj.releaseData(); });

  for (auto &&it : ret)
  {
    auto &fn_seq = it.second;
    // prepare() is called here
    fn_seq->iterate([&](exec::IFunction &ifunc) { ifunc.prepare(); });
  }

  return ret;
}
```

glistening:
I believe this change is valid at least for the current code base.
But for future (including support dynamic tensor) and readability, I decide to keep `prepare()`. Thank you for review.

PR #11449, opened 2023-09-01
[tflite2circle] Revise to use new mio

seanshpark:
This will revise to use new mio-circle06, mio-tflite2121.


seanshpark:
for #11390 , from #11387

jinevening (approved):
LGTM

Issue #11374, opened 2023-08-29
[circle-mpqsolver] Provide more unit tests


stamalakhov:
### What

Let's introduce more unit-tests for 'circle-mpqsolver`.

### Why
We need to provide greater level of code coverage by unit-tests.

stamalakhov:
@jinevening 
Do we need more tests for `circle-mpqsolver`?

jinevening:
Recent result from http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/693/cobertura/

![image](https://github.com/Samsung/ONE/assets/5449554/ab076eea-96d3-4a19-973d-361b13ff7007)

Things got better than https://github.com/Samsung/ONE/pull/11452#issuecomment-1754262236. I think you can close this issue. Thanks!

PR #11452, opened 2023-09-01
[circle-mpqsolver] Add unittests

stamalakhov:
This commit adds unit tests for sources in core and bisection folders.

Related: #11374
ONE-DCO-1.0-Signed-off-by: s.malakhov

### DIFF
--- a/compiler/circle-mpqsolver/CMakeLists.txt
+++ b/compiler/circle-mpqsolver/CMakeLists.txt
@@ -40,14 +40,25 @@ set(TEST_SOURCES
     "src/pattern/PatternResolver.cpp"
     "src/core/Dumper.cpp"
     "src/core/DumpingHooks.cpp"
+    "src/MPQSolver.cpp"
+    "src/bisection/BisectionSolver.cpp"
+    "src/core/DumpingHooks.cpp"
+    "src/core/SolverOutput.cpp"
+    "src/core/Evaluator.cpp"
 )
 
 nnas_find_package(GTest REQUIRED)
 GTest_AddTest(circle_mpqsolver_test ${TESTS} ${TEST_SOURCES})
 target_include_directories(circle_mpqsolver_test PRIVATE src)
 target_include_directories(circle_mpqsolver_test PRIVATE ${Jsoncpp_INCLUDE_DIRS})
+target_include_directories(circle_mpqsolver_test PUBLIC ${HDF5_INCLUDE_DIRS})
 target_link_libraries(circle_mpqsolver_test ${Jsoncpp_STATIC_LIB})
+target_link_libraries(circle_mpqsolver_test PUBLIC ${HDF5_CXX_LIBRARIES})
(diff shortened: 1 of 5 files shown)


stamalakhov:
@nnfw-bot test nncc-release

stamalakhov:
@jinevening 
Can i split this draft into mergeable PR's?

jinevening:
> Can i split this draft into mergeable PR's?

Yes. There are some duplicate codes, so please remove them before uploading.

stamalakhov:
> > Can i split this draft into mergeable PR's?
> 
> Yes. There are some duplicate codes, so please remove them before uploading.

Ok. Thank you.

seanshpark:
@stamalakhov , I was thinking of unit test with function calls with some hard-coded data...
but it seems you're using HDF5 files... this wasn't expected.
is it not possible with some hard coded data to call target APIs ?

seanshpark:
I've looked into `compiler/circle-mpqsolver/src/core/Evaluator.cpp` file and found that
one important looking method
```
WholeOutput compute_outputs(const luci::Module *module, const std::string &h5file)
```
accepts HDF5 file path... and it's inside anonymous namespace...

from this design, adding unit test like I wrote, provide some hard-coded data, looks impossible.

what I could say is that...
- separate file I/O from data processing
- test data processing in unit tests

stamalakhov:
It's possible only if `DatasetEvaluator` will be extended to receive `raw tensor data` input data, and `BisectionSolver` will be extended to receive raw `DatasetEvaluator`.

seanshpark:
> It's possible only if DatasetEvaluator will be extended to receive raw tensor data input data, and BisectionSolver will be extended to receive raw DatasetEvaluator.

that would be adding APIs that won't be used. I don't expect this is realistic.

stamalakhov:
@jinevening 
Only negative tests can be added for `BisectionSolver` and `Evaluator`.
1. Add them?
2. or m.b. to remove all `hdf5` reading operations from `Evaluator` and rework `BisectionSolver` and `Evaluator` ?
3. or for now there are enough tests?

jinevening:
From http://npuci.mooo.com:8080/job/nnfw/job/master/job/daily-compiler-coverage/656/cobertura/ (10/10 report)

Columns: Files, Classes, Methods, Lines
image

You can add negative tests to increase test coverage. `BisectionSolver.cpp` is not tested at all, so it would be effective to find a way to test it.

It would be better to extract data reader from evaluator, so that we can test evaluator without h5 files. However that will require significant refactoring, so let's leave it as a future work.

stamalakhov:
> You can add negative tests to increase test coverage. `BisectionSolver.cpp` is not tested at all, so it would be effective to find a way to test it.

ok.



> It would be better to extract data reader from evaluator, so that we can test evaluator without h5 files. 

ok.

 Thank you.

stamalakhov:
This one can be closed as new one #11849 is opened.

Issue #11454, opened 2023-09-02
[onert] Support conv2d hybrid


glistening:
Let's support conv2d hybrid.

- weight: int8 symm, per-channel
- input/output: float

It is a child issue of #11047.
The draft is #11318.

PR #11455, opened 2023-09-02
[onert] Check conv2d hybrid validity

glistening:
It adds validation for conv2d hybrid op.
It ensure conv2d weight is per-channel quantized.

Related: #11454, #11318

I added negative test.

```
$ Product/x86_64-linux.debug/out/unittest/nnfw_api_gtest --gtest_filter=GenModelTest.neg_OneOp_Conv2D_I8_Hybrid*
Note: Google Test filter = GenModelTest.neg_OneOp_Conv2D_I8_Hybrid*
[==========] Running 1 test from 1 test suite.
[----------] 1 test from GenModelTest
[ RUN      ] GenModelTest.neg_OneOp_Conv2D_I8_Hybrid_PerLayer
Error during model loading : OperationValidator failed at line 203
Failed model loading as expected.
[       OK ] GenModelTest.neg_OneOp_Conv2D_I8_Hybrid_PerLayer (3 ms)
[----------] 1 test from GenModelTest (3 ms total)

[==========] 1 test from 1 test suite ran. (4 ms total)
[  PASSED  ] 1 test.
```

### DIFF
--- a/tests/nnfw_api/src/one_op_tests/Conv2D.test.cc
+++ b/tests/nnfw_api/src/one_op_tests/Conv2D.test.cc
@@ -276,3 +276,29 @@ TEST_F(GenModelTest, neg_OneOp_Conv2D_I8_NonZero_ZeroPoints)
 
   SUCCEED();
 }
+
+TEST_F(GenModelTest, neg_OneOp_Conv2D_I8_Hybrid_PerLayer)
+{
+  CircleGen cgen;
+  std::vector<int8_t> weight_data{1, 2, 3, 4, 5, 6, 7, 8, 9};
+  uint32_t weight_buf = cgen.addBuffer(weight_data);
+  std::vector<float> bias_data{0, 2, 4};
+  uint32_t bias_buf = cgen.addBuffer(bias_data);
+  int in = cgen.addTensor({{1, 1, 1, 3}, circle::TensorType::TensorType_FLOAT32});
+  // Hybrid does not support per-layer.
+  std::vector<float> weight_scales = {0.5};
+  std::vector<int64_t> weight_zeropoints = {0};
+  int weight = cgen.addTensor({{3, 1, 1, 3}, circle::TensorType::TensorType_INT8, weight_buf},
+                              weight_scales, weight_zeropoints);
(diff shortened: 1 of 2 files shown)


tests/nnfw_api/src/one_op_tests/Conv2D.test.cc
```diff
@@ -276,3 +276,29 @@ TEST_F(GenModelTest, neg_OneOp_Conv2D_I8_NonZero_ZeroPoints)
...
+
+TEST_F(GenModelTest, neg_OneOp_Conv2D_I8_Hybrid_PerLayer)
+{
+  CircleGen cgen;
+  std::vector<int8_t> weight_data{1, 2, 3, 4, 5, 6, 7, 8, 9};
+  uint32_t weight_buf = cgen.addBuffer(weight_data);
+  std::vector<float> bias_data{0, 2, 4};
+  uint32_t bias_buf = cgen.addBuffer(bias_data);
+  int in = cgen.addTensor({{1, 1, 1, 3}, circle::TensorType::TensorType_FLOAT32});
+  // Hybrid does not support per-layer.
```
ragmani, line 288: What is `per-layer`?

ragmani: I realized that it's per-tensor.
(Optional) It would be better to change it to per-tensor.

glistening: I used per-layer because `circle-quantizer` uses the term in source and command line argument. I am okay for both term. But I will change as you suggested.

ragmani: It's just optional. The term is already in use, so I think it's better to consider changing the term layer.

ragmani (approved):
LGTM

hseok-oh:
I don't think we need to block handling `per-tensor` hybrid on `OperationValidator` because it is valid type on TFLite (generated under v2).
Before blocking this type, we should decide to not supporting TFLite file from v1.x any more.

glistening:
@hseok-oh told me `OperationValidator` is for checking it is valid model in term of specification, not for checking runtime capability. I will try to move the check to other location, not in `OperationValidator`.

glistening:
I will make another PR to check in backend (`ConvolutionLayer`).

PR #11456, opened 2023-09-02
[onert/cker] Introduce PortableAsymmetricQuantizeFloats

glistening:
It introduces PortableAsymmetricQuantizeFloats.
Implementation is copied from TensorFlow.

Related: https://github.com/Samsung/ONE/issues/11454, https://github.com/Samsung/ONE/pull/11318

### DIFF
--- a/compute/cker/include/cker/PortableTensorUtils.h
+++ b/compute/cker/include/cker/PortableTensorUtils.h
@@ -144,6 +144,60 @@ inline void PortableSymmetricQuantizeFloats(const float *values, const int size,
   }
 }
 
+inline void PortableAsymmetricQuantizeFloats(const float *values, const int size,
+                                             int8_t *quantized_values, float *scaling_factor,
+                                             int32_t *offset)
+{
+  /* Copied from TensorFlow PortableAsymmetricQuantizeFloats */
+  const int32_t kMinScale = -128;
+  const int32_t kMaxScale = 127;
+  const double qmin_double = kMinScale;
+  const double qmax_double = kMaxScale;
+  const auto minmax = std::minmax_element(values, values + size);
+  const double rmin = static_cast<double>(std::min(0.0f, *minmax.first));
+  const double rmax = static_cast<double>(std::max(0.0f, *minmax.second));
+  if (rmin == rmax)
+  {
+    memset(quantized_values, 0, size * sizeof(int8_t));
+    *scaling_factor = 1;
+    *offset = 0;
+    return;
+  }
+  else
+  {
+    double scale = (rmax - rmin) / (qmax_double - qmin_double);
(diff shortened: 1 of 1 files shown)


compute/cker/include/cker/PortableTensorUtils.h
```diff
@@ -144,6 +144,60 @@ inline void PortableSymmetricQuantizeFloats(const float *values, const int size,
...
+  const double qmin_double = kMinScale;
+  const double qmax_double = kMaxScale;
+  const auto minmax = std::minmax_element(values, values + size);
+  const double rmin = static_cast<double>(std::min(0.0f, *minmax.first));
+  const double rmax = static_cast<double>(std::max(0.0f, *minmax.second));
+  if (rmin == rmax)
+  {
+    memset(quantized_values, 0, size * sizeof(int8_t));
+    *scaling_factor = 1;
+    *offset = 0;
```
ragmani, line 163: `offset`(zero point) cannot be 0 just because of `rmin == rmax`. `offset`(zero point) is a value that indicates where the float value 0 is located in the range of quantized values.

compute/cker/include/cker/PortableTensorUtils.h
```diff
@@ -144,6 +144,60 @@ inline void PortableSymmetricQuantizeFloats(const float *values, const int size,
...
+    {
+      nudged_zero_point = kMinScale;
+    }
+    else if (zero_point_double >= qmax_double)
+    {
+      nudged_zero_point = kMaxScale;
+    }
+    else
+    {
+      nudged_zero_point = static_cast<int8_t>(round(zero_point_double));
```
ragmani, line 187: ditto. The value just cast from the double value 0 cannot be a zero point.

compute/cker/include/cker/PortableTensorUtils.h
glistening: As I understand  this code from tensorflow lite, it is the case all values are zero. It looks okay.

ragmani: I figured out that this condition applies only if all inputs are `0.0f` because of the following code.
```c++
  const double rmin = static_cast<double>(std::min(0.0f, *minmax.first));
  const double rmax = static_cast<double>(std::max(0.0f, *minmax.second));
```

compute/cker/include/cker/PortableTensorUtils.h
ragmani: After I calculated it by directly entering values, I figure out that `zero_point_double` is zero point, not `0.0f`.
I'm sorry for the confusion.

ragmani (approved):
LGTM

hseok-oh (approved):

PR #11461, opened 2023-09-03
[tfldump] Print TransposeConv ActFunc

seanshpark:
This will enable to print TransposeConv activation function.

### DIFF
--- a/compiler/tfldump/src/OpPrinter.cpp
+++ b/compiler/tfldump/src/OpPrinter.cpp
@@ -643,6 +643,8 @@ public:
       os << "Padding(" << params->padding() << ") ";
       os << "Stride.W(" << params->stride_w() << ") ";
       os << "Stride.H(" << params->stride_h() << ") ";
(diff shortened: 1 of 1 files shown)


seanshpark:
for #11390 , from #11387

ragmani (approved):
LGTM

jinevening (approved):
LGTM

jinevening (approved):
LGTM

BTW, what happens if we dump tflite of previous schema? Will it raise an error or just skip printing?

seanshpark:
> what happens if we dump tflite of previous schema? Will it raise an error or just skip printing?

when FB reads a file with old schema version it reads as default value,
which would be `NONE` for TransposeConv activation.
This is a dump of old tflite file with new tfldump
```
O(0:0) TRANSPOSE_CONV
    Padding(0) Stride.W(1) Stride.H(1) Activation(NONE)
    I T(0:0) out_shape
    I T(0:1) ker
    I T(0:2) ifm
    O T(0:3) ofm
```

PR #11466, opened 2023-09-04
[record-minmax] Update function signature

jinevening:
This updates function signature of update_qparam.

### DIFF
--- a/compiler/record-minmax/include/MinMaxComputer.h
+++ b/compiler/record-minmax/include/MinMaxComputer.h
@@ -17,7 +17,12 @@
 #ifndef __RECORD_MINMAX_MINMAXCOMPUTER_H__
 #define __RECORD_MINMAX_MINMAXCOMPUTER_H__
 
-#include "MinMaxObserver.h"
+#include "MinMaxVectors.h"
+
+#include <luci/IR/CircleNode.h>
+
+#include <unordered_map>
+#include <memory>
 
 namespace record_minmax
 {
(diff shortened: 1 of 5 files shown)


compiler/record-minmax/include/MinMaxComputer.h
```diff
@@ -33,8 +38,8 @@ class MinMaxComputer
   virtual ~MinMaxComputer() = default;
 
   // Child class must implement this
-  // TODO Use proper input signature
```
jinevening, line 36: This PR resolves this TODO.

`MinMaxObserver` and `MinMaxComputer` are not coupled anymore, which makes it easy to write unittests for MinMaxComputer.

compiler/record-minmax/include/MinMaxComputer.h
```diff
@@ -33,8 +38,8 @@ class MinMaxComputer
...
-  virtual void update_qparam(MinMaxObserver *observer) = 0;
+  virtual void
+  update_qparam(const std::unordered_map<const luci::CircleNode *, MinMaxVectors> *minmax_map) = 0;
```
seanshpark, line 42: (optional) may look better with alias , something like
```cxx
using MapNode2MinMaxVectors = std::unordered_map<const luci::CircleNode *, MinMaxVectors> *;
```

seanshpark (approved):
LGTM

compiler/record-minmax/include/MinMaxComputer.h
jinevening: Hmm.. I'd like to avoid `using` in the header file. https://jmmv.dev/2013/12/header-files-avoid-c-using-directives.html#the-rule

PR #11467, opened 2023-09-04
[luci] Support TransposeConv activation

seanshpark:
This will revise to support TransposeConv activation.

### DIFF
--- a/compiler/luci/pass/src/FuseBatchNormWithTConvPass.cpp
+++ b/compiler/luci/pass/src/FuseBatchNormWithTConvPass.cpp
@@ -87,6 +87,9 @@ bool fused_batch_norm_with_tconv(luci::CircleAdd *add)
     return false;
   if (not luci::fill(&scale, &tconv).with_commutative_args_of(mul))
     return false;
+  // skip if tconv has fused activation
+  if (tconv->fusedActivationFunction() != luci::FusedActFunc::NONE)
+    return false;
 
   // check scale and shift constant attributes
   // TODO maybe rank check is not needed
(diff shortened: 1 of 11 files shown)


seanshpark:
for #11390 , from #11387

compiler/luci/pass/src/FuseBatchNormWithTConvPass.cpp
```diff
@@ -215,6 +218,7 @@ bool fused_batch_norm_with_tconv(luci::CircleAdd *add)
   fused_tconv->stride()->h(tconv->stride()->h());
   fused_tconv->stride()->w(tconv->stride()->w());
   fused_tconv->name(name + "/TransposeConv");
+  fused_tconv->fusedActivationFunction(tconv->fusedActivationFunction());
```
jinevening, line 221:
```suggestion
  // TODO Remove following ReLU Ops and set proper fused activation function
  fused_tconv->fusedActivationFunction(luci::FusedActFunc::NONE);
```

jinevening: `tconv->fusedActivationFunction()` is always NONE, so there will be no problem, but logically fused tconv should not have the same activation function with tconv.

seanshpark: It's not always NONE. One of our internal model is RELU with new TF2.12.1, which I first started upgrading the schema version.

seanshpark: Thanks!, I'll update with NOTE.

jinevening (approved):
LGTM

This change seems incompatible with the September release of our backend compiler (which is synced with circle schema version 5).

seanshpark:
> This change seems incompatible with the September release of our backend compiler

Yes, our September release will go with TF2.8.0 (1.24.0).

jinevening:
More details.

This change will make `FuseActivationFunctionPass` (included in O1) fuse activation functions to the preceding TConv. Our backend compiler may not handle such a model correctly.

seanshpark:
> This change will make FuseActivationFunctionPass (included in O1) fuse activation functions to the preceding TConv

In upcoming PRs, `FuseActivationFunctionPass` for `TransposeConv` will be canceled.
And `XpSepActFromTransposeConvPass` will be introduced to separate fused activation from `TransposeConv`. This Pass will be removed someday when all backends supports fused activation.

PR #11468, opened 2023-09-04
[tflite2circle] Set TransposeConv activation

seanshpark:
This will enable to set TransposeConv activation.


seanshpark:
for #11390 , from #11387

jinevening (approved):
LGTM

PR #11481, opened 2023-09-05
[onert] Fix Conv2D per-channel check condition

glistening:
It fixes wrong condition and name in per-channel quantization.


hseok-oh (approved):
LGTM

PR #11500, opened 2023-09-07
[luci/pass] Introduce XpSepActFromTransposeConvPass

seanshpark:
This will introduce XpSepActFromTransposeConvPass that will separate activation function from TransposeConv.

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/XpSepActFromTransposeConvPass.h
+++ b/compiler/luci/pass/include/luci/Pass/XpSepActFromTransposeConvPass.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


seanshpark:
for #11390 , from #11387

jinevening:
Better to apply this only to fp32 tconv.

seanshpark:
> Better to apply this only to fp32 tconv.

Can I get the reason for this?

jinevening:
The current code does not handle qparam correctly.

seanshpark:
> The current code does not handle qparam correctly.

Can you please describe it more? I don't understand your point.

jinevening:
> Can you please describe it more? I don't understand your point.

Sorry. Let me describe more details. This pass decomposes TConv + Relu into TConv and Relu.

If the model was quantized (i.e., TConv + Relu has qparam), the decomposed TConv and Relu should also have proper quantization parameters. However, the current code does not consider setting the qparams.

seanshpark:
@jinevening , PTAL again

jinevening (approved):
LGTM Thank you!

Issue #11280, opened 2023-08-18
[one-quantize] Support name aliasing in quantization configuration file


jinevening:
## What

Let's support aliasing of layer names (ex: "alternate" or other name) in quantization config file.

## Why

@seanshpark 's comment

We have a plan to migrate to onnx2circle tool someday and this tool will have different naming rule.
IMHO, if we can provide multiple names for a node, we can have some room (but complicated) for smooth migration and less confusion.

So, current is like, for example

"names": [ "a", "b", "c" ],
if we can provide alternative names for the node, like, "a" can be "a1" or "a2",

"names": ["a", "b", "c" ],
"alternate": {
  "a" : ["a1", "a2"],
  "b" : ["b1", "b2"],
}
if a node with name "a" is not found then search in alternate
if a node is not found in also in alternate its invalid and stop the tool.

seanshpark:
@jinevening , if it's OK, I'll work on this.

seanshpark:
Prepare test materials
- use `onnx_conv2d_conv2d.onnx` model in `one-cmds`
- prepare two venv with TF2.8.0 and TF2.12.1
- ...

seanshpark:
Test qconfig.json file for `T(0:12)`
```json
{
  "default_quantization_dtype" : "uint8",
  "default_granularity" : "layer",
  "layers" : [
    {
      "names" : [
        "onnx_tf_prefix_Relu_3;Add_1;convolution_1;Const_3"
      ],
      "alternate": [
        {
          "onnx_tf_prefix_Relu_3;Add_1;convolution_1;Const_3":
            [
              "onnx_tf_prefix_Relu_3;Add_1;convolution_1;Const"
            ]
        }
      ],
      "dtype" : "int16",
      "granularity" : "channel"
    }
  ]
}
```

seanshpark:
First approach experiment:
```diff
     struct LayerParam
     {
       std::string name;
       std::string dtype;
       std::string granularity;
+      std::vector<std::string> altnames;
     };
```

seanshpark:
Work items
- How to test this in CI ?
- ...

jinevening:
We may need to determine how to handle the case that there is an existing tensor with the altname. Maybe throw an exception?

seanshpark:
For multiple output testing, let's add Split Op.

seanshpark:
> We may need to determine how to handle the case that there is an existing tensor with the altname. 

Such as this?
[...]

PR #11513, opened 2023-09-08
DRAFT: circle-quantize with alternative names

seanshpark:
on-going draft to support alternative names.


seanshpark:
for #11280

seanshpark:
done

PR #11519, opened 2023-09-08
[luci/profile] Circle Source and Origin refactor

tomdol:
This commit contains the changes which are intended to refactor the code
of the Origin and Source classes in the profiling module(library).

ONE-DCO-1.0-Signed-off-by: Tomasz Dolbniak <t.dolbniak@partner.samsung.com>

### DIFF
--- a/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
+++ b/compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
@@ -32,7 +32,7 @@ protected:
   struct Source
   {
   public:
-    std::string name(void) const { return _name; }
+    const std::string& name(void) const { return _name; }
     void name(const std::string &name) { _name = name; }
 
     uint32_t id(void) const { return _id; }
--- a/compiler/luci/profile/src/CircleNodeOrigin.cpp
+++ b/compiler/luci/profile/src/CircleNodeOrigin.cpp
@@ -80,16 +80,13 @@ class CompositeOrigin final : public luci::CircleNodeOrigin
 public:
   CompositeOrigin() = delete;
 
-  template <typename T> CompositeOrigin(T origins)
+  template <typename T> CompositeOrigin(const T& origins)
   {
     if (origins.size() == 0)
       throw std::invalid_argument("No origins provided");
 
-    for (auto &origin : origins)
-    {
-      if (origin != nullptr)
-        _origins.emplace_back(origin);
-    }
+    std::copy_if(std::begin(origins), std::end(origins), std::back_inserter(_origins),
+                 [](const auto &origin) { return origin != nullptr; });
   }
 
 public:
@@ -99,10 +96,8 @@ public:
 
     for (auto &origin : _origins)
     {
-      for (auto source : origin->sources())
-      {
-        res.emplace(source);
-      }
+      const auto sources = origin->sources();
+      res.insert(std::begin(sources), std::end(sources));
     }
 
     return res;
--- a/compiler/luci/profile/src/CircleNodeOrigin.cpp
+++ b/compiler/luci/profile/src/CircleNodeOrigin.cpp
@@ -18,7 +18,9 @@
 
 #include <loco.h>
 
+#include <algorithm>
 #include <cassert>
+#include <iterator>
 #include <vector>
 
 namespace


compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
```diff
@@ -32,7 +32,7 @@ class CircleNodeOrigin
   struct Source
   {
   public:
-    std::string name(void) const { return _name; }
+    const std::string& name(void) const { return _name; }
```
seanshpark, line 35: plz do not change to `const`

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -80,16 +80,13 @@ class CompositeOrigin final : public luci::CircleNodeOrigin
...
   {
     if (origins.size() == 0)
       throw std::invalid_argument("No origins provided");
 
-    for (auto &origin : origins)
-    {
-      if (origin != nullptr)
-        _origins.emplace_back(origin);
-    }
+    std::copy_if(std::begin(origins), std::end(origins), std::back_inserter(_origins),
```
seanshpark, line 88: plz add header for `copy_if`, `back_inserter`

seanshpark:
plz follow https://github.com/Samsung/ONE/wiki/ONE-Developer%27s-Certificate-of-Origin

seanshpark:
next time, please leave first commit message as is and add a comment for your explanation;
```
The constructor of CompositeOrigin now avoids unnecessary double copy of shared_ptr. Perhaps the current approach was intended to use move semantics but because of the lvalue used in the for loop and the const initializer_list used in the composite_origin helper function (+ the usage of emplace_back with lvalue reference) there were unnecessary copies made. The new approach still copies the shared pointers but it's now done only once.

The name of the Source is now returned by const reference which prevents a new string being created each time the name() method is called.

Finally I've used the STL algorithms in place of raw loops in a few places.
```

seanshpark:
I don't understand why this PR change is `cleanup`

seanshpark:
It would be better to put single context per PR for better understanding.
This PR seems to have multiple changes.

compiler/luci/profile/src/CircleNodeOrigin.cpp
seanshpark: why is this change needed?

compiler/luci/profile/src/CircleNodeOrigin.cpp
```diff
@@ -99,10 +96,8 @@ class CompositeOrigin final : public luci::CircleNodeOrigin
 
     for (auto &origin : _origins)
     {
-      for (auto source : origin->sources())
-      {
-        res.emplace(source);
-      }
+      const auto sources = origin->sources();
+      res.insert(std::begin(sources), std::end(sources));
```
seanshpark, line 100: why is this changed needed?

compiler/luci/profile/src/CircleNodeOrigin.cpp
tomdol: This change just follows the approach which suggests using standard library algorithms over raw loops. It's a simple refactor of the existing code, not backed by any requirement.

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
tomdol: Does this introduce an API break or is this return-by-copy intended? This particular change is an optimisation in a way because it avoids unnecessary copies at the call site.

seanshpark: I want to leave it as-is to make variable writable.

tomdol:
Reasoning behind this PR:
This commit contains the changes which are intended to clean up the code of the Origin and Source classes in the profiling module(library).

The constructor of `CompositeOrigin` now avoids unnecessary double copy of `shared_ptr`. Perhaps the current approach was intended to use move semantics but because of the lvalue used in the for loop and the `const initializer_list` used in the `composite_origin` helper function (+ the usage of `emplace_back` with lvalue reference) there were unnecessary copies made. The new approach still copies the shared pointers but it's now done only once.

The name of the `Source` is now returned by const reference which prevents a new string being created each time the `name()` method is called.

Finally I've used the STL algorithms in place of raw loops in a few places.

compiler/luci/profile/src/CircleNodeOrigin.cpp
tomdol: This is one more approach to refactor the raw loop with a standard library method instead. The `std::set::insert` is able to insert a range of elements by taking 2 iterators.

compiler/luci/profile/include/luci/Profile/CircleNodeOrigin.h
tomdol: The writability of the returned value is controlled by the cv-specifiers of the variable itself, not the return value specification. If the return-by-value is kept we get an unnecessary copy if the caller actually needs to refer to the name but doesn't necessarily need a copy of this internal string.
```
Source s;
std::string name1 = s.name();
const std::string name2 = s.name();
const std::string& name2 = s.name(); // unnecessary copy avoided
```
An example of such situation is looking for a source by name:
```
const luci::Node node = (...);
const auto origin = luci::get_origin(node);
for (const auto* source : origin->sources()) {
    if (source->name() == "some_name") { // the returned string doesn't have to be a copy
        // do something with this source
    }
}
```
and this is exactly what I encountered in the NPU compiler which depends on this code.

seanshpark:
@llFreetimell , as you initiated these codes, I'll follow your opinion.  (personally I prefer existing code)

Issue #11487, opened 2023-09-06
[onert] How to bind NNFW CAPI with Python


blue4683:
Related #11368
 
# Overview

We are currently binding `runtime/onert/api/include/nnfw.h`(NNFW CAPI) with `Python` and almost finish binding `nnfw.h` as it is.

But, as we proceeded with Python binding, we found that we need to think about the binding method.

In `C++`, data can be accessed through a pointer, so processed data can be used even if it is not returned. So `C++` function only returns `NNFW_STATUS`. But, because of this situation, `Python` is hard to use the data processed by `C++`.

We do not fully understand `nnfw.h` so we may not know what `nnfw.h` does. So we are going to proceed with the binding in the direction ONE wants. So we want to discuss about this.

We thought of two ways of binding.

# Way 1

```cpp
// in C++, get input tensorinfo
NNFW_STATUS input_tensorinfo(uint32_t index)
{
    return nnfw_input_tensorinfo(session, index, &ti);
}
```

```python
# in Python, get input tensorinfo
state = session.input_tensorinfo(0)
if state != NNFW_STATUS.NNFW_STATUS_NO_ERROR:
    print('nnfw_input_tensorinfo error')
    return
```

The first is that `Python` only sends the data required for `C++` handling all data. When `Python` sends data to `C++`, `Python` receives `NNFW_STATUS` to verify that it worked properly. If processed data(input, output, …) should be used in Python, it should be imported using functions (`get_input`, `get_output`, …).

# Way 2

```cpp
// In C++, get input tensorinfo
nnfw_tensorinfo input_tensorinfo(uint32_t index, nnfw_tensorinfo ti)
{
    NNFW_STATUS state = nnfw_input_tensorinfo(session, index, &ti);
    if (state != NNFW_STATUS_NO_ERROR)
    {
        exit(1);
    }
    return ti;
}
```

```python
# In Python, get input tensorinfo
ti = nnfw_tensorinfo()
ti = session.input_tensorinfo(0, ti)
# Handle with processed ti in Python
input_elements = num_elems(ti)
input = [0.0] * input_elements
input = session.set_input(0, input, ti, input_elements)
```

The second way is to send the data processed in `C++` to `Python`, too. `C++` sends not `NNFW_STATUS`, but processed data through `C++` function. If the `C++` function returns an inappropriate `NNFW_STATUS`, `C++` handles the error.

# Finish

[...]

Issue #11368, opened 2023-08-28
[onert] Introduce ONE Runtime Python binding


KangInyeong:
## What
Let's package and publish onert's NNFW API after binding with Python.
  

## Why
- ONE provides APIs for efficient AI development across multiple hardware and software platforms.
- However, ONE APIs are written in the C/C++ language, so this makes high barriers to start to develop.
- NNFW API is one of APIs provided from ONE Runtime(onert)
- Binding NNFW API to Python can provide an environment where AI models can be easily built, optimized, and executed within devices in various environments, so it can be highly competitive.

### Expected benefits
 - Even those who know only Python can easily use the ONE Runtime NNFW API.
 - By publishing to PyPI, it will be easy to install and use API within python code.
    
  
## Tasks
- NNFW API Binding with pybind11
- Packaging & Publishing Python module to PyPI
- Test Python API with Rpi4
- Make a tutorial for easily use

glistening:
Welcome! I think new directory of `runtime/onert/python/api` would be introduced as working directory for python API binding. 

@Samsung/one_onert, @hseok-oh, @chunseoklee Please give your opinion.

@KangInyeong, @jaemaning Please don't wait for our decision. You may work in any directory. The location will be reviewed when you push a draft PR. Relocating home for python binding would not require much time. :)

ragmani:
I agree with @glistening. `runtime/onert/python/api` looks good.

I think we can consider the location of `nnfw_api_pybind11_tutorial.py` depending on what role it is.
If you have a plan to implement `nnfw_api_pybind11_tutorial.py` with bound python modules without installing the published python package, it would be a good choice to put it in `runtime/onert/python/api` as well.
Or If you mean the tutorial implemented with the published python package, another location may be better.

Anyway, please don't wait for our decision as @glistening said.

hseok-oh:
IMO, it's enough to use `runtime/onert/python` without `api` directory because python will be used for API only on onert. 

Additionally, we may need to rename  `runtime/onert/api` to `runtime/onert/capi`.

glistening:
When I suggest `python/api`, I was thinking of the following structures:

```
$ tree runtime/onert/python/
runtime/onert/python/
├── api
├── samples
└── tests
```

But I am okay for @hseok-oh's suggestion.

We may put 

- `tests` under `tests/python_api` (in the same level of `tests/nnfw_api`.
- `samples` under `runtime/onert/sample` (though it was for c/c++).

[...]

PR #11536, opened 2023-09-15
DRAFT: Runtime cross-build process including nnfw capi python binding work

phj799:
Related #11368
cross-build the runtime with added NNFW CAPI bindings derived from issue #11487


hseok-oh:
I'll close this because it is outdated.

Issue #11551, opened 2023-09-19
[q-extract-torch] PyTorch Quantization Parameter extract


ssafyWooJeong:
## What
Let's develop `PyTorch Quantization Parameter extract`, a tool to extract quantization parameters in Quantized PyTorch model, to be used by `q-implant`.
#11254 

## Why
* ONE has its own quantization tool, but it has limitations in applying latest quantization techniques implemented in PyTorch.
* There is `q-implant` module which can import quantization parameters into circle, in development.
* We will implement `q-extract-torch`, which can extract quantization parameters for `q-implant`
* Expected benefits
   
  *  The latest quantization techniques implemented in PyTorch can be applied to the circle model
  *  Circle model will have improved accuracy and performance by implanting extracted quantization parameters

## Tasks
* Define `q-extract-torch`'s input arguments.
* Define data format (quantization parameters + weights of the model).
* Implement `q-extract-torch`.
* Mapping names between PyTorch model and Circle model.
* Test `q-extract-torch`

## Structure
```
- q-extract-torch
  - test
    - example modules like res/PyTorchExamples/examples
  - include 
    - circle     → pics 
  - Torch2Circle.py   →   Convert PyTorch to Circle
  - Torch_Circle_Mapper.py    →   Generate mapping data between PyTorch and Circle by using Tensor data
  - Torch_Extractor.py     →   Extract/Save quantization parameters from PyTorch
  - Torch_Extract_Helper.py     →   For convenient use
  - Test_Ruuner.py
```

## Notes
It uses `tflite2circle` and `pics` to make name mapping between PyTorch and Circle.

seanshpark:
this issue looks identical to #11254 . why did you add this?

ssafyWooJeong:
#11254 needs qparam.json and numpy files to implant quantization parameters to circle.
This issue handles python script to export quantization parameters(qparam.json and numpy files) from a PyTorch Model.

seanshpark:
1/ It's a bit hard to understand what this tool from reading the comment
2/ I've talked offline with @jinevening and IIUC , this tool is to 
- extract quantization data from pytorch model
- save the data for `q-implant` tool

thus this tool works with `q-implant` but is not `q-implant`
and using word `export` doesn't seem to be appropriate.
- center of operation for this is `q-implant`
- this tool will extract data from pytorch model
- this tool will save/convert data to `q-implant` tool

jinevening:
[...]

Issue #11254, opened 2023-08-16
[q-implant] Quantization parameter importer


jinevening:
## What

Let's develop `q-implant`, a tool to import quantization parameters generated by external toolchains (ex: PyTorch) to a circle model.

## Why

- ONE has its own quantization tool, but it has limitations in applying latest quantization techniques implemented in various frameworks. Especially, QAT (Quantization-Aware Training) or advanced PTQ (Post-Training Quantization) techniques usually require a significant amount of GPU resources, which is difficult to integrate with ONE. 
- We will implement `q-implant`, which can import quantization parameters gathered by an external quantization framework (ex: PyTorch) to a circle model.
- Expected benefits
  - The latest quantization techniques implemented in various frameworks can be applied to the circle model.
  - Circle model will have improved accuracy and performance.

## Tasks

- Define input format (quantization parameters + weights of the model).
- Implement `q-implant`.
- Test `q-implant`.

## Usage

Example (Assumed input format is json (`param.json`))

```sh
q-implant --input input.circle --param param.json --output output.circle
```

The above command will generate a quantized circle model (`output.circle`) from an fp32 model (`input.circle`) using `param.json`.

batcheu:
Let me ask a (dumb) question. 😅
Can we expect to merge this functionality as a additional option of onecc quantizer ?

jinevening:
> Can we expect to merge this functionality as a additional option of onecc quantizer ?

Very good question :) I think q-implant will be an option of `onecc quantize`. I can't think of another case for now.

jinevening:
## Input format

I'm thinking `.json` + `.npy` files (qparam) as an input to `q-implant`. The json file will describe qparam info, and npy files will have the contents.

For example, let's assume a torch model with a single Op (Conv2d).

Torch code (`permute` is inserted here and there to support NHWC input/output)
```py
import torch
import torch.nn as nn
import torch.nn.functional as F

class Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.ker = torch.tensor([[[[0.311261,0.920864]]]], dtype = torch.float32)
        self.bias = torch.tensor([0.502937], dtype = torch.float32)
        self.layer0 = nn.Conv2d(2, 1, (1,1), stride = (1,1), dilation = (1,1))
        self.layer0.weight = nn.Parameter(self.ker.permute(0,3,1,2))
        self.layer0.bias = nn.Parameter(self.bias)
    def forward(self, ifm):
[...]

PR #11552, opened 2023-09-19
DRAFT: Introduce PyTorch Quantization Parameter Export for q-implant

ssafyWooJeong:
Draft to support PyTorch Quantization Parameter Export for `q-implant`.(#11254)

This PR uses `tflite2circle` and `pics` to make name mapping between PyTorch and Circle.

for #11551 

```
- q-implant
   - script 
     - include
       - circle    → pics 
     - Torch2Circle.py
     - Torch_Circle_Mapper.py
     - Torch_QParam_Exporter.py
     - TorchExtractor.py
```

ONE-DCO-1.0-Signed-off-by: WooJeong Jeong wj6349@gmail.com


seanshpark:
@ssafyWooJeong , when you fire PR, plz leave PR message as-is from commit message and add addition comment for more information.

ssafyWooJeong:
I got it. We will keep that in mind to make following PR.

jinevening:
Are you going to include this inside q-implant? I think this tool has a different role from q-implant, so it would be better to make a separate module. And, there is no test for the codes. It would be better to include tests too.

ssafyWooJeong:
Ok, I understood.
Then close both issue and draft pr and make another one when ready?
And we will think about test too.

jinevening:
Please close only draft.

PR #11554, opened 2023-09-19
[onecc-docker] Revisit onecc-docker command

hseok-oh:
This commit updates `onecc-doecker` command.
- Change ubuntu version to 20.04
- Update python version requirement

### DIFF
--- a/compiler/onecc-docker/debian/control
+++ b/compiler/onecc-docker/debian/control
@@ -9,5 +9,5 @@ Homepage: https://github.com/Samsung/ONE
 Package: onecc-docker
 Architecture: amd64
 Multi-Arch: foreign
-Depends: ${misc:Depends}, ${shlibs:Depends}, python3.8
+Depends: ${misc:Depends}, ${shlibs:Depends}, python3
 Description: On-device Neural Engine docker package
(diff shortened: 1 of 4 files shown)


compiler/onecc-docker/debian/control
```diff
@@ -9,5 +9,5 @@ Homepage: https://github.com/Samsung/ONE
 Package: onecc-docker
 Architecture: amd64
 Multi-Arch: foreign
-Depends: ${misc:Depends}, ${shlibs:Depends}, python3.8
+Depends: ${misc:Depends}, ${shlibs:Depends}, python3
```
dayo09, line 12: :+1:

dayo09: Should the directory name be `compiler/onecc-docker/focal/control`?

jyoungyun (approved):
LGTM

compiler/onecc-docker/debian/control
jyoungyun: @dayo09 This file is located in `onecc-docker/debian/control`. It does not have OS version name in the directory path. :)

dayo09: Oh, my mistake. I read 'debian' as  'bionic'. :sweat:

PR #11563, opened 2023-09-20
[onert-micro] Fix kernels CMake file

BalyshevArtem:
This commit fixes kernels CMakeLists file for onert-micro to move above `if ENABLE_TEST` statement 

for #11264


SlavikMIPT (approved):
LGTM

PR #11572, opened 2023-09-22
[luci-value-test] Raise tolernace for Conv2D_U8_000

seanshpark:
Conv2D_U8_000 sometimes fail cause of random input value.
This will raise tolerance for value difference.


jinevening (approved):
LGTM

PR #11577, opened 2023-09-25
[luci-compute] Download neon-to-sse source

hseok-oh:
Let's add dependency to "NEON2SSE" source. It is used for TFLite x64 computation kernel.


hseok-oh:
I found this when I try x64 tizen build.

seanshpark (approved):
LGTM

Issue #11582, opened 2023-09-25
[luci-interpreter] Handle int32 overflow in SimpleMemoryManager


jinevening:
## What

 Let's handle int32 overflow in SimpleMemoryManager.

## Why

SimpleMemoryManager uses `int32 num_elements()` when computing the size of allocated tensor. Max value of int32 = 2,147,483,647. There is a case that raises overflow in im2col buffer of Conv2D.

jinevening:
Done

PR #11583, opened 2023-09-25
DRAFT: Fix overflow in huge Conv2D

jinevening:
This fixes int32 overflow in huge Conv2D Op.

---
https://github.com/Samsung/ONE/issues/11582

### DIFF
--- a/compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
+++ b/compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
@@ -60,6 +60,17 @@ public:
     return result;
   }
 
+  // TODO Replace num_elements
+  int64_t large_num_elements() const
+  {
+    int64_t result = 1;
+    for (const int32_t dim : _dims)
+    {
+      result *= dim;
+    }
+    return result;
+  }
+
   bool operator==(const Shape &other) const { return _dims == other._dims; }
 
   bool operator!=(const Shape &other) const { return !operator==(other); }
--- a/compiler/luci-interpreter/pal/linux/PALConv2d.h
+++ b/compiler/luci-interpreter/pal/linux/PALConv2d.h
@@ -111,9 +111,11 @@ static inline void SetupScratchpadTensor(luci_interpreter::Tensor *scratchpad,
     const int32_t output_width = output_shape.Dims(2);
 
     auto data_type_size = static_cast<int32_t>(luci_interpreter::getDataTypeSize(input_data_type));
-    int32_t scratchpad_size = batches * output_width * output_height * input_depth * filter_height *
-                              filter_width * data_type_size;
-    luci_interpreter::Shape scratchpad_shape{scratchpad_size};
+    // im2col_shape
+    // data_type_size is added because we use U8 for scratchpad buffer dtype
+    luci_interpreter::Shape scratchpad_shape{batches, output_height, output_width,
+                                             input_depth * filter_height * filter_width,
+                                             data_type_size};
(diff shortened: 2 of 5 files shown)


compiler/luci-interpreter/pal/linux/PALConv2d.h
```diff
@@ -111,9 +111,11 @@ static inline void SetupScratchpadTensor(luci_interpreter::Tensor *scratchpad,
     const int32_t output_width = output_shape.Dims(2);
 
     auto data_type_size = static_cast<int32_t>(luci_interpreter::getDataTypeSize(input_data_type));
-    int32_t scratchpad_size = batches * output_width * output_height * input_depth * filter_height *
-                              filter_width * data_type_size;
```
jinevening, line 115: Overflow happens here.

compiler/luci-interpreter/src/SimpleMemoryManager.cpp
```diff
@@ -30,7 +30,9 @@ void SimpleMemoryManager::allocate_memory(luci_interpreter::Tensor &tensor)
     release_memory(tensor);
   }
   const auto element_size = getDataTypeSize(tensor.element_type());
-  const auto num_elements = tensor.shape().num_elements();
+
+  // Use large_num_elements to avoid overflow
+  const auto num_elements = tensor.shape().large_num_elements();
```
jinevening, line 35: With `num_elements()`, overflow also happens here.

compiler/luci-interpreter/src/SimpleMemoryManager.cpp
```diff
@@ -30,7 +30,9 @@ void SimpleMemoryManager::allocate_memory(luci_interpreter::Tensor &tensor)
...
 
   auto *data = new uint8_t[num_elements * element_size];
```
seanshpark, line 37: Q) no need to check new success/fail? (as we're now allocating large size)

jinevening: Currently, SimpleMemoryManager just throws std::bad_alloc. It would be possible to wrap `new` with try-catch to improve error handling.

seanshpark: >  just throws std::bad_alloc

I think this will do :)

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
```diff
@@ -60,6 +60,17 @@ class Shape
     return result;
   }
 
+  // TODO Replace num_elements
+  int64_t large_num_elements() const
```
glistening, line 64: @jinevening Why do you use signed int here? Does `large_num_elements()` returns negative number? It could be happen when some of `dim` has `-1`, which means `undefined-yet`.

compiler/luci-interpreter/src/SimpleMemoryManager.cpp
glistening: As `num_elements` can be negative, `new` may throw `std::bad_array_new_length`. It could be happen when the shape has `-1`, which means undefined-yet. It seems `luci-interpreter` will not distinguish the error from huge size and undefined shape at this moment.

compiler/luci-interpreter/include/luci_interpreter/core/Tensor.h
jinevening: Just because existing `num_elements()` returns signed int (`int32_t`). I think it is good to use `uint64_t`!

PR #11593, opened 2023-09-26
[q-implant] Quantization Parameter forward propagation

ssafyWooJeong:
Operators(such as transpose, reshape and etc.) can be added while model conversion. Those operators are not present at original model, so corresponding quantization parameter could not exist on `qparam.json`. For those operators, forward quantization parameter to make appropriate graph.

ONE-DCO-1.0-Signed-off-by: WooJeong Jeong <wj6349@gmail.com>

### DIFF
--- a/compiler/q-implant/include/QImplant.h
+++ b/compiler/q-implant/include/QImplant.h
@@ -28,7 +28,7 @@ class QImplant final
 {
 public:
   QImplant(const std::string &path) : _path(path) {}
-  
+
   void write(loco::Graph *g);
 
 private:
--- a/compiler/q-implant/include/QImplant.h
+++ b/compiler/q-implant/include/QImplant.h
@@ -28,13 +28,13 @@ class QImplant final
 {
 public:
   QImplant(const std::string &path) : _path(path) {}
-
   void write(loco::Graph *g);
 
 private:
-  const std::string &_path;
-
   void forward_qparam(loco::Graph *g);
+
+private:
+  const std::string &_path;
 };
 
 } // namespace q_implant
--- a/compiler/q-implant/include/QImplant.h
+++ b/compiler/q-implant/include/QImplant.h
@@ -28,6 +28,7 @@ class QImplant final
 {
 public:
   QImplant(const std::string &path) : _path(path) {}
+  
   void write(loco::Graph *g);
 
 private:
--- a/compiler/q-implant/include/QImplant.h
+++ b/compiler/q-implant/include/QImplant.h
@@ -33,6 +33,8 @@ public:
 
 private:
   const std::string &_path;
+
+  void forward_qparam(loco::Graph *g);
 };
 
 } // namespace q_implant
--- a/compiler/q-implant/src/QImplant.cpp
+++ b/compiler/q-implant/src/QImplant.cpp
@@ -32,6 +32,10 @@ using namespace q_implant;
 
 namespace
 {
+// TODO: Find opcodes which can be appended in conversion
+std::set<luci::CircleOpcode> appendable_operator_opcode{
+  luci::CircleOpcode::CONCATENATION, luci::CircleOpcode::RESHAPE, luci::CircleOpcode::SPLIT,
+  luci::CircleOpcode::TRANSPOSE};
 
 // Return directory path of given file path
 // TODO Find a platform-independent way to do this
@@ -103,6 +107,8 @@ Json::Value load_json(const std::string &path)
 
 void set_dtype(luci::CircleNode *node, loco::DataType dtype) { node->dtype(dtype); }
 
+void set_dtype(luci::CircleNode *origin, luci::CircleNode *dest) { dest->dtype(origin->dtype()); }
+
 void set_scale(luci::CircleNode *node, const std::string &scale_path)
 {
   assert(node);               // FIX CALLER UNLESS
@@ -141,6 +147,20 @@ void set_quantized_dimension(luci::CircleNode *node, const uint32_t quantized_di
   assert(node->quantparam()); // FIX CALLER UNLESS
 
   node->quantparam()->quantized_dimension = quantized_dimension;
+}
+
+void set_quantparam(luci::CircleNode *origin, luci::CircleNode *dest)
+{
+  assert(origin);
+  auto origin_quantparam = origin->quantparam();
+  assert(origin_quantparam);
+
+  assert(dest);
+  auto dest_quantparam = dest->quantparam();
+  assert(dest_quantparam);
+  dest_quantparam->zerop = origin_quantparam->zerop;
+  dest_quantparam->scale = origin_quantparam->scale;
+  dest_quantparam->quantized_dimension = origin_quantparam->quantized_dimension;
 }
 
 template <loco::DataType DT> void set_value(luci::CircleConst *node, const std::string &value_path)
@@ -248,6 +268,50 @@ void QImplant::write(loco::Graph *g)
       const auto value_path = dir_path + '/' + tensor["value"].asString();
 
       set_value(const_node, value_path, dtype);
+    }
+  }
+
+  std::set<luci::CircleNode *> visit;
+  std::deque<luci::CircleNode *> que;
+  for (auto input_node : loco::input_nodes(g))
+  {
+    auto node = loco::must_cast<luci::CircleNode *>(input_node);
+    que.emplace_back(node);
+  }
+
+  // check whether node can share its successor node quantization parameter
+  while (!que.empty())
+  {
+    auto node = que.front();
+    que.pop_front();
+
+    // skip when node is output
+    if (node->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
+    {
+      continue;
+    }
+
+    auto quantparam = node->quantparam();
+    THROW_UNLESS(quantparam);
+
+    for (auto child : loco::succs(node))
+    {
+      auto child_node = loco::must_cast<luci::CircleNode *>(child);
+      if (visit.find(child_node) != visit.end())
+        continue;
+      visit.emplace(child_node);
+      que.emplace_back(child_node);
+
+      if (child_node->quantparam() == nullptr)
+      {
+        if (appendable_operator_opcode.find(child_node->opcode()) ==
+            appendable_operator_opcode.end())
+          continue;
+
+        set_dtype(node, child_node);
+        child_node->quantparam(std::make_unique<luci::CircleQuantParam>());
+        set_quantparam(node, child_node);
+      }
     }
   }
 
--- a/compiler/q-implant/src/QImplant.cpp
+++ b/compiler/q-implant/src/QImplant.cpp
@@ -32,11 +32,6 @@ using namespace q_implant;
 
 namespace
 {
-// TODO: Find opcodes which can be appended in conversion
-std::set<luci::CircleOpcode> appendable_operator_opcode{
-  luci::CircleOpcode::CONCATENATION, luci::CircleOpcode::RESHAPE, luci::CircleOpcode::SPLIT,
-  luci::CircleOpcode::TRANSPOSE};
-
 // Return directory path of given file path
 // TODO Find a platform-independent way to do this
 std::string directory_path(const std::string &file_path)
@@ -107,7 +102,7 @@ Json::Value load_json(const std::string &path)
 
 void set_dtype(luci::CircleNode *node, loco::DataType dtype) { node->dtype(dtype); }
 
-void set_dtype(luci::CircleNode *origin, luci::CircleNode *dest) { dest->dtype(origin->dtype()); }
+void copy_dtype(const luci::CircleNode *src, luci::CircleNode *dest) { dest->dtype(src->dtype()); }
 
 void set_scale(luci::CircleNode *node, const std::string &scale_path)
 {
@@ -147,20 +142,6 @@ void set_quantized_dimension(luci::CircleNode *node, const uint32_t quantized_di
   assert(node->quantparam()); // FIX CALLER UNLESS
 
   node->quantparam()->quantized_dimension = quantized_dimension;
-}
-
-void set_quantparam(luci::CircleNode *origin, luci::CircleNode *dest)
-{
-  assert(origin);
-  auto origin_quantparam = origin->quantparam();
-  assert(origin_quantparam);
-
-  assert(dest);
-  auto dest_quantparam = dest->quantparam();
-  assert(dest_quantparam);
-  dest_quantparam->zerop = origin_quantparam->zerop;
-  dest_quantparam->scale = origin_quantparam->scale;
-  dest_quantparam->quantized_dimension = origin_quantparam->quantized_dimension;
 }
 
 template <loco::DataType DT> void set_value(luci::CircleConst *node, const std::string &value_path)
@@ -271,49 +252,7 @@ void QImplant::write(loco::Graph *g)
     }
   }
 
-  std::set<luci::CircleNode *> visit;
-  std::deque<luci::CircleNode *> que;
-  for (auto input_node : loco::input_nodes(g))
-  {
-    auto node = loco::must_cast<luci::CircleNode *>(input_node);
-    que.emplace_back(node);
-  }
-
-  // check whether node can share its successor node quantization parameter
-  while (!que.empty())
-  {
-    auto node = que.front();
-    que.pop_front();
-
-    // skip when node is output
-    if (node->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
-    {
-      continue;
-    }
-
-    auto quantparam = node->quantparam();
-    THROW_UNLESS(quantparam);
-
-    for (auto child : loco::succs(node))
-    {
-      auto child_node = loco::must_cast<luci::CircleNode *>(child);
-      if (visit.find(child_node) != visit.end())
-        continue;
-      visit.emplace(child_node);
-      que.emplace_back(child_node);
-
-      if (child_node->quantparam() == nullptr)
-      {
-        if (appendable_operator_opcode.find(child_node->opcode()) ==
-            appendable_operator_opcode.end())
-          continue;
-
-        set_dtype(node, child_node);
-        child_node->quantparam(std::make_unique<luci::CircleQuantParam>());
-        set_quantparam(node, child_node);
-      }
-    }
-  }
+  forward_qparam(g);
 
   // Update output nodes
   auto graph_outputs = g->outputs();
@@ -346,4 +285,38 @@ void QImplant::write(loco::Graph *g)
   }
 }
 
+void QImplant::forward_qparam(loco::Graph *g)
+{
+  std::set<luci::CircleOpcode> forwardable_opcode{
+    luci::CircleOpcode::RESHAPE, luci::CircleOpcode::SPLIT, luci::CircleOpcode::TRANSPOSE};
+
+  auto forwardable = [&forwardable_opcode](luci::CircleOpcode opcode) {
+    return forwardable_opcode.find(opcode) != forwardable_opcode.end();
+  };
+
+  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
+  {
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+    // skip when node is output
+    if (circle_node->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
+      continue;
+
+    auto quantparam = circle_node->quantparam();
+    if (quantparam == nullptr)
+      continue;
+
+    for (auto successor : loco::succs(node))
+    {
+      auto successor_node = loco::must_cast<luci::CircleNode *>(successor);
+      if (successor_node->quantparam() == nullptr)
+      {
+        if (!forwardable(successor_node->opcode()))
+          continue;
+        assert(circle_node->quantparam());
+        copy_quantparam(circle_node, successor_node);
+        copy_dtype(circle_node, successor_node);
+      }
+    }
+  }
+}
 #undef THROW_UNLESS
--- a/compiler/q-implant/src/QImplant.cpp
+++ b/compiler/q-implant/src/QImplant.cpp
@@ -32,6 +32,7 @@ using namespace q_implant;
 
 namespace
 {
+
 // Return directory path of given file path
 // TODO Find a platform-independent way to do this
 std::string directory_path(const std::string &file_path)
@@ -287,19 +288,26 @@ void QImplant::write(loco::Graph *g)
 
 void QImplant::forward_qparam(loco::Graph *g)
 {
-  std::set<luci::CircleOpcode> forwardable_opcode{
-    luci::CircleOpcode::RESHAPE, luci::CircleOpcode::SPLIT, luci::CircleOpcode::TRANSPOSE};
+  /*
+   * TODO: add comment about how to add to the set
+   *
+   * If the operator doesn't change input tensor value,
+   * (Operator don't change input quantization parameter to output tensor)
+   * the operator's quantization parameter can be forwarded
+   */
+  std::set<luci::CircleOpcode> forwardable_opcode;
+  forwardable_opcode.emplace(luci::CircleOpcode::RESHAPE);
+  forwardable_opcode.emplace(luci::CircleOpcode::SPLIT);
+  forwardable_opcode.emplace(luci::CircleOpcode::TRANSPOSE);
+  // TODO add more Ops
 
   auto forwardable = [&forwardable_opcode](luci::CircleOpcode opcode) {
     return forwardable_opcode.find(opcode) != forwardable_opcode.end();
   };
 
-  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
   {
     auto circle_node = loco::must_cast<luci::CircleNode *>(node);
-    // skip when node is output
-    if (circle_node->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
-      continue;
 
     auto quantparam = circle_node->quantparam();
     if (quantparam == nullptr)
@@ -312,11 +320,11 @@ void QImplant::forward_qparam(loco::Graph *g)
       {
         if (!forwardable(successor_node->opcode()))
           continue;
-        assert(circle_node->quantparam());
         copy_quantparam(circle_node, successor_node);
         copy_dtype(circle_node, successor_node);
       }
     }
   }
 }
+
 #undef THROW_UNLESS
--- a/compiler/q-implant/src/QImplant.cpp
+++ b/compiler/q-implant/src/QImplant.cpp
@@ -298,6 +298,7 @@ void QImplant::forward_qparam(loco::Graph *g)
   std::set<luci::CircleOpcode> forwardable_opcode;
   forwardable_opcode.emplace(luci::CircleOpcode::RESHAPE);
   forwardable_opcode.emplace(luci::CircleOpcode::SPLIT);
+  forwardable_opcode.emplace(luci::CircleOpcode::CIRCLESPLITOUT);
   forwardable_opcode.emplace(luci::CircleOpcode::TRANSPOSE);
   // TODO add more Ops
 


ssafyWooJeong:
for: #11254 #11594

jinevening:
It seems that you want to propagate qparams in forward direction (from input to output). You can take a look at `PropagateQParamForwardPass`. It would be better to use the pass to avoid duplicate codes.

compiler/q-implant/src/QImplant.cpp
```diff
@@ -32,6 +32,10 @@ using namespace q_implant;
 
 namespace
 {
+// TODO: Find opcodes which can be appended in conversion
+std::set<luci::CircleOpcode> appendable_operator_opcode{
+  luci::CircleOpcode::CONCATENATION, luci::CircleOpcode::RESHAPE, luci::CircleOpcode::SPLIT,
```
jinevening, line 37: For concatenation, qparam has to be propagated backward.

PTAL `propagate_concat_quantparam` function in `PropagateQParamBackwardPass.cpp`.

compiler/q-implant/src/QImplant.cpp
```diff
@@ -32,6 +32,10 @@ using namespace q_implant;
 
 namespace
 {
+// TODO: Find opcodes which can be appended in conversion
+std::set<luci::CircleOpcode> appendable_operator_opcode{
```
seanshpark, line 36: plz put this NONE STATIC function or struct/class member.

seanshpark:
plz revise title in shorter one so that title doesn't split like `which are ad…`

ssafyWooJeong:
> plz revise title in shorter one so that title doesn't split like `which are ad…`

changed title into 'Quantization Parameter forward propagation' as @jinevening mentioned.
>you want to propagate qparams in forward direction (from input to output).

>plz put this in NONE STATIC function or struct/class member.
Will change it into none static function when not using `PropagateQParamForwardPass` which @jinevening  suggested.

Thank you.

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong:
```
luci::PropagateQParamForwardPass forward_pass;
forward_pass.run(g);
```

If I understand `PropagateQParamForwardPass` correctly, running `PropagateQParamForwardPass` will propagate quantization parameter when operations like Reshape, Transpose.

```
bool copy_qparam(luci::CircleQuantParam *src, luci::CircleQuantParam *dst)
{
  assert(src->scale.size() == dst->scale.size());
  assert(src->zerop.size() == dst->zerop.size());
```

`copy_qparam` assume that source and destination node have same scale,zerop size, so we need to assign source's size into destination, right?

I'm not sure about it's correct or not `PropagateQParamForwardPass` maybe generally used in `quantized model` , so don't care about dtype. (Maybe the IR Graph is assumed as quantized correctly)

So, when I run the code by using `PropagateQParamForwardPass`, exception occured due to remaining float. (Quantization Parameter copied, but dtype not changed)

I not sure about changing dtype with using `PropagateQParamForwardPass` is possible or not.

<details>
<summary>Code</summary>
  std::set<luci::CircleNode *> visit;
  std::deque<luci::CircleNode *> que;
  for (auto input_node : loco::input_nodes(g))
  {
    auto node = loco::must_cast<luci::CircleNode *>(input_node);
    que.emplace_back(node);
  }

  while (!que.empty())
  {
    auto node = que.front();
    que.pop_front();

    if (node->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
    {
      continue;
    }

    auto node_qparam = node->quantparam();
    assert(node_qparam);

    for (auto child : loco::succs(node))
    {
      auto child_node = loco::must_cast<luci::CircleNode *>(child);
      if (visit.find(child_node) != visit.end())
        continue;
      visit.emplace(child_node);
      que.emplace_back(child_node);

      if (child_node->quantparam() == nullptr)
      {
        child_node->quantparam(std::make_unique<luci::CircleQuantParam>());
        auto child_qparam = child_node->quantparam();
        child_qparam->scale.assign(node_qparam->scale.size(), 0);
        child_qparam->zerop.assign(node_qparam->zerop.size(), 0);
      }

    }
  }

  luci::PropagateQParamForwardPass forward_pass;
  forward_pass.run(g);
</details>

jinevening: I see. So, we may need to do some preprocessing or modify `PropagateQParamForwardPass` to use it here. Thanks.

jinevening: Please remove concat from `appendable_operator_opcode`.

compiler/q-implant/src/QImplant.cpp
```diff
@@ -103,6 +107,8 @@ Json::Value load_json(const std::string &path)
 
 void set_dtype(luci::CircleNode *node, loco::DataType dtype) { node->dtype(dtype); }
 
+void set_dtype(luci::CircleNode *origin, luci::CircleNode *dest) { dest->dtype(origin->dtype()); }
```
jinevening, line 110:
```suggestion
void copy_dtype(const luci::CircleNode *src, luci::CircleNode *dest) { dest->dtype(src->dtype()); }
```

compiler/q-implant/src/QImplant.cpp
```diff
@@ -143,6 +149,20 @@ void set_quantized_dimension(luci::CircleNode *node, const uint32_t quantized_di
   node->quantparam()->quantized_dimension = quantized_dimension;
 }
 
+void set_quantparam(luci::CircleNode *origin, luci::CircleNode *dest)
```
jinevening, line 152:
```suggestion
void copy_quantparam(const luci::CircleNode *src, luci::CircleNode *dest)
```

compiler/q-implant/src/QImplant.cpp
```diff
@@ -251,6 +271,50 @@ void QImplant::write(loco::Graph *g)
...
+      que.emplace_back(child_node);
+
+      if (child_node->quantparam() == nullptr)
+      {
+        if (appendable_operator_opcode.find(child_node->opcode()) ==
+            appendable_operator_opcode.end())
+          continue;
+
+        set_dtype(node, child_node);
+        child_node->quantparam(std::make_unique<luci::CircleQuantParam>());
```
jinevening, line 312: Better to move this inside `copy_quantparam`.

compiler/q-implant/src/QImplant.cpp
```diff
@@ -251,6 +271,50 @@ void QImplant::write(loco::Graph *g)
...
+  while (!que.empty())
+  {
+    auto node = que.front();
+    que.pop_front();
+
+    // skip when node is output
+    if (node->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
+    {
+      continue;
+    }
```
jinevening, line 292: For consistency.

```suggestion
      continue;
```

compiler/q-implant/src/QImplant.cpp
```diff
@@ -251,6 +271,50 @@ void QImplant::write(loco::Graph *g)
...
+
+    auto quantparam = node->quantparam();
+    THROW_UNLESS(quantparam);
```
jinevening, line 295: `node` can be a non-quantizable type (ex: int32 or bool). `continue` would be enough.

Exception has to be used carefully, because users will instantly experience the failure. So, we typically throw an exception when the input of program (circle model and param.json in this case) is not valid, or users misuse the tool.

compiler/q-implant/src/QImplant.cpp
```diff
@@ -251,6 +271,50 @@ void QImplant::write(loco::Graph *g)
     }
   }
 
+  std::set<luci::CircleNode *> visit;
+  std::deque<luci::CircleNode *> que;
```
seanshpark, line 275: please rename `visit`, `que` to some meaningful word.

compiler/q-implant/src/QImplant.cpp
```diff
@@ -251,6 +271,50 @@ void QImplant::write(loco::Graph *g)
...
+    // skip when node is output
+    if (node->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
+    {
+      continue;
+    }
+
+    auto quantparam = node->quantparam();
+    THROW_UNLESS(quantparam);
+
+    for (auto child : loco::succs(node))
```
seanshpark, line 297: plz rename `child` to something else. graph/node has no child concept.

jinevening:
Could you extract the code as a separate function or class? `write` function is becoming too long.

It seems that you used queue just to traverse the graph. We already have helper functions for that purpose, ex: `loco::postorder_traversal` or `Pass`. Could you use them?

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: Replaced it to newly created class member `forward_qparam`

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: changed as suggested!

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: Found `copy_quantparam` on  [`CircleQuantParam.cpp`](https://github.com/Samsung/ONE/blob/6c77165d79bedb5fe4e4a3c09a1f4dafe99f58bd/compiler/luci/lang/src/CircleQuantParam.cpp#L28)

It's purpose is same, and to avoid duplicated code mentioned before,
removed it from `q-implant` and use `copy_quantparam` from `CircleQuantParam`

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: For using `copy_quantparam` from `CircleQuantParam`, just put `copy_dtype` as before.

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: I intend to make `exception` early when `exception` occurs later.
I never thought about non-quantizable type before, and just continuing would works fine!

So just `continue` would be better!

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: As suggestion about using `loco::postorder_traversal` by @jinevening,
replaced it to `loco::postorder_traversal`

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: I never knew about that, and renamed to `successor` as is.

Thank you.

compiler/q-implant/include/QImplant.h
```diff
@@ -33,6 +33,8 @@ class QImplant final
 
 private:
   const std::string &_path;
+
+  void forward_qparam(loco::Graph *g);
```
seanshpark, line 37: plz put this like
```cxx
class QImplant final
{
public:
  QImplant(const std::string &path) : _path(path) {}
  void write(loco::Graph *g);

private:
  void forward_qparam(loco::Graph *g);

private:
  const std::string &_path;
```

compiler/q-implant/src/QImplant.cpp
```diff
@@ -32,7 +32,6 @@ using namespace q_implant;
 
 namespace
 {
-
```
seanshpark, line 35: plz do not remove this line

compiler/q-implant/src/QImplant.cpp
```diff
@@ -282,4 +285,38 @@ void QImplant::write(loco::Graph *g)
...
+      {
+        if (!forwardable(successor_node->opcode()))
+          continue;
+        assert(circle_node->quantparam());
+        copy_quantparam(circle_node, successor_node);
+        copy_dtype(circle_node, successor_node);
+      }
+    }
+  }
+}
```
seanshpark, line 321: plz add empty line after `}`

compiler/q-implant/src/QImplant.cpp
```diff
@@ -282,4 +285,38 @@ void QImplant::write(loco::Graph *g)
   }
 }
 
+void QImplant::forward_qparam(loco::Graph *g)
+{
+  std::set<luci::CircleOpcode> forwardable_opcode{
+    luci::CircleOpcode::RESHAPE, luci::CircleOpcode::SPLIT, luci::CircleOpcode::TRANSPOSE};
```
seanshpark, line 291: plz add comment why only three Ops are listed here

seanshpark: (optional) use `forwardable_opcode.emplace()` to add these three for future extendability

compiler/q-implant/src/QImplant.cpp
```diff
@@ -282,4 +285,38 @@ void QImplant::write(loco::Graph *g)
...
+
+  auto forwardable = [&forwardable_opcode](luci::CircleOpcode opcode) {
+    return forwardable_opcode.find(opcode) != forwardable_opcode.end();
+  };
+
+  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
```
seanshpark, line 297: use `loco::active_nodes()`

compiler/q-implant/src/QImplant.cpp
```diff
@@ -282,4 +285,38 @@ void QImplant::write(loco::Graph *g)
...
+      continue;
+
+    for (auto successor : loco::succs(node))
+    {
+      auto successor_node = loco::must_cast<luci::CircleNode *>(successor);
+      if (successor_node->quantparam() == nullptr)
+      {
+        if (!forwardable(successor_node->opcode()))
+          continue;
+        assert(circle_node->quantparam());
```
seanshpark, line 315: plz hoist this line

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: During model conversion(onnx->tf) we faced only these operators until now.
Honestly not sure about which operators can be added on model conversion period and can use predecessor quantization parameter.

When found/faced others, we will add here.

At first I put elements vertically, and ran `nnas format'. I surprised that `nnas` make it horizontally.
So using `emplace` would be best option, because we can read vertically and can extend list easily.

Lastly, `add comment` means comment in here or add comment on `cpp`?

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: I found that `postorder_traversal` can give duplicate Node.

If I understand well, using `loco::active_nodes()` just intend to remove duplicated node, right?

`active_nodes` uses `set` to remove duplication, and I intend to process `in-order` from input
`set` orders elements, but I not sure about it will keep node order. (I think `set` will order by address of node, not sequence)

To remove duplication and keep node order, can I use `loco::postorder_traversal` and remove duplication with `set`?

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: `copy_quantparam` in `CircleQuantParam.cpp` check source quantparam is null or not, so It would unnecessary as you mention.

I will remove it.

Thank you!

compiler/q-implant/src/QImplant.cpp
seanshpark: > Honestly not sure about which operators can be added on model conversion period

Then, plz add decision factors how you selected these Ops.

> When found/faced others, we will add here.

You mean by until SSAFY?

> Lastly, add comment means comment in here or add comment on cpp?

In the cpp source code.

compiler/q-implant/src/QImplant.cpp
seanshpark: > I think set will order by address of node, not sequence

circle files are ordered in execution sequence, which nodes are created in that sequence which will have memory address of that sequence.
please look inside `loco::active_nodes()` source code.

compiler/q-implant/src/QImplant.cpp
seanshpark: `circle_node` does not change in this `for (auto successor : loco::succs(node))` loop.
so we can hoist out of this `for` loop.
but then, there is
```cxx
    auto quantparam = circle_node->quantparam();
    if (quantparam == nullptr)
      continue;
```
which, we don't have to do assertion check.

I didn't check with `copy_quantparam`

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: >Then, plz add decision factors how you selected these Ops.

- Operators can be added during model conversion
- Operators are not quantiazable, so can pass input tensor's quantization parameter to output tensor.

It's our thoughts, would you think It can be the decision factor what you mentioned?

> You mean by until SSAFY?

Not sure about, but maybe..?
If I have time to look `onnx-tf`, I will find if any other operators can be added.

> In the cpp source code.

If the decision factor above is appropriate, I will place it to cpp with your other suggestions and commit it.

Thank you.

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: > nodes are created in that sequence which will have memory address of that sequence.

Never know about that node will be placed as execution sequence!
The address of node can guarantee it, so `loco::active_nodes()` would be great solution!

compiler/q-implant/src/QImplant.cpp
seanshpark: > It's our thoughts, would you think It can be the decision factor what you mentioned?

nope. how did you added `RESHAPE` ?

ssafyWooJeong: I'm not sure how to answer because knowledge is limited.

- Operators can be added during model conversion
As we remember, we saw reshape was added on `onnx` to `tf` conversion, but if you suspicious of it, we can't answer about that.
(Don't have much knowledges, saw it in low version of torch/onnx-tf so not know it was wrong conversion or not, can't remember which model conversion make it)

[Reshape](https://www.tensorflow.org/api_docs/python/tf/reshape)
Reshape have input tensor, and shape tensor(in int), so not quantizable.
As we know, reshape operator don't change quantization paramter. (input/output tensor's quantization parameter same)

Last, if `qparam.json` contain quantization parameter(original model in other frame have reshape operator) of reshape operator, it would not work (because it have quantization parameter).

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: If I understand correctly, you mean that we can do it like
```
if(!circle_node->quantparam())
  continue;
```

compiler/q-implant/src/QImplant.cpp
```diff
@@ -282,4 +285,38 @@ void QImplant::write(loco::Graph *g)
...
+  auto forwardable = [&forwardable_opcode](luci::CircleOpcode opcode) {
+    return forwardable_opcode.find(opcode) != forwardable_opcode.end();
+  };
+
+  for (auto node : loco::postorder_traversal(loco::output_nodes(g)))
+  {
+    auto circle_node = loco::must_cast<luci::CircleNode *>(node);
+    // skip when node is output
+    if (circle_node->opcode() == luci::CircleOpcode::CIRCLEOUTPUT)
+      continue;
```
jinevening, line 302: This code seems unnecessary.

jinevening: Output nodes have no successors, so they will be ignored in the below loop.

compiler/q-implant/src/QImplant.cpp
seanshpark: in https://github.com/Samsung/ONE/pull/11593#discussion_r1336867644

> If I understand PropagateQParamForwardPass correctly, running PropagateQParamForwardPass will propagate quantization parameter when operations like Reshape, Transpose.

what was your understanding?

compiler/q-implant/src/QImplant.cpp
seanshpark: > If I understand correctly, you mean that we can do it like

I don't know why the code is related with `assert(circle_node->quantparam());` removal.

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: When reading the code, `PropagateQParamForwardPass` will copy predecessor node's quantization parameter to node like reshape, transpose.

So I understood that if I run the `PropagateQParamForwardPass`, some operators like Reshape or Transpose can get quantization parameter from predecessor.

> Then, plz add decision factors how you selected these Ops.

Is asking why there are three operators in list

> how did you added RESHAPE ?

Is saying that we should exclude reshape because reshape operator would not be added at model conversion(onnx -> tf) period. (What we saw/remember could be wrong, and I'm not 100% sure whether reshape can be added or not on model conversion)

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: >plz hoist this line this line looks unnecessary

First, I only heard about `hoist` in `javascript`, and not sure about the meaning of sentence.
And when I checked review at `File changed`, the review is attached at `assert` line.
So I understood that assert line is unnecessary.

> circle_node does not change in this for (auto successor : loco::succs(node)) loop.
so we can hoist out of this for loop.

I understood that `quantparam` is only used at `if` condition, so declaring `quantparam` is unnecessary.

> If I understand correctly, you mean that we can do it like

So at here, I asked about (want to check)  `hoist` means that don't have to declare variable and just use it in `if` condition.
I didn't meant that removal assertion or not.

Sorry about lack of my knowledge and English..

Thank you

compiler/q-implant/src/QImplant.cpp
seanshpark: > When reading the code, PropagateQParamForwardPass will copy ...

Do you mean, the code wrote by @jinevening ?

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: >Output nodes have no successors, so they will be ignored in the below loop.

I missed that.
I will reflect it.

Thank you

compiler/q-implant/src/QImplant.cpp
seanshpark: > I asked about (want to check) hoist means

FYI https://en.wikipedia.org/wiki/Loop-invariant_code_motion

ssafyWooJeong: Now I got what you meant.

`assert(circle_node->quantparam());` is not related to successor_node, so need to check it not in side of successor loop but in side of node loop.

> but then, there is

But we have already checked quantparam, so there is no need to check it. so you strikethrough the `hoist` sentence.
I knew about loop-invariant conditions, but don't know about `hoist` means.
So when you say the line is unnecessary, I start to check near code and when checking `copy_quantparam`, it checks `qparam` so I just thought `unnecessary` is checking qparam even the function process well. 

Sorry about lack of conversation.

Thank you

compiler/q-implant/src/QImplant.cpp
ssafyWooJeong: Yes, when I reading `PropagateQParamForwardPass` on luci/pass, it looks like copying quantization parameter when some operator like reshape, transpose.
So `PropagateQParamForwardPass` was thought to be used in `q-implant`, but when looked closely,

> copy_qparam assume that source and destination node have same scale,zerop size, so we need to assign source's size into destination, right?

`copy_qparam` on `PropagateQParamForwardPass` assume that destination model already have same scale/zero point size. `q-implant` could not have quantization parameter (assume using it to not quantized model to make quantized model), so we have to assign proper size for not quantized model.

> when I run the code by using PropagateQParamForwardPass, exception occured due to remaining float. (Quantization Parameter copied, but dtype not changed)

When used it, dtype not copied, and need to find another solution.

> It seems that you used queue just to traverse the graph. We already have helper functions for that purpose, ex: loco::postorder_traversal or Pass. Could you use them?

@jinevening recommended to use postorder_traversal.

> what was your understanding?

That's what I understood, and not sure about relation between `PropagateQParamForwardPass` and `Reshape`

seanshpark: > ... relation between PropagateQParamForwardPass and Reshape

RESHAPE was the first item in the code, so I asked the reason, it was not about specific `Reshape`.
You should have answered like "I don't know cause the original code was written from @jinevening".

Anyway, my suggestion is like
```cxx
  // TODO: add comment about how to add to the set
  std::set<luci::CircleOpcode> forwardable_opcode;
  forwardable_opcode.emplace(luci::CircleOpcode::RESHAPE);
  forwardable_opcode.emplace(luci::CircleOpcode::SPLIT);
  forwardable_opcode.emplace(luci::CircleOpcode::TRANSPOSE);
  // TODO add more Ops
```

ssafyWooJeong: >nope. how did you added RESHAPE ?

I understood it as why did we added `reshape` itself.

Firstly, we didn't know about `luci::PropagateQParamForwardPass` until @jinevening recognized what we want to do is same/similar with  `PropagateQParamForwardPass` and asked to check possible to use to prevent duplicated code yesterday.

> Honestly not sure about which operators can be added on model conversion period

What I'm not sure about, is don't know all operators can be added during model conversion, so we placed what we saw on conversion for now. 

>>Then, plz add decision factors how you selected these Ops.

>  Operators can be added during model conversion
 Operators are not quantiazable, so can pass input tensor's quantization parameter to output tensor.

These are 'as is' how we selected, `decision factors` may mean which operators can be placed on list (common attribute that operators in list must satisfied).
- Operator does not change quantization parameter. (min, max, zero point, scale)
- reshape, transpose and split are just change tensor value's position/dimension, and not change individual value so quantization parameter does not change.
- So input tensor's quantization parameter and output tensor's quantization parameter must be same.


That's what all I know, not sure that answer ,is what you hoped, or not.


>Anyway, my suggestion is like

And I think it is good for people to read and expand, I will reflect it.

Thank you.

seanshpark: > and not change individual value so quantization parameter does not change.


@jinevening , does this hold for channel wise?

jinevening: > does this hold for channel wise?

Yes. Activation is quantized layer-wise, not channel-wise.

compiler/q-implant/src/QImplant.cpp
```diff
@@ -287,19 +288,26 @@ void QImplant::write(loco::Graph *g)
...
+  /*
+   * TODO: add comment about how to add to the set
+   *
+   * If the operator doesn't change input tensor value,
+   * (Operator don't change input quantization parameter to output tensor)
+   * the operator's quantization parameter can be forwarded
+   */
+  std::set<luci::CircleOpcode> forwardable_opcode;
+  forwardable_opcode.emplace(luci::CircleOpcode::RESHAPE);
+  forwardable_opcode.emplace(luci::CircleOpcode::SPLIT);
```
jinevening, line 300: Qparam has to be forwarded to CircleSplitOut, which is the output of CircleSplit.

```suggestion
  forwardable_opcode.emplace(luci::CircleOpcode::SPLIT);
  forwardable_opcode.emplace(luci::CircleOpcode::CIRCLESPLITOUT);
```

jinevening: LGTM after this change.

ssafyWooJeong: Reflected the suggestion!

compiler/q-implant/include/QImplant.h
```diff
@@ -28,9 +28,11 @@ class QImplant final
 {
 public:
   QImplant(const std::string &path) : _path(path) {}
-
```
seanshpark, line 31: please move this format change in another PR.

seanshpark: Q) why remove this line? personally this helps reading the code and looks better.

ssafyWooJeong: > please move this format change in another PR.

1) Does it means that add function signature in .h in another PR, and add implementation in this PR? (Maybe rebase this branch to merged one?)
2) Or, just `format change` means that only line removal?

> Q) why remove this line? personally this helps reading the code and looks better.

And I also thought that line separation looks better too.
I just removed the line by thinking about there is convention that I don't know.

> plz put this like
```
class QImplant final
{
public:
  QImplant(const std::string &path) : _path(path) {}
  void write(loco::Graph *g);

private:
  void forward_qparam(loco::Graph *g);

private:
  const std::string &_path;
```

I will keep the line after check what I have to do.
When case 1, I will separate `.h` first, and change this PR.
Or when case 2, I will just keep the line in this PR.

Thank You!

seanshpark: > Or, just format change means that only line removal?

this one.

ssafyWooJeong: Add the line as before.

Thank You!

compiler/q-implant/include/QImplant.h
```diff
@@ -28,9 +28,12 @@ class QImplant final
...
+
```
jinevening, line 31: Trailing whitespace.

```suggestion
  
```

seanshpark (approved):
LGTM

jinevening (approved):
LGTM

Issue #11596, opened 2023-09-26
[circle-quantizer] Refactor vector of LayerParam


seanshpark:
To extend passing multiple sets of `std::vector<LayerParam>`,
let's define a new TYPE for `std::vector<std::shared_ptr<LayerParam>>`

seanshpark:
for https://github.com/Samsung/ONE/issues/11280#issuecomment-1734826226

seanshpark:
Current usage 
```cxx
      auto l = std::make_shared<LayerParam>();
      {
        l->name = layer["name"].asString();
        l->dtype = layer["dtype"].asString();
        l->granularity = layer["granularity"].asString();
      }
      p.emplace_back(l);
```
```cxx
      for (auto layer_param : layer_params)
      {
        LayerInfo info;
        {
          info.name = layer_param->name;
          info.dtype = str_to_dtype(layer_param->dtype);
          info.granularity = str_to_granularity(layer_param->granularity);
        }
        ctx->layers_info.emplace_back(info);
      }
```

seanshpark:
Steps
1/ revise with `using NEWTYPE = std::vector<std::shared_ptr<LayerParam>>;`
2/ introduce `struct NEWTYPE` ?

seanshpark:
`LayerParamsSet` was added

PR #11598, opened 2023-09-26
[luci/pass] Introduce LayerParms

seanshpark:
This will simplify methods with LayerParams type alias.


seanshpark:
for #11596 , from #11513

seanshpark:
@jinevening PTAL

jinevening (approved):
LGTM

Issue #11542, opened 2023-09-18
Compiler FE: Support nn.Linear + nn.GELU + nn.Linear Torch model


jinevening:
## What

Let's support import/optimize/quantize of the `nn.Linear` -> `nn.GELU` -> `nn.Linear` pattern in torch.

## Why

Existing `fuse_gelu` option only works for limited patterns. `nn.Linear` -> `nn.GELU` -> `nn.Linear` pattern is not supported.

A major issue is that some Ops of GELU are fused with adjacent FC layers during import (onnx->tflite), so GELU cannot be fused. 

Onnx pattern
![image](https://github.com/Samsung/ONE/assets/5449554/28f2b9a1-cb91-4a50-848e-69ae6a77c24c)

Circle pattern
![image](https://github.com/Samsung/ONE/assets/5449554/7264016e-b1e0-4325-8a3d-be4f9fb729db)
(Note that **Div before Erf is fused with the preceding FC**, and **the last Mul is fused with the next FC**)

There seems two options.
1. Implement a new pass to substitute the above pattern into FC + GELU + FC
  - This may handle nn.Linear -> nn.GELU -> nn.Linear pattern, which is very specific.
2. Use circle-mlir
  - This can handle more general patterns, but need some time to become stable (?). CC @seanshpark

jinevening:
NOTE This task is not necessary if `experimental_disable_batchmatmul_unfold` is True.

seanshpark:
@jinevening , do you have any captured image of `expected circle` model graph?

> Use circle-mlir

I expect _circle-mlir_ to be released next year.

jinevening:
> do you have any captured image of expected circle model graph?

I don't have the expected graph, but there is one generated by previous one-compiler (1.25).

![image](https://github.com/Samsung/ONE/assets/5449554/7ee2d663-f1a2-4a44-bcf1-6a42879d0c43)

The above pattern is in fact the same with `FC (BatchMatMul + Add) -> Gelu -> FC (BatchMatMul)`, which I expect.

jinevening:
All done.

I've checked that our target models (deit, efficientformer) have FC-Gelu-FC patterns with O1 option.

PR #11604, opened 2023-09-27
DRAFT: Introduce extract_gelu_from_opt_fc option

jinevening:
On going draft to support extract_gelu_from_opt_fc.

---
Related to: #11542


jinevening:
I couldn't add a new test in `luci-pass-value-test`, because the original pattern contains `Erf`, which is not supported by TFLiteInterpreter .

So, I manually did value check as below. We have an onnx model which contains the target pattern.

1. Convert onnx to circle -> Slice the target Ops -> Apply `extract_gelu_from_opt_fc` option -> A
2. Convert onnx to tflite (with [SELECT_TF_OPS option](https://github.com/tensorflow/tensorflow/issues/60663#issuecomment-1581532987))  -> Slice the target Ops -> B
3. Compare execution results of A and B using `luci_eval_verifier.py` in `luci-value-test`.

NOTE B can be executed with TFLiteInterpreter, because `SELECT_TF_OPS` option is turned on in `tf2tfliteV2.py`.

Step 3 passed in my local machine. I want to share the test models, but they are not uploaded to github (perhaps due to github upload issue..).

jinevening:
Commit plan
- luci
- circle2circle
- exclude.lst
- res
- circle2circle-dredd-test
- one-cmds

jinevening:
For the bias pattern, I've done value test similarly with https://github.com/Samsung/ONE/pull/11604#issuecomment-1736718249.

jinevening:
All done

PR #11608, opened 2023-09-27
[luci] Introduce ReplaceWithFCGeluFCPass 

jinevening:
This introduces ReplaceWithFCGeluFCPass and its test.

---
Related to: https://github.com/Samsung/ONE/issues/11542
Draft PR: #11604

### DIFF
--- a/compiler/luci/pass/include/luci/Pass/ReplaceWithFCGeluFCPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ReplaceWithFCGeluFCPass.h
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_REPLACE_WITH_FC_GELU_FC_PASS_H__
+#define __LUCI_REPLACE_WITH_FC_GELU_FC_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/include/luci/Pass/ExtractGeluFromOptFCPass.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace luci
+{
+
+/**
+ * @brief  Class to extract Gelu from a certain Op pattern
+ *
+ * To see the target Op pattern, please visit its implementation.
+ */
+struct ExtractGeluFromOptFCPass final : public logo::Pass
```
seanshpark, line 30: Q) what is `Opt` in `..FromOptFC...` ?

jinevening: Opt means "Optimized". Help message will contain the full meaning.

From draft
```cpp
  add_switch(arser, "--extract_gelu_from_opt_fc",
             "Extract gelu operator from optimized fc pattern");
```

seanshpark: Thanks for the explanation :)
But then, it's a bit ambiguous about the meaning `optimized`...
Is it like mul/add constant fused?

As help message is in another source file, can you add some explanation here too ?

seanshpark: (optional) I think I had some misunderstanding about this pass...
From what this pass does and from the words of existing pass class...
how about `ReplaceWithFCGeluFCPass` or `TransformToFCGeluFCPass`

jinevening: > Is it like mul/add constant fused?

Yes. `div` and `mul` in gelu Op are fused with the preceding/succeeding FCs.

> As help message is in another source file, can you add some explanation here too ?

Sure. I'll update :)

> how about ReplaceWithFCGeluFCPass or TransformToFCGeluFCPass ?

Existing passes have the form of `ReplaceAwithB` or `TransformAToB`. The suggested names have no information about `A` (because of complex pattern..). Would it be ok?

seanshpark: > The suggested names have no information about A (because of complex pattern..). Would it be ok?

Yes, missing `A` was the concern... -_-;;;
I think it's OK as I don't have better idea.

seanshpark:
Ah, please revise commit message title too.

jinevening:
@nnfw-bot test nncc-release

PR #11611, opened 2023-09-27
[circle-quantizer] Extract QConfReader

seanshpark:
This will extract QConfReader for reading qconf json file.

### DIFF
--- a/compiler/circle-quantizer/src/CircleQuantizer.cpp
+++ b/compiler/circle-quantizer/src/CircleQuantizer.cpp
@@ -35,8 +35,37 @@ using LayerParams = luci::CircleQuantizer::Options::LayerParams;
 using Algorithms = luci::CircleQuantizer::Options::Algorithm;
 using AlgorithmParameters = luci::CircleQuantizer::Options::AlgorithmParameters;
 
+struct QConfReader
+{
+  void init(const std::string &filename)
+  {
+    std::ifstream ifs;
+
+    ifs.open(filename);
+
+    // Failed to open cfg file
+    if (not ifs.is_open())
+      throw std::runtime_error("Cannot open config file. " + filename);
+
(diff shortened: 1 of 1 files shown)


seanshpark:
for #11280 , from draft  #11513

seanshpark:
@jinevening ping

compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -38,8 +38,40 @@ using LayerParams = luci::CircleQuantizer::Options::LayerParams;
 using Algorithms = luci::CircleQuantizer::Options::Algorithm;
 using AlgorithmParameters = luci::CircleQuantizer::Options::AlgorithmParameters;
 
+struct QConfReader
+{
+  void init(std::string &filename)
```
jinevening, line 43:
```suggestion
  void init(const std::string &filename)
```

compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -38,8 +38,40 @@ using LayerParams = luci::CircleQuantizer::Options::LayerParams;
...
+    // Failed to open cfg file
+    if (not _ifs.is_open())
+      throw std::runtime_error("Cannot open config file. " + filename);
+
+    JSONCPP_STRING errs;
+    Json::CharReaderBuilder builder;
+
+    // Failed to parse
+    if (not parseFromStream(builder, _ifs, &_root, &errs))
+      throw std::runtime_error("Cannot parse config file (json format). " + errs);
```
jinevening, line 56: It would be better to close `_ifs` right after parsing, rather than in destructor.

jinevening (approved):
LGTM Thank you!

PR #11646, opened 2023-10-04
[circle-quantizer] Read alternate layer params

seanshpark:
This will enable to read alternate layer params.

### DIFF
--- a/compiler/circle-quantizer/src/CircleQuantizer.cpp
+++ b/compiler/circle-quantizer/src/CircleQuantizer.cpp
@@ -118,7 +118,7 @@ LayerParams read_layer_params(std::string &filename)
   return p;
 }
 
-LayerParamsSet read_layer_params_set(const std::string &filename)
+LayerParamsSet read_layer_params_set(std::string &filename)
 {
   LayerParamsSet lpss;
 
--- a/compiler/circle-quantizer/src/CircleQuantizer.cpp
+++ b/compiler/circle-quantizer/src/CircleQuantizer.cpp
@@ -32,6 +32,7 @@
 
 using LayerParam = luci::CircleQuantizer::Options::LayerParam;
 using LayerParams = luci::CircleQuantizer::Options::LayerParams;
+using LayerParamsSet = luci::CircleQuantizer::Options::LayerParamsSet;
 using Algorithms = luci::CircleQuantizer::Options::Algorithm;
 using AlgorithmParameters = luci::CircleQuantizer::Options::AlgorithmParameters;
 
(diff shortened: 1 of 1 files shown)


seanshpark:
for #11280 , from draft  #11513

seanshpark:
@nnfw-bot test nncc-release

compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -117,6 +118,46 @@ LayerParams read_layer_params(std::string &filename)
   return p;
 }
 
+LayerParamsSet read_layer_params_set(std::string &filename)
```
jinevening, line 121:
```suggestion
LayerParamsSet read_layer_params_set(const std::string &filename)
```

seanshpark: FYI need to add `const` also to existing `read_layer_params()`

compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -117,6 +118,46 @@ LayerParams read_layer_params(std::string &filename)
   return p;
 }
 
+LayerParamsSet read_layer_params_set(const std::string &filename)
```
jinevening, line 121: Hmm.. it is quite difficult for me to understand the behavior. I found how `alternate` is handled in https://github.com/Samsung/ONE/issues/11280#issuecomment-1734650521.

`read_layer_params_set` returns a LayerParamsSet which includes three LayerParams

LayerParams (default, based on "name" and "names") -> vector incldues "onnx_tf_prefix_Relu_1;Add;convolution;Const_1"
LayerParams (alternate, from "tf2121") -> vector includes "onnx_tf_prefix_/relu1/Relu;Add;convolution;Const_2"
LayerParams (alternate, from "o2c") -> vector includes "Conv_0;Relu_1"

From https://github.com/Samsung/ONE/pull/11647, among the three params, only 1 valid LayerParams should exist.

It seems quite different from the originally discussed interface in https://github.com/Samsung/ONE/issues/11280#issue-1856190615. Is it intended?

seanshpark: > Is it intended?

Yup. changed to avoid complicate situations and to make things simple :)

jinevening: Ok, then could you explain how the original example can be handled with the new interface?

Original interface
```
"names": [ "a", "b", "c" ],
if we can provide alternative names for the node, like, "a" can be "a1" or "a2",

"names": ["a", "b", "c" ],
"alternate": {
"a" : ["a1", "a2"],
"b" : ["b1", "b2"],
}
```

New interface
```
{
"names": ["a", "b", "c"]
"alternate" : {
  "v1" : ["a1", "b", "c"],
  "v2" : ["a2", "b", "c"],
  "v3" : ["a", "b1", "c"],
  ...
}
}
```
IIUC, there will be 3 (a, a1, a2) X 3 (b, b1, b2) X 1 = 9 configurations. Did I understand correctly?

compiler/circle-quantizer/src/CircleQuantizer.cpp
```diff
@@ -118,7 +118,7 @@ LayerParams read_layer_params(std::string &filename)
   return p;
 }
 
-LayerParamsSet read_layer_params_set(std::string &filename)
+LayerParamsSet read_layer_params_set(const std::string &filename)
```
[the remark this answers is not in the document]
jinevening: Ah, it needs further changes. Thanks.

compiler/circle-quantizer/src/CircleQuantizer.cpp
seanshpark: alternate was not designed(?) for combinations; sorry for not describing about this in the first place.

intention of original interface is to support more (version B and C) from current version A
- from version A: ["a", "b", "c" ],
- from version B: ["a1", "b1", "c" ],
- from version C: ["a2", "b2", "c" ],

which simple can be described in new version as
```json
{
"names": ["a", "b", "c"]
"alternate" : {
  "B" : ["a1", "b1", "c"],
  "C" : ["a2", "b2", "c"],
}
}
```

jinevening: Ah, I see. So, it is not intended to cover `["a1", "b", "c"]`. Thanks!

jinevening (approved):
LGTM

shs-park (commented):
LGTM!

PR #11480, opened 2023-09-05
[Draft] Add MaxPool2D for training

zetwhite:
This PR is to add MaxPool2D operation for training. 

ONE-DCO-1.0-Signed-off-by: SeungHui Youn <sseung.youn@samsung.com>

------------------------------------------------------
* You can check the **result** of this draft here - https://github.com/Samsung/ONE/issues/11427#issuecomment-1747817428

Issue #11427, opened 2023-09-01
[onert] Introduce MaxPool op for training


glistening:
`onert` started to support training. It only covers a limited of operators at this moment.
Someone may start to support other ops by referring existing other training op implementation.

Related: #11389

zetwhite:
I was looking into it. 
Is it okay to assign myself?

zetwhite:
I think someone might wonder about the progress of this issue. 
So, Here is a note to share the progress.

-----------------------------------------------

I made a draft here with some uniitests here https://github.com/Samsung/ONE/pull/11480.
But I didn't check the integrated one(full pass test) which starts with `*.circle` file, yet. 

But as you might know, We have another work in office for two weeks. :cry: 
So, This work will be suspended for a while. I'll try to finish this one as soon as possible.

zetwhite:
* (progress updated) 

https://github.com/Samsung/ONE/pull/11480(Draft) is ready.
It works well with under model and the training data.


[pool_mnist_model.zip](https://github.com/Samsung/ONE/files/12809198/pool_mnist_model.zip)

zetwhite:
(progress updated) 

https://github.com/Samsung/ONE/pull/11480(Draft) is ready. It works well with under model and the training data.
So, I will make PRs from the draft from now on. 

* model and data : [pool_mnist_model.zip](https://github.com/Samsung/ONE/files/12809198/pool_mnist_model.zip)

```bash
➜ ~/Workspace/ONE/Product/out/bin/onert_train --modelfile pool_mnist_model/pool_model.circle --load_input:raw pool_mnist_model/mnist_x_train.1000.bin --load_expected:raw pool_mnist_model/mnist_y_train.1000.bin --data_length 1000 -v 100 --learning_rate 0.01 --epoch 10  
Model Expected Filename pool_mnist_model/mnist_y_train.1000.bin
Model Input Filename pool_mnist_model/mnist_x_train.1000.bin
Model Filename pool_mnist_model/pool_model.circle
Epoch 1/10 - 55.257ms/step - loss: [0] 13.8349
Epoch 2/10 - 55.186ms/step - loss: [0] 1.9618
Epoch 3/10 - 55.166ms/step - loss: [0] 0.8555
Epoch 4/10 - 55.243ms/step - loss: [0] 0.5257
Epoch 5/10 - 55.164ms/step - loss: [0] 0.3861
Epoch 6/10 - 55.200ms/step - loss: [0] 0.3138
[...]

Issue #11248, opened 2023-08-11
[circle+] Formula to calculate gradient


jyoungyun:
## Purpose

This issue is where we record what formula we used when calculating the gradient of backward propagation. Based on the formula here, we will implement a gradient, and for details of the implementation, please refer to the source code.

jyoungyun:
# FullyConnected

## formula to calculate gradient
Input : $X$
Weights $W$
Output : $Z$
Bias : $B$

### FullyConnected Forward
##### Formula
$Z = XW + B$

##### Fullyconnected Formula
fullyconnected's `weights` is already transposed. So $W$ here is transposed of the above $W$.
$Z = fc(X, W)$

### FullyConnected Backward
#### Input
##### Formula
$\frac{\partial L}{\partial X}$ = $\frac{\partial L}{\partial Z} W^T $

##### Fullyconnected Formula
$\frac{\partial L}{\partial X}$ =  fc($\frac{\partial L}{\partial Z}, W^T) $

#### Weights
##### Formula
$\frac{\partial L}{\partial W} = (\frac{\partial L}{\partial Z})^T X$

##### Fullyconnected Formula
$\frac{\partial L}{\partial W} = fc((\frac{\partial L}{\partial Z})^T, X^T)$

#### Bias
##### Formula
$$
\begin{flalign}
\frac{\partial L}{\partial B} = 
\begin{bmatrix}
    \frac{\partial L}{\partial z_{11}} + \frac{\partial L}{\partial z_{21}} \\
    \frac{\partial L}{\partial z_{12}} + \frac{\partial L}{\partial z_{22}} \\
\end{bmatrix} 
\end{flalign}
$$

_This code is copied from #11204._

jyoungyun:
## MSE (Mean Square Error)

### Forward

### $\text{Error}(y, y_{t}) = \cfrac{1}{N} \sum_{\text{i=1 to N}} (y_{i} - y_{it})^2$

### Backward

[...]

PR #11686, opened 2023-10-06
[onert] Enable max2d PoolLayer for training

zetwhite:
This PR is to enable max2d PoolLayer for training.

This PR fills unimplemented part of PoolLayer: 

  * implements shape-inference visit() for pool2d
  * implelments kernel-generator visit() for pool2d
  * Implements train PoolLayer op's forward, backward functions and ckernel
  
---------------------------------------------------

related issue : https://github.com/Samsung/ONE/issues/11427 
from draft : https://github.com/Samsung/ONE/pull/11480

### DIFF
--- a/compute/cker/include/cker/train/operation/MaxPool.h
+++ b/compute/cker/include/cker/train/operation/MaxPool.h
@@ -1,0 +1,162 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __NNFW_CKER_TRAIN_OPERATION_MAXPOOL_H__
+#define __NNFW_CKER_TRAIN_OPERATION_MAXPOOL_H__
+
+#include "cker/Shape.h"
+#include "cker/Utils.h"
+#include "cker/eigen/Utils.h"
+
+#include <Eigen/Core>
+
+namespace nnfw
+{
+namespace cker
+{
+namespace train
+{
+
+// Most of the logic except 'arg_max_index' related is copy-paste from
+// https://github.com/Samsung/ONE/blob/a380292/compute/cker/include/cker/operation/MaxPool.h#L42-L88
+// 'arg_max_index' is to record max-arguments' index to apply gradient later.
+inline void MaxPool2D(const PoolParams &params, const Shape &input_shape, const float *input_data,
+                      const Shape &output_shape, float *output_data, int *arg_max_index)
+{
+  assert(input_shape.DimensionsCount() == 4);
+  assert(output_shape.DimensionsCount() == 4);
+  assert(input_shape.Dims(0) == output_shape.Dims(0)); // MaxPool2D doesn't change batch
+  assert(input_shape.Dims(3) == output_shape.Dims(3)); // MaxPool2D doesn't change depth
+
+  const int batches = MatchingDim(input_shape, 0, output_shape, 0);
+  const int input_height = input_shape.Dims(1);
+  const int input_width = input_shape.Dims(2);
+  const int output_height = output_shape.Dims(1);
+  const int output_width = output_shape.Dims(2);
+  const int filter_height = params.filter_height;
+  const int filter_width = params.filter_width;
+  const int pad_height = params.padding_values.height;
+  const int pad_width = params.padding_values.width;
+  const int stride_height = params.stride_height;
+  const int stride_width = params.stride_width;
+
+  const auto in_mat = MapAsMatrixWithLastDimAsRows(input_data, input_shape);
+  auto out_mat = MapAsMatrixWithLastDimAsRows(output_data, output_shape);
+  auto arg_max_index_mat = MapAsMatrixWithLastDimAsRows(arg_max_index, output_shape);
+
+  // initialize output area
+  std::fill(output_data, output_data + output_shape.FlatSize(), 0.0);
+  std::fill(arg_max_index, arg_max_index + output_shape.FlatSize(), -1);
+
+  // initialize projected area with lowest float
+  const int h_start =
+    (pad_height < filter_height) ? 0 : (pad_height - filter_height) / stride_height + 1;
+  const int h_end = std::min((input_height + pad_height - 1) / stride_height + 1, output_height);
+
+  const int w_start =
+    (pad_width < filter_width) ? 0 : (pad_width - filter_width) / stride_width + 1;
+  const int w_end = std::min((input_width + pad_width - 1) / stride_width + 1, output_width);
+
+  for (int b = 0; b < batches; ++b)
+  {
+    for (int h_idx = h_start; h_idx < h_end; h_idx++)
+    {
+      for (int w_idx = w_start; w_idx < w_end; w_idx++)
+      {
+        const int offset = NodeOffset(b, h_idx, w_idx, output_height, output_width);
+        out_mat.col(offset).setConstant(std::numeric_limits<float>::lowest());
+      }
+    }
+  }
+
+  for (int b = 0; b < batches; ++b)
+  {
+    for (int h = 0; h < input_height; ++h)
+    {
+      for (int w = 0; w < input_width; ++w)
+      {
+        // (h_start, h_end) * (w_start, w_end) is the range that the input
+        // vector projects to.
+        int hpad = h + pad_height;
+        int wpad = w + pad_width;
+
+        int h_start = (hpad < filter_height) ? 0 : (hpad - filter_height) / stride_height + 1;
+        int h_end = std::min(hpad / stride_height + 1, output_height);
+
+        int w_start = (wpad < filter_width) ? 0 : (wpad - filter_width) / stride_width + 1;
+        int w_end = std::min(wpad / stride_width + 1, output_width);
+
+        // compute elementwise sum
+        for (int ph = h_start; ph < h_end; ++ph)
+        {
+          for (int pw = w_start; pw < w_end; ++pw)
+          {
+            const int out_offset = NodeOffset(b, ph, pw, output_height, output_width);
+            const int in_offset = NodeOffset(b, h, w, input_height, input_width);
+
+            const auto out_vector = out_mat.col(out_offset);
+            const auto in_vector = in_mat.col(in_offset);
+
+            // update arg_max_index_mat
+            arg_max_index_mat.col(out_offset) =
+              (out_vector.array() < in_vector.array())
+                .select(in_offset, arg_max_index_mat.col(out_offset));
+
+            // update out_mat
+            out_mat.col(out_offset) = out_vector.cwiseMax(in_vector);
+          }
+        }
+      }
+    }
+  }
+
+  out_mat.cwiseMin(params.float_activation_min).cwiseMax(params.float_activation_max);
+}
+
+inline void MaxPool2DGrad(const Shape &deriv_output_shape, const float *deriv_output_data,
+                          const int *arg_max_index, const Shape &deriv_input_shape,
+                          float *deriv_input_data)
+{
+  assert(deriv_input_shape.DimensionsCount() == 4);
+  assert(deriv_output_shape.DimensionsCount() == 4);
+
+  // initialize deriv_input_data
+  std::fill(deriv_input_data, deriv_input_data + deriv_input_shape.FlatSize(), 0.0);
+
+  const int depth = MatchingDim(deriv_input_shape, 3, deriv_output_shape, 3);
+  const auto deriv_out_mat = MapAsMatrixWithLastDimAsRows(deriv_output_data, deriv_output_shape);
+  auto arg_max_index_mat = MapAsMatrixWithLastDimAsRows(arg_max_index, deriv_output_shape);
+  auto deriv_in_mat = MapAsMatrixWithLastDimAsRows(deriv_input_data, deriv_input_shape);
+
+  for (int col_index = 0; col_index < deriv_out_mat.cols(); col_index++)
+  {
+    auto arg_indices = arg_max_index_mat.col(col_index);
+    for (int d = 0; d < depth; d++)
+    {
+      // output value is from padding, so nothing to propagate
+      if (arg_indices(d) == -1)
+        continue;
+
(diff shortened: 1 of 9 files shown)


runtime/onert/backend/cpu/ops/PoolLayer.h
```diff
@@ -53,10 +53,11 @@ class PoolLayer : public ::onert::exec::IFunction
 
   void run() override;
 
-private:
+protected:
   const IPortableTensor *_input;
   IPortableTensor *_output;
```
zetwhite, line 58: These need to be accessible by `train/op/PoolLayer`.

runtime/onert/backend/train/KernelGenerator.h
```diff
@@ -50,9 +50,11 @@ class KernelGenerator : public backend::train::KernelGeneratorBase
   void visit(const ir::train::operation::ElementwiseActivation &) override;
   void visit(const ir::train::operation::FullyConnected &) override;
   void visit(const ir::train::operation::Loss &) override;
+  void visit(const ir::train::operation::Pool2D &) override;
   void visit(const ir::train::operation::Reshape &node) override;
 
 private:
+  const ir::Operands &_ctx;
```
zetwhite, line 57: This is necessary to get `pool layer`'s input and output tensor shape.
In/Out shape is used to calculate the pool layer's padding.

runtime/onert/backend/train/ops/PoolLayer.cc
```diff
@@ -25,7 +31,104 @@ namespace train
...
+  switch (pool_type)
+  {
+    case train::ops::PoolType::kMax:
+      return cpu::ops::PoolType::kMax;
+    default:
+      throw std::runtime_error("PoolLayer: Unsupported pool type yet");
+  }
+}
+
+class MaxPool2D final : public TrainingKernelRegistry
```
zetwhite, line 48: This class is a wrapper of  "max pool" related things. 

It provides
	* forward and backward interfaces for maxpooling (call proper kernel) 
	* holds additional memory used to record the max-arguments' index.

runtime/onert/backend/train/ops/PoolLayer.cc
```diff
@@ -35,14 +147,32 @@ void PoolLayer::configure(const IPortableTensor *input, const uint32_t paddingLe
...
+  cpu::ops::PoolLayer::configure(input, paddingLeft, paddingRight, paddingTop, paddingBottom,
+                                 strideWidth, strideHeight, kernelWidth, kernelHeight, activation,
+                                 output, convert_to_infer(op_type));
+
+  if (output->data_type() != OperandType::FLOAT32)
+  {
+    throw std::runtime_error("PoolLayer : Unsupported data type for training");
+  }
+
+  // ready training kernel
```
zetwhite, line 169: Both Inference and training kernel is ready in compile time.
In Runtime, depends on `bool training` value, It decides which kernel to use.

compute/cker/include/cker/train/operation/MaxPool.h
```diff
@@ -0,0 +1,131 @@
...
+
+// Most of the logic except 'arg_max_index' related is copy-paste from
+// https://github.com/Samsung/ONE/blob/a380292/compute/cker/include/cker/operation/MaxPool.h#L42-L88
+// 'arg_max_index' is to record max-arguments' index to apply gradient later.
+inline void MaxPool2D(const PoolParams &params, const Shape &input_shape, const float *input_data,
+                      const Shape &output_shape, float *output_data, int *arg_max_index)
+{
+  assert(input_shape.DimensionsCount() == 4);
+  assert(output_shape.DimensionsCount() == 4);
+  assert(input_shape.Dims(3) == output_shape.Dims(3)); // MaxPool2D doesn't change depth
```
ragmani, line 41: (Optional) How about adding to check batches?
```suggestion
  assert(input_shape.Dims(0) == output_shape.Dims(0)); // MaxPool2D doesn't change batch
  assert(input_shape.Dims(3) == output_shape.Dims(3)); // MaxPool2D doesn't change depth
```

runtime/onert/backend/train/KernelGenerator.h
ragmani: (Optional) How about using the base member `_tgraph` instead of adding this member (because performance is not important where this member is used) ?

runtime/onert/backend/train/ops/PoolLayer.cc
```diff
@@ -25,7 +31,113 @@ namespace train
 namespace ops
 {
 
-PoolLayer::PoolLayer() : cpu::ops::PoolLayer()
+namespace
+{
+
+cpu::ops::PoolType convert_to_infer(const train::ops::PoolType &pool_type)
```
ragmani, line 37: (Optional)
```suggestion
cpu::ops::PoolType convertToInfer(const train::ops::PoolType &pool_type)
```
Although there are many places in onert that did not follow naming conventions, it would be better to use camel case that starts with lowercase.

compute/cker/include/cker/train/operation/MaxPool.h
jyoungyun: The first dimension is already being checked in the `MatchingDim` function below. So this overlaps. :)

runtime/onert/backend/train/ops/PoolLayer.cc
```diff
@@ -25,7 +31,113 @@ namespace train
...
+    // maxpool forward
+    nnfw::cker::train::MaxPool2D(_op_params, getShape(in), getBuffer<float>(in), out_shape,
+                                 out_data, getBuffer<int>(arg_max_index));
+
+    // activation forward
+    for (int i = 0; i < out_shape.FlatSize(); ++i)
+    {
+      out_data[i] = nnfw::cker::ActivationFunctionWithMinMax(
+        out_data[i], _op_params.float_activation_min, _op_params.float_activation_max);
+    }
```
jyoungyun, line 106: (Q) Why do you implement this activation function here? The original MaxPool.h kernel code implements this function inside the cker. Could you explain your intention?

compute/cker/include/cker/train/operation/MaxPool.h
```diff
@@ -0,0 +1,131 @@
...
+  const int input_width = input_shape.Dims(2);
+  const int output_height = output_shape.Dims(1);
+  const int output_width = output_shape.Dims(2);
+  const int stride_height = params.stride_height;
+  const int stride_width = params.stride_width;
+
+  const auto in_mat = MapAsMatrixWithLastDimAsRows(input_data, input_shape);
+  auto out_mat = MapAsMatrixWithLastDimAsRows(output_data, output_shape);
+  auto arg_max_index_mat = MapAsMatrixWithLastDimAsRows(arg_max_index, output_shape);
+
```
jyoungyun, line 54: Could you explain the grad formula for MaxPool function [here](https://github.com/Samsung/ONE/issues/11248)? It will help me understand your code. :)

runtime/onert/backend/train/ops/PoolLayer.cc
zetwhite: I hope that `forward` and `backward` functions are matched somehow. 
Since the `backward` explicitly calls gradient-calculation function for `maxpool` and `relu`, I hope that `forward` has a similar look.  

I also thought that if `avgpool` is introduced later, this activation-related part could be exported to a common-thing (to share between `avgpool` and `maxpool`).
To export this common part, I thought it might better to have logic explicit here rather than inside the cker.

zetwhite: But I don't have a strong opinion about this part. 
So, If you think it might be better to move inside the `cker`, please let me know. :smile:

compute/cker/include/cker/train/operation/MaxPool.h
```diff
@@ -0,0 +1,131 @@
...
+            // update out_mat
+            out_mat.col(out_offset) = out_vector.cwiseMax(in_vector);
+          }
+        }
+      }
+    }
+  }
+}
+
+inline void MaxPool2DGrad(const Shape &deriv_output_shape, const float *deriv_output_data,
```
zetwhite, line 102: Additionally also here, `MaxPool2DGrad` only do maxpool thing, not activation related. 
So, To make corresponded `MaxPool2D`, I thought it looks more natural that `MaxPool2D` only does max pooling, not activation.

runtime/onert/backend/train/ops/PoolLayer.cc
zetwhite: [Additionally, Also here ](https://github.com/Samsung/ONE/pull/11686/files/4c2d6bbd23ce01ef3c3212bc213045fb5d7a5ab8#r1351950822)

compute/cker/include/cker/train/operation/MaxPool.h
ragmani: I think it's not bad because it doesn't affect performance.

compute/cker/include/cker/train/operation/MaxPool.h
glistening: @zetwhite We handles several types (e.g. max, avg, ...) of pooling in `PoolLayer`. It handles activation (including no activation) using `*_activation_min/max` in parameter. It is the way how `tflite` handles activation. We used similar (or same) kernel if there is no reason to do differently. Thus, handling differently only in `MaxPool`'s `forward` may make others confused. I prefer to keep the same style.

runtime/onert/backend/train/ops/PoolLayer.cc
glistening: I like the similar structure between `forward` and `backward` in same operator. But for me, keeping `forward` consistent to existing inference code as much as possible is more important. Please find more https://github.com/Samsung/ONE/pull/11686/files#r1353573206.

compute/cker/include/cker/train/operation/MaxPool.h
zetwhite: @glistening Oh, I fully understood your intention. I'll move activation part to cker.

runtime/onert/backend/train/ops/PoolLayer.cc
ragmani: You can use `cwiseMin()` and `cwieMax()` of `out_mat` in MaxPool.h of cker like below.
https://github.com/Samsung/ONE/blob/31fc7a8267e38cd0879e85b8e6f4cfd942a010f7/compute/cker/include/cker/eigen/eigen_gemm_eigen.h#L84-L87

compute/cker/include/cker/train/operation/MaxPool.h
zetwhite: updated!

compute/cker/include/cker/train/operation/MaxPool.h
zetwhite: updated!

runtime/onert/backend/train/KernelGenerator.h
zetwhite: Oh, I see. updated!

runtime/onert/backend/train/ops/PoolLayer.cc
zetwhite: updated

compute/cker/src/train/MaxPool.test.cc
```diff
@@ -0,0 +1,188 @@
...
+
+  int b;
+  int h;
+  int w;
+  int d;
+};
+
+} // namespace
+
+TEST(CKer_Operation, MaxPool2D_Depth1)
```
zetwhite, line 42: Test Results

```bash
➜ Product/out/unittest/test_cker --gtest_filter='*MaxPool*'  
Running main() from /home/sseung/Workspace/ONE/externals/GTEST/googletest/src/gtest_main.cc
Note: Google Test filter = *MaxPool*
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from CKer_Operation
[ RUN      ] CKer_Operation.MaxPool2D_Depth1
[       OK ] CKer_Operation.MaxPool2D_Depth1 (0 ms)
[ RUN      ] CKer_Operation.MaxPool2D_Depth2
[       OK ] CKer_Operation.MaxPool2D_Depth2 (0 ms)
[----------] 2 tests from CKer_Operation (0 ms total)

[==========] 2 tests from 1 test suite ran. (0 ms total)
[  PASSED  ] 2 tests.
```

compute/cker/include/cker/train/operation/MaxPool.h
```diff
@@ -0,0 +1,134 @@
...
+            const int out_offset = NodeOffset(b, ph, pw, output_height, output_width);
+            const int in_offset = NodeOffset(b, h, w, input_height, input_width);
+
+            const auto out_vector = out_mat.col(out_offset);
+            const auto in_vector = in_mat.col(in_offset);
+
+            // update arg_max_index_mat
+            arg_max_index_mat.col(out_offset) =
+              (out_vector.array() < in_vector.array())
+                .select(in_offset, arg_max_index_mat.col(out_offset));
```
zetwhite, line 92: (notes) 

eigen `select` syntax from (  https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#a65e78cfcbc9852e6923bebff4323ddca )

```
(condition).select(if-selected, else-selected);
```
```cpp
MatrixXi m(3, 3);
m << 1, 2, 3,
     4, 5, 6,
     7, 8, 9;
m = (m.array() >= 5).select(-m, m);
cout << m << endl;

// output : 
// 1 2 3
// 4 -5 -6
// -7 -8 -9
```

ragmani:
Could you add negative tests? We should have more negative tests than positive tests.

compute/cker/include/cker/train/operation/MaxPool.h
zetwhite: I added here - https://github.com/Samsung/ONE/issues/11248#issuecomment-1756798349

zetwhite:
> Could you add negative tests? 

Oh I see. I'll add more negative (also padding) tests.

zetwhite:
> Oh I see. I'll add more negative (also padding) tests.

(Progress updates) While working on this, i found some suspicious part with a padding example. 
That's why it took some time.  :cry:

compute/cker/src/train/MaxPool.test.cc
```diff
@@ -0,0 +1,279 @@
...
+    MaxPoolOpVerifier<float> verifier(op_param, in, out);
+
+    std::vector<float> input = {0, 0, 0, 0};
+    std::vector<float> expected_output = {0, 0, 0, 0};
+    verifier.verifyForward(input, expected_output);
+
+    std::vector<float> output_deriv = {0.1, 0.1, 0.1, 0.2};
+    std::vector<float> expected_input_deriv = {0.1, 0.1, 0.1, 0.1};
+    verifier.verifyBackward(output_deriv, expected_input_deriv, false);
+  }
```
zetwhite, line 278: (lately updated) Here, I updated tc with padding and negative one. (https://github.com/Samsung/ONE/pull/11686#issuecomment-1757001754)

ragmani (approved):
LGTM

glistening (approved):
LGTM

jyoungyun (approved):
LGTM

PR #11688, opened 2023-10-10
[infra] Revise modules list to remove old mio

seanshpark:
This will revise modules list to remove old mio-tflite280, mio-circle05.


seanshpark:
for #11390

jinevening (approved):
LGTM

PR #11560, opened 2023-09-19
[circle-mpqsolver] Introduce PatternSolver

stamalakhov:
This draft introduces preliminary version of PatternSolver.
for the model from https://github.com/Samsung/ONE/issues/11543#issuecomment-1722941372 it produced the following pattern:
image

on `circle-mpqsolver` level its usage is as follows:
```
./build/compiler/circle-mpqsolver/circle-mpqsolver 
 --input_model .../LAYER_NORM/layernorm.min_max.circle
 --output_model  .../LAYER_NORM/quantized.mpq.circle",
 --patterns --LayerNorm
```
TODO: m.b. tests?
Related: #11543
ONE-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

Issue #11543, opened 2023-09-18
[circle-mpqsolver] Introduce pattern-based solver


jinevening:
## What

Let's introduce pattern-based solver.

## Why

We can use pre-defined patterns to improve accuracy of some operators (ex: Softmax, LayerNorm).

stamalakhov:
@jinevening 
May be i could try to resolve this one?

jinevening:
Of course :)

stamalakhov:
Do you mean to attach some (e.g. `.json`) file with restrictions on node `quantization` type depending on nodes type (like Softmax..) and/or depending on their name (`frozen` nodes)? or am i missing something?

jinevening:
The interface is not fixed. We may need to discuss that too. I'm thinking that this feature should not require users to write node names. It will automatically find target nodes based on some patterns.

One possible scenario is to introduce a new solver, e.g., PatternSolver.

```ini
[one-quantize]
input_path=model.circle
quantized_dtype=uint8
granualrity=channel
ampq=True
ampq_algorithm=pattern
mixed_layernorm=True
mixed_softmax=True
output_path=model.q.circle
```

---
For now, we have two target patterns (softmax and layernorm). But softmax pattern is not fixed, so it will be better to start with layernorm pattern.

Target pattern
- Mul-Rsqrt chain (Mul, Mean, Add, Rsqrt) -> int16
- The rest (Mean, Sub, Mul, Add) -> uint8

![image](https://github.com/Samsung/ONE/assets/5449554/0e9e90c7-f593-4108-a1da-4b4786b1e249)

jinevening:
Here is fp32 circle model for the layernorm pattern.

[layernorm.zip](https://github.com/Samsung/ONE/files/12645758/layernorm.zip)

stamalakhov:
> For now, we have two target patterns (softmax and layernorm). But softmax pattern is not fixed, so it will be better to start with layernorm pattern.
> 
> Target pattern
> 
>     * Mul-Rsqrt chain (Mul, Mean, Add, Rsqrt) -> int16
> 
>     * The rest (Mean, Sub, Mul, Add) -> uint8

Ahhh. Understood. Ok. Thank you.

stamalakhov:
@jinevening 
[...]

PR #11693, opened 2023-10-10
[circle-mpqsolver] Introduce PatternSolver

stamalakhov:
This commit introduces PatternSolver
to get optimal mixed precision model.

Its correctness is tested at https://github.com/Samsung/ONE/pull/11560.

Draft: https://github.com/Samsung/ONE/pull/11560
Related: https://github.com/Samsung/ONE/issues/11543

### DIFF
--- a/compiler/circle-mpqsolver/src/pattern/PatternSolver.cpp
+++ b/compiler/circle-mpqsolver/src/pattern/PatternSolver.cpp
@@ -16,9 +16,14 @@
 
 #include "PatternSolver.h"
 
+#include "PatternResolver.h"
+
 #include <iostream>
 
 using namespace mpqsolver::pattern;
+
+using LayerParam = luci::CircleQuantizer::Options::LayerParam;
+using LayerParams = luci::CircleQuantizer::Options::LayerParams;
 
 PatternSolver::PatternSolver(const std::string &input_quantization,
                              const std::string &output_quantization,
@@ -44,4 +49,51 @@ std::unique_ptr<luci::Module> PatternSolver::run(const std::string &module_path)
   }
 
   return module;
+}
+
+void PatternSolver::set_mpq_options(MPQOptions &options) { _options = options; }
+
+LayerParams PatternSolver::get_frozen_params() const
+{
+  LayerParams params;
+  for (auto node_to_param : _frozen._node_to_param)
+  {
+    params.push_back(std::make_shared<LayerParam>(node_to_param.second));
+  }
+
+  return params;
+}
+
+void PatternSolver::resolve_patterns(luci::Module *module)
+{
+  _frozen._node_to_param.clear();
+
+  for (auto pattern : _options._patterns)
+  {
+    std::unique_ptr<pattern::PatternResolver> resolver;
+    switch (pattern)
+    {
+      case QuantizationPattern::Q8LayerNormWithQ16Variance:
+        resolver = std::make_unique<pattern::Q8LayerNormWithQ16VarianceResolver>();
+        break;
+      default:
+        throw std::runtime_error("Unsupported pattern to resolve");
+    }
+
+    auto const resolved = resolver->resolve(module);
+    for (auto node_param : resolved)
+    {
+      auto const frozen = _frozen._node_to_param.find(node_param.first);
+      if (frozen == _frozen._node_to_param.end())
+      {
+        // node was not previously defined - just set it (no ambiguity)
+        _frozen._node_to_param[node_param.first] = node_param.second;
+      }
+      else if (frozen->second.dtype != node_param.second.dtype)
+      {
+        // ambiguity (incoming description conflicts with current)
+        throw std::runtime_error("Resolved patterns contradict each other");
+      }
+    }
+  }
 }
--- a/compiler/circle-mpqsolver/src/pattern/PatternSolver.cpp
+++ b/compiler/circle-mpqsolver/src/pattern/PatternSolver.cpp
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


jinevening (approved):
LGTM

Hmm.. It found that some methods in `MPQSolver` have to be moved to `PatternSolver`.

```cpp
  /**
   * @brief set quantization options
   */
  void set_mpq_options(MPQOptions &options);

  /**
   * @brief fill _frozen with prescribed quantization parameters of resolved nodes
   */
  void resolve_patterns(luci::Module *module);

  /**
   * @brief resolve Q8LayerNormWithQ16Variance pattern
   */
  void resolve_layer_norm_pattern(luci::Module *module);

  /**
   * @brief transform _frozen nodes to Quantizer friendly form
   */
  luci::CircleQuantizer::Options::LayerParams get_frozen_params() const;
```

Those methods are only used in `PatternSolver`.

compiler/circle-mpqsolver/src/pattern/PatternSolver.cpp
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  resolve_patterns(module.get());
+
+  auto layer_params = get_frozen_params();
+
+  if (!_quantizer->quantize(module.get(), "uint8", layer_params))
+  {
+    std::cerr << "ERROR: Failed to quantize model" << std::endl;
+    return nullptr;
```
seanshpark, line 47: Q) above throws runtime_error. this line returns nullptr.
what is the decision point for this?

compiler/circle-mpqsolver/src/pattern/PatternSolver.cpp
```diff
@@ -0,0 +1,51 @@
...
+  MPQOptions options{patterns};
+  set_mpq_options(options);
+}
+
+std::unique_ptr<luci::Module> PatternSolver::run(const std::string &module_path)
+{
+  auto module = read_module(module_path);
+  if (!module)
+  {
+    throw std::runtime_error("Failed to load module");
```
seanshpark, line 37: `module` or `model` ?

stamalakhov: Ahh. Sorry . `Model` sounds more user friendly.

stamalakhov:
> Hmm.. I found that some methods in `MPQSolver` have to be moved to `PatternSolver`.

@jinevening 
But they will be useful for `Bisection` solver or any other solvers, which will use frozen parameters.
Should i move them into 'PatternSolver?

compiler/circle-mpqsolver/src/pattern/PatternSolver.cpp
stamalakhov: I thought that:
throw will notify user that input data is incorrect, while quantize may fail due to incorrect output parameters.
I'll remove `throw` as `Bisection` doesn't throw in that case.

jinevening:
> But they will be useful for Bisection solver or any other solvers, which will use frozen parameters.
Should i move them into 'PatternSolver?

Currently Bisection solver does not use those functions. Do you have a plan to support combination of pattern solver and bisection solver? If so, it would be better to make a new class e.g., `PatternResolver`, which includes those methods, and let `BisectionSolver` and `PatternSolver` inherit `PatternResolver`.

stamalakhov:
> > But they will be useful for Bisection solver or any other solvers, which will use frozen parameters.
> > Should i move them into 'PatternSolver?
> 
> Currently Bisection solver does not use those functions. Do you have a plan to support combination of pattern solver and bisection solver? If so, it would be better to make a new class e.g., `PatternResolver`, which includes those methods, and let `BisectionSolver` and `PatternSolver` inherit `PatternResolver`.
> 
> This can be done in a separate PR.

ok.

stamalakhov:
@jinevening 
done.

compiler/circle-mpqsolver/src/pattern/PatternSolver.cpp
stamalakhov: Ahhh. ok.

stamalakhov: @seanshpark
done.

stamalakhov: For `bisection` it will be included into separate PR.

seanshpark: > throw will notify user that input data is incorrect, while quantize may fail due to incorrect output parameters.
I'll remove throw as Bisection doesn't throw in that case.

`std::cerr << "ERROR: Failed to quantize model" << std::endl;` also notifies to the user with same format. I'm still not certain that two may exit in different ways.

stamalakhov: ok. I'll change it to throw also.

stamalakhov: done

compiler/circle-mpqsolver/src/pattern/PatternSolver.cpp
```diff
@@ -0,0 +1,102 @@
...
+  MPQOptions options{patterns};
+  set_mpq_options(options);
+}
+
+std::unique_ptr<luci::Module> PatternSolver::run(const std::string &module_path)
+{
+  auto module = read_module(module_path);
+  if (!module)
+  {
+    throw std::runtime_error("Failed to load model");
```
seanshpark, line 42: in `MPQSolver::read_module()`,
```cxx
  if (module.get() == nullptr)
  {
    std::cerr << "ERROR: Failed to load " << path << std::endl;
    return nullptr;
  }
```
which seems to show two `Failed to load` message.
IMHO, it maybe difficult to solve this issue in this PR so please fire a new issue for this problem, and go as-is in this PR.

stamalakhov: Ahh. Ok.

jinevening (approved):
LGTM Thank you

PR #11700, opened 2023-10-11
[luci] Fake quantize Exp

jinevening:
This supports fake quantization of Exp op.


seanshpark (approved):
LGTM

PR #11721, opened 2023-10-13
[one-cmds] Support HardSwish decomposing option

shs-park:
This commit adds the missing description of the previously implemented `decompose_hardswish` option.

ONE-DCO-1.0-Signed-off-by: Seungho Henry Park <shs.park@samsung.com>

---

Related to: #10930

### DIFF
--- a/compiler/one-cmds/how-to-use-one-commands.txt
+++ b/compiler/one-cmds/how-to-use-one-commands.txt
@@ -149,6 +149,7 @@ one-optimize
 one-optimize provides network or operator transformation shown below.
 
 Current transformation options are
+- decompose_hardswish : This will decompose hardswish operator to Add, Mul and Relu6 operators.
 - decompose_softmax : This will decompose softmax into multiple operators for special backends.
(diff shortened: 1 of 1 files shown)


compiler/one-cmds/how-to-use-one-commands.txt
```diff
@@ -149,6 +149,7 @@ one-optimize
 one-optimize provides network or operator transformation shown below.
 
 Current transformation options are
+- decompose_hardswish : This will decompose hardswish into multiple operators for special backends.
```
seanshpark, line 152: in `circle2circle`, description is `Decompose HardSwish operator to Add, Mul and Relu6 operators`. Do you want to go with current one and fix also `circle2circle` or revise this to follow `circle2circle` ?

shs-park: I think it would be better to edit this description to the one on `circle2circle`.
I will update the commit. 
Thanks for pointing it out. 😺

seanshpark (approved):
LGTM

Issue #11717, opened 2023-10-13
[one-quantize] Save recorded min/max


jinevening:
## What

Let's make an option to save min/max recorded during calibration

## Why

Min/max values are useful for optimization in some backends.

jinevening:
@BalyshevArtem Thanks for taking this task!

BalyshevArtem:
I think issue can be closed, as #11783 and #11793 are merged ? :)

jinevening:
Yes. Thank you for contribution!

PR #11730, opened 2023-10-17
[DRAFT][one-quantize] Save recorded min/max

BalyshevArtem:
This commit saves recorded min/max

for issue: #11717

### DIFF
--- a/compiler/luci/pass/include/luci/CircleQuantizer.h
+++ b/compiler/luci/pass/include/luci/CircleQuantizer.h
@@ -85,6 +85,7 @@ public:
       Quantize_input_type,
       Quantize_output_type,
       Quantize_TF_style_maxpool,
+      Quantize_save_min_max,
     };
 
     virtual ~Options() = default;
--- a/compiler/luci/pass/src/QuantizeWeights.cpp
+++ b/compiler/luci/pass/src/QuantizeWeights.cpp
@@ -407,8 +407,11 @@ void QuantizeWeights::quantize_weights(luci::CircleConst *weights)
     {
       sym_wquant_per_channel(weights, scaling_factor, channel_dim_index);
     }
-    quantparam->min.clear();
-    quantparam->max.clear();
+    if (not save_min_max)
+    {
+      quantparam->min.clear();
+      quantparam->max.clear();
+    }
     quantparam->quantized_dimension = channel_dim_index;
   }
   // Find min/max per layer-wise
@@ -449,8 +452,11 @@ void QuantizeWeights::quantize_weights(luci::CircleConst *weights)
     auto min = quantparam->min[0];
     auto scaling_factor = quantparam->scale[0];
     asym_wquant_per_layer(weights, min, scaling_factor);
-    quantparam->min.clear();
-    quantparam->max.clear();
+    if (not save_min_max)
+    {
+      quantparam->min.clear();
+      quantparam->max.clear();
+    }
   }
 }
 void QuantizeWeights::visit(luci::CircleConv2D *node)
--- a/compiler/luci/pass/src/QuantizeWeights.cpp
+++ b/compiler/luci/pass/src/QuantizeWeights.cpp
@@ -407,11 +407,6 @@ void QuantizeWeights::quantize_weights(luci::CircleConst *weights)
     {
       sym_wquant_per_channel(weights, scaling_factor, channel_dim_index);
     }
-    if (not save_min_max)
-    {
-      quantparam->min.clear();
-      quantparam->max.clear();
-    }
     quantparam->quantized_dimension = channel_dim_index;
   }
   // Find min/max per layer-wise
@@ -452,11 +447,6 @@ void QuantizeWeights::quantize_weights(luci::CircleConst *weights)
     auto min = quantparam->min[0];
     auto scaling_factor = quantparam->scale[0];
     asym_wquant_per_layer(weights, min, scaling_factor);
-    if (not save_min_max)
-    {
-      quantparam->min.clear();
-      quantparam->max.clear();
-    }
   }
 }
 void QuantizeWeights::visit(luci::CircleConv2D *node)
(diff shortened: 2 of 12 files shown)


compiler/luci/pass/include/luci/CircleQuantizer.h
```diff
@@ -85,6 +85,7 @@ class CircleQuantizer final
       Quantize_input_type,
       Quantize_output_type,
       Quantize_TF_style_maxpool,
+      Quantize_save_min_max,
```
seanshpark, line 88: plz split luci / circle-quantize to separate PRs

seanshpark (changes requested):
plz split luci , circle-quantize to separate PRs

seanshpark:
plz add some tests for these

compiler/luci/pass/src/QuantizeWeights.cpp
```diff
@@ -407,8 +407,11 @@ void QuantizeWeights::quantize_weights(luci::CircleConst *weights)
     {
       sym_wquant_per_channel(weights, scaling_factor, channel_dim_index);
     }
-    quantparam->min.clear();
-    quantparam->max.clear();
+    if (not save_min_max)
+    {
+      quantparam->min.clear();
+      quantparam->max.clear();
+    }
```
jinevening, line 414: I guess that this code can be removed? Because they will be removed in QuantizeWithMinMax if save_min_max is false. 
```suggestion
```

compiler/luci/pass/src/QuantizeWeights.cpp
```diff
@@ -449,8 +452,11 @@ void QuantizeWeights::quantize_weights(luci::CircleConst *weights)
     auto min = quantparam->min[0];
     auto scaling_factor = quantparam->scale[0];
     asym_wquant_per_layer(weights, min, scaling_factor);
-    quantparam->min.clear();
-    quantparam->max.clear();
+    if (not save_min_max)
+    {
+      quantparam->min.clear();
+      quantparam->max.clear();
+    }
```
jinevening, line 459: Ditto.

```suggestion
```

compiler/luci/pass/src/QuantizeWeights.h
```diff
@@ -29,14 +29,16 @@ namespace luci
  */
 struct QuantizeWeights final : public luci::CircleNodeMutableVisitor<void>
 {
-  QuantizeWeights(loco::DataType input, loco::DataType output, QuantizationGranularity gr)
-    : input_type(input), output_type(output), granularity(gr)
+  QuantizeWeights(loco::DataType input, loco::DataType output, QuantizationGranularity gr,
+                  bool is_save_min_max)
+    : input_type(input), output_type(output), granularity(gr), save_min_max(is_save_min_max)
```
jinevening, line 34: The change seems unnecessary. PTAL https://github.com/Samsung/ONE/pull/11730/files#r1362939612

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
```diff
@@ -577,13 +577,29 @@ bool QuantizeWithMinMaxPass::run(loco::Graph *g)
...
+    // Save min max values
+    if (_ctx->save_min_max)
+    {
+      const auto quant_param = circle_node->quantparam();
+      if (quant_param != nullptr)
+      {
+        node_to_min_values[circle_node] = quant_param->min;
+        node_to_max_values[circle_node] = quant_param->max;
+      }
+    }
```
jinevening, line 601: Q) Is there a reason to save min/max values here and restore later? Isn't it possible to maintain min/max values and remove them only when `_ctx->save_min_max = true`?

jinevening:
> plz add some tests for these

About testing, is it possible to check existence of min/max using `circledump` (ex: like `one-quantize_016.test`)?

compiler/luci/pass/src/QuantizeWeights.cpp
BalyshevArtem: I think yes, it can be good idea, thank you!

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
BalyshevArtem: As I understand there are special passes (https://github.com/Samsung/ONE/blob/11478d7ccffbc3bd55eedfe1a4d1498ba992c808/compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp#L629-L642) for some operations, that can remove forget about min,max values. 
I thought there might be more such places where values are implicitly forgotten, so I decided to keep it that way for sure, but if it would be better to change the methods themselves that create new quantum parameters (for example `make_predefined_qparam` and `quant_const`), then I will do so.
 Probably the second option is to change these methods so that they always save and then at the end, depending on the value of the flag, either delete or save these values.

BalyshevArtem:
> About testing, is it possible to check existence of min/max using `circledump` (ex: like `one-quantize_016.test`)?

Actually, I'm not sure I know how to do it, could you tell me how to do it? :)
I was thinking of just adding a new test to the QuantizeWithMinMaxPass.test file, where after applying this pass, check that each node has the min max values. How about this way, is it ok?

compiler/luci/pass/src/QuantizeWeights.cpp
BalyshevArtem: Done

jinevening:
> How about this way, is it ok?

Yes. That looks ok.

compiler/luci/pass/src/QuantizeWithMinMaxPass.cpp
jinevening: The previous way saves a map (node pointer -> min/max values) at the beginning. But the saved pointers can be invalidated, because nodes can be removed and newly created during quantization. 

I see that you already made a change. Thanks!

jinevening:
Please fix build error

```
[2023-10-18T11:11:05.683Z] Segmentation fault (core dumped)
```

jinevening:
@BalyshevArtem Looks good to me in general. Could you upload PRs?

BalyshevArtem:
everything merged - close it

PR #11734, opened 2023-10-18
[CFE/minmax-embedder] Introduce driver

glistening:
It introduces minmax-embedder driver.

### DIFF
--- a/compiler/minmax-embedder/CMakeLists.txt
+++ b/compiler/minmax-embedder/CMakeLists.txt
@@ -6,7 +6,9 @@ if(NOT HDF5_FOUND)
 endif(NOT HDF5_FOUND)
 
 file(GLOB_RECURSE SOURCES "src/*.cpp")
+file(GLOB_RECURSE DRIVER "driver/*.cpp")
 file(GLOB_RECURSE TESTS "src/*.test.cpp")
+list(REMOVE_ITEM SOURCES ${DRIVER})
 list(REMOVE_ITEM SOURCES ${TESTS})
 
 #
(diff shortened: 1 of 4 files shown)


compiler/minmax-embedder/driver/Driver.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  EmbedderOptions opt{min_percentile, max_percentile};
+  try
+  {
+    Embedder().embed(output_path, circle_path, minmax_path, opt);
+  }
+  catch (const std::runtime_error &err)
+  {
+    std::cout << err.what() << std::endl;
+    std::cout << arser;
```
jinevening, line 74: Is it intended to print `arser`?

compiler/minmax-embedder/CMakeLists.txt
```diff
@@ -39,3 +41,16 @@ nnas_find_package(GTest REQUIRED)
...
 target_link_libraries(minmax_embedder_test minmax_embedder)
+
+#
+# Driver
+#
+add_executable(minmax_embedder_driver "${DRIVER}")
+target_link_libraries(minmax_embedder_driver minmax_embedder)
+target_link_libraries(minmax_embedder_driver arser)
+target_link_libraries(minmax_embedder_driver safemain)
+target_link_libraries(minmax_embedder_driver vconone)
```
jinevening, line 52: You may need to update `requires.cmake` to include these modules.

jinevening (approved):
LGTM

compiler/minmax-embedder/requires.cmake
```diff
@@ -1,2 +1,5 @@
+require("arser")
 require("loco")
 require("luci")
+require("safemain")
+require("vconone")
```
hseok-oh, line 5: This change forces to build `arser`, `safemain`, and `vonone` when we want to build library only and don't need unittest and driver. 
When we want to build library only for runtime's ODC, we will meet this issue. 
To avoid this issue, we should divide library and driver project, like other compiler projects.

hseok-oh: Above comment is my misunderstand. Forget that.

hseok-oh (approved):
LGTM

PR #11725, opened 2023-10-16
[luci/pass] Introduce fuse_slice_with_tconv

stamalakhov:
This draft introduces fuse_slice_with_tconv pass.

Running `onecc` 
```
[onecc]
one-import-tflite=False
one-import-tf=False
one-import-bcq=False
one-import-onnx=False
one-optimize=True
one-quantize=False
one-codegen=False
one-profile=False

[one-optimize]
input_path=/home/stanislav/repos/WorkSpaces/fuse_tconv_slice/model_original.circle
output_path=/home/stanislav/repos/WorkSpaces/fuse_tconv_slice/model_fused.opt.circle
fuse_slice_with_tconv=True
````
successfully produced  result (see attachment).
[fuse_tconv_slice.zip](https://github.com/Samsung/ONE/files/12930049/fuse_tconv_slice.zip)

Related: #10960
ONE-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

Issue #10960, opened 2023-06-30
Compiler FE: Fuse Slice with TransposeConv


jinevening:
## What

Let's fuse slice Op with TransposeConv.

The below transformation is possible.

Before
```
TConv (padding: valid) -> Slice (Slice edges)
```

After
```
TConv (padding: same)
```

It is especially effective when batchnorm (channel wise MUL/ADD) appears after `TConv -> Slice` pattern, because batchnorm can be fused with the preceding TConv.

Useful link to understand padding of TConv: https://towardsdatascience.com/understand-transposed-convolutions-and-build-your-own-transposed-convolution-layer-from-scratch-4f5d97b2967

stamalakhov:
@jinevening 
M.b. i can try to resolve this issue?

jinevening:
Of course. Thank you :)

jinevening:
Let's close this issue. Thanks @stamalakhov

PR #11735, opened 2023-10-18
[luci/pass] Introduce FuseSliceWithTConvPass

stamalakhov:
This commit introduces FuseSliceWithTConvPass
for fusing slice with tconv if possible.

Its correctness is tested at #11725.
Draft: #11725
Related: https://github.com/Samsung/ONE/issues/10960

### DIFF
--- a/compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
+++ b/compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
@@ -1,0 +1,288 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FuseSliceWithTConvPass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/Profile/CircleNodeOrigin.h>
+#include <luci/Service/CircleNodeClone.h>
+#include <luci/Service/Nodes/CircleConst.h>
+
+namespace
+{
+
+#define RETURN_FALSE_UNLESS(cond) \
+  if (not(cond))                  \
+    return false;
+
+inline int32_t compute_input_size(luci::Padding padding, int32_t image_size, int32_t filter_size,
+                                  int32_t stride)
+{
+  switch (padding)
+  {
+    case luci::Padding::SAME:
+      return (image_size + stride - 1) / stride;
+    case luci::Padding::VALID:
+      return (image_size + stride - filter_size) / stride;
+    default:
+      throw std::runtime_error("Unsupported padding");
+  }
+}
+
+inline int32_t extract_pad_value(int32_t stride, int32_t in_size, int32_t filter_size,
+                                 int32_t out_size)
+{
+  const int32_t padding = ((in_size - 1) * stride + filter_size - out_size) / 2;
+  return padding > 0 ? padding : 0;
+}
+
+inline uint32_t cal_offset(const luci::CircleConst *node, const uint32_t *indices)
+{
+  return indices[0] * node->dim(1).value() * node->dim(2).value() * node->dim(3).value() +
+         indices[1] * node->dim(2).value() * node->dim(3).value() +
+         indices[2] * node->dim(3).value() + indices[3];
+}
+
+luci::Padding compute_padding(const luci::CircleTransposeConv *tconv, int32_t out_height,
+                              int32_t out_width, int32_t pad_top, int32_t pad_left)
+{
+  auto const filter = dynamic_cast<luci::CircleConst *>(tconv->filter());
+  if (!filter)
+    return luci::Padding::UNDEFINED;
+
+  auto tconv_shape = dynamic_cast<luci::CircleConst *>(tconv->inputSizes());
+  if (!tconv_shape)
+    return luci::Padding::UNDEFINED;
+
+  luci::Padding padding = luci::Padding::UNDEFINED;
+  std::initializer_list<luci::Padding> paddings_to_check = {luci::Padding::VALID,
+                                                            luci::Padding::SAME};
+
+  auto const filter_height = filter->dim(1).value();
+  auto const filter_width = filter->dim(2).value();
+  auto const stride_height = tconv->stride()->h();
+  auto const stride_width = tconv->stride()->w();
+
+  for (auto padding_to_check : paddings_to_check)
+  {
+    auto const in_height =
+      compute_input_size(padding_to_check, out_height, filter_height, stride_height);
+    auto const pad_top_virtual =
+      extract_pad_value(stride_height, in_height, filter_height, out_height);
+    if (pad_top_virtual != pad_top)
+      continue;
+
+    auto const in_width =
+      compute_input_size(padding_to_check, out_width, filter_width, stride_width);
+    auto const pad_left_virtual =
+      extract_pad_value(stride_width, in_width, filter_width, out_width);
+    if (pad_left_virtual == pad_left)
+    {
+      padding = padding_to_check; // correct padding is found
+      break;
+    }
+  }
+
+  return padding;
+}
+
+/**
+ *  Fuse Slice with CircleTransposeConv if possible
+ *
+ *  NOTE: In case predecessor of slice is tconv, we can try to merge slice with tconv,
+ *  because spatial slice is reduction so as padding for tconv,
+ *  while channels slice reduction can be directly modeled in tconv.
+ *  For now there is no option to set explicitely pad values for
+ *  CircleTransposeConv. Only using VALID/SAME and output shape is the only way
+ *  to set pad values. That is why not all numerical values of pad are legal for such
+ *  transform.
+ *
+ *  BEFORE
+ *                    |
+ *           [CircleTransposeConv]
+ *                    |
+ *               [CircleSlice]
+ *                    |
+ *
+ *  AFTER
+ *                    |
+ *            [CircleTransposeConv] (with m.b. changed padding, output shape, and filter/bias)
+ *                    |
+ *
+ */
+
+bool fuse_slice_with_tconv(luci::CircleSlice *slice)
+{
+  // NOTE: assume NHWC layout
+  auto tconv = dynamic_cast<luci::CircleTransposeConv *>(slice->input());
+  RETURN_FALSE_UNLESS(tconv != nullptr);
+
+  // offset
+  auto begin = dynamic_cast<luci::CircleConst *>(slice->begin());
+  // sanity check
+  RETURN_FALSE_UNLESS(begin != nullptr && begin->dtype() == loco::DataType::S32 &&
+                      begin->rank() == 1);
+
+  // output shape
+  auto out_shape = dynamic_cast<luci::CircleConst *>(slice->size());
+  // sanity check
+  RETURN_FALSE_UNLESS(out_shape != nullptr && out_shape->dtype() == loco::DataType::S32 &&
+                      out_shape->rank() == 1);
+
+  // output shape of tconv
+  auto tconv_shape = dynamic_cast<luci::CircleConst *>(tconv->inputSizes());
+  // sanity check
+  RETURN_FALSE_UNLESS(tconv_shape != nullptr && tconv_shape->dtype() == loco::DataType::S32 &&
+                      tconv_shape->rank() == 1);
+
+  // no update if batch dimension is processed in slice
+  RETURN_FALSE_UNLESS(begin->at<loco::DataType::S32>(0) == 0 &&
+                      out_shape->at<loco::DataType::S32>(0) ==
+                        tconv_shape->at<loco::DataType::S32>(0));
+
+  // filter
+  auto const tconv_filter = dynamic_cast<luci::CircleConst *>(tconv->filter());
+  // sanity check
+  RETURN_FALSE_UNLESS(tconv_filter != nullptr && tconv_filter->rank() == 4 &&
+                      tconv_filter->dtype() == loco::DataType::FLOAT32);
+
+  // bias
+  auto const tconv_bias = dynamic_cast<luci::CircleConst *>(tconv->bias());
+  // Only support const bias
+  // TODO Support non-const bias
+  RETURN_FALSE_UNLESS(tconv_bias != nullptr && tconv_bias->rank() == 1 &&
+                      tconv_bias->dtype() == loco::DataType::FLOAT32);
+
+  auto const out_height = out_shape->at<loco::DataType::S32>(1);
+  auto const out_width = out_shape->at<loco::DataType::S32>(2);
+
+  auto const pad_top = begin->at<loco::DataType::S32>(1);
+  auto const pad_left = begin->at<loco::DataType::S32>(2);
+
+  // As there is no option to set numerical values of pad explicitly for CircleTransposeConv
+  // we need to be sure that interpretation of PADDING + OUTPUT_SHAPE will produce
+  // the pad values, defined by slice. If possible compute_padding will return correct
+  // padding value, otherwise it will return UNDEFINED
+  auto const padding = compute_padding(tconv, out_height, out_width, pad_top, pad_left);
+  if (padding == luci::Padding::UNDEFINED)
+    return false; // impossible to fuse
+
+  auto const out_channels = out_shape->at<loco::DataType::S32>(3);
+  // update filter and bias in case it's needed
+  loco::Node *fused_filter = tconv->filter();
+  loco::Node *fused_bias = tconv->bias();
+  // Channel-direction slice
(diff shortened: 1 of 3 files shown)


compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,286 @@
...
+    if (pad_top_virtual != pad_top)
+      continue;
+
+    auto const in_width =
+      compute_input_size(padding_to_check, out_width, filter_width, stride_width);
+    auto const pad_left_virtual =
+      extract_pad_value(stride_width, in_width, filter_width, out_width);
+    if (pad_left_virtual == pad_left)
+    {
+      padding = padding_to_check; // adissible paddding is found
```
seanshpark, line 94: what is `adissible` ?

seanshpark: `paddding` -> `padding` typo?

stamalakhov: Ahh. Sorry. Type. I meant `admissible`.

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,286 @@
...
+bool fuse_slice_with_tconv(luci::CircleSlice *slice)
+{
+  // NOTE: assume NHWC layout
+  auto tconv = dynamic_cast<luci::CircleTransposeConv *>(slice->input());
+  RETURN_FALSE_UNLESS(tconv != nullptr);
+
+  auto const tconv_users = loco::succs(tconv);
+  // in general no fusion is possible if tconv is used by many successors
+  RETURN_FALSE_UNLESS(tconv_users.size() == 1);
+
```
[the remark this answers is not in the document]
stamalakhov: @jinevening 
As far i understand:
in case all users of `tconv` are different slices, then this pass will produce as many `TConv`s as users. Is it intended?

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,286 @@
...
+  {
+    auto const tconv_filter = dynamic_cast<luci::CircleConst *>(tconv->filter());
+    // sanity check
+    RETURN_FALSE_UNLESS(tconv_filter != nullptr && tconv_filter->rank() == 4 &&
+                        tconv_filter->dtype() == loco::DataType::FLOAT32);
+
+    auto const in_channels = tconv_filter->dim(3).value();
+
+    luci::CircleConst *fused_tconv_filter =
+      loco::must_cast<luci::CircleConst *>(luci::clone_node(tconv_filter, slice->graph()));
```
jinevening, line 187: Node is created in this line, meaning that graph is changed. But in the below line, this function can return false, e.g., if bias is non-const.

Could you check all conditions first and then create nodes?

stamalakhov: ok

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,286 @@
...
+  auto const padding = compute_padding(tconv, out_height, out_width, pad_top, pad_left);
+  if (padding == luci::Padding::UNDEFINED)
+    return false; // no fuse is possible
+
+  auto const out_channels = size->at<loco::DataType::S32>(3);
+  // update filter and bias in case it's needed
+  loco::Node *fused_filter = tconv->filter();
+  loco::Node *fused_bias = tconv->bias();
+  if (begin->at<loco::DataType::S32>(3) != 0 ||
+      out_channels != tconv_shape->at<loco::DataType::S32>(3))
```
jinevening, line 177:
```suggestion
  // Channel-direction slice
  if (begin->at<loco::DataType::S32>(3) != 0 ||
      out_channels != tconv_shape->at<loco::DataType::S32>(3))
```

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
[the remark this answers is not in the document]
stamalakhov: @jinevening
Should i remove this check or change comment?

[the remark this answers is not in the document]
jinevening: Yes. That's how our passes work in general :) We can limit the case to work just for a single user (e.g., for peformance). Then we may update the comment (the current comment is somewhat misleading).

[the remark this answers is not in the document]
stamalakhov: ok

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
stamalakhov: > `paddding` -> `padding` typo?

done.

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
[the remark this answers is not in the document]
stamalakhov: done

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,287 @@
...
+    if (pad_top_virtual != pad_top)
+      continue;
+
+    auto const in_width =
+      compute_input_size(padding_to_check, out_width, filter_width, stride_width);
+    auto const pad_left_virtual =
+      extract_pad_value(stride_width, in_width, filter_width, out_width);
+    if (pad_left_virtual == pad_left)
+    {
+      padding = padding_to_check; // correct padding is found
```
jinevening, line 94: Q) Is it okay not to check bottom/right padding?

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,287 @@
...
+  RETURN_FALSE_UNLESS(tconv != nullptr);
+
+  // offset
+  auto begin = dynamic_cast<luci::CircleConst *>(slice->begin());
+  // sanity check
+  RETURN_FALSE_UNLESS(begin != nullptr && begin->dtype() == loco::DataType::S32 &&
+                      begin->rank() == 1);
+
+  // output shape
+  auto size = dynamic_cast<luci::CircleConst *>(slice->size());
```
jinevening, line 140: Better to rename the variable for clarity.

```suggestion
  auto out_shape = dynamic_cast<luci::CircleConst *>(slice->size());
```

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,287 @@
...
+  auto const padding = compute_padding(tconv, out_height, out_width, pad_top, pad_left);
+  if (padding == luci::Padding::UNDEFINED)
+    return false; // no fuse is possible
+
+  auto const out_channels = size->at<loco::DataType::S32>(3);
+  // update filter and bias in case it's needed
+  loco::Node *fused_filter = tconv->filter();
+  loco::Node *fused_bias = tconv->bias();
+  // Channel-direction slice
+  // Weights/Bias are removed.
```
jinevening, line 185:
```suggestion
  // Corresponding weights/bias of TConv is sliced.
```

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
stamalakhov: > Q) Is it okay not to check bottom/right padding?

As far as i know - yes. It's okay.  Bottom/right paddings are just for correcting size of resulting TConv dimension.

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
stamalakhov: ok

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,287 @@
...
+  auto const pad_top = begin->at<loco::DataType::S32>(1);
+  auto const pad_left = begin->at<loco::DataType::S32>(2);
+
+  // As there is no option to set numerical values of pad explicitly for CircleTransposeConv
+  // we need to be sure that interpretation of PADDING + OUTPUT_SHAPE will produce
+  // the pad values, defined by slice. If possible compute_padding will return correct
+  // padding value, otherwise it will return UNDEFINED
+  auto const padding = compute_padding(tconv, out_height, out_width, pad_top, pad_left);
+  if (padding == luci::Padding::UNDEFINED)
+    return false; // no fuse is possible
```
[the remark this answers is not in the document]
stamalakhov: ok

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,287 @@
...
+  // Channel-direction slice
+  // Weights/Bias are removed.
+  if (begin->at<loco::DataType::S32>(3) != 0 ||
+      out_channels != tconv_shape->at<loco::DataType::S32>(3))
+  {
+    // fused filter
+    auto const in_channels = tconv_filter->dim(3).value();
+
+    luci::CircleConst *fused_tconv_filter =
+      loco::must_cast<luci::CircleConst *>(luci::clone_node(tconv_filter, slice->graph()));
```
jinevening, line 193:
```suggestion
    luci::CircleConst *fused_tconv_filter = luci::clone(tconv_filter);
```

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,287 @@
...
+ *  transform.
+ *
+ *  BEFORE
+ *                    |
+ *           [CircleTransposeConv]
+ *                    |
+ *               [CircleSlice]
+ *                    |
+ *
+ *  AFTER (if possible)
```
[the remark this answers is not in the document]
stamalakhov: ok

seanshpark:
@stamalakhov , how about adding several test recipes for padding variations in your DRAFT PR?

stamalakhov:
> @stamalakhov , how about adding several test recipes for padding variations in your DRAFT PR?

ok.

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,287 @@
...
+        }
+      }
+    }
+    fused_tconv_filter->name(tconv_filter->name() + "/FusedSlice");
+    luci::add_origin(fused_tconv_filter, luci::get_origin(tconv_shape));
+    fused_filter = fused_tconv_filter;
+
+    // fused bias
+    luci::CircleConst *fused_tconv_bias =
+      loco::must_cast<luci::CircleConst *>(luci::clone_node(tconv_bias, slice->graph()));
```
jinevening, line 223:
```suggestion
    luci::CircleConst *fused_tconv_bias = luci::clone(tconv_bias);
```

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,286 @@
...
+      tconv_bias->at<loco::DataType::FLOAT32>(c) = data;
+    }
+
+    fused_tconv_bias->name(tconv_bias->name() + "/FusedSlice");
+    luci::add_origin(fused_tconv_bias, luci::get_origin(tconv_bias));
+    fused_bias = fused_tconv_bias;
+  }
+
+  auto *fused_tconv_shape =
+    loco::must_cast<luci::CircleConst *>(luci::clone_node(tconv_shape, slice->graph()));
```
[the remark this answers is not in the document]
stamalakhov: Yes. Sorry. I'll fix that.

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,287 @@
...
+  auto const tconv_filter = dynamic_cast<luci::CircleConst *>(tconv->filter());
+  // sanity check
+  RETURN_FALSE_UNLESS(tconv_filter != nullptr && tconv_filter->rank() == 4 &&
+                      tconv_filter->dtype() == loco::DataType::FLOAT32);
+
+  // bias
+  auto const tconv_bias = dynamic_cast<luci::CircleConst *>(tconv->bias());
+  // sanity check
+  RETURN_FALSE_UNLESS(tconv_bias != nullptr && tconv_bias->rank() == 1 &&
+                      tconv_bias->dtype() == loco::DataType::FLOAT32);
```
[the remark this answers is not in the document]
stamalakhov: done

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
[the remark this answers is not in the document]
stamalakhov: done

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
jinevening: Can you add a test for this case in draft, i.e., top padding != bottom padding?

stamalakhov: ok

stamalakhov: @jinevening 
To clarify:
https://github.com/Samsung/ONE/blob/b284f4f63ca5049385a7dd8c47a8e535b86617e9/compiler/luci-interpreter/src/kernels/TransposeConv.cpp#L73-L76

TConv kernel doesn't use pad_bottom or pad_right, they are used implicitly via spatial size reduction.
So this PR doesn't use bottom/right pads.
The only effort made in this PR is to make sure values for top/left padding, (which are the only values used after deciphering `VALID/SAME+kernel+shape=>(top/left pads)`) will be the same as those used in `luci-interpreter'. 
Actually current test in draft (it's yet `TConvSlice_000/test.recipe`) uses (pad_top = pad_left = 0)  while (pad_bottom = pad_right = 3).

jinevening: You've already added tests. Thanks.

For `Net_TConv_Slice_000`,
- Before
  - [1, 8, 8, 32] ->  TConv [1, 18, 18, 32] (valid) -> Slice [1, 15, 15, 32]
  - Slice Op slices like `pad_top` = 0, `pad_bottom` = 3.
- After
  - [1, 8, 8, 32] ->  TConv [1, 15, 15, 32] (same)
  - Padding would be `pad_top` = 1, `pad_bottom` = 2 to make them balanced as much as possible.

The outputs would be different, but the test passed.. Could you help my understanding? What am I missing here?

stamalakhov: > Could you help my understanding?

@jinevening 
As far as i understand,
the only thing that matters for interpreter is `_padding_height/_padding_left` at https://github.com/Samsung/ONE/blob/b284f4f63ca5049385a7dd8c47a8e535b86617e9/compiler/luci-interpreter/src/kernels/TransposeConv.cpp#L73-L76.

> * [1, 8, 8, 32] ->  TConv [1, 15, 15, 32] (same)

To my knowledge padding of resulting TConv would be `valid`, please see the results of running pass in attachment.
[Net_TConv_Slice_000..zip](https://github.com/Samsung/ONE/files/13050547/Net_TConv_Slice_000.zip)
so resulting pads for `valid` tconv would be (pad_top = 0, pad_bottom = 3), the same as in `slice`.

Please note that the code in PR tries to produce such values for tconv's `padding`, which would produce the same values for `_padding_height/_padding_left`  in interpreter.

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
```diff
@@ -0,0 +1,289 @@
...
+{
+  switch (padding)
+  {
+    case luci::Padding::SAME:
+      return (image_size + stride - 1) / stride;
+    case luci::Padding::VALID:
+      return (image_size + stride - filter_size) / stride;
+    default:
+      assert(false);
+      return 0;
```
jinevening, line 42: Please throw an exception in this case.

stamalakhov: Ahh. ok.

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
stamalakhov: @jinevening 
Just a note:
I believe that seamless fusion of `Slice` with `TConv` would be possible if there was something like `PaddedTConv` with an option to explicitly set pad numerical values.

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
stamalakhov: done

compiler/luci/pass/src/FuseSliceWithTConvPass.cpp
stamalakhov: > Thanks for explanation.

Glad to be helpful.

jinevening (approved):
LGTM

@stamalakhov Can you check if there is any side effect on our backend compiler after this conversion is added to `O1` option? This pass will affect a lot of models, so we may need to check regression to add this into `O1`.

stamalakhov:
> Can you check if there is any side effect on our backend compiler after this conversion is added to `O1` option? This pass will affect a lot of models, so we may need to check regression to add this into `O1`.

ok. I'll try.

seanshpark (approved):
LGTM

PR #11745, opened 2023-10-20
[tools/onert_train] Change default learning rate

jyoungyun:
This commit changes the default learning rate value to 0.001.


jyoungyun:
The value `0.001` is not an absolute value. However, we usually compare the result of tensorflow, and tensorflow sets the 0.001 as the default learning rate. So it changes the default value to make it easier to test.

hseok-oh (approved):
LGTM

ragmani (approved):
LGTM

Issue #11741, opened 2023-10-19
[tflchef/circlechef] Needs of dedicated type for custom operator


mhs4670go:
## What

Let's introduce a dedicated type for custom operators in the recipe.

## Why

Recipe has `Operation` message as below.

https://github.com/Samsung/ONE/blob/b284f4f63ca5049385a7dd8c47a8e535b86617e9/compiler/tflchef/proto/tflchef.proto#L554-L558

And, `type` key decides operator codes. For example, `BroadcastTo` operator can be created with below recipe. Even though the operator is custom one, it's gonna be find until `tflchef` has [its information](https://github.com/Samsung/ONE/tree/b284f4f63ca5049385a7dd8c47a8e535b86617e9/compiler/tflchef/core/src/CustomOp).
```
operation {
  type: "BroadcastTo"
  input: "bc_input"
  input: "bc_shape"
  output: "bc_ofm"
}
```

But, as the version of tensorflow goes up, some of custom operators have become built-in operators. Which means that there is no way for the recipe to distinguish between a custom operator and a built-in operator because both operator codes are written into the same key - `type` key.

Even though some custom operator becomes built-in operators, we should support both of them because of backward compatibility - for example, `BroadcastTo` as a custom op and `BroadcastTo` as a built-in op.

So, let's revisit this policy like below.

## TO-BE

- Custom operators' type in the recipe is `Custom`.
- `Operation` message introduces `custom_code` optional string key.

Then, above recipe would be like belows.

### BroadcastTo as a custom operator
```
operation {
  type: "Custom"
  custom_code: "BroadcastTo"
  input: "bc_input"
  input: "bc_shape"
  output: "bc_ofm"
}
```

### BroadcastTo as a built-in operator
```
operation {
  type: "BroadcastTo"
  input: "bc_input"
  input: "bc_shape"
  output: "bc_ofm"
}
```

Seunghui98:
@mhs4670go 
All done !! :)

PR #11750, opened 2023-10-21
[Draft][tflchef/circlechef] Use a dedicated type for custom operators

Seunghui98:
On going draft to use a dedicated type for custom operators in the recipe.

---
Related Issue : https://github.com/Samsung/ONE/issues/11741

### DIFF
--- a/compiler/circlechef/core/src/ModelChef.cpp
+++ b/compiler/circlechef/core/src/ModelChef.cpp
@@ -135,9 +135,11 @@ gather_builtincode_map(const ::circlechef::ModelRecipe &model_recipe)
 
   for (const auto &operation : model_recipe.operation())
   {
-    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
-    if (op_chef->code() == circle::BuiltinOperator_CUSTOM)
+    if ((operation.has_extype() && operation.extype() == "Custom") ||
+        (operation.has_type() && operation.type() == "Custom"))
       continue;
+
+    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
 
     // Various operation version is unified as the highest version among them
     if (builtin_map.find(op_chef->code()) == builtin_map.end() ||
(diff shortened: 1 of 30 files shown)


compiler/circlechef/core/src/ModelChef.cpp
```diff
@@ -171,6 +177,12 @@ std::set<std::string> gather_customcode_set(const ::circlechef::ModelRecipe &mod
   std::set<std::string> customcode_set;
   for (const auto &operation : model_recipe.operation())
   {
+    if (operation.type() == "Custom")
+    {
+      customcode_set.insert(operation.custom_code());
+      continue;
+    }
```
mhs4670go, line 184: These codes can substitute below codes.
```cxx
auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
if (op_chef->code() == circle::BuiltinOperator_CUSTOM)
  customcode_set.insert(operation.type());
```

Which means above codes should be removed.

compiler/circlechef/core/src/ModelChef.cpp
```diff
@@ -182,6 +194,12 @@ std::set<std::string> gather_customcode_set(const ::circlechef::ModelRecipe &mod
     const auto &graph = model_recipe.graph(g);
     for (const auto &operation : graph.operation())
     {
+      if (operation.type() == "Custom")
```
mhs4670go, line 197: ditto

res/TensorFlowLiteRecipes/All_000/test.recipe
```diff
@@ -18,13 +18,14 @@ operand {
   }
 }
 operation {
-  type: "All"
+  type: "Custom"
```
seanshpark, line 21: to land separate PRs for `compiler` and `res` files,
we need to introduce some intermediate changes
for example,
1/ in `res`, introduce `extype: "Custom"` and `custom_code: "All"`
2/ in `tflchef`, if `extype` exist, use that, if not use `type`
3/ in `res`, change `type` and remove `extype`
4/ in `tflchef` remove related with `extype`

Seunghui98: Thank you for your comment!! I will upload the code PR as you guided me. :)

compiler/tflchef/core/src/CustomOp/AddV2.cpp
```diff
@@ -29,7 +29,11 @@ AddV2Chef::custom_value(flatbuffers::FlatBufferBuilder &fbb) const
 {
   auto &operation = (*_operation);
 
-  assert(operation.type() == "AddV2");
+  if (operation.has_extype() && operation.extype() == "Custom")
```
seanshpark, line 32: why not check `type() == `Custom` too ?

compiler/tflchef/core/src/CustomOp/AddV2.cpp
```diff
@@ -29,7 +29,11 @@ AddV2Chef::custom_value(flatbuffers::FlatBufferBuilder &fbb) const
...
+  {
+    assert(operation.has_custom_code());
+    assert(operation.custom_code() == "AddV2");
+  }
```
seanshpark, line 36: as these blocks are repeated several times, would be better to add a method

PR #11749, opened 2023-10-20
[circledump] OpPrinter should print enum name for Padding

glistening:
OpPrinter printed enum value, not enum name for Padding.
It caused circledump's output as binary.
Now, it will print enum Padding name.

ONE-DCO-1.0-Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>

Related: https://github.com/Samsung/ONE/pull/11699#issuecomment-1772717795

With this PR, padding is printed correctly.

### BEFORE
```
$ circledump Net_Conv_Relu6_000.circle
...
O(0:0) CONV_2D 
    Padding(^@) Stride.W(1) Stride.H(1) Dilation.W(1) Dilation.H(1) Activation(NONE)
    I T(0:0) Placeholder
    I T(0:1) Conv2D_1
    I T(0:2) Conv2D_2
    O T(0:4) Conv2D_11
O(0:1) RELU6 
    I T(0:4) Conv2D_11
    O T(0:5) ReLU6
O(0:2) CONV_2D 
    Padding(^@) Stride.W(1) Stride.H(1) Dilation.W(1) Dilation.H(1) Activation(NONE)
    I T(0:5) ReLU6
    I T(0:3) Conv2D_21
    I T(0:2) Conv2D_2
    O T(0:6) Conv2D_22
O(0:3) RELU6 
    I T(0:6) Conv2D_22
    O T(0:7) ReLU6_1

Inputs/Outputs: I(input)/O(output) T(tensor index) OperandName
I T(0:0) Placeholder
O T(0:7) ReLU6_1
```

### AFTER
```
$ circledump Net_Conv_Relu6_000.circle
...
O(0:0) CONV_2D 
    Padding(SAME) Stride.W(1) Stride.H(1) Dilation.W(1) Dilation.H(1) Activation(NONE)
    I T(0:0) Placeholder
    I T(0:1) Conv2D_1
    I T(0:2) Conv2D_2
    O T(0:4) Conv2D_11
O(0:1) RELU6 
    I T(0:4) Conv2D_11
    O T(0:5) ReLU6
O(0:2) CONV_2D 
    Padding(SAME) Stride.W(1) Stride.H(1) Dilation.W(1) Dilation.H(1) Activation(NONE)
    I T(0:5) ReLU6
    I T(0:3) Conv2D_21
    I T(0:2) Conv2D_2
    O T(0:6) Conv2D_22
O(0:3) RELU6 
    I T(0:6) Conv2D_22
    O T(0:7) ReLU6_1

[...]

PR #11758, opened 2023-10-24
[tfldump] OpPrinter should print enum name for Padding

glistening:
OpPrinter printed enum value, not enum name for Padding. It caused tfldump's output as binary.
Now, it will print enum Padding name.

It is a twin PR for `tfldump`.

Related: #11749


hseok-oh (approved):
LGTM

PR #11765, opened 2023-10-24
[onert] Rename quantization API

glistening:
It refines quantization API to use consistent style with nnfw train API.

### DIFF
--- a/runtime/onert/api/include/nnfw_experimental.h
+++ b/runtime/onert/api/include/nnfw_experimental.h
@@ -337,8 +337,8 @@ NNFW_STATUS nnfw_train_expected_tensorinfo(nnfw_session *session, uint32_t index
  *  On-Device Quantization APIs
  *
  * On-Device Quantization APIs are designed to be used in the following order
- * 1. nnfw_set_quantization_type
- * 2. nnfw_set_quantized_model_path
+ * 1. nnfw_quantize_set_qtype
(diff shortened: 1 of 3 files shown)


glistening:
It is 1st step to refine quantization API. I wrote this PR as a suggestion to get your opinion.

Training API has the prefix of `nnfw_train_...`.
It would be good to use the same convention for Quantization API.

runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -382,7 +382,7 @@ NNFW_STATUS nnfw_set_quantization_type(nnfw_session *session, NNFW_QUANTIZE_TYPE
  * @param[in] path    Quantized model path
  * @return    @c NNFW_STATUS_NO_ERROR if successful, otherwise return @c NNFW_STATUS_ERROR
  */
-NNFW_STATUS nnfw_set_quantized_model_path(nnfw_session *session, const char *path);
+NNFW_STATUS nnfw_quantize_set_qmodel_path(nnfw_session *session, const char *path);
```
glistening, line 385: I am thinking of `working directory`. Because there are several files (e.g. minmax recording, ... to be saved. `quantized model` is not the only one. Also, we need to generate the quantized model under some rule, not arbitrary name by user.

chunseoklee:
prefix nnfw_quantize looks fine to me.

Issue #11760, opened 2023-10-24
[luci-interpreter] Support cos, sin


seanshpark:
Out internal model has `cos` and `sin` Op, which would be nice to have value test.
To support value testing, we need to implement `luci-interpreter` kernels for these.

Support `F32` is enough for now.

seanshpark:
done

PR #11778, opened 2023-10-25
DRAFT luci-interpreter Cos, Sin and Tile

seanshpark:
on-going draft to support cos, sin in luci-interpreter.


seanshpark:
for  #11760

PR #11785, opened 2023-10-25
[tflchef] Use extype and custom_code type for custom operators

Seunghui98:
This introduces extype and custom_code for custom operators.

---

### DIFF
--- a/compiler/tflchef/core/src/ModelChef.cpp
+++ b/compiler/tflchef/core/src/ModelChef.cpp
@@ -141,10 +141,10 @@ gather_builtincode_map(const ::tflchef::ModelRecipe &model_recipe)
 
   for (const auto &operation : model_recipe.operation())
   {
-    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
-    if (op_chef->code() == tflite::BuiltinOperator_CUSTOM)
+    if ((operation.has_extype() && operation.extype() == "Custom") || operation.type() == "Custom")
       continue;
 
+    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
     // Various operation version is unified as the highest version among them
     if (builtin_map.find(op_chef->code()) == builtin_map.end() ||
         builtin_map[op_chef->code()] < operation.version())
@@ -157,10 +157,11 @@ gather_builtincode_map(const ::tflchef::ModelRecipe &model_recipe)
     const auto &graph = model_recipe.graph(g);
     for (const auto &operation : graph.operation())
     {
-      auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
-      if (op_chef->code() == tflite::BuiltinOperator_CUSTOM)
+      if ((operation.has_extype() && operation.extype() == "Custom") ||
+          operation.type() == "Custom")
         continue;
 
+      auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
       // Various operation version is unified as the highest version among them
       if (builtin_map.find(op_chef->code()) == builtin_map.end() ||
           builtin_map[op_chef->code()] < operation.version())
@@ -177,9 +178,11 @@ std::set<std::string> gather_customcode_set(const ::tflchef::ModelRecipe &model_
   std::set<std::string> customcode_set;
   for (const auto &operation : model_recipe.operation())
   {
-    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
-    if (op_chef->code() == tflite::BuiltinOperator_CUSTOM)
-      customcode_set.insert(operation.type());
+    if ((operation.has_extype() && operation.extype() == "Custom") || operation.type() == "Custom")
+    {
+      assert(operation.has_custom_code());
+      customcode_set.insert(operation.custom_code());
+    }
   }
 
   // Add ops used in Graphs(subgraphs)
@@ -188,9 +191,12 @@ std::set<std::string> gather_customcode_set(const ::tflchef::ModelRecipe &model_
     const auto &graph = model_recipe.graph(g);
     for (const auto &operation : graph.operation())
     {
-      auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
-      if (op_chef->code() == tflite::BuiltinOperator_CUSTOM)
-        customcode_set.insert(operation.type());
+      if ((operation.has_extype() && operation.extype() == "Custom") ||
+          operation.type() == "Custom")
+      {
+        assert(operation.has_custom_code());
+        customcode_set.insert(operation.custom_code());
+      }
     }
   }
 
@@ -619,7 +625,11 @@ template <typename T> std::map<std::string, int32_t> cook_graph(const T &graph, 
   {
     assert(operation.has_type());
 
-    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
+    std::string op_type = operation.type();
+    if (operation.has_custom_code())
+      op_type = operation.custom_code();
+
+    auto op_chef = op_chef_registry().lookup(op_type).create(&operation);
 
     // Create 'inputs'
     std::vector<int32_t> input_vec = as_dataset(operation.input()).map(lookup).vectorize();
@@ -650,7 +660,9 @@ template <typename T> std::map<std::string, int32_t> cook_graph(const T &graph, 
     // custom operator
     else
     {
-      auto op_it = std::find(custom_code_vec.begin(), custom_code_vec.end(), operation.type());
+      assert(operation.has_custom_code());
+      auto op_it =
+        std::find(custom_code_vec.begin(), custom_code_vec.end(), operation.custom_code());
       assert(op_it != custom_code_vec.end());
       opcode_index = builtin_code_map.size();
       opcode_index += std::distance(custom_code_vec.begin(), op_it);
--- a/compiler/tflchef/core/src/ModelChef.cpp
+++ b/compiler/tflchef/core/src/ModelChef.cpp
@@ -141,7 +141,7 @@ gather_builtincode_map(const ::tflchef::ModelRecipe &model_recipe)
 
   for (const auto &operation : model_recipe.operation())
   {
-    if ((operation.has_extype() && operation.extype() == "Custom") || operation.type() == "Custom")
+    if (operation.type() == "Custom" || (operation.has_extype() && operation.extype() == "Custom"))
       continue;
 
     auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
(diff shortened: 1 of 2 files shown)


compiler/tflchef/core/src/CustomOp/AddV2.cpp
```diff
@@ -29,7 +29,7 @@ AddV2Chef::custom_value(flatbuffers::FlatBufferBuilder &fbb) const
 {
   auto &operation = (*_operation);
 
-  assert(operation.type() == "AddV2");
+  assert(operation.custom_code() == "AddV2");
```
seanshpark, line 32: I thought there will be `type` check as `CustomOp`.
Draft doesn't do the check. Am I missing something?

seanshpark: Q) what happens when user missed to add `custom_code` in the `.recipe` file?

Seunghui98: @seanshpark 
Thank you for comments! 
> Q) what happens when user missed to add custom_code in the .recipe file?

Perhaps, if the user forgets to add `custom_code` in recipe file, it will be caught in the assert code... 😅
What do you think about changing this part to a code that checks with `custom_type` if there is `extype` and `type` if there is no `extype`?

seanshpark: logic I think would be is described in https://github.com/Samsung/ONE/pull/11750#discussion_r1368165801 . `extype` or `type` check if `CustomOp` would be `2/` .

seanshpark: > it will be caught in the assert code... 

which one? please point out in DRAFT

seanshpark: >  if there is extype and type if there is no extype?

I'm not sure how this question is about my `Q) what happens when user missed ... `
Can you come again?

compiler/tflchef/core/src/CustomOp/AddV2.cpp
```diff
@@ -29,7 +29,11 @@ AddV2Chef::custom_value(flatbuffers::FlatBufferBuilder &fbb) const
 {
   auto &operation = (*_operation);
 
-  assert(operation.type() == "AddV2");
+  if (operation.has_extype() && operation.extype() == "Custom")
+  {
+    assert(operation.has_custom_code());
+    assert(operation.custom_code() == "AddV2");
+  }
```
seanshpark, line 36: plz read comment in DRAFT

compiler/tflchef/core/src/ModelChef.cpp
```diff
@@ -141,10 +141,11 @@ gather_builtincode_map(const ::tflchef::ModelRecipe &model_recipe)
 
   for (const auto &operation : model_recipe.operation())
   {
-    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
-    if (op_chef->code() == tflite::BuiltinOperator_CUSTOM)
+    if ((operation.has_extype() && operation.extype() == "Custom") ||
+        (operation.has_type() && operation.type() == "Custom"))
       continue;
```
Seunghui98, line 146: First of all, if there is a `extype` or the type is `Custom`, the code is written to determine that it is a custom op.
--> Logic to refer to (https://github.com/Samsung/ONE/pull/11750#discussion_r1368165801)
If the `extype` in recipe files is removed in step `3)`, everything related to the `extype` will be deleted again in the step `4)`

compiler/tflchef/core/src/OpChef.h
```diff
@@ -45,4 +45,14 @@ struct OpChefFactory
...
+static void CheckCustomOp(const tflchef::Operation operation, std::string op_type)
+{
+  if ((operation.has_extype() && operation.extype() == "Custom") ||
+      (operation.has_type() && operation.type() == "Custom"))
+  {
+    assert(operation.has_custom_code());
+    assert(operation.custom_code() == op_type);
+  }
+};
+
```
Seunghui98, line 57: I created a function here to avoid duplication.

Seunghui98:
@seanshpark @mhs4670go

 As you(@seanshpark ) advised, I rewrote the code.
Could you review the code again ??

compiler/tflchef/core/src/CustomOp/AddV2.cpp
```diff
@@ -28,8 +28,7 @@ flatbuffers::Offset<flatbuffers::Vector<uint8_t>>
 AddV2Chef::custom_value(flatbuffers::FlatBufferBuilder &fbb) const
 {
   auto &operation = (*_operation);
-
```
seanshpark, line 31: it would be not to remove this empty line

compiler/tflchef/core/src/OpChef.h
```diff
@@ -45,4 +45,14 @@ struct OpChefFactory
   virtual std::unique_ptr<OpChef> create(const tflchef::Operation *operation) const = 0;
 };
 
+static void CheckCustomOp(const tflchef::Operation operation, std::string op_type)
```
seanshpark, line 48: plz do not use `static` in header. there will be duplicate static functions across the callers.

seanshpark:
@Seunghui98 , this PR has multiple context of changes and title/description of this PR doesn't seem to explain the changes.

compiler/tflchef/core/src/OpChef.h
Seunghui98: @seanshpark 
Oh... I didn't know that part, Thank you for letting me know.
I don't know if I should create an `OpChef.cpp` file as an implementation for the `OpChef.h` header and
put the function in it.
Could you advise me where it would be good...?

seanshpark: > Could you advise me where it would be good...?

Please think of what the purpose of this function, and the scope of the functionality.

And looking in other global methods in `Convert.h`, please follow existing snake_case convention.

And does `CheckCustomOp` express the purpose of new method?

Naming is difficult and needs some experience, which may require try and error.

compiler/tflchef/core/src/OpChef.h
seanshpark: Q) Can you explain more about what you mean by `avoid duplication` ?

Seunghui98:
First of all, this PR will be reflected after [the previous PR](https://github.com/Samsung/ONE/pull/11813).
So, the associated logic is [here](https://github.com/Samsung/ONE/pull/11750#discussion_r1368165801).
The extype will be used for a while and will be deleted later.
Could you review the code again? (@seanshpark, @mhs4670go)

compiler/tflchef/core/src/ModelChef.cpp
```diff
@@ -177,9 +178,11 @@ std::set<std::string> gather_customcode_set(const ::tflchef::ModelRecipe &model_
   std::set<std::string> customcode_set;
   for (const auto &operation : model_recipe.operation())
   {
-    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
-    if (op_chef->code() == tflite::BuiltinOperator_CUSTOM)
-      customcode_set.insert(operation.type());
+    if ((operation.has_extype() && operation.extype() == "Custom") || operation.type() == "Custom")
+    {
+      assert(operation.has_custom_code());
```
mhs4670go, line 183: On second thought, this assert always will be true because `custom_code` is optional element, which means that it has a default value. So, this and the assert in the `check_custom_op_value` method can be removed.

mhs4670go: The check should change into check if given string is empty - default value.

Seunghui98: @mhs4670go 

Thank you for your comment ! :)
I applied it regarding comments, 
Could you look at it again ?

compiler/tflchef/core/src/ModelChef.cpp
```diff
@@ -177,9 +178,11 @@ std::set<std::string> gather_customcode_set(const ::tflchef::ModelRecipe &model_
   std::set<std::string> customcode_set;
   for (const auto &operation : model_recipe.operation())
   {
-    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
-    if (op_chef->code() == tflite::BuiltinOperator_CUSTOM)
-      customcode_set.insert(operation.type());
+    if ((operation.has_extype() && operation.extype() == "Custom") || operation.type() == "Custom")
+    {
+      assert(operation.custom_code() != "");
```
mhs4670go, line 183: This one would be more readable.
```suggestion
      assert(not operation.custom_code().empty());
```

compiler/tflchef/core/src/ModelChef.cpp
```diff
@@ -141,10 +141,10 @@ gather_builtincode_map(const ::tflchef::ModelRecipe &model_recipe)
 
   for (const auto &operation : model_recipe.operation())
   {
-    auto op_chef = op_chef_registry().lookup(operation.type()).create(&operation);
-    if (op_chef->code() == tflite::BuiltinOperator_CUSTOM)
+    if ((operation.has_extype() && operation.extype() == "Custom") || operation.type() == "Custom")
```
seanshpark, line 144: optional) as `type` is key and `extype` will be removed.
```suggestion
    if (operation.type() == "Custom" || (operation.has_extype() && operation.extype() == "Custom"))
```

compiler/tflchef/core/src/OpUtils.cpp
```diff
@@ -20,9 +20,8 @@
 
 void check_custom_op_value(const tflchef::Operation operation, std::string op_type)
 {
-  if ((operation.has_extype() && operation.extype() == "Custom") || operation.type() == "Custom")
+  if (operation.type() == "Custom" || (operation.has_extype() && operation.extype() == "Custom"))
```
seanshpark, line 23: let's change this in another PR. I don't think this change is related with purpose of this PR.

compiler/tflchef/core/src/ModelChef.cpp
```diff
@@ -650,7 +660,9 @@ template <typename T> std::map<std::string, int32_t> cook_graph(const T &graph,
     // custom operator
     else
     {
-      auto op_it = std::find(custom_code_vec.begin(), custom_code_vec.end(), operation.type());
+      assert(not operation.custom_code().empty());
+      auto op_it =
+        std::find(custom_code_vec.begin(), custom_code_vec.end(), operation.custom_code());
```
seanshpark, line 665: can we do this?
```suggestion
      auto op_it =
        std::find(custom_code_vec.begin(), custom_code_vec.end(), op_type);
```

seanshpark: actually for better readability, I hope the code be like this without line split
```suggestion
      auto custom_code = operation.custom_code();
      auto op_it = std::find(custom_code_vec.begin(), custom_code_vec.end(), custom_code);
```

Seunghui98: @seanshpark 

Oh, yes! we can.
Thank you for letting me know. :)

Seunghui98:
@seanshpark @mhs4670go 

I reflected the comments in code.
Could you review the code again??

Seunghui98:
@seanshpark @mhs4670go 
I reflected comments in `Apply comment` commit.
Could you review the code again??

seanshpark (approved):
LGTM

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM:)

PR #11797, opened 2023-10-26
[luci-value-test] Fix to use addevaltol

seanshpark:
This will fix Conv2D_U8_000 to use addevaltol()


seanshpark:
Hope this will fix CI failures.

jinevening (approved):
LGTM

Issue #11779, opened 2023-10-25
[luci-interpreter] Enable Tile Op


seanshpark:
Lets' enable TILE Op in luci-interpreter.

seanshpark:
Ah, CircleTile exist ... :)
Need to enable in `luci-interpreter`

seanshpark:
done

PR #11799, opened 2023-10-26
[luci-interpreter] Enable Tile

seanshpark:
This will enable excution of CircleTile Op.


seanshpark:
for  #11779 , from #11778

compiler/luci-interpreter/src/kernels/Tile.test.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+TEST(TileTest, MultiplesDTypeInvalid_NEG)
+{
+  std::unique_ptr<IMemoryManager> memory_manager = std::make_unique<TestMemoryManager>();
+  Shape input_shape{1, 3};
+  std::vector<float> input_data{1.0f, 2.0f, 3.0f};
+  Tensor input_tensor =
+    makeInputTensor<DataType::FLOAT32>(input_shape, input_data, memory_manager.get());
+  Shape mul_shape{3};
+  std::vector<float> mul_data{1.0f, 2.0f, 3.0f};
```
jinevening, line 62: This test checks wrong dtype, so it would be better to use the right shape.

```suggestion
  Shape mul_shape{2};
  std::vector<float> mul_data{1.0f, 2.0f};
```

compiler/luci-interpreter/src/kernels/Tile.test.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+TEST(TileTest, Float)
+{
+  std::unique_ptr<IMemoryManager> memory_manager = std::make_unique<TestMemoryManager>();
+  Shape input_shape{1, 3};
+  std::vector<float> input_data{1.0f, 2.0f, 3.0f};
+  Tensor input_tensor =
+    makeInputTensor<DataType::FLOAT32>(input_shape, input_data, memory_manager.get());
+  Shape mul_shape{2};
+  std::vector<int32_t> mul_data{1, 2};
```
jinevening, line 39: Could you add a test case for mul_data {2, 1}? The output shape will be {2, 3}, output data will be {1, 1, 2, 2, 3, 3}

compiler/luci-interpreter/src/kernels/Tile.cpp
```diff
@@ -0,0 +1,122 @@
...
+  {
+    const T *in_end = in_data + in_size;
+    T *new_out_data = std::copy(in_data, in_end, out_data);
+    in_data = out_data;
+    out_data = new_out_data;
+  }
+}
+
+template <typename T, typename M>
+std::pair<int, int> TileOneDimension(const tflite::RuntimeShape &in_dimensions, const T *in_data,
```
jinevening, line 77: The return value is not used. Why this function has return value?

seanshpark: return is used in below
```
   std::tie(stride_size, tiled_stride_size) =
      TileOneDimension(in_dimensions, copy_from_data, multiples, copy_to_data, dimension + 1)
```
code itself is from TensorFlow -_-;

compiler/luci-interpreter/src/kernels/Tile.test.cpp
seanshpark: renamed this test to `FloatMul12` and added `FloatMul21` for your suggestion case.

compiler/luci-interpreter/src/kernels/Tile.test.cpp
```diff
@@ -0,0 +1,127 @@
...
+  memory_manager->allocate_memory(output_tensor);
+  kernel.execute();
+
+  std::vector<int32_t> ref_output_shape{2, 3};
+  std::vector<float> ref_output_data{1.0f, 2.0f, 3.0f, 1.0f, 2.0f, 3.0f};
+  EXPECT_THAT(extractTensorData<float>(output_tensor), FloatArrayNear(ref_output_data));
+  EXPECT_THAT(extractTensorShape(output_tensor), ::testing::ElementsAreArray(ref_output_shape));
+}
+
+TEST(TileTest, MultiplesShapeInvalid_NEG)
```
seanshpark, line 77: added this case for wrong shape

jinevening (approved):
LGTM Thank you

PR #11802, opened 2023-10-27
[luci-value-test] Enable Cos, Sin and Tile

seanshpark:
This will enable value test of Cos, Sin And Tile Ops.


seanshpark:
for #11760,  #11779 , from #11778

jinevening (approved):
LGTM

PR #11810, opened 2023-10-27
draft: [luci-interpreter] Support CumSum operation

batcheu:
This is a draft commit enabled CumSum Op in luci-interpreter.


batcheu:
@seanshpark Could you take a look whether the modification is enough?
I've checked the result of circle-interpreter with converted CumSum circle and origin ONNX file.
If it it enough, then I'll revise the PRs and push them with separated ones.

seanshpark:
@batcheu , please refer at #10930 for enabling new Op :)

batcheu:
> @batcheu , please refer at #10930 for enabling new Op :)

Oh, I think I misunderstood. I thought that we just modify interpreter for value test.
You want me to add the CumSum operation to ONE compiler in a complete way, right?

seanshpark:
> You want me to add the CumSum operation to ONE compiler in a complete way, right?

Yes, as there is no `CircleCumSum` IR, we have to add this IR first.
To test this IR is read/wrote correctly, there are several test modules to enable to be sure.
And also luci-interpreter is needed for our onnx2circle tool, value test module is also necessary.

batcheu:
> > You want me to add the CumSum operation to ONE compiler in a complete way, right?
> 
> Yes, as there is no `CircleCumSum` IR, we have to add this IR first. To test this IR is read/wrote correctly, there are several test modules to enable to be sure. And also luci-interpreter is needed for our onnx2circle tool, value test module is also necessary.

Okay, I'll revise the this PR first as a complete set of adding operation following steps in #10930.
Thanks :)

res/TensorFlowLiteRecipes/CumSum_000/test.recipe
```diff
@@ -0,0 +1,45 @@
+operand {
```
seanshpark, line 1: plz add `test.reverse` to both recipes and check tflchef reverse is working OK.

batcheu: Ah, I didn't check what .reverse file is for and just ignored them because all of them were empty. 
Now I found it works like as a flag to determine whether applying tflite-reverse test.
I'll add .reverse file to enable tflite-reverse test either.
FYI, I've checked it by manual `(e.g. tflchef-reverse test.tflite test.recipe)`, and it was working.

PR #11811, opened 2023-10-27
[onert-micro] Optimize While op

BalyshevArtem:
This commit adds optimization for While operation by checking condition graph and don't run it.

for issue: #11812


SlavikMIPT (changes requested):
It seems that there is an error in unit test (and kernel as well)

PR #11813, opened 2023-10-30
[tflchef] Introduce a op util function to check custom codes

Seunghui98:
This introduces a function that check codes of CustomOp to avoid duplication.

---

### DIFF
--- a/compiler/tflchef/core/src/OpUtils.cpp
+++ b/compiler/tflchef/core/src/OpUtils.cpp
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


Seunghui98:
Here is the related PR that came out first. -> https://github.com/Samsung/ONE/pull/11785
There are many changes here, so I split the PR.
After this PR is merged, the next `ModelChef.cpp` in tflchef will be reflect.

compiler/tflchef/core/src/OpUtils.cpp
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+#include "OpUtils.h"
+
+#include <stdexcept>
+
+void check_custom_op_value(const tflchef::Operation operation, std::string op_type)
+{
+  if ((operation.has_extype() && operation.extype() == "Custom") ||
+      (operation.has_type() && operation.type() == "Custom"))
```
seanshpark, line 24: As I know, `ex_type` is optional and `operation.type()` can be null, and we will remove this later.
But do we have to call `operation.has_type()` too?

Seunghui98: > But do we have to call `operation.has_type()` too?

No. Come to think of it, it doesn't seem necessary. :)
I pushed the code reflected in your comment.
Could you review the code again??

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):
LGTM

PR #11824, opened 2023-10-30
DRAFT: Common Subexpression Elimination

jinevening:
On going draft to support CSE.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to:  https://github.com/Samsung/ONE/issues/11704

Issue #11704, opened 2023-10-12
[one-optimize] Perform CSE on Quantize Op


jinevening:
## What

Let's perform CSE (common subexpression elimination) on duplicate Quantize Op.

![image](https://github.com/Samsung/ONE/assets/5449554/d9c4a5da-493f-44b1-af42-c5f0141efa82)

## Why

That pattern appears in mixed-precision quantization of layer normalization.

jinevening:
Done.

PR #11825, opened 2023-10-30
[luci] Add CommonSubExpressionEliminationPass

jinevening:
This adds a new pass CommonSubExpressionEliminationPass.

---
Related to: https://github.com/Samsung/ONE/issues/11704
Draft PR: #11824

### DIFF
--- a/compiler/luci/pass/src/CommonSubExpressionEliminationPass.cpp
+++ b/compiler/luci/pass/src/CommonSubExpressionEliminationPass.cpp
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


compiler/luci/pass/src/CommonSubExpressionEliminationPass.cpp
```diff
@@ -0,0 +1,87 @@
...
+    if (auto saved_node = cache.get(expr))
+    {
+      loco::replace(cnode).with(saved_node);
+      changed = true;
+    }
+    // Cache miss
+    else
+    {
+      cache.put(expr, cnode);
+    }
```
jinevening, line 81: This shows how CSE works.

For `node` in circle graphs,
1. Create an expression from `node` (node with its inputs).
2. If the same expression appeared before, use the expression.
3. Otherwise, save the expression.

Op-specific logic is required in step 2 to determine two expressions are the same. It is implemented in `bool operator==(const Expression &x, const Expression &y)` in ExpressionCache.cpp.

compiler/luci/pass/src/helpers/ExpressionCache.cpp
```diff
@@ -0,0 +1,171 @@
...
+  return key;
+}
+
+bool operator==(const Expression &x, const Expression &y)
+{
+  if (x.inputs != y.inputs)
+    return false;
+
+  // Check general (non-op-specific) attributes
+  if (not same_common_attributes(x.op, y.op))
```
seanshpark, line 136: I don't fully understand the logic, but what if there are two FCs that are identical but their weight values are different?

jinevening: If two FCs have different weights, then they are not equal (= not common subexpression), so CSE is not applied.

Currently this pass only supports Quantize Op. To support a new Op, we need to implement op-specific logic to check if two expressions are equal in `bool operator==(const Expression &x, const Expression &y)`.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #11654, opened 2023-10-04
[one-optimize] Optimize redundant Transposes around Reshape pattern


jinevening:
## What

Let's optimize the below pattern.
![image](https://github.com/Samsung/ONE/assets/5449554/48e7dfc6-bd5f-4abc-8a75-61859e7f8435)

The pattern can be converted to a single `Reshape`.

## Why

The above pattern appears in efficientformer.

jinevening:
[A, B, C, D] -> (pre) Transpose -> Reshape -> (post) Transpose -> [A, B x C, D]

Reshape merges multiple dimensions into one (B, C in the example). Pre/post transpose should preserve the order of the input.

stamalakhov:
@jinevening 
M.b i could try to resolve this task?

zetwhite:
@jinevening @stamalakhov 

If you don't mind, Can I try this one?
I asked about this issue to @jinevening offline, And I found that it is quite a good issue for the first try.
So, I'd like to learn the frontend part through this issue. 

( Sorry for confusion 😢 )

stamalakhov:
> Can I try this one?

@zetwhite 
Yes. Sure. Sorry for confusion. Thank you.

zetwhite:
First target<sup>[1]</sup> is done. So close this PR.

<sub> [1]  First target : I decided to target `input rank >= output rank` cases only, instead of all possible tensor.</sub>

PR #11833, opened 2023-10-31
Draft: Add Transpose-Reshape-Transpose optmiation pass

zetwhite:
to resovle : https://github.com/Samsung/ONE/issues/11654


compiler/luci/pass/src/FoldTransposesAroundReshapePass.cpp
```diff
@@ -0,0 +1,283 @@
...
+#include <luci/Profile/CircleNodeOrigin.h>
+
+#include <list>
+#include <vector>
+#include <queue>
+
+namespace
+{
+
+class PatternSimulator
```
zetwhite, line 29: I'll revisit this part!

jinevening:
Quick comments

1. We typically use `fold_XXX` for constant folding passes. Could you rename the pass to `RemoveUnnecesssaryTransposeNetPass` ?

Why?
- I think the optimization can be applied to operators other than Reshape. (ex: Tranpose -> Squeeze -> Transpose)
- There is a pass named `RemoveUnnecessaryReshapeNetPass`, which removes redundant Reshape Ops from `Reshape -> Op -> Reshape` pattern.
- This pass optimizes `Transpose -> Op -> Transpose`, so it would be better to use a similar name.

circle2circle option name would be just `remove_unnecessary_transpose` (no Net at the end).

2. Please add `luci-pass-value-test` and `circle2circle-dredd-test` after finishing implementation.

zetwhite:
> Quick comments

@jinevening 
Thank you for your comments. I applied it.

And the draft is ready.
Could you take another look? Thank you in advance.

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.test.cpp
```diff
@@ -0,0 +1,395 @@
...
+   *   A  B   C   D
+   *      |
+   * (2, 12, 12, 15)
+   *  D  A   B   C
+   *      |
+   * (2, 36, 12, 5)
+   *  D  ABCa ABCb ABCc
+   *      |
+   * (36, 12, 5, 2)
+   * ABCa ABCb ABCc D
```
zetwhite, line 242: How `tags` working, I think this example might be helpful.

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+  void simulate();
+
+  bool areTransposesUnnecessary() const;
+
+  luci::CircleReshape *createAlternative(loco::Graph *graph);
+
+private:
+  void _init_shape(const luci::CircleNode *in_tensor);
```
seanshpark, line 158: Q) ONE compiler uses `_` prefix in variables not methods. Do you have any particular reason using this?

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+  {
+    if (not mid_reshape)
+      return false;
+
+    const auto shape_node = dynamic_cast<luci::CircleConst *>(mid_reshape->shape());
+    if (shape_node == nullptr)
+      return false;
+
+    if (loco::succs(mid_reshape).size() != 1)
+      return false;
```
jinevening, line 341: # of successors check is unnecessary.

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+  }
+  const auto front_transpose = dynamic_cast<luci::CircleTranspose *>(mid_reshape->tensor());
+  {
+    if (not front_transpose)
+      return false;
+
+    if (loco::succs(front_transpose).size() != 1)
+      return false;
```
jinevening, line 349: ditto

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+  void _simulate_reshape(const luci::CircleReshape *reshape_node);
+
+  const luci::CircleTranspose *_front_transpose;
+  const luci::CircleReshape *_mid_reshape;
+  const luci::CircleTranspose *_back_transpose;
+
+  // 'tag' is used to track where each elements come from.
+  // If there are indentical tags between elements, `sub_tag` are  used to distinguish.
+  const char main_tag;
+  const char sub_tag;
```
jinevening, line 169:
```suggestion
  const char _main_tag;
  const char _sub_tag;
```

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+bool RemoveUnnecessaryTransposeNetPass::run(loco::Graph *g)
+{
+  bool changed = false;
+
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    if (auto transpose_node = dynamic_cast<luci::CircleTranspose *>(node))
+    {
+      if (remove_no_effect_reshape(transpose_node))
```
seanshpark, line 379: logic is quite long ... TLDR.
1/ please add a diagram like other pass.
2/ please a some comment explaining the logic so that others and future you can understand without difficulties.

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.test.cpp
seanshpark: Q) is there any particular reason using ABCD ?
for Tensors, we usually use HW, CHW nad NCHW and I think you already know this.

jinevening:
Please fix the build error.

You can run `./nncc configure` with `-DENABLE_STRICT_BUILD=ON` to detect such an error.

---
[2023-11-06T02:54:40.244Z] /opt/jenkins_agent/workspace/nnfw/master/pr-nncc-release@2/compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp: In member function ‘bool {anonymous}::PatternSimulator::areTransposesUnnecessary() const’:

[2023-11-06T02:54:40.244Z] /opt/jenkins_agent/workspace/nnfw/master/pr-nncc-release@2/compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp:277:22: error: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<{anonymous}::Shape::ShapeElement>::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]

[2023-11-06T02:54:40.244Z]    for (auto i = 0; i < _shape.elements.size(); i++)

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+  {
+    uint32_t value;
+    std::vector<char> tags;
+  };
+
+public:
+  void add_elem(int v)
+  {
+    ShapeElement se;
+    se.value = v;
```
seanshpark, line 42: `v` is `int` but `value` is `uint32_t`.

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
zetwhite: Thank you for reviewing :smile: 

I catch your point. 
Definity, The code is a bit long, and it is hard to catch the overall things. 
I'll try to add some comments to explain the overall logic.

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+    elements.push_back(ShapeElement(se));
+  }
+
+  void add_elem_with_tag(int v, char tag)
+  {
+    ShapeElement se;
+    se.value = v;
+    se.tags.push_back(tag);
```
seanshpark, line 50: only on `char` is added. `tags` name has `s` for plural.

seanshpark, line 50: if this is used for number, how about using `uint8_t` ?

seanshpark, line 50: Q) why not use `tag` as `int32_t` ?

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+  uint32_t _start_idx; // [start_idx, _end_idx) points Shape.elements range
+  uint32_t _end_idx;
+  uint32_t _product; // product of [se[_start_idx], se[_end_idx -1]]
+
+  std::vector<Shape::ShapeElement> &se;
+
+public:
+  SECursor(Shape &s) : _start_idx(0), _end_idx(1), se(s.elements)
+  {
+    _product = s.elements.at(0).value;
```
seanshpark, line 71: this assumes there is at least one element in `s.elements`.
I'm not sure what happens when is not not met.
maybe add some assertion?

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+  {
+    ShapeElement se;
+    se.value = v;
+    se.tags.push_back(tag);
+    elements.push_back(ShapeElement(se));
+  }
+
+  void add_elem(ShapeElement e) { elements.push_back(e); }
+
+  std::vector<ShapeElement> elements;
```
seanshpark, line 56:
```suggestion
  std::vector<ShapeElement> _elements;
```

compiler/luci/pass/src/RemoveUnnecessaryTransposeNetPass.cpp
```diff
@@ -0,0 +1,387 @@
...
+};
+
+class SECursor
+{
+private:
+  uint32_t _start_idx; // [start_idx, _end_idx) points Shape.elements range
+  uint32_t _end_idx;
+  uint32_t _product; // product of [se[_start_idx], se[_end_idx -1]]
+
+  std::vector<Shape::ShapeElement> &se;
```
seanshpark, line 66:
```suggestion
  std::vector<Shape::ShapeElement> &_se;
```

PR #11840, opened 2023-10-31
[DRAFT] Support custom GRU

BalyshevArtem:
This draft supports custom GRU, adds conversion in circle2cirlce as new pass, and adds new kernel in onert-micro


chunseoklee:
@BalyshevArtem Can we extend this to multiple GRU layer pattern like While-while-while-strideslice ?

BalyshevArtem:
> Can we extend this to multiple GRU layer pattern like While-while-while-strideslice ?
I think yes
Could you please provide tensorflow or tflite model to obtain such graph?

chunseoklee:
@BalyshevArtem Let's try to merge this PR into master. For this, we need to do : 

- [ ] extend to multiple GRU layer pattern like While-while-while-strideslice (skip if it can be merged w/o this extension)
- [ ] value check for Custom GRU op

compiler/luci/lang/include/luci/IR/Module.h
```diff
@@ -46,6 +46,8 @@ class Module final
 public:
   void add(std::unique_ptr<loco::Graph> &&g);
 
+  void remove(size_t idx);
```
seanshpark, line 49: Is it possible to make a copy of module without this `idx` subgraph and replace it?

compiler/luci/pass/include/luci/CircleOptimizer.h
```diff
@@ -34,6 +34,7 @@ class CircleOptimizer final
   {
     enum Algorithm
     {
+      FuseUnrolledGRUAsCustomGRU,
```
seanshpark, line 37: plz add to bottom not on the top

BalyshevArtem: Actually, it is first draft, I will post new one with changes only in `compiler` part.

Issue #11814, opened 2023-10-30
Compiler FE: List up unused projects


hseok-oh:
Let's list unused compiler frontend projects to remove or disable build.

seanshpark:
>  remove or disable build.

+1 to `disable build`

@lemmaa , your opinion ?

seanshpark:
I'll work on this task :)

seanshpark:
#11848 Draft implementation:
- exclude module (`compiler/something`) when there exist `exclude.me` file
- if `BUILD_WHITELIST` is given, ignore this feature, build all items in the list

@lemmaa , any comments?

seanshpark:
@lemmaa PTAL

lemmaa:
> Let's list unused compiler frontend projects to remove or disable build.

@hseok-oh ,  what's the purpose of this work? 

My guess is that there are many possible reasons for this, such as 

- Optimizing CI build resources, 
- Reducing the difficulty of source code maintenance, 
- Etc. 

In any case, the reason, purpose, etc. must be clear so that there is no confusion in the discussion and conclusion.

lemmaa:
> #11848 Draft implementation:
> 
> * exclude module (`compiler/something`) when there exist `exclude.me` file in that folder
> * if `-DBUILD_WHITELIST=...` is given, ignore this feature and build all modules in the list
> * if `-DENABLE_EXCLUDE_ME=OFF` is given, ignore this feature and build all modules
> * excluded module may fail in build in the future but we won't know that
> * we can re-enable excluded module when ever needed again
>   * any build failure should be fixed at that time
> 
> @lemmaa , any comments?

@seanshpark , I agree with your draft design. However, it would be nice to check what @hseok-oh's original intention was and organize the details.

hseok-oh:
> > Let's list unused compiler frontend projects to remove or disable build.
> 
> @hseok-oh , what's the purpose of this work?
> 
> My guess is that there are many possible reasons for this, such as
> 
>     * Optimizing CI build resources,
> 
>     * Reducing the difficulty of source code maintenance,
> 
>     * Etc.
> 
> 
> In any case, the reason, purpose, etc. must be clear so that there is no confusion in the discussion and conclusion.
> 
> And, what do you think are excluded? :)

Reason to remove: maybe not used any more
- ex) projects for caffe tools

Reason to disable build: reduce CI build resource
-  xx) `coco` and related, `ann-api` and related, etc...

lemmaa:
[...]

PR #11848, opened 2023-11-01
DRAFT CFE enable to exclude modules

seanshpark:
on-going draft to exclude modules in CFE.


seanshpark:
for #11814

seanshpark:
with `nncc-debug` build, this and PR builds after this PR is mostly about 1h30m ... 
doesn't seem to change build time ...

seanshpark:
folder count in `build/debug` with
```
find . -mindepth 1 -maxdepth 1 -type d | wc -l
```
before: `151`
after: ...

seanshpark:
done for now

PR #11867, opened 2023-11-02
[circlechef] Remove extype used for custom op

Seunghui98:
This commit includes removing extype used for custom op because it is no longer necessary.

---


Seunghui98:
The order of this PR is the step 4 [here](https://github.com/Samsung/ONE/pull/11750#discussion_r1368165801).

Issue #11842, opened 2023-11-01
Compiler FE: Enable CumSum Op


batcheu:
Lets' enable CUMSUM Op

- [x] tflchef
- [x] res/TensorFlowLiteRecipes
- [x] tfldump (nothing to do)
- [x] tflite2circle
- [x] circledump (nothing to do)
- [x] luci.lang
- [x] luci.import
- [x] luci.export 
- [x] luci.logex
- [x] luci.service
- [x] luci.partition
- [x] luci.tests
- [x] luci_interpreter
- [x] luci-value-test
- [x] circle-part-value-test (nothing to do, tested in draft PR)

seanshpark:
@batcheu , the list in the first comment is the sequence of PRs to land.
But you will have failure in `common-artifacts`.
To resolve this, first add to exclude list in `common-artifacts`,
and then land necessary codes then remove from the exclude list.

Doing this will let CI test the landed codes. (we all know that draft is working)

batcheu:
> @batcheu , the list in the first comment is the sequence of PRs to land. But you will have failure in `common-artifacts`. To resolve this, first add to exclude list in `common-artifacts`, and then land necessary codes then remove from the exclude list.
> 
> Doing this will let CI test the landed codes. (we all know that draft is working)

Okay, so let me check if I understand correctly.
First, I have to add a addtionnal commit adding CumSum in exclude.list into between updating `tlchef` commit  and adding a recipes commit.  And then in the final step, remove CumSum from exclude.list to let CI test the operation added.
Am I right?

seanshpark:
> Am I right?

I think so :)

batcheu:
Done ! ;)

PR #11871, opened 2023-11-03
[tflchef] Support CumSum operation

batcheu:
This adds support for CumSum operation in tflchef.

---
Draft PR : https://github.com/Samsung/ONE/pull/11810


hseok-oh:
@nnfw-bot test all

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

compiler/tflchef/proto/tflchef.proto
```diff
@@ -274,6 +274,11 @@ message CosOptions {
   // None
 }
 
+message CumSumOptions {
+  optional bool exclusive = 1;
+  optional bool reverse = 2;
```
shs-park, line 279: The CumSum `reverse` option seems to have a default value - false.
How about adding it? =)

```suggestion
  optional bool exclusive = 1;
  optional bool reverse = 2 [default = false];
```

Note: https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/cumsum

compiler/tflchef/tflite/src/Op/include/CumSum.h
```diff
@@ -0,0 +1,38 @@
...
+{
+public:
+  void filler(const tflite::Operator *op, TFliteImport *import,
+              tflchef::ModelRecipe *model_recipe) const override;
+  tflchef::Operation *build(const tflite::Operator *op, TFliteImport *import,
+                            tflchef::ModelRecipe *model_recipe) const override;
+};
+
+} // namespace tflchef
+#endif // __TFLITE_OP_CUMSUM_H__
```
shs-park, line 38: It's not a big deal, but...
How about adding an empty line for consistency with the other header files? =)

```suggestion
} // namespace tflchef

#endif // __TFLITE_OP_CUMSUM_H__
```

shs-park (approved):
I left a few comments.
Please take a look.

compiler/tflchef/proto/tflchef.proto
seanshpark: I also think adding default looks good :)

batcheu: Thanks, I'll apply it. 👍

compiler/tflchef/tflite/src/Op/include/CumSum.h
batcheu: ditto 👍

compiler/tflchef/proto/tflchef.proto
batcheu: FYI, the default value of boolean type in protocol buffer v2 is `false`.
https://protobuf.dev/programming-guides/proto2/#optional:~:text=For%20bools%2C%20the%20default%20value%20is%20false.
Even though, the explicit assignment of default value may be helpful for preventing human errors.

In addition to, Protocol buffer v3 doesn't allow non-zero default value 🤔.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #11873, opened 2023-11-03
[luci/service] Support CumSum operation

batcheu:
This adds support for CumSum operation in luci service.

---
Draft PR : https://github.com/Samsung/ONE/pull/11810

### DIFF
--- a/compiler/luci/service/src/CircleTypeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleTypeInferenceRule.cpp
@@ -93,6 +93,11 @@ struct TypeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::DataT
 
   loco::DataType visit(const luci::CircleCos *node) final { return luci::dtype_get(node->x()); }
 
+  loco::DataType visit(const luci::CircleCumSum *node) final
+  {
+    return luci::dtype_get(node->input());
+  }
+
   loco::DataType visit(const luci::CircleCustom *node) final
   {
     if (node->custom_code() == "BatchMatMulV2")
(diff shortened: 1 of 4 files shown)


hseok-oh:
@nnfw-bot test all

compiler/luci/service/src/CircleTypeInferenceRule.cpp
```diff
@@ -93,6 +93,8 @@ struct TypeInferenceAlgorithm final : public luci::CircleNodeVisitor<loco::DataT
 
   loco::DataType visit(const luci::CircleCos *node) final { return luci::dtype_get(node->x()); }
 
+  loco::DataType visit(const luci::CircleCumSum *node) final { return node->dtype(); }
```
mhs4670go, line 96: Maybe? Because CumSum has same dtype with its input.
```suggestion
  loco::DataType visit(const luci::CircleCumSum *node) final { return luci::dtype_get(node->input()); }
```

batcheu: Both of them have same type, but it's better to following dtype of input as you commented. :)
Thanks :smile:

seanshpark: > Both of them have same type

To be exact, `node-dtype()` is from the file dtype.
`luci::dtype_get(node->input())` is setting dtype from the input.
Logically, I think the second one is better as we're doing dtype inference.
As I understand `CumSum` output dtype follows input dtype.

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM thank you!

PR #11878, opened 2023-11-06
[common-artifacts] Exclude CumSum operation

batcheu:
This disables the test of CumSum operation until it's fully supported.

---

### DIFF
--- a/compiler/common-artifacts/exclude.lst
+++ b/compiler/common-artifacts/exclude.lst
@@ -5,6 +5,7 @@
 
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
+optimize(CumSum_000)
 
 ## CircleRecipes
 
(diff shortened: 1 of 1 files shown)


seanshpark (approved):
LGTM

compiler/common-artifacts/exclude.lst
```diff
@@ -5,6 +5,7 @@
 
 #[[ optimize : Exclude from circle optimization(circle2circle) ]]
 ## TensorFlowLiteRecipes
+optimize(CumSum_000)
```
mhs4670go, line 8: Just for note, the build will work without this line because there's no optimization pass related `CumSum` that would make it failure.

mhs4670go (approved):
LGTM

compiler/common-artifacts/exclude.lst
batcheu: Thanks for comment.
By the way, while it tests with execution of circle2circle, it complained about no implementation in luci/export regarding to CumSum. And it breaks the build. 
So, the line will be removed after luci/export commit is merged.

mhs4670go: Ah, right. Sorry. I was mistaken.

PR #11701, opened 2023-10-11
DRAFT: Support BroadcastTo

Seunghui98:
On going draft to support BroadcastTo

ONE-DCO-1.0-Signed-off-by: SeungHui Lee <shsh1004.lee@samsung.com>

---
Related Issue : https://github.com/Samsung/ONE/issues/10784

Issue #10784, opened 2023-05-15
Compier FE: Support BroadcastTo Operator


jinevening:
## What

Let's support `BroadcastTo` Operator.

TODO
- [x] tflite2circle
- [x] tfldump (nothing to do)
- [x] circledump (nothing to do)
- [x] exclude in common-artifacts
- [x] tflchef
- [x] res recipes
- [x] luci/lang
- [x] luci/import
- [x] luci/service
- [x] luci/partition
- [x] luci/logex
- [x] luci/export
- [x] luci/tests
- [x] luci-interpreter
- [x] remove exclude in common-artifacts
- [x] circle2circle
- [x] quantizer (verifier, fake quantization, const input) (nothing to do)
- [x] luci-value-py-test

seanshpark:
@Seunghui98 , for #11877 , 
Items in the first comment is order of PRs that is suggested for testing the landed codes in CI.
Please follow
I've slightly modified the list with `common-artifacts`

You can post multiple PRs if there are no dependencies.

seanshpark:
`compiler/luci/pass/src/ResolveCustomOpAddPass.cpp` is code that check `CircleCustom` with `BroadcastTo`
I think we need to revise this.

seanshpark:
When we read old circle v0.5 model files with `CircleCustom` with custom code having `BroadcastTop`, we have to convert this to `CircleBroadcastTo`.
Please consider this too.

CC @mhs4670go

mhs4670go:
After discussion on the offline, we'll introduce below things.
- New pass that processes `BroadcastTo(built-in) - Custom(AddV2)` pattern
- New pass that changes custom op into built-in op

Seunghui98:
@seanshpark 

Can I add the tflchef module and then the recipe file?
if i create a test.recipe for BroadcastTo, the build succeeds. 
but, If i create a test.reverse file together, the build fails and the following error message is generated.

```
[ 16%] Built target tflchef-file
[ 16%] Built target tflchef-reverse
[ 16%] Generating BroadcastTo_000.gen.recipe
terminate called after throwing an instance of 'std::runtime_error'
  what():  Not supported: (invalid custom)
Aborted (core dumped)
make[2]: *** [compiler/tflchef/tests/CMakeFiles/tflchef_testfiles.dir/build.make:4908: compiler/tflchef/tests/BroadcastTo_000.gen.recipe] Error 134
[...]

PR #10737, opened 2023-04-28
DRAFT: Support GeLU

jinevening:
On going draft to support GeLU.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/10692

PR #11886, opened 2023-11-06
[common-artifacts] Exclude Broadcast_To

Seunghui98:
This adds Broadcast_To_001 to exclude.lst.

---


seanshpark:
like in  #11878 , there are 3 places to exclude in common-artifacts.
have you tested with (1) with current HEAD, (2) with recipes, (3) build check ?

Seunghui98:
> have you tested with (1) with current HEAD, (2) with recipes, (3) build check ?

Yes. The build part was successful with the recipe file. 
There is no optimizer part of circle2circle, so I did not add optimizer exclude, 
should I put it in to add module function independently? :)

seanshpark:
> There is no optimizer part of circle2circle, so I did not add optimizer exclude,

OK, thanks :)

> should I put it in to add module function independently? :)

nope :)

seanshpark (approved):
LGTM

PR #11888, opened 2023-11-07
[compiler] Check exclude.me and exclude

seanshpark:
This will enable to check exclude.me file to exclude the module.


seanshpark:
for #11814 , from  #11848

hseok-oh (approved):
LGTM

PR #11898, opened 2023-11-07
[cker] Rename derivative to grad or incoming

ragmani:
This commit renames using the term "derivative" to the term "grad" or "incoming".

### DIFF
--- a/compute/cker/include/cker/train/operation/MaxPool.h
+++ b/compute/cker/include/cker/train/operation/MaxPool.h
@@ -126,22 +126,21 @@ inline void MaxPool2D(const PoolParams &params, const Shape &input_shape, const 
   out_mat.cwiseMin(params.float_activation_min).cwiseMax(params.float_activation_max);
 }
 
-inline void MaxPool2DGrad(const Shape &deriv_output_shape, const float *deriv_output_data,
-                          const int *arg_max_index, const Shape &deriv_input_shape,
-                          float *deriv_input_data)
+inline void MaxPool2DGrad(const Shape &incoming_shape, const float *incoming_data,
+                          const int *arg_max_index, const Shape &grad_shape, float *grad_data)
 {
-  assert(deriv_input_shape.DimensionsCount() == 4);
-  assert(deriv_output_shape.DimensionsCount() == 4);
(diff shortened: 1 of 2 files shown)


compute/cker/include/cker/train/operation/MaxPool.h
```diff
@@ -126,22 +126,23 @@ inline void MaxPool2D(const PoolParams &params, const Shape &input_shape, const
   out_mat.cwiseMin(params.float_activation_min).cwiseMax(params.float_activation_max);
 }
 
-inline void MaxPool2DGrad(const Shape &deriv_output_shape, const float *deriv_output_data,
-                          const int *arg_max_index, const Shape &deriv_input_shape,
-                          float *deriv_input_data)
+inline void MaxPool2DGrad(const Shape &incoming_output_shape, const float *incoming_output_data,
+                          const int *arg_max_index, const Shape &back_prop_input_shape,
+                          float *back_prop_input_data)
```
zetwhite, line 131: I re-read this part. 
I feel that the `output` and `input` is a bit hard to catch. 

  * `output-derivation` means derivation which is corresponded to my(layer's) output-tensor. 
  * `input-derivation` means derivation which is corresponded to my(layer's) input-tensor.

For `Cker` level (Not in `Layer` Level), under naming looks better.  
```suggestion
inline void MaxPool2DGrad(const Shape &incoming_shape, const float *incoming_data,
                          const int *arg_max_index, const Shape &grad_shape,
                          float *grad_data)
```

zetwhite: > In Cker level (Not in Layer Level), For clarify, under naming looks better. 

@ragmani Could you share your opinion about this?
If you agree with me, How about close this PR and Let me re-name with just `incoming` and `grad`.

ragmani: Your suggestion looks better. Thanks, I will change it.

ragmani: I renamed it to incoming or grad.

zetwhite (approved):
LGTM :+1: 
Thank you!

glistening (approved):

PR #11908, opened 2023-11-07
[circlechef] Tidy nnkit from requires.cmake

seanshpark:
This will tidy nnkit from requires.cmake.


seanshpark:
`nnkit` is not used in `circlechef`

mhs4670go (approved):
LGTM

PR #11909, opened 2023-11-07
[tflchef] Tidy nnkit from requires.cmake

seanshpark:
This will tidy nnkit from requires.cmake.


seanshpark:
`nnkit` is not used in `tflchef`
found this while doing #11814

mhs4670go (approved):
LGTM

PR #11932, opened 2023-11-08
[tflchef] Support BroadcastTo

Seunghui98:
This supports BroadcastTo Op in tflchef.

---


seanshpark (approved):
LGTM

compiler/tflchef/tflite/src/Op/BroadcastTo.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace tflchef
+{
+
+void TFliteOpBroadcastTo::filler(const tflite::Operator *op, TFliteImport *import,
+                                 tflchef::ModelRecipe *model_recipe) const
+{
+  const std::vector<int32_t> &inputs = as_index_vector(op->inputs());
+
+  bool hasShape = (inputs.size() == 2);
```
mhs4670go, line 30: It seems that BroadcastTo operator always has two inputs.

https://github.com/tensorflow/tensorflow/blob/1b8f5c396f0c016ebe81fe1af029e6f205c926a4/tensorflow/lite/kernels/broadcast_to.cc#L90

```suggestion
  assert (inputs.size() == 2);
```

Seunghui98: Thank you for the comment ! :)
I modified this part.
Could you review the code again?

seanshpark (approved):
LGTM

compiler/tflchef/tflite/src/Op/BroadcastTo.cpp
seanshpark: @Seunghui98 , if this part was copied from other Op, which one was it?

compiler/tflchef/proto/tflchef.proto
```diff
@@ -291,6 +291,10 @@ message BatchToSpaceNDOptions {
   // None
 }
 
+message BroadcastToOptions {
+  // None
```
batcheu, line 295: It seems that there's no available option for boradcast_to operation.
https://www.tensorflow.org/api_docs/python/tf/broadcast_to#args

Do we need this empty message type? If it is just for confirmation of no option, I'd like leave as comment.
Protobuf compiler will generate related code and data structure even for empty message ;(

seanshpark: > Do we need this empty message type? 

I assume design of this protobuf is to explicitly add empty for empty option.

batcheu: I could build tflchef without any error after removing the empty message.

compiler/tflchef/tflite/src/Op/BroadcastTo.cpp
mhs4670go: I heard it is from Reshape op.

mhs4670go (approved):
LGTM

compiler/tflchef/proto/tflchef.proto
batcheu: > > Do we need this empty message type?
> 
> I assume design of this protobuf is to explicitly add empty for empty option.

I didn't know about that. But I just want to warn that unmecessery code generation 🤔
However, in current .proto, it seems that there're many missing options which is not used as like `BroadcastTo`.

batcheu (approved):
LGTM

PR #11942, opened 2023-11-08
[infra] Use cmake new options

hseok-oh:
This commit updates commands to use cmake new options.
- `-S` and `-B`: source and build directoy
- `--install` with `--prefix`: install and override install prefix


infra/nnfw/command/configure
```diff
@@ -2,25 +2,5 @@
...
-  if [ ! -d "${INSTALL_PATH}" ]; then
-    echo "[ERROR] Fail to make install directory '${INSTALL_PATH}'"
-    echo "        Please make '${INSTALL_PATH}' directory first"
-    exit 255
-  fi
-fi
-
-cd "${BUILD_PATH}"
-cmake "${NNFW_PROJECT_PATH}"/infra/nnfw -DCMAKE_INSTALL_PREFIX="${INSTALL_PATH}" "$@"
+# Require cmake version >= 3.13
```
batcheu, line 5: Don't we need version check? Although, required version is very old, it is better to notify the user with specific version required rather quitting with kind of 'Unsuppoted option' error.

batcheu: It may be helpful to get the cmake version only (wo/ banner) with following command.
```
cmake -E capabilities |  grep -o '"string":"[^"]*' | grep -o '[^"]*$'
```

batcheu:
Just for curiosity, will you update nncc's either?

hseok-oh:
> Just for curiosity, will you update nncc's either?

No

1) `nncc` has differrent cmake minimum requirement
2) `nncc` does not have install command

infra/nnfw/command/configure
batcheu: Ah yes, but cmake command will fail before the cmake_minimum_requred() macro.
My worries was about lacking information of failure. If cmake version is older than required version, error message would be something like this.

```
$cmake -S ./infra/nnfw
CMake Error: Unknown argument -S
CMake Error: Run 'cmake --help' for all supported options.
```

It would be nice if it just tells checking the version of cmake.
```
$cmake -S ./infra/nnfw
[ Error ] Your cmake version (x.x.x) is too old, minimum required version is 3.16.3.
```

However, this is just my preference, optional opinion. Don't mind it ;)

hseok-oh: I tried to add error message when cmake version is old. But I decided to not add message because it is not simple to check cmake version on bash script working both on ubuntu and tizen.

glistening (approved):
LGTM

infra/nnfw/command/install
```diff
@@ -7,5 +7,5 @@ if [[ ! -d "${BUILD_PATH}" ]]; then
   exit 255
 fi
 
-cd ${BUILD_PATH}
-make install
+# Require cmake version >= 3.15
+cmake --install ${BUILD_PATH} "$@"
```
batcheu, line 11: I've tried to modify install path to %{buildroot} as we talked in offline, but it fails.
Too much files (e.g. test packages under %{test_install_path}) are installed to modified pathes.
Anyway, current modification just works without any errors. 👍

batcheu (approved):
LGTM

PR #11959, opened 2023-11-08
[tflchef] Revise check_custom_op_value parameters

seanshpark:
This will revise check_custom_op_value with reference parameters to reduce stack memory.


seanshpark:
Our internal analysis tool warns this as
> The value expected as 1st parameter 'operation' in this procedure has the size of 912 bytes. Passing structures of such size by value can lead to unnecessary performance loss.

seanshpark:
@Seunghui98 PTAL

mhs4670go (approved):
LGTM

PR #11961, opened 2023-11-08
[moco-log] Add exclude.me

seanshpark:
This will add exclude.me to exclude this module from build.


seanshpark:
for #11814 , from  #11848

jinevening (approved):
I think 1 approval is enough for this kind of PR :)

seanshpark:
> I think 1 approval is enough for this kind of PR :)

Ah, thanks :)
Change itself is simple but I want to be sure this module is checked for necessity which I could miss.

mhs4670go (approved):
LGTM

PR #11989, opened 2023-11-10
[onert] Adjust whether to cache cpu conv weights

ragmani:
This commit introduces a boolean param whether to check whether to cache weights for ConvolutionLayer of cpu backend.

### DIFF
--- a/runtime/onert/backend/cpu/KernelGenerator.cc
+++ b/runtime/onert/backend/cpu/KernelGenerator.cc
@@ -309,6 +309,9 @@ void KernelGenerator::visit(const ir::operation::Conv2D &node)
   const auto activation = node.param().activation;
   const auto param_padding = node.param().padding;
   const auto dilation = node.param().dilation;
+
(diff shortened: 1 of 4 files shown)


glistening:
@ragmani Do you mean `filter` by `weight`? or both `filter` and `bias`?

ragmani:
@glistening  `filter` means `weights`. Tensorflow conv kernel that computes gradients uses the term. Anyway this PR does not use the term `filter`.

glistening:
TensorFlow Lite uses `filter` and `bias` instead of `weight` and `bias`. I like this way because it is clear. People use `weight` as meaning of parameters.

I am still confused whether this PR would like to control cache `filter` and `bias`? Or just `filter`. From your answer, it seems to mean `filter` only because the code is copied from TensorFlow, and they use `weight` for `filter`.

glistening (approved):
LGTM

hseok-oh (approved):

PR #12012, opened 2023-11-13
Add BroadcastTo_001 

Seunghui98:
This adds a recipe file for builtin BroadcastTo.

---

### DIFF
--- a/res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
+++ b/res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
@@ -1,0 +1,29 @@
+# BroadcastTo Op supports op version 2 and 3.
+# BroadcastTo of Float type is version 2 and Quantized BroadcastTo is version 3.
+# We are using tflite kernels for float values. so, the version should be 2.
+# refer https://github.com/tensorflow/tensorflow/pull/46224/files
+operand {
+  name: "bc_input"
+  type: FLOAT32
+  shape { dim: 2 dim: 3 }
+}
+operand {
+  name: "bc_shape"
+  type: INT32
+  shape { dim: 3 }
+  filler { tag: "explicit" arg: "1" arg: "2" arg: "3" }
+}
+operand {
+  name: "bc_ofm"
+  type: FLOAT32
+  shape { dim: 1 dim: 2 dim: 3 }
+}
+operation {
+  type: "BroadcastTo"
+  input: "bc_input"
+  input: "bc_shape"
+  output: "bc_ofm"
+  version: 2
+}
+input: "bc_input"
+output: "bc_ofm"


Seunghui98:
since `version` can be both 2 and 3 ** [reference](https://github.com/tensorflow/tensorflow/blob/ce2d218fc88c20b1d81abb017eabd436779865e4/tensorflow/lite/kernels/register.cc#L306-L311), recipe files was created to test both versions.

seanshpark:
plz split for "Add" and "Change"

res/TensorFlowLiteRecipes/BroadcastTo_000/test.recipe
```diff
@@ -15,11 +15,11 @@ operand {
   shape { dim: 1 dim: 2 dim: 3 }
 }
 operation {
-  type: "Custom"
+  type: "BroadcastTo"
   input: "bc_input"
   input: "bc_shape"
   output: "bc_ofm"
-  custom_code: "BroadcastTo"
+  version: 2
```
seanshpark, line 22: Q) why is this changed?

Seunghui98: @seanshpark 
> Q) why is this changed?

`BroadcastTo` was newly introduced as a built-in op, so I modified it to fit the existing broadcastTo_000 recipe file. 
Would it be better to make `BroadcastTo_001` and `BroadcastTo_002` recipe files separately for Builtin op test use?

seanshpark: > Would it be better to make BroadcastTo_001 and BroadcastTo_002 recipe files separately for Builtin op test use?

Yes

Seunghui98: @seanshpark 

Thank you for your advice !! 
I applied comments. :)

mhs4670go (dismissed):
LGTM

res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
```diff
@@ -0,0 +1,25 @@
...
+  name: "bc_ofm"
+  type: FLOAT32
+  shape { dim: 1 dim: 2 dim: 3 }
+}
+operation {
+  type: "BroadcastTo"
+  input: "bc_input"
+  input: "bc_shape"
+  output: "bc_ofm"
+  version: 2
```
seanshpark, line 22: why is this version added?

res/TensorFlowLiteRecipes/BroadcastTo_002/test.recipe
```diff
@@ -0,0 +1,25 @@
...
+  name: "bc_ofm"
+  type: FLOAT32
+  shape { dim: 1 dim: 2 dim: 3 }
+}
+operation {
+  type: "BroadcastTo"
+  input: "bc_input"
+  input: "bc_shape"
+  output: "bc_ofm"
+  version: 3
```
seanshpark, line 22: why is this 3?

res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
Seunghui98: @seanshpark 
This is because `luci-value-test` failure occurs when the value of the `version` is 1(default value). ->[error message](https://github.com/Samsung/ONE/pull/11701#issuecomment-1767390375).


In tensorflow, the version of the `builtin` op of `broadcastTo` is shown as 2 or 3, 
so I created recipe files to test both of these. :) ** [reference](https://github.com/tensorflow/tensorflow/blob/ce2d218fc88c20b1d81abb017eabd436779865e4/tensorflow/lite/kernels/register.cc#L306-L311)

seanshpark: OK, thanks. Please add a comment that we have to use version 2 for this Op.

res/TensorFlowLiteRecipes/BroadcastTo_002/test.recipe
seanshpark: Do we have any conditional branch or test with version is 2 or 3 ?
if not, I don't think we have to add version 3.

mhs4670go: Or, we can add the check if the version is 2 or 3 in the `validate` function of `compiler/luci/import/src/Nodes/CircleBroadcastTo.cpp`.

```c++
bool CircleBroadcastToGraphBuilder::validate(const ValidateArgs &args) const
{
  auto const bt_version = args.reader.opcodes().at(args.op.opcode_index)->version();
  // The version one of broadcast to op won't be not supported
  // https://github.com/tensorflow/tensorflow/blob/ce2d218fc88c20b1d81abb017eabd436779865e4/tensorflow/lite/kernels/register.cc#L306-L311
  if (bt_version != 2 && bt_version != 3)
    return false;

  return GraphBuilder::validate(args, 2);
}
```

seanshpark: As I know, for all other Ops we didn't do any version checking: min, max values.
I don't know our IR has version... -_-;
For now, let's skip this version check in import.

mhs4670go: Okay. Then how about use version 3? I'm not sure but latest version might be good..?

seanshpark: > ... but latest version might be good..?

Found that circle IR doesn't have version thus doesn't import version per Op.
The version may effect tflite interpreter for value check.
Any version is OK to me with working value check.

mhs4670go (approved):
LGTM

res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
```diff
@@ -0,0 +1,26 @@
+# We have to use version 3 for BroadcastTo Op.
```
Seunghui98, line 1: @seanshpark 

To mark this, I wrote annotations on the recipe file. 
Should I mark it on other parts? :)

res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
```diff
@@ -0,0 +1,27 @@
+# We have to use the latest version of 3 for BroadcastTo Op.
+# https://github.com/tensorflow/tensorflow/blob/ce2d218fc88c20b1d81abb017eabd436779865e4/tensorflow/lite/kernels/register.cc#L306-L311
```
Seunghui98, line 2: @seanshpark 

> OK, thanks. Please add a comment that we have to use version 2 for this Op.

To mark this, I wrote annotations on the recipe file. 
Should I mark it on other parts? :)

seanshpark: > Should I mark it on other parts? :)

Sorry but I don't understand your question.

> We have to use the latest version of 3 for BroadcastTo Op.

Why is it so? why not 2? URL you've added doesn't seem to explain this.

My comment about "we have to use version 2" is that this Op is the first Op that requires version > 1. I wanted some explanation what you understand about this.

mhs4670go: I've found related PR on tensorflow. (https://github.com/tensorflow/tensorflow/pull/46224/files)

```cxx
// The version one of broadcast to op won't be not supported since the
    // version one was rollbacked and the builtin op code number has been
    // changed because of builtin op code shortage problem.
    // Quantized broadcast_to is version 3
    case BuiltinOperator_BROADCAST_TO:
      if (op_sig.inputs.at(0).type == kTfLiteInt8 ||
          op_sig.inputs.at(0).type == kTfLiteInt16) {
        return 3;
      }
      return 2;
```

We are using tflite kernels for float values. So, the version should be 2.

res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
```diff
@@ -0,0 +1,30 @@
+# BroadcastTo Op supports op version 2 and 3.
+# BroadcastTo of Float type is version 2 and Quantized BroadcastTo is version 3.
+# We are using tflite kernels for float values. so, the version should be 2.
+# https://github.com/tensorflow/tensorflow/blob/ce2d218fc88c20b1d81abb017eabd436779865e4/tensorflow/lite/kernels/register.cc#L306-L311
+# https://github.com/tensorflow/tensorflow/pull/46224/files
```
Seunghui98, line 5: @seanshpark 

based on the comments you gave, I commented in detail so that users can understand the version. 
Are the comments i added okay? :)

res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
```diff
@@ -0,0 +1,30 @@
+# BroadcastTo Op supports op version 2 and 3.
+# BroadcastTo of Float type is version 2 and Quantized BroadcastTo is version 3.
+# We are using tflite kernels for float values. so, the version should be 2.
+# https://github.com/tensorflow/tensorflow/blob/ce2d218fc88c20b1d81abb017eabd436779865e4/tensorflow/lite/kernels/register.cc#L306-L311
```
seanshpark, line 4: I think this can be removed. this line is too long and 46224 explains about the version.

res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
seanshpark, line 5:
```suggestion
# refer https://github.com/tensorflow/tensorflow/pull/46224/files
```

Seunghui98: Thank you a lot !!

seanshpark:
```
# BroadcastTo of Float type is version 2 and Quantized BroadcastTo is version 3.
# We are using tflite kernels for float values. so, the version should be 2.
```
looks fine. I didn't leave any comments. why was it removed?

res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
```diff
@@ -0,0 +1,28 @@
+# BroadcastTo of Float type is version 2 and Quantized BroadcastTo is version 3.
+# We are using tflite kernels for float values. so, the version should be 2.
+# refer https://github.com/tensorflow/tensorflow/pull/46224/files
```
seanshpark, line 3: this comment is for 1~3

seanshpark, line 3: this comment is for 3

seanshpark, line 3: this comment is for 3

seanshpark:
UI doesn't seem to distinguish much for comment for 1~3 vs 3 only.
there is `Comment on lines +1 to +3` but it's small and not noticeable...

res/TensorFlowLiteRecipes/BroadcastTo_001/test.recipe
Seunghui98, line 3: @seanshpark 

I misunderstood to remove code lines 1 to 4. 😅 
I modified it to 1~3 lines again, is it okay?

seanshpark: > I modified it to 1~3 lines again, is it okay?

I don't quite understand your question.

Seunghui98: @seanshpark 
You gave comments to one line, but I misunderstood to delete the 1 to 3 code lines because of UI. 😅
I modified it again to reflect the comments you gave me, is it okay?

seanshpark: Yes, looks OK. We have some trouble in communication. I hope it gets better next time.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #12024, opened 2023-11-13
[luci/service] Support BroadcastTo

Seunghui98:
This supports BroadcastTo in luci service.

---

### DIFF
--- a/compiler/luci/service/src/CircleShapeInferenceRule.cpp
+++ b/compiler/luci/service/src/CircleShapeInferenceRule.cpp
@@ -492,6 +492,36 @@ loco::NodeShape infer_batchmatmul_shape(const loco::TensorShape &x_shape,
   return loco::NodeShape{output_shape};
 }
 
+loco::NodeShape infer_broadcast_to(const luci::CircleBroadcastTo *node)
+{
+  const loco::DataType S32 = loco::DataType::S32;
+
+  loco::TensorShape shape_by_input;
+  {
+    LUCI_ASSERT(node->shape(), "2nd input shape() should not be nullptr");
+
+    // Only support node's shape() is CircleConst with S32
+    auto const_shape_node = dynamic_cast<luci::CircleConst *>(node->shape());
+    if (const_shape_node != nullptr)
+    {
+      LUCI_ASSERT(const_shape_node->dtype() == S32, "Only support int32 CircleConst");
+
+      shape_by_input.rank(const_shape_node->size<S32>());
+      for (uint32_t axis = 0; axis < shape_by_input.rank(); ++axis)
+      {
+        shape_by_input.dim(axis) = const_shape_node->at<S32>(axis);
+      }
+    }
+    else
+    {
+      // We use shape from the node itself
+      shape_by_input = own_shape(node);
+    }
+  }
+
+  return loco::NodeShape{shape_by_input};
+}
+
 loco::NodeShape infer_concatenation(const luci::CircleConcatenation *node)
 {
   // TODO Support when CircleConcatenation has 0 input
@@ -2013,6 +2043,11 @@ public:
   loco::NodeShape visit(const luci::CircleBatchToSpaceND *node) final
   {
     return infer_batch_to_space_nd(node);
+  }
+
+  loco::NodeShape visit(const luci::CircleBroadcastTo *node) final
+  {
+    return infer_broadcast_to(node);
   }
 
   loco::NodeShape visit(const luci::CircleCast *node) final { return use_x(node); }
--- a/compiler/luci/service/src/Nodes/CircleBroadcastTo.test.cpp
+++ b/compiler/luci/service/src/Nodes/CircleBroadcastTo.test.cpp
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -492,6 +492,38 @@ loco::NodeShape infer_batchmatmul_shape(const loco::TensorShape &x_shape,
...
+      shape_by_input.rank(const_shape_node->size<S32>());
+      for (uint32_t axis = 0; axis < shape_by_input.rank(); ++axis)
+      {
+        shape_by_input.dim(axis) = const_shape_node->at<S32>(axis);
+      }
+    }
+    else
+    {
+      // We use shape from the node itself
+      shape_by_input = own_shape(node);
```
mhs4670go, line 520: @seanshpark Could I have a one question about `own_shape`?

I don't remember the detail but this api has been introduced for a shape inference of OutputDummy things. I think other operators should not use this api because their shape should be inferenced from our shape inference algorithm. But, many operators' inference use this api. Is it intended this way?

seanshpark: > But, many operators' inference use this api. 

When shape inference is not possible, like there is unknown or impossible for some reason or we need to calculate from input values then currently there is no way of shape inference. for this, `own_shape()` or `use_own()` is used.

there's a comment in the method
```cxx
loco::TensorShape own_shape(const luci::CircleNode *node)
{
  loco::TensorShape shape;
  shape.rank(node->rank());
  for (uint32_t r = 0; r < node->rank(); ++r)
  {
    // Shape inference rules in this file did not consider unknown dimension.
    // If some node has unknown dimension, 0 is inserted and wrong shape
    // inference was done as a result.
    // To fix this, new shape inference algorithm is being implemented.
    // Until new inference algorithm is fully implemented, unknown dimension
    // would be represented as 1 along with TFLite expression.
    shape.dim(r) = node->dim(r).known() ? node->dim(r).value() : 1;
  }
  return shape;
}
```

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -492,6 +492,38 @@ loco::NodeShape infer_batchmatmul_shape(const loco::TensorShape &x_shape,
...
+{
+  LOGGER(l);
+
+  const loco::DataType S32 = loco::DataType::S32;
+
+  loco::TensorShape shape_by_input;
+  {
+    LUCI_ASSERT(node->shape(), "2nd input shape() should not be nullptr");
+
+    // Only support node's shape() is CircleConst with S32
```
seanshpark, line 505: what is `532` ?

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -492,6 +492,38 @@ loco::NodeShape infer_batchmatmul_shape(const loco::TensorShape &x_shape,
   return loco::NodeShape{output_shape};
 }
 
+loco::NodeShape infer_broadcast_to(const luci::CircleBroadcastTo *node)
+{
+  LOGGER(l);
```
seanshpark, line 497: where is `l` used?

compiler/luci/service/src/Nodes/CircleBroadcastTo.test.cpp
```diff
@@ -0,0 +1,33 @@
...
+  auto node_broadcastTo = g->nodes()->create<luci::CircleBroadcastTo>();
+
+  auto gc = loco::make_graph();
+  auto cloned = luci::clone_node(node_broadcastTo, gc.get());
+  ASSERT_NE(nullptr, cloned);
+  ASSERT_EQ(gc.get(), cloned->graph());
+
+  auto cloned_broadcastTo = dynamic_cast<luci::CircleBroadcastTo *>(cloned);
+  ASSERT_NE(nullptr, cloned_broadcastTo);
+}
```
seanshpark, line 33: maybe we can a `_NEG` test case where `shape()` nullptr.
`S64` type `shape()` can be enabled in the future so this won't be necessary.

compiler/luci/service/src/CircleShapeInferenceRule.cpp
Seunghui98: @seanshpark 

`infer_broadcast_to` function is based on the `infer_reshape function`. -> [code](https://github.com/Samsung/ONE/blob/d423975e86813289158c12760a879df90787ebf9/compiler/luci/service/src/CircleShapeInferenceRule.cpp#L1046)
Is it better to remove LOGGER?

compiler/luci/service/src/Nodes/CircleBroadcastTo.test.cpp
Seunghui98: @seanshpark 
Thank you for your comments ! :)

compiler/luci/service/src/CircleShapeInferenceRule.cpp
seanshpark: I think you don't know what is `LOGGER(l)`. Please follow the codes and how it's used.
And you'll get the answer.

Seunghui98: @seanshpark 
Oh, I knew I'd have to get rid of it because `LOGGER(l)` only declares here and doesn't use it from behind.
Thank you for letting me know. :)

compiler/luci/service/src/Nodes/CircleBroadcastTo.test.cpp
Seunghui98: @seanshpark 
I added a new `_NEG` test. :)
Could you review codes again??

seanshpark: My suggestion was about the code in shape inference and but the added negative test case is about cloning. 
As other test cases are not like this, and adding shape inference negative test case is another topic, and the new negative test case you've added seems strange for
```
auto cloned_broadcastTo = dynamic_cast<luci::CircleReshape *>(cloned);
```
, let's remove this new negative test case.

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

seanshpark (approved):
LGTM

PR #12031, opened 2023-11-14
[onert] Support TrainableGraph dump in dot

glistening:
It supports dot dumping for TrainableGraph during TrainingCompiler's compilation.

```
$ GRAPH_DOT_DUMP=1 \
Product/x86_64-linux.debug/out/bin/onert_train \
mnist.circle \
--load_input:raw mnist_train.i160.bin \
--load_expected:raw mnist_train.o160.bin \
--optimizer 1 \
--data_length 160 \
--epoch 1

$ ls *.dot
-rw-rw-r-- 1 gyu gyu  578 Nov 14 10:57 after_loss_insertion-0.dot
-rw-rw-r-- 1 gyu gyu 2061 Nov 14 10:52 after_lower_subg-0.dot
-rw-rw-r-- 1 gyu gyu  843 Nov 14 10:52 before_loss_insertion-0.dot
```

(I will upload the screenshot after I got to home.)

You can see the graph in dumped dot by downloading `.dot` and opening with `xdot` or something.

[training_dots.tar.gz](https://github.com/Samsung/ONE/files/13343736/training_dots.tar.gz)

### DIFF
--- a/runtime/onert/core/src/compiler/train/TrainingCompiler.cc
+++ b/runtime/onert/core/src/compiler/train/TrainingCompiler.cc
@@ -134,6 +134,17 @@ std::shared_ptr<CompilerArtifact> TrainingCompiler::compile(void)
   // operation
   _model.reset();
 
+  // TODO Handle dump level for each model
+  auto dump_level = static_cast<dumper::dot::DotDumper::Level>(_options->graph_dump_level);
+  onert::dumper::dot::DotDumper dot_dumper(dump_level);
+
+  for (const auto &pair : trainable_subgraphs)
+  {
+    const auto &subg_index = pair.first;
+    const auto &subg = pair.second;
+    dot_dumper.dump(*subg, nnfw::misc::str("before_loss_insertion-", subg_index.value()));
+  }
+
   // Apply pass for trainable subgraphs
   for (auto &&pair : trainable_subgraphs)
   {
@@ -144,6 +155,13 @@ std::shared_ptr<CompilerArtifact> TrainingCompiler::compile(void)
       .append(std::make_unique<train::pass::LossInsertionPass>(*trainable_subg, &_training_info,
                                                                subg_index))
       .run();
+  }
+
+  for (const auto &pair : trainable_subgraphs)
+  {
+    const auto &subg_index = pair.first;
+    const auto &subg = pair.second;
+    dot_dumper.dump(*subg, nnfw::misc::str("after_loss_insertion-", subg_index.value()));
   }
 
   // Change input shape according to batch_size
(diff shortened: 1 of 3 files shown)


runtime/onert/core/src/compiler/train/TrainingCompiler.cc
```diff
@@ -194,8 +208,8 @@ std::shared_ptr<CompilerArtifact> TrainingCompiler::compile(void)
   for (const auto &pair : lowered_subgs)
   {
     const auto &subg_index = pair.first;
-    const auto &lowered_subg = pair.second;
-    dot_dumper.dump(*lowered_subg, nnfw::misc::str("after_lower_subg-", subg_index.value()));
+    const auto &subg = pair.second;
+    dot_dumper.dump(*subg, nnfw::misc::str("after_lower_subg-", subg_index.value()));
```
glistening, line 212: It would be good to factor out this repeated code as a function. `TrainingCompiler.cc` is not the only place, same pattern are found in `Compiler.cc`. It is out of the scope of this PR.

jyoungyun (dismissed):
LGTM

runtime/onert/core/src/compiler/train/TrainingCompiler.cc
ragmani, line 212: `pair.second` is not a `TrainableGraph`. If you want to call `DotDumper::dump(const ir::train::TrainableGraph &graph, const std::string &tag)`, how about changing this code like below?
```suggestion
    const auto &trainable_subg = pair.second->trainable_graph();
    dot_dumper.dump(trainable_subg, nnfw::misc::str("after_lower_subg-", subg_index.value()));
```

glistening: I think it is beyond this PR. I just kept the original working. Do you want to fix this in this PR?

glistening: @ragmani I removed the change for dumping lowered graph.


I don't understand fully about original code. It is the reason I kept the original behavior. It uses `ILoweredGraph` version.

```cc
void DotDumper::dump(const compiler::ILoweredGraph &lowered_graph, const std::string &tag)
```

not
```cc
void DotDumper::dump(const ir::Graph &graph, const std::string &tag)
```

It is beyond this PR.

runtime/onert/core/src/compiler/train/TrainingCompiler.cc
```diff
@@ -134,6 +134,17 @@ std::shared_ptr<CompilerArtifact> TrainingCompiler::compile(void)
   // operation
   _model.reset();
 
+  // TODO Handle dump level for each model
+  auto dump_level = static_cast<dumper::dot::DotDumper::Level>(_options->graph_dump_level);
+  onert::dumper::dot::DotDumper dot_dumper(dump_level);
```
glistening, line 139: These lines are moved from lines 169~172.

glistening:
With this PR, you can see the following graphs.

### Before Loss Insertion


### After Loss Insertion

Something went wrong. I will update tomorrow.

### Just for your information, After lowering

ragmani (approved):
LGTM

jyoungyun (approved):
LGTM

Issue #11892, opened 2023-11-07
[onert_train] segfault during onert_train mnist model (dense,dense,softmax)


glistening:
For a model, which is basically a mnist model[^1]

[^1]: https://www.tensorflow.org/datasets/keras_example#step_2_create_and_train_the_model

```py
model = tf.keras.models.Sequential([
  tf.keras.layers.Flatten(input_shape=(28, 28)),
  tf.keras.layers.Dense(128, activation='relu'),
  tf.keras.layers.Dense(10)
])
```

but it has an additional softmax, maybe the code is like the following:

```py
model = tf.keras.models.Sequential([
  tf.keras.layers.Flatten(input_shape=(28, 28)),
  tf.keras.layers.Dense(128, activation='relu'),
  tf.keras.layers.Dense(10),
  tf.keras.layers.Softmax()
])
```

`onert_train` got segmentation fault with debug build.

```
[Switching to Thread 0x7ffe677fe640 (LWP 127076)]
0x00007ffff53d4685 in _mm_load_ps (__P=0x7ffff45ba838) at /usr/lib/gcc/x86_64-linux-gnu/9/include/xmmintrin.h:927
927	  return *(__m128 *)__P;
(gdb) bt
#0  0x00007ffff53d4685 in _mm_load_ps (__P=0x7ffff45ba838) at /usr/lib/gcc/x86_64-linux-gnu/9/include/xmmintrin.h:927
#1  Eigen::internal::pload<float __vector(4)>(Eigen::internal::unpacket_traits<float __vector(4)>::type const*) (from=0x7ffff45ba838) at /ssd/z/ONE/externals/TENSORFLOW-2.8.0-EIGEN/Eigen/src/Core/arch/SSE/PacketMath.h:714
#2  0x00007ffff5413ff6 in Eigen::internal::ploadt<float __vector(4), 16>(Eigen::internal::unpacket_traits<float __vector(4)>::type const*) (from=0x7ffff45ba838) at /ssd/z/ONE/externals/TENSORFLOW-2.8.0-EIGEN/Eigen/src/Core/GenericPacketMath.h:985
#3  Eigen::internal::ploadt_ro<float __vector(4), 16>(Eigen::internal::unpacket_traits<float __vector(4)>::type const*) (from=0x7ffff45ba838) at /ssd/z/ONE/externals/TENSORFLOW-2.8.0-EIGEN/Eigen/src/Core/GenericPacketMath.h:1009
[...]

PR #12033, opened 2023-11-14
[tools/onert_train] Remove data_length parameter from onert_train tool

jyoungyun:
This commit removes data_length input parameter. It calcaulates the data length value using the input and expected data files.

Related comment: https://github.com/Samsung/ONE/issues/11892#issuecomment-1797888717

### DIFF
--- a/tests/tools/onert_train/src/rawdataloader.cc
+++ b/tests/tools/onert_train/src/rawdataloader.cc
@@ -23,12 +23,46 @@
 
 namespace onert_train
 {
-
-Generator RawDataLoader::loadData(const std::string &input_file, const std::string &expected_file,
-                                  const std::vector<nnfw_tensorinfo> &input_infos,
-                                  const std::vector<nnfw_tensorinfo> &expected_infos,
-                                  const uint32_t data_length, const uint32_t batch_size)
+uint64_t getRawTensorSize(const std::vector<nnfw_tensorinfo> &infos)
 {
+  // NOTE The input_infos has already applied the batch_size.
+  //      In order to know the size of the tensor stored in the actual file,
+  //      the batch_size information must be excluded.
+  uint64_t total = 0;
+  for (uint32_t i = 0; i < infos.size(); ++i)
+  {
+    total += (bufsize_for(&infos[i]) / infos[i].dims[0]);
+  }
+  return total;
+}
+} // namespace onert_train
+
+namespace onert_train
+{
+std::tuple<Generator, uint32_t>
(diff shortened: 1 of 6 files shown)


tests/tools/onert_train/src/rawdataloader.cc
```diff
@@ -45,33 +79,24 @@ Generator RawDataLoader::loadData(const std::string &input_file, const std::stri
...
+      }
+      for (uint32_t i = 0; i < expected_infos.size(); ++i)
+      {
+        auto bufsz = bufsize_for(&expected_infos[i]);
+        _expected_file.seekg(expected_origins[i] + idx * bufsz, std::ios::beg);
+        _expected_file.read(reinterpret_cast<char *>(expecteds[i].data()), bufsz);
+      }
+      return true;
+    },
+    data_length);
```
glistening, line 99: I am not sure it is safe to return lambda to caller using `std::make_tuple`. `std::make_tuple` uses reference. If all variables from your lambda (e.g. captured variables of your lambda (`input_infos` and `expected_infos`)) and so on... are guaranteed to outlive than the receiver of `std::make_tuple`. It may be okay. However, it is hard to make sure manually by reading code. Static code analyzer may help.

jyoungyun: > Thus, all variables from your lambda (e.g. captured variables of your lambda (`input_infos` and `expected_infos`...) needs to be guaranteed to outlive than the receiver of `std::make_tuple`.

Yes, the captured variables is guaranteed to live outside this function because it is located in the same location as the caller function. I don't know exactly how to improve the readability of this code, but I will make some changes.

glistening: @jyoungyun I am not saying the captured variables what I've mentioned above. I already checked it. My concern is there may be more captured stuff under hood, which I am missing. If you're sure, I am okay.

glistening (approved):
LGTM

tests/tools/onert_train/src/rawdataloader.cc
ragmani, line 99: (Optional) This function is too long and has multiple roles. It would be better to have a separate function to get data length.

ragmani (approved):
LGTM

tests/tools/onert_train/src/rawdataloader.cc
jyoungyun: I considered several designs for this. But I couldn't find a better way. I will update this code to increase readability next time. Thank you for feedback. :)

PR #12036, opened 2023-11-14
[cker] Set struct PoolParams initial value

hseok-oh:
This commit defines initial value of struct PoolParams fields.

---

Draft: #12013

### DIFF
--- a/compute/cker/include/cker/Types.h
+++ b/compute/cker/include/cker/Types.h
@@ -80,17 +80,17 @@ enum class BroadcastableOpCategory : uint8_t
 
 struct PoolParams
 {
-  PaddingValues padding_values;
-  int stride_height;
-  int stride_width;
-  int filter_height;
-  int filter_width;
+  PaddingValues padding_values{0, 0};
+  int stride_height{0};
+  int stride_width{0};
+  int filter_height{0};
+  int filter_width{0};
   // uint8, etc, activation params.
-  int32_t quantized_activation_min;
-  int32_t quantized_activation_max;
+  int32_t quantized_activation_min{0};
+  int32_t quantized_activation_max{0};
   // float activation params.
-  float float_activation_min;
-  float float_activation_max;
+  float float_activation_min{0.0f};
+  float float_activation_max{0.0f};
 };
 
 struct SoftmaxParams


glistening:
@zetwhite I wondered why only `PoolParams` needs initial values. I talked with @hseok-oh, and I've found the access to uninitialized value happens during `train` backend `maxpool` layer. I am okay to fix this issue with this PR. It may be more safe. But what is the meaning of `...activation_{min,max}` are both zero.... I would like to fix the `maxpool` training code also.

compute/cker/include/cker/Types.h
```diff
@@ -91,6 +91,14 @@ struct PoolParams
...
   float float_activation_min;
   float float_activation_max;
+
+  // Default constructor with initial values
+  PoolParams()
+    : padding_values{0, 0}, stride_height(0), stride_width(0), filter_height(0), filter_width(0),
+      quantized_activation_min(0), quantized_activation_max(0), float_activation_min(0.0f),
+      float_activation_max(0.0f)
+  {
+  }
```
jyoungyun, line 101: (Q) Why do we have to set the initial value for pool params?

zetwhite:
> ... I would like to fix the `maxpool` training code also.

I'll take a look where `backend` access uninitialized param.

glistening:
There are a lot of ...Params in `cker/Types.h`. I wondered why struct `PoolParams` made static error and needs to be handled in different way.

I found the reason:

Most params (which has been used for inference) are constructed before MaxPool constructor.

<sub> backend/cpu/ops/PoolLayer.cc </sub>
```cc
#define POOLING_PARAMETERS                              \
  nnfw::cker::PoolParams op_params;                     \
  op_params.stride_height = strideHeight;               \
  op_params.stride_width = strideWidth;                 \
  op_params.filter_height = kernelHeight;               \
  op_params.filter_width = kernelWidth;                 \
  op_params.padding_values.height = (int8_t)paddingTop; \
  op_params.padding_values.width = (int8_t)paddingLeft; \
  op_params.float_activation_min = 0;                   \
  op_params.float_activation_max = 0;                   \
  op_params.quantized_activation_min = 0;               \
  op_params.quantized_activation_max = 0;

void PoolLayer::configure(const IPortableTensor *input, const uint32_t paddingLeft, const uint32_t,
                          const uint32_t paddingTop, const uint32_t, const uint32_t strideWidth,
                          const uint32_t strideHeight, const uint32_t kernelWidth,
                          const uint32_t kernelHeight, const ir::Activation activation,
                          IPortableTensor *output, const PoolType op_type)
{
  assert(input != nullptr);
  assert(output != nullptr);

  _input = input;
  _output = output;

  POOLING_PARAMETERS

  switch (_input->data_type())
  {
    case OperandType::FLOAT32:
    {
      float output_activation_min = 0;
      float output_activation_max = 0;
      CalculateActivationRange<float>(activation, &output_activation_min, &output_activation_max);
      op_params.float_activation_min = output_activation_min;
      op_params.float_activation_max = output_activation_max;

      _kernel = generateKernelGeneric<float>(op_params, op_type);
      break;
...
```

@zetwhite or @ragmani I hope to handle in consistent way. Is there any reason to make difference between `inference` and `training`?

glistening:
@hseok-oh I prefer in-class member initializer than introducing constructor. The former will keep this class as an aggregate (if we uses compiler > c++14). Also, it may be kept as a standard-layout. Also, `BinaryArithmeticOpParam` uses the in-class member initialize, it looks consistent.

compute/cker/include/cker/Types.h
ragmani: When default constructor is called, Initializing default members prevents garbage values ​​from being stored in members. So it can prevent potential bugs.

hseok-oh:
> @hseok-oh I prefer in-class member initializer than introducing constructor. The former will keep this class as an aggregate (if we uses compiler > c++14). Also, it may be kept as a standard-layout. Also, `BinaryArithmeticOpParam` uses the in-class member initialize, it looks consistent.

@glistening Then you want to change `PoolParams` struct to class, right?

compute/cker/include/cker/Types.h
jyoungyun: Ah, I wondered why PoolParam needs to be initialized here. :)

jyoungyun (approved):
LGTM

compute/cker/include/cker/Types.h
```diff
@@ -80,17 +80,17 @@ enum class BroadcastableOpCategory : uint8_t
 
 struct PoolParams
 {
-  PaddingValues padding_values;
-  int stride_height;
-  int stride_width;
-  int filter_height;
-  int filter_width;
+  PaddingValues padding_values{0, 0};
+  int stride_height{0};
```
glistening, line 84: I intended
```suggestion
  int stride_height = 0;
```
style for consistency and better readability. But, it is okay!

glistening (approved):
LGTM

PR #12037, opened 2023-11-14
[infra] Remove nncc docker-run command

hseok-oh:
This commit removes nnfw docker-run, docker-nncc and docker-shell  command. 
We can use nnas commands for these.


seanshpark (approved):
LGTM

PR #12041, opened 2023-11-14
[onert] Fix uninit member in PoolLayer for training

zetwhite:
This PR is to fix an uninitialized member-variable problem in PoolLayer. 
This is to resolve an issue found by coverity.

* background : https://github.com/Samsung/ONE/pull/12036#issuecomment-1809559677 
* related coverity report [link](http://10.113.139.115:8081/reports.htm#v13368/p10494/fileInstanceId=621180639&defectInstanceId=130733937&mergedDefectId=1734530) 

### DIFF
--- a/runtime/onert/backend/train/ops/PoolLayer.cc
+++ b/runtime/onert/backend/train/ops/PoolLayer.cc
@@ -52,8 +52,8 @@ private:
   const IPortableTensor *_output;
   nnfw::cker::PoolParams _op_params;
 
-  std::unique_ptr<Tensor> _act_back_prop_output;
-  std::unique_ptr<Tensor> _arg_max_index;
+  std::unique_ptr<Tensor> _act_back_prop_output = nullptr;
+  std::unique_ptr<Tensor> _arg_max_index = nullptr;
 
 public:
   MaxPool2D(const uint32_t paddingLeft, const uint32_t, const uint32_t, const uint32_t paddingTop,
@@ -71,6 +71,8 @@ public:
       _op_params.padding_values.width = (int8_t)paddingLeft;
       CalculateActivationRange<float>(activation, &_op_params.float_activation_min,
                                       &_op_params.float_activation_max);
+      _op_params.quantized_activation_min = 0; // not used in training
+      _op_params.quantized_activation_max = 0; // not used in training
     }
 
     _arg_max_index = std::make_unique<Tensor>(_output->get_info(), _output->layout());


runtime/onert/backend/train/ops/PoolLayer.cc
```diff
@@ -52,8 +52,8 @@ class MaxPool2D final : public TrainingKernelRegistry
   const IPortableTensor *_output;
   nnfw::cker::PoolParams _op_params;
 
-  std::unique_ptr<Tensor> _act_back_prop_output;
-  std::unique_ptr<Tensor> _arg_max_index;
+  std::unique_ptr<Tensor> _act_back_prop_output = nullptr;
+  std::unique_ptr<Tensor> _arg_max_index = nullptr;
```
zetwhite, line 56: This is not caused by Coverity.

While reading the code, I just found out it is better to assign `nullptr` explicitly.

runtime/onert/backend/train/ops/PoolLayer.cc
```diff
@@ -71,6 +71,8 @@ class MaxPool2D final : public TrainingKernelRegistry
       _op_params.padding_values.width = (int8_t)paddingLeft;
       CalculateActivationRange<float>(activation, &_op_params.float_activation_min,
                                       &_op_params.float_activation_max);
+      _op_params.quantized_activation_min = 0; // not used in training
+      _op_params.quantized_activation_max = 0; // not used in training
```
zetwhite, line 75: These lines are to resolve Coverity warning.

hseok-oh:

hseok-oh (approved):
LGTM

glistening:
@zetwhite Please find https://github.com/Samsung/ONE/pull/12036#issuecomment-1811645472.

runtime/onert/backend/train/ops/PoolLayer.cc
zetwhite: I first thought it is better to assign `nullptr` "explicitly", although `unique_ptr` has a default null value.
I thought it might help read the code, a little bit. 

On second thought, "`unique_ptr` has null value on default" is well-known.
So, Without explicit assigning also looks good for me.

Better to remove explicit assigning to focus on the uinitialization-issue.

jyoungyun: (suggestion)
How about using the initializer list for consistency with other class member?

runtime/onert/backend/train/ops/PoolLayer.cc
```diff
@@ -52,8 +52,8 @@ class MaxPool2D final : public TrainingKernelRegistry
   const IPortableTensor *_output;
   nnfw::cker::PoolParams _op_params;
```
glistening, line 53:
```suggestion
  nnfw::cker::PoolParams _op_params{};
```

If several suggestions requires much changes, and you're busy now, this may be a simple solution. It will initialize `PoolParams` with zero.

zetwhite:
I talked about this with @glistening and here is the conclusion : 

* don't fix the poollayer implementation now.
   * move `PoolParams` to `MaxPool2D::TrainingKernelRegistry` to `train::ops::PoolLayer` class is helpful (for later, to introducing other Pool OP - AveragePool, MinPool ..etc)
   * But it isn't necessary right now. (So I will handle it later)

* Initialization will be handled in the `PoolParams` structure (fixed by this PR : https://github.com/Samsung/ONE/pull/12036) 

So, close this PR.

runtime/onert/backend/train/ops/PoolLayer.cc
zetwhite: > How about using the initializer list for consistency with other class member? 

Thanks for the suggestion! 

But I changed the mind and decided not to handle the un-initialization issue in PoolLayer. 
  * Because Initialization in PoolParams(like https://github.com/Samsung/ONE/pull/12036) looks enough. So Duplicated-initialization here is a bit redundant. 

So, I closed this PR. Thanks for reviewing.

Issue #12020, opened 2023-11-13
[circle-mpqsolver] Dump qconfig file in PatternSolver


jinevening:
## What

Let's dump qconfig file after PatternSolver runs.

## Why

Users can make their own qconfig files applied with pre-defined patterns.

## How

I think it would be ok to dump qconfig file if `--save_intermediate` option is turned on, which is similar to bisection solver.

stamalakhov:
@jinevening 
M.b. i can try to resolve this one?

seanshpark:
seems done

PR #12042, opened 2023-11-14
[circle-mpqsolver] Dump config

stamalakhov:
This commit dumps final qconfig if save_intermediate option is on.

Running `patternsolver` with `--save_intermediate` option turned to `on`, successfully produced `FinalConfiguration.mpq.json` with correct parameters. 
Please see attachement for result:
[out_3.zip](https://github.com/Samsung/ONE/files/13349884/out_3.zip)

Related: #12020

### DIFF
--- a/compiler/circle-mpqsolver/src/core/DumpingHooks.h
+++ b/compiler/circle-mpqsolver/src/core/DumpingHooks.h
@@ -48,7 +48,7 @@ public:
   virtual void onQuantized(luci::Module *module) const override;
 
   /**
-   * @brief called on the start of iterative search
+   * @brief called on the start of mpq search
    */
   virtual void onBeginSolver(const std::string &model_path, float q8error, float q16error) override;
 
@@ -64,7 +64,7 @@ public:
                               float error) override;
 
   /**
-   * @brief called at the end of iterative search
+   * @brief called at the end of mpq search
    */
   virtual void onEndSolver(const LayerParams &layers, const std::string &def_dtype,
                            float qerror) override;


stamalakhov:
@jinevening 
Bisection currently doesn't use quantization `context`, so i've updated `Dumper` only for `PatternSolver` (by hardcoding `granularity` to `channel` in `BisectionSolver`).

compiler/circle-mpqsolver/src/core/SolverHooks.h
```diff
@@ -61,6 +61,15 @@ class SolverHooks
    */
   virtual void on_end_solver(const LayerParams &layers, const std::string &def_dtype,
                              float qerror) = 0;
```
jinevening, line 63: I think this should be replaced with new `on_end_solver`?

stamalakhov: > I think this should be replaced with new `on_end_solver`? Or at least TODO comment needs to be inserted.

ok.

stamalakhov: I've posted comment for now.

stamalakhov: @jinevening 
Please ignore comment above. I've merged these functions into single one
```
 virtual void on_end_solver(const LayerParams &layers, const std::string &def_dtype,
                             const std::string *def_granularity, const float *qerror);
```
so that unapplicable parameters are passed as nullptr.
Please let me know if this design is not satisfactory.

stamalakhov: @jinevening 
Can i try to split into mergeable PRs?

stamalakhov: @jinevening 
Please take another look.

jinevening: Sorry for late reply. I forgot this issue for a while (maybe missed when I came back from leave).

> Please let me know if this design is not satisfactory.
Or it's better to introduce some structure like Quantizer Context or don't use pointers (so that empty string and -1 error will mean unapplicability)?

Making a struct looks good to me. BTW, is it necessary to use hard-coded values like "channel" or "uint8"? Now those information is saved in Quantizer::Context.

stamalakhov: > Making a struct looks good to me. BTW, is it necessary to use hard-coded values like "channel" or "uint8"? Now those information is saved in Quantizer::Context.

Ahhh. I'll try to remove hard-coded values. Thank you.

stamalakhov: @jinevening 
done

compiler/circle-mpqsolver/src/core/Dumper.cpp
```diff
@@ -80,18 +80,19 @@ void Dumper::prepareDirectory(const std::string &dir_path) const
 }
 
 void Dumper::dumpMPQConfiguration(const LayerParams &layers, const std::string &def_dtype,
-                                  int step) const
+                                  const std::string &default_granularity, int step) const
```
jinevening, line 83:
```suggestion
                                  const std::string &def_granularity, int step) const
```

compiler/circle-mpqsolver/src/core/DumpingHooks.h
```diff
@@ -68,14 +68,15 @@ class DumpingHooks final : public QuantizerHook, public SolverHooks
    * @brief called at the end of iterative search
    */
   virtual void on_end_solver(const LayerParams &layers, const std::string &def_dtype,
-                             float qerror) override;
+                             const std::string *def_granularity, const float *qerror) override;
```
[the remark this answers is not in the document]
stamalakhov: > If _ctx is given, def_dtype and def_granularity would be unnecessary arguments?

@jinevening 
For `bisection` def_dtype is needed, because we can quantize to `int16` the whole network.
As for `def_granularity` - you are right, we don't need. I'll remove it.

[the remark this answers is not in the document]
stamalakhov: @jinevening 
done.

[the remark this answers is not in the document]
stamalakhov: @jinevening 
Please take another look. `BisectionSolver` still uses explicit "uint8"/"int16". Do i have to change `BisectionSolver`, to use only context?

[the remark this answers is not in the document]
jinevening: Looks okay to me. Let's proceed.

stamalakhov:
AS all PR's are merged, this one can be closed.

Issue #11692, opened 2023-10-10
[onert] Implement circle+(training) parameter 


zetwhite:
Let's implement a training param importer that gets hyper-parameters from the circle+ file. 

### notes 
* For now, `onert_train` gets hyper-parameters from command line options.

zetwhite:
(Progress Updates) 

Now, the draft (https://github.com/Samsung/ONE/pull/11740) works well with this file. 

* model_file : [mnist_with_meta_circle.zip](https://github.com/Samsung/ONE/files/13112257/mnist_with_meta_circle.zip)
   * epoch : 2, batch : 64, optimizer : sgd, learning_rate : 0.01 
   * have only FC and Relu operation 

```
➜ ~/Workspace/ONE/Product/out/bin/onert_train --modelfile mnist_with_meta.circle --load_input:raw pool_mnist_model/mnist_x_train.1000.bin --load_expected:raw pool_mnist_model/mnist_y_train.1000.bin --data_length 1000 -v 100                                


Model Expected Filename pool_mnist_model/mnist_y_train.1000.bin
Model Input Filename pool_mnist_model/mnist_x_train.1000.bin
Model Filename mnist_with_meta.circle
Epoch 1/2 - 50.340ms/step - loss: [0] 0.2309
Epoch 2/2 - 50.130ms/step - loss: [0] 0.2185
===================================
MODEL_LOAD   takes 1.486 ms
PREPARE      takes 8.617 ms
EXECUTE      takes 1508.700 ms
- MEAN     :  1508.700 ms
- MAX      :  1508.700 ms
- MIN      :  1508.700 ms
- GEOMEAN  :  1508.700 ms
```

zetwhite:
* (additionally) Yesterday, I asked @jyoungyun for advice on the current implementation details. Since she gave me a much better idea, I have a plan to fix a current implementation.

zetwhite:
### (Progress Updated) 

I cleaned the code from previous draft and upload it as version-2 ( https://github.com/Samsung/ONE/pull/12045 ).
You can check it is working well with the file : [mnist.zip](https://github.com/Samsung/ONE/files/13371786/mnist.zip) 
```
➜ ~/Workspace/ONE/Product/out/bin/onert_train --modelfile mnist.circle+ --load_input:raw pool_mnist_model/mnist_x_train.1000.bin --load_expected:raw pool_mnist_model/mnist_y_train.1000.bin --data_length 1000 -v 100
[...]

PR #12045, opened 2023-11-16
Draft-v2: Implement Circle+ Loader

zetwhite:


zetwhite:
#### notes 
This draft 
   * adds an API that fills `nnfw_train_info` which resides in `onert_train` from the model-file. 
   * the API create `ir::TrainingInfo` and fills `nnfw_train_info` 
image

infra/nnfw/cmake/CfgOptionFlags.cmake
```diff
@@ -23,6 +23,7 @@ option(BUILD_ONERT_RUN "Build onert_run" ON)
 option(BUILD_ONERT_TRAIN "Build onert_train" ON)
 option(BUILD_TFLITE_LOADER "Build TensorFlow Lite loader" ON)
 option(BUILD_CIRCLE_LOADER "Build circle loader" ON)
+option(BUILD_TRAININFO_LOADER "Build circle traininfo loader" ON)
```
jyoungyun, line 26: How about using `circlr+` word instead of `traininfo` ? Because this pr is for loading circle+ file format. :)

runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -203,6 +203,22 @@ typedef struct nnfw_train_info
   NNFW_TRAIN_OPTIMIZER opt = NNFW_TRAIN_OPTIMIZER_SGD;
 } nnfw_train_info;
 
+constexpr char traininfo_metadata_name[] = "CIRCLE_TRAINING";
+
```
jyoungyun, line 207: This constexpr is intended to implement a circle+ loader. It is not for the users. Then how about moving this expression into the source file?

runtime/onert/frontend/circle/include/circle_loader.h
```diff
@@ -25,7 +25,7 @@ namespace onert
 {
 namespace circle_loader
```
jyoungyun, line 26: You can use different namespace for circle+ format.

tests/tools/onert_train/src/onert_train.cc
```diff
@@ -59,19 +59,96 @@ int main(const int argc, char **argv)
...
+      else if (args.useSingleModel())
+        NNPR_ENSURE_STATUS(nnfw_load_model_from_file(session, model_file.c_str()));
       else
-        NNPR_ENSURE_STATUS(nnfw_load_model_from_file(session, args.getPackageFilename().c_str()));
+        NNPR_ENSURE_STATUS(nnfw_load_model_from_file(session, model_file.c_str()));
     });
 
+    // If failed to load train_info from model file,
+    // or train_info is given through arugment
+    if (tinfo_loaded == false || args.isTrainingInfo())
```
jyoungyun, line 83: This `tinfo_loaded` is always `false` value. Is it your intention?

tests/tools/onert_train/src/onert_train.cc
```diff
@@ -119,42 +196,6 @@ int main(const int argc, char **argv)
...
-    // prepare training info
-    nnfw_train_info tri;
-    tri.batch_size = args.getBatchSize();
-    tri.learning_rate = args.getLearningRate();
-    tri.loss = convertLossType(args.getLossType());
-    tri.opt = convertOptType(args.getOptimizerType());
-
-    // prepare execution
-
-    // TODO When nnfw_{prepare|run} are failed, can't catch the time
```
jyoungyun, line 157: Please keep this comment. :)

tests/tools/onert_train/src/onert_train.cc
```diff
@@ -59,19 +59,96 @@ int main(const int argc, char **argv)
...
+        switch (opt)
+        {
+          case NNFW_TRAIN_OPTIMIZER_ADAM:
+            return "adam";
+          case NNFW_TRAIN_OPTIMIZER_SGD:
+            return "sgd";
+          default:
+            return "unkwown";
+        }
+      };
```
jyoungyun, line 141: (suggestion) It would be better to implement it as a separate function. :)

tests/tools/onert_train/src/args.cc
```diff
@@ -274,6 +274,17 @@ void Args::Parse(const int argc, char **argv)
...
+    if (!vm.count("data_length"))
+      throw boost::program_options::error(std::string("data_length option is mandatory."));
+
+    auto is_given = [&vm](const std::string &s) -> bool { return not vm[s].defaulted(); };
+
+    if (is_given("batch_size") && is_given("learning_rate") && is_given("loss") &&
+        is_given("optimizer"))
+    {
+      _is_training_info = true;
+    }
```
jyoungyun, line 287: It would be better to check if there is a given parameter in main function instead of using `_is_training_info` variable.

glistening:
@zetwhite Here are some points what I've talked you offline:

1. I would like to remove training info's default values from several difference places.
(e.g. flatbuffers, command-line, API structure default, ...)
It would be good to keep just 1 place (which is reasonable) or no place).
For example, I think command-line tool (e.g. `onert-train` specifies training info to **override** default value, it may be good to remove all default values from command-line tool.

2. Training information timing needs to be determined.
While the model itself is loaded during `nnfw_load_...`, training info was constructed in `nnfw_train_prepare`. According to this decision, the implementation will be changed.

3. Where to put circle_traininginfo.fbs and its generated file.
I would like to create `circlex_schema` (or `circle+_schema` in the same level of `circle_schema`.).

infra/nnfw/cmake/CfgOptionFlags.cmake
zetwhite: I intentionally choose the word `traininfo` instead of `circle+`.

Because `circle+` range is a bit ambiguous. 
Which does `circle+` range for?
   (1) both `circle format` and `train info`
   (2) only `train info`
 
I thought `circle+` means (1). 
But this loader mainly focuses on loading `train info` from metadata. 
So, That's why I choose `traininfo loader`.

jyoungyun: I thought circle+ contains all meta information included in circle+ file format.
This circle+ loader loads all meta information from circle+ file. And each function can select the meta data it is interested in.

runtime/onert/api/include/nnfw_experimental.h
zetwhite: Thanks for pointing out.
Updated!

tests/tools/onert_train/src/onert_train.cc
zetwhite: Ah, that was my mistake.
updated!

tests/tools/onert_train/src/onert_train.cc
zetwhite: updated!

runtime/onert/frontend/circle/src/circle_loader.cc
```diff
@@ -228,11 +228,11 @@ void CircleLoader::loadBCQFullyConnected(const Operator *op, ir::Graph &subg)
 
 } // namespace
 
-std::unique_ptr<ir::Model> loadModel(const std::string &filename)
+std::unique_ptr<ir::Model> loadModel(const std::string &filename, bool load_metadata)
 {
   auto model = std::make_unique<ir::Model>();
   CircleLoader loader(model);
-  loader.loadFromFile(filename);
+  loader.loadFromFile(filename, load_metadata);
```
zetwhite, line 235: (notes) I modified `circle_loader::loadModel` a bit to load all metadata into `ir::Model`.

runtime/onert/core/include/ir/Model.h
```diff
@@ -175,8 +175,26 @@ class Model
...
+  }
+
+  std::shared_ptr<const ir::Data> get_metadata(const std::string name) const
+  {
+    return _metadatas.at(name);
+  }
+
+private:
+  std::unordered_map<std::string, std::shared_ptr<const ir::Data>> _metadatas;
+};
```
zetwhite, line 197: * `ir::Model` has a new field name `_metadata` to save (metadata-name, data) pairs.
* `_metadata` field is filled while loading circle_model.

infra/nnfw/cmake/CfgOptionFlags.cmake
zetwhite: Oh, I see your points.

But still for me, the words "circle+ loader" look like it is responsible for loading both circle(subgraphs and tensors) and its metadata(train info and .. etc).

Therefore, How about the word `circletr_meta_loader` ? 
+ ) I also thought the word `circle+_meta_loader`, but I'm afraid that string "+" cause trouble later.

zetwhite:
> @zetwhite Here are some points what I've talked you offline: 

Thanks for the summary.

I found out how-to-append the loader on the current implementation wasn't clear. :sweat_smile:  
I'd better clarify the points and ask circle+'s member opinions offline.

tests/tools/onert_train/src/args.cc
zetwhite: I agree.

But for now, It is hard to check in the main. 
Because `Arg` has a default value, It is hard to figure out whether the value is given or default in main. 

https://github.com/Samsung/ONE/blob/8243c77d683af17786e3ae6e752dd28ffb98bd41/tests/tools/onert_train/src/args.cc#L212-L222

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -189,6 +191,8 @@ std::unique_ptr<onert::ir::Model> loadModel(const std::string filename,
     return onert::tflite_loader::loadModel(filename.c_str());
   if (model_type == "circle")
     return onert::circle_loader::loadModel(filename.c_str());
+  if (model_type == "circle+")
+    return onert::circle_loader::loadModel(filename.c_str(), /*load metadata*/ true);
```
zetwhite, line 195: (notes)

runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "ir/train/TrainingInfo.h"
+#include "ir/Model.h"
+
+namespace onert
+{
+namespace traininfo_loader
+{
+
+std::unique_ptr<ir::train::TrainingInfo> loadTrainingInfo(std::shared_ptr<const ir::Model> model);
```
zetwhite, line 28: (notes)

PR #12047, opened 2023-11-16
[onert] Introduce LossMeanSquaredErrorLayer in train backend

jyoungyun:
This commit introduces LossMeanSqauredErrorLayer in train backend. This code is from LossLayer. LossLayer becomes a base class and other loss types inherit LossLayer to implement each loss layer.

### DIFF
--- a/runtime/onert/backend/train/KernelGenerator.cc
+++ b/runtime/onert/backend/train/KernelGenerator.cc
@@ -19,7 +19,7 @@
 #include "ops/ConvolutionLayer.h"
 #include "ops/ElementwiseActivationLayer.h"
 #include "ops/FullyConnectedLayer.h"
-#include "ops/LossLayer.h"
+#include "ops/LossMeanSquaredErrorLayer.h"
 #include "ops/GradientApplier.h"
 #include "ops/PoolLayer.h"
 #include "ops/ReshapeLayer.h"
@@ -50,17 +50,6 @@ convertElementwiseActivationType(ir::operation::ElementwiseActivation::Type type
   {
     case ir::operation::ElementwiseActivation::Type::RELU:
       return ops::ElementwiseActivationType::kReLU;
-    default:
-      throw std::runtime_error("train KernelGenerator : Not supported operation yet");
-  }
-}
-
-ops::LossType convertLossType(ir::operation::Loss::Type type_ir)
-{
-  switch (type_ir)
-  {
-    case ir::operation::Loss::Type::MEAN_SQUARED_ERROR:
-      return ops::LossType::kMSE;
     default:
       throw std::runtime_error("train KernelGenerator : Not supported operation yet");
   }
@@ -252,14 +241,23 @@ void KernelGenerator::visit(const ir::train::operation::Loss &node)
   auto y_true_tensor = _tensor_reg->getPortableTensor(y_true_index);
 
   auto back_prop_y_pred_tensor = _tensor_reg->getBackPropTensor(y_pred_index);
-  auto fn = std::make_unique<ops::LossLayer>();
 
-  fn->configure(y_pred_tensor, y_true_tensor, output_tensor, back_prop_y_pred_tensor,
-                convertLossType(node.param().op_type));
+  auto op_type = node.param().op_type;
 
-  _return_fn = std::move(fn);
-
-  UNUSED_RELEASE(convertPoolType);
+  switch (op_type)
+  {
+    case ir::operation::Loss::Type::MEAN_SQUARED_ERROR:
+    {
+      auto fn = std::make_unique<ops::LossMeanSquaredErrorLayer>();
+      fn->configure(y_pred_tensor, y_true_tensor, output_tensor, back_prop_y_pred_tensor);
+      _return_fn = std::move(fn);
+      break;
+    }
+    case ir::operation::Loss::Type::CATEGORICAL_CROSSENTROPY:
+    default:
+      throw std::runtime_error("LossLayer: unsupported loss type");
+      break;
+  }
 }
 
 void KernelGenerator::visit(const ir::train::operation::Pool2D &node)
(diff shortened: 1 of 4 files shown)


jyoungyun:
This PR helps to introduce LossCategoricalCrossentropyLayer easily. :)

runtime/onert/backend/train/KernelGenerator.cc
```diff
@@ -252,14 +241,22 @@ void KernelGenerator::visit(const ir::train::operation::Loss &node)
...
-                convertLossType(node.param().op_type));
+  auto op_type = node.param().op_type;
 
-  _return_fn = std::move(fn);
+  std::unique_ptr<ops::LossLayer> fn;
+  switch (op_type)
+  {
+    case ir::operation::Loss::Type::MEAN_SQUARED_ERROR:
+      fn = std::make_unique<ops::LossMeanSquaredErrorLayer>();
+      dynamic_cast<ops::LossMeanSquaredErrorLayer *>(fn.get())->configure(
```
glistening, line 252: @jyoungyun `dynamic_cast` has performance issue. I prefer not to use `dynamic_cast` except that it is not avoidable. I don't think much, but is it possible to remove `dynamic_cast` here?

ragmani: I agree with @glistening . And there is no reason to use `dynamic_cast`.

runtime/onert/backend/train/KernelGenerator.cc
```diff
@@ -252,14 +241,22 @@ void KernelGenerator::visit(const ir::train::operation::Loss &node)
...
+        y_pred_tensor, y_true_tensor, output_tensor, back_prop_y_pred_tensor);
+      break;
+    default:
+      throw std::runtime_error("LossLayer: unsupported loss type");
+  }
```
ragmani, line 257: How about the below code?
```suggestion
  switch (op_type)
  {
    case ir::operation::Loss::Type::MEAN_SQUARED_ERROR:
      auto fn = std::make_unique<ops::LossMeanSquaredErrorLayer>();
      fn->configure(
        y_pred_tensor, y_true_tensor, output_tensor, back_prop_y_pred_tensor);
      _return_fn = std::move(fn); 
      break;
    default:
      throw std::runtime_error("LossLayer: unsupported loss type");
  }
```

runtime/onert/backend/train/KernelGenerator.cc
```diff
@@ -252,14 +241,22 @@ void KernelGenerator::visit(const ir::train::operation::Loss &node)
...
-                convertLossType(node.param().op_type));
+  auto op_type = node.param().op_type;
 
-  _return_fn = std::move(fn);
+  std::unique_ptr<ops::LossLayer> fn;
+  switch (op_type)
+  {
+    case ir::operation::Loss::Type::MEAN_SQUARED_ERROR:
+      fn = std::make_unique<ops::LossMeanSquaredErrorLayer>();
+      static_cast<ops::LossMeanSquaredErrorLayer *>(fn.get())->configure(
```
ragmani, line 252: `static_cast` for performing downcast does not perform any safety checks. If you want to keep using this downcasting way you can use `nnfw::misc::polymorphic_downcast` instead of `static_cast`.

And `unique_ptr` can be used just like a pointer. Calling `get()` seems to be wasteful.
```suggestion
      nnfw::misc::polymorphic_downcast<ops::LossMeanSquaredErrorLayer *>(fn)->configure(
```

ragmani, line 252: Alternatively, you can pass those arguments as constructor parameters without having to define and call `configure` like below.
```suggestion
      fn = std::make_unique<ops::LossMeanSquaredErrorLayer>(
```

runtime/onert/backend/train/KernelGenerator.cc
```diff
@@ -252,14 +241,22 @@ void KernelGenerator::visit(const ir::train::operation::Loss &node)
...
+        y_pred_tensor, y_true_tensor, output_tensor, back_prop_y_pred_tensor);
```
ragmani, line 253: Alternatively, you can pass those arguments as constructor parameters without having to define and call `configure` like below.
```suggestion
      fn = std::make_unique<ops::LossMeanSquaredErrorLayer>(
        y_pred_tensor, y_true_tensor, output_tensor, back_prop_y_pred_tensor);
```

runtime/onert/backend/train/KernelGenerator.cc
jyoungyun: Thank you for suggesting several ways for this pr. :) This suggestion is most explicitly implemented and I prefer this way.

ragmani (approved):
LGTM

runtime/onert/backend/train/ops/LossMeanSquaredErrorLayer.cc
```diff
@@ -0,0 +1,69 @@
...
+void LossMeanSquaredErrorLayer::configure(const IPortableTensor *y_pred,
+                                          const IPortableTensor *y_true, IPortableTensor *output,
+                                          IPortableTensor *back_prop_y_pred)
+{
+  LossLayer::configure(y_pred, y_true, output, back_prop_y_pred);
+}
+
+void LossMeanSquaredErrorLayer::forward(bool)
+{
+  // TODO Implement this
```
glistening, line 40: @jyoungyun What does "Implement this" mean? FLOT32 case seems implemented. Maybe, is this comment for else case? If so, throw seems enough.

jyoungyun: Ｙｅｓ，　ｔｈｉｓ　ｆｕｎｃｔｉｏｎ　ｎｅｅｄｓ　ｔｏ　ｂｅ　ｉｍｐｌｅｍｅｎｔｅｄ　ｆｏｒ　ｄｉｆｆｅｒｅｎｔ　ｄａｔａ　ｔｙｐｅｓ．

jyoungyun: Yes, this function needs to be implemented for different data types.

glistening: I considered this comment was there before you implement `FLAT32`, and left by mistake. Because as I wrote in the above, 

>    throw std::runtime_error("LossMeanSquaredErrorLayer: unsupported data type");

already speaks the same thing.

glistening (approved):
LGTM

PR #12058, opened 2023-11-17
[luci-interpreter] Support BroadcastTo

Seunghui98:
This supports BroadcastTo in luci-interpreter.

---


compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <stdexcept>
+#include <iostream>
+
+namespace luci_interpreter
+{
+
+namespace kernels
+{
+
+static Shape extractShapeFromTensor(const Tensor *tensor)
```
seanshpark, line 31:
```suggestion
namespace
{

Shape extractShapeFromTensor(const Tensor *tensor)
```

compiler/luci-interpreter/pal/linux/pal.cmake
```diff
@@ -68,7 +68,9 @@ macro(add_pal_to_target TGT)
     # instead add sources with visitors in this library
     set(PAL_SOURCES ${TensorFlowSource_DIR}/tensorflow/lite/kernels/internal/tensor_utils.cc
             ${TensorFlowSource_DIR}/tensorflow/lite/kernels/internal/reference/portable_tensor_utils.cc
-            ${TensorFlowSource_DIR}/tensorflow/lite/kernels/internal/quantization_util.cc)
+            ${TensorFlowSource_DIR}/tensorflow/lite/kernels/internal/quantization_util.cc
+            ${TensorFlowSource_DIR}/tensorflow/lite/kernels/kernel_util.cc
+            ${TensorFlowSource_DIR}/tensorflow/lite/c/common.c)
```
seanshpark, line 73: Q) why are these added?

compiler/luci-interpreter/src/kernels/BroadcastTo.test.cpp
```diff
@@ -0,0 +1,92 @@
...
+                 {1, 2, 3},
+                 /*shape_data*/
+                 {2, 3},
+                 /*output_data*/
+                 {
+                   1, 2, 3, // Row 1
+                   1, 2, 3, // Row 2
+                 });
+  SUCCEED();
+}
```
seanshpark, line 87: can you please add some negative tests?

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+  Shape shape(tensor->shape().num_elements());
+  if (tensor->element_type() == DataType::S32)
+  {
+    const auto *shape_data = tensor->data<int32_t>();
+    for (int i = 0; i < tensor->shape().num_elements(); ++i)
+    {
+      shape.dim(i) = shape_data[i];
```
seanshpark, line 39: Q) is there no cases where shape_data[i] < 0 ?

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+  else if (tensor->element_type() == DataType::S64)
+  {
+    const auto *shape_data = tensor->data<int64_t>();
+    for (int i = 0; i < tensor->shape().num_elements(); ++i)
+    {
+      shape.dim(i) = static_cast<int32_t>(shape_data[i]);
```
jinevening, line 47: Can you check value overflow?

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
jinevening, line 31:
```suggestion
// TODO Extract this function to Utils.h
static Shape extractShapeFromTensor(const Tensor *tensor)
```

compiler/luci-interpreter/pal/linux/PALBroadcastTo.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include <tensorflow/lite/kernels/internal/reference/broadcast_to.h>
+
+namespace luci_interpreter_pal
+{
+static inline void BroadcastTo(const tflite::RuntimeShape &input_shape, const char *input_data,
+                               const tflite::RuntimeShape &output_shape, char *output_data,
+                               TfLiteType data_type)
+{
+  constexpr int kMaxDims = 5;
```
jinevening, line 28: Can you add a comment about the value? Why `kMaxDims` is 5?

compiler/luci-interpreter/src/kernels/BroadcastTo.test.cpp
```diff
@@ -0,0 +1,92 @@
...
+  kernel.execute();
+
+  EXPECT_THAT(extractTensorData<float>(output_tensor), FloatArrayNear(output_data));
+}
+
+class BroadcastToTest : public ::testing::Test
+{
+};
+
+TEST_F(BroadcastToTest, Simple)
```
jinevening, line 59:
```suggestion
TEST_F(BroadcastToTest, SimpleS32)
```

compiler/luci-interpreter/src/kernels/BroadcastTo.test.cpp
```diff
@@ -0,0 +1,92 @@
...
+                 {2, 3},
+                 /*output_data*/
+                 {
+                   1, 2, 3, // Row 1
+                   1, 2, 3, // Row 2
+                 });
+  SUCCEED();
+}
+
+TEST_F(BroadcastToTest, SupportS64)
```
jinevening, line 74:
```suggestion
TEST_F(BroadcastToTest, SimpleS64)
```

compiler/luci-interpreter/src/kernels/BroadcastTo.test.cpp
```diff
@@ -0,0 +1,92 @@
...
+                 /*shape_data*/
+                 {2, 3},
+                 /*output_data*/
+                 {
+                   1, 2, 3, // Row 1
+                   1, 2, 3, // Row 2
+                 });
+  SUCCEED();
+}
+
```
jinevening, line 88: Can you add some negative tests?

compiler/luci-interpreter/pal/linux/pal.cmake
Seunghui98: @seanshpark 

when I added the `BroadcastTo` op related code to the luci-interpreter, the following error occurred at the time of build.
```
[ 60%] Linking CXX executable luci_interpreter_memory_manager_test
/usr/bin/ld: libluci_interpreter.so: undefined reference to `tflite::TfLiteTypeGetSize(TfLiteType)'
collect2: error: ld returned 1 exit status
make[2]: *** [compiler/luci-interpreter/src/CMakeFiles/luci_interpreter_memory_manager_test.dir/build.make:124: compiler/luci-interpreter/src/luci_interpreter_memory_manager_test] Error 1
make[1]: *** [CMakeFiles/Makefile2:19323: compiler/luci-interpreter/src/CMakeFiles/luci_interpreter_memory_manager_test.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```
So with the help of @mhs4670go , he asked `BalyshevArtem` about this problem, and I got the following [advice]( https://github.com/Samsung/ONE/pull/11701#issuecomment-1759228267).
Perhaps the version of tensorflow has changed, resulting in the addition of library import.

compiler/luci-interpreter/src/kernels/BroadcastTo.test.cpp
Seunghui98: yes ! Thanks you for comments :)

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
seanshpark: Plz check tflite kernel and how it does.

Seunghui98: @seanshpark 

with the case of `shape_data[i] < 0`, I set the args value in the recipe file and proceeded with the build and test, but it worked well without failure. 
Should I add a new case where shape_data[i] < 0 ?

seanshpark: Have to checked tflite kernel source codes with BroadcastTo shape input?

Seunghui98: > Have to checked tflite kernel source codes with BroadcastTo shape input?

Perhaps if the dim of `broadcast_to` shape input becomes negative number, the dim value will be automatically adjusted by other dim values of the shape.

I found this in the numpy library.
For the first time, I declared a numpy array with a shape (1, 4, 3), and then I recreated it with (-4, 6) using `Reshape`. 

Therefore, it was confirmed that the dim value of the negative number was automatically dynamically calculated by the later 6 dim value and converted to (2, 6).

The following is the code, and the value output for this.

- python code
```
import numpy as np

A = np.array([[[1, 2, 3], [1, 2, 3], [1, 2, 3], [1, 2, 3]]])

print(np.shape(A))

A.shape = (-4, 6)
print(A)

print(np.shape(A))
```

- print log
```
(1, 4, 3)
[[1 2 3 1 2 3]
 [1 2 3 1 2 3]]
(2, 6)
```

In addition, the `shape` of `broadcast_to` in the `tensorflow` follows the characteristics of the `tf.shape` of the tensorflow, and when the dim of the `tf.shape` is negative number or `None`, the value of dim is dynamically changed like the characteristics of the `numpy`.
-> [reference](https://www.tensorflow.org/api_docs/python/tf/shape)

So I think i should add assign codes a negative value in this function to prevent it from entering the shape dim value.

So I think i need to add an `assert` code to prevent negative values from being set in the shape dim value in this function. :)

compiler/luci-interpreter/pal/linux/PALBroadcastTo.h
Seunghui98: @jinevening 

If the current `kMaxDims` is set to 8 specified in [tensorflow](https://github.com/tensorflow/tensorflow/blob/1b8f5c396f0c016ebe81fe1af029e6f205c926a4/tensorflow/lite/kernels/broadcast_to.cc#L34), 
memory out-of-bound error occurs in relation to `memcpy` in ubuntu 18.04. 
This has a compiler bug related to gcc version, so the value of `kMaxDims` seems to be supported up to 5.
In ubuntu 18.04, the basic version of gcc is 7.4, and it seems that there is a bug related to built_in_memcpy only in this version. This is because the upper versions, Ubuntu 20.04 and 22.04, have a basic version of gcc of 9.4 and no related errors occur here.
i posted a bug-related issue on [tensorflow](https://github.com/tensorflow/tensorflow/issues/62440),
but I didn't get the exact answer related to the bug. 😅
So I'm going to add the code set to `kMaxDims` as 5, and I'm going to add bug-related comments.

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,105 @@
...
+  // Check if output shape is broadcastable from input shape
+  int input_rank = input()->shape().num_dims();
+  int output_rank = output_shape.num_dims();
+  int extending_rank = output_rank - input_rank;
+
+  for (int idx = 0; idx < input_rank; ++idx)
+  {
+    LUCI_INTERPRETER_CHECK(input()->shape().dim(idx) == 1 ||
+                           input()->shape().dim(idx) == output_shape.dim(extending_rank + idx));
+  }
```
Seunghui98, line 78: @seanshpark 

> Have to checked tflite kernel source codes with BroadcastTo shape input?

I added code to check broadcastable value from like [broadcast_to.cc of tflite-kernels](https://github.com/tensorflow/tensorflow/blob/1b8f5c396f0c016ebe81fe1af029e6f205c926a4/tensorflow/lite/kernels/broadcast_to.cc#L60-L76).

new code checks the shape input value here.

>  is there no cases where shape_data[i] < 0 ?

If the shape input is negative dim values, it is checked that it is not broadcastable because it is different from the dim values of the input shape (If the dim value of input shape is positive).


> Then, what happens if the dim value of the input shape is negative?

In the `ONE` recipe, if the dim value of the input shape is set to a negative number, the following protobuf-related ERROR occurs, so it cannot be possible in the recipe.

```
[ 13%] Generating BroadcastTo_000.tflite
[libprotobuf ERROR /home/leeseunghui/workspace/ONE/externals/PROTOBUF/src/google/protobuf/text_format.cc:288] Error parsing text-format tflchef.ModelRecipe: 4:16: Expected integer, got: "-2"
ERROR: Failed to parse recipe 'BroadcastTo_000.recipe'
make[2]: *** [compiler/luci/tests/CMakeFiles/luci_testfiles.dir/build.make:3047: compiler/luci/tests/BroadcastTo_000.tflite] Error 255
make[1]: *** [CMakeFiles/Makefile2:6671: compiler/luci/tests/CMakeFiles/luci_testfiles.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```


And, I additionally added the test code to check the broadcastable.

compiler/luci-interpreter/src/kernels/BroadcastTo.test.cpp
Seunghui98: @jinevening 

Thank you a lot! 
I reflected it in the code regarding the related comments, 
so Could you review it again?

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  // Check if output shape is broadcastable from input shape
+  int input_rank = input()->shape().num_dims();
+  int output_rank = output_shape.num_dims();
+  int extending_rank = output_rank - input_rank;
+
+  for (int idx = 0; idx < input_rank; ++idx)
+  {
+    LUCI_INTERPRETER_CHECK(input()->shape().dim(idx) == 1 ||
+                           input()->shape().dim(idx) == output_shape.dim(extending_rank + idx));
```
seanshpark, line 77: can you please explain why add `idx` with `extending_rank` for output_shape?

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
seanshpark: 1/ please share the recipe file. 
2/ why have you used `-2` ?

Seunghui98: > 1/ please share the recipe file.
```
# BroadcastTo Op supports op version 2 and 3.
# BroadcastTo of Float type is version 2 and Quantized BroadcastTo is version 3.
# We are using tflite kernels for float values. so, the version should be 2.
# refer https://github.com/tensorflow/tensorflow/pull/46224/files
operand {
  name: "bc_input"
  type: FLOAT32
  shape { dim: "-2" dim: "3" }
}
operand {
  name: "bc_shape"
  type: INT32
  shape { dim: 3 }
  filler { tag: "explicit" arg: "1" arg: "-2" arg: "3" }
}
operand {
  name: "bc_ofm"
  type: FLOAT32
  shape { dim: "1" dim: "-2" dim: "3" }
}
operation {
  type: "BroadcastTo"
  input: "bc_input"
  input: "bc_shape"
  output: "bc_ofm"
  version: 2
}
input: "bc_input"
output: "bc_ofm"
```

> 2/ why have you used -2 ? what is the meaning of -2 ?

I just used `-2` to test if it is possible when dim values of `input` and `shape` are both negative. (In this case, it is set to any number that is judged to be broadcastable in the broadcast_to.cc of tflite kernel.)

seanshpark: minus values should work. have you checked other recipe files?

> I just used -2 to test if

As I understand you expect `-2` should also work. is this correct?

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
Seunghui98: `broadcast_to` operation can stretch the rank (if the rank of the shape is larger than the rank of the input), or the dim value can be stretched. (If the dim value of input shape is 1). 

So, like the [broadcast_to](https://github.com/tensorflow/tensorflow/blob/b487f0246c5804978f86f0a6bf2add1b911af23c/tensorflow/lite/kernels/broadcast_to.cc#L69-L76) of the tflite kernel, codes check whether it is `broadcastable` using `id` and `extending_rank`.

In order to help you understand the above explanations, I attached an example of the result of output shape after operation operation.
1) Stretch rank after operation operation.
2) Stretch dim after operation operation.
```
input shape : (2, 3)
shape shape : (1, 2, 3)
output shape : (1, 2, 3)  -> rank stretch

input shape : (1, 3)
shape shape : (3, 3)
output shape : (3, 3) -> dim stretch
```

seanshpark: I don't understand your example with how it's related with `extending_rank`.
Your explanation just jumps to the conclusion without any middle steps.

compiler/luci-interpreter/pal/linux/PALBroadcastTo.h
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#ifndef LUCI_INTERPRETER_PAL_BROADCASTTO_H
+#define LUCI_INTERPRETER_PAL_BROADCASTTO_H
+
+#include <tensorflow/lite/kernels/internal/reference/broadcast_to.h>
+
+namespace luci_interpreter_pal
+{
+static inline void BroadcastTo(const tflite::RuntimeShape &input_shape, const char *input_data,
```
seanshpark, line 24:
```suggestion
{

static inline void BroadcastTo(const tflite::RuntimeShape &input_shape, const char *input_data,
```

compiler/luci-interpreter/pal/linux/PALBroadcastTo.h
```diff
@@ -0,0 +1,37 @@
...
+                               TfLiteType data_type)
+{
+  // BroadcastTo op supports up to 8 kMaxDims in tensorflow.
+  // but, currently we support up to 5 dims because there is a compiler bug in 7.4.0 gcc version.
+  // https://github.com/tensorflow/tensorflow/blob/932af96ae91b4fa647dc50ad0f14c3e0b60affab/tensorflow/lite/kernels/broadcast_to.cc#L118
+  constexpr int kMaxDims = 5;
+  tflite::reference_ops::BroadcastTo<kMaxDims>(input_shape, input_data, output_shape, output_data,
+                                               data_type);
+}
+} // namespace luci_interpreter_pal
```
seanshpark, line 35:
```suggestion
}

} // namespace luci_interpreter_pal
```

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include "kernels/Utils.h"
+
+#include "PALBroadcastTo.h"
+
+#include <stdexcept>
+
+namespace luci_interpreter
+{
+
+namespace kernels
```
seanshpark, line 27:
```suggestion
namespace luci_interpreter
{
namespace kernels
```

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+
+namespace
+{
+// TODO Extract this function to Utils.h
+Shape extractShapeFromTensor(const Tensor *tensor)
```
seanshpark, line 33:
```suggestion
{

// TODO Extract this function to Utils.h
Shape extractShapeFromTensor(const Tensor *tensor)
```

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+  LUCI_INTERPRETER_CHECK(input()->element_type() == output()->element_type());
+
+  Shape output_shape = extractShapeFromTensor(shape());
+
+  int input_rank = input()->shape().num_dims();
+  int output_rank = output_shape.num_dims();
+
+  // Ensures the shape is 1D tensor
+  LUCI_INTERPRETER_CHECK(output_rank >= 1);
```
seanshpark, line 78: `is 1D tensor` vs `output_rank >= 1` doesn't fit

compiler/luci-interpreter/src/kernels/BroadcastTo.test.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "kernels/BroadcastTo.h"
+#include "kernels/TestUtils.h"
+#include "luci_interpreter/TestMemoryManager.h"
+
+namespace luci_interpreter
+{
+namespace kernels
+{
+namespace
+{
```
seanshpark, line 27:
```suggestion
{

namespace
{
```

compiler/luci-interpreter/src/kernels/BroadcastTo.test.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  BroadcastTo kernel(&input_tensor, &shape_tensor, &output_tensor);
+
+  EXPECT_ANY_THROW(kernel.configure());
+}
+
+} // namespace
+} // namespace kernels
+
+} // namespace luci_interpreter
```
seanshpark, line 118:
```suggestion
} // namespace

} // namespace kernels
} // namespace luci_interpreter
```

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
Seunghui98: The official document explains BroadcastTo op as follows. -> [link](https://www.tensorflow.org/api_docs/python/tf/broadcast_to)

> Broadcasting is the process of making arrays to have compatible shapes for arithmetic operations. Two shapes are compatible if for each dimension pair they are either equal or one of them is one.

> When broadcasting, if a tensor has fewer axes than necessary its shape is padded on the left with ones. 

Considering this case, it seems that `broadcast_to.cc` in tflite kernel approaches the index with `idx` and `extending_rank` variables. :)

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,121 @@
...
+  // Ensures output rank is not less than input rank
+  LUCI_INTERPRETER_CHECK(input_rank <= output_rank);
+
+  // Check if output shape is broadcastable from input shape
+  int extending_rank = output_rank - input_rank;
+  for (int idx = 0; idx < input_rank; ++idx)
+  {
+    LUCI_INTERPRETER_CHECK(input()->shape().dim(idx) == 1 ||
+                           input()->shape().dim(idx) == output_shape.dim(extending_rank + idx));
+  }
```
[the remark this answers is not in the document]
seanshpark: Q) why is there `input()->shape().dim(idx) == 1` ?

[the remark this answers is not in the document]
Seunghui98: @seanshpark 

> Q) why is there input()->shape().dim(idx) == 1 ?

This is a code that checks whether `broadcast_to` is valid in tensorflow core `broadcast_to_op.cc`. -> [link](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/bcast.h#L190-L226)
The code part below is difficult to understand 😅, 
but there is a description of how to check valid in the annotation.

```
  // Going through each dimension starting from the inner-most
  // dimension, compares dimension of x and y. They are compatible if
  // they are equal or either is 1.
```

so, it seems that the tflite kernel also checks whether the dim value of `input()->shape()` and `output()->shape()` have the same dim value or dim value of `input()->shape()` is 1 with the following code.

```
LUCI_INTERPRETER_CHECK(input()->shape().dim(idx) == 1 ||
                       input()->shape().dim(idx) == output_shape.dim(extending_rank + idx));
```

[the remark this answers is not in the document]
seanshpark: > Since TensorFlow is row-major, the right-most dimension

what do you know about `row-major`?

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+void BroadcastTo::execute() const
+{
+  switch (input()->element_type())
+  {
+    case DataType::FLOAT32:
+      evalFloat();
+      break;
+    default:
+      throw std::runtime_error("Unsupported type.");
```
seanshpark, line 109:
```suggestion
      throw std::runtime_error("luci-intp BroadcastTo Unsupported type.");
```

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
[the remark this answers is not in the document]
seanshpark: > in more detail in the core of tensorflow)

why tensorflow? our circle follows tflite not tensorflow.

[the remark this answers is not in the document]
Seunghui98: We talked together offline, 
but in this regard, I write comments for someone to look for later. :)

> what do you know about row-major?

i don't know much about `row-major`.
The reason for bringing the above rule was to find the rule of the second broadcast_to op related to the above question. (2. ~).

> why tensorflow? our circle follows tflite not tensorflow.

i looked up the `tensorflow` with the idea that I could take a closer look at the characteristics of op after seeing the annotation of the codes in the `broadcast_to` of tflite kernel that annotation brought the characteristics of tensorflow broadcast_to_op. :)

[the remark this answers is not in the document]
seanshpark: Thanks for the explanation !

> error: ‘extending_rank’ was not declared in this scope

I don't know why this happened cause this variable existed in your change.

seanshpark (approved):
LGTM thank you!

compiler/luci-interpreter/pal/linux/PALBroadcastTo.h
```diff
@@ -0,0 +1,39 @@
...
+namespace luci_interpreter_pal
+{
+
+static inline void BroadcastTo(const tflite::RuntimeShape &input_shape, const char *input_data,
+                               const tflite::RuntimeShape &output_shape, char *output_data,
+                               TfLiteType data_type)
+{
+  // BroadcastTo op supports up to 8 kMaxDims in tensorflow.
+  // but, currently we support up to 5 dims because there is a compiler bug in 7.4.0 gcc version.
+  // https://github.com/tensorflow/tensorflow/blob/932af96ae91b4fa647dc50ad0f14c3e0b60affab/tensorflow/lite/kernels/broadcast_to.cc#L118
```
mhs4670go, line 31: Just seven length will do the trick:)
```suggestion
  // https://github.com/tensorflow/tensorflow/blob/932af96/tensorflow/lite/kernels/broadcast_to.cc#L118
```

mhs4670go (approved):
LGTM. Maybe my comment can be applied in another PR as it is just a small change.

jinevening (approved):
LGTM

compiler/luci-interpreter/src/kernels/BroadcastTo.cpp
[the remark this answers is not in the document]
Seunghui98: @seanshpark 
> I don't know why this happened cause this variable existed in your change.

Oh, i accidentally copied another log. 😅 
I'm sorry I corrected it again to the log that occurs in that error. :)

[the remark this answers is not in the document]
Seunghui98: @seanshpark 
> I don't know why this happened cause this variable existed in your change.

Oh, i accidentally copied another log. 😅 
I'm sorry I corrected it again to the log that occurs in that error. :)

Issue #12060, opened 2023-11-20
[luci] BatchMatMul shape inference


seanshpark:
One of our internal model has problem with shape inference of BatchMatMul.

> circle2circle: ERROR: Internal Exception. x_rhs and y_lhs should be same 
> compiler/luci/service/src/CircleShapeInferenceRule.cpp:486

which is 
```cxx
  if (x_rhs.known() && y_lhs.known() && not(x_rhs == y_lhs))
    INTERNAL_EXN("x_rhs and y_lhs should be same");
```
https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/batch-mat-mul shows that there is no condition check for this...

seanshpark:
Nodes from the model:
![a831d9c6-d5ca-4bd4-b92b-ab1c4c3e3bb5](https://github.com/Samsung/ONE/assets/4616940/14b89ae8-1385-41ba-9287-b60559bf5cbc)

jinevening:
I guess that this issue can be closed?

PR #12061, opened 2023-11-20
[luci/svc] Revise BatchMatMul shape inference

seanshpark:
This will revise BatchMatMul shape inference to pass and show log for shape condition.


seanshpark:
for #12060

compiler/luci/service/src/CircleShapeInferenceRule.cpp
```diff
@@ -483,7 +483,10 @@ loco::NodeShape infer_batchmatmul_shape(const loco::TensorShape &x_shape,
   loco::Dimension y_rhs = adj_y ? y_shape.dim(y_rank - 2) : y_shape.dim(y_rank - 1);
 
   if (x_rhs.known() && y_lhs.known() && not(x_rhs == y_lhs))
```
jinevening, line 485: Is it okay for `x_rhs != y_lhs` ? TFLite interpreter (and circle interpreter) does not allows such a shape.

seanshpark: I think the cause is from circle-mlir which seems fix is needed.
Thanks for taking a look at tflite interpreter codes :)

seanshpark:
closing as strange shape was from circle-mir converter...

Issue #12070, opened 2023-11-21
[ONE] efficientformer_v2_s1_baseline.onnx subgraph IndexError


Aeren1564:
I tried the convert the following ONNX model into quantized circle model:

[model.onnx.zip](https://github.com/Samsung/ONE/files/13422337/model.onnx.zip)

(which is a subgraph extracted from [efficientformer_v2_s1_baseline.onnx.zip](https://github.com/Samsung/ONE/files/13422365/efficientformer_v2_s1_baseline.onnx.zip) with [onnx extractor](https://onnx.ai/onnx/api/utils.html))

and got the following error message:

```
...

  warnings.warn(
WARNING:absl:`/patch_embed/patch_embed_5/Mul_1_output_0` is not a valid tf.function parameter name. Sanitizing to `arg__patch_embed_patch_embed_5_Mul_1_output_0`.
WARNING:tensorflow:`/patch_embed/patch_embed_5/Mul_1_output_0` is not a valid node name. Accepted names conform to Regex /re.compile('^[A-Za-z0-9.][A-Za-z0-9_.\\\\/>-]*$')/
WARNING:tensorflow:`/patch_embed/patch_embed_5/Mul_1_output_0` is not a valid node name. Accepted names conform to Regex /re.compile('^[A-Za-z0-9.][A-Za-z0-9_.\\\\/>-]*$')/
one-import-onnx: StagingError: in user code:


    IndexError: tuple index out of range
```

This is the config file I used.

```asm
[onecc]
one-import-tf=False
one-import-tflite=False
one-import-bcq=False
one-import-onnx=True
one-optimize=True
one-quantize=True
one-pack=False
one-codegen=False
include=O1

[one-import-onnx]
input_path=./model.onnx
output_path=./model.circle

[one-optimize]
input_path=./model.circle
output_path=./model.opt.circle
convert_nchw_to_nhwc=True

[one-quantize]
input_path=./model.opt.circle
output_path=./model.opt.q8.circle
granularity=channel
quantized_dtype=uint8
```

seanshpark:
Reason: I/O name includes `/` which TensorFlow doesn't accept(don't know why... old version was OK)
I'll prepare a patch for this.

seanshpark:
@Aeren1564 , as code has landed, plz try with next release(tomorrow morning).

PR #12072, opened 2023-11-21
[one-cmds] Revise one-import-onnx I/O name sanitizier

seanshpark:
This will revise one-import-onnx I/O name sanitizer with '/'
- add prefix with 'a_' is added unconditionally as sanitized name may have multiple '_'

### DIFF
--- a/compiler/one-cmds/one-import-onnx
+++ b/compiler/one-cmds/one-import-onnx
@@ -70,21 +70,21 @@ class TidyIONames:
         for idx in range(0, len(self.onnx_model.graph.input)):
             name = self.onnx_model.graph.input[idx].name
             if not name in self.initializers:
-                if '.' in name or ':' in name or name[:1].isdigit():
+                if '.' in name or ':' in name or '/' in name:
                     self.input_nodes.append(name)
                     name_alt = name.replace('.', '_')
                     name_alt = name_alt.replace(':', '_')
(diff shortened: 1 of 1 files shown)


seanshpark:
for #12070

TF shows recommended name as
```
Accepted names conform to Regex /re.compile('^[A-Za-z0-9.][A-Za-z0-9_.\\\\/>-]*$')/
```

compiler/one-cmds/one-import-onnx
```diff
@@ -70,21 +70,21 @@ class TidyIONames:
...
             if not name in self.initializers:
-                if '.' in name or ':' in name or name[:1].isdigit():
+                if '.' in name or ':' in name or '/' in name:
                     self.input_nodes.append(name)
                     name_alt = name.replace('.', '_')
                     name_alt = name_alt.replace(':', '_')
-                    if name_alt[:1].isdigit():
-                        name_alt = 'a_' + name_alt
+                    name_alt = name_alt.replace('/', '_')
+                    name_alt = 'a_' + name_alt  # prevent start with digit or _
```
seanshpark, line 78: unconditionally add `a_` prefix to make it simple if there are multiple `_` in front of the name. so we don't need to check if name starts with numbers.

mhs4670go (approved):
LGTM

PR #12073, opened 2023-11-21
[onert] Define LossCode and LossInfo as separated files

jyoungyun:
This commit implements LossCode and LossInfo as separated files and use the same structure in Compiler, TrainingInfo and Loss IR.

### DIFF
--- a/runtime/onert/core/include/ir/operation/Loss.h
+++ b/runtime/onert/core/include/ir/operation/Loss.h
@@ -36,35 +36,12 @@ public:
     // TODO Add more inputs if necessary
   };
 
-  // NOTE It is not yet determined how to get the information of the previous activation when
-  //      generating kernels of Loss operation for each backend. If it is determined to get it
-  //      from the object of this class, we have to consider whether to change this enum class.
-  enum class Type
-  {
-    MEAN_SQUARED_ERROR,
-    CATEGORICAL_CROSSENTROPY
-  };
-
-  struct Param
-  {
-    Type op_type;
-    // TODO Add more params if necessary
-    Param() : op_type(Type::MEAN_SQUARED_ERROR) {}
-  };
-
 public:
-  Loss(const OperandIndexSequence &inputs, const OperandIndexSequence &outputs, const Param &param);
+  Loss(const OperandIndexSequence &inputs, const OperandIndexSequence &outputs);
 
 public:
   void accept(OperationVisitor &v) const override;
-  std::string name() const override;
   OpCode opcode() const final { return OpCode::Loss; }
-
-public:
-  const Param &param() const { return _param; }
-
-private:
-  Param _param;
 };
 
 } // namespace operation
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -1186,14 +1186,14 @@ NNFW_STATUS nnfw_session::train_prepare(const nnfw_train_info *info)
 
     auto convertLossType = [](const int &type) {
       if (type == NNFW_TRAIN_LOSS_MEAN_SQUARED_ERROR)
-        return onert::ir::operation::Loss::Type::MEAN_SQUARED_ERROR;
+        return onert::ir::train::LossCode::MeanSquaredError;
       if (type == NNFW_TRAIN_LOSS_CATEGORICAL_CROSSENTROPY)
-        return onert::ir::operation::Loss::Type::CATEGORICAL_CROSSENTROPY;
+        return onert::ir::train::LossCode::CategoricalCrossentropy;
       else
         throw std::runtime_error("not supported loss type");
     };
-    onert::compiler::train::LossInfo loss_info;
-    loss_info.type = convertLossType(tinfo.loss);
+    onert::ir::train::LossInfo loss_info;
+    loss_info.loss_code = convertLossType(tinfo.loss);
 
     auto convertOptType = [](const int &type) {
       if (type == NNFW_TRAIN_OPTIMIZER_SGD)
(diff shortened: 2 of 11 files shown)


runtime/onert/core/include/ir/operation/Loss.h
```diff
@@ -36,35 +36,16 @@ class Loss : public Operation
...
-  };
-
 public:
-  Loss(const OperandIndexSequence &inputs, const OperandIndexSequence &outputs, const Param &param);
+  Loss(const OperandIndexSequence &inputs, const OperandIndexSequence &outputs);
 
 public:
   void accept(OperationVisitor &v) const override;
-  std::string name() const override;
+  std::string name() const override { return _loss_name + toString(opcode()); };
```
ragmani, line 44: The name seems to be duplicated. Is `_loss_name` necessary?
```suggestion
  std::string name() const override { return toString(opcode()); };
```

jyoungyun: The `_loss_name` is necessary to print the detailed loss type. The opcode() only prints `Loss`, but `_loss_name` has one of the detailed loss name of `MeanSquaredError` or `CategoricalCrossentropy`.

jyoungyun: GraphDump calls the operation through ir::Operation visitor to print the loss name. However, the Loss in ir::Operation does not have detailed train information. It is included in ir::train::Operation. Thus, the _loss_name will be updated from the derived class to the base class.

ragmani: I see. How about removing moving this method to `ir::train::operation::Loss`? If so, you can remove `_loss_name` and implement more simply like below.
```suggestion
  std::string name() const override { return toString(_param.loss_code); };
```

jyoungyun: If this function moves to the `ir::train::operation::Loss`, the GraphDump does not print the `_param.loss_code` because GraphDump uses the `ir::OperationVisitor` not `ir::train::OperationVisitor`... :(

zetwhite (approved):
LGTM :+1:

runtime/onert/core/include/ir/operation/Loss.h
glistening: I think `Loss` is similar to `BinaryArithmetic` in inference operation. Why do we need to define `Loss` as a concrete op? Is there any problem to define `MeanSquaredError` and `CategoricalCrossEntrophy` as operations respectively?

jyoungyun: I thought about it. Loss operation is only necessary in the train kernel, but onert is designed to operate each operation in `ir::operation`. Therefore, if `LossLayer` is not defined in `ir::operation` and each loss function such as `MeanSquaredError` or `CategoricalCrossentropy` is implemented, this operation should be located in `ir::operation` rather than `ir::train::operation`. If we can implement the operation in `ir::train::operation` without defining in `ir::operation`, your approach seems appropriate.

In this PR, I'd like to introduce `LossCode` and `LossInfo` classes. Therefore, it would be good to discuss it on a separate issue. Thank you for your comment.

ragmani: > If this function moves to the ir::train::operation::Loss, the GraphDump does not print the _param.loss_code because GraphDump uses the ir::OperationVisitor not ir::train::OperationVisitor... :(

Does `ir::OperationVisitor` block overloading virtual functions?

ragmani: > Is there any problem to define MeanSquaredError and CategoricalCrossEntrophy as operations respectively?

As far as I know, the reason why `ir::Operation` that can define a operation among multiple operations such as `BinaryArithmetic` are introduced is duplicated code. I don't know other problems.

glistening: > In this PR, I'd like to introduce `LossCode` and `LossInfo` classes. Therefore, it would be good to discuss it on a separate issue. Thank you for your comment.

Okay. Let's consider this approach later.

glistening: > I thought about it. Loss operation is only necessary in the train kernel, but onert is designed to operate each operation in `ir::operation`. Therefore, if `LossLayer` is not defined in `ir::operation` and each loss function such as `MeanSquaredError` or `CategoricalCrossentropy` is implemented, this operation should be located in `ir::operation` rather than `ir::train::operation`. If we can implement the operation in `ir::train::operation` without defining in `ir::operation`, your approach seems appropriate.

Just a note:

`TrainableOperator` already has its own operator list. - `runtime/onert/core/include/ir/train/Operations.lst`. Thus, at a glance, I don't find problem.

jyoungyun: @glistening @ragmani 
I finally understood you guys suggestion. I thought of implementing the `name()` function in derived classes such as 'MeanSquaredError` or `CategoricalCrossentropy` classes. In this case, I thought it couldn't do that because these classes were derived classes and the operation visitor couldn't call the derived class function. Now I know that `name()` can be implemented in `ir::train::operation::Loss` class, not `ir::operation::Loss` class. I updated this PR. Thank you.

jyoungyun: There is one think we need to know. The output varies depending on what GraphDumper hands over.

```
[   GraphDumper  ]   %11 = @4_MeanSquaredErrorLoss(%9, %14)
```

glistening (approved):
LGTM. (I want to try to remove `ir::operation::Loss` after merging this.)

zetwhite (approved):

ragmani (approved):
LGTM

glistening:
> LGTM. (I want to try to remove `ir::operation::Loss` after merging this.)

I took a look. It requires more changes than I guessed. It is related how to reuse the existing code, which is based and designed for inference, with less effort.

PR #11849, opened 2023-11-01
[circle-mpqsolver] Add Bisection Unitests

stamalakhov:
This draft refactors bisection to enable in-memory data input for Evaluator and BisectionSolver.

Previous draft: #11452
Related: #11374 
ONE-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

PR #12085, opened 2023-11-22
[circle-mpqsolver] Introduce getSampleInput

stamalakhov:
This commit introduces getSampleInput for using in tests.

Draft: https://github.com/Samsung/ONE/pull/11849
Related: https://github.com/Samsung/ONE/issues/11374

### DIFF
--- a/compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
+++ b/compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
@@ -212,13 +212,13 @@ bool isFileExists(const std::string &path)
 namespace data_utils
 {
 
-class SampleDataInput final : public core::DataProvider
+class SingleDataProvider final : public core::DataProvider
 {
 public:
-  SampleDataInput() : _num_of_samples(1) {}
-  size_t numSamples() const override { return _num_of_samples; }
-  uint32_t numInputs(uint32_t sample) const override { return 1; }
-  void getSampleInput(uint32_t sample, uint32_t input, core::InputData &data) const
+  SingleDataProvider() = default;
+  size_t numSamples() const override { return 1; }
+  uint32_t numInputs(uint32_t) const override { return 1; }
+  void getSampleInput(uint32_t, uint32_t, core::InputData &data) const
   {
     size_t size = data.data().size() / sizeof(float);
     auto floats = reinterpret_cast<float *>(data.data().data());
@@ -227,14 +227,11 @@ public:
       floats[idx] = 0.f; // or any other value
     }
   }
-
-private:
-  uint32_t _num_of_samples;
 };
 
-std::unique_ptr<mpqsolver::core::DataProvider> getSampleInput()
+std::unique_ptr<mpqsolver::core::DataProvider> getSingleDataProvider()
 {
-  return std::make_unique<SampleDataInput>();
+  return std::make_unique<SingleDataProvider>();
 }
 
 } // namespace data_utils
--- a/compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
+++ b/compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
@@ -208,5 +208,36 @@ bool isFileExists(const std::string &path)
 }
 
 } // namespace io_utils
+
+namespace data_utils
+{
+
+class SampleDataInput final : public core::DataProvider
+{
+public:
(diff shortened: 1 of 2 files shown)


compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
```diff
@@ -208,5 +208,36 @@ bool isFileExists(const std::string &path)
...
+    size_t size = data.data().size() / sizeof(float);
+    auto floats = reinterpret_cast<float *>(data.data().data());
+    for (uint32_t idx = 0; idx < size; idx++)
+    {
+      floats[idx] = 0.f; // or any other value
+    }
+  }
+
+private:
+  uint32_t _num_of_samples;
```
seanshpark, line 232: I don't know much about `DataProvider`,
if `SampleDataInput` only provides single sample, then `numInputs` uses `1` not `_num_of_samples` the `_num_of_samples` seems useless.

compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
```diff
@@ -208,5 +208,36 @@ bool isFileExists(const std::string &path)
 }
 
 } // namespace io_utils
+
+namespace data_utils
+{
+
+class SampleDataInput final : public core::DataProvider
```
seanshpark, line 215: optional) not understanding much about this but how about `SingleDataProvider` ?

compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
```diff
@@ -208,5 +208,36 @@ bool isFileExists(const std::string &path)
...
+{
+public:
+  SampleDataInput() : _num_of_samples(1) {}
+  size_t numSamples() const override { return _num_of_samples; }
+  uint32_t numInputs(uint32_t sample) const override { return 1; }
```
seanshpark, line 220:
```suggestion
  uint32_t numInputs(uint32_t) const override { return 1; }
```

compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
```diff
@@ -208,5 +208,36 @@ bool isFileExists(const std::string &path)
...
+  void getSampleInput(uint32_t sample, uint32_t input, core::InputData &data) const
```
seanshpark, line 221:
```suggestion
  void getSampleInput(uint32_t, uint32_t, core::InputData &data) const
```

compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
```diff
@@ -208,5 +208,36 @@ bool isFileExists(const std::string &path)
...
+  {
+    size_t size = data.data().size() / sizeof(float);
```
seanshpark, line 223: Q) just curious and my be stupid question, if we can use number of elements in `core::InputData` are used, why do we pass `sample` and `input`?

compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
stamalakhov: Ahh. Yes. Sorry. You're right. I'll remove it.

compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
stamalakhov: > optional) not understanding much about this but how about `SingleDataProvider` ?

Yes. Looks better. Thank you.

compiler/circle-mpqsolver/src/core/TestHelper.test.cpp
stamalakhov: > Q) just curious and my be stupid question, if we can use number of elements in `core::InputData` are used, why do we pass `sample` and `input`?

@seanshpark
`Sample` stands for specific `data_point`, this `data_point` in its turn can have any number of tensors, so it's two-dimensional, that is why we need two parameters to query specific data. `size` it's just a number of elements in `data`. These parameters are unrelated. So we need them all. Quantity of inputs for specific `sample` can be queried by `numInputs`.

seanshpark (approved):
LGTM thank you!

jinevening (approved):
LGTM

PR #12106, opened 2023-11-23
[tools/onert_train] Add loss_reduction_type option

jyoungyun:
This commit adds `--loss_reduction_type` option.
The default value is 0, which means to follow the model setting or
onert train setting.

~~waiting for #12105~~

### DIFF
--- a/tests/tools/onert_train/src/onert_train.cc
+++ b/tests/tools/onert_train/src/onert_train.cc
@@ -132,6 +132,21 @@ int main(const int argc, char **argv)
       }
     };
 
+    auto convertLossReductionType = [](int type) {
+      switch (type)
+      {
+        case 0:
+          return NNFW_TRAIN_LOSS_REDUCTION_INVALID;
+        case 1:
+          return NNFW_TRAIN_LOSS_REDUCTION_SUM_OVER_BATCH_SIZE;
+        case 2:
+          return NNFW_TRAIN_LOSS_REDUCTION_SUM;
+        default:
+          std::cerr << "E: not supported loss reduction type" << std::endl;
+          exit(-1);
+      }
+    };
+
     auto convertOptType = [](int type) {
       switch (type)
       {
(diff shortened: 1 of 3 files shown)


tests/tools/onert_train/src/onert_train.cc
```diff
@@ -132,6 +132,21 @@ int main(const int argc, char **argv)
       }
     };
 
+    auto convertLossReductionType = [](int type) {
+      switch (type)
+      {
+        case 0:
+          return NNFW_TRAIN_LOSS_REDUCTION_INVALID;
```
ragmani, line 139: `default_value` of args is 0, but it is converted to `NNFW_TRAIN_LOSS_REDUCTION_INVALID` here. Does `NNFW_TRAIN_LOSS_REDUCTION_INVALID` mean "Use default setting (Model parameter or ONERT train setting)" in onert?

jyoungyun: Yes. If you want to use the default setting in model parameter or onert training setting, you can pass the `0` in the `onert_train` tool. The `onert_train` tool converts the `0` into  `NNFW_TRAIN_LOSS_REDUCTION_INVALID` value. If the onert obtains the `NNFW_TRAIN_LOSS_REDUCTION_INVALID` value, it means that there is no additional setting value for the loss reduction type. In this case, the onert will use the model reduction type or onert own reduction type setting.

https://github.com/Samsung/ONE/blob/58abcdb96eaaf2a70287dc0d0fb10b57ede06c9d/runtime/onert/core/include/compiler/train/TrainingInfo.h#L49-L51

jyoungyun:
```bash
$ ./Product/out/bin/onert_train --help
...
  --loss_reduction_type arg (=0)        Loss Reduction type
                                        0: Use default setting (Model parameter or ONERT train 
                                        setting)
                                        1: SUM_OVER_BATCH_SIZE
                                        2: SUM
```

tests/tools/onert_train/src/onert_train.cc
ragmani: I see. I just couldn't matched the word, "INVALID" and the sentence, "Use default setting".

ragmani (approved):
LGTM

hseok-oh (approved):
LGTM

PR #12107, opened 2023-11-23
[tools] Introduce random_data_generator for onert_train

jyoungyun:
This commit introduces random_data_generator for onert_train. This tool generates random data file using onert_run tool.


zetwhite (commented):
I read the code and its logic (and readme) was quite clear.

But I have an uncertain part:  
It looks like this tool generates output from random input using a "model" files.
Does the "model" file have to be already a well-trained one? or the one needed to be trained..?

jyoungyun:
> The model file used is already a well-trained one? or the one needed to be trained..?

It doesn't matter if the model file is trained or not. When we implement new training operation, we sometimes use sample model files with new operators that you are interested in. In this case, there is no dataset for this new operator. This tool help you create a suitable dataset for the new model.

zetwhite (approved):

ragmani:
> It doesn't matter if the model file is trained or not. When we implement new training operation, we sometimes use sample model files with new operators that you are interested in. In this case, there is no dataset for this new operator. This tool help you create a suitable dataset for the new model.

I guess loss values are always 0 when running `onert_train` with any models and datasets to be generated by this generator. If so, gradients will also be 0. So I didn't understand why this generator is necessary? Is the purpose of this generator just to create datasets so that we can call `onert_train`? What is the difference from input datasets and output datasets filled with random values?

jyoungyun:
> I guess loss values are always 0 when running `onert_train` with any models and datasets generated by this generator. If so, gradients will also be 0. That means we cannot validate backwarding phases too with datasets generated by this generator.  So I didn't understand why this generator is necessary. Is the purpose of this generator just to create datasets so that we can call `onert_train`? Could I get the purpose of this generator?

When using the same input and output data file, the loss value is always `0`. So I create multiple datasets and test them using different input/output datasets. It doesn't mean much, but check if it works well by calculating the loss value. This tool is not required if we can handle all test cases using the actual dataset. Do you think we can test all training operators using the actual dataset?

ragmani:
> When using the same input and output data file, the loss value is always 0. So I create multiple datasets and test them using different input/output datasets. It doesn't mean much, but check if it works well by calculating the loss value. This tool is not required if we can handle all test cases using the actual dataset. Do you think we can test all training operators using the actual dataset?

No, I didn't mean that. I couldn't find any reason why `onert_run` runs to generate datasets. If we have to use different input/output datasets, what is the difference from generating inputs/outputs with random data?

jyoungyun:
@ragmani 

> No, I didn't mean that. I couldn't find any reason why `onert_run` runs to generate datasets. If we have to use different input/output datasets, what is the difference from generating  input/output datasets with random data?

In fact, I did not consider the case of not using the `onert_run` tool to create a dataset. Using `onert_run` has the advantage(?) of getting `0` loss if train-related things are properly implemented when passing through co-generated input/output dataset. We can fill the dataset with arbitrary data. Is the method better than using `onert_run` tool? If true, I will modify this PR.

zetwhite:
> what is the difference from generating input/output datasets with random data? 

I personally thought that this tool help generate output data with some restriction.
For example, "binary classifier" expect output (`y-hat`) as boolean value (0 or 1).
With this tool, It could easily generate output with boolean value. 
Otherwise, Making output data with restriction might be a bit troublesome.

ragmani:
> In fact, I did not consider the case of not using the onert_run tool to create a dataset. Using onert_run has the advantage(?) of getting 0 loss if train-related things are properly implemented when passing through co-generated input/output dataset.

If you mean that co-generated input/output datasets can validate all forwarding phases, I also think it can be an advantage of using `onert_run`.

> We can fill the dataset with arbitrary data. Is the method better than using onert_run tool?

If we can validate all training phases by arbitrary datasets, I think it's a good method.
But if not, I didn't find the other better methods. How about manipulating output datasets generated by `onert_run` so that loss value is not 0? But I'm not sure if it's a better method.

ragmani:
> I personally thought that this tool help generate output data with some restriction.

> For example, "binary classifier" expect output (y-hat) as boolean value (0 or 1).
With this tool, It could easily generate output with boolean value.
Otherwise, Making output data with restriction might be a bit troublesome.

@zetwhite 
I didn't catch what you worried about. `onert` deals with bool type as 8 bit bool. If so, we can make a generator that generates output dataset with random data in 8 bit boolean type like `onert`.

zetwhite:
> 
> @zetwhite 
> > I personally thought that this tool help generate output data with some restriction.
> 
> > For example, "binary classifier" expect output (y-hat) as boolean value (0 or 1).
> With this tool, It could easily generate output with boolean value.
> Otherwise, Making output data with restriction might be a bit troublesome.
> 
> I didn't catch what you worried about. `onert` deals with bool type as 8 bit bool. If so, we can make a generator that generates output dataset with random data in 8 bit boolean type like output of `onert`. 

Oh I see. If so, I couldn't find a reason that we should generate output data using model inference(onert_run).

jyoungyun:
> If we can validate all training phases by arbitrary datasets, I think it's a good method.
But if not, I didn't find the other better methods. How about manipulating output datasets generated by `onert_run` so that loss value is not 0? But I'm not sure if it's a better method.

I'm looking for a way to implement it so that the loss does not become `0`. I will update this PR soon. :)

PR #12111, opened 2023-11-24
[onert] Remove ir::train::operation

glistening:
It removes ir::train::operation by adding clone() in IOperation. It removes op to train::op converting and visitor.

```
$ Product/x86_64-linux.debug/out/bin/onert_train mnist.circle --load_input:raw mnist_train.i160.bin --load_expected:raw mnist_train.o160.bin 
Model Expected Filename mnist_train.o160.bin
Model Input Filename mnist_train.i160.bin
Model Filename mnist.circle
Epoch 1/5 - time: 22.016ms/step - loss: [0] 0.2247
Epoch 2/5 - time: 21.306ms/step - loss: [0] 0.2198
Epoch 3/5 - time: 21.321ms/step - loss: [0] 0.2153
Epoch 4/5 - time: 21.290ms/step - loss: [0] 0.2111
Epoch 5/5 - time: 21.281ms/step - loss: [0] 0.2071
MODEL_LOAD   takes 1.4030 ms
PREPARE      takes 8.5810 ms
EXECUTE      takes 536.7260 ms
- Epoch 1      takes 110.0820 ms
- Epoch 2      takes 106.5320 ms
- Epoch 3      takes 106.6060 ms
- Epoch 4      takes 106.4480 ms
- Epoch 5      takes 106.4050 ms
```


glistening:
The only meaningful difference between `IOperation` and `ITrainableOperation` is:

```cc
  virtual std::unique_ptr<ITrainableOperation> clone() const = 0;
```

By moving `clone()` to `IOperation`, we can get rid of

- `ITrainableOperation`
- runtime/onert/core/{include,src}/ir/train/operation/*.{h,cpp}
- *TrainableConverter*

about 1334 lines.

jyoungyun:
I'd like your suggestion. If there is only one that is really different, it would be better to merge into one. :) One thing I'm concerned about is that something that only training feature needs may happen in the future. This is because we have not yet fully implemented training. Anyway, I think it is a good attempt. 👍

hseok-oh:
I agree this PR. Additionally, I think we are ready to remove build flag `ENABLE_ONERT_TRAIN` on core.

glistening:
@ragmani PTAL

ragmani (changes requested):
`onert` still only performs some of training features, and we do not yet understand much about training features. And the features you try to remove from `onert` was introduced to increase the scalability of training features for each operation in onert. To remove the feature to be removed, I think we need to convince ourselves that those features are not needed or introduce something to replace them.
- The features you try to remove in this PR.
  - `ITrainableOperation` is an interface that can have functions to be called for training operations.
  - IR for each operation that contains information for training.

jyoungyun:
If it is difficult to implement this work right now, how about making a list of what we want to do in the future for refactoring and write it down there?

glistening:
> If trainable operations in `ir/train/operation/*` have field members, it's meaningful. It is a prototype pattern applied for perfect deep copying for trainable operations. So we have to remove them with trainable operations instead of moving `clone()` to `IOperation` if trainable operations are removed. If trainable operations are not removed, I think it's better to keep it as is.

I know. I am asking that it is likely to happen. I believe it would be better to use other design instead of adding too much like current.

glistening:
@chunseoklee, @zetwhite Please give your opinion. Or we may discuss more offline.

ragmani:
> I know. I am asking that it is likely to happen. I believe it would be better to use other design instead of adding too much like current. 

I mean in the sentence that I recommend you do not move `clone()` to `IOperation` even if trainable operations are removed since you tried to move `clone()` to `IOperation`.

glistening:
> > I know. I am asking that it is likely to happen. I believe it would be better to use other design instead of adding too much like current.
> 
> I mean in the sentence that I recommend you do not move `clone()` to `IOperation` even if trainable operations are removed since you tried to move `clone()` to `IOperation`.

Exactly, I don't think `clone` is the method for `ITrainableOperator`. It is very general method, which can be used in other classes. We may define this method as a kind of Mixin.

zetwhite:
> @chunseoklee, @zetwhite Please give your opinion. Or we may discuss more offline.

I started to read the code now. I'll leave my opinion soon.

ragmani:
> Exactly, I don't think `clone` is the method for `ITrainableOperator`. It is very general method, which can be used in other classes. We may define this method as a kind of Mixin.

There is already `OperationCloner`. If you don't want to use `OperationCloner`, replace it with using the `clone()` methods. It's duplicated.

zetwhite:
> I started to read the code now. I'll leave my opinion soon.

I read the code - It removes `ITrainableOp` and related visitor.
Removing duplication looks good to me. 

I don't have any strong opinion about this change.

ragmani:
> Almost all code is duplication of code. I felt a bad smell from current structure.

I don't think it's a bad smell. Rather, the two options you mentioned violate SRP for clean code. But if you want to remove it, do it. I won't be involved in removing the features anymore.

glistening:
> I don't think it's a bad smell. Rather, the two options you mentioned violate SRP for clean code. But if you want to remove it, do it. I won't be involved in removing the features anymore. Please just apply [#12111 (comment)](https://github.com/Samsung/ONE/pull/12111#issuecomment-1829035707).


I think `ir::train::operator` is one of the best example of the violation SRP.

`ir::train::operator` has both roles of `inference` and `training` by coping existing code.

As I wrote in this PR, `ir::train::operator` is nothing but `ir::operator` + `clone()`.

ragmani:
> Also, I don't think `OperationCloner` is same to `clone()`. `OperationCloner` is a visitor implementation while clone() can be used any time it want to make a clone for a specific operator.

The implementation way of both is different, but the purpose is the same. Both return `unique_ptr<IOperation>` cloned from an operation. Just, `clone()` only has the possibility that it can deep-copy perfectly. But, if you think keeping both is not a smell, I won't be involved in it too.

glistening:
> > Also, I don't think `OperationCloner` is same to `clone()`. `OperationCloner` is a visitor implementation while clone() can be used any time it want to make a clone for a specific operator.
> 
> The implementation way of both is different, but the purpose is the same. Both return `unique_ptr<IOperation>` cloned from an operation. Just, `clone()` only has the possibility that it can deep-copy perfectly. But, if you think keeping both is not a smell, I won't be involved in it too.

It may be the difference between you and me.

I don't want to make `visitor` to make `clone`. There is no reason to do double dispatch to make a clone.

Similarly, I don't want to introduce `ir::train::operation` and its related changes to make `Operation` clonnable.

ragmani:
> > I don't think it's a bad smell. Rather, the two options you mentioned violate SRP for clean code. But if you want to remove it, do it. I won't be involved in removing the features anymore. Please just apply [#12111 (comment)]

> As I wrote in this PR, ir::train::operator is nothing but ir::operator + clone().
clone() is not a feature of trainable operator in my view.

If `ir::train::operator` is removed, `ir::operator + clone()` will be duplicated with `OperationCloner` in purpose.

glistening:
@ragmani Moving `clone`() or using `OperatorCloner` is not the point of this issue. I just want to get opinion the direction of removing `ir::train::operator`. It is the reason I added `No Merge` ahead. I will use `OperatorCloner` instead of moving `clone()` if other changes are okay to you.

ragmani:
> Moving `clone()` or using `OperatorCloner` is not the point of this issue. I just want to get opinion the direction of removing `ir::train::operator`

I know it. If `ir::train::operation` is removed, `ir::operation` + `clone()` will be duplicated with `OperationCloner` in purpose. So I mentioned it. However, if you think keeping both is not a smell, I won't be involved in it too.

ragmani:
> It is the reason I added `No Merge` ahead. I will use `OperatorCloner` instead of moving `clone()` if other changes are okay to you.

In my opinion, other changes are not okay, but I won't be involved in removing `ir::train::operation` anymore.

glistening:
As @ragmani suggested, I removed `clone()` from `IOperation` (though it is beyond this PR in my view). It now uses `OperationCloner` instead of previous `ITrainableOperation::cone()`.

Issue #12100, opened 2023-11-23
[luci] Extract compute codes from luci/pass to luci-compute


seanshpark:
For first step, let's relocate computing codes in pass for constant folding to luci-compute module.

Objective of this task
- remove direct dependency with tflite kernels; headers, structures, etc
- provide constant folding
- provide shape inference
- ...

For
- DepthwiseConv2D
- FullyConnected

seanshpark:
for #10088

seanshpark:
shape inference for `DepthwiseConv2D` has common `infer_conv2d_type()` with `Conv2D`
- not sure its worth moving to `luci-compute` for these two
- but can real result while doing so...

seanshpark:
`FoldFullyConnectedPass` looks stopped handling `keep_num_dims`

seanshpark:
Getting min/max from `fused_act_func` looks better moved into `luci-compute`
- looks better with `to_compute` to convert type than getting min/max from type in luci/pass
- but this needs own `FusedActFunc` type

seanshpark:
done for now

PR #12114, opened 2023-11-27
[luci-compute] Introduce tflite_shape

seanshpark:
This will introduce tflite_shape to convert shape.


seanshpark:
for #12100 , from  #12099

jinevening:
Convert.h/cpp seems too general. How about renaming into `ConvertShape`?

seanshpark:
> How about renaming into ConvertShape?

There will be convert also for `PaddingType`, `FullyConnectedWeightsFormat`

jinevening (approved):
LGTM

PR #12119, opened 2023-11-27
[infra/debian] Make link for libluci_compute

seanshpark:
This will add lib link for libluci_compute.so file.


mhs4670go:
Maybe `luci_compute` library should be installed first.

seanshpark:
will be rebased after  #12125 lands

mhs4670go (approved):
LGTM

PR #12123, opened 2023-11-27
Draft: update xnnpack

hseok-oh:
It reduces warning message.
- tizen build is not working yet


hseok-oh:
It will be handed with #12701

PR #12125, opened 2023-11-28
[luci-compute] Revise to be a library

seanshpark:
This will revise interface to be a libary and installed.


seanshpark:
for #12100 , from  #12099

mhs4670go (approved):
LGTM

PR #12135, opened 2023-11-28
[onert] Apply reduction type to Loss

jyoungyun:
This commit applies a reduction type to the Loss operation. It changes the loss output shape to match the batch size using a static shape inferer. The loss information is stored in the TrainingExecutor and the loss value is returned according to the reduction type.

### DIFF
--- a/compute/cker/include/cker/train/operation/Loss.h
+++ b/compute/cker/include/cker/train/operation/Loss.h
@@ -29,31 +29,31 @@ namespace cker
 namespace train
 {
 
+template <typename T> inline T square(T value) { return value * value; }
+
 template <typename T>
 inline void MSE(const Shape &y_pred_shape, const T *y_pred_data, const Shape &y_true_shape,
                 const T *y_true_data, const Shape &output_shape, T *output_data)
 {
-  // TODO Consider Reduction
-  if (output_shape != Shape{1})
-    throw std::runtime_error("cker::MSE: output_shape != Shape{1}");
+  if (output_shape.DimensionsCount() != 1)
+    throw std::runtime_error("cker::MSE: output dimension count should be 1");
+  if (output_shape.Dims(0) != y_pred_shape.Dims(0))
+    throw std::runtime_error("cker::MSE: output and y_pred do not have the same batch");
   if (y_pred_shape != y_true_shape)
     throw std::runtime_error("cker::MSE: y_pred_shape != y_true_shape");
 
-  const auto y_pred = MapAsMatrixWithLastDimAsRows(y_pred_data, y_pred_shape);
-  const auto y_true = MapAsMatrixWithLastDimAsRows(y_true_data, y_true_shape);
+  const auto batch = y_pred_shape.Dims(0);
+  const auto size = FlatSizeSkipDim(y_pred_shape, 0);
 
-  double squared_sum = 0.0f;
-  for (size_t c = 0; c < (size_t)y_pred.cols(); ++c)
+  for (int b = 0; b < batch; ++b)
   {
-    for (size_t r = 0; r < (size_t)y_pred.rows(); ++r)
+    float sum = 0.f;
+    for (int i = 0; i < size; ++i)
     {
-      double error = y_pred.coeff(r, c) - y_true.coeff(r, c);
-      squared_sum += (error * error);
+      sum += square(y_pred_data[b * size + i] - y_true_data[b * size + i]);
     }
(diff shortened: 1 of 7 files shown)


runtime/onert/core/src/exec/train/TrainableExecutor.cc
```diff
@@ -195,8 +195,16 @@ float TrainableExecutor::getLoss(const ir::IOIndex &pred_io_ind) const
   if (loss_ind.undefined())
     throw std::runtime_error{"Loss " + std::to_string(loss_ind.value()) + " is not defined."};
   backend::ITensor *tensor = _tensor_regs.getITensor(loss_ind);
-  auto loss_buf = reinterpret_cast<float *>(tensor->buffer());
-  return *loss_buf;
+  double sum = 0.f;
```
ragmani, line 198: (Optional) According to https://en.cppreference.com/w/cpp/language/floating_literal, the suffix "f" means `float`. For the literal meaning double, "f" must be removed such as "0.".
But there is no problem with the suffix "f".

runtime/onert/core/src/exec/train/TrainableExecutor.cc
```diff
@@ -33,10 +33,10 @@ TrainableExecutor::TrainableExecutor(
   backend::train::TrainableBackendContexts &&backend_contexts,
   const compiler::train::TensorRegistries &tensor_regs,
   compiler::train::TrainableCodeMap &&code_map, const std::vector<ir::OperationIndex> &order,
-  const util::TracingCtx *tracing_ctx)
+  const util::TracingCtx *tracing_ctx, const ir::train::LossInfo &loss_info)
```
ragmani, line 36: (Optional) It would be better to reflect `LossInfo` to the `TrainableCodeMap` for Loss op instead of adding it as a param here.

ragmani (dismissed):
LGTM

runtime/onert/core/src/exec/train/TrainableExecutor.cc
jyoungyun: @ragmani 
It's a good suggestion. I will try to use `TrainableCodeMap` for `LossInfo`. I will take a closer look and update it again.

jyoungyun: Well, when I check the code, the `TrainableCodeMap` is a map structure that has `TrainableCodeAndInfo` for each operation. I don't think `LossInfo` is information related to each operation. Rahter, it is related to `TrainableExecutor` getting loss value. So in this case, I think the current location of `LossInfo` is good. What do you think about it?

ragmani: I think it's better to reflect the whole `LossInfo` to the kernels for Loss op if `LossInfo` is not changed. So I guess loss values can be reflected in output of Loss if reduction type is not changed.
Anyway, it's an optional suggestion.

ragmani (dismissed):

ragmani (approved):

hseok-oh:
```
[2023-11-29T01:01:50.479Z] [ RUN      ] CKer_Operation.neg_LossMSE
[2023-11-29T01:01:50.479Z] double free or corruption (out)
[2023-11-29T01:01:50.479Z] /jenkins/aws/workspace/nnfw/master/pr-onert-cross-debug/Product/out/test/command/unittest: line 78:   520 Aborted                 (core dumped) $TEST_BIN $(get_gtest_option)
[2023-11-29T01:01:50.480Z] Product/out/unittest/test_cker failed... return code: 134
```

compute/cker/include/cker/train/operation/Loss.h
```diff
@@ -29,31 +29,31 @@ namespace cker
 namespace train
 {
 
+template <typename T> inline T square(T value) { return value * value; }
```
ragmani, line 32: (Optional) How about using `std::pow`?

ragmani (approved):
LGTM

compute/cker/include/cker/train/operation/Loss.h
ragmani: `std::pow(x, 2)` is faster than `x*x` in your link.

ragmani: > std::pow(x, 2) is faster than x*x in your link.

Sorry, I was mistaken.

hseok-oh (approved):

PR #12136, opened 2023-11-28
[luci/import] Revise validate_minmax with int16

seanshpark:
This will revise validate_minmax method to support INT16 quantized model.


mhs4670go (approved):
LGTM

PR #12137, opened 2023-11-28
[cker] Introduce FloorMod kernels

ragmani:
This commit Introduces kernels for FloorMod operator.
  - Support element-wise FloorMod in general data type
  - Support broadcast FloorMod in general data type
  - Introduce validating introduced kernels

### DIFF
--- a/compute/cker/src/FloorMod.test.cc
+++ b/compute/cker/src/FloorMod.test.cc
@@ -1,0 +1,155 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


jyoungyun (approved):
LGTM

compute/cker/src/FloorMod.test.cc
```diff
@@ -0,0 +1,155 @@
...
+      ASSERT_EQ(output[i], expected_output[i]);
+  }
+
+  // Simple Float
+  {
+    // Shape: {1, 2, 2, 1}
+    std::vector<float> input1 = {10, 9, 11, 3};
+    // Shape: {1, 2, 2, 1}
+    std::vector<float> input2 = {2, 2, 3, 4};
+    std::vector<float> expected_output = {0, 1, 2, 3};
```
glistening, line 97: (optional) It may be better to cover float numbers like:
```suggestion
    std::vector<float> input1 = {10.5, 5.5, 11, 3};
    // Shape: {1, 2, 2, 1}
    std::vector<float> input2 = {2, 3.2, 3, 4};
    std::vector<float> expected_output = {0.5, 2.3, 2, 3};
```

glistening (approved):
LGTM

compute/cker/src/FloorMod.test.cc
ragmani: This test was from tensorflow. I don't want to change it. I will add real number test with point to one_op_tests instead.

PR #12138, opened 2023-11-28
[luci-compute] Introduce initial DepthwiseConv2D

seanshpark:
This will introduce initial compute code for DepthwiseConv2D.

### DIFF
--- a/compiler/luci-compute/include/luci_compute/DepthwiseConv2D.h
+++ b/compiler/luci-compute/include/luci_compute/DepthwiseConv2D.h
@@ -1,0 +1,81 @@
+/* Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_COMPUTE_DEPTHWISE_CONV2D_H__
+#define __LUCI_COMPUTE_DEPTHWISE_CONV2D_H__
+
+#include "Types.h"
+
+#include <loco/IR/TensorShape.h>
+
+namespace luci
+{
+namespace compute
+{
+
+// TODO extract some common for multiple Ops
+class DepthwiseConv2D
+{
+public:
+  DepthwiseConv2D() = default;
+
+public:
+  DepthwiseParams &params(void) { return _params; }
+
+  void input(const loco::TensorShape &shape, const float *data)
+  {
+    _input_shape = shape;
+    _input_data = data;
+  }
+
+  void filter(const loco::TensorShape &shape, const float *data)
+  {
+    _filter_shape = shape;
+    _filter_data = data;
+  }
+
+  void bias(const loco::TensorShape &shape, const float *data)
+  {
+    _bias_shape = shape;
+    _bias_data = data;
+  }
+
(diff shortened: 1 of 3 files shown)


seanshpark:
for #12100 , from  #12099

compiler/luci-compute/src/DepthwiseConv2D.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  _output_shape.rank(4);
+  _output_shape.dim(0) = input_batches;
+  _output_shape.dim(1) = output_height;
+  _output_shape.dim(2) = output_width;
+  _output_shape.dim(3) = filter_channels_out;
+
+  // TODO compute padding
+  _params.padding_values.height = 0;
+  _params.padding_values.width = 0;
```
jinevening, line 70: Is it ok to hard-code this?

compiler/luci-compute/src/DepthwiseConv2D.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  if (_params.depth_multiplier != static_cast<int32_t>(filter_channels_out / input_depth))
+    return false; // wrong depth multiplier value
+
+  if (_bias_shape.dim(0).value() != filter_channels_out)
+    return false; // unsupported bias value
+
+  // TODO compute output
+  auto output_height = 0;
+  auto output_width = 0;
```
jinevening, line 60: Are you going to update this code?

compiler/luci-compute/src/DepthwiseConv2D.cpp
seanshpark: There will be actual computation in next PR.
I've left `// TODO compute padding` comment.
If you are curious about the next change, please look at #12099

compiler/luci-compute/src/DepthwiseConv2D.cpp
seanshpark: Yes, `// TODO compute output` comment is for that.

seanshpark:
> This code is very similar to luci-interpreter codes. Is it necessary to re-implement?

The situation: 
- please read  #12100
- luci/pass already has these codes.
- first step is to relocate luci/pass to luci-compute, including this PR
- relocating itself cannot be just copy, as current luci/pass codes are somewhat complicated.
- next step (not sure when) is to refactor luci-interpreter so that to use common computation codes of luci-compute.
- refactoring luci-inpterpreter may not be easy cause of multiple computation backends

Why do we have this situation?
- some of our contributes from the past made this situation with duplicate codes

seanshpark:
> I think if you're going to re-implement new kernels, 

Not re-implementation. I's relocating the luci/pass codes with some cutting the complicated codes in current luci/pass copes with some APIs.

> we need to test them as luci-interpreter does.

That will be in another step. This and upcoming PRs are relocating so that these codes itself is going to be tested from existing luci/pass and other integration tests.

Currently adding also unit tests is too much for me.

jinevening:
Thanks for the replies. When I first see this PR, I thought that code is quite different from the one in the original pass, so I didn't recognize that it's just relocation. I understand it is difficult to extract computation code from the original code. I'll approve the PR as it would have been tested on the draft.

`luci-compute` is a separate module (independent of luci/pass), so I think unittest for each kernel is necessary. I made https://github.com/Samsung/ONE/issues/12150 for that task.

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM

compiler/luci-compute/include/luci_compute/DepthwiseConv2D.h
```diff
@@ -0,0 +1,81 @@
...
+  void output(float *data) { _output_data = data; }
+
+public:
+  bool prepare(void);
+  const loco::TensorShape &outputShape(void) const { return _output_shape; }
+  void compute(void);
+
+private:
+  // param to pass to compute kernel
+  DepthwiseParams _params;
```
mhs4670go, line 64: @seanshpark Our static analysis tool complains about this line for not initializing `_params` inside the default constructor. And so does `FullyConnected.h`.

seanshpark: I'll prepare a change;

PR #12143, opened 2023-11-29
[onert] Introduce backwarding graph order for training

jyoungyun:
This commit introduces backwarding graph order for training. It starts with the Loss op and find the backwarding order.

### DIFF
--- a/runtime/onert/core/src/exec/train/TrainableExecutor.cc
+++ b/runtime/onert/core/src/exec/train/TrainableExecutor.cc
@@ -32,9 +32,13 @@ TrainableExecutor::TrainableExecutor(
   std::unique_ptr<compiler::train::LoweredTrainableGraph> lowered_graph,
   backend::train::TrainableBackendContexts &&backend_contexts,
   const compiler::train::TensorRegistries &tensor_regs,
-  compiler::train::TrainableCodeMap &&code_map, const std::vector<ir::OperationIndex> &order,
-  const util::TracingCtx *tracing_ctx, const ir::train::LossInfo &loss_info)
-  : _lowered_graph{std::move(lowered_graph)}, _backend_contexts{std::move(backend_contexts)},
+  compiler::train::TrainableCodeMap &&code_map,
+  const std::vector<ir::OperationIndex> &forward_order,
+  const std::vector<ir::OperationIndex> &backward_order, const util::TracingCtx *tracing_ctx,
+  const ir::train::LossInfo &loss_info)
+  : _code_map{std::move(code_map)}, _forward_order{std::move(forward_order)},
+    _backward_order{std::move(backward_order)}, _lowered_graph{std::move(lowered_graph)},
+    _backend_contexts{std::move(backend_contexts)},
     _trainable_graph{_lowered_graph->trainable_graph()}, _tensor_regs{std::move(tensor_regs)},
     _mutex(), _tracing_ctx(tracing_ctx), _loss_info(loss_info)
 {
@@ -50,12 +54,6 @@ TrainableExecutor::TrainableExecutor(
   };
   build_tensor_list(_trainable_graph.getInputs(), _input_tensors);
   build_tensor_list(_trainable_graph.getOutputs(), _output_tensors);
-
-  for (auto &&index : order)
-  {
-    auto &trainable_code = code_map.at(index);
-    _code.emplace_back(std::move(trainable_code));
-  }
 }
 
 void TrainableExecutor::execute(const std::vector<backend::IPortableTensor *> &,
@@ -110,8 +108,9 @@ void TrainableExecutor::forwardImpl(bool training)
     auto profiling_subg_index = _tracing_ctx->getSubgraphIndex(&_trainable_graph.graph());
 
     _subject.notifySubgraphBegin(profiling_subg_index);
-    for (auto &&code : _code)
+    for (auto &&index : _forward_order)
     {
+      const auto &code = _code_map.at(index);
       const auto backend = code.lower_info->backend();
 // TODO : Move ruy profiler into ExecutionObserver
 #ifdef RUY_PROFILER
@@ -128,8 +127,9 @@ void TrainableExecutor::forwardImpl(bool training)
   }
   else
   {
-    for (auto &&code : _code)
+    for (auto &&index : _forward_order)
     {
+      const auto &code = _code_map.at(index);
 // TODO : Move ruy profiler into ExecutionObserver
 #ifdef RUY_PROFILER
       ruy::profiler::ScopeLabel label(code.op->name());
@@ -157,9 +157,9 @@ void TrainableExecutor::backwardImpl(uint32_t training_step)
     auto profiling_subg_index = _tracing_ctx->getSubgraphIndex(&_trainable_graph.graph());
 
     _subject.notifySubgraphBegin(profiling_subg_index);
-    for (auto it = _code.rbegin(); it != _code.rend(); ++it)
+    for (auto &&index : _backward_order)
     {
-      const auto &code = *it;
+      const auto &code = _code_map.at(index);
       const auto backend = code.lower_info->backend();
 // TODO : Move ruy profiler into ExecutionObserver
 #ifdef RUY_PROFILER
@@ -176,9 +176,9 @@ void TrainableExecutor::backwardImpl(uint32_t training_step)
   }
   else
   {
-    for (auto it = _code.rbegin(); it != _code.rend(); ++it)
+    for (auto &&index : _backward_order)
     {
-      const auto &code = *it;
+      const auto &code = _code_map.at(index);
 // TODO : Move ruy profiler into ExecutionObserver
 #ifdef RUY_PROFILER
       ruy::profiler::ScopeLabel label(code.op->name());
--- a/runtime/onert/core/src/exec/train/TrainableExecutor.h
+++ b/runtime/onert/core/src/exec/train/TrainableExecutor.h
@@ -49,7 +49,8 @@ public:
                     backend::train::TrainableBackendContexts &&backend_contexts,
                     const compiler::train::TensorRegistries &tensor_regs,
                     compiler::train::TrainableCodeMap &&code_map,
-                    const std::vector<ir::OperationIndex> &order,
+                    const std::vector<ir::OperationIndex> &forward_order,
+                    const std::vector<ir::OperationIndex> &backward_order,
                     const util::TracingCtx *tracing_ctx, const ir::train::LossInfo &training_info);
 
 public:
@@ -90,7 +91,9 @@ private:
   void backwardImpl(uint32_t training_step);
 
 private:
-  std::vector<compiler::train::TrainableCodeAndInfo> _code;
+  compiler::train::TrainableCodeMap _code_map;
+  std::vector<ir::OperationIndex> _forward_order;
+  std::vector<ir::OperationIndex> _backward_order;
   ExecutionObservee _subject;
   std::shared_ptr<ir::OperationIndexMap<int64_t>> _indexed_ranks;
   std::unique_ptr<compiler::train::LoweredTrainableGraph> _lowered_graph;
--- a/runtime/onert/core/include/ir/train/TrainableGraph.h
+++ b/runtime/onert/core/include/ir/train/TrainableGraph.h
@@ -129,7 +129,7 @@ public:
 
 public:
   std::vector<ir::OperationIndex> topolSortOperations() const;
-  // TODO Support topological sort for backwarding
+  std::vector<ir::OperationIndex> btopolSortOperations() const;
 
 private:
   Graph _graph;
--- a/runtime/onert/core/src/compiler/ExecutorFactory.cc
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.cc
@@ -749,9 +749,15 @@ exec::IExecutor *ExecutorFactory::createTrainableExecutor(
     (lowered_graph->graph().getInputs() + lowered_graph->graph().getOutputs()) |
       ir::Remove::DUPLICATED | ir::Remove::UNDEFINED);
 
-  // linearize
+  // linearize for forwarding
   auto order = Linear::linearize(*lowered_graph);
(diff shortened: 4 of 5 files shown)


runtime/onert/core/include/compiler/train/TrainableCodeMap.h
```diff
@@ -35,12 +35,12 @@ struct TrainableCodeAndInfo
...
   const OperationLowerInfo *lower_info;
   // TODO Change to TrainableFnSequence
-  std::unique_ptr<exec::train::TrainableFnSequence> tn_seq;
+  std::shared_ptr<exec::train::TrainableFnSequence> tn_seq;
 
   TrainableCodeAndInfo(const ir::OperationIndex op_ind, const ir::train::ITrainableOperation *op,
                        const OperationLowerInfo *lower_info,
-                       std::unique_ptr<exec::train::TrainableFnSequence> &&tn_seq)
-    : op_ind{op_ind}, op{op}, lower_info{lower_info}, tn_seq{std::move(tn_seq)}
+                       std::shared_ptr<exec::train::TrainableFnSequence> &&tn_seq)
```
ragmani, line 42:
```suggestion
                       std::shared_ptr<exec::train::TrainableFnSequence> tn_seq)
```

jyoungyun:
```
[ ExecutorFactory] Linearize for forwarding order
[    Linearize   ] %10 = @5_Permute(%0)
[    Linearize   ] %4 = @0_Reshape(%10, %2)
[    Linearize   ] %5 = @1_FullyConnected(%4, %3, %1)
[    Linearize   ] %6 = @2_Softmax(%5)
[    Linearize   ] %9 = @4_Permute(%6)
[    Linearize   ] %11 = @6_Permute(%7)
[    Linearize   ] %8 = @3_MeanSquaredErrorLoss(%6, %11)
[ ExecutorFactory] Linearize for backwarding order
[    Linearize   ] %8 = @3_MeanSquaredErrorLoss(%6, %11)
[    Linearize   ] %6 = @2_Softmax(%5)
[    Linearize   ] %5 = @1_FullyConnected(%4, %3, %1)
[    Linearize   ] %4 = @0_Reshape(%10, %2)
[    Linearize   ] %10 = @5_Permute(%0)
[    Linearize   ] %11 = @6_Permute(%7)
```

runtime/onert/core/include/ir/Graph.h
```diff
@@ -119,6 +119,7 @@ class Graph : public IGraph
   // Topological sort
 public:
   std::vector<ir::OperationIndex> topolSortOperations() const;
+  std::vector<ir::OperationIndex> btopolSortOperations() const;
```
glistening, line 122: `Graph` is used for `ir::operation`. `btopolSortOperations` is only for training. I don't want to add this in `ir::Graph`.

runtime/onert/core/src/exec/train/TrainableExecutor.cc
```diff
@@ -33,7 +33,7 @@ TrainableExecutor::TrainableExecutor(
   backend::train::TrainableBackendContexts &&backend_contexts,
   const compiler::train::TensorRegistries &tensor_regs,
   compiler::train::TrainableCodeMap &&code_map, const std::vector<ir::OperationIndex> &order,
-  const util::TracingCtx *tracing_ctx)
+  const std::vector<ir::OperationIndex> &border, const util::TracingCtx *tracing_ctx)
```
glistening, line 36: _border_ is existing word. It would be good to use other name.

runtime/onert/core/src/exec/train/TrainableExecutor.h
```diff
@@ -90,6 +91,7 @@ class TrainableExecutor : public IExecutor
 
 private:
   std::vector<compiler::train::TrainableCodeAndInfo> _code;
+  std::vector<compiler::train::TrainableCodeAndInfo> _bcode;
```
glistening, line 94: I don't understand why `_bcode` is needed. Currently `TrainableFnSequence` already has `backward()`. What is the difference?

jyoungyun: The `TrainableExecutor` uses `_code` to call the `forward()` and `backward()` functions. The `_code` contains the order in which operation is called. This commit introduces `_bcode` for backwarding order. The current onert calls the backwarding function in the reverse direction of `_code` order. However, the backwarding should start with `Loss` operator. So this pr searches for a new order for backwarding.

runtime/onert/core/include/ir/Graph.h
ragmani: @jyoungyun How about moving it to `TrainableGraph`?

jyoungyun: Okay, it's a good idea. However even though this function is moved to the `TrainableGraph`, the Linear still needs the `blinearize()` function. I'm worried about it...

jyoungyun: @ragmani I fixed this PR. I do not make a wrapper function for `btopolSortOperations()` in `Linear` class.

ragmani: Hm... I think it's okay to add it to there.
If we have to put it in a train directory, how about introducing a new `Linear` class for training?

runtime/onert/core/src/exec/train/TrainableExecutor.h
glistening: I don't understand fully in detail. Could you please show me the full draft, which uses the added members and methods?

jyoungyun: There is no draft code. This PR works completely with this PR alone. Use of this member can be found in the `TrainableExecutor.cc` file.

ragmani (dismissed):
LGTM

runtime/onert/core/src/exec/train/TrainableExecutor.h
glistening: I think you want to traverse in backward order for the same code in `_code`. Then, backward order iterator is what you need as I understand. Is it right?

If so, we may add `iterate()` and `biterate()` in `TrainableCodeMap` (preferred but it is currently just an alias) or `TrainableCodeAndInfo`.

glistening: I wrote here I explained offline:

https://github.com/Samsung/ONE/blob/e48b702468fc683deafbbde69edb960ad7eb2638/runtime/onert/core/include/util/ObjectManager.h#L186-L192

You can introduce `back_iterate`. In this method, you can make backward ordering and iterate and call your function.

glistening (changes requested):
If you just want to traverse in backward order with the same `TrainableCodeandInfo`, adding another `TrainableCodeInfo` member of `_bcode` seems overkill.

glistening:
@jyoungyun told me `_code` is not a class, but just a `std::vector(TrainableCodeAndInfo)`. There is no way to add method. Let me think more. Or anyone can suggest better approach.

glistening:
I would like to roll back `unique_ptr` → `shared_ptr` in `TrainableCodeMap.h`.
We don't need to make `shared_ptr`, `backward_order` is enough.

Also, `backward_order` is required in `backwardImpl()` only.
It can be created in `backwardImpl()`.

```diff
diff --git a/runtime/onert/core/include/compiler/train/TrainableCodeMap.h b/runtime/onert/core/include/compiler/train/TrainableCodeMap.h
index 6c754d48c..35915b507 100644
--- a/runtime/onert/core/include/compiler/train/TrainableCodeMap.h
+++ b/runtime/onert/core/include/compiler/train/TrainableCodeMap.h
@@ -35,12 +35,12 @@ struct TrainableCodeAndInfo
   const ir::train::ITrainableOperation *op;
   const OperationLowerInfo *lower_info;
   // TODO Change to TrainableFnSequence
-  std::shared_ptr<exec::train::TrainableFnSequence> tn_seq;
+  std::unique_ptr<exec::train::TrainableFnSequence> tn_seq;
 
   TrainableCodeAndInfo(const ir::OperationIndex op_ind, const ir::train::ITrainableOperation *op,
                        const OperationLowerInfo *lower_info,
-                       std::shared_ptr<exec::train::TrainableFnSequence> tn_seq)
-    : op_ind{op_ind}, op{op}, lower_info{lower_info}, tn_seq{tn_seq}
+                       std::unique_ptr<exec::train::TrainableFnSequence> &&tn_seq)
+    : op_ind{op_ind}, op{op}, lower_info{lower_info}, tn_seq(std::move(tn_seq))
   {
   }
 };
diff --git a/runtime/onert/core/src/compiler/ExecutorFactory.cc b/runtime/onert/core/src/compiler/ExecutorFactory.cc
index c20d709da..50ee8ea9f 100644
--- a/runtime/onert/core/src/compiler/ExecutorFactory.cc
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.cc
@@ -891,7 +891,6 @@ exec::IExecutor *ExecutorFactory::createTrainableExecutor(
                                                  tensor_regs,
                                                  std::move(code_map),
                                                  order,
-                                                 backward_order,
                                                  tracing_ctx};
 
   if (!options->trace_filepath.empty())
diff --git a/runtime/onert/core/src/exec/train/TrainableExecutor.cc b/runtime/onert/core/src/exec/train/TrainableExecutor.cc
index 3972c185e..ed010cb24 100644
--- a/runtime/onert/core/src/exec/train/TrainableExecutor.cc
+++ b/runtime/onert/core/src/exec/train/TrainableExecutor.cc
@@ -33,7 +33,7 @@ TrainableExecutor::TrainableExecutor(
   backend::train::TrainableBackendContexts &&backend_contexts,
   const compiler::train::TensorRegistries &tensor_regs,
   compiler::train::TrainableCodeMap &&code_map, const std::vector<ir::OperationIndex> &order,
-  const std::vector<ir::OperationIndex> &backward_order, const util::TracingCtx *tracing_ctx)
+  const util::TracingCtx *tracing_ctx)
   : _lowered_graph{std::move(lowered_graph)}, _backend_contexts{std::move(backend_contexts)},
     _trainable_graph{_lowered_graph->trainable_graph()}, _tensor_regs{std::move(tensor_regs)},
     _mutex(), _tracing_ctx(tracing_ctx)
@@ -53,14 +53,7 @@ TrainableExecutor::TrainableExecutor(
 
   for (auto &&index : order)
   {
-    auto &trainable_code = code_map.at(index);
-    _code.emplace_back(trainable_code);
-  }
-
-  for (auto &&index : backward_order)
-  {
-    auto &trainable_code = code_map.at(index);
-    _bcode.emplace_back(trainable_code);
+    _code.emplace_back(std::move(code_map.at(index)));
   }
 }
 
@@ -158,13 +151,16 @@ void TrainableExecutor::backward(const IODescription &, uint32_t training_step)
 
 void TrainableExecutor::backwardImpl(uint32_t training_step)
 {
+  auto backward_order = _trainable_graph.btopolSortOperations();
   if (_tracing_ctx)
   {
     auto profiling_subg_index = _tracing_ctx->getSubgraphIndex(&_trainable_graph.graph());
 
     _subject.notifySubgraphBegin(profiling_subg_index);
-    for (auto &&code : _bcode)
+
+    for (auto &&index : backward_order)
     {
+      auto &code = _code.at(index.value());
       const auto backend = code.lower_info->backend();
 // TODO : Move ruy profiler into ExecutionObserver
 #ifdef RUY_PROFILER
@@ -181,8 +177,9 @@ void TrainableExecutor::backwardImpl(uint32_t training_step)
   }
   else
   {
-    for (auto &&code : _bcode)
+    for (auto &&index : backward_order)
     {
+      auto &code = _code.at(index.value());
 // TODO : Move ruy profiler into ExecutionObserver
 #ifdef RUY_PROFILER
       ruy::profiler::ScopeLabel label(code.op->name());
diff --git a/runtime/onert/core/src/exec/train/TrainableExecutor.h b/runtime/onert/core/src/exec/train/TrainableExecutor.h
index c2296b4e9..6b645305f 100644
--- a/runtime/onert/core/src/exec/train/TrainableExecutor.h
+++ b/runtime/onert/core/src/exec/train/TrainableExecutor.h
@@ -49,7 +49,6 @@ public:
                     const compiler::train::TensorRegistries &tensor_regs,
                     compiler::train::TrainableCodeMap &&code_map,
                     const std::vector<ir::OperationIndex> &order,
-                    const std::vector<ir::OperationIndex> &backward_order,
                     const util::TracingCtx *tracing_ctx);
 
 public:
@@ -91,7 +90,6 @@ private:
 
 private:
   std::vector<compiler::train::TrainableCodeAndInfo> _code;
-  std::vector<compiler::train::TrainableCodeAndInfo> _bcode;
   ExecutionObservee _subject;
   std::shared_ptr<ir::OperationIndexMap<int64_t>> _indexed_ranks;
   std::unique_ptr<compiler::train::LoweredTrainableGraph> _lowered_graph;
```

runtime/onert/core/include/compiler/train/TrainableCodeMap.h
```diff
@@ -35,12 +35,12 @@ struct TrainableCodeAndInfo
   const ir::train::ITrainableOperation *op;
   const OperationLowerInfo *lower_info;
   // TODO Change to TrainableFnSequence
-  std::unique_ptr<exec::train::TrainableFnSequence> tn_seq;
+  std::shared_ptr<exec::train::TrainableFnSequence> tn_seq;
```
glistening, line 38: I don't want to change `unique_ptr` to `shared_ptr` for back iteration.

ragmani:
> Also, `backward_order` is required in backwardImpl() only.
> It can be created in `backwardImpl()`.

I think it would be better to move parts that affect performance(latency) to compilation of `onert` if possible. So please don't call topological sort in `backwardImpl()`.

glistening:
@ragmani Good point. It would be better to construct ordering before `backwardImpl()`.

Anyway, my main point is that I don't want to introduce `_bcode` when we only need backward_ordering.

jyoungyun:
@glistening

> Anyway, my main point is that I don't want to introduce `_bcode` when we only need backward_ordering.

I made some changes about this. I didn't copy the `TrainableCodeAndInfo` for backwarding and store the forwarding and backwarding orders by sharing the same `code_map`.

glistening (approved):
LGTM :+1:

PR #12145, opened 2023-11-29
[cker] Make Mul support bool type

ragmani:
This commit makes the kernel for Mul support boolean type too.
  - Add mul op unit tests
  - Support bool type

### DIFF
--- a/compute/cker/src/Mul.test.cc
+++ b/compute/cker/src/Mul.test.cc
@@ -1,0 +1,210 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cker/operation/BinaryArithmeticOps.h>
+
+#include <gtest/gtest.h>
+#include <vector>
+
+TEST(CKer_Operation, Mul)
+{
+  // Simple
+  {
+    // Shape: {1, 2, 2, 1}
+    std::vector<int32_t> input1 = {10, 9, 11, 3};
+    // Shape: {1, 2, 2, 1}
+    std::vector<int32_t> input2 = {2, 2, 3, 4};
+    std::vector<int32_t> expected_output = {20, 18, 33, 12};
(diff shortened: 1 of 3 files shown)


compute/cker/src/Mul.test.cc
```diff
@@ -0,0 +1,210 @@
+/*
+ * Copyright (c) 2019 Samsung Electronics Co., Ltd. All Rights Reserved
```
ragmani, line 2:
```suggestion
 * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
```

glistening:
@ragmani At first, I thought you are introducing a new data type of `Mul bool`. Your commit message body seems to describe better. Could you please update commit title like _"[cker] Make Mul support bool type"_ ?

ragmani:
> At first, I thought you are introducing a new data type of Mul bool. Your commit message body seems to describe better. Could you please update commit title like "[cker] Make Mul support bool type" ?

Changed it.

jyoungyun (approved):
LGTM

glistening:
@ragmani 

Mul for bool type seems to be same to `LogicalAnd`. If so, why don't you use the existing `LogicalAnd.h`?

`LogicalAnd` does not require function call for each elementwise `and` operation while this PR requires lambda.

### This PR
```cc
return [](const bool &a, const bool &b) -> bool { return a && b; };
```

### LogicalAnd
```cc
template <typename T>
inline void LogicalAndElementwise(const Shape &shape, const T *input1_data, const T *input2_data,
                                  T *output_data)
{

  int num_elements = shape.FlatSize();

  for (int t = 0; t < num_elements; t++)
  {
    output_data[t] = input1_data[t] && input2_data[t];
  }
}
```

ragmani:
> Mul for bool type seems to be same to LogicalAnd. If so, why don't you use the existing LogicalAnd.h?

> LogicalAnd does not require function call for each elementwise and operation while this PR requires lambda.

I had no idea about existence of `LogocalAnd` kernel.
Most `Mul` ops use `BinaryArithmetic` kernels that can process `BinaryArithmeticOpParam`, but `LogicalAnd` does not. I think it would be better to use a `BinaryArithmetic` kernel for consistency and maintenance even if the `BinaryArithmetic` kernel does not process `BinaryArithmeticOpParam` now.

glistening:
@ragmani 
I have different view. `cker` is our abstraction of kernel computation. `cker` is there for code reuse. If same computation is required, there is no reason to have duplicated `cker` code. More code requires more maintenance cost. In this case, you added test case for bool type Mul. 

Furthermore, it is likely to add performance penalty as I mentioned above. (We don't need lambda for logical and.)

ragmani:
@glistening 
If other members are agreed with you, I will change `Mul` bool op to use `LogicalAnd` instead of `BinaryArithmetic`.

@Samsung/one_onert Please give me your opinions.

hseok-oh:
>Please give me your opinions.

No matter which one is selected.

glistening:
I am waiting for 1 additional approval.

hseok-oh (approved):

PR #12141, opened 2023-11-29
[onert] Introduce LossCategoricalCrossentropy layer

jyoungyun:
This commit introduces LossCategoricalCrossentropy layer and param structure.

ONE-DCO-1.0-Signed-off-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #12151, opened 2023-11-30
[onert] Introduces LossCategoricalCrossentropyInfo structure

jyoungyun:
This commit introduces LossCategoricalCrossentropyInfo structure instead of using union.


jyoungyun:
@ragmani 

I implemented it from [your comment](https://github.com/Samsung/ONE/pull/12141#discussion_r1408708279). Did I understand correctly what you said?

runtime/onert/backend/train/KernelGenerator.cc
```diff
@@ -257,9 +256,11 @@ void KernelGenerator::visit(const ir::train::operation::Loss &node)
     }
     case ir::train::LossCode::CategoricalCrossentropy:
     {
+      auto loss_info =
+        reinterpret_cast<const ir::train::LossCategoricalCrossentropyInfo &>(node.param());
```
ragmani, line 260:
```suggestion
        nnfw::misc::polymorphic_downcast<const ir::train::LossCategoricalCrossentropyInfo &>(node.param());
```

ragmani:
@jyoungyun 

> I implemented it according to https://github.com/Samsung/ONE/pull/12141#discussion_r1408708279. Did I understand correctly what you said?

The way I was thinking of applying polymorphism was to applying polymorphism to `LossParam`. But it's optional. I think this way is also good.

runtime/onert/backend/train/KernelGenerator.cc
jyoungyun: I tried to this downcast function but it occured `source type is not polymorphic` error. I didn't think about it much. I'll take a closer look.

PR #12153, opened 2023-11-30
[luci-compute] DWConv2D compute output/padding

seanshpark:
This will enable to compute output and padding for DepthwiseConv2D.


seanshpark:
for #12100 , from  #12099

jinevening (approved):

mhs4670go (approved):
LGTM

PR #12155, opened 2023-11-30
[luci-compute] Introduce initial FullyConnected

seanshpark:
This will introduce initial compute code for FullyConnected.

### DIFF
--- a/compiler/luci-compute/include/luci_compute/FullyConnected.h
+++ b/compiler/luci-compute/include/luci_compute/FullyConnected.h
@@ -1,0 +1,86 @@
+/* Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


seanshpark:
for #12100 , from  #12099

compiler/luci-compute/include/luci_compute/FullyConnected.h
```diff
@@ -0,0 +1,86 @@
...
+  FullyConnectedParams _params;
+  // new param from tflite version 5
+  bool _keep_num_dims = false;
+  // shape and data for inputs
+  loco::TensorShape _input_shape;
+  loco::TensorShape _weights_shape;
+  loco::TensorShape _bias_shape;
+  const float *_input_data = nullptr;
+  const float *_weights_data = nullptr;
+  const float *_bias_data = nullptr;
```
jinevening, line 76: Q) Does luci-compute only support fp32 kernel?

seanshpark: Current luci-interpreter supports U8, S8 and FP32.
This FC implementation comes fro current luci/pass with only FP32.
We may expand with other dtypes later.

jinevening (approved):
LGTM

compiler/luci-compute/include/luci_compute/FullyConnected.h
seanshpark: thanks for the suggestion. I've added https://github.com/Samsung/ONE/issues/10088#issuecomment-1833136497 for future work :)

mhs4670go (approved):
LGTM

PR #12158, opened 2023-11-30
[luci-compute] Enable FullyConnected weights_format conversion

seanshpark:
This will enable FullyConnected weights_format conversion.

### DIFF
--- a/compiler/luci-compute/src/ConvertTypes.cpp
+++ b/compiler/luci-compute/src/ConvertTypes.cpp
@@ -16,6 +16,7 @@
 #include "ConvertTypes.h"
 
 #include <cassert>
+#include <stdexcept>
 
 namespace luci
 {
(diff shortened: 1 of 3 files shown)


seanshpark:
for #12100 , from  #12099

mhs4670go (approved):
LGTM

compiler/luci-compute/src/ConvertTypes.cpp
```diff
@@ -46,5 +46,18 @@ tflite::PaddingType tflite_padding(const PaddingType type)
...
 
+tflite::FullyConnectedWeightsFormat tflite_weights_format(const FullyConnectedWeightsFormat type)
+{
+  switch (type)
+  {
+    case FullyConnectedWeightsFormat::kDefault:
+      return tflite::FullyConnectedWeightsFormat::kDefault;
+    case FullyConnectedWeightsFormat::kShuffled4x16Int8:
+      return tflite::FullyConnectedWeightsFormat::kShuffled4x16Int8;
+  }
```
jinevening, line 57: Better to insert default case for error handling? The below assert would be ignored in release mode.

seanshpark: `FullyConnectedWeightsFormat` has only two. logically `default` will not work.

jinevening: Yes. That holds for now, but if a new format is added, the switch statement will miss the added format.

seanshpark: > the switch statement will miss the added format.

well, that will make us to revise all related codes not only here.

seanshpark: > The below assert would be ignored in release mode.

Do you think throw would be better?

My background of adding `assert` is that logically we assume other case won't exist here but just break at debug build just in case and also assume caller should have filtered it. for throwing, there may be a chance with that value but we won't/can't handle that.

jinevening: > Do you think throw would be better?

Yes. If a new format is added, then luci-compute may silently produce wrong results in release mode. In that case, it'll be difficult to find where is the problem among importer, optimizer, quantizer, and compiler.

> assert is that logically we assume other case won't exist here

The assumption would not hold anymore if a new format is added.

jinevening (approved):
LGTM Thank you!

mhs4670go (approved):
LGTM

PR #12159, opened 2023-11-30
[luci-compute] Revise to have FusedActFunc

seanshpark:
This will revise compute classes to have FusedActFunc as inputs.


seanshpark:
for https://github.com/Samsung/ONE/issues/12100#issuecomment-1833259371 , from #12099

PR #12166, opened 2023-11-30
[luci-compute] Fix case for output_shape method

seanshpark:
This will fix to snake_case for output_shape method.


jinevening (approved):
LGTM

Issue #12130, opened 2023-11-28
[onert] Remove frontend layout feature


ragmani:
## What

Let's remove frontend layout feature.

## Why
`onert` supports only NHWC model(circle, tflite) and probably will supports only NHWC model in the future. If so, the feature is meaningless.

## How
- Change default layout to `ir::Layout::UNKNOWN`
- Remove frontend_layout from `ir::Graph`.
- Remove layout from `compiler::PermuteFactor`.
- Remove layout from `compiler::OperationLowerInfo`.
- Add layout to `compiler::OperandLowerInfo`.
- Add layout to `ir::Operand` to support ops like `Conv2D` that has layout.
- Delegate dealing with layouts to backends.
  - Remove layout feature for each backend.
  - Enable each backend to update their partial graph

Draft #12224

glistening:
`onert` is not the only user of `circle` schema. Other tools (like ONE front end and/or NPU Compiler) may make use of `data_format`.

> onert supports only NHWC model(circle, tflite)

I am not sure it is valid. We've supported some model in `NHWC` at least for some operators a long time ago.

ragmani:
> Then, will we remove data_format field in circle schema?

Thanks for your information. I didn't know circle schema has `data_format`. I will stop this task.

ragmani:
@glistening 
> onert is not the only user of circle schema. Other tools (like ONE front end and/or NPU Compiler) may make use of data_format.

What is ONE front and NPU Compiler? `onert` get models from them?

> I am not sure it is valid. We've supported some model in NHWC at least for some operators a long time ago.

OK, I will stop this task.

glistening:
@ragmani 

> What is ONE front and NPU Compiler? onert get models from them? I didn't catch the reason why you are talking about other tools.

It was my comment about @hseok-oh's 

> Then, will we remove `data_format` field in `circle` schema?


> What is ONE front and NPU Compiler? 

CFE in this repo, and NPU_Compiler in our internal repo.

hseok-oh:
But if we don't use `data_format` field on compiler frontend and runtime (in other word, use `NHWC` only as default value), I think we can consider removing this feature.

ragmani:
> But if we don't use data_format field on compiler frontend and runtime (in other word, use NHWC only as default value),

I also felt that compiler frontend converts models(i.e. ONNX model) to circle models in `NHWC` when I was working on compiler frontend.

hseok-oh:
[...]

PR #12168, opened 2023-12-01
DRAFT CFe deprecate circle data_format

seanshpark:
on-going draft to deprecate circle data_format field.


seanshpark:
for #12130

seanshpark:
check done

PR #12162, opened 2023-11-30
[DRAFT] Introduce luci-value-py-test

mhs4670go:
This commit introduces luci-value-py-test to reduce long test duration.

Related: #12071
ONE-DCO-1.0-Signed-off-by: seongwoo <mhs4670go@naver.com>

Issue #12071, opened 2023-11-21
Compiler FE: Too slow ctest time


mhs4670go:
## What

I think the ctest test like `luci-value-test` takes too long.. We can resolve this with smaller input size of the recipes.

jinevening:
Input size is really a major reason of slow test? For similar tests, e.g., one-cmds test, the time taken for each test is similar regardless of input size of models. I guess that a major bottleneck is running python script which loads a huge framework (tensorflow) over and over again.

seanshpark:
yes, huge framework.. but we may do some duration testing with tensor size ...

mhs4670go:
@jinevening I see your point. Actually, I didn't think that much about this yet. I'll consider what you said. 

First, I list up take-too-long tests. Let's start reducing the duration of longest one - luci-value-test.

```
      Start 23: tflchef_test
23/76 Test #23: tflchef_test ...........................   Passed    1.29 sec
      Start 43: luci_unit_readtest
43/76 Test #43: luci_unit_readtest .....................   Passed    1.85 sec
      Start 44: luci_unit_writetest
44/76 Test #44: luci_unit_writetest ....................   Passed    1.89 sec
      Start 56: record_minmax_conversion_test
56/76 Test #56: record_minmax_conversion_test ..........   Passed   18.16 sec
      Start 57: q-implant-qparam-test
57/76 Test #57: q-implant-qparam-test ..................   Passed    3.36 sec
      Start 58: pota_fake_wquant_test
58/76 Test #58: pota_fake_wquant_test ..................   Passed    4.87 sec
      Start 59: pota_record_minmax_test
59/76 Test #59: pota_record_minmax_test ................   Passed    9.06 sec
      Start 60: pota_quantization_test
60/76 Test #60: pota_quantization_test .................   Passed    4.95 sec
      Start 61: pota_fake_wquant_test_with_config
61/76 Test #61: pota_fake_wquant_test_with_config ......   Passed    3.87 sec
      Start 62: pota_parallel_record_minmax_test
62/76 Test #62: pota_parallel_record_minmax_test .......   Passed    9.66 sec
      Start 63: pota_quantization_test_with_config
63/76 Test #63: pota_quantization_test_with_config .....   Passed    7.94 sec
      Start 64: pota_wo_quantization_test
64/76 Test #64: pota_wo_quantization_test ..............   Passed    0.61 sec
      Start 65: minmax_embedder_value_test
[...]

PR #12169, opened 2023-12-01
[common-artifacts] Install pytest

mhs4670go:
This commit installs pytest package into venv.

Draft: #12162
Related: #12071 


seanshpark (approved):
LGTM

PR #12170, opened 2023-12-01
[luci-compute] Revise to throw for invalid padding

seanshpark:
This will revise to throw for invalid input for tflite_padding method.


seanshpark:
from https://github.com/Samsung/ONE/pull/12158#discussion_r1411524548 suggestion

PR #12171, opened 2023-12-01
[luci-compute] Revise to set min/max from activation

seanshpark:
This will revise DepthwiseConv2D, FullyConnected to set activation min/max from activation function.


seanshpark:
for #12100 , from  #12099

mhs4670go (approved):
LGTM

jinevening (approved):
LGTM

PR #12173, opened 2023-12-01
[luci-value-py-test] Introduce pytest for luci value test

mhs4670go:
This commit introudces a pytest for luci value test.

Draft: https://github.com/Samsung/ONE/pull/12162
Related: https://github.com/Samsung/ONE/issues/12071

### DIFF
--- a/compiler/luci-value-py-test/requires.cmake
+++ b/compiler/luci-value-py-test/requires.cmake
@@ -1,1 +1,2 @@
 require("common-artifacts")
+require("luci-eval-driver")


seanshpark (dismissed):
LGTM

compiler/luci-value-py-test/requires.cmake
```diff
@@ -0,0 +1 @@
+require("common-artifacts")
```
jinevening, line 1:
```suggestion
require("common-artifacts")
require("luci-eval-driver")
```

PR #12175, opened 2023-12-03
[luci/pass] Introduce to_compute helper methods

seanshpark:
This will introduce to_compute conversion helper methods.

### DIFF
--- a/compiler/luci/pass/src/helpers/Compute.cpp
+++ b/compiler/luci/pass/src/helpers/Compute.cpp
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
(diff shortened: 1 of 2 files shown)


seanshpark:
for #12100 , from  #12099

compiler/luci/pass/src/helpers/Compute.cpp
```diff
@@ -0,0 +1,72 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Compute.h"
+
+#include <cassert>
+
```
jinevening, line 20: This header seems unused.

PR #12176, opened 2023-12-03
[luci/pass] Extract is_same_shape helper method

seanshpark:
This will extract is_same_shape helper method to separaet module.


seanshpark:
for #12100 , from  #12099

mhs4670go (approved):
LGTM

Issue #12148, opened 2023-11-29
[luci-interpreter] Enable SelectV2 Op 


ragmani:
## What

Let's enable SelectV2 Op from `Tensor.Expand` of ONNX model.

onnx.Where is circle.Select or circle.SelectV2

## Tasks
- [x] Remove broadcast from Select op
  - #12183
- [x] luci-interpreter
  - #12181
- [x] luci-value-test

Draft : #12149

ragmani:
Done

PR #12181, opened 2023-12-04
[luci-interpreter] Support SelectV2 op

ragmani:
This makes luci-interpreter support SelectV2 op.

---

For https://github.com/Samsung/ONE/issues/12148


compiler/luci-interpreter/src/kernels/Select.h
```diff
@@ -42,8 +42,6 @@ class Select : public Kernel
   void evalFloat() const;
 
 private:
-  // for SelectV2
-  bool _requires_broadcast = false;
```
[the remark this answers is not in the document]
ragmani: OK I will separate it.

seanshpark (dismissed):
LGTM thank you!

ragmani:
@jinevening PTAL

compiler/luci-interpreter/src/kernels/SelectV2.cpp
```diff
@@ -0,0 +1,97 @@
...
+      evalFloat();
+      break;
+    default:
+      throw std::runtime_error("luci-intp SelectV2 unsupported type.");
+  }
+}
+
+void SelectV2::evalFloat() const
+{
+
```
jinevening, line 73:
```suggestion
```

compiler/luci-interpreter/src/kernels/SelectV2.cpp
```diff
@@ -0,0 +1,97 @@
...
+  auto cond_shape = condition()->shape();
+  auto cond_num_dims = cond_shape.num_dims();
+  auto t_shape = t()->shape();
+  auto e_shape = e()->shape();
+
+  bool is_input_condition_scalar = cond_num_dims == 0;
+  bool has_rank_one_input_condition = cond_num_dims == 1 && cond_shape.dim(0) == t_shape.dim(0);
+
+  _has_low_rank_input_condition =
+    (is_input_condition_scalar || has_rank_one_input_condition) && t_shape == e_shape;
```
jinevening, line 53: This behavior is different from tflite kernel.

```suggestion
  _has_low_rank_input_condition =
    is_input_condition_scalar || has_rank_one_input_condition;
```

`t_shape == e_shape` is necessary? It seems that `has_low_rank_input_condition` is only used by select v1.

ragmani: > `t_shape == e_shape` is necessary?

Yes, it is necessary if select v2 kernel uses `RankOneSelect()`.

> `has_low_rank_input_condition` is only used by select v1.

I just implemented it following below conditions. 

### TFL SelectV2 op

https://www.tensorflow.org/mlir/tfl_ops#tflselect_v2_tflselectv2op

Select values of 'x' if the corresponding value of 'condition' is true or the value of 'y' if false. There are valid condition input sizes:

1. Either the same shape (in which case the select is elementwise), or
2. Broadcastable shapes between 'condition', 'x' and 'y'.

Anyway, I will change this kernel to use only `BroadcastSelect4DSlow()` as your advice.

compiler/luci-interpreter/src/kernels/SelectV2.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+void SelectV2::configure()
+{
+  LUCI_INTERPRETER_CHECK(condition()->element_type() == DataType::BOOL);
+  LUCI_INTERPRETER_CHECK(t()->element_type() == e()->element_type());
+  LUCI_INTERPRETER_CHECK(t()->element_type() == output()->element_type());
+
+  auto cond_shape = condition()->shape();
+  auto cond_num_dims = cond_shape.num_dims();
```
jinevening, line 44: This variable is unused.

```suggestion
```

jinevening (approved):
LGTM

seanshpark (approved):
LGTM

PR #12182, opened 2023-12-04
[luci-compute] tflite_shape throw for unknown

seanshpark:
This will revise to throw for unknwon value with tflite_shape method.


seanshpark:
to be sure we have known shape for calling compute kernel codes.

jinevening (approved):
LGTM

PR #12183, opened 2023-12-04
[luci-interpreter] Remove broadcast feature from Select op

ragmani:
This commit removes broadcast feature from Select op.

ONE-DCO-1.0-Signed-off-by: ragmani ragmani0216@gmail.com

---

For #12148


seanshpark (approved):
LGTM

PR #12189, opened 2023-12-05
[luci-compute] Enable unit test with ConvertTypes

seanshpark:
This will enable unit test with ConvertTypes methods.


seanshpark:
for #12100 , from  #12099

mhs4670go (approved):
LGTM

seanshpark:
@jinevening ping

Issue #12177, opened 2023-12-04
[mio-circle07] Upgrade circle-schema


seanshpark:
Let's introduce mio-circle07 and revise dependent modules.
- new circle schema #12167 is now available

seanshpark:
major change(s)
- `data_format` field is deprecated

seanshpark:
close as done

PR #12195, opened 2023-12-05
[mio-circle07] Introduce mio-circle for v0.7

seanshpark:
This will introduce mio-circle for v0.7


seanshpark:
for #12177 , from  #12178
- code itself is copied from mio-circle06 with `data_format` removed.

jinevening (approved):
LGTM

mhs4670go (approved):
LGTM.

I come up with two options to reduce duplicate codes (of course we don't have to reduce them)

1. If we pull out duplicate APIs into a single module, we can just make `mio-circleX` depend on it.
2. Or, we can make the schema version a parameter for cmake build. Then, just single `mio-circle` module generates its header files according to the given version.

PR #12198, opened 2023-12-05
[infra] Include mio-circle07

seanshpark:
This will include mio-circle07 in infra file list.


seanshpark:
for #12177 , from  #12178

mhs4670go (approved):
LGTM

PR #12199, opened 2023-12-05
[luci-pass-value-py-test] Introduce pytest

mhs4670go:
This commit introduces a pytest for luci-pass-value-test.

Related: #12071

### DIFF
--- a/compiler/luci-pass-value-py-test/conftest.py
+++ b/compiler/luci-pass-value-py-test/conftest.py
@@ -1,0 +1,40 @@
+import re
+
+
+def extract_test_args(s):
+    p = re.compile('eval\\((.*)\\)')
+    result = p.search(s)
+    return result.group(1)
+
+
+def pytest_addoption(parser):
+    parser.addoption("--test_list", action="store", help="Path to test list")
+    parser.addoption(
+        "--tflite_dir", action="store", help="Directory including tflite file")
+    parser.addoption(
+        "--circle_dir", action="store", help="Directory including circle file")
+    parser.addoption(
+        "--luci_eval_driver", action="store", help="Path to luci eval driver")
+
+
+def pytest_generate_tests(metafunc):
+    list_path = metafunc.config.getoption('test_list')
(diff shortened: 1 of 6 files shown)


mhs4670go:
### Before

```bash
Start 75: luci_pass_value_test
75/76 Test #75: luci_pass_value_test ...................   Passed   63.73 sec
```

### After

```bash
$ ./nncc test -R luci_pass_value_py_test                    
Test project /home/seongwoo/ONE/build
    Start 70: luci_pass_value_py_test
1/1 Test #70: luci_pass_value_py_test ..........   Passed    1.98 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   1.98 sec
```

compiler/luci-pass-value-py-test/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+    tflite_dir = metafunc.config.getoption('tflite_dir')
+    circle_dir = metafunc.config.getoption('circle_dir')
+    eval_driver_path = metafunc.config.getoption('luci_eval_driver')
+    if list_path is None:
+        tests_default_tol = []
+    else:
+        with open(list_path) as f:
+            contents = [line.rstrip() for line in f]
+
+        comment_removed = [line for line in contents if not line.startswith('#')]
```
seanshpark, line 31: Q) is it OK to if there is `#` at the end like this?
```
addeval(model, pass) # just comment
```

compiler/luci-pass-value-py-test/requires.cmake
```diff
@@ -0,0 +1,2 @@
+require("common-artifacts")
+require("luci-eval-driver")
```
jinevening, line 2:
```suggestion
require("luci-eval-driver")
require("circle2circle")
```

compiler/luci-pass-value-py-test/conftest.py
mhs4670go: @seanshpark Yes. `re` module match just exact regex for this.
```python
p = re.compile('eval\((.*)\)')
```

PR #12201, opened 2023-12-05
[luci-value-test] Turn off the test

mhs4670go:
This commit excludes the module from build list.

This test is replaced by `luci-value-py-test` for reducing test duration.


seanshpark:
Change is OK
I'll confirm this after https://github.com/Samsung/ONE/issues/12071#issuecomment-1841648762 is done :)

seanshpark (approved):
LGTM

PR #12207, opened 2023-12-06
[infra/nncc] Note for utcount

seanshpark:
This will add a note for utcount as it is not maintained anymore.


hseok-oh (approved):
LGTM

PR #12209, opened 2023-12-06
[circle-operator] Revise to use mio-circle07

seanshpark:
This will revise to use mio-circle07.


seanshpark:
for  #12177 , from draft  #12178

PR #12206, opened 2023-12-06
DRAFT Revise nncc/Makefile.arm32

seanshpark:
on-going draft to fix compiler frontend arm32 build.

Signed-off-by: SaeHie Park <saehie.park@gmail.com>

PR #12221, opened 2023-12-06
[infra] Introduce test_venv in Makefile.arm32

seanshpark:
This will introduce test_venv for ARM32 target device in Makefile.arm32 to prepare python venv which needs python environment.


seanshpark:
for https://github.com/Samsung/ONE/issues/12071#issuecomment-1841648762 , from #12206

mhs4670go (approved):
LGTM

PR #12223, opened 2023-12-06
[onert] Implement nnfw_train_export_circle 

glistening:
- Introduce OriginIndex in Operand
- Introduce export_circle to classes from api to core
- Fix to copy the whole operand info during training phase
- To show, how to use
  - onert_train, circle_dump (just for buffer printing) are added temporarily

### DIFF
--- a/runtime/onert/core/include/backend/basic/train/TrainableBackendContextHelpers.h
+++ b/runtime/onert/core/include/backend/basic/train/TrainableBackendContextHelpers.h
@@ -43,9 +43,7 @@ ITensorRegistry *genTensors(backend::train::TrainableBackendContext &ctx,
       return;
     // NOTE Assuming there is no layout changes (Always assume NHWC or UNKNOWN)
     assert(tgraph.layout() != ir::Layout::NCHW);
-    ir::OperandInfo backend_info{obj.shape(), obj.typeInfo(), obj.info().memAllocType(),
-                                 obj.isConstant()};
-    tensor_builder->registerTensorInfo(ind, backend_info, ir::Layout::NHWC);
+    tensor_builder->registerTensorInfo(ind, obj.info(), ir::Layout::NHWC);
   });
 
   // For the executors that does not have fixed linear execution order:
--- a/runtime/onert/core/include/exec/Execution.h
+++ b/runtime/onert/core/include/exec/Execution.h
@@ -158,6 +158,14 @@ public:
    * @return @c float Loss value
    */
   float getLoss(const ir::IOIndex &ind);
+
+  /**
+   * @brief     Export circle
+   * @note      It should be called after training
+   * @param[in] path   Output path
+   * @throw     runtime_error on wrong path or permission denied
+   */
+  void export_circle(const std::string &path) const;
 #endif // ONERT_TRAIN
 
   ir::Shape getInputShape(ir::IOIndex ind) const;
--- a/runtime/onert/core/src/dumper/text/GraphDumper.cc
+++ b/runtime/onert/core/src/dumper/text/GraphDumper.cc
@@ -87,8 +87,10 @@ void dumpGraph(const ir::Graph &graph)
     const auto &op = graph.operations().at(op_ind);
     VERBOSE(GraphDumper) << "  " << formatOperation(op, op_ind) << "\n";
   }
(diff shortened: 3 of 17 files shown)


glistening:
I will introduce another module (`CircleMutator` class) whose role is to generate circle based on original circle. It fundamentally requires `circle_schema_generated.h`. However, it is located under `frontend`.

I don't want to make reverse dependency from `core` or `backend` to `frontend`.
Thus, I suggested to move `frontend/circle_schema` out of `frontend` so that any module which requires `circle_schema_generated.h` can use it.

But @hseok-oh told me he is going to relocate schema-related things.

Thus, I will make a different PR, which will include `CircleMutator` and put the implementation under `api` to workaround `circle-schema` reverse dependency temporarily.

runtime/onert/core/include/backend/basic/train/TrainableBackendContextHelpers.h
```diff
@@ -43,9 +43,7 @@ ITensorRegistry *genTensors(backend::train::TrainableBackendContext &ctx,
       return;
     // NOTE Assuming there is no layout changes (Always assume NHWC or UNKNOWN)
     assert(tgraph.layout() != ir::Layout::NCHW);
-    ir::OperandInfo backend_info{obj.shape(), obj.typeInfo(), obj.info().memAllocType(),
-                                 obj.isConstant()};
-    tensor_builder->registerTensorInfo(ind, backend_info, ir::Layout::NHWC);
+    tensor_builder->registerTensorInfo(ind, obj.info(), ir::Layout::NHWC);
```
glistening, line 46: I found new `originIndex` is missing after tensor builder phase. It is because `OperandInfo` is constructed with listed members only constructor. It would be better to use copy constructor.

runtime/onert/core/src/exec/train/TrainableExecutor.cc
```diff
@@ -207,6 +207,19 @@ float TrainableExecutor::getLoss(const ir::IOIndex &pred_io_ind) const
...
+  (void)path;
+  _tensor_regs.iterateTrainableTensors(
+    [&](const ir::OperandIndex &idx, backend::train::ITrainableTensor *tensor) {
+      printf("[G] Tensor %u: origin = %u, buffer (addr, size) = (%p, %zu)\n", idx.value(),
+             tensor->get_info().originIndex().value(), tensor->buffer(), tensor->total_size());
+      // I did not check buffer size or type. It is just quick check for temporary
+      float *fbuf = (float *)tensor->buffer();
+      printf("              buffer = %f, %f, %f, %f,...\n", fbuf[0], fbuf[1], fbuf[2], fbuf[3]);
+    });
+}
```
glistening, line 221: It is just to show how to use `iterateTrainableTensors` and make sure `buf` information is correct. It will be replaced with circle generation. But we cannot include `circle_schema_generated.h` from `core`.

runtime/onert/core/src/dumper/text/GraphDumper.cc
```diff
@@ -87,8 +87,10 @@ void dumpGraph(const ir::Graph &graph)
     const auto &op = graph.operations().at(op_ind);
     VERBOSE(GraphDumper) << "  " << formatOperation(op, op_ind) << "\n";
   }
+  graph.operands().iterate([&](const ir::OperandIndex &idx, const ir::Operand &oprd) {
+    VERBOSE(GraphDumper) << "  Origin(" << idx << "): " << oprd.originIndex() << std::endl;
+  });
```
glistening, line 92: It helps to track Operands information after each phases. It helps to find the problem like https://github.com/Samsung/ONE/pull/12223/files#r1416719944.

runtime/onert/core/include/exec/Execution.h
```diff
@@ -158,6 +158,14 @@ class Execution
...
    */
   float getLoss(const ir::IOIndex &ind);
+
+  /**
+   * @brief     Export circle
+   * @note      It should be called after training
+   * @param[in] path   Output path
+   * @throw     runtime_error on wrong path or permission denied
+   */
+  void export_circle(const std::string &path) const;
```
glistening, line 168: Note that it is enclosed with `#ifdef ONERT_TRAIN`. `#ifdef ONERT_TRAIN` should be removed.

glistening:
See #12245, except for the implementation under `api` will go out of `api`.

PR #12224, opened 2023-12-06
[Draft][onert] Remove frontend layout

ragmani:
This draft removes frontend layout feature.


runtime/onert/api/include/nnfw.h
```diff
@@ -389,7 +389,7 @@ NNFW_STATUS nnfw_output_size(nnfw_session *session, uint32_t *number);
 /**
  * @brief Set the layout of an input
  *
- * The input that does not call this has NNFW_LAYOUT_NHWC layout
+ * The input that does not call this has NNFW_LAYOUT_NONE layout
```
hseok-oh, line 392: What is model input layer if the input is used as input of `Conv` op?

ragmani: I think the layout should be `NNFW_LAYOUT_NHWC`. I didn't work for it yet in this draft.

PR #12227, opened 2023-12-06
[mio-circle06] Exclude from build

seanshpark:
This will make mio-circle06 to be excluded from genenral build.


seanshpark:
for  #12177 , from draft  #12178

mhs4670go (approved):
LGTM

PR #12230, opened 2023-12-06
[luci-compute] Make compiler happy with switch default

seanshpark:
This will revise to make compiler happy with default case for switch.


mhs4670go (approved):
LGTM

Issue #12127, opened 2023-11-28
[onert] Support Bool8 type of Mul op


ragmani:
## What

Let's support Bool8 type of Mul op.

## Why
Currently Bool8 type of Mul op is not supported yet.

## Tasks
- [x] Introduce Mul bool kernels
  - #12145

- [x] Apply Mul bool kernels to onert
  - #12231

Draft : #12144

ragmani:
Done.

PR #12231, opened 2023-12-06
[onert] Apply bool type kernel of Mul op

ragmani:
This commit applies bool type kernel of Mul op.
  - Apply the kernel to onert
  - Modify the kernel so that it can be called implicitly too

---

for #12127

### DIFF
--- a/compute/cker/include/cker/operation/BinaryArithmeticOps.h
+++ b/compute/cker/include/cker/operation/BinaryArithmeticOps.h
@@ -221,11 +221,11 @@ BinaryArithmeticOp(const BinaryArithmeticOpParam &params, const Shape &input1_sh
 template <BinaryArithmeticOpType op_type, typename T>
 inline typename std::enable_if_t<!is_quant8<T>::value && std::is_same<T, bool>::value>
 BinaryArithmeticOp(const BinaryArithmeticOpParam &params, const Shape &input1_shape,
-                   const bool *input1_data, const Shape &input2_shape, const bool *input2_data,
-                   const Shape &output_shape, bool *output_data)
+                   const T *input1_data, const Shape &input2_shape, const T *input2_data,
+                   const Shape &output_shape, T *output_data)
 {
   reference::BinaryArithmeticOp(params, input1_shape, input1_data, input2_shape, input2_data,
-                                output_shape, output_data, GetBinaryArtithmeticFn<op_type, bool>());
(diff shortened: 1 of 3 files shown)


jyoungyun (dismissed):
LGTM

compute/cker/include/cker/operation/BinaryArithmeticOps.h
```diff
@@ -221,11 +221,11 @@ BinaryArithmeticOp(const BinaryArithmeticOpParam &params, const Shape &input1_sh
...
 inline typename std::enable_if_t<!is_quant8<T>::value && std::is_same<T, bool>::value>
 BinaryArithmeticOp(const BinaryArithmeticOpParam &params, const Shape &input1_shape,
-                   const bool *input1_data, const Shape &input2_shape, const bool *input2_data,
-                   const Shape &output_shape, bool *output_data)
+                   const T *input1_data, const Shape &input2_shape, const T *input2_data,
+                   const Shape &output_shape, T *output_data)
 {
   reference::BinaryArithmeticOp(params, input1_shape, input1_data, input2_shape, input2_data,
-                                output_shape, output_data, GetBinaryArtithmeticFn<op_type, bool>());
+                                output_shape, output_data, GetBinaryArtithmeticFn<op_type, T>());
```
hseok-oh, line 228: I have a concern about using bool type array because tensorflow lite and our runtime uses 8bit for bool type, but c++'s sizeof(bool) is not restricts as 8bit. I think it has similar risky with `int`.

ragmani: Although c++'s `sizeof(bool)` is not restricted as 1 byte, most of `sizeof(bool)` seems to be 1 bytes. How about throw an exception when `sizeof(bool)` is not 1 byte? The place is where data is converted in onert?

hseok-oh: Why don't we use `static_assert`? 

ex)
https://github.com/Samsung/ONE/blob/8180a7ce28f147ea43fa143840eb23c3d0f5708a/compiler/luci/import/src/Importer.test.cpp#L119

https://github.com/Samsung/ONE/blob/8180a7ce28f147ea43fa143840eb23c3d0f5708a/compiler/enco/core/src/Support/Debugging.h#L27

jyoungyun (approved):
LGTM

hseok-oh (approved):
LGTM

PR #12233, opened 2023-12-07
[infra/onert] Remove mio-circle06 for luci build

seanshpark:
This will update luci build phase on runtime build to remove mio-circle06.


seanshpark:
@hseok-oh ping

PR #12234, opened 2023-12-07
[infra] Remove mio-circle06 from compiler list

seanshpark:
This will remove mio-circle06 from compiler build list.


mhs4670go (approved):
LGTM

PR #12235, opened 2023-12-07
[infra/nncc] Remove mio-circle06 from Makefile.arm32

seanshpark:
This will remove mio-circle06 from Makefile.arm32 build list.


mhs4670go (approved):
LGTM

PR #12238, opened 2023-12-07
[one-cmds] Fix import-onnx sanitize

seanshpark:
This will fix import-onnx sanitize method to also handle names starting with digit,
that was accidently removed in last change.


mhs4670go (approved):
LGTM

PR #12243, opened 2023-12-08
[circle-mpqsolver] Revise protected

stamalakhov:
This commit redeclares protected memebers as private for final classes.


jinevening (approved):
LGTM

PR #12247, opened 2023-12-11
[luci/pass] Reset shape status for SplitVOut

seanshpark:
This will reset shape status for CircleSplitVOut too for NCHW to NHWC conversion.


seanshpark:
@jinevening , plz check this if this is needed. I'll drop this if not.

jinevening:
This seems not necessary. #12254 will resolve the problem.

Issue #12248, opened 2023-12-11
CFE: ConvertNCHWToNHWCPass with CircleSplitV


seanshpark:
ConvertNCHWToNHWCPass with `CircleSplitV` is called twice for internal model.
So the axis is change like 1 -> 3 -> 2 which is incorrect.

seanshpark:
ping @jinevening

seanshpark:
@jinevening , I've check with unit test model (#12249) 
```
circle2circle --convert_nchw_to_nhwc SplitV_001.circle SplitV_001.opt.circle
```
this has not problem. 

may have to look for other related nodes ... -_-;

jinevening:
Our target model has multiple split_v ops sharing a single axis node. So, the value of the axis node was updated multiple times, which is wrong.

#12254 will fix the problem.

jinevening:
#12254 was merged. Let's close this issue.

PR #12249, opened 2023-12-11
DRAFT test SplitV with nchw to nhwc and shapeinf

seanshpark:
on-going draft to test SplitV with nchw to nhwc and shape inference.


seanshpark:
for #12248

seanshpark:
done testing

PR #12250, opened 2023-12-11
[circle-part-value-test] Turn off the test

mhs4670go:
This commit turns off the test as this test is replaced by circle-part-value-py-test.

Related: #12071 


seanshpark (approved):
LGTM

PR #12251, opened 2023-12-11
[infra] Add circle-part-value-py-test to compiler modules

mhs4670go:
This commit adds circle-part-value-py-test to compiler modules script and remove circle-part-value-test.


seanshpark (approved):
LGTM

Issue #12241, opened 2023-12-07
[luci/service] Concat shape infer fails if input shape is unknown


seanshpark:
When one of `CircleConcat` input shape is unknown for some reason, it drops error message which is hard to know.
```
ERROR: vector::_M_range_check: __n (which is 1) >= this->size() (which is 0)
```

Let's fix this to show some information about the input.

This unknown Op is `CustomOp` with no I/O shapes.

seanshpark:
done

PR #12257, opened 2023-12-11
[luci/service] Show better message for Concatenation

seanshpark:
This will show better error message for Concatenation if input shapes are incompatible.


seanshpark:
for  #12241

jinevening (approved):
LGTM

PR #12124, opened 2023-11-27
[DRAFT] Fuse Reshape-Mean-Reshape-Div pattern

BalyshevArtem:
This draft fuses reshape-mean-reshape-div pattern.

for issue: #11956

ONE-DCO-1.0-Signed-off-by: Artem Balyshev <a.balyshev@samsung.com>

Issue #11956, opened 2023-11-08
[one-optimize] Optimize Mul/Reshape/Div pattern


BalyshevArtem:
## What 

Let's optimize Reshape/Mean/Mul/Reshape/Div pattern, where Mul(x_mul, y_mul), y_mul - is const scalar value and Div(x_div, y_div), x_div - is const scalar value:

![original](https://github.com/Samsung/ONE/assets/43214667/79d12d8e-1b99-4ab2-ad29-534438e0b6f2)

as Mean/Div where Div(x_new, y_new), x_new const scalar value equal to (x_div / x_mul):

![Screenshot from 2023-11-10 13-44-09](https://github.com/Samsung/ONE/assets/43214667/c19f284c-4bdf-4ef5-9675-e6bf3633b7d6)

## Why 

Our target model has this pattern

jinevening:
@BalyshevArtem Could you give a background of this work?

The draft (#11995) would work only for a very specific pattern. If this task is urgent, it would be ok to proceed with it. However, we typically make a more generally-applicable pass. For example, it would be possible to forward Reshape Op across Mean/Mul/Div. Then, two Reshape Ops will be fused, so the final Ops will become Mean->Div->Reshape.

BalyshevArtem:
> @BalyshevArtem Could you give a background of this work?
> 
> The draft (#11995) would work only for a very specific pattern. If this task is urgent, it would be ok to proceed with it. However, we typically make a more generally-applicable pass. For example, it would be possible to forward Reshape Op across Mean/Mul/Div. Then, two Reshape Ops will be fused, so the final Ops will become Mean->Div->Reshape.

No, nothing urgent, just found such pattern in some target models. So, better co create 3 passes: 1)forward Reshape op across Mean/Mul/Div, 2)  Then fusing two reshape ops, 3) And fusing Mul/Div as Div, right?

jinevening:
We already have 2). 1) and 3) would be enough.

BalyshevArtem:
> We already have 2). 1) and 3) would be enough.

Introduce new #12124 draft. To optimize this pass I added CircleDiv, CircleMul, CircleMean to forwardReshape op pass (https://github.com/Samsung/ONE/pull/12124#discussion_r1406314803, https://github.com/Samsung/ONE/pull/12124#discussion_r1406315209, https://github.com/Samsung/ONE/pull/12124#discussion_r1406318707).
[...]

PR #12254, opened 2023-12-11
[luci] Fix SplitV bug in ConvertNCHWToNHWC

jinevening:
This creates a new axis rather than in-place update.

ONE-DCO-1.0-Signed-off-by: Hyukjin Jeong <hj1.jeong@samsung.com>

---
Related to: https://github.com/Samsung/ONE/issues/12248

PR #12261, opened 2023-12-11
[luci/pass] Extend ForwardReshapeToUnaryOpPass with Mean op

BalyshevArtem:
This commit extends ForwardReshapeToUnaryOpPass with Mean op.

for issue: https://github.com/Samsung/ONE/issues/11956
from draft https://github.com/Samsung/ONE/pull/12124

### DIFF
--- a/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
+++ b/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
@@ -96,6 +96,7 @@ bool forward_reshape(luci::CircleReshape *reshape, luci::CircleMean *mean, uint3
   auto *shape_reshape = loco::must_cast<luci::CircleConst *>(new_reshape->shape());
   assert(shape_reshape->dtype() == loco::DataType::S32);     // FIX_CALLER_UNLESS
   assert(axis < shape_reshape->size<loco::DataType::S32>()); // FIX_CALLER_UNLESS
+  // Mean reduction will make value to '1'
   shape_reshape->at<loco::DataType::S32>(axis) = 1;
 
   // Do shape inference for this node again.
--- a/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
+++ b/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
@@ -285,7 +285,7 @@ protected:
 
     for (int32_t i = 0; i <= axis_value; ++i)
     {
-      if (reshape_input->dim(i).value() != reshape->dim(i).value())
+      if (not reshape_input->dim(i).known() or reshape_input->dim(i).value() != reshape->dim(i).value())
         return false;
     }
 
--- a/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
+++ b/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
@@ -243,6 +243,34 @@ protected:
     return false;
   }
 
+  /**
+   * Graph example:
+   *
+   *  BEFORE
+   *               [Input]
+   *              (3, 4, 4)                 [Shape_Const = (1, -1, 4)]
+   *                  |                     |
+   *              [Reshape] ----------------
+   *              (1, 12, 4)
+   *                  |
+   *        [Mean, keep_dims = true]
+   *              (1, 12, 1)
+   *                  |
+   *               [Output]
+   *
+   *  AFTER
+   *               [Input]
+   *              (3, 4, 4)
+   *                  |
+   *         [Mean, keep_dims = true]
+   *              (3, 4, 1)                 [Shape_Const = (1, -1, 1)]
+   *                  |                     |
+   *              [Reshape]-----------------
+   *              (1, 12, 1)
+   *                  |
+   *              [Output]
+   *
+   */
   bool visit(luci::CircleMean *node)
   {
     luci::CircleReshape *reshape = nullptr;
--- a/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
+++ b/compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
@@ -78,6 +78,33 @@ luci::CircleReshape *create_cloned_reshape(luci::CircleReshape *reshape)
   return new_reshape;
 }
 
+bool forward_reshape(luci::CircleReshape *reshape, luci::CircleMean *mean, uint32_t axis)
+{
+  assert(reshape != nullptr); // FIX_CALLER_UNLESS
+  assert(mean != nullptr);    // FIX_CALLER_UNLESS
+
+  auto new_reshape = create_cloned_reshape(reshape);
+  if (not new_reshape)
+    return false;
+
+  // reconnect network
+  loco::replace(mean).with(new_reshape);
+  mean->input(reshape->tensor());
+  new_reshape->tensor(mean);
+
+  // Change const shape axis value
+  auto *shape_reshape = loco::must_cast<luci::CircleConst *>(new_reshape->shape());
+  assert(shape_reshape->dtype() == loco::DataType::S32);     // FIX_CALLER_UNLESS
+  assert(axis < shape_reshape->size<loco::DataType::S32>()); // FIX_CALLER_UNLESS
+  shape_reshape->at<loco::DataType::S32>(axis) = 1;
+
+  // Do shape inference for this node again.
+  mean->shape_status(luci::ShapeStatus::UNDEFINED);
+  reshape->shape_status(luci::ShapeStatus::UNDEFINED);
+
+  return true;
+}
+
 bool forward_reshape(luci::CircleReshape *reshape, luci::CircleAbs *abs)
 {
   assert(reshape != nullptr); // FIX_CALLER_UNLESS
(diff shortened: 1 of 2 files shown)


compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
```diff
@@ -201,6 +277,9 @@ namespace luci
  *          |            |           |
  *
  *   UnaryOp: CircleNeg, ...
+ *   Note: For CircleMean in which the axis is a scalar
+ *   constant and reshape Op does not change the axis on which the mean is
+ *   taken, the Reshape Op can be forwarded.
```
seanshpark, line 282:
```suggestion
 *   Note: For CircleMean in which the axis is a scalar
 *         constant and reshape Op does not change the axis on which the mean is
 *         taken, the Reshape Op can be forwarded.
```

BalyshevArtem: Done

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
```diff
@@ -201,6 +277,9 @@ namespace luci
  *          |            |           |
  *
  *   UnaryOp: CircleNeg, ...
+ *   Note: For CircleMean in which the axis is a scalar
+ *      constant and reshape Op does not change the axis on which the mean is
+ *      taken, the Reshape Op can be forwarded.
```
seanshpark, line 282:
```suggestion
 *   Note: For CircleMean in which the axis is a scalar
 *         constant and reshape Op does not change the axis on which the mean is
 *         taken, the Reshape Op can be forwarded.
```

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
```diff
@@ -156,6 +183,55 @@ class ForwardReshape final : public luci::CircleNodeMutableVisitor<bool>
...
+
+    // reshape shouldn't change rank
+    if (reshape_input->rank() != reshape->rank())
+      return false;
+
+    assert(reshape_input->rank() > static_cast<uint32_t>(axis_value));
+
+    for (int32_t i = 0; i <= axis_value; ++i)
+    {
+      if (reshape_input->dim(i).value() != reshape->dim(i).value())
```
seanshpark, line 228: maybe have to check `reshape_input->dim(i).known()` as input may have unknown shape.

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
```diff
@@ -76,6 +76,33 @@ luci::CircleReshape *create_cloned_reshape(luci::CircleReshape *reshape)
...
+  // reconnect network
+  loco::replace(mean).with(new_reshape);
+  mean->input(reshape->tensor());
+  new_reshape->tensor(mean);
+
+  // Change const shape axis value
+  auto *shape_reshape = loco::must_cast<luci::CircleConst *>(new_reshape->shape());
+  assert(shape_reshape->dtype() == loco::DataType::S32);     // FIX_CALLER_UNLESS
+  assert(axis < shape_reshape->size<loco::DataType::S32>()); // FIX_CALLER_UNLESS
+  shape_reshape->at<loco::DataType::S32>(axis) = 1;
```
seanshpark, line 97: Q) why put `1` ?

BalyshevArtem: Since we are swapping the reshape and mean operation, then after applying mean first, the axis tensor dimension values will be equal to 1, and accordingly it is necessary to change the reshape_const value so that it corresponds to the new input for reshape.

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
BalyshevArtem: Done

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
```diff
@@ -156,6 +183,55 @@ class ForwardReshape final : public luci::CircleNodeMutableVisitor<bool>
...
+    if (axis->dtype() != loco::DataType::S32)
+      return false;
+
+    // Should be scalar
+    if (axis->size<loco::DataType::S32>() != 1)
+      return false;
+
+    // axis value
+    int32_t axis_value;
+    axis_value = axis->at<loco::DataType::S32>(0);
```
jinevening, line 206: For simplicity
```suggestion
    auto axis_value = axis->at<loco::DataType::S32>(0);
```

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
```diff
@@ -156,6 +183,55 @@ class ForwardReshape final : public luci::CircleNodeMutableVisitor<bool>
...
+      axis_value += static_cast<int32_t>(reshape->rank());
+
+    assert(axis_value >= 0);
+
+    if (node->keep_dims() != true)
+      return false;
+
+    auto *reshape_input = dynamic_cast<luci::CircleNode *>(reshape->tensor());
+    if (reshape_input == nullptr)
+      return false;
```
jinevening, line 218:
```suggestion
    auto reshape_input = loco::must_cast<luci::CircleNode *>(reshape->tensor());
```

BalyshevArtem: Done, thank you

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
seanshpark: Thanks for the long explanation, but I don't understand (sorry for my ignorance) why it is equal to 1. Can you give more information about this?

jinevening: @BalyshevArtem Forwarding reshape over mean seems not trivial. Could you add a test as you did in `FuseMulWithDivPass` ?  It would be good to draw BEFORE and AFTER graph for readers.

BalyshevArtem: > Could you add a test as you did in `FuseMulWithDivPass` ? It would be good to draw BEFORE and AFTER graph for readers.

Added test.


> Can you give more information about this?

after the mean operation on the axis dimension, the value will always be equal to one. Therefore, shape const should also change its axis value by one:
```
/**
 *  Simple graph for test
 *
 *  BEFORE
 *               [Input]
 *              (3, 4, 4)                 [Shape_Const = (1, -1, 4)]
 *                  |                               |
 *              [Reshape] ----------------
 *              (1, 12, 4)
 *                  |
 *        [Mean, keep_dims = true]
 *              (1, 12, 1)
 *                  |
 *               [Output]
 *
 *  AFTER
 *               [Input]
 *              (3, 4, 4)
 *                  |
 *         [Mean, keep_dims = true]
 *              (3, 4, 1)                 [Shape_Const = (1, -1, 1)]
 *                  |                               |
 *              [Reshape]-----------------
 *              (1, 12, 1)
 *                  |
 *              [Output]
 *
 */
```

seanshpark: Thanks for the explanation! Now I understand :)

seanshpark: It would be nice to put BEFORE/AFTER diagram in the source

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
```diff
@@ -78,6 +78,33 @@ luci::CircleReshape *create_cloned_reshape(luci::CircleReshape *reshape)
...
+  // reconnect network
+  loco::replace(mean).with(new_reshape);
+  mean->input(reshape->tensor());
+  new_reshape->tensor(mean);
+
+  // Change const shape axis value
+  auto *shape_reshape = loco::must_cast<luci::CircleConst *>(new_reshape->shape());
+  assert(shape_reshape->dtype() == loco::DataType::S32);     // FIX_CALLER_UNLESS
+  assert(axis < shape_reshape->size<loco::DataType::S32>()); // FIX_CALLER_UNLESS
+  shape_reshape->at<loco::DataType::S32>(axis) = 1;
```
seanshpark, line 99: Like #12254 , if `shape_reshape` is shared, there would an error.

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
seanshpark: Found graph in test code. let's move that to here.

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
BalyshevArtem: above we use cloned `new_reshape` node : `auto new_reshape = create_cloned_reshape(reshape);`, where `shape_reshape` also cloned as new node: `luci::CircleConst *cloned_shape = clone_shape(reshape);`.

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
BalyshevArtem: > Found graph in test code. let's move that to here.

Done

compiler/luci/pass/src/ForwardReshapeToUnaryOpPass.cpp
seanshpark, line 99: just one more thing for future beginners,
```suggestion
  // Mean reduction will make value to '1'
  shape_reshape->at<loco::DataType::S32>(axis) = 1;
```

BalyshevArtem: done

seanshpark (approved):
LGTM thank you!

jinevening (approved):
LGTM

PR #12246, opened 2023-12-11
[onert] Implement nnfw_train_export_circle (in api)

glistening:
- Introduce OriginIndex in Operand
- Fix to copy the whole operand info during training phase
- Introduce iterateTrainableTensors from api to core
- Introduce _model_path in nnfw_session

ONE-DCO-1.0-Signed-off-by: Sanggyu Lee [sg5.lee@samsung.com](mailto:sg5.lee@samsung.com)

Related: #12223

PR #12273, opened 2023-12-12
[tools/onert_train] Introduce --export_path option

glistening:
It introduces --export_path to specify the path to export circle.

Draft: #12246

### DIFF
--- a/tests/tools/onert_train/src/args.cc
+++ b/tests/tools/onert_train/src/args.cc
@@ -150,6 +150,8 @@ void Args::Initialize(void)
     }
   };
 
+  auto process_export_path = [&](const std::string &path) { _export_model_filename = path; };
+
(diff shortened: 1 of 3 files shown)


tests/tools/onert_train/src/args.cc
```diff
@@ -150,6 +150,8 @@ void Args::Initialize(void)
     }
   };
 
+  auto process_export_path = [&](const std::string &path) { _export_model_filename = path; };
```
jyoungyun, line 153: What happens if the file is already created?

jyoungyun (approved):
LGTM

zetwhite (approved):
LGTM

tests/tools/onert_train/src/args.cc
glistening: At this moment, it just overwrites the existing name of file. We may decide the policy.

PR #12278, opened 2023-12-13
[one-cmds] Add constant.py dependency

jinevening:
This adds constant.py to cmake dependency for automatic recompilation.


compiler/one-cmds/CMakeLists.txt
```diff
@@ -168,7 +169,7 @@ add_custom_command(OUTPUT ${O1_CFG_FILE_BIN}
           --constant ${O1_OPTION}
           --format cfg
           --output_path ${O1_CFG_FILE_BIN}
-  DEPENDS ${CONSTANT_EXPORTING_SCRIPT}
+  DEPENDS ${CONSTANT_EXPORTING_SCRIPT} ${CONSTANT_SCRIPT}
```
jinevening, line 172: `export_constant.py` internally uses `constant.py`, but it was not considered. So, even though O1 list in `constant.py` is updated, the actual O1 list in the `O1.cfg` is not updated if there is a previously-built file.

This PR fixes the issue, so now O1 list is updated when `constant.py` is updated.

seanshpark (approved):
LGTM

zetwhite (approved):
LGTM

mhs4670go (approved):
LGTM

PR #12292, opened 2023-12-14
[onert] mmap circle file for exporting circle

glistening:
It ensure to mmap circle file with getting file size.

Draft: https://github.com/Samsung/ONE/pull/12246

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -37,6 +37,10 @@
 #include <dirent.h>
 #include <misc/string_helpers.h>
 
+#include <fcntl.h>    // O_RDONLY
+#include <sys/mman.h> // mmap, munmap
+#include <sys/stat.h> // fstat
+#include <unistd.h>   // close
 /*
  * API does not accept string argument longer than max length below
  */
@@ -1450,8 +1454,47 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
     return NNFW_STATUS_INVALID_STATE;
   }
 
-  // NYI
-  return NNFW_STATUS_ERROR;
+  auto getFdSize = [](int fd) -> off_t {
+    if (fd < 0)
+      return -1L;
+
+    struct stat file_stat;
+    if (fstat(fd, &file_stat) != 0)
+      return -1L;
+
+    return file_stat.st_size;
+  };
+
+  auto ensure_mmap = [getFdSize](const char *file_path,
+                                 /* out */ size_t &mmapped_buf_sz) -> uint8_t * {
+    int fd = open(file_path, O_RDWR);
+    auto sz = getFdSize(fd);
+    if (sz < 0)
+      return nullptr;
+
+    auto buf = static_cast<uint8_t *>(mmap(NULL, sz, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0));
+    if (buf == MAP_FAILED)
+    {
+      close(fd);
+      return nullptr;
+    }
+    // msync, munmap requires `size_t` while fstat returns `off_t`.
+    // `off_t` may be long or long long.
+    // `size_t` may be uint32 or uint64.
+    // `sz` is not negative by if-statement above.
+    assert(sizeof(off_t) == sizeof(size_t));
+    mmapped_buf_sz = static_cast<size_t>(sz);
+    return buf;
+  };
+
+  size_t mmapped_buf_sz;
+  uint8_t *mmapped_buf = ensure_mmap(path, mmapped_buf_sz);
+  if (mmapped_buf == nullptr)
+    return NNFW_STATUS_INVALID_FILE;
+
+  if (munmap(mmapped_buf, mmapped_buf_sz) == -1)
+    return NNFW_STATUS_ERROR;
+  return NNFW_STATUS_NO_ERROR;
 }
 
 bool nnfw_session::isStatePreparedTraining()
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -1467,7 +1467,8 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
       // file_stat.st_size is `off_t` while msync, munmap requires `size_t`
       // `off_t` may be long or long long.
       // `size_t` may be uint32 or uint64.
-      static_assert(sizeof(off_t) == sizeof(size_t), "sizeof(off_t) != sizeof(size_t)");
+      static_assert(sizeof(off_t) <= sizeof(size_t),
+                    "size_t should be bigger than off_t in positive range");
       _buf_sz = static_cast<size_t>(file_stat.st_size);
       _buf = mmap(NULL, _buf_sz, PROT_READ | PROT_WRITE, MAP_SHARED, _fd, 0);
       return _buf != MAP_FAILED;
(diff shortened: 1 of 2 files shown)


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1450,8 +1454,47 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
...
+
+    return file_stat.st_size;
+  };
+
+  auto ensure_mmap = [getFdSize](const char *file_path,
+                                 /* out */ size_t &mmapped_buf_sz) -> uint8_t * {
+    int fd = open(file_path, O_RDWR);
+    auto sz = getFdSize(fd);
+    if (sz < 0)
+      return nullptr;
```
jyoungyun, line 1473:
```suggestion
    if (sz < 0) {
      close(fd);
      return nullptr;
    }
```

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1450,8 +1454,47 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
...
+      close(fd);
+      return nullptr;
+    }
+    // msync, munmap requires `size_t` while fstat returns `off_t`.
+    // `off_t` may be long or long long.
+    // `size_t` may be uint32 or uint64.
+    // `sz` is not negative by if-statement above.
+    assert(sizeof(off_t) == sizeof(size_t));
+    mmapped_buf_sz = static_cast<size_t>(sz);
+    return buf;
```
jyoungyun, line 1487: When does the `fd` close?

runtime/onert/api/src/nnfw_api_internal.cc
YongseopKim: 👍

YongseopKim: and below code could be used

```c++
int fd = open(file_path, O_RDWR);
if (fd < 0) {
  // handle error
}
```

glistening:
@jyoungyun, @YongseopKim I've updated PR to make sure `fd` is closed.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1450,8 +1454,64 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
...
+
+    bool ensure_mmap()
+    {
+      struct stat file_stat;
+      if (fstat(_fd, &file_stat) != 0)
+        return false;
+      // file_stat.st_size is `off_t` while msync, munmap requires `size_t`
+      // `off_t` may be long or long long.
+      // `size_t` may be uint32 or uint64.
+      assert(sizeof(off_t) == sizeof(size_t));
```
ragmani, line 1470:
```suggestion
      static_assert(sizeof(off_t) == sizeof(size_t));
```

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1450,8 +1454,64 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
...
+      {
+        ::close(_fd);
+        _fd = -1; // mark as cleaned up
+      }
+      return ret;
+    }
+
+    ~MMappedFile() { close(); }
+
+    uint8_t *buf() const { return static_cast<uint8_t *>(_buf); }
```
ragmani, line 1496:
```suggestion
    uint8_t *buf() const { return reinterpret_cast<uint8_t *>(_buf); }
```

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1450,8 +1454,64 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
...
+      struct stat file_stat;
+      if (fstat(_fd, &file_stat) != 0)
+        return false;
+      // file_stat.st_size is `off_t` while msync, munmap requires `size_t`
+      // `off_t` may be long or long long.
+      // `size_t` may be uint32 or uint64.
+      assert(sizeof(off_t) == sizeof(size_t));
+      _buf_sz = static_cast<size_t>(file_stat.st_size);
+      _buf =
+        static_cast<uint8_t *>(mmap(NULL, _buf_sz, PROT_READ | PROT_WRITE, MAP_SHARED, _fd, 0));
```
ragmani, line 1473:
```suggestion
        reinterpret_cast<uint8_t *>(mmap(NULL, _buf_sz, PROT_READ | PROT_WRITE, MAP_SHARED, _fd, 0));
```

ragmani (dismissed):
LGTM +0.5 (I don't know about mmap well)

runtime/onert/api/src/nnfw_api_internal.cc
glistening: @ragmani [Reference](https://en.cppreference.com/w/cpp/language/static_assert) says `static_assert( bool-constexpr )` is available since c++17. It may require additional change. Let's see how our CI says.

ragmani: If so, how about using `static_assert( bool-constexpr , message )` ?

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1450,8 +1454,62 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
...
+
+    bool ensure_mmap()
+    {
+      struct stat file_stat;
+      if (fstat(_fd, &file_stat) != 0)
+        return false;
+      // file_stat.st_size is `off_t` while msync, munmap requires `size_t`
+      // `off_t` may be long or long long.
+      // `size_t` may be uint32 or uint64.
+      static_assert(sizeof(off_t) == sizeof(size_t), "sizeof(off_t) != sizeof(size_t)");
```
ragmani, line 1470: As far as I know, we can know `bool-constexpr` without message of `static_assert.

ragmani (commented):
LGTM +0.5 (I don't know about mmap well)

ragmani (approved):

ragmani (dismissed):
LGTM

runtime/onert/api/src/nnfw_api_internal.cc
glistening: Yes, I made the message more meaningful. Also, I've updated `cond` because it is okay that `size_t` have bigger size than `off_t`.

glistening:
@ragmani CI shows `static_assert` failure during Tizen build. Tizen seems to use 32 bit `size_t`, 64 bit `off_t`. Thus, I change `static_assert` to runtime range check. Thank you for helping me find the error earlier. :+1:

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1462,13 +1462,10 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
     bool ensure_mmap()
     {
       struct stat file_stat;
-      if (fstat(_fd, &file_stat) != 0)
+      if (fstat(_fd, &file_stat) != 0 || file_stat.st_size < 0 ||
+          static_cast<size_t>(file_stat.st_size) > SIZE_MAX)
```
ragmani, line 1466: Is always `static_cast<size_t>(file_stat.st_size) > SIZE_MAX` true?

glistening: :cry: You're right. `size_t` should be `uint64_t`.

glistening:
@Samsung/one_onert PTAL

chunseoklee (approved):

YongseopKim (approved):

PR #12152, opened 2023-11-30
[draft] Circle+ Importer v3 implementation

zetwhite:
List of updated 

```
runtime/onert
  ㄴ api 
		   * support new API listed below 
	           * nnfw_train_get_traininfo
	           * nnfw_train_get_batchsize
	        * nnfw_load_model_from_modelfile : parse(load) metadata into session._train_info 
  ㄴ core
			* ir/Model : hold metadata name and related buffer(ir/Data)
  ㄴ frontend
			* base_loader : load metadata(copy buffer) while loading a model
			* traininfo_loader : .... 
  ㄴ tests/tools/onert_train
			* update cmdline argument to have default(invalid) value 
			* update onert_train to use new api 
```

PR #12294, opened 2023-12-14
[onert] Introduce metadata to model and loader

zetwhite:
This PR introduces a metadata field to ir::Model and related loading logic.

draft : https://github.com/Samsung/ONE/pull/12152 
issue : https://github.com/Samsung/ONE/issues/11692

### DIFF
--- a/runtime/onert/core/include/ir/Model.h
+++ b/runtime/onert/core/include/ir/Model.h
@@ -177,9 +177,9 @@ private:
   std::shared_ptr<backend::custom::IKernelBuilder> _kernel_builder;
 
 public:
-  void add_metadata(const std::string &name, std::shared_ptr<const ir::Data> data)
+  void add_metadata(const std::string &name, std::unique_ptr<const ir::Data> data)
   {
-    _metadatas.emplace(name, data);
+    _metadatas.emplace(name, std::move(data));
   }
 
   bool is_metadata_exist(const std::string &name) const
@@ -187,13 +187,20 @@ public:
     return _metadatas.find(name) != _metadatas.end();
   }
 
-  std::shared_ptr<const ir::Data> get_metadata(const std::string name) const
+  std::unique_ptr<const ir::Data> get_metadata(const std::string name)
   {
-    return _metadatas.at(name);
+    auto m = _metadatas.find(name);
+
+    if (m == _metadatas.end())
+      throw std::out_of_range{"no meatdata named " + name};
+
+    auto data = std::move(m->second);
+    _metadatas.erase(m);
+    return data;
   }
 
 private:
-  std::unordered_map<std::string, std::shared_ptr<const ir::Data>> _metadatas;
+  std::unordered_map<std::string, std::unique_ptr<const ir::Data>> _metadatas;
 };
 } // namespace ir
 } // namespace onert
--- a/runtime/onert/core/include/ir/Model.h
+++ b/runtime/onert/core/include/ir/Model.h
@@ -187,7 +187,8 @@ public:
     return _metadatas.find(name) != _metadatas.end();
   }
 
-  std::unique_ptr<const ir::Data> get_metadata(const std::string name)
+  // NOTE The corresponding metadata is deleted from the model and returned
+  std::unique_ptr<const ir::Data> extract_metadata(const std::string name)
   {
     auto m = _metadatas.find(name);
 
--- a/runtime/onert/core/include/ir/Model.h
+++ b/runtime/onert/core/include/ir/Model.h
@@ -175,8 +175,26 @@ public:
 
 private:
   std::shared_ptr<backend::custom::IKernelBuilder> _kernel_builder;
-};
 
+public:
+  void add_metadata(const std::string &name, std::shared_ptr<const ir::Data> data)
+  {
+    _metadatas.emplace(name, data);
+  }
+
+  bool is_metadata_exist(const std::string &name) const
+  {
+    return _metadatas.find(name) != _metadatas.end();
+  }
+
+  std::shared_ptr<const ir::Data> get_metadata(const std::string name) const
+  {
+    return _metadatas.at(name);
+  }
+
+private:
+  std::unordered_map<std::string, std::shared_ptr<const ir::Data>> _metadatas;
+};
 } // namespace ir
 } // namespace onert
 
--- a/runtime/onert/core/include/ir/Model.h
+++ b/runtime/onert/core/include/ir/Model.h
@@ -182,7 +182,7 @@ public:
     _metadatas.emplace(name, std::move(data));
   }
 
-  bool is_metadata_exist(const std::string &name) const
+  bool exists_metadata(const std::string &name) const
   {
     return _metadatas.find(name) != _metadatas.end();
   }
--- a/runtime/onert/frontend/base_loader/include/base_loader.h
+++ b/runtime/onert/frontend/base_loader/include/base_loader.h
@@ -249,11 +249,7 @@ template <typename LoaderDomain>
 std::unique_ptr<ir::Data>
 BaseLoader<LoaderDomain>::BaseLoader::loadMetadata(const uint32_t buffer_idx)
 {
-  if (_domain_model == nullptr)
-  {
-    throw std::runtime_error{"fail to access _domain_model"};
-  }
-
+  assert(_domain_model != nullptr);
   const auto *data = _domain_model->buffers()->Get(buffer_idx)->data();
   if (_fd == -1) // Model is from memory
   {
--- a/runtime/onert/frontend/base_loader/include/base_loader.h
+++ b/runtime/onert/frontend/base_loader/include/base_loader.h
@@ -47,6 +47,7 @@ protected:
   using Buffer = typename LoaderDomain::Buffer;
   using BuiltinOperator = typename LoaderDomain::BuiltinOperator;
   using CustomOptionsFormat = typename LoaderDomain::CustomOptionsFormat;
+  using Metadata = typename LoaderDomain::Metadata;
   using Model = typename LoaderDomain::Model;
   using Operator = typename LoaderDomain::Operator;
   using Padding = typename LoaderDomain::Padding;
@@ -122,6 +123,7 @@ protected:
   }
 
 private:
+  std::unique_ptr<ir::Data> loadMetadata(const uint32_t buffer_idx);
   virtual std::unique_ptr<ir::Graph> loadSubgraph(const SubGraph *subg) = 0;
   // Operations
   template <typename OpIR, typename... Args>
@@ -241,6 +243,35 @@ void BaseLoader<LoaderDomain>::BaseLoader::loadFromBuffer(uint8_t *buffer, size_
   _base = buffer;
   _verifier = std::make_unique<Verifier>(reinterpret_cast<const std::uint8_t *>(_base), size);
   loadModel();
+}
+
+template <typename LoaderDomain>
+std::unique_ptr<ir::Data>
+BaseLoader<LoaderDomain>::BaseLoader::loadMetadata(const uint32_t buffer_idx)
+{
+  if (_domain_model == nullptr)
+  {
+    throw std::runtime_error{"fail to access _domain_model"};
+  }
+
+  const auto *data = _domain_model->buffers()->Get(buffer_idx)->data();
+  if (_fd == -1) // Model is from memory
+  {
+    return std::make_unique<ir::ExternalData>(data->data(), data->size());
+  }
+  else // Model is loaded(mmap'd) from a file
+  {
+    size_t data_size = data->size();
+    ptrdiff_t offset_start = data->data() - _base;
+    ptrdiff_t offset_end = offset_start + data_size;
+
+    ptrdiff_t page_start = (offset_start / _pagesize) * _pagesize;
+    size_t mapping_size = offset_end - page_start;
+
+    // Since metadata is not access often in inference/training time, always use mmaped-data
+    // Ref : https://github.com/Samsung/ONE/issues/3961#issuecomment-681750231
+    return std::make_unique<ir::MMapedData>(_fd, page_start, mapping_size, offset_start, data_size);
+  }
 }
 
 template <typename LoaderDomain>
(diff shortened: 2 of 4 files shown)


runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -47,6 +47,7 @@ template <typename LoaderDomain> class BaseLoader
   using Buffer = typename LoaderDomain::Buffer;
   using BuiltinOperator = typename LoaderDomain::BuiltinOperator;
   using CustomOptionsFormat = typename LoaderDomain::CustomOptionsFormat;
+  using Metadata = typename LoaderDomain::Metadata;
```
zetwhite, line 50: Since both `tflite` and `circle` have metadata, Let add `Metadata` and `loadMetadata` in base_loader.

tflite_schema.fbs
https://github.com/Samsung/ONE/blob/11620233e1a303dcc60f48fc7103cb55d84270cc/runtime/onert/frontend/tflite/tflite_schema.fbs#L1239-L1244 

circle_schema.fbs (v0.7) 
https://github.com/Samsung/ONE/blob/11620233e1a303dcc60f48fc7103cb55d84270cc/res/CircleSchema/0.7/circle_schema.fbs#L1616-L1621

runtime/onert/core/include/ir/Model.h
```diff
@@ -175,8 +175,26 @@ class Model
...
+    return _metadatas.find(name) != _metadatas.end();
+  }
+
+  std::shared_ptr<const ir::Data> get_metadata(const std::string name) const
+  {
+    return _metadatas.at(name);
+  }
+
+private:
+  std::unordered_map<std::string, std::shared_ptr<const ir::Data>> _metadatas;
```
zetwhite, line 196: Let's make a Model to hold metadata-name and related buffer(Data). 

I have an plan to parse this data like this : (draft code - https://github.com/Samsung/ONE/pull/12152/files#r1426184845 )

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -243,6 +245,35 @@ void BaseLoader<LoaderDomain>::BaseLoader::loadFromBuffer(uint8_t *buffer, size_
...
 
+template <typename LoaderDomain>
+std::unique_ptr<ir::Data>
+BaseLoader<LoaderDomain>::BaseLoader::loadMetadata(const uint32_t buffer_idx)
+{
+  if (_domain_model == nullptr)
+  {
+    throw std::runtime_error{"fail to access _domain_model"};
+  }
+
```
zetwhite, line 256:
```suggestion
  assert(_domain_model != nullptr); 
```

runtime/onert/core/include/ir/Model.h
```diff
@@ -175,8 +175,26 @@ class Model
...
+
+  bool is_metadata_exist(const std::string &name) const
+  {
+    return _metadatas.find(name) != _metadatas.end();
+  }
+
+  std::shared_ptr<const ir::Data> get_metadata(const std::string name) const
+  {
+    return _metadatas.at(name);
+  }
```
jyoungyun, line 193: How about moving these public functions to one place on top public specifier?

runtime/onert/core/include/ir/Model.h
```diff
@@ -175,8 +175,26 @@ class Model
 
 private:
   std::shared_ptr<backend::custom::IKernelBuilder> _kernel_builder;
-};
 
+public:
+  void add_metadata(const std::string &name, std::shared_ptr<const ir::Data> data)
```
jyoungyun, line 180: How about following the same naming convention like other functions? Other functions use CamelCase naming convention.

zetwhite: Thanks for pointing out.

I tried to, but it's look like that some method uses snake_case, and some uses CamelCase
Is there a more recommended rule? 

https://github.com/zetwhite/ONE/blob/eba3f1d21a0dfdcc600f8d9d8982fae51939bc6c/runtime/onert/core/include/ir/Model.h#L128-L162

runtime/onert/core/include/ir/Model.h
zetwhite: As I understood, This file(Model.h) follows this rule. 

```
public: 
   ... some interface to access the field. 
private : 
   ... filed
``` 

* `_kernel_builder` and its related interface
https://github.com/Samsung/ONE/blob/44b7707080e52b0d90a70800edf1a942371988ed/runtime/onert/core/include/ir/Model.h#L164-L177 

* `_subgraph` and its related interface
https://github.com/Samsung/ONE/blob/44b7707080e52b0d90a70800edf1a942371988ed/runtime/onert/core/include/ir/Model.h#L139-L161

zetwhite: > As I understood, This file(Model.h) follows this rule.

So I tried the follow the convention as I understood.
Do you still think it is better to move the public method to the top?  
If you think so, I'd like to follow :)

runtime/onert/core/include/ir/Model.h
zetwhite, line 196: (note) better to use `std::unique_ptr<const ir::Data>`

runtime/onert/core/include/ir/Model.h
jyoungyun: I understood that these functions were implemented in the same way as other functions in this file. Keep it as it is. :)

runtime/onert/core/include/ir/Model.h
jyoungyun: Well, when I looked onert code again, there is no naming convention rule. So you don't need to update it.

hseok-oh (dismissed):
LGTM

runtime/onert/core/include/ir/Model.h
```diff
@@ -175,8 +175,26 @@ class Model
...
+  {
+    _metadatas.emplace(name, data);
+  }
+
+  bool is_metadata_exist(const std::string &name) const
```
glistening, line 185: What about `has_metadata`?

```suggestion
  bool has_metadata(const std::string &name) const
```

runtime/onert/core/include/ir/Model.h
glistening: Just my curiosity: I didn't see `datas` often before. it looked a bit, perhaps data is already plural form. Anyway, s after `data` helps me that it is a collection of `metadata`. : )

runtime/onert/core/include/ir/Model.h
ragmani: How about?
```suggestion
  bool exists_metadata(const std::string &name) const
```

glistening: I am okay both `has_metadata()` and `exists_metadata()`. I also thought `exists_metadata` as another candidate.

runtime/onert/core/include/ir/Model.h
zetwhite: I've also thought about `metadatas`.
For me, `Metadata` looks like one word (not `meta` and `data`), and I'd like to emphasize a 'collection' of metadata.
So, As you guessed, I tried to use `metadatas`, despite it is grammarly a bit wrong.

ragmani (dismissed):
LGTM

glistening:
I am waiting for `is_metadata_exist` is renamed.

runtime/onert/core/include/ir/Model.h
zetwhite: updated to use `unique_ptr`

runtime/onert/core/include/ir/Model.h
zetwhite: thanks for the suggestion! 
updated.

zetwhite:
updated method name `exists_metadata` - https://github.com/Samsung/ONE/pull/12294/commits/2eddc0623dba5235d992132210c6f0a5d67225b2 
updated get_metadata to return `unique_ptr` - https://github.com/Samsung/ONE/pull/12294/commits/2eddc0623dba5235d992132210c6f0a5d67225b2

runtime/onert/core/include/ir/Model.h
```diff
@@ -177,23 +177,30 @@ class Model
...
+  std::unique_ptr<const ir::Data> get_metadata(const std::string name)
   {
-    return _metadatas.at(name);
+    auto m = _metadatas.find(name);
+
+    if (m == _metadatas.end())
+      throw std::out_of_range{"no meatdata named " + name};
+
+    auto data = std::move(m->second);
+    _metadatas.erase(m);
```
glistening, line 198: @zetwhite It doesn't look intuitive to erase the metadata after `get_metadata`. Do we need to return `unique_ptr`?

hseok-oh (dismissed):

runtime/onert/core/include/ir/Model.h
zetwhite: Umm, I've also thought about returning "const raw ptr" instead of `unique_ptr`.

I think that returning `unique_ptr` is much safer.
I'm a little afraid that under situation happened with "const raw ptr". 

```cpp

 // load model from file
std::unique_ptr<ir::Model> model = laodModel();
 
// get buffer
const ir::Data* metadata_buffer = model->get_metadata("CIRCLE_TRAINING"); 

// delete a model
// Since the model holds unique_ptr of metadata, all metadata are also deleted.
model.reset(); 

// access already freed memory space
metadata_buffer->base
```

zetwhite: Since `ir::Model` doesn't know how methods can be used, I thought it is better to use `unique_ptr` for safety. 
Do you have better idea about this, plz let me know! 

+) If erasing metadata after`get_metadata` is un-intutive, we could find a bettter naming instead of `get_metadata`- :thinking: :thinking:  maybe `extract_metadata`?

ragmani: `extract_metadata` looks better to me.

glistening: I don't object to use `unique_ptr`. `unique_ptr` is good thing to prevent memory leak. I just don't get the user surprised the metadata is removed after `get_metadata`. If you choose to remove metadata (though I don't like it), it would be better to rename which says clearly the metadata will not be available once you get it. At least for me, `extract` is not good one. I tried to find more appropriate word, but I don't have idea. 

Once again, I don't oppose this PR if it gets two approvals. It will be merged.

zetwhite: > How about introducing a class for metadata into onert? If so, you can prevent the problem you worried about by adjusting the introduced class (i.g. introducing getting const reference of metadata class into ir::Model and deleting copy constructor of metadata class). Furthermore, you can take advantage of easily adding functions related to metadata.

 @ragmani Thanks for thinking together. I failed to fully understand your suggestion. 
If we return a 'const reference of metadata' from `get_metadata`, Does it extend the lifetime of metadata, even if the model instance is deleted? 

> Furthermore, you can take advantage of easily adding functions related to metadata. 

(not important, but for sharing opinion) 
Umm.. I think there are not many functions needed for metadata. Since metadata in a model is just a buffer(not parsed yet), there aren't many common things between metadata. 
So, This advantage is not very attractive to me.

zetwhite: For now, I'd like to rename `get_metadata` to `extract_metadata` and add some comments for `ir::Model` user.

If I could get 2 approvals, let it proceed. 
Or Let me try to find another solution. 

(note) I get the word `extract` from c++17 std.
>  unlinks the node that contains the element pointed to by position and returns a node handle that owns it. 
(https://en.cppreference.com/w/cpp/container/map/extract)

glistening: Wow, I didn't know `map.extract()`. Then, we may replace the body of `extract_metadata`:
```cc
    auto m = _metadatas.find(name);

    if (m == _metadatas.end())
      throw std::out_of_range{"no meatdata named " + name};

    auto data = std::move(m->second);
    _metadatas.erase(m);
    return data;
```
with:
```cc
   // After c++17
   return std::move(_metadatas.extract(name).mapped());
```

runtime/onert/frontend/base_loader/include/base_loader.h
```diff
@@ -243,6 +245,31 @@ void BaseLoader<LoaderDomain>::BaseLoader::loadFromBuffer(uint8_t *buffer, size_
...
+    size_t data_size = data->size();
+    ptrdiff_t offset_start = data->data() - _base;
+    ptrdiff_t offset_end = offset_start + data_size;
+
+    ptrdiff_t page_start = (offset_start / _pagesize) * _pagesize;
+    size_t mapping_size = offset_end - page_start;
+
+    // Since metadata is not access often in inference/training time, always use mmaped-data
+    // Ref : https://github.com/Samsung/ONE/issues/3961#issuecomment-681750231
+    return std::make_unique<ir::MMapedData>(_fd, page_start, mapping_size, offset_start, data_size);
```
glistening, line 269: `_fd` must outlive returned `std::unique_ptr<ir::Data>`. Currently @zetwhite is going to convert `ir::Data` to `TrainingInfo` in model loading phase. Thus, it will be safe at this moment.

zetwhite: Ah, I also worried about this part. 

As I searched, there is no need to keep `_fd` open after mmaped.

--------------------------

I refered this : https://stackoverflow.com/questions/17490033/do-i-need-to-keep-a-file-open-after-calling-mmap-on-it  


* From Posix manpage [ref](https://pubs.opengroup.org/onlinepubs/7908799/xsh/mmap.html#:~:text=The%20mmap()%20function%20adds%20an%20extra%20reference%20to%20the%20file%20associated%20with%20the%20file%20descriptor%20fildes%20which%20is%20not%20removed%20by%20a%20subsequent%20close()%20on%20that%20file%20descriptor.%20This%20reference%20is%20removed%20when%20there%20are%20no%20more%20mappings%20to%20the%20file.) 

> The mmap() function adds an extra reference to the file associated with the file descriptor fildes which is not removed by a subsequent close() on that file descriptor. This reference is removed when there are no more mappings to the file. 

* From Linux manpage [ref](https://linux.die.net/man/2/mmap#:~:text=On%20the%20other%20hand%2C%20closing%20the%20file%20descriptor%20does%20not%20unmap%20the%20region.) 

> On the other hand, closing the file descriptor does not unmap the region.

runtime/onert/core/include/ir/Model.h
ragmani: > If we return a 'const reference of metadata' from get_metadata, Does it extend the lifetime of metadata, even if the model instance is deleted?

No, it does not extend lifetime of metadata. If the reference exists and the model instance is deleted, the reference is a dangling reference.

> (not important, but for sharing opinion)
Umm.. I think there are not many functions needed for metadata. Since metadata in a model is just a buffer(not parsed yet), there aren't many common things between metadata.
So, This advantage is not very attractive to me.

I see, it was just a suggestion.

glistening (approved):
LGTM :+1:

ragmani (approved):
LGTM

PR #12296, opened 2023-12-14
[onert/test] Add util to print nnfw_train_info

zetwhite:
This PR aims to print nnfw_train_info in onert_train.

### DIFF
--- a/tests/tools/onert_train/src/nnfw_util.cc
+++ b/tests/tools/onert_train/src/nnfw_util.cc
@@ -16,7 +16,10 @@
 
 #include <cassert>
 #include <string>
+#include <iostream>
+
 #include "nnfw.h"
+#include "nnfw_experimental.h"
 
 namespace onert_train
 {
@@ -44,6 +47,73 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
     sizeof(int16_t), /* NNFW_TYPE_TENSOR_QUANT16_SYMM_SIGNED = 7 */
   };
   return elmsize[ti->dtype] * num_elems(ti);
+}
+
+std::ostream &operator<<(std::ostream &os, const NNFW_TRAIN_OPTIMIZER &opt)
+{
+  switch (opt)
+  {
+    case NNFW_TRAIN_OPTIMIZER_ADAM:
+      os << "adam";
+      break;
+    case NNFW_TRAIN_OPTIMIZER_SGD:
+      os << "sgd";
+      break;
+    default:
+      os << "unsupported optimzier";
+  }
+  return os;
+}
+
+std::ostream &operator<<(std::ostream &os, const NNFW_TRAIN_LOSS &loss)
+{
+  switch (loss)
+  {
+    case NNFW_TRAIN_LOSS_MEAN_SQUARED_ERROR:
+      os << "mean squared error";
+      break;
+    case NNFW_TRAIN_LOSS_CATEGORICAL_CROSSENTROPY:
+      os << "categorical crossentropy";
+      break;
+    default:
+      os << "unsupported loss";
+  }
+  return os;
+}
+
+std::ostream &operator<<(std::ostream &os, const NNFW_TRAIN_LOSS_REDUCTION &loss_reduction)
+{
+  switch (loss_reduction)
+  {
+    case NNFW_TRAIN_LOSS_REDUCTION_INVALID:
+      os << "invalid (use default)";
+      break;
+    case NNFW_TRAIN_LOSS_REDUCTION_SUM_OVER_BATCH_SIZE:
+      os << "sum over batch size";
+      break;
+    case NNFW_TRAIN_LOSS_REDUCTION_SUM:
+      os << "sum";
+      break;
+    default:
+      os << "unsupported reduction type";
+      break;
+  }
+  return os;
+}
+
+std::ostream &operator<<(std::ostream &os, const nnfw_loss_info &loss_info)
+{
+  os << "{loss= " << loss_info.loss << ", type= " << loss_info.reduction_type << "}";
+  return os;
+}
+
+std::ostream &operator<<(std::ostream &os, const nnfw_train_info &info)
+{
+  os << "- learning_rate   = " << info.learning_rate << "\n";
+  os << "- batch_size      = " << info.batch_size << "\n";
+  os << "- loss_info       = " << info.loss_info << "\n";
+  os << "- optimizer       = " << info.opt << "\n";
+  return os;
 }
 
 } // namespace onert_train
--- a/tests/tools/onert_train/src/nnfw_util.h
+++ b/tests/tools/onert_train/src/nnfw_util.h
@@ -18,6 +18,7 @@
 #define __ONERT_TRAIN_NNFW_UTIL_H__
 
 #include "nnfw.h"
+#include "nnfw_experimental.h"
 
 #define NNPR_ENSURE_STATUS(a)        \
   do                                 \
@@ -32,6 +33,7 @@ namespace onert_train
 {
 uint64_t num_elems(const nnfw_tensorinfo *ti);
 uint64_t bufsize_for(const nnfw_tensorinfo *ti);
-} // end of namespace onert_train
 
+std::ostream &operator<<(std::ostream &os, const nnfw_train_info &info);
+} // end of namespace onert_train
 #endif // __ONERT_TRAIN_NNFW_UTIL_H__
--- a/tests/tools/onert_train/src/onert_train.cc
+++ b/tests/tools/onert_train/src/onert_train.cc
@@ -168,6 +168,9 @@ int main(const int argc, char **argv)
     tri.loss_info.reduction_type = convertLossReductionType(args.getLossReductionType());
     tri.opt = convertOptType(args.getOptimizerType());
 
+    std::cout << "== training parameter ==" << std::endl;
+    std::cout << tri;
+    std::cout << "========================" << std::endl;
     // prepare execution
 
     // TODO When nnfw_{prepare|run} are failed, can't catch the time


tests/tools/onert_train/src/onert_train.cc
```diff
@@ -168,6 +168,9 @@ int main(const int argc, char **argv)
     tri.loss_info.reduction_type = convertLossReductionType(args.getLossReductionType());
     tri.opt = convertOptType(args.getOptimizerType());
 
+    std::cout << "== training parameter ==" << std::endl;
+    std::cout << tri;
+    std::cout << "========================" << std::endl;
```
zetwhite, line 173: printing example
``` 
➜ ~/Workspace/ONE/Product/out/bin/onert_train --modelfile mnist_with_meta.circle --load_input:raw pool_mnist_model/mnist_x_train.1000.bin --load_expected:raw pool_mnist_model/mnist_y_train.1000.bin --batch_size=10 --optimizer=1
Model Expected Filename pool_mnist_model/mnist_y_train.1000.bin
Model Input Filename pool_mnist_model/mnist_x_train.1000.bin
Model Filename mnist_with_meta.circle
== training parameter ==
- learning_rate   = 0.001
- batch_size      = 10
- loss_info       = {loss = mean squared error, type = invalid (use default)}
- optimizer       = adam
========================
Epoch 1/5 - time: 7.842ms/step - loss: [0] 0.0675
Epoch 2/5 - time: 7.843ms/step - loss: [0] 0.0359
Epoch 3/5 - time: 7.801ms/step - loss: [0] 0.0260
Epoch 4/5 - time: 7.800ms/step - loss: [0] 0.0197
Epoch 5/5 - time: 7.840ms/step - loss: [0] 0.0157
MODEL_LOAD   takes 1.3320 ms
PREPARE      takes 9.8060 ms
EXECUTE      takes 3935.2690 ms
- Epoch 1      takes 784.1750 ms
- Epoch 2      takes 784.3300 ms
- Epoch 3      takes 780.0620 ms
- Epoch 4      takes 779.9520 ms
- Epoch 5      takes 784.0440 ms
```

tests/tools/onert_train/src/nnfw_util.h
```diff
@@ -32,6 +33,7 @@ namespace onert_train
 {
 uint64_t num_elems(const nnfw_tensorinfo *ti);
 uint64_t bufsize_for(const nnfw_tensorinfo *ti);
-} // end of namespace onert_train
 
+std::ostream &operator<<(std::ostream &os, const nnfw_train_info &info);
```
zetwhite, line 37: ### Background 

`nnfw_train_info` is a structure that appears to API-user, to set/get the training parameters.

After importer implemented, There are many ways to set training parameters : 
   * using cmd line input
   * using model file (metadata include traininginfo)
   * if nothing given, default parameter will be used

So, I'd like to
   * print training parameter in onert_train to show which one is selected
   * hide detailed print-logic to keep main function simple -> so, hide detailed things in nnfw_util.c

zetwhite: @jyoungyun Could you share your opinion about this PR?

tests/tools/onert_train/src/nnfw_util.cc
```diff
@@ -46,4 +49,71 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
...
+  switch (opt)
+  {
+    case NNFW_TRAIN_OPTIMIZER_ADAM:
+      os << "adam";
+      break;
+    case NNFW_TRAIN_OPTIMIZER_SGD:
+      os << "sgd";
+      break;
+    default:
+      os << "unsupported optimzier";
```
jyoungyun, line 63:
```suggestion
      os << "unsupported optimizer";
      break;
```

tests/tools/onert_train/src/nnfw_util.cc
```diff
@@ -46,4 +49,71 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
...
+  switch (loss)
+  {
+    case NNFW_TRAIN_LOSS_MEAN_SQUARED_ERROR:
+      os << "mean squared error";
+      break;
+    case NNFW_TRAIN_LOSS_CATEGORICAL_CROSSENTROPY:
+      os << "categorical crossentropy";
+      break;
+    default:
+      os << "unsupported loss";
```
jyoungyun, line 79:
```suggestion
      os << "unsupported loss";
      break;
```

tests/tools/onert_train/src/nnfw_util.cc
```diff
@@ -46,4 +49,71 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
...
+    default:
+      os << "unsupported reduction type";
+      break;
+  }
+  return os;
+}
+
+std::ostream &operator<<(std::ostream &os, const nnfw_loss_info &loss_info)
+{
+  os << "{loss= " << loss_info.loss << ", type= " << loss_info.reduction_type << "}";
```
jyoungyun, line 106:
```suggestion
  os << "{loss= " << loss_info.loss << ", reduction_type= " << loss_info.reduction_type << "}";
```

tests/tools/onert_train/src/nnfw_util.cc
```diff
@@ -46,4 +49,71 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
...
+  }
+  return os;
+}
+
+std::ostream &operator<<(std::ostream &os, const NNFW_TRAIN_LOSS_REDUCTION &loss_reduction)
+{
+  switch (loss_reduction)
+  {
+    case NNFW_TRAIN_LOSS_REDUCTION_INVALID:
+      os << "invalid (use default)";
```
jyoungyun, line 89:
```suggestion
      os << "use default setting (model parameter or ONERT train setting)";
```

jyoungyun (commented):
Some of the comments I wrote make the string long. Please modify it properly.

glistening:
@zetwhite Why don't you put this dump logic for `nnfw_train_info` under somewhere under `api`? It is likely to be used out of `onert_train`.

Also, I want to move `nnfw_util.*` to somewhere under `api` as it is used both `onert_run` and `onert_train`, and it started to diverse.

glistening (dismissed):
As I mentioned in https://github.com/Samsung/ONE/pull/12296#issuecomment-1857305218, I think the helper function could be located under `api`. But it is okay to move later when it becomes necessary.

tests/tools/onert_train/src/nnfw_util.cc
zetwhite: updated

tests/tools/onert_train/src/nnfw_util.cc
zetwhite: updated

`reduction_type` is a bit long to display, I rather choose `reduction`.

tests/tools/onert_train/src/nnfw_util.cc
zetwhite: updated!

tests/tools/onert_train/src/nnfw_util.cc
```diff
@@ -46,4 +49,73 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
...
```
zetwhite, line 91: I've thought about this part. 

Since it is a "dump" function, its role is just printing out the 'nnfw_train_info' value.
Interpreting the 'invalid' meaning as 'use default' or 'use model parameter' is a bit out of this function's role. 

So, Renaming `NNFW_TRAIN_LOSS_REDUCTION_INVALID` into `NNFW_TRAIN_LOSS_REUDCTION_DEFAULT` looks a better option for me.  
After renaming the enum, fixing this dump function to print 'use default option' (instead of printing invalid) looks good to me. 

> So, Renaming `NNFW_TRAIN_LOSS_REDUCTION_INVALID` into `NNFW_TRAIN_LOSS_REUDCTION_DEFAULT` looks a better option for me.   

It is an out of this PR purpose, so let me fix it later.

tests/tools/onert_train/src/nnfw_util.cc
zetwhite: I tried it, but it looks too long to display.

I append my opinion about how to print the `INVALID` option here - https://github.com/Samsung/ONE/pull/12296#discussion_r1429608402 

So, How about merging this PR as-is, and trying to reconfigure it later? 
Since I also have a plan to introduce a default(invalid) enum type for other configurations (`NNFW_TRAIN_OPTIMIZER`, `NNFW_TRIAN_LOSS`), I think i could fix it together.

jyoungyun: Well, however this `invalid` option means using model parameter or ONERT training setting. If we leave a `invalid` log about this option, the user may think that this option is invalid. How about using `use default setting` only for display?

zetwhite:
> @zetwhite Why don't you put this dump logic for nnfw_train_info under somewhere under api? It is likely to be used out of onert_train.

@glistening I checked some space for dump function in api.

```bash
➜ tree include  

 ├──      nnfw_experimental.h  
 ├──      nnfw.h  
 ├──      nnfw_internal.h  
 ├──      nnfw_uitl.h  
 └──      nnfw_version.h  

```

Since all functions listed in `nnfw_experimental.h` and `nnfw.h` return NNFW_STATUS, Moving the dump function in those headers looks inappropriate.
So, I'm thinking about adding a new header named `nnfw_util.h` for the utility things.
Thinking about `nnfw_util.h`, I have a question -  Does utility API have to support C interface?

zetwhite:
> So, I'm thinking about adding a new header named nnfw_util.h for the utility things.
 
If so, it might need to change. (because the current PR use `ostream` which is only supported in C++)
if so, it might take some time, so I'd like to move it into `api` later.

tests/tools/onert_train/src/nnfw_util.cc
zetwhite: >  If we leave a invalid log about this option, the user may think that this option is invalid. 

I agree with this point.
updated!

jyoungyun (approved):
LGTM

jyoungyun:
> @zetwhite Why don't you put this dump logic for `nnfw_train_info` under somewhere under `api`? It is likely to be used out of `onert_train`.
> 
> Also, I want to move `nnfw_util.*` to somewhere under `api` as it is used both `onert_run` and `onert_train`, and it started to diverse.

Change of `nnfw_util.*` can also be used in the onert when the user enables the `ONERT_LOG_ENABLE` option. I think the purpose of this PR just shows the parameter of `onert_train`. So I approve this PR as it is.

glistening (approved):
LGTM

Issue #12293, opened 2023-12-14
[onert] Remove #ifdef ONERT_TRAIN


glistening:
We need to remove `#ifdef ONERT_TRAIN` as much as possible.
Hopefully, remove the whole occurrences.

They are introduced when we're doing prototyping and in a hurry.

I am not the author of many of #ifdef ONERT_TRAIN.
It needs the help of the original authors.

- [ ] getBufSize (nnfw_api_internal.cc)
- [ ] ...

glistening:
I would like to start with seemingly simple one.

@jyoungyun May I ask why `getBufSize(...)` needs to be enclosed with `#ifdef ONERT_TRAIN`.

https://github.com/Samsung/ONE/blob/11620233e1a303dcc60f48fc7103cb55d84270cc/runtime/onert/api/src/nnfw_api_internal.cc#L199-L221

It is not training-dependent code.

We may move the following existing code into somewhere.

https://github.com/Samsung/ONE/blob/11620233e1a303dcc60f48fc7103cb55d84270cc/tests/tools/onert_run/src/nnfw_util.cc#L23-L47

jyoungyun:
> May I ask why `getBufSize(...)` needs to be enclosed with `#ifdef ONERT_TRAIN`.

When we first started ONERT training, we decided to use `ONERT_TRAIN` definition if the code is used for only ONERT train. So I wrote the code like this.

At the beginning of the project, there were many concerns about the direct merge of ONERT training code   into ONERT from the perspective of file size, memory size, release criteria and performance. Is there anything more concerns about these?

glistening:
Closed as #12301 is merged.

PR #12297, opened 2023-12-14
[onert] Move private static method from ExecutorFactory.h

glistening:
It moves private static method implementation from ExecutorFactory.h to
ExecutorFactory.cc.

Related: #12293

### DIFF
--- a/runtime/onert/core/src/compiler/ExecutorFactory.h
+++ b/runtime/onert/core/src/compiler/ExecutorFactory.h
@@ -72,19 +72,6 @@ private:
   ExecutorFactory();
 
 private:
-  static void prepareMigrantTensors(compiler::ILoweredGraph &lowered_graph,
-                                    const backend::BackendContexts &backend_contexts);
-  static void prepareBuiltinBackend(const TensorRegistries &tensor_regs,
-                                    const std::shared_ptr<exec::IExecutors> &executors,
-                                    const backend::BackendContexts &backend_contexts,
-                                    const ir::ModelIndex &index);
-  static std::deque<std::pair<const backend::Backend *, backend::BackendContext *>>
-  orderBackendContext(const backend::BackendContexts &backend_contexts);
-
-  static exec::IExecutor *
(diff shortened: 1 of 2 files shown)


runtime/onert/core/src/compiler/ExecutorFactory.h
```diff
@@ -21,14 +21,7 @@
...
 #include "backend/ITensor.h"
 
-#ifdef ONERT_TRAIN
-#include "backend/train/TrainableBackendContext.h"
-#endif // ONERT_TRAIN
 #include "compiler/LoweredGraph.h"
-#ifdef ONERT_TRAIN
-#include "compiler/train/LoweredTrainableGraph.h"
-#include "compiler/train/TrainingInfo.h"
-#endif // ONERT_TRAIN
```
glistening, line 31: #ifdef ... and include train related-headers are replaced with forward declaration below.

runtime/onert/core/src/compiler/ExecutorFactory.h
```diff
@@ -69,36 +71,6 @@ class ExecutorFactory
...
-  static void
-  prepareMigrantTensors(compiler::ILoweredGraph &lowered_graph,
-                        const backend::train::TrainableBackendContexts &backend_contexts);
-  static exec::IExecutor *
-  createTrainableExecutor(std::unique_ptr<compiler::train::LoweredTrainableGraph> lowered_graph,
-                          const std::shared_ptr<exec::IExecutors> &executors,
-                          const ExecutorFactoryArgs &args,
-                          const compiler::train::TrainingInfo &training_info);
-#endif // ONERT_TRAIN
-
```
glistening, line 101: Remove the private methods declaration.

glistening:
Changes in `ExecutorFactory.cc` are 
- the reordering. That is, put `create...` first, then ExecutorFactory constructor comes.
- Remove `ExecutionFactory::`. It is not the static method of ExecutorFactory any longer. It is now static function.

glistening:
After #12301, it is not necessary.

PR #12300, opened 2023-12-14
[onert] Introduce nnfw_train_set_output api

jyoungyun:
This commit introduces nnfw_train_set_output api.
If user sets the output buffer using this function, user can obtain the output result executed with trained weight.


glistening (approved):
LGTM

PR #12302, opened 2023-12-14
[tools/onert_train] Add metric argument

jyoungyun:
This commit adds metric argument.
It simply calculates the metric value using the variables.

### DIFF
--- a/tests/tools/onert_train/src/args.cc
+++ b/tests/tools/onert_train/src/args.cc
@@ -228,6 +228,9 @@ void Args::Initialize(void)
       "Optimizer type\n"
       "0: SGD (default)\n"
       "1: Adam\n")
+    ("metric", po::value<int>()->default_value(-1)->notifier([&] (const auto &v) { _metric_type = v; }),
+      "Metricy type\n"
+      "  Simply calculates the metric value using the variables (default: none)\n")
     ("verbose_level,v", po::value<int>()->default_value(0)->notifier([&](const auto &v) { _verbose_level = v; }),
          "Verbose level\n"
          "0: prints the only result. Messages btw run don't print\n"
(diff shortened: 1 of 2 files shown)


tests/tools/onert_train/src/args.cc
```diff
@@ -228,6 +228,9 @@ void Args::Initialize(void)
       "Optimizer type\n"
       "0: SGD (default)\n"
       "1: Adam\n")
+    ("metric", po::value<int>()->default_value(-1)->notifier([&] (const auto &v) { _metric_type = v; }),
+      "Metricy type\n"
+      "  Simply calculates the metric value using the variables (default: none)\n")
```
glistening, line 233: @jyoungyun Apparently, there is no choice at this moment. Do we need this argument at this time?

jyoungyun: Yes, after other functions are merged, it will be a meaningful argument.
If you want, I can add `categorical accuracy` type even though there is no concrete implementation now.

glistening (approved):
LGTM

tests/tools/onert_train/src/args.h
```diff
@@ -59,6 +59,7 @@ class Args
   const int getLossType(void) const { return _loss_type; }
   const int getLossReductionType(void) const { return _loss_reduction_type; }
   const int getOptimizerType(void) const { return _optimizer_type; }
+  const int getMetricType(void) const { return _metric_type; }
```
ragmani, line 62: (Optional) Returning `const int` value is meaningless.
```suggestion
  int getMetricType(void) const { return _metric_type; }
```

ragmani (approved):
LGTM

tests/tools/onert_train/src/args.h
hseok-oh: Let's change this after merging this PR.

PR #12265, opened 2023-12-12
[Draft] Implement Validation pass for train (2nd version)

jyoungyun:
Related issue: #12204

PR #12303, opened 2023-12-14
[tools/onert_train] Introduce Metric class

jyoungyun:
This commit introduces Metric class.
Metric class calculates the metric value using the output and expected data. Only categorical accuracy is supported now.

Draft: https://github.com/Samsung/ONE/pull/12265

### DIFF
--- a/tests/tools/onert_train/CMakeLists.txt
+++ b/tests/tools/onert_train/CMakeLists.txt
@@ -12,7 +12,7 @@ list(APPEND ONERT_TRAIN_SRCS "src/nnfw_util.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/randomgen.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/rawformatter.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/rawdataloader.cc")
-list(APPEND ONERT_TRAIN_SRCS "src/metric.cc")
+list(APPEND ONERT_TRAIN_SRCS "src/metrics.cc")
 
 nnfw_find_package(Boost REQUIRED program_options)
 nnfw_find_package(HDF5 QUIET)
--- a/tests/tools/onert_train/CMakeLists.txt
+++ b/tests/tools/onert_train/CMakeLists.txt
@@ -12,6 +12,7 @@ list(APPEND ONERT_TRAIN_SRCS "src/nnfw_util.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/randomgen.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/rawformatter.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/rawdataloader.cc")
+list(APPEND ONERT_TRAIN_SRCS "src/metric.cc")
 
 nnfw_find_package(Boost REQUIRED program_options)
 nnfw_find_package(HDF5 QUIET)
--- a/tests/tools/onert_train/src/metric.cc
+++ b/tests/tools/onert_train/src/metric.cc
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "metric.h"
+#include "nnfw_util.h"
+
+#include <algorithm>
+#include <stdexcept>
+
+namespace onert_train
+{
+
+Metric::Metric(const std::vector<Allocation> &output, const std::vector<Allocation> &expected,
+               const std::vector<nnfw_tensorinfo> &infos)
+  : _output{output}, _expected{expected}, _infos{infos}
+{
+  // DO NOTHING
+}
+
+template <typename T>
+float Metric::categoricalAccuracy(const T *output, const T *expected, uint32_t batch, uint64_t size)
+{
+  int correct = 0;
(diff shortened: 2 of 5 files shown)


tests/tools/onert_train/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ list(APPEND ONERT_TRAIN_SRCS "src/nnfw_util.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/randomgen.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/rawformatter.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/rawdataloader.cc")
+list(APPEND ONERT_TRAIN_SRCS "src/metric.cc")
```
glistening, line 15: (beyond this PR) @jyoungyun What about to use glob? instead of adding each file. It is the usual way we use under `/runtime/onert`. Previously, for `onert_run`, I've appended all the files in `onert_run`. But `onert_train` has more files, and it is likely to have more files.

tests/tools/onert_train/src/metric.cc
```diff
@@ -0,0 +1,66 @@
...
+
+Metric::Metric(const std::vector<Allocation> &output, const std::vector<Allocation> &expected,
+               const std::vector<nnfw_tensorinfo> &infos)
+  : _output{output}, _expected{expected}, _infos{infos}
+{
+  // DO NOTHING
+}
+
+template <typename T>
+float Metric::categoricalAccuracy(const T *output, const T *expected, int32_t batch, int32_t size)
```
glistening, line 34: @jyoungyun May I ask why we use `int32_t` for `size`?
- signed type halves the max size.
- Also, we need to check whether it is negative. (Currently, it does not check.)

tests/tools/onert_train/CMakeLists.txt
jyoungyun: We can use `glob` to collect all cc files. I will fix it later.

tests/tools/onert_train/src/metric.cc
jyoungyun: I fixed it to use `uint64_t` for `size`.

tests/tools/onert_train/src/metric.h
```diff
@@ -0,0 +1,47 @@
...
+private:
+  template <typename T>
+  float categoricalAccuracy(const T *output, const T *expected, uint32_t batch, uint64_t size);
+
+public:
+  float categoricalAccuracy(int32_t index);
+
+private:
+  const std::vector<Allocation> &_output;
+  const std::vector<Allocation> &_expected;
```
glistening, line 41: @jyoungyun Is it necessary to have `_output` and `_expected` as members? `categoricalAccuracy(const T *output, const T *expected, uint32_t batch, uint64_t size)` seems to have everything needs to calculate. I expect `Metric` is set of algorithms for calculating several kind of accuracies.

jyoungyun: Yes, the metric class contains the _output and _expected results as members to access each index data easily.
Here is an example for this class. I want to match the usability of this function with `nnfw_train_get_loss` function.

```
            // get validation loss and accuracy
            Metric metric(output_data, expected_data, expected_infos);
            for (int32_t i = 0; i < num_expecteds; ++i)
            {
              float temp = 0.f;
              NNPR_ENSURE_STATUS(nnfw_train_get_loss(session, i, &temp));
              losses[i] += temp;
              if (args.getMetricType() == 0)
                metrics[i] += metric.categoricalAccuracy(i);
            }
          }
```

chunseoklee (approved):

tests/tools/onert_train/src/metric.cc
```diff
@@ -0,0 +1,66 @@
...
+
+template <typename T>
+float Metric::categoricalAccuracy(const T *output, const T *expected, uint32_t batch, uint64_t size)
+{
+  int correct = 0;
+  for (int b = 0; b < batch; ++b)
+  {
+    int b_offset = b * size;
+    std::vector<T> boutput(output + b_offset, output + b_offset + size);
+    std::vector<T> bexpected(expected + b_offset, expected + b_offset + size);
```
ragmani, line 41: (Optional)
```suggestion
    int begin_offset = b * size;
    int end_offset = begin_offset + size;
    std::vector<T> boutput(output + b_offset, output + b_offset + size);
    std::vector<T> bexpected(expected + b_offset, expected + b_offset + size);
```

tests/tools/onert_train/src/metric.cc
```diff
@@ -0,0 +1,66 @@
...
+
+float Metric::categoricalAccuracy(int32_t index)
+{
+  auto batch = _infos[index].dims[0];
+  auto size = num_elems(&_infos[index]) / batch;
+  switch (_infos[index].dtype)
+  {
+    case NNFW_TYPE_TENSOR_FLOAT32:
+      return categoricalAccuracy(static_cast<const float *>(_output[index].data()),
+                                 static_cast<const float *>(_expected[index].data()), batch, size);
```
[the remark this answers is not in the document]
ragmani: @jyoungyun 
I guess it's not a similar case. A void pointer in C++ is a special pointer that can point to objects of any data type, but`uint8_t` pointer is not a special poiter. So I'm not sure if `static_cast`ing between pointers to other numerical types is always valid.
According to `https://www.mooclab.club/threads/how-do-you-explain-the-differences-among-static_cast-reinterpret_cast-const_cast-and.4079/`, "`static_cast` performs implicit conversions, the reverses of implicit standard conversions, and (possibly unsafe) base to derived conversions." and "reinterpret_cast converts one pointer to another without changing the address, or converts between pointers and their numerical (integer) values."
I recommend using `reinterpret_cast` for casting between pointers to other numerical types.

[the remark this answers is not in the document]
ragmani: I talked offline with @jyoungyun . It was my mistaken. The returned value of `_output[index].data()` is `void *`, not `uint8_t *`. If so, this code looks fine to me.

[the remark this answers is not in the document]
jyoungyun: The `_output[index].data()` returns the void pointer. This data class is implemented on the `onert_train` itself. :) If the `data()` returns `uint8_t` data, it should use `reinterpret_cast` as you said. However, in this case, the void pointer is returned, so I think `static_cast` is correct for this case.

ragmani (approved):
LGTM

glistening (approved):
LGTM

PR #12318, opened 2023-12-18
[luci/pass] Introduce ResovleFormerCustomOpPass 

Seunghui98:
This commit introduces ResolveFormerCustomOpPass to luci.
ResolveFormerCustomOpPass changes a specific custom op into a builtin op from schema version upgrade.

---

### DIFF
--- a/compiler/luci/pass/include/luci/CircleOptimizer.h
+++ b/compiler/luci/pass/include/luci/CircleOptimizer.h
@@ -52,6 +52,7 @@ public:
       ResolveCustomOpMatMul,
       ResolveCustomOpMaxPoolWithArgmax,
       ResolveCustomOpSplitV,
+      ResolveFormerCustomOp,
       FoldAddV2,
       FoldCast,
       FoldDensify,
--- a/compiler/luci/pass/include/luci/Pass/ResolveFormerCustomOpPass.h
+++ b/compiler/luci/pass/include/luci/Pass/ResolveFormerCustomOpPass.h
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_RESOLVE_FORMER_CUSTOM_OP_PASS_H__
+#define __LUCI_RESOLVE_FORMER_CUSTOM_OP_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief Class to convert a custom operator to a built-in operator.
+ *
+ * @details This pass changes a op formerly used as a custom op to builtin op
+ *          from schema version upgrade.
+ */
+struct ResolveFormerCustomOpPass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::ResolveFormerCustomOpPass"; }
+
+  bool run(loco::Graph *g) final;
+};
+
+} // namespace luci
+
+#endif // __LUCI_RESOLVE_FORMER_CUSTOM_OP_PASS_H__
--- a/compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
+++ b/compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
@@ -1,0 +1,99 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/ResolveFormerCustomOpPass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/AttrFusedActFunc.h>
+#include <luci/Profile/CircleNodeOrigin.h>
+
+#include <flatbuffers/flexbuffers.h>
+
+namespace
+{
+
+bool resolve_with_BroadcastTo(luci::CircleCustom *node)
+{
+  // check if the number of inputs is 2.
+  if (node->numInputs() != 2)
+    return false;
+
+  auto input = loco::must_cast<luci::CircleNode *>(node->inputs(0));
+
+  // check if shape are support data types
+  auto shape = loco::must_cast<luci::CircleNode *>(node->inputs(1));
+  if (shape->dtype() != loco::DataType::S32 && shape->dtype() != loco::DataType::S64)
+    return false;
+
+  auto name = node->name();
+  assert(name.length() > 0);
+
+  auto broadcastTo = node->graph()->nodes()->create<luci::CircleBroadcastTo>();
+  broadcastTo->input(input);
+  broadcastTo->shape(shape);
+  broadcastTo->name(name);
+  luci::add_origin(broadcastTo, luci::get_origin(node));
+
+  auto customOut = loco::succs(node);
+
+  assert(customOut.size() == 1);
+
+  // check if the data type of output is same with the one of the input feature map.
+  auto output = loco::must_cast<luci::CircleNode *>(*customOut.begin());
+  if (input->dtype() != output->dtype())
+    return false;
+
+  replace(*customOut.begin()).with(broadcastTo);
+
+  return true;
+}
+
+bool resolve_custom_op(luci::CircleCustom *node)
+{
+  const std::string custom_code = node->custom_code();
+
+  if (custom_code == "BroadcastTo")
+  {
+    return resolve_with_BroadcastTo(node);
+  }
+  // TODO add more custom codes
+
+  return false;
+}
+
+} // namespace
+
+namespace luci
+{
+
+bool ResolveFormerCustomOpPass::run(loco::Graph *g)
+{
+  bool changed = false;
+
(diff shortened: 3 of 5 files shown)


compiler/luci/pass/include/luci/Pass/ResolveCustomOpPass.h
```diff
@@ -0,0 +1,40 @@
...
+  ResolveCustomOpPass() {}
+
+  const char *name(void) const final { return "luci::ResolveCustomOpPass"; }
+
+  bool run(loco::Graph *g) final;
+};
+
+} // namespace luci
+
+#endif // __LUCI_RESOLVE_CUSTOM_OP_PASS_H__
```
Seunghui98, line 40: The overall description of the new pass is as follows.

- The discussion of the New pass, `ResovleCustomOpPass`, came from here.
-> https://github.com/Samsung/ONE/issues/10784#issuecomment-1797862821
- This new pass is a pass that changes op, which was previously supported by `custom` op, 
when it can now be supported by `built-in` op.

compiler/luci/pass/include/luci/CircleOptimizer.h
```diff
@@ -47,6 +47,7 @@ class CircleOptimizer final
       FuseMulWithConv,
       FuseMulWithDiv,
       FuseTransposeWithMean,
+      ResolveCustomOp,
```
Seunghui98, line 50: The overall description of the new pass is as follows.

- The discussion of the New pass, `ResovleCustomOpPass`, came from here.
-> https://github.com/Samsung/ONE/issues/10784#issuecomment-1797862821
- This new pass is a pass that changes op, which was previously supported by `custom` op, 
when it can now be supported by `built-in` op.

seanshpark: Q) are you going to support conversion of all custom ops to built-in ops?

Seunghui98: @seanshpark 

> Q) are you going to support conversion of all custom ops to built-in ops?

No. This pass will only support specific `custom` op that currently come into the builtin op. 
Until now, only BroadcastTo op is applicable to this case. :)

seanshpark: There exist
image

Your new pass name looks like it will absorb all existing ones.

Seunghui98: @seanshpark 

Oh... Thank you for letting me know :)
Could you recommend the name of new pass ?? 😀

mhs4670go: ResolveExCustomOp..?? AdaptExCustomOp..??

hseok-oh:
@nnfw-bot test all

compiler/luci/pass/include/luci/CircleOptimizer.h
Seunghui98: @mhs4670go 
> ResolveExCustomOp

I think it is a good.
Thank you for your comments :)

seanshpark:
@Seunghui98 , can you explain how `ResolveExCustomOp` is related this this change?
What does `Ex` stand for?

Seunghui98:
@seanshpark 

> can you explain how ResolveExCustomOp is related this this change?

`ResolveExCustomOp` is a pass that convert `CircleCustom` using previous circle shema version(below latest version) to `Circle BuiltinOp` supported by the latest circle schema version.
so, **this change** includes that first identifies whether it is `CircleCustom`, checks whether it can be changed to `Builtin op` and converts it.


> What does Ex stand for?

`Ex` means `previous` or `before`.
so, `ExCustomOp` is currently supported as a builtin op based on the circle schema version, but it means that it was a custom op before.

seanshpark:
Optional) please us `[luci/pass]` prefix in the title so that to follow previous PRs

> Ex means previous or before.

My experience was about `Extended` so this confused me.
- https://stackoverflow.com/questions/20613172/what-does-suffix-ex-mean-in-function-name

Can you use `ResolvePrevCustomOpPass` not to get confused?

compiler/luci/pass/include/luci/Pass/ResolveExCustomOpPass.h
```diff
@@ -0,0 +1,40 @@
...
+#define __LUCI_RESOLVE_EX_CUSTOM_OP_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief Class to convert a custom operator to a built-in operator.
+ * pass to which ops that used to be custom op but changed to builtin ops will be applied.
```
seanshpark, line 27: 1/ First sentence ends with period, completes the sentence but the second sentence starts with `pass`, lower case, which looks like two sentence has some relation but I am not sure...
Can you please fix this?

2/ Please add the reason "from schema version upgrade" ?

compiler/luci/pass/src/ResolveExCustomOpPass.cpp
```diff
@@ -0,0 +1,84 @@
...
+    broadcastTo->name(name);
+
+    auto customOut = loco::succs(node);
+
+    assert(customOut.size() == 1);
+    replace(*customOut.begin()).with(broadcastTo);
+
+    return true;
+  }
+
```
seanshpark, line 58:
```suggestion
  }
  // TODO add more custom codes
```

compiler/luci/pass/src/ResolveExCustomOpPass.cpp
```diff
@@ -0,0 +1,84 @@
...
+namespace
+{
+
+bool resolve_custom_op(luci::CircleCustom *node)
+{
+  const std::string custom_code = node->custom_code();
+
+  if (custom_code == "BroadcastTo")
+  {
+    // check if input are support data types
```
seanshpark, line 33: optional) maybe introduce separate method for "BroadcastTo"

Seunghui98:
@seanshpark 
> Can you use ResolvePrevCustomOpPass not to get confused?
> Add: How about ResolveFormerCustomOpPass ?

I think both are good ideas. :)
but, to choose one, `ResolveFormerCustomOpPass` seems to be a pass name that helps users understand more clearly.
Thank you for your recommendation.

compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+#include "luci/Pass/ResolveFormerCustomOpPass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/IR/AttrFusedActFunc.h>
+#include <luci/Profile/CircleNodeOrigin.h>
+
+#include <flatbuffers/flexbuffers.h>
+namespace
```
seanshpark, line 24:
```suggestion
#include <flatbuffers/flexbuffers.h>

namespace
```

seanshpark:
https://github.com/Samsung/ONE/pull/12318#pullrequestreview-1796132205 isn't applied

compiler/luci/pass/include/luci/Pass/ResolveFormerCustomOpPass.h
```diff
@@ -0,0 +1,40 @@
...
+{
+
+/**
+ * @brief Class to convert a custom operator to a built-in operator.
+ * Pass to which ops that used to be custom op but changed to builtin ops will be applied.
+ */
+struct ResolveFormerCustomOpPass final : public logo::Pass
+{
+  ResolveFormerCustomOpPass() {}
+
```
mhs4670go, line 32: I think this default constructor just can be omitted if there is no specific reason to add it.

```suggestion
```

compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
```diff
@@ -0,0 +1,90 @@
...
+{
+
+bool resolve_with_BroadcastTo(luci::CircleCustom *node)
+{
```
mhs4670go, line 28: I think we need to check firstly if the number of inputs matches what it should be.
```suggestion
{
if (node->numInputs() != 2)
    return false;
```

compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
```diff
@@ -0,0 +1,90 @@
...
+  // check if input are support data types
+  auto input = loco::must_cast<luci::CircleNode *>(node->inputs(0));
+  if (input->dtype() != loco::DataType::FLOAT32)
+    return false;
```
mhs4670go, line 32: I just referred [this page](https://www.tensorflow.org/api_docs/python/tf/broadcast_to) but there seemed to not have any supported data types. 

> A Tensor. Has the same type as input.

We can just check if the data type of output is same with the one of the input feature map.

compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  auto name = node->name();
+  assert(name.length() > 0);
+
+  auto broadcastTo = node->graph()->nodes()->create<luci::CircleBroadcastTo>();
+  broadcastTo->input(input);
+  broadcastTo->shape(shape);
+  broadcastTo->name(name);
+
```
mhs4670go, line 46: We should add an origin about this new op.
```cxx
luci::add_origin(broadcastTo, luci::get_origin(node));
```

Seunghui98: @mhs4670go 

Thank you a lot !! :)
Could you explain what function `luci::add_origin()` has?

compiler/luci/pass/src/ResolveFormerCustomOpPass.test.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "luci/Pass/ResolveFormerCustomOpPass.h"
+
+#include <gtest/gtest.h>
+
+TEST(ResolveFormerCustomOpPassTest, name)
+{
+  luci::ResolveFormerCustomOpPass pass;
+  auto const name = pass.name();
+  ASSERT_NE(nullptr, name);
+}
```
seanshpark, line 26: (in another PR)
It would be nice to have some positive and negative tests for new introduced codes.
I don't see them in the draft. Can you apply them in your draft PR?
1/ unit test codes in this `ResolveFormerCustomOpPass.test.cpp` file
2/ integration test in `luci-pass-value-py-test` module

Seunghui98: @seanshpark 
> I don't see them in the draft. Can you apply them in your draft PR?

Thank you for your guide! :)
I will apply it in Draft PR including some positive and negative tests.

seanshpark (approved):
LGTM

compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
mhs4670go: As I said in the offline, "origin" feature has been introduced for knowing where the newly created operator come from during the optimization process. 

`luci::add_origin` has two parameters - former is newly created one and the latter is the origin.

mhs4670go (approved):
LGTM

Issue #12263, opened 2023-12-11
[luci-pass/onert-micro] Convert GRU pattern to CustomGRU


BalyshevArtem:
### What

Let's add pass to convert pattern (which is formed when converting a GRU operation from tensorflow to tflite) to CustomGRU and support it in onert-micro

### Why

Some target models for onert-micro contains such pattern. This conversion accelerates the execution of target models in onert-micro.

### How

- [ ] Introduce FuseGRUpatternAsCustomGRU pass in luci/pass
- [ ] Support CustomGRU in onert-micro

BalyshevArtem:
First draft to check impact #11840 of such conversion

BalyshevArtem:
## Before conversion

Main graph:

![main_g](https://github.com/Samsung/ONE/assets/43214667/00c73ffa-9476-486a-a3cd-d4c0d58ce895)

Cond graph:

![cond_g](https://github.com/Samsung/ONE/assets/43214667/10d32350-89fe-4044-82f3-7a167b0ac74a)

Body graph:

![body_g](https://github.com/Samsung/ONE/assets/43214667/3f6d60f7-b4e3-48ca-8673-4d59292cf085)

BalyshevArtem:
## After conversion:

![opt_g](https://github.com/Samsung/ONE/assets/43214667/4284de36-1a37-43d4-ae65-3eac56a7f09b)

seanshpark:
Are you going to use `custom_gru` as the op type?
Op is `CustomOp` and custom code is `custom_gru`.  This is odd.
I would be better to rename this to some meaningful name.
What kind of GRU is this? Plz add some details about this.

seanshpark:
@lemmaa , @hseok-oh , what do you think of adding new Circle Op  `CIR_GRU = -5` in circle schema?
- `CIR_` prefix is to avoid future conflict when `GRU` is added in tflite , just optional, not the main point.

BalyshevArtem:
> Are you going to use `custom_gru` as the op type? Op is `CustomOp` and custom code is `custom_gru`. This is odd. I would be better to rename this to some meaningful name. What kind of GRU is this? Plz add some details about this.

Yes, use `CustomOp` with custom code - `CUSTOM_GRU`.


> I would be better to rename this to some meaningful name

Smth like `CIRCLE_CUSTOM_GRU`?



> What kind of GRU is this?

[...]

PR #12319, opened 2023-12-18
[DRAFT] Support CircleGRU

BalyshevArtem:
This draft supports CircleGRU in compiler.

for issue: #12320
related issue: #12263


compiler/circlechef/circle/src/CircleOpRegistry.h
```diff
@@ -58,6 +58,7 @@ class CircleOpRegistry
     REG_TFL_OP(BATCH_MATMUL, CircleOpBatchMatMul);
     REG_TFL_OP(BCQ_FULLY_CONNECTED, CircleOpBCQFullyConnected);
     REG_TFL_OP(BCQ_GATHER, CircleOpBCQGather);
+    REG_TFL_OP(CIR_GRU, CircleOpCircleGRU);
```
seanshpark, line 61:
```suggestion
    REG_TFL_OP(GRU, CircleOpGRU);
```

seanshpark: or,
1/ to prepare tflite introduces GRU,
2/ two `Circle` looks odd
```suggestion
    REG_TFL_OP(CIR_GRU, CircleOpCirGRU);
```

compiler/circlechef/proto/circlechef.proto
```diff
@@ -64,6 +64,7 @@ enum Activation {
   NONE = 0;
   RELU = 1;
   RELU6 = 3;
+  TANH = 4;
```
seanshpark, line 67: Q) is this necessary for our model?

compiler/luci/pass/src/FuseGRUPass.cpp
```diff
@@ -0,0 +1,477 @@
...
+          num_split_out++;
+          break;
+        default:
+          return false;
+      }
+    }
+
+    if (num_fc != 2 or num_mul != 3 or num_logistic != 2 or num_split != 2 or num_add != 6 or
+        num_gather != 1 or num_reshape != 1 or num_sub != 1 or num_tanh != 1 or num_split_out != 6)
+      return false;
```
seanshpark, line 241: I don't think counting elements can provide correct condition.
And tracking the graph for exact Op type looks laborious job.
But we may have to do with some good graph traversal codes.

seanshpark:
May need `res/TensorFlowLiteRecipes/Net_DecomposedGru_000` for testings
- circle2circle-dredd-recipe-test
- luci-pass-value-py-test

You may know that you can create `Net_DecomposedGru_000` from tflite file with `tflchef-reverse` tool.

seanshpark:
Q) do you not have a plan to implement luci-interpreter ?

compiler/luci/pass/src/FuseGRUPass.cpp
```diff
@@ -0,0 +1,477 @@
...
+  auto cloned = graph->nodes()->create<luci::CircleConst>();
+
+  if (cloned != nullptr)
+  {
+    // dtype/shape
+    cloned->dtype(node->dtype());
+    cloned->rank(node->rank());
+
+    // values
+    switch (node->dtype())
```
seanshpark, line 322: Q) do we need all these implemented types?

compiler/luci/pass/src/FuseGRUPass.cpp
```diff
@@ -0,0 +1,477 @@
...
+{
+  bool changed = false;
+
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    auto while_out_node = dynamic_cast<luci::CircleWhileOut *>(node);
+    if (not while_out_node)
+      continue;
+
+    if (fuse_gru(while_out_node))
```
seanshpark, line 470: Q) are other sub graphs like the body graph removed?
I don't recall we had some method to remove graph in the Module.

compiler/circlechef/circle/src/CircleOpRegistry.h
BalyshevArtem: I think second variant better to avoid future problems with tflite GRU operation

BalyshevArtem:
> Q) do you not have a plan to implement luci-interpreter ?

I'm not sure if this operation is needed in luci-interpreter, so now I don't have this in plan. But if you think it's better to add it, I'll do it)

seanshpark:
> But if you think it's better to add it, I'll do it)

Current purpose I have is to validate fusion implementation through value test (as mathematical equivalence). 
If adding the kernel takes heavy work, we may fire a task issue and postpone to later time.
I'll let you decide by your situation.

compiler/luci/pass/src/FuseGRUPass.cpp
BalyshevArtem: > Q) are other sub graphs like the body graph removed?

Actually - no. We need to support some methods to remove subgraphs in the Module

BalyshevArtem: What do you think about add a new option that will leave only the main sub-graph, and delete others? Smth like `--use_only_main_graph` and use fuse_gru and use_only_main_graph together. Otherwise, we need to transfer/save the luci::Module entity to CircleOptimizer to remove unnecessary graphs

compiler/luci/pass/src/FuseGRUPass.cpp
BalyshevArtem: Okay, I will change it)

compiler/circlechef/circle/src/CircleOpRegistry.h
BalyshevArtem: Done

compiler/circlechef/proto/circlechef.proto
BalyshevArtem: Remove it, thank you

seanshpark:
>  `tflite_import.num_subgraph() == 1' failed.

current tflchef-reverse seems to supports only single subgraph model.
you can expand this to support multiple subgraphs or post an issue to request this feature.

BalyshevArtem:
> May need `res/TensorFlowLiteRecipes/Net_DecomposedGru_000` for testings

Done, added it

BalyshevArtem:
@seanshpark, can I split this draft into PRs? For `luci/pass/src/FuseGRUPass` pass I think better create another draft to solve this problem https://github.com/Samsung/ONE/pull/12319#discussion_r1453249304 and add luci-interpreter part to check values.

compiler/luci/pass/src/FuseGRUPass.cpp
seanshpark: > Otherwise, we need to transfer/save the luci::Module entity to CircleOptimizer to remove unnecessary graphs

Maybe some kind of eliminate dead sub-graph...
1/ As of now, all sub-graphs having ID > 0 are used by some other sub-graph (usually ID == 0).
(I'm not sure this will hold in the future)
2/ We can use `void CircleOptimizer::optimize(luci::Module *m) const`
3/ Add something like `EliminateDeadSubgraphPass` (maybe with `--elminate-dead-graph` option or run this for default without any options)
4/ Traverse main graph and mark sub-graphs that are used
5/ Delete sub-graphs that are not marked

Can this be feasible ?

seanshpark:
> can I split this draft into PRs? 

Yes, let's go :)

compiler/luci/pass/src/FuseGRUPass.cpp
BalyshevArtem: > Can this be feasible ?

I think yes, good idea to add `EliminateDeadSubgraphPass` and run this for default

PR #12322, opened 2023-12-19
[infra/nncc] Revise Makefile.arm32

seanshpark:
This will revise Makefile.arm32 to replace with luci-value-py-test.


seanshpark:
for https://github.com/Samsung/ONE/issues/12071#issuecomment-1841648762

mhs4670go (approved):
LGTM

PR #12323, opened 2023-12-19
[onert/api] Store circle on nnfw_train_export_circle

glistening:
It creates new circle using trained weights on nnfw_train_export_circle.

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -29,6 +29,7 @@
 #include "ir/OpCode.h"
 #include "util/TracingCtx.h"
 #include "odc/QuantizeManager.h"
+#include "circle_schema_generated.h"
 
 #include <fstream>
 #include <iostream>
@@ -1500,6 +1501,8 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
       return _buf != MAP_FAILED;
     }
 
+    bool sync() { return msync(_buf, _buf_sz, MS_SYNC) == 0; }
+
     bool close()
     {
       bool ret = false;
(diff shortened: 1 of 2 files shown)


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1531,7 +1534,47 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
...
+      auto buf_idx = subg->tensors()->Get(idx.value())->buffer();
+      const ::circle::Buffer *buffer = (*model->buffers())[buf_idx];
+      if (!buffer || !buffer->data())
+        throw std::runtime_error("Buffer for trainable tensors is invalid");
+
+      const flatbuffers::Vector<uint8_t> *array = buffer->data();
+      if (!array)
+        throw std::runtime_error("Data for trainable tensor's buffer is invalid");
+
+      uint8_t *org_buf = const_cast<uint8_t *>(array->Data());
```
jyoungyun, line 1559: I suggest to use const data type here and use `const_cast` in `memcpy` function directly so that the pointer address does not change in the middle.

glistening: @jyoungyun I don't understand what you're concerning by _the pointer address does not change in the middle_. Also, as you can see a few lines below, I would like to check whether `org_buf` is null or not before passing it to `memcpy`.

Anyway, I applied my understanding of your suggestion.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1531,7 +1534,46 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
...
+      auto buf_idx = subg->tensors()->Get(idx.value())->buffer();
+      const ::circle::Buffer *buffer = (*model->buffers())[buf_idx];
+      if (!buffer || !buffer->data())
+        throw std::runtime_error("Buffer for trainable tensors is invalid");
+
+      const flatbuffers::Vector<uint8_t> *array = buffer->data();
+      if (!array)
+        throw std::runtime_error("Data for trainable tensor's buffer is invalid");
+
+      if (!array->Data())
```
glistening, line 1559:
```suggestion
      const uint8_t *org_buf = array->Data();
      if (!org_buf)
```

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1531,7 +1534,46 @@ NNFW_STATUS nnfw_session::train_export_circle(const char *path)
...
+        throw std::runtime_error("Data for trainable tensor's buffer is invalid");
+
+      auto org_buf_sz = array->size();
+      if (org_buf_sz != tensor->total_size())
+        throw std::runtime_error("Trained tensor buffer size does not match original tensor's one");
+
+      memcpy(const_cast<uint8_t *>(array->Data()), tensor->buffer(), org_buf_sz);
```
glistening, line 1566:
```suggestion
      memcpy(const_cast<uint8_t *>(org_buf), tensor->buffer(), org_buf_sz);
```

Or you may want this.

jyoungyun: Yes, right. I think this change is better than before. :)

glistening:
@jyoungyun I've updated as you suggested.
@hseok-oh or/and @Samsung/one_onert PTAL

glistening:
For reviewers,

- Train and Export
```
$ Product/x86_64-linux.debug/out/bin/onert_train \
mnist.circle \
--load_input:raw mnist.input \
--load_expected:raw mnist.expected \
--epoch 1 \
--export_path tr.circle
```
- Dump tensor information with const buffers
```
$ build/compiler/circle-tensordump/circle-tensordump --tensors tr.circle | egrep "buffer|shape" -B 1 -A 3 > after.txt

$ build/compiler/circle-tensordump/circle-tensordump --tensors mnist.circle | egrep "buffer|shape" -B 1 -A 3 > before.txt
```
- Find differences:
```
$ diff before.txt after.txt
...
41c41
<      └── data  : 0.182483, 0.171852, 0.158546, ...
---
>      └── data  : 0.18249, 0.171859, 0.158555, ...
...
```

hseok-oh (approved):

hseok-oh:
@nnfw-bot test all

PR #12339, opened 2023-12-20
[tools/onert_train] Apply validation split value

jyoungyun:
This commit applies the validation split value.
If the user sets the validataion split value, the data geneator splits the given dataset into train and validation datasets. Then, the training process uses the train dataset, and the validation process uses the validation dataset.

Draft: #12265


ragmani (approved):
LGTM

glistening (approved):
LGTM

PR #12340, opened 2023-12-20
[onert] Introduce ir::train::TrainingInfo

zetwhite:
This PR introduces ir::train::TrainingInfo into core.

draft : https://github.com/Samsung/ONE/pull/12152

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -25,8 +25,10 @@
 #include "tflite_loader.h"
 #include "trix_loader.h"
 #include "json/json.h"
+#include "ir/Model.h"
 #include "ir/NNPkg.h"
 #include "ir/OpCode.h"
+#include "ir/train/TrainingInfo.h"
 #include "util/TracingCtx.h"
 #include "odc/QuantizeManager.h"
 #include "circle_schema_generated.h"
@@ -226,7 +228,7 @@ uint64_t getBufSize(const nnfw_tensorinfo *info)
 
 nnfw_session::nnfw_session()
   : _nnpkg{nullptr}, _coptions{}, _compiler_artifact{nullptr}, _execution{nullptr},
-    _kernel_registry{nullptr}, _quant_manager{nullptr}
+    _kernel_registry{nullptr}, _train_info{nullptr}, _quant_manager{nullptr}
 {
   // DO NOTHING
 }
@@ -275,6 +277,8 @@ NNFW_STATUS nnfw_session::load_circle_from_buffer(uint8_t *buffer, size_t size)
     // TODO: Update _model_path if necessary
     _nnpkg = std::make_shared<onert::ir::NNPkg>(std::move(model));
     _coptions.push_back(onert::compiler::CompilerOptions::fromGlobalConfig());
+    // TODO load TrainingInfo from model, using TraininfoLoader
+    _train_info = onert::ir::train::TrainingInfo::createDefaultInfo();
     _state = State::MODEL_LOADED;
   }
   catch (const std::exception &e)
@@ -316,6 +320,8 @@ NNFW_STATUS nnfw_session::load_model_from_modelfile(const char *model_file_path)
     _model_path = std::string(model_file_path);
     _nnpkg = std::make_shared<onert::ir::NNPkg>(std::move(model));
     _coptions.push_back(onert::compiler::CompilerOptions::fromGlobalConfig());
+    // TODO load TrainingInfo from model, using TraininfoLoader
+    _train_info = onert::ir::train::TrainingInfo::createDefaultInfo();
     _state = State::MODEL_LOADED;
   }
   catch (const std::exception &e)
@@ -400,6 +406,8 @@ NNFW_STATUS nnfw_session::load_model_from_nnpackage(const char *package_dir)
       model->bindKernelBuilder(_kernel_registry->getBuilder());
       _nnpkg->push(onert::ir::ModelIndex{i}, std::move(model));
       _coptions.push_back(onert::compiler::CompilerOptions::fromGlobalConfig());
+      // TODO load TrainingInfo from model, using TraininfoLoader
+      _train_info = onert::ir::train::TrainingInfo::createDefaultInfo();
     }
 
     auto toIODesc = [](std::string str) {
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -25,10 +25,8 @@
 #include "tflite_loader.h"
 #include "trix_loader.h"
 #include "json/json.h"
-#include "ir/Model.h"
 #include "ir/NNPkg.h"
 #include "ir/OpCode.h"
-#include "ir/train/TrainingInfo.h"
 #include "util/TracingCtx.h"
 #include "odc/QuantizeManager.h"
 #include "circle_schema_generated.h"
@@ -228,7 +226,7 @@ uint64_t getBufSize(const nnfw_tensorinfo *info)
 
 nnfw_session::nnfw_session()
   : _nnpkg{nullptr}, _coptions{}, _compiler_artifact{nullptr}, _execution{nullptr},
-    _kernel_registry{nullptr}, _train_info{nullptr}, _quant_manager{nullptr}
+    _kernel_registry{nullptr}, _quant_manager{nullptr}
 {
   // DO NOTHING
 }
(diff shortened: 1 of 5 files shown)


jyoungyun:
(suggestion)
How about separating this PR into two PRs, one is to define `ir::train::TrainingInfo` and the other is to add  `TrainingInfo` to `nnfw_session`. And is changing `compile::train::TrainingInfo` to `ir::train::TrainingInfo` a big change? If you don't mind, I suggest updating the codes related to `compile::train::TrainingInfo`  at once in one pr.

runtime/onert/core/src/ir/train/TrainingInfo.cc
```diff
@@ -0,0 +1,72 @@
...
+    loss_info.reduction_type = ir::train::LossReductionType::SumOverBatchSize;
+  }
+  info->setLossInfo(loss_info);
+
+  onert::ir::train::OptimizerInfo opt_info;
+  {
+    opt_info.optim_code = ir::train::OptimizerCode::SGD;
+    opt_info.learning_rate = 0.001f;
+  }
+  info->setOptimizerInfo(opt_info);
```
jyoungyun, line 44: Should we initialize these variables? I think there is no need to redefine it here because there is already an initial value.

runtime/onert/core/include/ir/train/TrainingInfo.h
```diff
@@ -0,0 +1,68 @@
...
+{
+namespace train
+{
+
+class TrainingInfo final
+{
+public:
+  TrainingInfo() : _loss_info(), _optimizer_info(), _batch_size(0), _epoch(0) {}
+  // named constructor : create a default valid TrainingInfo
+  static std::unique_ptr<TrainingInfo> createDefaultInfo();
```
zetwhite, line 39: Q: Why did you use the named constructor?  

A: I think it is confusing that some `ir` default-constructors make an invalid one, but sometimes not.  
So, I prefer to give a named constructor, to show the caller that this `ir` is filled with default 'valid' value.

runtime/onert/core/include/ir/train/TrainingInfo.h
```diff
@@ -0,0 +1,68 @@
...
+  uint32_t batchSize() const { return _batch_size; }
+  uint32_t epoch() const { return _epoch; }
+
+  // setter
+  void setBatchSize(const uint32_t batch_size) { _batch_size = batch_size; }
+  void setLossInfo(const LossInfo &loss_info) { _loss_info = loss_info; }
+  void setOptimizerInfo(const OptimizerInfo &optimizer_info) { _optimizer_info = optimizer_info; }
+  void setEpoch(uint32_t epoch) { _epoch = epoch; }
+
+  bool isValid() const;
```
zetwhite, line 55: I'd like to check the TrainingInfo validation, using this method in `train_prepare`. 
in draft, `isValid()` is used here - https://github.com/Samsung/ONE/pull/12152/files#r1432281266 

We can check the "combination of each field" is fine with this method.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -400,6 +406,8 @@ NNFW_STATUS nnfw_session::load_model_from_nnpackage(const char *package_dir)
       model->bindKernelBuilder(_kernel_registry->getBuilder());
       _nnpkg->push(onert::ir::ModelIndex{i}, std::move(model));
       _coptions.push_back(onert::compiler::CompilerOptions::fromGlobalConfig());
+      // TODO load TrainingInfo from model, using TraininfoLoader
+      _train_info = onert::ir::train::TrainingInfo::createDefaultInfo();
```
zetwhite, line 410: I have a plan about TODO comment like this. 
A draft code related with this TODO - https://github.com/Samsung/ONE/pull/12152/files#r1426184845

zetwhite:
> How about separating this PR into two PRs, one is to define `ir::train::TrainingInfo` and the other is to add `TrainingInfo` to `nnfw_session`. 

Thank you for the suggestion! No problem. I'll separate it into two :) 

> And is changing `compile::train::TrainingInfo` to `ir::train::TrainingInfo` a big change? If you don't mind, I suggest updating the codes related to `compile::train::TrainingInfo` at once in one pr.

Originally, I'd like to complete loader-things first. and replace compile::TrainingInfo with `ir` later.
Because 
   - replacing is not really necessary for the loader.
   - `compile::train::TrainingInfo` is used in Compiler, Executor and TrainableOperationConverter,
      Since I'm not familiar with these, I'd like to replace it carefully. 

So, It is fine, I'd like to replace `compiler::TraininnInfo` with `ir::TrainingInfo` later. 
But if you think it is better to do now, I'd like to follow. So, please share your opinion. Thank you :

runtime/onert/core/src/ir/train/TrainingInfo.cc
zetwhite: I append my idea about the named constructor here - https://github.com/Samsung/ONE/pull/12340/files/21d8ef90f0277b3f9df5f822d26cfd8adf56372a#r1432279409. 

The initial constructor has initially "invalid" value by default. 
The named constructor returns a default "valid" info, to be used when training_info isn't given by a model or user.

runtime/onert/core/include/ir/train/TrainingInfo.h
```diff
@@ -0,0 +1,68 @@
...
+#include "ir/train/LossInfo.h"
+
+namespace onert
+{
+namespace ir
+{
+namespace train
+{
+
+class TrainingInfo final
```
zetwhite, line 34: (plan) This TrainingInfo will be saved in nnfw_session : https://github.com/Samsung/ONE/pull/12342/commits/8fa3ffaec360916b60eba426e533da61952fa9e0

jyoungyun:
> Thank you for the suggestion! No problem. I'll separate it into two :) 

Thank you :)

> Originally, I planned to complete loader-things first. and replace `compile::TrainingInfo` with `ir` later.
> Because 
>    - replacing is not really necessary for the loader.
>    - `compile::train::TrainingInfo` is used in Compiler, Executor and TrainableOperationConverter,
>       Since I'm not familiar with these, I'd like to replace it carefully. 
> 
> So, It is fine, I'd like to replace `compiler::TraininnInfo` with `ir::TrainingInfo` later. 

I respect your opinion. It may take more time to replace `compiler::TrainingInfo` with `ir::TrainingInfo`. So we continue this change as it is and apply for replacement later. :)

runtime/onert/core/src/ir/train/TrainingInfo.cc
zetwhite: I removed the default valid training info after an offline discussion.

runtime/onert/core/include/ir/train/TrainingInfo.h
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __ONERT_IR_TRAIN_TRAINING_INFO_H__
+#define __ONERT_IR_TRAIN_TRAINING_INFO_H__
+
+#include <memory>
```
jyoungyun, line 21:
```suggestion
```

jyoungyun (approved):
LGTM Thank you!

zetwhite:
> LGTM Thank you!

updated once more, because of code formatting issue :sweat_smile:

glistening (approved):
LGTM

runtime/onert/core/src/ir/train/TrainingInfo.cc
```diff
@@ -0,0 +1,49 @@
...
+
+bool TrainingInfo::isValid() const
+{
+  if (_batch_size == 0)
+    return false;
+
+  if (_optimizer_info.optim_code == onert::ir::train::OptimizerCode::Invalid)
+    return false;
+
+  if (_optimizer_info.learning_rate == 0.0f)
```
ragmani, line 34: If learning rate is a negative value, outputs will move away from the minimum. In other words, outputs will move to reverse what gradient descent does, and nn model won't be trained.
So, it would be better to change the condition if this function is for checking valid info of `TrainingInfo`, not default values.
```suggestion
  if (_optimizer_info.learning_rate <= 0.0f)
```

ragmani (approved):

jyoungyun (approved):
LGTM

runtime/onert/core/src/ir/train/TrainingInfo.cc
glistening: @zetwhite or @ragmani Please update this in another PR.

PR #12360, opened 2023-12-21
[onert] Update TrainingInfo::isValid() condition

zetwhite:
This PR update TrainingInfo::isValid() condtion to check learning_rate has positive value.

### DIFF
--- a/runtime/onert/core/src/ir/train/TrainingInfo.cc
+++ b/runtime/onert/core/src/ir/train/TrainingInfo.cc
@@ -31,7 +31,7 @@ bool TrainingInfo::isValid() const
   if (_optimizer_info.optim_code == onert::ir::train::OptimizerCode::Invalid)
     return false;
 
-  if (_optimizer_info.learning_rate == 0.0f)
+  if (_optimizer_info.learning_rate <= 0.0f)
     return false;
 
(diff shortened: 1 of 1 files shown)


runtime/onert/core/src/ir/train/TrainingInfo.cc
```diff
@@ -31,7 +31,7 @@ bool TrainingInfo::isValid() const
   if (_optimizer_info.optim_code == onert::ir::train::OptimizerCode::Invalid)
     return false;
 
-  if (_optimizer_info.learning_rate == 0.0f)
+  if (_optimizer_info.learning_rate <= 0.0f)
```
zetwhite, line 34: background : https://github.com/Samsung/ONE/pull/12340#discussion_r1432642877 

> (Optional)
> If learning rate is a negative value, outputs will move away from the minimum after training. In other words, outputs will move to reverse what gradient descent does, and nn model won't be trained at all.
So, it would be better to change the condition if this function is for checking valid info of `TrainingInfo`, not default values.

SlavikMIPT (approved):
LGTM

ragmani (approved):
LGTM

glistening (approved):
LGTM

PR #12362, opened 2023-12-26
[onert] Introduce AUTO loss reduction type

jyoungyun:
This commit introduces AUTO loss reduction type. The experimental API does not have Invalid loss reduction type because the user does not have to select the INVALID type. Instead, INVALID type is changed to AUTO type. AUTO type means that the reduction option will be determined by the context. In contrast, onert ir supports both Invalid and Auto types. The default value of loss info in onert ir is Invalid type to determine whether the value has been received from the user.

### DIFF
--- a/runtime/onert/api/include/nnfw_experimental.h
+++ b/runtime/onert/api/include/nnfw_experimental.h
@@ -182,8 +182,8 @@ typedef enum
 
 typedef enum
 {
-  /** Invalid */
-  NNFW_TRAIN_LOSS_REDUCTION_INVALID = 0,
+  /** Auto */
+  NNFW_TRAIN_LOSS_REDUCTION_AUTO = 0,
   /** Scalar sum divided by number of elements in losses */
   NNFW_TRAIN_LOSS_REDUCTION_SUM_OVER_BATCH_SIZE = 1,
   /** Scalar sum of weighted losses */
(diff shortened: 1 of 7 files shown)


runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -182,8 +182,8 @@ typedef enum
 
 typedef enum
 {
-  /** Invalid */
-  NNFW_TRAIN_LOSS_REDUCTION_INVALID = 0,
+  /** Auto */
+  NNFW_TRAIN_LOSS_REDUCTION_AUTO = 0,
```
zetwhite, line 186: How about keep both `invalid` and `auto` here? 

Since a api for getting train_info - `nnfw_train_get_traininfo` will be added, There `invalid` is also necessary in API space  to represent(`LossReductionType::Auto` in core).

zetwhite: Otherwise, looks good to me. 👍

jyoungyun: How about adding `Invalid` enum to both `Loss` and `Optimizer` at once? :)

zetwhite: > How about adding Invalid enum to both Loss and Optimizer at the same time? 

Oh I see. No problem. :smile: 
If so, I could add `invalid` in same time for both `core/ir/..` and `api/../nnfw_experimental`, when it become necessary.

zetwhite (approved):
LGTM :+1:

ragmani (approved):
LGTM

glistening:
Typo: introdues → introduces in commit body

runtime/onert/core/include/compiler/train/TrainingInfo.h
```diff
@@ -47,8 +47,9 @@ class TrainingInfo
   {
     _loss_info = loss_info;
 
-    // If the reduction type is not specified, it uses SumOverBatchSize.
-    if (_loss_info.reduction_type == ir::train::LossReductionType::Invalid)
+    // Always use SumOverBatchSize type
+    // TODO Support different type
+    if (_loss_info.reduction_type == ir::train::LossReductionType::Auto)
       _loss_info.reduction_type = ir::train::LossReductionType::SumOverBatchSize;
   }
```
zetwhite, line 54: (reference link for another PR)

jyoungyun:
> Typo: introdues → introduces in commit body

I fixed the commit message.

glistening (approved):
LGTM

PR #12367, opened 2023-12-26
[luci-value-py-test] Enable SelectV2

ragmani:
This commit enables SelectV2 Op value test.

ONE-DCO-1.0-Signed-off-by: ragmani ragmani0216@gmail.com

---

For #12148


mhs4670go:
Could you apply this change in the `luci-value-py-test`? `luci-value-test` module has been deprecated.

ragmani:
> Could you apply this change in the luci-value-py-test? luci-value-test module has been deprecated.

Is the same way as `luci-value-test` to enable tests?

ragmani:
> And can I see the issue that luci-value-test has been deprecated. I found the issue https://github.com/Samsung/ONE/issues/12071, but couldn't find any related comment.

I heard from @mhs4670go offline that compiler project is deprecated if `exclude.me` file exists in the project.

mhs4670go:
> Is the same way as `luci-value-test` to enable tests? 

Yeah, you can apply the change the way `luci-value-test` does. Just remove the sharp sign from the below codes.
https://github.com/Samsung/ONE/blob/be49160015d223839d4032e8edc6efa0f860f540/compiler/luci-value-py-test/test.lst#L142-L144

> And can I see the issue that `luci-value-test` has been deprecated. I found the issue #12071, but couldn't find any related comment.

https://github.com/Samsung/ONE/pull/12367#issuecomment-1869474779

mhs4670go:
You can modify the commit title, too. `luci-value-test` -> `luci-value-py-test`.

seanshpark (approved):
LGTM thank you!

mhs4670go (approved):

PR #12378, opened 2023-12-28
[luci/pass] Revise ResolveFormerCustomOpPass with condition

seanshpark:
This will revise esolveFormerCustomOpPass with condition check to early return false so that new CircleBroadcastTo object won't be wasted.

### DIFF
--- a/compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
+++ b/compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
@@ -38,6 +38,14 @@ bool resolve_with_BroadcastTo(luci::CircleCustom *node)
   if (shape->dtype() != loco::DataType::S32 && shape->dtype() != loco::DataType::S64)
     return false;
 
+  auto customOut = loco::succs(node);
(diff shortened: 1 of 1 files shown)


compiler/luci/pass/src/ResolveFormerCustomOpPass.cpp
```diff
@@ -47,15 +55,6 @@ bool resolve_with_BroadcastTo(luci::CircleCustom *node)
...
   luci::add_origin(broadcastTo, luci::get_origin(node));
 
-  auto customOut = loco::succs(node);
-
-  assert(customOut.size() == 1);
-
-  // check if the data type of output is same with the one of the input feature map.
-  auto output = loco::must_cast<luci::CircleNode *>(*customOut.begin());
-  if (input->dtype() != output->dtype())
-    return false;
```
seanshpark, line 57: when return false is called, `broadcastTo` objected is useless.
this is not critical as not used nodes will be cleaned later.

Seunghui98 (approved):
LGTM Thank you ! :)

PR #12379, opened 2023-12-28
[tools/onert_train] Add vdata_length check logic

jyoungyun:
This commit adds vdata_length check logic.
If the user does not give the validation_split value, the vdata_length
is 0 by default and it does not execute validation loop. In this case,
the error check logic is meaningless.

### DIFF
--- a/tests/tools/onert_train/src/onert_train.cc
+++ b/tests/tools/onert_train/src/onert_train.cc
@@ -255,7 +255,10 @@ int main(const int argc, char **argv)
       exit(-1);
     }
 
-    if (vdata_length < tri.batch_size)
+    // If the user does not give the validation_split value,
+    // the vdata_length is 0 by default and it does not execute
+    // validation loop.
+    if (vdata_length != 0 && vdata_length < tri.batch_size)
     {
       std::cerr << "E: validation data is not enough for validation."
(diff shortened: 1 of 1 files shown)


tests/tools/onert_train/src/onert_train.cc
```diff
@@ -255,14 +255,6 @@ int main(const int argc, char **argv)
       exit(-1);
     }
 
-    if (vdata_length < tri.batch_size)
```
YongseopKim, line 258: Q. Is the flow working as you think? 

I mean, I read below code in this file

```cpp
          const int num_valid_step = vdata_length / tri.batch_size;
          for (uint32_t n = 0; n < num_valid_step; ++n)
```

if `(vdata_len < tri.batch_size)` is true, the above code(running validation) doesn't enter the loop. I'm curious whether it's okay or not.

(IHMO, warning log is necessary)

jyoungyun: I updated this PR to add vdata_length value check logic to the if condition.

zetwhite (approved):
LGTM :+1:

YongseopKim (approved):
LGTM 👍

PR #12380, opened 2023-12-28
[onert] Implement ReduceMean train backend

jyoungyun:
This commit implements ReduceMean train backend.

Draft: #12358

### DIFF
--- a/runtime/onert/backend/train/ops/MeanLayer.cc
+++ b/runtime/onert/backend/train/ops/MeanLayer.cc
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2023 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "MeanLayer.h"
+
+#include "OperationUtils.h"
+
+#include <cker/Shape.h>
+#include <cker/train/operation/ReduceMean.h>
+#include <cker/operation/BinaryArithmeticOps.h>
+
+namespace onert
+{
+namespace backend
+{
+namespace train
+{
+namespace ops
+{
+
+MeanLayer::MeanLayer()
+  : cpu::ops::MeanLayer(), _back_prop_input{nullptr}, _back_prop_output{nullptr}
+{
+  // DO NOTHING
+}
+
+void MeanLayer::configure(const IPortableTensor *input, const IPortableTensor *axes,
+                          IPortableTensor *output, bool keep_dims, IPortableTensor *back_prop_input,
+                          const IPortableTensor *back_prop_output)
+{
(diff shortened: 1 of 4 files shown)


runtime/onert/backend/train/ops/MeanLayer.cc
```diff
@@ -0,0 +1,85 @@
...
+  nnfw::cker::Shape keep_dim_shape;
+  if (_keep_dims == false)
+  {
+    keep_dim_shape.ReplaceWith(getShape(_input));
+    assert(getShape(_axes).DimensionsCount() == 1);
+    for (int i = 0; i < getShape(_axes).Dims(0); ++i)
+    {
+      keep_dim_shape.SetDim(reinterpret_cast<int32_t *>(_axes->buffer())[i], 1);
+    }
+  }
```
zetwhite, line 63: I'm not sure about this part. 

I thought axes is just an one integer value (axis = 0, axis = 1 ..).

But It looks like axes is represetend in tensor in onert. :thinking:

jyoungyun: Yes, the axes is a tensor in onert. And AFAIK, the `axes` tensor is one dimension tensor. It contains multiple axes to reduce.

This part is not necessary in the formula, but it is necessary to use Broadcast in eigen. If the `keep_dims` is false, this operation reduces the given axis dimension. So if the input shape of X is (2,2) and the axis is 0, the output shape of reduce mean is (2,). In eigen, the (2,) shape can not broadcast to the (2,2). So it changes the (2,) shape to the (1, 2) shape. This part changes the `_back_prop_input` shape to a shape that broadcast is possible.

zetwhite: > Yes, the axes is a tensor in onert. And AFAIK, the axes tensor is one dimension tensor. It contains multiple axes to reduce. 

Aha, It contains **multiple** axes to reduce. This point it what I missed. 

Thanks for the kind explanation! I clearly understood :+1:

zetwhite (approved):
LGTM :+1:

runtime/onert/backend/train/ops/MeanLayer.cc
```diff
@@ -0,0 +1,85 @@
...
+  else
+  {
+    keep_dim_shape.ReplaceWith(getShape(_back_prop_input));
+  }
```
ragmani, line 67: (Optional)
You can use `getReducerAxes()` in `OperationUtils.h`.

runtime/onert/backend/train/ops/MeanLayer.cc
```diff
@@ -0,0 +1,85 @@
...
+
+void MeanLayer::forward(bool) { cpu::ops::MeanLayer::run(); }
+
+void MeanLayer::backward()
+{
+  nnfw::cker::Shape keep_dim_shape;
+  if (_keep_dims == false)
+  {
+    keep_dim_shape.ReplaceWith(getShape(_input));
+    assert(getShape(_axes).DimensionsCount() == 1);
```
ragmani, line 58: We may consider the case that rank is 0.
```suggestion
    assert(_getShape(_axes).DimensionsCount() == 0 || getShape(_axes).DimensionsCount() == 1);
```
or
```suggestion
    assert(_axes->getShape().rank() == 0 || _axes->getShape().rank() == 1);
```

runtime/onert/backend/train/ops/MeanLayer.cc
jyoungyun: I applied `getReducerAxes()`. PTAL again.

zetwhite (approved):

ragmani (approved):
LGTM

jyoungyun:
@Samsung/one_onert PTAL

PR #12381, opened 2023-12-28
[onert/api] Move model null check from export_circle

glistening:
It moves model null check before dereferencing model.


jyoungyun (approved):
LGTM

zetwhite (approved):
LGTM

PR #12383, opened 2023-12-28
[onert] Remove training info from nnfw_train_prepare

zetwhite:
This PR removes training info from 'nnfw_train_prepare' argument.
It also makes 'nnfw_train_prepare' to use training info from session.

related : https://github.com/Samsung/ONE/issues/11692

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -1242,7 +1242,7 @@ NNFW_STATUS nnfw_session::train_set_traininfo(const nnfw_train_info *info)
   return NNFW_STATUS_NO_ERROR;
 }
 
-NNFW_STATUS nnfw_session::train_prepare(const nnfw_train_info *info)
+NNFW_STATUS nnfw_session::train_prepare()
 {
   // We may need different state to represent training model is loaded
   if (!isStateModelLoaded())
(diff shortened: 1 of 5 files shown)


zetwhite:
A bit hard to review. 
Better to split it smaller PR

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1187,56 +1260,16 @@ NNFW_STATUS nnfw_session::train_prepare(const nnfw_train_info *info)
...
-    onert::ir::train::TrainingInfo training_info;
-    training_info.setBatchSize(tinfo.batch_size);
-    training_info.setLossInfo(loss_info);
-    training_info.setOptimizerInfo(opt_info);
+    if (not _train_info->isValid())
+      throw std::runtime_error{"training info is not valid"};
 
     auto compiler =
-      onert::compiler::CompilerFactory::get().create(_nnpkg, _coptions, &training_info);
+      onert::compiler::CompilerFactory::get().create(_nnpkg, _coptions, _train_info.get());
```
zetwhite, line 1272: (note) nnfw_train_prepare update

jyoungyun:
The `nnfw_session` in nnfw_api_internal.h contains TrainingInfo as an unique_ptr object.
https://github.com/Samsung/ONE/blob/64b87a6618e3883e5ff42023de91a3cd7c05cf5d/runtime/onert/api/src/nnfw_api_internal.h#L214
This PR obtains the address of `_train_info` and passes it to `TrainingCompiler`.
```
onert::compiler::CompilerFactory::get().create(_nnpkg, _coptions, _train_info.get());
```
The `TrainingCompiler` stores a copy of `_train_info`.
https://github.com/Samsung/ONE/blob/64b87a6618e3883e5ff42023de91a3cd7c05cf5d/runtime/onert/core/src/compiler/train/TrainingCompiler.cc#L50-L51
The `TrainingCompiler` creates a `TrainableExecutor` using `createTrainableExecutor`.
https://github.com/Samsung/ONE/blob/64b87a6618e3883e5ff42023de91a3cd7c05cf5d/runtime/onert/core/src/compiler/train/TrainingCompiler.cc#L288-L289
The `createTrainableExecutor` copies optimizer and loss information from `_training_info`.
And it is no longer used.

At first, I thought that you should use `std::move` instead of `get()` api. However, while the session is alive, the `_train_info` pointer is alive and other functions use the value of this pointer. So I think it's okay.

In addition, `nnfw_session` has `_training_step` to manage epoch count. The `TrainingInfo` also contains epoch information. So the `_training_step` can be merged into the `TrainingInfo` and `std::move` should not be performed because this epoch count must be maintained while the session is alive.

I leave a long comment, but what I want to say is "LGTM". Thank you :)

glistening:
@zetwhite Please split `api` and `tools` into two PRs.

jyoungyun (approved):

glistening (approved):
LGTM. It breaks API compatibility. Let's merge at once.

jyoungyun:
> Please split this into two PRs: (`api` and `tools`)

If split this into two PRs, it makes build error.

zetwhite:
@jyoungyun 

Thank you a lot for checking detailed things :) 

>  However, while the session is alive, the _train_info pointer is alive and other functions use the value of this pointer. 

I agree. Also `nnfw_train_get_traininfo`(not yet merged in master branch, but planned) needs to access  _train_info in session. 

> The TrainingCompiler stores a copy of _train_info. 

I think this copy is redundant. 
Maybe, we could remove this by updating TrainingCompiler holds raw pointer, instead of structure itself.
(similar to _options under code) 
https://github.com/Samsung/ONE/blob/64b87a6618e3883e5ff42023de91a3cd7c05cf5d/runtime/onert/core/src/compiler/train/TrainingCompiler.h#L72-L75

jyoungyun:
> I think this copy is redundant.
> Maybe, we could remove this by updating TrainingCompiler holds raw pointer, instead of structure itself.

I agree this point :)
And I'm making a PR for merging the `_training_step` into `TrainingInfo`. If this commit is merged, the TrainingInfo will be more useful in various ways.

zetwhite:
> In addition, nnfw_session has _training_step to manage epoch count.  

Let me add some background about this.

I added epoch to TrainingInfo, Because circle+ schema(training info flatbuffer schema) holds epoch information. (sorry for inner-link [:link:](https://github.sec.samsung.net/one-project/circle-x/blob/91fc9550d914a3c62bf41e62e007e1a2f5f37b59/spec/circle_training.fbs#L113))
 
Since there is no way to save a model file's info except saving in session._train_info, 
I put the model file's epoch in the session, just in case. 
But there is no exact plan to use epoch in session.

jyoungyun:
The epoch in circle+ schema is not necessary information. Because the epoch count is not related to the training model. We should remove the epoch info from circle+ schema. :)

PR #12384, opened 2023-12-28
[onert/api] Add nnfw_train_set_traininfo()

zetwhite:
This PR adds `nnfw_train_set_traininfo()` API into nnfw_experimental header. 
The `nnfw_train_set_traininfo()` API is to set train_info in the session.

ONE-DCO-1.0-Signed-off-by: SeungHui Youn sseung.youn@samsung.com

### DIFF
--- a/runtime/onert/api/include/nnfw_experimental.h
+++ b/runtime/onert/api/include/nnfw_experimental.h
@@ -221,7 +221,7 @@ typedef struct nnfw_train_info
 } nnfw_train_info;
 
 /**
- * @brief Set train_info into session
+ * @brief Set training info into session
  * @note  This function should be called after calling {@link nnfw_load_model_from_file}
  *        and before calling {@link nnfw_train_prepare}
  *
--- a/runtime/onert/api/include/nnfw_experimental.h
+++ b/runtime/onert/api/include/nnfw_experimental.h
@@ -222,8 +222,8 @@ typedef struct nnfw_train_info
 
 /**
  * @brief Set train_info into session
- * @note  This function should be called in MODEL_LOADED state,
- *        after {@link nnfw_load_model_from_file} before {@link nnfw_prepare}
+ * @note  This function should be called after calling {@link nnfw_load_model_from_file}
+ *        and before calling {@link nnfw_train_prepare}
  *
  *  @return @c NNFW_STATUS_NO_ERROR If successful
  */
--- a/runtime/onert/api/include/nnfw_experimental.h
+++ b/runtime/onert/api/include/nnfw_experimental.h
@@ -221,6 +221,15 @@ typedef struct nnfw_train_info
 } nnfw_train_info;
 
 /**
+ * @brief Set train_info into session
+ * @note  This function should be called in MODEL_LOADED state,
+ *        after {@link nnfw_load_model_from_file} before {@link nnfw_prepare}
+ *
+ *  @return @c NNFW_STATUS_NO_ERROR If successful
+ */
+NNFW_STATUS nnfw_train_set_traininfo(nnfw_session *session, const nnfw_train_info *train_info);
+
+/**
  * @brief Prepare session to be ready for training
  * @note  The session will be entered into training mode
  *
--- a/runtime/onert/api/include/nnfw_experimental.h
+++ b/runtime/onert/api/include/nnfw_experimental.h
@@ -221,7 +221,7 @@ typedef struct nnfw_train_info
 } nnfw_train_info;
 
 /**
- * @brief Set training info into session
+ * @brief Set training info
  * @note  This function should be called after calling {@link nnfw_load_model_from_file}
  *        and before calling {@link nnfw_train_prepare}
  *
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -1178,8 +1178,6 @@ NNFW_STATUS nnfw_session::train_set_traininfo(const nnfw_train_info *info)
   if (not isStateModelLoaded())
   {
     std::cerr << "Error during nnfw_session::train_set_traininfo : invalid state" << std::endl;
-    std::cerr << "nnfw_session::train_set_traininfo must be called in MODEL_LOADED state"
-              << std::endl;
     return NNFW_STATUS_INVALID_STATE;
   }
 
@@ -1191,12 +1189,11 @@ NNFW_STATUS nnfw_session::train_set_traininfo(const nnfw_train_info *info)
 
   // after model loaded, it ensures that _train_info is not nullptr
   assert(_train_info != nullptr);
-  assert(info != nullptr);
 
   auto convertLossType = [](const int &type) {
     if (type == NNFW_TRAIN_LOSS_MEAN_SQUARED_ERROR)
       return onert::ir::train::LossCode::MeanSquaredError;
-    if (type == NNFW_TRAIN_LOSS_CATEGORICAL_CROSSENTROPY)
+    else if (type == NNFW_TRAIN_LOSS_CATEGORICAL_CROSSENTROPY)
       return onert::ir::train::LossCode::CategoricalCrossentropy;
     else
       throw std::runtime_error("not supported loss type");
@@ -1204,7 +1201,7 @@ NNFW_STATUS nnfw_session::train_set_traininfo(const nnfw_train_info *info)
 
   auto convertLossReductionType = [](const int &type) {
     if (type == NNFW_TRAIN_LOSS_REDUCTION_AUTO)
-      return onert::ir::train::LossReductionType::Invalid;
+      return onert::ir::train::LossReductionType::Auto;
     else if (type == NNFW_TRAIN_LOSS_REDUCTION_SUM_OVER_BATCH_SIZE)
       return onert::ir::train::LossReductionType::SumOverBatchSize;
     else if (type == NNFW_TRAIN_LOSS_REDUCTION_SUM)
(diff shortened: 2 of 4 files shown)


runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -220,6 +220,15 @@ typedef struct nnfw_train_info
   NNFW_TRAIN_OPTIMIZER opt = NNFW_TRAIN_OPTIMIZER_SGD;
 } nnfw_train_info;
```
zetwhite, line 222: ### For reviewers

This PR just adds one new API (_train_info setter - named `nnfw_train_set_traininfo`). 
But I have a plan to change `nnfw_train_prepare` a bit. 
The overall direction is like this. 

#### AS-IS 
* `nnfw_train_prepare()` gets train_info and compiles training model using the given train_info


#### After
* `nnfw_set_train_traininfo()` set train_info into session
* `nnfw_train_prepare()` compile training model using the train_info inside the session

runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -220,6 +220,15 @@ typedef struct nnfw_train_info
...
 
+/**
+ * @brief Set train_info into session
+ * @note  This function should be called in MODEL_LOADED state,
+ *        after {@link nnfw_load_model_from_file} before {@link nnfw_prepare}
+ *
+ *  @return @c NNFW_STATUS_NO_ERROR If successful
+ */
+NNFW_STATUS nnfw_train_set_traininfo(nnfw_session *session, const nnfw_train_info *train_info);
+
```
zetwhite, line 231: **This API can called after `model_loaded`.** 

Because.. If the user set train_info before the model loaded,
It is possible that train_info in the session is overwritten when the model is loaded.

So, To avoid unexpected overwritten while model loaded, Let's make sure that `nnfw_train_set_traininfo` called after model loaded.

runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -220,6 +220,15 @@ typedef struct nnfw_train_info
   NNFW_TRAIN_OPTIMIZER opt = NNFW_TRAIN_OPTIMIZER_SGD;
 } nnfw_train_info;
 
+/**
+ * @brief Set train_info into session
+ * @note  This function should be called in MODEL_LOADED state,
```
ragmani, line 225: User cannot see `MODEL_LOADED` state, how about removing the content related to `MODEL_LOADED`?

zetwhite: Updated :)

runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -220,6 +220,15 @@ typedef struct nnfw_train_info
...
 } nnfw_train_info;
 
+/**
+ * @brief Set train_info into session
+ * @note  This function should be called after calling {@link nnfw_load_model_from_file}
+ *        and before calling {@link nnfw_prepare}
+ *
+ *  @return @c NNFW_STATUS_NO_ERROR If successful
+ */
+NNFW_STATUS nnfw_train_set_traininfo(nnfw_session *session, const nnfw_train_info *train_info);
```
zetwhite, line 230: **This API should be called after model_loaded.**

Because.. If the user set train_info before the model loaded,
train_info in the session can be overwritten while the model loaded.
This can be confusing.

So, To avoid unexpected overwritten while the model loaded, Let's make sure that nnfw_train_set_traininfo is called after the model loaded.

runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -220,6 +220,15 @@ typedef struct nnfw_train_info
   NNFW_TRAIN_OPTIMIZER opt = NNFW_TRAIN_OPTIMIZER_SGD;
 } nnfw_train_info;
 
+/**
+ * @brief Set train_info into session
+ * @note  This function should be called after calling {@link nnfw_load_model_from_file}
+ *        and before calling {@link nnfw_prepare}
```
ragmani, line 226: Maybe?
```suggestion
 *        and before calling {@link nnfw_train_prepare}
```

BTW, is there anything else to be modified in `nnfw_train_prepare` due to this commit?

zetwhite: Ah, Updated. 

> BTW, is there anything else to be modified in nnfw_train_prepare due to this commit? 

I'd like to update `nnfw_train_prepare` like this.
* remove `nnfw_train_info` argument and update to use session's train_info.  
https://github.com/Samsung/ONE/pull/12383/files#r1437471834

runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -220,6 +220,15 @@ typedef struct nnfw_train_info
   NNFW_TRAIN_OPTIMIZER opt = NNFW_TRAIN_OPTIMIZER_SGD;
 } nnfw_train_info;
 
+/**
+ * @brief Set train_info into session
```
glistening, line 224: Please don't use variable name for description for user.
```suggestion
 * @brief Set training info
```

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1173,6 +1173,78 @@ NNFW_STATUS nnfw_session::set_backends_per_operation(const char *backend_setting
   return NNFW_STATUS_NO_ERROR;
 }
 
+NNFW_STATUS nnfw_session::train_set_traininfo(const nnfw_train_info *info)
+{
+  if (not isStateModelLoaded())
+  {
+    std::cerr << "Error during nnfw_session::train_set_traininfo : invalid state" << std::endl;
+    std::cerr << "nnfw_session::train_set_traininfo must be called in MODEL_LOADED state"
```
jyoungyun, line 1181:
```suggestion
```

Invalid state error is enough. The `MODEL_LOADED` state is internal status. So the user will not be able to understand what it means.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1173,6 +1173,78 @@ NNFW_STATUS nnfw_session::set_backends_per_operation(const char *backend_setting
...
+
+  if (info == nullptr)
+  {
+    std::cerr << "nnfw_session::train_set_traininfo : info is nullptr" << std::endl;
+    return NNFW_STATUS_UNEXPECTED_NULL;
+  }
+
+  // after model loaded, it ensures that _train_info is not nullptr
+  assert(_train_info != nullptr);
+  assert(info != nullptr);
```
jyoungyun, line 1194:
```suggestion
```
This error check logic is duplicated.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1173,6 +1173,78 @@ NNFW_STATUS nnfw_session::set_backends_per_operation(const char *backend_setting
...
+
+  auto convertLossType = [](const int &type) {
+    if (type == NNFW_TRAIN_LOSS_MEAN_SQUARED_ERROR)
+      return onert::ir::train::LossCode::MeanSquaredError;
+    if (type == NNFW_TRAIN_LOSS_CATEGORICAL_CROSSENTROPY)
```
jyoungyun, line 1199:
```suggestion
    else if (type == NNFW_TRAIN_LOSS_CATEGORICAL_CROSSENTROPY)
```

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1173,6 +1173,78 @@ NNFW_STATUS nnfw_session::set_backends_per_operation(const char *backend_setting
...
+      return onert::ir::train::LossCode::CategoricalCrossentropy;
+    else
+      throw std::runtime_error("not supported loss type");
+  };
+
+  auto convertLossReductionType = [](const int &type) {
+    if (type == NNFW_TRAIN_LOSS_REDUCTION_AUTO)
+      return onert::ir::train::LossReductionType::Invalid;
```
jyoungyun, line 1207:
```suggestion
      return onert::ir::train::LossReductionType::Auto;
```
Maybe it should be Auto type?

runtime/onert/api/include/nnfw_experimental.h
zetwhite: Thank you. Updated :)

runtime/onert/api/src/nnfw_api_internal.cc
zetwhite: Updated :)

runtime/onert/api/src/nnfw_api_internal.cc
zetwhite: Thank you for catching :) 
Updated

jyoungyun (approved):
LGTM 👍

runtime/onert/api/include/nnfw_experimental.h
ragmani: I see, there are changes in the draft PR. I will review it later on the PR that it will be applied.

ragmani (approved):
LGTM

runtime/onert/api/include/nnfw_experimental.h
glistening: Please remove 'into session" in description as I suggested. I think "into session" is better to be removed.

runtime/onert/api/include/nnfw_experimental.h
```diff
@@ -220,6 +220,15 @@ typedef struct nnfw_train_info
   NNFW_TRAIN_OPTIMIZER opt = NNFW_TRAIN_OPTIMIZER_SGD;
 } nnfw_train_info;
 
+/**
+ * @brief Set training info into session
```
zetwhite, line 224:
```suggestion
 * @brief Set training info
```

runtime/onert/api/include/nnfw_experimental.h
zetwhite: updated :)

glistening (approved):
LGTM

jyoungyun (approved):
LGTM :)

ragmani (approved):

Issue #12110, opened 2023-11-24
[onert] Reduce the code changes for training


glistening:
As basic training is working, I would like to clean up and reduce code changes which was introduced for training in tight schedule.

- [ ] #12111
- [ ] Duplication of Compiler: `TrainingCompiler` and `Compiler`
- [ ] Duplication of `ITensorRegistry`
- [ ] ...

### Why?

It is hard to identify what is changed because there are much of duplicated code.

It would be good to 

1) remove code duplication or
2) at least make the difference noticeable.

### An example

For example, as 1st step, I tried to remove `ir::train::operation::SOME_OP`,
which has almost same to `ir::operation::SOME_OP` except for `visit()` and `clone()`.

I can remove a lot of code, and it seems to work.
I would like to hear other's opinion.

(ADDED) You can see several opinions and concerns from #12111.

glistening:
https://github.com/Samsung/ONE/blob/7b6efc699a99d2947a502f9813f4c5c9c1598656/runtime/onert/core/src/compiler/Compiler.cc#L47-L197

vs

https://github.com/Samsung/ONE/blob/7b6efc699a99d2947a502f9813f4c5c9c1598656/runtime/onert/core/src/compiler/train/TrainingCompiler.cc#L60-L298

glistening:
https://github.com/Samsung/ONE/blob/7b6efc699a99d2947a502f9813f4c5c9c1598656/runtime/onert/core/include/backend/ITensorRegistry.h#L74-L144

https://github.com/Samsung/ONE/blob/7b6efc699a99d2947a502f9813f4c5c9c1598656/runtime/onert/core/include/backend/train/ITensorRegistry.h#L64-L245

PR #12385, opened 2023-12-28
[onert] Implement DepthwiseConv2D shape inferer

jyoungyun:
This commit implements DepthwiseConv2D shape inferer.

### DIFF
--- a/runtime/onert/core/include/compiler/StaticShapeInferer.h
+++ b/runtime/onert/core/include/compiler/StaticShapeInferer.h
@@ -126,6 +126,7 @@ private:
   void visit(const ir::operation::Comparison &op) override;
   void visit(const ir::operation::Concat &op) override;
   void visit(const ir::operation::Conv2D &op) override;
+  void visit(const ir::operation::DepthwiseConv2D &op) override;
   void visit(const ir::operation::ElementwiseActivation &op) override;
   void visit(const ir::operation::ElementwiseBinary &op) override;
   void visit(const ir::operation::ElementwiseUnary &op) override;
--- a/runtime/onert/core/src/compiler/StaticShapeInferer.cc
+++ b/runtime/onert/core/src/compiler/StaticShapeInferer.cc
@@ -453,6 +453,23 @@ void StaticShapeInferer::visit(const ir::operation::Conv2D &op)
   output.info().shape(new_shape);
 }
 
+void StaticShapeInferer::visit(const ir::operation::DepthwiseConv2D &op)
+{
+  auto &operands = _lowered_subg->graph().operands();
+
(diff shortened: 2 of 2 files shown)


glistening:
It is a part of #12110. Shape inference code are duplicated as the training operation grows. I tried to remove the duplication (e.g. #12111). But you and @ragmani don't want the approach. If so, it would be good if we have alternative solution to reduce code duplication.

jyoungyun:
@glistening  This PR is not related to the train feature.

glistening (approved):
LGTM. This PR itself is ok. But it would be nice if we can reduce the code duplication.

(ADD) Ah, it is not under `train`. May I ask why it became necessary?

jyoungyun:
> May I ask why it became necessary?

If the batch size is changed, the output shape of DepthwiseConv2D is not changed even if the input is changed.
This is because the ShapeInferer visit function of the DepthwiseConv2D was not implemented so I fixed it.

glistening:
@jyoungyun I am not familiar with batch_size change. Here are several questions:

1. I guess `batch_size` is different to the original `batch_size` which is written in model file. What was the original `batch_size` ?
2. How did you change `batch_size`? I guess you used `nnfw_apply_tensorinfo` or `nnfw_set_input_tensorinfo`. Which one did you use?
3. Is https://github.com/Samsung/ONE/pull/12385#issuecomment-1871654379 the log of "before-this-PR" or "after-this-PR"?

jyoungyun:
> 1. I guess `batch_size` is different to the original `batch_size` which is written in model file. What was the original `batch_size` ?

The original batch_size is written in model file. Almost the batch size in our model files is 1.

> 2. How did you change `batch_size`? I guess you used `nnfw_apply_tensorinfo` or `nnfw_set_input_tensorinfo`. Which one did you use?

If you change the batch size of input operation, this automatically changes the batch size of all ops through the shape inferer.

https://github.com/Samsung/ONE/blob/64b87a6618e3883e5ff42023de91a3cd7c05cf5d/runtime/onert/core/src/compiler/train/TrainingCompiler.cc#L167-L182

> 3. Is the log of https://github.com/Samsung/ONE/pull/12385#issuecomment-1871654379 "before-this-PR" or "after-this-PR"?

This log is "before-this-PR".

zetwhite (approved):
LGTM

glistening:
> Changing the batch size of an input automatically changes the batch size of all ops through the shape inferer.

My question was not about this. I asked how the batch size was set. I guess it was done one of two what I've guessed. Anyway, it is beyond this PR.

> This log is "before-this-PR".

Yes, I thought the shape is not correct. I don't know what is the value of "after-this-PR". But as the core was there before this PR and has test. I guess you've tested.

glistening (approved):
LGTM

PR #12390, opened 2023-12-29
[onert] Merge _training_step into TrainingInfo

jyoungyun:
This commit merges _training_step into TrainingInfo. It removes duplicated variable.

related comment: https://github.com/Samsung/ONE/pull/12383#issuecomment-1871732299

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -1264,8 +1264,8 @@ NNFW_STATUS nnfw_session::train_prepare()
     if (not _train_info->isValid())
       throw std::runtime_error{"training info is not valid"};
 
-    // initialize epoch count
-    _train_info->epoch() = 0;
+    // initialize trainingStep count
+    _train_info->trainingStep() = 0;
 
     auto compiler =
       onert::compiler::CompilerFactory::get().create(_nnpkg, _coptions, _train_info.get());
@@ -1452,7 +1452,7 @@ NNFW_STATUS nnfw_session::train_run(bool update_weights)
   {
     if (update_weights)
     {
-      auto &training_step = _train_info->epoch();
+      auto &training_step = _train_info->trainingStep();
       _execution->train(training_step++);
     }
     else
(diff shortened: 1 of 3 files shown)


zetwhite (dismissed):
LGTM

runtime/onert/core/include/ir/train/TrainingInfo.h
```diff
@@ -43,7 +43,7 @@ class TrainingInfo final
   const LossInfo &lossInfo() const { return _loss_info; }
   const OptimizerInfo &optimizerInfo() const { return _optimizer_info; }
   uint32_t batchSize() const { return _batch_size; }
-  uint32_t epoch() const { return _epoch; }
+  uint32_t &epoch() { return _epoch; }
```
ragmani, line 46: Could you add a member function that returns it by const reference? 
```suggestion
  const uint32_t &epoch() const { return _epoch; }
```

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1449,7 +1452,8 @@ NNFW_STATUS nnfw_session::train_run(bool update_weights)
   {
     if (update_weights)
     {
-      _execution->train(_training_step++);
+      auto &training_step = _train_info->epoch();
```
ragmani, line 1455: `training_step` and `epoch` are different. Why did you merge them?

jyoungyun: There is no need to keep the `epoch` number in `TrainingInfo` because the epoch information depends on the user's argument, not related to the model.

And the `training_step` is located directly in the `Session` object. However, we introduced the `TrainingInfo` to session object. This variable can be merged into the `TrainingInfo` to manage information related to training in one place.

So I tried to merge `training_step` into the `TrainingInfo` while changing the name to `epoch`.

How about merging the `training_step` variable to the `TrainingInfo` while keeping the same name?

ragmani: The name `training_step` looks better.

jyoungyun: > The name `training_step` looks better.

This code is outdated. Please see the new patch. :)

ragmani (approved):
LGTM

zetwhite (approved):

Issue #12388, opened 2023-12-29
[onert] Support ReLU6 for training


zetwhite:
### What

Let's support ReLU6 for training

parent issue : https://github.com/Samsung/ONE/issues/12325 

### Task

* draft : https://github.com/Samsung/ONE/pull/12395 

- [x] Add ReLU6 ckernel https://github.com/Samsung/ONE/pull/12476
- [x] Apply ReLU6 to elementwise activation layer https://github.com/Samsung/ONE/pull/12487
- [x] Apply ReLU6 to fused activation of each layer 
        - https://github.com/Samsung/ONE/pull/12492 
        - https://github.com/Samsung/ONE/pull/12502

zetwhite:
Draft : https://github.com/Samsung/ONE/pull/12395 

I checked both fused activation, and stand-alone activation works well. 

# fused one
[model.zip](https://github.com/Samsung/ONE/files/13936487/model.zip)

* tested env
```
- learning_rate   = 0.001
- batch_size      = 32
- loss_info       = {loss = mean squared error, reduction = sum over batch size}
- optimizer       = adam 
```


* tensorflow
```
Epoch 1/5
32/32 [==============================] - 0s 947us/step - loss: 0.0518 - mean_squared_error: 0.0518
Epoch 2/5
32/32 [==============================] - 0s 847us/step - loss: 0.0364 - mean_squared_error: 0.0364
Epoch 3/5
32/32 [==============================] - 0s 876us/step - loss: 0.0277 - mean_squared_error: 0.0277
Epoch 4/5
32/32 [==============================] - 0s 898us/step - loss: 0.0223 - mean_squared_error: 0.0223
Epoch 5/5
32/32 [==============================] - 0s 862us/step - loss: 0.0186 - mean_squared_error: 0.0186

```

* onert_train
```
Epoch 1/5 - time: 37.853ms/step - loss: [0] 0.0518
Epoch 2/5 - time: 37.927ms/step - loss: [0] 0.0361
Epoch 3/5 - time: 37.778ms/step - loss: [0] 0.0272
Epoch 4/5 - time: 37.771ms/step - loss: [0] 0.0218
Epoch 5/5 - time: 38.018ms/step - loss: [0] 0.0180

``` 


# not fused one
[model.zip](https://github.com/Samsung/ONE/files/13936464/model.zip)


* tested env 
```
- learning_rate   = 0.001
- batch_size      = 32
[...]

PR #12395, opened 2024-01-02
Draft: [onert] support ReLU6 training

zetwhite:
This draft is to support ReLU6 training feature.

 - Add ReLU6Grad cker
 - Extract fused activation gradient calculation part to reuse
 - Add Relu6 to EW activation layer 

related : https://github.com/Samsung/ONE/issues/12388

### DIFF
--- a/compute/cker/include/cker/train/operation/ReLU6.h
+++ b/compute/cker/include/cker/train/operation/ReLU6.h
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2024 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __NNFW_CKER_TRAIN_OPERATION_RELU6_H__
+#define __NNFW_CKER_TRAIN_OPERATION_RELU6_H__
+
+#include "cker/Shape.h"
+#include "cker/eigen/Utils.h"
+#include <Eigen/Core>
+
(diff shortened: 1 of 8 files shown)


compute/cker/include/cker/train/operation/ReLU6.h
```diff
@@ -0,0 +1,51 @@
...
+                      const Shape &grad_shape, float *grad_data)
+{
+  const auto output_map = MapAsVector(output_data, output_shape);
+  const auto incoming_map = MapAsVector(incoming_data, incoming_shape);
+  auto grad_map = MapAsVector(grad_data, grad_shape);
+
+  if (output_shape == incoming_shape && output_shape == grad_shape)
+    grad_map.array() =
+      incoming_map.array() *
+      (0.0f < output_map.array() && output_map.array() < 6.0f).template cast<float>();
```
zetwhite, line 42: (note)

`array()` has no cost - it just change the view of data, for element-wise operation

https://eigen.tuxfamily.org/dox/group__QuickRefPage.html#:~:text=Recall%20that%20.array()%20has%20no%20cost%2C%20it%20only%20changes%20the%20available%20API%20and%20interpretation%20of%20the%20data

runtime/onert/backend/train/ops/OperationUtils.cc
```diff
@@ -0,0 +1,65 @@
...
+                                   getShape(input_backprop), getBuffer<float>(input_backprop),
+                                   getShape(output_backprop), getBuffer<float>(output_backprop));
+      res = output_backprop;
+      break;
+    default:
+      throw std::runtime_error("Unsupported activation type yet");
+  }
+
+  return res;
+}
```
zetwhite, line 60: @ragmani 

While I was supporting `relu6`, 
I found out that each layer(Convolution, FullyConnected and Pool for now) implements activation backpropagation on its own. 
 
I'd like to extract fused activation's backpropagation part as OperationUtil to remove code duplication. 

But before going on, I'd like to hear your opinion. Do you think it is fine?  
( I'm a bit afraid that there's a reason it has the current implementation/structure. )

ragmani: Sorry for the late reply. Perhaps, I missed your comment.

zetwhite: No problem! 
I didn't ask again, because it wasn't in a hurry at all.

zetwhite:
related code is all merged. so close this Draft :)

PR #12397, opened 2024-01-03
[luci/pass] Revise FoldCastPass to early check

seanshpark:
This will revise FoldCastPass to earcky check types to skip casting,
not to unnecessary create CircleConst.


jinevening (approved):
LGTM

(optional) There exist the same `if` statements below. I think we can reduce duplication by moving constant creation logic inside the below if-statements.

seanshpark:
> I think we can reduce duplication by moving constant creation logic inside the below if-statements.

I left constant creation outside of `if` statements to preserve original authors writings.
Instead, I'll propose another way...

PR #12403, opened 2024-01-03
[onert] Update TrainingInfo::isValid() to return error mesage

zetwhite:
This PR is to update TrainingInfo::isValid() to return error message.

### DIFF
--- a/runtime/onert/api/src/nnfw_api_internal.cc
+++ b/runtime/onert/api/src/nnfw_api_internal.cc
@@ -1346,8 +1346,9 @@ NNFW_STATUS nnfw_session::train_prepare()
 
   try
   {
-    if (not _train_info->isValid())
-      throw std::runtime_error{"training info is not valid"};
+    const auto res = _train_info->isValid();
+    if (not res.valid)
+      throw std::runtime_error{"training info is not valid \n" + res.error_msg};
 
     auto compiler =
       onert::compiler::CompilerFactory::get().create(_nnpkg, _coptions, _train_info.get());
--- a/runtime/onert/core/include/ir/train/TrainingInfo.h
+++ b/runtime/onert/core/include/ir/train/TrainingInfo.h
@@ -59,7 +59,13 @@ public:
   void setOptimizerInfo(const OptimizerInfo &optimizer_info) { _optimizer_info = optimizer_info; }
   void setEpoch(uint32_t epoch) { _epoch = epoch; }
 
-  bool isValid() const;
+  struct ValidationResult
+  {
+    bool valid;
+    std::string error_msg;
+  };
+
+  ValidationResult isValid() const;
 
 private:
   LossInfo _loss_info;
--- a/runtime/onert/core/src/ir/train/TrainingInfo.cc
+++ b/runtime/onert/core/src/ir/train/TrainingInfo.cc
@@ -23,25 +23,43 @@ namespace ir
 namespace train
 {
 
-bool TrainingInfo::isValid() const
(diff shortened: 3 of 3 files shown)


runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1346,8 +1346,9 @@ NNFW_STATUS nnfw_session::train_prepare()
 
   try
   {
-    if (not _train_info->isValid())
-      throw std::runtime_error{"training info is not valid"};
+    const auto res = _train_info->isValid();
+    if (not res.valid)
+      throw std::runtime_error{"training info is not valid \n" + res.error_msg};
```
zetwhite, line 1351: After this PR, `train_prepare` show that - why training info is invalid. 

### Before this PR 

When the user sets invalid training info, 
`train_prepare`fails without showing which training information is invalid.

```
./onert_train --modelfile mnist_with_meta.circle --load_input:raw pool_mnist_model/mnist_x_train.1000.bin --load_expected:raw pool_mnist_model/mnist_y_train.1000.bin --epoch 10  --learning_rate=0 --batch_size=0

Model Expected Filename pool_mnist_model/mnist_y_train.1000.bin
Model Input Filename pool_mnist_model/mnist_x_train.1000.bin
Model Filename mnist_with_meta.circle
== training parameter ==
- learning_rate   = 0
- batch_size      = 0
- loss_info       = {loss = mean squared error, reduction = use automatic reduction type}
- optimizer       = sgd
========================
Error during nnfw_session::train_prepare : training info is not valid 
```

### After this PR 

`train_prepare` shows the message like 'batch size must not be zero', 'learning rate must be positive'. 

```bash 
./onert_train --modelfile mnist_with_meta.circle --load_input:raw pool_mnist_model/mnist_x_train.1000.bin --load_expected:raw pool_mnist_model/mnist_y_train.1000.bin --epoch 10  --learning_rate=0 --batch_size=0

Model Expected Filename pool_mnist_model/mnist_y_train.1000.bin
Model Input Filename pool_mnist_model/mnist_x_train.1000.bin
Model Filename mnist_with_meta.circle
== training parameter ==
- learning_rate   = 0
- batch_size      = 0
- loss_info       = {loss = mean squared error, reduction = use automatic reduction type}
- optimizer       = sgd
========================
Error during nnfw_session::train_prepare : training info is not valid 
batch size must not be zero
learning rate must be positive

```

jyoungyun:
I don't want to interrupt you, but I'm not sure if this PR is really necessary. I think it would be better to give a well-described document than to print information of all parameters as error messages.

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -1346,8 +1346,9 @@ NNFW_STATUS nnfw_session::train_prepare()
 
   try
   {
-    if (not _train_info->isValid())
-      throw std::runtime_error{"training info is not valid"};
+    const auto res = _train_info->isValid();
+    if (not res.valid)
```
glistening, line 1350: Just a note. Runtime have not used `not`, we've used `!` so far. I am okay for using alternative tokens.

zetwhite:
> I don't want to interrupt you, but I'm not sure if this PR is really necessary. 

Thanks for the opinion :) 
As you pointed out, this isn't not the one which is really necessary.
My intention was providing a kind API - pointing out detailed reason why it fails to compile(prepare).
I'm also satisfied with giving detailed description, instead of giving detailed error message.

zetwhite:
Let's close this one for now.
I'll reopen this when it become really necessary.

PR #12408, opened 2024-01-05
[onert/api] Load TrainingInfo from the session model

zetwhite:
This PR updates model loading api to load TrainingInfo from the model.

related : https://github.com/Samsung/ONE/issues/11692


runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
```diff
@@ -27,7 +27,7 @@ namespace train
 namespace traininfo_loader
 {
 
-static constexpr char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
+static const constexpr char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
```
glistening, line 30: @zetwhite May I ask why `const` is added here?

zetwhite: About this part, I made a separate PR. 
Could you check this one? https://github.com/Samsung/ONE/pull/12410

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -317,8 +331,7 @@ NNFW_STATUS nnfw_session::load_model_from_modelfile(const char *model_file_path)
     _model_path = std::string(model_file_path);
     _nnpkg = std::make_shared<onert::ir::NNPkg>(std::move(model));
     _coptions.push_back(onert::compiler::CompilerOptions::fromGlobalConfig());
-    // TODO load TrainingInfo from model, using TraininfoLoader
-    _train_info = std::make_unique<onert::ir::train::TrainingInfo>();
+    _train_info = loadTrainingInfo(_nnpkg->primary_model());
```
zetwhite, line 334: I checked with gdb that `_train_info` is loaded as expected in my local env. 

```bash
# run without training argument - cmd line argument 
➜ gdb --args \
~/Workspace/ONE_second/Product/build/bin/onert_train \
--modelfile mnist_with_meta.circle \
--load_input:raw pool_mnist_model/mnist_x_train.1000.bin \
--load_expected:raw pool_mnist_model/mnist_y_train.1000.bin
 ... 
Breakpoint 1, nnfw_session::load_model_from_modelfile ... 

# before loadTrainingInfo, _train_info is nullptr
334	    _train_info = loadTrainingInfo(_nnpkg->primary_model());
(gdb) p _train_info
$1 = std::unique_ptr<onert::ir::train::TrainingInfo> = {get() = 0x0}

# after laodTrainingInfo, _train_info is filled with circle file values 
(gdb) n
335	    _state = State::MODEL_LOADED;
(gdb) print *_train_info
$2 = {_loss_info = {loss_code = onert::ir::train::LossCode::MeanSquaredError,  .... } 


```

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -202,6 +203,18 @@ std::unique_ptr<onert::ir::Model> loadModel(const std::string filename,
   return std::unique_ptr<onert::ir::Model>(nullptr);
 }
 
+std::unique_ptr<onert::ir::train::TrainingInfo>
+loadTrainingInfo(std::shared_ptr<onert::ir::Model> &&model)
```
ragmani, line 207:
```suggestion
loadTrainingInfo(const std::shared_ptr<onert::ir::Model> &model)
```

zetwhite: `extract_metadata` (used in line 212) is not a const function, It could not have `const` here.

zetwhite: Also, `_nnpkg->primary_model()` (in this file, lie 292) returns shared_ptr; 
Here, shared_ptr itself looks like rvalue.

https://github.com/Samsung/ONE/blob/c925d8ae740700a1293d6a27f57c51e447650026/runtime/onert/core/include/ir/NNPkg.h#L92 

I'd like to reference shared_ptr itself (to not increase reference count), I used rvalue reference `std::shared_ptr<onert::ir::Model>&&`.

ragmani: > extract_metadata (used in line 212) is not a const function, It could not have const here.

`const std::shared_ptr<Model>` does not mean the object what the `std::shared_ptr` points to (`const Model *`), so it's OK if `const std::share` calls a const function of the object what it points to.

ragmani: > Also, _nnpkg->primary_model() (in this file, lie 292) returns shared_ptr.
Here, shared_ptr itself looks like rvalue.
> I'd like to reference shared_ptr itself (to not increase reference count), I used rvalue reference std::shared_ptr<onert::ir::Model>&&.

It seems to return value, not reference. I think there is no way not to increase the reference count unless you change `primary_model()`.

zetwhite: Aha, Thank you a lot for clarifying - what really happened in this line :+1: :+1:

I updated it as you first suggested! - use const lvalue reference. :smile:

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -202,6 +203,18 @@ std::unique_ptr<onert::ir::Model> loadModel(const std::string filename,
   return std::unique_ptr<onert::ir::Model>(nullptr);
 }
 
+std::unique_ptr<onert::ir::train::TrainingInfo>
+loadTrainingInfo(const std::shared_ptr<onert::ir::Model> &model)
+{
+  const auto tinfo_name = onert::train::traininfo_loader::TRAININFO_METADATA_NAME;
```
jyoungyun, line 209: For a while, I considered this `tinfo_name` should be `const auto &` not `const auto`. Soon I found out the `TRAINING_METADATA_NAME` is a char pointer and realized that there is no need to use `&`. In order to enhance the code readability, it would be a good solution to use the `std::string` instead of `char pointer`. I made a PR about my suggestion #12424. This is just my suggestion and I don't force it. :)

jyoungyun (approved):
LGTM

runtime/onert/api/src/nnfw_api_internal.cc
zetwhite: Thank you for the suggestion and the draft PR! :smile: 
I like the way you suggested - it improves readability and efficiency. 

But as you know, I already fixed traininfo_loader once :sweat_smile: - https://github.com/Samsung/ONE/pull/12410. 
And  I'd like to close this issue - https://github.com/Samsung/ONE/issues/11692#issuecomment-1855114192 as soon as possible. (After this PR, one last thing is to remove the default argument from onert_train.) 

So, If you don't mind I'd like to merge this PR as it is and... fix it to use std::string instead of char* in the next PR.

jyoungyun: My suggestion requires several changes except for this PR. So I want to merge it as it is. :)
If you don't mind, I can apply my suggestion after merged this PR with other reviewer's approval.

zetwhite: > If you don't mind, I can apply my suggestion after merged this PR with other reviewer's approval.

I appreciate it :grinning: Thank you.

zetwhite:
@ragmani @glistening Could you take another look?

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -24,6 +24,7 @@
 #include "circle_loader.h"
 #include "tflite_loader.h"
 #include "trix_loader.h"
+#include "traininfo_loader.h"
```
glistening, line 27: (Out of curiosity) Why do you add `#include` here? It is not the end of `#include`, and it is not in alphabetical order. If you intended almost-alphabetical order, it would be better to put before `trix_loader.h`.

glistening (approved):
LGTM

ragmani (approved):
LGTM

runtime/onert/api/src/nnfw_api_internal.cc
```diff
@@ -202,6 +203,18 @@ std::unique_ptr<onert::ir::Model> loadModel(const std::string filename,
   return std::unique_ptr<onert::ir::Model>(nullptr);
 }
 
+std::unique_ptr<onert::ir::train::TrainingInfo>
+loadTrainingInfo(const std::shared_ptr<onert::ir::Model> &model)
+{
+  const auto tinfo_name = onert::train::traininfo_loader::TRAININFO_METADATA_NAME;
+  if (model->exists_metadata(tinfo_name))
+  {
+    const auto buffer = model->extract_metadata(tinfo_name);
```
ragmani, line 212: I have a question regardless to this PR.
Is `model` going to always have `metadata` as external data(i.g. `ir::MMapedData`) in the future?

ragmani: I'll ask the above question in another issue since this PR is closed.

zetwhite: Let me clarify the word first, to make sure that we have the same understanding. 

* `ir::CachedData` : holds **copy** of original data
* `ir::ExternalData` : points original data, we have to make sure that the original data is not lost.

zetwhite: > Is model going to have metadata as only external data(i.g. ir::MMapedData) in the future? 

@ragmani 
For now, only `TrainingInfo` is passed through Model's Metadata.
Since TrainingInfo is extracted when **loaded** time, we could use `ExternalData`.
If Metadata is used(extracted) at another time, I'm not sure that the original buffer (mmaped data) is not damaged. 

So, About your question, For now, Yes only external data is used in metadata.
About the future, I'm not 100% sure.

ragmani: This PR was closed. So I copied the above comment into #12433

PR #12410, opened 2024-01-05
[onert/frontend] Revisit traininfo metadata name

zetwhite:
This PR adds 'const' keyword to the traininfo metadata name.
It also adds 'extern' keyword to follow one-definition-rule.

### DIFF
--- a/runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
+++ b/runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
@@ -27,7 +27,7 @@ namespace train
 namespace traininfo_loader
 {
 
-static constexpr char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
+static const char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
 
 std::unique_ptr<ir::train::TrainingInfo> loadTrainingInfo(const uint8_t *buffer, const size_t size);
 
--- a/runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
+++ b/runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
@@ -27,7 +27,7 @@ namespace train
 namespace traininfo_loader
 {
 
-static const char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
+extern const char *const TRAININFO_METADATA_NAME;
 
 std::unique_ptr<ir::train::TrainingInfo> loadTrainingInfo(const uint8_t *buffer, const size_t size);
 
--- a/runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
+++ b/runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
@@ -27,6 +27,7 @@ namespace train
 namespace traininfo_loader
 {
 
+// TODO change this line to use inline variable after C++17
 extern const char *const TRAININFO_METADATA_NAME;
 
 std::unique_ptr<ir::train::TrainingInfo> loadTrainingInfo(const uint8_t *buffer, const size_t size);
--- a/runtime/onert/frontend/circle_traininfo/src/traininfo_loader.cc
+++ b/runtime/onert/frontend/circle_traininfo/src/traininfo_loader.cc
@@ -25,6 +25,8 @@ namespace train
 namespace traininfo_loader
 {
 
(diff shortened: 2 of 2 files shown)


runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
```diff
@@ -27,7 +27,7 @@ namespace train
 namespace traininfo_loader
 {
 
-static constexpr char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
+static const char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
```
zetwhite, line 30: While I was on the next step, I found an error :  
```bash 
ONE/runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h:30:56: 
error: ISO C++ forbids converting a string constant to ‘char*’ [-Werror=write-strings]
   30 | static constexpr char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
      |                                                        ^~~~~~~~~~~~~~~~~

``` 

It looks like `constexpr` doesn't guarantee that char-array(string) content is `const`.
So, For clearness, Let's just remove `constexpr` and use `const` instead.

-------------------------------------------------------------


I refered stack-overflow https://stackoverflow.com/questions/30561104/const-constexpr-char-vs-constexpr-char.
I tried to find an official C++ document about this, But it's hard to find a clear explanation about this.

runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
```diff
@@ -27,7 +27,7 @@ namespace train
 namespace traininfo_loader
 {
 
-static constexpr char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
+extern const char *const TRAININFO_METADATA_NAME;
```
zetwhite, line 30: While I was on the next step ([next PR](https://github.com/Samsung/ONE/pull/12408)), I found an error :  
```bash 
ONE/runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h:30:56: 
error: ISO C++ forbids converting a string constant to ‘char*’ [-Werror=write-strings]
   30 | static constexpr char *const TRAININFO_METADATA_NAME = "CIRCLE_TRAINING";
      |                                                        ^~~~~~~~~~~~~~~~~

``` 

It looks like `constexpr char*` doesn't mean that char-array(string) content is constant.
So, For clearness, Let's just remove `constexpr` and use `const` instead.

-------------------------------------------------------------


I refered stack-overflow https://stackoverflow.com/questions/30561104/const-constexpr-char-vs-constexpr-char.
I tried to find an official C++ document about this, But it's hard to find a clear explanation about this.

ragmani: Could you add a comment like below:
```
// TODO Change to inline variables since C++17
```

zetwhite: PTAL :)

YongseopKim: 👍

YongseopKim (approved):
LGTM and I have a question.

by zetwhite

> This PR adds 'const' keyword to the traininfo metadata name.
> It also adds 'extern' keyword to follow one-definition-rule.

Where can I find the reference to `one-definition-rule`?

zetwhite:
> Where can I find the reference to `one-definition-rule`?

Ah, I didn't put the reference about `one-definition-rule`.
I put things below : 

*  [cpp-reference](https://en.cppreference.com/w/cpp/language/definition#:~:text=Only%20one%20definition%20of%20any%20variable%2C%20function%2C%20class%20type%2C%20enumeration%20type%2C%20concept(since%20C%2B%2B20)%20or%20template%20is%20allowed%20in%20any%20one%20translation%20unit)
  Here, You could find an official document about one-definition-rule 

* best way to declare and define global variable : [link-from-stackoveflow](https://edisciplinas.usp.br/pluginfile.php/5453726/mod_resource/content/0/Extern%20Global%20Variable.pdf )
  Since the official document is a bit broad, I recommend reading this note (I found this on StackOverflow)  
  This one is not highly reliable as an official document, but easy to catch this PR's changes.

jyoungyun (approved):
LGTM

ragmani (approved):
LGTM

PR #12412, opened 2024-01-05
[tools/onert_train] Add *_UNDEF to the dump utility

zetwhite:
This PR adds *_UNDEF enum to the ostream print utility.

### DIFF
--- a/tests/tools/onert_train/src/nnfw_util.cc
+++ b/tests/tools/onert_train/src/nnfw_util.cc
@@ -53,6 +53,9 @@ std::ostream &operator<<(std::ostream &os, const NNFW_TRAIN_OPTIMIZER &opt)
 {
   switch (opt)
   {
+    case NNFW_TRAIN_OPTIMIZER_UNDEFINED:
+      os << "undefined";
+      break;
     case NNFW_TRAIN_OPTIMIZER_ADAM:
       os << "adam";
       break;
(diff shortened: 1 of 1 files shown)


tests/tools/onert_train/src/nnfw_util.cc
```diff
@@ -53,6 +53,9 @@ std::ostream &operator<<(std::ostream &os, const NNFW_TRAIN_OPTIMIZER &opt)
 {
   switch (opt)
   {
+    case NNFW_TRAIN_OPTIMIZER_UNDEFINED:
+      os << "undefined";
```
YongseopKim, line 57: (Optional items in the future or just stuff)
- I believe the best way is to define a static map or a function to return only these const strings
- This is only used by the `operator<<`
- However, I'm not sure this is easy way in cpp. (I've not worked with cpp for about two years so I'm now not familiar with cpp.)

``cpp
    case NNFW_TRAIN_OPTIMIZER_UNDEFINED:
      os << nnfw_train_optimizer_strs[opt];
```

YongseopKim (approved):
LGTM 👍

jyoungyun (approved):
LGTM

tests/tools/onert_train/src/nnfw_util.cc
ragmani: > However, I'm not sure this is easy way in cpp. (I've not worked with cpp for about two years so I'm now not familiar with cpp.)

@zetwhite 
You can refer to 
https://github.com/Samsung/ONE/blob/f165eaf7553ba65f5f80eeda1500a810f5cd331d/runtime/onert/core/src/ir/operation/Pool2D.cc#L37-L45

ragmani (approved):
LGTM

Issue #12418, opened 2024-01-08
[onert] Let's remove AUTO loss reduction type 


zetwhite:
### What

Let's remove AUTO keyword in loss reduction type in training. 

### Background

In the offline talk last week, we decided to remove the `AUTO` keyword for loss reduction type. 

We considered below reasons : 

   * The meaning of `Auto` and `Default` is interpreted differently by person. 

   * Pytorch defines `None`, `Mean`, `Sum` for loss reduction type (not `Auto`) and it looks like there is no problem with this.

zetwhite:
resolved by https://github.com/Samsung/ONE/pull/12419

PR #12391, opened 2023-12-29
[onert/api] Add nnfw_train_get_traininfo

zetwhite:
This PR adds nnfw_train_get_traininfo API.
This api is to get training info from the session.

ONE-DCO-1.0-Signed-off-by: SeungHui Youn <sseung.youn@samsung.com>

related : https://github.com/Samsung/ONE/issues/11692

PR #12419, opened 2024-01-08
[onert] Remove AUTO from train loss reduction type

zetwhite:
This PR removes AUTO from train loss reduction type.

related issue : https://github.com/Samsung/ONE/issues/12418

### DIFF
--- a/tests/tools/onert_train/src/args.cc
+++ b/tests/tools/onert_train/src/args.cc
@@ -228,10 +228,10 @@ void Args::Initialize(void)
         "Loss type\n"
         "0: MEAN_SQUARED_ERROR (default)\n"
         "1: CATEGORICAL_CROSSENTROPY\n")
-    ("loss_reduction_type", po::value<int>()->default_value(1)->notifier([&] (const auto &v) { _loss_reduction_type = v; }),
+    ("loss_reduction_type", po::value<int>()->default_value(0)->notifier([&] (const auto &v) { _loss_reduction_type = v; }),
         "Loss Reduction type\n"
-        "1: SUM_OVER_BATCH_SIZE(default)\n"
-        "2: SUM\n")
+        "0: SUM_OVER_BATCH_SIZE(default)\n"
+        "1: SUM\n")
     ("optimizer", po::value<int>()->default_value(0)->notifier([&] (const auto &v) { _optimizer_type = v; }),
       "Optimizer type\n"
       "0: SGD (default)\n"
--- a/tests/tools/onert_train/src/args.cc
+++ b/tests/tools/onert_train/src/args.cc
@@ -228,10 +228,10 @@ void Args::Initialize(void)
         "Loss type\n"
         "0: MEAN_SQUARED_ERROR (default)\n"
         "1: CATEGORICAL_CROSSENTROPY\n")
-    ("loss_reduction_type", po::value<int>()->default_value(0)->notifier([&] (const auto &v) { _loss_reduction_type = v; }),
+    ("loss_reduction_type", po::value<int>()->default_value(1)->notifier([&] (const auto &v) { _loss_reduction_type = v; }),
         "Loss Reduction type\n"
-        "0: SUM_OVER_BATCH_SIZE(default)\n"
-        "1: SUM\n")
+        "1: SUM_OVER_BATCH_SIZE(default)\n"
+        "2: SUM\n")
     ("optimizer", po::value<int>()->default_value(0)->notifier([&] (const auto &v) { _optimizer_type = v; }),
       "Optimizer type\n"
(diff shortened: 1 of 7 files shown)


tests/tools/onert_train/src/args.cc
```diff
@@ -230,9 +230,8 @@ void Args::Initialize(void)
         "1: CATEGORICAL_CROSSENTROPY\n")
     ("loss_reduction_type", po::value<int>()->default_value(0)->notifier([&] (const auto &v) { _loss_reduction_type = v; }),
         "Loss Reduction type\n"
-        "0: AUTO (default)\n"
-        "1: SUM_OVER_BATCH_SIZE\n"
-        "2: SUM\n")
+        "0: SUM_OVER_BATCH_SIZE(default)\n"
+        "1: SUM\n")
```
zetwhite, line 234: (warning) 

Here, `onert_train` argument is modified. 
This could affect developer's local test-bed.

YongseopKim (dismissed):
LGTM 👍 

I'm worried one thing - the enums value match.
- The `NNFW_TRAIN_LOSS_REDUCTION` enum in `runtime/onert/api/include/nnfw_experimental.h` 
- onert_train's arg `loss_reduction_type` 
- would confuse somewhat.

We must not forget to handle them carefully.

(Just IMO, you can ignore.)

zetwhite:
> I'm worried one thing - the enums value match.

Ah, That's right!  I missed that point, thank you a lot for your careful look. 

My minor worry is that It is fine to update `nnfw_experimental.h` values. :thinking: 
Because It could break some client codes that depend on this enum value.

IMO, we could update nnfw_experiment.h for now.  
  * Because there is no client of this training feature except `onert_train`. 
  * Also, I already had a history of updating `nnfw_experimental.h` enum value - https://github.com/Samsung/ONE/pull/12391/files 

So, Let's just update enum value to avoid value miss-match :smile:

YongseopKim:
@zetwhite oh. You don't have to consider my opinion right now. You can take your time enough and consider it later. 😄

YongseopKim (approved):
LGTM 💯

tests/tools/onert_train/src/args.cc
jyoungyun, line 234:
```suggestion
        "1: SUM_OVER_BATCH_SIZE(default)\n"
        "2: SUM\n")
```

tests/tools/onert_train/src/args.cc
```diff
@@ -230,9 +230,8 @@ void Args::Initialize(void)
         "1: CATEGORICAL_CROSSENTROPY\n")
     ("loss_reduction_type", po::value<int>()->default_value(0)->notifier([&] (const auto &v) { _loss_reduction_type = v; }),
```
jyoungyun, line 231:
```suggestion
    ("loss_reduction_type", po::value<int>()->default_value(1)->notifier([&] (const auto &v) { _loss_reduction_type = v; }),
```

tests/tools/onert_train/src/onert_train.cc
```diff
@@ -138,10 +138,8 @@ int main(const int argc, char **argv)
       switch (type)
       {
         case 0:
```
jyoungyun, line 140:
```suggestion
        case 1:
```

tests/tools/onert_train/src/onert_train.cc
```diff
@@ -138,10 +138,8 @@ int main(const int argc, char **argv)
...
-          return NNFW_TRAIN_LOSS_REDUCTION_AUTO;
-        case 1:
           return NNFW_TRAIN_LOSS_REDUCTION_SUM_OVER_BATCH_SIZE;
-        case 2:
+        case 1:
```
jyoungyun, line 142:
```suggestion
        case 2:
```

jyoungyun (commented):
How about this one? :)

zetwhite:
> How about this one? :)

No problem :)  

Just for curiosity, May I ask the reason for reverting the args value?
Maybe not change onert_train's behavior? or to avoid value mismatch between nnfw_experiment and onert_train (mentioned here - https://github.com/Samsung/ONE/pull/12419#pullrequestreview-1808847922)

jyoungyun:
> Maybe not change onert_train's behavior? or to avoid value mismatch between nnfw_experiment and onert_train 

Changing the value can satisfy both issues even though it does not match with other parameter's in `onert_train`.

jyoungyun (approved):
LGTM

hseok-oh:
@zetwhite Please resolve conflicts

jyoungyun (approved):
LGTM

YongseopKim (approved):
LGTM 🥇

PR #12424, opened 2024-01-09
[Draft] Use std::string instead of char pointer

jyoungyun:
This commit uses std::string instead of char pointer. In c++ area,
std::string is more popular way than char pointer to manage character
array.


runtime/onert/frontend/circle_traininfo/include/traininfo_loader.h
```diff
@@ -28,7 +28,7 @@ namespace traininfo_loader
 {
 
 // TODO change this line to use inline variable after C++17
-extern const char *const TRAININFO_METADATA_NAME;
+extern const std::string TRAININFO_METADATA_NAME;
```
glistening, line 31: @jyoungyun What is the benefit of using `std::string`? `char*` is traditional way for constant string. I am not sure it is safe and better to declare and use `std::string` for this case. I would keep the traditional way since there is no problem and is proven way.

glistening: Rather, sometimes it is not safe to define an instance of class because there may be dependency between classes. Initialization order is complicated.

jyoungyun: In C++, we usually use `std::string`. I didn't force to use it and it is just suggestion.
I have already written down why I suggest this one in the related issue. 

> Rather, sometimes it is not safe to define an instance of class because there may be dependency between classes. Initialization order is complicated.

And it is not related to this PR's suggestion.

glistening: No. My comment is about possible side-effect. Why it is not recommended to define an instance in static level. It is not the reason that `std::string` is popular in c++. We don't manipulate the string. We just need a plain constant data. I don't find the clear reason to change in this way.

> I have already written down why I suggest this one in the related issue.

Could you please leave the link to issue in this PR?

jyoungyun: I don't understand what kinds of possible side-effect you said there could be. However, I found the `std::string` has a second copy of the string to place it on the heap. And in c++ 11, the `const char*` is a suitable type for a string-literal. When c++17 comes out, we can also use `std::string_view` instead of `const char*`.

For the above mentioned reason, I close this PR.

> Could you please leave the link to issue in this PR?

You can find the reason why I suggested this PR [here](https://github.com/Samsung/ONE/pull/12408#discussion_r1445520929).

glistening: For my concern, please find https://stackoverflow.com/questions/12452707/why-static-and-global-variables-of-class-type-are-dangerous. It is likely that it is safe to define `std::string` in global variables though I don't investigated and will not. If it turned out it is safe, I don't object this change. But as I wrote, I don't find the benefit because the traditional style is more readable for me.

Issue #11744, opened 2023-10-20
[onert-micro] restore DC score


chunseoklee:
DC in SAM Score is dropped to 3.86 (Note that it was 4.18 at onert-micro 1.0.0 release)

[DC Score link](https://analysishub.sec.samsung.net/service/analyses/2432084) : Note that this link will be outdated soon. Any way, At this point, DC score is below 4 (3.86). Thus, we cannot release onert-micro with this score.

Let's restore this.

chunseoklee:
<body link="#0563C1" vlink="#954F72">



No. | Metric | BlockKey | Path | LineStart | LineEnd | LineCount | Token | Violation
-- | -- | -- | -- | -- | -- | -- | -- | --
1 | DC | 1 | onert-micro/luci-interpreter/src/kernels/AveragePool2D.cpp | 42 | 116 | 75 | 469 | Y
2 | DC | 1 | onert-micro/luci-interpreter/src/kernels/MaxPool2D.cpp | 53 | 125 | 73 | 469 | Y
3 | DC | 7 | onert-micro/luci-interpreter/src/kernels/Split.cpp | 42 | 93 | 52 | 243 | Y
4 | DC | 7 | onert-micro/luci-interpreter/src/kernels/SplitV.cpp | 41 | 92 | 52 | 243 | Y
5 | DC | 2 | onert-micro/luci-interpreter/src/kernels/SVDF.cpp | 37 | 82 | 46 | 346 | Y
6 | DC | 3 | onert-micro/luci-interpreter/src/kernels/FloorDiv.cpp | 42 | 87 | 46 | 296 | Y
7 | DC | 3 | onert-micro/luci-interpreter/src/kernels/FloorMod.cpp | 42 | 87 | 46 | 296 | Y
8 | DC | 4 | onert-micro/luci-interpreter/src/kernels/Maximum.cpp | 37 | 78 | 42 | 278 | Y
9 | DC | 4 | onert-micro/luci-interpreter/src/kernels/Minimum.cpp | 37 | 78 | 42 | 278 | Y
10 | DC | 6 | onert-micro/luci-interpreter/pal/common/PALSub.h | 38 | 77 | 40 | 245 | Y
11 | DC | 6 | onert-micro/luci-interpreter/pal/common/PALAddCommon.h | 43 | 82 | 40 | 245 | Y
[...]

PR #12428, opened 2024-01-09
[onert-micro] Reduce duplicate code in pool layers

ragmani:
This commit reduces duplicate code in pool layers.
  - Introduce Pool2DCommon.h that has common functions for pool layers.
    - configure_kernel_CirclePool2DCommon : unifies configure function of pool layers.
    - createPoolParams : unifies creating parameters of pool layers.
  - Apply common function for pool layers.

---

For #11744

### DIFF
--- a/onert-micro/luci-interpreter/src/kernels/Pool2DCommon.h
+++ b/onert-micro/luci-interpreter/src/kernels/Pool2DCommon.h
@@ -54,16 +54,10 @@ inline void configure_kernel_CirclePool2DCommon(const circle::Operator *cur_op,
 }
 
 inline luci_interpreter_pal::PoolParams createPoolParams(const circle::Operator *cur_op,
-                                                         BaseRuntimeGraph *runtime_graph)
+                                                         const kernels::SISOKernel &siso_kernel)
 {
-  const auto input_index = cur_op->inputs()->operator[](0);
-  const auto output_index = cur_op->outputs()->operator[](0);
-
-  assert(input_index != -1);
-  assert(output_index != -1);
-
-  const auto input = runtime_graph->getCircleTensorByIndex(input_index);
-  auto output = runtime_graph->getCircleTensorByIndex(output_index);
+  const auto input = siso_kernel.input();
+  const auto output = siso_kernel.output();
 
   const auto *options = cur_op->builtin_options_as_Pool2DOptions();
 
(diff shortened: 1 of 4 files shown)


onert-micro/luci-interpreter/src/kernels/Pool2DCommon.h
```diff
@@ -0,0 +1,128 @@
...
+    LUCI_INTERPRETER_CHECK(
+      std::abs(Tensor::scale(siso_kernel.output()) - Tensor::scale(siso_kernel.input())) <= 1.0e-6);
+    LUCI_INTERPRETER_CHECK(Tensor::zero_point(siso_kernel.input()) == 0 &&
+                           Tensor::zero_point(siso_kernel.output()) == 0);
+  }
+#endif // DIS_QUANT
+}
+
+inline luci_interpreter_pal::PoolParams createPoolParams(const circle::Operator *cur_op,
+                                                         BaseRuntimeGraph *runtime_graph)
```
BalyshevArtem, line 57: I think it's better to pass `const kernels::SISOKernel &siso_kernel` parameter here so as not to read the input and output tensor twice, since this has already been done in the calling function.
What do you think?)

BalyshevArtem (commented):
Everything apart from one small comment looks great, thanks)

onert-micro/luci-interpreter/src/kernels/Pool2DCommon.h
ragmani: Do you mean to pass `const kernels::SISOKernel &siso_kernel` instead of `BaseRuntimeGraph *runtime_graph`? It looks better.

onert-micro/luci-interpreter/src/kernels/Pool2DCommon.h
```diff
@@ -0,0 +1,128 @@
...
+}
+
+inline luci_interpreter_pal::PoolParams createPoolParams(const circle::Operator *cur_op,
+                                                         const kernels::SISOKernel &siso_kernel)
+{
+  const auto input_index = cur_op->inputs()->operator[](0);
+  const auto output_index = cur_op->outputs()->operator[](0);
+
+  assert(input_index != -1);
+  assert(output_index != -1);
```
BalyshevArtem, line 63: so, now input_index and output_index can be removed, I think

BalyshevArtem (approved):
LGTM, thank you!

BalyshevArtem:
@chunseoklee, PTAL)

chunseoklee (approved):

PR #12433, opened 2024-01-10
[onert] Change metadata pointer type to ExternalData

ragmani:
This commit changes metadata pointer type to ExternalData.

### DIFF
--- a/runtime/onert/core/include/ir/Model.h
+++ b/runtime/onert/core/include/ir/Model.h
@@ -177,7 +177,7 @@ private:
   std::shared_ptr<backend::custom::IKernelBuilder> _kernel_builder;
 
 public:
-  void add_metadata(const std::string &name, std::unique_ptr<const ir::Data> data)
+  void add_metadata(const std::string &name, std::unique_ptr<const ir::ExternalData> data)
   {
     _metadatas.emplace(name, std::move(data));
   }
@@ -187,8 +187,10 @@ public:
     return _metadatas.find(name) != _metadatas.end();
   }
 
+  bool exists_metadata() const { return _metadatas.size() != 0; }
+
   // NOTE The corresponding metadata is deleted from the model and returned
-  std::unique_ptr<const ir::Data> extract_metadata(const std::string name)
+  std::unique_ptr<const ir::ExternalData> extract_metadata(const std::string name)
   {
     auto m = _metadatas.find(name);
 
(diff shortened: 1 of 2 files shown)


zetwhite:
Thanks for trying :smile: 

I think we can use `ExternalData` for now and change it to use `ir::Data` (include `CachedData`) later when we need it. 
I put my detailed opinion here - https://github.com/Samsung/ONE/pull/12408#discussion_r1446889319.

ragmani:
I have a question regardless to this PR.
Is `model` going to have `metadata` as only external data(i.g. `ir::MMapedData`) in the future?

_Originally posted by @ragmani in https://github.com/Samsung/ONE/pull/12408#discussion_r1446861799_

ragmani:
Let me clarify the word first, to make sure that we have the same understanding. 

* `ir::CachedData` : holds **copy** of original data
* `ir::ExternalData` : points original data, we have to make sure that the original data is not lost.

_Originally posted by @zetwhite in https://github.com/Samsung/ONE/pull/12408#discussion_r1446885827_

ragmani:
> Is model going to have metadata as only external data(i.g. ir::MMapedData) in the future? 

@ragmani 
For now, only `TrainingInfo` is passed through Model's Metadata.
Since TrainingInfo is extracted when **loaded** time, we could use `ExternalData`.
If Metadata is used(extracted) at another time, I'm not sure that the original buffer (mmaped data) is not damaged. 

So, About your question, For now, Yes only external data is used in metadata.
About the future, I'm not 100% sure.

_Originally posted by @zetwhite in https://github.com/Samsung/ONE/pull/12408#discussion_r1446889319_

ragmani:
> * `ir::ExternalData` : points original data, we have to make sure that the original data is not lost.

@zetwhite 
Can we make sure if original data(external data) is not lost?

ragmani:
As talked offline with @zetwhite,
1. We confirmed that mmaped original data(external data) is not lost unless we call `munmap()`. So `metadata` will not be danging pointer in the current code.
2. @zetwhite is going to resolve the problem of having metadata information twice in memory.(i.g. `TrainingInfo` and mmaped metadata for `TrainingInfo`)

For those reasons, I close this PR.

zetwhite:
> For those reasons, I close this PR. Sorry for interrupting you due to my misunderstanding.

No problem at all. 
I appreciate your interest and by this chance, I could rethink about overall direction:)

Issue #12429, opened 2024-01-09
[luci] Applying an optimization pass before the shape inference


mhs4670go:
## What

![image](https://github.com/Samsung/ONE/assets/5449554/85c86059-3bcd-42bb-9f73-29b61f748523)

Above graph is from #12046. The issue is for introducing a constant folding pass related its operators: Shape/StridedSlice/Pack. The thing is that optimization passes run **only after applying shape and data type inference**.

https://github.com/Samsung/ONE/blob/3dc37ed1ea7a8c92afa3a3b7126d3b3d6c72836c/compiler/luci/service/src/CircleShapeInferenceRule.cpp#L1610

But, the shape of `TransposeConv` in the graph **can't be inferred** because its first arity, `inputSizes`, **is not constant**.

To resolve this issue, there are two things that I come up with as of now.

1. Support non-const `inputSizes`
2. Run this constant folding pass before the shape inference

Former option seems to be difficult because we have no related infrastructure as of now. So, second one looks okay where the pass assumes only valid shape would be used.

@seanshpark @jinevening Could you take a look at this issue?

seanshpark:
@mhs4670go , can we close this issue?

PR #12435, opened 2024-01-10
[luci] Use own shape in the shape inference

mhs4670go:
This commit adds a logic where the shape inference of CircleTransposeConv uses a node's own shape when one of inputs is not constant.

Related: https://github.com/Samsung/ONE/issues/12429#issuecomment-1884092665
Draft: #12434


seanshpark (approved):
LGTM

PR #12443, opened 2024-01-11
[tools/onert_train] Add use-model-parameter to argument

zetwhite:
This PR adds use-model-parameter to onert_train::args.

### DIFF
--- a/tests/tools/onert_train/src/args.cc
+++ b/tests/tools/onert_train/src/args.cc
@@ -222,18 +222,24 @@ void Args::Initialize(void)
     )
     ("mem_poll,m", po::value<bool>()->default_value(false)->notifier([&](const auto &v) { _mem_poll = v; }), "Check memory polling (default: false)")
     ("epoch", po::value<int>()->default_value(5)->notifier([&](const auto &v) { _epoch = v; }), "Epoch number (default: 5)")
-    ("batch_size", po::value<int>()->default_value(32)->notifier([&](const auto &v) { _batch_size = v; }), "Batch size (default: 32)")
-    ("learning_rate", po::value<float>()->default_value(0.001)->notifier([&](const auto &v) { _learning_rate = v; }), "Learning rate (default: 0.001)")
+    // TODO make model file parameter to be default, after circle+ file generating tool puslished
+    ("batch_size", po::value<int>()->default_value(32)->notifier([&](const auto &v) { _batch_size = v; }), 
+        "Batch size (default: 32), If 0 is given, use model parameter")
+    ("learning_rate", po::value<float>()->default_value(0.001)->notifier([&](const auto &v) { _learning_rate = v; }), 
+        "Learning rate (default: 0.001), If 0.0 is given, use model file parameter")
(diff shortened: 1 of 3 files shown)


tests/tools/onert_train/src/args.cc
```diff
@@ -222,18 +222,24 @@ void Args::Initialize(void)
     )
     ("mem_poll,m", po::value<bool>()->default_value(false)->notifier([&](const auto &v) { _mem_poll = v; }), "Check memory polling (default: false)")
     ("epoch", po::value<int>()->default_value(5)->notifier([&](const auto &v) { _epoch = v; }), "Epoch number (default: 5)")
-    ("batch_size", po::value<int>()->default_value(32)->notifier([&](const auto &v) { _batch_size = v; }), "Batch size (default: 32)")
-    ("learning_rate", po::value<float>()->default_value(0.001)->notifier([&](const auto &v) { _learning_rate = v; }), "Learning rate (default: 0.001)")
+    // TODO make model file parameter to be default, after circle+ file generating tool puslished
```
zetwhite, line 225:
```suggestion
    // TODO make model file parameter be default, after training info include file generating tool introduced 
```

tests/tools/onert_train/src/args.cc
```diff
@@ -222,18 +222,24 @@ void Args::Initialize(void)
     )
     ("mem_poll,m", po::value<bool>()->default_value(false)->notifier([&](const auto &v) { _mem_poll = v; }), "Check memory polling (default: false)")
     ("epoch", po::value<int>()->default_value(5)->notifier([&](const auto &v) { _epoch = v; }), "Epoch number (default: 5)")
-    ("batch_size", po::value<int>()->default_value(32)->notifier([&](const auto &v) { _batch_size = v; }), "Batch size (default: 32)")
-    ("learning_rate", po::value<float>()->default_value(0.001)->notifier([&](const auto &v) { _learning_rate = v; }), "Learning rate (default: 0.001)")
+    // TODO make model file parameter default, after training info include file generating tool introduced 
+    ("batch_size", po::value<int>()->default_value(32)->notifier([&](const auto &v) { _batch_size = v; }),
```
jyoungyun, line 226: What about setting the 0 as a default value to use the model parameter when the user does not give this parameter?

jyoungyun: If the user provides a circle+ model with some training parameters included in the model without any specific `onert_train` parameter, the train parameters of the given model may be automatically overwritten because the `onert_train` has its own default value. This can cause confusion for the user.

And if the user gives the circle model, not circle+ model, what about throwing an error to force the user to set the proper train parameter?

zetwhite: I see your point.
As I understood, your intention is "Let's make model file parameter default".   
I agree with your point. 

I first thought that we had to sustain the command line default value, 
Because there is NO 'circle+ generating tools' in public GitHub.
Since public users can't make circle+ file easily, I thought removing the default value could cause some discomfort. 

But, actually, There are no users for onert_train except developers :) 
So, Let's just remove it for quick development!

PR #12445, opened 2024-01-11
Draft: Remove default value from onert_train 

zetwhite:
This draft tries
  * remove the default option for train_info
     * to avoid always overwriting model file train_info
  * update `onert_train/args` use NNFW_TRAIN_...  directly
     * to avoid additional parsing (int to NNFW_TRAIN_..) in onert_train main

### DIFF
--- a/infra/cmake/packages/BoostConfig.cmake
+++ b/infra/cmake/packages/BoostConfig.cmake
@@ -60,7 +60,7 @@ endfunction(_Boost_Build)
 if (NOT BUILD_BOOST)
   # BoostConfig.cmake does not honor QUIET argument at least till cmake 1.70.0.
   # Thus, don't try to find_package if you're not entirely sure you have boost.
-  find_package(Boost 1.58.0 QUIET COMPONENTS log program_options filesystem system)
+  find_package(Boost 1.84.0 QUIET COMPONENTS log program_options filesystem system)
   if(Boost_FOUND)
     return()
   endif()
@@ -84,5 +84,5 @@ if(BUILD_BOOST)
   set(Boost_USE_STATIC_LIBS ON)
 
   # We built boost library so update Boost variables.
-  find_package(Boost 1.58.0 QUIET COMPONENTS log program_options filesystem system)
+  find_package(Boost 1.84.0 QUIET COMPONENTS log program_options filesystem system)
 endif(BUILD_BOOST)
--- a/tests/tools/onert_train/CMakeLists.txt
+++ b/tests/tools/onert_train/CMakeLists.txt
@@ -6,6 +6,10 @@ if(NOT BUILD_ONERT)
   return()
 endif(NOT BUILD_ONERT)
 
+# Use C++17 for onert_train
+# TODO Remove this when we use C++17 for all runtime directories
+set(CMAKE_CXX_STANDARD 17)
+
 list(APPEND ONERT_TRAIN_SRCS "src/onert_train.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/args.cc")
 list(APPEND ONERT_TRAIN_SRCS "src/nnfw_util.cc")
--- a/tests/tools/onert_train/src/args.cc
+++ b/tests/tools/onert_train/src/args.cc
@@ -15,6 +15,7 @@
  */
 
 #include "args.h"
+#include "nnfw_util.h"
 
 #include <functional>
 #include <iostream>
@@ -199,6 +200,40 @@ void Args::Initialize(void)
     }
   };
 
+  auto check_range = [](const std::initializer_list<int> arg_list, const int value,
+                        const std::string &arg_name = "") {
+    for (const auto &i : arg_list)
+    {
+      if (i == value)
+        return;
+    }
+    throw std::runtime_error{arg_name + " " + std::to_string(value) + " is unsupported argument"};
+  };
+
+  auto loss_help_msg = [](const std::initializer_list<int> arg_list) -> std::string {
+    std::string res = "Loss type (default: use model file parameter)\n";
+    for (const auto &i : arg_list)
+      res += std::to_string(i) + ": " + toString(static_cast<NNFW_TRAIN_LOSS>(i)) + "\n";
+    res.erase(res.length() - 1); // remove last \n
+    return res;
+  };
+
+  auto loss_reduction_type_help_msg = [](const std::initializer_list<int> arg_list) -> std::string {
+    std::string res = "Loss Reduction type (default: use model file parameter)\n";
+    for (const auto &i : arg_list)
+      res += std::to_string(i) + ": " + toString(static_cast<NNFW_TRAIN_LOSS_REDUCTION>(i)) + "\n";
+    res.erase(res.length() - 1); // remove last \n
+    return res;
+  };
+
+  auto optimizer_help_msg = [](const std::initializer_list<int> arg_list) -> std::string {
+    std::string res = "Optimizer (default: use model file parameter)\n";
+    for (const auto &i : arg_list)
+      res += std::to_string(i) + ": " + toString(static_cast<NNFW_TRAIN_OPTIMIZER>(i)) + "\n";
+    res.erase(res.length() - 1); // remove last \n
+    return res;
+  };
+
   // General options
   po::options_description general("General options", 100);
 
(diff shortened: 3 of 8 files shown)


tests/tools/onert_train/src/onert_train.cc
```diff
@@ -121,45 +121,6 @@ int main(const int argc, char **argv)
...
-      {
-        case 1:
-          return NNFW_TRAIN_OPTIMIZER_SGD;
-        case 2:
-          return NNFW_TRAIN_OPTIMIZER_ADAM;
-        default:
-          std::cerr << "E: not supported optimizer type" << std::endl;
-          exit(-1);
-      }
-    };
```
zetwhite, line 161: ## Motivation

I'd like to remove this part from `onert_train`. 
Because this part highly depends on `onert_train/args.h `. 
   * Now, If we update argument list in onert_train/args.h, we have to update this onert_train main function too.

tests/tools/onert_train/src/nnfw_util.cc
```diff
@@ -49,63 +50,50 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
...
+  return name_map.at(loss);
+}
+
+std::string toString(const NNFW_TRAIN_LOSS_REDUCTION &loss_rdt)
+{
+  static const std::unordered_map<NNFW_TRAIN_LOSS_REDUCTION, std::string> name_map{
+    {NNFW_TRAIN_LOSS_REDUCTION_UNDEFINED, "undefined"},
+    {NNFW_TRAIN_LOSS_REDUCTION_SUM_OVER_BATCH_SIZE, "sum over batch size"},
+    {NNFW_TRAIN_LOSS_REDUCTION_SUM, "sum"}};
+  return name_map.at(loss_rdt);
```
zetwhite, line 79: (notes)

tests/tools/onert_train/src/args.cc
```diff
@@ -199,6 +200,17 @@ void Args::Initialize(void)
     }
   };
 
+  auto loss_help_msg = []() -> std::string {
+    std::string res = "Loss type\n";
+
+    constexpr int printed_list[2] = {1, 2};
+    for (auto &i : printed_list)
+      res += std::to_string(i) + ": " + toString(static_cast<NNFW_TRAIN_LOSS>(i)) + "\n";
```
zetwhite, line 208: ### Motivation

For the help message, Let's reuse the related string defined in nnfw_utils.h
https://github.com/Samsung/ONE/pull/12445/files#r1448423248

tests/tools/onert_train/src/types.h
```diff
@@ -17,11 +17,58 @@
...
 
+#include <type_traits>
+#include <stdexcept>
+
 namespace onert_train
 {
 
 using TensorShape = std::vector<int>;
 
+template <class T, class Enable = void> struct Optional
```
hseok-oh, line 28: IMO, we can use c++17 for `onert_train`. Otherwise, we can use `boost::optional`.

tests/tools/onert_train/src/nnfw_util.h
```diff
@@ -36,6 +36,10 @@ namespace onert_train
 uint64_t num_elems(const nnfw_tensorinfo *ti);
 uint64_t bufsize_for(const nnfw_tensorinfo *ti);
 
+std::string toString(const NNFW_TRAIN_OPTIMIZER &opt);
+std::string toString(const NNFW_TRAIN_LOSS &loss);
+std::string toString(const NNFW_TRAIN_LOSS_REDUCTION &loss_rdt);
 std::ostream &operator<<(std::ostream &os, const nnfw_train_info &info);
+
```
zetwhite, line 43:
```suggestion
```

tests/tools/onert_train/src/types.h
zetwhite: Thanks for giving the idea.
For this draft, Let me try to use c++17.

tests/tools/onert_train/CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ if(NOT BUILD_ONERT)
   return()
 endif(NOT BUILD_ONERT)
 
+# Use C++17 for onert_train
+# TODO Removethis when we use C++17 for all runtime directories
+set(CMAKE_CXX_STANDARD 17)
```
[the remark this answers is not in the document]
zetwhite: It causes trouble with boost. 

```
android.release/overlay/include/boost/smart_ptr/detail/shared_count.hpp:395:33: error: no template named 'auto_ptr' in namespace 'std'

[2024-01-15T06:13:13.272Z]     explicit shared_count( std::auto_ptr<Y> & r ): pi_( new sp_counted_impl_p<Y>( r.get() ) )

[2024-01-15T06:13:13.272Z]                            
```

tests/tools/onert_train/CMakeLists.txt
```diff
@@ -33,6 +37,8 @@ endif(HDF5_FOUND)
 
 target_include_directories(onert_train PRIVATE src)
 target_include_directories(onert_train PRIVATE ${Boost_INCLUDE_DIRS})
+# To compile boost with c++17
```
zetwhite, line 40: To remove auto_ptr in boost which is not compatible with c++17 
```suggestion
# To remove auto_ptr in boost which is not compatible with c++17  
```

tests/tools/onert_train/src/args.h
```diff
@@ -81,11 +89,11 @@ class Args
...
-  int _batch_size;
-  float _learning_rate;
-  int _loss_type;
-  int _loss_reduction_type;
-  int _optimizer_type;
+  std::optional<int> _batch_size;
+  std::optional<float> _learning_rate;
+  std::optional<NNFW_TRAIN_LOSS> _loss_type;
+  std::optional<NNFW_TRAIN_LOSS_REDUCTION> _loss_reduction_type;
+  std::optional<NNFW_TRAIN_OPTIMIZER> _optimizer_type;
```
zetwhite, line 96: @jyoungyun 

I'd like to hear your opinion about this change.

#### AS-IS
* `onert_train/args.h` holds integer value and parse it to enum `NNFW_TRAIN_*` in `onert_train`'s main function.

#### TO-BE
* `onert_train/args.h` directly holds enum `NNFW_TRAIN_*`.
* Since it holds an enum value, to distinguish the not-given value use `optional` class.
 

Since boost has a problem with c++17, I'm trying to find another way,  But before going further, I'd like to hear your opinion about the overall direction.

tests/tools/onert_train/src/args.cc
```diff
@@ -222,20 +257,17 @@ void Args::Initialize(void)
...
-        "1: SUM_OVER_BATCH_SIZE(default)\n"
-        "2: SUM")
-    ("optimizer", po::value<int>()->default_value(1)->notifier([&] (const auto &v) { _optimizer_type = v; }),
-      "Optimizer type\n"
-      "1: SGD (default)\n"
-      "2: Adam")
+    ("batch_size", po::value<int>()->notifier([&](const auto &v) { _batch_size = v; }), "Batch size (default: use model file parameter)\n")
+    ("learning_rate", po::value<float>()->notifier([&](const auto &v) { _learning_rate = v; }), "Learning rate (default: use model file parameter)") 
+    ("loss",
+        po::value<int>()->notifier([&] (const auto &v) { check_range({1, 2}, v, "loss"); _loss_type = static_cast<NNFW_TRAIN_LOSS>(v); }),
```
zetwhite, line 263: here `check_range` assure that valid integer is always passed!

tests/tools/onert_train/src/nnfw_util.cc
```diff
@@ -49,63 +50,50 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
...
 }
 
+std::string toString(const NNFW_TRAIN_OPTIMIZER &opt)
+{
+  static const std::unordered_map<NNFW_TRAIN_OPTIMIZER, std::string> name_map{
+    {NNFW_TRAIN_OPTIMIZER_UNDEFINED, "undefined"},
+    {NNFW_TRAIN_OPTIMIZER_ADAM, "adam"},
+    {NNFW_TRAIN_OPTIMIZER_SGD, "sgd"},
+  };
+  return name_map.at(opt);
```
jyoungyun, line 60: (Q) How does it work if it receives a value that is not defined in enum in the `toString` method?

tests/tools/onert_train/src/args.cc
```diff
@@ -199,6 +200,40 @@ void Args::Initialize(void)
...
+    return res;
+  };
+
+  auto optimizer_help_msg = [](const std::initializer_list<int> arg_list) -> std::string {
+    std::string res = "Optimizer (default: use model file parameter)\n";
+    for (const auto &i : arg_list)
+      res += std::to_string(i) + ": " + toString(static_cast<NNFW_TRAIN_OPTIMIZER>(i)) + "\n";
+    res.erase(res.length() - 1); // remove last \n
+    return res;
+  };
```
zetwhite, line 235: .

tests/tools/onert_train/src/nnfw_util.cc
zetwhite: It might throw an error in line `name_map.at()` line.

tests/tools/onert_train/src/args.cc
```diff
@@ -199,6 +200,40 @@ void Args::Initialize(void)
...
   };
 
+  auto check_range = [](const std::initializer_list<int> arg_list, const int value,
+                        const std::string &arg_name = "") {
+    for (const auto &i : arg_list)
+    {
+      if (i == value)
+        return;
+    }
+    throw std::runtime_error{arg_name + " " + std::to_string(value) + " is unsupported argument"};
```
jyoungyun, line 210: IMO, it would be better to print error message instead of throwing an error.

```
E: Fail to run by runtime error:loss 5 is unsupported argument
```

zetwhite:
@nnfw-bot test onert-android-release

tests/tools/onert_train/src/args.cc
```diff
@@ -222,20 +257,17 @@ void Args::Initialize(void)
...
-        "2: CATEGORICAL_CROSSENTROPY")
-    ("loss_reduction_type", po::value<int>()->default_value(1)->notifier([&] (const auto &v) { _loss_reduction_type = v; }),
-        "Loss Reduction type\n"
-        "1: SUM_OVER_BATCH_SIZE(default)\n"
-        "2: SUM")
-    ("optimizer", po::value<int>()->default_value(1)->notifier([&] (const auto &v) { _optimizer_type = v; }),
-      "Optimizer type\n"
-      "1: SGD (default)\n"
-      "2: Adam")
+    ("batch_size", po::value<int>()->notifier([&](const auto &v) { _batch_size = v; }), "Batch size (default: use model file parameter)\n")
```
jyoungyun, line 260: (suggestion)
How about setting a default value for batch_size? Since the batch size is not related to the model file, so onert_train can provide a default batch size value to the user.

TensorFlow sets 32 as the default batch size, and PyTorch sets 1 as the default batch size. In ONERT, if the data length is not divisible by the batch size, it does not train the rest of the dataset. Therefore, I suggest setting 1 as the batch size for onert_train.

infra/cmake/packages/BoostConfig.cmake
```diff
@@ -60,7 +60,7 @@ endfunction(_Boost_Build)
 if (NOT BUILD_BOOST)
   # BoostConfig.cmake does not honor QUIET argument at least till cmake 1.70.0.
   # Thus, don't try to find_package if you're not entirely sure you have boost.
-  find_package(Boost 1.58.0 QUIET COMPONENTS log program_options filesystem system)
+  find_package(Boost 1.84.0 QUIET COMPONENTS log program_options filesystem system)
```
zetwhite, line 63: Boost 1.58.0 does not support C++17. 
Since `auto_ptr`is deprecated in c++17, if we compile with boost old version, it throws an error message like `auto_ptr` is not found in `std` namespace. 

https://www.boost.org/users/history/version_1_58_0.html#:~:text=Boost%27s%20primary%20test,4.4.7%2C%204.8.4%2C%204.9.2 

> Boost's primary test compilers are:
> Linux:
> Clang: 3.4
> Clang, C++14: 3.5, 3.6
> GCC: 4.4.7, 4.9.2
> GCC, C++98: 4.8.1, 4.8.2
> GCC, C++11: 4.4.7, 4.8.4, 4.


----------------------

So, Let's upgrade Boost version.

tests/tools/onert_train/src/args.cc
zetwhite: updated:) 
I make it to print an error message to std::cerr and end with `exit(1)`.  
Is it right that you intended?

tests/tools/onert_train/src/args.cc
zetwhite: I updated like this : [`f0e231d` (#12445)](https://github.com/Samsung/ONE/pull/12445/commits/f0e231d30761c640dd7b096ffd173c1d27cfea45) 

Since the onert_train main function has to know whether the value of `batch_size` is user-given or not, It is hard to remove `optional` here. 
I updated batch size decision logic in the onert_train main function. 

Is this way fine with you?

tests/tools/onert_train/src/args.cc
```diff
@@ -199,6 +200,41 @@ void Args::Initialize(void)
...
+    return res;
+  };
+
+  auto optimizer_help_msg = [](const std::initializer_list<int> arg_list) -> std::string {
+    std::string res = "Optimizer (default: use model file parameter)\n";
+    for (const auto &i : arg_list)
+      res += std::to_string(i) + ": " + toString(static_cast<NNFW_TRAIN_OPTIMIZER>(i)) + "\n";
+    res.erase(res.length() - 1); // remove last \n
+    return res;
+  };
```
zetwhite, line 236: (note) 

print help message - to show how to use `onert_train`

infra/cmake/packages/BoostConfig.cmake
ragmani: Just for convey information, `auto_ptr` is removed in c++17.

PR #12448, opened 2024-01-11
[cker] Introduce DepthwiseConv2D gradient kernel

jyoungyun:
This commit introduces DepthwiseConv2D gradient kernel. The depthwise_conv_op.h file came from TensorFlow code. This commit transforms the depthwise_conv related code to fit onert and execute this kernel function.

Draft: #12392

### DIFF
--- a/compute/cker/include/cker/eigen/depthwise_conv_op.h
+++ b/compute/cker/include/cker/eigen/depthwise_conv_op.h
@@ -1,0 +1,956 @@
+/*
+ * Copyright (c) 2024 Samsung Electronics Co., Ltd. All Rights Reserved
+ * Copyright 2015 The TensorFlow Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __NNFW_CKER_EIGEN_DEPTHWISE_CONV_OP_H__
+#define __NNFW_CKER_EIGEN_DEPTHWISE_CONV_OP_H__
+
+// From tensorflow/core/kernels/depthwise_conv_grad_op.cc
+#define EIGEN_USE_THREADS
+
+#include <thread>
+#include "unsupported/Eigen/CXX11/Tensor"
+#include "cker/operation/Helper/Tensor.h"
+
+// From tensorflow/core/kernels/depthwise_conv_op.h
+namespace nnfw
+{
+namespace cker
+{
+namespace depthwise_conv_op
+{
+
+template <typename Device, typename T> struct LaunchDepthwiseConvBackpropInputOp
+{
+  void operator()(int batch, int in_rows, int in_cols, int in_depth, int filter_rows,
+                  int filter_cols, int depth_multiplier, int stride, int pad_rows, int pad_cols,
+                  int out_rows, int out_cols, int out_depth, const T *out_backprop, const T *filter,
+                  T *in_backprop);
+};
+
+template <typename Device, typename T> struct LaunchDepthwiseConvBackpropFilterOp
+{
+  void operator()(int batch, int in_rows, int in_cols, int in_depth, int filter_rows,
+                  int filter_cols, int depth_multiplier, int stride, int pad_rows, int pad_cols,
+                  int out_rows, int out_cols, int out_depth, const T *out_backprop, const T *input,
+                  T *filter_backprop);
+};
+
+namespace functor
+{
+
+// Pads 'filter' to vector-register boundary along its inner dimension:
+//   filter_inner_dim_size = in_depth * depth_multiplier
+// Requires 'filter' to have the following storage order:
+//   [filter_rows, filter_cols, in_depth, depth_multiplier]
+// Returns zero-padded filter in 'padded_filter'.
+//
+// EX:
+//   in_depth = 3, depth_multiplier = 2, filter [2, 2], register_width = 4
+//   So we have a total of 3 * 2 = 6 filters, each of spatial size 2 x 2.
+//
+//   filter [rows, cols, in_depth, depth_multiplier]
+//     [u0, v0, w0, x0] [y0, z0, u1, v1] [w1, x1, y1, z1]
+//     [u2, v2, w2, x2] [y2, z2, u3, v3] [w3, x3, y3, z3]
+//
+//   padded_filter [rows, cols, in_depth, depth_multiplier]
+//     [u0, v0, w0, x0] [y0, z0, 0, 0] [u1, v1, w1, x1] [y1, z1, 0, 0]
+//     [u2, v2, w2, x2] [y2, z2, 0, 0] [u3, v3, w3, x3] [y3, z3, 0, 0]
+
+template <typename T> struct DepthwiseFilterPadOp
+{
+  void operator()(int, int, int, int, int filter_rows, int filter_cols, int, int, int, int, int,
+                  int, int out_depth, const T *filter, T *padded_filter)
+  {
+    typedef typename Eigen::internal::packet_traits<T>::type Packet;
+    static const int64_t kPacketSize = (sizeof(Packet) / sizeof(T));
+
+    // Calculate vectorized and scalar lengths of filter's inner dimension.
+    const int64_t filter_inner_dim_size = out_depth;
+    const int64_t vectorized_size = (filter_inner_dim_size / kPacketSize) * kPacketSize;
+    const int64_t scalar_size = filter_inner_dim_size - vectorized_size;
+    // Calculate required padding and padded output buffer stride.
+    const int64_t pad_size = scalar_size > 0 ? kPacketSize - scalar_size : 0;
+    const int64_t padded_filter_stride = vectorized_size + kPacketSize;
+
+    const int64_t filter_spatial_size = filter_rows * filter_cols;
+    for (int64_t i = 0; i < filter_spatial_size; ++i)
+    {
+      const int64_t input_base = i * filter_inner_dim_size;
+      const int64_t output_base = i * padded_filter_stride;
+      // Write vectorized length of filter's inner dimension to output.
+      for (int64_t j = 0; j < vectorized_size; j += kPacketSize)
+      {
+        const auto v = Eigen::internal::ploadu<Packet>(filter + input_base + j);
+        Eigen::internal::pstoreu<T>(padded_filter + output_base + j, v);
+      }
+      // Write scalar length of filter's inner dimension to output.
+      for (int64_t j = 0; j < scalar_size; ++j)
+      {
+        padded_filter[output_base + vectorized_size + j] = filter[input_base + vectorized_size + j];
+      }
+      // Pad the remainder of output to vector-register boundary.
+      for (int64_t j = 0; j < pad_size; ++j)
+      {
+        padded_filter[output_base + vectorized_size + scalar_size + j] = static_cast<T>(0);
+      }
+    }
+  }
+};
+
+// Copies data from local region in 'input' specified by 'out_r' and 'out_'c'
+// to 'input_buffer'. The copied data is replicated by factor
+// 'depth_multiplier', and padded to vector register-width boundaries so
+// that it is aligned for efficient traversal and vector multiply-add by the
+// depthwise kernel.
+//
+// EX:
+//   in_depth = 3, depth_multiplier = 2, filter [2, 2], register_width = 4
+//
+//   input: [batch, in_rows, in_cols, in_depth]
+//
+//     [a0, a1, a2, b0, b1, b2, ..., e0, e1, e2, f0, f1, f2, ...]
+//
+//   input_buffer (register boundaries shown):
+//     [a0, a0, a1, a1] [a2, a2, 0, 0]   in_row = 0, in_col = 0
+//     [b0, b0, b1, b1] [b2, b2, 0, 0]   in_row = 0, in_col = 1
+//     [e0, e0, e1, e1] [e2, e2, 0, 0]   in_row = 1, in_col = 0
+//     [f0, f0, f1, f1] [f2, f2, 0, 0]   in_row = 1, in_col = 1
+//
+// Returns replicated and padded data from specified input region in
+// 'input_buffer'.
+
+template <typename T> struct DepthwiseInputCopyOp
+{
+  void operator()(int, int in_rows, int in_cols, int in_depth, int filter_rows, int filter_cols,
+                  int depth_multiplier, int stride, int pad_rows, int pad_cols, int, int,
+                  int out_depth, const int64_t padded_filter_inner_dim_size, const int64_t out_r,
+                  const int64_t out_c, const T *input, T *input_buffer)
+  {
+    typedef typename Eigen::internal::packet_traits<T>::type Packet;
+    static const int64_t kPacketSize = Eigen::internal::packet_traits<T>::size;
+
+    const int64_t kDepth = depth_multiplier;
+    // Calculate vectorized and scalar (residual) lengths for 'in_depth'.
+    const int64_t input_vectorized_size = (in_depth / kPacketSize) * kPacketSize;
+    const int64_t input_scalar_size = in_depth - input_vectorized_size;
+
+    // Calculate output padding length.
+    const int64_t output_scalar_size = out_depth % kPacketSize;
+    const int64_t output_pad_size = output_scalar_size > 0 ? kPacketSize - output_scalar_size : 0;
+
+    // Iterate through all rows x cols reading 'in_depth' from 'input' and
+    // replicating by 'depth_multiplier' into 'input_buffer' (otherwise
+    // zero-padding input buffer as needed).
+    auto *in_buf = input_buffer;
+    const int64_t in_r_start = out_r * stride - pad_rows;
+    const int64_t in_c_start = out_c * stride - pad_cols;
+
+    // TODO: add a ploaddup variant for depth == 2 if needed.
+    if (kDepth > 1 && kDepth <= kPacketSize)
+    {
+      for (int64_t f_r = 0; f_r < filter_rows; ++f_r)
+      {
+        const int64_t in_r = in_r_start + f_r;
+
+        for (int64_t f_c = 0; f_c < filter_cols; ++f_c)
+        {
+          const int64_t in_c = in_c_start + f_c;
+
+          if (in_r >= 0 && in_r < in_rows && in_c >= 0 && in_c < in_cols)
+          {
+            const auto *in = input + (in_r * in_cols + in_c) * in_depth;
+            int64_t limit = in_depth;
+            // This will overwrite up to kPacketSize next elements,
+            // this is ok on all iterations except the last one, since
+            // we will write correct values on a next iteration.
+            if (f_c == filter_cols - 1)
+            {
+              limit -= (kPacketSize - kDepth) / kDepth + 1;
+              if (limit < 0)
+              {
+                limit = 0;
+              }
+            }
+            // Copy vectorized portion of inner dimension.
+            for (int64_t d = 0; d < limit; d++)
+            {
+              const auto p = Eigen::internal::pset1<Packet>(in[d]);
+              Eigen::internal::pstoreu<T>(in_buf, p);
+              in_buf += kDepth;
+            }
+
+            // Copy the scalar portion.
+            for (int64_t d = limit; d < in_depth; d++)
+            {
+              const auto value = in[d];
+              for (int64_t dm = 0; dm < kDepth; dm++)
+              {
+                in_buf[dm] = value;
+              }
+              in_buf += kDepth;
+            }
+
+            // Pad the remainder of the output to vector register boundary.
+            for (int64_t d = 0; d < output_pad_size; ++d)
+            {
+              in_buf[d] = static_cast<T>(0);
+            }
+            in_buf += output_pad_size;
+          }
+          else
+          {
+            // Zero pad.
+            memset(in_buf, 0, sizeof(T) * padded_filter_inner_dim_size);
+            in_buf += padded_filter_inner_dim_size;
+          }
+        }
+      }
+    }
+    else if (kDepth > kPacketSize)
+    {
+      // Calculate vectorized and scalar (residual) lengths for
+      // 'depth_multiplier'. This is used to efficiently replicate data for
+      // when 'depth_multiplier' > kPacketSize.
+      const int64_t dm_vectorized_size = (kDepth / kPacketSize) * kPacketSize;
+
+      for (int64_t f_r = 0; f_r < filter_rows; ++f_r)
+      {
+        const int64_t in_r = in_r_start + f_r;
+
+        for (int64_t f_c = 0; f_c < filter_cols; ++f_c)
+        {
+          const int64_t in_c = in_c_start + f_c;
+
+          if (in_r >= 0 && in_r < in_rows && in_c >= 0 && in_c < in_cols)
+          {
+            const auto *in = input + (in_r * in_cols + in_c) * in_depth;
+            // Copy vectorized portion of inner dimension.
+            for (int64_t d = 0; d < in_depth; d++)
+            {
+              const auto p = Eigen::internal::pset1<Packet>(in[d]);
+              for (int64_t dm = 0; dm < dm_vectorized_size; dm += kPacketSize)
+              {
+                Eigen::internal::pstoreu<T>(in_buf + dm, p);
+              }
+              // Overlapping store for the remainder.
+              Eigen::internal::pstoreu<T>(in_buf + kDepth - kPacketSize, p);
+              in_buf += kDepth;
+            }
+            // Pad the remainder of the output to vector register boundary.
+            for (int64_t d = 0; d < output_pad_size; ++d)
+            {
+              in_buf[d] = static_cast<T>(0);
+            }
+            in_buf += output_pad_size;
+          }
+          else
+          {
+            // Zero pad.
+            memset(in_buf, 0, sizeof(T) * padded_filter_inner_dim_size);
+            in_buf += padded_filter_inner_dim_size;
+          }
+        }
+      }
+    }
+    else if (kDepth == 1)
+    {
+      for (int64_t f_r = 0; f_r < filter_rows; ++f_r)
+      {
+        const int64_t in_r = in_r_start + f_r;
+
+        for (int64_t f_c = 0; f_c < filter_cols; ++f_c)
+        {
+          const int64_t in_c = in_c_start + f_c;
+
+          if (in_r >= 0 && in_r < in_rows && in_c >= 0 && in_c < in_cols)
+          {
+            const auto *in = input + (in_r * in_cols + in_c) * in_depth;
+            for (int64_t d = 0; d < input_vectorized_size; d += kPacketSize)
+            {
+              const auto p = Eigen::internal::ploadu<Packet>(in + d);
+              Eigen::internal::pstoreu<T>(in_buf, p);
+              in_buf += kPacketSize;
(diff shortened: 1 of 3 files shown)


YongseopKim (approved):
LGTM 👍 
And IMO the title is better if `[cker] Introduce ...` because these codes are not in onert.

glistening:
@jyoungyun I am waiting for the title change as @YongseopKim commented: `[onert]` to `[cker]`

jyoungyun:
Oh, I was missing the `[cker]` header. I will update it.

glistening (approved):
LGTM

compute/cker/include/cker/eigen/depthwise_conv_op.h
```diff
@@ -0,0 +1,956 @@
...
+//
+//   'buffer' (register boundaries shown):
+//
+//     [f00, f01, f10, f11] [f20, f21, 0, 0]   in_row = 0, in_col = 0
+//     [e00, e01, e10, e11] [e20, e21, 0, 0]   in_row = 0, in_col = 1
+//     [b00, b01, b10, b11] [b20, b21, 0, 0]   in_row = 1, in_col = 0
+//     [a00, a01, a10, a11] [a20, a21, 0, 0]   in_row = 1, in_col = 1
+//
+template <typename T>
+static void CopyOutputBackpropRegion(int, int, int, int, int filter_rows_, int filter_cols_, int,
```
ragmani, line 356: I guess there is no reason to use static keyword. Why did you use static keyword?
```suggestion
void CopyOutputBackpropRegion(int, int, int, int, int filter_rows_, int filter_cols_, int,
```

compute/cker/include/cker/eigen/depthwise_conv_op.h
```diff
@@ -0,0 +1,956 @@
...
+//                            ([b00, b01, b10, b11] x [u2, v2, w2, x2]) +
+//                            ([a00, a01, a10, a11] x [u3, v3, w3, x3])
+//
+//   Reduction step along depth-multiplier dimension:
+//
+//     [q00, q01, q10, q11] [q20, q21, 0, 0] -> [r0, r1, r2, 0]
+//
+
+template <typename T>
+static void ComputeBackpropInput(int, int, int in_cols, int in_depth_, int filter_rows,
```
ragmani, line 455: ditto

compute/cker/include/cker/eigen/depthwise_conv_op.h
```diff
@@ -0,0 +1,956 @@
...
+//
+template <typename T>
+static void CopyOutputBackpropRegion(int, int, int, int, int filter_rows_, int filter_cols_, int,
+                                     int stride_, int pad_rows_, int pad_cols_, int out_rows_,
+                                     int out_cols_, int out_depth,
+                                     const int64_t padded_filter_inner_dim_size, const int64_t in_r,
+                                     const int64_t in_c, const T *out_backprop, T *buffer)
+{
+  typedef typename Eigen::internal::packet_traits<T>::type Packet;
+  static const int64_t kPacketSize = (sizeof(Packet) / sizeof(T));
```
ragmani, line 363: How about using local variable instead of using static variable?
```suggestion
  const int64_t kPacketSize = (sizeof(Packet) / sizeof(T));
```

ragmani: Also, you can replace this with an inline namespace variable since C++17.

compute/cker/include/cker/train/operation/DepthwiseConv.h
```diff
@@ -0,0 +1,112 @@
...
+{
+
+class DepthwiseConv
+{
+public:
+  DepthwiseConv() = default;
+
+  template <typename T> int64_t kPacketSize() const
+  {
+    typedef typename Eigen::internal::packet_traits<T>::type Packet;
```
ragmani, line 38: (Optional) You can replace `typedef` with _alias declaration_
```suggestion
    using packet_t = typename Eigen::internal::packet_traits<T>::type;
```

compute/cker/src/train/DepthwiseConv.test.cc
```diff
@@ -0,0 +1,410 @@
...
+    calculateInputGradExpected(params, incoming_shape, incoming_data, filter_shape, filter_data,
+                               grad_shape, expected.data());
+
+    _dconv_kernel->backpropInput(params, incoming_shape, incoming_data, filter_shape, filter_data,
+                                 _padded_filter.data(), grad_shape, gradient.data(),
+                                 _use_padded_filter, _filter_buffers.data(),
+                                 _filter_dim_buffers.data());
+
+    for (size_t i = 0; i < gradient.size(); ++i)
+      EXPECT_NEAR(gradient[i], expected[i], 1e-3f);
```
ragmani, line 80: There is no checking `_filter_buffers` and `_filter_dim_buffers` even if they are updated. Could you add the verification to check them?

compute/cker/src/train/DepthwiseConv.test.cc
```diff
@@ -0,0 +1,410 @@
...
+
+    calculateFilterGradExpected(params, incoming_shape, incoming_data, input_shape, input_data,
+                                filter_grad_shape, expected.data());
+
+    _dconv_kernel->backpropFilter(params, incoming_shape, incoming_data, input_shape, input_data,
+                                  filter_grad_shape, gradient.data(), _padded_filter.data(),
+                                  _filter_buffers.data());
+
+    for (size_t i = 0; i < gradient.size(); ++i)
+      EXPECT_NEAR(gradient[i], expected[i], 1e-3f);
```
ragmani, line 112: Here too. There is no checking `_padded_filter` and `_filter_buffers`.

compute/cker/include/cker/eigen/depthwise_conv_op.h
```diff
@@ -0,0 +1,956 @@
...
+    const int64_t input_bytes = out_rows * out_cols * out_depth * sizeof(T);
+    const int64_t output_bytes = in_rows * in_cols * in_depth * sizeof(T);
+    const int64_t compute_cycles = in_rows * in_cols * out_depth * batch;
+    const Eigen::TensorOpCost cost(input_bytes, output_bytes, compute_cycles);
+    d.parallelFor(batch, cost, shard);
+  }
+};
+
+template <typename T>
+static void
```
ragmani, line 639:
```suggestion
void
```

compute/cker/include/cker/train/operation/DepthwiseConv.h
ragmani: I checked that this code was from tensorflow. You can ignore the above comment.

compute/cker/include/cker/eigen/depthwise_conv_op.h
ragmani: I checked that this code was from tensorflow. You can ignore the above comment.

compute/cker/include/cker/train/operation/DepthwiseConv.h
```diff
@@ -0,0 +1,112 @@
...
+{
+namespace train
+{
+
+class DepthwiseConv
+{
+public:
+  DepthwiseConv() = default;
+
+  template <typename T> int64_t kPacketSize() const
```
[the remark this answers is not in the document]
jyoungyun: @ragmani 
If `DepthwiseConv` class is a class template, then the `DepthwiseConvolutionLayer` in train backend should have each type of `DepthwiseConv` class. Is this a good approach? Since other things except for this function are almost the same, I only made this function as a template function.

compute/cker/src/train/DepthwiseConv.test.cc
```diff
@@ -0,0 +1,410 @@
...
+
+    _use_padded_filter = (out_depth % kPacketSize) == 0 ? false : true;
+    {
+      nnfw::cker::Shape padded_filter_shape(
+        {batch, filter_spatial_size, padded_filter_inner_dim_size});
+      _padded_filter.resize(padded_filter_shape.FlatSize());
+    }
+
+    {
+      const int thread_count = _dconv_kernel->getThreadCount() + 1;
```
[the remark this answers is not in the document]
jyoungyun: The meaning of `+1` is for main thread. The `numThreads()` in the eigen library only returns the thread count except for the main thread. This function can be executed in the main thread depending on the situation. So I added `1` for main thread.

compute/cker/include/cker/eigen/depthwise_conv_op.h
jyoungyun: Yes, this code came from TensorFlow. I want to keep the existing code. :)

compute/cker/include/cker/train/operation/DepthwiseConv.h
jyoungyun: This code also came from TensorFlow. How about keeping the existing code for maintaining?

compute/cker/src/train/DepthwiseConv.test.cc
jyoungyun: This buffers are used for temporary memory. Should I verify the temp memory?

compute/cker/src/train/DepthwiseConv.test.cc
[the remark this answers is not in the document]
glistening: I don't fully understand. Out-of-curiosity, I searched tensorflow code (e.g. depthwise_conv_op.cc), it seems to use worker thread counts. I don't find `+`, but I found the following:
```cc
auto worker_threads = *(ctx->device()->tensorflow_cpu_worker_threads());
```
I guess, the `thread_count` should be the number of worker threads.
Is the main thread is part of worker threads?

compute/cker/include/cker/train/operation/DepthwiseConv.h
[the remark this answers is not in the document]
ragmani: I understand why you used function template. You can use polymorphism to make `DepthwiseConvolutionLayer` independent from type of `DepthwiszeConv`, but I think it's not a good approach.

I still think it's better to use class template because there may be ambiguity in using a different type from the kernel functions below when using function template. But it would be better to keep this code as is for now as long as there is no alternative.

compute/cker/include/cker/train/operation/DepthwiseConv.h
ragmani: Isn't this code in this file that can be modified?
Anyway, this is optional.

compute/cker/src/train/DepthwiseConv.test.cc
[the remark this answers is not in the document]
ragmani: Is there any reason to add 1 for main thread in the thread count? As far as I know, main thread is usually not used in optimized kernel.

compute/cker/src/train/DepthwiseConv.test.cc
ragmani: It seems that there was a misunderstanding about what we talked about offline in the past. I think It's better to manage temporary memories in kernel if they are only used in kernel?
Could you move temporary memories into kernel?

compute/cker/include/cker/eigen/depthwise_conv_op.h
ragmani: I think it's better to remove `static` even if this file came from tensorflow.

compute/cker/include/cker/eigen/depthwise_conv_op.h
ragmani: I think it's better to remove static even if this file came from tensorflow.

compute/cker/src/train/DepthwiseConv.test.cc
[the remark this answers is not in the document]
jyoungyun: > Is there any reason to add 1 for main thread in the thread count? As far as I know, main thread is usually not used in optimized kernel.

Could you give a link related to `main thread is usually not used in optimized kernel`? This is because when I tested this kernel using several cases, the main thread was also used. Especially if the batch size is 1, it usually uses the main thread. And in other cases, which thread is being used depends on the situation.

https://github.com/libigl/eigen/blob/1f05f51517ec4fd91eed711e0f89e97a7c028c0e/unsupported/Eigen/CXX11/src/Tensor/TensorDeviceThreadPool.h#L175-L179
```
    if (n <= 1 || numThreads() == 1 ||
        CostModel::numThreads(n, cost, static_cast<int>(numThreads())) == 1) {
      f(0, n);
      return;
    }
```
In this code, n is a batch_size. And the batch size is 1, it works on main thread.

```
(gdb) info threads
  Id   Target Id                                       Frame 
  1    Thread 0x7ffff7a48780 (LWP 1326007) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=0, limit=1) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
  2    Thread 0x7ffff797e700 (LWP 1326011) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0c9d8) at ../sysdeps/nptl/futex-internal.h:183
  3    Thread 0x7ffff717d700 (LWP 1326012) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0ca58) at ../sysdeps/nptl/futex-internal.h:183
  4    Thread 0x7ffff697c700 (LWP 1326013) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0cad8) at ../sysdeps/nptl/futex-internal.h:183
  5    Thread 0x7ffff617b700 (LWP 1326014) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0cb58) at ../sysdeps/nptl/futex-internal.h:183
  6    Thread 0x7ffff597a700 (LWP 1326015) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0cbd8) at ../sysdeps/nptl/futex-internal.h:183
  7    Thread 0x7ffff5179700 (LWP 1326016) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0cc58) at ../sysdeps/nptl/futex-internal.h:183
  8    Thread 0x7ffff4978700 (LWP 1326017) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0ccd8) at ../sysdeps/nptl/futex-internal.h:183
  9    Thread 0x7ffff4177700 (LWP 1326018) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0cd58) at ../sysdeps/nptl/futex-internal.h:183
  10   Thread 0x7ffff3976700 (LWP 1326019) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0cdd8) at ../sysdeps/nptl/futex-internal.h:183
  11   Thread 0x7ffff3175700 (LWP 1326020) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0ce58) at ../sysdeps/nptl/futex-internal.h:183
  12   Thread 0x7ffff2974700 (LWP 1326021) "test_cker" futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x555555b0ced8) at ../sysdeps/nptl/futex-internal.h:183
  13   Thread 0x7ffff2173700 (LWP 1326022) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=8, limit=9) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
  14   Thread 0x7ffff1972700 (LWP 1326023) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=9, limit=10) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
  15   Thread 0x7ffff1171700 (LWP 1326024) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=6, limit=7) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
  16   Thread 0x7ffff0970700 (LWP 1326025) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=4, limit=5) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
  17   Thread 0x7ffff016f700 (LWP 1326026) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=3, limit=4) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
  18   Thread 0x7fffef96e700 (LWP 1326027) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=7, limit=8) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
* 19   Thread 0x7fffef16d700 (LWP 1326028) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=5, limit=6) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
  20   Thread 0x7fffee96c700 (LWP 1326029) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=1, limit=2) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
  21   Thread 0x7fffee16b700 (LWP 1326030) "test_cker" nnfw::cker::depthwise_conv_op::LaunchDepthwiseConvBackpropInputOp<Eigen::ThreadPoolDevice, float>::operator()(int, int, int, int, int, int, int, int, int, int, int, int, int, float const*, float const*, float*, float*, bool, float*, float*)::{lambda(long, long)#1}::operator()(long, long) const (this=0x555555b0c560, start=2, limit=3) at /home/jyoung/git/npu/ONE/compute/cker/include/cker/eigen/depthwise_conv_op.h:592
```

And in the multiple batch size case, the main thread is also used to execute the `f` function. So I added the `+1` for main thread.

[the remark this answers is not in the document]
jyoungyun: https://github.com/libigl/eigen/blob/1f05f51517ec4fd91eed711e0f89e97a7c028c0e/unsupported/Eigen/CXX11/src/Tensor/TensorDeviceThreadPool.h#L241-L242

You can find the code that directly executes the user function. Using or not using the main thread is implementation-dependent and not a set rule. Additionally, in the Eigen library, it uses the main thread, so I need to add an additional memory buffer for the main thread.

compute/cker/include/cker/train/operation/DepthwiseConv.h
jyoungyun: I prefer to keep the existing code for maintaining issue. The cker code almost(?) came from TensorFlow. Even a simple expression may be difficult to search if it does not match. For this reason, I copied the TensorFlow code with a minimum of modification.

compute/cker/src/train/DepthwiseConv.test.cc
jyoungyun: I don't understand what you said. Why do you suggest moving these temporary memories into the kernel? If these memories are located in the backend kernel, we can reuse these temporary memory for optimization.

compute/cker/src/train/DepthwiseConv.test.cc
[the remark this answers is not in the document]
glistening: 1. our `ruy` backend provides a way to limit the max_threads from environment variables.

runtime/onert/backend/ruy/ExternalContext.h

```c
class ExternalContext
{
private:
  static const int kDefaultNumThreadpoolThreads = 4;

public:
  ExternalContext() : _ruy_context(new ::ruy::Context)
  {
    setMaxNumThreads(onert::util::getConfigInt(onert::util::config::RUY_THREADS));
  }

  void setMaxNumThreads(int max_num_threads)
  {
    const int target_num_threads =
      max_num_threads > -1 ? max_num_threads : kDefaultNumThreadpoolThreads;
    _ruy_context->set_max_num_threads(target_num_threads);
  }
```
If I remember correctly, the thread count does not include main threads. Main thread waits for worker threads.

2. I don't know much about eigen internal, but at a glance,

```c
    if (n <= 1 || numThreads() == 1 ||
        CostModel::numThreads(n, cost, static_cast<int>(numThreads())) == 1) {
      f(0, n);
      return;
    }
```

seems to be the case for 1 batch. I don't understand why it spawns 21 or more threads.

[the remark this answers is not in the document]
jyoungyun: > If I remember correctly, the thread count does not include main threads. Main thread waits for worker threads.

I didn't say that the thread count includes main thread. The thread count returns 20 (in x64 machine) and main thread is not included in that count. So I added `+1` for main thread. And it is an eigen library. This code has nothing to do with ruy backend behaviour.

> seems to be the case for 1 batch. I don't understand why it spawns 21 or more threads.

I already commented the code that executes the operation in main thread before the main thread waits.

ragmani:
> [2024-01-16T02:01:00.847Z] Cannot contact xu4-ubuntu-01: hudson.remoting.RequestAbortedException: java.nio.channels.ClosedChannelException

@nnfw-bot test onert-cross-release

ragmani:
@nnfw-bot test onert-cross-release

compute/cker/src/train/DepthwiseConv.test.cc
ragmani: As we talked about offline, I suggested moving those memories into the kernel since those memories are only used in kernel. But I couldn't find the way for cker to manage and optimize temporary memories. I think temporary memories would be managed directly in train backend. In short, it would be better to keep it as is.

compute/cker/src/train/DepthwiseConv.test.cc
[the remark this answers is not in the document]
ragmani: @jyoungyun 

> seems to be the case for 1 batch. I don't understand why it spawns 21 or more threads.

I agree with @glistening . I think the code including the code below does not explain why main thread should be added to the thread count.
```c++
    Barrier barrier(static_cast<unsigned int>(block_count));
    std::function<void(Index, Index)> handleRange;
    handleRange = [=, &handleRange, &barrier, &f](Index first, Index last) {
      if (last - first <= block_size) {
        // Single block or less, execute directly.
        f(first, last);
        barrier.Notify();
        return;
      }
      // Split into halves and submit to the pool.
      Index mid = first + divup((last - first) / 2, block_size) * block_size;
      pool_->Schedule([=, &handleRange]() { handleRange(mid, last); });
      handleRange(first, mid);
    };
    handleRange(0, n);
    barrier.Wait();
  }
```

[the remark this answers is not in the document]
jyoungyun: @ragmani In the above code, the main thread schedules for multi-threading and also executes the `handleRange` itself.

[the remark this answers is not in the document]
glistening: > @glistening
> 
> > If I remember correctly, the thread count does not include main threads. Main thread waits for worker threads.
> 
> I didn't say that the thread count includes main thread. And that's why I added `1` to the thread counts. The thread count returns 20 (in x64 machine) and main thread is not included in that count. So I added `1` for main thread. And it is an eigen library. This code has nothing to do with ruy backend behaviour.

I guess I already understood what you say. I want to keep our code that uses external library consistent. A few years ago, there was some requirements to limit the total number of threads. I guess `onert::util::config::RUY_THREADS` make it easy to control how many threads will be allowed for ruy kernel. But this PR of eigen case is different. It is hard to control and understand how many threads are used/allowed for this kernel.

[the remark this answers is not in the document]
jyoungyun: @glistening 

> What is returned by `_dconv_kernel->getThreadCount()`? Is the number controllable?

It returns the core number. And we couldn't change this value.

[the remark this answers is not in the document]
ragmani: > @ragmani In the above code, the main thread schedules for multi-threading and also executes the handleRange itself.

You are right. Please ignore my comment.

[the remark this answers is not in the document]
ragmani: @jyoungyun 

> In the above code, the main thread schedules for multi-threading and also executes the `handleRange` itself.

It seems that you are right. Main thread may call the code below.

```c++
      if (last - first <= block_size) {
        // Single block or less, execute directly.
        f(first, last);
        barrier.Notify();
        return;
      }
```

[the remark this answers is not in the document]
ragmani: @glistening 
> I guess I already understood what you say. I want to keep our code that uses external library consistent. A few years ago, there was some requirement to limit the total number of threads. I guess onert::util::config::RUY_THREADS make it easy to control how many threads will be allowed for ruy kernel. Also, we can tune how many threads are better by changing the number of thread.

`cker` has already been using `Eigen::ThreadPoolDevice`, which has as many threads as the number of cores. For example, `Conv` kernel for inference, optimizer kernels for training, `BroadcastTo` kernel. If we only use kernels that easily controls the number of threads, we have to implement kernels ourselves to replace kernels that are hard to control. 

@jyoungyun 
Is there a way to apply the optimized `DepthwiseConv` kernels without calling `getThreadCount()`?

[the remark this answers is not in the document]
ragmani: If we controls the number of threads in `Eigen::ThreadPoolDevice`, we have to modify the code below:
https://github.com/Samsung/ONE/blob/2a4d8d60c77e45672ac3dfabf7970d6005074f93/compute/cker/include/cker/eigen/EigenSupport.h#L85-L114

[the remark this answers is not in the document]
glistening: @jyoungyun and @ragmani Thank you for explanation. For Eigen, the number of thread is determined by the number of core on the device. It is not controllable like `ruy` backend.

glistening (approved):
LGTM

ragmani (approved):
LGTM

YongseopKim (approved):

Issue #12455, opened 2024-01-12
[compiler] Check upgrade to c++17


seanshpark:
Let's check upgrade to c++17 turns out and deprecate using boost.

seanshpark:
flatbuffers `Length()` is going to be deprecated and suggests to use `size()`

seanshpark:
Ubuntu 18.04 link error when using `std::filesystem`
> undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
> undefined reference to `std::filesystem::create_directory(std::filesystem::__cxx11::path const&)'

glistening:
Some of compiler (especially quantizer-related things) are used for on-device compiler. And Tizen does not support c++17 yet.

seanshpark:
done for now

PR #12457, opened 2024-01-12
DRAFT compiler use c++17

seanshpark:
on-going draft to use c++17 and fixes


seanshpark:
for #12455

seanshpark:
done

PR #12464, opened 2024-01-15
[cirle-inspect] Revise to use size()

seanshpark:
This will revise to use size() as Length() is going to be deprecated.


seanshpark:
for https://github.com/Samsung/ONE/issues/12455#issuecomment-1888529622 , from draft #12457

mhs4670go (approved):
LGTM

PR #12470, opened 2024-01-15
[infra/nncc] Revise to use c++17

seanshpark:
This will revise compiler to use c++17.


seanshpark:
for #12455 , from draft #12457

jinevening (approved):
LGTM

PR #12407, opened 2024-01-04
[draft] Constant folding for Shape op

miusic:
This draft folds the Shape op as a constant. (WIP)
[ci skip]

---
related to: #12046

Issue #12046, opened 2023-11-16
[one-optimize] Constant folding for Shape/StridedSclie/Pack


jinevening:
## What

Let's fold Shape, StridedSlice, Pack operators.

![image](https://github.com/Samsung/ONE/assets/5449554/85c86059-3bcd-42bb-9f73-29b61f748523)

mhs4670go:
If it become not urgent work, maybe it could be good first issue for newbies.

miusic:
If this issue does not urgent, can I try to work on it?

And is [this issue](https://github.com/Samsung/ONE/issues/5598) good enough as a reference?

jinevening:
>  can I try to work on it?

@miusic Sure :) #5598 would be a good reference for you.

jinevening:
You can use the below recipe. Please take a look at `tflchef` to see how to generate a tflite model from the recipe.

```
operand {
  name: "functional_3/conv2d_transpose_31/stack/3"
  type: INT32
  shape {
  }
  quant {
    quantized_dimension: 0
  }
  is_variable: false
}
operand {
  name: "functional_3/up4/stack/3"
  type: INT32
  shape {
  }
  quant {
    quantized_dimension: 0
  }
  is_variable: false
}
operand {
  name: "functional_3/up5/stack/3"
  type: INT32
  shape {
  }
  quant {
    quantized_dimension: 0
  }
  is_variable: false
}
operand {
  name: "functional_3/conv2d_transpose_16/conv2d_transpose"
  type: FLOAT32
  shape {
    dim: 16
    dim: 5
    dim: 5
    dim: 32
  }
  filler {
    tag: "gaussian"
    arg: "0.0"
    arg: "0.1"
  }
  quant {
    quantized_dimension: 0
  }
  is_variable: false
}
operand {
  name: "functional_3/up6/strided_slice"
  type: INT32
  shape {
    dim: 1
  }
  filler {
    tag: "explicit"
    arg: "0"
  }
  quant {
    quantized_dimension: 0
  }
  is_variable: false
}
operand {
  name: "functional_3/up6/strided_slice1"
  type: INT32
  shape {
    dim: 1
  }
  filler {
    tag: "explicit"
    arg: "1"
  }
  quant {
    quantized_dimension: 0
  }
  is_variable: false
}
operand {
  name: "functional_3/tf_op_layer_split_3/split_3"
  type: FLOAT32
  shape {
    dim: 1
    dim: 32
    dim: 64
    dim: 32
[...]

PR #12478, opened 2024-01-16
[luci/pass] Introduce the pass to FoldShape

miusic:
This commit introduces a pass to fold Shape op

---
Related to: #12046
Draft: #12407

ONE-DCO-1.0-Signed-off-by: jihunnn-kim <jihunnn.kim@samsung.com>

### DIFF
--- a/compiler/luci/pass/src/FoldShapePass.cpp
+++ b/compiler/luci/pass/src/FoldShapePass.cpp
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) 2024 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FoldShapePass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/Profile/CircleNodeOrigin.h>
+
+namespace
+{
+
+template <loco::DataType OutType> luci::CircleConst *folding(luci::CircleShape *shape)
+{
+  auto input_node = loco::must_cast<luci::CircleNode *>(shape->input());
+  auto name = input_node->name();
+  assert(name.length() > 0);
+  auto shape_status = input_node->shape_status();
+  if (shape_status != luci::ShapeStatus::VALID)
+    return nullptr;
+  auto rank = input_node->rank();
+  // TODO support rank == 0 when possible
+  if (rank == 0)
+    return nullptr;
+  for (uint32_t i = 0; i < rank; i++)
+  {
+    auto dim = input_node->dim(i);
+    if (!dim.known())
+      return nullptr;
+  }
+
+  auto folded_shape = input_node->graph()->nodes()->create<luci::CircleConst>();
+  folded_shape->name(name + "_ConstShape");
+  folded_shape->dtype(OutType);
+  folded_shape->rank(1);
+  folded_shape->dim(0).set(rank);
+  luci::add_origin(folded_shape, luci::get_origin(shape));
+
+  folded_shape->size<OutType>(rank);
+  for (uint32_t i = 0; i < rank; i++)
+    folded_shape->at<OutType>(i) = input_node->dim(i).value();
+
+  return folded_shape;
+}
+
+// Fold Shape to const if the input shape is static
+template <loco::DataType OutType> bool fold_shape(luci::CircleShape *shape)
+{
+  auto folded_shape = folding<OutType>(shape);
+  if (not folded_shape)
+    return false;
+
+  loco::replace(shape).with(folded_shape);
+
+  return true;
+}
+
+} // namespace
+
+namespace luci
+{
+
+/**
+ * BEFORE
+ *
+ *     [CircleNode]
+ *           |
+ *     [CircleShape]
+ *           |
+ *     [CircleNode]
+ *
+ * AFTER
+ *
+ *     [CircleConst]  [CircleNode]
+ *           |
+ *     [CircleNode]
+ *
+ */
+bool FoldShapePass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    if (auto shape = dynamic_cast<luci::CircleShape *>(node))
+    {
+      auto out_type = shape->out_type();
+      switch (out_type)
+      {
+        case loco::DataType::S32:
+          if (fold_shape<loco::DataType::S32>(shape))
+            changed = true;
+          break;
+        case loco::DataType::S64:
+          if (fold_shape<loco::DataType::S64>(shape))
+            changed = true;
+          break;
+        default:
+          break;
+      }
+    }
+  }
+
+  return changed;
+}
+
+} // namespace luci
--- a/compiler/luci/pass/src/FoldShapePass.test.cpp
+++ b/compiler/luci/pass/src/FoldShapePass.test.cpp
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2024 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "luci/Pass/FoldShapePass.h"
+#include "PassTestGraphs.h"
+
+#include <luci/IR/CircleNodes.h>
+
+#include <gtest/gtest.h>
+
+namespace
+{
+
+template <loco::DataType OutType> class FoldShapeGraph : public luci::ConstantFoldingAddTestGraph
+{
+public:
+  FoldShapeGraph(std::vector<uint32_t> input_shape)
+    : luci::ConstantFoldingAddTestGraph(input_shape, OutType)
+  {
+    _x = _g.nodes()->template create<luci::CircleConst>();
+    _x->name("x");
+    _x->dtype(loco::DataType::S32);
+    _x->rank(input_shape.size());
+    for (uint32_t i = 0; i < input_shape.size(); i++)
+      _x->dim(i).set(input_shape.at(i));
+    _x->shape_status(luci::ShapeStatus::VALID);
+
+    _shape = _g.nodes()->template create<luci::CircleShape>();
+    _shape->name("shape");
+    _shape->out_type(OutType);
+    _shape->input(_x);
+    _shape->shape({4});
+    _shape->rank(1);
+    _shape->dim(0).set(4);
+  }
+
+  loco::Node *createFoldedPattern() override { return _shape; }
+
+protected:
+  luci::CircleShape *_shape = nullptr;
+  luci::CircleConst *_x = nullptr;
+};
+
+/**
+ *  Graph that has a Shape Op
+ *
+ *    BEFORE
+ *
+ *                     [CircleConst]
+ *                           |
+ *    [CircleInput]    [CircleShape]
+ *              \       /
+ *             [CircleAdd]
+ *                  |
+ *            [CircleOutput]
+ *
+ *    AFTER
+ *
+ *    [CircleInput]    [CircleConst]
+ *              \       /
+ *             [CircleAdd]
+ *                  |
+ *            [CircleOutput]
+ *
+ */
+class FoldShapePassGraphTest : public FoldShapeGraph<loco::DataType::S32>, public ::testing::Test
+{
+public:
+  FoldShapePassGraphTest() : FoldShapeGraph<loco::DataType::S32>({1, 8, 8, 64}) {}
+
+  virtual void SetUp() { init(); }
+};
+
+} // namespace
+
+TEST(FoldShapePassTest, name)
+{
+  luci::FoldShapePass pass;
+  auto const name = pass.name();
+  ASSERT_NE(nullptr, name);
+}
+
+TEST_F(FoldShapePassGraphTest, fold_shape)
+{
+  luci::FoldShapePass pass;
+  while (pass.run(graph()))
+    ;
+
+  auto folded_const = getFoldedPattern();
+  EXPECT_NE(nullptr, folded_const);
+
+  // Check type, shape, values of folded shape
+  EXPECT_EQ(loco::DataType::S32, folded_const->dtype());
+  EXPECT_EQ(1, folded_const->rank());
+  EXPECT_EQ(4, folded_const->dim(0).value());
+  EXPECT_EQ(1, folded_const->at<loco::DataType::S32>(0));
+  EXPECT_EQ(8, folded_const->at<loco::DataType::S32>(1));
+  EXPECT_EQ(8, folded_const->at<loco::DataType::S32>(2));
+  EXPECT_EQ(64, folded_const->at<loco::DataType::S32>(3));
+}
+
+TEST_F(FoldShapePassGraphTest, undefined_shape_NEG)
+{
+  _x->shape_status(luci::ShapeStatus::UNDEFINED);
+
+  luci::FoldShapePass pass;
+  EXPECT_FALSE(pass.run(graph()));
+}
+
+TEST_F(FoldShapePassGraphTest, unallowed_rank_NEG)
+{
+  _x->rank(0);
+
+  luci::FoldShapePass pass;
+  EXPECT_FALSE(pass.run(graph()));
+}
+
+TEST_F(FoldShapePassGraphTest, unknown_dimension_NEG)
+{
+  _x->dim(0).unset();
+
+  luci::FoldShapePass pass;
+  EXPECT_FALSE(pass.run(graph()));
+}
--- a/compiler/luci/pass/include/luci/Pass/FoldShapePass.h
+++ b/compiler/luci/pass/include/luci/Pass/FoldShapePass.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2024 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __LUCI_FOLD_SHAPE_PASS_H__
+#define __LUCI_FOLD_SHAPE_PASS_H__
+
+#include <logo/Pass.h>
+
+namespace luci
+{
+
+/**
+ * @brief  Class to fold Shape to a constant tensor
+ */
+struct FoldShapePass final : public logo::Pass
+{
+  const char *name(void) const final { return "luci::FoldShapePass"; }
+
+  bool run(loco::Graph *g) final;
+};
+
+} // namespace luci
+
+#endif // __LUCI_FOLD_SHAPE_PASS_H__


seanshpark:
1/ plz read and follow https://github.com/Samsung/ONE/wiki/ONE-Developer%27s-Certificate-of-Origin
2/ do not remove signed-off in PR message

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,91 @@
...
+  auto folded_shape = input_node->graph()->nodes()->create<luci::CircleConst>();
+  folded_shape->name(name + "_ConstShape");
+  folded_shape->dtype(OutType);
+  folded_shape->rank(1);
+  folded_shape->dim(0).set(rank);
+  luci::add_origin(folded_shape, luci::get_origin(shape));
+
+  folded_shape->size<OutType>(rank);
+  for (uint32_t i = 0; i < rank; i++)
+    folded_shape->at<OutType>(i) = input_node->dim(i).value();
```
seanshpark, line 46: what happens when dim is unknown? is it acceptable?

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+  return folded_shape;
+}
+
+/**
+ * Fold Shape to const if the input shape is fixed
+ **/
```
seanshpark, line 53: this looks like for doxygen but it is not.
plz go with doxygen format or just plain text comment.

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+} // namespace
+
+namespace luci
+{
+
+/**
+ * Constant Folding for Shape Op
+ **/
```
seanshpark, line 72: ditto

compiler/luci/pass/src/FoldShapePass.cpp
seanshpark: `value()` is `uint32_t` but may be different with `OutType`

seanshpark:
you have unit test in your draft. why not add it to this PR?

miusic:
>1/ plz read and follow https://github.com/Samsung/ONE/wiki/ONE-Developer%27s-Certificate-of-Origin
2/ do not remove signed-off in PR message

Sorry for missing the signed-off tag.. I'll update soon!

miusic:
> you have unit test in your draft. why not add it to this PR?

I have planned to make the PR like below (refered #5707)

- Introduce the pass to fold Shape op
- Add test for FoldShapePass
- Add option to fold Shape op
- Support for FoldShape pass

Is it better to change the order?

seanshpark:
> Is it better to change the order?

It would be better to add pass and pass.test in one PR.
IMHO changes for `CircleOptimizer` is in different context.

compiler/luci/pass/src/FoldShapePass.cpp
miusic: > what happens when dim is unknown? is it acceptable?

Thank you for pointing this out!
Since it is not acceptable as you said, I'll update to check if the `dim` is kwown or not 😄

miusic: > `value()` is `uint32_t` but may be different with `OutType`

Yes, it is bit different now.. but I'm not sure which is the proper way
For now, I just implemented it as follow the default `out_type` value of `tf.shape`, which is signed-32bit integer [[link](https://www.tensorflow.org/api_docs/python/tf/shape)]

Is it better to match as unsigned-32 as follwing the data type of `value()`?

seanshpark: You made this PR with `if (fold_shape<loco::DataType::S32>(shape))` call.
Why did you use `S32` ?

miusic: > Why did you use S32 ?

First of all, my intend was following the default `out_type` of `tf.shape` (`S32`)

But, IIUC, the `out_type` of `CircleShape` can be designated by the user
So I tried to use `shape->out_type()` like below, but it was failed 😥

```cpp
auto out_type = shape->out_type();
if (fold_shape<out_type>(shape))
```

Therefore I made this PR with `S32` first, and added `TODO Support more data types` comment

If there is anything I missed, please let me know :)

seanshpark: > TODO Support more data types

what types are you planning to support after this PR?

do you know what output types exist for `Shape` Op?

miusic:
## Unit test

`./build/compiler/luci/pass/luci_pass_test --gtest_filter=FoldShape*`

```bash
Note: Google Test filter = FoldShape*
[==========] Running 2 tests from 2 test suites.
[----------] 1 test from FoldShapePassTest
[ RUN      ] FoldShapePassTest.name
[       OK ] FoldShapePassTest.name (0 ms)
[----------] 1 test from FoldShapePassTest (0 ms total)

[----------] 1 test from FoldShapeGraphTest
[ RUN      ] FoldShapeGraphTest.fold_shape
[       OK ] FoldShapeGraphTest.fold_shape (0 ms)
[----------] 1 test from FoldShapeGraphTest (0 ms total)

[==========] 2 tests from 2 test suites ran. (0 ms total)
[  PASSED  ] 2 tests.
```

compiler/luci/pass/src/FoldShapePass.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  // Check type, shape, values of folded shape
+  EXPECT_EQ(loco::DataType::S32, folded_const->dtype());
+  EXPECT_EQ(1, folded_const->rank());
+  EXPECT_EQ(4, folded_const->dim(0).value());
+  EXPECT_EQ(1, folded_const->at<loco::DataType::S32>(0));
+  EXPECT_EQ(8, folded_const->at<loco::DataType::S32>(1));
+  EXPECT_EQ(8, folded_const->at<loco::DataType::S32>(2));
+  EXPECT_EQ(64, folded_const->at<loco::DataType::S32>(3));
+}
```
seanshpark, line 109: add negative tests that makes `count_of_pos() <= count_of_neg()`

miusic:
I'll separate this PR into `[luci/pass] Introduce the pass to FoldShape` and `[luci] Introduce the FoldShape` (latter is about `CircleOptimizer`)

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+#include "luci/Pass/FoldShapePass.h"
+
+#include <luci/IR/CircleNodes.h>
+#include <luci/Profile/CircleNodeOrigin.h>
+
+namespace
+{
+
```
seanshpark, line 24: plz add a comment of text graph like other passes that shows how the graph changes before-after the pass.

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,93 @@
...
+  return true;
+}
+
+} // namespace
+
+namespace luci
+{
+
+// Constant Folding for Shape Op
```
seanshpark, line 74: this comment seems unncessary

compiler/luci/pass/src/FoldShapePass.test.cpp
miusic: Ok, I'll update it soon!

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,93 @@
...
+  luci::add_origin(folded_shape, luci::get_origin(shape));
+
+  folded_shape->size<OutType>(rank);
+  for (uint32_t i = 0; i < rank; i++)
+    folded_shape->at<OutType>(i) = input_node->dim(i).value();
+
+  return folded_shape;
+}
+
+// Fold Shape to const if the input shape is fixed
```
seanshpark, line 57: q) what is the meaning of `fixed` ?

miusic: It seems more proper to say `valid`
Sorry for confusion 😅

seanshpark: `valid` also looks vague . 
or can you explain the meaning of `valid` here?

compiler/luci/pass/src/FoldShapePass.cpp
miusic: > what types are you planning to support after this PR?

For now, I planned to support `S64` only, and the details are as follow

> do you know what output types exist for Shape Op?

I guess that the `Shape` supports `S32` and `S64`, due to the following statement from [tf.shape](https://www.tensorflow.org/api_docs/python/tf/shape)

> The specified output type of the operation (int32 or int64).

But I'm not sure that these two types cover all the types that supported...
This is the reason why I remained it as a `TODO` item

compiler/luci/pass/src/FoldShapePass.cpp
miusic: Ok! I'll update

(Just out of curiosity) Where is the proper place for showing how the graph changes before-after pass?

Some pass commented it at the source of the pass file (_~~Pass.cpp_)
(e.g. `FoldDensifyPass.cpp` / `FoldDequantizePass.cpp`),
and some pass commented it at the test file (_~~Pass.test.cpp_)
(e.g. `FoldCastPass.test.cpp`),
and some pass commented it both of the source file (_~~Pass.cpp_) and the test file (_~~Pass.test.cpp_)
(e.g. `FoldDepthwiseConv2DPass.cpp`, `FoldDepthwiseConv2DPass.test.cpp` / `FoldFullyConnectedPass.cpp`, `FoldFullyConnectedPass.test.cpp`)

Is it the best that showing how the graph changes at the both of the source file and the test file..?

compiler/luci/pass/src/FoldShapePass.cpp
seanshpark: "whose shape is known, not dynamic" and "VALID" have different points of view.

please search the internet for shape and stuff how others represent and how our codes are implemented.

compiler/luci/pass/src/FoldShapePass.cpp
seanshpark: What is your understanding by my asking to add a graph?
What is the help of the graph comment in the source do you get?
What is the help of the graph comment in the test do you get?

compiler/luci/pass/src/FoldShapePass.cpp
miusic: Sorry for the vague explanation

I'll find more explicit expression refering your advice :)

compiler/luci/pass/src/FoldShapePass.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+ *
+ *      [Input]   [Shape]
+ *           \    /
+ *            [Add]
+ *              |
+ *           [Output]
+ *
+ *    AFTER
+ *
+ *         [CircleConst]
```
jinevening, line 71: BEFORE has input, output, add, so this should be updated.

compiler/luci/pass/src/FoldShapePass.cpp
miusic: I think that **static** is proper to explain what I want to say

below is the meaning of each word what I understand

**static shape** -

compiler/luci/pass/src/FoldShapePass.test.cpp
miusic: May I ask some more details..? I faild to catch the point 😥

compiler/luci/pass/src/FoldShapePass.cpp
miusic: I think that **static** is proper word to explain what I want to say

---
below is the meaning of the words what I understand

**static shape** - the shape that is determinized at graph definition time (fully-known shape)
It can folded as a constant before run the graph
(e.g. [1, 8, 8, 64])

**dynamic shape** - the shape that is determinized at graph execution time (partially-known shape)
It cannot folded as a constant because we don't know the output of the **shape** node before run the graph
(e.g. [None, 8, 8, 64])

**valid shape** - the shape is known (fully-known or partially known)
This term includes **static shape** and **dynamic shape**
If we don't know any of the dimension value, it is unknown shape (e.g. [None, None, None, None])

miusic: I think that **static** is proper word to explain what I want to say

---
below is the meaning of the words what I understand

**static shape ** - the shape that is determinized at graph definition time (fully-known shape)
It can folded as a constant before run the graph
(e.g. [1, 8, 8, 64])

**dynamic shape** - the shape that is determinized at graph execution time (partially-known shape / unknown shape / unknown shape and rank)
It cannot folded as a constant because we don't know the output of the **shape** node before run the graph
(e.g. [None, 8, 8, 64])

**valid shape** - the shape is known (fully-known or partially known)

compiler/luci/pass/src/FoldShapePass.test.cpp
miusic: updated!

miusic: ## Unit test

`./build/compiler/luci/pass/luci_pass_test --gtest_filter=FoldShape*`

---
```bash
Note: Google Test filter = FoldShape*
[==========] Running 5 tests from 2 test suites.
[----------] 1 test from FoldShapePassTest
[ RUN      ] FoldShapePassTest.name
[       OK ] FoldShapePassTest.name (0 ms)
[----------] 1 test from FoldShapePassTest (0 ms total)

[----------] 4 tests from FoldShapeGraphTest
[ RUN      ] FoldShapeGraphTest.fold_shape
[       OK ] FoldShapeGraphTest.fold_shape (2 ms)
[ RUN      ] FoldShapeGraphTest.undefined_shape_NEG
[       OK ] FoldShapeGraphTest.undefined_shape_NEG (0 ms)
[ RUN      ] FoldShapeGraphTest.rank_mismatch_NEG
[       OK ] FoldShapeGraphTest.rank_mismatch_NEG (0 ms)
[ RUN      ] FoldShapeGraphTest.unknown_dimension_NEG
[       OK ] FoldShapeGraphTest.unknown_dimension_NEG (0 ms)
[----------] 4 tests from FoldShapeGraphTest (2 ms total)

[==========] 5 tests from 2 test suites ran. (2 ms total)
[  PASSED  ] 5 tests.
```

compiler/luci/pass/src/FoldShapePass.test.cpp
```diff
@@ -0,0 +1,133 @@
...
```
jinevening, line 71: Please update the comment as follows :)

```suggestion
 *      [Input]   [Const]
 *           \    /
 *            [Add]
 *              |
 *           [Output]
```

miusic: Updated! :D

miusic:
> 2/ do not remove signed-off in PR message

I didn't remove the signed-off in PR message but it does not appear automatically.. but it was added in commit message
Is there no problem if I add this message manually?

seanshpark:
> Is there no problem if I add this message manually?

no problem.

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * BEFORE
+ *
+ *        [Input]
+ *           |
+ *        [Shape]
+ *           |
+ *        [Output]
+ *           |
```
seanshpark, line 82: `|` connected to `[Output]` looks strange. what is your intention of this?

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,108 @@
...
+  auto rank = input_node->rank();
+  assert(name.length() > 0);
+  if (shape_status != luci::ShapeStatus::VALID)
+    return nullptr;
+  if (rank <= 0)
+    return nullptr;
+  for (uint32_t i = 0; i < rank; i++)
+  {
+    auto dim = input_node->dim(i);
+    if (!dim.known()) // If dim is unknown, return nullptr
```
seanshpark, line 39: I don't think `// If dim is unknown, return nullptr` is necessary.
It just explains the code itself.

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,108 @@
...
+namespace
+{
+
+template <loco::DataType OutType> luci::CircleConst *folding(luci::CircleShape *shape)
+{
+  auto input_node = loco::must_cast<luci::CircleNode *>(shape->input());
+  auto name = input_node->name();
+  auto shape_status = input_node->shape_status();
+  auto rank = input_node->rank();
+  assert(name.length() > 0);
```
seanshpark, line 31: let's move this line just after `auto name = ... `line

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,108 @@
...
+  if (shape_status != luci::ShapeStatus::VALID)
+    return nullptr;
+  if (rank <= 0)
```
seanshpark, line 34: have you checked with rank 0 constant?

seanshpark: can `rank` be minus?

compiler/luci/pass/src/FoldShapePass.cpp
miusic: Did you mean `|` below `[Output]`, right?

My intention was that there can be connected graph after the `[Output]`... but it looks unnecessary as I see it now

I'll remove it, thank you!

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,108 @@
...
+ */
+bool FoldShapePass::run(loco::Graph *g)
+{
+  bool changed = false;
+  for (auto node : loco::active_nodes(loco::output_nodes(g)))
+  {
+    if (auto shape = dynamic_cast<luci::CircleShape *>(node))
+    {
+      // TODO Support more data types
+      if (fold_shape<loco::DataType::S32>(shape))
```
seanshpark, line 98: `CircleShape` has `out_type`. why not use this?

compiler/luci/pass/src/FoldShapePass.cpp
miusic: For consistency, is it better to move the lines which check the constraint of `shape_status` and `rank` to just after the declaration of `shape_status` and `rank` respectively?

compiler/luci/pass/src/FoldShapePass.test.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "PassTestGraphs.h"
+
+#include <luci/IR/CircleNodes.h>
+
+#include <gtest/gtest.h>
+
+namespace
+{
+
+template <loco::DataType FromT, loco::DataType ToT>
```
seanshpark, line 27: Q) what is the meaning of `FromT` and `ToT` in this test?

compiler/luci/pass/src/FoldShapePass.cpp
miusic: I've tried like below but failed.. 😥 https://github.com/Samsung/ONE/pull/12478#discussion_r1452918067

```cpp
auto out_type = shape->out_type();
if (fold_shape<out_type>(shape))
```

If the way I've tried was wrong, then may I ask you what I have to refer to? (code, document, or keyword for search..?)

compiler/luci/pass/src/FoldShapePass.test.cpp
```diff
@@ -0,0 +1,137 @@
...
+  luci::CircleShape *_shape = nullptr;
+  luci::CircleConst *_x = nullptr;
+};
+
+/**
+ *  Graph that has a Shape Op
+ *
+ *    BEFORE
+ *
+ *      [Input]   [Shape]
```
seanshpark, line 63: Can you add input of `[Shape]` ?

compiler/luci/pass/src/FoldShapePass.cpp
seanshpark: I don't think there is strict rules for this.
Usually we add assert check for the variable just after declaration for condition check.

compiler/luci/pass/src/FoldShapePass.cpp
seanshpark: something like this
```cxx
auto out_type = shape->out_type();
switch(out_type)
{
case S32: ...
case S64: ...
default: skip
}
```

compiler/luci/pass/src/FoldShapePass.test.cpp
miusic: `FromT` denotes the type of input, and `ToT` denotes the type of output

As I see it now, it looks possible to remove `FromT` and it looks more clear to replace `ToT` as `OutType`

compiler/luci/pass/src/FoldShapePass.test.cpp
miusic: Sorry but I failed to catch the point..

Did you mean that adding an `[Input]` node before `[Shape]` node like below, am I right?

```cpp
 *    BEFORE
 * 
 *                [Input]
 *                   |
 *      [Input]   [Shape]
 *           \    /
 *            [Add]
 *              |
 *           [Output]
```

seanshpark: I see input of `Shape` is `Const` if I am not wrong.

compiler/luci/pass/src/FoldShapePass.cpp
miusic: > have you checked with rank 0 constant?

The test case `rank_mismatch_NEG` in `FoldShapePass.test.cpp` is for this case!

> can `rank` be minus?

No.. AFAIK, `rank` can be a non-negative value (zero or positive integer)

Then, it might be fine just check if the `rank` is zero or not

compiler/luci/pass/src/FoldShapePass.cpp
miusic: Fine! I'll update this part :)

> Usually we add assert check for the variable just after declaration for condition check.

I agree this idea

compiler/luci/pass/src/FoldShapePass.cpp
seanshpark: > The test case rank_mismatch_NEG in FoldShapePass.test.cpp is for this case!

Not this.
Please check with TF or pytorch and create a rank 0 constant and check how the frameworks act.
Please check with recipie/tflchef with rank 0 constant if this is possible or not.

miusic: Ok, thank you for your advice!

I'll search about that, and check again! :)

miusic: > You may have to dig to "what is rank 0 constant?"

What I found is that the constant whose rank is 0 represents scalar value

> Please check with TF or pytorch and create a rank 0 constant and check how the frameworks act.

I created a scalar value with simple tensorflow code like below, and printed the value to check its rank

```python
import tensorflow as tf

def main():
  const = tf.constant(1.0)
  rank = tf.rank(const)
  print("constant -> ", const)
  print("rank -> ", rank)
  
  return 0

if __name__ == '__main__':
  main()
```
```bash
constant ->  tf.Tensor(1.0, shape=(), dtype=float32)
rank ->  tf.Tensor(0, shape=(), dtype=int32)
```

Through this result, I concluded that `rank 0 constant` is a scalar

When I run the code a constant tensor, the result shows the rank is 1

```python
import tensorflow as tf

def main():
  const = tf.constant([1.0])
  rank = tf.rank(const)
  print("constant -> ", const)
  print("rank -> ", rank)
  
  return 0

if __name__ == '__main__':
  main()
```
```bash
constant ->  tf.Tensor([1.], shape=(1,), dtype=float32)
rank ->  tf.Tensor(1, shape=(), dtype=int32)
```

---
> Please check with recipie/tflchef with rank 0 constant if this is possible or not.

And then, I made a test model with rank 0 constant with below recipe

```bash
operand {
  name: "input1"
  type: FLOAT32
  shape {
  }
}
operand {
  name: "input2"
  type: FLOAT32
  shape {
  }
  filler {
    tag: "explicit"
    arg: "20"
  }
}
operand {
  name: "out_shape1"
  type: INT32
  shape {
  }
}
operand {
  name: "output"
  type: INT32
  shape {
  }
}
operation {
  type: "Shape"
  shape_options {
    out_type: INT32
  }
  input: "input1"
  output: "out_shape1"
}
operation {
  type: "Add"
  input: "out_shape1"
  input: "input2"
  output: "output"
}
input: "input1"
input: "input2"
output: "output"
```

rank0

It was possible to make a model with rank 0 constant, but it was failed to fold the `Shape` op due to the condition below (rank is 0)

```cpp
if (rank <= 0)
```

miusic: If I change the condition to accept folding when the input is rank 0 constant like below, the result after folding `Shape` becoms `[0]` (rank 1, one-dimensional tensor with value as 0)

```cpp
if (rank < 0)
```

but I think this result is not reasonable, so it should be rejected to fold when the input is rank 0 constant
(because the scalar values have no shape, it looks strange that the output of the `Shape` has a value)

seanshpark: > ... , so it should be rejected to fold when the input is rank 0 constant

How does TF or pytorch act?

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+  auto input_node = loco::must_cast<luci::CircleNode *>(shape->input());
+  auto name = input_node->name();
+  assert(name.length() > 0);
+  auto shape_status = input_node->shape_status();
+  if (shape_status != luci::ShapeStatus::VALID)
+    return nullptr;
+  auto rank = input_node->rank();
+  if (rank == 0)
+    return nullptr;
```
seanshpark, line 35: let's check this how TF or pytorch act.

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,115 @@
...
+ *        [Input]
+ *           |
+ *        [Shape]
+ *           |
+ *        [Output]
+ *
+ * AFTER
+ *
+ *      [CircleConst]
+ *
```
seanshpark, line 86: where did `[Input]` and `[Output]` go?
Q) what do you mean by `[Input]` and `[Output]` ? or, what are these?

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,115 @@
...
+        case loco::DataType::S32:
+          if (fold_shape<loco::DataType::S32>(shape))
+            changed = true;
+          break;
+        case loco::DataType::S64:
+          if (fold_shape<loco::DataType::S64>(shape))
+            changed = true;
+          break;
+        default:
+          changed = false;
```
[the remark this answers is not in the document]
seanshpark: I think you don't understand how `run()` methods work.
Please explain after you look in to other Passes and the purpose of them.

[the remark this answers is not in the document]
miusic: > Why have to changed to `changed = false;` ?

My intention was just re-confirm the value of `changed` as it was declared at first (`bool changed = false;`)

If `Shape` op was successfully folded and graph changed, then it returns `true`
or if it was failed to fold the graph (for some reason...), then it returns `false`

It was my intention, but the code what I wrote was wrong... 😥

---
> Please explain after you look in to other Passes and the purpose of them.

I understanded the way how `run()` works as like below

1. Find the node traversing the entire graph to apply the pass (In this case, it becomes `Shape` node)
2. If find the node, try to apply the pass (Here, it is folding the operation)
3. If the pass was applied without problem, then the variable `changed` is marked as `true`
4. If the pass was not applied to any of the nodes, the variable `changed` remain as `false`

---
I've missed that there is a for-loop and the final value of the `changed` should be the `OR` of the result from each node 😅

This part should be changed like below!
```cpp
default:
  break;
```

compiler/luci/pass/src/FoldShapePass.cpp
miusic: > Q) what do you mean by `[Input]` and `[Output]` ? or, what are these?

I wanted to say that there is a tensor as an `[Input]` of `Shape` (that has a static shape... which can be folded as a constant), and then it returns a tensor as an `[Output]` (rank 1 tensor whose values are the each dimension of input tensor)

And after the `Shape` is folded as a constant, the `[Output]` will be replaced as a `[CircleConst]`

This is what I wanted to say... but the diagram seems insufficient to explain it 🤔


Is it more proper to comment like below..?

```cpp
 * BEFORE
 *
 *           |
 *        [Shape]
 *           |
 *
 * AFTER
 *
 *      [CircleConst]
```

`|` above the `[Shape]` is an input tensor, and `|` below the `[Shape]` is an output tensor

seanshpark: `[CircleSomething]` in the diagram is intended to express the circle node IR.
If you used `[CircleConst]` then it would be the `CircleConst` IR node.
Using `[Shape]`, not `[CircleShape]` would not follow this and also 
you have two different expression rules in your diagram.

Using `[CircleNode]` is used as don't care about specific IR but it is a circle IR node.

I think you may have copied `[Input]` and `[Output]` from some where else...
In circle IR, there is `CircleInput` and `CircleOutput` to express inputs and outputs of the graph. I was curious if you knew this when adding the `[Input]` and `[Output]` in the graph.

> where did [Input] and [Output] go?

This question is about why are replaced nodes not in the diagram.
To be exact, all the pass don't remove the node, just changes the connection.
If you look in to some of the graphs, but not all, there are diagrams that shows dangling replaced nodes. This is to express nature of the Pass. Real deletion is done in some other Pass, for that purpose.

miusic: First, I truley appreciate to your kind response :)

>In circle IR, there is `CircleInput` and `CircleOutput` to express inputs and outputs of the graph. I was curious if you knew this when adding the `[Input]` and `[Output]` in the graph.

I didn't know... I just thought that the `Input`/`Output` means `before`/`after` node of `Shape` operation respectively (Input of `Shape` and output of `Shape`)

> To be exact, all the pass don't remove the node, just changes the connection.

Then I guess, according to this, below diagram is more exact to explain the behavior of this pass. Am I right..?

```cpp
 * BEFORE
 * 
 *     [CircleInput]
 *           |
 *        [Shape]
 *           |
 *     [CircleOutput]
 *
 * AFTER
 *
 *     [CircleInput]
 *
 *     [CircleConst]
 *           |
 *     [CircleOutput]
 *
```

---
> If you look in to some of the graphs, but not all, there are diagrams that shows dangling replaced nodes.

Yes, I can find it in `FuseAddWithTConvPass.cpp`

compiler/luci/pass/src/FoldShapePass.cpp
miusic: I think that you said this with same intention of the comment from https://github.com/Samsung/ONE/pull/12478#discussion_r1465561647
So I reply here together! 😃

According to the annotation from `tf.shape`, it returns an empty vector

>   `tf.shape` returns a 1-D integer tensor representing the shape of `input`.
  For a scalar input, the tensor returned has a shape of (0,) and its value is
  the empty vector (i.e. []).

So if we accept the scalar as a valid input, it seems more reasonable to return an empty vector
But I'm not sure if it is possible...

compiler/luci/pass/src/FoldShapePass.cpp
seanshpark: If this `CircleInput` or `CircleOutput` is not a requirement, it would be better to use just `CircleNode` as these can be any kind of circle nodes.

compiler/luci/pass/src/FoldShapePass.cpp
seanshpark:
```cxx
  auto rank = input_node->rank();
  if (rank == 0)
    return nullptr;
```cxx
->
```
  auto rank = input_node->rank();
  // support rank == 0 when possible
  if (rank == 0)
    return nullptr;
```

File: compiler/luci/pass/src/FoldShapePass.cpp [hunk 20]
```diff
@@ -0,0 +1,119 @@
...
+ *
+ *     [CircleNode]
+ *           |
+ *     [CircleShape]
+ *           |
+ *     [CircleNode]
+ *
+ * AFTER
+ *
+ *     [Circlenode]
```
seanshpark, line 85:
```suggestion
 *     [CircleNode]
```

compiler/luci/pass/src/FoldShapePass.cpp
```diff
@@ -0,0 +1,119 @@
...
+ *
+ *     [CircleConst]
+ *           |
+ *     [CircleNode]
```
seanshpark, line 89:
```
 * AFTER
 *
 *     [CircleConst]  [CircleNode]
 *           |
 *      [CircleNode]
 *
```

seanshpark: Input of `CircleReshape`, `CircleNode` maybe connected to other nodes.
In this diagram, it would be better to draw like above

compiler/luci/pass/src/FoldShapePass.test.cpp
```diff
@@ -0,0 +1,135 @@
...
+ *              |
+ *           [Output]
+ *
+ *    AFTER
+ *
+ *      [Input]   [Const]
+ *           \    /
+ *            [Add]
+ *              |
+ *           [Output]
```
seanshpark, line 74: `.cpp` used `Circle` prefix. It would be better to match naming convention.

compiler/luci/pass/src/FoldShapePass.cpp
miusic: > // support rank == 0 when possible

I understood this comment as a `TODO` item.. am I right?

seanshpark: yup

compiler/luci/pass/src/FoldShapePass.test.cpp
```diff
@@ -0,0 +1,135 @@
...
+  luci::CircleShape *_shape = nullptr;
+  luci::CircleConst *_x = nullptr;
+};
+
+/**
+ *  Graph that has a Shape Op
+ *
+ *    BEFORE
+ *
+ *    [CircleInput]   [CircleShape]
```
seanshpark, line 62: `[CircleShape]` has a `CircleConst` as input

compiler/luci/pass/src/FoldShapePass.test.cpp
```diff
@@ -0,0 +1,135 @@
...
+} // namespace
+
+TEST(FoldShapePassTest, name)
+{
+  luci::FoldShapePass pass;
+  auto const name = pass.name();
+  ASSERT_NE(nullptr, name);
+}
+
+TEST_F(FoldShapeGraphTest, fold_shape)
```
seanshpark, line 94: how about `FoldShapePassGraphTest` ?

compiler/luci/pass/src/FoldShapePass.test.cpp
miusic: I think that you want to say same thing from https://github.com/Samsung/ONE/pull/12478#discussion_r1462870644

I think what you said means the tensor whose shape is `[1, 8, 8, 64]` is input of `[CircleShape]` (just in this test code..!)

This is true, but I'm bit confused because it is hard to decide that which diagram is more proper to explain the graph

I think there are two possible options. Please let me know how do you think about it 😄

---
### option 1
```cpp
/**
 *  Graph that has a Shape Op
 *
 *    BEFORE
 *
 *                    [CircleConst]
 *                          |
 *    [CircleInput]   [CircleShape]
 *              \      /
 *             [CircleAdd]
 *                 |
 *            [CircleOutput]
 *
 *    AFTER
 *
 *    [CircleInput]   [CircleConst]
 *              \      /
 *             [CircleAdd]
 *                 |
 *            [CircleOutput]
 *
 */
```

---
### option 2
```cpp
/**
 *  Graph that has a Shape Op
 *
 *    BEFORE
 *
 *             [CircleInput]
 *                 |      \
 *                 |    [CircleShape]
 *                 |      /
 *             [CircleAdd]
 *                 |
 *            [CircleOutput]
 *
 *    AFTER
 *
 *    [CircleInput]   [CircleConst]
 *              \      /
 *             [CircleAdd]
 *                 |
 *            [CircleOutput]
 *
 */
```

---
IMHO, second options looks more reasonable due to below reaseons

- The input of `Shape` and `Add` is identical (at here, `[1, 8, 8, 64]`)
- According to https://github.com/Samsung/ONE/pull/12478#discussion_r1465910319, pass itself does not remove any node
  > To be exact, all the pass don't remove the node, just changes the connection.

  But, if we choose first option, the `[CircleShape]` node disappear
  Or, to avoid this, the `[CircleShape]` node should be marked as dangling node, but it looks bit strange (because `[CircleShape]` was folded and was replaced as `[CircleConst]`...) and I think that it can cause confusion


Consequently, I updated the comment as following the second option now
But if there is any error or something I misunderstood, please let me know :)

compiler/luci/pass/src/FoldShapePass.test.cpp
miusic: It looks more clear! Updated :)

compiler/luci/pass/src/FoldShapePass.test.cpp
miusic: replied at https://github.com/Samsung/ONE/pull/12478#discussion_r1467279976 together! :D

compiler/luci/pass/src/FoldShapePass.test.cpp
seanshpark: Q) how can the graph of test, `FoldShapeGraph`, you wrote can have two options?

miusic: Actually, that is the confusing point to me... 😥

In this test, the input of the graph is constant, and it works as a input of `CircleShape`

So I thought like that the input of the `[CircleShape]` can be marked as `[CircleInput]` and `[CircleConst]` both, thus I thought that there are two options

To **option 1**, the input of `[CircleShape]` was marked as `[CircleConst]`
To **option 2**, the input of `[CircleShape]` was marked as `[CircleInput]`

but I'm not sure which one is right...

seanshpark: > To option 1, the input of [CircleShape] was marked as [CircleConst]

What do you mean by `marked` ?

> To option 2, the input of [CircleShape] was marked as [CircleInput]

What do you mean by `[CircleInput]` ?

miusic: > Consequently, I updated the comment as following the second option now

And I've pushed the code without the comment change by mistake 😅
It is updated now!

seanshpark: I think you don't understand the diagram in the comment.

seanshpark (approved):
LGTM

compiler/luci/pass/src/FoldShapePass.test.cpp
miusic: I've totally misunderstood about it..! 😢

(For note)
After offline talk with @seanshpark, I understood what the comments mean

> `[CircleSomething]` in the diagram is intended to express the circle node IR.

Like already said in https://github.com/Samsung/ONE/pull/12478?notification_referrer_id=NT_kwDOAq4Po7M5MDkzNDY2MDgwOjQ0OTYxNjk5#discussion_r1465910319, `[CircleSomething]` is a Circle IR node, and it is implemented as a class at the code

It means that there are a classes named `CircleInput`/`CircleOutput`/`CircleShape`/`CircleConst` and so on

So the nodes which were commented here represent each correspond node for the test graph!

jinevening (approved):
LGTM

Issue #12053, opened 2023-11-17
Introduce a new tool to generate h5 file (quantization dataset) from numpy files


jinevening:
## What

Let's introduce a new tool to pack numpy files to h5-format quantization dataset.

## Why

There is a similar tool, `rawdata2hdf5`, but users often try to convert numpy files rather than raw data.

So, it would be helpful to provide a new tool e.g., `numpy2hdf5`.

jinevening:
Here is a script I locally used.

```python
import h5py as h5
import numpy as np
import argparse
import glob
import os

def get_parser():
    """Create and return given the argument parser"""
    parser = argparse.ArgumentParser(
        description='command line tool to convert numpy files to hdf5 file')
    parser.add_argument(
        "-i",
        "--input_dir",
        type=str,
        help=
        "Path to the directory where numpy files are saved",
        required=True)
    parser.add_argument(
        "-o",
        "--output_path",
        type=str,
        help="Path to the output hdf5 file.",
        required=True)
    return parser

def create_hdf5(input_dir, output_path):
    """Create the hdf5 file using raw data files listed in input_dir"""
    h5_file = h5.File(output_path, 'w')
    group = h5_file.create_group("value")

    numpy_paths = glob.glob(os.path.join(input_dir, '*'))

    num_converted = 0 
    for numpy_path in numpy_paths:
        numpy_data = np.load(numpy_path)
        sample = group.create_group(str(num_converted))
        sample.create_dataset('0', data=numpy_data)
        num_converted += 1

    h5_file.close()

    print("Numpy data have been packaged to " + output_path)
    print("Number of packaged data: " + str(num_converted))

def main():
    parser = get_parser()

    args = parser.parse_args()

    create_hdf5(args.input_dir, args.output_path)

if __name__ == '__main__':
    main()
```

mhs4670go:
@jinevening Maybe we can introduce a new module for data format conversion where the scripts are all there. Or we could just one single driver for them. What's your take on this?

jinevening:
That seems a good idea. It'll be good to introduce a unified driver (ex: `create-h5-dataset`), which combines `rawdata2hdf5` and `numpy2hdf5`.

seanshpark:
As I understand these hdf5 files are for quantization. I hope `quantization` or abbreviation word is used, then that can help understand it's purpose.

jinevening:
How about `one-create-quant-dataset`? It would be good to sync the prefix (`one-`) with existing one cli.

[...]

PR #12479, opened 2024-01-16
[rawdata2hdf5] Add warning to deprecate rawdata2hdf5

Seunghui98:
This commit adds warning to deprecate rawdata2hdf5.

---
Related Issue: https://github.com/Samsung/ONE/issues/12053
Draft PR: https://github.com/Samsung/ONE/pull/12186

### DIFF
--- a/compiler/rawdata2hdf5/rawdata2hdf5
+++ b/compiler/rawdata2hdf5/rawdata2hdf5
@@ -24,6 +24,7 @@ import numpy as np
 import argparse
 import glob
 import os
+import warnings
 
 
 def get_parser():
@@ -105,6 +106,7 @@ def create_hdf5(data_list, output_path):
 
 
 def main():
+    warnings.warn("rawdata2hdf5 will be deprecated. Please use one-create-quant-dataset.");
     parser = get_parser()
 
     args = parser.parse_args()


compiler/rawdata2hdf5/rawdata2hdf5
```diff
@@ -105,6 +106,7 @@ def create_hdf5(data_list, output_path):
 
 
 def main():
+    warnings.warn("[rawdata2hdf5] section will be deprecated. Please use [one-create-quant-dataset] section.");
```
seanshpark, line 109: Q) what is `[rawdata2hdf5] section` ?

Seunghui98: @seanshpark 

> Q) what is [rawdata2hdf5] section ?

Simply, it means the tool of `rawdata2hdf5`. 😅
I was thinking about what would be better for this warning message, 
and I used a message referring to this. --- > [reference](https://github.com/Samsung/ONE/blob/c91b55370e28993569059233f744a0085d0dbbea/compiler/one-cmds/onelib/CfgRunner.py#L66)


Is it better to use the word `tool` than the word `section`?

seanshpark: @jinevening , any comments?

seanshpark: > and I used a message referring to this. 

do you understand the word `section` from the reference?

jinevening: @Seunghui98  `rawdata2hdf5` can be called in the command line, so `section` seems not appropriate. Could you update the message as below?

```suggestion
    warnings.warn("rawdata2hdf5 will be deprecated. Please use one-create-quant-dataset.");
```

Seunghui98: @jinevening 
Thank you for your comment :)

seanshpark: FYI `section` came from https://en.wikipedia.org/wiki/INI_file#Sections

jinevening (approved):
LGTM

PR #12487, opened 2024-01-17
[onert] Apply ReLU6Grad to ElementwiseActivationLayer

zetwhite:
This PR applies ReLU6Grad to ElementwiseActivationLayer.

issue : https://github.com/Samsung/ONE/issues/12388 
draft : https://github.com/Samsung/ONE/pull/12395

### DIFF
--- a/runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
+++ b/runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
@@ -19,6 +19,7 @@
 #include "OperationUtils.h"
 
 #include <cker/train/operation/ReLU.h>
+#include <cker/train/operation/ReLU6.h>
 
 namespace onert
 {
@@ -54,22 +55,31 @@ void ElementwiseActivationLayer::configure(const IPortableTensor *input, IPortab
     case ElementwiseActivationType::kReLU:
       if (input->data_type() == OperandType::FLOAT32)
       {
-        if (alpha == std::numeric_limits<float>::infinity() && beta == 0.f)
+        if ((alpha == std::numeric_limits<float>::infinity() || alpha == 6.0f) && beta == 0.f)
         {
           cpu::ops::ElementwiseActivationLayer::configure(
             input, output, alpha, beta, cpu::ops::ElementwiseActivationType::kReLU);
 
-          _backward_kernel = [](const IPortableTensor *output, const IPortableTensor *incoming,
-                                IPortableTensor *outgoing) {
-            nnfw::cker::train::ReLUGrad(getShape(output), getBuffer<float>(output),
-                                        getShape(incoming), getBuffer<float>(incoming),
-                                        getShape(outgoing), getBuffer<float>(outgoing));
+          auto relu_cker = [&alpha]() {
+            if (alpha == std::numeric_limits<float>::infinity())
+              return nnfw::cker::train::ReLUGrad;
+            else if (alpha == 6.0f)
+              return nnfw::cker::train::ReLU6Grad;
+            else
+              throw std::runtime_error{"no supported relu kernel"};
+          }();
+
+          _backward_kernel = [relu_cker](const IPortableTensor *output,
+                                         const IPortableTensor *incoming,
+                                         IPortableTensor *outgoing) {
+            relu_cker(getShape(output), getBuffer<float>(output), getShape(incoming),
+                      getBuffer<float>(incoming), getShape(outgoing), getBuffer<float>(outgoing));
           };
         }
         else
         {
-          throw std::runtime_error("train ElementwiseActivationLayer : This layer does not "
-                                   "suppport other ReLU except for ReLU(0-inf)");
+          throw std::runtime_error(
+            "train ElementwiseActivationLayer : Unsupported ReLU activation type");
         }
       }
       else


runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
```diff
@@ -54,22 +55,31 @@ void ElementwiseActivationLayer::configure(const IPortableTensor *input, IPortab
...
-                                        getShape(incoming), getBuffer<float>(incoming),
-                                        getShape(outgoing), getBuffer<float>(outgoing));
+          auto relu_cker = [&alpha]() {
+            if (alpha == std::numeric_limits<float>::infinity())
+              return nnfw::cker::train::ReLUGrad;
+            else if (alpha == 6.0f)
+              return nnfw::cker::train::ReLU6Grad;
+            else
+              throw std::runtime_error{"no supported relu kernel"};
+          }();
```
zetwhite, line 70: (note) select kernel by alpha value

runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
```diff
@@ -54,22 +55,31 @@ void ElementwiseActivationLayer::configure(const IPortableTensor *input, IPortab
...
+            relu_cker(getShape(output), getBuffer<float>(output), getShape(incoming),
+                      getBuffer<float>(incoming), getShape(outgoing), getBuffer<float>(outgoing));
           };
         }
         else
         {
           throw std::runtime_error("train ElementwiseActivationLayer : This layer does not "
-                                   "suppport other ReLU except for ReLU(0-inf)");
+                                   "suppport other ReLU except for ReLU(0-inf) and ReLU6(0-6)");
         }
```
zetwhite, line 83: ### NOTE
# WIP

After This PR, the training feature supports `ReLU6`. (not fused one) 

I also checked loss value isn't different with tensorflow. 



* TensorFlow
```
Epoch 1/5
50/50 [==============================] - 0s 617us/step - loss: 0.0794 - mean_squared_error: 0.0794
Epoch 2/5
50/50 [==============================] - 0s 542us/step - loss: 0.0586 - mean_squared_error: 0.0586
Epoch 3/5
50/50 [==============================] - 0s 533us/step - loss: 0.0505 - mean_squared_error: 0.0505
Epoch 4/5
50/50 [==============================] - 0s 542us/step - loss: 0.0461 - mean_squared_error: 0.0461
Epoch 5/5
50/50 [==============================] - 0s 513us/step - loss: 0.0431 - mean_squared_error: 0.0431
```

* onert_train
```
- learning_rate   = 0.001
- batch_size      = 20
- loss_info       = {loss = mean squared error, reduction = sum over batch size}
- optimizer       = adam
========================
Epoch 1/5 - time: 3.351ms/step - loss: [0] 0.0794
Epoch 2/5 - time: 3.375ms/step - loss: [0] 0.0586
Epoch 3/5 - time: 3.409ms/step - loss: [0] 0.0505
Epoch 4/5 - time: 3.655ms/step - loss: [0] 0.0461
Epoch 5/5 - time: 3.915ms/step - loss: [0] 0.0431
```

runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
```diff
@@ -54,22 +55,31 @@ void ElementwiseActivationLayer::configure(const IPortableTensor *input, IPortab
...
+            else
+              throw std::runtime_error{"no supported relu kernel"};
+          }();
+
+          _backward_kernel = [relu_cker](const IPortableTensor *output,
+                                         const IPortableTensor *incoming,
+                                         IPortableTensor *outgoing) {
+            relu_cker(getShape(output), getBuffer<float>(output), getShape(incoming),
+                      getBuffer<float>(incoming), getShape(outgoing), getBuffer<float>(outgoing));
           };
```
zetwhite, line 77: (note) create `_backend_kernel`, by capturing selected `relu_cker`

ragmani (dismissed):
LGTM

runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
```diff
@@ -54,22 +55,31 @@ void ElementwiseActivationLayer::configure(const IPortableTensor *input, IPortab
...
         }
         else
         {
           throw std::runtime_error("train ElementwiseActivationLayer : This layer does not "
-                                   "suppport other ReLU except for ReLU(0-inf)");
+                                   "suppport other ReLU except for ReLU(0-inf) and ReLU6(0-6)");
```
glistening, line 82: (optional) It would be good not to list the supported ReLU types here. We need to update the message whenever supported list is updated. What about something like "Not supported ReLU type activation".

glistening (dismissed):
LGTM

jyoungyun (dismissed):
LGTM

runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
zetwhite: Looks better!
updated :)

ragmani (approved):

runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
```diff
@@ -54,22 +55,31 @@ void ElementwiseActivationLayer::configure(const IPortableTensor *input, IPortab
...
         {
           cpu::ops::ElementwiseActivationLayer::configure(
             input, output, alpha, beta, cpu::ops::ElementwiseActivationType::kReLU);
 
-          _backward_kernel = [](const IPortableTensor *output, const IPortableTensor *incoming,
-                                IPortableTensor *outgoing) {
-            nnfw::cker::train::ReLUGrad(getShape(output), getBuffer<float>(output),
-                                        getShape(incoming), getBuffer<float>(incoming),
-                                        getShape(outgoing), getBuffer<float>(outgoing));
+          auto relu_cker = [&alpha]() {
```
ragmani, line 63: `alpha` is a local variable. Please use capture by value.
```suggestion
          auto relu_cker = [alpha]() {
```

zetwhite: I thought reference is fine,
Because this lambda function is invoked right now - when it is defined. 
This lambda function is also used in only local!
Is there anything I miss-understood here?

Anyway, alpha is just float type, removing & is no problem with me.

runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
zetwhite, line 70:
```suggestion
          }();
```
lambda is invoked here. 
and it returns a function pointer to relu_cker.

runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
ragmani: Ah, I thought `relu_cker is a lambda, but it's not.

ragmani (approved):

runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
zetwhite: Ah, definitely.. this using is tricky. :sweat_smile:

jyoungyun (approved):
LGTM

runtime/onert/backend/train/ops/ElementwiseActivationLayer.cc
```diff
@@ -54,22 +55,31 @@ void ElementwiseActivationLayer::configure(const IPortableTensor *input, IPortab
...
+            if (alpha == std::numeric_limits<float>::infinity())
+              return nnfw::cker::train::ReLUGrad;
+            else if (alpha == 6.0f)
```
YongseopKim, line 66: Q. Is it okay to compare float value with float value? I'm not sure....

zetwhite: Could you explain a little more? 

At first glance, I failed to catch your worries.
I thought It is natural to compare one float with another.

glistening: Generally not, but some range of integer is exactly represented in float format.

For float, it has 24-bit significand. 6 is exactly representalbe.

To make sure,

https://en.wikipedia.org/wiki/Single-precision_floating-point_format

> Integers between 0 and 16777216 can be exactly represented

YongseopKim: Almost it returns true

zetwhite: Aha, I understood @YongseopKim's point.  
There is a floating-point (representataion) error issue.

YongseopKim: https://codereview.stackexchange.com/questions/275266/c-compare-floats

YongseopKim: FYI, this is just comment. Do not count me as the number of review.

glistening: > Almost it returns true but rarely could return false. If you already know this and then code it, it would be okay. But if not, we could consider it. 😀

I don't understand. For the case in range (0 and 16777216), it is always safe to compare as I understand. Maybe I don't understand your point yet.

ragmani: As far as I know, `(6.0f == 6.0f)` is always true. They have the same precision since they are `float` literal.
But when one of them is a double or a long double like `(6.0f == 6.0)`, I'm not sure if they have the same precision.

YongseopKim: > It seems that both exponent value and fraction value of 6.0 and 6.0f are the same.

This is my point.

ragmani: >> It seems that both exponent value and fraction value of 6.0 and 6.0f are the same.
> This is my point.

I updated about it in `https://github.com/Samsung/ONE/pull/12487#discussion_r1456812278`. In short, if comparing value is 6.0, there is no problem.

YongseopKim: And I already told the term `approximate equality`.

if we have a consensus, because of value 6, it is okay. I'm curious what sth happens if the floating point's precision is too big.
- https://beta.boost.org/doc/libs/1_68_0/libs/math/doc/html/math_toolkit/float_comparison.html

YongseopKim: My comment is not pointing out that this is a problem.

PR #12492, opened 2024-01-17
[onert] Extract fused activation backprop to OperationUtils

zetwhite:
This PR extracts fused activation backpropagation part to OperationUtils. 

issue : https://github.com/Samsung/ONE/issues/12388 
draft : https://github.com/Samsung/ONE/pull/12395

### DIFF
--- a/runtime/onert/backend/train/ops/ConvolutionLayer.cc
+++ b/runtime/onert/backend/train/ops/ConvolutionLayer.cc
@@ -134,21 +134,16 @@ void ConvolutionLayer::backwardFloat32()
 {
   // Calculate gradient for activation
   const IPortableTensor *backprop_act;
-  switch (_activation)
+  try
   {
-    case ir::Activation::NONE:
-      backprop_act = _back_prop_output;
-      break;
-    case ir::Activation::RELU:
-      nnfw::cker::train::ReLUGrad(getShape(_output), getBuffer<float>(_output),
-                                  getShape(_back_prop_output), getBuffer<float>(_back_prop_output),
-                                  getShape(_act_back_prop_output.get()),
-                                  getBuffer<float>(_act_back_prop_output.get()));
-      backprop_act = _act_back_prop_output.get();
-      break;
-    default:
-      throw std::runtime_error("train ConvolutionLayer: Unsupported activation type yet");
+    backprop_act =
+      backpropActivation(_activation, _output, _back_prop_output, _act_back_prop_output.get());
   }
+  catch (const std::exception &e)
+  {
+    throw std::runtime_error{"train ConvolutionLayer: " + std::string(e.what())};
+  }
+  assert(backprop_act != nullptr);
 
   // Initialize conv params for training kernels
   nnfw::cker::ConvParams conv_train_params;
--- a/runtime/onert/backend/train/ops/FullyConnectedLayer.cc
+++ b/runtime/onert/backend/train/ops/FullyConnectedLayer.cc
@@ -134,21 +134,16 @@ void FullyConnectedLayer::backwardFloat32()
 {
   // Calculate gradient for activation
   const IPortableTensor *backprop_act;
-  switch (_activation)
+  try
   {
-    case ir::Activation::NONE:
-      backprop_act = _back_prop_output;
-      break;
-    case ir::Activation::RELU:
-      nnfw::cker::train::ReLUGrad(getShape(_output), getBuffer<float>(_output),
-                                  getShape(_back_prop_output), getBuffer<float>(_back_prop_output),
-                                  getShape(_act_back_prop_output.get()),
-                                  getBuffer<float>(_act_back_prop_output.get()));
-      backprop_act = _act_back_prop_output.get();
-      break;
-    default:
-      throw std::runtime_error("train FullyConnectedLayer: Unsupported activation type yet");
+    backprop_act =
+      backpropActivation(_activation, _output, _back_prop_output, _act_back_prop_output.get());
   }
+  catch (const std::exception &e)
+  {
+    throw std::runtime_error{"train FullyConnectedLayer: " + std::string(e.what())};
+  }
+  assert(backprop_act != nullptr);
 
   // Initialize TransposeParams
   nnfw::cker::TransposeParams transpose_param;
--- a/runtime/onert/backend/train/ops/OperationUtils.cc
+++ b/runtime/onert/backend/train/ops/OperationUtils.cc
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2024 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "OperationUtils.h"
+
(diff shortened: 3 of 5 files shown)


runtime/onert/backend/train/ops/FullyConnectedLayer.cc
```diff
@@ -134,6 +134,18 @@ void FullyConnectedLayer::backwardFloat32()
...
   const IPortableTensor *backprop_act;
+  try
+  {
+    backprop_act =
+      backpropActivation(_activation, _output, _back_prop_output, _act_back_prop_output.get());
+  }
+  catch (const std::exception &e)
+  {
+    throw std::runtime_error{"train FullyConnectedLayer: " + std::string(e.what())};
+  }
```
zetwhite, line 145: caller part

runtime/onert/backend/train/ops/ConvolutionLayer.cc
```diff
@@ -134,21 +134,16 @@ void ConvolutionLayer::backwardFloat32()
 {
```
zetwhite, line 134: ### for reviewers

This PR is much easier to review with split diff view.

runtime/onert/backend/train/ops/OperationUtils.h
```diff
@@ -36,6 +36,31 @@ using cpu::ops::getNumberOfDimensions;
...
+ *                        In other words, outcoming gradient of current layer's acitvation
+ *                        If activation is NONE, this param isn't necessary
+ * @return tensor that holds backpropagate result of activation
+ *         If activation is NONE, just return input_backprop
+ */
+const IPortableTensor *backpropActivation(const ir::Activation &activation,
+                                          const IPortableTensor *output,
+                                          const IPortableTensor *input_backprop,
+                                          IPortableTensor *output_backprop = nullptr);
+
```
zetwhite, line 63: This is a kind of refactoring PR. 

I checked that `onert_train` doesn't affect with these changes, by making dummy models and check loss value aren't changed.

* Tested Dummy Model 1 
* Tested Dummy Model2

runtime/onert/backend/train/ops/OperationUtils.cc
```diff
@@ -0,0 +1,65 @@
...
+                                  getShape(input_backprop), getBuffer<float>(input_backprop),
+                                  getShape(output_backprop), getBuffer<float>(output_backprop));
+      break;
+    // TODO: Add ReLU6
+    default:
+      throw std::runtime_error("Unsupported activation type yet");
+  }
+  return output_backprop;
+}
+
```
zetwhite, line 61: #### For reviewer

This refactoring aims to make space - that helps add other activation kernels. 
Without these changes, we have to modify every layer to support new (fused) activation.

runtime/onert/backend/train/ops/OperationUtils.h
zetwhite: BTW, while doing this, I felt the necessity of an integrated test.

runtime/onert/backend/train/ops/ConvolutionLayer.cc
ragmani: Is there a difference from https://github.com/Samsung/ONE/pull/12492/files?

runtime/onert/backend/train/ops/FullyConnectedLayer.cc
```diff
@@ -134,21 +134,16 @@ void FullyConnectedLayer::backwardFloat32()
 {
   // Calculate gradient for activation
   const IPortableTensor *backprop_act;
-  switch (_activation)
+  try
```
ragmani, line 137: I'm not sure if adding a try block here just for a little better message. Try blocks affect performance.

runtime/onert/backend/train/ops/OperationUtils.h
YongseopKim: 👍

runtime/onert/backend/train/ops/ConvolutionLayer.cc
zetwhite: > Is there a difference from [#12492 (files)](https://github.com/Samsung/ONE/pull/12492/files)?

There is no difference! 
It just forces to have a split view. (not unified view) 
( It looks to cause some confusion. So I removed the line )

runtime/onert/backend/train/ops/FullyConnectedLayer.cc
YongseopKim: I agree with @ragmani .

YongseopKim: However, the message like `train FuulyConnectedLayer` is very useful for debug/devel, I think.

YongseopKim: Hmm... how to achieve two goals?

YongseopKim (approved):
LGTM 👍

runtime/onert/backend/train/ops/FullyConnectedLayer.cc
zetwhite: Yes.. This try-catch is only for adding `train ...Layer` in error message. 

`Try..Catch` also heart readability a bit. :thinking:

zetwhite: How about wrapping try{} .. in more higher level? 

Maybe here? 

### AS-IS 

https://github.com/Samsung/ONE/blob/e50a233e60eeaa9cf37ea1cc16eb25646b088911/runtime/onert/core/src/exec/train/TrainableFnSequence.cc#L34-L40 

### TO-BE 

```cc
//TrainableSequence.cc 
void TrainableFnSequence::backward(uint32_t training_step)
{
  for (auto it = _functions.rbegin(); it != _functions.rend(); ++it)
  {
    try{
      (*it)->backward();
    }
    catch{ 
       // print layer name and detailed message here 
    }   
}

Comment (ragmani) [reply runtime/onert/backend/train/ops/FullyConnectedLayer.cc:137]: I don't know whether higher level is more efficient. Rather, I think the currrent code is better in error handling.

Comment (zetwhite) [reply runtime/onert/backend/train/ops/FullyConnectedLayer.cc:137]: > I don't know whether higher level is more efficient.  

I thought the above approach because it could give some consistency in the error message format.
Like... Inside the `backward()`, Any function could throw an error in any way, and `catch{}` gives a kind of tag -  where it problem resides. 

Yes, As you pointed out, this way isn't efficient either. 

> Do you think it's necessary to add a try block just for branching error messages? 

I also like to remove try{} block. 
But I think adding a message to show which layer is problematic is necessary.
That message could help the user(developer) a lot - If a user finds a problematic layer(node) easily, They could take a walk-around - like removing a problematic layer from the model and retrying.

Comment (zetwhite) [reply runtime/onert/backend/train/ops/FullyConnectedLayer.cc:137]: > Is there any way to branch error message?

:thinking: I could make the function(`backpropActivation`) return nullptr when error occurs, instead of thorwing an error.
If so, caller should check the result value isn't nullptr; if result value is nullptr, caller might throw an error like before. 

Maybe caller part looks like this : 
```cpp
const IPortableTensor *backprop_act =  backpropActivation(_activation, _output, _back_prop_output, _act_back_prop_output.get());

if(backprop_act ==  nullptr)
   throw runtime_error{"train Convolution layer : error ... "};
```

ragmani: > I also like to remove try{} block.
But I think adding a message to show which layer is problematic is necessary.
That message could help the user(developer) a lot - If a user finds a problematic layer(node) easily, They could take a walk-around - like removing a problematic layer from the model and retrying.

I think developers can see which layer is problematic thought debugger. So, I'm not sure if it's worth adding a try bock.

@ragmani Could you share your opinion about this direction? :)

It looks better in performance. But, it seems unclear what message to write. I don't know there is another way to branch error messages. 🤔, Let's keep the current code and modify it later if there is a performance issue.

ragmani (approved):
LGTM

jyoungyun (approved):
LGTM

PR #12500, opened 2024-01-18
[onert-micro] Enable layer-wise hybrid FC

BalyshevArtem:
This commit enables layer-wise hybrid FC.

Need for some internal models.


SlavikMIPT (approved):
LGTM

PR #12509, opened 2024-01-22
[luci/pass] Revise ResolveCustomOpAddPass to process BroadcastTo

Seunghui98:
This will revise ResolveCustomOpAddPass to process built-in BroadcastTo. 
In addition, more unit tests were added to ResolveCustomOpAddPass to test BroadcastTo-AddV2 pattern.

---

### DIFF
--- a/compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
@@ -25,7 +25,7 @@
 namespace
 {
 
-/// @brief Returns the number of Custom("BroadcastTo") among node's inputs
+/// @brief Returns the number of BroadcastTo among node's inputs
 int32_t num_of_broadcast_to(const luci::CircleNode *node)
 {
   int32_t bt_cnt = 0;
@@ -40,6 +40,11 @@ int32_t num_of_broadcast_to(const luci::CircleNode *node)
         auto custom = loco::must_cast<luci::CircleCustom *>(inputOut->input());
         if (custom->custom_code() == "BroadcastTo")
           ++bt_cnt;
+        break;
+      }
+      case luci::CircleOpcode::BROADCAST_TO:
+      {
+        ++bt_cnt;
         break;
       }
       default:
@@ -65,6 +70,12 @@ int32_t get_broadcastTo_index_among_inputs_of(luci::CircleCustom *cop)
       if (broadcastTo->custom_code() == "BroadcastTo")
         return idx;
     }
+    else
+    {
+      auto broadcastTo = dynamic_cast<luci::CircleBroadcastTo *>(cop->inputs(idx));
+      if (broadcastTo)
+        return idx;
+    }
   }
 
   return -1;
@@ -72,7 +83,25 @@ int32_t get_broadcastTo_index_among_inputs_of(luci::CircleCustom *cop)
 
 // NOTE Broadcasting of input `Const` is skipped cause `Add` will do the broadcasting.
 // TODO Implement broadcasting to the `Const` input.
-/** BEFORE
+/**
+ * [Pattern1]
+ *  BEFORE
+ *                                  [CircleConst]
+ *                                        |
+ *        [CircleNode]         [BroadcastTo(Builtin)]
+ *              \                       /
+ *               \                     /
+ *                \                   /
+ *               [AddV2(CircleCustom)]
+ *  AFTER
+ *
+ *         [CircleConst]         [CircleNode]
+ *                   \           /
+ *                    \         /
+ *                    [CircleAdd]
+ *
+ * [Pattern2]
+ *  BEFORE
  *                                  [CircleConst]
  *                                        |
  *        [CircleNode]         [BroadcastTo(CircleCustom)]
@@ -94,19 +123,41 @@ bool resolve_with_BroadcastTo(luci::CircleCustom *addv2)
   if (broadcastTo_idx == -1)
     return false;
 
-  auto input = loco::must_cast<const luci::CircleCustomOut *>(addv2->inputs(broadcastTo_idx));
-  auto broadcastTo = loco::must_cast<luci::CircleCustom *>(input->input());
-
   auto name = addv2->name();
   assert(name.length() > 0);
+
+  auto input = loco::must_cast<const luci::CircleNode *>(addv2->inputs(broadcastTo_idx));
+  luci::CircleNode *bc = nullptr;
+  loco::Node *bc_input = nullptr;
+
+  switch (input->opcode())
+  {
+    case luci::CircleOpcode::BROADCAST_TO:
+    {
+      auto broadcastTo = loco::must_cast<luci::CircleBroadcastTo *>(addv2->inputs(broadcastTo_idx));
+      bc = broadcastTo;
+      bc_input = broadcastTo->input();
+      break;
+    }
+    case luci::CircleOpcode::CIRCLECUSTOMOUT:
+    {
+      auto inputOut =
+        loco::must_cast<const luci::CircleCustomOut *>(addv2->inputs(broadcastTo_idx));
+      auto braodcastTo = loco::must_cast<luci::CircleCustom *>(inputOut->input());
+      bc = braodcastTo;
+      bc_input = braodcastTo->inputs(0);
+      break;
+    }
+    default:
+      return false;
+  }
 
   auto add = addv2->graph()->nodes()->create<luci::CircleAdd>();
   add->fusedActivationFunction(luci::FusedActFunc::NONE);
   add->x(addv2->inputs(1 - broadcastTo_idx));
-  add->y(broadcastTo->inputs(0));
+  add->y(bc_input);
   add->name(name + "/Add");
-  luci::add_origin(
-    add, luci::composite_origin({luci::get_origin(broadcastTo), luci::get_origin(addv2)}));
+  luci::add_origin(add, luci::composite_origin({luci::get_origin(bc), luci::get_origin(addv2)}));
 
   auto customOut = loco::succs(addv2);
   assert(customOut.size() == 1);
--- a/compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
+++ b/compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
@@ -17,13 +17,140 @@
 #include "luci/Pass/ResolveCustomOpAddPass.h"
 
 #include <gtest/gtest.h>
+#include <luci/test/TestIOGraph.h>
+
+#include <luci/IR/CircleNodes.h>
+
+using namespace luci::test;
 
 namespace
 {
 
+/**
+ *   Test graph with Custom(AddV2) to resolve
+ *
+ * [Pattern 1]
+ *        [Input]   [BroadcastTo]
+ *             \         /
+ *           [Custom(AddV2)]
+ *                  |
+ *             [CustomOut]
+ *                  |
+ *               [Output]
+ *
+ * [Pattern 2]
+ *        [Input]   [Custom(BroadcastTo)]
+ *             \         /
+ *           [Custom(AddV2)]
+ *                  |
+ *             [CustomOut]
+ *                  |
+ *               [Output]
+ */
+class BroadcastToAddGraphlet
+{
+public:
+  BroadcastToAddGraphlet() = default;
+
+public:
+  void init(loco::Graph *g)
+  {
+    _addV2 = g->nodes()->create<luci::CircleCustom>(2, 1);
+    _addV2->custom_code("AddV2");
+    _addV2->shape({2, 3});
+    _addV2->dtype(loco::DataType::FLOAT32);
+    _addV2->name("addV2");
+
+    // Const as BroadcastTo input
+    _broadcastTo_input = g->nodes()->create<luci::CircleConst>();
+    _broadcastTo_input->dtype(loco::DataType::FLOAT32);
+    _broadcastTo_input->shape({1, 3});
+    _broadcastTo_input->size<loco::DataType::FLOAT32>(3);
+    _broadcastTo_input->at<loco::DataType::FLOAT32>(0) = 1;
+    _broadcastTo_input->at<loco::DataType::FLOAT32>(1) = 2;
+    _broadcastTo_input->at<loco::DataType::FLOAT32>(2) = 3;
+
+    // Const as BroadcastTo shape
+    auto broadcastTo_shape = g->nodes()->create<luci::CircleConst>();
+    broadcastTo_shape->dtype(loco::DataType::S32);
+    broadcastTo_shape->shape({2});
+    broadcastTo_shape->size<loco::DataType::S32>(2);
+    broadcastTo_shape->at<loco::DataType::S32>(0) = 2;
+    broadcastTo_shape->at<loco::DataType::S32>(1) = 3;
+
+    _custom_broadcastTo = g->nodes()->create<luci::CircleCustom>(2, 1);
+    _custom_broadcastTo->custom_code("BroadcastTo");
+    _custom_broadcastTo->dtype(loco::DataType::FLOAT32);
+    _custom_broadcastTo->shape({2, 3});
+    _custom_broadcastTo->name("BroadcastTo");
+
+    _custom_broadcastTo->inputs(0, _broadcastTo_input);
+    _custom_broadcastTo->inputs(1, broadcastTo_shape);
+
+    _custom_broadcastTo_out = g->nodes()->create<luci::CircleCustomOut>();
+    _custom_broadcastTo->custom_code("BroadcastTo");
+    _custom_broadcastTo_out->shape({2, 3});
+    _custom_broadcastTo_out->dtype(loco::DataType::FLOAT32);
+    _custom_broadcastTo_out->index(0);
+    _custom_broadcastTo_out->input(_custom_broadcastTo);
+
+    _builtin_broadcastTo = g->nodes()->create<luci::CircleBroadcastTo>();
+    _builtin_broadcastTo->dtype(loco::DataType::FLOAT32);
+    _builtin_broadcastTo->name("BroadcastTo");
+
+    _builtin_broadcastTo->input(_broadcastTo_input);
+    _builtin_broadcastTo->shape(broadcastTo_shape);
+
+    _addV2_out = g->nodes()->create<luci::CircleCustomOut>();
+    _addV2_out->shape({2, 3});
+    _addV2_out->dtype(loco::DataType::FLOAT32);
+    _addV2_out->index(0);
+    _addV2_out->input(_addV2);
+  }
+
+public:
+  luci::CircleCustom *addV2() { return _addV2; }
+  luci::CircleBroadcastTo *builtin_broadcastTo() { return _builtin_broadcastTo; }
+
+protected:
+  luci::CircleCustom *_addV2 = nullptr;
+  luci::CircleCustomOut *_addV2_out = nullptr;
+  luci::CircleCustom *_custom_broadcastTo = nullptr;
+  luci::CircleBroadcastTo *_builtin_broadcastTo = nullptr;
+  luci::CircleCustomOut *_custom_broadcastTo_out = nullptr;
+  luci::CircleConst *_broadcastTo_input = nullptr;
+};
+
+class BroadcastToAddV2Graph : public TestIGraphlet,
+                              public TestOsGraphlet<1>,
+                              public BroadcastToAddGraphlet
+{
+public:
+  BroadcastToAddV2Graph() = default;
+
+  void init(const bool &isCustomBroadcastTo)
+  {
+    TestIGraphlet::init(g(), {2, 3});
+    TestOsGraphlet<1>::init(g(), {{2, 3}});
+    BroadcastToAddGraphlet::init(g());
+
+    // connect Input and Output to test graph
+    _addV2->inputs(0, input());
+
+    if (isCustomBroadcastTo)
+      _addV2->inputs(1, _custom_broadcastTo_out);
+    else
+      _addV2->inputs(1, _builtin_broadcastTo);
+
+    _addV2_out->input(_addV2);
+    output(0)->from(_addV2_out);
+  }
+};
+
 class ResolveCustomOpAddPassTest : public ::testing::Test
 {
 public:
+  BroadcastToAddV2Graph _g;
   luci::ResolveCustomOpAddPass _pass;
 };
 
@@ -33,4 +160,70 @@ TEST_F(ResolveCustomOpAddPassTest, name)
 {
   auto const name = _pass.name();
   ASSERT_NE(nullptr, name);
+}
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_CustomBroadcastTo)
+{
+  _g.init(true);
+
+  // check if Custom(AddV2) exists before the pass
+  auto addV2_out = dynamic_cast<luci::CircleCustomOut *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, addV2_out);
+  auto addV2 = dynamic_cast<luci::CircleCustom *>(addV2_out->input());
+  EXPECT_NE(nullptr, addV2);
+  EXPECT_TRUE("AddV2" == addV2->custom_code());
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  // check if Custom(AddV2) is converted to Add
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  // check if Custom(BroadcastTo) is removed
+  auto input_y = dynamic_cast<luci::CircleConst *>(add->y());
+  EXPECT_NE(nullptr, input_y);
+}
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_BuiltinBroadcastTo)
+{
+  _g.init(false);
+
+  // check if Custom(AddV2) exists before the pass
+  auto addV2_out = dynamic_cast<luci::CircleCustomOut *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, addV2_out);
+  auto addV2 = dynamic_cast<luci::CircleCustom *>(addV2_out->input());
+  EXPECT_NE(nullptr, addV2);
+  EXPECT_TRUE("AddV2" == addV2->custom_code());
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  // check if Custom(AddV2) is converted to Add
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  // check if BroadcastTo is removed
+  auto input_y = dynamic_cast<luci::CircleConst *>(add->y());
+  EXPECT_NE(nullptr, input_y);
+}
+
+TEST_F(ResolveCustomOpAddPassTest, wrong_custom_code_NEG)
+{
+  _g.init(false);
+
+  _g.addV2()->custom_code("UNSUPORTED_CUSTOM_CODE");
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(false, ret);
+}
+
+TEST_F(ResolveCustomOpAddPassTest, wrong_input_type_NEG)
+{
+  _g.init(false);
+
+  _g.builtin_broadcastTo()->dtype(loco::DataType::BOOL);
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(false, ret);
 }


mhs4670go:
There are two things that I think you should know when revising this pass.

1. `CircleXXXOut` node

Nodes whose suffix are "Out" is for the nodes like `CircleCustom`, `CircleSplitV`, etc where the number of their outputs is not fixed. You can check their builder class if `build_out` is implemented. The "Out" nodes is unique for each node. For example, `CircleCustom` has `CircleCustomOut` outputs, and `CircleSplitV` has `CircleSplitVOut` for its outputs.

Considering this fact, below comments are not right. Because "Builtin" BroadcastTo operator doesn't need `CircleCustomOut` node.

```cxx
/** BEFORE
 *                                  [CircleConst]
 *                                        |
 *        [CircleNode]         [BroadcastTo(CircleCustom or Builtin)]
 *              \                         |
 *               \                [CircleCustomOUt]
 *                \                   /
 *               [AddV2(CircleCustom)]
 *
```

Then, why did this pass return true?

2. Pass only resolves Custom AddV2 node

Current pass returns true even though resolving BroadcastTo operator failed. That is why the name of this pass is ResolveCustomOp"Add".
```cxx
  if (resolve_with_BroadcastTo(addv2))
    return true;

  auto name = addv2->name();
  assert(name.length() > 0);

  auto add = addv2->graph()->nodes()->create<luci::CircleAdd>();
  add->fusedActivationFunction(luci::FusedActFunc::NONE);
  add->x(addv2->inputs(0));
  add->y(addv2->inputs(1));
  add->name(name + "/Add");
  luci::add_origin(add, luci::get_origin(addv2));

  auto customOut = loco::succs(addv2);
  assert(customOut.size() == 1);
  replace(*customOut.begin()).with(add);

  return true;
```

You can see that even though `resolve_with_BroadcastTo` function returns `false`, custom AddV2 is replaced by built-in Add node.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,10 +17,200 @@
...
+  _g.addV2()->inputs(1, _g.builtin_broadcastTo());
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  auto broadcastTo = dynamic_cast<luci::CircleBroadcastTo *>(add->y());
+  EXPECT_NE(nullptr, broadcastTo);
```
mhs4670go, line 184: The BroadcastTo operator has to be gone according to your comments.

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -46,12 +46,35 @@ int32_t get_broadcastTo_index_among_inputs_of(luci::CircleCustom *cop)
         bc_idx = static_cast<int32_t>(idx);
       }
     }
+    else
+    {
+      auto broadcastTo = dynamic_cast<luci::CircleBroadcastTo *>(cop->inputs(idx));
+      if (broadcastTo)
+        return idx;
```
seanshpark, line 53: how does `This function assumes there is only one BroadcastTo node among its inputs.` condition check here?

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -46,12 +46,35 @@ int32_t get_broadcastTo_index_among_inputs_of(luci::CircleCustom *cop)
...
+    }
   }
 
   return bc_idx;
 }
-
-/** BEFORE
+/**
```
seanshpark, line 59:
```suggestion

/**
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,10 +17,203 @@
...
+namespace
+{
+
+/**
+ *  graph having Built-in operator BroadcastTo
+ *
+ *
+ *        [Input]   [BroadcastTo]
+ *             \         /
+ *             [CircleAdd]
```
seanshpark, line 35: Isn't this CircleCustom(AddV2) ?

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,10 +17,203 @@
...
+    BroadcastToAddGraphlet::init(g());
+
+    // connect graph
+    _addV2->inputs(0, input());
+
+    output(0)->from(_addV2_out);
+  }
+};
+
+class PassTest : public ::testing::Test
```
seanshpark, line 139: plz give a name for this test not `PassTest`

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
mhs4670go: @Seunghui98 Hmm.. I think there are two things to consider.

1. Using opcode

Dynamically casting two or more class make the check complicated. If you rather check their opcodes and use them with `must_cast`, the logic looks simpler.
```cxx
// ..
auto input = must_cast<const luci::CircleNode *>(cop->inputs(idx));
switch (input->opcode())
{
  case luci::CircleOpcode::BROADCAST_TO:
  {
    auto broadcastTo = loco::must_cast<luci::CircleBroadcastTo *>(input->input());
    // ..
  }
  case luci::CircleOpcode::CUSTOM
  {
    auto broadcastTo = loco::must_cast<luci::CircleCustom *>(input->input());
    // ..
  }
}
```

2. Introduce a function that checks the assumption

As two or more classes are being considered, assumption check might as well become a dedicated function. Otherwise, the logic would be complicated because current function does two things.

```cxx
int32_t get_broadcastTo_index_among_inputs_of(luci::CircleCustom *cop)
{
  if (num_of_broadcast_to(cop) != 1) return -1;
  // or
  if (not has_single_broadcast_to(cop)) return -1;
  
  // ..
}
```

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -73,19 +96,33 @@ bool resolve_with_BroadcastTo(luci::CircleCustom *addv2)
...
 
   auto name = addv2->name();
   assert(name.length() > 0);
 
   auto add = addv2->graph()->nodes()->create<luci::CircleAdd>();
   add->fusedActivationFunction(luci::FusedActFunc::NONE);
   add->x(addv2->inputs(1 - broadcastTo_idx));
-  add->y(broadcastTo->inputs(0));
+
+  // BroadcastTo(Builtin)
```
mhs4670go, line 108: Original logic does two things sequentially.

1. find where the broadast_to operator is;it returns false if there is no such op.
2. create a new add operator.

But, this PR is creating a add operator first, which could become obsolete if there is no op. So, getting the op first will be efficient.

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
Seunghui98: @seanshpark 

>how does This function assumes there is only one BroadcastTo node among its inputs. condition checked here?

no, there isn't condition checked. 😅
so, As in the comment given by @mhs4670go , I will introduce a function that pass can check condition of the inputs. :)
Thank you for letting me know !!

Seunghui98: @mhs4670go 

Thank you for your comments !!
I will reflect these in this PR and upload again. :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: @seanshpark 

okay.. I set `ResolveCustomOpAddPassTest` as the name in my local,
but the name was already used in the unit test in [here](https://github.com/Samsung/ONE/blob/31295c97fc8f10e630c105a31f131683a9ad4548/compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp#L21).
so, a conflict occurred in gtest. 😅
How about temporarily setting the name to `PassTest` in this PR,
and modifying it by uploading another PR in following order..?

1) Modify unit name test from `TEST` to `TEST_F` using `PassTest`. -> [here](https://github.com/Samsung/ONE/blob/31295c97fc8f10e630c105a31f131683a9ad4548/compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp#L21)

2) Change the name of `PassTest` to `ResolveCustomOpAddPassTest` and the related parts.

seanshpark: > How about temporarily setting the name to PassTest in this PR,

Fire a PR that changes existing one to something else if you want to keep `ResolveCustomOpAddPassTest` for the new ones.

Seunghui98: @seanshpark 

Okay !! I think the method you provided is less complicated, 
so I'll change the existing one first.
Thank you a lot !! :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: > Isn't this CircleCustom(AddV2) ?

yes. I missed CircleCustom(AddV2) step.
I will revise it. :)

Seunghui98:
@seanshpark @mhs4670go 
This PR(https://github.com/Samsung/ONE/pull/12527) was merged.
so, I revised this PR with the comments I received.

so, Could you review the code again? :)

Seunghui98:
@nnfw-bot test nncc-debug
@nnfw-bot test nncc-release

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -25,10 +25,42 @@
...
+        ++bt_cnt;
+        break;
+      }
+      case luci::CircleOpcode::CIRCLECUSTOMOUT:
+      {
+        auto inputOut = loco::must_cast<const luci::CircleCustomOut *>(cop->inputs(idx));
+        auto broadcastTo = loco::must_cast<luci::CircleCustom *>(inputOut->input());
+        if (broadcastTo->custom_code() == "BroadcastTo")
+          ++bt_cnt;
+        break;
```
seanshpark, line 48: this part looks duplicate with `get_broadcastTo_index_among_inputs_of`

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -46,12 +78,36 @@ int32_t get_broadcastTo_index_among_inputs_of(luci::CircleCustom *cop)
...
   return bc_idx;
 }
 
-/** BEFORE
+/**
+ * [Pattern1]
+ * BEFORE
+ *                                  [CircleConst]
+ *                                        |
+ *        [CircleNode]         [BroadcastTo(Builtin)]
```
seanshpark, line 97: pattern1 and patern2 looks misaligned. please check.

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -73,19 +129,39 @@ bool resolve_with_BroadcastTo(luci::CircleCustom *addv2)
...
+      auto inputOut =
+        loco::must_cast<const luci::CircleCustomOut *>(addv2->inputs(broadcastTo_idx));
+      auto broadcastTo = loco::must_cast<luci::CircleCustom *>(inputOut->input());
+      add->y(broadcastTo->inputs(0));
+      luci::add_origin(
+        add, luci::composite_origin({luci::get_origin(broadcastTo), luci::get_origin(addv2)}));
+      break;
+    }
+    default:
+      return false;
```
seanshpark, line 161: `auto add = addv2->graph()->nodes()->create<luci::CircleAdd>();` is useless for this case.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,129 @@
...
 namespace
 {
 
+/**
+ *    BroadcastTo-AddV2 graph pattern
+ *
+ *
+ *        [Input]   [BroadcastTo]
+ *             \         /
+ *         [CircleCustom(AddV2)]
```
seanshpark, line 35: Fix use one convention. `Circle` or not `Circle`

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
Seunghui98: @seanshpark 

> auto add = addv2->graph()->nodes()->create<luci::CircleAdd>(); is useless for this case.

I'm asking because I don't know 😅,
is there a case where `CircleAdd` comes in here?

seanshpark: > is there a case where CircleAdd comes in here?

what do you mean by comes in here?

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -61,7 +109,7 @@ int32_t get_broadcastTo_index_among_inputs_of(luci::CircleCustom *cop)
  *               [AddV2(CircleCustom)]
  *  AFTER
  *
- *         [CircleConst]         [CircleNode]
+ *         [CircleNode]         [CircleConst]
```
seanshpark, line 112: plz don't change what is not related.

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
Seunghui98: @seanshpark 

> what do you mean by comes in here?
At first, I misunderstood `CircleAdd` as input to switch-case statement. 😅

I revised the related code again so that `auto add = addv2->graph()->nodes()->create<luci::CircleAdd>();` can be declared and used. :)

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
Seunghui98: Okay, I will update it. 😅

Seunghui98: I updated it. :)

compiler/luci/pass/src/ResolveCustomOpAddPass.cpp
```diff
@@ -25,43 +25,91 @@
...
       auto broadcastTo = loco::must_cast<luci::CircleCustom *>(input->input());
       if (broadcastTo->custom_code() == "BroadcastTo")
-      {
-        // check if there is only one BroadcastTo node
-        if (bc_idx != -1)
-        {
-          return -1;
-        }
-        bc_idx = static_cast<int32_t>(idx);
-      }
```
Seunghui98, line 47: @seanshpark 

> this part looks duplicate with get_broadcastTo_index_among_inputs_of.

I removed the part that duplicate with `num_of_broadcast_to` from `get_broadcastTo_index_among_inputs_of `, 
is that okay ?

seanshpark: > ... is that okay ?

@mhs4670go PTAL

for me, I think there are multiple context of changes, it would be better to split this PR.

mhs4670go: Looks okay. Then, could you post a PR first that revises existing pass with `num_of_broadcast_to`?

Seunghui98: > Then, could you post a PR first that revises existing pass with num_of_broadcast_to?

yes! Yes. I will upload new PR first. :)

Seunghui98: @seanshpark @mhs4670go 

Another PR(https://github.com/Samsung/ONE/pull/12575) was merged first,
so Could you review this PR again? :)

Seunghui98:
@seanshpark @mhs4670go 

Another PR(https://github.com/Samsung/ONE/pull/12575) was merged,
so Could you review this PR again? :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,129 @@
...
+
+using namespace luci::test;
 
 namespace
 {
 
+/**
+ *    graph having Custom operator AddV2
+ *
+ *
```
seanshpark, line 32:
```suggestion
 *
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,129 @@
...
+ *        [Input]   [BroadcastTo]
```
seanshpark, line 33: There is `Custom(BroadcastTo)` and `BroadcastTo` depending on the test.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,129 @@
...
+public:
+  BroadcastToAddV2Graph() = default;
+
+  void init(void)
+  {
+    TestIGraphlet::init(g(), {2, 3});
+    TestOsGraphlet<1>::init(g(), {{2, 3}});
+    BroadcastToAddGraphlet::init(g());
+
+    // connect graph
```
seanshpark, line 132: this comment doesn't seem to help understanding the code

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,129 @@
...
+    _addV2->inputs(0, input());
+
```
seanshpark, line 134: `_custom_broadcastTo` , `_builtin_broadcastTo` connection is done in test, not here
adding a note about this would help understanding the code

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +150,69 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  _g.init();
+  _g.addV2()->inputs(1, _g.custom_broadcastTo_out());
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  auto broadcastTo_input = dynamic_cast<const luci::CircleConst *>(add->y());
```
seanshpark, line 165: `broadcastTo_input` ? it's casting with `CircleConst` 
this gives confusion.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +150,69 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  _g.init();
+  _g.addV2()->inputs(1, _g.builtin_broadcastTo());
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  auto broadcastTo_input = dynamic_cast<luci::CircleConst *>(add->y());
```
seanshpark, line 188: ditto

Seunghui98: @seanshpark 
Thank you for your comments!
I modified `input` to `input_0` and `broadcastTo_input` to `input_1` so as not to confuse. :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,136 @@
...
+
+#include <luci/IR/CircleNodes.h>
+
+using namespace luci::test;
 
 namespace
 {
 
+/**
+ *    graph having Custom operator AddV2
```
seanshpark, line 30: > graph having Custom operator AddV2

How does this sentence help understand the test?

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,136 @@
...
+ * [Pattern 1]
+ *        [Input]   [BroadcastTo]
+ *             \         /
+ *           [Custom(AddV2)]
+ *                  |
+ *             [CustomOut]
+ *                  |
+ *               [Output]
+ * [Pattern 2]
```
seanshpark, line 39:
```suggestion
 *               [Output]
 *
 * [Pattern 2]
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,136 @@
...
+    _custom_broadcastTo_out->dtype(loco::DataType::FLOAT32);
+    _custom_broadcastTo_out->index(0);
+    _custom_broadcastTo_out->input(_custom_broadcastTo);
+
+    // Builtin BroadcastTo
+    _builtin_broadcastTo = g->nodes()->create<luci::CircleBroadcastTo>();
+    _builtin_broadcastTo->dtype(loco::DataType::FLOAT32);
+    _builtin_broadcastTo->name("BroadcastTo");
+    _builtin_broadcastTo->input(input);
+    _builtin_broadcastTo->shape(shape);
```
seanshpark, line 102: above `_custom_broadcastTo` has empty line between.
match adding empty line or without empty line.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,136 @@
...
+public:
+  BroadcastToAddV2Graph() = default;
+
+  void init(void)
+  {
+    TestIGraphlet::init(g(), {2, 3});
+    TestOsGraphlet<1>::init(g(), {{2, 3}});
+    BroadcastToAddGraphlet::init(g());
+
+    // connect CircleInput to graph
```
seanshpark, line 139: maybe ?
```suggestion
    // connect Input and Output to test graph
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +157,73 @@ TEST_F(ResolveCustomOpAddPassTest, name)
   auto const name = _pass.name();
   ASSERT_NE(nullptr, name);
 }
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_CustomBroadcastTo)
+{
+  _g.init();
+
+  // connect Custom(BroadcastTo) to graph
```
seanshpark, line 165: word `graph` is also used here. not sure what `graph` means.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +157,73 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  auto input_0 = dynamic_cast<luci::CircleNode *>(add->x());
+  EXPECT_NE(nullptr, input_0);
+  EXPECT_EQ(2, input_0->dim(0));
+  EXPECT_EQ(3, input_0->dim(1));
+
+  auto input_1 = dynamic_cast<const luci::CircleConst *>(add->y());
+  EXPECT_NE(nullptr, input_1);
+  EXPECT_EQ(1, input_1->at<loco::DataType::FLOAT32>(0));
+  EXPECT_EQ(2, input_1->at<loco::DataType::FLOAT32>(1));
+  EXPECT_EQ(3, input_1->at<loco::DataType::FLOAT32>(2));
```
seanshpark, line 183: I can't cat what these checks are doing.
I would help understand with some explanation comments.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
seanshpark: My comment at https://github.com/Samsung/ONE/pull/12509#pullrequestreview-1855170135 suggested to change two empty lines to one but you've removed all two,
which makes readability drop.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +157,73 @@ TEST_F(ResolveCustomOpAddPassTest, name)
   auto const name = _pass.name();
   ASSERT_NE(nullptr, name);
 }
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_CustomBroadcastTo)
+{
+  _g.init();
+
+  // connect Custom(BroadcastTo) to graph
+  _g.addV2()->inputs(1, _g.custom_broadcastTo_out());
```
seanshpark, line 166: purpose of `init()` is to do connection of test graphlet and `TestIOGraphlet`s.
how about adding two `init()` methods for connecting `_custom_broadcastTo` or `_builtin_broadcastTo` ?

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: @seanshpark 
> How does this sentence help understand the test?

Referring to [here](https://github.com/Samsung/ONE/blob/90fe4e6d5b6a8c6246487971fd98cf96df863b0b/compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.test.cpp#L40), this sentence was written to explain the graph of AddV2 used in the test.
Do you think this setence needs further explanation?


> My comment at https://github.com/Samsung/ONE/pull/12509#pullrequestreview-1855170135 suggested to change two empty lines to one but you've removed all two, which makes readability drop.

Oh.... I will revise it and upload again. :)

seanshpark: > explain the graph of AddV2 used in the test.

I don't know how this explains about the test,
There is a AddV2... so what is it have any relation with the test itself?

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: @seanshpark 
> how about adding two init() methods for connecting _custom_broadcastTo or _builtin_broadcastTo ?

I think it's a good idea. :)
Then, Would it be better to make a function that is set to `Builtin` or `Custom` according to a argument?

seanshpark: > Then, Would it be better to make a function that is set to Builtin or Custom according to a argument?

That is also OK.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: @seanshpark 

well... Thinking about your comment, 
this sentence doesn't seem to be right to explain the test itself. :)

I thought of the following sentence to explain the test according to graph below,
is it okay..? :)

```
/**
<<<<<<<<<<<<<<<<<<<<<<<<<<<
 *   Resolved AddV2 graph in the test looks like below.
 <<<<<<<<<<<<<<<<<<<<<<<<<<<  
 * 
 * [Pattern 1]
 *        [Input]   [BroadcastTo]
 *             \         /
 *           [Custom(AddV2)]
 *                  |
 *             [CustomOut]
 *                  |
 *               [Output]
 * 
 * [Pattern 2]
 *        [Input]   [Custom(BroadcastTo)]
 *             \         /
 *           [Custom(AddV2)]
 *                  |
 *             [CustomOut]
 *                  |
 *               [Output]
 */
```

seanshpark: > Resolved AddV2 graph in the test looks like below.

`Test graph with Custom(AddV2) to resolve` ?

Seunghui98: @seanshpark 
Oh.. I think this sentence is a good idea 
because it's easier to understand the test  :)
I will reflect it and upload soon. Thank you a lot !

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: @seanshpark 

I reflected comments in this PR :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,149 @@
...
+    // CircleConst (BroadcastTo-shape)
+    auto shape = g->nodes()->create<luci::CircleConst>();
+    shape->dtype(loco::DataType::S32);
+    shape->shape({2});
+    shape->size<loco::DataType::S32>(2);
+    shape->at<loco::DataType::S32>(0) = 2;
+    shape->at<loco::DataType::S32>(1) = 3;
+
+    if (isCustomBroadcastTo)
+    {
```
seanshpark, line 83: I mean put these into `BroadcastToAddV2Graph::init()`

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
seanshpark: > // check if inputs are set well after changing from Custom(AddV2) to Add.

I mean what are the numbers like 0 and 1 in
```
  EXPECT_EQ(2, input_0->dim(0));
  EXPECT_EQ(3, input_0->dim(1));
```
and 0, 1, 2 in
```
  EXPECT_EQ(1, input_1->at<loco::DataType::FLOAT32>(0));
  EXPECT_EQ(2, input_1->at<loco::DataType::FLOAT32>(1));
  EXPECT_EQ(3, input_1->at<loco::DataType::FLOAT32>(2));
```
?

seanshpark: > set well

word `well` is ambiguous . plz be more specific.

Seunghui98: @seanshpark 

I changed `input_0` and `input_1` to `input_x` and `input_y` and the comment to make it easier to understand as follows.

```
  // check input_x and input_y to see if Custom(AddV2) is resolved.
  auto input_x = dynamic_cast<luci::CircleNode *>(add->x());
  EXPECT_NE(nullptr, input_x);
  EXPECT_EQ(2, input_x->dim(0));
  EXPECT_EQ(3, input_x->dim(1));

  auto input_y = dynamic_cast<const luci::CircleConst *>(add->y());
  EXPECT_NE(nullptr, input_y);
  EXPECT_EQ(1, input_y->at<loco::DataType::FLOAT32>(0));
  EXPECT_EQ(2, input_y->at<loco::DataType::FLOAT32>(1));
  EXPECT_EQ(3, input_y->at<loco::DataType::FLOAT32>(2));
```

And FYI, Because `CircleInput` and input of `BroadcastTo` change to `x` and `y` if it is resolved from `Custom(AddV2)` to `Add`, so check the value of each dim here.
Would it be better to add more comments to explain this? :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: I revised it. Thank you a lot :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
seanshpark: 1/ what's the difference of changing `input_0` to `intput_x` ?

2/ I asked about the numbers, not the variables.
to add more, why are the values `2, 3` and `1, 2, 3` compared?
what are the meanings of these number?

3/ what are the purpose of these tests?

Seunghui98: @seanshpark 

> 1/ what's the difference of changing input_0 to intput_x ?

There is no meaningful difference. 😅
but, It's just... To understand the test easily, the names of the variables are matched to the options `x` and `y` of `Add`.

> 2/ I asked about the numbers, not the variables.
to add more, why are the values 2, 3 and 1, 2, 3 compared?
what are the meanings of these number?

the value 2, 3 are the **shape** of `CircleInput`, 
and values 1, 2, 3 are the **values** of `BroadcastTo-input()`.
--->
```
    // CircleConst (BroadcastTo-input)
    auto input = g->nodes()->create<luci::CircleConst>();
    input->dtype(loco::DataType::FLOAT32);
    input->shape({1, 3});
    input->size<loco::DataType::FLOAT32>(3);
    input->at<loco::DataType::FLOAT32>(0) = 1;
    input->at<loco::DataType::FLOAT32>(1) = 2;
    input->at<loco::DataType::FLOAT32>(2) = 3;
```


> 3/ what are the purpose of these tests?
if `Custom(AddV2)` is resolved to `Add` through pass, because `CircleInput` and `BroadcastTo->input()` are connected to `Add`'s x and y, so I test to confirm this.

seanshpark: > the value 2, 3 are the shape of CircleInput,

really?

Seunghui98: @seanshpark 

> plz add some comments about this and use DEFINES not just 1, 2, 3 so that readers can understand what is check and why it is checked.

Okay, I will add comments as you said.
Thank you for your advice from the user's point of view all the time. :)


> really? why do we check this?
what's the point of pass?


Well... I'm sorry that I made a mistake of setting the shape in `CircleInput` instead of the values of input. 😂

Before making the test codes, I referred to the existing [ResolveCustomOpSplitVPass.test.cpp](https://github.com/Samsung/ONE/blob/356901d7a13c270252919834b581096e2cf35663/compiler/luci/pass/src/ResolveCustomOpSplitVPass.test.cpp#L113), [ResolveCustomOpBatchMatMulPass.test.cpp](https://github.com/Samsung/ONE/blob/356901d7a13c270252919834b581096e2cf35663/compiler/luci/pass/src/ResolveCustomOpBatchMatMulPass.test.cpp#L101), and i misunderstood that only the shape was set in `CircleInput`.
so, it seems that the test needs to be modified to **check the values of CircleInput** rather than to check the shape of CircleInput.

And this is the other part, 
As I thought about it while answering your question, this pass only contains the function of converting `Custom(AddV2)` to `Add`.
Therefore, the test only needs to check whether `x` and `y` of `Add` are set without checking the values. The code for this is as follow, Could I change it like this?

```
TEST_F(ResolveCustomOpAddPassTest, simple_test_CustomBroadcastTo)
{
  _g.init(true);

  auto ret = _pass.run(_g.g());
  EXPECT_EQ(true, ret);

  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
  EXPECT_NE(nullptr, add);

  <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
  // check input_x and input_y to see if Custom(AddV2) is resolved.
  // if the pass is run, CircleInput will be connected to x of add and BroadcastTo->input() will be connected to y of add.
  auto input_x = dynamic_cast<luci::CircleNode *>(add->x());
  EXPECT_NE(nullptr, input_x);

  auto input_y = dynamic_cast<const luci::CircleConst *>(add->y());
  EXPECT_NE(nullptr, input_y);
 <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
}
```

Seunghui98: @seanshpark 

> plz add some comments about this and use DEFINES not just 1, 2, 3 so that readers can understand what is check and why it is checked.

Okay, I will add comments as follows. Thank you a lot :)
```
  // check if inputs are set after changing from Custom(AddV2) to Add.
  // if the pass runs, x and y of Custom(AddV2) will be connected to x and y of Add.
```

> really? why do we check this?
what's the point of pass?

well... When the pass is executed, x and y of `Custom(AddV2)` are connected to x and y of `Add`. 
Then, since the shape and values of x and y are maintained even after they are connected, 
I think this should be checked.

seanshpark: >  and use DEFINES not just 1, 2, 3 so that

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +165,81 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  // check if inputs are set after changing from Custom(AddV2) to Add.
+  // if the pass runs, x and y of Custom(AddV2) will be connected to x and y of Add.
+  auto input_x = dynamic_cast<luci::CircleInput *>(add->x());
+  EXPECT_NE(nullptr, input_x);
+
+  EXPECT_EQ(2, input_x->dim(0));
+  EXPECT_EQ(3, input_x->dim(1));
```
seanshpark, line 185: why do we need to check these?

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +165,81 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+
+  auto input_y = dynamic_cast<const luci::CircleConst *>(add->y());
+  EXPECT_NE(nullptr, input_y);
+
+  EXPECT_EQ(1, input_y->dim(0));
+  EXPECT_EQ(3, input_y->dim(1));
```
seanshpark, line 191: ditto

seanshpark, line 191: @mhs4670go , as I know, BroadcastTo output shape should be changed to by doing broadcast to the shape input, but skipped doing so cause `Add` Op will do the broadcasting. I think we need to add some comments about this in the main code and here. What's your comment?

mhs4670go: You're right. That makes it easy to know about the pass.

Seunghui98: @mhs4670go 
> I think we need to add some NOTE comments about this in the main code and here. 
> You're right. That makes it easy to know about the pass.

I will add some NOTE comments to understand this. Thank you!

Seunghui98:
@seanshpark @mhs4670go 

After that offline discussion, I modified the code again.
1/ If this pass runs, x and y of Custom(AddV2) will be connected to x and y of Add.
So, I checked this using the pointer, not the shape or values of input.

2/ I added NOTE comments to know about broadcasting being skipped.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +170,61 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  // if this pass runs, x and y of Custom(AddV2) will be connected to x and y of Add.
+  // so, check if x and y of Custom(AddV2) Add are the equal to x and y of Add.
+  EXPECT_EQ(_g.input_x(), add->x());
```
seanshpark, line 186: Is this must have? or good to have?

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +170,61 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+
+  // When adding two input values of differenct shapes, Add performs broadcasting according to NumPy
+  // rules. so, because BroadcastTo is skipped, BraodcastTo->input() will be connected to y of Add.
```
seanshpark, line 189: This comment is something not what I expected.
Please fire a PR about this in another PR with a `NOTE` comment in Pass code.

Seunghui98: @seanshpark 
Okay. I will do that! :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: @seanshpark 

> Is this must have? or good to have?

I think this is good to have.
This is because i think we can check the following functions in the pass.

1/ x of Add is set to CircleInput. 
2/ y of Add is set to BroadcastTo->input().  


```
=========ResolveCustomOpAddPass.cpp==========
...
...
  add->x(addv2->inputs(1 - broadcastTo_idx));
  add->y(bc_input);
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: I uploaded a new PR related to this.:) https://github.com/Samsung/ONE/pull/12615

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
Seunghui98: Well noted. Thank you :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,55 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  auto addV2 = dynamic_cast<luci::CircleCustom *>(_g.addV2());
+  EXPECT_NE(nullptr, addV2);
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  // check if Custom(AddV2) is converted to Add after this pass is executed.
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+}
```
Seunghui98, line 184: @seanshpark @mhs4670go 

After the offline discussion, 
I modified the code so that this can test the main features of the pass (`Custom(AddV2)`->`Add`).

1/ if pass is not executed, AddV2 is Custom
2/ if pass is executed, Custom(AddV2) will be converted to Add.

Could you reivew this PR again? :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,55 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
```
Seunghui98, line 200: @seanshpark @mhs4670go 

After the offline discussion, 
I modified the code so that this can test the main features of the pass (`Custom(AddV2)`->`Add`).

1/ if pass is not executed, AddV2 is Custom
2/ if pass is executed, AddV2 will be converted to Add.

Could you reivew this PR again? :)

seanshpark: Test name is `simple_test_BuiltinBroadcastTo`
there is no test for builtin BroadcastTo op.
what is the intention of this test?

Seunghui98: @seanshpark 

> what is the intention of this test?

This is a test for the pass that converts 'AddV2' with **Builtin-BroadcastTo as input**  into Add. 😅
Come to think of it, the TestName doesn't seem to fit, so I'll change it to `simple_test_input_Custom_BroadcastTo`.

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  // check if Custom(AddV2) is converted to Add after this pass is executed.
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  // check if broadcasting of input `Const` is skipped.
+  auto input_y = dynamic_cast<luci::CircleConst *>(add->y());
+  EXPECT_NE(nullptr, input_y);
```
Seunghui98, line 187: @seanshpark @mhs4670go 

I reflected this after offline discussion. :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  EXPECT_NE(nullptr, addV2);
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  // check if Custom(AddV2) is converted to Add after this pass is executed.
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  // check if broadcasting of input `Const` is skipped.
```
seanshpark, line 205: what do you mean by `skipped` ?

Seunghui98: @seanshpark 

> what do you mean by skipped ?

This meaning is that BroadcastTo is removed, and only BroadcastTo`s input `Const` is used. :)

Seunghui98: I revised comments. Thank you for letting me know :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  EXPECT_NE(nullptr, addV2);
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  // check if Custom(AddV2) is converted to Add after this pass is executed.
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  // check if broadcasting is skipped and Add directly uses the `Const` input of broadcastTo.
```
seanshpark, line 185:
```suggestion
  // check if BroadcastTo is removed
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  _g.init(true);
+
+  // check if addV2 is Custom before this pass is executed.
+  auto addV2 = dynamic_cast<luci::CircleCustom *>(_g.addV2());
+  EXPECT_NE(nullptr, addV2);
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  // check if Custom(AddV2) is converted to Add after this pass is executed.
```
seanshpark, line 181:
```suggestion
  // check if Custom(AddV2) is converted to Add
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  // check if broadcasting is skipped and Add directly uses the `Const` input of broadcastTo.
```
seanshpark, line 205:
```suggestion
  // check if BroadcastTo is removed
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
   auto const name = _pass.name();
   ASSERT_NE(nullptr, name);
 }
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_CustomBroadcastTo)
+{
+  _g.init(true);
+
+  // check if addV2 is Custom before this pass is executed.
```
seanshpark, line 174:
```suggestion
  // check if Custom(AddV2) exist before the pass
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  // check if broadcasting is skipped and Add directly uses the `Const` input of broadcastTo.
+  auto input_y = dynamic_cast<luci::CircleConst *>(add->y());
+  EXPECT_NE(nullptr, input_y);
+}
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_BuiltinBroadcastTo)
+{
+  _g.init(false);
+
+  // check if AddV2 is Custom before this pass is executed.
```
seanshpark, line 194:
```suggestion
  // check if Custom(AddV2) exist before the pass
```

Seunghui98: I reflected these. Thank you a lot !! :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  EXPECT_NE(nullptr, addV2);
+
+  auto ret = _pass.run(_g.g());
+  EXPECT_EQ(true, ret);
+
+  // check if Custom(AddV2) is converted to Add
+  auto add = dynamic_cast<luci::CircleAdd *>(_g.output(0)->from());
+  EXPECT_NE(nullptr, add);
+
+  // check if Broadcasting is removed
```
seanshpark, line 185:
```suggestion
  // check if Custom(BroadcastTo) is removed
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
```
seanshpark, line 205:
```suggestion
  // check if BroadcastTo is removed
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  auto input_y = dynamic_cast<luci::CircleConst *>(add->y());
+  EXPECT_NE(nullptr, input_y);
+}
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_BuiltinBroadcastTo)
+{
+  _g.init(false);
+
+  // check if Custom(AddV2) exists before the pass
+  auto addV2 = dynamic_cast<luci::CircleCustom *>(_g.addV2());
```
seanshpark, line 195:
```suggestion
  auto addV2 = dynamic_cast<luci::CircleCustom *>(_g.output(0)->from());
  EXPECT_NE(nullptr, addV2);
  EXPECT_TRUE("AddV2" == addV2->custom_code());
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +166,63 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  EXPECT_NE(nullptr, addV2);
```
seanshpark, line 196:
```suggestion
  auto addV2 = dynamic_cast<luci::CircleCustom *>(_g.output(0)->from());
  EXPECT_NE(nullptr, addV2);
  EXPECT_TRUE("AddV2" == addV2->custom_code());
```

Seunghui98: I modified the code appropriately by reflecting your comment as follows. :)
```
// check if Custom(AddV2) exists before the pass
  auto addV2_out = dynamic_cast<luci::CircleCustomOut *>(_g.output(0)->from());
  auto addV2 = dynamic_cast<luci::CircleCustom *>(addV2_out->input());
  EXPECT_NE(nullptr, addV2);
  EXPECT_TRUE("AddV2" == addV2->custom_code());
```

Seunghui98:
```suggestion
   // check if Custom(AddV2) exists before the pass
  auto addV2_out = dynamic_cast<luci::CircleCustomOut *>(_g.output(0)->from());
  auto addV2 = dynamic_cast<luci::CircleCustom *>(addV2_out->input());
  EXPECT_NE(nullptr, addV2);
  EXPECT_TRUE("AddV2" == addV2->custom_code());
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,146 @@
...
+ */
+class BroadcastToAddGraphlet
+{
+public:
+  BroadcastToAddGraphlet() = default;
+
+public:
+  void init(loco::Graph *g)
+  {
+    // CircleCustom(AddV2)
```
seanshpark, line 58:
```suggestion
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,146 @@
...
+    // CircleConst (BroadcastTo-input)
+    _broadcastTo_input = g->nodes()->create<luci::CircleConst>();
+    _broadcastTo_input->dtype(loco::DataType::FLOAT32);
+    _broadcastTo_input->shape({1, 3});
+    _broadcastTo_input->size<loco::DataType::FLOAT32>(3);
+    _broadcastTo_input->at<loco::DataType::FLOAT32>(0) = 1;
+    _broadcastTo_input->at<loco::DataType::FLOAT32>(1) = 2;
+    _broadcastTo_input->at<loco::DataType::FLOAT32>(2) = 3;
+
+    // CircleConst (BroadcastTo-shape)
```
seanshpark, line 74:
```suggestion
    // Const as BroadcastTo shape
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,146 @@
...
+    auto broadcastTo_shape = g->nodes()->create<luci::CircleConst>();
+    broadcastTo_shape->dtype(loco::DataType::S32);
+    broadcastTo_shape->shape({2});
+    broadcastTo_shape->size<loco::DataType::S32>(2);
+    broadcastTo_shape->at<loco::DataType::S32>(0) = 2;
+    broadcastTo_shape->at<loco::DataType::S32>(1) = 3;
+
+    // Custom BroadcastTo
```
seanshpark, line 82:
```suggestion
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,146 @@
...
+    _custom_broadcastTo->inputs(1, broadcastTo_shape);
+
+    _custom_broadcastTo_out = g->nodes()->create<luci::CircleCustomOut>();
+    _custom_broadcastTo->custom_code("BroadcastTo");
+    _custom_broadcastTo_out->shape({2, 3});
+    _custom_broadcastTo_out->dtype(loco::DataType::FLOAT32);
+    _custom_broadcastTo_out->index(0);
+    _custom_broadcastTo_out->input(_custom_broadcastTo);
+
+    // Builtin BroadcastTo
```
seanshpark, line 99:
```suggestion
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,146 @@
...
+    _builtin_broadcastTo = g->nodes()->create<luci::CircleBroadcastTo>();
+    _builtin_broadcastTo->dtype(loco::DataType::FLOAT32);
+    _builtin_broadcastTo->name("BroadcastTo");
+
+    _builtin_broadcastTo->input(_broadcastTo_input);
+    _builtin_broadcastTo->shape(broadcastTo_shape);
+
+    // CircleCustomOut
```
seanshpark, line 107:
```suggestion
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -17,13 +17,146 @@
...
+    _addV2_out->shape({2, 3});
+    _addV2_out->dtype(loco::DataType::FLOAT32);
+    _addV2_out->index(0);
+    _addV2_out->input(_addV2);
+  }
+
+public:
+  luci::CircleCustom *addV2() { return _addV2; }
+  luci::CircleCustom *custom_broadcastTo() { return _custom_broadcastTo; }
+  luci::CircleCustomOut *custom_broadcastTo_out() { return _custom_broadcastTo_out; }
```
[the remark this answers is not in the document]
Seunghui98: @seanshpark 
I revised this PR base on your comments.
but, `addV2()` is used in the following `wrong_custom_code_NEG` test, so I left it. :)
```
TEST_F(ResolveCustomOpAddPassTest, wrong_custom_code_NEG)
{
  _g.init(false);

  _g.addV2()->custom_code("UNSUPORTED_CUSTOM_CODE");

  auto ret = _pass.run(_g.g());
  EXPECT_EQ(false, ret);
}
```

Thank you for your comments :)

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +161,67 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
 }
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_CustomBroadcastTo)
+{
+  _g.init(true);
+
+  // check if Custom(AddV2) exists before the pass
+  auto addV2_out = dynamic_cast<luci::CircleCustomOut *>(_g.output(0)->from());
+  auto addV2 = dynamic_cast<luci::CircleCustom *>(addV2_out->input());
+  EXPECT_NE(nullptr, addV2);
```
seanshpark, line 172:
```suggestion
  EXPECT_NE(nullptr, addV2);
  auto addV2 = dynamic_cast<luci::CircleCustom *>(addV2_out->input());
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +161,67 @@ TEST_F(ResolveCustomOpAddPassTest, name)
   auto const name = _pass.name();
   ASSERT_NE(nullptr, name);
 }
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_CustomBroadcastTo)
+{
+  _g.init(true);
+
+  // check if Custom(AddV2) exists before the pass
+  auto addV2_out = dynamic_cast<luci::CircleCustomOut *>(_g.output(0)->from());
```
seanshpark, line 170:
```suggestion
  auto addV2_out = dynamic_cast<luci::CircleCustomOut *>(_g.output(0)->from());
  EXPECT_NE(nullptr, addV2_out);
```

compiler/luci/pass/src/ResolveCustomOpAddPass.test.cpp
```diff
@@ -34,3 +161,67 @@ TEST_F(ResolveCustomOpAddPassTest, name)
...
+  auto input_y = dynamic_cast<luci::CircleConst *>(add->y());
+  EXPECT_NE(nullptr, input_y);
+}
+
+TEST_F(ResolveCustomOpAddPassTest, simple_test_BuiltinBroadcastTo)
+{
+  _g.init(false);
+
+  // check if Custom(AddV2) exists before the pass
+  auto addV2_out = dynamic_cast<luci::CircleCustomOut *>(_g.output(0)->from());
```
seanshpark, line 192:
```suggestion
  auto addV2_out = dynamic_cast<luci::CircleCustomOut *>(_g.output(0)->from());
  EXPECT_NE(nullptr, addV2_out);
```

seanshpark (approved):
LGTM

mhs4670go (approved):
LGTM

PR #12517, opened 2024-01-23
[tools/onert_train] Add virtual destructor for DataLoader

zetwhite:
This PR adds a virtual destructor for the onert_train's DataLoader class. Since DataLoader is base class, it should have virtual destructor.

### DIFF
--- a/tests/tools/onert_train/src/dataloader.h
+++ b/tests/tools/onert_train/src/dataloader.h
@@ -41,6 +41,7 @@ public:
   {
     // DO NOTHING
   }
+  virtual ~DataLoader() = default;
 
   virtual std::tuple<Generator, uint32_t>
   loadData(const uint32_t batch_size, const float from = 0.0f, const float to = 1.0f) = 0;


tests/tools/onert_train/src/dataloader.h
```diff
@@ -41,6 +41,7 @@ class DataLoader
   {
     // DO NOTHING
   }
+  virtual ~DataLoader() = default;
```
zetwhite, line 44: #### (for reviewers) Background of this PR 


I found a warning message while doing android-cross-build. 
So this PR is to suppress warning messages.

``` bash 
/home/ubuntu/android-sdk/ndk/20.0.5594570/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1/memory:2339:5: 

warning: delete called on 'onert_train::DataLoader' that is abstract but has non-virtual destructor
      [-Wdelete-non-virtual-dtor]
    delete __ptr;
    ^
```

YongseopKim: To check my understanding this, I leave a comment. 😄 (Nowadays I felt C++ is too hard for me...)

if there are some child classes that inherit to `DataLoader` that has no virtual destructor(dtor), as you already know, all of child classes calls parent's `DataLoader::destructor()`.

```cpp
// ChildDataLoader -> DataLoadr
// if DataLoadr has no destructor, 
// default and not virtual DataLoader's destructor is appended implicitly by compiler.
DataLoader *loader = new ChildDataLoader();
delete loader; // DataLoader::destructor() is called
```

Virtual destructor can make this right logic.
```cpp
// ChildDataLoader -> DataLoadr
// if DataLoadr has virtual destructor
DataLoader *loader = new ChildDataLoader();
delete loader; // ChildDataLoader::destructor() is called
```

YongseopKim: And, I have another question. Compiler's warning message cannot be permitted? or can be ignored? I'm asking about ONE's rule.

YongseopKim (approved):
LGTM 👍

tests/tools/onert_train/src/dataloader.h
jyoungyun: > Virtual destructors
> Deleting an object through pointer to base invokes undefined behavior unless the destructor in the base class is [virtual](https://en.cppreference.com/w/cpp/language/virtual):
> from https://en.cppreference.com/w/cpp/language/destructor

jyoungyun (approved):
LGTM

tests/tools/onert_train/src/dataloader.h
zetwhite: > Virtual destructor can make this right logic.

exactly right!

zetwhite: >  Compiler's warning message cannot be permitted? or can be ignored? I'm asking about ONE's rule.

@YongseopKim 

I put my understanding here. 
Since I'm a newbie to CMake, there could be some misunderstanding. 


There is interface named `nnfw_common`, this interface makes it warnings into error. 
(`-Werror` options force it. [refer](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#:~:text=%2D-,Werror,-%C2%B6)  ) 

https://github.com/Samsung/ONE/blob/31295c97fc8f10e630c105a31f131683a9ad4548/infra/nnfw/CMakeLists.txt#L99-L102  


So, the component(library) with `nnfw_common` does not permit any warning.
Whereas, the component without `nnfw_common` permits a warning. (in other words, just ignore the warning)
Since `tools/onert_train` doesn't note `nnfw_common`, `onert_train` has ignored the warning messages.

YongseopKim: I got it, but Why hasn't there been a problem so far? Is it because of NDK compiler?
- current gcc compiler doesn't make warning about this
- only NDK compiler makes warning about this

Is it right?

zetwhite: I guess so. 
But not 100% sure.

YongseopKim: By the way, I am asking this because it is a compiler feature that prevents us from making errors easily. If we can prevent it by turning on the compiler warning option, it would be better to turn it on in the GCC as well.

But I'm not sure whetehr gcc can do it or not. 

/cc @hseok-oh @ragmani

PR #12499, opened 2024-01-18
draft for trainable pad op

YongseopKim:
#12413 

ONE-DCO-1.0-Signed-off-by: Yongseop Kim <yons.kim@samsung.com>

Issue #12413, opened 2024-01-08
[onert] Introduce Pad op for training


YongseopKim:
Parent issue: https://github.com/Samsung/ONE/issues/12325

Let's introduce Pad op for training.

YongseopKim:
`transpose` -> `Pad`

ONERT_LOG_ENABLE_1
```
[   GraphDumper  ] {
[   GraphDumper  ]     %3 =   @0_Conv2D(  %0,  %1,  %2)
[   GraphDumper  ]     %6 =   @1_DepthwiseConv2D(  %3,  %4,  %5)
[   GraphDumper  ]     %9 =   @2_Conv2D(  %6,  %7,  %8)
[   GraphDumper  ]    %12 =   @3_Conv2D(  %9, %10, %11)
[   GraphDumper  ]    %14 =   @4_Pad( %12, %13)
[   GraphDumper  ]    %17 =   @5_DepthwiseConv2D( %14, %15, %16)
[   GraphDumper  ]    %20 =   @6_Conv2D( %17, %18, %19)
[   GraphDumper  ]    %23 =   @7_Conv2D( %20, %21, %22)
[   GraphDumper  ]    %26 =   @8_DepthwiseConv2D( %23, %24, %25)
[   GraphDumper  ]    %29 =   @9_Conv2D( %26, %27, %28)
[   GraphDumper  ]    %30 =  @10_Add( %20, %29)
[   GraphDumper  ]    %33 =  @11_Conv2D( %30, %31, %32)
[   GraphDumper  ]    %34 =  @12_Pad( %33, %13)
[   GraphDumper  ]    %37 =  @13_DepthwiseConv2D( %34, %35, %36)
[   GraphDumper  ]    %40 =  @14_Conv2D( %37, %38, %39)
[   GraphDumper  ]    %43 =  @15_Conv2D( %40, %41, %42)
[   GraphDumper  ]    %46 =  @16_DepthwiseConv2D( %43, %44, %45)
[   GraphDumper  ]    %49 =  @17_Conv2D( %46, %47, %48)
[   GraphDumper  ]    %50 =  @18_Add( %40, %49)
[   GraphDumper  ]    %53 =  @19_Conv2D( %50, %51, %52)
[   GraphDumper  ]    %56 =  @20_DepthwiseConv2D( %53, %54, %55)
[   GraphDumper  ]    %59 =  @21_Conv2D( %56, %57, %58)
[...]

PR #12522, opened 2024-01-24
[cker/train] Introduce Pad op in train

YongseopKim:
This commit introduces Pad op in cker/train.

---

for: https://github.com/Samsung/ONE/issues/12413
draft: https://github.com/Samsung/ONE/pull/12499

### DIFF
--- a/compute/cker/include/cker/Types.h
+++ b/compute/cker/include/cker/Types.h
@@ -397,6 +397,12 @@ struct LeakyReluParams
   float alpha;
 };
 
+struct PadParams
+{
+  int32_t data[8];
+  int32_t rank;
+};
+
 enum class Order
 {
   kColMajor,
--- a/compute/cker/include/cker/train/operation/Pad.h
+++ b/compute/cker/include/cker/train/operation/Pad.h
@@ -54,11 +54,13 @@ inline void Depad(const int32_t *padding_data, int32_t pad_rank, const Shape &in
   using PaddingList = std::vector<PaddingInfo>;
 
   assert(output_shape.DimensionsCount() == input_shape.DimensionsCount());
+  assert(output_shape.DimensionsCount() == pad_rank);
 
   PaddingList padding_list(pad_rank);
   for (int32_t n = 0; n < pad_rank; ++n)
   {
     const int32_t *from = padding_data + (n * 2);
+    assert(from[0] >= 0 && from[1] >= 0);
     padding_list[n] = {from[0], from[1]};
   }
   for (int32_t i = 0; i < pad_rank; ++i)
@@ -118,6 +120,7 @@ inline void Depad(const int32_t *padding_data, int32_t pad_rank, const Shape &in
       }
       break;
     }
+    // NOTE: Assume that tensors' memory format is NHWC. In cker, it cannot be checked.
     case 4:
     {
       const int32_t out_cube = output_shape.Dims(0);
--- a/compute/cker/include/cker/train/operation/Pad.h
+++ b/compute/cker/include/cker/train/operation/Pad.h
@@ -36,19 +36,9 @@ namespace train
  *             ->    [C,2,3,C]
  *             ->    [C,C,C,C]
  */
-template <typename T>
-inline void Pad(const int32_t *padding_data, int32_t pad_rank, const Shape &input_shape,
-                const T *input_data, const Shape &output_shape, T *output_data,
-                const T *constant_value_data)
-{
-  // Use nnfw::cker::Pad directly
-  nnfw::cker::Pad<T>(padding_data, pad_rank, input_shape, input_data, output_shape, output_data,
-                     constant_value_data);
-}
-
 /*
- * input_data(backward_output_data) will be transformed by backward of PAD operation with padding
- * options to output_data(backward_input_data)
+ * input_data(backward_output_data) will be transformed by backward of PAD operation (Depad) with
+ * padding options to output_data(backward_input_data)
  *
  * input_data(backward_output_data)  ->  output_data(backward_input_data)
  *   [C,C,C,C]                       ->    [0,1]
@@ -119,10 +109,9 @@ inline void Depad(const int32_t *padding_data, int32_t pad_rank, const Shape &in
       {
         for (auto h = 0; h < out_height; ++h)
         {
-          // TODO: Remove unnecessary literal value 0
           const auto in_offset =
-            (d + padding_depth) * in_plain_size + (h + padding_top) * in_width + (0 + padding_left);
-          const auto out_offset = (d * out_plain_size) + (h * out_width) + (0);
+            (d + padding_depth) * in_plain_size + (h + padding_top) * in_width + (padding_left);
+          const auto out_offset = (d * out_plain_size) + (h * out_width);
           // copy a row of input data to output data
           std::memcpy(output_data + out_offset, input_data + in_offset, out_width * sizeof(T));
         }
@@ -152,12 +141,10 @@ inline void Depad(const int32_t *padding_data, int32_t pad_rank, const Shape &in
         {
           for (auto h = 0; h < out_height; ++h)
           {
-            // TODO: Remove unnecessary literal value 0
             const auto in_offset = (c + padding_cube) * in_cube_size +
                                    (d + padding_depth) * in_plain_size +
-                                   (h + padding_top) * in_width + (0 + padding_left);
-            const auto out_offset =
-              (c * out_cube_size) + (d * out_plain_size) + (h * out_width) + (0);
+                                   (h + padding_top) * in_width + (padding_left);
+            const auto out_offset = (c * out_cube_size) + (d * out_plain_size) + (h * out_width);
             // copy a row of input data to output data
             std::memcpy(output_data + out_offset, input_data + in_offset, out_width * sizeof(T));
           }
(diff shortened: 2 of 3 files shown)


compute/cker/include/cker/train/operation/Pad.h
```diff
@@ -0,0 +1,181 @@
...
+ *             ->    [C,C,C,C]
+ */
+template <typename T>
+inline void Pad(const int32_t *padding_data, int32_t pad_rank, const Shape &input_shape,
+                const T *input_data, const Shape &output_shape, T *output_data,
+                const T *constant_value_data)
+{
+  // Use nnfw::cker::Pad directly
+  nnfw::cker::Pad<T>(padding_data, pad_rank, input_shape, input_data, output_shape, output_data, constant_value_data);
+}
```
jyoungyun, line 45: IMHO, there is no need to re-define this function. We can use `nnfw::cker::Pad` directly in train backend. It makes unnecessary function call and we have to manage two set of code (Pad in train/operation and Pad in cpu/operation). Also it makes binary size bigger. (Of course, it's a very trivial issue. 😄)

compute/cker/include/cker/train/operation/Pad.h
```diff
@@ -0,0 +1,178 @@
...
+ * output_data
+ *
+ * input_data  ->  output_data
+ *   [0,1]     ->    [C,C,C,C]
+ *   [2,3]     ->    [C,0,1,C]
+ *             ->    [C,2,3,C]
+ *             ->    [C,C,C,C]
+ */
+template <typename T>
+inline void Pad(const int32_t *padding_data, int32_t pad_rank, const Shape &input_shape,
```
YongseopKim, line 40: My new patch overwrites @jyoungyun 's comment. I write here again instead of @jyoungyun .

https://github.com/Samsung/ONE/pull/12522#discussion_r1464521239

> IMHO, there is no need to re-define this function. We can use nnfw::cker::Pad directly in train backend. It makes unnecessary function call and we have to manage two set of code (Pad in train/operation and Pad in cpu/operation). Also it makes binary size bigger. (Of course, it's a very trivial issue. 😄)

YongseopKim: I agree with @jyoungyun . I'm going to use `nnfw::ckwer::Pad` in train backend.

compute/cker/include/cker/train/operation/Pad.h
YongseopKim: https://github.com/Samsung/ONE/pull/12522/files#r1465679532

YongseopKim:
PTAL @jyoungyun and @Samsung/one_onert

compute/cker/include/cker/Types.h
```diff
@@ -397,6 +397,12 @@ struct LeakyReluParams
   float alpha;
 };
 
+struct PadParams
+{
+  int32_t data[8];
+  int32_t rank;
+};
```
jyoungyun, line 404: 👍 

It would be good to use this param in cpu backend too.

jyoungyun (approved):
LGTM

compute/cker/include/cker/train/operation/Pad.h
```diff
@@ -0,0 +1,165 @@
...
+  for (int32_t n = 0; n < pad_rank; ++n)
+  {
+    const int32_t *from = padding_data + (n * 2);
+    padding_list[n] = {from[0], from[1]};
+  }
+  for (int32_t i = 0; i < pad_rank; ++i)
+  {
+    assert(output_shape.Dims(i) ==
+           input_shape.Dims(i) - padding_list[i].first - padding_list[i].second);
+  }
```
zetwhite, line 68: May I ask - what is stored in `padding_data`? 

It looks like it saves `padding_data` saves some padding info.

* padding_info[i] : i-th dimension of padding info
   * padding_info[i].first : padding of top
   * padding_info[i].second : padding-left

compute/cker/include/cker/train/operation/Pad.h
```diff
@@ -0,0 +1,165 @@
...
+
+  switch (pad_rank)
+  {
+    case 0:
+    case 1:
+    {
+      const int32_t out_width = output_shape.Dims(0);
+      const int32_t padding_left = padding_list[0].first;
+      std::memcpy(output_data, input_data + padding_left, out_width * sizeof(T));
```
zetwhite, line 77: Does `input_data` and `output_data` memory layout is in order of `HWD`?

YongseopKim: That's a good point. Let me check it!

YongseopKim: After examining the Pad layer code, I have discovered that the Pad layer does not take into account any channel order of a rank 4 tensor but only implements NHWC tensor. It is necessary to add an assertion code that it must be NHWC.

YongseopKim: And @zetwhite discussed about padding rank, I realize that padding rank should be checkecd.

zetwhite: Aha, My curiosity is somehow resolved by your offline explaination.

If `plan_rank` = 1 assure that 'input_shape rank is also 1', 
I think there is no problem with code :)

YongseopKim: I'll upload a new commit with two patches
1. check NHWC at rank-4
2. check pad rank

YongseopKim: > check NHWC at rank-4

I realize that this cannot be checked in cker level. I'll add code about that into backend code of PadLayer soon.

YongseopKim:
PTAL

compute/cker/include/cker/train/operation/Pad.h
```diff
@@ -0,0 +1,168 @@
...
+        for (auto d = 0; d < out_depth; ++d)
+        {
+          for (auto h = 0; h < out_height; ++h)
+          {
+            const auto in_offset = (c + padding_cube) * in_cube_size +
+                                   (d + padding_depth) * in_plain_size +
+                                   (h + padding_top) * in_width + (padding_left);
+            const auto out_offset = (c * out_cube_size) + (d * out_plain_size) + (h * out_width);
+            // copy a row of input data to output data
+            std::memcpy(output_data + out_offset, input_data + in_offset, out_width * sizeof(T));
```
zetwhite, line 152: I'm a bit afraid that I might be wrong.

But this offset calculation looks like `NCHW`. 
```
const auto in_offset = (c + padding_cube) * in_cube_size +     // n * CHW
                                   (d + padding_depth) * in_plain_size +    // c * HW 
                                   (h + padding_top) * in_width                   // h * W 
                                   + padding_left                                         // w 
``` 

I referred this : https://oneapi-src.github.io/oneDNN/dev_guide_understanding_memory_formats.html#nchw

YongseopKim: IHMO, in cker, the logic here doesn't matter NCHW/NHWC. In the view of Pad kernel, it just does copying. If it might be NCHW, the code var name will be changed.

If I'm wrong, I'll leave a comment.

zetwhite: > IHMO, in cker, the logic here doesn't matter NCHW/NHWC.  

I agree somehow. 

IMO, layout in onert(`ir::Layout::NCHW`) just means`nnfw::cker::Shape`'s element order. 
In the tensor level, offset-calculation can be different. 

( I feel similar things while working on MaxPool cker, before )

YongseopKim: Okay now I'm sure somewhat.

At the train backend level, tensors' layer assume NHWC.
```c++
// backend/train/Config.cc
ir::Layout Config::supportLayout(const ir::IOperation &, ir::Layout) { return ir::Layout::NHWC; }

// backend/train/BackendContext.cc
backend::train::ITensorRegistry *BackendContext::genTrainingTensors()
{
  const ir::train::TrainableGraph &tgraph = *trainable_graph();
  auto tensor_builder = _tensor_builder;

  tgraph.operands().iterate([&](const ir::OperandIndex &ind, const ir::Operand &obj) {
    if (external_operands().contains(ind))
      return;
    // NOTE Assuming there is no layout changes (Always assume NHWC or UNKNOWN)
    assert(tgraph.layout() != ir::Layout::NCHW);

    // TODO Different shape of back propagation tensor
    ir::OperandInfo backend_info{obj.shape(), obj.typeInfo(), obj.info().memAllocType(),
                                 obj.isConstant()};
    tensor_builder->registerBackwardTensorInfo(ind, backend_info, ir::Layout::NHWC);
  });

  // TODO Plan tensor builds to reduce peak memory usage
  tgraph.operands().iterate([&](const ir::OperandIndex &ind, const ir::Operand &) {
    if (tensor_builder->isRegisteredBackward(ind))
      tensor_builder->notifyBackwardFirstUse(ind);
  });

  tensor_builder->allocateBackward();

  return _tensor_registry.get();
}
```

Therefore, at the rank-4 tensor in cker, it would be handled as NHWC.

I fixed words from specific width/height/depth to abstract(less-specific) row/col/plain. This means that CHW/HWC is not important point in cker.

YongseopKim:
PTAL @jyoungyun and @zetwhite

zetwhite (approved):
LGTM 👍

YongseopKim:
PTAL @jyoungyun

jyoungyun (approved):
LGTM

YongseopKim:
ping @Samsung/one_onert

compute/cker/include/cker/train/operation/Pad.h
zetwhite: Aha, I think I also got it. 
Thank you  :+1:

Issue #12520, opened 2024-01-24
[onert] Let's remove default value from onert_train/args


zetwhite:
# Background

### Motivation1
For now, `onert_train` can set training info in 2 ways. 
* using model file (by circle file metadata) 
* using onert_train command line input

Since command line input always has a default value, `onert_train` always choose to use the cmd line info instead of choosing the model file's.

So, Let's 
  * remove the default value(info)
  * make it choose to use the model file training info, if training info is not given from cmd line.  

### Motivation2

While working on motivation1, I found out that 
 * `onert_train/args` just save an integer value
 * `onert_train` main does parsing - from an integer value to the corresponding enum `NNFW_TRAIN_*`.  

Since `onert_train/args` holds an integer value, It is hard to remove the default value. More exactly, it is hard to represent that 'nothing is given'.  

Additionally parsing (integer to NNfW_TRAIN enum) is not a kind of training logic, IMO it is better to remove it from the main. 

---------------------

# Suggestion

In such background, I suggest this direction: https://github.com/Samsung/ONE/pull/12445 

### AS-IS

* `onert_train/args`: hold an integer value
* `onert_train/onert_train`: parse integer value to `NNFW_TRAIN_*` enum and always overwrite session's training info. 

### TO-BE 
* `onert_train/args` : directly hold an enum `NNFW_TRAIN_*` and to represent nothing is given use `std::optional`. 
* `onert_train/args` : check cmd line parameter is given or not, and then decide whether to overwrite the session's training info or not. 

### Side effects of my suggestion 
To merge the draft, it expects some additional changes :  
  * Since `std::optional` is introduced in C++17, C++17 needs to be used to compile `onert_train`.
  * Since `Boost v1.58.0` used in android-build does not support C++17, boost version has to be upgraded. 
      * Boost supports C++17 after v1.65.0, But the latest Boost supports various compilers including both Clang and GCC.  
      * So I'd lie to upgrade to v1.84.0 which is the latest. 
      * https://github.com/Samsung/ONE/pull/12445#discussion_r1463308097

zetwhite:
@Samsung/one_onert 

Could you briefly check draft #12445 and leave any opinion about the change? 
I'd like to ask if it is fine to use C++17 for onert_train and upgrade the boost version to the latest.

jyoungyun:
> I'd like to ask if it is fine to use C++17 for onert_train and upgrade the boost version to the latest.

👍

zetwhite:
[...]

PR #12528, opened 2024-01-25
[tools/onert_train] Refactor print logic of train info

zetwhite:
This PR refactors print logic of training information. 
It introduces NNFW_TRAIN_* enum to string map, which is referred when dump(print) function called.

-------------------------

draft : https://github.com/Samsung/ONE/pull/12445 
in the process to resolve : https://github.com/Samsung/ONE/issues/12520

### DIFF
--- a/tests/tools/onert_train/src/nnfw_util.cc
+++ b/tests/tools/onert_train/src/nnfw_util.cc
@@ -50,7 +50,7 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
   return elmsize[ti->dtype] * num_elems(ti);
 }
 
-std::string toString(const NNFW_TRAIN_OPTIMIZER &opt)
+std::string toString(NNFW_TRAIN_OPTIMIZER opt)
 {
   static const std::unordered_map<NNFW_TRAIN_OPTIMIZER, std::string> name_map{
     {NNFW_TRAIN_OPTIMIZER_UNDEFINED, "undefined"},
@@ -60,7 +60,7 @@ std::string toString(const NNFW_TRAIN_OPTIMIZER &opt)
   return name_map.at(opt);
 }
 
-std::string toString(const NNFW_TRAIN_LOSS &loss)
+std::string toString(NNFW_TRAIN_LOSS loss)
 {
   static const std::unordered_map<NNFW_TRAIN_LOSS, std::string> name_map{
     {NNFW_TRAIN_LOSS_UNDEFINED, "undefined"},
@@ -70,7 +70,7 @@ std::string toString(const NNFW_TRAIN_LOSS &loss)
   return name_map.at(loss);
 }
 
-std::string toString(const NNFW_TRAIN_LOSS_REDUCTION &loss_rdt)
+std::string toString(NNFW_TRAIN_LOSS_REDUCTION loss_rdt)
 {
   static const std::unordered_map<NNFW_TRAIN_LOSS_REDUCTION, std::string> name_map{
     {NNFW_TRAIN_LOSS_REDUCTION_UNDEFINED, "undefined"},
--- a/tests/tools/onert_train/src/nnfw_util.cc
+++ b/tests/tools/onert_train/src/nnfw_util.cc
@@ -17,6 +17,7 @@
 #include <cassert>
 #include <string>
 #include <iostream>
+#include <unordered_map>
 
 #include "nnfw.h"
 #include "nnfw_experimental.h"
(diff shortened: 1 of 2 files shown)


tests/tools/onert_train/src/nnfw_util.cc
```diff
@@ -49,63 +50,50 @@ uint64_t bufsize_for(const nnfw_tensorinfo *ti)
...
+
+std::string toString(const NNFW_TRAIN_LOSS_REDUCTION &loss_rdt)
+{
+  static const std::unordered_map<NNFW_TRAIN_LOSS_REDUCTION, std::string> name_map{
+    {NNFW_TRAIN_LOSS_REDUCTION_UNDEFINED, "undefined"},
+    {NNFW_TRAIN_LOSS_REDUCTION_SUM_OVER_BATCH_SIZE, "sum over batch size"},
+    {NNFW_TRAIN_LOSS_REDUCTION_SUM, "sum"}};
+  return name_map.at(loss_rdt);
+}
+
```
zetwhite, line 81: Another motivation to use a map is here: https://github.com/Samsung/ONE/pull/12412#discussion_r1444102192 

/cc @ragmani @YongseopKim

tests/tools/onert_train/src/nnfw_util.h
```diff
@@ -36,6 +36,9 @@ namespace onert_train
 uint64_t num_elems(const nnfw_tensorinfo *ti);
 uint64_t bufsize_for(const nnfw_tensorinfo *ti);
 
+std::string toString(const NNFW_TRAIN_OPTIMIZER &opt);
+std::string toString(const NNFW_TRAIN_LOSS &loss);
+std::string toString(const NNFW_TRAIN_LOSS_REDUCTION &loss_rdt);
```
zetwhite, line 41: I have a plan to use this function later, here : 

 https://github.com/Samsung/ONE/pull/12445#r1466143798

zetwhite: Oh, I didn't think much about this part. 

Either way(passing by copy, or cost reference) looks good to me. 

-------------------------------------------------------

IHMO, 
It is trivial but one good point of using `const reference` is that we can give consistency to the `toString` function. 

If we want to add `toString` function of another class/struct(not enum), later. 
We might make a function like `string toString(const type&)` to avoid an additional copy.
I think it is a kind of consistency - that a similar purpose function has a similar definition.

jyoungyun (dismissed):
LGTM

tests/tools/onert_train/src/nnfw_util.h
YongseopKim: > enum is okay to be copied

means that it's like `toString(int32_t loss_rdt);`. It's just copying 4 bytes memory. That's why I said **okay**. 😄

It depends on you's idea. For me, I cannot feel satisfied with the following code.
```cpp
// for consistency of function signature
// Is the following also okay?
std::string toString(const bool &sth_bool);
```

YongseopKim: Of course, it's a trivial.

YongseopKim (approved):
LGTM

tests/tools/onert_train/src/nnfw_util.h
zetwhite: > // for consistency of function signature
> // Is the following also okay? 
> std::string toString(const bool &sth_bool);  

Hmm.. Definitely, it looks awkward. 
Your point makes sense :smile: 

I'd better update to remove the reference here.

zetwhite:
@YongseopKim @jyoungyun PTAL

tests/tools/onert_train/src/nnfw_util.h
```diff
@@ -36,6 +36,9 @@ namespace onert_train
 uint64_t num_elems(const nnfw_tensorinfo *ti);
 uint64_t bufsize_for(const nnfw_tensorinfo *ti);
 
+std::string toString(NNFW_TRAIN_OPTIMIZER opt);
+std::string toString(NNFW_TRAIN_LOSS loss);
+std::string toString(NNFW_TRAIN_LOSS_REDUCTION loss_rdt);
```
zetwhite, line 41: I have a plan to use this function later, here : 

 https://github.com/Samsung/ONE/pull/12445#r1466143798

YongseopKim (approved):
LGTM 👍

jyoungyun (approved):
LGTM

ragmani (approved):
LGTM

Issue #12505, opened 2024-01-22
[onert] Introduce on-device codegen feature


jyoungyun:
## What

Let's introduce on-device codegen feature.

This issue is a part of United AI framework.
Related issue: #12491 

Draft: #12504 

### Task

onert
- [x] #12531
- [x] #12536 
- [x] #12553
- [x] #12583 

onert_run
- [x] #12537
- [x] #12584

jyoungyun:
Finally Done!

PR #12504, opened 2024-01-22
[Draft] [onert] Introduce on-device codegen feature

jyoungyun:
This commit adds compliation APIs to nnfw_experimental.h file.

ONE-DCO-1.0-Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>
Co-authored-by: Hyeongseok Oh <hseok82.oh@samsung.com>
Co-authored-by: Jiyoung Yun <jy910.yun@samsung.com>

PR #12531, opened 2024-01-25
[onert] Introduce ICodegen class

jyoungyun:
This commit introduces ICodegen class.
The backend target compiler and Codegen loader will refer to this file together and will be implemented with the same structure.

Co-authored-by: Sanggyu Lee <sg5.lee@samsung.com>

Related issue: #12505 
Draft PR: #12504

### DIFF
--- a/runtime/onert/core/include/odc/ICodegen.h
+++ b/runtime/onert/core/include/odc/ICodegen.h
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2024 Samsung Electronics Co., Ltd. All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __ONERT_ODC_ICODEGEN_H__
+#define __ONERT_ODC_ICODEGEN_H__
+
+// NOTE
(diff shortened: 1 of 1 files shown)


runtime/onert/core/include/odc/ICodegen.h
```diff
@@ -0,0 +1,50 @@
...
+{
+namespace odc
+{
+
+class ICodegen
+{
+public:
+  virtual ~ICodegen() = default;
+
+  virtual int codegen(const char *in, const char *out) = 0;
```
YongseopKim, line 44: I'm not sure if I fully understand on-device-compile(r), but let me review it. 😄
1. `in` and `out` are filename? and the file name must contain extension?
2. I recommend `std::string` than `const char *` because this code is in c++
3. returning value is int. This is too c-stylish. How about throwing or enum?

YongseopKim: I discussed with @jyoungyun about it, and I obtained the C-style code in C++ that was already placed in the 'odc' directory. Therefore, I now agree with it.

However, there are some points in the source code within this 'odc' directory that need to be refactored. In the future, we will do so after reaching a consensus.

YongseopKim (approved):
LGTM

runtime/onert/core/include/odc/ICodegen.h
jyoungyun: @YongseopKim 

> 1. `in` and `out` are filename? and the file name must contain extension?

Yes, they are filename string. There are no rules for extension. This function just loads the specified input filename and saves data in the specified output filename regardless of the file extension.

> 2. I recommend `std::string` than `const char *` because this code is in c++

I want to merge this code as it is. This is because I want to maintain the same structure as Quantizer code. If we need, I will do refactoring later.

> 3. returning value is int. This is too c-stylish. How about throwing or enum?

This code is implemented in the C library. Therefore, it returns Errno as the c-style retrun value. If we define enum, we need to manage another enum between the two modules. It would be better if there is less data shared between the two.

hseok-oh (approved):
LGTM
I agree to change to use `std::string` in `odc` later.

